PR #853, opened 2015-01-17
Code will be mounted, no need to copy files

ibnesayeed:
Copying files from the current directory to container's code directory and mounting the same directory afterwards makes no sense to me.

Signed-off-by: Sawood Alam <ibnesayeed@gmail.com>

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -78,7 +78,7 @@ We define our Python dependencies in a file called `requirements.txt`:
 Next, we want to create a Docker image containing all of our app's dependencies. We specify how to build one using a file called `Dockerfile`:
 
     FROM python:2.7
-    ADD . /code
+    RUN mkdir /code
     WORKDIR /code
     RUN pip install -r requirements.txt
 


thaJeztah:
I *think* (but I agree, the docs could be clearer on this), this is to support this workflow;

- When *building* the container, your code is added to the image/container. For "production" you are able to deploy the image as-is, no need to mount a volume.
- During *development*, you mount your local files *over* the files in the container. You can edit your local files (the files in the volume), without having to re-build the container on each change.
- Once the changes are ready, re-build the container (which copies the current version of your files *into* the image), and deploy the image to production, where it is run *without* a volume/bind-mount, so that it uses the files inside the image.

Does it make sense that way?

ibnesayeed:
@thaJeztah I got your point, but in that case a separate `fig.yml` file will be required for deployment, right? Even if it was intentional the way it is, I think getting started guides should not add confusing stuff unless explained.

thaJeztah:
Yup, I'm with you there. Perhaps a separate section should be added to explain that workflow.

bfirsh:
Thanks @ibnesayeed. As @thaJeztah said, this is so you can use the same image in production. We're going to improve the docs around how Compose can be used in production, see #944.

Issue #836, opened 2015-01-12
Fig caching mounted volumes


jgadbois:
After updating fig, I'm having a problem where fig seems to be caching the local working directory and using it's own version instead of using my local directory.  For instance in my fig.yml I have

```
  volumes:
    - .:/var/www/html/
```

But file changes don't show in the web server.

In another project, where a volume is mounted the same way, switching branches in my local directory does not seem to update the code in the mounted volume.

Is this a new configuration item I am missing or something?

aanand:
Can you reproduce the bug using just `docker run -v ...` and not Fig? If so, it's a problem with your setup (if you're using boot2docker or some other VM-based solution, it's possible that your folder sharing isn't working properly).

Maxim-Filimonov:
@aanand I'm facing similar issue but only with `fig up`.
See example(looks for BetterErrors config which i have removed locally):
```bash
> tail config/environments/development.rb
  # Adds additional error checking when serving assets at runtime.
  # Checks for improperly declared sprockets dependencies.
  # Raises helpful error messages.
  config.assets.raise_runtime_errors = true

  # Raises error for missing translations
  # config.action_view.raise_on_missing_translations = true

  config.cache_store = :memory_store
end

> fig up app
Attaching to app_1
app_1 |   # Raises error for missing translations
app_1 |   # config.action_view.raise_on_missing_translations = true
app_1 |
app_1 |   config.cache_store = :memory_store
app_1 |
app_1 |   # Allow any ip as we are running it in docker container
=== THIS SHOULD NOT BE HERE app_1 |   BetterErrors::Middleware.allow_ip! =======
app_1 | end

> fig run app
  # Adds additional error checking when serving assets at runtime.
  # Checks for improperly declared sprockets dependencies.
  # Raises helpful error messages.
  config.assets.raise_runtime_errors = true

  # Raises error for missing translations
  # config.action_view.raise_on_missing_translations = true

  config.cache_store = :memory_store
end
> docker run -v /vagrant:/myapp <CONTAINER_ID> tail config/environments/development.rb
  # Adds additional error checking when serving assets at runtime.
  # Checks for improperly declared sprockets dependencies.
  # Raises helpful error messages.
  config.assets.raise_runtime_errors = true

  # Raises error for missing translations
  # config.action_view.raise_on_missing_translations = true

  config.cache_store = :memory_store
end
```

Fig.yml:
```
app:
  volumes:
    - /vagrant:/myapp
[...]

PR #711, opened 2014-12-06
Fix volumes on recreate

dnephin:
Resolves: #447 (and probably some other related issues)

Instead of using `volumes_from` on the intermediate container, inspect the volumes, and copy them over the to intermediate container and from the intermediate to the final container.

Issue #447, opened 2014-08-26
reload fig.yml to mount the changed volumes


japrescott:
When I build a container/service with fig and then later change the directory structure and adjust the volume mounts in fig.yml, my container always wants to mount the old directory. This happens also when I rebuild the container with the `--no-cache` flag.
Only way I have found to get the volumes to be updated is by removing all containers and their images (and the layers) and rebuilding. Interestingly, I've found that when I run the container with `fig run [container] /bin/bash`, the volumes are updated. Using `fig up` does not update the paths to the volumes.
I think this is a bug, but it very well could be that I am missing a point..

PR #858, opened 2015-01-20
Preserve individual volumes on recreate

dnephin:
Resolves: #836

When a container is being recreated (during `docker-compose up`) we use `volumes-from` to copy every volume over to the new container. This has lead to a lot of confusion (there are numerous github issues, and stackoverflow questions about it).

Instead of blanket copying every volume with `volumes-from` this PR changes the behaviour to only copy over container data volumes that are still present in the configuration. Host volumes, and volumes that were removed from the configuration will no longer be copied over (although host volumes are still applied, they just aren't copied) .

#711 was the first attempt, but wasn't compatible with docker 1.4.1+
Related Issue #447

### DIFF
--- a/compose/service.py
+++ b/compose/service.py
@@ -3,14 +3,14 @@ from __future__ import absolute_import
 from collections import namedtuple
 import logging
 import re
-from operator import attrgetter
 import sys
-import six
+from operator import attrgetter
 
+import six
 from docker.errors import APIError
 from docker.utils import create_host_config, LogConfig
 
-from .config import DOCKER_CONFIG_KEYS
+from .config import DOCKER_CONFIG_KEYS, merge_environment
 from .container import Container, get_container_name
 from .progress_stream import stream_output, StreamOutputError
 
@@ -183,10 +183,10 @@ class Service(object):
         Create a container for this service. If the image doesn't exist, attempt to pull
         it.
         """
-        override_options['volumes_from'] = self._get_volumes_from(previous_container)
         container_options = self._get_container_create_options(
             override_options,
             one_off=one_off,
+            previous_container=previous_container,
         )
 
         if (do_build and
@@ -247,6 +247,12 @@ class Service(object):
         self.client.rename(
             container.id,
             '%s_%s' % (container.short_id, container.name))
+
+        override_options = dict(
+            override_options,
+            environment=merge_environment(
+                override_options.get('environment'),
+                {'affinity:container': '=' + container.id}))
         new_container = self.create_container(
             do_build=False,
             previous_container=container,
@@ -326,7 +332,7 @@ class Service(object):
             links.append((external_link, link_name))
         return links
 
-    def _get_volumes_from(self, previous_container=None):
+    def _get_volumes_from(self):
         volumes_from = []
         for volume_source in self.volumes_from:
             if isinstance(volume_source, Service):
@@ -338,9 +344,6 @@ class Service(object):
 
             elif isinstance(volume_source, Container):
                 volumes_from.append(volume_source.id)
-
-        if previous_container:
-            volumes_from.append(previous_container.id)
 
         return volumes_from
 
@@ -363,7 +366,11 @@ class Service(object):
 
         return net
 
-    def _get_container_create_options(self, override_options, one_off=False):
+    def _get_container_create_options(
+            self,
+            override_options,
+            one_off=False,
+            previous_container=None):
         container_options = dict(
             (k, self.options[k])
             for k in DOCKER_CONFIG_KEYS if k in self.options)
@@ -396,10 +403,18 @@ class Service(object):
                 ports.append(port)
             container_options['ports'] = ports
 
+        override_options['binds'] = merge_volume_bindings(
+            container_options.get('volumes') or [],
+            previous_container)
+
         if 'volumes' in container_options:
             container_options['volumes'] = dict(
                 (parse_volume_spec(v).internal, {})
                 for v in container_options['volumes'])
+
+        container_options['environment'] = merge_environment(
+            self.options.get('environment'),
+            override_options.get('environment'))
 
         if self.can_be_built():
             container_options['image'] = self.full_name
@@ -417,11 +432,6 @@ class Service(object):
     def _get_container_host_config(self, override_options, one_off=False):
         options = dict(self.options, **override_options)
         port_bindings = build_port_bindings(options.get('ports') or [])
-
-        volume_bindings = dict(
-            build_volume_binding(parse_volume_spec(volume))
-            for volume in options.get('volumes') or []
-            if ':' in volume)
 
         privileged = options.get('privileged', False)
         cap_add = options.get('cap_add', None)
@@ -447,8 +457,8 @@ class Service(object):
         return create_host_config(
             links=self._get_links(link_to_self=one_off),
             port_bindings=port_bindings,
-            binds=volume_bindings,
-            volumes_from=options.get('volumes_from'),
+            binds=options.get('binds'),
+            volumes_from=self._get_volumes_from(),
             privileged=privileged,
             network_mode=self._get_net(),
             devices=devices,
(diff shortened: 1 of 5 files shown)


aanand:
Looks like tests are failing on 1.2.0 - this line fixed it:

```patch
diff --git a/tests/integration/service_test.py b/tests/integration/service_test.py
index 5e00ec1..6df2301 100644
--- a/tests/integration/service_test.py
+++ b/tests/integration/service_test.py
@@ -148,6 +148,7 @@ class ServiceTest(DockerClientTestCase):
         self.assertIn('FOO=1', old_container.get('Config.Env'))
         self.assertEqual(old_container.name, 'figtest_db_1')
         service.start_container(old_container)
+        old_container.inspect() # reload volumes
         volume_path = old_container.get('Volumes')['/etc']

         num_containers_before = len(self.client.containers(all=True))
```

However, there's a more serious regression - see https://github.com/docker/fig/pull/859. The new logic introduces in #711 doesn't preserve volumes that are declared in the image.

What if the volume copying logic were:

- copy all volume binds from an existing container
- add/overwrite all volumes with explicit host paths from `fig.yml`
- add any missing volumes *without* explicit host paths from `fig.yml`

This wouldn't handle the case where a volume is deleted from `fig.yml`, but at least it'd handle the case where a volume is *remapped* to a different host path.

compose/service.py
```diff
@@ -519,6 +522,59 @@ def pull(self, insecure_registry=False):
...
+    """Find the container data volumes that are in `volumes_option`, and return
+    a mapping of volume bindings for those volumes.
+    """
+    volumes = []
+
+    volumes_option = volumes_option or []
+    container_volumes = container.get('Volumes') or {}
+    original_volumes = set(volumes_option + container_volumes.keys())
+
+    for volume in original_volumes:
```
dnephin: This works in docker 1.2 to fix the test added in #859

Using docker 1.4.1 it seems like docker is refusing to override the volume with a `bind`.  This function is returning the correct value, but when `container.start(binds=...)` is called, the old volume remains.

Maybe `volumes_from` has to be used to override the volume? Maybe this is a bug in docker?  I'm not really clear on what the expected behaviour is.

thaJeztah: Maybe volumes-upper-guru @cpuguy83 has some time to have a look? Think you want this solved in the final release?

cpuguy83: Have you tried against Docker master?
You should indeed be able to override an existing volume, and volumeFrom have priority over all.

andreyst: @cpuguy83 Per @dnephin request adding info from my case here: https://gist.github.com/andreyst/e213c18ef32dcb28dc5e

This works if I try to mount something else like /tmp. If I ln -s the dir I need in /tmp and try to pass /tmp/dirname as a volume, it still doesn't work. If then I pass /tmp as a volume, it does work, but if then I try to cd to it from a container, it gives me "No such file or directory".

dnephin: I tried to create a minimal reproduction of this failure, but when I tried with the docker cli I was not able to reproduce the failure.

https://gist.github.com/dnephin/e584a7741a514d631632

From the docker inspect they look almost identical, except you can see in the fig example the volume is not set to the path specified by `binds`.  

I tried to use mitmproxy to inspect the differences between the http requests being used, but it seems like the docker cli ignores `HTTP_PROXY` (I was able to get the requests made by fig at least).

I'm struggling to find the difference between this example and the test case. Does anyone know of a way to make the docker cli use a proxy?

thaJeztah: @dnephin Support for proxies was added in Docker 1.5 via https://github.com/docker/docker/pull/9951 (docs changes for it are in https://github.com/docker/docker/pull/10192) not sure if there's another way to specify this. Given that it's just a one line change, you could try to build a 1.4.1 version with just that one change added?

(Sorry if it's a lame answer, but perhaps the links are useful :))

dnephin:
Ok, This is working for 1.4.1 now, but requires api version 1.5 , which I think is  1.4 ?

So blocking on #886 and makes compose incompatible with earlier versions.

To support older versions we'd have to detect the server version and pass the binds to `start` instead of `create`. Not sure if it's worth it.

thaJeztah:
If the docs are correct, API v1.15 corresponds with docker 1.3 (https://docs.docker.com/v1.3/reference/api/docker_remote_api/). In my opinion, that's reasonable. Several CVEs were addressed in docker 1.3.x, so people should be encouraged to run that version at minimum.

Otherwise, people can always run Fig 1.0.1 (that won't dissapear) :smile:

cpuguy83:
But, keep in mind, there should be no issue with doing this on start (at least from the Docker side)!

dnephin:
@cpuguy83 from https://gist.github.com/dnephin/e584a7741a514d631632 it looks like (at least in 1.4.1) the binds in start are not overriding those from the Dockerfile. Maybe that's fixed in 1.5, I haven't had a chance to try it yet.

cpuguy83:
@dnephin I think I've found the problem in Docker.
It's this check here: https://github.com/docker/docker/blob/v1.4.1/daemon/volumes.go#L83

For volumes that are created by docker, `IsBindMount` would always be false.
This may not work on master, let me do some tests and see if I can get a fix in for 1.5.

cpuguy83:
And that one may be different actually.
Testing fig against master seems to be OK.

```yaml
# fig.yml
web:
  build: .
  volumes:
    - "/foo"
    - "/bar"
```

```Dockerfile
# Dockerfile
FROM busybox
ADD . /foo
VOLUME /foo
VOLUME /bar
CMD exec top
```

Does this look right?

dnephin:
I don't think you'll be able to reproduce it with master, since it's still using `volumes_from`, and it only manifests when you recreate a container.

This gist does reproduce it: https://gist.github.com/dnephin/083a1052b8293a5b46a4

It prints the volumes from both containers. They should be the same, but the second container has a different volume for `/data`.

cpuguy83:
Yep, this will be fixed by docker/docker#10365

dnephin:
I've updated the title and description of this issue to better reflect the changes.

I'd love to chat about getting this into 1.2.  I think this is still the most common issue people have with `docker-compose` (see the long list of related tickets above, there are also multiple stack-overflow questions about it).

This PR will need to be updated once we remove the intermediary container (in #874), but there is nothing in docker 1.5 (or coming in docker 1.6) that actually fixes the issue addressed by this PR.

How do you feel about putting this into the 1.2 milestone?

thaJeztah:
> volumes that were removed from the configuration will no longer be copied over

This *does* have the nasty side-effect that those volumes will be left behind (orphaned), which makes it very difficult to access them or clean then up until docker implements volume management

dnephin:
That is true, but this is basically the case with any `docker run -v` command that is run, right?

thaJeztah:
Yes, but with those I have *some* control, ie `docker rm -v <container>`. 

Given how many questions I've seen from people unaware of this ("docker leaks containers", "docker eats my disk space"), it should be documented properly.

Lol, maybe compose should create a dummy container and attach all old volumes to that :)

In general, Docker *really* needs better volume management. There are various proposals, so I hope it someday arrives natively.

Apart from those concerns, I agree with this PR

dnephin:
Ok, this is rebased with master and tests are passing.

aanand:
Nice.

One more thing: Swarm uses `volumes_from` to co-schedule containers. It won't have that to go on any more, so we'll need to explicitly co-schedule with an environment variable:

    affinity:container==previous_container_id

dnephin:
Cool, is swarm still using environment variables for this? The README suggests it's using labels now as well. Maybe both are supported? I wonder if it makes sense to just use labels?

dnephin:
Added the affinity.  I was also able to cleanup some of the `merge_environment()` logic for `run` that was in `cli.main`.

compose/service.py
```diff
@@ -246,6 +246,12 @@ def recreate_container(self, container, **override_options):
         self.client.rename(
             container.id,
             '%s_%s' % (container.short_id, container.name))
+
+        override_options = dict(
+            override_options,
+            environment=merge_environment(
+                override_options.get('environment'),
+                {'affinity:container': container.id}))
```
aanand: Hmm, it needs to come out as `affinity:container==<id>`, i.e. a double-equals. Maybe we should add a test for it?

(I'm going to investigate running Compose's test suite against Swarm, btw)

dnephin: Oh, I did add a test for it, but I didn't realize it required the `==`.

If I set the value to `=<id>` that should work, right ?

dnephin:
Using `==` now

aanand:
LGTM

PR #859, opened 2015-01-20
Test for preservation of volumes declared in images

aanand:
This test succeeds against 72095f54b26650affed47c3b888d77572d6ecbf0 (i.e. before the merge of #711). We weren't testing this case before.


dnephin:
LGTM once rebased with #863 (and the test passes)

aanand:
Rebased and tests pass.

Issue #843, opened 2015-01-14
fig up tries to run incorrect command


jbulchergist:
I have a *data container* build from *scratch* with nothing but a file called `/true` init, so that fig is able to start up the container (it would be nice if I didn't even need that file). Fig starts the container fine the first time, but attempting to run `fig up` on the service after bringing it down results in an odd message:

`Cannot start container 12341234...: exec: "/bin/echo": stat /bin/echo: no such file or directory`

I don't want fig runing echo - I want to run `/true`, if anything.

```Dockerfile
FROM scratch

COPY true /true

CMD [ "/true" ]
```

Shouldn't fig respect `CMD`? Can I make it respect `CMD`?

dnephin:
I think this is a duplicate of #613 and #517

dnephin:
Tracking this in #613, the workaround for now is to use `busybox` or a container that has `/bin/echo`

Issue #613, opened 2014-11-04
Fig data-only containers


ElessarWebb:
Using this simple configuration:

```yaml
data:
  image: base/arch
  volumes:
    - /var/lib/mysql
  command: true

mysql:
  build: .
  volumes_from:
    - data
```

I would expect that the mysql data is persisted on the data-only container 'data'. But writing anything to /var/lib/mysql is *NOT* persisted across runs. Doing this with plain docker does work.

Am I overlooking something, or is this either a bug or a feature of fig?

The fig commands I'm trying to use are:

```
fig run mysql init
fig run mysql start
```

The mysql image has an entrypoint that takes init and start command. The first one populates /var/lib/mysql, the second one starts the mysql server with /var/lib/mysql as it's datadir. Both run, but the second one fails because /var/lib/mysql is empty.

The following is a minimal example of it failing:

```
fig run mysql /bin/bash -c "touch /var/lib/mysql/testfile && ls -l /var/lib/mysql"
fig run mysql /bin/bash -c "ls -l /var/lib/mysql"
```

Where the first `ls -l` shows that the file was created, where as the second one returns empty.

Using docker 1.3.0 and fig 1.0.0

drags:
Ran into a similar issue while debugging with a coworker and wanted to leave this comment in case it helped anyone else.

One of our users had an issue with data not persisting between fig runs. We found that the docker daemon running inside of boot2docker seemed to be ignoring volume configs completely (whether we used volumes_from with a data container, or even a static volume mount). The container that wound up getting spawned would always have a docker-daemon created volume mount (similar to when a Dockerfile contains a VOLUME line and no other volume information is provided at run time, and that image did contain a VOLUME line for that dir). Also, it would create a new version of the data container each run, rather than re-using the initially created one.

While we didn't find the root cause of the issue we noticed that the docker client (on the OSX side) was 1.3.2, whereas the docker server was running 1.4.0. After upgrading boot2docker and getting the OSX docker client up to 1.4.0 the volumes are mounting as expected and data is persisting.

Hope this helps someone with a similar issue.

Starefossen:
I'm having a hard time getting persistent volumes from my data only containers. I had to use @dnephin's workaround `fig up -d data` in order for fig/docker to keep the volume persistent between `fig up`.

Docker version 1.3.3 (build d344625), Fig 1.0.1 (master@55095e)

dnephin:
#711 has been merged which I believe resolves this issue. It will be in the next release

[...]

PR #517, opened 2014-10-03
Add an explicit noop option for minimal containers that do not have echo

vizowl:
In order to support data only volumes using minimal images #514 I have added an option that allows a no-op command to be explicitly specified - currently it is assumed to be echo which isn't necessarily available on a minimal image.

I have added a test and updated the documentation.

I am concerned that this approach essentially exposes an internal feature ...

PR #860, opened 2015-01-20
service.py: configurable entrypoint

### LINKED ISSUE #514: Second `up` rewrites entry point
The second time you `fig up`, fig rewrites the entry point on a container. To reproduce this issue, start with the following fig file:

```yaml
whomovedmyentrypoint:
    image: scratch
    entrypoint: fail
```

Then, run `fig up` twice, inspecting the entry point each time:

```
$ fig up; docker inspect --format='{{.Config.Entrypoint}}' $(docker ps -ql); fig up; docker inspect --format='{{.Config.Entrypoint}}' $(docker ps -ql)
Creating foo_whomovedmyentrypoint_1...
Cannot start container 5206410efc5dce757cc0146d5dd3b28ac3cf37e23100015ea8565adf832edeb1: exec: "/fail": stat /fail: no such file or directory
[/fail]
Recreating foo_whomovedmyentrypoint_1...
Cannot start container 3fc0d214031cf7a0f955501f43e546381139610bd30ee8df98b9d938579707b9: exec: "echo": executable file not found in $PATH
[echo]
```

The use case for _not_ changing the entry point is that the minimal container just for housing volumes (documented by Docker as [Data Volume Containers](https://docs.docker.com/userguide/dockervolumes/)) is little more than scratch with directories and a single, no-op executable like [true-asm](https://github.com/tianon/dockerfiles/tree/master/true).

piranha:
In case your image is small enough to not contain /bin/echo, fig will 
not work properly failing with "stat /bin/echo: no such file or directory".

Fixes #514, #613, #843

### DIFF
--- a/fig/service.py
+++ b/fig/service.py
@@ -284,7 +284,7 @@ class Service(object):
         intermediate_container = Container.create(
             self.client,
             image=container.image,
-            entrypoint=['/bin/echo'],
+            entrypoint=intermediate_options.get('entrypoint', '/bin/echo'),
             command=[],
             detach=True,
         )


dnephin:
There was a bunch of discussion about this issue in #517 as well. I had [proposed this solution](https://github.com/docker/fig/pull/517#issuecomment-58070969), but someone correctly pointed out that although this works well for data-only containers, it does not work for regular runnable containers, because it would start that container during the recreate process.

aanand:
With Docker 1.5 we can use renaming and drop intermediate containers entirely: https://github.com/docker/fig/issues/874

dnephin:
We've added renaming, so we don't need this intermediate container anymore

PR #863, opened 2015-01-20
Revert #711 from dnephin/fix_volumes_on_recreate

dnephin:
Turns out this needs more testing.

There are some significant differences between docker 1.2  and 1.4.1 that aren't handled properly.

It would be awesome if we could setup wercker to test against at least a couple different versions of dockerd.

### DIFF
--- a/fig/project.py
+++ b/fig/project.py
@@ -176,14 +176,18 @@ class Project(object):
            do_build=True):
         running_containers = []
         for service in self.get_services(service_names, include_links=start_links):
-            create_func = (service.recreate_containers if recreate
-                           else service.start_or_create_containers)
-
-            for container in create_func(
-                    insecure_registry=insecure_registry,
-                    detach=detach,
-                    do_build=do_build):
-                running_containers.append(container)
+            if recreate:
+                for (_, container) in service.recreate_containers(
+                        insecure_registry=insecure_registry,
+                        detach=detach,
+                        do_build=do_build):
+                    running_containers.append(container)
+            else:
+                for container in service.start_or_create_containers(
+                        insecure_registry=insecure_registry,
+                        detach=detach,
+                        do_build=do_build):
+                    running_containers.append(container)
 
         return running_containers
 
--- a/fig/service.py
+++ b/fig/service.py
@@ -242,9 +242,8 @@ class Service(object):
 
     def recreate_containers(self, insecure_registry=False, do_build=True, **override_options):
         """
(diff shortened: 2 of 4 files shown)


aanand:
LGTM

bfirsh:
@dnephin Wercker is very solidly stuck at a single Docker version, unfortunately.

What we can do is get Fig tested on [Docker's Jenkins instance](https://jenkins.dockerproject.com/). That'll let us test on Windows and other cool things like that.

cpuguy83:
@dnephin Please let me know if you need any help on this.
I'm not a pythonist at all, but know the volumes side well and can fix any issues you're seeing on the Docker side of this.

Also would be a great help to test and make sure the Docker 1.5 RC is working!

dnephin:
Thanks @cpuguy83  I'm working on this in #858

I tried to reproduce with docker cli, but was not able to reproduce the error https://gist.github.com/dnephin/e584a7741a514d631632

Next step is to send the requests through mitmproxy to find out why they are different (I believe they should be the same).  I was informed the docker cli doesn't support proxy settings until 1.5 anyway, so I'll have to try with 1.5

cpuguy83:
@dnephin This is probably because the CLI does everything on the create endpoint.

dnephin:
Oh, from the docker-py docs I wasn't aware that was possible, but that makes a lot of sense.

Does it get passed as `binds

cpuguy83:
I think @shin- Added this to docker-py recently.
Basically, as of docker 1.2(?) you can pass in anything form HostConfig (read: everything that you were passing in at start) in during create time. https://docs.docker.com/reference/api/docker_remote_api_v1.16/#create-a-container

So it still looks the same, except the start options are nested within the main contianer config as HostConfig.  This should look like the `docker inspect` output.

dnephin:
Ok `Binds` is missing from docker-py master https://github.com/docker/docker-py/blob/master/docker/client.py#L206

I should have looked at the official API docs instead of just the docker-py source.  I'll try that.

shin-:
[We do support Binds, but you're looking in the wrong place :)](https://github.com/docker/docker-py/blob/master/docker/utils/utils.py#L340-L341)

Let me know if I can help!

dnephin:
Ah yes, realized it has to go through `host_config`. 

I believe it's working with `host_config` ! I need to do some cleanup, and we need to resolve #886 so we can unpin `version='1.14'`, but otherwise this is great progress!

Thanks @cpuguy83  @shin-

PR #868, opened 2015-01-21
Missing router.php file in php command

tdutrion:
The `router.php`was missing in the `php -S`command, and the router was therefore never called.

### DIFF
--- a/wordpress.html
+++ b/wordpress.html
@@ -32,7 +32,7 @@ ADD . /code
 <p>Next up, <code>fig.yml</code> starts our web service and a separate MySQL instance:</p>
 <div class="highlight"><pre><code class="text language-text" data-lang="text">web:
   build: .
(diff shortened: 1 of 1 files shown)


aanand:
I can't reproduce the problem you mention - the Wordpress tutorial works just fine for me.

Has anyone else had trouble with it?

prestalyon:
It does actually work because index.php is called, but the router.php file
is never called in the current example.

Solutions are either set /code/router.php at the end of the command or
completely remove the router.php file.

Whereas it would be tempting to just remove it, and the basic WordPress
will work just fine, problems occurs when using plugins such as CSS or js
concatenation plugins where php scripts are directly called. In such cases,
index.php would catch the request instead of the targeted file, and
therefore result in an error.

dnephin:
You're right, we don't need this `router.php` at all.  Let's remove it from the docs instead of including it.

PR #879, opened 2015-01-25
Update bash completion for compose 1.1.0-rc1

albers:
This adds `docker-compose run --service-ports`.
Support for `docker-compose kill -s SIGNAL` and `docker-compose up --no-build` was already added in my last PR.
With this, bash completion is in sync with the additions mentioned in the release notes.

Issue #864, opened 2015-01-20
Support fig.yml and show a deprecation warning


aanand:

thaJeztah:
+1 would be welcome to give me time to update all my repos :)

ibnesayeed:
If there are both `fig.yml` as well as `compose.yml` in the same directory, later should take the precedence. Should the deprecation warning still be shown or can it just be ignored?

aanand:
The `.yaml` support complicates this too. I think my preferred logic would be:

- Try the following filenames until one exists: `docker-compose.yml`, `docker-compose.yaml`, `fig.yml`, `fig.yaml`
- If there's more than one file with any of the above names, explicitly say which one we're using.
- If we're using `docker-compose.yaml`, show the standard nagging message about file extensions.
- If we're using `fig.*`, show a deprecation warning.

ibnesayeed:
I thought the file name will be `compose.yml` instead of `docker-compose.yml`, or may be both. Short names are good unless they create any confusion.

aanand:
Name length is definitely an issue, but I think it's *more* of an issue for the binary - whose name is being typed over and over - than the configuration file.

I don't love `docker-compose` as a binary name, but I don't mind it all as a filename. Indeed, it's a good clue to someone who's just cloned a repository that it uses Docker.

ibnesayeed:
I am not sure why the binary name was chosen to be `docker-compose` instead of just `compose`, was there any name conflict with any existing package?

funkyfuture:
this is getting off-topic:
"[Ecplicit is better than implicit](https://www.python.org/dev/peps/pep-0020/)" is a good mantra.

i suggest you set an `alias` in your `.bashrc` or set a link in your `$PATH`.
though a short form that is canonically used in Docker's community would support communications.
`dco`? `doco`, `docomp`? but anyone who wants to discuss that should create a new issue for that. i don't want to discuss it.

and yes, `pip search compose` actually lists a package.

albers:
I updated the bash completion according to aanands suggestion, see #887.

PR #887, opened 2015-01-26
Bash completion supports fig.yml and other legacy filenames

albers:
~~This builds on #879 Update bash completion for compose 1.1.0-rc1.~~
It adds support for legacy stack composition filenames as suggested by aanand in #864.


albers:
rebased on master so that it can be merged independently from #879

aanand:
LGTM

dnephin:
LGTM

Issue #750, opened 2014-12-22
Fig tutorial for Ruby on Rails doesn't work


lblackburn:
Using ruby:2.1.5 and rails ~> 4.2.0

When I fig up I can't access the rails app on 192.168.59.103:3000 (using boot2docker)

zhulinpinyu:
+1

madwire:
I had the same problem just bind to 0.0.0.0 e.g
```
command: bundle exec rails s -p 3000 -b '0.0.0.0'
```
See if that works.

aanand:
Is this the cause? From the [Rails 4.2 release notes](http://edgeguides.rubyonrails.org/4_2_release_notes.html#default-host-for-rails-server):

> Due to a [change in Rack](https://github.com/rack/rack/commit/28b014484a8ac0bbb388e7eaeeef159598ec64fc), `rails server` now listens on `localhost` instead of `0.0.0.0` by default.

If so, we just need to update the `fig.yml` in the tutorial as @madwire suggests.

zhulinpinyu:
I had the same problem.
my solution is
        
    app server use  `unicorn`, not WEBrick

is worked for me.

pekeler:
Ran into same problem with my Rails 4.1.7 app. The bind parameter makes it work again. But now `fig up` is not showing rails' stdout anymore.
```
bash-3.2$ fig up
Recreating myapp_db_1...
Recreating myapp_web_1...
Attaching to myapp_db_1, myapp_web_1
db_1  | LOG:  database system was shut down at 2015-02-15 18:30:35 UTC
db_1  | LOG:  database system is ready to accept connections
db_1  | LOG:  autovacuum launcher started
```
Ugh, this all used to work last month.

pekeler:
... and suddenly it shows up again
```
bash-3.2$ fig up
Recreating myapp_db_1...
Recreating myapp_web_1...
Attaching to myapp_db_1, myapp_web_1
db_1  | LOG:  database system was shut down at 2015-02-15 19:08:17 UTC
db_1  | LOG:  database system is ready to accept connections
db_1  | LOG:  autovacuum launcher started
web_1 | => Booting Thin
web_1 | => Rails 4.1.7 application starting in development on http://0.0.0.0:3000
...
```

looks like something didn't flush.

dnephin:
Looks like this was fixed, it now binds to `0.0.0.0`.

PR #810, opened 2015-01-07
Tweaks to the rails tutorial to bring it inline with rails 4.2 release

madwire:
Due to a change in Rack, rails server now listens on localhost instead of 0.0.0.0 by default.

Please also see #750

PR #888, opened 2015-01-27
Fix problem with not binding to public interfaces

FlorianLoch:
The problem of not binding to all, including the public interfaces, is also discussed in issue #750. That causes the fig.yml in the tutorial to not work.


FlorianLoch:
Sorry for "double fixing" - pull request #810 already fixes the problem.

PR #891, opened 2015-01-27
Fix test for image-declared volumes

aanand:
A stray 'fig' got lost in the merge post-rename, meaning the containers weren't being cleaned up properly. Test was failing against Docker 1.4.1.


dnephin:
LGTM

PR #894, opened 2015-01-28
Fix database.yml file with port and database name

SaitoWu:
the `postgres` database is the default database, you can't use it.

port should be filled cause the default strategy is connect by socks file

### DIFF
--- a/docs/rails.md
+++ b/docs/rails.md
@@ -69,11 +69,11 @@ Open up your newly-generated `database.yml`. Replace its contents with the follo
     development: &default
       adapter: postgresql
       encoding: unicode
-      database: postgres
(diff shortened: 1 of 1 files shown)


aanand:
Regarding the first point: our usage of the `postgres` database is in line with that recommended by the [official postgres image](https://registry.hub.docker.com/_/postgres/) and the [Postgres documentation](http://www.postgresql.org/docs/9.3/interactive/app-initdb.html):

> The postgres database is a default database meant for use by users, utilities and third party applications.

Regarding the second: are you getting a connection error? Could you paste the terminal output?

SaitoWu:
Can't reproduce the result, close for now.

PR #900, opened 2015-01-29
Ship 1.1.0-rc2

aanand:
The release notes for 1.1.0-rc2 will be an an amalgamation of the CHANGELOG.md entries for both RCs.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,17 @@
 Change log
 ==========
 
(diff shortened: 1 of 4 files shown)


aanand:
Rebased.

dnephin:
I think we update tab completion as well, right?

LGTM

aanand:
Yeah. Updated.

PR #916, opened 2015-02-02
Fix version string - using pep0440

robsonpeixoto:
Using the PEP 0440 conversion to version name: https://www.python.org/dev/peps/pep-0440/#examples-of-compliant-version-schemes

Signed-off-by: Robson Peixoto <robsonpeixoto@gmail.com>


aanand:
Thanks. The horse has bolted on this one, but for future releases we should consider using the PEP0440 convention.

PR #922, opened 2015-02-05
Tweak intro

aanand:
We shouldn't yet be recommending production use - for now, let's continue to emphasise the development use case and mention staging/CI.

cc @bfirsh @fredlf @SvenDowideit


fredlf:
LGTM

PR #925, opened 2015-02-06
Link correction

stevepereira:


aanand:
`.html` is correct - the docs will eventually be hosted on docs.docker.com as rendered HTML.

stevepereira:
....
Why not merge and then update it when 'eventually' becomes 'actually'? Leaving it broken until 'eventually' seems like something of a usability issue.

aanand:
Because the docs aren't meant to be consumed at github.com/docker/fig. Their current location is www.fig.sh, where they work just fine.

stevepereira:
Aha!

:) well done then!

Issue #219, opened 2014-05-11
Python 3?


analytically:
Ubuntu 14.04 defaults to Python 3.4 be default, so let's make fig compatible.

- Use https://github.com/bufordtaylor/python-texttable instead of texttable
- Upgrade websocket-client to v0.14 (will be released on May 15)
- Upgrade PyYAML to 3.11
- In fig's source, change

```
except StreamOutputError, e:
```
to
```
except StreamOutputError as e:
```

- Add '3.4' to .travis.yml

I'll submit a pull request when websocket-client 0.14 is released.

dnephin:
#440 is in master!  the next release should support python 3

franklinyu:
The [Fedora package](https://apps.fedoraproject.org/packages/docker-compose) still depends on Python 2, which should be unnecessary since the [Arch Linux package](https://www.archlinux.org/packages/?name=docker-compose) depends only on Python 3 packages. Is the maintainer of the Fedora package also in this repo? Or I need to submit bug report there?

PR #926, opened 2015-02-06
fix syntax errors indicated by pip

allustin:
wthen we tried install docker-compose with pip - we get an syntax error


aanand:
Are you using Python 3? Support is in the works - see #219

aanand:
(This is still a change worth making, but this PR needs squashing to a single commit and signing off)

allustin:
@aanand yes - we use Python 3.4 .Thanks.

PR #929, opened 2015-02-06
have setup.py read requirements.txt

larsks:
I was surprised to find the list of requirements was replicated inside 
setup.py as well in requirements.txt/requirements-dev.txt.  This change has
setup.py read the files so that the requirements only need to be maintained in
one place.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,5 +1,5 @@
 mock >= 1.0.1
 nose
-git+https://github.com/pyinstaller/pyinstaller.git@12e40471c77f588ea5be352f7219c873ddaae056#egg=pyinstaller
+pyinstaller
 unittest2
 flake8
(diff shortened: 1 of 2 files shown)


dnephin:
This is absolutely the wrong way to do setup.py/requirements.txt.

The list is not duplicated. `setup.py:install_requires` lists the minimal required versions for a package. `requirements.txt` pins exact versions so that the build is repeatable. 

See https://caremad.io/2013/07/setup-vs-requirement/ for more details

larsks:
Well, I disagree, but I'm not the one maintaining the code so if you're happy with it, more power to you!

bfirsh:
Thanks for the contribution @larsks, but @dnephin has put a lot of thought into getting this right! Out of interest – why do you disagree?

bfirsh:
@dnephin I wonder if it'd be worth adding some comments documenting which requirements go where?

dnephin:
Linking to https://caremad.io/2013/07/setup-vs-requirement/  definitely wouldn't hurt

PR #951, opened 2015-02-11
README.md: Update "Docker Compose" => "fig"

msabramo:


dnephin:
This is going in the wrong direction. Fig is being renaming to docker compose.

bfirsh:
He has a point... people are landing on this page for documentation expecting it to be Fig. /cc @aanand

msabramo:
This was the first of me hearing of the fig to Docker Compose name change so that's why I was confused.  But I had looked at fig a long time ago and just recently remembered it and starting looking at using it. So I'm:

1. Coming with prior memory of the "fig" name. 
2.  Not following mailing lists and such 

Like any name change, it may cause a bit of confusion and then it will work itself out.

aanand:
Perhaps we should temporarily add a "(previously known as Fig)" note to the README.

PR #962, opened 2015-02-12
Add syntax highlighting to yml.md

Starefossen:
Added `yml` syntax highlighting to yml-example code.


aanand:
Thanks. @fredlf? Do we have a policy on syntax highlighting hints of any kind?

dnephin:
It looks like this is being handled in the new docs (http://docs.docker.com/compose/yml/)

Issue #886, opened 2015-01-26
Run CI tests on multiple Docker versions


bfirsh:
Our CI tests should run on several recent versions of Docker. The best way to do that would probably be to use [Docker's Jenkins instance](https://jenkins.dockerproject.com/), managed by @jfrazelle I think.

jfrazelle:
I have ideas!!! I was telling @aanand lol but I might just try something out with a Dockerfile and multiple `ADD` lines pulling the binaries from get.docker.com, then running some scripts or something

thaJeztah:
Wouldn't the storage of those versions interfere with each other? Or is a full cleanup needed between starting each version?

Liking the creativity :)

jfrazelle:
well if you run the daemon in a container much like the integration-cli tests in docker you cna have setup and tear down and all that mumbo jumbo :D

thaJeztah:
mumbo jumbo. My favorite!

bfirsh:
@jfrazelle If we test on the same infrastructure as the engine do we get testing on Windows and OS X for free? :D

jfrazelle:
not osx and windows is still experimentally testing

dnephin:
PRs #858 and #958 are both blocked on being able to run tests against a newer docker than the one available from wercker.

aanand:
Update: see #1005, which uses Docker-in-Docker to test against multiple versions.

bfirsh:
Put this back on the backlog because with #1066 we may only need to test against Docker 1.6. :)

aanand:
Closed by #1005.

PR #972, opened 2015-02-16
Make Swarm schedule containers on the same node when they are linked together

aanand:
... by specifying all HostConfig at create time. This is required for [Swarm integration](https://github.com/docker/fig/issues/814): the cluster needs to know about config like `links` and `volumes_from` at create time so that it can co-schedule containers. This is particularly important because we rely on `volumes_from` when recreating containers - if the old/intermediate/new containers aren't co-scheduled, they'll fail to start.

Because the ability to specify HostConfig at create time was introduced in Docker 1.4, that's our new minimum version requirement. The install docs have been updated accordingly.

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -328,20 +328,20 @@ class TopLevelCommand(Command):
         if options['--user']:
             container_options['user'] = options.get('--user')
 
+        if not options['--service-ports']:
+            container_options['ports'] = []
+
         container = service.create_container(
             one_off=True,
             insecure_registry=insecure_registry,
             **container_options
         )
 
-        service_ports = None
-        if options['--service-ports']:
-            service_ports = service.options['ports']
         if options['-d']:
-            service.start_container(container, ports=service_ports, one_off=True)
+            service.start_container(container)
             print(container.name)
         else:
-            service.start_container(container, ports=service_ports, one_off=True)
+            service.start_container(container)
             dockerpty.start(project.client, container.id, interactive=not options['-T'])
             exit_code = container.wait()
             if options['--rm']:
--- a/compose/cli/docker_client.py
+++ b/compose/cli/docker_client.py
@@ -32,4 +32,4 @@ def docker_client():
         )
 
     timeout = int(os.environ.get('DOCKER_CLIENT_TIMEOUT', 60))
-    return Client(base_url=base_url, tls=tls_config, version='1.14', timeout=timeout)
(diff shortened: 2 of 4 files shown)


cpuguy83:
Looks like HostConfig on create has been available since 1.3.0 - https://github.com/docker/docker/blob/v1.3.0/daemon/create.go#L31

dnephin:
Overall I like it.  

This is going to conflict pretty heavily with #858, but it should be manageable. 

Both this PR and #858 are kind of blocked on #886.  Do you know if there is any word on when we'll be able to run the test suite against docker > 1.3 ?

aanand:
@cpuguy83 Interesting - looks like this wasn't made official until API version 1.15: http://docs.docker.com/reference/api/docker_remote_api/#v115

@dnephin Yeah, we really need a better setup. I'm looking into it.

aanand:
Rebased.

compose/cli/main.py
```diff
@@ -328,20 +328,20 @@ def run(self, project, options):
...
-        service_ports = None
-        if options['--service-ports']:
-            service_ports = service.options['ports']
         if options['-d']:
-            service.start_container(container, ports=service_ports, one_off=True)
+            service.start_container(container)
             print(container.name)
         else:
-            service.start_container(container, ports=service_ports, one_off=True)
+            service.start_container(container)
```
dnephin: minor (since it's existing), this is being done in both branches of the is/else, so it should really happen before the branch.

dnephin:
LGTM (assuming there is still some place to run the automated test suite with the new dependency on docker 1.4.1)

bfirsh:
@dnephin Now testing on Jenkins against multiple Docker versions! Should this not work on Docker <1.4 though? Hm. Seems like it is.

dnephin:
@bfirsh I believe the only issue was the api version change (`API_VERSION = '1.15'`) would require docker 1.3.x, where as wercker was on 1.2.x right?  I guess my comment should have said `1.3.x`, I got the version wrong.

zk000133:
Built Docker Compose from the 'set-host-config-at-create-time' branch.
If I have any exposed ports declared in the docker-compose.yml file, Swarm will complain when doing a compose up: "unable to find a node with port xxxx available"
Scheduling single containers with bind ports on a swarm works.

aanand:
I've reverted the documentation changes. The API version has changed, but 1.3 is still the minimum version. (This is largely irrelevant anyway, as we'll be making 1.6 the minimum in the next release.)

PR #973, opened 2015-02-16
Better support for DOCKER_TLS_VERIFY environment variable.

Vudumen:
Signed-off-by: Balint Laszlo BILLER <vudumen@vudumen.hu>

### DIFF
--- a/compose/cli/docker_client.py
+++ b/compose/cli/docker_client.py
@@ -21,7 +21,7 @@ def docker_client():
     try:
         tls_verify = int(tls_verify) > 0
     except ValueError:
-        tls_verify = tls_verify in ["true", "yes", "on", "enabled", "enable"]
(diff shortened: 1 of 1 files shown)


aanand:
Thanks for the contribution, but we want to stay exactly in line with [how the Docker client reads the variable](https://github.com/docker/docker/blob/04b0bf4c27bed6c07f84b5ad5c252917f912f0cd/docker/flags.go#L16).

```go
var (
	dockerCertPath  = os.Getenv("DOCKER_CERT_PATH")
	dockerTlsVerify = os.Getenv("DOCKER_TLS_VERIFY") != ""
)
```

If you can convince the maintainers to accept your change, then I'll happily make the same change in Compose.

dnephin:
Going to close this. If docker starts to support the proposed version, feel free to re-open

PR #977, opened 2015-02-17
[fix] strict client verify as ENV variable

mosquito:
Sometimes I don't want check the server certificate.

### DIFF
--- a/compose/cli/docker_client.py
+++ b/compose/cli/docker_client.py
@@ -17,6 +17,7 @@ def docker_client():
     tls_config = None
 
     if os.environ.get('DOCKER_TLS_VERIFY', '') != '':
+        tls_strict = bool(os.environ.get('DOCKER_TLS_STRICT', '1'))
         parts = base_url.split('://', 1)
         base_url = '%s://%s' % ('https', parts[1])
 
@@ -25,7 +26,7 @@ def docker_client():
 
         tls_config = tls.TLSConfig(
             ssl_version=ssl.PROTOCOL_TLSv1,
-            verify=True,
+            verify=tls_strict,
             assert_hostname=False,
             client_cert=client_cert,
             ca_cert=ca_cert,


cpuguy83:
I believe the Docker CLI is looking for `DOCKER_TLS_VERIFY`, might be good to use that instead.

mosquito:
Unfortunately there is no. That is unexpected. Docker CLI is only use this object.

cpuguy83:
Yes, if you look in the code just above, it's looking to see if DOCKER_TLS_VERIFY is set at all.  If it's set, then it assumes you want TLS verification.
so `unset DOCKER_TLS_VERIFY` or `export DOCKER_TLS_VERIFY=""` would take care of that.

mosquito:
I want to use TLS, but server has self-signed certificate. And I have a problem:
```docker --tls=true --tlscacert=Docker_CA.crt --tlscert=cert.cer --tlskey=key.pem --host=tcp://172.17.10.201:4142 ps
FATA[0002] An error occurred trying to connect: Get https://172.17.10.201:4142/v1.17/containers/json: x509: cannot validate certificate for 172.17.10.201 because it doesn't contain any IP SANs```

**Comment by mosquito** [comment]
The docker cli split thats:

```
$ docker --help
Usage: docker [OPTIONS] COMMAND [arg...]

A self-sufficient runtime for linux containers.

Options:
  ...............
  --tls=false                                     Use TLS; implied by --tlsverify flag
  --tlscacert="/Users/mosquito/.docker/ca.pem"    Trust only remotes providing a certificate signed by the CA given here
  --tlscert="/Users/mosquito/.docker/cert.pem"    Path to TLS certificate file
  --tlskey="/Users/mosquito/.docker/key.pem"      Path to TLS key file
  --tlsverify=false                               Use TLS and verify the remote (daemon: verify client, client: verify daemon)
  ^^^^^^^^^^^^^^^
```

**Comment by cpuguy83** [comment]
Yes, it looks like you need to add the host IP to the server key, or use the hostname that the key was generated for to connect to.

**Comment by mosquito** [comment]
Yes, that's may resolve my problem. But I will have to recreate certificates for all clients, when I will change server ip. And this split for behaviour the fig and the docker cli.

**Comment by cpuguy83** [comment]
So long as the server key is signed by the same CA as the client key, you should not need to change a clients.

**Comment by mosquito** [comment]
Ok. That's true. What about different behavior?

File: compose/cli/docker_client.py [hunk 1]
```diff
@@ -25,7 +26,7 @@ def docker_client():
 
         tls_config = tls.TLSConfig(
             ssl_version=ssl.PROTOCOL_TLSv1,
-            verify=True,
```
mosquito: Value hardcoded. It's ok?

jgrowl:
Is there a technical reason why compose would not respect both DOCKER_TLS and DOCKER_TLS_VERIFY? I would still like to use TLS without verification.

aanand:
DOCKER_TLS doesn't seem to be an environment variable recognised by the Docker daemon.

Arguably it should be (and work exactly like the `--tls` flag), at which point Compose could support it too, but we don't want to invent new DOCKER_* environment variables that aren't supported by the daemon.

mosquito:
https://github.com/docker/docker/blob/master/docker/flags.go#L29

Variable looks like DOCKER_TLS_VERIFY. Should I rename it in the request.

mosquito:
https://github.com/docker/docker/blob/master/docker/flags.go#L29

Variable looks like DOCKER_TLS_VERIFY.

Should I user DOCKER_TLS_STRICT value from DOCKER_TLS_VERIFY?

jgrowl:
As an alternative to using a different environment variable, it is pretty straight forward to add --tls and --tlsverify flags to compose to match docker's behavior. I haven't created a PR just yet as I wanted to do a little more testing. 

See: https://github.com/docker/compose/compare/master...jgrowl:master

There is a warning that would need to be suppressed for a more docker-compose-ish warning:
```
/usr/local/lib/python2.7/dist-packages/requests-2.6.2-py2.7.egg/requests/packages/urllib3/connectionpool.py:768: InsecureRequestWarning: Unverified HTTPS request is being made. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.org/en/latest/security.html
  InsecureRequestWarning)
```

I also had to set the ca to None since it complains when you have a ca file with verify turned off.

aanand:
@jgrowl Looks like a good start - looking forward to the PR.

dnephin:
We've moved all of this logic in to `dockerpy` at `docker.utils.kwargs_from_env`.

Any changes (and further discussion) will need to be opened as a PR on that repo.

Issue #75, opened 2014-02-04
Add support for registry authentication


ahromis:
I was having issues with the cert chain, so I followed what is on this thread for docker-osx: https://github.com/dotcloud/docker/issues/3683 .  I'm able to do a docker login just fine using a new binary on mac os x.

However when using fig and having the Dockerfile pull from the same private registry I get the following error using fig build:

Step 1 : FROM ssl-repo.com/image
Pulling repository ssl-repo.com/image
Error: Authentication is required.

Does fig read the .dockercfg?  Any other ideas on what the issue could be here?

JamesStewy:
@rafael adding `/v1/` to the address still works for me with Docker 1.4.1. Which program (Docker or Fig) doesn't work and what error are you receiving? Also, what is the version of the Docker Registry you are trying to connect to?

mborho:
Adding /v1/ works for me as well with docker 1.0.1 and fig 1.0.1. 

Besides that,  ~/.dockercfg has to be readable for fig....

rafael:
@JamesStewy I'm on docker 1.4.1 and fig 1.0.1. The version of the registry it's 0.8.1. The program that doesn't work is fig. I'm able to pull and push images from the registry when using docker directly. I'm getting the following error:

```
Traceback (most recent call last):
  File "/usr/local/bin/fig", line 9, in <module>
    load_entry_point('fig==1.0.1', 'console_scripts', 'fig')()
  File "/Library/Python/2.7/site-packages/fig/cli/main.py", line 31, in main
    command.sys_dispatch()
  File "/Library/Python/2.7/site-packages/fig/cli/docopt_command.py", line 21, in sys_dispatch
    self.dispatch(sys.argv[1:], None)
  File "/Library/Python/2.7/site-packages/fig/cli/command.py", line 28, in dispatch
    super(Command, self).dispatch(*args, **kwargs)
  File "/Library/Python/2.7/site-packages/fig/cli/docopt_command.py", line 24, in dispatch
    self.perform_command(*self.parse(argv, global_options))
  File "/Library/Python/2.7/site-packages/fig/cli/command.py", line 56, in perform_command
    handler(project, command_options)
  File "/Library/Python/2.7/site-packages/fig/cli/main.py", line 427, in up
    insecure_registry=insecure_registry,
  File "/Library/Python/2.7/site-packages/fig/project.py", line 174, in up
    for (_, container) in service.recreate_containers(insecure_registry=insecure_registry):
[...]

PR #997, opened 2015-02-20
Use docker-py 1.0.0

abesto:
This seems to fix an issue where `docker-compose` tries to connect to private docker registry via HTTP instead of HTTPS, so the auth data in `.dockercfg` doesn't match the URL, causing authentication to fail.

(Also, `1.0.0` is a Good Thing)

Let me know if `< 2.0.0` is too permissive. Minimum version is `1.0.0` because `Client.timeout` is a public attribute from `1.0.0`, breaking the (so far not public) api of `Client._timeout`.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 PyYAML==3.10
-docker-py==0.7.1
+docker-py==1.0.0
 dockerpty==0.3.2
 docopt==0.6.1
 requests==2.2.1
--- a/setup.py
+++ b/setup.py
@@ -30,7 +30,7 @@ install_requires = [
     'requests >= 2.2.1, < 3',
     'texttable >= 0.8.1, < 0.9',
     'websocket-client >= 0.11.0, < 1.0',
-    'docker-py >= 0.6.0, < 0.8',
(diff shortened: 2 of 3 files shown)


abesto:
Oh, the tests failed. Will fix.

dnephin:
LGTM, `2.0.0` should be correct as long as `docker-py` follows semver

aanand:
Does this close #75?

aanand:
OK, it doesn't - that's waiting on https://github.com/docker/docker-py/pull/507. Never mind.

Semver is nice and all, but I'd rather play it safe and keep the minor version sticky, so `< 1.1.0` would be better. We hardcode the version that ends up in the binaries anyhow.

abesto:
Sorry for the late response. Fair enough, restricted to `< 1.1.0`. I can squash the two commits tomorrow if it makes a difference.

aanand:
Yes please!

abesto:
Is done. :)

aanand:
LGTM

abesto:
Yay! Thanks :)

PR #999, opened 2015-02-21
Stack traces when DOCKER_HOST is invalid Issue:945

KingsleyKelly:
Throws an error if the docker client is attempting to use TLS via a unix socket through not setting a Docker Home env variable. Includes test, and new TLSParameterError class.

The Docker Client model throws this error but behaves differently in Compose as TLS is configured prior to creating the Client object in docker_client.py. Suppresses the stack traces through the error class.

This is my first pull request against Compose so let me know if I've made some incorrect assumptions or my style is poor.

### DIFF
--- a/compose/cli/docker_client.py
+++ b/compose/cli/docker_client.py
@@ -1,5 +1,6 @@
 from docker import Client
 from docker import tls
+from . import errors
 import ssl
 import os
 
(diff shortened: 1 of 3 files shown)


dnephin:
Thanks for the contribution!

Sorry it's taken so long for us to respond. This code has moved into https://github.com/docker/docker-py, so we'd need to make this change in that repo.

PR #1005, opened 2015-02-23
Run tests using Docker-in-Docker so we can test multiple versions

aanand:
This build will fail on Wercker because it depends on `docker build -f FILE`, which isn't available in Wercker's Docker version (1.2.0).

Next step: get running on Jenkins!

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,18 @@
 FROM debian:wheezy
-RUN apt-get update -qq && apt-get install -qy python python-pip python-dev git && apt-get clean
+
+RUN apt-get update -qq
+
+# Compose dependencies
+RUN apt-get install -qqy python python-pip python-dev git
+
+# Test dependencies
+RUN apt-get install -qqy apt-transport-https ca-certificates curl lxc iptables
+RUN curl https://get.docker.com/builds/Linux/x86_64/docker-1.3.3 > /usr/local/bin/docker-1.3.3 && chmod +x /usr/local/bin/docker-1.3.3
+RUN curl https://get.docker.com/builds/Linux/x86_64/docker-1.4.1 > /usr/local/bin/docker-1.4.1 && chmod +x /usr/local/bin/docker-1.4.1
+RUN curl https://get.docker.com/builds/Linux/x86_64/docker-1.5.0 > /usr/local/bin/docker-1.5.0 && chmod +x /usr/local/bin/docker-1.5.0
+
+RUN apt-get clean
+
 RUN useradd -d /home/user -m -s /bin/bash user
 WORKDIR /code/
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,17 +1,27 @@
 FROM debian:wheezy
 
-RUN apt-get update -qq
+RUN set -ex; \
+    apt-get update -qq; \
+    apt-get install -y \
+        python \
+        python-pip \
+        python-dev \
+        git \
+        apt-transport-https \
+        ca-certificates \
+        curl \
+        lxc \
+        iptables \
+    ; \
+    rm -rf /var/lib/apt/lists/*
 
-# Compose dependencies
-RUN apt-get install -qqy python python-pip python-dev git
+ENV ALL_DOCKER_VERSIONS 1.3.3 1.4.1 1.5.0
 
-# Test dependencies
-RUN apt-get install -qqy apt-transport-https ca-certificates curl lxc iptables
-RUN curl https://get.docker.com/builds/Linux/x86_64/docker-1.3.3 > /usr/local/bin/docker-1.3.3 && chmod +x /usr/local/bin/docker-1.3.3
-RUN curl https://get.docker.com/builds/Linux/x86_64/docker-1.4.1 > /usr/local/bin/docker-1.4.1 && chmod +x /usr/local/bin/docker-1.4.1
-RUN curl https://get.docker.com/builds/Linux/x86_64/docker-1.5.0 > /usr/local/bin/docker-1.5.0 && chmod +x /usr/local/bin/docker-1.5.0
-
-RUN apt-get clean
+RUN set -ex; \
+    for v in ${ALL_DOCKER_VERSIONS}; do \
+        curl https://get.docker.com/builds/Linux/x86_64/docker-$v -o /usr/local/bin/docker-$v; \
+        chmod +x /usr/local/bin/docker-$v; \
+    done
 
 RUN useradd -d /home/user -m -s /bin/bash user
 WORKDIR /code/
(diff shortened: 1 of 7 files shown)


jfrazelle:
:+1:

script/test
```diff
@@ -1,5 +1,19 @@
-#!/bin/sh
+#!/bin/bash
+# See CONTRIBUTING.md for usage.
+
 set -ex
+
 docker build -t docker-compose .
```
jfrazelle: I would add a tag to each of these that is the commit sha so `docker-compose:$(git rev-parse --short HEAD)`

jfrazelle: otherwise for parallel builds things are going to get tricky

script/test
```diff
@@ -1,5 +1,19 @@
...
+docker build -f Dockerfile.tests -t docker-compose-tests .
+
+if [ "$DOCKER_VERSIONS" == "" ]; then
+  DOCKER_VERSIONS="1.5.0"
+elif [ "$DOCKER_VERSIONS" == "all" ]; then
+  DOCKER_VERSIONS="1.3.3 1.4.1 1.5.0"
+fi
+
+for version in $DOCKER_VERSIONS; do
+  docker run --privileged --rm -e "DOCKER_VERSION=$version" docker-compose-tests nosetests "$@"
```
jfrazelle: then before this line a echo "running container for $version"

jfrazelle:
ping @tianon bash god, this going to be tested on jenkins

jfrazelle:
Can we also keep in mind I'm not just going to put "script/test" into Jenkins I'm going to put the docker commands because who is to say someone won't change this into something that we don't want to execute so can we try to keep it concise :)

jfrazelle:
We can also set this up next week when I am in your neck of the woods \o/

aanand:
OK, I've set it up so that you can kick the whole thing off with a single `docker run`. This requires two levels of Docker nesting (an outer one where we mount `docker.sock`, an inner one where we run Docker-in-Docker), but it works.

You should be able to just run:

```bash
TAG="docker-compose:$(git rev-parse --short HEAD)"

docker build -t "$TAG" .

docker run \
  --rm \
  --volume="/var/run/docker.sock:/var/run/docker.sock" \
  -e "DOCKER_VERSIONS=all" \
  -e "TAG=$TAG" \
  --entrypoint="script/test-versions" \
  "$TAG" \
  --verbose
```

jfrazelle:
You rule!!!!

jfrazelle:
I think its hanging https://jenkins.dockerproject.com/job/Compose%20Master/7/label=ubuntu-aufs-prs/console :( I dont understand

Dockerfile
```diff
@@ -1,5 +1,18 @@
...
+# Compose dependencies
+RUN apt-get install -qqy python python-pip python-dev git
+
+# Test dependencies
+RUN apt-get install -qqy apt-transport-https ca-certificates curl lxc iptables
+RUN curl https://get.docker.com/builds/Linux/x86_64/docker-1.3.3 > /usr/local/bin/docker-1.3.3 && chmod +x /usr/local/bin/docker-1.3.3
+RUN curl https://get.docker.com/builds/Linux/x86_64/docker-1.4.1 > /usr/local/bin/docker-1.4.1 && chmod +x /usr/local/bin/docker-1.4.1
+RUN curl https://get.docker.com/builds/Linux/x86_64/docker-1.5.0 > /usr/local/bin/docker-1.5.0 && chmod +x /usr/local/bin/docker-1.5.0
+
+RUN apt-get clean
```
bfirsh: This needs to be run on the same line as `apt-get install` otherwise it'll actually make the image _bigger_. (The debs are still in the layers before then this adds another layer that needs downloading.)

tianon: It also won't actually do anything: https://github.com/docker/docker/blob/738bdec4908755761b27b7d15baeefb1a824f770/contrib/mkimage/debootstrap#L72-L95

The pattern we usually use is `RUN apt-get update && apt-get install -y ... && rm -rf /var/lib/apt/lists/*`.

script/wrapdocker
```diff
@@ -0,0 +1,98 @@
+#!/bin/bash
+# Adapted from https://github.com/jpetazzo/dind
```
tianon: Why not just use the canonical https://github.com/docker/docker/blob/738bdec4908755761b27b7d15baeefb1a824f770/project/dind instead?

script/test-versions
```diff
@@ -0,0 +1,26 @@
...
+
+set -e
+
+>&2 echo "Running lint checks"
+flake8 compose
+
+if [ "$DOCKER_VERSIONS" == "" ]; then
+  DOCKER_VERSIONS="1.5.0"
+elif [ "$DOCKER_VERSIONS" == "all" ]; then
+  DOCKER_VERSIONS="1.3.3 1.4.1 1.5.0"
```
tianon: If you set this in the `Dockerfile` instead, you could have the entire list of "supported" Docker versions in a single place: (in case that's an interesting property for you)
```Dockerfile
ENV SUPPORTED_DOCKER_VERSIONS 1.5.0 1.4.1 1.3.3
RUN set -e \
    && for v in ${SUPPORTED_DOCKER_VERSIONS}; do \
        curl https://get.docker.com/builds/Linux/x86_64/docker-$v -o /usr/local/bin/docker-$v; \
        chmod +x /usr/local/bin/docker-$v; \
    done
```

tianon:
Looks sane to me :+1:

jfrazelle:
LGTM https://jenkins.dockerproject.com/job/Compose%20Master/label=ubuntu-aufs-prs/8/console merge it and i will turn all on

dnephin:
Nice!  It would be awesome to have this run against multiple python versions as well. Even if it's only python2.6+latest docker, python2.7+latest docker.

At some point we'll want to add python3.4+latest docker

jfrazelle:
merge it merge it merge it

bfirsh:
Hells yeah

bfirsh:
ping @jfrazelle

bfirsh:
When we get the 1.6 RC, I wonder if we can test against that too...

jfrazelle:
You would just add t to the Dockerfile from the looks of it

bfirsh:
Oh, yeah. Assumed that'd be in Jenkins. Nice.

aanand:
Reactivated webhooks, rebased #1088, all looks good.
screen shot 2015-03-17 at 16 32 02

aanand:
We're not validating the DCO any more - that's very bad. Where's the best place to do that? Presumably not inside a `docker run` - we don't put `.git` inside the Docker image.

aanand:
We could mount the code directory as a volume, and then I can add a `script/validate-dco` line to the top of `script/test-versions`. @jfrazelle: sound sensible?

jfrazelle:
ya totally sgtm

aanand:
@jfrazelle Cool - see https://github.com/docker/compose/pull/1128.

PR #1007, opened 2015-02-23
Add -f flag as option on rm.

### LINKED ISSUE #752: Proposal: `fig rm -f` shortcut for `fig rm --force`
Would it be possible to create a shortcut for `fig rm --force`? That would be cool to type four letters less :)

Currently, the output looks as follows:
```
$ fig rm -f
Remove stopped service containers.

Usage: rm [options] [SERVICE...]

Options:
    --force   Don't ask to confirm removal
    -v        Remove volumes associated with containers
```

Thanks,
Pawel

KingsleyKelly:
This fixes #752

Just added another flag to rm and a test. I'm not sure if this is addressed by any other pull requests, but I couldn't find them.

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -238,8 +238,8 @@ class TopLevelCommand(Command):
         Usage: rm [options] [SERVICE...]
 
         Options:
-            --force   Don't ask to confirm removal
-            -v        Remove volumes associated with containers
+            -f, --force   Don't ask to confirm removal
+            -v            Remove volumes associated with containers
         """
         all_containers = project.containers(service_names=options['SERVICE'], stopped=True)
         stopped_containers = [c for c in all_containers if not c.is_running]
--- a/tests/integration/cli_test.py
+++ b/tests/integration/cli_test.py
@@ -295,6 +295,12 @@ class CLITestCase(DockerClientTestCase):
         self.assertEqual(len(service.containers(stopped=True)), 1)
         self.command.dispatch(['rm', '--force'], None)
         self.assertEqual(len(service.containers(stopped=True)), 0)
+        service = self.project.get_service('simple')
+        service.create_container()
+        service.kill()
+        self.assertEqual(len(service.containers(stopped=True)), 1)
+        self.command.dispatch(['rm', '-f'], None)
+        self.assertEqual(len(service.containers(stopped=True)), 0)
 
     def test_kill(self):
         self.command.dispatch(['up', '-d'], None)


albers:
Please give me a ping if this is accepted so that I can add the flag to bash completion.

compose/cli/main.py
```diff
@@ -238,15 +238,15 @@ def rm(self, project, options):
...
+            -f, --force   Don't ask to confirm removal
+            -v            Remove volumes associated with containers
         """
         all_containers = project.containers(service_names=options['SERVICE'], stopped=True)
         stopped_containers = [c for c in all_containers if not c.is_running]
 
         if len(stopped_containers) > 0:
             print("Going to remove", list_containers(stopped_containers))
-            if options.get('--force') \
+            if options.get('--force') or options.get('-f') \
```
aanand: Is this necessary? I believe docopt normalises `-f` to `--force`.

KingsleyKelly: You are entirely right. I seem to have fixed a bug that was fixed! Sorry, I also tested this against my current version of fig where the flag failed and thought the above had fixed it. I can close this pull request if that makes the most sense.

aanand: No need. You can make a change while keeping it to one commit with `git commit --amend`, and then push that to your branch with `git push --force`.

aanand: To be clear: it's your change to the *docs* for rm that causes `-f` to work as a short version of `--force`. That's how docopt works - you write the docs and it implements the flags.

But we don't need the check for `options.get('-f')` because docopt is smart and just sets `options['--force']` if it sees `-f`. So this line can be reverted.

KingsleyKelly: Ok that's awesome. Have amended the commit.

aanand:
LGTM

dnephin:
LGTM, but it seems like the wercker build hasn't run against this PR for some reason?

KingsleyKelly:
I tried amending the commit and opening a new pull request, neither of which triggered wercker. I'm unfamiliar with the service, is this likely to be an error on my end?

dnephin:
Nope, I don't think it's on your end. I know @aanand is working on having the test suite run on jenkins, I'm not sure if wercker integration was disabled, or if there is an error on their end.

aanand:
Looks like Wercker is still running but we've lost the indicator, possibly because I added hooks for Jenkins and it's confusing GitHub about what to show.

[The build was successful](https://app.wercker.com/#build/54f01bead9b1463663464b61), so I'm going to merge.

PR #1010, opened 2015-02-24
Log "creating container" when scaling

bfirsh:
If an image needs pulling, it just looks like it's hanging.


aanand:
LGTM

PR #1011, opened 2015-02-24
Add a --pid=host feature to expose the host PID space to the container

sdake:
Docker 1.5.0 introduces a --pid=host feature which is fantastic for monitoring
containers from the baremetal system or atomically upgrading or rolling back
a deployed application.

For more details of a real-life use case, check out:
http://sdake.io/2015/01/28/an-atomic-upgrade-process-for-openstack-compute-nodes/

Docker-DCO-1.1-Signed-off-by: Steven Dake <stdake@cisco.com> sdake@github

### DIFF
--- a/compose/config.py
+++ b/compose/config.py
@@ -20,6 +20,7 @@ DOCKER_CONFIG_KEYS = [
     'links',
     'mem_limit',
     'net',
+    'pid',
     'ports',
     'privileged',
     'restart',
--- a/compose/service.py
+++ b/compose/service.py
@@ -25,6 +25,7 @@ DOCKER_START_KEYS = [
     'dns_search',
     'env_file',
     'net',
+    'pid',
     'privileged',
     'restart',
 ]
@@ -434,6 +435,7 @@ class Service(object):
         privileged = options.get('privileged', False)
         cap_add = options.get('cap_add', None)
         cap_drop = options.get('cap_drop', None)
+        pid = options.get('pid', None)
 
         dns = options.get('dns', None)
         if isinstance(dns, six.string_types):
@@ -457,6 +459,7 @@ class Service(object):
             restart_policy=restart,
             cap_add=cap_add,
             cap_drop=cap_drop,
+            pid_mode=pid
         )
 
     def _get_image_name(self, image):
--- a/docs/yml.md
+++ b/docs/yml.md
@@ -264,6 +264,16 @@ net: "none"
 net: "container:[name or id]"
 net: "host"
 ```
+### pid
+
+```
+pid: "host"
+```
+
+Sets the PID mode to the host PID mode.  This turns on sharing between
+container and the host operating system the PID address space.  Containers
+launched with this flag will be able to access and manipulate other
+containers in the bare-metal machine's namespace and vise-versa.
 
 ### dns
 
--- a/tests/integration/service_test.py
+++ b/tests/integration/service_test.py
@@ -419,6 +419,17 @@ class ServiceTest(DockerClientTestCase):
         container = create_and_start_container(service)
         self.assertEqual(container.get('HostConfig.NetworkMode'), 'host')
 
+    def test_pid_mode_none_defined(self):
+        service = self.create_service('web', pid=None)
+        container = create_and_start_container(service)
+        print 'STEAK %s' % (container.get('HostConfig.PidMode'))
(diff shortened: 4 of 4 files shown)


sdake:
Would a maintainer mind adding feedback on this pull request?

Thanks
-steve

compose/container.py
```diff
@@ -14,6 +14,7 @@ def __init__(self, client, dictionary, has_been_inspected=False):
         self.client = client
         self.dictionary = dictionary
         self.has_been_inspected = has_been_inspected
+        self._version = '1.17'
```
dnephin: I believe this line is unnecessary

compose/cli/docker_client.py
```diff
@@ -32,4 +32,4 @@ def docker_client():
         )
 
     timeout = int(os.environ.get('DOCKER_CLIENT_TIMEOUT', 60))
-    return Client(base_url=base_url, tls=tls_config, version='1.14', timeout=timeout)
+    return Client(base_url=base_url, tls=tls_config, version='1.17', timeout=timeout)
```
dnephin: I believe this will be blocked on moving the build to run on jenkins. Wercker doesn't support later versions of docker

compose/container.py
sdake: Reading the docker-py clinet.py code, it may be appear your correct.  I did actually test this in a real world environment, and I thought everything was necessary at the time, but I was doing the job at 3am, so its possible I made an error.

iI'll test without it next week and either submit results or submit a patch revision with this change reverted.

compose/cli/docker_client.py
sdake: That is a pretty serious sadface :(  Is there a published schedule for that transition or an unpublished estimated date?

Thanks
-steve

dnephin: cc @aanand 

There was an issue about something related, but it's moved to the backlog now. I don't know of a related issue I can link you.

sdake:
@dnephin Thanks a whole bunch for the review!  I'll either fix the changes or provide a rationale why line

dnephin:
Great, thanks for the contribution!

compose/cli/docker_client.py
aanand: Jenkins is still WIP, but it's a priority.

aanand:
Looks fine, but blocking on Jenkins :( soon, hopefully.

sdake:
@dnephin I addressed your comments and have tested the rebased master works properly for the pid:host use case.

Regards
-steve

Dockerfile
```diff
@@ -15,7 +15,7 @@ RUN set -ex; \
     ; \
     rm -rf /var/lib/apt/lists/*
 
-ENV ALL_DOCKER_VERSIONS 1.3.3 1.4.1 1.5.0
+ENV ALL_DOCKER_VERSIONS 1.5.0
```
dnephin: We still want to test against older versions, but I think we can disable the tests that we expect to fail when the version is older.


If we accept api `version` from the environment then we can set it based on the version we're testing against.

compose/cli/docker_client.py
dnephin:
```python

version=os.environ.get('COMPOSE_API_VERSION', '1.17')
```

sdake:
This patch has been rebased onto master which runs docker-dev downloaded from curl.

sdake:
@dnephin Any chance you can take another look and merge this pull request?  I think it is ready to go.

Regards
-steve

dnephin:
LGTM

I'll wait for @aanand to have another look and merge

aanand:
I like it - just needs a rebase and an integration test.

sdake:
@aanand rebased and passing master test cases - will add functional test today/tomorrow.

sdake:
@aanand Added two integration tests.  One tests that the docker daemon returns a pidmode=net when passed the pid=host operation via service creation, and one tests that docker daemon returns '' for pidmode when passed None (which is the other condition the service creates.  Testing that pidMode actually works correctly is a job for the docker test cases - I think we are covered there.

Please advise if you would like additional work.

Thanks!
-steve

aanand:
Yes, absolutely - if the container has the right configuration set, Compose has done its job as far as testing goes. Thanks.

LGTM

PR #1017, opened 2015-02-25
Document Swarm integration

aanand:
While Swarm integration is in a primitive state, we won't talk about it in the official Compose docs.


bfirsh:
Shall we link to it from the roadmap?

aanand:
Done.

bfirsh:
Woot!

bfirsh:
swarm

PR #1027, opened 2015-02-25
Ship 1.1.0

aanand:


CHANGES.md
```diff
@@ -1,21 +1,8 @@
 Change log
 ==========
 
-1.1.0-rc2 (2015-01-29)
```
thaJeztah: Is there still a full history available somewhere after this has been removed? It's sometimes handy to be able to check when/in which version something was fixed/introduced.

aanand: We weren't planning on it, but the git tags will stick around, so that'll help when narrowing things down.

bfirsh:
LGTM

marksteve:
:tada:

PR #1038, opened 2015-02-27
Add container IP address to "docker-compose ps" output

cschwede:


compose/container.py
```diff
@@ -63,6 +63,11 @@ def number(self):
             return None
 
     @property
+    def ip(self):
+        self.inspect_if_not_inspected()
```
dnephin: You don't need to call `self.inspect_if_not_inspected()` explicitly anymore, it's called by `self.get(...)`

dnephin:
Some explanation about why you'd like to make this change would be good.

My initial thought is that this table is already pretty full, and I'm not sure about including IP address by default.  Maybe an option to included extended details could work.

jbarlow83:
@dnephin It is not obvious what IP addresses containers will be assigned to. After poking around I found that `docker inspect $(docker ps -q) | grep IPA` can answer this question, but I expected to find information about containers started by -compose within -compose.

dnephin:
Thanks for the PR.  We've had a bunch of different requests for `ps`, so I'm going to aggregate them all into #2369 so it'll be easier to design a solution that works for most of them.

I don't think we'll be able to accept this PR as is.

PR #1054, opened 2015-03-03
possiblity to override the client version 

httpPrincess:
it is not the best idea to hard code the version of the client. At least the possibility to override this should be provided.

### DIFF
--- a/compose/cli/docker_client.py
+++ b/compose/cli/docker_client.py
@@ -32,4 +32,5 @@ def docker_client():
         )
 
     timeout = int(os.environ.get('DOCKER_CLIENT_TIMEOUT', 60))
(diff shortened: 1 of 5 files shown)


dnephin:
I think we should make this change, the only question is about what the environment variable should be called.  I believe @aanand has suggested `COMPOSE_CLIENT_VERSION`.

Please squash your changes into a single commit

httpPrincess:
as you wished

dnephin:
I'm not sure why that first commit from albers is showing up. I think that's already in master?

httpPrincess:
I had some strange problems with squashing, perhaps this is the reason why i appears here.

dnephin:
You should be able to rebase with master to get rid of it

httpPrincess:
forget it.

PR #1077, opened 2015-03-06
Add ps -s option to print service names.

tpounds:
A new -s option to support printing names in addition to images. Useful in scripting scenarios since compose uses service names natively for most commands.


albers:
:+1: 
This should also be added to bash completion [here](https://github.com/docker/compose/blob/master/contrib/completion/bash/docker-compose#L177).

tpounds:
Updated with bash completion support. Thanks @albers!

albers:
Completion LGTM. Thanks!

dnephin:
LGTM

aanand:
Could you add it to zsh as well, now that we've merged #1325?

aanand:
(Also, could `ps -s` be used by the completion scripts instead of hacky manual YAML file parsing?)

albers:
@aanand Sadly, `ps -s` won't help here as it only lists existing containers. For completion, I also need the names of services without existing containers. I do not know of any other way to get the full service list other than parsing the yaml.

What I really miss here is some kind of `ps -a -s` that lists services regardless of container creation state.

tpounds:
@aanand Updated with zsh completion. I'm going to leave the YAML parsing as-is for now per @albers feedback. LMK if I should address anything else.

albers:
@aanand @tpounds What do you think of an `-a` option that is only valid when `-s` is given that makes compose output the full list of services?
The new `-s` option is nice, but as it only cares for services with existing instances, it only covers a subset of scripting scenarios. See my above mentioned usecase in bash completion: extractig the full service list.

aanand:
I've put some thoughts down in https://github.com/docker/compose/issues/1498 - let's move the discussion there.

aanand:
LGTM

dnephin:
LGTM

aanand:
Thinking about it, since this overlaps with the requirements laid out in #1498, I retract my LGTM - we should figure out the best way to serve all these needs, which might involve an entirely new command.

tpounds:
@aanand @albers I agree a more general command as mentioned in #1498 would be useful for many scenarios. Do you think there is any value in providing the context aware options to `docker-compose ps` in addition to a new command as well? My thought is something consistent with what `docker ps` provides.

aanand:
@tpounds It could be useful to support more of `docker ps`' options in `docker-compose ps`, but I don't think they would share any code.

For example:

- `docker-compose services --filter` as discussed in #1498 would be filtering *services* and doing it *client-side* by looking at the YAML file and what's running.
- `docker ps --filter` just passes filters for *containers* along to the Docker API, which will filter them *server-side*.

dnephin:
I think there is value in this as-is (use case https://github.com/docker/compose/issues/374#issuecomment-126401655).

dnephin:
I think the proposal in #1953 might be more appropriate for this feature. `ps` is the status of containers, so just displaying service names doesn't fit all that well.

dnephin:
Thanks for the contribution!

Sorry it's taken us a while to respond. We'd like to support this feature with the `docker-compose config` command proposal in #1953.

Issue #988, opened 2015-02-19
Proposal: Types, imports and external services


aanand:
This proposal describes an enhancement to `docker-compose.yml` that will enable the importing of selected bits of configuration from other Compose configuration files, and the definition of ‘external’ or ‘dummy’ services. It is intended to serve three use cases:

1. I deploy my app in multiple environments - development, staging, production. Some configuration is different in each environment; some services only need to exist in some environments. I want to maintain a configuration file for each environment without repeating myself.

2. I have multiple apps which perform different tasks but re-use some services. I don’t want to repeat the definitions of those services in each app’s configuration.

3. I have a defined dependency, such as a database, in my app. When I’m developing the app, I want Compose to spin up a local database container to link to from my other services. When running the app in production, I instead want to point my other services at an already-running, separately-managed database service, which may not be managed with Compose or even Docker.

It is **related** to #495/#845, in that use case 1 is partially served by parameterisation of configuration values. It does not replace that functionality, but complements it.

It is **distinct** from #318/#758, in that it does not serve this use case:

- I have two apps. App A has link dependencies on app B. When I type `docker-compose up` in app A’s directory, I want Compose to first spin up app B and *then* app A, with cross-app links in place.

However, that use case could be served with an implementation that builds on the enhancements proposed here.

## Service types

A new configuration key, `type`, can be specified on a service defined in `docker-compose.yml`. It’s optional, and its value defaults to `"container"`. There are three possible types of value.

### "container"

Denotes that this service consists of one or more homogenous Docker containers, configured using the options specified here. This is exactly how `docker-compose.yml` services are defined today.

```yaml
web:
  type: container # optional
  build: .
  ports:
    - 80:8000
  links:
    - db
```

### Path to another Compose file

Denotes that this service is defined in another file. The file’s path, and the name of the service within that file, are supplied here.

```yaml
db:
  type: common.yml#db
```

Assuming `common.yml` defines a `db` service, this is equivalent to copying and pasting its configuration here.

Configuration can also be overridden:

```yaml
db:
  type: common.yml#db
  environment:
    POSTGRES_USER: devpass
    POSTGRES_PASSWORD: devuser
```

### "external"

Denotes an externally-defined service. Its location, and any other configuration, are supplied here.

```yaml
db:
  type: external
  host: 1.2.3.4
  port: 5432
  environment:
    POSTGRES_USER: produser
    POSTGRES_PASSWORD: prodpass
```

[...]

PR #1080, opened 2015-03-07
WIP: Service types

aanand:
This is an unfinished first attempt at implementing the "service types" proposed in #988, to feel out what it's like to implement. It differs from what's proposed there in a few ways:

1. Instead of specifying `type: path/to/file.yml#servicename`, you go:

    ```yaml
    type: copy
    path: path/to/file.yml
    service: servicename
    ```

   This is mostly to speed up implementation for now.

2. The `copy` type is equivalent to plucking a single service dictionary out of another file and merging the two (including the smart merging of links and environment variables). It doesn't pull in transitive dependencies or look at any other services in the other file.

   Service names in `links`/`volumes_from` therefore have "dynamic scope", not "lexical scope" - if I pull in something with `links: ["db"]`, it'll get linked to my locally-defined `db` service.

   This is much simpler than trying to implement a full-on import system, but doesn't cover as many use cases.

3. What #988 calls `external`, I here call `remote`. It feels like a better name to me, but I'm willing to be persuaded otherwise.

### DIFF
--- a/compose/cli/command.py
+++ b/compose/cli/command.py
@@ -4,9 +4,9 @@ from requests.exceptions import ConnectionError, SSLError
 import logging
 import os
 import re
-import yaml
 import six
 
+from .. import config
 from ..project import Project
 from ..service import ConfigError
 from .docopt_command import DocoptCommand
@@ -69,18 +69,11 @@ class Command(DocoptCommand):
             return verbose_proxy.VerboseProxy('docker', client)
         return client
 
-    def get_config(self, config_path):
-        try:
-            with open(config_path, 'r') as fh:
-                return yaml.safe_load(fh)
-        except IOError as e:
-            raise errors.UserError(six.text_type(e))
-
     def get_project(self, config_path, project_name=None, verbose=False):
         try:
-            return Project.from_config(
+            return Project.from_dicts(
                 self.get_project_name(config_path, project_name),
-                self.get_config(config_path),
+                config.load(config_path),
                 self.get_client(verbose=verbose))
         except ConfigError as e:
             raise errors.UserError(six.text_type(e))
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -12,7 +12,8 @@ import dockerpty
 
 from .. import __version__
 from ..project import NoSuchService, ConfigurationError
-from ..service import BuildError, CannotBeScaledError, parse_environment
+from ..service import BuildError, CannotBeScaledError
+from ..config import parse_environment
 from .command import Command
 from .docopt_command import NoSuchCommand
 from .errors import UserError
--- a/compose/config.py
+++ b/compose/config.py
@@ -51,7 +51,8 @@ DOCKER_CONFIG_HINTS = {
 
 
 def load(filename):
-    return from_dictionary(load_yaml(filename))
+    working_dir = os.path.dirname(filename)
+    return from_dictionary(load_yaml(filename), working_dir=working_dir)
 
 
 def load_yaml(filename):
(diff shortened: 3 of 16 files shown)


bfirsh:
This looks like a great start. Some initial thoughts:

1) `path` and `service` being top-level config items seems ambiguous. Perhaps they could be child items of type somehow, or perhaps grouped under some other key

2) I don't like automatically including links from other files because the scope it is not explicit and can cause breakage.

e.g. `development.yml`:
```
web:
   build: .
   link: db
db:
   image: postgres
```

`production.yml`:
```
web:
   type: copy
db:
   # ... some production database
```

Just looking at `production.yml`, it is not obvious that the web server depends on the db. Also, if I change the name of `db` in `development.yml`, it will presumably silently break.

This is an implicit/explicit problem. Implicitly including links is going to cause breakage and ambiguous scope. I can imagine always having to run `docker-compose run web env` to figure out what is being linked to, instead of it being clearly listed under the `links` key.

What's the downside of always explicitly defining links? I can't think of one.

3) I much prefer the word "include". It describes really well what is going on here. I don't think "I need to copy my web service into this file", I think "I want to include my web service".

4) "remote" implies remote over a network, which is not necessarily true.

I think writing some docs might help us iron out some of the design and naming issues.

aanand:
> `path` and `service` being top-level config items seems ambiguous. Perhaps they could be child items of type somehow, or perhaps grouped under some other key

Maybe the type *is* the key:

```yaml
myweb:
  copy:
    path: common.yml
    service: web
```

> Implicitly including links is going to cause breakage and ambiguous scope.

Agreed; that's why an import system which imports dependencies is better. Still, it'll take much more time to implement, and if we can address a significant number of use cases without doing so, great. So let's go with no importing of links.

What should we do if we encounter a `links` option in the service you're trying to include? I'm inclined to say it should be an error.

> I much prefer the word "include". It describes really well what is going on here.

With regards to naming, I'm trying to come up with names that won't clash with things we might want to do in future. See the other use cases described in #988 - two potential ones are including a service along with its dependencies, and linking to a service in another Compose app.

If we implement those, and decide to invent new types for them, they'll need names too. Perhaps this is overly cautious, but I don't want us to paint ourselves into a corner.

Still, if we implement the inclusion of dependencies later, it could be done as an additional option:

```yaml
myweb:
  include:
    path: common.yml
    service: web
    include_links: true
```

So maybe `include` is fine - it's certainly more descriptive.

Another thing worth noting: in languages where the word `include` appears today (off the top of my head, C and PHP), it means "include an entire file". People are going to be surprised when it doesn't do that in Compose.

> I think writing some docs might help us iron out some of the design and naming issues.

Agreed. I'll draft a quick usage guide.

bfirsh:
```
myweb:
  include:
    path: common.yml
    service: web
```
:+1: Love this. Including was never really a type anyway. The type is "Docker container" or whatever.

It's a great idea having extra options. With different words, you have to try and remember what those different words do. (Does copying do X? Or is that including?)

> Another thing worth noting: in languages where the word include appears today (off the top of my head, C and PHP), it means "include an entire file".

Maybe "extends" is better? I like how that makes them feel a bit like classes. Extending, redefining attributes, etc etc.

> What should we do if we encounter a links option in the service you're trying to include? I'm inclined to say it should be an error.

It could just not copy links. You always redefine links.

aanand:
`extends` is good.

It occurs to me now that you should be able to not specify a file path, and thus refer to a service defined in the current file. People have already asked for that feature.

Also, maybe `path` should be called `file`.

> It could just not copy links. You always redefine links.

Sure, I'm just anticipating surprised users - a lot of "my links don't work with `extends`" bug reports, much like all the "my ports don't work with `fig run`" reports. But since you can immediately redefine them, maybe it's not going to annoy people too much.

bfirsh:
> Also, maybe path should be called file.

:+1:

> I'm just anticipating surprised users - a lot of "my links don't work with extends" bug reports, much like all the "my ports don't work with fig run" reports. 

An Engine feature that would make this much better is if image could say what links or environment variables it needs setting and refuse to start without them set. If you included a web service without linking to the db, it would then say "hey, you need to connect this web service to a db".

In the meantime, I think doing something as simple as possible first, then improving it is best. I think we can get away with some clear documentation and a warning ("hey, you extended a service which was linked to "db", but you haven't linked to that in the extended version. you need to explicitly set links.").

aanand:
First attempt at docs for `extends`: #1088

aanand:
Work now continues in #1088.

Issue #961, opened 2015-02-12
Add support of --read-only docker flag


spiddy:
Now that we have the possibility to run docker container on read-only filesystem with docker v1.5 It would be really cool to have it on docker-compose.yml format also.

That would be very useful in order to create _immutable infrastructures_.

compose.yml example format could follow the same format as privileged boolean flag:

<pre>
nginx:
  image: nginx:1.7.9
  read_only: true
</pre>

spiddy:
pending of docker/docker-py#483 to be released in order to get a PR ready.

diogomonica:
@spiddy what is the status of this?

spiddy:
There was another PR done for the job, but since it was stuck with a conflict, I pushed my changes. hope this doesn't confuse people.

aanand:
Closed by https://github.com/docker/compose/pull/1335

PR #1084, opened 2015-03-09
add support for read_only parameter

lsde:
add support for specifying read_only flag as per https://github.com/docker/compose/issues/961


aanand:
Nice, thanks! Needs a test, some documentation and your commit needs signing off.

lsde:
ok, added test, simple docs and signed.

lsde:
The problem is that Wrecker doesn't have 1.5 box to make tests pass, what would you propose to get this merged?

dnephin:
This was done in #1335

Issue #129, opened 2014-03-03
Validate Compose file and produce better error messages


aanand:
There are a lot of ways to confuse Compose with a badly-formed `docker-compose.yml`. This file should be checked when it is read and good error messages should be produced to help users fix the problem.

This could be done by producing a schema for the file, like was done in https://github.com/docker/compose/pull/1348.

These sorts of things should be checked:

- [x] Top-level object is a dict
- [x] Each key (a service name) maps to a dict (https://github.com/orchardup/fig/pull/128)
- [ ] String config options (e.g. `image`) - are numbers allowed? When? Don't accept YAML's eager interpretation of values (e.g. `true`) which should be quoted (#1788)
- [x] List config options (e.g. `volumes`) either *must* be a list, or we helpfully wrap single values in a list for you
- [x] String-or-list config options (e.g. `command`, `entrypoint`)
- [x] Dictionary config options (e.g. `environment`)
- [x] Format of various options (e.g. `volumes` - see #260)

References: #117 #127

draghuram:
@aanand, Is there any consensus on the best way to implement the validation? I wanted to check first before I give it a shot.

dnephin:
Not at all consensus, but I have been thinking that `jsonschema` would be a good option. It's language agnostic (so other tools interested in supporting the compose format can do so), and it should support validation for all the cases here.

draghuram:
@dnephin, I will try to come up with schema using "JSON schema" and we can take it from there.

funkyfuture:
@draghuram what's your status on this? *maybe* i will have the time to take care of this within the next days.

draghuram:
@funkyfuture As you can see, I created a PR with all the work that I have done. Basically, there is a JSON schema file and code to validate the configuration against that. At some point, I was able to run all the tests successfully but that was some time back. I think there are two more things that need to be done:

- Write tests covering invalid configurations
- Make sure that Schema covers everything that we need.

Please feel free to build on top of this PR as I may not be able to spend time on this issue.

kumavis:
+1 to keeping yaml
+1 to validating the schema

mnowster:
I majority of these concerns should now be covered/addressed by https://github.com/docker/compose/pull/1808

KamilKopaczyk:
Could it be more strict? Perhaps allowed values for keys such as _restart_?

Lately i was investigating why one of my containers isn't restarting, when others do. I looked at my yaml file and everything was ok. Except for a typo i didn't see ;)
[...]

PR #1087, opened 2015-03-10
Allows empty `links` in YAML configuration file.

### LINKED ISSUE #1086: Cannot service with an empty list of links
The following config drives *compose* crazy:

```yaml
service:
  build: .
  links:
```

Altough this is valid YAML., *compose* raises the following Exception: `TypeError: 'NoneType' object is not iterable`

This should be allowed.

tristan0x:
Fixed `TypeError: 'NoneType' object is not iterable` issue when service declares an empty list of `links`.
Fixes #1086 

Signed-off-by: Tristan Carel <tristan@cogniteev.com>

### DIFF
--- a/compose/project.py
+++ b/compose/project.py
@@ -17,7 +17,7 @@ def sort_service_dicts(services):
     sorted_services = []
 
     def get_service_names(links):
-        return [link.split(':')[0] for link in links]
+        return [link.split(':')[0] for link in links or []]
 
     def visit(n):
         if n['name'] in temporary_marked:
@@ -118,7 +118,7 @@ class Project(object):
     def get_links(self, service_dict):
         links = []
         if 'links' in service_dict:
-            for link in service_dict.get('links', []):
+            for link in service_dict.get('links', []) or []:
                 if ':' in link:
                     service_name, link_name = link.split(':', 1)
                 else:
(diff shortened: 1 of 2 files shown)


aanand:
Thanks, but I'd rather not allow `links` to be `None`. It'd be better to throw a helpful error.

tristan0x:
Yet, this is valid YAML:

```yaml
console:
   image: busybox:latest
   links:
   command: /bin/sleep 300
```

In that case, `links` are `None`. It is what the yaml library returns. But it still means **empty list of links**.

Do you expect compose to throw an helpful error for such config?

If you don't, here is a patch proposal!
I understand you prefer a clean model in `compose.project.Project` class, without stuff like `None` `links`. So what about *fixing* the YAML returned by `compose.cli.command.Command.get_config` member method by replacing `None` by `[]` for a bunch of keys:

- links
- external_links
- ports
- volumes
- volumes_from
- env_file

mgoelzer:
@tristan0x:  Why was this closed without merging?  I often run into this "NoneType is not iterable" problem when I forget to remove an empty YAML key, and your change seemed like a step in the right direction.  If anything, I think your change should be expanded to the other YAML keys (as you suggest in your Mar 12 comment).

tristan0x:
@mgoelzer : This issue is part of something bigger which is configuration validation. Please see #129
Thank anyway for your support :)

PR #1097, opened 2015-03-12
Add script/shell

aanand:
Useful for running tests repeatedly without going through a `docker build` every time - even when cached, it's a pain for quickly iterating on unit tests.


bfirsh:
LGTM

PR #1099, opened 2015-03-12
Fix env file resolution

aanand:
Environment variable file paths should be considered relative to the compose file, not to the current directory. These are usually the same, unless the user specifies an alternate one in a different directory with `-f`.


compose/config.py
```diff
@@ -0,0 +1,230 @@
...
+
+    if 'links' in base or 'links' in override:
+        d['links'] = merge_links(base, override)
+
+    for k in ALLOWED_KEYS:
+        if k not in ['links', 'environment']:
+            if k in override:
+                d[k] = override[k]
+
+    return d
```
dnephin: It looks like this (`merge_service_dicts`) isn't used yet. I guess this was pulled in from the other branch?

dnephin:
I like it

What about host volume binds with relative paths?  Do you plan on addressing those in a separate branch? It seems like volume paths should probably also be relative to the `docker-compose.yml` to be consistent.

aanand:
Yes, good point. I'll update it so volume paths are treated similarly.

(That's going to be fun, because we expand `~` and environment variables in volume paths.)

aanand:
After some thought, I'm going to do volume paths separately. This is rebased and ready to go.

bfirsh:
LGTM

Issue #1092, opened 2015-03-11
No output from "run" if STDIN is not a TTY


abesto:
We hit this issue when using `docker-compose run` on Jenkins to run integration tests.

```sh
echo -e "main:\n  image: hello-world" > docker-compose.yml
docker-compose run main < /dev/null
```
produces no output at all. Adding `-T` makes no difference.

A hack to work around the issue (on OSX), sometimes prints only a character "H", sometimes the proper output:
```sh
echo -e "main:\n  image: hello-world" > docker-compose.yml
script -q /dev/null docker-compose run main < /dev/null
```

The same on Linux:
```sh
echo -e "main:\n  image: hello-world" > docker-compose.yml
script -q -c "docker-compose run main" /dev/null < /dev/null
```

bschmeck:
I've run into the same problem, and just opened #1103 to stop creating interactive dockerptys when running in an environment without a TTY.

The reason your given examples don't work when given `-T` is because dockerpty expects [multiplexed streams](https://docs.docker.com/reference/api/docker_remote_api_v1.9/#attach-to-a-container) when running without a TTY.  Dockerpty then closes after failing to demux STDOUT.

Dockerpty's demuxer implementation is [here.](https://github.com/d11wtq/dockerpty/blob/master/dockerpty/io.py#L147))

emcfins:
# dibs

strokirk:
Same for us, running `docker-compose -p ${PROJECT_ID} run -T --rm web py.test --create-db -x` produces no output, which makes figuring out which test failed take extra effort.

bschmeck:
FWIW, I traced this issue back to dockerpty: https://github.com/d11wtq/dockerpty/issues/32

When running a manually patched version of dockerpty, I see the full output from Jenkins and haven't run into any problems.

emk:
Thank you for figuring this out!

We're also hitting this error when running `docker-compose` under GoCD. Unfortunately, it makes it impossible to debug what our integration/deployment server is doing with Docker.

What steps need to be taken to get all the necessary patches written and merged? I can spend some cycles on this if it helps. `docker-compose` is seriously great, and we'd love to get this fixed.

aanand:
https://github.com/d11wtq/dockerpty/issues/32 sounds reasonable to me. Let's give it a shot and see if it breaks anything.

emk:
[...]

PR #1103, opened 2015-03-13
Use the tty variable to specify whether to create an interactive dockerpty.

bschmeck:
When run in an environment without a TTY, compose will launch docker containers without ttys, but will still create interactive dockerptys, unless the -T flag is given.  This PR ensures that whenever compose launches containers without ttys, the dockerptys are non-interactive.

I'm not sure how to write tests for this, but am happy to try if I can be pointed in the right direction.

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -341,7 +341,7 @@ class TopLevelCommand(Command):
             print(container.name)
         else:
             service.start_container(container, ports=service_ports, one_off=True)
-            dockerpty.start(project.client, container.id, interactive=not options['-T'])
+            dockerpty.start(project.client, container.id, interactive=tty)
             exit_code = container.wait()
             if options['--rm']:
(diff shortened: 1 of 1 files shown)


aanand:
Is this logic correct? You *can* do interactive non-TTY containers with Docker, using `docker run -i` without `-t`.

bschmeck:
The logic might not be correct, I'm a little out of my depth.  (I can't think of a reason you'd want an interactive non-TTY container.)

I was surprised to find that by default compose starts a container without a TTY, yet expects an interactive PTY.  This bit me when trying to use compose with Jenkins, and based on #1092 I'm not the only one who's run into this.

Maybe the proper solution is to just call greater attention to this in the documentation for the -T flag?

dnephin:
I think we have this fixed in docker-compose 1.3

PR #1112, opened 2015-03-13
Remove wercker status from readme

bfirsh:
Because it doesn't really work anymore and looks scary when it's broken. Jenkins is coming soon!


dnephin:
LGTM

Issue #1082, opened 2015-03-08
Support request: tests fail due port-blocking container


funkyfuture:
whenever i run tests with `docker run -v /var/run/docker.sock:/var/run/docker.sock --rm --entrypoint nosetests docker-compose`, two tests are failing. from what i read i'd say that there's still a container listening to the same port that is tried to bound to.

```
.....Estty: standard input: Inappropriate ioctl for device
.stty: standard input: Inappropriate ioctl for device
.stty: standard input: Inappropriate ioctl for device
........E............................................................................................................................................................................
======================================================================
ERROR: test_port (tests.integration.cli_test.CLITestCase)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/code/tests/integration/cli_test.py", line 379, in test_port
    self.command.dispatch(['up', '-d'], None)
  File "/code/compose/cli/command.py", line 27, in dispatch
    super(Command, self).dispatch(*args, **kwargs)
  File "/code/compose/cli/docopt_command.py", line 24, in dispatch
    self.perform_command(*self.parse(argv, global_options))
  File "/code/compose/cli/command.py", line 59, in perform_command
    handler(project, command_options)
  File "/code/compose/cli/main.py", line 443, in up
    do_build=not options['--no-build'],
  File "/code/compose/project.py", line 185, in up
    do_build=do_build):
  File "/code/compose/service.py", line 261, in recreate_containers
    self.start_container(container)
  File "/code/compose/service.py", line 342, in start_container
    cap_drop=cap_drop,
  File "/code/compose/container.py", line 121, in start
    return self.client.start(self.id, **options)
  File "/usr/local/lib/python2.7/dist-packages/docker/client.py", line 867, in start
    self._raise_for_status(res)
  File "/usr/local/lib/python2.7/dist-packages/docker/client.py", line 95, in _raise_for_status
    raise errors.APIError(e, response, explanation=explanation)
APIError: 500 Server Error: Internal Server Error ("Cannot start container 24396d264e8472ae0dc93c04ebd7cbd4b35f9b8f444d74701d720dc922433a34: Error starting userland proxy: listen tcp 0.0.0.0:9999: bind: address already in use")
-------------------- >> begin captured logging << --------------------
requests.packages.urllib3.connectionpool: DEBUG: "GET /run/docker.sock/v1.14/containers/json?all=1&limit=-1&trunc_cmd=1&size=0 HTTP/1.1" 200 None
[...]

PR #1118, opened 2015-03-16
Change port in ports-composefile to 49152

funkyfuture:
This shall lower the propability to interfere with another service (e.g. the
WebUI of an application) that is running on the machine where tests are run.

Signed-off-by: funkyfuture <funkyfuture@riseup.net>

Solves #1082


bfirsh:
LGTM

dnephin:
LGTM

Issue #1066, opened 2015-03-05
Specify project and service with container labels


bfirsh:
Labels are likely going to land in Docker 1.6: https://github.com/docker/docker/pull/9882

If we used these to specify what project and service a container is part of, a whole load of stuff becomes a lot simpler. We don't have to do gross things like `project, service, n = container_name.split("_", 3)`.

Questions:

 - Do we still want to give containers names? It's useful to be able to reference containers like that (`docker exec project_web_1`) and we want something useful to show up in `docker ps`
 - Do we still want to give containers integer references of some kind? It's pretty handy to be able to have a human readable reference to a container

thaJeztah:
:+1: :)

> Do we still want to give containers names?

Why not? I'd say, keep the convention to name containers that are managed by `compose`. However, there are various feature-requests to be able to override the name, e.g. by allowing a `name: com.my-container.app` in `docker-compose.yml`

> Do we still want to give containers integer references of some kind

I think that's still needed if scaling is possible. ie, you can use a (custom)name for a container, but a second instance should still have a unique name (suffix). This should also be stored in a label, though, to allow filtering.

dnephin:
>  I'd say, keep the convention to name containers that are managed by compose. However, there are various feature-requests to be able to override the name,

+1. The names are still nice to have, but being able to customize them is good too

montells:
+1 Names are usefull to me.

zkanda:
Please keep the names... :+1:

thaJeztah:
Happy to inform that support for labels (https://github.com/docker/docker/pull/9882) was merged and will be included in the upcoming 1.6 release of Docker.

Also, I think there are currently a number of related issues in this area (https://github.com/docker/compose/issues/652, https://github.com/docker/compose/issues/869, https://github.com/docker/compose/issues/941 among others); perhaps the discussion should be focussed in *this* issue (but that's just a suggestion) :)

bfirsh:
Ahhwwww yeah.

aanm:
@bfirsh  I want to start working around this, can I start PR stuff?

dnephin:
I'm going to take a first pass at this to use labels instead of the naming convention to identify containers.

aanm:
@dnephin 
Take a look at https://github.com/docker/compose/pull/1269

dnephin:
[...]

Issue #846, opened 2015-01-15
Next-generation configuration format


aanand:
There are a whole load of things that people are asking for in `fig.yml`. I want to pull them together here so that we can discuss configuration at a high level and avoid incremental bloat and viscosity.

(**Important note:** I haven’t the least bit of interest in discussing filenames or alternatives to YAML in this issue. A separate issue, sure, but don’t shed those bikes in here.)

I also want to discuss what I see as an emerging boundary between *portable* and *non-portable* configuration. By way of analogy, let's look at the Dockerfile format, which aims to be fully portable. For example:

- `VOLUME` lets you specify paths inside the container that should be volumes, but doesn't let you specify a host path.
- `EXPOSE` lets you specify a port to expose, but not what port on the host to map it to.

As a result, a Docker image will work anywhere - it isn't coupled to any features of the host environment, such as filesystem layout or available ports. This enables people to build tools on Docker which can run any image without worrying about how it's configured, an abstraction which is only possible because Dockerfile *enforces* separation.

With Fig we've created a configuration format for a group of containers - let's call it an app. Most of what you specify in `fig.yml` (i.e. most container configuration plus links between defined containers) is, to my eyes, portable - but not all of it. We let you specify:

- volume paths on the host
- ports to expose on the host
- links to externally-managed containers (#544, merged but not released)
- `build` keys which reference paths outside the current directory

There have also been requests/PRs for support for:

- service scaling directives (#630)
- order of container startup (#235)
- parameterisation of configuration values, probably based on environment variables local to the shell in which Fig is being invoked (#426, #495, #845)
- spinning up other Fig apps which an app depends on, identified by the path to their code/configuration (#318, #758)

All of this looks highly *non-portable* to me - if we continue to support such features in `fig.yml`, it’ll never be suitable as an abstract app definition, and apps will remain coupled to particular characteristics of the systems they’re running on. As we move towards a future where deploying Docker apps on multiple hosts (e.g. with Swarm) or to multiple environments (dev, CI, staging, production), this will become more and more of a pain point.

So I want to talk about how we might redesign Fig’s configuration such that we can support the use cases those features would serve - all of them real problems that real users have faced - and simultaneously achieve an app definition format that is as portable as a Dockerfile.

If we want to do this without sacrificing a significant amount of usability, one approach I’ve thought of is to define two formats:

- A core app definition file, analogous to Dockerfile, which only allows portable configuration.
[...]

PR #1124, opened 2015-03-17
Implemented labels on docker compose

aanm:
yml example file
```yml
web:
  labels:
   - ram=2g
  build: .
  command: python app.py
  ports:
   - "5000:5000"
  volumes:
   - .:/code
  links:
   - redis
redis:
  labels:
   - storage=ssd
   - ram=4g
   - production
  image: redis
```

Do not merge yet.
Depends on https://github.com/docker/docker-py/pull/529
Signed-off-by: André Martins <martins@noironetworks.com>

### DIFF
--- a/compose/service.py
+++ b/compose/service.py
@@ -9,7 +9,7 @@ import sys
 
 from docker.errors import APIError
 
-from .config import DOCKER_CONFIG_KEYS
+from .config import DOCKER_CONFIG_KEYS, load
 from .container import Container, get_container_name
 from .progress_stream import stream_output, StreamOutputError
 
@@ -29,6 +29,7 @@ DOCKER_START_KEYS = [
 
 VALID_NAME_CHARS = '[a-zA-Z0-9]'
 
+LABEL_PREFIX = ''
 
 class BuildError(Exception):
     def __init__(self, service, reason):
@@ -48,6 +49,9 @@ VolumeSpec = namedtuple('VolumeSpec', 'external internal mode')
 
 
 ServiceName = namedtuple('ServiceName', 'project service number')
+
+
+LabelSpec = namedtuple('LabelSpec', 'key value')
 
 
 class Service(object):
@@ -181,7 +185,6 @@ class Service(object):
                 self.can_be_built() and
                 not self.client.images(name=self.full_name)):
             self.build()
-
         try:
             return Container.create(self.client, **container_options)
         except APIError as e:
@@ -296,7 +299,8 @@ class Service(object):
             self,
             insecure_registry=False,
             detach=False,
-            do_build=True):
+            do_build=True,
+            **override_options):
         containers = self.containers(stopped=True)
 
         if not containers:
@@ -305,6 +309,7 @@ class Service(object):
                 insecure_registry=insecure_registry,
                 detach=detach,
                 do_build=do_build,
+                **override_options
             )
             return [self.start_container(new_container)]
         else:
@@ -398,6 +403,51 @@ class Service(object):
         container_options['name'] = self._next_container_name(
             self.containers(stopped=True, one_off=one_off),
             one_off)
+
+        if 'labels' in container_options:
+            container_options['labels'] = dict(
+                (LABEL_PREFIX + key, container_options['labels'][key])
+                for key in container_options['labels'])
+
+        if 'remove_labels' in container_options:
+            service_cfg_lbls = container_options.get('labels')
+            if service_cfg_lbls is not None:
+                regex = container_options.get('remove_labels')
+                a = re.compile(regex)
+                container_options['labels'] = dict(
+                    (k, v) for (k,v) in service_cfg_lbls.iteritems()
+                    if not a.match(k))
+            del container_options['remove_labels']
+
+        if 'labels_file' in container_options:
+            # Merge labels from config with labels from --labels-file
+            label_cfg_file = container_options.get('labels_file')
+            label_list_cfg_file = load(label_cfg_file)
+            service_cmd_opts = [a_serv_lbl for a_serv_lbl in label_list_cfg_file
+                                if a_serv_lbl.get('name') == self.name
+                                and a_serv_lbl.get('labels') is not None]
+            if len(service_cmd_opts) == 0:
+                raise ConfigError ('No labels defined for service \'' +
+                                   self.name + '\' in ' +
+                                   container_options.get('labels_file'))
+            service_cmd_lbls = service_cmd_opts[0]['labels']
+            service_cfg_lbls = container_options.get('labels')
+            prefix_cmd = ""
+            prefix_cfg = ""
+            if 'prefix_labels' in container_options:
+                prefix_cmd = ".cmd:"
+                prefix_cfg = ".cfg:"
+                del container_options['prefix_labels']
+            temp_labels = dict()
+            if service_cfg_lbls is not None:
+                temp_labels.update(
+                    (prefix_cfg + key, service_cfg_lbls[key]) for key in service_cfg_lbls)
+            temp_labels.update(
+                (prefix_cmd + key, service_cmd_lbls[key]) for key in service_cmd_lbls)
+
+            container_options['labels'] = temp_labels
+            del container_options['labels_file']
+
 
         # If a qualified hostname was given, split it into an
         # unqualified hostname and a domainname unless domainname
@@ -552,7 +602,6 @@ def parse_volume_spec(volume_config):
                           "one of: rw, ro." % (volume_config, mode))
 
     return VolumeSpec(external, internal, mode)
-
 
 def parse_repository_tag(s):
     if ":" not in s:
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -13,7 +13,7 @@ import dockerpty
 from .. import __version__
 from ..project import NoSuchService, ConfigurationError
 from ..service import BuildError, CannotBeScaledError
-from ..config import parse_environment
+from ..config import parse_environment, load
 from .command import Command
 from .docopt_command import NoSuchCommand
 from .errors import UserError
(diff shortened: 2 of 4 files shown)


thaJeztah:
Thanks! Wondering; would it make sense to use a dict/hash-map (whatchamacallit in yaml) in stead of an array? I.e;

```yaml
web:
    labels:
        ram: 2g
        storage: ssd
```

aanm:
@thaJeztah
While coding I wondered that too.
Also, what about implementing labels via command line for example
```
docker-compose up web --label ram=8g
```
That would overwrite the settings in the yaml file, described here https://github.com/docker/docker/issues/11187 (An end-to-end example)

thaJeztah:
I like that. Wrt hash/array; I'll leave that to the maintainers for now. both will work, but a hash more closely matches the JSON in the Docker API

Also, the docker labels feature supports a `--label-file`. I wonder how that should be implemented in compose (1 file per service?)

dnephin:
It was mentioned a few times in the labels PR that labels pretty closely resembles environment variables. I suspect that will be true for our implementation here as well.

For environment variables both lists and dicts are supported in the config, `env_file: <path>` to point at a file, and `-e` as a param to `run`.  

I could see labels working the same way: `label_file: ...` in the config, `--label` as a param to `docker-compose run` (not `docker-compose up`)

compose/service.py
```diff
@@ -399,6 +401,12 @@ def _get_container_create_options(self, override_options, one_off=False):
             self.containers(stopped=True, one_off=one_off),
             one_off)
 
+        if 'labels' in container_options:
+            labels_dict = ['io.docker.compose:' + label for label in container_options['labels']]
```
dnephin: We should not  be namespacing labels here, it should be left up to the developer.  

We will likely use a namespace for our internal labels in #1066, but any user specified labels should be unmodified I think.

thaJeztah: Agreed; labels set by the user should be sent to docker as-is.

thaJeztah:
> For environment variables both lists and dicts are supported in the config

True. The reason I thought a dict would be more appropriate because (contrary to env-variables), the storage for "labels" in Docker also is a dict. Label *keys* (names) are unique and a dict more verbosely expresses that.

An array should work, but would (probably) require more handling, because latter values for the same label should overwrite former ones.

> I could see labels working the same way: label_file: ...

Funny fact; Docker actually supports *multiple* env-files. The contents of those files are combined / latter values overwrite former values here as well.

I was wondering if it should be possible to specify a `--label-file` on the command line. Problem is that it's not possible to specify *which service* the file should be applied to (the same obviously is true for `--label`); will it be applied to *all* services? Is that the desired result? (not sure)

compose/service.py
tgraf: A motivation for a separate prefix for labels specified in the composition definition I can see is to be able to differ between the labels specified in the composition and the labels specified with `docker-compose run --label`.

This implements separation of concern between labels specified by the developer and whoever is running the composition in a particular environment. Otherwise `docker-compose run` invocations are required to always specify and overwrite all possible labels used in an environment if the guy running the workloads wants to avoid a developer mistakenly specifying a label used for operational concerns.

Simple example:
You are using a label `profile=production` which influences resource allocation and scheduling. You do not want compositions to inherit these privileges unless the label was specified by the operator invoking `docker-compose run`.

A richer write-up of this separation of concern can be found here:
https://github.com/docker/docker/issues/11187

thaJeztah: Ideally, labels used for scheduling (Swarm) should be namespaced and perhaps not even defined as *labels* in `docker-compose.yml` (even though they will be set as labels), e.g.

```yaml
web:
    constraints:
        storage: ssd
        memory: 2g
    labels:
        hello: world
```

the "constraints" will then result in a `--label com.docker.swarm.constraint=storage=ssd` (or similar)

I think Swarm is planning on namespacing their labels as well, but I'll have to check.

aanm:
Thanks for the feedback guys.
@thaJeztah 
Changing to a dict simplified the code, thanks.
Regarding to specifying labels via command line I was thinking something like:
``` 
run [options] [--label serv:KEY[=VAL]...] [-e KEY=VAL...] SERVICE [COMMAND] [ARGS...]
   --label serv:KEY[=VAL]   Specifies a label for a given SERVICE. If
                                        no SERVICE is given the label will be used
                                        on every service.
   --no-prefix-label     Don't prefix 'cfg' and 'cmd' when --label is used.
   --label-file               Configuration yml file with all the labels.
```
The problem I have is, for example, with this command:
```
docker-compose run --label web:myEnv=draft redis:myEnv=complete
```
Now, is the `redis:myEnv=complete` a SERVICE or an argument from label?
Accordingly with the docs a service name is only composed by [a-zA-Z0-9]. So, I'll assume that everything before the `:` is the SERVICE. Thus, it will not be possible to give a single label via command line for every service started.

I can also create an option to read from a .yml that has the labels predefined by the operator.
labels.yml
```yaml
web:
    constraints:
        storage: ssd
        memory: 2g
    labels:
        hello: world
```
That would result in something like:
```json
        "Labels": {
            "io.docker.compose.cmdline:memory": "2g" ->  The one inside labels.yml from the OP
            "io.docker.compose.cfgfile:memory": "99g" -> The one inside compose.yml from the dev
        },
```
What are your thoughts on this?
@dnephin 
You said earlier "`--label` as a param to docker-compose run (not docker-compose up)" Why can't be both?

compose/service.py
tgraf: Agreed that the constraints even though stored as labels should not all share the same namespace. It must be known what is a constraint, what is an identification label, what is a capability label, etc. 

An operator should be able to say at `docker-compose run` time to ignore all previous constraints and use a new set. This should not require the operator to overwrite each constraint separately. Allowing for this separation of concern is absolutely critical. It must be possible to tie applications to infrastructure without storing this glue in the application itself.

So instead of separating compose and runtime labels with a namespace, an alternative could be a `--remove-labels` option which allows for wildcards:

```yaml
web:
    labels:
         constraints:
            storage: ssd
            memory: 2
         generic:
            hello: world
```

To overwrite all constraints:

`[...] --remove-labels 'constraints.*' --label constraints.memory=4g`

tgraf:
@aanm 
The `--label` option could simply support a single path argument to read labels from a `.yml` instead. This does not fully resolve the developer vs. operator labels though without a way to remove or replace whole label namespaces. An additional `--remove-labels` might be an easy to understand API for the operator:

* No option provided
  * Labels are inherited 1:1, no special operator concerns.
* `--remove-labels 'constraints.*'`
  * Remove all constraints but keep other labels, operator only wants to provide own constraints
* `--remove-labels`
  * Remove all labels, I'm not trusting the labels specified at all.

I'm sure @aanand has his own thoughts on this ;-)

aanm:
```bash
$ sudo docker-compose help up
    ...
    --labels-file FILE          Uses all labels, in the given yaml FILE for the
                                SERVICEs to be up.
    --no-prefix-labels          Don't prefix '.cfg' and '.cmd' when --labels-file
                                is used.
    --remove-labels [REGEX]     Removes all labels. If REGEX is given, removes
                                labels that match the given REGular EXpression.
                                (Note: This option has less priority than the
                                given --labels-file)
    ...
$ sudo docker-compose help run
    ...
    --labels-file FILE          Uses all labels, in the given yaml FILE for the
                                SERVICE to be run.
    --no-prefix-labels          Don't prefix '.cfg' and '.cmd' when --labels-file
                                is used.
    --remove-labels [REGEX]     Removes all labels. If REGEX is given, removes
                                labels that match the given REGular EXpression.
                                (Note: This option has less priority than the
                                given --labels-file)
    ...
$ cat docker-compose.yml
web:
  labels:
    dev.attrib.color: white
    dev.attrib.shape: square
  build: .
...
redis:
  labels:
    storage: ssd
    ram: 4g
    profile: production
  image: redis

$ cat labels.yml 
web:
  labels:
    color: green
redis:
  labels:
    storage: ssd
    ram: 2g
    profile: production
$ sudo docker-compose up --labels-file labels.yml --remove-labels 'dev.*'
$ sudo docker inspect dockercompose_web_1 | grep Labels -B1 -A6
        "Image": "dockercompose_web",
        "Labels": {
            ".cmd:color": "green"
        },
        "MacAddress": "",
        "Memory": 0,
        "MemorySwap": 0,
        "NetworkDisabled": false,
$ sudo docker inspect dockercompose_redis_1 | grep Labels -B1 -A8
        "Image": "redis:latest",
        "Labels": {
            ".cfg:profile": "production",
            ".cfg:ram": "4g",
            ".cfg:storage": "ssd",
            ".cmd:profile": "production",
            ".cmd:ram": "2g",
            ".cmd:storage": "ssd"
        },
        "MacAddress": "",
```
Note the absence of `dev.attrib` in `dockercompose_web_1` ;-)
Feel free to add more suggestions.

thaJeztah:
Thanks @aanm! Hm, my thoughts;

> `docker-compose run --label web:myEnv=draft redis:myEnv=complete`
> Now, is the redis:myEnv=complete a SERVICE or an argument from label

Not sure if you meant write that, but the example should be `--label web:foo.... --label redis:foo....` (i.e. the `--label` flag must be repeated for each label.

I like the idea, though, of using a `service:` prefix.


Some other thoughts;

- I don't think the labels should be automatically prefixed/namespaced (with `.cfg` / `.cmd`); Docker doesn't do this, so Compose shouldn't do this either. Adding `.cfg` and `.cmd` namespaces is (IMO) overcomplicating things; labels should follow the same order of preference as Docker does (e.g. CLI > YAML) 
- The `--label-file` should accept a newline-delimited file. I think a YAML format will also be covered by 
the new config format (https://github.com/docker/compose/issues/846)
- ***If*** a shorthand is implemented, I'd prefer `--label-ns=my.name.space` over `--no-prefix-label`

Perhaps more importantly; I'm fine with adding more features later, but please in a follow-up PR. Keep the initial implementation simple and **only** implement what's supported by Docker itself, i.e.;

- Being able to specify labels in `docker-compose.yml` (no automatic namespaces/prefixes)
- Being able to specify labels via the CLI (`--label`) (initially, applied to *all* services)
- Being able to read in a `--label-file` (initially, applied to *all* services)

Those should not require a lot of discussion, which makes it easier to merge.

In a follow-up/separate PR, `--label [service:]label[=value]` could be implemented. This may require some discussion, wrt compatibility

All features wrt https://github.com/docker/docker/issues/11187 etc., should be in a follow up. Currently, https://github.com/docker/docker/issues/11187 is just a *proposal* and it doesn't make sense to implement anything before it's even accepted in Docker itself. It's fine to *discuss* options for implementing it in Compose, but best kept in a separate issue.

However, I'm not a maintainer, just my personal opinion!

aanm:
> I don't think the labels should be automatically prefixed/namespaced (with .cfg / .cmd); Docker doesn't do this, so Compose shouldn't do this either. Adding .cfg and .cmd namespaces is (IMO) overcomplicating things; labels should follow the same order of preference as Docker does (e.g. CLI > YAML)

Yeah, I agree, it doesn't look "good" with .cfg/.cmd. Since docker uses CLI > YAML I'll do it the same way.

> The --label-file should accept a newline-delimited file. I think a YAML format will also be covered by the new config format (#846)

What's a "newline-delimited file"? This -> https://en.wikipedia.org/wiki/Line_Delimited_JSON#Example_Output

> If a shorthand is implemented, I'd prefer --label-ns=my.name.space over --no-prefix-label

`my.name.space` would be applied to **all** labels?

> Perhaps more importantly; I'm fine with adding more features later, but please in a follow-up PR. Keep the initial implementation simple and only implement what's supported by Docker itself, i.e.;

Okay ;-) I will:
* remove the .cfg/.cmd thing
* specify labels via the CLI (--label) (initially, applied to all services)
* make tests for everything
* create a new PR for the '--remove-labels'? (not sure if I should keep this here or move it to a new PR)

Thanks for the feedback!

thaJeztah:
> What's a "newline-delimited file"?

The label file uses the same format as `--env-file`. You can find an example in this section; https://docs.docker.com/reference/commandline/cli/#examples_8

Basically, it's just a single `key=value` per line.

> create a new PR for the '--remove-labels'?

I would personally move it to a new PR, yes, so that it can be discussed without upholding this PR.

But again, I'm not a maintainer; they could have a different opinion here :)

gourao:
Who consumes (or acts) upon these labels?  From what I understand so far, from this PR (and also from https://github.com/docker/docker/pull/9882), I see a way for opaque labels to be passed from the compose yaml spec or from the dockerfile all the way to the docker daemon.  But since these are opaque, docker itself is not acting on the labels other than preserving them.

So my question is, who are these labels intended for and how do we go about acting on them?  

My interest for example would be around getting the storage specific labels and making the storage implementation honor the requested labels.

thaJeztah:
At this moment, labels are only that; a label. No software "acts" on them. Docker enables filtering images/containers based on labels and you are able to read the labels using `docker inspect`.

It *is* possible to have software make use of those labels (e.g. Composer itself, to store the project name, or Swarm to schedule containers), but for now that should be "out of scope" for this PR.

aanand:
Thanks for making a start on this. I think this PR does much more than the minimum necessary for a useful labels feature:

- As @thaJeztah said, I don't think we ought to be pre/postfixing label names.
- I think `labels_file` can be implemented separately.
- Any command-line flags are a nicety - `labels` in `docker-compose.yml` is the real value-add.
- `remove_labels` is a confusing feature. I'm not convinced it's useful.

Furthermore, it puts a lot of logic in `main.py` and `service.py` which should really be in `config.py`.

I've made a start on an MVP labels feature in #1139. I think the next good thing to implement would be `labels_file`, if you want to take a crack at that - have a look at the implementation of `env_file`.

PR #1138, opened 2015-03-19
Resolve links set as DNS to their ip address.

rubenbe:
When a dns section contains names of linked services,
the names are replaced with the container addresses.

Example:

```yaml
mydns:
    image: ubuntu:14.04
    command: sleep 300 
web:
    image: ubuntu:14.04
    command: cat /etc/resolv.conf
    links:
      - mydns:dns
    dns:
       - dns
       - 8.8.8.8
```

This will configure the following resolv.conf in the "web" container:

```
nameserver 172.17.10.113
nameserver 8.8.8.8
```

### DIFF
--- a/compose/service.py
+++ b/compose/service.py
@@ -270,7 +270,7 @@ class Service(object):
             if ':' in volume)
 
         privileged = options.get('privileged', False)
-        dns = options.get('dns', None)
+        dns = options.get('dns', [])
         dns_search = options.get('dns_search', None)
         cap_add = options.get('cap_add', None)
         cap_drop = options.get('cap_drop', None)
@@ -284,7 +284,7 @@ class Service(object):
             volumes_from=self._get_volumes_from(intermediate_container),
             privileged=privileged,
             network_mode=self._get_net(),
-            dns=dns,
+            dns=self._get_dns_resolved(dns),
             dns_search=dns_search,
             restart_policy=restart,
             cap_add=cap_add,
@@ -331,6 +331,18 @@ class Service(object):
     def _next_container_number(self, all_containers):
         numbers = [parse_name(c.name).number for c in all_containers]
         return 1 if not numbers else max(numbers) + 1
+
+    def _get_dns_resolved(self, dns):
+        if type(dns) != list:
+            dns = [dns]
+        containers = dict([(link_name, service.containers()) for service, link_name in self.links])
+        dns_resolved = []
+        for d in dns:
+            if d in containers.keys():
+                dns_resolved.extend([c.get('NetworkSettings.IPAddress') for c in containers[d]])
+            else:
+                dns_resolved.append(d)
+        return dns_resolved
 
     def _get_links(self, link_to_self):
         links = []
--- a/docs/yml.md
+++ b/docs/yml.md
@@ -189,7 +189,7 @@ net: "host"
 
 ### dns
 
-Custom DNS servers. Can be a single value or a list.
+Custom DNS servers. Can be a single value or a list. Names of linked containers are resolved.
 
 ```
 dns: 8.8.8.8
--- a/tests/integration/service_test.py
+++ b/tests/integration/service_test.py
@@ -434,6 +434,28 @@ class ServiceTest(DockerClientTestCase):
         container = create_and_start_container(service)
         self.assertEqual(container.get('HostConfig.Dns'), ['8.8.8.8', '9.9.9.9'])
 
+    def test_dns_named_single_value(self):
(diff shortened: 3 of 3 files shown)


dnephin:
Interesting idea, I can see this being quite useful, especially if you want to integrate with one of the many dynamic-dns-for-containers-as-a-container options.

It might be worth it to make the syntax `container:dns` to be a bit more explicit (I believe this is the syntax for `net`).

@aanand @bfirsh what do you think?

docs/yml.md
```diff
@@ -189,7 +189,7 @@ net: "host"
 
 ### dns
 
-Custom DNS servers. Can be a single value or a list.
+Custom DNS servers. Can be a single value or a list. Names of linked containers are resolved.
```
dnephin: This could use a bit more documentation, maybe an example below?

compose/service.py
```diff
@@ -332,6 +332,18 @@ def _next_container_number(self, all_containers):
...
+        if type(dns) != list:
+            dns = [dns]
+        containers = dict([(link_name, service.containers()) for service, link_name in self.links])
+        dns_resolved = []
+        for d in dns:
+            if d in containers.keys():
+                dns_resolved.extend([c.get('NetworkSettings.IPAddress') for c in containers[d]])
+            else:
+                dns_resolved.append(d)
+        return dns_resolved
```
dnephin: It seems like this doesn't need to be restricted to just linked containers. It should be able to use *any* container with the appropriate name (maybe a container that isn't even part of the composition).

If this were to use `container:<name>` syntax as I've suggested, you should be able to extract a lot of shared code from [project.get_net()](https://github.com/docker/compose/blob/9931cd2c4cd6683cef5ded0d8f15c71c7a9f7741/compose/project.py#L161).

The "lookup container first by service name, then by container name" would be identical (which is great for consistency).

thaJeztah:
Won't this be a problem if the linked service is restarted and changes its IP address?

Also (not sure if that would help here), I think some changes were made recently to dynamically update resolv.conf in some cases.

dnephin:
Ya, I don't think this should use linked services. It should work with a container outside of the composition.

thaJeztah:
Well, then fun thing of course is that a linked container will be in `/etc/hosts` and automatically updated on changes; in which case using just "dns" for the DNS (in stead of the IP-address) would make sense.

rubenbe:
@thaJeztah I've tried adding "dns" directly in /etc/resolv.conf, but that didn't work.
This is also according to the resolv.conf(5) man page (http://linux.die.net/man/5/resolv.conf) which states that nameservers should be defined by their ip address.

qzio:
I'm :+1:  on this, or at least something like this.

I want to run a dns resolver container that can be used by all the other containers, to be able to take advantage of the dns part of consul.

aanand:
Interesting idea. @dnephin's `container:<name>` syntax makes sense.

Two things:

- It seems like this change should be made in Docker, i.e. `docker run --dns=container:<name>`
- How hard is it to use a linked container as a DNS server *without* this change?

lalyos:
The usual workaround for using a dns runing is a different containers is to let the dns container:
- expose the port 53:53 and bind to 0.0.0.0 insisde
- expose the port 53 explicitly to `172.17.42.1:53/udp`

This way other containers can refer to it as `--dns 172.17.42.1` or whatever your bridge_ip is. As an alternative you can use --net=host so can bind to the host ip.

As a consul specific description see the README at https://github.com/progrium/docker-consul

sttts:
:+1:  this feature would be great, in my case to model distributed systems with circular dependencies. Manual links don't work in that case.

steve-jansen:
@lalyos Thank you for highlighting the `172.17.42.1:53:53/udp` workaround!  Helped me solve a problem with binding the network to the host on CircleCI!  :beers:

thaJeztah:
@steve-jansen @lalyos please note that with docker 1.9, the IP-address is not guaranteed to be 172.17.42.1; https://github.com/docker/docker/issues/17305

dnephin:
Thanks again for the PR.

Since this PR was first opened a lot has changed. We now have user defined networks in Docker, and in the next release they should be using an embedded DNS server to handle "discovery". 

There should be much less of a need for these dns discovery services now that it's built into Docker.

I'm going to close this PR since this functionality should be handled in the next release. If you're interested in following the proposals you can find more information here:
* https://github.com/docker/libnetwork/issues/737
* https://github.com/docker/libnetwork/issues/767
* https://github.com/docker/docker/issues/18699

PR #1139, opened 2015-03-19
WIP: `labels` option in docker-compose.yml

aanand:
```yaml
labels:
  com.example.description: "Accounting webapp"
  com.example.department: "Finance"
```

```yaml
labels:
  - "com.example.description=Accounting webapp"
  - "com.example.department=Finance"
```

This is WIP for two reasons:

1. It tests against a development version of Docker which is just bundled in the repo. Not nice.
2. It uses an unreleased docker-py implementation of labels (https://github.com/docker/docker-py/pull/532).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,13 +15,18 @@ RUN set -ex; \
     ; \
     rm -rf /var/lib/apt/lists/*
 
-ENV ALL_DOCKER_VERSIONS 1.3.3 1.4.1 1.5.0
+# ENV ALL_DOCKER_VERSIONS 1.6.0
 
-RUN set -ex; \
-    for v in ${ALL_DOCKER_VERSIONS}; do \
-        curl https://get.docker.com/builds/Linux/x86_64/docker-$v -o /usr/local/bin/docker-$v; \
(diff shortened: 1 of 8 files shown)


thaJeztah:
I'm still in preference of the first notation (a dict). I realise an array more closely matches the way ENV-vars are currently implemented, but a dict makes it more verbose that label names (keys) are unique. In addition, I can imagine a dict will be easier to handle, e.g., when merging labels from various sources (yaml, CLI, label-file), but that's a lesser concern.

aanand:
@thaJeztah I also prefer the dict form, and I feel the same way about environment variables too - but for consistency between the two features, and between Compose and the Docker CLI, I think it's important to support both.

Both env vars and labels are normalised to a dictionary when we read them, so it's purely cosmetic.

thaJeztah:
@aanand no problem! Just emphasising the "dict" approach in the documentation should do :)

dnephin:
I like it

aanand:
I've force-unmerged this, since the docker binary is kind of big and we don't want to make everyone who clones docker/compose download it. Will re-merge it once #1144 is addressed.

PR #1142, opened 2015-03-20
Added support of option mac-address

calou:


JonasT:
Will this be fixed & merged soon? It seems to be an essential option for an IPv6 workaround due to the mess that is IPv6 support in docker right now (see https://github.com/docker/docker/issues/13481 for details), so it would be really very helpful if this was added.

Also, it appears the merge conflict seems to be just the docs part which shouldn't be hard to fix, so what's the hold up?

dnephin:
I think the tests were failing originally because we weren't running against a new-enough version of docker. That should be resolved now.

With a rebase this should be ready to go.

calou:
I just rebased but the tests are not still working.
I got the following error: Unable to find image 'docker-compose:963f0fc' locally
I don't know for sure what the problem is or what I did wrong.

compose/config.py
```diff
@@ -37,6 +37,7 @@
     'volumes',
     'volumes_from',
     'working_dir',
+    'mac_address',
```
dnephin: Please add this entry to the list in sorted order

calou: Done

dnephin:
It would be good to squash these into a single commit, otherwise looks good!

calou:
I squashed the commits, seems like the test results are good.

dnephin:
LGTM

aanand:
LGTM

calou:
Thanks

PR #1146, opened 2015-03-21
Label file support

aanm:
Implemented label-file support for docker-compose.yml

I think that everything is working, unless some tests fail.
@aanand what do you think?

### DIFF
--- a/compose/config/config.py
+++ b/compose/config/config.py
@@ -62,6 +62,7 @@ DOCKER_CONFIG_KEYS = [
     'image',
     'ipc',
     'labels',
+    'label_file',
     'links',
     'mac_address',
     'mem_limit',
@@ -511,6 +512,18 @@ def resolve_build_args(build):
     return dict(filter(None, (resolve_env_var(k, v) for k, v in six.iteritems(args))))
 
 
+def resolve_labels(service_dict):
+    """Unpack any labels variables from an label_file, if set.
+    Interpolate labels values if set.
+    """
+    labels = {}
+    for label_file in service_dict.get('label_file', []):
+        labels.update(labels_from_file(label_file))
+
+    labels.update(parse_labels(service_dict.get('label_file')))
+    return dict(resolve_labels(k, v) for k, v in six.iteritems(labels))
+
+
 def validate_extended_service_dict(service_dict, filename, service):
     error_prefix = "Cannot extend service '%s' in %s:" % (service, filename)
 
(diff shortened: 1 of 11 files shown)


aanm:
The `extends` option deprecate this PR

aanand:
I wouldn't say it deprecates it - `labels_file` would still be useful for setting the same labels across multiple heterogeneous services in a file - but it's true that it adds slightly less value now.

thaJeztah:
Also, labels-file is something that docker itself supports so might still be desirable to bring parity with docker.

aanand:
Yep - if people are using `docker run --labels-file` and want to switch to Compose, they're gonna want a replacement for it.

aanm:
Right now the option is not available on CLI, do you think it should be?

aanand:
Thanks for working on this. No, it's not imperative that we support it in the CLI.

The labels file support as you've implemented it differs significantly from Docker's `--label-file` support. I see the value in it, but it's more than we need for feature parity, and as far as I can tell the file format isn't backwards-compatible with the Docker CLI. It'd be better for this PR to just support Docker's simple label file format.

aanm:
@aanand I see that `env_file` doesn't seem to have the same behavior of docker's `env-file`. It seems like `compose` deals with all env by himself without giving `docker daemon` a chance of know that `env-file` was used on `compose`. Am I wrong?

thaJeztah:
@aanm the point is, that users should be able to create an env-file, use it with docker, and when creating their "composition", use *the same file* in docker-compose. For that to work, docker-compose and docker should use the same file-format.

aanm:
@thaJeztah ah! I was seeing things wrongly. I'm going to make necessary changes then.

aanm:
@aanand Will you merge this in the next days? Please tell me so I will solve the conflicts.

compose/config.py
```diff
@@ -344,6 +346,63 @@ def env_vars_from_file(filename):
...
+    if not os.path.exists(filename):
+        raise ConfigurationError("Couldn't find label file: %s" % filename)
+
+    labels = {}
+    for line in open(filename, 'r'):
+        line = line.strip()
+        if line and not line.startswith('#'):
+            k, v = split_label(line)
+            labels[k] = v
+    return labels
```
dnephin: A lot of this code is almost identical to  the `env_file` equivalent. We should just add a parameter to those functions and re-use the same code. I think that will make it easier to rebase.

aanand:
Sorry for the delay - if you rebase this and make the refactoring @dnephin, we'll merge it.

aanm:
@aanand
I've changed the code to be updated with the current code.
The tests errors are caused by this pull request?

aanand:
Yep, looks like you've got some formatting issues:

```
compose/config/config.py:342:1: E302 expected 2 blank lines, found 1
compose/config/config.py:346:11: E225 missing whitespace around operator
compose/config/config.py:350:5: F821 undefined name 'env'
compose/config/config.py:351:64: F821 undefined name 'env'
compose/config/config.py:353:1: E302 expected 2 blank lines, found 1
compose/config/config.py:543:1: E302 expected 2 blank lines, found 1
compose/config/config.py:677:1: F811 redefinition of unused 'split_label' from line 543
tests/integration/service_test.py:862:5: E303 too many blank lines (2)
```

PR #1157, opened 2015-03-23
Support an api version from the environment

dnephin:
Right now I can't even run the test suite because it expects a version of the API that hasn't been officially released. 

By allowing the version to be specified from the environment developers can at least the older features in compose without upgrading the dockerd.


dnephin:
Replaced by #1780

PR #958, opened 2015-02-12
Adding extra_hosts param to docker-compose

birdy-:
Adding extra_hosts parameter with doc and tests

PR #848, opened 2015-01-15
Add support for docker's client's --add-host command

sampwing:
Updated docker-py to the most recent version ( 0.7.1), which allows assigning values to extra_hosts (similar to ---add-host.)  Updated fig to allow passing in these parameters and added tests.

Implementation requested in the following issues:
#597 
#754

Issue #754, opened 2014-12-23
docker-compose.yml should support every docker run option


bfirsh:
Notable things missing:

 - [x] add-host https://github.com/docker/fig/issues/597 https://github.com/docker/compose/pull/1158
 - [x] cpuset https://github.com/docker/compose/pull/685
 - [x] device https://github.com/docker/compose/commit/df87bd91
 - [x] memswap_limit https://github.com/docker/fig/pull/685
 - [x] mac_address added in this commit: https://github.com/docker/compose/commit/b76ac6e6

The following run options aren't in docker-py; cpuset-mems, cpu-period, cpu-quota, oom-kill-disable, sig-proxy

Related: #363 

Examples: https://github.com/docker/compose/pull/293 https://github.com/docker/compose/pull/830

mishak87:
Any progress with add-host https://github.com/docker/fig/issues/597? In docker-py support is already present https://github.com/docker/docker-py/pull/400.

mishak87:
I would be for:
```yaml
hosts:
   hostname: ip
```

dommmel:
AFAICT support for "device" is also present in docker-py (see https://github.com/docker/docker-py/pull/373)

gpolyn:
+1 for #597

zhucegep:
I really need this feature: #597

mishak87:
@zhucegep you can try [crane](https://github.com/michaelsauter/crane) which supports it

razic:
@bfirsh are `memswap_limit` and `cpuset` not integrated into `docker-compose` yet?

razic:
also i'm trying to use `cpu_shares` which looks to be defined here, but doesn't seem to actually be working... https://github.com/docker/fig/blob/master/compose/service.py#L21

zhucegep:
@mishak87  thank you, I had implement this feature on environment variable.

usage:
```
environment:
    HOSTS:  "zookeeper zookeeper.example.com\n172.17.42.1 my.own.com"
```

impl:

```
#! /bin/bash
echo  "# host config by docker" >> /etc/hosts
echo  "$HOSTS" | while read line 
do
  link_service_name=`echo $line | cut -d" " -f1`
  ip=$link_service_name
  # if does not contain "." then change it to ip from env
  [[ $ip =~ "." ]] || ip=`env | grep TCP=tcp:// | grep -i $link_service_name | cut -d"/" -f3 | cut -d":" -f1 | sort | uniq`
[...]

Issue #597, opened 2014-10-30
fig.yml support for --add-host?


coderlol:
Any chance fig.yml can support the docker run option --add-host?

PR #1158, opened 2015-03-23
Add extra_hosts to yml configuration --add-hosts

chernjie:
Add extra_hosts to yml configuration

Followup from:
- #597
- #754 
- #848 
- #958 

### Example:
```yml
extra_hosts:
 - "docker:162.242.195.82"
 - "fig:50.31.209.229"
```

Signed-off-by: CJ <lim@chernjie.com>

### DIFF
--- a/compose/config.py
+++ b/compose/config.py
@@ -43,6 +43,8 @@ ALLOWED_KEYS = DOCKER_CONFIG_KEYS + [
 DOCKER_CONFIG_HINTS = {
     'cpu_share': 'cpu_shares',
     'add_host': 'extra_hosts',
+    'hosts': 'extra_hosts',
+    'extra_host': 'extra_hosts',
     'link': 'links',
     'port': 'ports',
     'privilege': 'privileged',
--- a/compose/config.py
+++ b/compose/config.py
@@ -15,6 +15,7 @@ DOCKER_CONFIG_KEYS = [
     'entrypoint',
     'env_file',
     'environment',
+    'extra_hosts',
     'hostname',
     'image',
     'links',
@@ -41,6 +42,7 @@ ALLOWED_KEYS = DOCKER_CONFIG_KEYS + [
 
 DOCKER_CONFIG_HINTS = {
     'cpu_share': 'cpu_shares',
+    'add_host': 'extra_hosts',
     'link': 'links',
     'port': 'ports',
     'privilege': 'privileged',
--- a/compose/service.py
+++ b/compose/service.py
@@ -629,7 +629,17 @@ def build_extra_hosts(extra_hosts_config):
     if extra_hosts_config is None:
         return None
 
-    if isinstance(extra_hosts_config, list):
-        return dict(r.split(':') for r in extra_hosts_config)
-    else:
-        return dict([extra_hosts_config.split(':')])
+    if isinstance(extra_hosts_config, basestring):
+        extra_hosts_config = [extra_hosts_config]
+
+    extra_hosts_dict = {}
+    for extra_hosts_line in extra_hosts_config:
+        if isinstance(extra_hosts_line, dict):
+            # already interpreted as a dict (depends on pyyaml version)
+            extra_hosts_dict.update(extra_hosts_line)
+        else:
+            # not already interpreted as a dict
+            host, ip = extra_hosts_line.split(':')
+            extra_hosts_dict.update({host.strip(): ip.strip()})
+
+    return extra_hosts_dict
--- a/compose/service.py
+++ b/compose/service.py
@@ -23,6 +23,7 @@ DOCKER_START_KEYS = [
     'dns',
     'dns_search',
     'env_file',
+    'extra_hosts',
     'net',
     'pid',
     'privileged',
@@ -448,6 +449,8 @@ class Service(object):
 
         restart = parse_restart_spec(options.get('restart', None))
 
+        extra_hosts = build_extra_hosts(options.get('extra_hosts', None))
+
         return create_host_config(
             links=self._get_links(link_to_self=one_off),
             port_bindings=port_bindings,
@@ -460,6 +463,7 @@ class Service(object):
             restart_policy=restart,
             cap_add=cap_add,
             cap_drop=cap_drop,
+            extra_hosts=extra_hosts,
             pid_mode=pid
         )
 
@@ -619,3 +623,13 @@ def split_port(port):
 
     external_ip, external_port, internal_port = parts
     return internal_port, (external_ip, external_port or None)
+
+
+def build_extra_hosts(extra_hosts_config):
+    if extra_hosts_config is None:
+        return None
+
+    if isinstance(extra_hosts_config, list):
+        return dict(r.split(':') for r in extra_hosts_config)
+    else:
+        return dict([extra_hosts_config.split(':')])
--- a/compose/service.py
+++ b/compose/service.py
@@ -626,20 +626,25 @@ def split_port(port):
 
 
 def build_extra_hosts(extra_hosts_config):
-    if extra_hosts_config is None:
-        return None
+    if not extra_hosts_config:
+        return {}
 
-    if isinstance(extra_hosts_config, basestring):
-        extra_hosts_config = [extra_hosts_config]
-
-    extra_hosts_dict = {}
-    for extra_hosts_line in extra_hosts_config:
-        if isinstance(extra_hosts_line, dict):
-            # already interpreted as a dict (depends on pyyaml version)
-            extra_hosts_dict.update(extra_hosts_line)
-        else:
-            # not already interpreted as a dict
+    if isinstance(extra_hosts_config, list):
+        extra_hosts_dict = {}
+        for extra_hosts_line in extra_hosts_config:
+            if not isinstance(extra_hosts_line, six.string_types):
+                raise ConfigError(
+                    "extra_hosts_config \"%s\" must be either a list of strings or a string->string mapping," %
+                    extra_hosts_config
+                )
             host, ip = extra_hosts_line.split(':')
             extra_hosts_dict.update({host.strip(): ip.strip()})
+        extra_hosts_config = extra_hosts_dict
 
-    return extra_hosts_dict
+    if isinstance(extra_hosts_config, dict):
+        return extra_hosts_config
+
+    raise ConfigError(
+        "extra_hosts_config \"%s\" must be either a list of strings or a string->string mapping," %
+        extra_hosts_config
+    )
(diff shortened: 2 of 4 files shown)


chernjie:
@tdesvenain @sampwing I have signed your commits after rebasing from the latest master.

tdesvenain:
Great, thanks !

chernjie:
I am wondering if it is possible to add an alias to `extra_hosts`, e.g. `hosts`

pedrokiefer:
Shouldn't it be possible to add and extra_host that is another container, but not actually linked to the current container.

Something like:
```
web:
  build: .
worker:
  build: .
  extra_hosts:
     - web
```

aanand:
@pedrokiefer Why not just use a link? You'll get the `/etc/hosts` entry.

chernjie:
@pedrokiefer : What @aanand said, use `link` for containers, use `extra_hosts` for non-containers

pedrokiefer:
I had some kind of circular dependency between containers, so I thought ```extra_hosts``` would solve that to me.

chernjie:
rebased with CI fixes

chernjie:
@aanand I would appreciate your feedback/input with this PR. The integration tests are working well. Are there anything that I have missed?

compose/service.py
```diff
@@ -616,3 +620,23 @@ def split_port(port):
 
     external_ip, external_port, internal_port = parts
     return internal_port, (external_ip, external_port or None)
+
+
+def build_extra_hosts(extra_hosts_config):
+    if extra_hosts_config is None:
+        return None
+
+    if isinstance(extra_hosts_config, basestring):
```
aanand: `basestring` should be `six.string_types()` for Python 3 compatibility

aanand:
Nice. Only thing is that I think supporting the list-of-dicts format is really strange. I would rather support:

- list of `host:ip` strings
- (optionally) dictionary of `host` string to `ip` string

This would be consistent with `environment` and `labels`.

chernjie:
@aanand I do not have a strong opinion on this.
The slight difference between `list-of-dicts` vs `list-of-strings` is the spacing immediately after the colon `:`. These can easily be introduced from mistakes.

```yaml
dicts:
 docker: 162.242.195.82
 fig: 50.31.209.229
list-of-dicts:
 - docker: 162.242.195.82
 - fig: 50.31.209.229
list-of-strings:
 - docker:162.242.195.82
 - fig:50.31.209.229
```

That said, the current `build_extra_hosts` code supports all three.

@tdesvenain, do you have any input since you were the one who wrote the code?

aanand:
I agree that the syntax is error-prone (I blame YAML for being too permissive), but it feels like a hack to allow lists of dicts and we don't do it anywhere else.

Perhaps in the examples for list-of-strings, we should always quote them:

```yaml
list-of-strings:
 - "docker:162.242.195.82"
 - "fig:50.31.209.229"
```

chernjie:
@aanand updated the examples

tdesvenain:
Hi,
I'm afraid that many people who use to work with yaml will make the same
mistake,
In my humble opinion if we use the ':' character in docker-compose syntax,
we'd better use it the common way in yaml, as key/value separator.
On the top of that, we actually describe a mapping (host/ip), so dictionary
seems more accurate than string.

But actually if this introduces inconsistency, i'm ok with removing this.
There,
- we have to validate the format to alert user (before my commit we just
got an exception that looked like an issue)
- I agree that we should at least quote the whole string in documentation

2015-04-04 16:26 GMT+02:00 chernjie <notifications@github.com>:

> @aanand <https://github.com/aanand> updated the examples
>
> --
> Reply to this email directly or view it on GitHub
> <https://github.com/docker/compose/pull/1158#issuecomment-89586085>.
>



-- 
Thomas Desvenain

Téléphone : 09 51 37 35 18

tdesvenain:
After having further reflexion about it, i think my commits could be
reverted as they introduce too much inconsistency against other declarations

2015-04-04 19:01 GMT+02:00 Thomas Desvenain <thomas.desvenain@gmail.com>:

> Hi,
> I'm afraid that many people who use to work with yaml will make the same
> mistake,
> In my humble opinion if we use the ':' character in docker-compose syntax,
> we'd better use it the common way in yaml, as key/value separator.
> On the top of that, we actually describe a mapping (host/ip), so
> dictionary seems more accurate than string.
>
> But actually if this introduces inconsistency, i'm ok with removing this.
> There,
> - we have to validate the format to alert user (before my commit we just
> got an exception that looked like an issue)
> - I agree that we should at least quote the whole string in documentation
>
> 2015-04-04 16:26 GMT+02:00 chernjie <notifications@github.com>:
>
>> @aanand <https://github.com/aanand> updated the examples
>>
>> --
>> Reply to this email directly or view it on GitHub
>> <https://github.com/docker/compose/pull/1158#issuecomment-89586085>.
>>
>
>
>
> --
> Thomas Desvenain
>
> Téléphone : 09 51 37 35 18
>



-- 
Thomas Desvenain

Téléphone : 09 51 37 35 18

chernjie:
@aanand @tdesvenain I have explicitly removed support for `list-of-dicts`, added some new test to make sure that only list and mappings are supported.

compose/service.py
```diff
@@ -619,3 +623,32 @@ def split_port(port):
...
+
+
+def build_extra_hosts(extra_hosts_config):
+    if not extra_hosts_config:
+        return {}
+
+    if isinstance(extra_hosts_config, list):
+        extra_hosts_dict = {}
+        for extra_hosts_line in extra_hosts_config:
+            if isinstance(extra_hosts_line, dict):
```
aanand: I think it'd be better to check `if not isinstance(extra_hosts_line, six.string_types)` and raise an error saying `extra_hosts_config must be either a list of strings or a string->string mapping`. We shouldn't need to do anything special with the list-of-dicts case.

docs/yml.md
```diff
@@ -87,6 +87,23 @@ external_links:
...
 
+### extra_hosts
+
+Add hostname mappings. Use the same values as the docker client `--add-host` parameter.
+
+```
+extra_hosts:
+ - "docker:162.242.195.82"
+ - "fig:50.31.209.229"
+```
```
aanand: "docker" and "fig" are strange host names - why would I have a host called "fig"? (cc @moxiegirl - any best practices here?)

Could you add an example with a dict? Best if the strings are quoted, e.g.

```
extra_hosts:
  somehost: "162.242.195.82"
  otherhost: "50.31.209.229"
```

chernjie: Updated examples

compose/service.py
chernjie: Updated code

chernjie:
hmmm... @aanand Jenkins stopped running a few days ago:
https://jenkins.dockerproject.com/job/Compose-PRs/
image

compose/service.py
tdesvenain: Where the code is updated ?
Did you fix the unit tests i wrote which rely with dict format ?

2015-04-14 16:01 GMT+02:00 chernjie <notifications@github.com>:

> In compose/service.py
> <https://github.com/docker/compose/pull/1158#discussion_r28328697>:
>
> > @@ -619,3 +623,32 @@ def split_port(port):
> >
> >      external_ip, external_port, internal_port = parts
> >      return internal_port, (external_ip, external_port or None)
> > +
> > +
> > +def build_extra_hosts(extra_hosts_config):
> > +    if not extra_hosts_config:
> > +        return {}
> > +
> > +    if isinstance(extra_hosts_config, list):
> > +        extra_hosts_dict = {}
> > +        for extra_hosts_line in extra_hosts_config:
> > +            if isinstance(extra_hosts_line, dict):
>
> Updated code
>
> --
> Reply to this email directly or view it on GitHub
> <https://github.com/docker/compose/pull/1158/files#r28328697>.
>



-- 
Thomas Desvenain

Téléphone : 09 51 37 35 18

aanand:
Gave it a kick and it's up again. Looks like you need to wrap your lists in `set(...)` when asserting equality so it ignores ordering.

chernjie:
Added `.sort` instead of `set`, would this work?
cc: @aanand

chernjie:
@aanand any updates to this PR? Any feedback would be appreciated

aanand:
`set(...)` is preferable for consistency with other tests.

chernjie:
@aanand Updated with `set(...)`

aanand:
Fantastic. Could you squash and rebase? Thanks.

chernjie:
@aanand I have rebased and squashed all my commits. I have left the other author's contribution as individual commit so as not to lose their contribution information. I hope that's okay

compose/config.py
```diff
@@ -41,6 +42,7 @@
 
 DOCKER_CONFIG_HINTS = {
     'cpu_share': 'cpu_shares',
+    'add_host': 'extra_hosts',
```
dnephin: maybe `'extra_host': 'extra_hosts',` as well ?

compose/service.py
```diff
@@ -619,3 +623,28 @@ def split_port(port):
...
+                raise ConfigError(
+                    "extra_hosts_config \"%s\" must be either a list of strings or a string->string mapping," %
+                    extra_hosts_config
+                )
+            host, ip = extra_hosts_line.split(':')
+            extra_hosts_dict.update({host.strip(): ip.strip()})
+        extra_hosts_config = extra_hosts_dict
+
+    if isinstance(extra_hosts_config, dict):
+        return extra_hosts_config
```
dnephin: It might be a good idea to only support one of these formats.  We've supported both in the past, but I'm not sure that's necessary going forward. It seems to cause confusion and extra code.

If we decide to keep lists (or both), I would suggest `<host>=<ip>` (equals instead of colon) syntax instead. If you forget the quotes it confuses yaml into things its a mapping, and that can be very counter-intuitive to users.  If we use an equals, it should understand that it is a string, not a mapping.

If we decide to keep mappings (or both), we should probably do some basic validation of the keys and values, maybe even just assert that they are Truthy (not None or empty strings).

dnephin: Oh, I just realized this discussion has already happened. I'll defer to whatever @aanand decides now that I'm registered my concerns.

dnephin:
I like this, I think it addresses the "external" type described in #988

LGTM

compose/config.py
chernjie: Added @ 86a08c0

aanand:
LGTM

rysiekpl:
Oh this is sweet, and I'm kind of craving this feature right now. Any date set for 1.3.0 release?

PR #1160, opened 2015-03-23
Fix volume merging

aanand:
There were a couple of silly mistakes in the volume merging code. Fixed, with regression tests.

### DIFF
--- a/tests/unit/config_test.py
+++ b/tests/unit/config_test.py
@@ -40,26 +40,44 @@ class ConfigTest(unittest.TestCase):
 
 
 class MergeTest(unittest.TestCase):
-    def test_merge_volumes(self):
+    def test_merge_volumes_empty(self):
         service_dict = config.merge_service_dicts({}, {})
         self.assertNotIn('volumes', service_dict)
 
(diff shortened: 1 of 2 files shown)


tests/unit/config_test.py
```diff
@@ -39,6 +39,42 @@ def test_config_validation(self):
...
+            {'volumes': ['/bar:/code', '/quux:/data']},
+        )
+        self.assertEqual(set(service_dict['volumes']), set(['/bar:/code', '/quux:/data']))
+
+        service_dict = config.merge_service_dicts(
+            {'volumes': ['/foo:/code', '/quux:/data']},
+            {'volumes': ['/bar:/code', '/data']},
+        )
+        self.assertEqual(set(service_dict['volumes']), set(['/bar:/code', '/data']))
+
```
dnephin: Normally this would be 6 separate tests, so the test names can describe the case that is being tested.  It's not completely clear to which the difference between each of these without inspecting each in detail.

aanand: Agreed - I've split them out.

dnephin:
LGTM

PR #1161, opened 2015-03-23
Revert labels and dependencies

aanand:
When we updated our docker-py and requests versions in #1100, we inadvertently introduced a dependency on a version of requests that [doesn't play nicely with PyInstaller](http://stackoverflow.com/questions/28775276/pyinstaller-error-importerror-no-module-named-requests-packages-chardet-sys).

Since this hasn't been fixed yet, and the only feature we have that depends on an up-to-date docker-py is [labels](https://github.com/docker/compose/pull/1145), I'm backing it out and resetting our pinned versions.

We can still specify a wider version range in `setup.py` for people who install with pip and want to use newer versions of docker-py or requests, but for the purposes of building the binary, we're sticking with docker-py 1.0.0 and requests 2.2.1 for now.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,18 +15,13 @@ RUN set -ex; \
     ; \
     rm -rf /var/lib/apt/lists/*
 
-# ENV ALL_DOCKER_VERSIONS 1.6.0
+ENV ALL_DOCKER_VERSIONS 1.3.3 1.4.1 1.5.0
 
-# RUN set -ex; \
-#     for v in ${ALL_DOCKER_VERSIONS}; do \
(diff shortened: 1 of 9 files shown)


dnephin:
I am curious, have you tried with https://pypi.python.org/pypi/requests/2.6.0 ? It might be easier to use a newer version instead of reverting? Seems like there was a change related to the problem described in the SO: https://github.com/kennethreitz/requests/compare/v2.5.3...v2.6.0

aanand:
Sadly, using requests 2.6.0 doesn't seem to appease PyInstaller.

dnephin:
That's too bad,

LGTM

aanand:
FYI, I've reported that we're still having the issue at the bottom of this thread: https://github.com/kennethreitz/requests/issues/2465#issuecomment-85642105

PR #1172, opened 2015-03-24
Fix regression in Dns and DnsSearch settings

aanand:
Dns and DnsSearch are coming out as `[""]` instead of `None` when no value is provided.


dnephin:
LGTM

Issue #1168, opened 2015-03-24
compose run with commit option


madwire:
It would be great if compose had a commit option on the run command, this will help to improve productivity with various package manages. For example if Im building a rails app and I would like to change/update a dependancy. I would have to rebuild the docker image every time which can take some time. If after the run command it could commit the changes made to the docker image then it would save a load of time.

Something like this? `docker-compose run -c web bundle update`

I have build a basic shell script which I currently use which is great, but it would be better if its built in.

```shell
#!/bin/bash

dir=${PWD##*/} # Get current directory name 
image_base=${dir//_/} # remove underscores
service_name=$1 # Service name
docker_image_name=$image_base'_'$service_name # build docker compose image name

docker-compose run $service_name $2 $3 $4 $5 $6 # pass args to service
run_container=$(docker ps -l -q) #get the last run container id
docker commit $run_container $docker_image_name #commit the image changes
docker rm $run_container # remove the run container
```
How I use the script is like this `docker-compose-commit web bundle update`

madwire:
So I've open a PR for this #1178

dnephin:
See https://github.com/docker/compose/pull/1178#issuecomment-134369653 for discussion about alternatives to this approach.

PR #1178, opened 2015-03-25
Add commit to run command

madwire:
Added a commit option on the run command, this will help to improve productivity with various package manages. For example if Im building a rails app and I would like to change/update a dependancy. I would have to rebuild the docker image every time which can take some time. If after the run command it could commit the changes made to the docker image then it would save a load of time.

Also see #1168 

Signed-off-by: Richard Adams <richard@madwire.co.uk>

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -283,6 +283,8 @@ class TopLevelCommand(Command):
                                   to the host.
             -T                    Disable pseudo-tty allocation. By default `docker-compose run`
                                   allocates a TTY.
+            -c                    Commit the container's file changes or settings after the run command
+                                  into the service image.
         """
         service = project.get_service(options['SERVICE'])
 
@@ -344,6 +346,13 @@ class TopLevelCommand(Command):
             service.start_container(container)
             dockerpty.start(project.client, container.id, interactive=not options['-T'])
             exit_code = container.wait()
+            if options['-c'] and exit_code == 0:
+                # if the container successfully exited then commit the changes to the service image
+                log.info("Committing %s..." % service.full_name)
+                project.client.commit(
+                    container=container.id,
+                    repository=service.full_name
+                )
             if options['--rm']:
                 log.info("Removing %s..." % container.name)
                 project.client.remove_container(container.id)
--- a/tests/integration/cli_test.py
+++ b/tests/integration/cli_test.py
@@ -318,6 +318,48 @@ class CLITestCase(DockerClientTestCase):
         self.assertIn("0.0.0.0", port_random)
         self.assertEqual(port_assigned, "0.0.0.0:49152")
 
+    @patch('dockerpty.start')
+    def test_run_service_with_commit(self, _):
+        self.command.base_dir = 'tests/fixtures/simple-dockerfile'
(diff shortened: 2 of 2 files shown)


dnephin:
I don't think this is going to work. The container is discarded and recreated on the next `up`, so that commit isn't going to really do anything.

More generally I don't think this really belongs in compose. It doesn't really have anything to do with composition. If you're just iterating on a `Dockerfile` you can do that just the default docker cli.

In general I think commit is only used in rare situations where you don't care about repeatable builds, usually you want to rebuild the container.

There are some other ways to work around the problem you describe. Best practice is to group all your `apt-get install` all together in a single line, but while developing the container you can just keeping adding new `RUN` lines. That way all your previous installs are cached, and you don't have to wait for them again.  At the end when you're done you can just merge them together into a single `RUN` line.

Another option is to use http://jpetazzo.github.io/2014/06/17/transparent-squid-proxy-docker/ to cache the downloads, which saves a bunch of time.

madwire:
@dnephin Sorry I want to make it clear this is not to be used for OS packages, Just for application dependencies manages like Bundler, NPM, Pip etc.

I do agree with you that in most cases this should not be the way to go. But some dependency manages create lock files after you have ran them and a quick commit of the installed files and lock file does speed things up. The lock files are also synced back to the end users app directory so this is repeatable.

As for a proxy cache that only works for packages that have already been compiled or don't need to be, which is not the case for all.

In general you would only use it in rare situations yes, but it doesn't mean it doesn't have a place.

sorby69:
+1

aanand:
:-1: The problem you describe is a real and annoying one, but I don't want to encourage non-repeatable practices like this.

madwire:
@aanand Sorry I may be missing the bigger picture, could you elaborate on what you mean by non-repeatable?

I've been using a bash script equivalent for about a month and only on running bundle update and it works great. Fast, reliable and repeatable.

DVG:
:+1: This is one of the big pain points in using docker/compose for a developing application where application-level dependencies (things that don't go in a Dockerfile) change frequently. It can take 5-10 minutes to do a bundle install from scratch for a Rails project, versus a few seconds to just install the new gemdependency. Being able to quickly automate things like that would be a big plus in favor of docker as a development environment.

dnephin:
You can always use volumes during development to handle  the changing of dependencies.

dnephin:
Thanks for the contribution! Unfortunately this is not something we want to include in compose.

Committing the result of a run command is not a repeatable process (there are too many external factors which can't be accounted for). There are already a few options for dealing with the problem this intends to fix:

* use a cache to reduce download time
* optimize the install process (https://github.com/Yelp/venv-update and http://pythonwheels.com/ are both good options for python. Unfortunately I don't know enough about ruby to suggest options)
* build as part of a "builder" container that uses a host volume mount, produce a tarball, then add the tarball to the image (https://github.com/openshift/source-to-image describes that pattern). This is very similar to what you're doing with commit, but it's more explicit about what is happening.
* use a volume mount for the source in development (where the dependencies change frequently), and use an `ADD` for CI/CD (where they don't change nearly as often).

Issue #1184, opened 2015-03-26
pep8 for config files and tests


akoskaaa:
I was scanning through the config files during installing and I noticed that some of them and some test files are not pep8 compatible.

akoskaaa:
#dibs

PR #1185, opened 2015-03-26
Make test files and config files pep8 valid

akoskaaa:
### What does it do?
Upon reading some config files I noticed they are not pep8 valid, so I ran the command on the root of compose. There were only issues in config files and tests. This pull request proposes a solution for this, in the name of clean code.
Resolves https://github.com/docker/compose/issues/1184

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -8,7 +8,7 @@ deps =
     -rrequirements-dev.txt
 commands =
     nosetests -v {posargs}
-    flake8
+    flake8 compose tests setup.py
 
 [flake8]
 # ignore line-length for now
--- a/tox.ini
+++ b/tox.ini
@@ -8,7 +8,7 @@ deps =
     -rrequirements-dev.txt
 commands =
     nosetests -v {posargs}
-    flake8 compose
+    flake8
 
 [flake8]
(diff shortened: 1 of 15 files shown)


dnephin:
Nice!

Now that all the files are compliant please add `tests`  to the `flake8` line in `script/test-versions` and `tox.ini` so that any new lines are checked as well.

akoskaaa:
Sweet, thanks for noticing. The flake8 tests actually caught a few more stuff this way. Will fix and create a new commit when I get the green light.

bfirsh:
Nice. LGTM

tox.ini
```diff
@@ -8,7 +8,7 @@ deps =
     -rrequirements-dev.txt
 commands =
     nosetests -v {posargs}
-    flake8 compose
+    flake8
```
dnephin: This can cause problems when you have extra files in the directory. If you were to create a another virtualenv it will try and check every file in that virtualenv.

I think this should be:

```
flake8 compose tests setup.py
```

akoskaaa: Completely valid point. Fixed in 4e0f555

aanand:
Thanks!

Issue #1102, opened 2015-03-12
ports not accepting a range in yml


awsmsrc:
the command following command works as of docker 1.5:
```
docker run -p 3000-3002:3000-3002 port_test
```
specifying ports this way in the docker-compose.yml always results in a cannot start container exit status 2

tvcutsem:
Ran into the same issue yesterday. Specifying the port range as a string also doesn't help. Here's the offending section from my yml file:

```
ports:
  - "50000-50050:50000-50050"
```

dougborg:
We have run into this issue as well. We were expecting something like this to "just work":

```yml
ports:
  - "1000-1019"
```

uvgroovy:
#dibs

awsmsrc:
gonna work on this one at the docker party tonight

uvgroovy:
hi @awsmsrc, I already started working on FYI - do you want to help us?

DXist:
+1

thecodethinker:
Has this been in any release yet?

awsmsrc:
yeah would be good to get an update @aanand

claudeatsafe:
+1

aanand:
It's been implemented by @uvgroovy in https://github.com/docker/compose/pull/1191, but needs rebasing.

uvgroovy:
@aanand  rebasing done

reidrac:
+!

willbarrett:
Looks like the implementation just needs docs and a rebase. I'm sure we'd all be grateful to @uvgroovy if that happened soon :)

uvgroovy:
Hi,
I just realized that missed the documentation request from the PR... sorry for that - on it right now
thanks @willbarrett

tpires:
Can someone close this? Looks like this is already merged [here](https://github.com/docker/compose/pull/1827).

jasonm-makerbot:
:+1:

brunoborges:
Why only "ports:" supports range? Why not "expose:" too?

Also, is it possible to fix the documentation to go into more detail on the differences between ports and expose? It is still not clear on what happens in each case.

vpillac:
Hi, my docker-compose files stopped working after updating to docker-compose version 1.5.2 (from MacOS Docker Toolbox 1.9.1).

```
Service 'xxx' configuration key 'expose' '0' is invalid: should be of the format 'PORT[/PROTOCOL]'
```

While the file contains
```
  expose:
[...]

PR #1075, opened 2015-03-06
Support alternate Dockerfile name.

KyleJamesWalker:
Add support for compose support Docker's `-f`/`--file` option from within your docker-compose.yml file.

```yaml
app:
  build: .
  dockerfile: Dockerfile-alternate
```

I quickly threw this in, and it seems to work for me, please let me know if anything else is needed to get this into master.

PR #1191, opened 2015-03-26
add port range support to docker compose

uvgroovy:
fixes https://github.com/docker/compose/issues/1102
tests passed


aanand:
Thanks for contributing!

- Could you add some integration tests?
- Could any of this logic be moved into docker-py?

compose/service.py
```diff
@@ -590,17 +590,51 @@ def build_volume_binding(volume_spec):
...
+        port, protocol = parts
+        protocol = "/" + protocol
+
+    parts = str(port).split('-')
+
+    if len(parts) == 1:
+        return ["%s%s" % (port, protocol)]
+
+    if len(parts) == 2:
+        return ["%s%s" % (p, protocol) for p in range(int(parts[0]), int(parts[1]) + 1)]
```
dnephin: Can anyone confirm this needs to be handled by the client? 

It seems like it should be handled by dockerd. I can't find any reference to it in the remote api docs.

uvgroovy: When I looked at the docker API, the data structure that I saw there only contained one port, that's why I chose to implement it on the client.
see ExposedPorts at https://docs.docker.com/reference/api/docker_remote_api_v1.17/

uvgroovy: Also looking at the docker client source code, you can also see that the ports are used individually.
see runconfig/parse.go line 229-238 (the line with ParsePortRange)

uvgroovy: As for integration tests - I can add those later tonight, but let's first decide if the general idea is acceptable. i.e. it should be done here, vs the docker server and docker-py

aanand: Looks like Docker does indeed do it client-side, so that's where we'll need to do it too. Moving it to the server is a discussion to be had on docker/docker.

Paging @shin-: does processing of port ranges belong in docker-py, in your opinion?

shin-: Hmm, possibly. But probably as a function in the tools module rather than inside the API methods.

aanand: @shin- Thanks, I agree - I think we can move `build_port_bindings` and its dependencies (`to_port_range`, `split_port`, `add_port_mapping`) all over to `docker.utils`, perhaps as a submodule (`docker.utils.ports`).

@uvgroovy Could you submit a PR to docker-py?

uvgroovy: @aanand So basically move build_port_bindings and deps to docker.utils.ports and then import it in compose/service.py?

I can do that, just one question - 
doing that will temporary break the code, unless requirments-dev is updated.
so basically, i think we need to:
1. add build_port_bindings to docker-py
2. release new docker-py
3. remove build_port_bindings from compose/service.py, import it from docker-py and update requirements-dev.txt to the new docker-py version

sounds good? any comments\feedback?

aanand: I think it should go:

1. Submit PR to docker-py
2. Update this PR with the code changes and set the docker-py requirement in `requirements.txt` to point at your repo and branch with a git URL (see the pyinstaller requirement in `requirements-dev.txt` for an example)
3. When the docker-py PR is merged, update the git URL to point at the docker-py merge commit
4. When the next docker-py is released, update `requirements.txt` to the stable version
5. Merge this PR

uvgroovy: sounds good.

dougborg:
Hey @uvgroovy, just wanted to check in with you to see if this was still moving forward. Do you need any help with testing, etc.? Let me know. I'd be glad to help any way I can.

uvgroovy:
Hi, Sorry for the delay, jewish holidays are slowing me down...
I got the changes to my branch in the docker-py repository, and now just trying to verify that it works with compose, and then i'll submit a PR for docker-py

uvgroovy:
I updated the version of requests to version 2.5.2 to match the one required from docker-py;

but now, I encounter this error:
ImportError: No module named 'requests.packages.chardet.sys'
see:
http://stackoverflow.com/questions/28775276/pyinstaller-error-importerror-no-module-named-requests-packages-chardet-sys

@dougborg  \ @aanand would love for some advice :)

dougborg:
I tried to reproduce this error on my local development environment with your branch, but was unable to. Can you provide any more details as to what your setup looks like and how you are invoking your tests?

I am running:

OSX Yosemite 10.10.2
docker-machine version 0.2.0 (48bea37)
docker-compose 1.1.0 (ee64e0d)
docker 1.6-rc4

uvgroovy:
So basically, When I tried to build compose with the docker-py repo directly. this is my requirements.txt:
```
PyYAML==3.10
git+https://github.com/uvgroovy/docker-py.git@8028c988ce363345988bc23444f08e27394a7fb5#egg=docker-py
dockerpty==0.3.2
docopt==0.6.1
requests==2.2.1
six==1.7.3
texttable==0.8.2
websocket-client==0.11.0
```

it complained that requests 2.2.1 is too old:
```
./script/test
...
Installed /usr/local/lib/python2.7/dist-packages/docker_compose-1.1.0-py2.7.egg
Processing dependencies for docker-compose==1.1.0
error: Installed distribution requests 2.2.1 conflicts with requirement requests>=2.5.2
...
```
 so I changed the requirements.txt for compose in my branch to contain requests 2.5.2 and got the error i mentioned.
```
 ImportError: No module named 'requests.packages.chardet.sys'
```

awsmsrc:
hey guys whats happening with this?
Anything I can do to help?

aanand:
Same issue as https://github.com/docker/compose/pull/1075 and https://github.com/docker/compose/pull/1269 - we're waiting for an upstream fix in requests (https://github.com/kennethreitz/requests/pull/2533) to make it into a release.

uvgroovy:
@aanand , perhaps can you assist with merge the docker-py PR?
 https://github.com/docker/docker-py/pull/544

uvgroovy:
requests 2.6 was released with a fix to this issue. updating both docker-py and compose to 2.6.2 seems to be working (tried it in my own laptop and the tests pass) - but here jenkins is specifically looking for requests>=2.2.1,<2.6
and as a result uses 
Best match: requests 2.5.3

I couldn't find a reference to requests "<2.6" in the repo it self.

Any idea why that happens\how to fix this?

aanand:
Sorry I missed this. Needs rebasing on master, then it should pick up the new requests version range.

uvgroovy:
i re-did my commit on the master from upstream compose. all should be well now

aanm:
How will this behave if we scale containers that are greater than the number of ports available?

uvgroovy:
@aanm  not sure what you are asking. this PR just allows you to specify multiple ports in the configuration. it is as if you wrote them individually (i.e. the code 'translates' the port range to multiple single ports for the rest of the program).

aanm:
Currently, compose sends a user error if you try to scale a container that has ports defined. For example if you have `port: 8888` and you want to scale that service to 2, it won't work because the `compose.yml` only has one port defined. What I was asking is, if the same occurs if you have `port: 8888-8889` and you try to scale that service to 3.

aanand:
@aanm The same will happen, yes.

It would be nice to be able to specify a set of ports where each one is assigned to a single container in the service, but it's not what the Docker CLI's port range syntax means, so we'd need to come up with an alternate way of specifying it.

aanm:
@aanand I think it would also be nice to NOT throw an error on the case I was talking about, either with a range or a single port specification. I think it would be better to just throw a warning. I'm saying this because compose + swarm fail if we try to scale up with a single port defined.

dnephin:
Replacing the error with a warning seems appropriate now that swarm makes that case possible.

aanm:
@dnephin https://github.com/docker/compose/pull/1466

uvgroovy:
@aanand re-did the commit to the updated compose master.

aanand:
Thanks!

LGTM

aanand:
Strike that - still needs docs.

uvgroovy:
@aanand sorry for the delay (i missed your last comment) - I added documentation to yml.md - please let me know if that works.

Also, while it seems that tests failed, it actually looks like an IT issue:
"unable to access 'https://github.com/docker/compose.git/': server certificate verification failed"
any way to re-run the tests?

docs/yml.md
```diff
@@ -102,7 +102,7 @@ An entry with the ip address and hostname will be created in `/etc/hosts` inside
 ### ports
 
 Expose ports. Either specify both ports (`HOST:CONTAINER`), or just the container
-port (a random host port will be chosen).
+port (a random host port will be chosen). You can specify a port range instead of a single port (`START-END`). If you use a range for the container ports, you may specify a range for the host ports as well. both ranges must be of equal size.
```
aanand: ping @moxiegirl

My take: this seems to suggest on first reading that you can specify a range of container ports but a single host port, which I don't think is correct. Perhaps:

***

Expose ports. You can specify just a container port (which will be mapped to a random host port), or both (`HOST:CONTAINER`).

You can optionally pass in ranges of port numbers. If you're specifying host ports as well as container ports, the host and container port ranges must be the same size.

docs/yml.md
```diff
@@ -102,7 +102,7 @@ An entry with the ip address and hostname will be created in `/etc/hosts` inside
 ### ports
 
 Expose ports. Either specify both ports (`HOST:CONTAINER`), or just the container
```
moxiegirl: @uvgroovy thanks for the contribution. We just sorted this same material out on a docker/docker pr.  So, let's try it here:

---> https://gist.github.com/moxiegirl/3dded271198e4107c4d2

### ports

Makes an exposed port accessible on a host and the ports are available to
any client that can reach that host. Docker binds the exposed port to a random
port on the host within an *ephemeral port range* defined by
`/proc/sys/net/ipv4/ip_local_port_range`. You can also map to a specific port or range of ports.

Acceptable formats for the `ports1 value are:

* `containerPort`
* `ip:hostPort:containerPort`
* `ip::containerPort`
* `hostPort:containerPort`

You can specify a range for both the `hostPort` and the `containerPort` values.
When specifying ranges, the container port values in the range must match the
number of host port values in the range, for example, 
`1234-1236:1234-1236/tcp`. Once a host is running, use the 'docker port' command
to see the actual mapping.

The following configuration shows examples of the port formats in use:

    ports:
     - "3000"
     - "3000-3005"
     - "8000:8000"
     - "9090-9091:8080-8081"
     - "49100:22"
     - "127.0.0.1:8001:8001"
     - "127.0.0.1:5000-5010:5000-5010"


>**Note:** When mapping ports, in the `hostPort:containerPort` format, you may
>experience erroneous results when using a container port lower than 60. This
>happens because YAML parses numbers in the format `xx:yy` as sexagesimal (base
>60). To avoid this problem, always explicitly specify your port
>mappings as strings.

aanand: Looks good. The only thing I'd change is this line:

> Once a host is running, use the `docker port` command to see the actual mapping.

This should instead read `docker-compose port`.

moxiegirl: @aanand updated comment and the gist.

yangxuesong:
the lastest release 1.4.0rc2 still not support port range.
when will it be available?

mnowster:
Merged in https://github.com/docker/compose/pull/1827

PR #1162, opened 2015-03-24
Update getting started docs

paulczar:
addresses #1154

Signed-off-by: Paul Czarkowski <username.taken@gmail.com>

PR #1199, opened 2015-03-26
using the same files in presentation and quick start to fix #1148

### LINKED ISSUE #1148: Compose documentation
On the main Docker Compose documentation (https://docs.docker.com/compose/),starting at the top, it describes three steps that need to be taken to get things up and running with a small sample using Python and Postgres (not the Python, Flask, and Redis example later).  It isn't clear for the new user that this first example is not an App, especially when the steps include what looks to be "docker-compose up" in the description which indicates that something should execute.

I believe it would be more clear to the novice that the first example isn't something that will run if the example Dockerfile and docker-compose.yml file were removed; leaving on the description of the three steps.  Later, the Quick Start section describes a fully functioning example with all the necessary files.

The confusion created here for the novice could dissuade them from attempting to use Compose.

chanezon:
Signed-off-by: Patrick Chanezon <patlist@chanezon.com>

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -20,11 +20,9 @@ reproduced anywhere:
 
 ```Dockerfile
 FROM python:2.7
-WORKDIR /code
-ADD requirements.txt /code/
-RUN pip install -r requirements.txt
 ADD . /code
-CMD python app.py
+WORKDIR /code
+RUN pip install -r requirements.txt
 ```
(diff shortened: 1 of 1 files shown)


chanezon:
After chatting with @aanand, closing this and adding to https://github.com/docker/compose/pull/1162 instead

Issue #1081, opened 2015-03-08
docker-compose won't install [Linux mint 17.1 | Python3 | Anaconda3]


ruxi:
According to the [docs](http://docs.docker.com/compose/install/), these are the commands to install compose:
```
curl -L https://github.com/docker/compose/releases/download/1.1.0/docker-compose-`uname -s`-`uname -m` > /usr/local/bin/docker-compose

 chmod +x /usr/local/bin/docker-compose
```

When I do that, this is what returned:
```
bash: /usr/local/bin/docker-compose: Permission denied
```

So I tried the pip method
```
 sudo pip install -U docker-compose
```

It also gave me some errors:
```
Traceback (most recent call last):
  File "/usr/bin/pip", line 9, in <module>
    load_entry_point('pip==1.5.4', 'console_scripts', 'pip')()
  File "/usr/lib/python2.7/dist-packages/pkg_resources.py", line 351, in load_entry_point
    return get_distribution(dist).load_entry_point(group, name)
  File "/usr/lib/python2.7/dist-packages/pkg_resources.py", line 2363, in load_entry_point
    return ep.load()
  File "/usr/lib/python2.7/dist-packages/pkg_resources.py", line 2088, in load
    entry = __import__(self.module_name, globals(),globals(), ['__name__'])
  File "/usr/lib/python2.7/dist-packages/pip/__init__.py", line 11, in <module>
    from pip.vcs import git, mercurial, subversion, bazaar  # noqa
  File "/usr/lib/python2.7/dist-packages/pip/vcs/mercurial.py", line 9, in <module>
    from pip.download import path_to_url
  File "/usr/lib/python2.7/dist-packages/pip/download.py", line 25, in <module>
    from requests.compat import IncompleteRead
ImportError: cannot import name IncompleteRead
```

docker-compose --version

```
Traceback (most recent call last):
  File "/home/username-ssd/anaconda3/bin/docker-compose", line 9, in <module>
    load_entry_point('docker-compose==1.1.0', 'console_scripts', 'docker-compose')()
  File "/home/username-ssd/anaconda3/lib/python3.4/site-packages/setuptools-12.0.4-py3.4.egg/pkg_resources/__init__.py", line 519, in load_entry_point
  File "/home/username-ssd/anaconda3/lib/python3.4/site-packages/setuptools-12.0.4-py3.4.egg/pkg_resources/__init__.py", line 2630, in load_entry_point
[...]

PR #1201, opened 2015-03-27
Update install documentation

asveepay:
#1081 sudo may be required to install on Ubuntu and other Linux flavors

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -23,7 +23,7 @@ To install Compose, run the following commands:
     curl -L https://github.com/docker/compose/releases/download/1.1.0/docker-compose-`uname -s`-`uname -m` > /usr/local/bin/docker-compose
     chmod +x /usr/local/bin/docker-compose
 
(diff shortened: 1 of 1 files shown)


aanand:
Thanks!

It's actually nothing to do with Linux - it happens when `/usr/local/bin` isn't writable by the current user. The message should therefore be more generic - perhaps something like:

> **Note:** If you get a "Permission denied" error, your `/usr/local/bin` directory probably isn't writable and you'll need to install Compose as the superuser. Run `sudo -i`, then the two commands above, then `exit`.

If you prefix it with a `>`, as I've done above, it'll be formatted like other hints in the documentation (e.g. [here](http://docs.docker.com/examples/nodejs_web_app/)).

ping @fredlf

asveepay:
Updated as specified

aanand:
Thanks. Could you squash to one commit and sign it off please?

PR #1202, opened 2015-03-27
WIP: Jenkins script

aanand:
I've moved the DCO validation step out of `script/test-versions`, where it really doesn't belong, and added `script/ci`, which:

1. Performs DCO validation
2. Runs tests against all Docker versions
3. Builds the Linux binary

Currently the last step is failing with an odd `Permission denied` error, which I'm still investigating.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,6 +25,9 @@ RUN set -ex; \
     curl https://test.docker.com/builds/Linux/x86_64/docker-1.6.0-rc2 -o /usr/local/bin/docker-1.6.0-rc2; \
     chmod +x /usr/local/bin/docker-1.6.0-rc2
 
+# Set the default Docker to be run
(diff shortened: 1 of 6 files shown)


dnephin:
sounds like a good idea to me

aanand:
OK, should be working now.

@jfrazelle - could you try updating the Jenkins command to:

    $ TAG="docker-compose:$(git rev-parse --short HEAD)"
    $ docker build -t "$TAG" .
    $ docker run --rm --volume="/var/run/docker.sock:/var/run/docker.sock" --volume="$(pwd):/code" -e "TAG=$TAG" --entrypoint="script/ci" "$TAG"

jfrazelle:
will dpo!

jfrazelle:
done!

PR #1205, opened 2015-03-27
[cli] run --rm overrides restart: always

### LINKED ISSUE #1013: docker-compose run command continuously restarting
##### Description of problem:

An unexpected behaviour is encountered when using `docker-compose run` with services defining `restart: always`.

`docker version`:
```
Client version: 1.5.0
Client API version: 1.17
Go version (client): go1.4.1
Git commit (client): a8a31ef
OS/Arch (client): linux/amd64
Server version: 1.5.0
Server API version: 1.17
Go version (server): go1.4.1
Git commit (server): a8a31ef
```

`uname -a`:
Linux docker01 3.13.0-45-generic #74-Ubuntu SMP Tue Jan 13 19:36:28 UTC 2015 x86_64 x86_64 x86_64 GNU/Linux

##### Environment details :
Host running inside a VMWare VM.

##### How reproducible:

Use this docker-compose file :

```yaml
db:
    image: mysql
    restart: always
    environment:
        MYSQL_ROOT_PASSWORD: mypass
```

##### Steps to Reproduce:

1.
```Shell
$ docker-compose up -d
Creating mysql_db_1...
```

2.
```
$ docker-compose run --rm db sh -c 'exec mysql -h db -uroot -p"$DB_ENV_MYSQL_ROOT_PASSWORD"'
mysql> exit
Removing mysql_db_run_1...
Conflict, You cannot remove a running container. Stop the container before attempting removal or use -f
```

##### Actual Results:

`mysql_db_run_1` is still running.

##### Expected Results:

mysql_db_run_1 stopped and removed.

##### Additional info:

If I don't specify the '--rm' option, the result is the same but without the conflict message.

It looks like `docker-compose run` started the container with 'restart: always' option. I did not find a way to override this behavior from the `docker-compose run` command. Maybe `docker-compose run` shall ignore `restart: always` in docker-compose.yml (at least since the `--rm` flag has been specified) or provide a way to override this ?

josephpage:
Fix #1013 

I think docs is not needed because it's an intuitive behavior.

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -325,6 +325,9 @@ class TopLevelCommand(Command):
         if options['--entrypoint']:
             container_options['entrypoint'] = options.get('--entrypoint')
 
+        if options['--rm']:
+            container_options['restart'] = None
+
         if options['--user']:
             container_options['user'] = options.get('--user')
 
--- a/tests/unit/cli_test.py
+++ b/tests/unit/cli_test.py
@@ -134,6 +134,55 @@ class CLITestCase(unittest.TestCase):
             call_kwargs['environment'],
             {'FOO': 'ONE', 'BAR': 'NEW', 'OTHER': 'THREE'})
 
+    def test_run_service_with_restart_always(self):
+        command = TopLevelCommand()
+        mock_client = mock.create_autospec(docker.Client)
+        mock_project = mock.Mock()
+        mock_project.get_service.return_value = Service(
+            'service',
+            client=mock_client,
+            restart='always',
+            image='someimage')
+        command.run(mock_project, {
+            'SERVICE': 'service',
+            'COMMAND': None,
+            '-e': [],
+            '--user': None,
+            '--no-deps': None,
+            '--allow-insecure-ssl': None,
+            '-d': True,
+            '-T': None,
+            '--entrypoint': None,
+            '--service-ports': None,
+            '--rm': None,
+        })
+        _, _, call_kwargs = mock_client.create_container.mock_calls[0]
+        self.assertEquals(call_kwargs['host_config']['RestartPolicy']['Name'], 'always')
+
+        command = TopLevelCommand()
+        mock_client = mock.create_autospec(docker.Client)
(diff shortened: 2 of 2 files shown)


josephpage:
All tests pass on local, why they failed with Jenkins ?

aanand:
Your commit needs signing off:

    $ git commit --amend --reuse-message=HEAD --signoff
    $ git push --force

aanand:
Also, could we get an integration test please? Thanks!

josephpage:
@aanand Because the container is created and immediately removed by `docker-compose run --rm` command, it's difficult to test that the container doesn't have a restart policy.
Note that the `--rm` option is not tested today.

What do you think about it ?

```python
    @patch('dockerpty.start')
    def test_run_service_with_restart_always(self, __):
        name = 'simple'
        self.command.base_dir = 'tests/fixtures/restart-composefile'

        self.command.dispatch(['run', name, '/bin/echo', 'helloworld'], None)
        service = self.project.get_service(name)
        container = service.containers(one_off=True)[0]
        self.assertEqual(container.get('HostConfig.RestartPolicy.Name'), 'always')

        self.command.dispatch(['run', '--rm', name, '/bin/echo', 'helloworld'], None)
        self.assertEqual(len(service.containers()), 0)
        self.assertEqual(len(service.containers(stopped=True)), 0)
```

with this `tests/fixtures/restart-composefile/docker-compose.yml` : 

```yml
simple:
  image: busybox:latest
  restart: always
```

aanand:
Hm, good point. In that case, maybe just a unit test - use `mock_client` to check that `create_container()` is called with `restart=None`.

josephpage:
Just added a unit test. Seems good to you @aanand ?

jaimegildesagredo:
@aanand some updates here? we would need this fix merged. If I can be of some help just ping me.

dnephin:
LGTM

aanand:
OK, looks like the CLI code is too tightly coupled to improve that unit test without a lot of refactoring.

LGTM

Issue #1194, opened 2015-03-26
Path supplied to `build` should be relative to compose file


aanand:
As is now the case with paths supplied to `env_file` and `volumes`.

Issue #1209, opened 2015-03-27
Add support for pointing to tarball contexts in the 'build' parameter.


moysesb:
Since the docker client used by compose (docker-py) supports receiving externally generated tarballs as a parameter for its build() method, this alternative should be available in the project configuration file. i.e.:

> service:
> 
> > build: path/to/tarball.tar.gz

The  contents within the compressed file should pass the same level of validation as the currently supported directory paths (for example, 'there must be a Dockerfile at the root inside the tar archive'). Aside from that, the general idea is that the archive is some kind of externally validated artifact, auto-generated by a continuous integration stack, so there should be no need for parsing .dockerignore. 

If this sounds like a good idea I can start working on a PR (I already have an working prototype).

dnephin:
It seems reasonable to me that `build:` would support tarballs as well

aanand:
Seems like this change belongs in docker-py.

moysesb:
@aanand docker-py already has this feature implemented, the issue is that compose does not use it. As it's currently implemented, compose assumes the value of `build:` to be a directory path, the value is passed "as is" to docker-py. The feature being requested is that compose should call docker-py's build method with the right combination of arguments when the `build:` parameter points to a tarball. In services.py:
```python
def build(self, no_cache=False):
   log.info('Building %s...' % self.name)
"""
 - Here it should be determined what kind of resource self.options['build'] points to.
  - If it's a tarball, the call to self.client.build should look something like:
    build_output = self.client.build(
        fileobj=open(self.options['build'], 'rb'),
        custom_context=True,
        encoding='gzip',
        tag=self.full_name,
        stream=True,
        rm=True,
        nocache=no_cache,
    )
  - If it's a directory, use the already existing call:
"""
build_output = self.client.build(
    self.options['build'],
    tag=self.full_name,
    stream=True,
    rm=True,
    nocache=no_cache,
)
```
The point is that the actual context that docker-py transmits over to the docker daemon is a tarball, whether the caller passes a directory path, a tarball or a `Dockerfile` directly. As it is now, compose only knows how to pass a directory, so if I generate a tarball externally as part of some build process that has nothing to do with compose, I have to gratuitously decompress it and then have docker-py compress it again (in python!) before sending it to docker.

moysesb:
@aanand It would certainly be possible to change docker-py so that it automatically detects when the context being passed is already compressed (instead of relying on the combination-of-parameters logic that it uses today). 
[...]

PR #1213, opened 2015-03-29
Make value of 'build:' relative to the yml file.

moysesb:
This fixes issue #1194 and is a prerequisite for implementing #1209

### DIFF
--- a/compose/config.py
+++ b/compose/config.py
@@ -171,6 +171,9 @@ def process_container_options(service_dict, working_dir=None):
     if 'volumes' in service_dict:
         service_dict['volumes'] = resolve_host_paths(service_dict['volumes'], working_dir=working_dir)
 
+    if 'build' in service_dict:
+        service_dict['build'] = resolve_build_path(service_dict['build'], working_dir=working_dir)
+
     return service_dict
 
 
@@ -328,6 +331,17 @@ def resolve_host_path(volume, working_dir):
         return "%s:%s" % (expand_path(working_dir, host_path), container_path)
     else:
         return container_path
+
+
+def resolve_build_path(build_path, working_dir=None):
+    if working_dir is None:
+        raise Exception("No working_dir passed to resolve_build_path")
+
+    _path = expand_path(working_dir, build_path)
+    if not os.path.exists(_path) or not os.access(_path, os.R_OK):
+        raise ConfigurationError("build path %s either does not exist or is not accessible." % _path)
+    else:
+        return _path
 
 
 def merge_volumes(base, override):
--- a/tests/unit/config_test.py
+++ b/tests/unit/config_test.py
@@ -383,3 +383,36 @@ class ExtendsTest(unittest.TestCase):
         ]
 
         self.assertEqual(set(dicts[0]['volumes']), set(paths))
+
+
+class BuildPathTest(unittest.TestCase):
+    def setUp(self):
+        self.abs_context_path = os.path.join(os.getcwd(), 'tests/fixtures/build-ctx')
+
+    def test_nonexistent_path(self):
+        options = {'build': 'nonexistent.path'}
+        self.assertRaises(
+            config.ConfigurationError,
+            lambda: config.make_service_dict('foo', options, 'tests/fixtures/build-path'),
+        )
+
+    def test_relative_path(self):
+        relative_build_path = '../build-ctx/'
+        service_dict = config.make_service_dict(
+            'relpath',
+            {'build': relative_build_path},
+            working_dir='tests/fixtures/build-path'
+        )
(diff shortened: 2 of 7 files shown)


compose/config.py
```diff
@@ -86,7 +86,8 @@ def make_service_dict(self, name, service_dict):
         service_dict['name'] = name
         service_dict = resolve_environment(service_dict, working_dir=self.working_dir)
         service_dict = self.resolve_extends(service_dict)
-        return process_container_options(service_dict, working_dir=self.working_dir)
+        service_dict = process_container_options(service_dict, working_dir=self.working_dir)
+        return resolve_build_path(service_dict, working_dir=self.working_dir)
```
aanand: Could we instead call `resolve_build_path` from `process_container_options`? I think that'd be conceptually cleaner.

tests/unit/config_test.py
```diff
@@ -40,6 +40,50 @@ def test_config_validation(self):
...
+    def test_nonexistent_path(self):
+        options = {'build': 'nonexistent.path'}
+        self.assertRaises(
+            config.ConfigurationError,
+            lambda: config.make_service_dict('foo', options, 'tests/fixtures/build-path'),
+        )
+
+    def test_relative_path(self):
+        relative_build_path = '../build-ctx/'
+        resolved_build_path = '/code/tests/fixtures/build-ctx'
```
aanand: This will probably fail when `nosetests` is run outside the testing container (which needs to be done when e.g. we're working on docker-py). Could you replace it with `os.path.abspath('tests/fixtures/build-ctx')`, here and in the other tests?

aanand:
Thanks so much! Looks really good. Only some small changes to make - see commit comments.

compose/config.py
moysesb: Will do. It sure looks better in `process_container_options`.

tests/unit/config_test.py
moysesb: So, in `resolved_build_path = '/code/tests/fixtures/build-ctx'` the `/code` part should be dynamic, but still absolute (the `resolve_build_path` in `config` puts an absolute path there). Is there a way for me to discover from the test what is the proper 'prefix' when the test is run outside the testing container? Would you say os.getcwd() is safe in this case, i.e. `resolved_build_path = os.path.join(os.getcwd(), 'tests/fixtures/build-ctx')`

aanand: That should work fine, but as far as I can tell it's equivalent to `os.path.abspath('tests/fixtures/build-ctx')`, which is simpler.

compose/config.py
```diff
@@ -311,6 +314,18 @@ def resolve_host_path(volume, working_dir):
         return container_path
 
 
+def resolve_build_path(build_path, working_dir=None):
+    if working_dir is None:
+        raise Exception("No working_dir passed to resolve_build_path")
+
+    _path = os.path.join(working_dir, build_path)
+    _path = os.path.normpath(os.path.abspath(_path))
+    if (not os.path.exists(_path) or not os.access(_path, os.R_OK)):
```
aanand: You can use the `expand_path` helper here, like `get_env_files` and `resolve_host_path` do.

aanand:
Almost there. Once you've done that last one, could you rebase and squash? Thanks.

moysesb:
All tests passed locally, the failure seems to be unrelated to my changes. Is there something I can do?

aanand:
Nope, was a Jenkins hiccup. Passing now.

LGTM

aanand:
Ah, needs a rebase thanks to #1226 - sorry!

moysesb:
Shifted `BuildPathTest` down to be the last class in the file, as it should be ;)

aanand:
Thanks. Could you rebase to the current master? I'm seeing confusing extra stuff in the tests diff.

moysesb:
My bad, I should have run the diff myself before pushing, should be good now.

aanand:
:star: LGTM :star:

aanand:
ping @dnephin - last bugfix for 1.2, so I'll roll another RC after this is merged!

compose/config.py
```diff
@@ -330,6 +333,17 @@ def resolve_host_path(volume, working_dir):
         return container_path
 
 
+def resolve_build_path(build_path, working_dir=None):
+    if working_dir is None:
+        raise Exception("No working_dir passed to resolve_build_path")
```
dnephin: if `working_dir` is required it really shouldn't be a `kwarg`, it should  be just a regular positional argument

dnephin:
Just a minor style nit, nothing that would block a merge.

LGTM

PR #1221, opened 2015-03-30
Update Swarm doc

### LINKED ISSUE #1208: Document that containers are scheduled onto the same node on Swarm
#972 made containers schedule onto the same node when running Compose apps on Swarm. We should probably document this in SWARM.md.

aanand:
- Co-scheduling will now work, so we can remove the stuff about `volumes_from` and `net` and manual affinity filters.

- Added a section about building.

Closes #1208.

### DIFF
--- a/SWARM.md
+++ b/SWARM.md
@@ -9,43 +9,24 @@ Still, Compose and Swarm can be useful in a “batch processing” scenario (whe
 
 A number of things need to happen before full integration is achieved, which are documented below.
 
-Re-deploying containers with `docker-compose up`
-------------------------------------------------
-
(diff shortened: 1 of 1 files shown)


SWARM.md
```diff
@@ -9,43 +9,24 @@ Still, Compose and Swarm can be useful in a “batch processing” scenario (whe
...
-  environment:
-    - "affinity:container==myproject_db_*"
-db:
-  image: postgres
-```
+Building
+--------
 
-Here, we express an affinity filter on all web containers, saying that each one must run alongside a container whose name begins with `myproject_db_`.
+`docker build` against a Swarm cluster is not implemented, so for now the `build` option will not work - you will need to manually build your service's image, push it somewhere and use `image` to instruct Compose to pull it. Here's an example using the Docker Hub:
```
bfirsh: Could be an opportunity here to link off to doc about how to do multiple environments with Compose. In dev, you can use `build: .`, in production you can use `image: myusername/web` and point at Swarm.

bfirsh:
LGTM

Issue #1167, opened 2015-03-24
Incomplete env. variable substitution in the "volumes" section


pavlo:
Hey guys, it looks like Compose does not handle env. variables substitution if there're two vars like shown below: 

    volumes:
      - ${APP_DIR_ON_HOST}:${APP_DIR_ON_CONTAINER}

in this case the first var (APP_DIR_ON_HOST) gets expanded just fine while the second one stays unchanged - it gets passed literally as a string `${APP_DIR_ON_CONTAINER}`

Shouldn't it expand the second one as well? I would expect that.

Please let me know if you want me to provide more detailed example.
P.

wallnerryan:
#dibs for docker boston, looking into this now

wallnerryan:
confirmed the issue

export DHOST=/tmp/
export DCONTAINER=/host/tmp/

```
mycontainer:
   image: ubuntu:12.04
   volumes:
      - ${DHOST}:${DCONTAINER}
   command: /bin/bash -c "while true; do echo hello world; sleep 1; done"
```

```
 [46] → docker exec -it compose_mycontainer_1 /bin/bash
root@90c7c4204386:/# ls
${DCONTAINER}  bin  boot  dev  etc  home  lib  lib64  media  mnt  opt  proc  root  run  sbin  selinux  srv  sys  tmp  usr  var
```

wallnerryan:
Fixed this, but haven't tested enough yet. will post branch soon
```
wallnr @ ~/Desktop/dockerparty/compose-examples
 [100] → cat compose.yml 
mycontainer:
   image: ubuntu:12.04
   volumes:
      - ${DHOST}:${DCONTAINER}
   command: /bin/bash -c "while true; do echo hello world; sleep 1; done"

wallnr @ ~/Desktop/dockerparty/compose-examples
 [101] → echo ${DHOST}
/tmp/

wallnr @ ~/Desktop/dockerparty/compose-examples
 [102] → echo ${DCONTAINER}
/host/tmp/

wallnr @ ~/Desktop/dockerparty/compose-examples
 [103] → docker-compose -f compose.yml up -d
Creating composeexamples_mycontainer_1...

wallnr @ ~/Desktop/dockerparty/compose-examples
 [104] → docker exec  -it composeexamples_mycontainer_1 /bin/bash
root@48483b48bfc0:/# ls host/tmp/
aberr  k5_skip  tce  tcloop
root@48483b48bfc0:/# 
```

[...]

PR #1223, opened 2015-03-30
Bugfix 1167

wallnerryan:
Aims to fix https://github.com/docker/compose/issues/1167

With this fix you can use

```
volumes:
  - ${APP_DIR_ON_HOST}:${APP_DIR_ON_CONTAINER}
```
Or
```
volumes:
  - $APP_DIR_ON_HOST:$APP_DIR_ON_CONTAINER

```

Or regular path names, only checks happen when using local env vars for volume identification


wallnerryan:
may need to squash those commits, rebases and merged and it duplicated some of the commit messages.

barraq:
:+1:

dnephin:
This is now handled by #1765

PR #1225, opened 2015-03-30
When extending, `build` replaces `image` and vice versa

### LINKED ISSUE #1222: When extending, `build` should replace `image` and vice versa
Currently, if a service with `build` is extended by a service that sets `image`, the resulting configuration will have both options, resulting in an error.

aanand:
Closes #1222.


dnephin:
LGTM

Issue #1228, opened 2015-03-31
building should support `docker build --file` option


iffy:
I want to mimic something like `docker build --file something.Dockerfile .` in `docker-compose.yml`  But the `--file` option is not supported.  From `docker build --help`:

```
-f, --file=""        Name of the Dockerfile(Default is 'Dockerfile' at context root)
```

I'm happy to attempt a patch.  It looks like a name needs to be chosen for the `docker-compose.yml` name but that docker-py supports this already: https://github.com/docker/docker-py/blob/master/docker/client.py#L304

dnephin:
Duplicate of #834

PR #1230, opened 2015-03-31
Add support for dockerfile when building

iffy:
This is my feeble attempt at fixing this.  The builds currently fail; I think because of an old version of docker-py.  The required docker-py support was added by this: https://github.com/docker/docker-py/commit/0713488faca9a473135021ad5425bf216ab54ba4 but I can't tell what version that is included in.

This is to fix issue #1228

### DIFF
--- a/compose/config.py
+++ b/compose/config.py
@@ -11,6 +11,7 @@ DOCKER_CONFIG_KEYS = [
     'detach',
     'dns',
     'dns_search',
+    'dockerfile',
     'domainname',
     'entrypoint',
     'env_file',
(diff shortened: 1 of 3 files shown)


iffy:
I didn't sign off the commit, apparently.  Well... I'll leave this fix to someone else who is in the groove with this project.

aanand:
Thanks for the contribution! We're getting close with https://github.com/docker/compose/pull/1075, but there's some low-hanging fruit in https://github.com/docker/compose/issues/754 if you feel like adding support for something else.

For future reference, to sign off a commit:

    git commit --signoff

To fix a commit that isn't signed off:

    git commit --amend --reuse-message=HEAD --signoff

To push your fixed commit if you've already pushed the old one:

    git push --force

iffy:
Oh, it's a known issue being worked on :)  That's good.  And thanks for the signoff tip -- I'll do that for the next one.  Would a programmatic future-proof way of doing #754 be possible?  For example:

```
run-opts: "-foo -bar -whatever=who"
```

aanand:
Afraid that wouldn't work, since we don't shell out to `docker run` - we talk directly to the remote API.

There could be a key for directly setting extra properties on the container JSON, but we'd be opening the door to a lot of misuse.

PR #1231, opened 2015-03-31
Test against Docker 1.6 RC2 only

aanand:
While Compose 1.2 will support Docker 1.3 and up, Compose 1.3 will introduce several features that rely on later versions - including labels, which were added in Docker 1.6.

We talked briefly about being able to specify an API version with an environment variable, but after some thought I don't think the combinatorial explosion of Docker versions and API versions is going to be worth the maintenance trouble.

Meanwhile, multiple PRs (#1011, #1075) can't be merged because they're waiting on us dropping support for older versions. Accordingly, I've updated the Dockerfile so that from now on we're only testing against Docker 1.6.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,18 +15,14 @@ RUN set -ex; \
     ; \
     rm -rf /var/lib/apt/lists/*
 
-ENV ALL_DOCKER_VERSIONS 1.3.3 1.4.1 1.5.0 1.6.0-rc2
+ENV ALL_DOCKER_VERSIONS 1.6.0-rc2
 
 RUN set -ex; \
-    for v in 1.3.3 1.4.1 1.5.0; do \
-        curl https://get.docker.com/builds/Linux/x86_64/docker-$v -o /usr/local/bin/docker-$v; \
-        chmod +x /usr/local/bin/docker-$v; \
-    done; \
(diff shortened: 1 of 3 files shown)


dnephin:
Once #1066 goes in the dependency on 1.6+ makes a lot of sense, so depending on it early seems fine to me.

It would be cool to still test against 2 versions, when there are multiple versions that are compatible, (ex 1.6 and 1.7 when it's available). It can help catch non-backwards compatible changes in docker releases. So we might still want #1157 again eventually, but we can talk about it when 1.7 comes around.

I agree the 3+ versions matrix is not worth it.

LGTM

thaJeztah:
For those that can't (or don't want to) update Docker, there's still the older releases; most new features will only be useful for Docker 1.6, so I think it makes sense.

Perhaps add a simple compatibility matrix table to the readme and/or Documentation, something like;

Compose/Fig release | Docker versions supported | Link
---- | --- | ---
Compose 1.2 | 1.6 | [release notes / download](https://github.com/docker/compose/releases/tag/1.2.0rc2)
Compose 1.1 | 1.3 - 1.5 | [release notes / download](https://github.com/docker/compose/releases/tag/1.1.0)

Issue #45, opened 2014-01-22
The project name should be configurable


bfirsh:
Ideally in `fig.yml` (how do we have general config in there? maybe we need a top level `services` key?), but maybe in a separate `.fig-project` file or something.

mikehaertl:
+1 

Any news on this? There was an interesting suggestion by @jderusse here, which i like very much:
https://github.com/docker/compose/issues/745#issuecomment-74028861

jderusse:
The problem with your proposals, is that you'r mixing project configuration and service declaration. it's less readable and adding a keyword (projet or .project) can break backward compatibility.

I realy thinks it's time for docker-compose to improve it schema by adding a top level structure wich contains services. 
```
containers:
  web:
    ...
  db:
    ...
```

With such structure, we'll be able to improve the schema like
```yaml
schema-version: 1.1
project_name: foo
containers:
  web:
    ...
  db:
    ...
```
or 
```yaml
schema-version: 2.0
project: 
  name: foo
  default_env_file: ./bar
containers:
  web:
    ...
  db:
    ...
```

Notice the property schema-version which help us keep backword compatibilities with schema evolutions.
And to keep BC with current schema, in compose/project.py::from_config
```
    @classmethod
    def from_config(cls, name, config, client):
        if 'schema-version' not in config or not isinstance(config['schema-version'], basestring)
            config = {
                'schema-version': '1.0',
                'containers': config
            }
        dicts = []
        for service_name, service in list(config.containers.items()):
            if not isinstance(service, dict):
                raise ConfigurationError('Service "%s" doesn\'t have any configuration options. All top level keys in your docker-compose.yml must map to a dictionary of configuration options.' % service_name)
            service['name'] = service_name
            dicts.append(service)
        return cls.from_dicts(name, dicts, client)
```

If I've time, i'll work on a patch proposal in the nexts days.

yegortokmakov:
I have two projects that are using docker compose, one dependent on the other. For this use-case it will be great to have something like dependent projects:

```yaml
schema-version: 2.0
project: 
  name: foo
  default_env_file: ./bar
require:
  project2:
    config: PATH | URL
    services:
      - serv1
      - serv2
    detached: true
containers:
  web:
    ...
  db:
    ...
```

I am posting it there because it seems like relative to #1233 
Does it makes sense? If yes, I will make it based on @jderusse code.

dnephin:
This sounds a lot like #318

phazei:
+1
[...]

PR #463, opened 2014-09-01
Configure project name from the config file

dnephin:
Resolve #45 and adds `.project` required for #318 and #210 

Still needs documentation, opening PR for discussion regarding the top level key to use for this config (currently `.project`).

PR #1233, opened 2015-03-31
Advanced config file

### LINKED ISSUE #210: Global env variables
I'd be great to be able to declare 'global' env variables, that are set for all containers.

jderusse:
This PR improve the config schema by moving the services under a new top level node `services`. 
```yaml
version: 1.0
services:
  db:
    image: postgres
  web:
    build: .
    command: python manage.py runserver 0.0.0.0:8000
    volumes:
      - .:/code
    ports:
      - "8000:8000"
    links:
      - db
```

Why ?

* Thank to the version property, the schema is extendable and we can keep Backward Compatibility
* We can add global parameters (like name, gobal env, global extends, ...)
* We can manage several versions of the schema (ie if a parameter is renamed in a future version).

What the difference with #463 ?

This PR does not mix service declaration and project configuration:

* More readable
* No collision between service's name and project key

I implement an usage of this new schema with an optional property `project` which replace the directory_name

```yaml
version: 1.1
project: foo
services:
  db:
    image: postgres
  web:
    build: .
```

This should fix tickets #45, #745
Then we could extends this schema to fix #210, #318

note: The script **does not** break backward compatibility :smiley: even if someone already use a service named `project` or `.project` or `version` or wathever

### DIFF
--- a/compose/cli/command.py
+++ b/compose/cli/command.py
@@ -71,14 +71,15 @@ class Command(DocoptCommand):
 
     def get_project(self, config_path, project_name=None, verbose=False):
         try:
+            config_dict = config.load(config_path)
             return Project.from_dicts(
-                self.get_project_name(config_path, project_name),
-                config.load(config_path),
+                self.get_project_name(config_dict, project_name),
+                config_dict,
                 self.get_client(verbose=verbose))
         except ConfigError as e:
             raise errors.UserError(six.text_type(e))
 
-    def get_project_name(self, config_path, project_name=None):
+    def get_project_name(self, config_dict, project_name=None):
         def normalize_name(name):
             return re.sub(r'[^a-z0-9]', '', name.lower())
 
@@ -90,9 +91,10 @@ class Command(DocoptCommand):
         if project_name is not None:
             return normalize_name(project_name)
 
-        project = os.path.basename(os.path.dirname(os.path.abspath(config_path)))
-        if project:
-            return normalize_name(project)
+        if config_dict:
+            project_name = config_dict['project']
+            if project_name:
+                return normalize_name(project_name)
 
         return 'default'
 
--- a/compose/config.py
+++ b/compose/config.py
@@ -1,6 +1,8 @@
 import os
 import yaml
 import six
+import compose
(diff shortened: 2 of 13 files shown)


csarrazi:
Bump

yosifkit:
Ping @aanand, @bfirsh, @dnephin, @mnowster: any word on whether this is acceptable?  Or reasons as to why not (besides the needed DCO sign-off)?

Having a project name in the yml (which is only part of this PR) would go a long way to help swarm integration since name conflicts occur often in multi-team environments.

aanand:
@yosifkit The discussion over whether or not the project name belongs in the config file is https://github.com/docker/compose/issues/745. Until that's resolved, no PR which implements such a change is going to be acceptable.

docteurklein:
I'd like to see a "services" entrypoint as well.
Not talking about the project name.
@aanand I think the project name being included or not is just a possibility.
Once you moved the services from global, you're free too add more keys at the root of the file,
and avoid clashes with future potential keys (like metadata, project name, …).
 
I'd like to add a "commands" key at the root of the file to enable a third party tool to read its config inside the compose yml file.

michaeljs1990:
Anyway this can be modified to support a / separator instead of _ optionally with a flag something like...

```
version: 1.1
project: foo
services:
  delimiter: /
  db:
    image: postgres
  web:
    build: .
```

would build services/db and services/web. If it could be made to do that this would be a super helpful pr that would fix many issues that I have to write custom scripts for right now.

dnephin:
Thanks for the contribution! Sorry it's taken a while to respond.  

Our focus so far on the config has been to improve the validation and error messages. In the upcoming 1.5.0 release we've got a new `jsonschema` for handling the validation and much improved error messages.

I think we'd like to add a version to the schema, which may happen as part of #2113.  If we're going to break backwards compatibility (or even just introduce a new format), we'd like to make a bunch of changes all at once (not just adding a version in isolation).

Sorry we weren't ready to make this breaking change when this PR was first submitted.  Our configuration code has changed significantly since then, so I don't think we'll be able to merge this PR.

PR #1244, opened 2015-04-02
Add support for every build context type supported by the daemon

moysesb:
This is a Proposal PR.

Implements #1209 (and more) using some of the ideas from https://github.com/docker/docker-py/issues/538

This changeset adds support for specifying the build path in any of the formats supported by docker-py's implementation of the build() operation. I expanded the section on `build:` in docs/yml.md with a brief explanation of what kind of values are supported. The implementation is mostly in a new file: `compose/context.py`.


Dockerfile
```diff
@@ -12,6 +12,7 @@ RUN set -ex; \
         curl \
         lxc \
         iptables \
+	liblzma-dev \
```
thaJeztah: Nit; indentation looks off

docs/yml.md
```diff
@@ -29,15 +29,44 @@ image: a4bc65fd
...
 ### build
 
-Path to a directory containing a Dockerfile. When the value supplied is a 
-relative path, it is interpreted as relative to the location of the yml file 
-itself. This directory is also the build context that is sent to the Docker daemon.
+Path to a context containing a Dockerfile. When the value supplied is a relative filesystem
+path, it is interpreted as relative to the location of the config file.
+
+The value supplied to this parameter can have any of the forms accepted by the docker remote
+api ```/build``` endpoint. Namely:
```
thaJeztah: Single back-ticks are sufficient for inline code in MarkDown (use triple back-ticks (or 4 spaces indentation) for code "blocks")

moysesb:
ping @aanand. Any word on whether this is being considered for merging? Thanks!

aanand:
Thanks for the PR! I'm theoretically in support of the feature, but:

- Why are we doing LZMA decompression client-side? Is that how the Docker CLI does it? I'm not relishing the bloat there.

- A lot of this logic looks like it belongs in docker-py. I see from https://github.com/docker/docker-py/issues/538 that it's non-trivial to overhaul `build` and `create_container` at this time, but the logic in `compose.context` could at least be moved to `docker.utils`.

- I deeply appreciate your including documentation. I don't think it needs to be that comprehensive, though - we should instead make sure the [docker build docs](http://docs.docker.com/reference/commandline/cli/#build) cover everything and link to them, and then show a short list of example values for `build`.

moysesb:
* The lzma thing is just a workaround for the fact that python2's tarfile module doesn't support the lzma format. The rationale is that compose should (perhaps) perform a preliminary sanity check on the archive before submitting it to the daemon, since the upload can be expensive if the file is large. The only way to make sure it's a valid archive on python2 is to include a check using the lzma module. Note that there's no actual decompression on the client side, just a peek to see if the tarfile/lzma module recognizes the archive format.

* I agree that most of this logic should be in docker-py. I'll wait for the maintainers to review 
https://github.com/docker/docker-py/issues/538 and then move `context.py` there.

moysesb:
@aanand  Moved `context.py` to docker-py and pushed a new commit here. I understand there's a discussion going on about how much validation should be performed by compose on the value of `build` (in #1372) - the implementation of `create_context_from_path()` in my docker-py fork does validate the supplied path, so maybe that should be taken into account.

Dockerfile
```diff
@@ -12,6 +12,7 @@ RUN set -ex; \
         curl \
         lxc \
         iptables \
+        liblzma-dev \
```
aanand: How widely installed is liblzma? We should make sure it exists on all the main already-supported Linux distros and OS X.

moysesb: Oh, this is a leftover from when context.py was in compose, there's no actual need for this dependency here. The lzma compression is handled as a special case in docker-py now: under py3 the tarfile module has built-in support for lzma (presuming the xz-utils headers where available when interpreter was built) but under py2 docker-py just accepts a file with an '.xz' extension as a lzma compressed tarball and passes it on to the build call without 'peeking' into it.

PR #1253, opened 2015-04-03
Add guide to using Compose in production

### LINKED ISSUE #944: Document how to use Compose in production
We should acknowledge the fact that it is possible to use Docker in production, but it is not recommended. We should also explain that we are working on improved support for production and link to the roadmap.

Options:

 - You can run your Compose app on a single server
 - You could use Compose to manage configuration for a single service (e.g. putting `fig up web` in your upstart script or similar)
 - You can point Compose at Swarm, but this integration is experimental and Swarm is in beta

/cc @fredlf

aanand:
Closes #944.

ping @fredlf

### DIFF
--- a/docs/mkdocs.yml
+++ b/docs/mkdocs.yml
@@ -1,5 +1,6 @@
 
 - ['compose/index.md', 'User Guide', 'Docker Compose' ]
+- ['compose/production.md', 'User Guide', 'Using Compose in production' ]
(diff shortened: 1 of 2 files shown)


fredlf:
I'm going to merge this and then edit it in a separate PR. 

FWIW, I feel like the docs LGTM rules for these projects are still not well defined or socialized. We should address that at some point.

PR #1258, opened 2015-04-06
Prepping for 1.6 release.

fredlf:
Adds release notes and edits/revises new Compose in production doc.


bfirsh:
LGTM

Might be worth stating up front that Compose only works on a single server. That isn't particularly clear, and is only implied right at the end when we say Swarm support is experimental. This can be fixed in a follow-up PR though.

PR #1260, opened 2015-04-07
Improve missing DOCKER_HOST stack traces

phuu:
Adds a more user-friendly message when DOCKER_HOST is not set.

One possible addition would be suggest fixes — perhaps `eval` the docker-machine env output — but that might be making too much of an assumption about the user's setup.

### DIFF
--- a/compose/cli/docker_client.py
+++ b/compose/cli/docker_client.py
@@ -1,5 +1,6 @@
 from docker import Client
 from docker import tls
+from ..service import ConfigError
 import ssl
 import os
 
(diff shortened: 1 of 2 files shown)


phuu:
Sorry for the broken build — can't run tests locally. Will try to rectify.

aanand:
Thanks!

I don't think having DOCKER_HOST not set is an error - it'll use the Unix socket in that case.

This might all go away if we just use docker-py's [kwargs_from_env](https://github.com/docker/docker-py/blob/1845d3b13fd7d964dea448fb82a29bde085c5d0f/docker/utils/utils.py#L279-L296) function.

phuu:
Oh ok — so this can be closed? I'm not very an advanced docker user but just thought fixing that rather hairy set of stack traces would help the next person who hits them.

dnephin:
Yes, I think we can close this. Not setting a `DOCKER_HOST` is just fine, it will use a default that is the default socket that docker will run on.

PR #1261, opened 2015-04-07
Fix vars in volume paths

### LINKED ISSUE #1255: volumes: absolute path becomes relative if it is in an environment variable
This used to work with ``fig``, but no longer seems to work in ``docker-compose==1.2.0rc3``:

```
myservice:
  volumes:
    - ${DOCKER_CERT_PATH}:/certs
```
``DOCKER_CERT_PATH`` contains an absolute path to a directory (e.g ``/Users/me/.boot2docker/certs/boot2docker-vm``).
Instead of mounting that, it will make it relative to the directory of the ``docker-compose.yml`` file.

Probably the code checks somewhere if it starts with ``/`` before expanding the variable, since setting it to ``- /${DOCKER_CERT_PATH}:/certs`` works.

aanand:
Closes #1255. Regressed with the `config.py` refactoring.

### DIFF
--- a/compose/config.py
+++ b/compose/config.py
@@ -328,6 +328,8 @@ def resolve_host_paths(volumes, working_dir=None):
 def resolve_host_path(volume, working_dir):
     container_path, host_path = split_volume(volume)
     if host_path is not None:
+        host_path = os.path.expanduser(host_path)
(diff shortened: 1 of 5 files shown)


bfirsh:
LGTM

PR #1293, opened 2015-04-15
1224: Check that image or build is specified.

mchasal:
Added the check to make sure that exactly one of the options is specified and a test case to check.
Also had to add an image option to a bunch of the existing test cases that started failing because of the new check.

Will anxiously await any review comments. 

Thanks


aanand:
LGTM

dnephin:
LGTM

dnephin:
Thanks for fixing this!

mchasal:
Thanks all. Happy to contribute and looking to do more in the near future.

PR #1299, opened 2015-04-16
Fix #1295 by single command string for install

### LINKED ISSUE #1295: Installation failure on curl command
Despite following the installation command, installation fails:
```
$ sudo curl -L https://github.com/docker/compose/releases/download/1.1.0/docker-compose-`uname -s`-`uname -m` > /usr/local/bin/docker-compose
-bash: /usr/local/bin/docker-compose: Permission denied
```
Platform:
```
Mac OS X 10.10.3 Yosemite
$ curl --version
curl 7.41.0 (x86_64-apple-darwin14.1.0) libcurl/7.41.0 OpenSSL/1.0.2a zlib/1.2.8 libidn/1.29
Protocols: dict file ftp ftps gopher http https imap imaps pop3 pop3s rtsp smb smbs smtp smtps telnet tftp 
Features: IDN IPv6 Largefile NTLM NTLM_WB SSL libz TLS-SRP UnixSockets 
```

ain:


docs/install.md
```diff
@@ -18,12 +18,9 @@ First, install Docker version 1.3 or greater:
...
 ### Install Compose
 
-To install Compose, run the following commands:
+To install Compose, run the following command:
 
-    curl -L https://github.com/docker/compose/releases/download/1.2.0/docker-compose-`uname -s`-`uname -m` > /usr/local/bin/docker-compose
-    chmod +x /usr/local/bin/docker-compose
-
-> Note: If you get a "Permission denied" error, your `/usr/local/bin` directory probably isn't writable and you'll need to install Compose as the superuser. Run `sudo -i`, then the two commands above, then `exit`.
+    sudo bash -c "curl -L https://github.com/docker/compose/releases/download/1.2.0/docker-compose-`uname -s`-`uname -m` > /usr/local/bin/docker-compose && chmod +x /usr/local/bin/docker-compose"
```
kojiromike: Many users (such as [Homebrew](http://brew.sh/) users) will have `/usr/local` owned by a non-root user, but this approach will result in the actual `docker-compose` file being owned by `root` regardless of the ownership of the rest of the directory. That can cause confusion down the line.

I acknowledge that software security is fully a PITA, but does Docker/Compose really want to encourage users to use `sudo` to install it by default? Users who don't want to understand these issues should simply use a package manager like `pip` or homebrew itself` to install `docker-compose`.

ain: I'm using MacPorts as many others who do not prefer do have their package
manager rely on OS and have clean separation. Same for users on Linux.

The only better way than the one proposed, could be the original Brew/Ports
package so one could install it from there. Or have RVM/NVM-like
installation requiring path update.


-- 
Sent from Gmail Mobile

kojiromike: Exactly. Users who want the easy, "it just works" solution should use whatever package manager they like. Users who want a custom install should:

1. Not need to be told about `sudo`.
2. Be given a sane default that doesn't require sudo, such as `pip` or `~/bin`.

ain: I'm not sure if you realise this, but the curl command given in the documentation doesn't even work with sudo. Just test.

aanand:
I'd rather not make people use `sudo` if it's not necessary.

mnowster:
Apologies for slow response on this PR. Since it was opened changes have been subsequently made to the install docs which cover installing without `sudo` and alternative install methods, eg `pip`.

https://github.com/docker/compose/blob/master/docs/install.md#alternative-install-options

PR #1302, opened 2015-04-17
Adds silent flag to install cURL command

iMerica:
The download progress isn't helpful to users installing Compose.

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -20,7 +20,7 @@ First, install Docker version 1.3 or greater:
 
 To install Compose, run the following commands:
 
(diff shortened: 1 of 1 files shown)


bfirsh:
LGTM

aanand:
I'm :-1: on this - I like having a progress indicator when downloading anything larger than a text file.

If anyone disagrees vehemently, feel free to make your case! Otherwise I'll close this.

dnephin:
Thanks for the PR. I agree with @aanand here, it's better to keep the progress information on by default.

iMerica:
Sure np.

PR #1320, opened 2015-04-21
Add memswap_limit support

sgirones:
Add memswap_limit support

Refs: https://github.com/docker/compose/issues/754

Signed-off-by: Salvador Girones <salvadorgirones@gmail.com>

### DIFF
--- a/compose/config.py
+++ b/compose/config.py
@@ -19,6 +19,7 @@ DOCKER_CONFIG_KEYS = [
     'image',
     'links',
     'mem_limit',
+    'memswap_limit',
     'net',
     'pid',
     'ports',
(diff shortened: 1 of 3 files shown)


sgirones:
If integration tests create real containers, I guess the host is not configured for MemorySwap limitation

aanand:
You could try writing a unit test instead, similar to the one in https://github.com/docker/compose/pull/1205

It's messy, but it'll motivate us to refactor that part ( ͡° ͜ʖ ͡°)

dnephin:
added in fc26982132b4a08f4efa2365a8cd668f94ff004e

Issue #1287, opened 2015-04-13
Allow ENV variables in docker-compose.yml


mattes:
docker-compose.yml
```
web:
  image: some-image
  volumes:
    - $PWD:/home/web
```

It would be really cool if ``env`` variables would get replaced with their actual values during ``docker-compose up``.

mattes:
ref: https://github.com/docker/compose/issues/939

hauleth:
:+1:

It will allow tools like `invoker` to run Composer apps in background.

pwaller:
Looks like a duplicate of #495

aanand:
Closing in favour of https://github.com/docker/compose/issues/1377

PR #685, opened 2014-11-30
Added support of cpuset and memswap_limit options in fig.yml.

arieb:
I added support for cpuset and memswap_limit options. unfortunately cpuset support was only added on docker-py 0.6, so i updated the setup.py and requirements.txt to use 0.6 (i saw there is already an open PR on this #659 ).

PR #76, opened 2014-02-04
Feature/env

SkinyMonkey:
This feature allow you to use environment variables in the fig.yml file.
It simply expand any found environment variable (ex: $HOST) to the value defined in the environment running fig.

useful example : 

I'm using Skydock (https://github.com/crosbymichael/skydock) along with skydns.

When i launch a skydns container by hand I need to specify the ip of the docker0 interface so that docker inject a new nameserver into /etc/resolv.conf:

ex:

Here the ip is "172.17.42.1"

`docker run -d -dns 172.17.42.1 -p 172.17.42.1:53:53/udp -name skydns crosbymichael/skydns -nameserver 8.8.8.8:53 -domain docker`

A nice way to be more flexible, compute what is the address automatically for example, is to add an environment variable in the equation : 

`env DNS_IP=172.17.42.1 docker run -d -dns $DNS_IP -p $DNS_IP:53:53/udp -name skydns crosbymichael/skydns -nameserver 8.8.8.8:53 -domain docker`

Wouldn't it be nice to be able to do the same thing in a fig.yml file? 

ex:

`env DNS_IP=172.17.42.1 fig up`

with this fig.yml:

```
skydns:
    image: crosbymichael/skydns
    ports:
        - $DNS_IP:53:53/udp
        - 8080:8080
    dns:
        - $DNS_IP
    command: "-nameserver 8.8.8.8:53 -domain docker"
```

I wanted to add a flag to make this expansion optional, just in case. What do you think?

PR #1338, opened 2015-04-24
inject ENV variables to yaml

mattes:
inject ENV variables to yaml

https://github.com/docker/compose/issues/1287

### DIFF
--- a/compose/config.py
+++ b/compose/config.py
@@ -407,9 +407,15 @@ def get_service_name_from_net(net_config):
 def load_yaml(filename):
     try:
         with open(filename, 'r') as fh:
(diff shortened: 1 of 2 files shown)


funkyfuture:
#685 does this as well and allows defaults.

dnephin:
#76 took a very similar approach. The problem with the approach is that it makes it really easy to corrupt your yaml, or even allows unexpected environment variables to add new sections, services or even commands to compose, which could be used maliciously.

I strongly believe environment substitution needs to happen on a per-fiend basis. It's more work, but it's much safer.

dnephin:
This is done in #1765

Issue #919, opened 2015-02-03
Intermediate Container Incorrectly Assumes /bin/echo Available


alunduil:
When using the etcd image from the CoreOS guys: quay.io/coreos/etcd, I get the following error from docker-compose:

```
Cannot start container a6da0c7e877b1075696d64802f7e159b7660e5cd395064b891ab5712c74bc266: exec: "/bin/echo": stat /bin/echo: no such file or directory
```

This is when starting the following datastore service from docker-compose.yml:

```
datastore:
  image: quay.io/coreos/etcd:latest
  ports:
   - "4001:4001"
```

I found https://github.com/docker/fig/blob/fd30920aac060b03d6e67ec5c3170f670a73d00e/compose/service.py#L288, which indicates that this assumption is hard coded and containers *must* provide /bin/echo in order to be usable by docker-compose.  Is there any way we can remove this assumption so a greater number of containers are usable?

tianon:
This is related to the discussion in https://github.com/docker/fig/pull/517 :+1:

bfirsh:
:+1: solution is https://github.com/docker/fig/issues/874 I believe

tristan0x:
I encountered the same issue while using minimalist [tianon/true](https://registry.hub.docker.com/u/tianon/true/) Docker image that only provides `/true`utility.
I am using Docker 1.5 and Docker Compose 1.1.0.
You can use the following YAML file to reproduce the issue

```yaml
data:
  image: tianon/true
  volumes:
    - /data
```

```sh
$ docker-compose up
Creating tmp_data_1...
Attaching to tmp_data_1
tmp_data_1 exited with code 0
Gracefully stopping... (press Ctrl+C again to force)

$ docker-compose up
Recreating tmp_data_1...
Cannot start container     3c29600d4043393936497d39c9c8c355f2427f4a34c86ea65d4ed3d315c50960: exec: "/bin/echo": stat /bin/echo: no such file or directory
```

As I want to keep data-only container as small as possible, my workaround is yet another [cogniteev/truecho](https://registry.hub.docker.com/u/cogniteev/truecho/) Docker image based on `tianon/true` that provides `/bin/echo` executable.

tj:
having this issue with nsqio/nsq as well

```
Client version: 1.5.0
Client API version: 1.17
Go version (client): go1.4.1
Git commit (client): a8a31ef
OS/Arch (client): darwin/amd64
Server version: 1.5.0
Server API version: 1.17
[...]

PR #1349, opened 2015-04-24
Rename container when recreating it

### LINKED ISSUE #874: Drop intermediate container and use renaming instead
With the merging of https://github.com/docker/docker/pull/8570, Docker 1.5 will support container renaming. This means we can stop creating an "intermediate container" when recreating containers - their only purpose is to free up the container name while preserving volumes.

This will stop Compose working on Docker <1.5, but I think it's worth it: as well as generally reducing complexity, it'll remove the problem that #516 and #517 attempt to solve.

dnephin:
Resolves #874 (and a bunch of related bug reports)

### DIFF
--- a/compose/service.py
+++ b/compose/service.py
@@ -30,6 +30,7 @@ DOCKER_START_KEYS = [
     'pid',
     'privileged',
     'restart',
+    'volumes_from',
 ]
 
 VALID_NAME_CHARS = '[a-zA-Z0-9]'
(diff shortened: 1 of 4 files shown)


compose/service.py
```diff
@@ -241,29 +245,15 @@ def recreate_container(self, container, **override_options):
...
-            detach=True,
-            host_config=create_host_config(volumes_from=[container.id]),
-        )
-        intermediate_container.start()
-        intermediate_container.wait()
-        container.remove()
-
-        options = dict(override_options)
+        # Use a hopefully unique container name by prepending the short id
+        self.client.rename(container.id, container.short_id + container.name)
```
dnephin: I think there is one possible failure case here. If compose crashes between this line , and the `remove()` line, the old container will be left with the temporary name.

The next time compose runs, it wont recreate (since the container has a different name now), and any volumes will be new instances.  The old volumes will still be attached to the temporary container, so they shouldn't be lost.

aanand: This is also possible right now, isn't it? If Compose crashes after creating the intermediate container but before creating the new one, it won't recreate.

It's not ideal, but at least it isn't a regression ^_^

aanand: Tiny thing, but right now this generates names like `c8a6039dcecounter_web_1`. Could we stick an underscore between the short ID and the name?

aanand:
Should close https://github.com/docker/compose/issues/919 too.

dnephin:
Rebased, and added the underscore to the renamed container name

aanand:
giphy

PR #1352, opened 2015-04-24
Use cool new IRCCloud links for IRC channel

bfirsh:
IRCCloud have a slick new experience for joining specific channels. This will ask you to sign up for IRCCloud, then put you straight into the channel.

@jonty <3

@fredlf Do you think we should add this to the docs page? Happy to go through other projects and add it too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ Installation and documentation
 ------------------------------
 
 - Full documentation is available on [Docker's website](http://docs.docker.com/compose/).
-- Hop into #docker-compose on Freenode if you have any questions.
+- If you have any questions, you can talk in real-time with other developers in the #docker-compose IRC channel on Freenode. [Click here to join using IRCCloud.](https://www.irccloud.com/invite?hostname=irc.freenode.net&channel=%23docker-compose)
 
 Contributing
 ------------


thaJeztah:
I *do* like IRCCloud, but wondering if Docker should use it as the default link (given that there are other options), perhaps as a suggestion (if you don't have an IRC client , there's IRCCloud)

bfirsh:
@thaJeztah mm. agreed. updated. what do you think?

thaJeztah:
Better :-)

(Hm, perhaps we should re-write the IRC quick start to use IRCCloud https://docs.docker.com/project/get-help/)

@moxiegirl wdyt on using IRCCloud in the examples and adding a link to all projects? Too opinionated?

Jonty:
Possibly worth mentioning that the landing page & widget also has an irc:// link to the channel for other IRC clients.

Here's the full list of options for linking to IRCCloud channels: https://jsfiddle.net/cfgo4g2d/7/embedded/result/

moxiegirl:
@thaJeztah Yes, I've been thinking of moving http://docs.docker.com/project/get-help/ to using IRCloud instead of webchat.  I just haven't had any bandwidth to update.

@bfirsh If you want to update that would be great. Here is an issue: https://github.com/docker/docker/issues/12854. 

If not, your current change LGTM.

aanand:
LGTM

thaJeztah:
@Jonty ah, thanks for listing those options

> the landing page & widget also has an irc:// link to the channel for other IRC clients.

Perhaps it *can* be used as the default link, I just wanted to be a bit careful; does Docker want to "advertise" IRCCloud as the preferred client (or not), or keep it generic.

Not against it, just didn't want it to be done without overthinking (I'm using IRCCloud myself)

Jonty:
We're talking about perhaps making the "using other IRC clients" section a lot better, to make sure people get the best experience regardless of their chosen client. Is there any information you'd specifically like included?

In other news, we've pushed all the badge/link types to a dedicated page with a tool for making them: https://www.irccloud.com/badges

thaJeztah:
> Is there any information you'd specifically like included?

I'm currently traveling, but will have a proper look later (@moxiegirl ?). Thanks for thinking with us though; really appreciated!

PR #1200, opened 2015-03-26
paulczar fixes plus example file

chanezon:
Signed-off-by: Patrick Chanezon <patlist@chanezon.com>
Includes @paulczar https://github.com/docker/compose/compare/master...paulczar:beginner_doc plus an example yml file at the top

PR #1353, opened 2015-04-25
Update index.md

justinwhite:
Removed a few lines to improve clarity and consistency of the documentation.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -15,36 +15,6 @@ be done to get it running.
 Compose is great for development environments, staging servers, and CI. We don't
 recommend that you use it in production yet.
 
-Using Compose is basically a three-step process.
-
-First, you define your app's environment with a `Dockerfile` so it can be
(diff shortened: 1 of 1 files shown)


aanand:
Thanks - good idea - but we've already got an overhaul to the front page waiting in https://github.com/docker/compose/pull/1200.

justinwhite:
Gotcha. Missed that one. I think my beef was mostly with the `ADD . /code` because `/code` is mounted as a volume a few lines below. Thought it'd help avoid confusion.

aanand:
Ah yes. We've got one open for that in https://github.com/docker/compose/pull/1186, but it's stalled. If the author doesn't have time to do the fixes we need, feel free to step in and finish it!

dnephin:
Looks like we've already made a similar change http://docs.docker.com/compose/

Thanks for the contribution!

PR #1269, opened 2015-04-08
Implement 'labels' option

aanand:
Using unstable versions of docker-py (for label support) and requests, because docker-py requires a newer version but [this fix](https://github.com/kennethreitz/requests/pull/2533) hasn't landed yet.

Sadly, even after that fix, the binary still isn't working:

```
+ dist/docker-compose-Linux-x86_64 --version
Traceback (most recent call last):
  File "<string>", line 2, in <module>
  File "/usr/local/lib/python2.7/dist-packages/PyInstaller/loader/pyi_importers.py", line 270, in load_module
    exec(bytecode, module.__dict__)
  File "/code/build/docker-compose/out00-PYZ.pyz/compose", line 2, in <module>
  File "/usr/local/lib/python2.7/dist-packages/PyInstaller/loader/pyi_importers.py", line 270, in load_module
    exec(bytecode, module.__dict__)
  File "/code/build/docker-compose/out00-PYZ.pyz/compose.service", line 11, in <module>
ImportError: No module named docker.errors
```

Might be another weird requests/pyinstaller bug - going to see if I can produce a minimal failing test case outside of Compose.

Issue #652, opened 2014-11-17
Support for explicit container names


prologic:
Hi, Before I submit a PR for this I wanted to gauge the reaction/feedback first.

See: https://github.com/therealprologic/fig/commit/1653955606881c9d9268cfcf8044cfde01128b1c

This adds optional explicit container names to fig. Use-case:

```
autodock:
    image: prologic/autodock
    name: autodock
    ports:
        - 1338:1338
    volumes:
        - /var/run/docker.sock:/var/run/docker.sock

hipache:
    image: hipache
    name: hipache
    ports:
        - 80:80
        - 443:443
```

hadim:
+1

aanand:
This would be good, particularly the sharing use-case. `fig scale` would need to exit with a helpful error.

Off the top of my head I'm not sure how much of an ordeal this would be to implement, but once labels are in Docker and we're no longer relying on names (https://github.com/docker/docker/pull/9882), it'll be a lot easier.

funkyfuture:
@therealprologic i haven't tried your commits yet (propably tomorrow), but is there anything i can help you with to make it PR-ready?

baptistedonaux:
+1

frankscholten:
+1

MrMMorris:
+1! 

I was trying to have one of our developers run compose and it broke because their root dir name is different than mine

thaJeztah:
Happy to inform that support for labels (https://github.com/docker/docker/pull/9882) was merged and will be included in the upcoming 1.6 release of Docker.

Also, I think there are currently a number of related issues in this area (https://github.com/docker/compose/issues/652, https://github.com/docker/compose/issues/869, https://github.com/docker/compose/issues/941 among others); perhaps the discussion should be focussed on the global proposal https://github.com/docker/compose/issues/1066 (but that's just a suggestion)

kcmerrill:
I've read and re-read the proposal but I don't understand how adding labels allows one to customize explicitly the container name? 

If anybody has any additional insight they'd be willing to share I'd love additional information. For me, explicit container names in docker-compose/fig are essential to using reverse proxies based on container names.

thaJeztah:
@kcmerrill Right now, Compose uses a naming convention (`projectname_servicename_instance`) in order to "identify" which containers belong to which project, because there are no other ways to do that.

Now that labels are possible, labels can be used to "identify" the containers, without having to use a special naming.

For example, by starting a container named `foobar` but adding some labels;

[...]

PR #1356, opened 2015-04-27
Use labels instead of names to identify containers

### LINKED ISSUE #1066: Specify project and service with container labels

dnephin:
Resolves #1066 

Should be compatible with #1269

Also includes:
* a fix for the integration test suite, it was leaving behind one_off containers (the first commit)
* a small refactor of `create_container` and `next_container_number()`. I moved the logging into `create_container()` instead of having it happen in all the callers, and changed it so that `find_next_number()` only happens once.

### DIFF
--- a/compose/container.py
+++ b/compose/container.py
@@ -4,6 +4,8 @@ from __future__ import absolute_import
 import six
 from functools import reduce
 
+from .const import LABEL_CONTAINER_NUMBER, LABEL_SERVICE
+
 
 class Container(object):
     """
@@ -58,14 +60,11 @@ class Container(object):
 
     @property
     def name_without_project(self):
-        return '_'.join(self.dictionary['Name'].split('_')[1:])
+        return '{0}_{1}'.format(self.labels.get(LABEL_SERVICE), self.number)
 
     @property
     def number(self):
-        try:
-            return int(self.name.split('_')[-1])
-        except ValueError:
-            return None
+        return int(self.labels.get(LABEL_CONTAINER_NUMBER) or 0)
 
     @property
     def ports(self):
@@ -159,6 +158,7 @@ class Container(object):
         self.has_been_inspected = True
         return self.dictionary
 
+    # TODO: only used by tests, move to test module
     def links(self):
         links = []
         for container in self.client.containers():
--- a/compose/container.py
+++ b/compose/container.py
@@ -64,7 +64,11 @@ class Container(object):
 
     @property
     def number(self):
-        return int(self.labels.get(LABEL_CONTAINER_NUMBER) or 0)
+        number = self.labels.get(LABEL_CONTAINER_NUMBER)
+        if not number:
+            raise ValueError("Container {0} does not have a {1} label".format(
+                self.short_id, LABEL_CONTAINER_NUMBER))
+        return int(number)
 
     @property
     def ports(self):
--- a/compose/service.py
+++ b/compose/service.py
@@ -10,8 +10,16 @@ import six
 from docker.errors import APIError
 from docker.utils import create_host_config, LogConfig
 
+from . import __version__
 from .config import DOCKER_CONFIG_KEYS, merge_environment
-from .container import Container, get_container_name
+from .const import (
+    LABEL_CONTAINER_NUMBER,
+    LABEL_ONE_OFF,
+    LABEL_PROJECT,
+    LABEL_SERVICE,
+    LABEL_VERSION,
+)
+from .container import Container
 from .progress_stream import stream_output, StreamOutputError
 
 log = logging.getLogger(__name__)
@@ -79,27 +87,17 @@ class Service(object):
 
     def containers(self, stopped=False, one_off=False):
         return [Container.from_ps(self.client, container)
-                for container in self.client.containers(all=stopped)
-                if self.has_container(container, one_off=one_off)]
-
-    def has_container(self, container, one_off=False):
-        """Return True if `container` was created to fulfill this service."""
-        name = get_container_name(container)
-        if not name or not is_valid_name(name, one_off):
-            return False
-        project, name, _number = parse_name(name)
-        return project == self.project and name == self.name
+                for container in self.client.containers(
+                    all=stopped,
+                    filters={'label': self.labels(one_off=one_off)})]
 
     def get_container(self, number=1):
         """Return a :class:`compose.container.Container` for this service. The
         container must be active, and match `number`.
         """
-        for container in self.client.containers():
-            if not self.has_container(container):
-                continue
-            _, _, container_number = parse_name(get_container_name(container))
-            if container_number == number:
-                return Container.from_ps(self.client, container)
+        labels = self.labels() + ['{0}={1}'.format(LABEL_CONTAINER_NUMBER, number)]
+        for container in self.client.containers(filters={'label': labels}):
+            return Container.from_ps(self.client, container)
 
         raise ValueError("No container found for %s_%s" % (self.name, number))
 
@@ -138,7 +136,6 @@ class Service(object):
         # Create enough containers
         containers = self.containers(stopped=True)
         while len(containers) < desired_num:
-            log.info("Creating %s..." % self._next_container_name(containers))
             containers.append(self.create_container(detach=True))
 
         running_containers = []
@@ -178,6 +175,7 @@ class Service(object):
                          insecure_registry=False,
                          do_build=True,
                          previous_container=None,
+                         number=None,
                          **override_options):
         """
         Create a container for this service. If the image doesn't exist, attempt to pull
@@ -185,6 +183,7 @@ class Service(object):
         """
         container_options = self._get_container_create_options(
             override_options,
+            number or self._next_container_number(one_off=one_off),
             one_off=one_off,
             previous_container=previous_container,
         )
@@ -209,7 +208,6 @@ class Service(object):
         """
         containers = self.containers(stopped=True)
         if not containers:
-            log.info("Creating %s..." % self._next_container_name(containers))
             container = self.create_container(
                 insecure_registry=insecure_registry,
                 do_build=do_build,
@@ -256,6 +254,7 @@ class Service(object):
         new_container = self.create_container(
             do_build=False,
             previous_container=container,
+            number=container.labels.get(LABEL_CONTAINER_NUMBER),
             **override_options)
         self.start_container(new_container)
         container.remove()
@@ -280,7 +279,6 @@ class Service(object):
         containers = self.containers(stopped=True)
 
         if not containers:
-            log.info("Creating %s..." % self._next_container_name(containers))
             new_container = self.create_container(
                 insecure_registry=insecure_registry,
                 detach=detach,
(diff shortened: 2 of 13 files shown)


compose/service.py
```diff
@@ -306,14 +302,19 @@ def get_net_name(self):
...
-        return '_'.join(bits + [str(self._next_container_number(all_containers))])
-
-    def _next_container_number(self, all_containers):
-        numbers = [parse_name(c.name).number for c in all_containers]
+    def get_container_name(self, number, one_off=False):
+        # TODO: Implement issue #652 here
+        return build_container_name(self.project, self.name, number, one_off)
+
+    # TODO: this would benefit from github.com/docker/docker/pull/11943
+    # to remove the need to inspect every container
```
thaJeztah: Not very familiar with python, but does this currently require to inspect *every* container, or the containers after filtering by project (and service name)? Trying to determine how important https://github.com/docker/docker/pull/11943 is for this.

dnephin: Only the containers after filtering, so it's not so bad

thaJeztah: Ah thanks, I was expecting that, but wasn't sure :smile:

aanand:
Fantastic stuff, thank you. Will merge once docker-py label support has landed.

aanand:
Will this be backwards-compatible with existing Compose projects? i.e. if I've got containers running that I started with the current stable version, then I upgrade and run `docker-compose up`, will it recreate or will it ignore them and start new ones?

thaJeztah:
Hm, perhaps interesting thought to add the version of compose that was used to create the containers as a label as well, in case there's some migration needed in future.

dnephin:
Ah, good point. I think that might be a problem. I suspect it will look for containers using labels, find none, then attempt to create them, running into a 409 conflict error from docker.

I think we'll need to do something to address that. 

We could release one version that falls back to trying to lookup by name, or
we could add labels to `create_container` in one version, and not actually use them to lookup containers until the next version.

Both of those run into the issue where it requires users to upgrade one version at a time though, skipping the intermediate version would still break things.

Another option, which is usually my preference, is to provide a migration script, which would read a `docker-compose.yml` and recreate the containers adding labels. We could document that the options are either: 
1. `docker-compose rm` using the old version, to get a fresh create
2. use the migration script to add labels to existing containers

@thaJeztah I think a compose version label would be a good idea

aanand:
Perhaps if `Project.containers` or `Service.containers` returns an empty list, we could then check for containers using the old method and prompt the user to upgrade (using a migration script/command) if we find any. Something like:

```python
def containers(self, *args, **kwargs):
    containers = self._containers_using_labels(*args, **kwargs)
    if not containers:
        self._check_for_old_containers()
    return containers

def _check_for_old_containers(self):
    containers = self._containers_using_names()
    if containers:
        names = [c.name for c in containers]
        raise UserError(
            """
            Found containers that were created using an old version of Compose:
            {}

            You have two options:
              - Migrate them with `docker-compose migrate-containers-1.3`
              - Remove them with `docker rm -f {}`
            """.format("\n".join(names), " ".join(names))
        )
```

aanand:
This would mean checking for containers would still be potentially slow when there aren't any in the current project, but compared to how slow it is now we're only adding a single API call (doing a `ps` with a label filter), which shouldn't be too costly if it's not going to find anything.

We can remove it in a later version once we're satisfied that most people have upgraded.

thaJeztah:
@aanand I like that idea. If compose had a configuration file (`~/.docker/compose.json`), it could be configurable and disabled if people don't need it anymore :)

(Docker itself is working on some changes in the config files / structure)

ahmetalpbalkan:
@dnephin liked the idea a lot. Does this PR remove service name entirely from the container name or still keep it around? I guess for readability of 'docker ps' output it might be nice to have them. How does the labels look like when containers are started from `compose`?

aanand:
We keep container names, but they're now "write-only" - we don't look at them, except to figure out what the next numeric suffix should be. (I think eventually we should drop numeric suffixes in favour of something random, so that generating the next one doesn't involve an API call).

Also, once this is merged, we'll be able to implement support for specifying custom container names.

For that reason, it might make sense to revisit the output of `docker-compose ps`. Since container names are no longer going to be consistently structured, there should probably be a "Service" column:

```
Service        Name                    Command                    State   Ports
----------------------------------------------------------------------------------------
redis          MyCoolContainerName     /usr/local/bin/run         Up
web            composetest_web_a1b2c3  /bin/sh -c python app.py   Up      5000->5000/tcp
```

> How do the labels on the containers look like when containers are started from `compose`?

Something like this:

```
com.docker.compose.project=composetest
com.docker.compose.service=web
com.docker.compose.container_number=1
com.docker.compose.oneoff=False
```

thaJeztah:
>except to figure out what the next numeric suffix should be

`MAX(com.docker.compose.container_number) + 1` ? Don't think the container-name is required for that?

>there should probably be a "Service" column

Or even a "Project" column; compose should now be able to get all containers that are managed by compose (filter by `--filter label=com.docker.compose.project` should probably work), so that even *all* services for *all* projects could be shown and (partly) managed.

aanand:
> Don't think the container-name is required for that?

True - we can just look at the labels - but it's still an extra API call versus generating something random.

> Or even a "Project" column

That might be cool: `docker-compose ps --all-projects` or something.

thaJeztah:
> That might be cool: docker-compose ps --all-projects or something.

Yes! Thinking if we should keep track of the location of the yaml file where the project was started from (the project "home" directory)

aanand:
Actually, yes, that's a good idea and one I meant to suggest earlier:

```
com.docker.compose.project.name=composetest
com.docker.compose.project.path=/Users/me/src/composetest
```

thaJeztah:
> com.docker.compose.project.path=/Users/me/src/composetest

Think that should be the location of the config (yaml), now that all paths are relative to that file (iirc), but yeah, something like that.

On a side-note, there's a breaking change being worked on in Docker currently. I'm personally not really happy about that. Not sure if compose *itself* is affected directly, but compose *users* definitely, if they use bind-mounted volumes to mount their source code during development. See:
https://github.com/docker/docker/pull/13025#issuecomment-99683333

thaJeztah:
> com.docker.compose.project.path=/Users/me/src/composetest

Think that should be the location of the config (yaml), now that all paths are relative to that file (iirc), but yeah, something like that.

On a side-note, there's a breaking change being worked on in Docker currently. I'm personally not really happy about that. Not sure if compose *itself* is affected directly, but compose *users* definitely, if they use bind-mounted volumes to mount their source code during development. See:
https://github.com/docker/docker/pull/13025#issuecomment-99683333

dnephin:
I'm thinking `com.docker.compose.config=/full/path/to/compose.yml`. It's possible to specify a different name for the compose file, so pointing at the path isn't really enough information, where as you can always get the root path used by the project by doing `dirname <config file>`.

dnephin:
We don't actually have access to the config path from `Service` right now, so I'm thinking we should wait to add `com.docker.compose.config`. It will be a larger change to pass the path around through Project, into the Service, and this change is already pretty large as it is. It's also not actually used for anything yet.

I've added the compose version, that was easy enough.

I'm going to focus on migration and backwards compatibility.

aanand:
:+1:

thaJeztah:
> this change is already pretty large as it is

Fully agree! one step at a time, no need to mingle multiple features into one.
The suggestions are nice for a future feature being added and don't relate to the Switch to using labels directly, just *depend* on it.

dnephin:
Ok, the latest commit adds a warning when no containers are found, and adds a `--migrate-to-labels` option which recreates containers with labels.  It's a bit weird right now because you still have to specify some command with `--migrate-to-labels`, because the parsing of args happens in a base class. Would like some way to improve that, but I wanted to get this first pass out for feedback first.

thaJeztah:
Sounds good; got a couple of minor nits, mainly personal taste, but I put them here anyway in case you like them;

- I think the label constants could be easier to recognize if the `_LABEL` suffix was turned into a prefix, e.g. `LABEL_CONTAINER_NUMBER`
- Perhaps the `com.docker.compose.container_number` should use a "dash" instead of an underscore; it's more in line with the "reversed domain name" namespace, as domain names / urls generally don't use an underscore
- Perhaps rename the `--migrate-to....` option to simply `--migrate`
- I spotted one `#TODO` in the tests; does that still apply?

great work, thanks so much!

aanand:
Looking fantastic. I'll echo all @thaJeztah's points, except that I think `--migrate-to-labels` is better - `--migrate` could mean any number of things, and we don't plan to keep the option around anyway.

- Could it be a command? `docker-compose migrate-to-labels`
- A test for migration would be good.

dnephin:
The migration could be a command, that would fix the weirdness, but it also seemed a bit weird to make it a full command, since it will be removed eventually.

I'll add a test as well.

The TODO in the tests should be fine for now, it'll be important for  #652 which is why I added it. I'll add the ticket number to the TODO.

dnephin:
Alright, migration test added, and appears to be passing. There was actually a bug in the previous version, it would try to add labels to all projects, instead of just the current one.

thaJeztah:
> but it also seemed a bit weird to make it a full command, since it will be removed eventually.

True, not a big issue I think, as long as the docs mention it's a temporary method. The migration is a nice extra and will only be used once for a project, so *"shrugs"*

Changes look good to me (but again, I'm not a python coder, so just a general "feel" by glancing over it). I think what's remaining is;

- Updates to the docs (a mention of the new approach, description of the labels, perhaps a how-to for manually creating containers to be included in a project)
- I wonder if we need tests for containers with an incomplete set of labels (thinking of this, if people manually create containers for a project - **do we support that?**)
- Updates to the shell-completion scripts where needed (can be another PR, and I'm sure @albers is willing to help out if needed :))

Again, great work!

aanand:
> thinking of this, if people manually create containers for a project - **do we support that?**

Definitely not.

thaJeztah:
> Definitely not.

:+1: definitely worth a mention in the docs. People **will** try, which is fine, but it should be clear that they're on their own in that case.

compose/container.py
```diff
@@ -54,14 +56,11 @@ def name(self):
...
-        return '_'.join(self.dictionary['Name'].split('_')[1:])
+        return '{0}_{1}'.format(self.labels.get(LABEL_SERVICE), self.number)
 
     @property
     def number(self):
-        try:
-            return int(self.name.split('_')[-1])
-        except ValueError:
-            return None
+        return int(self.labels.get(LABEL_CONTAINER_NUMBER) or 0)
```
aanand: Is it ever expected/OK for a container not to have a number label? I feel like we should error out if we don't find one, unless I've missed a case. Same for the service label.

aanand:
@moxiegirl @fredlf - your opinions on docs would be useful here. For context, we're making a change to Compose that means people will need to migrate their containers to a new format - but we're only going to keep it around for one version before dropping support for old containers.

Compose will show a warning if it detects old-style containers, prompt you to run `docker-compose migrate_to_labels`, and show the URL of a document that explains what's going on.

My question is: where should the document go? We don't plan to keep it around for a long time, so perhaps docs.docker.com isn't the right place and it should just go in a Markdown file on GitHub.

moxiegirl:
@aanand My recommendations would be:

* include an *Upgrade Compose* section in the compose release notes with instructions
* add an *Upgrade Compose* section to the installation docs (http://docs.docker.com/compose/install/) with also a mention

I would recommend the section in the release notes would only exist if we know there can be confusion. So, if no issues exist in the next release, the section goes away.  For the *Upgrade Compose* section in the install --- that would remain a stable touchpoint even if it has a single line to rerun the `curl` in the next release.

Unfortunately, I have no cycles this release to add the section to install, if you agree and want to add it, I'll have time to review.

aanand:
@moxiegirl Thanks - I'll do that. I've opened an issue for it: #1425

dnephin:
Rebased, and added an error to `Container.number`

aanand:
LGTM

thaJeztah:
Great job guys! Thanks

funkyfuture:
bravo! :dancers: :dancer: :dancers: 

(only a matter of gusto atm, but shouldn't anything that is used package-wide like the constants in `const.py` not rather be defined in `__init__.py`? so everything is accessible just by an `import compose`.)

Issue #1350, opened 2015-04-24
Using extends to replace "build" with "image" results in an error when build directory path does not exist


turtlemonvh:
# Reproducing the Error

Create 2 files.

    # compose_base.yml
    sleeper:
      build: /tmp/doesnotexist

    # compose_local.yml
    sleeper:
      extends:
        file: compose_base.yml
        service: sleeper
      image: ubuntu:12.04
      command: sleep 1000

Run 

    docker-compose -f compose_local.yml up -d

You will get the error:

    build path /tmp/doesnotexist either does not exist or is not accessible.

According to [the docs](http://docs.docker.com/compose/extends/#adding-and-overriding-configuration)

> In the case of build and image, using one in the local service causes Compose to discard the other, if it was defined in the original service.

Since that section is replaced, a reference to an invalid path should not result in an error.

# Diagnosis

The root of the problem is that [`other_loader.make_service_dict` is called before the configurations are merged](https://github.com/docker/compose/blob/4926f8aef629631032327542a56ae35099807005/compose/config.py#L112).  Part of `make_service_dict` is a call to [process_container_options](https://github.com/docker/compose/blob/4926f8aef629631032327542a56ae35099807005/compose/config.py#L162) which does a check on existence of the build directory.

I realize this is done this way to make recursive loading of parent configs easy, but I think that those checks are not necessary if the configuration being built is a configuration that is being extended from *and* that part of the configuration is being overwritten.

# Why its a problem

If I want to build an image on one machine, then just run the image I build on another machine (in production), I'd like to be able to use the same base config file in both environments.

With the current bug / implementation we have 2 options:

1. Create another compose.yml file just for building that inherits from base and defines build directories for each section. The base compose file would then have to define an `image` section because [one or the other is required](https://github.com/docker/compose/blob/master/compose/service.py#L63).
1. Create empty directories on the machine where docker-compose expects to find build directories

Both of these are pretty silly options.

turtlemonvh:
One possible solution would be to only resolve the host paths defined in the container options when on the final level of the inheritance stack.  That is, if you're looking at something you're extending, don't both resolving its paths.  This could be handled with a configuration flag.

If the community is open to this option, I'll submit a PR for this on Monday.

turtlemonvh:
Pull request: https://github.com/docker/compose/pull/1357

turtlemonvh:
This can be closed now that #1357 is merged.

ebmeierj:
[...]

PR #1357, opened 2015-04-27
Fix for #1350, nonexisting build path in parent section causes extending section to fail

turtlemonvh:
Now we can have `build` sections in a base yml file that don't exist if those sections are overridden with `image` or `build` sections in a child.

### DIFF
--- a/compose/config.py
+++ b/compose/config.py
@@ -64,6 +64,7 @@ def from_dictionary(dictionary, working_dir=None, filename=None):
             raise ConfigurationError('Service "%s" doesn\'t have any configuration options. All top level keys in your docker-compose.yml must map to a dictionary of configuration options.' % service_name)
         loader = ServiceLoader(working_dir=working_dir, filename=filename)
         service_dict = loader.make_service_dict(service_name, service_dict)
+        validate_paths(service_dict)
         service_dicts.append(service_dict)
 
     return service_dicts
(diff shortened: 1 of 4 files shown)


aanand:
Thanks for the fix, and the bug report! I think this is a bit of a messy solution though. It'd be better to just move all validation of paths (including those passed to `volumes`) to another method which we only call once all the `extends` stuff is done - so in `from_dictionary` we'd go:

```python
service_dict = loader.make_service_dict(service_name, service_dict)
validate_paths(service_dict)
service_dicts.append(service_dict)
```

turtlemonvh:
Thanks @aanand.  I agree the original was not beautiful - I was trying to change the control flow as little as possible.

Please let me know if the changes are what you were looking for.

tests/unit/config_test.py
```diff
@@ -407,7 +422,7 @@ def test_nonexistent_path(self):
         options = {'build': 'nonexistent.path'}
         self.assertRaises(
             config.ConfigurationError,
-            lambda: config.make_service_dict('foo', options, 'tests/fixtures/build-path'),
+            lambda: config.validate_paths(config.make_service_dict('foo', options, 'tests/fixtures/build-path')),
```
aanand: Could this lambda call `from_dictionary` instead of `valid_paths(make_service_dict)`? I'm worried about reaching too far into the internals - it makes the tests more stable, but could hide regressions if the code is restructured.

compose/config.py
```diff
@@ -170,10 +174,16 @@ def process_container_options(service_dict, working_dir=None):
...
+            service_dict['volumes'],
+            working_dir=working_dir
+        )
 
     if 'build' in service_dict:
-        service_dict['build'] = resolve_build_path(service_dict['build'], working_dir=working_dir)
+        service_dict['build'] = resolve_build_path(
+            service_dict['build'],
+            working_dir=working_dir
+        )
```
aanand: There are a bunch of unrelated formatting changes here - could you revert them?

aanand:
Thanks, looks better. Couple of comments inline.

aanand:
Looks great. Could you squash to one commit?

turtlemonvh:
Thanks again @aanand.  I made the changes you requested.  Let me know if you'd like me to squash all of that into one commit.

turtlemonvh:
Just saw your comment - squashing now.

turtlemonvh:
@aanand Squashed.

Thanks for the guidance on this!

aanand:
LGTM

dnephin:
LGTM

PR #1359, opened 2015-04-27
Remove DCO validation from CI script

aanand:
Now that @GordonTheTurtle is checking it, we don't need to do it in Jenkins.


dnephin:
LGTM

PR #1360, opened 2015-04-27
Remove wercker.yml

aanand:


dnephin:
LGTM

elhachmi:
Hi,
Can I know, why did you stop to use wercker, because I plan to use it for a private projects

PR #1371, opened 2015-04-29
Added new policy for assigning hostnames to container replicas

kikkomep:
Up to now replicas of a container (created using the docker-compose command 'scale')  have the same hostname of their 'master'.
This PR defines a new policy for assigning a proper hostname to container replicas according to the pattern: <master_hostname> + "_r" + <#replica_instance>.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,12 @@
 Change log
 ==========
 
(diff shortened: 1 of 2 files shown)


dnephin:
In many ways, this is trying to solve the same problem as #1131 I believe

Issue #1369, opened 2015-04-29
Support build urls


jonaseck2:
Docker supports both local and url build paths. docker-compose should do the same.

dnephin:
Duplicate of #1209 which has  some more discussion

PR #1372, opened 2015-04-29
Added support for url build paths

### LINKED ISSUE #491: Unable to fig build a github repo / works with docker build
Using `docker build` I can pull and build an image directly from Github, for example: 
```docker build -t java https://github.com/StrictlyBusiness/docker-java8.git```

When I try to do the same using `fig build` and the following `fig.yml`:
```
webapp:
  build: https://github.com/StrictlyBusiness/docker-java8.git
```

it shows `Building webapp...` then after ~40 seconds, it shows just a single closing bracket `)`.  For example:
```
$ fig build
Building webapp...
)
```

I haven't found a way to get any more details on what's happening behind the scenes in fig to know why it's failing.

jonaseck2:
Implements #1369 
Fixes #491

Shamelessly stole url validation logic from docker.

Manual testing works fine but I've been unable to run the unit tests due to a 404 error from docker when running script/test.

### DIFF
--- a/compose/config.py
+++ b/compose/config.py
@@ -57,6 +57,14 @@ DOCKER_CONFIG_HINTS = {
     'workdir': 'working_dir',
 }
 
+DOCKER_VALID_URL_PREFIXES = (
+    'http://',
+    'https://',
+    'git://',
+    'github.com/',
+    'git@',
+)
+
 
 def load(filename):
     working_dir = os.path.dirname(filename)
@@ -356,14 +364,21 @@ def resolve_host_path(volume, working_dir):
 def resolve_build_path(build_path, working_dir=None):
     if working_dir is None:
         raise Exception("No working_dir passed to resolve_build_path")
-    return expand_path(working_dir, build_path)
+    if is_url(build_path):
+        return build_path
+    else:
+        return expand_path(working_dir, build_path)
 
 
 def validate_paths(service_dict):
     if 'build' in service_dict:
         build_path = service_dict['build']
-        if not os.path.exists(build_path) or not os.access(build_path, os.R_OK):
-            raise ConfigurationError("build path %s either does not exist or is not accessible." % build_path)
+        if (not os.path.exists(build_path) or not os.access(build_path, os.R_OK)) and not is_url(build_path):
+            raise ConfigurationError("build path %s either does not exist, is not accessible or is not a valid url." % build_path)
+
+
+def is_url(build_path):
+    return build_path.startswith(DOCKER_VALID_URL_PREFIXES)
 
 
 def merge_volumes(base, override):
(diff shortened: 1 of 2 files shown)


aanand:
Nice. I think that path validation logic belongs in docker-py, though.

funkyfuture:
i'm afraid it belongs to the config-validation. in #1355 the validation-strategy is to accept nothing that is not defined as being allowed.

jonaseck2:
I'll implement it as an UrlValidator instead

jonaseck2:
Would this fix be a candidate for a 1.2.1 release? I'm unable to use 1.2.0 due to the broken url support.

aanand:
@funkyfuture Nonetheless, it's not specific to Compose, and should at least be a utility function under `docker.utils`.

@jonaseck2 Is it a regression? I don't think Compose ever supported build URLs.

jonaseck2:
1.1.0 works like a charm with https://github.com/jonaseck2/freezing-wookie/blob/master/docker-compose.yml.

jonaseck2:
Before 1.2.0 I'm assuming that compose just transparently forwarded the build path to docker and let docker validate the build path. In that sense, adding validation of the build path only limits what docker compose allows itself to pass to docker. I think it's a better idea to separate concerns between docker and compose and leave validation of argument values to docker.

Initially, my solution to the issue was to check if the build path was a file path by testing if the relative path didn't exist and the expanded path did, otherwise just send it to docker as is.

Perhaps this is a better solution after all, both specifically and for compose in general?

funkyfuture:
> Nonetheless, it's not specific to Compose, and should at least be a utility function under docker.utils.

i didn't meant to deny that. but that it's crucial to do a rehearsal of a composition before the the band goes to stage. i'm kind of traumatized by the expierence of data-loss due to a misconfiguration.

so, my remark was like a loud thinking to remind me. in the meantime i added an url-validation to the buildpath-check in #1355.

aanand:
> I think it's a better idea to separate concerns between docker and compose and leave validation of argument values to docker.

*Some* validation is going to have to happen client-side, because the client needs to know whether to create a tarball or not. But docker-py is already doing that.

What if we just remove the build path validation code?

jonaseck2:
Yes, and just leave enough logic to figure out if it's a file path and apply expansion or not

dnephin:
LGTM

aanand:
Could we get a unit test for invalid URLs?

dnephin:
I believe this overlaps with #1244 which has docs and a test case

dnephin:
An integration test and some docs (which could reference  https://docs.docker.com/reference/commandline/build/) would be great.

Will need a rebase as well to update with master

dnephin:
I'd like to get this into the 1.5.1 release. If you don't have any time to update it, let me know, and I will rebase it and add an integration test.

jonaseck2:
Please do. I've been trying to get this done for quite some time now and I haven't found the time for it.

dnephin:
Rebased in #2430

Issue #1346, opened 2015-04-24
Support new Docker networking idioms


aanand:
Docker 1.7 will introduce the top-level concept of a "network", accompanied by new CLI and API functionality. Compose should add support for networks, which might involve one or both of:

- putting all containers in a project-wide network
- adding a `network` key so that users can enable more specific inter-service communication

aanand:
[Pinging @dave-tucker and @mavenugo in case I've made any basic errors here.]

Here's the shortest possible explanation of how networks will be used in the Docker CLI:

```
$ docker run -d --name c1 ...
$ docker run -d --name c2 ...
$ docker networks create mynetwork
$ docker networks attach mynetwork c1 foo
$ docker networks attach mynetwork c2 bar
```

`c1` and `c2` can now communicate with one another within the network `mynetwork`. Each has an IP address in `mynetwork`, and each one is addressable by its given hostname (`foo`, `bar`) by other containers in `mynetwork`.

Behind the scenes, a network can be implemented by one of many drivers. There'll be a basic driver enabling basic single-host inter-container communication, but people can implement more capable drivers which can provide cross-host communication (as well as, presumably, all sorts of other cool stuff).

I see two obvious approaches we could take to networking UX in Compose:

1.  **Create a default per-project network, and put all containers in it under the service name**

    This is cool because it involves no new stuff to learn, and in fact users now have *less* to know about - you delete your `links` entry and everything will still work:

    ```yaml
    web:
      build: .
      ports: ["80:8000"]
    redis:
      image: redis
    ```

    Behind the scenes, when you run `docker-compose up`, Compose will:

    - Create a network, presumably named after the project
    - Attach the `web` container to the network with the name `web`
    - Attach the `redis` container to the network with the name `redis`

    Then the web container can connect to the hostname `redis`, just as with links.

2.  **Require users to explicitly specify networks**

    Here, you'd have a `networks` key for explicitly putting containers in networks, and anything not explicitly attached would be isolated:

    ```yaml
    web:
      build: .
      ports: ["80:8000"]
      networks:
        - mynetwork
    redis:
      image: redis
      networks:
        - mynetwork
    redisbackup:
      build: ./redis-backup
      volumes_from: redis
      # not attached to any networks
    ```

    You'd also (presumably) be able to specify explicit hostnames to join as:

    ```yaml
    networks:
      - mynetwork:redis
      - othernetwork:persistence
    ```

Here are some things to think about:

[...]

PR #1375, opened 2015-04-30
Share /etc/hosts option

kikkomep:
This PR adds a new option ```--share-etc-hosts``` to ```up``` and ```scale``` commands allowing hostname resolution between containers.

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -12,6 +12,7 @@ import dockerpty
 
 from .. import __version__
 from ..project import NoSuchService, ConfigurationError
+from ..service import update_etc_hosts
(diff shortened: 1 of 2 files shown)


aanand:
Thanks for the PR. It's an interesting idea, but hostname resolution is what links (and the [upcoming networking idioms](https://github.com/docker/compose/issues/1346)) are for. This looks like a bit of a hack. What's the use case?

kikkomep:
The problem using ```links``` is that they do not support cyclic references between containers but I see that the [upcoming network idioms](#1346) do that.

bfirsh:
Thanks for the contribution @kikkomep, but this will indeed work once Docker's new networking is supported. See #1346. :)

PR #1381, opened 2015-05-01
Show proper command in help text of build subcommand

sherter:
The help text of the build subcommand suggested to use 'compose build' (instead of 'docker-compose build') to rebuild images.


aanand:
LGTM

dnephin:
LGTM

Issue #842, opened 2015-01-13
Add `security-opt` option for fig.yml


sigelinde:
Docker 1.3+ support `--security-opt` option for `docker run`, but fig is not.
http://blog.docker.com/2014/10/docker-1-3-signed-images-process-injection-security-options-mac-shared-directories/

lsowen:
Does anyone know if there are any plans to support this?

lsowen:
Looks like most of the heavy lifting was already completed in docker-py (docker/docker-py@1845d3b13fd7d964dea448fb82a29bde085c5d0f).  I've implemented the few changes in docker-compose required to support it in my fork: https://github.com/lsowen/compose

thaJeztah:
@lsowen if you have implemented this, perhaps you can create a pull-request?

lsowen:
@thaJeztah It currently relies on a development version of docker-py, so my plan was to wait until it had been released on pypi.  Should I go ahead and create a pull request before then?

thaJeztah:
@lsowen I'm not a maintainer here, but you can create a PR with the changes, mentioning that it relies on a newer version of `docker-py`, that hasn't been released yet. (Or something similar :))

lsowen:
Looks like docker-py v1.2.0 was just released, so I'm planning on updating/squashing and creating a PR.

lsowen:
Just created #1382.

lsowen:
This can be closed now because #1382 has been merged.

PR #1382, opened 2015-05-02
Add security_opt as a docker-compose.yml option

lsowen:
Addresses #842 by exposes the security_opt support recently implemented by docker-py in version 1.2.1.

### DIFF
--- a/compose/config.py
+++ b/compose/config.py
@@ -30,6 +30,7 @@ DOCKER_CONFIG_KEYS = [
     'ports',
     'privileged',
     'restart',
+    'security_opt',
     'stdin_open',
     'tty',
     'user',
--- a/compose/service.py
+++ b/compose/service.py
@@ -42,6 +42,7 @@ DOCKER_START_KEYS = [
     'privileged',
     'restart',
     'volumes_from',
+    'security_opt',
 ]
 
 VALID_NAME_CHARS = '[a-zA-Z0-9]'
@@ -595,6 +596,7 @@ class Service(object):
         cap_drop = options.get('cap_drop', None)
         log_config = LogConfig(type=options.get('log_driver', 'json-file'))
         pid = options.get('pid', None)
+        security_opt = options.get('security_opt', None)
 
         dns = options.get('dns', None)
         if isinstance(dns, six.string_types):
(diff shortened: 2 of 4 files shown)


thaJeztah:
Thanks!

I think this needs some changes to the documentation as well (and a simple example)

lsowen:
@thaJeztah , good feedback.  I will work on adding that this weekend.

aanand:
Thanks!

1. Was `security_opt` really only added in 1.2.1? I'm in favour of upgrading of course, but if it's not relevant to the feature it'd be better to do it in a separate PR. `requirements.txt` would also need updating.

2. Could you add an integration test?

3. Documentation would be great, as @thaJeztah said.

lsowen:
@aanand I am planning to add documentation.  And unfortunately, it was only added in v1.2.1 (added in commit https://github.com/docker/docker-py/commit/1845d3b13fd7d964dea448fb82a29bde085c5d0f).

lsowen:
@aanand Any suggestions or pointers on how to create some portable tests?  This applies selinux users/roles/types, and most default ones I don't think will let docker even start the instance.  I created a custom one for my use case, but to use it you have to compile/install it, which doesn't seem like a viable option.

aanand:
I don't know much about selinux, but you could have a look at the Docker Engine tests to see if they did anything clever (or ask in #docker-dev on Freenode).

Alternatively you can take the approach we use in some unit tests, e.g. [this one](https://github.com/docker/compose/blob/89789c54add776d3a9262ed4b4f7a682f1cbdb42/tests/unit/cli_test.py#L139-L186) - mock out the `docker.Client` object and inspect what arguments it was called with to make sure the `security_opt` config made it through.

It's pretty messy (we need to refactor that CLI code), but it works.

lsowen:
Looks like in docker they test that the passed in option is parsed correctly (https://github.com/docker/docker/blob/e6ec703df31f04c3ea3a6aeb19881494a6ae58ae/daemon/daemon_unit_test.go).

And I guess we don't really care in our testing if the security_opt was applied correctly (as we can assume proper behavior by docker itself), I think I will go down the route of making sure that the security_opt gets into the `docker.Client` correctly.

JonLundy:
+1 on adding this into docker-compose!

aanand:
Be good to get this in. Just needs:

- rebase
- unit test
- now that we're on docker-py 1.2.2 in master, the changes to requirements.txt can be removed

lsowen:
@aanand I finally added tests around the new behavior.  Please let me know if you think there is anything else that needs to be accomplished to get this merged.  Thanks!

aanand:
LGTM

bfirsh:
LGTM

Thanks @lsowen!

Issue #1303, opened 2015-04-17
Support `--log-driver`


colinrymer:
Docker 1.6 adds support for the `--log-driver` option. It'd be great to be able to set this in the `docker-compose.yml` file.

chernjie:
+1

just bump into this problem

doomrobo:
+1
Looks like compose and docker-py will have to be updated to support the --log-driver

pavlo:
+1

cmoro-deusto:
Just upgraded to docker v1.6 and happily migrated all my oldishe bash start/stop scripts to compose only to find that ```--log-driver``` is not yet supported. This would be great. 

If you need help beta testing this just let me know. 

Thanks a lot for you awesome work!

StyleEX:
+1

chernjie:
keep a watch out for https://github.com/docker/docker-py/issues/547

cmoro-deusto:
thanks @chernjie

lwcolton:
It looks like https://github.com/docker/docker-py/issues/547 has been resolved.  +1 to adding log driver support to docker compose.

cyppan:
+1

strootman:
+1

saschaarthur:
well im ultra bad at python i tried to start with:

--- a/compose/config.py
+++ b/compose/config.py
@@ -21,6 +21,7 @@ DOCKER_CONFIG_KEYS = [
     'image',
     'labels',
     'links',
+    'log-driver',
     'mem_limit',
     'net',
     'pid',

--- a/compose/service.py
+++ b/compose/service.py
@@ -25,6 +25,7 @@ DOCKER_START_KEYS = [
     'env_file',
     'extra_hosts',
     'net',
+    'log-driver',
     'pid',
     'privileged',
     'restart',

now compose accepts "log-driver" argument in *.yml but it seems its not passing it, can anybody verify / help out? :)

vdemeester:
@saschaarthur This would look like this : https://github.com/vdemeester/compose/commit/e324022980fb0ce927fc7b22c63cfaab1e6ac762

I'm looking at how to test it (unit & integration tests)

vdemeester:
The other questions is : With ``-log-driver`` it's possible to pass out config (with a set of *key/value*). How would it translate in a ``docker-compose.yml`` file. Something like that would do ?

```yaml
web:
    image: busybox
    log_config: 
        type: json-file
        configs:
            - key1: val1
```

[...]

PR #1388, opened 2015-05-05
Add support for log-driver as a docker-compose.yml option

vdemeester:
Discussed in #1303, as ``docker-py`` version 1.2.1 support this, we can add it as an option for ``docker-compose.yml`` files.

It's kept simple for now, the config files would look like :

```yaml
web1:
    image: busybox
    log_driver: syslog
web2:
    image: busybox
    log_driver: json-file
web3:
    image: busybox
    log_driver: none
```

The choice of ``log_driver`` is to keep it coherent with the ``--log-driver`` option of ``docker run`` (even though it's ``LogConfig`` in the API).

### DIFF
--- a/docs/yml.md
+++ b/docs/yml.md
@@ -271,6 +271,20 @@ labels:
   - "com.example.label-with-empty-value"
 ```
 
+### log driver
+
+Specify a logging driver for the service's containers, as with the ``--log-driver`` option for docker run ([documented here](http://docs.docker.com/reference/run/#logging-drivers-log-driver)).
+
+Allowed values are currently ``json-file``, ``syslog`` and ``none``. The list will change over time as more drivers are added to the Docker engine.
+
+The default value is json-file.
+
+```
+log_driver: "json-file"
+log_driver: "syslog"
+log_driver: "none"
+```
+
 ### net
 
 Networking mode. Use the same values as the docker client `--net` parameter.
--- a/compose/config.py
+++ b/compose/config.py
@@ -23,6 +23,7 @@ DOCKER_CONFIG_KEYS = [
     'links',
     'mem_limit',
     'net',
+    'log_driver',
     'pid',
     'ports',
     'privileged',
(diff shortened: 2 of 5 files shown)


docs/yml.md
```diff
@@ -271,6 +271,16 @@ labels:
   - "com.example.label-with-empty-value"
 ```
 
+### log drivers
+
+Set the log driver to use for the container, same as the ``--log-driver=""`` option of ``docker run``. The possible value are currently : ``json-file``, ``syslog`` and ``none``. This list will elvolve with the addition of the driver support in the driver engine. The default value is the same as for docker, it is ``json-file``.
```
aanand: This reads a bit confusingly. How about:

> Specify a logging driver for the service's containers, as with the `--log-driver` option for `docker run` ([documented here](http://docs.docker.com/reference/run/#logging-drivers-log-driver)).
>
> Allowed values are currently `json-file`, `syslog` and `none`. The list will change over time as more drivers are added to the Docker engine.
>
> The default value is `json-file`.

aanand: The header for this section should read "log_driver" too.

aanand:
Thanks! Minor documentation comments; otherwise, looks great.

docs/yml.md
vdemeester: Yeah, sounds way better :) I'll update it according to it.

vdemeester:
@aanand took care of your comments ;)

aanand:
Thanks!

LGTM

dnephin:
LGTM

nick-allen:
Is `extends` going to acknowledge the `log_driver` setting?

aanand:
@nick-allen `log_driver` should get passed through when you extend a service, yes.

flower-pot:
From what I can tell, this does not cover setting `--log-opt` correct? Are there plans to allow doing so?

vdemeester:
@flower-pot it's probably already there  https://github.com/docker/compose/pull/1665 (for the next release 1.4.0)

PR #1390, opened 2015-05-06
docker-compose down

MathewT:
This PR adds docker-compose down as a simple alias for docker-compose stop.     docker-compose down seems a bit more intuitive for many than docker-compose stop.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,11 @@
 Change log
 ==========
 
+1.3.0 (2015-04-16)
+------------------
+
+- `docker-compose` now supports `down` option, which behaves exactly the same as the `stop` option.
+
 1.2.0 (2015-04-16)
 ------------------
 
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -94,6 +94,7 @@ class TopLevelCommand(Command):
       scale     Set number of containers for a service
       start     Start services
       stop      Stop services
+      down      Down services (same as stop but symmetrical to up)
       up        Create and start containers
 
     """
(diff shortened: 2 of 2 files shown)


MathewT:
My apologies for not signing the commit per the rules.  I executed the instructions above.  Please let me know if anything else is required.

rvion:
do you really think an alias is a good idea ?
in my opinion, it only adds confusion.
There is no evidence that it's more intuitive.
Also, all past tutorials will mention stop, so It might confuse the reader about why there is 2 commands. 
People will think the 2 commands are different.
I would -1 this feature

Can't you just add an alias in your personnal config ?

rvion:
Is there some "authority" checking those PR to assure the coherence of the whole ?

keiththomps:
:+1: The opposite of `up` is `down`. Seems more intuitive to me.

dnephin:
Thanks for the PR!

Instead of duplicating the `stop` command, could you try this:

1) rename `stop` to `down`
2) alias `stop` to `down`

```python

    def down(...):
        ...

    stop = down
````

3) I wonder if `docopt` will support an alias, instead of a second line:

```
   down, stop     Stop services

```

Once that's done, please squash your commits to a single commit.

We haven't discussed this much, but I think it probably doesn't hurt to have an alias if users find it more intuitive.

dnephin:
I've been thinking about `stop` vs `down`, and I think `down`and `stop` are actually different.  Maybe we should implement a `down`, but I don't think it will be an alias of `stop` after all. I think `down` would also remove containers.

Let's continue the discussion in #2277

I'm going to close this PR for now, and we can revisit once the we've figured out the correct behaviour.

PR #1396, opened 2015-05-07
Fix markdown formatting issue

albers:
A missing blank line caused a heading not to be recognized as such, see https://docs.docker.com/compose/completion/

### DIFF
--- a/docs/completion.md
+++ b/docs/completion.md
@@ -3,14 +3,12 @@ layout: default
 title: Command Completion
 ---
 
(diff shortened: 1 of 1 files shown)


aanand:
Thanks. It looks like those `-----` headings should be `##` headings too - would you mind fixing that?

albers:
@aanand Updated.

aanand:
LGTM

dnephin:
LGTM

Issue #1395, opened 2015-05-07
Perform actions in parallel


iMerica:
Hey guys,

I just want to gauge interest in this before I start working on it. Is this something others would find useful?

My project's yml file has grown so big that `docker-compose up -d` and `docker-compose stop` take about a minute so this would be super awesome if once the parent containers are started/stopped, the child containers would start/stop asynchronously.

aanand:
This makes a lot of sense, yes. A lot of stuff Compose does could be parallelised. Off the top of my head:

1. Container creation is order-dependent, but chunks of it could still be done in parallel - e.g. if both `web` and `worker` depend on `db`, then once `db` is created, `web` and `worker` can be created in parallel.

2. Container starting is also order-dependent in the same way.

3. Container stopping/killing should be completely order-independent - if it's all going down, it doesn't matter what order we do it in.

4. Container deletion should also be completely order-independent.

cmoro-deusto:
+1

bfirsh:
Replaced by https://github.com/docker/compose/issues/1652

PR #1399, opened 2015-05-08
Only recreate what's changed

### LINKED ISSUE #1345: Only recreate containers whose configuration needs updating
Currently, `docker-compose up` deletes and recreates every container, so as to make sure any changes to `docker-compose.yml` have been picked up. Instead, it should inspect container configuration to see if they *need* recreating, and do the minimum necessary.

aanand:
Closes #1345. To try it out:

    $ docker-compose up -d --x-smart-recreate

Here's what needs fixing:

- [x] If the image tag (whether arrived at via `image` or `build`) has changed, this should trigger a recreate.
- [x] When a service is recreated, dependent services via `links`, `volumes_from` and `net:container` should also be recreated, whether or not their configuration has changed.
- [x] If a container is up-to-date but not running, it should be started.


aanand:
After giving it some thought, it'd be far easier to just label every container with a hash generated from the service dict that was used to create it, and compare hashes to see if a container has changed.

This means we can avoid doing any of the configuration diffing stuff, which is (a) hairy (b) going to be a total pain to debug.

Thoughts?

aanand:
After giving it some thought, it'd be far easier to just label every container with a hash generated from the service dict that was used to create it, and compare hashes to see if a container needs recreating.

This means we can avoid doing any of the configuration diffing stuff, which is (a) hairy to implement (b) full of corner cases (c) going to be a total pain to debug.

Thoughts?

dnephin:
I agree, I like the idea of using a hash. It should make things a lot easier.

aanand:
Pushed an initial implementation using hashes.

aanand:
Downstream dependent services are now recreated when an upstream dependency is recreated.

aanand:
This is now ready for review.

aanand:
One thing worth noting: the `--no-build` flag has been rendered somewhat useless, as we now *have* to hit the API (to determine the image ID) before creating containers. Still, we're calling `inspect_image()` rather than `images()`, so perhaps it won't be a huge performance regression.

Compose now shows a warning if you pass `--no-build` but the image doesn't exist.

aanand:
Just rebasing this now that #1356 is in!

compose/project.py
```diff
@@ -195,24 +195,59 @@ def build(self, service_names=None, no_cache=False):
...
+
+        for service in services:
+            updated_dependencies = [
+                name
+                for name in service.get_dependency_names()
+                if name in plans
+                and plans[name].action == 'recreate'
+            ]
+
+            if updated_dependencies:
```
dnephin: When we were chatting about #1395 I suggested to support the feature we'd probably want to store the [DAG](http://en.wikipedia.org/wiki/Directed_acyclic_graph) of the services in `compose.project.sort_service_dicts()` instead of flattening it immediately.

I think that data structure would make this easier as well. Instead of having to lookup the dependencies for each service, we could immediately add recreate plans for all dependencies. When it's available we could consider using it here.

compose/service.py
```diff
@@ -183,49 +193,154 @@ def create_container(self,
...
+        if self.can_be_built():
+            return self.full_name
+        else:
+            return self.options['image']
+
+    def converge(self,
+                 allow_recreate=True,
+                 smart_recreate=False,
+                 insecure_registry=False,
+                 do_build=True):
```
dnephin: This isn't called anymore, is it ?

aanand: True - I'll remove it.

compose/project.py
aanand: Yeah, I think that'd be a good refactor as part of implementing #1395.

aanand:
Rebased and removed `converge()` from `Service`.

aanand:
Whoops, the tests are still using `converge()`. I'll move it into a test helper in a separate PR.

aanand:
Build is green.

dnephin:
LGTM

prologic:
This feature isn't it any published version of docker-compose yet right?

wader:
@prologic it was introduced in 1.3

prologic:
As a default (*no special cil option*) feature? :)

wader:
Sorry, not default, as a option to `up`, see `docker-compose up --help`

prologic:
```#!bash
dc up --help
    --force-recreate       Recreate containers even if their configuration and
                           image haven't changed. Incompatible with --no-recreate.
```

This one I guess :)

wader:
Hmm what version is that?
```
$ docker-compose up --help | grep recreate
    --x-smart-recreate     Only recreate containers whose configuration or
    --no-recreate          If containers already exist, don't recreate them.
```

prologic:
```#!bash
$ dc --version
docker-compose version: 1.4.0

$ dc up --help | grep recreate
volumes). To prevent Compose from picking up changes, use the `--no-recreate`
If you want to force Compose to stop and recreate all containers, use the
`--force-recreate` flag.
    --force-recreate       Recreate containers even if their configuration and
                           image haven't changed. Incompatible with --no-recreate.
    --no-recreate          If containers already exist, don't recreate them.
```

wader:
Your totally right, i'm running 1.3.1 and should probably update :)

From 1.4 changelog https://github.com/docker/compose/blob/master/CHANGELOG.md
```
The experimental --x-smart-recreate flag which introduced this feature in Compose 1.3.0 has been removed, and a --force-recreate flag has been added for when you want to recreate everything.
```

prologic:
:+1: Awesome :)

PR #1400, opened 2015-05-08
Added devices config handling and device HostConfig handling

DanElbert:
I'm new to contributing to compose, but this is a stab at adding the device option to the yml config, as per issue #754.

### DIFF
--- a/docs/yml.md
+++ b/docs/yml.md
@@ -29,8 +29,8 @@ image: a4bc65fd
 
 ### build
 
-Path to a directory containing a Dockerfile. When the value supplied is a 
-relative path, it is interpreted as relative to the location of the yml file 
+Path to a directory containing a Dockerfile. When the value supplied is a
+relative path, it is interpreted as relative to the location of the yml file
 itself. This directory is also the build context that is sent to the Docker daemon.
 
 Compose will build and tag it with a generated name, and use that image thereafter.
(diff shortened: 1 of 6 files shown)


dnephin:
I think this is a good start. I believe we also need to update `compose/config.py` to be able to merge the lists of devices when using `extends`

DanElbert:
I've updated the pull request.

docs/yml.md
```diff
@@ -342,6 +342,15 @@ dns_search:
   - dc2.example.com
 ```
 
+### devices
+
+Devices option.  List of device mappings.  Uses the same format as the --device docker client create option.
```
aanand: - First sentence is unnecessary
- `--device` should be in backticks
- Wrap to 80 chars please!

aanand:
Thanks!

For the sake of consistency, we should support a mapping as well as a list. Is that supported? If so, it should be documented.

aanand:
> For the sake of consistency, we should support a mapping as well as a list.

I retract this, as I was thinking about environment variables. We don't support YAML mappings for volumes. Sorry for the noise.

This looks good - just needs squashing to a single commit.

DanElbert:
Thanks for the feedback.  I was just going to ask about the mapping. :)

DanElbert:
All squashed.  Let me know if I missed anything else.

aanand:
LGTM

dnephin:
LGTM

PR #1404, opened 2015-05-11
Expand Expose statement to include IPAddr access

kentoj:
Since I only use Docker Containers directly via their IP address I don't bind any ports of my containers directly to my Host Machine. If the expose for Docker Compose works just like that of the Dockerfile then this definition should be updated to reflect that a port can be exposed on the container and then accessible by using the IP address of the container along with the port specified in the "EXPOSE' statement.

Edit: I forgot to add two words on line 138 before submitting the pull request. It should read "accessible to linked services as well as directly..." I forgot to add the words "as well" to transition the statement properly.

If you accept this PR I'll add those words or I can submit another one with the corrected wording.

### DIFF
--- a/docs/yml.md
+++ b/docs/yml.md
@@ -134,8 +134,8 @@ ports:
 
 ### expose
 
-Expose ports without publishing them to the host machine - they'll only be
-accessible to linked services. Only the internal port can be specified.
+Expose ports without publishing them to the host machine - this will make the ports 
+accessible to linked services as directly on the container when accessing it by its IP address assigned from the Docker Host. Only the internal port can be specified.
 
 ```
 expose:
--- a/docs/yml.md
+++ b/docs/yml.md
@@ -135,7 +135,7 @@ ports:
 ### expose
 
 Expose ports without publishing them to the host machine - this will make the ports 
-accessible to linked services as directly on the container when accessing it by its IP address assigned from the Docker Host. Only the internal port can be specified.
+accessible to linked services as well as directly on the container when accessing it by its IP address assigned from the Docker Host. Only the internal port can be specified.
 
 ```
 expose:


aanand:
It's true that the container is reachable at its IP address, but accessing it as such is not really a recommended practice for users of Compose, so I'm reluctant to encourage it. Anyone else have strong feelings about this?

dnephin:
I think this wording is actually misleading as well.

@kentoj something I learned recently is that you don't actually have to expose ports at all. It does't make them available, they're already available without you having to expose them.  I think expose is just used to provide environment variables for links. So if you're connecting to containers with `container_id:port` you don't even need to expose.

kentoj:
@aanand My particular use case with compose involves setting up containers to be DNS-resolvable hosts in a subdomain to provide name-based access to containers everywhere on my network that can see the DNS server. With the proper network traffic routing rules in place and a simple watcher script to update a DNS based off events from the Docker Event stream I have a nice setup where I spin up a container and immediately I can access it from any machine on my network by its domain name.

Since I register the name of my container and its IP dynamically with my DNS then I judge that I need to access Docker containers by IP address whether starting them up manually or via docker compose. You mention (@aanand) that accessing a container by its IP address is not really a recommended practice for users of Compose. Do you think what you said relates to my situation and use case?

Since my DNS server translates the name to an IP address for a Docker Container I see accessing a container by its domain name as nearly equivalent to accessing it by it's IP address.

The clarification I suggested is not necessarily valid given the comment by @dnephin though the wording of the documentation seems to misrepresent the EXPOSE as the _only_ way to access a container on a specific port. For beginners, like I was a few short months ago, this is very misleading and for a long time I thought that was the only way. I can see how the primary intended use case for different tools would or would not encourage the access of containers by their IP address.

Nevertheless, while I could be missing the point of how to best use Docker Compose the use case of having a docker host be a routable, resolvable host on a network seems very compelling to me from a automation and configuration management point of view. It definitely seems to me like Compose could accommodate the use case of a routable, resolvable host on a network for every started container.

Please inform me as to what you think on the matter if you see the value of my use case differently. I definitely have more thoughts on the issue and am running into some roadblocks with it already and I would like to become aware of the best thinking on these matters.

dnephin:
@kentoj are you interested in updating the docs to reflect that `expose` is really only a "hint" to other services? or should we close this PR?

kentoj:
Yes, I am. I think the wording could be improved from what I have done. It wasn't obvious to me as a beginner that the EXPOSE command was only acting as a publishing mechanism, not some access control mechanism.

aanand:
Closing due to inactivity. Feel free to open another PR if you'd like to make any clarifying changes to the `expose` docs as @dnephin suggested.

Issue #964, opened 2015-02-13
Request: Get all exposed project ports


Starefossen:
It would have been a nice addition to the `docker-compose port` command if you would just get all exposed project ports without having to probe each individual service and "internal" port.

For the following `docker-compose.yml` file:

``` yml
service1:
  ports:
    - "80"
    - "443"
service2:
  ports:
    - "1234"
```

The `docker-compose port` command with no arguments would return something like this:

```
service1:
  80:0.0.0.0:49201
  443:0.0.0.0:49202

service2:
  1234:0.0.0.0:49203
```

kojiromike:
What if there's more than one instance of the same service (scale>1)? How would you expect the output in that case?

Starefossen:
Like this?:

```
service1:
  80:0.0.0.0:49201
  80:0.0.0.0:49202
  443:0.0.0.0:49203
  443:0.0.0.0:49203

service2:
  1234:0.0.0.0:49204
  1234:0.0.0.0:49205
```

geerlingguy:
+1 to this—I would like `docker-compose port` with no options to print off all the ports for all the containers managed by the `compose` file. It's annoying in some cases to have to run multiple `docker-compose port [service]` commands (and for scripting, calling that multiple times is slow, at least on my Mac).

PR #1406, opened 2015-05-12
Fixing docker-compose port with scale (#667)

### LINKED ISSUE #667: Fig port with index not working with fig scale
Running `fig scale api=2` creates the following with output from `fig ps`:

```
            Name                          Command               State           Ports
---------------------------------------------------------------------------------------------
testcom_api_1   pm2 start /app/app.js --wa ...   Up      0.0.0.0:49153->80/tcp
testcom_api_2   pm2 start /app/app.js --wa ...   Up      0.0.0.0:49154->80/tcp
```
Running `fig port api 80` only returns the first container when we expected it to return both, is this intended?

Trying to use the `fig port --index=1 api 80`  or `fig port --index=2 api 80`  returns "No container found for api_1" and "No container found for api_2" respectively

vdemeester:
Fixes #667 and Closes #735 (taking over it). The PR #735 is old and no integration tests provided so I'm submitting a new one.

Another thing, not related. The current behavior is to show the first port by default if no ``--index`` provided. What if we change this to display all port (prefixed with the name/id/number), kinda like ``docker port`` ? I guess it would be better to discuss this in another issue/PR (but https://github.com/docker/compose/issues/964 is a little about that already). 

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -170,13 +170,14 @@ class TopLevelCommand(Command):
         Usage: port [options] SERVICE PRIVATE_PORT
 
         Options:
-            --protocol=proto  tcp or udp (defaults to tcp)
+            --protocol=proto  tcp or udp [default: tcp]
             --index=index     index of the container if there are multiple
-                              instances of a service (defaults to 1)
+                              instances of a service [default: 1]
         """
+        index = int(options.get('--index'))
         service = project.get_service(options['SERVICE'])
         try:
-            container = service.get_container(number=options.get('--index') or 1)
+            container = service.get_container(number=index)
         except ValueError as e:
             raise UserError(str(e))
         print(container.get_local_port(
--- a/tests/fixtures/ports-composefile-scale/docker-compose.yml
+++ b/tests/fixtures/ports-composefile-scale/docker-compose.yml
@@ -1,0 +1,6 @@
+
+simple:
+  image: busybox:latest
+  command: /bin/sleep 300
+  ports:
+    - '3000'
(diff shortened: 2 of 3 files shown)


compose/cli/main.py
```diff
@@ -165,15 +165,21 @@ def port(self, project, options):
...
         """
-        service = project.get_service(options['SERVICE'])
+        index = int(options.get('--index')) - 1
+        containers = sorted(
+            project.containers(service_names=[options['SERVICE']], stopped=True) +
+            project.containers(service_names=[options['SERVICE']], one_off=True),
+            key=attrgetter('name'))
         try:
-            container = service.get_container(number=options.get('--index') or 1)
+            container = containers[index]
```
dnephin: I don't think this is right.  This should be handled by `service.get_container()` like the old implementation.

dnephin:
> What if we change this to display all port (prefixed with the name/id/number), kinda like docker port ?

I think that would be cool, but it doesn't have to happen as part of this PR

compose/cli/main.py
vdemeester: Yeah my bad, fixing it ;)

vdemeester:
@dnephin yes definitely not in this one. I'll prepare another one for it ;).

vdemeester:
I've update it this morning (remove an useless ``except``).

/cc @aanand @bfirsh

vdemeester:
… and rebased :)

aanand:
LGTM

dnephin:
LGTM

PR #1415, opened 2015-05-15
Fix race condition in `docker-compose run`

aanand:
We shouldn't start the container before handing it off to dockerpty - dockerpty will start it after attaching, which is the correct order. Otherwise the container might exit before we attach to it, which can lead to weird bugs.

For example, it looks as though the data streamed from an `/attach` call is different if the container has exited. Try running `docker-compose run SERVICE seq 10` - I frequently get marching output like this:

```
$ docker-compose run web seq 10
1
 2
  3
   4
    5
     6
      7
       8
        9
         10
```

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -341,7 +341,6 @@ class TopLevelCommand(Command):
             service.start_container(container)
             print(container.name)
         else:
-            service.start_container(container)
             dockerpty.start(project.client, container.id, interactive=not options['-T'])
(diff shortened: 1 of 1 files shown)


aanand:
Thinking about it, that specific bug - an absence of `\r` characters in the output stream if the container has exited - is arguably a Docker bug. But anyway: we shouldn't start the container before attaching.

dnephin:
LGTM

Issue #1188, opened 2015-03-26
Support ro option in volumes_from


caiguanhao:
docker-compose 1.1.0 doesn't support `ro` in `volumes_from` yet.

**error message**:

```
Service "container" mounts volumes from "nginx_logs_1:ro", which is not the name of a service or container.
```

**docker-compose.yml**:

```yml
container:
  build: .
  ports:
    - '19198:22'
  volumes_from:
    - 'nginx_logs_1:ro'
```

**docker command that works**:

```
docker run -d -p=19198:22 --volumes-from=nginx_logs_1:ro --name=ssh_container_1 ssh_container
```

vdemeester:
#dibs on that one :)

PR #1420, opened 2015-05-16
Add support for options (ro/rw) in volumes_from (#1188)

vdemeester:
Add support for ``volumes_from`` with read-only, with integration and unit tests, related to #1188.

```yaml
container:
  build: .
  ports:
    - '19198:22'
  volumes_from:
    - 'nginx_logs_1:ro'
```

This changes a bit how ``Service`` treats its ``volumes_from`` attribute, as it now asks for a ``tuple`` of ``(object, string)`` where the object can be a ``Service`` or ``Container`` and the string could be ``None`` or a option like ``'ro'``, ``'rw'``.

I think there is rooms of improvement in the implementation, so reviews are more than welcome :wink:.

### DIFF
--- a/compose/project.py
+++ b/compose/project.py
@@ -7,9 +7,9 @@ from docker.errors import APIError
 
 from .config import get_service_name_from_net, ConfigurationError
 from .const import DEFAULT_TIMEOUT, LABEL_PROJECT, LABEL_SERVICE, LABEL_ONE_OFF
+from .service import Service, parse_volume_from_spec, VolumeFromSpec
 from .container import Container
 from .legacy import check_for_legacy_containers
-from .service import Service
 from .utils import parallel_execute
 
 log = logging.getLogger(__name__)
@@ -24,12 +24,15 @@ def sort_service_dicts(services):
     def get_service_names(links):
         return [link.split(':')[0] for link in links]
 
+    def get_service_names_from_volumes_from(volumes_from):
+        return [volume_from.split(':')[0] for volume_from in volumes_from]
+
     def get_service_dependents(service_dict, services):
         name = service_dict['name']
         return [
             service for service in services
             if (name in get_service_names(service.get('links', [])) or
-                name in service.get('volumes_from', []) or
+                name in get_service_names_from_volumes_from(service.get('volumes_from', [])) or
                 name == get_service_name_from_net(service.get('net')))
         ]
 
@@ -158,16 +161,19 @@ class Project(object):
     def get_volumes_from(self, service_dict):
         volumes_from = []
         if 'volumes_from' in service_dict:
-            for volume_name in service_dict.get('volumes_from', []):
+            for volume_from_config in service_dict.get('volumes_from', []):
+                volume_from_spec = parse_volume_from_spec(volume_from_config)
+                # Get service
                 try:
-                    service = self.get_service(volume_name)
-                    volumes_from.append(service)
+                    service_name = self.get_service(volume_from_spec.source)
+                    volume_from_spec = VolumeFromSpec(service_name, volume_from_spec.mode)
                 except NoSuchService:
                     try:
-                        container = Container.from_id(self.client, volume_name)
-                        volumes_from.append(container)
+                        container_name = Container.from_id(self.client, volume_from_spec.source)
+                        volume_from_spec = VolumeFromSpec(container_name, volume_from_spec.mode)
                     except APIError:
-                        raise ConfigurationError('Service "%s" mounts volumes from "%s", which is not the name of a service or container.' % (service_dict['name'], volume_name))
+                        raise ConfigurationError('Service "%s" mounts volumes from "%s", which is not the name of a service or container.' % (volume_from_config, volume_from_spec.source))
+                volumes_from.append(volume_from_spec)
             del service_dict['volumes_from']
         return volumes_from
 
--- a/compose/service.py
+++ b/compose/service.py
@@ -4,7 +4,6 @@ from collections import namedtuple
 import logging
 import re
 import sys
-from operator import attrgetter
 
 import six
 from docker.errors import APIError
@@ -71,6 +70,9 @@ class NoSuchImageError(Exception):
 
 
 VolumeSpec = namedtuple('VolumeSpec', 'external internal mode')
+
+
+VolumeFromSpec = namedtuple('VolumeFromSpec', 'source mode')
 
 
 ServiceName = namedtuple('ServiceName', 'project service number')
@@ -440,7 +442,7 @@ class Service(object):
         return [s.name for (s, _) in self.links]
 
     def get_volumes_from_names(self):
-        return [s.name for s in self.volumes_from if isinstance(s, Service)]
+        return [s.source.name for s in self.volumes_from if isinstance(s.source, Service)]
 
     def get_net_name(self):
         if isinstance(self.net, Service):
@@ -485,16 +487,9 @@ class Service(object):
 
     def _get_volumes_from(self):
         volumes_from = []
-        for volume_source in self.volumes_from:
-            if isinstance(volume_source, Service):
-                containers = volume_source.containers(stopped=True)
-                if not containers:
-                    volumes_from.append(volume_source.create_container().id)
-                else:
-                    volumes_from.extend(map(attrgetter('id'), containers))
-
-            elif isinstance(volume_source, Container):
-                volumes_from.append(volume_source.id)
+        for volume_from_spec in self.volumes_from:
+            volumes = build_volume_from(volume_from_spec)
+            volumes_from.extend(volumes)
 
         return volumes_from
 
@@ -810,6 +805,38 @@ def parse_volume_spec(volume_config):
         parts.append('rw')
 
     return VolumeSpec(*parts)
+
+
+def build_volume_from(volume_from_spec):
+    volumes_from = []
+    if isinstance(volume_from_spec.source, Service):
+        containers = volume_from_spec.source.containers(stopped=True)
+        if not containers:
+            volumes_from = ["{}:{}".format(volume_from_spec.source.create_container().id, volume_from_spec.mode)]
+        else:
+            volumes_from = ["{}:{}".format(container.id, volume_from_spec.mode) for container in containers]
+    elif isinstance(volume_from_spec.source, Container):
+        volumes_from = ["{}:{}".format(volume_from_spec.source.id, volume_from_spec.mode)]
+    return volumes_from
+
+
+def parse_volume_from_spec(volume_from_config):
+    parts = volume_from_config.split(':')
+    if len(parts) > 2:
+        raise ConfigError("Volume %s has incorrect format, should be "
+                          "external:internal[:mode]" % volume_from_config)
+
+    if len(parts) == 1:
+        source = parts[0]
+        mode = 'rw'
+    else:
+        source, mode = parts
+
+    if mode not in ('rw', 'ro'):
+        raise ConfigError("VolumeFrom %s has invalid mode (%s), should be "
+                          "one of: rw, ro." % (volume_from_config, mode))
+
+    return VolumeFromSpec(source, mode)
 
 
 # Ports
--- a/tests/integration/project_test.py
+++ b/tests/integration/project_test.py
@@ -3,6 +3,7 @@ from __future__ import unicode_literals
 from compose import config
 from compose.const import LABEL_PROJECT
 from compose.project import Project
+from compose.service import VolumeFromSpec
 from compose.container import Container
 from .testcases import DockerClientTestCase
 
@@ -68,7 +69,7 @@ class ProjectTest(DockerClientTestCase):
         )
         db = project.get_service('db')
         data = project.get_service('data')
-        self.assertEqual(db.volumes_from, [data])
+        self.assertEqual(db.volumes_from, [VolumeFromSpec(data, 'rw')])
 
     def test_volumes_from_container(self):
         data_container = Container.create(
@@ -89,7 +90,7 @@ class ProjectTest(DockerClientTestCase):
             client=self.client,
         )
         db = project.get_service('db')
-        self.assertEqual(db.volumes_from, [data_container])
+        self.assertEqual(db._get_volumes_from(), [data_container.id + ':rw'])
 
     def test_net_from_service(self):
         project = Project.from_dicts(
--- a/tests/integration/service_test.py
+++ b/tests/integration/service_test.py
@@ -22,6 +22,7 @@ from compose.service import (
     ConvergencePlan,
     Service,
     build_extra_hosts,
+    VolumeFromSpec,
 )
 from compose.container import Container
 from .testcases import DockerClientTestCase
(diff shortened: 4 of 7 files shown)


compose/service.py
```diff
@@ -334,16 +334,21 @@ def _get_links(self, link_to_self):
 
     def _get_volumes_from(self):
         volumes_from = []
-        for volume_source in self.volumes_from:
+        for volume_from in self.volumes_from:
+            volume_source, mount_option = volume_from
```
dnephin: A namedtuple would be good, that way you could do `volume_from.source` and `volume_from.mount_option`.

vdemeester:
@dnephin updated with a ``namedtuple`` (defined in ``config.py``, wasn't really sure about that..)

funkyfuture:
i suggest to put the declaration of `VolumesFrom` into `__init.py__`. it's now placed in a module that itself doesn't make use of it, but various others. this would also prevent later possible circular references through imports.

vdemeester:
@funkyfuture ah yes, good idea, updated my PR with that :wink:

funkyfuture:
you may change the issue-title as not only `ro` is supported as suffix.

compose/__init__.py
```diff
@@ -1,4 +1,8 @@
 from __future__ import unicode_literals
 from .service import Service  # noqa:flake8
 
+from collections import namedtuple
+
 __version__ = '1.3.0dev'
+
+VolumeFrom = namedtuple("VolumeFrom", 'source mount_option')
```
dnephin: minor point, but this is only used in one module right now (outside of tests), so it would be good to keep it local to that module.  

It's good to keep `__init__.py` as small as possible

dnephin:
I think this is looking good.

At some point we're going to need to refactor part of this. It seems like the logic overlaps a lot with `net` handling and is pretty similar between `config` and `service` modules, but that shouldn't happen as part of this change.

compose/__init__.py
vdemeester: Yeah that's right.. I moved it back to project.py — i'll see if janky is happy.. I'm on a EDGE network right now, it's taking for ever to build/test it on my laptop :cry: .

vdemeester:
I've update and rebase it against master (as there was a small conflict with a test rename).

funkyfuture:
well, it's not in `config.py` anymore. `project.py` seems more proper.

however, i learned yesterday that CPython 2.6 doesn't have `NamedTuple` in its library. but for some reason that interpreter is supposed to be supported.

vdemeester:
@funkyfuture ah :disappointed:, not cool.. 

> With the 2.6.9 release, and five years after its first release, the Python 2.6 series is now officially retired. All official maintenance for Python 2.6, including security patches, has ended.

Are we sure we want to continue supporting a version that is not supported by the python foundation since 2 years ? :P

aanand:
Looks like OS X has shipped with 2.7 since Mavericks. I think we can probably drop 2.6 support now.

aanand:
I like the look of this. Just needs docs.

vdemeester:
Oh yeah, I didn't think of updating documentation. I'll update this in a few hours ;-)

dnephin:
>  CPython 2.6 doesn't have NamedTuple

Sure it does! https://docs.python.org/2.6/library/collections.html#collections.namedtuple

It was added in 2.6, we use it in a few places already in compose.

vdemeester:
Documentation updated.

funkyfuture:
> Sure it does!

sorry for the confusion and driving this off-topic. i wrongly assumed `sys.version_info` was a named tuple. so, thank you for the correction. however, dropping 2.6 seems cool.

tests/integration/project_test.py
```diff
@@ -67,7 +87,30 @@ def test_volumes_from_container(self):
             client=self.client,
         )
         db = project.get_service('db')
-        self.assertEqual(db.volumes_from, [data_container])
+        self.assertEqual(db._get_volumes_from(), [data_container.id])
+
+        project.kill()
+        project.remove_stopped()
```
dnephin: These two lines wont run if the test fails. The test case should take care of this for you, so you can remove it.

tests/integration/project_test.py
```diff
@@ -67,7 +87,30 @@ def test_volumes_from_container(self):
...
+                    'volumes_from': ['composetest_data_container:ro'],
+                },
+            }),
+            client=self.client,
+        )
+        db = project.get_service('db')
+        self.assertEqual(db._get_volumes_from(), [data_container.id + ':ro'])
 
         project.kill()
         project.remove_stopped()
```
dnephin: Same here, if you don't mind removing it, that would be great.

tests/integration/service_test.py
```diff
@@ -236,8 +237,8 @@ def test_create_container_with_home_and_env_var_in_volume_path(self):
     def test_create_container_with_volumes_from(self):
         volume_service = self.create_service('data')
         volume_container_1 = volume_service.create_container()
-        volume_container_2 = Container.create(self.client, image='busybox:latest', command=["top"])
-        host_service = self.create_service('host', volumes_from=[volume_service, volume_container_2])
+        volume_container_2 = Container.create(self.client, image='busybox:latest', command=["/bin/sleep", "300"])
```
dnephin: I think this is a bad merge, it should be `top`

tests/integration/service_test.py
```diff
@@ -245,6 +246,18 @@ def test_create_container_with_volumes_from(self):
         self.assertIn(volume_container_2.id,
                       host_container.get('HostConfig.VolumesFrom'))
 
+    def test_create_container_with_volumes_from_read_only(self):
+        volume_service = self.create_service('data')
+        volume_container_1 = volume_service.create_container()
+        volume_container_2 = Container.create(self.client, image='busybox:latest', command=["/bin/sleep", "300"])
```
dnephin: Please use `top`, we changed everything over to use it so the tests would run faster

vdemeester:
@dnephin Updated :wink:

aanand:
As @samalba reports in https://github.com/docker/compose/issues/1493, it'd be good to support all the modes the Docker daemon supports. This PR might already do that, but it'd be good to test and document it.

vdemeester:
@aanand ah yes, I should update the documentation, and probably tests as well then :wink:

vdemeester:
Well, it's just gonna work with Docker 1.7.0 as it's a pretty recent change (https://github.com/docker/docker/commit/af7d17a6c906d348eb0929766a81fd04fc173c44). 

@aanand How should we handle that ?

vdemeester:
Few bugs on the build with 1.7.0-rc1, I'll wait for https://github.com/docker/compose/pull/1497 to be merged then :). I've update the Dockerfile to only use the 1.7.0-rc1, but I'll update this when needed :wink:.

aanand:
I've implemented support for any mode with the `volumes` key in https://github.com/docker/compose/pull/1515, so here it just needs to be implemented for `volumes_from`.

vdemeester:
@aanand ok :wink:. I think that's already what's implemented (I look at #1515 and this was nothing I modified :stuck_out_tongue:)

aanand:
Great.

I think the dependency on Docker 1.7 can actually be relaxed, since my understanding is that the volume mode support might be backported to the 1.6 series.

This means that your integration tests won't pass, but as long as there's a test in `test/unit/service.py` that uses a mock client (like [test_create_with_special_volume_mode](https://github.com/docker/compose/pull/1515/files#diff-bca0836805752707d5852da846492291R395) in my PR) it should be fine to remove the integration tests.

vdemeester:
I've rebased it since #1497 has been merged :wink:.

vdemeester:
ohhh @aanand, just got what you meant… :sweat_smile: I'll update it this week end then :smile:

vdemeester:
@aanand just rebased it (there was a conflict on the yaml file) and removed the integration tests as said.

vdemeester:
@aanand @dnephin @bfirsh Anything more to do on this one ? :blush:

compose/project.py
```diff
@@ -13,6 +14,8 @@
 
 log = logging.getLogger(__name__)
 
+VolumeFrom = namedtuple("VolumeFrom", 'source mount_option')
```
aanand: For the sake of consistency, could you:

1. rename `mount_option` to `mode`
2. rename `VolumeFrom` to `VolumesFromSpec`
3. move it into `compose.service` and import it here

aanand: (We should really move all those named tuples into their own file, along with any related functions for parsing them)

compose/project.py
```diff
@@ -158,15 +164,19 @@ def get_volumes_from(self, service_dict):
         volumes_from = []
         if 'volumes_from' in service_dict:
             for volume_name in service_dict.get('volumes_from', []):
+                if ':' in volume_name:
+                    service_name, mount_option = volume_name.split(':', 1)
+                else:
+                    service_name, mount_option = volume_name, None
```
aanand: Could you extract this logic into a function, and do the same with the string-building logic in `Service._get_volumes_from()`? Say, `parse_volumes_from()` and `build_volumes_from()`.

compose/project.py
vdemeester: Ok, will do :wink:.

Moving them in their own file along with related functions should be in another PR I think, right ?

compose/service.py
```diff
@@ -442,16 +442,20 @@ def _get_links(self, link_to_self):
...
+            else:
+                mount_option = ''
+            if isinstance(volume_from.source, Service):
+                containers = volume_from.source.containers(stopped=True)
                 if not containers:
-                    volumes_from.append(volume_source.create_container().id)
+                    volumes_from.append(volume_from.source.create_container().id + mount_option)
                 else:
-                    volumes_from.extend(map(attrgetter('id'), containers))
+                    volumes_from.extend([container_id + mount_option for container_id in map(attrgetter('id'), containers)])
```
aanand: Now that this is a comprehension, there's not much point in keeping the `map`+`attrgetter` stuff - it can just iterate over `containers`.

compose/project.py
aanand: Agreed.

compose/project.py
vdemeester: Ok.

thaJeztah:
ping @aanand PTAL; looks like the PR was updated :)

vdemeester:
@thaJeztah he did, he did ;-). I still have to update some code though :-).

thaJeztah:
@vdemeester :+1: wasn't sure if you forgot to ping him :smile:

compose/service.py
```diff
@@ -770,6 +765,47 @@ def parse_volume_spec(volume_config):
...
+        if not containers:
+            volumes_from = ["{}:{}".format(volume_from_spec.source.create_container().id, volume_from_spec.mode)]
+        else:
+            volumes_from = ["{}:{}".format(container.id, volume_from_spec.mode) for container in containers]
+    elif isinstance(volume_from_spec.source, Container):
+        volumes_from = ["{}:{}".format(volume_from_spec.source.id, volume_from_spec.mode)]
+    return volumes_from
+
+
+def parse_volume_from_spec(volume_from_config, get_service, client):
```
vdemeester: I did something a little strange I feel, because I need to call ``get_service`` or use ``client`` to get the source (service or container). I ended up passing the method and the client reference as parameters.

compose/service.py
```diff
@@ -857,3 +893,12 @@ def build_extra_hosts(extra_hosts_config):
         "extra_hosts_config \"%s\" must be either a list of strings or a string->string mapping," %
         extra_hosts_config
     )
+
+
+class NoSuchService(Exception):
```
vdemeester: This one was defined in ``project.py`` so moved here to avoid circular dependency.

compose/service.py
aanand: Sorry, I was unclear. I was thinking that just the string splitting and concatenation logic should be moved into utility functions - the part where we determine whether the source is a service or a container (as well as the corresponding logic in `Service._get_volumes_from` where we check whether we're looking at an instance of Service or Container) can stay where it is.

vdemeester: Don't be sorry :wink: ; that's what I understood, but… I though I could do both in there, to avoid updating VolumeFromSpec afterwards. But now that I'm looking at it, it feels really bad :sweat_smile:, I'll fix that :wink:.

compose/project.py
```diff
@@ -157,16 +160,19 @@ def get_links(self, service_dict):
...
         if 'volumes_from' in service_dict:
-            for volume_name in service_dict.get('volumes_from', []):
+            for volume_from_config in service_dict.get('volumes_from', []):
+                volume_from_spec = parse_volume_from_spec(volume_from_config)
+                # Get service
                 try:
-                    service = self.get_service(volume_name)
-                    volumes_from.append(service)
+                    service_name = self.get_service(volume_from_spec.source)
+                    volume_from_spec = VolumeFromSpec(service_name, volume_from_spec.mode)
```
vdemeester: @aanand I wasn't sure if ``parse_volume_from_spec`` should return a ``VolumeFromSpec`` as I need to create a new one few lines later..

vdemeester:
@aanand I think i've update the code according to all your comments :)

vdemeester:
@aanand rebased it (but seems like janky is not happy :sob:)

thaJeztah:
kicked Janky (in case it helps :-))

vdemeester:
@thaJeztah it did :wink:

vdemeester:
ping @aanand @mnowster

compose/project.py
```diff
@@ -24,12 +24,15 @@ def sort_service_dicts(services):
     def get_service_names(links):
         return [link.split(':')[0] for link in links]
 
+    def get_service_names_from_volumes_from(volumes_from):
+        return [volume_from.split(':')[0] for volume_from in volumes_from]
```
aanand: Could we also use `parse_volume_from_spec` here? Like:

```python
return [
    parse_volume_from_spec(vf).source
    for vf in volumes_from
]
```

compose/service.py
```diff
@@ -812,6 +807,38 @@ def parse_volume_spec(volume_config):
...
+
+    if len(parts) == 1:
+        source = parts[0]
+        mode = 'rw'
+    else:
+        source, mode = parts
+
+    if mode not in ('rw', 'ro'):
+        raise ConfigError("VolumeFrom %s has invalid mode (%s), should be "
+                          "one of: rw, ro." % (volume_from_config, mode))
```
aanand: We shouldn't validate the mode client-side - we can leave that to the daemon.

aanand:
After stepping back, I'm a little concerned that in various parts of the code, a `VolumeFromSpec`'s `source` property can be either a string or a service/container. I don't think we can escape the service/container conflation, but adding strings into the mix feels bad to me.

(I'm aware this was informed by my suggestions, which might not have been very good ones. Sorry.)

ping @dnephin @mnowster - what do you think of the code?

compose/project.py
dnephin: +1

dnephin:
Sorry it's taken me a while to look at this.

I think overall it looks great, but I share the concern about using `VolumeFromSpec` with different types for source.  I'd even say let's create another namedtuple `Volume`, that will probably have the same fields, but `source` is expected to be a container or service, instead of a string.

That should make it a bit more clear, and maybe in the future those types will diverage.

mnowster:
Hi there @vdemeester , would you have the time to address the feedback from @dnephin ?  We'd like to get this PR in for 1.5. 

If you don't have the time, no worries, I can pick this up(your contributions will remain in tact). Let me know, thanks :)

vdemeester:
@mnowster waw it's been a while :sweat_smile:. There is kind of a lot of conflicts and my head is a little lost on `compose` on that one :sweat:, so feel free to *pick it up* :wink:.

mnowster:
@vdemeester I know :disappointed:  I'm sorry it's taken so long, been so much going on but we're trying to get better at responding. No problem, I'll pick it up, thank you for all your work on this so far, your commits will remain so you'll be credited :+1: :smiley:

PR #1423, opened 2015-05-17
[WIP] add an option to publish all ports

docteurklein:
I'm not even sure it works.
There is no tests.

I'm just pushing this to see if you guys think it worths to be added.

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -441,6 +441,7 @@ class TopLevelCommand(Command):
             --no-color             Produce monochrome output.
             --no-deps              Don't start linked services.
             --no-recreate          If containers already exist, don't recreate them.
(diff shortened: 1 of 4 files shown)


compose/cli/main.py
```diff
@@ -454,11 +455,13 @@ def up(self, project, options):
         start_deps = not options['--no-deps']
         recreate = not options['--no-recreate']
         service_names = options['SERVICE']
+        publish_all_ports = not options['-P']
```
docteurklein: in fact the `not` seems wrong.

funkyfuture:
as you make it a default, it possibly breaks existing setups.

dnephin:
http://docs.docker.com/compose/django/

I don't think this is something we want to support. You can specify ports to expose from the configuration, but we don't really want flags on `up` that can change the composition. The flags are just to change the command behaviour and output.

PR #1437, opened 2015-05-21
Project.containers with service_names

### LINKED ISSUE #1434: Unable to show logs from just a single service
Running the latest build from master branch:

When I try to `docker-compose logs MYSERVICE` I get the entire stack of services log output.

dnephin:
Fixes #1434 and adds tests for this case, I guess we were missing coverage here.


aanand:
LGTM

PR #1440, opened 2015-05-21
Legacy fixes

aanand:
Fixed two bugs:

- One-off containers were included in the warning log messages, which can make for unreadable output when there are lots (as there often are).

- Compose was attempting to recreate one-off containers as normal containers when migrating.

Did some cleanup:

- Rename `migration` module to `legacy` to make its legacy-ness explicit

- Move `check_for_legacy_containers` into `legacy` module

- Fix migration test so it can be run in isolation


dnephin:
LGTM

PR #1441, opened 2015-05-21
Bail out immediately if there are legacy containers

aanand:
Sits on top of #1440.

If there are legacy containers, then in almost all cases it's not safe to keep doing whatever we were about to do. Additionally, the error output was pretty difficult to read.

I've changed it so we display a single message with all the names, plus directions for removing the containers if the user doesn't want to migrate, and then exit.


dnephin:
LGTM

bfirsh:
LGTM

PR #1444, opened 2015-05-21
Migrate containers in dependency order

aanand:
Sits on top of #1441.

This fixes a bug introduced by #1441 where migration would fail with an error if a downstream container was migrated before its upstream dependencies (e.g. web before db), due to `check_for_legacy_containers()` being implicitly called when we fetch `links`, `volumes_from` or `net` dependencies.

### DIFF
--- a/compose/legacy.py
+++ b/compose/legacy.py
@@ -35,15 +35,15 @@ def check_for_legacy_containers(
     and warn the user that those containers may need to be migrated to
     using labels, so that compose can find them.
     """
-    names = list(get_legacy_container_names(
+    containers = list(get_legacy_containers(
         client,
         project,
         services,
         stopped=stopped,
         one_off=one_off))
 
-    if names:
-        raise LegacyContainersError(names)
+    if containers:
+        raise LegacyContainersError([c.name for c in containers])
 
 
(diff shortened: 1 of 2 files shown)


compose/legacy.py
```diff
@@ -72,26 +72,31 @@ def add_labels(project, container, name):
...
         stopped=False,
         one_off=False):
 
-    for container in client.containers(all=stopped):
-        name = get_container_name(container)
-        for service in services:
+    containers = client.containers(all=stopped)
+
+    for service in services:
+        for container in containers:
```
dnephin: Ok, so if I understand this correctly, the change here is that now we're iterating over things in the order of the passed in services, instead of just the container list order (which is basically undefined).

We have to assume the services were passed in the correct order, but that's probably safe to assume, since that's how we store them generally.

Does that sound right?

aanand: Yes, sorry, the actual *fix* is a bit subtle - it relies on the fact that `Project` keeps its services in order, and so `project.service_names` returns them in order. That's a widely relied-upon constraint, though, so I don't think this is brittle.

dnephin:
LGTM

PR #1445, opened 2015-05-21
Use 'top' instead of 'sleep' as a dummy command

aanand:
When `sleep` is PID 1 it doesn't respond to SIGTERM, which means a call to `/containers/:id/stop` blocks for 10 seconds before SIGKILL is invoked. `top` responds properly and exits.

On my machine, this change cuts just over 2 minutes, or around 35%, off the total test suite running time. Not bad.


cpuguy83:
Very nice. Also keep in mind I think top doesn't exit 0 with SIGTERM

dnephin:
LGTM

PR #1452, opened 2015-05-23
Support json serialization of env vars

cjelly:
Env vars are great for injecting config into Docker containers, but when you begin to need nested config options, they become a bit cumbersome (imagine variables for each bit of info about a specific resource or service that your application needs to be able to connect to). This change enables docker-compose users to specify yaml dictionaries or lists under an environment variable option, which is then serialized to JSON for deserialization in the container.

Example:

```
environment:
   DATABASES:
      mysql_db:
         host: localhost
         username: root
         password: 
```
would result in

```
$ echo $DATABASES
$ {"mysql_db": {"host": "localhost", "username": "root", "password": ""}}
```

In theory you could store your entire app config under one env var if you really wanted to.

### DIFF
--- a/compose/config.py
+++ b/compose/config.py
@@ -1,6 +1,7 @@
 import os
 import yaml
 import six
+import json
 
 
 DOCKER_CONFIG_KEYS = [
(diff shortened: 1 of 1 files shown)


dnephin:
Thanks for the contribution!

While this is an interesting idea, I don't think compose is the right place to handle it.

You could accomplish this by using `env_file`, and `json.dump()` values into the file, or you could set host environment variables and use a variable without a value in the `docker-compose.yml` to pass it in.

cjelly:
Thanks for the feedback - I'll close this out.

PR #1459, opened 2015-05-26
Update description of Compose

bfirsh:
"Define and run multi-container applications with Docker"

Not just development environments, and "complex" is not clear and
not really true.


aanand:
LGTM

Issue #1378, opened 2015-04-30
Allow scaling services with static ports on Swarm with affinity


frol:
Please, look at this config:

```yml
mynode:
    image: myimage
    environment:
        - "affinity:container!=myproject_mynode_*"
    ports:
        - "50010:50010"
```

Once I try to scale it with `docker-compose scale mynode=3` it rejects to scale with the following error:

> Service "mynode" cannot be scaled because it specifies a port on the host. If multiple containers for this service were created, the port would clash.
> 
> Remove the ":" from the port definition in docker-compose.yml so Docker can choose a random port for each container.

However, the ports won't clash because of the affinity.

Now, I'm forced to use `net: host` to avoid this.

aanand:
Yes, we should perhaps just remove that check now that Swarm can schedule multiple containers with the same port open. I'm not sure if it ever really helped anyone.

jamshid:
Getting rid of the error message would be great. Would fix https://github.com/docker/compose/issues/1133 too.

wricardo:
:+1: for getting rid of this error message

aanm:
@frol You can close this, it has been fixed.

frol:
@aanm Cool! Thanks!

PR #1466, opened 2015-05-26
Modified scale awareness from exception to warning

aanm:
It starts not making sense only use docker-compose on a single host environment. I'm proposing removing the scale awareness of port collision from exception to a warning.
Fixes: #1378
Signed-off-by: André Martins <martins@noironetworks.com>

### DIFF
--- a/compose/service.py
+++ b/compose/service.py
@@ -55,10 +55,6 @@ class BuildError(Exception):
         self.reason = reason
 
 
-class CannotBeScaledError(Exception):
-    pass
-
-
 class ConfigError(ValueError):
     pass
 
@@ -154,7 +150,9 @@ class Service(object):
         - removes all stopped containers
         """
         if not self.can_be_scaled():
-            raise CannotBeScaledError()
+            log.warn('Service %s specifies a port on the host. If multiple containers '
+                     'for this service are created on a single host, the port will clash.'
+                     % self.name)
 
         # Create enough containers
         containers = self.containers(stopped=True)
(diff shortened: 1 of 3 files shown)


dnephin:
I think this is a good idea.  I'll defer to others about the text of the warning.

compose/service.py
```diff
@@ -154,7 +150,9 @@ def scale(self, desired_num):
         - removes all stopped containers
         """
         if not self.can_be_scaled():
-            raise CannotBeScaledError()
+            log.warn('Service %s only specifies a port on the host. If multiple '
+                     'containers for this service are created on a single host, '
+                     'the port will clash.' % self.name)
```
aanand: Pretty clear warning, but I don't think we need "only".

aanm: Oh, you're right, makes more sense without the "only"

dnephin:
LGTM

aanand:
LGTM

chanwit:
@aanand @dnephin This is a trivial change, but important. Is it possible to include this change into 1.3?

aanand:
ping @bfirsh - reckon we could consider this a "bug fix" and include it in 1.3?

bfirsh:
:+1:

aanand:
Nice. Let's do another RC with this fix then (after a few days, just in case any more issues crop up).

chanwit:
Thank you!

wader:
It should be possible to use scale and `ip::containerPort` format right? adjust `can_be_scaled`?

PR #1468, opened 2015-05-27
Extend up -t to pass timeout to stop running containers

thieman:
This PR extends the `timeout` option on `docker-compose up` to include the timeout as a parameter when stopping containers that are already running at the time of `docker-compose up` execution. Currently, this timeout parameter is only applied to the stopping of containers at the termination of a `docker-compose up` command that is run without the `-d` (detach) parameter. That behavior remains unchanged.

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -447,9 +447,9 @@ class TopLevelCommand(Command):
                                    image needs to be updated. (EXPERIMENTAL)
             --no-recreate          If containers already exist, don't recreate them.
             --no-build             Don't build an image, even if it's missing
-            -t, --timeout TIMEOUT  When attached, use this timeout in seconds
-                                   for the shutdown. (default: 10)
-
+            -t, --timeout TIMEOUT  Use this timeout in seconds for container shutdown
+                                   when attached or when containers are already
+                                   running. (default: 10)
         """
         insecure_registry = options['--allow-insecure-ssl']
         detached = options['-d']
@@ -460,6 +460,7 @@ class TopLevelCommand(Command):
         allow_recreate = not options['--no-recreate']
         smart_recreate = options['--x-smart-recreate']
         service_names = options['SERVICE']
+        timeout = int(options['--timeout']) if options['--timeout'] is not None else None
 
         project.up(
             service_names=service_names,
@@ -468,6 +469,7 @@ class TopLevelCommand(Command):
             smart_recreate=smart_recreate,
             insecure_registry=insecure_registry,
             do_build=not options['--no-build'],
+            timeout=timeout
         )
 
         to_attach = [c for s in project.get_services(service_names) for c in s.containers()]
@@ -485,8 +487,7 @@ class TopLevelCommand(Command):
                 signal.signal(signal.SIGINT, handler)
 
                 print("Gracefully stopping... (press Ctrl+C again to force)")
-                timeout = options.get('--timeout')
-                params = {} if timeout is None else {'timeout': int(timeout)}
+                params = {} if timeout is None else {'timeout': timeout}
                 project.stop(service_names=service_names, **params)
 
     def migrate_to_labels(self, project, _options):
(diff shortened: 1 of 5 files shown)


tests/unit/service_test.py
```diff
@@ -254,6 +254,21 @@ def test_recreate_container(self, _):
...
+        service.image = lambda: {'Id': 'abc123'}
+        new_container = service.recreate_container(mock_container, timeout=1)
+
+        mock_container.stop.assert_called_once_with(timeout=1)
+        self.mock_client.rename.assert_called_once_with(
+            mock_container.id,
+            '%s_%s' % (mock_container.short_id, mock_container.name))
+
+        new_container.start.assert_called_once_with()
+        mock_container.remove.assert_called_once_with()
```
aanand: I think we just need to test the call to `stop` with `timeout=1` here - the other calls are already being tested in `test_recreate_container`.

compose/service.py
```diff
@@ -359,7 +362,8 @@ def recreate_container(self,
         """
         log.info("Recreating %s..." % container.name)
         try:
-            container.stop()
+            stop_params = {} if timeout is None else {'timeout': int(timeout)}
+            container.stop(**stop_params)
```
dnephin: I think you should be able to do just 

```
container.stop(timeout=timeout)
```

Since you'e already casted timeout to an int, or it's None already.

compose/cli/main.py
```diff
@@ -485,8 +487,7 @@ def handler(signal, frame):
                 signal.signal(signal.SIGINT, handler)
 
                 print("Gracefully stopping... (press Ctrl+C again to force)")
-                timeout = options.get('--timeout')
-                params = {} if timeout is None else {'timeout': int(timeout)}
+                params = {} if timeout is None else {'timeout': timeout}
                 project.stop(service_names=service_names, **params)
```
dnephin: Now that you have a timeout as an int or None, I think you can do 

```
project.stop(service_names=service_names, timeout=timeout)
```

tests/unit/service_test.py
```diff
@@ -254,6 +254,15 @@ def test_recreate_container(self, _):
         new_container.start.assert_called_once_with()
         mock_container.remove.assert_called_once_with()
 
+    @mock.patch('compose.service.Container', autospec=True)
+    def test_recreate_container_with_timeout(self, _):
+        mock_container = mock.create_autospec(Container)
+        service = Service('foo', client=self.mock_client, image='someimage')
+        service.image = lambda: {'Id': 'abc123'}
```
dnephin: The unit-under-test (in this case `Service`) really shouldn't be modified, otherwise you risk having the unit behave different, and the test not reflect normal operation.

Could this be provided as a return value from mock_client instead?

```
mock_client.inspect_image.return_value = {'Id': ...}
```

dnephin:
Overall I think this will be a good change, just some minor style requests

thieman:
Docker-py seems to be freaking out when `timeout` is passed as `None`. That's unfortunate. I'll revert the timeout changes and see if I can get the mock change to pass.

thieman:
Review comments integrated (where possible) and commits squashed.

aanand:
Thanks. Looks like we can still remove that second `int()` cast, even if we have to keep the `params` dict thing?

thieman:
Sorry, thought it might have been worth keeping for defensive purposes. Changed now.

compose/cli/main.py
dnephin: Ah right, because `docker-py` specifies `timeout=10` instead of `timeout=None`. That's unfortunate.

dnephin:
LGTM, I think we could also fix the `timeout=None` issue by setting it to our own default, instead of None, but it's fine like this.

dnephin:
This change will also allow us to speed up the integration test suite. Right now ~70s are spent waiting to stop containers (mostly in the `tests.integration.state_test` suite). If we can set this to 0.5s timeout, it should be a lot faster.

dnephin:
Thanks for the contribution!

This commit was already merged as part of another PR. It was either cherry-picked or rebased, so the sha changed, and github didn't realize it had been merged, but the commit is still yours (c24d5380e69a2b27a4c7b0c34ff05aac5794de26).

PR #1471, opened 2015-05-27
Script to prepare OSX build environment

aanand:
This will make sure we build on an up-to-date Python. This is relevant right now because `requests` logs `InsecurePlatformWarning` messages on Python <2.7.9.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -53,16 +53,20 @@ you can specify a test directory, file, module, class or method:
 
 ## Building binaries
 
-Linux:
+`script/build-linux` will build the Linux binary inside a Docker container:
 
     $ script/build-linux
 
-OS X:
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -53,17 +53,17 @@ you can specify a test directory, file, module, class or method:
 
 ## Building binaries
 
-Linux:
+`script/build-linux` will build the Linux binary inside a Docker container:
 
     $ script/build-linux
 
-OS X:
+On OS X, you should build inside a Mountain Lion VM for proper compatibility.
```
bfirsh: Might be worth pointing out that this is optional.... if you know the binary is going on only be run on Mavericks, then you can build it on Mavericks.

aanand: Yeah. Reworded.

bfirsh:
giphy-4

PR #1474, opened 2015-05-28
Fix SSL "CERTIFICATE_VERIFY_FAILED" error

aanand:
Python 2.7.9+ is very sensitive about the OpenSSL version used - this is the root cause of #890, which used to only come up for people who'd installed with Pip but is now happening in the Compose OSX binary, because we're now building it with Python 2.7.9 (which is necessary because otherwise `requests` will complain - rightly - that we're not making verified HTTPS connections).

In summary, the magic combination is **Python 2.7.9** and **OpenSSL 1.0.1**.

This PR changes two things:

1. `docker-compose --version` outputs Python and OpenSSL version information, as well as Compose's version.
2. `script/prepare-osx` makes sure that Python 2.7.9 and OpenSSL 1.0.1 are installed and activated in the Mac VM.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,9 +3,11 @@ FROM debian:wheezy
 RUN set -ex; \
     apt-get update -qq; \
     apt-get install -y \
-        python \
-        python-pip \
-        python-dev \
+        gcc \
+        make \
+        zlib1g \
+        zlib1g-dev \
+        libssl-dev \
         git \
         apt-transport-https \
         ca-certificates \
@@ -14,6 +16,37 @@ RUN set -ex; \
         iptables \
     ; \
     rm -rf /var/lib/apt/lists/*
+
+# Build Python 2.7.9 from source
+RUN set -ex; \
+    curl -LO https://www.python.org/ftp/python/2.7.9/Python-2.7.9.tgz; \
+    tar -xzf Python-2.7.9.tgz; \
+    cd Python-2.7.9; \
+    ./configure --enable-shared; \
+    make; \
+    make install; \
+    cd ..; \
+    rm -rf /Python-2.7.9; \
+    rm Python-2.7.9.tgz
+
+# Make libpython findable
+ENV LD_LIBRARY_PATH /usr/local/lib
+
+# Install setuptools
+RUN set -ex; \
+    curl -LO https://bootstrap.pypa.io/ez_setup.py; \
+    python ez_setup.py; \
+    rm ez_setup.py
+
+# Install pip
+RUN set -ex; \
+    curl -LO https://pypi.python.org/packages/source/p/pip/pip-7.0.1.tar.gz; \
+    tar -xzf pip-7.0.1.tar.gz; \
+    cd pip-7.0.1; \
+    python setup.py install; \
+    cd ..; \
+    rm -rf pip-7.0.1; \
+    rm pip-7.0.1.tar.gz
 
 ENV ALL_DOCKER_VERSIONS 1.6.0
 
--- a/compose/cli/utils.py
+++ b/compose/cli/utils.py
@@ -8,6 +8,7 @@ from docker import version as docker_py_version
 import os
 import platform
 import subprocess
+import ssl
 
 
 def yesno(prompt, default=None):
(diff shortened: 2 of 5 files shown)


fxdgear:
:+1:

fxdgear:
just to add I've verified the same with python 2.7.10 and OpenSSL 1.0.1j

aanand:
@fxdgear Interesting - Python 2.7.10 won't even import `ssl` for me with OpenSSL 1.0.1j:

```
$ python -V
Python 2.7.10

$ brew switch openssl 1.0.2a-1
$ python -c 'import ssl; print ssl.OPENSSL_VERSION'
OpenSSL 1.0.2a 19 Mar 2015

$ brew switch openssl 1.0.1j
$ python -c 'import ssl; print ssl.OPENSSL_VERSION'
  File "<string>", line 1, in <module>
  File "/usr/local/Cellar/python/2.7.10/Frameworks/Python.framework/Versions/2.7/lib/python2.7/ssl.py", line 97, in <module>
    import _ssl             # if we can't import it, let the error propagate
ImportError: dlopen(/usr/local/Cellar/python/2.7.10/Frameworks/Python.framework/Versions/2.7/lib/python2.7/lib-dynload/_ssl.so, 2): Symbol not found: _SSL_CTX_set_alpn_protos
  Referenced from: /usr/local/Cellar/python/2.7.10/Frameworks/Python.framework/Versions/2.7/lib/python2.7/lib-dynload/_ssl.so
  Expected in: /usr/local/opt/openssl/lib/libssl.1.0.0.dylib
 in /usr/local/Cellar/python/2.7.10/Frameworks/Python.framework/Versions/2.7/lib/python2.7/lib-dynload/_ssl.so
```

dnephin:
lgtm (but I know very little about osx and brew)

aanand:
Just pushed another commit to make sure we use Python 2.7.9 in the Docker image, so it's used to build the Linux binary. It means building Python from source, but it works.

aanand:
I've built some binaries with this fix. Please try them out if you've been experiencing SSL issues:

http://cl.ly/3W3a2S3t2c32/download/docker-compose-Linux-x86_64
http://cl.ly/0i00310l3x27/download/docker-compose-Darwin-x86_64

adambiggs:
I can confirm that the Darwin binary works.

steverobbins:
:+1: Darwin

Dockerfile
```diff
@@ -15,6 +17,37 @@ RUN set -ex; \
...
+RUN set -ex; \
+    curl -LO https://www.python.org/ftp/python/2.7.9/Python-2.7.9.tgz; \
+    tar -xzf Python-2.7.9.tgz; \
+    cd Python-2.7.9; \
+    ./configure --enable-shared; \
+    make; \
+    make install; \
+    cd ..; \
+    rm -rf /Python-2.7.9; \
+    rm Python-2.7.9.tgz
```
dnephin: Hmm, this added 4min to the jenkins build, which is unfortunate.

Could we use ubuntu vivid, which seems to have 2.7.9  (http://packages.ubuntu.com/vivid/python) ?

That way we could install from a package without having to build everything. I believe there was another PR for python3 support where @funkyfuture found that using ubuntu made things easier, but I forget the details of that now.

aanand: It is unfortunate, but hopefully it's temporary - if pyOpenSSL can fix its import-related oddities, hopefully it'll work with PyInstaller and we can go back to using the package.

The problem with using a newer distro is that it upgrades glibc, which is bad for compatibility with older distros.

dnephin:
LGTM

compose/cli/utils.py
```diff
@@ -132,6 +133,7 @@ def get_version_info(scope):
     elif scope == 'full':
         return versioninfo + '\n' \
             + "docker-py version: %s\n" % docker_py_version \
-            + "%s version: %s" % (platform.python_implementation(), platform.python_version())
+            + "%s version: %s\n" % (platform.python_implementation(), platform.python_version()) \
+            + "OpenSSL version: %s" % ssl.OPENSSL_VERSION
```
dnephin: This broke python 2.6 compatiblity

`AttributeError: 'module' object has no attribute 'OPENSSL_VERSION' `

which I guess is ok, we just need to remove it from tox.ini

aanand: Sounds good - I don't think we need to support Python 2.6.

funkyfuture: :clap: :clap: :clap:

yopeso:
The binary does not work in Linux although we have good certificate.

PR #1467, opened 2015-05-26
Change docker-compose kill integration tests

dnephin:
Attempt to fix #759

Lets see what jenkins thinks about these changes. They pass locally, but they basically invert the previous logic.

PR #1480, opened 2015-05-29
Change kill SIGINT test to use SIGSTOP

### LINKED ISSUE #759: fig kill integration tests fail locally on master
I'm seeing two test failures on master locally:

```
test_kill_interrupted_service (tests.integration.cli_test.CLITestCase)
test_kill_signal_sigint (tests.integration.cli_test.CLITestCase)
```

These were added recently #611 (9abdd337b5146643084205ad949c017b6ab843a4). For some reason they aren't failing in wercker.

I'm not sure how these tests pass. From what I can see the SIGINT should actually shutdown the container, but the tests expect the container to be running after a SIGINT.

Am I missing something?

bfirsh:
I think the original intention of the original test was the check that
different signals work, I think. This does this -- it sends a signal
that doesn't cause the container to stop.

Closes #759. Replaces #1467.

### DIFF
--- a/tests/integration/cli_test.py
+++ b/tests/integration/cli_test.py
@@ -360,22 +360,22 @@ class CLITestCase(DockerClientTestCase):
         self.assertEqual(len(service.containers(stopped=True)), 1)
         self.assertFalse(service.containers(stopped=True)[0].is_running)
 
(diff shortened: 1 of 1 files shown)


aanand:
LGTM

PR #1482, opened 2015-05-29
Make it possible to run tests remotely

bfirsh:
Don't mount code and make the build smaller. See commit comments for detail...


aanand:
LGTM

Issue #1377, opened 2015-04-30
Interpolate environment variables in docker-compose.yml


aanand:
(I'm creating a fresh issue for this as the [old one](https://github.com/docker/compose/issues/495) has accumulated rather a lot of baggage.)

It should be possible to pass in environment variables to the value of any* configuration entry in `docker-compose.yml`. A lot of people want to do it, it's good for portability and I'm satisfied it's not going to cause chaos.

I have some reckons.

### Required variables and optional defaults

It's useful to be able to specify that a variable that *must* be present in the environment, i.e. that Compose will refuse to run if it isn't. However, this will be a pain when you've got lots of them, so it should either be something you explicitly enable, or it should be possible to specify a default value.

The MVP implementation does not need to have either feature, but there should be a clear path to implementing both in a backwards-compatible way.

### Syntax

There's a strong case for implementing an established standard, as long as it's not heavyweight - our requirements for functionality are minimal.

-   [POSIX parameter expansion](http://pubs.opengroup.org/onlinepubs/009695399/utilities/xcu_chap02.html#tag_02_06_02) is OK. It has a few too many features, but we could implement a subset of them:

    - `${VARIABLE}` - outputs empty string if `VARIABLE` is unset
    - `${VARIABLE-default}` - outputs `default` if `VARIABLE` is unset
    - `${VARIABLE?}` - errors out if `VARIABLE` is unset

    https://github.com/docker/compose/pull/845 implemented a Bash-style `${VARIABLE:default}` syntax, which is similar to POSIX parameter expansion but slightly different.

-   [Python's format syntax](https://docs.python.org/2/library/string.html#formatstrings) would be trivial to implement, but is tied to the Python language, and has rather too many features.

### Implementation

Python's `os.path.expandvars` function implements the most basic case of POSIX parameter expansion:

```
>>> from os.path import expandvars
>>> expandvars('${HOME}')
'/Users/aanand'
```

However, it's got at least 2 problems:

1.  An unset variable doesn't expand to an empty string - instead, it results in no expansion:

    ```
    >>> expandvars('${UNSET}')
    '${UNSET}'
    ```

2.  Malformed syntax doesn't error out - instead, it also results in no expansion:

    ```
    >>> expandvars('${HOME')
    '${HOME'
    ```

[...]

PR #1488, opened 2015-05-30
Support reading config from stdin

dnephin:
Related to #1377

By supporting configuration from stdin, it becomes a lot easier for users with complex templating requirements to use a separate tool for templating, and pass the final config into docker-compose directly.

I wanted to ensure that the `if filename == '-'` check only happened in one place, so I had to refactor some of `compose.cli.command`.  `get_config_path` was moved to `compose.config` to support this.

### DIFF
--- a/compose/config.py
+++ b/compose/config.py
@@ -1,6 +1,12 @@
+import logging
 import os
+import sys
 import yaml
+from collections import namedtuple
+
 import six
+
+from compose.cli.utils import find_candidates_in_parent_dirs
 
 
 DOCKER_CONFIG_KEYS = [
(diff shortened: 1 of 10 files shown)


compose/config.py
```diff
@@ -63,12 +69,57 @@
...
+    if filename == '-':
+        return ConfigDetails(yaml.safe_load(sys.stdin), os.getcwd(), None)
+
+    filename = get_config_path(base_dir, filename)
+    return ConfigDetails(load_yaml(filename), os.path.dirname(filename), filename)
+
 
+def get_config_path(base_dir, file_path):
+    if file_path:
+        return os.path.join(base_dir, file_path)
```
aanand: I think I'd find this bit of logic less confusing if it were moved into the function above. At that point, I think the functions' responsibilities are a bit better-separated, and they could also be renamed. Say:

```python
def resolve_config_details(base_dir, filename):
    if filename == '-':
        return ConfigDetails(yaml.safe_load(sys.stdin), os.getcwd(), None)

    if filename is None:
        file_path = find_config_file(base_dir)
    else:
        file_path = os.path.join(base_dir, filename)

    return ConfigDetails(load_yaml(file_path), os.path.dirname(file_path), file_path)


def find_config_file(base_dir):
    (candidates, path) = find_candidates_in_parent_dirs(SUPPORTED_FILENAMES, base_dir)
    # ...
```

dnephin: I had just copied the old `get_config_path()` from `Command`, but now that we have another place to move it, I think that makes sense.

aanm:
So it won't be possible to use a compose-file and stdin at the same time? This would be interesting when people want to overwrite stuff from compose-file without creating a new file. Such as:
```bash
$cat my-compose.yml 
web:
  ports: 1234:5000
$ docker-compose up -f my-compose.yml --stdin-extends '{web: {ports: 5000:5000}}'
```
Which would result in running web with ports 5000:5000 instead of 1234.
(I know the format is wrong but you get the idea ;-) )

dnephin:
@aanm that's correct, the stdin would replace the file. I think this is pretty standard for unix utilities.  If you need something from a file, it would be up to you to read the original file, and merge it into the final config you send to stdin.

aanand:
@aanm What you describe is interesting (and partly covered by the `extends` option), but orthogonal to the idea of reading input from stdin - it'd apply just as much in the case of having two files, rather than one file + stdin.

@dnephin Come to think of it, what's the behaviour of `extends` in config read from stdin? Looks like it'll resolve paths relative to the current directory, but that should be documented somewhere.

dnephin:
True, I should add that with the small doc change I made about using `-` for file.

aanm:
@aanand I thinking that could some some users problems when they want to scale
something like:
 ```bash
count=0
while true;
let count=count+1
docker-compose up -f my-compose.yml --stdin-extends '{web: {ports: $count:5000}}'
do
```
My point is too use environment variables has input for configuration in compose. I think that would solve some problems that users are getting.

@dnephin Does docker itself has this feature that you are PR?

aanand:
@aanm For environment variable interpolation, see #1377.

dnephin:
Updated the docs, and cleaned up `get_config_path()`

dnephin:
Rebased again! Should be ready

aanand:
LGTM

PR #1489, opened 2015-05-30
Faster integration tests

dnephin:
Use labels to identify images and containers to remove, and cleanup after the run (instead of before). Also use a very short timeout for `up`

I also removed the kill/remove_stopped from some tests since that doesn't actually properly clean things up (and we won't want contributors to copy that pattern).

### DIFF
--- a/tests/integration/service_test.py
+++ b/tests/integration/service_test.py
@@ -235,7 +235,12 @@ class ServiceTest(DockerClientTestCase):
     def test_create_container_with_volumes_from(self):
         volume_service = self.create_service('data')
         volume_container_1 = volume_service.create_container()
-        volume_container_2 = Container.create(self.client, image='busybox:latest', command=["top"])
+        volume_container_2 = Container.create(
+            self.client,
+            image='busybox:latest',
+            command=["top"],
+            labels={LABEL_PROJECT: 'composetest'},
+        )
         host_service = self.create_service('host', volumes_from=[volume_service, volume_container_2])
         host_container = host_service.create_container()
         host_service.start_container(host_container)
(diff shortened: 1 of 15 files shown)


dnephin:
With the changes from @thieman integration tests are now ~70s (~40%) faster

thaJeztah:
> With the changes from @thieman integration tests are now ~70s (~40%) faster

Wow! Interested to hear; does filtering by label make a big difference? If it does, I can imagine that other docker projects can benefit from that trick as well!

dnephin:
The filtering by label did almost nothing for the jenkins run (I believe it's because of the docker-in-docker method it uses to run the suite).  On my desktop it makes a difference because I have a bunch of unrelated images and containers. I was seeing ~10% improvement with the labels change with ~300 images and about the same number of containers.

thaJeztah:
@dnephin thanks for the info, I don't think that's worth changing then in the other projects (at least, no huge timesaver)

tests/integration/service_test.py
```diff
@@ -245,72 +249,6 @@ def test_create_container_with_volumes_from(self):
...
-        )
-
-        old_container = create_and_start_container(service)
-        self.assertEqual(old_container.get('Volumes').keys(), ['/data'])
-        volume_path = old_container.get('Volumes')['/data']
-
-        new_container = service.converge()[0]
-        self.assertEqual(new_container.get('Volumes').keys(), ['/data'])
-        self.assertEqual(new_container.get('Volumes')['/data'], volume_path)
-
```
aanand: Is this logic still being tested properly anywhere?

dnephin: Isn't this just testing `converge()` which is a test-only helper?

I'll have to verify that there are tests for `execute_convergence_plan()`

aanand: It was originally testing the behaviour of `up`. Arguably the tests belong elsewhere - I just want to make sure that we're not discarding any important coverage.

dnephin: Ah , makes sense. I think I'll have to restore these and just rename them

dnephin:
Restored the tests for execute_convergence_plan

aanand:
LGTM

PR #1335, opened 2015-04-24
docker-compose create --readonly

chernjie:
mount the container's root filesystem as read only. Same as `--read-only`

```yaml
read_only: true
```

Signed-off: CJ <lim@chernjie.com>

PR #1491, opened 2015-05-31
Add support of --read-only docker flag

spiddy:
Add support of --read-only docker flag (#961)

### DIFF
--- a/compose/service.py
+++ b/compose/service.py
@@ -594,6 +594,7 @@ class Service(object):
         privileged = options.get('privileged', False)
         cap_add = options.get('cap_add', None)
         cap_drop = options.get('cap_drop', None)
+        read_only = options.get('read_only', None)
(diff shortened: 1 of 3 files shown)


dnephin:
I think you also need to add `read_only` to the config key lists in `config.py` in order to support it in the config

spiddy:
I think what you mention was already added in the master (https://github.com/docker/compose/blob/master/compose/config.py#L21)

I tested it using the following docker-compose.yml as example:

```
demo:
  image: ubuntu:14.04
  command: touch /hi
  read_only: true
```

```
➜  ~  docker-compose up  
Creating compose_demo_1...
Attaching to compose_demo_1
demo_1 | touch: cannot touch '/hi': Read-only file system
compose_demo_1 exited with code 1
```

Are you refering to a different use case or is this sufficient?

dnephin:
Oh ya, this support already exists in master! It was done in #1335

PR #1497, opened 2015-06-01
Run tests against Docker 1.7 RC2

aanand:
There are a few test regressions (not application regressions, thankfully). I've fixed some but not all.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -48,11 +48,13 @@ RUN set -ex; \
     rm -rf pip-7.0.1; \
     rm pip-7.0.1.tar.gz
 
-ENV ALL_DOCKER_VERSIONS 1.6.0
+ENV ALL_DOCKER_VERSIONS 1.6.0 1.7.0-rc2
 
 RUN set -ex; \
     curl https://get.docker.com/builds/Linux/x86_64/docker-1.6.0 -o /usr/local/bin/docker-1.6.0; \
-    chmod +x /usr/local/bin/docker-1.6.0
+    chmod +x /usr/local/bin/docker-1.6.0; \
+    curl https://test.docker.com/builds/Linux/x86_64/docker-1.7.0-rc2 -o /usr/local/bin/docker-1.7.0-rc2; \
+    chmod +x /usr/local/bin/docker-1.7.0-rc2
 
 # Set the default Docker to be run
 RUN ln -s /usr/local/bin/docker-1.6.0 /usr/local/bin/docker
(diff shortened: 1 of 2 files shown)


compose/container.py
```diff
@@ -115,6 +115,14 @@ def environment(self):
     def is_running(self):
         return self.get('State.Running')
 
+    @property
+    def cpu_set(self):
+        return self.get('HostConfig.CpusetCpus') or self.get('Config.Cpuset')
```
bfirsh: Is this a bug in the engine? Shouldn't the API should remain stable?

dnephin:
The kill tests are failing on jenkins the same way they were for me in #759 

Here's what I did to get it working on my laptop: https://github.com/docker/compose/compare/docker:master...dnephin:broken_kill_integration_tests

It does change the nature of the tests though

compose/container.py
aanand: Yes. Reported in https://github.com/docker/docker/issues/13663 and reverted my workarounds.

aanand:
@dnephin Seems to be a Docker 1.7 regression - whatever signal I send to the container (even, say, SIGWINCH), it exits. I'm investigating.

aanand:
Opened https://github.com/docker/docker/issues/13665 for the two kill test failures.

So, of the five failures:

- Two are caused by https://github.com/docker/docker/issues/13665
- Two are caused by https://github.com/docker/docker/issues/13663
- One is fixed by https://github.com/docker/docker/pull/13535, which should land in 1.7 RC2

aanand:
Update: docker/docker#13665 is fixed by docker/docker#13666. Two bugs down, one to go.

runcom:
Will take care of the last one really soon :)

runcom:
all good now, last bug was fixed @aanand

aanand:
Amazing, thanks. I've updated it to run against Docker master to confirm.

So that the build remains stable, though, I'll wait for RC2 and update it to run against that before merging.

vdemeester:
Ok tried out with 1.7.0-rc2 and it is working :wink: !

aanand:
All green against RC2 .

aanand:
Rebased.

dnephin:
LGTM

thaJeztah:
Just in time for 1.7-rc3, which will be pushed out later today :smile:

Moving to rc3 should hopefully be less problematic now that these issues have been solved.

Issue #1503, opened 2015-06-03
Inject chosen container name in container


octplane:
Because it can be desirable for monitoring, having the container name accessible in the container could be useful.

Use case include:
- monitoring the docker from outside
- monitoring the applications inside the docker
- wanting to merge the data with coherent data.

Currently, there is not way to get the container name apart from being outside the docker, or being able to talk to the docker daemon. This is clearly not something we want from out inner applications. I suggest passing a simple environment variable could do the trick in most case.

dnephin:
I think with the new networking support coming in the next release, you'll be able to use hostname for this instead of container name.

dnephin:
https://github.com/docker/compose/pull/2334#issuecomment-156276148 has another use case for this.

I think this may also address the issue raised in #1131

Using a predefined environment variable (as implemented in #1504) seems like a reasonable way to implement this. The only other option would be to set the hostname, but we saw that was disruptive to other use cases.

One potential problem is that containers can be renamed, but the environment variable would not get updated on a rename (there is no way to update the environment). So when it gets renamed, the value would be incorrect.

Another option might be to use `add-host` to add another line to `/etc/hosts`.  We won't know the container IP address yet, so we can't set it to that, but maybe there is some other alias that can be used?

I don't know of any other way we can set this in the container.

okgemaltoprg:
What about /etc/container_name (we already have debian_version, system-release, etc.). Anybody can grep/cat the value and it can be changed on the fly.

okgemaltoprg:
So, the final decision is just to use uid (hostname) and ignore the fact that there are use cases where it's simply not enough?

dnephin:
There hasn't been any decision made about this yet.

The case you describe in #2334 seems to be pretty rare, and maybe is more of an issue with Registrator being too limited about how it registers containers. Other container-dns solutions support ways to override the name.

> What about /etc/container_name

I don't think there are any facilities for adding files to the container at runtime, and we won't want to be adding an extra volume to every container.

The approach of using an environment variable breaks when containers are renamed, so we don't yet have a reliable alternative to using the unique hostname.

okgemaltoprg:
:/ I agree with you on the statements. However, I thought that Registrator+Consul isn't so rare :) I solved it already with custom microservice managing alternative hosts as I want it, but that's not perfect solution and I will start looking for alternatives soon.

> Other container-dns solutions support ways to override the name.

Could you help me and name some?

dnephin:
[...]

PR #1504, opened 2015-06-03
Fix for #1503

octplane:
(too naive ?) Fix for #1503 
- inject container name at creation in environment variables when
  available. The variable is name COMPOSE_CONTAINER_NAME

### DIFF
--- a/compose/service.py
+++ b/compose/service.py
@@ -217,6 +217,7 @@ class Service(object):
         )
 
         if 'name' in container_options:
+            container_options['environment'].setdefault('COMPOSE_CONTAINER_NAME', container_options['name'])
             log.info("Creating %s..." % container_options['name'])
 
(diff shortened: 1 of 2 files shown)


jippi:
+1

dnephin:
In the next release of docker and compose there will be support for network aliases backed by an embedded DNS server. This should allow you to query for the name of a container from dns instead of passing it in.

I think querying for the name is a more reliable solution because if the container name changes, you'll get the new name.  With an environment variable you'd get the old name.

Thanks for the PR, but I don't think we'll be able to accept it at this time, so I'm going to close it.

octplane:
np, I've had ample time to find solutions to this problem since last June.

varunchitturi:
> np, I've had ample time to find solutions to this problem since last June.

Could you please mention the solutions you have found.

PR #1521, opened 2015-06-06
Validate that service names passed to Project.containers aren't bogus.

dano:
The `Project.containers` method doesn't validate the list of service names passed to it, which leads to behavior like this:

```
dan@dandesk:~/test$ sudo docker-compose logs madeupnamesdfg
Attaching to 
```
The invalid service name just gets ignored. This behavior is seen with `ps` and `rm`, in addition to `logs`. However, to be consistent with all the other docker-compose commands, a `NoSuchService` exception should be raised instead.

With this patch, the previous example does just that:
```
dan@dandesk:~/test$ sudo docker-compose logs madeupnamesdfg
No such service: madeupnamesdfg
```
Let me know if this needs a unit test, or if you'd prefer something other than using `get_services` (like a dedicated `validate_service_names` method) to do the validation.

### DIFF
--- a/compose/project.py
+++ b/compose/project.py
@@ -274,6 +274,9 @@ class Project(object):
             service.remove_stopped(**options)
 
     def containers(self, service_names=None, stopped=False, one_off=False):
+        if service_names:
+            # Will raise NoSuchService if one of the names is invalid
+            self.get_services(service_names)
         containers = [
             Container.from_ps(self.client, container)
             for container in self.client.containers(
--- a/compose/project.py
+++ b/compose/project.py
@@ -99,6 +99,16 @@ class Project(object):
 
         raise NoSuchService(name)
 
+    def validate_service_names(self, service_names):
+        """
+        Validate that the given list of service names only contains valid
+        services. Raises NoSuchService if one of the names is invalid.
+        """
+        valid_names = self.service_names
+        for name in service_names:
+            if name not in valid_names:
+                raise NoSuchService(name)
+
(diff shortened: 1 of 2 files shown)


dano:
Ugh, sorry about the missed signature. What's the right way to fix that?

aanand:
Thanks! Could you add a test to `tests/integration/cli_test.py` that checks that an error is shown when you run `docker-compose logs madeupname`?

dano:
@aanand Done.

aanand:
LGTM

compose/project.py
```diff
@@ -274,6 +274,9 @@ def remove_stopped(self, service_names=None, **options):
             service.remove_stopped(**options)
 
     def containers(self, service_names=None, stopped=False, one_off=False):
+        if service_names:
+            # Will raise NoSuchService if one of the names is invalid
+            self.get_services(service_names)
```
dnephin: I guess this works, I'm just a little weary that we're using this for it's side effects, instead of using an interface specifically for the purpose of validating the service names. I feel like this is going to break or make it harder to refactor later on.  I'd like to investigate our options for exposing this in a more redirect way.

dano: Yeah, I made a comment along those lines in the pull request itself - it would be very easy to add a `validate_service_names` method that *just* makes sure the services actually exist, if you'd prefer.

dnephin:
LGTM

aanand:
LGTM

PR #1526, opened 2015-06-08
Remove Service.start_or_create_containers()

aanand:
It's used in one place, which is a test, and that test doesn't even need to start the container.


dnephin:
LGTM

PR #1527, opened 2015-06-08
Remove logging on run --rm

aanand:
`docker run --rm` doesn't log that it's removing the container, and neither should we.


dnephin:
LGTM

Issue #1483, opened 2015-05-29
docker-compose run --rm does not remove container


villlem:
`docker-compose run` version 1.3.0 RC1 does not remove container when called with `--rm` option. Those containers does not appear in `docker-compose ps` command either.

hason:
The same error for ``docker-compose 1.2.0``:

```
$ docker-compose run --rm composer composer install --ansi
```

```
$ docker-compose ps
Name   Command   State   Ports 
------------------------------
```

```
$ docker ps -a
CONTAINER ID        IMAGE                        COMMAND                CREATED              STATUS                          PORTS                      NAMES
50b8489c451a        my_composer:latest        "composer install --   About a minute ago   Exited (0) About a minute ago                              my_composer_run_11   
7cd7af8939e4        my_composer:latest        "composer install --   3 minutes ago        Exited (0) 3 minutes ago                                   my_composer_run_10   
```

josephpage:
Same error.
I will investigate soon.

vascofg:
+1

josephpage:
Seems that execution stops after dockerpty.start :  https://github.com/docker/compose/blob/master/compose/cli/main.py#L348
I have no idea why

pikeas:
+1. Minimal reproducible example: `docker run --rm my_service true` for any defined service.

```
~/D/c/test docker-compose run --rm client true
Creating test_client_run_1...
~/D/c/test docker-compose run --rm client true                                                                                                                                                                       Creating testj_client_run_2...
~/D/c/test docker ps -a                                                                                                                                                                                              CONTAINER ID        IMAGE                    COMMAND                CREATED             STATUS                     PORTS               NAMES
1bcc06f0a0f3        test_client:latest   "true"                 1 seconds ago       Exited (0) 1 seconds ago                       testj_client_run_2
8c2340786437        test_client:latest   "true"                 4 seconds ago       Exited (0) 3 seconds ago                       test_client_run_1

Happening with latest `master` branch.

villlem:
Looks like regression between dockerpty v0.3.2 and v0.3.3. If i install compose 1.2.0 and force dockerpty to 0.3.2, clients are removed as expected. I'm sorry but I have no time to explore this further :( maybe next week.

aanand:
Fix in https://github.com/d11wtq/dockerpty/pull/44.

villlem:
Problem fixed. After upgrade to dockerpty 0.3.4, containers are removed as expected. Many thanks.

[...]

PR #1529, opened 2015-06-08
Update dockerpty to 0.3.4

### LINKED ISSUE #1424: docker-compose --rm always returns 1
It doesn't matter how I kick it, it always returns 1 and my ansible task fails! However, without `--rm` it returns 0.

docker-compose.yml
```
test:
  image: debian:wheezy
```
Ubuntu 14.04.2 LTS
Docker version 1.6.2, build 7c8fca2
docker-compose 1.2.0
```
root@megabox:~# docker-compose run --rm test true
root@megabox:~# echo $?
1
```

I believe the exit code comes from `compose/cli/main.py:346`, however, `--rm` doesn't seem to affect the exit code so this doesn't make much sense...
```
    ¦   ¦   exit_code = container.wait()
    ¦   ¦   if options['--rm']:
    ¦   ¦   ¦   log.info("Removing %s..." % container.name)
    ¦   ¦   ¦   project.client.remove_container(container.id)
    ¦   ¦   sys.exit(exit_code)
```
How come it's always 1?
It should also be noted that the containers still show up in `docker ps -a` despite using `--rm`, so I'm guessing something is wrong.

I thought it might be apparmor at first so I disabled it `service apparmor teardown` but no change.

Any ideas? Thanks!

aanand:
Closes #1424 and #1483.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 PyYAML==3.10
 docker-py==1.2.2
-dockerpty==0.3.3
+dockerpty==0.3.4
 docopt==0.6.1
 requests==2.6.1
 six==1.7.3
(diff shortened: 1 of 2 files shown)


dnephin:
LGTM

PR #1531, opened 2015-06-09
Test that data volumes now survive a crash when recreating

### LINKED ISSUE #1418: Test recreate behaviour once we're using labels
Once #1356 is merged, `docker-compose up` should become more fault-tolerant.

Currently, if `up` crashes [between the renaming of the old container and the creation of the new one](https://github.com/docker/compose/blob/c8022457ebc94a50ee3acc41e8efa31f5181b839/compose/service.py#L247-L259), on its next invocation it will ignore the old container entirely, which can lead to loss of data if there were volumes attached.

However, once we're using labels - which aren't affected by the rename - it'll detect the old container and attempt to recreate it once again. The fact that it's been renamed no longer matters.

We should add a test for this (by deliberately making `recreate_container` abort in the right place), confirm that it doesn't work on master and that it does work once #1356 is merged.

aanand:
Closes #1418.

### DIFF
--- a/tests/integration/resilience_test.py
+++ b/tests/integration/resilience_test.py
@@ -1,0 +1,37 @@
+from __future__ import unicode_literals
+from __future__ import absolute_import
+
+import mock
+
+from compose.project import Project
+from .testcases import DockerClientTestCase
+
+
+class ResilienceTest(DockerClientTestCase):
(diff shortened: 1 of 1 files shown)


dnephin:
LGTM

PR #1540, opened 2015-06-12
Environment variable interpolation in docker-compose.yml

sebglazebrook:
Relating to https://github.com/docker/compose/issues/1377

I know there has been a couple of PR's around this of recent. Here's my two cents.

Note:

- This is my first time writing Python so I'm assuming my code is missing a lot of pythonisms.
- Also was going to do a bit more clean up and structure changes but I wanted someone to look over it first to let me know if I'm on the right path.

Any/all feedback welcome.

### DIFF
--- a/compose/config.py
+++ b/compose/config.py
@@ -469,21 +469,28 @@ def load_yaml(filename):
         raise ConfigurationError(six.text_type(e))
 
 
-class TemplateVariable(object):
-    def __init__(self, env_var_name, default_value=None, required=False):
-        self.env_var_name = env_var_name
-        self.default_value = default_value
-        self.required = required
+class ComposeTemplate(object):
 
-    def value(self):
-        result = os.environ.get(self.env_var_name, None)
-        if result:
-            return result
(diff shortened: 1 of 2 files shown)


aanand:
Thanks for making a start!

A few things:

- Variable expansion within strings is entirely uncoupled from Compose's duties, and can therefore live in a separate module (or, even better, a separate Python package entirely).

- A lot of the classes and class methods you've got can be reduced to simple top-level functions, which is much more Pythonic.

- I'm wary of regular expressions, as explained in #1377. We need to make sure it's robust and produces good error messages for malformed input.

It might be worth looking at / building upon @kojiromike's beginnings-of-an-implementation of POSIX PE: https://github.com/kojiromike/parameter_expansion

timfallmk:
FWIW this would likely solve #1553

funkyfuture:
@timfallmk what are YOU doing here? ;-)

dnephin:
This was solved with #1765

PR #1544, opened 2015-06-12
Fix volume binds de-duplication

### LINKED ISSUE #983: volume mapping does not work properly when host path is the same
when i try to create a container with a volume mapping that uses the same host path, its not working

    docker run --rm -ti --name test1 -v /var/run/docker.sock:/var/run/docker.sock -v /var/run/docker.sock:/tmp/sock -v /usr/bin/docker:/usr/bin/docker busybox /usr/bin/docker inspect --format='{{.Volumes}}' test1
gives me

    map[/tmp/sock:/run/docker.sock /usr/bin/docker:/usr/bin/docker /var/run/docker.sock:/run/docker.sock]

when i run this `fig.yml` via `fig -p fig up`:

    test2:
      image: busybox
      volumes:
        - /var/run/docker.sock:/tmp/sock
        - /usr/bin/docker:/usr/bin/docker
      command: /usr/bin/docker inspect --format={{.Volumes}} fig_test2_1
i get the following

    Creating fig_test2_1...
    Attaching to fig_test2_1
    test2_1 | map[/tmp/sock:/var/lib/docker/vfs/dir/20b2479134a353d63b754900fd9363fc704bdba3dd436dba70b96052457e7618 /usr/bin/docker:/usr/bin/docker /var/run/docker.sock:/run/docker.sock]
    fig_test2_1 exited with code 0
    Gracefully stopping... (press Ctrl+C again to force)

As you can see the first mapping with `/var/run/docker.sock` is ignored and only the latter is mapped.

aanand:
Volume binds are currently deduped based on the host path. They should be deduped based on the container path.

Closes #983 and also fixes a regression introduced by #858 where, if an existing container for a service reports a different host path for a volume than was supplied (e.g. by prepending `/mnt/sda1`, as sometimes happens), a binding for each host path is passed to the new container's config, which is an error.

Waiting on https://github.com/docker/docker-py/pull/635.

### DIFF
--- a/compose/service.py
+++ b/compose/service.py
@@ -731,7 +731,7 @@ def merge_volume_bindings(volumes_option, previous_container):
         volume_bindings.update(
             get_container_data_volumes(previous_container, volumes_option))
 
-    return volume_bindings
+    return volume_bindings.values()
 
 
 def get_container_data_volumes(container, volumes_option):
@@ -763,8 +763,7 @@ def get_container_data_volumes(container, volumes_option):
 
 
 def build_volume_binding(volume_spec):
-    internal = {'bind': volume_spec.internal, 'ro': volume_spec.mode == 'ro'}
-    return volume_spec.external, internal
+    return volume_spec.internal, "{}:{}:{}".format(*volume_spec)
 
 
 def parse_volume_spec(volume_config):
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 PyYAML==3.10
-docker-py==1.2.2
+docker-py==1.2.3-rc1
 dockerpty==0.3.4
 docopt==0.6.1
 requests==2.6.1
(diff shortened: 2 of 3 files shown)


aanand:
The docker-py fix is now out in 1.2.3-rc1, so I've updated `requirements.txt`. This PR is ready to go.

dnephin:
LGTM

pikeas:
Running compose from master, run is currently broken. utils:convert_volume_binds(binds) expects a dict, this commit replaces it with a list of values().

Let me know if I should open an issue instead.

PS - Seems like this should have been caught by basic integration testing...?

dnephin:
@pikeas You need to install the right version of docker-py, otherwise you'll hit that error.

I think we forgot to increase the minimum required version in setup.py to reflect that. It needs 1.2.3-rc1

aanand:
Good catch - I'll update `setup.py` in another PR.

aanand:
See #1550.

pikeas:
Also needs to resolve conflicting requirement for `requests<2.7,>=2.6.1` in compose.

aanand:
@pikeas What does that conflict with?

pikeas:
@aanand Hm...disregard, broken when manually updating docker-py (before this PR was merged), fine reinstalling from scratch just now.

PR #1545, opened 2015-06-14
Updated for new documentation tooling

moxiegirl:
Updating with changes
Updating for Hugo
Adding a README'
moving index.md compose-overview.md in links

ping @aanand @ehazlett This is the tooling for documentation intended for 1.7.  Can I get an LGTM please?


moxiegirl:
### The "docs RC plan"

This is the first step to an "docs RC build" if you will.  

After I pull this PR and the other 6, I'm going to do a build to a staging bucket on S3.  That build will include the new theme which is dropping around noon Monday. I'll page through every page on the RC build:

- look for layout issues
- checking all the links
- making sure images appear

Each problem will be noted and then I'll do a set of follow on PRs to fix issues.  That's why I'm wanting to pull this set today, it allows for an RC cycle.  And then, also, if necessary one more before the release this week.

thaJeztah:
ping @dnephin and @bfirsh as well for new docs tooling

ehazlett:
This looks OK to me but I'm not that familiar with the code base.  I'll defer to @bfirsh and @aanand.

aanand:
LGTM

PR #1550, opened 2015-06-15
Update setup.py with new docker-py minimum

aanand:
I updated `requirements.txt` in #1544, but forgot to update `setup.py`.


dnephin:
LGTM

Issue #1317, opened 2015-04-21
Add option for self-signed certificates for DOCKER_HOST over TLS


dekobon:
With docker you can specify a Docker host that listens exclusively with TLS. Then you can use the parameter --tls and the environment variable DOCKER_CERT_PATH to connect to a Docker host with self-signed certificates.

However, when using Docker compose, you can't connect to a host using self-signed certs because the only option to enable TLS based Docker hosts is DOCKER_TLS_VERIFY and that environment variable has TLS verify baked into it. I need something more like a vanilla DOCKER_TLS environment variable or a CLI parameter that is equivalent of "docker --tls"

eviking:
+1

jgrowl:
+1

dnephin:
compose supports http://docs.docker.com/compose/reference/overview/#docker-cert-path now

PR #1569, opened 2015-06-18
Adds additional TLS configuration CLI arguments

jgrowl:
A few notes: :notes: 

- I am ignoring urllib3 InsecureRequestWarning as it gets displayed for every request and is rather spammy. I think it makes sense to show a custom error once instead since the user has to explicitly set verification off. I am using the same text from urllib3's InsecureRequestWarning, but it might make sense to modify it for compose.

- --tls and --tlsverify arguments can bet set equal to a few values that docker seems to support: true, t, 1. 

- --tls argument without any assignment does not work the way it does with docker. I could not figure out how to achieve that behavior with docopt.

See #1317 #977

### DIFF
--- a/compose/cli/command.py
+++ b/compose/cli/command.py
@@ -61,12 +61,17 @@ class Command(DocoptCommand):
         project = self.get_project(
             self.get_config_path(explicit_config_path),
             project_name=options.get('--project-name'),
+            tls=options.get('--tls') in ("true", "t", "1"),
+            tls_ca_cert=options.get('--tlscacert'),
+            tls_cert=options.get('--tlscert'),
+            tls_key=options.get('--tlskey'),
+            tls_verify=options.get('--tlsverify') in ("true", "t", "1"),
             verbose=options.get('--verbose'))
 
         handler(project, command_options)
 
-    def get_client(self, verbose=False):
-        client = docker_client()
+    def get_client(self, tls=False, tls_ca_cert=None, tls_cert=None, tls_key=None, tls_verify=False, verbose=False):
+        client = docker_client(tls, tls_ca_cert, tls_cert, tls_key, tls_verify)
         if verbose:
             version_info = six.iteritems(client.version())
             log.info("Compose version %s", __version__)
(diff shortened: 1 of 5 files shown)


ekristen:
Can we get this updated and merged? Docker Compose needs to support TLS without client authentication. Thanks!

compose/cli/command.py
```diff
@@ -61,12 +61,17 @@ def perform_command(self, options, handler, command_options):
         project = self.get_project(
             self.get_config_path(explicit_config_path),
             project_name=options.get('--project-name'),
+            tls=options.get('--tls'),
+            tls_ca_cert=options.get('--tlscacert'),
+            tls_cert=options.get('--tlscert'),
+            tls_key=options.get('--tlskey'),
+            tls_verify=options.get('--tlsverify'),
```
dnephin: Instead of passing all of these around as separate kwargs I think we should use a namedtuple and pass around a single object:

```
TLSOptions = namedtuple('TLSOptions', 'enabled ca_cert cert key verify')
```

compose/cli/docopt_command.py
```diff
@@ -21,7 +24,9 @@ def sys_dispatch(self):
         self.dispatch(sys.argv[1:], None)
 
     def dispatch(self, argv, global_options):
-        self.perform_command(*self.parse(argv, global_options))
+        with warnings.catch_warnings():
+            warnings.filterwarnings(action='ignore', category=InsecureRequestWarning)
+            self.perform_command(*self.parse(argv, global_options))
```
dnephin: I'm not sure about this. I think this warning can be raised in other situations (when the host doesn't have the proper certs package installed) and we'd be masking those errors. 

I think I'd prefer to just keep warning.

dnephin:
Thanks for the contribution!

Needs a rebase, and I've made a couple comments

dnephin:
Oh, and docs

compose/cli/docker_client.py
```diff
@@ -16,19 +18,31 @@ def docker_client():
...
+
+        tls_config = docker_tls.TLSConfig(
             ssl_version=ssl.PROTOCOL_TLSv1,
-            verify=True,
+            verify=tls_verify,
             assert_hostname=False,
-            client_cert=client_cert,
-            ca_cert=ca_cert,
+            client_cert=(tls_cert, tls_key),
+            ca_cert=tls_ca_cert,
```
aanand: See https://github.com/docker/compose/pull/1744 - we want to use docker-py's `kwargs_from_env` rather than doing all this logic ourselves. That might well mean making changes to docker-py.

dnephin:
Thanks for the PR!

I'm going to close this due to lack of activity. If you're still interesting in contributing please rebase, and re-open this PR.

ekristen:
@dnephin can we just get an official maintainer to implement this? This is a much needed feature.

dnephin:
I created #2670 to track it. I believe we already support this from environment variables, don't we?

PR #1579, opened 2015-06-19
Don't set network mode when none is specified

bfirsh:
Setting a value overrides the new default network option.


aanand:
Opened #1580 with a unit test.

PR #1581, opened 2015-06-19
Add env/cli cross references in cli.md

Chouser:
Just noticed that there are multiple ways to set the same options, but the docs don't make this clear. Hope this helps.


dnephin:
Thanks for the contribution!

I think this would be great to include.  Would you mind adding a link to the documents as well? 

You'll need to rebase on master as well.

mnowster:
Closing in favour of https://github.com/docker/compose/pull/2273

PR #1584, opened 2015-06-21
Added timeout option on scale down action

rvichery:
As this is already supported on up, stop and restart actions, I think It could be great to have this feature on a scale down action as well!

Thanks!


dnephin:
Thanks for the contribution! Looks like this was added already in (9d0bbdf8dd503ff149409d4f8cca18852df9dea7)

PR #1586, opened 2015-06-21
Replace backtick code blocks with indentation

### LINKED ISSUE #1583: Remove ``` method from code samples in docs subdirectory
GitHub-Flavored-Markdown supports the ``` method. Hugo doesn't handle GFM. Hugo delegate all its markdown rendering to Blackfriday (a Go lib). https://github.com/russross/blackfriday34
So, they render as lines. Using indent is the correct method. 

We need to switch to the indent method in the compose docs.  The docs.docker.com files are using https://highlightjs.org/ which supports a lot of formats including YAML and Dockerfiles.

Closes https://github.com/docker/docs-base/issues/41

aanand:
Closes #1583 

ping @moxiegirl

### DIFF
--- a/docs/extends.md
+++ b/docs/extends.md
@@ -28,25 +28,21 @@ the configuration around.
 When defining any service in `docker-compose.yml`, you can declare that you are
 extending another service like this:
 
-```yaml
-web:
-  extends:
-    file: common-services.yml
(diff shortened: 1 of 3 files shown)


moxiegirl:
LGTM

PR #1603, opened 2015-06-25
Add --help to bash completion

albers:
`--help` works for all commands. But it is not mentioned in the individual help messages, probably intentionally to save screen space. 
As `--help` is very helpful (pun), I added support for it to bash completion.


aanand:
Nice. We should add it to zsh completion as well (ping @sdurrheimer)

LGTM

dnephin:
LGTM

Issue #1519, opened 2015-06-05
Relax restrictions on service names


aanand:
Now that we're not relying on a strict container name format, we should be able to allow more characters in service names, as has often been requested (see #799, #869, #941, #1300, #1384).

- Docker allows `[a-zA-Z0-9_.-]` ([daemon.go](https://github.com/docker/docker/blob/04c6f09fbdf60c7765cc4cb78883faaa9d971fa5/daemon/daemon.go#L56))

- Compose allows just `[a-zA-Z0-9]` ([service.py](https://github.com/docker/compose/blob/7d2a89427c59774a8cbf503a57cb9f3b0d47d1fe/compose/service.py#L49))

Any good reason we can't allow `_.-` now?

thaJeztah:
:+1: I'll finally be able to build my 

`a_.--.__.--.__.--.__.--.__.--.__.--.__.--.__.--._` and `127.0.0.1` images :D

aanand:
Also relevant: allowed characters in image tags, since those are also based on service name. Seems to be the same chars though ([tags.go](https://github.com/docker/docker/blob/04c6f09fbdf60c7765cc4cb78883faaa9d971fa5/graph/tags/tags.go)).

erichonkanen:
+1

zstarer:
+1 just ran into this after a long day. that's fun. container name: -_-

broegner:
Is there anything new here? Will this change be part of the next RC or is it already baken into the current one?

aanand:
1.3 is in code freeze, so this will not be in the next release.

phazei:
Is there anything happening with support on this? 
In this comment https://github.com/docker/compose/issues/472#issuecomment-91252147 it's mentioned that there should be some support for this in 1.3.

I'm using 1.3 and it doesn't seem to be any different.  Will there be a 1.3.1 or will we need to wait till 1.4 till anything other than [a-z0-9] is allowed for reasons of service/container/host compatibility/bug?

aanand:
This won't be in 1.3, or any 1.3.x release. 1.4 is the likely target.

aanand:
Closed by #1624.

TristanCP:
"Any good reason we can't allow _.- now?"

There is a good reason for "_", since service names are used as host names in URLs :  

"Note that while upper and lower case letters are allowed in domain
names, no significance is attached to the case.  That is, two names with
the same spelling but different case are to be treated as if identical.

[...]

PR #1624, opened 2015-07-01
1519 relax service name restrictions

mnowster:
Allow _.- characters to be used in Service and Project names, as requested per this issue: https://github.com/docker/compose/issues/1519

### DIFF
--- a/compose/service.py
+++ b/compose/service.py
@@ -47,7 +47,7 @@ DOCKER_START_KEYS = [
     'security_opt',
 ]
 
-VALID_NAME_CHARS = '[a-zA-Z0-9]'
+VALID_NAME_CHARS = '[a-zA-Z0-9\._\-]'
 
 
 class BuildError(Exception):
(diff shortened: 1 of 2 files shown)


dnephin:
LGTM

aanand:
LGTM

thaJeztah:
@aanand will this require a migration path? e.g., what happens if I currently have a project started from a directory named `example.com`, and I upgrade Docker Compose? Will it still be able to correlate the running containers with the project?

(Or is the automatic generation of project names, based in the project-directory not changed)

aanand:
@thaJeztah Good point, but that hasn't been changed - it's handled by `normalize_name()` in [command.py](https://github.com/docker/compose/blob/ae96fc0071f4b899e4a95023a80c103468646350/compose/cli/command.py#L85). If we ever *do* change it, we'll need a migration path in place.

thaJeztah:
@aanand :+1: wasn't sure here, so thought I'd better check

marcuslinke:
> @thaJeztah Good point, but that hasn't been changed - it's handled by normalize_name() in command.py. If we ever do change it, we'll need a migration path in place.

@aanand As I see in current master `normalize_name()` is called in each case. So is there currently any chance to use a project name containing dashes? If not then #1519 should be reopened.

PR #1632, opened 2015-07-02
Extends file default behaviour and fixes circular reference bug

mnowster:
If no 'file' key is specified in the extends_options then by default we will look for the service within the same file.

This also includes a fix for circular reference checking which wasn't working entirely as desired.

### DIFF
--- a/compose/config.py
+++ b/compose/config.py
@@ -158,7 +158,7 @@ class ServiceLoader(object):
         if 'extends' not in service_dict:
             return service_dict
 
-        extends_options = process_extends_options(service_dict['name'], service_dict['extends'])
+        extends_options = validate_extends_options(service_dict['name'], service_dict['extends'])
 
         if self.working_dir is None:
             raise Exception("No working_dir passed to ServiceLoader()")
@@ -190,7 +190,7 @@ class ServiceLoader(object):
         return (self.filename, name)
 
 
-def process_extends_options(service_name, extends_options):
+def validate_extends_options(service_name, extends_options):
     error_prefix = "Invalid 'extends' configuration for %s:" % service_name
 
     if not isinstance(extends_options, dict):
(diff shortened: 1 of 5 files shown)


dnephin:
What's the story here? I think this would be good to add. Did it get scraped for a new PR?

aanand:
It needs more work - we were pairing on it - but was closed in error.

aanand:
LGTM

dnephin:
I like it, could be squashed down to 1 or 2 commits I think

mnowster:
@dnephin How strongly is the squash feeling? I don't feel comfortable going for squashing it down as there's a lot going on in this PR. Started out as a feature and then discovered and needed to include fixing some circular bugs, which I then couldn't do without some refactoring. The commits help communicate the why for these decisions.

This may also be because I'm newer to this project and squashing confidence comes later.

dnephin:
@mnowster I'll defer to @aanand. I don't have strong feelings about it, it's just been the convention with compose to squash down to a few commits. I've always been asked to do the same :)

mnowster:
@dnephin ok, cool cool. Maybe that's something I could pair with @aanand on, so I don't mess it up :) thanks for reviewing the PR!

aanand:
There's a strong case to be made for making it 3 commits (2 bugfixes + 1 feature) but I don't think it'd be worth the time involved, and the current sequence of commits contains good information, so I think it's fine.

bfirsh:
Does this need docs for the new behaviour?

mnowster:
@bfirsh yes, good spot, thank you. I have updated the docs here: https://github.com/docker/compose/pull/1666

ekohl:
For future reference, if you use ```Closes #issue``` or ```Fixes #issue``` then github will automatically close #issue when it's merged.

mnowster:
@ekohl I had forgotten that, thanks for the reminder :balloon:

PR #1633, opened 2015-07-02
Added env var parsing to volume container paths

### LINKED ISSUE #1626: Docker Compose parses host paths for ENV substitution, but not container paths
[This comment](https://github.com/docker/compose/issues/551#issuecomment-97411619) in #551 indicates that the following syntax should work for passing SSH_AUTH_SOCK into a container:

```
volumes:
    - $SSH_AUTH_SOCK:$SSH_AUTH_SOCK
environment:
    - SSH_AUTH_SOCK
```

However, [resolve_host_path in config.py](https://github.com/docker/compose/blob/master/compose/config.py#L355) does exactly as it's named, it resolves the host path but not the container path.

This means that the snippet pasted above results in a volume like this:

`['/tmp/ssh-4NokVWqj63/agent.10499:$SSH_AUTH_SOCK']`

Instead of resolve_host_paths/resolve_host_path, the functions should probably be resolve_path/resolve_paths and resolve both host and container paths.

jeffk:
This commit adds environment variable parsing to the container side of the volume mapping in configs.  Currently only host side paths are evaluated for environment variable substitution. This commit fixes #1626.

The common use case for this is mounting SSH agent sockets in a container, but it's useful anywhere you may want to have external variables effect volume paths. An example of mounting an SSH agent socket, which currently fails, due to the host side not being evaluated: 

```
volumes:
    - $SSH_AUTH_SOCK:$SSH_AUTH_SOCK
environment:
    - SSH_AUTH_SOCK
```

### DIFF
--- a/compose/config.py
+++ b/compose/config.py
@@ -184,7 +184,7 @@ def process_container_options(service_dict, working_dir=None):
     service_dict = service_dict.copy()
 
     if 'volumes' in service_dict:
-        service_dict['volumes'] = resolve_host_paths(service_dict['volumes'], working_dir=working_dir)
+        service_dict['volumes'] = resolve_volume_paths(service_dict['volumes'], working_dir=working_dir)
 
     if 'build' in service_dict:
         service_dict['build'] = resolve_build_path(service_dict['build'], working_dir=working_dir)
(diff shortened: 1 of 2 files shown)


dnephin:
While I don't think it's absolutely necessarily for this case (as I commented in the issue), I don't think it hurts to add this until we figure out #1377 , and the extra test is good to have as well.

LGTM

aanand:
LGTM

thaJeztah:
Thanks for contributing @jeffk !

PR #1637, opened 2015-07-03
Fix a unit test that was failing occasionally

dnephin:
I'm not sure how this ever worked, but it was failing occasionally while I was running unit tests locally.  The `working_dir` is actually a kwarg to this function, not part of the dict.


aanand:
Good catch. Needs a rebase.

dnephin:
Oh, apparently this was also fixed in one of the recently merged branches. Going to close this.

PR #1643, opened 2015-07-03
Show an error on 'run' when there are legacy one-off containers

### LINKED ISSUE #1609: Running one-off containers does not work in 1.3, "is already in use"
I used to run tests in docker by running one-off containers, but that is not possible in 1.3.

When i try i says
```
$ docker-compose run backend npm test
Conflict. The name "app_backend_run_1" is already in use by container e46cb8f520ad. You have to delete (or rename) that container to be able to reuse that name.
```

Same commando that I used before.

I have tried running `docker-compose up` and also `docker-compose migrate-to-labels`

How do you run one-off containers, like the docs propose
https://github.com/docker/compose/blob/master/docs/cli.md#run

aanand:
Also warn the user about the one-off containers in the standard error message about legacy containers.

Closes #1609.

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -34,7 +34,7 @@ def main():
     except KeyboardInterrupt:
         log.error("\nAborting.")
         sys.exit(1)
-    except (UserError, NoSuchService, ConfigurationError, legacy.LegacyContainersError) as e:
+    except (UserError, NoSuchService, ConfigurationError, legacy.LegacyError) as e:
         log.error(e.msg)
         sys.exit(1)
     except NoSuchCommand as e:
@@ -336,12 +336,22 @@ class TopLevelCommand(Command):
         if not options['--service-ports']:
             container_options['ports'] = []
 
-        container = service.create_container(
-            quiet=True,
-            one_off=True,
-            insecure_registry=insecure_registry,
-            **container_options
-        )
+        try:
+            container = service.create_container(
+                quiet=True,
+                one_off=True,
+                insecure_registry=insecure_registry,
+                **container_options
+            )
+        except APIError as e:
+            legacy.check_for_legacy_containers(
+                project.client,
+                project.name,
+                [service.name],
+                allow_one_off=False,
+            )
+
+            raise e
 
         if options['-d']:
             service.start_container(container)
--- a/compose/legacy.py
+++ b/compose/legacy.py
@@ -1,6 +1,7 @@
 import logging
 import re
 
+from .const import LABEL_VERSION
 from .container import get_container_name, Container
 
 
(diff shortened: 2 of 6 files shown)


compose/legacy.py
```diff
@@ -86,34 +126,66 @@ def migrate_project_to_labels(project):
...
+                continue
 
+            container = Container.from_ps(client, container)
+
+            if is_one_off:
+                legacy_one_off_containers.append(container)
+            else:
+                legacy_containers.append(container)
+
+    return (legacy_containers, legacy_one_off_containers)
```
dnephin: I think this implementation could be a little simpler.  Instead of always returning both, it could take a `one_off=False` kwarg (like a bunch of functions do currently).

If it's True return only the one_off containers, of it's False only the non-one_off containers.

That removes the need for the tuple return in `check_name`, `validate_name`, and this function.

I removes the unused return value from `migrate_project_to_labels`, and the logic in `check_for_legacy_containers` will be about the same.

compose/legacy.py
```diff
@@ -24,41 +25,81 @@
...
-        raise LegacyContainersError([c.name for c in containers])
+        raise LegacyContainersError(
+            [c.name for c in containers],
+            [c.name for c in one_off_containers],
+        )
+
+    if one_off_containers and not allow_one_off:
+        raise LegacyOneOffContainersError(
+            [c.name for c in one_off_containers],
+        )
```
dnephin: With the suggested edit to `get_legacy_containers`, I think this would be

```python

containers = get_legacy_containers(..., one_off=False)
if containers:
    raise ...

if not allow_one_off and get_legacy_containers(..., one_off=True):
    riase ...
```

compose/legacy.py
dnephin: I guess the disadvantage there is the extra query for containers, which is relatively slow, but we're already adding one slow query to `docker-compose run` either way

compose/cli/main.py
```diff
@@ -292,6 +292,13 @@ def run(self, project, options):
         """
         service = project.get_service(options['SERVICE'])
 
+        legacy.check_for_legacy_containers(
+            project.client,
+            project.name,
+            [service.name],
+            allow_one_off=False,
+        )
```
dnephin: Since this is not a one-time cost, it might be nice to provide a `--skip-legacy-check` for users that know their containers are gone, and don't want to incur the extra overhead of the "list all containers" api call.

compose/legacy.py
aanand: Yeah - that was why I did it this way, but it does add a fair bit of complexity. As you say, I don't think it's worth the performance savings in the end. Will redo.

compose/cli/main.py
aanand: Hm, good point - this will continue to be slow post-migration.

aanand:
OK, I've made the code changes you suggested, and also changed the behaviour of `docker-compose run` so that it only does the check if it fails to create the container, so performance in the normal case isn't affected.

dnephin:
LGTM

PR #1644, opened 2015-07-03
Stop 'rm' and 'ps' listing services not defined in the current file

### LINKED ISSUE #1596: `rm` reports that it will delete services that aren't defined in the current file
Hello, 
I have a `data.yml` that declares various data containers that should not be deleted in normal circumstances.

In the main `docker-compose.yml` these containers are referenced by name in `volumes_from`.

With this setup under 1.2 `docker-compose rm` deletes only containers from `docker-compose.yml`. 
Since 1.3 it tries to delete also the data containers, and this breaks my deploy.

I think that the yml file should be used to restrict the containers found by the label lookup, unless some specific command line option is given.

As a workaround, is there another way to prevent deletion of data containers?

aanand:
If you have an alternate YAML file with different services defined, containers for those services will be shown in `docker-compose ps` even if you don't pass that file in.

Furthermore, `docker-compose rm` will claim that it's going to remove them, but actually won't.

Closes #1596.

### DIFF
--- a/compose/project.py
+++ b/compose/project.py
@@ -288,6 +288,9 @@ class Project(object):
     def containers(self, service_names=None, stopped=False, one_off=False):
         if service_names:
             self.validate_service_names(service_names)
+        else:
+            service_names = self.service_names
+
         containers = [
             Container.from_ps(self.client, container)
             for container in self.client.containers(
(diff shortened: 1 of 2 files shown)


dnephin:
Interesting, I hadn't considered this case. 

With this change, the canonical source of truth for "what is in this composition" changes from "any container with the project label" to "containers matching both the project label, and a service label in the active config".

It seems a little strange to me that you'd have services with the same project name, and they wouldn't be considered part of the same composition, but I guess this is the behaviour in docker-compose <1.3.

Might be worth considering if the "any container with the project label" definition makes sense long-term, but for now,

LGTM

aanand:
Yeah, I agree that we should step back at some point and figure out what logic we really want, but right now it's a regression.

Issue #1641, opened 2015-07-03
docker compose doesn't (always) print error


haf:
E.g.

```
﻿﻿Compose version 1.3.1
Docker base_url: http://172.17.8.100:2375
Docker version: KernelVersion=4.0.5, Arch=amd64, ApiVersion=1.18, Version=1.6.2, GitCommit=7c8fca2-dirty, Os=linux, GoVersion=go1.4.2
docker containers <- (all=True, filters={u'label': [u'com.docker.compose.project=heka', u'com.docker.compose.service=heka', u'com.docker.compose.oneoff=False']})
docker containers -> (list with 1 items)
Recreating heka_heka_1...
docker stop <- (u'6db7f0825056efa25ac888849e4a678c7e3b8a300c07963f9743d8983431718d')
docker stop -> None
docker rename <- (u'6db7f0825056efa25ac888849e4a678c7e3b8a300c07963f9743d8983431718d', u'6db7f08250_heka_heka_1')
docker rename -> None
docker inspect_container <- (u'6db7f0825056efa25ac888849e4a678c7e3b8a300c07963f9743d8983431718d')
docker inspect_container -> {u'AppArmorProfile': u'',
 u'Args': [u'--config', u'/etc/hekad'],
 u'Config': {u'AttachStderr': False,
             u'AttachStdin': False,
             u'AttachStdout': False,
             u'Cmd': [u'hekad', u'--config', u'/etc/hekad'],
             u'CpuShares': 0,
             u'Cpuset': u'',
             u'Domainname': u'',
             u'Entrypoint': None,
...
docker inspect_image <- ('haaf/heka:latest')
docker inspect_image -> {u'Architecture': u'amd64',
 u'Author': u'',
 u'Comment': u'',
 u'Config': {u'AttachStderr': False,
             u'AttachStdin': False,
             u'AttachStdout': False,
             u'Cmd': [u'hekad', u'--config', u'/etc/hekad'],
             u'CpuShares': 0,
             u'Cpuset': u'',
             u'Domainname': u'',
...
docker inspect_image <- ('haaf/heka:latest')
docker inspect_image -> {u'Architecture': u'amd64',
 u'Author': u'',
 u'Comment': u'',
 u'Config': {u'AttachStderr': False,
             u'AttachStdin': False,
             u'AttachStdout': False,
             u'Cmd': [u'hekad', u'--config', u'/etc/hekad'],
             u'CpuShares': 0,
             u'Cpuset': u'',
[...]

Issue #1176, opened 2015-03-25
Compose 'up' command fails after second call


jlesquembre:
With this simple .yml file:

```yaml
web:
  image: alpine:3.1
  command: echo "Hello World"
```

First time you run `docker-compose up`, everything works as expected:

```
Creating test_web_1...
Attaching to test_web_1
web_1 | Hello World
test_web_1 exited with code 0
Gracefully stopping... (press Ctrl+C again to force)
```

But second time, compose doesn't display any output and never exits:

```
Recreating test_web_1...
Attaching to
```

Probably related to #753

dnephin:
Yup, duplicate of #753

dnephin:
Going to track this issue in #753

PR #1645, opened 2015-07-03
Fix race condition in docker-compose up

### LINKED ISSUE #753: Fig does not exit if container stops before attaching
When running fig up, if a container stops before attaching it seems like Fig will only attach to the containers that are left running. The output from the stopped container is also not displayed.

```
$ cat > fig.yml
web:
  build: ubuntu
  command: echo hello world
db:
  image: postgres:9.1
$ fig up
Creating example_web_1...
Creating example_db_1...
Attaching to example_db_1
db_1 | The files belonging to this database system will be owned by user "postgres".
db_1 | This user must also own the server process.
...
```

dnephin:
This feels really familiar, and I could have sworn we had fixed this already, but I can't find the PR.

Resolves #753, #1176, #1641

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -453,7 +453,7 @@ class TopLevelCommand(Command):
         service_names = options['SERVICE']
         timeout = float(options.get('--timeout') or DEFAULT_TIMEOUT)
 
-        project.up(
+        to_attach = project.up(
             service_names=service_names,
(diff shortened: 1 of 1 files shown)


aanand:
I was also sure we'd fixed it, but clearly not!

aanand:
LGTM

Issue #1561, opened 2015-06-17
x-smart-recreate crashes when modified image name/tag


josephpage:
Just got this error with docker-compose 1.3.0rc2 :

```
$ docker-compose up --x-smart-recreate
Traceback (most recent call last):
  File "<string>", line 3, in <module>
  File "/code/build/docker-compose/out00-PYZ.pyz/compose.cli.main", line 31, in main
  File "/code/build/docker-compose/out00-PYZ.pyz/compose.cli.docopt_command", line 21, in sys_dispatch
  File "/code/build/docker-compose/out00-PYZ.pyz/compose.cli.command", line 34, in dispatch
  File "/code/build/docker-compose/out00-PYZ.pyz/compose.cli.docopt_command", line 24, in dispatch
  File "/code/build/docker-compose/out00-PYZ.pyz/compose.cli.command", line 66, in perform_command
  File "/code/build/docker-compose/out00-PYZ.pyz/compose.cli.main", line 469, in up
  File "/code/build/docker-compose/out00-PYZ.pyz/compose.project", line 221, in up
  File "/code/build/docker-compose/out00-PYZ.pyz/compose.project", line 261, in _get_convergence_plans
  File "/code/build/docker-compose/out00-PYZ.pyz/compose.service", line 285, in convergence_plan
  File "/code/build/docker-compose/out00-PYZ.pyz/compose.service", line 299, in _containers_have_diverged
  File "/code/build/docker-compose/out00-PYZ.pyz/compose.service", line 415, in config_hash
  File "/code/build/docker-compose/out00-PYZ.pyz/compose.service", line 420, in config_dict
TypeError: 'NoneType' object has no attribute '__getitem__'
```

Resolved after `docker-compose pull`

aanand:
Can you provide more information please? What sequence of commands did you type? What does your YAML look like? Does the problem only occur if you use `--x-smart-recreate`?

josephpage:
Simple case : 

```
$ cat <<EOF > docker-compose.yml
example:
  image: debian:latest
  command: echo "hello world"
EOF

$ docker-compose up -d --x-smart-recreate
Pulling example (debian:latest)...
latest: Pulling from debian
64e5325c0d9d: Pull complete
bf84c1d84a8f: Already exists
debian:latest: The image you are pulling has been verified. Important: image verification is a tech preview feature and should not be relied on to provide security.
[...]

PR #1650, opened 2015-07-06
Fix NoneType on smart recreate

Vektah:
This fixes issue #1561 

smart-recreate hashes a dictionary of properties of the service, one of those is the image. The image can be null if it has not yet been fetched. By leaving it out of the hash in this case it will always mismatch causing the change to be recognised.

Alternatively the image existence could be checked for earlier.

Steps to reproduce bug from @josephpage:
```
$ mkdir demo
$ cd demo
$ cat <<EOF > docker-compose.yml
example:
  image: debian:latest
  command: echo "hello world"
EOF

$ docker-compose up -d --x-smart-recreate
Pulling example (debian:latest)...
latest: Pulling from debian
64e5325c0d9d: Pull complete
bf84c1d84a8f: Already exists
debian:latest: The image you are pulling has been verified. Important: image verification is a tech preview feature and should not be relied on to provide security.
Digest: sha256:2613dd69166e1bcc0a3e4b1f7cfe30d3dfde7762aea0e2f467632bda681d9765
Status: Downloaded newer image for debian:latest
Creating demo_example_1...
=> OK

$ cat <<EOF > docker-compose.yml
example:
  image: debian:8.1
  command: echo "hello world"
EOF

$ docker-compose up -d --x-smart-recreate
# BOOM
```

### DIFF
--- a/compose/service.py
+++ b/compose/service.py
@@ -402,10 +402,14 @@ class Service(object):
         return json_hash(self.config_dict())
 
     def config_dict(self):
-        return {
+        dict = {
             'options': self.options,
-            'image_id': self.image()['Id'],
         }
+
+        if self.image() is not None:
+            dict['image_id'] = self.image()['Id']
+
+        return dict
 
     def get_dependency_names(self):
         net_name = self.get_net_name()


compose/service.py
```diff
@@ -402,11 +402,15 @@ def config_hash(self):
...
 
     def config_dict(self):
-        return {
+        dict = {
             'options': self.options,
-            'image_id': self.image()['Id'],
         }
 
+        if self.image() is not None:
+            dict['image_id'] = self.image()['Id']
```
dnephin: How about:

```python

return {
    ...
    'image_id': (self.image() or {}).get('Id')
}
```

That way the API call is only performed once, and I think it would be the same behaviour

dnephin:
I'm not sure if this is going to be correct in all cases, but I think it does address the immediate issue.

I wonder if it would be better to use the image name in the case where the image Id doesn't exist yet?

aanand:
Thanks for the fix @Vektah, but this isn't quite right.

If the image we want to use doesn't exist, the service's containers *definitely* need recreating, so we can bail out of trying to create a config hash and just mark the service for recreating. I've implemented this in https://github.com/docker/compose/pull/1658.

PR #1659, opened 2015-07-06
Allow scaling services that have port binding of the form `host_ip::container_port`

viranch:
Compose gives up scaling if a service has a port binding of the form: `192.168.1.2::80` (bind to a specific host IP but a random host port, for the container port 80) with the error:

```
Service "foo" cannot be scaled because it specifies a port on the host. If multiple containers for this service were created, the port would clash.

Remove the ":" from the port definition in docker-compose.yml so Docker can choose a random port for each container.
```

Compose wrongly assumes that a host port is specified because there's a `:` in the port binding.

This PR fixes this.

### DIFF
--- a/compose/service.py
+++ b/compose/service.py
@@ -770,10 +770,28 @@ class Service(object):
         return self.options.get('container_name')
 
     def specifies_host_port(self):
-        for port in self.options.get('ports', []):
-            if ':' in str(port):
+        def has_host_port(binding):
+            _, external_bindings = split_port(binding)
+
+            # there are no external bindings
+            if external_bindings is None:
+                return False
+
+            # we only need to check the first binding from the range
+            external_binding = external_bindings[0]
+
+            # non-tuple binding means there is a host port specified
+            if not isinstance(external_binding, tuple):
                 return True
-        return False
+
+            # extract actual host port from tuple of (host_ip, host_port)
+            _, host_port = external_binding
+            if host_port is not None:
+                return True
+
+            return False
+
+        return any(has_host_port(binding) for binding in self.options.get('ports', []))
 
     def pull(self, ignore_pull_failures=False):
         if 'image' not in self.options:
--- a/tests/unit/service_test.py
+++ b/tests/unit/service_test.py
@@ -444,6 +444,68 @@ class ServiceTest(unittest.TestCase):
         }
         self.assertEqual(config_dict, expected)
 
+    def test_specifies_host_port_with_no_ports(self):
+        service = Service(
+            'foo',
+            image='foo')
+        self.assertEqual(service.specifies_host_port(), False)
+
+    def test_specifies_host_port_with_container_port(self):
+        service = Service(
+            'foo',
+            image='foo',
+            ports=["2000"])
+        self.assertEqual(service.specifies_host_port(), False)
+
+    def test_specifies_host_port_with_host_port(self):
+        service = Service(
+            'foo',
+            image='foo',
+            ports=["1000:2000"])
+        self.assertEqual(service.specifies_host_port(), True)
+
+    def test_specifies_host_port_with_host_ip_no_port(self):
+        service = Service(
+            'foo',
+            image='foo',
+            ports=["127.0.0.1::2000"])
+        self.assertEqual(service.specifies_host_port(), False)
+
+    def test_specifies_host_port_with_host_ip_and_port(self):
+        service = Service(
+            'foo',
+            image='foo',
+            ports=["127.0.0.1:1000:2000"])
+        self.assertEqual(service.specifies_host_port(), True)
+
+    def test_specifies_host_port_with_container_port_range(self):
+        service = Service(
+            'foo',
+            image='foo',
+            ports=["2000-3000"])
+        self.assertEqual(service.specifies_host_port(), False)
+
+    def test_specifies_host_port_with_host_port_range(self):
+        service = Service(
+            'foo',
+            image='foo',
+            ports=["1000-2000:2000-3000"])
+        self.assertEqual(service.specifies_host_port(), True)
+
+    def test_specifies_host_port_with_host_ip_no_port_range(self):
+        service = Service(
+            'foo',
+            image='foo',
+            ports=["127.0.0.1::2000-3000"])
+        self.assertEqual(service.specifies_host_port(), False)
+
+    def test_specifies_host_port_with_host_ip_and_port_range(self):
+        service = Service(
+            'foo',
+            image='foo',
+            ports=["127.0.0.1:1000-2000:2000-3000"])
+        self.assertEqual(service.specifies_host_port(), True)
+
 
 class NetTestCase(unittest.TestCase):
 


aanand:
Thanks.

1. Rather than inspecting the string to see if there's an external port specified, it'd be better call `split_port()` and inspect its output.
2. It'd be good to have some unit tests for `can_be_scaled`.

compose/service.py
```diff
@@ -679,7 +679,8 @@ def labels(self, one_off=False):
 
     def can_be_scaled(self):
         for port in self.options.get('ports', []):
-            if ':' in str(port):
+            external = split_port(port)[1]
+            if external and not (type(external) == tuple and external[1] == None):
```
aanand: I think that if you're checking that `type(external) == tuple`, then checking it for truthiness is unnecessary. And I think [isinstance is better than comparing types](http://stackoverflow.com/a/1549854). So this can just be:

```python
if isinstance(external, tuple) and external[1] == None:
```

aanand: Oh, and [PEP 8](https://www.python.org/dev/peps/pep-0008/) says to use `is None`, not `== None`.

> Comparisons to singletons like None should always be done with is or is not , never the equality operators

viranch: > then checking it for truthiness is unnecessary

You mean `external[1] == None` is not required? I think it is, `external[1]` corresponds to host port, if its specified `can_be_scaled` should fail, else succeed.

compose/service.py
```diff
@@ -678,10 +678,21 @@ def labels(self, one_off=False):
         ]
 
     def can_be_scaled(self):
-        for port in self.options.get('ports', []):
-            if ':' in str(port):
+        # TODO: this function should ideally move to docker.utils.ports
+        # get rid of it once its accepted into docker-py
+        def has_external_port(external_binding):
+            if external_binding is None:
```
viranch: We don't need to check this at all IMO because [docker-py's `split_port`](https://github.com/docker/docker-py/blob/master/docker/utils/ports/ports.py#L55) never returns a list with `None` elements as the 2nd return value.

compose/service.py
```diff
@@ -678,10 +678,21 @@ def labels(self, one_off=False):
...
                 return False
-        return True
+            if isinstance(external_binding, tuple) and \
+                    (len(external_binding) < 2 or external_binding[1] is None):
```
viranch: I also doubt we need to check the length of the tuple given [docker-py's `split_port`](https://github.com/docker/docker-py/blob/master/docker/utils/ports/ports.py#L83) always returns a dual-length tuple (second element may be `None`, but it will still be there).

compose/service.py
```diff
@@ -678,10 +678,21 @@ def labels(self, one_off=False):
...
+                return False
+            return True
+
+        return not any(
+            has_external_port(external_binding)
+            for binding in self.options.get('ports', [])
+            for external_binding in (ports.split_port(binding)[1] or [])
+        )
```
viranch: Considering the above 2 comments, this function pretty much boils down to:

```python
def can_be_scaled(self):
    return all(
        isinstance(external_binding, tuple) and external_binding[1] is None # takes care of "127.0.0.1::80" format (returns True), False for "80:80" format
        for binding in self.options.get('ports', [])
        for external_binding in (ports.split_port(binding)[1] or []) # takes care of "80" format: split_port(...)[1] will be None, evaluating all([]) to True
    )
```

viranch:
I think we might also need tests covering port ranges, given docker-py's ports util support it.

viranch:
@aanand Ping :) Let me know about my comments on the diff, I'll rebase from master given the refactor of `can_be_scaled()` in https://github.com/docker/compose/commit/9ffe69a572e862eb90017cceddd65d76c4eed555

compose/service.py
aanand: If it's going to be a public function in docker-py, it needs to be able to handle all valid port bindings, not just the subset created by `split_port()`.

viranch: If we do what I have proposed in https://github.com/docker/compose/pull/1659/files#r34191487 then we don't have to put this `has_external_bindings()` function in docker-py as a public function at all, we can simply inspect what docker-py's `split_port()` provides us.

viranch:
Rebased with latest master, seems like the CI build failed due to docker errors.

```
+ docker run --rm -t --privileged --volume=/var/jenkins_home/workspace/Compose-PRs/.git:/code/.git --name compose-prs1012 --volume=/var/run/docker.sock:/var/run/docker.sock -e DOCKER_VERSIONS=all -e TAG=compose:454b52f --entrypoint=script/ci compose:454b52f --verbose
Running lint checks
Running tests against Docker 1.7.1
Waiting for Docker to start...
Docker failed to start
time="2015-08-08T08:26:11.302291821Z" level=warning msg="Udev sync is not supported. This will lead to unexpected behavior, data loss and errors. For more information, see https://docs.docker.com/reference/commandline/cli/#daemon-storage-driver-option" 
time="2015-08-08T08:26:11.303233892Z" level=error msg="There are no more loopback devices available." 
time="2015-08-08T08:26:11.303417028Z" level=fatal msg="Error starting daemon: error initializing graphdriver: loopback mounting failed" 
Build step 'Execute shell' marked build as failure
```

dnephin:
Resolves #1102

tests/unit/service_test.py
```diff
@@ -150,6 +150,40 @@ def test_build_port_bindings_with_nonmatching_internal_ports(self):
         self.assertEqual(port_bindings["1000"], [("127.0.0.1", "1000")])
         self.assertEqual(port_bindings["2000"], [("127.0.0.1", "2000")])
 
+    def test_specifies_host_port_with_host_ip_and_port(self):
+        service = Service(
+            'foo',
+            image='foo',
+            ports=["127.0.0.1:1000:2000"])
+        self.assertEqual(service.specifies_host_port(), True)
```
dnephin: Minor nit, now that we're using `py.test` we actually get better error messages if you use `assert service.specifies_host_port()` or `assert not ...`.

Would you mind changing the assertions to use bare `assert` ?

compose/service.py
```diff
@@ -749,10 +749,12 @@ def custom_container_name(self):
...
-        for port in self.options.get('ports', []):
-            if ':' in str(port):
-                return True
-        return False
+        return any(
+            # Binding has an external port
+            not (isinstance(external_binding, tuple) and external_binding[1] is None)
+            for binding in self.options.get('ports', [])
+            for external_binding in (ports.split_port(binding)[1] or [])
+        )
```
dnephin: Using numeric indexes to split tuples is generally discouraged.

I think this could be improved slightly by splitting out some of the logic into a nested function

```python

def has_host_port(binding):
    _, external_binding = binding
    if not external_binding:
        return False
    ...


return any(has_host_port(binding) for binding in self.options.get('ports', []))
```

tests/unit/service_test.py
viranch: Rest of the service_test.py still uses `self.assertEqual`, I don't want to break the convention. What you're suggesting could go in as a separate PR that updates all tests IMO.

viranch:
Oops, bad push closed this, reopening.

viranch:
Rebased with latest master.

dnephin:
LGTM

mnowster:
LGTM. Thanks @viranch :+1:

PR #1660, opened 2015-07-06
Add in memswap_limit run option

mnowster:
Add in missing memswap_limit run option that was included in issue https://github.com/docker/compose/issues/754


dnephin:
LGTM

PR #1666, opened 2015-07-07
Docs for `file` default behaviour

mnowster:
Change in behaviour, `file` key is optional and if not set the
default is to look within the same file as `extends` is defined.

### DIFF
--- a/docs/extends.md
+++ b/docs/extends.md
@@ -233,11 +233,13 @@ manually keep both environments in sync.
 ### Reference
 
 You can use `extends` on any service together with other configuration keys. It
-always expects a dictionary that should always contain two keys: `file` and
-`service`.
+always expects a dictionary that should always contain the key: `service` and optionally the `file` key.
 
-The `file` key specifies which file to look in. It can be an absolute path or a
-relative one&mdash;if relative, it's treated as relative to the current file.
+The `file` key specifies the location of a Compose configuration file defining
+the extension. The `file` value can be an absolute or relative path. If you
+specify a relative path, Docker Compose treats it as relative to the location
+of the current file. If you don't specify a `file`, Compose looks in the
+current configuration file.
 
 The `service` key specifies the name of the service to extend, for example `web`
 or `database`.
--- a/docs/yml.md
+++ b/docs/yml.md
@@ -217,6 +217,9 @@ and adds `ports` and `links` configuration. It overrides one of the defined
 environment variables (DEBUG) with a new value, and the other one
 (SEND_EMAILS) is left untouched.
 
+The `file` key is optional, if it is not set then Compose will look for the
+service within the current file.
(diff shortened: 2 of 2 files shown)


docs/extends.md
```diff
@@ -233,11 +233,9 @@ manually keep both environments in sync.
 ### Reference
 
 You can use `extends` on any service together with other configuration keys. It
-always expects a dictionary that should always contain two keys: `file` and
-`service`.
+always expects a dictionary that should always contain the key: `service` and optionally the `file` key.
 
-The `file` key specifies which file to look in. It can be an absolute path or a
-relative one&mdash;if relative, it's treated as relative to the current file.
+The `file` key specifies which file to look in. If the `file` key is not set then the default behaviour is to look in the current file that is specifying the extends. When setting the `file` key, it can be an absolute path or a relative one&mdash;if relative, it's treated as relative to the current file.
```
aanand: I find "the current file that is specifying the extends" a bit confusing. How about:

> The `file` key specifies which file to look for the service in. If it's not set, Compose will look in the current file.

ping @moxiegirl for style opinions regarding em dashes

mnowster:
@aanand ah yes, thank you, that wording is clearer. I haven't touched the emdash, that was already there :cat2: 

@moxiegirl hi, can you take a look and let me know if you have any feedback please? Thank you :)

docs/extends.md
```diff
@@ -233,11 +233,9 @@ manually keep both environments in sync.
 ### Reference
 
 You can use `extends` on any service together with other configuration keys. It
-always expects a dictionary that should always contain two keys: `file` and
-`service`.
+always expects a dictionary that should always contain the key: `service` and optionally the `file` key.
 
-The `file` key specifies which file to look in. It can be an absolute path or a
-relative one&mdash;if relative, it's treated as relative to the current file.
+The `file` key specifies which file to look for the service in. If it's not set, Compose will look in the current file. When setting the `file` key, it can be an absolute path or a relative one&mdash;if relative, it's treated as relative to the current file.
```
moxiegirl: The `file` is a bit ambiguous ---  any arbitrary file or a compose configuration file? Also what does it define the service or the extension---either could work.  Also, avoid future tense.

So, a rewrite would look like this:

The `file` key specifies the location of a Compose configuration file defining the extension.  The `file` value can be an absolute or relative path.  If you specify a relative path, Docker Compose treats it as relative to the location of the current file. If you don't specify a `file`, Compose looks in the current configuration file.  

### Editorial guidance on emdash
image
image

mnowster:
@moxiegirl this is really helpful feedback, thank you. I struggled when writing this sentence, too many instances of using the word 'file' has made the word seem weird to me now!

aanand:
LGTM

PR #1671, opened 2015-07-08
Added load_image yml option for optionally loading a tar image

philwrenn:
I am using a CI tool to build my dockers images. This addition allows me to bypass pulling from a registry if a tar image is available locally.

### DIFF
--- a/compose/config.py
+++ b/compose/config.py
@@ -53,7 +53,8 @@ ALLOWED_KEYS = DOCKER_CONFIG_KEYS + [
     'dockerfile',
     'expose',
     'external_links',
-    'name',
+    'load_image',
+    'name'
 ]
 
 DOCKER_CONFIG_HINTS = {
--- a/compose/service.py
+++ b/compose/service.py
@@ -2,6 +2,7 @@ from __future__ import unicode_literals
 from __future__ import absolute_import
 from collections import namedtuple
 import logging
+import os
 import re
 import sys
 from operator import attrgetter
@@ -240,6 +241,10 @@ class Service(object):
                 self.build()
             else:
                 raise NeedsBuildError(self)
+
+        elif self.can_be_loaded():
+            with open(self.options['load_image'], 'r') as data:
+                self.client.load_image(data=data)
         else:
             self.pull(insecure_registry=insecure_registry)
 
@@ -682,6 +687,9 @@ class Service(object):
 
     def can_be_built(self):
         return 'build' in self.options
+
+    def can_be_loaded(self):
+        return 'load_image' in self.options and os.path.exists(self.options['load_image'])
 
     @property
     def full_name(self):
--- a/docs/yml.md
+++ b/docs/yml.md
@@ -28,6 +28,14 @@ pull if it doesn't exist locally.
     image: orchardup/postgresql
     image: a4bc65fd
 
+### load_image
+
+Path to tar archive of an image to load. Use with `image` to load from a
+tar archive instead of pulling from a registry when the image doesn't
+exist locally.
+
+    load_image: /path/to/image/tar
+
 ### build
 
 Path to a directory containing a Dockerfile. When the value supplied is a


dnephin:
Thanks for the contribution!

I think if we're going to do this, we should have `image` support a path to a tarball instead of adding a new key. There's already a bunch of extra logic to handle the mutually exclusive `image` and `build` options, and this key would add yet more complexity to it.

Supporting path on `image` is still a bit of a hard sell, since it's not supported directly by docker engine.

philwrenn:
I understand wanting to be judicious with adding new keys; considering that my approach is probably not best. Since adding a new key won't work, and using the most logical existing key would be a "hard sell." Am I correct in assuming that image tarball support in general is a long shot?

dnephin:
I think it might be too early to make that call. It would be good to understand your use case a bit better.

Am I correct in saying that it's designed to be used as a local cache?  Instead of pulling an image, the tarball would be made available locally?

How does the tarball get there? 

Is it built locally then exported? If that's the case, wouldn't it already be available in the list if images on the host?

If it's build somewhere else, why isn't the registry being used to distribute it?

philwrenn:
We have a Jenkins CI server in our office that we use to build docker images for our developers. We keep the images synced out to our developers' systems for quick access. So yes I guess I would call that a cache.

dnephin:
Why not use a crontab with `docker pull` to make them available locally, instead of a tarball ?

philwrenn:
Doesn't docker pull only work with a docker registry instance?

dnephin:
Yes, that's true. You would have to setup a local private registry to host the images.

philwrenn:
I realize it's not the way, but I am still interested in using tarballs.  So even if just for my own use, I am going rewrite this so `image` can accept an image tarball. Maintaining a patch for my own build is certainly easier than writing my own solution from the ground up.

If you were going to add tarball support to `image` how would you determine that the string was a tarball and not an image name? Would you think just testing if the path exists would be sufficient?

dnephin:
Yup, I think that's what I would do in that case

aanand:
> If you were going to add tarball support to `image` how would you determine that the string was a tarball and not an image name? Would you think just testing if the path exists would be sufficient?

I'm a little wary of logic like this because of cases like misspelled/misplaced files, so let's sketch it out:

```python
def get_image_tag(image):
    if os.path.exists(image):
        return import_tarball(image)

    if not image_exists(image):
        client.pull(image)

    return image
```

So if the file doesn't exist we'll treat the path as an image tag and attempt to pull it. Here's what can happen when we do that:

```
$ docker pull file.tar
Pulling repository docker.io/library/file.tar
Error: image library/file.tar:latest not found
$ docker pull dir/file.tar
Pulling repository docker.io/dir/file.tar
Error: image dir/file.tar:latest not found
$ docker pull dir.with.dots/file.tar
Error response from daemon: unable to ping registry endpoint https://dir.with.dots/v0/
v2 ping attempt failed with error: Get https://dir.with.dots/v2/: dial tcp: lookup dir.with.dots: no such host
 v1 ping attempt failed with error: Get https://dir.with.dots/v1/_ping: dial tcp: lookup dir.with.dots: no such host
```

And in the probably extreme edge case where the directory prefix is the name of an actual Docker registry:

```
$ docker pull quay.io/file.tar
Pulling repository quay.io/file.tar
Error: Status 403 trying to pull repository file.tar: "{\"error\": \"Permission Denied (user: aanand)\"}"
```

In each of these cases, we'd need to make a decision about what error message to show the user, not knowing if their intention was to import a tarball or refer to an image tag. It would be quite easy to accidentally confuse them. For that reason, I'd be cautious about overriding the meaning of `image`.

I know *we're* not planning to actually implement this functionality right now in any case, but those are my thoughts - I hope they help if you choose to try and implement it yourself.

philip-dakno:
@aanand and @dnephin  I really appreciate you thoughts on this. Thanks.

PR #440, opened 2014-08-24
Support python 3

dnephin:
Progress on #219

The test suite passes, but I ran into some errors while trying to run it under python 3.4 manually (https://github.com/d11wtq/dockerpty/issues/8 is the first).

It might be a good idea to have the test suite running python3 even if it doesn't work for real yet. That way new code will have to be compatible and it'll be less work to fix the remaining python3 issues (as we add tests which expose the problems).

PR #1672, opened 2015-07-08
Misc. Tweaks for Python 3 Support

adamkrellenstein:
Pretty uncontroversial stuff, I think. Doesn't alter Python 2 execution.


dnephin:
Looks familiar #440

dnephin:
Thanks for the contribution.  I believe I have all of these in #440 which also enables the tests for py3 so we don't get regressions.  I hope to get that branch merged soon.

Issue #1661, opened 2015-07-06
Feature request: add scale parameter in yml


jmmills:
As a user of compose (formally fig), I would like to be able to specify the number of nodes started for any given definition (a.k.a scale) from inside the manifest (yaml configuration file), so that I can ship my cluster definition with my service orchestration. 

E.g. syntax:

```
worker:
    build: rqworker
    scale: 5
    links:
       - redis
    command: rqworker -u tcp://redis 
```

aanm:
@jmmills Can't you start your containers such as: "docker-compose scale worker=5" and that service would start with 5?

jmmills:
@aanm Yes, but I think that functionality should be mirrored as a default in the service definition. Maybe I have a minimal set of workers that should always be running and I want that to be clearly declared as a default.

prologic:
@@aanm Do you expect that a ``docker-compose up`` should take the ``scale`` parameter into account? The only problem with this that I see is that we're introducing parameters into the declarative configuration that are **NOT** compatible with the underlying Docker / Docker API concepts but very specific to Docker Compose.

If we were to do this going forward; I'd suggest something like:

```
#!yaml
worker:
    build: rqworker
    $scale: 5
    links:
       - redis
    command: rqworker -u tcp://redis 
```

Where ``$<parameter>`` denotes a Docker Compose specific thing.

aanand:
We've gone back and forth on whether `scale` belongs in the YAML; there's been a PR to add it before (#630). I'm still of the opinion that most people shouldn't put scale numbers in their config, because it makes it less portable, but I understand the use cases for doing so.

Now that we've got a rudimentary way to augment Compose files with `extends` (and hopefully better ones soon - #1380, #758), the concerns I raised in https://github.com/docker/compose/pull/630#issuecomment-69210279 are perhaps less of an issue.

jamshid:
I'd like to set `scale=0` in my yml for test-related services that I don't normally want started. I only want to create those services explicitly with `docker-compose run` or an explicit `scale` .

jmmills:
@jamshid I've often wanted that, a definition that sets up an environment but doesn't run by default. I've been relegated to creating a base image (which a zero/no-op scale would also help with) in which I run my unit tests out of (via ```docker run```), and then my container composition consumes the base image. 

Something like this seems pretty useful for dev configurations

```
myproject:
    build: .
    command: nosetests
    scale: 0
    links:
       - redis
redis:
    image: redis
apiserver:
[...]

PR #1676, opened 2015-07-09
WIP: Docker networking support

aanand:
This is starting out as a docs-only PR, meant to get early feedback on the direction of Docker's networking support, from the viewpoint of Compose. It's scoped to what we plan to implement in the first stable release.

Your comments are welcomed.

Replaces #1346.

ping @dave-tucker @bfirsh @shykes @mavenugo @lxpollitt @monadic


nazar-pc:
Does links going to be dropped from Docker itself like you did in docs? Never heard about that.
In my opinion it was quite useful to be strict about who can access whom.
Creating separate network for each app is OK, but few network for app without really any need is overkill I think.

Also you do not currently mention how to provide aliases like `db:mysql` in links with publishing.

Also
> Looking up the name `web` will return the IP address of one of them, but Docker and Compose do not provide any guarantees about which one

Is quite a big issue, there should be a way to differentiate them, adding suffixes would allow, for instance, load balancer to discover all web servers automatically to do its job.
Or labels might be used here? I think it worth to mention this explicitly as recommended way to do things.

docs/networking.md
```diff
@@ -0,0 +1,114 @@
...
+        - "other-name"
+        - "web"
+
+You can stop a service's containers from being published at all by passing in an empty list.
+
+    web:
+      build: .
+      publish: []
+
+## Configure your own networks
```
dave-tucker: We may want to push this out until we've fixed some of the implementation issues with regard to having containers attached to multiple networks.

docs/networking.md
```diff
@@ -0,0 +1,114 @@
...
+
+If the default behaviour of a single app-wide network isn't flexible enough, you can configure your own networks and specify what networks each container should join.
+
+Instead of defining your services at the top-level in `docker-compose.yml`, place them under a `services` key, and define networks under a top-level `networks` key.
+
+Here's how you would set up a 2-network, 3-container app where the `nginx` and `db` services cannot discover or communicate with each other directly:
+
```
lxpollitt: I think this section should focus on providing the flexibility to run multiple apps in a single network.  For example, `docker compose up -d --net=<network>` (and any equivalents of that which might feature directly within the Compose file if we want to support specifying networks in the Compose file).

I don't think support for containers connecting to multiple networks in 1.8 timeframe is realistic.  I think there are broad range of libnetwork, network driver and service discovery interaction and implementation issues that need to be resolved to make that work sensibly.  I suggest restricting each container to a single network for 1.8.

docs/networking.md
```diff
@@ -0,0 +1,114 @@
...
+When you run `docker-compose up`, the following happens:
+
+1. A network called `myapp` is created.
+2. A container is created using `web`'s configuration. It joins the network `myapp` under the name `web`.
+3. A container is created using `db`'s configuration. It joins the network `myapp` under the name `db`.
+
+Each container can now look up the hostname `web` or `db` and get back the appropriate container's IP address. For example, `web`'s application code could connect to the URL `postgres://db:5432` and start using the Postgres database.
+
+Because `web` explicitly maps a port, it's also accessible from the outside world via port 8000 on your Docker host's network interface.
```
lxpollitt: In 1.8 timeframe is it likely that `ports` will be supported by libnetwork or the plugin multi-host network drivers will support this?  So is this really just an option when using the bridge network?  If so then in the short term I think it would be worth tweaking the wording to finesse this.  e.g. "When testing on a single host system using the bridge network you can..." or similar.

Longer term I think `ports` could map sensibly to plug-in drivers where NAT is required at the boundary of the network.  The drivers could implement this in a variety of ways depending on their architecture. e.g. either via specific NAT device, or via fully distributed NAT similar to what K8s does today, or any other way that fits with the driver.  (This would need some careful consideration though.  Is it a network driver issue?  Or when we have service discovery drivers is it a service discovery driver issue?  Or will there be separate load balancer drivers?  I suggest this is all for discussion later way beyond the scope of this first version of this document.)

docs/networking.md
```diff
@@ -0,0 +1,114 @@
...
+[menu.main]
+parent="smn_workw_compose"
+weight=6
++++
+<![end-metadata]-->
+
+
+# Networking in Compose
+
+By default, Compose sets up a single default [network](http://TODO/docker-networking-docs) for your app. Each container for a service joins the default network and is discoverable via DNS under the service's name.
```
squaremo: You may want to say "via the system resolver" or similar here, instead of DNS specifically. Some solutions (e.g., the incumbent) will use `/etc/hosts` as a mechanism.

docs/networking.md
squaremo: Yep. Everything works neatly in the example given, but it's easy to construct a scenario which is not so clear; for instance if there are containers published as `foo` in both networks, what answer should `web` get if it asks for `foo`? (You could answer "just don't do that" I suppose, but it would be better to come up with a model that avoids such an ambiguity).

monadic: +1

docs/networking.md
```diff
@@ -0,0 +1,114 @@
...
+
+> **Note:** Your app's network is given the same name as the "project name", which is based on the name of the directory it lives in. See the [CLI docs](cli.md#p-project-name-name) for how to override it.
```
dnephin: Has there been any discussion about namespacing these like `com.docker.compose.<project_name>` or a little less verbose `composition.<project_name>` ?

moxiegirl:
@aanand @bfirsh If this is experimental in the next release, do y'all want to create an experimental directory? Then we can point to this directory from release notes and blogs.  Ping me back on email/slack if that doesn't work for you.

docs/networking.md
aanand: OK, I'm just going to remove this section for now.

docs/networking.md
aanand: (cc @dave-tucker)

Does this mean that the web container needs to be both on the `bridge` network and the app-specific network? If so, will Compose need to specify that, or will everything join `bridge` by default?

Also, given the "container in multiple networks" problem won't be solved generically for the first release, would this need to be a specially-handled case on the Engine side?

docs/networking.md
aanand: Good question. I would be in favour of namespacing - to most Compose users, the network name is an implementation detail that they won't spend any time looking at (unlike container names, which appear in `docker ps` and generally all over the place), so verbosity isn't much of a concern. `com.docker.compose.*` sounds good to me.

docs/networking.md
aanand: Good point. I've reworded this sentence.

docs/networking.md
lxpollitt: Note that we need to be slightly careful of how we namespace network names since there is a reasonable chance they may end up featuring in service discovery namespacing at some point in the future.  e.g. myapp.mynetwork.mycompany.com as a DNS entry (with the DNS search path including .mynetwork.mycompany.com by default for myapp's containers, so a simple app doesn't need to worry about this namespacing).

docs/networking.md
lxpollitt: @dave-tucker might be best placed to comment on what will be implemented in 1.8 timeframe.

My best guess is:
- Plugin network drivers will not support `ports`. (They are not currently passed over the libnetwork API, and adding this in the "right way" will need a reasonable amount of thought and community consensus building.)
- The `bridge` driver will continue to support `ports`.
- Plugin network drivers will continue to support their own mechanisms for providing NAT style function if it features in their solution.  For example, Weave have `weave export` amongst other things.

aanand: That all makes sense. I just want to ensure the dev use case still works: web needs to be able to talk to db (so it needs to be on the app's network), but it also needs to talk to the outside world (so it needs to be on the `bridge` network).

My understanding is that those are two separate networks, and so it needs to be able to join both.

lxpollitt: I'll let @dave-tucker comment, but having a container join two networks was one of the scenarios we were trying to avoid to 1.8 (because it needs a decent amount of thought to get right across the broad range of plugin network drivers).  So I was expecting the `ports` directive would only do anything if the network you've specified is the `bridge` network (either because that's what is set up as the default network, or because you explicitly specified that).  For the plugin drivers (via libnetwork) the `ports` directive would be ignored and it's up to the user to additionally issue whatever commands required (if any) to make web accessible from the outside world (e.g. `weave expose` or similar in the case of Weave).

dave-tucker: ok here's the deal. currently `-p`doesn't work properly with the multi-host driver. as a workaround, what has been proposed is that a container that want's to expose ports will be (silently) connected to the bridge network on the host to allow this. This isn't the best workaround, but it should be transparent to compose. 

@lxpollitt we should chat re: handling or `-p` in drivers as this is something I think should be part of the driver api to make composes' job easier

dave-tucker: @lxpollitt having just checked through the code i see it's passed as part of `epInfo` in `CreateEndpoint` - are we saying that currently drivers either a) don't see this or b) see it, but don't honour the user intent of publishing ports?

aanand: Thanks for the clarification. Since this document doesn't go as far as to mention network drivers (or multi-host) at all, I think it's fair to leave this sentence as is. Perhaps there should be another section about "Running an app on multiple hosts", which both explains the driver architecture and points out its limitations re: ports.

shykes: A design for handling -p has been discussed in significant detail. It does not require changing the driver API.

lxpollitt: @dave-tucker Let me double check.  (I based my comments about the port option not being passed to remote drivers on what someone had told me.  I didn't check the code myself.)

lxpollitt: @dave-tucker It looks like it _is_ in the API now.  (It wasn't when we first started writing the Calico driver, but was added some time later, probably pre DockerCon but hard to say.)  I think that means in theory remote drivers could implement the port mapping.  I don't have a feel for how many have or how people may have interpreted any requirement to do so.  There seems to be little documentation around what libnetwork is expecting drivers to do with the parameter (and no comments in the code either) so your guess is as good as mine.  I think the fact you say it doesn't currently work in multi-host and are proposing to silently attach to the bridge also is an indication greater clarity in this area would be helpful.  (I'm slightly hesitant about endorsing the silently connecting to the bridge workaround without understanding more about exactly how this is intended to work.  e.g. How is the container's namespace configured to ensure traffic goes down the right interface etc.)  Perhaps the discussion @shykes refers to is somewhere public and will help clarify for those of us who aren't already familiar with it?

lxpollitt: I should probably reiterate, that I do see the port mapping option as being important for a range of drivers and deployment scenarios.  (For example, I quite like Kubernetes approach of allowing every host to act as a port mapping redirect.  It makes a lot of sense to me in public cloud deployments given the current limitations of cloud APIs.)  I'm just trying to clarify what is supported today vs what is longer term, and how the feature is intended to behave longer term.

aanand:
Just an update: I've started on an implementation of the networking API in docker-py: https://github.com/docker/docker-py/pull/729

Following conversation with @mavenugo, we're considering renaming "networks" to "scopes" (at the Docker Engine and API level), and removing all talk of *networking*, since that's not the real developer-level concern (rather, it's service discovery).

I'm going to push an update to the docs to that effect.

moxiegirl:
@aanand Can you give me a definition of *scope* in this context?  I was looking through the PR and trying to replace "network" with "scope" as used...it wasn't coming out clean.  Scope is a synonym for range.  And depending on how you use it in context could cause confusion.

aanand:
@moxiegirl Sure. It's similar to how programming languages define "scope", as in "a context in which a particular name means a particular thing".

I'll ignore Compose for a moment and demonstrate using hypothetical Docker client syntax.

Suppose I have two applications. I create a *scope* for each, called say `app1` and `app2`.

```
$ docker service scope create app1
$ docker service scope create app2
$ docker service scope ls
ID             NAME       NETWORK DRIVER  DISCOVERY DRIVER
e722ff412f35   app1       overlay         default
f832ff40dfb7   app2       overlay         default
```

In each app's scope, there is a set of *services*, each with a name that's unique within that scope.

A container in that app's scope can *publish* as a *backend* for a particular service.

`docker run` has a `--service` flag for publishing a container as a backend for a service, creating the service if it doesn't already exist.

```
$ docker run -d --service app1.db postgres
abc123

$ docker run -d --service app2.db mysql
def456

$ docker service ls
NAME      SCOPE          BACKEND        IP         PORTS
db        app1           abc123         10.42.0.1  5432:5432
db        app2           def456         10.42.0.3  3306:3306
```

Within the `app1` scope, the hostname `db` resolves to the IP of `app1`'s database container.

```
$ docker run --service app1.web app1-web-image ping db
PING db (10.42.0.1) 56(84) bytes of data.
64 bytes from db (10.42.0.1): icmp_seq=1 ttl=64 time=997 ms
64 bytes from db (10.42.0.1): icmp_seq=2 ttl=64 time=0.037 ms
^C
```

Within the `app2` scope, the hostname `db` resolves to the IP of `app2`'s database container.

```
$ docker run --service app2.web app2-web-image ping db
PING db (10.42.0.3) 56(84) bytes of data.
64 bytes from db (10.42.0.3): icmp_seq=1 ttl=64 time=997 ms
64 bytes from db (10.42.0.3): icmp_seq=2 ttl=64 time=0.037 ms
^C
```

moxiegirl:
@aanand So, I'm expecting the work `scope` to appear in the `networking.md` file you have provided and it isn't.  In the documentation, I'm expecting to see how the scope related to the network as compose understands it, I see in your pseudo command above a `NETWORK DRIVER` column and no mention of driver in `networking.md` either...  Are you planning on updating `network.md`?

Also, I'm still waiting to hear if this is "experimental" or going out finished...whatcha thinking?

mavenugo:
@moxiegirl thats correct. it will appear in networking.md shortly. Also, the word `NETWORK DRIVER` and `NETWORK PLUGINS` are used interchangably. To be technically correct, `NETWORK DRIVER` is the right usage. Plugins provide a way for drivers to be external to the docker daemon process.

the experimental networking features will be part of 1.9 proper release & it is appropriate to have that assumption in compose.

lxpollitt:
@aanand, @mavenugo: Can you confirm that the `scope` applies to both service discovery and reachability. i.e. If I bypass service discovery and just try direct IP address access then the scope limits that too.  (I think that was the plan for the old `network` concept, but wanted to double check in case the thinking has moved on for `scope`.)

mavenugo:
@lxpollitt thats correct. `scope` covers both service discovery and service reachability. It is enabled by default and the users will get these features out of the box. I would say that the thinking is converging well in this direction.

moxiegirl:
@mavenugo I'm sorted thanks. @aanand When  you are ready for a docs review let me know.

shykes:
Alex, yes that is correct. They are still networks, nothing has changed. We
are just experimenting with different words to call them something other
than "network". Honestly people seem pretty confused by the word "scope" so
far so we might switch back. In any case, the underlying concepts have not
changed.

lxpollitt:
@shykes Do you have any feel for whether it is general developers (i.e. users of Docker) or network people (i.e. libnetwork plug-in implementors) getting confused?  Or is it just people in general?  I have a networking focus in my interests here, but I really don't mind what it is called, so long as its definition is clear and I don't foresee implementation challenges.  Given that `scope` here is being defined as both a limit on service discovery and a limit on reachability (or at least default discoverability and reachability), then it seems perfectly reasonable to me. But people do get hung up on names surprisingly often!

Where it gets trickier from for service discover and networking implementations is around whether a single container can exist in / be connected to multiple scopes and what the expected behavior is in those cases.  I don't know what the latest thinking on that is, but that's probably a discussion for another day.

nazar-pc:
For me personally it is tricky because "scope" is not directly associated with network alone - Docker have FS layer with its own drivers, for instance. Why not just call it "network", no one will be confused about what it actually does.

shykes:
On Wed, Aug 26, 2015 at 8:24 PM, Alex Pollitt <notifications@github.com>
wrote:

> @shykes <https://github.com/shykes> Do you have any feel for whether it
> is general developers (i.e. users of Docker) or network people (i.e.
> libnetwork plug-in implementors) getting confused? Or is it just people in
> general? I have a networking focus in my interests here, but I really don't
> mind what it is called, so long as its definition is clear and I don't
> foresee implementation challenges. Given that scope here is being defined
> as both a limit on service discovery and a limit on reachability (or at
> least default discoverability and reachability), then it seems perfectly
> reasonable to me. But people do get hung up on names surprisingly often!
>
The biggest confusion with "scope" is simply that it's very generic, and
not obviously a networking feature. A scope for what? That's the problem in
a nutshell.

The problem with "network" was that traditional network engineers would get
hung up on the technical accuracy, because it's a virtualized network, not
necessarily a physical one.

> Where it gets trickier from for service discover and networking
> implementations is around whether a single container can exist in / be
> connected to multiple scopes and what the expected behavior is in those
> cases. I don't know what the latest thinking on that is, but that's
> probably a discussion for another day.
>

Yes, a single container can be connected to any number of scopes, and that
can be changed dynamically at any time. That is a major advantage of the
new Docker networking model.

The expected behavior is rather straightforward:

- The container is discoverable as different names from from different
scopes. Depending on the scope of a particular lookup, it will be reachable
at a different IP address.
- Avoiding conflicts between IP addresses is left as an exercise to the
reader. As long as you use the same plugin for all networks, it should work
out (if your plugin is not broken). Conflicts across plugins are possible.
We are thinking about a possible, separate extension point to allow IP
allocation to be specified separately from any particular connectivity
plugin.

- When the container is making a hostname lookup (outbound discovery),
there is the possibility of a name conflict (there might be 3 different
services called "db" on the 3 networks visible to the container). In this
case discovery will apply a priority order. That order could then be made
configurable.

aanand:
> When the container is making a hostname lookup (outbound discovery),
there is the possibility of a name conflict (there might be 3 different
services called "db" on the 3 networks visible to the container). In this
case discovery will apply a priority order. That order could then be made
configurable.

Is it also the case that a container can choose to avoid conflict by performing network-specific lookups, e.g. `db.network-a`, `db.network-b`, `db.network-c`?

mavenugo:
@aanand thats correct. if apps are smart to use the scope as part of the lookup, it should work seamlessly. The priority logic is applicable in a case when there is a service name conflict if scope is not included during lookup.

moxiegirl:
I'm +1 for going with network rather than introducing a new term like `scope`.  Yes, network guru's might find `scope` less than precise.  If we always went around using the "highly educated" terminology,  I'd be typing this with my "phalanges."  My point isn't that we should ignore these educated specialists, rather that they are more likely a subset of users...choosing words that make a specialized subset happy can sometimes cause overhead for the larger user population.

shykes:
Good question Aanand. My opinion: yes but with a caveat.

1) yes, the container should be able to make network-specific lookups. The
most useful use cases will be A) manual troubleshooting ("can I ping foo
from here?") or B) smart load-balancers or other multi-network-aware
containers.

2) but, the network names and IDs should not be exposed directly to the
container. Rather they should be namespaced, with perhaps a way cor the
user to specify the container-specific alias of each network.

TLDR: just because you call your network "prod" doesn't mean containers can
automatically lookup db.prod.

aanand:
**Attention:** this has been updated with the docs for the final planned functionality for Docker 1.9 / Compose 1.5. Please take a second look.

bfirsh:
Nice!

My only concern is that removing links is a huge change for users. It might be worth sprinkling around some helpful notes. "links are no longer used in Compose 1.5... check out the networking docs to read about the future"

Also - presumably we will force users to update to Docker 1.9 in Compose 1.5? If a user doesn't want to upgrade to 1.9, we should make sure users can still find the Compose 1.4 docs with "links", etc.

ibuildthecloud:
@mavenugo @aanand Does the section `Configure how services are published` still apply?

ibuildthecloud:
How does a compose project use any network driver besides the default driver?  Is it possible for a compose project to join an existing network?

ibuildthecloud:
I'm not fully comfortable with this proposal.  First off, I'm glad to see the docker networking UX finally land in 1.9, but I think the jury's still out on how users will use networks.  Here's my observations thus far...

Users really want cross host networking.  That has been said loud and clear.  Beyond that simple request I don't have a clear picture of why users want networks as a first class object from an application perspective.  I work with a good amount of users and I rarely get requests where people want VPC-like or OpenStack Neutron like functionality in Docker from an application perspective.  What I do get is a lot of requests where people want to plug docker into an existing network fabric.  OVS, VLANs, or something crazy.  They have a network managed by another team and they just need the Docker containers to play nice with the existing networking model.  So you can see that this is more of an administrative function and not a user/developer function.  Once Docker 1.9 is out I will be working with a lot of users to address these one off networking requirements.

I would much rather see compose hold off on using networks as a core part of the functionality.  I think compose should support `net: foo`, but beyond that don't change until we understand better how networking will be used.  My fear is that if compose makes core assumption of networking it will make it not possible to use compose in certain situations.

For example, I am working with some one now that must plug Docker into OVS in an extremely custom way.  The way in which they are doing this, having a single Docker network for their custom set up makes sense, and then use that network as the default.  They have no requirement to create multiple "networks" in Docker.  Now if I want to deploy two compose templates and both have a container named "foo" the only way I can do that after this proposed change would be to create two networks.  But... in the scenario I described they can't create two networks.  The only way I can support two network with this OVS approach is if I write a driver that creates "pseudo" networks.  Now this just gets really odd.

What compose really needs is service discovery namespacing, not networks.  Again, I think compose should just hold off a bit.  Lets see how users used Docker networks first.

dnephin:
I think this proposal supports the situation you describe with `publish:`.  Using `publish` you could either vary the name using environment variable interpolation (#1765 new in 1.5.0) or by using an override to name the services (#2051, also new in 1.5.0).  I believe that would let you run two copies on the same network. The network you join would be based on the project name, so you can configure that as well.

That said, I think it would be good to allow users to customize the default network (if they don't want it to be the project name), and to allow them to completely disable creating a default network.  I don't know if those features will be available for the initial release.

aanand:
A small update: following discussion with @mavenugo and @dave-tucker, we're going to use `hostname` for configuring the name by which a container joins the network.

How it'll work is: if you've set a `hostname` in `docker-compose.yml`, Compose sends that to the Engine, which uses it as the name for the container to join under. If you *haven't* set a `hostname`, Compose sets it to the service name for you.

This means that joining under multiple names, or no names at all, will not be supported for v1 - we'll be holding off on those until there's an actual service discovery API.

I've updated the doc:

https://github.com/aanand/fig/blob/3648c2c52bd8fb98f046f29d10640996934901c4/docs/networking.md#configure-how-services-are-published

ibuildthecloud:
@aanand How does scaling work now.  From what I understand you are saying either service name from compose or the hostname will be set to the container name (ie `--name`).  This seems to indicate you can't scale the service anymore.

aanand:
@ibuildthecloud The container name isn't used anywhere. If you've set `hostname` in your `docker-compose.yml`:

```yaml
web:
  hostname: mywebservice
```

then Compose will send `"Hostname": "mywebservice"` when creating each container for web.

If you haven't set a `hostname`, then Compose will send `"Hostname": "web"`.

In both cases, each container will join the network under that name. If you create multiple containers, they'll join under the same name. If you look up the hostname `web` inside any container that's connected to the network, you'll get the IP address of one of the web containers back, but precisely which one is left up to Linux. Under the hood, there are simply multiple entries in each container's `/etc/hosts`.

This is basically the same behaviour that scaling gives you today. It's not very useful, but it's a start - we need a proper load-balancing solution to make it viable.

pwaller:
@aanand is there a way to specify container scaling in the yml, so that `docker-compose up` brings up a functional cluster?

aanand:
@pwaller see #1661 - it's still under discussion.

mnowster:
The link to the git commit in the requirements isn't working, I think that's why the tests are failing.

The code itself LGTM.

aanand:
This is now ready to merge, other than the fact that it still depends on https://github.com/docker/docker-py/pull/729 and https://github.com/docker/docker/pull/16645.

It skips the networking-related tests unless running against a 1.9.x daemon - once there's a 1.9.0 RC out, thanks to @dnephin's work in #2069 they *should* automatically start running.

albers:
I will add this to bash completion once it is merged.

dnephin:
Merging as part of #2191

Issue #1651, opened 2015-07-06
Stop/kill/remove in parallel


bfirsh:
Stopping a project can often take a long time, sometimes even longer than doing "up".

When hitting ctrl-c in "up", we should stop all of the containers in parallel to speed it up. This can also be done when doing "stop", "kill" and "rm".

PR #1687, opened 2015-07-14
1651 parallelise stopping containers

mnowster:
As per issue: https://github.com/docker/compose/issues/1651 Stopping/killing/removing containers can be very slow, by doing it in parallel we gain significant speed improvements. Rough benchmarking was 60% faster with this approach.

### DIFF
--- a/compose/project.py
+++ b/compose/project.py
@@ -1,15 +1,16 @@
 from __future__ import unicode_literals
 from __future__ import absolute_import
-import logging
 from functools import reduce
+import logging
 
 from docker.errors import APIError
 
 from .config import get_service_name_from_net, ConfigurationError
-from .const import LABEL_PROJECT, LABEL_SERVICE, LABEL_ONE_OFF, DEFAULT_TIMEOUT
-from .service import Service
+from .const import DEFAULT_TIMEOUT, LABEL_PROJECT, LABEL_SERVICE, LABEL_ONE_OFF
 from .container import Container
 from .legacy import check_for_legacy_containers
+from .service import Service
+from .utils import parallel_execute
 
 log = logging.getLogger(__name__)
 
@@ -197,12 +198,15 @@ class Project(object):
             service.start(**options)
 
     def stop(self, service_names=None, **options):
-        for service in reversed(self.get_services(service_names)):
-            service.stop(**options)
+        parallel_execute("stop", self.containers(service_names), "Stopping", "Stopped", **options)
 
     def kill(self, service_names=None, **options):
-        for service in reversed(self.get_services(service_names)):
-            service.kill(**options)
+        parallel_execute("kill", self.containers(service_names), "Killing", "Killed", **options)
+
+    def remove_stopped(self, service_names=None, **options):
+        all_containers = self.containers(service_names, stopped=True)
+        stopped_containers = [c for c in all_containers if not c.is_running]
+        parallel_execute("remove", stopped_containers, "Removing", "Removed", **options)
 
     def restart(self, service_names=None, **options):
         for service in self.get_services(service_names):
@@ -283,10 +287,6 @@ class Project(object):
     def pull(self, service_names=None, insecure_registry=False):
         for service in self.get_services(service_names, include_deps=True):
             service.pull(insecure_registry=insecure_registry)
-
-    def remove_stopped(self, service_names=None, **options):
-        for service in self.get_services(service_names):
-            service.remove_stopped(**options)
 
     def containers(self, service_names=None, stopped=False, one_off=False):
         if service_names:
--- a/compose/const.py
+++ b/compose/const.py
@@ -1,4 +1,5 @@
 
+DEFAULT_MAX_WORKERS = 5
 DEFAULT_TIMEOUT = 10
 LABEL_CONTAINER_NUMBER = 'com.docker.compose.container-number'
 LABEL_ONE_OFF = 'com.docker.compose.oneoff'
--- a/compose/service.py
+++ b/compose/service.py
@@ -129,6 +129,7 @@ class Service(object):
         for c in self.containers(stopped=True):
             self.start_container_if_stopped(c, **options)
 
+    # TODO: remove these functions, project takes care of starting/stopping,
     def stop(self, **options):
         for c in self.containers():
             log.info("Stopping %s..." % c.name)
@@ -143,6 +144,8 @@ class Service(object):
         for c in self.containers():
             log.info("Restarting %s..." % c.name)
             c.restart(**options)
+
+    # end TODO
 
     def scale(self, desired_num):
         """
--- a/compose/utils.py
+++ b/compose/utils.py
@@ -1,5 +1,39 @@
-import json
 import hashlib
+import json
+import logging
+import os
+
+import concurrent.futures
+
+from .const import DEFAULT_MAX_WORKERS
+
+
+log = logging.getLogger(__name__)
+
+
+def parallel_execute(command, containers, doing_msg, done_msg, **options):
+    """
+    Execute a given command upon a list of containers in parallel.
+    """
+    max_workers = os.environ.get('MAX_WORKERS', DEFAULT_MAX_WORKERS)
+
+    def container_command_execute(container, command, **options):
+        log.info("{} {}...".format(doing_msg, container.name))
+        return getattr(container, command)(**options)
+
+    with concurrent.futures.ThreadPoolExecutor(max_workers=max_workers) as executor:
+        future_container = {
+            executor.submit(
+                container_command_execute,
+                container,
+                command,
+                **options
+            ): container for container in containers
+        }
+
+        for future in concurrent.futures.as_completed(future_container):
+            container = future_container[future]
+            log.info("{} {}".format(done_msg, container.name))
 
 
 def json_hash(obj):
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -193,6 +193,12 @@ the daemon.
 
 Configures the path to the `ca.pem`, `cert.pem`, and `key.pem` files used for TLS verification. Defaults to `~/.docker`.
 
+### DEFAULT\_MAX\_WORKERS
+
+Configures the maximum number of worker threads to be used when executing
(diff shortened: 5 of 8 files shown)


mnowster:
WIP: Now fixing up some things after feedback.

aanand:
Minor thing: I'm getting "..." at the end of lines which report that something's been done, not that it's in progress, e.g. "Stopped counter_redis_1...".

We should (probably separately) revisit the output anyhow - it'd be good to have something like `docker pull`'s output, where concurrent operations are reported and live-updated, one per line.

compose/project.py
```diff
@@ -54,7 +54,7 @@ def visit(n):
     return sorted_services
 
 
-class Project(object):
+class Project(ParallelContainerManager):
```
dnephin: I would love it if we could avoid inheritance here. I think an instance of`ParallelContainerManager()` could be an attribute instead of a base class. The only method it uses from `Project` is `self.containers()`. Instead of passing in the service of service_names to the calls, the list of containers could be passed in.

`Project.stop()` could then call `self.parallel_manager.execute('stop', containers, **options)`

compose/mixins.py
```diff
@@ -0,0 +1,51 @@
...
+        """
+        From a list of service names get all the containers and execute the
+        desired command on each of them.
+        """
+        if command == 'remove':
+            # we only want containers that are stopped/not running
+            all_containers = self.containers(service_names, stopped=True)
+            containers = [c for c in all_containers if not c.is_running]
+        else:
+            containers = self.containers(service_names)
```
dnephin: Removing the inheritance would remove the need for this branch logic as well

compose/cli/main.py
```diff
@@ -141,12 +141,20 @@ def kill(self, project, options):
...
-                              Default signal is SIGKILL.
+            -s SIGNAL                  SIGNAL to send to the container.
+                                       Default signal is SIGKILL.
+            --max-workers NUM_WORKERS  Specify max number of workers to use to stop containers.
+                                       (default: 5)
+
+
         """
         signal = options.get('-s', 'SIGKILL')
+        max_workers = float(options.get('--max-workers') or DEFAULT_MAX_WORKERS)
```
dnephin: I think this should be `int()` instead of `float()` since we're casting to an int later and it never gets used as a `float()`, right?

compose/project.py
```diff
@@ -197,23 +198,24 @@ def start(self, service_names=None, **options):
             service.start(**options)
 
     def stop(self, service_names=None, **options):
-        for service in reversed(self.get_services(service_names)):
-            service.stop(**options)
+        return self.container_executor("stop", service_names, **options)
```
dnephin: These `return`, but it looks like `container_executor()` doesn't return anything. Should it return the container list, or can we remove this return statement?

dnephin:
Looks great, my only real concern is the inheritance

compose/project.py
mnowster: I don't understand what your real concern is here? You want to avoid inheritance but not giving me any specific reasons why/what problems you can see. Mixins are a common pattern and a good first step to pull this functionality out as we explore further cases of parallelisification. Help me to understand what your intent is here.

This was a more iterative approach of make it work first, refactor it, improve. So the style of passing in services is a reflection of those steps. `stop` and `kill` have the same logic for creating the container list, so I'd rather not duplicate that logic in project.py when I can have it once in here with a single code branch for remove, which is a different flow case. Granted, at the time I had `restart` as well but then had to back that out.

compose/cli/main.py
mnowster: Yes, good spot, thanks. I'm re-working these right now anyway and moving max_workers out into an ENV variable.

compose/project.py
mnowster: Yep, good catch, thank you. I'll remove.

mnowster:
@aanand I think I can fix it to stop doing that for these cases in this PR and then yes, agree, I'll open up a new PR and pull the output in line with `docker pull`

compose/project.py
aanand: My main (only?) problem with the use of inheritance here is that it's hiding the degree of coupling between `ParallelContainerManager` and any class which mixes it in. The contract seems to be:

1. The subclass must implement `containers(self, service_names, stopped)`. It returns a list, whose elements must implement `name` and `is_running`.
2. The subclass must have a `log` attribute, which is an instance of `logging.Logger` (or implements the same interface).

In the second case, I think having a global file-level `log` object removes that coupling. In the first case, it's beginning to feel like `ParallelContainerManager` knows a bit too much about `Project`, or even about containers.

Suppose we had a single method:

```python
def parallel_execute(objects,
                     method,
                     doing_message=None,
                     done_message=None,
                     max_workers=None,
                     args=None,
                     kwargs=None):
    """
    Executes the named method on each object in `objects` in parallel, passing
    along `args` and `kwargs` to each execution.

    Returns `None` once all executions have completed.

    If not `None`, `max_workers` should be an `int` specifying the maximum
    number of concurrent workers to use.

    If not `None`, `doing_message` should be a function which takes a single
    object and returns a string to output once the method has started.

    If not `None`, `done_message` should be a function which takes a single
    object and returns a string to output once the method has finished.
    """
    # ...
```

`project.py` would then look something like:

```python
def stop(self, service_names=None, **options):
    parallel_execute(
        self.containers(service_names),
        'stop',
        doing_message=lambda c: "{}: Stopping...".format(c.name),
        done_message=lambda c: "{}: Stopped".format(c.name),
        kwargs=options,
    )

def kill(self, service_names=None, **options):
    parallel_execute(
        self.containers(service_names),
        'kill',
        doing_message=lambda c: "{}: Killing...".format(c.name),
        done_message=lambda c: "{}: Killed".format(c.name),
        kwargs=options,
    )

def remove(self, service_names=None, **options):
    parallel_execute(
        [c for c in self.containers(service_names) if not c.is_running],
        'remove',
        doing_message=lambda c: "{}: Removing...".format(c.name),
        done_message=lambda c: "{}: Removed".format(c.name),
        kwargs=options,
    )
```

We've taken a duplication hit here, but what I like is that we've got the scary parallel stuff squirrelled away in as small and sealed a box as possible.

dnephin: >  You want to avoid inheritance but not giving me any specific reasons why/what problems you can see. 

I see inheritance as the problem.  There is already quite a bit written on the topic ("inheritance vs composition" or "why inheritance is wrong/bad/evil"). You can always represent the same logic using composition (instead of inheritance), and in my experience the code is always more maintainable and easier to read when composition is used.

I don't think this feature necessitates the additional of inheritance (since it's only really sharing a single param that can easily be passed in instead of the service names). It really doesn't even need to be a class, since it's just a single method with no state. It could be just a function.

> Mixins are a common pattern and a good first step to pull this functionality out as we explore further cases of parallelisification

I think this is where we disagree. I see mixins are an anti-pattern, and one that can (and should) be avoided. If this were just a function, you could use it anywhere, without depending on the class having a `containers()` method.  If we want to refactor Project later and move things around, we won't need to modify this container manager, because it doesn't depend on `Project` at all.

> `stop` and `kill` have the same logic for creating the container list, so I'd rather not duplicate that logic

It's not duplicated! you already have a function which handles the logic (`self.containers()`) you're just called a function twice, not duplicating any logic. If you wanted to go a step further and remove the existing duplication, you could make a single method on `Project` that takes the "signal" as a param.  This would removes even more duplication, and still avoid the branch logic and inheritance.

mnowster:
@aanand @dnephin thank you both for your explanations. Trying to convey intent and context via text in a box in github is hard.

I think some context of my choice in path was lost as I pre-emptively squashed some commits, I shall work on adapting my commit strategy for future. As I'm currently sat right next to @aanand there was also some discussion as I explained my path, which I will try harder to capture and write down.

@dnephin yeh, I think we have different experiences and design approaches(I've been coding for over 10 years and have seen many design approaches both functional and OOP work well in different contexts), which is Great :) I'm way more pragmatic than I am dogmatic re:design etc and honestly think having diversity in ideas/experience is a good thing and will provide a source of many possibilities and angles/suggestions in future. I appreciate we haven't met/chatted yet so is difficult to gauge conversation.

Anyway, moving code around is no problem, so I've re-factored in and squashed all the things :wind_chime:

compose/project.py
```diff
@@ -1,16 +1,20 @@
...
+import logging
+import os
 
+import concurrent.futures
 from docker.errors import APIError
 
 from .config import get_service_name_from_net, ConfigurationError
-from .const import LABEL_PROJECT, LABEL_SERVICE, LABEL_ONE_OFF, DEFAULT_TIMEOUT
+from .const import (DEFAULT_MAX_WORKERS, DEFAULT_TIMEOUT, LABEL_PROJECT,
+                    LABEL_SERVICE, LABEL_ONE_OFF)
```
aanand: Could we break this out to one per line, so future changes are less disruptive?

compose/project.py
```diff
@@ -326,6 +329,30 @@ def _inject_deps(self, acc, service):
...
+                    container,
+                    command,
+                    **options
+                ): container for container in containers
+            }
+
+            for future in concurrent.futures.as_completed(future_container):
+                container = future_container[future]
+                log.info("{} {}".format(done_msg, container.name))
+
```
aanand: Looks great. Since this doesn't use `self` any more, could it be moved into `compose.utils`?

compose/project.py
```diff
@@ -326,6 +329,30 @@ def _inject_deps(self, acc, service):
         dep_services.append(service)
         return acc + dep_services
 
+    def _parallel_executor(self, command, containers, doing_msg, done_msg, **options):
```
aanand: Minor thing, but I'd prefer a verb (`execute`?) over a noun.

compose/project.py
mnowster: You don't normally split them out individually onto each line, PEP8 recommends a few options for dealing with long imports, hence me putting them in brackets. 

What I could have done to be more helpful was in a separate commit, re-order them. Then another one adding in the import, I conflated the two which feels more disruptive. I'll keep this in mind for next time.

aanand:
LGTM

Issue #1590, opened 2015-06-22
Compose doesn't read ~/.docker/config.json


sschepens:
Seems like with docker 1.7 and docker-compose 1.3.1 authentication is broken and docker-compose fails to download private repositories on docker hub.
This used to work with docker 1.6 and compose 1.2.

leftyhitchens:
+1 one, I am also experiencing this issue.

jbiel:
See https://github.com/docker/docker/issues/14092

juhasz:
No, docker/docker#14092 is about private registries, but compose currently can't even authenticate to dockerhub. But it's not that simple:

Before Docker 1.7.0 docker registry authentications were stored in the `~/.dockercfg`, but now they are stored in `~/.docker/config.json`.

With updating to Docker 1.7.0 the contents of `~/.dockercfg` were migrated, but the old file wasn't removed. (See docker/docker#12009.)

But docker compose (or I think it's a docker-py issue), still use `~/.dockercfg`. So if you were logged in before, docker-compose will work with DockerHub authentication, but if you log in with a freshly installed 1.7.0 Docker, compose will not find the authentication config.

jbiel:
Apologies, my mistake.

volcomism:
+1

compose.progress_stream.StreamOutputError: Error: image blah/blah:latest not found

paulsjv:
@volcomism exact same error message here.  DC 1.3.1 Docker 1.7.0

aanand:
We need to update to docker-py 1.3.0 once it's out.

shanielh:
Same issue here, I managed to fix it by copying and modifying contents of `~/.dockercfg` (from `~/.docker/config.json`)

aanand:
Closed in #1688.

caiges:
I'm running into this with with docker 1.7.1 and docker-compose 1.3.2. I can connect to our private registry using docker directly but docker-compose fails authentication. @tnoeding

aanand:
@caiges do you have a `~/.dockercfg` or a `~/.docker/config.json`? If the former, it's not this bug.

aanand:
If the latter, please comment in https://github.com/docker/compose/issues/75 with your version info + registry version.

mark878:
FYI: I have his same problem with docker 1.7.1 and I tried docker-compose 1.4RC2

aanand:
@mark878 do you have a `~/.dockercfg` or a `~/.docker/config.json`?

[...]

PR #1688, opened 2015-07-14
Use docker-py 1.3.0

### LINKED ISSUE #1229: Output missing with container in TTY mode
I'm trying to run several containers with TTY mode enabled (`tty: true` in config). When I enable TTY mode, commands such as `compose up` and `compose logs` do not show the container's output. However I can run `docker logs $cid` and the output shows up just fine.
The output does show up when the container exits, but not until then. And when it does show up, the first several bytes of the stream are missing. For example:

```
hello:
  image: busybox
  command: sh -c 'echo hello world; sleep 2; echo done'
  tty: true
```

```
# compose up
Recreating code_hello_1...
Attaching to code_hello_1
hello_1 | rld
hello_1 | done
code_hello_1 exited with code 0
Gracefully stopping... (press Ctrl+C again to force)
```

Both lines showed up at the exact same time instead of having a 2 second pause, and note that the `hello wo` is missing.

This is using current master https://github.com/docker/compose/commit/43369cda9ce5f0fe4860c7160cfa6550d31e2417

aanand:
Closes #1229, #1590.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,8 +1,8 @@
 PyYAML==3.10
-docker-py==1.2.3
+docker-py==1.3.0
 dockerpty==0.3.4
 docopt==0.6.1
 requests==2.6.1
 six==1.7.3
 texttable==0.8.2
-websocket-client==0.11.0
+websocket-client==0.32.0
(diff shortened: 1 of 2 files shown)


dnephin:
LGTM

PR #1692, opened 2015-07-14
Add test coverage support

bfirsh:
Prints out results on console and puts HTML report in `coverage-html`.

Could also be integrated with coveralls if we wanted: https://coveralls.io/

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -2,3 +2,4 @@
 build
 dist
 venv
+coverage-html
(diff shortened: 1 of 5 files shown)


bfirsh:
ping @aanand @mnowster

mnowster:
I was just thinking I could do with the output of test coverage to see if I've got all the config tests covered. Win, thanks :balloon: 

LGTM

Issue #1691, opened 2015-07-14
Non-TTY causing bad wrapping, suggest remove setting width.


ekristen:
@ralphtheninja and I have stumbled across a problem while trying to programmatically use docker-compose via node.js and exec. If compose is not in a TTY it defaults to 80, which causes some wrapping that is difficult to parse.

Basically when trying to script this from a non-TTY, this is what happens ...

```
Name             Command             State              Ports       
-------------------------------------------------------------------------
workstation_ngin   nginx -g daemon    Up                 443/tcp, 0.0.0.0 
x_1                off;                                  :8888->80/tcp    
```

FWIW, Docker Machine does not appear to look for a TTY (at least a brief look at the code didn't turn anything up)

https://github.com/docker/compose/blob/master/compose/cli/formatter.py#L7-L12

Is there a reason for having this? If not it would be nice to remove. I am more than willing to submit a PR to do so.

ralphtheninja:
:+1:

aanand:
I don't know that there's any particular reason for it, no.

It would also be good to bring Compose's table formatting in line with Docker and Machine's.

PR #1693, opened 2015-07-14
Do not wrap table cells

ekristen:
Fixed https://github.com/docker/compose/issues/1691

### DIFF
--- a/compose/cli/formatter.py
+++ b/compose/cli/formatter.py
@@ -3,9 +3,18 @@ from __future__ import absolute_import
 import os
 import texttable
 
+
+def get_tty_width():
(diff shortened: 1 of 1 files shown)


aanand:
This removes the width-detection behaviour and wrapping entirely, which makes for unpleasant terminal output in the normal case. It would be best to disable wrapping only if we can't retrieve a terminal size.

aanand:
(This would simply entail changing `80` to `0`, I think.)

ekristen:
@aanand I fixed it to be 0

dnephin:
LGTM

aanand:
LGTM

ralphtheninja:
:+1:

PR #1701, opened 2015-07-15
Reformat and sync command descriptions

aanand:
CLI banners were out of sync with the wording in the docs, plus there were some unwrapped lines.

ping @moxiegirl

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -111,11 +111,11 @@ class TopLevelCommand(Command):
 
     def build(self, project, options):
         """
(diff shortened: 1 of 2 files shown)


moxiegirl:
Hi @aanand Thanks for doing this.  One global comment which is future tense "will start" > "starts"  in general best practice is:

image

aanand:
Thanks. I've updated it so we're using present tense, except for the `up` docs, since they've changed in #1702 anyway.

aanand:
Superseded by #1715.

PR #1702, opened 2015-07-15
Smart recreate by default

### LINKED ISSUE #1654: Stabilise state convergence
In #1399, a basic version of state convergence was hidden behind the `--x-smart-recreate` flag. This should be promoted to a stable feature and become the default behaviour.

A flag should be added to `up` to support the old behaviour where all containers are stopped and recreated.

aanand:
Closes #1654.

ping @moxiegirl (extra docs changes in this one)

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -127,15 +127,20 @@ Stops running containers without removing them. They can be started again with
 
 Builds, (re)creates, starts, and attaches to containers for a service.
 
-Linked services will be started, unless they are already running.
+Unless they are already running, this command also starts any linked services.
 
-By default, `docker-compose up` will aggregate the output of each container and,
-when it exits, all containers will be stopped. Running `docker-compose up -d`,
-will start the containers in the background and leave them running.
+The `docker-compose up` command aggregates the output of each container. When
+the command exits, all containers are stopped. Running `docker-compose up -d`
+starts the containers in the background and leaves them running.
 
-By default, if there are existing containers for a service, `docker-compose up` will stop and recreate them (preserving mounted volumes with [volumes-from]), so that changes in `docker-compose.yml` are picked up. If you do not want containers stopped and recreated, use `docker-compose up --no-recreate`. This will still start any stopped containers, if needed.
+If there are existing containers for a service, and the service's configuration
+or image was changed after the container's creation, `docker-compose up` picks
+up the changes by stopping and recreating the containers (preserving mounted
+volumes). To prevent Compose from picking up changes, use the `--no-recreate`
+flag.
 
-[volumes-from]: http://docs.docker.io/en/latest/use/working_with_volumes/
+If you want to force Compose to stop and recreate all containers, use the
+`--force-recreate` flag.
 
 ## Options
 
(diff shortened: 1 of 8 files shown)


docs/cli.md
```diff
@@ -129,13 +129,25 @@ Builds, (re)creates, starts, and attaches to containers for a service.
 
 Linked services will be started, unless they are already running.
 
-By default, `docker-compose up` will aggregate the output of each container and,
-when it exits, all containers will be stopped. Running `docker-compose up -d`,
-will start the containers in the background and leave them running.
+By default, `docker-compose up` will aggregate the output of each container
+and, when it exits, all containers will be stopped. Running
+`docker-compose up -d`, will start the containers in the background and leave
+them running.
```
moxiegirl: ### up

Builds, (re)creates, starts, and attaches to containers for a service.

Unless they are already running, link services start.

By default, `docker-compose up` aggregates the output of each container. When it exits, all containers are stopped. Running
`docker-compose up -d`, starts the containers in the background and leaves
them running.

By default, if there are existing containers for a service, and the service's
configuration or image has changed since they were created, `docker-compose up`
stops and recreates them (preserving mounted volumes), so that the changes
are picked up.

If you do not want containers stopped and recreated, use `--no-recreate` to start any stopped containers.

If you want to force compose to stop and recreate all containers, use
the `--force-recreate` flag.

mnowster:
Ok, think that's a LGTM :palm_tree:

docs/cli.md
```diff
@@ -71,21 +71,20 @@ For example,
...
 One-off commands are started in new containers with the same configuration as a
-normal container for that service, so volumes, links, etc will all be created as
+normal container for that service, so volumes, links, etc are all created as
 expected. When using `run`, there are two differences from bringing up a
 container normally:
 
-1. the command will be overridden with the one specified. So, if you run
-`docker-compose run web bash`, the container's web command (which could default
-to, e.g., `python app.py`) will be overridden to `bash`
+1. The command is overridden with the one specified. So, if you run
```
moxiegirl: These should just be bullets.  You should only use a numbered list on procedures or steps in a process.

docs/cli.md
```diff
@@ -71,21 +71,20 @@ For example,
 
     $ docker-compose run web python manage.py shell
 
-will start the `web` service and then run `manage.py shell` in python.
-Note that by default, linked services will also be started, unless they are
-already running.
+starts the `web` service and then runs `manage.py shell` in python. Note that
```
moxiegirl: python > Python

docs/cli.md
```diff
@@ -71,21 +71,20 @@ For example,
...
+by default, linked services are also started, unless they are already running.
 
 One-off commands are started in new containers with the same configuration as a
```
moxiegirl: What do you mean by a "one-off" command? In comparison to what? Is there an area in the compose documentation where we are explaining the differences between one-off commands and the XXX commands?

"normal container" implies there are "abnormal containers"  so if compose has the notion of different types of containers we should identify what are the various kinds of containers.  I think here you just mean something like you can start a container with a command that runs and exits or with a long running command like a daemon/service...but we should explain that concept in our compose docs. Where do we have that?

aanand:
Because #1701 has turned into #1715, I've rebased this so it's independent of it. It still contains @moxiegirl's updated description for the `up` command, so it should be good to go.

PR #1703, opened 2015-07-15
Rename max workers environment variable

aanand:
We should prefix our variables with `COMPOSE_`. Also, the name in the docs was wrong (`DEFAULT_MAX_WORKERS` instead of `MAX_WORKERS`), but that needs to change anyhow.


mnowster:
*face-palm* :crying_cat_face: 

Thanks for spotting/fixing.

PR #1710, opened 2015-07-16
Improve output for parallel command

mnowster:
This approach takes the style of replacing the output message, in
place, when the command has finished executing. Bringing it a bit
more inline with what `docker pull` does and improves readability for
the user.


mnowster:
@aanand this now includes the flush statement, so I think should be good to merge if you're happy.

aanand:
LGTM

PR #1711, opened 2015-07-16
Add container_name option for specifying a custom container name

### LINKED ISSUE #1518: Support custom container names
Requested in https://github.com/docker/compose/issues/652 and https://github.com/docker/compose/issues/1347.

It would be nice to allow users to customise container names, perhaps with a `name` field:

```yaml
web:
  image: myimage
  name: webapp
```

```
$ docker-compose up -d
$ docker-compose ps
CONTAINER ID        COMMAND             CREATED              STATUS              NAMES
747cb66d30f5        "python app.py"     About a minute ago   Up About a minute   webapp
```

However, it might not play nicely with `docker-compose scale`, since names are supposed to be unique.

Some options, keeping in mind that sequential suffixes are probably going away (#1516):

1. Don't allow services with a custom name to be scaled.

2.  Automatically append a suffix to the name of each container:

    `webapp_1a2b3c`, `webapp_4d5e6f`, `webapp_7a8b9c`

    Easy enough, but the feature becomes somewhat less useful at that point.

3. Automatically append a suffix to the name of all but the *first* container:

    `webapp`, `webapp_1a2b3c`, `webapp_4d5e6f`

    It works, but it's more complicated. What does the logic for creating/scaling look like now?

aanand:
Closes #1518.

```yaml
web:
  build: .
  container_name: my-web-container
```

Went with option 1 - a service with a custom name simply can't be scaled beyond 1 container. We let the Docker daemon enforce that.

### DIFF
--- a/compose/service.py
+++ b/compose/service.py
@@ -157,6 +157,12 @@ class Service(object):
         - starts containers until there are at least `desired_num` running
         - removes all stopped containers
         """
+        if self.custom_container_name() and desired_num > 1:
+            log.warn('The "%s" service is using the custom container name "%s". '
+                     'Docker requires each container to have a unique name. '
+                     'Remove the custom name to scale the service.'
+                     % (self.name, self.custom_container_name()))
+
         if self.specifies_host_port():
             log.warn('The "%s" service specifies a port on the host. If multiple containers '
                      'for this service are created on a single host, the port will clash.'
@@ -531,7 +537,8 @@ class Service(object):
             for k in DOCKER_CONFIG_KEYS if k in self.options)
         container_options.update(override_options)
 
-        container_options['name'] = self.get_container_name(number, one_off)
+        container_options['name'] = self.custom_container_name() \
+            or self.get_container_name(number, one_off)
 
         if add_config_hash:
             config_hash = self.config_hash()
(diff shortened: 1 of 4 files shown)


aanand:
ping @moxiegirl

compose/service.py
```diff
@@ -157,7 +157,12 @@ def scale(self, desired_num):
         - starts containers until there are at least `desired_num` running
         - removes all stopped containers
         """
-        if not self.can_be_scaled():
+        if self.custom_name() and desired_num > 1:
+            log.warn('Service %s specifies a custom container name. The Docker daemon '
```
moxiegirl: 'Service %s is attempting to use a container named  % self.name. Docker requires each container have a unique name.'

tests/integration/service_test.py
```diff
@@ -699,6 +699,13 @@ def test_empty_labels(self):
         for name in labels_list:
             self.assertIn((name, ''), labels)
 
+    def test_custom_container_name(self):
```
mnowster: Is it worth us also having a test that demonstrates that an error is raised if you try to call scale with a named container?

mnowster: Discussed this in person. For the record, we can look at increasing tests after code freeze depending on priority of other tasks.

mnowster:
LGTM.

PR #1713, opened 2015-07-16
Increase max-workers

mnowster:
There's significant speed improvement by having more workers. This
value still shouldn't cause anyone's machines to melt/explode.


aanand:
LGTM

PR #1714, opened 2015-07-16
Make scale timeout configurable, default to 10

### LINKED ISSUE #1707: Add a --timeout parameter to `docker-compose scale`
AFAIK, when you scale down a service, the excess containers get an stop command.
However, they can get killed prematurely if the stop is not quick enough.
Allowing for a stop timeout as it's done in 'docker stop' would solve this.

aanand:
Rather than forcing a timeout of 1 second when scaling a service down, use the default timeout of 10 seconds and make it configurable via the `--timeout` flag, as with `stop`, `restart` and `up`.

This makes scaling down slow when containers don't handle SIGINT properly, but that will be less of an issue when we're stopping them in parallel, which will be implemented soon.

Closes #1707.

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -372,8 +372,14 @@ class TopLevelCommand(Command):
 
             $ docker-compose scale web=2 worker=3
 
-        Usage: scale [SERVICE=NUM...]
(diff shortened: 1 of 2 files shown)


dnephin:
LGTM

PR #1715, opened 2015-07-17
Updating Compose documentation

moxiegirl:
- Split out commands into individual pages for maintainability
- Add full usage in commands pages for usability
- Updated description of the run command
- Carries and expands on doc change in #1701

ping @aanand

### DIFF
--- a/docs/reference/docker-compose.md
+++ b/docs/reference/docker-compose.md
@@ -1,0 +1,55 @@
+<!--[metadata]>
++++
+title = "docker-compose"
+description = "docker-compose Command Binary"
+keywords = ["fig, composition, compose, docker, orchestration, cli,  docker-compose"]
+[menu.main]
+parent = "smn_compose_cli"
+weight=-2	
++++
+<![end-metadata]-->
+
+
+# docker-compose Command
+
+```
+Usage:
+  docker-compose [options] [COMMAND] [ARGS...]
+  docker-compose -h|--help
+
+Options:
+  -f, --file FILE           Specify an alternate compose file (default: docker-compose.yml)
+  -p, --project-name NAME   Specify an alternate project name (default: directory name)
+  --verbose                 Show more output
+  -v, --version             Print version and exit
+
+Commands:
+  build              Build or rebuild services
+  help               Get help on a command
+  kill               Kill containers
+  logs               View output from containers
+  port               Print the public port for a port binding
+  ps                 List containers
+  pull               Pulls service images
+  restart            Restart services
+  rm                 Remove stopped containers
+  run                Run a one-off command
+  scale              Set number of containers for a service
+  start              Start services
+  stop               Stop services
+  up                 Create and start containers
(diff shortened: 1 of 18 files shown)


docs/reference/docker-compose.md
```diff
@@ -0,0 +1,52 @@
...
+
+ Specify what file to read configuration from. If not provided, Compose will look
+ for `docker-compose.yml` in the current working directory, and then each parent
+ directory successively, until found.
+
+ Use a `-` as the filename to read configuration from stdin. When stdin is used
+ all paths in the configuration will be relative to the current working
+ directory.
+
+ Specifies an alternate project name (default: current directory name)
```
aanand: Looks like the descriptions for the global `-f` and `-p` options have ended up here without any context.

docs/reference/run.md
```diff
@@ -0,0 +1,54 @@
...
+--entrypoint CMD      Override the entrypoint of the image.
+-e KEY=VAL            Set an environment variable (can be used multiple times)
+-u, --user=""         Run as specified username or uid
+--no-deps             Don't start linked services.
+--rm                  Remove container after run. Ignored in detached mode.
+--service-ports       Run command with the service's ports enabled and mapped to the host.
+-T                    Disable pseudo-tty allocation. By default `docker-compose run` allocates a TTY.
+```
+
+Runs a one-time command against a service. For example, the following command starts the `web` service and runs `manage.py shell` in `python`.
```
aanand: "runs `manage.py shell` in `python`" feels a bit strange to me - makes it sound like `python` might be a container, or an image. Maybe:

> For example, the following command starts a container with the `web` service's image and configuration, and `python manage.py shell` as its command.

Also, I think `bash` would be a better example here, so people who don't know Django aren't thrown off wondering what `manage.py shell` is.

> For example, the following command starts a container with the `web` service's image and configuration, and `bash` as its command.
>
>     $ docker-compose run web bash

docs/reference/index.md
```diff
@@ -0,0 +1,16 @@
...
+keywords = ["fig, composition, compose, docker, orchestration, cli,  reference"]
+[menu.main]
+identifier = "smn_compose_cli"
+parent = "smn_compose_ref"	
++++
+<![end-metadata]-->
+
+## Compose CLI reference
+
+The following pages describe the usage information for the Docker Compose subcommands. You can also see this information by running `docker-compose [SUBCOMMAND] --help` from the command line.
```
aanand: When we say "The following pages", where are these pages going to be listed? Is it just in the sidebar?

docs/reference/docker-compose.md
moxiegirl: You are right. I'll fix that.

docs/reference/index.md
moxiegirl: This is something sloppy on my part, I forgot to do this.

moxiegirl:
comments incorporated. ping @aanand

aanand:
LGTM

PR #1717, opened 2015-07-17
Fix regression in docs for 'up'

aanand:
The new docs for `up` (since #1702 was merged) were lost in #1715.

ping @moxiegirl


moxiegirl:
LGTM

PR #1725, opened 2015-07-19
Updating with the latest image

moxiegirl:
Changing to base image from master


aanand:
LGTM

PR #1731, opened 2015-07-20
Make parallel tasks interruptible with Ctrl-C

### LINKED ISSUE #1729: Pressing ctrl-C twice to force kill no longer works
It used to do this:

```
Stopping composedemo_web_1...
^CKilling composedemo_web_1...
Killing composedemo_redis_1...
```

This now doesn't seem to work:

```
composedemo_web_1: Stopping...
composedemo_redis_1: Stopped
^C^C^C^C^C^C^C^C^C^C%
```

aanand:
The `concurrent.futures` backport doesn't play well with KeyboardInterrupt, so I'm using Thread and Queue instead.

Since thread pooling would likely be a pain to implement, I've just removed `COMPOSE_MAX_WORKERS` for now. We'll implement it later if we decide we need it.

Closes #1729.

### DIFF
--- a/compose/utils.py
+++ b/compose/utils.py
@@ -2,13 +2,11 @@ import codecs
 import hashlib
 import json
 import logging
-import os
 import sys
 
 from docker.errors import APIError
-import concurrent.futures
-
-from .const import DEFAULT_MAX_WORKERS
+from Queue import Queue, Empty
+from threading import Thread
 
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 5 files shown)


compose/utils.py
```diff
@@ -26,25 +23,33 @@ def parallel_execute(command, containers, doing_msg, done_msg, **options):
...
-            executor.submit(
-                container_command_execute,
-                container,
-                command,
-                **options
-            ): container for container in containers
-        }
-
-        for future in concurrent.futures.as_completed(future_container):
-            container = future_container[future]
```
dnephin: I'm curious if something like this would work

```python

try:
    for future in concurrent.futures.as_completed(future_container):		
        container = future_container[future]
except KeyboardInterupt:
    for future in future_container.values():
        future.cancel()

```

mnowster: I've just been playing around with something similar, I think it only works for futures that have finished or not yet started, won't bubble up to kill those that have already started work

mnowster:
LGTM

PR #1734, opened 2015-07-20
Parallelise scale

mnowster:
An approach to do scale making use of the parallel execute functionality.

Includes a re-factor of the parallelise functionality to decouple it from container.

Some remaining things that need improvement to be done in separate PRs:
- better error handling
- update tests

### DIFF
--- a/compose/project.py
+++ b/compose/project.py
@@ -198,15 +198,30 @@ class Project(object):
             service.start(**options)
 
     def stop(self, service_names=None, **options):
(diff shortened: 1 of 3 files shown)


mnowster:
Trying to name these things so they make sense but are generic is not easy. Not sure how happy I feel with these names but I've been staring at it for so long now it all looks :sunglasses: any feedback welcome: ping @aanand  @dnephin

aanand:
LGTM

PR #1736, opened 2015-07-20
Deprecate --allow-insecure-ssl

aanand:
Waiting on https://github.com/docker/docker-py/pull/681 to be merged and released. We'll be rid of "HTTPS endpoint unresponsive and insecure mode isn't enabled" once and for all.

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -26,6 +26,11 @@ from .utils import yesno, get_version_info
 
 log = logging.getLogger(__name__)
 
+INSECURE_SSL_WARNING = """
+Warning: --allow-insecure-ssl is deprecated and has no effect.
+It will be removed in a future version of Compose.
+"""
+
 
 def main():
     setup_logging()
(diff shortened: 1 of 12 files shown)


compose/cli/main.py
```diff
@@ -232,13 +237,13 @@ def pull(self, project, options):
         Usage: pull [options] [SERVICE...]
 
         Options:
-            --allow-insecure-ssl    Allow insecure connections to the docker
-                                    registry
+            --allow-insecure-ssl    Deprecated - no effect.
```
bfirsh: Can we remove from the help text altogether? That's what Engine does.

dnephin: I think it's better to keep it. Removing it immediately would break any scripts which include this flag. Leaving it here as a no-op gives people more time to remove it without having it break on upgrade.

aanand: Yeah, the drawback (as well as the benefit) of using `docopt` is that you can't have undocumented command line flags.

bfirsh: That sucks. Oh well.

aanand:
https://github.com/docker/docker-py/pull/681 is out in docker-py 1.3.1. I've updated `setup.py` and `requirements.txt`, so this is now ready to go.

mnowster:
LGTM

NigelThorne:
I can see you have removed the "insecure-registry" options, but what is the alternative? How do you run a registry without ssh, and then connect a docker instance to it?

Cheers
Nigel

aanand:
@NigelThorne You should be able to do that with no extra configuration. Are you having problems? If so, please open an issue!

PR #1744, opened 2015-07-21
Remove custom docker client initialization logic

aanand:
Use `docker.utils.kwargs_from_env` instead.

### DIFF
--- a/compose/cli/docker_client.py
+++ b/compose/cli/docker_client.py
@@ -1,8 +1,7 @@
 import os
-import ssl
 
 from docker import Client
-from docker import tls
+from docker.utils import kwargs_from_env
 
 
 def docker_client():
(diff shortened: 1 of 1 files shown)


compose/cli/docker_client.py
```diff
@@ -9,27 +8,9 @@ def docker_client():
     Returns a docker-py client configured using environment variables
     according to the same logic as the official Docker client.
     """
-    cert_path = os.environ.get('DOCKER_CERT_PATH', '')
-    if cert_path == '':
-        cert_path = os.path.join(os.environ.get('HOME', ''), '.docker')
```
dnephin: From what I can see `kwargs_from_env()` doesn't include this default for a cert path, is that correct?  Is this path no longer used?

aanand: Well spotted. Looks like that logic does indeed still exist in Docker:

- [read DOCKER_CERT_PATH](https://github.com/docker/docker/blob/master/docker/flags.go#L28)
- [set it to cliconfig.ConfigDir() if empty](https://github.com/docker/docker/blob/master/docker/flags.go#L76)
- [cliconfig sets it to $HOME/.docker if empty](https://github.com/docker/docker/blob/master/cliconfig/config.go#L33)

So I think we should add it to docker-py first - this PR isn't urgent.

aanand: https://github.com/docker/docker-py/pull/686

thaJeztah:
@aanand I see https://github.com/docker/docker-py/pull/686 was merged

aanand:
Yep - we can merge this as soon as the next docker-py release is out.

mnowster:
Closing in favour of https://github.com/docker/compose/pull/2078

PR #1746, opened 2015-07-21
Remove logging test

aanand:
It doesn't do much other than cause the remainder of the test suite to generate lots of junk output.


mnowster:
LGTM

PR #1755, opened 2015-07-22
Don't use custom name for one-off containers

### LINKED ISSUE #1753: container_name syntax breaks docker-compose run
When using the container_name syntax added in #1711 "docker-compose run" fails with a message like the following:

    Conflict. The name "myservice" is already in use by container bc5601faea86. You have to delete (or rename) that container to be able to reuse that name.

Example docker-compose.yml:

    myservice:
        container_name: myservice_name
        image: ubuntu:latest
        command: tail -f /dev/null

Commands run:

    docker-compose up -d
    docker-compose run --rm myservice /bin/sleep 3

"docker-compose run" is using the same directive to name its container, as can be verified by changing the container_name in docker-compose.yml after starting the service, and running:

    docker-compose run -d myservice tail -f /dev/null

I changed the service name to "myservice_name_another", which became the instantiated container's name.

aanand:
Closes #1753.

### DIFF
--- a/compose/service.py
+++ b/compose/service.py
@@ -577,8 +577,10 @@ class Service(object):
             for k in DOCKER_CONFIG_KEYS if k in self.options)
         container_options.update(override_options)
 
-        container_options['name'] = self.custom_container_name() \
-            or self.get_container_name(number, one_off)
+        if self.custom_container_name() and not one_off:
+            container_options['name'] = self.custom_container_name()
+        else:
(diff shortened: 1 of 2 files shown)


dnephin:
While this fixes the immediate problem. I know there are cases where I'd actually want this custom name for the one off. Maybe we could add `--name` to `docker-compose run` allowing naming of one-off containers?  I guess it doesn't need to happen right now.

dnephin:
LGTM

PR #1758, opened 2015-07-22
Improve scale test coverage

mnowster:


tests/integration/service_test.py
```diff
@@ -549,6 +549,119 @@ def test_scale(self):
...
+        container = service.create_container(number=next_number, quiet=True)
+        container.start()
+
+        self.assertTrue(container.is_running)
+        self.assertEqual(len(service.containers()), 1)
+
+        service.scale(1)
+
+        self.assertEqual(len(service.containers()), 1)
+        self.assertTrue(container.is_running)
```
aanand: I don't think `is_running` - or anything else on Container - updates unless you explicitly call `inspect()` on it again.

mnowster: Thanks for spotting this, I didn't realise that. Have updated.

aanand:
LGTM

PR #1759, opened 2015-07-22
Update install.md to reference 1.3.3

jgeiger:
Let people download the newest release when viewing the docs.


aanand:
Thanks. I already had a commit sitting around to update `install.md` and `CHANGES.md` which I hadn't pushed - thanks for reminding me! Replacing with #1762.

PR #1774, opened 2015-07-27
Contributors can build public docs with compose docs in context

moxiegirl:
- Copy down all the other public repo docs and build
- Changed base image
- Users building compose docs can see their changes in the context of the entire docs.docker.com

### DIFF
--- a/docs/Dockerfile
+++ b/docs/Dockerfile
@@ -6,6 +6,14 @@ COPY . /src
 
 COPY . /docs/content/compose/
 
+RUN svn checkout https://github.com/docker/docker/trunk/docs /docs/content/docker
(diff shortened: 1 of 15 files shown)


aanand:
LGTM

aanand:
Scratch that - `script/docs` is no longer working for me (I get `/bin/sh: 1: /src/pre-process.sh: not found` during the build step). Is it still the right way to build the docs?

moxiegirl:
@aanand script/docs looks like something that was used to support mkdocs.  To build documentation, go to docs directory and type `make docs`

aanand:
Okay. I'll remove it in a separate PR.

moxiegirl:
Thanks @aanand

PR #1502, opened 2015-06-03
Add volume_driver support to compose

lukemarsden:
https://github.com/docker/docker/pull/13161 landed in docker master.

This PR adds support for the `volume_driver` directive to compose, and avoids an ambiguity with a compose feature which expands local paths.

Depends on https://github.com/docker/docker-py/pull/628

Thanks to @aanand for your help with this!

PR #1777, opened 2015-07-28
Update API version to 1.19

aanand:
Required by #1502.

### DIFF
--- a/compose/cli/docker_client.py
+++ b/compose/cli/docker_client.py
@@ -14,7 +14,7 @@ def docker_client():
         cert_path = os.path.join(os.environ.get('HOME', ''), '.docker')
 
     base_url = os.environ.get('DOCKER_HOST')
-    api_version = os.environ.get('COMPOSE_API_VERSION', '1.18')
+    api_version = os.environ.get('COMPOSE_API_VERSION', '1.19')
 
     tls_config = None
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -48,16 +48,14 @@ RUN set -ex; \
     rm -rf pip-7.0.1; \
     rm pip-7.0.1.tar.gz
 
-ENV ALL_DOCKER_VERSIONS 1.6.2 1.7.1
+ENV ALL_DOCKER_VERSIONS 1.7.1
 
 RUN set -ex; \
-    curl https://get.docker.com/builds/Linux/x86_64/docker-1.6.2 -o /usr/local/bin/docker-1.6.2; \
-    chmod +x /usr/local/bin/docker-1.6.2; \
     curl https://get.docker.com/builds/Linux/x86_64/docker-1.7.1 -o /usr/local/bin/docker-1.7.1; \
     chmod +x /usr/local/bin/docker-1.7.1
 
 # Set the default Docker to be run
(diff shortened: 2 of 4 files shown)


compose/cli/docker_client.py
```diff
@@ -32,4 +32,4 @@ def docker_client():
         )
 
     timeout = int(os.environ.get('DOCKER_CLIENT_TIMEOUT', 60))
-    return Client(base_url=base_url, tls=tls_config, version='1.18', timeout=timeout)
+    return Client(base_url=base_url, tls=tls_config, version='1.19', timeout=timeout)
```
dnephin: I've been using `auto` here when I write things with `docker-py` (https://github.com/docker/docker-py/blob/master/docker/clientbase.py#L63). I know we've talked about this before.

If we change this to `1.19` everyone is forced to update to the latest dockerd even if they don't use the one minor feature that requires it.  Some users are going to want to upgrade to the latest docker-compose to pickup bug fixes and new features, without upgrading dockerd immediately.

I think it would be good to either use `auto` or support it with #1157

gheart:
I'd just like to add that I believe the variable name is more accurate and the placement of the environ.get is in a better spot in [#1780](https://github.com/docker/compose/pull/1780) . Also, I'm extremely passionate about this project and it would be a personal milestone for me to have contributed to the community for the first time (if that means anything which it might not).

compose/cli/docker_client.py
gheart: I'd just like to add that I believe the variable name is more accurate and the placement of the environ.get is in a better spot in [#1780](https://github.com/docker/compose/pull/1780). Also, I'm extremely passionate about this project and it would be a personal milestone for me to have contributed to the community for the first time (if that means anything which it might not). I would give anything to be the one to help with this.

aanand: `auto` is cool but introduces another HTTP request to every invocation of `docker-compose`, which I don't think is an acceptable slowdown.

As for making users upgrade: yeah, it sucks, but once Compose 1.4 is released, Docker 1.6 will be two versions old. Part of the reason we do lockstep releases of the whole platform is so that we don't have to cater to older versions (although we've been good about that so far).

I'm OK with adding the environment variable if it defaults to `1.19` and we tell Docker 1.6 users that they can set it back if they want but it won't be officially supported.

@gheart Your variable name is more descriptive, but inventing new `DOCKER_` variables that don't exist in the Docker client is not Compose's job, which is why we prefix ours with `COMPOSE_`.

Since you're eager to get a contribution in, if you rename the variable to `COMPOSE_API_VERSION`, and update your PR (squashed to one commit), I'd be happy to merge it. I'll then rebase this PR to update the default version to `1.19`.

gheart: Oh my gosh, thank you so much!!! Will do it right now!!

gheart: @aanand I've done just like you asked. I've renamed the env var to ```COMPOSE_API_VERSION``` and squashed to a single commit: [#1780](https://github.com/docker/compose/pull/1780/). Please don't hesitate to let me know if you need anything else.

aanand:
Rebased.

dnephin:
LGTM

aanand:
Tests pass locally.

PR #1778, opened 2015-07-28
Remove custom docs script

aanand:
Pass straight through to `cd docs && make docs`.

ping @moxiegirl - I figured it was best to keep the script around so there's always one reliable way to build the docs, should the workflow ever change again.


moxiegirl:
@aanand The one reliable way going forward should be `make docs`.

aanand:
OK, I've removed `script/docs`.

moxiegirl:
Thanks!

moxiegirl:
LGTM

Issue #271, opened 2014-06-25
Binary on darwin fails with "illegal instruction" on some machines


tomstuart:
The current `fig` release doesn’t work for me:

```
$ fig
-bash: fig: command not found

$ curl -L https://github.com/orchardup/fig/releases/download/0.4.2/darwin > /usr/local/bin/fig
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   329  100   329    0     0    329      0  0:00:01 --:--:--  0:00:01   329
100 3860k  100 3860k    0     0   625k      0  0:00:06  0:00:06 --:--:--  842k

$ chmod +x /usr/local/bin/fig

$ fig
Illegal instruction: 4
```

I’m running OS X 10.9.3 on a 1.6 GHz Intel Core 2 Duo (Late 2010 MacBook Air).

pwaller:
We hit this over here, too.

ramschmaerchen:
I'd really love to see docker-compose working on vanilla OSX.

fig: 1.1.0-rc2
python: 2.7.5
OS X: 10.9.5

Model Name: MacBook
Model Identifier: MacBook5,2 (Early 2009)
Processor Name: Intel Core 2 Duo
Processor Speed: 2 GHz
Number of Processors: 1
Total Number of Cores: 2
L2 Cache: 3 MB
Memory: 4 GB
Bus Speed: 1.07 GHz

NodeGuy:
:+1:

  Model Name:	MacBook Pro
  Model Identifier:	MacBookPro5,4
  Processor Name:	Intel Core 2 Duo
  Processor Speed:	2.53 GHz
  Number of Processors:	1
  Total Number of Cores:	2
  L2 Cache:	3 MB
  Memory:	4 GB
  Bus Speed:	1.07 GHz
  Boot ROM Version:	MBP53.00AC.B03
  SMC Version (system):	1.49f2

tim545:
$ sysctl -n machdep.cpu.brand_string
Intel(R) Core(TM) i7 CPU       M 640  @ 2.80GHz

Had the same issue but 
`brew install python`
`pip install -U pip`
fixed it for me

Not sure why but installing python by downloading their package installer (2.7.9) and install Fig with curl gave me the `Illegal instruction: 4` error when running `fig --version`

ramschmaerchen:
[...]

PR #1779, opened 2015-07-28
Add hint about OS X binary compatibility

aanand:
As reported by many people in #271, older Macs have trouble with the Compose binary. I've added a hint to the install instructions directing them to install using Pip instead.

ping @moxiegirl

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -35,6 +35,8 @@ To install Compose, run the following commands:
 Optionally, you can also install [command completion](completion.md) for the
 bash and zsh shell.
 
+> **Note:** Some older Mac OS X CPU architectures are incompatible with the binary. If you receive an "Illegal instruction: 4" error after installing, you should install using the `pip` command instead.
+
 Compose is available for OS X and 64-bit Linux. If you're on another platform,
 Compose can also be installed as a Python package:
 


docs/install.md
```diff
@@ -43,6 +43,8 @@ Compose can also be installed as a Python package:
 No further steps are required; Compose should now be successfully installed.
 You can test the installation by running `docker-compose --version`.
 
+> Note: If you get an error like "Illegal instruction: 4" after installing using the binary on Mac OS X, your CPU architecture isn't supported and you should install using Pip instead.
```
moxiegirl: @aanand move this note after the sentence starting "Optionally,..."

Change Note: to

 **Note**: Some older Mac OS X CPU architectures are incompatible with  installing the binary using `curl`.  If you receive an "Illegal instruction: 4" error after installing using `curl`, you should install using the `pip` command instead.

aanand: Done

thaJeztah:
LGTM, thx

josephpage:
:+1:

aanand:
@moxiegirl My only problem with this wording is it might imply that `curl` is the problem, not the binary. Do you think that's an issue?

moxiegirl:
@aanand yeah, you have a point now that I read it.  If you remove "with curl" I'm cool. LGTM

PR #1780, opened 2015-07-28
Allow API version specification via env var

gheart:
Hard-coding the API version to '1.18' with the docker-py constructor will
cause the docker-py logic at https://github.com/docker/docker-py/blob/master/docker/client.py#L143-L146
to always fail, which will cause authentication issues if you're using a
remote daemon using API version 1.19 (boot2docker running docker 1.7.1) - regardless of the API version of the registry.

Allow the user to set the API version via an environment variable. If the variable is not present, it will still default to '1.18,' like it does today.

### DIFF
--- a/compose/cli/docker_client.py
+++ b/compose/cli/docker_client.py
@@ -14,6 +14,8 @@ def docker_client():
         cert_path = os.path.join(os.environ.get('HOME', ''), '.docker')
 
     base_url = os.environ.get('DOCKER_HOST')
+    api_version = os.environ.get('COMPOSE_API_VERSION', '1.18')
+
     tls_config = None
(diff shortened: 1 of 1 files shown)


dnephin:
Duplicate of #1157 ?

gheart:
Dang, you beat me to it! This would have been my first open source contribution :( Oh well, I'm just glad that it's getting fixed for everyone!

gheart:
Although, I'd argue my variable names are more accurate :)

aanand:
CI is broken, but I've confirmed that the tests pass locally.

LGTM

dnephin:
LGTM

PR #1798, opened 2015-07-31
Alphabetise reference list

mbailey:
Bring in line with Glossary. https://docs.docker.com/reference/glossary/

Alphabetising list makes makes parsing by humans easier.

### DIFF
--- a/docs/yml.md
+++ b/docs/yml.md
@@ -19,15 +19,6 @@ As with `docker run`, options specified in the Dockerfile (e.g., `CMD`,
 `EXPOSE`, `VOLUME`, `ENV`) are respected by default - you don't need to
 specify them again in `docker-compose.yml`.
 
-### image
-
-Tag or partial image ID. Can be local or remote - Compose will attempt to
-pull if it doesn't exist locally.
-
-    image: ubuntu
-    image: orchardup/postgresql
-    image: a4bc65fd
-
 ### build
 
 Path to a directory containing a Dockerfile. When the value supplied is a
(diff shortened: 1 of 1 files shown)


aanand:
ping @moxiegirl 

I'm uncertain about this - while the current ordering is almost certainly suboptimal (it's the kind of thing that can very easily deteriorate with a PR-centric workflow), I'd rather prioritise the configuration options that people, especially beginners, are most likely to be looking for.

A closer analogue might be the [docker run reference](http://docs.docker.com/reference/run/), which groups options logically (something we could also consider).

moxiegirl:
@mbailey thank you for the contribution. 

@aanand references are generally arranged alphabetically -- this is helpful to experienced people moving fast and beginners who might be new to Compose but are not new to the alphabetical convention.  Teaching beginners the important elements of the configuration is a great idea. I think we should do it explicitly rather than relying on them to glean it from the page organization.  How about a section at the beginning   "For new Compose users" and then a quick list of tips... 

As for the `run` reference, I plan to fix that wagon. :-/

mbailey:
Thanks guys! I love living in a world where you can send a pull request for docs on a website and get fast feedback from real people. Can't wait for government to catch up with us. :-)

aanand:
OK, in that case this should be fine after a rebase. Looks like the source of the merge conflict is a68ee0d9c2d3969e161ca973cfbe3e62bcb3dd2e, so make sure those changes are still included!

mnowster:
@mbailey heya, can you update this PR with a rebase please? We can then give it final review and get it merged?

:heart: good docs

moxiegirl:
@aanand @mnowster Y'all want me to carry this one?

mnowster:
@moxiegirl s'ok, I'll pick it up and rebase it :)

mnowster:
Closing this in favour of the one I've rebased and made an additional change in. I've cherry-picked your commit so your contribution is still intact, thanks @mbailey  :)

https://github.com/docker/compose/pull/2066

PR #1799, opened 2015-07-31
Remove useless postgres 'port' configuration

d2bit:
If I'm not wrong `ports` should be `expose`, but the image is already exposing that port, so it could be cleaned.
I've seen it's already done in [Djanjo version](https://github.com/docker/compose/blob/master/docs/django.md).

BTW, why could be the reason to use `ports` without setting the host port number up? I really don't get that option :confused: 

Really nice work in composer, I like it so much!!

### DIFF
--- a/docs/rails.md
+++ b/docs/rails.md
@@ -40,8 +40,6 @@ Finally, `docker-compose.yml` is where the magic happens. This file describes th
 
     db:
       image: postgres
-      ports:
-        - "5432"
     web:
       build: .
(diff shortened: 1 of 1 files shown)


gponsu:
:+1:

aanand:
I think this is a holdover from before the `postgres` image existed and we were using an image that didn't expose the port by default.

LGTM

dnephin:
LGTM

Issue #1085, opened 2015-03-10
Epic: Windows support


bfirsh:
This is a meta-issue for keeping track of all the things we're doing around Windows support for Compose.

Various things that are part of this:

 - **Building a binary on Windows.** PyInstaller, the method we currently use to build binaries for Linux and OS X, also works with Windows. In theory this should just work. It would be useful to have build scripts etc to automate this.
 - **Run Compose inside a container.** This may be be easier than building a native binary for Windows. https://github.com/docker/compose/issues/169 https://github.com/docker/compose/pull/1806
 - **Making Compose Windows compatible.** I expect this will be mostly making sure paths work and [getting the TTYs to work](https://github.com/docker/compose/issues/417).
 - **Volume sharing.** When running Compose on OS X, it expects `/Users` to be mounted in the right place so volumes work. I wonder if we need something added to Machine/boot2docker to make volumes work as expected on Windows. ([The Compose quick start guide](http://docs.docker.com/compose/) has a demonstration of how volumes are used to mount code inside a container.)
 - **Run CI tests on Windows.** We don't all test on Windows, so we'd have to be careful about regressions. https://github.com/docker/compose/issues/886

Steps:
 - [x] Investigate whether binaries will build cleanly on Windows https://github.com/docker/compose/issues/1051
 - [x] Investigate whether running Compose inside a container will be easier than running it natively on Windows

Related issues:
 - https://github.com/docker/compose/issues/598

mikehaertl:
So you talk about a native Windows version of `docker-compose` here? 

As there isn't even a `docker` version for windows yet, I wonder if it wouldn't make more sense to focus on including `docker-compose` into the boot2docker VM for now:

https://github.com/boot2docker/boot2docker/issues/603

This is something that blocks many developers on Windows machines. And even though there are workarounds, I'd consider them hacks. Windows users should not have such a hard time to work with `docker-compose`.

funkyfuture:
Using Linux containers on Windows. Yeah, that's epic.

dduportal:
@mikehaertl The docker client is not a requirement for docker-compose (which embed its own python-written client). For example, we use https://github.com/rhuss/docker-maven-plugin in my companyy, which can contact our boot2docker, but running a JVM process in Windows.

So there are two main use cases :
* Running Compose inside boot2docker (at OS level or inside a container)
* Running compose from your Host OS

=> In the two cases, compiling and running is just some plumbing. The main goal will be making that usable.
[...]

PR #1806, opened 2015-08-03
Docker image and bash script for running compose in a container

### LINKED ISSUE #169: Epic: Make Compose a Docker container
$ docker run compose up

**Update, 2015-03-10:** #619 is the first step towards this, but we still have to pass loads of options to `docker run` to make it work properly. Next step is figuring out if we can make any changes to Docker so we don't have to pass these options – e.g. some way of setting default options for an image, some way of passing the Docker socket through to the container, etc.

### LINKED ISSUE #1662: Avoid Managing a non-containerized Python Runtime
Docker Compose is great, but if you have to manage a Python runtime outside of a container, you're losing some of the value of Docker itself. I've created [a proof-of-concept of containerizing docker-compose itself](https://github.com/kojiromike/docker-compose-wrapper), so that you can use docker-compose from within Docker, thus avoiding the need to manage Python dependencies outside a container. All you need is docker (and, frankly, bash, but that can be dealt with.)

Anyway, we're using this now. Hope you like it.

dnephin:
Resolves #169
Resolves #1662

The image is ~50Mb.

Thoughts?

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -40,19 +40,37 @@ To install Compose, do the following:
 
         curl -L https://github.com/docker/compose/releases/download/VERSION_NUM/docker-compose-`uname -s`-`uname -m` > /usr/local/bin/docker-compose
 
-     If you have problems installing with `curl`, you can use `pip` instead: `pip install -U docker-compose`
+     If you have problems installing with `curl`, see
+     [Alternative Install Options](#alternative-install-options).
 
-4. Apply executable permissions to the binary:
+5. Apply executable permissions to the binary:
 
         $ chmod +x /usr/local/bin/docker-compose
 
-5.  Optionally, install [command completion](completion.md) for the
+6.  Optionally, install [command completion](completion.md) for the
 `bash` and `zsh` shell.
 
-6. Test the installation.
+7. Test the installation.
 
         $ docker-compose --version
         docker-compose version: 1.4.2
+
+
+## Alternative install options
+
+### Install using pip
+
+    $ sudo pip install -U docker-compose
+
+
+### Install as a container
+
+Compose can also be run inside a container, from a small bash script wrapper.
+To install compose as a container run:
+
+    $ curl -L https://github.com/docker/compose/releases/download/1.5.0/compose-run > /usr/local/bin/docker-compose
+    $ chmod +x /usr/local/bin/docker-compose
+
 
 ## Upgrading
 
--- a/script/run
+++ b/script/run
@@ -1,0 +1,48 @@
+#!/bin/bash
+#
+# Run docker-compose in a container
+#
+# This script will attempt to mirror the host paths by using volumes for the
+# following paths:
+#   * $(pwd)
+#   * $(dirname $COMPOSE_FILE) if it's set
+#   * $HOME if it's set
+#
+# You can add additional volumes (or any docker run options) using
+# the $COMPOSE_OPTIONS environment variable.
+#
+
+
+set -e
+
+VERSION="1.4.0dev"
+# TODO: move this to an official repo
+IMAGE="dnephin/docker-compose:$VERSION"
+
+
+# Setup options for connecting to docker host
+if [ -z "$DOCKER_HOST" ]; then
+    DOCKER_HOST="/var/run/docker.sock"
+fi
+if [ -S "$DOCKER_HOST" ]; then
+    DOCKER_ADDR="-v $DOCKER_HOST:$DOCKER_HOST -e DOCKER_HOST"
+else
+    DOCKER_ADDR="-e DOCKER_HOST"
+fi
+
+
+# Setup volume mounts for compose config and context
+VOLUMES="-v $(pwd):$(pwd)"
+if [ -n "$COMPOSE_FILE" ]; then
+    compose_dir=$(dirname $COMPOSE_FILE)
+fi
+# TODO: also check --file argument
+if [ -n "$compose_dir" ]; then
+    VOLUMES="$VOLUMES -v $compose_dir:$compose_dir"
+fi
+if [ -n "$HOME" ]; then
+    VOLUMES="$VOLUMES -v $HOME:$HOME"
+fi
+
+
+exec docker run --rm -ti $DOCKER_ADDR $COMPOSE_OPTIONS $VOLUMES -w $(pwd) $IMAGE $@
--- a/.dockerignore
+++ b/.dockerignore
@@ -4,6 +4,6 @@
 .tox
 build
 coverage-html
-dist
 docs/_site
 venv
+.tox
--- a/Dockerfile.run
+++ b/Dockerfile.run
@@ -1,0 +1,15 @@
+
+FROM    alpine:edge
+RUN     apk -U add \
+            python \
+            py-pip
+
+COPY    requirements.txt /code/requirements.txt
+RUN     pip install -r /code/requirements.txt
+
+ENV     VERSION 1.4.0dev
+
+COPY    dist/docker-compose-$VERSION.tar.gz /code/docker-compose/
+RUN     pip install /code/docker-compose/docker-compose-$VERSION/
+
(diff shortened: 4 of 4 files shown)


script/run
```diff
@@ -0,0 +1,48 @@
...
+VOLUMES="-v $(pwd):$(pwd)"
+if [ -n "$COMPOSE_FILE" ]; then
+    compose_dir=$(dirname $COMPOSE_FILE)
+fi
+# TODO: also check --file argument
+if [ -n "$compose_dir" ]; then
+    VOLUMES="$VOLUMES -v $compose_dir:$compose_dir"
+fi
+if [ -n "$HOME" ]; then
+    VOLUMES="$VOLUMES -v $HOME:$HOME"
```
aanand: Does everything work fine if `$(pwd)` is a subdir of `$HOME` and we get overlapping volume mounts?

dnephin: Ya, that doesn't seem to be a problem from what I can tell

script/run
```diff
@@ -0,0 +1,48 @@
...
+fi
+
+
+exec docker run -ti $DOCKER_ADDR $COMPOSE_OPTIONS $VOLUMES -w $(pwd) $IMAGE $@
```
dnephin: I guess this needs `--rm` as well

bfirsh:
Interesting.

Would this make Windows support easier? Paths, TTYs, etc... https://github.com/docker/compose/issues/1085

dnephin:
It would run easily enough in cygwin I think, but I don't think windows comes with bash, right?

We could re-create this bash script in golang, and publish it as a statically linked binary, which might make it more portable.

bfirsh:
Sorry - yes. Assuming we have a similar script or binary running on Windows that runs Compose in a container, would Compose running inside a container work on Windows?

aanand:
Yes, it would. The other assumption is that volume mounting would 'just work', as it does on OS X (as long as you're within your home directory).

docs/install.md
```diff
@@ -42,6 +47,21 @@ Compose can also be installed as a Python package:
 
     $ sudo pip install -U docker-compose
 
+### Container install
+
+Compose can also be run inside a container, from a small bash script wrapper.
+To install compose as a container run:
+
+    curl -L https://github.com/docker/compose/releases/download/1.3.3/compose-run > /usr/local/bin/docker-compose
```
thaJeztah: What if we added compose-run to the image itself, using docker as the distribution mechanism, and a small install script to echo/output it, e.g.;

```
# To install the script wrapper, run
$ docker run --rm docker-compose install > /usr/local/bin/docker-compose
$ chmod +x /usr/local/bin/docker-compose

# Validate it's working
$ docker-compose --version
docker-compose version: 1.4.0rc3
```

dnephin: That's a cool idea, I think that could work.

aanand: That is a cool idea. We should probably put an explicit version tag in, so we can change the run script if we need to.

bfirsh: Could it be a separate image? If you install Compose with the Toolbox, the "install" command might be confusing.

dnephin: Ya, since this command is really just printing out the bash script, we could probably call it `echo-run-script` or something like that?

thaJeztah: The advantage of it being in the same image, is that the image is pulled during "install". The first "run" is instant then :-)

mnowster:
@dnephin if you can do a rebase, I'll review.

dnephin:
Rebased

docs/install.md
```diff
@@ -40,20 +40,38 @@ To install Compose, do the following:
 
         curl -L https://github.com/docker/compose/releases/download/VERSION_NUM/docker-compose-`uname -s`-`uname -m` > /usr/local/bin/docker-compose
 
-     If you have problems installing with `curl`, you can use `pip` instead: `pip install -U docker-compose`
+     If you have problems installing with `curl`, see alternative install
+     options below.
```
mnowster: Minor edit: Can we make this a link to the `alternative install` section? Then we can remove the, 'below'.

docs/install.md
```diff
@@ -40,20 +40,38 @@ To install Compose, do the following:
...
 `bash` and `zsh` shell.
 
-6. Test the installation.
+7. Test the installation.
 
         $ docker-compose --version
         docker-compose version: 1.4.2
 
+
+## Alternative Install Options
```
mnowster: We don't want `Install` and `Options` to be capitalised here.

docs/install.md
dnephin: Ya, it seems like there is no standard way to do that with markdown?  I guess I have to do something like this https://gohugo.io/extras/crossreferences/ , I'll try it out

script/run
```diff
@@ -0,0 +1,48 @@
+#!/bin/bash
+#
+# Run docker-compose in a container
+#
+# This script will attempt to mirror the host paths by using volumes for the
```
mnowster: You write really helpful doc strings for your bash scripts :+1:

docs/install.md
dnephin: I tested it out locally, seems to work

docs/install.md
```diff
@@ -40,20 +40,38 @@ To install Compose, do the following:
...
+
+### Source install
+
+    $ sudo pip install -U docker-compose
+
+
+### Container install
```
mnowster: I would suggest `Install as a container` as a more descriptive header.

docs/install.md
```diff
@@ -40,20 +40,38 @@ To install Compose, do the following:
...
-6. Test the installation.
+7. Test the installation.
 
         $ docker-compose --version
         docker-compose version: 1.4.2
 
+
+## Alternative Install Options
+
+### Source install
```
mnowster: How do you feel if we just put `pip install` here. `Source` makes me think of downloading zip files, make and build scripts.

docs/install.md
```diff
@@ -40,20 +40,38 @@ To install Compose, do the following:
...
+
+    $ sudo pip install -U docker-compose
+
+
+### Install as a container
+
+Compose can also be run inside a container, from a small bash script wrapper.
+To install compose as a container run:
+
+    curl -L https://github.com/docker/compose/releases/download/1.5.0/compose-run > /usr/local/bin/docker-compose
```
mnowster: I think this is missing a $ at the beginning so it'll get rendered as a command.

dnephin:
Cool, I've made those changes

mnowster:
LGTM

mnowster:
Come on janky :racehorse: :horse_racing:

script/run
```diff
@@ -0,0 +1,48 @@
...
+# You can add additional volumes (or any docker run options) using
+# the $COMPOSE_OPTIONS environment variable.
+#
+
+
+set -e
+
+VERSION="1.4.0dev"
+# TODO: move this to an official repo
+IMAGE="dnephin/docker-compose:$VERSION"
```
dnephin: I'll open an issue to get this image published to a better repo, and add notes to the release process doc about updating it on release.

script/run
```diff
@@ -0,0 +1,48 @@
...
+fi
+if [ -S "$DOCKER_HOST" ]; then
+    DOCKER_ADDR="-v $DOCKER_HOST:$DOCKER_HOST -e DOCKER_HOST"
+else
+    DOCKER_ADDR="-e DOCKER_HOST"
+fi
+
+
+# Setup volume mounts for compose config and context
+VOLUMES="-v $(pwd):$(pwd)"
```
kojiromike: Why not use "$PWD" (which will be set by the shell) and avoid the subshells here?

dnephin: That would probably work as well.  There are times when `$PWD` isn't set (if you fork/exec without a shell for example).

Does this cause a problem?

kojiromike: No serious problem, just calling out the unnecessary subshell.

PR #1808, opened 2015-08-04
129 validate compose yml

mnowster:
Introducing schema validation for compose yml.

Utilising jsonschema, we have a defined schema we can pass to it so it can validate the config that the user has specified.

The validation errors that are raised replicate existing behaviour and print out human readable error messages. It also expands to cover new scenarios and provide helpful error messages rather than stack traces.

Functionality in here should address each of the points raised in this issue: https://github.com/docker/compose/issues/129

:sparkles: :cat2:


aanand:
Looking great.

thaJeztah:
Nice!! thank you :-)

mnowster:
@aanand ok, I've squashed commits as far as I'm happy too without it becoming difficult to follow, imo.

compose/config/schema.json
```diff
@@ -0,0 +1,144 @@
...
+        "mac_address": {"type": "string"},
+        "mem_limit": {"type": "number"},
+        "memswap_limit": {"type": "number"},
+        "name": {"type": "string"},
+        "net": {"type": "string"},
+        "pid": {"type": "string"},
+
+        "ports": {
+          "oneOf": [
+            {"type": "string", "format": "ports"},
```
aanand: We don't actually support a string value for `ports` - it must be a list.

funkyfuture:
ciao @mnowster, as i was already working on a [config-validation](#1355), and formulated a lot of invalid dictionaries on that behalf. i tested these against your implementation and *80* of them failed. (and it's a total of 80, what baffles me the more. but i checked with debug-statements twice.) 
shall i open a pull request on your repo to integrate these?

i'm in a mix of feelings here. i was about to do some work on my implementation, respectively the used validation-library first. (it deserves a better handling of error-messages incl. l11n.) another point that imo makes that library preferable is the possibility to use yaml for the schema, but that may be possible with jsonschema too.

on the other hand i'm relieved that i don't necessarily have that on my agenda anymore. and your implementation certainly is less invasive to the config-module than mine. seems to have gotten some proper refactoring in the meantime. :-)

funkyfuture:
oh, the heat here… i found out that my tests do not cover the code, as they assume that validation happens after calling `ServiceLoader.make_service_dict`. 

and it's to be criticized that it doesn't, as config-mixins are left out of validation. furthermore it makes more sense if you use the `ServiceLoader` in client-code.

i'll look into the test-behaviour with `config.load` when it cooled down.

funkyfuture:
i opened a pull request with test-generators that will test a bunch of valid and invalid configurations. of these tests, 42 are failing.

i also roughly reviewed what i did in #1355 so far and want to emphasize which checks are imo missing here. my paradigm for that is the assumption that any configuration-error can possibly lead to data-loss and thus any validation must take racetime-conditions into consideration and anticipate any possible human- or machine-generated mistake.

- validate complete configuration including anything resolved into it (paths, extends-, environment-files and such)
- validate values for keys with a restricted set of allowed values (capabilities, net, memory, etc)
- validate accessibility for referenced filesystem-objects
- validate legal posix-paths for mount-points in containers
- validate network-configuration like hostnames, ips, matching port-ranges, …

furthermore a nice-to-have:
- propose keys that best fuzzy-match an invalid config-key

mnowster:
Hi @funkyfuture,

Thank you for your contribution to the initial spike on the approach.

I reviewed both your initial PR and draghuram's with @aanand and @bfirsh, and we decided on the jsonschema approach. An initial scope was agreed that we were seeking to replace stack traces with better error messages but doesn't go much further than what we already with existing validation functionality, eg nothing extra for network config as the docker daemon takes care of this.

The requirement to get validation schema in for 1.5 release was increased in priority, as is the way of open source projects, so I took over the issue.

Thanks again for contributing.

:balloon:

mnowster:
ping @aanand , I've finished fixing up a few issues. We should be good to go now, unless you find anything :palm_tree:

bfirsh:
Does the schema get included in the Python source package? It might need adding to MANIFEST.in or something or rather...

funkyfuture:
propably it's not a good idea to discuss this here, but i don't want to open a new issue unless you actually regard my concerns as an issue (or if you prefer: user story).

first of all, let me understand you correctly @mnowster:
> The requirement to get validation schema in for 1.5 release was increased in priority, as is the way of open source projects

there surely are diverse ways to develop an open-source-project. do you rather mean "as is the way of agile development"? that'd make sense to me. otherwise i'd respond that there is always an alternative. if you mean "agile" and if this is the paradigm, me and my team will reconsider whether Compose is still the horse to bet on for orchestration in our production environments.

> eg nothing extra for network config as the docker daemon takes care of this.

i remember this argument, but i also recall that it was regarded as a non-satisfying answer, because the Docker daemon handles single containers, not a set of 'em. thus, a misconfigured composition of containers can run into unknown outcomes if one container fails to start, but the others are started anyway.

to not validate anything that get's mixed in is imo a significant design flaw.

my perspective on this is one of a paranoid and extra-cautious sysadmin. is this perspective relevant for y'all? if so, what do you propose how to deal with it?

aanand:
@funkyfuture You're right - this isn't the place to discuss whether or not to validate things on the client that are already validated on the daemon. That's a complicated question, and you should feel free to open a separate issue for that.

As @mnowster said, this PR is an incremental step that should result in a significant reduction in cryptic stack traces for Compose users. We don't need to have a completely laid-out plan for our general approach to validation in order to merge it.

mnowster:
@bfirsh ah, that is a good point, I'll look into it, thank you.

aanand:
LGTM

PR #1783, opened 2015-07-29
WIP: Use dockerswarm/dind image instead of doing docker-in-docker ourselves

aanand:
This simplifies the test suite, removing a layer of nesting and allowing us to abstract away the philosophical terror of Docker-in-Docker. Unfortunately I get intermittent connection errors:

```
$ script/test

(...build output...)

Running lint checks
Running tests against Docker 1.7.1
++ docker run -d --privileged --volume=/var/lib/docker --expose=2375 dockerswarm/dind:1.7.1 docker -d -H tcp://0.0.0.0:2375
+ daemon_container_id=ebebf491e47a54e41b9221d072720a1cfc7f3c892c91079981e19f695c12c6b7
+ docker run --rm --link=ebebf491e47a54e41b9221d072720a1cfc7f3c892c91079981e19f695c12c6b7:docker --env=DOCKER_HOST=tcp://docker:2375 --entrypoint=nosetests docker-compose:2759ab5
.........stty: standard input: Inappropriate ioctl for device
.stty: standard input: Inappropriate ioctl for device
.stty: standard input: Inappropriate ioctl for device
........................................................................................................................E.........E.........................................................................................................................................................................................................
======================================================================
ERROR: test_start_container_builds_images (tests.integration.service_test.ServiceTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/code/tests/integration/service_test.py", line 431, in test_start_container_builds_images
    self.assertIn('success', container.logs())
  File "/code/compose/container.py", line 158, in logs
    return self.client.logs(self.id, *args, **kwargs)
  File "/usr/local/lib/python2.7/site-packages/docker/utils/decorators.py", line 20, in wrapped
    return f(self, resource_id, *args, **kwargs)
  File "/usr/local/lib/python2.7/site-packages/docker/client.py", line 561, in logs
    return self._get_result(container, stream, res)
  File "/usr/local/lib/python2.7/site-packages/docker/clientbase.py", line 242, in _get_result
    cont = self.inspect_container(container)
  File "/usr/local/lib/python2.7/site-packages/docker/utils/decorators.py", line 20, in wrapped
    return f(self, resource_id, *args, **kwargs)
  File "/usr/local/lib/python2.7/site-packages/docker/client.py", line 482, in inspect_container
    self._get(self._url("/containers/{0}/json".format(container))),
  File "/usr/local/lib/python2.7/site-packages/docker/clientbase.py", line 86, in _get
[...]

PR #1815, opened 2015-08-06
Abort tests if daemon fails to start

aanand:
Currently the test suite hangs if our Docker-in-Docker daemon fails to start, which is pretty useless. This makes it abort if that happens and print the daemon logs so we can see what's gone wrong.

Longer term we should switch to using the dind image - see #1783 - but that's not quite working well enough yet.


dnephin:
LGTM

PR #1817, opened 2015-08-06
Update volume tests for clarity

aanand:
- Better method names.

- Environment variable syntax in volume paths, even when a driver is specified, now *will* be processed (the test wasn't testing it  properly). However, `~` will still *not* expand to the user's home  directory.


mnowster:
LGTM

PR #1825, opened 2015-08-07
Avoid duplicate warnings if an unset env variable is used multiple times

aanand:
Suppose a `docker-compose.yml` has a section like this:

```yaml
labels:
  - "foo=${FOO}"
  - "bar=${BAR}"
  - "foobar=${FOO}${BAR}"
```

Currently, Compose will print 4 warnings when any `docker-compose` command is run:

```
$ docker-compose up
The FOO variable is not set. Substituting a blank string.
The BAR variable is not set. Substituting a blank string.
The FOO variable is not set. Substituting a blank string.
The BAR variable is not set. Substituting a blank string.
...
```

This changes it so we only print one warning per variable:

```
$ docker-compose up
The FOO variable is not set. Substituting a blank string.
The BAR variable is not set. Substituting a blank string.
...
```

### DIFF
--- a/compose/config/interpolation.py
+++ b/compose/config/interpolation.py
@@ -10,13 +10,15 @@ log = logging.getLogger(__name__)
 
 
 def interpolate_environment_variables(config):
+    mapping = BlankDefaultDict(os.environ)
+
     return dict(
-        (service_name, process_service(service_name, service_dict))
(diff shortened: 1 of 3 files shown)


mnowster:
LGTM

mnowster:
I do like features that make helpful messages more helpful. Command line tool design FTW :speedboat:

PR #1833, opened 2015-08-10
Show a warning when a relative path is specified without "./"

### LINKED ISSUE #1823: Support named volumes
Currently, Compose will treat the part of a `volumes` entry before the `:` (if present) in two different ways:

- If you've specified a `volume_driver`, it will treat it as a *volume name* and pass it straight through to the Docker API.

- Otherwise, it will treat it as a filesystem path and perform expansion of relative paths and the `~` character.

However, Docker will eventually be adding support for volumes as a top-level concept (`docker volumes`), and therefore implicitly for *named local volumes* - e.g. a volume named `data` which refers to an arbitrary filesystem path on the host machine, which you can mount with `docker run -v data:/data`.

The Docker daemon decides whether to treat the value as a path or a volume name based on [whether it's an absolute path](https://github.com/docker/docker/blob/72c04ab87ce16ab592e66da1753c0dc6cdc89517/daemon/volumes_unix.go#L404-L411), which on Unix means [it starts with a slash](https://github.com/golang/go/blob/2adc4e892704bb99f2d16c0c09777987cb6bed35/src/path/filepath/path_unix.go#L11-L14) and on Windows is, uh, [complicated](https://github.com/golang/go/blob/2adc4e892704bb99f2d16c0c09777987cb6bed35/src/path/filepath/path_windows.go#L16-L27). But we've got an extra challenge because Compose allows relative paths and `~`. We need a way of deciding whether or not to perform path expansion before passing the value to the API.

We have a number of options. In increasing order of my personal preference:

1. Try passing the value straight through to the API. If it fails, perform expansion and try again. I don't like this because it complicates the client-side flow a lot (we can't know what the right value is until it's time to create a container) and requires multiple API calls.

2. Perform expansion on the value, see if it's a valid local path, and use that if so; otherwise, pass it straight through. I don't like this because it means we'll do completely different things if you misspell a path, forget to create it etc.

3. Treat the value as a path if it contains (or perhaps *starts with*) any of `.~/`. This is simple enough, but will break backwards compatibility. Entries like this one will now be treated differently:

    ```yaml
    volumes:
      - data:/data
    ```

    Where this previously meant "mount the `data` subdirectory of the current directory", it would now mean "mount the volume named `data`".

    If we go down this route, it might be best to put out a deprecation warning in the next release whenever we see host paths that don't start with one of the special characters, so users have a release cycle to update their YAML (i.e. to change `data` to `./data`), and then make the breaking change in the subsequent release.

    If we do *that*, we need to do it now - 1.4 will be out very soon.

Thoughts?

aanand:
Closes #1823 by implementing the third option described there.

### DIFF
--- a/compose/config/config.py
+++ b/compose/config/config.py
@@ -90,6 +90,13 @@ SUPPORTED_FILENAMES = [
 ]
 
 
+PATH_START_CHARS = [
+    '/',
+    '.',
+    '~',
+]
+
+
 log = logging.getLogger(__name__)
 
 
@@ -260,7 +267,7 @@ def process_container_options(service_dict, working_dir=None):
         raise ConfigurationError("Invalid 'memswap_limit' configuration for %s service: when defining 'memswap_limit' you must set 'mem_limit' as well" % service_dict['name'])
 
     if 'volumes' in service_dict and service_dict.get('volume_driver') is None:
-        service_dict['volumes'] = resolve_volume_paths(service_dict['volumes'], working_dir=working_dir)
+        service_dict['volumes'] = resolve_volume_paths(service_dict, working_dir=working_dir)
 
     if 'build' in service_dict:
         service_dict['build'] = resolve_build_path(service_dict['build'], working_dir=working_dir)
@@ -421,17 +428,31 @@ def env_vars_from_file(filename):
     return env
 
 
-def resolve_volume_paths(volumes, working_dir=None):
+def resolve_volume_paths(service_dict, working_dir=None):
     if working_dir is None:
         raise Exception("No working_dir passed to resolve_volume_paths()")
 
-    return [resolve_volume_path(v, working_dir) for v in volumes]
+    return [
+        resolve_volume_path(v, working_dir, service_dict['name'])
+        for v in service_dict['volumes']
+    ]
 
 
-def resolve_volume_path(volume, working_dir):
+def resolve_volume_path(volume, working_dir, service_name):
     container_path, host_path = split_path_mapping(volume)
     container_path = os.path.expanduser(container_path)
+
     if host_path is not None:
+        if not any(host_path.startswith(c) for c in PATH_START_CHARS):
+            log.warn(
+                'Warning: the mapping "{0}" in the volumes config for '
+                'service "{1}" is ambiguous. In a future version of Docker, '
+                'it will designate a "named" volume '
+                '(see https://github.com/docker/docker/pull/14242). '
+                'To prevent unexpected behaviour, change it to "./{0}"'
+                .format(volume, service_name)
+            )
+
         host_path = os.path.expanduser(host_path)
         return "%s:%s" % (expand_path(working_dir, host_path), container_path)
     else:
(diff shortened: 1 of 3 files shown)


tests/unit/config_test.py
```diff
@@ -117,6 +117,33 @@ def test_volume_binding_with_home(self):
...
+
+            make_dict(volumes=['data:/container/path'], volume_driver='mydriver')
+            self.assertEqual(0, warn.call_count)
+
+            make_dict(volumes=['data:/container/path'])
+            self.assertEqual(1, warn.call_count)
+            warning = warn.call_args[0][0]
+            self.assertIn('"data:/container/path"', warning)
+            self.assertIn('"./data:/container/path"', warning)
+
```
thaJeztah: are, e.g. `.profile`, `~tmp`, `~`, `.` valid paths, or should they always contain a slash? (just thinking of some variations)

aanand: Good question. I'd say `~` and `.` are fine, but `.profile` and `~tmp` should be prefixed. Will fix.

thaJeztah: thx!

aanand: Thinking about it, it's better to keep the logic simple - "if the volume starts with `[.~/]`, we treat it as a path - otherwise, we treat it as a name" - as it's something users have to worry about. So I'm going to rule that all of those values are valid paths.

thaJeztah: you may be right; I don't think `~tmp` would be a valid name for a volume, so should be fine

compose/config/config.py
```diff
@@ -431,7 +438,15 @@ def resolve_volume_paths(volumes, working_dir=None):
 def resolve_volume_path(volume, working_dir):
     container_path, host_path = split_path_mapping(volume)
     container_path = os.path.expanduser(container_path)
+
     if host_path is not None:
+        if not any(host_path.startswith(c) for c in PATH_START_CHARS):
+            log.warn(
```
bfirsh: It isn't clear this warning is about a volume - maybe worth saying it's in the "volumes" section of a specific service?

Might also be explaining why or linking off to an issue explaining why.

aanand: Agreed - I've updated it so that it specifies the service and links to https://github.com/docker/docker/pull/14242.

mnowster:
LGTM

thaJeztah:
Thanks @aanand

compose/config/config.py
```diff
@@ -421,17 +428,31 @@ def env_vars_from_file(filename):
...
+    ]
 
 
-def resolve_volume_path(volume, working_dir):
+def resolve_volume_path(volume, working_dir, service_name):
     container_path, host_path = split_path_mapping(volume)
     container_path = os.path.expanduser(container_path)
+
     if host_path is not None:
+        if not any(host_path.startswith(c) for c in PATH_START_CHARS):
```
dnephin: https://docs.python.org/2/library/stdtypes.html#str.startswith

Pretty minor, `startswith()` also accepts a tuple, so this could be

```python
if not host_path.startswith(PATH_START_CHARS):
    ...
```

(if `PATH_START_CHARS` is changed to tuple)

dnephin:
I'm late to the party, but also LGTM. I think this is the right way to handle it.

compose/config/config.py
thaJeztah: nice! (complete python n00b ;-))

karlkfi:
This has an undesirable effect when combined with environment variables: #1867

PR #1836, opened 2015-08-10
Use overlay driver in tests

aanand:
Using the overlay driver for the Docker daemon we start in our test script will hopefully make CI more stable.


bfirsh:
LGTM

PR #1843, opened 2015-08-11
Add privileged to avoid volume permission denied

kunyan:


kunyan:
I think add "privileged:true" is not the best solution.
because it will make container have same permission as host
btw, when user use manage.py create something, the file is belong to root....(but we want it belong to current user)
it seems not have a perfect solution for this case.

dnephin:
I'm not sure why this is necessary. Why do you need privileged here?

kunyan:
@dnephin  When I create a django project follow the guide https://docs.docker.com/compose/django/, I found when I executed
```
docker-compose run web django-admin.py startproject composeexample .
```

some files created by django are belong  to root, I can't edit them. so I'm trying to find out a way to make them belong to current user. do you have any idea about this ?

I have try again, add "privileged:true" is not necessary, only add "user: '1000:1000'" is good enough.

dnephin:
One way to deal with that is to create a user in the container with the same uid as your host user. Then use the `USER` Dockerfile directive to set that user as the active user.

Another option is to make sure that the files are created on the host first, before compose runs.

Another way is to use a container to edit the files as root.

I'm going to close this PR (since it doesn't fix the issue), and open a new issue (#1936) about updating some of these examples.

PR #1846, opened 2015-08-11
Fix mem_limit and memswap_limit regression

aanand:
Since API version 1.19, they must be specified in HostConfig.


dnephin:
LGTM

PR #1852, opened 2015-08-12
Fix ports validation

mnowster:
I had misunderstood the valid formats allowed for ports. They must
always be in a list.


aanand:
LGTM

PR #1853, opened 2015-08-12
Only use overlay driver in CI

aanand:
People using Linux machines to work on Compose might not have overlay installed.


dnephin:
LGTM

Issue #1845, opened 2015-08-11
TypeError when service name is a number


juliantaylor:
```
1:
    build: .
```

leads to a type error and not a useful error message like invalid name

```
Traceback (most recent call last):
  File "<string>", line 3, in <module>
  File "/code/build/docker-compose/out00-PYZ.pyz/compose.cli.main", line 39, in main
  File "/code/build/docker-compose/out00-PYZ.pyz/compose.cli.docopt_command", line 21, in sys_dispatch
  File "/code/build/docker-compose/out00-PYZ.pyz/compose.cli.command", line 27, in dispatch
  File "/code/build/docker-compose/out00-PYZ.pyz/compose.cli.docopt_command", line 24, in dispatch
  File "/code/build/docker-compose/out00-PYZ.pyz/compose.cli.command", line 57, in perform_command
  File "/code/build/docker-compose/out00-PYZ.pyz/compose.cli.command", line 79, in get_project
  File "/code/build/docker-compose/out00-PYZ.pyz/compose.project", line 85, in from_dicts
  File "/code/build/docker-compose/out00-PYZ.pyz/compose.service", line 85, in __init__
  File "/code/build/docker-compose/out00-PYZ.pyz/re", line 141, in match
TypeError: expected string or buffer
```

aanand:
@mnowster: Will #1808 catch this?

mnowster:
@aanand yes, only these chars are allowed `[a-zA-Z0-9._-]`

juliantaylor:
a number matches that pattern

mnowster:
sorry, yes you're absolutely right @juliantaylor , I need to drink some more coffee before re-reading my regexes.

mnowster:
I'll add to my list to fix :+1:

juliantaylor:
I assume it could be a valid service name, compose might just need to convert integer dictionary keys to strings.

mnowster:
@juliantaylor yep, I think it could be a valid service name. This will get sorted in a separate PR rather than #1808 
Thanks for reporting the issue!

mnowster:
Fixed by https://github.com/docker/compose/pull/1855 but I didn't put the fixes in the comment so it didn't close this. *face-palm*

PR #1855, opened 2015-08-12
A number is a valid service name

mnowster:
While we technically allow a service name to be a number, it was causing the code
to blow up with a TypeError, as reported https://github.com/docker/compose/issues/1845. 

This is due to it expecting a string or a buffer. So I took the approach to ensure all
service names get converted to strings.

### DIFF
--- a/compose/config/config.py
+++ b/compose/config/config.py
@@ -13,7 +13,11 @@ from .errors import (
     CircularReference,
     ComposeFileNotFound,
 )
-from .validation import validate_against_schema
+from .validation import (
+    validate_against_schema,
+    validate_service_names,
+    validate_top_level_object
+)
 
 
 DOCKER_CONFIG_KEYS = [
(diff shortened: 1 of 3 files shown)


dnephin:
I was thinking the problem is that yaml treats a number as an integer, instead of a string. So the developer should probably write:

```yaml
'2': 
    ...
```

I'm not sure about casting these to strings. Maybe we could just warn the developer of the type and have them fix the yaml?

mnowster:
@dnephin Yeh I was in two minds about it myself, partly because I didn't want to force the user to do extra work for something that feels like it *should* work but then conflicted of casting everything to a string :8ball: 

Thinking about it some more and talking it over with @aanand, I think you're right and this is a good suggestion, thanks. Importantly with your approach we can always change our minds later, if we go with cast all to strings, we can't as easily change track.

aanand:
LGTM

PR #1858, opened 2015-08-12
Fix regression in `up`

### LINKED ISSUE #1856: Since 1.4, `docker-compose up container` logs all containers, not just those passed as argument
`docker-compose up lb wsgi` now also logs other containers, not just those explicitly passed in argument ("lb" and "wsgi" in this example).
So instead of having just the logs I'm interested in like before, I'm now flooded with logs of verbose containers like varnish.

aanand:
When running `docker-compose up A B`, logs are shown for those services *and all their dependencies*, which is not what the user expects. This regressed in #1645.

Not wild about the variable names here - would appreciate a second pair of eyes.

Closes #1856.

### DIFF
--- a/compose/project.py
+++ b/compose/project.py
@@ -245,26 +245,30 @@ class Project(object):
         if force_recreate and not allow_recreate:
             raise ValueError("force_recreate and allow_recreate are in conflict")
 
-        services = self.get_services(service_names, include_deps=start_deps)
+        services = self.get_services(service_names, include_deps=False)
+        all_services = self.get_services(service_names, include_deps=start_deps)
 
-        for service in services:
+        for service in all_services:
             service.remove_duplicate_containers()
 
         plans = self._get_convergence_plans(
-            services,
+            all_services,
             allow_recreate=allow_recreate,
             force_recreate=force_recreate,
         )
 
-        return [
-            container
-            for service in services
-            for container in service.execute_convergence_plan(
+        containers_to_return = []
+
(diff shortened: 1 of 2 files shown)


compose/project.py
```diff
@@ -245,26 +245,30 @@ def up(self,
...
+            containers = service.execute_convergence_plan(
                 plans[service.name],
                 do_build=do_build,
                 timeout=timeout
             )
-        ]
+            if service in services:
+                containers_to_return += containers
+
+        return containers_to_return
```
dnephin: I'm not sure about this. I think `project.up()` should still return all the containers it created.

I would think the filtering should happen in the `cli.main.up()` (or it can call a function to do the filtering) since that's the layer that deals with user expectations. 

I think it's conceivable that something else might want to know all the containers created by up (I say this because I've actually written code that has that expectation).

aanand: Fair enough - only problem with doing it in `cli.main.up` is that it can't be tested.

aanand: (easily)

dnephin: Ya, that's true. At some point I think it might make sense to split up some of the larger cli functions into some smaller functions (to make it easier to test), but that doesn't need to happen now.

In this case I think everything that happens after  `project.up()` could be in a separate function:

```python

if detached:
    return

print_logs_for_up(containers, service_names, monochrome, time)
```

dnephin:
Closed by #1931

Issue #1826, opened 2015-08-07
Stack trace when trying to extend a nonexistent service


aanand:
```yaml
web:
  image: busybox
  extends:
    service: foo
```

```
$ docker-compose up -d
Traceback (most recent call last):
  File "/Users/aanand/.virtualenvs/docker-compose/bin/docker-compose", line 9, in <module>
    load_entry_point('docker-compose==1.4.0.dev0', 'console_scripts', 'docker-compose')()
  File "/Users/aanand/work/docker/compose/compose/cli/main.py", line 39, in main
    command.sys_dispatch()
  File "/Users/aanand/work/docker/compose/compose/cli/docopt_command.py", line 21, in sys_dispatch
    self.dispatch(sys.argv[1:], None)
  File "/Users/aanand/work/docker/compose/compose/cli/command.py", line 27, in dispatch
    super(Command, self).dispatch(*args, **kwargs)
  File "/Users/aanand/work/docker/compose/compose/cli/docopt_command.py", line 24, in dispatch
    self.perform_command(*self.parse(argv, global_options))
  File "/Users/aanand/work/docker/compose/compose/cli/command.py", line 57, in perform_command
    verbose=options.get('--verbose'))
  File "/Users/aanand/work/docker/compose/compose/cli/command.py", line 78, in get_project
    config.load(config_details),
  File "/Users/aanand/work/docker/compose/compose/config/config.py", line 132, in load
    service_dict = loader.make_service_dict(service_name, service_dict)
  File "/Users/aanand/work/docker/compose/compose/config/config.py", line 156, in make_service_dict
    service_dict = self.resolve_extends(service_dict)
  File "/Users/aanand/work/docker/compose/compose/config/config.py", line 183, in resolve_extends
    other_service_dict = other_config[extends_options['service']]
KeyError: 'foo'
```

PR #1859, opened 2015-08-12
Raise configuration error when trying to extend service that does not exist

mrfuxi:
Fix for issue: #1826 


aanand:
LGTM

Issue #1851, opened 2015-08-12
container_name directive doesn't affect entries in docker-compose logs


dunk:
When using the container_name directive in the docker-compose.yml file, entries in docker-compose logs still show up with the container_name_1 naming scheme. It's non-critical, but a little inconsistent.

aanand:
Agreed - we should fix this.

PR #1862, opened 2015-08-13
Use custom container name in logs

mrfuxi:
Use custom container name in logs

### DIFF
--- a/compose/container.py
+++ b/compose/container.py
@@ -6,6 +6,7 @@ from functools import reduce
 import six
 
 from .const import LABEL_CONTAINER_NUMBER
+from .const import LABEL_PROJECT
 from .const import LABEL_SERVICE
 
 
@@ -70,7 +71,12 @@ class Container(object):
 
     @property
     def name_without_project(self):
-        return '{0}_{1}'.format(self.service, self.number)
+        project = self.labels.get(LABEL_PROJECT)
+
+        if self.name.startswith('{0}_{1}'.format(project, self.service)):
+            return '{0}_{1}'.format(self.service, self.number)
+        else:
+            return self.name
 
     @property
     def number(self):
--- a/tests/unit/container_test.py
+++ b/tests/unit/container_test.py
@@ -83,8 +83,14 @@ class ContainerTest(unittest.TestCase):
         self.assertEqual(container.name, "composetest_db_1")
 
     def test_name_without_project(self):
+        self.container_dict['Name'] = "/composetest_web_7"
         container = Container(None, self.container_dict, has_been_inspected=True)
         self.assertEqual(container.name_without_project, "web_7")
+
+    def test_name_without_project_custom_container_name(self):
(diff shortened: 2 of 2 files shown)


mrfuxi:
Feedback regarding proposed solution would be appreciated

aanand:
Thanks! I think this approach is a bit weird. I suppose what makes things complicated is that the container doesn't know whether it has a custom name or not - only the Service knows that, and there's currently no direct link between them.

I think the right place for this logic to live is LogPrinter. I'm going to take a crack at it today.

mrfuxi:
Yeah, container does not know that it has a custom name (possibly it should).
`name_without_project` is also used in `compose/service.py`

While this particular solution is the the best, I think that having the logic in container would be more transparent

aanand:
@dnephin @mnowster - what do you think of this as a solution for #1851? It's a bit weird, but it seems to be the lowest-impact fix.

mnowster:
This doesn't feel right to me. I need a bit more time to think it over.

dnephin:
Originally I had the same concern (as you both expressed), but then I looked around a bit more, and I think this is actually a pretty reasonable way to do it.

I'm going to suggest a couple small change.

compose/container.py
```diff
@@ -64,7 +68,13 @@ def name(self):
 
     @property
     def name_without_project(self):
-        return '{0}_{1}'.format(self.labels.get(LABEL_SERVICE), self.number)
+        project = self.labels.get(LABEL_PROJECT)
+        service = self.labels.get(LABEL_SERVICE)
```
dnephin: We just merged a change which makes `self.service` a property. When you rebase I'd suggest using that directly.

compose/container.py
```diff
@@ -64,7 +68,13 @@ def name(self):
...
+
+        if '{0}_{1}'.format(project, service) in self.name:
+            return '{0}_{1}'.format(service, self.number)
```
dnephin: I'd like to make this a big more strict and say something like this:

```python
if self.name.startswith('{0}_{1}'.format(project, service)):
    return '{0}_{1}'.format(service, self.number)
else:
    return self.name
```

mrfuxi:
Does this mean I should finish the work?

dnephin:
@mrfuxi I'd say give @mnowster another day to look it over, and if everyone is on-board, then yes.

thaJeztah:
Just adding in case it's useful; a WIP to support custom log tags in Docker upstream; https://github.com/docker/docker/pull/15384

mnowster:
I spiked an idea but it didn't pan out and isn't worth it. Frustrating but hey, software is compromise. 

I agree with @dnephin's suggestion to make it more strict else we're making assumptions about what a user would do for custom naming, which we ideally shouldn't.

So if you're ok to make that change @mrfuxi then yeh I'm on board.

dnephin:
LGTM

PR #1870, opened 2015-08-14
Fix volume path warning

### LINKED ISSUE #1867: Environment Variables in volumes trigger named volume warning
Warning added in #1833 is triggered by volumes starting with environment variables (e.g. `${VOLUME_DIR}:/var/stuff`).

How would we know whether the env-var was a name or a path? 

Is there a way to wait to test for `.~/` until after env-vars are evaluated?

aanand:
Though this change has no effect on master, it will fix the problem described in #1867 when cherry-picked on top of `1.4.0`, because variable interpolation for volume paths is done differently there.

Closes #1867.


mnowster:
LGTM

Issue #1874, opened 2015-08-15
Combined command to kill and remove containers


screwgoth:
Most of the times, during development of a docker-compose.yml file, I found myself killing services and then removing the stopped containers immediately. To do this, I had to execute 2 docker-compose commands , first to kill the services and then to remove the containers and associated volumes. 
I think It would be very convenient if one command did the complete clean-up , viz., killing AND removing the stopped containers. 
So, I think it would be a great convenience option to have --rm in the kill command.

aanand:
Agree on the problem, but I'm not sure about the CLI design. The Docker client uses `rm -f` to mean "kill and remove"; ideally, `docker-compose rm -f` would also do that, rather than inventing another flag for `kill` that doesn't exist on its Docker client counterpart.

This is complicated, though, because Compose and the Docker client define `rm` and `-f` differently.

-   `docker rm CONTAINER`
    - removes CONTAINER if it's not running
    - if it is running, raises an error
-   `docker rm -f CONTAINER`
    - as above, but kills CONTAINER first (so it always gets removed)
- `docker-compose rm [SERVICE ...]`
    - removes all containers (optionally filtered by service) which are not running
    - prompts for confirmation before doing so
- `docker-compose rm -f [SERVICE ...]`
    - as above, but doesn't prompt for confirmation

If we change `docker-compose rm` to be more in line with `docker rm`, it'll break people's scripts in alarming ways that'll delete more containers than they were expecting. We'd have to go through a careful deprecation period.

albers:
I fully agree with @screwgoth that there should be a way to `docker-compose rm` running containers as `docker rm -f` does.
In addition, I'd like to get rid of `docker-compose rm` asking for confirmation at all. This is not how Unix commands should behave. They should either succeed or fail.

aanand:
@albers On my Mac, `rm` asks for confirmation when a file is read-only. I have to pass `-f` to stop it doing so.

Whether or not we drop confirmation entirely, we should still go through a period of deprecation before changing the behaviour. I propose a "hard deprecation" strategy as follows:

1. Add a `--kill` flag to `rm`, which means "remove all containers, killing them first if necessary".
2. Add a `--no-confirm` flag to `rm`, which means "don't ask for confirmation".
3. Remove the `--force` flag. If a user passes it in, exit immediately with an error explaining that the meaning of `--force` is going to change, and they should use `--no-confirm` instead to get the old behaviour.
4. Release a stable version of Compose.
[...]

PR #1875, opened 2015-08-15
Add rm --kill option to kill containers before removing them

screwgoth:
Most of the times, during development of a docker-compose.yml file, I found myself killing services and then removing the stopped containers immediately. To do this, I had to execute 2 docker-compose commands , first to kill the services and then to remove the containers and associated volumes. 
I think It would be very convenient if one command did the complete clean-up , viz., killing AND removing the stopped containers. 
So, I think it would be a great convenience option to have --rm in the kill command.

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -152,21 +152,10 @@ class TopLevelCommand(Command):
         Options:
             -s SIGNAL         SIGNAL to send to the container.
                               Default signal is SIGKILL.
-            --rm              Remove containers nd associated volumes forcefully
-                              after stopping them
         """
         signal = options.get('-s', 'SIGKILL')
 
         project.kill(service_names=options['SERVICE'], signal=signal)
-        if options['--rm']:
-            all_containers = project.containers(service_names=options['SERVICE'], stopped=True)
-            stopped_containers = [c for c in all_containers if not c.is_running]
-
-            if len(stopped_containers) > 0:
-                print("Going to remove", list_containers(stopped_containers))
-                project.remove_stopped(service_names=options['SERVICE'], v=True)
-            else:
-                print("No stopped containers")
 
     def logs(self, project, options):
         """
@@ -264,9 +253,14 @@ class TopLevelCommand(Command):
         Usage: rm [options] [SERVICE...]
 
         Options:
+            --kill        Kill all the running Services
             -f, --force   Don't ask to confirm removal
             -v            Remove volumes associated with containers
         """
+        if options['--kill']:
+            print("Going to kill all the Services")
+            project.kill(service_names=options['SERVICE'], signal='SIGKILL')
+
         all_containers = project.containers(service_names=options['SERVICE'], stopped=True)
         stopped_containers = [c for c in all_containers if not c.is_running]
 
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -257,16 +257,22 @@ class TopLevelCommand(Command):
             -f, --force   Don't ask to confirm removal
             -v            Remove volumes associated with containers
         """
+        killed=False
         if options['--kill']:
-            print("Going to kill all the Services")
-            project.kill(service_names=options['SERVICE'], signal='SIGKILL')
+            print("Going to kill all containers")
+            if options.get('--force') \
+                    or yesno("Are you sure? [yN] ", default=False):
+                        project.kill(service_names=options['SERVICE'], signal='SIGKILL')
+                        killed=True
+            else:
+                print("No containers killed")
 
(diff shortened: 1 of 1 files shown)


aanand:
Thanks, but I'm -1 on this - see https://github.com/docker/compose/issues/1874#issuecomment-131780195.

screwgoth:
Can someone help me on why the test failed on an issue which seems like has nothing to do with my code. Maybe I'm wrong, but I'm not able to understand the test for me to make an update:
https://jenkins.dockerproject.org/job/Compose-PRs/1073/console

aanand:
Thanks for updating.

1. You're killing the containers first, then asking for confirmation to remove them. We should ask for confirmation first, and make sure we list *all* the containers if you passed `--kill`.

2. It'd be better to perform the kill and remove for each container in parallel, rather than kill them all and *then* remove them all. Perhaps `Project` should have a new method (this is off the top of my head, not guaranteed to work):

    ```python
    def kill_and_remove(self, service_names=None)
        parallel_execute(
            objects=self.containers(service_names, stopped=True),
            obj_callable=lambda c: c.remove(force=True),
            msg_index=lambda c: c.name,
            msg="Removing"
        )
    ```

screwgoth:
Hi @aanand, 
But that's not what `docker-compose kill` does. Without parameters, it just kills all the containers without asking permission. What I'm trying to say is, no confirmation is required in `docker` for killing containers.

One thing that we could do, is when `docker-compose rm --kill` is passed, ask whether all the containers should be killed and removed and then go ahead and kill & remove them, behaviour similar ro `docker-compose rm`. Does that make sense ?

aanand:
> One thing that we could do, is when `docker-compose rm --kill` is passed, ask whether all the containers should be killed and removed and then go ahead and kill & remove them, behaviour similar ro `docker-compose rm`. Does that make sense ?

That's exactly what I was proposing. Sorry if I was unclear.

screwgoth:
Updated as per discussion. 
Following are some prominent test cases:
1. If all container services are up and running `docker-compose rm --kill` will ask if all the services are to be killed. If answered "y", it will kill all the containers and remove the stopped containers. If "N", then no containers are killed and since none are stopped, none are removed.
2. However, if some containers are already in stopped state when `docker-compose rm --kill` is called, and the user answers "N", no containers will be killed, but for all the previously  stopped containers , normal rm behaviour follows, i.e. the user will be asked confirmation to remove the containers.
3. `docker-compose rm --kill -f` will promptly kill all the containers and remove all the resultant stopped containers.

aanand:
I think it's weird to confirm 2 separate actions. I'd rather there were a single confirmation. This would work better:

```bash
$ docker-compose ps
    Name           Command       State     Ports
------------------------------------------------
counter_web_1   python app.py   Up
counter_web_2   python app.py   Up
counter_web_3   python app.py   Exit 137

# default behaviour stays the same

$ docker-compose rm
Going to remove counter_web_3
Are you sure? [yN] n

# confirm kill+remove for all containers

$ docker-compose rm --kill
Going to kill and remove counter_web_1, counter_web_2, counter_web_3
Are you sure? [yN] n

# remove the stopped container

$ docker-compose rm  --force
counter_web_3

# kill+remove the running containers 

$ docker-compose rm --kill --force
counter_web_1
counter_web_2
```

Note how we say "Going to kill and remove counter_web_1, counter_web_2, counter_web_3". This is technically false - counter_web_3 is already stopped and isn't going to be killed again - but I don't think that inaccuracy matters, and the flow is dramatically simpler.

screwgoth:
But this is exactly what my patch does. Although, admittedly, it does not mention "kill and remove" just "Going to kill all containers". I can update it with "Going to kill and remove all the containers" , just so that we don't have to deal with the inaccuracy you mentioned.

Other than that, this is behaving exactly as you have demoed above.

aanand:
OK, I misread the code and I see that you're using the `killed` variable to make sure we don't ask twice for confirmation. Sorry about that.

On to point 2 of https://github.com/docker/compose/pull/1875#issuecomment-132532098, then: it'd be both faster and cleaner (from a code perspective) if the kill+remove step was a single, parallelised job. For one thing, that `killed` variable could be thrown away.

Could you try and implement what I suggested in the code snippet?

aanand:
Bump - @screwgoth, any interest in updating this?

aanand:
As mentioned in https://github.com/docker/compose/issues/1874#issuecomment-140283789, it'd be nice to have a `-k` alias for `--kill` too.

dnephin:
We now have `docker-compose down` (#2646) which allows you to do these operations together, so going to close this PR.

Issue #1863, opened 2015-08-13
docker-compose ignores the global settings of log-driver


mahnunchik:
/etc/default/docker 
```
DOCKER_OPTS='
-H tcp://0.0.0.0:2376
-H unix:///var/run/docker.sock
--storage-driver aufs
--tlsverify
--tlscacert /etc/docker/ca.pem
--tlscert /etc/docker/server.pem
--tlskey /etc/docker/server-key.pem
--label provider=generic
--log-driver=syslog
'
```

docker-compose.yml
```yml
web:
  build: ./
  ports:
    - "80:80"
  links:
    - db:DB
db:
  image: mariadb:latest
```

Following command starts application with default `json-file` log-driver:
```
$ docker-compose up -d
```
Despite the global specified `syslog` log-driver.

```bash
$ docker -v
Docker version 1.8.1, build d12ea79
$ docker-compose -v
docker-compose version: 1.4.0

```

aanand:
[This code](https://github.com/docker/compose/blob/ed31673069b417c1f9761812bfb7ea2b82c09ae1/compose/service.py#L648-L651) is probably the culprit.

We should probably just set `log_config` to `None` if `options.get('log_driver')` and `options.get('log_opt')` are both `None`.

mnowster:
I've done some investigation and we can't set `log_config` to `None`, that will get interpreted as meaning disable logging.

Further investigation with @aanand looking at the docker code also has brought to light that you _can_ have `log_opt` without setting a `log_driver`.

This is going to need a PR doing on docker-py to remove the extraneous validation of `log_driver` types and let the docker-daemon handle if you set a `log_driver` to None but also send `log_opt`.

The other benefit of removing that code from docker-py is that it'll reduce the bottleneck of getting support in for new/more log drivers, which is a Good Thing.

Then we can change the default behaviour of `docker-compose` to not set `log_driver` to `json-file` and instead we should only send that option if it is set.

thaJeztah:
Might be of interest that it looks like there's currently a regression in docker, causing `--log-opt` settings for the daemon to be ignored; see this PR for more information: https://github.com/docker/docker/pull/15611

mnowster:
@thaJeztah ah, I wasn't aware of that, thanks for the heads up.

diist:
I have this issue with docker 1.8.2 and docker-compose 1.4.0 

[...]

PR #1878, opened 2015-08-17
Change default `log_driver` value

mnowster:
If `log_driver` is not set in our .yml don't set a default of 'json-file', as this overrides daemon level config options as per issue: https://github.com/docker/compose/issues/1863

Send an empty string and let docker figure out it's default/read from it's global settings.

Fixes https://github.com/docker/compose/issues/1863

### DIFF
--- a/compose/service.py
+++ b/compose/service.py
@@ -657,7 +657,7 @@ class Service(object):
         cap_add = options.get('cap_add', None)
         cap_drop = options.get('cap_drop', None)
         log_config = LogConfig(
-            type=options.get('log_driver', 'json-file'),
+            type=options.get('log_driver', ""),
             config=options.get('log_opt', None)
         )
         pid = options.get('pid', None)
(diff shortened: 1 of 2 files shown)


mnowster:
Note: the tests will fail until the docker-py branch is merged/released.

aanand:
LGTM

compose/service.py
```diff
@@ -657,7 +657,7 @@ def _get_container_host_config(self, override_options, one_off=False):
         cap_add = options.get('cap_add', None)
         cap_drop = options.get('cap_drop', None)
         log_config = LogConfig(
-            type=options.get('log_driver', 'json-file'),
+            type=options.get('log_driver', ""),
             config=options.get('log_opt', None)
         )
```
dnephin: Just noticed there was no test added to verify the bug in #1863 was fixed by this change

mnowster: That's not what happened. I added one, then removed it and added the test at the docker-py level and at the time, it showed up a docker bug.

https://github.com/docker/docker-py/pull/718/files#diff-347ad10bf29330a41397ea485670f027R285

https://github.com/docker/docker/issues/15633

Which has now been fixed, so we can remove the skip test from docker-py.

dnephin: Ah, cool

PR #1879, opened 2015-08-17
Handle all exceptions from API

mnowster:
If we get back an error that wasn't an APIError, it was causing the
thread to hang. This will catch the Exception and re-raise it, so we 
keep the stacktrace.


mnowster:
ping @aanand for review, I pulled the exception handling out of the other PR as requested :wind_chime:

aanand:
LGTM :+1: :+1: :+1:

PR #1886, opened 2015-08-18
Stop using deprecated `create_host_config` from utils

mnowster:
To ensure backwards compatibility that `network_mode` gets set to  `default` when we have an api version high enough we need to use `docker.client.create_host_config` and not the deprecated `docker.utils.create_host_config`.

### DIFF
--- a/compose/service.py
+++ b/compose/service.py
@@ -11,7 +11,6 @@ from operator import attrgetter
 import enum
 import six
 from docker.errors import APIError
-from docker.utils import create_host_config
 from docker.utils import LogConfig
 from docker.utils.ports import build_port_bindings
 from docker.utils.ports import split_port
@@ -678,7 +677,7 @@ class Service(object):
 
         devices = options.get('devices', None)
 
-        return create_host_config(
+        return self.client.create_host_config(
             links=self._get_links(link_to_self=one_off),
             port_bindings=port_bindings,
             binds=options.get('binds'),
--- a/tests/integration/service_test.py
+++ b/tests/integration/service_test.py
@@ -813,6 +813,13 @@ class ServiceTest(DockerClientTestCase):
         for k, v in {'FILE_DEF': 'F1', 'FILE_DEF_EMPTY': '', 'ENV_DEF': 'E3', 'NO_DEF': ''}.items():
             self.assertEqual(env[k], v)
 
+    def test_with_high_enough_api_version_we_get_default_network_mode(self):
+        # TODO: remove this test once minimum docker version is 1.8.x
+        with mock.patch.object(self.client, '_version', '1.20'):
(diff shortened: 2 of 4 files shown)


compose/service.py
```diff
@@ -667,7 +667,10 @@ def _get_container_host_config(self, override_options, one_off=False):
 
         devices = options.get('devices', None)
 
+        version = os.environ.get('COMPOSE_API_VERSION', '1.19')
+
         return create_host_config(
+            version=version,
```
dnephin: It's too bad `docker.Client()._version` is "private". It would have been nice to just use `self.client._version` instead of reading it from the environment again (which maybe is still an option)?

If we have to read it again, we should probably move the `COMPOSE_API_VERSION` string to `const.py`, and create a `DEFAULT_API_VERSION` in `const.py` and use it instead of hardcoding `1.19` in both places.

mnowster: Yep, I agree with you. I did think of doing that refactoring while I was in there and the reason I didn't do these was I'm already now 3 levels away from the original bug I was working on! So decided I had to draw a line somewhere and decided to stop.

mnowster:
Ok, this does work but I don't know why the requirements.txt isn't picking up the point to the docker-py commit which includes the `Client.create_host_config` :crystal_ball:

tests/integration/service_test.py
```diff
@@ -813,6 +813,12 @@ def test_resolve_env(self):
         for k, v in {'FILE_DEF': 'F1', 'FILE_DEF_EMPTY': '', 'ENV_DEF': 'E3', 'NO_DEF': ''}.items():
             self.assertEqual(env[k], v)
 
+    def test_with_high_enough_api_version_we_get_default_network_mode(self):
+        with mock.patch.object(self.client, '_version', '1.20'):
+            service = self.create_service('web')
+            service_config = service._get_container_host_config({})
+            self.assertEquals(service_config['NetworkMode'], 'default')
```
dnephin: This test is passing in CI, but my understanding is that it should fail against docker 1.7.1 since the client api version would be ahead of the server.  Any idea why it's not failing?

mnowster: Hmm, that's weird, because that was the failure I was getting with docker 1.8 until @aanand helped me change the way it was written and patch the version number on the client object.

dnephin: Oh, I think it's because this doesn't actually hit the API, it's only hitting the client, which checks for this version.

I think we'll be able to remove this test once we upgrade the minimum version to docker 1.8.x.  Mind adding a TODO here so we remember we can remove it?

mnowster: Ah, right. Yep, done.

dnephin:
LGTM

Issue #1816, opened 2015-08-06
Failed to initialize logging driver: Unix syslog delivery error


flower-pot:
I wanted to try out the log-driver and log-opt added in #1388 and #1665.

My `docker-compose.yml` looks like this:

```
web:
  image: flowerpot/goif
  ports:
    - "8080:8080"
  log_driver: "syslog"
  log_opt:
    address: "udp://syslog-server:514"
```

However, when I do `docker-compose up` I get the following error:

```Failed to initialize logging driver: Unix syslog delivery error```

Which is weird because this works:

```docker run --rm -p 8080:8080 --log-driver=syslog --log-opt syslog-address=udp://syslog-server:514 flowerpot/goif```

Shouldn't that be equivalent?

Versions:

```
docker-compose version: 1.4.0rc3
```

```
Client version: 1.7.1
Client API version: 1.19
Go version (client): go1.4.2
Git commit (client): 786b29d
OS/Arch (client): darwin/amd64
Server version: 1.7.1
Server API version: 1.19
Go version (server): go1.4.2
Git commit (server): 786b29d
OS/Arch (server): linux/amd64
```

aanand:
Under `log_opt`, shouldn't the `address` key be `syslog-address`?

flower-pot:
You are absolutely right, sorry about that...

PR #1897, opened 2015-08-20
log_opt: change address to syslog-address

qzio:
The documentation for `log_opt` stated to use `address` but that gives an error.

If I use `syslog-address` everything works as expected.

Also, see #1816 which is closed...

Not sure if I am missing something obviously, but this error cost me a couple of hours ;)

### DIFF
--- a/compose/config/schema.json
+++ b/compose/config/schema.json
@@ -56,16 +56,9 @@
         "image": {"type": "string"},
         "labels": {"$ref": "#/definitions/list_or_dict"},
         "links": {"type": "array", "items": {"type": "string"}, "uniqueItems": true},
+
         "log_driver": {"type": "string"},
-
-        "log_opt": {
-          "type": "object",
-
-          "properties": {
-            "syslog-address": {"type": "string"}
-          },
-          "required": ["syslog-address"]
-        },
+        "log_opt": {"type": "object"},
 
         "mac_address": {"type": "string"},
         "mem_limit": {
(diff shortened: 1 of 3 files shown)


olleolleolle:
The docs that confused were here: https://docs.docker.com/compose/yml/#log-driver

compose/config/schema.json
```diff
@@ -62,9 +62,9 @@
           "type": "object",
 
           "properties": {
-            "address": {"type": "string"}
+            "syslog-address": {"type": "string"}
           },
-          "required": ["address"]
+          "required": ["syslog-address"]
```
aanand: Unless I've misunderstood, we shouldn't be validating properties in here at all - we don't know what options the driver will accept.

aanand: (ping @mnowster)

mnowster: @aanand you are correct, this is a mistake. We should remove:

```
          "properties": {
            "address": {"type": "string"}
          },
          "required": ["address"]
```

mnowster:
@qzio sorry about that, my mistake as I was wrapping my head around how the `log-opt` should work.

qzio:
@mnowster It's ok, I'm just glad compose even support log_driver and log_opt :)

Would be nice to get a fix in for other people though.

mnowster:
LGTM

mnowster:
Thanks for spotting it and submitting a fix @qzio , much appreciated :)

PR #1903, opened 2015-08-21
Use official python docker images instead of compiling python manually

dopry:
I'm one of those developers that gets impatient waiting for feedback from a build server.  It seems like every time I've submitted or update a PR on compose I need to wait for docker build, and cache doesn't seem to get leveraged. On 6/1/2015 @aanand introduced manual compilation of python 2.7.9 to suppress some unsecure TLS warnings. The compile take several minutes. This change updates the docker file to use the official python:2.7.10-wheezy container to eliminate the compile time.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,13 +1,8 @@
-FROM debian:wheezy
+FROM python:2.7.10-wheezy
 
 RUN set -ex; \
     apt-get update -qq; \
     apt-get install -y \
-        gcc \
-        make \
-        zlib1g \
-        zlib1g-dev \
-        libssl-dev \
         git \
         apt-transport-https \
         ca-certificates \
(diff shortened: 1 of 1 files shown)


dnephin:
I agree we need to make the build faster.

This may be one option. When we go to merge in python3 support and run tests against it, we'll have to `apt-get install python3`, but that's probably fine.

I believe the reason the build cache isn't being hit is because all the builds get run using docker-in-docker. So the build cache isn't actually on the host, it's in a container that gets removed after the build ends. I'd like to fix it so that we're actually able to take advantage of the build cache.

dopry:
I don't think it's an either or situation. I think now you can shave off 20-30% of the build time with this PR. Then later when time and testing allows improve the build pipeline as a whole.

aanand:
@dnephin The build step isn't done in docker-in-docker - it's done from outside our scripts as part of the Jenkins job.

I've seen the cache both hit and miss in the past; my guess would be that something is aggressively clearing or pruning it because of all the builds being run. (@jfrazelle - can you shed any light there?)

It's true that we don't need to build Python 2.7.9 from source any more, since boot2docker 1.8.0 fixed the [underlying issue](https://github.com/boot2docker/boot2docker/issues/808), so this should be fine. It needs a rebase, though.

dnephin:
Ah, my mistake. I need to look into the build process some more.

It looks like the last step of the build is a `docker rmi` which would remove all the layers, right? So we'll only hit the cache if another build runs at the same time.

dopry:
Rebase pushed.

On Mon, Aug 24, 2015 at 12:00 PM, Daniel Nephin <notifications@github.com>
wrote:

> Ah, my mistake. I need to look into the build process some more.
>
> It looks like the last step of the build is a docker rmi which would
> remove all the layers, right? So we'll only hit the cache if another build
> runs at the same time.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/docker/compose/pull/1903#issuecomment-134263021>.
>



-- 
Darrel O'Pry
The Spry Group, LLC.
http://www.spry-group.com
718-355-9767 x101

dopry:
Unfortunately I can't take credit for breaking the build... that seems to be in master.

jfrazelle:
we have a cron job clean up every so often

aanand:
Build is green. LGTM

dopry:
Now you're building both python 2.x and 3.x in the Dockerfile and the builds are even slower. You know most people create a container for each version of python they need to test against or for all the versions of python they need to test against and push it to an image repository to reduce their build times in their CI pipeline.

dnephin:
Using separate images for each python version is definitely an option. I wont save us much time unless we can run the builds in parallel. I'll open an issue about it.

Thanks to @jfrazelle we're no longer deleting the layers after the build, so we actually hit the cache more often.  I observed some of my builds being completely cached.

I think there is still a cronjob that removes them after some time, and I'm not sure exactly how often that is, but I'll investigate that further as well.

PR #1905, opened 2015-08-22
Update the docs build system

moxiegirl:
- reducing the number of RUN suggested by Mbauer
- run the same script run on the docs.docker.com publish step
- fixing the link and image structure so they work in GitHub
- new sed scripts on docs.docker.com allow for good sedding on GHFM


moxiegirl:
ping @aanand @bfirsh

docs/reference/index.md
```diff
@@ -11,19 +11,20 @@ parent = "smn_compose_ref"
...
 
-The following pages describe the usage information for the [docker-compose](/reference/docker-compose.md) subcommands. You can also see this information by running `docker-compose [SUBCOMMAND] --help` from the command line.
+The following pages describe the usage information for the [docker-compose](reference/docker-compose.md) subcommands. You can also see this information by running `docker-compose [SUBCOMMAND] --help` from the command line.
+
+* [build](../reference/build.md)
+* [help](../reference/help.md)
+* [kill](../reference/kill.md)          
+* [ps](../reference/ps.md)
+* [restart](../reference/restart.md)
+* [run](../referencerun.md)
```
thaJeztah: Looks like you inadvertently removed slashes after `reference` here (and below)

docs/Dockerfile
```diff
@@ -1,26 +1,26 @@
...
-# 4 Change ](word.md) to ](/project/word)
-# 5 Remove .md extension from link text
-# 6 Change ](../ to ](/project/word) 
-# 7 Change ](../../ to ](/project/ --> not implemented
-# 
-# 
-RUN /src/pre-process.sh /docs
+COPY . /docs/content/compose
+ 
+
```
thaJeztah: Just a nit; would be nice to have a comment describe what the touch-up script does

docs/Dockerfile
```diff
@@ -1,26 +1,26 @@
...
+RUN wget https://raw.githubusercontent.com/docker/docs.docker.com/master/touch-up.sh 
+RUN chmod 777 touch-up.sh && mv touch-up.sh /src && cat /dev/null > /docs/build.json
+
+# To get the git info for this repo
```
thaJeztah: Stray comment?

docs/Dockerfile
```diff
@@ -1,26 +1,26 @@
...
+
+RUN /src/touch-up.sh /docs
```
thaJeztah: Thinking we should combine the lines, and use curl here, as we do in other locations, also we don't need the intermediate download / move. `cat /dev/null` .. would a "touch" do here? Or is the build.json file already present?

Something like this (writing by head, not tested yet);
```Dockerfile
RUN curl -sSL https://raw.githubusercontent.com/docker/docs.docker.com/master/touch-up.sh > /src/touch-up.sh \
 && chmod +x /src/touch-up.sh \
 && touch /docs/build.json \
 && /src/touch-up.sh /docs
```

thaJeztah:
> fixing the link and image structure so they work in GitHub

💝 that is awesome! Very happy to see that! Thanks @moxiegirl !

docs/Dockerfile
thaJeztah: Hm, thinking if we could move some steps to `ONBUILD` instructions in the docs-base image, then we don't have to repeat them in each repository.

thaJeztah: lol, just noticed this is coming from the docs.docker.com repo; I wasn't aware this was opened up now (think it was private before?)

moxiegirl: @thaJeztah  docs.docker.com is open.   I don't want to move this into `docs-base` for a couple of reasons. It would hide the processing step and I want users to be able to find it and I'm trying to keep `docs-base` for theme only.  `docs.docker.com` is home to our build scripts.

docs/Dockerfile
moxiegirl: Agreed. I had it and I removed it during a test.

docs/reference/index.md
moxiegirl: Thanks...I was thinking last night Friday late is a bad time to do this....I need a reviewer.  Viola!

docs/Dockerfile
```diff
@@ -1,26 +1,28 @@
...
-# 7 Change ](../../ to ](/project/ --> not implemented
-# 
-# 
-RUN /src/pre-process.sh /docs
+# Copy local Compose changes to the build
+#
+COPY . /src
+COPY . /docs/content/compose
+ 
+# Run the build ssed over the markdown to conver to Hugo crosss-references
```
thaJeztah: s/conver/convert/
s/crosss/cross/

🙊

docs/Dockerfile
```diff
@@ -1,26 +1,28 @@
...
+#
+RUN curl -sSL https://raw.githubusercontent.com/docker/docs.docker.com/master/touch-up.sh > /src/touch-up.sh \
+ && chmod +x /src/touch-up.sh \
+ && touch /docs/build.json \
+ && /src/touch-up.sh /docs
```
thaJeztah: Github complains about missing new line (line break) at the end of the file

In this case, you might want to add it; I don't know if Dockerfile parsing breaks if the last new line is missing. Would be a good test case btw :-)

thaJeztah:
Thanks, Mary. Caught two more nits. My computer is running updates right now, so didn't test it :sweat_smile:

docs/Dockerfile
moxiegirl: @thaJeztah What would be the test, running the build? Builds fine. :grinning:

thaJeztah: Alright then \o/

moxiegirl:
I'm going to close this for a bit. I will reopen.  I want to do a bit more research.

PR #1914, opened 2015-08-24
Export COMPOSE_FILE

bbirand:
The environment variable is not used by `docker-compose` without the `export` line..


aanand:
Thanks. LGTM

PR #1920, opened 2015-08-25
Use py.test as a test runner

dnephin:
I split out the "build" dependencies from the "test" dependencies to cleanup tox.ini a bit. pyinstaller doesn't install under python3.

I verified that the coverage report is still in the right place.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -91,7 +91,7 @@ RUN pip install -r requirements-dev.txt
 RUN pip install tox==2.1.1
 
 ADD . /code/
-RUN python setup.py install
+RUN pip install --no-deps -e /code
 
 RUN chown -R user /code/
 
--- a/requirements-build.txt
+++ b/requirements-build.txt
@@ -1,0 +1,1 @@
+git+https://github.com/pyinstaller/pyinstaller.git@12e40471c77f588ea5be352f7219c873ddaae056#egg=pyinstaller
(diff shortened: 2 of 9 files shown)


Dockerfile
```diff
@@ -91,7 +91,7 @@ RUN pip install -r requirements-dev.txt
 RUN pip install tox==2.1.1
 
 ADD . /code/
-RUN python setup.py install
+RUN pip install --no-deps -e /code
```
dnephin: This change makes it so that dependencies from `setup.py` aren't reinstalled, and so that files aren't copied. It just creates a link to the source files in `/code`, so this build step is faster.

aanand: Oh nice - will this also prevent the problem we've had before where, if `setup.py` disagrees somehow with `requirements.txt`, then this step will silently reinstall different versions of dependencies?

dnephin: Yup, exactly. Previously this was installing new versions of dependencies .Now it doesn't do any re-install of dependencies, so only what's listed in `requirements.txt` gets installed (but if we somehow miss a dependency in requirements.txt it fails, which I think is good).

dnephin:
Hit another test error, created https://github.com/docker/docker/issues/15995

dnephin:
A green build!

This is ready for more eyes

aanand:
LGTM. Is it ready to be merged, then?

dnephin:
I think so, I'm happy with it

PR #1924, opened 2015-08-25
Fix `docker-compose logs` on python3

dnephin:
I think this is the last incompatibility with python3. Thanks to @funkyfuture  and @ouziel-slama for calling out this issue a while ago and opening PRs against my branch.

I've updated the unit tests for this function to properly reflect the types we're getting back from `docker-py`. In both py2 and py3 it returns bytes objects, so now our unittests do the same, and we convert everything to unicode in `split_buffer()`.

Includes a couple other cleanup commits for moving a test file to match the naming convention, remove two functions from `utils.py` that are not used anywhere.


aanand:
LGTM

Issue #1909, opened 2015-08-24
Interpolate and validate services included by extends


mnowster:
When a service is extended, we're not currently interpolating and validating that it is a valid service to extend. We should be.

PR #1927, opened 2015-08-26
Validate and interpolate extended services

mnowster:
In order to run interpolation and schema validation against a service that we want to extend, I did some re-factoring, which ultimately made it easier and clearer to then be able to perform those functions.

The `resolve_extends` has always been a bit of an overloaded function, it previously was doing more than what the word resolve meant to me. It was constructing, validating, logic, validating then constructing. Rather than insert more complexity into an already complex function, I've done some refactoring which ultimately I feel helps tighten up the design a bit.

Ultimately, this now splits validation into two parts, an initial field validation, which we run when we call `Config.load` and a second validation, of service schema validation, when we call `make_service_dict`. I've also moved some functions to be within the class, as they provide no benefit living outside of it and it helps clear up the intent of the scope of that functionality. Internal workings of the ServiceLoader.

I'm not stating that this is perfect code but it does achieve our goal of the feature, some of the code has been improved while I'm there and gives us some room to think about how we might iterate on this in the future. It highlights some design approaches too, which there is never only one true answer for.

*Note*: Please note the commit messages, don't just look at the giant diff. It'll help demonstrate how I de-constructed this problem and why I've chosen to improve/move some things around.

:sparkles: :ok_woman: 

Fixes https://github.com/docker/compose/issues/1909

### DIFF
--- a/compose/config/config.py
+++ b/compose/config/config.py
@@ -142,8 +142,12 @@ def load(config_details):
     service_dicts = []
 
     for service_name, service_dict in list(processed_config.items()):
-        loader = ServiceLoader(working_dir=working_dir, filename=filename)
-        service_dict = loader.make_service_dict(service_name, service_dict)
+        loader = ServiceLoader(
+            working_dir=working_dir,
+            filename=filename,
+            service_name=service_name,
+            service_dict=service_dict)
+        service_dict = loader.make_service_dict()
         validate_paths(service_dict)
         service_dicts.append(service_dict)
 
@@ -151,7 +155,7 @@ def load(config_details):
 
 
 class ServiceLoader(object):
-    def __init__(self, working_dir, filename, already_seen=None):
+    def __init__(self, working_dir, filename, service_name, service_dict, already_seen=None):
         if working_dir is None:
             raise Exception("No working_dir passed to ServiceLoader()")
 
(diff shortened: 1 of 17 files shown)


aanand:
I've had a first pass and it looks great. Thanks for the helpful commit messages.

One question: will the following setup work?

**docker-compose.yml:**

```yaml
web:
  extends:
    file: common.yml
    service: common-config

  build: .
```

**common.yml:**

```yaml
common-config:
  environment:
    - FOO=1
```

```
$ docker-compose up
```

Note how `common-config` is, by itself, invalid - it doesn't specify `build` or `image` - but `web` in `docker-compose.yml` *is* valid, because it adds a `build` key.

I think that this setup definitely *should* work, and that we should have an explicit test for it. Apologies if we already do and I've missed it.

mnowster:
@aanand that's a good question and I believe the answer is yes it will work, I covered that case in the schema here: https://github.com/docker/compose/blob/master/compose/config/schema.json#L119

You are right, we should totally have a test to cover that scenario. I will add one :+1:

mnowster:
After further investigation, this needs a bit more work on this to resolve the situation @aanand has described. Agile development of iterating FTW :muscle:

aanand:
LGTM

dnephin:
LGTM

aanand:
validation-road-sign-6644977

PR #1930, opened 2015-08-26
Allow setting a one-off container name

dnephin:
In #1755 we removed the custom name from one-off containers, which left them with the default name.

This adds a `--name` param to `docker-compose run` to name the container

### DIFF
--- a/tests/unit/service_test.py
+++ b/tests/unit/service_test.py
@@ -150,6 +150,19 @@ class ServiceTest(unittest.TestCase):
         self.assertEqual(opts['hostname'], 'name.sub', 'hostname')
         self.assertEqual(opts['domainname'], 'domain.tld', 'domainname')
 
+    def test_get_container_create_options_with_name_option(self):
+        service = Service(
+            'foo',
+            image='foo',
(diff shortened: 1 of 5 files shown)


tests/unit/service_test.py
```diff
@@ -99,14 +99,12 @@ def test_get_volumes_from_service_no_container(self):
 
     def test_split_domainname_none(self):
         service = Service('foo', image='foo', hostname='name', client=self.mock_client)
-        self.mock_client.containers.return_value = []
```
dnephin: These mock return values haven't been necessary for a while, since we refactored `_get_container_create_options()` to not make any api calls.

aanand: Nice. Sorry to be picky, but would you mind doing this cleanup (and the `__` -> `_` changes) in a separate commit?

dnephin: Cool, I can do that

dnephin:
Split into 2 commits, cleanup then actual feature

aanand:
LGTM

PR #1931, opened 2015-08-26
Only show logs for listed services in `up`

### LINKED ISSUE #1856: Since 1.4, `docker-compose up container` logs all containers, not just those passed as argument

dnephin:
I implemented what I was suggesting in #1858 

Closes #1856

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -538,19 +538,8 @@ class TopLevelCommand(Command):
         )
 
         if not detached:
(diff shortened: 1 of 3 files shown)


aanand:
LGTM

Issue #1908, opened 2015-08-24
`dockerfile` and `image` shouldn't be allowed in combination


aanand:
A file like this should result in a validation error:

```yaml
web:
  image: myimage
  dockerfile: Dockerfile.alt
```

This should also be documented.

PR #1932, opened 2015-08-26
Do not allow to specify both image and dockerfile in configuration.

mrfuxi:
Config validation mentioned in #1908

### DIFF
--- a/tests/unit/config_test.py
+++ b/tests/unit/config_test.py
@@ -191,6 +191,17 @@ class ConfigTest(unittest.TestCase):
                 )
             )
 
+    def test_config_image_and_dockerfile_raise_validation_error(self):
+        expected_error_msg = "Service 'web' has both an image and alternate Dockerfile."
+        with self.assertRaisesRegexp(ConfigurationError, expected_error_msg):
+            config.load(
+                config.ConfigDetails(
+                    {'web': {'image': 'busybox', 'dockerfile': 'Dockerfile.alt'}},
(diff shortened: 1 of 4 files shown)


dnephin:
LGTM

Build failures are known (#1922) not related to your change.

A little more context could be included about why build/dockerfile is incompatible with image, but I  don't think that's necessary to merge this PR.

mrfuxi:
We could quote exact reason form error message: 

> Using `image` together with either `build` or `dockerfile` is not allowed. Attempting to do so results in an error. A service can either be built to image or use an existing image, not both.

tests/unit/config_test.py
```diff
@@ -191,6 +191,28 @@ def test_invalid_list_of_strings_format(self):
                 )
             )
 
+    def test_config_image_and_build_raise_validation_error(self):
```
mnowster: This is a duplicate test, we already have the image and build case covered here: https://github.com/mrfuxi/compose/blob/image-dockerfile-validation/tests/unit/config_test.py#L131

mnowster:
Once the duplicate test has been removed, everything else is LGTM. 

Thanks @mrfuxi :star:

mrfuxi:
Ok, will remove the test (have not noticed it earlier). Should I update docs or leave them as they are?

dnephin:
I think they're ok as they are

mrfuxi:
Extra test if gone

PR #1933, opened 2015-08-26
Fix flaky test

### LINKED ISSUE #1922: tests.integration.service_test test_scale_with_custom_container_name_outputs_warning

dnephin:
Branched from #1939
Resolves #1922 

This works around the problem of containers not being removed correctly by using a different name for the service.  Also adds a new label to all containers so that if we end up with test pollution we can track the container back to the test that created it.

This is definitely just a workaround, but I think the root issue is actually a race condition in the docker engine so will take much longer to fix.  I don't like having every branch fail our test suite. So I'd like to push this workaround, since the failure isn't directly related to compose itself.

### DIFF
--- a/tests/integration/state_test.py
+++ b/tests/integration/state_test.py
@@ -199,7 +199,6 @@ class ServiceStateTest(DockerClientTestCase):
 
         self.assertEqual([c.is_running for c in containers], [False, True])
 
-        web = self.create_service('web', **options)
         self.assertEqual(
             ('start', containers[0:1]),
             web.convergence_plan(),
--- a/compose/config/config.py
+++ b/compose/config/config.py
@@ -358,7 +358,7 @@ def parse_environment(environment):
         return dict(split_env(e) for e in environment)
 
     if isinstance(environment, dict):
-        return environment
+        return dict(environment)
 
     raise ConfigurationError(
         "environment \"%s\" must be a list or mapping," %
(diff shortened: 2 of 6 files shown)


dnephin:
https://jenkins.dockerproject.org/job/Compose-PRs/1179/console

dnephin:
It worked! build is green

tests/integration/state_test.py
```diff
@@ -199,7 +199,6 @@ def test_trigger_start(self):
 
         self.assertEqual([c.is_running for c in containers], [False, True])
 
-        web = self.create_service('web', **options)
```
aanand: What's the purpose of this change?

dnephin: I found it while debugging. It's a no-op since line 193 does the same thing.

aanand:
LGTM

PR #1935, opened 2015-08-27
Normalise ignore files

aanand:
- Consistent order and contents (where possible)
- Prepend `.gitignore` paths with slashes where appropriate


dnephin:
LGTM, same build failure from #1922

PR #1948, opened 2015-08-28
Support TLS without Verification

ekristen:
It is possible for the docker daemon to be run with TLS without `--tlsverify`, example is `docker -H tcp://192.168.100.100:2376 --tls ps` 

This PR allows it so that if you specify `DOCKER_TLS` as non empty env var, that the connection will get setup to not verify the connection.

Signed-off-by: Erik Kristensen <erik@erikkristensen.com>

### DIFF
--- a/compose/cli/docker_client.py
+++ b/compose/cli/docker_client.py
@@ -19,6 +19,16 @@ def docker_client():
 
     tls_config = None
 
+    if os.environ.get('DOCKER_TLS', '') != '':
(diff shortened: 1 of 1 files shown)


dnephin:
#977 was a related PR, please see some of the discussion in that PR.

dnephin:
We've moved all of this logic in to `docker-py` at `docker.utils.kwargs_from_env`.

Any changes (and further discussion) will need to be opened as a PR on that repo.

PR #1950, opened 2015-08-31
WIP: Use pytest-xdist to run tests in parallel

dnephin:
Branched from #1920

Most runs seem to have a few flakes, I'm submitting this PR to see how they run on jenkins.


aanand:
Nice.

dnephin:
Still 3 or 4 failures per run unfortunately.  

I was able to set a unique project name per test case, but I think there's going to be some work to address the flakes.  Still trying to figure out if these are problems with the test (overlapping constants) or if they're failures in docker engine.

dnephin:
The only tests that fail now are ones that do a `build`. We need a way to assign a dynamic label to the image to do proper cleanup.  This is going to require more work, so closing this PR for now.

PR #1952, opened 2015-08-31
Bundle schema.json in binary

aanand:
The built binary doesn't include `schema.json`, causing it to fail when doing anything less trivial than `docker-compose version`. Fixed with a PyInstaller hook that explicitly copies it in.

I'm not sure `script/hooks` is the best place for it - suggestions welcome.


dnephin:
A couple ideas for including the data:

* pyinstaller really should support https://pythonhosted.org/setuptools/setuptools.html#including-data-files (using `package_data={'compose.config': ['*.json'}]`), but I have no idea if it does
* http://pythonhosted.org/PyInstaller/#adding-files-to-the-bundle makes me believe that we could add something like this to `docker-compose.spec`

```python
exe = EXE(
   pyz,
   ...,
   [('SCHEMA', '/my/project/schema.json', 'DATA')]
   ....)
```

A last resort might be to use hooks, but specify the path in the `docker-compose.spec`. I think one of the kwargs is `hookspath`.

aanand:
`docker-compose.spec` isn't currently under version control - could/should it be? It's a generated file and contains the absolute path to the repository on my local machine, but it's Python code, so I suppose we can make it generic.

dnephin:
Oh, I wasn't aware of that. I guess it's created automatically by pyinstaller?  

I'm not sure if that's any better, but the docs seem to indicate that editing it and customizing it is a supported way of doing things (at least from what I read).

dnephin:
Oh ya, the first point in http://pythonhosted.org/PyInstaller/#using-spec-files "When you want to bundle data files with the app.". I think that's us

aanand:
Updated to use `docker-compose.spec`, which is now versioned.

dnephin:
LGTM (once jenkins is happy with it)

aanand:
Rebased after merging #1920 caused a conflict

dnephin:
Tests failed on #1955, which sucks because we don't get any real coverage for this until the very last step of the ci script (which gets skipped on earlier failures).

aanand:
Rebased after #1927 split out `schema.json` into two files

mnowster:
LGTM

Issue #424, opened 2014-08-19
bind mount (volumes) not allowed for files – only directorys


chymian:
hey,
trying to mount a single file (instead of a directory) throws an error:
Cannot start container bc0f924401841f2ed92c088cb8089cadad2359126b9f6a3ff15b6cb657835fb0: setup mount namespace bind mounts mounting /etc/eb8/freeIPA/server/etc/krb5.conf into /var/lib/docker/btrfs/subvolumes/bc0f924401841f2ed92c088cb8089cadad2359126b9f6a3ff15b6cb657835fb0/etc/krb5.conf not a directory

it's allowed in docker!!!

fig.yml
```server:
    build: docker-freeipa
    ports:
        - "2222:22"
        - "53"
        - "80:80"
        - "443:443"
        - "389:389"
        - "636:636"
        - "88:88"
        - "464:464"
        - "123:123"

    environment:
        PASSWORD:   **************
        FORWARDER:  192.168.***.***

    hostname:     freeipa
    domainname:   ****.*****

    privileged:   true

    volumes:
        - /etc/eb8/freeIPA/server/etc/httpd/conf.d/:/etc/httpd/conf.d
        - /etc/eb8/freeIPA/server/etc/httpd/conf/:/etc/httpd/conf
        - /etc/eb8/freeIPA/server/etc/ipa/:/etc/ipa
        - /etc/eb8/freeIPA/server/etc/krb5.conf:/etc/krb5.conf
        - /etc/eb8/freeIPA/server/etc/pki-ca/:/etc/pki-ca
        - /etc/eb8/freeIPA/server/etc/ssh/:/etc/ssh
        - /etc/eb8/freeIPA/server/etc/sssd/:/etc/sssd
        - /etc/eb8/freeIPA/server/root/:/root
        - /etc/eb8/freeIPA/server/var/cache/ipa:/var/cache/ipa
        - /etc/eb8/freeIPA/server/var/lib/dirsrv/:/var/lib/dirsrv
        - /etc/eb8/freeIPA/server/var/lib/ipa-client/:/var/lib/ipa-client
        - /etc/eb8/freeIPA/server/var/lib/ipa/:/var/lib/ipa
        - /etc/eb8/freeIPA/server/var/log/:/var/log
```

ryanjm:
[...]

PR #1954, opened 2015-09-01
Allow for user relative paths on EXTEND

nhumrich:
Previously, user paths `~/somepath` did not resolve correctly.

For example:

    web:
      extends:
        file: ~/some/path.yml
        service: web

The `~/` currently doesnt work, you get the following error:

    [Errno 2] No such file or directory: u'/home/USER/folder/~/some/path/.yml'

This fixes that issue.

### DIFF
--- a/tests/unit/config_test.py
+++ b/tests/unit/config_test.py
@@ -950,6 +950,27 @@ class ExtendsTest(unittest.TestCase):
             load_from_filename('tests/fixtures/extends/nonexistent-service.yml')
 
 
+class ExpandPathTest(unittest.TestCase):
+    working_dir = '/home/user/somedir'
+
+    def test_expand_path_normal(self):
+        result = config.expand_path(self.working_dir, 'myfile')
+        self.assertEqual(result, self.working_dir + '/' + 'myfile')
+
+    def test_expand_path_absolute(self):
+        abs_path = '/home/user/otherdir/somefile'
+        result = config.expand_path(self.working_dir, abs_path)
+        self.assertEqual(result, abs_path)
+
+    def test_expand_path_with_tilde(self):
+        test_path = '~/otherdir/somefile'
+        with mock.patch.dict(os.environ):
+            os.environ['HOME'] = user_path = '/home/user/'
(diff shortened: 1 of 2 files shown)


dnephin:
I think this resolves #424

A unit test would be great

mnowster:
`expand_path` is used by a few other functions as well as `extends`, like `resolve_build_path` and `resolve_volume_path` for example. Are there any potential concerns about allowing userpaths to be expanded in those situations as well? I can't think of any but thought it was worth raising @dnephin @aanand 

The approach and test is LGTM from me.

dnephin:
I can't think of any potential problems with allowing `~` expansion on those paths.  

I wonder how this works on windows.

nhumrich:
`expand_user` works fine on windows, it expands to `C:\Users\USERNAME\` or something similar. The tests will not work on windows because the paths use `\` instead of `/` as the path sep, but most of the tests in this file have the same problem.

tests/unit/config_test.py
```diff
@@ -950,6 +950,26 @@ def test_load_throws_error_when_base_service_does_not_exist(self):
...
+        abs_path = '/home/user/otherdir/somefile'
+        result = config.expand_path(self.working_dir, abs_path)
+        self.assertEqual(result, abs_path)
+
+    def test_expand_path_with_tilde(self):
+        user_path = '~/otherdir/somefile'
+        result = config.expand_path(self.working_dir, user_path)
+        self.assertFalse('~' in result)
+        self.assertFalse(result.startswith(self.working_dir))
+        self.assertTrue(result.endswith(user_path[1:]))
```
nhumrich: That sounds reasonable. Ill try it out.

nhumrich:
looks like an intermittent error. How do I re-run tests?

dnephin:
I've got a fix for that failure in #2018, once it's merged you can rebase and you should be fine. Don't worry about it until then.

dnephin:
LGTM

aanand:
LGTM

joeldenning:
love this!

Issue #1955, opened 2015-09-01
test_scale_with_api_returns_unexpected_exception fails frequently on jenkins


dnephin:
`tests/integration/service_test.py ServiceTest.test_scale_with_api_returns_unexpected_exception` aka `Test that when scaling if the API returns an error, that is not of type` fails in many of our jenkins run

* https://jenkins.dockerproject.org/job/Compose-PRs/1213/console

with error:
```
Traceback (most recent call last):
  File "/code/tests/integration/testcases.py", line 22, in tearDown
    self.client.remove_container(c['Id'])
  File "/code/.tox/py27/lib/python2.7/site-packages/docker/utils/decorators.py", line 20, in wrapped
    return f(self, resource_id, *args, **kwargs)
  File "/code/.tox/py27/lib/python2.7/site-packages/docker/client.py", line 696, in remove_container
    self._raise_for_status(res)
  File "/code/.tox/py27/lib/python2.7/site-packages/docker/clientbase.py", line 103, in _raise_for_status
    raise errors.NotFound(e, response, explanation=explanation)
NotFound: 404 Client Error: Not Found ("Cannot destroy container faeacc68496793312980e36e20be6c7ae2a04a9b5f98c12f815a2235963ae834: Driver overlay failed to remove root filesystem faeacc68496793312980e36e20be6c7ae2a04a9b5f98c12f815a2235963ae834: stat /var/lib/docker/overlay/faeacc68496793312980e36e20be6c7ae2a04a9b5f98c12f815a2235963ae834: no such file or directory")
```

This is a docker engine bug, but something about this specific test triggers it frequently. Looking at the test, I can't really tell what it might be.

The failure is coming from `tearDown()`. We could wrap this cleanup in a try/except, which would have the added benefit of completing the rest of the cleanup.

dnephin:
Ok, I spent some time looking into this today, and I've made some progress. 


If I add a `time.sleep(0.1)` [here](https://github.com/docker/compose/blob/1.4.0/tests/integration/testcases.py#L21) between the `kill` and the `rm` it doesn't fail. So it likely has to do with container state.

If I inspect the container before the `rm` it's in state created, but the directory for the container does not exist on disk.

I'll continue trying to track down the error in the engine, but for now I think our options are either to try/except and skip the remove, or attempt removing the container a couple times with a short sleep between attempts.

dnephin:
[...]

PR #1960, opened 2015-09-02
Fix config_hash context to properly represent the service

### LINKED ISSUE #1757: Removing a service causes "up" to fail
Run this app:

```
web:
  image: nginx
  links:
   - redis
redis:
  image: redis
```

Now change to this:

```
web:
  image: nginx
```

and then:

```
$ docker-compose up
Starting composedemo_web_1...
Cannot start container 757b1fb7911f12a2cb06a22c7de192da687f5042daf6a4db30d76a487f71f94a: Cannot link to a non running container: /composedemo_redis_1 AS /composedemo_web_1/composedemo_redis_1
```

dnephin:
Fixes #1757, from the issue:

> Our config hash is missing a few fields (links, net, volumes_from, external_links), because we strip them from self.options. When the config hash doesn't change (because the change in configuration is in one of those fields), no container is recreated and we get this failure.

First and forth commit have the fix, the middle two are cleanup.


dnephin:
Ok, cleanup is in, the one test failure is #1955

aanand:
LGTM

bfirsh:
nice!

PR #1962, opened 2015-09-02
Create a ConvergenceStrategy enum

dnephin:
This is a small refactor so that we don't have to validate the correctness of convergence plan args in each layer (we were doing it 3 times).

The two booleans settings represent 3 valid states, and 1 invalid state. We can encapsulate those states with an enum. This way we only have to valid the arguments once (in the cli layer), and we can pass around the enum, which we know can only be in a correct state.

`enum34` is a backport of the core `enum` module from python3.4.

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -19,6 +19,7 @@ from ..progress_stream import StreamOutputError
 from ..project import ConfigurationError
 from ..project import NoSuchService
 from ..service import BuildError
+from ..service import ConvergenceStrategy
 from ..service import NeedsBuildError
 from .command import Command
 from .docopt_command import NoSuchCommand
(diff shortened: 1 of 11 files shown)


compose/cli/main.py
```diff
@@ -582,6 +574,21 @@ def version(self, project, options):
...
+    if force_recreate and not allow_recreate:
+        raise UserError("--force-recreate and --no-recreate cannot be combined.")
+
+    if force_recreate:
+        return ConvergenceStrategy.always
+
+    if not allow_recreate:
+        return ConvergenceStrategy.never
+
+    return ConvergenceStrategy.changed
```
aanand: Nice. Now that this is isolated, I think we can stop doing double-negatives and replace the `allow_recreate` variable with a `no_recreate` variable, right?

dnephin: Yup, that's an easy change, I'll do that

aanand: It makes me think we should've gone with a `--strategy` option, rather than multiple boolean options, but that's a discussion for another time

dnephin:
Build is green!

aanand:
LGTM

PR #1918, opened 2015-08-25
Fix scripts/test and add to note about pre-commit hook to contributing

dnephin:
@mnowster this should fix it

PR #1964, opened 2015-09-02
Remove .git from .dockerignore

lmesz:
Hello,

I have read the guideline how to contribute to compose, but when I tried to execute tests, pre-commit failed. So tests didn't have chance to run at all.
This was because tests run in a container in which the code was added but .git directory was on the exclude list, but pre-commit needs it.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,6 +1,5 @@
 *.egg-info
 .coverage
-.git
 .tox
 build
 coverage-html


dnephin:
@lmesz thanks for the contribution! The fix for this issue is in #1918 which has been stuck behind some unrelated test failures.  We should get it merged soon.

I don't know if we want to include `.git` in the image itself

dnephin:
And that fix was just merged, so you should be all set. Let me know if you have any problems with the new master.

PR #1966, opened 2015-09-02
Expose the CLI reference pages a bit better from the index

dnephin:
I noticed while reading over our docs that it was really hard for me to find the environment variables and command reference.  I think this change will make the situation better by cross-linking the reference pages to each other, and by adding a "where to go next" section to the overview page.

Also adds docs for a missing cli environment variable.

### DIFF
--- a/docs/reference/overview.md
+++ b/docs/reference/overview.md
@@ -31,6 +31,26 @@ Setting this is optional. If you do not set this, the `COMPOSE_PROJECT_NAME` def
 
 Specify the file containing the compose configuration. If not provided, Compose looks for a file named  `docker-compose.yml` in the current directory and then each parent directory in succession until a file by that name is found.
 
+### COMPOSE\_API\_VERSION
+
+The Docker API only supports requests from clients which report a specific
+version. If you receive a `client and server don't have same version error` using
+`docker-compose`, you can workaround this error by setting this environment
+variable. Set the version value to match the server version.
+
+Setting this variable is intended as a workaround for situations where you need
+to run temporarily with a mismatch between the client and server version. For
+example, if you can upgrade the client but need to wait to upgrade the server.
+
+Running with this variable set and a known mismatch does prevent some Docker
+features from working properly. The exact features that fail would depend on the
+Docker client and server versions. For this reason, running with this variable
+set is only intended as a workaround and it is not officially supported.
+
+If you run into problems running with this set, resolve the mismatch through
+upgrade and remove this setting to see if your problems resolve before notifying
+support.
+
 ### DOCKER\_HOST
 
 Sets the URL of the `docker` daemon. As with the Docker client, defaults to `unix:///var/run/docker.sock`.
--- a/docs/reference/overview.md
+++ b/docs/reference/overview.md
@@ -14,6 +14,13 @@ weight=-2
 
 This section describes the subcommands you can use with the `docker-compose` command.  You can run subcommand against one or more services. To run against a specific service, you supply the service name from your compose configuration. If you do not specify the service name, the command runs against all the services in your configuration.
 
+
+## Commands
+
+* [docker-compose Command](docker-compose.md)
+* [CLI Reference](index.md)
+
+
 ## Environment Variables
 
 Several environment variables are available for you to configure the Docker Compose command-line behaviour.
@@ -68,11 +75,6 @@ Configures the path to the `ca.pem`, `cert.pem`, and `key.pem` files used for TL
 
 Configures the time (in seconds) a request to the Docker daemon is allowed to hang before Compose considers
 it failed. Defaults to 60 seconds.
-
-
-
-
-
 
 
 ## Compose documentation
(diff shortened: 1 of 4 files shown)


docs/reference/overview.md
```diff
@@ -31,6 +38,12 @@ Setting this is optional. If you do not set this, the `COMPOSE_PROJECT_NAME` def
 
 Specify the file containing the compose configuration. If not provided, Compose looks for a file named  `docker-compose.yml` in the current directory and then each parent directory in succession until a file by that name is found.
 
+### COMPOSE\_API\_VERSION
+
+Specify the api version the client should report. If you get an error about
+`client and server don't have same version` you may use this environment
+variable to have the client report a version compatible with the server.
```
aanand: We should make it clear that this is not really safe or "supported".

dnephin: I wasn't sure how to word it. I was initially thinking of something like "Setting an api version may cause some features to fail, and is not officially supported."

aanand:
Seems like a good idea to me. (ping @moxiegirl)

Not strictly related to your changes, but I find the distinction between `reference/index.md`, `reference/overview.md` and `reference/docker-compose.md` rather confusing. What purpose does each serve?

Furthermore, the Compose command's environment variables are documented on a page called "Introduction to the CLI", whereas the page called "Compose environment variables reference" documents the environment variables that populate linked containers (which we expressly tell people not to use).

dnephin:
I completely agree with both of those points, but I didn't want to make too many changes up-front.

moxiegirl:
@aanand The `index.md` file has a special roll in the static file generation.  First, it serves as the attachment for the menu label:

image

It also allows for this kind of reference: `http://docs.docker.com/compose/reference/` where as overview `http://docs.docker.com/compose/reference/overview` .  I have plans in the next iteration of the docs to make these menu items also act as links not just as labels. Until that happens, I have the `overview.md` files around.

dnephin:
>  I have plans in the next iteration of the docs to make these menu items also act as links not just as labels

Nice, I've wanted this!

@moxiegirl what do you think about the doc changes in this PR ?

docs/reference/overview.md
thaJeztah: We had some discussion on this in docker; https://github.com/docker/docker/pull/15964#discussion_r38387671

It's not yet merged, but it will be described as a "power user" feature, and "use at your own risk"

moxiegirl: This is a bit ambiguous so more words clarify a bit:

The Docker command line client supports a specific version of the Docker API.  If you receive a `client and server don't have same version` error using the command line, setting this environment variable prevents this error. Set it to the version you want the client to report. typically a version compatible with the server.

dnephin: Hmm, the intent has changed a bit with this wording and now it's not really correct, how about:

> The Docker API only supports requests from clients which report a specific version.  If you receive a `client and server don't have same version` error using `docker-compose`, setting this environment variable prevents this error. Set it to the version you want the client to report. typically the server version.

> **NOTE**: Setting a different API version will prevent some features from working, and is not officially supported.

moxiegirl: Is it the specific version or a "matching" version? I think from your rewrite (noice!) same is true. So lets say matching.  

"report. typically" -> "report, typically"

Then I got to your note and hit a wall. Why are we suggesting they do this if we don't support it?  Which features do I lose if I **do** do it.  Why wouldn't I just fix the problem with the client/API mismatch ---> if this was in the discussion...let me know.  We kind of need to summarize the answers here to make the doc useful.

dnephin: "matching" sounds good.

>  Which features do I lose if I do do it.

That will depend on the version of the client and the server. It will be different for basically every combination, so we can't accurately enumerate them here.

> Why wouldn't I just fix the problem with the client/API mismatch

Sometimes (especially at larger orgs, or when there are shared hosts where the users don't necessarily have root) it takes a lot longer to upgrade the server. The server might only be upgraded every second or third version, where as a developer can easily upgrade the client (docker-compose) at any time.

So this environment variable is a workaround for when you want bug fixes or new features from the client, but you can't get a newer server just yet.

moxiegirl: @dnephin thanks for the answers.  Let's try something like what you see below.  It explains the situation much more completely and gives them a way to troubleshoot problems (assuming I got it right.) Also, I think we should notify Docker Support about this setting --- have y'all talked to them about it?

------>

The Docker API only supports requests from clients which report a specific version.  If you receive a `client and server don't have same version` error using `docker-compose`, you can workaround this error by setting this environment variable. Set the version value  to match the server version.

Setting this variable is intended as a workaround for situations where you need to run temporarily with a mismatch between the client and server version.  For example, if you can upgrade the client but need to wait to upgrade the server. 

Running with this variable set and a known mismatch does prevent some Docker features from working properly.  The exact features that fail would depend on the Docker client and server versions. For this reason,  running with this variable set is only intended as a workaround and it is not officially supported. 

If you run into problems running with this set, resolve the mismatch through upgrade and remove this setting to see if your problems resolve before notifying support.

dnephin: > I think we should notify Docker Support about this setting

I agree, what's the right way to communicate this type of thing with support?


That rewording sounds great! Thanks!

dnephin:
Wording for the `COMPOSE_API_VERSION` updated

moxiegirl:
@dnephin I talked with Andy and support and let him know.  I think it is 4X4 from here out. LGTM @dnephin from me.

aanand:
LGTM

PR #1968, opened 2015-09-03
Added csv option to ps

lmesz:
In our use-case we process docker-compose ps output programatically.
It is easier to process csv format instead of table.


aanand:
Thanks for the contribution. It's a good idea to have machine-readable output, but format-wise my vote definitely goes to JSON over CSV.

lmesz:
Okay I modify the request, commits and tests

lmesz:
Hm... Think about it and I close this PR and create a new one with patches, proper branch name etc...

PR #1970, opened 2015-09-03
fix expose schema definition

mnowster:
It was incorrectly defined in the schema that `expose` was an array of numbers, when it is valid to be either an array of numbers or an array of strings.

### DIFF
--- a/tests/unit/config_test.py
+++ b/tests/unit/config_test.py
@@ -239,6 +239,21 @@ class ConfigTest(unittest.TestCase):
                 )
             )
 
+    def test_valid_config_which_allows_two_type_definitions(self):
+        expose_values = [["8000"], [8000]]
+        for expose in expose_values:
+            service = config.load(
+                config.ConfigDetails(
+                    {'web': {
+                        'image': 'busybox',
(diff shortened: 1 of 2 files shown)


tests/unit/config_test.py
```diff
@@ -239,6 +239,21 @@ def test_config_extra_hosts_list_of_dicts_validation_error(self):
                 )
             )
 
+    def test_config_allows_either_type(self):
```
dnephin: Either type of "..."?  maybe `test_config_allows_expose_as_int_or_string` ?

mnowster: It was supposed to be generic, I'm trying not to fall into trap of testing every single schema field, more covering each aspect of the different kinds of schema functionality. So this could be expose or another field where the definition states that it can be either of 2 things. So really I'm testing `"type": ["something", "somethingelse"]` does what I expect it to do in the schema.

mnowster: The use case that highlighted this, was expose, so I used that.

dnephin:
LGTM

aanand:
LGTM

PR #1974, opened 2015-09-03
Bump requests to 2.7 to fix the ResponseNotReady() error

dnephin:
We've seen the occasional 

```
requests.exceptions.ConnectionError: ('Connection aborted.', ResponseNotReady('Request-sent',))
```

in #1783 and I've seen it a few times running against a `docker-machine` virtualbox.

This is fixed in `requests == 2.7` https://github.com/kennethreitz/requests/issues/2579

I also noticed that tox was missing a default posargs, which caused it to run  tests against other virtualenvs (jenkins isn't effected because there wouldn't be any on jenkins).

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,7 +3,7 @@ docker-py==1.3.1
 dockerpty==0.3.4
 docopt==0.6.1
 jsonschema==2.5.1
-requests==2.6.1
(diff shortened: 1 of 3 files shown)


aanand:
LGTM

Issue #1788, opened 2015-07-29
docker-compose does not pass environment variables true/false unchanged


jsonxr:
docker-compose does not pass environment variable 'true', 'false', 'yes', 'no' (without quotes) unchanged to the container. YML parser recognizes these values as a boolean and converts them to either True or False. This is confusing since environment variables are generally just a string.  I would expect all values to be passed to the container exactly as defined in the yml file as if they were a string.

Simple work around is to force them to be a string by quoting all environment variables.

test.sh

    echo $SHOW

docker-compose.yml

    test:
      image: ubuntu
      environment:
        SHOW: true
      command: sh /tmp/test.sh
      volumes:
        - './test.sh:/tmp/test.sh'

Expected Output:

    test_1 | true

Actual Output:

    test_1 | True

## Work Around:
Put all environment variables in quotes to force it to behave as a string.

    test:
      image: ubuntu
      environment:
        SHOW: 'true'
      command: sh /tmp/test.sh
      volumes:
        - './test.sh:/tmp/test.sh'

aanand:
Yep, this should be part of #129

PR #1976, opened 2015-09-03
Add Hint: Enclose bool variables in quotes when using dictionary

marb-dev:
Environment variables like `true` and `false` should be enclosed in quotes when using a dictionary; otherwise, their value will be parsed into `True` or `False` in the container, which makes them harder to be used within scripts.
Covers #1788

Signed-off-by: Markus Brück <marb-dev@users.noreply.github.com>

### DIFF
--- a/docs/yml.md
+++ b/docs/yml.md
@@ -184,17 +184,21 @@ Mount all of the volumes from another service or container.
 
 ### environment
 
-Add environment variables. You can use either an array or a dictionary.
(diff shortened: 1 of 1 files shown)


aanand:
It absolutely sucks that Compose does this. Instead of telling users to work around it, we should treat it as a bug and fix it. There's already an issue for that - https://github.com/docker/compose/issues/1788 - and we're making it a priority for Compose 1.5.

As such, I'm going to close this PR. Thank you for contributing!

PR #1982, opened 2015-09-04
Remove trailing white space from docs/index.md

mnowster:
Not sure how this got past the pre-commit hooks before being merged, but I've fixed it.


dnephin:
Ah, I noticed this failing after a rebased as well, nice!

I think the reason it got by pre-commit is because our CI doesn't run on docs-only changes, I'm not sure why.. but we should fix that.

LGTM

PR #1990, opened 2015-09-05
Remove or space out suspension dots after service name

### LINKED ISSUE #1975: up should not add 3 dots after container name. Makes slow to copy/paste
There are added 3 dots after the container name when doing `up`

```
# docker-compose up
Starting dockercomposetest_helloworld_1...
Attaching to dockercomposetest_helloworld_1
```
which makes copy/pasing the name on Linux slow, as you can't just double  click on the name without also getting the 3 dots. So one would have to drag the mouse cursor over the name and make sure not to get any of the dots included.

**Proposal**

Don't add the 3 dots.

shin-:
Fixes #1975

### DIFF
--- a/compose/service.py
+++ b/compose/service.py
@@ -143,27 +143,27 @@ class Service(object):
     # TODO: remove these functions, project takes care of starting/stopping,
     def stop(self, **options):
         for c in self.containers():
-            log.info("Stopping %s..." % c.name)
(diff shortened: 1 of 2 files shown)


aanand:
LGTM

thaJeztah:
Thanks @shin- !

PR #1992, opened 2015-09-05
Ps output in json

lmesz:
So in our environment we process docker-compose output programmatically.
It would be great if the output would be in machine readable format.
Previously I chose csv, but after @aanand comment I reworked the PR and made the ouput in json format.

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -1,6 +1,7 @@
 from __future__ import print_function
 from __future__ import unicode_literals
 
+import json
 import logging
 import re
 import signal
@@ -253,6 +254,7 @@ class TopLevelCommand(DocoptCommand):
 
         Options:
             -q    Only display IDs
+            --json    List containers in JSON format
         """
         containers = sorted(
             project.containers(service_names=options['SERVICE'], stopped=True) +
@@ -280,7 +282,10 @@ class TopLevelCommand(DocoptCommand):
                     container.human_readable_state,
                     container.human_readable_ports,
                 ])
-            print(Formatter().table(headers, rows))
+            if (options['--json']):
+                print(json.dumps(rows))
+            else:
+                print(Formatter().table(headers, rows))
 
     def pull(self, project, options):
         """
--- a/docs/reference/ps.md
+++ b/docs/reference/ps.md
@@ -16,6 +16,7 @@ Usage: ps [options] [SERVICE...]
 
 Options:
 -q    Only display IDs
+--json    List containers in JSON format
 ```
 
 Lists containers.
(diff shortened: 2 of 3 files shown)


lmesz:
Hm... How can I rerun tests ?
Based on the output It doesn't look like it cause my patch, looks like something wrong with the underlaying fs, but check again.

thaJeztah:
I see the tests are still running now, but if you need the tests to be restarted, ask anytime in the #docker-dev IRC channel if a maintainer can restart (mention compose and the PR number)

lmesz:
Okay thank you I didn't know that.
The tests are still failing in CI, on my local machine all of them are green.
It still looks like the overlayfs has some problem

lmesz:
Yeayyy, finally all greeen :)

aanand:
Rather than outputting an array of arrays - which isn't the easiest thing to process programmatically - why don't we just output an array of container config dictionaries? It'll be much more useful to anyone trying to script Compose.

compose/cli/formatter.py
```diff
@@ -23,3 +24,7 @@ def table(self, headers, rows):
         table.set_chars(['-', '|', '+', '-'])
 
         return table.draw()
+
+    def json(self, headers, rows):
+        rows.insert(0, headers)
+        return json.dumps(rows)
```
dnephin: I don't think this should be part of `Formatter()`. I see json output as a separate alternative to this human readable formatting.  Let's just call `json.dumps()` directly from `main`.

compose/cli/main.py
```diff
@@ -234,6 +234,7 @@ def ps(self, project, options):
 
         Options:
             -q    Only display IDs
+            --json    List containers in JSON format
```
dnephin: I'm wondering if we should actually support this as a top level option on `docker-compose` instead of an option on each subcommand.

lmesz: E.g.: docker-compose --json logs or up -d ? I think it would be the best.
If you agree update the PR's patches based on this advice.

compose/cli/formatter.py
lmesz: Okay I update PR.

lmesz:
@aanand ...and fix the array of array issue that you mentioned

compose/cli/main.py
```diff
@@ -280,7 +282,10 @@ def ps(self, project, options):
                     container.human_readable_state,
                     container.human_readable_ports,
                 ])
-            print(Formatter().table(headers, rows))
+            if (options['--json']):
+                print(json.dumps(rows))
```
dnephin: The json output should be a json object so that each field has a key. I believe with this formatting it would be just an array of fields without any keys.

dnephin:
Thanks for the PR.  We've had a bunch of different requests for ps, so I'm going to aggregate them all into #2369 so it'll be easier to design a solution that works for most of them.

I don't think we'll be able to accept this PR as is. The json format need a bit of work.  If you're still interested in making those changes, that would be great.

lmesz:
Hali,

Yes I still interested in, and sorry I didn't have time for the mentioned fixes. But I hope I will.
And understand that you said and close this PR.

PR #1993, opened 2015-09-05
Touchup "Extending services in Compose"

charleswhchan:
(The rendered version, if it's easier for review: 
https://github.com/charleswhchan/compose/blob/patch-2/docs/extends.md)

### DIFF
--- a/docs/extends.md
+++ b/docs/extends.md
@@ -45,8 +45,8 @@ looks like this:
         - "/data"
 
 In this case, you'll get exactly the same result as if you wrote
-`docker-compose.yml` with that `build`, `ports` and `volumes` configuration
-defined directly under `web`.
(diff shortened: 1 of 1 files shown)


aanand:
Thanks for the contribution!

In my opinion, this guide is going to need to change significantly as a result of #2051, since the process of configuring an app for multiple environments should now be simplified. Perhaps we should incorporate some of these changes when revising it.

dnephin:
Ya, I'm about to start working on #2141 which will make significant changes to this section.

I'll cherry-pick this commit first and try and preserve some of the edits.

charleswhchan:
Thanks @dnephin.

dnephin:
The new version is out.  I tried to incorporate some of these changes.  It did change pretty drastically, so I wasn't able to apply the cherry-pick, but hopefully the new version is better.

Thanks again for the PR!

Issue #1689, opened 2015-07-14
ipc as yml option?


saschaarthur:
Hey,

is it possible to use --IPC via yml file?

saschaarthur:
bump ? :+1:

titilambert:
+1

yank1:
+1

PR #1995, opened 2015-09-06
Added support for IPC namespaces, fixes GH-1689

predakanga:
Wasn't sure about the protocol regarding tests for minor changes like this, but there are no edge cases regarding the IPC setting - it's None, or a string, nothing more.

Signed-off-by: Lachlan Pease <predatory.kangaroo@gmail.com>

### DIFF
--- a/compose/config/config.py
+++ b/compose/config/config.py
@@ -36,6 +36,7 @@ DOCKER_CONFIG_KEYS = [
     'extra_hosts',
     'hostname',
     'image',
+    'ipc',
     'labels',
     'links',
     'mac_address',
--- a/compose/config/fields_schema.json
+++ b/compose/config/fields_schema.json
@@ -59,6 +59,7 @@
         "external_links": {"type": "array", "items": {"type": "string"}, "uniqueItems": true},
         "hostname": {"type": "string"},
         "image": {"type": "string"},
+        "ipc": {"type": "string"},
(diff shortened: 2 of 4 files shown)


mnowster:
Hi @predakanga , thanks for this. I double checked in docker-py and there is indeed support in `create_host_config` for ipc so yeh, we should support it too. 

You raise a good question about tests protocol. Looking at the test coverage we have for `_get_container_host_config`, we don't have a unit test for every option and this doesn't contain extra logic, so I'm OK with it.

LGTM. We're having some issues with Jenkins at the moment, the failure is unrelated to the code changes in this PR.

dnephin:
code looks good, but I think we should add it to the docs in `docs/yml.md` as well

predakanga:
Thanks for the feedback, @mnowster and @dnephin.

I'm adding the documentation at the minute, but noticed two other disparities - docker-compose is missing UTS and Ulimits configuration (see [docker-create](https://docs.docker.com/reference/commandline/create/)).

UTS isn't in docker-py yet, but Ulimits is - should I fold ulimit support into this PR, to save some time?

mnowster:
Hi @predakanga , ah good spot with the missing options, thanks for that :palm_tree: 

If you have the time, it'd be super helpful if you could create a separate issue and PR for `Ulimits`. We try to keep PRs to mostly just doing one thing, where possible, helps us review them.

Issue #1987, opened 2015-09-05
Extend an entire Compose file


bfirsh:
It is a common pattern to define a base Compose file, and then define Compose files for different environments with a few small changes.

It is currently possible to extend single services, but it is very verbose to include a large number of services and make a small change to one of them (for example, setting `RAILS_ENV=production`).

It should be possible to extend a Compose file with a complete set of services from another Compose file. All of those services will be copied into the Compose file, as if you were extending each of the services individually:

 - If you don't define a service in the child file, it is copied as-is.
 - If you do define a service, it should behave as if you've extended that single service.
 - There is no way of undefining a service (yet)

This is an intentionally simple first step, and I am intentionally not defining a syntax so we can discuss.

Design questions:

 - This is the first top-level configuration we have added to Compose. How should we do this?

Related issues / suggested designs: https://github.com/docker/compose/issues/318 #1380 [dcao-merge](https://github.com/dnephin/compose-addons#dcao-merge)

[(This is part of an initiative to define an app once in a way that can be used across dev, test and prod.)](https://github.com/docker/compose/issues/1784)

aanand:
Another related design is @dnephin's `dcao-merge` https://github.com/dnephin/compose-addons#dcao-merge

andrerom:
+100


#1 BUT, I guess you imply this but should be mentioned: it should keep the volume and link setup so you don't need to re define that.

#2 because of #1 it will have different semantics then extend. And personally I think it should allow composition, so that means "import" semantics. Example import application 1 and 2 and make them talk together by means of link/own-container..

aanm:
@bfirsh Let's assume a dev has a Compose file `comp-dev.yml` and op has a different one `comp-op.yml`, will be possible to run compose has:
`docker-compose --extends op-comp.yml -f comp-dev.yml up`

dnephin:
If one config depends on another, I would expect that dependency to be declared as part of the contents of the file, instead of a command line argument.

aanm:
It doesn't, it belongs to the operator. This way, the operator only needs to have its own compose file for its own infrastructure. Thus, for 10000 developers with 10000 different compose files, the single operator could easily override all options the he intended on any developer's compose file as:
```
for file in *-dev.yml; do
 docker-compose --extends op-comp.yml -f $file up -d
done
```

aanand:
After some discussion, @dnephin and I came up with the following:

[...]

PR #1999, opened 2015-09-07
Allow using `docker-compose.yml.dist`

iamluc:
Having one `docker-compose.yml` file in a project can sometime lead to conflicts with others developers as we all need to adapt it to our needs (ports, volumes, environments, etc..).

With this PR, I suggest to allow compose to look at `docker-compose.yml.dist` if `docker-compose.yml` does not exist.
So the `.dist` file can be added to the VCS of the project, but can still be overrided by creating the file `docker-compose.yml` (excluded in the VCS).

This strategy is used for example by [phpunit](https://phpunit.de)

### DIFF
--- a/compose/config/config.py
+++ b/compose/config/config.py
@@ -71,6 +71,7 @@ ALLOWED_KEYS = DOCKER_CONFIG_KEYS + [
 
 SUPPORTED_FILENAMES = [
     'docker-compose.yml',
+    'docker-compose.yml.dist',
     'docker-compose.yaml',
     'fig.yml',
     'fig.yaml',
@@ -109,7 +110,7 @@ def get_config_path(base_dir):
 
     winner = candidates[0]
 
-    if len(candidates) > 1:
+    if len(candidates) > 1 and not (len(candidates) == 2 and candidates[1].endswith('.dist')):
         log.warn("Found multiple config files with supported names: %s", ", ".join(candidates))
         log.warn("Using %s\n", winner)
 
--- a/docs/index.md
+++ b/docs/index.md
@@ -206,7 +206,7 @@ At this point, you have seen the basics of how Compose works.
 
 ## Release Notes
 
-To see a detailed list of changes for past and current releases of Docker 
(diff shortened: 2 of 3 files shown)


thaJeztah:
First impression: nice, I like it :+1:

dnephin:
Since this is a new convention, I think it's going to require some discussion.

My initial impression is that it might be good to support some kind of fallback (in addition to the "legacy filename" fallbacks we already do).  I'm not sure what the filenames should be, or if we should actually make it configurable.

#745 proposed supporting a `.compose` directory which would allow us to include configuration for this kind of thing (meta-configuration).

If we don't allow it to be configurable, I'd prefer to keep the `.yaml` extension, but I'm not sure how that would look.

It's also work mentioning that this can be accomplished pretty easily within a team or organization with a bash script wrapper and `-f` or `COMPOSE_FILE`

iamluc:
```
It's also work mentioning that this can be accomplished pretty easily within a team or organization with a bash script wrapper and -f or COMPOSE_FILE

**Comment by iamluc** [comment]
```
It's also work mentioning that this can be accomplished pretty easily within a team or organization with a bash script wrapper and -f or COMPOSE_FILE
```
Sure, but wrappers are often a pain because they are differents on each projects: path/name, options, etc...
IMHO, a standard is better

thaJeztah:
Yes, adding a wrapper is easy to do, but having it "officially" in compose will add to a standardized way to distribute a compose app (in stead of each project using their own way).

I think if a `docker-compose.yml.dist` is found, but *no* `docker-compose.yml`, Compose should output a warning, e.g. `**WARNING** using 'docker-compose.yml.dist' you should create your own project-file ..`

aanand:
See https://github.com/docker/compose/issues/1987#issuecomment-139632238 for an alternative solution which also serves the multi-environment use case.

mnowster:
Closing this as we now have the alternative solution done here https://github.com/docker/compose/pull/2051

PR #2000, opened 2015-09-07
Disallow booleans in environment

### LINKED ISSUE #1788: docker-compose does not pass environment variables true/false unchanged

mnowster:
This change will force people to put boolean values in quotes, thus ensuring
that the value gets passed through as intended.

Fixes #1788

### DIFF
--- a/compose/config/fields_schema.json
+++ b/compose/config/fields_schema.json
@@ -36,7 +36,15 @@
 
         "environment": {
           "oneOf": [
-            {"type": "object"},
+            {
+              "type": "object",
+              "patternProperties": {
+                "^[a-zA-Z0-9_]+$": {
+                  "type": ["string", "number"]
+                }
+              },
+              "additionalProperties": false
+            },
             {"type": "array", "items": {"type": "string"}, "uniqueItems": true}
           ]
         },
--- a/compose/config/fields_schema.json
+++ b/compose/config/fields_schema.json
@@ -39,7 +39,7 @@
             {
               "type": "object",
               "patternProperties": {
-                "^[a-zA-Z0-9_]+$": {
+                "^[^-]+$": {
                   "type": ["string", "number", "boolean"],
                   "format": "environment"
                 }
--- a/compose/config/fields_schema.json
+++ b/compose/config/fields_schema.json
@@ -40,7 +40,8 @@
               "type": "object",
               "patternProperties": {
                 "^[a-zA-Z0-9_]+$": {
-                  "type": ["string", "number"]
+                  "type": ["string", "number", "boolean"],
+                  "format": "environment"
                 }
               },
               "additionalProperties": false
--- a/compose/config/validation.py
+++ b/compose/config/validation.py
@@ -1,4 +1,5 @@
 import json
+import logging
 import os
 from functools import wraps
 
@@ -9,6 +10,9 @@ from jsonschema import RefResolver
 from jsonschema import ValidationError
 
 from .errors import ConfigurationError
+
+
+log = logging.getLogger(__name__)
 
 
 DOCKER_CONFIG_HINTS = {
@@ -41,6 +45,21 @@ def format_ports(instance):
         split_port(instance)
     except ValueError:
         return False
+    return True
+
+
+@FormatChecker.cls_checks(format="environment")
+def format_boolean_in_environment(instance):
+    """
+    Check if there is a boolean in the environment and display a warning.
+    Always return True here so the validation won't raise an error.
+    """
+    if isinstance(instance, bool):
+        log.warn(
+            "Warning: There is a boolean value, {0} in the 'environment' key.\n"
+            "Environment variables can only be strings.\nPlease add quotes to any boolean values to make them string "
+            "(eg, '{0}').\nThis warning will become an error in a future release. \r\n".format(instance)
+        )
     return True
 
 
@@ -259,15 +278,17 @@ def process_errors(errors, service_name=None):
 
 def validate_against_fields_schema(config):
     schema_filename = "fields_schema.json"
-    return _validate_against_schema(config, schema_filename)
+    format_checkers = ["ports", "environment"]
+    return _validate_against_schema(config, schema_filename, format_checkers)
 
 
 def validate_against_service_schema(config, service_name):
     schema_filename = "service_schema.json"
-    return _validate_against_schema(config, schema_filename, service_name)
+    format_checkers = ["ports"]
+    return _validate_against_schema(config, schema_filename, format_checkers, service_name)
 
 
-def _validate_against_schema(config, schema_filename, service_name=None):
+def _validate_against_schema(config, schema_filename, format_checker=[], service_name=None):
     config_source_dir = os.path.dirname(os.path.abspath(__file__))
     schema_file = os.path.join(config_source_dir, schema_filename)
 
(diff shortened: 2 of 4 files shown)


mnowster:
:person_frowning: the test failures are not to do with the changes, as far as I can tell.

aanand:
This will likely break a lot of existing files. Although the fix is simple, I think my vote would go to auto-converting boolean values to `"true"` and `"false"`. @dnephin, @shin- - thoughts?

shin-:
Yeah, on top of that, forcing explicit quotes is really unwieldy. I'd rather we convert boolean env variables to strings instead.

The small downside is that `yes` and `no` will be transformed if they're not quoted, but that's the case right now anyway.

compose/config/fields_schema.json
```diff
@@ -37,7 +37,15 @@
 
         "environment": {
           "oneOf": [
-            {"type": "object"},
+            {
+              "type": "object",
+              "patternProperties": {
+                "^[a-zA-Z0-9_]+$": {
```
dnephin: I think we should actually relax this to be any character (or maybe anything but dash).  While bash may not support everything, swarm at least needs colons (http://docs.docker.com/swarm/scheduler/filter/#affinity-filter), so it might be easier to just allow most characters for now.

aanand:
> The small downside is that `yes` and `no` will be transformed if they're not quoted, but that's the case right now anyway.

Oh, yuck. That's a pretty strong downside in my opinion (and reminds me - we've had the same problem with other options, e.g. `command: yes` becomes `"Command": ["True"]`).

compose/config/fields_schema.json
```diff
@@ -37,7 +37,15 @@
 
         "environment": {
```
dnephin: should this use the `list_or_dict` schema? maybe there are differences between them?

dnephin:
I really don't like converting types to other types. I prefer the strict option (as it's implemented) of failing, and documenting it in the reference. If someone wants a string, they should use a string!

If we suspect this is a common enough case, maybe we could allow for bools, but add a warning that it will be deprecated in the future. Give them a chance to upgrade and fix their config to use strings, and then remove it in the next release?

aanand:
> If we suspect this is a common enough case, maybe we could allow for bools, but add a warning that it will be deprecated in the future. Give them a chance to upgrade and fix their config to use strings, and then remove it in the next release?

+1. I do suspect it's common - there must be lots out there like this:

```yaml
environment:
  DEBUG: true
```

shin-:
> If someone wants a string, they should use a string!

But in most cases a literal value (without explicit quotes) will translate to a string in YAML, which is... fine in _most_ situations.

compose/config/fields_schema.json
mnowster: I didn't use `list_or_dict` for this one as we define some extra constraints in the `patternProperties` for the allowed key names. I was being restrictive cos I was worried about character compatibility for different systems.

compose/config/validation.py
```diff
@@ -44,6 +48,20 @@ def format_ports(instance):
     return True
 
 
+@FormatChecker.cls_checks(
+    format="environment")
```
dnephin: minor: this could be a single line

compose/config/fields_schema.json
dnephin: I think these constraints here are the same ones we'd want for `labels`, and most likely for `extra_hosts` as well (where are the two places that `list_or_dict` is used).

What do you think about moving this to be the new definition of `list_or_dict` ?

If we do that, I think we'd want to change the format to something like `non_bool_scalar`

compose/config/validation.py
```diff
@@ -44,6 +48,20 @@ def format_ports(instance):
...
+def format_boolean_in_environment(instance):
+    """
+    Check if there is a boolean in the environment and display a warning.
+    Always return True here so the validation won't raise an error.
+    """
+    if isinstance(instance, bool):
+        log.warn("Warning: There is a boolean value in the 'environment' key,")
+        log.warn("this will be deprecated in the future.")
+        log.warn("Please wrap booleans in quotes to ensure they're preserved. \r\n")
```
dnephin: I think we could include a bit more info in this error message, maybe something like:

```python
log.warn("Warning: There is a boolean value '{inst}' in the 'environment' key. Environment variables "
         "can only be strings. Please add quotes to any boolean values to make them strings"
         " (ex: '{inst}'). This warning will become an error in a future release.".format(inst=instance))
```

compose/config/validation.py
```diff
@@ -216,15 +234,17 @@ def _parse_valid_types_from_schema(schema):
...
     schema_filename = "fields_schema.json"
-    return _validate_against_schema(config, schema_filename)
+    format_checkers = ["ports", "environment"]
+    return _validate_against_schema(config, schema_filename, format_checkers)
 
 
 def validate_against_service_schema(config, service_name):
     schema_filename = "service_schema.json"
-    return _validate_against_schema(config, schema_filename, service_name)
+    format_checkers = ["ports"]
```
dnephin: Why is only the ports used here, but both are used above?

Would we get double warnings if we use the `environment` format here?

compose/config/fields_schema.json
mnowster: I honestly can't remember now why I thought they had separate requirements and I can't find it in my notes. If you can't think of any issues with doing that, I think that's a good suggestion, I will amend. Thanks.

compose/config/validation.py
mnowster: I had been trying to do something clever previously with calling the `log.warn` instead of an exception but it didn't pan out. So yep, I'll make it one line.

compose/config/validation.py
mnowster: Exactly, it gave double warnings. Currently the code path doesn't get to validate the full service schema unless it's gone through the fields validation, so thought that was the best place for it.

compose/config/validation.py
mnowster: Agreed that is clearer, will amend.

compose/config/fields_schema.json
dnephin: Oh, hmm, so I guess dashes are valid in label and hostnames, so that is one issue. We could just relax this to "any character" which would make it usable for all three.

compose/config/validation.py
dnephin: Cool, maybe an inline comment for that so we remember why these lists are different?

dnephin: Thinking about this some more, I think there might be a better way to handle this. Instead of repeating the validation in the second schema, why not do something like this:

1) validate all the fields against `fields_schema.json`
2) remove the `$ref: fields_schema.json#/definitions/service` from `service_schema.json` and only validate the required fields exist. The other fields can be handled by `additionalProperties: true`, since we've already validated that all the fields are correct.

I think that way the format warning only logs once.

I'd actually be in favor of dropping support for abstract base services, but that's a topic for another time.

mnowster: I feel like this is outside scope of this PR. Can we come back to that later in a separate PR as I think depending on how some of the future extend/import work is going to happen, this indeed will need to be re-factored but I'd like space for us to have a separate discussion around that.

Heh, yeh abstract base services are a bit of a pain.

compose/config/fields_schema.json
mnowster: I think that would be better in a separate PR as it effects other fields. I'll look at doing that after this one has gone in.

dnephin: cool, that sounds fair

aanand:
LGTM

Issue #1997, opened 2015-09-06
Entrypoint should be allowed to be a string


dnephin:
### Entrypoint as string

```yaml
web:
    image: alpine:edge
    entrypoint: 'sh'
```

`Service 'web' configuration key 'entrypoint' contains an invalid type, it should be an array`

I think string should be valid.

thaJeztah:
`['foo', 'bar']` is not valid JSON (use double quotes for strings `["foo", "bar"]`); could that be the problem here?

dnephin:
@thaJeztah nope, this is yaml (I verified that pyyaml parses it correctly).

The error is that `uniqueItems` is being set to True

thaJeztah:
@dnephin ah, right, I wasn't sure if compose would send the *literal* command to Docker, resulting in errors on Docker's side. Sorry for the noise here :+1:

mnowster:
@dnephin I've split this into 2 issues, as they're very different. Error message is now over at https://github.com/docker/compose/issues/2002. 

Thanks for catching the `entrypoint` mis-type, there's probably going to be a few more of those lurking :eyes:

dnephin:
makes sense, thanks for looking into it!

PR #2001, opened 2015-09-07
Allow entrypoint to be a list or string

mnowster:
Previously it was set to only allow a list.

Fixes https://github.com/docker/compose/issues/1997


aanand:
LGTM

PR #2005, opened 2015-09-08
Fix error message and class name from Boot2Docker to DockerMachine

lnmunhoz:
When I tried to run `docker-compose build` I get the error message "Couldn't connect to Docker daemon - you might need to run boot2docker up". But now the daemon uses `docker-machine` by default and not `boot2docker`.

### DIFF
--- a/compose/cli/errors.py
+++ b/compose/cli/errors.py
@@ -40,10 +40,10 @@ class DockerNotFoundGeneric(UserError):
         """)
 
 
-class ConnectionErrorBoot2Docker(UserError):
+class ConnectionErrorDockerMachine(UserError):
     def __init__(self):
-        super(ConnectionErrorBoot2Docker, self).__init__("""
-        Couldn't connect to Docker daemon - you might need to run `boot2docker up`.
+        super(ConnectionErrorDockerMachine, self).__init__("""
+        Couldn't connect to Docker daemon - you might need to run `docker-machine start default`.
         """)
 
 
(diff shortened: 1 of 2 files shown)


compose/cli/errors.py
```diff
@@ -40,10 +40,10 @@ def __init__(self):
         """)
 
 
-class ConnectionErrorBoot2Docker(UserError):
+class ConnectionErrorDockerMachine(UserError):
     def __init__(self):
-        super(ConnectionErrorBoot2Docker, self).__init__("""
-        Couldn't connect to Docker daemon - you might need to run `boot2docker up`.
+        super(ConnectionErrorDockerMachine, self).__init__("""
+        Couldn't connect to Docker daemon - you might need to run `docker-machine start default`.
```
dnephin: Where does `default` come from?  It looks like the machine docs use `dev` (https://docs.docker.com/machine/get-started/).

Is this the machine name that toolbox uses?

dnephin:
Good idea, we should update this.

Thanks for the contribution!

aanand:
I believe `default` is the name used by Toolbox, yes - ping @jeffdm

lnmunhoz:
@dnephin It is the default machine that toolbox creates. 
See in https://docs.docker.com/installation/mac/

dnephin:
In that case, I wonder if we should have the machine docs use this name as well.


LGTM, the test failure is unrelated and already ticketed.

lnmunhoz:
@dnephin I don't know if the test is related with the class names that I have changed. How to fix this?

lnmunhoz:
@dnephin Oh I just saw that is unrelated :+1:

mnowster:
LGTM

PR #2010, opened 2015-09-08
Set the terminal title to what Docker-Compose is currently doing

jrabbit:
I find it difficult when my builds generate a lot of output to track exactly whats going on. So I hunted down the xterm title control sequences and xprop method to store the current title so we clean up after ourselves in case the shell does not automatically (zsh/oh-my-zsh seems to).

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -115,6 +115,7 @@ class TopLevelCommand(Command):
       help               Get help on a command
       kill               Kill containers
       logs               View output from containers
+      pause              Pause services
       port               Print the public port for a port binding
       ps                 List containers
       pull               Pulls service images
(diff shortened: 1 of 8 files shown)


jrabbit:
Do I need to sign-off the rest of the commits?

thaJeztah:
All commits need to be signed, but in most cases, it's preferred to have only a single commit in a PR, so you'll need to "squash" them. The contribution guide of the docker documentation describes how you can do that; https://docs.docker.com/project/work-issue/#pull-and-rebase-frequently

jrabbit:
I think I fixed it but I'm not super-proficient with rebase so my work should be double-checked

dnephin:
Thanks for the contribution!

Unfortunately this feels to me like a concern that is outside the scope of compose. There are many different execution environments and terminals, and adding code for any specific one risks introducing bugs in others.

Isn't this something the terminal can handle itself? It should be possible to configure the terminal to set the title based on the command (at least in some terminals).

jrabbit:
The issue from where I am is that docker-compose yes is automatically in the title but what it's building is a mystery if its past my scrollback maximum

mnowster:
@jrabbit in most terminal preferences you can set what your scrollback maximum is, or indeed set it to unlimited.

@dnephin is right, this is outside of scope of compose I'm afraid.

PR #2011, opened 2015-09-08
Add more github label areas

dnephin:
Adds `area/volumes` and `area/packaging`, since those are two rather large areas that aren't covered by the existing ones.

Also set a fixed python  version for the import ordering hook. Otherwise developers who run this with python3 will see `enum` re-ordered.


aanand:
LGTM

Issue #2002, opened 2015-09-07
Error message for oneOf validation isn't bubbling up fine enough detail


mnowster:
(reported by @dnephin)

### Duplicate strings in command

```yaml
web:
    image: alpine:edge
    command: ['sh', 'sh']
```

`Service 'web' configuration key 'command' contains an invalid type, ...`

If I do the same thing with `entrypoint`, I get a better error message `Service 'web' configuration key 'entrypoint' value ['sh', 'sh'] has non-unique element`, but it's still an error

mnowster:
There are reasons for this, it's because one is a oneOf validation error and the other is a single type with a constraint violation of non unique. Parsing the oneOf validation error is more complex and at the moment is giving the simplest error message it can. The other we get the nicer error message for free as part of the error tree returned by jsonschema.

I'm doing a bit of spiking of some refactoring to attempt some improvements to oneOf error message parsing :crystal_ball:

dnephin:
I think there are two things going on here as well, the error message is one, but the other is that the array is set to `uniqueItems: true`, but it should be `false`.  Is that covered by this issue or #1997 ?

mnowster:
Ah, I missed the wanting `uniqueItems` to be false, sorry. I can create a separate issue for that one too. It'll help me keep track of them as I fix :hammer: 

To aid my own understanding of how these values are used, what's the use case for allowing non unique items in commands?

aanand:
Well a command list is just a Unix command-line invocation broken up into a list, so could contain any old stuff, e.g. `["my-app", "--arg", "value", "--arg-2", "value"]`

mnowster:
Thanks for that @aanand :+1:

PR #2014, opened 2015-09-09
Improve oneOf schema validation messaging

mnowster:
Refactor to do a bit more work to provide finer grained error messages when a oneOf schema ValidationError occurs.

Closes https://github.com/docker/compose/issues/2002

### DIFF
--- a/compose/config/validation.py
+++ b/compose/config/validation.py
@@ -107,16 +107,6 @@ def process_errors(errors, service_name=None):
     def _clean_error_message(message):
         return message.replace("u'", "'")
 
-    def _parse_valid_types_from_schema(schema):
-        """
-        Our defined types using $ref in the schema require some extra parsing
-        retrieve a helpful type for error message display.
-        """
-        if '$ref' in schema:
-            return schema['$ref'].replace("#/definitions/", "").replace("_", " ")
-        else:
-            return str(schema['type'])
-
     def _parse_valid_types_from_validator(validator):
         """
         A validator value can be either an array of valid types or a string of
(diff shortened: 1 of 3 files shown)


compose/config/validation.py
```diff
@@ -107,15 +107,73 @@ def _parse_key_from_error_msg(error):
...
+            msg = "contains non unique items, please remove duplicates from {}".format(
+                uniqueness[0].instance
+            )
+            return msg
+
+        types = [context.validator_value for context in error.context if context.validator == 'type']
+        if len(types) == 1:
+            valid_types = _parse_valid_types_from_validator(types[0])
+        else:
+            valid_types = _parse_valid_types_from_validator(types)
```
dnephin: I don't understand this if/else, why is the len == 1 a special case?

compose/config/validation.py
```diff
@@ -107,15 +107,73 @@ def _parse_key_from_error_msg(error):
...
         """
-        if '$ref' in schema:
-            return schema['$ref'].replace("#/definitions/", "").replace("_", " ")
+        pre_msg_type_prefix = "a"
+        last_msg_type_prefix = "a"
+        types_requiring_an = ["array", "object"]
+
+        if isinstance(validator, list):
+            last_type = validator.pop()
+            types_from_validator = ", ".join(validator)
```
dnephin: If I'm not mistaken, this only works for lists with two items, is that correct?

compose/config/validation.py
```diff
@@ -107,15 +107,73 @@ def _parse_key_from_error_msg(error):
...
+            msg = "{} {} or {} {}".format(
+                pre_msg_type_prefix,
+                types_from_validator,
+                last_msg_type_prefix,
+                last_type
+            )
         else:
-            return str(schema['type'])
+            if validator in types_requiring_an:
+                pre_msg_type_prefix = "an"
```
dnephin: This check seems to be repeated a few times. I think it would be good to make this a function, which would let us remove all the prefix variables and `types_requiring_an` from this function

```python

def anglicize_validator(validator):
    if validator in ["array", "object"]:
        return 'an ' + validator
    return 'a ' + validator
```

I think this change would return the size of this function quite a bit.

aanand: Yeah, I think I would prefer this too.

mnowster:
Thanks @dnephin for the refactor suggestion, have amended and it is much clearer. It also meant I could remove the no longer needed special case if, which was a way to work around a one element list.

compose/config/validation.py
```diff
@@ -107,15 +113,60 @@ def _parse_key_from_error_msg(error):
...
-        retrieve a helpful type for error message display.
+        A validator value can be either an array of valid types or a string of
+        a valid type. Parse the valid types and prefix with the correct article.
         """
-        if '$ref' in schema:
-            return schema['$ref'].replace("#/definitions/", "").replace("_", " ")
+        if isinstance(validator, list):
+            if len(validator) >= 2:
+                first_type = anglicize_validator(validator.pop(0))
+                last_type = anglicize_validator(validator.pop())
```
dnephin: I'm not sure if validator is re-used at all, but I would probably err on the side of caution here and not do any mutation.

Something like this would work:

```python
first_type, last_type, validator = validator[0], validator[-1], validator[1:-1]
```

mnowster: It doesn't get re-used but I don't have a strong opinion against doing this, so have amended :)

aanand:
LGTM

PR #2016, opened 2015-09-09
Remove uniqueness constraint for command

mnowster:
The command value can be a list, which would be a Unix command-line
invocation broken up into individual values, thus needing the ability to
have non unique values.


dnephin:
LGTM

aanand:
LGTM

Issue #2007, opened 2015-09-08
config schema allows for `detach` field


dnephin:
I see `detach: {type: boolean}` in `fields_schema.json`, but I don't see it referenced in the docs or code (it's a cli param for `run`, but not a config param).

Was this added by mistake? I think we should remove it.

aanand:
Yeah, that looks wrong to me.

PR #2017, opened 2015-09-09
Remove mistaken field

mnowster:
detach is a run param, not a config param. Oops, sorry!

Fixes https://github.com/docker/compose/issues/2007


dnephin:
LGTM

PR #2018, opened 2015-09-09
 Handle unexpected errors, but don't ignore background threads

### LINKED ISSUE #1955: test_scale_with_api_returns_unexpected_exception fails frequently on jenkins

dnephin:
Fixes #1955 

See the issue for description of the problem.

This does change the behaviour of `parallel_execute()` with unexpected exceptions (prints error instead of raises the exception). In order to handle this properly and still raise an exception a much larger refactor would be necessary (to attempt to join or kill unfinished threads), so I went with this change instead.

Let me know if you feel it's important to raise the unexpected exception and I can take another pass.

### DIFF
--- a/compose/utils.py
+++ b/compose/utils.py
@@ -21,7 +21,6 @@ def parallel_execute(objects, obj_callable, msg_index, msg):
     """
     stream = get_output_stream(sys.stdout)
     lines = []
-    errors = {}
 
     for obj in objects:
         write_out_msg(stream, lines, msg_index(obj), msg)
@@ -29,16 +28,17 @@ def parallel_execute(objects, obj_callable, msg_index, msg):
     q = Queue()
 
     def inner_execute_function(an_callable, parameter, msg_index):
+        error = None
         try:
             result = an_callable(parameter)
         except APIError as e:
-            errors[msg_index] = e.explanation
+            error = e.explanation
             result = "error"
         except Exception as e:
-            errors[msg_index] = e
+            error = e
             result = 'unexpected_exception'
 
-        q.put((msg_index, result))
+        q.put((msg_index, result, error))
 
     for an_object in objects:
         t = Thread(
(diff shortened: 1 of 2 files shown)


dnephin:
Restored the `raise error` behaviour

compose/utils.py
```diff
@@ -29,16 +28,16 @@ def parallel_execute(objects, obj_callable, msg_index, msg):
...
-            errors[msg_index] = e.explanation
+            error = e.explanation
             result = "error"
-        except Exception as e:
-            errors[msg_index] = e
+        except Exception as error:
             result = 'unexpected_exception'
 
-        q.put((msg_index, result))
+        q.put((msg_index, result, error))
```
dnephin: Apparently this breaks in py3 with 

```
UnboundLocalError: local variable 'error' referenced before assignment
```

If I change the second except to `except Exception as e: error = e` it succeeds.

I guess there is something special about using a variable in the exception handler

dnephin:
Build is green!

aanand:
green_field_wallpaper-wide

PR #2019, opened 2015-09-09
Allow to extend service using shorthand notation. Closes #1989

### LINKED ISSUE #1989: "service" key in extending should be optional
This is unnecessarily verbose:

```
base:
  ...
web:
  extends:
    service: base
  ...
```

You should be able to just do this:

```
base:
  ...
web:
  extends: base
  ...
```

mrfuxi:
I looked at using a shorthand notation to extend a service (Issue #1989)
Seams to be working ok, but validation errors got more generic. Can someone advise how to make them as specific as they where before?

### DIFF
--- a/compose/config/validation.py
+++ b/compose/config/validation.py
@@ -4,6 +4,7 @@ import os
 import sys
 from functools import wraps
 
+import six
 from docker.utils.ports import split_port
 from jsonschema import Draft4Validator
 from jsonschema import FormatChecker
@@ -162,10 +163,25 @@ def process_errors(errors, service_name=None):
         Inspecting the context value of a ValidationError gives us information about
         which sub schema failed and which kind of error it is.
         """
+
+        required = [context for context in error.context if context.validator == 'required']
+        if required:
+            return required[0].message
+
+        additionalProperties = [context for context in error.context if context.validator == 'additionalProperties']
+        if additionalProperties:
+            invalid_config_key = _parse_key_from_error_msg(additionalProperties[0])
+            return "contains unsupported option: '{}'".format(invalid_config_key)
+
         constraint = [context for context in error.context if len(context.path) > 0]
         if constraint:
             valid_types = _parse_valid_types_from_validator(constraint[0].validator_value)
-            msg = "contains {}, which is an invalid type, it should be {}".format(
+            invalid_config_key = "".join(
+                "'{}' ".format(fragment) for fragment in constraint[0].path
+                if isinstance(fragment, six.string_types)
+            )
+            msg = "{}contains {}, which is an invalid type, it should be {}".format(
+                invalid_config_key,
                 constraint[0].instance,
                 valid_types
             )
--- a/compose/config/config.py
+++ b/compose/config/config.py
@@ -275,15 +275,19 @@ class ServiceLoader(object):
         self.service_dict['environment'] = env
 
     def validate_and_construct_extends(self):
+        extends = self.service_dict['extends']
+        if not isinstance(extends, dict):
+            extends = {'service': extends}
+
         validate_extends_file_path(
             self.service_name,
-            self.service_dict['extends'],
+            extends,
             self.filename
         )
         self.extended_config_path = self.get_extended_config_path(
-            self.service_dict['extends']
+            extends
         )
-        self.extended_service_name = self.service_dict['extends']['service']
+        self.extended_service_name = extends['service']
 
         full_extended_config = pre_process_config(
             load_yaml(self.extended_config_path)
--- a/compose/config/fields_schema.json
+++ b/compose/config/fields_schema.json
@@ -57,14 +57,21 @@
         },
 
         "extends": {
-          "type": "object",
+          "oneOf": [
+            {
+              "type": "string"
+            },
+            {
+              "type": "object",
 
-          "properties": {
-            "service": {"type": "string"},
-            "file": {"type": "string"}
-          },
-          "required": ["service"],
-          "additionalProperties": false
+              "properties": {
+                "service": {"type": "string"},
+                "file": {"type": "string"}
+              },
+              "required": ["service"],
+              "additionalProperties": false
+            }
+          ]
         },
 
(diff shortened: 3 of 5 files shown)


dnephin:
The generic validation errors are being worked on in #2014

mrfuxi:
Thanks for pointing it out.
So would it be better to wait until that PR is merged and I will adjust my code accordingly, or should I "fix" errors by expecting more generic errors msgs?

dnephin:
I'm actually not sure, but I think the answer is wait for it to be merged, because it will handle all cases of `oneOf`

mrfuxi:
Thank you. I will wait for #2014 then

mrfuxi:
Can someone have a quick look at this.
If it needs any more work, please let me know, I'm happy to do the changes

compose/config/validation.py
```diff
@@ -162,10 +164,25 @@ def _parse_oneof_validator(error):
...
         """
+
+        required = [context for context in error.context if context.validator == 'required']
+        if required:
+            return required[0].message
+
+        additionalProperties = [context for context in error.context if context.validator == 'additionalProperties']
+        if additionalProperties:
+            invalid_config_key = _parse_key_from_error_msg(additionalProperties[0])
+            return "contains unsupported option: '{}'".format(invalid_config_key)
```
dnephin: What's this block for?

compose/config/validation.py
```diff
@@ -121,15 +122,16 @@ def anglicize_validator(validator):
     return 'a ' + validator
 
 
+def _parse_key_from_error_msg(error):
```
mnowster: This doesn't belong out here, it's a private internal function for `process_errors` and not to be used outside of it.

mrfuxi: Sorry for that, at some point I had a reason to move it out but it's no longer valid

compose/config/validation.py
mrfuxi: It's providing better error messages for oneOf (tests.unit.config.config_test).

additionalProperties:
- test_config_invalid_environment_dict_key_raises_validation_error
- test_extends_validation_invalid_key

required:
- test_extends_validation_empty_dictionary
- test_extends_validation_missing_service_key

mrfuxi:
bump

compose/config/validation.py
```diff
@@ -162,10 +163,25 @@ def _parse_oneof_validator(error):
...
+        additionalProperties = [context for context in error.context if context.validator == 'additionalProperties']
+        if additionalProperties:
+            invalid_config_key = _parse_key_from_error_msg(additionalProperties[0])
+            return "contains unsupported option: '{}'".format(invalid_config_key)
+
         constraint = [context for context in error.context if len(context.path) > 0]
         if constraint:
             valid_types = _parse_valid_types_from_validator(constraint[0].validator_value)
-            msg = "contains {}, which is an invalid type, it should be {}".format(
+            path = [
```
mnowster: This feels a bit like overkill to me. The contents of the `path` deque will be string keys, so I don't think we need to test for the instance of string here.

The depth of config keys is only two deep. So we already have the service name, already have the config key before this function is called, so we can confidently at this point look at just `constraint[0].path`.

`        if constraint:
            valid_types = _parse_valid_types_from_validator(constraint[0].validator_value)
            msg = "'{}' contains {}, which is an invalid type, it should be {}".format(
                "".join(constraint[0].path),
                constraint[0].instance,
                valid_types
            )
`

mrfuxi: I can't be 100% certain but sometimes you get index here (to the array)

mnowster:
One minor change requested, rest is LGTM.

Sorry it took a while @mrfuxi we are knee deep in trying to get ready for an rc.

compose/config/validation.py
mnowster: Oh, hmm. Are you able to replicate in what situation that happens? As I wouldn't expect it too.

mrfuxi: I will try to. Most likely there is a test for that or I will try to create one. If not I will fallow the suggestion

mrfuxi:
Np @mnowster I don't want to be too annoying when you're bussy :)

compose/config/validation.py
mrfuxi: With suggested change 2 tests fail:
- test_invalid_list_of_strings_format

```
E           AssertionError: "Service 'web' configuration key 'command' contains 1, which is an invalid type, it should be a string" does not match "Validation failed, reason(s):
E           Service 'web' configuration key 'command' '0' contains 1, which is an invalid type, it should be a string"
```

- test_config_extra_hosts_list_of_dicts_validation_error

```
E           AssertionError: "key 'extra_hosts' contains {'somehost': '162.242.195.82'}, which is an invalid type, it should be a string" does not match "Validation failed, reason(s):
E           Service 'web' configuration key 'extra_hosts' '0' contains {'somehost': '162.242.195.82'}, which is an invalid type, it should be a string"
```

mnowster: Aaaah, I see what's happening. I was focussed on `extends`, which being a dict would work happily, but the fields you've chosen in the tests highlights their `oneOf` contains a type of list with a format of items within that list. Right. Gotcha. Good catch :+1: 

Ok. I think we can make it a little clearer what's going on, I know the validator constraint is called `path`, what we're doing is trying to find the config key that contains something invalid. So how about something like:

```
            invalid_config_key = "".join(
                ["'{}' ".format(key) for key in constraint[0].path
                    if isinstance(key, six.string_types)
                ]
            )
```

mnowster:
We've got some problems with jenkins atm :cloud:  once they're resolved, I'll kick off a rebuild of this.

mrfuxi:
Cool, so far appveyor is way more stable :)

mnowster:
LGTM and green so I'm merging.

Can you open a separate PR please @mrfuxi to include docs for this change? I should have spotted that earlier, apols.

mrfuxi:
Sorry for that @mnowster 
I will do it asap

PR #2023, opened 2015-09-10
implement exec

### LINKED ISSUE #593: Support docker exec command
I've just learned about the new docker [exec](https://docs.docker.com/reference/commandline/cli/#exec) command. Wouldn't it be great to add it to fig ?

I think it really makes sense with fig, as you could run `fig up` and then exec some commands (like a shell) to debug your running containers if something happen.
Of course this is already possible by running `docker exec -it your_container cmd`, but I think that it wouldn't hurt to have `fig exec service cmd`.

I'll try to give it a shot and submit a PR.

TomasTomecek:
Resolves #593 

Still wip

- [x] basic functionality
- [x] tests
- [x] docs
- [x] interactive exec (`docker exec -ti` equiv)

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -43,7 +43,7 @@ from .utils import yesno
 
 
 if not IS_WINDOWS_PLATFORM:
-    from dockerpty.pty import PseudoTerminal, RunOperation
+    from dockerpty.pty import PseudoTerminal, RunOperation, ExecOperation
 
 log = logging.getLogger(__name__)
 console_handler = logging.StreamHandler(sys.stderr)
@@ -152,6 +152,7 @@ class TopLevelCommand(DocoptCommand):
       create             Create services
       down               Stop and remove containers, networks, images, and volumes
       events             Receive real time events from containers
+      exec               Execute a command in a running container
       help               Get help on a command
       kill               Kill containers
       logs               View output from containers
@@ -297,6 +298,57 @@ class TopLevelCommand(DocoptCommand):
             formatter = json_format_event if options['--json'] else format_event
             print(formatter(event))
             sys.stdout.flush()
+
+    def exec_command(self, project, options):
+        """
+        Execute a command in a running container
+
+        Usage: exec [options] SERVICE COMMAND [ARGS...]
+
+        Options:
+            -d                Detached mode: Run command in the background.
+            --privileged      Give extended privileges to the process.
+            --user USER       Run the command as this user.
+            -T                Disable pseudo-tty allocation. By default `docker-compose exec`
+                              allocates a TTY.
+            --index=index     index of the container if there are multiple
+                              instances of a service [default: 1]
+        """
+        index = int(options.get('--index'))
+        service = project.get_service(options['SERVICE'])
+        try:
+            container = service.get_container(number=index)
+        except ValueError as e:
+            raise UserError(str(e))
+        command = [options['COMMAND']] + options['ARGS']
+        tty = not options["-T"]
+
+        create_exec_options = {
+            "privileged": options["--privileged"],
+            "user": options["--user"],
+            "tty": tty,
+            "stdin": tty,
+        }
+
+        exec_id = container.create_exec(command, **create_exec_options)
+
+        if options['-d']:
+            container.start_exec(exec_id, tty=tty)
+            return
+
+        signals.set_signal_handler_to_shutdown()
+        try:
+            operation = ExecOperation(
+                    project.client,
+                    exec_id,
+                    interactive=tty,
+            )
+            pty = PseudoTerminal(project.client, operation)
+            pty.start()
+        except signals.ShutdownException:
+            log.info("received shutdown exception: closing")
+        exit_code = project.client.exec_inspect(exec_id).get("ExitCode")
+        sys.exit(exit_code)
 
     def help(self, project, options):
         """
--- a/compose/container.py
+++ b/compose/container.py
@@ -216,6 +216,12 @@ class Container(object):
     def remove(self, **options):
         return self.client.remove_container(self.id, **options)
 
+    def create_exec(self, command, **options):
+        return self.client.exec_create(self.id, command, **options)
+
+    def start_exec(self, exec_id, **options):
+        return self.client.exec_start(exec_id, **options)
+
     def rename_to_tmp_name(self):
         """Rename the container to a hopefully unique temporary container name
         by prepending the short id.
--- a/compose/cli/docopt_command.py
+++ b/compose/cli/docopt_command.py
@@ -43,6 +43,10 @@ class DocoptCommand(object):
 
     def get_handler(self, command):
         command = command.replace('-', '_')
+        # we certainly want to have "exec" command, since that's what docker client has
+        # but in python exec is a keyword
+        if command == "exec":
+            command = "exec_command"
 
         if not hasattr(self, command):
             raise NoSuchCommand(command, self)
--- a/docs/reference/exec.md
+++ b/docs/reference/exec.md
@@ -1,0 +1,29 @@
+<!--[metadata]>
++++
+title = "exec"
+description = "exec"
+keywords = ["fig, composition, compose, docker, orchestration, cli,  exec"]
+[menu.main]
+identifier="exec.compose"
+parent = "smn_compose_cli"
++++
+<![end-metadata]-->
+
+# exec
+
+```
+Usage: exec [options] SERVICE COMMAND [ARGS...]
+
+Options:
+-d                Detached mode: Run command in the background.
+--privileged      Give extended privileges to the process.
+--user USER       Run the command as this user.
+-T                Disable pseudo-tty allocation. By default `docker-compose exec`
+                  allocates a TTY.
+--index=index     index of the container if there are multiple
+                  instances of a service [default: 1]
+```
+
+This is equivalent of `docker exec`. With this subcommand you can run arbitrary
+commands in your services. Commands are by default allocating a TTY, so you can
+do e.g. `docker-compose exec web sh` to get an interactive prompt.
--- a/tests/acceptance/cli_test.py
+++ b/tests/acceptance/cli_test.py
@@ -752,6 +752,24 @@ class CLITestCase(DockerClientTestCase):
         self.project.stop(['simple'])
         wait_on_condition(ContainerCountCondition(self.project, 0))
 
+    def test_exec_without_tty(self):
+        self.base_dir = 'tests/fixtures/links-composefile'
+        self.dispatch(['up', '-d', 'console'])
+        self.assertEqual(len(self.project.containers()), 1)
+
+        stdout, stderr = self.dispatch(['exec', '-T', 'console', 'ls', '-1d', '/'])
+        self.assertEquals(stdout, "/\n")
+        self.assertEquals(stderr, "")
+
+    def test_exec_custom_user(self):
+        self.base_dir = 'tests/fixtures/links-composefile'
+        self.dispatch(['up', '-d', 'console'])
+        self.assertEqual(len(self.project.containers()), 1)
+
+        stdout, stderr = self.dispatch(['exec', '-T', '--user=operator', 'console', 'whoami'])
+        self.assertEquals(stdout, "operator\n")
+        self.assertEquals(stderr, "")
+
     def test_run_service_without_links(self):
         self.base_dir = 'tests/fixtures/links-composefile'
         self.dispatch(['run', 'console', '/bin/true'])


TomasTomecek:
@aanand @dnephin could I get some feedback here?

compose/service.py
```diff
@@ -308,6 +308,10 @@ def create_container(self,
 
         return Container.create(self.client, **container_options)
 
+    def exec_command(self, command, **kwargs):
+        for c in self.containers(filters={'status': 'running'}):
```
dnephin: Instead of the filter use `stopped=False` (which is the default).

compose/container.py
```diff
@@ -45,6 +45,14 @@ def create(cls, client, **options):
         response = client.create_container(**options)
         return cls.from_id(client, response['Id'])
 
+    @classmethod
+    def exec_command(cls, client, c_id, command, **options):
+        detach = options.get("detach")
```
dnephin: I think this is a problem. You can't really really attach to a service because it represents multiple containers.

dnephin:
I think there is still an issue with `exec` (similar to attach) in that it's fundamentally a container oriented command, not a service oriented one. The primary reason to use it is to get interactive access to a running container, but compose is concerned with the service abstraction, not individual containers.

`docker exec container_name` works just as well for interacting with a compose container.

So I'm not exactly sure how `docker-compose exec` should work.

TomasTomecek:
The reason I started working on this is

1. I wanted this feature, for me it's easier to type

    `docker-compose exec web bash`

    rather than

    `docker exec -ti i-have-no-idea-about-the container-name bash`

2. number of `+1`s in #593 

I agree that `exec` might not make sense if a service is composed of multiple containers. On the other hand, user may be interested in running the command in any of the containers (so this could be configurable).

TomasTomecek:
@dnephin any update?

TomasTomecek:
@aanand @bfirsh @dnephin once again, any update?

Just FYI, k8s support `exec`: https://cloud.google.com/container-engine/docs/kubectl/exec

aanand:
`docker-compose port` works on a single container (it has an optional `--index` parameter to specify which one, but that will probably become useless when we ditch sequential numbers).

I agree that the UX for doing an `exec` is not good, so a `docker-compose exec` which attaches to an arbitrary container sounds OK to me.

There was another PR/proposal for an `exec` which executed the specified command on *all* containers for the service, but I'm not convinced that's a good enough idea to warrant baking into Compose, and it's useless for interactive programs like bash.

This particular implementation looks like a good start, but I think it absolutely needs to support interactivity (via `dockerpty`) to be useful.

Thoughts?

dnephin:
The equivalent of that k8s exec command is the standard `docker exec`, you still need to specify the pod and the container.

TomasTomecek:
> This particular implementation looks like a good start, but I think it absolutely needs to support interactivity (via `dockerpty`) to be useful.

Absolutely. I was just waiting for you guys to get some feedback that this is the feature you would be interested in.

I'm planning to propose a patch to `dockerpty`. Without that support, I feel like that this feature is worthless. I'll get back to you once I have something to share.

> you still need to specify the pod and the container.

Looks like that container argument is optional, snippet from docs:

```
// get output from running 'date' from pod 123456-7890, using the first container by default
$ kubectl exec 123456-7890 date
```

TomasTomecek:
rebased and rewritten (added interactive support with updated dockerpty)

thaJeztah:
Looks like a test is failing, not sure if it's related

TomasTomecek:
@thaJeztah tests are untouched, am waiting for design review first

TomasTomecek:
@aanand @dnephin can we get this moving so it lands into next compose release?

compose/cli/main.py
```diff
@@ -243,6 +244,45 @@ def create(self, project, options):
...
+        """
+        Execute command in running service.
+
+        Usage: exec [options] SERVICE COMMAND [ARGS...]
+
+        Options:
+            -d                Detached mode: Run command in the background.
+            --privileged      Give extended privileges to the process.
+            --user USER       Run the command as this user.
+            -T                Disable pseudo-tty allocation. By default `docker-compose run`
```
dnephin: `s/run/exec/`

compose/service.py
```diff
@@ -249,6 +249,19 @@ def create_container(self,
...
+            container = self.get_container(number=kwargs["index"])
+        except KeyError:
+            container = self.get_container()
+        return Container.create_exec(self.client, container.id, command, **kwargs)
+
+    def exec_start(self, exec_id, **kwargs):
+        return Container.start_exec(self.client, exec_id, **kwargs)
+
+    def get_exec_exit_code(self, exec_id):
+        return self.client.exec_inspect(exec_id).get("ExitCode")
```
dnephin: Since these are container operations I think we should omit these 3 methods and just call the `Container` methods directly from the cli command. The `port` command is an example of this. It grabs the correct container first then operates directly on the container instead of the service.

TomasTomecek: This is the feedback I was looking for since I'm a bit loss in `compose`'s logic.

compose/cli/main.py
TomasTomecek: fixed (well spotted)

TomasTomecek:
rebased, updated

TomasTomecek:
those test failures seem unrelated to this PR

compose/container.py
```diff
@@ -46,6 +46,24 @@ def create(cls, client, **options):
         response = client.create_container(**options)
         return cls.from_id(client, response['Id'])
 
+    @classmethod
+    def create_exec(cls, client, c_id, command, **options):
```
dnephin: `client` is already available as `self.client` (same for the 2 methods below)

compose/container.py
```diff
@@ -46,6 +46,24 @@ def create(cls, client, **options):
...
+        privileged = options.get("privileged")
+        user = options.get("user")
+        tty = options.get("tty")
+        return client.exec_create(c_id, command, privileged=privileged, tty=tty, stdin=tty,
+                                  user=user)
```
dnephin: I think `c_id` here should come from `self.id`, so it can be removed from the method args as well.

dnephin:
Those test failures are unrelated, yes

TomasTomecek:
@dnephin thanks for pointers; converted class methods to methods

compose/cli/main.py
```diff
@@ -133,6 +134,7 @@ class TopLevelCommand(DocoptCommand):
       create             Create services
       down               Stop and remove containers, networks, images, and volumes
       events             Receive real time events from containers
+      exec               Execute command in running service
```
dnephin: I would probably say "Execute a command in a running container", to be clear that it's only run in a single container.

dnephin:
Looks like there is a flake8 error.  Code changes look good

TomasTomecek:
@dnephin thank you for the code review

Updated, added docs.

Regarding tests, wasn't sure which one should I write. I started with acceptance.

TomasTomecek:
regarding docs test failure: that's what happened on my machine, had to comment it out; @moxiegirl maybe?

TomasTomecek:
From my PoV, this is ready for final review and possible merge.

TomasTomecek:
fixed test failure (EOF in docs)

compose/container.py
```diff
@@ -212,6 +212,21 @@ def restart(self, **options):
...
+        privileged = options.get("privileged")
+        user = options.get("user")
+        tty = options.get("tty")
+        return self.client.exec_create(
+            self.id, command, privileged=privileged, tty=tty, stdin=tty, user=user)
+
+    def start_exec(self, exec_id, **options):
+        detach = options.get("detach")
+        tty = options.get("tty")
+        return self.client.exec_start(exec_id, detach=detach, tty=tty)
```
aanand: Small thing, but:  in these two methods, rather than extracting the kwargs we recognise and silently discarding anything else, why don't we just pass the whole dict off to `exec_create`/`exec_start`? That way if, a kwarg gets mistyped in future code changes, we find out about it immediately (because the docker-py method will raise an error).

compose/container.py
```diff
@@ -212,6 +212,21 @@ def restart(self, **options):
...
+
+    def get_exec_exit_code(self, exec_id):
+        return self.client.exec_inspect(exec_id).get("ExitCode")
```
aanand: Also a minor thing: this isn't a container-related method, so I feel like we could just delete it and move this line to `exec_command()` in `main.py`.

compose/container.py
TomasTomecek: Makes sense to me even though I feel like it's more error prone. But! With good integration tests it should be fine. Will refactor.

compose/container.py
TomasTomecek: can do

TomasTomecek:
updated per @aanand's suggestions

dnephin:
retest this please

TomasTomecek:
rebased, but it didn't help :/ now there's some java exception

SvenDowideit:
retest this please

TomasTomecek:
rebased, and... IT'S ALL GREEN! :green_apple: :cocktail:

compose/cli/main.py
```diff
@@ -296,6 +297,57 @@ def json_format_event(event):
...
+
+        try:
+            operation = ExecOperation(
+                    project.client,
+                    exec_id,
+                    interactive=tty,
+            )
+            pty = PseudoTerminal(project.client, operation)
+            pty.start()
+        except signals.ShutdownException:
```
dnephin: This is only raised if you register the signal handlers, but I don't see that happening here.

dnephin: I think even in the shutdown case you will want to use the exit code from the container, no?

dnephin:
sorry, noticed one more thing

TomasTomecek:
I'm really bad at copy-pasting. I'm sorry about that.

Should be fixed now.

To be honest, I'm unsure how that signal handing should suppose to work. When I ran compose like this
```shell
docker-compose exec asd sh
```
and then hit `^d` I expected the exception to be raised. That didn't happen.

TomasTomecek:
rebased b/c of the ^ unrelated test failure

dnephin:
LGTM

TomasTomecek:
rebased

aanand:
logo_exec_menor

LGTM

TomasTomecek:
\o/

This was a ride!

willejs:
:ok_hand:

darkn3rd:
Will this also have `exec:`? for `docker-compose.yml`?  I need this for a  wrapper script use case, in order to do setup before the process starts.

TomasTomecek:
@darkn3rd no, it's a cli command; feel free to open an issue with the feature request

thaJeztah:
@darkn3rd perhaps having an entrypoint script would be a solution for that? I think the "exec" would be executed *after* the image's main `CMD` is started, so probably not the best solution to perform these kind of actions.

TomasTomecek:
@thaJeztah to be honest I totally understand @darkn3rd's proposal: I've run into a similar issue: I have a web app with database started by compose. As soon as it's up, I want to populate database with some initial data and am not very keen on doing:

```
$ docker-compose up -d
$ docker-compose exec populate-database.sh
```

Would be awesome if compose had support for hooks.

thaJeztah:
@TomasTomecek well, the regular approach is to have the entrypoint script take care of that, for example, see the [Initializing a fresh instance](https://hub.docker.com/_/mysql/) section for the MySQL image (and [the script itself](https://github.com/docker-library/mysql/blob/d7e71447ff3ad3c579ac3cca8c0dc090d11d4baf/5.7/docker-entrypoint.sh#L96-L104))

Overall, `docker exec` is really designed for debugging purposes only. Also, initializing a database can be done through a "utility" container, that connects to the database-container to populate the database if needed

dnephin:
It would be even better to do it as part of `build` instead of in the entrypoint script. That way it is cached and the DB is ready to handle requests immediately, instead of waiting for initialization when it starts.

thaJeztah:
@dnephin yeah, problem is that the official images declare a `VOLUME`, so it's not possible to pre-populate databases if you're using the official images (unless you alter the configuration and use a different location I guess)

TomasTomecek:
@thaJeztah to me, custom entrypoint scripts are pure hack: initializing a service and running a service are two completely different operations which shouldn't be mixed in a single script -- when there is something wrong with such script, you need to debug it in order to have the service running in some form; on top of that, it's not clear what such script does (since usually it's several hundreds of LoC of cryptic shell code).

I like the idea with utility container, since it's clearly separates spinning database from its initialization.

@dnephin my issue with baking data inside database during build is that you need to *run* the database during build -- usually that's not a trivial thing. Not even saying that you have an image which can be only used within only one application in a single type of environment (dev/stage/prod).

mccrodp:
Hey all, great work getting this merged, hope I'm not hijacking...was looking to use this exact functionality using ````docker-compose````, rather than spawning a new container with run.

I see it in the docs now here: https://docs.docker.com/compose/reference/exec/

However, I get an error after running and don't see exec in the command list on the cli.

````No such command: exec````

I did an upgrade using ````docker-machine```` to v1.11.0 of boot2docker, so I presume I should be able to test this PR? Many thanks!

thaJeztah:
@mccrodp did you download the latest version of docker toolbox, or performed `docker-machine upgrade`? `docker-machine upgrade` only updates the VM to the latest version of boot2docker, but doesn't upgrade docker machine itself, not the other tools (such as docker compose). Does `docker-compose --version` show that you're running 1.7.0?

mccrodp:
@thaJeztah ooops, you were dead right! I had only ran the upgrade from the command line rather than the whole toolbox app (I was still on 1.6.2). Thanks for your help. 

All working now with ````docker-compose exec````. Awesome PR, thanks all involved!

PR #2026, opened 2015-09-10
Fix type for `tty` & `stdin_open`

ggtools:
`tty` & `stdin_open` are currently expected to be `string` while they should be `boolean`. This causes incompatibility with existing `docker-compose.yml`.


dnephin:
Thanks for the contribution!

LGTM

aanand:
LGTM

(Should we have tests for stuff like this?)

PR #2030, opened 2015-09-11
[WIP] Extend only abstract services

### LINKED ISSUE #1988: Extend a base service from the same file
It is a common pattern to define a base service which you want several services to extend, but it's currently very verbose to do this:

_docker-compose-base.yml_

```
base:
  environment:
    FOO: bar
```

_docker-compose.yml_

```
web:
  extends:
    file: docker-compose-base.yml
    service: base
  ...
db:
  extends:
    file: docker-compose-base.yml
    service: base
  ...
```

It should be possible to extend an incomplete service in a single compose file. Something like this:

```
base:
  abstract: true
  environment:
    FOO: bar
web:
  extends:
    service: base
  ...
db:
  extends:
    service: base
```

Design decisions:
- How should a service be marked as a service that shouldn't be run? (I've used "abstract" in the example above for the sake of illustration, pinched from Django's "abstract" models.)

mrfuxi:
Closes #1988

### DIFF
--- a/compose/config/config.py
+++ b/compose/config/config.py
@@ -152,7 +152,8 @@ def load(config_details):
             service_dict=service_dict)
         service_dict = loader.make_service_dict()
         validate_paths(service_dict)
-        service_dicts.append(service_dict)
+        if not service_dict.get('abstract', False):
+            service_dicts.append(service_dict)
 
     return service_dicts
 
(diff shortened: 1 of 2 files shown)


mrfuxi:
@bfirsh I went with abstract (I spend too much time with Django to think differently)

mnowster:
Dealing with abstract services isn't going to be that easy I'm afraid, it'll break validation in the current way validation is working.

I would pause doing anymore on this, it needs more discussion/agreement on what we'll do to deal with it and how validation will look/what's possible. I'm not convinced that we even need this functionality yet and don't want you to waste your time.

PR #2032, opened 2015-09-11
Flag to skip all pull errors when pulling images

### LINKED ISSUE #2028: docker-compose pull doesn't work with mixure of official and local images
I use my own image build process in my development environment. This process create images localy so I can test them. I also use mose official images (databases) in my `docker-compose.yml`. With this combination `docker-compose pull` command fail when its tries to pull my local only images and if I'm not lucky no database images are pulled.

I would like to from time to time pull new images from repository and so far I have to list them all. Is it possible to pass some flag to `pull`  that it tries to pull all images and only report those that failed (instead of stop pulling others)?

villlem:
Closes #2028

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -270,6 +270,7 @@ class TopLevelCommand(Command):
         Usage: pull [options] [SERVICE...]
 
         Options:
+            --ignore-pull-failures  Pull what it can and ignores images with pull failures.
             --allow-insecure-ssl    Deprecated - no effect.
         """
         if options['--allow-insecure-ssl']:
@@ -277,6 +278,7 @@ class TopLevelCommand(Command):
 
         project.pull(
             service_names=options['SERVICE'],
+            ignore_pull_failures=options.get('--ignore-pull-failures')
         )
 
     def rm(self, project, options):
--- a/compose/service.py
+++ b/compose/service.py
@@ -769,7 +769,7 @@ class Service(object):
                 return True
         return False
 
-    def pull(self):
+    def pull(self, ignore_pull_failures=False):
         if 'image' not in self.options:
             return
 
@@ -781,7 +781,14 @@ class Service(object):
             tag=tag,
             stream=True,
         )
-        stream_output(output, sys.stdout)
+
+        try:
+            stream_output(output, sys.stdout)
+        except StreamOutputError as e:
+            if not ignore_pull_failures:
+                raise
+            else:
+                log.error(six.text_type(e))
 
 
 class Net(object):
--- a/compose/project.py
+++ b/compose/project.py
@@ -311,9 +311,9 @@ class Project(object):
 
         return plans
 
-    def pull(self, service_names=None):
+    def pull(self, service_names=None, ignore_pull_failures=False):
         for service in self.get_services(service_names, include_deps=True):
-            service.pull()
+            service.pull(ignore_pull_failures)
 
     def containers(self, service_names=None, stopped=False, one_off=False):
         if service_names:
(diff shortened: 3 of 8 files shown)


compose/cli/main.py
```diff
@@ -275,6 +276,7 @@ def pull(self, project, options):
 
         project.pull(
             service_names=options['SERVICE'],
+            skip_missing=True if options.get('--skip-missing') else False
```
dnephin: You don't need an if/else here, just `skip_missing=options.get('--skip-missing')`

compose/service.py
```diff
@@ -768,7 +768,7 @@ def specifies_host_port(self):
                 return True
         return False
 
-    def pull(self):
+    def pull(self, skip_missing=True):
```
dnephin: This should default to False to be backwards compatible

compose/service.py
```diff
@@ -780,7 +780,13 @@ def pull(self):
...
             stream=True,
         )
-        stream_output(output, sys.stdout)
+        if not skip_missing:
+            stream_output(output, sys.stdout)
+        else:
+            try:
+                stream_output(output, sys.stdout)
+            except StreamOutputError as e:
+                log.error(six.text_type(e))
```
dnephin: I'd like to suggest a slightly different implementation:

```python

try:
    stream_output(output, sys.stdout)
except StreamOutputError as e:
    if not skip_missing:
        raise
    log.error(six.text_type(e))
    
```

villlem:
I've integrated @dnephin's comment. PTAL when you get chance.

dnephin:
LGTM, thanks!

compose/cli/main.py
```diff
@@ -268,6 +268,7 @@ def pull(self, project, options):
         Usage: pull [options] [SERVICE...]
 
         Options:
+            --skip-missing          Skip unpullable images.
```
dnephin: actually, I think we could be a little more informative here.  Maybe something like "Continue pulling images, even if one fails to pull".  

I'll let others comment on wording.

villlem: I like your wording. But the flag it self should be called `--skip-failed` or something like this.

mnowster: I agree with @dnephin here around the messaging. A suggestion; how about  `Continue pulling images, irrespective of failures`

The language we use here to describe this flag is really important to be clear. We don't know that images are missing. I would prefer `--continue-on-error`

mnowster: It'd also be to get @moxiegirl 's opinion on this wording, as I think it is a little tricky to convey clearly what we want to the user.

moxiegirl: Hmm, it seems like you aren't really skipping as much as ignoring.

`--ignore-pull-failures  Builds what it can and ignores images with pull failures.`

I also like `--continue-on-error` if it really continues everything regardless of error...but if they operation can error out it isn't really true.

mnowster: Thanks @moxiegirl , that is a very good point, I think your suggestion is clearer and more accurate.

@villlem are you ok to update the flag and message in line with @moxiegirl 's wording please?

villlem: Sounds good to me. I also update parameter names to be consistent. I there somethig else, some other doc that needs to be updated as well?

mnowster: There is the `pull` command docs file that'll need updating with the new option as well. I think that's it but I'll double check now.

https://github.com/docker/compose/blob/master/docs/reference/pull.md

Thank you for your work on this, it's appreciated.

villlem:
I've updated flags, docs and params names with @moxiegirl's wording. PTAL when you get chance. Thank you for all your comments. Let me know if there is something I should do.

mnowster:
LGTM. Thanks for your work @villlem

PR #2035, opened 2015-09-11
Fix schema.json MANIFEST.in entry

geron:
docker-compose up doesn't run after install without this.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -4,7 +4,7 @@ include requirements.txt
 include requirements-dev.txt
 include tox.ini
 include *.md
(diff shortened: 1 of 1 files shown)


dnephin:
Oh ya, we forgot to fix this.  Thanks!

Maybe we could do `include compose/config/*.json` so we can further change the name if we need to.

Just needs a commit signoff

geron:
Changed to *.json and signed off.

dnephin:
LGTM, that test failure will be fixed by #1955

aanand:
LGTM

PR #2040, opened 2015-09-14
include logo in readme and docs

### LINKED ISSUE #2024: create a logo for compose
Fig used to have a such a nice logo! Why doesn't compose have one?

fig

TomasTomecek:
Resolves #2024

Signed-off-by: Tomas Tomecek <ttomecek@redhat.com>


README.md
```diff
@@ -1,5 +1,7 @@
 Docker Compose
 ==============
+![Docker Compose](https://github.com/docker/compose/blob/master/docs/logo.png).
```
dnephin: I don't think this url will work, doesn't it need to be something like `https://raw.githubusercontent.com/docker/compose/master/docs/logo.png` ?

docs/index.md
```diff
@@ -8,9 +8,10 @@ parent="smn_workw_compose"
 +++
 <![end-metadata]-->
 
-
 # Overview of Docker Compose
 
+<img src="/compose/logo.png" width="200" height="200" />
```
dnephin: Can we use the markdown syntax here as well?

dnephin:
Thanks for the contribution!

README.md
thaJeztah: In Swarm they used this; https://github.com/docker/swarm/commit/accf600c327eeff416fb23d4ca12ad12b6ea89c5

thaJeztah: ^^ which, I think, would have the advantage that it also works on a Forked repo :-)

dnephin: Oh ya,  +1 , that is nice

thaJeztah: I don't know if the .png needs to be in the same path for this to work, but it's worth trying

docs/index.md
TomasTomecek: My point here was to resize the image b/c it's too big.

TomasTomecek:
So I changed it to mimic what swarm has.

dnephin:
I think we need to resize the image, it's too big for the README

thaJeztah:
Looks like there are [variations of Markdown that support image dimensions](http://stackoverflow.com/questions/14675913/how-to-change-image-size-markdown), but unfortunately, [not on GitHub](http://stackoverflow.com/questions/22485796/markdown-smaller-images-sizes-not-supported-by-github) (unless they implemented it in the mean time).

I suggest to add a resized copy to the root of the repo, and preserve the large size for the docs? Is the 200x200px size ok?

@TomasTomecek I can create a resized and optimized version of the png for you if you need, comment here if you want me to do that (might be tomorrow, as I'm in the EU timezone, and it's quite late now :smile:)

TomasTomecek:
done, resized to 200px (looks pretty small to be honest)

thaJeztah:
thanks! LGTM :-)

dnephin:
LGTM

mnowster:
LGTM

PR #1494, opened 2015-06-01
Add docker build --pull command.

glogiotatidis:
This waits for https://github.com/docker/docker-py/pull/624

PR #2041, opened 2015-09-14
Add new --pull option in build.

ggtools:
The `--pull` option from `docker build` is not available on compose. Add it.

### DIFF
--- a/tests/integration/cli_test.py
+++ b/tests/integration/cli_test.py
@@ -98,20 +98,56 @@ class CLITestCase(DockerClientTestCase):
             'sha256:38a203e1986cf79639cfb9b2e1d6e773de84002feea2d4eb006b52004ee8502d)...')
 
     @mock.patch('sys.stdout', new_callable=StringIO)
+    def test_build_plain(self, mock_stdout):
+        self.command.base_dir = 'tests/fixtures/simple-dockerfile'
+        self.command.dispatch(['build', 'simple'], None)
+
+        mock_stdout.truncate(0)
+        cache_indicator = 'Using cache'
+        pull_indicator = 'Status: Image is up to date for busybox:latest'
+        self.command.dispatch(['build', 'simple'], None)
+        output = mock_stdout.getvalue()
+        self.assertIn(cache_indicator, output)
+        self.assertNotIn(pull_indicator, output)
+
+    @mock.patch('sys.stdout', new_callable=StringIO)
     def test_build_no_cache(self, mock_stdout):
         self.command.base_dir = 'tests/fixtures/simple-dockerfile'
         self.command.dispatch(['build', 'simple'], None)
 
         mock_stdout.truncate(0)
         cache_indicator = 'Using cache'
-        self.command.dispatch(['build', 'simple'], None)
-        output = mock_stdout.getvalue()
-        self.assertIn(cache_indicator, output)
-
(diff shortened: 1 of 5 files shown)


tests/integration/cli_test.py
```diff
@@ -104,14 +104,22 @@ def test_build_no_cache(self, mock_stdout):
...
         mock_stdout.truncate(0)
         self.command.dispatch(['build', '--no-cache', 'simple'], None)
         output = mock_stdout.getvalue()
         self.assertNotIn(cache_indicator, output)
+        self.assertNotIn(pull_indicator, output)
+
+        mock_stdout.truncate(0)
+        self.command.dispatch(['build', '--pull', 'simple'], None)
+        output = mock_stdout.getvalue()
+        self.assertIn(pull_indicator, output)
```
dnephin: This should be split into a separate test case `def test_build_with_pull(self)`

dnephin:
Thanks for the contribution!  This is actually the 3rd(!) attempt to get this in, the previous ones were abandoned by the submitter, so I hope we can get this merged. We should be able to get it in for the next release.

I'll sort out the janky issue, just needs one small change to the test.

This will replace #1494

ggtools:
Build didn't pass but it's quite strange the pull tests are passing and the build ends with: `Build timed out (after 45 minutes). Marking the build as aborted.`

dnephin:
Don't worry about the build failures, we're having issues with jenkins right now.

I'll kick off another run when it's working again.

sdurrheimer:
I will add this to the zsh completion after merge.

@albers Incoming `docker-compose build --pull` option.

albers:
And I'll add it to bash completion.
Thanks @sdurrheimer for the ping.

dnephin:
The build is green, but I think we should split `test_build_with_pull` into two functions.  @ggtools would you mind making that change?

ggtools:
Actually if I'm splitting `test_build_with_pull` I might also split the other method and have 4 methods covering all possible configurations of `--no-cache` and `--pull`. What do you think of it?

dnephin:
That sounds great, thanks!

tests/integration/cli_test.py
```diff
@@ -98,20 +98,56 @@ def test_pull_with_digest(self, mock_logging):
...
 
         mock_stdout.truncate(0)
         cache_indicator = 'Using cache'
+        pull_indicator = 'Status: Image is up to date for busybox:latest'
         self.command.dispatch(['build', 'simple'], None)
         output = mock_stdout.getvalue()
         self.assertIn(cache_indicator, output)
+        self.assertNotIn(pull_indicator, output)
+
+    @mock.patch('sys.stdout', new_callable=StringIO)
```
dnephin: As an aside, I think it's a mistake we do this patch as a decorator.  It makes it harder to debug tests with `pdb`.  I've been moving some of them to be context managers around just the single line that needs the patch.

Using the `with mock.patch()` around just the line that needs it also removes the need to do `mock_stdout.truncate(0)`, since it would only wrap the stdout we care about.


Since this is pretty common in our test suite, I wouldn't ask you to change it as part of this PR, I think we can address that in a separate cleanup branch and change them all at once.

dnephin:
LGTM

aanand:
LGTM

Issue #1532, opened 2015-06-09
Support --user option in "docker-compose up"


mrzechonek:
I need to pass `--user` option to run orchestrated containers under my own UID. This is mostly because of mounted host volumes, and I'd like the dockerized app to generate files owned by me, not the root.

With `docker-compose up`, this is not possible, at least not directly. Right now I'm using a crazy workaround:

```bash
NAME=`compose run -d --user="$UID" someservicename`
docker rename $NAME ${NAME/_run/}
```

which is suboptiomal, to be gentle.

dnephin:
I'm not sure I understand, `docker-compose run --user` is an option, and the `docker-compose.yml` supports the `user` key (http://docs.docker.com/compose/yml/#working95dir-entrypoint-user-hostname-domainname-mem95limit-privileged-restart-stdin95open-tty-cpu95shares).

I guess we'd need to support environment variable resolution in the `user` field so that you can set it to `$UID` is that right? #1377

mrzechonek:
Hi. Yeah, $UID expansion would do the trick in this case. Still, --user option for 'up' command (not only for 'run') would be useful to start the whole project as particular user.

The point is to allow specifying the user without touching yml.

aanand:
> start the whole project as particular user

It never occurred to me that people might want to do this. Could you elaborate on the use case?

aanand:
Never mind, I scrolled up. Am I right in thinking you're on Linux? When using boot2docker it creates files with sensible permissions.

I wonder if there's a way to configure the Docker daemon to do this for every volume it mounts.

mrzechonek:
Yes, on Linux.

mrzechonek:
I know I could use a data only container and inspect/copy files from it, but a host volume is much more convenient and requires less scripting.

josephpage:
`up --user` could be dangerous if some containers already specify a user.

@mrzechonek I don't think that modifying a container because of its environment is a good practice.

mrzechonek:
@josephpage yeah, it might be. $UID expansion would do the trick though.

Pretty much the only thing I need is proper permissions on host volumes. I don't know of any other way to achieve that, besides running container process as $UID...

I would actually prefer this feature to be present in the Damon itself, like @aanand said. Or maybe an option to mount a volume in such a mode (a storage driver maybe?).

wildsurfer:
@mrzechonek, have you found a workaround?

mrzechonek:
[...]

PR #2042, opened 2015-09-14
Proposal: Run not one-off

dnephin:
(opening this PR to discuss, still needs tests)

There have been a couple occasions where it would have been useful to use `docker-compose run` to start a container, but have that container labeled and named as if it were a regular container (created by `up`), not a "one-off" container.  

Scenario 1 is using `docker-compose` with a process supervisor or init system. If you wanted to use a single `docker-compose.yml` for the entire system, but you wanted the supervisor to handle each service as a separate process, you could use `docker-compose run --no-deps --not-one-off` to start each container.  You could start them all at once (in parallel) , and as dependencies becomes available, the services would eventually stop failing and go into the `up` state.

Scenario 2 was more of a workaround for another issue, but in https://github.com/docker/compose/issues/1532#issuecomment-140015148 but may still be useful even after that issue is fixed.

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -351,6 +351,8 @@ class TopLevelCommand(DocoptCommand):
             -p, --publish=[]      Publish a container's port(s) to the host
             --service-ports       Run command with the service's ports enabled and mapped
                                   to the host.
+            --not-one-off         Run the container as part of the composition,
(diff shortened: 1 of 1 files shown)


dnephin:
@bukzor we talked about this a while ago re https://github.com/yelp/pgctl

PR #2045, opened 2015-09-14
Clean before doing a build

dnephin:
While building the linux binary for the 1.4.1 release I ran into a bunch of failures. I finally tracked them down to the `./build` directory.

This change will force cleanup all cached files to make sure we don't accidentally include old files in the binaries.

### DIFF
--- a/script/build-linux
+++ b/script/build-linux
@@ -2,6 +2,8 @@
 
 set -ex
 
+./script/clean
+
 TAG="docker-compose"
 docker build -t "$TAG" .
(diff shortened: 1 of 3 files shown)


mnowster:
@dnephin could you copy and put that reason why we need to clean before doing the build in the commit message please? Else we'll lose that context info when the PR gets merged.

dnephin:
Sounds good, I updated the commit message

mnowster:
Cool. LGTM.

PR #2046, opened 2015-09-15
Allow env_file to format lines as `export FOO=bar`

simonista:
Take it or leave it, but I thought it might be cool to allow the env_file to be formatted like a file that you could source in bash, for mixed development environments (some docker, some native).  Inspired by https://github.com/bkeepers/dotenv, which allows this.

### DIFF
--- a/compose/config/config.py
+++ b/compose/config/config.py
@@ -409,6 +409,8 @@ def env_vars_from_file(filename):
     for line in open(filename, 'r'):
         line = line.strip()
         if line and not line.startswith('#'):
+            if line.startswith('export '):
+                line = line.replace('export ', '', 1)
             k, v = split_env(line)
             env[k] = v
     return env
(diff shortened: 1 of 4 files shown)


thaJeztah:
Problem is, that such file would be incompatible with docker itself, which just uses newline separated key/values pairs. I like the suggestion, but not sure this is something that should be supported

dnephin:
Thanks for the contribution!

As @thaJeztah mentioned, we want to keep feature parity with what is supported by the docker engine, so we can't accept this PR at this time.

simonista:
Thanks, makes sense.  Do you think it would be worth taking this PR over to the main docker repo?

aanand:
You can already use the plain env file to populate your bash environment variables:

    export $(cat my-env-file)

I use this myself on personal projects and it works great.

PR #2047, opened 2015-09-15
Fixed HTML entity

michael-k:
See
- Broken entity: https://docs.docker.com/compose/install/#upgrading
- Entity for “EN DASH”: http://www.w3schools.com/charsets/ref_utf_punctuation.asp

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -69,7 +69,7 @@ to preserve) you can migrate them with the following command:
 
     $ docker-compose migrate-to-labels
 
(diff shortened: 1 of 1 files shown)


mnowster:
Hi @michael-k, I think this PR needs a rebase, please can you do that so we can review? Thanks :+1:

michael-k:
The `&endash;` (instead of `&ndash;`) is no longer in the code.

PR #2051, opened 2015-09-15
Extend compose files by allowing multiple files

### LINKED ISSUE #1987: Extend an entire Compose file

dnephin:
Resolves #1987

Add support for  specifying multiple `-f` arguments to `docker-compose`. Multiple files are merged onto each other to allow users to customize the composition for different environments.

The default overrides file is `docker-compose.override.yml` which is applied over the default `docker-compose.yml` if it exists.

### DIFF
--- a/compose/config/config.py
+++ b/compose/config/config.py
@@ -104,6 +104,8 @@ def find(base_dir, filenames):
         filenames = [os.path.join(base_dir, f) for f in filenames]
     else:
         filenames = get_default_config_files(base_dir)
+
+    log.debug("Using configuration files: {}".format(",".join(filenames)))
     return ConfigDetails(
         os.path.dirname(filenames[0]),
         [ConfigFile(f, load_yaml(f)) for f in filenames])
--- a/compose/config/config.py
+++ b/compose/config/config.py
@@ -2,7 +2,6 @@ import logging
 import os
 import sys
 from collections import namedtuple
-from functools import reduce
 
 import six
 import yaml
@@ -89,9 +88,22 @@ PATH_START_CHARS = [
 log = logging.getLogger(__name__)
 
 
-ConfigDetails = namedtuple('ConfigDetails', 'working_dir configs')
+class ConfigDetails(namedtuple('_ConfigDetails', 'working_dir config_files')):
+    """
+    :param working_dir: the directory to use for relative paths in the config
+    :type  working_dir: string
+    :param config_files: list of configuration files to load
+    :type  config_files: list of :class:`ConfigFile`
+     """
 
-ConfigFile = namedtuple('ConfigFile', 'filename config')
+
+class ConfigFile(namedtuple('_ConfigFile', 'filename config')):
+    """
+    :param filename: filename of the config file
+    :type  filename: string
+    :param config: contents of the config file
+    :type  config: :class:`dict`
+    """
 
 
 def find(base_dir, filenames):
@@ -170,10 +182,19 @@ def pre_process_config(config):
 
 
 def load(config_details):
-    working_dir, configs = config_details
+    """Load the configuration from a working directory and a list of
+    configuration files.  Files are loaded in order, and merged on top
+    of each other to create the final configuration.
+
+    Return a fully interpolated, extended and validated configuration.
+    """
 
     def build_service(filename, service_name, service_dict):
-        loader = ServiceLoader(working_dir, filename, service_name, service_dict)
+        loader = ServiceLoader(
+            config_details.working_dir,
+            filename,
+            service_name,
+            service_dict)
         service_dict = loader.make_service_dict()
         validate_paths(service_dict)
         return service_dict
@@ -187,21 +208,19 @@ def load(config_details):
         ]
 
     def merge_services(base, override):
+        all_service_names = set(base) | set(override)
         return {
             name: merge_service_dicts(base.get(name, {}), override.get(name, {}))
-            for name in set(base) | set(override)
+            for name in all_service_names
         }
 
-    def combine_configs(base, override):
-        service_dicts = load_file(base.filename, base.config)
-        if not override:
-            return service_dicts
+    config_file = config_details.config_files[0]
+    for next_file in config_details.config_files[1:]:
+        config_file = ConfigFile(
+            config_file.filename,
+            merge_services(config_file.config, next_file.config))
 
-        return ConfigFile(
-            override.filename,
-            merge_services(base.config, override.config))
-
-    return reduce(combine_configs, configs + [None])
+    return load_file(config_file.filename, config_file.config)
 
 
 class ServiceLoader(object):
--- a/compose/config/config.py
+++ b/compose/config/config.py
@@ -17,7 +17,6 @@ from .validation import validate_extended_service_exists
 from .validation import validate_extends_file_path
 from .validation import validate_service_names
 from .validation import validate_top_level_object
-from compose.cli.utils import find_candidates_in_parent_dirs
 
 
 DOCKER_CONFIG_KEYS = [
@@ -103,13 +102,13 @@ def find(base_dir, filenames):
     if filenames:
         filenames = [os.path.join(base_dir, f) for f in filenames]
     else:
-        filenames = [get_config_path(base_dir)]
+        filenames = get_default_config_path(base_dir)
     return ConfigDetails(
         os.path.dirname(filenames[0]),
         [ConfigFile(f, load_yaml(f)) for f in filenames])
 
 
-def get_config_path(base_dir):
+def get_default_config_path(base_dir):
     (candidates, path) = find_candidates_in_parent_dirs(SUPPORTED_FILENAMES, base_dir)
 
     if len(candidates) == 0:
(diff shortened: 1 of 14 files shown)


dnephin:
Ok, I've added some basic docs to the reference, cc @moxiegirl for review


I think we'll probably want to add an example which uses this new default override file, but I hope we can work on that as a separate branch.

compose/config/config.py
```diff
@@ -122,7 +129,31 @@ def get_config_path(base_dir):
         log.warn("%s is deprecated and will not be supported in future. "
                  "Please rename your config file to docker-compose.yml\n" % winner)
 
-    return os.path.join(path, winner)
+    return [os.path.join(path, winner)] + get_default_override_file(path)
+
+
+def get_default_override_file(path):
+    override_filename = os.path.join(path, DEFAULT_OVERRIDE_FILENAME)
+    return [override_filename] if os.path.exists(override_filename) else []
```
moxiegirl: Do you want to `log.info` if you find an override file?  You might be doing this somewhere and I missed it.

dnephin: We don't currently log anything about which file is picked if we hit the defaults. A `log.debug()` might be good for debugging.

moxiegirl: Works for me. Troubleshooting is the exact use case I was thinking of. :-D

docs/reference/docker-compose.md
```diff
@@ -41,6 +41,7 @@ Commands:
   unpause            Unpause services
   up                 Create and start containers
   migrate-to-labels  Recreate containers to add labels
+  version            Show the Docker-Compose version information
```
moxiegirl: @dnephin After discussion, I think we need to clarify and illustrate more as a way of teaching.

--->




# docker-compose Command

```
Usage:
  docker-compose [-f=<arg>...] [-p NAME] [-v] [COMMAND] [ARGS...]
  docker-compose -h|--help

Options:
  -v, --version             Print version and exit

Commands:
  build              Build or rebuild services
  help               Get help on a command
  kill               Kill containers
  logs               View output from containers
  pause              Pause services
  port               Print the public port for a port binding
  ps                 List containers
  pull               Pulls service images
  restart            Restart services
  rm                 Remove stopped containers
  run                Run a one-off command
  scale              Set number of containers for a service
  migrate-to-labels  Recreate containers to add labels
```

The Docker Compose binary. You use this command to build and manage multiple services in Docker containers.

Use the `-f` flag to specify the location of a Compose configuration file. You can supply multiple `-f` configuration files. When you supply multiple files, Compose combines them into a single configuration. Compose builds the configuration in the order you supply the files. Subsequent files files override and add to their successors.

For example, consider this command line:

```bash
$ docker-compose -f docker-compose.yml -f docker-compose.admin.yml run backup_db`
```

The `docker-compose.yml` file might specify a `webapp` service.

```
webapp:
  build: .
  ports:
    - "8000:8000"
  volumes:
    - "/data"
```

If the `docker-compose.admin.yml` also specifies this same service, any matching fields are override the previous file. New values, add to the `webapp` service configuration.

```
webapp:
  build: /var/specification
  environment:
    - DEBUG=1
```

Use a `-f` with `-` (dash) as the filename to read the configuration from
stdin. When stdin is used all paths in the configuration are
relative to the current working directory.

The `-f` flag is optional. If you don't provide this flag on the command line,
Compose traverses the working directory and its subdirectories looking for a
`docker-compose.yml` and a `docker-compose.override.yml` file. You must supply
at least the `docker-compose.yml` file. If both files are present, Compose
combines the two files into a single configuration. The configuration in the
`docker-compose.override.yml` file is applied over and in addition to the values
in the `docker-compose.yml` file.

Each configuration has a project name. If you supply a `-p` flag, you can
specify a project name. If you don't specify the flag, Compose uses the current
directory name.

## Where to go next

* [CLI environment variables](overview.md)
* [Command line reference](index.md)

compose/config/config.py
```diff
@@ -76,6 +76,7 @@
     'fig.yaml',
 ]
 
+DEFAULT_OVERRIDE_FILENAME = 'docker-compose.override.yml'
```
mnowster: Other defaults we've put in `const.py`, does this belong in there with them as well?

compose/config/config.py
```diff
@@ -132,29 +163,42 @@ def pre_process_config(config):
...
+    def combine_configs(base, override):
+        service_dicts = load_file(base.filename, base.config)
+        if not override:
+            return service_dicts
+
+        return ConfigFile(
+            override.filename,
+            merge_services(base.config, override.config))
+
+    return reduce(combine_configs, configs + [None])
```
aanand: This took me a long time to wrap my head around. I think it could be a lot less smart.

- `reduce` always, always makes my head hurt. Here in particular, `combine_configs` is returning a completely different type in the intermediate vs. the terminal case (list of dicts vs. `ConfigFile`). Could we make it just a plain loop?

- I had to stare at the set union in `merge_services` for a bit. Could we assign it to a variable so its purpose is clear?

    ```python
    all_service_names = set(base) | set(override)
    return {
        ...
        for name in all_service_names
    }
    ```

- `configs` is a strange name - could we change it to `config_files`, since (if I understand it correctly) it's a list of `ConfigFile` objects? We should also do the same with `ConfigDetails.configs`.

    Also, we should probably document the attributes (including types) of both `ConfigDetails` and `ConfigFile`.

- The nested functions make `load` a difficult function to skim because of how much I have to keep in my head, and they happen in between two lines of code, making the flow difficult to follow. Could we either define them all at the very top, or move them out to the module level?

compose/config/config.py
dnephin: I think this belongs with the list of `SUPPORTED_FILENAMES` which is just above this in this module.

`const.py` is good for constants that are used in multiple places, but if a value is only used in a single module (like this one) I prefer to keep it together with the code that uses it.

compose/config/config.py
dnephin: > Also, we should probably document the attributes (including types) of both ConfigDetails and ConfigFile.

We don't really do this in a consistent way right now. I'd love to use [sphinx](http://sphinx-doc.org/) for all the api docs so the documentation is consistent and machine readable.

dnephin:
@aanand  I think I've addressed your comments

Docs are updated as well, and added debug logging for filenames

compose/config/config.py
```diff
@@ -133,29 +178,42 @@ def pre_process_config(config):
...
+        return [
+            build_service(filename, name, service_config)
+            for name, service_config in processed_config.items()
+        ]
+
+    def merge_services(base, override):
+        all_service_names = set(base) | set(override)
+        return {
+            name: merge_service_dicts(base.get(name, {}), override.get(name, {}))
+            for name in all_service_names
```
mnowster: I think this level of terseness effects the readability a bit too much and could do with expanding out, being a bit more verbose. However, I think that's something we can address in a separate PR and not block this one for now.

compose/config/config.py
```diff
@@ -133,29 +178,42 @@ def pre_process_config(config):
     Pre validation checks and processing of the config file to interpolate env
     vars returning a config dict ready to be tested against the schema.
     """
-    config = interpolate_environment_variables(config)
-    return config
+    return interpolate_environment_variables(config)
 
 
 def load(config_details):
-    config, working_dir, filename = config_details
```
mnowster: Given how much is happening in this function, could we have a doc string please?

mnowster:
Overall I think this is good and great to get in before code freeze and some concerns around increasing level of verbosity is non blocking, which can be addressed outside of this PR. :palm_tree: 

LGTM.

compose/config/config.py
dnephin: I find it hard to be verbose (at least in code!) and always aim for greater density.  Let me know how I can make this more readable.

compose/config/config.py
dnephin: Cool, docstring added

Vanuan:
Somehow I thought that `docker-compose.override.yml` would be applied even if I specified additional file through `-f`. Isn't that the case?

dnephin:
It is not the case.

The default is `-f docker-compose.yml -f docker-compose.override.yml` . If you specify any `-f` the default is not used.

Vanuan:
> The default is -f docker-compose.yml -f docker-compose.override.yml

If it were defaults, it would fail with "docker-compose.override.yml is not found"
Is there a flag to ignore the "file doesn't exist" error?

dnephin:
I left out that bit, but yes, the override is considered optional. There is no flag to define an optional config

greggwon:
I have a complex C++ application that needs to build with a specific set of tools and libraries. I want to perform the build inside of a container, with everything installed into system directories /usr/lib, /lib, /usr/local/bin, /usr/local/lib etc.  To do the build, I mount an "external: false" volume on /usr/local/src.  In the Dockerfile, I just install linux with the build tools, and then cd to /usr/local/src, and do "git clone" to download the app which has submodule dependencies for the right versions of things for each branch/tag version I have.  The build works.  But, I then need to have an saved container without the /usr/local/src mount (and some secrets for GitHub access) that I can then use to run the application on the target hardware/environment.  How can I startup the container again, without the volume mounted nor the secrets in view?

glours:
hey @greggwon 
You should use multi-stage builds for this, a stage to build your source and a production one which only contains your final binary. 
https://docs.docker.com/build/building/multi-stage/

For the secret and git credentials you can take a look to https://docs.docker.com/engine/reference/builder/#run---mounttypesecret and https://docs.docker.com/engine/reference/builder/#run---mounttypessh

PR #2054, opened 2015-09-15
Choose API version automatically based on server

rgbkrk:
See [auto in docker-py Client API](http://docker-py.readthedocs.org/en/latest/api/index.html?highlight=auto).

If the user doesn't specify the API version, this uses 'auto' mode from docker-py, which detects the server version for you.


dnephin:
I think this was replaced by #2215

rgbkrk:
Sure was, thanks.

PR #2065, opened 2015-09-17
Use dockerswarm/dind image for ci

### LINKED ISSUE #2049: `DOCKER_VERSIONS=all script/test` doesn't work any more
It relies on the `ALL_DOCKER_VERSIONS` variable being set when `script/test-versions` runs, which is no longer the case now that `script/test` invokes it directly rather than running it inside a container.

dnephin:
Replaces #1783
Fixes #2049

The last commit includes some cleanup of the `Dockerfile`. Instead of installing everything three times, we can install it just twice (once for each python version).


mnowster:
BOOM. LGTM.

Issue #2037, opened 2015-09-13
 The system cannot find the file specified: u'\\fields_schema.json


alexandrev:
First of all, I want to apologize because I don't know if this is a bug or a misconfiguration on my side, but I tried to look for this problem but I don't have any response back. 

I'm trying to use the docker-compose in Windows. I'm trying the pull request https://github.com/docker/compose/pull/1957 and I'm getting this error when I try to run an up command:

jsonschema.exceptions.RefResolutionError: <urlopen error [Error 2] The system cannot find the file specified: u'\\fields_schema.json'>

This is the full stacktrace I'm getting:

Traceback (most recent call last):
  File "<string>", line 3, in <module>
  File "D:\Projects\GitHub\compose\build\docker-compose\out00-PYZ.pyz\compose.cli.main", line 51, in main
  File "D:\Projects\GitHub\compose\build\docker-compose\out00-PYZ.pyz\compose.cli.docopt_command", line 23, in sys_dispa
tch
  File "D:\Projects\GitHub\compose\build\docker-compose\out00-PYZ.pyz\compose.cli.command", line 32, in dispatch
  File "D:\Projects\GitHub\compose\build\docker-compose\out00-PYZ.pyz\compose.cli.docopt_command", line 26, in dispatch
  File "D:\Projects\GitHub\compose\build\docker-compose\out00-PYZ.pyz\compose.cli.main", line 146, in perform_command
  File "D:\Projects\GitHub\compose\build\docker-compose\out00-PYZ.pyz\compose.cli.command", line 62, in perform_command
  File "D:\Projects\GitHub\compose\build\docker-compose\out00-PYZ.pyz\compose.cli.command", line 83, in get_project
  File "D:\Projects\GitHub\compose\build\docker-compose\out00-PYZ.pyz\compose.config.config", line 153, in load
  File "D:\Projects\GitHub\compose\build\docker-compose\out00-PYZ.pyz\compose.config.config", line 187, in make_service_
dict
  File "D:\Projects\GitHub\compose\build\docker-compose\out00-PYZ.pyz\compose.config.validation", line 224, in validate_
against_service_schema
  File "D:\Projects\GitHub\compose\build\docker-compose\out00-PYZ.pyz\compose.config.validation", line 237, in _validate
_against_schema
  File "D:\Projects\GitHub\compose\build\docker-compose\out00-PYZ.pyz\jsonschema.validators", line 98, in iter_errors
[...]

PR #2068, opened 2015-09-18
Ensure RefResolver works across operating systems

mnowster:
The way paths were being done previously to load the schema meant they weren't working on windows, which meant the RefResolver couldn't find the *schema.json files to load in for successful validation.

Fixes https://github.com/docker/compose/issues/2037


aanand:
resolved_stamp

PR #2069, opened 2015-09-18
Test against a list of versions generated from docker/docker tags

dnephin:
Instead of having to update the docker versions we test against, query the github api and test against the most recent releases for the last two `major.minor` releases.

### DIFF
--- a/script/versions.py
+++ b/script/versions.py
@@ -1,0 +1,139 @@
+#!/usr/bin/env python
+"""
+Query the github API for the git tags of a project, and return a list of
+version tags for recent releases, or the default release.
+
+The default release is the most recent non-RC version.
+
+Recent is a list of unqiue major.minor versions, where each is the most
+recent version in the series.
+
+For example, if the list of versions is:
+
+    1.8.0-rc2
+    1.8.0-rc1
+    1.7.1
+    1.7.0
+    1.7.0-rc1
+    1.6.2
+    1.6.1
+
+`default` would return `1.7.1` and
+`recent -n 3` would return `1.8.0-rc2 1.7.1 1.6.2`
+"""
+from __future__ import print_function
+
+import argparse
+import itertools
+import operator
+from collections import namedtuple
+
+import requests
+
+
+GITHUB_API = 'https://api.github.com/repos'
+
+
+class Version(namedtuple('_Version', 'major minor patch rc')):
+
+    @classmethod
+    def parse(cls, version):
+        version = version.lstrip('v')
+        version, _, rc = version.partition('-')
+        major, minor, patch = version.split('.', 3)
+        return cls(int(major), int(minor), int(patch), rc)
+
+    @property
+    def major_minor(self):
+        return self.major, self.minor
+
+    @property
+    def order(self):
+        """Return a representation that allows this object to be sorted
+        correctly with the default comparator.
+        """
+        # rc releases should appear before official releases
+        rc = (0, self.rc) if self.rc else (1, )
+        return (self.major, self.minor, self.patch) + rc
+
+    def __str__(self):
+        rc = '-{}'.format(self.rc) if self.rc else ''
+        return '.'.join(map(str, self[:3])) + rc
+
+
+def group_versions(versions):
+    """Group versions by `major.minor` releases.
+
+    Example:
+
+        >>> group_versions([
+                Version(1, 0, 0),
+                Version(2, 0, 0, 'rc1'),
+                Version(2, 0, 0),
+                Version(2, 1, 0),
+            ])
+
+        [
+            [Version(1, 0, 0)],
+            [Version(2, 0, 0), Version(2, 0, 0, 'rc1')],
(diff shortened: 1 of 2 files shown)


script/versions.py
```diff
@@ -0,0 +1,115 @@
...
+        major, minor, patch = version.split('.', 3)
+        return cls(int(major), int(minor), int(patch), rc)
+
+    @property
+    def major_minor(self):
+        return self.major, self.minor
+
+    @property
+    def order(self):
+        return self[:3] + (None, self.rc) if self.rc else ('', )
```
mnowster: Could you expand this out to not be on one line please? I'm not sure I've properly understood what's the intent and why you need to do this calculation. `rc` will already be an empty string if it didn't exist after the `parse` function right? Why do you want a `(None, self.rc)` returned. What piece of context have I missed.

dnephin: This could definitely use a docstring to explain it.

The issue is that "default ordering" for this object would put rc releases after the canonical release.  This `(None, self.rc)` forces any release that was an rc release to come before the canonical (non-rc) release because None is sorted to the front of the sequence before any other character.

mnowster: Ah right, I wasn't expecting that `None` would get sorted in front of any characters. 

A docstring would also be helpful, I would still expand it out. It's not bash, you can have more than 1 line of code :dancers: I feel having an if branch would help readability here.

script/versions.py
```diff
@@ -0,0 +1,115 @@
...
+
+    def __str__(self):
+        rc = '-{}'.format(self.rc) if self.rc else ''
+        return '.'.join(map(str, self[:3])) + rc
+
+
+def group_versions(versions):
+    return list(
+        list(releases)
```
mnowster: Are you using list here as a shorthand to ensure uniqueness? I would also expand this code out to be a little more verbose :)

dnephin: lists aren't unique!

No, I'm using lists so that I can index into the sequence. `itertools.groupby()` returns a generator which can't be indexed.

I really dislike adding unnecessary code. This is pretty succinct and uses clear variables names.  One of the great features of python is that you don't have to write overly verbose code to express your ideas. Every additional line of code is an opportunity for a bug to exist.  Fewer lines of code is fewer bugs!  I understand the need for clear variable names, and whitespace to separate ideas, but I think this function is better expressed in a functional way (as it is written) than expanding it out into multiple lines.

This function reads pretty well to me:

"List the releases for all versions grouped by the major_minor attribute".

mnowster: I made a stupid mistake, I was thinking of sets while reading this, thinking if we used sets then that would give you the unique versions. Anyway. Disregard that bit.

What I'm trying to express is that to me, expanding this out isn't unnecessary code. This is harder to read. It reads well to you because you wrote it. As someone reading it to grok the solution and what problem it is solving, I'm sharing with you that this approach is harder to read in that it takes longer to figure out what's happening here. 

Balancing the line between very succinct code and over verbose code, we want somewhere in the middle. The rest of the codebase is written aiming for that somewhere in the middle. We don't have lots of functions that are inline like this, so I think it's fair to ask for it to be expanded out and made to be as readable as everything else we have.

Maybe worth pinging this @shin- and @aanand to see if it is just me who finds this harder to read. I appreciate my personal preferences and past experiences with code like this, means I'm not as much of a fan of denser functional coding style. 

I'm not blocking the PR on this, I am expressing my opinion.

script/versions.py
aanand: Yeah this is a bit of a puzzle to me, especially as I don't know the operator precedence between `+` and `if` off the top of my head.

Furthermore, I'd feel better if we didn't require readers of the code to know that fact about sorting. Instead of `None` and `''`, how about 0 and 1?

```python
if self.rc:
    return (self.major, self.minor, self.patch, 0, self.rc)
else:
    return (self.major, self.minor, self.patch, 1)
```

I know it's verbose, but at least it's really clear what the return value will look like in both cases, so I don't have to construct a tuple in my head and consult the class definition when reading.

dnephin: Ok, I've pushed a new version of this that I think makes it a lot more readable. It removes the need to know about operator precedence and it's explicit about the parts being returned on the return line.

script/versions.py
aanand: I think an example of the return format in the docstring would go a long way.

```python
>>> group_versions([
        Version.parse("1.0.0"),
        Version.parse("1.0.1"),
        Version.parse("1.1.0"),
        Version.parse("1.2.0"),
    ])
[
    [Version.parse("1.0.0"), Version.parse("1.0.1")],
    [Version.parse("1.1.0")],
    [Version.parse("1.2.0")],
]
```

In fact, docstring examples on most of these functions would be good, 'cause it's not known to the reader what the JSON returned by GitHub looks like.

dnephin:
Added more docstrings with example responses and reference to the github api docs

aanand:
LGTM

PR #2071, opened 2015-09-18
Restore the dist volume mount for building linux binaries

dnephin:
I removed the `/code` volume in #2065 so that we used the code from the image, but didn't realize we still needed `/code/dist` for the binary.


tombee:
Thanks for this!

aanand:
LGTM

PR #2083, opened 2015-09-22
Fix a test case that assumes busybox image id

dnephin:
Using the old busybox image id was causing failures in the 1.4.2 release branch, and it probably fails on master as well for any host that has updated busybox (and removed the old layers).


mnowster:
LGTM

PR #2088, opened 2015-09-23
2050 relative volume paths windows

mnowster:
Make relative volume paths work on windows.

### DIFF
--- a/compose/config/config.py
+++ b/compose/config/config.py
@@ -16,7 +16,6 @@ from .validation import validate_extended_service_exists
 from .validation import validate_extends_file_path
 from .validation import validate_service_names
 from .validation import validate_top_level_object
-from compose.const import IS_WINDOWS_PLATFORM
 
 
 DOCKER_CONFIG_KEYS = [
@@ -542,19 +541,13 @@ def path_mappings_from_dict(d):
     return [join_path_mapping(v) for v in d.items()]
 
 
-def split_path_mapping(path):
-    if IS_WINDOWS_PLATFORM:
-        if len(path.split(":")) > 2:
-            parts = path.split(":")
-            host = ":".join(parts[0:2])
-            container = ":".join(parts[2:])
-            return (container, host)
+def split_path_mapping(volume_path):
+    drive, volume_config = os.path.splitdrive(volume_path)
+    if ':' in volume_config:
+        (host, container) = volume_config.split(':', 1)
+        return (container, drive + host)
     else:
-        if ':' in path:
-            host, container = path.split(':', 1)
-            return (container, host)
-
-    return (path, None)
+        return (volume_path, None)
 
 
 def join_path_mapping(pair):
--- a/compose/config/config.py
+++ b/compose/config/config.py
@@ -16,6 +16,7 @@ from .validation import validate_extended_service_exists
 from .validation import validate_extends_file_path
 from .validation import validate_service_names
 from .validation import validate_top_level_object
+from compose.const import IS_WINDOWS_PLATFORM
 
 
 DOCKER_CONFIG_KEYS = [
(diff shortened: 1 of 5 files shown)


compose/const.py
```diff
@@ -2,6 +2,7 @@
 import sys
 
 DEFAULT_TIMEOUT = 10
+IS_WINDOWS_PLATFORM = (sys.platform == "win32")
```
dnephin: this is now duplicated with line 12, we should remove one of them

mnowster: Dammit. Good catch, thanks, I missed that in the rebase.

compose/service.py
```diff
@@ -937,7 +938,19 @@ def build_volume_binding(volume_spec):
...
+    parts to be returned as a valid VolumeSpec tuple.
+    """
     parts = volume_config.split(':')
+
+    if IS_WINDOWS_PLATFORM:
+        # relative paths in windows expand to include the drive, eg C:\
+        # so we join the first 2 parts back together to count as one
+        windows_parts = [":".join(parts[0:2])]
+        windows_parts.extend(parts[2:])
+        parts = windows_parts
```
dnephin: Could we do this as part of the config module instead?  If we have config always return a proper `VolumeSpec()` we don't have to worry about it twice

compose/config/config.py
```diff
@@ -541,12 +542,19 @@ def path_mappings_from_dict(d):
...
+            host = ":".join(parts[0:2])
+            container = ":".join(parts[2:])
+            return (container, host)
     else:
-        return (string, None)
+        if ':' in path:
+            host, container = path.split(':', 1)
+            return (container, host)
+
+    return (path, None)
```
dnephin: This feels like it's likely to hit a bunch of edge cases.  Would this work ?

```python

drive, volume_config = os.path.splitdrive(volume_config)
if ':' in volume_config:
    (host, container) = string.split(':', 1)
    return (container, drive + host)
else:
    return (drive + volume_config, None)
```

compose/service.py
mnowster: I'd rather do that as a separate PR to this. This is making it work first without too big a change, then we can refactor and improve. I have a feeling there'll be some refactoring required with the other `volumes_from` PR as well.

compose/config/config.py
mnowster: Let's find out!

dnephin:
appveyor is now setup to run on PRs, so you should be able to just push this branch with a new commit and have it run

mnowster:
Ok, the other test failures on appveyor are separate to what's going on in here. So this is good for final review/merge.

tests/unit/config/config_test.py
```diff
@@ -1138,7 +1138,21 @@ def test_expand_path_with_tilde(self):
         self.assertEqual(result, user_path + 'otherdir/somefile')
 
 
-@pytest.mark.xfail(IS_WINDOWS_PLATFORM, reason='paths use slash')
```
dnephin: The test failures are because this decorator was removed from `BuildPathTest`.

I don't think we should merge with a red build. Please re-add this `xfail` so the build is green.

compose/config/config.py
```diff
@@ -541,12 +541,13 @@ def path_mappings_from_dict(d):
...
-        (host, container) = string.split(':', 1)
-        return (container, host)
+def split_path_mapping(volume_path):
+    drive, volume_config = os.path.splitdrive(volume_path)
+    if ':' in volume_config:
+        (host, container) = volume_config.split(':', 1)
+        return (container, drive + host)
     else:
-        return (string, None)
+        return (volume_path, None)
```
dnephin: We stripped the drive from `volume_path`, don't we need to re-add it?

I guess in this case it should always be the empty string, since the container path can't be a windows path?

This is really just a question, not a blocking issue.

compose/service.py
dnephin: That's fair. In that case can we use the same `splitdrive()` logic here as well?

tests/unit/config/config_test.py
mnowster: Dammit, I fucked up the rebase and didn't spot this.

I know we don't merge on red build, I had assumed it was something else.

compose/config/config.py
mnowster: `volume_path` doesn't get changed, it's the result of `splitdrive` that we use. So rather than concating the two together again, we can return the `volume_path` as it was passed in.

dnephin: Ahh, I misread sorry, you are right.  In my example I had re-used the variable, and I didn't read this properly.

mnowster:
I'm falling into a pile of yaks with a chainsaw.

PR #2113, opened 2015-09-30
Introduce new config format with volume support and versioning

shin-:
This is a work in progress (volume support is not implemented) submitted early so we can start discussing the design approach and new schema.

### DIFF
--- a/compose/config/config.py
+++ b/compose/config/config.py
@@ -165,6 +165,27 @@ def find_candidates_in_parent_dirs(filenames, path):
     return (candidates, path)
 
 
+def get_services(config, version):
+    if version == 2:
+        return config.get('services', {})
+    else:
+        return config
+
+
+@validate_top_level_object
+def get_config_version(config):
+    return config.get('version')
+
+
+@validate_top_level_object
+def pre_process_config(config):
+    """
+    Pre validation checks and processing of the config file to interpolate env
+    vars returning a config dict ready to be tested against the schema.
+    """
+    return interpolate_environment_variables(config, get_config_version(config))
+
+
 def load(config_details):
     """Load the configuration from a working directory and a list of
     configuration files.  Files are loaded in order, and merged on top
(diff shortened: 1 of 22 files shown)


compose/config/config.py
```diff
@@ -171,6 +171,18 @@ def find_candidates_in_parent_dirs(filenames, path):
...
+def get_services(config, version):
+    if version == 2:
+        return config.get('services', {})
+    else:
+        return config
+
+
+@validate_top_level_object
+def get_config_version(config):
+    return config.get('version')
```
mnowster: This doesn't need to be in a separate function. It's a dict lookup so can just be a line in `pre_process_config` function and then that way we're not doing `validate_top_level_object` twice.

compose/config/interpolation.py
```diff
@@ -8,13 +8,22 @@
 log = logging.getLogger(__name__)
 
 
-def interpolate_environment_variables(config):
+def interpolate_environment_variables(config, version):
     mapping = BlankDefaultDict(os.environ)
+    cfg = config.copy()
```
mnowster: I'm generally against abbreviated variable names because they are harder to read, in python we value verbosity because it gives readability to the code and clearer intent. Please use descriptive variable names.

compose/config/config.py
```diff
@@ -171,6 +171,18 @@ def find_candidates_in_parent_dirs(filenames, path):
     return (candidates, path)
 
 
+def get_services(config, version):
+    if version == 2:
+        return config.get('services', {})
+    else:
+        return config
```
dnephin: Right now this version check is done in 4 places!  This is going to make it much harder to maintain the code. We should check for the config version in only 1 place. This should also result in a smaller change to support the new version.

Since the variation between versions is at the top level of the config, it should be easy enough right now.  Something like this:

1. rename the current `load()` function to `load_services()`
2. create a new `namedtuple('Config', 'service version volumes')`
3. create a new `load()` function that works something like this:

    ```python

    def get_config_version(config_details):
        ... # stubbed out, but returns the version and asserts the version in all files is the same

    def load(config_details):
        version = get_config_version(config_details)
        if not version: # user v1
            return load_services(config_details)
        if version == 2:
            return load_v2(config_details)

        raise UnknownVersionErrror(...)
    ```

4. `load_services()` would return `Config(services, "1", [])`
5. the new `load_v2()` would call `load_services()` to handle the services part, and call another method to handle the volumes validation, and return a `Config` object

This way `load()` always returns the same structure, and the rest of the code doesn't have to know anything about config version.

compose/config/fields_schema_v2.json
```diff
@@ -0,0 +1,170 @@
...
+      "patternProperties": {
+        "^[a-zA-Z0-9._-]+$": {
+          "$ref": "#/definitions/volume"
+        }
+      }
+    }
+  },
+
+  "definitions": {
+    "service": {
```
dnephin: If this is the same definition from v1 we should definitely share it with a `$ref`. If it's not we should consider breaking it down so that we can share parts of it.

compose/config/fields_schema_v2.json
```diff
@@ -0,0 +1,170 @@
...
+        "memswap_limit": ["mem_limit"]
+      },
+      "additionalProperties": false
+    },
+
+    "volume": {
+      "type": "object",
+      "properties": {
+        "driver": {"type": "string"},
+        "driver_opts": {"type": "object", "additionalProperties": true}
```
dnephin: I think we should do something like what we do for `environment`, where we restrict the object to values of  `"type": ["string", "number", "boolean"]`. Does that apply here?

compose/config/fields_schema_v2.json
```diff
@@ -0,0 +1,170 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+
+  "type": "object",
+  "properties": {
+    "version": {
+      "enum": [2]
```
dnephin: Minor, but we'll probably want to do semver here, so maybe a string of "2.0" so the version type is consistent when we increment it

michaeljs1990:
Sorry, just so i understand correctly... Docker compose is going to be switching to using json now instead of yaml files? Just want to get ready for this change since i'll be upgrading as soon as the new version comes out.

dnephin:
@michaeljs1990 no we're not switching to json. We can use jsonschema with yaml configuration since they deserialize into the same python structure.

michaeljs1990:
@dnephin thanks, sorry for the confusion.

shin-:
Closing in favor of #2421 (because the code has undergone significant changes since, rewriting the PR had become easier than rebasing the changes).

PR #2114, opened 2015-09-30
Added support for cgroup_parent

mohitsoni:
This change adds cgroup-parent support to compose project. It allows
each service to specify a 'cgroup_parent' option.

Related:
1. docker/docker-py#716 (Enables this PR)
2. mohitsoni/compose-executor#3 (Depends on this PR)

### DIFF
--- a/tests/unit/service_test.py
+++ b/tests/unit/service_test.py
@@ -331,6 +331,13 @@ class ServiceTest(unittest.TestCase):
         service.create_container(do_build=False)
         self.assertFalse(self.mock_client.build.called)
 
+    def test_create_container_no_build_cgroup_parent(self):
+        service = Service('foo', client=self.mock_client, build='.')
+        service.image = lambda: {'Id': 'abc123'}
+
+        service.create_container(do_build=False, cgroup_parent='test')
+        self.assertFalse(self.mock_client.build.called)
+
     def test_create_container_no_build_but_needs_build(self):
         service = Service('foo', client=self.mock_client, build='.')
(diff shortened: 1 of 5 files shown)


mnowster:
Thanks for the contribution @mohitsoni 

The `cgroup-parent` option will need adding to the docs as well. https://github.com/docker/compose/blob/master/docs/yml.md.

If you do a `git rebase -i HEAD~2` then you can `fixup` the typo commit into the previous one.

tests/unit/service_test.py
```diff
@@ -331,6 +331,13 @@ def test_create_container_no_build(self):
         service.create_container(do_build=False)
         self.assertFalse(self.mock_client.build.called)
 
+    def test_create_container_no_build_cgroup_parent(self):
+        service = Service('foo', client=self.mock_client, build='.')
+        service.image = lambda: {'Id': 'abc123'}
+
+        service.create_container(do_build=False, cgroup_parent='test')
+        self.assertFalse(self.mock_client.build.called)
```
mnowster: I don't see how this test is testing that our `cgroup_parent` option was passed in correctly. I think this is the the wrong place to write the test for the option.

Maybe something along the lines to this test, https://github.com/docker/compose/blob/master/tests/unit/service_test.py#L111
where we can see that the `cgroup_parent`, if specified, is sent to create_host_config.

mnowster:
The test needs updating, other than that the rest is LGTM :+1:

mnowster:
Closing in favour of https://github.com/docker/compose/pull/2189

I picked up and did the test change, so we can get this in for 1.5

Issue #2092, opened 2015-09-24
Specify both image name and build path


bfirsh:
It should be possible to specify the name of an image, and the path it can be built from. Like so:

```
web:
  image: hyperglobalmegacorp/web
  build: .
```

This lets you do various things:

1. Bootstrap an environment by pulling pre-built master images from the Hub so you don't have to build them locally when setting up an initial environment (`docker-compose pull; docker-compose up`)
2. Lets you easily push images that you have built
3. Lets you customise the name of the built image
4. Makes it easier to use the same Compose file in development and production (production can pull images in advance or ignore `build`)

Related: #213

aanand:
Use cases 2, 3, and 4 are compelling. The first one makes me nervous - means that at any given time, `web`'s image could have come from either a registry or a local build. Still, it might not be an issue in practice.

@dnephin implemented a `tag` key in https://github.com/docker/compose/pull/457, which was eventually closed for possibly-obsolete reasons:

> After using this branch for a bit, I realized that doing the tagging implicitly during `build` is probably not ideal. On a shared host which builds and tests many images (jenkins), you probably don't want to tag things until after the tests have completed.

As discussed in the following comments, this could be worked around by adding a git SHA to the tag, which is possible now that we have environment variable expansion:

```yaml
web:
  image: hyperglobalmegacorp/web:$SHA
  build: .
```

bfirsh:
Many users have specifically asked for this to fix use case 1. Their builds take a long time and they want to on board users quickly. They get round it by having shell scripts to pull and tag the images.

I've heard fewer people ask for 2 and 3, and nobody's asked for 4.

aanand:
To be honest I'm not that worried about it, and it's a situation with pretty clear semantics - after running `docker pull SERVICE` or `docker build SERVICE`, you can always be sure of where the current image came from. So I think it's fine.

dnephin:
This issue was originally raised quite a while ago (#627)

I'm +1 on being able to specify both

At a previous gig I worked on a tool to handle the problem in (1). Unfortunately we didn't get a chance to open-source it, but it was basically what you describe (pull before a build). That way you use `pull` as a way to replicate the build cache.

> 1) Lets you pull pre-built images from the Hub so you don't have to build them locally (docker-compose pull; docker-compose up). Their builds take a long time and they want to on board users quickly. They get round it by having shell scripts to pull and tag the images.

[...]

PR #2118, opened 2015-09-30
added -v (version) support to build and up commands

goloveychuk:
Use case: you are on active developing and should frequently change branch (version). 
This commit will give an ability to put a version while build or up. 
fig build -v dev www - will build image with name "project_www:dev"
fig up -v dev www - will run container for image with name "project_www:dev" (by default - latest)
Don't know if it's frequent use case, your choice.

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -158,10 +158,12 @@ class TopLevelCommand(Command):
         Options:
             --no-cache  Do not use cache when building the image.
             --pull      Always attempt to pull a newer version of the image.
+            -v VERSION  Put a version, latest by default
         """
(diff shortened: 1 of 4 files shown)


dnephin:
I think we would want to allow for a customized name before we support separate tags for image names. See #2092

If we allow customized names, you'd be able to do this with just a variable in the conrfig, which I think I prefer over a command line flag.

dnephin:
Thanks for the PR! Unfortunately I don't think we'll be able to accept this change. We'd like to support this in a more general way (not just by injecting a version tag).

The plan for supporting this is outlined in #2092, which will allow for a variable to set the image name or tag.

Issue #2050, opened 2015-09-15
Problem with relative paths in volumes in Windows


alexandrev:
I'm using the docker-compose version 1.5.0 with the pull-request https://github.com/docker/compose/pull/1957. When I'm trying to start a group of containers which one of them has a relative path in the volume section like this:

 volumes:
  - "./data:/var/run/docker.sock"

I'm getting this error:

b'invalid mode for volumes-from: /var/run/docker.sock'

dnephin:
This error suggests it's using `volumes_from` not `volumes`, could you post your full `docker-compose.yml` ?

alexandrev:
Of course. I've been trying different posiblities. This is the full one right now.

centosjava:
 build: base
bw6agent:
 build: bw6
 links:
  - "ambassador:tea"
 ports:
  - "9091:9091"
  - "5050:5050"
  - "9090:9090"
  - "8079:8079"
 hostname: bw6agent
tea:
 build: tea
 links:
  - "ambassador:bw6agent"
 ports:
  - "8777:8777"
 hostname: tea
ambassador:
 image: cpuguy83/docker-grand-ambassador
 volumes:
  - "./data/docker.sock:/var/run/docker.sock"
 command: "-name tibcobw6_bw6agent_1 -name tibcobw6_tea_1"

christianbundy:
@alexandrev You can use [fenced code blocks](https://help.github.com/articles/github-flavored-markdown/#fenced-code-blocks) with `yaml` [syntax highlighting](https://help.github.com/articles/github-flavored-markdown/#syntax-highlighting) to format your code.

```yaml
centosjava:
  build: base
bw6agent:
  build: bw6
  links:
    - "ambassador:tea"
  ports:
    - "9091:9091"
    - "5050:5050"
    - "9090:9090"
    - "8079:8079"
  hostname: bw6agent
tea:
  build: tea
  links:
    - "ambassador:bw6agent"
  ports:
    - "8777:8777"
  hostname: tea
ambassador:
  image: cpuguy83/docker-grand-ambassador
  volumes:
    - "./data/docker.sock:/var/run/docker.sock"
  command: "-name tibcobw6_bw6agent_1 -name tibcobw6_tea_1"
```

alexandrev:
Thanks a lot @christianbundy and sorry for don't do it this way at first.

msamblanet:
I am seeing a similar issue on a ```docker-compose up -d```

```yaml
connect:
[...]

PR #2122, opened 2015-10-01
2050 ensure windows volume paths are compatible with engine

mnowster:
This fixes absolute windows paths having drives which caused the split on `:` logic to break. It also ensures the path is compatible with engine, which requires a linux path, including the drive.

Fixes https://github.com/docker/compose/issues/2050

### DIFF
--- a/compose/service.py
+++ b/compose/service.py
@@ -20,6 +20,7 @@ from .config import DOCKER_CONFIG_KEYS
 from .config import merge_environment
 from .config.validation import VALID_NAME_CHARS
 from .const import DEFAULT_TIMEOUT
+from .const import IS_WINDOWS_PLATFORM
 from .const import LABEL_CONFIG_HASH
 from .const import LABEL_CONTAINER_NUMBER
 from .const import LABEL_ONE_OFF
@@ -937,7 +938,20 @@ def build_volume_binding(volume_spec):
 
 
 def parse_volume_spec(volume_config):
+    """
+    A volume_config string, which is a path, split it into external:internal[:mode]
+    parts to be returned as a valid VolumeSpec tuple.
+    """
     parts = volume_config.split(':')
+
+    if IS_WINDOWS_PLATFORM:
+        # relative paths in windows expand to include the drive, eg C:\
+        # so we join the first 2 parts back together to count as one
+        drive, volume_path = os.path.splitdrive(volume_config)
+        windows_parts = volume_path.split(":")
+        windows_parts[0] = os.path.join(drive, windows_parts[0])
+        parts = windows_parts
+
     if len(parts) > 3:
         raise ConfigError("Volume %s has incorrect format, should be "
                           "external:internal[:mode]" % volume_config)
--- a/compose/service.py
+++ b/compose/service.py
@@ -937,33 +937,54 @@ def build_volume_binding(volume_spec):
     return volume_spec.internal, "{}:{}:{}".format(*volume_spec)
 
 
+def normalize_paths_for_engine(external_path, internal_path):
+    """
+    Windows paths, c:\my\path\shiny, need to be changed to be compatible with
+    the Engine. Volume paths are expected to be linux style /c/my/path/shiny/
+    """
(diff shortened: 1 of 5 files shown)


compose/service.py
```diff
@@ -936,20 +937,51 @@ def build_volume_binding(volume_spec):
...
     parts = volume_config.split(':')
+
+    if IS_WINDOWS_PLATFORM:
+        # relative paths in windows expand to include the drive, eg C:\
+        # so we join the first 2 parts back together to count as one
+        drive, volume_path = os.path.splitdrive(volume_config)
+        windows_parts = volume_path.split(":")
+
+        if drive:
+            windows_parts[0] = "/{}{}".format(drive.replace(':', '/'), windows_parts[0])
```
aanand: Reading `ensure_engine_compatibility`'s docstring, I sort of expected this bit of logic to happen over there. Would it be simple enough to move it there?

mnowster: Yep, good point. I'll move it.

compose/service.py
```diff
@@ -936,20 +937,55 @@ def build_volume_binding(volume_spec):
     return volume_spec.internal, "{}:{}:{}".format(*volume_spec)
 
 
+def ensure_engine_compatibility(external_path, internal_path):
```
dnephin: Minor naming nit: this function name seems a lot more general then what it's doing. "ensure" to me means "validate", and "engine compatibility" is a much larger issue than just paths.

How about `normalize_host_path()`  (or something along those lines) ?

compose/service.py
```diff
@@ -936,20 +937,55 @@ def build_volume_binding(volume_spec):
...
     else:
-        external = os.path.normpath(parts[0])
-        internal = os.path.normpath(parts[1])
+        external, internal = ensure_engine_compatibility(os.path.normpath(parts[0]), os.path.normpath(parts[1]))
 
-    mode = parts[2] if len(parts) == 3 else 'rw'
+    try:
+        mode = parts[2]
+    except IndexError:
+        mode = 'rw'
```
dnephin: Why this change from if/else to try/except? If we expected the common case to be that 'rw' was specified, than I think this would be faster, but I think the more common case is actually no mode, which makes the if/else much faster.

compose/service.py
mnowster: This was a tough one to name, I was nearly going to invent a word :crystal_ball: but I chatted with @aanand about it as well. I originally was going to use `transform_for_engine` or something similar so that it is very explicit that we are changing the way these paths look for the specific need of the engine.
`normalize_host_path` doesn't capture the specific need of we are doing this for the engine. I understand the point that there are larger issues than paths, could be `ensure_paths_are_engine_compatible` is clearer?

compose/service.py
mnowster: Aah ok, I thought the common case was it would be specified, if it's the other way around then yeh, agree, I'll change it back to an if.

compose/service.py
dnephin: Does it matter why we're doing it? I usually keep naming to "what" it's doing, and leave the "why" for the docstring.

To me this is a "normalize" operation not "ensure". "ensure" is validation/assertions, this is "do something in some cases, but ignore it in others".

`normalize_paths_for_engine()` would work for me (even though "engine" might not always be relevant, it's more that we're normalizing them for posix, so my first choice would probably be `normalize_paths_to_posix()` or something like that).

mnowster: That naming, was me trying to describe "what" we're doing and part of that what included the "why". When naming things, especially tricky things like this, I try to think about the person who has to come after me and fix it or modify the code, so I lean towards verbosity and whys as that will aid them to understand why the code is the way it is and pick it up quicker than it took me to figure it out the first time.

Ultimately we're agreeing with each other but with different words :)

I'm happier with `normalize_paths_for_engine` for now, as this is a workaround a hack within the engine itself. Then when that is no longer the case, we can rename it again. Cool.

aanand:
LGTM

PR #2124, opened 2015-10-01
Don't expand volume names

aanand:
Only expand volume host paths if they begin with a dot.

This is a breaking change. The deprecation warning preparing users for this change has been removed.

### DIFF
--- a/compose/config/config.py
+++ b/compose/config/config.py
@@ -78,13 +78,6 @@ SUPPORTED_FILENAMES = [
 
 DEFAULT_OVERRIDE_FILENAME = 'docker-compose.override.yml'
 
-PATH_START_CHARS = [
-    '/',
-    '.',
-    '~',
-]
-
-
 log = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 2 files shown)


compose/config/config.py
```diff
@@ -495,18 +488,10 @@ def resolve_volume_path(volume, working_dir, service_name):
...
-            log.warn(
-                'Warning: the mapping "{0}:{1}" in the volumes config for '
-                'service "{2}" is ambiguous. In a future version of Docker, '
-                'it will designate a "named" volume '
-                '(see https://github.com/docker/docker/pull/14242). '
-                'To prevent unexpected behaviour, change it to "./{0}:{1}"'
-                .format(host_path, container_path, service_name)
-            )
-
+        if host_path.startswith('.'):
```
mnowster: If the path starts with a `~`, does this mean it won't get expanded now?

compose/config/config.py
```diff
@@ -495,18 +488,10 @@ def resolve_volume_path(volume, working_dir, service_name):
...
+            host_path = expand_path(working_dir, host_path)
         host_path = os.path.expanduser(host_path)
-        return "%s:%s" % (expand_path(working_dir, host_path), container_path)
+        return "%s:%s" % (host_path, container_path)
```
mnowster: Minor cleanup point: while we're here, we could update this line to use `.format()` for the string rather than the old skool `%s` .

tests/unit/config/config_test.py
```diff
@@ -434,59 +440,39 @@ def test_volume_binding_with_home(self):
...
-        with mock.patch('compose.config.config.log.warn') as warn:
-            make_dict(volumes=['/container/path'])
-            self.assertEqual(0, warn.call_count)
-
-            make_dict(volumes=['/data:/container/path'])
-            self.assertEqual(0, warn.call_count)
-
-            make_dict(volumes=['.:/container/path'])
-            self.assertEqual(0, warn.call_count)
+    def test_name_does_not_expand(self):
```
mnowster: yay, I like the splitting up of these tests :space_invader:

compose/config/config.py
aanand: No, it'll still get expanded - `~` is handled by the `os.path.expanduser` line below.

compose/config/config.py
aanand: Good point - have done.

mnowster:
LGTM

PR #2125, opened 2015-10-01
Pass all DOCKER_ env vars to py.test

aanand:
This ensures that `tox` will run against SSL-protected Docker daemons.


dnephin:
LGTM, one test failed so rebuilding. Hopefully we don't have another flaky test already...

dnephin:
Rebuild worked, let's hope that flake is extremely rare (it's the first I've seen).

PR #2126, opened 2015-10-01
1188 support ro options in volumes from

mnowster:
Rebased version of this PR https://github.com/docker/compose/pull/1420

### DIFF
--- a/compose/project.py
+++ b/compose/project.py
@@ -37,7 +37,10 @@ def sort_service_dicts(services):
         return [link.split(':')[0] for link in links]
 
     def get_service_names_from_volumes_from(volumes_from):
-        return [volume_from.split(':')[0] for volume_from in volumes_from]
+        return [
+            parse_volume_from_spec(volume_from).source
+            for volume_from in volumes_from
+        ]
 
     def get_service_dependents(service_dict, services):
         name = service_dict['name']
@@ -195,7 +198,7 @@ class Project(object):
                         raise ConfigurationError(
                             'Service "%s" mounts volumes from "%s", which is '
                             'not the name of a service or container.' % (
-                                volume_from_config,
+                                service_dict['name'],
                                 volume_from_spec.source))
                 volumes_from.append(volume_from_spec)
             del service_dict['volumes_from']
--- a/compose/project.py
+++ b/compose/project.py
@@ -17,8 +17,10 @@ from .legacy import check_for_legacy_containers
 from .service import ContainerNet
 from .service import ConvergenceStrategy
 from .service import Net
+from .service import parse_volume_from_spec
 from .service import Service
 from .service import ServiceNet
+from .service import VolumeFromSpec
 from .utils import parallel_execute
 
 
@@ -34,12 +36,15 @@ def sort_service_dicts(services):
     def get_service_names(links):
         return [link.split(':')[0] for link in links]
 
+    def get_service_names_from_volumes_from(volumes_from):
+        return [volume_from.split(':')[0] for volume_from in volumes_from]
+
     def get_service_dependents(service_dict, services):
         name = service_dict['name']
         return [
             service for service in services
             if (name in get_service_names(service.get('links', [])) or
-                name in service.get('volumes_from', []) or
+                name in get_service_names_from_volumes_from(service.get('volumes_from', [])) or
                 name == get_service_name_from_net(service.get('net')))
         ]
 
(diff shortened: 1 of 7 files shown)


compose/service.py
```diff
@@ -953,6 +949,35 @@ def parse_volume_spec(volume_config):
...
+    volumes_from = []
+    if isinstance(volume_from_spec.source, Service):
+        containers = volume_from_spec.source.containers(stopped=True)
+        if not containers:
+            volumes_from = ["{}:{}".format(volume_from_spec.source.create_container().id, volume_from_spec.mode)]
+        else:
+            volumes_from = ["{}:{}".format(container.id, volume_from_spec.mode) for container in containers]
+    elif isinstance(volume_from_spec.source, Container):
+        volumes_from = ["{}:{}".format(volume_from_spec.source.id, volume_from_spec.mode)]
+    return volumes_from
```
dnephin: Instead of initializing a list and then overriding it in each branch, can we return immediately from each branch?

```python

if isisntance(...):
   return ...

if isinstance(...):
   return

return []
```

compose/project.py
```diff
@@ -176,20 +184,23 @@ def get_links(self, service_dict):
...
+                        container_name = Container.from_id(self.client, volume_from_spec.source)
+                        volume_from_spec = VolumeFromSpec(container_name, volume_from_spec.mode)
                     except APIError:
                         raise ConfigurationError(
                             'Service "%s" mounts volumes from "%s", which is '
                             'not the name of a service or container.' % (
                                 service_dict['name'],
-                                volume_name))
+                                volume_from_spec.source))
+                volumes_from.append(volume_from_spec)
```
dnephin: We talked about it being a little strange that `VolumeFromSpec().source` is a string in some cases, and an object in others, but I think it's not that big of a deal in the end, so I'm ok with it.

dnephin:
Just one minor request, otherwise looks good

compose/service.py
mnowster: Yup.

compose/project.py
mnowster: Yeh I agree with you it is a bit strange. I did have a go at refactoring it, but threw away my attempt as it wasn't right(but useful as I subsequently learnt more about `volumes_from`). If we have time during bug fix we can take another crack at refactoring it but agree it's not blocking for rn.

compose/service.py
```diff
@@ -988,6 +984,38 @@ def parse_volume_spec(volume_config):
...
+    """
+    if isinstance(volume_from_spec.source, Service):
+        containers = volume_from_spec.source.containers(stopped=True)
+        if not containers:
+            return ["{}:{}".format(volume_from_spec.source.create_container().id, volume_from_spec.mode)]
+
+        container = containers[0]
+        return ["{}:{}".format(container.id, volume_from_spec.mode)]
+    elif isinstance(volume_from_spec.source, Container):
+        return ["{}:{}".format(volume_from_spec.source.id, volume_from_spec.mode)]
```
dnephin: It looks like we lost the default case of `return []`

docs/yml.md
```diff
@@ -346,11 +346,13 @@ should always begin with `.` or `..`.
 
 ### volumes_from
 
-Mount all of the volumes from another service or container.
+Mount all of the volumes from another service or container, with the
+supported flags by docker : ``ro``, ``rw``.
```
aanand: Wording's a bit strange here. Maybe:

> Mount all of the volumes from another service or container, optionally specifying read-only access (`ro`) or read-write (`rw`).

compose/service.py
mnowster: Yep, I removed it when I was changing them to return statements, because I realised, I _think_ that it isn't a valid case/return value.

By this point in the code, the `volume_from`  *has* to be either a service or container as done by this bit of code, https://github.com/docker/compose/blob/master/compose/project.py#L176 which gets called before it's value is passed into `Service`. So there should not be a situation where we `return []`, does that make sense?

docs/yml.md
mnowster: Good point, I'll update.

dnephin:
LGTM

dnephin:
I think the "Fixed tests" commit could be squashed into the previous one, but not a big deal.

mnowster:
@dnephin I have squashed the commit :)

aanand:
LGTM

PR #2130, opened 2015-10-02
Add flag for stops all containers if any container was stopped.

### LINKED ISSUE #2474: Preserve previous exit semantics and behavior
Prior to docker-compose 1.5.x, if a single container exited, the entire set of compose's running containers exited as well. (This is actually a rather major semantic change for 1.4.2 -> 1.5.x IMO.)

There are still cases where the previous behavior is more desirable. For one example: running an integration test suite using [dockerized Selenium container(s)](https://github.com/SeleniumHQ/docker-selenium) -- ideally when the integration tests exit, the entire docker compose process would exit as well.

Is there any way to enable this prior behavior through a flag or other mechanism?

scipetr:
Fixes #2474

The pull request [#1754](https://github.com/docker/compose/pull/1754) has changed the behavior of the commands `docker-compose up`.

I added the option `--cascade-stop` to return to the old behavior.

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -590,24 +590,32 @@ class TopLevelCommand(DocoptCommand):
         Usage: up [options] [SERVICE...]
 
         Options:
-            -d                     Detached mode: Run containers in the background,
-                                   print new container names.
-            --no-color             Produce monochrome output.
-            --no-deps              Don't start linked services.
-            --force-recreate       Recreate containers even if their configuration and
-                                   image haven't changed. Incompatible with --no-recreate.
-            --no-recreate          If containers already exist, don't recreate them.
-                                   Incompatible with --force-recreate.
-            --no-build             Don't build an image, even if it's missing
-            -t, --timeout TIMEOUT  Use this timeout in seconds for container shutdown
-                                   when attached or when containers are already
-                                   running. (default: 10)
+            -d                         Detached mode: Run containers in the background,
+                                       print new container names.
+                                       Incompatible with --abort-on-container-exit.
+            --no-color                 Produce monochrome output.
+            --no-deps                  Don't start linked services.
+            --force-recreate           Recreate containers even if their configuration
+                                       and image haven't changed.
+                                       Incompatible with --no-recreate.
+            --no-recreate              If containers already exist, don't recreate them.
+                                       Incompatible with --force-recreate.
+            --no-build                 Don't build an image, even if it's missing
+            --abort-on-container-exit  Stops all containers if any container was stopped.
+                                       Incompatible with -d.
+            -t, --timeout TIMEOUT      Use this timeout in seconds for container shutdown
+                                       when attached or when containers are already
+                                       running. (default: 10)
         """
         monochrome = options['--no-color']
         start_deps = not options['--no-deps']
+        cascade_stop = options['--abort-on-container-exit']
         service_names = options['SERVICE']
         timeout = int(options.get('--timeout') or DEFAULT_TIMEOUT)
         detached = options.get('-d')
+
+        if detached and cascade_stop:
+            raise UserError("--abort-on-container-exit and -d cannot be combined.")
 
         to_attach = project.up(
             service_names=service_names,
@@ -619,7 +627,7 @@ class TopLevelCommand(DocoptCommand):
         )
 
         if not detached:
-            log_printer = build_log_printer(to_attach, service_names, monochrome)
+            log_printer = build_log_printer(to_attach, service_names, monochrome, cascade_stop)
             attach_to_logs(project, log_printer, service_names, timeout)
 
     def version(self, project, options):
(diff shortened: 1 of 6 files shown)


dnephin:
In https://github.com/docker/compose/pull/1754#issuecomment-154218084 I propose some ways to make this work without adding a flag. Do those suggestions work for your case?

If not, could you tell us more about why this is necessary?

AndersKOlsson:
+1

Added a use case description in https://github.com/docker/compose/pull/1754#issuecomment-157670924

mbdas:
I agree the old behavior is needed. One can think of one step further where individual containers can optionally label if on its exit, all should die. By default the new behavior will hold true. A collection of containers would vary from use case to use case. So, giving more choice would help the developer to define the behavior.

simonvanderveldt:
+1 Added our use case to #1754 as well.

compose/cli/main.py
```diff
@@ -565,7 +567,7 @@ def up(self, project, options):
         )
 
         if not detached:
-            log_printer = build_log_printer(to_attach, service_names, monochrome)
+            log_printer = build_log_printer(to_attach, service_names, monochrome, cascade_stop)
             attach_to_logs(project, log_printer, service_names, timeout)
```
dnephin: I think we need a check here before this condition to raise an error if this new flag is used with `-d`, since they don't work together, we shouldn't just ignore it in the `-d` case.

compose/cli/main.py
```diff
@@ -545,12 +545,14 @@ def up(self, project, options):
             --no-recreate          If containers already exist, don't recreate them.
                                    Incompatible with --force-recreate.
             --no-build             Don't build an image, even if it's missing
+            --cascade-stop         Stops all containers if any container was stopped.
```
dnephin: I'm not sure about this name for the flag, I don't think it really captures the behaviour correctly. Let's try and brainstorm some other options.

I'll throw in a couple options:
* `--shutdown-on-first-container-exit` (feels a little too verbose, but I think captures the behaviour well)
* `--shutdown-on-exit` (more concise, less clear)
* `--stop-on-exit`

tests/unit/multiplexer_test.py
```diff
@@ -43,3 +44,16 @@ def problematic_iterator():
 
         with self.assertRaises(Problem):
             list(mux.loop())
+
+    def test_cascade_stop(self):
+        mux = Multiplexer([
+            ((lambda x: sleep(0.01) or x)(x) for x in ['after 0.01 sec T1', 'after 0.02 sec T1', 'after 0.03 sec T1']),
+            ((lambda x: sleep(0.02) or x)(x) for x in ['after 0.02 sec T2', 'after 0.04 sec T2', 'after 0.06 sec T2']),
+        ], cascade_stop=True)
```
dnephin: It would be great to test this without sleeping, but then I guess a failure would freeze the test suite.

Could we have one stream without any sleeps, and another that emits one event then sleeps for 0.01?  That way we should be sleeping at most just the 0.1s, and the one that doesn't sleep should always finish before that.

compose/cli/main.py
aanand: - We've got a real verbosity/clarity problem here, yeah. Is "on-container-exit" a good compromise?
- Perhaps "abort" instead of "shutdown" or "stop"?

```
docker-compose up --abort-on-container-exit
```

dnephin: sounds good to me

mbdas:
One quick question how does container restart policy be handled? On failure 5 should mean that the restart of 5 times should not be treated as stopped and kill all containers. What if one of the containers have restart always in compose? Then this flag would be ignored for that container?

dnephin:
Yes, that is one of the reasons we changed this behaviour, it's incompatible with restart policy.  Maybe we should add a warning if this flag is used and any service has a restart policy set, informing users that a restart policy won't work unless you run with `-d`.

mbdas:
Why is "-d" coming in context with this new flag? Whether compose is running the containers in foreground or background, this flag should be applicable in both situations, isn't that the case?

As for restart policy, one can use docker inspect to check the restart max and current restart count to not take an action with this kill all flag but take an action of killing all containers when the threshold is crossed for that container. For any container that has restart always policy, this flag will completely ignore that container lifecycle for abort policy. compose is almost playing a role of container manager, so to support the restart policy would be a great thing.

dnephin:
Container restart policy is supported with the default behaviour (without this flag). I don't want to add even more complexity to support this legacy behaviour. I think we should just warn that if you're using this flag, you don't get restarts.

If you want restarts, don't don't use `--abort-on-container-exit`. They are semantically incompatible.

dnephin:
Feature freeze for the 1.6 release is one week from today. If you are not able to make these changes, please comment on the issue. Someone else might be interested in carrying this PR and getting it in for the release.

aanand:
LGTM

dnephin:
LGTM

Issue #2135, opened 2015-10-03
[Windows] ValueError: LogConfig.type must be one of (json-file, syslog, none)


alexandrev:
I'm getting this error when I was trying to run a docker-compose up command:

```bash
Status: Downloaded newer image for cpuguy83/docker-grand-ambassador:latest===>]     32 B/32 B6 MB
Traceback (most recent call last):
  File "C:\Python34\Scripts\docker-compose-script.py", line 9, in <module>
    load_entry_point('docker-compose==1.5.0dev', 'console_scripts', 'docker-compose')()
  File "d:\projects\github\compose\compose\cli\main.py", line 51, in main
    command.sys_dispatch()
  File "d:\projects\github\compose\compose\cli\docopt_command.py", line 23, in sys_dispatch
    self.dispatch(sys.argv[1:], None)
  File "d:\projects\github\compose\compose\cli\command.py", line 32, in dispatch
    super(Command, self).dispatch(*args, **kwargs)
  File "d:\projects\github\compose\compose\cli\docopt_command.py", line 26, in dispatch
    self.perform_command(*self.parse(argv, global_options))
  File "d:\projects\github\compose\compose\cli\main.py", line 146, in perform_command
    return super(TopLevelCommand, self).perform_command(options, *args, **kwargs)
  File "d:\projects\github\compose\compose\cli\command.py", line 64, in perform_command
    handler(project, command_options)
  File "d:\projects\github\compose\compose\cli\main.py", line 553, in up
    timeout=timeout
  File "d:\projects\github\compose\compose\project.py", line 283, in up
    for service in services
  File "d:\projects\github\compose\compose\project.py", line 287, in <listcomp>
    timeout=timeout
  File "d:\projects\github\compose\compose\service.py", line 399, in execute_convergence_plan
    do_build=do_build,
  File "d:\projects\github\compose\compose\service.py", line 303, in create_container
    previous_container=previous_container,
  File "d:\projects\github\compose\compose\service.py", line 648, in _get_container_create_options
    one_off=one_off)
  File "d:\projects\github\compose\compose\service.py", line 661, in _get_container_host_config
    config=options.get('log_opt', None)
  File "C:\Python34\lib\site-packages\docker\utils\types.py", line 27, in __init__
    ', '.join(self.types._values)
ValueError: LogConfig.type must be one of (json-file, syslog, none)´´´

dnephin:
This error looks like it's using an older version of docker-py.

alexandrev:
[...]

PR #2136, opened 2015-10-03
Defining default value for log driver.

alexandrev:
Defining default value for log driver. For the available options I chose
json-file to improve the compability across different OS enviromnents.

This configuration was causing an error I reported in #2135.

### DIFF
--- a/compose/service.py
+++ b/compose/service.py
@@ -658,7 +658,7 @@ class Service(object):
         cap_add = options.get('cap_add', None)
         cap_drop = options.get('cap_drop', None)
         log_config = LogConfig(
-            type=options.get('log_driver', ""),
+            type=options.get('log_driver', "json-file"),
             config=options.get('log_opt', None)
         )
         pid = options.get('pid', None)


alexandrev:
Hi, I think now I sign-off my PR as the link you give me. Sorry for don't knowing the rules. My bad. I hope everything is OK now.

mnowster:
Hi @alexandrev, we set the default to be a "" rather than `json-file` on purpose. The extra context is described in this PR https://github.com/docker/compose/pull/1878 if that's helpful.

If you would like the `log_driver` to be `json-file` then you can set that in your .yml, https://docs.docker.com/compose/yml/#log-driver

alexandrev:
Hi @mnowster,

 

First of all, I don’t want to make any offense to your work and decisions as a team, because I’m a newbie here and I have no right to make this kind of assumptions. At first, I want to apologize to do a change that you already discuss about it and make a decision, so please forgive me for that.

 

Then, I’m not sure that I agree with your decision because, with the current code you have a problem of interoperatiblity between environments. Now, I can use a docker-compose.yml in my Linux environment (or MacOS X I think) and then, the same file is crashing on Windows client (not Windows engine, only client). And it is crashing, it is not showing a message or something like that, only crash.

 

This behavior make the docker-compose.yml not cross-platform and I think we should work to you have the least thing environment dependant in your docker-compose.yml. And that’s why I think we should define a default value for Windows. Maybe is not in docker-compose code but in docker-client code I don’t know or maybe we should do it only for Windows platform but I think the current behavior is not the best one.

 

And once again,  I apologize if these words offends you. It’s not its mean but to help to do the best component for every platform.

 

Thanks for your time and your amazing work

dnephin:
@alexandrev I agree we need to fix the issue,but this isn't the right fix for it.  Let's continue triaging the issue in #2135

mnowster:
Hi @alexandrev , no offense taken at all, I appreciate the time and effort you're contributing with us to ensure we're as compatible on windows as we want to be.

I linked to the PR to help with the context as to why we made that decision and the intent is that it should still be compatible with windows, so if it isn't, that's a bug and we'll figure that out and fix it :)

PR #2140, opened 2015-10-05
Reduce the scope of sys.stdout patching

dnephin:
Using `@mock.patch()` on the entire test method makes it really difficult to debug test failures in these functions. Adding a breakpoint (with `import pdb; pdb.set_trace()`) fails because there is no stdout for the prompt.

This change moves the patch to be around only the one call that needs it. This comes with some added benefits:
* the tests are now more explicit about what is supposed to generate the output that is being tested
* there's no need to `truncate()` the buffer


aanand:
LGTM

PR #2144, opened 2015-10-06
Remove unnecessary router.php from wordpress example.

dnephin:
Replaces #868

Apparently this file is not used, and it works without it, so let's just remove it.


aanand:
Confirmed - works just fine without. Perhaps Wordpress has been updated or something.

LGTM

PR #2145, opened 2015-10-06
Update release scripts for release image

### LINKED ISSUE #2129: Move compose image to an org on docker hub
We've just merged the docs for using compose in a container, but the script currently points at `dnephin/docker-compose`. We should move the image to something like `dockercompose/compose` (to match `dockerswarm/`).  Or we could make it an "official" image and just call it "compose".

We also need to update our release process docs to bump the version in this script and publish the new image.

dnephin:
Fixes #2129

Adds steps to the release scripts for building the "image release"


mnowster:
LGTM

Issue #2146, opened 2015-10-07
Warning: There is a boolean value, True in the 'environment' key.


atrauzzi:
Trying to run compose on a fresh Windows 10 and Docker Toolbox install.  Haven't customized or modified the environment, running in the msys based terminal that comes with toolbox.

I get the following error:

```
Warning: There is a boolean value, True in the 'environment' key.
Environment variables can only be strings.
Please add quotes to any boolean values to make them string (eg, 'True').
This warning will become an error in a future release.
```

My YAML files don't have booleans in them, so I'm not sure where it's finding this boolean value or where I should be changing it if it's something on my end.

dnephin:
Please include a copy of your `docker-compose.yml`, or at least a sanitized version of the `environment` section to help us debug this issue.

It would be nice to have more detail in the warning, but I'm not sure if that's possible with the formatters we're using the generate it.

atrauzzi:
```
data:
  container_name: project_dev_data
  image: busybox
  volumes:
    - /var/lib/mysql
    - ./storage:/var/www/storage

mysql:
  container_name: project_dev_mysql
  image: mariadb

  ports:
    - "3306:3306"

  volumes_from:
    - data

  environment:
    MYSQL_ROOT_PASSWORD:
    MYSQL_DATABASE: project
    MYSQL_ALLOW_EMPTY_PASSWORD: yes

web:

  container_name: project_dev_web
  build: .

  volumes_from:
    - data

  links:
    - mysql:mysql

  ports:
    - "8000:8000"
    - "8080:8080"

  volumes:
    - .:/var/www
    - ~/.ssh:/root/.ssh

  volumes_from:
    - data
```

Would it be that `yes` that's causing this issue?

aanand:
Correct - lots of things resolve to booleans in YAML: http://yaml.org/type/bool.html

Put it in quotes (`"yes"`) to make sure it goes through unchanged.

atrauzzi:
Alright, will try once I'm back at that machine, thanks! :)

PR #2149, opened 2015-10-07
Improve boolean warning message.

mnowster:
Including examples of more boolean types, eg yes/N as it's not
always immediately clear that they are treated as booleans by YAML.

As raised in https://github.com/docker/compose/issues/2146


dnephin:
LGTM

Issue #1953, opened 2015-09-01
Feature Request: `docker-compose config` Command


behemphi:
It would be pretty nice to have a `docker-compose validate` command that would validate the structure of the YAML file. The specific use case I have in mind is for CI pipelines.

* Return true if valid YAML of Docker-Compose format
* Return false otherwise

dnephin:
I've been thinking about this, I think it's a good idea.  I was considering naming the command `docker-compose config` instead of `validate`.

The default action would be:
* validate the config
* on success print the config to stdout with variables interpolated and with extends merged in, and exit status 0
* on failure print the error to stderr and exit status 1

With `-q` don't print the config, only errors
With `--services` only print the service names (in correct dependency order), not the full config (replaces #1077)

behemphi:
The command name is not terribly important to me, and `config` makes sense for sure.

Your idea on the output is much better than my own. Helps with debugging at development time too that way. 

+1 for @dnephin

aanand:
Yeah, this is a great idea. See also the featureset I sketched out in https://github.com/docker/compose/issues/1498, which would make shell completion much cleaner and less buggy.

dnephin:
Oh yes, I think these could definitely be merged.  What do you think about naming (`service` vs `config`) ?

aanand:
`config` makes sense to me if (by default) it's going to output the full YAML.

However, looking at #1498 again, the `--with-running-container` and `--with-stopped-container` commands make it begin to feel like conflation. (*One command* to either print out YAML or list services that have a container running?)

dnephin:
Ya, I agree with that. `--with-running-container` and `--with-stopped-container` feel a bit like they could be provided by `ps`. The distinction I was running with was:

* `config` - never hits the api, only deals with values from the config, prints config or parts of it (service names)
* `ps` - hits the api, lists containers

behemphi:
$0.02

My intent was around keeping a compose file in a project and having my CI system validate it.  @dnephin suggestion extends that suggestion to help with the development.

`ps` is an operational tool to tell me, as an ops person, what is running at any given time.  I would exercise this in a "trusty but verify" situation, or when something is wrong (alert comes in) and I use the command to see if a service is down according to docker-compose.  

### TL;DR

I see these functions as orthogonal in purpose and intent.

PR #2153, opened 2015-10-07
ls command to list services in a project

semarj:
Something I've found useful when jumping around to different projects, and forgetting what each one is called.

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -2,6 +2,7 @@ from __future__ import print_function
 from __future__ import unicode_literals
 
 import logging
+import os
 import re
 import signal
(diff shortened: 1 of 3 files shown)


dnephin:
Thanks for the contribution!  

We'd like to add a feature like this, but more along the lines of #1953.  So `docker-compose config --services` would list the services. There are a few other related features that we'd like to group together into a single sub-command `config`.

If you're interested in updating your PR I'd be happy to work with you to get it merged.

semarj:
I'd be happy to work on that if no one gets to it before me.  I'll reopen this if/when I do.

Issue #1070, opened 2015-03-05
Provide "master" builds on master.dockerproject.com


thaJeztah:
This is a continuation on https://github.com/docker/compose/pull/730#issuecomment-77418324

It would be really useful if pre-build binaries of docker-compose are added to https://master.dockerproject.com. This would enable people to download a working version of the latest version for testing if they like "living on the edge".

Also useful to confirm if an issue is resolved ("could you download the latest master to check if the issue is resolved?")

dnephin:
I'd like to do this.

dnephin:
Master builds are now available at  https://dl.bintray.com/docker-compose/master/

thaJeztah:
Thanks @dnephin !

schmunk42:
@dnephin @thaJeztah These builds seem outdated, I just got `1.9.0dev` - you should remove or update them.

dnephin:
If I remember correctly bintray has a weird thing where it doesn't let you push updated to a version after 6 months of the original publishing date. I went in an clicked "update version" on all the master versions, so maybe it will start working now after the next commit to master.

cc @shin-  next time you merge something mind pinging me and we can check to see if this worked?

schmunk42:
@dnephin @shin- That would be great. Btw: Would it also be possible to build binaries on/from the Docker images. My Python knowledge is pretty limited, with Go it seems no problem.

dnephin:
not sure what you mean

schmunk42:
I would like to use these tools on a Docker image, so a static binary would be preferred.

For the Docker CLI I can use a multi-stage build, like so:

```
FROM docker:17.09-rc as docker
FROM dmstr/roj:0.4.0-rc1
COPY --from=docker /usr/local/bin/docker /opt/local/bin/docker
```

I tried this with `docker-compose` also, but there are much more dependencies (on the image which can be build from this repo).

Iristyle:
Linux and Darwin are great... 

Windows would be even better if possible.


Thanks!

PR #2154, opened 2015-10-07
Publish master builds to bintray

dnephin:
Implements #1070 

Using travis-ci we can publish builds to bintray at https://dl.bintray.com/docker-compose/master/

The deploy step only runs on the master branch (by default). We can look to get it setup to work on tags as well, if we are interested in using this for scheduled releases as well.

Also added the bintray upload to appveyor config. The appveyor deploy scripts runs on any branch, but never on pull requests.

In both cases the API key is hidden to anyone submitting a PR, so there's no risk of a malicious PR exposing it. The API key is configured from the web UI of travis-ci and appveyor.

### DIFF
--- a/script/build-linux
+++ b/script/build-linux
@@ -5,7 +5,7 @@ set -ex
 ./script/clean
 
 TAG="docker-compose"
-docker build -t "$TAG" .
+docker build -t "$TAG" . | tail -n 200
 docker run \
     --rm --entrypoint="script/build-linux-inner" \
     -v $(pwd)/dist:/code/dist \
--- a/script/build-windows.ps1
+++ b/script/build-windows.ps1
@@ -29,7 +29,6 @@
 #        .\script\build-windows.ps1
 
 $ErrorActionPreference = "Stop"
-Set-PSDebug -trace 1
 
 # Remove virtualenv
 if (Test-Path venv) {
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -5,6 +5,7 @@
     - id: check-docstring-first
     - id: check-merge-conflict
     - id: check-yaml
+    - id: check-json
     - id: debug-statements
     - id: end-of-file-fixer
     - id: flake8
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,3 +17,13 @@ install: ./script/travis/install
 script:
   - ./script/travis/ci
   - ./script/travis/build-binary
+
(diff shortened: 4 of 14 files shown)


script/build-windows.ps1
```diff
@@ -29,7 +29,6 @@
 #        .\script\build-windows.ps1
 
 $ErrorActionPreference = "Stop"
-Set-PSDebug -trace 1
```
dnephin: I made this change because it was causing (literally) 9000 lines of unnecessary output in appveyor from the "find .pyc files" script.  Hopefully we can do without it.

script/build-linux
```diff
@@ -5,7 +5,7 @@ set -ex
 ./script/clean
 
 TAG="docker-compose"
-docker build -t "$TAG" .
+docker build -t "$TAG" . | tail -n 200
```
bfirsh: Is this causing a problem? Might risk us losing useful debugging output if there is an issue...

bfirsh:
Is it possible to push to bintray from jenkins? That could mean we don't have to add another service, and would mean builds would only get pushed if they're green on jenkins.

script/build-linux
dnephin: Ya, I made this change because on travis it causes 9000+ lines of debug output, which causes the log buffer to fill and requires us to download the log instead of being able to use the UI.  We could fix it by adding `> /dev/null` to some of the run lines in the `Dockerfile`, but then we'd lose it in all cases (and that's just a lot more work to track down).

On jenkins the image is already build and this line only hits the cache, so if there were issues we'd already see them the first time it runs the build.  If there are critical failures we should see them in the last 200 lines.

dnephin:
>  Is it possible to push to bintray from jenkins? That could mean we don't have to add another service, and would mean builds would only get pushed if they're green on jenkins.

I explored jenkins first, but it's not a good option at this time. This only build the master branch, so the build would already need to be green before we merged it into master.

script/build-windows.ps1
mnowster: This is helpful piece of context, can you add it to the commit message so we don't lose it when we merge please?

dnephin:
Added note about removing debug trace from the ps1 script to the commit message.

michaeljs1990:
awesome, thank you for this. I am testing the osx binary right now and it seems to be working great.

mnowster:
LGTM

Issue #2108, opened 2015-09-29
ImportError: No module named 'Queue'


talolard:
When trying to launch docker-compose in a python 3 environment I get the following error
 ````python
(analytics4)tal@tal-desktop:~/dev/analytics/data_migration$ docker-compose up
Traceback (most recent call last):
  File "/home/tal/.virtualenvs/analytics4/bin/docker-compose", line 9, in <module>
    load_entry_point('docker-compose==1.4.2', 'console_scripts', 'docker-compose')()
  File "/home/tal/.virtualenvs/analytics4/local/lib/python3.4/site-packages/pkg_resources.py", line 356, in load_entry_point
    return get_distribution(dist).load_entry_point(group, name)
  File "/home/tal/.virtualenvs/analytics4/local/lib/python3.4/site-packages/pkg_resources.py", line 2431, in load_entry_point
    return ep.load()
  File "/home/tal/.virtualenvs/analytics4/local/lib/python3.4/site-packages/pkg_resources.py", line 2147, in load
    ['__name__'])
  File "/home/tal/.virtualenvs/analytics4/local/lib/python3.4/site-packages/compose/cli/main.py", line 16, in <module>
    from ..project import NoSuchService, ConfigurationError
  File "/home/tal/.virtualenvs/analytics4/local/lib/python3.4/site-packages/compose/project.py", line 12, in <module>
    from .service import ContainerNet
  File "/home/tal/.virtualenvs/analytics4/local/lib/python3.4/site-packages/compose/service.py", line 28, in <module>
    from .utils import json_hash, parallel_execute
  File "/home/tal/.virtualenvs/analytics4/local/lib/python3.4/site-packages/compose/utils.py", line 8, in <module>
    from Queue import Queue, Empty
ImportError: No module named 'Queue'
```
Easy enough fix, 
````python
from Queue import Queue, Empty 
````

Should be 
````python
try:
    from Queue import Queue, Empty 
except:
    from queue import Queue, Empty 
````

**Comment by mnowster** [comment]
Hi @talolard what version of docker-compose are you running?

I believe we've already fixed this in master so should go out with the upcoming 1.5 release.

**Comment by joshuataylor** [comment]
Confirmed that using the latest git checkout solved this for me.

If you're using pip, it's pretty straight forward:
```
pip uninstall docker-compose
pip install git+git://github.com/docker/compose.git
```

**Comment by chiefy** [comment]
I'm not a python guy, so not sure, but...

```
[...]

Issue #1305, opened 2015-04-17
pkg_resources.VersionConflict: (requests 2.2.1 (/usr/lib/python2.7/dist-packages), Requirement.parse('requests>=2.5.2'))


MrMMorris:
version 1.2.0

`docker-compose run blah`
```
Traceback (most recent call last):
  File "/usr/local/bin/docker-compose", line 5, in <module>
    from pkg_resources import load_entry_point
  File "build/bdist.linux-x86_64/egg/pkg_resources.py", line 2749, in <module>
    
  File "build/bdist.linux-x86_64/egg/pkg_resources.py", line 446, in _build_master
    equal ``sys.path``.)
  File "build/bdist.linux-x86_64/egg/pkg_resources.py", line 459, in _build_from_requirements
    
  File "build/bdist.linux-x86_64/egg/pkg_resources.py", line 632, in resolve
    error_info = {}
pkg_resources.VersionConflict: (requests 2.2.1 (/usr/lib/python2.7/dist-packages), Requirement.parse('requests>=2.5.2'))
```

MrMMorris:
`sudo pip install --upgrade docker-compose` fixes it

MrMMorris:
I guess I'll leave it open in case you need to do something on your end :smile:

scotty-c:
To fix the issue run 
````
pip install requests==2.5.3 && pip install docker-compose
````

timani:
Thanks @scotty-c, I had the same error and the problem was I had requests 2.2.1 installed. 

After specifying requests==2.5.3 it worked for me.

tlightsky:
sudo pip install --upgrade docker-compose fixes it
works for me

lwcolton:
This is still a real problem.  By pinning the exact version of requests in setup.py, you prevent anyone from running docker-compose on a system with an upgraded version.  For instance, at this time setup.py requires requests < 2.7, when the latest available version is 2.7   So anyone who wants to run the latest version of the requests library on a system with docker-compose...    can't.    In my testing docker-compose is very much compatible with requests 2.7     So instead of forcing a certain version of that library, it might be nicer to just say  "requests >= 2.6.1" or whatever the lowest working version is.  Then in the README list the tested version for people who care.

lwcolton:
pkg_resources.DistributionNotFound: The 'requests<2.7,>=2.6.1' distribution was not found and is required by docker-compose
root@hlbghlabdocker01:~# pip show requests
---
Metadata-Version: 2.0
[...]

Issue #1290, opened 2015-04-14
Docker-compose doesn't start, complains about missing urllib.parse


JonasT:
I installed docker-compose through pip, but it doesn't run:

```
jonas@cyberman#18: docker-compose 
Traceback (most recent call last):
  File "/usr/bin/docker-compose", line 9, in <module>
    load_entry_point('docker-compose==1.1.0', 'console_scripts', 'docker-compose')()
  File "/usr/lib/python2.7/site-packages/pkg_resources.py", line 348, in load_entry_point
    return get_distribution(dist).load_entry_point(group, name)
  File "/usr/lib/python2.7/site-packages/pkg_resources.py", line 2311, in load_entry_point
    return ep.load()
  File "/usr/lib/python2.7/site-packages/pkg_resources.py", line 2025, in load
    entry = __import__(self.module_name, globals(),globals(), ['__name__'])
  File "/usr/lib/python2.7/site-packages/compose/__init__.py", line 2, in <module>
    from .service import Service  # noqa:flake8
  File "/usr/lib/python2.7/site-packages/compose/service.py", line 10, in <module>
    from docker.errors import APIError
  File "/usr/lib/python2.7/site-packages/docker/__init__.py", line 20, in <module>
    from .client import Client # flake8: noqa
  File "/usr/lib/python2.7/site-packages/docker/client.py", line 34, in <module>
    import websocket
  File "/usr/lib/python2.7/site-packages/websocket/__init__.py", line 22, in <module>
    from ._core import *
  File "/usr/lib/python2.7/site-packages/websocket/_core.py", line 41, in <module>
    from ._url import *
  File "/usr/lib/python2.7/site-packages/websocket/_url.py", line 23, in <module>
    from six.moves.urllib.parse import urlparse
ImportError: No module named urllib.parse
jonas@cyberman#18:
```

I suppose some dependency is missing here? I just installed docker-compose from the latest pip package:
```
pip install -U docker-compose
```
(There's a separate pip3, so I know this used the python2 pip, just in case that's important)

Edit: /usr/bin/python defaults to python2.7 unlike previously stated.

JonasT:
jonas@cyberman#26: /usr/bin/python2 --version
Python 2.7.8
jonas@cyberman#26: /usr/bin/python --version
Python 2.7.8
jonas@cyberman#27: pip show docker-compose
---
Name: docker-compose
Version: 1.1.0
Location: /usr/lib/python2.7/site-packages
[...]

Issue #1288, opened 2015-04-14
ImportError: No module named urllib.parse


Code-Vortex:
I am getting this error and don't know what should i do,
I have tried to update `six` using `pip install -U six` but its up to date `Requirement already up-to-date: six in /usr/lib/python2.7/site-packages`
what should i do in order to solve this issue ?
```
λ - ~ $ python --version
Python 2.7.8
λ - ~ $ docker-compose --version
Traceback (most recent call last):
  File "/usr/bin/docker-compose", line 9, in <module>
    load_entry_point('docker-compose==1.1.0', 'console_scripts', 'docker-compose')()
  File "/usr/lib/python2.7/site-packages/pkg_resources.py", line 348, in load_entry_point
    return get_distribution(dist).load_entry_point(group, name)
  File "/usr/lib/python2.7/site-packages/pkg_resources.py", line 2311, in load_entry_point
    return ep.load()
  File "/usr/lib/python2.7/site-packages/pkg_resources.py", line 2025, in load
    entry = __import__(self.module_name, globals(),globals(), ['__name__'])
  File "/usr/lib/python2.7/site-packages/compose/__init__.py", line 2, in <module>
    from .service import Service  # noqa:flake8
  File "/usr/lib/python2.7/site-packages/compose/service.py", line 10, in <module>
    from docker.errors import APIError
  File "/usr/lib/python2.7/site-packages/docker/__init__.py", line 20, in <module>
    from .client import Client # flake8: noqa
  File "/usr/lib/python2.7/site-packages/docker/client.py", line 34, in <module>
    import websocket
  File "/usr/lib/python2.7/site-packages/websocket/__init__.py", line 22, in <module>
    from ._core import *
  File "/usr/lib/python2.7/site-packages/websocket/_core.py", line 41, in <module>
    from ._url import *
  File "/usr/lib/python2.7/site-packages/websocket/_url.py", line 23, in <module>
    from six.moves.urllib.parse import urlparse
ImportError: No module named urllib.parse
```

dnephin:
What if you do `pip freeze | grep six` ?

I suspect the system installed six is older, the `pip install -U` only installs a user specific version

Code-Vortex:
the result of `pip freeze | grep six`:

```
λ - ~ $ pip freeze | grep six
six==1.9.0
```
and according to [python package index](https://pypi.python.org/pypi/six) this is the most recent version

Code-Vortex:
Is there something i have to do or this is a bug ?!

[...]

PR #2159, opened 2015-10-08
Update pip install instructions to be more reliable

### LINKED ISSUE #1996: Installing docker-compose through pip will sometimes(?) not upgrade websocket to required version
Installing docker-compose through pip will sometimes(?) not upgrade websocket to required version. This means:

1. User uses ```pip install docker-compose```.

2. User tries to use it. This error appears:
```
[root@localhost Server]# docker-compose
  File "/usr/bin/docker-compose", line 9, in <module>
    load_entry_point('docker-compose==1.4.0', 'console_scripts', 'docker-compose')()
  File "/usr/lib/python2.7/site-packages/pkg_resources/__init__.py", line 558, in load_entry_point
    return get_distribution(dist).load_entry_point(group, name)
  File "/usr/lib/python2.7/site-packages/pkg_resources/__init__.py", line 2682, in load_entry_point
    return ep.load()
  File "/usr/lib/python2.7/site-packages/pkg_resources/__init__.py", line 2355, in load
    return self.resolve()
  File "/usr/lib/python2.7/site-packages/pkg_resources/__init__.py", line 2361, in resolve
    module = __import__(self.module_name, fromlist=['__name__'], level=0)
  File "/usr/lib/python2.7/site-packages/compose/cli/main.py", line 10, in <module>
    from docker.errors import APIError
  File "/usr/lib/python2.7/site-packages/docker/__init__.py", line 20, in <module>
    from .client import Client, AutoVersionClient # flake8: noqa
  File "/usr/lib/python2.7/site-packages/docker/client.py", line 23, in <module>
    from . import clientbase
  File "/usr/lib/python2.7/site-packages/docker/clientbase.py", line 7, in <module>
    import websocket
  File "/usr/lib/python2.7/site-packages/websocket/__init__.py", line 22, in <module>
    from ._core import *
  File "/usr/lib/python2.7/site-packages/websocket/_core.py", line 41, in <module>
    from ._url import *
  File "/usr/lib/python2.7/site-packages/websocket/_url.py", line 23, in <module>
    from six.moves.urllib.parse import urlparse
ImportError: No module named urllib.parse
[root@localhost Server]#
```

3 . User needs to google around and dig for this error to find out that this is required to make things work: ```sudo pip install -U websocket```

This is not exactly how the install experience is supposed to be. Why does this not work properly per default? Also check out https://github.com/docker/compose/issues/1288 to see that it's a quite a large amount of people affected.

Please figure out why pip isn't properly installing the required websocket dependency, and make it work per default for all users.

### LINKED ISSUE #2101: [Usability/Feature Request] Do something better than failing with a cryptic ImportError backtrace when websocket package conflicts with distribution package
When installing docker-compose through pip, the installation can conflict with existing websockets package (already provided by the distribution/not installed through pip) as described here: #1996 

While this can be worked around by the user and I understand this is somewhat expected given the limitations of pip and the inherent problem of installing something system-wide with distribution python packages around and no use of virtualenv, I believe there are ways to improve the handling of this:

---

*Idea 1:* Catch the ImportError of the websocket package and display a *helpful message* which describes the problem, and suggests to the user to either wget docker-compose as a finished working package or remove the distribution's websocket package and try reinstalling it with pip to get a compatible version.

*Idea 2:* Figure out how to change the pip package so it somehow installs the correct websockets thing right away even with distribution packages around. (I have no idea of pip so I guess this might be impossible, I'm just adding it here in case there are ways to do this and nobody has bothered to try it)

---

At least the first one should be easily doable and would be a vast improvement for people running into this issue. If other approaches are despite being possible too much effort, please at least consider the first idea of a simple more descriptive helpful error message.

dnephin:
Every week we get one or two reports of failures to install from pip (#2108, #1996, #2101, #1305, #1288, #1290 are some recent ones I was able to find without much effort).

We use some very common packages (six, requests) which are newer than the version installed on most systems.  This is always going to lead to problems.

I think we should suggest that people use a virtualenv to install with pip to prevent these issues.  If they're a less experienced python user it should be a better experience for them.

Fixes #1996 
Fixes #2101

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -60,7 +60,14 @@ To install Compose, do the following:
 
 ### Install using pip
 
-    $ sudo pip install -U docker-compose
+Compose can be installed from [pypi](https://pypi.python.org/pypi/docker-compose)
+using `pip`.  If you install using `pip` it is highly recommended that you use a
+[virtualenv](https://virtualenv.pypa.io/en/latest/) because many operating systems
+have python system packages that conflict with docker-compose dependencies. See
+the [virtualenv tutorial](http://docs.python-guide.org/en/latest/dev/virtualenvs/)
+to get started.
+
+    $ pip install docker-compose
 
 
 ### Install as a container


docs/install.md
```diff
@@ -60,7 +60,16 @@ To install Compose, do the following:
 
 ### Install using pip
 
-    $ sudo pip install -U docker-compose
+Compose can be installed from [pypi](https://pypi.python.org/pypi/docker-compose)
+using `pip`.  If you install using `pip` it is highly recommended that you use a
+[virtualenv](https://virtualenv.pypa.io/en/latest/) because many operating systems
+have python system packages that conflict with docker-compose dependencies.
```
mnowster: I agree with this comment and us recommending using pip. 

I'm not convinced we should give the steps, below, on how to do it, as there is so much room there for people's personal setup/preferences that I don't think we should be prescribing that.

dnephin: That's fair, my thinking is this: If someone is familiar with virtualenv, they're just going to do things their way regardless of the steps we outline.  If someone is not familiar with virtualenv, calling out the steps to use it (and have the executable available without having to source the virtualenv) is going to be really helpful to them.

Maybe we can address this by explaining that this example is just a suggestion?

mnowster: I think I'd rather we link them to elsewhere to learn about it, that'll take them through the setup steps which already includes explanation so they can learn why they're doing those steps. Like maybe http://docs.python-guide.org/en/latest/dev/virtualenvs/ this is a good one as it also includes the additional tool of `virtualenvwrapper`.

dnephin:
Cool, I've made those changes. I linked to the virtualenv tutorial instead, and reverting the `pip install` example to be just the regular install.

I removed the `sudo` and `-U` because I think those are wrong in the common case.  The pip command will let you know to run `-U` if you need it, so it shouldn't be too confusing in the rare case when someone does need it.

mnowster:
LGTM

PR #2161, opened 2015-10-08
WIP: Add changelog for 1.5.0

dnephin:
**DON'T MERGE**

Opening a PR to make sure I haven't missed anything, and as an FYI so we don't duplicate the effort.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,52 @@
 Change log
 ==========
 
+1.5.0 (2015-10-13)
+------------------
+
+Major Features
+
+-   Compose is now available on windows.
+-   Environment variable can be used in the compose file. See
+    https://github.com/docker/compose/blob/129092b7/docs/yml.md#variable-substitution
+-   Multiple compose files can be specified, allowing you to override
+    setting in the default compose file. See
+    https://github.com/docker/compose/blob/129092b7/docs/reference/docker-compose.md
+    for more details.
+-   Configuration validation is now a lot more strict
+-   `up` now waits for all services to exit before shutting down
+-   Support for the new docker networking can be enabled with
+    the `--x-networking` flag
+
+New Features
+
+-   `volumes_from` now supports a mode option allowing for read-only
+    `volumes_from`
+-   Volumes that don't start with a path indicator (`.` or `/`) will now be
+    treated as a named volume. Previously this was a warning.
+-   `--pull` flag added to `build`
+-   `--ignore-pull-failures` flag added to `pull`
+-   Support for the `ipc` field added to the compose file
+-   Containers created by `run` can now be named with the `--name` flag
+-   If you install Compose with pip or use it as a library, it now
+    works with Python 3
+-   `image` field now supports image digests (in addition to ids and tags)
+-   `ports` now supports ranges of ports
+-   `--publish` flag added to `run`
+-   New subcommands `pause` and `unpause`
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,47 @@
...
 ==========
 
+1.5.0 (2015-10-13)
+------------------
+
+Major Features
+
+-   Compose is now available on windows.
+-   Environment variable can be used in the compose file. See
+    https://github.com/docker/compose/blob/master/docs/yml.md#variable-substitution
```
thaJeztah: To prevent these links to break in future, consider linking to a specific commit, or to the document in the release branch/tag

CHANGELOG.md
```diff
@@ -1,6 +1,47 @@
 Change log
 ==========
 
+1.5.0 (2015-10-13)
+------------------
+
+Major Features
+
+-   Compose is now available on windows.
```
mnowster: I think we may need to indicate somehow that there are some known issues with relative volume paths on windows.

mnowster: If https://github.com/docker/compose/pull/2181 makes it into the release candidate, you can ignore my comment above.

dnephin: cool, I think it will make it in

CHANGELOG.md
dnephin: Normally I'd like to link to a tag, but we don't create the tag until after the official release, so anyone reading these as part of an rc won't get a correct link.

I also can't link to a branch because we'll drop the bump-1.5.0 branch after release.

I guess that only leaves a commit, which is an ugly URL, but not sure what else to do

CHANGELOG.md
```diff
@@ -1,6 +1,47 @@
...
+-   Compose can be installed and run as a docker image
+-   `volumes_from` now supports a mode option allowing for read-only
+    `volumes_from`
+-   Volumes that don't start with a path indicator (`.` or `/`) will now be
+    treated as a named volume. Previously this was a warning.
+-   `--pull` flag added to `build`
+-   `--ignore-pull-failures` flag added to `pull`
+-   Support for the `ipc` field added to the compose file
+-   Containers created by `run` can now be named with the `--name` flag
+-   Python 3 support
```
bfirsh: "If you install Compose with PyPi or as a Python library, it now works with Python 3."

Most people probably don't know they're using Python. ;)

dnephin: Ah, good call!

CHANGELOG.md
```diff
@@ -1,6 +1,47 @@
...
+    https://github.com/docker/compose/blob/master/docs/reference/docker-compose.md
+    for more details.
+-   Configuration validation is now a lot more strict
+-   `up` now waits for all services to exit before shutting down
+-   Support for the new docker networking can be enabled with
+    the `--x-networking` flag
+
+New Features
+
+-   Compose can be installed and run as a docker image
```
bfirsh: It might be worth sticking this at the bottom, clarifying that it's an experimental feature - it might confuse users if they think this is the default. (How does this work? Do I have to change what I'm doing? etc)

dnephin:
Updated based on feedback

CHANGELOG.md
thaJeztah: Good point :/

Perhaps leave it like this during the RC period, and update together with the bump?
Is there a tag created for the RC's?

PR #2162, opened 2015-10-08
Docs changes, fixed some links, removed older links, etc

dnephin:
This is a collection of small docs changes. Each commit should be well isolated from the others.

cc @moxiegirl 

* Moved `COPY` to the end of the `Dockerfile` so that rebuilding doesn't re-download all the external repos. It makes it much faster to re-run `make -C docs`. Is this change ok for the larger docs build?
* I removed some links from the footer menu on most pages. 
 * The "environment variable" page is basically deprecated, so I don't think we want to link to it from everywhere.
 * The "Command completion" doc is only really useful during installation, and we already link to it from the install docs.
 * The reference section doesn't really need a link back to the getting started guides. I think these guides are only really useful when someone is "getting started" with compose. Once they're hitting the reference section it's not going to be useful for them.
* Fixed cross-section links in the reference section

### DIFF
--- a/docs/completion.md
+++ b/docs/completion.md
@@ -66,4 +66,3 @@ Enjoy working with Compose faster and with less typos!
 - [Get started with WordPress](wordpress.md)
 - [Command line reference](/reference)
 - [Yaml file reference](yml.md)
-- [Compose environment variables](env.md)
--- a/docs/completion.md
+++ b/docs/completion.md
@@ -1,6 +1,6 @@
 <!--[metadata]>
 +++
-title = "Command Completion"
+title = "Command-line Completion"
 description = "Compose CLI reference"
 keywords = ["fig, composition, compose, docker, orchestration, cli,  reference"]
 [menu.main]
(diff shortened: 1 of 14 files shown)


docs/completion.md
```diff
@@ -1,6 +1,6 @@
 <!--[metadata]>
 +++
-title = "Command Completion"
+title = "Command-line Completion"
```
albers: Schouldn't this be commandline?

dnephin: I was going by the wikipedia term for it (https://en.wikipedia.org/wiki/Command-line_completion).

moxiegirl:
@albers @dnephin `command-line interface` with `interface` because both` command` and `line` modify interface.  Alone just `command line` --- `command` is modifier to just `line` in this construction.   So, `command-line completion` is correct.  We are most likely inconsistent in the docs.  We should correct as we find them.

@dnephin Regarding the `Dockerfile` change that works.  We are in the process of final testing on some infrastructure work and if that goes well, will be updating these anyway.  Thanks for the lesson too.

RE the other changes LGTM.  They are minor so I'm good with a merge.

albers:
@moxiegirl Thanks for clarification.

mnowster:
LGTM

Issue #2089, opened 2015-09-23
Formal specification of Compose file


bfirsh:
We now have a JSON Schema definition of a Compose file:

https://github.com/docker/compose/blob/master/compose/config/service_schema.json https://github.com/docker/compose/blob/master/compose/config/fields_schema.json

We should probably:
- Put this somewhere so it can be shared by libcompose https://github.com/docker/libcompose/issues/34 . This could probably just be in same repository (like Docker's image specification in https://github.com/docker/distribution ), but at some point we might want to be adventurous and put it in a separate repository.
- Attach a human readable definition of the specification. This could be just a high-level overview, which points at the JSON schema and docs. Or perhaps adapted from the YML definition in the documentation. Or perhaps auto-generated somehow. Not really sure what's best here.

dnephin:
As part of this I think we need to add a version and follow semver when we make changes.

I think it might also be a good idea to try and combine the two schemas into a single schema. I think we can still validate against nested schemas, so we should be able to continue to support abstract base services even with a single schema.

mnowster:
+1 on semver.

It started out as a single schema but was split to cater for the 2 stages of validation and abstract services. Schema has it's limitations and I'm not sure how re-joining it _and_ catering for the abstract base would work. Be interesting to see how that one pans out and if you have plans, do a spike on it to see what's possible, before we release and share with libcompose.

I'm not super convincedif it'd be worth the amount of effort and complexity to re-join. Hmm. Would need more headspace to think about it some more and with access to how other people would intend to use it.

dnephin:
You can treat every nested schema exactly the same way you treat a top-level schema in a different file. So using a single file schema, we can support the 2 stage validation by passing in the path to the nested schema for the first phase.

mnowster:
When I tried that previously, I couldn't get it to work, I wonder if that's because I hadn't got the RefResolver working at that point. Hmm. Interesting. Ok cool. I'll be happy for someone else to work on that area of code for a while, I need a break from it :boat:

dnephin:
Oh ya, I remember that getting the RefResolver to work was really difficult when I did it

dnephin:
We should address #2299 before we publish a format spec as well

thaJeztah:
[...]

PR #2164, opened 2015-10-09
Rename "yml reference" to "Compose file reference"

dnephin:
I think we should start referring to the configuration file as the "Compose file". 

1) We're talking about formalizing the spec (#2089), so it would be good to starting referring to it by a proper name early
2) Calling it the "yml file" I think puts the focus on the wrong thing. Technically compose works with json files as well (because yml is just a superset of json). 
3) This reference doc includes the docs for variable substitution now (which is not related to yml)

I haven't renamed the document so that we don't break any links. I'm not sure if there is any way to alias or redirect from the old url.

I also made a small change to the `volume_driver` section. I think it would be good to start splitting out this catch-all section, so I started with one small change.

cc @moxiegirl

### DIFF
--- a/docs/completion.md
+++ b/docs/completion.md
@@ -65,4 +65,4 @@ Enjoy working with Compose faster and with less typos!
 - [Get started with Rails](rails.md)
 - [Get started with WordPress](wordpress.md)
 - [Command line reference](./reference/index.md)
-- [Yaml file reference](yml.md)
+- [Compose file reference](yml.md)
--- a/docs/completion.md
+++ b/docs/completion.md
@@ -65,4 +65,4 @@ Enjoy working with Compose faster and with less typos!
 - [Get started with Rails](rails.md)
 - [Get started with WordPress](wordpress.md)
 - [Command line reference](./reference/index.md)
-- [Compose file reference](yml.md)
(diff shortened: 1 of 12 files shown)


moxiegirl:
@dnephin You can rename the file to `compose-file.md` and then put an `alias` in  the metadata (See https://gohugo.io/extras/aliases/)

```
aliases = ["/compose/yml"]
```

bfirsh:
LGTM

mnowster:
@dnephin 'cos I've merged some other docs PRs this now needs a rebase and then I'll happily merge it.

dnephin:
Cool, I'm going to make a few more changes to this. I'll move the `yml.md` to `compose-file.md` and add an alias.

dnephin:
Ok, renamed yml.md and added an alias for the old url

thaJeztah:
@dnephin needs a rebase now :'(

dnephin:
rebased

thaJeztah:
LGTM, thanks!

mnowster:
@dnephin dammit, sorry, I merged the django docs PR which means this needs rebasing again.

dnephin:
rebased

mnowster:
LGTM.

PR #2166, opened 2015-10-10
Add Bridged Networking Capability

mdaue:
Proposed pull request adds a network module that provides interfaces and netdefs keywords to support  provision of bridged networking on Linux hosts in addition to the existing Docker networking capability. Using the new interfaces keyword (that is removed before Docker keyword validation), define your network IPv4/6 addresses and assign them to container interfaces. Using the new netdefs keyword, define which networks (bridges) the interface(s) should attach to. Both traditional brctl and openvswitch bridges are supported at this time.

All tests pass. New keywords are popped from services before passing it to Docker.


mdaue:
I read the part about each commit having to have a signature too late.

dnephin:
It looks like this has a bunch of old unrelated commits. Could you check your git history and rebase on master?

Issue #2128, opened 2015-10-02
Relative path, .:/data, does not work on windows


mnowster:
Due to us now using `splitdrive` in `split_path_mapping` it reads the `.:` as a drive, rather than a relative path.

mnowster:
For reference, this is the `splitdrive` code. https://github.com/python/cpython/blob/829b49cbd2e4b1d573470da79ca844b730120f3d/Lib/ntpath.py#L120

atrauzzi:
Awesome!  What's my best option for pulling down the latest change and trying it out?

mnowster:
@atrauzzi if you hang tight, there should be a release candidate sooooooooooon :)

dnephin:
https://github.com/docker/compose/releases/download/1.5.0rc1/docker-compose-Windows-x86_64.exe

In the future if you're interested in trying out pre-release features there is also https://dl.bintray.com/docker-compose/master/docker-compose-Windows-x86_64.exe

atrauzzi:
@mnowster :)

@dnephin Awesome, I'd love to help make more work for you guys! ;)

nemild:
@mnowster, @dnephin  Thanks for the fix, but we seem to have the same issue on docker-compose on Windows, even with docker-compose 1.5.1

I used the relative path for volume: `.:/myapp`, and that folder is empty when I do a `docker-compose up`.

If it makes it easier, The relative path doesn't really matter to me - I'm fine with absolute, but even if I replace the relative path with that (my pwd spits out: /d/mercurial/bch ) - it doesn't work. The `myapp` folder in the container is still empty.

On a separate note, how does one debug this? I can't shell into docker as I do on Mac due to interactive mode not being available on Windows. So I've resorted to doing `up`, and having it run an `ls` or `pwd` command, rather than the normal commands.

More details:
OS: Windows 8.1
OS (uname -a): MINGW64_NT-6.3 2.1.0(0.288/5/3) 2015-6-28 x86_64 Msys
Docker 1.9.1, API version: 1.2.1, build a34a1d5, Git commit a34a1d5, Client OS/Arch: windows/amd64, Server OS/Arch: linux/amd64
Docker-Compose 1.5.1

Full docker-compose:
```yaml
db:
  image: postgres
  ports:
    - "5432:5432"

web:
  build: .
  command: bash -c "(foreman run rails s -p 3000 -b '0.0.0.0' &) && ($$(npm bin)/gulp)"
  volumes:
    - .:/myapp
[...]

PR #2175, opened 2015-10-12
2128 relative path on windows

mnowster:
Expand volume paths much earlier so that we don't give relative paths to `splitdrive`.

Should fix https://github.com/docker/compose/issues/2128

### DIFF
--- a/compose/config/errors.py
+++ b/compose/config/errors.py
@@ -1,3 +1,7 @@
+class ConfigError(ValueError):
+    pass
+
+
 class ConfigurationError(Exception):
     def __init__(self, msg):
         self.msg = msg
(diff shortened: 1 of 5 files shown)


compose/config/errors.py
```diff
@@ -1,3 +1,7 @@
+class ConfigError(ValueError):
+    pass
```
dnephin: A TODO about this being legacy and us needing to consolidate to use the `ConfigurationError` would be good here I think

mnowster: Ok, good idea.

mnowster:
Closing in favour of https://github.com/docker/compose/pull/2181

PR #2181, opened 2015-10-13
2128 split drive not handling relative paths

mnowster:
Should fix https://github.com/docker/compose/issues/2128

### DIFF
--- a/compose/config/config.py
+++ b/compose/config/config.py
@@ -485,7 +485,6 @@ def resolve_volume_paths(service_dict, working_dir=None):
 
 def resolve_volume_path(volume, working_dir, service_name):
     container_path, host_path = split_path_mapping(volume)
-    container_path = os.path.expanduser(container_path)
 
     if host_path is not None:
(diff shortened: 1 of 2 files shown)


dnephin:
LGTM

compose/config/config.py
```diff
@@ -485,7 +485,6 @@ def resolve_volume_paths(service_dict, working_dir=None):
 
 def resolve_volume_path(volume, working_dir, service_name):
     container_path, host_path = split_path_mapping(volume)
-    container_path = os.path.expanduser(container_path)
```
dnephin: Oh wait, why is this removed?

mnowster: Because `container_path` should never be relative, so we don't need to call `expanduser` on it. @aanand spotted it last time when I was working through understanding the paths situation.

dnephin: I guess it would expand to the wrong path, that's true.  We added this in 1.4.0, so we should probably call this out in the release notes as a bug fix.

PR #2193, opened 2015-10-14
Docs for shorthand notation of extends. Issue #1989

mrfuxi:
Docs missing in PR #2019

### DIFF
--- a/docs/extends.md
+++ b/docs/extends.md
@@ -59,6 +59,10 @@ You can go further and define (or re-define) configuration locally in
         - DEBUG=1
       cpu_shares: 5
 
+    important_web:
+      extends: web
+      cpu_shares: 10
+
 You can also write other services and link your `web` service to them:
 
     web:
@@ -233,7 +237,8 @@ manually keep both environments in sync.
 ### Reference
 
 You can use `extends` on any service together with other configuration keys. It
-always expects a dictionary that should always contain the key: `service` and optionally the `file` key.
+expects a dictionary that contains a `service` key and optionally a `file` key.
+The `extends` key can also take a string, whose value is the name of a `service` defined in the same file.
 
 The `file` key specifies the location of a Compose configuration file defining
 the extension. The `file` value can be an absolute or relative path. If you


docs/extends.md
```diff
@@ -233,7 +237,8 @@ manually keep both environments in sync.
 ### Reference
 
 You can use `extends` on any service together with other configuration keys. It
-always expects a dictionary that should always contain the key: `service` and optionally the `file` key.
+expects either a dictionary that should always contain the key: `service` and
```
mnowster: Try not to use `simply` in docs, it can be very frustrating for those new to this where nothing feels simple!

How about something like this maybe:

```
The `extends` key can also take a string, whose value is the name of a `service` defined in the same file.
```

mrfuxi: You're right. Done.

mnowster:
LGTM.

@thaJeztah would you be able to give this a review for me as well please? Extends is always a tricky one to explain written down, so feedback appreciated.

thaJeztah:
@mnowster thanks for the ping, will give it a look

docs/extends.md
```diff
@@ -233,7 +237,9 @@ manually keep both environments in sync.
 ### Reference
 
 You can use `extends` on any service together with other configuration keys. It
-always expects a dictionary that should always contain the key: `service` and optionally the `file` key.
+expects a dictionary that should always contain the key: `service` and
```
thaJeztah: I'd drop the "should" and "always", something like:

```Markdown
expects a dictionary that contains a `service` key and optionally a `file` key.
The `extends` key can also take a string, whose value is the name of a `service`
defined in the same file.
```

mrfuxi: It does sound better. Done

thaJeztah:
Thanks @mrfuxi, LGTM

mnowster:
Thanks @thaJeztah , you're so good with words, that does sound even better :sparkles: 

Also thanks @mrfuxi for working with us to make these changes :+1: 

LGTM.

mrfuxi:
Np. I have to find a new issue now :)

PR #2196, opened 2015-10-15
Remove incorrectly placed comment

md2perpe:
I'm not sure if it should be there at all, but at least it should hardly be where it currently is located.


dnephin:
Nice find, this looks correct. 

LGTM

mnowster:
LGTM.

dnephin:
Thanks!

Issue #1752, opened 2015-07-22
`docker-compose pull SERVICE` pulls dependencies


aanand:
If a service `web` depends on a service `db`, and both services specify an image, then `docker-compose pull web` will pull the images for both db and web, rather than just web. There's no `--no-deps` workaround either.

gilclark:
I changed include_links to include_deps.  So it was generalizing the notion of a working with a dependency.  In the past if one was pulling an image and it had links to another image that dependency would be pulled too.  Now it's the same idea just more generalized (links, volumes, net).

aanand:
I've just realised it was always passing `True` as the argument name. Apologies for not actually reading the diff. I've deleted that part of the issue description so people don't get confused by it.

cweagans:
Just opened a PR for this: #2199

PR #2199, opened 2015-10-15
docker-compose pull SERVICE should not pull SERVICE's dependencies

cweagans:
Fixes https://github.com/docker/compose/issues/1752

### DIFF
--- a/compose/project.py
+++ b/compose/project.py
@@ -335,7 +335,7 @@ class Project(object):
         return plans
 
     def pull(self, service_names=None, ignore_pull_failures=False):
-        for service in self.get_services(service_names, include_deps=True):
+        for service in self.get_services(service_names, include_deps=False):
             service.pull(ignore_pull_failures)
 
     def containers(self, service_names=None, stopped=False, one_off=False):


compose/project.py
```diff
@@ -335,7 +335,10 @@ def _get_convergence_plans(self, services, strategy):
         return plans
 
     def pull(self, service_names=None, ignore_pull_failures=False):
-        for service in self.get_services(service_names, include_deps=True):
+        pull_dependencies = True
+        if service_names:
+            pull_dependencies = False
+        for service in self.get_services(service_names, include_deps=pull_dependencies):
```
dnephin: Let's just do `include_deps=False` here. If there are no services_names it's already going to pull everything, so the behaviour should be the same.

dnephin:
Thanks for fixing this!

compose/project.py
cweagans: :+1:

cweagans:
@dnephin Pushed a new commit per your comment earlier.

dnephin:
LGTM (assuming jenky is green)

cweagans:
Is there a process for backporting fixes to an earlier version of Compose? Or is it more of an "upgrade to get new stuff and bug fixes" kind of thing?

mnowster:
@cweagans As a general guide it's more of an upgrade to get fixes. Imbetween major releases we'll do minor ones which include high priority fixes rather than making people wait for 2months for the next major release.

This will go into the 1.5 release. 

LGTM.

cweagans:
@mnowster Roger that. Thanks for the merge.

PR #2200, opened 2015-10-16
Revert networking-related changes to getting started guides

aanand:
While networking support is still experimental and behind a flag, we should leave the getting started guides as they are.


dnephin:
LGTM

mnowster:
LGTM

Issue #229, opened 2014-05-23
Switch to using hyphens as a separator in hostnames


kung-foo:
The hostname format used by fig creates names that are not strictly valid. 

Current pseudo-code:

``` python
name = '_'.join([project, container, instance_num])
```

This generates names like `cluster_hadoop_1`. Underscores are not valid (though in practice _most_ components are tolerant of this). 

Valid names should match `[a-zA-Z0-9\-]+`.

I came across this error when trying to test out some hadoop/hdfs containers and hadoop bailed with an exception saying that `hdfs://flume_hadoop_1/` was not valid URI (even though it was in `/etc/hosts` and `ping flume_hadoop_1` worked just fine).

See: http://en.wikipedia.org/wiki/Hostname#Restrictions_on_valid_host_names

Changing this to dashes is easy, but it would break existing configurations that depend on hard coded container names.

dnephin:
We could add aliases for dashes when we implement #2312, keep both underscores and dashes for a release or two, then drop support for underscores in hostnames.

susu:
Same here, cannot run spark cluster with docker-compose. Showstopper 
for docker-compose, falling back to "manual docker".

aanand:
@susu Why is this a showstopper? The problem of underscores in hostnames is easily worked around, either with link aliases or by changing service names.

susu:
Nope, because spark somehow extracts the underscored hostname of docker
container (projectname_container_1.default_network), and try to use it
(which is an invalid hostname in URI). Anyway, falling back to version 1
"solved" the issue.

aanand:
> spark somehow extracts the underscored hostname of docker container (projectname_container_1.default_network), and try to use it

However it's doing that, that's the wrong thing to do. Is this an image you're using from the Hub, or did you build it yourself?

Switching back to version 1 is not a future-proof solution.

gordontyler:
@aanand The workaround I used for this was to create a network so I could control the name:

```
docker network create spark
```

Use that as an external network in my docker-compose.yml file:

```
networks:
  spark:
    external: true
```

Connect each service to that network:

```
services:
  spark-master:
    image: spark
    command: org.apache.spark.deploy.master.Master --host spark-master
    networks:
      - spark
```

And finally override the container name and hostname in the spark worker service:

```
  spark-worker:
    image: spark
    container_name: "spark-worker"
    hostname: "spark-worker"
[...]

PR #2201, opened 2015-10-16
Replace underscores with hyphens in a container name

kikkomep:
This PR replaces underscores with hyphens when used as separators in a container name. 
The use of hyphens is important when the container name coincides with the hostname of the container: indeed underscores produce invalid hostnames (see [restrictions on valid hostnames](https://en.wikipedia.org/wiki/Hostname#Restrictions_on_valid_host_names)).

### DIFF
--- a/compose/service.py
+++ b/compose/service.py
@@ -853,7 +853,7 @@ def build_container_name(project, service, number, one_off=False):
     bits = [project, service]
     if one_off:
         bits.append('run')
-    return '_'.join(bits + [str(number)])
+    return '-'.join(bits + [str(number)])
 
 
 # Images


dnephin:
Thanks for the PR!

This issue has been raised a few times now, unfortunately making this change now would be a pretty major breaking change. Since there are ways to work around the hostnames, we'd like to avoid this breaking change. Even if we did want to make the change, I think this would be the wrong place for it. We need to add another alias to the links, not adjust the container name.

Links set the service name as a hostname as well, so if you aren't scaling a service you can use the service name (Ex: `web`, or `db`) as the hostname. In practice we've seen that most implementations will accept hostnames with underscores anyway.

Going forward the new networking feature will work with the service name as well. We can look to add aliases that uses dashes, as a way to transition to a new naming scheme.  That work is being tracked in #2312

dnephin:
The original issue for this is #229

Issue #2123, opened 2015-10-01
"variable is not set. Substituting a blank string." being shown when env variables from links used in directives


shrikrishnaholla:
I'm working with master (which is currently at dabf1e8657674014a5bc89f99edbf2fe0629bb71), and while building or starting, I'm getting the above warning, and those commands won't work

Ex: I am using a service

```yaml
rcli:
  image: redis:latest
  links:
    - redis
  command: >
     sh -c 'redis-cli -h "$REDIS_PORT_6379_TCP_ADDR" -p "$REDIS_PORT_6379_TCP_PORT"'
```

Here, since the env variable `REDIS_PORT_6379_TCP_ADDR` is obtained because of linking, and it's non-existent when building, or before starting the `redis` service, I can't use this service. It substitutes the value with an empty string.

```
The REDIS_PORT_6379_TCP_ADDR variable is not set. Substituting a blank string.
The REDIS_PORT_6379_TCP_PORT variable is not set. Substituting a blank string.
```

shrikrishnaholla:
This is also reproducible if the `CMD` directive in the `Dockerfile` uses any env variable

```
CMD fluentd --gemfile /etc/fluent/Gemfile -c /fluentd/etc/$FLUENTD_CONF -p /fluentd/plugins $FLUENTD_OPT
```

Seems like command string is being evaluated or something by `compose` or `docker-py`, and that is being done prematurely

villlem:
This is due https://github.com/docker/compose/issues/1377. For your case double `$`
On the other note
`$REDIS_PORT_6379_TCP_ADDR` - using those env is deprecated.

dnephin:
I think maybe we should document this error explicitly, I suspect a few more people will hit this on release.

It should only apply to variables in the config though, not the `Dockerfile`.

shrikrishnaholla:
@dnephin My bad. It's not hitting Dockerfile. I had replicated the command in the `command` section in `docker-compose.yml`, which was causing the error

shrikrishnaholla:
@villlem 
> `$REDIS_PORT_6379_TCP_ADDR` - using those env is deprecated.

What is the recommended way? I'm using whatever I read in the [docs](https://docs.docker.com/userguide/dockerlinks/#environment-variables)

villlem:
@shrikrishnaholla: Docker populate /etc/host inside container with names from link and also do not change ports. But documentation is confusing. 

See https://docs.docker.com/compose/env/ for details.

[...]

PR #2204, opened 2015-10-16
Attempt to document escaping env vars

mnowster:
To help people who may run into a warning about their env vars being set to empty strings, when they don't realise that they need to escape it with a double dollar sign.

Helps to clarify / fixes https://github.com/docker/compose/issues/2123

### DIFF
--- a/docs/compose-file.md
+++ b/docs/compose-file.md
@@ -428,9 +428,18 @@ Both `$VARIABLE` and `${VARIABLE}` syntax are supported. Extended shell-style
 features, such as `${VARIABLE-default}` and `${VARIABLE/foo/bar}`, are not
 supported.
 
-If you need to put a literal dollar sign in a configuration value, use a double
-dollar sign (`$$`).
+You can use a `$$` (double-dollar sign) when your configuration needs a literal
+dollar sign. This also prevents Compose from interpolating a value, so a `$$`
+allows you to refer to environment variables that you don't want processed by
+Compose.
 
+    web:
+      build: .
+      command: "$$VAR_NOT_INTERPOLATED_BY_COMPOSE"
+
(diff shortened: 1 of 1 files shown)


mnowster:
@thaJeztah @moxiegirl I had a go at trying to write something to capture the warning to help users with this as it could be confusing, as suggested by @dnephin in the issue. Any feedback welcome :)

docs/compose-file.md
```diff
@@ -428,9 +428,17 @@ Both `$VARIABLE` and `${VARIABLE}` syntax are supported. Extended shell-style
 features, such as `${VARIABLE-default}` and `${VARIABLE/foo/bar}`, are not
 supported.
 
-If you need to put a literal dollar sign in a configuration value, use a double
-dollar sign (`$$`).
+When you want to set an environment variable that is not to be interpolated by
+Compose, or the configuration value needs a literal dollar sign you can use a
+double dollar sign (`$$`).
```
moxiegirl: You can use a `$$` (double-dollar sign) when your configuration needs a literal dollar sign.  This also prevents Compose from interpolating a value, so `$$` useful for referring to environment variables but disabling processing by Compose.

<snip>

If you forget an use a single `$` Compose interprets the value as an environment variable and warns:

<snip>

mnowster: @moxiegirl  ah yeh, I see what you're saying and the difference in how it reads. I want to write in a way that's accessible to everyone. I'll update it, thank you for your feedback :smile_cat:

moxiegirl:
@mnowster thank you, that scan's much better. LGTM

thaJeztah:
Thanks @mnowster, LGTM too :+1:

PR #2206, opened 2015-10-16
extends must be copied from override file

### LINKED ISSUE #2205: "extends" and multiple yml files don't work together
I'm currently trying to use the new "multiple files" (#2051) feature and got stuck on extends not working anymore when used together.

Example:
```
# common.yml
baseservice:
  labels:
    - "my.label=my-value"
  environment:
    MY_ENV: "MY_VALUE"

# base.yml
service1:
  extends:
    file: common.yml
    service: baseservice
...more stuff...

# dev.yml
service2:
  extends:
    file: common.yml
    service: baseservice
...more stuff...
```
When started with:
```
docker-compose -f base.yml -f dev.yml up
```
the services do not include the labels and env vars from "baseservice".

Just for testing purposes, I tried to use a random name for the extends/file field in dev.yml and it did not complain about not found files, which looks like it doesn't even try to open the file.

dnephin:
Fixes #2205

`merge_service_dicts()` which is used for extends doesn't copy extends, but we need to copy extends for overriding of files.

In the test for this fix I've used a `tempdir` for creating the necessary file instead of using a real file. I think this makes the test easier to understand because the fixtures are inline, instead of a separate file.

If we're happy with this approach, I think we could convert a few of our other unit tests to use this setup.

### DIFF
--- a/tests/unit/config/config_test.py
+++ b/tests/unit/config/config_test.py
@@ -156,6 +156,43 @@ class ConfigTest(unittest.TestCase):
             config.load(details)
         assert 'Top level object needs to be a dictionary' in exc.exconly()
 
+    def test_load_with_multiple_files_and_extends_in_override_file(self):
+        base_file = config.ConfigFile(
+            'base.yaml',
+            {
+                'web': {'image': 'example/web'},
+            })
+        override_file = config.ConfigFile(
+            'override.yaml',
+            {
+                'web': {
+                    'extends': {
+                        'file': 'common.yml',
+                        'service': 'base',
+                    },
+                    'volumes': ['/home/user/project:/code'],
+                },
+            })
+        details = config.ConfigDetails('.', [base_file, override_file])
+
+        tmpdir = py.test.ensuretemp('config_test')
+        tmpdir.join('common.yml').write("""
+            base:
+              labels: ['label=one']
+        """)
+        with tmpdir.as_cwd():
+            service_dicts = config.load(details)
+
+        expected = [
+            {
+                'name': 'web',
+                'image': 'example/web',
+                'volumes': ['/home/user/project:/code'],
+                'labels': {'label': 'one'},
+            },
+        ]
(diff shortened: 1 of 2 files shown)


tests/unit/config/config_test.py
```diff
@@ -133,6 +134,43 @@ def test_load_with_multiple_files(self):
...
+                    'extends': {
+                        'file': 'common.yml',
+                        'service': 'base',
+                    },
+                    'volumes': ['/home/user/project:/code'],
+                },
+            })
+        details = config.ConfigDetails('.', [base_file, override_file])
+
+        tmpdir = py.test.ensuretemp('config_test')
```
mnowster: It's not immediately clear why you're doing this? How does creating a temporary directory fit into the description of what this test is doing around extends? Could this have been achieved with a fixture?

dnephin: > Could this have been achieved with a fixture?

If we go with the `py.test` way of doing fixtures, maybe. With unittest fixtures, no.

Fixtures are just "common test setup", but there is always other test setup that isn't all that common. This is just setting up the environment so the test exercises the expected functionality.

It's creating a `common.yml` which is referenced from the config above.

I'm not sure how to make this more clear.

dnephin: Added more detail to the PR description about why this is different

mnowster:
I'm LGTM for both code and adopting the new testing approach. I understand now and can see the benefit of having the detail of what you're testing inline with the test.

mnowster:
@aanand are you cool with the direction of this testing approach?

aanand:
Yeah, I like it. Fixtures are good in that an actual directory with actual files is nice and easy to understand, but I never liked the coupling-at-a-distance.

dnephin:
Ya, I think for end-to-end tests having the files could be good, but for unit tests it's easier to just do it inline.  Most of our files exist for our integration tests. I think there are only a couple used by unit tests.

dnephin:
Rebased

PR #2207, opened 2015-10-16
Refactor of ServiceLoader to make it immutable

dnephin:
Mutable objects are harder to debug and harder to reason about.  `ServiceLoader` was almost immutable. There was just a single function which set fields for a second function. Instead of mutating the object, we can pass those values as parameters to the next function.

~~Also fixed a subtle bug with `env_file`. It was being removed from the service_dict, which would result in changes to `env_file` not being reflected in the config_hash.~~  Actually, I think this was probably fine, because the `environment` section would still change with a change to `env_file`. 

Updated some docstrings to be [pep257](https://www.python.org/dev/peps/pep-0257/) compliant (text should start on the first line, not a newline after the `"""`).

Removed `resolve_environment()` from `ServiceLoader()`. It's not really part of the core concern of the class (which is to handle `extends`).

This diff is a lot more readable using [split diff](https://github.com/docker/compose/pull/2207/files?diff=split)

### DIFF
--- a/compose/config/config.py
+++ b/compose/config/config.py
@@ -212,9 +212,16 @@ def load(config_details):
 
 
 class ServiceLoader(object):
-    def __init__(self, working_dir, filename, service_name, service_dict, already_seen=None):
+    def __init__(
+        self,
+        working_dir,
+        filename,
+        service_name,
+        service_dict,
+        already_seen=None
+    ):
         if working_dir is None:
-            raise Exception("No working_dir passed to ServiceLoader()")
+            raise ValueError("No working_dir passed to ServiceLoader()")
 
         self.working_dir = os.path.abspath(working_dir)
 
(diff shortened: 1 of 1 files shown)


compose/config/config.py
```diff
@@ -238,114 +245,106 @@ def detect_cycle(self, name):
...
+def resolve_environment(working_dir, service_dict):
+    """Unpack any environment variables from an env_file, if set.
+    Interpolate environment values if set.
+    """
+    if 'environment' not in service_dict and 'env_file' not in service_dict:
+        return {}
+
+    env = {}
+    if 'env_file' in service_dict:
+        for f in get_env_files(service_dict, working_dir=working_dir):
```
mnowster: Try not to use single letter variable names. It doesn't hurt to expand this out to say something like `env_file` and help being clear/readable.

mnowster:
There is way too much happening in here for a single commit.

Any purely stylistic/pep changes should be in a single commit. It helps reduce the diff noise and when you're trying to reason about why something has changed.

"Removed resolve_environment() from ServiceLoader()" Should also be a separate commit, it is unrelated to the other changes and isn't reflected in the current commit message.

Overall, it's not clear what this PR is setting out to achieve and why, what problem is it fixing?

compose/config/config.py
dnephin: This code was just moved from the class, I did not change it significantly.  In 02c52ae673a66c7a8f6455611d8561d8f6954383 you moved this function into  the class and didn't change this variable name! 

but I can make the change now I guess.

dnephin:
> Overall, it's not clear what this PR is setting out to achieve and why, what problem is it fixing?

I covered that in the PR description, but I'll try to add some more details to it.

compose/config/config.py
mnowster: Hah, I am not perfect either! While you're here making changes, may as well improve it :)

mnowster:
Think some source of confusion was the making it immutable isn't linked to the `env` bug, which now I see is scratched out. 

So I think what you're saying, that this is code cleanup to make it immutable because you think this will improve code design/ability to debug? It's not tied to a specific problem.

dnephin:
Yup, exactly

mnowster:
Got it :) 

If you make them separate commits and then the PR description is the first paragraph that'd be helpful.

mnowster:
Just needs a rebase and then I'll merge.

dnephin:
rebased

PR #2208, opened 2015-10-16
Fix check for tags in release script `make-branch`

dnephin:
Ran into another bug while starting to cherry-pick some branches into rc2.

I'm not sure if this worked if there isn't a branch with the same name, but if there is a branch with the same name it fails without the `--tags`.  This is only a problem when we're branching from a git tag (which is why I didn't see it in the previous round of fixes).

### DIFF
--- a/script/release/make-branch
+++ b/script/release/make-branch
@@ -46,7 +46,7 @@ if [ -z "$REMOTE" ]; then
 fi
 
 # handle the difference between a branch and a tag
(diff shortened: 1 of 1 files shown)


mnowster:
LGTM. Could you copy some of the context of _why_ from the PR message and include it in the commit please? It's valuable info.

dnephin:
Added more context to the commit message.

PR #2209, opened 2015-10-16
Give the user a better error message when the yaml file is invalid

dnephin:
reported here: http://stackoverflow.com/questions/33114202/docker-compose-not-working-on-debian-8-2

Couldn't find an existing issue for it.

If there are yaml parsing errors the user would get a stacktrace, so the important details (line and column number with the error) were kind of hard to see.

This fixes the issue by removing the stacktrace and only printing the relevant error message:

```
yaml.scanner.ScannerError: mapping values are not allowed here
  in "./examples/invalid-yaml/dc.yml", line 3, column 20
```

### DIFF
--- a/tests/unit/config/config_test.py
+++ b/tests/unit/config/config_test.py
@@ -5,6 +5,7 @@ import shutil
 import tempfile
 from operator import itemgetter
 
+import py
 import pytest
 
 from compose.config import config
@@ -348,6 +349,18 @@ class ConfigTest(unittest.TestCase):
                     'filename.yml'
                 )
             )
+
+    def test_load_yaml_with_yaml_error(self):
+        tmpdir = py.test.ensuretemp('invalid_yaml_test')
+        invalid_yaml_file = tmpdir.join('docker-compose.yml')
+        invalid_yaml_file.write("""
+            web:
+              this is bogus: ok: what
+        """)
+        with pytest.raises(ConfigurationError) as exc:
+            config.load_yaml(str(invalid_yaml_file))
+
+        assert 'line 3, column 32' in exc.exconly()
 
 
 class InterpolationTest(unittest.TestCase):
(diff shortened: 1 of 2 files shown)


tests/unit/config/config_test.py
```diff
@@ -349,6 +350,18 @@ def test_config_invalid_environment_dict_key_raises_validation_error(self):
...
 
+    def test_load_yaml_with_yaml_error(self):
+        tmpdir = py.test.ensuretemp('invalid_yaml_test')
+        invalid_yaml_file = tmpdir.join('docker-compose.yml')
+        invalid_yaml_file.write("""
+            web:
+              this is bogus: ok: what
+        """)
+        with pytest.raises(ConfigurationError) as exc:
+            config.load_yaml(str(invalid_yaml_file))
```
mnowster: What's the context behind doing this rather than having a 2 line fixture file?

tests/unit/config/config_test.py
```diff
@@ -349,6 +350,18 @@ def test_config_invalid_environment_dict_key_raises_validation_error(self):
                 )
             )
 
+    def test_load_yaml_with_yaml_error(self):
+        tmpdir = py.test.ensuretemp('invalid_yaml_test')
+        invalid_yaml_file = tmpdir.join('docker-compose.yml')
+        invalid_yaml_file.write("""
```
mnowster: What's the context behind doing this rather than having a 2 line fixture file?

dnephin: The test is a lot more clear when you can see the contents of the file inline with the test. You don't have to go digging around for the file to see what it's testing.

It also prevents someone else from incorrectly using this file for another test case.

An invalid `yaml` file would also cause the `pre-commit` check to fail. We could go adding exceptions for these, but it's unnecessary if we inline it.

I can see the advantage to a separate file if the file is large enough that editing it inline is annoying. For something this small I don't see any advantage to having a separate file.

mnowster:
LGTM.

PR #2210, opened 2015-10-16
Properly validate files when multiple files are used

### LINKED ISSUE #2203: Stack trace when an override file is empty
If I create an empty `docker-compose.override.yml` in a project, `docker-compose up` throws the following error:

```
  File "/compose/build/docker-compose/out00-PYZ.pyz/compose.cli.main", line 53, in main
  File "/compose/build/docker-compose/out00-PYZ.pyz/compose.cli.docopt_command", line 23, in sys_dispatch
  File "/compose/build/docker-compose/out00-PYZ.pyz/compose.cli.docopt_command", line 26, in dispatch
  File "/compose/build/docker-compose/out00-PYZ.pyz/compose.cli.main", line 163, in perform_command
  File "/compose/build/docker-compose/out00-PYZ.pyz/compose.cli.command", line 54, in project_from_options
  File "/compose/build/docker-compose/out00-PYZ.pyz/compose.cli.command", line 90, in get_project
  File "/compose/build/docker-compose/out00-PYZ.pyz/compose.config.config", line 215, in load
  File "/compose/build/docker-compose/out00-PYZ.pyz/compose.config.config", line 205, in merge_services
TypeError: 'NoneType' object is not iterable
```

We should instead show the same `Top level object needs to be a dictionary` error that you get if your `docker-compose.yml` file is empty.

dnephin:
Fixes #2203 

Removed the single-use decorators so the functionality can be used directly as a function.

### DIFF
--- a/compose/config/config.py
+++ b/compose/config/config.py
@@ -14,7 +14,6 @@ from .validation import validate_against_fields_schema
 from .validation import validate_against_service_schema
 from .validation import validate_extended_service_exists
 from .validation import validate_extends_file_path
-from .validation import validate_service_names
 from .validation import validate_top_level_object
 
 
@@ -165,16 +164,6 @@ def find_candidates_in_parent_dirs(filenames, path):
     return (candidates, path)
 
 
-@validate_top_level_object
-@validate_service_names
-def pre_process_config(config):
-    """
-    Pre validation checks and processing of the config file to interpolate env
-    vars returning a config dict ready to be tested against the schema.
-    """
-    return interpolate_environment_variables(config)
-
-
 def load(config_details):
     """Load the configuration from a working directory and a list of
     configuration files.  Files are loaded in order, and merged on top
@@ -194,7 +183,7 @@ def load(config_details):
         return service_dict
 
     def load_file(filename, config):
-        processed_config = pre_process_config(config)
+        processed_config = interpolate_environment_variables(config)
         validate_against_fields_schema(processed_config)
         return [
             build_service(filename, name, service_config)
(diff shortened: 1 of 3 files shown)


compose/config/config.py
```diff
@@ -165,13 +164,11 @@ def find_candidates_in_parent_dirs(filenames, path):
     return (candidates, path)
 
 
-@validate_top_level_object
-@validate_service_names
```
mnowster: You've removed `validate_service_names` without putting it back in elsewhere. Was this on purpose? You still need this validation to happen, validating a `top_level_object` as a dict on it's own isn't enough.

compose/config/config.py
```diff
@@ -165,13 +164,11 @@ def find_candidates_in_parent_dirs(filenames, path):
...
 def pre_process_config(config):
-    """
-    Pre validation checks and processing of the config file to interpolate env
+    """Pre validation checks and processing of the config file to interpolate env
     vars returning a config dict ready to be tested against the schema.
     """
+    validate_top_level_object(config)
```
mnowster: Given the direction you're going with this approach, I don't think it makes sense to have this being called as part of `pre_process_config`. As you've already called `validate_top_level_object` for each config file, so there's no need to include it in this function and call it again.

compose/config/config.py
dnephin: I moved it into `validate_top_level_object()` so they're always called together now.

compose/config/config.py
dnephin: Ya, i was going back and forth on this one. This is also called from the `extends` resolving.  I was considering just dropping this method and calling both functions. I think I'll do that.

compose/config/config.py
mnowster: dammit, sorry, I missed that.

compose/config/config.py
mnowster: Yeh, I think given how the code has changed that it doesn't fit into the model of how things work now and the different places it needs to be called. So I'm +1 with it being dropped.

dnephin:
Removed `pre_process_config()`

mnowster:
BOOM. 

LGTM>

PR #2212, opened 2015-10-17
fix problem with bash completion in old bash

albers:
There is a problem with bash completion running in old bash versions. These versions are still shipped with current MacBooks. See #16348 for the corresponding Docker issue and its details.

As docker-compose bash completion also uses the problematic bash feature, this PR applies the same fix as used in Docker, see #16348.

Steps to reproduce the problem:
On a MacBook, completion of `docker-compose run -e` will result in

    docker-compose run -e bash: compopt: command not found

Likewise, completion of `docker-compose scale` will yield

    docker-compose scale bash: compopt: command not found

### DIFF
--- a/contrib/completion/bash/docker-compose
+++ b/contrib/completion/bash/docker-compose
@@ -17,6 +17,12 @@
 #    . ~/.docker-compose-completion.sh
 
 
+# suppress trailing whitespace
+__docker_compose_nospace() {
+	# compopt is not available in ancient bash versions
+	type compopt &>/dev/null && compopt -o nospace
+}
+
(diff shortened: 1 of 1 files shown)


albers:
This is a minor issue, no urgent need to include in upcoming release.

thaJeztah:
Thanks @albers, tested and LGTM

thaJeztah:
> This is a minor issue, no urgent need to include in upcoming release.

Still would be good though, because it's not the best experience if the completion is broken :+1:

mnowster:
LGTM. Thanks @albers :sparkles:

PR #2213, opened 2015-10-17
Trim Dockerfile

viranch:
Pipe curl's download directly to extract/execute program to reduce number of commands and unnecessary removal of downloaded files.


dnephin:
LGTM

mnowster:
LGTM

PR #2215, opened 2015-10-18
Rely on 'auto' for DEFAULT_API_VERSION

rgbkrk:
docker-py supports this directly

### DIFF
--- a/compose/cli/docker_client.py
+++ b/compose/cli/docker_client.py
@@ -9,7 +9,7 @@ from ..const import HTTP_TIMEOUT
 log = logging.getLogger(__name__)
 
 
-DEFAULT_API_VERSION = '1.19'
+DEFAULT_API_VERSION = 'auto'
 
 
 def docker_client(version=None):


rgbkrk:
I'm assuming you'd want me to [set the `COMPOSE_API_VERSION` environment variable instead](https://github.com/docker/compose/blob/26ff7296969a38ffa61b548b1d2d0eeb08e0264a/docs/reference/overview.md#compose_api_version) or to always be upgraded. Is there any easy way to map from API version to Compose verson so I can install this into a virtualenv (instead of what I get from docker toolbox)?

dnephin:
We've discussed a change like this in the past, but we're hesitant to make `auto` the default for two reasons:
* it adds an extra api call to every command
* it can might it more difficult to debug failures, if compose is used with an older version of the engine that doesn't support a required feature. (which is why we have this long warning here https://github.com/docker/compose/blob/master/docs/reference/overview.md#compose_api_version)

It's probably worth keeping the discussion open, but I'm not sure if we can make this change

dnephin:
> Is there any easy way to map from API version to Compose verson

I don't think so. It largely depends on what features require from the API.  We generally try to support the latest two versions of  the API. So in the new 1.5.0-rc1 we're supporting 1.20 and 1.21.

rgbkrk:
Yeah, I was just in a case of where the API mismatched on one level (1.20 vs 1.19 IIRC), so I wondered what I needed to pin compose to. Seems like the env var will do the trick in the future.

It's too bad the call to `/version` can't happen on the first run during compose so it's not a call for everytime.

PR #2218, opened 2015-10-18
Clean orphan containers

### LINKED ISSUE #1477: Stop containers that have been removed from docker-compose.yml
Apologies if this is covered somewhere, but I wasn't able to find anything.

The new `--x-smart-recreate` is working really nicely for changes to running containers and new containers, but if I remove a container from container from the YAML definition it just stays running.

It'd be fantastic to support automatic removal if that's at all possible.

cmehay:
This pull request adds to compose a `clean` command to clean-up orphaned containers when `docker-compose.yml` file is edited.

The first commit adds `clean` command which provides two options to stop and remove orphan containers. Using `--keep` only stop running orphaned containers, `--force` will kill and remove containers without confirmation.

`ps` gets the new flag `--all` to list orphaned containers, output can be filtered using service names. Service names will match orphan containers if `--all` flag is present.

`up` gets the new flag `--clean` to kill orphaned container. This flag will not remove orphaned containers, only `clean` command will.

The second commit adds unit tests about those changes.

Please tell me if any of my edits may alter any good practice about the project.

This is my very first participation on Github to a big project like compose. I really hope I didn't make any mistake in the pull request procedure. Please be indulgent if any :)

Fixes #1477

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -146,6 +146,7 @@ class TopLevelCommand(DocoptCommand):
       start              Start services
       stop               Stop services
       unpause            Unpause services
+      clean              Remove orphan containers
       up                 Create and start containers
       migrate-to-labels  Recreate containers to add labels
       version            Show the Docker-Compose version information
@@ -265,12 +266,42 @@ class TopLevelCommand(DocoptCommand):
         Usage: ps [options] [SERVICE...]
 
         Options:
-            -q    Only display IDs
+            -q          Only display IDs
+            -a, --all   Display orphan containers
         """
-        containers = sorted(
-            project.containers(service_names=options['SERVICE'], stopped=True) +
-            project.containers(service_names=options['SERVICE'], one_off=True),
-            key=attrgetter('name'))
+        orphan = options.get('--all', False)
+        containers = []
+        services = [[s] for s in options['SERVICE']] \
+            if len(options['SERVICE']) else [[]]
+
+        for service in services:
+            try:
+                # We should bypass NoSuchService exception to handle
+                # orphan filtering
+                containers = sorted(containers +
+                                    project.containers(service_names=service,
+                                                       stopped=True) +
+                                    project.containers(service_names=service,
+                                                       one_off=True),
+                                    key=attrgetter('name'))
+            except NoSuchService:
+                if not orphan:
+                    project.validate_service_names(options['SERVICE'])
+        nb_containers = len(containers)
+
+        if orphan:
+            containers = containers + sorted(
+                project.containers(service_names=options['SERVICE'],
+                                   stopped=True,
+                                   orphan=True) +
+                project.containers(service_names=options['SERVICE'],
+                                   one_off=True,
+                                   orphan=True),
+                key=attrgetter('name'))
+
+        if len(containers) is 0:
+            # Now we are using validate_service_names to raise NoSuchService
+            project.validate_service_names(options['SERVICE'])
 
         if options['-q']:
             for container in containers:
@@ -283,14 +314,17 @@ class TopLevelCommand(DocoptCommand):
                 'Ports',
             ]
             rows = []
-            for container in containers:
+            for nb, container in enumerate(containers):
                 command = container.human_readable_command
+                state = container.human_readable_state
                 if len(command) > 30:
                     command = '%s ...' % command[:26]
+                if nb >= nb_containers:
+                    state = '%s (orphan)' % state
                 rows.append([
                     container.name,
                     command,
-                    container.human_readable_state,
+                    state,
                     container.human_readable_ports,
                 ])
             print(Formatter().table(headers, rows))
@@ -492,6 +526,38 @@ class TopLevelCommand(DocoptCommand):
         """
         project.unpause(service_names=options['SERVICE'])
 
+    def clean(self, project, options):
+        """
+        Kill and remove orphan containers.
+
+        If any container from a previous version of docker-compose.yml is running,
+        this container will be killed and removed.
+
+        If the container is not running, it will be removed.
+
+        Orphan containers are identified using labels.
+
+        To prevent Compose from removing containers, use the `--keep` flag.
+
+        Usage: clean [options]
+
+        Options:
+            --keep          Do not remove containers.
+            -f, --force     Don't ask to confirm removal.
+        """
+
+        orphan_containers = (project.containers(orphan=True, stopped=True) +
+                             project.containers(orphan=True, one_off=True))
+        keep = options.get('--keep', False)
+        if len(orphan_containers) is 0:
+            print("No orphan containers")
+        elif keep or options.get('--force'):
+            project.clean(keep=keep)
+        else:
+            print("Going to remove", list_containers(orphan_containers))
+            if yesno("Are you sure? [yN] ", default=False):
+                project.clean(keep=keep)
+
     def up(self, project, options):
         """
         Builds, (re)creates, starts, and attaches to containers for a service.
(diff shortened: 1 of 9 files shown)


compose/cli/main.py
```diff
@@ -140,6 +140,7 @@ class TopLevelCommand(DocoptCommand):
       start              Start services
       stop               Stop services
       unpause            Unpause services
+      clean              Remove orphan containers
```
dnephin: I'd like some feedback from others here. I worry that `clean` might be too generic and it won't be obvious from the name what it's supposed to do.

compose/cli/main.py
```diff
@@ -273,17 +301,13 @@ def ps(self, project, options):
...
-                rows.append([
-                    container.name,
-                    command,
-                    container.human_readable_state,
-                    container.human_readable_ports,
-                ])
+            rows = table(containers)
+            if orphan:
+                headers.append('Misc')
+                rows += table(orphan_containers, misc='orphan')
```
dnephin: I wonder if we need this. Thinking about `docker ps -a` it doesn't modify the columns, it just adds more rows.

I think we include this detail in the "State" column instead. Maybe as "Up (orphan)".

compose/cli/main.py
```diff
@@ -256,16 +257,43 @@ def ps(self, project, options):
...
+        orphan = options.get('--all', False)
         containers = sorted(
             project.containers(service_names=options['SERVICE'], stopped=True) +
             project.containers(service_names=options['SERVICE'], one_off=True),
             key=attrgetter('name'))
 
+        orphan_containers = sorted(
+            project.containers(stopped=True, orphan=True) +
+            project.containers(one_off=True, orphan=True),
+            key=attrgetter('name'))
```
dnephin: It would be good to avoid these extra API calls when `orphan` is `False`.  How about something like this:

```python

if orphan:
    containers = sorted(
        containers +
        project.containers(...) + 
        project.containers(...),
        key=attrgetter('name'))
```

I think keeping them in the same list will simplify the `-q` case as well.

compose/project.py
```diff
@@ -284,12 +284,31 @@ def build(self, service_names=None, no_cache=False, pull=False):
...
+            msg_index=lambda c: c.name,
+            msg="Killing"
+        )
+        if not keep:
+            parallel_execute(
+                objects=all_containers,
+                obj_callable=lambda c: c.remove(),
+                msg_index=lambda c: c.name,
+                msg="Removing"
+            )
```
dnephin: It would be good to use the same `parallel_execute()` for both tasks, but not entirely necessary.

compose/project.py
```diff
@@ -360,7 +389,7 @@ def matches_service_names(container):
                 self.service_names,
             )
 
-        return [c for c in containers if matches_service_names(c)]
+        return [c for c in containers if matches_service_names(c, orphan)]
```
dnephin: Since this is a nested function you actually already have access to `orphan` from the parent scope, and don't need to pass it in here.  You can drop the extra parameter to `matches_service_names()` as well.

dnephin:
Thanks for the contribution!

Since this is a relative large change (new subcommand, new feature) it may take a little while to get this reviewed and merged, but I think we should include some support for dealing with unmanaged containers.

compose/cli/main.py
cmehay: I did this to keep orphaned containers at the end of the list, I thought it was more clear to not mix up orphaned containers with active services, but I will change this.

cmehay:
Thanks for your review :)

I will fix the elements you pointed out.

cmehay:
I changed the behavior of `ps --all` sub command.

`orphan` is added in State collum as you suggested out, and orphaned containers can be filtered using  service names. So using `--all` flag won't be raise an exception if an orphaned container with a missing service name exists.

The behavior `ps` without `--all` flag is untouched.

ps: I just figured out that I forgot to add multiple service names in unit test for `ps` sub command. Sorry about that, I will amend my commit latter to add them.

thaJeztah:
Perhaps this should just be a flag on `rm` in stead of a subcommand, i.e. `docker-compose rm --clean`. I'm not sure I would expect the `--clean` flag to be on `docker-compose up

(just my thoughts)

dnephin:
@thaJeztah I think that's a good point. We may not need a whole new subcommand for this. Adding `--include-orphaned` to `stop` and `rm` would be enough.

For `up` I think the rational is that `up` is supposed to update the composition to the new state defined in the compose file. If services are removed, they are just lost. So I can see an argument for `up` having a flag to remove orphaned containers as well.

@cmehay what do you think about dropping the new subcommand and doing this with flags to `stop`, `rm`, and `up` (which you already have) ?

cmehay:
@dnephin @thaJeztah I thing it would be much safer to add a sub command for this task. I mean, is `rm --include-orphaned` will kill all my containers including orphaned one ? I don't think this is what user will expect.

By the way, if we add a flag to `rm` or `stop` for orphaned containers, a managed service could be killed or removed accidentally, using clean command is much safer IMHO.

Finally, for `up` command, as you said, I see `up` as idempotent command which set services as they are described in yml file, so `--clean` flag for this command is useful I think

thaJeztah:
@cmehay `rm` doesn't remove running containers, so in that respect it's safe(-ish), unless data volumes are involved of course.

(Thinking out loud here) Perhaps we should have a similar approach to how docker handles "dangling" images and volumes; i.e., something like `docker-compose rm $(docker-compose ps -a --filter orphaned=true)`

Otoh, this may become too complicated for users, idk

thaJeztah:
For reference; there has been quite some discussion around implementing a `docker clean` command; https://github.com/docker/docker/issues/928

cmehay:
I updated unit test and `ps` method to fix behavior when results are filtered using service names.

I'm not very happy of what I've done in this method, I tricked with exceptions to list every containers, but the behavior of `ps --all` is now consistent.

cmehay:
@thaJeztah regarding the discussion about `docker clean`, the difficulty with `docker` is that very difficult to determine which containers should be removed or not using this sub command, this is probably why they didn't find a solution to this problem.

Regarding `compose`, we know exactly which containers will be removed, so this would be easier to find a solution :smile:

jberkus:
I can't comment on the implementation, but I could really use this feature.  When I'm testing a new docker-compose file, I often end up with dozens of orphan, stopped containers.

cmehay:
I would like to say something else to defend the idea of subcommand. I went to a docker meetup with Solomon Hykes few weeks ago, and he said that the important thing of docker is that docker lets people do the way they want to do.

I agree with that, people use containers many ways, so I think compose should provide a way to clean unmanaged containers the safer way as possible for the many ways docker is used.

Adding flag to existing subcommands if not very safe as these subcommands can alter managed containers, it would be harder to filter services, should display ambiguous informations about containers, and be result legit containers to be stopped or removed...

`clean` subcommand will never remove managed containers, it will remove only containers from services which has been removed from `docker-compose.yml` file, for instance it can be used safely if you're using exited containers.

I really hope people share my point of view, and please let me know if you want me to improve anything in the current implementation.

BTW, I'm really sorry about my English, I really hope this is not too painful to read :bow:

thaJeztah:
> BTW, I'm really sorry about my English, I really hope this is not too painful to read :bow:

No worries there, @cmehay, had no problem whatsoever to read! Many (if not most) contributors don't have English as their primary language (including myself) :smile:

cmehay:
The code has been rebased from master.

The unit test of `clean` subcommand seems to randomly fail due to the latency of exiting container. The test seems to be performed before the third container which run `true` cmd is actually exited.

Adding `time.sleep(1)` in the test should be enough but I wonder if this is acceptable...

I can also change the test to not test with exited container, as you prefer.

compose/project.py
cmehay: I'm not sure to know how to do this :confused:

compose/cli/main.py
cmehay: Is `cleanup` would be better?

cmehay:
New paradigm on containers and services is driving this MR obsolete.

PR #2223, opened 2015-10-19
Powershell script to run compose in a container.

au-phiware:
As requested by @dnephin [here](https://github.com/docker/compose/issues/1085#issuecomment-146599493) (sorry it took so long).

This script assumes the typical environment that Windows users operate,
namely, VirtualBox running the boot2docker ISO, managed by
docker-machine. I wrote this script for my Windows using colleagues and
first placed it in the public domain as a gist:
https://gist.github.com/au-phiware/25213e72c80040f398ba
In short, that script works for me.

I have adapted that script to use the (yet to be) official image
(docker/compose:latest) and also added an extra environment variable to
provide additional options to docker run.

### DIFF
--- a/script/run.ps1
+++ b/script/run.ps1
@@ -1,0 +1,22 @@
+# Run docker-compose in a container via boot2docker.
+#
+# The current directory will be mirrored as a volume and additional
+# volumes (or any other options) can be mounted by using
+# $Env:DOCKER_COMPOSE_OPTIONS.
+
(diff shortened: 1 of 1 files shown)


dnephin:
Thanks for submitting this as a PR!

thaJeztah:
Sweet! Thanks @au-phiware!

mnowster:
LGTM. Thanks for doing this @au-phiware

dnephin:
LGTM

It doesn't yet support all the same options as the linux/osx bash version, but that's ok for now

PR #2232, opened 2015-10-20
WIP: Upgrade to api 1.20

### LINKED ISSUE #2131: Update compose for remote API version 1.20
It looks like there were a few backwards incompatibilities in 1.20.  It appears that swarm doesn't support older versions of the API, so if we want to run against swarm, we need to support the latest API.

dnephin:
**Don't merge** I'd like to wait until we're done with the 1.5.0 release so that we don't run into conflicts while cherry-picking.  Opening the PR now to verify against jenkins that the work is done.

Fixes #2131 

Upgrade our tests and api interaction to be compatible with api version 1.20.

### DIFF
--- a/compose/cli/docker_client.py
+++ b/compose/cli/docker_client.py
@@ -9,7 +9,7 @@ from ..const import HTTP_TIMEOUT
 log = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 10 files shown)


dnephin:
This was carried in #2421

Issue #1910, opened 2015-08-24
Stable Compose / Swarm integration


bfirsh:
It should be possible to point Compose at Swarm and have it run any Compose app that runs on a single Engine.

This is also being tracked on Swarm's side here: https://github.com/docker/swarm/issues/1156
### Sub-issues
- [ ] Run the Compose test suite against a Swarm cluster
- [x] https://github.com/docker/swarm/issues/1244
- [x] https://github.com/docker/compose/issues/1913
- [ ] https://github.com/docker/compose/issues/1528
- [ ] #2131
- [x] https://github.com/docker/swarm/issues/1257
- [ ] #2169
- [x] https://github.com/docker/swarm/issues/1293
- [x] https://github.com/docker/swarm/issues/1316

dnephin:
Related issues: #1913, #1528

I think step 1 here might be "run the compose test suite against a swarm cluster".

bfirsh:
:+1: thanks

dnephin:
#1090 is related, but might already be fixed

dnephin:
This is blocked on https://github.com/docker/swarm/issues/1244

dnephin:
Possible #2374

dnephin:
I've created a new swarm label: https://github.com/docker/compose/labels/swarm to track related issues, instead of having to come update the checklist in this issue.

bfirsh:
Wow what a throwback. Hi everyone. 😂

PR #2233, opened 2015-10-20
Cleanup some unit tests and whitespace

dnephin:
While I was working on #1910 I did a bit of cleanup. The branch for swarm compatibility is already quite large,  so I pulled this out as a separate branch

* Remove some unnecessary newlines
* Remove a unittest that was attempting to test behaviour that was removed a while ago
* Updated some unit tests to use mocks instead of a custom fake


tests/unit/service_test.py
```diff
@@ -351,44 +351,37 @@ def test_parse_repository_tag(self):
...
-            self.assertEqual('someimage', repo)
-            self.assertEqual('latest', tag)
-            images.append({'Id': 'abc123'})
-            return []
-
-        service.image = lambda *args, **kwargs: mock_get_image(images)
-        self.mock_client.pull = pull
-
-        service.create_container()
-        self.assertEqual(1, len(images))
```
dnephin: This is already tested as `test_pull_image_no_tag()`

mnowster:
LGTM

PR #2236, opened 2015-10-21
Use inspect network to query for an existing network

dnephin:
And more tests for get_network()

Also changed the version check to use an api version:
* because the tests have to reference the api version anyway to do inspection
* using an engine version breaks against swarm, and we want to be able to run against swarm


mnowster:
LGTM>

Issue #1756, opened 2015-07-22
"up" displays output from previous runs


bfirsh:
When an app has been run before, `up` displays heaps of output from previous runs of the containers:

```
$ docker-compose up
Starting composedemo_redis_1...
Starting composedemo_web_1...
Attaching to composedemo_redis_1, composedemo_web_1
web_1   |  * Running on http://0.0.0.0:5000/ (Press CTRL+C to quit)
web_1   |  * Restarting with stat
redis_1 | 1:M 20 Jul 11:29:44.582 # User requested shutdown...
web_1   |  * Running on http://0.0.0.0:5000/ (Press CTRL+C to quit)
redis_1 | 1:M 20 Jul 11:29:44.582 * Saving the final RDB snapshot before exiting.
web_1   |  * Restarting with stat
redis_1 | 1:M 20 Jul 11:29:44.586 * DB saved on disk
redis_1 | 1:M 20 Jul 11:29:44.586 # Redis is now ready to exit, bye bye...
redis_1 | 1:C 20 Jul 11:31:03.505 # Warning: no config file specified, using the default config. In order to specify a config file use redis-server /path/to/redis.conf
redis_1 |                 _._
redis_1 |            _.-``__ ''-._
redis_1 |       _.-``    `.  `_.  ''-._           Redis 3.0.2 (00000000/0) 64 bit
redis_1 |   .-`` .-```.  ```\/    _.,_ ''-._
redis_1 |  (    '      ,       .-`  | `,    )     Running in standalone mode
redis_1 |  |`-._`-...-` __...-.``-._|'` _.-'|     Port: 6379
redis_1 |  |    `-._   `._    /     _.-'    |     PID: 1
redis_1 |   `-._    `-._  `-./  _.-'    _.-'
redis_1 |  |`-._`-._    `-.__.-'    _.-'_.-'|
redis_1 |  |    `-._`-._        _.-'_.-'    |           http://redis.io
redis_1 |   `-._    `-._`-.__.-'_.-'    _.-'
redis_1 |  |`-._`-._    `-.__.-'    _.-'_.-'|
redis_1 |  |    `-._`-._        _.-'_.-'    |
redis_1 |   `-._    `-._`-.__.-'_.-'    _.-'
redis_1 |       `-._    `-.__.-'    _.-'
redis_1 |           `-._        _.-'
redis_1 |               `-.__.-'
redis_1 |
[...]

PR #2254, opened 2015-10-26
1756 dont display output from previous runs

mnowster:
This approach fixes to prevent `up` from displaying output of previous runs.

This approach is minimal viable changes to get the desired fix in with minimum disruption to rest of the structure of the code and ensures our users get a fix for 1.5 rather than having to wait until 1.6 when we'll tackle more of the logging features.

Fixes https://github.com/docker/compose/issues/1756

### DIFF
--- a/compose/cli/log_printer.py
+++ b/compose/cli/log_printer.py
@@ -73,9 +73,13 @@ def build_no_log_generator(container, prefix, color_func):
 
 
 def build_log_generator(container, prefix, color_func):
-    # Attach to container before log printer starts running
-    stream = container.attach(stdout=True, stderr=True,  stream=True, logs=True)
-    line_generator = split_buffer(stream)
+    # if the container doesn't have a log_stream we need to attach to container
+    # before log printer starts running
+    if container.log_stream is None:
+        stream = container.attach(stdout=True, stderr=True,  stream=True, logs=True)
+        line_generator = split_buffer(stream)
+    else:
+        line_generator = split_buffer(container.log_stream)
 
     for line in line_generator:
         yield prefix + line
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -565,16 +565,18 @@ class TopLevelCommand(DocoptCommand):
         start_deps = not options['--no-deps']
         service_names = options['SERVICE']
         timeout = int(options.get('--timeout') or DEFAULT_TIMEOUT)
+        detached = options.get('-d')
 
         to_attach = project.up(
             service_names=service_names,
             start_deps=start_deps,
             strategy=convergence_strategy_from_opts(options),
             do_build=not options['--no-build'],
-            timeout=timeout
+            timeout=timeout,
+            detached=detached
         )
 
-        if not options['-d']:
+        if not detached:
             log_printer = build_log_printer(to_attach, service_names, monochrome)
             attach_to_logs(project, log_printer, service_names, timeout)
 
@@ -636,7 +638,10 @@ def convergence_strategy_from_opts(options):
 
 def build_log_printer(containers, service_names, monochrome):
     if service_names:
-        containers = [c for c in containers if c.service in service_names]
+        containers = [
+            container
+            for container in containers if container.service in service_names
+        ]
     return LogPrinter(containers, monochrome=monochrome)
 
 
--- a/compose/project.py
+++ b/compose/project.py
@@ -290,7 +290,8 @@ class Project(object):
            start_deps=True,
            strategy=ConvergenceStrategy.changed,
            do_build=True,
-           timeout=DEFAULT_TIMEOUT):
+           timeout=DEFAULT_TIMEOUT,
+           detached=False):
 
         services = self.get_services(service_names, include_deps=start_deps)
 
@@ -308,7 +309,8 @@ class Project(object):
             for container in service.execute_convergence_plan(
                 plans[service.name],
                 do_build=do_build,
-                timeout=timeout
+                timeout=timeout,
+                detached=detached
             )
         ]
 
--- a/compose/service.py
+++ b/compose/service.py
@@ -435,17 +435,6 @@ class Service(object):
         else:
             raise Exception("Invalid action: {}".format(action))
 
-    def _recreate_stop_container(self, container, timeout):
-        try:
-            container.stop(timeout=timeout)
-        except APIError as e:
-            if (e.response.status_code == 500
-                    and e.explanation
-                    and 'no such process' in str(e.explanation)):
-                pass
-            else:
-                raise
-
     def recreate_container(self,
                            container,
                            timeout=DEFAULT_TIMEOUT,
@@ -458,7 +447,7 @@ class Service(object):
         """
         log.info("Recreating %s" % container.name)
 
-        self._recreate_stop_container(container, timeout)
+        container.stop(timeout=timeout)
         container.rename_to_tmp_name()
         new_container = self.create_container(
             do_build=False,
--- a/compose/service.py
+++ b/compose/service.py
@@ -395,11 +395,17 @@ class Service(object):
     def execute_convergence_plan(self,
                                  plan,
                                  do_build=True,
-                                 timeout=DEFAULT_TIMEOUT):
+                                 timeout=DEFAULT_TIMEOUT,
+                                 detached=False):
         (action, containers) = plan
+        should_attach_logs = not detached
 
         if action == 'create':
             container = self.create_container(do_build=do_build)
+
+            if should_attach_logs:
+                container.attach_log_stream()
+
             self.start_container(container)
 
             return [container]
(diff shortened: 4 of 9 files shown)


aanand:
Nice.

compose/service.py
```diff
@@ -448,11 +479,26 @@ def recreate_container(self,
             else:
                 raise
 
+    def _recreate_rename_container(self, container):
         # Use a hopefully unique container name by prepending the short id
         self.client.rename(
             container.id,
-            '%s_%s' % (container.short_id, container.name))
+            '%s_%s' % (container.short_id, container.name)
+        )
```
dnephin: I think the `Container` class would be a better place for this new function. It only uses `self.client` and `container` (which are both available in `Container`).

Maybe something like `def rename_to_tmp_name(self):` ?

compose/service.py
```diff
@@ -448,11 +479,26 @@ def recreate_container(self,
             else:
                 raise
```
dnephin: I'm pretty sure we don't need the try/except anymore (it can be just a plain `container.stop(timeout=timeout)`).

The api returns a 304 for already stopped containers which doesn't raise an error. Looks like it was added over a year ago b573b87a92d12bad82a41377e8cbaa891ac0dfb9. The API has been fixed since then.

If you don't feel comfortable removing the try/except in this PR, maybe just add a TODO for now and we can remove it later?

compose/cli/main.py
```diff
@@ -565,17 +565,25 @@ def up(self, project, options):
...
-        if not options['-d']:
-            log_printer = build_log_printer(to_attach, service_names, monochrome)
+        if not detached:
+            filtered_streams = [(container, thread) for container, thread in to_attach if thread is not None]
+            containers, log_streams = zip(*filtered_streams)
+            log_printer = LogPrinter(
+                list(containers),
+                monochrome=monochrome,
+                log_streams=list(log_streams)
+            )
```
dnephin: It looks like you removed `build_log_printer` but then reproduced most of the logic inline and didn't remove the function?

Could this be moved back into `build_log_printer` ?

compose/cli/log_printer.py
```diff
@@ -36,10 +37,18 @@ def no_color(text):
...
+            # threads attached to containers
+            for color_func, container, stream in zip(color_funcs, self.containers, self.log_streams):
+                generator_func = build_log_generator
+                prefix = color_func(build_log_prefix(container, prefix_width))
+                yield generator_func(container, prefix, color_func, stream)
+        else:
+            for color_func, container in zip(color_funcs, self.containers):
+                generator_func = get_log_generator(container)
+                prefix = color_func(build_log_prefix(container, prefix_width))
+                yield generator_func(container, prefix, color_func)
```
dnephin: I think a lot more of this code can be shared between the branches by making a small change to `get_log_generator()`. I was hoping to have a chance today to prototype that, but I've run out of time.

I think we can get it down to a single line in each if/else branch inside the for loop

compose/project.py
```diff
@@ -302,15 +303,17 @@ def up(self,
...
                 plans[service.name],
                 do_build=do_build,
-                timeout=timeout
+                timeout=timeout,
+                detached=detached
             )
-        ]
+            containers += converged_results
+
+        return containers
```
dnephin: This doesn't need to change, right? It's just an added parameter.  I think the compression version is more clear.

compose/service.py
```diff
@@ -395,49 +395,80 @@ def _containers_have_diverged(self, containers):
...
             container = self.create_container(do_build=do_build)
+
+            log_stream = None
+            if container.has_api_logs and not detached:
+                log_stream = container.attach(stdout=True, stderr=True, stream=True)
+
             self.start_container(container)
 
-            return [container]
+            return zip([container], [log_stream])
```
dnephin: It would be really awesome if we could keep the same return value. I think a lot of the complexity in this change is having to accommodate this difference.

Since our log_stream comes from a container, I think we could make it a method on Container.

```python

class Container(object):
    def __init__(self, ...):
       ...
       self._log_stream = None


    def init_log_stream(self):
        if not container.has_api_logs:
            return
        self._log_stream = self.attach(stdout=True, stderr=True, stream=True)

    def get_log_stream(self):
        return self._log_stream
```

It's unfortunate that it wouldn't be immutable , but Container isn't immutable as it is, so I guess that's not so bad.

In this function I think we'll be able to restore the returning of just a list of containers, and it can do:

````python
    if not detached:
        container.init_log_stream()
````

File: compose/service.py [hunk 7]
```diff
@@ -448,11 +479,26 @@ def recreate_container(self,
...
         # Use a hopefully unique container name by prepending the short id
         self.client.rename(
             container.id,
-            '%s_%s' % (container.short_id, container.name))
+            '%s_%s' % (container.short_id, container.name)
+        )
+
+    def recreate_container(self,
+                           container,
+                           timeout=DEFAULT_TIMEOUT):
```
dnephin: Instead of duplicating this, with my proposed change to add init_log_stream/get_log_stream to Container, I think we can just add an `init_log_stream=False` kwarg here, and set it to `True` from `execute_convergence_plan()` to remove the need for duplication.

compose/service.py
```diff
@@ -395,27 +395,57 @@ def _containers_have_diverged(self, containers):
...
-                                 timeout=DEFAULT_TIMEOUT):
+                                 timeout=DEFAULT_TIMEOUT,
+                                 detached=False):
         (action, containers) = plan
 
         if action == 'create':
             container = self.create_container(do_build=do_build)
+
+            if container.has_api_logs and not detached:
+                container.attach_log_stream()
```
dnephin: `attach_log_stream()` already checks for `has_api_logs()` (which I think is a good place to do it), so I think we can remove it from here and do just `if not detached:`

compose/service.py
```diff
@@ -395,27 +395,57 @@ def _containers_have_diverged(self, containers):
...
         elif action == 'start':
-            for c in containers:
-                self.start_container_if_stopped(c)
+            for container in containers:
+                if not container.is_running:
+                    if container.has_api_logs and not detached:
+                        container.attach_log_stream()
+
+                    log.info("Starting %s" % container.name)
+                    self.start_container(container)
```
dnephin: If we add a flag `attach_logs` to `start_container_if_stopped()` which calls this. I think this would be:

```python

for container in containers:
    self.start_container_if_stopped(container, attach_logs=not detached)
```

which removes the need to duplicate anything

compose/service.py
```diff
@@ -395,27 +395,57 @@ def _containers_have_diverged(self, containers):
...
+                # attach
+                if new_container.has_api_logs and not detached:
+                    new_container.attach_log_stream()
+                recreated_containers.append(new_container)
+
+                # restart
+                self.start_container(new_container)
+                container.remove()
+
+            return recreated_containers
```
dnephin: I think we can do the same thing here.

If we add a flag `attach_logs` to `recreate_container()` which calls `attach_log_stream()`. I think this would be:

```python

return [
    self.recreate_container(container, timeout=timeout, attach_logs=not detached)
    for container in containers
]
```

which removes the need to duplicate anything

dnephin:
This is looking great, thanks for making the changes

Just a couple smaller things to remove some of the duplicate and I think this will be ready

mnowster:
@dnephin ok, I think I've addressed all feedback. I'm just running all tests locally now.

Yeh by introducing the `attach_logs` flag enabled me to remove that duplication so good call thank you :+1:

compose/service.py
```diff
@@ -395,27 +395,35 @@ def _containers_have_diverged(self, containers):
...
+
+            if not detached:
+                container.attach_log_stream()
+
             self.start_container(container)
 
             return [container]
 
         elif action == 'recreate':
+            should_attach_logs = not detached
```
dnephin: This could be moved out of the if/elif/elif block since it's used in all three branches, but super minor, not a blocker

dnephin:
LGTM!

Thanks for getting this fixed, there will be many happy users!

compose/service.py
mnowster: ~Good point, done :)

mnowster:
@dnephin can confirm all tests pass on my local machine btw :sparkles:

dnephin:
tests pass locally

thaJeztah:
\o/ thanks @mnowster !

bfirsh:
woot!

szuliq:
Not sure if it was intended but attaching to a detached container still duplicates logs. 

Using `docker-compose==1.5.2` the following

    docker-compose up -d xxx
    docker-compose up xxx

prints all the logs from the backlog of `xxx` whereas

    docker-compose up xxx

when `xxx` is not running in the background does not.

When I say backlog, I mean since the container was created, not since it was last run "interactively".

I think I know why it's that way. Attaching probably is equivalent of `docker-compose logs xxx` with `Ctrl+C` binding.

Nevertheless my guts tell me the calls to `docker-compose up xxx` should behave in the same way regardless of the `xxx` running in the background or not.

Pomax:
Seeing a complete log-since-creation scrolling by again, was there a regression that nullified this change?

Issue #2248, opened 2015-10-23
Network aliases created with full container name & not short name


skelethan:
Per https://github.com/docker/compose/blob/0409b287ad3ede39282c199d611456aab15dca71/docs/networking.md

![image](https://cloud.githubusercontent.com/assets/9012349/10705720/92822f6a-79a5-11e5-8b02-813db4097ffe.png)

However, what I am seeing is that the aliases are created with the full container name from compose e.g. context_web_1

This eliminates the benefit of consistent aliases, as compose will dynamically assign network alias depending on number of containers & where they are run from.

My simple test below.
![image](https://cloud.githubusercontent.com/assets/9012349/10705780/2d89ddf0-79a6-11e5-81e8-9e335507ed3f.png)

![image](https://cloud.githubusercontent.com/assets/9012349/10705807/71bceea4-79a6-11e5-86d0-f54f05f3a9cd.png)

![image](https://cloud.githubusercontent.com/assets/9012349/10705776/21e4a638-79a6-11e5-9207-552c4df5df1a.png)

dnephin:
I'm able to reproduce this.

cc @aanand, @mavenugo

According to https://github.com/docker/compose/pull/1676#issuecomment-145562152 the container is supposed to join the network using the hostname, but that is not what's happening with docker engine 1.9.0-rc2.

The hostname is correctly set to the service name (ex: "web"), but it's joining the network as "project_web_1" (according to the /etc/hosts in other containers on the network). This is what I see:

```
172.18.0.2      networking_other_1
172.18.0.2      networking_other_1.networking
172.18.0.3      networking_web_1
172.18.0.3      networking_web_1.networking
```

Did this change? Do we need to change something on the compose side?

dnephin:
@skelethan thanks for reporting this. Unfortunately we weren't able to fix this for the current release so we've updated our documentation to match what is supported right now (the container joins the networking using the container name).

In the next release we'll have support for aliases which should support a much nicer short names.

Teudimundo:
Is there any workaround that could be used? Since addresses with `_` are not valid FQDN this bug can cause some pain.

dnephin:
@Teudimundo yup, [the docs](https://github.com/docker/compose/blob/master/docs/networking.md#configure-how-services-are-published) mention that you can set a custom container name with `container_name:`.

Teudimundo:
Thank you, i missed that.

pwaller:
[...]

PR #2258, opened 2015-10-26
Fix running one-off containers with --x-networking

dnephin:
Found this while debugging #2248 

We need to disable linking to self when we run with networking enabled.

`run` sets up links from the one-off container to the "canonical" container, which breaks when used with `--x-networking`. The link to the canonical container it's not something the user has asked for, it's just something we do automatically.

In this case it seems appropriate to not create the link at all, since it will work the same way without the link.

### DIFF
--- a/compose/service.py
+++ b/compose/service.py
@@ -539,6 +539,9 @@ class Service(object):
         return 1 if not numbers else max(numbers) + 1
 
     def _get_links(self, link_to_self):
+        if self.use_networking:
+            return []
+
         links = []
         for service, link_name in self.links:
             for container in service.containers():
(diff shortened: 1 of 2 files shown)


dnephin:
The only test failure is https://github.com/docker/docker/issues/15995

compose/service.py
```diff
@@ -540,6 +540,10 @@ def _next_container_number(self, one_off=False):
 
     def _get_links(self, link_to_self):
         links = []
+
+        if self.use_networking:
```
mnowster: Minor: Could you put this line before the `links = []` line. It helps express the intent a little clearer that if we're using networking that `_get_links` is to do nothing.

shin-:
LGTM

compose/service.py
dnephin: will do

dnephin:
done

mnowster:
LGTM. Tests pass on my machine.

Issue #1322, opened 2015-04-22
Support ulimit options


divideby0:
Now that Docker 1.6 supports ulimit options, it would be nice to have this as an option for containers managed by compose, similar to `mem_limit`, `cpu_shares`, etc.

janober:
+1

aanand:
Both are already supported: http://docs.docker.com/compose/yml/#working95dir-entrypoint-user-hostname-domainname-mem95limit-privileged-restart-stdin95open-tty-cpu95shares

divideby0:
To clarify, `mem_limit` and `cpu_shares` are supported but it doesn't appear that `ulimit` is supported yet.  Ulimit options were only recently added in the Docker 1.6 engine:

https://blog.docker.com/2015/04/docker-release-1-6/

aanand:
Sorry, I misread. Yes, `ulimit` support would be good. Looks like docker-py added support in https://github.com/docker/docker-py/pull/554, so it should be simple to do.

corbinu:
+1 need for couchbase/server image

thaJeztah:
Also worth noting that docker recently added support for ulimit to `docker build` as well. I wonder if compose should apply the limits both for build and run?

tpires:
+1

lu4nm3:
+1

kfox1111:
+1

adrienkohlbecker:
+1

crquan:
+1

indigodp:
+1

trukhinyuri:
+1

Bondza:
+1

crquan:
+1

shanielh:
+1

rschmukler:
+1

KevinGreene:
Working on this now. The ulimits argument in docker-py takes a list of dictionaries, with the key "name", optional keys "soft" and "hard", e.g. `{'name': 'nofile', 'soft': 8096, 'hard': 16192}`

The expected YAML entry in a Docker compose file for this would be:

```yaml
ulimits:
  nofile:
    soft: 8096
    hard: 16192
```

Correct?

divideby0:
That sounds good to me, though I'm not certain if the existing YAML parse supports deeply nested values. If no, I think the alternative would also be acceptable:

```yaml
ulimits:
  nofile: "8096:16912"
```

Note that the latter syntax more closely matches what's actually provided to the Docker daemon, for example:

```bash
[...]

PR #2261, opened 2015-10-27
Added ulimits to docker-compose

KevinGreene:
I added ulimits functionality as outlined in #1322, but I have a few style concerns.

### How to handle error checking

As far as I could tell, this was the most nested config option, so I wasn't sure the appropriate amount of type checking to put into the function. You will currently get a docker-compose level error if you put in something that isn't a dict for the first level, or a dict or an int for the second level. It also handles putting in a key other than `soft` or `hard`. However, if you put in something like

```
ulimits:
  nofile:
    soft: ThisIsAnError
    hard: 60000
```

You get an error from docker-py, not docker-compose, that states `ValueError: Ulimit.soft must be an integer`.

### Should we handle `soft:hard` style of setting limits?

This isn't exposed in docker-py at all, but should we add the ability to say
```
ulimits:
  nofile: "8096:16192"
```

It's relatively easy to do, but I was worried about continuing to build up this relatively small feature. I'm not sure if docker-compose wants the kitchen sink when it comes to options.

If either of these should be changed, let me know and I can fix ASAP.

### DIFF
--- a/compose/config/fields_schema.json
+++ b/compose/config/fields_schema.json
@@ -116,6 +116,7 @@
         "security_opt": {"type": "array", "items": {"type": "string"}, "uniqueItems": true},
         "stdin_open": {"type": "boolean"},
         "tty": {"type": "boolean"},
+        "ulimits": {"type": "object"},
         "user": {"type": "string"},
         "volumes": {"type": "array", "items": {"type": "string"}, "uniqueItems": true},
         "volume_driver": {"type": "string"},
--- a/compose/config/fields_schema.json
+++ b/compose/config/fields_schema.json
@@ -116,7 +116,25 @@
         "security_opt": {"type": "array", "items": {"type": "string"}, "uniqueItems": true},
         "stdin_open": {"type": "boolean"},
         "tty": {"type": "boolean"},
-        "ulimits": {"type": "object"},
+        "ulimits": {
+          "type": "object",
+          "patternProperties": {
+            "^[a-z]+$": {
+              "oneOf": [
+                {"type": "integer"},
+                {
+                  "type":"object",
+                  "properties": {
+                    "hard": {"type": "integer"},
+                    "soft": {"type": "integer"}
+                  },
+                  "required": ["soft", "hard"],
+                  "additionalProperties": false
+                }
+              ]
+            }
+          }
+        },
         "user": {"type": "string"},
         "volumes": {"type": "array", "items": {"type": "string"}, "uniqueItems": true},
         "volume_driver": {"type": "string"},
--- a/compose/service.py
+++ b/compose/service.py
@@ -1085,11 +1085,6 @@ def build_ulimits(ulimit_config):
         if isinstance(soft_hard_values, six.integer_types):
             ulimits.append({'name': limit_name, 'soft': soft_hard_values, 'hard': soft_hard_values})
         elif isinstance(soft_hard_values, dict):
-            if not soft_hard_values['soft'] <= soft_hard_values['hard']:
-                raise ConfigError(
-                    "ulimit_config \"%s\" cannot contain a 'soft' value higher than 'hard' value" %
-                    ulimit_config
-                )
             ulimit_dict = {'name': limit_name}
             ulimit_dict.update(soft_hard_values)
             ulimits.append(ulimit_dict)
--- a/compose/service.py
+++ b/compose/service.py
@@ -1078,26 +1078,19 @@ def parse_restart_spec(restart_config):
 
 
 def build_ulimits(ulimit_config):
-    valid_keys = set(['hard', 'soft'])
-
     if not ulimit_config:
         return None
     ulimits = []
     for limit_name, soft_hard_values in six.iteritems(ulimit_config):
         if isinstance(soft_hard_values, six.integer_types):
-            ulimits.append({'Name': limit_name, 'soft': soft_hard_values, 'hard': soft_hard_values})
+            ulimits.append({'name': limit_name, 'soft': soft_hard_values, 'hard': soft_hard_values})
         elif isinstance(soft_hard_values, dict):
-            if not set(soft_hard_values) == valid_keys:
-                raise ConfigError(
-                    "ulimit_config \"%s\" must contain both 'hard' and 'soft' as secondary keys, and nothing else" %
-                    ulimit_config
-                )
             if not soft_hard_values['soft'] <= soft_hard_values['hard']:
                 raise ConfigError(
                     "ulimit_config \"%s\" cannot contain a 'soft' value higher than 'hard' value" %
                     ulimit_config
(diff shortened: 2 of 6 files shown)


compose/config/fields_schema.json
```diff
@@ -116,6 +116,7 @@
         "security_opt": {"type": "array", "items": {"type": "string"}, "uniqueItems": true},
         "stdin_open": {"type": "boolean"},
         "tty": {"type": "boolean"},
+        "ulimits": {"type": "object"},
```
dnephin: I think this is where most, or possibly all, the error checking should happen. Instead of just "object" you can specify the nested schema. There's a partial example of this with the environment key.

KevinGreene: Ah, that makes sense. I'll update this in < 12 hours.

dnephin:
Thanks for the PR!  I think this is looking good, I'll have another look over it tomorrow

KevinGreene:
@dnephin I migrated the ulimits schema checking to `fields_schema.json` instead of directly in the `service.py` file.

Let me know if it looks like anything else can be fixed / improved. Thanks!

tests/integration/service_test.py
```diff
@@ -161,6 +162,36 @@ def test_build_extra_hosts(self):
             {'www.example.com': '192.168.0.17',
              'api.example.com': '192.168.0.18'})
 
+    def test_build_ulimits(self):
+        def sort_ulimit_conf(ulimit_conf):
+            return sorted(ulimit_conf, key=lambda k: k['Name'])
+        # invalid options
```
mnowster: There's a bit too much going on in this test, can you split these out into separate tests please?  `test_build_ulimits_with_invalid_options`, `test_build_ulimits_with_dicts`, `test_build_ulimits_with_integers`

compose/config/fields_schema.json
```diff
@@ -116,6 +116,25 @@
...
+                    "hard": {"type": "integer"},
+                    "soft": {"type": "integer"}
+                  },
+                  "required": ["soft", "hard"],
+                  "additionalProperties": false
+                }
+              ]
+            }
+          }
+        },
```
dnephin: Schema looks good to me

compose/service.py
```diff
@@ -1069,6 +1071,28 @@ def parse_restart_spec(restart_config):
...
+# Ulimits
+
+
+def build_ulimits(ulimit_config):
+    if not ulimit_config:
+        return None
+    ulimits = []
+    for limit_name, soft_hard_values in six.iteritems(ulimit_config):
+        if isinstance(soft_hard_values, six.integer_types):
+            ulimits.append({'Name': limit_name, 'soft': soft_hard_values, 'hard': soft_hard_values})
```
dnephin: Minor: it looks like `docker-py` supports both `name` as well (and it's the first option it checks). Could we use `name` (lowercase) here to be consistent with the other fields?

script/test-versions
```diff
@@ -5,10 +5,6 @@
 set -e
 
 >&2 echo "Running lint checks"
-docker run --rm \
-  ${GIT_VOLUME} \
-  --entrypoint="tox" \
-  "$TAG" -e pre-commit
```
dnephin: I think this was removed by mistake?  We want to keep this linting check

KevinGreene: Yes, definitely by mistake. I have Tox running locally, but for some reason when running `script/test` tox failed to even start. I'll undo that and squash.

compose/service.py
KevinGreene: I agree, that makes more sense.

tests/integration/service_test.py
KevinGreene: Can do.

KevinGreene:
All of the most recent issues (Name -> name, undo removal of tox, and more split out tests) are now in the PR.

compose/service.py
```diff
@@ -1069,6 +1071,28 @@ def parse_restart_spec(restart_config):
...
+def build_ulimits(ulimit_config):
+    if not ulimit_config:
+        return None
+    ulimits = []
+    for limit_name, soft_hard_values in six.iteritems(ulimit_config):
+        if isinstance(soft_hard_values, six.integer_types):
+            ulimits.append({'name': limit_name, 'soft': soft_hard_values, 'hard': soft_hard_values})
+        elif isinstance(soft_hard_values, dict):
+            if not soft_hard_values['soft'] <= soft_hard_values['hard']:
+                raise ConfigError(
```
mnowster: I think you want to use a `ConfigurationError` here, that's the exception error we use. The `ConfigError` is a legacy one that we need to phase out.

mnowster: Wait, hang on, I'm in a different file. Sorry.

The checking of these values should go in as a separate `validate_ulimits` function over in `config.py` and raise a `ConfigurationError`. We want to raise configuration errors early and make them the concern of the config not the service.

compose/config/fields_schema.json
mnowster: Yep, schema looks right.

compose/service.py
KevinGreene: Just to make sure I start down the right path, you're saying adding that validation by calling a validate_ulimits function somewhere around https://github.com/docker/compose/blob/master/compose/config/config.py#L348, correct?

Is the ultimate goal to factor out all of the ConfigErrors from service.py?

dnephin: Yup, that's correct

dawidmalina:
Is there any chance that this feature will be released with 1.5 version of compose?

dnephin:
Unfortunately not, feature freeze for 1.5.0 was a few weeks ago.

dawidmalina:
When do you think it will be merged to master and released?

KevinGreene:
Is there anything more that is needed for this PR?

I see the Janky build is failing, but I'm not sure why. The error messages I see are ```Couldn't run auplink before unmount: exec: \"auplink\": executable file not found in $PATH"```, but I'm not sure how to go about debugging.

dnephin:
If you rebase again, we merged some improvements to the build yesterday that hide a bunch of the verbose logging.  The build should pass with the rebase on master.

dnephin:
LGTM (once build is green)

docs/compose-file.md
```diff
@@ -348,6 +348,17 @@ Override the default labeling scheme for each container.
         - label:user:USER
         - label:role:ROLE
 
+### ulimits
+
+Override the default ulimits for a container. You can either use a number
+to set the hard and soft limits, or specify them in a dictionary.
```
aanand: Wording's a bit strange here. Maybe "You can either specify a single limit as an integer or soft/hard limits as a dictionary".

dnephin:
I've rebased this in #2363 

Thanks for the contribution!

PR #2262, opened 2015-10-27
Added --force-rm to compose build.

### LINKED ISSUE #2253: docker-compose build leaves dangling containers on fail
With this Dockerfile
```dockerfile
FROM busybox

RUN exit 1
```

and docker-compose.yml
```yml
service:
  build: .
```

Running 
```bash
docker-compose build service
```
will fail because of the "RUN exit 1" and will leave in intermediate container there (which is completely unusable and useless).

There is an option for docker called --force-rm which fixes this but I see no option in compose for this.

Could there be one added?

Thanks!

adrian-budau:
Fixes #2253 

It's a flag passed to docker build that removes the intermediate
containers left behind on fail builds.

Signed-off-by: Adrian Budau <budau.adi@gmail.com>

### DIFF
--- a/compose/project.py
+++ b/compose/project.py
@@ -278,10 +278,10 @@ class Project(object):
         for service in self.get_services(service_names):
             service.restart(**options)
 
-    def build(self, service_names=None, no_cache=False, pull=False):
+    def build(self, service_names=None, no_cache=False, pull=False, force_rm=False):
         for service in self.get_services(service_names):
             if service.can_be_built():
-                service.build(no_cache, pull)
+                service.build(no_cache, pull, force_rm)
             else:
                 log.info('%s uses an image, skipping' % service.name)
 
--- a/contrib/completion/bash/docker-compose
+++ b/contrib/completion/bash/docker-compose
@@ -93,7 +93,7 @@ __docker_compose_services_stopped() {
 _docker_compose_build() {
 	case "$cur" in
 		-*)
-			COMPREPLY=( $( compgen -W "--help --no-cache --pull" -- "$cur" ) )
+			COMPREPLY=( $( compgen -W "--help --force-rm --no-cache --pull" -- "$cur" ) )
 			;;
 		*)
 			__docker_compose_services_from_build
--- a/tests/fixtures/simple-failing-dockerfile/Dockerfile
+++ b/tests/fixtures/simple-failing-dockerfile/Dockerfile
@@ -1,0 +1,7 @@
+FROM busybox:latest
+LABEL com.docker.compose.test_image=true
+LABEL com.docker.compose.test_failing_image=true
+# With the following label the container wil be cleaned up automatically
+# Must be kept in sync with LABEL_PROJECT from compose/const.py
+LABEL com.docker.compose.project=composetest
+RUN exit 1
--- a/tests/integration/cli_test.py
+++ b/tests/integration/cli_test.py
@@ -13,7 +13,9 @@ from compose.cli.command import get_project
 from compose.cli.docker_client import docker_client
 from compose.cli.errors import UserError
 from compose.cli.main import TopLevelCommand
+from compose.container import Container
 from compose.project import NoSuchService
+from compose.service import BuildError
 
 
 class CLITestCase(DockerClientTestCase):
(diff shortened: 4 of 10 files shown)


compose/project.py
```diff
@@ -278,10 +278,10 @@ def restart(self, service_names=None, **options):
         for service in self.get_services(service_names):
             service.restart(**options)
 
-    def build(self, service_names=None, no_cache=False, pull=False):
+    def build(self, service_names=None, force_rm=False, no_cache=False, pull=False):
         for service in self.get_services(service_names):
             if service.can_be_built():
-                service.build(no_cache, pull)
+                service.build(force_rm, no_cache, pull)
```
dnephin: Minor note here: If someone is using compose as a library (which we don't officially support), and they're incorrectly using positional args for kwargs (like we do here), adding the new kwarg to the beginning of the arg list will break code calling this function.

Adding it to the end of the arg list is a bit nicer.

tests/fixtures/simple-failing-dockerfile/Dockerfile
```diff
@@ -0,0 +1,3 @@
+FROM busybox:latest
+LABEL com.docker.compose.test_failing_image=true
```
dnephin: Please use the existing label (`com.docker.compose.test_image`) for this, so it will be cleaned up without you having to add any cleanup to each test.

dnephin: Oh, I see how this is being used.

I think you can include both labels

tests/integration/cli_test.py
```diff
@@ -155,6 +157,47 @@ def test_build_no_cache_pull(self):
...
+        ]
+
+        containers = [Container.from_ps(self.project.client, c) for c in self.project.client.containers(
+            all=True,
+            filters={"label": labels})]
+
+        # First remove the containers, because if the assertion fails, tearDown won't be able to clean up
+        container_count = len(containers)
+        for container in containers:
+            container.remove(force=True)
```
dnephin: To be a little more explicit about this (without needing a comment), please use [`self.addCleanup()`](https://docs.python.org/2/library/unittest.html#unittest.TestCase.addCleanup)

dnephin:
Thanks for the PR!

Just a couple minor requests, otherwise it's looking good.

adrian-budau:
Hi Daniel.

I'll try with addCleanup when I get home but I feel it won't solve anything
because during the teardown it tries to remove all images, and this will
error out because of the dangling containers.

I'll search for something similar addCleanup that is called before teardown

adrian-budau:
I added the test_image back and also  ```com.docker.compose.project=composetest ``` which allows it to automatically be cleaned on tearDown.

From what I can tell it's not finding some images in the janky tests, so the fail there does not seem my fault.

dnephin:
Yup, the janky error is not related. I've restarted it now

dnephin:
LGTM (once we get a green build)

adrian-budau:
Should I wait until master is green again or should I rebase against the last green master?

dnephin:
I'll restart the build again. We fixed the nodes, but there is still one flaky test we need to disable on 1.8.3.  I don't think you need to rebase again (just yet).

dnephin:
There's a fix to disable the flaky test in master now. If you rebase on master the build should be green.

contrib/completion/bash/docker-compose
```diff
@@ -93,7 +93,7 @@ __docker_compose_services_stopped() {
 _docker_compose_build() {
 	case "$cur" in
 		-*)
-			COMPREPLY=( $( compgen -W "--help --no-cache --pull" -- "$cur" ) )
+			COMPREPLY=( $( compgen -W "--help --force-rm --no-cache --pull" -- "$cur" ) )
```
albers: Please keep the options in alphabetical order here.
Otherwise, bash completion LGTM.
Thanks, Adrian!

aanand:
LGTM after rebase

dnephin:
I've rebased this, and it will be merged in #2361

dnephin:
Thanks for the contribution!

Issue #2259, opened 2015-10-26
Unicode error in volume paths


dnephin:
Reported here: https://github.com/docker/compose/issues/2234#issuecomment-151007248

kuailedubage:
So sorry to reply so late due to the time zone problem:-).It occurs `UnicodeEncodeError` if the `host_path` of volume contains non-ascii characters.But it works OK when I replace "{}:{}" with u"{}:{}" in line 508 of config.py.

mnowster:
@kuailedubage ah no worries, thank you for reporting it :) I've submitted a fix in #2264

PR #2264, opened 2015-10-27
Handle non-ascii volume directories

mnowster:
Fixes https://github.com/docker/compose/issues/2259

### DIFF
--- a/tests/unit/config/config_test.py
+++ b/tests/unit/config/config_test.py
@@ -573,6 +573,11 @@ class VolumeConfigTest(unittest.TestCase):
         }, working_dir='.')
         self.assertEqual(d['volumes'], ['~:/data'])
 
+    def test_volume_path_with_non_ascii_directory(self):
+        volume = u'/Füü/data:/data'
+        container_path = config.resolve_volume_path(volume, ".", "test")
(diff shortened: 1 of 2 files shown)


dnephin:
LGTM

tests/unit/config/config_test.py
```diff
@@ -573,6 +573,11 @@ def test_home_directory_with_driver_does_not_expand(self):
         }, working_dir='.')
         self.assertEqual(d['volumes'], ['~:/data'])
 
+    def test_volume_path_with_non_ascii_directory(self):
+        volume = '/Füü/data:/data'
+        container_path = config.resolve_volume_path(volume, ".", "test")
+        self.assertEqual(container_path, '/Füü/data:/data')
```
dnephin: I can confirm this test fails on python2 (windows and linux). 

I was able to fix it by chaining this to:

```python
volume = u'/Füü/data:/data'
```

and

```python
self.assertEqual(container_path, volume)
```

which matches a real run, because yaml.load() returns unicode strings (not bytes).

dnephin:
LGTM

kuailedubage:
Thank you all:-).

PR #2265, opened 2015-10-27
Update tutorials and small fix with rails tutorial

### LINKED ISSUE #1936: Update compose examples
I've never tried the example projects, but I've noticed we get a decent number of PRs and issues opened about them (ex: #1843, and many of our [oldest PRs](https://github.com/docker/compose/pulls?q=is%3Aopen+is%3Apr+label%3Akind%2Fdocumentation)).

It may be that we only need to make a couple small changes to smooth out the examples, but I'm opening this issue to track the work.

Related:
*  #750
* #1843
* #254

dnephin:
Fixes #1936

* adds a note about `chown` and the ownership of files (addresses #1843)
* we merged a PR that added `Gemfile.lock` to the `Dockerfile`, but it broke the rails tutorial.  Adding an  empty one fixes it
* added the full path for the `database.yml` so it's clear what file needs to be edited.

cc @moxiegirl for docs review

### DIFF
--- a/docs/django.md
+++ b/docs/django.md
@@ -110,8 +110,20 @@ In this step, you create a Django started project by building the image from the
 
 3. After the `docker-compose` command completes, list the contents of your project.
 
-        $ ls
-        Dockerfile       docker-compose.yml          composeexample       manage.py        requirements.txt
+        $ ls -l
+        drwxr-xr-x 2 root   root   composeexample
+        -rw-rw-r-- 1 user   user   docker-compose.yml
+        -rw-rw-r-- 1 user   user   Dockerfile
+        -rwxr-xr-x 1 root   root   manage.py
+        -rw-rw-r-- 1 user   user   requirements.txt
+
+    The files `django-admin` created are owned by root. This happens because
+    the container runs as the `root` user.
+
+4. Change the ownership of the new files.
+
+        sudo chown -R $USER:$USER .
+
 
 ## Connect the database
 
--- a/docs/rails.md
+++ b/docs/rails.md
@@ -37,6 +37,10 @@ Next, create a bootstrap `Gemfile` which just loads Rails. It'll be overwritten 
     source 'https://rubygems.org'
     gem 'rails', '4.2.0'
 
+You'll need an empty `Gemfile.lock` in order to build our `Dockerfile`.
+
+    $ touch Gemfile.lock
+
 Finally, `docker-compose.yml` is where the magic happens. This file describes the services that comprise your app (a database and a web app), how to get each one's Docker image (the database just runs on a pre-made PostgreSQL image, and the web app is built from the current directory), and the configuration needed to link them together and expose the web app's port.
 
     db:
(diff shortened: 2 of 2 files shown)


docs/django.md
```diff
@@ -113,6 +113,13 @@ In this step, you create a Django started project by building the image from the
         $ ls
```
moxiegirl: This is a note that should be part of  a step. Start by correcting step 3 to do an `ls -al`  Then...under that put:

  The files `django-admin` created are owned by root. This happens because the
  container runs as the `root` user.  

 4. Change the ownership of the new files.

      `sudo chown -R $USER:$USER .`

docs/rails.md
```diff
@@ -37,6 +37,10 @@ Next, create a bootstrap `Gemfile` which just loads Rails. It'll be overwritten
     source 'https://rubygems.org'
     gem 'rails', '4.2.0'
 
+We'll also need an empty `Gemfile.lock` in order to build our `Dockerfile`.
```
moxiegirl: You'll need an empty `Gemfile.lock` file to build the `Dockerfile`.

image

docs/rails.md
```diff
@@ -80,6 +84,12 @@ rebuild.)
 
     $ docker-compose build
 
+
+> *Note:* You'll notice the files created by `rails new` are owned
```
moxiegirl: Replace with the information I gave you above...as this is not in procedure but narrative style, tweak in this way

The files `django-admin` created are owned by root. This happens because the  container runs as the `root` user.  Change the ownership of the new files.

      `sudo chown -R $USER:$USER .`

docs/rails.md
```diff
@@ -87,8 +97,7 @@ expects a database to be running on `localhost` - so you need to point it at the
 `db` container instead. You also need to change the database and username to
 align with the defaults set by the `postgres` image.
 
-Open up your newly-generated `database.yml` file. Replace its contents with the
-following:
+Replace the contents of `config/database.yml` with the following configuration.
```
moxiegirl: ...with the following:

dnephin:
I've made those changes

moxiegirl:
LGTM Thanks @dnephin merge at will as far as I'm concerned.

docs/django.md
```diff
@@ -110,8 +110,20 @@ In this step, you create a Django started project by building the image from the
...
+        -rw-rw-r-- 1 user   user   Dockerfile
+        -rwxr-xr-x 1 root   root   manage.py
+        -rw-rw-r-- 1 user   user   requirements.txt
+
+    The files `django-admin` created are owned by root. This happens because
+    the container runs as the `root` user.
+
+4. Change the ownership of the new files.
+
+        sudo chown -R $USER:$USER .
```
romanlevin: This does not fix the `IOError: [Errno 13] Permission denied: '/code/manage.py'`  issue, since it fails the docker-compose command, and the files are never created in the project directory.

romanlevin: Nope, sorry, this is happening because of SELinux on Fedora.

PR #2276, opened 2015-10-28
Update docs about networking for current release

### LINKED ISSUE #2248: Network aliases created with full container name & not short name

dnephin:
cc @dave-tucker, @aanand 

I've updated the compose networking docs to match what the engine supports for this release.

Fixes #2248

### DIFF
--- a/docs/networking.md
+++ b/docs/networking.md
@@ -14,7 +14,11 @@ weight=6
 
 > **Note:** Compose's networking support is experimental, and must be explicitly enabled with the `docker-compose --x-networking` flag.
 
-Compose sets up a single default [network](/engine/reference/commandline/network_create.md) for your app. Each container for a service joins the default network and is both *reachable* by other containers on that network, and *discoverable* by them at a hostname identical to the service's name.
+Compose sets up a single default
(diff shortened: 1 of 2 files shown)


dave-tucker:
LGTM

mnowster:
I think I understand why the scaling section has been removed, do you think it'd be worth including a section/statement about scaling that describes why it's not currently available? I can see users being a bit confused that it's not there/asking about it maybe?

Everything else is LGTM :)

dnephin:
With this setup scaling is available, it's just not that useful because you'd have to do some form of client-side load balancing or have an external, but that's how it was with links.

I think the section existed for the "hostname" version of this doc because the behaviour was changing from links, but now that it's the same, I don't think we need to call it out in the document.

mnowster:
@dnephin ah I see, ok, cool.

PR #2278, opened 2015-10-28
Disable flaky test against 1.8.3

dnephin:
Now that 1.9.0rc3 is out this test is passing reliably, but it still fails against the older engine version. Let's disable it for the older version so we have reliable green builds again.


mnowster:
LGTM

PR #2020, opened 2015-09-09
Fix compose up with other log drivers

dnephin:
Fixes #1947

Update the docs to say that custom log drivers make it impossible to view the logs.

Update log printing to print a reasonable warning when an alternative log_driver is set.

Also some refactoring in the last commit.

PR #2283, opened 2015-10-29
Logs are available for all log drivers except for none.

dnephin:
When I implemented #2020 I think I was confused about how these log drivers worked.  It turns out we can't get logs from other drivers, but attach does work!

This was reported on the docker-user@ mailing list.

### DIFF
--- a/compose/container.py
+++ b/compose/container.py
@@ -145,7 +145,7 @@ class Container(object):
     @property
     def has_api_logs(self):
         log_type = self.log_driver
-        return not log_type or log_type == 'json-file'
+        return not log_type or log_type != 'none'
 
     def attach_log_stream(self):
         """A log stream can only be attached if the container uses a json-file


mnowster:
I think we need to make a change in https://github.com/docker/compose/blob/master/compose/cli/log_printer.py#L60 to inspect the type of driver as well maybe? As the check for `has_api_logs` won't be enough by itself now, this will try and get logs for things we can't get logs from.

dnephin:
I think that's ok, this is the code path I was testing.  maybe the name needs to be changed now?  I believe the only way we get logs at this time is by using attach, we don't use the logs api endpoint.

mnowster:
Ah right, sorry I got confused there for a moment.

Yeh, it could do with being changed but not necessarily as part of this, can be done as part of the changes to use the logs api endpoint.

LGTM. Merge away :sparkles:

PR #2287, opened 2015-10-29
Ensure network exists when calling run before up

### LINKED ISSUE #2284: Running a service before `up` with networking enabled fails
Steps to reproduce (against docker 1.9.0-rc3):

1. Create a new project, or set a new project name
2. Run any service with the new project name `docker-compose --x-networking run service bash`

You'll get an error:
`Cannot start container b631dcb1feea803af2d06471789fe3646a3b6f59af04bc771f66fd70137a024b: network networking not found`

We need to create the network as part of `run` as well as `up` I think.

shin-:
Fixes #2284

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -380,6 +380,8 @@ class TopLevelCommand(DocoptCommand):
                     start_deps=True,
                     strategy=ConvergenceStrategy.never,
                 )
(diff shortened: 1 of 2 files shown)


shin-:
Added integration test.

dnephin:
LGTM

Issue #2289, opened 2015-10-29
Multiple file parameter does not use the correct working directory


LoicMahieu:
Related: #2051 

I would like to use multiple `-f` parameter for running multiple docker-compose in one.
Example:
Project A needs Project B. There are not in the same directory.
- Project A docker-compose.yml:
```yml
app:
  image: some/image
  command: /project-A-start.sh
  working_dir: /app
  volumes:
    - .:/app
  ports:
    - "3001:3000"
  links:
    - api
```

- Project B docker-compose.yml:
```yml
api:
  image: some/other/image
  command: /project-B-start.sh
  working_dir: /app
  volumes:
    - .:/app
  ports:
    - "3000:3000"
```

 So I use the command in project A:
```up
docker-compose -f docker-compose.yml -f ../project-b/docker-compose.yml up
```

But `api` container from project B does not start because command `/project-B-start.sh` can not be found. It appears the volume/cwd used for `../project-b/docker-compose.yml` is not `../project-b` but the initial directory, the project A. (Hope it's clear)
Furthermore, it turns out that it is the first `-f` parameter that define the directory. If I inverse the two parameter, the directory used is now `project-b` and now it is `/project-A-start.sh` that can not be found.

Is this behavior by design ? In my case the perfect behavior would be to use a separate "cwd" for each docker-compose.

Thanks for your help.

---

Docker: 1.8.0
Docker Compose: 1.5.0rc2

dnephin:
Thanks for the report!  

I'm actually working on better documentation for this feature as we speak, so this feedback is helpful.

The order of the files changing the working directory is by design. The first file is always the base and the following files are applied over it in the order they are included. This isn't important for your example case, but it is when a service is defined in both files. We need a well defined order so that we know which file overrides the other.

I think the current working directory being relative to the first file is also necessary. There's no requirement that the overriding files are completely valid compose files. They may contain services that are missing a `build` or `image` (effectively abstract), or they may contain links to services that aren't defined locally.

Since each file is just a fragment, it's extremely difficult to keep track of which fragments are relative to which path.

I'll make sure to include both of these details in the documentation.

-----

To support your use case, you can remove the `volumes` when including it from a separate project, and use the code in the image.

To support that workflow, you could split out the Compose file into two:

Project B `docker-compose.yml`:

```yml
api:
  image: some/other/image
  command: /project-B-start.sh
  working_dir: /app
```

Project B `docker-compose.override.yml`:

```yml
api:
  volumes:
[...]

Issue #1786, opened 2015-07-29
Document best practices for how to manage releases in production


bfirsh:
We currently have [some lightweight documentation](https://docs.docker.com/compose/production/) about how to use Compose in production, but this could do with improvements:
- How to manage releases, particularly when deploying through the Docker Hub (e.g. in development build images, in production use image from Hub)
- A step-by-step guide to deploying a Compose app from dev through to production
- Some examples of how to deploy apps using init scripts
- Some examples of how to deploy apps using Swarm
- How to deploy apps on a single server (particularly useful for internal tools, etc)

Resources
- [Examples of production Compose files](https://github.com/search?utf8=%E2%9C%93&q=in%3Apath+docker-compose-prod.yml+extension%3Ayml&type=Code&ref=searchresults)

funkyfuture:
> init scripts

and service-configs for systemd, that thing will take over. i thought of a command to generate these configs.

dnephin:
There's some previous discussion and requests in #93, #1264, #1730, #1035. I've closed those tickets in favour of this one, but linking to them for reference.

soulrebel:
I would like some recommendations on how to deploy config files through docker/compose/swarm. 
We have a setup (that was recommended by consultants) that makes images with config files and declares volumes to export them. This looks good in principle, but it does not work as expected in a number of cases.

dnephin:
Putting the configs into a "config" image that just exposes a volume seems like a reasonable way to do it. I'd be great to hear more about the cases where it doesn't work, either here or in a new issue.

soulrebel:
@dnephin If you go down the road of one single configuration container and use --volumes-from you sacrifice some security (every container sees all configs) but it looks easy to set up and nice: it's immutable and does not use any host fs path. 

Once you operate outside localhost and do not recreate everything at each run, you start learning the subtleties of the --volumes-from and compose recreation policies: the config image is updated and restarted, but client container still mount their current volumes unless they are recreated as well for independent reasons. This took a while to be noticed and left us with a workaround of deleting the old config container whenever the config changes.

Another solution would seem to be avoiding immutability and change data inside the same volumes, running a cp or similar. At this point it would just be easier to pull the configs from a git repo and skip the image build altogether... which was the stateful solution I originally had in mind. If you want no fixed host path, you need a config data-only container and a config copier.

[...]

Issue #1784, opened 2015-07-29
Epic: Define an application once in a way that can be used in dev, test and prod


bfirsh:
It should be possible to define a single file that can abstractly describe the high-level structure of application. This file should be able to be used across development, test and production with no modification, but may need additional configuration files or environment variables to describe the variances between environments.

**Related things:**

 - [x] [`extends` in Compose files](https://docs.docker.com/compose/extends/). This was a good start, but is still pretty clunky for achieving the above.
 - [x] [Interpolate environment variables in Compose files](https://github.com/docker/compose/issues/1377)
 - [x] Research use cases and what users want to do
 - [x] [Extend an entire Compose file](https://github.com/docker/compose/issues/1987)
 - [x] Recommend patterns for achieving what users want to do (see also #1786)

(This list is not a complete list of what it will take to achieve this goal, and may change as we research what users want.)

**Research:**

- Here is a recommended setup for how to run a Django app in both dev and production with Compose, using two different Compose files: https://github.com/pydanny/cookiecutter-django/pull/284
- User feedback on extends and where it falls short for this use case: https://github.com/docker/compose/issues/1380
- [Example production Compose files](https://github.com/search?utf8=%E2%9C%93&q=in%3Apath+docker-compose-prod.yml+extension%3Ayml&type=Code&ref=searchresults)

aanand:
#1380 provides some user feedback on `extends` and where it falls short for this use case.

dnephin:
I think https://github.com/dnephin/compose-addons#dcao-merge addresses some of this. It provides a convenient way to vary a configuration but keep a similar base.

dnephin:
This is now documented in `extends.md`

PR #2290, opened 2015-10-29
Docs for using multiple compose files

### LINKED ISSUE #2141: Multiple environment getting started
* Rework the getting started for extends so that it works for multiple environments.
* Change the URL if necessary (redirect for extends URL)
* Daniel to do a pass Mary to edit

dnephin:
Resolves #2141, #2289, also related to #1784 and #1786

cc @moxiegirl , @bfirsh

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -13,8 +13,8 @@ DOCKER_ENVS := \
 	-e TIMEOUT
 # note: we _cannot_ add "-e DOCKER_BUILDTAGS" here because even if it's unset in the shell, that would shadow the "ENV DOCKER_BUILDTAGS" set in our Dockerfile, which is very important for our official builds
 
-# to allow `make DOCSDIR=docs docs-shell` (to create a bind mount in docs)
-DOCS_MOUNT := $(if $(DOCSDIR),-v $(CURDIR)/$(DOCSDIR):/$(DOCSDIR))
+# to allow `make DOCSDIR=1 docs-shell` (to create a bind mount in docs)
+DOCS_MOUNT := $(if $(DOCSDIR),-v $(CURDIR):/docs/content/compose)
 
 # to allow `make DOCSPORT=9000 docs`
 DOCSPORT := 8000
@@ -37,7 +37,7 @@ GITCOMMIT := $(shell git rev-parse --short HEAD 2>/dev/null)
 default: docs
 
 docs: docs-build
-	$(DOCKER_RUN_DOCS) -p $(if $(DOCSPORT),$(DOCSPORT):)8000 -e DOCKERHOST "$(DOCKER_DOCS_IMAGE)" hugo server --port=$(DOCSPORT) --baseUrl=$(HUGO_BASE_URL) --bind=$(HUGO_BIND_IP)
+	$(DOCKER_RUN_DOCS) -p $(if $(DOCSPORT),$(DOCSPORT):)8000 -e DOCKERHOST "$(DOCKER_DOCS_IMAGE)" hugo server --port=$(DOCSPORT) --baseUrl=$(HUGO_BASE_URL) --bind=$(HUGO_BIND_IP) --watch
 
 docs-draft: docs-build
 	$(DOCKER_RUN_DOCS) -p $(if $(DOCSPORT),$(DOCSPORT):)8000 -e DOCKERHOST "$(DOCKER_DOCS_IMAGE)" hugo server --buildDrafts="true" --port=$(DOCSPORT) --baseUrl=$(HUGO_BASE_URL) --bind=$(HUGO_BIND_IP)
--- a/docs/compose-file.md
+++ b/docs/compose-file.md
@@ -169,21 +169,21 @@ Extend another service, in the current file or another, optionally overriding
 configuration.
 
 You can use `extends` on any service together with other configuration keys.
-The value must be a dictionary with the key: `service` and may optionally have
-the `file` key.
+The `extends` value must be a dictionary defined with a required `service`
+and an optional `file` key.
 
     extends:
       file: common.yml
       service: webapp
 
-The `file` key specifies the location of a Compose configuration file defining
-the service which is being extended. The `file` value can be an absolute or
-relative path. If you specify a relative path, Docker Compose treats it as
-relative to the location of the current file. If you don't specify a `file`,
-Compose looks in the current configuration file.
+The `service` the name of the service being extended, for example
+`web` or `database`. The `file` is the location of a Compose configuration
+file defining that service.
 
-The `service` key specifies the name of the service to extend, for example `web`
-or `database`.
+If you omit the `file` Compose looks for the service configuration in the
+current file. The `file` value can be an absolute or relative path. If you
+specify a relative path, Compose treats it as relative to the location of the
+current file.
 
 You can extend a service that itself extends another. You can extend
 indefinitely. Compose does not support circular references and `docker-compose`
--- a/docs/compose-file.md
+++ b/docs/compose-file.md
@@ -168,44 +168,29 @@ accessible to linked services. Only the internal port can be specified.
 Extend another service, in the current file or another, optionally overriding
 configuration.
 
-Here's a simple example. Suppose we have 2 files - **common.yml** and
-**development.yml**. We can use `extends` to define a service in
-**development.yml** which uses configuration defined in **common.yml**:
+You can use `extends` on any service together with other configuration keys.
+The value must be a dictionary with the key: `service` and may optionally have
+the `file` key.
 
-**common.yml**
+    extends:
+      file: common.yml
+      service: webapp
 
-    webapp:
-      build: ./webapp
-      environment:
-        - DEBUG=false
-        - SEND_EMAILS=false
+The `file` key specifies the location of a Compose configuration file defining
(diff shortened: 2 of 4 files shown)


docs/Makefile
```diff
@@ -37,7 +37,7 @@ GITCOMMIT := $(shell git rev-parse --short HEAD 2>/dev/null)
 default: docs
 
 docs: docs-build
-	$(DOCKER_RUN_DOCS) -p $(if $(DOCSPORT),$(DOCSPORT):)8000 -e DOCKERHOST "$(DOCKER_DOCS_IMAGE)" hugo server --port=$(DOCSPORT) --baseUrl=$(HUGO_BASE_URL) --bind=$(HUGO_BIND_IP)
+	$(DOCKER_RUN_DOCS) -p $(if $(DOCSPORT),$(DOCSPORT):)8000 -e DOCKERHOST "$(DOCKER_DOCS_IMAGE)" hugo server --port=$(DOCSPORT) --baseUrl=$(HUGO_BASE_URL) --bind=$(HUGO_BIND_IP) --watch
```
dnephin: cc @SvenDowideit as well on this change.  I'm pretty sure the old setup wasn't setting the correct volume path.  With this I was able to see my changes immediately by just refreshing.

I hear hugo does livereload as well, but I don't see it injecting the livereload.js

docs/compose-file.md
```diff
@@ -163,44 +163,29 @@ accessible to linked services. Only the internal port can be specified.
 Extend another service, in the current file or another, optionally overriding
 configuration.
 
-Here's a simple example. Suppose we have 2 files - **common.yml** and
-**development.yml**. We can use `extends` to define a service in
-**development.yml** which uses configuration defined in **common.yml**:
+You can use `extends` on any service together with other configuration keys.
```
moxiegirl: You can use `extends` on any service together with other configuration keys.
The `extends` value must be a dictionary defined with a required `service` and an optional `file` key.

    extends:
      file: common.yml
      service: webapp

The `service` is what is being extended, , for example `web` or `database`. the
`file` is the location of a Compose configuration file defining that service. If
you omit the `file` Compose looks for the service configuration in the current
file. The `file` value can be an absolute or relative path. If you specify a
relative path, Compose treats it as relative to the location of the current
file. If you don't specify a `file`, Compose looks in the current configuration
file.

You can extend a service that itself extends another. You can extend
indefinitely. Compose does not support circular references and `docker-compose`
returns an error if it encounters one.

docs/extends.md
```diff
@@ -77,183 +85,45 @@ You can also write other services and link your `web` service to them:
...
-          volumes:
-            - .:/code
-          links:
-            - redis
-          environment:
-            - REDIS_HOST=redis
-        redis:
-          image: redis
+Extending an individual service is useful when you have multiple services that
+have a common configuration.  In this example we have a composition with
```
moxiegirl: In this example, the composition two services:

docs/extends.md
```diff
@@ -337,24 +210,152 @@ with locally-defined values taking precedence:
...
+## Multiple Compose files
 
-    # local service
-    volumes:
-      - /local-dir/bar:/bar
-      - /local-dir/baz/:baz
+Using multiple Compose files enables you to customize a composition for
+different environments or different workflows.
+
+### Understanding multiple Compose files
```
moxiegirl: ### Understanding multiple Compose files

By default, Commpose reads two files, a `docker-compose.yml` and an optional `docker-compose.override.yml` file. By convention, the `docker-compose.yml` contains your base configuration. The override file, as it name implies, can contain configuration overrides for existing services or entirely new new services.

If a service is defined in both files, Compose merges the configurations using the same rules as the `extends` field (see [Adding and overriding
configuration](#adding-and-overriding-configuration)), with one exception.  If a
service contains `links` or `volumes_from` those fields are copied over and
replace any values in the original service, in the same way single-valued fields
are copied.

To use multiple override files, or an override file with a different name, you
can use the `-f` option to specify the list of files. Compose merges files in
the order they're specified on the command line. See the [`docker-compose`
command reference](./reference/docker-compose.md) for more information about
using `-f`.

When you use multiple configuration files, you must make sure all paths in the
files are relative to the base Compose file (the first Compose file specified
with `-f`). This is required because override files are need not be valid
Compose files. Override files can contain small fragments of configuration.
Tracking which fragment of a service is relative to which path is difficult and
confusing, so to keep paths easier to understand, all paths must be defined
relative to the base file.

docs/extends.md
```diff
@@ -337,24 +210,152 @@ with locally-defined values taking precedence:
...
+information about using `-f`.
+
+When multiple files are used, all paths in the files are relative to the
+base Compose file (the first Compose file specified with `-f`). This is necessary
+because override files are not always valid Compose files, they may be just small
+fragments of configuration. Tracking which fragment of a service is relative to
+which path is difficult and confusing, so to keep paths easier to understand,
+all paths are kept relative to the base file.
+
+### Example use case
```
moxiegirl: No royal we's and everything happens in the now not future soooo...

--->

### Example use case

In this section are two common use cases for multiple compose files: changing a
composition for different environments, and running administrative tasks
against a composition.

#### Different environments

A common use case for multiple files is changing a development composition
for a production-like environment (which may be production, staging or CI).
To support these differences, you can split your Compose configuration into
a few different files:

Start with a base file that defines the canonical configuration for the
services.

**docker-compose.yml**

    web:
      image: example/my_web_app:latest
      links:
        - db
        - cache

    db:
      image: postgres:latest

    cache:
      image: redis:latest

The example development environment exposes some ports to the
host, mount our code as a volume, and build the web image.

**docker-compose.override.yml**


    web:
      build: .
      volumes:
        - '.:/code'
      ports:
        - 8883:80
      environment:
        DEBUG: 'true'

    db:
      command: '-d'
      ports:
        - 5432:5432

    cache:
      ports:
        - 6379:6379

When you run `docker-compose up` it reads the overrides automatically.

Now, it might be nicet to also use this composition in a production environment. So, create another override file (which might be stored in a different git
repo or managed by a different team).

**docker-compose.prod.yml**

    web:
      ports:
        - 80:80
      environment:
        PRODUCTION: 'true'

    cache:
      environment:
        TTL: '500'

To deploy with this production Compose file you can run

    docker-compose -f docker-compose.yml -f docker-compose.prod.yml up -d

This deploys all three services using the configuration in
`docker-compose.yml` and `docker-compose.prod.yml` (but not the
dev configuration in `docker-compose.override.yml`).

See [production](production.md) for more information about Compose in
production.

#### Administrative tasks

Another common use case is running adhoc or administrative tasks against one
or more services in a composition. This example demonstrates running a
database backup.

Start with a **docker-compose.yml**.

    web:
      image: example/my_web_app:latest
      links:
        - db

    db:
      image: postgres:latest

In a **docker-compose.admin.yml**  add a new service to run the database
export or backup.

    dbadmin:
      build: database_admin/
      links:
        - db


To start a normal environment run `docker-compose up -d`. To run a database
backup, use the `docker-compose.admin.yml` as well.

    docker-compose -f docker-compose.yml -f docker-compose.admin.yml \
        run dbadmin db-backup

dnephin:
Ok, "We" is now purged from the docs

moxiegirl:
LGTM

docs/extends.md
```diff
@@ -262,13 +132,11 @@ locally. This ensures dependencies between services are clearly visible when
 reading the current file. Defining these locally also ensures changes to the
 referenced file don't result in breakage.
 
-If a configuration option is defined in both the original service and the local
-service, the local value either *override*s or *extend*s the definition of the
-original service. This works differently for other configuration options.
+If a configuration option is defined in both the original service the local
+service, the local value *replaces* or *rextends* the original value.
```
aanand: Typo: "rextends"

docs/extends.md
```diff
@@ -337,24 +210,155 @@ with locally-defined values taking precedence:
...
-      - /local-dir/bar:/bar
-      - /local-dir/baz/:baz
+Using multiple Compose files enables you to customize a composition for
+different environments or different workflows.
+
+### Understanding multiple Compose files
+
+By default, Compose reads two files, a `docker-compose.yml` and an optional
+`docker-compose.override.yml` file. By convention, the `docker-compose.yml`
+contains your base configuration. The override file, as it's name implies, can
```
aanand: it's -> its

docs/extends.md
```diff
@@ -337,24 +210,155 @@ with locally-defined values taking precedence:
...
+contain configuration overrides for existing services or entirely new new
```
aanand: new new

docs/extends.md
```diff
@@ -337,24 +210,155 @@ with locally-defined values taking precedence:
...
+command reference](./reference/docker-compose.md) for more information about
+using `-f`.
+
+When you use multiple configuration files, you must make sure all paths in the
+files are relative to the base Compose file (the first Compose file specified
+with `-f`). This is required because override files need not be valid
+Compose files. Override files can contain small fragments of configuration.
+Tracking which fragment of a service is relative to which path is difficult and
+confusing, so to keep paths easier to understand, all paths must be defined
+relative to the base file.
```
aanand: Perhaps make this paragraph a note (i.e. prefix it with `> **Note:**`)

aanand:
LGTM

PR #2295, opened 2015-10-30
Don't set a default network driver, let the server decide

### LINKED ISSUE #2292: default network
I'm not sure I'm reading correctly, but it seems like compose is choosing `bridge` by default as default network driver : https://github.com/docker/compose/blob/master/compose/project.py#L86

We should let the daemon decide instead, same as this PR https://github.com/docker/docker/pull/17504

dnephin:
Fixes #2292

### DIFF
--- a/compose/project.py
+++ b/compose/project.py
@@ -83,7 +83,7 @@ class Project(object):
         self.services = services
         self.client = client
         self.use_networking = use_networking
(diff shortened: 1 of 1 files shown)


dnephin:
> joffrey [1:39 PM] On phone so can't leave a comment, but lgtm @dnephin

PR #2298, opened 2015-10-30
Fix networking tests to work with new API in engine rc4 

dnephin:
Update tests to work with https://github.com/docker/docker/pull/17536


shin-:
LGTM :+1:

PR #2300, opened 2015-10-30
Use "unicode_literals" in all files

dnephin:
Based on #2306 because docopt fails when given unicode strings as args.

We recently fixed a bunch of unicode bugs, many of which were fixed by just using a unicode literal.  To prevent any regressions in this area, we can force python2 to use unicode literals.

I've also added a pre-commit check that will add this import to new files for us (it actually did all the work to add the missing imports in this PR!).

The second commit fixes a test that broke with this change (because the new repr would include the `u''` prefix on the string repr.  Since we're printing out a value from the config, I think using a json reprinstead of the python repr is more correct.  We could also use the yaml repr, but that is harder to read in a condensed error message (and will pretty often be the same as the json repr).

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -14,7 +14,12 @@
     - id: requirements-txt-fixer
     - id: trailing-whitespace
 - repo: git://github.com/asottile/reorder_python_imports
-  sha: 3d86483455ab5bd06cc1069fdd5ac57be5463f10
+  sha: v0.1.0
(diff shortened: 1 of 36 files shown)


dnephin:
Unfortunately docopt doesn't like parsing parameters that are unicode strings in py27.  I'll need some way to work around that.

dnephin:
rebased

aanand:
LGTM

PR #2304, opened 2015-11-02
Remove service.start_container()

dnephin:
It has been an unnecessary wrapper around container.start() for a little while now, so we can call it directly.

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -448,7 +448,7 @@ class TopLevelCommand(DocoptCommand):
             raise e
 
         if detach:
-            service.start_container(container)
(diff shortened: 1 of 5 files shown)


aanand:
Nice. Needs a rebase.

dnephin:
rebased

aanand:
Looks like Jenkins is failing because we need a `1.9.0-rc5` tag for the `dind` image.

dnephin:
hmm, ya, I just created it.I guess I didn't wait long enough before rebuilding

aanand:
LGTM

PR #2306, opened 2015-11-03
Update cli test suite to run the executable in a child process

dnephin:
This change allows us to re-use the test suite against other executables (like the one in libcompose).  It should also make it easier to test the behaviour of `run`, and support #2300


aanand:
Nice.

LGTM

Issue #1490, opened 2015-05-31
Console animations printed to the log files as new lines


ahmetalpbalkan:
I have a system that execs the `compose`command and that logs the output (stderr/stdout output streams redirected) to a log file.

Apparently the `docker up -d` is showing some image pull animations (obviously contains ANSI escape sequences to clear lines or delete chars) and when those get logged to the log file, I just see bunch of empty lines.

Not a big deal, mostly a cosmetics issue but wanted to bring up if it's easy to fix.

Output:

```
Creating compose307929553_blog_1...
Pulling image ghost:latest...
latest: Pulling from ghost






















ghost:latest: The image you are pulling has been verified. Important: image verification is a tech preview feature and should not be relied on to provide security.
Digest: sha256:b415d8cdf5aa809b95def3e122251b80fe6f99c557bd37ba9d7af2523089511a
Status: Downloaded newer image for ghost:latest
Creating compose307929553_cache_1...
Pulling image memcached:latest...
latest: Pulling from memcached










memcached:latest: The image you are pulling has been verified. Important: image verification is a tech preview feature and should not be relied on to provide security.
Digest: sha256:61a047af4f93bf3750658197594e5145beadf02f1236b411034b9ce6cfa61140
Status: Downloaded newer image for memcached:latest
```

dnephin:
Is this on windows?

ahmetalpbalkan:
@dnephin nope. This is linux. I'm just running `docker-compose` through Go (`os/exec.Command#CombinedOutput`) and in the output I'm getting I see those empty lines in the returned data.

PR #2311, opened 2015-11-03
Removed unneeded newlines on docker-compose pull with no tty

yvespp:
"docker-compose pull &> test.log" would print a lot of newlines, this pull request fixes this.

old output:

    (env)lodur:compose yves$ docker-compose -f ../../../../../../compose.yaml pull  &> test.log
    (env)lodur:compose yves$ cat test.log 
    Pulling redis (redis:latest)...
    latest: Pulling from library/redis
    
    
    
    
    
    
    
    
    
    
    
    
    
    
    
    
    
    Digest: sha256:cef5e2b2218eec7f2619bc115e69207c2a222d265ca6dc8e8f97b5a454f3510b
    Status: Image is up to date for redis:latest

new output:

    (env)lodur:compose yves$ docker-compose -f ../../../../../../compose.yaml pull  &> test.log
    (env)lodur:compose yves$ cat test.log 
    Pulling redis (redis:latest)...
    latest: Pulling from library/redis
    Digest: sha256:cef5e2b2218eec7f2619bc115e69207c2a222d265ca6dc8e8f97b5a454f3510b
    Status: Image is up to date for redis:latest

### DIFF
--- a/compose/progress_stream.py
+++ b/compose/progress_stream.py
@@ -14,8 +14,14 @@ def stream_output(output, stream):
 
     for event in utils.json_stream(output):
         all_events.append(event)
+        is_progress_event = 'progress' in event or 'progressDetail' in event
 
-        if 'progress' in event or 'progressDetail' in event:
+        if not is_progress_event:
+            print_output_event(event, stream, is_terminal)
+            stream.flush()
+
+        # if it's a progress event and we have a terminal, then display the progress bars
+        elif is_terminal:
             image_id = event.get('id')
             if not image_id:
                 continue
@@ -27,17 +33,16 @@ def stream_output(output, stream):
                 stream.write("\n")
                 diff = 0
 
-            if is_terminal:
-                # move cursor up `diff` rows
-                stream.write("%c[%dA" % (27, diff))
+            # move cursor up `diff` rows
+            stream.write("%c[%dA" % (27, diff))
 
-        print_output_event(event, stream, is_terminal)
+            print_output_event(event, stream, is_terminal)
 
-        if 'id' in event and is_terminal:
-            # move cursor back down
-            stream.write("%c[%dB" % (27, diff))
+            if 'id' in event:
+                # move cursor back down
+                stream.write("%c[%dB" % (27, diff))
 
-        stream.flush()
+            stream.flush()
 
     return all_events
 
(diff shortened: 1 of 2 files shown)


dnephin:
Fixes #1490

compose/progress_stream.py
```diff
@@ -27,15 +30,18 @@ def stream_output(output, stream):
...
+            if 'id' in event:
+                # move cursor back down
+                stream.write("%c[%dB" % (27, diff))
 
-        if 'id' in event and is_terminal:
-            # move cursor back down
-            stream.write("%c[%dB" % (27, diff))
+        # print all other events
+        elif not is_progress_event:
+            print_output_event(event, stream, is_terminal)
```
dnephin: To make this a bit easier to read, and remove the need for the second check of `is_progress_event`, could we switch this around and do:

```python

if not is_progress_event:
    print_output_event(event, stream, is_terminal)
    continue

# The rest of the logic below here
...

```

dnephin:
Thanks for the PR! I think this will fix the issue, but could use a new test case in `tests/unit/progress_stream_test.py`.

Don't worry about he appveyor failure, it's unrelated to your branch.

It would also be good to squash these commits into a single commit.

yvespp:
Merged the commits, changed the if and added tests.

dnephin:
LGTM

compose/progress_stream.py
```diff
@@ -14,8 +14,14 @@ def stream_output(output, stream):
...
         all_events.append(event)
+        is_progress_event = 'progress' in event or 'progressDetail' in event
 
-        if 'progress' in event or 'progressDetail' in event:
+        if not is_progress_event:
+            print_output_event(event, stream, is_terminal)
+            stream.flush()
+
+        # if it's a progress event and we have a terminal, then display the progress bars
+        elif is_terminal:
```
dnephin: I took another look at this with @aanand 

I think by making this a couple of guard statements it would be slightly easier to understand. How about this:

```python
if not is_progress_event:
    ...
   continue

if not is_terminal:
    continue

...
```

That way the rest of the logic is outside of the if/elif block

dnephin:
Thanks for the PR and fix!

We merged this in #2371 after rebasing

PR #2314, opened 2015-11-04
Fix `up` "requires build" error

### LINKED ISSUE #1679: docker-compose up assumes --no-build
It seems if I try to run docker-compose up with a yml that uses `build: .` and the image does not already exist, it acts like  --no-build is passed.

This is on version 1.3.1

dnephin:
Fixes #1679


aanand:
LGTM

PR #2325, opened 2015-11-04
Fix jenkins CI

dnephin:
by using an older docker version to match the host


dnephin:
appveyor is still broken, but this fixes jenkins at least

shin-:
lgtm!

PR #2331, opened 2015-11-05
Flush the log stream after writing each line

### LINKED ISSUE #1549: Flush log buffer when printing logs
```bash
 $ docker-compose up -d && docker-compose logs &> composer.log
```
Someone correct me if I am wrong here, but what happens in the above is that nothing gets written to `composer.log` until `docker-compose logs` is killed, at which time stdout/stderr are unbuffered and written to `composer.log`?

dnephin:
Fixes #1549

Includes some refactoring of log_printer_test to support checking for flush(), and so that each test calls the unit-under-test directly, instead of through a helper function.


shin-:
LGTM!

chiefy:
:+1: thanks!

PR #2334, opened 2015-11-05
Don't set the hostname to the service name with networking.

### LINKED ISSUE #2327: Don't set hostname to service name when using --x-networking
Using the new networking features, when you start a container.
The name is correctly `<folder>_<service>_1` but the hostname is `<service>`

It's impossible the scale the containers in some case because all the containers get the same hostname.

dnephin:
Fixes #2327

Since it's not being used for anything, I think we can undo this part of the networking experiment. When we get alias support there should be some other mechanism we can use to set the aliases.

cc @aanand @vieux

### DIFF
--- a/compose/service.py
+++ b/compose/service.py
@@ -599,9 +599,6 @@ class Service(object):
             container_options['hostname'] = parts[0]
             container_options['domainname'] = parts[2]
 
-        if 'hostname' not in container_options and self.use_networking:
-            container_options['hostname'] = self.name
-
         if 'ports' in container_options or 'expose' in self.options:
             ports = []
             all_ports = container_options.get('ports', []) + self.options.get('expose', [])
--- a/tests/acceptance/cli_test.py
+++ b/tests/acceptance/cli_test.py
@@ -205,7 +205,6 @@ class CLITestCase(DockerClientTestCase):
             containers = service.containers()
             self.assertEqual(len(containers), 1)
             self.assertIn(containers[0].id, network['Containers'])
-            self.assertEqual(containers[0].get('Config.Hostname'), service.name)
 
         web_container = self.project.get_service('web').containers()[0]
         self.assertFalse(web_container.get('HostConfig.Links'))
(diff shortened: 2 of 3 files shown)


vieux:
LGTM

vieux:
thanks @dnephin

vieux:
@abronan you might want to use this next week.

abronan:
Indeed, LGTM.

aanand:
LGTM

okgemaltoprg:
What if I'd like have quite opposite? Set hostname to full service name? E.g. test_myapp_1 instead of myapp only? The reason is following:
I have to get exposed dynamic port from within the container to be able to set Hazelcast bind port on public IP.

Previously (Docker 1.8.x), mistakenly using "experimental" /etc/hosts mapping, I could easily connect my container IP with my container name (because it was listed together with the rest of the services running on default subnet) and use it as filter for list of services gathered from Consul.

Now my container could be only uid (default Docker /etc/hosts mapping) or service name (generated by Docker-compose) where neither of it helps discovering my real container name (test_myapp_1) published by Registrator to Consul.
There are multiple ways how to solve it, however none of them is as trivial as the previous one.

Any ideas? Could it be optional?

dnephin:
Does it need to be the container_name? Could it be the unique  container id, which you can get from `$(hostname)` ?

okgemaltoprg:
That's the trick - (gliderlabs) Registrator registers IP (default=server/host), container_name, published ports. There's no way (please correct me if I'm wrong) how to force it to register container id (as tag for example).
So, in case there's more than one service running on one host (for HA), one have to filter response from Consul somehow. The only filter available in that case is container_name which is currently unknown again inside the container (talking about production hardened setup, not running on host network or mounting Docker socket, or exposing Docker api).

Previously, I used container name "experimentally" added to /etc/hosts (registered by Registrator in Consul by default) to filter running services on that specific host and get dynamically assigned port as result, then export it as Java variable and start Tomcat.

curl --connect-timeout 1 -m 1 --no-keepalive -sS http://host:8500/v1/health/service/oobs-hazelcast | jq --arg HOST_IP ${HOST_IP} --arg CONTAINER_NAME ${CONTAINER_NAME} -r '.[] | (.Service.Port | tostring) + "-" + .Node.Address + "-" + .Service.ID | select(contains($HOST_IP)) | select(contains($CONTAINER_NAME))'

dnephin:
Ok, this sounds a lot like #1503, let's continue the discussion there

Issue #2308, opened 2015-11-03
Container volume masks host volume in the compose file


mekanixdk:
I have a little snippet in `docker-compose.yml` like:

```
volumes:
  - ${HOST_PATH}:/container/path
```
If I change the value of `HOST_PATH` any subsequent `up` or `build`, the container will still mount the volume with the *old* value of `HOST_PATH`.

Is this because `volumes` is not a run time option like `docker -v` or is this a bug?

aramalipoor:
You're right it's only used when creating the container, I think using `up --force-recreate` flag will create a new container with new volumes.

dnephin:
I can't reproduce this issue with master .What version of docker-compose are you running?

Here's what I ran:

```yaml
web:                                                                                                                                                                                                                                                                              
  image: 'alpine:edge'                                                                                                                                                                                                                                                            
  volumes:                                                                                                                                                                                                                                                                        
    - '${HOST_PATH}:/contents'                                                                                                                                                                                                                                                    
  command: 'ls /contents'   
```

```
HOST_PATH=`pwd`/examples docker-compose up
# Prints the list of files in examples

HOST_PATH=. docker-compose up
# Prints a different list of files in the current directory
```

mekanixdk:
I was using 1.5.0 RC1 and RC2. I can see that 1.5.0 have just been released. I will try that with your example above.

My container looks like this:
```
mysql:
  container_name:
    dungbeetle_mysql
  image:
    mysql:5.6
  environment:
    MYSQL_ROOT_PASSWORD: password
  expose:
    - "3306"
  ports:
    - "3306:3306"
  volumes:
    - ${DUNGBEETLE_MYSQL_DATA}:/var/lib/mysql
  hostname: database
```

mekanixdk:
@dnephin you example worked like you said.

I am a noob regarding docker, but here is a theory of what I am seeing.

In the mysql Dockerfile `VOLUME` is defined (`VOLUME /var/lib/mysql`), and as I read the docs the container creates that "volume" on the host `/somewhere/where/docker/containers/are/stored/too`

In `docker-compose.yml` I define:

```
volumes:
  - ${HOST_PATH}:/var/lib/mysql
```
Ie. I try to define *where* that "volume" should be located on the host.

It *appears* like this overwrite the `VOLUME` in Dockerfile, and consequently at build time I can define my own path.

[...]

PR #2335, opened 2015-11-05
Some small refactoring around volumes

dnephin:
This is some cleanup I did while working on #2308. I'm splitting it off as a separate PR so it can be reviewed separately.

Changes are hopefully well explained in the git commit message.

### DIFF
--- a/compose/config/config.py
+++ b/compose/config/config.py
@@ -252,7 +252,7 @@ class ServiceLoader(object):
         if not self.already_seen:
             validate_against_service_schema(service_dict, self.service_name)
 
-        return process_container_options(service_dict, working_dir=self.working_dir)
+        return process_container_options(self.working_dir, service_dict)
 
     def validate_and_construct_extends(self):
         extends = self.service_dict['extends']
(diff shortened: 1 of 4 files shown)


compose/config/config.py
```diff
@@ -345,14 +345,14 @@ def validate_extended_service_dict(service_dict, filename, service):
                 "%s services with 'net: container' cannot be extended" % error_prefix)
 
 
-def process_container_options(service_dict, working_dir=None):
-    service_dict = service_dict.copy()
+def process_container_options(working_dir, service_dict):
+    service_dict = dict(service_dict)
```
shin-: What's the idea behind this change? Isn't the previous version "safer"?

shin-:
Other than the one question, LGTM

compose/config/config.py
dnephin: I think they are able the same thing. They would both make a shallow copy, one using the constructor and another using the copy method.  I wasn't able to find much about the differences.

I think for the case where `service_dict` is actually a `dict` they would be equivalent, if `service_dict` is `UserDict`, `OrderedDict`, or something else, it would force it to be a regular `dict` using this method, and would preserve the original type using `copy()`.

In practice it probably doesn't make any difference.  I think I only changed it because I'm used to writing `dict()`, but I can change it back.

shin-: `dict()` would also make something like `[(1, 2)]` into `{ 1: 2 } `, transforming potentially garbage data into a dictionary, while copy() would fail there. It's probably irrelevant, so I'm fine with the change.

PR #2336, opened 2015-11-05
Add a warning when the host volume in the compose file is masked by a data volume

### LINKED ISSUE #2308: Container volume masks host volume in the compose file

dnephin:
Fixes #2308

Context in the discussion below

### DIFF
--- a/compose/service.py
+++ b/compose/service.py
@@ -918,8 +918,10 @@ def merge_volume_bindings(volumes_option, previous_container):
         if volume.external)
 
     if previous_container:
+        data_volumes = get_container_data_volumes(previous_container, volumes)
+        warn_on_masked_volume(volumes, data_volumes, previous_container.service)
         volume_bindings.update(
-            get_container_data_volumes(previous_container, volumes))
+            build_volume_binding(volume) for volume in data_volumes)
 
     return list(volume_bindings.values())
 
@@ -929,7 +931,6 @@ def get_container_data_volumes(container, volumes_option):
     a mapping of volume bindings for those volumes.
     """
     volumes = []
-
     container_volumes = container.get('Volumes') or {}
     image_volumes = [
         parse_volume_spec(volume)
@@ -949,9 +950,27 @@ def get_container_data_volumes(container, volumes_option):
 
         # Copy existing volume from old container
         volume = volume._replace(external=volume_path)
-        volumes.append(build_volume_binding(volume))
+        volumes.append(volume)
 
-    return dict(volumes)
+    return volumes
+
+
+def warn_on_masked_volume(volumes_option, container_volumes, service):
+    container_volumes = dict(
+        (volume.internal, volume.external)
+        for volume in container_volumes)
+
+    for volume in volumes_option:
+        if container_volumes.get(volume.internal) != volume.external:
+            log.warn((
+                "Service \"{service}\" is using volume \"{volume}\" from the "
+                "previous container. Host mapping \"{host_path}\" has no effect. "
+                "Remove the existing containers (with `docker-compose rm {service}`) "
+                "to use the host volume mapping."
+            ).format(
+                service=service,
+                volume=volume.internal,
+                host_path=volume.external))
 
 
 def build_volume_binding(volume_spec):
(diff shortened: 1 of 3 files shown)


shin-:
Is this warning going to show even if the host path doesn't change? I can see users being confused by that when there's nothing wrong with their service config in the first place.

dnephin:
It will show anytime the container is ignoring the `volume:` setting in the Compose file.  I can't think of a reason that you'd define a host volume mapping in the config, but for some reason want to ignore it and use the data volume from a container.

The most common case for this is when there is a `VOLUME` in their Dockerfile for a service.

If the user starts with a simple Compose file (one without volumes) and runs `docker-compose up` they will get a container with a container data volume. They later decide to add a host mapping for that service at the same path. When they run `docker-compose up` again, they'll see this warning, because the config change they made is being ignored.

If they `docker-compose rm` before running `up`, there will be no warning.

If they never `up` before adding the `volume: ` section to the service in the Compose file, there will be no warning.

I think that is how it should  be working now. It should only warn when something is actually being ignored from the Compose file.

dnephin:
Looking at the test again, I see what you're saying. I need to look at it again to see if it's actually testing the right thing.

dnephin:
I fixed up the logic and the test. It was warning anytime a host volume existed at the same path as an image volume, even if the host volume was being used.  

Now it only warns when the host volume is being ignored.

compose/service.py
```diff
@@ -951,9 +952,26 @@ def get_container_data_volumes(container, volumes_option):
...
+    container_volumes = dict(
+        (volume.internal, volume.external)
+        for volume in container_volumes)
+
+    for volume in volumes_option:
+        if container_volumes.get(volume.internal) != volume.external:
+            log.warn((
+                "Service \"{service}\" is using volume \"{volume}\" from the "
+                "previous container. Host mapping \"{host_path}\" has no effect. "
+                "Remove the existing containers to use the host volume mapping."
```
aanand: Could print the command to use here - `docker-compose rm {service}`

dnephin: Ya, I started with that, but it gets a bit messy if they've used a different name for the command. I would use `sys.argv[0]`, but that often has a full path on it, so I'd have to `dirname()` to truncate.

I could make it `"%s rm %s" % (dirname(sys.argv[0]), service)`

aanand: True, but I don't think we really need to serve that case. We already print `docker-compose` in several other places.

dnephin: ok

dnephin:
`docker-compose rm {service}` added to the warning message

aanand:
LGTM

doozr:
I'm seeing a problem with this implementation; if I start a container with a host volume mounted then change the configuration in some other way (e.g. adding an evironment variable) then I still get the warning when the container is recreated.

Furthermore, if I add a host volume that was not previously there, either as a host or data volume, then the warning appears.

In summary:

* if a container is recreated with an identical host volume, it still shows the warning even though the correct directory is mounted
* if a container is recreated with a new host volume that never previously existed as either host nor data volume, the warning appears and makes no sense.

doozr:
For example, take this original file:

    test:
        image: busybox
        volumes:
            - "/home/vagrant:/vagrant"
        environment:
            TEST: 1

If I modify it thus and run `docker-compose up -d`, I would not expect the warning, but the warning still appears:

    test:
        image: busybox
        volumes:
            - "/home/vagrant:/vagrant"
        environment:
            TEST: 2

Then, if I add a new volume that has never been declared:

    test:
        image: busybox
        volumes:
            - "/home/vagrant:/vagrant"
            - "/tmp:/hosttmp"
        environment:
            TEST: 2

Then I get a warning about both volumes.

dnephin:
Thanks for the bug report. I've got a fix out here: #2390

PR #2341, opened 2015-11-06
Only create the default network if at least one service needs it

### LINKED ISSUE #2275: Only create the default network if it's necessary
Our docs (https://github.com/docker/compose/blob/master/docs/networking.md#custom-container-network-modes) claim that we only create the network if it's going to be used, but currently we always create the network when `--x-networking` is set. 

We should only be creating it if at least one container uses the default network.

dnephin:
Fixes #2275

### DIFF
--- a/compose/project.py
+++ b/compose/project.py
@@ -300,7 +300,7 @@ class Project(object):
 
         plans = self._get_convergence_plans(services, strategy)
 
-        if self.use_networking:
+        if self.use_networking and self.uses_default_network():
             self.ensure_network_exists()
 
         return [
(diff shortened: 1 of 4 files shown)


shin-:
LGTM!

Am I the only one who feels that `mode` is a bad name for that property?

dnephin:
Ya, that property was hard to name. It has to represent the difference between each implementation. In some contexts it makes sense (host mode vs bridge mode).

dnephin:
Fixed the test, and `s/all/any/`

shin-:
:+1:

PR #2349, opened 2015-11-09
Validate all files before merging them

dnephin:
Fixes #2321

Details in the commit messages.

The third commit contains another bug fix around extends and the service_name that gets reported.

PR #2350, opened 2015-11-09
Remove name field from the config schema

### LINKED ISSUE #2299: "name" is not a valid service config field
Example:

```yaml
web:
    image: "alpine:edge"
    name: "different name"
```

This config would be valid, but shouldn't be.  I think `name` is included because we modify the config to move the name in as a field, but I think we should validate it separately as part of the "top level" validation instead.

The example config should raise an error about an unsupported key.

dnephin:
Fixes #2299 

Details in the commit message. Only the last two commits are for this issue, the rest come from #2349 . Trying to do these as separate branches would have resolved it many conflicts later, so I've based one fix on the other.

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -13,12 +13,12 @@ from requests.exceptions import ReadTimeout
 
 from .. import __version__
 from .. import legacy
+from ..config import ConfigurationError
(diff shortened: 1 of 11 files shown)


shin-:
LGTM!

PR #2351, opened 2015-11-09
Update service config_dict computation to include volumes_from mode

### LINKED ISSUE #2322: Changing the `volumes_from` mode doesn’t trigger a recreate on that service
I have a `docker-compose.yml`:

```yaml
redis:
  image: redis
  volumes_from: ["data"]
data:
  image: tianon/true
  volumes: ["/data"]
```

If I start the app with `up -d` and then change the `volumes_from` line to add a mode (e.g. `volumes_from: ["data:ro"]`), the `redis` service isn't recreated when I run `up -d` again:

```
$ docker-compose up -d
Starting counter_data_1
counter_redis_1 is up-to-date
```

shin-:
Ensure config_hash is updated when volumes_from mode is changed, and service is recreated on next up as a result.

Fixes #2322

### DIFF
--- a/compose/service.py
+++ b/compose/service.py
@@ -502,7 +502,9 @@ class Service(object):
             'image_id': self.image()['Id'],
             'links': self.get_link_names(),
             'net': self.net.id,
-            'volumes_from': self.get_volumes_from_names(),
+            'volumes_from': [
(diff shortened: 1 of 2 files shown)


dnephin:
LGTM

PR #2355, opened 2015-11-10
Refactor process_error and include a filename as part of the validation error messages

### LINKED ISSUE #2323: Validation errors should specify the file they originate from
Right now, validation errors don't include the filename, which will make problems hard to track down if you're using multiple files.

Of course, we can't validate files in isolation because override files aren't required to be valid standalone Compose files, which makes this complicated.

Perhaps we could do something like this (pseudocode):

```python
config = load_config(filenames[0])
# on failure, will output an error message specifying the name of the first file
validate_config(config, filenames[0])

for filename in filenames[1:]:
    config = merge_config(config, filename)
    # on failure, will output an error message specifying the name of the current file
    validate_config(config, filename)
```

dnephin:
Fixes #2323 

The refactoring started as a way to reduce the complexity of `process_errors()`. It started as a function with cyclomatic complexity of 26! I refactored it into a few smaller functions, each of which is under a cyclomatic complexity of 8.

Part of the refactoring included adding ids to some of the schemas. We were guessing at which schema was being validated by arbitrary hints (the type of path). By using a schema id we're much less likely to break these validation messages in the future, as we add new schemas and sub-schemas.

The last commit adds filenames to validation messages, and adjusts the tests to match.

### DIFF
--- a/compose/config/fields_schema.json
+++ b/compose/config/fields_schema.json
@@ -2,15 +2,18 @@
   "$schema": "http://json-schema.org/draft-04/schema#",
 
   "type": "object",
+  "id": "fields_schema.json",
 
   "patternProperties": {
     "^[a-zA-Z0-9._-]+$": {
       "$ref": "#/definitions/service"
     }
   },
+  "additionalProperties": false,
 
   "definitions": {
     "service": {
+      "id": "#/definitions/service",
       "type": "object",
 
       "properties": {
@@ -167,6 +170,5 @@
       ]
     }
 
-  },
-  "additionalProperties": false
+  }
 }
--- a/compose/config/config.py
+++ b/compose/config/config.py
@@ -229,9 +229,9 @@ def load(config_details):
 
 
 def process_config_file(config_file, service_name=None):
-    validate_top_level_object(config_file.config)
+    validate_top_level_object(config_file)
     processed_config = interpolate_environment_variables(config_file.config)
-    validate_against_fields_schema(processed_config)
+    validate_against_fields_schema(processed_config, config_file.filename)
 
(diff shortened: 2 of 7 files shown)


compose/config/fields_schema.json
```diff
@@ -2,15 +2,18 @@
   "$schema": "http://json-schema.org/draft-04/schema#",
 
   "type": "object",
+  "id": "fields_schema.json",
```
shin-: I'm not too fond of having the filename being repeated here, especially because this object describes the root schema (dictionary of services).

Alternatives: `services`, `compose_services`, `services_list` ?

compose/config/fields_schema.json
```diff
@@ -2,15 +2,18 @@
...
   "patternProperties": {
     "^[a-zA-Z0-9._-]+$": {
       "$ref": "#/definitions/service"
     }
   },
+  "additionalProperties": false,
 
   "definitions": {
     "service": {
+      "id": "#/definitions/service",
```
shin-: If we change the path of that definition to something else we'll have to change the ID as well, which goes against the idea of what an ID is supposed to be.

Alternatives: `service_definition`, `service_description`, `service_properties`?

dnephin:
rebased out the intermediate commits, just waiting on the build to run

dnephin:
We chatted about the schema id issue offline. I would agree that it seems redundant (and more brittle) to use an id that patches the path, but according to https://github.com/json-schema/json-schema/wiki/The-%22id%22-conundrum#how-to-fix-that this is the correct  way to handle ids.

shin-:
LGTM :+1:

PR #2360, opened 2015-11-10
Use exit code 1 when encountering a ReadTimeout

### LINKED ISSUE #2357: When COMPOSE_HTTP_TIMEOUT is reached the exit code should be non-zero
We often run into timeout issues when using `docker-compose` in [circleci](https://circleci.com/) to run unit tests, for example:

```
$docker-compose run yiic create_unit_test_db
ERROR: An HTTP request took too long to complete. Retry with --verbose to obtain debug information.
If you encounter this issue regularly because of slow network conditions, consider setting COMPOSE_HTTP_TIMEOUT to a higher value (current value: 60).
```

The problem is that the exit code that `docker-compose` returns is zero, meaning the build continues, whereas it should be stopped.

---

compose v1.5.0. It was installed with `sudo pip install --upgrade docker-compose` as part of the `circle.yml` file.

shin-:
Fixes #2357

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -80,6 +80,7 @@ def main():
             "If you encounter this issue regularly because of slow network conditions, consider setting "
             "COMPOSE_HTTP_TIMEOUT to a higher value (current value: %s)." % HTTP_TIMEOUT
         )
+        sys.exit(1)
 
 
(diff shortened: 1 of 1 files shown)


dnephin:
LGTM

PR #2373, opened 2015-11-11
Allow dashes in environment variable names

### LINKED ISSUE #2370: Need support for dashes in environment variable names
The Java spring-boot library accepts core config through a standardized set of environment variables. Spring boot provides a mechanism to convert their properties to environment, [see here](http://docs.spring.io/spring-boot/docs/current/reference/html/howto-properties-and-configuration.html#howto-change-the-location-of-external-properties). Some of these standard spring properties have dashes in them.

These properties worked under previous versions of compose but are no longer allowed in 1.5. Running a compose file like this

```
app:
  build: .
  expose:
   - "80"
  links:
    - db:db
  environment:
    - SERVER_PORT=80
    - SPRING_DATASOURCE_URL=jdbc:postgresql://db:5432/demo
    - SPRING_DATASOURCE_USERNAME=username
    - SPRING_DATASOURCE_PASSWORD=password
    - SPRING_DATASOURCE_DRIVER-CLASS-NAME=org.postgresql.Driver
    - SPRING_JPA_HIBERNATE_DDL-AUTO=none
```

Now fails with the error 

`Service 'app' configuration key 'environment' contains unsupported option: 'SPRING_JPA_HIBERNATE_DDL-AUTO'`

shin-:
See http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap08.html

> Environment variable names used by the utilities in the Shell and Utilities volume of POSIX.1-2008 consist solely of uppercase letters, digits, and the <underscore> ( '_' ) from the characters defined in Portable Character Set and do not begin with a digit. **Other characters may be permitted by an implementation; applications shall tolerate the presence of such names.**

Fixes #2370

### DIFF
--- a/compose/config/fields_schema.json
+++ b/compose/config/fields_schema.json
@@ -40,7 +40,7 @@
             {
               "type": "object",
               "patternProperties": {
-                "^[^-]+$": {
+                ".+": {
                   "type": ["string", "number", "boolean", "null"],
                   "format": "environment"
                 }
--- a/tests/unit/config/config_test.py
+++ b/tests/unit/config/config_test.py
@@ -470,20 +470,18 @@ class ConfigTest(unittest.TestCase):
         self.assertTrue(mock_logging.warn.called)
         self.assertTrue(expected_warning_msg in mock_logging.warn.call_args[0][0])
 
-    def test_config_invalid_environment_dict_key_raises_validation_error(self):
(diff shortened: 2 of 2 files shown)


dnephin:
LGTM once build is green

thaJeztah:
ah, yes, we recently removed all validation for env-vars. although many names are not valid according to the specs, we discussed that and decided that it's the responsibility of the software running *inside* the container to handle invalid names, not docker

PR #2389, opened 2015-11-13
Use uname to build target name for different platforms

StefanScherer:
In preparation to build docker-compose for Linux on ARM we should replace the fixed target name with the values from `uname -s` and `uname -m`.


dnephin:
LGTM

shin-:
LGTM!

PR #2390, opened 2015-11-13
Fix extra warnings on masked volumes

dnephin:
Fixes https://github.com/docker/compose/pull/2336#issuecomment-156423384

### DIFF
--- a/tests/unit/service_test.py
+++ b/tests/unit/service_test.py
@@ -26,6 +26,8 @@ from compose.service import parse_volume_spec
 from compose.service import Service
 from compose.service import ServiceNet
 from compose.service import VolumeFromSpec
+from compose.service import VolumeSpec
+from compose.service import warn_on_masked_volume
 
 
 class ServiceTest(unittest.TestCase):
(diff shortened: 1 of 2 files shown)


shin-:
LGTM!

tests/unit/service_test.py
```diff
@@ -750,6 +752,39 @@ def test_different_host_path_in_container_json(self):
...
+        container_volumes = [
+            VolumeSpec('/var/lib/docker/path', '/path', 'rw'),
+            VolumeSpec('/var/lib/docker/path', '/other', 'rw'),
+        ]
+        service = 'service_name'
+
+        with mock.patch('compose.service.log') as mock_log:
+            warn_on_masked_volume(volumes_option, container_volumes, service)
+
+        mock_log.warn.called_once_with(mock.ANY)
```
aanand: Does this not need to be `assert`-ed?

dnephin: Oops, yes, and the patches should be using `autospec`. Opening a PR for that now

PR #2392, opened 2015-11-13
docker-compose events

### LINKED ISSUE #1510: Proposal: docker-compose events
There have been a few requests for supporting some form of "hooks" system within compose (#74, #1341).

A feature which runs commands on the host would add a lot of complexity to compose and the compose configuration. Another option is to support these features by providing a way for external tools to run the command triggered by an event.

[docker events](http://docs.docker.com/reference/api/docker_remote_api_v1.18/#monitor-dockers-events) provides an event stream of all docker events, but would still require filtering.

`docker-compose events` could provide a similar interface by filtering the events returned by `/events` and return a stream of only the events related to the active compose project. The event stream could also include a new field `"service": <service_name>` in addition to the fields already provided by `/events`

dnephin:
Fixes #1510 

Adds a new command `events` for tailing the event stream for containers in the composition.

The event fields match the new api fields being introduced in 1.10 (https://github.com/docker/docker/pull/18888/)

This gist shows an example of using these events to create event hooks in under 20 lines of bash:
https://gist.github.com/dnephin/c56b479810f863527e90

Any suggestion on where this can be included in the docs?

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -256,8 +256,10 @@ class TopLevelCommand(DocoptCommand):
             --json      Output events as a stream of json objects
         """
         def format_event(event):
-            return ("{time}: service={service} event={event} "
-                    "container={container} image={image}").format(**event)
+            attributes = ["%s=%s" % item for item in event['attributes'].items()]
+            return ("{time} {type} {action} {id} ({attrs})").format(
+                attrs=", ".join(sorted(attributes)),
+                **event)
 
         def json_format_event(event):
             event['time'] = event['time'].isoformat()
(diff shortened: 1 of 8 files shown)


dnephin:
I've updated this PR, and included an example of using the events to create the often requested  "hooks" feature.

dnephin:
I've noticed that I have to use `PYTHONUNBUFFERED=1` to get these to flush properly. We might need to add some force flushing to stdout.

dnephin:
Ok, I've added the `flush()` so that users won't have to know about `PYTHONUNBUFFERED=1`

aanand:
Nice. Tested locally, looks great.

LGTM

CodeTahoe:
@dnephin re: docs - Maybe it can go in it's own section as "events" under bash completion under the heading of "create multi-container applications"

Issue #2008, opened 2015-09-08
Improve validation schema for list_or_dict fields


dnephin:
Right now `extra_hosts` and `labels` use "one of: object or array", but no validation is done on the contents of the `object`.

I think this can be improved by creating a schema for the object that uses `patternProperties`,  so that it will fail if someone uses nested objects, or arrays within the object.

```json
{
  "type": "object",
  "patternProperties": {"type": "string"},
}
```

dnephin:
Apparently you've already have a fix for  this in #2000!

dnephin:
I've updated this issue to be about the `list_or_dict` schema, since it has the same issue, and isn't fixed by #2000

PR #2405, opened 2015-11-17
Move all validation to config

dnephin:
To support #1953 we need to be able to fully validate the configuration without creating a client, so all validation needs to be moved to the config package.

Includes a fix for #2008

This removes the deprecated `service.ConfigError`, and moves all the validation to `config`. As part of this change, there is a new `finalize_service()` function which takes care of creating domain objects for things like Volumes, VolumesFrom, etc. By creating these objects during the config loading we don't have to re-parse the config again from the project and service modules.

### DIFF
--- a/compose/cli/command.py
+++ b/compose/cli/command.py
@@ -14,7 +14,6 @@ from . import errors
 from . import verbose_proxy
 from .. import config
 from ..project import Project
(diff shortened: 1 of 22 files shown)


aanand:
LGTM

PR #2407, opened 2015-11-17
New `docker-compose config` command

### LINKED ISSUE #1953: Feature Request: `docker-compose config` Command

dnephin:
Fixes #1953 

Adds a new command for validating and printing the Compose file

### DIFF
--- a/compose/cli/command.py
+++ b/compose/cli/command.py
@@ -46,7 +46,7 @@ def friendly_error_message():
 def project_from_options(base_dir, options):
     return get_project(
         base_dir,
-        get_config_path(options.get('--file')),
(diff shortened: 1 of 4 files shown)


aanand:
Needs a rebase, but otherwise looks great.

dnephin:
Rebased, added the `--quiet` alias, and fixed the wording (I think)

aanand:
LGTM

PR #2409, opened 2015-11-17
Fix texttable dep. 0.8.2 was removed from pypi.

dnephin:
The old version isn't available anymore


shin-:
LGTM as long as tests pass :+1:

PR #2417, opened 2015-11-18
Add the git sha to version output

### LINKED ISSUE #2377: docker-compose version should show git hash
It would be nice if `docker-compose --version` showed also the Git SHA which the "binary" was compiled with.  Having some struggles with compiling Compose, I wanted to verify that I was in fact using the correct version.

dnephin:
Fixes #2377

cc @nathanleclaire

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@
 /docs/_site
 /venv
 README.rst
+compose/GITSHA
--- a/Dockerfile.run
+++ b/Dockerfile.run
@@ -8,6 +8,6 @@ COPY    requirements.txt /code/requirements.txt
 RUN     pip install -r /code/requirements.txt
 
 ADD     dist/docker-compose-release.tar.gz /code/docker-compose
-RUN     pip install /code/docker-compose/docker-compose-*
+RUN     pip install --no-deps /code/docker-compose/docker-compose-*
 
 ENTRYPOINT ["/usr/bin/docker-compose"]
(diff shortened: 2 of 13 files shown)


nathanleclaire:
Nice Dan!

shin-:
LGTM!

aanand:
Does this change preserve compatibility with the [completion installation commands](https://github.com/docker/compose/blob/master/docs/completion.md), which depend on the output of `docker-compose --version`?

dnephin:
I don't see anything in the completion script that reads the `--version` output. I also tried them both out quickly, and didn't have any issues.

@albers did I miss something?

aanand:
It's the installation instructions in `docs/completion.md` that I'm worried about - e.g. for bash:

     curl -L https://raw.githubusercontent.com/docker/compose/$(docker-compose --version | awk 'NR==1{print $NF}')/contrib/completion/bash/docker-compose > /etc/bash_completion.d/docker-compose

The `docker-compose --version | awk 'NR==1{print $NF}'` part might break if the output of `--version` changes.

dnephin:
Oh, I missed that, sorry. 

It actually still works. The `awk` expression returns the short git sha, which is valid for that part of the url: https://raw.githubusercontent.com/docker/compose/e406702/contrib/completion/bash/docker-compose

albers:
The download URL was constructed before `docker-compose version --short` existed.
I suggest to use that expression in the download URL. It's shorter and safer.

albers:
created #2512

PR #2420, opened 2015-11-18
run.sh script: Also pass DOCKER_TLS_VERIFY and DOCKER_CERT_PATH env vars to compose container

simonvanderveldt:
The script as is doesn't pass `DOCKER_TLS_VERIFY` and `DOCKER_CERT_PATH` which means it fails when used with a TLS enabled docker daemon (like the one created by machine for local development).

Signed-off-by: Simon van der Veldt <simon.vanderveldt@gmail.com>


dnephin:
LGTM

shin-:
LGTM

PR #2421, opened 2015-11-19
Add support for declaring named volumes in compose files

### LINKED ISSUE #2110: Add support for volume declaration in compose file
Using the volume API added in Docker 1.9 (recently implemented support in docker-py, see https://github.com/docker/docker-py/pull/764)

shin-:
* Bump default API version to 1.21 (required for named volume management)
* Introduce new, versioned compose file format while maintaining support
  for current (legacy) format
* Test updates to reflect changes made to the internal API

Fixes #2110, replaces #2113 

WIP:
- [x] Tests
- [x] Documentation

### DIFF
--- a/compose/config/config.py
+++ b/compose/config/config.py
@@ -10,6 +10,7 @@ from collections import namedtuple
 import six
 import yaml
 
+from ..const import COMPOSEFILE_VERSIONS
 from .errors import CircularReference
 from .errors import ComposeFileNotFound
 from .errors import ConfigurationError
@@ -24,6 +25,7 @@ from .validation import validate_against_fields_schema
 from .validation import validate_against_service_schema
 from .validation import validate_extends_file_path
 from .validation import validate_top_level_object
+from .validation import validate_top_level_service_objects
 
 
 DOCKER_CONFIG_KEYS = [
@@ -161,13 +163,24 @@ def find(base_dir, filenames):
 
 def get_config_version(config_details):
     def get_version(config):
-        validate_top_level_object(config)
-        return config.config.get('version')
+        if config.config is None:
+            return None
+        version = config.config.get('version', 1)
+        if isinstance(version, dict):
+            version = 1
+        return version
+
     main_file = config_details.config_files[0]
+    validate_top_level_object(main_file)
     version = get_version(main_file)
     for next_file in config_details.config_files[1:]:
+        validate_top_level_object(next_file)
         next_file_version = get_version(next_file)
-        if version != next_file_version:
+        if version is None:
+            version = next_file_version
+            continue
+
+        if version != next_file_version and next_file_version is not None:
             raise ConfigurationError(
                 "Version mismatch: main file {0} specifies version {1} but "
                 "extension file {2} uses version {3}".format(
@@ -224,6 +237,9 @@ def load(config_details):
     Return a fully interpolated, extended and validated configuration.
     """
     version = get_config_version(config_details)
+    if version not in COMPOSEFILE_VERSIONS:
+        raise ConfigurationError('Invalid config version provided: {0}'.format(version))
+
     processed_files = []
     for config_file in config_details.config_files:
         processed_files.append(
@@ -231,9 +247,10 @@ def load(config_details):
         )
     config_details = config_details._replace(config_files=processed_files)
 
-    if not version or isinstance(version, dict):
+    if version == 1:
         service_dicts = load_services(
-            config_details.working_dir, config_details.config_files
+            config_details.working_dir, config_details.config_files,
+            version
         )
         volumes = {}
     elif version == 2:
@@ -242,11 +259,9 @@ def load(config_details):
             for f in config_details.config_files
         ]
         service_dicts = load_services(
-            config_details.working_dir, config_files
+            config_details.working_dir, config_files, version
         )
         volumes = load_volumes(config_details.config_files)
-    else:
-        raise ConfigurationError('Invalid config version provided: {0}'.format(version))
 
     return Config(version, service_dicts, volumes)
 
@@ -259,14 +274,14 @@ def load_volumes(config_files):
     return volumes
 
 
-def load_services(working_dir, config_files):
+def load_services(working_dir, config_files, version):
     def build_service(filename, service_name, service_dict):
         service_config = ServiceConfig.with_abs_paths(
             working_dir,
             filename,
             service_name,
             service_dict)
-        resolver = ServiceExtendsResolver(service_config)
+        resolver = ServiceExtendsResolver(service_config, version)
         service_dict = process_service(resolver.run())
 
         # TODO: move to validate_service()
@@ -301,8 +316,8 @@ def load_services(working_dir, config_files):
 
 
 def process_config_file(config_file, service_name=None, version=None):
-    validate_top_level_object(config_file)
-    processed_config = interpolate_environment_variables(config_file.config)
+    validate_top_level_service_objects(config_file, version)
+    processed_config = interpolate_environment_variables(config_file.config, version)
     validate_against_fields_schema(
         processed_config, config_file.filename, version
     )
@@ -316,10 +331,11 @@ def process_config_file(config_file, service_name=None, version=None):
 
 
 class ServiceExtendsResolver(object):
-    def __init__(self, service_config, already_seen=None):
+    def __init__(self, service_config, version, already_seen=None):
         self.service_config = service_config
         self.working_dir = service_config.working_dir
         self.already_seen = already_seen or []
+        self.version = version
 
     @property
     def signature(self):
@@ -348,7 +364,8 @@ class ServiceExtendsResolver(object):
 
         extended_file = process_config_file(
             ConfigFile.from_filename(config_path),
-            service_name=service_name)
+            service_name=service_name, version=self.version
+        )
         service_config = extended_file.config[service_name]
         return config_path, service_config, service_name
 
@@ -359,6 +376,7 @@ class ServiceExtendsResolver(object):
                 extended_config_path,
                 service_name,
                 service_dict),
+            self.version,
             already_seen=self.already_seen + [self.signature])
 
         service_config = resolver.run()
--- a/compose/config/config.py
+++ b/compose/config/config.py
@@ -118,6 +118,9 @@ class ConfigFile(namedtuple('_ConfigFile', 'filename config')):
     def from_filename(cls, filename):
         return cls(filename, load_yaml(filename))
 
+    def get_service_dicts(self, version):
+        return self.config if version == 1 else self.config.get('services', {})
+
 
 class Config(namedtuple('_Config', 'version services volumes')):
     """
@@ -164,9 +167,11 @@ def find(base_dir, filenames):
 def get_config_version(config_details):
     def get_version(config):
         if config.config is None:
-            return None
+            return 1
         version = config.config.get('version', 1)
         if isinstance(version, dict):
+            # in that case 'version' is probably a service name, so assume
+            # this is a legacy (version=1) file
             version = 1
         return version
 
@@ -176,9 +181,6 @@ def get_config_version(config_details):
     for next_file in config_details.config_files[1:]:
         validate_top_level_object(next_file)
         next_file_version = get_version(next_file)
-        if version is None:
-            version = next_file_version
-            continue
 
         if version != next_file_version and next_file_version is not None:
             raise ConfigurationError(
@@ -316,8 +318,16 @@ def load_services(working_dir, config_files, version):
 
 
 def process_config_file(config_file, version, service_name=None):
-    validate_top_level_service_objects(config_file, version)
-    processed_config = interpolate_environment_variables(config_file.config, version)
+    service_dicts = config_file.get_service_dicts(version)
+    validate_top_level_service_objects(
+        config_file.filename, service_dicts
+    )
+    interpolated_config = interpolate_environment_variables(service_dicts)
+    if version == 2:
+        processed_config = dict(config_file.config)
+        processed_config.update({'services': interpolated_config})
+    if version == 1:
+        processed_config = interpolated_config
     validate_against_fields_schema(
         processed_config, config_file.filename, version
     )
--- a/compose/config/config.py
+++ b/compose/config/config.py
@@ -315,7 +315,7 @@ def load_services(working_dir, config_files, version):
     return build_services(config_file)
 
 
-def process_config_file(config_file, service_name=None, version=None):
+def process_config_file(config_file, version, service_name=None):
     validate_top_level_service_objects(config_file, version)
     processed_config = interpolate_environment_variables(config_file.config, version)
     validate_against_fields_schema(
@@ -364,7 +364,7 @@ class ServiceExtendsResolver(object):
 
         extended_file = process_config_file(
             ConfigFile.from_filename(config_path),
-            service_name=service_name, version=self.version
+            version=self.version, service_name=service_name
         )
         service_config = extended_file.config[service_name]
         return config_path, service_config, service_name
--- a/compose/config/config.py
+++ b/compose/config/config.py
@@ -117,6 +117,17 @@ class ConfigFile(namedtuple('_ConfigFile', 'filename config')):
         return cls(filename, load_yaml(filename))
 
 
+class Config(namedtuple('_Config', 'version services volumes')):
+    """
+    :param version: configuration version
+    :type  version: int
+    :param services: List of service description dictionaries
+    :type  services: :class:`list`
+    :param volumes: List of volume description dictionaries
+    :type  volumes: :class:`list`
+    """
+
+
 class ServiceConfig(namedtuple('_ServiceConfig', 'working_dir filename name config')):
 
     @classmethod
@@ -146,6 +157,24 @@ def find(base_dir, filenames):
     return ConfigDetails(
         os.path.dirname(filenames[0]),
         [ConfigFile.from_filename(f) for f in filenames])
+
+
+def get_config_version(config_details):
+    def get_version(config):
+        validate_top_level_object(config)
+        return config.config.get('version')
+    main_file = config_details.config_files[0]
+    version = get_version(main_file)
+    for next_file in config_details.config_files[1:]:
+        next_file_version = get_version(next_file)
+        if version != next_file_version:
+            raise ConfigurationError(
+                "Version mismatch: main file {0} specifies version {1} but "
+                "extension file {2} uses version {3}".format(
+                    main_file.filename, version, next_file.filename, next_file_version
+                )
+            )
+    return version
 
 
 def get_default_config_files(base_dir):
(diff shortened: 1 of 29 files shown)


compose/config/config.py
```diff
@@ -217,20 +282,20 @@ def merge_services(base, override):
...
+    config_file = config_files[0]
+    for next_file in config_files[1:]:
         config = merge_services(config_file.config, next_file.config)
         config_file = config_file._replace(config=config)
 
     return build_services(config_file)
 
 
-def process_config_file(config_file, service_name=None):
+def process_config_file(config_file, service_name=None, version=None):
```
dnephin: I think `ServiceExtendsResolver.validate_and_construct_extends()` will need to call this with a version as well. 

Since `version` is a required parameter can we make it a position arg instead of a kwarg?

compose/config/config.py
```diff
@@ -133,6 +144,24 @@ def find(base_dir, filenames):
         [ConfigFile.from_filename(f) for f in filenames])
 
 
+def get_config_version(config_details):
+    def get_version(config):
+        validate_top_level_object(config)
+        return config.config.get('version')
```
dnephin: Could we set a default version of `1` here, so that the rest of the code dones't need to know about the "no version is version 1" assumption?  We can keep that logic in this one place. 

I guess that will require a `git mv` of the fields schema file.

compose/config/fields_schema_v2.json
```diff
@@ -0,0 +1,43 @@
...
+  "properties": {
+    "version": {
+      "enum": [2]
+    },
+    "services": {
+      "type": "object",
+      "patternProperties": {
+        "^[a-zA-Z0-9._-]+$": {
+          "$ref": "fields_schema.json#/definitions/service"
+        }
```
dnephin: I'm hoping we'll be able to refactor the schema so that this service name constraint only exists in once place (instead of both here and the v1 schema).  I think it can wait until later though.

compose/config/fields_schema_v2.json
```diff
@@ -0,0 +1,43 @@
...
+
+  "definitions": {
+    "volume": {
+      "type": "object",
+      "properties": {
+        "driver": {"type": "string"},
+        "driver_opts": {
+          "type": "object",
+          "patternProperties": {
+            "^.+$": {"type": ["boolean", "string", "number"]}
```
dnephin: I think we probably want to remove `"boolean"` from this list. We include it in other sections for backwards compatibility, but it often doesn't do what the user wants (ex: it can translate `yes`, `y`, `True` into `true` without the user knowing about it).

compose/config/validation.py
```diff
@@ -269,10 +269,13 @@ def format_error_message(error, service_name):
     return '\n'.join(format_error_message(error, service_name) for error in errors)
 
 
-def validate_against_fields_schema(config, filename):
+def validate_against_fields_schema(config, filename, version=None):
+    schema_filename = "fields_schema.json"
+    if version:
+        schema_filename = "fields_schema_v{0}.json".format(version)
```
dnephin: I think with the suggested change to making a default version of 1 we can drop the `if` here, and make `version` a required position arg.

compose/volume.py
```diff
@@ -0,0 +1,19 @@
...
+class Volume(object):
+    def __init__(self, client, project, name, driver=None, driver_opts=None):
+        self.client = client
+        self.project = project
+        self.name = name
+        self.driver = driver
+        self.driver_opts = driver_opts
+
+    def create(self):
+        return self.client.create_volume(self.name, self.driver, self.driver_opts)
```
dnephin: What happens if the volume already exists? Is this just a No-op?

Probably good for a test case

docs/compose-file.md
```diff
@@ -24,6 +24,64 @@ As with `docker run`, options specified in the Dockerfile (e.g., `CMD`,
 `EXPOSE`, `VOLUME`, `ENV`) are respected by default - you don't need to
 specify them again in `docker-compose.yml`.
 
+## Versioning
+
+It is possible to use different versions of the `compose.yml` format.
+Below are the formats currently supported by compose.
```
dnephin: I think we might want to start by only talking about the latest version of the config up front, with a section at the bottom which covers v1 (and maybe a link to the V1 section from here).

We'll have to include some notes about how to migrate to the v2 config in the section at the bottom.

Does that sound right @moxiegirl ?

compose/project.py
```diff
@@ -115,6 +117,14 @@ def from_dicts(cls, name, service_dicts, client, use_networking=False, network_d
                     net=net,
                     volumes_from=volumes_from,
                     **service_dict))
+        if config_data.volumes:
+            for vol_name, data in config_data.volumes.items():
+                project.volumes.append(
+                    Volume(
+                        client=client, project=name, name=vol_name,
+                        driver=data.get('driver'), driver_opts=data.get('driver_opts')
+                    )
```
dnephin: I think we will probably need to prefix the volume_name here with the project name, so that we continue to support multiple isolated environments on the same host (documented as one of the key features of compose http://docs.docker.com/compose/#Multiple-isolated-environments-on-a-single-host).

compose/config/config.py
```diff
@@ -188,10 +217,46 @@ def load(config_details):
...
     """
+    version = get_config_version(config_details)
+    processed_files = []
+    for config_file in config_details.config_files:
+        processed_files.append(
+            process_config_file(config_file, version=version)
+        )
+    config_details = config_details._replace(config_files=processed_files)
+
+    if not version or isinstance(version, dict):
```
dnephin: When would version be a `dict` ?

compose/config/config.py
```diff
@@ -188,10 +217,46 @@ def load(config_details):
...
+        service_dicts = load_services(
+            config_details.working_dir, config_details.config_files
+        )
```
dnephin: Since this function takes both parts of a `config_details`, I would just have it accept a single arg of `config_details`  instead of splitting it here. What do you think?

```python
service_dicts = load_services(config_details)
```

compose/config/config.py
```diff
@@ -133,6 +144,24 @@ def find(base_dir, filenames):
         [ConfigFile.from_filename(f) for f in filenames])
 
 
+def get_config_version(config_details):
+    def get_version(config):
+        validate_top_level_object(config)
```
dnephin: This `validate_top_level_object()` is now called twice for the same config. Once here, and again below as part of `process_config_file()`

It would be nice to only have it happen once, and I think this will actually fail now that `validate_top_level_object()` also validates that the top level keys are services names with values of objects.

I guess `validate_top_level_object()` could be split up again. The first step of `load()` could be to validate that every `ConfigFile` has a valid `config` dict. Then this `get_config_version()` can proceed without doing any more validation, and `process_config_file()` can call the rest of validation.

docs/compose-file.md
moxiegirl: @dnephin Agreed the organization is off.  `## Service configuration reference`  I would most likely put under this heading a paragraph about the fact there are multiple versions.  Then put a cross -reference to the version section to follow.

Then move `## Versioning` to follow the main body of configuration material

This section would need to answer:

* define whether the configuration version matches the product version 
* under what conditions can I use  a different version; for example does the version of the configuration file match or differ from the Compose version
* can I mix syntax in the same config file (prob not but should say)
* can I use two config files that use differing versions of the syntax (I'm thinking also of how the override files interact with the main config under the version conditions)
* provide some guidance to the reader under what circumstances this version mixing is useful

I'll work with the new compose writer @londoncalling to build this suggested organization with @shin- .

compose/project.py
shin-: Updated volume creation to be prefixed with the project name (logic in the Volume class)

shin-:
I think I addressed most of the points brought up in the earlier review, except for the documentation part.

compose/config/config.py
```diff
@@ -149,6 +162,35 @@ def find(base_dir, filenames):
...
+
+    main_file = config_details.config_files[0]
+    validate_top_level_object(main_file)
+    version = get_version(main_file)
+    for next_file in config_details.config_files[1:]:
+        validate_top_level_object(next_file)
+        next_file_version = get_version(next_file)
+        if version is None:
+            version = next_file_version
+            continue
```
aanand: If I'm reading this right, you could have a legacy main file and a version-1 override file, and the ultimate version would be 1.

It scares me that the version decided upon is dependent on the override files. I think it'd be better if we enforced that all files must be the same version.

aanand: OK, I think I might be misreading this. The possible versions are actually 1 and 2, so what does it mean when we return `None` from `get_version`?

compose/project.py
```diff
@@ -205,6 +215,15 @@ def kill(self, service_names=None, **options):
     def remove_stopped(self, service_names=None, **options):
         parallel.parallel_remove(self.containers(service_names, stopped=True), options)
 
+    def initialize_volumes(self):
+        try:
+            for volume in self.volumes:
+                volume.create()
+        except NotFound:
+            raise ConfigurationError(
+                'Volume %s sepcifies nonexistent driver %s' % (volume.name, volume.driver)
```
aanand: Typo: `sepcifies`

tests/integration/project_test.py
```diff
@@ -439,3 +445,85 @@ def test_unscale_after_restart(self):
...
+
+    def test_project_up_volumes(self):
+        vol_name = 'composetests_{0:x}'.format(random.getrandbits(32))
+        full_vol_name = 'composetest_{0}'.format(vol_name)
+        config_data = config.Config(
+            2, [{
+                'name': 'web',
+                'image': 'busybox:latest',
+                'command': 'top'
+            }], {vol_name: {'driver': 'local'}}
```
aanand: Keyword args would make this a bit more readable

tests/integration/project_test.py
```diff
@@ -439,3 +445,85 @@ def test_unscale_after_restart(self):
         project.up()
         service = project.get_service('web')
         self.assertEqual(len(service.containers()), 1)
+
+    def test_project_up_volumes(self):
+        vol_name = 'composetests_{0:x}'.format(random.getrandbits(32))
```
aanand: Bit strange that this comes out as `composetest_composetests_<hexchars>` - I don't think two prefixes are necessary

tests/integration/project_test.py
```diff
@@ -439,3 +445,85 @@ def test_unscale_after_restart(self):
...
+        project = Project.from_config(
+            name='composetest',
+            config_data=config_data, client=self.client
+        )
+        project.up()
+        self.assertEqual(len(project.containers()), 1)
+
+        volume_data = self.client.inspect_volume(full_vol_name)
+        self.assertEqual(volume_data['Name'], full_vol_name)
+        self.assertEqual(volume_data['Driver'], 'local')
```
aanand: Would be good to have an integration test which ensures that we don't attempt to create volumes again. (I know there's a lower-level one in `volume_test.py` but it'd be good to have a higher-level test in case of refactoring)

aanand:
Looking really good.

We need to decide what to do if the user changes volume configuration (e.g. switches drivers). We should definitely *not* just destroy it and recreate it with the new config - perhaps by default we should show an error, telling the user to remove the existing volume if they're happy to do so.

aanand:
Just realised a problem with the above: if the user hasn't specified a driver, then we don't know what the "correct" driver is, which means that we don't know whether it needs to change.

There are analogous problems with container config (e.g. auto-generated hostnames), which is why we went down the `config_hash` route, but that depends on labels, and I don't believe volumes can have labels.

We might have a similar problem with networks, too.

compose/config/config.py
shin-: When one config file is empty, as tested [here](https://github.com/docker/compose/blob/master/tests/unit/config/config_test.py#L173). In that case we can't assume the version to be a specific one, but we don't want to fail either.

shin-:
So, from my investigation, the docker API will raise an error if you try to change the driver for an existing volume:

```
In [3]: c.create_volume('skystar', driver='local')
[snip]
APIError: 500 Server Error: Internal Server Error ("A volume name skystar already exists with the nfs driver. Choose a different volume name.")
```

So, it would simply be a matter of handling this error gracefully. Issuing a create on a volume that already exists with the same driver is essentially a no-op (although depending on the driver implementation, it may let you update the driver options. It seems to be the case for [docker-volume-netshare](https://github.com/gondor/docker-volume-netshare/blob/25a88a2f8dbf4b317a802709ab6edc1e444b2ca2/netshare/drivers/mounts.go#L64-L65) which I'm using to test out the behavior).

TL;DR proposal:
- Always issue a create on the defined volumes during the `up` phase.
- Gracefully handle API 500 if driver changes in the config.

@aanand

dnephin:
How do we gracefully handle the 500? Is that a warning or an error?  

I think the proposal sounds good. Avoiding the "if not exists: create" allows us to avoid possible race conditions with other clients.  Let's check with someone familiar with the volume driver to make sure it matches their expectations.

cc @cpuguy83 , does the proposal from @shin- in the comment above match your expectations for  volumes drivers? (A create for an existing volume is a no-op when the driver/options are identical).

cpuguy83:
It's correct.
We also currently don't look at opts.
Names are unique, and if you try to create a new volume with the same name against a different driver it will fail.
Same name, same driver (and Docker has it in memory) will be a no-op. -- this is the weird bit...

compose/project.py
erikh: can we override this somehow? We prefix our volumes in volplugin with the tenant name, and I'm fairly certain others are doing similar syntaxes with the volume name.

compose/config/config.py
aanand: Can't we just assume the version is 1? An "empty" v2 file should look like this, in my opinion:

```yaml
version: 2
```

dmp42:
> I'll work with the new compose writer @londoncalling to build this suggested organization with @shin- .

Ping @moxiegirl @londoncalling @shin- what's the status on this?

compose/config/config.py
shin-: If you think that's better, that's an easy change to make.

shin-: Updated af9291e664fcfc17498ded4fa9574c4ea7e0a5d8

compose/config/config.py
```diff
@@ -149,6 +162,32 @@ def find(base_dir, filenames):
         [ConfigFile.from_filename(f) for f in filenames])
 
 
+def get_config_version(config_details):
+    def get_version(config):
+        if config.config is None:
+            return 1
+        version = config.config.get('version', 1)
+        if isinstance(version, dict):
+            version = 1
```
dnephin: In what case would `config.config['version']`  be a dict?

Isn't this either `None` (because the `version` key doesn't exist) or an int?

shin-: If the user for some reason has a service called version in a legacy compose file.

compose/config/config.py
```diff
@@ -149,6 +162,32 @@ def find(base_dir, filenames):
...
+        next_file_version = get_version(next_file)
+
+        if version != next_file_version and next_file_version is not None:
+            raise ConfigurationError(
+                "Version mismatch: main file {0} specifies version {1} but "
+                "extension file {2} uses version {3}".format(
+                    main_file.filename, version, next_file.filename, next_file_version
+                )
+            )
+    return version
```
dnephin: I think we will probably need to do a similar version checking for extends, but I think we should wait until this is merged since it's already so large.

compose/config/config.py
dnephin: Ahh, right! Mind adding an inline comment about that? I'm sure I will forget.

shin-: Sure, will do.

compose/config/config.py
```diff
@@ -199,14 +238,52 @@ def load(config_details):
 
     Return a fully interpolated, extended and validated configuration.
     """
+    version = get_config_version(config_details)
+    if version not in COMPOSEFILE_VERSIONS:
+        raise ConfigurationError('Invalid config version provided: {0}'.format(version))
```
dnephin: By checking `COMPOSEFILE_VERSIONS` here we now have the source of truth for "what versions do we support" in two places, below in the if/elif, and in the const.

What do you think about removing this check and moving the error to an `else` in the branch below ?

shin-: I was doing this before but I think that moving it to the else was too late for some reason. I'm gonna test it again.

dnephin: Cool, if for some reason this needs to happen after processing files , the block below that iterates over the config_details could be moved into a function and called from both branches.

moxiegirl:
@dmp42 V's just ramping up on Compose.  If there is a rush on this, let us know and we can carry in a later PR.

compose/config/config.py
```diff
@@ -232,20 +309,20 @@ def merge_services(base, override):
...
     return build_services(config_file)
 
 
-def process_config_file(config_file, service_name=None):
-    validate_top_level_object(config_file)
-    processed_config = interpolate_environment_variables(config_file.config)
-    validate_against_fields_schema(processed_config, config_file.filename)
+def process_config_file(config_file, version, service_name=None):
+    validate_top_level_service_objects(config_file, version)
+    processed_config = interpolate_environment_variables(config_file.config, version)
```
dnephin: I notice both of these functions have a very similar line added to them to lookup the service_dicts. 

It would be nice to abstract that away a little better, and remove the duplication.  Maybe either a function `get_service_dict(config_file, version)` called from here, and only pass in the service_dicts to each of these?

Or maybe a property or method on the `ConfigFile` class for returning them?

shin-: Which 2 functions are you talking about? `validate_top_level_service_objects` and `interpolate_environment_variables` ?

dnephin: I think this change will also let you remove the version check in the return of `interpolate_environment_variables`, putting the config back in the right place can be handled here where we're forced to be aware of differences between versions.

compose/project.py
dnephin: @erikh If you want to manage your volumes externally, you'll be able to point at any external volumes as you would normally. This prefixing is only for volumes created locally as part of the compose file.

compose/project.py
```diff
@@ -205,6 +215,27 @@ def kill(self, service_names=None, **options):
...
+    def initialize_volumes(self):
+        try:
+            for volume in self.volumes:
+                volume.create()
+        except NotFound:
+            raise ConfigurationError(
+                'Volume %s specifies nonexistent driver %s' % (volume.name, volume.driver)
+            )
+        except APIError as e:
+            if 'Choose a different volume name' in str(e):
```
dnephin: This is really unfortunate. The volumes API should really be returning a proper 4xx code for this. It's a shame we have to resort to string comparison for this.

compose/config/config.py
dnephin: Right, those two

compose/project.py
shin-: Yeah, unfortunately right now this is a generic 500 ISE.
```
APIError: 500 Server Error: Internal Server Error ("A volume name skystar already exists with the nfs driver. Choose a different volume name.")
``

File: compose/project.py (hunk 8 above)
Comment (shin-) [inline compose/project.py]: @dnephin Actually, I haven't implemented that yet - it's complex enough that I'd rather make it a different PR, if we believe it's required.

Comment (dnephin) [inline compose/project.py]: Oh ya, just looking at `service.py` it seems the user would actually have to specify a value with the prefix included, is that right?

Another PR sounds fine to me.  

By the time this is in a release, we should have some way to use externally created volumes. #2502 is the issue for it.

Comment (erikh) [inline compose/project.py]: Ah ok! I misunderstood. Thank you!

File: compose/project.py (hunk 21 above)
Comment (thaJeztah) [inline compose/project.py]: @dnephin @shin- have you opened an issue for that in docker/docker? Sounds like a good improvement (I think I've seen a discussion around this, so there *may* already be an issue for it)

**Comment by shin-** [comment]
@dnephin @aanand Do we need anything else to proceed with merging this?

File: compose/project.py (hunk 8 above)
Comment (shin-) [inline compose/project.py]: I now have a branch handling the behavior discussed above: https://github.com/shin-/compose/commit/f432df4c3ede60f1eb6fc3e121887ece8326b3a8 .

File: script/test-versions [hunk 22]
```diff
@@ -18,7 +18,7 @@ get_versions="docker run --rm
 if [ "$DOCKER_VERSIONS" == "" ]; then
   DOCKER_VERSIONS="$($get_versions default)"
 elif [ "$DOCKER_VERSIONS" == "all" ]; then
-  DOCKER_VERSIONS="$($get_versions recent -n 2)"
+  DOCKER_VERSIONS="$($get_versions recent -n 1)"
```
dnephin: Mind adding a TODO here to set this back to 2 after the first RC is released?

compose/config/config.py
dnephin: I still think this is an issue, but not enough to block this branch, so I guess we can discuss it after this is merged.

dnephin:
LGTM, I think we can address the remaining (documentation, and using existing volumes) after this is merged

shin-:
Squashed some commits + rebased.

PR #2427, opened 2015-11-19
Add note about required pip version.

### LINKED ISSUE #2425: Dependency for functools32 not installed by pip
Installing docker-compose in CentOS under a virtualenv fails with this:

```
  File "/opt/docker/odoo-grupoesoc/pyenv/bin/docker-compose", line 5, in <module>
    from pkg_resources import load_entry_point
  File "/opt/docker/odoo-grupoesoc/pyenv/lib/python2.7/site-packages/pkg_resources/__init__.py", line 3095, in <module>
    @_call_aside
  File "/opt/docker/odoo-grupoesoc/pyenv/lib/python2.7/site-packages/pkg_resources/__init__.py", line 3081, in _call_aside
    f(*args, **kwargs)
  File "/opt/docker/odoo-grupoesoc/pyenv/lib/python2.7/site-packages/pkg_resources/__init__.py", line 3108, in _initialize_master_working_set
    working_set = WorkingSet._build_master()
  File "/opt/docker/odoo-grupoesoc/pyenv/lib/python2.7/site-packages/pkg_resources/__init__.py", line 658, in _build_master
    ws.require(__requires__)
  File "/opt/docker/odoo-grupoesoc/pyenv/lib/python2.7/site-packages/pkg_resources/__init__.py", line 959, in require
    needed = self.resolve(parse_requirements(requirements))
  File "/opt/docker/odoo-grupoesoc/pyenv/lib/python2.7/site-packages/pkg_resources/__init__.py", line 846, in resolve
    raise DistributionNotFound(req, requirers)
pkg_resources.DistributionNotFound: The 'functools32' distribution was not found and is required by jsonschema
```

Steps to reproduce:

```
virtualenv-2.7 pyenv
source ./pyenv/bin/activate
pip install docker-compose
docker-compose pull
```

Workaround: `pip install functools32`

Related to https://github.com/docker/compose/issues/2222?

dnephin:
Fixes #2425

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -70,6 +70,7 @@ to get started.
 
     $ pip install docker-compose
 
+> **Note:** pip version 6.0 or greater is required
 
 ### Install as a container
 


docs/install.md
```diff
@@ -70,6 +70,7 @@ to get started.
 
     $ pip install docker-compose
 
+> **Note:** pip version 6.0 or greater is required
```
shin-: Kind of odd to use a blockquote marker to emphasize this line. It's probably fine without it (and more semantically sound)?

Yajo:
Why not add it to requirements.txt?

dnephin:
@Yajo pip is the tool that reads the `requirements.txt`, so there's no opportunity to install it before other dependencies.

docs/install.md
dnephin: I would agree with you that it's not semantically correct. However this is the convention we use all over in the docs (not just in compose, but all the docker projects).

I think the reason is that markdown is so limited, it doesn't provide any other way to create a note like this with the proper style (something like http://sphinx-doc.org/markup/para.html#directive-note).

Yet another reason I think markdown is incomplete for documentation purposes!

aanand:
LGTM

PR #2430, opened 2015-11-19
Build from url

### LINKED ISSUE #491: Unable to fig build a github repo / works with docker build

dnephin:
Rebased #1372 and added more tests and docs
Fixes #491

### DIFF
--- a/compose/config/config.py
+++ b/compose/config/config.py
@@ -564,7 +564,8 @@ def validate_paths(service_dict):
             (not os.path.exists(build_path) or not os.access(build_path, os.R_OK))
         ):
             raise ConfigurationError(
-                "build path %s either does not exist or is not accessible." % build_path)
+                "build path %s either does not exist, is not accessible, "
+                "or is not a valid URL." % build_path)
 
 
 def merge_path_mappings(base, override):
--- a/compose/config/config.py
+++ b/compose/config/config.py
@@ -76,6 +76,13 @@ ALLOWED_KEYS = DOCKER_CONFIG_KEYS + [
     'external_links',
 ]
 
+DOCKER_VALID_URL_PREFIXES = (
+    'http://',
+    'https://',
+    'git://',
+    'github.com/',
+    'git@',
+)
 
 SUPPORTED_FILENAMES = [
     'docker-compose.yml',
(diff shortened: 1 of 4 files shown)


tests/unit/config/config_test.py
```diff
@@ -1393,6 +1393,34 @@ def test_from_file(self):
...
+            }, '.', None))
+            assert service_dict[0]['build'] == valid_url
+
+    def test_invalid_url_in_build_path(self):
+        invalid_urls = [
+            'example.com/bogus',
+            'ftp://example.com/',
+            '/path/does/not/exist',
+        ]
+        for valid_url in invalid_urls:
```
aanand: This variable should probably be called something different

compose/config/config.py
```diff
@@ -524,11 +531,26 @@ def resolve_volume_path(working_dir, volume):
...
         build_path = service_dict['build']
-        if not os.path.exists(build_path) or not os.access(build_path, os.R_OK):
-            raise ConfigurationError("build path %s either does not exist or is not accessible." % build_path)
+        if (
+            not is_url(build_path) and
+            (not os.path.exists(build_path) or not os.access(build_path, os.R_OK))
+        ):
+            raise ConfigurationError(
+                "build path %s either does not exist, is not accessible, "
+                "or was not recognized as a url." % build_path)
```
aanand: Extremely minor thing: perhaps "...or is not a valid URL", to make it clear that the *input* is the problem, rather than that some internal operation failed.

aanand:
Great stuff.

dnephin:
Updated with those changes

aanand:
LGTM

PR #2436, opened 2015-11-20
Reduce cyclomatic complexity

dnephin:
In an attempt to bring cyclomatic complexity down to the suggested value of 10 I've refactored the two most complex functions, which gets us pretty close (12). 

Also makes `pause` and `unpause` operations happen in parallel.

### DIFF
--- a/compose/container.py
+++ b/compose/container.py
@@ -1,6 +1,7 @@
 from __future__ import absolute_import
 from __future__ import unicode_literals
 
+import operator
 from functools import reduce
 
 import six
@@ -8,6 +9,7 @@ import six
 from .const import LABEL_CONTAINER_NUMBER
 from .const import LABEL_PROJECT
 from .const import LABEL_SERVICE
+from compose.utils import parallel_execute
 
 
 class Container(object):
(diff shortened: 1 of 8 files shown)


dnephin:
Split view (https://github.com/docker/compose/pull/2436/files?diff=split) makes this a lot easier to read

compose/container.py
```diff
@@ -250,3 +252,40 @@ def get_container_name(container):
...
+def parallel_unpause(containers, options):
+    parallel_operation(containers, 'unpause', 'Unpausing')
+
+
+def parallel_kill(containers, options):
+    parallel_operation(containers, 'kill', 'Killing')
+
+
+def parallel_restart(containers, options):
+    parallel_operation(containers, 'restart', 'Restarting')
```
dnephin: I wasn't sure where to put these.

What do you think about a new module `parallel.py`, and moving all of these plus the `parlallel_execute()` from utils.py into the new module?

compose/utils.py
```diff
@@ -17,58 +17,50 @@
...
 
 
-def parallel_execute(objects, obj_callable, msg_index, msg):
-    """
-    For a given list of objects, call the callable passing in the first
+def perform_operation(func, arg, callback, index):
+    try:
+        callback((index, func(arg)))
+    except Exception as e:
+        callback((index, e))
```
dnephin: This used to be the nested function `inner_execute_function()`

compose/utils.py
```diff
@@ -17,58 +17,50 @@
...
-            pass
+            continue
+
+        if isinstance(result, APIError):
+            errors[msg_index] = "error", result
+            writer.write(msg_index, 'error')
+        elif isinstance(result, Exception):
+            errors[msg_index] = "unexpected_exception", result
+        else:
+            writer.write(msg_index, 'done')
```
dnephin: I noticed we had effectively the same branch conditions in both the nested function and the "collect" phase. Instead of having both, we can just check the result type for errors.

compose/utils.py
```diff
@@ -80,6 +72,36 @@ def inner_execute_function(an_callable, parameter, msg_index):
...
+        position = self.lines.index(obj_index)
+        diff = len(self.lines) - position
+        # move up
+        self.stream.write("%c[%dA" % (27, diff))
+        # erase
+        self.stream.write("%c[2K\r" % 27)
+        self.stream.write("{} {} ... {}\r".format(self.msg, obj_index, status))
+        # move back down
+        self.stream.write("%c[%dB" % (27, diff))
+        self.stream.flush()
```
dnephin: Making `write_out_msg()` a class allows us to:
* remove the need for the caller to be aware of `lines`, the state is managed within this class
* remove the need to pass around `stream` and the default message. It's always the same, so we can pass it in as a dependency to the constructor
* drop the if/else condition, they are now separate methods

compose/container.py
```diff
@@ -250,3 +252,40 @@ def get_container_name(container):
...
+def parallel_unpause(containers, options):
+    parallel_operation(containers, 'unpause', options, 'Unpausing')
+
+
+def parallel_kill(containers, options):
+    parallel_operation(containers, 'kill', options, 'Killing')
+
+
+def parallel_restart(containers, options):
+    parallel_operation(containers, 'restart', options, 'Restarting')
```
dnephin: I wasn't sure where to put these. I put them into `container.py` since they are effectively operations on containers.

What do you think about a new module `parallel.py`, and moving all of these plus the `parlallel_execute()` from utils.py into the new module?

aanand: Sounds good, yep.

aanand:
I like it!

dnephin:
Cool, parallel_execute and the other methods moved to a new module

aanand:
LGTM

PR #2438, opened 2015-11-20
Fix `build --pull` test flakes

dnephin:
These tests have flaked at least a couple times today. This should prevent it.

The test checks for the "already up to date" text.

Example failure: https://jenkins.dockerproject.org/job/Compose-PRs/1860/console


aanand:
LGTM

PR #2450, opened 2015-11-23
Fix env_file overrides with extends

### LINKED ISSUE #2446: env_file doesn't get extended
https://github.com/vipconsult/dockerfiles

docker-compose -f always-on-prod.yml --x-networking up -d

this extends always-on.yml which on it side extends common.yml which has 
env_file:
        - env 

always-on-prod.yml defines its own env_file
env_file:
        - env-prod

the problem is that envs defined in the env-prod don't override the envs defined in common.yml 

I think this worked as expected before I upgraded from  1.5.0

Docker version 1.9.0, build 76d6bc9
docker-compose version: 1.6.0dev

linux debian jessie

dnephin:
Based on #2405
Fixes #2446

### DIFF
--- a/compose/config/config.py
+++ b/compose/config/config.py
@@ -324,16 +324,13 @@ class ServiceExtendsResolver(object):
         return filename
 
 
-def resolve_environment(service_config):
+def resolve_environment(service_dict):
     """Unpack any environment variables from an env_file, if set.
     Interpolate environment values if set.
     """
-    service_dict = service_config.config
-
     env = {}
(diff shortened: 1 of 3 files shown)


aanand:
Should this be rebased now that #2405 is in?

dnephin:
I can rebase it to update github, but it won't change the actual merge

dnephin:
Ok, rebased

aanand:
Great. Makes the PR easier to read at least!

LGTM

PR #2458, opened 2015-11-25
Image and build together

### LINKED ISSUE #2092: Specify both image name and build path

dnephin:
Fixes #2092 

Allow both `image` and `build` in a service config, and use `image` as the image name when building, if it exists.

### DIFF
--- a/compose/config/config.py
+++ b/compose/config/config.py
@@ -305,7 +305,8 @@ def load_services(working_dir, config_files, version):
         return {
             name: merge_service_dicts_from_files(
                 base.get(name, {}),
-                override.get(name, {}))
+                override.get(name, {}),
+                version)
             for name in all_service_names
         }
 
@@ -397,7 +398,10 @@ class ServiceExtendsResolver(object):
             service_name,
         )
 
-        return merge_service_dicts(other_service_dict, self.service_config.config)
+        return merge_service_dicts(
+            other_service_dict,
+            self.service_config.config,
+            self.version)
 
     def get_extended_config_path(self, extends_options):
         """Service we are extending either has a value for 'file' set, which we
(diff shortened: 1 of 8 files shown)


dnephin:
Rebased, now only applies to V2 config.

aanand:
LGTM

graingert:
@dnephin does this allow something like:

```yml
# docker-compose.yml
production:
   - image: ham/spam
   - build: ./spam
development:
   - image: ham/spam-dev
   - build: ./spam
   - dockerfile: Dockerfile-dev
```

```
# spam/Dockerfile
FROM ubuntu
RUN apt-get install nginx npm
COPY . /usr/bin/src/
RUN npm build
EXPOSE 80 443
CMD ["nginx", "-g", "daemon off;"]
```

```
# spam/Dockerfile-dev
FROM ham/spam # build on parent ham/spam.
RUN npm start # start live reloader
```

aanand:
@graingert Yes, but you have to use the v2 file format to use `image` and `build` together:

```yaml
version: 2
services:
  production:
    image: ham/spam
    build: ./spam
  development:
    image: ham/spam-dev
    build:
      context: ./spam
      dockerfile: Dockerfile-dev
```

You probably wouldn't actually define development and production as separate services in the same file, though. More likely you'd put the core config in `docker-compose.yml` and the development override in `docker-compose.override.yml`:

```yaml
# docker-compose.yml
version: 2
services:
  spam:
     image: ham/spam
     build: ./spam

# docker-compose.override.yml
version: 2
services:
  spam:
    image: ham/spam-dev
    build:
      context: ./spam
      dockerfile: Dockerfile-dev
```

graingert:
@aanand How does docker-compose know it needs to build "ham/spam" for `docker-compose.override.yml` to work?

aanand:
I see what you were getting at now. Compose doesn't support dependent builds - you should use a tool like [Shipwright](https://github.com/6si/shipwright) for that.

graingert:
@aanand ah nice, thanks

dnephin:
I've been prototyping a tool in https://github.com/dnephin/dobi that would support building chains of images (as well as a bunch of other things). There are a few [docs](https://github.com/dnephin/dobi/tree/master/docs) and [examples](https://github.com/dnephin/dobi/tree/master/examples), but it could definitely use more, I need one for the ordered build use case.

PR #2464, opened 2015-11-26
Validate `expose`

### LINKED ISSUE #2406: Validate the format of `expose`
The other day I expend some hours trying to investigate why my docker container was not exposing some ports. And understanding that was a docker problem, and not an app problem.

In dockerfile there is **expose** to say the ports, but **ports** in docker-compose. Ok, there is a change.

But why cannot say that "warning, expose syntax is not valid in docker-compose, you meant ports?" or something like this?

aanand:
I made a bunch of fixes and refactorings to `ports` validation before implementing `expose` validation - see the commit log.

Closes #2406.

### DIFF
--- a/compose/config/fields_schema.json
+++ b/compose/config/fields_schema.json
@@ -83,16 +83,8 @@
         "ports": {
           "type": "array",
           "items": {
-            "oneOf": [
-              {
-                "type": "string",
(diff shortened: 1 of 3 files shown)


dnephin:
LGTM

PR #2466, opened 2015-11-27
Removes the deprecated --allow-insecure-ssl flag

### LINKED ISSUE #2368: Remove `--allow-insecure-ssl` flag
This flag became a no-op warning in 1.4. So there have been two full releases which warned about this deprecated flag.

I think we can drop it as part of the 1.6 release. What do you think?

dnephin:
Fixes #2368


aanand:
Tests failed on `CLITestCase.test_up_handles_sigterm` - looks unrelated?

dnephin:
Yup, pretty sure this is a test flake because the 5 second timeout is too aggressive. I'll put up another PR to increase the timeout.

aanand:
LGTM

PR #2467, opened 2015-11-27
Remove migrate-to-labels.

### LINKED ISSUE #2367: Remove `migrate-to-labels`
Migrate to labels was added in 1.3 to upgrade from 1.2.

I think we can probably drop it for the 1.6 release.  What do you think?

If someone still needs to migrate, they can always use 1.5.

dnephin:
Fixes #2367

Remove legacy support and migrate-to-labels


aanand:
Lovely.

If we don't check for legacy containers any more, what will happen if there *are* legacy containers? Will it result in a name conflict when we try and create new ones?

dnephin:
I think that's right, we'll get an error about container name already exists

aanand:
LGTM

PR #2491, opened 2015-12-03
WIP: Bump 1.5.2

dnephin:


aanand:
Nice. Any idea what's going on with the test failures? I guess the `env_file` fixes depend on a big chain of refactoring commits, so it's a bit of a tangle. Maybe something got left out?

dnephin:
I figured out the test failure. There was a refactoring done between two sets of changes which added `env_file` to the list of fields to merge in `merge_services`. I missed that change when resolving a merge conflict.  I've added it back in, and it looks like test are passing now!

dnephin:
Fixed the wording in the changelog as well. Once janky runs green I think this is ready to go

PR #2492, opened 2015-12-03
Stop warning about ".yaml" file extension

### LINKED ISSUE #2330: Stop warning about ".yaml" file extension.
Currently compose outputs a warning when using extensions other than `.yml` for the config file:

```
WARNING: Please be aware that .yml is the expected extension in most cases, and using .yaml can cause compatibility issues in future.
```

However, the creators of YAML request that we use `.yaml` when possible. [[1]](http://www.yaml.org/faq.html)

Additionally, I don't think it's fruitful to warn users about this particular "compatibility issue"; things like which character is used for `\n` in text files or using uppercase or non-ASCII characters in file names can also be "compatibility issues" when multiple platforms are involved, yet most programs don't force a user of a particular platform to use a convention which is compatible with other platforms.

Just my two cents. If the devs agree that this can/should be removed, I'll make a PR.

jake-low:
Fixes #2330.

### DIFF
--- a/compose/config/config.py
+++ b/compose/config/config.py
@@ -161,11 +161,6 @@ def get_default_config_files(base_dir):
         log.warn("Found multiple config files with supported names: %s", ", ".join(candidates))
         log.warn("Using %s\n", winner)
 
-    if winner == 'docker-compose.yaml':
-        log.warn("Please be aware that .yml is the expected extension "
-                 "in most cases, and using .yaml can cause compatibility "
-                 "issues in future.\n")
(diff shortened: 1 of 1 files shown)


dnephin:
As long as we have a warning when multiple candidates exist, I think it's probably ok that we accept different files as the default. Especially considering that `.yaml` is the suggested extension.

aanand:
Agreed.

LGTM

dnephin:
LGTM

PR #2494, opened 2015-12-03
Limit log memory usage for huge logs

romuald:
When using the log command logs with lots of logs, dockek-compose will accumulate logs in memory faster than they can be streamed on sys.stdout

Add an optional maximum queue size in Multiplexer to make incommingstreams block

With this patch, I've reduced the memory usage of `docker-compose logs` for my current containers from 250Mb to ~ 40Mb

### DIFF
--- a/compose/cli/log_printer.py
+++ b/compose/cli/log_printer.py
@@ -24,7 +24,7 @@ class LogPrinter(object):
 
         prefix_width = max_name_width(self.containers)
         generators = list(self._make_log_generators(self.monochrome, prefix_width))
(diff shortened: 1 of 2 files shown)


dnephin:
LGTM

dnephin:
Thanks for the Fix!

Needs a rebase, and the value 10 should probably be a constant.

dnephin:
Closing since no response from the OP, and this branch is going to require a lot of rework to merge, now that the log printer code has changed significantly.

PR #2512, opened 2015-12-07
Use more robust download URL for completions

albers:
The download URLs for the bash and zsh completions are currently computed by parsing the output of `docker-compose --version`. This is ugly and error-prone, see #2417 for an example (though compatibility was preserved in that case).

With `docker-compose version --short` in place, we now have a more robust way of getting the version string.


thaJeztah:
Much cleaner as well, nice!

aanand:
LGTM

dnephin:
Nicer for release versions, but does break the command for master builds. I think that's fine.

LGTM (just waiting on janky)

PR #2518, opened 2015-12-08
Add bash completion for config subcommand

albers:
This adds bash completion for the new `docker-compose config` subcommand, ref #2407

/cc @sdurrheimer new command, please add to zsh completion.


dnephin:
LGTM, Thanks!

aanand:
LGTM

PR #2521, opened 2015-12-09
Add method to get more info about container status

jeanpralo:
Signed-off-by: Jean Praloran <jeanpralo@gmail.com>

### DIFF
--- a/compose/container.py
+++ b/compose/container.py
@@ -115,6 +115,8 @@ class Container(object):
     def human_readable_state(self):
         if self.is_paused:
             return 'Paused'
+        if self.is_restarting:
+            return 'Restarting'
         if self.is_running:
             return 'Ghost' if self.get('State.Ghost') else 'Up'
         else:
@@ -133,6 +135,10 @@ class Container(object):
     @property
     def is_running(self):
         return self.get('State.Running')
+
+    @property
+    def is_restarting(self):
+        return self.get('State.Restarting')
 
     @property
     def is_paused(self):


compose/container.py
```diff
@@ -131,10 +133,18 @@ def environment(self):
         return dict(var.split("=", 1) for var in self.get('Config.Env') or [])
 
     @property
+    def status(self):
+        return self.get('State.Status')
```
aanand: Is this actually used anywhere? I don't think there's much point adding it if not.

aanand:
I like the change to `human_readable_state` so that we can see when a container's restarting. Just one comment inline.

compose/container.py
jeanpralo: well I am not using it directly inside docker-compose, I figured this could be helpfull for people like me who import compose directly into their python code. 

So I don't really mind if it does not make it through.

jeanpralo:
Yeah will really like to see this merged since right now a container that keeps on restarting if it has a restart flag set to always will say it's up.

compose/container.py
aanand: OK - that makes sense, but as we're not prioritising that use case right now, I'd rather not add it.

jeanpralo:
@dnephin i have removed the status method :)

dnephin:
LGTM

aanand:
Needs squashing to one commit, then LGTM

jeanpralo:
@aanand done :)

Issue #1135, opened 2015-03-19
Cannot open self /usr/bin/docker-compose or archive /usr/bin/docker-compose.pkg


kokaz:
I installed `docker-compose` on my server I reconnect later it said :

``` sh
Cannot open self /usr/bin/docker-compose or archive /usr/bin/docker-compose.pkg
```

I have to reinstall it all the time :/

Thanks for your help!

jasherai:
:+1: I'm experiencing this too... not sure what extra info would be useful to help debug

funkyfuture:
how have you installed it? what operating system are you using? what's the output of `which docker-compose`?

jasherai:
I've installed via the curl + chmod technique and I'm on ubuntu 14.10 (utopic). And output as requested...
```
pmehta@phaze:~$ which docker-compose
/usr/local/bin/docker-compose
```

kokaz:
I've installed `docker-compose` with `curl` and `chmod` method too on a CentOS 7.
```
$ which docker-compose                                                                                                                                              [19:24:40]
/usr/local/bin/docker-compose
```

funkyfuture:
does it work when you install compose with pip?

hajaalin:
I had same problem, it was because of permissions. This helped:
`sudo chmod a+rx /usr/local/bin/docker-compose`

funkyfuture:
@hajaalin that seems odd. which OS is that? how did you install it?

@jasherai we never had problem with an installation via `pip` on Ubuntu. and also `python setup.py develop`

hajaalin:
@funkyfuture I installed with `curl` and `chmod` method. This is Ubuntu 14.04.2, but not fully standard, the distro is maintained by our university. I don't know if there is some tweak that affects default file permissions, for example.

jasherai:
@funkyfuture I've been running the pip version of the install since your suggestion and it has been stable since. This seems specific to the curl method. I haven't tried the altered permissions that @hajaalin suggested so can't give feedback on that.

djsly:
Same problem here

```
-bash-4.1$ docker-compose -p docker ps
Couldn't connect to Docker daemon at http+unix://var/run/docker.sock - is it running?

If it's at a non-standard location, specify the URL with the DOCKER_HOST environment variable.
-bash-4.1$ sudo docker-compose -p docker ps
Cannot open self /usr/bin/docker-compose or archive /usr/bin/docker-compose.pkg
[...]

PR #2523, opened 2015-12-09
Docs don't mention docker-compose also requires read permission.

XavierTalpe:
When installing docker-compose on Ubuntu I ran into the following problem:
````
Cannot open self /usr/local/bin/docker-compose or archive /usr/local/bin/docker-compose.pkg
````

Based on some Googling and information in https://github.com/docker/compose/issues/1135, adding the read permission solved the issue for me and other people.

### DIFF
--- a/contrib/completion/bash/docker-compose
+++ b/contrib/completion/bash/docker-compose
@@ -102,6 +102,11 @@ _docker_compose_build() {
 }
 
 
+_docker_compose_config() {
+	COMPREPLY=( $( compgen -W "--help --quiet -q --services" -- "$cur" ) )
+}
+
+
 _docker_compose_docker_compose() {
 	case "$prev" in
 		--file|-f)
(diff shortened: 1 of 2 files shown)


aanand:
Looks like this needs a rebase.

dnephin:
How did you download the binary? It's very strange that it would be downloaded without read permissions.  Are you sure the issue wasn't the user ownership ?

XavierTalpe:
@dnephin Good point. I redid the steps I did yesterday and now it seems to work... /confused. The read flag is indeed set by default. Not sure how I managed to run into this issue. I guess this PR can be closed as invalid.

aanand:
I suppose it wouldn't hurt to add `+rx` though?

dnephin:
Since it's not necessary, I think it might just serve to confuse people instead of actually fixing anything.

PR #2526, opened 2015-12-10
Ignore extra coverage files

dnephin:
Ignore extra coverage files that are created because we run acceptance tests in a subprocess.

These files have the process id in their name, so they wont be removed by the normal coverage cleanup on each run.

Now doing the cleanup as part of the normal `./script/clean`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,6 @@
 *.egg-info
 *.pyc
-/.coverage
+.coverage*
 /.tox
 /build
 /coverage-html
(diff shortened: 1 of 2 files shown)


aanand:
Seems to make Windows unhappy :/

dnephin:
Ya, using find isn't going to work.  I've changed this PR to instead just `.gitignore` the files, and have `./script/clean` remove them

aanand:
LGTM

PR #2527, opened 2015-12-10
Fix dns and dns_search when used strings and without extends.

### LINKED ISSUE #2515: ERROR: json: cannot unmarshal string into Go value of type []string
Software versions:
```bash
$ docker-compose -v
docker-compose version 1.5.2, build 7240ff3
$ docker -v
Docker version 1.9.1, build a34a1d5
```

compose.yml:

```yml
mycont:
  #tty: true
  #entrypoint: /sbin/init
  env_file: ".env"
  build: "."
  container_name: "mycont"
  domainname: "example.lab"
  hostname: "mycont"
  dns: "192.168.128.3"
  dns_search: "example.lab"
  restart: "always"
  expose:
    - "389"
    - "80"
  extra_hosts:
    - "ldap.example.lab:192.168.1.10"
    - "ldap:192.168.1.10"
    - "tgnpdc:192.168.1.10"
  # http://blog.oddbit.com/2014/08/11/four-ways-to-connect-a-docker/
  net: "bridge"
  ports:
    - "389:389"
    - "80:80"
```

Error:

```bash
$ docker-compose up -d
Recreating 177deb9c24_177deb9c24_177deb9c24_177deb9c24_177deb9c24_177deb9c24_177deb9c24_177deb9c24_177deb9c24_177deb9c24_177deb9c24_177deb9c24_177deb9c24_177deb9c24_177deb9c24_mycont
ERROR: json: cannot unmarshal string into Go value of type []string
```

I suspect that this is because latest updates installed.

dnephin:
Fixes #2515

### DIFF
--- a/tests/unit/config/config_test.py
+++ b/tests/unit/config/config_test.py
@@ -535,6 +535,23 @@ class ConfigTest(unittest.TestCase):
             }))
         assert "which is an invalid type" in exc.exconly()
 
+    def test_normalize_dns_options(self):
+        actual = config.load(build_config_details({
+            'web': {
+                'image': 'alpine',
+                'dns': '8.8.8.8',
+                'dns_search': 'domain.local',
+            }
+        }))
+        assert actual == [
+            {
+                'name': 'web',
+                'image': 'alpine',
+                'dns': ['8.8.8.8'],
+                'dns_search': ['domain.local'],
+            }
+        ]
+
 
 class PortsTest(unittest.TestCase):
     INVALID_PORTS_TYPES = [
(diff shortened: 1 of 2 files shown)


tests/unit/config/config_test.py
```diff
@@ -1080,8 +1097,8 @@ def test_resolve_environment_nonexistent_file(self):
                 {'foo': {'image': 'example', 'env_file': 'nonexistent.env'}},
                 working_dir='tests/fixtures/env'))
 
-            assert 'Couldn\'t find env file' in exc.exconly()
-            assert 'nonexistent.env' in exc.exconly()
+        assert 'Couldn\'t find env file' in exc.exconly()
+        assert 'nonexistent.env' in exc.exconly()
```
dnephin: These assertions weren't running because they were in the `with` block. I noticed this while looking around to make sure we had a test that covered the case where `env_file` was a string (this test covers it).

aanand:
LGTM

Issue #2163, opened 2015-10-09
Proposal: new build config section


dnephin:
#### Motivation for this proposal
* https://github.com/docker/docker/pull/15182 will be in 1.9.0 allow for build args
* new users are often confused when the `environment` section isn't applied during the build phase

#### Proposal

Move all build related configuration into a `build:` section in the service.  Example

```yaml

web:
  build:
    context: .
    dockerfile: Dockerfile.name
    args:
       key: value
  ...
```

This should help clarify the fields that are related to the build phase (and the fields that are not).

#### Backwards compatibility

On the first release both `build` and `dockerfile` would still be allowed at the top level of the service definition.  A `dockerfile` at the top level would print a deprecation warning.

On the following release `dockerfile` at the top level of the service definition would be removed, and it must be under the `build` section.  For trivial cases, a config of `build: path/to/context` would still be supported.

Related #2111

thaJeztah:
I like this approach, sgtm. For the deprecation of the top-level fields; perhaps Compose should add a deprecation policy similar to Engine; https://github.com/docker/docker/blob/master/docs/misc/deprecated.md. This gives users time to adapt, and a clear notice when a feature will be removed,

blaggacao:
+1 can't wait... :dancer:

mosop:
:+1:

marcellodesales:
:+1:

lechen26:
is this option was added to docker-compose? the support for build-args?

Silex:
:+1:

ducva:
:+1:

GarrettHeel:
Working on this in #2540

levivm:
+1

dnephin:
Merged in #2653, carried from #2540

marcellodesales:
Thanks everyone! :+1:  Can't wait for this release!

ericm24:
+1 Really need --build-arg capability with docker-compose!!  Will be MOST GLAD when it is added and released!

marcellodesales:
@ericm24 +1 :) Can't wait too! :)

Issue #2111, opened 2015-09-30
Add support for --build-args


kismet010:
As this docker/docker#15182 functionality has been merged, it would be great to manage with Compose

I.e.:

Dockerfile with default variables
```
ARG vhost=default.com
RUN echo $vhost > ...
```

Overriding through docker-composer.yml
```
web:
  build: ...
  arguments:
    - vhost: dev.com
```

Or an hierarchy of extended environments
```
production:
  extends: web
  arguments:
    - vhost: my-final-domain.com
```

dnephin:
I've been wanting to make a change to the build config, something like this:

```yaml

web:
  build:
    context: .
    dockerfile: dockerfiles/path/Dockerfile
    arguments:
      arg1: ...
```

dnephin:
I created a proposal for this configuration change at #2163

smboykin:
+1 on this feature request

snelis:
+1 on passing environment variables onto the build step via compose

lsborg:
+1

Silex:
+1

marcellodesales:
I'm stuck at Compose right now... :( Please please add this support...

I can sucessfully build with the the --build-arg because I need the HTTP_PROXY on my build servers... :(

```sh
[root@pppdc9prda2x npmo-docker-compose]# docker build --build-arg HTTP_PROXY=$HTTP_PROXY -t configure -f ./configure.Dockerfile .
Sending build context to Docker daemon 141.8 kB
Step 1 : FROM node:4-onbuild
# Executing 3 build triggers...
Step 1 : COPY package.json /usr/src/app/
 ---> Using cache
Step 1 : RUN npm install
 ---> Using cache
Step 1 : COPY . /usr/src/app
 ---> 52f129b79095
Removing intermediate container 00cb3b4651d4
Step 2 : CMD npm run configure
 ---> Running in 7d0642695c58
 ---> db140f15fe39
Removing intermediate container 7d0642695c58
Successfully built db140f15fe39
```

Build from compose
----

I still get a build stuck trying to connect to resources... I would propose:

```
  build_args:
    - HTTP_PROXY=$HTTP_PROXY
```

```yml
licenseData:
  image: busybox
  volumes:
    - .license.json:/usr/src/app/.license.json

configure:
  build: .
  dockerfile: configure.Dockerfile
  build_args:
    - HTTP_PROXY=$HTTP_PROXY
  volumes_from:
    - "licenseData"
  environment:
[...]

PR #2540, opened 2015-12-12
Add support for build arguments

GarrettHeel:
Fixes # #2163

### DIFF
--- a/compose/config/config.py
+++ b/compose/config/config.py
@@ -2,6 +2,7 @@ from __future__ import absolute_import
 from __future__ import unicode_literals
 
 import codecs
+import functools
 import logging
 import operator
 import os
@@ -437,6 +438,11 @@ def resolve_environment(service_dict):
     return dict(resolve_env_var(k, v) for k, v in six.iteritems(env))
 
 
+def resolve_build_args(build):
+    args = parse_build_arguments(build.get('args'))
+    return dict(resolve_env_var(k, v) for k, v in six.iteritems(args))
+
+
 def validate_extended_service_dict(service_dict, filename, service):
     error_prefix = "Cannot extend service '%s' in %s:" % (service, filename)
 
@@ -474,11 +480,15 @@ def process_service(service_config):
             for path in to_list(service_dict['env_file'])
         ]
 
+    if 'build' in service_dict:
+        if isinstance(service_dict['build'], six.string_types):
+            service_dict['build'] = resolve_build_path(working_dir, service_dict['build'])
+        elif isinstance(service_dict['build'], dict) and 'context' in service_dict['build']:
+            path = service_dict['build']['context']
+            service_dict['build']['context'] = resolve_build_path(working_dir, path)
+
     if 'volumes' in service_dict and service_dict.get('volume_driver') is None:
         service_dict['volumes'] = resolve_volume_paths(working_dir, service_dict)
-
-    if 'build' in service_dict:
-        service_dict['build'] = resolve_build_path(working_dir, service_dict['build'])
 
     if 'labels' in service_dict:
         service_dict['labels'] = parse_labels(service_dict['labels'])
@@ -514,6 +524,8 @@ def finalize_service(service_config):
 
     if 'restart' in service_dict:
         service_dict['restart'] = parse_restart_spec(service_dict['restart'])
+
+    normalize_build(service_dict, service_config.working_dir)
 
     return normalize_v1_service_format(service_dict)
 
@@ -579,8 +591,29 @@ def merge_service_dicts(base, override, version):
 
     if version == 1:
         legacy_v1_merge_image_or_build(d, base, override)
+    else:
+        merge_build(d, base, override)
 
     return d
+
+
+def merge_build(output, base, override):
+    build = {}
+
+    if 'build' in base:
+        if isinstance(base['build'], six.string_types):
+            build['context'] = base['build']
+        else:
+            build.update(base['build'])
+
+    if 'build' in override:
+        if isinstance(override['build'], six.string_types):
+            build['context'] = override['build']
+        else:
+            build.update(override['build'])
+
+    if build:
+        output['build'] = build
 
 
 def legacy_v1_merge_image_or_build(output, base, override):
@@ -602,22 +635,6 @@ def merge_environment(base, override):
     return env
 
 
-def parse_environment(environment):
-    if not environment:
-        return {}
-
-    if isinstance(environment, list):
-        return dict(split_env(e) for e in environment)
-
-    if isinstance(environment, dict):
-        return dict(environment)
-
-    raise ConfigurationError(
-        "environment \"%s\" must be a list or mapping," %
-        environment
-    )
-
-
 def split_env(env):
     if isinstance(env, six.binary_type):
         env = env.decode('utf-8', 'replace')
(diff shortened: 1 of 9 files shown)


dnephin:
Thanks for the PR!  Please see the proposed format in #2163 for how we'd like to implement this,

GarrettHeel:
@dnephin  Ah - I see, that makes sense. Is this something that someone @ dockerhq is implementing or should I tackle that issue as a whole?

dnephin:
@GarrettHeel no one is working on it yet, so if you'd like to pick it up, that would be great

Let me know if you have any questions, either here or on irc

GarrettHeel:
@dnephin I had a go at implementing this, could you take a look?

aanand:
Looks like a great start. I think it'd be better to only support `build` as an object in the new version of the file format, however.

We should probably extract the config versioning stuff out of #2421 and get it into master so it's not blocking PRs like this one.

compose/config/config.py
```diff
@@ -25,6 +25,11 @@
 from .validation import validate_extends_file_path
 from .validation import validate_top_level_object
 
+# basestring not available or needed in python3
+try:
+    basestring
+except NameError:
+    basestring = str
```
dnephin: Please use `six.string_types` instead of this try/except

docs/compose-file.md
```diff
@@ -41,9 +55,47 @@ Compose will build and tag it with a generated name, and use that image thereaft
...
+
+
+#### args
+
+Add build arguments. You can use either an array or a dictionary. Any
+boolean values; true, false, yes, no, need to be enclosed in quotes to ensure
+they are not converted to True or False by the YML parser.
+
+Build arguments with only a key are resolved to their environment value on the
+machine Compose is running on, which can be helpful for secret or host-specific values.
```
dnephin: The [dockerfile docs](https://docs.docker.com/engine/reference/builder/#arg) say not to use this for secrets. The reasons aren't really clear (maybe the values are stored in the image after?), but I think it's best we don't contradict the documentation.

tests/integration/testcases.py
```diff
@@ -38,11 +38,15 @@ def create_service(self, name, **kwargs):
         if 'command' not in kwargs:
             kwargs['command'] = ["top"]
 
+        client = self.client
+        if 'client' in kwargs:
+            client = kwargs.pop('client')
```
dnephin: We're going to be upgrading the minimum API version to 1.21 pretty soon, so we'll be able to drop this.  Tests will fail for now against the older version.  

Once #2421 is merged, we'll only test against 1.21, and I think that branch is a pre-req for these changes so that we can apply this to only the v2 config.

compose/config/config.py
```diff
@@ -493,6 +517,22 @@ def parse_environment(environment):
...
+    if isinstance(arguments, list):
+        return dict(split_env(e) for e in arguments)
+
+    if isinstance(arguments, dict):
+        return dict(arguments)
+
+    raise ConfigurationError(
+        "build arguments \"%s\" must be a list or mapping," %
+        arguments
+    )
```
dnephin: Since this is almost identical to `parse_environment()` (and really `parse_labels()` is as well), I think we should have just one function that takes two params for the differences between them:

```python

def parse_dict_or_list(split_func, type_name, arguments):
    ...

parse_build_arguments = functools.partial(parse_dict_or_list, split_env, 'build arguments')
parse_environment = functools.partial(parse_dict_or_list, split_env, 'build arguments')
parse_labels = functools.partial(parse_dict_or_list, split_label, 'labels')
```

These could be single line functions instead of using `functools.partial`, I'm *partial* to functools.

dnephin:
Thanks! this is looking great. I think we'll need a couple days to get  #2421 merged so that we have version support and we can add this to the v2 config.

GarrettHeel:
Fixed all of your comments. I'll keep an eye on #2421 and rebase when it's merged

GarrettHeel:
@dnephin @aanand did we still want to support the `build: .` shortcut for version 2 or always expect a dict?

dnephin:
I like the idea of supporting `build: .` for the case when you don't have any other build related settings. As soon as you need args, or dockerfile, you need to use the dict.

BlkChockr:
:+1: @dnephin, but that one has to be well documented, I think. Otherwise, people will probably get a headache trying to use both `build: .` and the dict for the settings at the same time ;)

Also: In which version and when will this (presumably) be released? I really need this! :)

dnephin:
I've added the version milestone (1.6)

marcellodesales:
@dnephin Is there anything people can help to get this built? I've been waiting on this to see #2111... I'd like to build from source and test...

dnephin:
We've made some rather large changes to support the v2 format, and I think this branch and #2458 will cause a few merge conflicts with each other.

Either way, I think you can expect to see something in master by the end of the week.

GarrettHeel:
I've rebased this off #2421 and #2458 now.

The documentation could use some work, especially between versions. I.e `dockerfile` being moved into the the `build` object in version 2.

Let me know what you think.

marcellodesales:
@dnephin I will have some time tomorrow to build and run this local... I really need this feature! :+1: Thanks!

dnephin:
Needs another rebase.

@GarrettHeel just want to let you know that we've got code freeze coming up soon, and we want to get this in. We may need to carry this PR and make some small changes to get it merged. #2653 is rebased on the latest master with a few small changes.

I'll leave a few comments here about the docs changes, but we might just end up pulling them into #2653 if we run out of time.

compose/config/config.py
```diff
@@ -430,6 +431,12 @@ def resolve_environment(service_dict):
     return dict(resolve_env_var(k, v) for k, v in six.iteritems(env))
 
 
+def resolve_build_args(build):
+    args = {}
+    args.update(parse_build_arguments(build.get('args')))
```
dnephin: I think since `parse_build_arguments()` always returns a new dict, these two lines could be:

```python
args = parse_build_arguments(build.get('args'))
```

compose/config/config.py
```diff
@@ -668,7 +734,13 @@ def is_url(build_path):
 
 def validate_paths(service_dict):
     if 'build' in service_dict:
-        build_path = service_dict['build']
+        build = service_dict.get('build', {})
+
+        if isinstance(build, six.string_types):
+            build_path = build
+        elif isinstance(build, dict) and 'context' in build:
+            build_path = build['context']
```
dnephin: I would expect that `build` would be normalized to always be a `dict()` by the time it gets here, do we really need the string case still?

If we do need it, could we have `process_service()` always return a dict for `build` so we don't have to worry about the variation after that point.

compose/config/validation.py
```diff
@@ -151,18 +151,29 @@ def handle_error_for_schema_with_id(error, service_name):
             VALID_NAME_CHARS)
 
     if schema_id == '#/definitions/constraints':
+        # Build context could in 'build' or 'build.context' and dockerfile could be
+        # in 'dockerfile' or 'build.dockerfile'
```
dnephin: With my proposed change to have `process_service()` always return build as a dict, I don't think we'll need to worry about this here either. We can assume a consistent representation.

docs/compose-file.md
```diff
@@ -89,6 +90,30 @@ definition.
...
+      build: .
+
+    version: 2
+    services:
+      web:
+        build:
+          context: .
+          dockerfile: Dockerfile-alternate
+          args:
+            buildno: 1
```
dnephin: We have #2622 to finalize this, but I think what we'd like to do is only include the v2 docs in this section, and have a separate section at the bottom for migration, and differences between the versions.

compose/config/config.py
GarrettHeel: The issue is that if we normalize `build` in `process_service` then v1 schema validation will fail. We could potentially add a `normalize_build()` immediately after `validate_against_service_schema`, but then we still need conditional logic in `validation.py` that you also commented on.

dnephin: Ah true, so I guess it can't be normalized before finalize_service(), ok

GarrettHeel:
I've rebased this again off master

shin-:
Closed via #2653

dnephin:
@GarrettHeel thanks again for all your work getting this in!

marcellodesales:
@GarrettHeel Thanks all!!! So waited feature!

gggeek:
cant wait for compose 1.6 to have this!

marcellodesales:
@gggeek The most complete docker-compose yet!

PR #2547, opened 2015-12-14
Add docker-compose create command.

### LINKED ISSUE #1125: docker-compose create
Possibility to just create but not run containers. This would be a useful feature for systemd based startups.

seguins:
Adds a new command  for creating a service without starting it.

Closes #1125

### DIFF
--- a/compose/project.py
+++ b/compose/project.py
@@ -123,6 +123,12 @@ class Project(object):
             [uniques.append(s) for s in services if s not in uniques]
             return uniques
 
+    def get_services_without_duplicate(self, service_names=None, include_deps=False):
+        services = self.get_services(service_names, include_deps)
+        for service in services:
+            service.remove_duplicate_containers()
+        return services
+
     def get_links(self, service_dict):
         links = []
         if 'links' in service_dict:
@@ -215,6 +221,14 @@ class Project(object):
             else:
                 log.info('%s uses an image, skipping' % service.name)
 
+    def create(self, service_names=None, strategy=ConvergenceStrategy.changed, do_build=True):
+        services = self.get_services_without_duplicate(service_names, include_deps=True)
+
+        plans = self._get_convergence_plans(services, strategy)
+
+        for service in services:
+            service.execute_convergence_plan(plans[service.name], do_build, detached=True, start=False)
+
     def up(self,
            service_names=None,
            start_deps=True,
@@ -223,10 +237,7 @@ class Project(object):
            timeout=DEFAULT_TIMEOUT,
            detached=False):
 
-        services = self.get_services(service_names, include_deps=start_deps)
-
-        for service in services:
-            service.remove_duplicate_containers()
+        services = self.get_services_without_duplicate(service_names, include_deps=start_deps)
 
         plans = self._get_convergence_plans(services, strategy)
 
--- a/tests/acceptance/cli_test.py
+++ b/tests/acceptance/cli_test.py
@@ -264,6 +264,52 @@ class CLITestCase(DockerClientTestCase):
         ]
         assert not containers
 
+    def test_create(self):
+        self.dispatch(['create'])
+        service = self.project.get_service('simple')
+        another = self.project.get_service('another')
+        self.assertEqual(len(service.containers()), 0)
+        self.assertEqual(len(another.containers()), 0)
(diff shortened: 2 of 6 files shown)


aanand:
LGTM

compose/project.py
```diff
@@ -215,6 +221,22 @@ def build(self, service_names=None, no_cache=False, pull=False, force_rm=False):
...
+
+        plans = self._get_convergence_plans(services, strategy)
+
+        for service in services:
+            plan = plans[service.name]
+            if plan.action == 'create':
+                service.create_container(do_build=True)
+            elif plan.action == 'recreate':
+                service.recreate_container(
+                    service.containers(stopped=True)[0],
```
dnephin: I don't think this is right, there could be more than one container at some point.

compose/project.py
```diff
@@ -215,6 +221,22 @@ def build(self, service_names=None, no_cache=False, pull=False, force_rm=False):
...
+                    do_build=do_build,
+                    start_new_container=False
+                )
```
dnephin: I think it would be good to keep this logic in `service.py`, since it's very similar to the existing `execute_convergence_plan`. Maybe `service.execute_create_convergence_plan()` ? It's a bit wordy, but at least it's clear.

tests/acceptance/cli_test.py
```diff
@@ -264,6 +264,74 @@ def test_build_failed_forcerm(self):
...
+        self.base_dir = 'tests/fixtures/links-composefile'
+        self.dispatch(['create', 'web'], None)
+        web = self.project.get_service('web')
+        db = self.project.get_service('db')
+        console = self.project.get_service('console')
+        self.assertEqual(len(web.containers()), 0)
+        self.assertEqual(len(db.containers()), 0)
+        self.assertEqual(len(web.containers(stopped=True)), 1)
+        self.assertEqual(len(db.containers(stopped=True)), 1)
+        self.assertEqual(len(console.containers(stopped=True)), 0)
```
dnephin: Following the [testing pyramid](http://martinfowler.com/bliki/TestPyramid.html) I think the above two tests (create_twice and with_link) should be integration tests in `tests/integration/`. We already have plenty of end-to-end coverage with the other tests you've added.

compose/project.py
aanand: `service. execute_convergence_plan(start=False)`?

seguins: @dnephin Thanks for the feedback, I will apply the solution of @aanand which seems cleaner.

compose/project.py
seguins: Indeed, it will be removed and fixed by the call to ```service.execute_convergence_plan```

compose/project.py
dnephin: I'm a little concerned that is going to make `execute_convergence_plan()` a little overly complex.  We have a max-complexity set to 12, so I guess as long as we don't have to relax that I'm cool with it.

seguins:
@dnephin thanks for the review.
I updated the commit : 
* ```project.create``` uses now ```service.execute_convergence_plan```
* I removed two acceptance tests and add new integration tests

compose/project.py
```diff
@@ -215,6 +221,14 @@ def build(self, service_names=None, no_cache=False, pull=False, force_rm=False):
             else:
                 log.info('%s uses an image, skipping' % service.name)
 
+    def create(self, service_names=None, strategy=ConvergenceStrategy.changed, do_build=True):
+        services = self.get_services_without_duplicate(service_names, include_deps=True)
+
+        plans = self._get_convergence_plans(services, strategy)
+
+        for service in services:
+            service.execute_convergence_plan(plans[service.name], do_build, start=False)
```
dnephin: Just one more small thing, I think we want to avoid calling attach here (it's an unnecessary API call), so this should add `detached=True`  as well

dnephin:
Looks good, just one more small thing

seguins:
Done

dnephin:
LGTM

thaJeztah:
@seguins looks like this needs a rebase :cry:

thaJeztah:
Oh, maybe not, it doesn't say "merge conflict", it's probably github being over restrictive (sorry for the noise :smile:)

PR #2550, opened 2015-12-15
Start, restart, pause and unpause exit with non-zero if nothing to do

### LINKED ISSUE #1236: docker-compose restart ought to exit with non-zero if it didn't actually restart
If you have no containers running and you do `docker-compose restart` it exits with 0.  It failed to restart so it should exit with non-zero.  This will let you do something like:

```
docker-compose restart || docker-compose up -d
```

dbonev:
When there are no containers to restart (pause, unpause or start respectively), compose will now exit with non-zero status, printing an error message. This allows for usage like
```
docker-compose restart || docker-compose up
```
Fixes #1236 

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -265,7 +265,8 @@ class TopLevelCommand(DocoptCommand):
 
         Usage: pause [SERVICE...]
         """
-        project.pause(service_names=options['SERVICE'])
+        containers = project.pause(service_names=options['SERVICE'])
+        exit_if(not containers, 'No containers to pause', 1)
 
     def port(self, project, options):
         """
@@ -476,7 +477,8 @@ class TopLevelCommand(DocoptCommand):
 
         Usage: start [SERVICE...]
         """
-        project.start(service_names=options['SERVICE'])
+        containers = project.start(service_names=options['SERVICE'])
+        exit_if(not containers, 'No containers to start', 1)
 
     def stop(self, project, options):
         """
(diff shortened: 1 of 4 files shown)


compose/cli/main.py
```diff
@@ -170,6 +170,11 @@ def perform_command(self, options, handler, command_options):
         with friendly_error_message():
             handler(project, command_options)
 
+    def exit_if(self, condition, message, exit_code):
+        if condition:
+            log.error(message)
+            sys.exit(exit_code)
```
dnephin: Since this method doesn't use `self`, can we make it a module level function? There are a bunch of these at the bottom of the file.

dnephin:
Thanks for the PR!

I think this looks good, just a couple small requests.

I should also note that in the case where containers already exist in the desired state there won't be any errors, but I think that's correct behaviour.

compose/cli/main.py
dnephin: I also wonder if it would be more correct  to raise an exception here, to be handled by `main()` above, which already does a bunch of this kind of thing.  That way we'd only have `sys.exit()` in a single function.

dbonev:
@dnephin Thanks for the feedback. Done both:
 - exit_if() is now module-level and moved to the bottom of the file;
 - raising SystemExit instead of manually calling sys.exit()

Thanks

aanand:
LGTM

dnephin:
LGTM

PR #1169, opened 2015-03-24
Clarified that 'rm' removes data and containers

ChrisChinchilla:
As per issue - https://github.com/docker/compose/issues/190

Signed-off-by: Chris Ward <chris@crate.io>

PR #2554, opened 2015-12-17
Clarify behaviour of 'rm'

aanand:
Replaces #1169, making a clear distinction between data inside/outside volumes, and letting the user know how to get at volumes for containers which have been removed (now that we have a `docker volume` command ✨).


dnephin:
LGTM

Issue #2503, opened 2015-12-04
Move logging configuration under `logging` key


aanand:
Right now we have `log_driver` and `log_opt`. It'd be cleaner to have them both under a `logging` key:

```yaml
logging:
  driver: syslog
  options:
    syslog-address: "tcp://192.168.0.42:123"
```

An additional advantage: overriding a service's `logging` option would be cleaner, as you only have to override a single key (you're never going to want to override `log_driver` but *not* `log_opt`, because options are driver-specific).

dbonev:
#dibs

aanand:
Closed by #2626

PR #2561, opened 2015-12-19
log_driver and log_opt moved to logging key.

dbonev:
Moving ```log_driver``` and ```log_opt``` to a new ```logging``` key:
```
logging:
    driver:
    options:
```
Implements #2503

### DIFF
--- a/compose/service.py
+++ b/compose/service.py
@@ -510,6 +510,13 @@ class Service(object):
 
         return volumes_from
 
+    def get_logging_options(self):
+        logging_dict = self.options.get('logging', {})
+        return {
+            'log_driver': logging_dict.get('driver', ""),
+            'log_opt': logging_dict.get('options', None)
+        }
+
     def _get_container_create_options(
             self,
             override_options,
@@ -522,6 +529,8 @@ class Service(object):
             (k, self.options[k])
             for k in DOCKER_CONFIG_KEYS if k in self.options)
         container_options.update(override_options)
+
+        container_options.update(self.get_logging_options())
 
         if self.custom_container_name() and not one_off:
             container_options['name'] = self.custom_container_name()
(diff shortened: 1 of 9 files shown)


compose/service.py
```diff
@@ -940,3 +941,20 @@ def build_ulimits(ulimit_config):
...
+        config=log_options
+    )
+
+
+def set_log_options(container_options):
+    if ('logging' in container_options):
+        logging_dict = container_options['logging']
+        container_options['log_driver'] = logging_dict.get('driver', "")
+        container_options['log_opt'] = logging_dict.get('options', None)
+        del container_options['logging']
```
aanand: I'd prefer if we didn't perform mutation in a utility function like this, as long as there's a nice way to do it without. Maybe:

1.  Move `'logging'` out of `DOCKER_CONFIG_OPTIONS` and into `ALLOWED_KEYS`.

2.  In `in _get_container_create_options`:

    ```python
    container_options.update(self.get_logging_options())
    ```

3. `get_logging_options`:

    ```python
    def get_logging_options(self):
        logging_dict = self.options.get('logging', {})
        return {
            'log_driver': logging_dict.get('driver', ""),
            'log_opt': logging_dict.get('options', None),
        }

aanand:
Nice, thanks. Ideally the `logging` key would only apply for users of the new file format, so this is blocking on #2421. You could implement it on top of that branch so that it's ready to go once that PR is merged.

dbonev:
@aanand Thanks for the feedback.
>I'd prefer if we didn't perform mutation in a utility function like this, as long as there's a nice way to do it without.

Note taken -- did it this way.
>so this is blocking on #2421.

I'll make sure it applies cleanly on top of #2421 branch. 
Thanks.

PR #2564, opened 2015-12-21
Declare networks in Compose file

### LINKED ISSUE #2478: Proposal: Specify networks in the Compose file
Related issues: #2470, #2476

Users should be able to specify networks in the Compose file, other than the default network, which are created on-demand. This would be useful for specifying drivers and parameters, as well as for creating multiple networks to lock down reachability between services, e.g. for security reasons.

This necessitates new Compose file syntax in the form of a new top-level key for specifying networks, and a new service-level key for specifying multiple networks to join. I propose `networks` and `publish`:

```yaml
networks:
  front:
    driver: my-custom-driver
    options: # driver-specific options
      foo: "a"
      bar: "b"
  back:
    driver: my-custom-driver

services:
  nginx:
    image: username/nginx
    publish:
      - front

  web:
    image: username/web
    publish:
      - front
      - back

  db:
    image: username/db
    publish:
      - back
```

This creates a 2-network, 3-service app where `nginx` can talk to `web` and `web` to `db`, but `nginx` cannot talk directly to `db`.

### `networks`: mapping or list?

If networks are specified as a mapping, it works really well in the case where you're specifying a driver/options. But in the (probably common) case where you just want a network and the default driver, it'd be kind of a pain to have to specify anything more than the name. Trouble is, specifying *just* the name is kind of weird/awkward with a mapping.

Let's suppose the `back` network in the example above doesn't need to specify a driver or options:

```yaml
networks:
  front:
    driver: my-custom-driver
    options:
      foo: "a"
      bar: "b"

  back: {}

  # Or allow nulls?
  back:
```

Conversely, we could use a list instead:

```yaml
networks:
  - name: front
    driver: my-custom-driver
    options:
      foo: "a"
      bar: "b"

  - name: back

  # Or allow just a string?
  - back
```

We now need to check for duplicate names, but perhaps that's OK.

### Default network

- If the user explicitly defines networks, should we still create a default network?

- If a service specifies networks to connect to with `publish`, should it also connect to the default network?

- Could we somehow support changing the name of the default network with this syntax, as requested in #2470?

### Aliases

Networking plans to support aliases under which a container can join a network (https://github.com/docker/libnetwork/issues/737), so Compose should support those. Perhaps via colon syntax:

```yaml
web:
  publish:
    - front:web-app
    - back
```

Or we could use mapping syntax, but then we run into a similar problem with the "no alias" case as described previously with the `networks` key:

```yaml
web:
  publish:
    front: web-app
    back:
```

Arguably this isn't so bad, since we already permit something similar with entries in `environment`.

### Overriding with multiple Compose files

If `docker-compose.yml` and `docker-compose.override.yml` both specify networks, how should we handle networks that are defined in both files? Should a latterly-defined network completely replace the configuration for a previously-defined network with the same name, or should we do some level of merging of keys? If so, how deeply should we merge it - should we descend into sub-keys of `options`, for example?

aanand:
Closes #2478.

TODO:
- [x] Preserve existing behaviour for legacy files
- [x] Decide on API version switching logic
- [x] Remove the `net` option in v2
- [x] Add the `networks` option, enabling services to connect to multiple networks
- [x] Don't connect containers to the `bridge` network by default
- [x] Special-case the external networks `host` and `bridge`
- [x] Support the `external` option for all other externally-defined networks

TODO in separate PRs:
- Aliasing (automatic, not user-specified)
- Validation

### DIFF
--- a/compose/config/config.py
+++ b/compose/config/config.py
@@ -139,14 +139,16 @@ class ConfigFile(namedtuple('_ConfigFile', 'filename config')):
         return {} if self.version == 1 else self.config.get('volumes', {})
 
 
-class Config(namedtuple('_Config', 'version services volumes')):
+class Config(namedtuple('_Config', 'version services volumes networks')):
     """
     :param version: configuration version
     :type  version: int
     :param services: List of service description dictionaries
     :type  services: :class:`list`
-    :param volumes: List of volume description dictionaries
-    :type  volumes: :class:`list`
+    :param volumes: Dictionary mapping volume names to description dictionaries
+    :type  volumes: :class:`dict`
+    :param networks: Dictionary mapping network names to description dictionaries
+    :type  networks: :class:`dict`
     """
 
 
@@ -256,39 +258,44 @@ def load(config_details):
     config_details = config_details._replace(config_files=processed_files)
 
     main_file = config_details.config_files[0]
-    volumes = load_volumes(config_details.config_files)
+    volumes = load_mapping(config_details.config_files, 'volumes', 'Volume')
+    networks = load_mapping(config_details.config_files, 'networks', 'Network')
     service_dicts = load_services(
         config_details.working_dir,
         main_file.filename,
         [file.get_service_dicts() for file in config_details.config_files],
         main_file.version)
-    return Config(main_file.version, service_dicts, volumes)
+    return Config(main_file.version, service_dicts, volumes, networks)
 
 
-def load_volumes(config_files):
-    volumes = {}
+def load_mapping(config_files, key, entity_type):
+    mapping = {}
+
     for config_file in config_files:
-        for name, volume_config in config_file.get_volumes().items():
-            volumes[name] = volume_config or {}
-            if not volume_config:
+        for name, config in config_file.config.get(key, {}).items():
+            mapping[name] = config or {}
+            if not config:
                 continue
 
-            external = volume_config.get('external')
+            external = config.get('external')
             if external:
-                if len(volume_config.keys()) > 1:
+                if len(config.keys()) > 1:
                     raise ConfigurationError(
-                        'Volume {0} declared as external but specifies'
-                        ' additional attributes ({1}). '.format(
+                        '{} {} declared as external but specifies'
+                        ' additional attributes ({}). '.format(
+                            entity_type,
                             name,
-                            ', '.join([k for k in volume_config.keys() if k != 'external'])
+                            ', '.join([k for k in config.keys() if k != 'external'])
                         )
                     )
                 if isinstance(external, dict):
-                    volume_config['external_name'] = external.get('name')
+                    config['external_name'] = external.get('name')
                 else:
-                    volume_config['external_name'] = name
+                    config['external_name'] = name
 
-    return volumes
+            mapping[name] = config
+
+    return mapping
 
 
 def load_services(working_dir, filename, service_configs, version):
--- a/compose/config/fields_schema_v2.json
+++ b/compose/config/fields_schema_v2.json
@@ -17,6 +17,15 @@
       },
       "additionalProperties": false
     },
+    "networks": {
+      "id": "#/properties/networks",
+      "type": "object",
+      "patternProperties": {
+        "^[a-zA-Z0-9._-]+$": {
+          "$ref": "#/definitions/network"
+        }
+      }
+    },
     "volumes": {
       "id": "#/properties/volumes",
       "type": "object",
@@ -30,6 +39,10 @@
   },
 
   "definitions": {
+    "network": {
+      "id": "#/definitions/network",
+      "type": "object"
+    },
     "volume": {
       "id": "#/definitions/volume",
       "type": ["object", "null"],
--- a/compose/project.py
+++ b/compose/project.py
@@ -64,7 +64,9 @@ class Project(object):
                 custom_networks.append(
                     Network(
                         client=client, project=name, name=network_name,
-                        driver=data.get('driver'), driver_opts=data.get('driver_opts')
+                        driver=data.get('driver'),
+                        driver_opts=data.get('driver_opts'),
+                        external_name=data.get('external_name'),
                     )
                 )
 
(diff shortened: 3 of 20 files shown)


compose/config/fields_schema_v2.json
```diff
@@ -30,6 +39,10 @@
   },
 
   "definitions": {
+    "network": {
+      "id": "#/definitions/network",
+      "type": "object"
+    },
```
dnephin: I assume this is temporary and the keys will be defined when the alias support is flushed out?

compose/project.py
```diff
@@ -399,37 +457,6 @@ def matches_service_names(container):
...
-            log.info(
-                'Creating network "{}" with {}'
-                .format(self.default_network_name, driver_name)
-            )
-            self.client.create_network(self.default_network_name, driver=self.network_driver)
-
-    def remove_network(self):
-        network = self.get_network()
-        if network:
-            self.client.remove_network(network['Id'])
```
dnephin: I think we lost this (which was previously unused, but now is used by `down`).

I think this can be `Network.remove()` now?

compose/service.py
```diff
@@ -477,9 +486,6 @@ def _next_container_number(self, one_off=False):
         return 1 if not numbers else max(numbers) + 1
 
     def _get_links(self, link_to_self):
-        if self.use_networking:
-            return []
-
```
dnephin: I think we still need a check here, otherwise `run` is going to create the `link_to_self`.  It can probably just be `is self.networks: return []` ?

tests/unit/service_test.py
```diff
@@ -484,14 +484,6 @@ def test_specifies_host_port_with_host_ip_and_port_range(self):
...
         self.assertEqual(service.specifies_host_port(), True)
 
-    def test_get_links_with_networking(self):
-        service = Service(
-            'foo',
-            image='foo',
-            links=[(Service('one'), 'one')],
-            use_networking=True)
-        self.assertEqual(service._get_links(link_to_self=True), [])
-
```
dnephin: I think this test is still important for the `run` case.

compose/project.py
```diff
@@ -145,6 +171,21 @@ def get_services_without_duplicate(self, service_names=None, include_deps=False)
...
+                networks.append(name)
+            else:
+                matches = [n for n in network_definitions if n.name == name]
+                if matches:
+                    networks.append(matches[0].full_name)
+                else:
+                    raise ConfigurationError(
+                        'Service "{}" uses an undefined network "{}"'
+                        .format(service_dict['name'], name))
+        return networks
```
dnephin: Not a blocker, but this doesn't use `self` so it doesn't need to be a method on `Project` at all. It can be just a function at the module level.

I think that's good for consistency as well, since it's called from a `@classmethod` for the `Project` is fully initialized.

Maybe it could actually be a function in the `network.py` module ?

compose/project.py
```diff
@@ -55,23 +58,47 @@ def from_config(cls, name, config_data, client):
...
                     use_networking=use_networking,
+                    networks=networks,
                     links=links,
                     net=net,
                     volumes_from=volumes_from,
                     **service_dict))
+
+        project.networks += custom_networks
+        if project.uses_default_network():
+            project.networks.append(project.default_network)
```
dnephin: I think this is duplicated on `initialize_network()` ?

compose/config/fields_schema_v2.json
aanand: Yeah, I'll define the keys properly as part of the "validation" work.

compose/project.py
aanand: Yep, have implemented this locally as part of rebasing on the `down` PR - pushing soon.

compose/service.py
aanand: Ah yes. In fact, it still needs to be `if self.use_networking` - it's possible for a service in a v2 app to have no networks (by defining `networks: []`).

tests/unit/service_test.py
aanand: Yep.

compose/project.py
aanand: So it is - removing.

compose/config/config.py
```diff
@@ -256,39 +258,44 @@ def load(config_details):
...
-def load_volumes(config_files):
-    volumes = {}
+def load_mapping(config_files, key, entity_type):
+    mapping = {}
+
     for config_file in config_files:
-        for name, volume_config in config_file.get_volumes().items():
-            volumes[name] = volume_config or {}
-            if not volume_config:
+        for name, config in config_file.config.get(key, {}).items():
```
dnephin: I think there's a subtle edge case here for V1 config. If someone has a service named `volume` or `network` this is going to crash, or break in some unexpected way.

I think we should use the `get_volumes()` function I added to `ConfigFile`, and create a similar one for `get_networks()`. They switch on version, so we won't hit that case.

Instead of `key` you can take `func_name` and do:

`getattr(config_file, func_name)().items()`

compose/project.py
```diff
@@ -55,23 +58,47 @@ def from_config(cls, name, config_data, client):
...
+                    )
+                )
 
         for service_dict in config_data.services:
-            links = project.get_links(service_dict)
+            if use_networking:
+                networks = get_networks(
+                    service_dict,
+                    custom_networks + [project.default_network])
+                net = Net(networks[0]) if networks else Net("none")
```
dnephin: I think this means we'll attempt to re-connect to the first network twice?  Maybe that's fine? I don't really know.

compose/project.py
```diff
@@ -251,7 +275,7 @@ def initialize_volumes(self):
     def down(self, remove_image_type, include_volumes):
         self.stop()
         self.remove_stopped(v=include_volumes)
-        self.remove_network()
+        self.remove_default_network()
```
dnephin: I think this needs to remove all networks, not just the default.  Since there's a TODO on not removing external, I guess this could be a TODO as well

dnephin:
I'd be fine fixing these as follow-ups, so LGTM

PR #2566, opened 2015-12-22
Fix `run` when using a non-default log driver

### LINKED ISSUE #2448: Using gelf log driver and lose the ability to `fig run <container> bash`
Im using `gelf` logging in my docker-compose file and I lose the ability to use `fig run <container> bash` because it gives the error "Error attaching: configured logging reader does not support reading". An example config is:

```
my_container:
  image: link-to-image
  log_driver: "gelf"
  log_opt:
    gelf-address: udp://0.0.0.0:12201
    gelf-tag: my_container
```

I understand that I lose the ability to use `docker-compose logs` or `docker-compose logs my_container` but I would expect `docker-compose run my_container bash` to let me into the container as I am overriding the CMD line manually.

```
docker-compose version: 1.5.1
Docker version 1.9.0, build 76d6bc9
```

alf:
Fixes #2448

As soon as @aanand releases a new version of dockerpty, this change should be enough to fix #2448. I'm opening the PR early to give time for feedback etc. Maybe there are tests that need to be written or something for the PR to be acceptable?

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -667,7 +667,7 @@ def run_one_off_container(container_options, project, service, options):
         sys.exit(1)
 
     set_signal_handler(shutdown)
-    dockerpty.start(project.client, container.id, interactive=not options['-T'])
+    dockerpty.start(project.client, container.id, interactive=not options['-T'], logs=False)
     exit_code = container.wait()
     remove_container()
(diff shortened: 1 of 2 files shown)


alf:
@aanand It would be great if you could make a release of dockerpty that includes the logs-option.

PR #2584, opened 2015-12-28
Update `volumes_from` docs to state default

mjgpy3:
[Proof of read-write](https://github.com/docker/compose/blob/cfb1b37da22242dc67d5123772b3fa4518458504/compose/config/types.py#L26).

I found myself wondering what the default was a couple of times, and finally decided to change it :)


dnephin:
LGTM

aanand:
LGTM

PR #2585, opened 2015-12-28
Fix signal handlers by moving shutdown logic out of handler

### LINKED ISSUE #2543: regression in 1.5.2 Ctrl+C force stop is not working
In previous version 1.5.1 was working. Tested in two computers, in one directly with 1.5.2, in the other was working with 1.5.1 after update to 1.5.2, no.

So it stops because of the timeout

Example output, but here you cannot see nothing strange:
```
Stopping pep-container ... done
Stopping idm ... done
Stopping pdp-container ... done
```

### LINKED ISSUE #2580: Regression in 1.5.2 re: hitting Control-C when running `up`
It looks like there is a regression in docker-compose 1.5.2 when compared with 1.5.1.

Previously, after running `docker-compose up` in a terminal window, pressing `Control-C` (i.e. sending SIGINT from the same shell window) caused services to enter the "Stopping" phase.  However, when I upgraded to 1.5.2, services now immediately go to "Killing."

Note that if I go to a _different_ terminal window and send a SIGINT or SIGTERM to that process using `kill`, then the services _do_ go to the "Stopping" phase.  So the issue looks like it is restricted to hitting `Control-C` from the same shell window.  I'm observing this on Mac OS X 10.11 using Python 3.5.1.

One more thing.  I did notice one improvement with 1.5.2.  Previously, sending a SIGTERM to `docker-compose up` from a different window didn't seem to work and stop services like SIGINT.  With 1.5.2, it now does, and behaves the same as sending SIGINT from a different window.

So the issue now is with hitting Control-C from the same window.  My guess is that this might have to do with the fact that hitting Control-C in a terminal window sends a SIGINT not just to the parent process, but also to all child processes (and there is at least one child with docker-compose).  So maybe docker-compose is treating that as receiving two distinct SIGINT signals, causing it to go immediately into the "Killing" phase (as if the user hit Control-C twice).

dnephin:
Fixes #2543
Fixes #2580

As suggested by @cjerdonek in #2543, the issue with our shutdown signal handlers may be that we're doing too much in the handler itself. This change moves all the shutdown into a try/except in the main flow, and only raises an exception in the handler.

With this change #2543 seems to be resolved. If anyone is still able to reproduce the issue, please let me know.

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -3,7 +3,6 @@ from __future__ import unicode_literals
 
 import logging
 import re
-import signal
 import sys
 from inspect import getdoc
 from operator import attrgetter
@@ -12,6 +11,7 @@ import yaml
 from docker.errors import APIError
 from requests.exceptions import ReadTimeout
 
+from . import signals
 from .. import __version__
 from ..config import config
 from ..config import ConfigurationError
@@ -655,20 +655,19 @@ def run_one_off_container(container_options, project, service, options):
         if options['--rm']:
             project.client.remove_container(container.id, force=True)
 
-    def force_shutdown(signal, frame):
+    signals.set_signal_handler_to_shutdown()
+    try:
+        try:
+            dockerpty.start(project.client, container.id, interactive=not options['-T'])
+            exit_code = container.wait()
+        except signals.ShutdownException:
+            project.client.stop(container.id)
+            exit_code = 1
+    except signals.ShutdownException:
         project.client.kill(container.id)
         remove_container(force=True)
         sys.exit(2)
 
-    def shutdown(signal, frame):
-        set_signal_handler(force_shutdown)
-        project.client.stop(container.id)
-        remove_container()
-        sys.exit(1)
-
-    set_signal_handler(shutdown)
-    dockerpty.start(project.client, container.id, interactive=not options['-T'])
-    exit_code = container.wait()
     remove_container()
     sys.exit(exit_code)
 
@@ -683,24 +682,18 @@ def build_log_printer(containers, service_names, monochrome):
 
 
 def attach_to_logs(project, log_printer, service_names, timeout):
+    print("Attaching to", list_containers(log_printer.containers))
+    signals.set_signal_handler_to_shutdown()
 
-    def force_shutdown(signal, frame):
+    try:
+        try:
+            log_printer.run()
+        except signals.ShutdownException:
+            print("Gracefully stopping... (press Ctrl+C again to force)")
+            project.stop(service_names=service_names, timeout=timeout)
+    except signals.ShutdownException:
         project.kill(service_names=service_names)
         sys.exit(2)
-
-    def shutdown(signal, frame):
-        set_signal_handler(force_shutdown)
-        print("Gracefully stopping... (press Ctrl+C again to force)")
-        project.stop(service_names=service_names, timeout=timeout)
-
-    print("Attaching to", list_containers(log_printer.containers))
-    set_signal_handler(shutdown)
-    log_printer.run()
-
-
-def set_signal_handler(handler):
-    signal.signal(signal.SIGINT, handler)
-    signal.signal(signal.SIGTERM, handler)
 
 
 def list_containers(containers):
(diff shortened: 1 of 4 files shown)


compose/cli/main.py
```diff
@@ -655,20 +655,19 @@ def remove_container(force=False):
...
-
-    def shutdown(signal, frame):
-        set_signal_handler(force_shutdown)
-        project.client.stop(container.id)
-        remove_container()
-        sys.exit(1)
+    signals.set_signal_handler_to_shutdown()
+    try:
+        dockerpty.start(project.client, container.id, interactive=not options['-T'])
+        exit_code = container.wait()
```
cjerdonek: It looks like there are some race conditions around here, for example if ForceShutdownException occurs at this line.

dnephin: There are definitely a few race conditions here. I'm not sure if it's possible to reliably handle all of them. 

I'd rather not make the code a lot uglier to handle a case we can't reproduce in practice.  I'll see if I can get any of them to happen.

cjerdonek: > For example, if the second signal fires before set_signal_handler(force_shutdown) is set, there isn't much we can do about that.
Yes, but at least in that scenario the user can Control-C again. In the scenario I mentioned above, the user doesn't have a second chance.

I'm not sure if it's possible with this approach to handle all of the essential cases cleanly. I proposed an alternative approach in the issue discussion earlier today, but it would take more work. I would favor any approach that moves closer to some of the ideas I mentioned. I think it would make subsequent changes easier to do and think about, etc. With this type of exception handling, it is not so easy, and the logic is repeated twice (and would need to be more times if additional operations are added).

dnephin: The logic is not repeated. `up` and `run` do **different** things. The shutdown operation is different.  

We gain nothing by moving them together, it would actually make the code harder to follow because we'd have unrelated shutdown happening in the same function.

dnephin:
I've simplified the implementation by removing the need to set a second signal handler with a different exception.  Now the same exception will be raised again to perform the force shutdown.

In addition to the automated coverage, I've tested this with
```
kill $pid; kill $pid
```

When this is run against a `docker-compose up` process, the second `kill` doesn't do anything.  Sending another kill at any time after this will correctly trigger the kill.

```
kill $pid; sleep 0.0001; kill $pid
```

With this command the kill shutdown is performed correctly.

cjerdonek:
It looks like your latest patch is susceptible to additional race conditions whereby two or three near-concurrent signals can cause your except blocks to be skipped entirely (so that additional signals after that won't get you back on track). This is mainly a consequence of issues with asynchronous exceptions in Python. My advice would be not to raise any exceptions from your signal handler and to use a pattern like what I suggested in [this comment](https://github.com/docker/compose/issues/2543#issuecomment-167901816) (minus raising the interrupt exception at the end).

dnephin:
> susceptible to additional race conditions whereby two or three near-concurrent signals can cause your except blocks to be skipped entirely

I was able to reproduce this one using `parallel kill $pid ::: $pid ::: $pid`, but it's very hard to reproduce.

Either way, I've pushed a new version which fixes the issue. The behaviour is now in-line with my previous comment. Subsequent signals after a short delay will properly run the kill operation.

Your suggestion exhibits the same problems as I describe in my comment: https://gist.github.com/dnephin/a0a7a155fdea2ba17627

I believe this is unavoidable as one signal can interrupt the previous signal handler. It is not a race condition.

cjerdonek:
Yes, signal handlers can interrupt signal handlers.  But no, my suggestion is not susceptible to the more important race condition that I described, namely that your except blocks can _bypassed entirely_ in the presence of asynchronous exceptions so that no additional signals at a later (even delayed) time will result in your cleanup.

In my suggestion with a signal handler with no interrupt exception at the end, the cleanup / end of the signal handlers is guaranteed to run.  Yes, it is true that setting the flag can be missed at the beginning.  But additional signals later on _are_ guaranteed to result in the final cleanup / kill unlike with your patch.

shin-:
LGTM as far as I'm concerned. If there are issues not covered by this specific PR, we can always improve on it in the next iteration.

Needs to be rebased however.

Issue #2586, opened 2015-12-29
Option to specify CWD


vaab:
Skimming through the issues, it seems that you've chosen to set the paths of `docker-compose.yml` (for instance the `build` path) relative to `docker-compose.yml` file's own location.

This doesn't suits my current needs, and I do not think you should take this setup for granted to the point where we do not have any options to bypass this behavior. I'll have to make gruesome hacks to go around this.

So as a proposition, why not simply add an optional general argument `--cwd PATH`  (as the `-f FILENAME` is), that would allow more flexibility and compatibility with other behaviors ?

I'm using a generated `docker-compose.yml` to hook Juju style of services description. And this CWD following the config file is a pain in the ass currently (and it has already generated quite some puzzling it seems) , so if at least we could prevent this behavior with an optional argument (no compatibility break), this would be great.

PR #2594, opened 2016-01-02
Added ability to change the working directory via a command-line flag

dbonev:
Using the --cwd PATH option from the CLI, it is now possible to change the working directory to a path different from that of the compose file.

Implements #2586

Signed-off-by: Dimitar Bonev dimitar.bonev@gmail.com

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -160,6 +160,8 @@ class TopLevelCommand(object):
       --skip-hostname-check       Don't check the daemon's hostname against the name specified
                                   in the client certificate (for example if your docker host
                                   is an IP address)
+      --project-directory PATH    Specify an alternate working directory
+                                  (default: the path of the compose file)
 
     Commands:
       build              Build or rebuild services
--- a/compose/config/config.py
+++ b/compose/config/config.py
@@ -218,10 +218,10 @@ class ServiceConfig(namedtuple('_ServiceConfig', 'working_dir filename name conf
             config)
 
 
-def find(base_dir, filenames, environment):
+def find(base_dir, filenames, environment, override_dir='.'):
     if filenames == ['-']:
         return ConfigDetails(
-            os.getcwd(),
+            os.path.abspath(override_dir),
             [ConfigFile(None, yaml.safe_load(sys.stdin))],
             environment
         )
@@ -233,7 +233,7 @@ def find(base_dir, filenames, environment):
 
     log.debug("Using configuration files: {}".format(",".join(filenames)))
     return ConfigDetails(
-        os.path.dirname(filenames[0]),
+        override_dir or os.path.dirname(filenames[0]),
         [ConfigFile.from_filename(f) for f in filenames],
         environment
     )
--- a/compose/cli/command.py
+++ b/compose/cli/command.py
@@ -28,7 +28,8 @@ def project_from_options(project_dir, options):
         verbose=options.get('--verbose'),
         host=options.get('--host'),
         tls_config=tls_config_from_options(options),
-        environment=environment
+        environment=environment,
+        override_dir=options.get('--project-directory'),
     )
 
 
@@ -59,10 +60,10 @@ def get_client(environment, verbose=False, version=None, tls_config=None, host=N
 
 
 def get_project(project_dir, config_path=None, project_name=None, verbose=False,
-                host=None, tls_config=None, environment=None):
+                host=None, tls_config=None, environment=None, override_dir=None):
     if not environment:
         environment = Environment.from_env_file(project_dir)
-    config_details = config.find(project_dir, config_path, environment)
+    config_details = config.find(project_dir, config_path, environment, override_dir)
     project_name = get_project_name(
         config_details.working_dir, project_name, environment
     )
--- a/docs/reference/overview.md
+++ b/docs/reference/overview.md
@@ -39,6 +39,8 @@ Options:
   --skip-hostname-check       Don't check the daemon's hostname against the name specified
                               in the client certificate (for example if your docker host
                               is an IP address)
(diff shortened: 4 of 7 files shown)


dbonev:
Hmm... all tests look fine on my machine. It's not clear to me why the Jenkins build failed (also the errors look somewhat strange and unrelated).

dnephin:
I've restarted the build, I think the failures were unrelated.

LGTM when tests are green

compose/config/config.py
```diff
@@ -133,10 +133,10 @@ def with_abs_paths(cls, working_dir, filename, name, config):
             config)
 
 
-def find(base_dir, filenames):
+def find(base_dir, filenames, override_dir=None):
     if filenames == ['-']:
         return ConfigDetails(
-            os.getcwd(),
+            os.getcwd() if not override_dir else override_dir,
```
aanand: Could this be shortened to `override_dir or os.getcwd()`?

compose/config/config.py
```diff
@@ -146,7 +146,7 @@ def find(base_dir, filenames):
 
     log.debug("Using configuration files: {}".format(",".join(filenames)))
     return ConfigDetails(
-        os.path.dirname(filenames[0]),
+        os.path.dirname(filenames[0]) if not override_dir else override_dir,
```
aanand: Similarly, could this be shortened to `override_dir or os.path.dirname(filenames[0])`?

aanand:
Should we have an acceptance test for this?

dnephin:
Ya, I guess the test doesn't cover all the code changes

dbonev:
@aanand, @dnephin Many thanks for the feedback.
I changed the code as requested and added acceptance tests. Unfortunately, the Jenkins build failed with an (I think) unrelated timeout error in a totally unrelated test. :disappointed: 

Again, it's all green on my machine, but please let me know if I'm missing something or something doesn't look right.

dnephin:
I think it's unrelated, I restarted the build

dbonev:
@dnephin Thanks

dnephin:
LGTM

dbonev:
Synced to the ```HEAD```. The test failure seems unrelated.

simonvanderveldt:
@dbonev What does `cwd` stand for? Compose working directory?
Isn't this more of a compose project thing and thus more analogous to `project_name` should be named `project_directory`?
Especially since by default the project name is the directory name containing the `docker-compose.yml` file (or the one where `docker-compose` is run from?).

Asking as I think it will probably be pretty confusing having different types of `working` directories.

Also, what's the exact use case for this? Does it include building and volume mounting `./<filename>` as well? And override/additional `docker-compose.*.yml` files?

dnephin:
Considering that I already made that mistake, I think I agree. Calling it project-directory might be good.

I think a good use-case is when you use stdin instead of a file path, you can set the project directory to something other than the current working directory.

dbonev:
OK, I'll resubmit as soon as I can. I'll change ```cwd``` to ```project-directory```.

dbonev:
Done. Sorry for the delay. Kinda overloaded lately.

compose/cli/main.py
```diff
@@ -144,6 +144,8 @@ class TopLevelCommand(DocoptCommand):
       -f, --file FILE           Specify an alternate compose file (default: docker-compose.yml)
       -p, --project-name NAME   Specify an alternate project name (default: directory name)
       --verbose                 Show more output
+      --project-directory PATH                Specify an alternate working directory
```
dnephin: looks like the formatting here is off

dnephin:
Just one small issue with formatting, otherwise LGTM

aanand:
Yep - LGTM after the formatting fix

simonvanderveldt:
I think it would be a good idea to add a test to make sure that a volume mounted file from the project-dir works.

compose/cli/main.py
```diff
@@ -144,6 +144,8 @@ class TopLevelCommand(DocoptCommand):
       -f, --file FILE           Specify an alternate compose file (default: docker-compose.yml)
       -p, --project-name NAME   Specify an alternate project name (default: directory name)
       --verbose                 Show more output
+      --project-directory PATH  Specify an alternate working directory
+                                (default: the path of the compose file)
```
dnephin: hmm, I think this should say "the directory of the compose file" or "the dirname of the compose file".

compose/config/config.py
```diff
@@ -205,10 +205,10 @@ def with_abs_paths(cls, working_dir, filename, name, config):
             config)
 
 
-def find(base_dir, filenames):
+def find(base_dir, filenames, override_dir=None):
     if filenames == ['-']:
         return ConfigDetails(
-            os.getcwd(),
+            override_dir or os.getcwd(),
```
dnephin: I just noticed something which I missed earlier.

We already have `base_dir` here. I think setting `project-directory` should be equivalent to setting `base_dir`. I think we can probably re-use it. Instead of using `os.getcwd()`, this can be `os.abspath(base_dir)`, so when we pass in the default of `.` it will be expanded to the same path.

Sorry I didn't notice this earlier.

bfirsh:
Thanks for the contribution, @dbonev! Would you mind addressing the comments and rebasing, then we can get this in. 💃

dbonev:
@bfirsh  Yeah, I'll look at it asap. Sorry for the long delay (totally my bad, been quite busy lately).

andrerom:
Side: Is this only to be able to use different `.env` files?

If so a perhaps more direct fix would be to be able to set path to `.env` file *(including different name so you can have `.env_dev`)* either via argument and/or environment variable like partly proposed in description of #3399.

shin-:
Rebased and merged in #4589 - Thank you!

PR #2604, opened 2016-01-05
Improve the Wordpress tutorial by removing the deprecated docker links

kevinsimper:
I updated the wordpress guide and I did test it on my on computer.

It was necessary to add `-p wordpress` because you don't necessarily know what the docker-compose project is going to be called unless set explicit.

### DIFF
--- a/docs/wordpress.md
+++ b/docs/wordpress.md
@@ -46,8 +46,6 @@ and a separate MySQL instance:
       command: php -S 0.0.0.0:8000 -t /code
       ports:
         - "8000:8000"
-      links:
-        - db
       volumes:
         - .:/code
     db:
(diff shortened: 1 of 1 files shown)


aanand:
Thanks. I don't think we want to update the guides until networking is out of the experimental stage, at which point

- there'll be no need for the `--x-networking` flag
- there'll be no need to specify the full db container name, and therefore no need to specify the project name, as the `db` hostname will resolve correctly.

kevinsimper:
Okay, that cool, I am just very exited about networks! :dancers: 

> there'll be no need to specify the full db container name, and therefore no need to specify the project name, as the db hostname will resolve correctly.

In docker-compose 1.5.2 is all dns names prefixed with the project name. Is that coming in next release?

dnephin:
Yes, in 1.6 the containers now have the same aliases as before (the service name, and container short id).

Thanks for this PR. We've updated the tutorials for the new version of Compose. Those changes should be published soon.

kevinsimper:
:+1:

Issue #2481, opened 2015-12-01
Previous container volumes warnings even with `--force-recreate`


pandeiro:
When calling `docker-compose up --force-recreate`, I see warnings about volumes from previous containers being used, even though, as I understand it (per conversion w/ @dnephin on #docker-compose), this should be a non-issue.

```
WARNING: Service "frontend" is using volume "/.foo" from the previous container. Host mapping "/home/user/.foo" has no effect. Remove the existing containers (with `docker-compose rm frontend`) to use the host volume mapping.
```

dnephin:
This is a regression in 1.5.1, which is fixed for 1.5.2 (to be released soon).


You can grab a version of master: https://dl.bintray.com/docker-compose/master/ if you'd like to try it out before the 1.5.2 release.

Related issue: #2390

aanand:
1.5.2 is out now:

https://github.com/docker/compose/releases/tag/1.5.2

tian2992:
I have this same issue on docker-compose version 1.5.2 
```
WARNING: Service "data" is using volume "/var/lib/postgresql" from the previous container. Host mapping "None" has no effect. Remove the existing containers (with `docker-compose rm data`) to use the host volume mapping.
```
```
docker-compose version 1.5.2, build unknown
```

dnephin:
The warning is still there, it should only be printed when a volume is actually being masked. 

Could you include your `docker-compose.yml` and the output of a docker inspect for the container?

jrhite:
I just upgraded and see the same thing:

`docker-compose version 1.5.2, build 7240ff3`

```
$ docker-compose -f myapp-staging.yml up -d --force-recreate web
Recreating myapp_data_1
WARNING: Service "data" is using volume "/var/lib/postgresql" from the previous container. Host mapping "None" has no effect. Remove the existing containers (with `docker-compose rm data`) to use the host volume mapping.
Recreating myapp_postgres_1
Recreating myapp_web_1
WARNING: Service "web" is using volume "/usr/src/app/myapp/ui/dist/static" from the previous container. Host mapping "None" has no effect. Remove the existing containers (with `docker-compose rm web`) to use the host volume mapping.
WARNING: Service "web" is using volume "/usr/src/app/landingpage" from the previous container. Host mapping "None" has no effect. Remove the existing containers (with `docker-compose rm web`) to use the host volume mapping.
```

docker-compose.yml:
```
web:
  restart: 'always'
  build: ./web
  expose:
    - '8000'
[...]

PR #2609, opened 2016-01-05
Don't warn when the container volume is a compose option

dnephin:
Fixes: https://github.com/docker/compose/issues/2481#issuecomment-168874417

Another case where the volume masking warning is incorrect. 

If the volume comes from the Compose options the host paths will be different, but there is nothing being masked, so we shouldn't warn here either.


shin-:
Needs rebasing, otherwise LGTM.

PR #2612, opened 2016-01-06
Fix extends with multiple files

dnephin:
Fixes https://github.com/docker/compose/commit/64447879d2f5a2fe5b8b50819b6620b759715a9d#commitcomment-15296425

cc @loic


loic:
Tested it, all good. Thanks for the quick turnaround!

shin-:
LGTM!

Issue #2502, opened 2015-12-04
Syntax for locally-declared versus externally-created entities


aanand:
In the next version of the Compose file format, we plan to implement support for declaring networks and volumes in the Compose file, for use by services declared in the same application (see #2421, #2478).

It's likely that, as @dnephin raises in point 2 of https://github.com/docker/compose/issues/2478#issuecomment-160784494, we'll need to prefix the names of networks/volumes with the project name so that they don't conflict with networks/volumes declared in other applications.

This raises the issue of how to refer to *locally-declared* versus *externally-created* networks/volumes. Our requirements are:

1.  I want to define a network/volume in my Compose file with a particular name, and later refer to it by that name (without having to prefix it with an unpredictable project name). It'll be defined in a block looking something like this:

    ```yaml
    networks:
      locally-defined-network:
        driver: default
    volumes:
      locally-defined-volume:
        driver: default
    ```

2.  I want to create a network/volume manually via `docker network create` or `docker volume create` with a particular name, and later refer to it by that name.

    Note that this is already possible: you can refer to an externally-created network with `net`, or an externally-created volume with `volumes`. (We're *not* obliged to carry the existing syntax over, however - we're defining a new structure for the Compose file, so it's acceptable to introduce changes as long as the old syntax continues to work for files using the old format.)

When looking at a name in `net` or `volumes`, it's **not** an acceptable solution to first check for a locally-defined entity of the appropriate type that matches the name, and fall back to assuming that it's an externally-created entity if we don't find one. As described in #2501, this is error-prone and could make the file harder to read for users who aren't fluent in the Compose file format.

I have a couple of ideas as to how we could approach it. A key question, unanswered by either, is of which case should be considered the *default* - local versus external.

### Separate keys

Instead of overloading `net` and `volumes` with two different behaviours, we could split them into separate config options (we can bikeshed the actual prefix/suffix later if we choose to go down this route):

```yaml
local_net: locally-defined-network
net: externally-created-network
local_volumes: ["locally-defined-volume:/data"]
volumes: ["externally-created-volume:/data"]
```

or, more in line with with `links` and `external_links`:

```yaml
net: locally-defined-network
external_net: externally-created-network
volumes: ["locally-defined-volume:/data"]
external_volumes: ["externally-created-volume:/data"]
```

Arguably, the example above implies that old-style volume bind mounts like `.:/code` should also live under `external_volumes`, rather than `volumes`, which feels a bit weird.

### Sigils

[...]

PR #2621, opened 2016-01-07
Don't recreate pre-existing volumes.

shin-:
During the `initialize_volumes` phase, if a volume using the non-namespaced name already exists, don't create the namespaced equivalent.


dnephin:
Closing since we're going  with https://github.com/docker/compose/issues/2502#issuecomment-170112554

PR #2625, opened 2016-01-08
Allow Entrypoints to be Lists

kojiromike:
Currently the documentation for compose says that entrypoints have to be strings. I gather this is to model them after `docker run --entrypoint "blah"`. But docker-compose accepts `command`s that are lists, and used to accept list `entrypoint`s as well. In fact, they still work, except for the validation preventing their definition.

I'd like to have a discussion about this – I'm happy to enhance this PR by updating the documentation as well, if the compose team is amenable to it.

### DIFF
--- a/compose/config/service_schema_v1.json
+++ b/compose/config/service_schema_v1.json
@@ -34,7 +34,12 @@
         "dns_search": {"$ref": "#/definitions/string_or_list"},
         "dockerfile": {"type": "string"},
         "domainname": {"type": "string"},
-        "entrypoint": {"$ref": "#/definitions/string_or_list"},
+        "entrypoint": {
+          "oneOf": [
+            {"type": "string"},
+            {"type": "array", "items": {"type": "string"}}
+          ]
+        },
         "env_file": {"$ref": "#/definitions/string_or_list"},
         "environment": {"$ref": "#/definitions/list_or_dict"},
 
--- a/docs/compose-file.md
+++ b/docs/compose-file.md
@@ -120,6 +120,10 @@ Override the default command.
 
     command: bundle exec thin -p 3000
 
+The command can also be a list, in a manner similar to [dockerfile](https://docs.docker.com/engine/reference/builder/#cmd):
+
+    command: [bundle, exec, thin, -p, 3000]
+
 ### cgroup_parent
 
 Specify an optional parent cgroup for the container.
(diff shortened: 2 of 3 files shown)


dnephin:
Ah, the problem with `list_or_string` is that it requires unique items. I think this change makes sense, entrypoint is supposed to accept an array as well.

LGTM

kojiromike:
Updated to cover documentation.

docs/compose-file.md
```diff
@@ -120,6 +120,15 @@ Override the default command.
...
     command: bundle exec thin -p 3000
 
+The command can also be a list, in a manner similar to [dockerfile](https://docs.docker.com/engine/reference/builder/#cmd):
+
+    command:
+        - bundle
+        - exec
+        - thin
+        - -p
+        - 3000
```
dnephin: Maybe a minor point, but I would personally use the `[]` form of a list for something like this:

```yaml
command: [bundle, exec, thin, -p, 3000]
```

docs/compose-file.md
```diff
@@ -174,6 +183,22 @@ specified using the `build` key.
 
 Using `dockerfile` together with `image` is not allowed. Attempting to do so results in an error.
 
+### entrypoint
+
+Override the default entrypoint.
```
dnephin: Nice, I think we should move this into it's own section

dnephin:
cc @moxiegirl for the docs changes

moxiegirl:
LGTM  Thanks @kojiromike

kojiromike:
Cool. Let me know if you want me to tweak anything else before merge.

dnephin:
LGTM

kojiromike:
Rebased because someone moved code from field_schema to service_schema.

dnephin:
I've got a fix for the build failure in #2639 if you want to rebase on that (to get the tests passing again). It should be merged in a minute.

kojiromike:
Rebased

compose/config/service_schema_v1.json
```diff
@@ -34,7 +34,12 @@
         "dns_search": {"$ref": "#/definitions/string_or_list"},
         "dockerfile": {"type": "string"},
         "domainname": {"type": "string"},
-        "entrypoint": {"$ref": "#/definitions/string_or_list"},
+        "entrypoint": {
+          "oneOf": [
+            {"type": "string"},
+            {"type": "array", "items": {"type": "string"}}
+          ]
+        },
```
dnephin: We're going to need this change in `compose/config/service_schema_v2.json` as well now, sorry about that.

Coping it over should be the only change.

kojiromike:
Done

dnephin:
LGTM, Thanks!

docs/compose-file.md
```diff
@@ -460,10 +480,6 @@ Each of these is a single value, analogous to its
     cpu_quota: 50000
     cpuset: 0,1
 
-    entrypoint: /code/entrypoint.sh
-    user: postgresql
-    working_dir: /code
-
```
dnephin: Looks like a couple extra lines got removed here, could you restore these two?

aanand:
LGTM

PR #2626, opened 2016-01-08
Move logging config under separate key in v2 compose files

shin-:
Implements #2503, based on #2561


tests/integration/service_test.py
```diff
@@ -902,7 +902,7 @@ def test_log_drive_empty_default_jsonfile(self):
         self.assertFalse(log_config['Config'])
 
     def test_log_drive_none(self):
-        service = self.create_service('web', log_driver='none')
+        service = self.create_service('web', logging={'driver': 'none'})
```
dnephin: I think we want to keep these existing tests and just rename them to be V1 tests, and add new ones for the V2 format

dnephin: nevermind, I guess at this point the config has been normalized to the v2 format.

compose/service.py
```diff
@@ -523,6 +530,8 @@ def _get_container_create_options(
             for k in DOCKER_CONFIG_KEYS if k in self.options)
         container_options.update(override_options)
 
+        container_options.update(self.get_logging_options())
+
```
dnephin: Why is this required? I thought we only had to provide it to the HostConfig section which is done below.

compose/config/config.py
```diff
@@ -505,6 +504,16 @@ def finalize_service(service_config):
...
 
+    if 'log_driver' in service_dict or 'log_opt' in service_dict:
+        if 'logging' not in service_dict:
+            service_dict['logging'] = {}
+        if 'log_driver' in service_dict:
+            service_dict['logging']['driver'] = service_dict['log_driver']
+            del service_dict['log_driver']
+        if 'log_opt' in service_dict:
+            service_dict['logging']['options'] = service_dict['log_opt']
+            del service_dict['log_opt']
```
dnephin: Could we put this into a new function called something like `normalize_v1_format()` (and call it from here)? That way when we go to remove V1, it's clear which normalize steps are done for legacy compatibility, and which ones are still required.

tests/acceptance/cli_test.py
```diff
@@ -716,6 +716,57 @@ def test_start_no_containers(self):
...
+            ],
+            None)
+
+        containers = self.project.containers()
+        self.assertEqual(len(containers), 2)
+
+        another = self.project.get_service('another').containers()[0]
+        log_config = another.get('HostConfig.LogConfig')
+        self.assertTrue(log_config)
+        self.assertEqual(log_config.get('Type'), 'none')
```
dnephin: I think this last test could probably be an integration test, instead of an acceptance test, since it's concerned with more fine grained behaviour, not just the high level feature, but it's not a blocker.

dnephin:
A question and a request, but generally looks good.

I'd really like to merge the fields_schema and service_schema into a single file, but I guess it's not urgent.

compose/service.py
shin-: You're right, this doesn't look useful since those keys will be removed a few lines later.

shin-:
@dnephin Addressed everything - PTAL.

compose/config/config.py
dnephin: A very small style nit (not a blocker, more of just an fyi), you could write this function with just 3 lines:

```python
service_dict.setdefault('logging', {})
service_dict['logging']['driver'] = service_dict.pop('log_driver', '')
service_dict['logging']['options'] = service_dict.pop('log_opt', None)
```

Since we set these keys to the default blank values later on anyway, it doesn't hurt to set them to blank values earlier on.

dnephin:
LGTM! thanks for making that test change

aanand:
LGTM

PR #2627, opened 2016-01-10
Update roadmap

bfirsh:
Bring it inline with current plans:

- There should be ways of delaying/defining service startup
- Use in production is not necessarily about the command-line
  tool, but also improving file format, integrations, new tools,
  etc.

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -1,13 +1,21 @@
 # Roadmap
 
+## An even better tool for development environments
+
+Compose is a great tool for development environments, but it could be even better. For example:
+
+- It should be possible to define hostnames for containers which work from the host machine, e.g. “mywebcontainer.local”. This is needed by apps comprising multiple web services which generate links to one another (e.g. a frontend website and a separate admin webapp)
+
 ## More than just development environments
 
-Over time we will extend Compose's remit to cover test, staging and production environments. This is not a simple task, and will take many incremental improvements such as:
+Compose currently works really well in development, but we want to make the Compose file format better for test, staging, and production environments. To support these use cases, there will need to be improvements to the file format, improvements to the command-line tool, integrations with other tools, and perhaps new tools altogether.
+
+Some specific things we are considering:
 
 - Compose currently will attempt to get your application into the correct state when running `up`, but it has a number of shortcomings:
   - It should roll back to a known good state if it fails.
   - It should allow a user to check the actions it is about to perform before running them.
-- It should be possible to partially modify the config file for different environments (dev/test/staging/prod), passing in e.g. custom ports or volume mount paths. ([#1377](https://github.com/docker/compose/issues/1377))
+- It should be possible to partially modify the config file for different environments (dev/test/staging/prod), passing in e.g. custom ports, volume mount paths, or volume drivers. ([#1377](https://github.com/docker/compose/issues/1377))
 - Compose should recommend a technique for zero-downtime deploys.
 - It should be possible to continuously attempt to keep an application in the correct state, instead of just performing `up` a single time.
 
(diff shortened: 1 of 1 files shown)


ROADMAP.md
```diff
@@ -1,8 +1,19 @@
 # Roadmap
 
+## An even better tool for development environments
+
+Compose is a great tool for development environments, but it could be even better. For example:
+
+- [Compose could watch your code and automatically kick off builds when something changes.](https://github.com/docker/fig/issues/184)
```
dnephin: Could we discuss dropping this from the roadmap? It doesn't feel appropriate for compose. It's trivial to do this with an external tool that watches files and calls build/up when one changes. We could just document that, instead of building it in.

ROADMAP.md
```diff
@@ -1,8 +1,19 @@
...
+- It should be possible to define hostnames for containers which work from the host machine, e.g. “mywebcontainer.local”. This is needed by apps comprising multiple web services which generate links to one another (e.g. a frontend website and a separate admin webapp)
+- There should be a way to delay starting services until dependent services are ready.
```
simonvanderveldt: This is already addressed in #3030 without including it in Compose.

ROADMAP.md
simonvanderveldt: +1 for @dnephin's comment, there are a lot of tools available that provide this functionality so it seems unnecessary to add it to Compose.
Maybe it's a good idea to add a suggestion how to do this to the FAQ?

dnephin: True, wouldn't hurt to provide an example of it somewhere.

ROADMAP.md
bfirsh: already addressed! thanks! :+1:

ROADMAP.md
bfirsh: I _really_ like the idea of this feature, but happy to remove it if it's contentious.

@aanand WDYT?

dnephin: As a feature it's a very nice thing to have, but I don't think it needs to be integrated. I think it's better to document it, and provide a recipe to implement it if you're not already familiar with existing tools. There are a LOT of portability issues with file watching and it's not something we need to deal with ourself. There are already tons of existing tools that can handle this for you.

bfirsh: Removed this for the sake of it not being controversial ;)

bfirsh:
ping @aanand @dnephin

dnephin:
LGTM

aanand:
LGTM

ROADMAP.md
```diff
@@ -1,13 +1,21 @@
...
+Compose currently works really well in development, but we want to make the Compose file format better for test, staging, and production environments. To support these use cases, there will need to be improvements to the file format, improvements to the command-line tool, integrations with other tools, and perhaps new tools altogether.
+
+Some specific things we are considering:
 
 - Compose currently will attempt to get your application into the correct state when running `up`, but it has a number of shortcomings:
   - It should roll back to a known good state if it fails.
   - It should allow a user to check the actions it is about to perform before running them.
-- It should be possible to partially modify the config file for different environments (dev/test/staging/prod), passing in e.g. custom ports or volume mount paths. ([#1377](https://github.com/docker/compose/issues/1377))
+- It should be possible to partially modify the config file for different environments (dev/test/staging/prod), passing in e.g. custom ports, volume mount paths, or volume drivers. ([#1377](https://github.com/docker/compose/issues/1377))
```
andrerom: This issue is closed so it is a bit unclear if this has been solved or not.

For this also option to be able to have several different .env files *(the global one that does not need to be hard coded into `docker-compoe.yml` files)*, so you can have one for prod and one for dev could have been referred to here *(argument to set different .env file would be best option, but current pr/issue to set different cwd kind of solves it also)*.

PR #2635, opened 2016-01-12
Use networking for version 2 Compose files

aanand:
- Remove `--x-networking` and `--x-network-driver`
- There's now no way to set a network driver - this will be added back with the `networks` key

TODO:

- [x] Disable the use of `links` in v2 files

TODO in a separate PR:

- Use network-wide aliasing to make each service reachable at its own name

### DIFF
--- a/compose/cli/command.py
+++ b/compose/cli/command.py
@@ -13,6 +13,7 @@ from requests.exceptions import SSLError
 from . import errors
 from . import verbose_proxy
 from .. import config
+from ..const import API_VERSIONS
 from ..project import Project
 from .docker_client import docker_client
 from .utils import call_silently
@@ -77,7 +78,10 @@ def get_project(base_dir, config_path=None, project_name=None, verbose=False):
     config_details = config.find(base_dir, config_path)
     project_name = get_project_name(config_details.working_dir, project_name)
     config_data = config.load(config_details)
-    api_version = '1.21' if config_data.version < 2 else None
+
+    api_version = os.environ.get(
+        'COMPOSE_API_VERSION',
+        API_VERSIONS[config_data.version])
     client = get_client(verbose=verbose, version=api_version)
 
     return Project.from_config(project_name, config_data, client)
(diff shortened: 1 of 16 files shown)


compose/cli/command.py
```diff
@@ -75,18 +73,14 @@ def get_client(verbose=False, version=None):
     return client
 
 
-def get_project(base_dir, config_path=None, project_name=None, verbose=False,
-                use_networking=False, network_driver=None):
+def get_project(base_dir, config_path=None, project_name=None, verbose=False):
     config_details = config.find(base_dir, config_path)
+    project_name = get_project_name(config_details.working_dir, project_name)
+    config_data = config.load(config_details)
+    api_version = '1.21' if config_data.version < 2 else None
```
dnephin: Since we're planning on supporting this for a little while longer, maybe we should have a mapping of config_version -> api_version instead of the constant `DEFAULT_API_VERSION`.  

That way the default api versions are defined together in the same place.

tests/acceptance/cli_test.py
```diff
@@ -355,13 +355,13 @@ def test_up_without_networking(self):
     def test_up_with_networking(self):
         self.require_api_version('1.21')
```
dnephin: We can remove this now

tests/acceptance/cli_test.py
```diff
@@ -653,11 +653,11 @@ def test_run_with_custom_name(self):
     def test_run_with_networking(self):
         self.require_api_version('1.21')
         client = docker_client(version='1.21')
```
dnephin: We can remove this as well, and use `self.client`

tests/acceptance/cli_test.py
```diff
@@ -336,13 +335,10 @@ def test_up_attached(self):
...
 
     def test_up_without_networking(self):
-        self.require_api_version('1.21')
-
         self.base_dir = 'tests/fixtures/links-composefile'
         self.dispatch(['up', '-d'], None)
-        client = docker_client(version='1.21')
 
-        networks = client.networks(names=[self.project.name])
+        networks = self.client.networks(names=[self.project.name])
```
dnephin: I think this needs to use `self.project.default_network_name` now

aanand: Good catch, fixed.

dnephin:
LGTM!

shin-:
LGTM :+1:

PR #2638, opened 2016-01-12
 Add bash completion for `docker-compose events`

albers:
This adds bash completion for #2392.

I'm not sure what services I should suggest for that command. I opted for all services. Should I use only running ones instead?


dnephin:
All services is correct.

LGTM

Thanks!

aanand:
LGTM

PR #2639, opened 2016-01-12
Fix pep8 errors from the new pep8 release.

dnephin:


shin-:
LGTM!

dnephin:
The janky build is past the pre-commit step, so I'm going to merge this to unblock people

PR #2646, opened 2016-01-12
docker-compose down

### LINKED ISSUE #1874: Combined command to kill and remove containers

### LINKED ISSUE #2277: Proposal: docker-compose down (opposite of up)
I ran `docker-compose up` to start my instance. It'd be nice (and a consistent API) if I could also use `docker-compose down` to turn it off. It could even be a silent alias for the relevant command.

### LINKED ISSUE #2279: 1.5.0-rc2: Compose does not remove the network during rm
```
$ docker-compose --x-networking up -d
Creating network "counter" with driver "bridge"
Creating counter_web_1
Creating counter_mongo_1

$ docker-compose --x-networking stop
Stopping counter_mongo_1 ... done
Stopping counter_web_1 ... done

$ docker-compose --x-networking rm
Going to remove counter_mongo_1, counter_web_1
Are you sure? [yN] y
Removing counter_mongo_1 ... done
Removing counter_web_1 ... done

$ docker network ls
NETWORK ID          NAME                DRIVER
d84962a207f8        counter             bridge
f60408bc3b3b        bridge              bridge
13a612e10bd7        none                null
517effa88694        host                host
```

After, `rm` I still have a network created with the project name
I would expect this to be removed

NOTE: This also causes an issue with Swarm that I will raise another issue for

dnephin:
Fixes #2279
Fixes #2277
Fixes #1874

Implements a new `down` command which should be capable of stopping and removing all resources created by `up` (with the correct flags).

I'm not sure how this is going to work with networks, if the network happens to have other containers on it. Does the remove fail?

I think the behaviour of down should be that it logs an error and continues if one of the resources isn't removed, and reports a proper exit code if anything failed.  That isn't implemented yet, but I wanted to get this out before code freeze.

### DIFF
--- a/compose/project.py
+++ b/compose/project.py
@@ -272,7 +272,7 @@ class Project(object):
 
     def down(self, remove_image_type, include_volumes):
         self.stop()
-        self.remove_stopped()
+        self.remove_stopped(v=include_volumes)
         self.remove_network()
 
         if include_volumes:
@@ -441,6 +441,7 @@ class Project(object):
             return
         network = self.get_network()
         if network:
+            log.info("Removing network %s", self.default_network_name)
             self.client.remove_network(network['Id'])
 
     def uses_default_network(self):
--- a/compose/project.py
+++ b/compose/project.py
@@ -270,6 +270,24 @@ class Project(object):
                     )
                 )
 
+    def down(self, remove_image_type, include_volumes):
+        self.stop()
+        self.remove_stopped()
+        self.remove_network()
+
+        if include_volumes:
+            self.remove_volumes()
+
+        self.remove_images(remove_image_type)
+
+    def remove_images(self, remove_image_type):
+        for service in self.get_services():
+            service.remove_image(remove_image_type)
+
+    def remove_volumes(self):
+        for volume in self.volumes:
+            volume.remove()
+
     def restart(self, service_names=None, **options):
         containers = self.containers(service_names, stopped=True)
         parallel.parallel_restart(containers, options)
@@ -419,6 +437,8 @@ class Project(object):
             self.client.create_network(self.default_network_name, driver=self.network_driver)
 
     def remove_network(self):
+        if not self.use_networking:
+            return
         network = self.get_network()
         if network:
             self.client.remove_network(network['Id'])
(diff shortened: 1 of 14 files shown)


blah238:
Cool, I was just looking for this command, surprised it didn't already exist! :+1:

compose/project.py
```diff
@@ -253,6 +253,24 @@ def initialize_volumes(self):
...
+
+        self.remove_images(remove_image_type)
+
+    def remove_images(self, remove_image_type):
+        for service in self.get_services():
+            service.remove_image(remove_image_type)
+
+    def remove_volumes(self):
+        for volume in self.volumes:
+            volume.remove()
```
dnephin: TODO: never remove external volumes

compose/project.py
```diff
@@ -253,6 +253,24 @@ def initialize_volumes(self):
                     )
                 )
 
+    def down(self, remove_image_type, include_volumes):
+        self.stop()
+        self.remove_stopped(v=include_volumes)
+        self.remove_network()
```
dnephin: TODO: this needs to be graceful if the remove fails

aanand:
Are you planning to finish the TODOs in this PR or separately? In any case, LGTM after rebase.

dnephin:
Rebased and made it so that external volumes are never removed. I also fixed the help text to say that only containers and networks are removed by default.

I'm going to leave the "handle network remove failures gracefully" TODO until after networks are fully implemented, I expect behaviour may change slightly.

dnephin:
Rebased again, going to merge

albers:
@dnephin `docker-compose up` supports services as arguments. Does this make sense for `docker-compose down` as well?

dnephin:
I considered that, but I don't think so.  `down` is meant to teardown everything (ex: networks, volumes) which aren't specific to a single service, and will fail if they happen before all the services are stopped and removed.

Instead of trying to handle all those edge cases I opted for not allowing service names. If someone finds a really critical case for it, we can always add it later.

albers:
@dnephin Makes sense. I stumbled over this only because of symmetry considerations when working on the completion.

PR #2647, opened 2016-01-13
Support for external volumes

shin-:
See https://github.com/docker/compose/issues/2502#issuecomment-170112554

### DIFF
--- a/compose/config/config.py
+++ b/compose/config/config.py
@@ -273,6 +273,13 @@ def load_volumes(config_files):
     for config_file in config_files:
         for name, volume_config in config_file.config.get('volumes', {}).items():
             volumes.update({name: volume_config})
+            external = volume_config.get('external')
+            if external:
+                if isinstance(external, dict):
+                    volume_config['external_name'] = external.get('name')
+                else:
+                    volume_config['external_name'] = name
+
     return volumes
 
 
--- a/compose/config/fields_schema_v2.json
+++ b/compose/config/fields_schema_v2.json
@@ -41,6 +41,13 @@
             "^.+$": {"type": ["string", "number"]}
           },
           "additionalProperties": false
+        },
+        "external": {
+          "type": ["boolean", "object"],
+          "properties": {
+            "name": {"type": "string"}
+          },
+          "additionalProperties": false
         }
       }
     }
--- a/compose/config/fields_schema_v2.json
+++ b/compose/config/fields_schema_v2.json
@@ -32,24 +32,32 @@
   "definitions": {
     "volume": {
       "id": "#/definitions/volume",
-      "type": "object",
-      "properties": {
-        "driver": {"type": "string"},
-        "driver_opts": {
-          "type": "object",
-          "patternProperties": {
-            "^.+$": {"type": ["string", "number"]}
-          },
-          "additionalProperties": false
+      "oneOf": [{
+        "type": "object",
+        "properties": {
+          "driver": {"type": "string"},
+          "driver_opts": {
+            "type": "object",
+            "patternProperties": {
+              "^.+$": {"type": ["string", "number"]}
+            },
+            "additionalProperties": false
+          }
         },
-        "external": {
-          "type": ["boolean", "object"],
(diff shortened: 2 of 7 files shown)


compose/volume.py
```diff
@@ -1,14 +1,23 @@
...
         self.project = project
         self.name = name
         self.driver = driver
         self.driver_opts = driver_opts
+        self.external_name = None
+        if external:
+            if isinstance(external, dict):
+                self.external_name = external.get('name')
+            else:
+                self.external_name = self.name
```
dnephin: I wonder if we should do this normalization in the config module, since we do the other related operations there.

dnephin:
LGTM

compose/project.py
```diff
@@ -235,6 +237,18 @@ def remove_stopped(self, service_names=None, **options):
     def initialize_volumes(self):
         try:
             for volume in self.volumes:
+                if volume.external:
+                    log.info(
+                        'Volume {0} declared as external. No new '
+                        'volume will be created.'.format(volume.name)
+                    )
```
aanand: Small thing - I think this could be demoted to a `log.debug`.

compose/project.py
```diff
@@ -235,6 +237,18 @@ def remove_stopped(self, service_names=None, **options):
...
+                    if not volume.exists():
+                        raise ConfigurationError(
+                            'Volume {0} declared as external, but could not be'
+                            ' found. Please create the volume manually and try'
+                            ' again.'.format(volume.full_name)
```
aanand: We could suggest the command to use here: `docker volume create --name={name}`

compose/volume.py
aanand: Agreed, that'd be good.

shin-: "External name" is not really a part of the configuration though, it's mostly an internal attribute. I'd be weary of polluting the config with it unless it's needed earlier in the process.

dnephin: Sure it is, it's in the config as `{'external': {'name': ...}}`

compose/config/fields_schema_v2.json
```diff
@@ -41,6 +41,13 @@
             "^.+$": {"type": ["string", "number"]}
           },
           "additionalProperties": false
+        },
+        "external": {
+          "type": ["boolean", "object"],
+          "properties": {
+            "name": {"type": "string"}
+          },
+          "additionalProperties": false
```
dnephin: Oh, I just realized, I think we're missing a constraint. We want external and driver/driver_opts to be mutually exclusive, right ?

I think you can set a constraint that if external is used, driver and driver_opts are invalid.

shin-: Well, yeah, the 2 other options won't really do anything if `external` is specified. I'll see if JSON schema has a thing for that.

compose/config/config.py
```diff
@@ -273,6 +273,13 @@ def load_volumes(config_files):
     for config_file in config_files:
         for name, volume_config in config_file.config.get('volumes', {}).items():
             volumes.update({name: volume_config})
+            external = volume_config.get('external')
+            if external:
+                if isinstance(external, dict):
+                    volume_config['external_name'] = external.get('name')
+                else:
+                    volume_config['external_name'] = name
```
aanand: This looks like it will set `external_name` to `None` if I type:

```yaml
vol:
  external: {}
```

First branch should probably read `volume_config['external_name'] = external.get('name', name)`

compose/config/fields_schema_v2.json
dnephin: We had a version of it with build vs image that you can look at (it's gone now, but it was in 1.5.2).  I'm not sure if there is any other way, other than using  oneOf: `requires`, `not: requires`

compose/config/config.py
shin-: `external` would evaluate to `False` in that case.

shin-:
@dnephin Should look good now.

dnephin:
LGTM!

Issue #2501, opened 2015-12-04
Proposal: Make `net` and `volumes_from`'s container/service support explicit


aanand:
Right now, you can put both service and container names in `volumes_from`:

```yaml
db:
  image: postgres
  volumes_from: ["data", "something-else"]

data:
  image: tianon/true
  volumes: ["/var/lib/postgresql/data"]
```

Compose will interpret the `data` entry in `volumes_from` as referring to the `data` service, and the `something-else` entry as referring to an already-created container.

The same is true of `net`: you can type `net: "container:<container-name>"` or `net: "container:<service-name>"`.

This is error-prone: if you mistype a service name, you'll get an error saying that no such container exists, which is misleading. Furthermore, it makes it harder to tell what's going on for someone who's not aware that you can mix container and service names.

We should make the distinction explicit.

### net

In the case of `net`, I think `net: "container:<service-name>"` should become `net: "service:<service-name>"`. Both uses of `net` are uncommon, so we needn't make them especially concise.

### volumes_from

In the case of `volumes_from`, it's less obvious what to do:

1. We could rule that container and service names must be prefixed with `container:` or `service:`.

2. We could rule that container names must be prefixed with `container:`, and un-prefixed names are assumed to be service names. This is probably how most people use `volumes_from`, so it makes sense to make the default case more concise.

3. We could rule the opposite way: service names must be prefixed with `service:`, and un-prefixed names are assumed to be container names. This means that the container name case maps to the Docker CLI's `--volumes-from=container-name` flag, but also means that most users will have to do more typing.

I personally like option 2 best.

aanand:
One other solution I forgot to mention: similar to how we have two separate config keys - `links` and `external_links` - for defining links to locally-defined services versus externally-created containers, we could do something similar and have `volumes_from` and `external_volumes_from`.

dnephin:
I also like option 2

aanand:
This can now be closed:

- #2649 sorted out `volumes_from`

- `--net container:...` makes little sense in a post-networking (Compose file v2) world, and changing it for pre-networking (Compose file v1) would break backwards compatibility, so there's nothing to do here.

funkytaco:
Can somebody explain the documentation on this?

https://docs.docker.com/compose/compose-file/ says:

>net
>Networking mode. Use the same values as the docker client --net parameter.

    net: "bridge"
    net: "none"
    net: "container:[name or id]"
    net: "host"

[...]

PR #2649, opened 2016-01-13
Make volumes from more explicit in V2 config format

dnephin:
Fixes the "volumes_from" section of #2501 

Includes some refactoring of version checking.  `version` becomes a cached_property of a `ConfigFile`. That way we don't have to pass it around along with the config.

### DIFF
--- a/compose/config/config.py
+++ b/compose/config/config.py
@@ -10,6 +10,7 @@ from collections import namedtuple
 
 import six
 import yaml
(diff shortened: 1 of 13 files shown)


dnephin:
Fixed that bug, adding a couple missing test cases for parsing volumes_from, and rebased on top of the latest external volumes changes.

aanand:
LGTM

PR #2650, opened 2016-01-13
Interpolate variables in the new sections of the Compose file

dnephin:
Adds interpolation support for `volumes` and `networks` sections of the compose file.

Also adds `ConfigFile.get_networks()` and uses it as part of `load_mapping()` to avoid the rare bug where someone might use the name `volumes` or `networks` as a service name in a V1 config.


dnephin:
actually, this fails when using a blank volume config, so I think it needs another test

dnephin:
I'll wait for networks to be merged and add them to this PR as well

dnephin:
Rebased and added support for networks.

aanand:
LGTM

PR #2653, opened 2016-01-14
Support for build arguments

shin-:
Based on #2540 

Resolved merge conflicts + small change to move legacy format handling in the appropriate method.

### DIFF
--- a/compose/config/config.py
+++ b/compose/config/config.py
@@ -562,6 +562,12 @@ def normalize_v1_service_format(service_dict):
             service_dict['logging']['options'] = service_dict['log_opt']
             del service_dict['log_opt']
 
+    if 'dockerfile' in service_dict:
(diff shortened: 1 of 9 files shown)


dnephin:
Looks like there is a test failure

compose/config/config.py
```diff
@@ -690,6 +749,26 @@ def resolve_volume_path(working_dir, volume):
         return container_path
 
 
+def normalize_build(service_dict, working_dir):
+    build = {}
+
+    # supported in V1 only
+    if 'dockerfile' in service_dict:
+        build['dockerfile'] = service_dict.pop('dockerfile')
```
dnephin: I think we had deleted this in a previous rebase, right? Now that it's in the V1 function

aanand:
LGTM

PR #2654, opened 2016-01-14
Fix flaky multiplex test

dnephin:
We just added this test in #2130, and it's already failed once:
https://ci.appveyor.com/project/dnephin/compose/build/make_volumes_from_more_explicit-366#L764

I forgot that I wanted to change it before we merged it, so doing it now. Instead of relying on precise timing and scheduling of threads, make one really fast (no sleep) stream, and one really slow (long sleep) stream.  

I believe this gives us the same test coverage without sleeping or working about it flaking.


aanand:
LGTM

PR #2655, opened 2016-01-14
Properly validate volume definition

shin-:
+ Test valid empty volume definitions


dnephin:
Works for me. 

Maybe a test case for this new error case?

Do you want to update the two `driver: default` references in the docs in this PR as well?

shin-:
There should be one already that I included in the external volumes PR.

dnephin:
Ah, right, there is, cool

aanand:
LGTM

PR #2658, opened 2016-01-14
Error when the project name is invalid for the default image name

dnephin:
Users will get a more appropriate error message now, instead of one about invalid image name.


aanand:
LGTM

PR #2659, opened 2016-01-14
Containers join each network aliased to their service's name

aanand:
TODO:

- [x] Run against a dev build of Engine
- [x] Tag v2-only tests and don't run them against Engine 1.9


dnephin:
LGTM!

PR #2662, opened 2016-01-14
Fix config printing unicode objects

### LINKED ISSUE #2644: config` includes python objects instead of string
```PS C:\Users\Administrator\test> docker-compose version
docker-compose version 1.6.0dev, build 304a44a
docker-py version: 1.6.0
CPython version: 2.7.11
OpenSSL version: OpenSSL 1.0.2d 9 Jul 2015```

Given the docker-compose.yml:
```my_nginx:
   build: .
   ports: 
     - 8080:80
   volumes:
     - ".\testdir:\foo"
```
docker-compose config

```
PS C:\Users\Administrator\test> docker-compose config
my_nginx:
  build: !!python/unicode 'C:\Users\Administrator\test'
  ports:
  - 8080:80
  volumes:
  - !!python/object/new:compose.config.types.VolumeSpec
    - !!python/unicode '/c/Users/Administrator/test/testdir'
    - !!python/unicode '/foo'
    - !!python/unicode 'rw'
```

dnephin:
Fixes #2644 

Also fixes:
* `down` now removes all networks, except for external networks
* `config` prints volumes/networks/version

Currently `config` will print a v2 config regardless of the input.  If we keep it this way, I think it could be used as a way for people to automatically convert their v1 config to v2 configs.

Trying to have it print a v1 config might be a challenge, because we normalize to the V2 format, so we'd have to undo all of those operations. I'd rather not do that.


dnephin:
Actually, I guess it doesn't exactly work as a migration script because it resolves variables/extends/paths, so it might actually be more work to convert those back.

shin-:
LGTM

PR #2665, opened 2016-01-15
Validate that an extended config file has the same version as the base

### LINKED ISSUE #2623: Check version consistency for extends directives
See: https://github.com/docker/compose/pull/2421#discussion_r47576562

dnephin:
Fixes #2623

In order to compare files we need access to the `ConfigFile` object, so I swapped out some version and filename parameters for the full `ConfigFile` instead, then call our existing function which compares the versions from a list of `ConfigFile` objects.


shin-:
LGTM :+1:

PR #1686, opened 2015-07-13
.partition() is cleaner than .split() for getting both parts

cclauss:
Signed-off-by: Chris Clauss <cclauss@bluewin.ch>

PR #2669, opened 2016-01-15
.partition() is cleaner than .split() for getting both parts

cclauss:
Signed-off-by: Chris Clauss cclauss@bluewin.ch

Attempts to redo some of https://github.com/docker/compose/pull/1686

### DIFF
--- a/compose/project.py
+++ b/compose/project.py
@@ -151,9 +151,7 @@ class Project(object):
 
         Raises NoSuchService if any of the named services do not exist.
         """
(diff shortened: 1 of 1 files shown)


cclauss:
I will have to reread singing and try again.  Closing for now.

dnephin:
I think the signing here is ok, the bot might not be able to find it because it's not the last line in the commit. If you move the comment to above the sign-off, I think it will work.

PR #2671, opened 2016-01-15
Fix linux master build

dnephin:
We don't include `git` in the `Dockerfile.run`, and it's needed to install dependencies. Since we're not actually pushing this image to the registry there's no reason to build it here.

This should be fixed by the time we release 1.6 because we'll have removed the git url from requirements.txt.

https://travis-ci.org/dnephin/compose/builds/102679653


shin-:
LGTM!

PR #2672, opened 2016-01-15
Organize scripts into directories

dnephin:
We've got a lot of scripts now, so I've tried to organize them into `build`, `test`, `run`, and `setup`.

We already had `travis` and `release`.


dnephin:
rebased and green!

aanand:
LGTM

Issue #2619, opened 2016-01-07
Dependency-ordered stop


IlyaSkriblovsky:
Hi!

Is it possible to acheive ordered stopping with Compose? My app needs to write something to DB when stopping gracefully.

@aanand said in #1395:

> Container stopping/killing should be completely order-independent - if it's all going down, it doesn't matter what order we do it in.

While this is true for almost any regular web app, I can't agree with that for something more stateful. For example, when my service receives stop signal, it gracefully closes all active connections and saves log records on each disconnection to DB. I'm trying to dockerize my app with Compose and this seems to be an issue :(  Thanks for advice.

For example, both SysV and Systemd have means for dependency-ordering both starting and stopping of services. I bet you won't like if your PC will unmount filesystems before all your apps is closed while shutting down :)

andrenarchy:
+1 for stopping processes in the reverse order of starting them. btw: this is also how [maestro-ng](https://github.com/signalfx/maestro-ng) handles it.

Concerning the quote of @aanand: I disagree because sending `SIGTERM` doesn't mean that everything is going down chaotically but actually gives the process a way to shut down gracefully.

IlyaSkriblovsky:
Any comments from maintainers?

If the general idea of dependency-ordered service stopping (using `links` and `volumes_from` as implicit dependency specifications) will be approved by project maintainers, I will try to implement it and create a pull request.

I've read almost all discussions here about introducing `depends_on` or other possible explicit dependency syntax, but failed to understand whether any of such ideas are approved and planned for implementation? Will Compose have any means of inter-service dependency tracking when linking will be eventually removed from Docker? Using `volumes_from` for stating dependencies looks like a dirty hack.

dnephin:
If we add a new `depends_on` key to the upcoming V2 config format (#2504) then it would make sense to allow it to be used for shutdown as well as startup.

I don't think we've made any decision here yet, but it does seem to be a highly requested feature, so it's very likely we will provide something.

I don't think we're ready for a PR just yet. If we do add it, it will likely only be for the V2 format, which isn't ready just yet. We also need to decide exactly what the scope of `depends_on` would be.

IlyaSkriblovsky:
The ordering by `links` and `volumes_from` can be implemented right now without introducing any additional syntax. Compose already does ordering while starting up, I just suggest to add same ordering to shutting down.

[...]

PR #2679, opened 2016-01-16
Parallel up/start/stop

IlyaSkriblovsky:
Starting and stopping containers in dependency-ordered manner as proposed in #2619

- Starting is currently done sequentially in dependency order. This change makes it parallel while obeying dependencies. It may speed up starting by around 30% in my tests.
- Stopping is currently done all-parallel, regardless of dependencies. This change make stopping obey dependencies. This may slower shutting down, but makes it possible to stop processes gracefully .
- Killing behavior is not changed

Seems like `dep_ordered_op` is a superset of `parallel_op` (with `get_deps=lambda x: []`). Should I merge them together?

Ugly `silent` flag was used because `Service` does `log.info` on each started container and I don't want to interfere.

### DIFF
--- a/compose/parallel.py
+++ b/compose/parallel.py
@@ -155,10 +155,6 @@ def parallel_remove(containers, options):
     parallel_operation(stopped_containers, 'remove', options, 'Removing')
 
 
-def parallel_stop(containers, options):
-    parallel_operation(containers, 'stop', options, 'Stopping')
-
-
 def parallel_start(containers, options):
     parallel_operation(containers, 'start', options, 'Starting')
 
--- a/compose/parallel.py
+++ b/compose/parallel.py
@@ -14,68 +14,98 @@ from compose.cli.signals import ShutdownException
 from compose.utils import get_output_stream
 
 
-def perform_operation(func, arg, callback, index):
-    try:
-        callback((index, func(arg)))
-    except Exception as e:
-        callback((index, e))
+def parallel_execute(objects, func, get_name, msg, get_deps=None):
+    """Runs func on objects in parallel while ensuring that func is
+    ran on object only after it is ran on all its dependencies.
 
-
-def parallel_execute(objects, func, index_func, msg):
-    """For a given list of objects, call the callable passing in the first
-    object we give it.
+    get_deps called on object must return a collection with its dependencies.
+    get_name called on object must return its name.
     """
     objects = list(objects)
     stream = get_output_stream(sys.stderr)
+
     writer = ParallelStreamWriter(stream, msg)
-    q = setup_queue(writer, objects, func, index_func)
+    for obj in objects:
+        writer.initialize(get_name(obj))
+
+    q = setup_queue(objects, func, get_deps, get_name)
 
     done = 0
     errors = {}
+    error_to_reraise = None
+    returned = [None] * len(objects)
 
     while done < len(objects):
         try:
-            msg_index, result = q.get(timeout=1)
+            obj, result, exception = q.get(timeout=1)
         except Empty:
             continue
         # See https://github.com/docker/compose/issues/189
         except thread.error:
             raise ShutdownException()
 
-        if isinstance(result, APIError):
-            errors[msg_index] = "error", result.explanation
-            writer.write(msg_index, 'error')
-        elif isinstance(result, Exception):
-            errors[msg_index] = "unexpected_exception", result
+        if exception is None:
+            writer.write(get_name(obj), 'done')
+            returned[objects.index(obj)] = result
+        elif isinstance(exception, APIError):
+            errors[get_name(obj)] = exception.explanation
+            writer.write(get_name(obj), 'error')
         else:
-            writer.write(msg_index, 'done')
+            errors[get_name(obj)] = exception
+            error_to_reraise = exception
+
         done += 1
 
-    if not errors:
-        return
+    for obj_name, error in errors.items():
+        stream.write("\nERROR: for {}  {}\n".format(obj_name, error))
 
-    stream.write("\n")
-    for msg_index, (result, error) in errors.items():
-        stream.write("ERROR: for {}  {} \n".format(msg_index, error))
-        if result == 'unexpected_exception':
-            raise error
+    if error_to_reraise:
+        raise error_to_reraise
+
+    return returned
 
 
-def setup_queue(writer, objects, func, index_func):
-    for obj in objects:
-        writer.initialize(index_func(obj))
+def _no_deps(x):
+    return []
 
-    q = Queue()
 
-    # TODO: limit the number of threads #1828
-    for obj in objects:
-        t = Thread(
-            target=perform_operation,
-            args=(func, obj, q.put, index_func(obj)))
-        t.daemon = True
-        t.start()
+def setup_queue(objects, func, get_deps, get_name):
+    if get_deps is None:
+        get_deps = _no_deps
 
-    return q
+    results = Queue()
+
+    started = set()   # objects, threads were started for
+    finished = set()  # already finished objects
+
+    def do_op(obj):
+        try:
+            result = func(obj)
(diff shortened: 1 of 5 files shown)


dnephin:
Impressive first draft, some high level comments:
* I think we should merge `dep_ordered_op` and `parallel_op` (as you suggest)
* `project.start()` is not actually what's called on `up`, there is a project.up()`. We might want to apply this to `up()` as well since that's what is most commonly used. I think start is rarely used.

IlyaSkriblovsky:
Thanks, forgot about `up`.

Am I correct, that test failures are not caused by my changes?

dnephin:
The `test_down` failure could be related if you changed the stdout of `project.stop()`, which I believe you've done.

IlyaSkriblovsky:
Ah, I've tested on Docker 1.9 on my machine and this test was skipped. Will fix that.

IlyaSkriblovsky:
I've tried to modify `up` to make it build, create and recreate containers in sequence, but to start them in parallel. But it turns out that I can't divide starting containers from creating them: for example if service `B` uses `container:A` network, creating `B` requires `A` to be already running.

So, I see two options:
- parallelize `up` by making it do all its work (build, create and start) in parallel (with dependency ordering, of course). But in that way the console output of building images will mix up.
- make it build when needs to be built in sequence and then create, recreate and start containers in parallel.

Do you agree that second option is the right way?

aanand:
Yes - I think all the building/pulling should happen first, and then (re)create+start can be as parallelised as possible.

IlyaSkriblovsky:
It seems like `test.acceptance.cli_test.py::CLITestCase::test_up_handles_sigint`/`sigterm` has a race condition. They start `up`, wait until there are expected number of containers started and then do `os.kill`. But `TopLevelCommand.up()` sets signal handlers only *after* it starts all containers (in `attach_to_logs()`). So, test method might send signal after containers is started, but before handlers is attached, and Compose process will be killed without proper shutdown, leaving containers running.

This issue is outside of this PR and not caused by my changes. But seems like my changes affected some timings and this issue now appears with greater probability than before (it appeared only in py27 in last jenkins build, but not in py34, for example).

I added a workaround just to make tests to pass by adding `, delay=1)` to `wait_on_condition` calls to make them send signals not earlier than after 1 second. This might be removed from final code.

dnephin:
I'll work on a fix for that issue

IlyaSkriblovsky:
What about the rest of PR? :)

dnephin:
Oops, I guess I had a "fixes (comment)" in another PR, and I guess that caused github to close this issue.

The test race condition should be fixed on master now

IlyaSkriblovsky:
Removed my workaround for #2784 and rebased on master

IlyaSkriblovsky:
rebased on master

tests/acceptance/cli_test.py
```diff
@@ -1227,7 +1228,7 @@ def test_events_json(self):
         os.kill(events_proc.pid, signal.SIGINT)
         result = wait_on_process(events_proc, returncode=1)
         lines = [json.loads(line) for line in result.stdout.rstrip().split('\n')]
-        assert [e['action'] for e in lines] == ['create', 'start', 'create', 'start']
+        self.assertEqual(Counter(e['action'] for e in lines), {'create': 2, 'start': 2})
```
dnephin: We prefer `assert` over `self.assertEqual`. `pytest` takes care of formatting the error properly, and generally does a better job than unittest at formatting it, and displaying relevant information.

compose/parallel.py
```diff
@@ -133,3 +92,86 @@ def parallel_kill(containers, options):
...
+    """Runs operation on objects in parallel while ensuring that operation is
+    ran on object only after it is ran on all its dependencies.
+
+    get_deps called on object must return a collection with its dependencies.
+    get_name called on object must return its name.
+    Setting silent to True disables any console output.
+    """
+    objects = list(objects)
+    stream = get_output_stream(sys.stderr)
+    writer_class = DummyParallelStreamWriter if silent else ParallelStreamWriter
```
dnephin: I don't think we need to worry about being silent. We can have start and up print a similar message. If they already have logging somewhere else we can move it here.

dnephin:
I think  the parallel function has changed a bit since this PR was first prepared. Some logic was extracted into a separate function to reduce the overall complexity, and I think the error handling was simplified a bit.

I think it would be good if we could keep as much as the existing implementation as possible, and just add a few args for handling dependencies. The big difference with this implementation is the `feed()` function, right? Otherwise it's approximately the same?

IlyaSkriblovsky:
Thanks for feedback. I will rebase this and try to extract queue processing from `dep_ordered_op` just like it was done to `parallel_op`. Lets see if it will be possible to share other parts of the function.

IlyaSkriblovsky:
Squashed and rebased onto master.

1. I've refactored my `dep_ordered_op` to be compatible with `parallel_execute` and to inherit its structure and some code parts.

2. New `parallel_execute` still handles silent mode (enabled by `msg=None`). This is to make it silent for `up` because output of `up` differs from output of other commands. Each line of `up` output may have different operation name (Starting/Creating/Recreating/Up-to-date). This is actually not a problem: we might replace `msg` argument of `parallel_execute` with function returning operation name for each object (service). But anyways, it will require removing logging from multiple methods of `Service` class: `execute_convergence_plan`, `create_container`, `recreate_container`, etc, or passing some sort of silence flag via arguments. Does it worth it? Please advice.

3. #2804 that was merged today adds test case which tests that `up` raises `APIError` on some broken config. This testcase is currently failing on this branch, because `up` was redone by using `parallel_execute` which consumes `APIError` and just logs them to console. What is the correct expected behavior of `up`? To reraise `APIError` like this testcase expects or to consume `APIError` just like other methods (`stop`/`restart`/`pause`) that was implemented by `parallel_execute`?

dnephin:
For 2 I agree about the logging with `up`. The way you have it now seems correct.

For 3, I think the test case will have to change. We can probably test that behaviour in a better way.

IlyaSkriblovsky:
Fixed test by checking `len(up()) == 0` (`up()` returns a list of successfully started containers)

IlyaSkriblovsky:
Weird, `tests/integration/project_test.py::ProjectTest::test_up_with_network_static_addresses` seems to pass well on my machine...

IlyaSkriblovsky:
I made temporary commit that makes `up` not to swallow APIErrors. Hope this will help to understand why `tests/integration/project_test.py::ProjectTest::test_up_with_network_static_addresses` is failing. Will reset this commit after that.

IlyaSkriblovsky:
Ok, I played a bit with failing `test_up_with_network_static_addresses` test, then reverted temporary commits back and now it suddenly passes :-/  So, it seems this test is somewhat unstable and this is not related to this PR.

So, can I do anything else for this PR?

aanand:
I've checked it out and it works nicely. I think the only issue I have is with output - we need to do some refactoring so that the output of `up` ("Recreating X", "Y is up-to-date" etc) is handled in the same way as that of `start`, `stop`, `kill` etc. Otherwise we could easily get jumbled output. But we can do that separately.

LGTM

compose/project.py
```diff
@@ -200,13 +201,40 @@ def get_network_mode(self, service_dict, networks):
...
+            services,
+            start_service,
+            operator.attrgetter('name'),
+            'Starting',
+            get_deps)
+
         return containers
 
     def stop(self, service_names=None, **options):
-        parallel.parallel_stop(self.containers(service_names), options)
```
dnephin: I think `parallel_stop()` can be deleted now. I don't think it's used anywhere else

dnephin:
LGTM, just one minor request to remove the old unused function

IlyaSkriblovsky:
Removed unused `parallel_stop` and rebased to master

PR #2688, opened 2016-01-18
POC: Run the full test suite on travisci

dnephin:
This runs the full test suite on travisci. The test suite finishes a lot faster than janky most of the time.

https://travis-ci.org/dnephin/compose/builds/103157899

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,10 +5,24 @@ language: python
 matrix:
   include:
     - os: linux
-      services:
-      - docker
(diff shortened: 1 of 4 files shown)


dnephin:
I think we can improve the build times quite a bit by fixing our Dockerfile to not build everything from source.  If we need a specific version of python we can build a deb package with the right version.

dnephin:
Closing this since it never seemed to be any more reliable than jenkins.

PR #2690, opened 2016-01-18
Allow custom ipam config

aanand:


dnephin:
LGTM

tests/acceptance/cli_test.py
```diff
@@ -485,6 +485,36 @@ def test_up_missing_network(self):
         assert 'Service "web" uses an undefined network "foo"' in result.stderr
 
     @v2_only()
+    def test_up_with_ipam_config(self):
```
dnephin: This seems like it might be better suited as an integration test instead of an end-to-end test. We're not really testing a compose feature here.

aanand: Sure - moved it to a `project.py` integration test.

dnephin:
LGTM

PR #2692, opened 2016-01-18
Catch TLSParameterErrors from docker-py

aanand:


shin-:
LGTM!

compose/cli/main.py
```diff
@@ -65,6 +66,11 @@ def main():
     except APIError as e:
         log.error(e.explanation)
         sys.exit(1)
+    except TLSParameterError:
+        log.error(
+            'TLS configuration is invalid - make sure your DOCKER_TLS_VERIFY and DOCKER_CERT_PATH are set correctly.\n'
+            'You might need to run `eval "$(docker-machine env default)"`')
+        sys.exit(1)
```
dnephin: This error is really specific to one section of the code, I don't think we need it in the general catch-all error handling.

Could we move it to `docker_client.py` and re-raise a `UserError` with this message?

Arguably a few of these other errors should be moved to more specific areas as well.

aanand: Good point - updated.

Some other errors might be moveable - the only ones that need to be here are those which don't happen until we make an API request, I suppose.

dnephin:
LGTM

aanand:
Cherry-picked into #2753

PR #2693, opened 2016-01-18
Add a migration script for v1 to v2 compose format

### LINKED ISSUE #2678: Provide a migration command or script for v1 to v2 Compose format
We should provide a migration script or command for converting from the v1 format to the v2 format. We'll need to use a "round trip" yaml parser like https://pypi.python.org/pypi/ruamel.yaml

I don't think we should change our yaml parser entirely, we can just use this library for the conversion and keep pyyaml for everything else.

I think we can do something like this:
 1. run the v1 config through `docker-compose config -q` to make sure it's valid
 2. run the convertion
 3. run the converted v2 config through `docker-compose config -q` again to make sure it's still valid. 

We could even compare the output from 1 and 3, and they should be the same.

Doing it as a script would be easier (we don't have to go adding ruamel and all it's dependencies, which will increase the size of the binary for everyone), but a new command `migrate-to-v2` is possible.

dnephin:
Fixes #2678

`ruamel.yaml` does an ok job of preserving header comments, formatting, aliases, and anchors.

But it is still pretty limited:
 * inline comments may still be lost
 * comments above lines seem to always be lost
 * in some cases (it seems to be when using indent=4) an end-of-line comment can be added with the wrong amount of space (none), confusing the pyyaml parser
* the order of services and service options changes
* quotes around most string scalars are lost

I don't know of any round trip yaml parser that is any better (in python or any other language). I doubt the quality of this script is ever really going to be good enough to suggest it as the primary way to migrate. Since the quality isn't great, I don't think we should be including it as a compose command.

I think we should provide this script as a docker image, and suggest it as an alternative option in the migration doc, for users with very large compose files.

I think for most compose files the user will be better off just using their editor to make the necessary changes.

### DIFF
--- a/contrib/migration/migrate-compose-file-v1-to-v2.py
+++ b/contrib/migration/migrate-compose-file-v1-to-v2.py
@@ -1,0 +1,99 @@
+#!/usr/bin/env python
+"""
+Migrate a Compose file from the V1 format in Compose 1.5 to the V2 format
+supported by Compose 1.6+
+"""
+from __future__ import absolute_import
+from __future__ import unicode_literals
+
+import argparse
+import logging
+import sys
+
+import ruamel.yaml
+
+
+log = logging.getLogger('migrate')
+
+
+def migrate(content):
+    data = ruamel.yaml.load(content, ruamel.yaml.RoundTripLoader)
+
+    service_names = data.keys()
+    for name, service in data.items():
+        # remove links and external links
+        service.pop('links', None)
+        external_links = service.pop('external_links', None)
+        if external_links:
+            log.warn(
+                "Service {name} has external_links: {ext}, which are no longer "
+                "supported. See https://docs.docker.com/compose/networking/ "
+                "for options on how to connect external containers to the "
+                "compose network.".format(name=name, ext=external_links))
+
+        # net is now networks
+        if 'net' in service:
+            service['networks'] = [service.pop('net')]
+
+        # create build section
+        if 'dockerfile' in service:
+            service['build'] = {
+                'context': service.pop('build'),
+                'dockerfile': service.pop('dockerfile'),
+            }
+
+        # create logging section
+        if 'log_driver' in service:
+            service['logging'] = {'driver': service.pop('log_driver')}
+            if 'log_opt' in service:
+                service['logging']['options'] = service.pop('log_opt')
+
+        # volumes_from prefix with 'container:'
+        for idx, volume_from in enumerate(service.get('volumes_from', [])):
+            if volume_from.split(':', 1)[0] not in service_names:
+                service['volumes_from'][idx] = 'container:%s' % volume_from
+
+    data['services'] = {name: data.pop(name) for name in data.keys()}
+    data['version'] = 2
+    return data
+
+
+def write(stream, new_format, indent, width):
+    ruamel.yaml.dump(
+        new_format,
+        stream,
+        Dumper=ruamel.yaml.RoundTripDumper,
+        indent=indent,
+        width=width)
+
+
+def parse_opts(args):
+    parser = argparse.ArgumentParser()
+    parser.add_argument("filename", help="Compose file filename.")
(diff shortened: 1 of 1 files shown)


contrib/migration/migrate-compose-file-v1-to-v2.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+def migrate(content):
+    data = ruamel.yaml.load(content, ruamel.yaml.RoundTripLoader)
+
+    service_names = data.keys()
+    for name, service in data.items():
+        # remove links and external links
+        service.pop('links', None)
+        service.pop('external_links', None)
```
aanand: Should we perhaps show a warning if the user has defined anything we can't seamlessly convert? No-alias `links` entries are fine and can be discarded, but anything else is probably going to break their app if it's missing.

dnephin: Makes sense, we could warn on `external_links`

dnephin:
Added the warning for external_links

aanand:
LGTM

aanand:
Cherry-picked into #2753

AvdN:
The indent level in ruamel.yaml is not preserved, but the code tries to preserve the starting column  of a comment (so they stay aligned). I checked and there was a bug, only appearing when reindenting to a larger value when a scalar was pushed against the comment. In version 0.10.21 this is fixed and an extra space inserted to keep scalar and comment apart.

Do you have an example of comments above lines disappearing? Please sent me some examples, so I can reproduce/fix.

I assume that the major problem with dropping of quotes lies in the fact that something like "2525:2525" is considered to be better kept in quotes because if you change `2525:2525` to `25:25`  PyYAML sees that as a sexagesimal. 

-  ruamel.yaml does put quotes around strings that would be sexagesimals, but that doesn't reduce the risk of `2525:2525` as mentioned above.
- sexagesimals are a YAML 1.1 artifact, not mentioned in the YAML 1.2 specification.  
- I can provide you with a PR on docker-compose with ruamel.yaml and **a loader that doesn't recognise sexagesimals at all (i.e. is only YAML 1.2 compliant)**, that way `- 25:25` will always be a list item that is a string scalar, and dropping the quotes on conversion is less off/no longer an issue.

Please let me know if you are interested.

dnephin:
@AvdN thanks for the bug fix with the comments.

For comments above lines disappearing, I was pretty sure I had a case to reproduce it, but I can't seem to anymore, so maybe it was also fixed, or maybe it's a very isolated case (and I forget now exactly where).

I don't know if there is a major problem with dropping quotes, it's just a difference from source and output. If someone happens to like to quote everything for some reason, those quotes will be gone. I don't think we need to worry about it much.

That change to drop support for sexagesimals might be a good idea, I think it's very unlikely anyone will actually want to write a number that way, and it has confused users before.  A PR to fix that would be awesome, thanks!

AvdN:
I agree that the support for YAML 1.1 specifics is more confusing that helpful.
The best solution would be to make ruamel.yaml aware of a possible `%YAML 1.x` header and take action based on that. It would have to default to 1.2 and when using the default possible give a one of warning if a construct would have given a different result  in 1.1. That is however a longer term task.

For me (handling question on StackOverflow) the problems with YAML 1.1 include:
- sexagesimals (`12:34:56` loading as integer 45296, and it cannot deal with 61 (leap)second minutes  anyway ;-) )
- octals ( 012 loads as integer 10, 012345678 loads a string "012345678"
- yes, no, on, off loading as True/False (and thus needing quotes).
If users of docker-compose actually *use* these PyYAML/YAML1.1 features, they almost certainly know what they are doing, most will put quotes around the constructs as needed and recommended in the docker-compose manual. 

To be clear switching to using ruamel.yaml without supporting the above 1.1 constructs would **not** affect existing code that already has quotes. That the converter doesn't write the quotes is a different issue, I'll look into that as well, I can already preserve literal scalars  (multi-line strings starting with `: |`), so the basics for that are there.

PR #2694, opened 2016-01-18
Fix some bugs in release scripts

dnephin:
* upload to pypi with twine fails because the tarball doesn't have a dash in the name, this removes the dash
* print the contribution count as well as the list of user names


aanand:
LGTM

PR #2696, opened 2016-01-19
Test that you can set the default network to be external

aanand:


dnephin:
LGTM

aanand:
Cherry-picked into #2753

PR #2699, opened 2016-01-19
Update Compose file documentation for version 2

### LINKED ISSUE #2622: Update compose file versioning documentation
See: https://github.com/docker/compose/pull/2421#discussion_r45494335

> `## Service configuration reference`  I would most likely put under this heading a paragraph about the fact there are multiple versions.  Then put a cross -reference to the version section to follow.

> Then move `## Versioning` to follow the main body of configuration material

> This section would need to answer:

> * define whether the configuration version matches the product version 
> * under what conditions can I use  a different version; for example does the version of the configuration file match or differ from the Compose version
> * can I mix syntax in the same config file (prob not but should say)
> * can I use two config files that use differing versions of the syntax (I'm thinking also of how the override files interact with the main config under the version conditions)
> * provide some guidance to the reader under what circumstances this version mixing is useful

aanand:
- Explain each version in its own section
- Explain how to upgrade from version 1 to 2
- Note which keys are restricted to particular versions
- A few corrections to the docs for version-specific keys

Closes #2622.

### DIFF
--- a/docs/compose-file.md
+++ b/docs/compose-file.md
@@ -318,6 +318,10 @@ container name and the link alias (`CONTAINER:ALIAS`).
      - project_db_1:mysql
      - project_db_1:postgresql
 
+> **Note:** If you're using the [version 2 file format](#version-2), the
+> externally-created containers must be connected to at least one of the same
+> networks as the service which is linking to them.
+
 ### extra_hosts
 
 Add hostname mappings. Use the same values as the docker client `--add-host` parameter.
@@ -358,27 +362,24 @@ It's recommended that you use reverse-DNS notation to prevent your labels from c
 
 ### links
 
-> [Version 1 file format](#version-1) only. In version 2 files, use
-> [networking](networking.md) for communication between containers.
-
 Link to containers in another service. Either specify both the service name and
-the link alias (`SERVICE:ALIAS`), or just the service name (which will also be
-used for the alias).
+a link alias (`SERVICE:ALIAS`), or just the service name.
 
-    links:
-     - db
-     - db:database
-     - redis
+    web:
+      links:
+       - db
+       - db:database
+       - redis
 
-An entry with the alias' name will be created in `/etc/hosts` inside containers
-for this service, e.g:
+Containers for the linked service will be reachable at a hostname identical to
+the alias, or the service name if no alias was specified.
 
-    172.17.2.186  db
-    172.17.2.186  database
-    172.17.2.187  redis
+Links also express dependency between services in the same way as
+[depends_on](#depends-on), so they determine the order of service startup.
 
-Environment variables will also be created - see the [environment variable
-reference](env.md) for details.
+> **Note:** If you define both links and [networks](#networks), services with
+> links between them must share at least one network in common in order to
+> communicate.
 
 ### logging
 
@@ -862,7 +863,6 @@ In the majority of cases, moving from version 1 to 2 is a very simple process:
 
 1. Indent the whole file by one level and put a `services:` key at the top.
 2. Add a `version: 2` line at the top of the file.
-3. Delete all `links` entries.
 
 It's more complicated if you're using particular configuration features:
 
@@ -879,14 +879,28 @@ It's more complicated if you're using particular configuration features:
           options:
             syslog-address: "tcp://192.168.0.42:123"
 
--   `links` with aliases: If you've defined a link with an alias such as
-    `myservice:db`, there's currently no equivalent to this in version 2. You
-    will have to refer to the service using its name (in this example,
-    `myservice`).
+-   `links` with environment variables: As documented in the
+    [environment variables reference](env.md), environment variables created by
+    links have been deprecated for some time. In the new Docker network system,
+    they have been removed. You should either connect directly to the
+    appropriate hostname or set the relevant environment variable yourself,
+    using the link hostname:
 
--   `external_links`: Links are deprecated, so you should use
-    [external networks](networking.md#using-externally-created-networks) to
-    communicate with containers outside the app.
+        web:
+          links:
+            - db
+          environment:
+            - DB_PORT=tcp://db:5432
+
+-   `external_links`: Compose uses Docker networks when running version 2
+    projects, so links behave slightly differently. In particular, two
+    containers must be connected to at least one network in common in order to
+    communicate, even if explicitly linked together.
+
+    Either connect the external container to your app's
+    [default network](networking.md), or connect both the external container and
+    your service's containers to an
+    [external network](networking.md#using-a-pre-existing-network).
 
 -   `net`: If you're using `host`, `bridge` or `none`, this is now replaced by
     `networks`:
--- a/docs/compose-file.md
+++ b/docs/compose-file.md
@@ -169,6 +169,29 @@ client create option.
     devices:
       - "/dev/ttyUSB0:/dev/ttyUSB0"
 
+### depends_on
+
+Express dependency between services, which has two effects:
+
+- `docker-compose up` will start services in dependency order. In the following
+  example, `db` and `redis` will be started before `web`.
+
+- `docker-compose up SERVICE` will automatically include `SERVICE`'s
+  dependencies. In the following example, `docker-compose up web` will also
+  create and start `db` and `redis`.
+
+    version: 2
+    services:
+      web:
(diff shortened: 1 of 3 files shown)


docs/compose-file.md
```diff
@@ -577,6 +603,211 @@ documentation for more information. Optional.
...
+          - back
+      redis:
+        image: redis
+        volumes:
+          - data:/var/lib/redis
+        networks:
+          - back
+    volumes:
+      data:
+        driver: default
```
dnephin: This isn't actually valid (from my tests at least). I think it needs to be either `local`, `{}`, or blank.  There's another instance of this in `docs/index.md`

dnephin:
cc @londoncalling for docs changes

docs/compose-file.md
```diff
@@ -92,28 +42,29 @@ definition.
 
 Configuration options that are applied at build time.
 
-In version 1 this must be given as a string representing the context.
+In the [version 1 file format](#version-1), `build` must be specified as a
+string containing the path to the build context.
```
dnephin: Should we maybe start with v2, and move the v1 format to the bottom of each section?

docs/compose-file.md
```diff
@@ -577,6 +603,211 @@ documentation for more information. Optional.
...
+## Network configuration reference
+
+The top-level `networks` key lets you specify networks to be created. For a full
+explanation of Compose's use of Docker networking features, see the
+[Networking guide](networking.md).
+
+### driver
+
+Specify which driver should be used for this network. Defaults to `bridge`. The
+Docker Engine will return an error if the driver is not available.
```
dnephin: I believe it actually defaults to `overlay` on swarm, so we might want to say the default depends on the engine.

docs/compose-file.md
```diff
@@ -577,6 +603,211 @@ documentation for more information. Optional.
...
+These differences are explained below.
+
+
+### Version 1
+
+Compose files that do not declare a version are considered "version 1". In
+those files, all the [services](#service-configuration-reference) are declared
+at the root of the document.
+
+Version 1 files are supported by **all versions of Compose**.
```
dnephin: We should probably say "all previous versions" or something like that.

I think we'll want to deprecate v1 after a few releases. I think it would be good to make that clear. It's currently supported, but will be dropped at some point, and we won't be adding any new features to V1.  So feature requests for things like build args will be rejected.

docs/compose-file.md
```diff
@@ -577,6 +603,211 @@ documentation for more information. Optional.
...
+      image: redis
+
+
+### Version 2
+
+Compose files using the version 2 syntax must indicate the version number at
+the root of the document. All [services](#service-configuration-reference)
+must be declared under the `services` key.
+
+Version 2 files are supported by **Compose 1.6.0+** and **Engine 1.10.0+**.
```
dnephin: and require **Engine 1.10.0+** ?

docs/compose-file.md
```diff
@@ -12,79 +12,29 @@ parent="smn_compose_ref"
...
+[volumes](#volume-configuration-reference).
+The default path for a Compose file is `./docker-compose.yml`.
 
-Each service defined in `docker-compose.yml` must specify exactly one of
-`image` or `build`. Other keys are optional, and are analogous to their
-`docker run` command-line counterparts.
+A service definition contains configuration which will be applied to each
+container started for that service, much like passing command-line parameters to
+`docker run`. Likewise, network and volume definitions are analogous to
+`docker network create` and `docker volume create`.
```
dnephin: We have had a couple reports from users about it being difficult to find the docs about variable interpolation.  It's in the menu on the right, but the menu is so long it's easy to miss (since it's at the bottom).

It would probably be good to mention variable substitution as part of the intro paragraph here, that it's supported in all the sections of the config, and link to the section at the bottom.
It doesn't need to happen as part of these changes though.

docs/compose-file.md
```diff
@@ -577,6 +603,211 @@ documentation for more information. Optional.
...
+            syslog-address: "tcp://192.168.0.42:123"
+
+-   `links` with aliases: If you've defined a link with an alias such as
+    `myservice:db`, there's currently no equivalent to this in version 2. You
+    will have to refer to the service using its name (in this example,
+    `myservice`).
+
+-   `external_links`: Links are deprecated, so you should use
+    [external networks](networking.md#using-externally-created-networks) to
+    communicate with containers outside the app.
```
dnephin: Or have the external container join the network by calling `docker network connect ...` ?

docs/compose-file.md
```diff
@@ -577,6 +603,211 @@ documentation for more information. Optional.
...
+      config:
+        - subnet: 172.28.0.0/16
+          ip_range: 172.28.5.0/24
+          gateway: 172.28.5.254
+          aux_addresses:
+            host1: 172.28.1.5
+            host2: 172.28.1.6
+            host3: 172.28.1.7
+
+
```
dnephin: We're missing docs for `external` here. It's covered in the narrative docs, but I think it needs to be in the reference.

I think we're also missing those docs for volumes.

docs/networking.md
```diff
@@ -99,11 +98,35 @@ Here's an example Compose file defining several networks. The `proxy` service is
...
-        # The 'outside' network is expected to already exist - Compose will not
-        # attempt to create it
         external: true
 
+You can also specify the name of the network separately from the name used to refer to it within the Compose file:
+
+    networks
+      outside:
+        external:
+          name: actual-name-of-network
```
dnephin: What do you think about moving this section to the reference to document `external` , and have a section here that talks about changing the defualt network using `external` instead?

docs/compose-file.md
dnephin: I'll fix the one in `docs/index.md`

londoncalling:
@dnephin fyi, taking a look at this today, thx!

docs/compose-file.md
aanand: I've added a paragraph about it.

aanand:
- Updated for @dnephin's feedback
- Documented `depends_on`
- Updated now that `links` are in v2

aanand:
@londoncalling Thoughts on the latest iteration?

docs/compose-file.md
```diff
@@ -562,20 +610,317 @@ subcommand documentation for more information.
...
+If set to `true`, specifies that this volume has been created outside of
+Compose. `docker-compose up` will not attempt to create it, and will raise
+an error if it doesn't exist.
+
+`external` cannot be used in conjunction with other volume configuration keys
+(`driver`, `driver_opts`).
+
+In the example below, instead of attemping to create a volume called
+`&lt;projectname&gt;_data`, Compose will look for an existing volume simply
```
londoncalling: fix formatting on project name

docs/compose-file.md
```diff
@@ -562,20 +610,317 @@ subcommand documentation for more information.
...
+
+If set to `true`, specifies that this network has been created outside of
+Compose. `docker-compose up` will not attempt to create it, and will raise
+an error if it doesn't exist.
+
+`external` cannot be used in conjunction with other network configuration keys
+(`driver`, `driver_opts`, `ipam`).
+
+In the example below, `proxy` is the gateway to the outside world. Instead of
+attemping to create a network called `&lt;projectname&gt;_outside`, Compose will
```
londoncalling: need to fix formatting for project name (2)

docs/networking.md
```diff
@@ -123,6 +134,17 @@ Instead of (or as well as) specifying your own networks, you can also change the
...
+## Using a pre-existing network
+
+If you want your containers to join a pre-existing network, use the [`external` option](compose-file.md#network-configuration-reference):
+
+    networks:
+      default:
+        external:
+          name: my-pre-existing-network
+
+Instead of attemping to create a network called `&lt;projectname&gt;_default`, Compose will look for a network called `my-pre-existing-network` and connect your app's containers to it.
```
londoncalling: need to fix formatting on project name (3 of 3)

londoncalling:
Hey @aanand , other than the minor copy-edit thing I just found (see previous comments), LGTM.

aanand:
Thanks. Just rebased and cleaned up and will merge once the build passes.

aanand:
Cherry-picked into #2753

PR #2706, opened 2016-01-19
Don't copy host volumes on recreate

dnephin:
I was doing some more testing of copy during recreate, and I noticed that if I removed a host volume, and there was a volume defined in the image for the same path or if you made it a container volume, you would continue to get the host volume instead.

This change makes it so that if you remove the host volume from the config, you won't keep getting the host volume.

### DIFF
--- a/compose/service.py
+++ b/compose/service.py
@@ -911,6 +911,10 @@ def get_container_data_volumes(container, volumes_option):
         if not mount:
             continue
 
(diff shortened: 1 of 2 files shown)


dnephin:
ping @shin- @aanand 

would be good to get this small fix into 1.6, I forgot to add the milestone earlier

shin-:
LGTM

aanand:
Cherry-picked into #2765

PR #2708, opened 2016-01-19
Implement depends_on

### LINKED ISSUE #2682: Proposal: New field to define ordered start and stop
Original implementation is #686

Now that networks don't require you to define service dependencies we could provide a new field to define an order, allowing applications to start and stop containers in a specific order.

The advantage to a new field (and not re-using an existing one), is that it allows applications that don't need an order to start and stop everything in parallel.  Order would only be applied explicitly if the user wants it.

This does not address healthchecking and the problems described in #374, but it's a prerequisite of any solution to that problem. It does solve issues like #2619 (and maybe others?).

Current proposal is to name the field `depends_on`

dnephin:
Fixes #2682

Second and third commit contain some cleanup for things I noticed while working on it.

Services will be ordered by the `depends_on` key.

### DIFF
--- a/compose/config/config.py
+++ b/compose/config/config.py
@@ -31,6 +31,7 @@ from .validation import validate_depends_on
 from .validation import validate_extends_file_path
 from .validation import validate_top_level_object
 from .validation import validate_top_level_service_objects
+from .validation import validate_ulimits
 
 
 DOCKER_CONFIG_KEYS = [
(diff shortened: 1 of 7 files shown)


aanand:
LGTM

schmunk42:
Just stumbled upon this merge :)
This will fix my issue described in https://github.com/docker/swarm/issues/1651#issuecomment-173124464, right? 

TL;dr: `links` were previously responsible for ordering

dnephin:
That's the idea, yes

cescoferraro:
Are there any instruction anywhere on how to install the latest compose so I can test this?

aanand:
@cescoferraro RC2 will be out with this soon.

aanand:
Cherry-picked into #2753

aanand:
@cescoferraro [RC2 is out now](https://github.com/docker/compose/releases/tag/1.6.0-rc2).

cescoferraro:
I have struggled with this for so long 😭😭 @aanand Great job!

PR #2710, opened 2016-01-20
Install python from deb packages instead of source in the CI Dockerfile

dnephin:
I think this should reduce the time of our test runs.

I had to upgrade enum34 to get it to install with the slightly newer version of python 3.4

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,59 +1,28 @@
-FROM debian:wheezy
+FROM debian:jessie
 
-RUN set -ex; \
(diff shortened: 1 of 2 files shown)


dnephin:
I missed the python-dev package I think, the build binary if failing.

dnephin:
Closing this since it didn't seem to save much time.

I think a better solution here will be to have a build matrix, and have separate docker images for each build target (py27, py34, linting, etc). That way each build can run in parallel, and only needs to build the parts it uses.

PR #2712, opened 2016-01-20
Fix `extends` when using v2 config format

jzvelc:
This fixes configuration error (Service not found) when extending a service defined in v2 format.

### DIFF
--- a/compose/config/config.py
+++ b/compose/config/config.py
@@ -134,6 +134,9 @@ class ConfigFile(namedtuple('_ConfigFile', 'filename config')):
             return 1
         return version
 
+    def get_service(self, name):
+        return self.get_service_dicts()[name]
+
     def get_service_dicts(self):
         return self.config if self.version == 1 else self.config.get('services', {})
 
@@ -351,19 +354,19 @@ def process_config_file(config_file, service_name=None):
 
     if config_file.version == 2:
         processed_config = dict(config_file.config)
-        processed_config['services'] = interpolated_config
+        processed_config['services'] = services = interpolated_config
         processed_config['volumes'] = interpolate_environment_variables(
             config_file.get_volumes(), 'volume')
         processed_config['networks'] = interpolate_environment_variables(
             config_file.get_networks(), 'network')
 
     if config_file.version == 1:
-        processed_config = interpolated_config
+        processed_config = services = interpolated_config
 
     config_file = config_file._replace(config=processed_config)
     validate_against_fields_schema(config_file)
 
-    if service_name and service_name not in processed_config:
+    if service_name and service_name not in services:
         raise ConfigurationError(
             "Cannot extend service '{}' in {}: Service not found".format(
                 service_name, config_file.filename))
@@ -408,7 +411,8 @@ class ServiceExtendsResolver(object):
         extended_file = process_config_file(
             extends_file,
             service_name=service_name)
-        service_config = extended_file.config[service_name]
+        service_config = extended_file.get_service(service_name)
+
         return config_path, service_config, service_name
 
     def resolve_extends(self, extended_config_path, service_dict, service_name):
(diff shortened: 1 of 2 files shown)


compose/config/config.py
```diff
@@ -363,7 +363,8 @@ def process_config_file(config_file, service_name=None):
     config_file = config_file._replace(config=processed_config)
     validate_against_fields_schema(config_file)
 
-    if service_name and service_name not in processed_config:
+    if service_name and (config_file.version == 1 and service_name not in processed_config or
+                         config_file.version == 2 and service_name not in processed_config['services']):
```
dnephin: Instead of branching again on the version, could we set a `services` variable to the correct value in the  branch above this?

```python
processed_config['services'] = services = interpolated_config

...

processed_config = services = interpolated_config

```

Then this can be just 

```python

if service_name and service_name in services:
    ...

```

compose/config/config.py
```diff
@@ -408,7 +409,13 @@ def validate_and_construct_extends(self):
         extended_file = process_config_file(
             extends_file,
             service_name=service_name)
-        service_config = extended_file.config[service_name]
+
+        if extended_file.version == 2:
+            service_config = extended_file.config['services'][service_name]
+
+        if extended_file.version == 1:
+            service_config = extended_file.config[service_name]
```
dnephin: We'd like to keep the version if/else in as few places as possible.  Right now most of it is done in the `ConfigFile` class (this `extended_file` is an instance of one).

Could you move this condition into a `ConfigFile.get_service(name)` so this call can be:

```python

service_config = extended_file.get_service(service_name)

```

dnephin:
Thanks for the bug report, and the PR to fix it!

jzvelc:
@dnephin Updated.

compose/config/config.py
```diff
@@ -134,6 +134,9 @@ def version(self):
             return 1
         return version
 
+    def get_service(self, name):
+        return self.get_service_dicts()[name]
```
dnephin: I think we might want to do either `.get(name)` or a try/except `KeyError` raise a ConfigurationError. It might not be possible for that condition to happen now, but as code changes it would be good to raise an appropriate error.

dnephin:
Thanks for making those changes.  I wonder if we should parameterize all the extends tests to run with both v1 and v2 configs, so that we can test all the cases.  I don't think we need to do that as part of this change.

aanand:
LGTM

aanand:
Cherry-picked into #2753

marcelmfs:
I just saw this happening again with docker-compose 1.15. Anybody else seeing the same?

mrit:
I have the same problem @marcelmfs

michaelclarkcuadrado:
@marcelmfs Just upgraded to 1.15 and all my extends are now broken

neven-s:
I've had the same problem and fixed it by updating docker to latest version (v17.06)

ssassi:
Not working for me. I have docker-compose 1.15 and docker v17.07.

emyller:
Not working for me either. Going back to docker-compose 1.14 solves the problem.

epollan:
@emyller this is affecting me, too.  docker 17.07 + docker-compose 1.15.  Are you on osx?  If so, do you mind sharing how you did the downgrade?

emyller:
@epollan I'm on Arch Linux (main dev box); had to clone `docker-compose-bin` at 1.14.x and build it. Not using Docker on my OSX, but I believe I would have to do something similar.

joaocc:
Having the same issue in ```docker-compose version 1.17.0, build ac53b73``` on OSX.

samupl:
Now I see the exact same behavior:

```$ docker-compose --version
docker-compose version 1.20.1, build 5d8c71b
```

What happens when I run an extended config:
```
$ docker-compose -f test.yml up
ERROR: Cannot extend service 'common.yml' in /home/js/test/test.yml: Service not found
```

test.yml:
```
version: '2'
services:

  redis:
    extends: common.yml
```

common.yml:
```
version: '2'
services:
  redis:
    image: redis
    volumes:
    - ./storage/redis:/data
```

PR #2723, opened 2016-01-21
Fix 'run' behaviour with networks

aanand:
- Test that one-off containers join all networks
- Don't set any aliases

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,8 +1,8 @@
 PyYAML==3.11
 cached-property==1.2.0
-dockerpty==0.3.4
 docopt==0.6.1
 enum34==1.0.4
+git+https://github.com/d11wtq/dockerpty.git@29b1394108b017ef3e3deaf00604a9eb99880d5e#egg=dockerpty
 git+https://github.com/docker/docker-py.git@master#egg=docker-py
 jsonschema==2.5.1
 requests==2.7.0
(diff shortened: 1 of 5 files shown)


dnephin:
LGTM

aanand:
Updated - we were relying on a non-feature of Engine that's about to get fixed. Proper fix is now in, and reliant on https://github.com/d11wtq/dockerpty/pull/54.

aanand:
https://github.com/d11wtq/dockerpty/pull/54 is merged.

requirements.txt
```diff
@@ -1,8 +1,8 @@
 PyYAML==3.11
 cached-property==1.2.0
-dockerpty==0.3.4
 docopt==0.6.1
 enum34==1.0.4
+git+https://github.com/d11wtq/dockerpty.git@master#egg=dockerpty
```
dnephin: I think we should be using a sha here, otherwise it could break without any changes to compose

mavenugo:
@aanand FYI. If a container needs to be connected to multiple networks, the recommended approach is -
1. create the container in 1 network with the required options
2. connect to other networks with its own network specific option
3. start the container

Though we allow connecting to networks after start, the application would assume that all the networks are connected to when it starts. Hence, it is better to follow the above sequence.

dnephin:
I think that's what we're doing now.

Got a link to the engine PR?

Other than the pinning `master`, LGTM

requirements.txt
aanand: True - same goes for docker-py, I'll fix that separately

aanand:
Cherry-picked into #2753

PR #2728, opened 2016-01-22
Match named volumes in service definitions with declared volumes

### LINKED ISSUE #2709: [1.6.0] Extra volumes are created when using new compose file format
I have __docker-compose.yml__ file in folder __nowsd__:
```yml
version: 2
volumes:
  data:
    driver: local
  ugc:
    driver: local
services:
  web:
    image: busybox
    volumes:
      - data:/var/log
      - ugc:/var/ugc
```
_$_ __docker-compose up__
```
Starting nowsd_web_1
Attaching to nowsd_web_1
nowsd_web_1 exited with code 0
```
_$_ __docker volume ls__
```
DRIVER              VOLUME NAME
local               nowsd_data
local               data
local               ugc
local               nowsd_ugc
```

You can see that two extra unscoped volumes has been created. When I inspect container nowsd_web_1 i can see that volumes _data_ and _ugc_ are attached to container instead of _nowsd_data_ and _nowsd_ugc_ respectively.

shin-:
Fixes #2709

### DIFF
--- a/compose/config/types.py
+++ b/compose/config/types.py
@@ -163,3 +163,9 @@ class VolumeSpec(namedtuple('_VolumeSpec', 'external internal mode')):
     def repr(self):
         external = self.external + ':' if self.external else ''
         return '{ext}{v.internal}:{v.mode}'.format(ext=external, v=self)
+
+    @property
+    def is_named_volume(self):
+        return self.external and not (
+            self.external.startswith('.') or self.external.startswith('/')
+        )
--- a/compose/config/types.py
+++ b/compose/config/types.py
@@ -166,6 +166,4 @@ class VolumeSpec(namedtuple('_VolumeSpec', 'external internal mode')):
 
     @property
     def is_named_volume(self):
-        return self.external and not (
-            self.external.startswith('.') or self.external.startswith('/')
-        )
+        return self.external and not self.external.startswith(('.', '/', '~'))
--- a/compose/project.py
+++ b/compose/project.py
@@ -74,6 +74,17 @@ class Project(object):
         if 'default' not in network_config:
             all_networks.append(project.default_network)
 
+        if config_data.volumes:
+            for vol_name, data in config_data.volumes.items():
+                project.volumes.append(
+                    Volume(
+                        client=client, project=name, name=vol_name,
+                        driver=data.get('driver'),
+                        driver_opts=data.get('driver_opts'),
+                        external_name=data.get('external_name')
+                    )
+                )
+
         for service_dict in config_data.services:
             if use_networking:
                 networks = get_networks(service_dict, all_networks)
@@ -86,6 +97,9 @@ class Project(object):
 
             volumes_from = get_volumes_from(project, service_dict)
 
+            if config_data.version == 2:
+                match_named_volumes(service_dict, project.volumes)
+
             project.services.append(
                 Service(
                     client=client,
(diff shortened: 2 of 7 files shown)


shin-:
Hmm, it's difficult to detect whether a VolumeSpec defines a bind volume or not. I'll finish this later, but the rest should be sound.

tests/integration/project_test.py
```diff
@@ -813,3 +813,39 @@ def test_initialize_volumes_inexistent_external_volume(self):
...
+        )
+        service = project.services[0]
+        self.assertEqual(service.name, 'simple')
+        volumes = service.options.get('volumes')
+        self.assertEqual(len(volumes), 1)
+        self.assertEqual(volumes[0].external, full_vol_name)
+        project.up()
+        engine_volumes = self.client.volumes()
+        self.assertIsNone(next(v for v in engine_volumes if v['Name'] == vol_name))
+        self.assertIsNotNone(next(v for v in engine_volumes if v['Name'] == full_vol_name))
```
dnephin: For the assertions, I think it might be more obvious what is being asserted if you do something like this:

```
container, = service.get_containers()
assert [mount['Name'] for mount in container.get('Mounts')] == [full_vol_name]
```

compose/config/types.py
```diff
@@ -163,3 +163,9 @@ def parse(cls, volume_config):
     def repr(self):
         external = self.external + ':' if self.external else ''
         return '{ext}{v.internal}:{v.mode}'.format(ext=external, v=self)
+
+    @property
+    def is_named_volume(self):
+        return self.external and not (
+            self.external.startswith('.') or self.external.startswith('/')
```
dnephin: I believe `~` is the other path we consider as a host bind. We do a similar check in `split_path_mapping()`.

I think we could combine these and do something like:

```python
def is_host_bind(path):
    return path.startswith(('.', '~', os.sep))
```

Note: startswith can take a tuple: https://docs.python.org/2/library/stdtypes.html#str.startswith

compose/project.py
```diff
@@ -473,6 +477,23 @@ def get_networks(service_dict, network_definitions):
...
+                (v for v in project_volumes if v.name == volume_spec.external),
+                None
+            )
+            if not declared_volume:
+                raise ConfigurationError(
+                    'Named volume "{0}" is used in service "{1}" but no'
+                    ' declaration was found in the volumes section.'.format(
+                        volume_spec.repr(), service_dict.get('name')
+                    )
+                )
```
dnephin: I think we should do this validation as part of `compose/config/config` similar to how we validate depends_on. That way we the `docker-compose config` command will catch the error (since that command is supposed to be able to validate a config).

Once we've validated that all the named volumes are defined, we can just add the prefix in service when we build the binds without worrying about looking it up.

compose/project.py
```diff
@@ -473,6 +477,23 @@ def get_networks(service_dict, network_definitions):
     return networks
 
 
+def match_named_volumes(service_dict, project_volumes):
+    for volume_spec in service_dict.get('volumes', []):
+        if volume_spec.is_named_volume:
+            declared_volume = next(
+                (v for v in project_volumes if v.name == volume_spec.external),
+                None
+            )
```
dnephin: Instead of doing a n^2 lookup, creating a dict keyed by external name would make this an n.

compose/config/types.py
shin-: Thanks!
However, `split_path_mapping` tests for a relative path, so it's slightly different.
Also, the paths in `VolumeSpec` are already normalized for the docker engine, so we don't need to consider OS variance.

shin-:
Thanks for the review @dnephin ! This should look good now.

aanand:
CI is upset for reasons seemingly unrelated to the code, but all tests pass locally, so LGTM.

aanand:
Cherry-picked into #2753

PR #2734, opened 2016-01-23
Empty env var #2359

### LINKED ISSUE #2359: Inconsistency in how compose handles empty environment variables
If you want to use the value of an environment variable that is available in the host you can:

```bash
$ HOST_ENV=value docker run -e HOST_ENV ubuntu:14.04 env
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
HOSTNAME=acb280b6edd9
HOST_ENV=value
HOME=/root
```

But, if the variable is not defined in the host, then it also doesn't get created in the container:

```bash
$ docker run -e HOST_ENV ubuntu:14.04 env
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
HOSTNAME=29a464f43ecb
HOME=/root
```

But, given the following `docker-compose.yml`:

```yaml
banana:
  image: ubuntu:14.04
  command: env
  environment:
    HOST_ENV:
```

We get an empty variable:

```bash
$ docker-compose run banana
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
HOSTNAME=9841adaf229b
TERM=xterm
HOST_ENV=
HOME=/root
```

Should this be considered a bug? I would expect the same behavior from both tools.

jrabbit:
closes #2359. I tried to keep the changes minimal. I also changed the behavior of env_file pass-throughs. This makes compose's behavior similar to docker run's.

### DIFF
--- a/compose/config/config.py
+++ b/compose/config/config.py
@@ -464,18 +464,12 @@ def resolve_environment(service_dict):
         env.update(env_vars_from_file(env_file))
 
     env.update(parse_environment(service_dict.get('environment')))
-    d = dict(resolve_env_var(k, v) for k, v in six.iteritems(env))
-    if '_' in d.keys():
-        del d['_']
-    return d
+    return dict(filter(None, (resolve_env_var(k, v) for k, v in six.iteritems(env))))
 
 
 def resolve_build_args(build):
     args = parse_build_arguments(build.get('args'))
-    d = dict(resolve_env_var(k, v) for k, v in six.iteritems(args))
-    if '_' in d.keys():
-        del d['_']
-    return d
+    return dict(filter(None, (resolve_env_var(k, v) for k, v in six.iteritems(args))))
 
 
 def validate_extended_service_dict(service_dict, filename, service):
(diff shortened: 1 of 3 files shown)


compose/config/config.py
```diff
@@ -464,12 +464,18 @@ def resolve_environment(service_dict):
...
+    return d
 
 
 def resolve_build_args(build):
     args = parse_build_arguments(build.get('args'))
-    return dict(resolve_env_var(k, v) for k, v in six.iteritems(args))
+    d = dict(resolve_env_var(k, v) for k, v in six.iteritems(args))
+    if '_' in d.keys():
+        del d['_']
+    return d
```
dnephin: Instead of using `'_'` as a special case, could we use the empty tuple (`()`).

If you `return ()` from `resolve_env_var()`, you can filter it here using:

```python
dict(filter(None, resolve_env_var(k, v) for k, v in six.iteritems(args)))
```

dnephin:
Thanks for the fix! Just one request for the special token used for "unset".

jrabbit:
I don't think that test failing has anything to do with this code though

dnephin:
Ya, the test failure is unrelated.

LGTM

shin-:
LGTM!

PR #2743, opened 2016-01-25
Implement network_mode option

### LINKED ISSUE #2701: New field `network_mode` for v2 format
Currently we support `host`, `bridge` and `none` as special cases in the list of networks, and we don't support `net: container:...` or `net: service:...`.

The new `network_mode` field would be a single scalar string value which supports any of the above, and we would remove this support from the  `networks` field.

aanand:
Closes #2701.


compose/service.py
```diff
@@ -468,7 +467,7 @@ def config_dict(self):
             'options': self.options,
             'image_id': self.image()['Id'],
             'links': self.get_link_names(),
-            'net': self.net.id,
+            'net': self.network_mode.id,
```
aanand: I didn't change the key from `'net'` here because it'll invalidate people's config hashes, but maybe that's not an issue?

compose/service.py
```diff
@@ -787,22 +786,22 @@ def pull(self, ignore_pull_failures=False):
                 log.error(six.text_type(e))
 
 
-class Net(object):
```
shin-: We are still using `Net`, `ContainerNet` and `ServiceNet` in `tests/integration/service_test.py` and `tests/unit/service_test.py`, causing failures.

aanand: Right, sorry, I meant to update those but forgot. Done. Only failures now are AUFS, not related to our code.

shin-:
I think this is good. Tests pass for me against dind-master. LGTM!

aanand:
Cherry-picked into #2753

PR #2744, opened 2016-01-25
Add pytest.ini

aanand:
Copied over from docker-py. This makes running `py.test` at the command line a bit less annoying.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -15,7 +15,7 @@ deps =
     -rrequirements.txt
     -rrequirements-dev.txt
 commands =
-    py.test -v -rxs \
+    py.test -v \
         --cov=compose \
         --cov-report html \
         --cov-report term \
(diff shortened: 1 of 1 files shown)


dnephin:
LGTM

pytest.ini
```diff
@@ -0,0 +1,2 @@
+[pytest]
+addopts = --tb=short -rxs
```
dnephin: If I remember correctly, I think we can actually add this to `tox.ini` and `py.test`  will read it the same way.

aanand: So it does. I've updated accordingly.

dnephin:
looks like all the jenkins nodes are dead, so the tests aren't running

LGTM

jfrazelle:
new feature @dnephin you like? ^

dnephin:
Ya! Being able to see the build failures from the PR is pretty cool

PR #2745, opened 2016-01-25
Fix computation of service lists in bash completion

albers:
Up to now, bash completion extracted services from "the" compose file (all services) or from container names (containers with special features, e.g. running ones).

This is no longer sufficient:
* compose files may be located in parent directories
* compose files may inherit from other files
* compose files may be enhanced or overridden by `docker-compose.override.yml`
* compose files may contain variables
* container names may be adjusted with the `container_name` directive

The completion of service names is therefore seriously broken. As some of these breaking changes are introduced in 1.6.0, I strongly suggest to include this patch in 1.6.0 as well.

This PR makes use of two new features in order to fix the shortcomings:
* `docker-compose config` gives access to the configuration as `compose` sees it
* the service name of a container is determined from its labels

There is still one feature missing: support for multiple `-f` options. I will add this in another PR.

/cc @sdurrheimer PTAL, maybe you can reuse the updated functions in zsh completion

### DIFF
--- a/contrib/completion/bash/docker-compose
+++ b/contrib/completion/bash/docker-compose
@@ -17,6 +17,10 @@
 #    . ~/.docker-compose-completion.sh
 
 
+__docker_compose_q() {
+	docker-compose 2>/dev/null ${compose_file:+-f $compose_file} ${compose_project:+-p $compose_project} "$@"
+}
+
 # suppress trailing whitespace
 __docker_compose_nospace() {
 	# compopt is not available in ancient bash versions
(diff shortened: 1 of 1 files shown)


albers:
@dnephin I think this is more than a cleanup. Completion of services is really broken as-is.
Please consider pulling it into 1.6.0.

aanand:
LGTM

dnephin:
LGTM

aanand:
Cherry-picked into #2765

PR #2746, opened 2016-01-25
Alias containers by short id

### LINKED ISSUE #2702: Containers should be discoverable from their container id
Containers currently join networks using the container name (default), and with their service name.

We'd like to change or deprecate the container name (#1516), so we should provide a second alias that allows a user to uniquely identify a container (when multiple containers have the same service name, when they are scaled).

We could either use the short container id, or `service-<short container id>`.

I think just container id might be better. It's already possible to query for this id from within the container (`$(hostname)`), but it's not possible to query for the service name.

aanand:
This necessitates a small change in how we connect to networks: rather than setting links and aliases for the one arbitrarily-picked network the container joins on creation, we disconnect and reconnect it immediately after creating it, so that we can set the short-id alias.

Closes #2702.

### DIFF
--- a/compose/service.py
+++ b/compose/service.py
@@ -430,12 +430,16 @@ class Service(object):
         return container
 
     def connect_container_to_networks(self, container):
-        one_off = (container.labels.get(LABEL_ONE_OFF) == "True")
+        connected_networks = container.get('NetworkSettings.Networks')
 
         for network in self.networks:
+            if network in connected_networks:
+                self.client.disconnect_container_from_network(
+                    container.id, network)
+
             self.client.connect_container_to_network(
                 container.id, network,
-                aliases=self._get_aliases(one_off=one_off),
+                aliases=self._get_aliases(container),
                 links=self._get_links(False),
             )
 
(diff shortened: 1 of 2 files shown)


shin-:
LGTM!

aanand:
Cherry-picked into #2753

compose/service.py
```diff
@@ -430,12 +430,16 @@ def start_container(self, container):
         return container
 
     def connect_container_to_networks(self, container):
-        one_off = (container.labels.get(LABEL_ONE_OFF) == "True")
+        connected_networks = container.get('NetworkSettings.Networks')
 
         for network in self.networks:
+            if network in connected_networks:
+                self.client.disconnect_container_from_network(
+                    container.id, network)
```
dnephin: Do we need to join a network when we create it? or could we just have it join all the networks here?

aanand: Yes, we need to set a `NetworkMode` or it'll default to `bridge` (and we'll just have to disconnect from that). I wondered about setting it to `none`, but it's then impossible to connect it to any networks.

dnephin: Good to know, thanks!

Might be worth making a feature request on engine for a way to accomplish this without connecting to a network originally.

PR #2748, opened 2016-01-25
fixed documentation about traversing yml files

schmunk42:
Actually compose is traversing "upwards" in the directory tree, looking in parent folders for a yml file. Using subdirectories wouldn't make sense, since you may end up finding multiple files.

Or I understood something completely wrong here :)


dnephin:
Yup, I think this was wrong,LGTM

aanand:
Thanks!

LGTM

aanand:
Cherry-picked into #2753

PR #2755, opened 2016-01-26
Fix rebase-bump-commit script

aanand:
Trim whitespace from `wc`'s output before constructing arguments to `git rebase`


dnephin:
Weird, I guess this is a difference in implementation from linux to osx?

I'd like to make an image which contains all the required programs for the release scripts.

LGTM

aanand:
Cherry-picked into #2765

PR #2753, opened 2016-01-26
Bump 1.6.0 rc2

aanand:
Do not merge.

PR #2756, opened 2016-01-26
Add back external links in v2

aanand:
It's perhaps a bit alarming that we don't have a test for this, but that's true of a lot of config options. We should figure out a strategy there.


shin-:
LGTM!

aanand:
Cherry-picked into #2753

PR #2757, opened 2016-01-26
Bump docker-py version to latest RC

shin-:


aanand:
This has exposed a bug, which I'll fix in a PR combined with this change.

aanand:
Replaced by #2761

Issue #2750, opened 2016-01-26
Compose support flow sequence written as a comma separated list within square brackets?


ruffsl:
I'm just testing a simple foo bar example using [variable substitution](https://docs.docker.com/compose/compose-file/#variable-substitution):

``` yaml
test:
  image: ubuntu
  volumes: ${FOO_BAR}
  command: ping 127.0.0.1
```

but am encountering this:

``` command
$ mkdir /tmp/foo
$ mkdir /tmp/bar
$ export FOO_BAR="['/tmp/bar:/bar', '/tmp/foo:/foo']"
$ docker-compose up
ERROR: Validation failed in file './docker-compose.yml', reason(s):
Service 'test' configuration key 'volumes' contains an invalid type, it should be an array
```

Does compose support this YAML spec, or am I doing something improper?
http://yaml.org/spec/1.2/spec.html#id2759963
relevant context: https://github.com/NVIDIA/nvidia-docker/issues/39

dnephin:
This is the same idea as #2730, only scalar strings are supported. You could set a single value, or have a variable for each, but you can't set an array in a variable.

tiry:
This is indeed the same issue.
Looks like based on the `service_schema` jsonschema we could try an "automatic cast" after the interpolation.

Issue #2730, opened 2016-01-22
Cast issue on int64 after variable interpolation in docker-compose.yml


tiry:
I am using compose variable interpolation because I would like to make part of the compose template configurable.

This works for some parameters, however, there seems to be an issue with parameters that expect an integer value (tested against 1.5.2 and 1.6 dev).

Typically, if I want to change the `cpu-shares` parameter I end up with an error :

```
ERROR: json: cannot unmarshal string into Go value of type int64
```

Looking at what happens in `interpolation.py` and checking the content of the interpolation result:
- when the `cpu-shares` is _"hard coded"_ the resulting dict is 
  
    { ... 'cpu_shares': 1024  ... }
- when the `cpu-shares` is set to use a variable using `cpu_shares: ${CPU_SHARES}` the resulting dict is 
  
    { ... 'cpu_shares': '1024' ... }

This looks to me like the root cause of the problem.

I am not sure what the best approach would be:
- should I try to add a tentative cast after the `Template(string).substitute` ?
- should I add some post processing in `config.py` or in `validate_against_fields_schema` ?

dnephin:
I don't  know of a good way to handle problem at this time, but I think it's something that does need to be fixed. We have the same issue with boolean fields.

tiry:
Just trying to discover the code and see how this could be fixed.

### Using Json Schema info

In `service_schema_v1.json` the `cpu_shares` is defined as

    "cpu_shares": {"type": ["number", "string"]},

I am not sure why `"string"` is accepted, based on what I can see in the `docker run` [doc](https://docs.docker.com/engine/reference/run/#cpu-share-constraint) it should not be the case (but I may very well have missed something).

I am asking because, if we had a single type in the schema, it could probably be used to do some automatic cast after the interpolation :

 - get the type of the field from the schema
 - do the cast from String as needed

The conversion could also be done as an error handler for the jsonschema validator, but I like the idea that the "automatic cast" is only applied for *"interpolated variables"*.

### Add conversion hint in the source template

An other option would be to add some hints in the templatised config file (ex `cpu_shares: int:${CPU_SHARES}`), but this does not seems like a clean approach (+ it is redundant info).

wu0407:
the similar problem,if i set
```
    logging:
      driver: "json-file"
      options:
        max-size: "10m"
        max-file: 3
```
docker-compose up end up
```
json: cannot unmarshal number into Go value of type string
```
docker version
```
[root@docker deploy]# docker version
Client:
[...]

PR #2759, opened 2016-01-26
Issue #2730 - better handle types for expanded variables

tiry:
Here is a changeset to handle automatic cast based on the types defined in the jsonschema.
I added test for my use case as well as for #2750.

My python is pretty rusted, but it seems to work as expected ... let me know what you think.

### DIFF
--- a/compose/config/interpolation.py
+++ b/compose/config/interpolation.py
@@ -1,6 +1,7 @@
 from __future__ import absolute_import
 from __future__ import unicode_literals
 
+import json
 import logging
 import os
 from string import Template
@@ -8,15 +9,17 @@ from string import Template
 import six
 
 from .errors import ConfigurationError
+
+
 log = logging.getLogger(__name__)
 
 
-def interpolate_environment_variables(config, section):
+def interpolate_environment_variables(config, section, schema):
     mapping = BlankDefaultDict(os.environ)
 
     def process_item(name, config_dict):
         return dict(
-            (key, interpolate_value(name, key, val, section, mapping))
+            (key, interpolate_value(name, key, val, section, mapping, schema))
             for key, val in (config_dict or {}).items()
         )
 
@@ -26,9 +29,46 @@ def interpolate_environment_variables(config, section):
     )
 
 
-def interpolate_value(name, config_key, value, section, mapping):
+def _cast_interpolated(interpolated_value, expected_type):
+
+    if expected_type == "string":
+        return None
+    elif expected_type == "number":
+        try:
+            return int(interpolated_value)
+        except ValueError:
+            return None
+    elif expected_type == "boolean":
+        if "true" == interpolated_value.lower():
+            return True
+        if "false" == interpolated_value.lower():
+            return False
+        return None
+    elif expected_type == "array":
+        if isinstance(interpolated_value, list):
+            return interpolated_value
+        return json.loads("{{\"data\":{0}}}".format(interpolated_value.replace("'", "\"")))["data"]
+
+
+def interpolate_value(name, config_key, value, section, mapping, schema):
     try:
-        return recursive_interpolate(value, mapping)
+        interpolated = recursive_interpolate(value, mapping)
+        if (interpolated != value):
+            # cast as needed
+            if config_key in schema["definitions"]["service"]["properties"]:
+                field_def = schema["definitions"]["service"]["properties"][config_key]
+                if "type" in field_def:
+                    allowed_types = field_def["type"]
+                    if isinstance(allowed_types, list):
+                        for allowed_type in allowed_types:
+                            converted = _cast_interpolated(interpolated, allowed_type)
+                            if converted is not None:
+                                return converted
+                    else:
+                        converted = _cast_interpolated(interpolated, allowed_types)
+                        if converted is not None:
+                            return converted
+        return interpolated
     except InvalidInterpolation as e:
         raise ConfigurationError(
             'Invalid interpolation format for "{config_key}" option '
@@ -41,7 +81,9 @@ def interpolate_value(name, config_key, value, section, mapping):
 
 def recursive_interpolate(obj, mapping):
     if isinstance(obj, six.string_types):
-        return interpolate(obj, mapping)
+        if '$' in obj:
+            return interpolate(obj, mapping)
+        return obj
     elif isinstance(obj, dict):
         return dict(
             (key, recursive_interpolate(val, mapping))
--- a/compose/config/interpolation.py
+++ b/compose/config/interpolation.py
@@ -30,9 +30,8 @@ def interpolate_environment_variables(config, section, schema):
 
 
 def _cast_interpolated_inside_list(interpolated_value, possible_types):
-    if "string" in possible_types:
-        possible_types.remove("string")
-    for possible_type in possible_types:
+    castable_types = [t for t in possible_types if t != "string"]
+    for possible_type in castable_types:
         v = _cast_interpolated(interpolated_value, possible_type)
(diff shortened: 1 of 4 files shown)


compose/config/validation.py
```diff
@@ -364,19 +364,33 @@ def validate_against_fields_schema(config_file):
...
         format_checker=["ports"],
         service_name=service_name)
 
 
+def load_jsonschema(prefix, version):
+    schema_filename = "{0}_schema_v{1}.json".format(prefix, version)
+    config_source_dir = os.path.dirname(os.path.abspath(__file__))
+
+    if sys.platform == "win32":
+        config_source_dir = config_source_dir.replace('\\', '/')
```
dnephin: I'm not sure why this is necessary. On windows I would expect `os.path` to return the correct paths.

compose/config/interpolation.py
```diff
@@ -41,7 +81,9 @@ def interpolate_value(name, config_key, value, section, mapping):
 
 def recursive_interpolate(obj, mapping):
     if isinstance(obj, six.string_types):
-        return interpolate(obj, mapping)
+        if '$' in obj:
+            return interpolate(obj, mapping)
+        return obj
```
dnephin: Was this necessary, or is it just a shortcut for the case where there are no variables?

compose/config/interpolation.py
```diff
@@ -26,9 +29,46 @@ def process_item(name, config_dict):
...
+                    if isinstance(allowed_types, list):
+                        for allowed_type in allowed_types:
+                            converted = _cast_interpolated(interpolated, allowed_type)
+                            if converted is not None:
+                                return converted
+                    else:
+                        converted = _cast_interpolated(interpolated, allowed_types)
+                        if converted is not None:
+                            return converted
+        return interpolated
```
dnephin: I think we could avoid doing this type inspection here. In `recursive_interpolate()` we can pull out the part of the schema relevant for the section we're going into and pass that as the value for `schema` when calling itself.

That way `cast_interpolated()` can be called from just the `else` block in `recursive_interpolate()`.

compose/config/interpolation.py
```diff
@@ -26,9 +29,46 @@ def process_item(name, config_dict):
...
+        try:
+            return int(interpolated_value)
+        except ValueError:
+            return None
+    elif expected_type == "boolean":
+        if "true" == interpolated_value.lower():
+            return True
+        if "false" == interpolated_value.lower():
+            return False
+        return None
```
dnephin: In the case where no casting happens, I think this function should just return the original value. That way there is no need to check for `None`. It can just return normally. It should simplify the callers logic a bit.

compose/config/interpolation.py
tiry: Yes, the point was to avoid doing the lookup in the json schema is anyway there is no interpolation.

tiry:
I'll try to integrate the feedback asap and will resubmit.

compose/config/interpolation.py
```diff
@@ -26,9 +29,46 @@ def process_item(name, config_dict):
...
+    elif expected_type == "array":
+        if isinstance(interpolated_value, list):
+            return interpolated_value
+        return json.loads("{{\"data\":{0}}}".format(interpolated_value.replace("'", "\"")))["data"]
+
```
dnephin: This seems overly complex.  I think supporting arrays is probably the most contentious part of this feature. I think we should try to keep it as straightforward as possible.

How about:

```python
return json.loads(interpolated_value)
```

Although, I wonder if it should actually be `yaml.safe_load()` since the config is yaml format. Using yaml would support json as well.

dnephin:
Thanks for the PR!   I think it's on the right track.

We'll need some time to discuss it for the 1.7 release, so it might be another couple weeks before we can give any more design feedback.

compose/config/interpolation.py
dnephin: I guess that doesn't hurt. I think it would be good to move this to the first line of `interpolate()`, that way `interpolate()` is the only function that knows or care about the variable format.  If we were to ever want to change to a different type of variable, the change would be contained to that one function.

compose/config/interpolation.py
tiry: As I said I have not written python for years, but it seems that `json.loads` does not work with an Array.
Looks like I am not the only one to face the issue 
=> http://stackoverflow.com/questions/10973614/convert-json-array-to-python-list

However, it may be just because of issue between simple quotes (') and double quotes (")

dnephin: I think that's it. `json` only supports double quotes for strings.

tiry:
What is the preferred process on your side : 

 - I amend my previous commit
 - I simply add more commits (and, may be, you squash then before merge)
 - I submit an other PR 

NB : I know this is a kind a "religious" question, so I prefer to ask before doing something inappropriate :)

compose/config/validation.py
tiry: the `if sys.platform == "win32":` was here in the code I moved ...

dnephin: Ah yes, I saw that, but I'm not sure why it's there either. This is mostly just a question, it may be necessary. If you can remove it and the appveyor tests pass (they run on windows) then I think it's safe to remove.

dnephin:
New commits are fine. You may be asked to squash at the end, right before merging.  Or you could also squash them now if you think it's warranted.

tiry:
Here is a new version.
I tried to incorporate most of the comments.

I say most, because while changing the code I realized that some cases where not handled, so I added more tests and adapted the implementation.
As a result, the `_cast_interpolated` is not as simple as I would have hoped ...

tiry:
Not sure I understand the problem on the Jenkins build: it is related to my changes or can it come from other changes ?

dnephin:
They look unrelated

tiry:
Ok, should I do something to restart the build / check ?

tiry:
Looks like I need to rebase my changes ... will do it when I find the time

flx42:
Any progress @tiry? Since this PR is solving #2750, it would be great for [nvidia-docker](https://github.com/NVIDIA/nvidia-docker). Thank you for your work!

tiry:
I did a rebase (+fixes) and everything seems to work on my side ... let's see what you CI says !

tiry:
It does not look like the build failure is related to my code.
Can someone restart the job (since I do not have the permission to do it myself on Jenkins), or should I do an empty commit ?

flx42:
I don't know how the CI works for `docker-compose`, but usually on GitHub you can just `git commit --amend --reset-author` and then `git push -f` (force push the modified commit on your branch) to trigger new checks.

tiry:
Fair enough, I was just looking for a cleaner option.

tiry:
Failed again so the error does not seem to be transient.
My understanding is that the problem is not related to my code change, if someone on `docker-compose` could take a look this would be helpful.

Thank you !

dnephin:
yes, the failures are unrelated, i've started another build now that the issue should be fixed.

tiry:
Great, let me know if you have further feedback !

tiry:
I could rebase once more, but I would need to have some feedback from you guys otherwise this is kind of pointless.

dnephin:
Yup, don't worry about rebasing just yet, we'll need to look over it again

ruffsl:
Any status on review of this PR? I'm here coming from the same nvidia-docker related camp (#2750 and https://github.com/NVIDIA/nvidia-docker/issues/39) as @flx42 .

shin-:
#5291

PR #2761, opened 2016-01-27
Remove ability to join bridge network + user-defined networks

aanand:
Incorporates #2757.

Containers connected to the bridge network can't have aliases, so it's simpler to rule that they can *either* be connected to the bridge network (via `network_mode: bridge`) *or* be connected to user-defined networks (via `networks` or the default network).


shin-:
LGTM!

aanand:
Cherry-picked into #2753

PR #2763, opened 2016-01-27
Add optional argument to pull dependencies on docker-compose pull.

kcboschert:
This adds the `--include-deps` argument to `docker-compose pull` to additionally pull all dependencies/links images for the specified SERVICE.

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -499,10 +499,12 @@ class TopLevelCommand(object):
 
         Options:
             --ignore-pull-failures  Pull what it can and ignores images with pull failures.
+            --include-deps          Pull dependencies.
         """
         self.project.pull(
             service_names=options['SERVICE'],
-            ignore_pull_failures=options.get('--ignore-pull-failures')
+            ignore_pull_failures=options.get('--ignore-pull-failures'),
+            include_deps=options.get('--include-deps'),
         )
 
     def rm(self, options):
--- a/compose/project.py
+++ b/compose/project.py
@@ -428,8 +428,8 @@ class Project(object):
 
         return plans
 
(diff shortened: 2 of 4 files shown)


dnephin:
Thanks for the PR.

Can you tell us a bit more about why this is necessary? Could you instead of `docker-compose pull --ignore-pull-failures` and just pull all the images?

kcboschert:
Dependencies used to be included by default with `docker-compose pull SERVICE`. It was reported as issue https://github.com/docker/compose/issues/1752, but the fix didn't provide a way to continue using the previous default.

`docker-compose pull --ignore-pull-failures` would attempt to pull all of the services in my compose.yml, while `docker-compose pull --include-deps SERVICE` would only pull the specified SERVICE and it's dependencies. The second test handles this case, but isn't very explicit. It's also ensuring that console's image is never pulled from https://github.com/docker/compose/blob/master/tests/fixtures/links-composefile/docker-compose.yml.

dnephin:
Thanks, this PR just needs to update the docs in `docs/reference/pull/md`, otherwise it's looking good

mtibben:
Do we anticipate this getting merged for the next release?

serkanyersen:
any update on this?

haplo:
This looks like a very straightforward change, what's preventing it from getting merged?

serkanyersen:
I've written a small script that can do this until this PR is merged. https://github.com/serkanyersen/dcpull

This script recursively finds all dependencies and calls `docker-compose pull` for each of them.

thiagofm:
Would be nice to have this merge, the current way it works is a bit counter intuitive.

prateeksarda:
when is the plan to have this feature under which release?

PR #2771, opened 2016-01-28
Stop `extends` from clobbering `depends_on` in the dependant service

rtlong:
Presently (using 1.6.0-rc2), this config:
```yaml
version: 2
services:
  foo:
    image: busybox
    command: echo foo

  bar:
    extends: foo
    command: echo bar
    depends_on:
      - biz

  biz:
    image: busybox
    command: echo biz
```
Results in the following resolved config after loading:
```yaml
networks: {}
services:
  bar:
    command: echo bar
    image: busybox
  biz:
    command: echo biz
    image: busybox
  foo:
    command: echo foo
    image: busybox
version: 2
volumes: {}
```
(It silently drops the `depends_on` set on `bar`.)

The fix I found is simple enough. I do think a test should be added to guard against this, but I'm very unfamiliar with both Python and the testing style used in this project, so I'd appreciate some guidance as to how to proceed to add a test.

### DIFF
--- a/compose/config/config.py
+++ b/compose/config/config.py
@@ -82,6 +82,7 @@ DOCKER_CONFIG_KEYS = [
 ALLOWED_KEYS = DOCKER_CONFIG_KEYS + [
     'build',
     'container_name',
+    'depends_on',
     'dockerfile',
     'expose',
     'external_links',


dnephin:
Thanks for the bug report. I think the correct behaviour here should actually be to error and say that `depends_on` can't be used with `extends` (it should work the same way as links).

I'll look to see why that isn't happening.

rtlong:
Actually, if I replace `depends_on` with `links` in my example, this does work as expected. You can't extend a service which itself defines links or dependencies but you can do this, currently. 

By the way, I'm curious: what is the technical reason for preventing one from extending a service with dependencies?

dnephin:
Ah yes, you're right. This fix looks good.

LGTM

dnephin:
Thanks for the fix, I noticed some related problems, so I'm going to carry this change in #2785

PR #2737, opened 2016-01-24
New navigation for 1.10 release

moxiegirl:
For the 1.10 release we are moving to product-name taxonomy to help users find documentation about a product.  This PR makes this menu:

image

Signed-off-by: Mary Anthony <mary@docker.com>

PR #2777, opened 2016-01-29
Re-order docs in menu, fix menu title capitalization, cleanup command reference

dnephin:
cc @moxiegirl 

First commit moves the quickstart guides before other docs
Second commit fixes capitalization (previously some were using lowercase)
Third commit tries to fix the command-line reference docs. The overview.md page was really just a reference to the environment variables, so I updated the title and page to match that (and added an alias).

### DIFF
--- a/docs/compose-file.md
+++ b/docs/compose-file.md
@@ -1,6 +1,6 @@
 <!--[metadata]>
 +++
-title = "Compose file reference"
+title = "Compose File Reference"
 description = "Compose file reference"
 keywords = ["fig, composition, compose, docker"]
 aliases = ["/compose/yml"]
--- a/docs/compose-file.md
+++ b/docs/compose-file.md
@@ -903,7 +903,8 @@ It's more complicated if you're using particular configuration features:
             syslog-address: "tcp://192.168.0.42:123"
 
 -   `links` with environment variables: As documented in the
-    [environment variables reference](env.md), environment variables created by
+    [environment variables reference](link-env-deprecated.md), environment variables
+    created by
     links have been deprecated for some time. In the new Docker network system,
     they have been removed. You should either connect directly to the
     appropriate hostname or set the relevant environment variable yourself,
--- a/docs/env.md
+++ b/docs/env.md
@@ -1,6 +1,6 @@
 <!--[metadata]>
 +++
-title = "Environment variables reference"
+title = "Environment Variables Reference"
 description = "Compose CLI reference"
 keywords = ["fig, composition, compose, docker, orchestration, cli,  reference"]
 [menu.main]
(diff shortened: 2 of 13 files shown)


aanand:
Great.

There might be confusion between `env.md` and `envvars.md`. Perhaps there should be a note at the top of each one explaining what it is and linking to the other one. We should perhaps rethink the URL slugs too.

dnephin:
The urls are actually `/compose/reference/envvars/` vs `/compose/envs`.

Why don't we move `envs.md` to `link-envs.md` and add an alias ?

dnephin:
I've made that change, and adjusted the menu titles, so it should be clear now. 

"Link environment variables" vs "CLI environment variables"

moxiegirl:
@dnephin Thanks for doing this. Can I get you to do one more thing? Can you put the content of `index.md` into the `overview.md` file and the content of `overview.md` into `index.md` --- I know its whack this is why I'm asking:

The current doc theme does not allow us to create a split menu: a menu whose label is both a link and a dropdown. Our theme redesign fixes this but it is in RFD, so earliest out is about March 2016. In the mean time

* the overview.md pages to show up in the menu.
* when a user clicks Overview, we will redirect go to https://docs.docker.com/compose/
* Later, after the redesign goes live, we will re-redirect in the unlikely event someone was able to book mark https://docs.docker.com/compose/overview

dnephin:
@moxiegirl this PR doesn't have an `overview.md` in `reference/` it's just an alias to `envvars.md`, but I think I see what you are saying.

If I understand correctly, somewhere in the docs build we will be adding redirects for every index.md to redirect to the  `./overview/` page, is that right?

I think what we have in `docs/reference/index.md` now is correct (it matches what we see at `docs/index.md`).

But to support the overview redirect I need to create a new page that can act as an overview to the CLI reference section.  I don't think we had an appropriate page for that before. I can restore the intro paragraph from the old overview.md page for that purpose.

dnephin:
Ok, I made that change. Let me know if that works

moxiegirl:
Hey Daniel, I'm very sorry for the slow reply -two new writers started today and Mondays are meetingful as it is. :-P

- You don't need to do anything to reference; this only applies to the top level directory
- In docs.docker.com/compose is going to the content of overview.md
- If we leave things as they are in this PR, the will be going to the next docs/index.md --- if you are cool with that; don't change anything at the docs/
- revert the work you did in reference

Slack me if you have questions. Sorry for the firedrill  --- this release yikes!

moxiegirl:
@dnephin Scott would like to  see "Docker Compose" in the top level menus.  Can you change the `title` on `docs/index.md` to be `Docker Compose` please?

dnephin:
> You don't need to do anything to reference; this only applies to the top level director

Well then I'm really confused because you just setup `index.md` and `overview.md` that way a few days ago in #2737. They seem correct to me the way they are now.

> the will be going to the next docs/index.md

Not sure what you mean.


I think I'm going to keep the changes to reference. I think it works better this way.

dnephin:
I'm updated the title to make it `Docker Compose`

dnephin:
I got a signoff from @moxiegirl over another channel

aanand:
Cherry-picked into #2765.

PR #2780, opened 2016-01-29
fix search order of project name

freeseacher:


dnephin:
Sorry, this is not correct. The environment variable should take precedence over the default.

If you're having a problem with the project name please open an issue.

Issue #2767, opened 2016-01-27
Show a better error when handling an unsupported file format version


aanand:
1.  If the user specifies `version: 1`, they've misunderstood how versioning works and we should show a helpful error. Right now, we show this:

        ERROR: In file './docker-compose.yml' service 'version' doesn't have any configuration options. All top level keys in your docker-compose.yml must map to a dictionary of configuration options.

2.  When the user specifies a version that we don't support (e.g. `2.1`), we should perhaps show an error along the lines of "version 2.1 of the file format isn't supported by this version of Compose", so that we're a little more future-proof. The downside is that old code can't know if the version the user has specified is an actual version that exists or is just mistyped. Thoughts?

dnephin:
I would omit "by this version of Compose", since that implies that it may be supported by some other version.

Something like "Unsupported compose file version: 2.1" 

Otherwise sounds good.

aanand:
Closed by #2783

plaindocs:
I'm getting this error, and don't fully understand it. 

I'm running a downloaded Dockerfile (https://github.com/danielquinn/paperless) on a Hypriot Raspberry PI (http://blog.hypriot.com/getting-started-with-docker-on-your-arm-device). 

Docker file contains:
    
    version: '2'

Docker version: 

    $ docker info
    Containers: 1
    Images: 12
    Server Version: 1.9.0

I'm not sure if that image runs on ARM, but that is what I'm trying to find out. Any clues?

PR #2783, opened 2016-01-29
Fix validation and version checking

### LINKED ISSUE #2766: Make version a string
It doesn't make sense for `version` to be a number: we likely want the next update to the file format to be `2.1`, not `3`, and we definitely don't want it to be a float.

aanand:
- Fields are now described as `web.image` or `services.web.image`, rather than `Service 'web' configuration key 'image'`
- Fixed bug where fields would come out as `Service 'services' configuration key 'web'` in v2 files
- `version` is now a string
- Show better error messages, point the user in the right direction

Closes #2766, #2767.


dnephin:
LGTM, 

Test failure is caused by merging a branch that added new test fixtures that used int versions.

aanand:
Updated to rectify that fixture. Only test failure is now in the unrelated `events` test.

aanand:
Cherry-picked into #2765.

PR #2784, opened 2016-01-29
Fix race condition with up and setting signal handlers

dnephin:
Fixes: https://github.com/docker/compose/pull/2679#issuecomment-174221963

Also print stdout on wait_for_container()
and adds a test for force shutdown

The old unit test is already covered by an acceptance test, so I removed it since the function no longer exists.


aanand:
LGTM

aanand:
Cherry-picked into #2765

PR #2785, opened 2016-01-29
Fix list options when used with extends and multiple files

dnephin:
From #2771

`depends_on` was being dropped when used with `extends`.

I noticed we were actually treating these list options differently, so this fixes them so they all act the same way.  All list options should be joined together, not replace.

When `merge_service_dicts()` is called to merge `extends` many of these values wont exist (because we validate that they aren't in the base in `validate_extended_service_dict()`. 

When `merge_service_dicts()` is called to merge files these options will now be joined correctly.

### DIFF
--- a/compose/config/config.py
+++ b/compose/config/config.py
@@ -26,6 +26,7 @@ from .sort_services import get_service_name_from_network_mode
 from .sort_services import sort_service_dicts
 from .types import parse_extra_hosts
 from .types import parse_restart_spec
+from .types import ServiceLink
 from .types import VolumeFromSpec
 from .types import VolumeSpec
 from .validation import match_named_volumes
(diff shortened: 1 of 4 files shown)


dnephin:
Hmm, I just realized that our docs actually say `links` and `volumes_from` are copied over like scalars. Why is that? It seems wrong

aanand:
I'm not sure. It does seem a bit wrong on the face of it.

In the case of `links`, if they're going to be combined, then we should make sure we handle alias collisions properly (the extending service's link should clobber the extended service's link).

dnephin:
I think that was previously an issue with `ports` as well. You could have overlapping port mappings if one uses a host ip, and the other only uses a host port. 

`volumes_from` also has an issue where you could have the same service but with different modes.

dnephin:
Ok, I've fixed `links` merging so that if the aliases match, only the override link will be used.

Also updated the docs for this change.

I don't think we need to worry about the ports/volumes_from case at this time, it hasn't really been an issue yet.

aanand:
Great. The `merge_sequence` stuff could perhaps use a unit test or two, but that's a small thing.

LGTM after rebase.

aanand:
Cherry-picked into #2765.

PR #2786, opened 2016-01-29
Fix a few bugs around networking and project initilization

dnephin:
* Volumes were not being created on `docker-compose run`, we were only calling `initialize_networks()`
* Networks were not included as part of the config_hash, so changes to the networks would not trigger a rebuild of a service
* Previously we were creating all networks, even if they weren't being used by any service. Now we only create networks used by at least one service, and warn if there are any that are not used.

Also moves some code out of `Project` into `ProjectNetworks` and `ProjectVolumes`. This allows us to construct the `networks` and `volumes` before the `Project`, and pass it into the constructor, instead of having to mutate the `Project` object.

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -693,7 +693,7 @@ def run_one_off_container(container_options, project, service, options):
                 start_deps=True,
                 strategy=ConvergenceStrategy.never)
 
-    project.initialize_networks()
+    project.initialize()
 
     container = service.create_container(
         quiet=True,
--- a/compose/network.py
+++ b/compose/network.py
@@ -133,12 +133,17 @@ class ProjectNetworks(object):
 
     @classmethod
     def from_services(cls, services, networks, use_networking):
-        networks = {
-            network: networks[network]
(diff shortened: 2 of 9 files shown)


aanand:
Nice. I guess if the config hash is already changing, we might as well change `net` to `network_mode` after all? Fine if that happens in a separate PR though.

I guess we never explicitly talked about whether networks that aren't in use should be created, but it sort of makes sense. If I `docker-compose up foo`, will only foo's networks (and its upstream dependencies' networks) be created? That would seem to follow logically.

LGTM

dnephin:
Hmm, that's not the case right now. The networks are filtered during project setup, not as part of the `up` command, so all defined networks are created.  We could still have `networks.initialize()` only create the required networks, but I think that is separate from this change.

Personally I think `docker-compose up foo`

dnephin:
I fixed the `if not /else`

aanand:
LGTM after rebase

aanand:
Cherry-picked into #2765.

PR #2795, opened 2016-01-31
issue #2794 - environment variable interpolation fails for $RANDOM

### LINKED ISSUE #2794: $RANDOM out of scope for environment variable interpolation.
My use case calls for a random integer to be used as a server id for a mysql replica setup. This can be passed via a `docker run` command using bash but not an option with docker-compose due to the nature of RANDOM not being listed in the static list of ENV. Naturally it's generated each call.

Docker example:

```
[moore@host test]$ docker run --rm busybox echo $RANDOM
6735
[moore@host test]$ docker run --rm busybox echo $RANDOM
25738
```

Compose example:
config...
_docker-compose.yml:_

```
randomTest:
   image: busybox
   container_name: rand
   command: echo $RANDOM
```

run...

```
$ docker-compose up
WARNING: The RANDOM variable is not set. Defaulting to a blank string.
Recreating rand
Attaching to rand
rand |
rand exited with code 0
```

Could be fixed by catching the instance of the RANDOM key before error and generate an appropriate random integer using python functionality therefore permitting $RANDOM to live in compose workflow too.

mysqlboy:
Fixes #2794

When using core docker we can instantiate a container interpolating $RANDOM (bash pseudo function) to obtain a runtime random integer value. This does not translate in compose because of python. The interpolation checks the key:values from the os.environ where $RANDOM doesn't persist.

My specific use case is mysql containers where the `server-id` requires a unique integer. Passing this to `command` reduces the amount custom, post-up scripting that needs to be performed. This could also prove useful in other cases.

Docker example:
```
[moore@host test]$ docker run --rm busybox echo $RANDOM
6735
[moore@host test]$ docker run --rm busybox echo $RANDOM
25738
```

Compose example
docker-compose.yml:
```
randomTest:
   image: busybox
   container_name: rand
   command: echo $RANDOM
```
test: (unpatched)
```
$ docker-compose up
WARNING: The RANDOM variable is not set. Defaulting to a blank string.
Recreating rand
Attaching to rand
rand |
rand exited with code 0
```
test: (patched)
```
$ docker-compose up
Recreating rand
Attaching to rand
rand | 1664
rand exited with code 0

$ docker-compose up
Recreating rand
Attaching to rand
rand | 8547
rand exited with code 0
```

### DIFF
--- a/compose/config/interpolation.py
+++ b/compose/config/interpolation.py
@@ -68,7 +68,7 @@ class BlankDefaultDict(dict):
     def __getitem__(self, key):
         if key == "RANDOM":
             from random import randint
-            rand = (randint(2,9999))
+            rand = (randint(2, 9999))
             return rand
         else:
             try:
--- a/compose/config/interpolation.py
+++ b/compose/config/interpolation.py
@@ -66,17 +66,22 @@ class BlankDefaultDict(dict):
         self.missing_keys = []
 
     def __getitem__(self, key):
-        try:
-            return super(BlankDefaultDict, self).__getitem__(key)
-        except KeyError:
-            if key not in self.missing_keys:
-                log.warn(
-                    "The {} variable is not set. Defaulting to a blank string."
-                    .format(key)
-                )
-                self.missing_keys.append(key)
+        if key == "RANDOM":
+            from random import randint
+            rand = (randint(2,9999))
+            return rand
+        else:
+            try:
+                return super(BlankDefaultDict, self).__getitem__(key)
+            except KeyError:
+                if key not in self.missing_keys:
+                    log.warn(
+                        "The {} variable is not set. Defaulting to a blank string."
+                        .format(key)
+                    )
+                    self.missing_keys.append(key)
 
-            return ""
+                return ""
 
 
 class InvalidInterpolation(Exception):


dnephin:
I would suggest using the container name for the unique id. You can get it using `$(hostname)` inside the container.

You can pretty easily use `$RANDOM` from inside the container as well. I'm not sure this is something we want to support in the variable interpolation.

mysqlboy:
I get it's niche. For my use case the container name would not suit since mysql's server-id is not an integer. Yes, $RANDOM can generally be called inside yet there's a beauty of calling

docker-compose scale master=1 replica=n to instantiate a larger collection of replicas in a cold command. There are workarounds but this helps no doubt. 

Of course it's not pretty and I understand if there's no desire to maintain this.

dnephin:
I've actually used the container id to generate a server-id for mysql before. I believe I just translated it into a base 10 number (the container id is a base 16 (aka hex) int), then shifted off a few of the bytes (because it was too large).

I think it was something like this:

```
echo "$((0x$(hostname) >> 24))"
```

dnephin:
Thanks for the PR. As discussed this isn't something we're interested in supporting. There are many ways to get a random number, but using the container id is generally the best way if you need it to relate to the container in some way.

While our variable interpolation looks very similar to bash variables it is not meant to be a feature compatible implementation of what bash supports.

referup-tarantegui:
I found that is very interesting set a hostname for the container with a $RANDOM variable. With this feature I can set a prefix and a random part to avoid duplicates. For example:

```
apache-$RANDOM
```

At the moment, I cannot set this variable, docker-compose tells me that this variable is empty.

mysqlboy:
@referup-tarantegui this is because $RANDOM is not a persisted variable it's the result of a function. Look at env and you'll see it's not there or blank (I don't recall). My patch can address it but then you're working wide of a supportable docker-compose binary.

PR #2802, opened 2016-02-01
Deploying to bintray from appveyor using the new bintray support

### LINKED ISSUE #2640: Windows master builds broken on bintray
The artifacts built by appveyor are fine, but on bintray they are 3 bytes.

Something about the deployment from appveyor to bintray is broken. Could be something appveyor changed, I haven't had a chance to look into it yet.

dnephin:
Fixes #2640

Tested with my compose fork and my own bintray bucket. The uploaded file is the correct size.

Using this new config, the deployment is configured in appveyor instead of in the `appveyor.yml`. There may be a way to configure it from the file as well, but the docs aren't up yet for how to do that.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -9,20 +9,16 @@ install:
 # Build the binary after tests
 build: false
 
-environment:
-  BINTRAY_USER: "docker-compose-roleuser"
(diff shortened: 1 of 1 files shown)


aanand:
LGTM

PR #2806, opened 2016-02-02
Run test containers in TTY mode

aanand:
This enables coloured output for tox and pytest.


dnephin:
Normally jenkins fails to run containers with tty enabled, but I guess since these run inside another container it works.

LGTM

PR #2808, opened 2016-02-02
Connect container to networks before starting it

### LINKED ISSUE #2805: Cannot resolve parent container IP address - with depends_on in config
I've an issue with the latest rc2 release and the depends_on feature. My config looks like this:
[docker-compose.yml](https://github.com/docker/compose/files/113802/docker-compose.yml.txt)

Getting error:
> front.example.org           | Message from VCC-compiler:
> front.example.org           | Backend host '"www.development.example.org"' could not be resolved to an IP address:
> front.example.org           | 	Name or service not known

So apparently the front.development.example.org container starts before it's able to resolve the host www.development.example.org to an IP address.
(I've attempted to manually sleep for two secs before starting front main application, and then it works fine.)

aanand:
Will hopefully fix #2805.

### DIFF
--- a/compose/service.py
+++ b/compose/service.py
@@ -424,8 +424,8 @@ class Service(object):
             return self.start_container(container)
 
     def start_container(self, container):
-        container.start()
         self.connect_container_to_networks(container)
+        container.start()
         return container
 
     def connect_container_to_networks(self, container):


dnephin:
LGTM

dnephin:
Actually, I take that break, it seems to be broken this way, I tried it with my setup, and I'm seeing:

```
     "Networks": {
                "cs_default": {
                    "IPAMConfig": null,
                    "Links": null,
                    "Aliases": null,
                    "NetworkID": "974b0e26f2746c5399cb6824321f7aa18eedfd20c567c4e100403a295dd35766",
                    "EndpointID": "",
                    "Gateway": "",
                    "IPAddress": "",
                    "IPPrefixLen": 0,
                    "IPv6Gateway": "",
                    "GlobalIPv6Address": "",
                    "GlobalIPv6PrefixLen": 0,
                    "MacAddress": ""
                }
            }

```
and they fail to connect to other containers.

mavenugo:
@dnephin I tried it in the engine directly -

```
  $ sudo docker create --net=temp1 --name=container1 -t busybox
88c3f1c424d71b7698468c3493161f5d3045579d478e9f05d5aef2fe6268a910
  $ sudo docker network disconnect temp1 container1
  $ sudo docker network connect --alias=c1 temp1 container1
  $ sudo docker start -i container1
  $ docker inspect container1

            "Networks": {
                "temp1": {
                    "IPAMConfig": null,
                    "Links": null,
                    "Aliases": null,
                    "NetworkID": "14399b5bf43d5754c648609844ac0ef5ffe4e0c57b85693e4c70396f85915848",
                    "EndpointID": "d90799fa3f3cf869c582ab2fca5d99bb5a51bd2e2355620acb591c30e5e4045d",
                    "Gateway": "172.22.0.1",
                    "IPAddress": "172.22.0.2",
                    "IPPrefixLen": 16,
                    "IPv6Gateway": "",
                    "GlobalIPv6Address": "",
                    "GlobalIPv6PrefixLen": 0,
                    "MacAddress": "02:42:ac:16:00:02"
                }
            }
```

It seems to be working fine.

aanand:
@mavenugo I'm seeing an issue where there are no aliases or links if I connect before starting:

```console
$ docker network create test-network
$ docker create --net=test-network --name=test-container busybox top
$ docker network disconnect test-network test-container
$ docker network connect --link=foo:bar --alias=baz test-network test-container
$ docker start test-container
$ docker inspect test-container
...
            "Networks": {
                "test-network": {
                    "IPAMConfig": null,
                    "Links": null,
                    "Aliases": null,
                    "NetworkID": "5278aeed659d70b6384625491cbb0be8ad5772052ebef75870520d81dfd3b44c",
                    "EndpointID": "7660cef24511060803159e11512a937830c4d7322d2ad2defb82299c3bf41422",
                    "Gateway": "172.20.0.1",
                    "IPAddress": "172.20.0.2",
                    "IPPrefixLen": 16,
                    "IPv6Gateway": "",
                    "GlobalIPv6Address": "",
                    "GlobalIPv6PrefixLen": 0,
                    "MacAddress": "02:42:ac:14:00:02"
                }
            }
...
```

I expect `Links` to be `["foo:bar"]` and `Aliases` to be `["baz"]`.

mavenugo:
@aanand ahh... @aboch resolved it yesterday https://github.com/docker/docker/pull/19943 and being this late in the release, we are not going to stop the engine release for this one.

aanand:
I presume the plan is for that fix go out in Engine 1.10.1? If so, we'll have to fix it in Compose 1.6.1, but the problem is that people will have to upgrade Compose in lockstep with Engine or their apps will suddenly break due to having no links or aliases.

That's going to be a terrible user experience, so I advocate strongly for getting the fix in for 1.10.0.

mavenugo:
@aanand so what is the downside of getting this PR in now in compose (knowing well that the short-id alias will not be available). When 1.10.1 is out, that will be fixed and hence there wont be a compatibility matrix issue ?

dnephin:
@mavenugo this PR does not work. Merging it would cause every v2 compose file to fail. Our tests are failing, and testing it manually has shown that it does not work.

mavenugo:
@dnephin my question is not that. My question is about the impact of not having the short-id as an `--alias` till 1.10.1

aanand:
Let's suppose we drop the short id alias.

We need to set the service name alias (`web`, `db`). We can now set that alias at container creation time *for one network*, but a container can potentially be connected to multiple networks, so we would still need to perform a series of `/connect` calls. If we do those before starting the container, then with Engine in its current state, the aliases won't be set. If we do them after, we get the race condition described in #2805.

In conclusion, if https://github.com/docker/docker/pull/19943 doesn't make it into 1.10.0, we have two choices on the Compose side:

1.  Don't fix the race condition. When containers start, they cannot expect hostnames for services to resolve immediately, and must back off and retry.

    When the fix lands in a subsequent Engine 1.10.1, we have to decide again whether to fix the race condition in Compose 1.6.1. If we do, then Compose 1.6.1 will break with Engine 1.10.0 - no aliases will be set. Users must upgrade Engine as well as Compose, or their apps will suddenly stop working.

    We could check the Engine version and refuse to run if it's <1.6.1, but that requires an extra API call on almost every Compose command. I've advocated not doing any pre-emptive version checking for a long time for that exact reason.

2. Fix the race condition, making aliases non-functional for people using multiple networks.

To my eye Option 1 looks bad and Option 2 looks unacceptable.

dnephin:
This sounds right to me. I think both options are pretty bad.

aanand:
OK, https://github.com/docker/docker/pull/19943 has been merged and I've confirmed locally that it fixes the tests, so once Engine RC4 is out then CI should pass for this and we should be able to merge it.

dnephin:
LGTM

seems to be working with manual test as well with docker 1.10-rc4

PR #2822, opened 2016-02-04
Update docker-py and dockerpty

aanand:
This won't pass until those versions are released, of course.


dnephin:
LGTM once it's green

aanand:
Cherry-picked into #2821.

PR #2826, opened 2016-02-04
Bump 1.7.0dev

aanand:


docs/install.md
```diff
@@ -18,7 +18,7 @@ first.
 
 To install Compose, do the following:
 
-1. Install Docker Engine version 1.7.1 or greater:
+1. Install Docker Engine:
```
aanand: The minimum Engine version now depends on the Compose file version you're using, so I thought it simpler to just omit it here rather than explain too much up front. Most people, if installing Compose, will probably be installing Engine too, and so they'll get the latest version anyway. Sound good?

dnephin: Ya, I like it

dnephin:
Build failed on missing dind image, I'm building it now.
Don't think we need to wait on it for this change.

LGTM

PR #2830, opened 2016-02-05
Add support for shm_size.  Fixes #2823.

### LINKED ISSUE #2823: Missing support for --shm-size
Support for a configurable /dev/shm size was added in docker/docker#16168 as part of Docker 1.10 (API version 1.22).  I am currently running into an issue where I need a larger /dev/shm size in a project that is managed by docker-compose and I am unable to configure it when running containers via docker-compose.

It should be similar to fields like `mem_limit` for actual implementation.

nubs:
shm_size controls the size of /dev/shm in the container and requires Docker 1.10 or newer (API version 1.22).  This requires docker-py 1.8.0 (docker/docker-py#923).

Similar to fields like `mem_limit`, `shm_size` may be specified as either an integer or a string (e.g., `64M`).

Updating docker-py to the master branch in order to get the unreleased dependency on `shm_size` there in place.

Signed-off-by: Spencer Rinehart <anubis@overthemonkey.com>

### DIFF
--- a/compose/service.py
+++ b/compose/service.py
@@ -57,6 +57,7 @@ DOCKER_START_KEYS = [
     'volumes_from',
     'security_opt',
     'cpu_quota',
+    'shm_size',
 ]
 
 
@@ -654,6 +655,7 @@ class Service(object):
             ipc_mode=options.get('ipc'),
             cgroup_parent=options.get('cgroup_parent'),
             cpu_quota=options.get('cpu_quota'),
+            shm_size=options.get('shm_size'),
         )
 
     def build(self, no_cache=False, pull=False, force_rm=False):
--- a/compose/service.py
+++ b/compose/service.py
@@ -40,6 +40,7 @@ DOCKER_START_KEYS = [
     'cap_add',
     'cap_drop',
     'cgroup_parent',
+    'cpu_quota',
     'devices',
     'dns',
     'dns_search',
@@ -54,10 +55,9 @@ DOCKER_START_KEYS = [
     'pid',
     'privileged',
     'restart',
-    'volumes_from',
     'security_opt',
-    'cpu_quota',
     'shm_size',
+    'volumes_from',
 ]
 
 
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,9 +1,9 @@
 PyYAML==3.11
 cached-property==1.2.0
-docker-py==1.7.0
 dockerpty==0.4.1
 docopt==0.6.1
 enum34==1.0.4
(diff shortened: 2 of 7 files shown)


compose/service.py
```diff
@@ -57,6 +57,7 @@
     'volumes_from',
     'security_opt',
     'cpu_quota',
+    'shm_size',
```
dnephin: This list is supposed to be in sorted order.

I see now that the last couple additions were not adding in order. Would you mind running a `:sort` on this section to move the `shm_size` into the right place (and the other two at the same time).

dnephin:
Looks good, just one small request

compose/service.py
nubs: Sure thing, will do.

nubs:
Repushed with sorted DOCKER_START_KEYS.

requirements.txt
```diff
@@ -1,9 +1,9 @@
 PyYAML==3.11
 cached-property==1.2.0
-docker-py==1.7.0
 dockerpty==0.4.1
 docopt==0.6.1
 enum34==1.0.4
+git+https://github.com/docker/docker-py.git@master#egg=docker-py
```
dnephin: Oh, I missed this before, please pin to a git sha instead of a branch. That way we don't accidentally pick up other docker-py changes.

dnephin:
Thanks for sorting.

LGTM with the one small changes to ping sha instead of a branch.

Looks like our jenkins CI is really backed up this morning because of unrelated issues. Might be a little while before the tests run.

nubs:
Updated with hash instead of branch for docker-py.  Chose the hash of the merge commit for my change upstream.

aanand:
LGTM

tests/integration/service_test.py
```diff
@@ -102,6 +102,12 @@ def test_create_container_with_cpu_quota(self):
         container.start()
         self.assertEqual(container.get('HostConfig.CpuQuota'), 40000)
 
+    def test_create_container_with_shm_size(self):
+        service = self.create_service('db', shm_size=67108864)
```
dms-cat: Isn't this the default value? `shm_size` doesn't seem to work - the following configuration still returns `"ShmSize": 67108864`:

```
  db:
    image: postgres:11.1
    shm_size: '2gb'
```

Ditto for `shm_size: 2147483648`.

dms-cat: [Others](https://stackoverflow.com/q/55416904/96588) seem to have the same problem.

PR #2832, opened 2016-02-05
Refactor CLI initialization and unit tests

aanand:
A series of refactors, roughly ordered with the most "quick fix"/"worth it" ones first. In particular, I'm not sure the refactors to the code in the last one are very good - it could use some fresh eyes.


dnephin:
First 5 commits look good, I've got some comments about the last.

aanand:
Rebased and removed the last two commits since it's not a great approach.

dnephin:
LGTM when build is green

PR #2851, opened 2016-02-08
Fix merge behavior for 'networks' and 'network_mode'.

### LINKED ISSUE #2839: compose silently ignore networks when extends or multiple compose files
Compose 1.6 silently ignore the "networks" option when the service extends from another one.

This version 2 docker file :
```YAML
version: '2'
services:
  foo:
    image: busybox
    command: echo foo
  bar:
   extends: foo
   command: echo bar
   networks:
      - mynetwork
networks:
  mynetwork:
```

Results in the following resolved config after loading:
 
```YAML
networks:
  mynetwork: {}
services:
  bar:
    command: echo bar
    image: busybox
  foo:
    command: echo foo
    image: busybox
version: '2.0'
volumes: {}
```

The "bar" container is not attach to "mynetwork".

### LINKED ISSUE #2840: Using docker-compose with multiple configuration files disables network_mode: "host" for services
When using docker-compose (v1.6) with multiple configuration files with "-f" like the following:

docker-compose -f docker-compose.yml -f preproduction.yml build
docker-compose -f docker-compose.yml -f preproduction.yml up -d

If services are using network_mode: "host" then this configuration is disabled (dropping to "bridge" mode)

cr7pt0gr4ph7:
As stated in #2839 and #2840, the `networks` and `network_mode` service configuration keys are ignored when multiple compose files are specified on the command line (or when extending services via `extends`).

The reason for this is that the relevant `merge_service_dicts` function in `compose/config/config.py` does not handle these configuration keys.
This simply enables handling of `network_mode` and `networks` via the existing mechanism, and additionally adds a test for the merge behavior of `networks`.

Fixes #2839
Fixes #2840

### DIFF
--- a/tests/unit/config/config_test.py
+++ b/tests/unit/config/config_test.py
@@ -1594,30 +1594,64 @@ class BuildOrImageMergeTest(unittest.TestCase):
         )
 
 
-class MergeListsTest(unittest.TestCase):
+class MergeListsTest(object):
+    def config_name(self):
+        return ""
+
+    def base_config(self):
+        return []
+
+    def override_config(self):
+        return []
+
+    def merged_config(self):
+        return set(self.base_config()) | set(self.override_config())
+
     def test_empty(self):
-        assert 'ports' not in config.merge_service_dicts({}, {}, DEFAULT_VERSION)
+        assert self.config_name() not in config.merge_service_dicts({}, {}, DEFAULT_VERSION)
 
     def test_no_override(self):
         service_dict = config.merge_service_dicts(
-            {'ports': ['10:8000', '9000']},
+            {self.config_name(): self.base_config()},
             {},
             DEFAULT_VERSION)
-        assert set(service_dict['ports']) == set(['10:8000', '9000'])
+        assert set(service_dict[self.config_name()]) == set(self.base_config())
 
     def test_no_base(self):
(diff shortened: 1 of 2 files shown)


tests/unit/config/config_test.py
```diff
@@ -1594,30 +1594,64 @@ def test_merge_build_or_image_override_with_other(self):
...
+
+class MergeNetworksTest(unittest.TestCase, MergeListsTest):
+    def config_name(self):
+        return 'networks'
+
+    def base_config(self):
+        return ['frontend', 'backend']
+
+    def override_config(self):
+        return ['monitoring']
```
dnephin: I think we'll eventually want to use http://pytest.org/latest/parametrize.html to support this type of test, instead of inheritance, but that doesn't need to happen now.

For now could we make these class fields instead of functions? Since their value doesn't change I think that will work.

```python

class MergeNetworksTest(unittest.TestCase, MergeListsTest):
    config_name = 'networks'
    base_config = ['frontend', 'backend']
    override_config = ['monitoring']

```

dnephin:
Thanks for the fix! I think the code changes look good, just one request for the tests.

dnephin:
LGTM

Thanks for the fix!

aanand:
LGTM

PR #2853, opened 2016-02-08
Use 12 characters for the short id

### LINKED ISSUE #2845: Container linking doesn't work reliably under docker-compose 1.6
Environment: docker 1.10, docker-compose 1.6, Ubuntu 15.10

Background:

I've just updated docker to 1.10 and docker-compose to 1.7 on my Ubuntu 15.10 box. I am using around 7-8 containers linked with docker-compose, and it was working perfectly under docker-compose 1.6, but with 1.7, some containers don't see other (linked) containers. This seems to be related to the changed name resolution (changed from /etc/hosts to the /etc/resolv.conf-based approach).

Let's say I have containers named A and B. B is linked to A, and the Java code running inside the container wants to connect to 'A'. B starts up, I am getting exceptions saying "cannot connect to ". Note that is the random (?) ID of the container, not its name.

1) How does B know the ID of A? Somehow the DNS must return this ID when trying to look up 'A'.
2) If the DNS returns this ID, then why can't B connect to A through its ID?

So somehow, something must have regressed between 1.6 and 1.7.

Note that linking works for some containers (always), and doesn't for some others (always).

dnephin:
Fixes #2845

Ref: https://github.com/docker/docker/blob/master/pkg/stringid/stringid.go

Looks like we've used 10 characters for short_id since the tool was named plum, but docker uses 12 characters.

### DIFF
--- a/compose/container.py
+++ b/compose/container.py
@@ -60,7 +60,7 @@ class Container(object):
 
     @property
     def short_id(self):
-        return self.id[:10]
+        return self.id[:12]
 
     @property
     def name(self):
--- a/tests/unit/container_test.py
+++ b/tests/unit/container_test.py
@@ -12,8 +12,9 @@ from compose.container import get_container_name
 class ContainerTest(unittest.TestCase):
 
     def setUp(self):
+        self.container_id = "abcabcabcbabc12345"
         self.container_dict = {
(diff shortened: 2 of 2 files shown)


shin-:
LGTM

PR #2858, opened 2016-02-08
Fix upgrading url in changelog

dnephin:
Reported by someone on twitter


dnephin:
Technically this link doesn't work either, because for some reason the hash urls on all the pages gained some weird serial id at the end, but assuming that is fixed, this should be correct.

aanand:
LGTM

PR #2859, opened 2016-02-09
Fix header type for external volume

JacobHayes:


dnephin:
Thanks! We already have a PR open to fix this: #2827

JacobHayes:
Oh, great! Sorry for the double pr :)

PR #2880, opened 2016-02-10
Merge jsonschemas to a single schema file per version

### LINKED ISSUE #2704: Merge `fields_schema_vx.json` and `service_schema_vx.json`
Currently we split the config schema into two files, one for the "top level" configuration and the other for the service definition. We can actually put everything into the same file and still do multi-stage validation.

1. Move `#/definitions/service` into `fields_schema_vx.json`
2. Move `#/definitions/constraints` into `fields_schema_vx.json` (and rename it to service-constraints or something like that)
4. Rename `fields_schema_vx.json` to `config_schema_vx.json`, and remove `service_schema.json`
5. Create a new definition called something like `top-level-config` which contains the top level properties (`services`, `networks`, `volumes`, and `version`) with `type: object`, without any validation of the properties.
3. Move the current root schema properties into a new definitions, one for each type `networks`, `volumes`, `services`.
4. Change the top level schema to be `allOf: ['top-level-config', 'networks', 'volumes', 'services']`

In the compose validation code we would never validate the root schema directly. First we validate against the `top-level-config` schema to make sure all the keys exist and are of the right type. Then as we validate each section we use the individual definitions. This allows us to raise an error which contains the name of the item (service name, volume name, etc) without having to guess at where the name is in within the schema.

I think this should block #2089

dnephin:
Fixes #2704 

I always found it really difficult to reason about the purpose of each of the schemas. They're now a single schema, and the previous validation functions are renamed to:
* `validate_against_config_schema` - which performs the jsonschema validation of the structure
* `validate_service_constraints` - which performs the constraint validation. The code to format the errors here is actually longer than it would be to just apply these constraints in python, but it is kind of nice that we can express the constraint from a schema, so I left them in.

This should make it easier to formalize the schema, because there is now a single schema for each version.


dnephin:
Test failure is just a flake , tests passed correctly in 3/4 runs

dnephin:
ping, this is rebased and green!

aanand:
LGTM

PR #2884, opened 2016-02-11
Updating Dockerfile

moxiegirl:
- remove tutorials, that repo is now private
- add notary as that component is now building
- add toolbox as that component is now building
- change ownership of kitematic to docker as that repo moved

@dnephin


moxiegirl:
author LGTM

dnephin:
LGTM,

Although i'm starting to think we should be doing this differently. There should probably be an automated build of this from the some repo, and locally we can just override the project specific, instead of having to update every project anytime one changes.

moxiegirl:
@dnephin Yes! I was just thinking of just this the other day.  I'll look into doing it as soon as UCP is out the door.  Sven should be working on other stuff.

moxiegirl:
@dnephin The beauty of doing it with images --- aside from the processing and lack of svn, is we should be able to give the "entire web" not just the oss projects.

PR #2887, opened 2016-02-11
Allow running 'docker-compose run' without ptty on Windows

mrkschan:
This patch enables running `docker-compose run -T` on Windows.

Signed-off-by: KS Chan <mrkschan@gmail.com>

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -458,11 +458,12 @@ class TopLevelCommand(DocoptCommand):
         """
         service = project.get_service(options['SERVICE'])
         detach = options['-d']
+        no_ptty = options['-T']
 
-        if IS_WINDOWS_PLATFORM and not detach:
+        if IS_WINDOWS_PLATFORM and (not detach and not no_ptty):
             raise UserError(
                 "Interactive mode is not yet supported on Windows.\n"
-                "Please pass the -d flag when using `docker-compose run`."
+                "Please pass either -d or -T flag when using `docker-compose run`."
             )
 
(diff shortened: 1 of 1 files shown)


mrkschan:
I'm not sure why `-T` is not allowed even though I can run `docker-compose run -T --rm django ls` successfully on Windows10.

dnephin:
LGTM

aanand:
This warrants further investigation because, to the best of my knowledge, `dockerpty` should fail to run on Windows (indeed, we don't even import it if we detect we're running on Windows).

@mrkschan: do you get output when you run that command? What happens if you don't pass `-T`? What shell are you using?

mrkschan:
@aanand I use babun https://babun.github.io/ with https://github.com/tiangolo/babun-docker enabled.

I'm not sure if you're talking about running `dockerpty`.

```
{ ~ }  » dockerpty                                                                    ~
zsh: command not found: dockerpty
```

mrkschan:
@aanand I got what you're saying now. What I did was `cd compose/ && pip install -U .` and then try to run `docker-compose`. That allowed me to run with `dockerpty`.

When I tried to build the binary using powershell via `scripts/build-windows.ps1`, the executable would complain missing `RunOperation`.

I now try to import `RunOperation` from `dockerpty` and see if a new executable would work on Windows.

mrkschan:
Closing this since we need dockerpty ported to windows in order to have `RunOperation`. I would stick to my old path: `docker-compose run -d ...` and `docker logs -f ...` instead.

PR #2888, opened 2016-02-11
Upgrade pyinstaller.

### LINKED ISSUE #2886: Dealing with noexec /tmp
When running docker-compose with a /tmp filesystem that is set to noexec in fstab docker-compose fails with error message similar to: "error while loading shared libraries: libz.so.1".
In earlier versions of docker-compose we have worked-around this by setting TMP environment-variable to point to some other directory where noexec is not set, but this doesn't work anymore in 1.6.0.
Is there a suggested workaround for this?

dnephin:
Fixes #2886 (hopefully)

### DIFF
--- a/requirements-build.txt
+++ b/requirements-build.txt
@@ -1,1 +1,1 @@
-pyinstaller==3.0
+pyinstaller==3.1.1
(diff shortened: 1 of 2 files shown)


shin-:
LGTM

PR #2893, opened 2016-02-12
Bring up all dependencies when running a single service.

### LINKED ISSUE #2833: run not looking at depends_on
I think that `docker-compose run <service>` should bring up the dependencies of that service, the same as `docker-compose up <service>` would. There is already the `--no-deps` flag for run, which implies that the default behavior is to bring up dependencies.

If this is the intended behavior, then I am not seeing it. If I have a `docker-compose.yml` file like this:

```yaml
version: '2'
services:
  web:
    image: postgres
    command: sh -c 'psql -h db -p 5432 -U postgres -d postgres -c "\list"'
    entrypoint: /tmp/entrypoint.sh
    volumes:
      - ./entrypoint.sh:/tmp/entrypoint.sh
    depends_on:
       - db
  db:
    image: postgres
```

Then running `docker-compose run web` won't bring up the `db` container.

shin-:
Fixes #2833

### DIFF
--- a/tests/acceptance/cli_test.py
+++ b/tests/acceptance/cli_test.py
@@ -738,6 +738,15 @@ class CLITestCase(DockerClientTestCase):
         self.assertEqual(len(db.containers()), 1)
         self.assertEqual(len(console.containers()), 0)
 
+    @v2_only()
+    def test_run_service_with_dependencies(self):
+        self.base_dir = 'tests/fixtures/v2-dependencies'
+        self.dispatch(['run', 'web', '/bin/true'], None)
+        db = self.project.get_service('db')
+        console = self.project.get_service('console')
+        self.assertEqual(len(db.containers()), 1)
+        self.assertEqual(len(console.containers()), 0)
+
     def test_run_with_no_deps(self):
         self.base_dir = 'tests/fixtures/links-composefile'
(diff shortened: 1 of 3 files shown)


tests/acceptance/cli_test.py
```diff
@@ -738,6 +738,14 @@ def test_run_service_with_links(self):
         self.assertEqual(len(db.containers()), 1)
         self.assertEqual(len(console.containers()), 0)
 
+    def test_run_service_with_dependencies(self):
+        self.base_dir = 'tests/fixtures/v2-dependencies'
```
dnephin: I think you forgot to `git add` this ?

shin-: Yup, I had. Fixed!

tests/acceptance/cli_test.py
```diff
@@ -738,6 +738,14 @@ def test_run_service_with_links(self):
         self.assertEqual(len(db.containers()), 1)
         self.assertEqual(len(console.containers()), 0)
 
+    def test_run_service_with_dependencies(self):
```
dnephin: I think this needs the `@v2_only()` decorator

dnephin:
LGTM

saulshanabrook:
Thanks @dnephin this is great :+1:

PR #2894, opened 2016-02-12
Only set a container affinity if there are volumes to copy over

### LINKED ISSUE #2866: Compose cannot recreate a container on Swarm if the container has an host port mapping
## What I have
- 5 nodes cluster
- Compose 1.6.0
- Swarm 1.1.0
- Engine 1.10.0
## What I do

With the following `docker-compose.yml`:

``` yaml
www:
  image: nginx
  ports:
    - "8888:80"
```

I run:

```
docker-compose up -d
docker-compose up -d --force-recreate
```
## What I expect to see

Container is created, then recreated.
## What I see instead

Container is created, but when trying to recreate it, I see:

```
Recreating repro_www_1
ERROR: unable to find a node that satisfies container==a0e31139c706d24a739b88034e1f0b266eb049b6b304580c3b29bf4535d1c958
```

(Where `a0e3...` is the ID of the existing container.)
## Cause

Compose provides a placement constraint to force the new container to be on the same host as the old one. But when there is a port mapping, this fails, even if the container has been stopped first, because Swarm doesn't deal with port mappings the same way as Engine does (a stopped container is still considered to use the port).

I attached the logs of `docker inspect` (before recreating the container) and the verbose output of `docker-compose up -d`.

[inspect.log.txt](https://github.com/docker/compose/files/123408/inspect.log.txt)
[compose.log.txt](https://github.com/docker/compose/files/123407/compose.log.txt)

dnephin:
Partially fixes #2866

Instead of always setting a container affinity, only set one if there are volumes to copy over from a previous container.

### DIFF
--- a/compose/service.py
+++ b/compose/service.py
@@ -592,20 +592,19 @@ class Service(object):
                     ports.append(port)
             container_options['ports'] = ports
 
-        override_options['binds'] = merge_volume_bindings(
-            container_options.get('volumes') or [],
-            previous_container)
-
-        if 'volumes' in container_options:
-            container_options['volumes'] = dict(
-                (v.internal, {}) for v in container_options['volumes'])
-
         container_options['environment'] = merge_environment(
             self.options.get('environment'),
             override_options.get('environment'))
 
-        if previous_container:
-            container_options['environment']['affinity:container'] = ('=' + previous_container.id)
+        binds, affinity = merge_volume_bindings(
+            container_options.get('volumes') or [],
+            previous_container)
+        override_options['binds'] = binds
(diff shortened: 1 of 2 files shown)


aanand:
Nice. Will it set affinity if there's an explicit volume bind to a host path or named volume?

dnephin:
The affinity is not set for host volumes or named volumes (they're both excluded by `get_container_data_volumes()` because of the `if volume.external` check). I believe this is the correct behaviour.

I added a named volume to an existing test to confirm named volumes.

dnephin:
rebased

aanand:
LGTM

PR #2903, opened 2016-02-12
driver_opts can only be of type string

### LINKED ISSUE #2899: driver parameter ignored in volumes section?
I'm having an issue using rexray driver with docker-compose 1.6.
I have following very simple file:
```
version: "2"
services:
  mysql:
    image: sameersbn/mysql:latest
    volumes:
      - mydata:/var/lib/mysql

volumes:
  mydata:
    driver: "rexray"
    driver_opts:
      volumetype: "gp2"
      size: 5
```

Given that file I suppose, that DE will create volume with Rexray volume driver?
But really volume is created with 'local' driver instead.
And I even can enter _any_ string as driver parameter, and volume will be created with 'local' driver.
But if I describe volume as external, provided existing rexray volume name, it will be successfully created and mounted.

Is it a bug or I'm doing something wrong?

shin-:
Fixes #2899

### DIFF
--- a/tests/unit/config/config_test.py
+++ b/tests/unit/config/config_test.py
@@ -231,6 +231,20 @@ class ConfigTest(unittest.TestCase):
         assert volumes['simple'] == {}
         assert volumes['other'] == {}
 
+    def test_volume_invalid_driver_opt(self):
+        config_details = build_config_details({
+            'version': '2',
+            'services': {
+                'simple': {'image': 'busybox'}
+            },
+            'volumes': {
+                'simple': {'driver_opts': {'size': 42}},
+            }
+        })
+        with pytest.raises(ConfigurationError) as exc:
+            config.load(config_details)
+        assert 'driver_opts.size contains an invalid type' in exc.exconly()
+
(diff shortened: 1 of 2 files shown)


tests/unit/config/config_test.py
```diff
@@ -231,6 +231,19 @@ def test_named_volume_config_empty(self):
...
+            'version': '2',
+            'services': {
+                'simple': {'image': 'busybox'}
+            },
+            'volumes': {
+                'simple': {'driver_opts': {'size': 42}},
+            }
+        })
+        with self.assertRaises(ConfigurationError):
+            config.load(config_details)
```
dnephin: Minor: we've been using `pytest.raises(...)` in most places. It's basically the same idea, but returns a different object which I think makes it easier to inspect errors. It will also let us transition off of `unittest.TestCase` classes.

Probably worth checking the error message here as well, since the error itself can be raised  for many different conditions.

shin-: :+1: updated

dnephin:
LGTM (when build is green)

PR #2905, opened 2016-02-12
Fix signal handling with pyinstaller binary

### LINKED ISSUE #2904: Pyinstaller has issues with signals
There's a bunch of history in #1040 and #2055.

We've tried multiple implementations of signal handlers, but each has their own set of issues, but **ONLY** when run from the frozen binary created by pyinstaller.

It looks like there is a very old issue in pyinstaller around this: https://github.com/pyinstaller/pyinstaller/issues/208

These problems can manifest in three ways:
* a `thread.error` when a signal interrupts a thread lock
* the signal handlers being completely ignored and raising a `KeynoardInterupt` instead
* the signal handlers being registered but the try/except to handle the except is skipped (this could be caused by the signal firing multiple times for a single `ctrl-c`, but I can't really verify that's what is happening)

dnephin:
Fixes #2904

If you have an OSX box, it would be good to test this out.  It seems to have completely fixed the issue on linux.

Upstream PR is https://github.com/pyinstaller/pyinstaller/pull/1822

### DIFF
--- a/requirements-build.txt
+++ b/requirements-build.txt
@@ -1,1 +1,3 @@
-pyinstaller==3.1.1
+#pyinstaller==3.1.1
+# pyinstaller 3.1.1 with fix for signals
+git+https://github.com/dnephin/pyinstaller.git@9e8cad86e6f19f16ad9aec4e16bc5c57454cb072#egg=pyinstaller


dnephin:
The problem with this version is that it leaves tmp files lying around.

I think we can address that by using a tempdir that is based on a hash of something, that way there will only ever be one copy lying around, which is probably ok.

dnephin:
This doesn't work, so closing for now

Issue #1516, opened 2015-06-05
Proposal: Stop using sequential container numbers


aanand:
Sequential container numbers (`myapp_web_1`, `myapp_web_2`) are problematic.
1. They're costly to maintain: we have to query the names of all containers for a service to determine what the next number should be. As apps get bigger, this is going to get worse.
2. They're error-prone: if a Swarm node with `myapp_web_LASTNUM` goes down at the same time as Compose is querying the cluster, it will incorrectly think `LASTNUM` is the next number we should use. Once the node comes up, there'll be a clash of container names.
3. They make it impossible to do multiple `docker-compose run`s in parallel.

As such, I think we should start using random suffixes. This may well break some scripts that rely on container names, but now that we're using labels to track containers, we should be explicit about making _no guarantees_ about container names.

villlem:
For scripts is useful ps command with -q option. If there is option for get my only one container there would be no need for some explicit naming.

SBoudrias:
I just wondered what were the plans for this feature? Is it planned soon?

dnephin:
Hey Simon!

We aren't doing anything about this in the 1.5 release (which will be out soon), but I think we may look to implement something in the 1.6 release which will be a few months away.

murphyke:
+1  I want to run parallel copies of a container in a data processing pipeline.  If there were a `--container-name=NAME` CLI flag, I would be in business, but random suffixes would work also.

ibizaman:
Another use case where this would be useful is if the prefix - the directory of the `docker-compose.yml` - is the same and the same service name is used.

For example, having two apps with the following structure:
```
myapp/
    doc/
    src/
        docker-compose.yml
myapp2/
    doc/
    src/
        docker-compose.yml
```
and assuming they both use a `mysql` service, this will result in two identically named services `src_mysql_1`.

A not so great solution would be to use the complete `$PWD` as a prefix.
But in fact, why do we even need to name the machines in the first place? This is not a troll question. :)

dnephin:
The project name is required to support [isolated environments](https://docs.docker.com/compose/overview/#multiple-isolated-environments-on-a-single-host).

caiofbpa:
:+1:

rvernica:
@dnephin, just wondering, are you still planning to remove the sequential container numbers?

kchudy:
I just came across this issue when CI executes tests in parallel (triggered by different commits). I handled it by generating a random name for the container like so:
```
[...]

PR #2906, opened 2016-02-12
Round-robin container attribution on service network_mode

lphuberdeau:
Before this patch, all slave containers would have the equivalent of "container:primary_1" as a network mode. The patch applies a round-robin logic to container attribution. I won't affect previously created containers, but at least allows scaling services that depend on the service network mode

Sample compose file:

```
primary:
    image: ubuntu
    command: sleep 60
slave:
    image: ubuntu
    command: sleep 60
    network_mode: "service:primary"
```

Sample set-up:

```
docker-compose up -d
docker-compose scale primary=2
docker-compose scale slave=5
```

### DIFF
--- a/compose/service.py
+++ b/compose/service.py
@@ -621,11 +621,12 @@ class Service(object):
 
         container_options['host_config'] = self._get_container_host_config(
             override_options,
-            one_off=one_off)
+            one_off=one_off,
+            number=number)
 
         return container_options
 
-    def _get_container_host_config(self, override_options, one_off=False):
+    def _get_container_host_config(self, override_options, one_off=False, number=1):
         options = dict(self.options, **override_options)
 
         logging_dict = options.get('logging', None)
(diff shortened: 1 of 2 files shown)


dnephin:
Thanks for the PR. We had a chance to discuss it briefly at our maintainers sync-up today.

Unfortunately I don't think we're going to be able to accept this change. It's an interesting idea, but I don't think it's a robust solution to the problem. This "pod-like"  functionality really needs to be supported by the engine running the container, not just the client.

We'd also like to stop using these sequence numbers, see #1516, so adding more features that depend on them will make this more difficult.

lphuberdeau:
Assigning the network mode to the first service in the list is rather arbitrary and prevents any kind of scaling. Without using numbers, random allocation would be better than the current situation. For small volumes, or when stricter relations are required, it would be unusable however.

For the random vs numbering, could it be optional? Random is good for some cases, but numbers re really useful when affinity is required. We have some cases where we need multiple workers per main containers, others where there is a one to one relationship, and others where it's the more trivial horizontal scaling.

I could convert the logic here to take the container name based on the selected scheme and apply the best logic possible.

shin-:
My understanding is that this change is now obsolete when using the v2/v3 format and new networking. I'll go ahead and close this.

PR #2907, opened 2016-02-13
Allow user to specify custom network aliases

### LINKED ISSUE #2829: `--net-alias` support in v2 Compose File
I am unsure if I am missing this in the docs or not, but I'm not seeing any way in Compose 1.6.0 that implements the new `--net-alias` support in Docker-Engine.

Here's the relevant add from the Docker Engine release notes: https://github.com/docker/docker/pull/19242

My specific end-user ask is a way to define multiple aliases for a container on a network. Something like this, perhaps?

```yml

networks:
    example

services:
    www:
        image: apache:2.4.18
        networks:
            example:
                aliases:
                    - localweb
                    - www.localaddress.com
````

### PR DESCRIPTION
**Pull request by shin-**
Closes #2829

### DIFF
--- a/compose/network.py
+++ b/compose/network.py
@@ -162,10 +162,7 @@ class ProjectNetworks(object):
 def get_network_aliases_for_service(service_dict):
     if 'network_mode' in service_dict:
         return {}
-    networks = service_dict.get('networks', ['default'])
-    if isinstance(networks, list):
-        return dict((net, []) for net in networks)
-
+    networks = service_dict.get('networks', {'default': None})
     return dict(
         (net, (config or {}).get('aliases', []))
         for net, config in networks.items()
--- a/compose/network.py
+++ b/compose/network.py
@@ -15,7 +15,7 @@ log = logging.getLogger(__name__)
 
 class Network(object):
     def __init__(self, client, project, name, driver=None, driver_opts=None,
-                 ipam=None, external_name=None, aliases=None):
+                 ipam=None, external_name=None):
         self.client = client
         self.project = project
         self.name = name
@@ -23,7 +23,6 @@ class Network(object):
         self.driver_opts = driver_opts
         self.ipam = create_ipam_config_from_dict(ipam)
         self.external_name = external_name
-        self.aliases = aliases or []
 
     def ensure(self):
         if self.external_name:
@@ -160,25 +159,32 @@ class ProjectNetworks(object):
             network.ensure()
 
 
-def get_network_names_for_service(service_dict):
+def get_network_aliases_for_service(service_dict):
     if 'network_mode' in service_dict:
-        return []
-    return service_dict.get('networks', ['default'])
+        return {}
+    networks = service_dict.get('networks', ['default'])
+    if isinstance(networks, list):
+        return dict((net, []) for net in networks)
+
+    return dict(
+        (net, (config or {}).get('aliases', []))
+        for net, config in networks.items()
+    )
+
+
+def get_network_names_for_service(service_dict):
+    return get_network_aliases_for_service(service_dict).keys()
 
 
 def get_networks(service_dict, network_definitions):
     networks = {}
-    aliases = service_dict.get('network_aliases', {})
-    for name in get_network_names_for_service(service_dict):
-        log.debug(name)
+    for name, aliases in get_network_aliases_for_service(service_dict).items():
         network = network_definitions.get(name)
         if network:
-            log.debug(aliases)
-            networks[network.full_name] = aliases.get(name, [])
+            networks[network.full_name] = aliases
         else:
             raise ConfigurationError(
                 'Service "{}" uses an undefined network "{}"'
                 .format(service_dict['name'], name))
 
-    log.debug(networks)
     return networks
--- a/compose/network.py
+++ b/compose/network.py
@@ -15,7 +15,7 @@ log = logging.getLogger(__name__)
 
 class Network(object):
     def __init__(self, client, project, name, driver=None, driver_opts=None,
-                 ipam=None, external_name=None):
+                 ipam=None, external_name=None, aliases=None):
         self.client = client
         self.project = project
         self.name = name
@@ -23,6 +23,7 @@ class Network(object):
         self.driver_opts = driver_opts
         self.ipam = create_ipam_config_from_dict(ipam)
         self.external_name = external_name
+        self.aliases = aliases or []
 
     def ensure(self):
         if self.external_name:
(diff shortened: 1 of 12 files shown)

### CODE REVIEW

**Comment by aanand** [comment]
Just catching up - the current design of `network_aliases` looks alright, but I personally preferred the dict form of `networks` proposed in #2829 - it looks neater, doesn't require inter-key validation (i.e. we don't have to check the `network_aliases` keys match the `networks` list) and gives us the ability to add more sub-keys in future.

Is it really an issue that `networks` can be either a list or a dict?

**Comment by shin-** [comment]
@aanand : 
From my understanding, the new format would have been a list of (dict or string), as such:
```yaml
web:
  networks:
    - one
    - another:
         aliases:
           - foo
```

Personally, I find that quite confusing and tedious - and that's with just two networks and a single alias. It is also fairly error-prone, see:

```yaml
web:
    networks:
        - one:
            aliases:
                - aaa
          another:
            aliases:
                - bbb
```
This looks right, yet the missing dash would make it invalid.

While having a separate parameter might look more verbose at first glance, I think it ends up being neater and more readable.

londoncalling:
@shin- LGTM re: compose-file.md

I had an idea about how to debug the Docs build, but my understanding of how to pull the code from the PR is apparently lacking. Let me know if you need help w/Docs build.

londoncalling:
@shin- Okay I figured it out :-) 
If you do the following, the docs build should work.

1. Remove this line from docs/Dockerfile: `RUN svn checkout https://github.com/docker/tutorials/trunk/docs /docs/content/tutorials`
3. Save and close the docs/Dockerfile
4. Commit the change to the Dockerfile, along with your other changes, and add it to this pull request

Let me know if you are okay with doing this yourself, or if you'd rather I do it as another PR, or carry this one.
Vicky

shin-:
Thanks @londoncalling ! I rebased the branch against master and things should look good now on that front.

shin-:
Test failure unrelated (aufs bug)

londoncalling:
Awesome  :-)

13josh:
Agreed with @aanand. I think the network_aliases is less desirable than what was suggested in the #2829.  It also makes for more complicated error checking and failure modes.  What if the user specifies an undefined network in network_aliases?  

```yaml
 networks:
  - some-network
 network_aliases:
  some-misspelled-network:
   - alias1
```

You can avoid having to do that mapping that of network_aliases to networks and just treat the aliases as arguments to the network call in the first place.  I'm not personally too worried about the potential for confusion a dash vs. no-dash case, particularly since the example you give is for the already existing 'networks' keyword.

I think it does _look_ tedious on first glance (that was my impression too), but you have to specify the network and you have to specify the alias *and* the network so it's tomato/tomatoe in that sense.

shin-:
Updated.

As discussed this morning, `networks` can either be a list of string or a dictionary. Dict keys are network names, values are service-specific network configuration (currently only `aliases`). Value can be null.

```yaml
networks:
  foo:
  bar:
    aliases:
      - alias1
      - alias3

File: docs/compose-file.md [hunk 1]
```diff
@@ -475,6 +475,19 @@ Networks to join, referencing entries under the
       - some-network
       - other-network
 
+#### aliases
+
+Alias names for this service on the specified network.
```
dnephin: Maybe add something about these aliases being usable as the hostname to connect to the service? I'm not sure if "name" is enough information here.

Something like this maybe:

> Aliases for this service on the network. Other servers can use this name to connect to this service.

@londoncalling what do you think?

compose/service.py
```diff
@@ -514,9 +514,9 @@ def _next_container_number(self, one_off=False):
 
     def _get_aliases(self, container):
         if container.labels.get(LABEL_ONE_OFF) == "True":
-            return []
+            return set()
 
-        return [self.name, container.short_id]
+        return set([self.name, container.short_id])
```
dnephin: minor: the `[]` are unnecessary here, you can actually do use a set literal now that we dropped python 2.6 support:

```python
return {self.name, container.short_id}
```

compose/network.py
```diff
@@ -159,18 +159,29 @@ def initialize(self):
...
-    return service_dict.get('networks', ['default'])
+        return {}
+    networks = service_dict.get('networks', ['default'])
+    if isinstance(networks, list):
+        return dict((net, []) for net in networks)
+
+    return dict(
+        (net, (config or {}).get('aliases', []))
+        for net, config in networks.items()
+    )
```
dnephin: I would actually expect that this normalization would need to happen in `compose/config/config.py`, otherwise merging networks from multiple files will probably fail when one is a list and the other is a dict.

Mind adding a test to `tests/unit/config/config_test.py` to confirm ?

londoncalling:
@shin- this might be overkill, but if all this is true ( per the original [proposal re: container aliases](https://github.com/docker/docker/issues/18699 ) this leverages), the it seems useful to include in the docs. Let me know what you think.

Aliases (alternative hostnames) for this service on the network. Other servers on the network can use either the hostname or this alias to connect to this service.  Since  alias is network-scoped:

* the same service can have different aliases when connected to another network.

* it is allowable to configure the same alias name to multiple containers in the same network.

dnephin:
I think that second sentence should be:

"Other services on the network can use either the service name or this alias to connect to this service"

Even though these all act as "hostnames" `hostname` is another field in the config, and I think using that word may confuse people.

shin-:
Rebased + updated

dnephin:
LGTM!

aanand:
I've got some docs comments but it needn't hold up the merge - we can fix it in a separate PR if others agree.

- I find the wording a bit odd, especially the way it implies that containers and services are the same thing.
- I don't think the ability to give the same alias to multiple services on the same network is enabled by aliases being network-scoped. It's simply the case that any number of containers on one network can have the same alias.
- It might be good to have a sort-of-realistic example.

How about:

---

Aliases (alternative hostnames) for this service on the network. Other containers on the same network can use either the service name or these aliases to connect to one of this service's containers.

Since `aliases` is network-scoped, the same service can have different aliases on different networks.

> **Note:** A network-wide alias can be shared by multiple containers, and even by multiple services. If it is, then exactly which container the name will resolve to is not guaranteed.

In the example below, the `db` service is reachable at the hostname `db` or `database` on the `new` network, and at `db` or `mysql` on the `legacy` network.

    version: 2

    services:
      web:
        build: ./web
        networks:
          - new

      worker:
        build: ./worker
        networks:
          - legacy

      db:
        image: mysql
        networks:
          new:
            aliases:
              - database
          legacy:
            aliases:
              - mysql

    networks:
      new:
      legacy:

londoncalling:
@shin- one small nit re: docs update: The code example doesn't look like it's formatted as code in compose.md `networks: some-network: aliases: - alias1 - alias3 other-network: aliases: - alias2`

londoncalling:
@aanand Thanks! this helps clarify things for me too.

@shin- why don't you merge your changes, and I'll pick them up and make the docs changes. I can open a PR or @aanand can, and ping me on it.

shin-:
Okay, merging!

PR #2913, opened 2016-02-13
A few fixes to the container-based install, which is broken on:

clkao:
* when pwd is /
* when stdin is not tty
* when invoking `run sh -c 'echo foo'`

### DIFF
--- a/script/run.sh
+++ b/script/run.sh
@@ -53,4 +53,4 @@ if [ -t 0 ]; then
     DOCKER_RUN_OPTIONS="$DOCKER_RUN_OPTIONS -i"
 fi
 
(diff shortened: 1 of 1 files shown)


script/run.sh
```diff
@@ -31,7 +31,9 @@ fi
 
 
 # Setup volume mounts for compose config and context
-VOLUMES="-v $(pwd):$(pwd)"
+if [ $(pwd) != '/' ]; then
```
dnephin: There should be double quotes around `"$(pwd)"` otherwise spaces will break this I think.

clkao: ok, edited the commit and adding quotes to the -w in run as well.

dnephin:
LGTM

aanand:
LGTM

PR #2932, opened 2016-02-16
reset colors after warning

AvdN:
If a warning is shown, and you happen to have no color setting in your (bash) prompt, the \033[37m setting stays active. With the message being hardly readable (light grey on my default light yellow background), that means the prompt is barely visible and you need to do `tput reset`.

(Would probably be better if the background color was set as well in case you have dark on light theme by default in your terminal.)

Signed-off-by: Anthon van der Neut anthon@mnt.org

### DIFF
--- a/contrib/migration/migrate-compose-file-v1-to-v2.py
+++ b/contrib/migration/migrate-compose-file-v1-to-v2.py
@@ -155,7 +155,7 @@ def parse_opts(args):
 
 
 def main(args):
-    logging.basicConfig(format='\033[33m%(levelname)s:\033[37m %(message)s\n')
+    logging.basicConfig(format='\033[33m%(levelname)s:\033[37m %(message)s\033[0m\n')
 
     opts = parse_opts(args)
 


dnephin:
Seems alright, although I don't see us doing this in `compose/cli/colors.py` either, so maybe this project exists outside of the migration script?

When you are seeing these colors, is it part of a migration or is it while running `docker-compose?

AvdN:
I see these while running the conversion utiltity ( `contrib/migration/migrate-compose-file-v1-to-v2.py` ) which emits the warning because my docker-compose.yml has undeclared external (host based) volumes. 
I haven't seen these in normal docker-compose use, but then all my projects are very similar, generated from a template, so I can't recall having  seen warnings there.
And of course if your bash/zsh/fish prompt sets colors in your prompt, chances are high you'll never notice, because that reset the coloring.

dnephin:
LGTM

AvdN:
I see these while running the conversion utility ( `contrib/migration/migrate-compose-file-v1-to-v2.py` ) which emits the warning because my docker-compose.yml has undeclared external (host based) volumes. 
I haven't seen these in normal docker-compose use, but then all my projects are very similar, generated from a template, so I can't recall having  seen warnings there.
And of course if your bash/zsh/fish prompt sets colours in your prompt, chances are high you'll never notice, because that reset the colouring.

aanand:
LGTM

PR #2934, opened 2016-02-16
Fix copying of volumes by using the name of the volume

### LINKED ISSUE #2912: DOC: Unexpected data loss when recreate containers
Hi everyone,

Im not sure if this is a docker, docker-compose or a "me" bug. But Im not getting my head around this for several hours. 

My simple docker-compose.yml
```
version: "2"
services:
  app:
    image: nginx:latest
    volumes_from:
      - storage
  storage:
    image: ubuntu:latest
    volumes:
      - /var/data
```
I bootstrap it with ```$ docker-compose up -d```.

Now I jump into "app" via ```docker exec -it docker_app_1 bash``` and add a test file within my new created volume.

```
root@8f5f850c30c7:/# touch /var/data/test
root@8f5f850c30c7:/# ls /var/data
test
```

Seems to work. Shutdown via ```$ docker-compose stop```. Now I add a second volume to my storage container:
```
version: "2"
services:
  app:
    image: nginx:latest
    volumes_from:
      - storage
  storage:
    image: ubuntu:latest
    volumes:
      - /var/data
      - /my/host/folder:/var/data2
```
This time this volume is backed by a host directory. Rerun:
```
$ docker-compose up -d
Recreating docker_storage_1
Recreating docker_app_1
```
Okay so far. Recreated my storage container, but should preserve existing volumes (docs). Lets check this:
```
$ docker exec -it docker_app_1 bash
root@8f5f850c30c7:/# ls /var/data
```
Oh my data is gone, but why? This is not the only time my data gots "lost" the last days. 
Its not 100% clear when this will happen. Simple running ```$ docker-compose up -d --force-recreate``` twice kills my data also. After the first launch my data is still there and after the second one its lost. Is this expected behavior? Recreate process seems to be broken in compose 1.6.0 :(

This happens on my local environment (OS X, docker 1.10.0, compose 1.6.0) and on my ubuntu-backed vServer (docker 1.10.1, compose 1.6.0).

Thanks for any advice

System:
```
$ docker-compose -v

$ docker version
Client:
 Version:      1.10.0
 API version:  1.22
 Go version:   go1.5.3
 Git commit:   590d5108
 Built:        Thu Feb  4 18:18:11 2016
 OS/Arch:      darwin/amd64

Server:
 Version:      1.10.0
 API version:  1.22
 Go version:   go1.5.3
 Git commit:   590d5108
 Built:        Thu Feb  4 19:55:25 2016
 OS/Arch:      linux/amd64
```

dnephin:
Fixes #2912 

This is a long overdue fix, that unfortunately caused a bug due to a change we made in 1.6.0.  In 1.6.0 we stopped copying host volumes from previous containers (because it's redundant).

However, the "copy volume" logic made any copied volumes appear as if they were host volumes instead of named volumes, which meant that on the second recreate we'd lose the volumes (and new ones would be created).  

This wasn't found by tests because we never recreated twice.

This fixes the 1.6.0, and should make this logic more resilient going forward. Instead of using the host path, we use the name (which didn't exist when this "copy volume" logic was first added). By using the name it's exactly like we're using a named volume.

### DIFF
--- a/compose/service.py
+++ b/compose/service.py
@@ -925,7 +925,7 @@ def get_container_data_volumes(container, volumes_option):
             continue
 
         # Copy existing volume from old container
-        volume = volume._replace(external=mount['Source'])
+        volume = volume._replace(external=mount['Name'])
         volumes.append(volume)
 
     return volumes
--- a/tests/integration/service_test.py
+++ b/tests/integration/service_test.py
@@ -273,6 +273,30 @@ class ServiceTest(DockerClientTestCase):
                           self.client.inspect_container,
                           old_container.id)
 
+    def test_execute_convergence_plan_recreate_twice(self):
+        service = self.create_service(
+            'db',
+            volumes=[VolumeSpec.parse('/etc')],
+            entrypoint=['top'],
+            command=['-d', '1'])
+
+        orig_container = service.create_container()
+        service.start_container(orig_container)
+
+        orig_container.inspect()  # reload volume data
+        volume_path = orig_container.get_mount('/etc')['Source']
+
+        # Do this twice to reproduce the bug
+        for _ in range(2):
+            new_container, = service.execute_convergence_plan(
+                ConvergencePlan('recreate', [orig_container]))
+
+            assert new_container.get_mount('/etc')['Source'] == volume_path
+            assert ('affinity:container==%s' % orig_container.id in
+                    new_container.get('Config.Env'))
+
+            orig_container = new_container
+
     def test_execute_convergence_plan_when_containers_are_stopped(self):
         service = self.create_service(
             'db',
--- a/tests/unit/service_test.py
+++ b/tests/unit/service_test.py
@@ -691,8 +691,8 @@ class ServiceVolumesTest(unittest.TestCase):
         }, has_been_inspected=True)
 
         expected = [
-            VolumeSpec.parse('/var/lib/docker/aaaaaaaa:/existing/volume:rw'),
-            VolumeSpec.parse('/var/lib/docker/cccccccc:/mnt/image/data:rw'),
+            VolumeSpec.parse('existingvolume:/existing/volume:rw'),
+            VolumeSpec.parse('imagedata:/mnt/image/data:rw'),
         ]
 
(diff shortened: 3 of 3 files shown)


dnephin:
I'm actually surprised this wasn't found until now (and there have only been 2 reports so far). It makes me believe that most people are using Compose for CI (where you always `rm -vf` everything) and not really relying on the copy-volume-on-recreate logic.

tests/integration/service_test.py
```diff
@@ -273,6 +273,30 @@ def test_execute_convergence_plan_recreate(self):
...
+            command=['-d', '1'])
+
+        orig_container = service.create_container()
+        service.start_container(orig_container)
+
+        orig_container.inspect()  # reload volume data
+        volume_path = orig_container.get_mount('/etc')['Source']
+
+        # Do this twice to reproduce the bug
+        for _ in xrange(2):
```
shin-: Doesn't exist in python 3. Just use `range(2)` instead - we probably have enough RAM for those 12 bytes.

compose/service.py
```diff
@@ -925,7 +925,7 @@ def get_container_data_volumes(container, volumes_option):
             continue
 
         # Copy existing volume from old container
-        volume = volume._replace(external=mount['Source'])
+        volume = volume._replace(external=mount['Name'])
```
shin-: Does that work with 1.9.x as well?

tests/integration/service_test.py
dnephin: Oops, right!

compose/service.py
dnephin: It should, I believe named volumes were in 1.9. Our test suite still runs against 1.9, so it should confirm for us.

shin-:
LGTM!

romankonz:
I just want to mention that i ran into this bug today with an (non-critical) application in production.
So, not everyone is using compose only for CI ;)

Thanks for fixing this!

pwaller:
Also to comment, we've hit upon it here too, but I found the existing bug report so didn't refile it or add a :+1:, I just subscribed. Glad to see this fixed and sitting tight for 1.6.1!

PR #2939, opened 2016-02-17
for 1.6.0 the version value needs to be a string

AvdN:
After conversion a file would immediately not load in docker-compose 1.6.0 with the message:

  ERROR: Version in "./converted.yml" is invalid - it should be a string.

Signed-off-by: Anthon van der Neut anthon@mnt.org

### DIFF
--- a/contrib/migration/migrate-compose-file-v1-to-v2.py
+++ b/contrib/migration/migrate-compose-file-v1-to-v2.py
@@ -33,7 +33,7 @@ def migrate(content):
 
     services = {name: data.pop(name) for name in data.keys()}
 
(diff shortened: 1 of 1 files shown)


AvdN:
I have no clue why janky failed. None of the current tests seem to actually use the contributed code. 
I closed the PR and reopened to force a rebuild, that didn't do it.

dnephin:
test failure is unrelated

LGTM

dnephin:
Thanks for the fix! we changed this late, and forgot to fix it in the migration script

shin-:
LGTM!

PR #2953, opened 2016-02-18
Enable interpolation of keys in compose file

dvaergiller:
This adds interpolation of all keys in the compose file except the
sections themselves. I have a need for doing that when, for example,
creating overlay networks based on some unique name:

```
networks:
  ${NAMESPACE}_overlay:
     driver: overlay

services:
  my_service:
    networks:
      - "${NAMESPACE}_overlay"
```

The reason for the slight code restructuring was to avoid code
duplication when implementing this. I do welcome feedback.

### DIFF
--- a/compose/config/interpolation.py
+++ b/compose/config/interpolation.py
@@ -14,27 +14,11 @@ log = logging.getLogger(__name__)
 def interpolate_environment_variables(config, section):
     mapping = BlankDefaultDict(os.environ)
 
-    def process_item(name, config_dict):
-        return dict(
-            (key, interpolate_value(name, key, val, section, mapping))
-            for key, val in (config_dict or {}).items()
-        )
-
-    return dict(
-        (name, process_item(name, config_dict))
-        for name, config_dict in config.items()
-    )
-
-
-def interpolate_value(name, config_key, value, section, mapping):
     try:
-        return recursive_interpolate(value, mapping)
+        return recursive_interpolate(config, mapping)
     except InvalidInterpolation as e:
         raise ConfigurationError(
-            'Invalid interpolation format for "{config_key}" option '
-            'in {section} "{name}": "{string}"'.format(
-                config_key=config_key,
-                name=name,
+            'Invalid interpolation format for "{string}" in {section}'.format(
                 section=section,
                 string=e.string))
 
@@ -44,7 +28,7 @@ def recursive_interpolate(obj, mapping):
         return interpolate(obj, mapping)
     elif isinstance(obj, dict):
         return dict(
-            (key, recursive_interpolate(val, mapping))
+            (interpolate(key, mapping), recursive_interpolate(val, mapping))
             for (key, val) in obj.items()
         )
     elif isinstance(obj, list):
--- a/compose/config/interpolation.py
+++ b/compose/config/interpolation.py
@@ -18,7 +18,7 @@ def interpolate_environment_variables(config, section):
         return recursive_interpolate(config, mapping)
     except InvalidInterpolation as e:
         raise ConfigurationError(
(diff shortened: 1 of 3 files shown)


dvaergiller:
So, I have a couple of failed tests. One seems unrelated to this patch. The other says "Differing items: {'other': None} != {'other': {}}"

Now, I could make the interpolation code replace None for {}, but is that really what we want? I would appreciate some help with this one.

dvaergiller:
Also, the master branch fails to run at all on my machine (OSX 10.11, Python 2.7), although the 1.6.0 tag works. I rebased the commits on 1.6.0 and everything seems to be working fine there. I now have two branches, interpolate_keys and interpolate_keys_1.6. The latter is mostly for my own use.

dnephin:
Could you explain a bit more why it's necessary to use a variable for the network name?  You could always make the network external, and then you would only need to put a variable in the `name:` field, which would work as-is.

dvaergiller:
Yes I could make the network external, but if I understand it correctly that means that I can not create overlays on the fly from compose.

What I need to be able to do is to run compose multiple times to deploy and take down a whole group of containers, connected to each other over a unique overlay so that they are isolated from instances previously set up by the same compose file:

NS=test1 docker-compose up
NS=test2 docker-compose up

I want this to result in two deployments on two different overlays. If there is already a way to do that then that would be great.

Another point is that this patch makes the behaviour more like I expected it to be. At first I just assumed that this is how it would work because "why wouldn't it"?

dnephin:
Yes, what you're describing is already a [key feature of Compose](https://docs.docker.com/compose/overview/#multiple-isolated-environments-on-a-single-host).

Start each instance with a different project name and they will use different network names.

dnephin:
> At first I just assumed that this is how it would work because "why wouldn't it"?

The mains reasons i can see are:
* it's unnecessary, the keys represent the structure and shouldn't change, and isolation is already supported
* it makes reporting errors a lot more difficult because the path to the error might be a variable

dvaergiller:
Ah. I had missed the little detail that the project name can be configured. That solves my use case. Thanks a lot.

Now that I have no direct use for this, and I don't know how anyone else would, let's close this?

dnephin:
That sounds appropriate.  Thanks for experimenting with this idea, I'm glad we could get it sorted out.

n-rodriguez:
> Now that I have no direct use for this, and I don't know how anyone else would,

@dnephin Hi there! Sorry to dig this out but now I have a use case : what if I want to templatize `configs` and `secrets` keys like :

```yml
secrets:
  # These are MySQL image passwords
  ${SWARMLET_APP}-${SWARMLET_ENV}__MYSQL_ROOT_PASSWORD:
    external: true
  ${SWARMLET_APP}-${SWARMLET_ENV}__MYSQL_PASSWORD:
    external: true

  # These are Postgres image passwords
  ${SWARMLET_APP}-${SWARMLET_ENV}__POSTGRES_PASSWORD:
    external: true

  # These are Sinatra image passwords
  ${SWARMLET_APP}-${SWARMLET_ENV}__MYSQL_DB_PASS:
    external: true
  ${SWARMLET_APP}-${SWARMLET_ENV}__POSTGRES_DB_PASS:
    external: true
  ${SWARMLET_APP}-${SWARMLET_ENV}__SENTRY_DSN:
    external: true

configs:
  ${SWARMLET_APP}-${SWARMLET_ENV}__CONFIG:
    external: true
```

That would help me a lot because today I need 2 docker-compose files in my repo 1 for `staging` and 1 for `production` because of this limitation.

I understand your objections ("it makes reporting errors a lot more difficult because the path to the error might be a variable") but as usual, with great power, great responsability.

Maybe it could be implemented with an optin option like `--interpolate-keys` and an environment variable ?

PR #2956, opened 2016-02-18
Abort on container exit fix

### LINKED ISSUE #2940: abort-on-container-exit does not work as expected
```
> uname -a
Linux mycomputername 3.13.0-77-generic #121-Ubuntu SMP Wed Jan 20 10:50:42 UTC 2016 x86_64 x86_64 x86_64 GNU/Linux
> docker -v
Docker version 1.10.1, build 9e83765
> docker-compose -v
```
According to the docs https://docs.docker.com/compose/reference/up/,
`--abort-on-container-exit` should
> Stop all containers if any container was stopped.

docker-compose.yml:
```
version: "2"

services:
    loop1:
        image: ubuntu:latest
        command: bash -c 'for i in {1..5}; do echo $$i && sleep 1; done'
        stop_signal: SIGKILL

    loop2:
        image: ubuntu:latest
        command: bash -c 'for i in {1..10}; do echo $$i && sleep 1; done'
        stop_signal: SIGKILL
```

After `docker-compose up --abort-on-container-exit` both containers start sending log messages.
When `loop1` exits, documentation suggests that `loop2` will also be stopped. This does not happen, which can be confirmed by executing `docker-compose ps` shortly after `loop1` exits.

An other issue which may be related:
When hitting `Ctrl-C` after `docker-compose up` I would expect this:
```
Stopping dockercompose_loop1_1 ... done
Stopping dockercompose_loop2_1 ... done
```
This is only happening occasionally, most of the time I get this instead:
```
^CERROR: Aborting.
```
Also, the containers remain up.

rbann:
This fixes #2940.

### DIFF
--- a/compose/cli/log_printer.py
+++ b/compose/cli/log_printer.py
@@ -5,6 +5,7 @@ import sys
 from itertools import cycle
 
 from . import colors
+from . import signals
 from .multiplexer import Multiplexer
 from compose import utils
 from compose.utils import split_buffer
@@ -41,7 +42,7 @@ class LogPrinter(object):
         for color_func, container in zip(color_funcs, self.containers):
             generator_func = get_log_generator(container)
             prefix = color_func(build_log_prefix(container, prefix_width))
-            yield generator_func(container, prefix, color_func)
+            yield generator_func(container, prefix, color_func, self.cascade_stop)
 
 
 def build_log_prefix(container, prefix_width):
@@ -64,7 +65,7 @@ def get_log_generator(container):
     return build_no_log_generator
 
 
-def build_no_log_generator(container, prefix, color_func):
+def build_no_log_generator(container, prefix, color_func, cascade_stop):
     """Return a generator that prints a warning about logs and waits for
     container to exit.
     """
@@ -72,9 +73,11 @@ def build_no_log_generator(container, prefix, color_func):
         prefix,
         container.log_driver)
     yield color_func(wait_on_exit(container))
+    if cascade_stop:
+        raise signals.CascadeStopException()
 
 
-def build_log_generator(container, prefix, color_func):
+def build_log_generator(container, prefix, color_func, cascade_stop):
     # if the container doesn't have a log_stream we need to attach to container
     # before log printer starts running
     if container.log_stream is None:
@@ -86,6 +89,8 @@ def build_log_generator(container, prefix, color_func):
     for line in line_generator:
         yield prefix + line
     yield color_func(wait_on_exit(container))
+    if cascade_stop:
+        raise signals.CascadeStopException()
 
 
 def wait_on_exit(container):
--- a/tests/acceptance/cli_test.py
+++ b/tests/acceptance/cli_test.py
@@ -746,14 +746,11 @@ class CLITestCase(DockerClientTestCase):
         os.kill(proc.pid, signal.SIGTERM)
         wait_on_condition(ContainerCountCondition(self.project, 0))
 
-    @v2_only()
     def test_up_handles_abort_on_container_exit(self):
-        self.base_dir = 'tests/fixtures/sleeps-composefile'
-        start_process(
-            self.base_dir,
-            ['-f', 'one-exits.yml', 'up', '-t', '200', '--abort-on-container-exit'])
+        start_process(self.base_dir, ['up', '--abort-on-container-exit'])
         wait_on_condition(ContainerCountCondition(self.project, 2))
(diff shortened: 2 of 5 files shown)


dnephin:
I don't think this solves the problem, see my comment in the issue.

tests/fixtures/sleeps-composefile/one-exits.yml
```diff
@@ -0,0 +1,6 @@
+simple:
+  image: busybox:latest
+  command: sleep 200
+another:
+  image: busybox:latest
+  command: sleep 0.5
```
dnephin: I think we could even do `echo something`, that way we're also testing that there isn't any race condition when containers exit immediately (which has been an issue before).

It also means the test doesn't need to wait 0.5 seconds to complete.

tests/acceptance/cli_test.py
```diff
@@ -718,6 +718,15 @@ def test_up_handles_force_shutdown(self):
         os.kill(proc.pid, signal.SIGTERM)
         wait_on_condition(ContainerCountCondition(self.project, 0))
 
+    @v2_only()
+    def test_up_handles_abort_on_container_exit(self):
+        self.base_dir = 'tests/fixtures/sleeps-composefile'
+        start_process(
+            self.base_dir,
+            ['-f', 'one-exits.yml', 'up', '-t', '200', '--abort-on-container-exit'])
```
dnephin: I think this `-t` timeout should be `-t 1`, The value is set for 200 just for that one test, because it's relevant in that case, but in this case we know that `sleep` requires a SIGKILL to shutdown, so we don't want to wait around for it.

tests/fixtures/sleeps-composefile/one-exits.yml
dnephin: Scratch that, I have a better idea. How about both services use `top` as the command. We use `top` because it exits properly on `SIGTERM`. I'll cover the testing changes above.

tests/acceptance/cli_test.py
```diff
@@ -718,6 +718,15 @@ def test_up_handles_force_shutdown(self):
...
+        wait_on_condition(ContainerCountCondition(self.project, 2))
+        wait_on_condition(ContainerCountCondition(self.project, 0), timeout=1)
```
dnephin: Ah, I guess this is why it needs to `sleep 0.5`, although on a loaded jenkins instance, this could still fail, even with the sleep. The `timeout=1` is also likely to flake a lot.

I'd like to propose this testing strategy using the `top` command.

```yaml
# Set project so teardown does the correct cleanup, you can find examples of this in other tests
self._project = get_project(self.base_dir, ['one-exits.yml'])
start_process(...)

wait_on_condition(ContainerCountCondition(self.project, 2))
self.project.stop(['simple'])
wait_on_condition(ContainerCountCondition(self.project, 0))
```

compose/cli/log_printer.py
```diff
@@ -41,7 +42,7 @@ def no_color(text):
         for color_func, container in zip(color_funcs, self.containers):
             generator_func = get_log_generator(container)
             prefix = color_func(build_log_prefix(container, prefix_width))
-            yield generator_func(container, prefix, color_func)
+            yield generator_func(container, prefix, color_func, self.cascade_stop)
```
dnephin: Instead of passing in `cascade_stop` here and having to duplicating the "raise exception" logic,  I think you can move it to the last line of `def run()` and it will work.

rbann: I do not like the duplication, but putting it in `run()` would not be OK, as it runs only when the `Multiplexer` exits, which does not happen when only one container exits.

dnephin: It must, because that is what the `cascade_stop` flag is doing, and also matches the behaviour I saw in testing.

rbann: Sure enough, just let the tests run and will push.

dnephin: So if we want to support the case 3 you mention in the issue, I think we'd need to make a couple more changes.

Instead of raising an exception here, I think we would just let this exit. Then from the call in `compose/cli/main.py` we can check the flag, and if it's set, we can run the shutdown from there.

rbann: OK, then I don't push this right now, but will do tests and implement this 'case 3' also.

dnephin:
Thanks for your work on this. We'd like to get this out soon, so I'm going to carry this PR and make a couple small changes to get it merged.

dnephin:
Carried in #2996

PR #2959, opened 2016-02-18
Replace PyYAML with ruamel.yaml

AvdN:
PyYAML supports YAML 1.1 which has features that require users
to quote some input. Main problem are sexagesimals, so that
"- 25:25" is interpreted as "- 1525". In ruamel.yaml this is
intpreted as if it was written "- '25:25'".
In addition the above, Yes/No/On/Off are no longer interpreted
as booleans, and scalars starting with a 0 and consisting of
numbers no longer as octals, but normal integers (use 0o12 if you
need octals).

The YAML 1.2 is implemented in ruamel.yaml for the RoundTripLoader.
In order to preserver safe loader, the VersionedResolver is combined
with the SafeConstructor in SafeLoader_1_2, which is used explicitly
to load YAML files.
A simple test is provided to confirm that sexagesimals no longer
break user input.

(Several lines that I never touched needed to be wrapped to let the
pre-commit hooks pass the flake8 test).

Signed-off-by: Anthon van der Neut a.van.der.neut@ruamel.eu

### DIFF
--- a/compose/cli/utils.py
+++ b/compose/cli/utils.py
@@ -8,6 +8,7 @@ import ssl
 import subprocess
 
 import docker
+import ruamel.yaml
 from six.moves import input
 
 import compose
@@ -68,13 +69,15 @@ def get_version_info(scope):
             "{}\n"
             "docker-py version: {}\n"
             "{} version: {}\n"
-            "OpenSSL version: {}"
+            "OpenSSL version: {}\n"
+            "ruamel.yaml version: {}"
         ).format(
             versioninfo,
             docker.version,
             platform.python_implementation(),
             platform.python_version(),
-            ssl.OPENSSL_VERSION)
+            ssl.OPENSSL_VERSION,
+            ruamel.yaml.__version__)
 
     raise ValueError("{} is not a valid version scope".format(scope))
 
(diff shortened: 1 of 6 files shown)


dnephin:
hmm, it is strange you had to make any file changes for the pre-commit hooks because we run them on every test run with `--all-files`. Maybe it was missing the config for some reason? (where we relax some rules).

I think I might have mistaken what you were proposing. I was hoping that we could keep pyyaml, but adjust the configuration to remove support for sexagesimals.  Is that not possible ?

My understanding is that ruamel.yaml is forked from pyyaml and provides some different class implementations for handling the round trip, is that right?

AvdN:
In addition to the roundtripping and the changes, there were many bugs fixed that were logged/PR-ed against PyYAML but were never acted upon in that project (it has essentially stalled since late 2014). 
Of course this can be done starting with the PyYAML code, but because e.g. the pattern matching rules are globals, applied when PyYAML is loaded, you would have to redo all of that and end up with a few hundred lines of code in docker-compose to make that happen.

AvdN:
I looked at the appveyor output, but it is not obvious to me any of my changes caused this to fail. Should I investigate that more closely?

dnephin:
I don't think we're ready to make such a big change to the yaml parser. I was hoping we could just drop the support for sexagesimals for now, since that's really the only issue we've seen frequently.

PR #2962, opened 2016-02-19
Update docs for volume_driver option

### LINKED ISSUE #2960: volume_driver in service config is unnecessary in v2 format
I believe `volume_driver` is only relevant for named volumes, so the driver would need to be specified in the `volumes` section, not the config section.

I think we can remove `volume_driver` from the service config for v2.

shin-:
Fixes #2960

### DIFF
--- a/docs/compose-file.md
+++ b/docs/compose-file.md
@@ -582,10 +582,11 @@ limit as an integer or soft/hard limits as a mapping.
 ### volumes, volume\_driver
 
 Mount paths or named volumes, optionally specifying a path on the host machine
-(`HOST:CONTAINER`), or an access mode (`HOST:CONTAINER:ro`). Named volumes can
-be specified with the
-[top-level `volumes` key](#volume-configuration-reference), but this is
-optional - the Docker Engine will create the volume if it doesn't exist.
+(`HOST:CONTAINER`), or an access mode (`HOST:CONTAINER:ro`).
+For [version 2 files](#version-2), named volumes need to be specified with the
+[top-level `volumes` key](#volume-configuration-reference).
+When using [version 1](#version-1), the Docker Engine will create the named
+volume automatically if it doesn't exist.
 
 You can mount a relative path on the host, which will expand relative to
 the directory of the Compose configuration file being used. Relative paths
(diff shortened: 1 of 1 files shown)


docs/compose-file.md
```diff
@@ -552,7 +552,7 @@ limit as an integer or soft/hard limits as a mapping.
         soft: 20000
         hard: 40000
 
-### volumes, volume\_driver
+### volumes
 
 Mount paths or named volumes, optionally specifying a path on the host machine
 (`HOST:CONTAINER`), or an access mode (`HOST:CONTAINER:ro`). Named volumes can
```
dnephin: There's a line just below this that says

> top-level `volumes` key, but this is optional - the Docker Engine will create the volume if it doesn't exist.

I think we need to move that to a **Note**: In V1 ...

dnephin:
Thanks for picking this up!

docs/compose-file.md
shin-: Hm, yeah. There's also no top-level volume key in v1, so I think that part needs a small rewrite.

shin-: Here's what I have:

> Mount paths or named volumes, optionally specifying a path on the host machine
(`HOST:CONTAINER`), or an access mode (`HOST:CONTAINER:ro`).

> With the [version 2 file format](#version-2), named volumes must be specified
with the [top-level `volumes` key](#volume-configuration-reference). If you're
using [version 1](#version-1), named volumes will be created implicitly when
declared in this section.

WDYT?

dnephin: works for me

dnephin:
It turns out I was wrong about this. @thaJeztah pointed me at https://github.com/docker/docker/pull/19190

So apparently we need to keep `volume_driver`, but we should document that it only applies to container volumes in v2, and not named volumes.

shin-:
Updated :+1: Now only a docs change.

docs/compose-file.md
```diff
@@ -580,11 +580,16 @@ should always begin with `.` or `..`.
...
-a `volume_driver`.
+If you do not use a host path, you may specify a `volume_driver`.
 
     volume_driver: mydriver
 
+Note that for [version 2 files](#version-2), this driver
+will not apply to named volumes (you should use the `driver` option when
+[declaring the volume](#volume-configuration-reference) instead).
+For [version 1](#version-1), both named volumes and container volumes will
+be afected.
```
aanand: Typo: "afected". But actually, I think better wording would be: "For version 1, both named volumes and container volumes will use the specified driver".

shin-: Fixed :+1:

dnephin:
We lost the fix to the section above where it says "Named volumes can
 be specified with the
 [top-level `volumes` key](#volume-configuration-reference), but this is
 optional"

shin-:
Good point, I updated it again.

dnephin:
LGTM!

Issue #2778, opened 2016-01-29
Add support for tmpfs


cwolfinger:
In the latest release candidate for 1.6.0 the --tmpfs options is not available.  Is it possible to add this to the service definition via some other mechanism.

Deshke:
and it is missing in the latest release and in the docs...

4alldigital:
would also like this as it was my main reason for updating to 1.10. To utilise better readonly environments and not have to expose /tmp directories to the host via volumes.  Any workarounds?

ct-clearhaus:
@4alldigital 
> Any workarounds?

We have been running read only for quite some time and have `/tmp` as a volume when necessary. This is fine since we can accept a "fresh" `/tmp` (and we set the sticky bit in the start script of the app).

PR #2978, opened 2016-02-20
Add support for docker run --tmpfs flag.

sllawap:
This adds support for the docker run --tmpfs flag as described in https://github.com/docker/compose/issues/2778


sllawap:
@GordonTheTurtle -- should be all set now. Sorry for the oversight; first time contributor!

dato:
I tried to test this PR, because I wanted to check whether syntax like `/tmp:size=10M` or `/home/foo/tmp:uid=NNN` would work.

However, I could not make the PR run. I think support for tmpfs is needed in docker-py, maybe?

Consider the following _docker-compose.yml_:

```yaml
version: "2"

services:
  test:
    image: busybox
    tmpfs: /run
    network_mode: none
```

I run: 

```
$ env PYTHONPATH=. bin/docker-compose up test
```

It fails with:

```pytb
  File "bin/docker-compose", line 3, in <module>
    main()
  File ".../src/compose/compose/cli/main.py", line 56, in main
    command.sys_dispatch()
  File ".../src/compose/compose/service.py", line 648, in _get_container_host_config
    tmpfs=options.get('tmpfs'),
  File ".../Library/Python/Envs/compose/lib/python2.7/site-packages/docker/api/container.py", line 141, in create_host_config
    return utils.create_host_config(*args, **kwargs)
TypeError: create_host_config() got an unexpected keyword argument 'tmpfs'
```

I looked at the [AppVeyor logs](https://ci.appveyor.com/project/docker/compose/build/master-1328), but I didn’t see the [`test_tmp()`](https://github.com/docker/compose/pull/2978/files#diff-ebf882803c8eac5a345368609d88cd80R874) test run, only the simpler [`test_tmpfs_option()`](https://github.com/docker/compose/pull/2978/files#diff-02a52ec1e2850804d6071dcb286cfebcR1184). I don’t really know if it the test was supposed to catch it or not.

compose/config/config.py
```diff
@@ -725,6 +729,7 @@ def merge_service_dicts(base, override, version):
         'external_links',
         'ports',
         'volumes_from',
+        'tmpfs',
```
dnephin: Since this can be a list or a string it needs to be in the list with `dns` and `dns_search`.

dnephin:
Thanks for the PR!

I've left one comment about a fix.
You'll also need to rebase on master, squash to one commit, and make sure that commit is signed.

If this requires a specific pre-release version of `docker-py` you can update the required version in `requirements.txt`. We're currently using a pinned version, so you might not need to make any changes there.

dnephin:
Carried in #3137

sllawap:
dnephin@ sorry for neglecting this, I was on vacation. Thank you for pushing this through!

Issue #2993, opened 2016-02-22
Handle booleans in environment/labels/extra_hosts


dnephin:
We've been warning when a boolean value is used in these fields because things like `y`, `yes`, `on`, `off`, are all converted to `True`, which is often not the intent of the user.

The warning says we're going to make it an error in the next release, so we should do that.

I think our options are:
1. make it an error
2. violate the yaml spec, and only accept `True`, and `False` as booleans, and treat everything else as a string.

dnephin:
I think (1) is more correct here, deviating too far from the yaml spec is going to be very confusing.

@aanand , @shin- what do you think?

I'd like to pick this up tomorrow.

aanand:
I agree that we should go with (1). It's very simple and clear to just say "this value should be a string".

Issue #2987, opened 2016-02-22
Err: contains [" item "], which is an invalid type, it should be a string, number, boolean, or a null


de1m:
Hi, I've a docker-compose.yml file 

```
httpd:
image: httpd
hostname: backup-new-test
privileged: 'true'
volumes:
  - /a/data/backup-new-test/httpdocs:/a/http
  - /a/data/backup-new-test/httpconf:/a/conf
restart: always
labels:
  backupage: 6M
  backup: /a/data/backup-new-test
  prerun:
    - touch /tmp/id.test
    - echo test >> /tmp/id.test
  postrun:
    - cat /tmp/id.test
    - rm /tmp/id.test
  strategy: off
  include:
    - /a/data/backup-new-test/inc1
    - /a/data/backup-new-test/inc2
    - /a/data/backup-new-test/inc3
  exclude:
    - /a/data/backup-new-test/ex1
    - /a/data/backup-new-test/ex2
    - /a/data/backup-new-test/ex3
```

when I run docker-compose -f /path/docker-compose.yml I get following error:

```
WARNING: There is a boolean value in the 'environment' key.
Environment variables can only be strings.
Please add quotes to any boolean values to make them string (eg, 'True', 'yes', 'N').
This warning will become an error in a future release.

ERROR: Validation failed in file '/a/run/backup-new-test/docker-compose.yml', reason(s):
httpd.privileged contains an invalid type, it should be a boolean
httpd.labels.prerun contains ["touch /tmp/id.test", "echo test >> /tmp/id.test"], which is an invalid type, it should be a string, number, boolean, or a null
```

If I run command

```
docker-compose -f /path/docker-compose.yml
```

without quotas on privileged or with quotas on prerun I get the same errors.
Please help me!

I use v1.6 of docker-compose with ubuntu 
Linux 3.13.0-78-generic #122-Ubuntu x86_64 x86_64 x86_64 GNU/Linux

dnephin:
1. Remove the quotes around `'true'` in the privileged field, it should be a boolean not a string
2. The warning about "environment" is wrong, that's a bug, I'll open a PR to fix it. The issue is actually with the `labels` section, not `environment` as it claims. I beleive `off` is being treated as a boolean, it needs to be quoted.
3. The error about "httpd.labels.prerun" is correct. You can't use arrays in the labels section, you need to make those literal json strings if you want them to have a json value:

```yaml

labels:
[...]

PR #2995, opened 2016-02-22
Fix warning about boolean values

dnephin:
The warning was only covering the "environment" section, but we use the definition in other sections as well. See #2987

We can fix this better in #2993


dnephin:
unlreated test failure

shin-:
LGTM

PR #2996, opened 2016-02-23
Fix abort on container exit

### LINKED ISSUE #2940: abort-on-container-exit does not work as expected

dnephin:
Carry #2956
Fixes  #2940

### DIFF
--- a/compose/cli/log_printer.py
+++ b/compose/cli/log_printer.py
@@ -5,7 +5,6 @@ import sys
 from itertools import cycle
 
 from . import colors
-from . import signals
 from .multiplexer import Multiplexer
 from compose import utils
 from compose.utils import split_buffer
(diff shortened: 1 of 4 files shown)


shin-:
LGTM

PR #2997, opened 2016-02-23
Fix env var without value

### LINKED ISSUE #2359: Inconsistency in how compose handles empty environment variables

dnephin:
Reverts #2734
Re-fixes #2359 with support for swarm `reschedule:on-node-failure` environment variables.

### DIFF
--- a/compose/config/config.py
+++ b/compose/config/config.py
@@ -512,12 +512,12 @@ def resolve_environment(service_dict):
         env.update(env_vars_from_file(env_file))
 
     env.update(parse_environment(service_dict.get('environment')))
-    return dict(filter(None, (resolve_env_var(k, v) for k, v in six.iteritems(env))))
+    return dict(resolve_env_var(k, v) for k, v in six.iteritems(env))
 
 
 def resolve_build_args(build):
     args = parse_build_arguments(build.get('args'))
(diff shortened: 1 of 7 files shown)


shin-:
LGTM

PR #3005, opened 2016-02-23
Corrected broken swarm documentation link

### LINKED ISSUE #3004: Swarm documentation broken link
The swarm documentation here: https://github.com/docker/compose/blob/master/SWARM.md was recently changed to point to this link: https://docs.docker.com/compose/swarm/ Which gives a 404 not found error

jeffkile:
Fixes #3004 - This is important because this page is the 3rd Google result for "docker compose swarm"

### DIFF
--- a/SWARM.md
+++ b/SWARM.md
@@ -1,1 +1,1 @@
-This file has moved to: https://docs.docker.com/compose/swarm/
+This file has moved to: https://github.com/docker/compose/blob/master/docs/swarm.md


jeffkile:
@aanand Can you take a look at this when you have a chance? Looks like you were the last to modify this link

dnephin:
Sorry, I don't think this change is correct. We're going to publish the docs to `docs.docker.com` later today, which will fix this broken link. 

Thanks for trying to fix it! You're right that this is the new source for the swarm docs, but we want to point at the published version.

PR #3011, opened 2016-02-24
Fix #2804: Add ipv4 and ipv6 static addressing

### LINKED ISSUE #2804: Define static IP address for container
It would be useful to be able to define a static IP address for a container within the `docker-compose.yml`.

I see that `docker run --ip xxx.xxx.xxx.xxx` is now available in [docker 1.10.0-rc2](https://github.com/docker/docker/releases/tag/v1.10.0-rc2). More details in the merged pull request: https://github.com/docker/docker/pull/19001

Example:

    docker network create --driver=bridge --subnet=192.168.0.0/24 --gateway=192.168.0.1 mynet
    docker run --ip=192.168.0.10 --net=mynet -ti some_image

mdaue:
Fixes #2804

Signed-off-by: Matt Daue <mattdaue@gmail.com>

### DIFF
--- a/compose/service.py
+++ b/compose/service.py
@@ -451,7 +451,10 @@ class Service(object):
     def connect_container_to_networks(self, container):
         connected_networks = container.get('NetworkSettings.Networks')
 
-        for network, aliases in self.networks.items():
+        for network, netdefs in self.networks.items():
+            aliases = netdefs.get('aliases', [])
+            ipv4_address = netdefs.get('ipv4_address', None)
+            ipv6_address = netdefs.get('ipv6_address', None)
             if network in connected_networks:
                 self.client.disconnect_container_from_network(
                     container.id, network)
@@ -459,7 +462,9 @@ class Service(object):
             self.client.connect_container_to_network(
                 container.id, network,
                 aliases=list(self._get_aliases(container).union(aliases)),
-                links=self._get_links(False),
+                ipv4_address=ipv4_address,
+                ipv6_address=ipv6_address,
+                links=self._get_links(False)
             )
 
     def remove_duplicate_containers(self, timeout=DEFAULT_TIMEOUT):
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,7 +3,7 @@ cached-property==1.2.0
 dockerpty==0.4.1
 docopt==0.6.1
 enum34==1.0.4
-git+https://github.com/docker/docker-py.git@81d8caaf36159bf1accd86eab2e157bf8dd071a9#egg=docker-py
+git+https://github.com/docker/docker-py.git@d8be3e0fce60fbe25be088b64bccbcee83effdb1#egg=docker-py
 jsonschema==2.5.1
 requests==2.7.0
 six==1.7.3
--- a/compose/config/config_schema_v2.0.json
+++ b/compose/config/config_schema_v2.0.json
@@ -152,7 +152,9 @@
                     {
                       "type": "object",
                       "properties": {
-                        "aliases": {"$ref": "#/definitions/list_of_strings"}
+                        "aliases": {"$ref": "#/definitions/list_of_strings"},
+                        "ipv4_address": {"type": "string"},
+                        "ipv6_address": {"type": "string"}
                       },
                       "additionalProperties": false
                     },
--- a/compose/network.py
+++ b/compose/network.py
@@ -159,26 +159,26 @@ class ProjectNetworks(object):
             network.ensure()
 
 
-def get_network_aliases_for_service(service_dict):
+def get_network_defs_for_service(service_dict):
     if 'network_mode' in service_dict:
         return {}
     networks = service_dict.get('networks', {'default': None})
     return dict(
-        (net, (config or {}).get('aliases', []))
+        (net, (config or {}))
         for net, config in networks.items()
     )
 
 
 def get_network_names_for_service(service_dict):
-    return get_network_aliases_for_service(service_dict).keys()
+    return get_network_defs_for_service(service_dict).keys()
 
 
 def get_networks(service_dict, network_definitions):
     networks = {}
-    for name, aliases in get_network_aliases_for_service(service_dict).items():
+    for name, netdef in get_network_defs_for_service(service_dict).items():
         network = network_definitions.get(name)
         if network:
-            networks[network.full_name] = aliases
+            networks[network.full_name] = netdef
         else:
             raise ConfigurationError(
                 'Service "{}" uses an undefined network "{}"'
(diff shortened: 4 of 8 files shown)


compose/service.py
```diff
@@ -451,16 +452,29 @@ def start_container(self, container):
...
             if network in connected_networks:
                 self.client.disconnect_container_from_network(
                     container.id, network)
 
-            self.client.connect_container_to_network(
-                container.id, network,
-                aliases=list(self._get_aliases(container).union(aliases)),
-                links=self._get_links(False),
-            )
+            if docker.version_info >= (1, 8, 0):
```
dnephin: We're going to set docker-py to a minimum version of 1.8.0, so we don't need this if/else branch.

For now you can set it to a git sha that contains the necessary commit in `requirements.txt`.

dnephin:
Thanks for the PR!

Just the one comment about the docker-py version check, otherwise I think this looks like it's on the right track.

Still needs a test case and docs.

clkao:
What happens when a service with static address gets scaled?  It should either be prevented or we should have some clever way for defining ip address corresponding to the service container index.

mdaue:
Interesting problem, I think it should probably be prevented for now. The main use case for static IP addressing seems to be deterministic kernel filtering to operate services (and possibly DNS). If we circumvent these expected rulesets with automagic IP addressing, it may defeat the purpose.

We could look into doing something with IPVS and perhaps a round-robin scheduling approach of exposing the static IP and scaling using internal IPs via load balancing.

mdaue:
tests pass locally with newer docker engine, I think janky fails with older docker engine: if \'client is newer than server\' in e.explanation:\nTypeError: \'str\' does not support the buffer interface\n'

.cache/v/cache/lastfailed
```diff
@@ -0,0 +1 @@
+{}
```
dnephin: I believe this file was added by mistake. Please `git rm` it.

dnephin:
This is looking good. Just one file that shouldn't be committed, and I think it would be good to squash the commits down to one or two commits.

dnephin:
The test failure is unrelated.

mdaue:
Removed inadvertent commit and squashed. Thanks!

dnephin:
LGTM

requirements.txt
```diff
@@ -3,7 +3,7 @@ cached-property==1.2.0
 dockerpty==0.4.1
 docopt==0.6.1
 enum34==1.0.4
-git+https://github.com/docker/docker-py.git@81d8caaf36159bf1accd86eab2e157bf8dd071a9#egg=docker-py
+git+https://github.com/mdaue/docker-py.git@d4a5bc4a86e9ab69368c80027439bca3ecc1e852#egg=docker-py
```
shin-: Can you comment on that change? What is currently missing from docker-py? Is there a PR for it?

dnephin: Oops, I missed this, good catch @shin- 

It depends on https://github.com/docker/docker-py/pull/936. Looks like that needs a rebase

shin-: Ah, cool. I just merged it. We can use SHA `d8be3e0fce60fbe25be088b64bccbcee83effdb1` for now.

mdaue:
Set requirements.txt to: git+https://github.com/docker/docker-py.git@d8be3e0fce60fbe25be088b64bccbcee83effdb1#egg=docker-py; this on @shin- suggestion

shin-:
LGTM once green :+1:

shin-:
It'll be in 1.7.0 - Target date simultaneous to the [Docker Engine 1.11.0 release](https://github.com/docker/docker/wiki#roadmap)

CRogers:
Is there a beta version that we can use? 3 months is a long time to wait :(

CRogers:
Oh, 04/07/2016 is american dates, only a couple of weeks to wait.

mavenugo:
@shin- @dnephin we should prevent `scale` for the service with static ip-address. I couldn't find that change in this PR. Is that handled ?

aanand:
@mavenugo from my manual testing, it looks like Engine will simply raise an error, which is fine. We could perhaps add a warning in Compose too though, to make the source of the problem clearer.

loomchild:
First of all thanks a lot for implementing this feature.

While testing it I found an issue - when I try to start an application after stopping it, I get the following error:

    docker.errors.APIError: 409 Client Error: Conflict ("b'cannot create network br-777330a67cde (877e0da144dacec77255ffaeee34e99d4d5529179a4e11d1f8e2fb7ca72ad793): conflicts with network 777330a67cdee4012567000bfedcb8d4d77dbf9101f1667cbf4f3dc976895414 (br-777330a67cde): networks have overlapping IPv4'")

Apparently the new network cannot be created second time because it was already created during the first run. However, this works for networks with non-static IPs - new ones are simply ignored.

mdaue:
@loomchild trying to duplicate your issue, could you please note your docker engine and docker-py versions?

mdaue:
also, if you can share your docker-compose.yml that would be helpful - thanks!

loomchild:
@mdaue I think I was too fast reporting a bug, I can't reproduce the error anymore. 

I was playing with different project names and it only occurs if I have two projects with different names trying to create network with the same range, which is expected I suppose.

As for the versions:

    Docker version 1.10.2, build c3959
    Docker-Compose commit 94b1862579049585c48a9f4039f275c02ebeb234 
    Docker-Py commit 7befe694bd21e3c54bb1d7825270ea4bd6864c13

Sorry for the noise.

mdaue:
No problem! Two projects accessing the same network is something that would have to be addressed in the ipam-handling code portion I think.

fordodone:
Thank you so much! works great engine 1.10.3 20f81dd compose 1.7.0dev 85e2fb6

daniel-smith-kaazing:
Is there a new ETA? I see Docker 1.10.0 is out as of yesterday and there is a 1.7.0 RC for Compose that supports this feature. When do we expect the RC to become an R, so to speak?

blah238:
@daniel-smith-kaazing https://github.com/docker/compose/releases/tag/1.7.0 was released today.

daniel-smith-kaazing:
Great news! Thank you!

*Dan Smith*

Senior Support Engineer  |  Phone: 650.960.8155 | Mobile: 408.835.1167
daniel.smith@kaazing.com <dorris.obrien@kaazing.com>

*------------------------------------------------------------*

KAAZING >|<
When Real-time Matters
Kaazing.com <http://www.kaazing.com/>  |  Download
<http://kaazing.com/download.html>  |  Blog <http://blog.kaazing.com/>  |
Twitter <https://twitter.com/kaazing>
*http://kaazing.com <http://kaazing.com/about-us/careers/>*

PR #3022, opened 2016-02-25
Update note about build + image with extends

### LINKED ISSUE #3018: `extends` with both `image` and `build` now work: update `extends` doc
[`extends` doc](https://docs.docker.com/compose/extends/) explains specific merge rules that apply to `image` and `build`:

> In the case of build and image, using one in the local service causes Compose to discard the other, if it was defined in the original service.

Since compose 1.6.0 and Compose File format v2 this limitation is not needed anymore because both `image` and `build` may be defined at the same time.
After testing it appears that `extends` has already relaxed this and merges with usual rules.

The documentation should be updated to reflect this.

aanand:
Closes #3018

### DIFF
--- a/docs/extends.md
+++ b/docs/extends.md
@@ -290,31 +290,17 @@ replaces the old value.
     # result
     command: python otherapp.py
 
-In the case of `build` and `image`, using one in the local service causes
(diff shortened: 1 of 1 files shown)


dnephin:
LGTM

PR #3030, opened 2016-02-26
Document ways to make services wait for dependencies

aanand:
- Moved the section to the top of the FAQ, on the assumption that it'll help people landing from searches
- Reword the question in the hope that it'll catch people searching for "startup order" or similar 
- Replace the paragraphs describing health checks with example implementations
- Also updated the next question while I was there, now that we support `stop_signal`

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -15,7 +15,7 @@ weight=90
 If you don’t see your question here, feel free to drop by `#docker-compose` on
 freenode IRC and ask the community.
 
-## Why do my services take 10 seconds to stop?
+## Why do my services take 10 seconds to recreate or stop?
 
 Compose stop attempts to stop a container by sending a `SIGTERM`. It then waits
 for a [default timeout of 10 seconds](./reference/stop.md).  After the timeout,
@@ -39,6 +39,12 @@ in your Dockerfile.
 
 * If you are able, modify the application that you're running to
 add an explicit signal handler for `SIGTERM`.
+
+* Set the `stop_signal` to a signal which the application knows how to handle:
+
+      web:
+        build: .
+        stop_signal: SIGINT
 
 * If you can't modify the application, wrap the application in a lightweight init
 system (like [s6](http://skarnet.org/software/s6/)) or a signal proxy (like
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -15,6 +15,76 @@ weight=90
 If you don’t see your question here, feel free to drop by `#docker-compose` on
 freenode IRC and ask the community.
 
+
+## How do I control the order of service startup? I need my database to be ready before my application starts.
+
+You can control the order of service startup with the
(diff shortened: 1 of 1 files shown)


docs/faq.md
```diff
@@ -40,6 +111,12 @@ in your Dockerfile.
 * If you are able, modify the application that you're running to
 add an explicit signal handler for `SIGTERM`.
 
+* Set the `stop_signal` to a signal which the application knows how to handle:
+
+      web:
+        build: .
+        stop_signal: SIGTERM
```
dnephin: `SIGTERM` is already the default, so I think this should be `SIGKILL` or `SIGINT`?

aanand: Oh whoops, I meant to put SIGINT yes

docs/faq.md
```diff
@@ -15,7 +15,78 @@ weight=90
...
+To handle this, your application should attempt to re-establish a connection to
+the database after a failure. If the application retries the connection,
+it should eventually be able to connect to the database.
+
+There are several ways to implement this:
+
+-   Use a tool such as [wait-for-it](https://github.com/vishnubob/wait-for-it)
+    or [dockerize](https://github.com/jwilder/dockerize). These are small
+    wrapper scripts which you can include in your application's image and will
+    wait for TCP to be available on a given host and port.
```
dnephin: I think this should be "a TCP socket" or something like that. Feels like there is a missing word here. Maybe even just "wait for service to be available".

docs/faq.md
```diff
@@ -15,7 +15,78 @@ weight=90
...
+The problem of waiting for a database to be ready is really just a subset of a
+much larger problem of distributed systems. In production, your database could
+become unavailable or move hosts at any time. Your application needs to be
+resilient to these types of failures.
+
+To handle this, your application should attempt to re-establish a connection to
+the database after a failure. If the application retries the connection,
+it should eventually be able to connect to the database.
+
+There are several ways to implement this:
```
dnephin: Technically these examples are not of application level retries, but workarounds for applications that don't properly retry.

dnephin:
LGTM when green

docs/faq.md
```diff
@@ -15,7 +15,77 @@ weight=90
...
+            build: .
+            ports:
+              - "80:8000"
+            depends_on:
+              - "db"
+            entrypoint: ./wait-for-it.sh db:5432
+          db:
+            image: postgres
+
+-   Write yor own wrapper script to perform a more application-specific health
```
simonvanderveldt: Minor typo here "yor"

simonvanderveldt:
@aanand Nice to see this addressed!

docs/faq.md
aanand: Thanks! Fixed it.

dnephin:
LGTM

bfirsh:
NICE! Missed this. :sparkling_heart:

PR #3032, opened 2016-02-26
Support multiple files in COMPOSE_FILE env var

### LINKED ISSUE #2395: Suggestion to use an env variable for -f file selection
The current option -f is great but error prone when deploying in production.

how about using a environment variable - something like 

$DOCKER-COMPOSER-ENV=prod

if not set  use docker-compose.yml and  show a warning no env set so using only the main file .....
if set use docker-compose.yml and extend using a file  based on the env 
docker-compose-$DOCKER-COMPOSER-ENV.yml 

$DOCKER-COMPOSER-ENV=prod
will run:
docker-compose -f docker-compose.yml -f docker-compose-prod.yml

dnephin:
Fixes #2395

This should make it easier to use multiple compose files when the files aren't the default.

### DIFF
--- a/compose/cli/command.py
+++ b/compose/cli/command.py
@@ -58,8 +58,10 @@ def get_config_path_from_options(options):
     if file_option:
         return file_option
 
-    config_file = os.environ.get('COMPOSE_FILE')
-    return [config_file] if config_file else None
+    config_files = os.environ.get('COMPOSE_FILE')
+    if config_files:
+        return config_files.split(os.pathsep)
+    return None
 
 
 def get_client(verbose=False, version=None):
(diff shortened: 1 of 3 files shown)


compose/const.py
```diff
@@ -15,6 +15,8 @@
 LABEL_VERSION = 'com.docker.compose.version'
 LABEL_CONFIG_HASH = 'com.docker.compose.config-hash'
 
+PATH_SEPARATOR = ';' if IS_WINDOWS_PLATFORM else ':'
```
aanand: I think we can use [`os.pathsep`](https://docs.python.org/2/library/os.html#os.pathsep) here?

> The character conventionally used by the operating system to separate search path components (as in `PATH`), such as `':'` for POSIX or `';'` for Windows. Also available via `os.path`.

dnephin: Oh nice, yes definitely. I couldn't find this earlier.

dnephin:
Now using `os.pathsep`

dnephin:
Test failure is unrelated

aanand:
LGTM

PR #3034, opened 2016-02-27
check --file/-f argument in run.sh

2k0ri:
It makes `docker-compose.yml` at any path be accessible without setting `COMPOSE_FILE`.

### DIFF
--- a/script/run/run.sh
+++ b/script/run/run.sh
@@ -34,10 +34,22 @@ fi
 if [ "$(pwd)" != '/' ]; then
     VOLUMES="-v $(pwd):$(pwd)"
 fi
+ARGS="$@"
+while (( $# > 0 )); do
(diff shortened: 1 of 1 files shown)


script/run/run.sh
```diff
@@ -34,10 +34,22 @@ fi
...
+        if [[ "$1" =~ = ]]; then
+            COMPOSE_FILE="${1#*=}"
+        else
+            COMPOSE_FILE="$2"
+        fi
+        break
+    fi
+    shift
+done
+set -- "$ARGS"
```
dnephin: Why ` =~ ^-f` instead of ` == "-f*"` ?

It would be great if this iteration could be done without having to shift off all the items.

dnephin:
Thanks for the PR. I have a couple questions about the implementation.

ijc:
PR grooming: no response from submitter.

PR #3039, opened 2016-02-28
Allowing null for build args

### LINKED ISSUE #3020: Build arguments with only a key don't work with object syntax, contradicting  documentation
[`build.args` doc](https://docs.docker.com/compose/compose-file/#args) states:

> Build arguments with only a key are resolved to their environment value on the machine Compose is running on.

It's the same feature as for [`environment`](https://docs.docker.com/compose/compose-file/#environment).

However, it doesn't work with object syntax:
```yaml
version: '2'
services:
  test:
    build:
      context: .
      args:
        BLA:
```
```shell
$ BLA=foo docker-compose config
services.test.build.args contains null, which is an invalid type, it should be a string, or a number
```
It should work (and maybe it shouldn't not accept a number, I don't know).

Array syntax does work:
```yaml
version: '2'
services:
  test2:
    build:
      context: .
      args:
        - BLA
```
```shell
$ BLA=foo docker-compose config
networks: {}
services:
  test2:
    build:
      args:
        BLA: foo
      context: /xxx
version: '2.0'
volumes: {}
```

Also, `build.args` doc is missing example for these specific points. `environment` has nice examples for that.

dbonev:
Fixes #3020

### DIFF
--- a/compose/config/config_schema_v2.0.json
+++ b/compose/config/config_schema_v2.0.json
@@ -58,20 +58,7 @@
               "properties": {
                 "context": {"type": "string"},
                 "dockerfile": {"type": "string"},
-                "args": {
-                  "oneOf": [
-                    {"$ref": "#/definitions/list_of_strings"},
-                    {
-                      "type": "object",
-                      "patternProperties": {
-                        "^.+$": {
-                          "type": ["string", "number"]
-                        }
-                      },
-                      "additionalProperties": false
-                    }
-                  ]
-                }
+                "args": {"$ref": "#/definitions/list_or_dict"}
               },
               "additionalProperties": false
             }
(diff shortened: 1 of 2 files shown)


dnephin:
Thanks for the PR!

Change looks correct. A quick unit test in `tests/unit/config/config_test.py` would be great.

I believe the current test failures are unrelated.

dnephin:
Looking at this again, I think the right fix might actually be to use `{$ref: #/definitions/list_or_dict},`. We're removing `boolean` from the accepted values, and at that point it should be identical to the schema here.

dbonev:
Done. Thx @dnephin

dnephin:
LGTM

shin-:
LGTM!

PR #3053, opened 2016-03-01
Don't allow booleans for mapping types

### LINKED ISSUE #2993: Handle booleans in environment/labels/extra_hosts

dnephin:
Fixes #2993

### DIFF
--- a/compose/config/config_schema_v1.json
+++ b/compose/config/config_schema_v1.json
@@ -156,8 +156,7 @@
           "type": "object",
           "patternProperties": {
             ".+": {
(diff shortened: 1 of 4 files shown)


dnephin:
test failure is unrelated

aanand:
LGTM

PR #2055, opened 2015-09-16
Fix stacktrace when ctrl-c stops logs

dnephin:
Fixes #189 

I had to test this manually with the binary, but I wasn't able to reproduce the error after this change.  Without the change I would get it ~80% of the runs (if I use the binary).

It's not pretty, but I'm not sure what else we can do about it.

PR #3056, opened 2016-03-01
Fix signal handling with pyinstaller.

### LINKED ISSUE #2904: Pyinstaller has issues with signals

dnephin:
Fixes #2904

In #2055 I added this `KeyboardInterrupt` to handle the pyinstaller issue of raising `thread.error` for some unknown reasons (probably a bug in pyinstaller). At the time this seemed to work, but after changing the signal handler logic to use exceptions, raising `KeyboardInterrupt` was no longer correct.

By instead raising  `ShutdownException` we drop into the correct try/except block when this bug occurs.

I've also adding a similar `try/except` for `thread.error` in the `parallel` module, since we use threads there as well, and we started to see the same issue during force shutdown.

This issue is difficult to test because it open happens with the pyinstaller binary, and even then it's not consistent. I did some manual testing and I wasn't able to reproduce the issue after making this change.

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -54,7 +54,7 @@ def main():
     try:
         command = TopLevelCommand()
         command.sys_dispatch()
-    except KeyboardInterrupt:
+    except (KeyboardInterrupt, signals.ShutdownException):
         log.error("Aborting.")
         sys.exit(1)
     except (UserError, NoSuchService, ConfigurationError) as e:
--- a/compose/parallel.py
+++ b/compose/parallel.py
@@ -6,9 +6,11 @@ import sys
 from threading import Thread
 
 from docker.errors import APIError
+from six.moves import _thread as thread
 from six.moves.queue import Empty
 from six.moves.queue import Queue
 
+from compose.cli.signals import ShutdownException
 from compose.utils import get_output_stream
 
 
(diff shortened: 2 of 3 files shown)


compose/cli/main.py
```diff
@@ -54,7 +54,7 @@ def main():
     try:
         command = TopLevelCommand()
         command.sys_dispatch()
-    except KeyboardInterrupt:
+    except (KeyboardInterrupt, signals.ShutdownException):
```
dnephin: This is added for the `logs` case, where we haven't setup any special try/except for this exception, so we can default to just treating it like a SIGINT.

compose/parallel.py
```diff
@@ -68,6 +61,23 @@ def parallel_execute(objects, func, index_func, msg):
...
+
+    # TODO: limit the number of threads #1828
+    for obj in objects:
+        t = Thread(
+            target=perform_operation,
+            args=(func, obj, q.put, index_func(obj)))
+        t.daemon = True
+        t.start()
+
+    return q
```
dnephin: This was extracted to reduce the complexity of `parallel_execute()` to keep it below the current threshold of 11. Adding the extra `except` brought it to 12.

shin-:
LGTM!

aanand:
LGTM

PR #3057, opened 2016-03-01
Add support for a default environment file

### LINKED ISSUE #2636: allow using variables from docker-compose.env in volume paths (and other configs in docker-compose.yml)
I know that it is already possible to use shell environment variables which will expand in the definition of e.g. the path of a mount volume

    volumes:
        - "$SOLR_HOME:/opt/solr/home"

But to set all the environment variables I need, this forces me to write some shell wrappers around the invocation of docker-compose (a one liner, admittedly, but I am a lazy bum...)

Since I already use a docker-compose.env file to define variables which are used in the construction of the containers, I wonder:
would it make sense to expand the variables in the docker-compose.env file so that they are available in docker-compose.yml ? (apart from an obvious loop problem of using them to define the value for 'env_file'...)

shin-:
Fixes #2636

Load and parse environment file from working dir.

~~No tests or docs yet.~~

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -10,7 +10,7 @@
     - id: end-of-file-fixer
     - id: flake8
     - id: name-tests-test
-      exclude: 'tests/(integration/testcases\.py)|(helpers\.py)'
+      exclude: 'tests/integration/testcases\.py'
     - id: requirements-txt-fixer
     - id: trailing-whitespace
 - repo: git://github.com/asottile/reorder_python_imports
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -10,7 +10,7 @@
     - id: end-of-file-fixer
     - id: flake8
     - id: name-tests-test
-      exclude: 'tests/(helpers\.py|integration/testcases\.py)'
+      exclude: 'tests/(integration/testcases\.py)|(helpers\.py)'
     - id: requirements-txt-fixer
     - id: trailing-whitespace
 - repo: git://github.com/asottile/reorder_python_imports
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -10,7 +10,7 @@
     - id: end-of-file-fixer
     - id: flake8
     - id: name-tests-test
-      exclude: 'tests/integration/testcases\.py'
+      exclude: 'tests/(integration/testcases\.py|helpers\.py)'
     - id: requirements-txt-fixer
     - id: trailing-whitespace
 - repo: git://github.com/asottile/reorder_python_imports
--- a/compose/cli/command.py
+++ b/compose/cli/command.py
@@ -34,7 +34,7 @@ def get_config_path_from_options(base_dir, options):
     if file_option:
         return file_option
 
-    environment = config.environment.Environment(base_dir)
+    environment = config.environment.Environment.from_env_file(base_dir)
     config_files = environment.get('COMPOSE_FILE')
     if config_files:
         return config_files.split(os.pathsep)
@@ -58,7 +58,7 @@ def get_project(project_dir, config_path=None, project_name=None, verbose=False,
     config_details = config.find(project_dir, config_path)
     project_name = get_project_name(config_details.working_dir, project_name)
     config_data = config.load(config_details)
-    environment = config.environment.Environment(project_dir)
+    environment = config.environment.Environment.from_env_file(project_dir)
 
     api_version = environment.get(
         'COMPOSE_API_VERSION',
@@ -75,7 +75,7 @@ def get_project_name(working_dir, project_name=None):
     def normalize_name(name):
         return re.sub(r'[^a-z0-9]', '', name.lower())
 
-    environment = config.environment.Environment(working_dir)
+    environment = config.environment.Environment.from_env_file(working_dir)
     project_name = project_name or environment.get('COMPOSE_PROJECT_NAME')
     if project_name:
         return normalize_name(project_name)
--- a/compose/cli/command.py
+++ b/compose/cli/command.py
@@ -9,6 +9,7 @@ import six
 
 from . import verbose_proxy
 from .. import config
+from ..config.environment import Environment
 from ..const import API_VERSIONS
 from ..project import Project
 from .docker_client import docker_client
@@ -34,7 +35,7 @@ def get_config_path_from_options(base_dir, options):
     if file_option:
         return file_option
 
-    environment = config.environment.Environment.from_env_file(base_dir)
+    environment = Environment.from_env_file(base_dir)
     config_files = environment.get('COMPOSE_FILE')
     if config_files:
         return config_files.split(os.pathsep)
@@ -58,7 +59,7 @@ def get_project(project_dir, config_path=None, project_name=None, verbose=False,
     config_details = config.find(project_dir, config_path)
     project_name = get_project_name(config_details.working_dir, project_name)
     config_data = config.load(config_details)
-    environment = config.environment.Environment.from_env_file(project_dir)
+    environment = Environment.from_env_file(project_dir)
 
     api_version = environment.get(
         'COMPOSE_API_VERSION',
@@ -75,7 +76,7 @@ def get_project_name(working_dir, project_name=None):
     def normalize_name(name):
         return re.sub(r'[^a-z0-9]', '', name.lower())
 
-    environment = config.environment.Environment.from_env_file(working_dir)
+    environment = Environment.from_env_file(working_dir)
     project_name = project_name or environment.get('COMPOSE_PROJECT_NAME')
     if project_name:
         return normalize_name(project_name)
--- a/compose/cli/command.py
+++ b/compose/cli/command.py
@@ -21,7 +21,7 @@ log = logging.getLogger(__name__)
 def project_from_options(project_dir, options):
     return get_project(
         project_dir,
-        get_config_path_from_options(options),
+        get_config_path_from_options(project_dir, options),
         project_name=options.get('--project-name'),
         verbose=options.get('--verbose'),
         host=options.get('--host'),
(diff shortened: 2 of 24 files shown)


compose/config/environment.py
```diff
@@ -0,0 +1,69 @@
...
+                self.missing_keys.append(key)
+
+            return ""
+
+
+class Environment(BlankDefaultDict):
+    __instance = None
+
+    @classmethod
+    def get_instance(cls, base_dir='.'):
```
shin-: The first version of this method had base_dir be `None` by default, and fail if it was called for the first time with no value. However, this caused errors with our test suite which does not necessarily calls the code through expected paths.

Current directory seems to be a sane default for 99.9% of cases.

.pre-commit-config.yaml
```diff
@@ -11,6 +11,7 @@
     - id: flake8
     - id: name-tests-test
       exclude: 'tests/integration/testcases.py'
+      exclude: 'tests/helpers.py'
```
dnephin: I don't think you can have two keys with the same value, can you?

dnephin:
Hmm, ya that's an interesting problem. We want to be able to set the project name and compose file from the `docker-compose.env` file, **but** in order to find the `docker-compose.env` we need to know the project directory, which is determined from the location of the compose file...

So we have a cyclic dependency that we can't really resolve. Maybe this isn't going to work for setting the `COMPOSE_FILE` value.

dnephin:
I would really like to avoid singletons, they are a mess to test and can make things a lot harder to refactor. If we remove the cyclic dependency problem from above, and just use the project directory as the location, do you think we could pass around an instance of this environment instead?

We could make it part of the `ConfigDetails`, which I believe is already passed everywhere.

shin-:
We could pass around an instance of the computed environment - it just means carrying an additional argument in a lot of functions in the `config` package, but maybe that's preferrable to the singleton approach.

dnephin:
If we add it to `ConfigDetails`, we'll only need to add an argument to: `process_config_file()`, `interpolate_config_section()`, and `interpolate_environment_variables()`. Three functions doesn't seem so bad. I would definitely prefer that over a singleton (which requires patching dozens of tests).

dnephin:
Oh, there is also `resolve_env_var()`. That adds another 5 or so I think.

shin-:
This should look saner now.

shin-:
Not sure why this specific test seems to fail on AppVeyor though. Works fine both on Janky and locally.

compose/config/config.py
```diff
@@ -302,9 +315,8 @@ def load(config_details):
         config_details.config_files, 'get_networks', 'Network'
     )
     service_dicts = load_services(
-        config_details.working_dir,
-        main_file,
-        [file.get_service_dicts() for file in config_details.config_files])
+        config_details, main_file,
+    )
```
dnephin: minor: this can be one line now

compose/config/environment.py
```diff
@@ -0,0 +1,51 @@
...
+                log.warn(
+                    "The {} variable is not set. Defaulting to a blank string."
+                    .format(key)
+                )
+                self.missing_keys.append(key)
+
+            return ""
+
+
+class Environment(BlankDefaultDict):
```
dnephin: I think we can make this just one class `Environment`, since `BlankDefaultDict` isn't used anymore

compose/config/environment.py
```diff
@@ -0,0 +1,51 @@
...
+    def __init__(self, base_dir):
+        super(Environment, self).__init__()
+        if base_dir:
+            self.load_environment_file(os.path.join(base_dir, '.env'))
+        self.update(os.environ)
```
dnephin: Instead of doing this in a constructor (which makes it harder to unit test), could we make this a `@classmethod` like `from_env_file(dir)` ?

I think it could replace `load_environment_file()` (with basically the same implementation).

compose/config/environment.py
```diff
@@ -0,0 +1,51 @@
...
+        mapping = {}
+        with open(path, 'r') as f:
+            for line in f.readlines():
+                line = line.strip()
+                if '=' not in line:
+                    raise ConfigurationError(
+                        'Invalid environment variable mapping in env file. '
+                        'Missing "=" in "{0}"'.format(line)
+                    )
+                mapping.__setitem__(*line.split('=', 1))
```
dnephin: Could this use `compose.config.config.env_vars_from_file()` ? I think the format should be the same.

compose/config/environment.py
dnephin: I think you'll be able to drop the `if base_dir` check, and instead just use `Environment({})` for testing.

dnephin:
I think this is looking good. Should be ready for tests and docs, just a couple comments.

compose/cli/command.py
```diff
@@ -47,18 +47,19 @@ def friendly_error_message():
...
 
 
-def get_config_path_from_options(options):
+def get_config_path_from_options(base_dir, options):
     file_option = options.get('--file')
     if file_option:
         return file_option
 
-    config_files = os.environ.get('COMPOSE_FILE')
+    environment = config.environment.Environment.from_env_file(base_dir)
```
aanand: This is a bit unwieldy - maybe we should import `Environment` at the top?

aanand:
This is looking nice.

What are the exact semantics of setting `COMPOSE_FILE` in the shell (or `--project-name` at the command line) versus setting it in the `.env` file, and how it affects where we search for `.env` and `docker-compose.yml`? We should document it clearly - it's going to confuse people.

dnephin:
After reading #3125 I just realize I think there is one spot missing. Since we're supporting the project name and filename from the `.env` file, we should also support the `DOCKER_*` and `COMPOSE_API_VERSION` env variables from the `.env` file as well.

I'm not sure if that is going to require a change the `docker-py` from_env function or not (to pass in a different dict).  We should  at least ticket that to make sure we get it into 1.7.0, if it's not part of this branch.

shin-:
I'm not sure what the use case is for having project-specific `DOCKER_*` environment vars. `COMPOSE_API_VERSION` is also dubious IMO, but I guess you could make a case for it?

shin-:
Actually COMPOSE_API_VERSION is supported. COMPOSE_HTTP_TIMEOUT is the one that's missing that I agree is an annoyance, but it should be fixable.

dnephin:
It's not necessarily that it's project specific, it's more that it's a way to persist the settings. I could see users just copying or linking a `docker-compose.env` for all of their projects so they don't have to eval anything.

For me, the biggest argument is more about consistency. We should support setting all the environment variables that compose supports (and we document).

aanand:
Agree. Plus I think there are valid use cases for project-specific `DOCKER_*` variables. You might have different apps that you're developing on different VMs with different versions of Engine, for example, or you might have an app that you're developing against a remote Engine so you can share it with people.

compose/cli/command.py
```diff
@@ -20,18 +21,19 @@
 def project_from_options(project_dir, options):
     return get_project(
         project_dir,
-        get_config_path_from_options(options),
+        get_config_path_from_options(project_dir, options),
         project_name=options.get('--project-name'),
         verbose=options.get('--verbose'),
     )
```
dnephin: It looks like all the calls to `Environment.from_env_file()` come from here. Instead of re-parsing the file multiple times, I think we could do it once here, and pass it through.

dnephin: We can also pass it to `config.find()`

shin-:
Updated to address https://github.com/docker/compose/pull/3057#discussion_r56228654

shin-:
Added support for `DOCKER_*` variables in `.env` file. Relies on docker/docker-py#1006

compose/config/environment.py
```diff
@@ -0,0 +1,74 @@
...
+            k, v = split_env(line)
+            env[k] = v
+    return env
+
+
+class Environment(dict):
+    def __init__(self, *args, **kwargs):
+        super(Environment, self).__init__(*args, **kwargs)
+        self.missing_keys = []
+        self.update(os.environ)
```
dnephin: I think this should happen in `from_env_file()` instead of the constructor. Generally having a constructor do any operation makes it a lot harder to test that class.

shin-: I don't think it'd be correct in that case - we want to be able to instantiate `Environment()` and have a copy of the current `os.environ`.

dnephin: You can just call the update before calling `cls()`

shin-: I mean outside of `from_env_file()`

dnephin: I don't understand, but I guess it's not a big deal. We can always change it later if necessary.

shin-:
It took a few tries, but the Windows issue is finally fixed!

compose/config/environment.py
shin-: Cases like this one: https://github.com/docker/compose/pull/3057/files#diff-44eaaf84aa39ec1255a2dd028a4e4e89R64

dnephin: Since that's only used during test, we don't actually make use of the `os.environ()` right ?

Even if we patched it, the patching context would be out of scope I think?

shin-:
Rebased.

dnephin:
LGTM

adrianolek:
https://docs.docker.com/compose/env-file/ says
> Compose supports declaring default environment variables in an environment file named .env and placed in the **same folder as your compose file**.

I have all my docker config in a `docker/` subdirectory and when I try to run docker-compose from the parent directory it doesn't seem to use the .env file.

```
/app_dir$ cat docker/.env
COMPOSE_PROJECT_NAME=foo-project
/app_dir$ docker-compose -f docker/docker-compose.yml up
Starting docker_node_1        # doesn't use COMPOSE_PROJECT_NAME
/app_dir$ cd docker
/app_dir/docker$ docker-compose -f docker-compose.yml up
Starting fooproject_node_1    # uses COMPOSE_PROJECT_NAME
...
```

Seems like either the documentation is wrong, and should say:
> Compose supports declaring default environment variables in an environment file named .env and placed in the **current directory**.

Or the implementation is wrong.

compose/config/environment.py
```diff
@@ -0,0 +1,93 @@
...
+        super(Environment, self).__init__(*args, **kwargs)
+        self.missing_keys = []
+
+    @classmethod
+    def from_env_file(cls, base_dir):
+        def _initialize():
+            result = cls()
+            if base_dir is None:
+                return result
+            env_file_path = os.path.join(base_dir, '.env')
```
laggingreflex: Is it possible to read the `env_file` setting here first? In case you might have a different `.env` filename (like `dev.env`) or multiple .env files?

PR #3063, opened 2016-03-02
add support for multiple compose files to bash completion

albers:
Since 1.6.0, Compose supports multiple compose files specified with `-f`.
These need to be passed to the docker invocations done by the completion.

/cc @sdurrheimer does this apply to zsh completion as well?


dnephin:
LGTM

shin-:
LGTM.

PR #3066, opened 2016-03-02
Don't call create on existing volumes

### LINKED ISSUE #3062: Only log volume create message if a volume is created
See https://github.com/docker/compose/pull/2883#issuecomment-190189471

shin-:
Fixes #3062

### DIFF
--- a/compose/volume.py
+++ b/compose/volume.py
@@ -3,7 +3,6 @@ from __future__ import unicode_literals
 
 import logging
 
-from docker.errors import APIError
 from docker.errors import NotFound
 
 from .config import ConfigurationError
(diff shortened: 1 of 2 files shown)


compose/volume.py
```diff
@@ -97,28 +97,32 @@ def initialize(self):
...
+                        'Creating volume "{0}" with {1} driver'.format(
+                            volume.full_name, volume.driver or 'default'
+                        )
                     )
-                )
-                volume.create()
+                    volume.create()
+                else:
+                    driver = volume.inspect()['Driver']
+                    if driver != volume.driver:
```
dnephin: Won't this fail in the case where no driver is specified, since `local != None` ?

shin-: Good catch. Updated + added a test to verify the behavior.

dnephin:
LGTM when build is green

shin-:
Test failure seems unrelated.

```
________________________ CLITestCase.test_pause_unpause ________________________

tests/acceptance/cli_test.py:1143: in test_pause_unpause

    self.dispatch(['up', '-d'], None)

tests/acceptance/cli_test.py:128: in dispatch

    return wait_on_process(proc, returncode=returncode)

tests/acceptance/cli_test.py:48: in wait_on_process

    assert proc.returncode == returncode

E   AssertionError: assert 1 == 0

E    +  where 1 = <subprocess.Popen object at 0x7f0896f10090>.returncode
```

Issue #3061, opened 2016-03-02
Make installer instructions great again


JLarky:
I want to say sorry that I didn't look into better place to say that, but I hope it will work out :)

I think that compose install instructions can be made more clear --- https://docs.docker.com/compose/install/

instead of talking about `sudo -i` and `exit` what if you would just use these commands:

```
$ curl -L https://github.com/docker/compose/releases/download/1.6.2/docker-compose-`uname -s`-`uname -m` > docker-compose
$ chmod +x docker-compose
$ sudo mv docker-compose /usr/local/bin/docker-compose
```

dnephin:
This sounds like an improvement to me.

PR #3069, opened 2016-03-03
Update install.md

cloojure:
Showed required use of sudo.

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -39,14 +39,16 @@ which the release page specifies, in your terminal.
 
      The following is an example command illustrating the format:
 
-        curl -L https://github.com/docker/compose/releases/download/1.6.2/docker-compose-`uname -s`-`uname -m` > /usr/local/bin/docker-compose
+        $ sudo su
+        $ curl -L https://github.com/docker/compose/releases/download/1.6.2/docker-compose-`uname -s`-`uname -m` > /usr/local/bin/docker-compose
+        $ exit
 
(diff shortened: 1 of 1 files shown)


dnephin:
I think #3061 is a better suggestion

JLarky:
and `sudo su` is frowned upon, that's why article itself mentions `sudo -i` just paragraph above

cloojure:
Hmmm, I'm not familiar with `sudo -i` vs `sudo su`.  Setting up the env from .profile & friends doesn't seem to affect it either way, does it?

JLarky:
all of this stuff is not needed anyway, just use `curl -o` https://github.com/docker/compose/pull/3076#issuecomment-192082391

JLarky:
as for `sudo -i`, it's just proper way to get into root, it is not big a deal in this case, but doing `sudo su` is just teaching kids bad words, so that later they would have to ask "isn't sudo su ok? I've seen some people in the internets do that". You can read more on topic [here](http://askubuntu.com/questions/376199/sudo-su-vs-sudo-i-vs-sudo-bin-bash-when-does-it-matter-which-is-used), it also has some links in it that could be helpful.

cloojure:
Oh, I see using "sudo curl ... -o /usr/local/bin/docker-compose" does away with the problematic shell redirect all together. OK.  #3076

dnephin:
Closing in favour of the suggestions in #3061

PR #3075, opened 2016-03-04
Display logs for new containers started after the command is run

### LINKED ISSUE #1115: ''docker-compose logs'' does not show logs for new containers
How to reproduce:

shell 1> docker-compose scale xxx=1
shell 2> docker-compose logs xxx
shell 1> docker-compose scale xxx=2
shell 1> (make xxx service produce logs in second container)

expected: shell 2 must show ''xxx_2 | ..." logs as well

### LINKED ISSUE #1866: logs doesn't exit if the container is already stopped
Related to #1858 (cc @aanand, @mnowster since this just came up)

If we attach the log printer to a container that has already exited, it never gets the "end of log" condition and `docker-compose logs` doesn't exist (the same is true for `up`).

In a way, this is related to the problem described in #1115, where logs for containers created by scale are not picked up.

I think we can fix both of those issues by adjusting the way we find containers to log. Instead of sending in a list of containers, we can do something like this:

1. start a thread which tails events (related to #1510)
2. when it receives a "create" event that matches the service name we want to print logs for, create a new thread to consume logs for that container
3. (in the case of `up`) start the containers
4. proceed with log tailing until all containers have exited (which will change the exit condition as well I think)

dnephin:
Fixes #1115
Fixes #1866

`logs -f` and `up` will now show logs for all containers, not just those that started before the command was run. They exit when the last container has exited and there is nothing left running. When `--abort-on-container-exit` is used, it will still exit after the first container exits.

Note 1:

When used with `up` the initial_containers use `attach`, but any new containers use the `logs` API call. This is necessary because there is no way to reliably attach to the container **before** the start event happens. This means that logs from previous runs for these containers will show up when the container is started. This is odd, but I don't see any way around it. The alternative is to use attach, but that would result in missing any logs that happen before the `attach` call.  while this is odd, I don't think it's a big problem because most containers are likely going to show up because of `scale`, so wont have old logs.

Note 2:

If a service is added to the Compose file after `up` or `logs` is run, the logs for that service will **not** show up because we always filter to only the services in the Compose file. We could easily change this, but if we do, I think we should do it consistently for all commands, not just in this case.

Note 3:

Any container started after `project.up()` returns the list of containers, but before the producer thread starts tailing events, will not have it's logs printed. This is somewhat unavoidable if we want to continue to use `attach` for containers created by `up`.


compose/cli/log_printer.py
```diff
@@ -71,32 +153,72 @@ def get_log_generator(container):
...
+
+
+def watch_events(thread_map, event_stream, presenters, thread_args):
+    for event in event_stream:
+        if event['action'] != 'start':
+            continue
+
+        if event['id'] in thread_map:
+            if thread_map[event['id']].is_alive():
+                continue
```
dnephin: I think there might be a race condition here, where a container could stop and start again immediately, but the tailer thread is still printing logs, so the thread is still alive.

I used to have a branch of `action == 'stop'`, which would pop the thread, I think I need to add that back to handle this case.

compose/cli/main.py
```diff
@@ -720,6 +724,9 @@ def up(self, project, options):
             raise UserError("--abort-on-container-exit and -d cannot be combined.")
 
         with up_shutdown_context(project, service_names, timeout, detached):
+            # start the event stream first so we don't lose any events
+            event_stream = project.events()
```
dnephin: hmm, I think this is a lie. The generator isn't going to start running until we actually try to read one event from it.  This might need more work.

dnephin: Or maybe this case isn't a big concern, because we already attach to existing containers, it's very unlikely that something is going to scale a container before we start the event stream.

compose/cli/log_printer.py
```diff
@@ -2,67 +2,145 @@
...
+        thread_map = build_thread_map(self.containers, thread_args)
+        start_producer_thread((thread_map, self.event_stream, thread_args))
+
+        for line in consume_queue(queue, self.cascade_stop):
+            remove_stopped_threads(thread_map)
+
+            if not line:
+                if not thread_map:
+                    return
+                continue
```
aanand: Wouldn't mind a comment above both the `return` and `continue` statements in this block explaining what condition we've reached. If I have it right, we're continuing (not producing a line of output) if a single container has stopped, and returning (stopping all output) if all containers have stopped?

compose/cli/log_printer.py
```diff
@@ -2,67 +2,145 @@
...
 
-        for color_func, container in zip(color_funcs, self.containers):
-            generator_func = get_log_generator(container)
-            prefix = color_func(build_log_prefix(container, prefix_width))
-            yield generator_func(container, prefix, color_func, self.log_args)
 
+def build_thread(container, presenters, queue, log_args):
+    tailer = Thread(
+        target=tail_container_logs,
+        args=(container, next(presenters), queue, log_args))
```
aanand: Feels a bit odd that `build_thread` takes a single container as an argument but the other sequence of things that we're advancing through in step (i.e. `presenters`) is passed in as an iterator to be advanced.

Might be cleaner to achieve the same thing by using a `zip` somewhere higher up the call stack and passing a single container + presenter to this function?

aanand:
LGTM, just a couple small code comments

compose/cli/log_printer.py
dnephin: Correct on the return. The continue is for any time we hit the timeout on `queue.get()`. 

A comment sounds appropriate, will add.

compose/cli/log_printer.py
dnephin: I think I had it that way for earlier iterations, but it meant that I had to deal with the arguments separately (instead of just being able to group presenters in with the other thread_args).

I'll take another look, maybe the caller is a better place to do this.

dnephin: Looking at this again, the reason container is singular is because from `watch_events()` we only have a single container. Using `presenters` instead of a single presenter lets us pass in all the thread args (except for the one unique arg of `container`) as a blob.  I think I only changed this in the last commit, so I can change it back if you want.

dnephin:
Rebased and added those comments.

compose/cli/log_printer.py
aanand: Yeah, I think I'd find it easier to understand if it was changed back. Thanks.

aanand:
LGTM

PR #3095, opened 2016-03-08
Refactor command dispatch and fix api version mismatch error

### LINKED ISSUE #3015: exceptions while handling of exceptions
it is difficult to trace/debug what is wrong with one service not willing to be brought up because during handling of exceptions other exceptions occured. The latter one can be traced to the first line
of[ `log_api_error()`](https://github.com/docker/compose/blob/master/compose/cli/main.py#L89)
there `client is newer than server` (str) is not of the same type as `e.explanation` (bytes). 

I am not sure where the bytestring come from but probably from the failed http request results which IIRC are bytestreams. At least on 3.5.1 (what I am running docker-compose in) you need to do::

```
def log_api_error(e):
    expl = e.explanation
    try:
        if isinstance(e.explanation, bytes):
            expl = e.explanation.decode('utf-8')
    except:
        pass
    if 'client is newer than server' in expl:

``` 
After changing that, I was able to see and address the real problem

If you are running under Python 2.7 you will not see this problem and utf-8 encoding should not be necessary. In any case a failed decoding leads to the `in` test to be as good as the original

dnephin:
Fixes #3015

The first few commits are refactoring of the command dispatch. Instead of a subclass/superclass there are now two separate classes. The first is concerned with:

* parsing docstrings
* displaying the top level help
* finding the appropriate handler function

The second is the `TopLevelCommand` which has all the handler functions for each subcommand. It now accepts a `Project` object to the constructor, instead of to each handler function. This would allow us to unit test the commands by passing in a mock of the project.

Finally I consolidated the connection error handling into one place. It was split between `main()`, and `friendly_error_message()`. It's now all in the `errors.py` module, and only more general errors are left in `main()`.

### DIFF
--- a/compose/cli/command.py
+++ b/compose/cli/command.py
@@ -1,52 +1,25 @@
 from __future__ import absolute_import
 from __future__ import unicode_literals
 
-import contextlib
 import logging
 import os
 import re
 
 import six
-from requests.exceptions import ConnectionError
-from requests.exceptions import SSLError
 
-from . import errors
 from . import verbose_proxy
 from .. import config
 from ..const import API_VERSIONS
 from ..project import Project
 from .docker_client import docker_client
-from .utils import call_silently
 from .utils import get_version_info
-from .utils import is_mac
-from .utils import is_ubuntu
 
 log = logging.getLogger(__name__)
 
 
-@contextlib.contextmanager
-def friendly_error_message():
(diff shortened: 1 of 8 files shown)


shin-:
LGTM.

aanand:
LGTM

Issue #1818, opened 2015-08-06
Epic: Distributable Compose apps


aanand:
(This is closely related to #1784.)

The Docker image format, along with the Hub, registry and Dockerfile build system, constitute a really good way to distribute single-container software. There's no equivalent for multi-container applications. The closest you can get is a Compose file that references a bunch of images hosted elsewhere.

If there were a way to package up a whole multi-container app - its container's images and configuration - for one-step distribution and deployment (as simple as `docker push`, `docker pull` and `docker run` make distributing single-container software), it would enable much smoother distribution of much more complex apps than a Docker registry can currently supply you with.

This might entail designing an artifact that's generated from a Compose file, much as a container image is generated from a Dockerfile.

There are also interesting use cases regarding deployment and rollback - if you could serialise/deserialise between the state of an app and an immutable artifact, rolling back a deployment would become a single logical step.

dnephin:
I've been thinking about this a bit, as it ties into #318. If the entire app can be pulled in that makes "including" it in another app much easier.

I was thinking along the lines of keeping the compose file and the images separate. That way you can re-use the docker-registry to pull the images, all you need is a way to distribute the `docker-compose.yml`. Keeping the images separate means you can pull things concurrently (instead of trying to pull a single large blob).

Sharing the `docker-compose.yml` is pretty trivial since it's a small text file, all we're missing is a registry for them.  If we had that, we could easily implement something like:

`docker-compose up --url http://compose-registry.docker.com/username/myapp.yml`

which would `GET` the `myapp.yml`, and proceed with a normal `up`.

Publishing the app could be something like `docker-compose publish`, which runs `docker push` for all the images, and a `POST` for the docker-compose.yml to the url.

thaJeztah:
A good start could be to support GitHub, similar to what docker supports with `docker build`; https://github.com/docker/docker/blob/master/docs/reference/commandline/build.md

E.g. `docker-compose (https://)github.com/me/myproject` would clone the project in the current directory, and build/start the project

dnephin:
I might be good to break down this problem into a few components:
* publish - how files are sent to a registry
* registry - how the files are stored/served
* consume - how to use a published file to start an app

I think a solution to each of these is should be pretty straightforward:

* publish - `docker-compose publish <url>` a fairly simple http POST
* registry - either docker hub, or some really simple CRUD add to receive and serve files
* consume - `docker-compose up --url <url>` (or something similar) http GET and proceed as usual

[...]

PR #3098, opened 2016-03-09
Support `-f url` to specify config url

yongjhih:
For example:

``` sh
docker-compose -f https://github.com/yongjhih/docker-parse-server/raw/master/docker-compose.yml up
```

### DIFF
--- a/compose/config/config.py
+++ b/compose/config/config.py
@@ -9,6 +9,7 @@ import string
 import sys
 from collections import namedtuple
 
+import requests
 import six
 import yaml
 from cached_property import cached_property
@@ -102,6 +103,11 @@ DOCKER_VALID_URL_PREFIXES = (
     'git://',
     'github.com/',
     'git@',
+)
+
+COMPOSE_VALID_URL_PREFIXES = (
+    'http://',
+    'https://',
 )
 
 SUPPORTED_FILENAMES = [
@@ -226,6 +232,13 @@ def find(base_dir, filenames, environment):
         )
 
     if filenames:
+        if is_compose_url(filenames[0]):
+            response = requests.get(filenames[0])
+            response.raise_for_status()
+            return ConfigDetails(
+                os.getcwd(),
+                [ConfigFile(None, yaml.safe_load(response.text))])
+
         filenames = [os.path.join(base_dir, f) for f in filenames]
     else:
         filenames = get_default_config_files(base_dir)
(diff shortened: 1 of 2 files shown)


compose/config/config.py
```diff
@@ -213,6 +214,9 @@ def find(base_dir, filenames):
             os.getcwd(),
             [ConfigFile(None, yaml.safe_load(sys.stdin))])
 
+    if filename.startswith('./http'):
```
sergeyklay: How about https://en.wikipedia.org/wiki/Lists_of_network_protocols

yongjhih: How about protocols testing for urllib?
I think we could support usual protocols.

changyuheng: I think `./http` is kind of hack. But @sergeyklay were you talking about supporting all the protocols? Or you're saying you prefer to support stdin and local file system only?

sergeyklay: I mean support of usual protocols, like ftp, http, etc.

dnephin:
Related to https://github.com/docker/compose/issues/1818#issuecomment-128416833

I like the idea, but I don't think we should overload `-f`. I think it would be better to make a new flag `--url` because downloading a remote file can fail in many more ways than just reading a local file.

compose/config/config.py
```diff
@@ -984,8 +994,11 @@ def has_uppercase(name):
 
 def load_yaml(filename):
     try:
-        with open(filename, 'r') as fh:
-            return yaml.safe_load(fh)
+        if is_compose_url(filename):
+            return yaml.safe_load(urllib.urlopen(filename))
```
dnephin: This is going to need a lot more error handling. There are a lot of failures that can happen here.

sergeyklay:
+1 for `--url`

compose/config/config.py
changyuheng: Actually, reference to the build system of Go or Java, I think we can simply support HTTP/HTTPS. Because they're much more popular them other protocols. curl http://stuff/to/download | docker-compose -f - is just a little bit long. And you know what, the container version of docker-compose cannot even read stdin properly. I think that's a bug but this can be a reason for us to consider the improvement.

pimterry:
I'd really like to use this, but it looks like the builds are broken right now, and just by minor things: a too long line in the tests and incorrectly ordered imports. @yongjhih, any chance you could fix those up soon?

compose/config/config.py
yongjhih: Allow error handling by `response.raise_for_status()`

yongjhih:
I used to `vim http://xxx` not `vim --url http://xxx`, so i did not think about `--url` option.

denismakogon:
+1 to previous reviews, --url should be the best option here, so in general, it looks good, but please consider to use different flag.

yongjhih:
@pimterry 
Fixed

PR #3102, opened 2016-03-10
Add flag to up/down to remove orphaned containers

### LINKED ISSUE #1477: Stop containers that have been removed from docker-compose.yml

shin-:
Fixes #1477

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -252,13 +252,15 @@ class TopLevelCommand(object):
         Usage: down [options]
 
         Options:
-            --rmi type      Remove images, type may be one of: 'all' to remove
-                            all images, or 'local' to remove only images that
-                            don't have an custom name set by the `image` field
-            -v, --volumes   Remove data volumes
+            --rmi type          Remove images, type may be one of: 'all' to remove
+                                all images, or 'local' to remove only images that
+                                don't have an custom name set by the `image` field
+            -v, --volumes       Remove data volumes
+            --remove-orphans    Remove containers for services not defined in
+                                the Compose file
         """
         image_type = image_type_from_opt('--rmi', options['--rmi'])
-        self.project.down(image_type, options['--volumes'])
+        self.project.down(image_type, options['--volumes'], options['--remove-orphans'])
 
     def events(self, options):
         """
@@ -324,9 +326,9 @@ class TopLevelCommand(object):
         signals.set_signal_handler_to_shutdown()
         try:
             operation = ExecOperation(
-                    self.project.client,
-                    exec_id,
-                    interactive=tty,
+                self.project.client,
+                exec_id,
+                interactive=tty,
             )
             pty = PseudoTerminal(self.project.client, operation)
             pty.start()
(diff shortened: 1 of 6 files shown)


compose/cli/main.py
```diff
@@ -269,13 +269,14 @@ def down(self, project, options):
...
         Options:
-            --rmi type      Remove images, type may be one of: 'all' to remove
-                            all images, or 'local' to remove only images that
-                            don't have an custom name set by the `image` field
-            -v, --volumes   Remove data volumes
+            --rmi type          Remove images, type may be one of: 'all' to remove
+                                all images, or 'local' to remove only images that
+                                don't have an custom name set by the `image` field
+            -v, --volumes       Remove data volumes
+            --remove-orphans    Remove orphaned containers
```
dnephin: I was thinking this could we `--remove-orphaned`, what do you think?

compose/project.py
```diff
@@ -356,23 +362,46 @@ def pull(self, service_names=None, ignore_pull_failures=False):
...
+                    ctnr.remove(force=True)
+                else:
+                    log.warning(
+                        'Found orphan container "{0}" for service "{1}". If '
+                        'you removed or renamed this service in your compose '
+                        'file, you can run this command with the '
+                        '--remove-orphans flag to clean it up.'.format(
+                            ctnr.name, service_name
+                        )
+                    )
```
dnephin: I wonder if we should also say: "If this container is part of a different Compose project you can set a unique project name using `-p` or COMPOSE_PROJECT_NAME".

shin-: The current message is already 3 lines long - I'm a bit worried about having too much information here.

compose/cli/main.py
shin-: I think `--remove-orphans` sounds better and it's one letter shorter. Grammatically I prefer it as well (i.e. "Remove orphaned what?")

I'll update the help text.

compose/project.py
aanand: Speaking of verbosity, it looks like this will be displayed multiple times if there are multiple orphaned containers/services? It'd be better to show just one message which lists all of them.

aanand:
I like it, just one comment on the warning message

shin-:
Updated to limit the number of warnings if multiple orphans are found.

compose/project.py
```diff
@@ -356,23 +362,50 @@ def pull(self, service_names=None, ignore_pull_failures=False):
...
+                ctnr.remove(force=True)
+        else:
+            log.warning(
+                'Found orphan containers ({0}) for this project. If '
+                'you removed or renamed this service in your compose '
+                'file, you can run this command with the '
+                '--remove-orphans flag to clean it up.'.format(
+                    ', '.join(["{}".format(ctnr.name) for ctnr in _find()])
+                )
+            )
```
dnephin: Isn't this being logged on every call now? 

Shouldn't there be something like `if not orphans: return` at the top?

shin-: Fixed + rebased

tests/integration/project_test.py
```diff
@@ -1055,3 +1057,47 @@ def test_project_up_named_volumes_in_binds(self):
...
+                'service1': {
+                    'image': 'busybox:latest',
+                    'command': 'top',
+                }
+            }
+        }
+
+        base_file = config.ConfigFile('base.yml', config_dict)
+        config_details = config.ConfigDetails('.', [base_file])
+        config_data = config.load(config_details)
```
dnephin: I think `build_config()` does this?

dnephin:
LGTM

shin-:
Rebased / squashed.

shin-:
I just updated the integration test to use v1 format so it can be tested against engine 1.9.1 as well.

Issue #2441, opened 2015-11-21
Improve variable interpolation, required flag and default value.


aramalipoor:
As first proposed in #1377 I think it would be very useful to have a way to set a variable as required, and  also set some default value.

```yaml
web:
   image: org/app:${VERSION?} # Required
```
Or
```yaml
web:
   image: org/app:${VERSION:latest} # Default Value
```

majidgolshadi:
:+1:

dustinblackman:
:+1:

armab:
:+1:

valeriangalliat:
:+1:

mseashor:
:+1:

shane-axiom:
+1, defaults for environment variables would be great. Simple syntax this would be great:

    mycontainer:
      image: ubuntu:latest
      ports:
        - ${THEPORT:8080}:8080

bennycornelissen:
:+1:

plouc:
I'm working on a project where people have heterogeneous env, I'm working on OSX and some are using Ubuntu, when using docker-machine with virtualbox provider, I have to put the full path to make Volume works:

```yaml
myservice:
  volumes:
    - ${HOST_VOLUME_PATH}/src:/src
```

The problem is now people who don't use virtualbox have to define the value, it should be great if we could  have a default value:

```yaml
myservice:
  volumes:
    - ${HOST_VOLUME_PATH:.}/src:/src
```

In other words :+1:

obazoud:
:+1:

jlrigau:
:+1:

scasagrande-d2l:
+1

leogamas:
:+1:

roll:
:+1:

asilluron:
:+1:

digitalLumberjack:
:+1:

hadim:
This feature is a must have. Any devs from the docker-compsose team working on it ? If not I could propose a PR.

dnephin:
While having defaults would be really nice, we haven't decided on any implementations yet.

I (personally) think it's more likely we go with something like #2636 for default values. It supports more use-cases, and doesn't require us to invent new syntax for defaults.

marcellodesales:
I'd like to see the same implementation as https://docs.docker.com/engine/reference/builder/#environment-replacement Right now, I'd like the ability to switch from 2 different implementations for the log driver.

Given the following Base implementation:

```yml
intuitive-syslog:
[...]

PR #3108, opened 2016-03-10
Support inline default values for interpolation

dnephin:
Implements part of #2441 to support inline default values.

I have no idea how robust this is, but it's a start.


compose/config/interpolation.py
```diff
@@ -53,9 +53,33 @@ def recursive_interpolate(obj, mapping):
...
+    # Modified from python2.7/string.py
+    def substitute(self, mapping):
+        # Helper function for .sub()
+        def convert(mo):
+            # Check the most common path first.
+            named = mo.group('named') or mo.group('braced')
+            if named is not None:
+                if ':-' in named:
+                    var, _, default = named.partition(':-')
+                    return mapping.get(var, default)
```
dnephin: These 3 lines are the only real substantial modifications.

I also cut out the arg/kwarg handling and just replaced it with a mapping arg, since that's how we use it.

aanand:
Nice. Maybe we should support `${VAR-default}` as well, i.e. substitute the default if VAR is unset, but not if it's set to an empty string? That way you can override an variable to the empty string if you want to.

dreyks:
oh, this is just what all of us need =) any hope this will be merged soon?

marcellodesales:
@aanand, @dnephin What needs to be done to get this in? Have been waiting for this for a while

nicooga:
What about using environment variables in key names? Like in:

~~~yml
version: '2'

services:
  postgres: {image: 'postgres:9.6'}
  redis: {image: 'redis:3.2.3'}

  web_server:
    build: ../sources/myapp
    depends_on: [postgres]
    env_file: ../env/$ENV.env
    links: [postgres]
    volumes:
      - myapp_${ENV}_source:/source
      - myapp_nginx_socks:/tmp/socks

  sidekiq:
    build: ../sources/myapp
    command: bundle exec sidekiq -e $ENV -c 2 -q default -q carrierwave
    depends_on: [postgres, redis]
    env_file: ../env/$ENV.env
    links: [postgres, redis]
    volumes_from: [web_server]

volumes:
  myapp_${ENV}_db: {external: true}
  myapp_${ENV}_source: {external: true}
  myapp_nginx_socks: {external: true}
~~~
For values it seems to be working right.

dnephin:
@marcellodesales I think we're going to try and get this into the next release for the v2.1 file format. 

#3653 adds the v2.1 file format, so once that is merged I can rebase this and change it to only apply to the new format.

marcellodesales:
@dnephin Sounds good... I can definitely be up to QA this... I reallllllly need it!

aanand:
I'd still like to support `${VAR-default}` syntax, if it's easy to implement.

dnephin:
Ok, rebased, added the `${VAR-default}` and gated the feature behind the Compose V2.1 version.

hadim:
Nice addition guys !

I am testing it at the moment the master git code and docker-engine 1.12 and it seems to work great except for volumes defaut values : 

```
version: '2.1'
services:

  test:
    container_name: test-name
    image: busybox
    command: "true"
    privileged: true
    volumes:
      - ${DATA_DIR-/home/hadim/test_data_dir}:/data:Z
```

I get the following error : 

```
ERROR: Invalid interpolation format for "volumes" option in service "test": "${DATA_DIR-/home/hadim/test_data_dir}:/data:Z"
```

Am I missing something or is that a bug ?

Thanks

shin-:
@hadim Looks like a bug - do you mind creating a separate issue for it? Thank you for your help!

hadim:
Sure.

glensc:
@dnephin https://github.com/docker/compose/pull/3108#issuecomment-248038373

i proposed new syntax for 2.1 here:
https://github.com/docker/compose/issues/3998#issuecomment-255069785

that shouldn't add complexity as keys can be still not interpolated and makes syntax more similar to other definitions.

AlJohri:
@shin- With this config: `ports: ["${PORT-8000}:${PORT-8000}"]`

I'm getting this error: `ERROR: Invalid interpolation format for "ports" option in service "app": "${PORT-8000}:${PORT-8000}"`

Am I doing something wrong?

shin-:
@AlJohri please share the output of `docker-compose version` and your entire Compose file.

tmack8001:
I also get this same issue where defaults aren't supported... and my docker-compose version is the latest non release candidate, 1.22.0. I'm trying to add an environment variable and default into the "docker_host" of the image.

```
my-cool-container
    image: ${DOCKER_HOST:-docker.something.internal.com}/namespace/my-cool-image:my-cool-tag
```

but I get the following error while executing `docker-compose up`:
```
ERROR: Invalid interpolation format for "image" option in service "my-cool-container": "${DOCKER_HOST:-docker.something.internal.com}/namespace/my-cool-image:my-cool-tag"
```

shin-:
>  Additionally when using the **2.1 file format**, it is possible to provide inline default values using typical shell syntax

https://docs.docker.com/compose/compose-file/compose-file-v2/#variable-substitution

Default values only with v2.1 or above.

tmack8001:
ah thanks @shin- 👍 I was still referring to v2 🤕

PR #3118, opened 2016-03-13
Implement #745: Ability to set default project name in YML

fesor:
Added `project` section in docker compose file, which allows users to define **default** project name in their yaml files:

```
version: "2"
project:
    default_name: "example"
```

This will help teams not to forgot about that. For example in my company i have only 5 project, which uses docker compose, and I already experience some problems since developers sometime forgot to add custom project name.

### Name resolution rules

 - Use custom name from `-p` cli option if provided
 - Use custom name from `COMPOSE_NAME` env variable if available
 - Use `default_name` option from `project` section of yml fileif provided
 - Use basename of project directory  as project default name

### Multiple compose files

If multiple compose yml files provided, `project` section will be used only from main one. 

docker-compose.yml:
```
version: "2"
project:
    default_name: foo
```

docker-compose.override.yml:
```
version: 2
project:
    default_name: bar
```

If we start docker-compose as usual (i.e. `docker-compose.override.yml` will override service definition in main file), then project name will be `foo`.

But if we change the order of files:

```
docker-compose -f docker-compose.override.yml -f docker-compose.yml config
```

then `bar` will be used as project name.

### DIFF
--- a/compose/cli/command.py
+++ b/compose/cli/command.py
@@ -55,8 +55,9 @@ def get_client(verbose=False, version=None, tls_config=None, host=None):
 def get_project(project_dir, config_path=None, project_name=None, verbose=False,
                 host=None, tls_config=None):
     config_details = config.find(project_dir, config_path)
-    project_name = get_project_name(config_details.working_dir, project_name)
     config_data = config.load(config_details)
+    default_project_name = config_data.project.get('default_name', None)
+    project_name = get_project_name(config_details.working_dir, project_name, default_project_name)
 
     api_version = os.environ.get(
         'COMPOSE_API_VERSION',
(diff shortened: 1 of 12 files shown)


sergeyklay:
Why twice :)

dnephin:
We discussed this a bit recently. Any `project_name` set in the Compose file should be just a default name, so it looks like the precendence order is correct in this PR. However, it's not clear from the name that it's only a "default", and not actually an override.

We weren't really able to come up with a good way to express this. Some ideas that have been suggested previously are:
* have a section called `project:` and make the key `default_name:` or `default_project_name:`
* make a top level key `default_project_name:`

I think it's going to be important to make it clear that this name is just a default.

fesor:
@dnephin good point.

I think `project` section gives more space for future changes. In this case I more like `default_name` option.

```
version: "2"
project:
    default_name: "app"
```

What do you think?

dnephin:
I personally like that option

ping @aanand , @shin-

fesor:
Closing this in flavor of https://github.com/docker/compose/pull/3057

adrianolek:
So was the idea of project name in yml abandoned? #3057 only allows to set project name via .env file.

fesor:
@adrianolek well... my implementation doesn't allow to override default name from another `docker-compose.yml` which makes it just pointless, you could just specify project name in `.env` and that's do the trick.

Also all other cases that I looked to easily solves via env variables (personally I just specify all this in CI build/deploy job as ENV variables and it works for me). Allowing to override default project name using multiple compose file just creates a lot of edge cases and this should be discussed.

PR #3122, opened 2016-03-14
Adding optional selection of SSL Version

KalleDK:
To allow users to be able to select another ssl version

(Windows boot2docker apparantly loves PROTOCOL_TLSv1)

This way you can change the version via env DOCKER_SSL_VERSION

Please comment, haven't done much

### DIFF
--- a/compose/cli/command.py
+++ b/compose/cli/command.py
@@ -4,6 +4,7 @@ from __future__ import unicode_literals
 import logging
 import os
 import re
+import ssl
 
 import six
 
@@ -37,8 +38,8 @@ def get_config_path_from_options(options):
     return None
 
 
-def get_client(verbose=False, version=None):
-    client = docker_client(version=version)
+def get_client(verbose=False, version=None, tls_version=None):
+    client = docker_client(version=version, tls_version=tls_version)
     if verbose:
         version_info = six.iteritems(client.version())
         log.info(get_version_info('full'))
@@ -57,7 +58,15 @@ def get_project(project_dir, config_path=None, project_name=None, verbose=False)
     api_version = os.environ.get(
         'COMPOSE_API_VERSION',
         API_VERSIONS[config_data.version])
-    client = get_client(verbose=verbose, version=api_version)
+    compose_tls_version = os.environ.get(
+        'COMPOSE_TLS_VERSION',
+        None)
+
+    tls_version = None
+    if compose_tls_version:
+        tls_version = ssl.getattr("PROTOCOL_{}".format(compose_tls_version))
+
+    client = get_client(verbose=verbose, version=api_version, tls_version=tls_version)
 
     return Project.from_config(project_name, config_data, client)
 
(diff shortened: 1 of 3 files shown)


dnephin:
Is `DOCKER_SSL_VERSION` something that is supported by the docker cli? I don't see any reference to it. If it's a Compose only flag it should be `COMPOSE_SSL_VERSION`.

Also needs docs in `docs/reference/envvars.md`, but lets see what @shin- and @aanand think about it first.

compose/cli/command.py
```diff
@@ -84,7 +85,15 @@ def get_project(base_dir, config_path=None, project_name=None, verbose=False):
     api_version = os.environ.get(
         'COMPOSE_API_VERSION',
         API_VERSIONS[config_data.version])
-    client = get_client(verbose=verbose, version=api_version)
+    docker_ssl_version = os.environ.get(
+        'DOCKER_SSL_VERSION',
+        None)
+
+    ssl_version = None
+    if docker_ssl_version == "TLSv1":
```
shin-: We need to add cases for TLSv1_1 and TLSv1_2 as well.

shin-:
It should be `COMPOSE_TLS_VERSION`, yes.
I like having in option, it's in line with the recent change https://github.com/docker/docker-py/pull/971

compose/cli/command.py
KalleDK: Yes, I didn't know which was supported by docker except for TLSv1

KalleDK:
@dnephin i found it in docker-py, don't know if thats compose specific

I'm using it from python on a windows connecting to boot2docker

KalleDK:
Changed to COMPOSE_TLS_VERSION and added docs

The default value is missing, as I don't know if we should set it, or use the one from docker-py

KalleDK:
Can't tell why the documentation test fail?

compose/cli/command.py
```diff
@@ -57,7 +58,19 @@ def get_project(project_dir, config_path=None, project_name=None, verbose=False)
...
-    client = get_client(verbose=verbose, version=api_version)
+    compose_tls_version = os.environ.get(
+        'COMPOSE_TLS_VERSION',
+        None)
+
+    tls_version = None
+    if compose_tls_version == "TLSv1":
+        tls_version = ssl.PROTOCOL_TLSv1
+    if compose_tls_version == "TLSv1_1":
+        tls_version = ssl.PROTOCOL_TLSv1_1
```
shin-: This will break on some clients where OpenSSL doesn't support TLS v1.1 / 1.2. You want to use `hasattr()`/`getattr()` instead.

KalleDK: Should it default to none or raise exception?

shin-: IMO It should raise an exception if the user is trying to use an unsupported `tls_version`.

KalleDK: It should raise an AttributeError now

KalleDK:
@shin- Do you need anymore for it to be accepted?

shin-:
I think we want to catch the `AttributeError` and give people a more explicit message when they try to use something that doesn't exist or isn't available. Otherwise it seems fine to me.

PR #3135, opened 2016-03-15
Make down idempotent, continue to remove resources if one is missing

### LINKED ISSUE #3132: docker-compose down shouldn't abort if resources have already been deleted
Running `docker-compose down` a second time should be a no-op, but it complains about the network having already been removed:

```console
$ docker-compose down
Stopping app_web_1 ... done
Stopping app_redis_1 ... done
Removing app_web_1 ... done
Removing app_redis_1 ... done
Removing network app_default

$ docker-compose down
Removing network app_default
ERROR: network app_default not found
```

The same thing happens with volumes if you pass `-v`.

dnephin:
Fixes #3132

### DIFF
--- a/compose/network.py
+++ b/compose/network.py
@@ -149,7 +149,10 @@ class ProjectNetworks(object):
         if not self.use_networking:
             return
         for network in self.networks.values():
-            network.remove()
+            try:
+                network.remove()
+            except NotFound:
+                log.warn("Network %s not found.", network.full_name)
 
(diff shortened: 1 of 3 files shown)


aanand:
Nice. Tiny nitpick: I think "Network" and "Volume" should be capitalised in the warnings.

dnephin:
done

aanand:
LGTM

PR #3139, opened 2016-03-15
Add support for TLS config command-line options

### LINKED ISSUE #1716: Add support for -H and TLS flags
The `docker-compose` command should support these options from the `docker` command:

```
  -H, --host=[]                      Daemon socket(s) to connect to 
  --tls=false                        Use TLS; implied by --tlsverify
  --tlscacert=~/.docker/ca.pem       Trust certs signed only by this CA
  --tlscert=~/.docker/cert.pem       Path to TLS certificate file
  --tlskey=~/.docker/key.pem         Path to TLS key file
  --tlsverify=false                  Use TLS and verify the remote
```

This will allow you to do `docker-compose $(docker-machine config dev) ...` etc.

shin-:
Fixes #1716

### DIFF
--- a/compose/cli/docker_client.py
+++ b/compose/cli/docker_client.py
@@ -22,7 +22,7 @@ def tls_config_from_options(options):
     cert = options.get('--tlscert')
     key = options.get('--tlskey')
     verify = options.get('--tlsverify')
-    hostname = urlparse(options.get('--host', '')).hostname
+    hostname = urlparse(options.get('--host') or '').hostname
 
     advanced_opts = any([ca_cert, cert, key, verify])
 
--- a/compose/cli/docker_client.py
+++ b/compose/cli/docker_client.py
@@ -22,14 +22,17 @@ def tls_config_from_options(options):
     key = options.get('--tlskey')
     verify = options.get('--tlsverify')
 
-    if tls is True:
+    advanced_opts = any([ca_cert, cert, key, verify])
+
+    if tls is True and not advanced_opts:
         return True
-    elif any([ca_cert, cert, key, verify]):
+    elif advanced_opts:
         client_cert = None
         if cert or key:
             client_cert = (cert, key)
         return TLSConfig(
-            client_cert=client_cert, verify=verify, ca_cert=ca_cert
+            client_cert=client_cert, verify=verify, ca_cert=ca_cert,
+            assert_hostname=options.get('--skip-hostname-check')
         )
     else:
         return None
--- a/compose/cli/docker_client.py
+++ b/compose/cli/docker_client.py
@@ -3,9 +3,11 @@ from __future__ import unicode_literals
 
 import logging
 import os
+from collections import namedtuple
 
 from docker import Client
 from docker.errors import TLSParameterError
+from docker.tls import TLSConfig
 from docker.utils import kwargs_from_env
 
 from ..const import HTTP_TIMEOUT
@@ -14,7 +16,24 @@ from .errors import UserError
 log = logging.getLogger(__name__)
 
 
-def docker_client(version=None):
+class TLSArgs(namedtuple('_TLSArgs', 'tls cert key ca_cert verify')):
+    @classmethod
+    def from_options(cls, options):
+        return cls(
+            tls=options.get('--tls', False),
+            ca_cert=options.get('--tlscacert'),
+            cert=options.get('--tlscert'),
+            key=options.get('--tlskey'),
+            verify=options.get('--tlsverify')
+        )
+
+    # def has_config(self):
+    #     return (
+    #         self.tls or self.ca_cert or self.cert or self.key or self.verify
+    #     )
+
+
+def docker_client(version=None, tls_args=None, host=None):
     """
     Returns a docker-py client configured using environment variables
     according to the same logic as the official Docker client.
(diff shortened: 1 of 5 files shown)


compose/cli/main.py
```diff
@@ -147,6 +147,13 @@ class TopLevelCommand(object):
       -p, --project-name NAME   Specify an alternate project name (default: directory name)
       --verbose                 Show more output
       -v, --version             Print version and exit
+      -H, --host HOST           Daemon socket to connect to
+
+      --tls                     Use TLS; implied by --tlsverify
+      --tlsacert                Trust certs signed only by this CA
```
tombee: Hey, should this be `--tlscacert`? :)

aanand:
Nice. Rather than create an intermediate `TLSArgs` object, how about just having a method `tls_config_from_options` which takes the options dict and returns a `TLSConfig` object? Then `docker_client()` can just take a `tls_config` kwarg.

tombee:
I've been trying this out with just TLS verification using the CA cert, but I can't seem to get it to work (same thing happens with and without `-H`).

```
$ docker-compose -H 192.168.1.16:443 --tlsverify --tlscacert ca.pem up
$ docker-compose --tlsverify --tlscacert ca.pem up

Define and run multi-container applications with Docker.

Usage:
  docker-compose [-f=<arg>...] [options] [COMMAND] [ARGS...]
  docker-compose -h|--help

Options:
  -v, --version             Print version and exit
  -H, --host HOST           Daemon socket to connect to

  --tls                     Use TLS; implied by --tlsverify
  --tlsacert                Trust certs signed only by this CA
  --tlscert                 Path to TLS certificate file
  --tlskey                  Path to TLS key file
  --tlsverify               Use TLS and verify the remote
```

aanand:
@tombee I think that's because of the typo you found - try passing `--tlsacert`.

I wish we could show a better error message when an arg isn't recognised.

tombee:
@aanand Tried that, I think the option is actually `--tlscacert`: https://github.com/docker/compose/pull/3139/files#diff-25307ba665155091ce1d48c75283dd6aR24 :)

Think it's just our help message that had the typo?

aanand:
@tombee We use docopt, so our help message is what the command line arg parser is generated from!

shin-:
Sorry, it's still wip. I'll push an update soon!

shin-:
Should look better now. @tombee if you wanna give it another shot!

compose/cli/docker_client.py
```diff
@@ -14,7 +15,27 @@
...
-def docker_client(version=None):
+def tls_config_from_options(options):
+    tls = options.get('--tls', False)
+    ca_cert = options.get('--tlscacert')
+    cert = options.get('--tlscert')
+    key = options.get('--tlskey')
+    verify = options.get('--tlsverify')
+
+    if tls is True:
+        return True
```
dnephin: Setting `--tls` makes it ignore all the other tls flags? is that how the docker-cli works?

shin-: I thought so, but checking again, it seems another valid option is `docker --tls --tlscert ... --tlskey ...`  which means "Authenticate with client certificate, do not authenticate server based on given CA" (what we would do here if `--tls` was omitted). I'll update to match.

tombee:
If I prefix with `tcp://`, then TLS doesn't seem to get enabled:

```
$ docker-compose -H tcp://192.168.1.16:443 --tlsverify --tlscacert ca.pem up
ERROR: Couldn't connect to Docker daemon at http://192.168.1.16:443 - is it running?
```

tombee:
Thanks for this @shin-!  Making a bit of progress, but still unable to use our CA cert for verification.

```
$ docker-compose -H 192.168.1.16:443 --tlsverify --tlscacert ca.pem up
ERROR: SSL error: hostname '192.168.1.16' doesn't match 'debian'
```

I've verified that I can use the same configuration with the engine CLI:

```
$ docker -H 192.168.1.16:443 --tlsverify --tlscacert ca.pem info
Containers: 23
Images: 24
Role: primary
Strategy: spread
Filters: health, port, dependency, affinity, constraint
Nodes: 1
 debian: 192.168.1.16:12376
...
```

Also seeing that I get a slightly different error if `--tls` is supplied even though our docs say `--tls=false - Use TLS; implied by --tlsverify`, so I was expecting the behaviour to be the same :)

```
$ docker-compose -H 192.168.1.16:443 --tls --tlsverify --tlscacert ca.pem up
ERROR: SSL error: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:581)
```

shin-:
- The `tcp://` address not being translated to an `https` endpoint is a docker-py bug.
- This `ERROR: SSL error: hostname '192.168.1.16' doesn't match 'debian'` is a case of `assert_hostname` being enabled by default. Don't know if we should disable it by default (docker engine seems to ignore it?) or add an option to do so.
- The issue with the `--tls` flag is linked to what Daniel mentions here https://github.com/docker/compose/pull/3139#discussion_r56447620 .

Thanks for your help with testing!

aanand:
We should definitely not check the hostname by default, as that'll break all Docker Machine local VMs. We could put an option in to enable hostname checking, but if the CLI doesn't do it, I think it's pretty low-priority.

shin-:
I left the option to disable the hostname check entirely, and modified the code a little bit to pass the user-specified hostname as the `assert_hostname`. It's not well documented but it does seem to be a feature of urllib3 / requests. This should work as expected, but if someone could test it with docker-machine that would be super helpful.

dnephin:
Looks like there is a python2.7 issue

dnephin:
LGTM, the one test flake is fixed in master now, if you want to rebase

tests/unit/cli_test.py
```diff
@@ -20,6 +21,85 @@
 from compose.project import Project
 
 
+class TLSConfigTestCase(unittest.TestCase):
```
dnephin: hmm, I think this should  be in `tests/unit/cli/docker_client_test.py`

dnephin:
LGTM

tombee:
@shin- Awesome work on this btw!

PR #3141, opened 2016-03-16
Add ability to remove one-off containers

shin-:
`rm` command now accepts a `--all` argument (shorthand `-a`) to remove one-off containers as well.

Going forward, the plan is to have this always be the case, hence a warning should be added for users when using `rm`

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -491,8 +491,12 @@ class TopLevelCommand(object):
         Options:
             -f, --force   Don't ask to confirm removal
             -v            Remove volumes associated with containers
+            -a, --all     Also remove one-off containers
         """
-        all_containers = self.project.containers(service_names=options['SERVICE'], stopped=True)
+        all_containers = self.project.containers(
+            service_names=options['SERVICE'], stopped=True,
+            one_off=(None if options.get('--all') else False)
+        )
         stopped_containers = [c for c in all_containers if not c.is_running]
 
         if len(stopped_containers) > 0:
@@ -501,7 +505,8 @@ class TopLevelCommand(object):
                     or yesno("Are you sure? [yN] ", default=False):
                 self.project.remove_stopped(
                     service_names=options['SERVICE'],
-                    v=options.get('-v', False)
+                    v=options.get('-v', False),
+                    one_off=options.get('--all')
                 )
         else:
             print("No stopped containers")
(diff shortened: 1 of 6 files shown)


compose/cli/main.py
```diff
@@ -486,8 +486,12 @@ def rm(self, options):
         Options:
             -f, --force   Don't ask to confirm removal
             -v            Remove volumes associated with containers
+            -a, --all     Also remove one-off containers
```
dnephin: I wonder if users know what "one-off" containers are, or if we should add "one-off containers created by docker-compose run".

compose/cli/main.py
```diff
@@ -496,7 +500,8 @@ def rm(self, options):
                     or yesno("Are you sure? [yN] ", default=False):
                 self.project.remove_stopped(
                     service_names=options['SERVICE'],
-                    v=options.get('-v', False)
+                    v=options.get('-v', False),
+                    one_off=options.get('--all')
```
dnephin: There's a bit of a mismatch here, where you do the `None if all else False` in the cli layer for one call, but it's done in the project layer for the other. I think this should be consistent.

I'm usually pretty weary of the True/False/None pattern. I think it's pretty non-intuitive.  Could we use another enum here (we have a bunch already in `compose/service.py` using enum34)?

The values could be: `include` (currently None), `exclude` (currently False), `only` ( currently True).

tests/acceptance/cli_test.py
```diff
@@ -1108,6 +1108,28 @@ def test_rm(self):
...
+        self.assertEqual(len(service.containers(stopped=True, one_off=True)), 0)
+
+        service.create_container(one_off=False)
+        service.create_container(one_off=True)
+        kill_service(service)
+        self.assertEqual(len(service.containers(stopped=True)), 1)
+        self.assertEqual(len(service.containers(stopped=True, one_off=True)), 1)
+        self.dispatch(['rm', '-f', '--all'], None)
+        self.assertEqual(len(service.containers(stopped=True)), 0)
+        self.assertEqual(len(service.containers(stopped=True, one_off=True)), 0)
```
dnephin: I think this can be two separate test cases?

compose/project.py
```diff
@@ -35,6 +36,20 @@
...
+class OneOffFilter(enum.Enum):
+    include = 0
+    exclude = 1
+    only = 2
+
+    @classmethod
+    def update_labels(cls, value, labels):
+        if value == cls.only:
+            labels.append('{0}={1}'.format(LABEL_ONE_OFF, "True"))
+        elif value == cls.exclude or value is False:
```
dnephin: Why `value is False` here? I think it might be safer to change the existing calls.

dnephin:
Replaced by #3178

PR #3156, opened 2016-03-17
Fix TypeError in Exception handling

graingert:
  File "/tmp/tmp.02tgGaAGtW/docker-compose/bin/docker-compose", line 11, in <module>
    sys.exit(main())
  File "/tmp/tmp.02tgGaAGtW/docker-compose/lib/python3.4/site-packages/compose/cli/main.py", line 68, in main
    log_api_error(e)
  File "/tmp/tmp.02tgGaAGtW/docker-compose/lib/python3.4/site-packages/compose/cli/main.py", line 89, in log_api_error
    if 'client is newer than server' in e.explanation:
TypeError: 'str' does not support the buffer interface

### DIFF
--- a/compose/cli/errors.py
+++ b/compose/cli/errors.py
@@ -66,7 +66,7 @@ def handle_connection_errors(client):
 
 
 def log_api_error(e, client_version):
-    if 'client is newer than server' not in e.explanation:
+    if b'client is newer than server' not in e.explanation:
         log.error(e.explanation)
         return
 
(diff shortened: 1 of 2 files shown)


aanand:
Thanks. Will this be compatible across Python 2 and 3? I'm not sure if `e.explanation` will be binary on both platforms.

graingert:
@aanand e.explanation will be bytes or None on both platforms.

aanand:
OK, LGTM

graingert:
looks like the tests were testing with unicode.

shin-:
LGTM!

dnephin:
LGTM

dnephin:
Thanks for the fix!

PR #3164, opened 2016-03-18
WIP: Fix output of 'config' for v1 files

aanand:
When run on a v1 file, `docker-compose config` returns invalid YAML:

```yaml
version: '1'
networks: {}
volumes: {}
services:
  foo:
    network_mode: "service:bar"
```

TODO:

- [x] Output a version 2 file
- [x] Set `network_mode: bridge`
- [x] Denormalise `restart`
- [x] Check for anything else that needs denormalising

### DIFF
--- a/compose/config/serialize.py
+++ b/compose/config/serialize.py
@@ -19,12 +19,14 @@ yaml.SafeDumper.add_representer(types.VolumeSpec, serialize_config_type)
 
 
 def serialize_config(config):
-    services = {service.pop('name'): service for service in config.services}
-
-    if config.version == V1:
-        for service_dict in services.values():
-            if 'network_mode' not in service_dict:
-                service_dict['network_mode'] = 'bridge'
+    denormalized_services = [
+        denormalize_service_dict(service_dict, config.version)
+        for service_dict in config.services
+    ]
+    services = {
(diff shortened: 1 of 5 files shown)


dnephin:
It's not invalid YAML, it's an invalid Compose file, right? The yaml still parses correctly.

I would really rather not implement all the logic to go from the normalized v2 format back into a V1 format, that we should be removing sooner rather than later.

Instead of converting to V1, why don't we just output a valid V2 format?

aanand:
Sorry, yes, it's not invalid YAML.

We could try and output a valid v2 file, but there'll be unpleasant edge cases. For example, we'll have to make sure every service has `network_mode: bridge` unless otherwise specified, to stop it from attempting to use networks.

Alternative plan: disable the `config` command for v1 files.

dnephin:
I prefer either of those options over converting to the V1 format.

Setting `network_mode: bridge` in the config module is actually a really interesting idea. I think it would allow us to remove the `if version` check from `compose/project.py`, which is another bonus.

I don't see us checking for version anywhere else, so that might be the only thing we need to change.

shin-:
#3243

shin-:
LGTM!

Issue #3106, opened 2016-03-10
`docker-compose up` hangs when tty parameter is true


rgorodischer:
Here's a minimal setup to reproduce the issue:

Dockerfile content:

```
FROM debian:jessie

CMD ["/bin/bash"]
```

docker-compose.yml content:

```
test-system:
  build: .
  stdin_open: true
  tty: true
```

Here's what happens when I try docker-compose up:

```
$ docker-compose up test-system
Creating testcompose_test-system_1
Attaching to testcompose_test-system_1
ERROR: Couldn't connect to Docker daemon at http+docker://localunixsocket - is it running?

If it's at a non-standard location, specify the URL with the DOCKER_HOST environment variable.
```

My user is added to `docker` group, and the output is the same when run from `sudo`. 

It manages to attach only with `docker-compose run` command.
I tried the experiment with same outcomes on `Docker version 1.9.1, build a34a1d5`, and compose versions `1.4.0`, `1.6.0`, and `1.6.2`.

UPD: 
docker info output:

```
$ docker info
Containers: 2
Images: 152
Server Version: 1.9.1
Storage Driver: aufs
 Root Dir: /var/lib/docker/aufs
 Backing Filesystem: extfs
 Dirs: 156
 Dirperm1 Supported: false
Execution Driver: native-0.2
Logging Driver: json-file
Kernel Version: 3.13.0-29-generic
Operating System: Ubuntu 14.04.2 LTS
CPUs: 8
Total Memory: 11.62 GiB
Name: romang-HP
ID: GDTJ:MBV5:DD5J:3IO2:T4K4:6A4P:QMXM:PLER:TIQA:2MIH:UZ52:VVDE
WARNING: No swap limit support
```

`docker-compose --verpose up` output:

```
$ docker-compose --verbose up test-system
compose.config.config.find: Using configuration files: ./docker-compose.yml
docker.auth.auth.load_config: File doesn't exist
compose.cli.command.get_client: docker-compose version 1.6.0, build d99cad6
docker-py version: 1.7.0
CPython version: 2.7.9
OpenSSL version: OpenSSL 1.0.1e 11 Feb 2013
compose.cli.command.get_client: Docker base_url: http+docker://localunixsocket
[...]

PR #3168, opened 2016-03-18
Handle all timeout errors consistently

dnephin:
Partially addresses the issues in #3106


aanand:
LGTM. Test failure seems unrelated?

dnephin:
most likely, but it seems similar to the test that I fixed in the first commit, so maybe I can fix this flake in the same way. I'll take a look.

dnephin:
Fix for the test flake is in #3181

PR #3137, opened 2016-03-15
Add support for docker run --tmpfs flag.

dnephin:
Carry #2978
Fixes #2778

PR #3179, opened 2016-03-21
Only allow tmpfs with the v2 format

dnephin:
I missed this in #3137

I believe tmpfs is only 1.10+ so we should only add it to v2.


shin-:
LGTM :+1:

PR #3181, opened 2016-03-21
Fix flaky partial_change state test

dnephin:
I believe the changes to run `up` in parallel are what caused this flake.

```
tests/integration/state_test.py:62: in test_partial_change
    self.assertEqual(len(new_containers), 2)
E   AssertionError: 1 != 2
```

There are two significant changes here:

1. the change to use `filter(ready, objects)` seems to fix the primary issue.
2. the change to use `command: 'top`` fixed another flake with this test, with a "can not stop this container", which I believe is caused by a race when the stop happens at the exact time the container exits itself naturally.

### DIFF
--- a/compose/parallel.py
+++ b/compose/parallel.py
@@ -32,8 +32,8 @@ def parallel_execute(objects, func, get_name, msg, get_deps=None):
 
     done = 0
     errors = {}
+    results = []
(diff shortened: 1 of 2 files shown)


aanand:
LGTM

PR #3187, opened 2016-03-23
Update wordpress example to use official images

bfirsh:
The orchardup images are very old and not maintained.


dnephin:
LGTM

PR #3193, opened 2016-03-24
WIP: Bump 1.7.0

dnephin:
There are still a couple PRs to cherry-pick, but I wrote up the release notes. It's late so I'll read over them again tomorrow to check for errors.


CHANGELOG.md
```diff
@@ -1,6 +1,82 @@
...
+    are printed. Use `-f` to get the old default behaviour.
+
+-   Booleans are no longer allows as keys for mappings in the Compose file
+    (for keys `environment`, `labels` and `extra_hosts`). Previously this
+    was a warning. Boolean values should be quoted so they become string values.
+
+New Features
+
+-   Add support for a `docker-compose.env` file to define default values for
+    any environment variables read by Compose.
```
aanand: It's just `.env`, right?

Maybe:

> Compose now looks for a `.env` file in the directory where it's run and will read any environment variables defined inside, if they're not already set in the shell environment. This lets you easily set defaults for variables used in the Compose file, or for any of the `COMPOSE_*` or `DOCKER_*` variables.

CHANGELOG.md
```diff
@@ -1,6 +1,82 @@
...
+
+-   Add a new flag to both `docker-compose up` and `docker-compose down` to
+    remove containers for services that were removed from the Compose file.
```
aanand: Would be good to explicitly say what the flag is

CHANGELOG.md
```diff
@@ -1,6 +1,82 @@
...
+
+-   Add support for removing containers created by `docker-compose run` to
+    `docker-compose rm` and `down`.
```
aanand: And here

aanand:
Rather than doing a million annoying comments, I just did some rewording in a gist: https://gist.github.com/aanand/fbf303fdb36f76a6b335

PR #3196, opened 2016-03-24
bash completion for new TLS options

albers:
This adds bash completion for #3139.

The new options are special ones: In several places, the completion script invokes `docker-compose` to gather information. The new options have to be used in these secondary invocations as well.
The existing solution (used for `--file` and `--project-name`) did not scale, so a refactoring step was prepended to the change.

Please schedule for 1.7.0 as the corresponding feature exists in that release.

### DIFF
--- a/contrib/completion/bash/docker-compose
+++ b/contrib/completion/bash/docker-compose
@@ -18,31 +18,28 @@
 
 
 __docker_compose_q() {
-	local file_args
-	if [ ${#compose_files[@]} -ne 0 ] ; then
-		file_args="${compose_files[@]/#/-f }"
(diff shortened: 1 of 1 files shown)


albers:
Hints: 
* The general idea of treating top-level options and the  `__docker_compose_to_extglob` and `__docker_compose_to_alternatives` functions are taken from the docker completion script.
* I tested the completion in the `albers/albers/bash-completion-mac` image, so I'm quite sure there are no problems for Mac users.

dnephin:
LGTM

aanand:
LGTM

PR #3206, opened 2016-03-24
bash completion for `docker-compose exec`

albers:
Please schedule for 1.7.0 as the corresponding feature is present in that release.


dnephin:
LGTM

PR #3218, opened 2016-03-25
Add zsh completion for 'docker-compose run -w --workdir'

sdurrheimer:
Ref. #3136

@albers thx for the ping

Signed-off-by: Steve Durrheimer <s.durrheimer@gmail.com>

PR #3217, opened 2016-03-25
Add zsh completion for 'docker-compose up --build'

sdurrheimer:
Ref. 1.7-rc1 changelog

@albers thx for the ping

Signed-off-by: Steve Durrheimer <s.durrheimer@gmail.com>

PR #3216, opened 2016-03-25
Add zsh completion for 'docker-compose logs -f --follow --tail -t --t…

sdurrheimer:
…imestamps'

Ref. #2720

@albers thx for the ping

Signed-off-by: Steve Durrheimer <s.durrheimer@gmail.com>

PR #3215, opened 2016-03-25
Add zsh completion for 'docker-compose exec' command

sdurrheimer:
Ref. #2023

@albers thx for the ping

Signed-off-by: Steve Durrheimer <s.durrheimer@gmail.com>

PR #3221, opened 2016-03-27
Fix zsh autocomplete

aeriksson:
This fixes a few issues with the zsh autocomplete script, and brings it up to date with the current docker-compose commands.

Changes:
- Uses `docker-compose config` instead of manually trying to find and parse the docker-compose file. This fixes a few issues:
  - Now works for version 2 docker-compose files.
  - Will work correctly even if the current working directory is a subdirectory of the folder with the docker-compose file.
- Now has all the flags for all the docker-compose commands.
- Command help texts are now in line with what `docker-compose help x` produces (fixed some typos and outdated/incorrect help texts).
- Now passes all relevant flags when shelling out to `docker-compose` and `docker`, so autocomplete works even if `--tls`, `--tlskey`, etc are set.
- Fixed issue where autocomplete for `build`, `pull` would only show some services.
- Fixed issue with filtering on already selected services when service names are substrings of one another.
- All functions are now prefixed with the same number of underscores.

### DIFF
--- a/contrib/completion/zsh/_docker-compose
+++ b/contrib/completion/zsh/_docker-compose
@@ -198,9 +198,9 @@ __docker-compose_subcommand() {
         (create)
             _arguments \
                 $opts_help \
-                "(--no-recreate --no-build)--force-recreate[Recreate containers even if their configuration and image haven't changed. Incompatible with --no-recreate.]" \
-                "(--force-recreate)--no-build[If containers already exist, don't recreate them. Incompatible with --force-recreate.]" \
-                "(--force-recreate)--no-recreate[Don't build an image, even if it's missing]" \
+                "(--no-recreate)--force-recreate[Recreate containers even if their configuration and image haven't changed. Incompatible with --no-recreate.]" \
+                "(--force-recreate)--no-recreate[If containers already exist, don't recreate them. Incompatible with --force-recreate.]" \
+                "--no-build[Don't build an image, even if it's missing.]" \
                 '*:services:__docker-compose_services_all' && ret=0
             ;;
         (down)
@@ -325,9 +325,9 @@ __docker-compose_subcommand() {
                 '--build[Build images before starting containers.]' \
                 '--no-color[Produce monochrome output.]' \
                 "--no-deps[Don't start linked services.]" \
-                "--force-recreate[Recreate containers even if their configuration and image haven't changed. Incompatible with --no-recreate.]" \
-                "--no-recreate[If containers already exist, don't recreate them.]" \
-                "--no-build[Don't build an image, even if it's missing]" \
+                "(--no-recreate)--force-recreate[Recreate containers even if their configuration and image haven't changed. Incompatible with --no-recreate.]" \
+                "(--force-recreate)--no-recreate[If containers already exist, don't recreate them. Incompatible with --force-recreate.]" \
+                "--no-build[Don't build an image, even if it's missing.]" \
                 "(-d)--abort-on-container-exit[Stops all containers if any container was stopped. Incompatible with -d.]" \
                 '(-t --timeout)'{-t,--timeout}"[Specify a shutdown timeout in seconds. (default: 10)]:seconds: " \
                 "--remove-orphans[Remove containers for services not defined in the Compose file]" \
--- a/contrib/completion/zsh/_docker-compose
+++ b/contrib/completion/zsh/_docker-compose
@@ -207,7 +207,8 @@ __docker-compose_subcommand() {
             _arguments \
                 $opts_help \
                 "--rmi[Remove images, type may be one of: 'all' to remove all images, or 'local' to remove only images that don't have an custom name set by the 'image' field]:type:(all local)" \
-                '(-v --volumes)'{-v,--volumes}"[Remove data volumes]" && ret=0
+                '(-v --volumes)'{-v,--volumes}"[Remove data volumes]" \
+                '--remove-orphans[Remove containers for services not defined in the Compose file]' && ret=0
             ;;
         (events)
             _arguments \
@@ -329,6 +330,7 @@ __docker-compose_subcommand() {
                 "--no-build[Don't build an image, even if it's missing]" \
                 "(-d)--abort-on-container-exit[Stops all containers if any container was stopped. Incompatible with -d.]" \
                 '(-t --timeout)'{-t,--timeout}"[Specify a shutdown timeout in seconds. (default: 10)]:seconds: " \
+                "--remove-orphans[Remove containers for services not defined in the Compose file]" \
                 '*:services:__docker-compose_services_all' && ret=0
             ;;
         (version)
--- a/contrib/completion/zsh/_docker-compose
+++ b/contrib/completion/zsh/_docker-compose
@@ -19,26 +19,13 @@
 #  * @felixr docker zsh completion script : https://github.com/felixr/docker-zsh-completion
 # -------------------------------------------------------------------------
 
-# For compatibility reasons, Compose and therefore its completion supports several
-# stack compositon files as listed here, in descending priority.
-# Support for these filenames might be dropped in some future version.
-__docker-compose_compose_file() {
-    local file
-    for file in docker-compose.y{,a}ml ; do
-        [ -e $file ] && {
-            echo $file
-            return
-        }
-    done
-    echo docker-compose.yml
-}
-
 # Extracts all service names from docker-compose.yml.
 ___docker-compose_all_services_in_compose_file() {
     local already_selected
     local -a services
     already_selected=$(echo $words | tr " " "|")
-    awk -F: '/^[a-zA-Z0-9]/{print $1}' "${compose_file:-$(__docker-compose_compose_file)}" 2>/dev/null | grep -Ev "$already_selected"
+    docker-compose config --services 2>/dev/null \
+        | grep -Ev "$already_selected"
 }
 
 # All services, even those without an existing container
@@ -57,7 +44,12 @@ ___docker-compose_services_with_key() {
     local -a buildable
     already_selected=$(echo $words | tr " " "|")
     # flatten sections to one line, then filter lines containing the key and return section name.
-    awk '/^[a-zA-Z0-9]/{printf "\n"};{printf $0;next;}' "${compose_file:-$(__docker-compose_compose_file)}" 2>/dev/null | awk -F: -v key=": +$1:" '$0 ~ key {print $1}' 2>/dev/null | grep -Ev "$already_selected"
+    docker-compose config 2>/dev/null \
+        | sed -n -e '/^services:/,/^[^ ]/p' \
+        | sed -n 's/^  //p' \
+        | awk '/^[a-zA-Z0-9]/{printf "\n"};{printf $0;next;}' \
+        | awk -F: -v key=": +$1:" '$0 ~ key {print $1}' \
+        | grep -Ev "$already_selected"
 }
 
 # All services that are defined by a Dockerfile reference
--- a/contrib/completion/zsh/_docker-compose
+++ b/contrib/completion/zsh/_docker-compose
@@ -185,7 +185,7 @@ __docker-compose_subcommand() {
             _arguments \
                 $opts_help \
                 '--force-rm[Always remove intermediate containers.]' \
-                '--no-cache[Do not use cache when building the image]' \
+                '--no-cache[Do not use cache when building the image.]' \
                 '--pull[Always attempt to pull a newer version of the image.]' \
                 '*:services:__docker-compose_services_from_build' && ret=0
             ;;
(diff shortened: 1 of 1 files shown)


aeriksson:
So I realized this PR conflicts with #3214 #3215 #3216 #3217 #3218. They weren't there when I set out to work on this :)

Anyway, this has the changes in those PRs, as well as some other stuff.

If those get merged first, I can rebase this branch accordingly.

dnephin:
cc @sdurrheimer

sdurrheimer:
@dnephin @aeriksson Yes, we should merge others PRs before considering merging this one in order to have a clear state (#3214 #3215 #3216 #3217 #3218).

I would then be able to proper review this after a rebase.

aeriksson:
@sdurrheimer rebased!

contrib/completion/zsh/_docker-compose
```diff
@@ -52,7 +52,8 @@ __docker-compose_services_with_key() {
         | sed -n -e '/^services:/,/^[^ ]/p' \
         | sed -n 's/^  //p' \
         | awk '/^[a-zA-Z0-9]/{printf "\n"};{printf $0;next;}' \
-        | awk -F: -v key=": +$1:" '$0 ~ key {print $1}' \
```
sdurrheimer: Can you squash that changed with @cbfe6d0 ?

contrib/completion/zsh/_docker-compose
```diff
@@ -29,7 +29,7 @@ __docker-compose_all_services_in_compose_file() {
     local -a services
     already_selected=$(echo $words | tr " " "|")
     __docker-compose_q config --services \
-        | grep -Ev "$already_selected"
```
sdurrheimer: Same here, can you squash those changes with @cbfe6d0 ?

contrib/completion/zsh/_docker-compose
```diff
@@ -20,7 +20,7 @@
 # -------------------------------------------------------------------------
 
 # Extracts all service names from docker-compose.yml.
-___docker-compose_all_services_in_compose_file() {
```
sdurrheimer: There is normally a meaning about triple underscore prefix which is sub-functions used by complete function. Ex: `__docker-compose_services_from_build` using `___docker-compose_services_with_key`. But I'm okay about using only double underscore prefix.

contrib/completion/zsh/_docker-compose
```diff
@@ -359,10 +362,17 @@ _docker-compose() {
 
     _arguments -C \
         '(- :)'{-h,--help}'[Get help]' \
-        '--verbose[Show more output]' \
-        '(- :)'{-v,--version}'[Print version and exit]' \
         '(-f --file)'{-f,--file}'[Specify an alternate docker-compose file (default: docker-compose.yml)]:file:_files -g "*.yml"' \
         '(-p --project-name)'{-p,--project-name}'[Specify an alternate project name (default: directory name)]:project name:' \
+        '--verbose[Show more output]' \
+        '(- :)'{-v,--version}'[Print version and exit]' \
+        '(-H --host)'{-H,--host}'[Daemon socket to connect to]:host:' \
```
sdurrheimer: Since you are alphabetically re-ordering `--verbose` and `-v --version`, can you do the same with the new options (`--tls` etc.) ?
I would also prefer to have separated commits per group of new command options, for a better changes history lisibility:
- One for `docker-compose` base options (`--tls` etc.)
- One for adding `--remove-orphans` to `up` and `down`.
- One making `--force-recreate`, `--no-recreate`, `--no-build` and `--build` common between `create ` and `up` commands. See my other comment.
- One for fixing little errors.

contrib/completion/zsh/_docker-compose
```diff
@@ -198,21 +198,23 @@ __docker-compose_subcommand() {
         (create)
             _arguments \
                 $opts_help \
-                "(--no-recreate --no-build)--force-recreate[Recreate containers even if their configuration and image haven't changed. Incompatible with --no-recreate.]" \
-                "(--force-recreate)--no-build[If containers already exist, don't recreate them. Incompatible with --force-recreate.]" \
-                "(--force-recreate)--no-recreate[Don't build an image, even if it's missing]" \
+                "(--no-recreate)--force-recreate[Recreate containers even if their configuration and image haven't changed. Incompatible with --no-recreate.]" \
```
sdurrheimer: Given that `--force-recreate`, `--no-recreate`, `--no-build` and `--build` are now the same for the `create` and `up` commands, can you extract those options into a separated variable like we do on the docker/docker zsh completion ? (https://github.com/docker/docker/blob/master/contrib/completion/zsh/_docker#L537)

sdurrheimer:
@aeriksson Thanks ! I've made some comments.

Normally, when changes are needed in `bash` and `zsh` completion for the `docker/compose` and `docker/docker` repos, we prefer separated PRs instead of grouping many changes in a single one. But, given that it is your first contribution about this (which is very welcomed), we can do with it this time. Don't know what @dnephin think about this.

aanand:
@sdurrheimer A single PR is absolutely fine with me - more important that the commits are separated out.

Regarding the changes to the options, it'd be nice as you say to have even more fine-grained commits there, but I'm not hugely bothered.

LGTM

contrib/completion/zsh/_docker-compose
aeriksson: Yeah, I'll break the commit up.

Regarding the sorting, I reordered the flags to match the order given by `docker-compose help x`. I figured this would make the script easier to maintain.

Do you want me to do alphabetical ordering instead?

contrib/completion/zsh/_docker-compose
aeriksson: I could join cbfe6d0 and 7e80ab4.

However, I'm not sure this is a good idea. They address two separate issues, which just happened to be on the same line (originally), so I think it's a good idea to leave them as two separate commits.

Especially if I'll be splitting de216f3 up into smaller commits.

contrib/completion/zsh/_docker-compose
sdurrheimer: ping @albers This might interest you given that we are using the same services parsing method.

contrib/completion/zsh/_docker-compose
sdurrheimer: I'm ok with that, we keep cbfe6d0 and 7e80ab4 as separated commits.

contrib/completion/zsh/_docker-compose
albers: @sdurrheimer Yes, that's identical with my version.
It improves readibility at the cost of one extra command invocation. I think that's OK.
I have two suggestions, though which i'll add below.
@aeriksson would you mind to adjust this function in the bash completion as well so that both completions stay aligned?

contrib/completion/zsh/_docker-compose
```diff
@@ -52,7 +52,8 @@ __docker-compose_services_with_key() {
         | sed -n -e '/^services:/,/^[^ ]/p' \
         | sed -n 's/^  //p' \
         | awk '/^[a-zA-Z0-9]/{printf "\n"};{printf $0;next;}' \
-        | awk -F: -v key=": +$1:" '$0 ~ key {print $1}' \
+        | grep " \+$1:" \
```
albers: Please do not drop the first `:` here. I'd like to be as specific as possible.

contrib/completion/zsh/_docker-compose
```diff
@@ -52,7 +52,8 @@ __docker-compose_services_with_key() {
...
+        | sed "s/:.*//g" \
         | grep -Ev "$already_selected"
```
albers: How about `cut -d: -f1`? This should be faster than `sed`.

contrib/completion/zsh/_docker-compose
aeriksson: @albers I'm dropping the first `:` on purpose :)
With the extra `:`, the autocomplete only works if the docker-compose file is ordered in a specific way.

For instance, while this service will be matched (assuming `$1` is `build`):
```yaml
my-app:
  build: ./my-dir
  ports:
    - 80
```
…this one won't:
```yaml
my-app:
  ports:
    - 80
  build: ./my-dir
```

contrib/completion/zsh/_docker-compose
aeriksson: @albers sure, I'll create a bash PR too

contrib/completion/zsh/_docker-compose
aeriksson: Sure thing

contrib/completion/zsh/_docker-compose
albers: @aeriksson yes, you're right. Thanks for spotting this.

contrib/completion/zsh/_docker-compose
sdurrheimer: > Regarding the sorting, I reordered the flags to match the order given by `docker-compose help x`. I figured this would make the script easier to maintain.

> Do you want me to do alphabetical ordering instead?

Oh my bad, it's fine like that.

sdurrheimer:
@aeriksson Any news ?

aeriksson:
@sdurrheimer Sorry, this sort of fell by the wayside. I've split the flag changes up into several commits and updated the flag help texts with the changes in master.

contrib/completion/zsh/_docker-compose
sdurrheimer: +1 on replacing `sed` by `cut` here.

contrib/completion/zsh/_docker-compose
sdurrheimer: Waiting for this change and the `sed` by `cut` replacement to do a final review with testing on my machine.

contrib/completion/zsh/_docker-compose
aeriksson: Done!

contrib/completion/zsh/_docker-compose
aeriksson: So as it turns out, `--build` is actually slightly different between `up` and `create`, so I've let that one be.

However, I've extracted `--force-recreate`, `--no-recreate`, `--no-build`, and all other duplicated flags into variables.

sdurrheimer:
LGTM

contrib/completion/zsh/_docker-compose
```diff
@@ -19,52 +19,49 @@
...
-            echo $file
-            return
-        }
-    done
-    echo docker-compose.yml
+__docker-compose_q() {
+    docker-compose 2>/dev/null $compose_options "$@"
 }
 
 # Extracts all service names from docker-compose.yml.
```
albers: Please update this comment

aeriksson: @albers what should the comment say?

I removed it to be in line with the bash script:
https://github.com/docker/compose/blob/master/contrib/completion/bash/docker-compose#L20-L22

albers: I'm fine with removing the comment.

sdurrheimer:
Hello, what is the state of this PR ?

aeriksson:
AFAICT it's ready to merge.

contrib/completion/zsh/_docker-compose
sdurrheimer: @aeriksson Can you do that change ?

aeriksson: @sdurrheimer not sure what the change is…

Maybe I misunderstood something, but I think @albers asked me to update that comment rather than removing it.

I then pointed out that I removed it because it was no longer needed, and there is no corresponding comment on the same function in the bash script.

I take what @albers wrote after to mean that after some deliberation, he was okay with the comment being removed.

Again, I might be wrong here, and I'd gladly add a comment if there's a need for one :)

albers: @aeriksson The comment refers to the previous way of harvesting the service names from docker-compose.yml. The comment pointed out that the service list had some limitations: it did not take into account several other configuration files that were added over time.

Now, we use an official API for that task: `docker-compose config`. We let `docker-compose` combine all the configuration sources. But the comment was not updated and now is wrong.

There are two ways to resolve this mismatch: 

1. change the comment to "Extracts all service names from the output of docker-compose config"
2. drop the comment

I favour solution 2 because the comment from solution 1 is trivial. The new code does not need a comment any more.

aeriksson: Yes, and solution 2 is what is implemented in this PR, so no further changes should be required, right?

sdurrheimer: @aeriksson: @albers is talking about the comment line 26/36

aeriksson: Oh, I see. Updated that comment to say "All services defined in docker-compose.yml" rather than "Extracts all service names from docker-compose.yml".

sdurrheimer:
Ok LGTM

sdurrheimer:
@dnephin @aanand Can we get this merged soon ? This PR fixes the `version: 2` support and is kinda blocking me my upcoming PRs for new docker-compose command like `bundle` or `push.`

dnephin:
LGTM

aanand:
LGTM, thanks!

PR #3222, opened 2016-03-28
Rails Docs: Add nodejs to apt-get install command

jonlemmon:
When using the latest version of Rails, the tutorial currently errors when running `docker-compose up` with the following error:

```
/usr/local/lib/ruby/gems/2.3.0/gems/bundler-1.11.2/lib/bundler/runtime.rb:80:in `rescue in block (2 levels) in require': There was an error while trying to load the gem 'uglifier'. (Bundler::GemRequireError)
```

Installing nodejs in the build fixes the issue.

### DIFF
--- a/docs/rails.md
+++ b/docs/rails.md
@@ -22,7 +22,7 @@ container. This is done using a file called `Dockerfile`. To begin with, the
 Dockerfile consists of:
 
     FROM ruby:2.2.0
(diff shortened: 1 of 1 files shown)


dnephin:
LGTM

bfirsh:
Because of course you need Node.js installed to use Ruby on Rails.

LGTM, regretfully

PR #3224, opened 2016-03-28
ignore eclipse project files in .gitignore

max8899:


dnephin:
It's not necessary to add this to the project repo. You can add it to either `.git/info/exclude` or you can set a global `gitignore` for your user.

Issue #3210, opened 2016-03-25
Inconsistent behaviour with CA verification flags


tombee:
After testing out the new TLS configuration flags (that should make compose consistent with engine).  In particular I'm testing the scenario where we want to do server verification using a CA root certificate. I'm seeing a few issues where I have to enable `--skip-hostname-check`:

I'm testing this by running the master build docker-compose binary inside a `debian:jessie` image:

Just to confirm, `-H <hostname:port> --tlsverify --tlscacert ca.pem info` works okay with the engine CLI:

```
root@b3ace98de2f8 $ docker -H 192.168.122.153:443 --tlsverify --tlscacert ca.pem info 
Containers: 22
 Running: 10
 Paused: 0
 Stopped: 12
Images: 18
...
```

But I can't use the same parameters with compose:

```
root@b3ace98de2f8 $ docker-compose -H 192.168.122.153:443 --tlsverify --tlscacert ca.pem up -d 
Traceback (most recent call last):
  File "<string>", line 3, in <module>
  File "compose/cli/main.py", line 56, in main
  File "compose/cli/main.py", line 107, in perform_command
  File "compose/cli/main.py", line 722, in up
  File "compose/project.py", line 371, in up
  File "compose/project.py", line 179, in get_services_without_duplicate
  File "compose/service.py", line 472, in remove_duplicate_containers
  File "compose/service.py", line 479, in duplicate_containers
  File "compose/service.py", line 145, in containers
  File ".tox/py27/lib/python2.7/site-packages/docker/api/container.py", line 69, in containers
  File ".tox/py27/lib/python2.7/site-packages/docker/utils/decorators.py", line 47, in inner
  File ".tox/py27/lib/python2.7/site-packages/docker/client.py", line 112, in _get
  File ".tox/py27/lib/python2.7/site-packages/requests/sessions.py", line 477, in get
  File ".tox/py27/lib/python2.7/site-packages/requests/sessions.py", line 465, in request
  File ".tox/py27/lib/python2.7/site-packages/requests/sessions.py", line 573, in send
  File ".tox/py27/lib/python2.7/site-packages/requests/adapters.py", line 370, in send
  File ".tox/py27/lib/python2.7/site-packages/requests/packages/urllib3/connectionpool.py", line 544, in urlopen
  File ".tox/py27/lib/python2.7/site-packages/requests/packages/urllib3/connectionpool.py", line 341, in _make_request
[...]

PR #3226, opened 2016-03-29
Fix assert_hostname logic in tls_config_from_options

shin-:
1. `assert_hostname=True` is an error (at least for recent versions of urllib3 / requests). The code has been updated so we don't set that value anymore.
2. Improved hostname detection (`urlparse` would fail if no protocol was specified)
3. Avoid unnecessary lookup / computation if `--skip-hostname-check` is set.

This should help a little with #3210 . The main issue that remains is when the CA declares a hostname different from the one we're connecting through, causing the `SSLError: hostname '127.0.0.1' doesn't match u'localhost'` error. While this seems correct to do, the docker CLI obviously behaves differently. 

We could elect to skip the hostname check by default, or try and retrieve the CA's fingerprint and use `assert_fingerprint` instead. From [urllib3 docs](http://urllib3.readthedocs.org/en/latest/pools.html):
> VerifiedHTTPSConnection uses one of `assert_fingerprint`, `assert_hostname` and host in this order to verify connections. If `assert_hostname` is False, no verification is done.

### DIFF
--- a/compose/cli/docker_client.py
+++ b/compose/cli/docker_client.py
@@ -21,24 +21,37 @@ def tls_config_from_options(options):
     cert = options.get('--tlscert')
     key = options.get('--tlskey')
     verify = options.get('--tlsverify')
-    hostname = urlparse(options.get('--host') or '').hostname
+    host = options.get('--host')
+    skip_hostname_check = options.get('--skip-hostname-check', False)
+
+    if not skip_hostname_check:
+        hostname = urlparse(host).hostname if host else None
+        # If the protocol is omitted, urlparse fails to extract the hostname.
+        # Make another attempt by appending a protocol.
+        if not hostname and host:
+            hostname = urlparse('tcp://{0}'.format(host)).hostname
 
     advanced_opts = any([ca_cert, cert, key, verify])
 
     if tls is True and not advanced_opts:
         return True
-    elif advanced_opts:
+    elif advanced_opts:  # --tls is a noop
         client_cert = None
         if cert or key:
             client_cert = (cert, key)
+
+        assert_hostname = None
+        if skip_hostname_check:
+            assert_hostname = False
+        elif hostname:
+            assert_hostname = hostname
+
         return TLSConfig(
             client_cert=client_cert, verify=verify, ca_cert=ca_cert,
-            assert_hostname=(
-                hostname or not options.get('--skip-hostname-check', False)
-            )
+            assert_hostname=assert_hostname
         )
-    else:
-        return None
+
+    return None
 
 
 def docker_client(environment, version=None, tls_config=None, host=None):
--- a/compose/cli/docker_client.py
+++ b/compose/cli/docker_client.py
@@ -49,7 +49,7 @@ def docker_client(environment, version=None, tls_config=None, host=None):
                  "Please use COMPOSE_HTTP_TIMEOUT instead.")
 
     try:
-        kwargs = kwargs_from_env(assert_hostname=False, environment=environment)
+        kwargs = kwargs_from_env(environment=environment)
     except TLSParameterError:
         raise UserError(
             "TLS configuration is invalid - make sure your DOCKER_TLS_VERIFY "
--- a/compose/cli/docker_client.py
+++ b/compose/cli/docker_client.py
@@ -7,7 +7,6 @@ from docker import Client
 from docker.errors import TLSParameterError
 from docker.tls import TLSConfig
 from docker.utils import kwargs_from_env
-from requests.utils import urlparse
 
 from ..const import HTTP_TIMEOUT
 from .errors import UserError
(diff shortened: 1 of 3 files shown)


dnephin:
Oops, i forgot to push my changes to the `dind` ci script, so this build failed on the next rc release. Fixing it now.

compose/cli/docker_client.py
```diff
@@ -21,24 +21,37 @@ def tls_config_from_options(options):
...
     key = options.get('--tlskey')
     verify = options.get('--tlsverify')
-    hostname = urlparse(options.get('--host') or '').hostname
+    host = options.get('--host')
+    skip_hostname_check = options.get('--skip-hostname-check', False)
+
+    if not skip_hostname_check:
+        hostname = urlparse(host).hostname if host else None
+        # If the protocol is omitted, urlparse fails to extract the hostname.
+        # Make another attempt by appending a protocol.
```
dnephin: prepending

dnephin:
LGTM when green

aanand:
I vote to skip the hostname check by default, if that's what the docker CLI does.

tombee:
Apologies for my ignorance on this one, have been having a look at the python docs for `assert_hostname` and trying to figure out what the consequences are.

Does this mean that it will still validate the server certificate?  But only verifying that the hostname you're connecting to is one of the SANs listed in the certificate, rather than forcing hostname resolution?

NathanMcCauley:
In general we should not skip hostname verification. I don't understand fully why this is necessary. Does the docker-cli have a special case for localhost that we're missing?

compose/cli/docker_client.py
```diff
@@ -21,24 +21,37 @@ def tls_config_from_options(options):
...
-    elif advanced_opts:
+    elif advanced_opts:  # --tls is a noop
         client_cert = None
         if cert or key:
             client_cert = (cert, key)
+
+        assert_hostname = None
+        if skip_hostname_check:
+            assert_hostname = False
+        elif hostname:
```
endophage: Dangerous use of `hostname` as it could be possible in a future change for it to become undefined. Would be good to give it a default starting value at the start of the func.

shin-:
@tombee This is what it comes down to: https://docs.python.org/2/library/ssl.html#ssl.match_hostname . It checks the `subjectAltName` field in the peer's (in our case, the server's) certificate for a match with the hostname we're using to connect. Verification against the certificate itself is still done even if we disable that particular check.

@NathanMcCauley I don't know a lot about the CLI, except that it's using its own implementation of the TLS protocol and isn't tied to OpenSSL like we are, so matching the behavior is more complicated than just using the same options. It's important to note though that the problem isn't limited to localhost, as seen in #3210 for example: `ERROR: SSL error: hostname '192.168.122.153' doesn't match 'myhost'`.

tombee:
If I create an `/etc/hosts` entry:

```
hacker    172.17.0.1`
```

Then I do:

```
$ docker-compose -H hacker:443 --tlsverify --tlscacert ca.pem --skip-hostname-check up -d
downloads_redis_1 is up-to-date
```

The Docker CLI with the same options rejects the connection since `hacker` isn't amongst the SANs:

```
$ docker -H hacker:443 --tlsverify --tlscacert ca.pem version
Client:
 Version:      1.11.0-rc2
 API version:  1.23
 Go version:   go1.5.3
 Git commit:   388f544
 Built:        Fri Mar 25 19:39:29 2016
 OS/Arch:      linux/amd64
An error occurred trying to connect: Get https://hacker:443/v1.23/version: x509: certificate is valid for debian, not hacker
```

Without the `--skip-hostname-check` flag, compose seems to have the correct behaviour.

```
$ docker-compose -H tcp://hacker:443 --tlsverify --tlscacert ca.pem up -d
ERROR: SSL error: hostname 'hacker' doesn't match 'debian'
```

So it seems like when we supply `--skip-hostname-check`, it turns off server cert verification entirely?

endophage:
@shin- @tombee I'm wondering if this is also part of the problem: https://github.com/shazow/urllib3/blob/master/urllib3/packages/ssl_match_hostname/_implementation.py#L89

I need to get my hands on a sample certificate to work out if that's part of the problem.

shin-:
@endophage Not really - urllib3 just backports the Python 3.3 version of the algorithm and uses it over the standard lib's version if it's older. See the `__init__.py` in the same folder https://github.com/shazow/urllib3/blob/master/urllib3/packages/ssl_match_hostname/__init__.py

endophage:
Ah ok. 

For the record, Sec team is strongly against skipping hostname check. @aanand as far as I can tell, the CLI isn't skipping it, it's handling IPs correctly.

shin-:
I'm working on backporting the IP address suport into docker-py.

shin-:
Fix here: https://github.com/docker/docker-py/pull/1013

shin-:
I don't know why this breaks when I try to use a pypi release for the docker-py dependency. Same code from git is working fine.

aanand:
This seems to be the issue:

```
Collecting py2-ipaddress>=3.4.1 (from docker-py==1.8.0rc4->-r requirements.txt (line 3))
  Using cached py2-ipaddress-3.4.1.tar.gz
    Complete output from command python setup.py egg_info:
    py2-ipaddress: For Python 3.x, please use the official ipaddress module.
```

Looks like `py2-ipaddress` refuses to install on py3.

shin-:
Yes, but it shouldn't try to install it, there's an environment marker (
https://www.python.org/dev/peps/pep-0345/#environment-markers ) on that
dependency.

shin-:
That one test failure looks like a flake. I think this is ready.

aanand:
LGTM

PR #3232, opened 2016-03-29
updating .gitignore to ignore .DS_Store and .idea

twitherspoon:


dnephin:
These don't really need to be checked into the project. They can be ignored by using `.git/info/exclude` or setting a global git ignore in `~/.gitconfig`

Issue #3191, opened 2016-03-23
docker-compose help subcommand does not list help


cloojure:
The result of `docker-compose help` is not helpful:

```
> docker-compose help
Get help on a command.

Usage: help COMMAND
```

`docker-compose` should follow the example of both `docker` and `docker-machine`, which print help/usage text when invoked alone or with the "help" subcommand:

```
> docker 
Usage: docker [OPTIONS] COMMAND [arg...]
       docker daemon [ --help | ... ]
       docker [ --help | -v | --version ]

A self-sufficient runtime for containers.
...
```

```
> docker help
Usage: docker [OPTIONS] COMMAND [arg...]
       docker daemon [ --help | ... ]
       docker [ --help | -v | --version ]

A self-sufficient runtime for containers.
...
```

ijc:
Issue grooming: Seems to have be fixed at some point:
```
$ docker-compose help | wc -l
57
$ docker-compose help | head -n 15
Define and run multi-container applications with Docker.

Usage:
  docker-compose [-f <arg>...] [options] [COMMAND] [ARGS...]
  docker-compose -h|--help

Options:
  -f, --file FILE             Specify an alternate compose file
                              (default: docker-compose.yml)
  -p, --project-name NAME     Specify an alternate project name
                              (default: directory name)
  --verbose                   Show more output
  --log-level LEVEL           Set log level (DEBUG, INFO, WARNING, ERROR, CRITICAL)
  --no-ansi                   Do not print ANSI control characters
  -v, --version               Print version and exit
```

PR #3234, opened 2016-03-29
3191 help cli feature

twitherspoon:
Addressing issue #3191. Added code to output the top level command options if docker-compose help with no command options provided. Provides similar functionality to other docker cli tools.

### DIFF
--- a/compose/cli/docopt_command.py
+++ b/compose/cli/docopt_command.py
@@ -34,10 +34,6 @@ class DocoptDispatcher(object):
         if docstring is None:
             raise NoSuchCommand(command, self)
 
-        # checks to see if the command is help and if so outputs top level command options
-        if command == 'help' and options['ARGS'] == []:
-            raise SystemExit(docstring + '\n-------------------\n\n' + command_help)
-
         command_options = docopt_full_help(docstring, options['ARGS'], options_first=True)
         return options, handler, command_options
 
--- a/compose/cli/docopt_command.py
+++ b/compose/cli/docopt_command.py
@@ -34,6 +34,10 @@ class DocoptDispatcher(object):
         if docstring is None:
             raise NoSuchCommand(command, self)
 
+        # checks to see if the command is help and if so outputs top level command options
+        if command == 'help' and options['ARGS'] == []:
+            raise SystemExit(docstring + '\n-------------------\n\n' + command_help)
+
         command_options = docopt_full_help(docstring, options['ARGS'], options_first=True)
         return options, handler, command_options
 
--- a/.gitignore
+++ b/.gitignore
@@ -9,5 +9,3 @@
 /venv
 README.rst
 compose/GITSHA
-.DS_Store
-.idea/
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,5 @@
 /venv
 README.rst
 compose/GITSHA
+.DS_Store
+.idea/


compose/cli/docopt_command.py
```diff
@@ -34,6 +34,10 @@ def parse(self, argv):
         if docstring is None:
             raise NoSuchCommand(command, self)
 
+        # checks to see if the command is help and if so outputs top level command options
+        if command == 'help' and options['ARGS'] == []:
+            raise SystemExit(docstring + '\n-------------------\n\n' + command_help)
```
albers: I'm not familiar with python, but this looks as if an error will be thrown, setting the exit code to non-zero and writing to stderr.
As `docker-compose help` is a legal invocation and not an error condition, the command's exit code should be 0 and the output should go to stdout.
Can you please verify this?

aanand:
Thanks. Perhaps it would make more sense to implement this logic in `main.py`, inside `TopLevelCommand.help()`?

compose/cli/docopt_command.py
twitherspoon: Interesting comment and question.  I would not consider myself a python developer.  Just a guy interested in learning Docker and Python and this seemed like a great way to dive in.  Now to your question.  In docopt_command.py I followed the pattern that was being used within that script.  If no command was provided an exception was raised to output the top level commands. If a docstring for the command was not provided an exception was raised.  Likewise I thought if the help command was not used with out a command such as 'docker-compose help run' or 'docker-compose help up' then the top level commands would be outputted to stdout since the help command is intended to be used with a command option and not by itself.

If this is pattern is not ideal or doesn't follow python best practices, might be a good task to place in the list for future clean up.  Thoughts?

twitherspoon:
I thought and tried to put the logic in main.py and here's what I found out and my difficulty.

main.py calls out to DocoptDispatcher in docopt_command.py very early to parse the command line arguments.  During the processing in docopt_command.py the docopt method returns an exception and errors out on line 13 of docopt_command.py when calling the docopt method.  I think it's because there aren't any arguments provided for command.  Because of that error it never actually returns back to main.py.

By the way love the inputs and questions.  Feedback like this is very valuable.

compose/cli/docopt_command.py
albers: @twitherspoon my comment is not python specific, it is about unix best practices. The `docker` command behaves like this. You can verify by checking the exit codes like this:

```bash
$ docker help ; echo $?  # exit code is 0 (OK)
$ docker nonsense ; echo $?  # exit code is 1 (error)
```

and the output channels with
```bash
$ docker help | grep container  # help lines containing "container" (stdout)
$ docker nonsense | grep container  # error message on stderr, lines without "container" appear
```

aanand:
I understand the difficulty - the way docopt works is a bit surprising.

You're absolutely right that it errors out because no argument was passed to `help`. This is because it's parsing the command based on the "Usage" string in `help`'s docstring:

```
Usage: help COMMAND
```

The solution is to wrap `COMMAND` in square brackets to signify that it's an optional argument, and then check the value of `options['COMMAND']` in the method body to decide what to do. Here's how it should look:

```python
@classmethod
def help(cls, options):
    """
    Get help on the Compose command or a subcommand.

    Usage: help [COMMAND]
    """
    if options['COMMAND']:
        subject = get_handler(cls, options['COMMAND'])
    else:
        subject = cls

    raise SystemExit(getdoc(subject))
```

twitherspoon:
@aanand Thanks for the feedback and direction.  Let me work on that.  I did change the the command, but ran into some issues, but need to dive deeper into that to see what it was.  Probably just something wrong I did.  I'll see if I can get it worked out and create a new pull request if I can fix it.

Thanks!

twitherspoon:
Closing addressed by pull request #3263

PR #3249, opened 2016-03-31
Better variable substitution example

aanand:
In most cases, it'd be quite silly to put your Postgres version in an environment variable.


dnephin:
LGTM

PR #3263, opened 2016-04-04
Added code to support no argument help command

twitherspoon:
Addressing issue #3191 . Added code to output the top level command options if docker-compose help with no command options provided. Provides similar functionality to other docker cli tools. Closing pull request #3234, because it is a fixed by this pull request to put the logic in main.py

Signed-off-by: Tony Witherspoon <tony.witherspoon@gmail.com>

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -355,10 +355,14 @@ class TopLevelCommand(object):
         """
         Get help on a command.
 
-        Usage: help COMMAND
+        Usage: help [COMMAND]
         """
-        handler = get_handler(cls, options['COMMAND'])
-        raise SystemExit(getdoc(handler))
+        if options['COMMAND']:
+            subject = get_handler(cls, options['COMMAND'])
+        else:
+            subject = cls
+
+        print(getdoc(subject))
 
     def kill(self, options):
         """
--- a/tests/unit/cli_test.py
+++ b/tests/unit/cli_test.py
@@ -5,7 +5,7 @@ from __future__ import unicode_literals
 import os
 import shutil
 import tempfile
-from StringIO import StringIO
+from io import StringIO
 
 import docker
 import py
--- a/tests/unit/cli_test.py
+++ b/tests/unit/cli_test.py
@@ -5,6 +5,7 @@ from __future__ import unicode_literals
 import os
 import shutil
 import tempfile
+from StringIO import StringIO
 
 import docker
 import py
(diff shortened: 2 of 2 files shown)


aanand:
LGTM

dnephin:
Code change looks good. There is a test that expects the old behaviour which needs to be updated.

twitherspoon:
I can work on that as well.  I'll look at the cli tests, but if you point me in the right direction that will get me going quicker.

dnephin:
check out the build failure links above (the appveyor one is easier to read).

It contains the name of the failing test and filename: https://ci.appveyor.com/project/docker/compose/build/master-1704#L442

twitherspoon:
I removed the code in cli_test.py that was expecting the help command to raise SystemExit exceptions

dnephin:
Sorry, I don't think we want to remove it, just to change the expectation in the test.

twitherspoon:
Not sure how else to do this.  The original main.py code looks like this:
```
handler = get_handler(cls, options['COMMAND'])
raise SystemExit(getdoc(handler))
```

So it always throws an exception with an error code of 1

My update doesn't throw an exception and completes with error code of 0

```
        if options['COMMAND']:
            subject = get_handler(cls, options['COMMAND'])
        else:
            subject = cls

        print(getdoc(subject))
```

I can however change it to raise SystemExit and throw the exception in main.py and the cli tests will complete just fine.

Thoughts?

dnephin:
I think the behaviour of the command is correct, it's just the test that needs to be adjusted. I think the new test should continue to assert that the help text is printed, the difference is only in how it receives the help text. Instead of pulling it out of the exception it needs to capture stdout.

I believe you can do that with `mock.patch(sys.stdout):`. `pytest` also provides `capsys` , but I'm not sure how to use that as a context manager.

twitherspoon:
Updated the tests to validated against the docker-compose help up command as suggested.  Again you guys are awesome.  Thanks for pointing me in the right direction.  Learning tons.

dnephin:
LGTM

aanand:
LGTM

metal3d:
Here, the test is broken. I see that jenkins is not tested for this pull request and I cannot have a valid PR (see #3348 ans #3351 for example)

This is this PR that breaks tests.

If I revert the 984f839d330599638abfc7bdfc0754dc4f5c6d71 commit, tests are ok at home.

twitherspoon:
Is there anything I need to do?  Is there a way to see the Jenkins error?

metal3d:
I did à PR to fix tests. Writing for validation

metal3d:
See #3356

twitherspoon:
Nice.  Just checked it out.  I see what you mean now.

dnephin:
Oops, I guess janky failed to run and we didn't notice...

metal3d:
I guess that was jenkins too because my (just installed) compose dev environment didn't checked test. That's the reason my PR #3348 was retested several times (import order problems, and so on)

As soon as I finally fixed my test requirements, the truth was comed to me, I saw the light (red) and spent a long time to  understand the long test list. :)

Glade to help ;)

PR #3274, opened 2016-04-06
Relax the windows check for dockerpty

### LINKED ISSUE #3096: Improve this check for Windows users to allow interactive mode?
Take a look here: https://github.com/docker/compose/blob/1655be6c5bcbc35d2a34bd4c464b8a91981aaa42/compose/cli/main.py#L550

Is there any chance the check performed here could be refined so that if the user is running via cmder/conemu or another Windows terminal, interactive mode `run` isn't blocked?

I'm actually calling `docker-compose` from an msys2 bash terminal run via cmder.  Perhaps that's enough to satisfy what's needed for interactive mode?

dnephin:
Instead of checking the platform, attempt the import, so that users on windows that are running in a compatible terminal will be able to use interactive run.

Fixes #3096

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -20,7 +20,6 @@ from ..config import parse_environment
 from ..config.environment import Environment
 from ..config.serialize import serialize_config
 from ..const import DEFAULT_TIMEOUT
-from ..const import IS_WINDOWS_PLATFORM
 from ..progress_stream import StreamOutputError
 from ..project import NoSuchService
 from ..project import OneOffFilter
@@ -43,8 +42,12 @@ from .utils import get_version_info
 from .utils import yesno
 
 
-if not IS_WINDOWS_PLATFORM:
+try:
+    import dockerpty
     from dockerpty.pty import PseudoTerminal, RunOperation, ExecOperation
+except ImportError:
+    dockerpty = None
+
 
 log = logging.getLogger(__name__)
 console_handler = logging.StreamHandler(sys.stderr)
@@ -570,7 +573,7 @@ class TopLevelCommand(object):
         service = self.project.get_service(options['SERVICE'])
         detach = options['-d']
 
-        if IS_WINDOWS_PLATFORM and not detach:
+        if not dockerpty and not detach:
             raise UserError(
                 "Interactive mode is not yet supported on Windows.\n"
                 "Please pass the -d flag when using `docker-compose run`."


dnephin:
@atrauzzi @friism

This should allow anyone using a compatible shell on windows to do `docker-compose run`.  If you have a minute would you mind testing it out?

atrauzzi:
@dnephin - Sure thing!  Remind me again what I have to do to try out a specific branch of compose?  Any way it can done via pip?

dnephin:
This should work: `pip install git+https://github.com/dnephin/compose.git@1d75e9e67b6a497752d93be14fccad46a8ff799d#egg=compose`

atrauzzi:
Hmm... Getting this:

```
  Could not find a version that satisfies the requirement docker-py<2,>1.7.2 (from docker-compose) (from versions: 0.0.4
, 0.0.5, 0.0.6, 0.1.0, 0.1.1, 0.1.2, 0.1.3, 0.1.4, 0.1.5, 0.2.0, 0.2.1, 0.2.2, 0.2.3, 0.3.0, 0.3.1, 0.3.2, 0.4.0, 0.5.0,
 0.5.1, 0.5.2, 0.5.3, 0.6.0, 0.7.0, 0.7.1, 0.7.2, 1.0.0, 1.1.0, 1.2.1, 1.2.2, 1.2.3rc1, 1.2.3, 1.3.0, 1.3.1, 1.4.0, 1.5.
0, 1.6.0, 1.7.0rc1, 1.7.0rc2, 1.7.0rc3, 1.7.0, 1.7.1, 1.7.2, 1.8.0rc2, 1.8.0rc4, 1.8.0rc5)
No matching distribution found for docker-py<2,>1.7.2 (from docker-compose)
```

friism:
One can also grab the .exe here: https://ci.appveyor.com/project/docker/compose/build/master-1710/artifacts

I'll try soon

atrauzzi:
Ah, perfect!  Will try that right away.

atrauzzi:
That executable gave me:

```
PS C:\Users\atrau\Development\laravel-drydock> ./docker-compose run webapp ls
WARNING: The UID variable is not set. Defaulting to a blank string.
ERROR: Interactive mode is not yet supported on Windows.
Please pass the -d flag when using `docker-compose run`.
```

dnephin:
@atrauzzi with which terminal?

atrauzzi:
I'm in what's hosting PowerShell on the current insider build of Windows.  So, Microsoft's.

atrauzzi:
Same thing happens with the terminal host that ships with git bash.

`docker-compose version 1.7.0dev, build 3e070ce`

That's the executable artifact from above.

dnephin:
Ok, I guess this doesn't work. I'll look into it some more.

Thanks for testing

atrauzzi:
I'll keep an eye on this ticket and will try to get you whatever feedback you need promptly.  Thanks so much :)

rwilliams:
Doesn't this come down to Windows not having 'fcntl'? I'm pretty sure that's why the import fails.

atrauzzi:
Quick update: Just tried on the insider build that released yesterday (`14316.rs1_release.160402-2217`), same result.

Worth noting that this release contains the new "Bash on Ubuntu on Windows" subsystem, so the default console host that I'm using definitely has VT100 support now.

rwilliams:
Doesn't this come down the the fact that we're using python on Windows and it doesn't have the fnctl module so the import of dockerpty won't work?

dnephin:
> Doesn't this come down the the fact that we're using python on Windows and it doesn't have the fnctl module so the import of dockerpty won't work

Yes, I think that's right.

Yes I'm able to run Compose using babun or cygwin.

atrauzzi:
So I'm guessing this means anywhere in compose that tries to use that library is going to have to be wrapped to support some Windows-specific alternative?

rwilliams:
> So I'm guessing this means anywhere in compose that tries to use that library is going to have to be wrapped to support some Windows-specific alternative?

Here's the original issue from dockerpty https://github.com/d11wtq/dockerpty/issues/7 One of the points from tamascsaba might have some good info on writing non-blocking io that works on both linux and windows

atrauzzi:
@dnephin - Is this use case no longer being considered or is it being addressed elsewhere?

dnephin:
This PR didn't do anything so I closed it. It's still something we need to fix.  #3096 is still an open issue.

PR #3275, opened 2016-04-06
Updated Wordpress tutorial

activatedgeek:
The new tutorial now uses official Wordpress Docker Image.

### DIFF
--- a/docs/wordpress.md
+++ b/docs/wordpress.md
@@ -22,7 +22,7 @@ with Docker containers. This quick-start guide demonstrates how to use Compose t
 
     You can name the directory something easy for you to remember. This directory is the context for your application image. The directory should only contain resources to build that image.
 
-    This project directory will contain a `Dockerfile`, a `docker-compose.yaml` file, along with a downloaded `wordpress` directory and a custom `wp-config.php`, all of which you will create in the following steps.
+    This project directory will contain a `docker-compose.yaml` file which will be complete in itself for a good starter wordpress project.
 
 2. Change directories into your project directory.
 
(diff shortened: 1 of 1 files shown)


activatedgeek:
@dnephin When can I expect this to go live?

dnephin:
@activatedgeek I don't have any estimate for you unfortunately.

cc @londoncalling for docs review

docs/wordpress.md
```diff
@@ -22,116 +22,47 @@ with Docker containers. This quick-start guide demonstrates how to use Compose t
...
+            environment:
+              WORDPRESS_DB_HOST: db:3306
+              WORDPRESS_DB_PASSWORD: wordpress
+
+    ***NOTE***: You can also use the drop in replacement `MariaDB` by simply replacing with the official `mariadb:10`.
 
 ### Build the project
 
-With those four new files in place, run `docker-compose up` from your project directory. This will pull and build the needed images, and then start the web and database containers.
+With those four new files in place, run `docker-compose up -d` from your project directory. This will pull and build the needed images, and then start the wordpress and database containers.
```
londoncalling: @dnephin something seems to be missing here, because as written:
* this doesn't make sense anymore "with those four new files in place" (which 4 files? because now we don't tell them to create them)
*  it doesn't run

I see that the `docker-compose.yml` file is calling different DB and Wordpress images, defining different volumes, wordpress, etc. but how is the .yml supposed to find what it needs?  Doesn't seem to be working for me.

Also there doesn't seem to be a reference point for `MariaDB` comment. Where would I see that referenced? In one of the containers? In another config file?

londoncalling:
@dnephin @activatedgeek please see my comments inline on the markdown file; I can't get the sample app to run by following the tutorial as is now

activatedgeek:
Hey @londoncalling, I updated the misleading old text in the document. Also I just ran a fresh project, created a folder, adding the docker compose file mentioned in the document and it works. I have done absolutely no modifications whatsoever to the `wordpress` or `mysql` docker images.

I have added an extra note as the containers take a couple of minutes to bootstrap on the first load. Once the setup is complete it should work as claimed.

londoncalling:
@activatedgeek yes it works now. Thanks for updating the wording, it reads very nicely. As for it not running before, I believe that was caused by some problems I was having with some in-development versions of Docker. All good now. LGTM! @dnephin

PR #3278, opened 2016-04-07
Parallel execution: check readiness of each object individually

shin-:
Prevents edge-cases where some objects would be processed multiple times

Fixes #3271

PR #3279, opened 2016-04-07
Add locks to collections used by parallel threads

shin-:
Built on top of #3278 

I don't think we've had any reports where this was an issue, and this might be overzealous, hence it's a separate PR.


dnephin:
I think this would have resolved the immediate issues. With the changes in #3291 we don't need locks anymore because we're keeping all the state in a single thread, so I think we can close this.

PR #3291, opened 2016-04-08
Parallel execution refactor/fixes

### LINKED ISSUE #3271: When updating a container definition with links: `ERROR: ... Conflict. The name ... is already in use by container ...` (1.7 regression)
**What I do:**

```
docker-compose -f - up -d <<EOF
rainbow:
    image: redis
    links:
      - red
      - green
      - blue

red:
    image: redis

green:
    image: redis

blue:
    image: redis
EOF
docker-compose -f - up -d <<EOF
rainbow:
    image: redis
    links:
      - red
      - green
      - blue

red:
    image: redis

green:
    image: redis

blue:
    image: redis
    links:
      - cyan
      - navy

cyan:
    image: redis

navy:
    image: redis
EOF
```

**What I expect:**

Container `blue` is updated since in the 2nd version it has links to `cyan` and `navy`

**What I see instead:**

```
ERROR: for cyan  Conflict. The name "/composebug_cyan_1" is already in use by container 7841361683d7af4c53cac8c698a09a79cb16c8c21d83e9705ff3c36ed3dcc26e. You have to remove (or rename) that container to be able to reuse that name.
```

**Extra notes:**

```
$ docker-compose -v
docker-compose version 1.7.0rc1, build 1ad8866
```

This is The Weirdest Thing Ever.

If I remove a container (e.g. `red`, `green`, or `navy`) everything works fine.

If I rollback to Compose 1.6.2, all works correctly.

If I bring up the 2nd stack directly (without going through the 1st stack) all works fine.

This is executed on a single Engine, after wiping out all containers; running 1.11-rc3.

I initially had the problem on a much bigger Compose file so I reduced it to the smallest thing that would still crash.

To make it easy to tweak/test/etc, get this gist: https://gist.github.com/jpetazzo/6902757218d61ad7e2f4bc07ff40053c

I'm stumped.

aanand:
- Fixes #3271
- Fixes another bug where if A depends on B and B fails, A is still executed
- Obviates the need for #3278 and #3279 by doing all set mutation in a single thread, to avoid race conditions and locks
- Adds basic logging and unit tests for `parallel` module

### DIFF
--- a/compose/parallel.py
+++ b/compose/parallel.py
@@ -64,18 +64,30 @@ def _no_deps(x):
     return []
 
 
+class State(object):
+    def __init__(self, objects):
+        self.objects = objects
+
+        self.started = set()   # objects being processed
+        self.finished = set()  # objects which have been processed
+        self.failed = set()    # objects which either failed or whose dependencies failed
+
+    def is_done(self):
+        return len(self.finished) + len(self.failed) >= len(self.objects)
+
+    def pending(self):
+        return set(self.objects) - self.started - self.finished - self.failed
+
+
 def parallel_execute_stream(objects, func, get_deps):
     if get_deps is None:
         get_deps = _no_deps
 
     results = Queue()
+    state = State(objects)
 
-    started = set()   # objects being processed
-    finished = set()  # objects which have been processed
-    failed = set()    # objects which either failed or whose dependencies failed
-
-    while len(finished) + len(failed) < len(objects):
-        for event in feed_queue(objects, func, get_deps, results, started, finished, failed):
+    while not state.is_done():
+        for event in feed_queue(objects, func, get_deps, results, state):
             yield event
 
         try:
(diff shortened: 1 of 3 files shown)


compose/parallel.py
```diff
@@ -68,40 +74,72 @@ def _no_deps(x):
...
         get_deps = _no_deps
 
     results = Queue()
+    output = Queue()
+
+    t = Thread(target=queue_consumer, args=(objects, func, get_deps, results, output))
+    t.daemon = True
+    t.start()
+
+    return output
```
dnephin: Do we actually need the consumer to run in a separate thread?

It seems like it could be part of the main thread, which would let us remove the second queue.  The main thread would consume results (as it does now), and call `queue_consumer()` to launch any new threads that can be started based on the event.

The end of `queue_consumer()` looks very similar to what the main thread is doing now.

Would that work?

compose/parallel.py
```diff
@@ -68,40 +64,83 @@ def _no_deps(x):
...
+    pending = state.pending()
+    log.debug('Pending: {}'.format(pending))
+
+    for obj in pending:
+        deps = get_deps(obj)
+
+        if any(dep in state.failed for dep in deps):
+            log.debug('{} has upstream errors - not processing'.format(obj))
+            yield (obj, None, UpstreamError())
+            state.failed.add(obj)
```
dnephin: This could push an event into `results` instead of yielding it. I think that change would remove the need to handle the return value from `feed_queue()` (there is no more return value), and it would mean that `state.failed.add()` would only happen in a single spot.

But that's a minor point, up to you.

dnephin:
LGTM

PR #3292, opened 2016-04-08
Remove extra ensure_image_exists() which causes duplicate builds.

### LINKED ISSUE #3286: "up --build" does not use updated image if containers already exist
I have a project with containers already created but stopped. When I do `up --build`, it just starts the containers and doesn't recreate them:

```
$ docker-compose up --build
Building web
...
Removing intermediate container d2dbcf581064
Successfully built ebe1c55658cd
Starting charingcross_db_1
Starting charingcross_web_1
...
```

Running a second time builds twice, but recreates it successfully:

```
$ docker-compose up -d
Building web
...
Removing intermediate container 47aa1ff5fd24
Successfully built 2f73f4815ea1
Starting charingcross_db_1
Recreating charingcross_web_1
Building web
...
Successfully built 2f73f4815ea1
Attaching to charingcross_db_1, charingcross_web_1
...
```

Compose 1.7.0rc1 / Engine 1.11.0rc3

dnephin:
Fixes #3286

`ensure_image_exists()` now happens before creating the convergence plan, so we generate the correct plan (instead of generating "start the old container").

In the case where something has change, we are now only passing `on_build=...` from the first call to `ensure_image_exists()`, so we don't end up building it twice.

### DIFF
--- a/compose/project.py
+++ b/compose/project.py
@@ -309,12 +309,13 @@ class Project(object):
     ):
         services = self.get_services_without_duplicate(service_names, include_deps=True)
 
+        for svc in services:
+            svc.ensure_image_exists(do_build=do_build)
         plans = self._get_convergence_plans(services, strategy)
 
         for service in services:
             service.execute_convergence_plan(
                 plans[service.name],
-                do_build,
                 detached=True,
                 start=False)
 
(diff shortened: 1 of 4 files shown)


aanand:
LGTM once green

PR #3297, opened 2016-04-10
Fix showing error service name when one of the services' name is invalid

max8899:
The problem is easy to reproduce, using the yaml file below:

a-zA-Z0-9._:
  image: busybox:latest
  command: top
invalid@name:
  image: busybox:latest
  command: top

and the result will be:
Invalid service name 'a-zA-Z0-9._' - only [a-zA-Z0-9._-] characters are allowed

Signed-off-by: Lei Gong xue177125184@gmail.com

### DIFF
--- a/compose/config/validation.py
+++ b/compose/config/validation.py
@@ -13,6 +13,7 @@ from jsonschema import Draft4Validator
 from jsonschema import FormatChecker
 from jsonschema import RefResolver
 from jsonschema import ValidationError
+from jsonschema import _utils
 
 from ..const import COMPOSEFILE_V1 as V1
 from .errors import ConfigurationError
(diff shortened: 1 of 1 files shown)


compose/config/validation.py
```diff
@@ -13,6 +13,7 @@
 from jsonschema import FormatChecker
 from jsonschema import RefResolver
 from jsonschema import ValidationError
+from jsonschema import _utils
```
dnephin: I don't think we should be using private functions from `jsonschema`.

dnephin:
So the problem is that we're reading the wrong value when creating the error message, is that right?

max8899:
Yes, it is. And the suggestion is to implement an same function? I was planning to do that but found "_utils" just do exactly the same thing.

dnephin:
It's a bit risky to depend on this function, as it could change or be removed, but I guess it's not any better to re-implement it.

LGTM when tests are green

shin-:
Closing ia #4529

PR #3302, opened 2016-04-11
feed_queue() refactor

aanand:
Put the event tuple into the results queue rather than yielding it from the function, as suggested by @dnephin in https://github.com/docker/compose/pull/3291#discussion_r59075116.

Also put some docstrings in and renamed a couple of functions.

### DIFF
--- a/compose/parallel.py
+++ b/compose/parallel.py
@@ -65,12 +65,19 @@ def _no_deps(x):
 
 
 class State(object):
+    """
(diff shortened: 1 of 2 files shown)


dnephin:
Code changes look good, but there is a test failure

aanand:
Weird, I don't get that locally. There might be a race - I'll investigate.

aanand:
OK, there was. Fixed.

PR #3310, opened 2016-04-12
Attempt to fix flaky logs test

aanand:
Rather than kill the `docker-compose` process, kill the one container it's tailing logs from.

In addition, make the `hello` container hang forever rather than for 2 seconds, just to rule out the possibility of creating a race condition if there's a significant delay in attaching.


dnephin:
`sleep 200` is 200 seconds not 2 seconds, but I guess tailing `/dev/null` is fine too. I don't think that will change much.

Killing the service instead of the logs process makes sense.

LGTM

PR #3337, opened 2016-04-15
Check for short id alias, and don't disconnect if it already exists

### LINKED ISSUE #3252: Networking doesn't work with Windows containers due to disconnecting containers from networks
See @taylorb-microsoft's comment here: https://github.com/docker/compose/commit/313c584185f3999eb7a9ad0b7792a251b0afcbf8#commitcomment-16546187

Windows doesn't support disconnecting containers from networks and isn't going to any time soon. It seems like we should get better APIs for updating aliases on containers instead of doing this which feels like hack.

This is blocking Compose working with Windows containers, I believe.

dnephin:
Fixes #3252

Related to https://github.com/docker/libnetwork/issues/996 and https://github.com/docker/docker/pull/21901

Windows engine doesn't have network disconnect, so we can check if the short id alias already exists, and skip the disconnect if it does exist.

cc @friism, @mavenugo

### DIFF
--- a/compose/service.py
+++ b/compose/service.py
@@ -461,10 +461,9 @@ class Service(object):
                     container.id,
                     network)
 
-            aliases = netdefs.get('aliases', [])
             self.client.connect_container_to_network(
                 container.id, network,
-                aliases=list(self._get_aliases(container).union(aliases)),
+                aliases=self._get_aliases(netdefs, container),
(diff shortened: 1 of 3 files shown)


dnephin:
@friism when used with https://github.com/docker/docker/pull/21901 this should be working. You can grab the binary from the appveyor build when it completes.

aanand:
LGTM

PR #3340, opened 2016-04-17
Fix #2922: Config does not catch missing links

### LINKED ISSUE #2922: config does not catch missing links
When a compose file describes a service _A_ that `links` to a service _B_ which is not described, the `up` command fails, yet the `config` command raises no concerns.

In the absence of a better command, `config` might be expected to reveal such problems ahead of the user invoking any further compose commands preventing possible downtime.

johnharris85:
Fixes #2922

### DIFF
--- a/compose/config/config.py
+++ b/compose/config/config.py
@@ -37,6 +37,7 @@ from .validation import validate_against_config_schema
 from .validation import validate_config_section
 from .validation import validate_depends_on
 from .validation import validate_extends_file_path
+from .validation import validate_links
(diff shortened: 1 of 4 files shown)


johnharris85:
Implementation is cloned from the checks for `depends_on`. Caused an issue with the `extends` test for extending services with links. The compose file fixture for that test had an invalid link so I 'fixed' it to only fail the once.

aanand:
Thanks!

LGTM

dnephin:
LGTM

PR #3348, opened 2016-04-19
Add command interpolation

metal3d:
Use $((cmd)) to be able to set a value from a shell command.

See #1377
See #3354

### DIFF
--- a/compose/config/interpolation.py
+++ b/compose/config/interpolation.py
@@ -2,12 +2,17 @@ from __future__ import absolute_import
 from __future__ import unicode_literals
 
 import logging
+import re
 from string import Template
+from subprocess import CalledProcessError
+from subprocess import check_output
 
 import six
 
 from .errors import ConfigurationError
+
 log = logging.getLogger(__name__)
+command_regexp = re.compile('\$\(\((.*?)\)\)')
 
 
 def interpolate_environment_variables(config, section, environment):
@@ -55,7 +60,31 @@ def interpolate(string, mapping):
     try:
         return Template(string).substitute(mapping)
     except ValueError:
+        has_command = False
+        for matches in command_regexp.finditer(string):
+            has_command = True
+            string = interpolate_command(string, matches.group(1), matches.group(0))
+        if has_command:
+            return interpolate(string, mapping)
+        else:
+            raise InvalidInterpolation(string)
+
+
+def interpolate_command(string, command, base):
+    try:
+        m = check_output(command, shell=True).strip()
+        string = string.replace(base, m.decode())
+    except CalledProcessError as e:
+        raise ConfigurationError(
+            '"{cmd}" command returned {ret} exit status.\n'
+            'Command output is:\n'
+            '{out}\n'.format(
+                cmd=e.cmd,
+                ret=e.returncode,
+                out=e.output.decode()))
+    except Exception as e:
         raise InvalidInterpolation(string)
+    return string
 
 
 class InvalidInterpolation(Exception):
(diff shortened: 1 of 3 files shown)


metal3d:
Can someone tell how can I locally test the "reorderImports" ?

PS: I add documentation in next commit when this test will be success

metal3d:
I didn't changed this files:

```
XFAIL tests/unit/config/config_test.py::VolumePathTest::test_split_path_mapping_with_windows_path
  does not have a drive

XFAIL tests/unit/config/types_test.py::TestVolumeSpec::()::test_parse_volume_windows_absolute_path
  does not have a drive
```

Help ?

dnephin:
Thanks for the PR!

Unfortunately I don't think this is something that we want to support. Adding this feature would mean that someone could run absolutely anything from a Compose file. I think in generally it is a very bad idea for a template to be able to execute code.

It's much better when the full context for a template is rendered ahead of time, and then applied to the template. This is what we have now with environment variables. A user can run these commands in a script and either generate a `.env` file or `export` the values so they are available to Compose.

Admittedly this is not as nice as it could be. It requires users to either run bash scripts or some other general purpose build tool (like `make`) which might be fine for some users, but not for others.

If we want to support execution of arbitrary commands on the host we would need a more explicit way to do so. In the past we've rejected other features (like "hooks") because executing tasks on the host is not something we really want to support for Compose.

bfirsh:
+1

Thanks for the contribution, but as @dnephin said, unfortunately we want to avoid anything in Compose which would make it not portable. If somebody writes a Compose file which expects a certain command to be available on the host, `docker-compose up` won't work on somebody else's computer.

If this could run inside a container somehow, then this might be a different matter... :D

metal3d:
I understand your opinion but I don't agree. :)

> I think in generally it is a very bad idea for a template to be able to execute code.

As far as I know, "compose" file is more a "Makefile" than a "template". Seeing #1377 we are several users thinking that a compose file should allow to get a "command output" to set some values.

And many template engines allow to run some commands (restricted somtimes) to get values (PHP, Bash, and so on) - that doesn't mean that template engines are bad, or you are wrong. That only mean that can be useful and often necessary. In #1377 you can see that many users need to get environment configuration to be set in compose file, and sometimes it's a pain as "having user and group ids for the current user"

A Makefile can execute some "bad command" and a compose file could mount volumes to remove the entire home directory (for example). So, in the same idea, volumes are dangerous.

Also I guess that the compatibility of command inside `$(())` is the same as environment variables interpolation. I've got some variables in my environment that are not set everywhere.

Compose files are generally used by developers (AFAIK) and as I mentioned in the given issue, to give an example, my team is not a sysadmin team, but a developers team. 

Asking them to set variables as this is very complicated:

```bash
GUID=$(id -g) UUID=$(id -u) docker-compose build
GUID=$(id -g) UUID=$(id -u) docker-compose up
# and so on
```

Yes, I made a Makefile to help us to run docker-compose on their computers, but that generates 2 files to maintain, only to get 2 values from the shell.

My PR allows that:

```yaml
#...
services:
    container:
        image: alpine
        build:
            context: docker/
            args:
                 GUID: $((id -g))
                 UUID: $((id -u))
        user: $((id -u)):$((id -g))
```

I admit that the problem is Windows. I cannot be sure that it works on that OS at this time (I've got no Windows here)

To be honest, my idea was born only for "id" command. 

Maybe given the UID and GID of current user or compose file owner as "special environment variable" can be a better idea. I can propose other PR for that.

Maybe we may restrict command list to "id", "sed" (without "-i", and no pipelines, for example by removing the "shell=True" from the `check_output()` function invocation.

What do you think ?

metal3d:
> A user can run these commands in a script and either generate a .env file or export the values so they are available to Compose.

Right, but that doesn't works with "build" directive, or maybe there is a bug.

metal3d:
See #3367 please

PR #3359, opened 2016-04-20
Clarify service networks documentation

bfirsh:
When jumping straight to this bit of the docs, it's not clear
that these are options under a service rather than the top-level
`networks` key. Added a service to make this super clear.


dnephin:
LGTM

PR #3360, opened 2016-04-20
Make validation errors less robotic

bfirsh:
```
Additional properties are not allowed ('dfjklhfsd' was unexpected)
Either specify a version of "2" (or "2.0") and place your service definitions under the `services` key, or omit the `version` key and place your service definitions at the root of the file to use version 1.
For more on the Compose file format versions, see https://docs.docker.com/compose/compose-file/
```

is now

```
Additional properties are not allowed ('dfjklhfsd' was unexpected)

You might be seeing this error because you're using the wrong Compose file version. Either specify a version of "2" (or "2.0") and place your service definitions under the `services` key, or omit the `version` key and place your service definitions at the root of the file to use version 1.
For more on the Compose file format versions, see https://docs.docker.com/compose/compose-file/
```

Hopefully this will help meatbag(s) parse this.

### DIFF
--- a/compose/config/errors.py
+++ b/compose/config/errors.py
@@ -3,10 +3,11 @@ from __future__ import unicode_literals
 
 
 VERSION_EXPLANATION = (
-    'Either specify a version of "2" (or "2.0") and place your service '
-    'definitions under the `services` key, or omit the `version` key and place '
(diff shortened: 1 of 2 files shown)


dnephin:
LGTM

PR #3364, opened 2016-04-21
service: detailed error messages for create and start

### LINKED ISSUE #3355: Include service name when printing an error message
We've been happily using docker compose for a couple of months in our project (~9 service, moderately complex setup). From time to time, usually when a new team member tries to set up the development environment, there's an issue with bringing up the environment. When this happens, all we get is some cryptic error message:
 * `No such file or directory`
 * `Container command not found or does not exist.`

At this point, we usually have to dig into the person's machine and check what's wrong. We usually have to: check container logs, check images, inspect containers, try to start them manually using `docker` directly. From time to time, we even have to check docker logs, start stuff with `strace`...

It would be so much better if compose printed way more information when an error occurs: service name, engine API call, all arguments... Ideally, engine would do the same by sending more detailed error messages.

TomasTomecek:
Fixes #3355 

this should help in situations like these: https://github.com/docker/docker/issues/18098#issuecomment-165518262

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -32,6 +32,7 @@ from ..service import BuildError
 from ..service import ConvergenceStrategy
 from ..service import ImageType
 from ..service import NeedsBuildError
+from ..service import OperationFailedError
 from .command import get_config_from_options
 from .command import project_from_options
 from .docopt_command import DocoptDispatcher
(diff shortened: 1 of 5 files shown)


dnephin:
LGTM, going to rebuild to see if test failures are flakes

TomasTomecek:
I was pretty naive thinking that I can change exception type and get away with it.

Fixed test failures.

dnephin:
LGTM

TomasTomecek:
rebased

aanand:
LGTM

PR #3373, opened 2016-04-22
Unset env vars behavior in 'run' mirroring engine

### LINKED ISSUE #3343: Passing host env var to run results in None value within container
If you `run` a container and pass an env var from the host just like you'd do with `docker run -e ENV_VAR` the env var gets set to `None` inside the container that's started.
Example commands:
```
export ENV_VAR='foo`
docker-compose run -e ENV_VAR <servicename>
```

IMHO the expected behaviour is that it works the same as [`docker run`](https://docs.docker.com/engine/reference/run/#env-environment-variables), i.e. if you only pass the host's env var name it's value should be set inside the container.

I couldn't find an issue or PR for this, but it's difficult to search for `env var` and `run`, so I might have missed it.
This is somewhat related to #363

shin-:
Unset env vars passed to `run` via command line options take the value of the system's var with the same name.

Fixes #3343

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -821,7 +821,9 @@ def build_container_options(options, detach, command):
     }
 
     if options['-e']:
-        container_options['environment'] = parse_environment(options['-e'])
+        container_options['environment'] = Environment.from_command_line(
+            parse_environment(options['-e'])
+        )
 
     if options['--entrypoint']:
         container_options['entrypoint'] = options.get('--entrypoint')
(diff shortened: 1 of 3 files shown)


shin-:
I've been trying to fix the test in python 3 but to no avail. No idea why I can't get any output.

dnephin:
LGTM when build is green.

I put this into 1.8.0, but I guess it could go into 1.7.1 as well, if we want

aanand:
Superseded by https://github.com/docker/compose/pull/3670

PR #3376, opened 2016-04-25
Support SOCKS proxy from HTTP[S]_PROXY env vars

brettdh:
Currently, this has to be done by monkey-patching `socket.socket`,
using the [PySocks](https://github.com/Anorov/PySocks) library to create a replacement that checks for
`ALL_PROXY` and routes connections through the proxy if present.

The hopefully-soon release of [Requests 2.10.0](https://github.com/kennethreitz/requests/pull/2953) with [this
update](https://github.com/shazow/urllib3/pull/762) to [urllib3](https://github.com/shazow/urllib3/blob/master/CHANGES.rst#114-2015-12-29) (assuming its adoption in
docker-py), should make this unnecessary, but I'm not sure a) what's
been keeping that from happening; or b) whether docker-compose would
be able to quickly adopt the brand-new version of Requests (it's
currently a few minor versions behind), so this seems a reasonable
stopgap.


brettdh:
Requests 2.10.0 is out now, so I'm going to test this soon with the version bump (and minus the monkey patch) to see if we can just bump the version instead.

brettdh:
Tested successfully without the monkeypatching. WFM with requests[socks]>=2.10.0

brettdh:
\*facepalm\*

Sorry for the Gordon noise; thought clicking "Update branch" would prompt me before doing silly things like adding an unsigned merge commit. Resolving with a force push shortly.

shin-:
Closing via #4990

PR #3384, opened 2016-04-26
Support combination of shorthand flag and equal sign for host option

### LINKED ISSUE #3264: Add standard Docker CLI flags for remote Daemon config?
It'd be pretty nice to be able to run commands like you can with the regular docker client, so I don't have to always inject stuff into the shell environment. For example:

    $ docker-compose $(docker-machine config default) up -d web

shin-:
Fixes #3264


dnephin:
Is there no way to fix this by changing the docopt doc?

Something like `--host=HOST` ?

Issue #3316, opened 2016-04-13
"KeyError: u'status'" with 1.7.0rc1 and swarm


jamshid:
I just set up a two-node docker swarm and get the below error when running "docker-compose up" on the below test.yml (it just verifies some services are available to docker).

Those url's are fine and curl's succeed. I don't know why docker-compose is reporting KeyError. 

[verbose.txt](https://github.com/docker/compose/files/216382/verbose.txt)
```
...
Exception in thread Thread-3:
Traceback (most recent call last):
  File "threading.py", line 810, in __bootstrap_inner
  File "threading.py", line 763, in run
  File "compose/cli/log_printer.py", line 190, in watch_events
  File "compose/project.py", line 344, in events
KeyError: u'status'

buildenv_test_1 exited with code 0
```

```
version: '2'

services:
  test:
    image: centos
    command: bash -e -c "set -o pipefail ; cat /etc/resolv.conf ;
              for url in
                  http://${DOCKER_BUILD_CACHE_HOST}:3142/acng-report.html
                  http://${DOCKER_BUILD_CACHE_HOST}:8181/starwars
                  http://${DOCKER_BUILD_CACHE_HOST}:8123
                  http://${DOCKER_BUILD_CACHE_HOST}:8081
                  http://${DOCKER_BUILD_CACHE_HOST}:8181
                  https://bitbucket.org
                  https://google.com
                  ; do
                  echo Verfiying $${url} ;
                  curl -I -fsS $${url} | grep -i -E 'HTTP/|Server:' | sed 's/^/^   /g' ||
                       { echo FAILED ; exit 1 ; } ;
              done ;
              echo 'Verifying curl dl.fedoraproject.org' ;
              curl -I -fsSL http://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm | grep -i -E 'HTTP/|Server:' | sed 's/^/^   /g' || { echo FAILED ; exit 1 ; } ;
              echo 'Verifying curl -x dl.fedoraproject.org' ;
              curl -I -fsSL -x 'http://${DOCKER_BUILD_CACHE_HOST}:8123' -H Cache-control:no-cache http://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm | grep -i -E 'HTTP/|Server:' | sed 's/^/^   /g' || { echo FAILED ; exit 1 ; } ;
              echo SUCCESS"
```
```
$ docker info
Containers: 32
 Running: 13
 Paused: 0
[...]

PR #3385, opened 2016-04-26
Skip event objects that don't contain a status field

shin-:
Workaround for #3316


dnephin:
An inline comment referring back to the original issue would be great, so we know why this workaround exists.

Otherwise LGTM

Issue #3361, opened 2016-04-21
[1.7] Logs do not get filtered by service properly when new containers start


JacobHayes:
When monitoring logs from a service (`docker-compose logs -f service1`), container starts from a different service will be shown in the log output. The feature in 1.7 to add new containers to exiting log outputs is nice, but should still be limited to containers of the same service when one is specified (ie: only when the scale changes).

To recreate:
```yaml
version: '2'
services:
    service1:
        image: alpine
        command: sh -c 'while true; do echo "service1"; sleep 1; done;'
    service2:
        image: alpine
        command: sh -c 'while true; do echo "service2"; sleep 1; done;'
```

```
docker-compose up -d service1 && \
eval '(sleep 5 && docker-compose up -d service2) &' && \
docker-compose logs -f service1
```

After five seconds, you will start to see output from both services, even though we just requested logs for `service1`.

dnephin:
Not sure why this is broken. We should be handling this because we filter events to just those that match the services names. So there should be no way to start watching a container for a service that isn't listed.

aronahl:
LogPrinter's watch_events method starts watching events for any unknown container, regardless of name.

This seems to fix it (only cursory testing):

```diff
diff --git a/compose/cli/log_printer.py b/compose/cli/log_printer.py
index b48462f..90b3c6c 100644
--- a/compose/cli/log_printer.py
+++ b/compose/cli/log_printer.py
@@ -62,13 +62,15 @@ class LogPrinter(object):
                  event_stream,
                  output=sys.stdout,
                  cascade_stop=False,
-                 log_args=None):
+                 log_args=None,
+                 services=None):
         self.containers = containers
         self.presenters = presenters
         self.event_stream = event_stream
         self.output = utils.get_output_stream(output)
         self.cascade_stop = cascade_stop
         self.log_args = log_args or {}
+        self.services = services
 
     def run(self):
         if not self.containers:
@@ -81,7 +83,8 @@ class LogPrinter(object):
             thread_map,
             self.event_stream,
             self.presenters,
-            thread_args))
+            thread_args,
+            self.services))
 
         for line in consume_queue(queue, self.cascade_stop):
             remove_stopped_threads(thread_map)
@@ -186,7 +189,7 @@ def start_producer_thread(thread_args):
     producer.start()
 
 
-def watch_events(thread_map, event_stream, presenters, thread_args):
+def watch_events(thread_map, event_stream, presenters, thread_args, services):
     for event in event_stream:
[...]

PR #3390, opened 2016-04-27
Add missing log event filter when using docker-compose logs.

aronahl:
Proposed fix for #3361


dnephin:
Code change looks good, thanks!

You'll just need to squash your commits and make sure the one commit is signed.

shin-:
LGTM, thanks!

Issue #3381, opened 2016-04-26
".env" file not read relative to docker-compose.yml file


andrerom:
[global/default ".env" file](https://docs.docker.com/compose/env-file/) is a great feature, allowing to set defaults for env variables without hardcoding reference to it inside each and every service definition using `env_file`, or having to wrap docker-compose in script to accomplish this.

However it does not seem to take folder of `docker-compose.yml` file into account, as implied by the doc:
> Compose supports declaring default environment variables in an environment file named .env, *and placed in the same folder as your compose file*.

#### Example:
```
docker-compose --verbose -f bin/.docker/docker-compose.yml up -d --no-recreate
compose.config.config.find: Using configuration files: ./bin/.docker/docker-compose.yml
WARNING: compose.config.environment.__getitem__: The DATABASE_USER variable is not set. Defaulting to a blank string.
WARNING: compose.config.environment.__getitem__: The DATABASE_PASSWORD variable is not set. Defaulting to a blank string.
WARNING: compose.config.environment.__getitem__: The DATABASE_NAME variable is not set. Defaulting to a blank string.
docker.auth.auth.load_config: File doesn't exist
compose.cli.command.get_client: docker-compose version 1.7.0, build 0d7bf73
```

### Suggested behavior
Maybe it should look for it in following order *(all before loading any docker-compose.yml files)*: 
- `pwd` *(execution folder)*
- folder of 1st docker-compose.yml
- [ ... ]

This would allow dev envimrnets to override prod env in example below with `.env` file in `dev/` folder:
```bash
 export COMPOSE_FILE="docker-compose.yml:dev/docker-compose.yml"
docker-compose up -d --no-recreate
```

andrerom:
Discussions with @dnephin revealed that this is most likely a doc issue, as it also allows setting `COMPOSE_FILE` implying it by design can only be taken from current working directory.

dnephin:
#2594 would provide one way to set the directory which contains the `.env` file

andrerom:
True, I was about to suggest a flag to override which env to use, but this would solve it as well.

cinemast:
In my opinion, changing the documentation does nit fix this issue. If you are combining multiple containers with the `-f` option, there is no way to use their local .env files.

Beside that, the documentation at https://docs.docker.com/compose/compose-file/ still promotes a different (and in my opinion more useful) behavior.

mdebord1:
[...]

PR #3399, opened 2016-04-29
Adds additional validation to 'env_vars_from_file'.

yorkedork:
The 'env_file' directive and feature precludes the use of the name '.env' in the path shared with 'docker-config.yml', regardless of whether or not it is enabled (or desired).

This change adds an additional validation to allow the use of this path provided it is not a file.

A real-world example: by convention, a project I contribute to generates a virtualenv at project-level named '.env'. The presence of this path results in immediate failure of a docker-compose invocation.

To address this issue more generally (and to cover more use cases), I would suggest that this implicit path '.env', relative to the current working directory (see #3381), be configurable, e.g.:

```
    @classmethod
    def from_env_file(cls, base_dir):
        def _initialize():
            result = cls()
            if base_dir is None:
                return result
-             env_file_path = os.path.join(base_dir, '.env')
+             env_file_path = os.path.join(base_dir, ENV_FILE_PATH)
            try:
                return cls(env_vars_from_file(env_file_path))
            except ConfigurationError:
                pass
            return result
        instance = _initialize()
        instance.update(os.environ)
        return instance
```

and see no reason why it should be hard-coded.

### DIFF
--- a/compose/config/environment.py
+++ b/compose/config/environment.py
@@ -28,6 +28,8 @@ def env_vars_from_file(filename):
     """
     if not os.path.exists(filename):
         raise ConfigurationError("Couldn't find env file: %s" % filename)
+    elif not os.path.isfile(filename):
+        raise ConfigurationError("%s is not a file." % (filename))
     env = {}
     for line in codecs.open(filename, 'r', 'utf-8'):
         line = line.strip()


andrefreitas:
+1 for this, a lot of people use virtualenv in .env path. I was making a PR for the same purpose.

shin-:
Thanks! LGTM.

ykifle:
+1 Always love it when I find an problem and then find that someone has already made a PR to fix it.

dnephin:
LGTM

PR #3400, opened 2016-04-30
Properly handle APIError failures in Project.up

### LINKED ISSUE #3333: Compose trying to attach to dead container after error reported?
Just noticed that there is an error from the `compose up` command but still says that it's trying to attach.  What's it trying to attach to here? 

    docker-compose up --build
    Recreating poc_1
    
    ERROR: for poc  Container command './entrypoint.sh' not found or does not exist.
    Attaching to

Might be just erroneous output.

shin-:
Fixes #3333

### DIFF
--- a/compose/project.py
+++ b/compose/project.py
@@ -390,13 +390,18 @@ class Project(object):
         def get_deps(service):
             return {self.get_service(dep) for dep in service.get_dependency_names()}
 
-        results = parallel.parallel_execute(
+        results, errors = parallel.parallel_execute(
             services,
             do,
             operator.attrgetter('name'),
             None,
             get_deps
         )
+        if errors:
+            raise ProjectError(
+                'Encountered errors while bringing up the project.'
+            )
+
         return [
             container
             for svc_containers in results
@@ -531,3 +536,7 @@ class NoSuchService(Exception):
 
     def __str__(self):
         return self.msg
+
+
+class ProjectError(Exception):
+    pass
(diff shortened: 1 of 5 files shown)


compose/project.py
```diff
@@ -390,13 +390,16 @@ def do(service):
...
-        results = parallel.parallel_execute(
+        results, errors = parallel.parallel_execute(
             services,
             do,
             operator.attrgetter('name'),
             None,
             get_deps
         )
+        if errors:
+            raise ProjectError('Errors encountered during up phase.')
```
dnephin: Maybe something like "Encountered errors while bringing up the project." ?

dnephin:
I think this is the right idea, just one suggestion on wording.

shin-:
Thanks! Updated the error message.

dnephin:
LGTM

compose/project.py
```diff
@@ -390,13 +390,18 @@ def do(service):
...
             services,
             do,
             operator.attrgetter('name'),
             None,
             get_deps
         )
+        if errors:
+            raise ProjectError(
+                'Encountered errors while bringing up the project.'
+            )
```
kojiromike: Unfortunately this doesn't really let us know which services errored out.

dnephin: I don't think that was the intent of this change. There was a different PR for that issue.

shin-: The detailed error with service information is reported during the `parallel_execute` call ( https://github.com/docker/compose/blob/master/compose/parallel.py#L57 )

PR #3356, opened 2016-04-20
Fix the tests from jenkins

metal3d:
Acceptance tests didn't set "help" command to return "0" EXIT_CODE.

close #3354
related #3263

Signed-off-by: Patrice Ferlet <metal3d@gmail.com>

PR #3418, opened 2016-05-04
Bump 1.7.1

shin-:


dnephin:
I think #3356 will get the tests passing, if you want to cherry-pick that into the branch as well

CHANGELOG.md
```diff
@@ -1,6 +1,50 @@
...
+  would not apply for newly created services.
+
+- Fixed a bug where unchanged services would sometimes be recreated in
+  in the up phase when using Compose with Python 3.
+
+- Fixed an issue where API errors encountered during the up phase would
+  not be recognized as a failure state by Compose.
+
+- Fixed an issue with Windows daemons where compose would try to disconnect
+  containers from their networks despite that feature not being available.
```
dnephin: It turns out this didn't actually fix the issue because the windows daemon doesn't use the same network driver.  We should probably drop this from the release notes.  Sorry I didn't communicate this earlier.

dnephin:
LGTM when green

PR #3439, opened 2016-05-10
Fixes #1135 - Use a new version of pyinstaller 3.2 to build a valid E…

### LINKED ISSUE #1135: Cannot open self /usr/bin/docker-compose or archive /usr/bin/docker-compose.pkg

garthy:
…LF file to stop prelinking breaking compose

### DIFF
--- a/requirements-build.txt
+++ b/requirements-build.txt
@@ -1,1 +1,1 @@
-pyinstaller==3.1.1
+pyinstaller==3.2


dnephin:
If you sign your commit and the build is green then we can get this into 1.8

dnephin:
Although it looks like the windows binary is broken with this change.

garthy:
sorry can't test on windows so didn't catch this.

This failed due to https://github.com/pyinstaller/pyinstaller/issues/1976 Looks like it will be fixed in a later version.

But it looks like there is a strange issue with github reporting tags that don't exist in the docker/docker repo causing a test script to fail

dnephin:
We've got #3558 open to use 3.2.1 once it gets released so going to close this PR.

PR #3449, opened 2016-05-11
Fix #3281: Unexpected result when using build args with default values

### LINKED ISSUE #3281: Unexpected result when using build args with default values
# Scenario:
## Files
`Dockerfile`:
```
FROM ubuntu:14.04

ARG FOO=1
RUN echo "-${FOO}-"

CMD /bin/bash
```

`docker-compose.yml`:
```
version: '2'

services:
  test:
    build:
      context: .
      args:
        - FOO
```

## Execution:
```
$ ./docker-compose-1.6.2 --verbose config
compose.config.config.find: Using configuration files: ./docker-compose.yml
networks: {}
services:
  test:
    build:
      args:
        FOO: None
      context: /home/riccardi/git/ses-docker/test-default-build-arg
version: '2.0'
volumes: {}

```

```
$ ./docker-compose-1.6.2 --verbose build
compose.config.config.find: Using configuration files: ./docker-compose.yml
docker.auth.auth.load_config: File doesn't exist
compose.cli.command.get_client: docker-compose version 1.6.2, build 4d72027
docker-py version: 1.7.2
CPython version: 2.7.9
OpenSSL version: OpenSSL 1.0.1e 11 Feb 2013
compose.cli.command.get_client: Docker base_url: http+docker://localunixsocket
compose.cli.command.get_client: Docker version: KernelVersion=4.2.0-35-generic, Os=linux, BuildTime=2016-03-10T15:54:52.312835708+00:00, ApiVersion=1.22, Version=1.10.3, GitCommit=20f81dd, Arch=amd64, GoVersion=go1.5.3
compose.service.build: Building test
compose.cli.verbose_proxy.proxy_callable: docker build <- (pull=False, stream=True, nocache=False, tag=u'testdefaultbuildarg_test', buildargs={u'FOO': 'None'}, rm=True, forcerm=False, path='/home/riccardi/git/ses-docker/test-default-build-arg', dockerfile=None)
docker.api.build._set_auth_headers: Looking for auth config
docker.api.build._set_auth_headers: No auth config in memory - loading from filesystem
docker.auth.auth.load_config: File doesn't exist
docker.api.build._set_auth_headers: No auth config found
compose.cli.verbose_proxy.proxy_callable: docker build -> <generator object _stream_helper at 0x7f56bafb3a50>
Step 1 : FROM ubuntu:14.04
 ---> b549a9959a66
Step 2 : ARG FOO=1
 ---> Using cache
 ---> 4774113d6ec5
Step 3 : RUN echo "-${FOO}-"
 ---> Running in dabd31837074
-None-
 ---> f8a99349af3b
Removing intermediate container dabd31837074
Step 4 : CMD /bin/bash
 ---> Running in 487f5e789c38
 ---> 6c484f426fb5
Removing intermediate container 487f5e789c38
Successfully built 6c484f426fb5
compose.cli.verbose_proxy.proxy_callable: docker close <- ()
compose.cli.verbose_proxy.proxy_callable: docker close -> None
```
( same result with 1.7.1-rc1 which includes PR #2938 )

# Issue
## Expected result:
prints `-1-`.

## Actual result:
prints `-None-`.

## Details:
Compose has no value for  `FOO` build arg from its environment, so it could either send an empty string to `docker build`, or better: not send this build arg to `docker build`.
The second one would be great: it would open the possibility to use the default value for the build arg as defined in the `Dockerfile`. ( For now the workaround is to duplicate the default values from `Dockerfile` to `.env`, only works with >=1.7.0).
The first one would be still better than the current behavior.

Current behavior: no value in Compose environment is represented in python as `None`, then casted to a string `"None"`, which is probably always bad.

Andrey9kin:
Fix the issue when build arg is set to None instead of empty string.
Usecase:

cat docker-compose.yml
....
    args:
      - http_proxy
      - https_proxy
      - no_proxy

If http_proxy, https_proxy, no_proxy environment variables are not defined
then http_proxy, https_proxy, no_proxy build args will be set to None which breaks all downloads

With this change build args will be set to empty string if corresponding env
variable is not defined

Signed-off-by: Andrey Devyatkin <andrey.a.devyatkin@gmail.com>

### DIFF
--- a/compose/service.py
+++ b/compose/service.py
@@ -701,12 +701,6 @@ class Service(object):
 
         build_opts = self.options.get('build', {})
         path = build_opts.get('context')
-        # If build argument is not defined and there is no environment variable
-        # with the same name then build argument value will be None
-        # Moreover it will be sent to the docker engine as None and then
-        # interpreted as string None which in many cases will fail the build
-        # That is why we filter out all pairs with value equal to None
-        buildargs = {k: v for k, v in build_opts.get('args', {}).items() if v != 'None'}
         # python2 os.path() doesn't support unicode, so we need to encode it to
         # a byte string
         if not six.PY3:
@@ -721,7 +715,7 @@ class Service(object):
             pull=pull,
             nocache=no_cache,
             dockerfile=build_opts.get('dockerfile', None),
-            buildargs=buildargs,
+            buildargs=build_opts.get('args', None),
         )
 
         try:
--- a/compose/service.py
+++ b/compose/service.py
@@ -701,6 +701,12 @@ class Service(object):
 
         build_opts = self.options.get('build', {})
         path = build_opts.get('context')
+        # If build argument is not defined and there is no environment variable
+        # with the same name then build argument value will be None
+        # Moreover it will be sent to the docker engine as None and then
+        # interpreted as string None which in many cases will fail the build
+        # That is why we filter out all pairs with value equal to None
+        buildargs = {k: v for k, v in build_opts.get('args', {}).items() if v != 'None'}
         # python2 os.path() doesn't support unicode, so we need to encode it to
         # a byte string
         if not six.PY3:
@@ -715,7 +721,7 @@ class Service(object):
             pull=pull,
             nocache=no_cache,
             dockerfile=build_opts.get('dockerfile', None),
-            buildargs=build_opts.get('args', None),
+            buildargs=buildargs,
         )
 
         try:
--- a/compose/utils.py
+++ b/compose/utils.py
@@ -95,4 +95,4 @@ def microseconds_from_time_nano(time_nano):
 
 
 def build_string_dict(source_dict):
-    return dict((k, str(v if v else '')) for k, v in source_dict.items())
+    return dict((k, str(v if v is not None else '')) for k, v in source_dict.items())
--- a/compose/utils.py
+++ b/compose/utils.py
@@ -95,4 +95,4 @@ def microseconds_from_time_nano(time_nano):
 
 
 def build_string_dict(source_dict):
-    return dict((k, str(v)) for k, v in source_dict.items())
+    return dict((k, str(v if v else '')) for k, v in source_dict.items())
--- a/tests/unit/config/config_test.py
+++ b/tests/unit/config/config_test.py
@@ -715,7 +715,7 @@ class ConfigTest(unittest.TestCase):
         ).services[0]
         assert 'args' in service['build']
         assert 'foo' in service['build']['args']
-        assert service['build']['args']['foo'] == 'None'
+        assert service['build']['args']['foo'] == ''
 
     def test_load_with_multiple_files_mismatched_networks_format(self):
         base_file = config.ConfigFile(
--- a/tests/unit/config/config_test.py
+++ b/tests/unit/config/config_test.py
@@ -717,6 +717,34 @@ class ConfigTest(unittest.TestCase):
         assert 'foo' in service['build']['args']
         assert service['build']['args']['foo'] == ''
 
+    # If build argument is None then it will be converted to the empty
+    # string. Make sure that int zero kept as it is, i.e. not converted to
+    # the empty string
+    def test_build_args_check_zero_preserved(self):
+        service = config.load(
+            build_config_details(
+                {
+                    'version': '2',
+                    'services': {
+                        'web': {
+                            'build': {
+                                'context': '.',
+                                'dockerfile': 'Dockerfile-alt',
+                                'args': {
+                                    'foo': 0
+                                }
+                            }
+                        }
+                    }
+                },
+                'tests/fixtures/extends',
+                'filename.yml'
+            )
+        ).services[0]
+        assert 'args' in service['build']
+        assert 'foo' in service['build']['args']
+        assert service['build']['args']['foo'] == '0'
+
     def test_load_with_multiple_files_mismatched_networks_format(self):
         base_file = config.ConfigFile(
             'base.yaml',
--- a/tests/unit/service_test.py
+++ b/tests/unit/service_test.py
@@ -445,7 +445,7 @@ class ServiceTest(unittest.TestCase):
             forcerm=False,
             nocache=False,
             rm=True,
-            buildargs={},
+            buildargs=None,
         )
 
     def test_ensure_image_exists_no_build(self):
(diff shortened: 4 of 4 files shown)


tests/unit/config/config_test.py
```diff
@@ -2121,7 +2121,7 @@ def test_resolve_environment_from_env_file_with_empty_values(self):
                 'FILE_DEF': u'bär',
                 'FILE_DEF_EMPTY': '',
                 'ENV_DEF': 'E3',
-                'NO_DEF': None
+                'NO_DEF': ''
```
dnephin: This change is wrong, the `None` is the correct value here, so I think the code change isn't correct to fix the issue.

tests/unit/config/config_test.py
```diff
@@ -2140,7 +2140,7 @@ def test_resolve_build_args(self):
         }
         self.assertEqual(
             resolve_build_args(build, Environment.from_env_file(build['context'])),
-            {'arg1': 'value1', 'empty_arg': '', 'env_arg': 'value2', 'no_env': None},
+            {'arg1': 'value1', 'empty_arg': '', 'env_arg': 'value2', 'no_env': ''},
```
dnephin: Same here, the old test is correct.

tests/unit/config/config_test.py
```diff
@@ -2077,7 +2077,7 @@ def test_resolve_environment(self):
             resolve_environment(
                 service_dict, Environment.from_env_file(None)
             ),
-            {'FILE_DEF': 'F1', 'FILE_DEF_EMPTY': '', 'ENV_DEF': 'E3', 'NO_DEF': None},
+            {'FILE_DEF': 'F1', 'FILE_DEF_EMPTY': '', 'ENV_DEF': 'E3', 'NO_DEF': ''},
```
dnephin: This value should be `None` as well.

compose/config/config.py
```diff
@@ -846,7 +846,7 @@ def resolve_env_var(key, val, environment):
     elif environment and key in environment:
         return key, environment[key]
     else:
-        return key, None
+        return key, ''
```
dnephin: I think this needs to stay `None`. The `None` needs to be handled somewhere else.

aanand:
@Andrey9kin Are you up for making the changes @dnephin suggested? We'd like to get this in for 1.8.0!

Andrey9kin:
@aanand last few weeks was quite intense so I didn't have time to submit a fix.
I'm planning to submit a patch before end of this week

Andrey9kin:
@aanand @dnephin new attempt - please take a look. Jenkins seems to fail due to environment flakiness which is unrelated to the change. Can I retrigger Jenkins build somehow?

I'm new to this code base so please advice if there is a better place to filter out those None's

compose/service.py
```diff
@@ -701,6 +701,12 @@ def build(self, no_cache=False, pull=False, force_rm=False):
 
         build_opts = self.options.get('build', {})
         path = build_opts.get('context')
+        # If build argument is not defined and there is no environment variable
+        # with the same name then build argument value will be None
+        # Moreover it will be sent to the docker engine as None and then
+        # interpreted as string None which in many cases will fail the build
+        # That is why we filter out all pairs with value equal to None
+        buildargs = {k: v for k, v in build_opts.get('args', {}).items() if v != 'None'}
```
aanand: I think this should be comparing against the literal value `None`, not the string `'None'`?

tests/unit/service_test.py
```diff
@@ -481,7 +481,33 @@ def test_ensure_image_exists_force_build(self):
             forcerm=False,
             nocache=False,
             rm=True,
-            buildargs=None,
+            buildargs={},
+        )
+
+    def test_ensure_filter_out_empty_build_args(self):
+        args = {u'no_proxy': 'None', u'https_proxy': 'something'}
```
aanand: Similarly, this should read `'no_proxy': None`

aanand:
OK, I've had a look. I think the right place to fix this is at the config parsing stage, rather than in `service.py`. I've tracked the culprit down to [`build_string_dict` in `utils.py`](https://github.com/docker/compose/blob/86b5ed1a84b5046089abcfded8922d6dc7e318a2/compose/utils.py#L97-L98), which calls `str()` on every value in the dictionary, resulting in `None -> 'None'`.

Here's a fix:

```diff
diff --git a/compose/utils.py b/compose/utils.py
index 494beea..6718d5c 100644
--- a/compose/utils.py
+++ b/compose/utils.py
@@ -95,4 +95,6 @@ def microseconds_from_time_nano(time_nano):


 def build_string_dict(source_dict):
-    return dict((k, str(v)) for k, v in source_dict.items())
+    def to_str(value):
+        return '' if value is None else value
+    return dict((k, to_str(v)) for k, v in source_dict.items())
```

I think the correct thing to do is pass unset args through as the empty string, not omit them entirely. This is consistent with how both `environment` and interpolation (`${VAR}`) works. Will this fix your use case?

Andrey9kin:
@aanand sure. Would it be Ok if I separate those two fixes into two pull requests so we have good logical division between issues?

aanand:
I think the fix you've currently got to `service.py` is unnecessary once we've fixed it in `utils.py`, so I think you can just disregard my code comments and implement it here.

Andrey9kin:
@aanand thanks for your guidance. I also felt that it should be a way to implement it better. Please take a look on updated version

dnephin:
LGTM

compose/utils.py
```diff
@@ -95,4 +95,4 @@ def microseconds_from_time_nano(time_nano):
 
 
 def build_string_dict(source_dict):
-    return dict((k, str(v)) for k, v in source_dict.items())
+    return dict((k, str(v if v else '')) for k, v in source_dict.items())
```
aanand: Only problem with this is it'll result in the empty string for the integer `0`, which probably isn't what the user expects. `v if v is not None else ''` is probably better. (Would be good to add an arg with a value of `0` to the test, too.)

Andrey9kin:
@aanand missed that. Thank you for your comment. Updated accordingly

shin-:
LGTM :+1:

Andrey9kin:
@aanand @dnephin do you have any other comments for this issue? If not shall we get it merged?

dnephin:
LGTM

PR #3460, opened 2016-05-13
Fix docs: startup-order example config

romantolkachyov:
Without "--" there is an error because `wait-for-it` tries to parse original command like their own arguments.


romantolkachyov:
Thank you for a comprehensive instructions! I have done it.

PR #3482, opened 2016-05-19
Fix bug where confirmation prompt doesn't show due to line buffering

### LINKED ISSUE #2897: `docker-compose rm` on windows waits for y/N answer before the question is shown.
When I do a __docker-compose rm__ it just shows a new empty line and then nothing.
By trial and error I found out that the "Are you sure?" question doesn't display but it waits for input.
Look at the flow below...
```
$ docker-compose rm
y
Removing afdriftmysql_phpmyadmin_1 ... done
Removing afdriftmysql_db_1 ... done
Going to remove afdriftmysql_phpmyadmin_1, afdriftmysql_db_1
Are you sure? [yN]
```

I would expect the question to be asked before having to answer it.

* Windows 7 Pro x64
* docker-toolbox 1.10.0
* git 2.6.2 x64 from git-scm.com

aanand:
Alternate solutions, according to [this Stack Overflow answer](http://stackoverflow.com/questions/107705/disable-output-buffering):

- Ensure that `PYTHONUNBUFFERED=1`. I can't see a way to do that from within the Python code.
- Wrap stdout in an object which flushes on every write. It would work, and would cover other cases too, but I don't want to think about how/whether it would mess with dockerpty.

Closes #2897.

### DIFF
--- a/compose/cli/utils.py
+++ b/compose/cli/utils.py
@@ -6,9 +6,9 @@ import os
 import platform
 import ssl
 import subprocess
+import sys
 
 import docker
-from six.moves import input
 
 import compose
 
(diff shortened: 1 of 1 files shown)


shin-:
LGTM!

PR #3488, opened 2016-05-20
Add support for creating internal networks

### LINKED ISSUE #3190: Ability to create internal networks missing
I cannot find a way to specify in the docker-compose file (even v2) that a network must be internal (as if it was created with `docker network create --internal ...`).

jgiannuzzi:
Fixes #3190

### DIFF
--- a/compose/config/config_schema_v2.0.json
+++ b/compose/config/config_schema_v2.0.json
@@ -246,7 +246,8 @@
             "name": {"type": "string"}
           },
           "additionalProperties": false
-        }
+        },
+        "internal": {"type": "boolean"}
       },
       "additionalProperties": false
     },
--- a/compose/network.py
+++ b/compose/network.py
@@ -15,7 +15,7 @@ log = logging.getLogger(__name__)
 
 class Network(object):
     def __init__(self, client, project, name, driver=None, driver_opts=None,
-                 ipam=None, external_name=None):
+                 ipam=None, external_name=None, internal=False):
         self.client = client
         self.project = project
         self.name = name
@@ -23,6 +23,7 @@ class Network(object):
         self.driver_opts = driver_opts
         self.ipam = create_ipam_config_from_dict(ipam)
         self.external_name = external_name
+        self.internal = internal
 
     def ensure(self):
         if self.external_name:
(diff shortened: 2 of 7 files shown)


shin-:
Thanks!

Oh man, this is going to be great...
```yaml
networks:
  foo:
    internal: true
    external: true
```

All jokes aside, code LGTM!

jgiannuzzi:
Rebased onto master

requirements.txt
```diff
@@ -1,9 +1,9 @@
 PyYAML==3.11
 cached-property==1.2.0
-docker-py==1.8.1
 dockerpty==0.4.1
 docopt==0.6.1
 enum34==1.0.4
+git+https://github.com/docker/docker-py.git@26f2b696ab61a6dd41a19c9730f85a6b572089b7#egg=docker-py
 jsonschema==2.5.1
```
denismakogon: Hi there, i'm not really sure that we should do something like this, can't we wait until docker-py will support internal networks (until new version will see the light)?

denismakogon:
LGTM +1

requirements.txt
jgiannuzzi: Well, that's what @shin- suggested here: https://github.com/docker/docker-py/pull/1068#issuecomment-220461252

shin-: Yes, we do this all the time, then switch back to a release version when moving closer to release date.

jgiannuzzi:
Oops, I did something stupid and pressed the "Update branch" button in GitHub, and then something even more stupid, I deleted the branch.

Is there any way to open this PR again?

aanand:
LGTM

Unfortunately we're past code freeze, so I've moved this into the 1.9.0 milestone.

jgiannuzzi:
Thanks, @aanand!

peterthomassen:
The new paragraph in the documentation (https://github.com/docker/compose/pull/3488/files#diff-b94299404e9b589177d56080e454dac3R862) sounds like `internal` only is usable in combination with `overlay` networks. Is that true?

It looks like the paragraph was half taken from https://docs.docker.com/engine/reference/commandline/network_create/#network-internal-mode where the other half may have been relevant as well.

PR #3489, opened 2016-05-20
Use latest OpenSSL version when building Mac binary on Travis

aanand:
This should fix the recent [build failures](https://travis-ci.org/docker/compose/jobs/131662563) on master, and it's also good from a security standpoint.


shin-:
LGTM!

Issue #3503, opened 2016-05-24
CLITestCase.test_events_human_readable fails due to timezone difference


denismakogon:
CLITestCase.test_events_human_readable fails due to wrong assumption that host where tests were launched will have the same date time as Docker daemon.

Example, instance works with US East coast time, but compose tests were launched at host that does have UTC + 3 date time.

So, assertion that is made `assert lines[0].startswith(datetime.date.today().isoformat())` is incorrect.

denismakogon:
Here's a trace:
```
    assert lines[0].startswith(datetime.date.today().isoformat())
E   AssertionError: assert <built-in method startswith of unicode object at 0x10ff034b0>('2016-05-24')
E    +  where <built-in method startswith of unicode object at 0x10ff034b0> = '2016-05-23 19:48:44.751620 container create 10cac4563e293c28d4af8a08cad4db342ecac0ee4d0a6e69d54b7f12fd673d6e (image=busybox:latest, name=simplecomposefile_simple_1)'.startswith
E    +  and   '2016-05-24' = <built-in method isoformat of datetime.date object at 0x10fe8d4b0>()
E    +    where <built-in method isoformat of datetime.date object at 0x10fe8d4b0> = datetime.date(2016, 5, 24).isoformat
E    +      where datetime.date(2016, 5, 24) = <built-in method today of type object at 0x10da88d00>()
E    +        where <built-in method today of type object at 0x10da88d00> = <type 'datetime.date'>.today
E    +          where <type 'datetime.date'> = datetime.date

```

denismakogon:
@shin- @dnephin hey guys, would you be kind to review this patch, thanks.

ijc:
Issue grooming: seems like the link PR was merged.

PR #3504, opened 2016-05-24
Issue-3503: Improve timestamp validation in tests

denismakogon:
CLITestCase.test_events_human_readable fails due to
 wrong assumption that host where tests were launched
 will have the same date time as Docker daemon.

 This fix introduces internal method for validating
 timestamp in Docker logs

### DIFF
--- a/tests/acceptance/cli_test.py
+++ b/tests/acceptance/cli_test.py
@@ -1473,6 +1473,17 @@ class CLITestCase(DockerClientTestCase):
         assert Counter(e['action'] for e in lines) == {'create': 2, 'start': 2}
 
     def test_events_human_readable(self):
+
+        def has_timestamp(string):
+            str_iso_date, str_iso_time, container_info = string.split(' ', 2)
+            try:
+                return isinstance(datetime.datetime.strptime(
+                    '%s %s' % (str_iso_date, str_iso_time),
+                    '%Y-%m-%d %H:%M:%S.%f'),
+                    datetime.datetime)
+            except ValueError:
+                return False
+
         events_proc = start_process(self.base_dir, ['events'])
         self.dispatch(['up', '-d', 'simple'])
         wait_on_condition(ContainerCountCondition(self.project, 1))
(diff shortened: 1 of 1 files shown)


denismakogon:
Fixes #3503

tests/acceptance/cli_test.py
```diff
@@ -1473,6 +1473,15 @@ def test_events_json(self):
         assert Counter(e['action'] for e in lines) == {'create': 2, 'start': 2}
 
     def test_events_human_readable(self):
+
+        def has_timestamp(string):
+            str_iso_datetime = ' '.join(string.split(' ')[:2])
+            try:
+                return isinstance(datetime.datetime.strptime(
+                    str_iso_datetime, '%Y-%m-%d %H:%M:%S.%f'), datetime.datetime)
```
dnephin: I don't think you need the `isinstance` here because `strptime` always returns a `datetime`. 

I would just `return datetime.datetime.strptime(...)`

tests/acceptance/cli_test.py
```diff
@@ -1473,6 +1473,15 @@ def test_events_json(self):
         assert Counter(e['action'] for e in lines) == {'create': 2, 'start': 2}
 
     def test_events_human_readable(self):
+
+        def has_timestamp(string):
+            str_iso_datetime = ' '.join(string.split(' ')[:2])
```
dnephin: You can do this in a more obvious way:

```python

str_iso_datetime, _ = line.split(' ', 2)

```

dnephin:
makes sense, just a couple comments

tests/acceptance/cli_test.py
denismakogon: So the reason I've used 'isinstance' because I thought True/False is more obvious rather than datetime and None or False. Such approach seemed to be more correct due to return statement semantic consistency, am I wrong?

tests/acceptance/cli_test.py
denismakogon: Agreed, will fix soon.

denismakogon:
@aanand @shin- Hey, guys, would you mind to review this patch?

tests/acceptance/cli_test.py
```diff
@@ -1473,6 +1473,18 @@ def test_events_json(self):
         assert Counter(e['action'] for e in lines) == {'create': 2, 'start': 2}
 
     def test_events_human_readable(self):
+
+        def has_timestamp(string):
+            print(string.split(' ', 2)[:-1])
```
aanand: I think this line can be removed?

denismakogon:
@aanand done.

shin-:
LGTM!

PR #3523, opened 2016-05-28
Fix Invalid Restart Policy

### LINKED ISSUE #3481: [1.7.1] Containers can be created with an invalid restart policy name
`docker-compose.yml`:

``` yaml
version: '2'
services:
  alpine:
    image: "alpine:latest"
    restart: unlesss-stopped
```

In the previous, `unlesss-stopped` is an invalid value for the restart policy name.

``` sh
sudo docker-compose up
```

Does not complain, and a container is created and started with an incorrect restart policy. Incorrect, based on `docker inspect` output.

 In contrast, this fails:

``` sh
sudo docker create --restart='unlesss-stopped' alpine
```

```
docker: invalid restart policy unlesss-stopped.
See 'docker create --help'.
```

**Context**

```
docker-compose version 1.7.1, build 6c29830
Docker version 1.11.1, build 5604cbe
```

johnharris85:
Fixes #3481.
Signed-off-by: John Harris <john@johnharris.io>

### DIFF
--- a/compose/config/types.py
+++ b/compose/config/types.py
@@ -75,13 +75,31 @@ class VolumeFromSpec(namedtuple('_VolumeFromSpec', 'source mode type')):
 
 
 def parse_restart_spec(restart_config):
+    allowed_restart_policies = [
+        'no',
+        'always',
+        'on-failure',
+        'unless-stopped',
+    ]
     if not restart_config:
         return None
+
     parts = restart_config.split(':')
+    restart_policy = parts[0]
+    if restart_policy not in allowed_restart_policies:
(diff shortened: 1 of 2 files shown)


johnharris85:
This look OK @dnephin @aanand ?

dnephin:
see my comment in #3481. I don't think we should be duplicating the supported policies here.

johnharris85:
Thanks @dnephin, only just saw it. Agree, I'll close.

johnharris85:
I have submitted an initial PR crack at fixing this in the engine here https://github.com/docker/docker/pull/24073.

PR #3540, opened 2016-06-01
Fix version function related to the new tags structure at the docker repository

GM-Alex:
The docker team changed it's tag structure which breaks the parse function of the version.py script. This version uses regex to grab the version and should be much more robust.


bfirsh:
@GM-Alex Thanks for this! Looks like this has fixed in part by https://github.com/docker/compose/commit/4bf5271ae2d53f8c6467642b6bd4c3372ed52da8 and this change collides with that. Do you think this is still required?

@dnephin WDYT?

dnephin:
I don't think this is correct. My fix is working for me on master. This change will treat doc tags as code releases, which is not correct and will return the wrong output.

GM-Alex:
@dnephin I agree if doc tags shouldn't be in there but the regex approach is much more robust, so you should consider to use regex anyway and check for doc tags or invalide version numbers by using regex.

dnephin:
I disagree: https://blog.codinghorror.com/regular-expressions-now-you-have-two-problems/

GM-Alex:
I'm not a big fan of using try catch instead of input validation, but your change fix the issue, so I'm fine if we don't use my approach.

GM-Alex:
@dnephin One more thing. I got now some time to read the article you mentioned above. And I would say that is not the case for this pull request. Jeff is saying that a over use of regex is bad and regex you don't understand. I totally agree, but in that case the regex is very simple and in my opinion regex is a good way to validate and parse small input, like version numbers.

PR #3543, opened 2016-06-02
Replace assertEquals with assertEqual in unittests since the former is getting dep…

milin:
…recated soon. Also, fix pep8 E309.

https://docs.python.org/2.7/library/unittest.html#deprecated-aliases


bfirsh:
LGTM

dnephin:
I think we'd like to convert these to `assert x == y` so that we get the nicer pytest failure messages, but this works for now.

LGTM

milin:
@dnephin Should I convert these to `assert x == y`? Only gotcha i see there is, we would have to convert rest of the test cases so everything is consistent, which might be a little more effort, and also a candidate for a fresh pr after this?

PR #3548, opened 2016-06-03
fix serialize restart spec with null string

realityone:
Signed-off-by: realityone <realityone@me.com>

Compose will meet an unexpected error when `restart` field is a null string, but it is valid in docker. 
~~so I think it should add a default value in `parse_restart_spec` function.~~

Here is the error info:

```
[realityone@rEimu ~/Downloads]$ cat docker-compose.yml 
restart-null:
  image: busybox
  restart: ''

[realityone@rEimu ~/Downloads]$ docker-compose -f docker-compose.yml config
  File "compose/cli/main.py", line 58, in main
  File "compose/cli/main.py", line 103, in perform_command
  File "compose/cli/main.py", line 242, in config
  File "compose/config/serialize.py", line 24, in serialize_config
  File "compose/config/serialize.py", line 49, in denormalize_service_dict
  File "compose/config/types.py", line 95, in serialize_restart_spec
TypeError: 'NoneType' object has no attribute '__getitem__'
docker-compose returned -1
[realityone@rEimu ~/Downloads]$ 
```

### DIFF
--- a/compose/cli/docker_client.py
+++ b/compose/cli/docker_client.py
@@ -45,10 +45,6 @@ def docker_client(environment, version=None, tls_config=None, host=None,
     Returns a docker-py client configured using environment variables
     according to the same logic as the official Docker client.
     """
-    if 'DOCKER_CLIENT_TIMEOUT' in environment:
-        log.warn("The DOCKER_CLIENT_TIMEOUT environment variable is deprecated.  "
-                 "Please use COMPOSE_HTTP_TIMEOUT instead.")
-
     try:
         kwargs = kwargs_from_env(environment=environment, ssl_version=tls_version)
     except TLSParameterError:
--- a/compose/cli/errors.py
+++ b/compose/cli/errors.py
@@ -13,7 +13,6 @@ from requests.exceptions import SSLError
 from requests.packages.urllib3.exceptions import ReadTimeoutError
 
 from ..const import API_VERSION_TO_ENGINE_VERSION
-from ..const import HTTP_TIMEOUT
 from .utils import call_silently
 from .utils import is_docker_for_mac_installed
 from .utils import is_mac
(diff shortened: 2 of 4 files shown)


dnephin:
I think we should probably fix this by changing `serialize_restart_spec()` instead of setting this default during parse.

realityone:
@dnephin
Thanks for advice, fix this in `serialize_restart_spec()` is more reasonable.

realityone:
CI failed. It seems like docker's incompatible API `/networks` return `null` instead of `[]`.

compose/config/types.py
```diff
@@ -92,6 +92,8 @@ def parse_restart_spec(restart_config):
 
 
 def serialize_restart_spec(restart_spec):
+    if not restart_spec:
+        return ''
```
dnephin: Hmm, I was thinking this would be `None`, I'm not sure if this works just as well.

realityone: Actually it can't be `None`, because we specified `"restart": {"type": "string"},` in `compose_schema_v*.json`, if we return `None` here, the `docker-compose config` command will output an invalid restart field.

```
(develop)[realityone@rEimu ~/compose]$ bin/docker-compose -f tests/fixtures/restart/docker-compose.yml config
networks: {}
services:
  always:
    image: busybox
    restart: always
  never:
    image: busybox
    restart: 'no'
  never-with-null:
    image: busybox
    restart: null
  on-failure:
    image: busybox
    restart: on-failure
  on-failure-5:
    image: busybox
    restart: on-failure:5
version: '2.0'
volumes: {}

(develop)[realityone@rEimu ~/compose]$ bin/docker-compose -f tests/fixtures/restart/docker-compose.yml config | bin/docker-compose -f - config
ERROR: The Compose file is invalid because:
services.never-with-null.restart contains an invalid type, it should be a string
```

dnephin: ok, great, thanks for confirming.

dnephin:
LGTM

PR #3559, opened 2016-06-07
Added bundler gem path as a volume to documentation 

madwire:
By adding the volume, the developer doesn’t need to install every single gem every time they want to add an extra dependency. It just makes the development process smoother.


shin-:
Closing as docs have moved.

If you haven't already, feel free to resubmit the PR to https://github.com/docker/docker.github.io

Thank you!

Issue #3525, opened 2016-05-28
Missing dependency on Ubuntu 16.04


dooman87:
I used original docker instruction (https://docs.docker.com/engine/installation/linux/ubuntulinux/) to install docker on ubuntu 16.04 that was created in GCP. It worked like a charm and manual is really great! 

However, docker compose didn't work out of the box:

```
:~$ docker-compose --version
Traceback (most recent call last):
  File "/usr/bin/docker-compose", line 9, in <module>
    load_entry_point('docker-compose==1.5.2', 'console_scripts', 'docker-compose')()
  File "/usr/lib/python2.7/dist-packages/pkg_resources/__init__.py", line 542, in load_entry_point
    return get_distribution(dist).load_entry_point(group, name)
  File "/usr/lib/python2.7/dist-packages/pkg_resources/__init__.py", line 2569, in load_entry_point
    return ep.load()
  File "/usr/lib/python2.7/dist-packages/pkg_resources/__init__.py", line 2229, in load
    return self.resolve()
  File "/usr/lib/python2.7/dist-packages/pkg_resources/__init__.py", line 2235, in resolve
    module = __import__(self.module_name, fromlist=['__name__'], level=0)
  File "/usr/lib/python2.7/dist-packages/compose/cli/main.py", line 11, in <module>
    from docker.errors import APIError
  File "/usr/lib/python2.7/dist-packages/docker/__init__.py", line 20, in <module>
    from .client import Client, AutoVersionClient, from_env # flake8: noqa
  File "/usr/lib/python2.7/dist-packages/docker/client.py", line 25, in <module>
    from . import api
  File "/usr/lib/python2.7/dist-packages/docker/api/__init__.py", line 2, in <module>
    from .build import BuildApiMixin
  File "/usr/lib/python2.7/dist-packages/docker/api/build.py", line 9, in <module>
    from .. import utils
  File "/usr/lib/python2.7/dist-packages/docker/utils/__init__.py", line 1, in <module>
    from .utils import (
  File "/usr/lib/python2.7/dist-packages/docker/utils/utils.py", line 33, in <module>
    from .. import tls
  File "/usr/lib/python2.7/dist-packages/docker/tls.py", line 5, in <module>
    from .ssladapter import ssladapter
  File "/usr/lib/python2.7/dist-packages/docker/ssladapter/__init__.py", line 1, in <module>
    from .ssladapter import SSLAdapter # flake8: noqa
[...]

PR #3566, opened 2016-06-08
Missing dependency on Ubuntu 16.04 #3525

dooman87:
* Updated install instruction.

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -51,6 +51,11 @@ which the release page specifies, in your terminal.
 6.  Optionally, install [command completion](completion.md) for the
 `bash` and `zsh` shell.
 
+     > Note: On Ubuntu 16.04 you also need to install python-ipaddress package:
+     ```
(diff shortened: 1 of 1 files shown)


dnephin:
I don't think this is correct. If you're using the binary install you shouldn't need to install any other packages.

kevinrood:
@dooman87 thank you for this update. I can confirm this helped me work around my setup issue on Ubuntu 16.04.

dooman87:
@dnephin Sorry, what you mean by "shouldn't need to install any other packages"? In the documentation the first step is to install docker-engine which will install python and other packages. However, you still need to install that additional package on Ubuntu 16.04.

dnephin:
I think the problem might be that our requirements.txt is not correct anymore, I'll submit a PR to fix that.

Are you install Compose by curling the binary, or are you using pip to install it?

dooman87:
I used curl.

Cheers,
Dmitry

PR #3572, opened 2016-06-09
Allowing rm option to stop containers, if required, before removing

rawkode:


rawkode:
No idea what happened to this. Closing!

Issue #3576, opened 2016-06-10
1.8.dev0 today: not using str argument to rstrip()


sanmai-NL:
`readline` method returns a `str`.
With CPython 3.5.1
@ https://github.com/docker/compose/tree/e502417df28e689d07006eb6e7a7e29729ddda82
```
Traceback (most recent call last):
  File "/opt/venvs/docker-compose/bin/docker-compose", line 9, in <module>
    load_entry_point('docker-compose==1.8.0.dev0', 'console_scripts', 'docker-compose')()
  File "/opt/venvs/docker-compose/lib64/python3.5/site-packages/compose/cli/main.py", line 58, in main
    command()
  File "/opt/venvs/docker-compose/lib64/python3.5/site-packages/compose/cli/main.py", line 109, in perform_command
    handler(command, command_options)
  File "/opt/venvs/docker-compose/lib64/python3.5/site-packages/compose/cli/main.py", line 553, in rm
    or yesno("Are you sure? [yN] ", default=False):
  File "/opt/venvs/docker-compose/lib64/python3.5/site-packages/compose/cli/utils.py", line 33, in yesno
    answer = input(prompt).strip().lower()
  File "/opt/venvs/docker-compose/lib64/python3.5/site-packages/compose/cli/utils.py", line 52, in input
    return sys.stdin.readline().rstrip(b'\n')
TypeError: rstrip arg must be None or str
```
Will submit PR for it.

sanmai-NL:
The offending code:
```py
def input(prompt):
    """
    Version of input (raw_input in Python 2) which forces a flush of sys.stdout
    to avoid problems where the prompt fails to appear due to line buffering
    """
    sys.stdout.write(prompt)
    sys.stdout.flush()
    return sys.stdin.readline().rstrip(b'\n')
```
Wouldn't disabling stdout buffering altogether using this CPython CLI option (with 3.5.1 and earlier versions): 
```
  -u     Force the binary I/O layers of stdout and stderr to be unbuffered.  stdin is always buffered.  The text I/O layer will still be line-buffered.
```
address the original issue adequately? I'm not aware of other use of standard output. 
By the way, I think standard error should be used for non-parsable output.

shin-:
Closing via #3577 - thanks!

PR #3577, opened 2016-06-10
Fix byte/str typing error

sanmai-NL:
See #3576.

### DIFF
--- a/compose/cli/utils.py
+++ b/compose/cli/utils.py
@@ -49,7 +49,7 @@ def input(prompt):
     """
     sys.stdout.write(prompt)
     sys.stdout.flush()
-    return sys.stdin.readline().rstrip(b'\n')
+    return sys.stdin.readline().rstrip('\n')
 
 
 def call_silently(*args, **kwargs):


shin-:
Thanks! Change LGTM.
@aanand WDYT?

aanand:
Will this break on Python 2?

shin-:
In Python 2, `b'...'` and `'...'` are one and the same, right?

aanand:
Usually, but we've got a `from __future__ import unicode_literals` at the top, so it'll be a Unicode string in both Python versions.

shin-:
Ah, good point.
I just tested it in ipython and it seems to work fine either way.
Python 2:
```python
In [3]: sys.stdin.readline().rstrip('\n')
hello
Out[3]: 'hello'

In [4]: sys.stdin.readline().rstrip(u'\n')
hello
Out[4]: u'hello'
```

Python 3:
```python
In [3]: sys.stdin.readline().rstrip(b'\n')
hello
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-3-cefaff8b347e> in <module>()
----> 1 sys.stdin.readline().rstrip(b'\n')

TypeError: rstrip arg must be None or str

In [5]: sys.stdin.readline().rstrip(u'\n')
hello
Out[5]: 'hello'
```

Python 2 with `unicode_literals`:
```
In [1]: from __future__ import unicode_literals

In [2]: import sys

In [3]: sys.stdin.readline().rstrip('\n')
hi
Out[3]: u'hi'
```

aanand:
Great. LGTM

PR #3579, opened 2016-06-10
Do not assign `None` to name for exceptions

sanmai-NL:


shin-:
I'm not sure what this is supposed to accomplish?

sanmai-NL:
PyLint complains about this, the alternative style is more flexible.

PR #3580, opened 2016-06-10
Add a minimal Pylint pre-commit hook

sanmai-NL:


sanmai-NL:
Can you review?

shin-:
We're already checking the code against flake8, so this seems redundant / pedantic.

PR #3589, opened 2016-06-14
Added support for oom_score_adj

### LINKED ISSUE #3565: Add support for "oom_score_adj" in manifest file
Docker (as of 1.10.0) supports a flag [`--oom-score-adj`](https://github.com/docker/docker/pull/16277), which allows adjustment of the oom-killer scoring for a container. This is useful because it allows you to protect your service from the oom-killer, or at least give it an idea of the priority of your containers so that it doesn't go wild and kill the important ones first.

I didn't see any documentation about how to wire this flag into a Compose manifest in the [reference docs](https://docs.docker.com/compose/compose-file/), nor did I find anything in the github issues/pr tracker by searching for "oom", so I assume it's unimplemented.

I'm down make a PR to wire up this flag, but first wanted to raise the issue first in case there was something I missed or if there was a reason it wasn't there.

Knetic:
Does what it says on the tin. 

Should close #3565.

### DIFF
--- a/compose/config/config.py
+++ b/compose/config/config.py
@@ -70,6 +70,7 @@ DOCKER_CONFIG_KEYS = [
     'mem_limit',
     'memswap_limit',
     'net',
+    'oom_score_adj'
     'pid',
     'ports',
     'privileged',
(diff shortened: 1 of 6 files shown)


Knetic:
Forgot to sign commit. Cancelling to fix.

PR #3601, opened 2016-06-14
Warn on missing digests, don't push/pull by default

aanand:
Safer default behaviour. Added a `--fetch-digests` flag to `bundle` to automatically push/pull.

```console
$ docker-compose bundle
ERROR: Some images are missing digests.

The following images need to be pushed:

    aanand/counter

The following images need to be pulled:

    redis

If this is OK, run `docker-compose bundle --fetch-digests`.

```

### DIFF
--- a/compose/bundle.py
+++ b/compose/bundle.py
@@ -40,6 +40,22 @@ SUPPORTED_KEYS = {
 VERSION = '0.1'
 
 
+class NeedsPush(Exception):
+    def __init__(self, image_name):
+        self.image_name = image_name
+
+
+class NeedsPull(Exception):
+    def __init__(self, image_name):
+        self.image_name = image_name
+
+
+class MissingDigests(Exception):
+    def __init__(self, needs_push, needs_pull):
+        self.needs_push = needs_push
+        self.needs_pull = needs_pull
+
+
 def serialize_bundle(config, image_digests):
     if config.networks:
         log.warn("Unsupported top level key 'networks' - ignoring")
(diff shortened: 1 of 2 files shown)


compose/cli/main.py
```diff
@@ -235,7 +241,26 @@ def bundle(self, config_options, options):
...
+
+                if e.needs_push:
+                    paras += ["The following images need to be pushed:", list_images(e.needs_push)]
+
+                if e.needs_pull:
+                    paras += ["The following images need to be pulled:", list_images(e.needs_pull)]
+
+                paras.append("If this is OK, run `docker-compose bundle --fetch-digests`.")
+
+                raise UserError("\n\n".join(paras))
```
dnephin: Minor nit, the context only needs to be around the `get_image_digests()` line not the entire block

compose/bundle.py
```diff
@@ -108,12 +143,15 @@ def get_image_digest(service):
...
         raise ValueError("Failed to get digest for %s" % service.name)
 
+    repo = parse_repository_tag(service.options['image'])[0]
     identifier = '{repo}@{digest}'.format(repo=repo, digest=digest)
 
     # Pull by digest so that image['RepoDigests'] is populated for next time
     # and we don't have to pull/push again
     service.client.pull(identifier)
 
+    log.info("Stored digest for {}".format(service.image_name))
```
dnephin: This extra pull isn't going to be necessary if we get https://github.com/docker/docker/pull/23503 into 1.12.

I wonder if we should remove it now (along with this info message).

dnephin:
Making pull optional makes sense to me, especially if we get https://github.com/docker/docker/pull/23503 into 1.12.

Making push conditional doesn't make as much sense. You're almost always going to need to push, it's the common path, and making it non-default seems strange.

compose/bundle.py
aanand: It'll still be necessary on pre-1.12 engines (although said engines don't support bundles, so that's arguably a minority case).

aanand:
It's simply that `push` is a destructive operation that could potentially affect other people, so I feel it doesn't hurt to be careful about it.

compose/bundle.py
dnephin: Ok, let's keep it for now.

dnephin:
That's fair, LGTM

PR #3612, opened 2016-06-16
Add some unit tests and an acceptance test for bundle

dnephin:
In addition to the tests, this PR also makes the following changes:
* the `Services`, and `Version` keys in the bundle are now capitalized to be consistent with the other keys. No change is required on the engine side because the go json parser accepts both cases.
* only re-pull by digest if the original pull didn't save the `RepoDigests`. This will be a significant improvement when https://github.com/docker/docker/pull/23503 is merged

The acceptance test for `bundle` is for the most trivial case, but the unit tests should cover most of the other behaviour.

I tried to write an acceptance test for `push` by starting a local registry. Unfortunately the authentication required by the engine to communicate with a registry makes this incredibly annoying so I think it'll have to wait.

### DIFF
--- a/compose/bundle.py
+++ b/compose/bundle.py
@@ -57,17 +57,7 @@ class MissingDigests(Exception):
 
 
 def serialize_bundle(config, image_digests):
-    if config.networks:
-        log.warn("Unsupported top level key 'networks' - ignoring")
-
-    if config.volumes:
-        log.warn("Unsupported top level key 'volumes' - ignoring")
-
-    return json.dumps(
-        to_bundle(config, image_digests),
-        indent=2,
-        sort_keys=True,
-    )
+    return json.dumps(to_bundle(config, image_digests), indent=2, sort_keys=True)
 
 
 def get_image_digests(project, allow_fetch=False):
@@ -99,7 +89,7 @@ def get_image_digest(service, allow_fetch=False):
             "required to generate a proper image digest for the bundle. Specify "
             "an image repo and tag with the 'image' option.".format(s=service))
 
-    separator = parse_repository_tag(service.options['image'])[2]
+    _, _, separator = parse_repository_tag(service.options['image'])
     # Compose file already uses a digest, no lookup required
     if separator == '@':
         return service.options['image']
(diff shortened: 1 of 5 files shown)


compose/bundle.py
```diff
@@ -143,24 +133,32 @@ def fetch_image_digest(service):
...
-    repo = parse_repository_tag(service.options['image'])[0]
+    repo, _, _ = parse_repository_tag(service.options['image'])
     identifier = '{repo}@{digest}'.format(repo=repo, digest=digest)
 
-    # Pull by digest so that image['RepoDigests'] is populated for next time
-    # and we don't have to pull/push again
-    service.client.pull(identifier)
-
-    log.info("Stored digest for {}".format(service.image_name))
+    # only do this is RepoTags isn't already populated
```
aanand: Small thing: "is" should be "if". And should "RepoTags" be "RepoDigests"?

tests/acceptance/cli_test.py
```diff
@@ -378,6 +379,32 @@ def test_build_failed_forcerm(self):
         ]
         assert not containers
 
+    def test_bundle_with_digests(self):
+        self.base_dir = 'tests/fixtures/bundle-with-digests/'
+        tmpdir = py.test.ensuretemp('cli_test_bundle')
+        self.addCleanup(tmpdir.remove)
+        filename = str(tmpdir.join('example.dsb'))
```
aanand: Another small thing: this should be `.dab` now

tests/integration/service_test.py
```diff
@@ -397,7 +397,7 @@ def test_execute_convergence_plan_when_host_volume_is_removed(self):
 
         assert not mock_log.warn.called
         assert (
-            [mount['Destination'] for mount in new_container.get('Mounts')],
+            [mount['Destination'] for mount in new_container.get('Mounts')] ==
```
aanand: This has now been fixed in #3619, but if GH isn't complaining, I guess the merge won't blow up.

aanand:
LGTM - just minor comments. Feel free to fix them here or in a follow-up.

The test failures are 1.12-related. Two are fixed by #3652, the other (I think) needs to be fixed in Engine.

dnephin:
fixed those issues

aanand:
LGTM

PR #3619, opened 2016-06-20
Fix assertion that was always true

jgiannuzzi:
An assertion was always evaluating to true in test_execute_convergence_plan_when_host_volume_is_removed.


dnephin:
LGTM

aanand:
Tests are broken against 1.12.0-rc2, but that's not the fault of this change. Confirmed that they pass locally against 1.11.2, so LGTM.

PR #3595, opened 2016-06-14
Add docker-compose push and docker-compose bundle

dnephin:
* `push` pushes services images to a registry
* `bundle` pulls/pushes as needed and builds a bundle file to be consumed by https://github.com/docker/docker/pull/23522

PR #3625, opened 2016-06-22
bash completion for `docker-compose bundle`

albers:
Ref: #3595
Please include in 1.8.0 as the feature is present there.


dnephin:
LGTM

PR #3626, opened 2016-06-22
bash completion for `docker-compose push`

albers:
Ref: #3595
Please include in 1.8.0 as the feature is present there.

I am not sure about which services to complete. Is "all services" correct? `docker-compose pull` restricts the completion to services based on images. I don't think this applies to `push`.


dnephin:
LGTM

PR #3652, opened 2016-06-27
Fix tests to accommodate short-id container alias

### LINKED ISSUE #3629: tests/acceptance test_run_* fails on 1.12.x engine
Something changed on the `1.12` engine where a container connected on a custom network, always have itself (the `shortid` it seems) as aliases. But two tests – `test_run_detached_connects_to_network` and `test_run_interactive_connects_to_network` — are asserting that the container created by compose have no aliases (`assert not config['Aliases']`), so they fail.

```bash
λ py.test -vs -tb=short -k network test/acceptance
# […]
_______________________________________________ CLITestCase.test_run_detached_connects_to_network _______________________________________________
tests/integration/testcases.py:50: in wrapper
    return f(self, *args, **kwargs)
tests/acceptance/cli_test.py:1155: in test_run_detached_connects_to_network
    assert not config['Aliases']
E   AssertionError: assert not ['dd597a597012']
_____________________________________________ CLITestCase.test_run_interactive_connects_to_network ______________________________________________
tests/integration/testcases.py:50: in wrapper
    return f(self, *args, **kwargs)
tests/acceptance/cli_test.py:1138: in test_run_interactive_connects_to_network
    assert not config['Aliases']
E   AssertionError: assert not ['f66515801541']
```

It's on master (and 1.8.0-rc1).

/cc @aanand @dnephin

aanand:
In Engine 1.12, containers join a network with a short-id alias by default. A couple of our tests don't expect this, so I've updated them to not mind whether the short-id alias is there or not. Once we drop support for Engine <1.12, we can make them explicitly expect it.

Closes #3629.

### DIFF
--- a/tests/acceptance/cli_test.py
+++ b/tests/acceptance/cli_test.py
@@ -1135,7 +1135,10 @@ class CLITestCase(DockerClientTestCase):
             ]
 
             for _, config in networks.items():
-                assert not config['Aliases']
+                # TODO: once we drop support for API <1.24, this can be changed to:
+                # assert config['Aliases'] == [container.short_id]
+                aliases = set(config['Aliases']) - set([container.short_id])
+                assert not aliases
 
     @v2_only()
     def test_run_detached_connects_to_network(self):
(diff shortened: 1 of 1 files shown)


aanand:
The remaining failure is, I believe, an Engine bug:

```
______________________ CLITestCase.test_up_with_links_v1 _______________________
tests/acceptance/cli_test.py:729: in test_up_with_links_v1
    assert networks == []
E   AssertionError: assert None == []
```

`networks` is the result of an API call to `/networks`, and should always be an array.

dnephin:
We could just relax this to be `assert not networks`.

dnephin:
LGTM

aanand:
We could, but I consider it an Engine bug, and so it's a good thing that we caught it. I've opened https://github.com/docker/docker/issues/24063 to track it.

dnephin:
I agree, it is just really nice to have green CI, even if we're ignoring a few bugs in our dependencies.

PR #3653, opened 2016-06-28
Add support for link-local IPs in service.networks definition

### LINKED ISSUE #3637: Add support for link-local IPs
Engine issue: https://github.com/docker/docker/pull/23415
~~Depends on: https://github.com/docker/docker-py/issues/1107~~ https://github.com/docker/docker-py/commit/986a14a152d20df6bb8622516ab2d18af73e5e20

shin-:
Fixes #3637


compose/config/config.py
```diff
@@ -171,7 +172,7 @@ def version(self):
         if version == '2':
             version = V2_0
 
-        if version != V2_0:
+        if version not in (V2_0, V2_1):
```
dnephin: minor: maybe `if version == V1` would be more correct here, since it's the only version that would not be supported (now and in the future).

setup.py
```diff
@@ -34,7 +34,7 @@ def find_version(*file_paths):
     'requests >= 2.6.1, < 2.8',
     'texttable >= 0.8.1, < 0.9',
     'websocket-client >= 0.32.0, < 1.0',
-    'docker-py == 1.9.0rc2',
+    'docker-py < 2.0',
```
dnephin: also `> 1.8.0`, right ?

compose/config/config.py
shin-: Actually, we want to raise an error here if the user inputs a non-yet released version (like "2.5" or "3"), so I think the test as it is is more correct.

dnephin: Ah, good point. I forgot this could still be any string.

shin-:
Rebased, included https://github.com/docker/docker-py/pull/1139

shin-:
Rebased!

docs/compose-file.md
```diff
@@ -623,7 +623,7 @@ An example:
 
 #### link_local_ips
 
-> [Version 2.1 file format](#version-2.1) only.
+> [Added in version 2.1](#version-2.1).
```
dnephin: Sorry, this probably needs to be "file format version 2.1" or something like that, so it's not confused with the Compose version, right?

PR #3655, opened 2016-06-28
Fix some test assertions to work with Engine 1.12

shin-:
This should let the tests on Jenkins / Appveyor pass again.


shin-:
Ah, nevermind, that's #3652 -_-

PR #3661, opened 2016-06-29
Add missing deps to requirements.txt

dnephin:
Looks like we were missing some deps in `requirements.txt`

Fixes #3566

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,9 +1,12 @@
 PyYAML==3.11
+backports.ssl-match-hostname==3.5.0.1; python_version < '3'
 cached-property==1.2.0
 docker-py==1.8.1
 dockerpty==0.4.1
 docopt==0.6.1
-enum34==1.0.4
+enum34==1.0.4; python_version < '3.4'
+functools32==3.2.3.post2; python_version < '3.2'
+ipaddress==1.0.16
 jsonschema==2.5.1
 requests==2.7.0
 six==1.7.3


dnephin:
Test failure is just the known one from engine.

I'm not sure how many people this will impact, so far only reported on ubuntu 16.04. We might want to consider it for the 1.8 release.

cc @aanand , @shin-

shin-:
I'm confused - shouldn't those be installed automatically when installing `docker-py`? If it helps, I don't mind. Just seems weird.

Also, maybe `setup.py` needs to be updated as well?

dnephin:
`setup.py` and `requirements.txt` are supposed to be different lists. `setup.py` is the list of direct dependencies (packages imported directly from Compose), where as `requirements.txt` is a fully resolved list of dependencies (include indirect dependencies) generated by `pip freeze`. https://caremad.io/2013/07/setup-vs-requirement/ is a longer write-up about it.

You're right that it should be installed anyway from the dependency in `docker-py`, I'm not sure why it would't be. So I guess it's possible that this doesn't fix #3566, but at the very least it should prevent our builds from breaking if one of these gets updated to a new major version.

aanand:
I can't reproduce the problem described in #3566 inside an `ubuntu:16.04` container, either with the binary or pip installation.

I also share @shin-'s confusion - the dependencies should be automatically installed as part of installing docker-py. Indeed, when I `pip install docker-compose` in the ubuntu container (after installing `python` and `python-pip`), they are:

```
Collecting enum34<2,>=1.0.4 (from docker-compose)
  Downloading enum34-1.1.6-py2-none-any.whl
Collecting ipaddress>=1.0.16; python_version < "3.3" (from docker-py<2,>=1.8.1->docker-compose)
  Downloading ipaddress-1.0.16-py27-none-any.whl
Collecting backports.ssl-match-hostname>=3.5; python_version < "3.5" (from docker-py<2,>=1.8.1->docker-compose)
  Downloading backports.ssl_match_hostname-3.5.0.1.tar.gz
Collecting functools32; python_version == "2.7" (from jsonschema<3,>=2.5.1->docker-compose)
  Downloading functools32-3.2.3-2.zip
```

aanand:
> So I guess it's possible that this doesn't fix #3566, but at the very least it should prevent our builds from breaking if one of these gets updated to a new major version.

Sounds OK to me. Any potential _drawbacks_ to pinning those versions?

dnephin:
There shouldn't be

aanand:
Okay, LGTM

dnephin:
Apparently the version of pip on travis OSX is old and doesn't support this https://travis-ci.org/docker/compose/jobs/141472659

I'll see if we can upgrade it

PR #3662, opened 2016-06-29
bash completion for `docker-compose bundle --fetch-digests`

albers:
Ref: #3601


albers:
Please schedule for 1.8.0 because the feature is present there.

aanand:
LGTM

Issue #3664, opened 2016-06-30
config does not handle unnamed volumes correct


lindt:
I do have several compose files. My goal is to combine them using "docker-compose config". The problem is, that docker-compose does not understand its own created file after that.

Given a compose file with an unnamed volume

``` yml
version: '2'
services:
  test:
    image: ubuntu
    volumes:
      - /config
```

When running 'docker-compose config'
Then the unnamed volume will be suffixed with ":rw"
Unnamed volumes are currently not allowed to be suffixed with ":rw".

This causes docker-compose to do not understand it's own format anymore.
Error message is like

```
Creating composeconfig_test_1

ERROR: for test  b'Invalid volume spec "rw": volumeabs: Invalid volume destination path: \'rw\' mount path must be absolute.'
ERROR: Encountered errors while bringing up the project.
```

Once running docker-compose config again on that file, the volume will be again suffixed with ":rw".

PR #3665, opened 2016-06-30
unnamed volumes do not accept a mode suffix

lindt:
as long as unnamed volumes do not accept mode (':rw') suffixes, `docker-compose config` should not add such a suffix.

This references issue #3664.

However I'm not sure if the missing support is the problem or just the wrong `docker-compose config`-output. But anyway `docker-compose config` should not output anything that is not a valid configuration file anymore.


shin-:
Thanks! Addressed the issue in #4562

PR #3669, opened 2016-06-30
Docs: Typo leading to possible extreme agony

chrisclark:
The postgres image expects a specific volume path. The docs had a typo in that path. This can cause a lot of agony. This is not hypothetical agony, but very real agony that was very recently experienced :)


dnephin:
LGTM

aanand:
Thanks! LGTM

PR #3670, opened 2016-06-30
[Carry #3373] Unset env vars behavior in 'run' mirroring engine

### LINKED ISSUE #3343: Passing host env var to run results in None value within container

aanand:
We appear to have a bug on Python 3.4 where the output of `run` is sometimes blank. This warrants further investigation, but for now I've fixed the test in #3373 so it doesn't depend on `run`.

Fixes #3343

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -19,6 +19,7 @@ from ..bundle import MissingDigests
 from ..bundle import serialize_bundle
 from ..config import ConfigurationError
 from ..config import parse_environment
+from ..config.environment import Environment
 from ..config.serialize import serialize_config
 from ..const import DEFAULT_TIMEOUT
 from ..const import IS_WINDOWS_PLATFORM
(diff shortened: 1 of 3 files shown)


shin-:
LGTM :+1:

aanand:
Only test failure is a known Engine bug that's fixed in master, so merging.

PR #3689, opened 2016-07-05
Suggest to run Docker for Mac if it isn't running

bfirsh:
Instead of suggesting docker-machine.

### DIFF
--- a/compose/cli/errors.py
+++ b/compose/cli/errors.py
@@ -15,6 +15,7 @@ from requests.packages.urllib3.exceptions import ReadTimeoutError
 from ..const import API_VERSION_TO_ENGINE_VERSION
 from ..const import HTTP_TIMEOUT
 from .utils import call_silently
+from .utils import is_docker_for_mac_installed
 from .utils import is_mac
 from .utils import is_ubuntu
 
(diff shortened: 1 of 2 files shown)


compose/cli/errors.py
```diff
@@ -55,6 +56,8 @@ def handle_connection_errors(client):
             if is_ubuntu():
                 exit_with_error(docker_not_found_ubuntu)
             exit_with_error(docker_not_found_generic)
+        if is_docker_for_mac_installed():
+            exit_with_error(conn_error_docker_for_mac)
```
dnephin: Should this go above the first `if which docker ` line? Or does docker for mac not include a docker binary in the path?

bfirsh: It does include a docker binary in the path, so putting it after produces the correct behaviour:

 * If Docker for Mac is not installed, there is no binary in the path, so it'll show an error telling you to install it.
 * If Docker for Mac is installed and not running, it'll show a message telling you to start it.

bfirsh: Docker for Mac also puts Compose in the path, so first scenario is unlikely. ;)

dnephin:
Build failures are real, otherwise LGTM

bfirsh:
fixed

dnephin:
Looks like RC3 is still breaking the test suite.

LGTM (test failures are not related)

aanand:
LGTM

aanand:
Those failures are actually on 1.11 and are my fault - fix is in https://github.com/docker/compose/pull/3691.

PR #3691, opened 2016-07-06
Fix alias tests on 1.11

aanand:
The fix in https://github.com/docker/compose/pull/3652 caused a regression when testing against 1.11.


dnephin:
LGTM

PR #3692, opened 2016-07-06
Fix some typos in cli_test.py

aanand:


shin-:
I believe these have been included in another PR since git doesn't detect any changes with the current master, so I'm closing this.

PR #3697, opened 2016-07-07
Pin base image to alpine:3.4 in Dockerfile.run

aanand:
Something regressed in `alpine:edge`, causing `pip install` to fail with `invalid command 'egg_info'`.


shin-:
LGTM :+1:

PR #3705, opened 2016-07-07
Fix timeout value in error message

### LINKED ISSUE #3596: COMPOSE_HTTP_TIMEOUT is set but incorrect value is print when timeout is reached
I set COMPOSE_HTTP_TIMEOUT to 300, but when it times out the error message says:
```
An HTTP request took too long to complete. Retry with --verbose to obtain debug information.
If you encounter this issue regularly because of slow network conditions, consider setting COMPOSE_HTTP_TIMEOUT to a higher value (current value: 60).
```

I looked at the source on master and it appears to still be using the const HTTP_TIMEOUT in the message.

This is:
Docker version 1.11.2, build b9f10c9

aanand:
Also, stop checking `DOCKER_CLIENT_TIMEOUT` - it's been deprecated since 1.5.0.

Closes #3596.

### DIFF
--- a/compose/cli/docker_client.py
+++ b/compose/cli/docker_client.py
@@ -45,10 +45,6 @@ def docker_client(environment, version=None, tls_config=None, host=None,
     Returns a docker-py client configured using environment variables
     according to the same logic as the official Docker client.
     """
-    if 'DOCKER_CLIENT_TIMEOUT' in environment:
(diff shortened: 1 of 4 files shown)


aanand:
ping @dnephin

dnephin:
LGTM

PR #3706, opened 2016-07-07
Updating build-args example in compose file

jpooler:
I think providing an example here helps make it clearer that you can't just pass the build-args without the Dockerfile expecting them.

### DIFF
--- a/docs/compose-file.md
+++ b/docs/compose-file.md
@@ -74,6 +74,14 @@ This will result in an image named `webapp` and tagged `tag`, built from `./dir`
 > -   Using `build` together with `image` is not allowed. Attempting to do so
 >     results in an error.
 
+Specifying args requires using [args](#args) in the Dockerfile for consumption.
+
+     FROM $CONTAINER:$TAG
+     ARG NODE_ENV
+     ENV NODE_ENV ${NODE_ENV}
+
+In combination with passing environmental variables to Compose via an environment file or your local shell, you are now able to specify build-time values.
+
 #### context
 
 > [Version 2 file format](#version-2) only. In version 1, just use
--- a/docs/compose-file.md
+++ b/docs/compose-file.md
@@ -119,6 +119,29 @@ Add build arguments. You can use either an array or a dictionary. Any
 boolean values; true, false, yes, no, need to be enclosed in quotes to ensure
 they are not converted to True or False by the YML parser.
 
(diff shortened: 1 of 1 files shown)


jpooler:
Not really sure how to resolve the appveyor issue...seems like a pip version problem on windows?

aanand:
Thanks, this is a good idea. Maybe it'd be better if it were part of the `args` docs, rather than the higher-level `build` docs. Something like:

```md
#### args

> [Version 2 file format](#version-2) only.

Add build arguments, which are environment variables accessible only during the
build process.

First, specify the arguments in your Dockerfile:

    ARG buildno
    ARG password

    RUN echo "Build number: $buildno"
    RUN script-requiring-password.sh "$password"

Then specify the arguments under the `build` key. You can pass either a mapping
or a list:

    build:
      context: .
      args:
        buildno: 1
        password: secret

    build:
      context: .
      args:
        - buildno=1
        - password=secret

Build arguments with no value are resolved to their environment value on the
machine Compose is running on.

    args:
      - buildno
      - password

> **Note**: YAML boolean values (`true`, `false`, `yes`, `no`, `on`, `off`) must
> be enclosed in quotes, so that the parser interprets them as strings.
```

jpooler:
Modified per your suggestion, appveyor still failing.

aanand:
Looks like an internal AppVeyor problem. It doesn't matter, since this is just a docs update.

LGTM, ping @londoncalling

aanand:
Carried in #3747

sxiii:
Hey guys! Is there a way to pass .env-file variable from docker-compose up (or docker-compose build) invocation into the Dockerfile execution environment without really weird hacks? I need to pass USER NAME variable that will be used in build-time environment to become this user. I've tried to do this today and spent over 5 hours with different approaches and not a single one allows me to use the pre-made .env file user variable. It works for me only when I hardcode user name into Dockerfile, sadly. Help would be much appreciated! Thanks...
P.S. Example .env:
```
version: '3'
services:
  terra-validator:
    image: terra-validator
    env_file:
      - terra-validator/variables.env
    build:
      context: terra-validator
      dockerfile: Dockerfile
      args:
        - NEWUSER=$NEWUSER
```
Dockerfile:
```
ARG NEWUSER
RUN adduser $NEWUSER
USER $NEWUSER
CMD whoami
```
Adduser command does not work whatsoever...

PR #3711, opened 2016-07-11
Improve bash completion for `run -e`

albers:
A minor improvement: The names of environment variables will now automatically have a `=` appended, which saves typing one char and clearly indicates that you should specify a value now.

Doesn't have to go into 1.8.0. (but can, of course).


aanand:
LGTM

albers:
There is an outstanding discussion (https://github.com/docker/docker/pull/24797#issuecomment-233661093) that this change might not be a good idea. Let's s wait for the outcome.

Issue #3712, opened 2016-07-11
support mem_reservation?


yank1:
HI , Do we support mem_reservation? 
If not, what the plan?

shin-:
#4563

PR #3714, opened 2016-07-11
add memory-reservation support

yank1:
The docker 1.12 would support memory-reservation.
It's very important production feature.

Ref #3712


aanand:
Flake8 is complaining - looks like you've got an extra blank line in `service_test.py`.

LGTM once build is green.

PR #3725, opened 2016-07-12

jakubsacha:
When you execute `docker-compose up --abort-on-container-exit` you may want to have access to return code from the container that exited.

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -830,7 +830,14 @@ class TopLevelCommand(object):
 
             if cascade_stop:
                 print("Aborting on container exit...")
+
+                exit_code = 0
+                for e in self.project.containers(service_names=options['SERVICE'], stopped=True):
+                    if not e.is_running:
+                        exit_code = e.exit_code
+
                 self.project.stop(service_names=service_names, timeout=timeout)
+                sys.exit(exit_code)
 
     @classmethod
     def version(cls, options):


aanand:
This implementation is potentially buggy - it's not guaranteed that we'll get the exit code of the first container that triggered the shutdown. Perhaps instead we should raise an exception when a container exits, if `cascade_stop` is true. We can then store the exit code on that exception and read it in `main.py`.

That aside, the idea is an interesting one. @dnephin - wdyt?

dnephin:
Exiting with a non-zero exit status makes sense, but I don't think we should promise the exit code from any particular container.

> you may want to have access to return code from the container that exited.

This is only ever going to work correctly if you have a single container. "The exit status from some arbitrary container that exited" is not really a valuable exit status. 

We're better off just exiting with `1` in all cases, and letting the user query containers using `docker inspect -f ...` if they need to know a specific exit status. If for some reason one container is "special", you can use `docker-compose run` and get the exit code that way. 

This flag was really just added to restore some backwards compatibility with a previous version. The "recommended" workflow for running something and getting the exit status is `docker-compose run`.

jakubsacha:
ATM in my project I use docker-compose run, as you suggested. The problem that I have is that with it I don't see the logs from all the containers. I see just the ones from the container I specify in run command.

>  it's not guaranteed that we'll get the exit code of the first container that triggered the shutdown
What i see here is that when cascade_stop is triggered we fetch the code from the stopped container, and then stop all the others.

I was also trying to throw some exception with the exit code what looks to me as bit better solution, but sadly in the log printer we don't have access to the containers objects.

Exiting with 1 when the container stopped for me makes sense only if the container exit code is other than 0. By that, you can indicate if you have an issue with execution or not. That would actually solve my problem as well.

The project I refer to is http://github.com/trivago/rumi 
https://github.com/trivago/rumi/blob/2dc4e8f02469a487140ffae14c7a5faeb7748e34/src/Process/RunningProcessesFactory.php

So, what do you guys recommend? I'm willing to improve pull request.

jakubsacha:
@aanand @dnephin ping

dnephin:
> Exiting with 1 when the container stopped for me makes sense only if the container exit code is other than 0. 

Sounds right

aanand:
OK, so the new logic will be: as soon as any container exits, `docker-compose up --abort-on-container-exit` will exit. The exit code will be 0 if the container in question exited with code 0; otherwise it will be 1.

Correct? If so, sounds good to me. Feel free to update the PR. What I said earlier about the implementation still holds - we need to make sure that we're looking at the exact container that triggered the cascade stop.

Finally, this needs a test for both the zero and non-zero exit case.

shin-:
Closing in favor of #4466

PR #3744, opened 2016-07-19
poc: Use dobi for project tasks

dnephin:
[dobi](https://github.com/dnephin/dobi) ([install](http://dnephin.github.io/dobi/install.html)) is a tool for configuring/automating build tasks.

This PR is an example of how it can be used to build, test, and release Compose.

The [config reference](https://dnephin.github.io/dobi/config.html) and [task docs](https://dnephin.github.io/dobi/tasks.html) should hopefully cover how this can be used.

Some notable changes:
- we're able to remove a few bash scripts
- the bash scripts that remain are run in a container, so are less likely to break because of differences in host environment
- `shell` is useful for getting a shell in the `builder` image for running the scripts adhoc
- an image/container for the release scripts should make them a lot more reliable to use on different hosts

cc @aanand, @shin-


aanand:
Really cool. Excited to see it develop.

When you have a moment, could you update the relevant docs (e.g. `CONTRIBUTING.md`, `RELEASE-PROCESS.md`) so it's clear what the developer needs to run?

dnephin:
I've updated CONTRIBUTING and added the doc tasks to `dobi.yaml` as well.

I'll get RELEASE-PROCESS.md when I create the releaser image, that's still a TODO.

dnephin:
Added the release commands and updated RELEASE-PROCESS.md

dnephin:
has anyone had a change to try this out? 

cc @bfirsh as well, in case you are interested

dnephin:
rebased!

dnephin:
Updated this again, now using the latest dobi 0.7

ping @shin- @aanand, let me know if you get a chance to try it out

ijc:
PR grooming: closing on the presumption that this is stale.

PR #3748, opened 2016-07-20
tearDown the project override at the end of each test case

ijc25:
self._project.client is a docker.client.Client, so creating a new self._project
leaks (via the embedded connection pool) a bunch of Unix socket file
descriptors for each test which overrides self.project using this mechanism.

In my tests I observed the test harness using 800-900 file descriptor, which is
OK on Linux with the default limit of 1024 but breaks on OSX (e.g. with
Docker4Mac) where the default limit is only 256. The failure can be provoked on
Linux too with `ulimit -n 256`.

With this fix I have observed the process ending with ~100 file descriptors
open, including 83 Unix sockets, so I think there is likely at least one more
leak lurking.

Signed-off-by: Ian Campbell <ian.campbell@docker.com>

### DIFF
--- a/tests/acceptance/cli_test.py
+++ b/tests/acceptance/cli_test.py
@@ -114,6 +114,8 @@ class CLITestCase(DockerClientTestCase):
             for n in networks:
                 if n['Name'].startswith('{}_'.format(self.project.name)):
                     self.client.remove_network(n['Name'])
+        if hasattr(self, '_project'):
+            del self._project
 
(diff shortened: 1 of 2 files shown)


aanand:
LGTM

ijc25:
@aanand since your LGTM I added a second patch which clears up the client class attr, which was accounting for most of the remaining fds.

dnephin:
LGTM

Issue #3754, opened 2016-07-21
depends_on Should Obey {{ State.Healthcheck.Status }} Before Launching Services


midnightconman:
Is it possible for the `depends_on` parameter to wait for a service to be in a "Healthy" state before starting services, if a healthcheck exists for the container? At the moment in 1.8.0-rc2 the `depends_on` parameter only verifies that the container is in a running state, via {{ State.Status.running }}. This would allow for better dependency management in docker-compose.

aanand:
Yes, we should probably do this. It will involve:

- Adding support for both setting the health check options for a container and querying its health state to [docker-py](https://github.com/docker/docker-py)
- Adding support for those options (command, interval, retries, timeout) to the Compose file, probably under a `health_check` key
- Updating the logic of `depends_on` to query containers' health state before considering a dependency to be ready

There are some important questions to answer, e.g.

- What qualifies a _service_ (i.e. multiple containers) as healthy? Is it healthy if and only if _all_ its containers are healthy?
- Should containers not join networks until they are healthy?
- etc

midnightconman:
I like the idea of not joining a network until healthy also, that would allow for auto joining of containers via the service dns record. Is it possible to also have a mechanism where containers would leave the service dns record if the healthcheck is failing? I think this portion might be a question for docker engine.

aanand:
Yes, that's the problem with implementing it in Compose - its state can't be monitored. I think that suggests we shouldn't look at a container's health before connecting it to the network.

gittycat:
I think that the health state of a service is separate from whether it is connected to the network. A service might need to connect to other services first before it becomes healthy.

The health check should also be done at the service level, not container level. It makes more sense now to think in terms of Services than containers, specially with Docker 1.12 (Service has become a 1st class citizen). A consumer of the service should not have to care that the service is made up of 1,2 or dozens of containers. What counts is that the service as a whole is considered ready to accept requests.
This means that each service would need some centralised way for its containers to report their status to. The DNS resolver is already used to report the IP of each service. It could probably be used to keep the health status as well.

aanand:
Quoting @dnephin in https://github.com/docker/compose/pull/3815#issuecomment-237610466:

> I haven't tried out healthchecks yet. Is there some api call we can use to do "wait on container state == HEALTHY"?
>
> If that exits, I think it would make sense to use it on depends_on. I'd rather not build polling into Compose to handle it.

[...]

PR #3755, opened 2016-07-22
Healthcheck/Readiness of the container functionality POC

vkravets:
I would like to propose one of case how Healthcheck/Readiness functionality can be implemented on compose side

I faced with case when I don't have access to some of images which I using, I mean that I cannot change the content of it... Thus "wait-for-it" and "docker-size" in not an appropriate solution for me.

This is only POC which I would like to show and if it will be convenient for community it will be great...
## **Description**

This commit add new keywords on service level:

**_healthcheck-cmd_** - the main keyword which define healthcheck command which will be executed to check health ("redianes") of the service. 
Possible formats:
- service:cmd - to execute remote command
- cmd - to execute on the current service

First case helps to execute command on remote service even on external (I think). "links" and "depends_on" helps here to make sure that host was already up. In any case such check performed by code

Second case helps to prepare and make sure that current service is up and running, or check that local command successfully execute

**_healthcheck-retries_** - define how times should retry to execute the healthcheck-cmd to make sure that the host is ready (default: 2)

**_healthcheck-delay_** - define the delay between retries (default: 5)

Current behavior if healthcheck will failed in the end of retries the up will filed.

**Out of scope**:
- stop and delete containers if command was failed (is it need?)
- tests
- documentation

In any case I want to see this functionality in compose, thus I'm ready to fix all the notices which you are find =)

Enjoy and I hope that this functionality will be in the compose very soon!

Best Regards,
Vladimir

P.S: Let me know everything which I should do to see this compose =)

### DIFF
--- a/compose/config/config_schema_v2.0.json
+++ b/compose/config/config_schema_v2.0.json
@@ -82,9 +82,16 @@
         "dns": {"$ref": "#/definitions/string_or_list"},
         "dns_search": {"$ref": "#/definitions/string_or_list"},
         "domainname": {"type": "string"},
-        "healthcheck-cmd": {"type": "string"},
-        "healthcheck-retries": {"type": "string"},
-        "healthcheck-delay": {"type": "string"},
+        "healthcheck": {
+          "type": "object",
+          "properties": {
+            "cmd": {"type":"string"},
+            "delay": {"type": "number"},
+            "retries": {"type": "number"}
+          },
+          "additionalProperties": false
+
+        },
         "entrypoint": {
           "oneOf": [
             {"type": "string"},
--- a/compose/config/config_schema_v2.0.json
+++ b/compose/config/config_schema_v2.0.json
@@ -82,6 +82,9 @@
         "dns": {"$ref": "#/definitions/string_or_list"},
         "dns_search": {"$ref": "#/definitions/string_or_list"},
         "domainname": {"type": "string"},
+        "healthcheck-cmd": {"type": "string"},
+        "healthcheck-retries": {"type": "string"},
+        "healthcheck-delay": {"type": "string"},
         "entrypoint": {
           "oneOf": [
             {"type": "string"},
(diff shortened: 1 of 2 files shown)


vkravets:
Jenkins build failed due to connection issue it's not related to build process failed

aanand:
Thanks for doing this! I think the basic interface makes sense, but for the implementation we should be making use of the healthcheck functionality already implemented in Engine 1.12. See https://github.com/docker/compose/issues/3754.

vkravets:
@aanand This functionality on the engine is good but it should be implement (done) in the image, and if I don't have ability to change image or rebuild in this case very difficult to monitor readiness of the service in the sequences of image's run and on the compose side.

This functionality which I proposed is not depend on the image configuration at all and can be controlled by compose. It's more flexible solution that  engine have... But both of this functionalities can be works together and make more control of up environment by compose.

aanand:
A healthcheck can be set on any container without modifying its image. `docker run` supports the following options:

```
      --health-cmd string           Command to run to check health
      --health-interval duration    Time between running the check
      --health-retries int          Consecutive failures needed to report unhealthy
      --health-timeout duration     Maximum time to allow one check to run
```

There is therefore no advantage to re-implementing health checks outside of Engine that I can see, and plenty of disadvantages.

midnightconman:
We also probably want to overwrite healthcheck functionality in compose as a block, ie:

```
services:
  web:
    image: alpine:3.4
    healthcheck:
      interval: 30s
      timeout: 5s
      retires: 3
      command: ["nc localhost 80 || exit 2"]
```

vkravets:
@aanand Heh it's seems that I read only documentation related to dockerfile =( and skip important things fro docker run... It's great but as I saw the code it will be more complicated to implement... In any case thats good that such functionality exists for docker dun.

But it's not supported remote checking, I think we should have such ability also...

@midnightconman Yep, your example more logical... and clear

Zarbis:
@aanand Are there any plans to make Compose healthcheck-aware?

aanand:
@Zarbis That's exactly what this discussion, as well as the one in https://github.com/docker/compose/issues/3754, is about.

shin-:
We went with #4267 instead (using the Docker Engine healthcheck feature)

Issue #3611, opened 2016-06-16
"docker-compose run --rm" does not remove volumes


MihaiB:
# Observed behavior

`docker-compose run --rm` removes the container but not the volumes.

```
$ cat docker-compose.yml 
version: '2'
services:
  mongo:
    image: mongo
$ docker ps -a
CONTAINER ID        IMAGE               COMMAND             CREATED             STATUS              PORTS               NAMES
$ docker volume ls
DRIVER              VOLUME NAME
$ docker-compose run --rm mongo true
$ docker ps -a
CONTAINER ID        IMAGE               COMMAND             CREATED             STATUS              PORTS               NAMES
$ docker volume ls
DRIVER              VOLUME NAME
local               4f4cb250adccee21a56b3f843d7d0a71263f61635ec01514fdb198b767f52959
local               8cbf8df775b3d451700490dc2283f12ce3fb10f2b7383144839aa9374dc13c53
$
```
# Expected behavior

I expect `docker-compose run --rm` to remove the volumes, just like `docker run --rm` does:

```
$ docker ps -a
CONTAINER ID        IMAGE               COMMAND             CREATED             STATUS              PORTS               NAMES
$ docker volume ls
DRIVER              VOLUME NAME
$ docker run --rm mongo true
$ docker ps -a
CONTAINER ID        IMAGE               COMMAND             CREATED             STATUS              PORTS               NAMES
$ docker volume ls
DRIVER              VOLUME NAME
$ 
```
# Versions

```
$ docker version
Client:
 Version:      1.11.2
 API version:  1.23
 Go version:   go1.5.4
 Git commit:   b9f10c9
 Built:        Wed Jun  1 22:00:43 2016
 OS/Arch:      linux/amd64

Server:
 Version:      1.11.2
 API version:  1.23
 Go version:   go1.5.4
 Git commit:   b9f10c9
 Built:        Wed Jun  1 22:00:43 2016
 OS/Arch:      linux/amd64

$ docker-compose version
docker-compose version 1.7.1, build 0a9ab35
docker-py version: 1.8.1
CPython version: 2.7.9
OpenSSL version: OpenSSL 1.0.1e 11 Feb 2013
```

sds:
Relatedly, `docker-compose rm --force -v <service>` does not remove the volume either, despite the log output saying otherwise.

```
$ docker-compose --version
docker-compose version 1.7.1, build 0a9ab35
$ docker-compose up -d kafka
[...]

PR #3756, opened 2016-07-22
Remove anonymous volumes when using run --rm.

### LINKED ISSUE #2419: docker-compose run --rm does not remove volumes
Running one-off commands with docker-compose does not delete volumes used by the container. This is different than `docker run --rm` which does remove volumes after the container is deleted.

nkovacs:
Named volumes will not be removed.
This is consistent with the behavior of docker run --rm.

Fixes #2419, #3611

### DIFF
--- a/tests/acceptance/cli_test.py
+++ b/tests/acceptance/cli_test.py
@@ -984,6 +984,36 @@ class CLITestCase(DockerClientTestCase):
             [u'/bin/true'],
         )
 
+    def test_run_rm(self):
+        self.base_dir = 'tests/fixtures/volume'
+        proc = start_process(self.base_dir, ['run', '--rm', 'test'])
+        wait_on_condition(ContainerStateCondition(
+            self.project.client,
+            'volume_test_run_1',
+            'running'))
+        service = self.project.get_service('test')
+        containers = service.containers(one_off=OneOffFilter.only)
+        self.assertEqual(len(containers), 1)
+        mounts = containers[0].get('Mounts')
+        for mount in mounts:
+            if mount['Destination'] == '/container-path':
+                anonymousName = mount['Name']
+                break
+        os.kill(proc.pid, signal.SIGINT)
+        wait_on_process(proc, 1)
+
+        self.assertEqual(len(service.containers(stopped=True, one_off=OneOffFilter.only)), 0)
+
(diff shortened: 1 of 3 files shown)


aanand:
Thanks! LGTM

ping @dnephin

tests/acceptance/cli_test.py
```diff
@@ -984,6 +984,37 @@ def test_run_without_command(self):
             [u'/bin/true'],
         )
 
+    def test_run_rm(self):
+        volumes = self.client.volumes()['Volumes']
+        if volumes is None:
+            volumesBefore = 0
+        else:
+            volumesBefore = len(volumes)
```
dnephin: I don't think this is the right way to test this. It's too prone to failure. It assumes the engine being used is not in use by anything else, which is not a good assumption. The test could also pass if no volume was created, even though the remove didn't work.

A better way to test this would be to inspect the container that is created, get the volume id, and assert that the volume doesn't exist when the container exits. It's a little more involved to write, but you can do it with `start_process()` and `wait_on_process()`.

You'll need to run a non-terminating command instead of `/bin/true`.

```python

start_process(...)
wait_on_condition(ContainerCountCondition(self.project, 1))
container, = self.project.get_service('test').containers(...)
self.dispatch(["stop"])

# inspect volume by id and assert a not found error
```

tests/acceptance/cli_test.py
```diff
@@ -984,6 +984,37 @@ def test_run_without_command(self):
...
+
+        volumes = self.client.volumes()['Volumes']
+        if volumes is None:
+            volumesAfter = 0
+        else:
+            volumesAfter = len(volumes)
+
+        self.assertEqual(volumesAfter - volumesBefore, 0)
+
+    def test_run_rm_keeps_named_volume(self):
```
dnephin: Could we do this as part of the same test case (`test_run_rm`) ? A single service with both an anonymous and named volume.

nkovacs:
Done. Do you want me to squash the commits?

bfirsh:
@nkovacs yes please!

klausseiler:
Hi,
I have one question, when will this fix be added to the docker:master branch?
It's now quite a while ago when @nkovacs fixed that.

PR #3760, opened 2016-07-24
Add user agent to API calls

bfirsh:
Thanks to @mgoelzer's [excellent work in Engine](https://github.com/docker/docker/pull/21306), client user agents are now passed through to registries. This was partially so we could figure out how many people were using Compose and on what versions.

So now it's in, we should probably actually set a user agent in Compose. ;)

Pending a docker-py release so we can vendor it properly. @shin- - do you think it'd be possible to get [this change](https://github.com/docker/docker-py/pull/1125) in docker-py 1.9.0 so we can get this in Compose 1.8.0?

### DIFF
--- a/compose/cli/docker_client.py
+++ b/compose/cli/docker_client.py
@@ -10,6 +10,7 @@ from docker.utils import kwargs_from_env
 
 from ..const import HTTP_TIMEOUT
 from .errors import UserError
+from .utils import generate_user_agent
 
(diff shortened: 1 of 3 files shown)


aanand:
LGTM once docker-py version is stable

aanand:
Needs rebase now that we've updated docker-py!

bfirsh:
yipee. rebased!

PR #3762, opened 2016-07-25
Rename --fetch-digests to --push-images and remove auto-pull

### LINKED ISSUE #3720: docker-compose bundle --fetch-digests isn't a communicative name
According to one rather alarmed test user, the name of the `--fetch-digests` flag didn't make it clear that images were going to be pushed. We should perhaps come up with a better one.

aanand:
Closes #3720.

### DIFF
--- a/compose/bundle.py
+++ b/compose/bundle.py
@@ -60,7 +60,7 @@ def serialize_bundle(config, image_digests):
     return json.dumps(to_bundle(config, image_digests), indent=2, sort_keys=True)
 
 
-def get_image_digests(project, allow_fetch=False):
+def get_image_digests(project, allow_push=False):
     digests = {}
     needs_push = set()
     needs_pull = set()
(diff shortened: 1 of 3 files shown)


compose/bundle.py
```diff
@@ -108,27 +108,24 @@ def get_image_digest(service, allow_fetch=False):
...
-            raise
+
+def push_image(service):
+    try:
+        digest = service.push()
+    except:
+        log.error(
+            "Failed to push image for service '{s.name}'. Please use an "
+            "image tag that can be pushed to a Docker "
+            "registry.".format(s=service))
```
dnephin: I think it would be good to include the original error message here as well. The push could fail with an auth error, which would make this message misleading, but at least if we include the original error message they should have an idea of how to fix it.

It could also fail with some temporary connection error.

dnephin:
sounds good, just one comment about the error

compose/bundle.py
aanand: Looks like the error is already logged, because we `raise` immediately after printing this message.

dnephin:
LGTM

PR #3778, opened 2016-07-27
1.8.0 release master changes

shin-:
https://github.com/docker/compose/blob/master/project/RELEASE-PROCESS.md#if-its-a-minor-release-1x0-rather-than-a-patch-release-1xy

cc @aanand @dnephin


dnephin:
LGTM!

PR #3784, opened 2016-07-29
Fix stacktrace when handling timeout error

### LINKED ISSUE #3783: compose 1.8.2. TypeError: log_timeout_error() takes exactly 1 argument (0 given)
I have a set of containers where the mysql one has a 'weird' shell script as entrypoint.
I can start mysql after starting the container via 'docker exec ... /etc/init.d/mysql start', but when I do so, the command line hangs (I can ctrl+c to exit the docker command, and the mysql service does indeed run).

I just updated to docker 1.12 and docker-compose 1.8.0, and it seems that now handling of the 'rogue' container is getting worse: after executing the above steps, I tried 'docker-compose ps' and I get:

```
  File "compose/cli/main.py", line 61, in main
  File "compose/cli/main.py", line 113, in perform_command
  File "contextlib.py", line 35, in __exit__
  File "compose/cli/errors.py", line 56, in handle_connection_errors
TypeError: log_timeout_error() takes exactly 1 argument (0 given)
docker-compose returned -1
```

the same happens if I run `docker-compose kill -s SIGKILL`

aanand:
Reported in #3783

### DIFF
--- a/compose/cli/errors.py
+++ b/compose/cli/errors.py
@@ -53,7 +53,7 @@ def handle_connection_errors(client):
         log_api_error(e, client.api_version)
         raise ConnectionError()
     except (ReadTimeout, socket.timeout) as e:
-        log_timeout_error()
+        log_timeout_error(client.timeout)
         raise ConnectionError()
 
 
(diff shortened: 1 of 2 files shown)


shin-:
LGTM :+1:

PR #3787, opened 2016-07-29
Ubuntu: install latest release always addition

digitalist:
maybe there's a point to add an Ubuntu one-liner for installing a latest release always, without installing python etc.


digitalist:
Thanks & sorry, I forgot that docker has some advanced pull request policies.

PR #3789, opened 2016-07-29
json_splitter: Don't break when buffer contains leading whitespace.

### LINKED ISSUE #3788: json_splitter fails when leading whitespace is present in buffer.

shin-:
Fixes #3788 

cc @dnephin

### DIFF
--- a/compose/utils.py
+++ b/compose/utils.py
@@ -5,11 +5,13 @@ import codecs
 import hashlib
 import json
 import json.decoder
+import logging
 
 import six
 
 
 json_decoder = json.JSONDecoder()
+log = logging.getLogger(__name__)
 
 
 def get_output_stream(stream):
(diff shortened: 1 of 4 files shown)


compose/utils.py
```diff
@@ -60,13 +62,21 @@ def split_buffer(stream, splitter=None, decoder=lambda a: a):
...
     if buffered:
-        yield decoder(buffered)
+        try:
+            yield decoder(buffered)
+        except ValueError:
+            log.error(
+                'Compose tried parsing the following chunk as a JSON object, '
+                'but failed:\n%s' % repr(buffered)
+            )
+            raise
```
aanand: Will this still result in a stack trace being printed? If so, it'd be good to handle the error such that that doesn't happen (perhaps by raising something that `main.py` can catch and handle).

aanand:
LGTM

compose/utils.py
```diff
@@ -60,13 +64,21 @@ def split_buffer(stream, splitter=None, decoder=lambda a: a):
...
     if buffered:
-        yield decoder(buffered)
+        try:
+            yield decoder(buffered)
+        except ValueError as e:
+            log.error(
+                'Compose tried parsing the following chunk as a JSON object, '
+                'but failed:\n%s' % repr(buffered)
+            )
+            raise StreamParseError(e)
```
dnephin: I don't think this is correct. It assumes `decoder()` is always going to return a `ValueError`, but it's going to depend on what function is passed as the decoder.

Is this still necessary with the fix below?

shin-: It's not necessary for this particular issue, but if we have another failure in this area of the code in the future, it would make it easier to debug.

We can modify the `except` clause to catch any `Exception` instead?

dnephin: `Exception` would work. The message is also incorrect though. It's not necessarily `JSON`, it could be anything, depending on what gets passed as `decoder()`.

shin-: Alright, I changed the `except` clause and modified the error message to be more generic.

dnephin:
LGTM

PR #3792, opened 2016-07-30
Ubuntu: install latest release always addition

digitalist:
Signed-off-by: Aleksey Elymanov <“strangeqargo@gmail.com”>

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -41,6 +41,9 @@ which the release page specifies, in your terminal.
 
         curl -L https://github.com/docker/compose/releases/download/1.8.0/docker-compose-`uname -s`-`uname -m` > /usr/local/bin/docker-compose
 
+     The following command installs latest release:
+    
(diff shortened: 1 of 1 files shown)


digitalist:
I removed `sudo` with dd, because `sudo` usage is described earlier

aanand:
Thanks! Interesting idea, but I think it needs more consideration. If we're going to give people a command that always installs the latest version, we should arguably come up with one that works with less shell magic - e.g. by setting up stable 'latest' download URLs that redirect.

digitalist:
@aanand  it's just Linux/ubuntu related code, so, I suppose it will work ok. 
Though your idea is better.
I guess, we can't patch documentation to get latest version without using JS, and JS browser dependency is not good for automation. 
So we must switch to server-side, is there some docker server-side code repository, or I should write someone to make this  "always stable" link script?

shin-:
This should be moved to https://github.com/docker/docker.github.io where the docs have migrated!

PR #3797, opened 2016-08-01
Add note to bundle docs about requiring an experimental Engine build

aanand:
ping @londoncalling

### DIFF
--- a/docs/bundles.md
+++ b/docs/bundles.md
@@ -50,6 +50,15 @@ Wrote bundle to vossibility-stack.dab
 
 ## Creating a stack from a bundle
 
+> **Note**: Because support for stacks and bundles is in the experimental stage,
+> you need to install an experimental build of Docker Engine to use it.
+>
+> If you're on Mac or Windows, download the “Beta channel” version of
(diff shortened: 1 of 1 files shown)


londoncalling:
LGTM 👍

docs/bundles.md
```diff
@@ -50,6 +50,15 @@ Wrote bundle to vossibility-stack.dab
 
 ## Creating a stack from a bundle
 
+> **Note**: Because support for stacks and bundles is in the experimental stage,
+> you need to install an experimental build of Docker Engine to use it.
+>
+> If you're on Mac or Windows, download the “Beta channel” version of
+> [Docker for Mac](https://docs.docker.com/docker-for-mac/) or
```
sergeyklay: This is weird because there is native non-beta Docker

londoncalling: Only the Beta Docker for Mac and Docker for Windows ships with and uses the experimental Docker Engine, while the non-Beta versions are packaged with and feature map to the current released GA of Docker Engine.

PR #3807, opened 2016-08-03
allows to extend extra_hosts when using several configuration files

### LINKED ISSUE #3733: allow to extend extra_hosts when using several configuration files
According to the [documentation](https://docs.docker.com/compose/extends/#/adding-and-overriding-configuration) _docker-compose_ allows to extend a configuration by using multiple configuration files.

This has become pretty handy for me. Unfortunately it does not work for all properties as expected, the _extra_hosts_ property gets replaced instead of extended.

Please see the following MWE:

``` yaml
# conf1.yml
version: "2"
services:
    test:
        image: ubuntu:14.04
        expose:
            - "80"
        extra_hosts:
            - "foo:1.2.3.4"

# conf2.yml
version: "2"
services:
    test:
        expose:
            - "443"
        extra_hosts:
            - "bar:5.6.7.8"
```

``` bash
$ docker-compose --version
docker-compose version 1.7.1, build 6c29830
$ docker-compose -f conf1.yml -f conf2.yml config
networks: {}
services:
  test:
    expose:
    - '80'
    - '443'
    extra_hosts:
      bar: 5.6.7.8
    image: ubuntu:14.04
version: '2.0'
volumes: {}
```

The _foo_ entry gets removed in the _extra_hosts_ list, but I would have expected the list to contain both hosts, _foo_ and _bar_, just as it worked with _expose_.

Can you please check if you can modify docker-compose so it behaves the same for _expose_ and _extra_hosts_ and extend instead of replace the entries?

optimisticanshul:
This fixes #3733

Signed-off-by: Anshul Sharma <optimisticanshul@gmail.com>

### DIFF
--- a/compose/config/config.py
+++ b/compose/config/config.py
@@ -750,7 +750,7 @@ def merge_service_dicts(base, override, version):
 
     for field in [
         'ports', 'cap_add', 'cap_drop', 'expose', 'external_links',
-        'security_opt', 'volumes_from', 'depends_on',
+        'security_opt', 'volumes_from', 'depends_on', 'extra_hosts',
     ]:
         md.merge_field(field, merge_unique_items_lists, default=[])
 


aanand:
Thanks! As well as signing your commit, this could use a unit test.

PR #3814, opened 2016-08-04
Disambiguate 'Swarm' in integration doc

### LINKED ISSUE #3804: Docker Swarm and Docker Engine Swarm should be clearly distinguished in documentation
I read the overview here - https://docs.docker.com/compose/swarm/ :

> If you’re using version 2, your app should work with no changes:
...
>Once you’ve got [swarm] running, deploying your app to it should be as simple as:
> `$ eval "$(docker-machine env --swarm <name of swarm master machine>)"`
>`$ docker-compose up`

But when running docker-compose against a swarm, it provides a warning (below) and runs all docker-compose instances against the targeted master.

> WARNING: The Docker Engine you're using is running in swarm mode.
> 
> Compose does not use swarm mode to deploy services to multiple nodes in a swarm. All containers will be scheduled on the current node.
> 
> To deploy your application across the swarm, use the bundle feature of the Docker experimental build.
> 
> More info:
> https://docs.docker.com/compose/bundles

This is a request to update the docs to point folks in the right direction (which I'm guessing is the `.dab` experiment)

aanand:
Closes #3804 

ping @londoncalling

### DIFF
--- a/docs/swarm.md
+++ b/docs/swarm.md
@@ -11,6 +11,10 @@ parent="workw_compose"
 
 # Using Compose with Swarm
 
+> **Note:** “Swarm” here refers to [Docker Swarm](/swarm/overview.md), a product separate from Docker Engine. It does _not_ refer to [swarm mode](/engine/swarm), which is a built-in feature of Docker Engine introduced in version 1.12.
+>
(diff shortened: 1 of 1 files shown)


londoncalling:
@aanand LGTM, I'm going to add info to clarify this in Machine docs as well.

bfirsh:
Confusing innit.

bfirsh:
LGTM

PR #3815, opened 2016-08-04
Add support for healthcheck options in configuration

madjam002:
I'm looking into adding healthcheck support to the Docker Compile yaml file.
Please see https://github.com/docker/docker-py/pull/1152 for the corresponding pull request over at docker-py.

Both of these PRs are a WIP, there are no tests, but I'd just like to see if I am on the right tracks and want to here your thoughts.

This is working for me locally and I'm testing it with a docker compose file with 12 services in which all have healthcheck options.

An example service config is:

```
myservice:
  image: ....
  healthcheck:
    interval: 5s
    timeout: 10s
    retries: 30
    command: exit 0
```

### DIFF
--- a/compose/config/config.py
+++ b/compose/config/config.py
@@ -86,6 +86,7 @@ DOCKER_CONFIG_KEYS = [
     'volumes',
     'volumes_from',
     'working_dir',
+    'healthcheck'
 ]
 
 ALLOWED_KEYS = DOCKER_CONFIG_KEYS + [
--- a/compose/config/config_schema_v2.0.json
+++ b/compose/config/config_schema_v2.0.json
@@ -82,6 +82,16 @@
         "dns": {"$ref": "#/definitions/string_or_list"},
         "dns_search": {"$ref": "#/definitions/string_or_list"},
         "domainname": {"type": "string"},
+        "healthcheck": {
+          "type": "object",
+          "properties": {
+            "interval": {"type":"string"},
+            "timeout": {"type":"string"},
+            "retries": {"type": "number"},
+            "command": {"type": "string"}
+          },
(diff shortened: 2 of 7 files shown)


aanand:
I think this is a good thing to try out. We should get it to a prototype stage so that people in #374 can try it out and see if it works for them.

madjam002:
@aanand Sounds good, I'll carry on with it then.

Surely for #374 this doesn't need to be implemented because you can just specify `HEALTHCHECK` in `Dockerfile`? This is just for overriding it on a per-container/service basis.

aanand:
True. We should support setting `healthcheck` purely in the interests of parity with the Engine. However, it's going to be confusing if we let you specify it but it's not respected by `depends_on`.

@dnephin @shin- thoughts?

dnephin:
I haven't tried out healthchecks yet. Is there some api call we can use to do "wait on container state == HEALTHY"?

If that exits, I think it would make sense to use it on depends_on. I'd rather not build polling into Compose to handle it.

madjam002:
@dnephin I'd be interested to know this too. In a tool that I'm working on I'm currently doing this by polling `docker inspect` to see if the container is healthy, but some kind of event or hook would be better.

midnightconman:
I verified that you can use `docker event -f container=<container_id> -f event="health_status: healthy"` for this.

From comment: https://github.com/docker/compose/issues/3754#issuecomment-238103196

eslammostafa:
can we use HEALTHCHECK yet?

djui:
Any progress on this PR?

aanand:
I've opened https://github.com/docker/compose/pull/4163 to finish this work.

Issue #3790, opened 2016-07-29
Support `--build-arg` in `docker-compose build`


agilgur5:
Similarly to how Compose now supports the `run -e ENV_VAR=` option to mirror the behavior of `docker`, it would be extremely useful for Compose to also support `build --build-arg ENV_VAR=`.

This would fix some of the issues in #3608 ; without this option, it's very hard to use the `Dockerfile`'s optionally overrideable `ARG` syntax (e.g. `ARG=YOU_CAN_OVERRIDE_ME_WITH_BUILD_ARG`).
One currently has to use the `docker-compose.yml` file's `args:` option, which effectively either hard-codes the build argument or requires one to use a `.env` file in conjunction that duplicates the `Dockerfile`'s default argument.

dattran-vn01:
Hello, I wanted to learn docker-compose code and created one [pull request](https://github.com/docker/compose/pull/3836) to support this

shin-:
Closing via #4564

PR #3837, opened 2016-08-11
Support --build-arg for build command

dattran-vn01:
Support --build-arg for build command. Discussed [here](https://github.com/docker/compose/issues/3790)

Signed-off-by: Dat Tran dattran.vn01@gmail.com

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -206,18 +206,26 @@ class TopLevelCommand(object):
         e.g. `composetest_db`. If you change a service's `Dockerfile` or the
         contents of its build directory, you can run `docker-compose build` to rebuild it.
 
-        Usage: build [options] [SERVICE...]
+        Usage: build [options] [--build-arg key=val...] [SERVICE...]
 
         Options:
-            --force-rm  Always remove intermediate containers.
-            --no-cache  Do not use cache when building the image.
-            --pull      Always attempt to pull a newer version of the image.
+            --force-rm              Always remove intermediate containers.
+            --no-cache              Do not use cache when building the image.
+            --pull                  Always attempt to pull a newer version of the image.
+            --build-arg key=val     Set build-time variables for one service.
         """
+        service_names = options['SERVICE']
+        build_args = options.get('--build-arg', None)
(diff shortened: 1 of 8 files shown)


shin-:
Fixed and merged in #4564 

Thank you for your help!

prestonprice57:
HELP! I can't get this to work. It seems like 
`docker-compose build --build-arg -t customer-core-dev-container:latest .` 
should work but when I run that I get:

> Services are built once and then tagged as project_service,
> e.g. composetest_db. If you change a service's Dockerfile or the
> contents of its build directory, you can run docker-compose build to rebuild it.
> 
> Usage: build [options] [SERVICE...]
> 
> Options:
>     --force-rm  Always remove intermediate containers.
>     --no-cache  Do not use cache when building the image.
>     --pull      Always attempt to pull a newer version of the image.

I've tried various configurations for the command but I get this no matter what.

shin-:
`docker-compose version`?

prestonprice57:
I've tried with both version 2 and 3. From what I understand it should at least be supported in the latest release (3.3) right?

prestonprice57:
Actually when I try using version 3.3 it says:
`ERROR: Version in "./docker-compose.yml" is unsupported. You might be seeing this error because you're using the wrong Compose file version. Either specify a supported version ("2.0", "2.1", "3.0")`

Either way, it doesn't work with 2.0 or 3.0

shin-:
No, like, literally, what's the output of the `docker-compose version` command?

prestonprice57:
Oh duh. This is the output:

> docker-compose version 1.11.2, build dfed245
> docker-py version: 2.1.0
> CPython version: 2.7.12
> OpenSSL version: OpenSSL 1.0.2j  26 Sep 2016

shin-:
@prestonprice57 `--build-arg` was added in [1.12.0](https://github.com/docker/compose/releases/tag/1.12.0).

prestonprice57:
Oh okay that makes sense. I thought I had the latest version. Oops. Thanks for the help!

PR #3856, opened 2016-08-19
Improve Dockerfile.run image

marcosnils:
This dockerfile generates a more lightweight image that works with the
current official dynamically linked generated binaries

Signed-off-by: Marcos Lilljedahl marcosnils@gmail.com

### DIFF
--- a/Dockerfile.run
+++ b/Dockerfile.run
@@ -1,13 +1,16 @@
+FROM alpine:3.4
 
-FROM    alpine:3.4
-RUN     apk -U add \
-            python \
-            py-pip
 
-COPY    requirements.txt /code/requirements.txt
-RUN     pip install -r /code/requirements.txt
+ENV GLIBC 2.23-r3
 
-ADD     dist/docker-compose-release.tar.gz /code/docker-compose
-RUN     pip install --no-deps /code/docker-compose/docker-compose-*
 
-ENTRYPOINT ["/usr/bin/docker-compose"]
+RUN apk update && apk add --no-cache openssl ca-certificates && \
+    wget -q -O /etc/apk/keys/sgerrand.rsa.pub https://raw.githubusercontent.com/sgerrand/alpine-pkg-glibc/master/sgerrand.rsa.pub && \
+    wget https://github.com/sgerrand/alpine-pkg-glibc/releases/download/$GLIBC/glibc-$GLIBC.apk && \
+    apk add --no-cache glibc-$GLIBC.apk && rm glibc-$GLIBC.apk && \
+    ln -s /lib/libz.so.1 /usr/glibc-compat/lib/ && \
+    ln -s /lib/libc.musl-x86_64.so.1 /usr/glibc-compat/lib
+
+COPY dist/docker-compose-Linux-x86_64 /usr/local/bin/docker-compose
+
+ENTRYPOINT ["docker-compose"]
--- a/script/build/image
+++ b/script/build/image
@@ -8,9 +8,10 @@ if [ -z "$1" ]; then
 fi
 
 TAG=$1
+
 VERSION="$(python setup.py --version)"
 
 ./script/build/write-git-sha
 python setup.py sdist
-cp dist/docker-compose-$VERSION.tar.gz dist/docker-compose-release.tar.gz
+./script/build/linux
 docker build -t docker/compose:$TAG -f Dockerfile.run .


marcosnils:
@aanand version is hardcoded in the URL, you'd probably want to change it so binary is taken from python builds. I'm not familiarized with compose build process but I guess it should be easy to change

Dockerfile.run
```diff
@@ -1,13 +1,14 @@
...
-FROM    alpine:3.4
-RUN     apk -U add \
-            python \
-            py-pip
 
-COPY    requirements.txt /code/requirements.txt
-RUN     pip install -r /code/requirements.txt
+RUN apk update && apk add --no-cache \
+    curl openssl ca-certificates \
+    && curl -L https://github.com/docker/compose/releases/download/1.8.0/docker-compose-`uname -s`-`uname -m` > /usr/local/bin/docker-compose \
```
dnephin: This image is build from `script/build/image`.

Ya, we should use `COPY dist/docker-compose-Linux /usr/local/bin/docker-compose`, the problem is that we `.dockerignore` the `dist/` directory because we don't want it in the context normally. It can be quiet large and slows down other build steps.

It's too bad that `docker build` doesn't support a flag to set the `.dockerignore` file.

I think we can work around this by doing a copy of the files in `script/build/image` to a temp directory and building it from there.

sanmai-NL:
@dnephin: What the status on this?

dnephin:
The PR needs to be updated to copy the local file, not download a file that wont exist until after the release is published.

marcosnils:
@dnephin can you please let me know exactly what needs to be changed so I update the PR?. I'm not familiarized with compose build process so IDK exactly what needs to be changed.

dnephin:
I did in: https://github.com/docker/compose/pull/3856#discussion_r79014587

* in `script/build/image` copy the binary out of `dist/` (which is ignored by `.dockerignore`) and move it to some other path, so it can be added with `COPY`
* remove the `curl` replace it with a `COPY`

Possibly also cleanup the temp file that was copied in the first step

marcosnils:
@dnephin thx for the clarification. I don't see that .dockerignore is actually ignoring the `dist` directory. Is there something I'm not seeing? 

https://github.com/docker/compose/blob/master/.dockerignore

marcosnils:
@dnephin taking a closer look and AFAICS seems like `dist/` doesn't have the compose binary. If I run the `image` script, it just bundles the compose source in a tar.gz and just calls `docker build` with the Dockerfile.run dockerfile. 

For what I understand, the binary needs to the built using the standard Dockerfile and then just send it to the Dockerfile.run target so it just can be copied there. This is what I meant when I said before that I don't understand compose build process. The previous Dockerfile.run actually built the binary using python, but in this case to keep the image small we just need to copy it inside there. 

As there's no Makefile or anything, it's difficult for me to understand the actual build process.

dnephin:
Oh apparently I removed it when I set this up: https://github.com/docker/compose/commit/39cea970b8d161ce6986d5ad2f14b63cb3ff3094#diff-f7c5b4068637e2def526f9bbc7200c4e

The binary is built by `script/build/linux` you should be able to call that from `script/build/image`

marcosnils:
@dnephin PR updated, I think we have a proper build pipeline now.

Dockerfile.run
```diff
@@ -1,13 +1,15 @@
+FROM alpine:3.4
 
-FROM    alpine:3.4
-RUN     apk -U add \
-            python \
-            py-pip
 
-COPY    requirements.txt /code/requirements.txt
-RUN     pip install -r /code/requirements.txt
+COPY dist/docker-compose-Linux-x86_64 /usr/local/bin/docker-compose
```
dnephin: Please move this after the `RUN` so that we can use the build cache for the `RUN` layer.

Dockerfile.run
```diff
@@ -1,13 +1,15 @@
...
 
-ADD     dist/docker-compose-release.tar.gz /code/docker-compose
-RUN     pip install --no-deps /code/docker-compose/docker-compose-*
+RUN apk update && apk add --no-cache \
+    curl openssl ca-certificates \
+    && chmod +x /usr/local/bin/docker-compose \
```
dnephin: I don't think this is necessary. It should already have `+x`

Dockerfile.run
```diff
@@ -1,13 +1,15 @@
...
+    && wget -q -O /etc/apk/keys/sgerrand.rsa.pub https://raw.githubusercontent.com/sgerrand/alpine-pkg-glibc/master/sgerrand.rsa.pub \
+    && wget https://github.com/sgerrand/alpine-pkg-glibc/releases/download/2.23-r3/glibc-2.23-r3.apk \
+    && apk add --no-cache glibc-2.23-r3.apk && rm glibc-2.23-r3.apk \
```
dnephin: Please move the version here to an env variable:

```
export GLIBC=2.23-r3;
```

It can be used in the `wget` and `apk add` lines. That way when we upgrade the version we only need to change it in one place, not 4.

Dockerfile.run
```diff
@@ -1,13 +1,15 @@
...
+    && ln -s /lib/libz.so.1 /usr/glibc-compat/lib/ \
+    && ln -s /lib/libc.musl-x86_64.so.1 /usr/glibc-compat/lib
```
dnephin: Minor: please move all the `&&` to the end of the previous lines. That's the style we use in other places.

marcosnils: I've seen this approach in other dockerfiles i.e: https://github.com/tianon/dockerfiles/blob/master/jenkins/Dockerfile and I personally find it better. Do you still want to change it?

marcosnils:
@dnephin applied suggestions and squashed commits.

Dockerfile.run
```diff
@@ -1,13 +1,17 @@
...
+ENV GLIBC 2.23-r3
 
-ADD     dist/docker-compose-release.tar.gz /code/docker-compose
-RUN     pip install --no-deps /code/docker-compose/docker-compose-*
 
-ENTRYPOINT ["/usr/bin/docker-compose"]
+RUN apk update && apk add --no-cache \
+    curl openssl ca-certificates \
+    && wget -q -O /etc/apk/keys/sgerrand.rsa.pub https://raw.githubusercontent.com/sgerrand/alpine-pkg-glibc/master/sgerrand.rsa.pub \
+    && wget https://github.com/sgerrand/alpine-pkg-glibc/releases/download/$GLIBC/glibc-$GLIBC.apk \
```
dnephin: I missed this earlier. You're adding `curl`, but using `wget` to download. Why is that? Do we actually need both?

Dockerfile.run
dnephin: Yes please, this project uses end of line operators

Dockerfile.run
marcosnils: Seems like we don't.. might be some old stuff. I'll remove it and update the PR.

marcosnils:
@dnephin done. Hopefully we're good to go now!

Dockerfile.run
```diff
@@ -1,13 +1,16 @@
...
+RUN apk update && apk add --no-cache openssl ca-certificates && \
+    wget -q -O /etc/apk/keys/sgerrand.rsa.pub https://raw.githubusercontent.com/sgerrand/alpine-pkg-glibc/master/sgerrand.rsa.pub && \
+    wget https://github.com/sgerrand/alpine-pkg-glibc/releases/download/$GLIBC/glibc-$GLIBC.apk && \
+    apk add --no-cache glibc-$GLIBC.apk && rm glibc-$GLIBC.apk && \
+    ln -s /lib/libz.so.1 /usr/glibc-compat/lib/ && \
+    ln -s /lib/libc.musl-x86_64.so.1 /usr/glibc-compat/lib
+
+COPY dist/docker-compose-Linux-x86_64 /usr/local/bin/docker-compose
+
+ENTRYPOINT ["docker-compose"]
```
panchoh: Hi!

I suggest using the full path to the entrypoint binary, to prevent a search in the PATH, which makes it slightly faster to run.

Thanks a lot for the good work!

marcosnils: @panchoh thx for the suggestion. I believe this shoudln't be a blocking issue as PATH lookup time for a tool seems negligent for something that's used as a tool in a container, right?

panchoh: Not a blocker at all, of course.  I personally prefer less moving parts.  In fact, I usually drop the entrypoints directly onto /, and then refer to it as ENTRYPOINT ["/entrypoint"].  But as I said, just a humble suggestion.

Cheers!

schmunk42: This should also be reference in docs/installation - glad I found this, would have spend hours trying to figure out why it does not run on alpine

marcosnils:
@dnephin @shin- ping

shin-:
@marcosnils This is on my radar! Thank you for your patience.

shin-:
Rebased and merged in #4571 

Thank you!

marcosnils:
@shin- :tada: WOHO!

PR #3867, opened 2016-08-22
Use posix shell instead of bash for run.sh

ncopa:
There are no bash specific uses in the script so we can use posix shell.

Signed-off-by: Natanael Copa natanael.copa@docker.com


bfirsh:
LGTM

shin-:
Thanks!

PR #3891, opened 2016-08-31
Add swarm advisory

Vanuan:
This is confusing without it.


Vanuan:
Is related to #3814, actually

PR #3922, opened 2016-09-08
Improve volumespec parsing on windows platforms

### LINKED ISSUE #3897: Docker-compose rejects windows style volume mappings
When building windows containers a compose file like:

```
foo:
    image: windowsservercore
    volumes:
        - C:\foo:C:\bar:ro
```
Results in the error:
ERROR: Volume c:\foo:c:\bar:ro has incorrect format, should be external:internal[:mode] 

But if I call docker directly with "docker run -v C:\foo:C:\bar:ro windowsservercore" it is accepted as expected.

shin-:
Fixes #3897

### DIFF
--- a/compose/config/config.py
+++ b/compose/config/config.py
@@ -3,7 +3,6 @@ from __future__ import unicode_literals
 
 import functools
 import logging
-import ntpath
 import os
 import string
 import sys
@@ -16,6 +15,7 @@ from cached_property import cached_property
 from ..const import COMPOSEFILE_V1 as V1
 from ..const import COMPOSEFILE_V2_0 as V2_0
 from ..utils import build_string_dict
+from ..utils import splitdrive
 from .environment import env_vars_from_file
 from .environment import Environment
 from .environment import split_env
@@ -942,13 +942,7 @@ def split_path_mapping(volume_path):
     path. Using splitdrive so windows absolute paths won't cause issues with
     splitting on ':'.
     """
-    # splitdrive is very naive, so handle special cases where we can be sure
-    # the first character is not a drive.
-    if (volume_path.startswith('.') or volume_path.startswith('~') or
-            volume_path.startswith('/')):
-        drive, volume_config = '', volume_path
-    else:
-        drive, volume_config = ntpath.splitdrive(volume_path)
+    drive, volume_config = splitdrive(volume_path)
 
     if ':' in volume_config:
         (host, container) = volume_config.split(':', 1)
--- a/compose/config/types.py
+++ b/compose/config/types.py
@@ -12,6 +12,7 @@ import six
 from compose.config.config import V1
 from compose.config.errors import ConfigurationError
 from compose.const import IS_WINDOWS_PLATFORM
+from compose.utils import splitdrive
 
 
 class VolumeFromSpec(namedtuple('_VolumeFromSpec', 'source mode type')):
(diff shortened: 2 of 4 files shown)


compose/config/types.py
```diff
@@ -118,58 +119,96 @@ def normalize_paths_for_engine(external_path, internal_path):
...
 
-        external_path = external_path.replace('\\', '/')
 
-    return external_path, internal_path.replace('\\', '/')
+def splitdrive(path):
+    if len(path) == 0:
+        return ('', '')
+    if path[0] in ['.', '\\', '/', '~']:
+        return ('', path)
+    return ntpath.splitdrive(path)
```
aanand: As far as I can tell, we expand volume host paths in `compose.config.config.process_service()` before we get to `VolumeSpec.parse()`. So paths here shouldn't ever start with `.` or `~`. Or is `VolumeSpec.parse()` used in other contexts where paths might be relative?

compose/config/types.py
```diff
@@ -118,58 +119,96 @@ def normalize_paths_for_engine(external_path, internal_path):
...
         if drive:
-            external_path = '/' + drive.lower().rstrip(':') + tail
+            path = '/' + drive.lower().rstrip(':') + tail
+
+        return path.replace('\\', '/')
+
+    if external_path:
+        external_path = _normalize(external_path)
+
+    return external_path, _normalize(internal_path)
```
aanand: Is `normalize_paths_for_engine()` still a useful function? Seems like we could just move `_normalize()` up to the top level (maybe renaming it to `normalize_path_for_engine()`, singular) and invoke it directly from `VolumeSpec._parse_win32()`.

compose/config/types.py
```diff
@@ -118,58 +119,96 @@ def normalize_paths_for_engine(external_path, internal_path):
...
+                    "Volume %s has incorrect format, should be "
+                    "external:internal[:mode]" % volume_config
+                )
+            if drive:
+                parts[0] = drive + parts[0]
+            external, internal = normalize_paths_for_engine(
+                os.path.normpath(external), os.path.normpath(parts[0])
+            )
+            if len(parts) > 1:
+                mode = parts[1]
```
aanand: Looks like there's some repeated logic here where we call `splitdrive`, call `split(':')` on the tail, then prepend the drive onto the first part. Could it be factored out into a helper function?

compose/config/types.py
shin-: Yes, that shouldn't be an issue in that scenario. I think it's still nice to have a `splitdrive` function that can handle those corner cases and be used elsewhere, however, but if you feel strongly about it I can remove those from the list.

compose/config/types.py
shin-: I think you're right, we could do without it.

compose/config/types.py
shin-: I've moved `splitdrive` to `compose.utils` so we can use it in both `VolumeSpec.parse` and `compose.config.split_path_mapping`, with the latter occuring before the path expansion.

compose/config/types.py
shin-: Done!

compose/config/config.py
```diff
@@ -944,11 +944,7 @@ def split_path_mapping(volume_path):
     """
     # splitdrive is very naive, so handle special cases where we can be sure
     # the first character is not a drive.
```
aanand: I think this comment is now untrue.

PR #3942, opened 2016-09-12
Only allow log streaming if logdriver is json-file

### LINKED ISSUE #3873: logs command quits if one container uses syslog.
If you run «`docker-compose logs`» and one container's logging driver is set to syslog, compose has an unexpected behaviour.
It may show logs from some of the containers that use default json-file logging driver, but exits as soon as it reaches a container using syslog. Sometimes it shows no logs at all.

docker-compose test file:
```
version: '2'

services:
  serviceA:
    image: ubuntu:16.04
    command: echo serviceA
    logging:
      driver: syslog
      options:
        syslog-address: "tcp://localhost:514"

  serviceB:
    image: ubuntu:16.04
    command: echo serviceB
```

Tests:
```
$ docker-compose logs
Attaching to testlogs_serviceB_1, testlogs_serviceA_1
ERROR: configured logging reader does not support reading
$ docker-compose logs
Attaching to testlogs_serviceB_1, testlogs_serviceA_1
serviceB_1  | serviceB
ERROR: configured logging reader does not support reading
$ 
```

I guess it should consistently show logs from containerA and containerB, like when logging driver is set to «none»:
```
$ docker-compose logs
Attaching to testlogs_serviceA_1, testlogs_serviceB_1
serviceA_1  | WARNING: no logs are available with the 'none' log driver
serviceB_1  | serviceB
$ 
```

Please note that I did not test how it behaves with other drivers.

Versions used:
Docker version 1.12.1, build 23cf638
docker-compose version 1.8.0, build f3628c7

shin-:
Fixes #3873

### DIFF
--- a/compose/container.py
+++ b/compose/container.py
@@ -163,7 +163,7 @@ class Container(object):
     @property
     def has_api_logs(self):
         log_type = self.log_driver
-        return not log_type or log_type != 'none'
+        return not log_type or log_type in ('json-file', 'journald')
 
     def attach_log_stream(self):
         """A log stream can only be attached if the container uses a json-file
--- a/tests/unit/container_test.py
+++ b/tests/unit/container_test.py
@@ -150,6 +150,34 @@ class ContainerTest(unittest.TestCase):
         container = Container(None, self.container_dict, has_been_inspected=True)
         assert container.short_id == self.container_id[:12]
 
+    def test_has_api_logs(self):
+        container_dict = {
+            'HostConfig': {
+                'LogConfig': {
+                    'Type': 'json-file'
+                }
+            }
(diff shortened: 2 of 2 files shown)


dnephin:
According to https://docs.docker.com/engine/admin/logging/overview/

> The docker logscommand is available only for the json-file and journald logging drivers.

I think these docs have been updated since this was implemented in Compose, because I remember reading them before and seeing that it was supposed to work for most of the drivers.

So I think the check should be `in ('json-file', 'journald')`

shin-:
@dnephin PTAL :)

dnephin:
LGTM when green

PR #3943, opened 2016-09-13
Fix integration test on Docker for Mac

### LINKED ISSUE #3831: Running integration tests fail on Docker For Mac
Only one test fails:

https://github.com/docker/compose/blob/28120148f4e1380096407626f846eda8f6970f36/tests/unit/cli/errors_test.py#L35

This is probably a side effect of #3689

aanand:
There's probably a better way to test this (i.e. without substring matching), but this is the most minimal fix.

Closes #3831.


shin-:
LGTM

PR #3954, opened 2016-09-16
Force default host on windows to the default TCP host (instead of npipe)

shin-:
cc @dnephin @aanand 

docker-py 1.10 defaults to using npipes, but we don't want to make that change in Compose in a patch version. This PR overrides docker-py's default.


dnephin:
LGTM when green

PR #3955, opened 2016-09-16
Make env_file optional

### LINKED ISSUE #3560: Optional environment file (env_file)
Hi !

I'm using docker compose for my project.
All my api keys are stored in a local file (ignored by git) and I inject it into my project with `env_file` in my `docker-compose.yml`.
Most of those variables are only for production, you can run the project with the file empty.
The problem is, I can't start my container if the file doesn't exist.
It could be a good idea to add a parameter to env_file to make it optional.

Cheers !
Dorian

akalipetis:
This makes sure:
- the actual file is not necessary, meaning that you don't have to create it each time you clone (given that usually `.env` files are ignored)
- you still get a warning if the file doesn't exist, so you can be aware of times where a file should have been loaded but eventually didn't, for example due to a typo

Fix #3560 

Signed-off-by: Antonis Kalipetis akalipetis@gmail.com

### DIFF
--- a/compose/config/config.py
+++ b/compose/config/config.py
@@ -598,7 +598,10 @@ def resolve_environment(service_dict, environment=None):
     """
     env = {}
     for env_file in service_dict.get('env_file', []):
-        env.update(env_vars_from_file(env_file))
+        try:
+            env.update(env_vars_from_file(env_file))
+        except ConfigurationError as exc:
+            log.warn('{} - ignoring'.format(exc.msg))
 
     env.update(parse_environment(service_dict.get('environment')))
     return dict(resolve_env_var(k, v, environment) for k, v in six.iteritems(env))
--- a/tests/unit/config/config_test.py
+++ b/tests/unit/config/config_test.py
@@ -2936,13 +2936,13 @@ class EnvTest(unittest.TestCase):
         )
 
     def test_resolve_environment_nonexistent_file(self):
-        with pytest.raises(ConfigurationError) as exc:
+        with mock.patch('compose.config.config.log') as mock_logging:
             config.load(build_config_details(
                 {'foo': {'image': 'example', 'env_file': 'nonexistent.env'}},
                 working_dir='tests/fixtures/env'))
 
-        assert 'Couldn\'t find env file' in exc.exconly()
-        assert 'nonexistent.env' in exc.exconly()
+        assert 'Couldn\'t find env file' in mock_logging.warn.call_args[0][0]
+        assert 'nonexistent.env' in mock_logging.warn.call_args[0][0]
 
     @mock.patch.dict(os.environ)
     def test_resolve_environment_from_env_file_with_empty_values(self):


akalipetis:
I have also updated the failing test and rebased with master.

akalipetis:
Hey, is this something we'd like to see in the codebase?
If not, I could just close it 😄 

Happy new year!

akalipetis:
Closing this one, since it didn't get much attention. Feel free to ping me to reopen it and rebase it if we believe it's a good fit 😃

mariocesar:
This is useful, what it needs to be merged?

atrauzzi:
Whoa whoa, would love to see this.

lukad:
What is preventing this from getting merged?

DracoBlue:
👍 please add this!

akalipetis:
Reopening since it was requested by more people.

/cc @shin- since the discussion over at #3560 is kind of stalled but there are still people wanting this, is there anything I can do to either merge or decide against this?

dariusgm:
Are there any new updates on this MR?

derks:
Agree, this would be a nice feature to have... in our use-case, `docker-compose.yml` is built for local development to work out-of-the-box, but for different types of ci-testing (i.e. `RAILS_ENV=test` vs `RAILS_ENV=development`) we need to change/override the environment.  

Could be made backward compatible if necessary so that both of the following would work:

Default/Required:

```yaml
env_file: docker-compose.env
```

Backward/Forward Compatible:

```yaml
env_file:
    path: docker-compose.env
    required: false
```

The above is similar to how the `build` directive is handled where the default accepts just a string/path:

```yaml
build: .
```

But also accepts a nested dict:

```yaml
build:
    context: .
    args:
        SOME_BUILD_ARG=1
```

Etc...

lucasdf:
I would love to see this accepted.

esco:
@akalipetis can this be merged?

alanjds:
@dnephin 
(hope you dont mind my spam here ;)

nilbus:
This would fit my needs as implemented.

I'm of the opinion that we break backward-compatibility here. The new behavior is the one I believe most people will want, so we shouldn't require that more verbose syntax (nested `required: false`) most of the time. In cases where ENV vars are required and cannot have defaults, applications can and often will fail meaningfully in their absence. I defer to the compose team's opinion though. I'm not sure what this behavior change would require for docker-compose format versioning.

Ilyes512:
I would really like to have this as well! +1!!!

Also: This is almost open for a year? Disappointed it's not merge earlier. Creating empty .env files just to satisfy the check is not acceptable.

akalipetis:
If @shin- @dnephin or another maintainer is happy with either this (optional by default) or the proposed (dual definition, with either value or dict) implementation above, I'd be happy to rebase/update the PR and get it ready to be merged 😄

compose/config/config.py
```diff
@@ -598,7 +598,10 @@ def resolve_environment(service_dict, environment=None):
     """
     env = {}
     for env_file in service_dict.get('env_file', []):
-        env.update(env_vars_from_file(env_file))
+        try:
+            env.update(env_vars_from_file(env_file))
+        except ConfigurationError as exc:
+            log.warn('{} - ignoring'.format(exc.msg))
```
nilbus: Though unlikely, a missing file isn't the only reason a ConfigurationError could be thrown. When the path exists but is not a file (e.g. a directory), a `ConfigurationError` [will also be thrown][1]. In that case, a blocking error seems more appropriate than a passive warning.

[1]: https://github.com/akalipetis/compose/blob/651283eef96a6ab7854b3b8ee57bc25467f27f75/compose/config/environment.py#L31-L32

Instead of a `try`/`except`, perhaps we should check file existence directly.

```python
if os.path.exists(filename):
  env.update(env_vars_from_file(env_file))
else:
  log.warn('env_file {} does not exist - ignoring'.format(env_file))
```

nilbus:
**Goal**: Allow new contributors† to get set up by simply running `docker-compose up`, without required extra steps (that vary by platform) like `touch .env`.

**Maintainers**, do you agree that this is a worthy goal, in line with the project's philosophy?  
Please confirm whether you think additional effort is here is worthwhile.

If so, we might also consider also:
- combining the implementation and test commits, as per the [FAQ for contributors][]
- updating the [relevant documentation][] to explicitly document the new behavior

[FAQ for contributors]: https://docs.docker.com/opensource/FAQ/#what-is-the-pre-pull-request-checklist
[relevant documentation]: https://github.com/docker/docker.github.io/blob/master/compose/env-file.md

The most significant consideration for this PR may be the backward-compatibility issue. These are the options I see:

1. Change the behavior *without* a docker-compose.yml API version change, considering this to be a compatible change. Processes already handle missing ENV vars by providing defaults, erroring, or behaving incorrectly. Docker Compose needs not add extra protection against missing ENV vars, and it this behavior was never explicitly intended. The [documentation][relevant documentation] does not specify the behavior when the env_file does not exist.

    This runs the (likely small) risk of offending users who rely on the assumption that docker-compose will never start containers with missing ENV vars, because they believe users will always create the env_file correctly, with all required variables. Alternatively, they may rely on the missing env_file error as a reminder to create the env_file with the required variables.

2. Change the behavior, and bump the Compose file version. Provide different implementations based on the version. Bumping the API version to 2.x and 3.x will encourage users to acknowledge this change.

3. Use the [more verbose but definitely backward compatible syntax][syntax] suggested by @derks:

    ```yaml
    env_file:
        path: docker-compose.env
        required: false
    ```

[syntax]: https://github.com/docker/compose/pull/3955#issuecomment-295402087

**Maintainers**, you know the project needs best. Which option seems the most beneficial to you? We're willing to carry out the work but need a decision on the direction.

Thanks! 😄 

† _to projects that use docker-compose_

DracoBlue:
:+1:

compose/config/config.py
rosskarchner: If @akalipetis makes this change, would this be merge-able?

shin-:
The same way that `docker run --env-file doesnotexist ...` results in an error, I don't think absent environment files should be ignored by Compose. I realize this PR proposes replacing the error with a warning, but those are too often ignored or misunderstood by users to have the desired impact.

Moreover, this would make Compose files as a whole less portable by encouraging people to share them without some of their supporting files, which feels like the wrong direction to take for our project.

Given these considerations, I'm sorry to say I have to refuse merging this PR. I appreciate the work @akalipetis put into it and the people who have reviewed and proposed changes, but the cons outweigh the pros here unfortunately.

DarthNato:
We want to implement potentially several files listed on env_file, some of these may not really be required depending on the environment we are running docker compose.

I would really like to see this implemented. At the end, it is environment variables we are talking about, we do expect them to change from environment to environment. Hence I would not consider the env_file as a supporting required file.

ytilis:
Completely agree with @DarthNato 

In our case we have frontend developers who connect to the dev database, and backend developers who connect to their local db. We use .env files to override the default connection settings. It's silly that one group currently has to create a blank .env file just to avoid breaking compose, since in our case, it truly is optional and not a required supporting file depending on your environment. 

If anything, requiring an addendum to our README that you'll need to make a blank .env file in some cases makes the compose file LESS portable, not more. 

At the very least can we support the more verbose syntax of 
```
env_file:
    path: docker-compose.env
    required: false
```

That way you have to explicitly state it's an optional file which should mitigate many of the concerns expressed in rejecting this ticket.

nilbus:
@shin- We also want portability, but I think this feature is the best way to achieve it.

In my case, I want the [exercism development setup](https://github.com/exercism/exercism.io/tree/master/docker) process to be one command after cloning the git repository:

    docker-compose up

We do however want to allow overriding defaults (e.g. providing an optional API key, changing the default API URL, etc.) via environment variables. The majority of Exercism developers won't do this, yet _everyone_ has to create a local .env file.

> This would …[be] encouraging people to share [docker-compose files] without some of their supporting files.

The problem is that .env files can't easily be tracked by git and also ignore local changes. Unlike other supporting files, .env is often expected to be modified locally with values specific to the local developer.

Furthermore, there's not really a cross-platform way that I can tell people to create an empty .env file (Windows 😑). For the audience I'm writing for, it's important that I do include instruction on how to do this.

> If anything, requiring an addendum to our README that you'll need to make a blank .env file in some cases makes the compose file less portable, not more.

Right on.

@shin- With this and the proposed `required: false` syntax in mind, would you reconsider?

rthideaway:
All folks here would welcome a feature where you can override variables defined inside .env file by your custom ones without any special changes to your git workflow (like using "git assume unchanged", etc..). By introducing require flag (as @nilbus proposed) you'll get a compromise for both parties, since it ensures backward compatibility and also introducing new feature.

JDiPierro:
Is there currently any way to specify default environment variables but allow a user to override them with custom values without dirtying the git tree?

rthideaway:
@JDiPierro the best way so far would be probably to introduce multiple env files like
```
env-prod.dist
env-stg.dist
env-dev.dist
```
And just copy one of those to `env` based on environment. The file `.env` will be in .gitignore. You can also symlink .env to one of those. Additionally if you don't want to have a huge amount of variables in .env files, you can set some default values directly in docker-composer.yml, for example like this:
```
  example-container:
    image: docker-image:${EXAMPLE_CONTAINER_VERSION:-v1.1}
    container_name: ${EXAMPLE_CONTAINER_NAME:-cool-container}
    restart: unless-stopped
    ports:
      - "${EXAMPLE_CONTAINER_IP:-0.0.0.0}:${EXAMPLE_CONTAINER_HTTP_PORT:-80}:80"
      - "${EXAMPLE_CONTAINER_IP:-0.0.0.0}:${EXAMPLE_CONTAINER_HTTPS_PORT:-443}:443"
    volumes:
      - ${EXAMPLE_CONTAINER_FILES_FOO-./volumes/example-container/foo}/conf.d:/etc/service/foo
      - ${EXAMPLE_CONTAINER_FILES_BAR:-./volumes/example-container/bar}/vhost.d:/etc/service/bar
    logging:
      options:
        max-size: ${EXAMPLE_CONTAINER_LOG_MAX_SIZE:-4m}
        max-file: ${EXAMPLE_CONTAINER_LOG_MAX_FILE:-10}
    networks:
      - example-network
```
Then your local-env.dist can have just couple of those, e.g."
```
EXAMPLE_CONTAINER_HTTP_PORT=80
EXAMPLE_CONTAINER_HTTPS_PORT=443
```
And you prod-env.dist:
```
EXAMPLE_CONTAINER_VERSION=v1.1
EXAMPLE_CONTAINER_NAME=cool-container
EXAMPLE_CONTAINER_IP=127.0.0.1
EXAMPLE_CONTAINER_HTTP_PORT=8080
EXAMPLE_CONTAINER_HTTPS_PORT=4433
EXAMPLE_CONTAINER_LOG_MAX_SIZE=10
EXAMPLE_CONTAINER_LOG_MAX_FILE=25
```

Would be easier if we could use the feature mentioned in this thread.

JDiPierro:
I found in the docs that Compose will merge `docker-compose.override.yml` with your `docker-compose.yml` by default so you can gitignore the `.override.yml` file and use that to add local-only environment variables (or any other container config) without dirtying the git tree :D

compose/config/config.py
AlJohri: bumping this, if we make this change, would this be merge-able @nilbus?

atrauzzi:
That's pretty cool, but remember -- if we want to be able to have the environment variables read by things other than compose, that approach won't work.

thomasnal:
Solution for the optional `.env` file is to use string interpolation in docker compose file.  Docker compose reads `.env` file by default if available to make the variable available in the scope of `docker-compose.yml`.  With that said,

```
# docker-compose.yml
version: '3'
app:
  image: busybox
  environment:
    MY: ${MY}
```
When you login to the container now, you will see a warning,
```
docker-compose up -d
docker-compose exec app bash -Ic "echo $MY"
WARNING: The MY variable is not set. Defaulting to a blank string.

```
Create the `.env` file,
```
echo "MY=hello" > .env
docker-compose restart
```
Now you will see content of the variable in the container,
```
docker-compose exec app bash -lc "echo $MY"
hello
```

With this solution, you can choose to either add or leave out the `.env` file.  The feature is described in [The ".env" file section](https://docs.docker.com/compose/environment-variables/).

Let me know if it works for your use cases.  I've just encountered this issue and come up with the solution in the current project I am building for a company.

anujbiyani:
@thomasnal I believe that doesn't address OP's use case; he's talking about the environment file that seeds the container itself, not the one that lets you use environment variables in `docker-compose.yml`.

The confusion is because the file `.env` is used by docker-compose the way you linked, but OP is using `.env` as one of the files in the `env_file` config option (https://docs.docker.com/compose/environment-variables/#the-env_file-configuration-option).

I like @ytilis approach [above](https://github.com/docker/compose/pull/3955#issuecomment-342946651) with `required: true/false`. It addresses @shin- 	's two concerns in https://github.com/docker/compose/pull/3955#issuecomment-341231156:
> I realize this PR proposes replacing the error with a warning, but those are too often ignored or misunderstood by users to have the desired impact.

Making the required-ness explicit should be good enough to allow replacing the error with a warning.

> Moreover, this would make Compose files as a whole less portable by encouraging people to share them without some of their supporting files, which feels like the wrong direction to take for our project.

Marking an env file as not required means that it isn't needed for the Compose file to be fully portable.

leonaves:
@shin- found a hacky workaround to this, as I'm sure many others have, would request you reconsider your approach, people are going to be doing brittle things to support essentially this which will eventually break and confuse the end users (who may not be the same people who put the hack in place). At the end of the day, most users of docker-compose are not the people who set up the config, so while you might think you're protecting them, not providing graceful support to people who do configure these environments results in them being subjected to hacks.

eddiemonge:
@shin- doing in a (looks like almost monthly) checkin to see if you would reconsider if the required flag was added.

shin-:
Not up to me anymore. Please reach out to the current maintainers.

mihow:
@KlaasH Would you be able to chime in here? It looks like you signed off on this but it was never merged. Thank you!

KlaasH:
I'm not a maintainer, I just had a [somewhat-related PR](https://github.com/docker/compose/pull/6800) merged.  Though not as related as it looks.  It was about `--env-file`, the top-level option to `docker-compose`, which overrides the file to be read for values to load into the project-level environment.  This PR was about `env_file`, the service-level `docker-compose.yml` parameter that sets the file defining the environment for a specific container/service.  (And both of those are distinct from `--env-file`, the option to `docker run`, which is service-level as well, but not a `docker-compose` option.)

Though on the subject of this issue and proposed change, the project-level `.env` file might be relevant.  `docker-compose` will load an `.env` file into the project-level environment if it's there, but won't complain if there isn't one.  If the `env_file` for the service just passes through the values from the project environment, you could have that be a tracked file, or just use an inline parameter like:
```
  environment:
    - AWS_ACCESS_KEY_ID=${AWS_ACCESS_KEY_ID:-}
    - AWS_SECRET_ACCESS_KEY=${AWS_SECRET_ACCESS_KEY:-}
    - VAR_WITH_DEFAULT_VAL=${VAR_WITH_DEFAULT_VAL:-the default value}
```
Then the values in the `.env` file would be used if they're there, but as long as there's either a value or a default value for each variable, the service will be happy.  Actually, even without default values it might do the job, since I believe trying to pass an environment variable that isn't set into a service produces a warning, not a fatal error.

alexdesignworks:
@KlaasH string interpolations would work, except for the cases when images use scripts that check if the variable is set rather than if it is empty. So, with `AWS_ACCESS_KEY_ID=${AWS_ACCESS_KEY_ID:-}`, the `AWS_ACCESS_KEY_ID` variable will be set. If those images are 3rd party - there is no way to unset these empty variables. The only way would be to comment-out variable in `docker-compose.yml`.

As for using `docker-compose.override.yml` - environment variables may be set for all containers as a YAML reference like so:
```
x-environment:
  &default-environment
  XDEBUG_ENABLE: "${XDEBUG_ENABLE:-}"

services:
  cli:
    environment:
      <<: *default-environment
  php:
    environment:
      <<: *default-environment
```
It is not possible to extend the YAML reference section in `docker-compose.override.yml`.

jlivermont:
> Completely agree with @DarthNato
> 
> In our case we have frontend developers who connect to the dev database, and backend developers who connect to their local db. We use .env files to override the default connection settings. It's silly that one group currently has to create a blank .env file just to avoid breaking compose, since in our case, it truly is optional and not a required supporting file depending on your environment.
> 
> If anything, requiring an addendum to our README that you'll need to make a blank .env file in some cases makes the compose file **less** portable, not more.
> 
> At the very least can we support the more verbose syntax of
> 
> ```
> env_file:
>     path: docker-compose.env
>     required: false
> ```
> 
> That way you have to explicitly state it's an optional file which should mitigate many of the concerns expressed in rejecting this ticket.

Like others have stated, running a bunch of `touch` commands to create env files is cumbersome and annoying.

Can we reconsider @ytilis's suggestion --^ from Nov 8, 2017 to add a `required` attribute that marks an env files as required or optional?  I think that that behavior should work for everyone.  By default, missing env files cause docker-compose to error, but if marked as optional, a warning is printed and everything continues.

tomasfejfar:
One important aspect of this issue is that it already behaves this way to some extent. 

```
dev:
  build: .
```
This loads .env file if it exists. If it does not exist it does nothing. 

```
dev:
  build: .
  envfile:
    - .env
```

This loads .env if exists, blows up if it does not...

nhooey:
@ytilis suggests adding an option for `env_file` directives that let the user specify that the file is not required, so a `.env` file ignored in version control that doesn't exist will not break Docker Compose before a user creates it.

[@ytilis' suggestion](https://github.com/docker/compose/pull/3955#issuecomment-342946651):
> At the very least can we support the more verbose syntax of
> 
> ```
> env_file:
>     path: docker-compose.env
>     required: false
> ```
> 
> That way you have to explicitly state it's an optional file which should mitigate many of the concerns expressed in rejecting this ticket.

Could we get agreement on this approach so someone can get a merge request in?

@ulyssessouza: I'm tagging you since @shin-, @dnephin, and @aanand have all seem to quit the project entirely.

saivan:
I have to agree with @tomasfejfar, this is inconsistent. Sometimes we have good reasons not to provide particular env files to particular users. Eg, an env file with development and production variables.

I'm fine with the `required` parameter, but not having this functionality at all is obviously a roadblock for a lot of users. So it would be great to see it added.

umlaeute:
my ugly hack for now is to have an empty file in my repository, that is read as the `env_file` unless an `ENV_FILE` environment variable is set in the `.env` file.

~~~yaml
    env_file:
      - ${ENV_FILE:-.empty}
~~~

using "no" env_file:
~~~sh
$ grep . .empty .env
grep: .env: No such file or directory
$ docker-compose up
[...]
~~~

using `.env` as env_file:
~~~sh
$ grep . .empty .env
.env:ENV_FILE=.env
.env:MYVAR0=foo
.env:MYVAR1=bar
$ docker-compose up
[...]
~~~

using some other env_file:
~~~sh
$ grep . .empty .env prod.env
.env:ENV_FILE=prod.env
prod.env:MYVAR0=hello
prod.env:MYVAR1=goodbye
$ docker-compose up
[...]
~~~

it's hacky at best, but serves the purpose of having only a single simple file containing the environment for a container and which can be present or not (and be git-excluded)

leonaves:
@ulyssessouza tagging again as this is a no-brainer in my eyes. If people are able to hack and workaround to achieve the same end then it might as well be explicitly supported. Doing it to protect consumers of these files only leads to put them at the mercy of workarounds. The people who want to setup their projects this way will find a way to do it supported or otherwise.

tomasbedrich:
If your file is named **exactly `.env`**, you can just delete `env_file` directive, list all your variables in `environment` section (without values) and it works. 🎉  [source](https://docs.docker.com/compose/environment-variables/)

Example:
```
version: '3'
services:
  foo:
    image: foo
    build: .
    environment:
      - FLASK_RUN_PORT
      - FLASK_ENV
#    env_file: .env # <-- delete this
```

cbows:
What's the status on this? I have a list of env_files where some of them are optional. Right now, docker-compose complains that they are missing and throws an error. Current workaround is to manually `touch` missing files before.
It would help a lot if there was an option to output only a warning on missing env_file instead of an error.

JonHolman:
Agreed, I encountered this today as well and I did the same thing `touch` the missing files before running compose.  Seems ridiculous.

remydt:
I am working on a "monorepo" with a single `docker-compose.yaml` file at the root of the projet. docker-compose raises an error because I don't have `env_file` for each service. It would be perfect if `env_file` could be optional.

leonaves:
@ulyssessouza Please take a look at this, and specifically [@umlaeute's comment](https://github.com/docker/compose/pull/3955#issuecomment-576872833), this is possible with weird hacks already, so why not support it an official way, otherwise people are going to become dependent on methods not recommended by the official team that may end up breaking.

mewtlu:
Appreciate tomasbedrich's workaround to this problem but I wholeheartedly agree with leonaves and others, I don't understand why a workaround/hack which dirties either git or the compose config would be seen as a better solution than simply supporting this?

5ay3h:
IMHO there is no reason why `env_file` of a not used/initiated service will be evaluated - can we make this operation lazy(er)?

Currently all other service's params (`image`, `build`'s context, `volumes`, etc) are also treated as suggested (not being evaluated unless used) - why `env_file` should be different?


@ulyssessouza 🙏?

mmx86:
In dev environment for each service our team wants to have:
`.default.env` (default env settings)
`.custom.env` (custom overrides)

We want developer to:
1. git clone repository
2. docker-compose up
3. work
We don't want developer to:
1. Copy env files from example env files for each service.
2. Do anything else to be able start development.

We could just push into repository empty `.custom.env` files, but git doesn't allow to ignore created files. And custom overrides MUST be ignored by git.
One more option is to force every developer to run `git update-index --assume-unchanged` on these files. It is also unacceptable. And we don't want to deal with hooks.

It would be a life saver to have
```
    env_file:
      - path: service538/.default.env
        required: true
      - path: service538/.custom.env
        required: false
```

bohdanyurov-gl:
So far every single tool or service I've worked with simply assumed all env files to be optional **by default**. I mean, if you need to have them as required they should be some config files or required parameters. I've got extremely confused noticing that docker-compose has this custom behavior that puts hundreds of people into a dirty-hack area.

Yogarine:
@shin- 
> The same way that `docker run --env-file doesnotexist ...` results in an error, I don't think absent environment files should be ignored by Compose. I realize this PR proposes replacing the error with a warning, but those are too often ignored or misunderstood by users to have the desired impact.
> 
> Moreover, this would make Compose files as a whole less portable by encouraging people to share them without some of their supporting files, which feels like the wrong direction to take for our project.
> 
> Given these considerations, I'm sorry to say I have to refuse merging this PR. I appreciate the work @akalipetis put into it and the people who have reviewed and proposed changes, but the cons outweigh the pros here unfortunately.

Sorry for necrobumping, but I find the argument that this would make Compose files less portable is conjecture at best. In practice, the lack of an optional include path actually limits the portability of our project. Let me elaborate:

At work we maintain a fairly large (I think) docker-compose.yml with about 10 services. This docker-compose.yml (and related .env files) is used for all environments along our DTAP flow from development all the way up to production.

With a combination of extends and environment variable interpolation we've been able to make it so it can be easily configured to a certain environment. The development environment configuration works out of the box with sensible defaults. However, sometimes for certain external services alternate configuration needs configured. Of course, this can be done bij editing the original env files's or `docker-compose.yml`, however this wil leave local changes in git.

There is no elegant, fool-proof, convenient way to easily override environment values in files that are ignored in git by default. The only way to do this somewhat seamlessly is by using `docker-compose.override.yml` which is, in fact, optional... 🙃  
However, `docker-compose.override.yml` is only optional as long as you don't define `COMPOSE_FILE` in your `.env`, which is something I do to easily include multiple docker-compose files without needing to use a handful of `-f` options. Not to mention overriding a compose file requires knowledge about the compose spec, but we have all kinds of people running our environment from QA to junior frontend devs.

As you can see, I'm trying to make my compose setup as portable as possible for other developers, and the lack of an optional env file either requires me to limit my options, jump through strange hoops or, ultimately, resort to managing files outside of my docker-compose setup. That is the _opposite_ of compatibility.

LiquidLemon:
I really wish this got merged. We often use `.env-local` files in conjunction with the usual `.env` to allow overriding the default configuration in our projects. Having to add a line of `touch .env-local` in every CI job which builds images using `docker-compose` is *not ideal*.

chrisdlangton:
> The same way that `docker run --env-file doesnotexist ...` results in an error, I don't think absent environment files should be ignored by Compose. I realize this PR proposes replacing the error with a warning, but those are too often ignored or misunderstood by users to have the desired impact.
> 
> Moreover, this would make Compose files as a whole less portable by encouraging people to share them without some of their supporting files, which feels like the wrong direction to take for our project.
> 
> Given these considerations, I'm sorry to say I have to refuse merging this PR. I appreciate the work @akalipetis put into it and the people who have reviewed and proposed changes, but the cons outweigh the pros here unfortunately.

This issue was closed with a comment that attempts to give a rationale, and that comment has an overwhelming response of thumbs down.

The comment gives a rationale that is **not relevant** to the OP issue.

The OP issue specifies that they have the file specified by `env_file` as an ignored file. i.e. a file called `.env` is the defacto file used for local development and it is mostly always ignored by git. It will **not exist** purposefully in non-local environments (like a CI environment) because that would not be secure but mosst importantly the git repo says this file can not exist due to the ignore rule.

The comment is completely ignorant to the OP specific requirement and issue, therefore the comment is not on topic and not a rational reason to close the issue.

luxalpa:
> The same way that `docker run --env-file doesnotexist ...` results in an error, I don't think absent environment files should be ignored by Compose. I realize this PR proposes replacing the error with a warning, but those are too often ignored or misunderstood by users to have the desired impact.
> 
> Moreover, this would make Compose files as a whole less portable by encouraging people to share them without some of their supporting files, which feels like the wrong direction to take for our project.
> 
> Given these considerations, I'm sorry to say I have to refuse merging this PR. I appreciate the work @akalipetis put into it and the people who have reviewed and proposed changes, but the cons outweigh the pros here unfortunately.

A "rationale" can not be based on feelings. This is an irrationale.

Are you telling your users that their problems won't be resolved in this project and that we should use a different project? Which one do you recommend? Or are you asking us to fork this? This is confusing.

deviantintegral:
I'm normally loathe to create new issues when clearly there's active discussion elsewhere, but given there's been no updates from maintainers that I see since 2017 I've opened a new issue. Hopefully that gets on a maintainer's radar with either suggestions on alternatives, what would be required to see something like this merged, or an updated "won't fix" response.

https://github.com/docker/compose/issues/9181

jarl-dk:
Why not just add a new feature?
Along with existing (required) `env_file:`, introduce `optional_env_file:`. Like
```yaml
services:
  rails-app:
    env_file:
      - .env.development
    optional_env_file:
      - .env.development.local
```

ledlamp:
> by encouraging people to share them without some of their supporting files

Bruh I aint sharing my env file of secret keys 😤

LudovicTOURMAN:
As a workaround, I'm using that syntax (pointing by default on `/dev/null` file):
```yaml
  my-service:
    image: my-service:1.2.3
    env_file:
      - ${MY_SERVICE_ENV_FILE:-/dev/null}
```

You can then either use: `docker compose up`
or `MY_SERVICE_ENV_FILE=my-service.env docker compose up`

ericbf:
Anyone still coming here (like I did) for this feature, it is implemented now, as of Docker Compose `v2.24.0`.

See [this comment](https://github.com/docker/compose/issues/3560#issuecomment-1933045798) by [BuriedStPatrick](https://github.com/BuriedStPatrick):
> Came across this issue while googling for a solution for my highly specific use case. To anyone else coming across this issue, this IS actually implemented and no longer requires any hacks as of Docker Compose `v2.24.0`:
> 
> ```yaml
> version: 3.9
> services:
>   my-service:
>     env_file:
>       - path: /path/to/optional/.env
>         required: false # Makes the file optional. Default value is 'true'
> ```
> 
> Docs: https://docs.docker.com/compose/environment-variables/set-environment-variables/#additional-information

PR #3991, opened 2016-09-28
Add support for "isolation" in config

### LINKED ISSUE #3906: docker run --isolation parameter not available
`docker run` has a parameter [--isolation](https://docs.docker.com/engine/reference/commandline/run/#/specify-isolation-technology-for-container-isolation) that can be used to set additional isolation modes when running Docker containers. For example, `docker run --isolation=hyperv` is commonly used on Windows with Hyper-V. Given that most of the other run options are already supported such as `security_opt` and `privileged`, I think it makes sense that `isolation` should be there too.

Here's an example of how I would like to use it in docker-compose.yml:
```none
version: '2'
services:
  sql:
    image: windowsservercore
    tty: true
    stdin_open: true
#    isolation: 'hyperv'
  
  web:
    image: windowsservercore
    tty: true
    stdin_open: true
#    isolation: 'hyperv'
 
networks:
  default:
    external:
      name: nat
```

shin-:
Fixes #3906

### DIFF
--- a/compose/config/config_schema_v2.1.json
+++ b/compose/config/config_schema_v2.1.json
@@ -123,6 +123,7 @@
         "hostname": {"type": "string"},
         "image": {"type": "string"},
         "ipc": {"type": "string"},
+        "isolation": {"type": "string"},
         "labels": {"$ref": "#/definitions/list_or_dict"},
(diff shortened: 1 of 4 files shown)


aanand:
LGTM

PR #3994, opened 2016-09-29
Add support for enable_ipv6 in network definition.

### LINKED ISSUE #3430: Docker API v1.23 has broke IPv6 in docker-compose defined networks
As @gaertner-mike noticed in #2804, and due to this change in Docker API v1.23 :

> POST /networks/create now supports enabling ipv6 on the network by setting the EnableIPv6 field (doing this with a label will no longer work).

IPv6 custom networks doesn't work anymore with docker-compose, which uses a label so far to enable IPv6 (`com.docker.network.enable_ipv6: "true"`)

shin-:
Fixes #3430

### DIFF
--- a/compose/config/config_schema_v2.1.json
+++ b/compose/config/config_schema_v2.1.json
@@ -246,7 +246,8 @@
             "name": {"type": "string"}
           },
           "additionalProperties": false
-        }
+        },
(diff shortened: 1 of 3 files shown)


aanand:
LGTM

PR #4004, opened 2016-10-05
Properly merge logging dictionaries in overriding configs

### LINKED ISSUE #4000: Values for logging are not merged while using "extends"
- common.yml
```
version: '2'
services:
  common:
    network_mode: "mynetwork"
    environment:
      - PRODUCTION=true
    restart: on-failure:5
    logging:
      driver: "awslogs"
      options:
        awslogs-region: "eu-central-1"
        awslogs-group: "MY_LOGGING"
```
- docker-compose.yml
```
webapp:
    extends:
      file: common.yml
      service: common
    image: some/image:latest
    container_name: webapp
    ports:
      - "80:80"
      - "443:443"
    logging:
      options:
        awslogs-stream: "webapp"
```

After running `docker-compose up -d` I get:
```
Creating webapp
ERROR: for webapp  Cannot create container for service webapp:
unknown log opt 'awslogs-stream' for json-file log driver
```

All other values are inherited correctly (i.e. network_mode, environment and restart). Is this expected behaviour?
In real `docker-compose.yml` I have ~10 services, all of them have common logging driver (awslogs), the same region and logging group. The only difference is this "awslogs-stream".
Do I really have to repeat the same section (logging) for every service?

shin-:
Fixes #4000

### DIFF
--- a/compose/config/config.py
+++ b/compose/config/config.py
@@ -760,6 +760,8 @@ def merge_service_dicts(base, override, version):
     for field in ['dns', 'dns_search', 'env_file', 'tmpfs']:
         md.merge_field(field, merge_list_or_string)
 
+    md.merge_field('logging', merge_logging)
+
     for field in set(ALLOWED_KEYS) - set(md):
         md.merge_scalar(field)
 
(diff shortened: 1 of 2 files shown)


shin-:
@dnephin We could compare the driver in the base and in the override. If they're identical, merge both dictionaries. If they're different, use only the override options. I don't know if that works well for n>2 though.

dnephin:
Ya, that would work. The logic is a little more complicated, but it's probably more correct

shin-:
Updated and added more tests to cover all cases. PTAL :)

aanand:
LGTM! Definitely needs documenting.

PR #4010, opened 2016-10-05
updated README per vnext branch plan

londoncalling:
@johndmulhausen @nathanleclaire @shin- @aanand @dnephin  

I'll be updating this again when I have PR template in place, and we've ironed label system for okay-to-publish, and some other details, but this is a start.

If you don't like this plan in general (see your Slack message from me) or we want to tweak it, we can do so as we go along. I'm setting this up for a trial run.

Signed-off-by: Victoria Bialas <victoria.bialas@docker.com>


londoncalling:
LGTM, then, merging

Issue #3923, opened 2016-09-08
On native Windows, paths should not be "UNIXized"


shin-:
See: https://github.com/docker/docker/pull/26294
Also pointed out by @PatrickLang in https://github.com/docker/compose/issues/3897#issuecomment-244260851

1. We can try and detect the type of daemon compose is connected to and normalize paths accordingly. I don't know if this is even possible, and it creates additional network roundtrips that will make Compose feel slower in some environments.
1. We do not normalize by default, and create additional documentation to inform non-native users of the format they should employ. This would be a breaking change, but is much simpler to implement.

cc @aanand @dnephin @friism WDYT?

Relevant: https://github.com/docker/docker/pull/23569#issuecomment-226615771

friism:
ping @dgageot since I think docker for windows already attempts some of this re-writing.

dgageot:
@shin- @friism I think option 1 would have to be done on every Api clients, including Docker cli. Option 2 would work but on environments like Docker for Windows, it would make it more difficult to get rid of our intermediate Api proxy that currently has to do mount rewriting to support a windows host talking to a linux guest mounting a windows fs.

cc @justincormack

thaJeztah:
I would prefer 2.; automatically converting the target path (inside the container) looks odd to me, and it should just preserve what the user specified.

Perhaps the engine already takes Windows paths into account (I'm not running on Windows), so `/` may already be converted to `C:\` on a Windows container?

ping @jhowardmsft any ideas?

jhowardmsft:
The Windows daemon expects Windows style paths for volumes such as -v on docker run. It will reject Unix style paths such as /c/Windows/system32 etc.There are some situations where / will be treated as c:\ though in a dockerfile. What is the context of the question?

thaJeztah:
@jhowardmsft mainly, if docker-compose should try to automatically detect the type of daemon, and if Unix paths are specified as target, but the daemon is Windows, to magically try to convert `/foo/bar` into `C:\foo\bar`

jhowardmsft:
I think it would need evaluating api by api. Unfortunately the engine isn't consistent and many can accept unix style paths assuming the drive is c if omitted, as mentioned above. Volume is somewhat more restrictive, at least from the CLI. Almost all APIs filepath.fromslash the input so / or \ can be used interchangeably by and large. I don't really have enough compose experience to provide more guidance.

friism:
[...]

PR #4026, opened 2016-10-11
Do not normalize volume paths on Windows by default

shin-:
Add environment variable `COMPOSE_CONVERT_WINDOWS_PATHS` to enable normalization if needed (Toolbox / Machine users)

This is a proposed fix for #3923.

- It might break some user projects for people using Toolbox / Machine, but the old behavior can be enabled using the newly introduced environment variable.
- It would not break projects for "Docker for Windows" users, as the proxy for D4W does that path conversion already.
- It would fix the issue for Docker on Windows Server users, which can not be worked around currently.

### DIFF
--- a/compose/config/types.py
+++ b/compose/config/types.py
@@ -5,6 +5,7 @@ from __future__ import absolute_import
 from __future__ import unicode_literals
 
 import os
+import re
 from collections import namedtuple
 
 import six
@@ -13,6 +14,8 @@ from compose.config.config import V1
 from compose.config.errors import ConfigurationError
 from compose.const import IS_WINDOWS_PLATFORM
 from compose.utils import splitdrive
+
+win32_root_path_pattern = re.compile(r'^[A-Za-z]\:\\.*')
 
 
 class VolumeFromSpec(namedtuple('_VolumeFromSpec', 'source mode type')):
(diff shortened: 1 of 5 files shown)


friism:
I smoke tested this, appears to work as advertised

shin-:
Test failure seems to be unrelated.

shin-:
@dnephin @aanand 
- Do we have a better name for that environment variable?
- Does the code here look sane?

Nathan said on Slack he would be open to adding that variable to the output of `docker-machine env`, which would avoid breakage for people upgrading both products at least. We probably still need a docs callout and a clear note in the changelog.
I'll try and submit a PR for the machine change, but if we can settle on a name first it'd definitely help.

dnephin:
Maybe `COMPOSE_CONVERT_WINDOWS_PATHS` ? instead of WIN32 ?

shin-:
`s/WIN32/WINDOWS/` done

tests/unit/service_test.py
```diff
@@ -786,11 +786,11 @@ def setUp(self):
         self.mock_client = mock.create_autospec(docker.Client)
 
     def test_build_volume_binding(self):
-        binding = build_volume_binding(VolumeSpec.parse('/outside:/inside'))
+        binding = build_volume_binding(VolumeSpec.parse('/outside:/inside', True))
```
aanand: Does the `normalize` argument need to be included here (and in the other calls in the file)? It seems irrelevant.

compose/config/types.py
```diff
@@ -14,6 +15,8 @@
 from compose.const import IS_WINDOWS_PLATFORM
 from compose.utils import splitdrive
 
+win32_root_path_pattern = re.compile(r'[A-Za-z]\:\\.*')
```
aanand: This should perhaps start with `^`, although it arguably doesn't matter (a path with `C:\` somewhere in the middle is clearly invalid anyway).

aanand:
I think this is the right thing to do. Hopefully the number of people experiencing breakage will be small.

PR #4029, opened 2016-10-11
Allow hyphens and underscores in COMPOSE_PROJECT_NAME

### LINKED ISSUE #4002: Allow hyphens (dashes) in COMPOSE_PROJECT_NAME
The services defined in `docker-compose.yml` allow hyphens in their name, which means that the resulting containers can obtain the following names:

```
myproject_some-service_1
myproject_some-other-service_1
```

However, when it is attempted to use `COMPOSE_PROJECT_NAME=my-project` instead of `myproject`, the names of the services do not change to the following:

```
my-project_some-service_1
my-project_some-other-service_1
```

The cause is [this line in `command.py`](https://github.com/docker/compose/blob/267a243f80817025d2fe0cebc603939f78841736/compose/cli/command.py#L120):

``` py
 return re.sub(r'[^a-z0-9]', '', name.lower())
```

I might be wrong, but there seems to be no reason for having inconsistent naming restrictions in different parts of the resulting container name. Could be go for this?

``` py
 return re.sub(r'[^a-z0-9-]', '', name.lower())
```

Some of my projects `haveprettylongnamesanditbecomeshardtoreadthemwithouthyphens`. `this-could-be-improved`.

kachkaev:
Fixes #4002

### DIFF
--- a/compose/cli/command.py
+++ b/compose/cli/command.py
@@ -117,7 +117,7 @@ def get_project(project_dir, config_path=None, project_name=None, verbose=False,
 
 def get_project_name(working_dir, project_name=None, environment=None):
     def normalize_name(name):
-        return re.sub(r'[^a-z0-9]', '', name.lower())
+        return re.sub(r'[^a-z0-9_-]', '', name.lower())
 
     if not environment:
         environment = Environment.from_env_file(working_dir)


kachkaev:
Looks like I'll need to update some tests too – they expect `-` and `_` to be removed. See justification in #4002.

shin-:
This needs to be considered more carefully. I'm closing for now as the origin repo being deleted / private is causing trouble with our CI system.

xjewer:
so, what wrong with CI, can you reopen this pr?

PR #4030, opened 2016-10-12
Bump docker-py dependency (includes several npipe fixes)

shin-:
https://github.com/docker/docker-py/releases/tag/1.10.4


friism:
@shin- this worked with named pipe on my machine

PR #4036, opened 2016-10-17
Fix TypeError : unorderable types: str() < int()

nbarbey:
While merging list items into a set, strings and ints are compared which is not possible.


shin-:
Thank you for you contribution!

Do you mind adding a test for this?

nbarbey:
Any news on this ?

PR #4040, opened 2016-10-18
Allowing running containers to be rm'd by stop flag

rawkode:
Signed-off-by: David McKay david@rawkode.com

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -635,6 +635,7 @@ class TopLevelCommand(object):
 
         Options:
             -f, --force   Don't ask to confirm removal
+            -s, --stop    Stop the containers, if required, before removing
             -v            Remove any anonymous volumes attached to containers
             -a, --all     Deprecated - no effect.
         """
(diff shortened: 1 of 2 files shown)


rawkode:
I don't believe my changes have caused this build to fail ... I hope. Hopefully somebody can advise :+1: 

@aanand ?

shin-:
Sorry, I accidentally broke `master` yesterday. It should work now if you rebase your changes.

rawkode:
Fixed. Thanks @shin-

rawkode:
@shin- Is there anything else I can provide to improve this PR, or is there doubt to it's usefulness?

rawkode:
Hi,

I know this has been idle  for a few months, but I believe we should look at this. I'm not overly sold on there being a `--stop`, but the disparity between `--force` with `compose` and the `docker-cli` should probably be addressed.

`--force` within compose removes confirmation but won't remove a running container, whereas `docker-cli`'s `--force` will `Force the removal of a running container (uses SIGKILL)`

Thoughts?

shin-:
Sorry for the months-long silence! I think adding the `--stop` flag makes sense. We probably shouldn't change the meaning of `--force` since people have probably come to rely on it. PR as it is right now looks good to me :+1:

PR #4041, opened 2016-10-18
Enable --volumes-from when running a container

eddwardo:
This is an attempt to enable --volumes-from when running `docker-compose
run`. Added test fails when option is passed, despite the fact I added
it to cli.
I tried to figure out what is missing but I can't find the underlying
issue. I'd like to hear from someone more experienced what's missing
here.


shin-:
Sorry, I accidentally broke `master` yesterday. It should work now if you rebase your changes.

PR #4043, opened 2016-10-19
allow multi-line env vars in env files

tmornini:
This PR is likely unusable as-is, but forms a basis to tackle #3527.

I'm not a Python developer, don't understand Python development, have no idea how to run the tests, and couldn't make heads-or-tails of the said tests, so I was unable to verify this code in any way. :-(

Hoping someone can whip it into shape and get it merged!

### DIFF
--- a/compose/config/environment.py
+++ b/compose/config/environment.py
@@ -21,23 +21,47 @@ def split_env(env):
     else:
         return env, None
 
+def strip_continuation(input):
+    return v[:-1].strip()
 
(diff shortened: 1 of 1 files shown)


tmornini:
@GordonTheTurtle OK, signed and rebased on master as well.

aanand:
Thanks. Unfortunately we can't accept this PR at present. I've commented here on why: https://github.com/docker/compose/issues/3527#issuecomment-256810187

Issue #3716, opened 2016-07-11
Build on Windows: Permission denied


styfle:
Performing a docker build works fine but using the build param in a docker-compose file does not work.

I am using the latest Windows beta `docker-compose 1.8.0-rc1 build 9bf6bc6`.

## Dockerfile
```Dockerfile
FROM mhart/alpine-node:6.3.0
RUN addgroup -S nodeapp && adduser -S -g nodeapp nodeapp 
WORKDIR /app
COPY . .
RUN npm install && npm run build && npm run test

CMD ["node", "src/server.js"]
```

## package.json
```
{
  "name": "gateway-job",
  "main": "src/server.js",
  "scripts": {
    "build": "tsc",
    "test": "tape tests/*.js"
  },
  "dependencies": {
  },
  "devDependencies": {
    "tape": "^4.6.0",
    "typescript": "^1.8.10",
    "typings": "^1.3.0"
  }
}
```

## docker-compose-dev.yml
```yml
version: '2'

networks:
    mynetwork:
        driver: bridge

services:
    gateway:
        build: ../gateway
        image: gateway:1.0.0
        restart: unless-stopped
        networks:
            default: {}
            mynetwork: {}
        volumes:
            - d:/code/aconfig:/etc/nginx/conf.d
        ports:
            - 8080:80
            - 443:443
    gateway-job:
        restart: unless-stopped
        build: ../gateway-job
        image: gateway-job:1.0.0
        volumes:
            - /var/run/docker.sock:/var/run/docker.sock:ro
            - d:/code/aconfig:/etc/nginx/conf.d
        environment:
            NODE_ENV: 'production'
```

## Output
```sh
D:\code\ms\compose-config>docker-compose -f docker-compose-dev.yml up --build
Creating network "composeconfig_default" with the default driver
Creating network "composeconfig_mynetwork" with driver "bridge"
Building gateway-job
Step 1 : FROM mhart/alpine-node:6.3.0
 ---> fffb49210fed
Step 2 : RUN addgroup -S nodeapp && adduser -S -g nodeapp nodeapp
 ---> Using cache
 ---> 98ade41576dd
Step 3 : WORKDIR /app
 ---> Using cache
 ---> 29bf7298d317
Step 4 : COPY . .
 ---> e7177c69fdd9
Removing intermediate container a6cbd8105879
Step 5 : RUN npm install && npm run build && npm run test
 ---> Running in 3182e729be8a
npm WARN gateway-job@1.0.0 No repository field.

[...]

PR #4048, opened 2016-10-19
Bump 1.9.0-rc1

shin-:
**Breaking changes**

- For users of Docker Toolbox / Machine on Windows, the path conversion
  for volumes is no longer done by default. To enable it, please set the
  `COMPOSE_CONVERT_WINDOWS_PATHS` variable.

New Features

- Interactive mode for `docker-compose run` and `docker-compose exec` is
  now supported on Windows platforms.

- Introduced version 2.1 of the `docker-compose.yml` specification. Using
  this version requires to be used with Docker Engine 1.12 or above.
    - Added support for setting volume labels and network labels in
  `docker-compose.yml`.
    - Added support for the `isolation` parameter in service definitions.
    - Added support for link-local IPs in the service networks definitions.

- Added support for inline defaults in variable interpolation.

- Added support for the `group_add` and `oom_score_adj` parameters in
  service definitions.

- Added support for the `internal` and `enable_ipv6` parameters in network
  definitions.

- Compose now defaults to using the `npipe` protocol on Windows.

- Overriding a `logging` configuration will now properly merge the `options`
  mappings if the `driver` values do not conflict.

Bug Fixes

- Fixed several bugs related to `npipe` protocol support on Windows.

- Fixed an issue with Windows paths being incorrectly converted when
  using Docker on Windows Server.

- Fixed a bug where an empty `restart` value would sometimes result in an
  exception being raised.

- Fixed an issue where service logs containing unicode characters would
  sometimes cause an error to occur.

- Fixed a bug where unicode values in environment variables would sometimes
  raise a unicode exception when retrieved.


CHANGELOG.md
```diff
@@ -1,6 +1,57 @@
...
 ==========
 
+1.9.0 (2016-10-20)
+-----------------
+
+**Breaking changes**
+
+- For users of Docker Toolbox / Machine on Windows, the path conversion
+  for volumes is no longer done by default. To enable it, please set the
+  `COMPOSE_CONVERT_WINDOWS_PATHS` variable.
```
aanand: We could be more explicit here. How about:

When using Compose with Docker Toolbox/Machine on Windows, volume paths are no longer converted from `C:\Users` to `/c/Users`-style by default. To re-enable this conversion so that your volumes keep working, set the environment variable `COMPOSE_CONVERT_WINDOWS_PATHS=1`. Users of Docker for Windows are not affected and do not need to set the variable.

CHANGELOG.md
```diff
@@ -1,6 +1,57 @@
...
+
+New Features
+
+- Interactive mode for `docker-compose run` and `docker-compose exec` is
+  now supported on Windows platforms.
```
aanand: Worth mentioning that a `docker` binary is required for this to work.

CHANGELOG.md
```diff
@@ -1,6 +1,57 @@
...
+
+- Introduced version 2.1 of the `docker-compose.yml` specification. Using
+  this version requires to be used with Docker Engine 1.12 or above.
+    - Added support for setting volume labels and network labels in
+  `docker-compose.yml`.
+    - Added support for the `isolation` parameter in service definitions.
+    - Added support for link-local IPs in the service networks definitions.
+
+- Added support for inline defaults in variable interpolation.
```
aanand: Could mention the syntax here, so people don't have to go hunting in the docs:

Added support for shell-style inline defaults in variable interpolation. The supported forms are `${FOO:-default}` (fall back if FOO is unset) and `${FOO-default}` (fall back if FOO is unset or empty).

CHANGELOG.md
```diff
@@ -1,6 +1,57 @@
...
+
+- Added support for the `group_add` and `oom_score_adj` parameters in
+  service definitions.
+
+- Added support for the `internal` and `enable_ipv6` parameters in network
+  definitions.
+
+- Compose now defaults to using the `npipe` protocol on Windows.
```
aanand: Is this also a breaking change for Toolbox users on Windows?

christianbundy:
Awesome, I believe this also resolves #3716 -- thanks for moving it through the pipeline!

shin-:
Thanks for the review @aanand , CHANGELOG has been tweaked accordingly :+1:

PR #4059, opened 2016-10-23
Detect stdin redirection on TTY

jsoriano:
When using `docker-compose exec` with TTY (without `-T` flag), stdin redirection provokes an exception:

```
$ docker-compose exec server cat < docker-compose.yaml
  File "/home/jaime/src/docker-compose/venv/bin/docker-compose", line 11, in <module>
    load_entry_point('docker-compose', 'console_scripts', 'docker-compose')()
  File "/home/jaime/src/docker-compose/compose/cli/main.py", line 64, in main
    command()
  File "/home/jaime/src/docker-compose/compose/cli/main.py", line 116, in perform_command
    handler(command, command_options)
  File "/home/jaime/src/docker-compose/compose/cli/main.py", line 461, in exec_command
    pty.start()
  File "build/bdist.linux-x86_64/egg/dockerpty/pty.py", line 338, in start

  File "build/bdist.linux-x86_64/egg/dockerpty/io.py", line 32, in set_blocking
    I/O classes. open() uses the file's blksize (as obtained by os.stat) if
ValueError: file descriptor cannot be a negative integer (-1)
```

It is also reported in #3352 

This code detects that stdin is not a TTY and gives a bit more meaningful error:

```
$ docker-compose exec server cat < docker-compose.yaml
ERROR: The input device is not a TTY
```

Apart of handling this error, I wonder if another different flag should be added to control if stdin should be kept open, as `docker exec -i` does.

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -6,6 +6,7 @@ import contextlib
 import functools
 import json
 import logging
+import os
 import pipes
 import re
 import subprocess
@@ -436,6 +437,9 @@ class TopLevelCommand(object):
             args += command
 
             sys.exit(call_docker(args))
+
+        if tty and not os.isatty(sys.stdin.fileno()):
+            raise UserError("The input device is not a TTY, try using -T flag")
 
         create_exec_options = {
             "privileged": options["--privileged"],
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -417,26 +417,7 @@ class TopLevelCommand(object):
         tty = not options["-T"]
 
         if IS_WINDOWS_PLATFORM and not detach:
-            args = ["exec"]
-
-            if options["-d"]:
-                args += ["--detach"]
-            else:
-                args += ["--interactive"]
-
-            if not options["-T"]:
-                args += ["--tty"]
-
-            if options["--privileged"]:
-                args += ["--privileged"]
-
-            if options["--user"]:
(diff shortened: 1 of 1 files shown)


compose/cli/main.py
```diff
@@ -437,6 +438,9 @@ def exec_command(self, options):
 
             sys.exit(call_docker(args))
 
+        if tty and not os.isatty(sys.stdin.fileno()):
```
clintam: should we do a similar thing on the run command?

jsoriano: ummm, I didn't see that run also had a -T flag, I gave a quick look now and the implementation of the flag looks completely different.
I'll give a deeper look to see if code can be a bit unified to have a more consistent behaviour.

ijc:
@jsoriano was this by any chance fixed (or at least made obsolete) by #5580? It looks like it probably was (at least in the default case with no `COMPOSE_INTERACTIVE_NO_CLI` set)?

If not are you still interested after all this time in try to improve this? Given the time that has passed I'm going to default to know and close for now, it's easy enough to reopen if I am wrong.

jsoriano:
@ijc yes, it seems that since 1.19.0 it prints and error instead of failing with an exception:
```
$ docker-compose --version
docker-compose version 1.19.0, build 9e633ef
$ docker-compose exec apache cat < docker-compose.yml
the input device is not a TTY
```

PR #4066, opened 2016-10-24
Fix schema divergence - add missing fields to compose 2.1 schema

shin-:
PRs that were made to add fields before 2.1 was a thing and weren't ported afterwards.


aanand:
LGTM

PR #3980, opened 2016-09-26
Attach interactively on Windows by shelling out

aanand:
As discussed in https://github.com/docker/docker/issues/26705, we can shell out to `docker start --attach` instead of using dockerpty.

The change is conservatively scoped to interactive mode on Windows so that no existing use cases are broken. Once we're more confident, we can expand outwards.

TODO:

- [x] User testing ([here's the binary!](https://ci.appveyor.com/api/buildjobs/raq027fhnl0c1xjn/artifacts/dist%2Fdocker-compose-Windows-x86_64.exe))
- [x] Sensible error message when no `docker` binary is found
- [x] <s>Find out if we need to do version checking on the `docker` binary (when was `docker start --attach` introduced?)</s> We don't - it's been there since 1.0.

Closes #3194.

PR #4067, opened 2016-10-24
Replace "which" calls with the portable find_executable function

shin-:
Addresses https://github.com/docker/compose/pull/3980#issuecomment-255580302


aanand:
LGTM

PR #4070, opened 2016-10-24
Use colorama to enable colored output on Windows

shin-:
I've seen it mentioned multiple times in issues but I can't find it again for some reason - Anyway, [colorama](https://pypi.python.org/pypi/colorama) takes care of colored output on Windows.


rubenhenriques:
+1

PR #4071, opened 2016-10-25
Fix path of the parent dir of COMPOSE_FILE

NiR-:
Currently the `run.sh` tries to bind-mount the parent directory of the `COMPOSE_FILE`, but uses a relative path rather than an absolute path. It leads to the following error (with `export COMPOSE_FILE=docker/dev.yml`): 
> docker: Error response from daemon: Invalid bind mount spec "docker:docker": Invalid volume destination path: 'docker' mount path must be absolute..
See 'docker run --help'.


aanand:
Thanks! LGTM

PR #4091, opened 2016-10-29
Add support for templatable hostname

dmvk:
Hi, first of all, thanks for this amazing piece of software!

I'm trying to set up dev/test environment for our ops team, which develops salt stack formulas. I want them to be able to easily scale number of salt minion instances, which will have instance number included in hostname.

It is really neccessary to have proper hostnames here, because distributed services deployed on these minions need to be able to talk to each other (no I don't want any form of service discovery here, because we have fixed hostnames for this services in production).

I think doing this for all containers by default would be really invasive breaking change, for people already using docker-compose, so I propose making `hostname` parameter templatable. Everyone can choose, whether he wants to use instance number in hostname or not and can even choose in which format hostname will be.

Example `docker-compose.yml`, would look something like this.

```
version: '2'
services:
  master:
    build: ./images/master
    hostname: master
    volumes:
      - ./env:/www/salt/env
  minion:
    build: ./images/minion
    hostname: minion{num}
    links:
      - master
```

What you guys think? Is this solution acceptable for you?

Thanks!!

Signed-off-by: David Moravek david.moravek@firma.seznam.cz

### DIFF
--- a/compose/service.py
+++ b/compose/service.py
@@ -622,16 +622,20 @@ class Service(object):
 
         container_options.setdefault('detach', True)
 
-        # If a qualified hostname was given, split it into an
-        # unqualified hostname and a domainname unless domainname
-        # was also given explicitly. This matches the behavior of
-        # the official Docker CLI in that scenario.
-        if ('hostname' in container_options and
-                'domainname' not in container_options and
-                '.' in container_options['hostname']):
-            parts = container_options['hostname'].partition('.')
-            container_options['hostname'] = parts[0]
-            container_options['domainname'] = parts[2]
+        if ('hostname' in container_options):
+
+            # If a qualified hostname was given, split it into an
+            # unqualified hostname and a domainname unless domainname
+            # was also given explicitly. This matches the behavior of
+            # the official Docker CLI in that scenario.
+            if ('domainname' not in container_options and '.' in container_options['hostname']):
+                parts = container_options['hostname'].partition('.')
+                container_options['hostname'] = parts[0]
+                container_options['domainname'] = parts[2]
+
+            if ('{num}' in container_options['hostname']):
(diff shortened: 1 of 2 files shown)


dmvk:
Anyone?

dmvk:
Ping

shin-:
Isn't this identical to #1131? After some debate, I believe it was decided that such a feature was a bad idea (https://github.com/docker/compose/pull/1131#issuecomment-154153543)

lucacri:
+1

mdunc:
@shin- , this would actually be very useful.  At least the option to set the hostname to match the container name would help.  For example, say I have a RabbitMQ service that is configured to use DNS discovery for automatic clustering.  The service in the docker-compose file has a network alias of "rabbitmq" so that lookups to this name return the IPs of every RabbitMQ container.  So far, so good, but here's where we run in a to problem.  RabbitMQ then does a reverse DNS lookup on these IPs and then attempts to connect to each hostname that is returned.  This reverse DNS lookup returns the _container name_ and not the real hostname.  Although it can still technically connect, clustering fails because what RabbitMQ tries to connect to **must** exactly match the hostname of the container.  Since it sees that it doesn't match, it considers the node down and unavailable to cluster.

Currently, I see no way within docker-compose to override the hostname value to get it to match the container name so that reverse DNS lookups match.  And I also see no options to make reverse lookups return the container ID instead of the name which would also resolve this issue.  The way I've worked around this is by having two separate rabbitmq services - one with a manually assigned container name and hostname with an alias of "rabbitmq" and another rabbitmq service that I can scale as needed.  This way, at least one of the rabbit nodes has something that actually matches reverse DNS lookups and the cluster can be created.  But this feels like such an awful way to work around that and there's got to be a better way.

PR #4098, opened 2016-11-01
Add whitelisted driver option added by the overlay driver

shin-:
This fixes an issue when using the overlay network driver.


aanand:
LGTM

PR #4104, opened 2016-11-02
Bump docker-py

shin-:
Includes docker/docker-py#1285 which should fix #4089

PR #4105, opened 2016-11-03
Bump 1.9.0 rc3

shin-:
Bug Fixes
- Fixed several bugs related to `npipe` protocol support on Windows.
- Fixed an issue where Compose would incorrectly detect a configuration
  mismatch for overlay networks.


shin-:
cc @dnephin @aanand 

See also: #4104 #4098

PR #4112, opened 2016-11-03
Remove docs checker from Jenkinsfile and use cleanWorkspace option on wrappedNode

mikedougherty:
/cc @joaofnfernandes @shin-


joaofnfernandes:
Sweet! 👍

PR #4115, opened 2016-11-04
Call check_remote_network_config from Network.ensure

shin-:
Somehow this was missed in #4098


aanand:
Could do with a test, but that can be done separately.

LGTM

PR #4146, opened 2016-11-14
Add stop grace period to v3

dnephin:
cc @aanand

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -24,7 +24,6 @@ from ..config import ConfigurationError
 from ..config import parse_environment
 from ..config.environment import Environment
 from ..config.serialize import serialize_config
-from ..const import DEFAULT_TIMEOUT
 from ..const import IS_WINDOWS_PLATFORM
 from ..errors import StreamParseError
 from ..progress_stream import StreamOutputError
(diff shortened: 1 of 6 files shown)


aanand:
Looks like this is breaking due to an exception in [docker-py's stop() method](https://github.com/docker/docker-py/blob/a665c8c/docker/api/container.py#L402):

```
______________________ ResilienceTest.test_start_failure _______________________
tests/integration/resilience_test.py:47: in test_start_failure
    self.project.up()
compose/project.py:383: in up
    include_deps=start_deps)
compose/project.py:178: in get_services_without_duplicate
    service.remove_duplicate_containers()
compose/service.py:497: in remove_duplicate_containers
    c.stop(timeout=timeout)
compose/container.py:202: in stop
    return self.client.stop(self.id, **options)
.tox/py34/lib/python3.4/site-packages/docker/utils/decorators.py:21: in wrapped
    return f(self, resource_id, *args, **kwargs)
.tox/py34/lib/python3.4/site-packages/docker/api/container.py:402: in stop
    timeout=(timeout + (self.timeout or 0)))
E   TypeError: unsupported operand type(s) for +: 'NoneType' and 'int'
```

I don't think `stop()` likes to be passed `timeout=None`. We should probably fix that in docker-py, but we could also fix it on the Compose side?

dnephin:
hmm, ya it should never be None. I think we should always default to the constant value we have in Compose. I'll find out why that isn't happening.

dnephin:
Oh ya, the issue is that I missed calling `stop_timeout()` for this call to `stop()`. 

Should be fixed now

aanand:
LGTM

PR #4150, opened 2016-11-15
Limit testing pool to Ubuntu hosts to avoid errors with dind

shin-:
We originally thought ZFS was the issue, but it seems non-ubuntu systems cause our tests to fail consistently as well due to the `dind` container not being able to run.

cc @mikedougherty @aanand @dnephin


mikedougherty:
LGTM

PR #4151, opened 2016-11-15
Handle new pull failures behavior in Engine 1.13

shin-:
- Now returns a 404 status
- Error message has been updated (`repository nonexisting-image not found: does not exist or no read access`) (test updated to support both forms)

cc @aanand @dnephin

### DIFF
--- a/compose/service.py
+++ b/compose/service.py
@@ -10,6 +10,7 @@ from operator import attrgetter
 import enum
 import six
 from docker.errors import APIError
+from docker.errors import NotFound
 from docker.utils import LogConfig
 from docker.utils.ports import build_port_bindings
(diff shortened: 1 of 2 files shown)


thaJeztah:
This changed in https://github.com/docker/docker/pull/28235

ping @dmcgowan is that change something to be versioned in the API?

dmcgowan:
@thaJeztah these messages which were being checked originated from the registry. With the referenced change I did try to add some consistency to the messages for the not found or permissions cases. 

> is that change something to be versioned in the API?

Should probably at least be something which is documented in the API. Currently it is not and to make it worse the API sets status codes based on content of the error message, which can come from the registry.

PR #4163, opened 2016-11-16
Implement 'healthcheck' option

aanand:
Waiting on updated docker-py.

### DIFF
--- a/compose/config/config_schema_v3.0.json
+++ b/compose/config/config_schema_v3.0.json
@@ -205,12 +205,13 @@
         "interval": {"type":"string"},
         "timeout": {"type":"string"},
         "retries": {"type": "number"},
-        "command": {
+        "test": {
           "oneOf": [
             {"type": "string"},
             {"type": "array", "items": {"type": "string"}}
           ]
-        }
+        },
+        "disable": {"type": "boolean"}
       },
       "additionalProperties": false
     },
--- a/compose/config/config.py
+++ b/compose/config/config.py
@@ -17,6 +17,7 @@ from ..const import COMPOSEFILE_V2_0 as V2_0
 from ..const import COMPOSEFILE_V2_1 as V2_1
 from ..const import COMPOSEFILE_V3_0 as V3_0
 from ..utils import build_string_dict
+from ..utils import parse_nanoseconds_int
 from ..utils import splitdrive
 from .environment import env_vars_from_file
 from .environment import Environment
@@ -65,6 +66,7 @@ DOCKER_CONFIG_KEYS = [
     'extra_hosts',
     'group_add',
     'hostname',
+    'healthcheck',
     'image',
     'ipc',
     'labels',
@@ -642,11 +644,38 @@ def process_service(service_config):
     if 'extra_hosts' in service_dict:
         service_dict['extra_hosts'] = parse_extra_hosts(service_dict['extra_hosts'])
 
+    if 'healthcheck' in service_dict:
+        service_dict['healthcheck'] = process_healthcheck(
+            service_dict['healthcheck'], service_config.name)
+
     for field in ['dns', 'dns_search', 'tmpfs']:
         if field in service_dict:
             service_dict[field] = to_list(service_dict[field])
 
     return service_dict
+
+
+def process_healthcheck(raw, service_name):
+    hc = {}
+
+    if raw.get('disable'):
+        if len(raw) > 1:
+            raise ConfigurationError(
+                'Service "{}" defines an invalid healthcheck: '
+                '"disable: true" cannot be combined with other options'
+                .format(service_name))
+        hc['test'] = ['NONE']
+    elif 'test' in raw:
+        hc['test'] = raw['test']
+
+    if 'interval' in raw:
+        hc['interval'] = parse_nanoseconds_int(raw['interval'])
+    if 'timeout' in raw:
+        hc['timeout'] = parse_nanoseconds_int(raw['timeout'])
+    if 'retries' in raw:
+        hc['retries'] = raw['retries']
+
+    return hc
 
 
 def finalize_service(service_config, service_names, version, environment):
--- a/compose/service.py
+++ b/compose/service.py
@@ -17,7 +17,6 @@ from docker.utils.ports import split_port
 
 from . import __version__
 from . import progress_stream
-from . import timeparse
 from .config import DOCKER_CONFIG_KEYS
 from .config import merge_environment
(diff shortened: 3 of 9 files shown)


aanand:
Updated to validate the `disabled` option (`disabled: true` can't be combined with other options) and to move the logic from `Service` into the config loading stage.

djui:
@aanand It seems that this change is "only" affecting the behaviour of a Docker container, not the behaviour of Docker-Compose, is this observation correct? Could you explain that the implications of this change are?

- What happens with the service (and other services) while being in "starting" state
- What happens with the service (and other services) when becoming "unhealthy"

i.e. Without having an external service monitoring the healthcheck result status there are no implications to the container nor to docker-compose's behaviour right?

aanand:
@djui Your suspicions are correct. This change lets you set healthchecks on a service's containers just like with `docker run --health-cmd`, but those healthchecks do not affect Compose's behaviour at all.

ryan-blunden:
Looking forward to this.

compose/config/config_schema_v3.0.json
```diff
@@ -205,12 +205,13 @@
         "interval": {"type":"string"},
         "timeout": {"type":"string"},
         "retries": {"type": "number"},
-        "command": {
```
shin-: Why is "test" preferrable to "command"? The references I can find for engine use "command" or "cmd", not "test".

compose/config/config_schema_v3.0.json
```diff
@@ -205,12 +205,13 @@
...
+        "test": {
           "oneOf": [
             {"type": "string"},
             {"type": "array", "items": {"type": "string"}}
           ]
-        }
+        },
+        "disable": {"type": "boolean"}
```
shin-: Should this be `disabled`?

shin-: Also, if we're adding this option, I don't think we need to support the `null` type for the healthcheck section anymore.

compose/config/config_schema_v3.0.json
vdemeester: @shin- https://github.com/docker/docker/blob/ac9226d8435316f45c123b51b9dad4da4d02ad81/api/types/container/config.go#L11

yaakov-berkovitch:
Are/Will the health command available through the compose file for stacks ?

itaylor:
I've been following this issue (and it's predecessors) for quite some time, as I think it will greatly benefit my usage of docker-compose and allow me to remove a bunch of hacky junk I have to do to get my services to start in the correct order.

I was happy to read in the release notes https://github.com/docker/compose/releases
That we're able to use it with the new 2.1 version of the compose file spec in version 1.10 of compose

However, there's a line in there:
> *Note:* This feature will not be ported to version 3 Compose files.

Could someone help me understand the reasoning behind this feature being released and immediately orphaned?  Is there something else in version 3 that replaces its functionality?

yaakov-berkovitch:
Very strange, because i'm using current in experimental version 1.13 rc4 with version 3 of compose file.
So ??

vdemeester:
> Could someone help me understand the reasoning behind this feature being released and immediately orphaned? Is there something else in version 3 that replaces its functionality?

The note is on `depends_on`, not on `healthcheck` though

shin-:
version 3 is the format to be used with the `docker stack` commands. It's using a different approach to providing the features `docker-compose` currently provides, and as such it's unlikely that healthcheck dependency will be implemented in version 3. 

On the other hand, version 2.x will continue to be updated (for the time being) and it's completely fine to keep using it if it fits your application stack better.

patrickleet:
Can someone point me to an example of this?

I have a healthcheck defined in my service A, and another service B that runs integration tests. service B requires service A to be running and healthy.

I'm using depends_on to depend on service A from service B, however, service B is not waiting for a healthy status, resulting in my tests to fail.

Here is my docker-compose file:

version: '2.1'

networks: 
  testing:

services:

  graphql-service-test:
    build:
      context: .
      dockerfile: Dockerfile.dev
    image: graphql-service-test
    ports:
      - 3010:3010
    volumes:
      - .:/home/service/
      - /home/service/node_modules/
    networks:
      - testing

  unit:
    extends:
      service: graphql-service-test
    command: npm run test

  staging: # the name of the service 
    build:
      context: .
      dockerfile: Dockerfile.dev
    volumes:
      - .:/home/service/
      - /home/service/node_modules/
    command: npm run test:staging
    environment:
      - GRAPHQL_URI=http://graphql-service-test:3010/graphql
    networks:
      - testing
    depends_on:
      - graphql-service-test

Niceplace:
@patrickleet did you manage to solve your issue ? I'm having the same kind of problem, here's my docker-compose.

```
---
version: "2.1"

services:
  rancher-server:
    restart: always
    image: rancher/server:stable
    container_name: rancher_server
    depends_on:
      - mysql
    healthcheck:
      test: ["CMD", "mysqladmin", "ping", "-h mysql","--silent"]
      interval: 1s
      timeout: 10s
      retries: 5
    networks:
      - rancher_network
    environment:
       - http_proxy="http://proxy.example.com:1337"
       - https_proxy="http://proxy.example.com:1337"
       - no_proxy="localhost,127.0.0.1,192.168.6.0/25"
       - NO_PROXY="localhost,127.0.0.1,192.168.6.0/25"
       - CATTLE_DB_CATTLE_DATABASE=mysql
       - CATTLE_DB_CATTLE_MYSQL_HOST=mysql
       - CATTLE_DB_CATTLE_MYSQL_NAME=cattle
       - CATTLE_DB_CATTLE_MYSQL_PORT=3306
       - CATTLE_DB_CATTLE_USERNAME=rancheruser
       - CATTLE_DB_CATTLE_PASSWORD=rancherpass
    ports:
      - "11080:8080"

  mysql:
    restart: always
    image: sameersbn/mysql:latest
    container_name: mysql
    networks:
      - rancher_network
    environment:
      - 'DB_NAME=cattle'
      - 'DB_USER=rancheruser'
      - 'DB_PASS=rancherpass'
      - 'MYSQL_CHARSET=utf8'
      - 'MYSQL_COLLATION=utf8_general_ci'
    expose:
      - "3306"
    volumes:
      - /app/docker-volumes/rancher/mysql/data:/var/lib/mysql:rw

```

inadarei:
Please note that the last example doesn't work anymore because "Version 3 no longer supports the condition form of depends_on." Back to the shell script: https://docs.docker.com/compose/startup-order/

Issue #4178, opened 2016-11-22
docker-compose is lacking pids-limit 


onorua:
Apparently pids-limit is not implemented at all, it is not mentioned neither in config_schema nor in source code. 
I've expected it to be on the same level as `shm_size` , is there a way to make it work or we need to create PR for it?

shin-:
Hi!

This would indeed require a PR.

disconnect3d:
Just for other people who would get here - the `pids_limit: X` in `docker-compose.yml` doesn't work for compose version 3. It works for 2.1, see https://github.com/docker/compose/issues/4792 for more info.

PR #4180, opened 2016-11-23
introducing pids_limit, fixing #4178

onorua:
Version 2.1+ will have pids_limit functionality, we could not make 2.0 because it is using < 1.23 API version.


shin-:
Thanks! Could you add some tests as well?

onorua:
sure, added

shin-:
Rebased and merged in #4565 

Thanks!

PR #4219, opened 2016-12-06
Use docker SDK 2.0

shin-:
`docker==2.0.0` has been released.


shin-:
Tests are failing because we don't have `dockerswarm/dind:1.12.4-rc1`.

PR #4226, opened 2016-12-06
Increase minimum version for compose format v3

dnephin:
We'd like to add some features (notably secrets) to v3 that will require the v1.25 API version so we need to require it now.

cc @aanand, @shin-


aanand:
LGTM

Issue #4047, opened 2016-10-19
docker-compose terminates if build directory is missing


hholst80:
Summary:

`docker-compose` seems to pre-check that all build directories exists before running any commands. Even for commands where the build directory is not needed, like `docker-compose ps` or `docker-compose run ...`

Command:

```
docker-compose ps
```

Result:

```
ERROR: build path /omnicoder either does not exist, is not accessible, or is not a valid URL.
```

Expected result:

```
        Name                       Command               State              Ports
---------------------------------------------------------------------------------------------
api-server_1   npm start                        Up      3000/tcp
cron_1         /usr/sbin/crond -f -d 8 -c ...   Up
disbatcher_1   /usr/bin/tini -- python -m ...   Up
mongodb_1      /entrypoint.sh mongod            Up      27017/tcp
nginx_1        nginx -g daemon off;             Up      0.0.0.0:443->443/tcp, 80/tcp
```

trodrigues:
Just ran into this issue myself, in a situation where someone was trying to run a service and getting an error about a directory not existing.

The file in question defines multiple services. One of these services can run just this application (a REST API) and another one can run this application, and also build and run an image for an UI which talks to this API.

The person in question wanted only the API so they ran the service which starts only the API. But because there is a definition for that other service which runs both API and UI in the same file, and because this person didn't have the directory where the UI is built from, they got an error, preventing them from running their desired service.

cyberjoac:
+1

CodeCorrupt:
I'm currently running into this issue with Jenkins deployment. I use the  `docker-compos build` command to build the images locally, then push them to the registry. From there I go to the production server and want to `docker-compose up` copying over only the compose file and I get this error.

I'm getting around the issue by simply creating an empty directory. but it's a rather clunky work around.

afeilulu:
+1

italktocomputers:
I can confirm this issue is present in docker-compose version 1.23.2, build 1110ad0.

yurii-piets:
I'm running into this same issue.

loeffel-io:
same here

ariebrainware:
Any update?

cbrunnkvist:
Same.

> I'm getting around the issue by simply creating an empty directory. but it's a rather clunky work around.

Some of my (monorepo-) services have their build context outside (`../xxx/yyy`) of where their compose files live, so this make any such workaround even more clunky for me currently. 

[...]

PR #4259, opened 2016-12-17
Fixes https://github.com/docker/compose/issues/4047

hholst80:


ijc:
PR grooming: closing as stale.

Thanks you for your submission and sorry that it never got looked at. If you are interested in continuing with this then I think it needs at least a more verbose commit message explaining the issue and how the fix addresses it as well as some tests for the specific behaviour added.

However after such a long time I'm going to default to assuming you have moved on and close. We can reopen easily enough if I'm wrong.

PR #4288, opened 2017-01-04
Fix 404 issue, change APIError to more accureate ImageNotFound

tntC4stl3:
Actually this issue was found in docker-compose v1.9.0, when running an ansible-playbook, but after checking the latest code, I think it still occurs.

The reason is the [code](https://github.com/docker/docker-py/blob/738cfdcdf96e7ff56f6bb3a4966e337187ba51c4/docker/errors.py#L23) to get images in `docker-py` has been changed for a long time. It doesn't just raise `APIError` for all exceptions, but check the response code and response content to raise two more exceptions `ImageNotFound` and `NotFound`.

So if we still check the `APIError`, we will never raise the `NoSuchImageError` and make other applications confused to use it.

### DIFF
--- a/compose/service.py
+++ b/compose/service.py
@@ -10,6 +10,7 @@ from operator import attrgetter
 import enum
 import six
 from docker.errors import APIError
+from docker.errors import ImageNotFound
 from docker.errors import NotFound
 from docker.types import LogConfig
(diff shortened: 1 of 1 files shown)


tntC4stl3:
Anything else needs me to do before merging this PR?

PR #4289, opened 2017-01-04
Fixing bash path problem.

allandequeiroz:
Fixing bash path problem, avoiding '**docker-compose-completion.sh: bash: bad interpreter: No such file or directory**' error

### DIFF
--- a/contrib/completion/bash/docker-compose
+++ b/contrib/completion/bash/docker-compose
@@ -1,4 +1,4 @@
-#!bash
+#!/bin/bash
 #
 # bash completion for docker-compose
 #


albers:
The completion script is intended to be sourced (e.g. `source docker-compose-completion.sh`).
If the script is executed as a command, it will have no effect at all because everything it contributes to the environment will be gone once it terminates.
So the problem that you reported seems to be an indicator of wrong usage.

The hasbang is only present in this file as a hint for editors.
I agree that the path should be correct (even if bash is not executed), so :+1: for your change. Thanks!

shin-:
Thank you!

Issue #4240, opened 2016-12-12
Error when mounting host path with Docker for Windows when using different DOCKER_HOST


giggio:
When using a different `DOCKER_HOST`, unless you set `COMPOSE_CONVERT_WINDOWS_PATHS=0`, any mount you specify on your docker-compose file will fail with:

````
ERROR: for db  Cannot create container for service db: create <volume>: "<volume>" includes invalid characters for a local volume name, only "[a-zA-Z0-9][a-zA-Z0-9_.-]" are allowed
ERROR: Encountered errors while bringing up the project.
````

This is **really** confusing. This used to work fine before.
I am on `1.9.0` build `2585387`.
`DOCKER_HOST` is set to a remote linux server.
Volumes is:

````yml
    volumes:
      - /opt/data/foo/html/:/var/www/html
````

Only works when `COMPOSE_CONVERT_WINDOWS_PATHS=0` is set or when it is set to `COMPOSE_CONVERT_WINDOWS_PATHS=1`. In other words, it only fails if `COMPOSE_CONVERT_WINDOWS_PATHS` is unset.

On the [1.9.0 release](https://github.com/docker/compose/releases/tag/1.9.0) you mention that:

> When using Compose with Docker Toolbox/Machine on Windows, volume paths are no longer converted from C:\Users to /c/Users-style by default. To re-enable this conversion so that your volumes keep working, set the environment variable COMPOSE_CONVERT_WINDOWS_PATHS=1. Users of Docker for Windows are not affected and do not need to set the variable.

The last part `Users of Docker for Windows are not affected and do not need to set the variable` is **not** true.

shin-:
Thank you for the report.

Is your `DOCKER_HOST` pointing to a remote Linux server?
And when you see the error, is your `COMPOSE_CONVERT_WINDOWS_PATHS` variable unset, or what value is it set to?
Finally, what happens when you set `COMPOSE_CONVERT_WINDOWS_PATHS` to `1` instead?

giggio:
@shin- Yes, `DOCKER_HOST` is pointing to a remote Linux server.
The error happens when `COMPOSE_CONVERT_WINDOWS_PATHS` is unset. When I set it to `COMPOSE_CONVERT_WINDOWS_PATHS=0` (in powershell `$env:COMPOSE_CONVERT_WINDOWS_PATHS=0`), then it works.
I hadn't tried before, but `COMPOSE_CONVERT_WINDOWS_PATHS=1` also works. Go figure...
[...]

PR #4294, opened 2017-01-04
Ensure falsy values in COMPOSE_CONVERT_WINDOWS_PATHS are properly recognized

shin-:
This was uncovered in #4240 but does not fix the issue - however, it should avoid some of the additional confusion around it.

### DIFF
--- a/compose/config/config.py
+++ b/compose/config/config.py
@@ -712,7 +712,7 @@ def finalize_service(service_config, service_names, version, environment):
     if 'volumes' in service_dict:
         service_dict['volumes'] = [
             VolumeSpec.parse(
-                v, environment.get('COMPOSE_CONVERT_WINDOWS_PATHS')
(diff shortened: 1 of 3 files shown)


cmosguy:
Hi @shin- and docker dev folks.  So is this wrapped up in the latest docker-compose release?  If I understand this fix correctly, Windows users are no longer required to set the COMPOSE_CONVERT_WINDOWS_PATHS variable?  Is this still a requirement?  Because this is limited in it's documentation and error outputs..

shin-:
Newer versions of Docker Machine / Toolbox will set the value for you if it detects it's necessary. Other setups ("Docker For Windows" included) don't need to set this variable. If you're still experiencing problems with host binds on Windows, please open a new issue.

Dominic-Mathina:
I still have this issue even after setting the environment variables

Issue #4306, opened 2017-01-09
docker-compose run ignores --rm when run with -d


clintam:
When running `docker-compose run --rm -d service` the remove option is silently ignored. Rather, I expected to fail fast with a validation message similar to docker

```
$ docker run --rm -d busybox
Conflicting options: --rm and -d
```

## Steps to reproduce

Create a simple docker compose:

```
service:
  image: busybox:latest
  command: echo simple
```

Run `docker-compose run --rm -d service`

Observe that the container is not removed:
```
$ docker ps -a
CONTAINER ID        IMAGE               COMMAND                  CREATED             STATUS                     PORTS               NAMES
ceb2bdadd3f1        busybox:latest      "echo simple"            6 seconds ago       Exited (0) 5 seconds ago                       test_service_run_1
```

ijc:
Closing along with https://github.com/docker/compose/pull/4307#issuecomment-494376415.

PR #4307, opened 2017-01-09
Disallow running with -d and —rm

clintam:
Currently, only -d is honoured. Failing fast on this
option is consistent with docker run —d —-rm <image>

Fixes: https://github.com/docker/compose/issues/4306
Signed-off-by: Clint Morgan <clint.a.m@gmail.com>

### DIFF
--- a/tests/unit/cli_test.py
+++ b/tests/unit/cli_test.py
@@ -127,7 +127,10 @@ class CLITestCase(unittest.TestCase):
         _, _, call_kwargs = mock_run_operation.mock_calls[0]
         assert call_kwargs['logs'] is False
 
-    def test_run_service_with_restart_always(self):
+    @pytest.mark.xfail(IS_WINDOWS_PLATFORM, reason="requires dockerpty")
+    @mock.patch('compose.cli.main.RunOperation', autospec=True)
+    @mock.patch('compose.cli.main.PseudoTerminal', autospec=True)
+    def test_run_service_with_restart_always(self, mock_pseudo_terminal, mock_run_operation):
         mock_client = mock.create_autospec(docker.APIClient)
 
         project = Project.from_config(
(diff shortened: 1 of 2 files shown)


tests/unit/cli_test.py
```diff
@@ -127,7 +127,10 @@ def test_run_interactive_passes_logs_false(self, mock_pseudo_terminal, mock_run_
         _, _, call_kwargs = mock_run_operation.mock_calls[0]
         assert call_kwargs['logs'] is False
 
-    def test_run_service_with_restart_always(self):
+    @pytest.mark.xfail(IS_WINDOWS_PLATFORM, reason="requires dockerpty")
+    @mock.patch('compose.cli.main.RunOperation', autospec=True)
+    @mock.patch('compose.cli.main.PseudoTerminal', autospec=True)
+    def test_run_service_with_restart_always(self, mock_pseudo_terminal, mock_run_operation):
```
shin-: Why is this being modified?

shin-:
Thank you for your contribution.

Note that the documentation currently specifies:
```
             --rm                  Remove container after run. Ignored in detached mode.
```

As a result, this would be a breaking change, and we have to weigh the impact of this on current use cases where people rely (maybe unknowingly) on this option being ignored.

tests/unit/cli_test.py
clintam: That existing test was calling with the now-invalid combo of -d and --rm. This path took advantage of the early exit from `run_one_off_container` and so did not need to worry about mocking the TTY behaviors.

This changes the test to :
* use -d: false (since the --rm is key to the behaviors under test)
* uses mocked versions of the Run/Ptty (as that path is now exercised)

clintam:
Thank you @shin- for the review. I overlooked that call out in the documentation; I agree that it should now be considered a breaking change.

One related consideration: --rm is *not* totally ignored when run with -d: It is used to compute the restart policy as None. (You can see this in the unit test I touched.)

I'm happy to defer to the maintainer's judgment here. Cheers,

ijc:
Sorry that this discussion seems to have petered out without closure one way or another.

I think after such a long time it's probably less disruptive to own the current behaviour rather than making a breaking change several years down the line. So I'll close this (and the linked issue I guess).

Thanks for your contribution.

Issue #1061, opened 2015-03-04
Add option to list non-running containers


hamiltont:
Can `docker-compose ps` have an option like `docker-compose ps -a` that shows containers which are non-running? Basically, I'd like  an easy method to see what's inside the `docker-compose.yml` file without having to cat the file and skim through it.

hamiltont:
Hm, so I just realized that `ps` lists non-running containers if they have exited. I suppose this should be a request to also list services with a "never started" status

aanand:
It might be good for `ps` to be more service-centric, yes.

draghuram:
#dibs

draghuram:
If "docker-compose ps -a" lists never started containers, wouldn't it be inconsistent with "docker ps -a" semantics?

technolo-g:
I would like to vote that we replicate the behavior of docker with 
`docker-compose ps` not showing exited or otherwise stopped containers and
`docker-compose ps -a` showing all containers running or not

nightscape:
:+1:

clintam:
I see what @draghuram means: `docker-compose ps` already lists 'stopped' containers, which is what `docker ps -a` provides. What we want here is to have a flag to also capture "never started" containers. There is no way to be "consistent" with `docker ps` here as docker would have no concept of containers that may be started in the future.

I can still see the value in such a mechanism as it will allow to check if any more work needs to be done to start all of the services in an environment. But it does feel a bit off right now in the difference between docker core.

Perhaps another approach would be to add a `--dry-run` flag to `docker-compose up` that would detail all of the stuff needed to bring an environment back up?

koooge:
It seems to be implemented by #6327

ijc:
Issue grooming: I agree with the previous comment. Closing.

PR #4311, opened 2017-01-09
Add option to list not-yet-started containers

clintam:
Add flag -a to the ps command.

Fixes: https://github.com/docker/compose/issues/1061

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -564,6 +564,7 @@ class TopLevelCommand(object):
 
         Options:
             -q    Only display IDs
+            -a   Show not started containers (default shows running and stopped)
         """
         containers = sorted(
             self.project.containers(service_names=options['SERVICE'], stopped=True) +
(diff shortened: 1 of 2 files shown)


clintam:
Example:
```
$ docker-compose ps

$ docker-compose ps -a
           Name               Command      State      Ports 
-----------------------------------------------------------
simplecomposefile_simple_1              Not started         
simplecomposefile_another_1             Not started         

$ docker-compose up -d
Creating simplecomposefile_simple_1
Creating simplecomposefile_another_1

$ docker-compose ps -a
           Name               Command   State   Ports 
-----------------------------------------------------
simplecomposefile_another_1   top       Up            
simplecomposefile_simple_1    top       Up            

```

The behavior  is different than `docker ps -a` where the 'a' means also show stopped containers.

Currently `docker-compose ps` already shows stopped containers. What is "missing" is the ability to see containers that have not yet been started. Doing this means that the container

ijc:
Hi @clintam, I'm sorry that nobody looked at this at the time.

I may be mistaken but it seems that `ps` now has a `-a` option since #6327 which I think covers the case you were addressing here too? (digging a bit further, it seems I managed to conclude the same when I closed #1061 a little while back, at least I'm consistent I suppose).

If I missing something and your case isn't handled (and you are still interested in fixing it, sorry again that it's taken so long to have someone look at this) then please do rebase and ping for reopening.

PR #4313, opened 2017-01-10
allow custom container names in zsh completion

paschdan:
when you define a container_name in your docker-compose.yml the completion won't work correctly.

example:

docker-compose.yml
```
version:"2"
services:
  my_service:
    image: busybox
    container_name: my_service
```

the current behavior is:
```
$ docker-compose run <TAB> ... my
$ docker-compose run my
```

expected behavior would be:

```
$ docker-compose run <TAB> ... my_service
$ docker-compose run my_service
```

this pull request tries to find out, if a service was automatically named by compose, or if a custom name was given by comparing if the last part of a container name is a number and the first part is the current directory.


shin-:
I don't believe this is correct - `docker-compose run` takes a service name as argument, not a container name, so the `container_name` parameter shouldn't have any influence on completion.

PR #4333, opened 2017-01-14
Add --resolve-image-digests option to docker-compose config command

### LINKED ISSUE #4332: Add option to resolve tags to hashes for config command
**Description**

Docker 1.13 adds the option to deploy stacks using Docker Compose version 3 files in addition to Distributed Application Bundle (DAB) files. When `docker-compose bundle` is used to produce a DAB, it fetches image digests and replaces image tags with digests in the compose project. This produces a DAB whose images are pinned to a specific digest, rather than a tag that can be changed to different digests over time.

`docker-compose config` can be used to “compile” or combine multiple compose files together, resolve `.env` variables, and handle all of Docker Compose's options to produce a single compose file suitable for use with `docker stack deploy --compose-file`. It would be useful to add an option to the `config` command that replaces image tags with digests—like the `bundle` command—to produce a compose file destined for stack deployments that acts like DAB files with respect to image pinning.

1. Given Docker Compose file(s) forming a project, call `docker-compose config`.

A combined/resolved Docker Compose configuration is outputted. Any image tags remain as tags.

An option on `docker-compose config` that causes Docker Compose to output image digests instead of tags.

The brief description of the `config` command says that is is to “Validate and view the compose file.” But, it is the only way to produce a Docker Compose file that uses features not supported by `docker stack deploy --compose-file`. I believe this feature request points to a broader need for `docker-compose config` to be treated more like `docker-compose bundle`, and be used for generating a single/compiled/resolved Compose file for a Docker Compose project. This includes adopting bundle's options (`--push-images` and `--output`).

kinghuang:
Add an option named `--digest` to the `docker-compose config` command. It pins images to specific image digests for compose files, just like the `docker-compose bundle` command for DAB files.

Resolves #4332.

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -263,43 +263,7 @@ class TopLevelCommand(object):
         if not output:
             output = "{}.dab".format(self.project.name)
 
-        with errors.handle_connection_errors(self.project.client):
-            try:
-                image_digests = get_image_digests(
-                    self.project,
-                    allow_push=options['--push-images'],
-                )
-            except MissingDigests as e:
-                def list_images(images):
-                    return "\n".join("    {}".format(name) for name in sorted(images))
-
-                paras = ["Some images are missing digests."]
-
-                if e.needs_push:
-                    command_hint = (
-                        "Use `docker-compose push {}` to push them. "
-                        "You can do this automatically with `docker-compose bundle --push-images`."
-                        .format(" ".join(sorted(e.needs_push)))
-                    )
-                    paras += [
-                        "The following images can be pushed:",
-                        list_images(e.needs_push),
-                        command_hint,
-                    ]
-
-                if e.needs_pull:
-                    command_hint = (
-                        "Use `docker-compose pull {}` to pull them. "
-                        .format(" ".join(sorted(e.needs_pull)))
-                    )
-
-                    paras += [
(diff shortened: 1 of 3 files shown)


shin-:
cc @dnephin for your thoughts - not sure if this was disccused before.

dnephin:
I don't know of any previous discussion about this, but it seems like a good idea to me. It's effectively a "pin the image" feature, which I think is good.

On the implementation, I think there is a lot of duplicated code. Can you have it share the code with `bundle` ?

kinghuang:
Sure, I can work on refactoring the digests code between `bundle` and `config`.

Does `--digest` sound ok for the option name? Or, would something like `--pin-tags` be clearer?

dnephin:
In the context of `docker-compose config` I think it could be something like `--pin-images-to-digest`, or `--resolve-image-digests`

kinghuang:
The option has been renamed to `--resolve-image-digests`, and the common code in `bundle` and `config` has been moved to a function named `image_digests_for_project`.

kinghuang:
@shin- What's the process for getting this into a release?

shin-:
I still need to take a closer look at it, but tentatively adding to the 1.12 milestone

Thanks for reminding me :+1:

PR #4351, opened 2017-01-19
Note explicit support for Python 3.5 and 3.6

felixonmars:


shin-:
Thanks!

We're not currently testing for those versions (we should!), which is why I would hold off on making anything official for the time being.

felixonmars:
Is it enough to change the tox environment? Or maybe some other infrastructures?

shin-:
3.6 support addressed in #5588

PR #4397, opened 2017-01-27
Add support for returning the exit value of a specific container

memory:
Current best practice for using docker-compose as a tool for continuous
integration requires fragile shell pipelines to query the exit status
of composed containers, e.g.:

http://stackoverflow.com/questions/29568352/using-docker-compose-with-ci-how-to-deal-with-exit-codes-and-daemonized-linked
http://blog.ministryofprogramming.com/docker-compose-and-exit-codes/

This PR adds a `--forward-exitval <container>` flag that allows
`docker-compose up` to return the exit value of a specified container.
The container may optionally have a number specified (foo_2) otherwise
the first is defaulted to.

Signed-off-by: Nathan J. Mehl <n@climate.com>


memory:
...ping?

compose/cli/main.py
```diff
@@ -863,6 +871,22 @@ def up(self, options):
             if cascade_stop:
                 print("Aborting on container exit...")
                 self.project.stop(service_names=service_names, timeout=timeout)
+                if forward_exitval:
+                    def is_us(container):
+                        return container.name_without_project == forward_exitval
+                    candidates = filter(is_us, all_containers)
+                    if not candidates:
+                        log.error('No containers matching the spec "%s" were run.',
+                                  forward_exitval)
```
shin-: Ideally we want to detect this before running the containers. Unfortunately we don't know how many instances of a certain service will be run, but we should at least check for a service name match before.

Additionally, I think the feature would be improved if the option just took a service name as argument. If the service isn't scaled, the behavior would be the same. If it is, return the first exit code in the list that isn't 0 (or 0 if no container exited abnormally)

compose/cli/main.py
```diff
@@ -826,17 +826,23 @@ def up(self, options):
                                        running. (default: 10)
             --remove-orphans           Remove containers for services not
                                        defined in the Compose file
+            --forward-exitval SERVICE  Return the exit value of the selected service container.
```
shin-: `--exit-code-from` would be a better name.

shin-: `/s/exit value/exit code/` in the docs part as well.

memory: Reasonable!

memory: Willdo.

compose/cli/main.py
```diff
@@ -826,17 +826,23 @@ def up(self, options):
...
         timeout = timeout_from_opts(options)
         remove_orphans = options['--remove-orphans']
         detached = options.get('-d')
+        forward_exitval = container_exitval_from_opts(options)
 
         if detached and cascade_stop:
             raise UserError("--abort-on-container-exit and -d cannot be combined.")
 
+        if forward_exitval and not cascade_stop:
+            raise UserError("--forward-exitval requires --abort-on-container-exit.")
```
memory: Legit.

compose/cli/main.py
```diff
@@ -850,9 +856,11 @@ def up(self, options):
             if detached:
                 return
 
+            all_containers = filter_containers_to_service_names(to_attach, service_names)
```
memory: 👍

compose/cli/main.py
```diff
@@ -863,6 +871,22 @@ def up(self, options):
             if cascade_stop:
                 print("Aborting on container exit...")
                 self.project.stop(service_names=service_names, timeout=timeout)
+                if forward_exitval:
+                    def is_us(container):
+                        return container.name_without_project == forward_exitval
+                    candidates = filter(is_us, all_containers)
```
memory: I actually tried that and ran into a scoping issue, but that was probably my own 2am incompetence -- I'll go commune with the documentation and try again.

compose/cli/main.py
memory: 👍

memory:
Sorry for the slow progress here; I've been travelling on business for the past few weeks.  Will try to have an update to the PR ready today or tomorrow.

memory:
@shin- okay, I think we're good to go here.  Have a look?

shin-:
Looks good, thanks!

I'm holding off on merging because I expect there will be conflicts with #4466 - so I'll wait until both are ready and merge manually then.

Issue #1652, opened 2015-07-06
Epic: Improve performance by running operations in parallel


bfirsh:
Many of Compose's actions can be performed in parallel rather than in sequence. For larger apps, this would result in enormously reduced startup/redeploy times.

This is a meta-issue for keeping track of all of the stuff happening around that, in rough priority order.
- [x] `docker-compose (stop|kill|remove)`: https://github.com/docker/compose/issues/1651
- [x] `docker-compose scale`: https://github.com/docker/compose/issues/1653
- [x] `docker-compose up`: Creating, pulling, building and starting in `up` can be done in parallel: #1984
- [x] `docker-compose pull`: Multiple services can be pulled in parallel.
- [x] `docker-compose pull --parallel` is the default (deprecate flag).

(Replaces #1395)

dnephin:
#931 is also related, doing pull and build in parallel (although I don't think build will benefit at all).

gkze:
Bump -- this would be amazing since we are having to deal with more and more services in containers that don't depend on each other, so this would help speed up the process immensely

aanand:
@gkze `stop`, `kill`, `rm` and `scale` now work in parallel - try it out in [Compose 1.4.0 RC2](https://github.com/docker/compose/releases/tag/1.4.0rc2).

k2xl:
Cool. What about dc build?

dnephin:
@k2xl because of this https://github.com/docker/docker/issues/9656 bug/limitation in docker doing build in parallel in compose would actually make things slower. Once that's fixed, it should be pretty easy to add.

IlyaSkriblovsky:
Hi! What you would suggest if my app needs to write something to DB when stopping gracefully? Is it possible to acheive ordered stopping with Compose?

@aanand said in #1395:
> 3. Container stopping/killing should be completely order-independent - if it's all going down, it doesn't matter what order we do it in.
While this is true for almost any regular web app, but I can't agree with that for something more stateful. For example, when my server receives stop signal, it gracefully closes all active connections and saves log records on each disconnection to DB. I'm trying to dockerize my app with Compose and this seems to be an issue :(  Thanks for advice.

willejs:
+1 looking forward to this feature.

khalilj:
+1

IlyaSkriblovsky:
Starting phase of `up` made parallel in #2679 (and also `start` and `stop`)

morlajb:
+1

[...]

PR #4428, opened 2017-02-04
Pull services in parallel

edsrzf:
Updates #1652

This PR adds a new `--parallel` flag on `docker-compose pull` that pulls images in parallel, which should hopefully speed up pulls for projects with lots of services. By default, with no flag, you get the old, non-parallel behavior.

`parallel_execute` now accepts an optional `limit` parameter that specifies the number of threads that actually run at one time. There are arguably more elegant approaches than mine, but I opted to keep the diff relatively small. I'm happy to get some feedback about this. I've somewhat arbitrarily chosen 5 as the limit for the number of parallel pulls. I'm also happy to break this part out into a separate PR since it's mostly independent from the rest of this PR and could be useful in other contexts.

The normal `docker-compose pull` output is silenced when pulling in parallel. This means we no longer get progress bars with `--parallel`, but the output is easier to understand.

I've tested this manually and it seems to work. I'm not sure how I'd go about
writing an automated test for it, but open to suggestions.

### DIFF
--- a/compose/project.py
+++ b/compose/project.py
@@ -456,7 +456,7 @@ class Project(object):
 
     def pull(self, service_names=None, ignore_pull_failures=False):
         def pull_service(service):
-            service.pull(ignore_pull_failures)
+            service.pull(ignore_pull_failures, True)
 
         services = self.get_services(service_names, include_deps=False)
         parallel.parallel_execute(
--- a/compose/project.py
+++ b/compose/project.py
@@ -463,7 +463,8 @@ class Project(object):
             services,
             pull_service,
             operator.attrgetter('name'),
-            'Pulling')
+            'Pulling',
+            limit=5)
 
     def push(self, service_names=None, ignore_push_failures=False):
         for service in self.get_services(service_names, include_deps=False):
--- a/compose/project.py
+++ b/compose/project.py
@@ -454,17 +454,22 @@ class Project(object):
 
         return plans
 
-    def pull(self, service_names=None, ignore_pull_failures=False):
-        def pull_service(service):
-            service.pull(ignore_pull_failures, True)
-
+    def pull(self, service_names=None, ignore_pull_failures=False, in_parallel=False):
         services = self.get_services(service_names, include_deps=False)
-        parallel.parallel_execute(
-            services,
-            pull_service,
-            operator.attrgetter('name'),
-            'Pulling',
-            limit=5)
+
+        if in_parallel:
+            def pull_service(service):
+                service.pull(ignore_pull_failures, True)
+
+            parallel.parallel_execute(
+                services,
+                pull_service,
+                operator.attrgetter('name'),
+                'Pulling',
+                limit=5)
+        else:
+            for service in services:
+                service.pull(ignore_pull_failures)
 
     def push(self, service_names=None, ignore_push_failures=False):
         for service in self.get_services(service_names, include_deps=False):
--- a/compose/project.py
+++ b/compose/project.py
@@ -454,10 +454,10 @@ class Project(object):
 
         return plans
 
-    def pull(self, service_names=None, ignore_pull_failures=False, in_parallel=False):
+    def pull(self, service_names=None, ignore_pull_failures=False, parallel_pull=False):
         services = self.get_services(service_names, include_deps=False)
 
-        if in_parallel:
+        if parallel_pull:
             def pull_service(service):
                 service.pull(ignore_pull_failures, True)
 
--- a/compose/project.py
+++ b/compose/project.py
@@ -455,8 +455,15 @@ class Project(object):
         return plans
 
     def pull(self, service_names=None, ignore_pull_failures=False):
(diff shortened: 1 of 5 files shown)


shin-:
Thanks! 

I'm honestly not sure how well this would scale, how the engine handles 10, maybe 20 simultaneous pulls - would this potentially cause timeouts or significant slowdowns on certain networks - how well does our client handle maintaining several connections open at the same time - not to mention it probably messes up the output in Compose.

In order for me to accept this, it would at least need to be an option behind a flag, e.g `docker-compose pull --parallel`, and disabled by default.

edsrzf:
Thanks for the feedback! I want to do this right. I'm fine with keeping it behind a `--parallel` flag to start with, but would also like to try and fix the problems so that this could, perhaps someday, be the default.

Admittedly, the output is a bit weird. If we could limit each pull to a single status line and maybe get rid of the progress bars, or at least confine them to one line, that might be more manageable.

I've been running with this on a project that has 27 different services and been having no real problems pulling all images in parallel, but I can imagine that it could cause problems with some setups. Maybe we could limit the maximum number of pulls running at a time?

I'll look at `parallel.py` and see what I can do for both of these things.

edsrzf:
I've updated. Here's what's new:

There's a `--parallel` flag on `docker-compose pull`. By default, you get the old, non-parallel behavior.

`parallel_execute` now accepts an optional `limit` parameter that specifies the number of threads that actually run at one time. There are arguably more elegant approaches than mine, but I opted to keep the diff relatively small. I'm happy to get some feedback about this. I've somewhat arbitrarily chosen 5 as the limit for the number of parallel pulls. I'm also happy to break this part out into a separate PR since it's mostly independent from the rest of this PR and could be useful in other contexts.

The normal `docker-compose pull` output is silenced when pulling in parallel. This means we no longer get progress bars with `--parallel`, but the output is easier to understand.

compose/project.py
```diff
@@ -454,9 +454,22 @@ def _get_convergence_plans(self, services, strategy):
 
         return plans
 
-    def pull(self, service_names=None, ignore_pull_failures=False):
-        for service in self.get_services(service_names, include_deps=False):
-            service.pull(ignore_pull_failures)
+    def pull(self, service_names=None, ignore_pull_failures=False, in_parallel=False):
```
shin-: Nit: can we rename `in_parallel` to `parallel_pull` instead?

compose/service.py
```diff
@@ -885,17 +885,19 @@ def has_host_port(binding):
 
         return any(has_host_port(binding) for binding in self.options.get('ports', []))
 
-    def pull(self, ignore_pull_failures=False):
+    def pull(self, ignore_pull_failures=False, silent=False):
```
shin-: Maybe instead of `silent`, have a `no_progress` argument that still logs the `Pulling ...` lines, but omits the progress_stream part.

compose/service.py
```diff
@@ -885,17 +885,19 @@ def has_host_port(binding):
...
         tag = tag or 'latest'
-        log.info('Pulling %s (%s%s%s)...' % (self.name, repo, separator, tag))
+        if not silent:
+            log.info('Pulling %s (%s%s%s)...' % (self.name, repo, separator, tag))
         try:
             output = self.client.pull(repo, tag=tag, stream=True)
-            return progress_stream.get_digest_from_pull(
-                stream_output(output, sys.stdout))
+            if not silent:
+                return progress_stream.get_digest_from_pull(
```
shin-: I think we still need the result of `get_digest_from_pull` here, but we need to modify the  `stream_output` call to not write to `stdout`. Maybe something like
```python
with open(os.devnull, 'w') as devnull:
  return progress_stream.get_digest_from_pull(
    stream_output(output, devnull)
  )
```

compose/service.py
edsrzf: There are a couple problems with doing this:

First, `parallel_execute` outputs lines very similar to `Service.pull`. It's

`Pulling service (image:tag) ...`

vs.

`Pulling service ...`

So we get the same thing output twice for each service, basically.

Second, `parallel_execute` outputs some terminal commands to go back and update older lines, and if anything else outputs something it messes up the terminal commands and the wrong lines are updated.

compose/service.py
edsrzf: I've got no problem doing this, but as far as I can tell no one uses the value returned from this method.

compose/service.py
shin-: Nevermind then - as long as we have some level of feedback for the user during the operation, that should be sufficient.

compose/service.py
shin-: The cost of doing this seems small enough - it might be used by people using Compose as a library, or we might want to use it in the future - so I'd say it's worth it.

edsrzf:
Updated! Issues from code review should be addressed. I also changed from using a `Queue` to a `Semaphore` to limit parallelism since it just made more sense. And finally, I added a test for `parallel_execute` when there's a limit.

Please take another look.

PR #4443, opened 2017-02-08
Referencing the right segment of code in compose/bundle.py

xulike666:
Signed-off-by: Aaron.L.Xu <likexu@harmonycloud.cn>


shin-:
Thank you!

PR #4444, opened 2017-02-08
Provide support for ARM 32bit

### LINKED ISSUE #3795: Add support for building Docker-Compose on ARM
Currently it is not possible to build Docker-Compose for ARM (e.g. Raspberry Pi) with the standard Dockerfile and scripts in scripts/build/linux and scripts/build/linux-entrypoint.

This is mainly due to the fact that pyinstaller does not come with a bootloader that supports ARM out of the box.  
### How to add ARM support

This [repository and PR](https://github.com/hypriot/arm-compose/pull/9) demonstrates what needs to be done to build Docker-Compose for ARM.

I would like to help and prepare a PR that adds support for building Compose for ARM.
Before I start I would like to get some feedback and some direction on how to go about this.

One way to implement this would be to [add some detection for the OS and the architecture](https://github.com/docker/docker/blob/master/hack/make/.detect-daemon-osarch) similar to how it is done in the docker/docker repo.

Based on the result one could either install standard pyinstaller or the version patched for ARM support. This switch could be implemented in scripts/build/linux-entrypoint.

What do you think?

jcberthon:
I wanted to use `docker-compose` on my Raspberry Pi 2 (ARM 32bit). There is no built version for it, so I attempted to build one myself. The provided commit implement the necessary change to support this architecture. Fixes #3795 

Based on the `master` branch and due to recent commits (which upgraded pip and a few other dependencies), the change is now very simple. Basically it revolves around changing the base image (`debian` is x86_64 only, so I've been using `armhf/debian` instead) and downloading the docker binary for armel architecture.

I've tried to adapt the script/test/default in order to validate the generated docker-compose binary, but it is an unfinished job as I'm stuck with the fact that there are no images available of `dind` for ARM (`dockerswarm/dind` only provide x86_64 images).

Anyway, I have 2 small projects which uses docker-compose.yml and both works. But I'm only using a subset of all of docker-compose features, and I do not have a Swarm configured to test those aspects as well. So my commit is only partially validated. But at least it works perfectly for what I'm using of it.

_Note: Try to use the `bump-1.10.0` branch (in order to get a "stable" version of docker-compose) is a bit more tricky and requires several additional changes w.r.t. this PR. Those are visible on [this branch of my fork](https://github.com/jcberthon/compose/tree/compose-armv7), this is just for your information._

### DIFF
--- a/Dockerfile.armhf
+++ b/Dockerfile.armhf
@@ -1,0 +1,71 @@
+FROM armhf/debian:wheezy
+
+RUN set -ex; \
+    apt-get update -qq; \
+    apt-get install -y \
+        locales \
+        gcc \
+        make \
+        zlib1g \
+        zlib1g-dev \
+        libssl-dev \
+        git \
+        ca-certificates \
+        curl \
+        libsqlite3-dev \
+        libbz2-dev \
+    ; \
+    rm -rf /var/lib/apt/lists/*
+
+RUN curl https://get.docker.com/builds/Linux/armel/docker-1.8.3 \
+        -o /usr/local/bin/docker && \
+    chmod +x /usr/local/bin/docker
+
+# Build Python 2.7.13 from source
+RUN set -ex; \
+    curl -L https://www.python.org/ftp/python/2.7.13/Python-2.7.13.tgz | tar -xz; \
+    cd Python-2.7.13; \
+    ./configure --enable-shared; \
+    make; \
+    make install; \
+    cd ..; \
+    rm -rf /Python-2.7.13
+
+# Build python 3.4 from source
+RUN set -ex; \
+    curl -L https://www.python.org/ftp/python/3.4.6/Python-3.4.6.tgz | tar -xz; \
+    cd Python-3.4.6; \
+    ./configure --enable-shared; \
+    make; \
+    make install; \
+    cd ..; \
+    rm -rf /Python-3.4.6
+
+# Make libpython findable
+ENV LD_LIBRARY_PATH /usr/local/lib
+
+# Install pip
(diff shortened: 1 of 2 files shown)


jcberthon:
I have found out that I would need to update the script/release/build-binaries in order to build for the right target platform. Actually, this script does not need to be changed, but the scripts it calls should be adapted.

I can try to work on that, but I'd like some feedback/guidance on how to do that so it fits best with the project "culture".

My idea would be to add an architecture detection function which could be called by scripts in order to pick the right Dockerfile. So basically, the script call the detection functions which returns "armhf" (for instance) and then check that a Dockerfile.armhf exists if not fails. For amd64 (the "default" architecture), we would search for Dockerfile simply (without an filename extension).

This would require of course that all Dockerfile have an equivalent for each CPU architecture.

What do you think?

shin-:
Hey, thanks for taking the time to look into this, and sorry for the delayed response!

This PR as it is looks good to me. As far as `./script/release/build-binaries` is concerned, I'm not sure it's necessary to update it since it's mostly an internal tool for me and other contributors to produce a new release, and I don't think it would be wise for us to make an official ARM release at that time.
Rather, it would be good to have a new `arm` script in `./script/build` for people to build their own binary from source if they are on an architecture that supports it. Let me know what you think!

jcberthon:
Hi @shin-

Great! That is fine with me. I can come up with such script. Do you prefer that I make a new PR then?

Given my limited free time, it might take a week or two before I find the time to write it.

shin-:
Yes, a separate PR is fine whenever you have time to do so. Thank you!

PR #4455, opened 2017-02-09
Fix version 3.1

dnephin:


compose/config/config.py
```diff
@@ -186,11 +186,6 @@ def version(self):
         if version == '3':
             version = V3_0
 
-        if version not in (V2_0, V2_1, V3_0):
```
ebriney: I'm surprised that there is no parsing for `3.1`version here
```
        if version == '3.1':
            version = V3_1
```
no?

compose/config/config.py
```diff
@@ -186,11 +186,6 @@ def version(self):
         if version == '3':
             version = V3_0
```
ebriney: I'm surprised that there is no parsing for 3.1version here

        if version == '3.1':
            version = V3_1
no?

dnephin: No. These are aliases for converting "3" -> "3.0". There is no alias for 3.1

ebriney: Oh yes, thanks.

Issue #2578, opened 2015-12-26
Proposal: ignored objects in compose file


GeorgeTaveras1231:
Recently, I've been writing compose files where many of the services had small similarities. The way in which I attempted to deal with this was by using `yaml`'s inheritance. The reason why I wanted to use `yaml`'s inheritance feature  was because using `docker-compose`'s `extends` feature felt like an overkill. It also did not help that in order to `extend` a service, the base service had to have a valid service configuration, otherwise, `docker-compose` would complain. Also, if I define random shared objects at the top level, when I run `docker-compose up -d` it will attempt to create docker containers based on those objects.

My proposal is to support a top level namespace that can be used to define any `yaml` object you want. This will provide a namespace to define shared properties and leverage `yaml`'s inheritance.

Here is an example, (leveraging the proposed format in #2478):

``` yaml
ignore:
  has_special_env: &has_special_env
     VAR_A: 1
     VAR_B: 2

services:
  a:
    environment:
      <<: *has_special_env
      CUSTOM_VAR: 3
  b:
    environment:
      <<: *has_special_env
```

It would be nice if this was included in #2504

dnephin:
We've been hesitant to add any special sections for the purpose of yaml anchors. You can always create an anchor in the first section it's used and refer to it again later.

Also overlaps a bit with #1655

GeorgeTaveras1231:
@dnephin You can do that in the scenario where the first section is a exact subset of the other. There may be scenarios where that is not the case. For example:

```yaml
services:
  a:
    environment:
      VAR_A: 1
      VAR_B: 2
      CUSTOM_VAR: 3
  b:
    environment:
       VAR_A: 1
      VAR_B: 2
      CUSTOM_VAR: 4
```

sblackstone:
I had a good use case for this just today, (as see in the dupe ticket #3947)

I have several services running delayed_job and they need to share most of the app's config, but not all - e.g. it shouldn't be trying to publish ports.

```
version: '2'

default: &default
  image: my_app
  build: .    
  volumes:
    - .:/app
  environment:      
    RAILS_ENV: development
    DATABASE_URL: mysql2://dba:abcd1234@10.0.0.1/my_database

services:      
  app:
    <<: *default
    ports:
      - "3000:3000"
    command: /app/docker-up.sh

  delayed_job_default:
    <<: *default
    command: bin/delayed_job --queue default run
```

dnephin:
Closing as duplicate of #2942

PR #4461, opened 2017-02-10
Allow an 'extra' top-level element in config files.

tbeadle:
This allows writers of docker-compose.yml files to factor out repeated
bits of configuration to be under an 'extra' top-level key, allowing the
use of yml anchors.

Signed-off-by: Tommy Beadle <tbeadle@gmail.com>

### DIFF
--- a/compose/config/config_schema_v2.0.json
+++ b/compose/config/config_schema_v2.0.json
@@ -38,6 +38,12 @@
         }
       },
       "additionalProperties": false
+    },
+
+    "extra": {
+      "id": "#/properties/extra",
+      "type": "object",
+      "additionalProperties": true
     }
   },
 
--- a/compose/config/config_schema_v2.1.json
+++ b/compose/config/config_schema_v2.1.json
@@ -38,6 +38,12 @@
         }
       },
       "additionalProperties": false
+    },
+
+    "extra": {
+      "id": "#/properties/extra",
+      "type": "object",
+      "additionalProperties": true
     }
   },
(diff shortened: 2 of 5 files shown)


shin-:
First of all, thank you for contributing to the project!

This was proposed some time ago in #2578 - now we have an actual implementation. This looks simple enough and non-intrusive. It would probably require an update in docker/docker as well.

You'll need to remove the change from the 3.0 format as it is now frozen.

cc @dnephin I know you were hesitant in the past about such a feature, do you have any new thoughts on this?

tbeadle:
@shin I figured it could be included in all of the formats 2.0 and greater because it's backwards compatible.  What would be required to be changed in docker/docker?  I didn't think any of the changes here would impact that at all.

dnephin:
I'm fine with supporting "extra" sections, but I think something like this is more common:

```json
    "patternProperties": {
        "x-.+": {}
    },
```

`docker/docker` now reads Compose files for `stack deploy`, so the change needs to be supported by both.

JanNash:
Since version 3 is frozen, I needed a way to share common configuration, searched around, found this and saw that there's seemingly not been any activity here since a few months, I hacked [a small workaround](https://github.com/JanNash/docker-noop), which I think I could share here (I'm not sure if this is the right place but feel free to tell me, where else I can share this info :))

I'm not going to elaborate on it here, since there's a readme in the repo.
Have a nice one, everyone ☮️

PR #4466, opened 2017-02-10

kbh2o:

### DIFF
--- a/compose/cli/log_printer.py
+++ b/compose/cli/log_printer.py
@@ -90,8 +90,9 @@ class LogPrinter(object):
             if self.cascade_stop:
                 matching_container = [cont.name for cont in self.containers if cont.name == line]
                 if line in matching_container:
-                    self.cascade_starter = line
-                    break
+                    # Returning the name of the container that started the
+                    # the cascade_stop so we can return the correct exit code
+                    return line
 
             if not line:
                 if not thread_map:
@@ -233,10 +234,7 @@ def consume_queue(queue, cascade_stop):
         if item.exc:
             raise item.exc
 
-        if item.is_stop:
-            if cascade_stop:
-                yield item.item
-            else:
-                continue
+        if item.is_stop and not cascade_stop:
+            continue
 
         yield item.item
--- a/compose/cli/log_printer.py
+++ b/compose/cli/log_printer.py
@@ -87,6 +87,12 @@ class LogPrinter(object):
         for line in consume_queue(queue, self.cascade_stop):
             remove_stopped_threads(thread_map)
 
+            if self.cascade_stop:
+                matching_container = [cont.name for cont in self.containers if cont.name == line]
+                if line in matching_container:
+                    self.cascade_starter = line
+                    break
+
             if not line:
                 if not thread_map:
                     # There are no running containers left to tail, so exit
@@ -132,8 +138,8 @@ class QueueItem(namedtuple('_QueueItem', 'item is_stop exc')):
         return cls(None, None, exc)
 
     @classmethod
-    def stop(cls):
-        return cls(None, True, None)
+    def stop(cls, item=None):
+        return cls(item, True, None)
 
 
 def tail_container_logs(container, presenter, queue, log_args):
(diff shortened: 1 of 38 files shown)


kbh2o:
bump

shin-:
Thanks!

Looking at Aanand's comment on the original PR (https://github.com/docker/compose/pull/3725#issuecomment-233958678), I think there's one important piece that's missing, with the code as it is currently there is no guarantee that the container that caused the cascade stop is the one we're looking at for the potential exit code change. 

To accomplish that, changes in the `log_printer` might be needed to identify the container that caused the interruption and retrieve that specific code.

kbh2o:
Thanks for the feedback!  I'll look into the `log_printer` and push up changes soon.

kbh2o:
I pushed up some changes that work the suggested way.  Before I modify/write additional tests, I would like to know if this method makes sense.

compose/cli/log_printer.py
```diff
@@ -87,6 +87,12 @@ def run(self):
         for line in consume_queue(queue, self.cascade_stop):
             remove_stopped_threads(thread_map)
 
+            if self.cascade_stop:
+                matching_container = [cont.name for cont in self.containers if cont.name == line]
+                if line in matching_container:
+                    self.cascade_starter = line
```
shin-: maybe `return line` here instead

compose/cli/main.py
```diff
@@ -914,7 +914,13 @@ def up(self, options):
```
shin-: `cascade_starter = log_printer.run()` (can be `None`)

compose/cli/log_printer.py
shin-: And add a short comment to explain why we're doing that.

compose/cli/log_printer.py
```diff
@@ -230,7 +235,7 @@ def consume_queue(queue, cascade_stop):
 
         if item.is_stop:
             if cascade_stop:
-                raise StopIteration
+                yield item.item
```
shin-: we already `yield item.item` unconditionally below, so it seems like this whole block can be changed to
```
if item.is_stop and not cascade_stop:
  continue
```

kbh2o:
Thanks for the feedback.  I pushed the changes up.  I have tried to fix signing the "merge from master" commit, but I am not having any luck.

shin-:
Yeah, ideally we don't want merge commits in PRs, they should be rebased against master instead. Not a huge deal. :)

tests/fixtures/abort-on-container-exit-0/digest.yml
```diff
@@ -0,0 +1,6 @@
+simple:
```
shin-: File is unused

tests/fixtures/abort-on-container-exit-0/ignore-pull-failures.yml
```diff
@@ -0,0 +1,6 @@
+simple:
```
shin-: File is unused

tests/fixtures/abort-on-container-exit-1/digest.yml
```diff
@@ -0,0 +1,6 @@
+simple:
```
shin-: File is unused

tests/fixtures/abort-on-container-exit-1/ignore-pull-failures.yml
```diff
@@ -0,0 +1,6 @@
+simple:
```
shin-: File is unused

compose/cli/main.py
```diff
@@ -886,11 +886,17 @@ def up(self, options):
...
-            log_printer.run()
+            cascade_starter = log_printer.run()
 
             if cascade_stop:
                 print("Aborting on container exit...")
+                exit_code = 0
+                for e in self.project.containers(service_names=options['SERVICE'], stopped=True):
+                    if (not e.is_running and cascade_starter == e.name):
+                        if not e.exit_code == 0:
+                            exit_code = e.exit_code
```
shin-: I think we can add a `break` here (inside the if block).

compose/cli/main.py
```diff
@@ -886,11 +886,18 @@ def up(self, options):
...
+                        break
```
shin-: Should indent to be inside the `if not e.exit_code == 0` block

kbh2o: oops. I'll update it.

kbh2o:
not sure what I did... but that doesn't look right... (-.-)

shin-:
Oops! Maybe create a new branch from master `git checkout -b new_branch_name master` and cherry-pick your commits into it. Make sure you've pulled the latest master.

kbh2o:
Created a new PR #4555 I'll close this one out.

Issue #4309, opened 2017-01-09
Add --cache-from


thomaschaaf:
With https://github.com/docker/docker-py/pull/1383 docker-py added cache-from to the sdk. We could integrate it into docker compose making https://github.com/docker/docker/pull/26839 available.

PR #4310, opened 2017-01-09
Implement cache from

thomaschaaf:
Fixes: https://github.com/docker/compose/issues/4309
Depends on: https://github.com/docker/docker-py/pull/1383

Signed-off-by: Thomas Schaaf <schaaf@komola.de>

PR #4514, opened 2017-02-20
Add cache_from as an available command

jyapayne:
This fixes the issues with PR #4310.

Fixes: #4309

Signed-off-by: Joey Payne <jyapayne@gmail.com>

### DIFF
--- a/compose/config/config_schema_v3.1.json
+++ b/compose/config/config_schema_v3.1.json
@@ -71,7 +71,8 @@
               "properties": {
                 "context": {"type": "string"},
                 "dockerfile": {"type": "string"},
-                "args": {"$ref": "#/definitions/list_or_dict"}
+                "args": {"$ref": "#/definitions/list_or_dict"},
+                "cache_from": {"type": "#/definitions/list_of_strings"}
               },
               "additionalProperties": false
             }
--- a/compose/service.py
+++ b/compose/service.py
@@ -802,6 +802,7 @@ class Service(object):
             nocache=no_cache,
             dockerfile=build_opts.get('dockerfile', None),
             buildargs=build_opts.get('args', None),
+            cache_from=build_opts.get('cache_from', None),
         )
 
         try:
--- a/tests/integration/service_test.py
+++ b/tests/integration/service_test.py
@@ -32,6 +32,7 @@ from compose.service import NetworkMode
 from compose.service import Service
 from tests.integration.testcases import v2_1_only
 from tests.integration.testcases import v2_only
+from tests.integration.testcases import v3_only
 
 
 def create_and_start_container(service, **override_options):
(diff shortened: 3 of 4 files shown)


jyapayne:
@shin- My jenkins build failed because the docker version in the Dockerfile is too low. In the Dockerfile, your repository lists

```dockerfile
RUN curl https://get.docker.com/builds/Linux/x86_64/docker-1.8.3 \
        -o /usr/local/bin/docker && \
    chmod +x /usr/local/bin/docker
```

as the method of installing docker. I realize that this is to keep the image lighter. Is there a way to update this docker version to the latest (1.13.1)?

shin-:
The Dockerfile does not need to be changed - the CLI version is irrelevant. You do need to use API version 1.25, which means the `cache_from` option can only be added to format 3.0+

jyapayne:
@shin- Ah, okay. I've updated the test and then schema to reflect the information you've provided. I'll close those other issues.

shin-:
@jyapayne Do you mind rebasing this against the newest `master`? That should resolve the Jenkins issue hopefully.

compose/config/config_schema_v3.0.json
```diff
@@ -60,7 +60,8 @@
               "properties": {
                 "context": {"type": "string"},
                 "dockerfile": {"type": "string"},
-                "args": {"$ref": "#/definitions/list_or_dict"}
+                "args": {"$ref": "#/definitions/list_or_dict"},
+                "cache_from": {"$ref": "#/definitions/list_of_strings"}
```
dnephin: Config v3.0 is frozen, so it can't be added here.

compose/config/config.py
```diff
@@ -101,6 +101,7 @@
 
 ALLOWED_KEYS = DOCKER_CONFIG_KEYS + [
     'build',
+    'cache_from',
```
dnephin: `cache_from` is under `build`, so it shouldn't be in this list. This list is only for the top level service keys

compose/config/config_schema_v3.0.json
jyapayne: Alright, so just to v3.1 then. I'll remove this.

compose/config/config.py
jyapayne: Alright, I'll remove it!

jyapayne:
@shin- rebased and issues mentioned fixed! Hopefully the build passes this time.

shin-:
Thank you!

compose/config/config_schema_v3.1.json
```diff
@@ -71,7 +71,8 @@
               "properties": {
                 "context": {"type": "string"},
                 "dockerfile": {"type": "string"},
-                "args": {"$ref": "#/definitions/list_or_dict"}
+                "args": {"$ref": "#/definitions/list_or_dict"},
+                "cache_from": {"type": "#/definitions/list_of_strings"}
```
dnephin: This doesn't seem to work. I think this should be `$ref` ?

jyapayne: Should I open a new PR then?

dnephin: Nope, it should be fixed by #4616

cerdiogenes:
I'm receiving `unsupported option: 'cache_from'`, is it supported in `docker-compose version 1.13.0, build 1719ceb`?

shin-:
Only for the following Compose file formats: 2.2, 3.2 and 3.3

cerdiogenes:
@shin- when using `docker-compose.override.yml` the cache_from options is ignored, is this a expected behavior?

shin-:
We fixed that bug in the 1.14 RC.

PR #4530, opened 2017-02-23
Replace "_" with "-" in build_container_name

mterron:
Replace "_" with "-" in build_container_name to create RFC 952 (https://tools.ietf.org/html/rfc952) compliant hostnames.
The valid tokens as per RFC are A-Z (case insensitive),0-9 and - (hyphen) with the first character being alphabetic.
RFC 1123 (https://tools.ietf.org/html/rfc1123) expands on this allowing the first character in the label to be a digit too.

Signed-off-by: Miguel Terron <miguel.a.terron@gmail.com>

### DIFF
--- a/tests/acceptance/cli_test.py
+++ b/tests/acceptance/cli_test.py
@@ -1918,8 +1918,8 @@ class CLITestCase(DockerClientTestCase):
         self.dispatch(['up', '-d'])
         result = self.dispatch(['top'])
 
-        self.assertIn('top_service_a', result.stdout)
-        self.assertIn('top_service_b', result.stdout)
+        self.assertIn('top-service-a', result.stdout)
+        self.assertIn('top-service-b', result.stdout)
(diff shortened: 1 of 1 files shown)


mterron:
Trying to understand what is failing but I'm quite lost, if someone wants to take a look it'll be appreciated.

On AppVeyor fail I don't get how the first test can ever be true since they are the same:

```python
    def test_name_without_project(self):
        self.container_dict['Name'] = "/composetest-web-7"
        container = Container(None, self.container_dict, has_been_inspected=True)
        self.assertEqual(container.name_without_project, "web-7")

    def test_name_without_project_custom_container_name(self):
        self.container_dict['Name'] = "/custom-name-of-container"
        container = Container(None, self.container_dict, has_been_inspected=True)
        self.assertEqual(container.name_without_project, "custom-name-of-container")
```
For Jenkins, it seems it fails to run apt-get, not much I can do about it I guess.

PR #4536, opened 2017-02-25
Align status output for parallel_execute

edsrzf:
Previously docker-compose would output lines that looked like:

    Starting service ... done
    Starting short ...
    Starting service-with-a-long-name ... done

It's difficult to scan down this output and get an idea of what's happening.

Now the statuses are aligned, and output looks like this:

    Starting service                  ... done
    Starting short                    ...
    Starting service-with-a-long-name ... done

To me, this is quite a bit easier to read.

### DIFF
--- a/compose/parallel.py
+++ b/compose/parallel.py
@@ -35,7 +35,8 @@ def parallel_execute(objects, func, get_name, msg, get_deps=None):
 
     writer = ParallelStreamWriter(stream, msg)
     for obj in objects:
-        writer.initialize(get_name(obj))
+        writer.add_object(get_name(obj))
+    writer.write_initial()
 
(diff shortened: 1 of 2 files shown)


edsrzf:
Ping! This seems relatively straight forward. Any chance of getting it into 1.12.0?

edsrzf:
Ping! Again! (I don't want to start @-mentioning people directly, but I will if I have to.)

edsrzf:
@shin- Have you had a chance to take a look at this? Sorry for the @-mention, but it was a last resort.

shin-:
I was somewhat on the fence about this one, but now that I consider it again, it's clearly an improvement.

Code and tests all look good.

Thank you for your patience, and your help!

PR #4543, opened 2017-02-28
Fixes #4522: Adds cli option --ignore-override

### LINKED ISSUE #4522: Add CLI option to ignore docker-compose.override.yml
`docker-compose up` and friends automatically read the `docker-compose.override.yml` file if present. Sometimes it would come handy to disable this behavior. Something like `docker-compose --ignore-override up`.

dennisgove:
For issue #4522, adds cli option --ignore-override

When option `--ignore-override` is provided the docker-compose.override.yml, if it exists, will not be considered for any command.

### DIFF
--- a/compose/cli/command.py
+++ b/compose/cli/command.py
@@ -33,7 +33,8 @@ def project_from_options(project_dir, options):
         verbose=options.get('--verbose'),
         host=host,
         tls_config=tls_config_from_options(options),
-        environment=environment
+        environment=environment,
+        ignore_override=options['--ignore-override']
     )
 
 
@@ -93,10 +94,10 @@ def get_client(environment, verbose=False, version=None, tls_config=None, host=N
 
 
 def get_project(project_dir, config_path=None, project_name=None, verbose=False,
-                host=None, tls_config=None, environment=None):
+                host=None, tls_config=None, environment=None, ignore_override=False):
     if not environment:
         environment = Environment.from_env_file(project_dir)
-    config_details = config.find(project_dir, config_path, environment)
+    config_details = config.find(project_dir, config_path, environment, ignore_override)
     project_name = get_project_name(
         config_details.working_dir, project_name, environment
     )
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -168,6 +168,7 @@ class TopLevelCommand(object):
       --skip-hostname-check       Don't check the daemon's hostname against the name specified
                                   in the client certificate (for example if your docker host
                                   is an IP address)
+      --ignore-override           Ignore docker-compose.override.yml file during this run
 
     Commands:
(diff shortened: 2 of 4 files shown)


shin-:
Sorry, I guess I'm failing to see the point of this feature - if you want to selectively ignore a partial Compose file some of the time, why not call it anything other than `docker-compose.override.yml` and simply use the `-f` flag when you want to include it?

This kind of feels like bloat / feature duplication to me.

dennisgove:
I certainly understand that sentiment. My thinking here is that because `docker-compose.override.yml` is automatically read if it exists that there should be a way to turn that off. In my mind it's really a 6 of one half dozen of the other whether that's accomplished with `--ignore-override` or explicit use of `-f docker-compose.yml` to accomplish the same result.

If you'd rather go with the `-f docker-compose.yml` then I think I'd like to change the documentation referenced in https://github.com/docker/docker.github.io/pull/2019 to at least mention that `-f` should be used when one wants to ignore the override file.

philtay:
@shin- What if a project already has a `docker-compose.override.yml` file but I have to use only `docker-compose.yml` to avoid side effects (networks and volumes creation, starting unrelated containers, etc)? I can use the `-f` option, but this forces me to specify the obvious every time I use Compose (i.e. an error-prone `docker-compose -f docker-compose.yml`). More in general: there should be a non-intrusive way to disable any automatic behavior.

dennisgove:
@shin- , what're your thoughts on this? The documentation change in https://github.com/docker/docker.github.io/pull/2019 was prematurely accepted so if you don't think this feature will make it in then I'd like to get that PR reverted.

shin-:
Um, I think in that case it's preferable to avoid feature duplication. I don't think `docker-compose --ignore-override up` would be superior in terms of user experience to `docker-compose -f docker-compose.yml up`.

> If you'd rather go with the -f docker-compose.yml then I think I'd like to change the documentation referenced in docker/docker.github.io#2019 to at least mention that -f should be used when one wants to ignore the override file.

I do agree with making that clearer if we can.

dennisgove:
Thanks very much for your help with this. I'll go ahead and submit a new PR to .io for the documentation update.

matiasanaya:
@shin- I need to run this in several directories, and since both `docker-compose.yml` and `docker-compose.yaml` are supported I have to run something similar to:
```
docker-compose -f docker-compose.yml config || docker-compose -f docker-compose.yaml config
```
This line is assuming the preference order and is also assuming these are the only supported files.

I understand you suggest naming `docker-compose.override.yml` differently and that maybe you don't think it's warranted to provide a `--ignore-override` but I'd argue that `docker-compose -f docker-compose.yml config` is not equivalent to the proposed `docker-compose --ignore-override config`.

PR #4546, opened 2017-03-01
Allowing log output to be filtered using '--filter-log'

rawkode:

I'm sure there should be a test for this, but I am unsure what that would look like. Some help would be appreciated :smile:

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -854,7 +854,7 @@ class TopLevelCommand(object):
                                        running. (default: 10)
             --remove-orphans           Remove containers for services not
                                        defined in the Compose file
-            --filter-log "SERVICES"    Only show log output from services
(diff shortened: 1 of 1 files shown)


rawkode:
Use-Case:

We all want the single command development environment, `docker-compose up`, but generally we only want log output from our application and not it's supporting services.

Given a `docker-compose.yml` with `elasticsearch`, `kafka`, `postgres` and our application, we can now issue `docker-compose up --filter-log application` and get to work

shin-:
Hi David,

Thank you for your contribution! I'm slightly confused here, is this adding significant value compared to e.g. `docker-compose up -d && docker-compose logs <application>`?

rawkode:
@shin- no, it's syntactic sugar, though I think the use case is common enough to warrant the flag

PR #4554, opened 2017-03-01
Ability to set "privileged" flag (boolean) from environment var

snowsky:
Signed-off-by: Hao Wang <hao.1.wang@gmail.com>


snowsky:
ping

shin-:
Should be fixed by #5291

PR #4555, opened 2017-03-02

kbh2o:
This replaces #4466 since I messed up when squashing commits.  


jakubsacha:
@kbh2o thanks for doing it!

PR #3723, opened 2016-07-12
add memory-reservation support

yank1:
The docker 1.12 would support memory-reservation.
It's very important production feature.

@aanand it's OK now
Ref #3712

PR #4563, opened 2017-03-03
Add mem_reservation option to service config in 2.0 and 2.1 formats

shin-:
Replaces #3723 
Implements #3712


AlexeyRokhin:
compose/config/config.py misses comma after mem_limit

PR #4564, opened 2017-03-03
support --build-arg for build command

shin-:
Rebased #3837 
Implements #3790

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -209,18 +209,26 @@ class TopLevelCommand(object):
         e.g. `composetest_db`. If you change a service's `Dockerfile` or the
         contents of its build directory, you can run `docker-compose build` to rebuild it.
 
-        Usage: build [options] [SERVICE...]
+        Usage: build [options] [--build-arg key=val...] [SERVICE...]
 
         Options:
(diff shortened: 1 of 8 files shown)


compose/cli/main.py
```diff
@@ -209,18 +209,26 @@ def build(self, options):
...
+            --force-rm              Always remove intermediate containers.
+            --no-cache              Do not use cache when building the image.
+            --pull                  Always attempt to pull a newer version of the image.
+            --build-arg key=val     Set build-time variables for one service.
         """
+        service_names = options['SERVICE']
+        build_args = options.get('--build-arg', None)
+
+        if not service_names and build_args:
+            raise UserError("Need service name for --build-arg option")
```
artagnon: Why do this? If the user doesn't specify a service, apply it to all services; the Dockerfile for a service can choose to ignore the argument entirely, so there's no problem (builds used to fail for unconsumed arguments, but that's no longer the case).

yairkukielka: Another alternative is a way to specify for what service the build-arg is for, even if you want to build all the services

davewood:
sorry for digging this up.

does that mean if i pass --build-arg i cannot build all services at once anymore?

Issue #3731, opened 2016-07-14
Set volume name


devent:
Hello,
can I set the volume name if I also use a volume driver? I need to set the volume name to something like vol1 for convoy. If that's not possible with compose, please make it a feature request.

```
docker-compose up
Creating postgresqlconvoy_postgres_data_1

ERROR: for postgres_data  Failed to execute: mount [/dev/mapper/postgresqlconvoy_vol1 /var/lib/rancher/convoy/devicemapper/mounts/postgresqlconvoy_vol1], output mount: wrong fs type, bad option, bad superblock on /dev/mapper/postgresqlconvoy_vol1,
       missing codepage or helper program, or other error

       In some cases useful info is found in syslog - try
       dmesg | tail or so.
, error exit status 32
ERROR: Encountered errors while bringing up the project.
Makefile:22: recipe for target 'run' failed
make: *** [run] Error 1
```

```
version: '2'

services:
  postgres_data:
    image: erwinnttdata/postgresql:latest
    restart: "no"
    command: "true"
    volumes:
      - "vol1:/var/lib/postgresql/data"
    volume_driver: "convoy"

  postgres:
    image: erwinnttdata/postgresql:latest
    restart: "always"
    depends_on:
      - postgres_data
    volumes_from:
      - postgres_data

volumes:
    vol1:
        driver: "convoy"
```

aanand:
No, it's not possible. If you want control over the name, create the volume yourself first and refer to it in the Compose file as `external`:

```yaml
volumes:
  vol1:
    external:
      name: my-volume-name
```

mrmunch:
Can we handle this as a change request?
It would be rellay nice if I can set the volume name for volumes I create with docker-compose.
For now, there will be a prefix added at the volume name, but it would be nice if we can avoid this.

Gallaecio:
Just to be clear, this is a request to support the following syntax:
```yaml
volumes:
  volume_name:
    name: docker_volume_name
```
Where the `name` property of a `volumes` entry would be to volumes what the `container_name` property of a `services` entry is to containers.

dcoraboeuf:
Some storage providers, like when using Digital Ocean Block Storage with RexRay, do not support underscores (_) and I'd need therefore to provide a custom name :(

dcoraboeuf:
I've created a PR at https://github.com/docker/compose/pull/4573 to have an additional `actual_name` property, which can be used as the "real name" when creating the volume.

shin-:
[...]

PR #4573, opened 2017-03-04
#3731 Allows volumes to specify an `actual_name` property

dcoraboeuf:
Allows volumes to specify an `actual_name` property, allowing to by pass some volume driver/hosting limitations.

### DIFF
--- a/compose/volume.py
+++ b/compose/volume.py
@@ -14,11 +14,12 @@ log = logging.getLogger(__name__)
 
 
 class Volume(object):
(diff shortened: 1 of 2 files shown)


dcoraboeuf:
I just realized that this change won't be enough when used with `docker stack deploy`... which is exactly the target I wished for.

shin-:
There is a proposal for that issue on the `docker stack` side here: https://github.com/docker/cli/issues/274#issuecomment-312315877

Once that's implemented, we'll make the necessary adaptations on our side. I'll close this PR in the meantime.

PR #4586, opened 2017-03-08
Add bash completion for `run --volume`

albers:
Bash completion for #4566.


shin-:
Thank you!

PR #4587, opened 2017-03-08
Fix bash completion for `docker-compose images`

albers:
Amends #4561.
Bash completion for the images command was missing:

    docker-compose i<tab>

Also changes formatting to use tabs instead of spaces.

Ping @sdurrheimer for zsh completion: new command `images`


shin-:
Thanks! :)

PR #4592, opened 2017-03-09
Add bash completion for `docker-compose rm --stop`

albers:
Bash completion for #4040.

If `--stop` was given on the command line, all services are completed. Else, (as before this change) only stopped services are completed.
We have the same logic for `docker rm --force`.


shin-:
Thank you!

PR #4605, opened 2017-03-11
Ensure network config matches remote for all properties

### LINKED ISSUE #4600: Network config not applied when docker-compose.yml changes
When configuration changes in the `networks` block of the docker-compose.yml it looks like docker doesn't recreate the network and apply the latest change.

Eg. running `docker-compose up` with the following docker-compose.yml
```
version: '3'

services:
  web:
    image: nginx:alpine
    ports:
      - 80
      - 443
    networks:
      secret_network:

networks:
  secret_network:
```

results in `docker ps` showing that the ports are correctly exposed
```
CONTAINER ID        IMAGE               COMMAND                  CREATED             STATUS              PORTS                                           NAMES
f438fcb53922        nginx:alpine        "nginx -g 'daemon ..."   5 seconds ago       Up 4 seconds        0.0.0.0:32805->80/tcp, 0.0.0.0:32804->443/tcp   insightinformatics_web_1
```

However, if we change the docker-compose.yml to the following and run `docker-compose up` again
```
version: '3'

services:
  web:
    image: nginx:alpine
    ports:
      - 80
      - 443
    networks:
      secret_network:

networks:
  secret_network:
    internal: true    # oops, forgot this!
```

then `docker ps` shows that the ports are still exposed when they should no longer be
```
CONTAINER ID        IMAGE               COMMAND                  CREATED             STATUS              PORTS                                           NAMES
f438fcb53922        nginx:alpine        "nginx -g 'daemon ..."   2 minutes ago       Up 7 seconds        0.0.0.0:32807->80/tcp, 0.0.0.0:32806->443/tcp   insightinformatics_web_1
```

Manually running `docker network rm insightinformatics_secret_network` to remove the network, then rerunning `docker-compose up` has the expected effect and secret_network is now internal and `docker ps` shows no ports are exposed.
```
CONTAINER ID        IMAGE               COMMAND                  CREATED             STATUS              PORTS               NAMES
f438fcb53922        nginx:alpine        "nginx -g 'daemon ..."   6 minutes ago       Up 10 seconds                           insightinformatics_web_1
```

I'm not sure what expected behaviour is here, but I was expecting it to update the network config in the same way that container config would be updated if a change was made and then applied.

```
Docker version 1.13.0, build 49bf474
docker-compose version 1.11.2, build f963d76f
```

shin-:
Fixes #4600

### DIFF
--- a/compose/network.py
+++ b/compose/network.py
@@ -126,22 +126,64 @@ def create_ipam_config_from_dict(ipam_dict):
     )
 
 
+class NetworkConfigChangedError(ConfigurationError):
+    def __init__(self, net_name, property_name):
+        super(NetworkConfigChangedError, self).__init__(
+            'Network "{}" needs to be recreated - {} has changed'.format(
+                net_name, property_name
+            )
+        )
+
+
+def check_remote_ipam_config(remote, local):
+    remote_ipam = remote.get('IPAM')
+    ipam_dict = create_ipam_config_from_dict(local.ipam)
+    if local.ipam.get('driver') and local.ipam.get('driver') != remote_ipam.get('Driver'):
+        raise NetworkConfigChangedError(local.full_name, 'IPAM driver')
+    if len(ipam_dict['Config']) != 0:
+        if len(ipam_dict['Config']) != len(remote_ipam['Config']):
+            raise NetworkConfigChangedError(local.full_name, 'IPAM configs')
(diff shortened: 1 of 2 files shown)


Novex:
Would the [labels](https://docs.docker.com/compose/compose-file/#labels-3) property also need to be checked?

PR #4616, opened 2017-03-13
Synchronize the schema with docker/docker

dnephin:
Related PR https://github.com/docker/docker/pull/31795

The v3.1 schema in both repos should be reverted to the version we released, and a new schema is created for the next release.

TODO: fix the code to support this version


compose/config/config.py
```diff
@@ -478,7 +475,7 @@ def process_config_file(config_file, environment, service_name=None):
         'service',
         environment)
 
-    if config_file.version in (V2_0, V2_1, V3_0, V3_1):
+    if config_file.version > V1:
```
shin-: That's a string comparison, right? Any chance this could break? Can we just say `if config_file.version != V1` instead?

compose/config/config.py
```diff
@@ -493,10 +490,6 @@ def process_config_file(config_file, environment, service_name=None):
             environment)
     elif config_file.version == V1:
```
shin-: This can be a simple `else` now, I believe.

shin-:
Some `ImportError`s still in the tests.

compose/config/config.py
dnephin: It works for now, but yes I could see it breaking later, especially if we add patch versions. A real `Version` object would be a lot safer. I'll use ` != V1` for now.

shin-:
https://github.com/docker/compose/blob/master/tests/fixtures/ports-composefile/expanded-notation.yml needs to be bumped to 3.2 for that last test to pass.

dnephin:
hmm, I made that change, but the test is still failing for me. I'm not sure what's going on. It seems to be loading the correct schema.

compose/config/config.py
```diff
@@ -493,19 +490,13 @@ def process_config_file(config_file, environment, service_name=None):
...
-                'secrets',
-                environment
-            )
-    elif config_file.version == V1:
-        processed_config = services
+        processed_config['secrets'] = interpolate_config_section(
+            config_file,
+            config_file.get_secrets(),
+            'secrets',
+            environment)
```
dnephin: secrets is already checked against the schema during validation and the config version in `Config`, so I think we can avoid checking it again here.

shin-:
@dnephin Added a few changes in 7b1900951199628f879acce219a26619827a0b35, including to the v3.2 schema itself. I think the typo in the schema's ID was causing the issue with the expanded ports not being validated properly for some reason.

compose/config/config.py
```diff
@@ -493,19 +490,14 @@ def process_config_file(config_file, environment, service_name=None):
             config_file.get_networks(),
             'network',
             environment)
-        if config_file.version in (V3_1,):
+        if config_file.version in (const.COMPOSEFILE_V3_1, const.COMPOSEFILE_V3_2):
```
dnephin: The reason I removed this line is because it adds additional steps every time we add a new version.

We should really be able to add a new version by just dropping in a schema, so I'm trying to move in that direction.

Is this check actually necessary? Don't we already check this in two places:
* first in schema validation
* again in `ConfigFile.get_secrets()`

compose/config/serialize.py
```diff
@@ -45,7 +46,7 @@ def denormalize_config(config):
         if 'external_name' in vol_conf:
             del vol_conf['external_name']
 
-    if config.version in (V3_1,):
+    if config.version in (V3_1, V3_2):
```
dnephin: Can we change this to `< V3_1` for the same reason? We don't want to have to remember to add to this list every time we change a version.

Maybe we should go even further and have a function like `version_supports_secrets(version)`, and use it both here and in `compose/config/config.py` ?

compose/config/serialize.py
```diff
@@ -103,7 +104,7 @@ def denormalize_service_dict(service_dict, version):
                 service_dict['healthcheck']['timeout']
             )
 
-    if 'ports' in service_dict and version != V3_1:
+    if 'ports' in service_dict and version not in (V3_2,):
```
dnephin: same here, maybe we should make this `< V3_2`

dnephin:
> think the typo in the schema's ID was causing the issue with the expanded ports not being validated properly

Ahh, that makes sense, nice find! I believe jsonschema caches schemas by id. Since the tests would have loaded the V3_1 schema first, it would have it cached.

dnephin:
Let me know what you think. I agree comparing strings with `<` is a bit worrisome, but I think we can fix that. https://pypi.python.org/pypi/semver looks like a good option.

Maybe we can do that as a follow up PR?

compose/config/config.py
shin-: Specifically in that case, it messes up with a validation step later on because there's an unexpected `secrets` key in the resulting dictionary

shin-:
I agree that our version checks are getting unwieldy now, but a follow-up PR would probably be a better place to fix it.

dnephin:
Cool, sounds good

shin-:
Cool, I'll merge this then :+1:

PR #4624, opened 2017-03-16
Pass the --disable-pip-version-check option to pip freeze

hashpling:
I noticed this in an environment with no direct internet access. A new version of docker-compose appeared to be hanging for a few seconds on every invocation and it turned out that this was due to pip attempting to check whether it was up to date.

The `--disable-pip-version-check` option prevents this and has been part of pip's interface since version 6.

### DIFF
--- a/compose/cli/__init__.py
+++ b/compose/cli/__init__.py
@@ -13,7 +13,9 @@ try:
     # https://github.com/docker/compose/issues/4481
     # https://github.com/pypa/pip/blob/master/pip/_vendor/__init__.py
     s_cmd = subprocess.Popen(
-        ['pip', 'freeze'], stderr=subprocess.PIPE, stdout=subprocess.PIPE
+        ['pip', '--disable-pip-version-check', 'freeze'],
+        stderr=subprocess.PIPE,
+        stdout=subprocess.PIPE
     )
     packages = s_cmd.communicate()[0].splitlines()
     dockerpy_installed = len(


jbrwilkinson:
LGTM

shin-:
The issue here is that this command would probably fail if the pip version on the user's machine is older than that. I understand the inconvenience when offline, but I'd like to aim for a better solution here.
Maybe a `COMPOSE_SKIP_PIP_CHECK` environment variable that just ignores the entire block.

hashpling:
That's a fair point, however I found it impossible to get an environment installed where pip was sufficiently old such that it doesn't have `--disable-pip-version-check` (i.e. 1.5.6) and that `python-requests` was sufficiently modern that `docker-compose` actually makes it as far as executing this test. If you try to get a really old pip to upgrade requests without upgrading itself, it ends up in a broken state where it can no longer upgrade anything.

I don't know that this is anything more than a hypothetical concern, but I'm happy to be proved wrong and do something better.

shin-:
@hashpling Thanks - I think I have found an alternative using an environment variable instead - can you check out #4698 and see if that solves your issue as well?

hashpling:
Thanks. In principle I'm happy for #4698 to replace this fix. I've ask about a couple of minor points over there. This can be closed once that lands or whenever you feel is more appropriate.

shin-:
Closing in favor of #4698 - thanks again for your help.

PR #4625, opened 2017-03-16
Add bash completion for `build --build-arg`

albers:
bash completion for #4564.


shin-:
Thanks!

PR #4626, opened 2017-03-16
Add bash completion for `--project-directory`

albers:
Bash completion for #4589

Also features a minor refactoring in a separate commit:
>The old names were supposed to designate variables that get propagated
to the Docker daemon. This is not true for all contained options and
therefore confused me.
The new names focus on their placement after the top level command.


shin-:
Thanks!

PR #4629, opened 2017-03-17
fix misspell "compatibility" in script/ci

fate-grand-order:
Signed-off-by: fate-grand-order <chenjg@harmonycloud.cn>


shin-:
Thank you!

PR #4632, opened 2017-03-18
Add support for expanded mount/volume notation

shin-:


compose/config/config.py
```diff
@@ -1030,7 +1030,13 @@ def resolve_volume_paths(working_dir, service_dict):
 
 
 def resolve_volume_path(working_dir, volume):
-    container_path, host_path = split_path_mapping(volume)
+    if isinstance(volume, dict):
+        host_path = volume.get('source')
+        container_path = volume.get('target')
+        if host_path and volume.get('read_only'):
+            container_path += ':ro'
```
dnephin: I think there is also `no-copy`(from volume.nocopy) and the mount propagation flags that would need to be appended here.

It might be safer to convert everything to a dict, instead of converting to a string?

shin-: Yeah, definitely agreed. I want to do a dict conversion eventually, but it involves much larger changes that I didn't want to make so close to the release. I want the change to offer the same amount of functionality as the legacy notation in the meantime. Good call on the missing `nocopy` handling - I'll make a follow-up PR for it.

Issue #3936, opened 2016-09-12
Send docker-compose to daemon mode without restart


aug-riedinger:
As from this [SO question](http://stackoverflow.com/questions/39439689/docker-compose-send-to-daemon-mode-without-restart?noredirect=1#comment66202483_39439689):

I like to `docker-compose up` first to check that everything started correctly.

Once everything's fine, I'd like to being able to send the command to daemon mode, without having to `docker-compose stop` and `docker-compose -d up` (two steps + service interruption).

shin-:
Maybe you could run `docker-compose -d up`, then `docker-compose logs` to check the status of your app. `logs` can then be exited at any time without affecting your application.

aug-riedinger:
Indeed, but since the `-d` option is not default, I tend to forget about it in the first place and then find myself *stuck*.

It is not so bad, but I reckon the feature would make sense, wouldn't it?

There are reasons for Unix's Ctrl-Z to have been created

shin-:
What about `Ctrl-Z`, `bg` and maybe `disown`, then? It seems to me UNIX already gives you all the tools you need in that case.

aug-riedinger:
I tried `Ctrl-z`, `bg` but could not exit the ssh session, receiving `There are stopped jobs.` I don't know about `disown` though, I'll try it out.

TristanCP:
So, can you confirm `disown` does the trick without side effects ?

PR #4634, opened 2017-03-18
Adopting background before stop and kill

### LINKED ISSUE #4560: key sequence to detach from `docker-compose up`
this has been topic in #291, however i want to make aware of this feature request a little more .

i vote for a key-sequence that you can use to detach from `docker-compose up` without stopping all containers (i.e. when you forgot to pass `-d`). There seems to be a key-sequence for attached docker-containers (`Ctrl-p` `Ctrl-c`). It seems plausible to reuse this in docker-compose.

rawkode:
Adopting 'background' first on Cntrl-C and allowing second and third taps to be stop and kill

Would fix #4560, though I am aware this changes current behaviour and perhaps should be tackled a different way.

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -10,6 +10,7 @@ import pipes
 import re
 import subprocess
 import sys
+import time
 from distutils.spawn import find_executable
 from inspect import getdoc
 from operator import attrgetter
@@ -51,7 +52,6 @@ from .log_printer import LogPrinter
 from .utils import get_version_info
 from .utils import human_readable_file_size
 from .utils import yesno
-
 
 if not IS_WINDOWS_PLATFORM:
     from dockerpty.pty import PseudoTerminal, RunOperation, ExecOperation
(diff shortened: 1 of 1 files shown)


rawkode:
Although I said this "changes current behaviour" ... it is a "soft" change and an additional press would replicate original behaviour.

Thoughts, @shin- ?

shin-:
Thanks for taking the time to submit this!

From a user experience standpoint, this seems like a net loss to me. It means pressing Ctrl-C now do 3 different things depending on how many times you press it, and one of those things is significantly different from the 2 other.

Also, as I mentioned in #3936, this can already be done with some UNIX magic. As a result, I would be happy to merge support for something like the `Ctrl-P Ctrl-C` combination (since it seems to be what the engine uses for attached containers), but anything short of that seems too subpar to justify.

rawkode:
Perhaps it's just my opinion, but control sequences, like C-P C-Q, are so esoteric.

Perhaps on C-c, we can prompt for input: Background, Stop or Kill? with a default after a few seconds to normal behaviour?

rawkode:
Or I could let my shell handle it I guess 😂

shin-:
> Perhaps it's just my opinion, but control sequences, like C-P C-Q, are so esoteric.

You won't hear me disagree on that one. But it would at least be somewhat consistent with Engine, so there's possible transfer of knowledge there.

> Perhaps on C-c, we can prompt for input: Background, Stop or Kill? with a default after a few seconds to normal behaviour?

Well, do you agree with me that this is still a straight-up worse UX for people who want to stop/kill, and only marginally better for people who want to detach? The tradeoff just doesn't seem worthwhile to me.

rawkode:
Yeah, I'm getting there 👍

PR #4641, opened 2017-03-20
Add bash completion for `pull --parallel`

albers:
Bash completion for #4428.
Ping @sdurrheimer for zsh completion.


shin-:
Thank you! =)

PR #4661, opened 2017-03-23
Support 'nocopy' mode for expanded volume syntax

shin-:
Follow-up to https://github.com/docker/compose/pull/4632#discussion_r106985480


shin-:
Test failure unrelated

Issue #4583, opened 2017-03-07
Limit disk space : Set storage driver options per container 


ssiegl-iwes:
Hello,

would it be possible to add the limits for disk space per container with docker compose.

Like: https://github.com/docker/docker/blob/master/docs/reference/commandline/run.md#set-storage-driver-options-per-container

BR

Stefan

PR #4680, opened 2017-03-30
Set storage options per container

Dineshs91:
Add the ability to set the storage option per container.

Fixes https://github.com/docker/compose/issues/4583

I've added this to compose version `3.2` only. Should I be adding it to all the `3.*` versions ?

### DIFF
--- a/compose/config/config_schema_v2.1.json
+++ b/compose/config/config_schema_v2.1.json
@@ -228,6 +228,7 @@
         "stdin_open": {"type": "boolean"},
         "stop_grace_period": {"type": "string", "format": "duration"},
         "stop_signal": {"type": "string"},
+        "storage_opt": {"type": "object"},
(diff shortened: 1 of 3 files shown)


shin-:
Hi! Thank you for submitting this.

The main issue with this is that `storage_opts` is a container configuration, not a (Swarm) service configuration, and as such, needs to be added to the 2.1 format, but has no place in the 3.x format.

Dineshs91:
I've updated the PR with the changes.

Dineshs91:
@shin- Can I specify the test to use a particular storage driver ? I think this is the reason why the test case is failing.

rammaram06:
@shin- Any plans on merging this..we are unable to use docker-compose becoz of this missing feature.

shin-:
Merged via #4956 

Thank you again for your help!

PR #4682, opened 2017-03-30
Add zsh completion for 'docker-compose images'

sdurrheimer:
ref. #4561

Signed-off-by: Steve Durrheimer <s.durrheimer@gmail.com>


shin-:
Thanks!

PR #4698, opened 2017-04-03
Prevent pip version checks when calling `pip freeze`

shin-:
Replace #4624 

Using the environment variable instead of the command-line option allows us to support versions of pip before the option was introduced.

Reference: https://github.com/pypa/pip/issues/2269#issuecomment-68002027

### DIFF
--- a/compose/cli/__init__.py
+++ b/compose/cli/__init__.py
@@ -2,6 +2,7 @@ from __future__ import absolute_import
 from __future__ import print_function
 from __future__ import unicode_literals
 
+import os
 import subprocess
 import sys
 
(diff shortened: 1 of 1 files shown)


hashpling:
So this did seem to help my issue but one thing that I noticed while verifying is that it picked up `pip` from a different PATH. This wasn't an issue in my set up but I wondered if it might be a problem more generally.

Was it intentional to clear the environment for the invoked `pip`? It might be a less intrusive change to use a copy of `os.environ` with the addition of `PIP_DISABLE_PIP_VERSION_CHECK=1` rather than a completely clean environment.

This may be my Python ignorance, but what's the benefit of using `{ str('PIP_DISABLE_PIP_VERSION_CHECK'): str('1') }` over simply doing `{'PIP_DISABLE_PIP_VERSION_CHECK': '1'}`?

shin-:
Thank you for the feedback!

> This may be my Python ignorance, but what's the benefit of using `{ str('PIP_DISABLE_PIP_VERSION_CHECK'): str('1') }` over simply doing `{'PIP_DISABLE_PIP_VERSION_CHECK': '1'}`?

It solves a [Windows-specific issue](http://stackoverflow.com/questions/12253014/why-does-popen-fail-on-windows-if-the-env-parameter-contains-a-unicode-object) where the environment can only be bytestrings on Python 2 and Unicode strings on Python 3.

hashpling:
I've test this latest version in our environment and it looks good to me. :+1:

shin-:
Thank you for verifying!

PR #4709, opened 2017-04-06
Make self.project_dir not hardcoded to '.' in TopLevelCommand

joshgordon:
I was digging through cli/main.py to try to figure some stuff out and I noticed something a bit odd. Figured I'd open a PR for a fix.

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -203,7 +203,7 @@ class TopLevelCommand(object):
 
     def __init__(self, project, project_dir='.'):
         self.project = project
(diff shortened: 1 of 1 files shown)


compose/cli/main.py
```diff
@@ -111,7 +111,7 @@ def perform_command(options, handler, command_options):
         handler(command, options, command_options)
         return
 
-    project = project_from_options('.', options)
+    project = project_from_options(self.project_dir, options)
```
shin-: This ought to be `command.project_dir`, I think.

PR #4716, opened 2017-04-08
Add support for service.init

### LINKED ISSUE #4415: Docker v1.13 init feature to use in docker compose
It would be great to have the docker v1.13 init feature in Compose v3 so as to provision the containers with docker compose with an init process running as PID 1 inside the containers.

shin-:
Fixes #4415


Silex:
@shin- : in what docker-compose version is this supposed to ship? Just tried with `16.1` and using `init: true` does not work. Is this a version 3 only feature?

shin-:
@Silex it's only available in v2.2 and 2.3 formats.

Silex:
@shin-: ah, okay thanks!

PR #4730, opened 2017-04-12
Add --quiet parameter to docker-compose pull, using existing silent parameter

### LINKED ISSUE #3770: Option to suppress progress
There should be an option that still shows the build steps, but not the actual progress bars which spray a huge amount of output on CI servers since the docker commands expect that your using a terminal that has the control codes to write over previous lines.

jbarciauskas:
Resolves #3770

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -634,11 +634,13 @@ class TopLevelCommand(object):
         Options:
             --ignore-pull-failures  Pull what it can and ignores images with pull failures.
             --parallel              Pull multiple images in parallel.
+            --quiet                 Pull without printing progress information
         """
         self.project.pull(
             service_names=options['SERVICE'],
             ignore_pull_failures=options.get('--ignore-pull-failures'),
-            parallel_pull=options.get('--parallel')
+            parallel_pull=options.get('--parallel'),
+            silent=options.get('--quiet'),
         )
 
     def push(self, options):
--- a/tests/acceptance/cli_test.py
+++ b/tests/acceptance/cli_test.py
@@ -431,6 +431,10 @@ class CLITestCase(DockerClientTestCase):
         assert ('repository nonexisting-image not found' in result.stderr or
                 'image library/nonexisting-image:latest not found' in result.stderr)
 
+    def test_pull_with_quiet(self):
+        assert self.dispatch(['pull', '--quiet']).stderr == ''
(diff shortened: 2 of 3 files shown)


compose/cli/main.py
```diff
@@ -633,11 +633,13 @@ def pull(self, options):
         Options:
             --ignore-pull-failures  Pull what it can and ignores images with pull failures.
             --parallel              Pull multiple images in parallel.
+            --quiet                 Pull without printing progress information
```
shin-: I think `--no-progress` might be a better, more descriptive name for this option. `--quiet` could mean a number of different things.

tests/acceptance/cli_test.py
```diff
@@ -436,6 +436,9 @@ def test_pull_with_ignore_pull_failures(self):
         assert ('repository nonexisting-image not found' in result.stderr or
                 'image library/nonexisting-image:latest not found' in result.stderr)
 
+    def test_pull_with_quiet(self):
+        assert self.dispatch(['pull', '--quiet']).stdout == ''
```
shin-: I don't think that test is correct - all Compose status messages are written on `stderr`, including progress bars.

shin-:
Thank you for submitting this! 

It looks pretty good, but I have a few comments that need to be addressed before we merge this.

jbarciauskas:
Thanks, I will get to them shortly. I wasn't sure about the test, I just
copied another existing one. I should have gone back to double check, sorry!

compose/cli/main.py
jbarciauskas: Thinking about this again, it does suppress *all* output, and it is essentially a pass-through for a flag named "silent", so --quiet seems appropriate.

tests/acceptance/cli_test.py
jbarciauskas: The effect of setting silent=True is that all output is suppressed, so I will test both.

compose/cli/main.py
jbarciauskas: @shin- had forgotten this was still lying around, happy to switch this if it's necessary to merge it

shin-:
Thank you for bumping this. Do you mind squashing those commits together? Code looks good otherwise.

jbarciauskas:
@shin- all set!

shin-:
LGTM, thanks!

PR #4735, opened 2017-04-13
Add support for labels during build

### LINKED ISSUE #4038: Proposal: Support for building images with labels
Currently docker-compose supports to add labels to containers. But we with pure docker we can add labels to images with `docker build --label mylabel .` (https://docs.docker.com/engine/reference/commandline/build/)
Since we recently also gained support for labelling networks and volumes: https://github.com/docker/compose/pull/3898 I suggest to implement this for building images as well.

It could look like this:

```
version: '2'
services:
  myservice:
    build:
      context: .
      labels:
        - mylabelOne= "labelTextOne"
        - mylabelTwo= "labelTextTwo"
```

ColinHebert:
Adds support for labels when building an images.
cf.
https://github.com/docker/docker-py/blob/0a97df1abc7236793b82041626088f54ad8d204f/docker/api/build.py#L94

Fixes #4038

### DIFF
--- a/compose/config/config_schema_v3.2.json
+++ b/compose/config/config_schema_v3.2.json
@@ -72,6 +72,7 @@
                 "context": {"type": "string"},
                 "dockerfile": {"type": "string"},
                 "args": {"$ref": "#/definitions/list_or_dict"},
+                "labels": {"$ref": "#/definitions/list_or_dict"},
                 "cache_from": {"$ref": "#/definitions/list_of_strings"}
               },
               "additionalProperties": false
--- a/compose/service.py
+++ b/compose/service.py
@@ -831,6 +831,7 @@ class Service(object):
             nocache=no_cache,
             dockerfile=build_opts.get('dockerfile', None),
             cache_from=build_opts.get('cache_from', None),
+            labels=build_opts.get('labels', None),
             buildargs=build_args
         )
 
--- a/tests/unit/config/config_test.py
+++ b/tests/unit/config/config_test.py
@@ -845,8 +845,7 @@ class ConfigTest(unittest.TestCase):
         ).services[0]
         assert 'labels' in service['build']
         assert 'label1' in service['build']['labels']
-        assert isinstance(service['build']['labels']['label1'], str)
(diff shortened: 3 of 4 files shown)


ColinHebert:
I have not idea how the current tests are working but this is what I used in my dev environment to make sure it works:

docker-compose.yaml
``` yaml
version: '3.2'

services:
  web:
    build:
      context: .
      labels:
        test: blah
        test2: test
```

Dockerfile
```Docker
FROM ubuntu
```

Execution
```
root# docker-compose build
Building web
Step 1/2 : FROM ubuntu
 ---> 0ef2e08ed3fa
Step 2/2 : LABEL "test" 'blah' "test2" 'test'
 ---> Using cache
 ---> 6c6c2d57f195
Successfully built 6c6c2d57f195
```

shin-:
Thanks!

I think if you want to add something to the v3 format, you should submit a PR in the Engine repo first - they're the source of truth as far as v3 is concerned.

ColinHebert:
Uh? That sounds a bit.. unweildy.
In any case, I've opened docker/docker#32632 that updates the docker-compose schema.

ColinHebert:
@dnephin @shin-, with the PR being approved (but not merged) on the moby/moby repo, what are the next steps to get this merged?

ColinHebert:
@shin- would it be possible to get this one merged for the next release?

shin-:
Ah yes, sorry - it just needs to be moved to the new 3.3 schema as per https://github.com/moby/moby/pull/32972

ColinHebert:
I see, how do you suggest we proceed?
There are other PRs that are attempting to modify the schema (see moby/moby#32743) are we importing the schema each time it's modified?

Also note that the schemas now live in docker/cli (see docker/cli#32 to get 3.3 in docker-cli)

shin-:
I don't think we are at a point where submodule would actually improve the workflow here. The reorganization at the engine level with the new Moby project and the docker/cli project split has caused this process to be a bit more involved than usual, but it's not the norm.

Either way: Merged via #4858

PR #4736, opened 2017-04-13
Repair bad import

### LINKED ISSUE #4734: Secrets not export via docker-compose config 3.2
Example:
```yml
version: '3.2'
services:
  rabbitmq:
    image: 'Asd'
    secrets:
      - source: api-config-v1
        target: config.toml

secrets:
  api-config-v1:
    file: ~/nx_configs/config.toml
```

`docker-compose config` will export:
```yml
networks: {}
services:
  rabbitmq:
    image: Asd
    secrets:
    - source: api-config-v1
      target: config.toml
version: '3.2'
volumes: {}
```

It works fine with version 3.1
I may found the problem:
https://github.com/docker/compose/blob/master/compose/config/serialize.py#L10
https://github.com/docker/compose/blob/master/compose/config/serialize.py#L12

I think it should be:
```
from compose.const import COMPOSEFILE_V2_2 as V2_2
from compose.const import COMPOSEFILE_V3_2 as V3_2
```
https://github.com/docker/compose/blob/master/compose/config/serialize.py#L53
because variable `V3_2` contains `3.1` and the version in the config is `3.2`

shin-:
Fixes #4734

### DIFF
--- a/compose/config/serialize.py
+++ b/compose/config/serialize.py
@@ -7,9 +7,9 @@ import yaml
 from compose.config import types
 from compose.const import COMPOSEFILE_V1 as V1
 from compose.const import COMPOSEFILE_V2_1 as V2_1
-from compose.const import COMPOSEFILE_V2_1 as V2_2
+from compose.const import COMPOSEFILE_V2_2 as V2_2
 from compose.const import COMPOSEFILE_V3_1 as V3_1
-from compose.const import COMPOSEFILE_V3_1 as V3_2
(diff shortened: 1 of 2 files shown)


compose/config/serialize.py
```diff
@@ -9,7 +9,7 @@
 from compose.const import COMPOSEFILE_V2_1 as V2_1
 from compose.const import COMPOSEFILE_V2_1 as V2_2
```
Fank: I think you need to change this line too

Issue #3267, opened 2016-04-05
"docker-compose run -T" inconsistently outputs stdout / stderr?


cjerdonek:
I've noticed that `docker-compose run -T` seems to behave inconsistently from run to run with respect to outputting stderr and stdout.

For example, on repeated invocations of the following command:

```
$ docker-compose run -T web python3 -c "import sys; print(\"err\\n\", file=sys.stderr); print(\"stdout\\nstdout\\nstdout\\nstdout\\n\")"
```

the "err" usually doesn't print out (but sometimes does), and the "stdout" messages usually (but don't always) print.

This is a follow-up to issue #3239.

cjerdonek:
I'm using Compose version 1.5.1. I'm hesitant to upgrade and try on the latest version though because of breaking changes the last time I tried.

dnephin:
I've tried this a few dozen times with both master and 1.7.0rc1. I always see both streams, I always see stderr on stderr and stdout on stdout.  I never see any missing output. Sometimes the order is inverted, but that's to be expected with separate streams that use different buffers.

I can't reproduce the issue.

cjerdonek:
Okay, other things I can try are -- (1) try this on a stripped-down container, and (2) try with the newest release of compose. I should be able to try (1) within the next couple days. (2) may take a bit longer.

dnephin:
I used the `python:3` image to test it.

You can always just download the `docker-compose` binary to a local path, you don't have to install it to the system right away if you just want to test it out.

cjerdonek:
Good points. Thanks. Yes, I don't have to upgrade Docker toolbox. I'll try to try within the next few days.

cjerdonek:
Okay, I tried reproducing with compose 1.7.0rc1 and the plain python:3 image. About a third of the time, the err stream doesn't output, but the stdout stream always does.  With compose 1.5.1 and the plain python:3 image, I observed the same as what I originally reported. So it looks like going from 1.5.1 to 1.7.0rc1 makes things a bit better, but there is still an issue with flakiness for me.

dnephin:
Are you sure it's not just an issue with the order? Sometimes it's output after the stdout lines.

Are you running these tests on OSX? It may be an OSX issue, I was running the tests on Ubuntu.

cjerdonek:
No, it's not just the order. The total output is only 5 or 6 lines, so it's easy to see all the output. And yes, it's Mac OS X (should have made that clear from the beginning).

[...]

Issue #3239, opened 2016-03-30
document that `run -T` is needed to preserve stdout / stderr streams


cjerdonek:
In Docker Compose 1.5.1, I'm noticing that `docker-compose run` sends stderr from the running process to stdout of the process calling `run`. Is this intended?

[This comment](https://github.com/docker/compose/issues/2115#issuecomment-193420167) states that--

> docker-compose prints informational messages to stderr, and container output to the same stream as it was written to in the container (stdout or stderr).

but this isn't what I'm observing.

aanand:
Yeah, that's not quite accurate - we should document this.

By default, `run` creates a container with a pseudo-tty attached. This necessarily means that all output goes to stdout. You can disable this with the `-T` option, which will get you stdout/stderr mapping.

```console
$ docker-compose run web sh -c "echo stdout; >&2 echo stderr" 2> /dev/null
stdout
stderr

$ docker-compose run -T web sh -c "echo stdout; >&2 echo stderr" 2> /dev/null
stdout

$ docker-compose run -T web sh -c "echo stdout; >&2 echo stderr" > /dev/null
stderr
```

cjerdonek:
Great. Thanks so much for the explanation, @aanand. Should this issue be relabeled / left open as a documentation issue?

aanand:
Yeah, I've reclassified it as a docs issue.

cjerdonek:
@aanand Hmm, so I tried the `-T` option with `docker-compose run`, and I'm seeing non-deterministic behavior. Sometimes stdout is captured, and other times it isn't.

Is this a known issue? Is there known flakiness when capturing stdout with `-T`? Thanks a lot.

aanand:
That's not a known issue, no. We've had flakes of that kind in the past though. If you can produce a minimal failing case, please open a new issue for it.

cjerdonek:
Okay, thanks, @aanand. I'll open a separate issue, but I was able to reproduce with the following:

    $ docker-compose run -T web python3 -c "import sys; print(\"err\\n\", file=sys.stderr); print(\"foo\\nfoo\\nfoo\\nfoo\\n\")"

Half the time "err" doesn't  print, and 1 out of 8 times or so, the "foo" lines don't print.

ijc:
Issue grooming: Docs now say:
```
    -T                    Disable pseudo-tty allocation. By default `docker-compose run`
                          allocates a TTY.
```
which I think is what was being discussed. Closing.

PR #4738, opened 2017-04-14
Capture the beginning of service output during run

loomchild:
### Issue

`docker-compose up` always prints the entire container output, but `docker-compose run` sometimes skips the beginning of it. 

I wasn't able to reproduce this issue directly on local Linux machine, but when running docker-compose in docker container I can reproduce it every time.

Steps to reproduce:

1. Create the following `docker-compose.yml`:

        version: '2'
        services:
            main:
                image: debian
                command: ["bash", "-c", "echo -n Running; for i in {1..3}; do echo -n .; sleep 0.1; done; echo"]

2. Create the following `Dockerfile`:

        FROM python:3 
        RUN pip install docker-compose
        COPY docker-compose.yml /

3. Build the image

        docker build -t loomchild/docker-compose-bug .

4. Pull the debian image to avoid doing it later (you can skip this step if you already have debian)

        docker pull debian

5. Run the container:

        docker run -it -v /var/run/docker.sock:/var/run/docker.sock loomchild/docker-compose-bug bash

6. Execute docker-compose inside the container:

        docker-compose run main

Expected output:

    Running...

Actual output:

    ..

In other words everything before first `sleep` instruction is not displayed. However, when I add an additional sleep instruction at the beginning so the command in `docker-compose.yml` looks like:

    command: ["bash", "-c", "sleep 0.1; echo -n Running; for i in {1..3}; do echo -n .; sleep 0.1; done; echo"]

And rebuild the container and re-execute the test, the output is correct.

The described issue could be related to #3239 / #3267 since the behavior described there seems to be pseudo random and it happens inside a VM.

### Analysis

This seems to be caused by the fact that we start the container before initializing pseudoterminal (from cli/main.py, `run_one_off_container()`):

    ...
    pty = PseudoTerminal(project.client, operation)
    sockets = pty.sockets()
    service.start_container(container)
    pty.start(sockets)
    exit_code = container.wait()
    ... 

I see that this code was introduced in below commit:

    commit f3e55568d1a7c111398876bec84da8ed8b42da7f
    Author: Aanand Prasad <aanand.prasad@gmail.com>
    Date:   Thu Jan 21 18:34:18 2016 +0000

        Fix interactive run with networking
    
        Make sure we connect the container to all required networks *after*
        starting the container and *before* hijacking the terminal.
    
However, at this time the `start_container()` function was different and connected container to its networks after starting the container, as above commit message suggests:

    def start_container(self, container):
        container.start()
        self.connect_container_to_networks(container)
        return container

In the current code the order is inverted:

    def start_container(self, container):
        self.connect_container_to_networks(container)
        try:
            container.start()
        except APIError as ex:
            raise OperationFailedError("Cannot start service %s: %s" % (self.name, ex.explanation))
        return container

### Proposed solution

Perhaps the sequence should be the following (see pull request):

1. connect container to networks
2. start pty
3. start container

I have tested the proposed fix locally and it resolves the original problem and doesn't break any of the automated tests. However, it can potentially reintroduce the aforementioned "interactive run with networking" bug - could you explain how to reproduce it?

Considering this issue occurs semi-randomly and only in specific circumstances, I am not sure how to correctly implement a test for it. However, seeing the extensive testing suite based on Docker, I'd be happy to dig deeper if you think it's useful and give me some hints. 

### Environment

docker-compose version

    docker-compose version 1.12.0, build b31ff33
    docker-py version: 2.2.1
    CPython version: 3.6.1
    OpenSSL version: OpenSSL 1.0.1t  3 May 2016

docker version

    Client:
     Version:      17.03.1-ce
     API version:  1.27
     Go version:   go1.7.5
     Git commit:   c6d412e32
     Built:        Fri Mar 24 00:39:57 2017
     OS/Arch:      linux/amd64

    Server:
     Version:      17.03.1-ce
     API version:  1.27 (minimum version 1.12)
     Go version:   go1.7.5
     Git commit:   c6d412e32
     Built:        Fri Mar 24 00:39:57 2017
     OS/Arch:      linux/amd64
     Experimental: false

    The output of docker info.
    docker-compose version

docker info

    Containers: 8
     Running: 0
     Paused: 0
     Stopped: 8
    Images: 774
    Server Version: 17.03.1-ce
    Storage Driver: aufs
     Backing Filesystem: extfs
     Dirs: 1298
     Dirperm1 Supported: true
    Logging Driver: json-file
    Cgroup Driver: cgroupfs
    Plugins: 
     Volume: local
     Network: bridge host macvlan null overlay
    Swarm: inactive
    Runtimes: runc
    Default Runtime: runc
    Init Binary: docker-init
    containerd version: 4ab9917febca54791c5f071a9d1f404867857fcc
    runc version: 54296cf40ad8143b62dbcaa1d90e520a2136ddfe
    init version: 949e6fa
    Security Options:
     seccomp
      Profile: default
    Kernel Version: 4.9.0-2-amd64
    Operating System: Debian GNU/Linux 9 (stretch)
    OSType: linux
    Architecture: x86_64
    CPUs: 4
    Total Memory: 7.508 GiB
    Name: stinky
    ID: 2WWX:OYMN:5QA4:L3VF:TOW2:2MN6:WCU5:FXY6:P7H4:NPW3:GRWS:5ITZ
    Debug Mode (client): false
    Debug Mode (server): false
    Username: loomchild
    Experimental: false
    Insecure Registries:
     127.0.0.0/8
    Live Restore Enabled: false

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -1160,8 +1160,9 @@ def run_one_off_container(container_options, project, service, options):
                 )
                 pty = PseudoTerminal(project.client, operation)
                 sockets = pty.sockets()
-                service.start_container(container)
+                service.connect_container_to_networks(container)
                 pty.start(sockets)
+                service.start_container_without_connecting_to_networks(container)
                 exit_code = container.wait()
         except signals.ShutdownException:
             project.client.stop(container.id)
--- a/compose/service.py
+++ b/compose/service.py
@@ -492,11 +492,7 @@ class Service(object):
 
     def start_container(self, container):
         self.connect_container_to_networks(container)
-        try:
-            container.start()
-        except APIError as ex:
-            raise OperationFailedError("Cannot start service %s: %s" % (self.name, ex.explanation))
-        return container
+        return self.start_container_without_connecting_to_networks(container)
 
     def connect_container_to_networks(self, container):
         connected_networks = container.get('NetworkSettings.Networks')
@@ -518,6 +514,13 @@ class Service(object):
                 links=self._get_links(False),
                 link_local_ips=netdefs.get('link_local_ips', None),
             )
+
+    def start_container_without_connecting_to_networks(self, container):
+        try:
+            container.start()
+        except APIError as ex:
+            raise OperationFailedError("Cannot start service %s: %s" % (self.name, ex.explanation))
+        return container
 
     def remove_duplicate_containers(self, timeout=None):
         for c in self.duplicate_containers():


compose/cli/main.py
```diff
@@ -1160,8 +1160,9 @@ def remove_container(force=False):
                 )
                 pty = PseudoTerminal(project.client, operation)
                 sockets = pty.sockets()
```
loomchild: I could simplify the code as follows:

    pty.start(sockets)
    service.start_container(container)

but I was worried to reintroduce a bug during interactive run with networking, as described in f3e55568d1a7c111398876bec84da8ed8b42da7f

shin-:
Thank you for the detailed writeup and proposed fix.

For context, Aanand's commit you're quoting is part of PR #2723, which has some tests attached. I want to assume that if those tests aren't breaking, that means the fix is preserved, but obviously it's good to be prudent.

I'll take a closer look and do a real review when I have some time, but at first glance this looks reasonable.

loomchild:
Interesting. I tried adding a couple of networks to my `docker-compose.yml` and when I run the following using patched docker-compose:

    docker-compose run main 

    ip link show

then all of them are correctly displayed. 

However, if I try to apply the simple fix mentioned in my commit comment (invert `pty.start(sockets)` and `service.start_container(container)` instructions) instead of the full fix, then only one network is connected. I believe it means that I am able to reproduce the original interactive run with networking issue fixed in f3e55568d1a7c111398876bec84da8ed8b42da7f. What's weird is that when I run the automatic test suite, then all tests are green, including the ones added in a14906fd35f5c12b33631456bb7bc3dfd0a0be36.

loomchild:
Hi, did you have some time to look into this? Would you be interested in merging this pull request, perform some more tests or drop it as obsolete? 

Please let me know if I can help somehow.

ulyssessouza:
Hello @loomchild. Thanks for the PR, I could not reproduce the issue anymore.

I'm closing this because the issue looks obsolete.
Please, reopen if that's still happening.

loomchild:
@ulyssessouza I also cannot reproduce it anymore, must've been fixed in the meantime. Thanks.

Issue #2496, opened 2015-12-03
Proposal: Replace scale command with a scale option in the compose file


dnephin:
From https://github.com/docker/compose/issues/1661#issuecomment-161071522

Having both a `scale` command and `scale` configuration option leads to a scenario with unexpected results.  Take this example:

Given a `docker-compose.yml` that defines a service `web` with `scale: 5`
Run `docker-compose up -d` to start 5 instances of the `web` service
Then run `docker-compose scale web=4` to down-scale  the web service.
Finally, run another `docker-compose up -d`.

How many instances do you expect of `web`?  Some might say `4` because the service was scaled independently.  Some might say `5` because the compose file said 5. 

In the case where there are `4` instances, the value in the compose file is ignored, which is undesirable. `up -d` should always bring services to the state described in the compose file.

In the case where `up -d` scales to 5, there is no reason to have a scale command at all. `up -d` is already handling scale.  With the new file overrides a user could even set the scale numbers in a `scale.yml`, and run `docker-compose -f docker-compose.yml -f scale.yml`, which supports keeping scale separate from application configuration.

This proposal is to deprecate the `scale` command (add a removal warning for 1 or 2 releases, before finally removing it), and instead use `up -d` to scale a service.

**Update**

To handle the case of overriding the scale number from the command line, `up` can use the `service=count` syntax that is currently available to scale (https://github.com/docker/compose/issues/1661#issuecomment-204865035).

nazar-pc:
:+1: Completely agree with proposal.

schickling:
Sounds good! :+1:

migueleliasweb:
:+1:

megastef:
+1

ahfeel:
+1

RobMaskell:
How abut having a minimum number of instances to scale in the docker-compose.yml and leave the scale command as it is, an override? That would seem to handle all the use cases people have proposed and what to do when the two different ways of scaling conflict.

nazar-pc:
Why might you need scale command if you can just do `docker-compose up -d`?

RobMaskell:
It leaves the value in the docker-compose.yml as declarative there should be a minimum of 3, and if there should be a minimum of three the docker-compose scale 2 would be bad and stopping someone from doing that might be a good thing

nazar-pc:
This proposal is about removing `scale` command and introduce option instead, so you'll not be able to do `docker-compose scale whatever=2`. I do not understand your comments in this context.

[...]

PR #4742, opened 2017-04-18
docker-compose up --scale, scale in config file

shin-:
Implements #2496 

This is a slight variant of Daniel's proposal in #2496. It adds a `--scale` flag instead of overloading the functionality of the services list, which works around the issue described in https://github.com/docker/compose/issues/1661#issuecomment-205354944

The `--scale` option will be available for all versions of the Compose file, but the `scale` config option will only be added in file format 2.2. As a result, `docker-compose scale` with a 2.2 config file will be prohibited.

`deploy.replicas` might be considered as synonym to `scale` in the v3 format, but would be an independent addition (i.e. not in this PR).


compose/cli/main.py
```diff
@@ -875,7 +874,7 @@ def up(self, options):
         If you want to force Compose to stop and recreate all containers, use the
         `--force-recreate` flag.
 
-        Usage: up [options] [SERVICE...]
+        Usage: up [options] [--scale SERVICE=NUM...] [SERVICE...]
```
shin-: I think we have to declare it separately because it's repeatable. Same goes for build for example:
`        Usage: build [options] [--build-arg key=val...] [SERVICE...]`

or run:
`        Usage: run [options] [-v VOLUME...] [-p PORT...] [-e KEY=VAL...] SERVICE [COMMAND] [ARGS...]`

dnephin: Ahh, cool, I didn't know that.

PR #4762, opened 2017-04-25
[Fix #4493] Support for Required Environment Variables

### LINKED ISSUE #4493: Support for Required Environment Variables
I'm creating a new request for specifying required environment variables in docker-compose.yml. This applies to both env vars used for [variable substitution](https://docs.docker.com/compose/compose-file/#variable-substitution), and also for those passed to containers in the `environment` block. This has been asked for in the past ( see #2441 and #1377), but issues have been closed without addressing this. @dnephin 's [implementation](https://github.com/docker/compose/pull/3108/commits/fe08be698d36d42a66839ce284989947220931cd) of default environment variables via `:-` suffix is the closest thing to what a required variable implementation would look like.

### Variable Substitution
**Current Behavior**
Environment variables used for variable substitution (ie: `${ENV_VAR}`) in docker-compose.yml are currently replaced with a blank string at runtime if they are not set. Compose throws a warning but continues running the container.

**Proposal**
Allow users to specify environment variables that _must be set and can not be empty_ by appending the variable with suffix **`?`**. This maintains backwards compatibility with current ENV VAR naming implementation (current ENV VARS keys can not contain '?'). The container would not run unless this variable was set via `.env` or exists and is not empty on the host.

> Another option for the suffix is **`:?`** which follows [POSIX](http://www.ibm.com/support/knowledgecenter/ssw_aix_71/com.ibm.aix.osdevice/korn_shell_parmsub1.htm)/[bash](https://www.gnu.org/software/bash/manual/bash.html#Shell-Parameter-Expansion) convention, but this gets a bit messy later on in the `environment` block due to **`:`** being a key/value separator. I also think [**`!`**](https://www.youtube.com/watch?v=2P5qbcRAXVk) is a fine option though it doesn't follow any familiar convention that I know of).

Example:
<pre>
volumes:
    - ${SOURCE_VOLUME<b><i>?</i></b>}:/app/code
</pre>

**Note:** Specifying a default/fallback ENV VAR value via `$VAR:-default` should not be allowed in combination with the "required" suffix. The "required" suffix throws an error on empty vars, while the "default" suffix catches empty vars and replaces them with the default. The two cannot not be combined.

### Environment Block

**Current behavior**
When a key with no value is placed in the `environment:` block (or `env_file`), compose inherits that variable from the host. If the variable does not exist on the host, the variable is not set and is not passed to the container, and no warnings are thrown. The container runs without the environment variable.

**Proposal**
Allow users to specify required environment variables in the `environment` block (or `env_file`), throwing a runtime error at container startup if any required vars are empty or do not exist.

Example:
```
environment:
    DEPLOYMENT?:
```

### Work Needed

I'm basing this off of @dnephin 's [implementation](https://github.com/docker/compose/pull/3108/commits/fe08be698d36d42a66839ce284989947220931cd) of default environment variables. I'm not sure yet how to implement similar in the `environment` block.

- `interpolation.py` will need to parse the suffix `?` as a "required and not empty" flag. Regex will look something like [this regex for the default env var suffix](https://github.com/docker/compose/pull/3108/commits/fe08be698d36d42a66839ce284989947220931cd#diff-b2372f2a81cc5911e04cc7df7d68675bR75).
- Throw error at runtime in [environment.py try/catch](https://github.com/docker/compose/blob/a3a9d8944a413897ae1f0305d16d6d1071487ad6/compose/config/environment.py#L80) if var is required and empty with error message "ERROR: The {} variable is required and not set." Do not continue to run container.
- Add interpolation testing for required vars
- Update documentation to reflect required variables spec

eliat123:
This is my first PR. 

I have added support for required environment variables - issue #4493. 
Required environment variables are regular environment variables ending with '?'.

Whenever a required environment variable is written in docker-compose configuration: for example:

   ```
 volumes:
     - ${MISSING?}:/app/code
 environment:
     - MISSING?

```
we get an error message such as:
```
ERROR: Invalid interpolation format for "volumes" option in service "web": "${MISSING?}:/app/code" caused by required environment variable "MISSING" not set
```
and 
```
ERROR: environment variable "MISSING?" is required
```
Please note:
* All tests are passing and added new tests for this feature.
* Did not find a version to place this in the changelog.md
* Documentation is on a different repo, I will add it once I get approval for this PR.
* Tried to follow @rsynnest suggestions to implement this feature.

### DIFF
--- a/compose/network.py
+++ b/compose/network.py
@@ -158,8 +158,8 @@ def check_remote_ipam_config(remote, local):
             if sorted(lc.get('AuxiliaryAddresses')) != sorted(rc.get('AuxiliaryAddresses')):
                 raise NetworkConfigChangedError(local.full_name, 'IPAM config aux_addresses')
 
-    remote_opts = remote_ipam.get('Options', {})
-    local_opts = local.ipam.get('options', {})
+    remote_opts = remote_ipam.get('Options') or {}
+    local_opts = local.ipam.get('options') or {}
     for k in set.union(set(remote_opts.keys()), set(local_opts.keys())):
         if remote_opts.get(k) != local_opts.get(k):
             raise NetworkConfigChangedError(local.full_name, 'IPAM option "{}"'.format(k))
--- a/tests/unit/network_test.py
+++ b/tests/unit/network_test.py
@@ -100,6 +100,44 @@ class NetworkTest(unittest.TestCase):
             {'Driver': 'overlay', 'Options': None}, net
         )
 
+    def test_check_remote_network_config_null_remote_ipam_options(self):
+        ipam_config = {
+            'driver': 'default',
+            'config': [
+                {'subnet': '172.0.0.1/16', },
+                {
+                    'subnet': '156.0.0.1/25',
+                    'gateway': '156.0.0.1',
+                    'aux_addresses': ['11.0.0.1', '24.25.26.27'],
+                    'ip_range': '156.0.0.1-254'
+                }
+            ]
+        }
+        net = Network(
+            None, 'compose_test', 'net1', 'bridge', ipam=ipam_config,
+        )
+
+        check_remote_network_config(
+            {
+                'Driver': 'bridge',
+                'Attachable': True,
+                'IPAM': {
+                    'Driver': 'default',
+                    'Config': [{
+                        'Subnet': '156.0.0.1/25',
+                        'Gateway': '156.0.0.1',
+                        'AuxiliaryAddresses': ['24.25.26.27', '11.0.0.1'],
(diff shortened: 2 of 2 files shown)


eliat123:
@GordonTheTurtle I have signed the commit. the jenkins build seems to be stuck it has been running for more than a day. can you check it?

eliat123:
I have recommitted my work as previous jenkins build was stuck, andrewhsu@docker.com has asked me to re-trigger a commit to initialize the build process.

eliat123:
@shin- I have accidentally removed my changes from my fork and this caused this issue to close, can you reopen the issue and I will resubmit the fix.

Issue #4771, opened 2017-04-27
.yaml extention does not work with override file.


gorzell:
The [current documentation](https://docs.docker.com/compose/compose-file/#service-configuration-reference) states that you can use either the `.yml` or `.yaml` extension for your compose file, however this does not appear to hold true for the `docker-compose.override` file.  In that case only `.yml` seems to work.  It would be really nice if the behavior here was consistent.

eliat123:
#dibs

shin-:
Fixed-by: #4859

PR #4777, opened 2017-04-28
4771 Fix for yaml extention does not work with override file

eliat123:
Signed-off-by: Eli Atzaba <eliat123@gmail.com>

Fix for #4771


eliat123:
@shin- Added raising an error if both yaml & yml files are present. ( with relevant tests )

shin-:
Rebased + merged: #4859 

Thank you!

PR #4800, opened 2017-05-06
use docker instead of docker-py when print version

luzfcb:
"docker-py" python package was renamed to "docker" on commit https://github.com/docker/docker-py/commit/04e943798691ba2663b91eaedd8ae2410f016d2c .

Use docker when printing the version message, to avoid confusion between the new name and the old name of the package.

### DIFF
--- a/compose/cli/utils.py
+++ b/compose/cli/utils.py
@@ -89,7 +89,7 @@ def get_version_info(scope):
     if scope == 'full':
         return (
             "{}\n"
-            "docker-py version: {}\n"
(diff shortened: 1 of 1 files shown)


ijc:
PR grooming: The source package for the Python module is still [`docker-py`](https://github.com/docker/docker-py) and just saying `docker` is ambiguous with the Docker engine, which is more confusing than the potential confusion with the Python library name IMHO. Thus, closing, sorry it took a while for someone to look at this.

PR #4808, opened 2017-05-08
Updated CLI help for docker-compose pull command

londoncalling:
### What's changed

- updated Help for the `pull` command per user feedback

### Related
- [Issue 3190 in `docker.github.io` docs repo](https://github.com/docker/docker.github.io/issues/3190)

- [Comment in PR 3197 expanded on pull command, added links to Compose samples](https://github.com/docker/docker.github.io/pull/3197#discussion_r115339449)

### Reviewers

@shin- @dnephin @joaofnfernandes 

Signed-off-by: Victoria Bialas <victoria.bialas@docker.com>

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -171,12 +171,12 @@ class TopLevelCommand(object):
                                   in the client certificate (for example if your docker host
                                   is an IP address)
       --project-directory PATH    Specify an alternate working directory
-                                  (default: the path of the compose file)
+                                  (default: the path of the Compose file)
 
     Commands:
       build              Build or rebuild services
       bundle             Generate a Docker bundle from the Compose file
-      config             Validate and view the compose file
+      config             Validate and view the Compose file
       create             Create services
       down               Stop and remove containers, networks, images, and volumes
       events             Receive real time events from containers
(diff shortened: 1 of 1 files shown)


joaofnfernandes:
LGTM

shin-:
Where does the mention of `docker-stack.yml` come from though? Is that Docker Cloud bleeding through? It's not mentioned anywhere else in the Compose project AFAIK.

londoncalling:
@shin- this is the `docker-stack.yml` that operates essentially as a Compose file and uses the `deploy` key, not related to the Cloud stack file. (That naming confusion needs to be addressed.) Should I not mention it at all here, I can take it out.

londoncalling:
@shin- you are right. It doesn't work with `docker-stack.yml`. I'll change the Help description.

shin-:
@londoncalling The term we use in other parts of the CLI is "Compose file", like in `docker-compose down`:

```
...
By default, the only things removed are:

- Containers for services defined in the Compose file
- Networks defined in the `networks` section of the Compose file
...
```

Or `docker-compose config`:

```
Validate and view the compose file.
...
```

It might be better to use the same term for consistency (also because `docker-compose.yml` is the default, but not the only possible name for a Compose file)

londoncalling:
@shin- this should be good now, and it past the CI build! :-) Can you merge it for me once you think it's okay?

PR #4824, opened 2017-05-15
fix python3.x _asdict() return None

mengskysama:
Signed-off-by: mengskysama <mengskysama@gmail.com>


shin-:
Ah, I missed that one. Thank you!

Related: #4798

Issue #4728, opened 2017-04-12
`run --volumes` replaces volume configuration rather than adding


mtibben:
Using [`run --volumes`](https://github.com/docker/compose/pull/4566) replaces all volumes in the docker-compose config.

It would be more useful to me to be able to specify the volumes to mount in addition to the ones in config. Perhaps `--append-volumes` or `--add-volumes`?

shin-:
Hi Michael,

By design, the volumes declared in config are never used for `docker-compose run`. The addition of `--volumes` doesn't change anything in that regard.

mtibben:
That doesn't seem correct @shin-, `run` does mount volumes:
e.g.
```
$ cat << EOF > docker-compose.yml
version: '2'
services:
  test:
    image: alpine
    volumes:
      - /test
EOF

$ docker-compose run test mount | grep test
/dev/sda1 on /test type ext4 (rw,relatime,data=ordered)
```

andyneff:
I have the same problem.

I agree with @mtibben, `docker run` does mount the volumes. Furthermore, I think this is good behavior, and definitely the desired behavior for me and my team. If it did not do this, it would severely limit the usability of docker-compose for us.

Since `--volume` was recently added in 1.12, I would vote for this being a bug.

A simple fix for this would be modifying _get_container_create_options and add

```
        override_options['volumes'] = (container_options.get('volumes', []) +
                                       override_options.get('volumes', []))
```

I like this fix because it lets me add to the volumes, not replace. Furthermore, because of the order, if I is `-v/--volume' to bind to a directory that is already specified in the docker-compose, the `-v/--volume` wins, and the original is not mounted.

andyneff:
@mtibben 
As hold over, here is a bash hack you can use to get all the mounts from `docker-compose config`

```bash
function docker-compose-volumes()
{
  local OLD_IFS="${IFS}"
  IFS=$'\n'
  DOCKER_VOLUME_FLAGS=(\
        $(sed -nE '/^services:$/b s; d;
                   :s; n; /^  '"${1}"':$/b f; /^ /b s; d
                   :f; n; /^    volumes:/b v; /^    /b f; d
                   :v; n; /^    -/{H; b v}; x; s|\n    - |\n-v\n|g; p'))
  IFS="${OLD_IFS}"
}

### Test it out ###

cat << EOF > docker-compose.yml
version: '2'
services:
  mytest:
    image: alpine
    volumes:
[...]

PR #4830, opened 2017-05-16
Change --volume behavior to add instead of replace mounts

andyneff:
Potential solution to #4728.

When using the `docker-compose run` command, rather than replacing all mount when using the `-v/--volume`, this will add to the list of mounts (potentially overriding mounts already in the `docker-compose.yml` file).

- [ ] Need to add unit tests

### DIFF
--- a/compose/service.py
+++ b/compose/service.py
@@ -716,6 +716,8 @@ class Service(object):
         container_options = dict(
             (k, self.options[k])
             for k in DOCKER_CONFIG_KEYS if k in self.options)
+        override_options['volumes'] = (container_options.get('volumes', []) +
+                                       override_options.get('volumes', []))
         container_options.update(override_options)
 
         if not container_options.get('name'):


mtibben:
Hi @shin-, will 1.14 include this fix?

shin-:
Unfortunately, i don't think it will be. It was still a WIP when I made the RC cut (I believe it still is? @andyneff do you need anything from me to move this forward?)

mtibben:
This is a simple fix. Add a test and :shipit:? Anything I can do to help?

shin-:
Yeah, I'll take care of adding the test in the near future and then we can merge it :+1:

M3kH:
Well actually this reverts my expected behaviour since now I've always to specify the volumes on the `docker-compose run` since the one in the `docker-compose.yml` would never be overwritten.

Would be maybe nice have an options for it has suggested somewhere else and deliver a distinction between `add` and `replace`.

Issue #4582, opened 2017-03-07
docker-compose does not support cpus parameter


AlexeyRokhin:
docker-compose does not support cpus parameter:
`
ERROR: The Compose file '.\.\docker-compose.test.yml' is invalid because:
Unsupported config option for services.test: 'cpus'
`
Please add support for --cpus parameter.

shin-:
Are you referring to this: https://docs.docker.com/compose/compose-file/#resources ?

If so, it needs to be a subkey of `services.test.deploy.limits` or `services.test.deploy.reservations`

AlexeyRokhin:
I meat docker-compose itself.
Version 2.x (which I currently use) of docker-compose .yml does not support cpus parameter (--cpus for docker run).
Version 3.x supports it but documentation for states that docker-compose ignores section **deploy**: https://docs.docker.com/compose/compose-file/#deploy

georgyturevich:
@shin- Hi Joffrey,

It is very important issue for Windows Server 2016 containers as --cpus/--cpu-count/--cpu-percent parameters is the only way to limit cpu consumption. So, just for now, docker-compos is not production ready for Windows containers as we can configure these parameters.

shin-:
Ah, I see. Those were added recently (Engine 1.13). We're most likely going to need a Compose file format 2.2 to support those.

georgyturevich:
@shin- Hi Joffrey,

I see that you referenced this issue to the pull request https://github.com/docker/compose/pull/4262 but it seems that they are not related between each other.

Btw, do we have any ETA for implementing this current feature?

AlexeyRokhin:
@shin- Hi
As I understand, docker-compose depends on docker-py. And it is required to implement cpus, cpu-count and cpu-percent in docker-py before docker-compose. Is it correct?

shin-:
@georgyturevich Sorry, indeed - #4262 is fixed by #4588.
ETA for this feature is Compose 1.13

@AlexeyRokhin Yes! I see you submitted https://github.com/docker/docker-py/pull/1532 - I'll take a closer look at it ASAP.

AlexeyRokhin:
@shin- Hi Joffrey
Pull request for compose is created:
https://github.com/docker/compose/pull/4835

uLan08:
Hello, I encountered this but I shouldn't because my docker-compose version is `1.13` and docker-compose file version is `2.2`. Any ideas why?

```sh
$ docker-compose --version
[...]

PR #4835, opened 2017-05-17
Add cpu_count, cpu_percent, cpus parameters.

AlexeyRokhin:
Fixes https://github.com/docker/compose/issues/4582
Uses docker-py 2.3.0 to support these parameters.


compose/config/config_schema_v2.2.json
```diff
@@ -74,8 +74,11 @@
           ]
         },
         "container_name": {"type": "string"},
+        "cpu_count": {"type": "integer", "minimum": 0},
+        "cpu_percent": {"type": "integer", "minimum": 0, "maximum": 100},
         "cpu_shares": {"type": ["number", "string"]},
         "cpu_quota": {"type": ["number", "string"]},
+        "cpus": {"type": ["number", "string"], "minimum": 0},
```
shin-: I don't think "string" is a valid type here.

compose/service.py
```diff
@@ -798,6 +801,10 @@ def _get_container_host_config(self, override_options, one_off=False):
             init_path = options.get('init')
             options['init'] = True
 
+        nano_cpus = None
+        if 'cpus' in options:
+            nano_cpus = int(options.get('cpus') * 1000000000)
```
shin-: We should check in `compose/config/validation.py` that the result of `cpus * 10^9` is not a decimal value, [same as here](https://github.com/moby/moby/blob/v17.05.0-ce/opts/opts.go#L378-L381)

compose/service.py
```diff
@@ -798,6 +801,13 @@ def _get_container_host_config(self, override_options, one_off=False):
             init_path = options.get('init')
             options['init'] = True
 
+        nano_cpus = None
+        if 'cpus' in options:
+            nano_cpus = options.get('cpus') * 1000000000
+            if isinstance(nano_cpus, float) and not nano_cpus.is_integer():
```
shin-: I'd really like us to do that check in `validation.py` so it can be caught when running `docker-compose config`

AlexeyRokhin:
Thank you for your time!

PR #4864, opened 2017-05-23
Add partial support for v3.3 credential_spec

shin-:
cc @dnephin 

Related: https://github.com/docker/cli/pull/71

### DIFF
--- a/compose/config/config.py
+++ b/compose/config/config.py
@@ -108,6 +108,7 @@ DOCKER_CONFIG_KEYS = [
 ALLOWED_KEYS = DOCKER_CONFIG_KEYS + [
     'build',
     'container_name',
+    'credential_spec',
     'dockerfile',
     'log_driver',
     'log_opt',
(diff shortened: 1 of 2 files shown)


compose/config/config.py
```diff
@@ -320,6 +321,32 @@ def find_candidates_in_parent_dirs(filenames, path):
...
+            )
+        )
+
+    if services_using_credential_spec:
+        log.warn(
+            warning_template.format(
+                services=", ".join(sorted(s['name'] for s in services_using_credential_spec)),
+                key='credential_spec'
+            )
+        )
```
dnephin: Maybe split this into a new function?

```python

def check_swarm_only_key(service_dicts, key):
    ...
```

Then `check_swarm_only_config` could be:

```python
def check_swarm_only_config(service_dicts):
    check_swarm_only_key(service_dicts, 'deploy')
    check_swarm_only_key(service_dicts, 'credential_spec')
```

shin-: Thanks - good call. Updated accordingly :+1:

Issue #4870, opened 2017-05-27
missing slash at docker-compose rm 


kasyapcm:

kasyapcm:
missing slash at docker-compose rm

PR #4871, opened 2017-05-27
missing slash at docker-compose rm #4870

kashcm:


ijc:
PR grooming: Thank you for your submission, and sorry it took a while for anyone to have a look but apart from the lack of correct DCO the way this is presented is common enough that I don't think we really needto change anything, so I'm going to close.

PR #4920, opened 2017-06-13
'pip' isn't always the python we're running under

jcshort:
On distributions that allow parallel python2 and python3 installation, the `pip` binary may not be the python we're running under.  Specifically, we ran into a case where we needed docker-py in python2 to support a specific version of Ansible, and decided to install docker-compose via python3 pip to avoid a collision.  Much to our chagrin, python3 docker-compose continued to complain about docker-py, because of its naive usage of `pip`.  And yes, I hear you when you say virtualenv.


shin-:
Unfortunately that would break our binary distribution (where `sys.executable == 'docker-compose'`).

I'm afraid you'll have to ignore the warning

PR #4928, opened 2017-06-15
Optimize "extends" without file specification

protoss-player:
Loading the same config file adds about 100ms per each extension service, which results in painfully slow CLI calls when a config consists of a couple of dozens of services.

This patch makes Compose re-use config files.


shin-:
LGTM - thank you!

PR #4933, opened 2017-06-16
Fix TopLevelCommand constructor to make use of all args

### LINKED ISSUE #4855: Bug at the initialization of the TopLevelCommand class
Found a bug [here](https://github.com/docker/compose/blob/master/compose/cli/main.py#L205-L207): 

the second parameter `project_dir` sent to that class  is not used.

JoseTomasTocino:
Fixes #4855


shin-:
#4709

PR #4944, opened 2017-06-20
Add update-docker-composer bash update script

uberhacker:
Keeping docker-compose updated is a challenge since this binary is not included in package updates.  This commit provides a bash shell script to update docker-compose to the latest version.

### DIFF
--- a/bin/update-docker-compose
+++ b/bin/update-docker-compose
@@ -14,7 +14,7 @@ if [ ! -z $COMPOSE ]; then
   cd compose
   git fetch --all
   LATEST_RELEASE=$(git tag -l | grep ^1 | sort -Vr | head -1)
-  RC=$(git tag -l | grep ^1 | sort -Vr | head -1 | grep -e '[[:print:]]')
+  RC=$(echo $LATEST_RELEASE | grep -e '[a-zA-Z]')
   if [ ! -z $RC ]; then
     STABLE_RELEASE=${LATEST_RELEASE/[\-a-zA-Z]*[0-9]*/}
     RELEASES=$(git tag -l | grep ^1 | sort -Vr | xargs)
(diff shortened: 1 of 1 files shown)


uberhacker:
@GordonTheTurtle @shin- Is there any chance this might be accepted or should I just close this PR?  As you probably already know, updating Compose is currently a total PITA in Linux.

hartwork:
Why does `pip install --upgrade docker-compose` not work for you?
My vote on closing this pull request.

uberhacker:
@hartwork: Because I didn't install via pip?  Also, are you sure the pip repository is updated as frequently as this project?  I guess this could work but it would be nice if it was documented as a recommended way to install.  There is still the need to have the ability to self-update if not installing via pip.

bin/update-docker-compose
```diff
@@ -0,0 +1,37 @@
...
+    rm -rf compose/
+  fi
+  git clone https://github.com/docker/compose.git
+  cd compose
+  git fetch --all
+  LATEST_RELEASE=$(git tag -l | grep ^1 | sort -Vr | head -1)
+  RC=$(echo $LATEST_RELEASE | grep -e '[a-zA-Z]')
+  if [ ! -z $RC ]; then
+    STABLE_RELEASE=${LATEST_RELEASE/[\-a-zA-Z]*[0-9]*/}
+    RELEASES=$(git tag -l | grep ^1 | sort -Vr | xargs)
```
hartwork: Have you checked `git describe`? Looks like this is trying to re-create it.

hartwork:
@uberhacker `pip install --upgrade <package>` is standard for pip, I don't see why docker-compose would need to document it in particular. I suspect most people use docker-compose through their distro or through pip. Even the from-git versions work with pip like this if you want as pip can install from tarballs containing `setup.py` files:
```console
# cd "$(mktemp -d)"                                  
# virtualenv foo
# source foo/bin/activate
# pip install https://github.com/docker/compose/archive/master.tar.gz
# docker-compose --version
```
So you could run `pip install --upgrade https://github.com/docker/compose/archive/master.tar.gz` every day and stay up to date, I think. Do you still need the script then?

PR #4951, opened 2017-06-23
Add a flag --no-ansi to remove control characters on parallel executions

cecton:
As discussed in #4780 with @shin- , I made this PR which adds a flag to disable the control characters of the parallel executions. It's handy when logging the output of Docker Compose to a file for instance.

I tried to respect the existing code styling but don't hesitate to make any remark.

I haven't make any test, I'm not sure what kind of test you would want.

Example:

```
[0] [14:18:08] ~/r/app-swarm-ui dev > ~/repos/compose/bin/docker-compose --no-ansi rm -vf resource | cat -
Removing appswarmui_resource_1 ... 
Removing appswarmui_resource_1: done
Going to remove appswarmui_resource_1
```

### DIFF
--- a/compose/parallel.py
+++ b/compose/parallel.py
@@ -26,7 +26,7 @@ log = logging.getLogger(__name__)
 STOP = object()
 
 
-def parallel_execute(objects, func, get_name, msg, get_deps=None, limit=None):
+def parallel_execute(objects, func, get_name, msg, get_deps=None, limit=None, noansi=False):
     """Runs func on objects in parallel while ensuring that func is
     ran on object only after it is ran on all its dependencies.
 
@@ -36,7 +36,7 @@ def parallel_execute(objects, func, get_name, msg, get_deps=None, limit=None):
     objects = list(objects)
     stream = get_output_stream(sys.stderr)
 
-    writer = ParallelStreamWriter(stream, msg)
+    writer = ParallelStreamWriter(stream, msg, noansi)
     for obj in objects:
         writer.add_object(get_name(obj))
     writer.write_initial()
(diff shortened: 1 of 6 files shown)


cecton:
@shin- I don't want to pressure but I've been waiting a month for a feedback here 😢 Can someone have a look?

cecton:
ah but the build just failed in Python 2.7 on Windows...

cecton:
@shin-  fixed and I added a test that will also ensure the alignment is good with the --no-ansi flag

shin-:
@cecton Thank you! Tests pass which is great, I'll take a closer look at the code in the coming days but we should be good.

cecton:
Thanks a lot @shin- :)

compose/parallel.py
```diff
@@ -255,26 +258,60 @@ def write(self, obj_index, status):
         self.stream.flush()
 
 
-def parallel_operation(containers, operation, options, message):
+class ParallelStreamWriterNoANSI(object):
```
shin-: There's a lot of code duplication happening here that we can probably avoid by adding a `noansi` parameter and a handful of conditionals to the original class.

compose/project.py
```diff
@@ -60,13 +60,15 @@ class Project(object):
     """
     A collection of services.
     """
-    def __init__(self, name, services, client, networks=None, volumes=None, config_version=None):
+    def __init__(self, name, services, client, networks=None, volumes=None, config_version=None,
+                 noansi=False):
```
shin-: I wish there was a better way to pass this value, as it strictly influences CLI / output and hence doesn't really belong on the `Project` class. Unfortunately this would probably require a code overhaul that's outside of the scope of this PR.

compose/parallel.py
cecton: Very good point XD and quite obvious. How could I miss that...

cecton:
ping @shin- done (I don't know why the jenkens build doesn't work but it doesn't seem related to my changes)

shin-:
Thank you!

PR #4974, opened 2017-06-28
Bump docker Python SDK version -> 2.4.2

shin-:
Signed-off-by: Joffrey F <joffrey@docker.com>

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -37,7 +37,7 @@ install_requires = [
     'requests >= 2.6.1, != 2.11.0, < 2.12',
     'texttable >= 0.8.1, < 0.9',
     'websocket-client >= 0.32.0, < 1.0',
-    'docker >= 2.3.0, < 3.0',
+    'docker >= 2.4.2, < 3.0',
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -37,7 +37,7 @@ def find_version(*file_paths):
     'requests >= 2.6.1, != 2.11.0, < 2.12',
     'texttable >= 0.8.1, < 0.9',
     'websocket-client >= 0.32.0, < 1.0',
-    'docker >= 2.3.0, < 3.0',
+    'docker >= 2.4.2, < 3.0',
```
joaoqalves: @shin- Taking into account that this may break a lot of builds around the world, perhaps it would be good to be a little bit more restrictive here. I mean, `docker >= 2.4.2, < 2.5.0` or something around these lines. What do you think about it?

shin-: I understand the concern with the recent event with the release of 2.4.0, but I still think overall the current schema has more advantages than downsides. If other dependencies bug happen in the future, I'll be happy to re-evaluate my stance on this, but I don't believe it'll come to this.

PR #4982, opened 2017-07-01
Always silence pull output with --parallel

edsrzf:
This is how things were prior to the addition of the --quiet flag.
Making it not silent produces output that's weird and difficult to read.


shin-:
Right, I remember seeing that - thought I'd fixed it before merging.

Thanks!

shin-:
#4938

Issue #4556, opened 2017-03-02
Make --parallel the default for pull


edsrzf:
#4428 added the `--parallel` flag to the `pull` command. What needs to be done to get that to the point where it's the default? Off the top of my head:

- [ ] Some experience with `--parallel` to gather feedback.
- [ ] Better, more detailed output. I have some ideas here.

Do we need anything else? Is there any reason that parallel pulling should *never* become the default?

jmreicha:
Has there been any progress on this?  I have been using `--parallel` for awhile, and it has been working well. 
 The only thing I think it is missing is some sort of progress indicator.

edsrzf:
I haven't made any steps towards this yet, no, aside from thinking about it. I agree some sort of progress feedback is necessary. At the very least I want to know which images are currently being pulled as the command is working.

tejasmanohar:
+100. I've been wanting parallel downloads for a while and never knew it was an option (because who checks flags for commands that already work :P)

kingbuzzman:
Any news on this? A `.dockerglobalsettings` or something like that we can set it would also work...

PR #4983, opened 2017-07-01
Show what's in progress when running parallel

edsrzf:
Prior to this commit, docker-compose only showed "done" or "error" once a parallel task was finished. This made it hard to tell which task(s) were actually in progress.

Now there's an additional "working" status that's shown in yellow when a task is in-progress. The main goal was to get this feedback when using `docker-compose pull --parallel`, as that's the only parallel task with a concurrency limit.

I implemented this by adding an additional "status" field to the result tuple in parallel.py. I'm happy to discuss alternatives, bikeshed over status names, whether constants should be used, etc.

Updates #4556

### DIFF
--- a/compose/parallel.py
+++ b/compose/parallel.py
@@ -14,6 +14,7 @@ from six.moves.queue import Queue
 
 from compose.cli.colors import green
 from compose.cli.colors import red
+from compose.cli.colors import yellow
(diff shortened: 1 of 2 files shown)


edsrzf:
This was long-ignored and eventually equivalent code made it in, so I'm closing.

Issue #3573, opened 2016-06-09
Option to disable check for orphans with docker-compose


rreinurm:
`docker-compose version 1.7.1, build 0a9ab35`

I am using docker-compose option -p to group containers which belong to same environment so I have visually easy naming convention in place like _test_service_1_. For example I have **test** environment where I run 50 projects meaning there can 50+ containers. Every project provides own container via docker-compose.yml file. If I execute `docker-compose -p **test** up -d` then docker-compose is busy to compare all those containers which were created from other 50 projects.

johnharris85:
I don't think I really understand the issue here? Is the title related to the content in some way? Are you able to give any more information?

rreinurm:
My issue is that I spread deployment of containers to multiple independent docker-compose files. Project A has own docker-compose.yml and project B, C ... etc have their own corresponding docker-compose yml. Every time I have to make redeploy in one of those project, docker-compose discovers all those other non-defined projects and reports them as orphans. I have quite long list of those reported orphans and I expect my number of services will grow. 

Although at the moment it's only warning message, it causes confusion and I would like to avoid that (plus performance benefit would be nice as well, but it's not so significant atm). My current options are either to concatenate existing docker-compose.yml files as one or stop using `--project` to group containers together by common naming convention.

rreinurm:
Maybe better approach is to define additional filter via labels in docker-compose.yml file or parameters so it would be still possible to find orphans, but they don't report orphans from other projects. In my case example would be `docker-compose --project-name test-environment --label projectA up -d` (although I would prefer to switch project-name and label values). In the end I would like to have naming convention place for easier filtering from web-ui like _test-environment_projectA_1_ What do you think about the approach in general?

lukem512:
:+1:

I too am facing an issue where the orphaned containers is expected, desired behaviour and I would like to suppress this warning. Perhaps a `--suppress-orphan-warning` switch could be added?

Carighan:
Same problem here. We have 12 docker-compose files, each using the same docker file to build different versions (which we want to control separately as they're not updated in sync) of the same system for different environments.

Each up-command, we get a slew of orphan-warnings for actually wanted containers.

johnharris85:
So this would just be a flag to up? Like `docker-compose -p test up -d --suppress-orphan-warning` ?

rreinurm:
[...]

PR #4992, opened 2017-07-04
Add --ignore-orphan-warning

johnharris85:
Addresses #3573.

This change means that you can specify `--remove-orphans` and `--ignore-orphan-warning` and it will just remove them. There is a different (better?) implementation that makes those two flags incompatible and specifying the `--ignore-orphan-warning` flag just skips even looking for orphans. If this flag is deemed useful enough to merge, then I may change the implementation in that direction to remove useless work.

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -377,9 +377,20 @@ class TopLevelCommand(object):
             -t, --timeout TIMEOUT   Specify a shutdown timeout in seconds.
                                     (default: 10)
         """
+        environment = Environment.from_env_file(self.project_dir)
+        ignore_orphans = environment.get_boolean('COMPOSE_IGNORE_ORPHANS')
+
+        if ignore_orphans and options['--remove-orphans']:
+            raise UserError("COMPOSE_IGNORE_ORPHANS and --remove-orphans cannot be combined.")
+
         image_type = image_type_from_opt('--rmi', options['--rmi'])
         timeout = timeout_from_opts(options)
-        self.project.down(image_type, options['--volumes'], options['--remove-orphans'], timeout=timeout)
+        self.project.down(
+            image_type,
+            options['--volumes'],
+            options['--remove-orphans'],
+            timeout=timeout,
+            ignore_orphans=ignore_orphans)
 
     def events(self, options):
         """
@@ -941,6 +952,12 @@ class TopLevelCommand(object):
         if detached and timeout:
             raise UserError("-d and --timeout cannot be combined.")
 
+        environment = Environment.from_env_file(self.project_dir)
+        ignore_orphans = environment.get_boolean('COMPOSE_IGNORE_ORPHANS')
+
+        if ignore_orphans and remove_orphans:
+            raise UserError("COMPOSE_IGNORE_ORPHANS and --remove-orphans cannot be combined.")
+
         if no_start:
             for excluded in ['-d', '--abort-on-container-exit', '--exit-code-from']:
                 if options.get(excluded):
@@ -955,6 +972,7 @@ class TopLevelCommand(object):
                 timeout=timeout,
                 detached=detached,
                 remove_orphans=remove_orphans,
+                ignore_orphans=ignore_orphans,
                 scale_override=parse_scale_args(options['--scale']),
                 start=not no_start
             )
--- a/tests/integration/project_test.py
+++ b/tests/integration/project_test.py
@@ -1633,6 +1633,31 @@ class ProjectTest(DockerClientTestCase):
             if ctnr.labels.get(LABEL_SERVICE) == 'service1'
         ]) == 0
 
+    def test_project_up_ignore_orphans(self):
+        config_dict = {
+            'service1': {
+                'image': 'busybox:latest',
+                'command': 'top',
+            }
+        }
+
+        config_data = load_config(config_dict)
+        project = Project.from_config(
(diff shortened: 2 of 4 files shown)


shin-:
Hmmm, I'm not sure a flag is the best option here. It's never going to be a really common option, even if it's helpful to some. Maybe this could be controlled using an environment variable (`COMPOSE_IGNORE_ORPHAN_WARNINGS`), similar to `COMPOSE_API_VERSION` or `COMPOSE_PATH_SEPARATOR`.

WDYT?

johnharris85:
Thanks @shin. Can see the point-of-view, however most of the folks requesting (on the issue) are encountering this when running automated CI, not really going to be a manual thing. Given the likely use case / application, better to have a readable flag on the command than have to set / unset / keep track of ENV vars in a CI script / pipeline? 

No religious feelings either way really.

shin-:
Yeah, even in light of that, I like the env var better. I'd rather not clutter the list of options for up/down with something that is going to be useful for <1% of the user base.

Let me know if you have time to make the changes, I'll be happy to help if needed.

johnharris85:
OK cool. Will change to an env var. Do you prefer this implementation or the improved one above in my comment that skips the check altogether if it's set?

shin-:
Probably better to skip the check altogether, yes :+1:

brandon-fryslie:
If the check is going to be skipped altogether, '--ignore-orphans' might be a better name for the flag.  It is more consistent with the '--remove-orphans' flag.

johnharris85:
Thanks @brandon-fryslie, this is going to be an env variable now as per discussion. `COMPOSE_IGNORE_ORPHANS` sounds good. Will get round to implementing that this week.

johnharris85:
This is getting some attention over on the original issue so I'm trying to push an update to the PR but am blocked by #5536. Any ideas?

compose/cli/main.py
```diff
@@ -377,9 +377,20 @@ def down(self, options):
             -t, --timeout TIMEOUT   Specify a shutdown timeout in seconds.
                                     (default: 10)
         """
+        environment = Environment.from_env_file(self.project_dir)
+        ignore_orphans = environment.get_boolean('COMPOSE_IGNORE_ORPHANS')
+
+        if ignore_orphans and options['--remove-orphans']:
+            log.info("COMPOSE_IGNORE_ORPHANS is set, --remove-orphans flag is being ignored.")
```
shin-: Maybe `log.warn`? You could even argue for erroring out here, since the two options are conceptually incompatible.

compose/cli/main.py
```diff
@@ -935,12 +946,18 @@ def up(self, options):
...
+        ignore_orphans = environment.get_boolean('COMPOSE_IGNORE_ORPHANS')
+
         if detached and (cascade_stop or exit_value_from):
             raise UserError("--abort-on-container-exit and -d cannot be combined.")
 
         if detached and timeout:
             raise UserError("-d and --timeout cannot be combined.")
 
+        if ignore_orphans and remove_orphans:
+            log.info("COMPOSE_IGNORE_ORPHANS is set, --remove-orphans flag is being ignored.")
```
shin-: ditto

tests/integration/project_test.py
```diff
@@ -1,6 +1,7 @@
 from __future__ import absolute_import
 from __future__ import unicode_literals
 
+import os
```
shin-: Not necessary anymore?

compose/cli/main.py
johnharris85: True, I guess accidentally specifying both here could cause some unexpected behavior if it 'fails' silently. Will change it.

johnharris85:
Also needs a PR to update [docs here](https://docs.docker.com/compose/reference/envvars/) on the `vnext-compose` branch. Have a PR ready to go once this is merged.

EugenMayer:
did i miss something (probably) but why is there no cli-parameter for this? Thanks !

phlax:
@EugenMayer seems like `COMPOSE_IGNORE_ORPHANS` env var works

danger89:
Update docs?

shin-:
@danger89 https://docs.docker.com/compose/reference/envvars/#compose_ignore_orphans

alexandernst:
The warning  is  still visible even after running `export COMPOSE_IGNORE_ORPHANS=True`. Can anybody confirm this?

alexandernst:
Edit: This patch is using `Environment.from_env_file(self.project_dir)`, which means that the `COMPOSE_IGNORE_ORPHANS ` variable is being fetched from `.env` file, which is clearly not what this PR was trying to cover. It's hard to believe this was tested in any way.

Please fix by replacing with `Environment. from_command_line ` 🙏

ExplosiveJam:
```
COMPOSE_IGNORE_ORPHANS=1 docker-compose -f <blabla> down
```
works for me.

egbertvanderhout:
It seems to depend on which command you use.
- [x] `up`
- [x] `down`
- [ ] `run`

The env var is read in `cli/main.py` but only on `up` and `down`. `run` is using `up` at a lower level while the env var `COMPOSE_IGNORE_ORPHANS` is read at a higher level.

My use case where I encounter this issue: having a application stack (nginx, django, postgres) running and then running a one off backup command via a second docker-compose file. It works though the warning about orphaned containers is breaking error checks in backups scripts.

PR #5000, opened 2017-07-06
Faster package querying

and3rson:
During the `docker-py` distribution check, `pip freeze` can be terribly slow if you have a lot of packages installed. Here's my case:

    $ time pip freeze > /dev/null
    real    0m3,930s
    user    0m3,787s
     sys     0m0,093s

    $ pip freeze | wc -l
    173

    $ time docker-compose 2> /dev/null
    real    0m3,443s
    user    0m3,277s
    sys     0m0,143s

After this fix it becomes more than 3x faster:

    $ time docker-compose 2> /dev/null 
    real    0m0,935s
    user    0m0,827s
    sys     0m0,100s

This feature was tested in Python 2.7 & 3.6.

### DIFF
--- a/compose/cli/__init__.py
+++ b/compose/cli/__init__.py
@@ -2,7 +2,6 @@ from __future__ import absolute_import
 from __future__ import print_function
 from __future__ import unicode_literals
 
-import os
 import subprocess
 import sys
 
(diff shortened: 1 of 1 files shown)


and3rson:
The Jenkins build seems to be hanging... Is it normal?

shin-:
https://github.com/docker/compose/pull/4920#issuecomment-310774092

shin-:
This part of the code has been iterated over many times. The current solution is the one that offers the best compromise for every declination of the software.

Issue #4963, opened 2017-06-27
docker-compose pull output


Niondir:
I'm confused every time I do `docker-compose up -d`

```
postgres is up-to-date
nginx is up-to-date
Recreating backend
dashboard is up-to-date
```

The container name is always in front, just for "recreating" it's at the end. I wish you could unify this to make spotting what was recreated and what is up-to-date a bit easier.

nashley:
[service.py](https://github.com/docker/compose/blob/5c5a40c3375693cadaab6d1449e13f4a4bda6047/compose/service.py#L510) looks like the relevant file.

How does `%s is out-of-date. Recreating...` sound?

nashley:
Or perhaps `%s is being recreated`:
```
postgres is up-to-date
nginx is up-to-date
backend is being recreated
dashboard is up-to-date
```

Niondir:
I like all of them. Maybe the first proposal is best, since the line gets a bit longer. You usually want to focus on the recreated stuff I guess.

nashley:
@shin- I made a (very simple) PR. Let me know if you think the message should be changed or if I missed a part of the PR process.

ijc:
Closing as discussed in #5002. Thanks.

PR #5002, opened 2017-07-06
[WIP] Improve the message displayed when recreating containers

nashley:
Change the 'recreating' message to match the 'up-to-date' message
	by having it start with the container name.

Signed-off-by: Nick Ashley <nick@shley.us>

### DIFF
--- a/compose/service.py
+++ b/compose/service.py
@@ -509,7 +509,7 @@ class Service(object):
         volumes can be copied to the new container, before the original
         container is removed.
         """
-        log.info("Recreating %s" % container.name)
+        log.info("%s is out-of-date. Recreating..." % container.name)
 
         container.stop(timeout=self.stop_timeout(timeout))
         container.rename_to_tmp_name()


nashley:
For #4963

nashley:
Do the "testing" and "documentation" requirements still apply to this small patch?
Where should I find the testing and documentation files to update?

nashley:
Thanks @GordonTheTurtle. Not sure how that merge commit snuck in there, but I removed it.

johnharris85:
Hey @nashley ! Testing and docs doesn't really apply to this I don't think. Sometimes these kinds of changes will require test changes as the test is checking the output / logs to match a specific string, but in this case it doesn't seem like it as everything passed.

Don't think it requires a docs change either as it's fairly minor, but FYI the docs are in another repo @ https://github.com/docker/docker.github.io.

:)

nashley:
Looks like the documentation actually contains the 'Recreating" string in the [compose/rails.md](https://github.com/docker/docker.github.io/blob/master/compose/rails.md#connect-the-database) file (see the third codeblock in that section, or just search for it on the page).
I couldn't find it anywhere else with a recursive grep search (`grep -r 'Recreating' . --exclude-dir .gitignore | less`).

Should I submit a PR for the documentation, or should I wait until we're set on what the new message should be?

ijc:
PR grooming: Thanks for this and sorry noone looked at it at the time.

Looking at the current output I see
```
Creating private_nginx_1 ... done
Creating private_nginx_2 ... done
```
vs on update:
```
Recreating private_nginx_1 ... done
Recreating private_nginx_2 ... done
```
Which I think has addressed (or more like obsoleted) your original issue.

So I am going to close this and the linked issue, thanks again for your submission though.

PR #5004, opened 2017-07-06
WIP: Add support for cidfile

### LINKED ISSUE #940: Suport docker --cidfile option
`docker run` has the flag `--cidfile <file>` that writes the container id to the specefied file so that it is possible to read it for other purposes.

It should be a option in the config for one image

``` yaml
web:
  image: mysql:latest
  cidfile: /tmp/mysql.cid
...
```

Supporting this might be a simpler solution to build before `exec` command is supported (as discussed in https://github.com/docker/fig/issues/593)

AndreLouisCaron:
**Warning**: Do not merge this (yet)!  `docker-py` does not yet support the CID file option when creating a container (see docker/docker-py#967 for a tentative PR that adds it).

This is a tentative change to add a `cidfile` option to services.

Fixes #940.

### DIFF
--- a/compose/config/config_schema_v1.json
+++ b/compose/config/config_schema_v1.json
@@ -22,6 +22,7 @@
         "cap_add": {"type": "array", "items": {"type": "string"}, "uniqueItems": true},
         "cap_drop": {"type": "array", "items": {"type": "string"}, "uniqueItems": true},
         "cgroup_parent": {"type": "string"},
+        "cidfile": {"type": "string"},
         "command": {
           "oneOf": [
             {"type": "string"},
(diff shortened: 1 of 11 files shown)


shin-:
In light of the recent discussion https://github.com/docker/docker-py/pull/967#issuecomment-313458043 , it seems the option was misunderstood altogether.

If we do want a `cid_file` config option that matches the Docker CLI, we need to retrieve the container ID(s) and write to the specified file from inside Compose. The `CIDFile` option in the API is essentially a no-op.

@AndreLouisCaron It's up to you if you want to keep working on this PR. My understanding is that the actual functionality doesn't answer your use-case, so it's totally fine if you'd rather close it instead.

AndreLouisCaron:
Indeed, this doesn't help my use case, but @joshuacox seemed very interested in the change.  I don't mind putting a bit more effort into this to land the change if it helps somebody else.

PR #5011, opened 2017-07-10
Add target property to build

### LINKED ISSUE #4790: Add target property to build
Compose should support specifying target stage in the build definition. https://github.com/moby/moby/pull/32496

@dnephin

lawliet89:
Fixes #4790

I had to introduce a new v2.3 Compose format to implement this.


philtay:
@dnephin I think this should be added to v3 as well.

sirlatrom:
Will this also be in a spec version 3.4? I know 3.x ignores and does not support the build option, but having target supported would allow having just one docker-compose.yml file.

Issue #4838, opened 2017-05-17
Feature Request: Access to composer environment variables


oliveiraev:
I have several applications that share the same docker-compose.yaml boilerplate. The **unique** difference is the application name.

I would love to replace all my ```$application_name``` occurrences with ```$COMPOSE_PROJECT_NAME```.

shin-:
I'm not sure I understand the request. Can you elaborate on what you mean? Do you want Compose to populate the container's environment with a `COMPOSE_PROJECT_NAME` variable?

oliveiraev:
For my own use case, the best scenario would be if ```COMPOSE_PROJECT_NAME``` was exposed at ```docker-compose.yaml``` level, so we could do things like
```
version: "3"
services:
    nginx:
        image: nginx:stable-alpine
    volumes:
        - .:/var/www/$COMPOSE_PROJECT_NAME
```
but, once ```COMPOSE_PROJECT_NAME``` is available at ```docker-compose.yaml``` level, I think that automagically we could also do
```
[...]
        environment:
            - $COMPOSE_PROJECT_NAME
```
when relevant to populate container's environment.

shin-:
I see. I believe that's the same ask as #2294. The main problem is it encourages people to use some anti-patterns in their Compose file. But if you really need it, it can already be done, provided you use a custom `COMPOSE_PROJECT_NAME` environment variable.

I'll close here, but feel free to continue the discussion on #2294

oliveiraev:
Sorry for missing that. I'ts exactly the same point. I'll follow there. Thank you.

PR #5026, opened 2017-07-14
Allow using COMPOSE_PROJECT_NAME in environment substitutions

AndreLouisCaron:
When the `COMPOSE_PROJECT_NAME` is defined in the calling environment, it is
available for substitution in the `docker-compose.yml` file.  However, this
cannot be relied upon because the project name can be inferred from the
parent directory name.

This change ensures the `COMPOSE_PROJECT_NAME` environment variable is always
available for subtitution in `docker-compose.yml` file regardless of the
provenance of the project name.

Signed-off-by: Andre Caron <andre.l.caron@gmail.com>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,14 @@
 Change log
 ==========
 
+Unreleased
+----------
+
+#### All formats
+
+- The `COMPOSE_PROJECT_NAME` environment variable is now always available for
+  substitutions even when the value is not explicitly set.
+
 1.14.0 (2017-06-19)
 -------------------
 
--- a/compose/cli/command.py
+++ b/compose/cli/command.py
@@ -88,6 +88,9 @@ def get_project(project_dir, config_path=None, project_name=None, verbose=False,
     project_name = get_project_name(
         config_details.working_dir, project_name, environment
     )
+    # Provide ``COMPOSE_PROJECT_NAME`` for substitutions even
+    # when the value is inferred from the directory name.
+    environment.setdefault('COMPOSE_PROJECT_NAME', project_name)
     config_data = config.load(config_details)
 
     api_version = environment.get(
--- a/tests/fixtures/longer-filename-composefile/docker-compose.yaml
+++ b/tests/fixtures/longer-filename-composefile/docker-compose.yaml
@@ -1,3 +1,5 @@
 definedinyamlnotyml:
   image: busybox:latest
   command: top
+  environment:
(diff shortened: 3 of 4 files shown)


AndreLouisCaron:
A bit of context: I'm trying to build a miniature chaos monkey for local setups in order to battle-test a distributed system that should be resilient to outages.  This change would be really useful to search for other containers in the same project.

shin-:
This was asked for in https://github.com/docker/compose/issues/4838#issuecomment-303194407 - it's part of a larger debate that's been hotly discussed on this issue tracker for a while now. But the gist of the issue is that we discourage people from using project name as a crutch for poor app design. 

I'm happy to go over alternatives to using project name in your app if that's helpful, but I don't think this PR takes things in the right direction for Compose.

AndreLouisCaron:
I went through the other issues you linked to and I'm not sure I understand your objection, but I won't insist if you really don't want this to happen.  Sure, I can work around the issue by using a `.env` file -- it's clunky, but it'll work.

Before trying to use the project name like this, I though I'd "do it right", but listing all containers in the same compose project from inside one of the containers is far from easy because it loses all of the context about the "compose project".  Feel free to recommend an alternate approach here, I'm all ears :-)

hartwork:
I need `${COMPOSE_PROJECT_NAME}` to have one service use the image of another service like this:
```
services:
  one:
    build:
      .............

  two:
    depends_on:
      - one
    image: ${COMPOSE_PROJECT_NAME}_one:latest
```
Use of `${COMPOSE_PROJECT_NAME:-git_repo_basename}` does not work well for me so I need to have `.env` with `COMPOSE_PROJECT_NAME = ...` around, which feels like a bugfix workaround.

If you know a way to do image re-use without use of `${COMPOSE_PROJECT_NAME}`, please let me know.

ijc:
I think this was intended to be closed around https://github.com/docker/compose/pull/5026#issuecomment-315474112 so I'll do so now.

WRT:
> If you know a way to do image re-use without use of ${COMPOSE_PROJECT_NAME}, please let me know.

I think the simple answer is that if you have two services which happen to have the same `build: .` or `build: <somedir>` then due to the caching and sharing done by the `docker build`er infrastructure you won't actually end up building things twice, you'll get the same built content tagged twice (once for each service) sharing layers etc due to the cache hits from whichever got built second. Or at a minimum I would say: if things are not being shared in this way then that is an issue worth raising and attempting to address.

PR #4794, opened 2017-05-05
add exception for windows networking

friism:
Fixes this:

```
λ docker-compose -f docker-compose.windows.yml up
ERROR: Network "musicstore_default" needs to be recreated - option "com.docker.network.windowsshim.hnsid" has changed
```

PR #5050, opened 2017-07-25
Add exception for networkname in windows networking

kirinrastogi:
Similar to #4794

Fixes:
`ERROR: Network "musicstore_default" needs to be recreated - option "com.docker.network.windowsshim.networkname" has changed`

docker_networkname

### DIFF
--- a/compose/network.py
+++ b/compose/network.py
@@ -18,7 +18,8 @@ log = logging.getLogger(__name__)
 
 OPTS_EXCEPTIONS = [
     'com.docker.network.driver.overlay.vxlanid_list',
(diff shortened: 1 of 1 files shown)


shin-:
Thank you!

shin-:
@friism Do you know if there is a way we could get some advance warning when the Windows Engine team adds these, so we can update the whitelist before it becomes a problem?

friism:
lemme ping them

friism:
@kallie-b do you know if you're going to be adding more of these hard-coded networks? It would be good to get a heads-up if you do.

PR #5054, opened 2017-07-25
Escape dollar sign in serialized config output

### LINKED ISSUE #5031: `docker-compose config` should escape the `$` sign
Say we have a `docker-compose.yml` contains the `$` sign, and it should be escaped, like:

```yml
version: '3.3'
services:
  exporter:
    image: prom/node-exporter
    command:
      - -option_without_quote=$$
      - '-option_single_quote=$$'
      - "-option_double_quote=$$"
```

However, when using `docker-compose config` to show this file, the `$` sign are not escaped anymore:

```yml
$ docker-compose -f sample.yml config
services:
  exporter:
    command:
    - -option_without_quote=$
    - -option_single_quote=$
    - -option_double_quote=$
    image: prom/node-exporter
version: '3.3'
```

This will cause the problem when the output will be used in `docker stack deploy` case, as following error will be raised:

`Invalid interpolation format for "command" option in services "exporter": "-option_without_quote=$". You may need to escape any $ with another $.`

shin-:
Fixes #5031

### DIFF
--- a/compose/config/serialize.py
+++ b/compose/config/serialize.py
@@ -21,8 +21,11 @@ def serialize_dict_type(dumper, data):
 
 
 def serialize_string(dumper, data):
-    """ Ensure boolean-like strings are quoted in the output """
+    """ Ensure boolean-like strings are quoted in the output and escape $ characters """
     representer = dumper.represent_str if six.PY3 else dumper.represent_unicode
+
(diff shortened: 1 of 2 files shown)


TotempaaltJ:
This broke my build pipeline.

PR #5055, opened 2017-07-26
only require colorama on windows

carlwgeorge:
Colorama is only useful on Windows by design.  Since it has no effect on other platforms, it makes sense to not require it universially.

### DIFF
--- a/compose/cli/colors.py
+++ b/compose/cli/colors.py
@@ -1,7 +1,7 @@
 from __future__ import absolute_import
 from __future__ import unicode_literals
 
-import colorama
+from ..const import IS_WINDOWS_PLATFORM
 
 NAMES = [
     'grey',
@@ -33,7 +33,9 @@ def make_color_fn(code):
     return lambda s: ansi_color(code, s)
 
 
-colorama.init(strip=False)
+if IS_WINDOWS_PLATFORM:
+    import colorama
+    colorama.init(strip=False)
 for (name, code) in get_pairs():
     globals()[name] = make_color_fn(code)
 
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,7 +4,7 @@ backports.ssl-match-hostname==3.5.0.1; python_version < '3'
 cached-property==1.3.0
 certifi==2017.4.17
 chardet==3.0.4
-colorama==0.3.9
+colorama==0.3.9; sys_platform == 'win32'
 docker==2.4.2
 docker-pycreds==0.2.1
 dockerpty==0.4.1
--- a/setup.py
+++ b/setup.py
@@ -31,7 +31,6 @@ def find_version(*file_paths):
 
 install_requires = [
     'cached-property >= 1.2.0, < 2',
-    'colorama >= 0.3.7, < 0.4',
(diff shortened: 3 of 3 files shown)


carlwgeorge:
AppVeyor appears to be failing because it doesn't recognize the `platform_system` environment marker.

```
SyntaxError: invalid expression: platform_system
```

That marker is valid as best I can tell from the [official docs](https://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-platform-specific-dependencies) as well as [PEP 508](https://www.python.org/dev/peps/pep-0508/#environment-markers).  The AppVeyor output also mentions an outdated version of pip.

```
You are using pip version 7.1.2, however version 9.0.1 is available.
```

Running the tests locally with pip 9.0.1 works.  Is there a way to make AppVeyor use a newer version of pip?

shin-:
Use the `sys_platform` marker instead, as defined in [PEP 496](https://www.python.org/dev/peps/pep-0496/)

carlwgeorge:
@shin- I was ready to switch to the `sys_platform` marker as suggested, but I noticed this in the PEP 496 status paragraph:

> After this PEP was initially drafted, PEP 508 was developed and submitted to fully specify the dependency declaration syntax, including environment markers. As a result, this PEP ended up being rejected in favour of the more comprehensive PEP 508.

That helped me track down the exact issue.  Pip isn't the problem, but rather setuptools (pkg_resources).  PEP 508 was [implemented in setuptools 20.2](https://github.com/pypa/setuptools/blob/20.2/CHANGES.txt#L9-L13).  `pkg_resources.evaluate_marker()` accepts a `platform_system` marker when using setuptools 20.2 or higher, but raises an exception with older versions.

Would setting `setup_requires=['setuptools >= 20.2']` be acceptable in order to be able to use the `platform_system` marker?  I don't know if that would be sufficient to make AppVeyor happy, but I wanted to at least suggest it.

shin-:
We're already using the `sys_platform` marker. Plus, the fact that it's compatible with slightly older versions of setuptools is a good thing as far as we're concerned. :)

compose/cli/colors.py
```diff
@@ -33,7 +33,9 @@ def make_color_fn(code):
     return lambda s: ansi_color(code, s)
 
 
-colorama.init(strip=False)
+if platform.system() == 'Windows':
```
shin-: I would like to use `const.IS_WINDOWS_PLATFORM` here, for consistency.

shin-:
Thank you!

carlwgeorge:
You're welcome.  It makes packaging easier for me.  I don't know if it would be considered useful for the project, but I've got RPM packages available for RHEL/CentOS 7 now.

* [1.9](https://bodhi.fedoraproject.org/updates/FEDORA-EPEL-2017-d409101507) in [EPEL](https://fedoraproject.org/wiki/EPEL) (highest allowed version, limited by Red Hat only having the Python SDK 1 available)
* [1.15.0](https://copr.fedorainfracloud.org/coprs/carlwgeorge/docker-compose/build/584099/) in [COPR](https://copr.fedorainfracloud.org/coprs/carlwgeorge/docker-compose/)

PR #5073, opened 2017-08-01
Fixed handling of SIGHUP to properly stop containers (fix #5072)

### LINKED ISSUE #5072: Containers run with `--rm` option have 'AutoRemove: false' and they are not auto-removed when connection lost
Containers run with `docker-compose run --rm` option are not auto-removed when connection losts. They are still running on the server. They also have `"AutoRemove": false` setting.

Example:

```
$ docker-compose run --rm --name multi_unicorn_222 unicorn  bash -i &

$ docker inspect multi_unicorn_222 | grep AutoRemove
            "AutoRemove": false,
```

Versions:

```
$ docker version
Client:
 Version:      17.06.0-ce
 API version:  1.30
 Go version:   go1.8.3
 Git commit:   02c1d87
 Built:        Fri Jun 23 21:23:31 2017
 OS/Arch:      linux/amd64

Server:
 Version:      17.06.0-ce
 API version:  1.30 (minimum version 1.12)
 Go version:   go1.8.3
 Git commit:   02c1d87
 Built:        Fri Jun 23 21:19:04 2017
 OS/Arch:      linux/amd64
 Experimental: false

$ docker-compose version
docker-compose version 1.6.2, build 4d72027
docker-py version: 1.7.2
CPython version: 2.7.9
OpenSSL version: OpenSSL 1.0.1e 11 Feb 2013
```

907th:
Signed-off-by: Alexey Chernenkov <laise@pisem.net>

### DIFF
--- a/compose/cli/signals.py
+++ b/compose/cli/signals.py
@@ -17,6 +17,7 @@ def shutdown(signal, frame):
 def set_signal_handler(handler):
     signal.signal(signal.SIGINT, handler)
     signal.signal(signal.SIGTERM, handler)
+    signal.signal(signal.SIGHUP, handler)
 
 
 def set_signal_handler_to_shutdown():
--- a/tests/acceptance/cli_test.py
+++ b/tests/acceptance/cli_test.py
@@ -1189,6 +1189,13 @@ class CLITestCase(DockerClientTestCase):
         os.kill(proc.pid, signal.SIGTERM)
         wait_on_condition(ContainerCountCondition(self.project, 0))
 
+    def test_up_handles_sighup(self):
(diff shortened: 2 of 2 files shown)


shin-:
Thank you for the submission!

Unfortunately, I'm afraid this change would have unwanted consequences - for example, if I `docker-compose up` without detaching and lose connection to the server I'm SSH'd to, my expectation would be that my application keeps running. This change would break that expectation.

I believe a better approach would be to set `AutoRemove` accordingly when creating a one-off container when the API supports it (API version > 1.25). That fix would leave out older Compose files, but with v2.x being forward compatible, I think that's a reasonable compromise to make.

907th:
@shin- You're right! I close it.

PR #5075, opened 2017-08-01
Add isolation parameter to v3 compose file schema (fixes #5069).

### LINKED ISSUE #5069: Isolation parameter does not work in compose files with versions >2.3
The isolation parameter was added in #3906, and was released in compose file version 2.1. However, it was never included in the version 3 schema file, so it causes errors with newer compose file versions.

The v3 schema documentation at https://docs.docker.com/compose/compose-file/#isolation indicates that it is still supported so it took me quite a while to find the problem. It seems that the line `"isolation": {"type": "string"}` (from [compose/config/config_schema_v2.3.json#L163](https://github.com/docker/compose/blob/master/compose/config/config_schema_v2.3.json#L163)) was left out in the move to v3.0, and hasn't appeared again since. The parameter is still in compose/service.py and the tests were never removed so this is probably an easy fix.

The compose file I'm working with is intended to be deployed to a swarm, so unfortunately I can't just downgrade it. I've set `"exec-opts": ["isolation=hyperv"]` in `C:\ProgramData\docker\config\daemon.json` on each node to change the default as a workaround for now.

Techwolfy:
Fix for isolation parameter errors in v3+ compose files.

### DIFF
--- a/compose/config/config_schema_v3.0.json
+++ b/compose/config/config_schema_v3.0.json
@@ -105,6 +105,7 @@
         "hostname": {"type": "string"},
         "image": {"type": "string"},
         "ipc": {"type": "string"},
+        "isolation": {"type": "string"},
         "labels": {"$ref": "#/definitions/list_or_dict"},
         "links": {"type": "array", "items": {"type": "string"}, "uniqueItems": true},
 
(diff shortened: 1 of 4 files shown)


shin-:
As I explained in https://github.com/docker/compose/issues/5069#issuecomment-319808451, we can not make this change unless it is an available option for Swarm services. Moreover, the v3 spec should be modified on the docker/cli repo first then reproduced here - not the other way around.

Issue #5091, opened 2017-08-09
Return non-zero exit code if --parallel pull fails


aaronmorgan:
If the pull fails when using `--parallel` the individual image shows `error` but the process exits cleanly, i.e. `$lastexitcode` and `$?` report `0`  and `True` respectively. 

Occurs regardless of whether there's one request and it fails, or many requests and one or more of them fail.

`docker-compose version 1.14.0, build c7bdf9e3`

PR #5094, opened 2017-08-09
Fix exit code 0 upon parallel pull failure.

aronahl:
Implements enhancement #5091.

### DIFF
--- a/compose/project.py
+++ b/compose/project.py
@@ -498,13 +498,15 @@ class Project(object):
             def pull_service(service):
                 service.pull(ignore_pull_failures, True)
 
-            parallel.parallel_execute(
+            _, errors = parallel.parallel_execute(
                 services,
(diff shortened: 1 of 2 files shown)


shin-:
Thanks! Would you mind adding a test for it as well?

aronahl:
Sure.

aronahl:
I added it to the same place I found the other pull tests in cli_test.py.

Do hash-dibs comments work for claiming issues in this repository?

shin-:
LGTM - thank you!

> Do hash-dibs comments work for claiming issues in this repository?

Hmm, I don't think so, why?

aronahl:
I was just curious as to a way to own any other bug that might come up so that I don't duplicate work.  The contributing doc points to a parent spec for moby that refers to the bot for assignments.

PR #5107, opened 2017-08-16
add errors detail message for up project failure

KevinTHU:
we find out that if we got a failure of `docker-compose up`, we cannot get the detail errors from log or stdout, this is not good for us maybe also other compose users, we hope compose maintainers could accept this PR for improving this problem. thanks~

### DIFF
--- a/compose/project.py
+++ b/compose/project.py
@@ -454,7 +454,8 @@ class Project(object):
         )
         if errors:
             raise ProjectError(
-                'Encountered errors while bringing up the project.'
+                'Encountered errors while bringing up the project.\n{0}'.format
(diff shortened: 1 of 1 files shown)


shin-:
The errors appear in the log at the time they occur. They do not need to be repeated out of context at the end of the process.

KevinTHU:
@shin- yes, you are right when we just use docker-compose only, however, if we use other python program to directly call `project.up()` method, we cannot get the right errors and exception~, I feel this is not friendly for compose to be integrated by others.

shin-:
@KevinTHU I see. Unfortunately this is not usage we want to encourage - please keep in mind that we make no guarantees about the stability of our API and that internals may change at any time.

PR #5112, opened 2017-08-16
Enable environment variable chaining

### LINKED ISSUE #4836: Environment variables chaining
given following ```docker-compose.yaml```
```
version: "3"
services:
    sample:
        image: alpine
        environment:
            - USER_PASSWORD
            - ROOT_PASSWORD=${ROOT_PASSWORD:-$USER_PASSWORD}
        command: sh -c "env | sort | grep _PASSWORD"
```
Invocation of ```$ USER_PASSWORD='my own password' docker-compose run``` ends up with

```
Starting sample_sample_1 ... 
Starting sample_sample_1 ... done
Attaching to sample_sample_1
sample_1  | ROOT_PASSWORD=$USER_PASSWORD
sample_1  | USER_PASSWORD=my own password
```

Is there any way to make ```ROOT_PASSWORD``` inherit from ```USER_PASSWORD``` value?

oliveiraev:
closes #4836

This patch let users to set things like...

```
services:
    mysql:
        image: mysql
        enviroment:
            MYSQL_PASSWORD: default_password
            MYSQL_ROOT_PASSWORD: ${MYSQL_ROOT_PASSWORD:-$MYSQL_PASSWORD}
```

signed-off-by: Evandro Oliveira <evandrofranco@gmail.com>

### DIFF
--- a/tests/unit/config/interpolation_test.py
+++ b/tests/unit/config/interpolation_test.py
@@ -146,3 +146,9 @@ def test_interpolate_missing_with_default(defaults_interpolator):
 def test_interpolate_with_empty_and_default_value(defaults_interpolator):
     assert defaults_interpolator("ok ${BAR:-def}") == "ok def"
     assert defaults_interpolator("ok ${BAR-def}") == "ok "
+
+
+def test_interpolate_with_chained_fallback_variables(defaults_interpolator):
+    assert defaults_interpolator("ok ${BAR:-$FOO}") == "ok first"
+    assert defaults_interpolator("ok ${BAR:-$BAZ:-$FOO}") == "ok first"
+    assert defaults_interpolator("ok ${BAR-$FOO}") == "ok "
--- a/compose/config/interpolation.py
+++ b/compose/config/interpolation.py
@@ -82,10 +82,12 @@ class TemplateWithDefaults(Template):
             if named is not None:
                 if ':-' in named:
                     var, _, default = named.partition(':-')
-                    return mapping.get(var) or default
+                    interpolated = self.__class__(default).substitute(mapping)
+                    return mapping.get(var) or interpolated or default
                 if '-' in named:
                     var, _, default = named.partition('-')
-                    return mapping.get(var, default)
+                    interpolated = self.__class__(default).substitute(mapping)
+                    return mapping.get(var, interpolated or default)
                 val = mapping[named]
                 return '%s' % (val,)
             if mo.group('escaped') is not None:


FFLSH:
LGTM - This would be a win for us too 👍

tests/unit/config/interpolation_test.py
```diff
@@ -146,3 +146,9 @@ def test_interpolate_missing_with_default(defaults_interpolator):
 def test_interpolate_with_empty_and_default_value(defaults_interpolator):
     assert defaults_interpolator("ok ${BAR:-def}") == "ok def"
     assert defaults_interpolator("ok ${BAR-def}") == "ok "
+
+
+def test_interpolate_with_chained_fallback_variables(defaults_interpolator):
+    assert defaults_interpolator("ok ${BAR:-$FOO}") == "ok first"
```
FFLSH: Would this be a better format?

${BAR:-${FOO}}

Syntax for compose interpolation is consistent, otherwise it is unclear if we are expecting interpolation from compose or the shell.

tests/unit/config/interpolation_test.py
```diff
@@ -146,3 +146,9 @@ def test_interpolate_missing_with_default(defaults_interpolator):
...
+    assert defaults_interpolator("ok ${BAR:-$BAZ:-$FOO}") == "ok first"
+    assert defaults_interpolator("ok ${BAR-$FOO}") == "ok "
```
FFLSH: Perhaps add another case to check coalesced interpolation:

`assert defaults_interpolator("ok ${BAR:-the $FOO def}") == "ok the first def"`

tests/unit/config/interpolation_test.py
oliveiraev: The simplest implementation that I've made throws "syntax error" while expanding nested curly braces. But I'll take a deeper look.

tests/unit/config/interpolation_test.py
oliveiraev: This shomehow conflicts whith the previows review, not?

How about
```python
assert defaults_interpolator("ok ${BAR:-the ${FOO} def}") == "ok the first def"
```
Is it ok?

tests/unit/config/interpolation_test.py
oliveiraev: Also, I should remember that compose **do** expands variables outside curly braces:

> Both $VARIABLE and ${VARIABLE} syntax are supported. Additionally when using the 2.1 file format, it is possible to provide inline default values using typical shell syntax:
https://docs.docker.com/compose/compose-file/#variable-substitution

And the same docs advise to use double-dollar when you want a single dollar:
> If you forget and use a single dollar sign ($), Compose interprets the value as an environment variable and will warn you:
>
> The VAR_NOT_INTERPOLATED_BY_COMPOSE is not set. Substituting an empty string.

So, there is no "shell expansion" inside docker-compose.yaml. Even ```$$variables``` are translated to literal ```$variables``` inside the containers, which is exactly what the [referred issue](#4836) states.

tests/unit/config/interpolation_test.py
FFLSH: Yes - depends on if the previous comment.

tests/unit/config/interpolation_test.py
FFLSH: Okay. As long as `the ${BAR:-$$FOO}` results in `the $FOO` and `the $BAR:-$FOO` results in `the first` (when $FOO has a value of `first`) then I am happy with what you have now.

FFLSH: Would this cause problems in the following case:

`APIURL=${APP_APIURL:-http://$ENVIRONMENTsite.com/api}`

Where `ENVIRONMENT` is `dev` I would expect `APIURL` to come out as `http://devsite.com/api`

FFLSH:
@oliveiraev Any progress on this after my review? I can help out if you need.

shin-:
See #5268

PR #5115, opened 2017-08-17
Add support for squash in build definition

### LINKED ISSUE #4235: Add support for the --squash flag
See https://github.com/docker/docker/pull/22641

shin-:
Fix #4235

### DIFF
--- a/compose/config/config.py
+++ b/compose/config/config.py
@@ -1019,6 +1019,7 @@ def merge_build(output, base, override):
     md.merge_scalar('dockerfile')
     md.merge_scalar('network')
     md.merge_scalar('target')
+    md.merge_scalar('squash')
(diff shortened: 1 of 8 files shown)


shin-:
cc @dnephin Is it too late to add this to the 3.4 definition?

dnephin:
We should still have time to add it.

Squash is experimental, and now that multi-stage build exists there was talk of removing it instead of making it non-experimental, so maybe we shouldn't support it?

shin-:
Oh, i was not aware. Do you know if there's a public issue discussing the removal?

In any case, I will shelf this until we have a decision then.

dnephin:
I don't think there was an open issue yet, just comments in other related issues, so I created https://github.com/moby/moby/issues/34565.

Issue #3866, opened 2016-08-22
Add support for shm_size during image build also


pzrinscak:
I need to set shm_size during image build to build official Oracle database image.

With docker it would be like this

docker build --shm-size=1G

seanfdnn:
Concur. Just ran into this limitation myself.

PR #5116, opened 2017-08-18
Added shm_size to build-opts (issue #3866)

hoogenm:
Additional option for docker-compose to enlarge *shm-size during builds*. I hope this provides a solution to #3866. Is this option useful for the official compose package and the (next) compose 3.x format?

```
version: '3.5'

services:
  example:
    build:
      context: .
      shm_size: 100663296 # 96 MB
```
EDIT: memory units like k/m/g now supported as well. To be able to use the option, please use the (provisional) compose file format 3.5 *or* format 2.3.


shin-:
Thank you! I think this is useful, there's only a few things:

1. It's probably too late to add anything to the 3.4 format right now, as the 17.07 engine release has already had a few RCs and is code-frozen. We can add a tentative 3.5 format for now, and also add the option to the v2.3 format
2. You'll need to squash those commits into a single one :)
3. You can use [`compose.utils.parse_bytes`](https://github.com/docker/compose/blob/master/compose/utils.py#L142-L146) to support values with units

cc @dnephin for v3 schema stuff.

dnephin:
Thanks, opened https://github.com/docker/cli/issues/464 to track

hoogenm:
@shin- and @dnephin: thank you for your timely comments / opening the issue to track the PR! In the mean time, I have made serious efforts to add the `shm_size` build-option to 1) a newly added (provisional) 3.5 format, add the option to the v2.3 format, 2) squash my commits into one single commit, and 3) support values with units through `compose.utils.parse_bytes` as suggested (if an integer value is provided without a unit, it will still be interpreted as 'bytes').

hoogenm:
@shin- and @dnephin:

I noticed that in the mean time, in the official master branch, the compose-schema `3.4` is renamed to `3.4-beta`. I may need to update my PR. Would you like me to merge the `3.4-beta`-change into my PR (to resolve the conflict reported)? Do you expect the version number to stay `3.4-beta` at the time 1.17.0 will be released?

Note that I had also tentatively added a `3.5` format for the `shm_size`-*build* option as requested. Would you like me to add the `-beta` postfix to that version number as well? Or, alternatively, can I 'backport' the `shm_size` to the `3.4-beta` format, now that the `3.4-beta` *version/number* is explicitly marked as beta? Feel free to let me know.

shin-:
@hoogenm Sorry, I was away for a couple weeks. I think the next engine release is in code freeze at the moment and will include 3.4, so we should continue to target 3.5 as originally planned. @dnephin WDYT?

dnephin:
3.5 sounds right

shin-:
PR looks good to me. We might have to exclude 3.5 from the accepted list until it's frozen in an engine release, but that's outside the scope of this proposal.

hoogenm:
Thank you for the comments / feedback, I appreciate it.

PR #5125, opened 2017-08-22
Fix --no-ansi flag not working properly

cecton:
I just realized while testing that the statuses have colors which are also control chars.


cecton:
hey @shin- ! I'm not sure how I manage to write such bad test the other day but I realized now that the 1.16.0rc1 has still issues with control characters.

I have improved the test to specifically test for control characters this time and also for testing without and without the noansi flag.

Sorry for the inconvenience!

shin-:
Thank you! Closing via #5133

Issue #2147, opened 2015-10-07
Make nightly builds available


michaeljs1990:
Possibly this is already done but I can't find anything about it. I tried running...

```
pip install https://github.com/docker/compose/archive/master.zip
```

But i get an odd error saying ImportError: No module named queue.

Anyway since this project on a weekly basis is merging in PRs that make my life significantly easier such as the recent addition of allowing multiple files with the -f flag I feel like it would be worthwhile to have a nightly available.

dnephin:
I think this issue is a duplicate of #1070 which is in progress.

However, I wouldn't expect the current master to be raising a queue error either. What OS and python version are you running?

If you extra the contents of master.zip and `pip install -r requirements.txt` does the error go away?

Thanks!

michaeljs1990:
I am running on version osx 10.11 and when I try and pip install -r requirements.txt i get the error found in this post http://stackoverflow.com/questions/31900008/oserror-errno-1-operation-not-permitted-when-installing-scrapy-in-osx-10-11. I don't want to mess with that one much :-1: since i have a feeling i'm a few short rm commands away from messing up my system at that point... Really don't like mac sometimes.

dnephin:
So I'm guessing that's python3, but compose master should be working against python3.

Could you paste the whole stack trace? I'd like to find out why it's failing to install

michaeljs1990:
sure thing.

```
pip install -r requirements.txt 
Requirement already satisfied (use --upgrade to upgrade): PyYAML==3.10 in /Library/Python/2.7/site-packages (from -r requirements.txt (line 1))
Requirement already satisfied (use --upgrade to upgrade): docker-py==1.4.0 in /Library/Python/2.7/site-packages (from -r requirements.txt (line 2))
Requirement already satisfied (use --upgrade to upgrade): dockerpty==0.3.4 in /Library/Python/2.7/site-packages (from -r requirements.txt (line 3))
Requirement already satisfied (use --upgrade to upgrade): docopt==0.6.1 in /Library/Python/2.7/site-packages (from -r requirements.txt (line 4))
Requirement already satisfied (use --upgrade to upgrade): enum34==1.0.4 in /Library/Python/2.7/site-packages (from -r requirements.txt (line 5))
Requirement already satisfied (use --upgrade to upgrade): jsonschema==2.5.1 in /Library/Python/2.7/site-packages (from -r requirements.txt (line 6))
[...]

PR #5129, opened 2017-08-24
Rename 3.4 schema to 3.4-beta

### LINKED ISSUE #5128: 3.4 not supported in 1.16-rc1
I try to test target option for 3.4 with
```
version: '3.4'
services:
  test:
    image: test
    build:
      context: .
      target: test
```

and got 

```
ERROR: Version in "./docker-compose.yml" is unsupported. You might be seeing this error because you're using the wrong Compose file version. Either specify a supported version (e.g "2.2" or "3.3") and place your service definitions under the `services` key, or omit the `version` key and place your service definitions at the root of the file to use version 1.
For more on the Compose file format versions, see https://docs.docker.com/compose/compose-file/
```
With '2.3' version it works

docker -v
Docker version 17.06.1-ce, build 874a737
docker-compose -v
docker-compose version 1.16.0-rc1, build 9bea70a

shin-:
On the Docker Engine side, the 3.4 format is still in development until the 17.08 release. This change will let people use the current version using the `3.4-beta` version marker.

Also adding the schema to `docker-compose.spec`, fixes #5128

### DIFF
--- a/compose/config/config_schema_v3.4-beta.json
+++ b/compose/config/config_schema_v3.4-beta.json
@@ -1,6 +1,6 @@
 {
   "$schema": "http://json-schema.org/draft-04/schema#",
-  "id": "config_schema_v3.4.json",
+  "id": "config_schema_v3.4-beta.json",
   "type": "object",
   "required": ["version"],
 
--- a/compose/const.py
+++ b/compose/const.py
@@ -31,7 +31,7 @@ COMPOSEFILE_V3_0 = ComposeVersion('3.0')
 COMPOSEFILE_V3_1 = ComposeVersion('3.1')
 COMPOSEFILE_V3_2 = ComposeVersion('3.2')
 COMPOSEFILE_V3_3 = ComposeVersion('3.3')
-COMPOSEFILE_V3_4 = ComposeVersion('3.4')
+COMPOSEFILE_V3_4 = ComposeVersion('3.4-beta')
 
 API_VERSIONS = {
     COMPOSEFILE_V1: '1.21',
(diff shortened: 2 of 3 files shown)


schmunk42:
@shin- How can I test `3.4-beta` or better how do I get a nightly (static) binary of `docker-compose`?

Related: https://github.com/docker/compose/issues/2147

shin-:
It's in 1.16.1, just use "3.4-beta" in the version field of your Compose file. Note that it will disappear in the next release though.

schmunk42:
But it appears it does not support the `x-[...]` syntax from `3.4` mentioned in https://github.com/docker/cli/pull/452

```
$ docker-compose -v
docker-compose version 1.16.1, build 6d1ac21
```

docker-compose.yml

```
version: '3.4-beta'
x-extends: &anchor
  - one
  - two
services:
  nginx:
    image: nginx
```

Error message

```
$ docker-compose config
Additional properties are not allowed ('x-extends' was unexpected)

You might be seeing this error because you're using the wrong Compose file version. Either specify a supported version (e.g "2.2" or "3.3") and place your service definitions under the `services` key, or omit the `version` key and place your service definitions at the root of the file to use version 1.
For more on the Compose file format versions, see https://docs.docker.com/compose/compose-file/
```

shin-:
Yes - it will be in regular 3.4 in the next release. https://github.com/docker/compose/pull/5140

schmunk42:
I understand. But is there any chance to get nightly builds of binaries? https://github.com/docker/compose/issues/2147

shin-:
@schmunk42 Yep, those should be up to date again, we checked with @dnephin this morning: https://dl.bintray.com/docker-compose/master/

magicalbanana:
@shin I forget, but when is this going to be released?

shin-:
Beginning of November

Issue #5062, opened 2017-07-28
Docker Compose "images" prints the registry port in the wrong column


AnthonyMastrean:
If I have a Docker image reference including a full registry host name and port

```
example.com:8443/example
```

It's not displayed properly in the images table

```
$ docker-compose images
   Container         Repository       Tag         Image Id       Size
----------------------------------------------------------------------
example_example_1   example.com   8443/example   abcdef123456   173 MB
```

shin-:
That's not actually a valid image reference - it'd need to follow the `<host>/<namespace>/<repo>` convention.

Even so, that data is retrieved from the engine API, so it's not a Compose bug. You would see the same issue in the output of `docker inspect example_example_1`

AnthonyMastrean:
OK, I'll take the report over to the docker/docker repository. I'm able to push and pull with just the registry host name, is it possible my registry (Nexus) is rewriting the URL behind the scenes or is the Docker API just being flexible?

I'll try the same with the proper image reference, too.

tossmilestone:
@shin- I think this is a Compose bug, and I have create a pull request to fix it, PTAL.

PR #5130, opened 2017-08-24
Fix 'images' prints wrong tag

tossmilestone:
Fixed issue #5062

Signed-off-by: He Xiaoxi <tossmilestone@gmail.com>


shin-:
Thank you for calling it out! I went with a slightly different, more concise approach in #5132

PR #5136, opened 2017-08-25
Add bash completion for `--no-ansi`

albers:
This adds bash completion for #5093.

Also added a comment to make clear why there are two places where global options are defined.

Please add this to _v1.16.0_ as the corresponding feature is added in that release.


shin-:
Thank you @albers !

PR #5186, opened 2017-09-12
Make result of docker-compose config with order

denverdino:
Signed-off-by: Li Yi <denverdino@gmail.com>

In current implementation, the config is using the dict, so the order for the result of ```docker-compose config``` is not ensured. 

In this PR, we replace the dict with OrderedDict, so we can make the result of docker-compose config consistent with the original template

### DIFF
--- a/compose/config/config.py
+++ b/compose/config/config.py
@@ -7,6 +7,7 @@ import os
 import string
 import sys
 from collections import namedtuple
+from collections import OrderedDict
 
 import six
 import yaml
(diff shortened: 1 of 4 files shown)


ijc:
Hi @denverdino, I'm sorry noone looked at this at the time.

Is the issue you were solving with these changes still present?

I tried with:
```yaml
web:
  build: .
  links:
    - consul
    - postgres
    - redis
  environment:
    - FOO=BAR
    - ABC=DEF
    - XYZ=XYZ

consul:
  image: consul:latest
postgres:
  image: postgres:latest
redis:
  image: redis:latest
```

and then `docker-compose -f 5186.yaml config` is producing the exact same output every time, so things don't appear to be being obviously reordered each time. OTOH I don't really see where in the code this is being arranged.

ulyssessouza:
Stale

PR #5195, opened 2017-09-18
Simple patch to allow s390x images to be built

FrenchBen:
Needs integration with CI and s390x machine integration
Build is currently manual and will need full CI integration or better build. 
Pyinstaller currently doesn't support Z machine thus can't build the compact binary.

/cc @shin-


ry4nz:
Can it built on top of alpine?  Looks like docker/compose:1.15.0 is on Alpine.

PR #5213, opened 2017-09-22
[cli] Lift artificial limitation on --build-arg

artagnon:
Currently, `docker-compose --build-arg` requires that a service be
specified as part of the command-line invocation. So,

  $ docker-compose build --build-arg nocache=`git rev-parse @` foom

works. However, when using out-of-band scripts to automate the build
process of several Docker containers (in a CI system, for instance), it
becomes difficult to specify exactly which service requires the
build-arg. Docker has supported Dockerfiles that ignore build-args for a
long time, so there is no problem is specifying spurious build-args to
builds that don't consume it.

The limitation on `docker-compose build` today is artificial, and there
are no other commands that require specifying a service. Allow
`--build-arg` to also match all services so this is possible:

  $ docker-compose build --build-arg nocache=`git rev-parse @`

Please refer to #3790 for discussion on the original feature.

Signed-off-by: Ramkumar Ramachandra <artagnon@gmail.com>

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -230,19 +230,15 @@ class TopLevelCommand(object):
             --force-rm              Always remove intermediate containers.
             --no-cache              Do not use cache when building the image.
             --pull                  Always attempt to pull a newer version of the image.
-            --build-arg key=val     Set build-time variables for one service.
+            --build-arg key=val     Set build-time variables for service(s).
         """
-        service_names = options['SERVICE']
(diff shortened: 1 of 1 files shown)


shin-:
Looks like the change in engine that allows this occurred in 1.13.0 [1], which corresponds to API version 1.25. We'll at least need to gate this feature behind a version check

[1] https://github.com/moby/moby/pull/27412

PR #5218, opened 2017-09-27
Revert 3.4-beta temporary rename

shin-:


sirlatrom:
I'd argue this would be enough for a new release, given how 0.16.0 and 0.16.1 can't use "3.4" and Docker Engine 17.09 CE can.

PR #5234, opened 2017-10-03
Build labels option: array form produces unmarshal error (fixes #5183)

### LINKED ISSUE #5183: Array form of #labels produces unmarshal error
Consider the following docker-compose.yml: 
```
version: '2.1'
services:
  labels_as_array:
    build:
      context: .
      labels:
        - "com.example.description=Description"
        - "com.example.department=Department"
        - "com.example.label-with-empty-value"
```
which is canonical sample from [reference v2](https://docs.docker.com/compose/compose-file/compose-file-v2/#labels)

The Dockerfile for the sample is irrelevalt and is simply ```FROM busybox```

However, running `docker-compose up`, produces:
```
$ docker-compose -v
docker-compose version 1.16.0-rc1, build 9bea70a
$ docker-compose up
Building labels_as_array
ERROR: json: cannot unmarshal array into Go value of type map[string]string
```

Note that the same sample is also listed for v3 [here](https://docs.docker.com/compose/compose-file/#labels), but changing the version to `'3.3'` in the sample above does not help.

garribas:


garribas:
Squashed and pushed, the 2 acceptance tests from #5242 are failing though.

PR #5237, opened 2017-10-05
docker-compose exec doesn't have -e option (fixes #4551)

### LINKED ISSUE #4551: docker-compose exec doesn't have -e option?
docker-compose **run**:
```
Run a one-off command on a service.

For example:

    $ docker-compose run web python manage.py shell

By default, linked services will be started, unless they are already
running. If you do not want to start linked services, use
`docker-compose run --no-deps SERVICE COMMAND [ARGS...]`.

Usage: run [options] [-p PORT...] [-e KEY=VAL...] SERVICE [COMMAND] [ARGS...]

Options:
    -d                    Detached mode: Run container in the background, print
                          new container name.
    --name NAME           Assign a name to the container
    --entrypoint CMD      Override the entrypoint of the image.
    -e KEY=VAL            Set an environment variable (can be used multiple times)
    -u, --user=""         Run as specified username or uid
    --no-deps             Don't start linked services.
    --rm                  Remove container after run. Ignored in detached mode.
    -p, --publish=[]      Publish a container's port(s) to the host
    --service-ports       Run command with the service's ports enabled and mapped
                          to the host.
    -T                    Disable pseudo-tty allocation. By default `docker-compose run`
                          allocates a TTY.
    -w, --workdir=""      Working directory inside the container
```
docker-compose **exec**:
```
Execute a command in a running container

Usage: exec [options] SERVICE COMMAND [ARGS...]

Options:
    -d                Detached mode: Run command in the background.
    --privileged      Give extended privileges to the process.
    --user USER       Run the command as this user.
    -T                Disable pseudo-tty allocation. By default `docker-compose exec`
                      allocates a TTY.
    --index=index     index of the container if there are multiple
                      instances of a service [default: 1]
```

Why doesn't exec support environment command? Documentation for docker-compose exec reads "This is equivalent of docker exec." And documentation for docker exec DOES list --env parameter.

garribas:

Had to refactor a little bit method TopLevelCommand.exec_command to reduce cyclomatic complexity and avoid pre commit hooks error.

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -14,6 +14,8 @@ from distutils.spawn import find_executable
 from inspect import getdoc
 from operator import attrgetter
 
+import docker
+
 from . import errors
 from . import signals
 from .. import __version__
@@ -402,7 +404,7 @@ class TopLevelCommand(object):
         """
         Execute a command in a running container
 
-        Usage: exec [options] SERVICE COMMAND [ARGS...]
+        Usage: exec [options] [-e KEY=VAL...] SERVICE COMMAND [ARGS...]
 
         Options:
             -d                Detached mode: Run command in the background.
@@ -412,10 +414,15 @@ class TopLevelCommand(object):
                               allocates a TTY.
             --index=index     index of the container if there are multiple
                               instances of a service [default: 1]
+            -e, --env KEY=VAL Set environment variables (can be used multiple times,
+                              not supported in API < 1.25)
         """
         index = int(options.get('--index'))
         service = self.project.get_service(options['SERVICE'])
         detach = options['-d']
+
+        if options['--env'] and docker.utils.version_lt(self.project.client.api_version, '1.25'):
+            raise UserError("Setting environment for exec is not supported in API < 1.25'")
 
         try:
             container = service.get_container(number=index)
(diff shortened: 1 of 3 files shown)


compose/cli/main.py
```diff
@@ -453,6 +437,10 @@ def exec_command(self, options):
             "stdin": tty,
         }
 
+        # setting environment for exec is not supported in API < 1.25'
+        if docker.utils.version_gte(self.project.client.api_version, '1.25'):
```
shin-: We should call out that limitation in the docstring above, and raise a `UserError` if the API version is too low.

tests/acceptance/cli_test.py
```diff
@@ -1320,6 +1321,31 @@ def test_exec_custom_user(self):
...
+            'exec',
+            '-T',
+            '-e', 'foo=notbar',
+            '--env', 'alpha=beta',
+            name,
+            'env',
+        ])
+
+        # env overridden
+        self.assertIn('foo=notbar', stdout)
```
shin-: nit: we can use `assert 'foo=notbar' in stdout` instead (and similarly for the lines below), which looks a bit nicer.

garribas:
Thanks for the feedback, made requested changes.

PR #5238, opened 2017-10-05
Search .env file in parent directories

hennevogel:
Extracts the methods responsible for this into their own module and uses this
module for searching yml and .env files. This follows the same rules as the yml
files.

So you can have base_dir/.env and no project_dir/.env or project_dir/.env.
Before .env file was only used if you invoke from the base directory.

### DIFF
--- a/compose/config/config.py
+++ b/compose/config/config.py
@@ -12,6 +12,7 @@ import six
 import yaml
 from cached_property import cached_property
 
+from . import default_config
 from . import types
 from .. import const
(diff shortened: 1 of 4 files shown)


hennevogel:
Not sure this warrants a new test as this whole `.env` file feature is basically untested...

hennevogel:
Also not sure what jenkins want's to tell me. Maybe someone can translate? 🤠

ulyssessouza:
Stale

PR #5241, opened 2017-10-06
Allow port ranges like 8000-8010:80 (fixes #5230)

### LINKED ISSUE #5230: Port range like 8000-8010:80
PR #4649 says that it fixes #4378, which is the usage of port range in docker-compose.
But it does not work (or I'm missing something).
I've made a very simple docker-compose to test it
```
version: '3.3'
services:
  nginx:
    image: nginx
    ports:
    - 4200-4500:80
```
with docker-compose version : 1.17.0dev (same result with 1.16.1)
and docker : 17.06.0-ce

I've checked the source code, and it seems that there is no test for this use case.
The docker-compose.yml in the test/fixtures/ports-composefile miss an case with port range.

garribas:

### DIFF
--- a/compose/config/types.py
+++ b/compose/config/types.py
@@ -310,6 +310,12 @@ class ServicePort(namedtuple('_ServicePort', 'target published protocol mode ext
                             result.append(
                                 cls(target, pub[1], proto, None, pub[0])
                             )
+                        elif isinstance(pub, six.string_types) and '-' in pub:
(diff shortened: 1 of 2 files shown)


garribas:
@guilbill A proper acceptance test can't be added until #5249 is fixed.

shin-:
Thanks - I went with a different solution in #5428 !

Issue #4801, opened 2017-05-08
Progress markers are not shown correctly for `docker-compose up`


jpetazzo:
Note: this might be a 1.13 regression, since I don't remember seeing it with 1.12.

**What I do:**

I have a simple Compose file with two NGINX servers.

```bash
cat >docker-compose.yml <<EOF
www1:
  image: nginx
www2:
  image: nginx
EOF
```

I start them.

```bash
docker-compose up -d
```

**What I expect to see:**

The radiant beauty of the first-class, Python-powered, terminal-aware, ANSI-enlightened CLI which makes Compose a perpetual joy and delight to use and brings happiness to my darkest hours, i.e.:

```
$ docker-compose up -d
Creating tmp_www1_1 ... done
Creating tmp_www2_1 ... done
Starting tmp_www1_1 ... done
Starting tmp_www2_1 ... done
```

**What I see instead:**

Stuff gets slightly glitchy, i.e.:

```
$ docker-compose up -d
Creating tmp_www2_1 ... 
Creating tmp_www1_1 ... 
Creating tmp_www2_1
Creating tmp_www1_1 ... done
```

**Additional notes:**

While `up` is playfully humoring me, `down` won't let me, ahem, *down*, if I may intend the pun:

```
$ docker-compose down
Stopping tmp_www1_1 ... done
Stopping tmp_www2_1 ... done
Removing tmp_www1_1 ... done
Removing tmp_www2_1 ... done
```

It appears that `scale` took sides with `up`:

```
$ docker-compose scale www1=2
Starting tmp_www1_1 ... done
Creating tmp_www1_2 ... 
Creating tmp_www1_2 ... done
```

**Is this an important issue?**

I don't know! What is important? Is the breeze that blows through the bubbles of dew on morning leaves of blue mint important? In the grand scheme of things, it's not a big deal, but I know that my Compose fellows are perfectionists, and that they shan't have any of this. 

In all seriousness, though: I'm so used to Compose embodying the eternal jewel of software perfection, that when I saw this, I thought my containers had been lost at sea. So I think it's fix-worthy.

♥,

stealthybox:
10/10 💯🎨 read it twice.

I couldn't reproduce this on Docker-for-Windows with **compose 1.12**:
```bash
λ docker-compose version
docker-compose version 1.12.0, build ee0f34e1
docker-py version: 2.2.1
CPython version: 2.7.13
OpenSSL version: OpenSSL 1.0.2j  26 Sep 2016
```
[...]

PR #5250, opened 2017-10-10
Progress markers are not shown correctly for docker-compose up (fixes…

garribas:
… #4801)

The `Project.up()` method chains multiple calls to `parallel_execute`.

Given N as the number of services, parallel_execute() is called 1 + N times, each call has no knowledge about the total number of lines to be displayed and as a result progress is displayed on the last line for all services:
```
Project.up()
-> 1 x parallel_execute()
-> N x Service.execute_convergence_plan -> Service._execute_convergence_create 
-> parallel_execute()
```

Refactored so that there's a way to know the overall services associated with a project (method Project.get_scaled_services()) and used ServiceName named tuple as argument to parallel_execute (instead of integers) to pass additional state.

### DIFF
--- a/compose/parallel.py
+++ b/compose/parallel.py
@@ -26,7 +26,7 @@ log = logging.getLogger(__name__)
 STOP = object()
 
 
-def parallel_execute(objects, func, get_name, msg, get_deps=None, limit=None):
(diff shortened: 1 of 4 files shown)


shin-:
Thanks, changes LGTM.

While testing, I found that we were still getting redundant messages when recreating containers, which should be fixed by #5362

PR #5266, opened 2017-10-14
Fix secret location with absolute paths

shin-:
Fix #5206


AndreaGiardini:
Sorry, I didn't have time to write the tests, glad you made it :)

PR #5268, opened 2017-10-16
Allow nested default variables

### LINKED ISSUE #4836: Environment variables chaining

FFLSH:
closes #5112
closes #4836

Building on #5112 (since abandoned) this introduces multi-level nested defaults for environment variables.

Signed-off-by: Fraser Lowndes <fraserlowndes@etech.net>

### DIFF
--- a/compose/config/interpolation.py
+++ b/compose/config/interpolation.py
@@ -71,7 +71,7 @@ def recursive_interpolate(obj, interpolator):
 
 
 class TemplateWithDefaults(Template):
-    idpattern = r'[_a-z][_a-z0-9]*(?::?-[^}]+)?'
+    idpattern = r'[_a-z][_a-z0-9]*(?::?-.+)?'
 
     # Modified from python2.7/string.py
     def substitute(self, mapping):
@@ -82,10 +82,12 @@ class TemplateWithDefaults(Template):
             if named is not None:
                 if ':-' in named:
                     var, _, default = named.partition(':-')
-                    return mapping.get(var) or default
+                    interpolated = self.__class__(default).substitute(mapping)
+                    return mapping.get(var) or interpolated or default
                 if '-' in named:
                     var, _, default = named.partition('-')
-                    return mapping.get(var, default)
+                    interpolated = self.__class__(default).substitute(mapping)
+                    return mapping.get(var, interpolated or default)
                 val = mapping[named]
                 return '%s' % (val,)
             if mo.group('escaped') is not None:
--- a/tests/unit/config/interpolation_test.py
+++ b/tests/unit/config/interpolation_test.py
@@ -125,6 +125,8 @@ def test_invalid_interpolation(defaults_interpolator):
         defaults_interpolator('${foo }')
     with pytest.raises(InvalidInterpolation):
         defaults_interpolator('${foo!}')
+    with pytest.raises(InvalidInterpolation):
+        defaults_interpolator('${foo:}')
 
 
 def test_interpolate_missing_no_default(defaults_interpolator):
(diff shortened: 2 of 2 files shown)


FFLSH:
@shin- PTAL 😄

shin-:
My main issue with this is that it's a breaking change and we have to be very careful with introducing it, especially if it diverges from the `docker stack deploy` behavior. 
Maybe we could introduce a different syntax to indicate the default value needs to be expanded as well, and only activate it in a newly introduced Compose format. But as it stands, it's just going to break too many apps to be viable.

FFLSH:
@shin- Could you clarify the breaking changes you are seeing? The only one I can think of is compose files with environment variables such as `$FOO:-$BAR` which currently resolve to the literal string `$BAR` but after this change will resolve to the value of `BAR` or an empty string.

shin-:
Yes.

FFLSH:
@shin- Yes, as in that is the only breaking change you are thinking of?

memotype:
Just chiming in with the +1. I'm trying to find a way to do something along the lines of 

```
volumes:
  - ${BUILDDIR:-${PWD}/build}:/mnt/build
```

Also, alternatively, it would be cool if you could use environment variables in the .env file, but that doesn't seem to work either, it just takes the "${...}" literally.

JasonAGiles:
I would actually see the exclusion of this functionality to be a bug. The syntax that this is emulating ([shell parameter expansion](http://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html)) allows for these to be nested recursively. I actually found it quite frustrating to encounter a construct which I thought to be familiar, only to be confused when I started getting errors because the parameters weren't resolving to the values I thought they were.

I know this PR is quite old, but would it be possible to reconsider your position on this being a breaking change versus it being a bug in the way this syntax was implemented?  Especially if the values took the form of `${FOO:-${BAR:-default}}` where the intention is clear that the user is trying to nest variable substitution and not substitute the raw value of `${BAR:-default}`

shin-:
There's nothing to reconsider. If `f(A) = B` in an official release, and `f(A) = C` in a subsequent release, it's a breaking change. There's no ideological principle in play here.

memotype:
> There's nothing to reconsider. If `f(A) = B` in an official release, and `f(A) = C` in a subsequent release, it's a breaking change. There's no ideological principle in play here.

@shin- I highly doubt anyone is actually using this construct in the way it's currently implemented. If anyone can find an existing example of someone specifying "${FOO:-$BAR}" in a way that they actually expect the literal string "$BAR" in their variable, I'll eat my shoe.

memotype:
I've started a thread on the community forums to see if anyone can provide a real-life example of using the current semantics intentionally. https://forums.docker.com/t/fr-use-of-nested-variables-in-default-expansion-foo-bar/60114

nijikokun:
I just ran into this. Is there a way around it?

nickjj:
I hope this gets added. It would be a big help in some cases.

For example, here's 1 use case https://github.com/docker/compose/issues/6439.

gustavovalverde:
How is this not being considered a bug?

I completely agree with this comment: https://github.com/docker/compose/pull/5268#issuecomment-427935602

No one is expecting the actual behavior (and I highly doubt that there's a real use case for it or someone using it that way on purpose). All the issues opened by this behavior are stating the same.

Even the documentation state that typical shell syntax can be used, which is not true with this bug:
image

ChanderG:
This is super disappointing.

Ricordel:
I was also trying something like this
```
volumes:
  - ${BUILDDIR:-${PWD}/build}:/mnt/build
```
and was disappointed by the result.

I understand the argument about breaking changes in general, but not so much here considering that docker-compose files are versioned. Can we at least expect a next major version to include this breaking change?

Or would it be acceptable to have a configuration switch in the compose file?

ulyssessouza:
Stale. Issues that it addresses are already closed

treksler:
This is insane. If this isn't a bug, i don't know what is.
This needs to be reopened and merged ASAP.

If this breaks the code of people who are relying on buggy behaviour, so be it. 
That is the case with literally all software bugs that are fixed.

memotype:
@treksler 

https://xkcd.com/1172/

nijikokun:
@ulyssessouza it's a bug, how is it stale?

CHeavyarms:
With the Mirantis acquisition, are we, the community, able to reopen this and appeal to rational decision making and potentially be met with less abrasive gatekeeping behavior, or are we still staring helplessly into an obstinate void?

BinaryBen:
This is a bug. Please implement PR

ubuvoid:
The existing behavior is unquestionably broken. Please fix.

maudrid:
Is there a way to work around this? I need this functionality.
It will save my days of work, now I need to keep merging 2 projects because I can't make this configurable.

ikedam:
Though docker-compose v1 finally haven't support this feature, compose v2 supports this since v2.2.0.

You'd better to try compose v2.18.0 or later as compose < v2.18.0 has a bug in the checking formats for nested variables.

PR #5278, opened 2017-10-17
allows to use dash in COMPOSE_PROJECT_NAME

ev4ern:
I created this containers with docker-py

```
docker ps -f name=vps --format {{.Names}}
svc-vps-manager.fpm
svc-vps-manager.mysql
svc-vps-manager.redis
```

Why I cannot use dashes in COMPOSE_PROJECT_NAME? 

There is no containers in `docker-compose ps`, because dashes were trimmed
```
docker-compose ps
```


shin-:
#4002

PR #5292, opened 2017-10-20
Validate if duplicate keys in yaml exist (fixes #5267)

### LINKED ISSUE #5267: Duplicate keys in docker-compose.yaml silently overwritten
If you have duplicate keys in your __docker-compose.yaml__ file, Docker Compose will silently overwrite the value of the first key. It might seem silly, but it is not hard to accidentally include duplicate keys when working with large files with many configuration options. (I just personally spent a long time debugging an issue which ultimately came down to duplicate `volumes` keys in my __docker-compose.yaml__ file which caused Elasticsearch config to not load correctly). It would be great if Docker Compose could catch problems like this early.

After perusing the code a little, it seems that this overwriting behavior comes from PyYAML, and there is a [related issue](https://github.com/yaml/pyyaml/issues/41) there. I still think it's a good idea to create an issue here for discussion.

One solution is to make a change in PyYAML, but there are other possible solutions. My Python knowledge is limited, but one user [appears to have created a Gist](https://gist.github.com/pypt/94d747fe5180851196eb) which can address this issue by modifying the main PyYAML class to raise an exception for duplicate keys. Alternatively, [someone on StackOverflow](https://stackoverflow.com/questions/44904290/getting-duplicate-keys-in-yaml-using-python) suggested a different Python package called [ruamel.yaml](https://pypi.python.org/pypi/ruamel.yaml/) which raises an exception for duplicate keys by default.

### Steps to Reproduce

I'll use the [hashicorp/http-echo](https://github.com/hashicorp/http-echo) image as the basis for a reproducible example.

Create a new directory and add a __docker-compose.yaml__ file with the following contents:

``` yaml
version: '3'
services:
  echo:
    image: hashicorp/http-echo
    ports: 
      - "5678:5678"
    command: "-text='Hello, world!'"
```

Start the service by running:

```
docker-compose up -d
```

Then send a request to the service through the exposed port:

```
curl http://localhost:5678
> Hello, world!
```

This works as expected. In order to see the failure case, modify __docker-compose.yaml__ to include a duplicate key for `ports`:

``` yaml
version: '3'
services:
  echo:
    image: hashicorp/http-echo
    ports: 
      - "5678:5678"
    command: "-text='Hello, world!'"
    ports:
      - "6789:6789"
```

If you try to send another request it will not work because port 5678 is no longer exposed:

```
curl http://localhost:5678
> curl: (7) Failed to connect to localhost port 5678: Connection refused
```

The first `ports` key in __docker-compose.yaml__ was silently overwritten by the second.

### OS/Platform Information

As far as I can tell this affects all versions of Docker Compose on all platforms, but I'll include my OS/Platform information just in case. I'm running on MacOS version 10.12.6. I'm using Docker for Mac, which [I installed on this page](https://store.docker.com/editions/community/docker-ce-desktop-mac).

Output of `docker-compose version`:

```
docker-compose version 1.16.1, build 6d1ac21
docker-py version: 2.5.1
CPython version: 2.7.12
OpenSSL version: OpenSSL 1.0.2j  26 Sep 2016
```

Output of `docker version`:

```
Client:
 Version:      17.09.0-ce
 API version:  1.32
 Go version:   go1.8.3
 Git commit:   afdb6d4
 Built:        Tue Sep 26 22:40:09 2017
 OS/Arch:      darwin/amd64

Server:
 Version:      17.09.0-ce
 API version:  1.32 (minimum version 1.12)
 Go version:   go1.8.3
 Git commit:   afdb6d4
 Built:        Tue Sep 26 22:45:38 2017
 OS/Arch:      linux/amd64
 Experimental: false
```

garribas:
Implemented using a python yaml linter (`yamllint`: http://yamllint.readthedocs.io/).

YAML files are linted before being parsed by PyYAML in function `load_yaml`, only the rule to detect duplicate keys is applied. If linting errors are found a `ConfigurationError` is raised.

Given a compose file with duplicates such as:
```
service:
  image: busybox:latest
    command: top
    command: duplicate

  environment:
    foo: bar
    foo: bar
```

Compose commands produce the following output:
```
$ docker-compose config
ERROR: The Compose file ./docker-compose.yml is invalid:
line 4, column 3: duplication of key "command" in mapping (key-duplicates)
line 8, column 5: duplication of key "foo" in mapping (key-duplicates)
```

### DIFF
--- a/compose/config/config.py
+++ b/compose/config/config.py
@@ -53,6 +53,7 @@ from .validation import validate_pid_mode
 from .validation import validate_service_constraints
 from .validation import validate_top_level_object
 from .validation import validate_ulimits
+from .validation import validate_yaml
 
 
 DOCKER_CONFIG_KEYS = [
@@ -1293,6 +1294,9 @@ def has_uppercase(name):
 def load_yaml(filename):
     try:
         with open(filename, 'r') as fh:
+            validate_yaml(fh)
+            # stream has been read by validation function, seek to beginning before parsing yaml
+            fh.seek(0)
             return yaml.safe_load(fh)
     except (IOError, yaml.YAMLError) as e:
         error_name = getattr(e, '__module__', '') + '.' + e.__class__.__name__
--- a/compose/config/validation.py
+++ b/compose/config/validation.py
@@ -13,6 +13,8 @@ from jsonschema import Draft4Validator
 from jsonschema import FormatChecker
 from jsonschema import RefResolver
 from jsonschema import ValidationError
+from yamllint import linter
+from yamllint.config import YamlLintConfig
 
 from ..const import COMPOSEFILE_V1 as V1
 from ..const import NANOCPUS_SCALE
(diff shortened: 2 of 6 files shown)


shin-:
To be honest, I'd be in favor of approaching this without adding more dependencies if at all possible. One solution we might want to explore first: https://gist.github.com/pypt/94d747fe5180851196eb

PR #5294, opened 2017-10-23
config_hash using oderdict to fix bug for sometimes no change services receated by compose using python3

KevinTHU:
Like the code diff, in python2 dict() is almost ordered in the same way, however in python3, dict() is almost randomly ordered, so we need to use oderdict to make sure the same key-value dict is saved in the same order, and so the hash of the service config 'config_hash' will be the same if the config is the same, however, now for python3 it is not（python2.7 is ok, like what mentioned above）, and this is what this PR fix for. Thanks for PTAL~

Signed-off-by: Xin He <hexin@jiedaibao.com>

### DIFF
--- a/compose/service.py
+++ b/compose/service.py
@@ -5,7 +5,7 @@ import logging
 import os
 import re
 import sys
-from collections import namedtuple
+from collections import namedtuple, OrderedDict
 from operator import attrgetter
 
 import enum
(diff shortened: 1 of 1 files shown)


shin-:
As you can see, [the `json_hash` function](https://github.com/docker/compose/blob/master/compose/utils.py#L103-L107) already sorts the keys, so the hash is stable no matter what.

KevinTHU:
@shin- sorry, I miss a condition that services need have more than two networks, this would cause these services frequently recreated by docker-compose, because the networks is not ordered by json_hash~， you can try it anytime~ easy to recurrent.

KevinTHU:
@shin- hi, for the new point, I have change the code, but because this PR is closed, the code is not updated, could you help to reopen it? I think this bug is some kind of import. PTAL~

shin-:
I checked with python 3, nested dictionaries are also ordered with the `sort_keys` option. 
```python
import json

d = {'b': 1, 'a': {'z': 2, 'x': 4, 'y': 3}}
json.dumps(x) # '{"b": 1, "a": {"x": 4, "z": 2, "y": 3}}'
json.dumps(x, sort_keys=True) # '{"a": {"x": 4, "y": 3, "z": 2}, "b": 1}'
```

Look, I believe that there may be an issue with hashes not being computed correctly, but maybe give me an example I can use to reproduce so we can diagnose this together?

KevinTHU:
@shin- ok, I will check it later, I am using python 3.4.5.

KevinTHU:
@shin- oh, I see, lilke my new PR code change, the disorder is not caused by json.dumps(), but cause by compose/network.py get_network_defs_for_service(service_dict), code below:
```
def get_network_defs_for_service(service_dict):
    if 'network_mode' in service_dict:
        return {}
    networks = service_dict.get('networks', {'default': None})
    return dict(
        (net, (config or {}))
        for net, config in networks.items()
    )
```
the dict() need change to be OrderedDict()
PTAL~

KevinTHU:
@shin- hi, excuse me, is there any comment on this PR？I think it is a problem for python3 compatibility.

PR #5300, opened 2017-10-23
Better installation instruction in release notes

reutsharabani:
Changed sample download script to use the built in `-o` optoin in `curl` instead of redicrecting stdout's output.

This allows users to prepend `sudo` to the snippet to make it work in common use cases where root permissions are needed to create the output file.

From `curl`:
    -o, --output <file>
              Write output to <file> instead of stdout.


reutsharabani:
@GordonTheTurtle not sure if comment is automagically generated but `git commit --amend -s --no-edit` is pretty neat, thanks :)

shin-:
Thanks!

PR #5305, opened 2017-10-24
CLI: Add --detach option for up, run and exec commands

pantuza:
This contribution allows the usage of --detach option when running:

docker-compose run --detach
docker-compose exec --detach
docker-compose up --detach

The behavior is the same as the -d option.

I think it would be one item from #363

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -405,7 +405,7 @@ class TopLevelCommand(object):
         Usage: exec [options] SERVICE COMMAND [ARGS...]
 
         Options:
-            -d                Detached mode: Run command in the background.
+            -d, --detach      Detached mode: Run command in the background.
             --privileged      Give extended privileges to the process.
             -u, --user USER   Run the command as this user.
             -T                Disable pseudo-tty allocation. By default `docker-compose exec`
@@ -415,7 +415,7 @@ class TopLevelCommand(object):
         """
         index = int(options.get('--index'))
         service = self.project.get_service(options['SERVICE'])
-        detach = options['-d']
+        detach = options.get('-d') or options.get('--detach')
 
         try:
             container = service.get_container(number=index)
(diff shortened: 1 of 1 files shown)


compose/cli/main.py
```diff
@@ -415,7 +415,7 @@ def exec_command(self, options):
         """
         index = int(options.get('--index'))
         service = self.project.get_service(options['SERVICE'])
-        detach = options['-d']
+        detach = options.get('-d') or options.get('--detach')
```
shin-: You can shorten this to `options.get('--detach')`, which will be set even if the user uses the short flag.

compose/cli/main.py
```diff
@@ -749,7 +749,7 @@ def run(self, options):
             -w, --workdir=""      Working directory inside the container
         """
         service = self.project.get_service(options['SERVICE'])
-        detach = options['-d']
+        detach = options.get('-d') or options.get('--detach')
```
shin-: ditto

compose/cli/main.py
```diff
@@ -928,14 +928,14 @@ def up(self, options):
         service_names = options['SERVICE']
         timeout = timeout_from_opts(options)
         remove_orphans = options['--remove-orphans']
-        detached = options.get('-d')
+        detached = options.get('-d') or options.get('--detach')
```
shin-: ditto

compose/cli/main.py
```diff
@@ -928,14 +928,14 @@ def up(self, options):
...
         no_start = options.get('--no-start')
 
         if detached and (cascade_stop or exit_value_from):
             raise UserError("--abort-on-container-exit and -d cannot be combined.")
 
         if no_start:
-            for excluded in ['-d', '--abort-on-container-exit', '--exit-code-from']:
+            for excluded in ['-d', '--detach', '--abort-on-container-exit', '--exit-code-from']:
```
shin-: Similarly here, we can remove `-d` from the list as it is redundant.

compose/cli/main.py
```diff
@@ -1177,7 +1177,7 @@ def run_one_off_container(container_options, project, service, options):
         one_off=True,
         **container_options)
 
-    if options['-d']:
+    if options.get('-d') or options.get('--detach'):
```
shin-: ditto

PR #5309, opened 2017-10-25
Add support for oom_kill_disable in service config

### LINKED ISSUE #2901: oom-kill-disable option not available
(Follow up to #754)

This might be somewhat awkward option in docker-compose.yml, but perhaps a pass-through argument for `docker-compose run`?

shin-:
Fixes #2901


Shinoby92:
got removed in 3.x but exist in 2.x 
could we add that please back to version 3.x? and I don't get it why even remove it again from version 3, after it got added in 2.x?

Issue #5310, opened 2017-10-25
docker-compose up --timeout does not appear to work


andyneff:
## Version

Docker version 17.09.0-ce, build afdb6d4
docker-compose version 1.17.0-rc1, build a0f95af

## Problem

Setting timeout on `docker-compose up` does not affect the shutdown timeout

`docker-compose.yml` file

```yaml
version: "3.3"
services:
  test2:
    image: debian:9
    command: nohup sleep 1000
```

(Also tried with version 2.0, and 2.3)

```
$ docker-compose up -d --timeout 123 test2
$ docker inspect issue_test2_1 # Shows no StopTimeout set at all
$ time docker-compose stop
Stopping issue_test2_1 ... done

real	0m11.473s
user	0m0.908s
sys	0m0.144s
```

(Also tried `down`)

## Expected behavior

```
$ docker run -it --rm --stop-timeout=121 --name test1 debian:9 nohup sleep 1000
$ docker inspect test1 # Show a StopTimeout of 121 set in the "Config" section
$ docker stop test1 # takes the ~121 seconds
```

## Other

Using `stop_grace_period` does work when using docker-compose commands, but not docker commands, as it appears that `docker-compose stop` or `down` just re-reads the `docker-compose.yml` file at stop time, and has nothing to do with how the container was started

## Objective

I am trying to set the StopTimeout in the actual container, so the daemon knows what to respect when it is shutdown (for example during a full system shutdown). Something like `stop_grace_period` doesn't help me, as that only tells `docker-compose` how long to wait, not the docker daemon.

shin-:
`--timeout` is intended to be used in interactive mode to set the timeout when issuing `SIGTERM`. This is in line with the option's help message: `Use this timeout in seconds for container shutdown when attached or when containers are already running. (default: 10)`, but I agree it could be more explicit.

If you want to apply a timeout to `docker-compose stop`, you'll have to use the `-t` flag with that command.

We should also error out if the user attempts to use `-d` and `-t` together.

andyneff:
@shin- I see now that I misunderstood what the `--timeout` was for, its for when you run `docker-compose up` and it has to shutdown the container first to start a new one, that's the timeout associated with that shutdown only (if I read the code correctly). I can see how the help message is saying that now, but yes, it could use a little clarifying.

[...]

PR #5314, opened 2017-10-26
Have stop_grace_period also set StopTimeout on create

andyneff:
Feature to handle #5310

### DIFF
--- a/compose/service.py
+++ b/compose/service.py
@@ -15,6 +15,7 @@ from docker.errors import ImageNotFound
 from docker.errors import NotFound
 from docker.types import LogConfig
 from docker.utils import version_lt
+from docker.utils import version_gte
 from docker.utils.ports import build_port_bindings
 from docker.utils.ports import split_port
 from docker.utils.utils import convert_tmpfs_mounts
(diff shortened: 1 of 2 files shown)


compose/service.py
```diff
@@ -15,6 +15,7 @@
 from docker.errors import NotFound
 from docker.types import LogConfig
 from docker.utils import version_lt
+from docker.utils import version_gte
```
shin-: Make sure those are ordered, right now you're failing the pre-commit checks.

compose/service.py
```diff
@@ -759,6 +760,11 @@ def _get_container_create_options(
             container_options['hostname'] = parts[0]
             container_options['domainname'] = parts[2]
 
+        if (version_gte(self.client.api_version, '1.25') and
+                'stop_grace_period' in self.options):
+            container_options['stop_timeout'] = parse_seconds_float(
```
shin-: use `self.stop_timeout(None)` instead

andyneff:
@shin- Changes made. Let me know if there's anything else I need to do.

shin-:
LGTM, thank you!

PR #5320, opened 2017-10-27
Improve performance by replacing call to pip freeze with import of do…

### LINKED ISSUE #5038: Improve performance by avoiding to call "pip freeze"
By calling `pip freeze`, in the main module, the actual code slow down every commands. In my case about 1 seconds par call, "just" to check if docker-compose is properly installed. I don't think it worst it.

Some tentative to improve this part had been made #5000 and #4698, but it still 1 second (in my case repeated tens time in my makefile)

Could we:
- remove this check and move it in a dedicated command `docker-compose debug`, or only on `docker-compose version/help` or when an exception is caugth.
- or add an env variable to skip this check

if would be happy to write the patch.

garribas:
…cker.version_info (fixes #5038)

Replaced call to pip freeze with a python one-liner that imports docker.version_info. Version 1.10.6 was the latest release of docker python client released as docker-py. This change results in executions of docker-compose being about 1.5x faster (affects all commands: help, info, etc.)

With call to pip freeze:

> $ time docker-compose
> 
> real	0m1.033s
> user	0m0.894s
> sys	0m0.130s

With call to import docker.version_info:

> $ time docker-compose
> 
> real	0m0.676s
> user	0m0.568s
> sys	0m0.085s

### DIFF
--- a/compose/cli/__init__.py
+++ b/compose/cli/__init__.py
@@ -2,43 +2,42 @@ from __future__ import absolute_import
 from __future__ import print_function
 from __future__ import unicode_literals
 
-import os
 import subprocess
 import sys
+from ast import literal_eval as make_tuple
 
 # Attempt to detect https://github.com/docker/compose/issues/4344
 try:
-    # We don't try importing pip because it messes with package imports
-    # on some Linux distros (Ubuntu, Fedora)
-    # https://github.com/docker/compose/issues/4425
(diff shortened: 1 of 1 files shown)


shin-:
Unfortunately, I don't think that will work reliably when both `docker` and `docker-py` are installed, which was the issue in the first place.

PR #3939, opened 2016-09-12
Add overwrite strategy options to config

CarstenHoyer:
This PR introduces two new configuration options; override_strategy and overwrite.

**override_strategy**: Can be either 'default' or 'overwrite'. Default does nothing except being explicit. 'Overwrite' overwrites the base config with the override config.

**overwrite**: If override_strategy is set, this does not apply. Otherwise it will look at all the fields in the overwrite array and again overwrite the base config with the override config.
- Relates to #2260
  https://github.com/docker/compose/issues/2260#issuecomment-246574209

Note there is currently no tests, I wanted to just see if there was any interest in this PR first.

PR #5354, opened 2017-11-08
Implement 'overwrite_mutlivals' config option

shcallaway:
In your `docker-compose.override.yml`, the `overwrite_multivals` config option allows you to specify which multivalue fields should be _overwritten_ instead of _concatenated with the base_. 

For example:

```
services:
  gateway:
    ports:
      - "80"
    overwrite_multivals:
      - ports
```

I am totally ripping off this guy's idea: #3939 

*This PR is a work in progress. Todo:*
* compose/config/validation.py
* Unit tests
* Documentation

### DIFF
--- a/compose/config/config.py
+++ b/compose/config/config.py
@@ -125,6 +125,7 @@ ALLOWED_KEYS = DOCKER_CONFIG_KEYS + [
     'network_mode',
     'init',
     'scale',
+    'overwrite_multivals'
 ]
 
 DOCKER_VALID_URL_PREFIXES = (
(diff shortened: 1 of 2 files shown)


shcallaway:
RFC: Do people care about/want a generalized `overwrite_multivals` option? Or do they only care about overwriting ports? A boolean option called `overwrite_ports` would be much simpler.

leonaves:
+1

drsharp:
@shcallaway I would vote for a generalized option... 

we have a use case where the `depends_on` section is different for some of us in development. For example, the base docker compose for a given service we have depends on two cache stores, one for worker jobs and one for key-value storage. In development, we just run a single Redis for both, but we need them separate in non-development environments. It would be nice to be able to override the `depends_on` section and rewrite it, but currently the overrides just append.

So consider this a vote in favor of something more general than just ports.

note, if I could "fish my wish" it would simply be that whatever key I put in an override would fully override that key, not append. That would mean if I have something like:

```yml
services:
  xyz:
    ports:
       - "443:9876"
    depends_on:
      - service-a
      - service-b
      - service-c
```

and I wanted it to also depend on `service-d` in my local development, then I'd have my override look like:

```yaml
services:
  xyz:
    depends_on:
      - service-a
      - service-b
      - service-c
      - service-d
```

Currently, to just add an extra service, I can just add that one (because it appends) but I would much prefer the flexibility of a per-section overwrite. For example, in the case I mentioned at the start, if the base uses `service-a`, `service-b`, and `service-c`, but I only want to depend on `service-a` and `service-b`, then a full per-key overwrite would allow me to set:

```yaml
services:
  xyz:
    depends_on:
      - service-a
      - service-b
```

and be good to go.

I know this may be a corner case, but the idea of only appending when it's a list just seems overly constrictive to em.

d0ruk:
Currently the only (besides the ```COMPOSE_FILE```) solution is;

 ```docker-compose -f docker-compose.yml -f docker-compose.prod.yml up```

The only difference between the dev and prod configs is some ```ports``` arrays. I can just do

 ```docker-compose -f docker-compose.yml up``` for prod, if this feature merged into upstream.

Anyone?

at0ms:
+1

shcallaway:
Closing because this particular design won't be approved. I may revisit this feature in the future.

PR #5364, opened 2017-11-10
Implement subnet config validation (fixes #4552)

### LINKED ISSUE #4552: config does not validate broken subnets
docker-compose 1.11.2

Broken config, due to missing variable - no error during `config`
```
networks:
  default:
    ipam:
      config:
      - subnet: ''
      driver: default
```

On `docker-compose up`

```
ERROR: Invalid subnet  : invalid CIDR address:
```

Should be thrown already during validation.

DrewRomanyk:
Add enhancement to config validation for subnet validation. This is my first Docker Compose PR, so please let me know if I missed anything!

ASSUMPTIONS:
* Only apply this to version 3.5
* Valid/Invalid as shown in the tests are correct for all subnets used by Docker Compose
* OS has support for socket.inet_pton() (As implemented by http://python-jsonschema.readthedocs.io/en/latest/validate/#validating-formats)

EDIT:
It seems that I am failing the Windows py27, as it appears to not support socket.inet_pton(). Should I make the solution not have to support socket.inet_pton(), but rather a very long regex to support all notations of ipv6 such as (https://www.regexpal.com/93988)?

### DIFF
--- a/compose/config/validation.py
+++ b/compose/config/validation.py
@@ -72,22 +72,24 @@ def format_expose(instance):
 def format_subnet_ip_address(instance):
     if isinstance(instance, six.string_types):
         if '/' not in instance:
-            raise ValidationError("should be of the format 'IP_ADDRESS/CIDR'")
+            raise ValidationError("'{0}' 75 should be of the format 'IP_ADDRESS/CIDR'".format(instance))
 
         ip_address, cidr = instance.split('/')
 
         if re.match(VALID_IPV4_FORMAT, ip_address):
             if not (re.match(VALID_IPV4_CIDR_FORMAT, cidr) and
                     all(0 <= int(component) <= 255 for component in ip_address.split("."))):
-                raise ValidationError("should be of the format 'IP_ADDRESS/CIDR'")
+                raise ValidationError(
+                    "'{0}' 83 should be of the format 'IP_ADDRESS/CIDR'".format(instance))
         elif re.match(VALID_IPV6_CIDR_FORMAT, cidr) and hasattr(socket, "inet_pton"):
             try:
                 if not (socket.inet_pton(socket.AF_INET6, ip_address)):
-                    raise ValidationError("should be of the format 'IP_ADDRESS/CIDR'")
+                    raise ValidationError(
+                        "'{0}' 88 should be of the format 'IP_ADDRESS/CIDR'".format(instance))
             except socket.error as e:
                 raise ValidationError(six.text_type(e))
         else:
-            raise ValidationError("should be of the format 'IP_ADDRESS/CIDR'")
+            raise ValidationError("'{0}' 92 should be of the format 'IP_ADDRESS/CIDR'".format(instance))
 
     return True
 
--- a/compose/config/validation.py
+++ b/compose/config/validation.py
@@ -5,6 +5,7 @@ import json
 import logging
 import os
 import re
+import socket
 import sys
 
 import six
@@ -43,6 +44,9 @@ DOCKER_CONFIG_HINTS = {
 
 VALID_NAME_CHARS = '[a-zA-Z0-9\._\-]'
 VALID_EXPOSE_FORMAT = r'^\d+(\-\d+)?(\/[a-zA-Z]+)?$'
+VALID_IPV4_FORMAT = r'^(\d{1,3}.){3}\d{1,3}$'
+VALID_IPV4_CIDR_FORMAT = r'^(\d|[1-2]\d|3[0-2])$'
+VALID_IPV6_CIDR_FORMAT = r'^(\d|[1-9]\d|1[0-1]\d|12[0-8])$'
 
 
 @FormatChecker.cls_checks(format="ports", raises=ValidationError)
@@ -60,6 +64,30 @@ def format_expose(instance):
         if not re.match(VALID_EXPOSE_FORMAT, instance):
             raise ValidationError(
                 "should be of the format 'PORT[/PROTOCOL]'")
+
+    return True
+
+
+@FormatChecker.cls_checks("subnet_ip_address", raises=ValidationError)
+def format_subnet_ip_address(instance):
+    if isinstance(instance, six.string_types):
+        if '/' not in instance:
+            raise ValidationError("should be of the format 'IP_ADDRESS/CIDR'")
+
+        ip_address, cidr = instance.split('/')
+
+        if re.match(VALID_IPV4_FORMAT, ip_address):
+            if not (re.match(VALID_IPV4_CIDR_FORMAT, cidr) and
+                    all(0 <= int(component) <= 255 for component in ip_address.split("."))):
+                raise ValidationError("should be of the format 'IP_ADDRESS/CIDR'")
+        elif re.match(VALID_IPV6_CIDR_FORMAT, cidr) and hasattr(socket, "inet_pton"):
+            try:
+                if not (socket.inet_pton(socket.AF_INET6, ip_address)):
+                    raise ValidationError("should be of the format 'IP_ADDRESS/CIDR'")
+            except socket.error as e:
+                raise ValidationError(six.text_type(e))
+        else:
+            raise ValidationError("should be of the format 'IP_ADDRESS/CIDR'")
 
     return True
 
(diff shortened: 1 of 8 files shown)


shin-:
Hi Drew!

I think you're off to a good start, but the test still seems to be failing. You can run them locally first using the `tox` command (you might need to install it with `pip install tox`). If you're debugging and only want to run a few tests at a time, you can use [`pytest`](https://docs.pytest.org/en/latest/usage.html) directly as well.

On the design end:
- Validation should be added to all versions that support the `subnet` configuration, i.e. 3.0 - 3.5
- You indeed can not rely on UNIX-only functions/libraries. A regex is a sensible solution here.

HTH!

DrewRomanyk:
Hi Joffrey!

Thanks for the feedback and for the advice! I did test using the script/test/default method, however it passed those tests on my mac.

As for the design bits, I added the validation for all the v3 config versions and removed that UNIX dependency. However, I found the regex for the IPV6 would be very huge, and decided to use the format string option to make it more readable, if however, you would prefer just one huge regex string, I can easily do that! Also for the subnet config, is CIRD required?

compose/config/validation.py
```diff
@@ -64,6 +90,26 @@ def format_expose(instance):
...
+@FormatChecker.cls_checks("subnet_ip_address", raises=ValidationError)
+def format_subnet_ip_address(instance):
+    if isinstance(instance, six.string_types):
+        if '/' not in instance:
+            raise ValidationError("should be of the format 'IP_ADDRESS/CIDR'")
+
+        ip_address, cidr = instance.split('/')
+
+        if re.match(VALID_REGEX_IPV4_ADDR, ip_address):
+            if not re.match(VALID_REGEX_IPV4_CIDR, cidr):
```
shin-: I don't think we need a regexp for this part - a simple int conversion should suffice.

compose/config/validation.py
```diff
@@ -64,6 +90,26 @@ def format_expose(instance):
...
+                raise ValidationError("should be of the format 'IP_ADDRESS/CIDR'")
```
shin-: Conventionally, the entire string is the CIDR format. The part after the slash is usually called the prefix. In our case, I think it's fine to say `"should use the CIDR format"` for this error message.

compose/config/validation.py
```diff
@@ -64,6 +90,26 @@ def format_expose(instance):
     return True
 
 
+@FormatChecker.cls_checks("subnet_ip_address", raises=ValidationError)
+def format_subnet_ip_address(instance):
+    if isinstance(instance, six.string_types):
```
shin-: It's unlikely this will ever be fed something that isn't a string, but if it does, this will break when you try to `split()` a few lines later.
I'd remove the type check altogether.

compose/config/validation.py
```diff
@@ -64,6 +90,26 @@ def format_expose(instance):
...
+        if '/' not in instance:
+            raise ValidationError("should be of the format 'IP_ADDRESS/CIDR'")
+
+        ip_address, cidr = instance.split('/')
```
shin-:
```
>>> format_subnet_ip_address('a/b/c')
ValueError: too many values to unpack
>>>
```

Since we never want more than 2 parts, we should do `instance.split('/', 1)` here.

compose/config/validation.py
shin-: In the same vein, you might want to update the variable and constant names to match.

tests/unit/config/config_test.py
```diff
@@ -2846,6 +2846,93 @@ def check_config(self, cfg):
...
+        "192.168.0.1/",
+        "192.168.0.1/33",
+        "192.168.0.1/01",
+        "192.168.0.1",
+        "fe80:0000:0000:0000:0204:61ff:fe9d:f156/sdfsdfs",
+        "fe80:0000:0000:0000:0204:61ff:fe9d:f156/",
+        "fe80:0000:0000:0000:0204:61ff:fe9d:f156/129",
+        "fe80:0000:0000:0000:0204:61ff:fe9d:f156/01",
+        "fe80:0000:0000:0000:0204:61ff:fe9d:f156",
+        "ge80:0000:0000:0000:0204:61ff:fe9d:f156/128",
```
shin-: Add a test for more than 1 slash as well!

compose/config/validation.py
DrewRomanyk: Would it be okay if I would just include the CIDR prefix check within the regex pattern then, as all CIDRs would require that prefix?

tests/unit/config/config_test.py
DrewRomanyk: Will do, good idea!

compose/config/validation.py
shin-: That would work too.

compose/config/validation.py
DrewRomanyk: Actually, it seems that this check is common for the format checker methods in order for it to go to the type checker. Otherwise, in the old code, it will fail at the '/' check or in the new code where I just use regex, at the re.match code.

So would you prefer me to use a try except to handle the errors, or just use already existing pattern used by FormatChecker ipv4/ipv6 checkers, and also the Docker Compose expose checker?

shin-: The other 2 `format_*` methods can take integers as well, which is why they check the type. Since we enforce `type: string` on subnet entries, it will exit before ever hitting this method if the type is invalid:
```
$ cat docker-compose.yml 
version: '3.5'
services:
  web:
    image: busybox
networks:
  default:
    ipam:
      config:
        - 'subnet': 1234
      driver: 'default'
$ docker-compose config
networks.default.ipam.config.subnet contains an invalid type, it should be a string
```

DrewRomanyk: Hey Joffrey!

If you are new format code with those commands then maybe our environments are different, otherwise, it appears that at least on my end it is running the format checker before running the type checker. When I remove the isinstance check and run those two same commands with the same docker compose file:
```
$ cat docker-compose.yaml 
version: '3.5'
services:
  web:
    image: busybox
networks:
  default:
    ipam:
      config:
        - 'subnet': 1234
      driver: 'default'
$ docker-compose config
...
...
...
compose/config/validation.py", line 94, in format_subnet_ip_address
    if not re.match(VALID_REGEX_IPV4_CIDR, instance) and \
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/re.py", line 141, in match
    return _compile(pattern, flags).match(string)
TypeError: expected string or buffer
```

And when I modify the 3.5 config json schema file, to where subnet requires a number type, the type checker still doesn't complain when I give it a boolean. And when I run the type:number/format:subnet with a valid subnet, it fails at the type checker, but goes through the format checker (I tested by a simple print statement). Could our Json schema libraries be out of sync then?

So then, in this case, could it be:
* Are our environments not in sync? Aka: I messed something up with my environment to where Format is running before Type check?
* I messed up setting the Json schema up somewhere?
* Is the solution with isinstance safer?

If I go inside the jsonschema._format.py code, here is a few examples of what I've seen, which they all seem to have that isinstance check even for string only types such as emails
```
@_checks_drafts("email")
def is_email(instance):
    if not isinstance(instance, str_types):
        return True
    return "@" in instance
@_checks_drafts(draft3="ip-address", draft4="ipv4")
def is_ipv4(instance):
    if not isinstance(instance, str_types):
        return True
    if not _ipv4_re.match(instance):
        return False
    return all(0 <= int(component) <= 255 for component in instance.split("."))
@_checks_drafts("regex", raises=re.error)
def is_regex(instance):
    if not isinstance(instance, str_types):
        return True
    return re.compile(instance)
@_checks_drafts(draft3="date", raises=ValueError)
def is_date(instance):
    if not isinstance(instance, str_types):
        return True
    return datetime.datetime.strptime(instance, "%Y-%m-%d")
@_checks_drafts(draft3="time", raises=ValueError)
def is_time(instance):
    if not isinstance(instance, str_types):
        return True
    return datetime.datetime.strptime(instance, "%H:%M:%S")
```

Thank you again for all the assistance/feedback Joffrey, I really appreciate it!!

shin-: Interesting, I'll check out your branch and take a closer look. I'll let you know.

shin-: Yup, you were right, my bad. I expected the type constraint to be enforced before format validation.

shin-:
LGTM, thank you!

PR #5370, opened 2017-11-11
Fix overriding volumes from cli when running a service

eddwardo:
Signed-off-by: Piotr Szymanski <p.szymanski@bidpy.pl>

In case where volumes are defined for a service in compose yaml file,
user might want to override this volume from cli command as such:
docker-compose run -v local:/overrided_remote foo

Currently, if user tries to override the volume from cli, command is
ignored.

Bug has been introduced in 1.14.0 afaik.


eddwardo:
Locally, this test passes.

shin-:
Can you clarify what bug this is supposed to fix?

Also, tests do need to pass on our CI. In this case, the issue is that you're not creating the mounted file on the remote host. Look at `test_run_one_off_with_volume_merge` for how to achieve this.

eddwardo:
@shin- Update the commit message:
```
In case where volumes are defined for a service in compose yaml file,
user might want to override this volume from cli command as such:
docker-compose run -v local:/overrided_remote foo

Currently, if user tries to override the volume from cli, command is
ignored.
```

eddwardo:
When can this land?

eddwardo:
Ping?

eddwardo:
This blocks my company from upgrading from 1.14.0 to the latest.
How can I get this PR to land?

eddwardo:
fixed

PR #5378, opened 2017-11-15
Allow x-project-name property in Compose file to define project name

### LINKED ISSUE #745: Proposal: make project-name persistent.

shin-:
Fixes #745 

Notes:
- This is not meant as an official addition to the Compose file format. Other implementations may or may not offer similar functionality.
- The order of precedence is as follows: `--project-name` CLI option > `COMPOSE_PROJECT_NAME` environment variable > `x-project-name` value (iff `COMPOSE_X_PROJECT_NAME` is enabled) > directory name

### DIFF
--- a/compose/cli/command.py
+++ b/compose/cli/command.py
@@ -85,9 +85,7 @@ def get_project(project_dir, config_path=None, project_name=None, verbose=False,
     if not environment:
         environment = Environment.from_env_file(project_dir)
     config_details = config.find(project_dir, config_path, environment, override_dir)
-    project_name = get_project_name(
-        config_details.working_dir, project_name, environment
-    )
+    project_name = get_project_name(config_details, project_name, environment)
     config_data = config.load(config_details)
 
     api_version = environment.get(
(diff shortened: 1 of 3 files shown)


thedeeno:
For my use case this is similar enough to using `COMPOSE_PROJECT_NAME` that I wouldn't get any value from these changes. Couple that with the magic `_x_` prefixes and this just seems too complex to merge imo.

adminliberty:
I would love to see this get added to Docker Compose. Is there any update on this?

jpringle03:
I would love to see this get added to Docker Compose. Is there any update on this?

shin-:
The feedback we received on this issue led us to believe this wasn't the right approach to tackling the problem. We're still thinking about ways to do this, but I think it's safe to say this isn't it (at least not in isolation)

jpringle03:
Thanks for the quick response @shin- 

I'd love to eventually see a way to define the project name within the compose file. Is there conversation happening anywhere else about how to address this?

ijc:
I guess this should have been closed with https://github.com/docker/compose/pull/5378#issuecomment-373142575, so I will.

Issue #1498, opened 2015-06-02
Make shell completion DRYer and less hacky


aanand:
Thanks to @albers and @sdurrheimer, we now have both bash and zsh completion, which is fantastic. However, they (understandably) both do a lot of hacky stuff to read the YAML file without actually parsing YAML properly, and they both implement this separately.

We could potentially add some commands/flags to Compose that would give them the information they need in a simple shell-parseable format. Going off a quick scan of the [bash script](https://github.com/docker/compose/blob/master/contrib/completion/bash/docker-compose), it looks like it'd be useful to be able to list:
- [x] names of all services in the file
- [ ] names of all services that have a given configuration option set - here, we'd even be able to evaluate `extends` to get the correct configuration, which I doubt would be any fun at all to implement in shell script
- [ ] names of services that have at least one running container
- [ ] names of services that have at least one stopped container

Perhaps a `docker-compose services` command:

```
$ docker-compose services
web
redis

$ docker-compose services --with-option="build"
web

$ docker-compose services --with-running-container
redis
```

albers:
Wow! That's fantastic. I'm happy to see completion growing in weight (I didn't dare to ask about such features before).

Another difficult to implement feature that this proposal solves (taken from the releasenotes):

> When invoked in a subdirectory of a project, `docker-compose` will now climb up through parent directories until it finds a `docker-compose.yml`.

thaJeztah:
Nice thinking!

```bash
$ docker-compose services --with-option="build"
$ docker-compose services --with-running-container
```

How about using a `--filter ....` flag for this, which is a closer match to the way this works in "Docker like", e.g.

```bash
$ docker-compose services --filter option=build
$ docker-compose services --filter status=running
```

(Going off-topic alert);
To be useful for other purposes, `docker-compose services` could output all information of the services by default (similar to `docker inspect`), and feature a `--format ..` flag that accepts a Go-template. However, I realize python !== golang, so that would involve implementing go-templates in Python (which is probably a bad idea, unless someone did that already)

tpounds:
+1 to provide a `--filter` flag since it's consistent with many docker list resource commands.

albers:
+1 for the `--filter` syntax.

sdurrheimer:
`--filter` sounds good :+1:

albers:
With the addition of multiple compose files (#2051), and the recently introduced search of compose.yml in parent directories, this feature is gaining weight.

[...]

PR #5384, opened 2017-11-18
Implement --filter flag for docker-compose config --services and use it in bash completion

ilinum:
# The Flag
This PR add `--filter` flag to `docker-compose config --services` in the style of `docker ps --filter`.

`--filter` takes one of the following arguments:
* `option` (can be of value `build` or `image`) filters out all services that don't have the corresponding option.
* `status` (takes values `running`, `paused`, or `stopped`) filters out all services that don't have at least one container that has the provided status.

Example: `docker-compose config --services --filter "status=running" --filter "option=build"`.

# Completion

The goal of this PR is to make completion scripts less hacky. Now, instead of manually parsing the config file, bash completion calls the `docker-compose config` command with appropriate filter arguments.

Note that this PR does not touch zsh completion because I am not familiar with zsh. It's probably good if somebody else implements it in a separate PR.


ilinum:
@albers git blame tells me you implemented most of the bash completion :) Do you mind looking at this PR?

compose/cli/main.py
```diff
@@ -287,14 +287,15 @@ def config(self, config_options, options):
...
-        Usage: config [options]
+        Usage: config [options] [-f KEY=VAL...]
 
         Options:
             --resolve-image-digests  Pin image tags to digests.
             -q, --quiet              Only validate the configuration, don't print
                                      anything.
             --services               Print the service names, one per line.
             --volumes                Print the volume names, one per line.
+            -f, --filter KEY=VAL     Filter services by a property (can be used multiple times)
```
ilinum: This probably deserves better documentation but I wasn't sure how to format it.

In particular, it would be good to provide values that filter takes.

albers:
I could not build docker-compose on Ubuntu 16.04. I tried to follow the instructions in https://github.com/docker/compose/blob/master/CONTRIBUTING.md but got
```bash
$ python setup.py develop
  File "setup.py", line 12, in <module>
    import pkg_resources
ImportError: No module named pkg_resources
```
Is there a way to just build the binary? I'm not familiar with python.

albers:
I think the `option` filter has a confusing name. What is an _option_? The only allowed values are `build` and `image`, but are these _options_?
I propose to extend the scope of this filter to be more generic. It could accept arbitrary keys and optionally values, e.g.
```bash
# all services with a "image" key
docker-compose config --services --filter key=image
# all services with "image: busybox:latest"
docker-compose config --services --filter key=image=busybox:latest
# all services with a label
docker-compose config --services --filter key=label
# all services with a "some-label" label
docker-compose config --services --filter key=label=some-label
# all services with a "some-label" label of value "some-value"
docker-compose config --services --filter key=label=some-label=some-value
```

albers:
Cross-linking
#1498 Make shell completion DRYer and less hacky
#1953 Feature Request: `docker-compose config` Command

shin-:
Hmm... I don't really see this as something we'll want to support in the long term, to be honest. There are a lot of intricacies involving nested properties, complex values, etc. that are going to take a lot of time to support down the line.

All in all, if this is a necessity for you, I'd recommend making it a separate project.

albers:
@shin- The extended syntax is not a requirement for me. My use case (#1498) is perfectly covered by this PR at its current state. I just felt that `opt=image` and `opt=build` are not self-explanatory, and I could not come up with better names. So I got the idea of a more generic solution. If this means too much work, I withdraw my proposal.

How about renaming to `opt=from-image` and `opt=from-build`? This is more intuitive and still leaves space for potential further extensions.

ilinum:
Thanks for looking at it!
@albers yes, changing `option` to `key` sounds like a good idea. 
Regarding your issue about building docker, [this stackoverflow post](https://stackoverflow.com/questions/7446187/no-module-named-pkg-resources) might help you (but I have not encountered this error).

shin-:
Oh, sorry - I didn't read this correctly the first time around. So this is intended to improve shell completion as per #1498. I'll have to take a closer look, but right off the bat, it seems that the `status` filter belongs on the `ps` subcommand rather than `config` (though the form would likely be different). Ideally, I would rather not have the `config` subcommand ever connect to a daemon.

@albers Since you are on Linux, you should be able to build the binary from source by running `./script/build/linux` with just a Docker installation.

ilinum:
Has anyone had a chance to look at this?

The problem with adding using `ps` is that `ps` outputs container hashes and completion needs services.

albers:
@shin- brought up a very good point. As its name clearly states, the `config` command is about configuration. The configuration is solely expressed by the compose files. It feels strange that a filter activates additional behaviour so that state (not part of the compose files) is also considered.

The `ps` command looks like the natural place to harvest information about running states, but it works on container level, not service level (what we need here). And it only considers existing containers.

A new top level command could help here:
```bash
docker-compose services --filter state=running
```
It would do the same as `docker-compose config --services` but offer service level filtering as implemented in this PR.

WDYT? Is such a command of general interest? Would the current use case justify a new command?
I always found it ab bit strange that `docker-compose` ís all about handling services but the only command for querying state lists containers.

shin-:
I'd rather keep the list of commands short if we can help it. Especially if we consider the very limited usefulness something like `docker-compose services` would have for the majority of users.

On the other hand, I think augmenting `ps` with a flag that lists services rather than containers and is fairly intuitive (we're just going up one level of abstraction) while preserving the non-interactive nature of the `config` subcommand.

ilinum:
That would make sense! I haven't used `docker-compose` very much but I feel like it could be useful. 

Alternatively, we could make `docker-compose ps` more service-oriented.

ilinum:
@shin- didn't see your comment until I already sent mine.

Would it make sense to allow filtering services in `ps` also by `build` and `image` or would that be better kept in `config`?

albers:
>  I think augmenting ps with a flag that lists services rather than containers makes sense and is fairly intuitive (we're just going up one level of abstraction) while preserving the non-interactive nature of the config subcommand.

That's a great idea.
What would the output look like? Given the following `ps` output
```bash
$ docker-compose ps
   Name            Command          State            Ports
------------------------------------------------------------------
hugo_web_1   nginx -g daemon off;   Up       0.0.0.0:32768->80/tcp
hugo_web_2   nginx -g daemon off;   Paused   0.0.0.0:32769->80/tcp
```
State and Ports are container scoped, so just showing the service scoped fields we would get
```bash
$ docker-compose ps --services
   Name          Command
-------------------------------
hugo_web   nginx -g daemon off;
```

albers:
@shin- 
> Since you are on Linux, you should be able to build the binary from source by running `./script/build/linux ` with just a Docker installation.

Thanks, that was exactly what I was looking for.

ilinum:
@albers @shin- the feedback should be addressed now.

Currently `docker-compose ps --services` just prints out the services (same as `docker-compose config --services`:

```
$ docker-compose ps --services
with_image
with_build
$ docker-compose ps --services --filter "key=build"
with_build

```

ilinum:
@albers @shin- any feedback?

shin-:
I'm focused on the things that need to get merged for the 1.18 release right now, which means this is lower priority for the next few days. I'll review this thoroughly as soon as I have a chance - thank you for your patience!

albers:
Thanks for the quick rework. @ilinum.
I think the filters and the `--services` mode fit very well into `docker-compose ps`. They work as expected for `--services` mode.

I still got two suggestions:
* The filters should also work in plain `ps` mode: `docker-compse ps --filter status=running` should only list running containers. Same for the `build|image` filter.
* `key` is not a good name. It suggests that any key from the compose file can be used here, which is not true. There is also a name collision with the help message: It states that filters have a KEY=VAL syntax. It is confusing that KEY is either `key` or `status`. I prefer `source=image|build` instead of `key`. 

Let's wait for @shin-'s review.

compose/cli/main.py
```diff
@@ -599,38 +599,47 @@ def ps(self, options):
...
-            key=attrgetter('name'))
-
-        if options['-q']:
-            for container in containers:
-                print(container.id)
+        if options['--services']:
+            filters = build_filters(options.get('--filter'))
+            services = self.project.services
+            if filters:
+                services = filter_services(filters, services, self.project)
```
shin-: Could we filter the containers directly, then iff the `--services` flag is provided, generate a `set()` of services from the filtered container list and print that out? That way, `--filter` could be used with both "modes".

compose/cli/main.py
```diff
@@ -599,38 +599,47 @@ def ps(self, options):
...
+            print('\n'.join(service.name for service in services))
```
shin-: We could just `return` after this line and avoid the additional indentation in the rest of the function.

tests/acceptance/cli_test.py
```diff
@@ -467,6 +467,35 @@ def test_ps_alternate_composefile(self):
         self.assertNotIn('multiplecomposefiles_another_1', result.stdout)
         self.assertIn('multiplecomposefiles_yetanother_1', result.stdout)
 
+    def test_ps_services_filter_option(self):
+        self.base_dir = 'tests/fixtures/ps-services-filter'
+        image = self.dispatch(['ps', '--services', '--filter', 'key=image'])
+        build = self.dispatch(['ps', '--services', '--filter', 'key=build'])
+        all_services = self.dispatch(['ps', '--services'])
+
+        self.assertIn('with_build', all_services.stdout)
```
shin-: Please use the more readable format `assert 'with_build' in all_services.stdout` for newly added tests!

compose/cli/main.py
```diff
@@ -1312,3 +1321,51 @@ def build_exec_command(options, container_id, command):
     args += [container_id]
     args += command
     return args
+
+
+def has_container_with_state(containers, state):
+    for container in containers:
+        states = {
```
shin-: I feel we can improve this by taking the instantiation outside of the loop and using lambdas, i.e.
```python
states = {
  'running': lambda c: c.is_running,
  'stopped': lambda c: not c.is_running,
 ...
}
if state not in states:
  raise UserError(...)
for container in containers:
  if states[state](container):
    return True
return False
```

shin-: Also, it should be trivial to support `restarting` as a state filter as well.

compose/cli/main.py
```diff
@@ -1312,3 +1321,51 @@ def build_exec_command(options, container_id, command):
...
+                for status in filters[f]:
+                    if not has_container_with_state(containers, status):
+                        return False
+            elif f == 'key':
+                for key in filters[f]:
+                    if key == 'image' or key == 'build':
+                        if key not in service.options:
+                            return False
+                    else:
+                        raise UserError("Invalid option: %s" % key)
```
shin-: Maybe `"Invalid value for key filter: %s"`  would be a more useful error message here?

compose/cli/main.py
```diff
@@ -1312,3 +1321,51 @@ def build_exec_command(options, container_id, command):
...
+            raise UserError("Invalid state: %s" % state)
+        if states[state]:
+            return True
+    return False
+
+
+def filter_services(filters, services, project):
+    def should_include(service):
+        for f in filters:
+            if f == 'status':
```
shin-: rather than `status`, this should be called `state` to match the terminology used elsewhere in the application and the Docker ecosystem.

compose/cli/main.py
```diff
@@ -1312,3 +1321,51 @@ def build_exec_command(options, container_id, command):
...
+
+def build_filters(args):
+    filters = {}
+    for arg in args:
+        if '=' not in arg:
+            raise UserError("Arguments to --filter should be in form KEY=VAL")
+        key, val = arg.split('=', 1)
+        if key not in filters:
+            filters[key] = []
+        filters[key].append(val)
```
shin-: Is there a use-case for having multiple values for the same filter? In other words, when would a user possibly need to do something like `docker-compose ps --filter state=paused --filter state=running`?

If there's nothing compelling, we could simplify the code a lot by allowing only one value per filter.

compose/cli/main.py
albers: Actually, the corresponding filter in `docker ps` is called `status`, too: https://docs.docker.com/engine/reference/commandline/ps/#filtering

compose/cli/main.py
albers: `docker ps supports` multiple filters of the same kind, so it might be worth implementing it here for symmetry reasons, too. But I'm OK with a simpler solution. In that case, there should be an error message for duplicate filter usage.

compose/cli/main.py
ilinum: This would not work as well. The reason is that `self.project.services` reads the services from the compose file while `self.project.containers` would omit containers that have never been started before.

compose/cli/main.py
ilinum: Yep! Addressed.

compose/cli/main.py
ilinum: Done for both!

compose/cli/main.py
ilinum: Done!

compose/cli/main.py
ilinum: Simplified the code, now you can only have one `--filter` argument.

tests/acceptance/cli_test.py
ilinum: Done!

ilinum:
Sorry for taking a while to respond! Most of the feedback should be addressed now.

shin-:
No worries, I assumed the holiday season kept you busy 👍 

Code LGTM now - having filters apply to container lists as well would be nice to have, but that can be done later. The last remaining issue is the naming, as @albers pointed out, of the `key` filter which is very non-descriptive and potentially confusing. I'm fine with using `source` instead. WDYT?

ilinum:
Yep, didn't see that comment earlier!
Renamed to `source`!

shin-:
Thanks @ilinum ! 💯

PR #5393, opened 2017-11-22
implement --timeout flag for docker-compose down

### LINKED ISSUE #3370: [Request] - Implement -t | --timeout option for `docker-compose down
There is the `-t | --timeout` option that can be passed to `docker-compose stop`, it would be nice to add this option to `docker-compose down`.

As `down` destroys a containers, it means we won't be using it again, so we should be allowed "not to care if it exits correctly. Alternatively it could be a `-k | --kill` flag so that instead of "stopping and removing container" it would "kill and remove container".

I understand that it should not be the default behavior because some linked containers can still prefer if the present container is stopped correctly, but there are such cases where it could speed things up.

For instance, I do a lot of debugging on an application that _unfortunately_ doesn't handle `SIGTERM` signal, and it is a pain to wait 10 seconds after running `down`, so I usually `kill` it first, then `down` it.

It's just a suggestion, what do you think?

summergirl21:
This PR implements  `--timeout` flag for `docker-compose down` similar to
`docker-compose stop --timeout`, that specifies a shutdown timeout in seconds.

Fix #3370

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -371,9 +371,12 @@ class TopLevelCommand(object):
                                 attached to containers.
             --remove-orphans    Remove containers for services not defined in the
                                 Compose file
+            -t, --timeout TIMEOUT      Specify a shutdown timeout in seconds.
+                                     (default: 10)
         """
         image_type = image_type_from_opt('--rmi', options['--rmi'])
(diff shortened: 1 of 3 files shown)


summergirl21:
Can someone please review this PR?

summergirl21:
@shin- Hoping to get this change merged, any feedback would be greatly appreciated!

PR #5403, opened 2017-11-26
[fix] scale option in docker compose file

cuckoohello:
Currently scale option not works in docker-compose.yml for version >=3, but the code supports this option. It's not working just because config schema missing scale option.

```
version: '3'
services:
  redis:
    image: redis:alpine
    restart: always
    scale: 3
    ports:
      - 6379:6379
```


shin-:
That is intentional. `scale` is meant to be a v2 option only.

Issue #5374, opened 2017-11-13
Please how I can limit memory in docker-compose build


Rikhart:
¿How I can limit memory in "docker-compose build" command?
I need to setup a limit for building because my tiny droplet is out of memory when run the building process.

Thanks for your answers

shin-:
It's not currently possible with the `docker-compose build` command. Consider building with `docker build` using the `-m` option for the time being.

n0tel:
The same question. Any possibility to limit `docker-compose build` ?

PR #5410, opened 2017-11-28
Adds support for a memory build-arg in docker-compose build.

smiller123:
Implements #5374. You can now use memory as a build-arg to docker-compose build.

Signed-off-by: Samantha Miller <samantha.a.miller123@gmail.com>

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -232,6 +232,7 @@ class TopLevelCommand(object):
             --force-rm              Always remove intermediate containers.
             --no-cache              Do not use cache when building the image.
             --pull                  Always attempt to pull a newer version of the image.
+            -m, --memory MEM        Sets memory limit for the bulid container.
             --build-arg key=val     Set build-time variables for one service.
         """
         service_names = options['SERVICE']
@@ -248,6 +249,7 @@ class TopLevelCommand(object):
             no_cache=bool(options.get('--no-cache', False)),
             pull=bool(options.get('--pull', False)),
             force_rm=bool(options.get('--force-rm', False)),
+            memory=options.get('--memory'),
             build_args=build_args)
 
     def bundle(self, config_options, options):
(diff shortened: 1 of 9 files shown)


shin-:
Thanks!

I think this would actually be more useful as a CLI option rather than a static configuration entry, since the memory one is willing to allocate to a build will mostly depend on the capacity of the machine it's executed on. WDYT?

smiller123:
That definitely sounds right. It's what I intended to do, it just took me a while to get my head wrapped around everything. Is this better?

compose/config/config.py
```diff
@@ -1020,6 +1020,7 @@ def to_dict(service):
     md.merge_scalar('network')
     md.merge_scalar('target')
     md.merge_scalar('shm_size')
+    md.merge_scalar('memory')
```
shin-: Is this still needed now that we removed the option from the schemas?

compose/cli/main.py
```diff
@@ -232,6 +232,7 @@ def build(self, options):
             --force-rm              Always remove intermediate containers.
             --no-cache              Do not use cache when building the image.
             --pull                  Always attempt to pull a newer version of the image.
+            --memory MEM            Sets memory limit for the bulid container.
```
shin-: Can we add the short form `-m` as well?

smiller123:
Done. Thanks!

arnoldelena:
Is this issue ready to be closed? I see that it was merged.

wuboy0307:
I think an entry to specify memory in docker-composer.yml is still very useful. Consider you can always set memory limit in .env file and pass into the yml. The entry now is not a static configuration, it's from an environment variable.

It's very useful when I have one special container that need large memory. I can set it in .env and pass it to yml. Anyone who also maintain the docker composer project won't need to remember to add --memory to the special large-memory image every time.

PR #5411, opened 2017-11-28
Fix depends on restart behavior, fixes #3397

### LINKED ISSUE #3397: depends_on restarts dependent services when up is called
I bring up two containers called master and slave where slave depends_on master.  I destroy master.  I then run `docker-compose up` and both slave and master are recreated even though nothing has changed about slave. 

```
/ $ docker-compose -v
/ $ docker -v
Docker version 1.11.1, build 5604cbe
```

DrewRomanyk:
Signed-off-by: Drew Romanyk <drewiswaycool@gmail.com>

This fixes the #3397 issue and adds an option flag to revert to the old behavior. 

ASSUMPTIONS:
If containers are stopped, recreate them
If containers have no links and have updated dependencies, recreate them

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -895,7 +895,7 @@ class TopLevelCommand(object):
             --no-deps                  Don't start linked services.
             --force-recreate           Recreate containers even if their configuration
                                        and image haven't changed.
-            --always-recreate-deps     Recreate dependant containers.
+            --always-recreate-deps     Recreate dependent containers.
                                        Incompatible with --no-recreate.
             --no-recreate              If containers already exist, don't recreate them.
                                        Incompatible with --force-recreate.
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -895,6 +895,7 @@ class TopLevelCommand(object):
             --no-deps                  Don't start linked services.
             --force-recreate           Recreate containers even if their configuration
                                        and image haven't changed.
+            --always-recreate-deps     Recreate dependant containers.
                                        Incompatible with --no-recreate.
             --no-recreate              If containers already exist, don't recreate them.
                                        Incompatible with --force-recreate.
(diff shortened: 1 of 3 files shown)


compose/project.py
```diff
@@ -508,7 +510,13 @@ def _get_convergence_plans(self, services, strategy):
                 log.debug('%s has upstream changes (%s)',
                           service.name,
                           ", ".join(updated_dependencies))
-                plan = service.convergence_plan(ConvergenceStrategy.always)
+                containers_stopped = any((not c.is_paused and not c.is_restarting and not c.is_running
```
shin-: I think you can do that more efficiently with `service.containers(stopped=True, filters={'status': ['created', 'exited']})`. See https://docs.docker.com/engine/api/v1.30/#operation/ContainerList for the filters docs

tests/integration/state_test.py
```diff
@@ -136,6 +136,17 @@ def test_change_middle(self):
...
+            set(['web_1']),
+        )
+
+    def test_change_middle_always_recreate_deps(self):
+        old_containers = self.run_up(self.cfg, **{"always_recreate_deps": True})
+
+        self.cfg['web']['environment'] = {'NEW_VAR': '1'}
+        new_containers = self.run_up(self.cfg, **{"always_recreate_deps": True})
+
         self.assertEqual(
```
shin-: Please use the `assert A == B` notation here instead (I know the other tests in this file use the `assertEqual` notation, but they were written before we switched to py.test. We'll transition them as well eventually)

compose/cli/main.py
```diff
@@ -895,6 +895,7 @@ def up(self, options):
             --no-deps                  Don't start linked services.
             --force-recreate           Recreate containers even if their configuration
                                        and image haven't changed.
+            --always-recreate-deps     Recreate dependant containers.
```
shin-: Let's use the "dependent" spelling instead, for consistency with https://github.com/docker/compose/blob/master/compose/config/sort_services.py#L34

compose/project.py
```diff
@@ -508,7 +510,13 @@ def _get_convergence_plans(self, services, strategy):
                 log.debug('%s has upstream changes (%s)',
                           service.name,
                           ", ".join(updated_dependencies))
-                plan = service.convergence_plan(ConvergenceStrategy.always)
+                containers_stopped = len(
```
shin-: Why not use `any()` here as well?

shin-:
Rebased and merged in #5541 

Thank you!

PR #5421, opened 2017-11-30
Add label config validation, fixes #4904

### LINKED ISSUE #4904: A label with a value of only numbers causes unmarshal error
We've been using a Git short hash as a label to identify builds, like so:

```
    labels:
      com.duo.access-gateway.builder.version: 336fb8d4
      com.duo.access-gateway.code.version: v1.4.3
      com.duo.access-gateway.built: Tue, 11 Apr 2017 13:02:06 -0000
```

This has worked great, until today, when I tried to call `docker-compose up` on a file with the following labels:

```
    labels:
      com.duo.access-gateway.builder.version: 89526235
      com.duo.access-gateway.code.version: v1.4.4-beta0
      com.duo.access-gateway.built: Fri, 02 Jun 2017 14:38:41 -0000
```

The following error was returned:

> ERROR: for access-gateway  Cannot create container for service access-gateway: json: cannot unmarshal number into Go value of type string

It took me a while to track this down, but it's due to `com.duo.access-gateway.builder.version` being all numbers. I'm going to quote our values now, but this doesn't seem like it should be an issue -- or if it should be an issue, perhaps a more helpful error message could be provided.

DrewRomanyk:
Signed-off-by: Drew Romanyk <drewiswaycool@gmail.com>

Add validation for labels, fixes #4904

ASSUMPTIONS
This applies to all labels & all versions of docker compose

EDIT:
Since this failed, my assumption is wrong, is there any documentation on where to find which labels want strings only? As the docker compose docs didnt seem to tell (https://docs.docker.com/compose/compose-file/#labels). Or is it just the build labels that can allow numbers?

### DIFF
--- a/compose/config/config_schema_v1.json
+++ b/compose/config/config_schema_v1.json
@@ -78,7 +78,7 @@
         "hostname": {"type": "string"},
         "image": {"type": "string"},
         "ipc": {"type": "string"},
-        "labels": {"$ref": "#/definitions/list_or_dict"},
+        "labels": {"$ref": "#/definitions/labels"},
         "links": {"type": "array", "items": {"type": "string"}, "uniqueItems": true},
         "log_driver": {"type": "string"},
         "log_opt": {"type": "object"},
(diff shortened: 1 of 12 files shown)


shin-:
I think that test is wrong, actually.

OTOH, now that we merged #5291 , there might be an opportunity to convert label values to string automatically instead. If you want to dig deeper that'd be a cool thing to do, but otherwise your PR is a good first step.

DrewRomanyk:
@shin- How should I handle cases where the labels could have '.' characters in them?

PR #5428, opened 2017-12-01
Allow port publish ranges

### LINKED ISSUE #5230: Port range like 8000-8010:80

shin-:
Fixes #5230

### DIFF
--- a/compose/config/types.py
+++ b/compose/config/types.py
@@ -319,11 +319,19 @@ class ServicePort(namedtuple('_ServicePort', 'target published protocol mode ext
         except ValueError:
             raise ConfigurationError('Invalid target port: {}'.format(target))
 
-        try:
-            if published:
-                published = int(published)
-        except ValueError:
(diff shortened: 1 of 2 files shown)


niklas88:
Using this to run a service multiple times on the host with `--scale webui=2` I get a warning:

    WARNING: The "webui" service specifies a port on the host. If multiple containers for this service are created on a single host, the port will clash.

Still it works afterwards, but it's a minor annoyance

PR #5432, opened 2017-12-04
Use random suffixes when naming containers instead of sequential numbers

### LINKED ISSUE #1516: Proposal: Stop using sequential container numbers

ilinum:
Fix #1516

This required a lot of changes to tests that relied on containers being sequentially numbered.

### DIFF
--- a/compose/utils.py
+++ b/compose/utils.py
@@ -7,6 +7,7 @@ import json
 import json.decoder
 import logging
 import ntpath
+import uuid
 
 import six
 from docker.errors import DockerException
@@ -143,3 +144,7 @@ def parse_bytes(n):
         return sdk_parse_bytes(n)
     except DockerException:
         return None
+
+
+def generate_uid():
+    return str(uuid.uuid4().hex)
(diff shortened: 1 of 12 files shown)


compose/utils.py
```diff
@@ -143,3 +144,7 @@ def parse_bytes(n):
         return sdk_parse_bytes(n)
     except DockerException:
         return None
+
+
+def generate_uid():
+    return str(uuid.uuid4().hex)
```
ilinum: This generates really long identifiers. Perhaps it would be worth making them 10 characters or so?

ilinum:
I can't seem to figure out why some tests are failing. 
When I run them locally, they all pass :(

ilinum:
I'm going to close this PR since I don't really have time to work on it.

I think it's mostly functional except for the flaky tests!

PR #5435, opened 2017-12-05
Raise error if up used with both -d and --timeout

### LINKED ISSUE #5434: docker-compose up should not allow -d and --timeout together
As mentioned in #5310, --timeout on docker-compose is meant to be used for interactive mode so it doesn't make sense to use in combination with -d. If a user provides both options it should result in an error.

summergirl21:
Raise an error if `-d` and `--timeout` both passed to `docker-compose up`.

Fix #5434 
Signed-off-by: Madeline Stager <stager.madeline@gmail.com>

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -889,8 +889,8 @@ class TopLevelCommand(object):
 
         Options:
             -d                         Detached mode: Run containers in the background,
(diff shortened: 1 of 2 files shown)


shin-:
Fixes #5310

PR #5441, opened 2017-12-06
Implemented --url option to specify url of compose file to use.

### LINKED ISSUE #2313: Add an argument to specify docker-compose using URL
Currently `docker-compose` only takes `docker-compose.yml` that has to be on the local file system. Just like `docker build`, it'll be good to provide a context for `docker-compose up -d`. This will allow to run an application from a github project.

summergirl21:
Implemented `--url` option for `docker-compose` that specifies a url to load the compose file from.
`docker-compose --url <url> [command]`

Fix #2313
Signed-off-by: Madeline Stager <stager.madeline@gmail.com>

### DIFF
--- a/compose/config/config.py
+++ b/compose/config/config.py
@@ -257,11 +257,18 @@ class ServiceConfig(namedtuple('_ServiceConfig', 'working_dir filename name conf
             config)
 
 
-def find(base_dir, filenames, environment, override_dir=None):
+def find(base_dir, filenames, environment, override_dir=None, url=None):
     if filenames == ['-']:
         return ConfigDetails(
             os.path.abspath(override_dir) if override_dir else os.getcwd(),
             [ConfigFile(None, yaml.safe_load(sys.stdin))],
+            environment
+        )
+    if url:
+        response = six.moves.urllib.request.urlopen(url)
+        return ConfigDetails(
+            os.path.abspath(override_dir) if override_dir else os.getcwd(),
+            [ConfigFile(None, yaml.safe_load(response.read()))],
             environment
         )
 
--- a/tests/acceptance/cli_test.py
+++ b/tests/acceptance/cli_test.py
@@ -782,6 +782,19 @@ class CLITestCase(DockerClientTestCase):
         assert 'Removing network v2full_default' in result.stderr
         assert 'Removing network v2full_front' in result.stderr
 
+    def test_down_url(self):
+        self.base_dir = '.'
+        url = 'https://raw.githubusercontent.com/docker/compose/' \
+              '771bf5f7ea059f1efede26bd9b164e05467d289e/tests/fixtures/' \
+              'simple-composefile/docker-compose.yml'
+        self._project = get_project(self.base_dir, override_dir=self.override_dir, url=url)
+
+        self.dispatch(['--url', url, 'up', '-d'])
+        wait_on_condition(ContainerCountCondition(self.project, 2))
+
+        self.dispatch(['--url', url, 'down'])
+        assert len(self.project.containers()) == 0
+
     def test_down_timeout(self):
         self.dispatch(['up', '-d'], None)
         service = self.project.get_service('simple')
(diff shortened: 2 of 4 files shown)


tests/acceptance/cli_test.py
```diff
@@ -782,6 +782,19 @@ def test_down(self):
         assert 'Removing network v2full_default' in result.stderr
         assert 'Removing network v2full_front' in result.stderr
 
+    def test_down_url(self):
+        self.base_dir = '.'
+        url = 'https://raw.githubusercontent.com/docker/compose/' \
+              '771bf5f7ea059f1efede26bd9b164e05467d289e/tests/fixtures/' \
+              'simple-composefile/docker-compose.yml'
```
johnharris85: Going out to a real url may cause this test to be flaky. I think the Engine tests this same functionality by spinning up a local static file server as part of the tests.

tests/acceptance/cli_test.py
```diff
@@ -816,6 +829,18 @@ def test_up_detached(self):
         self.assertFalse(container.get('Config.AttachStdout'))
         self.assertFalse(container.get('Config.AttachStdin'))
 
+    def test_up_url(self):
+        self.base_dir = '.'
+        url = 'https://raw.githubusercontent.com/docker/compose/' \
+              '771bf5f7ea059f1efede26bd9b164e05467d289e/tests/fixtures/' \
+              'simple-composefile/docker-compose.yml'
```
johnharris85: Ditto

compose/config/config.py
```diff
@@ -257,13 +257,20 @@ def with_abs_paths(cls, working_dir, filename, name, config):
...
-def find(base_dir, filenames, environment, override_dir=None):
+def find(base_dir, filenames, environment, override_dir=None, url=None):
     if filenames == ['-']:
         return ConfigDetails(
             os.path.abspath(override_dir) if override_dir else os.getcwd(),
             [ConfigFile(None, yaml.safe_load(sys.stdin))],
             environment
         )
+    if url:
+        response = six.moves.urllib.request.urlopen(url)
```
johnharris85: A lot could go wrong here so probably need some error handling somewhere. Not sure where this would propagate up to currently.

shin-:
There's a few issues with this, like overrides and cross-file extends not being supported (or maybe working in unexpected, possibly exploitable ways) which make me very wary of even considering the addition.

On top of that, I'm not sure what the benefits are overall - how is this different from `curl -L -o docker-compose.yml https://my.url/composefile && docker-compose up`?

micaelgallego:
Curl is not universally installed. Providing this feature in docker-compose would make downloading and installed multi-container apps as easy as single-container apps.

Please consider to implement this in docker-compose natively.

shin-:
@micaelgallego As I explained, the main concern here is limited functionality (cross-file extends will outright break), and subjecting users to possible security exploits related to these features.

As for the availability of `curl`, if only there were [a way to run arbitrary software](https://hub.docker.com/r/appropriate/curl/) on computers! That would truly be revolutionary :wink:

micaelgallego:
I see ;) Then you suggest execute something like: 

```docker run --rm -v $(pwd):/work -w /work appropriate/curl -L -o docker-compose.yml https://my.url/composefile && docker-compose up```

(NOTE: This command doesn't work)

Instead of:

```docker-compose --url https://my.url/composefile up```

I still prefer the second option ;)

**Comment by shin-** [comment]
You brought up the situation where `curl` might not be installed (and maybe it's not installable for some reason, you're on Windows or the ops guys won't let you install it on the server). In that case, a dockerized curl seems like a reasonable alternative.

**Comment by ia3andy** [comment]
> curl -L -o docker-compose.yml https://my.url/composefile && docker-compose up

Well your approach is creating a file in the current directory, which may not be wanted. Do you have a similar command to do this without creating that file?

Maybe the trade-off solution could be to add an option that takes the content of the compose file as input, then we just have to pipe it in?

**Comment by ia3andy** [comment]
BTW, you should maybe have a look to how `kubectl apply -f URL` is dealing with this problem (because they do support URL and files):
https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands

**Comment by ia3andy** [comment]
In my case, I want to give customers a way to deploy an application locally without having to download the file just with docker installed (the docker-compose is on github)

**Comment by jamshid** [comment]
I guess kubectl started supporting urls within the past couple of years? This answer did not mention it: https://stackoverflow.com/a/53499117/72717
URL support for `-f` would be a nice feature for docker-compose but you can use that suggested "curl pipe" approach with docker-compose:
```
curl -fsS https://.../docker-compose.yml | docker-compose -f - up -d
```

ia3andy:
> curl -fsS https://.../docker-compose.yml | docker-compose -f - up -d

Thanks @jamshid!

Issue #5445, opened 2017-12-06
logging HOST_CONFIG_KEY in compose/service.py


abuckenheimer:
I'm running into an issue where I can't create a container with a logging override:
```
Traceback (most recent call last):
  ...
  File "/home/ubuntu/dev/cosmos/app/hydrogen/hydrogen/docker.py", line 77, in test
    c = d.create_container(command=shlex.split(s.test_command), logging={})
  File "/home/ubuntu/miniconda3/lib/python3.6/site-packages/compose/service.py", line 295, in create_container
    return Container.create(self.client, **container_options)
  File "/home/ubuntu/miniconda3/lib/python3.6/site-packages/compose/container.py", line 46, in create
    response = client.create_container(**options)
TypeError: create_container() got an unexpected keyword argument 'logging'
```
Looking at the source in service.py I think the relevant issue is [here](https://github.com/docker/compose/blob/771bf5f7ea059f1efede26bd9b164e05467d289e/compose/service.py#L826) we try to pop off `HOST_CONFIG_KEYS` so we can add them instead nested inside the `'host_config'` option of `container_options`. However `HOST_CONFIG_KEYS` does not define `'logging'` thus leaving the option at the top level and causing the container to barf on init as this is unexpected. I think this should be fixed in one line just by adding `'logging'` to the `HOST_CONFIG_KEYS` and would be happy to make a pr for this.

shin-:
Please share the Compose files you're using.

abuckenheimer:
```
version: '2'
services:
  example:
    image: alpine
    command: ['echo', 'hello']
    logging:
      driver: awslogs
      options:
        awslogs-group: example_group
        awslogs-region: us-east-1
```

I'm trying to manually override the logging in the container so I can run it locally  (i.e. with json_file logs). An example of how I'm doing that is:

```
import os
from compose.cli.command import get_project
from compose.config.environment import Environment
project = get_project(os.getcwd(), ['docker-compose.yml'])
service = p.get_service('example')
container = service.create_container(logging={})
```

shin-:
Unfortunately we don't support using Compose as a Python library (#4542). If you really need to do that for some reason, I'd go with updating `service.options` with the desired `logging` value, e.g.

```python
service = p.get_service('example')
service.options['logging'] = {}
service.create_container()
```

HTH

abuckenheimer:
aha disappointing, if you have time I'm curious why the change I suggested is wrong otherwise thanks for the help.

PR #5446, opened 2017-12-06
added logging config key

abuckenheimer:
#5445


shin-:
That is not the correct fix.

Issue #3033, opened 2016-02-27
docker compose may flood target


dongluochen:
This is related to [Docker swarm issue 1859](https://github.com/docker/swarm/issues/1859). When docker-compose tries to scale web instance to 3000. It generates 3000 network connections to DOCKER_HOST simultaneously. I think both client (compose) and swarm manager cannot handle such load. I see a list of errors from it before compose exits. While swarm manager should handle such burst gracefully by queuing requests or rejecting requests. It'd be good that compose also have some rate limiting mechanism. 

docker-compose.yml:

```
web:
  image: nginx
  restart: always
  labels:
    com.docker.swarm.reschedule-policies: "[\"on-node-failure\"]"
```

```
Creating and starting 2998 ... done
Creating and starting 2999 ... done
Creating and starting 3000 ... done

ERROR: for 2  Get http://dong-coreos-9:2375/containers/0a9c180197ef697633208e467aff9c62350ed14634be0a3e27771941ff397658/json: dial tcp 172.19.35.107:2375: socket: too many open files 
ERROR: for 5  Get http://dong-coreos-6:2375/containers/fe6874e3bee55c793787e6093b1dec547a59423be3085c2bba034fb0acf4d902/json: dial tcp 172.19.35.104:2375: socket: too many open files 
ERROR: for 7  Container created but refresh didn't report it back 
ERROR: for 8  Container created but refresh didn't report it back 
ERROR: for 10  Get http://dong-coreos-15:2375/containers/de495e005bc2371f8d225a0c084da647ce8860683df75746e388c87021a98813/json: dial tcp 172.19.35.113:2375: socket: too many open files 
ERROR: for 13  Post http://dong-coreos-11:2375/v1.15/containers/32c3cef02a790a0d558307a80e7db04e2726160426c3a46ac64a5efc9fad6536/start: dial tcp 172.19.35.109:2375: socket: too many open files. Are you trying to connect to a TLS-enabled daemon without TLS? 
ERROR: for 14  Post http://dong-coreos-13:2375/v1.15/containers/create?name=crash_web_14: dial tcp 172.19.35.111:2375: socket: too many open files. Are you trying to connect to a TLS-enabled daemon without TLS? 
[...]

Issue #1828, opened 2015-08-07
Using "scale" exhausts system limits and then errors out


alvinr:
# Problem

Its possible to exhaust system resources by doing something like

```
$ docker-compose scale web=1000
```

It appears that we try to execute this request in parallel, causing the system limit of open files (in this case) to be exceed.

```
Exception in thread Thread-5:
Traceback (most recent call last):
  File "/compose/build/docker-compose/out00-PYZ.pyz/threading", line 810, in __bootstrap_inner
  File "/compose/build/docker-compose/out00-PYZ.pyz/threading", line 763, in run
  File "/compose/build/docker-compose/out00-PYZ.pyz/compose.utils", line 31, in inner_execute_function
  File "/compose/build/docker-compose/out00-PYZ.pyz/compose.service", line 220, in <lambda>
  File "/compose/build/docker-compose/out00-PYZ.pyz/compose.service", line 173, in create_and_start
  File "/compose/build/docker-compose/out00-PYZ.pyz/compose.service", line 268, in create_container
  File "/compose/build/docker-compose/out00-PYZ.pyz/compose.service", line 579, in _get_container_create_options
  File "/compose/build/docker-compose/out00-PYZ.pyz/compose.service", line 464, in config_hash
  File "/compose/build/docker-compose/out00-PYZ.pyz/compose.service", line 469, in config_dict
  File "/compose/build/docker-compose/out00-PYZ.pyz/compose.service", line 295, in image
  File "/compose/build/docker-compose/out00-PYZ.pyz/docker.utils.decorators", line 20, in wrapped
  File "/compose/build/docker-compose/out00-PYZ.pyz/docker.client", line 489, in inspect_image
  File "/compose/build/docker-compose/out00-PYZ.pyz/docker.clientbase", line 86, in _get
  File "/compose/build/docker-compose/out00-PYZ.pyz/requests.sessions", line 477, in get
  File "/compose/build/docker-compose/out00-PYZ.pyz/requests.sessions", line 465, in request
  File "/compose/build/docker-compose/out00-PYZ.pyz/requests.sessions", line 573, in send
  File "/compose/build/docker-compose/out00-PYZ.pyz/requests.adapters", line 370, in send
  File "/compose/build/docker-compose/out00-PYZ.pyz/requests.packages.urllib3.connectionpool", line 544, in urlopen
  File "/compose/build/docker-compose/out00-PYZ.pyz/requests.packages.urllib3.connectionpool", line 341, in _make_request
[...]

PR #5464, opened 2017-12-10
Add COMPOSE_PARALLEL_LIMIT to restrict global number of parallel oper…

Rozelette:
…ations

Addresses #3033 and #1828 

Adds a global limit configurable by setting a COMPOSE_PARALLEL_LIMIT environment variable, defaulting to 64.

Signed-off-by: Shea Rozmiarek <uberpanzermensch@gmail.com>

### DIFF
--- a/compose/parallel.py
+++ b/compose/parallel.py
@@ -15,6 +15,8 @@ from six.moves.queue import Queue
 from compose.cli.colors import green
 from compose.cli.colors import red
 from compose.cli.signals import ShutdownException
+from compose.config.environment import Environment
+from compose.const import PARALLEL_LIMIT
 from compose.errors import HealthCheckFailed
 from compose.errors import NoHealthCheckConfigured
 from compose.errors import OperationFailedError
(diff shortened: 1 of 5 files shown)


compose/parallel.py
```diff
@@ -26,6 +28,18 @@
 STOP = object()
 
 
+def get_configured_limit():
+    limit = Environment.from_command_line({'COMPOSE_PARALLEL_LIMIT': None})['COMPOSE_PARALLEL_LIMIT']
```
shin-: While I get why you're doing it like this, I think we're definitely looking to be able to set this value in the `.env` file as well.

shin-: I'd suggest looking at how we went with implementing the `--no-ansi` option in #5093 for guidance.

tests/unit/parallel_test.py
```diff
@@ -67,6 +69,31 @@ def f(obj):
     assert errors == {}
 
 
+def test_parallel_execute_with_global_limit():
```
shin-: Make sure to use a `@mock.patch.dict(os.environ)` decorator to avoid bleeding test values into the user's environment and prevent unexpected side-effects with other tests.

ashmrtn: This is no longer required as the code for the test no longer sets the environment variable the way it used to

shin-:
Thank you for your help! :+1:

johnharris85:
This probably needs adding to the [docs here](https://docs.docker.com/compose/reference/envvars/) on the vnext-compose branch as well right?

shin-:
I usually add everything in one go after RC1 goes out. Less context switching :)

johnharris85:
Makes sense :)

Issue #5124, opened 2017-08-22
Sequential restart when using `docker-compose up` with scale


corradio:
Hello,

It seems to me that when scale is > 1, and a new image is being pulled, using `docker-compose up` restarts all containers in parallel rather than sequentially.
Doing a cascading restart would ensure zero downtime.

Am I correct in my analysis?

Thank you,

Olivier

shin-:
Both are valid strategies depending on your use-case, but you're right that we don't currently offer an automated way to do zero-downtime deployments. In the meantime, you may want to take a look at https://github.com/docker/compose/issues/1786#issuecomment-294434786 which outlines how to do so manually.

corradio:
Hi @shin- , 
Unfortunately your method won't work as I have more than 1 container in production.
With previous versions of docker/docker-compose, it seems like containers were restarted sequentially. Did anything change?
It would be nice to be able to control the start strategy in docker-compose (sequential vs parallel)

shin-:
> With previous versions of docker/docker-compose, it seems like containers were restarted sequentially. Did anything change?

Unless you're looking at versions from 2 years ago, that wasn't the case.

corradio:
I was indeed working with a very old version (due to CI constraints).
Any idea if it's possible to add the feature I'm requesting? I'm up for contributing if needed.

domq:
Replying to the [original question](https://github.com/docker/compose/issues/5124#issue-251914680), it is still not possible to do sequential restarts. *However*, you can prevent `docker-compose up --scale` from restarting existing containers using [the `--no-recreate` flag](https://docs.docker.com/compose/reference/up/). This is enough to do zero-downtime restarts if you only run one serving instance (by momentarily scaling to two, then deleting the oldest container).

PR #5471, opened 2017-12-11
Allow limiting of the number of containers restarted in parallel for a scaled service

ashmrtn:
This patch addresses #5124 and adds a new command line flag for `docker-compose up` called the `rollout-factor`. The flag limits the number of containers that will be restarted in parallel, allowing the service to restart with zero-downtime.

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -916,6 +916,8 @@ class TopLevelCommand(object):
                                        Implies --abort-on-container-exit.
             --scale SERVICE=NUM        Scale SERVICE to NUM instances. Overrides the `scale`
                                        setting in the Compose file if present.
+            --rollout-factor NUM       The number of containers to restart at one time when
+                                       the scale factor is greater than 1.
         """
         start_deps = not options['--no-deps']
         exit_value_from = exitval_from_opts(options, self.project)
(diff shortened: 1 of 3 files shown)


ashmrtn:
As of now, this patch has no tests associated with it and I am not completely sure how to test it as it basically adds a command line flag and then uses pre-existing features like `parallel_execute`'s `limiter` object. Do you have any suggestions as to how to test this?

shin-:
I'm afraid that this might set up false expectations regarding the availability of the application during upgrade - there's definitely a scenario where the last container is getting restarted while the first one is still booting (i.e. not ready to receive connections). While the feature would be a nice to have, I think the implementation of it is ultimately more complex.

ashmrtn:
That is an excellent point that I completely neglected to think about when making this patch. From what I understand, docker-compose supported something vaguely similar when it allowed people to specify healthchecks for the `condition` form of `depends_on`. The fact that compose no longer supports that makes it feel like compose technically shouldn't support something like rolling out containers as it requires similar logic to check if a container is ready to receive requests. Do you have any thoughts on how to determine if a container is ready to receive requests that could be added to the current PR?

ijc:
> Do you have any thoughts on how to determine if a container is ready to receive requests that could be added to the current PR?

Is the `HEALTHCHECK` stuff ([builder docs](https://docs.docker.com/engine/reference/builder/#healthcheck), [compose docs](https://docs.docker.com/compose/compose-file/#healthcheck)) useful/relevant here? IOW only consider any new container good once it has entered the Healthy state?

PR #5474, opened 2017-12-11
Added prioritization of networks. Resolves #5042

### LINKED ISSUE #5042: Cannot determine which port mapping should apply to which network
I've been struggling with an issue while trying to document Compose support for a three tier app using two networks in vSphere Integrated Containers. Our documentation really needed improving and as part of that work, I'm putting together guidelines on how to get the best out of both technologies.

The issue I have is that for a container that uses multiple networks, the order in which they're brought up matters. For example, when using a docker client, I would do the following:

```
docker network create internal
docker create -d --name web --net external -p 80 myImage
docker network connect internal web
docker start web
```
In the above example, I'm asking for port 80 to be exposed on the external network. If I flip it around, instead I get a port mapping on the bridge network.

```
docker network create internal
docker create -d --name web --net internal -p 80 myImage
docker network connect external web
docker start web
```
docker-compose seems to always use the same order, regardless of how the compose file is written - creating the container with the internal network and then adding the external network before starting it.

My question is: Is there any way to influence this ordering? The current compose file format leave it somewhat ambiguous - it allows you to expose ports and specify multiple networks, but there's no way to determine which ports are exposed on which networks. 

Any help in resolving this would be appreciated. Thanks!

zaldaroga:
Signed-off-by: Zal Daroga <zaldaroga@yahoo.com>

### DIFF
--- a/compose/config/config_schema_v3.0.json
+++ b/compose/config/config_schema_v3.0.json
@@ -310,7 +310,8 @@
           "additionalProperties": false
         },
         "internal": {"type": "boolean"},
-        "labels": {"$ref": "#/definitions/list_or_dict"}
+        "labels": {"$ref": "#/definitions/list_or_dict"},
+        "priority": {"type": "number"}
       },
       "additionalProperties": false
     },
--- a/compose/service.py
+++ b/compose/service.py
@@ -6,6 +6,7 @@ import os
 import re
 import sys
 from collections import namedtuple
+from collections import OrderedDict
 from operator import attrgetter
 
 import enum
@@ -556,10 +557,15 @@ class Service(object):
             raise OperationFailedError("Cannot start service %s: %s" % (self.name, ex.explanation))
         return container
 
+    def prioritized_networks(self):
+        prioritized_networks = OrderedDict(
+            sorted(self.networks.items(), key=lambda t: t[1].get('priority', 0), reverse=True))
+        return prioritized_networks
+
     def connect_container_to_networks(self, container):
         connected_networks = container.get('NetworkSettings.Networks')
 
-        for network, netdefs in self.networks.items():
+        for network, netdefs in self.prioritized_networks().items():
             if network in connected_networks:
                 if short_id_alias_exists(container, network):
                     continue
--- a/compose/service.py
+++ b/compose/service.py
@@ -559,7 +559,7 @@ class Service(object):
 
     def prioritized_networks(self):
         prioritized_networks = OrderedDict(
-            sorted(self.networks.items(), key=lambda t: t[1].get('priority', 0), reverse=True))
+            sorted(self.networks.items(), key=lambda t: t[1].get('priority', 0) or 0, reverse=True))
         return prioritized_networks
 
     def connect_container_to_networks(self, container):
--- a/compose/service.py
+++ b/compose/service.py
@@ -574,6 +574,7 @@ class Service(object):
                     container.id,
                     network)
 
+            print('Connecting to {}'.format(network))
             self.client.connect_container_to_network(
                 container.id, network,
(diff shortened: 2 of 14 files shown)


zaldaroga:
cc @mhank

shin-:
Thank you for your contribution! Before I take a closer look, can you make sure the test suite passes? Let me know if you need help figuring out some of those.

zaldaroga:
Hi @shin- 
When I run the test suite locally, I don't encounter any failures, so I'm not too sure how to figure out what the problem is. From looking at the jenkins console log, it seems like all of the errors are displaying this:
> requests.exceptions.ReadTimeout: HTTPConnectionPool(host='docker', port=2375): Read timed out. (read timeout=60)
Do you know why this might be happening/what I might need to do to fix it?
Thanks,
Zal

shin-:
I believe that's a fluke. I re-triggered the tests, we'll see what it says this time around.

compose/config/config_schema_v3.0.json
```diff
@@ -310,7 +310,8 @@
           "additionalProperties": false
         },
         "internal": {"type": "boolean"},
-        "labels": {"$ref": "#/definitions/list_or_dict"}
+        "labels": {"$ref": "#/definitions/list_or_dict"},
+        "priority": {"type": "number"}
```
shin-: Unfortunately, we can't modify the 3.x schemas as they are frozen to conform to the shape they were released in in the Engine.

For the time being, it's fine to keep this as a 2.x feature.

compose/service.py
```diff
@@ -556,10 +557,15 @@ def start_container(self, container):
             raise OperationFailedError("Cannot start service %s: %s" % (self.name, ex.explanation))
         return container
 
+    def prioritized_networks(self):
+        prioritized_networks = OrderedDict(
```
shin-: We can return the value directly without using an intermediary variable here.

compose/service.py
```diff
@@ -556,10 +557,15 @@ def start_container(self, container):
             raise OperationFailedError("Cannot start service %s: %s" % (self.name, ex.explanation))
         return container
 
+    def prioritized_networks(self):
```
shin-: Could we add an `@property` decorator here?

compose/service.py
```diff
@@ -568,6 +574,7 @@ def connect_container_to_networks(self, container):
                     container.id,
                     network)
 
+            print('Connecting to {}'.format(network))
```
shin-: Instead of `print`, we probably want to use `log.debug` here. It's not particularly useful information if the application runs normally.

tests/acceptance/cli_test.py
```diff
@@ -1244,6 +1244,13 @@ def test_up_with_net_v1(self):
         assert foo_container.get('HostConfig.NetworkMode') == \
             'container:{}'.format(bar_container.id)
 
+    def test_up_ordered_networks(self):
+        self.base_dir = 'tests/fixtures/networks'
+        result = self.dispatch(['-f', 'ordered-networks.yml', 'up', '-d'])
+
+        assert 'Connecting to networks_buzz\nConnecting to networks_foo' \
```
shin-: If we're removing the `print` statement above, you'll at least need to add `--verbose` to the `dispatch` call above. 
But I'd be wary of relying on those 3 specific lines appearing consecutively all the time. I think a better test would involve different `mac_address` values, and checking that the network with the highest priority sets the `mac_address` for the container.
If you do it that way, it's also probably easier to do it as an integration test so you can change priority values and check the container configuration changes accordingly.

shin-:
Updated and merged in #5566 

Thanks @zaldaroga !

PR #5490, opened 2017-12-18
Convert mounts to legacy volumes if API version < 1.30

### LINKED ISSUE #5489: API Version mismatch between docker-compose and system docker-ce
**Relevant `docker-compose --verbose up` output**
```
urllib3.connectionpool._make_request: http://localhost:None "GET /v1.25/images/ormuco-vpn:compose-build/json HTTP/1.1" 200 2041
```

Then a little bit later...

```
compose.cli.verbose_proxy.proxy_callable: docker create_host_config <- (device_read_iops=None, mem_swappiness=None, links=[], oom_score_adj=None, blkio_weight=None, cpu_count=None, cpuset_cpus=None, dns_search=None, pid_mode=None, init_path=None, log_config={'Type': u'', 'Config': {}}, cpu_quota=None, read_only=None, cpu_percent=None, device_read_bps=None, storage_opt=None, init=None, dns=None, volumes_from=[], ipc_mode=None, mem_reservation=None, security_opt=None, shm_size=None, device_write_iops=None, dns_opt=None, cgroup_parent=None, group_add=None, network_mode=u'cerebrom_default', volume_driver=None, oom_kill_disable=None, userns_mode=None, tmpfs=None, nano_cpus=None, port_bindings={'9600/tcp': [None]}, isolation=None, memswap_limit=None, restart_policy=None, blkio_weight_device=None, devices=None, extra_hosts=None, binds=[], sysctls=None, pids_limit=None, device_write_bps=None, cap_add=None, mounts=[{'Source': u'/home/wayne/projects/ormuco/cerebrom/docker/ormuco-vpn/credentials/vpnc', 'ReadOnly': None, 'Type': 'bind', 'Target': '/etc/vpnc'}, {'Source': u'/home/wayne/projects/ormuco/cerebrom/docker/ormuco-vpn/credentials/ssh', 'ReadOnly': None, 'Type': 'bind', 'Target': '/ssh'}], mem_limit=None, cap_drop=None, privileged=False, ulimits=None, cpu_shares=None)
compose.parallel.parallel_execute_iter: Failed: <Container: 21d4214f35f5_cerebrom_ormuco-vpn_1 (21d421)>
compose.parallel.feed_queue: Pending: set([])

ERROR: for 21d4214f35f5_cerebrom_ormuco-vpn_1  mounts param is not supported in API versions < 1.30
compose.parallel.parallel_execute_iter: Failed: <Service: ormuco-vpn>
compose.parallel.feed_queue: Pending: set([<Service: proctor>])
compose.parallel.feed_queue: <Service: proctor> has upstream errors - not processing
compose.parallel.parallel_execute_iter: Failed: <Service: proctor>
compose.parallel.feed_queue: Pending: set([])

ERROR: for ormuco-vpn  mounts param is not supported in API versions < 1.30
  File "/home/wayne/.virtualenvs/cerebrom-5qiep0Vx/bin/docker-compose", line 11, in <module>
    sys.exit(main())
  File "/home/wayne/.virtualenvs/cerebrom-5qiep0Vx/local/lib/python2.7/site-packages/compose/cli/main.py", line 71, in main
    command()
  File "/home/wayne/.virtualenvs/cerebrom-5qiep0Vx/local/lib/python2.7/site-packages/compose/cli/main.py", line 124, in perform_command
    handler(command, command_options)
  File "/home/wayne/.virtualenvs/cerebrom-5qiep0Vx/local/lib/python2.7/site-packages/compose/cli/main.py", line 956, in up
    start=not no_start
  File "/home/wayne/.virtualenvs/cerebrom-5qiep0Vx/local/lib/python2.7/site-packages/compose/project.py", line 479, in up
    get_deps,
  File "/home/wayne/.virtualenvs/cerebrom-5qiep0Vx/local/lib/python2.7/site-packages/compose/parallel.py", line 80, in parallel_execute
    raise error_to_reraise
docker.errors.InvalidVersion: mounts param is not supported in API versions < 1.30
```

**`docker version` output**
```
Client:
 Version:      17.09.0-ce
 API version:  1.32
 Go version:   go1.8.3
 Git commit:   afdb6d4
 Built:        Tue Sep 26 22:42:09 2017
 OS/Arch:      linux/amd64

Server:
 Version:      17.09.0-ce
 API version:  1.32 (minimum version 1.12)
 Go version:   go1.8.3
 Git commit:   afdb6d4
 Built:        Tue Sep 26 22:40:48 2017
 OS/Arch:      linux/amd64
 Experimental: false
```

**`docker info` output**
```
Containers: 13
 Running: 4
 Paused: 0
 Stopped: 9
Images: 245
Server Version: 17.09.0-ce
Storage Driver: overlay
 Backing Filesystem: extfs
 Supports d_type: true
Logging Driver: json-file
Cgroup Driver: cgroupfs
Plugins:
 Volume: local
 Network: bridge host macvlan null overlay
 Log: awslogs fluentd gcplogs gelf journald json-file logentries splunk syslog
Swarm: inactive
Runtimes: runc
Default Runtime: runc
Init Binary: docker-init
containerd version: 06b9cb35161009dcb7123345749fef02f7cea8e0
runc version: 3f2f8b84a77f73d38244dd690525642a72156c64
init version: 949e6fa
Security Options:
 seccomp
  Profile: default
Kernel Version: 4.9.0-3-amd64
Operating System: Debian GNU/Linux 9 (stretch)
OSType: linux
Architecture: x86_64
CPUs: 12
Total Memory: 23.5GiB
Name: agora
ID: UW3Z:67B4:Y2UM:YQS6:RD6G:7IMS:NE5R:3P32:27YZ:ZJQD:44ZA:QVX5
Debug Mode (client): false
Debug Mode (server): false
Experimental: false
Insecure Registries:
 127.0.0.0/8
Live Restore Enabled: false
```

**`docker-compose --version` output**
```
docker-compose version 1.18.0-rc2, build 189468b
```

**my thoughts**
It seems strange to me that `docker-compose` attempts to use API version `1.25` rather than `1.32` as reported by `docker version` (see above). Is it because `docker-compose` `1.18.0-rc2` simply doesn't support API version `1.32`? Or am I doing something incorrectly that causes `docker-compose` to use API version `1.25`?

Also it's worth mentioning that I am attempting to use https://docs.docker.com/compose/compose-file/#long-syntax-3; whenever I exclude the `volumes` key from the service this problem affects the api version conflict doesn't occur. I guess the next thing I'll try is https://docs.docker.com/compose/compose-file/#short-syntax-3.

shin-:
Compose file v3.2 introduces long syntax (mount syntax) for containers but is pinned to API version 1.25, which doesn't support container mounts. This fix converts mounts to volumes if the API version is too low.

Fixes #5489

### DIFF
--- a/compose/service.py
+++ b/compose/service.py
@@ -785,34 +785,9 @@ class Service(object):
             self.options.get('labels'),
             override_options.get('labels'))
 
-        container_volumes = []
-        container_mounts = []
-        if 'volumes' in container_options:
-            container_volumes = [
-                v for v in container_options.get('volumes') if isinstance(v, VolumeSpec)
-            ]
-            container_mounts = [v for v in container_options.get('volumes') if isinstance(v, MountSpec)]
-
-        binds, affinity = merge_volume_bindings(
-            container_volumes, self.options.get('tmpfs') or [], previous_container,
-            container_mounts
+        container_options, override_options = self._build_container_volume_options(
+            previous_container, container_options, override_options
         )
-        override_options['binds'] = binds
-        container_options['environment'].update(affinity)
-
-        container_options['volumes'] = dict((v.internal, {}) for v in container_volumes or {})
-        override_options['mounts'] = [build_mount(v) for v in container_mounts] or None
-
-        secret_volumes = self.get_secret_volumes()
-        if secret_volumes:
-            if version_lt(self.client.api_version, '1.30'):
-                override_options['binds'].extend(v.legacy_repr() for v in secret_volumes)
-                container_options['volumes'].update(
-                    (v.target, {}) for v in secret_volumes
-                )
-            else:
-                override_options['mounts'] = override_options.get('mounts') or []
-                override_options['mounts'].extend([build_mount(v) for v in secret_volumes])
 
         container_options['image'] = self.image_name
 
@@ -837,6 +812,42 @@ class Service(object):
         container_options['environment'] = format_environment(
             container_options['environment'])
         return container_options
+
+    def _build_container_volume_options(self, previous_container, container_options, override_options):
+        container_volumes = []
+        container_mounts = []
+        if 'volumes' in container_options:
+            container_volumes = [
+                v for v in container_options.get('volumes') if isinstance(v, VolumeSpec)
+            ]
+            container_mounts = [v for v in container_options.get('volumes') if isinstance(v, MountSpec)]
+
+        binds, affinity = merge_volume_bindings(
+            container_volumes, self.options.get('tmpfs') or [], previous_container,
+            container_mounts
+        )
+        override_options['binds'] = binds
+        container_options['environment'].update(affinity)
+
+        container_options['volumes'] = dict((v.internal, {}) for v in container_volumes or {})
+        if version_gte(self.client.api_version, '1.30'):
+            override_options['mounts'] = [build_mount(v) for v in container_mounts] or None
+        else:
+            override_options['binds'].extend(m.legacy_repr() for m in container_mounts)
+            container_options['volumes'].update((m.target, {}) for m in container_mounts)
+
+        secret_volumes = self.get_secret_volumes()
+        if secret_volumes:
+            if version_lt(self.client.api_version, '1.30'):
+                override_options['binds'].extend(v.legacy_repr() for v in secret_volumes)
+                container_options['volumes'].update(
+                    (v.target, {}) for v in secret_volumes
+                )
+            else:
+                override_options['mounts'] = override_options.get('mounts') or []
+                override_options['mounts'].extend([build_mount(v) for v in secret_volumes])
+
+        return container_options, override_options
 
     def _get_container_host_config(self, override_options, one_off=False):
         options = dict(self.options, **override_options)
--- a/tests/integration/service_test.py
+++ b/tests/integration/service_test.py
@@ -13,6 +13,7 @@ from six import StringIO
 from six import text_type
 
 from .. import mock
+from .testcases import docker_client
 from .testcases import DockerClientTestCase
 from .testcases import get_links
 from .testcases import pull_busybox
@@ -320,6 +321,23 @@ class ServiceTest(DockerClientTestCase):
             'db',
             volumes=[MountSpec(type='volume', source=volume_name, target=container_path)]
         )
+        container = service.create_container()
+        service.start_container(container)
+        mount = container.get_mount(container_path)
+        assert mount
+        assert mount['Name'] == volume_name
+
+    @v3_only()
+    def test_create_container_with_legacy_mount(self):
+        # Ensure mounts are converted to volumes if API version < 1.30
+        # Needed to support long syntax in the 3.2 format
+        client = docker_client({}, version='1.25')
+        container_path = '/container-volume'
+        volume_name = 'composetest_abcde'
+        self.client.create_volume(volume_name)
+        service = Service('db', client=client, volumes=[
+            MountSpec(type='volume', source=volume_name, target=container_path)
+        ], image='busybox:latest', command=['top'], project='composetest')
         container = service.create_container()
         service.start_container(container)
         mount = container.get_mount(container_path)


thaJeztah:
Can we reconsider this change, and make docker-compose perform API version negotiation instead?

The volume ("binds") API has different semantics than the "mounts" API: the mounts API _explicitly_ does _not_ create the host-path when using bind-mounts. By switching to the volume API, we're introducing a change in behavior between the 3.2 and 3.3 (and up) docker-compose file formats.

For example, using the 3.3 format:

```bash
$ mkdir repro-5489 && cd repro-5489

$ cat > docker-compose.yml <<'EOF'
version: "3.3"
services:
  mysql:
    image: "nginx:alpine"
    volumes:
      - type: bind
        source: ./mysql_data
        target: /var/lib/mysql
EOF


$ docker-compose up -d

Compose does not use swarm mode to deploy services to multiple nodes in a swarm. All containers will be scheduled on the current node.

To deploy your application across the swarm, use `docker stack deploy`.

Creating network "repro5489_default" with the default driver
Creating repro5489_mysql_1 ... error

ERROR: for repro5489_mysql_1  Cannot create container for service mysql: invalid mount config for type "bind": bind source path does not exist

ERROR: for mysql  Cannot create container for service mysql: invalid mount config for type "bind": bind source path does not exist
```

The service fails to start because the host path is not found; this is the intended behavior.


However, the 3.2 file-format (due to using the "volumes" API) succeeds;

```
$ cat > docker-compose.yml <<'EOF'
version: "3.2"
services:
  mysql:
    image: "nginx:alpine"
    volumes:
      - type: bind
        source: ./mysql_data
        target: /var/lib/mysql
EOF


$ docker-compose up -d

Compose does not use swarm mode to deploy services to multiple nodes in a swarm. All containers will be scheduled on the current node.

To deploy your application across the swarm, use `docker stack deploy`.

Creating network "repro5489_default" with the default driver
Creating repro5489_mysql_1 ... done
```

After deploying the stack, there's now a `mysql_data` directory created on the host;

```bash
$ ls -la
total 8
drwxr-xr-x    4 sebastiaan  staff   136 Dec 21 17:47 .
drwxr-xr-x  106 sebastiaan  staff  3604 Dec 21 17:38 ..
-rw-r--r--    1 sebastiaan  staff   152 Dec 21 17:47 docker-compose.yml
drwxr-xr-x    2 sebastiaan  staff    68 Dec 21 17:47 mysql_data
```

This is _not_ the intended behavior; when using the advanced volume notation (and the Mounts API), the host-path should _not_ be created.


I think the correct sollution would be to:

- only specify the _minimum_ required version for a compose-file version (if needed, but these options could be set "per action"; for example: allow using a 3.5 compose file against a docker 1.12 daemon, but if the file uses a "config", that part will fail)
- docker-compose should perform API version negotiation (fall back to an older API version if the daemon does not support the current version)

/cc @dnephin @cpuguy83

dnephin:
> allow using a 3.5 compose file against a docker 1.12 daemon, but if the file uses a "config", that part will fail

I don't think there's such a thing as "part of the config failing". Either if failed to deploy or it didn't. I don't think it makes sense to downgrade if the config isn't going to work. It makes sense to have a hard requirement of a minimum version required to use the features of a compose file version. If you want to use older daemons, use an older compose file version. Older versions all still supported for this reason.

This is the way compose has always worked, and it's a lot more predictable than downgrading/version negotiation.

The behaviour in thiss PR seems correct to me, but we should make sure it's well documented.

thaJeztah:
> I don't think there's such a thing as "part of the config failing".

Perhaps badly worded; but should the deploy fail if a 3.5 file is used, but all options used are actually 3.0 options? (so non of the options require the 3.5 spec)? - agreed that adjusting the version is an option for that, but with multiple compose files, I can imagine that being problematic.

> it's a lot more predictable than downgrading/version negotiation.

If we don't use negotiation, should we raise the minimum specification for the 3.2 format, because the API version that it currently uses does not support the Mount API thats needed for the extended volumes format.

dnephin:
Changing the minimum version for 3.2 could break people. We could add a warning if the version is 3.2 and the compose file contains an expanded volume mount, informing the user that they may get unexpected results and should use 3.3.

thaJeztah:
> Changing the minimum version for 3.2 could break people.

Yes, so for that reason, I thought having a "minimum" api version (but use "current" API version) for the file would work. The advanced syntax never worked for that file-version, so doing so wouldn't break from that perspective.

shin-:
@thaJeztah To be clear, the "degrading" to using volumes with the advanced syntax was the only behavior for Compose up until now. The new release now uses mounts when it's available. If anything, some breakage may occur for users who were previously using a 3.3 or 3.4 file and relying on the folder being created.

I don't really want to mess with the API - format mapping, because these are predictable rules we depend on (both on the programmer side and the "support" side) and changes there could create a whole new set of problems. Adding a warning for 3.2 users sounds good though.

PR #5562, opened 2018-01-11
Fix #5465 by catching no-image exception

### LINKED ISSUE #5465: Building container fails after removing old image.
I found this issue while working on a fix for issue #4337.

By running the following sequence of commands on [this example repository](https://github.com/twelve17/dc-test):

```
git checkout tags/test2_run1
docker-compose build
docker rmi --force dctest_service1
docker-compose build
docker-compose up
```

Compose fails because the old image is never properly rebuilt from the old container:

```

Compose does not use swarm mode to deploy services to multiple nodes in a swarm. All containers will be scheduled on the current node.

To deploy your application across the swarm, use `docker stack deploy`.

Recreating dctest_service1_1 ... error

ERROR: for dctest_service1_1  no such image: sha256:bb41503c2e723478308998afc7df0886fac0811621b70d61b28ca36418a41bb7: No such image: sha256:bb41503c2e723478308998afc7df0886fac0811621b70d61b28ca36418a41bb7

ERROR: for service1  no such image: sha256:bb41503c2e723478308998afc7df0886fac0811621b70d61b28ca36418a41bb7: No such image: sha256:bb41503c2e723478308998afc7df0886fac0811621b70d61b28ca36418a41bb7
```

Dahca:
Fix #5465 by catching a no-image exception in get_container_data_volumes. Added a case to the unit test to check that an empty existing image doesn't result in an exception.

Closed old pull request #5466 because I did a rebase and it showed over 300 commits to merge, which I thought was messy.

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -974,24 +974,50 @@ class TopLevelCommand(object):
             raise UserError("COMPOSE_IGNORE_ORPHANS and --remove-orphans cannot be combined.")
 
         if no_start:
-            for excluded in ['-d', '--abort-on-container-exit', '--exit-code-from']:
-                if options.get(excluded):
-                    raise UserError('--no-start and {} cannot be combined.'.format(excluded))
+            opts = ['-d', '--abort-on-container-exit', '--exit-code-from']
+            for excluded in [x for x in opts if options.get(x)]:
+                raise UserError('--no-start and {} cannot be combined.'.format(excluded))
 
         with up_shutdown_context(self.project, service_names, timeout, detached):
-            to_attach = self.project.up(
-                service_names=service_names,
-                start_deps=start_deps,
-                strategy=convergence_strategy_from_opts(options),
-                do_build=build_action_from_opts(options),
-                timeout=timeout,
-                detached=detached,
-                remove_orphans=remove_orphans,
-                ignore_orphans=ignore_orphans,
-                scale_override=parse_scale_args(options['--scale']),
-                start=not no_start,
-                always_recreate_deps=always_recreate_deps
-            )
+            try:
+                to_attach = self.project.up(
+                    service_names=service_names,
+                    start_deps=start_deps,
+                    strategy=convergence_strategy_from_opts(options),
+                    do_build=build_action_from_opts(options),
+                    timeout=timeout,
+                    detached=detached,
+                    remove_orphans=remove_orphans,
+                    ignore_orphans=ignore_orphans,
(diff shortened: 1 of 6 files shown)


shin-:
@Dahca Thanks! The messy rebase was probably from when I re-aligned master with the release branch :confounded: sorry.

Do you mind squashing these commits together? Other than that I think we're good to go.

Dahca:
Hey @shin-, not a problem!

Is this squashed enough? I was trying to squash the whole thing, but when I do it always has a merge conflict when I do the rebase from upstream.

Dahca:
Looks like jenkins timed out -- is there a way to re-run the check?

compose/cli/main.py
```diff
@@ -979,21 +979,23 @@ def up(self, options):
...
         with up_shutdown_context(self.project, service_names, timeout, detached):
-            try:
-                to_attach = self.project.up(
+            def up(rebuild):
+                return self.project.up(
                     service_names=service_names,
                     start_deps=start_deps,
                     strategy=convergence_strategy_from_opts(options),
-                    do_build=build_action_from_opts(options),
+                    do_build=build_action_from_opts(options) if not rebuild else BuildAction.force,
```
Dahca: This actually doesn't fix the problem. When the build is forced, it still checks the images and throws the NoImageFound exception.

Dahca:
@shin- The bug is still present even with BuildAction.force.

compose/cli/main.py
shin-: Ah, that sucks :(

Can we add a test for it? I know there's a unit test there but it didn't catch anything with my changes, so we probably need something more comprehensive. Once we have a good test we can figure out what the best approach to fixing it is.

Dahca: Alright, I'll work on that.

tests/unit/service_test.py
```diff
@@ -926,7 +927,7 @@ def test_get_container_data_volumes(self):
         volumes, _ = get_container_data_volumes(container, options, ['/dev/tmpfs'], [])
         assert sorted(volumes) == sorted(expected)
 
-    def test_get_container_data_volumes_image_is_none(self):
+    def test_get_container_data_volumes_image_does_not_exist(self):
```
Dahca: @shin- Here's the updated test, which currently fails.

shin-: Mmmh ok, I'll take a look over the next couple of days. Thank you!

shin-:
Merged in #5583 - thank you for your help and patience!

PR #5569, opened 2018-01-12
Fix unicode error - python2

thomascrha:
If the name of a container doesn't exist and its name isn't utf-8 encoded you receive a stacktrace


shin-:
Thank you!

PR #5579, opened 2018-01-16
Add bash completion for `ps --services --filter`

albers:
This adds bash completion for the new options introduced in #5384.

The helper function `__docker_compose_map_key_of_current_option` was copied from Docker's completion.

Also removes duplicate mention of `--filter` from the help message of `docker-compose ps --help`:
```bash
Usage: ps [options] [--filter KEY=VAL] [SERVICE...]

Options:
    -q                   Only display IDs
    --services           Display services
    --filter KEY=VAL     Filter services by a property

```

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -619,7 +619,7 @@ class TopLevelCommand(object):
         """
         List containers.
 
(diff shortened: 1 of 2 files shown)


shin-:
Thanks! 👍

PR #5593, opened 2018-01-19
Support mixed use of TLS flags and TLS environment variables

### LINKED ISSUE #3937: docker-compose complains about invalid TLS configuration
I am running Docker 1.11.2, and so far, everything is fine. The Docker engine is exposed via TCP, TLS is enabled. The certificate is a self-signed one, but so far it has always been working. E.g., if you run

```
$ docker ps
```

everything is fine and the running processes are shown. Now I tried to use docker-compose to start an application. This failed with the following error message:

> TLS configuration is invalid - make sure your DOCKER_TLS_VERIFY and DOCKER_CERT_PATH are set correctly

The same error happens if you provide the parameters `--tlscert` and `--tlskey` manually. The certificate contains the host name, so it should not be related to #3365. I have read multiple times that setting `CURL_CA_BUNDLE` to an empty value should help, but it didn't.

I have experienced this behavior with docker-compose 1.7.0 and 1.8.0.

Any ideas?

### LINKED ISSUE #4651: docker-compose does not have a way to override tls version at the commandline
Using compose, if I set

COMPOSE_TLS_VERSION
DOCKER_TLS_VERIFY
DOCKER_CERT_PATH

docker-compose works fine and gives me no errors.  If I decide to use command line argument instead of these environment variables, I have no way to set which version of TLS I want to use.  For servers that do not support TLS version 1, I get the following errors:

ERROR: SSL error: [SSL: TLSV1_ALERT_PROTOCOL_VERSION] tlsv1 alert protocol version (_ssl.c:590)

Setting the COMPOSE_TLS_VERSION environment variable and using compose with TLS arguments doesn't seem to work.  Compose seems to ignore it if TLS cmdline options are used.

---

docker-compose binary version 1.9.0 and 1.11.2

shin-:
Fixes #4651 
Fixes #3937

### DIFF
--- a/compose/cli/command.py
+++ b/compose/cli/command.py
@@ -35,7 +35,7 @@ def project_from_options(project_dir, options):
         project_name=options.get('--project-name'),
         verbose=options.get('--verbose'),
         host=host,
-        tls_config=tls_config_from_options(options),
+        tls_config=tls_config_from_options(options, environment),
         environment=environment,
         override_dir=options.get('--project-directory'),
     )
--- a/compose/cli/docker_client.py
+++ b/compose/cli/docker_client.py
@@ -2,6 +2,7 @@ from __future__ import absolute_import
 from __future__ import unicode_literals
 
 import logging
+import os.path
 import ssl
 
 from docker import APIClient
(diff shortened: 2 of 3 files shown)


shin-:
Fixes #3937

PR #5596, opened 2018-01-23
Add up flag `--renew-anon-volumes` (shorthand -V)

### LINKED ISSUE #4337: "docker-compose up" not rebuilding container that has an underlying updated image
This one is a bit lengthy to explain, so I have created a [test project to reproduce this behavior with](https://github.com/twelve17/dc-test).  

As I understand it from the `docker compose` `up` documentation, the container should be rebuilt if an image was changed.  However, it seems here that `docker-compose up` is not noticing that a particular image changed, and thus is re-using an existing container which has pointers to an older image. At least that is the closest I can get to a theory.  More information below the version and info.

docker version:
```
Client:
 Version:      1.12.5
 API version:  1.24
 Go version:   go1.6.4
 Git commit:   7392c3b
 Built:        Fri Dec 16 02:42:17 2016
 OS/Arch:      linux/amd64

Server:
 Version:      1.12.5
 API version:  1.24
 Go version:   go1.6.4
 Git commit:   7392c3b
 Built:        Fri Dec 16 02:42:17 2016
 OS/Arch:      linux/amd64
```

docker info:
```
docker info
Containers: 5
 Running: 0
 Paused: 0
 Stopped: 5
Images: 32
Server Version: 1.12.5
Storage Driver: aufs
 Backing Filesystem: extfs
 Dirs: 85
 Dirperm1 Supported: true
Logging Driver: json-file
Cgroup Driver: cgroupfs
Plugins:
 Volume: local
 Network: bridge host null overlay
Swarm: inactive
Runtimes: runc
Default Runtime: runc
Security Options: apparmor seccomp
Kernel Version: 4.4.0-59-generic
Operating System: Ubuntu 16.04.1 LTS
OSType: linux
Architecture: x86_64
CPUs: 4
Total Memory: 7.682 GiB
Debug Mode (client): false
Debug Mode (server): false
Insecure Registries:
 127.0.0.0/8
```
I am including the contents of the `README.md` here to help explain the problem in detail.

# Use Case

- One simple nodejs project with a `Dockerfile`.
- One local NPM dependency used by the above project (copied to container via Dockerfile). The project refers to the dependency via a [local path](https://docs.npmjs.com/files/package.json#local-paths).
- The nodejs project has one web route (`/`) that prints the version of the local npm dependency from its `package.json`.  This is used to verify the results of the test case procedure.
- [The docker-compose.yml file](https://github.com/twelve17/dc-test/blob/master/docker-compose.yml#L23) uses [this volume technique](http://jdlm.info/articles/2016/03/06/lessons-building-node-app-docker.html) to overlay the host machine's source tree on top of the container's source tree and then overlaying the `node_modules` from the container on top of the first volume.  This allows changing of sources on the host machine while at the same time using the node_modules that were built for the container's platform.

## Steps to Reproduce

1. Clone [this repo](https://github.com/twelve17/dc-test).
2. Clean up any previous containers and images related to this repo's project via `docker rm` and `docker rmi`.
3. Check out the `test2_run1` tag. This state represents the project using version 1.0.0 of the local NPM dependency.
4. Do a `docker-compose build`.  All steps should run without any cache usage if step 2 was followed correctly.  Note the version of the local NPM dependency during the `npm install` command, e.g. `+-- my-npm@1.0.0`.
5. Do a `docker-compose up`.  Browse to `http://localhost:8000`.  The page should report version `1.0.0`.
6. Stop the running containers. (Ctrl-C on the terminal from which the `up` command was issued.)
7. Check out the `test2_run2` tag. This introduces [a small change](https://github.com/twelve17/dc-test/compare/test2_run1...test2_run2) to the NPM's `index.js` file, and a version bump in its `package.json` to `1.0.1`.
8. Do a `docker-compose build`.  **Only the instructions up to `COPY ./my-npm ...` should use a cache.** (E.g., the docker output prints `---> Using cache` for that instruction.)  All subsequent steps should be run by docker.  This is because the changes introduced in step 7 to the NPM package should have invalidated the cache for the `COPY ./my-npm ...` command, and, as a result, subsequent steps too.  Confirm that during the `npm install` command, the new version of the NPM is printed in the summary tree output, e.g. `+-- my-npm@1.0.1`.
9. Do a `docker-compose up`.  Browse to `http://localhost:8000`.  The page should report version `1.0.1`.

Expected behavior: Page in step 9 should report `1.0.1`.  That is, a change in the local npm should be reflected in the container via `docker-compose up`.

Actual behavior: Page in step 9 reports `1.0.0`.

**Note that docker itself is re-building images as expected.**  The observed issue is not that docker is re-using a cached image, as the output shows it re-running NPM install and showing the new version of the local NPM dependency.  The issue is that `docker-compose` is not seeing that the underlying images that comprise the `dctest_service1` container have been updated.

In fact, running bash in the container allows us to see that the container has the updated `my-npm` module files, but the `node_modules` version is stale:

```
  # docker exec -it dctest_service1_1 bash
  app@6bf2671b75c6:~/service1$ grep version  my-npm/package.json  node_modules/my-npm/package.json
  my-npm/package.json:  "version": "1.0.1",
  node_modules/my-npm/package.json:  "version": "1.0.0"
  app@6bf2671b75c6:~/service1$
```

Workaround: Use `docker rm` to remove the `dctest_service1` container.  Then re-run `docker-compose up`, which will re-create the container using the existing images.  Notable in this step is that no underlying images are re-built.  In re-creating the container, `docker-compose` seems to figure out to use the newer volume that has the updated `node_modules`.

See the [`output`](https://github.com/twelve17/dc-test/tree/master/output) directory for the output printed during the first run (steps 4 and 5) and the second run (steps 8 and 9).

### LINKED ISSUE #5400: Consider a --force-recreate-volumes flag
If I have an image that defines a volume, the container ends up with an anonymous docker volume. This leads to issues like #4476 where I can't add `volumes_from:` if I've previously forgotten it.

Another use case I've encountered is when you have an image that populates a volume:

```
FROM base/almost_empty:1.0.0

ENV AIRFLOW_HOME /usr/local/airflow
COPY ./foo ${AIRFLOW_HOME}/stuff/foo/

VOLUME ${AIRFLOW_HOME}/stuff
```

In this case, when I update to a newer version of my image, I want a new volume with the latest version of `foo`.

At the moment, I can do `docker-compose down && docker-compose up`, but this recreates every single container in my docker-compose.yml.

Alternatively, I can do `docker-compose rm xyz && docker-compose up`. However assumes that I know which containers to run `docker-compose rm` on.

Would you be willing to add a `--force-recreate-volumes` argument? This would be a huge help for these use cases.

shin-:
to avoid reusing the previous container's data

Fixes #4337 
Fixes #5400

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -936,7 +936,7 @@ class TopLevelCommand(object):
             --always-recreate-deps     Recreate dependent containers.
                                        Incompatible with --no-recreate.
             --no-recreate              If containers already exist, don't recreate
-                                       them. Incompatible with --force-recreate.
+                                       them. Incompatible with --force-recreate and -V.
             --no-build                 Don't build an image, even if it's missing.
             --no-start                 Don't start the services after creating them.
             --build                    Build images before starting containers.
@@ -945,8 +945,10 @@ class TopLevelCommand(object):
             -t, --timeout TIMEOUT      Use this timeout in seconds for container
                                        shutdown when attached or when containers are
                                        already running. (default: 10)
-            --remove-orphans           Remove containers for services not
-                                       defined in the Compose file
+            -V, --renew-anon-volumes   Recreate anonymous volumes instead of retrieving
+                                       data from the previous containers.
+            --remove-orphans           Remove containers for services not defined
+                                       in the Compose file.
             --exit-code-from SERVICE   Return the exit code of the selected service
                                        container. Implies --abort-on-container-exit.
             --scale SERVICE=NUM        Scale SERVICE to NUM instances. Overrides the
@@ -992,6 +994,7 @@ class TopLevelCommand(object):
                     start=not no_start,
                     always_recreate_deps=always_recreate_deps,
                     reset_container_image=rebuild,
+                    renew_anonymous_volumes=options.get('--renew-anon-volumes')
                 )
 
             try:
@@ -1083,10 +1086,14 @@ def compute_exit_code(exit_value_from, attached_containers, cascade_starter, all
 def convergence_strategy_from_opts(options):
     no_recreate = options['--no-recreate']
     force_recreate = options['--force-recreate']
+    renew_anonymous_volumes = options.get('--renew-anon-volumes')
     if force_recreate and no_recreate:
         raise UserError("--force-recreate and --no-recreate cannot be combined.")
 
-    if force_recreate:
+    if no_recreate and renew_anonymous_volumes:
+        raise UserError('--no-recreate and --renew-anon-volumes cannot be combined.')
+
+    if force_recreate or renew_anonymous_volumes:
         return ConvergenceStrategy.always
 
     if no_recreate:
--- a/compose/project.py
+++ b/compose/project.py
@@ -445,7 +445,8 @@ class Project(object):
            rescale=True,
            start=True,
            always_recreate_deps=False,
-           reset_container_image=False):
+           reset_container_image=False,
+           renew_anonymous_volumes=False):
 
         self.initialize()
         if not ignore_orphans:
@@ -474,7 +475,8 @@ class Project(object):
                 rescale=rescale,
                 start=start,
                 project_services=scaled_services,
-                reset_container_image=reset_container_image
+                reset_container_image=reset_container_image,
+                renew_anonymous_volumes=renew_anonymous_volumes,
             )
 
         def get_deps(service):
--- a/compose/service.py
+++ b/compose/service.py
@@ -409,7 +409,8 @@ class Service(object):
 
             return containers
 
-    def _execute_convergence_recreate(self, containers, scale, timeout, detached, start):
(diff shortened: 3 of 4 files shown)


Wilfred:
Thank you very much :)

basz:
nice, will this work with `docker stack deploy` too or just with `docker-compose up`?

shin-:
@basz Just `docker-compose`. `docker stack` issues and requests should be reported on the [docker/cli](https://github.com/docker/cli) repo.

MrBoutte101:
@shin currently waiting for this to be released. Any idea on what release to expect it on?

basz:
It has been released as rc, can be installed manually

MrBoutte101:
Oh okay, yeah I see it's in pre-release. Do you know when the next official release is, by chance?

basz:
nope

PR #5598, opened 2018-01-23
Use mock compatibility import

carlwgeorge:
`tests/__init__.py` already has a try/except statement to use mock from the standard library when possible.  Take advantage of it like other tests already do.


shin-:
LGTM, thanks!

PR #5604, opened 2018-01-24
Add bash completion for `up {--always-recreate-deps,--renew-anon-volumes}`

albers:
This adds support for two new options for `docker-compose up` that will be added in 1.19.0, see #5602.
- `--always-recreate-deps` was added in #5541
- `--renew-anon-volumes|-V` was added in #5596

Please include in 1.19.0 so that features and completion are in sync.


shin-:
Thank you! :100:

PR #5612, opened 2018-01-26
Lower issue severity to "warning" if `down` tries to remove nonexisting image

and800:
It seems that most of the times it's not a real error when compose cannot delete image that doesn't exist.
It just means that the project has been already downed, or nas never been built.

This patch brings consistency with networks and volumes removal.


shin-:
Thanks!

Sounds fine in principle. Could you add a test for it?

and800:
@shin- I will, but a few days later, being a bit busy now

and800:
I'm sorry, forgot about this one. Gonna setup a test and update the branch tonight

and800:
done

PR #5643, opened 2018-02-03
ps show health (resolves #5525)

### LINKED ISSUE #5525: Can docker-compose ps show health status?
The Docker CLI can show health status

```
$ docker container ls --filter name=example_example_1
CONTAINER ID        IMAGE            COMMAND                  CREATED             STATUS                 PORTS                                     NAMES
6e9eef9909f7        example:latest   "nginx -g 'daemon ..."   5 hours ago         Up 5 hours (healthy)   80/tcp, 443/tcp, 0.0.0.0:8080->8080/tcp   example_example_1
```

But, the Docker Compose CLI does not

```
$ docker-compose ps example
         Name            Command          State                       Ports
------------------------------------------------------------------------------------------------
example_example_1   nginx -g daemon off;   Up      443/tcp, 80/tcp, 0.0.0.0:8080->8080/tcp
```

It might be nice to display the same health status, especially for those of us using Compose 2.1 and the depends_on / condition / service_healthy feature.

dihmuzikien:
Add health string to `docker-compose ps` output. The command behaves like Docker CLI For starting `Up (health: starting)`
For no health check `Up`
For healthy or unhealthy `Up (healthy/unhealthy)`

### DIFF
--- a/compose/container.py
+++ b/compose/container.py
@@ -129,7 +129,7 @@ class Container(object):
         if self.is_restarting:
             return 'Restarting'
         if self.is_running:
-            return 'Ghost' if self.get('State.Ghost') else 'Up'
+            return 'Ghost' if self.get('State.Ghost') else self.human_readable_health_status
         else:
             return 'Exit %s' % self.get('State.ExitCode')
 
@@ -171,6 +171,18 @@ class Container(object):
     def has_api_logs(self):
         log_type = self.log_driver
         return not log_type or log_type in ('json-file', 'journald')
+
+    @property
+    def human_readable_health_status(self):
+        """ Generate UP status string with up time and health
+        """
+        status_string = 'Up'
+        container_status = self.get('State.Health.Status')
+        if container_status == 'starting':
+            status_string += ' (health: starting)'
+        elif container_status is not None:
+            status_string += ' (%s)' % container_status
+        return status_string
 
     def attach_log_stream(self):
         """A log stream can only be attached if the container uses a json-file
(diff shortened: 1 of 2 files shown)


jrd:
Tested today, works very good with or without containers that have a healthcheck status

compose/container.py
```diff
@@ -172,6 +172,19 @@ def has_api_logs(self):
         log_type = self.log_driver
         return not log_type or log_type in ('json-file', 'journald')
 
+    @property
+    def create_health_status_string(self):
```
shin-: If a method is marked with an `@property` decorator, it should have a property name instead of an action name. To be consistent with other similar properties, maybe rename this to `human_readable_health_status`?

compose/container.py
```diff
@@ -172,6 +172,19 @@ def has_api_logs(self):
...
+        """ Generate UP status string with up time and health
+        """
+        status_string = 'Up'
+        if 'Health' in self.get('State'):
+            container_status = self.get('State.Health.Status')
```
shin-: Because `get` will return `None` if the key is undefined, you can simplify this part of the code by writing
```python
container_status = self.get('State.Health.Status')
if container_status == 'starting':
  ...
elif container_status is not None:
  ...
```

dihmuzikien:
@shin- I made changes as requested. But I haven't able to figure out how to squash my commits (sorry,, still a newbie).

shin-:
@dihmuzikien Thank you! I'll take care of rebasing / squashing things together.

shin-:
Merged in bb16d9e951afbd62a9f22d02b2eb613c1ac3ef4c

Thank you!

Issue #5545, opened 2018-01-08
Implement pytest fixtures


waynr:
As a developer of python microservices I often find myself implementing pytest fixtures that yield container objects a la https://github.com/docker/docker-py . Often times what I would really prefer to do is access information about services/containers from within tests to provide access to my service running as a container that has access to other containers in the same network.

The [pytest-docker](https://github.com/AndreLouisCaron/pytest-docker/) library does this to some extent but it's current implementation and distance from `docker-compose` makes me think that it may not provide a stable long term interface for doing this. In AndreLouisCaron/pytest-docker#20 I have suggested to that author to consider porting his code to this repository, where I think there is a better opportunity to avoid breaking changes as new versions of  `docker-compose` are released in the future by writing tests that run here to ensure such such api changes necessitate changes in the docker pytest fixtures.

waynr:
So to be clear, I am posting this issue not so much as a suggestion that someone else implement it but to see how open project maintainers would be to a contribution of this nature.

waynr:
Okay so after peeking around at the docker-compose codebase a little I'm starting to think that it may not be necessary to provide pytest fixtures. Instead it looks like I can make use of `compose.config.config.load(...)` and `compose.project.Project` to achieve my end goal which is to make use of a set of services defined in a compose yaml file and manage their up/down lifecycles using pytest fixture scopes. Does that sound like a reasonable approach that's likely not to be borked by api changes?

shin-:
Hi @waynr,

[Our (unpopular) stance](https://github.com/docker/compose/issues/4542#issuecomment-283191533) on the matter is that we do not make any guarantees about the stability of the Compose API, as the only usecase we officially support is through the CLI. For the same reason, I would have to refuse a contribution of that nature to the repository that would suggest implicit support / guarantees.

Thank you for your understanding!

waynr:
Thanks for your prompt response @shin-. It looks to me like the API is already tested by virtue of its use in various test cases, but it's unfortunate that the project won't consider supporting a for realz documented and versioned API. I do understand that this would signfiicantly complicate your teams' efforts when it comes to supporting the main product -- the command line tool. But can I be blamed  yearning for programmatic access to such a great tool?? ;)

[...]

PR #5655, opened 2018-02-07
Pass-through keyword arguments to Docker client for image push and pull

lexsca:
Allow pushing to and pulling from private image repositories without having to store credentials in a config file. The Docker client already allows this via auth_config parameter. By passing through extra keyword arguments, new parameters can also be used when they become available.

This doesn't affect the command-line interface.  It only adds to the Python API.

### DIFF
--- a/compose/project.py
+++ b/compose/project.py
@@ -537,12 +537,13 @@ class Project(object):
 
         return plans
 
-    def pull(self, service_names=None, ignore_pull_failures=False, parallel_pull=False, silent=False):
+    def pull(self, service_names=None, ignore_pull_failures=False,
+             parallel_pull=False, silent=False, **kwargs):
         services = self.get_services(service_names, include_deps=False)
 
         if parallel_pull:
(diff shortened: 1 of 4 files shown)


shin-:
Hi @lexsca, thank you for your contribution! Unfortunately, I'm going to have to close this for the reasons I explained [here](https://github.com/docker/compose/issues/5545#issuecomment-356100870). People looking to leverage Compose's  interface programmatically should do so through a fork of the project rather than pull requests.

Thank you for your understanding!

lexsca:
No worries.  I didn't see that previously.  Totally understand.  Might be helpful to post something to that effect on the [top-level readme](https://github.com/docker/compose/master/README.md).  Having docker-compose [published on PyPI](https://pypi.python.org/pypi/docker-compose) led me to believe using the underlying Python API was a reasonable path.

shin-:
Fair point, I'll see if there's a good place where it makes sense to mention it. :+1:

Issue #4994, opened 2017-07-05
Docker secrets compatibility issue


hiagop:
I'm trying to deploy an IMAP server using docker-compose and secrets for the sensitive data management. I followed the [documentation for the long syntax form of creating secrets](https://docs.docker.com/compose/compose-file/#long-syntax-2). But anytime I do `docker-compose up -d` I get this warning:

> WARNING: Service "imap" uses secret "key" with uid, gid, or mode. These fields are not supported by this implementation of the Compose file
> WARNING: Service "imap" uses secret "cert" with uid, gid, or mode. These fields are not supported by this implementation of the Compose file

This actually doesn't prevent the containers from being deployed though. But none of the properties set in the docker-compose.yml file are applied to the generated files inside the container.

I have already checked for the [compatibility table](https://docs.docker.com/compose/compose-file/#compose-and-docker-compatibility-matrix), but according to it, I should not be getting this warning, since I'm running my containers with the following docker engine and docker-compose versions:

`$ docker version`:

> Client:
>  Version:      17.06.0-ce
>  API version:  1.30
>  Go version:   go1.8.3
>  Git commit:   02c1d87
>  Built:        Fri Jun 23 21:20:04 2017
>  OS/Arch:      linux/amd64
> 
> Server:
>  Version:      17.06.0-ce
>  API version:  1.30 (minimum version 1.12)
>  Go version:   go1.8.3
>  Git commit:   02c1d87
>  Built:        Fri Jun 23 21:18:59 2017
>  OS/Arch:      linux/amd64
>  Experimental: false

`$ docker-compose version`:

> docker-compose version 1.14.0, build c7bdf9e
> docker-py version: 2.3.0
> CPython version: 2.7.9
> OpenSSL version: OpenSSL 1.0.1t  3 May 2016

-------------------
This is my **docker-compose.yml** file:

```
version: '3.1'
services:
    imap:
        image: cyrus
        hostname: cyrus
        volumes:
            - volume1
            - volume2
            - volume3
            - volume4
        secrets:
            - source: key
              target: my.key
              uid: '0'
              gid: '109'
              mode: 0640
            - source: cert
              target: my.crt
              uid: '0'
              gid: '109'
              mode: 0640
        ports:
            - 110:110
            - 143:143
            - 993:993
            - 995:995
            - 4190:4190
        restart: always
        networks:
            - mail
    logger:
        image: gliderlabs/logspout
        hostname: logspout
        volumes:
[...]

PR #5684, opened 2018-02-17
Compatibility mode

### LINKED ISSUE #4513: How to specify Memory & CPU limit in version 3
I am unable to specify CPU & memory for services specified in version 3 . 

With version 2 it works fine with "mem_limit" & "cpu_shares" parameters under the services . But it fails while using version 3 , putting them under deploy section doesn't seem worthy unless i am using swarm mode . 

Can somebody help ? 

    version: "3"
    services:
      node:
        build:
          context: .
          dockerfile: ./docker-build/Dockerfile.node
        restart: always
        environment:
          - VIRTUAL_HOST=localhost
        volumes:
          - logs:/app/out/
        expose:
          - 8083
        command: ["npm","start"]
        cap_drop:
          - NET_ADMIN
          - SYS_ADMIN

shin-:
Using the `--compatibility` CLI flag, v3 files can be converted to their v2 equivalent, with `deploy` options translated when possible. `--compatibility` can be used with all commands.

The conversion is a "best effort" attempt and shouldn't be relied on for production deployments.

Fixes #4513

### DIFF
--- a/compose/config/config.py
+++ b/compose/config/config.py
@@ -896,6 +896,7 @@ def translate_resource_keys_to_container_config(resources_dict, service_dict):
         service_dict['mem_reservation'] = resources_dict['reservations'].get('memory')
         if 'cpus' in resources_dict['reservations']:
             return ['resources.reservations.cpus']
+    return []
 
 
 def convert_restart_policy(name):
@@ -919,7 +920,7 @@ def translate_deploy_keys_to_container_config(service_dict):
         if k in deploy_dict
     ]
 
-    if 'replicas' in deploy_dict and deploy_dict.get('mode') == 'replicated':
+    if 'replicas' in deploy_dict and deploy_dict.get('mode', 'replicated') == 'replicated':
         service_dict['scale'] = deploy_dict['replicas']
 
     if 'restart_policy' in deploy_dict:
--- a/compose/config/config.py
+++ b/compose/config/config.py
@@ -16,6 +16,7 @@ from . import types
 from .. import const
 from ..const import COMPOSEFILE_V1 as V1
 from ..const import COMPOSEFILE_V2_1 as V2_1
+from ..const import COMPOSEFILE_V2_3 as V2_3
 from ..const import COMPOSEFILE_V3_0 as V3_0
 from ..const import COMPOSEFILE_V3_4 as V3_4
 from ..utils import build_string_dict
@@ -341,7 +342,7 @@ def find_candidates_in_parent_dirs(filenames, path):
     return (candidates, path)
 
 
-def check_swarm_only_config(service_dicts):
+def check_swarm_only_config(service_dicts, compatibility=False):
     warning_template = (
         "Some services ({services}) use the '{key}' key, which will be ignored. "
         "Compose does not support '{key}' configuration - use "
@@ -357,13 +358,13 @@ def check_swarm_only_config(service_dicts):
                     key=key
                 )
             )
-
-    check_swarm_only_key(service_dicts, 'deploy')
+    if not compatibility:
+        check_swarm_only_key(service_dicts, 'deploy')
     check_swarm_only_key(service_dicts, 'credential_spec')
     check_swarm_only_key(service_dicts, 'configs')
 
 
-def load(config_details):
+def load(config_details, compatibility=False):
     """Load the configuration from a working directory and a list of
     configuration files.  Files are loaded in order, and merged on top
     of each other to create the final configuration.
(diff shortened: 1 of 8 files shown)


shin-:
cc @mefyl @vdemeester for review :eyeglasses:

thaJeztah:
But doesn’t the V3 format already support memory and cpu limits? ie;

```yaml
deploy:
      resources:
        limits:
          cpus: '0.50'
          memory: 50M
        reservations:
          cpus: '0.25'
          memory: 20M

**Comment by thaJeztah** [comment]
Oh; they were ignored by docker compose? Sorry, missed that

**Comment by shin-** [comment]
@thaJeztah Yes, the point here is to have a way to account for them in Compose as well. It's behind a flag because we can't really guarantee the behavior will match exactly what you would see with `stack deploy`. Does that make sense?

**Comment by dnephin** [comment]
Would it make sense make this an option on `docker-compose config` for converting a config from v3 to v2? That way you could inspect the changes to have some idea of what you're running.

**Comment by shin-** [comment]
@dnephin it works with `config` as well - it outputs a valid 2.3 file.

**Comment by thaJeztah** [comment]
> It's behind a flag because we can't really guarantee the behavior will match exactly what you would see with stack deploy. Does that make sense?

Wondering; things like `configs` and `secrets` already have a different implementation in docker-compose; if we document that there are differences, would it make sense to enable it by default (possibly an option to _disable_ it)?

**Comment by shin-** [comment]
I want to say no - the secrets emulation is super confusing to users as it is already (see feedback in https://github.com/docker/compose/issues/4994 for example). I think having users opt into the feature gives us a chance to setup expectations a little better.

**Comment by thaJeztah** [comment]
Ah, yes, then 👍 on the flag

**Comment by thiagolsfortunato** [comment]
When I executed docker-compose --compatibility config` return this message:

`WARNING: The following deploy sub-keys are not supported in compatibility mode and have been ignored: resources.reservations.cpus`

```
deploy:
      resources:
        limits:
          memory: 1G
          cpus: '1'
        reservations:
          memory: 512M
          cpus: '0.50'
```

Why flag `--compatibility` does not support resources.reservations.cpus and resources.limits.cpus ?!

**Comment by thaJeztah** [comment]
> Why flag `--compatibility` does not support `resources.reservations.cpus` and `resources.limits.cpus` equals memory ?!

Those options are scheduling options for swarm-mode services (i.e., they don't actually reserve the resources, only make swarm take those into account when scheduling containers on a node).

Docker Compose doesn't work with swarm mode (it deploys local containers), so there's no orchestrator to take those limits into account

**Comment by thiagolsfortunato** [comment]
For not swarm-mode, how can I manager memory and cpu resources ?

**Comment by thaJeztah** [comment]
Using the feature described here; the warning that's printed is about cpu _reservation_;

Using this compose file:

```yaml
version: '3.5'
services:
  web:
    image: nginx:alpine
    deploy:
      resources:
        limits:
          cpus: '0.50'
          memory: 50M
```


```
docker-compose --compatibility up -d
WARNING: The following deploy sub-keys are not supported in compatibility mode and have been ignored: resources.reservations.cpus
Creating network "compat_default" with the default driver
Creating compat_web_1 ... done
```

```
docker inspect compat_web_1

...
    "Memory": 52428800,
    "NanoCpus": 500000000,
...
```

compose/config/config.py
```diff
@@ -868,10 +873,80 @@ def finalize_service(service_config, service_names, version, environment):
...
     return normalize_v1_service_format(service_dict)
 
 
+def translate_resource_keys_to_container_config(resources_dict, service_dict):
+    if 'limits' in resources_dict:
+        service_dict['mem_limit'] = resources_dict['limits'].get('memory')
+        if 'cpus' in resources_dict['limits']:
+            service_dict['cpus'] = float(resources_dict['limits']['cpus'])
+    if 'reservations' in resources_dict:
+        service_dict['mem_reservation'] = resources_dict['reservations'].get('memory')
```
thaJeztah: @shin- I just noticed this; this part isn't correct, because `--memory-reservation` (although named similar) is something very different than `--reserve-memory` on `docker service`; see https://github.com/moby/moby/pull/14579

samson-sham:
Sorry, I have been searching a way to set memory limit in **non-swarm** mode with compose config v3.5, and I am reading back & forth amongst:
#4513, #5803, #4994 
also with documentations:
https://docs.docker.com/compose/compose-file/#resources

The reason why I have to use v3.5 is because I had setup the networks between containers, that worked perfect.
But I want to test a specific container with memory shortage, and I find no way to configure this with v3.

```
version: "3.5"

services:
    mq:
      image: rabbitmq:3.7.11-management-alpine
      hostname: amqphost
      ports:
        - "5672:5672"
        - "15672:15672"
      restart: always
      networks:
        - mqnet
      # mem_limit: 200m
      # mem_reservation: 128m
      # deploy:
      #   resources:
      #     limits:
      #       memory: 200m
      #     reservations:
      #       memory: 128m
networks:
  mqnet:
    name: mqnet
    driver: bridge
```

So here is my docker-compose, sorry I have no idea to do networking with swarm, and I have also tried with `--compatibility` , which obviously doesn't support `mem_limit` & `mem_reservation` .
I hate to ask this, but what should I do?

ZacharyBacon:
https://blog.csdn.net/ZacharyBacon/article/details/121739675 为何不通过cgroup限制资源，现在只是支持swarm模式下，如果是非swarm下，可以通过挂载cgroup的文件来限制容器的资源。感兴趣的话，可以看下我写的这篇文章。

lordraiden:
So now "limits" cpu/memory can be used with compatibility mode in a v3 file
But "cpu_shares", "cpu_quota", "cpuset", "mem_limit" can not be used unless you use a v2 file
right?

thaJeztah:
@lordraiden this is a 6 year old pull request; docker compose v2 now uses the compose-specification file format which does not use a version, and unifies the v2 and v3 features; docs are located here; https://docs.docker.com/compose/compose-file/

PR #5701, opened 2018-02-22
Add support for seccomp files

### LINKED ISSUE #5574: security_opt and seccomp
I'd like to make a `docker-compose.yml` containing something like this:

```yaml
services:
  desktop:
    security_opt:
      - seccomp:"./chrome.json"
```

This would correspond to the `docker run` option `--security-opt seccomp:./chrome.json`, where [`chrome.json` is this seccomp profile](https://raw.githubusercontent.com/jfrazelle/dotfiles/master/etc/docker/seccomp/chrome.json).

Is there any way to give the `seccomp` key an arbitrary JSON string?

shin-:
Fixes #5574

### DIFF
--- a/compose/config/types.py
+++ b/compose/config/types.py
@@ -4,6 +4,7 @@ Types for objects parsed from the configuration.
 from __future__ import absolute_import
 from __future__ import unicode_literals
 
+import json
 import ntpath
 import os
 import re
@@ -13,6 +14,7 @@ import six
 from docker.utils.ports import build_port_bindings
 
 from ..const import COMPOSEFILE_V1 as V1
+from ..utils import unquote_path
 from .errors import ConfigurationError
 from compose.const import IS_WINDOWS_PLATFORM
 from compose.utils import splitdrive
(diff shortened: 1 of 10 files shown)


compose/service.py
```diff
@@ -881,6 +881,11 @@ def _get_container_host_config(self, override_options, one_off=False):
             init_path = options.get('init')
             options['init'] = True
 
+        print(options.get('security_opt'))
```
mnottale: looks like a leftover debug statement

shin-: oops! Thank you for catching it!

compose/config/types.py
```diff
@@ -457,3 +459,30 @@ def normalize_port_dict(port):
...
+    @classmethod
+    def parse(cls, value):
+        # based on https://github.com/docker/cli/blob/9de1b162f/cli/command/container/opts.go#L673-L697
+        con = value.split('=', 2)
+        if len(con) == 1 and con[0] != 'no-new-privileges':
+            if ':' not in value:
+                raise ConfigurationError('Invalid security_opt: {}'.format(value))
+            con = value.split(':', 2)
+
+        if con[0] == 'seccomp' and con[1] != 'unconfined':
```
sprohaska: Previously, it was possible to pass a seccomp profile by embedding the JSON-encoded profile into the docker-copmose file. This won't work anymore, because every value other than `unconfined` is now interpreted as a path. Maybe add `and not con[1].startswith('{')` here to leave alone values that look like JSON.

Issue #5590, opened 2018-01-19
Docker-compose build writes by default "Building ..." to STDERR.


paulvanbladel:
Docker-compose build writes by default "Building _MyImage_" to STDERR where MyImage is the image I'm building.

That makes integration with a build process quite difficult, since I want to capture real errors.
Any workaround possible ?

shin-:
You should use exit codes to detect errors.

paulvanbladel:
Yes but why is something written to STDERR when exit code = 0 ?
When integrating with build and release tooling, StdErr is sometimes interpreted by the tooling.

shin-:
We use `stderr` to separate Compose status output from application (or in this case, build) output.
It certainly seems unintuitive for said tooling to assume an error is occurring based solely on message presence on `stderr` . Maybe it's configurable?

paulvanbladel:
Joffrey,
Thanks a lot for your continuous support.
Indeed it's configurable, but since a build/release process should always stop on first error it's quite convenient to just rely on error output and you can just call the docker commands without any further error exit ceremony.
So, in your scenario, presuming that your run a bash script, are you for each docker command that's involved doing something like:
```
docker this or that
OUT=$?
if [ $OUT -eq 0 ];then
echo "ok"
else
exit $OUT
fi
```
For me it sounds really irrational that something is dropped in error output when there is no error. 
I would use standard output for that, no?

shin-:
@paulvanbladel Would being able to configure the log level of the `docker-compose` CLI using a flag or environment variable be a satisfactory solution?

paulvanbladel:
Joffrey,
Not sure. I just noticed that also docker login writes to STDERR just for providing a warning:
```
WARNING! Using --password via the CLI is insecure******** Use --password-stdin********
```
Not sure if the introduction of a log level would change anything.

sladg:
I have come across this as well.
I'm using CI to deploy automatically. Problem is: If I turn off ,,Error on STDERR" I can't detect errors from other application. If I leave it ON, Docker-compose fails.

I genuinely think that using STDERR for warnings or separation of application and composer is a bad solution.

Warning is not error and information message is not an error as well, so why are we using STDERR?

sonuame:
Please fix this. Its making all my pipelines to fail

PR #5702, opened 2018-02-22
Add '--quiet' option to 'up' to pull silently.

mnottale:


shin-:
If this only applies to the `pull` part of the operation, maybe the option should be named less ambiguously, like `--quiet-pull`?

mnottale:
Renamed the option per your request.

shin-:
Fixes #5590

Issue #3447, opened 2016-05-11
Add --quiet or --silent option to docker-compose


dkirrane:
I'm trying to assign the output of `docker-compose run` to a variable in bash.

Currently docker-compose run can print to console stuff like:

```
docker-compose run serviceabc ls
WARNING: The MVN_OPT variable is not set. Defaulting to a blank string.
Starting serviceabc
dir1
dir2
dir3
```

Can a `-s` `--silent` or `-q` `--quiet` option be added to `docker-compose run` avoid printing anything other than the command output.

I guess this would also make sense for other `docker-compose` commands too.

shin-:
You can redirect stderr to `/dev/null` (i.e. `2>/dev/null`), which should have that exact effect.

elliotchance:
@shin- The quiet option is different from just sending all output nowhere, perhaps it's better explained as "don't show progress bars".

There should be an option that still shows a build steps, but not the actual progress bars which spray a huge amount of output on CI servers since the docker commands expect that your using a terminal that has the control codes to write over previous lines.

shin-:
@elliotchance That seems to be a different request from the one made by the original poster. Can you create a separate issue for it?

elliotchance:
@shin- thanks for letting me know :) #3770

jbarciauskas:
I'd specifically be interested in an option to quiet the download progress update, as it pollutes the travis logs

lelit:
Redirecting stderr works, but potentially hides real errors too, so I think adding --quiet would be better.

RealSalmon:
+1

I would find this useful as well. Redirecting stderr is an *OK* workaround, but it has caused me some minor frustration as it hides actual errors.

Bamieh:
do we have any updates on this issue?

jbarciauskas:
This should have been fixed and released in 1.15? https://github.com/docker/compose/pull/4730

lelit:
Only for `pull`, AFAICT.

shin-:
Fixed with the addition of `--log-level` in #5703

charlesreid1:
Note that `--log-level` should come before `up`, so up doesn't steal the flag, e.g.,

```
$ docker-compose --log-level ERROR up -d --build
```

Vanav:
It is not fixed:
```
# docker-compose --log-level CRITICAL restart container

Restarting compose_container_1 ... done
```

iabotsiTC:
[...]

PR #5703, opened 2018-02-22
New --log-level option.

mnottale:

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -100,7 +100,10 @@ def dispatch():
         {'options_first': True, 'version': get_version_info('compose')})
 
     options, handler, command_options = dispatcher.parse(sys.argv[1:])
-    setup_console_handler(console_handler, options.get('--verbose'), options.get('--no-ansi'))
+    setup_console_handler(console_handler,
+                          options.get('--verbose'),
+                          options.get('--no-ansi'),
+                          options.get("--log-level"))
     setup_parallel_logger(options.get('--no-ansi'))
     if options.get('--no-ansi'):
         command_options['--no-color'] = True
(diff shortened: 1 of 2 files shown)


compose/cli/main.py
```diff
@@ -139,18 +139,35 @@ def setup_parallel_logger(noansi):
...
+          'DEBUG': logging.DEBUG ,
+          'INFO': logging.INFO,
+          'WARNING': logging.WARNING,
+          'ERROR': logging.ERROR,
+          'CRITICAL': logging.CRITICAL,
+        }
+        loglevel = levels.get(level)
+        if loglevel is None:
+            try:
+                loglevel = int(level)
```
shin-: I think it'll be easier to document and easier to use if we only support string values.

compose/cli/main.py
```diff
@@ -139,18 +139,35 @@ def setup_parallel_logger(noansi):
...
+
+    if level is not None:
+        levels = {
+          'DEBUG': logging.DEBUG ,
+          'INFO': logging.INFO,
+          'WARNING': logging.WARNING,
+          'ERROR': logging.ERROR,
+          'CRITICAL': logging.CRITICAL,
+        }
+        loglevel = levels.get(level)
```
shin-: maybe we can do `level.upper()` here - not much reason to be case-sensitive.

shin-:
Fixes #5590 
Fixes #3447

compose/cli/main.py
```diff
@@ -100,7 +100,10 @@ def dispatch():
         {'options_first': True, 'version': get_version_info('compose')})
 
     options, handler, command_options = dispatcher.parse(sys.argv[1:])
-    setup_console_handler(console_handler, options.get('--verbose'), options.get('--no-ansi'))
+    setup_console_handler(console_handler,
+                          options.get('--verbose'),
+                          options.get('--no-ansi'),
+                          options.get("--log-level"))
```
ebuildy: What about adding also an env. variable here? (`COMPOSE_LOG_LEVEL`)

PR #5713, opened 2018-02-25
Update docker-compose config example in readme to use version 3

jgrasser:


shin-:
Thanks, but we'd like to keep it at 2.

PR #4811, opened 2017-05-10
'run' command can use network aliases for service

geigerzaehler:
This PR adds an `--use-alias` option to the `run` command that will add the service network aliases to the container. This is something request in #3492.

PR #5725, opened 2018-02-27
Add --use-aliases flag to run command

### LINKED ISSUE #3492: 1.7.1 doesn't set network aliases when using `run` command
I noticed what appears to be a regression upon upgrading from 1.7.0 to 1.7.1, where the network "aliases" are not defined when running `docker-compose run`. 

This is the simplest docker-compose.yml configuration I could create to illustrate the difference:

``` yaml
version: '2'

services:
  test:
    image: busybox
    networks:
      default:
        aliases:
          - busy
    command: sh

```

When running `docker-compose run test` on 1.7.1 the "Aliases" key is empty:

``` sh
> docker-compose -v
> docker-compose run test
> docker inspect --format '{{.NetworkSettings.Networks.composetest_default.Aliases}}' $(docker-compose ps -q)
[]
```

On 1.7.0 I see the defined alias:

``` sh
> docker-compose -v
> docker-compose run test
> docker inspect --format '{{.NetworkSettings.Networks.composetest_default.Aliases}}' $(docker-compose ps -q)
[busy]
```

I also tried using a defined network, but Aliases weren't created there either under 1.7.1. Aliases _are_ assigned as expected using `docker-compose up -d test`.

Is the 1.7.1 behavior expected, or a bug? In my case the new behavior is breaking a selenium test suite that I have created, which has two services which need bidirectional communication. My solution (under 1.7.0) was to `link` my "app" service to the "selenium" service, and define an `alias` for the app service (on the `default` network). Then I `docker-compose run` the "app" service in order to run a series of tests (I can attach the docker-compose.yml file for the actual service if necessary but it uses some proprietary images). Under 1.7.0 the containers can resolve each other using the defined aliases, but under 1.7.1 only "app" can see "selenium", but not the other way around.

shin-:
Rebase #4811 
Closes #4811 
Fixes #3492


NdubisiOnuora:
@shin-, which version will this be in? I can't see this on https://docs.docker.com/compose/reference/run/.

Issue #5627, opened 2018-01-30
docker-compose cli command eat previous lines from the console output


bscheshirwork:
```sh
$docker-compose -f ~/docker-codeception-run/docker-compose.yml up -d
```
will rewrite existing output by produced output
![peek 2018-01-30 16-51](https://user-images.githubusercontent.com/5769211/35569799-3c577736-05de-11e8-95b9-3ca488e272df.gif)

shin-:
Thanks for the report!

ghostsquad:
what version of compose?

bscheshirwork:
this happens after update to https://github.com/docker/compose/releases/tag/1.19.0-rc2.

bscheshirwork:
`docker-compose version 1.19.0-rc3, build cdd0282`
pacman still alive!

Greek64:
Ha, yeah. Just wanted to search if an issue about this already exists. +1

Even if it's just cosmetic, it is driving me crazy! (only happens on up command)

ghostsquad:
@Greek64 Ya, I thought it was really weird that up & down have different behavior. It seems they should be using the same code paths, and honestly that's probably the fix.

bscheshirwork:
So... I try to detect commit (I am not a python programmer)

The eater was called between versions 1.17.0 and 1.18.0 (see half check) but error is older: after add colors (and status) the `docker-compose up` process draw 2 line.

```sh
dev@dev-Aspire-V3-772:~$ curl -L https://github.com/docker/compose/releases/download/1.11.2/docker-compose-`uname -s`-`uname -m` > /usr/local/bin/docker-compose
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   617    0   617    0     0    749      0 --:--:-- --:--:-- --:--:--   749
100 8066k  100 8066k    0     0  1583k      0  0:00:05  0:00:05 --:--:-- 2050k
dev@dev-Aspire-V3-772:~$ chmod +x /usr/local/bin/docker-compose
dev@dev-Aspire-V3-772:~$ docker-compose -f ~/projects/docker-yii2-app-advanced/docker-codeception-run/docker-compose.yml down
Stopping dockercodeceptionrun_nginx_1 ... done
Stopping dockercodeceptionrun_php_1 ... done
Stopping dockercodeceptionrun_browser_1 ... done
Stopping dockercodeceptionrun_db_1 ... done
[...]

PR #5735, opened 2018-03-01
Fix a race condition in ParallelStreamWriter.

mnottale:

### DIFF
--- a/compose/parallel.py
+++ b/compose/parallel.py
@@ -4,6 +4,7 @@ from __future__ import unicode_literals
 import logging
 import operator
 import sys
+from threading import Lock
 from threading import Semaphore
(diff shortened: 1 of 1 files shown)


mnottale:
This likely fixes #5627 which I could not reproduce (tried 3 different terminals). But a race condition caused by multiple parallel calls to _write_ansi is consistent with the symptoms.

shin-:
Thanks :+1:

bscheshirwork:
@mnottale we can try to use my composition for reproduce it 

https://github.com/bscheshirwork/docker-yii2-app-advanced-rbac/blob/master/docker-codeception-run/docker-compose.yml

PR #5751, opened 2018-03-06
Add support for 'cpu_period' for compose v2.1-v2.3.

mnottale:


shin-:
Initial pass looks good. I think we want to call [`parse_nanoseconds_int`](https://github.com/docker/compose/blob/master/compose/utils.py#L122-L126) on the value if it's a string during config processing ; we can make that happen in `interpolation.py`

PR #5752, opened 2018-03-06
Revamp ParallelStreamWriter to fix display issues.

mnottale:
Previous code was failing in the presence of dependencies between services, and when doing 'up' on services in the stopped state.

### DIFF
--- a/tests/unit/parallel_test.py
+++ b/tests/unit/parallel_test.py
@@ -143,6 +143,7 @@ class ParallelTest(unittest.TestCase):
 
 
 def test_parallel_execute_alignment(capsys):
+    ParallelStreamWriter.instance = None
     results, errors = parallel_execute(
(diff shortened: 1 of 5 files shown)


mnottale:
Really fixes #5627

tests/unit/parallel_test.py
```diff
@@ -148,6 +148,7 @@ def test_parallel_execute_alignment(capsys):
         func=lambda x: x,
         get_name=six.text_type,
         msg="Aligning",
+        use_existing_writer=False
```
shin-: Is this parameter only used for tests? If so, wouldn't it make more sense to get rid of it and reset `ParallelStreamWriter.instance` to `None` in the `setUp` method?

ofek:
Great job!

PR #5755, opened 2018-03-06
Add new maintainers and move inactive maintainers to alumni

shin-:
cc @mefyl @mnottale Please check that your info is correct.

Moved the OGs to the alumni group as per https://github.com/docker/opensource/pull/48

cc @thaJeztah


mefyl:
LGTM

PR #5762, opened 2018-03-08
Add '--timestamps' option to 'up'.

mnottale:

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -997,6 +997,7 @@ class TopLevelCommand(object):
                                        container. Implies --abort-on-container-exit.
             --scale SERVICE=NUM        Scale SERVICE to NUM instances. Overrides the
                                        `scale` setting in the Compose file if present.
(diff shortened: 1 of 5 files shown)


mnottale:
Unfortunately the "attach" endpoint does not support timestamps, so we have to use "logs", which only works on started containers, hence the added complexity.

shin-:
Hm, that's concerning to me. I feel like we're going to have to deal with a lot of inconsistent results between regular `up` and `up` with  `--timestamps` and that's going to be a headache to maintain. The feature might not be worth the cost down the line.

mnottale:
Hmm, would it be worth it to add a `timestamps` options to `attach`? That seems like an overlook.

ijc:
> Hmm, would it be worth it to add a timestamps options to attach? That seems like an overlook.

Probably it would have been, but I know you've moved on to other things so I'm going to take a chance on closing this now -- if I was wrong and you are still interested please do reopen though.

Issue #1769, opened 2015-07-27
"docker-compose run" should support `--volumes-from`


jamshid:
Would be nice if `docker-compose run` supported all `docker run` features like `--volumes-from`. 

E.g. I have a service defined in my docker-compose.yml that can backup a file system. I want to use it with arbitrary containers, not a static list of containers specified in `docker-compose.yml`'s `volumes_from`.

```
$ docker-compose --version
docker-compose version: 1.3.2
```

Sorry this might be a dup of https://github.com/docker/compose/issues/136, but that issue got closed with a fix that doesn't seem to affect `docker-compose run`.

gioppoluca:
+1
Having compose able to execute the whole set of docker run options is fundamental to be able to reproduce the same things done by hand but using the docker-compose.yml

christianbundy:
:+1: Just banged my head against the wall because of this for the better part of an hour.

ifeatu:
+1 glad I just found this thread.  I think it just saved me some heartache.

anibali:
+1
I would like to see volumes as well as volumes-from

johnjelinek:
is it possible to create a new data volume for every new service you spin up?

For examples, I would like docker-compose to spin up 5 haproxy containers, one for each node in my docker swarm. Then I would like each one to attach to it's own data container.

christianbundy:
@johnjelinek yeah, but the only way I know of doing that is creating five different services is `docker-compose.yml` with different volume paths.

johnjelinek:
@christianbundy: why different volume paths? It seems that each data container should be able to handle the same path, right? Just different data in each data container.

christianbundy:
Sorry, I meant different volume paths on the host, not different volume paths in the container. If you used the same volume paths on the host they'd end up sharing data.

johnjelinek:
how would they share data? I only want one data container mount per haproxy container.

gittycat:
I just ran into this limitation. I expected `docker-compose run` to allow us to define volumes to mount, just as `docker run` does.

dobbs:
You can use environment variables in `docker-compose.yml` to achieve the same effect.

docker-compose.yml:
```yaml
bar:
  image: ubuntu:trusty
  volumes:
    - /tmp
  command: ["touch", "/tmp/baz"]
    
foo:
  image: ubuntu:trusty
  volumes_from:
    - "$VOLUMES_FROM"
```

Then:
```bash
$ export VOLUMES_FROM=bar
[...]

PR #5774, opened 2018-03-12
run: New option '--volumes-from'.

mnottale:


simonvanderveldt:
@mnottale Thx for this PR, could really use this functionality :)
Small question: Does it also work for volumes from containers not defined in the compose file?

shin-:
Moving out of 1.21.0 milestone while we figure out some stuff internally.

simonvanderveldt:
Has there been any progress on this?

Btw this fixes #1769 and would be the last item on the checklist in #363

Issue #3391, opened 2016-04-28
compose up --no-build Build Path Does Not Exist


dsanders11:
Currently if you do a `docker-compose up --no-build` for a `docker-compose.yml` file that looks like:

```
version: '2'

services:
  foo:
    build: "foo"
    image: foo:latest
```

And you're in an environment where the build folder doesn't exist, you'll get:

````
ERROR: build path /foo either does not exist, is not accessible, or is not a valid URL.
````

Creating an empty directory with that name allows things to continue as expected.

I'd like to have a single `docker-compose.yml` file that can be used for building on the developer side and deploying on the production side. As such I never want to build in production and only use the prebuilt images, hence the '--no-build' flag. The behavior of reading the build option in this case is unexpected and counter intuitive.

It looks like this is part of the validating the config logic, which is understandable, but it would be nice if no build was plumbed down so that a missing build path was not an error.

**Comment by metal3d** [comment]
Having image and build directive is an error I guess. Isn't it ?

**Comment by metal3d** [comment]
I precise:

This meand "please, build Dockerfile found in "./foo" directory"
```
build: foo
```
And this means "please use "foo" image, pull it if you don't find it":

```
image: foo:latest
```

So docker-compose doesn't know if it should pull "foo" image (or use the local one), or to build Dockerfile found in "./foo" directory. And more: because the "./foo" directory does'nt exists, docker-compose complain to not find the directory that you give to find your Dockerfile.

metal3d:
You received 2 messages from me for some comments: sorry ! I didn't read your message as I should and I didn't understand what you meant. Now I understand :) and you're absolutly right !

dnephin:
This is covered by #1973, closing as duplicate.

Issue #1973, opened 2015-09-03
Allow some commands to run without full config validation


dnephin:
This has come up a couple times in the last week (most recently in #1972).

Commands like `stop`, `kill`, `ps`, `logs`, `port`, `restart`, `rm`, `start` only need to know the list of service names, not any part of the service definition.

Commands like `build`, `pull` only need to care about one or two fields from the service definition (`build` and `dockerfile` for `build`, `image` for pull).

If we could split our config validation into two phases (a top level service name validation, and a full validation) we could support that first group of commands.

The second group would be more work, and require a three phase validation.

ChrisRut:
+1

ebmeierj:
:+1:

peterwhitesell:
+1

broneq:
+1

mattjohnson:
+1

dionmcm:
+1

Any progress on this?

Chexpir:
+1

CafeLungo:
Still need this, I believe. I just ran into `down` trying to validate build paths. We deploy with `up --no-build`, but cannot `down` the same images without the respective docker build paths present.

```
$ docker-compose -f docker-compose.yml down --volumes --rmi local
ERROR: build path /home/user/modules/app either does not exist, is not accessible, or is not a valid URL
```

Resorting to docker down commands as a workaround.

mustard123:
> Still need this, I believe. I just ran into `down` trying to validate build paths. We deploy with `up --no-build`, but cannot `down` the same images without the respective docker build paths present.
> 
> ```
> $ docker-compose -f docker-compose.yml down --volumes --rmi local
> ERROR: build path /home/user/modules/app either does not exist, is not accessible, or is not a valid URL
> ```
> 
> Resorting to docker down commands as a workaround.

yup having the same issue

paslandau:
I believe this issue is still valid and should be re-opened. I just ran into this problem due to the dual nature of `docker compose` to be used for **building** as well as **running** containers. In my case, I defined an `env_file` in a `docker-compose.yml` config file that does not exist during build time and makes the command fail - even though it is not needed for `build` at all. Error is:

```
open /path/to/compose-secrets.env: The system cannot find the file specified.
 ```

**Comment by KostaGorod** [comment]
this issue is still valid,
can't pull docker during deployment with missing **required** env file

-  Fails with an error (missing file):
```
   env_file:
      - path: .myservice.env
        required:  true
```


[...]

PR #5780, opened 2018-03-13
Skip build config section validation when passing no-build option

smasset:
Partial fix for #1973 only handling #3391 part


turbo:
Need this. Because docker compose with multiple docker compose files still fails even when a build is overwritten with an image, when the build path doesn't exist. There is no need to validate build paths when they are overwritten. I guess this PR provides a workaround.

compose/config/config.py
```diff
@@ -718,7 +732,7 @@ def process_service(service_config):
             for path in to_list(service_dict['env_file'])
         ]
 
-    if 'build' in service_dict:
+    if 'build' in service_dict and 'build' not in skipped_sections:
```
shin-: Most of the design for the feature is made in a generic fashion, but we end up just making a couple string matches (here and in `validate_paths`) which seems to defeat the purpose. What's the rationale for not removing the declared `skipped_sections` from the config entirely? Wouldn't it make #1973 simpler to implement down the line?

smasset: Thanks for the review.

Not much of a rationale except I didn't think about it and minimum entropy production principle...

Now that you mention it, I'm mostly fine with it except I'm not sure if we should change original config that downward components might need.

Pretty busy this week. Will see if I can update the PR if I get another sleepless night.

chmorgan:
Any updates on these changes? I'm trying to use a docker-compose.yml to bring services up after loading images and bumping into this issue as the build files are not present.

ijc:
> Pretty busy this week. Will see if I can update the PR if I get another sleepless night.

@smasset are you still planing to update?

glours:

PR #5788, opened 2018-03-15
Allow dash and underscore in project name.

mnottale:


mnottale:
Fixes #4002 .

Issue #5380, opened 2017-11-16
Support docker options cpu-rt-runtime and cpu-rt-period


waldbaer:
Hello docker-compose team,

I would be happy to have the docker run options "[cpu-rt-period](https://docs.docker.com/engine/reference/run/#runtime-constraints-on-resources)" and "[cpu-rt-runtime](https://docs.docker.com/engine/reference/run/#runtime-constraints-on-resources)" also available in docker-compose.
These options allow modification of the scheduling policy inside containers.

Based on my research these options are available in [docker-py](https://github.com/tz70s/docker-py/commit/d74f1bc3805176af229ae3d2e9e5dbe1f0b40a15), but are not released yet.

Best regards
Sebastian

devkid:
This is in docker-py as of [Dec 14th](https://github.com/docker/docker-py/blame/82e57a4c158c4ff13340a29319c6445d81ddc208/docker/types/containers.py#L122).

Would also like to have this feature.

shin-:
Implemented in #5798

rjmccabe3701:
is there a reason these aren't implemented in the 3.x schemas?

PR #5798, opened 2018-03-16
Add support for cpu_rt_period and cpu_rt_runtime.

mnottale:

### DIFF
--- a/compose/config/config.py
+++ b/compose/config/config.py
@@ -69,6 +69,8 @@ DOCKER_CONFIG_KEYS = [
     'cpu_percent',
     'cpu_period',
     'cpu_quota',
+    'cpu_rt_period',
+    'cpu_rt_runtime',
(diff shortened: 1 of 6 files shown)


mnottale:
Fixes #5380 .

compose/config/config_schema_v2.1.json
```diff
@@ -107,6 +107,8 @@
         "cpu_shares": {"type": ["number", "string"]},
         "cpu_quota": {"type": ["number", "string"]},
         "cpu_period": {"type": ["number", "string"]},
+        "cpu_rt_period": {"type": ["number", "string"]},
+        "cpu_rt_runtime": {"type": ["number", "string"]},
```
shin-: I know I probably said something else the other day, but these options aren't compatible with v2.1 - they were added in API 1.25, which is 2.2+

PR #5838, opened 2018-03-29
Bump docker SDK -> 3.2.0

### LINKED ISSUE #4926: Want a way to specify a dockerfile outside the context in the docker-compose dockerfile directive
I've read several discussions about the security reasons why paths outside the context are not permitted.  OK, but it seems to me that that should not apply to the dockerfile directive.  After all, you can specify a dockerfile outside the context when using the docker build command itself, using -f.  I'd like to be able to do the same thing in docker-compose, either relative to the context or relative to the dockerfile-compose.yml file.  My goal is to put all the dockerfile-specific config in a subdirectory next to my context directory.  I can do that with everything except the dockerfile itself.

shin-:
https://github.com/docker/docker-py/releases/tag/3.2.0

Resolves #4926

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,7 +2,7 @@ backports.ssl-match-hostname==3.5.0.1; python_version < '3'
 cached-property==1.3.0
 certifi==2017.4.17
 chardet==3.0.4
-docker==3.1.4
+docker==3.2.0
 docker-pycreds==0.2.1
 dockerpty==0.4.1
 docopt==0.6.2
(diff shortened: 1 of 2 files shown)


KevinSjoberg:
How would I go about using this with Compose now? Basically I want to use a build context in `/code/project-a`, but the `Dockerfile` and `docker-compose.yml` file is in `/code/project-b`. Standing in `/code/project-a` and running `docker-compose -f /code/project-b up` does not seem to work.

PR #5857, opened 2018-04-07
Add --hash opt for config command.

### LINKED ISSUE #2013: Expose config-hash checking
I'm writing a tool that uses docker-compose and it needs to tell the user whether they need to 'up' a service. To do this, I have reproduced the config-hash so I can check if the config has been changed, but the method of doing this changes between docker-compose versions, so it would be handy to be able to ask docker-compose whether it thinks the config has changed directly, without performing any other actions.

wfender:
This feature is needed for many people to anticipate and improve rolling update / restart tasks. Resolves #2013

Signed-off-by: Fender William <fender.william@gmail.com>

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -326,7 +326,7 @@ class TopLevelCommand(object):
                                      anything.
             --services               Print the service names, one per line.
             --volumes                Print the volume names, one per line.
-            --hash=""                Print the service config hash, one per line.
+            --hash="all"             Print the service config hash, one per line.
                                      Set "service1,service2" for a list of specified services.
         """
 
@@ -349,9 +349,9 @@ class TopLevelCommand(object):
             print('\n'.join(volume for volume in compose_config.volumes))
             return
 
-        if options.get('--hash') is not None:
+        if options['--hash'] is not None:
             self.project = project_from_options('.', self.toplevel_options)
-            if options['--hash'] == '':
+            if options['--hash'] == "all":
                 for service in self.project.services:
                     print('{} {}'.format(service.name, service.config_hash))
             else:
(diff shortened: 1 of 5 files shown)


wfender:
Ping.

compose/cli/main.py
```diff
@@ -318,15 +318,15 @@ def config(self, options):
         """
         Validate and view the Compose file.
 
-        Usage: config [options]
+        Usage: config [options] [SERVICE...]
```
shin-: This is going to be really awkward, because `docker-compose config service1` will be a no-op where people could reasonably expect an output containing only the specified service(s) and dependent resources.
I'm not suggesting implementing that (it would be a significant amount of work for a mostly useless option), but I'd like to have a better way to make sure it's obvious the filtering only applies to `--hash`.

1. Make `--hash` a string flag that accepts a comma-separated list of services (leave blank for all services)
2. Remove the option and just let users grep for the services they want to process.

I'm slightly partial to 2 because it's more UNIXy, but 1 may be reasonable too.

compose/cli/main.py
```diff
@@ -348,6 +348,18 @@ def config(self, options):
             print('\n'.join(volume for volume in compose_config.volumes))
             return
 
+        if options['--hash']:
+            self.project = project_from_options('.', self.toplevel_options)
+            with errors.handle_connection_errors(self.project.client):
```
shin-: Is this necessary? I don't think we're making any remote calls here.

compose/service.py
```diff
@@ -365,6 +365,13 @@ def convergence_plan(self, strategy=ConvergenceStrategy.changed):
 
         return ConvergencePlan('noop', containers)
 
+    def get_config_hash(self):
```
shin-: We're not writing Java here - let's just access the property directly :slightly_smiling_face:

wfender:
Here are some adaptation and fix following your comment and i chose the "arg" solution.

wfender:
Here are some adaptation and fix following your comment and i chose the "arg" solution ;)

wfender:
@shin- ?

shin-:
Rebased, updated and merged in #6126 

Thank you again!

PR #5867, opened 2018-04-11
Refactor bash completion for services

albers:
This aligns bash completion for services it to the same principles used in Docker's bash completion.

### current state
We have one function `___docker_compose_all_services_in_compose_file` that outputs a list of service names. Its name is outdated because the names no longer are parsed from compose files.
It is used twice for code that scales services where a suffix of `=` is required.

There are also `__docker_compose_services_{all,from_build,from_image,paused,running,stopped}`, which actually perform completion. 3 of these functions are only used once, one twice.
The really important functions are `__docker_compose_services_all` (used 11 times) and `__docker_compose_services_running` (6 invocations),

### design principles
First of all, the function names don't make clear in which category a function falls

- a function that generates output and has no side effects
- a function that runs in void context and is executed solely for its side effects, i.e. modifyig global environment variables for bash completion.

I prefer to have two functions

- `__xxx_yyy` for generating all values
- `__xxx_complete_yyy` that performs the actual completion in the context of bash's global variables and uses `__xxx_yyy`.

This separation of concerns also eases testing.

The existing functions also have minor inconsistencies and code duplication, These can be eliminated by parameterizing the new functions.

### changes

I introduced two new functions, `__docker_compose_services` and `__docker_compose_complete_services` for the separate concerns mentioned above.

I removed the least-used helper functions in favor of more generic parameterized base functions that can easily be extended in the future (as was done in Docker's completion).

So we now have code like 
```bash
__docker_compose_complete_services --filter source=build
```
instead of 
```bash
__docker_compose_services_from_build
```
_I admit that the old names can be more descriptive. But I'm not a big fan of exploding numbers of helper functions that are just used once and suffer from duplication._

### TLRD

So now the most important use cases are covered by `__docker_compose_complete_services` and `__docker_compose_complete_running_services`, while the one-shot use cases use `__docker_compose_complete_services --filter key=value`.

### DIFF
--- a/contrib/completion/bash/docker-compose
+++ b/contrib/completion/bash/docker-compose
@@ -81,41 +81,24 @@ __docker_compose_nospace() {
 	type compopt &>/dev/null && compopt -o nospace
 }
 
-# Extracts all service names from the compose file.
-___docker_compose_all_services_in_compose_file() {
-	__docker_compose_q config --services
+
+# Outputs a list of all defined services, regardless of their running state.
+# Arguments for `docker-compose ps` may be passed in order to filter the service list,
+# e.g. `status=running`.
+__docker_compose_services() {
+	__docker_compose_q ps --services "$@"
 }
 
-# All services, even those without an existing container
-__docker_compose_services_all() {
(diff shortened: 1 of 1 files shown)


shin-:
LGTM :) Thank you!

PR #5968, opened 2018-05-17
Fix bash completion for running services

albers:
In #5867, I introduced a regression: Bash completion for running services, e.g. in `docker-compose stop <tab>` does no longer complete services.

This PR fixes the bug.


shin-:
Thanks! :)

PR #5992, opened 2018-05-29
Add 3.7 schema and add rollback_config to it

vdemeester:
Mirrors https://github.com/docker/cli/pull/409. As for `update_config`, it will be ignore by `docker-compose` :wink: 

cc @thaJeztah @shin-


compose/const.py
```diff
@@ -67,4 +69,5 @@
     API_VERSIONS[COMPOSEFILE_V3_4]: '17.06.0',
     API_VERSIONS[COMPOSEFILE_V3_5]: '17.06.0',
     API_VERSIONS[COMPOSEFILE_V3_6]: '18.02.0',
+    API_VERSIONS[COMPOSEFILE_V3_7]: '18.02.0',
```
thaJeztah: 18.06.0?

vdemeester: well, I wasn't sure about that. From my understanding, it's the "lowest" API version supported. And so far, there is nothing in v3.7 that wasn't already supported by v3.6 ; this is why I did put the same version.

vdemeester:
Updated :angel:

shin-:
Thank you! :1st_place_medal:

thaJeztah:
Note that ideally, we'd have some other compose-file changes in as well;

- https://github.com/docker/cli/pull/479 - but haven't found the time yet to carry that; I tried, but ran into some issues
- https://github.com/docker/cli/pull/896 - templating driver option

@shin- not sure if you work with release-branches, so let me know if you were planning to do a release (because after it's in a release, we likely can't change the compose-file spec anymore)

@vdemeester for the above; issues; I'll try to find some time, but if you have some and can help 🤗

shin-:
@thaJeztah I don't think we'll do a release before at least mid-June, so things should be alright on that end.

thaJeztah:
Thanks! Feel free to ping me if things change in that respect :+1:

vdemeester:
@thaJeztah can take care of it :wink:

thaJeztah:
@vdemeester https://github.com/docker/cli/pull/479 could be tricky; when I had the CLI working, I think SwarmKit still didn't pick up the configuration (but perhaps I did something wrong); I can push that branch as a quick WIP

PR #6092, opened 2018-07-19
support newer minor version of requests

ofek:
@shin- Upon reviewing https://github.com/docker/docker-py/pull/2089 and https://github.com/requests/requests/blob/master/HISTORY.rst#2190-2018-06-12 I realized we can support `requests` 2.19.x for those of us who pin that without modifying Compose's `urllib3` support at all. :smile: 

2.19.x also includes better compatibility with Python 3.7!!! https://github.com/requests/requests/pull/4501

How does this sound?


shin-:
That seems reasonable - thank you! :+1:

PR #6100, opened 2018-07-20
Add progress messages to parallel pull

### LINKED ISSUE #5960: docker-compose pull does not show progress anymore
## Description of the issue
Recently, `docker-compose pull` has stopped showing the progress. For big images, it just takes forever to download, the progress is really helpful.

I remember that this should only happen when you use `--quiet` flag with the command.

```
```
docker-compose version 1.21.0, build 5920eb0
docker-py version: 3.2.1
CPython version: 3.6.4
```
Output of "docker version"
```
Client:
 Version:      18.05.0-ce-rc1
 API version:  1.37
 Go version:   go1.9.5
 Git commit:   33f00ce
 Built:        Thu Apr 26 00:58:56 2018
 OS/Arch:      darwin/amd64
 Experimental: false
 Orchestrator: swarm

Server:
 Engine:
  Version:      18.05.0-ce-rc1
  API version:  1.37 (minimum version 1.12)
  Go version:   go1.10.1
  Git commit:   33f00ce
  Built:        Thu Apr 26 01:06:49 2018
  OS/Arch:      linux/amd64
  Experimental: true

1. run `docker-compose pull`

### Observed result
screen shot 2018-05-14 at 11 21 12 am

### Expected result
It should show the progress, something like this:
screen shot 2018-05-14 at 11 31 26 am

## Additional information

macOS High Seirra version 10.13.4, docker-compose install method brew

shin-:
Resolves #5960 

Linux binary: https://1243-15045751-gh.circle-artifacts.com/0/docker-compose-Linux-x86_64
OSX binary: https://1242-15045751-gh.circle-artifacts.com/0/docker-compose-Darwin-x86_64
Windows binary: https://ci.appveyor.com/api/buildjobs/7j4fpw9brn1hjowq/artifacts/dist%2Fdocker-compose-Windows-x86_64.exe

### DIFF
--- a/compose/parallel.py
+++ b/compose/parallel.py
@@ -313,6 +313,13 @@ class ParallelStreamWriter(object):
             self._write_ansi(msg, obj_index, color_func(status))
 
 
+def get_stream_writer():
+    instance = ParallelStreamWriter.instance
+    if instance is None:
+        raise RuntimeError('ParallelStreamWriter has not yet been instantiated')
+    return instance
+
+
 def parallel_operation(containers, operation, options, message):
     parallel_execute(
         containers,
--- a/compose/progress_stream.py
+++ b/compose/progress_stream.py
@@ -19,12 +19,11 @@ def write_to_stream(s, stream):
 def stream_output(output, stream):
     is_terminal = hasattr(stream, 'isatty') and stream.isatty()
     stream = utils.get_output_stream(stream)
-    all_events = []
     lines = {}
     diff = 0
 
     for event in utils.json_stream(output):
-        all_events.append(event)
+        yield event
         is_progress_event = 'progress' in event or 'progressDetail' in event
 
         if not is_progress_event:
@@ -56,8 +55,6 @@ def stream_output(output, stream):
             write_to_stream("%c[%dB" % (27, diff), stream)
 
         stream.flush()
-
-    return all_events
 
 
 def print_output_event(event, stream, is_terminal):
--- a/compose/project.py
+++ b/compose/project.py
@@ -548,16 +548,37 @@ class Project(object):
     def pull(self, service_names=None, ignore_pull_failures=False, parallel_pull=False, silent=False,
              include_deps=False):
         services = self.get_services(service_names, include_deps)
+        msg = not silent and 'Pulling' or None
 
         if parallel_pull:
             def pull_service(service):
-                service.pull(ignore_pull_failures, True)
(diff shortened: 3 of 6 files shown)


metabrain:
Works great for me in Windows 10, thanks for the patch 👍

bkraul:
Any chance this could be merged?

leandro2r:
Waiting it. This change will be useful

m0nhawk:
+1 for the progress for parallel download!

bkraul:
@mnottale Niiiice! It's coming, then??

shin-:
Thanks for the feedback folks, and thank you @mnottale for reviewing!

benyanke:
I just got `ERROR: for sqlcmd  'NoneType' object is not iterable` when using the linxu binary mentioned above.

Is that something  that should be discussed here, or in a new ticket?

shin-:
Probably unrelated, please make a new issue

HaraldNordgren:
@shin- This is nice, thanks for the fix! Even better would be if it showed total progress instead of just per-layer, but I understand if that may be difficult/impossible to estimate without pre-fetching data.

elashpear:
It should at least be possible to show # of layers left to pull, ex:
```
Pulling db:  layer 5/10: downloading (50%)
Pulling api: layer 3/6: extracting (10%) 
```
Without this info, what we have right now is basically just an activity indicator rather than a progress indicator, which is still better than what we had before (nothing). But it's impossible to tell how far along anything is since with lots of layers it just cycles between downloading and extracting an unknown number of times.

gerroon:
Is this merged? I do not get any progress on Debian Buster

benyanke:
What version of docker-compose are you using? It's likely debian doesn't have the absolute latest version, due to how non-rolling distros work.

bkraul:
I would suggest using the pip (python3-pip) docker-compose deployment instead. It always gives you the latest.

benyanke:
Alternatively, running it within docker itself can work quite well. It's not 100% foolproof, but for most uses, this has been great for me:

```bash
# add to .bashrc
function docker-compose() {
  docker run \
    --interactive \
    --tty \
    --workdir "$PWD" \
    --rm \
    --volume /var/run/docker.sock:/var/run/docker.sock \
    --volume "$PWD":"$PWD" \
    docker/compose:latest \
    docker-compose $@
}


```

steakhutzeee:
Hi! 

Actually i'm on OMV5 over Armbian on and Odroid HC2.

Issue #4460, opened 2017-02-09
-f - does not work for in container script


cmc333333:
If I install using a container, e.g. 
```sh
curl -L https://github.com/docker/compose/releases/download/1.10.0/run.sh > ~/bin/docker-compose
```

I can't pipe in compose configurations:
```sh
~/bin/docker-compose -f - up < my-config.yml
```

If I install via pip, instead, I can.

PR #6104, opened 2018-07-24
Fixes pipe handling in container mode.

### LINKED ISSUE #4599: Docker-compose via container doesn't connect pipes correctly
This can be tested with something like the following

`cat db.sql | docker-compose run --rm myimage mysql --host=mysql --user=root --password=foobar my_database`

If docker-compose is installed using the instructions on githubs release page

https://github.com/docker/compose/release

then the above will work

If docker-compose is installed as a container, the above line will just give a mysql prompt - implying pipes aren't getting hooked up correctly?

glorpen:
Closes #4599, #4460

- adds a way to provide options from env in both cases (tty & non tty)
- allocates TTY only if both stdin & stdout are TTYs
- enables interactive mode if stdin is not TTY

Signed-off-by: Arkadiusz Dzięgiel <arkadiusz.dziegiel@glorpen.pl>

### DIFF
--- a/script/run/run.sh
+++ b/script/run/run.sh
@@ -47,10 +47,11 @@ if [ -n "$HOME" ]; then
 fi
 
 # Only allocate tty if we detect one
-if [ -t 1 ]; then
-    DOCKER_RUN_OPTIONS="-t"
-fi
 if [ -t 0 ]; then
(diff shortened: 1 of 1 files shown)


shin-:
LGTM, thanks!

PR #6140, opened 2018-08-15
Add randomly generated slug to container names to prevent collisions

### LINKED ISSUE #4688: service_run_+1
I have a two cronjob for `run` `php` service in same time.

Only one `run` command execute correctly
```
Cannot create container for service php: Conflict. The container name "/folder_php_run_1" is already in use by container {hashhere}. You have to remove (or rename) that container to be able to reuse that name.
Command exited with non-zero status 1
```

shin-:
Fixes #4688

### DIFF
--- a/compose/service.py
+++ b/compose/service.py
@@ -1,7 +1,6 @@
 from __future__ import absolute_import
 from __future__ import unicode_literals
 
-import itertools
 import logging
 import os
 import re
@@ -49,9 +48,11 @@ from .errors import OperationFailedError
 from .parallel import parallel_execute
 from .progress_stream import stream_output
 from .progress_stream import StreamOutputError
+from .utils import generate_random_id
 from .utils import json_hash
 from .utils import parse_bytes
 from .utils import parse_seconds_float
+from .utils import truncate_id
 
 
 log = logging.getLogger(__name__)
@@ -215,13 +216,17 @@ class Service(object):
             )
         )
 
-    def get_container(self, number=1):
+    def get_container(self, number=None):
         """Return a :class:`compose.container.Container` for this service. The
         container must be active, and match `number`.
         """
-
-        for container in self.containers(labels=['{0}={1}'.format(LABEL_CONTAINER_NUMBER, number)]):
-            return container
+        if number is not None and len(number) == 64:
+            for container in self.containers(labels=['{0}={1}'.format(LABEL_CONTAINER_NUMBER, number)]):
+                return container
+        else:
+            for container in self.containers():
+                if number is None or container.number.startswith(number):
+                    return container
 
         raise ValueError("No container found for %s_%s" % (self.name, number))
 
@@ -426,7 +431,6 @@ class Service(object):
         return has_diverged
 
     def _execute_convergence_create(self, scale, detached, start, project_services=None):
-            i = self._next_container_number()
 
             def create_and_start(service, n):
                 container = service.create_container(number=n, quiet=True)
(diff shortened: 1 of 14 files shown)


shin-:
Relating to #1516 , this fixes concerns 2 and 3. The costly queries are unfortunately still there, but this PR is a compromise intended to cause minimal disruption to existing workflows. 

While this can be seen as a breaking change in the sense that container names can no longer be inferred the way they could before, it should be noted that
1. Compose never made any guarantees about the naming pattern of the containers it creates
2. Relying on inferred patterns can result in unexpected errors when the sequentiality breaks, making it something to avoid wherever possible
3. In the rare cases where it is necessary, scripts will still be able to fetch individual containers based on the predictable prefix (the part before the slug), with the caveats mentioned above. The prefix is also backward-compatible.

compose/service.py
```diff
@@ -425,27 +427,32 @@ def _containers_have_diverged(self, containers):
...
+            container = service.create_container(number=n, quiet=True)
+            if not detached:
+                container.attach_log_stream()
+            if start:
+                self.start_container(container)
+            return container
 
-            return containers
+        containers, errors = parallel_execute(
+            [ServiceName(self.project, self.name, number, generate_random_id()) for number in [
```
Dimrok: Maybe I'm missing something but why use `number` here?
```
- [ServiceName(self.project, self.name, number, generate_random_id()) for number in [index for index in range(i, i + scale)]]
+ [ServiceName(self.project, self.name, index, generate_random_id()) for index in range(i, i + scale)]
```

compose/utils.py
```diff
@@ -151,3 +152,21 @@ def unquote_path(s):
     if s[0] == '"' and s[-1] == '"':
         return s[1:-1]
     return s
+
+
+def generate_random_id():
```
Dimrok: Why do we use such a complicated algorithm to generate a random identifier? And is there any specific reason we don't want a fully numerical ID?

Otherwise `uuid.uuid4().hex` sounds good to generate a hexadecimal identifier, no?

tests/acceptance/cli_test.py
```diff
@@ -956,13 +963,13 @@ def test_down(self):
...
-        assert 'Removing v2-full_other_1' in result.stderr
-        assert 'Removing v2-full_web_run_1' in result.stderr
-        assert 'Removing v2-full_web_run_2' in result.stderr
+        assert 'Stopping v2-full_web_' in result.stderr
+        assert 'Stopping v2-full_other_' in result.stderr
+        assert 'Stopping v2-full_web_run_' in result.stderr
+        assert 'Removing v2-full_web_' in result.stderr
+        assert 'Removing v2-full_other_' in result.stderr
+        assert 'Removing v2-full_web_run_' in result.stderr
+        assert 'Removing v2-full_web_run_' in result.stderr
```
Dimrok: Duplicated.

tests/acceptance/cli_test.py
```diff
@@ -2346,10 +2368,10 @@ def test_logs_tail(self):
         self.dispatch(['up'])
 
         result = self.dispatch(['logs', '--tail', '2'])
-        assert 'c\n' in result.stdout
-        assert 'd\n' in result.stdout
-        assert 'a\n' not in result.stdout
-        assert 'b\n' not in result.stdout
+        assert 'y\n' in result.stdout
```
Dimrok: Don't you like `a b c d` anymore?

compose/service.py
shin-: You're not missing anything, I screwed up. Thanks!

compose/utils.py
shin-: The code is a transposition of https://github.com/moby/moby/blob/master/pkg/stringid/stringid.go ; we could go with a UUID instead.

tests/acceptance/cli_test.py
shin-: Now that we have hex strings in the output, `a b c d` are not really safe characters anymore :sweat_smile:

Dimrok: Well, that's a good point.

PR #6141, opened 2018-08-15
Add MUSL libc version of docker-compose

### LINKED ISSUE #3465: running inside alpine linux
Hi,

I'm about to run docker-compose inside a container built with Dockerfile:

```
FROM alpine

RUN apk update && apk add curl
RUN curl -o /usr/local/bin/docker-compose -L https://github.com/docker/compose/releases/download/1.6.2/docker-compose-`uname -s`-`uname -m` && chmod +x /usr/local/bin/docker-compose
ENTRYPOINT ["/usr/local/bin/docker-compose"]
```

but then, running:
`docker run -v /var/run/docker.sock:/var/run/docker.sock compose --help`
leads to a `no such file or directory`error.
I read that this is due to docker-compose being binary compiled from python sources, using the glibc while alpine runs musl.
Would it be possible to get a docker-compose binary running on alpine?

andyneff:
Resolves #3465

rocker-compose is discontinued, and this should bring musl support into the mainstream docker-compose build path

Builds executable using alpine 3.4 so that python 3.5 is used to run `pyinstaller`. Resulting executable should work as far back as alpine 3.2, which should be a decent amount of backwards compatibility. When using alpine 3.1 or older on prior projects, I never got `pyinstaller` 3.3.1 working, so hopefully this isn't an issue.

The docker image essentially `pip download`s the `requirements-build.txt` file so that the correct version of `pyinstaller` is used. Then the `pyinstaller` is bootstrapped for musl, and turned into a wheel. After the tox environment is created, the wheel is installed. By the time the `requirements-build.txt` is used in the normal build path, `pyinstaller` is already installed an happy.


shin-:
Thanks for the contribution @andyneff !

I'm not sure I quite understand what spurred this PR - #3465 was resolved a while ago. Is anything wrong with the existing [Dockerfile.run](https://github.com/docker/compose/blob/master/Dockerfile.run)? 

The way I see it, if we merge this, we'll be committing to supporting an additional release binary, which can be a serious long-term engagement. And while that can make sense if a given platform or community is not being appropriately served by the existing solution, this doesn't seem to be the case here, but please let me know if I'm misunderstanding!

cc @garethr @thaJeztah for their thoughts as well

thaJeztah:
could use input from @justincormack @kolyshkin and @ncopa

andyneff:
@shin- There were some internal reasons that spun up the desire to have docker-compose on musl (alpine) without using glibc, and we just wanted the same convenience that the executable offered glibc. Also, I had  to solve this on another project of mine, so this was mostly copy and paste and paste for me.

As I saw it, 3465 said you can either:

1. Use it in the pure python form
2. Add glibc and use glibc
3. Bootstrap pyinstaller for musl

We decided the most appropriate thing was option 3.

It's was my hope you decide this is low impact enough and worth adding to the binary pool, especially with the hard part completed.

ncopa:
It may also be of interested that I [recently](https://git.alpinelinux.org/cgit/aports/commit/?id=e7a2d585a0faeedf75c763676bdd77768cfeb5f8) added docker-compose as an alpine package so you can `apk add docker-compose`

ulyssessouza:
Thank you very much @andyneff , but I have to drop this PR in favor of https://github.com/docker/compose/pull/6641

PR #6162, opened 2018-09-04
Use a specific tag of compose in powershell run script

mausch:
since there is no `docker/compose:latest`

Resolves #


ulyssessouza:
Hello @mausch ! Thank you for the PR, but this is fixed by https://github.com/docker/compose/pull/6641
which will introduce the `latest` tag in the next release

PR #6168, opened 2018-09-05
[armhf] Make Dockerfile.armhf compatible with main

### LINKED ISSUE #6167: Dockerfile.armhf should be updated
## Description of the issue

Dockerfile is now using python:3.6 as base image and Dockerfile.armhf still uses Debian as base image. Also, armhf version tries to build python what is not necessary since python official image has official support for ARM architecture.

Not applicable

1. Just check Dockerfile and Dockerfile.armhf
2. Check python:3.6 official image for ARM support

### Observed result

Outdated Dockerfile.armhf that may lead to different results from regular Dockerfile.

### Expected result

ARM and X86 (and hopefully other archs) shall produce similar (if not identical) results.

not applicable

## Additional information

Not applicable.

jrbenito:
Dockerfile now uses python:3.6 image while Dockerfile.armhf uses
debian. Python image is officially supported in ARM archtecture hence,
the now both dockerfiles differs only on dockerbins.tgz file version.

This was tested on a RaspberryPi3.

May we use environmental variables to select dockerbins.tgz?

Signed-off-by: Josenivaldo Benito Jr <jrbenito@benito.qsl.br>

Resolves #6167

### DIFF
--- a/Dockerfile.armhf
+++ b/Dockerfile.armhf
@@ -1,54 +1,20 @@
-FROM armhf/debian:wheezy
+FROM python:3.6
 
 RUN set -ex; \
     apt-get update -qq; \
     apt-get install -y \
         locales \
-        gcc \
-        make \
-        zlib1g \
-        zlib1g-dev \
-        libssl-dev \
-        git \
-        ca-certificates \
         curl \
-        libsqlite3-dev \
-        libbz2-dev \
-    ; \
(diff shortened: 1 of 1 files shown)


jrbenito:
This should be upmerged on top of PR#6061 since it presents a better overall solution.

shin-:
Thanks @jrbenito - changes LGTM!

PR #6176, opened 2018-09-10
optimizes logo - every bite counts!

mathieu-aubin:
Resolves - Not applicable

Optimizes/minifies logo, losslessly, too.. 
Why? because we can and every byte counts in my opinion

Thanks


shin-:
Thank you for your contribution @mathieu-aubin !

The way I look at it, this would just add another image to our `git` repo, effectively making it grow in size. If "every byte counts", isn't this deleterious to your cause?

shin-:
Thank you for your candor! As our GitHub repo only receives a miserly 20k views a day, I'll go ahead and say the worldwide bandwidth can take the hit if it makes our `git clone`s faster.

mathieu-aubin:
Just as expected, typical answer. This is the attitude that rules the world, i sincerely hope - for both my kids and yours - that a change happens. And no, im not making it personnal and yes, its just a stupid picture.

PR #6190, opened 2018-09-13
Allow dependencies to be brought down after 'run'

imrook:
Use case: Using `docker-compose run --rm -p $BUILD_NUM` in a CI environment to bring up all dependencies and run a set of tests and return an exit code reflecting the result of the test process.

Problem: Dependencies are left running after the container running the tests has terminated. Something similar to the following block must be used to ensure the desired behavior of stopping and removing all containers and returning the exit code from the test container.
~~~
docker-compose run --rm -p $BUILD_NUM app scripts/run_tests.sh
exit_code = $?
docker-compose down
exit $exit_code
~~~

This PR allows the above to be done with the following line:
~~~
docker-compose run --rm --rm-deps -p $BUILD_NUM app scripts/run_tests.sh
~~~

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -813,6 +813,7 @@ class TopLevelCommand(object):
             -u, --user=""         Run as specified username or uid
             --no-deps             Don't start linked services.
             --rm                  Remove container after run. Ignored in detached mode.
+            --rm-deps             Down dependencies after run. (Requires --rm)
             -p, --publish=[]      Publish a container's port(s) to the host
             --service-ports       Run command with the service's ports enabled and mapped
                                   to the host.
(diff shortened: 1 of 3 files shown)


shin-:
Hi @imrook - Thank you for your contribution!

In the future please create an issue first to lay out the problem and how you're looking to solve it so we can validate the design. 👼 

My primary concern here is that the exit code for `docker-compose run --rm-deps app scripts/run_tests.sh` will only be 0 if both the `run` operation and the `down` operation are successful, so you're losing a ton of information you would have otherwise had. Your first snippet, while longer, is objectively better code if you care about the container's exit code. 
I think in line with the UNIX philosophy principles, each Compose subcommand should [do one thing and do it well](https://en.wikipedia.org/wiki/Unix_philosophy#Do_One_Thing_and_Do_It_Well), and the `--rm-deps` flag most certainly runs contrary to this principle.

ijc:
Since there has been no response to @shin-'s comments I'm going to close this for now. If you are still interested in discussion around this usecase please do continue (or perhaps open a design issue first as suggested).

PR #6194, opened 2018-09-14
[#6026] Added -q, --quiet option to 'up' and 'down' commands to suppress parallel stream messages.

### LINKED ISSUE #6026: Feature Request: Provide --quiet option to 'up' command
## Description of the issue

There is currently no way to prevent the `up` command from producing extraneous (i.e. 'meta' messages, not produced by the service being run) output.
Supposedly `--log-level` can be used to suppress unwanted output but unfortunately, this has no effect on `up` (it looks like there are quite a few uses of `print()` in the cli layer which of course isn't affected by log levels). 

Example:

```
~$ docker-compose --log-level CRITICAL up service
Starting someproject_service_1 ... done
Attaching to someproject_service_1
someproject_service_1 exited with code 0
```

### Observed result

`up` always produces 'meta'-output.

### Expected result

To have an option of preventing `up` from generating meta output.

## Additional information

```
docker-compose version 1.21.2, build a133471
docker-py version: 3.3.0
CPython version: 3.6.5
```

BEllis:
Added -q, --quiet option to 'up' and 'down' commands to suppress parallel stream messages.

Resolves #6026

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -42,6 +42,7 @@ from ..service import ConvergenceStrategy
 from ..service import ImageType
 from ..service import NeedsBuildError
 from ..service import OperationFailedError
+from ..utils import set_quiet
 from .command import get_config_from_options
 from .command import project_from_options
 from .docopt_command import DocoptDispatcher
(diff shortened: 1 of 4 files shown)


shin-:
Thanks for the contribution @BEllis !
However, I would rather we fix `--log-level` to apply to everything Compose produces, rather than add more flags with overlapping functionality. I don't think this PR approaches the problem in the right way.

BEllis:
I did actually do that first, but then I thought it would cause side-effects for backwards compatibility.

I'm happy to change it back to take effect when logging-level < WARNING

BEllis:
I'll raise a separate PR with that approach.

BEllis:
Closing as superseded by #6217

PR #6204, opened 2018-09-19
Don't convert slashes for UNIX paths on Windows hosts

### LINKED ISSUE #5716: Allow linux like paths in volume source on Windows
I would expect the following two docker-compose files to work exactly the same but they do not.  The first one works with **C:/** but the second one with just **/** does not.  In most other places in docker like Dockerfiles, Linux like paths can be used with no issues.

# Works:
```
version: '3.3'
services:
  arch.samples.containers:
    volumes:
    - type: bind
      source: ${ArchSamplesContainersConfigPath-C:/etc/Arch.Samples.Containers}
      target: /etc/Arch.Samples.Containers
      read_only: true
      volume:
        nocopy: true
```

# Does not work:
```
version: '3.3'
services:
  arch.samples.containers:
    volumes:
    - type: bind
      source: ${ArchSamplesContainersConfigPath-/etc/Arch.Samples.Containers}
      target: /etc/Arch.Samples.Containers
      read_only: true
      volume:
        nocopy: true
```

Fails with "Cannot create container for service arch.samples.containers: invalid mount config for type "bind": bind source path does not exist"

Docker Version:
Client:
 Version:       18.02.0-ce
 API version:   1.36
 Go version:    go1.9.3
 Git commit:    fc4de44
 Built: Wed Feb  7 21:12:53 2018
 OS/Arch:       windows/amd64
 Experimental:  true
 Orchestrator:  kubernetes

Server:
 Engine:
  Version:      18.02.0-ce
  API version:  1.36 (minimum version 1.12)
  Go version:   go1.9.3
  Git commit:   fc4de44
  Built:        Wed Feb  7 21:20:15 2018
  OS/Arch:      linux/amd64
  Experimental: true

docker-compose version 1.19.0, build 9e633ef3
docker-py version: 2.7.0
CPython version: 2.7.14
OpenSSL version: OpenSSL 1.0.2k  26 Jan 2017

Windows Version:
Windows 10 Build 17107.rs4_release.180220-1350

shin-:
Fixes #5716 
Fixes docker/for-win#1829

### DIFF
--- a/compose/config/types.py
+++ b/compose/config/types.py
@@ -136,6 +136,20 @@ def normalize_path_for_engine(path):
     return path.replace('\\', '/')
 
 
+def normpath(path, win_host=False):
+    """ Custom path normalizer that handles Compose-specific edge cases like
+        UNIX paths on Windows hosts and vice-versa. """
+
+    sysnorm = ntpath.normpath if win_host else os.path.normpath
+    # If a path looks like a UNIX absolute path on Windows, it probably is;
+    # we'll need to revert the backslashes to forward slashes after normalization
+    flip_slashes = path.startswith('/') and IS_WINDOWS_PLATFORM
+    path = sysnorm(path)
+    if flip_slashes:
+        path = path.replace('\\', '/')
+    return path
+
+
 class MountSpec(object):
     options_map = {
         'volume': {
@@ -152,12 +166,11 @@ class MountSpec(object):
 
     @classmethod
     def parse(cls, mount_dict, normalize=False, win_host=False):
-        normpath = ntpath.normpath if win_host else os.path.normpath
         if mount_dict.get('source'):
             if mount_dict['type'] == 'tmpfs':
                 raise ConfigurationError('tmpfs mounts can not specify a source')
 
-            mount_dict['source'] = normpath(mount_dict['source'])
+            mount_dict['source'] = normpath(mount_dict['source'], win_host)
             if normalize:
                 mount_dict['source'] = normalize_path_for_engine(mount_dict['source'])
 
@@ -247,7 +260,7 @@ class VolumeSpec(namedtuple('_VolumeSpec', 'external internal mode')):
         else:
             external = parts[0]
             parts = separate_next_section(parts[1])
-            external = ntpath.normpath(external)
+            external = normpath(external, True)
             internal = parts[0]
             if len(parts) > 1:
                 if ':' in parts[1]:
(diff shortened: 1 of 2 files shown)


compose/config/types.py
```diff
@@ -136,6 +136,20 @@ def normalize_path_for_engine(path):
     return path.replace('\\', '/')
 
 
+def normpath(path, win_host=False):
+    """ Custom path normalizer that handles Compose-specific edge cases like
+        UNIX paths on Windows hosts and vice-versa. """
+
+    sysnorm = ntpath.normpath if win_host else os.path.normpath
```
Dimrok: Shouldn't this be:
```
-sysnorm = ntpath.normpath if win_host else os.path.normpath
+sysnorm = ntpath.normpath if win_host else posix.normpath
```

Dimrok:
Just to note everything related to path here:

- `docker-compose` runs on both Windows- and Unix-based distributions, so `os.path.normpath` can be either `posixpath` or `ntpath`.
- the host can be Linux running Linux containers, Windows running Windows containers and Windows running Linux containers.
- The volume definition can be `<linux>:<linux>`, `<windows>:<linux>` and `<windows>:<windows>`. 
- Docker for Widows running a Linux container has a proxy that does some path manipulation and validation:
  - if you write `C:/foo/bar:/bar`, the proxy will ensure the volume is mounted somewhere and translate the given path to `/mnt/c/foo/bar` (where the volume is mounted).
  - if you write `/c/foo/bar:/bar`, no validation is done but the path is still translated to `/mnt/c/foo/bar` (we don't want that).
  - You can have both Posix and NT path (e.g. `/var/run/docker.sock:/var/run/docker.sock` and `C:\\Users\\toto:/toto`) for the same container. Docker for Windows understands absolute linux path as being part of the host (in this case, bind-mounting the socket)
  - An _ill-formed_ path such as `C:/foo/bar` is translated to `/mnt/c/foo/bar`.
- Docker for Windows running Windows containers doesn't manipulate paths.
- Volume configuration like `C:\\Users\\foo:C:\\foo` works natively on Docker for Windows using Windows containers
- You can have multi-arch images so you have to be aware of the host 

The point is: We shouldn't normalize the paths at all. The engine is clever enough to understand what you meant. Or if we really want to (e.g. to change `/foo/bar/../shiny` to `/foo/shiny` if we want to be fancy), we shouldn't change the _type_ of paths the user provided.

compose/config/types.py
shin-: `win_host` will be true only if `COMPOSE_FORCE_WINDOWS_HOST` is set. In this case we want to produce NT style paths regardless of the client platform. In all other cases, we want to use the default normalization for the client system, hence the use of `os.path.normpath`

shin-:
I think you make a good point about not having to normalize paths ; that said, I'd want to do a thorough check first to make sure we're not breaking anything by doing so, and have a follow-up PR for that specifically. For the time being, if this fixes docker/for-win#1829, let's roll with it.

PR #6212, opened 2018-09-21
Evaluate the value of the env file vars.

### LINKED ISSUE #3527: support newlines in env files
Environment variable values can have new lines in them and docker-compose.yml already supports that because yml does:

```
test:
    image: alpine
    environment:
        a: "b\nc"
    command: sh -c "echo \"$$a\""
```

However, the env file syntax doesn't support this. It looks like it doesn't do anything more than split on newlines and the first `=` sign. https://github.com/docker/compose/blob/86b5ed1a84b5046089abcfded8922d6dc7e318a2/compose/config/environment.py#L16

One way would be to allow quoted strings in the env file. This is not backwards compatible because anyone who already has a quoted string in their env file would see unexpected behaviour. Hopefully someone else has a better idea?

wfender:
I propose this fix. It does not address the multiline problem cause to me it should be a new feature but at least it will allow to use new line and other stuff like this in env file.

Signed-off-by: Fender William <fender.william@gmail.com>

Resolves #3527

### DIFF
--- a/compose/config/environment.py
+++ b/compose/config/environment.py
@@ -37,6 +37,11 @@ def env_vars_from_file(filename):
             line = line.strip()
             if line and not line.startswith('#'):
                 k, v = split_env(line)
+                try:
+                    v = v.encode('utf-8').decode('unicode_escape')
+                except Exception:
+                    log.warning('Env file unicode escape failed for value: "{}".'.format(v))
+                    pass
                 env[k] = v
     return env
 
--- a/compose/config/environment.py
+++ b/compose/config/environment.py
@@ -5,6 +5,7 @@ import codecs
 import contextlib
 import logging
 import os
+import re
 
 import six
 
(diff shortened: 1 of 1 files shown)


shin-:
Thanks @wfender !

Looks like this broke unicode support (see test failures). As far as I can tell, it also misses the point [Aanand made here](https://github.com/docker/compose/issues/3527#issuecomment-256810187) which is that changes to `env_file` handling should happen upstream in `moby/moby` first, otherwise we're introducing a confusing discrepancy between Docker and Compose.

wfender:
Hi @shin- 
It wasn't the unicode support but just a warning that I've add ;) and i reduce the scope to evaluate only a double quoted value.

Like many others I'm disappointed that such basic problem could not be solved for so long time.
For moby, the issue is closed and will not be treated despite the many requests.
Unfortunately on the side of moby the discussions about an .env2 file doesn't seem conclusive and i think it's not the place to develop an .env2 file.

So like the 50+ comments I think we could at least evaluate the double quoted values and that will probably help the many people who do not benefit from swarm mode. (docker secret ...) for certificates managment and other kind of variables.
Otherwise if we want the problem to be solved on the docker engine side (and I'll understand that we don't want to create confusion with the engine support ) then why not close this discussion and send back to moby/moby. 
Thanks :)

ijc:
I came across this while doing some PR grooming.

> why not close this discussion and send back to moby/moby.

Given what I read above from shin and the linked comment I think this is indeed the right thing to do (although since https://github.com/docker/compose/issues/3527#issuecomment-256810187 was written I think this is now https://github.com/docker/cli rather than moby/moby since IIRC such things are handled client side).

Issue #3729, opened 2016-07-13
allow removing something in docker-compose.override.yml


stefanfoulis:
My usecase is that the base `docker-compose.yml` maps a certain port in the `ports` section. In my `docker-compose.override.yml` I'd like to change that port since I already have a different service running there on my host.
As far as I understand the current implementation it is only possible to add stuff, override extends the other file.
As far as I can tell there is no way to remove an entry. And I can't think of an obvious syntax for it.

aanand:
Correct. Best practice is to not include anything in `docker-compose.yml` that you might not want to have in another environment.

nkovacs:
My problem with your suggestion is that docker-compose.override.yml might not exist (it's an optional file the developer can create locally). I only want to use docker-compose.override.yml to locally customize ports, like this:

docker-compose.yml:

```
version: '2'
services:
    web:
        ports:
            - 80
            - 443
        ...
```

docker-compose.override.yml:

```
version: '2'
services:
    web:
        ports:
            - 32080:80
            - 32443:443
```

This isn't a big deal because this will only result in 80 being exposed twice, once as a random port and once as 32080. I think it would make sense in this case to only expose it as 32080 automatically (without having to specify anything in the override to remove the original config value)

tomekit:
Facing the same problem. I've got one specific host which shouldn't start the one of the docker containers for given service. Yeah, maybe my `docker-composer.yml` approach is not the best, but I can see the need of removing something.

paunin:
Does anyone got script to remove ports from docker-compose file ? :)

paunin:
Alright, there is no solution so far, so I got a simple port-remover 
https://github.com/paunin/docker-compose-v2-ports-remover

In general I dont think docker-compose should allow to do it, as then logic of merging many files is gonna be complex...

bluemoehre:
This was already solved here https://github.com/docker/compose/pull/3939

jedie:
It's also not possible to removed a service, isn't it?

OnkelTem:
Having an "override" thing that cannot do half of the job... well, it's something.

mqu:
I'd like to be able to override some services and disable some others.

sauron918:
Same issue, I'd like to have possibility to **remove** port, dns, etc in `docker-compose.override.yml` but not override or extend with another one..

keithy:
Anyone notice that the only thing that you can't do with docker-compose.overide.yml, is to actually override anything!

[...]

PR #5355, opened 2017-11-08
Implement 'overwrite_ports' config option

shcallaway:
An alternative to the `overwrite_multivals` config option proposed in #5354 .

In your `docker-compose.override.yml`, the `overwrite_ports` config option allows you to toggle whether ports should be _concatenated with the base_ (false) or _overwritten_ (true).

_This PR is a work in progress. Todo:_

* compose/config/validation.py
* Unit tests
* Documentation

PR #6213, opened 2018-09-21
Add override option to service configuration for ports

### LINKED ISSUE #2260: Overriding the set of values when extending the "ports" option.
The default behavior when extending a service or overriding a whole docker-compose.yml file is to concatenate the sets of values on the multi-value options: ports, expose, external_links, dns and dns_search.

The problem appears when for example you have two production hosts both serving on the port 80 and a single machine for development, there is no way of extending or overriding the services for a development environment that avoids the port collision on the host.

An example: lets assume there are two servers one is a REST API and the other is a webpage that uses the API, both listen on port 80 on the production host

* compose.yml on host A:
```
restserver:
  image: node
  ports:
    - "80:80"
```
* compose.yml on host B:
```
webserver:
  image: django
  ports:
    - "80:80"
```
On a development environment where you want to test this setup you would define:

* compose.override.yml on dev for A:
```
restserver:
  ports:
    - "127.0.0.1:3030:80"
```
* compose.override.yml on dev for B:
```
webserver:
  external_links:
    - "service_restserver_1:api.restserver.domain.com"
  ports:
    - "127.0.0.1:8080:80"
```
Of course the "ports" option is concatenated with the production and there is no way of running both containers on the same host because both try to bind to port 80.

Is there any workaround or yml specific syntax for this use case?

jovobe:
This adds the ability to explicitly override the ports configuration with a given docker-compose.override.yml. This is manly an updated version of this PR: #3939

# Problem
We have a `docker-compose.yml` which we want to modify by using a `docker-compose.override.yml`. According to the docs [[1]](https://docs.docker.com/compose/extends/#adding-and-overriding-configuration) the `ports` section is overridden as follows:

> For the multi-value options ports, expose, external_links, dns, dns_search, and tmpfs, Compose concatenates both sets of values:

This is bad in some cases: What if the base file contains a port binding which we want to change? This is currently not possible.

# Solution
Inspired by @CarstenHoyer, I used his idea to add a new field to the service configuration called `override`. In this field all override-fields can be set (currently only `ports` is supported). This causes `docker-compose` to ignore the port configuration from the base and only uses the override port configuration.

## Example
docker-compose.yml
```
version: '2.1'
services:
  web:
    image: example/web
    ports:
      - "80:80"
```

docker-compose.override.yml
```
version: '2.1'
services:
  web:
    ports:
      - "8080:80"
    override:
      - ports
```

Before this PR the result would be:
```
version: '2.1'
services:
  web:
    image: example/web
    ports:
      - "80:80"
      - "8080:80"
```

with this PR the result will be:
```
version: '2.1'
services:
  web:
    image: example/web
    ports:
      - "8080:80"
```

---
Signed-off-by: Johan M. von Behren <johan@vonbehren.eu>

Resolves #2260 and #3729

---
\[1\]: https://docs.docker.com/compose/extends/#adding-and-overriding-configuration

### DIFF
--- a/compose/config/config.py
+++ b/compose/config/config.py
@@ -1061,7 +1061,12 @@ def merge_service_dicts(base, override, version):
         md.merge_field(field, merge_list_or_string)
 
     md.merge_field('logging', merge_logging, default={})
-    merge_ports(md, base, override)
+
+    if 'ports' in md.override.get('override', []):
+        md.merge_field('ports', merge_override_items_lists, default=[])
+    else:
+        merge_ports(md, base, override)
+
     md.merge_field('blkio_config', merge_blkio_config, default={})
     md.merge_field('healthcheck', merge_healthchecks, default={})
     md.merge_field('deploy', merge_deploy, default={})
@@ -1218,6 +1223,10 @@ def merge_labels(base, override):
     labels = parse_labels(base)
     labels.update(parse_labels(override))
     return labels
+
+
+def merge_override_items_lists(base, override):
+    return sorted(set(override))
 
 
 def split_kv(kvpair):
--- a/compose/config/config_schema_v2.0.json
+++ b/compose/config/config_schema_v2.0.json
@@ -256,7 +256,8 @@
         "volumes": {"type": "array", "items": {"type": "string"}, "uniqueItems": true},
         "volume_driver": {"type": "string"},
         "volumes_from": {"type": "array", "items": {"type": "string"}, "uniqueItems": true},
-        "working_dir": {"type": "string"}
+        "working_dir": {"type": "string"},
+        "override": {"type": "array", "items": {"type": ["string"], "uniqueItems": true, "format": "override"}}
       },
 
       "dependencies": {
--- a/compose/config/config_schema_v2.1.json
+++ b/compose/config/config_schema_v2.1.json
@@ -289,7 +289,8 @@
         "volumes": {"type": "array", "items": {"type": "string"}, "uniqueItems": true},
         "volume_driver": {"type": "string"},
         "volumes_from": {"type": "array", "items": {"type": "string"}, "uniqueItems": true},
-        "working_dir": {"type": "string"}
+        "working_dir": {"type": "string"},
+        "override": {"type": "array", "items": {"type": ["string"], "uniqueItems": true, "format": "override"}}
       },
 
       "dependencies": {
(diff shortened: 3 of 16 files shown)


nyetwurk:
As a side note, "ports" is an interesting case. I don't think
```
ports:
 - "80:80"
 - "8080:80"
```
has any utility anyway. After the first `xx:80`, subsequent `yy:80`s should override any previous ones...

The `ports:` syntax is kind of screwy in this regard; it should be a dict where the second port is be the key for the first port as a value, rather than just a list of strings.

It is probably too late to revisit the `ports` keyword, but perhaps a separate keyword that indicates a dict with port mappings would be more useful, and would not require strange override behavior.

jovobe:
Any feedback on this @shin-?

shin-:
Thanks @jovobe for picking up and updating the previous PR.

My issue with this and the previous proposals is that I find the design extremely confusing and UX extremely poor (same for #5354). If we're going ahead with such a significant change, I want us to get it right, and I don't think we've found the right answer yet.

If you're invested in getting this resolved, would you be willing to create an issue dedicated to discussing the design for this feature? I'd start with an RFE to formalize what we're trying to achieve, and then we can start brainstorming what that looks like concretely in a Compose definition.

nyetwurk:
I agree, and still think to solution is to move away from an array and towards a dict, which provides the end user with all possible override possibilities (without requiring an `override` directive), except for adding multiple external port mappings for the same internal port, which, IMO isn't really useful in this context anyway.

shcallaway:
I'll admit that my PRs (#5354 and #5355) are lacking on the design/UX side. I just wanted to get something out there in order to see if there's demand. I'm going to close those PRs for the time being. At some point, I'll consider revisiting this feature. :)

ijc:
Was a design issue (as proposed in https://github.com/docker/compose/pull/6213#issuecomment-426308766) ever opened?

Coming at this now it seems to me that this is a usecase which would be best served these days by [docker app](https://github.com/docker/app) where you would simply make the port a parameter.

Given both of those I'm going to close this PR for now, if `docker app` doesn't satisfy everyone's needs (from what I can see it does satisfy the authors) I would suggest opening a design issue for your particular use case either here on on the docker/app repo.

nyetwurk:
`docker app` does not satisfy anything.

nyetwurk:
What purpose does closing this serve?

PR #6217, opened 2018-09-24
6026 suppress progress when logging level info or debug

### LINKED ISSUE #6026: Feature Request: Provide --quiet option to 'up' command

BEllis:
Resolves #6026

Setting --log-level to WARNING, ERROR or CRITICAL will now hide progress messages.


BEllis:
This PR has been idle for a couple of months now. Any chance of getting it reviewed?

joelgenter:
Checking in on this as it'd be nice to have. Is there something preventing this from moving forward?

jamiegau:
Yes.  I would like to do that as renewing certbot.  YOU HAVE TO HAVE ACCESS TO PORT 80/443.  So you have to pull down the container, run certbot renew, and bring it up again.  Only was as its a security issue.  cannot use alternate ports.
And looking into this, best path has been to simply pull it down and renew and start up again during off hours.
Using &> to pipe output to some place can work here, but its not a good path as you cannot detect any output due to error bringin up or down the container.

BEllis:
@ijc This better? The branch name has the wrong issue number, but I assume that's not a biggy.

BEllis:
I still think adding a separate -q flag may be a better solution, as I can think of scenarios for suppressing the parallel logging at all logging levels.

gene1wood:
@BEllis Would it make sense to rebase and resolve the conflicts on this PR? @shin- or @ijc is there an intent to merge this feature (in which case Ben should fix up the conflicts that have occurred over the past 22 months)? Or is this not destined to make it into docker compose?

glours:

PR #6221, opened 2018-09-24
Preserve volumes order as declared in the compose file.

### LINKED ISSUE #6091: Random order of mounting volumes
## Description of the issue

I have a few volumes, some should be mounted inside folders of already defined (and expected to be mounted) volumes, but it seems (with --verbose) docker-compose uses some different logic than the order of which volumes mounts is declared.

```
volumes:
          # data
          - $PWD/mounts/abc-wpfrontend:/srv/cms/thecode/public_html/frontendwp
          # data from production nas
          - /Volumes/DataHD/frontendwp/wp-content/plugins:/srv/cms/thecode/public_html/frontendwp/wp-content/plugins
          - /Volumes/DataHD/frontendwp/wp-content/uploads:/srv/cms/thecode/public_html/frontendwp/wp-content/uploads
          # logs
          - $PWD/logs/frontendwp/supervisor:/var/log/supervisor
          - $PWD/logs/frontendwp/apache2:/var/log/apache2
          - $PWD/logs/frontendwp/newrelic:/var/log/newrelic
```
From --verbose log:
```
compose.cli.verbose_proxy.proxy_callable: docker create_host_config -> {'Binds': ['/Volumes/WebDevHD/abc-dev-env/logs/frontendwp/supervisor:/var/log/supervisor:rw',
           '/Volumes/WebDevHD/abc-dev-env/logs/frontendwp/apache2:/var/log/apache2:rw',
           '/Volumes/DataHD/frontendwp/wp-content/plugins:/srv/cms/thecode/public_html/frontendwp/wp-content/plugins:rw',
           '/Volumes/WebDevHD/abc-dev-env/mounts/abc-wpfrontend:/srv/cms/thecode/public_html/frontendwp:rw',
           '/Volumes/WebDevHD/abc-dev-env/logs/frontendwp/newrelic:/var/log/newrelic:rw',
           '/Volumes/DataHD/frontendwp/wp-content/uploads:/srv/cms/thecode/public_html/frontendwp/wp-content/uploads:rw'],
```

```
```
docker-compose version 1.21.1, build 5a3f1a3
docker-py version: 3.3.0
CPython version: 3.6.4
```
Output of "docker version"
```
Client:
 Version:      18.03.1-ce
 API version:  1.37
 Go version:   go1.9.5
 Git commit:   9ee9f40
 Built:        Thu Apr 26 07:13:02 2018
 OS/Arch:      darwin/amd64
 Experimental: false
 Orchestrator: swarm

Server:
 Engine:
  Version:      18.03.1-ce
  API version:  1.37 (minimum version 1.12)
  Go version:   go1.9.5
  Git commit:   9ee9f40
  Built:        Thu Apr 26 07:22:38 2018
  OS/Arch:      linux/amd64
  Experimental: true
```
Output of "docker-compose config"
```

### Observed result

As it's going every run by the new way, sometimes I have properly mounted folders, and sometimes not, sometimes only one of them properly mounted as in the example above.

### Expected result

Mount volumes in the strict order as specified in yml file.

```
(if applicable)
```

## Additional information

Dimrok:
Resolves #6091.

 `tests/unit/service_test.py`, in `ServiceTest::test_volumes_order_is_preserved`

before the fix:
```
['/v0:/v0:rw', '/v1:/v1:rw', '/v2:/v2:rw', '/v3:/v3:rw', '/v4:/v4:rw', '/v5:/v5:rw']
['/v3:/v3:rw', '/v2:/v2:rw', '/v4:/v4:rw', '/v5:/v5:rw', '/v0:/v0:rw', '/v1:/v1:rw']
```

after the fix:
```
['/v0:/v0:rw', '/v1:/v1:rw', '/v2:/v2:rw', '/v3:/v3:rw', '/v4:/v4:rw', '/v5:/v5:rw']
['/v0:/v0:rw', '/v1:/v1:rw', '/v2:/v2:rw', '/v3:/v3:rw', '/v4:/v4:rw', '/v5:/v5:rw']
```

### DIFF
--- a/compose/utils.py
+++ b/compose/utils.py
@@ -176,6 +176,7 @@ def unique_everseen(iterable, key=lambda x: x):
     "List unique elements, preserving order. Remember all elements ever seen."
     seen = set()
     for element in iterable:
-        if key(element) not in seen:
-            seen.add(element)
+        unique_key = key(element)
+        if unique_key not in seen:
+            seen.add(unique_key)
             yield element
--- a/compose/utils.py
+++ b/compose/utils.py
@@ -170,3 +170,12 @@ def truncate_id(value):
     if len(value) > 12:
         return value[:12]
     return value
+
+
+def unique_everseen(iterable, key=lambda x: x):
+    "List unique elements, preserving order. Remember all elements ever seen."
+    seen = set()
+    for element in iterable:
+        if key(element) not in seen:
+            seen.add(element)
+            yield element
--- a/compose/service.py
+++ b/compose/service.py
@@ -56,6 +56,7 @@ from .utils import json_hash
 from .utils import parse_bytes
 from .utils import parse_seconds_float
 from .utils import truncate_id
+from .utils import unique_everseen
 
 
 log = logging.getLogger(__name__)
@@ -940,8 +941,9 @@ class Service(object):
                 override_options['mounts'] = override_options.get('mounts') or []
                 override_options['mounts'].extend([build_mount(v) for v in secret_volumes])
 
-        # Remove possible duplicates (see e.g. https://github.com/docker/compose/issues/5885)
-        override_options['binds'] = list(set(binds))
+        # Remove possible duplicates (see e.g. https://github.com/docker/compose/issues/5885).
+        # unique_everseen preserves order. (see https://github.com/docker/compose/issues/6091).
+        override_options['binds'] = list(unique_everseen(binds))
         return container_options, override_options
 
     def _get_container_host_config(self, override_options, one_off=False):
--- a/compose/service.py
+++ b/compose/service.py
@@ -1429,7 +1429,7 @@ def merge_volume_bindings(volumes, tmpfs, previous_container, mounts):
     """
     affinity = {}
 
-    volume_bindings = dict(
+    volume_bindings = OrderedDict(
         build_volume_binding(volume)
         for volume in volumes
         if volume.external
--- a/tests/unit/config/config_test.py
+++ b/tests/unit/config/config_test.py
@@ -8,6 +8,7 @@ import os
 import shutil
 import tempfile
 from operator import itemgetter
+from random import shuffle
 
 import py
 import pytest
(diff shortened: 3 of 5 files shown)


shin-:
@Dimrok Can you rebase against master so #6225 gets included? That should clear up the CI failure.

Thanks!

Dimrok:
@shin- Done. And it is now deterministic on py27 too.

compose/utils.py
```diff
@@ -170,3 +170,12 @@ def truncate_id(value):
     if len(value) > 12:
         return value[:12]
     return value
+
+
+def unique_everseen(iterable, key=lambda x: x):
```
shin-: Can you explain what the intent is for the `key` parameter? The only place you seem to use it is in the unit tests.

compose/utils.py
```diff
@@ -170,3 +170,12 @@ def truncate_id(value):
...
+    "List unique elements, preserving order. Remember all elements ever seen."
+    seen = set()
+    for element in iterable:
+        if key(element) not in seen:
```
shin-: You're testing for the presence of `key(element)` in the `set` but then add `element` to it, which won't work:
```
In [7]: def keyf(x):
   ...:     return 'key:' + x

In [8]: list(unique_everseen(['a', 'b', 'c', 'a'], keyf))
Out[8]: ['a', 'b', 'c', 'a']
```

compose/utils.py
shin-: And can we rename this to `remove_duplicates`? :innocent:

shin-: Oh, I just realized that it's based on the function with the same name in https://github.com/erikrose/more-itertools ; name is fine, in that respect. The `key` param should be fixed to work correctly (see below) or removed if we don't think we'll use it.

compose/utils.py
Dimrok: Oups, you are right. Fixing that.

I had to adapt for it to be python2.7-compliant and I missed that. 
I'll update the tests too.

compose/utils.py
Dimrok: Fixed by 5b9b519.

compose/utils.py
```diff
@@ -170,3 +170,13 @@ def truncate_id(value):
     if len(value) > 12:
         return value[:12]
     return value
+
+
+def unique_everseen(iterable, key=lambda x: x):
+    "List unique elements, preserving order. Remember all elements ever seen."
+    seen = set()
+    for element in iterable:
+        unique_key = key(element)
```
shin-: nit: `k` as a var name would have been completely fine here.

shin-:
Thank you 💯

PR #6234, opened 2018-09-27
Improved expanding source paths of volumes

### LINKED ISSUE #6233: Expanding volumes source paths
## Description of the issue
When volume defined with long syntax, its source paths which starts with `~` do not expand well as a paths started with `.`

```
docker-compose version 1.22.0, build f46880f
```

```
version: "3.4"

services:
  test:
    image: "test:${DISTRO:-trusty}"
    build:
      context: test/
      args: { version: "${DISTRO:-trusty}" }
    volumes:
      - type: bind
        read_only: true
        source: ~/tmp/test1
        target: /tmp/test1
      
      - ~/tmp/test2:/tmp/test2:ro
```

1. Define volume with use of long syntax and set source started with `~`
2. `docker-compose up -d --force-recreate <service>`

### Observed result
```
Recreating test_test_1 ... error

ERROR: for test_test_1  Cannot create container for service test: invalid mount config for type "bind": invalid mount path: '~/tmp/test1' mount path must be absolute

ERROR: for test  Cannot create container for service test: invalid mount config for type "bind": invalid mount path: '~/tmp/test1' mount path must be absolute
```

### Expected result
```
Recreating 48d9f6c670d9_test_test_1 ... done
```

## Additional information

`Ubuntu 18.04.1 LTS`

zasca:
defined with long syntax when paths starts with '~'
Resolves #6233

### DIFF
--- a/tests/unit/config/config_test.py
+++ b/tests/unit/config/config_test.py
@@ -1322,6 +1322,29 @@ class ConfigTest(unittest.TestCase):
         assert mount.type == 'bind'
         assert mount.source == expected_source
 
+    def test_load_bind_mount_relative_path_with_tilde(self):
+        base_file = config.ConfigFile(
+            'base.yaml', {
+                'version': '3.4',
+                'services': {
+                    'web': {
+                        'image': 'busybox:latest',
+                        'volumes': [
+                            {'type': 'bind', 'source': '~/web', 'target': '/web'},
+                        ],
+                    },
+                },
+            },
+        )
+
+        details = config.ConfigDetails('.', [base_file])
+        config_data = config.load(details)
+        mount = config_data.services[0].get('volumes')[0]
+        assert mount.target == '/web'
+        assert mount.type == 'bind'
+        assert (not mount.source.startswith('~')
+                and mount.source.endswith('{}web'.format(os.path.sep)))
+
     def test_config_invalid_ipam_config(self):
         with pytest.raises(ConfigurationError) as excinfo:
             config.load(
--- a/compose/config/config.py
+++ b/compose/config/config.py
@@ -1281,7 +1281,7 @@ def resolve_volume_paths(working_dir, service_dict):
 
 def resolve_volume_path(working_dir, volume):
     if isinstance(volume, dict):
-        if volume.get('source', '').startswith('.') and volume['type'] == 'bind':
+        if volume.get('source', '').startswith(('.', '~')) and volume['type'] == 'bind':
             volume['source'] = expand_path(working_dir, volume['source'])
         return volume
 


shin-:
Thanks! Can you try and add a test for this as well?

zasca:
If my unit test is correct, then path expanding do not properly work on Windows platform and need more complex patching.

tests/unit/config/config_test.py
```diff
@@ -1322,6 +1322,36 @@ def test_load_bind_mount_relative_path(self):
         assert mount.type == 'bind'
         assert mount.source == expected_source
 
+    @mock.patch.dict(os.environ)
+    def test_load_bind_mount_relative_path_with_tilda(self):
+        if IS_WINDOWS_PLATFORM:
```
shin-: I think we can do without that - assume that Python expands `~` correctly on all platforms, and change the test to verify that `mount.source` doesn't start with `~` and has `/web` as the last element of its path.

zasca: I updated the test as you suggest, but the result is the same.

zasca:
Now everything looks fine.

tests/unit/config/config_test.py
```diff
@@ -1322,6 +1322,29 @@ def test_load_bind_mount_relative_path(self):
...
+            },
+        )
+
+        details = config.ConfigDetails('.', [base_file])
+        config_data = config.load(details)
+        mount = config_data.services[0].get('volumes')[0]
+        assert mount.target == '/web'
+        assert mount.type == 'bind'
+        assert (not mount.source.startswith('~')
+                and mount.source.endswith('\\web'))
```
shin-: That's going to fail on UNIX systems. Use `endswith('{}web'.format(os.path.sep))` instead?

tests/unit/config/config_test.py
```diff
@@ -1322,6 +1322,29 @@ def test_load_bind_mount_relative_path(self):
         assert mount.type == 'bind'
         assert mount.source == expected_source
 
+    def test_load_bind_mount_relative_path_with_tilda(self):
```
shin-: nit: tild**e**

tests/unit/config/config_test.py
zasca: Your truth. Thanks.

tests/unit/config/config_test.py
zasca: Fixed.

zasca:
This PR with last changes is good for merging now?

zasca:
> @zasca Looks good now yeah - can you squash this down to a single commit?

Done

shin-:
Thanks!

Issue #6252, opened 2018-10-05
Use official Docker image as docker/compose base image


nvllsvm:
## Description of the issue
The official Docker image can be used to simplify the Docker Compose image (`docker/compose`). The official image is also based on Alpine Linux.

### Observed result
Building `Dockerfile.run` inherits from `alpine:3.6` and downloads the Docker binary from `https://download.docker.com/linux/static/stable/x86_64/docker-17.12.1-ce.tgz`.

### Expected result
Building `Dockerfile.run` inherits from `docker:17.12.1` and uses the inherited Docker binary

shin-:
The `docker` image includes a lot of things we don't need (basically the whole daemon / moby part of Docker). `docker/compose:1.23.0-rc1` is 25 MB including the `docker-compose` executable, whereas `docker:17.12.1` is already 37MB by itself (and newer versions have grown another ~10).

Based on that, I don't think the change is warranted.

nvllsvm:
@shin- Good call. Updated to use the binary from the official image via multi-stage.

PR #6253, opened 2018-10-05
Use docker binary from official Docker image

nvllsvm:
Resolves https://github.com/docker/compose/issues/6252

### DIFF
--- a/Dockerfile.run
+++ b/Dockerfile.run
@@ -1,7 +1,7 @@
+FROM docker:17.12.1 as docker
 FROM alpine:3.6
 
 ENV GLIBC 2.27-r0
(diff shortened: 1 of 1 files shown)


shin-:
We may be able to use a multi-stage build to leverage the official image instead ; something like this:
```Dockerfile
FROM docker:17.12.1 AS docker
FROM alpine:3.6

ENV GLIBC 2.27-r0

COPY --from=docker /usr/local/bin/docker /usr/local/bin/docker
...
```

That way we avoid repackaging unnecessary binaries like `runc` or `containerd`

shin-:
LGTM, but please add the signoff line to your commit as indicated here: https://github.com/docker/compose/pull/6253#issuecomment-427435165

nvllsvm:
Signed off.

Issue #6106, opened 2018-07-25
Tag release


ssadok:
Hi,
Can you tag stable and latest please for your releases ?

Thanks.

shin-:
Unless I'm missing something, you can just use GitHub's permalink for that: https://github.com/docker/compose/releases/latest

nvllsvm:
@shin- Is there any reason why there is no `latest` tag for the Docker image of Docker Compose?

shin-:
@nvllsvm There is: https://github.com/docker/compose/issues/5987#issuecomment-392113124

Issue #6006, opened 2018-06-06
Can't pull docker compose image from dockerhub


egel:
## Description of the issue

This issue is directly related to images on docker hub (https://hub.docker.com/r/docker/compose/)

Users can't pull the image from docker. Missing the image with tag `latest` 

## Context information (for bug reports)
-

## Steps to reproduce the issue

1. `docker pull docker/compose`
2. Observe the error in terminal

### Observed result
The error in terminal
    
    → docker pull docker/compose
    Using default tag: latest
    Error response from daemon: manifest for docker/compose:latest not found

### Expected result

    → docker pull docker/compose:latest
    latest: Pulling from docker/compose
    605ce1bd3f31: Already exists
    ed764f565fce: Pull complete
    988404a576a5: Pull complete
    d69030722a7d: Pull complete
    2d6379878250: Pull complete
    d14800f53520: Pull complete
    Digest: sha256:97f49d34b6897df9cf023489075eb4dca15760c6e40b553313af8e760614a2e5
    Status: Downloaded newer image for docker/compose:latest

## Suggested solution
In each stable relase push container with the `latest` tag.

## Additional information

OS: Mac OSX 10.13.4
docker version: 

    → docker info
    Containers: 12
     Running: 6
     Paused: 0
     Stopped: 6
    Images: 229
    Server Version: 18.03.1-ce
    Storage Driver: overlay2
     Backing Filesystem: extfs
     Supports d_type: true
     Native Overlay Diff: true
    Logging Driver: json-file
    Cgroup Driver: cgroupfs
    Plugins:
     Volume: local
     Network: bridge host ipvlan macvlan null overlay
     Log: awslogs fluentd gcplogs gelf journald json-file logentries splunk syslog
    Swarm: inactive
    Runtimes: runc
    Default Runtime: runc
    Init Binary: docker-init
    containerd version: 773c489c9c1b21a6d78b5c538cd395416ec50f88
    runc version: 4fc53a81fb7c994640722ac585fa9ca548971871
    init version: 949e6fa
    Security Options:
     seccomp
      Profile: default
    Kernel Version: 4.9.87-linuxkit-aufs
    Operating System: Docker for Mac
    OSType: linux
    Architecture: x86_64
    CPUs: 4
    Total Memory: 1.952GiB
    Name: linuxkit-025000000001
[...]

Issue #4287, opened 2017-01-03
Make https://hub.docker.com/r/docker/compose/ more usable


tmaier:
https://hub.docker.com/r/docker/compose/
Common use case for this repo: Run docker-compose within a CI pipeline, which is based on docker-in-docker.

As MVP, please do the following:
- [ ] Add a reference to the `Dockerfile` which is being used here
- [ ] Add automated build from repo
- [ ] Add automated build when the dependent image `alpine` has been updated
- [ ] Add a `latest` tag

On top of this, please consider
- [ ] Add explanation how to use this together with `docker:dind`

tmaier:
This might get obsolete if https://github.com/docker-library/docker/pull/30 gets merged.

mathroc:
I would also suggest to follow [Dockerfile best practices about entrypoint](https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#/entrypoint)

right now I can't use it in gitlab CI because it tries to run sh in the image but it results in trying to run `docker-compose sh`

tianon:
I'd be curious to know the opinion of Compose maintainers on https://github.com/docker-library/docker/pull/30 -- should `docker:compose` be created?  Should we instead continue to punt users asking for it to the officially-upstream-maintained `docker/compose` image?  Should we look to create a new `docker-compose` official image?

(Happy to help with or implement any of these, but I'd rather not move forward on any of this without the support and/or blessing of the Compose maintainers. :smile:)

yxliang01:
+1 for making docker-compose image as an official image @tianon

tianon:
My comment at https://github.com/docker/compose/issues/4287#issuecomment-305615695 is still pretty much verbatim the same set of questions I'd have for maintainers here (although I'd probably be less keen on the `docker:compose` variant now).

I'm happy to do the legwork to make/maintain an image for this, but I don't want to step on any toes or go against maintainer's wishes.

Issue #5987, opened 2018-05-25
Instead of version number why don't use 'latest' it will help writing scripts to install docker-compose with the latest version.


Kuchiriel:
Instead of version number why don't use 'latest' it will help writing scripts to install docker-compose with the latest version.

shin-:
If you're talking about the `docker/compose` image on Hub, it's meant to be exclusively used through the `run.sh` script, which is versioned.

We currently have no plans to add a `latest` tag.

Kuchiriel:
I am talking about the releases tab.

nvllsvm:
@shin- The run.sh cannot be used as a base image in, for example, a GitLab CI pipeline.

PR #6331, opened 2018-11-08
Push latest tag for docker/compose

nvllsvm:
Tag and push the `latest` tag when building and releasing `docker/compose` images.

Resolves https://github.com/docker/compose/issues/6106
Resolves https://github.com/docker/compose/issues/6006
Partly resolves https://github.com/docker/compose/issues/4287

### DIFF
--- a/script/build/image
+++ b/script/build/image
@@ -15,3 +15,4 @@ VERSION="$(python setup.py --version)"
 python setup.py sdist bdist_wheel
 ./script/build/linux
 docker build -t docker/compose:$TAG -f Dockerfile.run .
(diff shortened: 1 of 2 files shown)


shin-:
Thanks!
As explained [here](https://github.com/docker/compose/issues/5987#issuecomment-392113124), that's not something we're currently looking to do.

nvllsvm:
That's rather unfortunate as many of us in the community find that the run.sh script is completely inadequate for our needs. 

What's the big push for that anyway? Is there a grander vision I'm missing?

nvllsvm:
@rumpl @ulyssessouza - Any thoughts on reopening this?

PR #6346, opened 2018-11-16
Show failed services 'docker-compose start' when containers are not availabe

### LINKED ISSUE #5469: Docker-compose start strange behviour when containers are not available
When I use `docker-compose start ` before a `docker-compose up` I obtain the following result:
```
Starting mongo         ... done
Starting elasticsearch ... done
Starting transporter   ... done
Starting pdf           ... done
Starting back          ... done
Starting front         ... done
ERROR: No containers to start
```
This behaviour is pretty strange as the output is contradictory it says that it start all containers but the errors say the opposite.

collin5:
Resolves #5469

### DIFF
--- a/compose/parallel.py
+++ b/compose/parallel.py
@@ -50,7 +50,11 @@ def parallel_execute_watch(events, writer, errors, results, msg, get_name):
     error_to_reraise = None
     for obj, result, exception in events:
         if exception is None:
-            writer.write(msg, get_name(obj), 'done', green)
+            if callable(getattr(obj, 'containers', None)) and not obj.containers():
+                # If service has no containers started
+                writer.write(msg, get_name(obj), 'failed', red)
+            else:
+                writer.write(msg, get_name(obj), 'done', green)
             results.append(result)
         elif isinstance(exception, ImageNotFound):
             # This is to bubble up ImageNotFound exceptions to the client so we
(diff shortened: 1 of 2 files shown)


compose/parallel.py
```diff
@@ -49,8 +49,12 @@ def parallel_execute_watch(events, writer, errors, results, msg, get_name):
     """
     error_to_reraise = None
     for obj, result, exception in events:
+        from compose.service import Service # noqa
         if exception is None:
-            writer.write(msg, get_name(obj), 'done', green)
+            if isinstance(obj, Service) and not obj.containers():
```
shin-: Can we use duck-typing here instead and avoid the in-block import? e.g.

```python
if hasattr(obj, 'containers') and callable(obj.containers) and not obj.containers():
  ...
```

collin5:
Fixed. Thanks!

compose/parallel.py
```diff
@@ -50,7 +50,10 @@ def parallel_execute_watch(events, writer, errors, results, msg, get_name):
     error_to_reraise = None
     for obj, result, exception in events:
         if exception is None:
-            writer.write(msg, get_name(obj), 'done', green)
+            if callable(getattr(obj, 'containers', None)) and not obj.containers():
```
shin-: Sorry, one last thing - can you add a comment here to explain what problem this solves, as a reminder for our future selves? :)

collin5:
Fixed! Thanks 👍

PR #6351, opened 2018-11-18
Small improvements to README.md

hartwork:
Hi!

Just a few small improvements to the readme. Happy to drop unwanted parts or make adjustments.

Best, Sebastian

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ A `docker-compose.yml` looks like this:
         image: redis
 
 For more information about the Compose file, see the
-[Compose file reference](https://github.com/docker/docker.github.io/blob/master/compose/compose-file/compose-versioning.md)
+[Compose file reference](https://github.com/docker/docker.github.io/blob/master/compose/compose-file/compose-versioning.md).
 
 Compose has commands for managing the whole lifecycle of your application:
 
@@ -48,8 +48,8 @@ Installation and documentation
 ------------------------------
 
 - Full documentation is available on [Docker's website](https://docs.docker.com/compose/).
-- Code repository for Compose is on [GitHub](https://github.com/docker/compose)
-- If you find any problems please fill out an [issue](https://github.com/docker/compose/issues/new)
+- Code repository for Compose is on [GitHub](https://github.com/docker/compose).
+- If you find any problems please fill out an [issue](https://github.com/docker/compose/issues/new). Thank you!
 
 Contributing
 ------------
--- a/README.md
+++ b/README.md
@@ -48,7 +48,6 @@ Installation and documentation
 ------------------------------
 
 - Full documentation is available on [Docker's website](https://docs.docker.com/compose/).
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -28,14 +28,14 @@ A `docker-compose.yml` looks like this:
       web:
         build: .
         ports:
-         - "5000:5000"
```
shin-: The quotes are consistent with [our advice regarding port mappings](https://docs.docker.com/compose/compose-file/#short-syntax-1). 

> Note: When mapping ports in the `HOST:CONTAINER` format, you may experience erroneous results when using a container port lower than 60, because YAML parses numbers in the format `xx:yy` as a base-60 value. For this reason, we recommend always explicitly specifying your port mappings as strings.

In this instance, even though the two are equivalent, I think we should "practice what we preach" and use the quoted version.

README.md
```diff
@@ -48,9 +48,9 @@ Installation and documentation
 ------------------------------
 
 - Full documentation is available on [Docker's website](https://docs.docker.com/compose/).
-- If you have any questions, you can talk in real-time with other developers in the #docker-compose IRC channel on Freenode. [Click here to join using IRCCloud.](https://www.irccloud.com/invite?hostname=irc.freenode.net&channel=%23docker-compose)
-- Code repository for Compose is on [GitHub](https://github.com/docker/compose)
-- If you find any problems please fill out an [issue](https://github.com/docker/compose/issues/new)
+- If you have any questions, you can talk in real-time with other developers in the `#docker-compose` IRC channel on Freenode. [Click here to join using IRCCloud](https://www.irccloud.com/invite?hostname=irc.freenode.net&channel=%23docker-compose).
```
shin-: I don't know how active this is. It might make sense to remove this entirely.

README.md
```diff
@@ -48,9 +48,9 @@ Installation and documentation
...
+- Code repository for Compose is on [GitHub](https://github.com/docker/compose).
+- If you find any problems please fill out an [issue](https://github.com/docker/compose/issues/new). Thank you!
```
shin-: While we're at it, can you update that link to use https://github.com/docker/compose/issues/new/choose instead? Thanks!

README.md
hartwork: That's hard to argue against, okay :+1:

README.md
hartwork: Right now it's 40 nicks, 0 ops. Are you undecided or decided for removal? Please let me know whether you want me to drop IRC or not.

README.md
hartwork: That sounds doable :smiley:

README.md
shin-: Let's remove it as I don't think we have any official presence there anymore. Thanks 👍

hartwork: Done

shin-:
Looking good. Thank you!

PR #6359, opened 2018-11-22
Avoids pushing the same image more than once.

### LINKED ISSUE #6245: Docker-compose push tries to push the same image multiple times
## Description of the issue
If your docker-compose.yml file contains multiple usages of the same image used for multiple services, then when you docker-compose push it tries to push that image once for every time it is used. The server side processing of this is handled correctly for me, saying layer already exists, but it is still added wasted time as it verifies each layer N times where N = number of times you referenced the image.

```
```
❯ docker-compose version
docker-compose version 1.22.0, build f46880fe
docker-py version: 3.4.1
CPython version: 3.6.6
OpenSSL version: OpenSSL 1.1.0f  25 May 2017
```
Output of "docker version"
```
 docker version
Client:
 Version:           18.06.1-ce
 API version:       1.38
 Go version:        go1.10.3
 Git commit:        e68fc7a
 Built:             Tue Aug 21 17:25:02 2018
 Experimental:      false

Server:
 Engine:
  Version:          18.06.1-ce
  API version:      1.38 (minimum version 1.12)
  Go version:       go1.10.3
  Git commit:       e68fc7a
  Built:            Tue Aug 21 17:26:30 2018
  OS/Arch:          linux/amd64
  Experimental:     false
```
Output of "docker-compose config"
```
Omitting as I haven't made a minimal reproduction, let me know if you need one.

1.Define two services in docker-compose.yml using the same image (specifically a private registry one, but shouldn't matter)
2.run docker-compose push
3.

### Observed result
One push per usage of image:tag

### Expected result
One push per unique image:tag

```
(if applicable)
```

## Additional information

Fedora f28, dnf

ulyssessouza:
Keep of already pushed images in a list `unique_images` to avoid re-pushing it.
**Note**: The images `nginx` and `nginx:latest` will be considered as the same, since `latest` is a default tag 

Resolves #6245

### DIFF
--- a/compose/project.py
+++ b/compose/project.py
@@ -29,6 +29,7 @@ from .service import ContainerNetworkMode
 from .service import ContainerPidMode
 from .service import ConvergenceStrategy
 from .service import NetworkMode
+from .service import parse_repository_tag
 from .service import PidMode
 from .service import Service
 from .service import ServiceNetworkMode
@@ -592,8 +593,15 @@ class Project(object):
                 service.pull(ignore_pull_failures, silent=silent)
 
     def push(self, service_names=None, ignore_push_failures=False):
+        unique_images = set()
         for service in self.get_services(service_names, include_deps=False):
-            service.push(ignore_push_failures)
+            # Considering <image> and <image:latest> as the same
+            repo, tag, sep = parse_repository_tag(service.image_name)
+            service_image_name = sep.join((repo, tag)) if tag else sep.join((repo, 'latest'))
+
+            if service_image_name not in unique_images:
+                service.push(ignore_push_failures)
+                unique_images.add(service_image_name)
 
     def _labeled_containers(self, stopped=False, one_off=OneOffFilter.exclude):
         ctnrs = list(filter(None, [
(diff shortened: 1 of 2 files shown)


compose/project.py
```diff
@@ -592,8 +592,17 @@ def pull_service(service):
                 service.pull(ignore_pull_failures, silent=silent)
 
     def push(self, service_names=None, ignore_push_failures=False):
+        unique_images = []
```
shin-: Not critical, but this would be better as a `set()`

compose/project.py
```diff
@@ -592,8 +592,17 @@ def pull_service(service):
...
         for service in self.get_services(service_names, include_deps=False):
-            service.push(ignore_push_failures)
+            # Considering <image> and <image:latest> as the same
+            image_name_array = service.image_name.split(":")
```
shin-: You should use [parse_repository_tag()](https://github.com/docker/compose/blob/2b604c1e8b0df4cf0da0b657634544218fc3a79d/compose/service.py#L1397-L1421)

```python
repo, tag, sep = parse_repository_tag(service.image_name)
service_image_name = sep.join(repo, tag) if tag else sep.join(repo, 'latest')
```

compose/project.py
ulyssessouza: Yep!

compose/project.py
ulyssessouza: Cool! I was looking for something to parse it but didn't see that.

shin-:
Thanks! I just moved the test to the unit tests folder as it doesn't actually interact with the Docker client (and so we don't start accidentally trying to re-push busybox to the hub in our CI :))

PR #6364, opened 2018-11-22
Avoids misleading warning concerning env vars when perfoming an `exec` command

### LINKED ISSUE #6350: Incorrect environment variable warnings during `exec`
## Description of the issue
Compose warns of unset environment variables when running `exec` in an existing container even if the environment variables are already set.

```
docker-compose version 1.23.1, build b02f1306
docker-py version: 3.5.0
CPython version: 3.6.7
OpenSSL version: OpenSSL 1.1.0f  25 May 2017
```

```
Client:
 Version:           18.09.0
 API version:       1.39
 Go version:        go1.10.4
 Git commit:        4d60db4
 Built:             Wed Nov  7 00:49:01 2018
 Experimental:      false

 Engine:
  Version:          18.09.0
  Go version:       go1.10.4
  Git commit:       4d60db4
  Built:            Wed Nov  7 00:16:44 2018
  OS/Arch:          linux/amd64
  Experimental:     false
```

```
services:
  db:
    environment:
      MYSQL_DATABASE: database
      MYSQL_ROOT_PASSWORD: root
      TEST_VARIABLE: ''
    image: mysql:5.7.22
version: '3.2'
```

**Ouput of `cat docker-compose.yml`**

```
version: '3.2'

services:

  db:
    image: mysql:5.7.22
    environment:
      MYSQL_ROOT_PASSWORD: root
      MYSQL_DATABASE: database
      TEST_VARIABLE: ${TEST_VARIABLE}
```

1. Copy the contents above into a `docker-compose.yml` file.
2. Start the service with `TEST_VARIABLE=test docker-compose up -d`.
3. Exec any command into the `db` container, for example: `docker-compose exec db bash`

### Observed result
The system warns me with `WARNING: The TEST_VARIABLE variable is not set. Defaulting to a blank string.`
Running `env` in the container shows that the variable is present:
```
$ root@5b357d005cff:/# env
...
TEST_VARIABLE=test
...
```

### Expected result
I only expect to see this warning if the environment variable was not supplied when the services were started. The warning claims that the variable is not exported.

## Additional information

peek 2018-11-18 14-10

ulyssessouza:
Detects the execution on an`exec` command and sets the environment instance to `silent` mode.

Resolves #6350

### DIFF
--- a/compose/cli/command.py
+++ b/compose/cli/command.py
@@ -21,10 +21,26 @@ from .utils import get_version_info
 
 log = logging.getLogger(__name__)
 
+SILENT_COMMANDS = set((
+    'events',
+    'exec',
+    'kill',
+    'logs',
+    'pause',
+    'ps',
+    'restart',
+    'rm',
+    'start',
+    'stop',
+    'top',
+    'unpause',
+))
+
 
 def project_from_options(project_dir, options):
     override_dir = options.get('--project-directory')
     environment = Environment.from_env_file(override_dir or project_dir)
+    environment.silent = options.get('COMMAND', None) in SILENT_COMMANDS
     set_parallel_limit(environment)
 
     host = options.get('--host')
--- a/tests/integration/environment_test.py
+++ b/tests/integration/environment_test.py
@@ -1,0 +1,52 @@
+from __future__ import absolute_import
+from __future__ import unicode_literals
+
+import tempfile
+
+from ddt import data
+from ddt import ddt
+
+from .. import mock
+from compose.cli.command import project_from_options
+from tests.integration.testcases import DockerClientTestCase
+
+
+@ddt
+class EnvironmentTest(DockerClientTestCase):
+    @classmethod
+    def setUpClass(cls):
+        super(EnvironmentTest, cls).setUpClass()
+        cls.compose_file = tempfile.NamedTemporaryFile(mode='w+b')
+        cls.compose_file.write(bytes("""version: '3.2'
+services:
+  svc:
+    image: busybox:latest
+    environment:
+      TEST_VARIABLE: ${TEST_VARIABLE}""", encoding='utf-8'))
+        cls.compose_file.flush()
+
+    @classmethod
+    def tearDownClass(cls):
+        super(EnvironmentTest, cls).tearDownClass()
+        cls.compose_file.close()
+
+    @data('events',
+          'exec',
+          'kill',
+          'logs',
+          'pause',
+          'ps',
+          'restart',
+          'rm',
+          'start',
+          'stop',
+          'top',
+          'unpause')
+    def _test_no_warning_on_missing_host_environment_var_on_silent_commands(self, cmd):
+        options = {'COMMAND': cmd, '--file': [EnvironmentTest.compose_file.name]}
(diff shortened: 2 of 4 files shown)


compose/cli/command.py
```diff
@@ -25,6 +25,7 @@
 def project_from_options(project_dir, options):
     override_dir = options.get('--project-directory')
     environment = Environment.from_env_file(override_dir or project_dir)
+    environment.silent = options['COMMAND'] == 'exec'
```
mmphego: Perhaps, instead of `environment.silent = options['COMMAND'] == 'exec'`
to avoid `KeyErrors` consider:
 `environment.silent = options.get('COMMAND', 'exec') == 'exec'`

ulyssessouza: Hmmm... I agree about avoiding `KeyErrors`, but I would use:
`environment.silent = options.get('COMMAND', None) == 'exec'`

Since we would like to silence it if and only if it's an `exec` command.
WDYT @mmphego ?

mmphego: Fully agree.

shin-:
@ulyssessouza Thanks! Could you add a test for this? You can use [this test](https://github.com/docker/compose/blob/7ae632a9ee7530fcf81e212baa3e588f477ea862/tests/unit/project_test.py#L544-L550) as a guideline.

ulyssessouza:
@shin- Could you please tell me if there is a better way to implement this test?
In any case, this one is working fine :)

DomHudson:
Thanks for this! Do you think it would be worth adding this for the `logs` command also?

tests/integration/environment_test.py
```diff
@@ -0,0 +1,33 @@
...
+                                  '--privileged': False,
+                                  '--user': None,
+                                  '--workdir': None,
+                                  '--publish': None,
+                                  '--service-ports': None,
+                                  '-T': False,
+                                  'ARGS': [],
+                                  'COMMAND': 'bash',
+                                  'SERVICE': 'db'})
+            assert fake_log.warn.call_count == 0
```
mmphego: Perhaps instead of using `assert`, due to the fact that it might/will be removed when compiling to optimized byte code i.e. `assert` statements can be globally disabled with the `-O` command-line flag.

Rather consider doing it this way.
```python
if not fake_log.warn.call_count:
    raise AssertionError()
```

DomHudson: Just interested in this comment - do you run tests with the optimisation flag? Surely that'll cause almost all tests to be useless? Am I missing something?

mmphego: There are some tests that require a bit of optimization, in those cases, I do run with `-O` flag set and if I need an assertion I just raise it.

shin-: Thanks. In our case, consistency and having recognizable idioms is far more important than shaving a few seconds off our test suite.

ulyssessouza: @mmphego I understand your point, however I think that this kind of optimization has to be done once and for all in the whole project and a guideline written when this kind of change makes itself necessary.
Changing the check style just for this issue would introduce an heterogeneous style.

WDYT @shin- ?

PR #6366, opened 2018-11-23
 Updated Shell scripts 

mmphego:
- Double quote to prevent globbing and word splitting.
- Changed bash shebang from `#!/bin/bash` to `#!/usr/bin/env bash` for [portability](https://en.wikipedia.org/wiki/Shebang_%28Unix%29#Portability)

Resolves #

### DIFF
--- a/script/build/linux-entrypoint
+++ b/script/build/linux-entrypoint
@@ -5,11 +5,15 @@ set -ex
 TARGET=dist/docker-compose-$(uname -s)-$(uname -m)
 VENV=/code/.tox/py36
 
-mkdir -p `pwd`/dist
-chmod 777 `pwd`/dist
+mkdir -p "$(pwd)"/dist
+chmod 777 "$(pwd)"/dist
 
-$VENV/bin/pip install -q -r requirements-build.txt
-./script/build/write-git-sha
-su -c "$VENV/bin/pyinstaller docker-compose.spec" user
-mv dist/docker-compose $TARGET
-$TARGET version
+if [ ! -f "${VENV}/bin/pip" ]; then
+    "${VENV}"/bin/pip install -q -r requirements-build.txt
+    ./script/build/write-git-sha
+    su -c "${VENV}/bin/pyinstaller docker-compose.spec" user
+    mv dist/docker-compose "${TARGET}"
+    "${TARGET}" version
+else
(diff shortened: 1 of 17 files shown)


script/build/linux-entrypoint
```diff
@@ -5,11 +5,15 @@ set -ex
...
+chmod 777 "$(pwd)"/dist
+
+if [ ! -f "${VENV}/bin/pip" ]; then
+    "${VENV}"/bin/pip install -q -r requirements-build.txt
+    ./script/build/write-git-sha
+    su -c "${VENV}/bin/pyinstaller docker-compose.spec" user
+    mv dist/docker-compose "${TARGET}"
+    "${TARGET}" version
+else
+    echo "Python-Pip is not installed in your Env"
```
hartwork: `>&2` and `exit 1` might be missing here. What do you think?

script/build/linux-entrypoint
```diff
@@ -5,11 +5,15 @@ set -ex
...
+fi
```
hartwork: Looks like a trailing newline is missing here.

script/build/linux-entrypoint
mmphego: Thanks for the heads up - I have resolved this issue!

script/build/linux-entrypoint
mmphego: Added trailing newline.

script/release/setup-venv.sh
```diff
@@ -1,40 +1,39 @@
...
-$PYTHONBIN -m venv $VENV_DIR $VENV_FLAGS
+$PYTHONBIN -m venv "${VENV_DIR}" "${VENV_FLAGS}"
 
-VENV_PYTHONBIN=$VENV_DIR/bin/python
+VENV_PYTHONBIN="${VENV_DIR}"/bin/python
 
 if debian_based; then
-  curl https://bootstrap.pypa.io/get-pip.py -o $VENV_DIR/get-pip.py
-  $VENV_PYTHONBIN $VENV_DIR/get-pip.py
+  curl https://bootstrap.pypa.io/get-pip.py | ${VENV_PYTHONBIN}
```
mmphego: Refactored code by piping curl into Python.

PR #6390, opened 2018-12-01
Fix merge on networks section

### LINKED ISSUE #6369: Cannot override network settings.
## Description of the issue

Configuration merge doesn't seem to merge network subsections correctly.
Probably related to #2851.

```
docker-compose version 1.23.1, build unknown
docker-py version: 3.5.1
CPython version: 3.7.1
OpenSSL version: OpenSSL 1.1.1  11 Sep 2018
```

```
Client:
 Version:           18.09.0-ce
 API version:       1.39
 Go version:        go1.11.2
 Git commit:        4d60db472b
 Built:             Fri Nov  9 00:05:34 2018
 Experimental:      false

Server:
 Engine:
  Version:          18.09.0-ce
  Go version:       go1.11.2
  Git commit:       4d60db472b
  Built:            Fri Nov  9 00:05:11 2018
  OS/Arch:          linux/amd64
  Experimental:     false

```

1. Set an `alias` on a service network.
2. Add an `ip_address` on the override file.

### Observed result

The `alias` is removed

### Expected result

Both the `alias` and the `ip_address` are set.

### Example

```
// docker-compose.yml
version: '3.7'
services:
  foo:
    image: scratch
    networks:
      default:
        aliases:
          - foo.bar
```

```
// docker-compose.override.yml
version: '3.7'
services:
  foo:
    networks:
      default:
        ipv4_address: 123.234.123.234
```

```
$: docker-compose config
services:
  foo:
    image: scratch
    networks:
      default:
        ipv4_address: 123.234.123.234
version: '3.7'
```

## Additional information

OS: Manjaro Linux (kernel 4.18.19-1-MANJARO)
`docker-compose` install method: distro package manager

ulyssessouza:
Properly merges the `networks` section

Resolves #6369

### DIFF
--- a/compose/config/config.py
+++ b/compose/config/config.py
@@ -1040,7 +1040,6 @@ def merge_service_dicts(base, override, version):
     md.merge_mapping('environment', parse_environment)
     md.merge_mapping('labels', parse_labels)
     md.merge_mapping('ulimits', parse_flat_dict)
-    md.merge_mapping('networks', parse_networks)
     md.merge_mapping('sysctls', parse_sysctls)
     md.merge_mapping('depends_on', parse_depends_on)
     md.merge_mapping('storage_opt', parse_flat_dict)
@@ -1050,6 +1049,7 @@ def merge_service_dicts(base, override, version):
     md.merge_sequence('security_opt', types.SecurityOpt.parse)
     md.merge_mapping('extra_hosts', parse_extra_hosts)
 
+    md.merge_field('networks', merge_networks, default={})
     for field in ['volumes', 'devices']:
         md.merge_field(field, merge_path_mappings)
 
@@ -1152,6 +1152,22 @@ def merge_deploy(base, override):
         md['placement'] = dict(placement_md)
 
     return dict(md)
+
+
+def merge_networks(base, override):
+    merged_networks = {}
+    all_network_names = set(base) | set(override)
+    base = {k: {} for k in base} if isinstance(base, list) else base
+    override = {k: {} for k in override} if isinstance(override, list) else override
+    for network_name in all_network_names:
+        md = MergeDict(base.get(network_name, {}), override.get(network_name, {}))
+        md.merge_field('aliases', merge_unique_items_lists, [])
+        md.merge_field('link_local_ips', merge_unique_items_lists, [])
+        md.merge_scalar('priority')
+        md.merge_scalar('ipv4_address')
+        md.merge_scalar('ipv6_address')
+        merged_networks[network_name] = dict(md)
+    return merged_networks
 
 
 def merge_reservations(base, override):
--- a/tests/unit/config/config_test.py
+++ b/tests/unit/config/config_test.py
@@ -1085,8 +1085,43 @@ class ConfigTest(unittest.TestCase):
         details = config.ConfigDetails('.', [base_file, override_file])
         web_service = config.load(details).services[0]
         assert web_service['networks'] == {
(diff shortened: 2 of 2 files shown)


compose/config/config.py
```diff
@@ -1154,6 +1154,20 @@ def merge_deploy(base, override):
...
 
 
+def merge_networks(base, override):
+    merged_networks = {}
+    all_network_names = set(base) | set(override)
+    base = {k: {} for k in base} if isinstance(base, list) else base
+    override = {k: {} for k in override} if isinstance(override, list) else override
+    for network_name in all_network_names:
+        md = MergeDict(base.get(network_name, {}), override.get(network_name, {}))
+        md.merge_scalar('aliases')
```
shin-: This needs to be `md.merge_field('aliases', merge_unique_items_lists, [])` so the result of `aliases; ['a', 'b']` and `aliases: ['b', 'c']` is `aliases: ['a', 'b', 'c']` to be consistent with other, similar config lists.

compose/config/config.py
```diff
@@ -1154,6 +1154,20 @@ def merge_deploy(base, override):
...
+        md.merge_scalar('ipv4_address')
+        md.merge_scalar('ipv6_address')
+        merged_networks[network_name] = dict(md)
```
shin-: Missing [`linked_local_ips`](https://docs.docker.com/compose/compose-file/compose-file-v2/#link_local_ips) (merge strategy same as aliases) and [`priority`](https://docs.docker.com/compose/compose-file/compose-file-v2/#priority) (simple `merge_scalar` as it's just an int value)

tests/unit/config/config_test.py
```diff
@@ -3843,8 +3878,71 @@ def test_no_base(self):
...
+            {self.config_name: self.override_config},
+            DEFAULT_VERSION)
+        assert service_dict[self.config_name] == {
+            'default': {
+                'aliases': ['foo.bar', 'foo.baz'],
+                'ipv4_address': '123.234.123.234'
+            }
+        }
+
+    def test_all_properties(self):
```
shin-: It would be good to add `link_local_ips` and `priority` values in this test as well.

tests/unit/config/config_test.py
```diff
@@ -3843,8 +3878,71 @@ def test_no_base(self):
 
 class MergeNetworksTest(unittest.TestCase, MergeListsTest):
```
shin-: A test to check that lists (aliases and link_local_ips) get combined rather than overridden would be good to have as well.

compose/config/config.py
ulyssessouza: Cool! 
Actually I understood that the field would be overwritten instead of merged in this case. My bad, I should have asked before the implementation and not assumed that :/

compose/config/config.py
ulyssessouza: Adding

tests/unit/config/config_test.py
ulyssessouza: Adding

tests/unit/config/config_test.py
ulyssessouza: Adding

shin-:
You've got a couple test failures because the order of list elements in your assert is mismatched. I believe they get ordered lexically during the merge, so you'll just need to change `{'aliases': ['foo', 'bar']}` to `{'aliases': ['bar', 'foo']}`

ulyssessouza:
@shin- Yep! Just fixed that

Issue #2369, opened 2015-11-11
docker-compose ps enhancements


dnephin:
This issue is to aggregate all the requests for enhancements to `ps`.  
- include the swarm node name (#2366)
- some way to disable wrapping of lines, either with a flag, or default to no wrapping when the terminal is not a tty (#1414)
- machine readable output, as json (#1992)
- container ip address (#1038)
- include the short container id (#1414)
- support additional fields (#1414)

We need to find a way to support all of this without making the default unnecessarily verbose.  I think we should be keeping the default fairly close to what we have now.

One option is to add a bunch of flags to include new columns.

lmesz:
Hello,

Thanks to gather the PRs and issues around PS.
I think the the default should be as it is, but give an oportunity the user to customize the columns would be great. E.g.: docker ps:
         Name                        Command               State                          Ports                         
-----------------------------------------------------------------------------------------------------------------------
dockermaster_etcd_1       /etcd                            Up      0.0.0.0:2379->2379/tcp, 2380/tcp, 4001/tcp, 7001/tcp
dockermaster_etcd_2       /etcd                            Up      0.0.0.0:2380->2379/tcp, 2380/tcp, 4001/tcp, 7001/tcp

docker ps --columns name
dockermaster_etcd_1
dockermaster_etcd_2

And docker ps --help should show the available columns.
The "machine readable output, as json" would be also great. With --json should drop the ps output like a simple json array as you suggested the original PR.

achekulaev:
I wanted to bring up another notice. Currently `ps` returns container **name** as a first column
```
# docker-compose ps
     Name                    Command                State     Ports 
-------------------------------------------------------------------
fusion114_cli_1   /opt/startup.sh /bin/sh -c ...   Exit 137         
fusion114_db_1    /entrypoint.sh mysqld            Exit 128         
fusion114_web_1   /opt/startup.sh apache2 -D ...   Exit 0   
```

However that value is useless for any other docker-compose command as docker-compose can not work with container names but only works with services. 

```
# docker-compose ps fusion114_cli_1
No such service: fusion114_cli_1
```
```
# docker-compose ps cli
     Name                    Command                State     Ports 
-------------------------------------------------------------------
fusion114_cli_1   /opt/startup.sh /bin/sh -c ...   Exit 137   
```

It makes sense to drop Name column and introduce Service Name or Service column instead.

```
# docker-compose ps
Service                    Command                State     Ports 
-------------------------------------------------------------------
[...]

PR #6391, opened 2018-12-01
Add ps --columns option

koooge:
Hi there.  This is a proposal of kind/feature `docker-compose ps --columns <value>`. (It's like `docker ps --format <value>`. )
But docker-compose is written in python, so I feel `ps --format` with go template is not for docker-compose .

Currently, if you need just container names, you have to use both docker and docker-compose .
```
$ docker ps --filter "id=$(docker-compose ps -q redis)" --format "{{.Names}}"
```

this proposal makes you only using docker-compose
```
$ docker-compose --columns "Name" redis
```

refs: https://github.com/docker/compose/issues/2369


glours:

PR #6397, opened 2018-12-05
trim whitespaces arround '=' in env settings

hirochachacha:
This resolves a common mistake.


hirochachacha:
Thank you for taking a look. How about #6403 ?

PR #6403, opened 2018-12-07
reject environment variable that contains white spaces

hirochachacha:


compose/config/environment.py
```diff
@@ -10,27 +10,32 @@
...
     if isinstance(env, six.binary_type):
         env = env.decode('utf-8', 'replace')
+    key, value = None, None
     if '=' in env:
-        return env.split('=', 1)
+        key, value = env.split('=', 1)
     else:
-        return env, None
+        key = env
+    if ' ' in key:
```
shin-: `if string.whitespace in  key:` would be better and allow us to also catch tabs and the like.

compose/config/environment.py
```diff
@@ -10,27 +10,32 @@
...
 from .errors import ConfigurationError
+from .errors import EnvFileNotFound
 
 log = logging.getLogger(__name__)
 
 
 def split_env(env):
     if isinstance(env, six.binary_type):
         env = env.decode('utf-8', 'replace')
+    key, value = None, None
```
shin-:
```suggestion
    key = value = None
```

compose/config/environment.py
```diff
@@ -10,27 +10,32 @@
...
     if '=' in env:
-        return env.split('=', 1)
+        key, value = env.split('=', 1)
     else:
-        return env, None
+        key = env
+    if ' ' in key:
+        raise ConfigurationError("environment variable '%s' should not contains white spaces." % key)
```
shin-: Grammar and style nits:
```suggestion
        raise ConfigurationError("environment variable name '%s' may not contain whitespace characters." % key)
```

compose/config/environment.py
```diff
@@ -10,27 +10,32 @@
...
     """
     if not os.path.exists(filename):
-        raise ConfigurationError("Couldn't find env file: %s" % filename)
+        raise EnvFileNotFound("Couldn't find env file: %s" % filename)
     elif not os.path.isfile(filename):
-        raise ConfigurationError("%s is not a file." % (filename))
+        raise EnvFileNotFound("%s is not a file." % (filename))
     env = {}
     with contextlib.closing(codecs.open(filename, 'r', 'utf-8-sig')) as fileobj:
         for line in fileobj:
```
shin-: We should catch the `ConfigurationError` to provide more information about the file where the error occurred, so people can find the source of the issue more easily if they have several environment files.

```python
for line in fileobj:
  line = line.strip()
  if line and not line.startswith('#'):
    try:
      k, v = split_env(line)
      env[k] = v
    except ConfigurationError as e:
      raise ConfigurationError('In file {}: {}'.format(filename, e.msg)
```

compose/config/environment.py
hirochachacha: done

compose/config/environment.py
hirochachacha: done.

compose/config/environment.py
hirochachacha: done.

compose/config/environment.py
hirochachacha: `string.whitespace` is string, so I did something like:
```
for k in key:
  if k in string.whitespace:
```

shin-:
Thank you!

PR #6406, opened 2018-12-08
Error on duplicate mount points.

### LINKED ISSUE #5948: docker-compose should warn on duplicate mount points
## Description of the issue

The following configuration produces an error from docker-compose:

```yaml
version: '3'
services:
  webserver:
    image: nginx
    volumes:
    - /tmp/foo:/tmp/foo
    - /tmp/foo:/tmp/foo
```

```
services.webserver.volumes value ['/tmp/foo:/tmp/foo', '/tmp/foo:/tmp/foo'] has non-unique elements
```

However, the following variants do not:

```yaml
version: '3'
services:
  webserver:
    image: nginx
    volumes:
    - /tmp/foo:/tmp/foo
    - /tmp/foo:/tmp/foo:rw
```

```yaml
version: '3'
services:
  webserver:
    image: nginx
    volumes:
    - /x:/y
    - /z:/y
```

Whilst it's easy to spot in simple cases like this, we've been tripped up by this in large configurations, where we assume `docker-compose config` would error on invalid configurations.

```
docker-compose version 1.21.0, build 5920eb0
```

collin5:
**i.e** 
screen shot 2018-12-09 at 00 52 44

**For**
```yaml
version: '3'
services:
  webserver:
    image: nginx
    volumes:
    - /tmp/foo:/tmp/foo
    - /tmp/foo:/tmp/foo:rw

```

Resolves #5948

### DIFF
--- a/compose/config/config.py
+++ b/compose/config/config.py
@@ -8,6 +8,7 @@ import os
 import string
 import sys
 from collections import namedtuple
+from operator import itemgetter, attrgetter
 
 import six
 import yaml
@@ -835,6 +836,17 @@ def finalize_service_volumes(service_dict, environment):
                 finalized_volumes.append(MountSpec.parse(v, normalize, win_host))
             else:
                 finalized_volumes.append(VolumeSpec.parse(v, normalize, win_host))
+
+        duplicate_mounts = []
+        mounts = [v.as_volume_spec() if isinstance(v, MountSpec) else v for v in finalized_volumes]
+        for mount in mounts:
+            if list(map(attrgetter('internal'), mounts)).count(mount.internal) > 1:
+                duplicate_mounts.append(mount.repr())
+
+        if duplicate_mounts:
+            raise ConfigurationError("Duplicate mount points: volumes [%s]" % (
+                ', '.join(duplicate_mounts)))
+        
         service_dict['volumes'] = finalized_volumes
 
     return service_dict
--- a/compose/config/config.py
+++ b/compose/config/config.py
@@ -8,7 +8,7 @@ import os
 import string
 import sys
 from collections import namedtuple
-from operator import itemgetter, attrgetter
+from operator import attrgetter
 
 import six
 import yaml
@@ -844,9 +844,9 @@ def finalize_service_volumes(service_dict, environment):
                 duplicate_mounts.append(mount.repr())
 
         if duplicate_mounts:
-            raise ConfigurationError("Duplicate mount points: volumes [%s]" % (
+            raise ConfigurationError("Duplicate mount points: [%s]" % (
                 ', '.join(duplicate_mounts)))
-        
+
         service_dict['volumes'] = finalized_volumes
 
     return service_dict
--- a/compose/service.py
+++ b/compose/service.py
@@ -9,6 +9,7 @@ import sys
 from collections import namedtuple
 from collections import OrderedDict
 from operator import attrgetter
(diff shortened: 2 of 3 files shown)


shin-:
Thank you for looking into this! There are 2 things I would want to see change design-wise before we go into implementation details:
1. This should definitely be an error, rather than a warning, and we should prevent deployment if we encounter this.
2. I think we should detect this earlier, during config validation. `finalize_service_volumes` in `compose/config/config.py` would be a good place to do this IMO.

collin5:
I've updated this. You can now take another look. Thank you!

compose/config/config.py
```diff
@@ -835,6 +836,21 @@ def finalize_service_volumes(service_dict, environment):
...
+                                                     else _v for _v in finalized_volumes]]
+        for external, internal in mounts:
+            if external and list(map(itemgetter(0), mounts)).count(
+                    external) > 1 or list(map(itemgetter(1), mounts)).count(internal) > 1:
+                duplicate_mounts.append("%s:%s" % (external, internal))
+
+        if duplicate_mounts:
+            log.error("Duplicate mount points: [%s]" % (
+                ', '.join(duplicate_mounts)))
+            sys.exit(1)
```
shin-: You can remove the `log` and `sys.exit` calls and `raise ConfigurationError` instead,

compose/config/config.py
```diff
@@ -835,6 +836,21 @@ def finalize_service_volumes(service_dict, environment):
...
                 finalized_volumes.append(VolumeSpec.parse(v, normalize, win_host))
+
+        duplicate_mounts = []
+        mounts = [(v.external, v.internal) for v in [_v.as_volume_spec()
+                                                     if hasattr(_v, 'as_volume_spec')
+                                                     else _v for _v in finalized_volumes]]
+        for external, internal in mounts:
+            if external and list(map(itemgetter(0), mounts)).count(
+                    external) > 1 or list(map(itemgetter(1), mounts)).count(internal) > 1:
+                duplicate_mounts.append("%s:%s" % (external, internal))
```
shin-: We're only concerned about `v.internal` duplicates - Mounting the same folder or volume in several (distinct) places inside the container is completely valid. As such, I would rewrite it thus:
```python
duplicate_mounts = []
mounts = [v.as_volume_spec() if isinstance(v, MountSpec) else v for v in finalized_volumes]
for m in mounts:
  if list(map(attrgetter('internal', mounts)).count(m.internal) > 1:
    duplicate_mounts.append(m.repr())
```

tests/acceptance/cli_test.py
```diff
@@ -510,7 +515,7 @@ def test_config_v3(self):
                     }, {
                         'target': '/anonymous', 'type': 'volume'
                     }, {
-                        'source': 'foobar',
+                        'source': 'foobar2',
```
shin-: This can be reverted - as I pointed out above, it's valid for the same volume to be mounted in multiple places.

tests/fixtures/v3-full/docker-compose.yml
```diff
@@ -48,7 +48,7 @@ services:
       - type: volume
         target: /anonymous
       - type: volume
-        source: foobar
+        source: foobar2
```
shin-: This can be reverted as well.

collin5:
No worries! I've updated this. Thank you.

shin-:
LGTM, thanks!

Issue #4431, opened 2017-02-05
Update requests dependency


jgeerds:
A few days ago *requests* 2.13 has been released. Unfortunately, *docker-compose* still enforces *requests* `< 2.12`:

`'requests >= 2.6.1, != 2.11.0, < 2.12',`

Is there a reason for this? Changes are [mostly bugfixes](https://github.com/kennethreitz/requests/blob/master/HISTORY.rst).

It seems like some [distributions already ship](https://git.archlinux.org/svntogit/community.git/tree/trunk/PKGBUILD?h=packages/docker-compose) `docker-compose` with newer versions of `requests` without any issues.

shin-:
There are indeed reasons for this: 
- https://github.com/docker/docker-py/issues/1321
- https://github.com/docker/docker-py/issues/1155
- https://github.com/docker/docker-py/issues/811
- https://github.com/docker/docker-py/issues/148

We make a fairly unconventional use of `requests` (HTTP over UNIX sockets) which makes us particularly sensitive to updates of the requests package - even when they're just bugfixes.

neilsh:
Suggestion: I wonder if a sustainable solution going forward would be to submit unit tests to the `requests` project covering the unconventional functionality that docker-compose relies on, to help ensure their interface doesn't change with new versions.

hrabalvojta:
Isn't this reason why you have RC versions?

voidlily:
I tried changing the requirement line for requests in docker-compose's setup.py to match docker-py's as an initial pass and tested some very basic functionality for local dev. I'm not sure what would need to be done next to take this to completion (though I suspect unit tests submitted upstream as from a previous suggestion would be helpful to clarify the unconventional behavior docker-compose relies on).

My motivation for making this change on my end is that in a project I'm working on there's another package I'm using with a transitive dependency on a recent version of requests, causing a conflict with the most recent official release of docker-compose.

`requests>2.18.0`
 
`'requests >= 2.6.1, != 2.11.0, < 2.12',`

PR #6415, opened 2018-12-15
setup.py: allow all recent 2.x requests releases

jacmet:
Instead of having to update this for each new requests release.

It it not quite clear why the restriction was added in the first place in
commit b0480b4d04e (Bump SDK version to latest), but change it to simply
disallow the upcoming 3.0 release to match what is done for the other
modules.

Signed-off-by: Peter Korsgaard <peter@korsgaard.com>

Resolves #

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -33,7 +33,7 @@ install_requires = [
     'cached-property >= 1.2.0, < 2',
     'docopt >= 0.6.1, < 0.7',
     'PyYAML >= 3.10, < 4',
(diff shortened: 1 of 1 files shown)


shin-:
Hi @jacmet,

Thanks for the contribution. [As explained before](https://github.com/docker/compose/issues/4431#issuecomment-277846958), this is a deliberate decision on our part after a few occurrences of requests updates accidentally breaking for our usage. I'm satisfied with that strategy and not looking to reconsider it at the moment.

PR #6454, opened 2019-01-11
Resolve digests without pulling image

### LINKED ISSUE #5818: Add ability to resolve image digests without pulling entire images
PR #4333 added `--resolve-image-digests` to `docker-compose config`, which resolves image tags to their digest to form stable references. However, this relies on the images being available on the Docker host.

It would be useful if Docker Compose could fetch image digests without having to pull the image if the image doesn't exist locally. In CI jobs using Docker-in-Docker, pulling images just for tag digests can be a significant waste of time and network resources.

Digests can be obtained using `HEAD` requests against `/v2/<name>/manifests/<reference>` in the [v2 API](https://docs.docker.com/registry/spec/api/#manifest).

https://stackoverflow.com/questions/35186693/what-is-the-docker-registry-v2-api-endpoint-to-get-the-digest-for-an-image#35420411

rumpl:
If there is no image locally `docker-compose --resolve-image-digests`
will try and get the digest from the repository.

Signed-off-by: Djordje Lukic <djordje.lukic@docker.com>

Resolves #5818


compose/bundle.py
```diff
@@ -118,6 +109,33 @@ def get_image_digest(service, allow_push=False):
...
+                digest = '{image_name}@{digest}'.format(
+                    image_name=service.image_name,
+                    digest=distribution['Descriptor']['digest']
+                )
+        except NoSuchImageError:
+            action = 'build' if 'build' in service.options else 'pull'
+            raise UserError(
+                "Digest not found for service '{service}'. "
+                "Repository does not exist or may require 'docker login'"
+                .format(service=service.name, action=action))
```
shin-: action isn't used here and can be removed

compose/service.py
```diff
@@ -363,6 +363,12 @@ def ensure_image_exists(self, do_build=BuildAction.none, silent=False):
             "rebuild this image you must use `docker-compose build` or "
             "`docker-compose up --build`.".format(self.name))
 
+    def distribution(self):
```
shin-: I would go for a more explicit name here, like `get_image_registry_data()`

PR #6327, opened 2018-11-06
Add option for `--all` flag to `ps`

collin5:
Resolves #6271

PR #6461, opened 2019-01-15
Add bash completion for `ps --all|-a`

albers:
This adds bash completion for #6327.
Please include into v1.24.0 for feature parity.


albers:
@chris-crone Please pick into v1.24.0 because the corresponding feature will be present in this release.

chris-crone:
Added it to the milestone so that we cherry-pick it. We just have one outstanding issue before we do the next release.

Thanks for the contribution @albers!

PR #6490, opened 2019-01-30
Fix Flake8 lint

ulyssessouza:
This removes extra indentation and replace the use of `is` by `==` when
comparing strings

Signed-off-by: Ulysses Souza <ulysses.souza@docker.com>


ulyssessouza:
Dropping it in favor of https://github.com/docker/compose/pull/6501 that includes this commit

PR #6518, opened 2019-02-14
Bump python 3.6.8 -> 3.7.2

### LINKED ISSUE #6502: Python 3.6 EOL
Python 3.6.8 [will be the last release of Python 3.6](https://www.python.org/downloads/release/python-368/). As such, we should update to use Python 3.7 for builds.

ulyssessouza:
Resolves #6502 

Note that this PR tries to enforce the python version in the `Dockerfile` as `3.7.2-stretch` (same as `3.7` by now) to have a more predictable build.

Note also that the versions of `tox` and `virtualenv` get harmonized through the various configurations.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -13,7 +13,7 @@ jobs:
         command: sudo pip install --upgrade tox==2.1.1 virtualenv==16.2.0
     - run:
         name: unit tests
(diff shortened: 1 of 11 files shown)


chris-crone:
Can you please repush this to a branch on `docker/compose` so that the macOS and Windows CIs are run?

PR #6544, opened 2019-03-04
Add test and implementation for secret added after container has been…

### LINKED ISSUE #6543: Secrets not added to containers that have already been created
## Description of the issue

If I create a simple docker compose file without a secret attached to any container, run `docker-compose up`, then add the secret, and run `docker-compose up` again, then the secret is not added.

I ran this with the version of docker-compose from the master branch at commit [b09d8802](https://github.com/docker/compose/commit/b09d8802edb498a91a73bcd63603774c9b027b47)

```
```

```
Client:
 Version:           18.06.1-ce
 API version:       1.38
 Go version:        go1.10.4
 Git commit:        e68fc7a
 Built:             Mon Oct  1 14:25:31 2018
 Experimental:      false

Server:
 Engine:
  Version:          18.06.1-ce
  API version:      1.38 (minimum version 1.12)
  Go version:       go1.10.4
  Git commit:       e68fc7a
  Built:            Mon Oct  1 14:25:33 2018
  OS/Arch:          linux/amd64
  Experimental:     false
```

```
secrets:
  A:
    file: /home/catears/code/compose/tests/rmme/docker-compose.yml
services:
  nginx:
    command: ls /run/secrets
    image: nginx
    secrets:
    - source: A
version: '3.5'
```

```yaml
version: '3.5'

services:
  nginx:
    image: nginx
    command: ls /run/secrets

    # Run once, uncomment me, and run again!
    # secrets:
    #   - A
secrets:
  A:
    file: ./docker-compose.yml
```

1. Using the above compose-file, run `docker-compose up`.
2. Wait until it fails (no secrets so `/run/secrets` doesn't exist).
3. Uncomment/Add the `A` secret.
4. Run `docker-compose up` again.

### Observed result

The following output:

```
Starting rmme_nginx_1 ... done
Attaching to rmme_nginx_1
nginx_1  | ls: cannot access '/run/secrets': No such file or directory
rmme_nginx_1 exited with code 2
```

### Expected result

The following output (which happens if you do not have a container beforehand):

```
Creating rmme_nginx_1 ... done
Attaching to rmme_nginx_1
nginx_1  | A
rmme_nginx_1 exited with code 0
```

I poked around the source code and it seems to me that when starting a project, compose will compare the configurations using the result of `Service.config_dict()` and adding the following diff, seemed to solve it for me.

```
diff --git a/compose/service.py b/compose/service.py
index 6483f4f3..fc51eaac 100644
--- a/compose/service.py
+++ b/compose/service.py
@@ -686,6 +686,7 @@ class Service(object):
             'links': self.get_link_names(),
             'net': self.network_mode.id,
             'networks': self.networks,
+            'secrets': self.secrets,
             'volumes_from': [
                 (v.source.name, v.mode)
                 for v in self.volumes_from if isinstance(v.source, Service)
```

I'd love to do a PR for this, but I do not know how to properly write a test for it. Also, I don't know if it is correct to recreate the container because the secret was missing, but I figured this might be the case because `volumes_from`, `image_id` and `networks` changes would cause compose to recreate a container.

CatEars:
… created

The issue is that if a secret is added to the compose file, then it will
not notice that containers have diverged since last run, because secrets
are not part of the config_hash, which determines if the configuration of
a service is the same or not.

Signed-off-by: Henrik Adolfsson <catears13@gmail.com>

Resolves #6543 

Not sure I got the test to work properly, need some help with that if possible...

### DIFF
--- a/tests/integration/project_test.py
+++ b/tests/integration/project_test.py
@@ -1534,7 +1534,6 @@ class ProjectTest(DockerClientTestCase):
             config_data=config_data1,
         )
         project.up()
-        project.stop()
 
         # Then up with secret
         project = Project.from_config(
--- a/tests/integration/project_test.py
+++ b/tests/integration/project_test.py
@@ -1,6 +1,7 @@
 from __future__ import absolute_import
 from __future__ import unicode_literals
 
+import copy
 import json
 import os
 import random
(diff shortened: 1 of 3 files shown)


tests/integration/project_test.py
```diff
@@ -1496,6 +1497,61 @@ def test_project_up_with_secrets(self):
...
+        config_data2 = build_config(**config_input2)
+
+        # First up with non-secret
+        project = Project.from_config(
+            client=self.client,
+            name='composetest',
+            config_data=config_data1,
+        )
+        project.up()
+        project.stop()
```
ijc: I think you don't want to `stop` here, since your goal is to test that an update to an existing service works.

CatEars: Removed it in the latest commit, tests pass on my computer, so should pass on Jenkins.

CatEars:
I rebased from upstream master since one of the test files had changed.

PR #6556, opened 2019-03-08
script/test/default: Pickup and use eventual HTTP proxy

### LINKED ISSUE #2590: Test will fail if using a proxy
I am trying to solve another issue, and when I try to run the integration tests some of the tests build an image. But there is an issue on the container because some of them are doing apt-get update they will fail because of timeout.

joakimr-axis:
Without passing proper proxy settings to the docker build command, it
will fail. This patch mitigates that (and works if no proxy is set).

Resolves #2590


joakimr-axis:
There should not be all 9 commits in one pull request. I will fix this.

PR #6559, opened 2019-03-08
Pickup and use eventual HTTP proxy

### LINKED ISSUE #2590: Test will fail if using a proxy

joakimr-axis:
Without passing proper proxy settings to the docker build command, it
will fail. This patch mitigates that (and works if no proxy is set).

Change-Id: I81e50b173047dd4a92b5a38a4940eff3453c8b80

Resolves #2590

### DIFF
--- a/script/build/image
+++ b/script/build/image
@@ -16,5 +16,7 @@ echo "${DOCKER_COMPOSE_GITSHA}" > compose/GITSHA
 python setup.py sdist bdist_wheel
 
 docker build \
+    --build-arg http_proxy=$http_proxy \
+    --build-arg https_proxy=$https_proxy \
     --build-arg GIT_COMMIT="${DOCKER_COMPOSE_GITSHA}" \
     -t "${TAG}" .
--- a/script/build/linux
+++ b/script/build/linux
@@ -8,6 +8,8 @@ DOCKER_COMPOSE_GITSHA="$(script/build/write-git-sha)"
 TAG="docker/compose:tmp-glibc-linux-binary-${DOCKER_COMPOSE_GITSHA}"
 
 docker build -t "${TAG}" . \
+       --build-arg http_proxy=$http_proxy \
+       --build-arg https_proxy=$https_proxy \
        --build-arg BUILD_PLATFORM=debian \
        --build-arg GIT_COMMIT="${DOCKER_COMPOSE_GITSHA}"
 TMP_CONTAINER=$(docker create "${TAG}")
--- a/script/build/test-image
+++ b/script/build/test-image
@@ -13,6 +13,8 @@ IMAGE="docker/compose-tests"
 DOCKER_COMPOSE_GITSHA="$(script/build/write-git-sha)"
 docker build -t "${IMAGE}:${TAG}" . \
        --target build \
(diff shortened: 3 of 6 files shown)


joakimr-axis:
How should I interpret this output from Jenkins?

```
10:33:49 b6da999fed9c: Retrying in 2 seconds
10:33:50 b6da999fed9c: Retrying in 1 second
10:33:59 efa807d0493e: Pushed
10:33:59 received unexpected HTTP status: 500 Internal Server Error
[Pipeline] }
[Pipeline] // stage
[Pipeline] }
[Pipeline] // wrap
[Pipeline] }
[Pipeline] // wrap
[Pipeline] }
[Pipeline] // withDockerRegistry
[Pipeline] }
[Pipeline] // node
[Pipeline] End of Pipeline

GitHub has been notified of this commit’s build result

ERROR: script returned exit code 1
Finished: FAILURE
```

Internal error, why? Something I can do about it, and if so, how?

ijc:
The 500 is an external issue, the appropriate people have been paged. 

Sorry for the inconvenience.

ijc:
Thanks for the PR.

I suppose you have confirmed that https://docs.docker.com/network/proxy/ doesn't work for `docker build`?

There's some inconsistency here around when `http_proxy`+`https_proxy` are set vs `no_proxy`. I would have expected all three to be always set en bloc. What am I missing?

Lastly, `--env="no_proxy=$no_proxy,docker"` can expand to `--env="no_proxy=,docker"`, is that really valid? (specifically a leading `,` on the value)

joakimr-axis:
You are right, `no_proxy` is expected together with the other ones. I will fix that.
And I think `--env="${no_proxy:+docker,$no_proxy}"` is what we want to mitigare the other question.

ijc:
Thanks, please also answer:
> I suppose you have confirmed that https://docs.docker.com/network/proxy/ doesn't work for docker build?

joakimr-axis:
So I have been trying to understand where you are going with that question, not really with success, so please help me out here.
If you set the user's docker client proxy as per the instructions in https://docs.docker.com/network/proxy/#configure-the-docker-client, what is currently latest on master will build without further alterations (on Docker >= 17.07, according to the documentation).
Without setting the user's docker client proxy as per the instructions in https://docs.docker.com/network/proxy/#configure-the-docker-client, it will not (but with the additions in this pull request, it will tough).
Please let me know in a binary yes/no way if you would like me to make the suggested no_proxy updates or if you intend to scrap this PR so I can drop it immediately instead.

joakimr-axis:
> Thanks, please also answer:
> 
> > I suppose you have confirmed that https://docs.docker.com/network/proxy/ doesn't work for docker build?

It's been a while since that was written, so I don't really rembember. I do have this setting for my Docker client on the machine behind the proxy, and trying it out today it works with current master ([440c94ea7a7e62b3de50722120ca34c4e818205a](https://github.com/docker/compose/commit/440c94ea7a7e62b3de50722120ca34c4e818205a)) out of the box. I will thus ditch this PR.

PR #6561, opened 2019-03-08
Update Docker version to 18.09.3

joakimr-axis:
Change-Id: I1ac2a617f03443b9f7327dcddcde4af8275e32e0
Signed-off-by: Joakim Roubert <joakimr@axis.com>

Resolves #6560

PR #6566, opened 2019-03-08
Delete Dockerfile.armhf; Dockerfile now builds on armhf

### LINKED ISSUE #6167: Dockerfile.armhf should be updated

joakimr-axis:
The current Dockerfile builds for armhf so there is no longer need for
the separate Dockerfile.armhf.

Change-Id: I7a90c32cc53dd7c08404ca019f16ff6e31dc5c34

Resolves #6167


ijc:
Unfortunately I don't think they are precisely the same (ignoring the expected version bump) this is effectively switching from using https://download.docker.com/linux/static/stable/armhf/docker-17.12.0-ce.tgz as the source of the `docker` binary to using the `docker:18.06.1` image on hub.

However although the docker image on hub is multiplatform it doesn't seem to have armhf precisely:
```console
$ docker manifest inspect docker:18.06.1 | jq '.manifests[].platform | select (.architecture|contains("arm"))'
{
  "architecture": "arm",
  "os": "linux",
  "variant": "v6"
}
{
  "architecture": "arm64",
  "os": "linux",
  "variant": "v8"
}
```

`armhf` is variant=7, while the image only have v6 (probably because that is what the alpine bases support).

In that light I'm slightly reluctant to make the change without a better understanding of the implications.

A switch to pull the binary from "https://download.docker.com/linux/static/stable/armhf/docker-18.06.1-ce.tgz" would, I think, be ok though, or perhaps given #6561 (once CI is happy) we might as well flip to 18.09.3 for consistency (in fact you could just as well do that change in that PR too).

joakimr-axis:
> `armhf` is variant=7, while the image only have v6 (probably because that is what the alpine bases support).

Ok. I have built and run the result on a Raspberry Pi and on another armhf (armv7hf) device.

> A switch to pull the binary from "https://download.docker.com/linux/static/stable/armhf/docker-18.06.1-ce.tgz" would, I think, be ok though, or perhaps given #6561 (once CI is happy) we might as well flip to 18.09.3 for consistency (in fact you could just as well do that change in that PR too).

I will add this removal of the armhf file to that PR and then close this one!

There is something—to me—strange going in with "Internal error" in the Jenkins build that I do not know how to come to terms with; is it me or is something behind the scenes unhappy?

joakimr-axis:
As per Ian's suggestion, this patch was incorporated in [https://github.com/docker/compose/pull/6561](https://github.com/docker/compose/pull/6561). Hence, this PR will be closed.

PR #6572, opened 2019-03-11
run.sh: Performance improvements

### LINKED ISSUE #6571: Performance improvements for run.sh
run.sh contains some unnecessary statements that can be removed. For instance, a statement like

if [ -n "$foo" ]; then
   echo bar
fi

adds an extra check for the [ result that was already evaluated, and shell script best practice is then to write

[ -z "$foo" ] || echo bar

instead and save us from wasting some extra CPU cycles while decreasing the shell script footprint too.

Furthermore, there is no need to add quotes where not needed, and also not make external calls to external binaries like dirname (where the shell build-in string manipulation can do the job for us at a much lower cost).

I will add a pull request for this.

joakimr-axis:
Some shell script best practices applied; remove unnecessary extra if
statements, remove quoting where not needed, use shell internal
functionality instead of external call to dirname.

Change-Id: I71b64a82e7e907a85822bd147d7c116c87ffea09

Resolves #6571

### DIFF
--- a/script/run/run.sh
+++ b/script/run/run.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/sh -e
 #
 # Run docker-compose in a container
 #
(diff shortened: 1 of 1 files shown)


ulyssessouza:
https://github.com/docker/compose/issues/6571#issuecomment-471542853

PR #6574, opened 2019-03-11
Make test scripts POSIX compatible

### LINKED ISSUE #6573: Make test scripts POSIX compatible
There is no real need to use bash for the test scripts, when they could easily be made POSIX compatible.

joakimr-axis:
Change-Id: I69e412d86381f22287abb963295e2aab6c791ee4

Resolves #6573

### DIFF
--- a/script/test/all
+++ b/script/test/all
@@ -1,24 +1,24 @@
-#!/bin/bash
+#!/usr/bin/env sh
 # This should be run inside a container built from the Dockerfile
 # at the root of the repo - script/test will do it automatically.
 
 set -e
 
->&2 echo "Running lint checks"
+>&2 echo Running lint checks
 docker run --rm \
-  --tty \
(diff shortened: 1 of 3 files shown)


script/test/all
```diff
@@ -1,64 +1,60 @@
-#!/bin/bash
+#!/bin/sh -e
```
nvllsvm: Use `/usr/bin/env sh` instead. On macOS and many Linux distros, `/bin/sh` is bash. By using `env`, you give the user the opportunity to use a POSIX shell like dash by respecting their configured `$PATH`.

It also makes my life easier when I'm testing a script I want to be POSIX-compliant. :)

joakimr-axis: Will fix immediately, thank you for your input!

ulyssessouza:
Same as https://github.com/docker/compose/pull/6576#issuecomment-491804574

PR #6576, opened 2019-03-11
Make build scripts POSIX compatible

### LINKED ISSUE #6575: Make build scripts POSIX compatible
There is no real need to use bash for the build scripts, when they could easily be made POSIX compatible.

joakimr-axis:
Change-Id: Ib74ab851c869c76720f5d0460de2522a00212170

Resolves #6575

### DIFF
--- a/script/build/image
+++ b/script/build/image
@@ -1,20 +1,20 @@
-#!/bin/bash
+#!/usr/bin/env sh
 
 set -e
 
-if [ -z "$1" ]; then
-    >&2 echo "First argument must be image tag."
-    exit 1
-fi
+[ -n "$1" ] || {
+	>&2 echo "First argument must be image tag."
+	exit 1
+}
 
-TAG="$1"
+TAG=$1
 
-VERSION="$(python setup.py --version)"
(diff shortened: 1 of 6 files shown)


script/build/image
```diff
@@ -1,11 +1,9 @@
-#!/bin/bash
+#!/bin/sh -e
```
nvllsvm: Use `/usr/bin/env sh` instead. On macOS and many Linux distros, `/bin/sh` is bash. By using `env`, you give the user the opportunity to use a POSIX shell like dash by respecting their configured `$PATH`.

It also makes my life easier when I'm testing a script I want to be POSIX-compliant. :)

joakimr-axis: Thank you for your input, the patch has been updated.

ulyssessouza:
Hello @joakimr-axis! Thank you for the contribution, but unfortunately I have to decline it.
IMHO, your patch reduces the readability of the scripts without giving anything back. 

Please, before proposing a PR on a self proposed issue, wait for the confirmation of the need of the last one and also discussion about the implementation design.

Issue #3574, opened 2016-06-09
docker-compose up doesn't pull down latest image if the image exists locally


kag0:
It would be nice if there were an option to check for new versions of images when running [`docker-compose up`](https://docs.docker.com/compose/reference/up/). 

We already have this functionality with [`docker build --pull`](https://docs.docker.com/engine/reference/commandline/build/) as was discussed here https://github.com/docker/docker/issues/4238 and there is an open issue to bring `--pull` to `docker run` here https://github.com/docker/docker/issues/13331.

I propose adding `--pull` to `up` to always attempt to pull a newer version of the images in the compose file.

shin-:
Is there a reason why `docker-compose pull && docker-compose up` is impractical?

kag0:
I think most of the points for/against are the same as the ones discussed in the issue for adding `--pull` to `docker run`. They range from consistency, to ease of scripting/workflow, to swarm integration (out of curiosity, what does `docker-compose pull` do with swarm?). 
I don't think it's a major issue, but something to consider. The same type of users who desired the feature elsewhere would likely also enjoy it here.

decoomanj:
I'm trying to execute "docker-compose build", but it doesn't refresh the image referenced in the Dockerfile. "docker-compose pull" only refreshes the images inside the YML. When can we expect the "docker-compose build --pull" functionality (or similar)?

cressie176:
Instead of (or in addition to) adding --pull to the cli, what about adding something on the service definition in the docker-compose file?, e.g.

```
version: '2'

services:

  postgres:
    image: postgres
    container_name: postgres
    pull: true
    ports:
     - '5432:5432'
```

This way if there's a service I don't care about being latest and one I do, docker-compose won't waste time downloading images I'm not interested in

blockloop:
I came here looking for this feature because we use it in our production cluster of Kubernetes. There the tag is "imagePullPolicy" and it can be set to "IfNotPresent", "Always", or "Never." Something similar for a compose environment would be nice.

PengBAI:
In our case, we need to rebuild base image everyday to make sure the latest dependencies updated for applications. Docker compose up to pull the latest image with the same tag is a nice feature. Why not !

marcelovcpereira:
Any news about this?

sugun999:
Hi,

Any news about this issue?

jonathan-kosgei:
+1

ivytribe:
+1

shin-:
[...]

Issue #883, opened 2015-01-26
docker-compose build and docker build lead to different IDs


schmunk42:
When building containers for my project I noticed that, even if I build from the same Dockerfile, `fig` and `docker` build different image IDs in the end.

I also tried to use the same tag, but the images still were different.

Built with fig:

```
IMAGE               CREATED             CREATED BY                                      SIZE
c9d754ad4f35        5 minutes ago       /bin/sh -c #(nop) ADD dir:ce7a82aadebe53a9cff   16.92 MB
7601cc84a236        9 minutes ago       /bin/sh -c /usr/local/bin/composer install --   34.13 MB
712b4a974fed        10 minutes ago      /bin/sh -c #(nop) ADD file:bbb37f3a1cb51d4326   92.3 kB
87d612d4111e        10 minutes ago      /bin/sh -c #(nop) ADD file:afe6fb0fc00bfc90cb   2.734 kB
84d2d9e64dc3        10 minutes ago      /bin/sh -c chmod 0400 /root/.ssh/id_rsa-deplo   1.808 kB
16e392fd78c7        10 minutes ago      /bin/sh -c #(nop) ADD file:45af06b02553fb4264   81 B
48f0aa118a70        10 minutes ago      /bin/sh -c #(nop) ADD file:07f9f2dc4a310c8a29   129 B
ec26eedffc59        10 minutes ago      /bin/sh -c #(nop) ADD file:f140ceae508309fece   1.679 kB
--- same till here, layers above from Dockerfile in project ---
1f260e5c4f2f        11 days ago         /bin/sh -c sed -i.bak 's/user = www-data/user   44.63 kB
5ec1e62830a7        11 days ago         /bin/sh -c #(nop) ENV YII_DEBUG=true            0 B
e0a170b5adbc        11 days ago         /bin/sh -c #(nop) ENV YII_ENV=dev               0 B
[...]
```

Built with docker:

```
a25b0e21a23b        About a minute ago   /bin/sh -c #(nop) ADD dir:48b6469d85aa5dbac5d   16.92 MB
6683ef055846        35 minutes ago       /bin/sh -c /usr/local/bin/composer install --   34.13 MB
47a678036ce5        36 minutes ago       /bin/sh -c #(nop) ADD file:cd1a4d0926e8ab11d5   92.3 kB
3b6cd9a1b4e3        36 minutes ago       /bin/sh -c #(nop) ADD file:21c0f8485a3f156686   2.734 kB
[...]

PR #6584, opened 2019-03-15
feat: add --cli option

orisano:
i implemented --exec option on `docker-compose build` for to use buildkit.

This PR is prototype. please give me feedback.

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -264,7 +264,6 @@ class TopLevelCommand(object):
 
         Options:
             --build-arg key=val     Set build-time variables for services.
-            --cli                   Run docker build command
             --compress              Compress the build context using gzip.
             --force-rm              Always remove intermediate containers.
             -m, --memory MEM        Sets memory limit for the build container.
@@ -276,6 +275,9 @@ class TopLevelCommand(object):
         """
         service_names = options['SERVICE']
         build_args = options.get('--build-arg', None)
+        environment_file = options.get('--env-file')
+        environment = Environment.from_env_file(self.project_dir, environment_file)
+        native_builder = environment.get_boolean('COMPOSE_NATIVE_BUILDER')
         if build_args:
             if not service_names and docker.utils.version_lt(self.project.client.api_version, '1.25'):
                 raise UserError(
@@ -295,7 +297,7 @@ class TopLevelCommand(object):
             gzip=options.get('--compress', False),
             parallel_build=options.get('--parallel', False),
             silent=options.get('--quiet', False),
-            cli=bool(options.get('--cli', False)),
+            cli=native_builder,
         )
 
     def bundle(self, options):
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -264,8 +264,8 @@ class TopLevelCommand(object):
 
         Options:
             --build-arg key=val     Set build-time variables for services.
+            --cli                   Run docker build command
             --compress              Compress the build context using gzip.
-            --exec                  Run docker build command
             --force-rm              Always remove intermediate containers.
             -m, --memory MEM        Sets memory limit for the build container.
             --no-cache              Do not use cache when building the image.
@@ -295,7 +295,7 @@ class TopLevelCommand(object):
             gzip=options.get('--compress', False),
             parallel_build=options.get('--parallel', False),
             silent=options.get('--quiet', False),
-            _exec=bool(options.get('--exec', False)),
+            cli=bool(options.get('--cli', False)),
         )
 
     def bundle(self, options):
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -263,16 +263,16 @@ class TopLevelCommand(object):
         Usage: build [options] [--build-arg key=val...] [SERVICE...]
 
         Options:
+            --build-arg key=val     Set build-time variables for services.
             --compress              Compress the build context using gzip.
+            --exec                  Run docker build command
             --force-rm              Always remove intermediate containers.
+            -m, --memory MEM        Sets memory limit for the build container.
             --no-cache              Do not use cache when building the image.
             --no-rm                 Do not remove intermediate containers after a successful build.
-            --pull                  Always attempt to pull a newer version of the image.
-            -m, --memory MEM        Sets memory limit for the build container.
-            --build-arg key=val     Set build-time variables for services.
             --parallel              Build images in parallel.
+            --pull                  Always attempt to pull a newer version of the image.
             -q, --quiet             Don't print anything to STDOUT
-            --exec                  Run docker build command
         """
         service_names = options['SERVICE']
         build_args = options.get('--build-arg', None)
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -272,6 +272,7 @@ class TopLevelCommand(object):
             --build-arg key=val     Set build-time variables for services.
             --parallel              Build images in parallel.
             -q, --quiet             Don't print anything to STDOUT
+            --exec                  Run docker build command
         """
         service_names = options['SERVICE']
         build_args = options.get('--build-arg', None)
@@ -293,7 +294,8 @@ class TopLevelCommand(object):
             build_args=build_args,
             gzip=options.get('--compress', False),
             parallel_build=options.get('--parallel', False),
-            silent=options.get('--quiet', False)
+            silent=options.get('--quiet', False),
+            _exec=bool(options.get('--exec', False)),
         )
 
     def bundle(self, options):
--- a/compose/service.py
+++ b/compose/service.py
@@ -8,6 +8,7 @@ import os
 import re
 import subprocess
 import sys
+import tempfile
 from collections import namedtuple
 from collections import OrderedDict
 from operator import attrgetter
@@ -1793,15 +1794,28 @@ def exec_build(path=None, tag=None, quiet=False, fileobj=None,
     command_builder.add_arg("--platform", platform)
     command_builder.add_arg("--isolation", isolation)
     command_builder.add_arg("--progress", "plain")
+
+    iidfile = tempfile.mktemp()
+    command_builder.add_arg("--iidfile", iidfile)
     args = command_builder.build([path])
 
+    magic_word = "Successfully built "
+    appear = False
     with subprocess.Popen(args, stdout=subprocess.PIPE, universal_newlines=True) as p:
         while True:
             line = p.stdout.readline()
             if line == "":
                 break
-            line = line.replace("writing image sha256:", "Successfully built ")
+            if magic_word in line:
+                appear = True
             yield json.dumps({"stream": line})
+
+    with open(iidfile) as f:
+        image_id = f.readline().split(":")[1].strip()
+    os.remove(iidfile)
+
+    if not appear:
+        yield json.dumps({"stream": "{}{}\n".format(magic_word, image_id)})
 
 
 class _CommandBuilder(object):
--- a/compose/service.py
+++ b/compose/service.py
@@ -1780,7 +1780,6 @@ def cli_build(path, tag=None, quiet=False, fileobj=None,
     command_builder.add_arg("--network", network_mode)
     command_builder.add_flag("--no-cache", nocache)
     command_builder.add_arg("--platform", platform)
-    command_builder.add_arg("--progress", "plain")
     command_builder.add_flag("--pull", pull)
     command_builder.add_flag("--quiet", quiet)
     command_builder.add_flag("--rm", rm)
--- a/compose/service.py
+++ b/compose/service.py
@@ -1717,23 +1717,9 @@ def exec_build(path, tag=None, quiet=False, fileobj=None,
     """
     Args:
         path (str): Path to the directory containing the Dockerfile
-        fileobj: A file object to use as the Dockerfile. (Or a file-like
-            object)
-        tag (str): A tag to add to the final image
-        quiet (bool): Whether to return the status
-        nocache (bool): Don't use the cache when set to ``True``
-        rm (bool): Remove intermediate containers. The ``docker build``
-            command now defaults to ``--rm=true``, but we have kept the old
-            default of `False` to preserve backward compatibility
-        timeout (int): HTTP timeout
-        custom_context (bool): Optional if using ``fileobj``
-        encoding (str): The encoding for a stream. Set to ``gzip`` for
-            compressing
-        pull (bool): Downloads any updates to the FROM image in Dockerfiles
-        forcerm (bool): Always remove intermediate containers, even after
-            unsuccessful builds
-        dockerfile (str): path within the build context to the Dockerfile
         buildargs (dict): A dictionary of build arguments
+        cache_from (:py:class:`list`): A list of images used for build
+            cache resolution
         container_limits (dict): A dictionary of limits applied to each
             container created by the build process. Valid keys:
             - memory (int): set memory limit for build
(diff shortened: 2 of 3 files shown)


AkihiroSuda:
cc @tonistiigi @tiborvass

compose/service.py
```diff
@@ -1695,3 +1698,127 @@ def rewrite_build_path(path):
...
+    command_builder.add_arg("--isolation", isolation)
+    command_builder.add_arg("--progress", "plain")
+    args = command_builder.build([path])
+
+    with subprocess.Popen(args, stdout=subprocess.PIPE, universal_newlines=True) as p:
+        while True:
+            line = p.stdout.readline()
+            if line == "":
+                break
+            line = line.replace("writing image sha256:", "Successfully built ")
```
AkihiroSuda: why?

orisano: for bypass this.
https://github.com/docker/compose/blob/master/compose/service.py#L1110
```python
        for event in all_events:
            if 'stream' in event:
                match = re.search(r'Successfully built ([0-9a-f]+)', event.get('stream', ''))
                if match:
                    image_id = match.group(1)
```

AkihiroSuda: This doesn't seem robust.
Can we just append `Successfully built...` when it didn't appear from `docker build` CLI?

orisano: I do not know how to get image_id from stdout on 'docker build --progress=plain'.
Do you have any idea?

AkihiroSuda: `--iidfile`?

orisano: I tried to implement to use an iidfile method.

docteurklein:
cool! It's not the first time using the CLI is an option (see https://docs.docker.com/compose/reference/envvars/#compose_interactive_no_cli ). maybe there is a way to normalize this?

mipearson:
I'm keen to contribute here - is there any guidance from the compose team as to how they'd prefer this to be implemented?

tonistiigi:
Not a compose maintainer but this seems ok approach to me. We'll probably take a bit different stab at the same problem at https://github.com/tonistiigi/buildx/issues/13 but the solutions don't seem conflicting and have bit different use cases.

compose/cli/main.py
```diff
@@ -267,6 +267,7 @@ def build(self, options):
             --build-arg key=val     Set build-time variables for services.
             --parallel              Build images in parallel.
             -q, --quiet             Don't print anything to STDOUT
+            --exec                  Run docker build command
```
thaJeztah: could you keep this sorted alphabetically?

compose/cli/main.py
```diff
@@ -290,7 +291,8 @@ def build(self, options):
             build_args=build_args,
             gzip=options.get('--compress', False),
             parallel_build=options.get('--parallel', False),
-            silent=options.get('--quiet', False)
+            silent=options.get('--quiet', False),
+            _exec=bool(options.get('--exec', False)),
```
thaJeztah: here as well (I think)

thaJeztah: Perhaps other locations as well, but I'm not a python coder 😅 and not super-familiar with the codebase (just noticed the PR)

compose/cli/main.py
orisano: currently, this codebase is not sorted alphabetically.
should I do it?

compose/service.py
```diff
@@ -1695,3 +1699,135 @@ def rewrite_build_path(path):
         path = WINDOWS_LONGPATH_PREFIX + os.path.normpath(path)
 
     return path
+
+
+def exec_build(path=None, tag=None, quiet=False, fileobj=None,
```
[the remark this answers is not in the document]
orisano: `path=None` is following the original codebase.
https://github.com/docker/docker-py/blob/37e096f6add7e26ada3d6840ce9a9ce341bbdf23/docker/api/build.py#L16-L23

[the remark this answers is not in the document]
ulyssessouza: The point is that in this API `path` is optional because `fileobj` can be used instead and `fileobj` is not used in your PR. Checked in https://github.com/docker/docker-py/blob/37e096f6add7e26ada3d6840ce9a9ce341bbdf23/docker/api/build.py#L125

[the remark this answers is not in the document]
orisano: I understand it.
can i use incompatible signature?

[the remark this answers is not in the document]
orisano: @ulyssessouza I fixed it

compose/service.py
```diff
@@ -1070,7 +1073,8 @@ def build(self, no_cache=False, pull=False, force_rm=False, memory=None, build_a
                 'Impossible to perform platform-targeted builds for API version < 1.35'
             )
 
-        build_output = self.client.build(
+        build = self.client.build if not _exec else exec_build
```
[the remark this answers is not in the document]
orisano: fixed it!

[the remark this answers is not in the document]
orisano: @ulyssessouza i fixed it. ok?

compose/cli/main.py
ulyssessouza: Yep.

Sorted > Unsorted :smiley:

ulyssessouza:
A general doubt I have is about the name of the option... `--exec` sounds very internal pythonic for me.
From an user's point of view I see something like `--native`, `--cli` or anything more generic. 

WDYT @chris-crone @thaJeztah @rumpl @shin-

compose/service.py
```diff
@@ -1695,3 +1699,135 @@ def rewrite_build_path(path):
...
+            single layer.
+        extra_hosts (dict): Extra hosts to add to /etc/hosts in building
+            containers, as a mapping of hostname to IP address.
+        platform (str): Platform in the format ``os[/arch[/variant]]``
+        isolation (str): Isolation technology used during build.
+            Default: `None`.
+    Returns:
+        A generator for the build output.
+    """
+    command_builder = _CommandBuilder()
```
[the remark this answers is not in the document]
orisano: sorted!

compose/cli/main.py
orisano: sorted!

compose/service.py
```diff
@@ -1701,33 +1701,17 @@ def rewrite_build_path(path):
...
 
 
-def exec_build(path, tag=None, quiet=False, fileobj=None,
-               nocache=False, rm=False, timeout=None,
-               custom_context=False, encoding=None, pull=False,
-               forcerm=False, dockerfile=None, container_limits=None,
-               decode=False, buildargs=None, gzip=False, shmsize=None,
-               labels=None, cache_from=None, target=None, network_mode=None,
-               squash=None, extra_hosts=None, platform=None, isolation=None):
+def exec_build(path, buildargs=None, cache_from=None, container_limits=None, custom_context=False,
```
ulyssessouza: Ops... Here I think you've pushed the sorting a bit too far...
The problem here is that anyone using the method with just positional arguments would have the code broken.

orisano: reverted

ulyssessouza:
Please rebase your PR with upstream `master` to pass the CI

orisano:
rebased

ulgens:
ping @ulyssessouza The last request change is already done :slightly_smiling_face: Is there something else missing?

ulgens:
Is there anyway to read `--exec` parameter from environment?

ulyssessouza:
I think that we could get a better name for the option. `--cli` would be better, I think.

Another concern is when executing `docker-compose up`. 
By now, the behavior is to build when there is no image and then run it. And as you can guess, as the `up` command has no such option, the build triggered by `docker-compose up` will never result in a `buildkit` build.

orisano:
Ok, I rename to `--cli`.

orisano:
> By now, the behavior is to build when there is no image and then run it. And as you can guess, as the up command has no such option, the build triggered by docker-compose up will never result in a buildkit build.
I understood.

CpuID:
Anything holding this back now...? Would be great to have compose using buildkit :)

rumours86:
@ulyssessouza need your approve

thesciz:
Excited to have this go in! @ulyssessouza, save us!

nphmuller:
According to his GitHub status he's on vacation. I'm sure he's back soon-ish.

Happy vacation @ulyssessouza! :)

thaJeztah:
Two more thoughts on this (was having a chat with @chris-crone);

- If this is to support buildkit, should we make this look for `DOCKER_BUILDKIT=1` (instead of requiring the user to set a flag?) (it might have to look for the `~/.docker/config.json` file as well 🤔, but looking for the environment variable would be a good start)
- Taking it further; should we _default_ to using the CLI? I'd expect 99% of the docker-compose installs to also have the `docker` cli installed;
  - ❓ are there advantages of docker-compose's build over calling the CLI?
  -  👍 would solve differences in builds; https://github.com/docker/compose/issues/883
  - 👍 would allow taking advantages of BuildKit, and use a single preference to switch between buildkit and legacy builder (no need to configure separately for compose)
  - 👍 over time, buildkit will become the default; this would possibly make the transition easier
  - 👍 also over time, `docker build` will likely get the functionality that's currently being previewed through [`docker buildx bake`](https://github.com/docker/buildx/blob/aed531a8a97fbc41b2e37f37ca94f4a02228139e/README.md#buildx-bake-options-target), which allows building a docker-compose file in a much more efficient way (sharing the build-context, building all services in parallel, but sharing the cache where possible). That functionality will obsolete (e.g.) the `--parallel`, and automatically "do the right thing".
  - ⚠️ we should have logic to fall back to using compose in rare situations where the `docker` CLI is not present
  - ❓ do we know of any side-effects? there should be an opt-out to allow users to force using `docker-compose` in case there's corner-cases we didn't think of

dsvictor94:
@thaJeztah another advantage of not use the `--cli` flag is consistence across build.
If I missed nothing reading this PR the cli mode is only used if build with `docker-compose build` but not during other events that cause the build (e.g docker compose up --build)

lig:
@dsvictor94 btw, there is already a difference between `up --build` and `build` itself, i.e. there is no `build --pull` equivalent for `up --build` (see #3574).

thaJeztah:
> another advantage of not use the --cli flag is consistence across build.
If I missed nothing reading this PR the cli mode is only used if build with docker-compose build but not during other events that cause the build (e.g docker compose up --build)

Good point; didn't think of that 👍 

One concern that was raised is that using the `docker` CLI will result in different formatting of the logs that are shown during build; I _personally_ think that that "con" would be outweighed by all the "pros", but it could be a breaking change for some (so we definitely should consider having the opt-out)

Possible approach would be (all my 0.02c - I'm not a maintainer in this repository);

1. Mark the Compose built-in builder as "frozen" (in maintenance mode);
    - bug fixes only, but no new features in that area
    - significantly reduces engineering time spent on playing "cat and mouse" to to try and catch up with changes in the `docker` CLI builder. There is a lot of functionality being added in the CLI itself, the `dockerd` daemon (through [BuildKit](https://github.com/moby/buildkit)), and through projects, such as [`docker buildx`](https://github.com/docker/buildx), [`docker app`](https://github.com/docker/app), as well as `docker assemble` (Docker Enterprise only for now))
    - Some of those features won't be easy (or even _possible_) to implement in Compose itself, in addition to "duplicating" the work is probably not the best use of engineering.
2. Introduce using the `docker` CLI for build as an (opt-in) experimental feature (`COMPOSE_EXPERIMENTAL=1` or (e.g.) `COMPOSE_NATIVE_BUILDER=1`)
3. Encourage people to use that feature, and **collect feedback**
    - Catch side-effects that we didn't think about, or due to esoteric use-cases/situations "in the wild"
4. After X release(s), make it the default
5. Allow users to opt-out (`COMPOSE_NATIVE_BUILDER=0`)
6. Meanwhile, work on improvements in other projects mentioned to gain native support for Compose files (improving the experience further).

inductor:
@ulyssessouza Still on vacation? :joy:

Stono:
Would really like this too! Can we get it merged?

orisano:
what is the status of this PR now?
what should I do?

nguyensach:
👀

AkihiroSuda:
@rumpl @ulyssessouza  @tonistiigi  PTAL?

There has been huge demand for this

thaJeztah:
@akihirosuda wdyt about my proposed approach? https://github.com/docker/compose/pull/6584#issuecomment-502099098

Stono:
I've been using a branch build of this for a while and I have to echo the concerns about the fact that effectively shelling out to the `docker build` command means totally different interaction with docker compose (different logs etc).

All of our existing build infra that inferred things from those logs fundamentally would have to chance and as a result, we shelved the idea.

I think the compose implementation needs to keep the UX the same, regardless of the underlying builder.

thaJeztah:
There still would be an opt out of course, but perhaps you could describe a bit more in detail how you consume the build logs.

Would it help if the build returned a machine-parseable output (eg JSON/JSON-stream, or any suitable format)?

Note that if we want to fully take advantage of buildkit, that the output _will_ change; buildkit allows building all images concurrently, which means that build steps that are shared between services may be de-duplicated, and/or build asynchronously (and out of order) This may affect how the output is shown.

AnthonBerg:
Regarding the output formatting: Was the old docker build console output formatted to a spec or contract?

If not, then I suggest we allow ourselves to move on from it. The benefits brought by Buildkit are very, very noticeable, and it does naturally call for a different output style.

I don't want to come across as lacking an understanding of practical use of the classical output format! I propose that the `--progress plain` output could be transformed into the old style for those who are vested in it.

Stono:
I'm not saying the output shouldn't change, I'm just saying it should be a considered change is all :-)  At the moment the PR to do this is effectively just shelling out to docker, rather than using docker-py, so you're going to end up with fragmented implementations depending on which you use, and bugs in one tree rather than the other.  It doesn't feel like a mature implementation, I might as well write a quick YAML parser myself that converts it into a `docker build.....` command.

WRT output, anything which I can parse would be great, so for context we currently parse the `docker-compose` build logs looking for multi-stage builds, which we then tag accordingly and push to the registry as separate images, which are then pulled down again in future builds to speed up caching.

So imagine:
````
FROM something as stage1
...
FROM something as stage2
... 
FROM something as build-artefact
COPY --from=stage1 some-file .
COPY --from=stage2 some-file .
```

So we would `docker push` `build-artefact`, obviously as that's what we'll deploy.  It'll only contain however the final binaries from each stage.  However the developers want caching of `stage1` and `stage2` on their next build.  The agent their build runs on in CI won't have those layers, so they also need to be stored in the registry.

Therefore we push `stage1` and `stage2` up to the registry as well, tagged as `stage1` and `stage2`, and then in the next build we do `cache_from: [stage1, stage2]`.  And we do that by parsing the docker-compose output looking for ` : FROM`, with the image id being the preceding line.

It's crude, but it works, and massively reduces our overall build durations for multi-stage projects.

thaJeztah:
> Regarding the output formatting: Was the old docker build console output formatted to a spec or contract?

No, it was never a formalised output, but we did try to not break people if there was no need

> I don't want to come across as lacking an understanding of practical use of the classical output format! I propose that the --progress plain output could be transformed into the old style for those who are vested in it. Using a transformation shim somehow. Out of scope of responsibility of Docker-Compose, but as a friendly utility.

I had a short meeting with @ulyssessouza to discuss this proposal, and we were converging to something like that;

- compose currently doesn't handle the "nice" output of the `docker` cli when buildkit is enabled (because it doesn't detect that a TTY is attached when called through docker compose)
  - research if we can make it see the TTY, so that the output of the `docker` cli is printed correctly
- if possible have an option on the `docker` cli that returns the build output in a machine-parseable format
  - perhaps the "plain" output is suitable, but if not, another format _could_ be added
  - offer this as an option during build
  - as a (future) enhancement; implement code in `docker-compose` to convert that output into something more "like" the existing output of docker-compose
  - (this could be a nice project to work on for people from the community)

Other things we discussed:

If we consider this feature (using my proposed approach https://github.com/docker/compose/pull/6584#issuecomment-502099098)

Refine the approach further

1. start with adding _just_ the flags that `docker-compose build` currently offers

    ```
        --compress              Compress the build context using gzip.
        --force-rm              Always remove intermediate containers.
        --no-cache              Do not use cache when building the image.
        --pull                  Always attempt to pull a newer version of the image.
        -m, --memory MEM        Sets memory limit for the build container.
        --build-arg key=val     Set build-time variables for services.
        --parallel              Build images in parallel.
    ```

    Of note here:

    - `--compress` is not supported by the `docker` cli, and with BuildKit (which uses an interactive session to send the build-context) probably won't be implemented (or have advantages). This could be a "dummy" flag on the docker-compose CLI for backward-compatibility (and be ignored / print a warning)
    - `--parallel` is not useful when using BuildKit (it parallelises automatically), so add a "dummy" flag for this as well
    - `--force-rm` is only useful without BuildKit enabled (but would be added to the `docker-compose` cli)
    - `--build-arg key=val` is already implemented in docker-compose, so likely would be added, but requires some discussion / close look (see below)

2. Add flags that are useful to make optimum use of BuildKit

    - `--progress` (as discussed above)
    - other options?

3. have a close, and careful look at all other options/flags that `docker build` has

    Carefully looks at options that are on `docker build`, but currently not yet in the `docker-compose` CLI. Decide which flags should be added, and which ones should _not_ be added:

    - Is a flag supported by BuildKit?
      - If not; will it be supported in future?
      - if not; don't add the flag, as it would only be added to the legacy builder
    - Does the flag make sense when building the whole stack (compose-file: multiple images being built)?
      - if a flag/option should be applied to _individual_ builds, it may need to find its way into the docker-compose file spec instead, and _not_ be added as a flag (which would apply to _all_ builds)
      - for example,. `--iidfile` has to be different for _each_ build (alternative implementations could still be considered, for example, templating the filename, or specifying a directory in which `iidfile`'s will be written)
      - flags, such as `--label` need discussion (do you want to apply the same label to _all_ images?)
      - `--target`, `--platform`, `--file`, etc. need similar discussions


It's better to start with a _minimum_ set of options, and work from there, than to implement _all_ options without having a "way back". Removing an option will break people, and although this would be an "opt-in" feature initially, it's always harder to _deprecate_ a feature, than to _add_ a feature.

thaJeztah:
Having said the above; we're still interested in learning;

- How the build output is used by people
- What information is used from the output
  - name, tag, and/or digest of the image was built?
  - shell output from the containers/build steps? (if so: in all cases, or only if there's a failure?)
  - is shell output _parsed_ or just collected?
- other information?

AnthonBerg:
Personally, I would parse - and use - the output more if it were formatted in a more explicit and meaningful way (say, JSON). Regex parsing an arbitrary stream is surprisingly effective, but having it nailed down in a structure is much better: `docker build | jq result.layers[0].stderr` or `docker build | jq result.digest`. Or just passing the whole thing through jq for filtering and colorization.

Metadata about the build result would be useful. Any tags tagged, the final image digest, image size. Maybe env vars and build args that were used? Maybe a customizable set of information? Although one could build that out of `inspect` and `history` with the digest. The digest is the most important I guess, and it would be good to have it delivered a little bit more clearly and concisely than the old `docker build` style. Old style is `Successfully built 9434d04e2ab2` while the BuildKit `--progress plain` output is `#24 writing image sha256: 9434d04e2ab2db053866972c1ede497ef2c5d48bb4dfbc126964db8e81827386 done`, which I prefer.

I mostly use shell output when something goes wrong. I might use shell output to emit statistics if it were a bit more contained – what part of the output is from Docker? What's stdout or stderr from the build stage? Which stage?

I much like the auto-collapsed output of `--progress auto`. That only tells me about things from the build perspective itself, not the output of successful stages. I might find benefit in the full logs per stage being available somehow, just out-of-band of the build stdout itself.

AnthonBerg:
my tl;dr might be “decouple build metadata and logging from build implementation”

CpuID:
> Having said the above; we're still interested in learning;
> 
> * How the build output is used by people

I barely use the build output, usually only if something fails, human analysis

> * What information is used from the output
>   * name, tag, and/or digest of the image was built?

Occasionally I'll use the name/tag/digest built, but rarely. Usually feed in a predictable (but sometimes pre-generated) `-t` (for `docker build`) or `image` YAML attribute (for `docker-compose build`) rather than relying on something generated internally.

>   * shell output from the containers/build steps? (if so: in all cases, or only if there's a failure?)
>   * is shell output _parsed_ or just collected?

Collected for human analysis, only on failure usually. Once in a while you might want the output of something like `RUN env` when debugging some weird edge case, or `cat somefile` etc, but again usually not parsed, but if exposed (optionally via feature flag?) would be nice.

> * other information?

CpuID:
> * compose currently doesn't handle the "nice" output of the `docker` cli when buildkit is enabled (because it doesn't detect that a TTY is attached when called through docker compose)
>   
>   * research if we can make it see the TTY, so that the output of the `docker` cli is printed correctly

+1, would be nice to know it looks the same between both `docker` and `docker-compose`

> * if possible have an option on the `docker` cli that returns the build output in a machine-parseable format
>   
>   * perhaps the "plain" output is suitable, but if not, another format _could_ be added
>   * offer this as an option during build
>   * as a (future) enhancement; implement code in `docker-compose` to convert that output into something more "like" the existing output of docker-compose
>   * (this could be a nice project to work on for people from the community)

I think the most common part of the output that would be beneficial would be if a layer is cached or not (cache state).

One other nice to have (but notpossible via Buildkit? and independent of this PR) would be exposing intermediate layer image SHA/digests. I've had to on occasion grab the SHA/digest before a failure and run `docker run -it --rm sha sh` and poke around at the intermediate image and try interactively resolve why a step of a build failed.

jamesharris-garmin:
Have we addressed the changes needed to merge this PR?

cristicbz:
I'm a little worried this fell into the trap of the perfect being the enemy of the good. This PR offers an opt-in, which will offer a good path for folks who need buildkit and don't need build output. The path forward for using the CLI all the time would be good to determine, but it shouldn't block the simple use-case that this PR supports.

thaJeztah:
This PR adds a new flag, in addition to all flags that the CLI supports, which means that all of those would have to be supported for future releases

AnthonBerg:
If that’s a problem, then if this is merged and flagged as experimental it doesn’t have to be supported.

I’m personally running the fork that has the feature and will do so for the foreseeable future.

ulyssessouza:
Hello @orisano. I'm sorry for the late response. As you can see, I just pushed 3 commits related to the internal meetings with @thaJeztah and the team.
The commits include:
- Buildkit's nice TTY output by letting the --progress as "auto"
- Replace --cli by and envvar called `COMPOSE_NATIVE_BUILDER`. 
- Reduce the arguments passed to `docker build` to the already present in `docker-compose build`

PTAL @thaJeztah 

Please note this feature is EXPERIMENTAL and could be removed at any new version

thaJeztah:
Thanks @ulyssessouza - that looks good for a start!

The only thing I'm thinking is if we should add the `--progress` flag to compose as well, so that people can use `--progress=plain` in situations where a terminal is present, but you want the output in a more parsable format (I know we've done so in CI, where a TTY may be attached, you don't need the fancy output)

That could be done as a follow-up though (we'll need follow-ups to discuss / implement other flags anyway)

I haven't tried building/testing this PR yet; did you try if setting `DOCKER_BUILDKIT=1` / `DOCKER_BUILDKIT=0` as (environment) variable works as expected?

```
DOCKER_BUILDKIT=1 docker-compose build ...

DOCKER_BUILDKIT=0 docker-compose build ...
```

We'll also need to make sure the `docker-compose up --build` follows the same behavior, otherwise the cache is not shared between both.

Some other thoughts;

- Should we print a warning if someone has this feature enabled, and uses the `--compress` or `--parallel` flag? (I _think_ both won't do anything when using the CLI for building?)
- Possibly hide them if `COMPOSE_NATIVE_BUILDER=1` (but that's a nice to have)

abhaychrungoo:
I have taken this through some limited manual testing  on a Linux x86 box

Would it be fair to argue that this PR should produce similar behavior for
- docker-compose build && docker-compose up
- docker-compose up --build

Currently, the first uses buildkit, but the second does not,

thaJeztah:
> Would it be fair to argue that this PR should produce similar behavior for

Thanks for testing it! Yes, I think it should work for both (but currently doesn't) (see my comment above yours)

@ulyssessouza is it difficult to make `docker compose up --build` also call out to the CLI for the actual build(s) ?

orisano:
https://github.com/orisano/compose/commit/aac7193b0cf0ee989e9ffcd504d1b18d3a07d327
supports `docker-compose up --build`

ulyssessouza:
Hello all!

Now that the requirements of the PR/issue are quite solid, I just opened a new PR to manage this in a clear way. 
The new one includes all the previous commits and the support for `docker-compose up --build` that @orisano just added squashed in a single commit
So to continue the review of feature, please refer to https://github.com/docker/compose/pull/6865

Once again, thanks a lot for the contribution @orisano !

PR #6588, opened 2019-03-17
Default ?err to (missing) required VAR name. Fixed #6587.

### LINKED ISSUE #6587: Missing mandatory value interp ERROR should mention the variable by name if no custom ?err
**Version info**
```
 uname -a
Darwin some-mac 17.7.0 Darwin Kernel Version 17.7.0: Thu Dec 20 21:47:19 PST 2018; root:xnu-4570.71.22~1/RELEASE_X86_64 x86_64
 docker-compose --version
docker-compose version 1.23.2, build 1110ad01
```

**Summary**
When interpolating a mandatory variable, the default error message generated when the value is unset does not mention the missing variable name.  The error mentions all the other relevant context, just not the variable name itself, which is the most important fact.  It is true that a `?err` could be supplied, but this seems like a DRY and somewhat non-obvious syntax.

**Example**

Consider this trivial `docker-compose.yml`:
```
version: '3.7'
services:
  myservice:
    image: busybox
    environment:
      - MYENV=${MYVAR:?}
    command: ["/bin/sh", "-c", "env | grep MYENV"]
```

Then `docker-compose up` (with no extra environment set) yields:

```
ERROR: Missing mandatory value for "environment" option in service "myservice": 
```
Compose error provides:
- the section name `environment`
- the service name `myservice`

It does _not_ mention:
- the environment value (or any other key) being set: `MYENV`.  That sounds reasonable, but also:
- the (mandatory) variable being interpolated: `MYVAR`.

**Proposed change**
When the interpolation contains a mandatory sep, and does not define a custom error `?err`, Compose could simply default the error to the name of the missing variable.  This would yield for above e.g.:

```
ERROR: Missing mandatory value for "environment" option in service "myservice": MYVAR
```

That is, a missing `err` defaults to `?<variable-name>`.

**Other considerations or workarounds**
As mentioned, an `?err` can be defined to achieve much the same thing, e.g. the somewhat obtuse `- MYENV=${MYVAR:?MYVAR}` yields the desired:

```
ERROR: Missing mandatory value for "environment" option in service "myservice": MYVAR
```
However the syntax is unwieldy, and forces an unnecessary repetition of the variable name in order to produce a custom error message, whereas a standard error message should suffice.  This syntax risks being not universally-applied, and leaves the runner of `docker-compose` unsure of which variable(s) (of potentially many) are unset.

**Alternatives to consider**
The proposal is to default `err` to `?<variable-name>` for mandatory variable interpolation.  Another option would be to always pre-pend <variable-name> to `err` even when present.  This seems like a compatibility issue and doesn't do-as-I-ask ... if you want to specify a custom `?err`, then you are responsible for also mentioning the variable name, if that is how you want it reported.

javabrett:
Changes the default (no custom `?err`) error-message for a missing required variable ${MYVAR:?} to end with `: MYVAR` e.g.

```
ERROR: Missing mandatory value for "environment" option in service "myservice": MYVAR
```

Behaviour with `${MYVAR:?err}` remains unchanged.

Fixed #6587.

### DIFF
--- a/compose/config/interpolation.py
+++ b/compose/config/interpolation.py
@@ -64,12 +64,12 @@ def interpolate_value(name, config_key, value, section, interpolator):
                 string=e.string))
     except UnsetRequiredSubstitution as e:
         raise ConfigurationError(
-            'Missing mandatory value for "{config_key}" option in {section} "{name}": {err}'.format(
-                config_key=config_key,
-                name=name,
-                section=section,
-                err=e.err
-            )
+            'Missing mandatory value for "{config_key}" option interpolating {value} '
+            'in {section} "{name}": {err}'.format(config_key=config_key,
+                                                  value=value,
+                                                  name=name,
+                                                  section=section,
+                                                  err=e.err)
         )
 
 


jcsirot:
Thank you for this PR.
I'm wondering if we would not prefer always displaying the variable name in the error message along with the additional error message (independently of the fact whether it is empty or not). 
What do you think @ulyssessouza ?

javabrett:
Is there a preference developing for just forcing the variable into the message/template?  Need to update the PR?

Preferred format (with no `err`)?:

```
ERROR: Missing mandatory value for "environment" variable "MYVAR" in service "myservice":
```

That now has everything but the key e.g. `MYENV` in #6587.

@jcsirot @ulyssessouza

javabrett:
To progress this, I have pushed a second commit, which undoes the first option presented above, and instead outputs the format:

```
ERROR: Missing mandatory value for "environment" option interpolating ['MYENV=${MYVAR:?}'] in service "myservice":
```

Can make formatting changes if required.  I prefer this format now - it includes the exact string being interpolated, so you get both the missing variable `MYVAR` and the variable it would have been assigned to `MYENV`.

If you prefer the original proposal, just drop the second commit on this branch.

javabrett:
@ulyssessouza thanks, squashes and improved commit-log message.

ulyssessouza:
LGTM

PR #6592, opened 2019-03-19
Fixed depends_on recreation behaviour for issue #6589

### LINKED ISSUE #6589: depends_on recreation behaviour - all dependants recreated
## Description of the issue

After making a change to a service, any services marked as dependent (via `depends_on`) are recreated upon running `docker-compose up`

My understanding was that unless the `--always-recreate-deps` flag was used with `docker-compose up`, dependent containers should not be recreated.

This behaviour seems to be very similar to this old issue https://github.com/docker/compose/issues/3397 which was deemed to indeed be a bug and dependent containers should not always be restarted. I've had a look at the fix https://github.com/docker/compose/pull/5541 which was merged in for this and think the issue is caused simply by an extraneous `not` on this [line](https://github.com/docker/compose/pull/5541/files#diff-c8c735e30c07f4c7494904b4570d1541R526).

Given the following example composition
```
version: '3.7'
services:
  nginx1:
    image: nginx:latest

  nginx2:
    image: nginx:latest
    depends_on:
    - nginx1
```

1. `docker-compose up -d`
2. change `nginx1`'s image tag e.g.`alpine`
3. `docker-compose up -d`

### Observed result

1. `nginx1` is recreated as expected
1. `nginx2` container is also recreated, despite having *not* been updated
```
docker-compose up -d
Recreating miscellany_nginx1_1 ... done
Recreating miscellany_nginx2_1 ... done
```

### Expected result

1. `nginx1` is recreated as expected *only*.
1. `nginx2` container remains running
```
docker-compose up -d
Recreating miscellany_nginx1_1 ... done
miscellany_nginx2_1 is up-to-date
```

```
docker-compose version 1.23.1, build b02f130
docker-py version: 3.5.1
CPython version: 3.5.6
```

```
Client:
 Version:           18.09.0
 API version:       1.39
 Go version:        go1.10.4
 Git commit:        4d60db4
 Built:             Wed Nov  7 00:49:01 2018
 Experimental:      false

 Engine:
  Version:          18.09.0
  Go version:       go1.10.4
  Git commit:       4d60db4
  Built:            Wed Nov  7 00:16:44 2018
  OS/Arch:          linux/amd64
  Experimental:     false
```

```
services:
  nginx1:
    image: nginx:latest
  nginx2:
    depends_on:
    - nginx1
    image: nginx:latest
version: '3.7'
```

## Additional information
```
uname -a    
Linux joebox 4.15.0-45-generic #48-Ubuntu SMP Tue Jan 29 16:28:13 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux
```

joeweoj:
Previously any containers which did *not* have any links were always recreated.
In order to fix depends_on and preserve expected links recreation behaviour, we now only use the ConvergenceStrategy.always recreation strategy if the service defines links but the container does not have any.

Resolves #6589

### DIFF
--- a/compose/project.py
+++ b/compose/project.py
@@ -586,8 +586,10 @@ class Project(object):
                           ", ".join(updated_dependencies))
                 containers_stopped = any(
                     service.containers(stopped=True, filters={'status': ['created', 'exited']}))
-                has_links = any(c.get('HostConfig.Links') for c in service.containers())
-                if always_recreate_deps or containers_stopped or not has_links:
+                service_has_links = any(service.get_link_names())
+                container_has_links = any(c.get('HostConfig.Links') for c in service.containers())
+                should_recreate_for_links = service_has_links ^ container_has_links
+                if always_recreate_deps or containers_stopped or should_recreate_for_links:
                     plan = service.convergence_plan(ConvergenceStrategy.always)
                 else:
                     plan = service.convergence_plan(strategy)
--- a/tests/integration/state_test.py
+++ b/tests/integration/state_test.py
@@ -5,6 +5,8 @@ by `docker-compose up`.
 from __future__ import absolute_import
 from __future__ import unicode_literals
 
+import copy
+
 import py
 from docker.errors import ImageNotFound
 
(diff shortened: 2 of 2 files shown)


jcsirot:
Thank you for this PR. Please could you ass a test covering the bug it fixes.

joeweoj:
@jcsirot I have added some tests to verify the fixed `depends_on` behaviour (i verified the `links` tests still passed when i created the PR).

Let me know if you have any questions. Thanks

jcsirot:
thank you @joeweoj.
@ulyssessouza PTAL

tests/integration/state_test.py
```diff
@@ -209,6 +209,108 @@ def test_service_recreated_when_dependency_created(self):
...
+
+        assert set(c.service for c in new_containers - old_containers) == set(['web'])
+
+    def test_change_middle_always_recreate_deps(self):
+        old_containers = self.run_up(self.cfg, always_recreate_deps=True)
+
+        self.cfg['services']['web']['environment'] = {'NEW_VAR': '1'}
+        new_containers = self.run_up(self.cfg, always_recreate_deps=True)
+
+        assert set(c.service for c in new_containers - old_containers) == {'web', 'nginx'}
```
ijc: Shouldn't this be `set(['web', 'nginx'])`,  for consistency with (nearly) all the other tests if no other reason?

tests/integration/state_test.py
```diff
@@ -209,6 +209,108 @@ def test_service_recreated_when_dependency_created(self):
...
+
+        self.cfg['services']['web']['environment'] = {'NEW_VAR': '1'}
+        new_containers = self.run_up(self.cfg)
+
+        assert set(c.service for c in new_containers - old_containers) == set(['web'])
+
+    def test_change_middle_always_recreate_deps(self):
+        old_containers = self.run_up(self.cfg, always_recreate_deps=True)
+
+        self.cfg['services']['web']['environment'] = {'NEW_VAR': '1'}
```
ijc: Since `self.cfg` is shared between all the sub tests, and the previous test already did this, does this actually change anything? Does it matter that it might not?

For safety it might make sense to use a unique name for each test, or to only every use/mutate copies of the base cfg.

tests/integration/state_test.py
```diff
@@ -209,6 +209,108 @@ def test_service_recreated_when_dependency_created(self):
...
+        next_cfg = {
+            'version': '2',
+            'services': {
+                'web': {
+                    'image': 'busybox:latest',
+                    'command': 'tail -f /dev/null',
+                },
+                'nginx': self.cfg['services']['nginx'],
+            }
+        }
```
ijc: I think it'd be clearer what was going on if this instead took a deep copy of `self.cfg` and then did `delete(next_cfg.services.db)` + whatever other mutations are expected.

tests/integration/state_test.py
```diff
@@ -209,6 +209,108 @@ def test_service_recreated_when_dependency_created(self):
...
+
+        containers = self.run_up(self.cfg)
+        assert len(containers) == 3
+
+        project = self.make_project(self.cfg)
+        project.stop(timeout=1)
+
+        containers = self.run_up(next_cfg)
+        assert len(containers) == 2
```
ijc: Better to assert they are the expected/correct 2?

compose/project.py
```diff
@@ -586,8 +586,10 @@ def _get_convergence_plans(self, services, strategy, always_recreate_deps=False)
                           ", ".join(updated_dependencies))
                 containers_stopped = any(
                     service.containers(stopped=True, filters={'status': ['created', 'exited']}))
-                has_links = any(c.get('HostConfig.Links') for c in service.containers())
-                if always_recreate_deps or containers_stopped or not has_links:
+                service_has_links = any(service.get_link_names())
+                container_has_links = any(c.get('HostConfig.Links') for c in service.containers())
+                should_recreate_for_links = service_has_links ^ container_has_links
```
ijc: It took me a while to puzzle out this xor, essentially it is reconcilling what the service asks for (which may have changed) with what the running container has.I think my main mental barrier was the commit message saying "if the service defines links but the container does not have any" which is only half the story, since we also should reconcille if the service has no links but the container does have them. If that's correct would you mind updating the commit message to reflect that please.

ijc: I think the tests don't actually cover the case of adding / removing links dynamically (only close one is `test_service_removed_while_down` but the project is down for that change). Is that also right?

tests/integration/state_test.py
joeweoj: changed this and all other references to use `set['svc1', ...]`

tests/integration/state_test.py
joeweoj: yes good point. each test now uses a deep cloned copy of the config

tests/integration/state_test.py
joeweoj: agreed, have changed to 
```
        next_cfg = copy.deepcopy(self.cfg)
        del next_cfg['services']['db']
        del next_cfg['services']['web']['depends_on']
```

tests/integration/state_test.py
joeweoj: have changed any and all collection length assertions to be explicit

compose/project.py
joeweoj: i have added 2 new tests `test_dependency_removed` and `test_dependency_added` which should explicitly verify the adding/removing links dynamically behaves as expected

joeweoj: have also modified the commit message as requested. hopefully addresses your comment

tests/integration/state_test.py
```diff
@@ -209,6 +211,145 @@ def test_service_recreated_when_dependency_created(self):
...
+    def test_change_root_always_recreate_deps(self):
+        local_cfg = copy.deepcopy(self.cfg)
+        old_containers = self.run_up(local_cfg, always_recreate_deps=True)
+
+        local_cfg['services']['db']['environment'] = {'NEW_VAR': '1'}
+        new_containers = self.run_up(local_cfg, always_recreate_deps=True)
+
+        assert set(c.service for c in new_containers - old_containers) == {
+            'db', 'web', 'nginx'
+        }
```
ijc: another missing `set(...)`?

joeweoj:
@jcsirot now the change has been approved, can i ask what the next steps are?
we're obviously keen to get this fix into the next release. 
thanks

ijc:
Next step is I merge it :smile:. Done, thanks!

joeweoj:
lovely. thanks very much!

PR #6595, opened 2019-03-21
Fix Secrets mount source path while using Dockertoolbox

### LINKED ISSUE #6585: Secrets invalid mount config for type 'bind' mount path must be absolute
## Description of the issue
Docker compose does not convert secrets source path correctly while using docker toolbox (virtualbox).

```
docker-py version: 3.7.0
CPython version: 3.7.2
OpenSSL version: OpenSSL 1.1.0j  20 Nov 2018
```

```
Client:
 Version:           18.09.2
 API version:       1.39
 Go version:        go1.11.5
 Git commit:        62479626f2
 Built:             Wed Feb 13 09:02:49 2019
 Experimental:      false

 Engine:
  Version:          18.09.3
  Go version:       go1.10.8
  Git commit:       774a1f4
  Built:            Thu Feb 28 06:40:51 2019
  OS/Arch:          linux/amd64
  Experimental:     false
```

```
secrets:
  password:
    file: C:\Users\Joaom\teste\secrets\password
  user:
    file: C:\Users\Joaom\teste\secrets\user
services:
  rice:
    image: alpine:3.8
    secrets:
    - source: password
    - source: user
version: '3.6'                                                                                                                                                             
```

1. Install docker toolbox latest release from Docker toolbox repository on a Windows 10 Home Edition computer.
2. docker-compose up -d

### Observed result
Secrets are not mounted in the container, because compose is not converting path from Windows to unix Style correctly like it does for volumes.
### Expected result
secrets to be mounted successfully.

```
Creating network "teste_default" with the default driver
Creating teste_rice_1 ... error

ERROR: for teste_rice_1  Cannot create container for service rice: invalid mount config for type "bind": invalid mount path: 'C:/Users/Joaom/teste/secrets/password' mount path must be absolute

ERROR: for rice  Cannot create container for service rice: invalid mount config for type "bind": invalid mount path: 'C:/Users/Joaom/teste/secrets/password' mount path must be absolute
```

BluZz95:
Resolves #6585


compose/config/config.py
```diff
@@ -874,6 +874,20 @@ def finalize_service_volumes(service_dict, environment):
     return service_dict
 
 
+def finalize_service_secrets(service_dict, environment):
+    if 'secrets' in service_dict:
+        finalized_secrets = []
+        normalize = environment.get_boolean('COMPOSE_CONVERT_WINDOWS_PATHS')
+        win_host = environment.get_boolean('COMPOSE_FORCE_WINDOWS_HOST')
+        for v in service_dict['secrets']:
+            if isinstance(v, dict):
```
ijc: I'm unsure about this check, at this point I think what we must have in our hand is the result of an earlier call to `types.ServiceSecret.parse(s)` which is a subclass of `collections.namedtuple` with a `repr` that returns a `dict`.

So I think this if can be removed, i.e .making the `finalized_secrets.append(...)` unconditional.

The variant in `finalize_service_volumes` is different because in that context the thing in hand could be a `dict` (`MountSpec`) or a `tuple` (`VolumeSpec`) hence the check.

jmadeira95:
So i don't think i can make a similar function for secrets like the one for volumes (finalize_service_volumes).
From what i have seen, secrets are divided into 2-3 parts which are only later converged into a final object (MountSpec). 
The one part containing the source path is loaded through this function:

```
def load_mapping(config_files, get_func, entity_type, working_dir=None): 
    mapping = {}

        for name, config in getattr(config_file, get_func)().items():
            mapping[name] = config or {}
            if not config:
                continue

            external = config.get('external')
            if external:
                validate_external(entity_type, name, config, config_file.version)
                if isinstance(external, dict):
                    config['name'] = external.get('name')
                elif not config.get('name'):
                    config['name'] = name

            if 'driver_opts' in config:
                config['driver_opts'] = build_string_dict(
                    config['driver_opts']
                )

            if 'labels' in config:
                config['labels'] = parse_labels(config['labels'])

            if 'file' in config:
                config['file'] = expand_path(working_dir, config['file'])

    return mapping
````

 which is used to obtain a dict with the path of the secret file.

```
secrets = load_mapping(
        config_details.config_files, 'get_secrets', 'Secret', config_details.working_dir
    )
```

which i after apply the path conversion according to:

```
normalize = environment.get_boolean('COMPOSE_CONVERT_WINDOWS_PATHS')
win_host = environment.get_boolean('COMPOSE_FORCE_WINDOWS_HOST')
```

howcroft:
Hi there @ijc 
Thanks for supporting @jmadeira95 on getting this PR going in the right direction.
Other than maybe signing-off on all commits, what else would you say would be needed to get this "Mergeable"?

PR #6623, opened 2019-03-31
support PyYAML up to 5.x version

### LINKED ISSUE #6619: Allow PyYAML >= 5
The PyYAML team recently released version 5.1 of their lib to properly fix CVE-2017-18342 vulnerability. It would be great to add support for this version in docker-compose.

GeyseR:
Resolves #6619
The https://github.com/yaml/pyyaml/issues/265 ticket in the original PyYAML mention several backward-incompatible changes in v.5.1:
1) Deprecate yaml.load and add FullLoader and UnsafeLoader classes. docker-compose uses save yaml files parser, so it should be pretty safe to update to the new version.
2) Make default_flow_style=False. The only place in the docker-compose project where dump is used already set `default_flow_style` argument to `False`

Also, in the same ticket author suggests waiting for the next release if the project uses `yaml.add_constructor`, `YAMLObject`, `from_yaml`, `to_yaml`, which seems unrelated to the docker-compose project.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -17,7 +17,7 @@ paramiko==2.6.0
 pypiwin32==219; sys_platform == 'win32' and python_version < '3.6'
 pypiwin32==223; sys_platform == 'win32' and python_version >= '3.6'
 PySocks==1.7.1
-PyYAML==4.2b1
+PyYAML==5.3
 requests==2.22.0
 six==1.12.0
 subprocess32==3.5.4; python_version < '3.2'
--- a/setup.py
+++ b/setup.py
@@ -32,7 +32,7 @@ def find_version(*file_paths):
 install_requires = [
     'cached-property >= 1.2.0, < 2',
     'docopt >= 0.6.1, < 1',
-    'PyYAML >= 3.10, < 5',
+    'PyYAML >= 3.10, < 6',
     'requests >= 2.20.0, < 3',
     'texttable >= 0.9.0, < 2',
     'websocket-client >= 0.32.0, < 1',
--- a/tests/acceptance/cli_test.py
+++ b/tests/acceptance/cli_test.py
@@ -269,7 +269,7 @@ services:
         # assert there are no python objects encoded in the output
         assert '!!' not in result.stdout
 
-        output = yaml.load(result.stdout)
+        output = yaml.safe_load(result.stdout)
         expected = {
             'version': '2.0',
             'volumes': {'data': {'driver': 'local'}},
@@ -294,7 +294,7 @@ services:
     def test_config_restart(self):
         self.base_dir = 'tests/fixtures/restart'
         result = self.dispatch(['config'])
-        assert yaml.load(result.stdout) == {
+        assert yaml.safe_load(result.stdout) == {
             'version': '2.0',
             'services': {
                 'never': {
@@ -323,7 +323,7 @@ services:
     def test_config_external_network(self):
         self.base_dir = 'tests/fixtures/networks'
         result = self.dispatch(['-f', 'external-networks.yml', 'config'])
-        json_result = yaml.load(result.stdout)
+        json_result = yaml.safe_load(result.stdout)
         assert 'networks' in json_result
         assert json_result['networks'] == {
             'networks_foo': {
@@ -337,7 +337,7 @@ services:
     def test_config_with_dot_env(self):
         self.base_dir = 'tests/fixtures/default-env-file'
         result = self.dispatch(['config'])
-        json_result = yaml.load(result.stdout)
+        json_result = yaml.safe_load(result.stdout)
         assert json_result == {
             'services': {
                 'web': {
@@ -352,7 +352,7 @@ services:
     def test_config_with_env_file(self):
         self.base_dir = 'tests/fixtures/default-env-file'
         result = self.dispatch(['--env-file', '.env2', 'config'])
-        json_result = yaml.load(result.stdout)
+        json_result = yaml.safe_load(result.stdout)
         assert json_result == {
             'services': {
                 'web': {
@@ -367,7 +367,7 @@ services:
     def test_config_with_dot_env_and_override_dir(self):
         self.base_dir = 'tests/fixtures/default-env-file'
         result = self.dispatch(['--project-directory', 'alt/', 'config'])
-        json_result = yaml.load(result.stdout)
+        json_result = yaml.safe_load(result.stdout)
         assert json_result == {
             'services': {
                 'web': {
@@ -382,7 +382,7 @@ services:
     def test_config_external_volume_v2(self):
         self.base_dir = 'tests/fixtures/volumes'
         result = self.dispatch(['-f', 'external-volumes-v2.yml', 'config'])
-        json_result = yaml.load(result.stdout)
+        json_result = yaml.safe_load(result.stdout)
         assert 'volumes' in json_result
         assert json_result['volumes'] == {
             'foo': {
@@ -398,7 +398,7 @@ services:
     def test_config_external_volume_v2_x(self):
         self.base_dir = 'tests/fixtures/volumes'
         result = self.dispatch(['-f', 'external-volumes-v2-x.yml', 'config'])
-        json_result = yaml.load(result.stdout)
+        json_result = yaml.safe_load(result.stdout)
         assert 'volumes' in json_result
         assert json_result['volumes'] == {
             'foo': {
@@ -414,7 +414,7 @@ services:
     def test_config_external_volume_v3_x(self):
         self.base_dir = 'tests/fixtures/volumes'
         result = self.dispatch(['-f', 'external-volumes-v3-x.yml', 'config'])
-        json_result = yaml.load(result.stdout)
+        json_result = yaml.safe_load(result.stdout)
         assert 'volumes' in json_result
         assert json_result['volumes'] == {
             'foo': {
@@ -430,7 +430,7 @@ services:
     def test_config_external_volume_v3_4(self):
         self.base_dir = 'tests/fixtures/volumes'
         result = self.dispatch(['-f', 'external-volumes-v3-4.yml', 'config'])
-        json_result = yaml.load(result.stdout)
+        json_result = yaml.safe_load(result.stdout)
         assert 'volumes' in json_result
         assert json_result['volumes'] == {
             'foo': {
@@ -446,7 +446,7 @@ services:
     def test_config_external_network_v3_5(self):
         self.base_dir = 'tests/fixtures/networks'
         result = self.dispatch(['-f', 'external-networks-v3-5.yml', 'config'])
-        json_result = yaml.load(result.stdout)
+        json_result = yaml.safe_load(result.stdout)
         assert 'networks' in json_result
         assert json_result['networks'] == {
             'foo': {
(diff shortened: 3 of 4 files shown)


GeyseR:
Any thoughts?

chris-crone:
Hi @GeyseR!

Thanks for the PR. As @ijc said in the issue, it would be good to wait for pyyaml 5.2 because of known issues. When that's out, can you please update this PR?

GeyseR:
Do you think those issues are critical for the compose project? In the PR description, I've added an explanation of why we can allow 5.1 without any breakage of the compose functionality.

chris-crone:
Taking a more in depth look:
> Deprecation of `yaml.load`

 We use this in a couple of places so we should move to `yaml.safe_load` in the PR that bumps pyyaml.

> Also the logic for detecting special characters was broken, which results in strings with tabs (for example) to be dumped in plain style on python 2.7 on systems with sys.maxunicode <= 0xffff.

That is worrying as we support Python 2.7.

Thank you for this PR and for taking the time to investigate this properly. Given the Python 2.7 issue though, I think it's best to hold off until pyyaml 5.2.

GeyseR:
> We use this in a couple of places so we should move to yaml.safe_load in the PR that bumps pyyaml.
i've updated all calls to `yaml.load` to `yaml.save_load`. All those places were in tests, the actual package code already uses `yaml.safe_load`/`yaml.safe_dump` everywhere.

> That is worrying as we support Python 2.7.
yea, I didn't take it into account while I probably had to. I would agree that we should wait for the pyyaml v.5.2. Do you think we should keep this PR and set it On Hold, or open a new one when 5.2 will be released? (In case of the first option I can take care of updating existing PR according to the changes in the new pyyaml release)

chris-crone:
> i've updated all calls to yaml.load to yaml.save_load. All those places were in tests, the actual package code already uses yaml.safe_load/yaml.safe_dump everywhere.

Thank you!

> yea, I didn't take it into account while I probably had to. I would agree that we should wait for the pyyaml v.5.2. Do you think we should keep this PR and set it On Hold, or open a new one when 5.2 will be released? (In case of the first option I can take care of updating existing PR according to the changes in the new pyyaml release)

Since you've done the work of moving to `safe_load`, we can leave this open and wait for 5.2. It's useful to have this as we have the CI runs with the changes so we can see if they break anything.

richardhanson:
Unless I'm missing something, the Python 2 issue is not relevant here. Since the package code already uses `safe_load` everywhere and this is just a dependency and test update, can this be merged? I'm working with a pytest plugin that is affected by this.

Thank you for being cautious and careful!

graingert:
> That is worrying as we support Python 2.7.

although that seems to only be broken in `yaml.dump` which isn't used, except in an example script which uses ruamel.yaml.dump

https://github.com/docker/compose/blob/75d5eb010851ddc71fa5cfb3728e399fdbf08128/contrib/migration/migrate-compose-file-v1-to-v2.py#L136-L142

graingert:
ah, no there's a dump here:

https://github.com/docker/compose/blob/75d5eb010851ddc71fa5cfb3728e399fdbf08128/compose/config/serialize.py#L105-L112

hartwork:
@GeyseR can you rebase this against latest master to address the conflicts?

My vote for getting this into 1.25 (rather than 1.26).

hartwork:
There is one more call to `.load` that might be worth checking, this time with `ruamel`, though:
```diff
--- a/contrib/migration/migrate-compose-file-v1-to-v2.py
+++ b/contrib/migration/migrate-compose-file-v1-to-v2.py
@@ -19,7 +19,7 @@ log = logging.getLogger('migrate')
 
 
 def migrate(content):
-    data = ruamel.yaml.load(content, ruamel.yaml.RoundTripLoader)
+    data = ruamel.yaml.safe_load(content, ruamel.yaml.RoundTripLoader)
 
     service_names = data.keys()
 
```

GeyseR:
Hey @hartwork 
Thanks for looking into this issue and PR. I definitely support your opinion about releasing this as part of 1.25 release. I've rebased my changed against current master.
Unfortunately, the pyyaml team hasn't released 5.2 version yet, so it looks like there could be some incompatibility problems with Python 2.7

> There is one more call to .load that might be worth checking, this time with ruamel

compose uses ruamel's RoundTripLoader for loading files, that uses RoundTripConstructor that in turn inherits from SafeConstructor. So we should be ok there.

GeyseR:
rebased on current master

GeyseR:
Hi @ulyssessouza, I think we should still wait for pyyaml 5.2 to be released (hopefully it will happen soon, their team just released a 5.2b1 version)

GeyseR:
It seems that all of the incompatibles have been fixed and released in PyYAML 5.2.
I've updated this PR accordingly, so I think it is ready to merge in the master branch

setup.py
```diff
@@ -32,7 +32,7 @@ def find_version(*file_paths):
 install_requires = [
     'cached-property >= 1.2.0, < 2',
     'docopt >= 0.6.1, < 1',
-    'PyYAML >= 3.10, < 5',
+    'PyYAML >= 3.10, < 5.3',
```
graingert: do we suspect that pyYAML will make breaking changes in v5.3?

```suggestion
    'PyYAML >= 3.10, < 6',
```

GeyseR: I just wanted to be more conservative there. I think it would be better to review PyYAML's changelog before the next version update in setup.py.

graingert: Everything else is pinned at a major version

GeyseR: fair enough, +1 for consistency :)

graingert:
https://github.com/docker/compose/pull/7089/files

graingert:
@rumpl will this make it for 1.25.1?

graingert:
@ulyssessouza how many reviewers are required for a PR?

ulyssessouza:
@graingert A PR normally needs the approval of 2 maintainers. This PR will get merged just after `1.25.1` is released.

The reason is that `1.25.1-rc1` is on testing phase to be promoted to `1.25.1` and as this is a major version change, that would be better to have this tested in an RC before.
This does not explain why it's not merged. And this is because the release process is quite tricky, so we prefer to have a clean `master` branch just to have a smooth release.
This should change for 2020 thanks to the work of @ndeloof in https://github.com/docker/compose/pull/7079

I'm very sorry to have a PR that LGTM not merged for that long but we are working on "cleaning the house" to get better workflows.

graingert:
@ulyssessouza will it make a 1.25.x release? Eg py2 still supported?

ulyssessouza:
@graingert Probably not

kohtala:
So Python3 would work with PyYAML 5.x? Would it help to use [environment markers](https://www.python.org/dev/peps/pep-0508/#environment-markers) and provide different PyYAML requirement for python 2.7 and 3.x?

So in `setup.py`

    'PyYAML >= 3.10, < 5',

would be like

    'PyYAML >= 3.10, < 5; python_version == "2.7"',
    'PyYAML >= 3.10, < 6; python_version != "2.7"',

There are other choises of comparisons, but that should work for existing and future versions.

I am a little impatient as I am fighting now to use two libraries at the same time where other requires PyYAML 5. I regret moving to Pipenv as I find no way to tell it to ignore this requires-dist that does not affect our use.

graingert:
> So Python3 would work with PyYAML 5.x? Would it help to use [environment markers](https://www.python.org/dev/peps/pep-0508/#environment-markers) and provide different PyYAML requirement for python 2.7 and 3.x?
> 
> So in `setup.py`
> 
> ```
> 'PyYAML >= 3.10, < 5',
> ```
> 
> would be like
> 
> ```
> 'PyYAML >= 3.10, < 5; python_version == "2.7"',
> 'PyYAML >= 3.10, < 6; python_version != "2.7"',
> ```
> 
> There are other choises of comparisons, but that should work for existing and future versions.
> 
> I am a little impatient as I am fighting now to use two libraries at the same time where other requires PyYAML 5. I regret moving to Pipenv as I find no way to tell it to ignore this requires-dist that does not affect our use.

There would be no point in doing that because this PR provides pyaml 3.10 to 6 support for both py2 and py3

ulyssessouza:
Python2 will be dropped in version 1.26.0 anyways. Check https://github.com/docker/compose/pull/7031

graingert:
> Python2 will be dropped in version 1.26.0 anyways. Check #7031

This pr is compatible with a v1.25.x release though

requirements.txt
```diff
@@ -17,7 +17,7 @@ paramiko==2.6.0
 pypiwin32==219; sys_platform == 'win32' and python_version < '3.6'
 pypiwin32==223; sys_platform == 'win32' and python_version >= '3.6'
 PySocks==1.6.7
-PyYAML==4.2b1
+PyYAML==5.2
```
ulyssessouza: 5.3 has been released, that would be nice to give it a try

graingert:
```suggestion
PyYAML==5.3
```

GeyseR: 👍

ulyssessouza:
Just squash the commits and that's good for me

graingert:
@GeyseR this will also need a rebase to deal with the conflicts in requirements.txt

@ulyssessouza if you merge PRs that manually change requirements.txt and conflict with dependabot PRs dependabot will automatically rebase to fix the conflicts, this way you don't have to wait for a human to resolve the conflicts

kohtala:
> There would be no point in doing that because this PR provides pyaml 3.10 to 6 support for both py2 and py3

I know. I do not disrespects this PR. If this PR is really going in now, I am happy, but I was maybe unclear in asking if another PR could go in while waiting for this.

docker-compose can not be currently installed with some other libraries. It is a blocker, and a partial solution would be better than keeping those unaffected waiting.

graingert:
> > There would be no point in doing that because this PR provides pyaml 3.10 to 6 support for both py2 and py3
> 
> I know. I do not disrespects this PR. If this PR is really going in now, I am happy, but I was maybe unclear in asking if another PR could go in while waiting for this.
> 
> docker-compose can not be currently installed with some other libraries. It is a blocker, and a partial solution would be better than keeping those unaffected waiting.

There is no partial solution that is py3 only, the code change in this PR is required to support pyyaml~=5.2 on either py3 or py2

graingert:
@GeyseR can you rebase and squash everything into 1 commit?

GeyseR:
@graingert @ulyssessouza I've squashed commits and rebased changes against the actual master branch

ulyssessouza:
Thank you all and specially @GeyseR for the patience!

graingert:
@ulyssessouza great stuff! Do you know when the next release will be?

ulyssessouza:
That will be 1.26.0 for sure but for the when, that depends on the advancement on the [CI rework](https://github.com/docker/compose/pull/7133)

graingert:
@ulyssessouza I see docker-compose==1.25.2rc2 has `'PyYAML >= 3.10, < 6',` and I see https://github.com/docker/compose/releases/tag/v1.25.2 but it's not on pypi yet

hartwork:
It is now: https://pypi.org/project/docker-compose/1.25.2/

graingert:
@hartwork awesome I'm upgrading everywhere now!

Issue #6617, opened 2019-03-28
Pip install of docker-compose 1.24 fails


carlmjohnson:
## Description of the issue
Running `pip install docker-compose==1.23.2` succeeds but `pip install docker-compose` fails.

## Context information (for bug reports)
Error output is: `/usr/bin/python2 -m pip install --ignore-installed --no-user --prefix /tmp/pip-build-env-Jeq6Db --no-warn-script-location --no-binary :none: --only-binary :none: -i https://pypi.org/simple -- setuptools wheel "cffi>=1.1; python_implementation != 'PyPy'"" failed with error code 1 in None`

## Steps to reproduce the issue

Against docker:stable image run `apk add --quiet --no-cache bash py-pip && pip install --quiet docker-compose==1.24.0`

### Observed result
```
/usr/bin/python2 -m pip install --ignore-installed --no-user --prefix /tmp/pip-build-env-Jeq6Db --no-warn-script-location --no-binary :none: --only-binary :none: -i https://pypi.org/simple -- setuptools wheel "cffi>=1.1; python_implementation != 'PyPy'"" failed with error code 1 in None`
```
### Expected result
Successful install.

## Additional information

Running Docker-in-Docker for Gitlab test runner. Probably not relevant?

AnthonyMastrean:
@carlmjohnson are you running inside of an Alpine container? I am and I'm seeing the same issue. The top half of my `.gitlab-ci.yml` file looked like this:

```yml
image: docker:stable

services:
  - docker:stable-dind

before_script:
  - apk add --no-cache --quiet py-pip
  - pip install --quiet docker-compose
  - docker login --username "$CI_REGISTRY_USER" --password "$CI_REGISTRY_PASSWORD" "$CI_REGISTRY"
```

But, I had to change it to this

```yml
image: docker:stable

services:
  - docker:stable-dind

before_script:
  - apk add --no-cache --quiet py-pip
  - pip install --quiet docker-compose~=1.23.0
  - docker login --username "$CI_REGISTRY_USER" --password "$CI_REGISTRY_PASSWORD" "$CI_REGISTRY"
```

carlmjohnson:
Yes, my `.gitlab-ci.yml` was essentially the same, and my fix has been to use the prior version of docker-compose.

outofmemoryagain:
I can confirm the same issue, and same fix... revert to previous version.

kynetiv:
[...]

Issue #6614, opened 2019-03-27
Change Docker image entrypoint to shell


ebuildy:
**Is your feature request related to a problem? Please describe.**

Using the image ``docker/compose``, the entrypoint is ``docker-compose`` bin, this makes the image difficult to use inside CI tools such as GitlabCI that require a shell.

Also, as an official Docker® image, this image should work like image ``docker`` (see https://github.com/docker-library/docker/blob/6e5ccf3c71e86b98bb75600b4309ccf1bc87775e/18.06/docker-entrypoint.sh), they can execute ``docker`` command or default shell command.

**Describe the solution you'd like**

The entrypoint of ``docker/compose`` image should have the same logic than ``docker`` image. Or you could create a new flavor ``docker/compose:sh`` that have default shell entrypoint.

**Describe alternatives you've considered**
I can override the entrypoint or create a new public Docker image, extending ``docker/compose`` with the same entrypoint logic than ``docker`` image. 

**Additional context**
* gitlabCI Docker executor

ebuildy:
I have done this image > https://github.com/ebuildy/compose/blob/docker-image-entrypoint-shell/docker-image/entrypoint.sh work pretty well!

ijc:
> use inside CI tools such as GitlabCI that require a shell.

Please can you expand upon this requirement, what exactly does it mean that GitlabCI requires a shell? Do you have a reference for that? (You can assume I know ~0 about gitlabci)

Why would you need to run the shell within the `compose` image, instead of using e.g. `alpine` or something which needs a shell?

Why can you not use `--entrypoint` to override in cases where the default is not what you want?

ebuildy:
Ho you are right, I should explain first what gitlabCI does, according https://docs.gitlab.com/runner/executors/docker.html#the-entrypoint

> The Docker executor doesn’t overwrite the ENTRYPOINT of a Docker image.
> 
> That means that if your image defines the ENTRYPOINT and doesn’t allow to run scripts with CMD, the image will not work with the Docker executor.

As such, to use the ``docker/compose`` image, we need to override the entrypoint. This is not a big problem, but as the official image ``docker`` have another logic of the entrypoint, smarter, I found too bad that ``docker/compose`` do not. This is not consistent, and Docker® should propose official images with same logic.

ijc:
Please can you give an example of what you are trying to do and show why it doesn't work?

[...]

PR #6641, opened 2019-04-10
Refactor Dockerfiles for generating musl binaries

### LINKED ISSUE #6621: Add docker:stable to your build matrix
It seems like installing Compose on docker:stable inside a GitLab CI build is a pretty popular workflow (see #6617). In this issue, @ulyssessouza said:

> This is related to the new SSH support which added some new dependencies. On our test systems these dependencies are already available once pip is installed. https://github.com/docker/compose/issues/6617#issuecomment-477974325

**Should you add this configuration to your build matrix?**

ulyssessouza:
- Refactor `Dockerfile` to be used for tests and distribution on docker hub on `debian` and `alpine`
- Adapt test scripts to the new Dockerfiles' structure
- Adapt Jenkinsfile to add alpine to the test matrix

Signed-off-by: Ulysses Souza <ulysses.souza@docker.com>

Resolves #6621 #6617 #6614 #6106 #6006

Partially: #4287


Dockerfile.alpine
```diff
@@ -0,0 +1,42 @@
...
+    libffi-dev \
+    libgcc \
+    make \
+    musl-dev \
+    openssl \
+    openssl-dev \
+    python2 \
+    python2-dev \
+    zlib-dev && \
+    pip install virtualenv==16.2.0 tox==2.1.1
```
thaJeztah: I'd keep the `pip install` separate; it likely won't overwrite files that were installed by `apk add`, so doing it in a separate step will add a new layer, but won't make the image bigger.

Also worth to keep the same convention as for `apk add`, and split the packages that will be installed to one-per-line, sorted alphabetically (it makes the Dockerfile longer, but can help making it more maintainable)

```suggestion
RUN pip install \
    tox==2.1.1 \
    virtualenv==16.2.0
```

If these packages are expected to be updated individually, could even be two `RUN` lines

Dockerfile.alpine
```diff
@@ -0,0 +1,42 @@
...
+RUN script/build/linux-entrypoint
+
+ADD docker-compose-entrypoint.sh /usr/local/bin/
+COPY --from=docker:18.09.4 /usr/local/bin/docker /usr/local/bin/docker
+ENTRYPOINT ["sh", "/usr/local/bin/docker-compose-entrypoint.sh"]
+
+FROM alpine:3.9 AS runtime
+COPY --from=docker:18.09.4 /usr/local/bin/docker /usr/local/bin/docker
+COPY --from=build /usr/local/bin/docker-compose /usr/local/bin/docker-compose
+ADD docker-compose-entrypoint.sh /usr/local/bin/
```
thaJeztah:
```suggestion
COPY docker-compose-entrypoint.sh /usr/local/bin/
```

Dockerfile.alpine
```diff
@@ -0,0 +1,42 @@
...
+# FIXME(chris-crone): virtualenv 16.3.0 breaks build, force 16.2.0 until fixed
+
+WORKDIR /code/
+
+ADD . /code/
+RUN tox --notest
+RUN script/build/linux-entrypoint
+
+ADD docker-compose-entrypoint.sh /usr/local/bin/
+COPY --from=docker:18.09.4 /usr/local/bin/docker /usr/local/bin/docker
```
thaJeztah: Could consider parameterising this for easier `grep`-ing;

```Dockerfile
ARG DOCKER_VERSION=18.09.4

FROM docker:${DOCKER_VERSION} AS docker-cli
```

And then in this stage;

```Dockerfile
COPY --from=docker-cli /usr/local/bin/docker /usr/local/bin/docker
```

Dockerfile.alpine
```diff
@@ -0,0 +1,42 @@
...
+ENTRYPOINT ["sh", "/usr/local/bin/docker-compose-entrypoint.sh"]
+
+FROM alpine:3.9 AS runtime
+COPY --from=docker:18.09.4 /usr/local/bin/docker /usr/local/bin/docker
+COPY --from=build /usr/local/bin/docker-compose /usr/local/bin/docker-compose
+ADD docker-compose-entrypoint.sh /usr/local/bin/
+ENTRYPOINT ["sh", "/usr/local/bin/docker-compose-entrypoint.sh"]
```
thaJeztah: Tiny optimisation when using "classic" builder; this step can be done _before_ the `COPY` lines (although BuildKit will already be smart enough to cache it)

Dockerfile.alpine
```diff
@@ -0,0 +1,42 @@
+FROM python:3.7.3-alpine3.9 AS build
+
+ARG DOCKER_COMPOSE_GITSHA=unknown
+ENV DOCKER_COMPOSE_GITSHA=$DOCKER_COMPOSE_GITSHA
+
+ENV BUILD_BOOTLOADER=1
```
thaJeztah: These `ARG` and `ENV` lines don't influence the `apk add ...`, and are not needed until further down (or even at runtime?); move them lower in the Dockerfile to prevent busting the cache if not needed

Dockerfile.debian
```diff
@@ -0,0 +1,31 @@
+FROM python:3.7.3-stretch AS build
+ARG DOCKER_COMPOSE_GITSHA=unknown
+ENV DOCKER_COMPOSE_GITSHA=$DOCKER_COMPOSE_GITSHA
```
thaJeztah: Same comment here w.r.t. position of these

Dockerfile.debian
```diff
@@ -0,0 +1,31 @@
...
+    pip install virtualenv==16.2.0 tox==2.1.1
+# FIXME(chris-crone): virtualenv 16.3.0 breaks build, force 16.2.0 until fixed
+
+WORKDIR /code/
+
+ADD . /code/
+RUN tox --notest
+RUN script/build/linux-entrypoint
+
+ADD docker-compose-entrypoint.sh /usr/local/bin/
```
thaJeztah: Should these be in the `build` stage? or only in the `runtime` / `deploy` stage?

Dockerfile.debian
```diff
@@ -0,0 +1,31 @@
...
+    git \
+    libc-dev \
+    libgcc-6-dev \
+    locales \
+    make \
+    openssl && \
+    pip install virtualenv==16.2.0 tox==2.1.1
+# FIXME(chris-crone): virtualenv 16.3.0 breaks build, force 16.2.0 until fixed
+
+WORKDIR /code/
```
thaJeztah: Is there a reason for putting this in a separate `Dockerfile`? Might consider combining them;

```Dockerfile
ARG BUILD_PLATFORM=alpine
ARG DOCKER_VERSION=18.09.4

FROM docker:${DOCKER_VERSION} AS docker-cli
FROM debian:stretch-slim      AS runtime-debian
FROM alpine3.9                AS runtime-alpine

FROM python:3.7.3-alpine3.9   AS build-alpine
RUN apk add --no-cache \
 # etc


FROM python:3.7.3-stretch AS build-debian
ENV BUILD_BOOTLOADER=1
RUN apt-get update && apt-get install -y \
 # etc


# Use either alpine or debian for the build
FROM build-${BUILD_PLATFORM} AS build
WORKDIR /code/

ARG GIT_COMMIT=unknown
ENV DOCKER_COMPOSE_GITSHA=${GIT_COMMIT}
COPY . .
RUN tox --notest
RUN script/build/linux-entrypoint

# Use either alpine or debian for the deploy/runtime
FROM runtime-${BUILD_PLATFORM} AS runtime
ENTRYPOINT ["sh", "/usr/local/bin/docker-compose-entrypoint.sh"]
COPY --from=docker-cli /usr/local/bin/docker          /usr/local/bin/docker
COPY --from=build      /usr/local/bin/docker-compose /usr/local/bin/docker-compose
COPY docker-compose-entrypoint.sh /usr/local/bin/
```


Then;

```bash
EXPORT BUILD_PLATFORM=alpine
EXPORT GIT_COMMIT =$(git rev-parse --verify HEAD) 

docker build \
	-t docker-compose:${BUILD_PLATFORM} \
	--build-arg BUILD_PLATFORM \
	--build-arg GIT_COMMIT \
	.

EXPORT BUILD_PLATFORM=debian
docker build \
	-t docker-compose:${BUILD_PLATFORM} \
	--build-arg BUILD_PLATFORM \
	--build-arg GIT_COMMIT \
	.
```

Dockerfile.alpine
```diff
@@ -0,0 +1,42 @@
+FROM python:3.7.3-alpine3.9 AS build
+
+ARG DOCKER_COMPOSE_GITSHA=unknown
+ENV DOCKER_COMPOSE_GITSHA=$DOCKER_COMPOSE_GITSHA
```
thaJeztah: It's often a good idea to use a different name for the `ARG` than for the `ENV`. Both act in the same "space" (both are set as (environment) variables, which can lead to run situations where the `ENV` is always overridden by the `ARG` (I can find some examples, don't have them at hand)

So, might want to consider, e.g.;

```Dockerfile
ARG GIT_COMMIT=unknown
ENV DOCKER_COMPOSE_GITSHA=$GIT_COMMIT
```

(or vice-versa)

Dockerfile.alpine
```diff
@@ -0,0 +1,42 @@
...
+    openssl-dev \
+    python2 \
+    python2-dev \
+    zlib-dev && \
+    pip install virtualenv==16.2.0 tox==2.1.1
+# FIXME(chris-crone): virtualenv 16.3.0 breaks build, force 16.2.0 until fixed
+
+WORKDIR /code/
+
+ADD . /code/
```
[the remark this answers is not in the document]
thaJeztah: So `tox` installs dependencies, correct? (perhaps also good to add as an example to https://github.com/moby/buildkit/blob/master/frontend/dockerfile/docs/experimental.md#example-cache-go-packages)


What does it need to know what to install? `requirements.txt`? (or one of the other `requirements-xx` files?)

If so; consider;

```Dockerfile
# Install requirements
COPY requirements.txt .
RUN tox --notest

COPY . .
ARG GIT_COMMIT=unknown
ENV DOCKER_COMPOSE_GITSHA=${GIT_COMMIT}
RUN script/build/linux-entrypoint
```

That way:

```Dockerfile
COPY requirements.txt .
```

Will _only_ happen if `requirements.txt` changes, and

```Dockerfile
RUN tox --notest
```

Will _only_ be executed in that case as well, but not if you change anything else in your codebase

thaJeztah:
Here's an old PR I created with changes; some of those changes (combining the `COPY` steps) should not be done (bad idea at the time); but perhaps there's useful information in that PR as well (check the individual commits https://github.com/docker/compose/pull/6061)

Dockerfile.debian
ulyssessouza: Wow! That's exactly what I was thinking to do, my only question is if that would build both everytime or just the necessary stages for the final target.

thaJeztah: When using buildkit, only the stages that you need will get built. When using the classic build, everything will be built (all the more reasons to use buildkit, as it's awesome!)

script/build/linux-entrypoint
```diff
@@ -2,14 +2,35 @@
...
+VENV=${CODE_PATH}/.tox/py37
 
-mkdir -p `pwd`/dist
-chmod 777 `pwd`/dist
+cd ${CODE_PATH}
+mkdir -p dist
+chmod 777 dist
 
-$VENV/bin/pip install -q -r requirements-build.txt
+${VENV}/bin/pip3 install -q -r requirements-build.txt
```
joakimr-axis: There is no need to add these extra {} here. It just makes the code more cluttered and leads the code reader into thinking there is something weird going on that needs to be handled.  Had the variable name been e.g. V-ENV, you would have needed to state ${V-ENV}, but for a perfectly simple name like VENV, keep things simple.

script/build/linux-entrypoint
```diff
@@ -2,14 +2,35 @@
...
+
+# TODO(ulyssessouza) To check if really needed
 ./script/build/write-git-sha
-su -c "$VENV/bin/pyinstaller docker-compose.spec" user
-mv dist/docker-compose $TARGET
-$TARGET version
+
+export PATH="${CODE_PATH}/pyinstaller:${PATH}"
```
joakimr-axis: See previous comment regarding no need for {} here

script/build/linux-entrypoint
```diff
@@ -2,14 +2,35 @@
...
+
+if [ ! -z "${BUILD_BOOTLOADER}" ]; then
```
joakimr-axis: Same here

script/test/default
```diff
@@ -3,17 +3,18 @@
...
-DOCKERFILE="${DOCKERFILE:-Dockerfile}"
+DOCKERFILE="${DOCKERFILE:-Dockerfile.alpine}"
+DOCKER_BUILD_TARGET="${DOCKER_BUILD_TARGET:-build}"
 
 rm -rf coverage-html
 # Create the host directory so it's owned by $USER
 mkdir -p coverage-html
 
-docker build -f ${DOCKERFILE} -t "$TAG" .
+docker build -f ${DOCKERFILE} -t "${TAG}" --target "${DOCKER_BUILD_TARGET}" .
```
joakimr-axis: It would be cleaner to use

docker build -f $DOCKERFILE -t "$TAG" --target "$DOCKER_BUILD_TARGET"

here. Even though I cannot explicitly find that space is forbidden in docker tags (https://docs.docker.com/engine/reference/commandline/tag/), is it really allowed? Otherwise, using $TAG instead of "$TAG" would be preferred.

docker-compose-entrypoint.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/sh
+set -e
+
+# first arg is `-f` or `--some-option`
+if [ "${1#-}" != "$1" ]; then
+	set -- docker-compose "$@"
+fi
```
joakimr-axis: You could avoid doing the extra call to if by writing this as a one-liner:

[ "${1#-}" == "$1" ] || set -- docker-compose "$@"

script/build/linux-entrypoint
joakimr-axis: export PATH="$CODE_PATH/pyinstaller:$PATH"

is just fine here and there is no need for {} with those variable names.

thaJeztah: I think it's generally good practice to be consistent and use `{}` (but there might be two camps on that)

joakimr-axis: Yes, there is (two camps). :-)
When I see a variable like that and there is no obvious thing in the variable name that calls for having the {}, I think ah, there must be some substitution or such going on and then where it isn't I see that oh, that was just somebody using {} when it was not needed. In simple use lite $PATH, one immediately recognizes that there is a regular PATH and moves on directly.

Dockerfile.alpine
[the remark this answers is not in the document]
ulyssessouza: Sure! The point is that for now, I don't know exactly which files are needed on each step anymore.
I'm planning to refactor this part by the end of the process or in follow up.
But I'm completely aware.

Dockerfile.alpine
ulyssessouza: Thanks for that one!

Dockerfile.alpine
ulyssessouza: Done

Dockerfile.alpine
ulyssessouza: Done

Dockerfile.alpine
ulyssessouza: Done

Dockerfile.debian
ulyssessouza: Done

Dockerfile.debian
ulyssessouza: For the build too, since it's used for tests.
Actually the runtime is the one that goes to the hub and the build is the one that will be used by the CI for the tests

Dockerfile.alpine
ulyssessouza: Done

script/build/linux-entrypoint
ulyssessouza: IMO the use of {} places delimiters on when the variable name ends.
As you just said, there are cases where it's necessary, so to make the code style consistent, I prefer using it. Even my IDE asks for it. And by the end of the day that's a question of taste.

script/build/linux-entrypoint
ulyssessouza: https://github.com/docker/compose/pull/6641#discussion_r274918137

script/build/linux-entrypoint
ulyssessouza: https://github.com/docker/compose/pull/6641#discussion_r274918137

script/test/default
ulyssessouza: Here I just use the double quotes because the line before was using it.

docker-compose-entrypoint.sh
ulyssessouza: IMHO, your suggestion lacks readability, which is important in docker-compose.

And the second point on that, is that actually this code is just a copy from https://github.com/docker-library/docker/blob/69ab1f338fbbdc57fa6dae39308bcd2fc98560a8/18.09/docker-entrypoint.sh#L5

script/build/linux-entrypoint
ulyssessouza: https://github.com/docker/compose/pull/6641#discussion_r274918137

Dockerfile.alpine
ulyssessouza: In this case, I think that what we win is not worth losing the readability of having the `ENTRYPOINT` as last line of a target. 
@thaJeztah WDYT?

Dockerfile
```diff
@@ -1,36 +1,68 @@
...
+    python2.7-dev
 
-RUN useradd -d /home/user -m -s /bin/bash user
+FROM build-${BUILD_PLATFORM} AS build
 WORKDIR /code/
-
 # FIXME(chris-crone): virtualenv 16.3.0 breaks build, force 16.2.0 until fixed
-RUN pip install virtualenv==16.2.0
-RUN pip install tox==2.9.1
+RUN pip install virtualenv==16.2.0 tox==2.1.1
```
thaJeztah: I suspect it's ok to keep these separate so that they can be cached if one is updated (as long as files installed by both don't overlap, it won't result in smaller images if combined

Dockerfile
```diff
@@ -1,36 +1,68 @@
...
 
-ADD requirements.txt /code/
-ADD requirements-dev.txt /code/
-ADD .pre-commit-config.yaml /code/
-ADD setup.py /code/
-ADD tox.ini /code/
-ADD compose /code/compose/
-ADD README.md /code/
+ARG GIT_COMMIT=unknown
+ENV DOCKER_COMPOSE_GITSHA=$GIT_COMMIT
```
thaJeztah: I _think_ this `ENV` is only used by `script/build/linux-entrypoint`, so better to move it lower to prevent unwanted cache-busts

Dockerfile
```diff
@@ -1,36 +1,68 @@
...
+COPY requirements.txt \
```
thaJeztah: Same as for the `pip install`; keeping these separate may actually be better (but give it a try), so that caching is optimized. The files don't overlap, so combining `COPY` lines won't make the image smaller. It would reduce the number of layers, but this is an intermediate stage, so won't affect the final image

Dockerfile
```diff
@@ -1,36 +1,68 @@
...
+     requirements-dev.txt \
+     .pre-commit-config.yaml \
+     tox.ini \
+     setup.py \
+     README.md \
```
thaJeztah: Does `tox` need the `README.md`? I suspect this can be removed (it will be copied in the `COPY . . ` below

Dockerfile
ulyssessouza: Done

Dockerfile
ulyssessouza: Done

Dockerfile
ulyssessouza: Done

Dockerfile
ulyssessouza: Actually it does. I thought the same before. `tox` executes `python setup.py egg_info` which will check for the `README.md` informed in https://github.com/docker/compose/blob/master/setup.py#L80

Dockerfile
```diff
@@ -1,36 +1,70 @@
...
-RUN pip install tox==2.9.1
-
-ADD requirements.txt /code/
-ADD requirements-dev.txt /code/
-ADD .pre-commit-config.yaml /code/
-ADD setup.py /code/
-ADD tox.ini /code/
-ADD compose /code/compose/
-ADD README.md /code/
+RUN pip install tox==2.1.1
```
thaJeztah: This version changed from `2.9.1` to `2.1.1`, correct?

Dockerfile
```diff
@@ -1,36 +1,70 @@
...
+COPY tox.ini .
+COPY setup.py .
+COPY README.md .
+COPY compose compose/
 RUN tox --notest
+COPY . .
+ARG GIT_COMMIT=unknown
+ENV DOCKER_COMPOSE_GITSHA=$GIT_COMMIT
+RUN script/build/linux-entrypoint
+COPY docker-compose-entrypoint.sh /usr/local/bin/
```
thaJeztah: very minor nit; this file likely doesn't change much, so could be moved before the `COPY . . `, or even before the `COPY requirements.txt` (ordering from "less frequently changing" to "most frequently changing").

If this script (and `ENTRYPOINT` below is only used for `runtime`, might consider moving both to the start of the `runtime` stage)

Dockerfile
```diff
@@ -1,36 +1,70 @@
...
+
+ENTRYPOINT ["sh", "/usr/local/bin/docker-compose-entrypoint.sh"]
```
thaJeztah: See above regarding if this should be in the `build` stage at all, or only in `runtime`

Dockerfile
```diff
@@ -1,36 +1,70 @@
...
 
-ADD . /code/
-RUN chown -R user /code/
+FROM alpine:${ALPINE_VERSION} AS runtime-alpine
+FROM debian:${DEBIAN_VERSION} AS runtime-debian
+FROM runtime-${BUILD_PLATFORM} AS runtime
+COPY --from=docker-cli  /usr/local/bin/docker           /usr/local/bin/docker
+COPY --from=build       /usr/local/bin/docker-compose   /usr/local/bin/docker-compose
+COPY docker-compose-entrypoint.sh /usr/local/bin/
```
thaJeztah: Move this (end the `ENTRYPOINT`) before the `COPY --from` lines; I expect this file to rarely change, so no need to copy it each time the `build` changes.

Dockerfile
```diff
@@ -1,36 +1,70 @@
...
 
-ENTRYPOINT ["/code/.tox/py37/bin/docker-compose"]
+ENTRYPOINT ["sh", "/usr/local/bin/docker-compose-entrypoint.sh"]
```
thaJeztah: was the `sh` needed? If the `docker-compose-entrypoint.sh` has its executable bit set (`chmod +x`), won't it already start `/bin/sh` ?

script/test/default
thaJeztah: `-f` specifies a path to the Docekrfile, so good practice to put quotes around it (even though that will likely not be hit here

```suggestion
docker build -f "${DOCKERFILE}" -t "${TAG}" --target "${DOCKER_BUILD_TARGET}" .
```

script/build/test-image
```diff
@@ -7,11 +7,12 @@ if [ -z "$1" ]; then
...
-ctnr_id=$(docker create --entrypoint=tox docker-compose-tests:tmp)
-docker commit $ctnr_id docker/compose-tests:latest
-docker tag docker/compose-tests:latest docker/compose-tests:$TAG
-docker rm -f $ctnr_id
-docker rmi -f docker-compose-tests:tmp
+DOCKER_COMPOSE_GITSHA=$(script/build/write-git-sha)
+docker build -t "${IMAGE}:${TAG}" . \
+       --target build \
+       --build-arg BUILD_PLATFORM=debian \
+       --build-arg GIT_COMMIT=${DOCKER_COMPOSE_GITSHA}
```
thaJeztah:
```suggestion
       --build-arg GIT_COMMIT="${DOCKER_COMPOSE_GITSHA}"
```

script/build/linux-entrypoint
```diff
@@ -2,14 +2,36 @@
...
+    git checkout v3.4
+    ${VENV}/bin/python3 ./waf configure --no-lsb all
+    ${VENV}/bin/pip3 install ..
+    cd ${CODE_PATH}
+    rm -Rf /tmp/pyinstaller
+else
+    echo "NOT compiling bootloader!!!"
+fi
+
+${VENV}/bin/pyinstaller --exclude-module pycrypto --exclude-module PyInstaller docker-compose.spec
```
thaJeztah: might need quotes here

script/build/linux-entrypoint
```diff
@@ -2,14 +2,36 @@
...
+
+export PATH="${CODE_PATH}/pyinstaller:${PATH}"
+
+if [ ! -z "${BUILD_BOOTLOADER}" ]; then
+    # Build bootloader for alpine
+    git clone --single-branch --branch master https://github.com/pyinstaller/pyinstaller.git /tmp/pyinstaller
+    cd /tmp/pyinstaller/bootloader
+    git checkout v3.4
+    ${VENV}/bin/python3 ./waf configure --no-lsb all
+    ${VENV}/bin/pip3 install ..
```
thaJeztah: might need quotes here

script/build/image
```diff
@@ -11,7 +11,10 @@ TAG=$1
...
 
-./script/build/write-git-sha
+DOCKER_COMPOSE_GITSHA=$(script/build/write-git-sha)
+echo ${DOCKER_COMPOSE_GITSHA} > compose/GITSHA
 python setup.py sdist bdist_wheel
-./script/build/linux
-docker build -t docker/compose:$TAG -f Dockerfile.run .
+
+docker build \
+    --build-arg GIT_COMMIT=${DOCKER_COMPOSE_GITSHA} \
```
thaJeztah:
```suggestion
    --build-arg GIT_COMMIT="${DOCKER_COMPOSE_GITSHA}" \
```

Dockerfile
ulyssessouza: Ops! My bad... Maybe during the squash. Good catch.

Dockerfile
ulyssessouza: Done

Dockerfile
ulyssessouza: Done

Dockerfile
ulyssessouza: Yep. I tried that, but it doens't work :(

script/build/linux-entrypoint
ulyssessouza: Done

script/build/linux-entrypoint
ulyssessouza: Done

script/build/image
ulyssessouza: Done

script/build/test-image
ulyssessouza: Done

script/test/default
ulyssessouza: Done

Dockerfile
ulyssessouza: Done

Dockerfile
```diff
@@ -1,36 +1,68 @@
...
+    libc-dev \
+    libgcc-6-dev \
+    make \
+    openssl \
+    python2.7-dev
 
-RUN useradd -d /home/user -m -s /bin/bash user
+FROM build-${BUILD_PLATFORM} AS build
+ENTRYPOINT ["sh", "/usr/local/bin/docker-compose-entrypoint.sh"]
+COPY --from=docker-cli /usr/local/bin/docker /usr/local/bin/docker
```
thaJeztah: I think you moved the wrong line; should this have been

```
COPY docker-compose-entrypoint.sh /usr/local/bin/
```

(and still wondering if this is needed in the _build_ stage? if not, just remove these lines)

ulyssessouza: Actually that's not a move, but a new `COPY` needed when running the tests.

script/release/release/images.py
```diff
@@ -42,33 +47,46 @@ def build_images(self, repository, files):
...
     @property
     def image_names(self):
-        return [
+        images = [
             'docker/compose-tests:latest',
             'docker/compose-tests:{}'.format(self.version),
             'docker/compose:{}'.format(self.version)
         ]
+        if self.latest:
+            images.append('docker/compose:latest')
```
chris-crone: As part of this, we should add Debian tags for images. Taking the upcoming 1.25.0 release as an example, we should have:
```
latest [Alpine based 1.25.0]
debian [Debian based 1.25.0]
1.25.0, 1.25.0-alpine [Alpine based 1.25.0]
1.25.0-debian [Debian based 1.25.0]
```

I think that's the minimum level of granularity that we can go with for now. If there's demand, we can add more explicit tags like: `1.25.0-alpine3.9.3` or `1.25.0-stretch-20190326-slim`.

I hesitate doing this now as it adds something else that we'll need to maintain so would prefer some user demand first.

WDYT @thaJeztah?

Dockerfile
thaJeztah: Ah! Gotcha; wasn't aware the tests now needed the docker cli 👍

script/release/release/images.py
thaJeztah: Makes sense; perhaps we should follow the official images tagging scheme (taking  https://hub.docker.com/_/golang/ as example);

Alpine based:
 - `1.25.0`, `1.25`, `1`, `latest`
 - `1.25.0-alpine3.9`, `1.25-alpine3.9`, `1-alpine3.9`, `alpine3.9`
 - `1.25.0-alpine`, `1.25-alpine`, `1-alpine`, `alpine`

Debian based:
- `1.25.0-stretch`, `1.25-stretch`, `1-stretch`, `stretch`


Note:

- Golang has the "default" reversed (`latest` refers to Debian based); we might have to decide if we want to follow that pattern, or be explicit about `alpine` being the preferred
- Will there be Windows based variants? In that case, using `latest` (and tags without `-distroX`) should likely refer to Windowsservercore


Calling @tianon for advice 🤗

thaJeztah: FWIW, looking at the Golang repo; I wondered why `stretch` was picked, and not `debian-stretch` (so `debian-stretch` -> `debian` (more specific -> less specific), which would be in line with `alpine3.9` -> `alpine` (more specific -> less specific))

script/build/linux
```diff
@@ -4,10 +4,15 @@ set -ex
...
 ./script/clean
 
-TAG="docker-compose"
-docker build -t "$TAG" .
-docker run \
-    --rm --entrypoint="script/build/linux-entrypoint" \
-    -v $(pwd)/dist:/code/dist \
-    -v $(pwd)/.git:/code/.git \
-    "$TAG"
+TMP_CONTAINER="tmpcontainer"
```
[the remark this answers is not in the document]
ulyssessouza: Done

script/build/linux
```diff
@@ -4,10 +4,15 @@ set -ex
...
+TAG="docker/compose:tmp-glibc-linux-binary"
```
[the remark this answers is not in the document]
ulyssessouza: Done

script/release/release/images.py
ulyssessouza: What I just implemented is...
Given `1.25.0` as an example version, the following tags are produced:
For Alpine: `1.25.0-alpine`, `1.25.0`, `alpine`, `latest`
For Debian: `1.25.0-debian`, `debian`

Knowing that the following tags will roll when releasing the new version:
`alpine`, `latest`, `debian`

I suggest implementing the intermediate tags (like `1-alpine` or `1.25-debian`) in another PR, since it looks more like a new feature than something related to this one.

script/build/linux
```diff
@@ -4,10 +4,15 @@ set -ex
...
-    -v $(pwd)/.git:/code/.git \
-    "$TAG"
+DOCKER_COMPOSE_GITSHA="$(script/build/write-git-sha)"
+TMP_CONTAINER="tmpcontainer-${DOCKER_COMPOSE_GITSHA}"
+TAG="docker/compose:tmp-glibc-linux-binary-${DOCKER_COMPOSE_GITSHA}"
+
+docker build -t "${TAG}" . \
+       --build-arg BUILD_PLATFORM=debian \
+       --build-arg GIT_COMMIT="${DOCKER_COMPOSE_GITSHA}"
+docker create --name "${TMP_CONTAINER}" "${TAG}"
```
thaJeztah: alternatively; don't give it a name, and capture the container ID (which is printed when creating the container)

script/test/default
```diff
@@ -3,17 +3,18 @@
 
 set -ex
 
-TAG="docker-compose:$(git rev-parse --short HEAD)"
+TAG="docker-compose:alpine-$(git rev-parse --short HEAD)"
 
-# By default use the Dockerfile, but can be overridden to use an alternative file
+# By default use the Dockerfile.alpine, but can be overridden to use an alternative file
```
chris-crone: The default below is `Dockerfile`, I think you missed this when putting the Alpine and Debian builds into the same Dockerfile.

script/release/release/images.py
tianon: > FWIW, looking at the Golang repo; I wondered why `stretch` was picked, and not `debian-stretch`, so
> 
> * `debian-stretch` -> `debian` (more specific -> less specific), which would be in line with;
> * `alpine3.9` -> `alpine` (more specific -> less specific)

So, there's some history here.  The `stretch` alias was added because we needed a way to help users transition off Jessie, so we tagged with suite and had both for a bit.  Then Alpine had the same issue, so we added similar Alpine tags.  For repositories where the suite isn't as important (like `redis`), I believe we went with simply `debian` (my top-of-the-head example here might be wrong, I'm writing this from my phone, but there's at least one that does that).

The reason Debian is typically the default is because many upstreams don't officially support Alpine and/or musl (Go and Python being very notable examples; Go has upstream tests that fail hard on Alpine with nobody working on fixing them and Python doesn't seem to care much that Alpine even exists, so functionality breaks or acts funny in some cases).

So, all that is really to say I'd recommend proceeding with an Alpine default with extreme caution for something based on Python unless there's a really comprehensive integration test to verify the result. 👍

thaJeztah: Thanks Tianon!

script/test/default
ulyssessouza: Done

script/build/linux
ulyssessouza: Done

thaJeztah:
Some failures, but not sure if they're expected (also some tests look to be terminated)

```
14:09:22 + docker pull dockerbuildbot/compose:alpine-a309ebbce61484db33c925568805f8fb3794aad6
14:09:23 Error response from daemon: manifest for dockerbuildbot/compose:alpine-a309ebbce61484db33c925568805f8fb3794aad6 not found

14:16:47 + docker pull dockerbuildbot/compose:debian-a309ebbce61484db33c925568805f8fb3794aad6
14:16:48 Error response from daemon: manifest for dockerbuildbot/compose:debian-a309ebbce61484db33c925568805f8fb3794aad6 not found
```

Dockerfile
```diff
@@ -1,36 +1,71 @@
...
+    make \
+    openssl \
+    python2.7-dev
 
+FROM build-${BUILD_PLATFORM} AS build
+COPY docker-compose-entrypoint.sh /usr/local/bin/
+ENTRYPOINT ["sh", "/usr/local/bin/docker-compose-entrypoint.sh"]
+COPY --from=docker-cli /usr/local/bin/docker /usr/local/bin/docker
+WORKDIR /code/
 # FIXME(chris-crone): virtualenv 16.3.0 breaks build, force 16.2.0 until fixed
```
rumpl: Did we try with `16.4.x`?

chris-crone: I think that I did locally and it worked. I just didn't get a chance to put up a PR.

We only use virtualenv to build (we don't ship it) so we actually don't need to use the latest version. We should pin it though so that we don't suddenly get broken builds.

thaJeztah:
🎉

nvllsvm:
@ulyssessouza Any chance the `latest` tag could be manually pushed for `1.24.0`?

ulyssessouza:
@nvllsvm I don't think so. That would be better for us to have the tagging process to be done by the release script to avoid any kind of human interference. And makes more sense since the feature will be applied with the new release.

nvllsvm:
@ulyssessouza I see that 1.24.1 was just released, but no latest tag was pushed. Is this expected?

chris-crone:
@nvllsvm We need to think a little about what we tag as latest. We currently have a Debian (i.e.: built with glibc) based build and an Alpine one (i.e.: build with musl).

It might make sense to have a `debian` tag and an `alpine` one.

andyneff:
@ulyssessouza Does this mean that the new 1.25 RC binaries should work in an alipine (musl) docker without having glibc installed?

ulyssessouza:
@andyneff This means that we will have 2 docker images. As @chris-crone said, they will be tagged with it's respective base platform. Meaning `glibc` for `debian` and `libmusl` for `alpine`.

andyneff:
Ah, I found it. https://hub.docker.com/r/docker/compose/tags Thanks!

andyneff:
Looks like the `docker-compose` binary won't work in alpine 3.8 (but works in alpine 3.9 no problem) because of the same issue mentioned by @stieg in [#6292](https://github.com/docker/compose/issues/6292#issuecomment-445374792), and I can't get the image to build against alpine 3.8 for other reasons. Will alpine 3.8 ever be supported? (I think 3.7 is technically EOL?)

```Dockerfile
FROM docker/compose:1.25.0-rc4 as docker_compose
FROM alpine:3.8
COPY --from docker_compose /usr/local/bin/docker-compose /usr/local/bin/docker-compose
RUN /usr/local/bin/docker-compose
```

```
[6] Error loading Python lib '/tmp/_MEIceKiFf/libpython3.7m.so.1.0': dlopen: Error relocating /tmp/_MEIceKiFf/libpython3.7m.so.1.0: getrandom: symbol not found
```

PR #6647, opened 2019-04-15
Allow exec on one-off containers created with run

### LINKED ISSUE #6549: docker-compose exec work on "run" containers
## Example use case

`docker-compose.yml`

```
version: "3.2"
services:
  foo:
    image: debian:9
    command: sleep 10000
```

1. `docker-compose run -d foo`
2. `docker-compose exec foo bash`

### Current Result:

```
ERROR: No container found for foo_1
```

### Desired result

It would be nice if not only did `docker-compose exec` check for containers created by `up`, but also `run`. This is already done for commands like `down` and `ps`

**Example:** `docker-compose ps`

```
          Name              Command     State   Ports
-----------------------------------------------------
foo_foo_run_382b47cedaf4   sleep 10000  Up 
```

### Alternative solutions

If for some reason mixing the `up`ed and `run`ed containers together in `exec` is incompatible with use cases/maintaining backwards compatibility, then specifying a flag that either "Only execs using `run` containers" or combines the list of `up` and `exec` containers. Which ever you think is best if this is needed.

### Current work around

I currently have to have to use a mix of `bash`, `docker-compose ps`, or `docker ps` commands to find the container I want to exec into

collin5:
Resolves #6549

### DIFF
--- a/compose/service.py
+++ b/compose/service.py
@@ -234,7 +234,12 @@ class Service(object):
         """Return a :class:`compose.container.Container` for this service. The
         container must be active, and match `number`.
         """
-        for container in self.containers(labels=['{0}={1}'.format(LABEL_CONTAINER_NUMBER, number)]):
+        from compose.project import OneOffFilter  # noqa
+
+        labels = ['{0}={1}'.format(LABEL_CONTAINER_NUMBER, number)]
+        containers = self.containers(labels=labels) + \
+            self.containers(labels=labels, one_off=OneOffFilter.only)
+        for container in containers:
             return container
 
         raise ValueError("No container found for %s_%s" % (self.name, number))
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -891,6 +891,8 @@ class TopLevelCommand(object):
             command = service.options.get('command')
 
         options['stdin_open'] = service.options.get('stdin_open', True)
+        options['--label'] = options.get(
+            '--label', []) + ['%s=%s' % (LABEL_CONTAINER_NUMBER, len(service.containers()) + 1)]
 
         container_options = build_one_off_container_options(options, detach, command)
         run_one_off_container(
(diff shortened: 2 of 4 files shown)


ijc:
Thank you @collin5 for this PR.

Could you explain a bit more the reasoning behind the changes here. I see you are adding a new label to these one off containers, which suggests this is needed for some reason but I'm unclear what that is. If it is just to mirror the change in `get_container` to filter on that label for oneoff containers? In which case why is the label needed, is it not possible to filter for all the oneoffs for some reason?

Also, there is a `noqa` comment in there -- why?

collin5:
Hi @ijc, sorry for the delay on this. I'm adding the label to one-off containers because [get_container](https://github.com/docker/compose/blob/26e1a2dd31b41ff3bc8418de879a8a839a199548/compose/service.py#L226) expects a container `index` with the default being set to `1`. We are currently adding this label for all the containers created with `docker-compose up` here https://github.com/docker/compose/blob/26e1a2dd31b41ff3bc8418de879a8a839a199548/compose/service.py#L891-L897 This was not the case for the one-off containers created with `run` which is part or the reasons we couldn't retrieve them here https://github.com/docker/compose/blob/26e1a2dd31b41ff3bc8418de879a8a839a199548/compose/service.py#L230

Also, `noqa` is for the flake8 to ignore the module level import not being at top of file. Putting this at top introduces circular import errors so I thought this was a better approach to avoid refactoring files not related to this issue.

mafrosis:
@collin5 @ijc Any progress on this? It's bugged me for a long time :D

compose/service.py
```diff
@@ -234,7 +234,12 @@ def get_container(self, number=1):
         """Return a :class:`compose.container.Container` for this service. The
         container must be active, and match `number`.
         """
-        for container in self.containers(labels=['{0}={1}'.format(LABEL_CONTAINER_NUMBER, number)]):
+        from compose.project import OneOffFilter  # noqa
```
kingbuzzman, line 237: This feels like an avoidable hack, can we move `OneOffFilter` into a `enums.py` (or equivalent)? Then we can import it from both `service.py` and `project.py` circumventing this import.

glours:

PR #6649, opened 2019-04-15
support requests up to 2.22.0 version

GeyseR:
Signed-off-by: Sergey Fursov <geyser85@gmail.com>

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -17,7 +17,7 @@ pypiwin32==219; sys_platform == 'win32' and python_version < '3.6'
 pypiwin32==223; sys_platform == 'win32' and python_version >= '3.6'
 PySocks==1.6.7
 PyYAML==4.2b1
-requests==2.20.0
+requests==2.22.0
 six==1.10.0
 texttable==0.9.1
 urllib3==1.24.2; python_version == '3.3'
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -9,15 +9,15 @@ dockerpty==0.4.1
 docopt==0.6.2
 enum34==1.1.6; python_version < '3.4'
 functools32==3.2.3.post2; python_version < '3.2'
-idna==2.5
```
ulyssessouza: Why bumping `idna`?

ijc:
@GeyseR does the bump of idna relate to bumping requests somehow or is this an unrelated change?

GeyseR:
Hm, I don't think it's a required change for the upgrade. I'll rollback idna lib and update requests to 2.22.0

ijc:
Thanks! If we want to bump idna we should at least do it in a separate commit, if not a separate PR.

GeyseR:
Ready

GeyseR:
> I have zero clue what that has to do with your change
same for me. maybe Jenkins is not working with the rewtitten history of the branch? What do you think about submitting a new PR to check this assumption?

ijc:
@GeyseR I kicked it to have another go, but it failed again.

I don't think we need a fresh PR in the first instance, but would certainly be worth rebasing this one onto the current `master` and force pushing, maybe that will unconfuse something...

GeyseR:
hey @ijc, I've just done a rebase and force push, and it seems that the new Jenkins build has passed the previous fail point

ijc:
CI is green, thanks @GeyseR for your contribution → merging.

PR #6664, opened 2019-04-23
Dockerfile: Decrease amount of layers

### LINKED ISSUE #6663: Dockerfile: Group some RUN and COPY commands to decrease amount of layers
Some RUN and COPY commands in the Dockerfile could be grouped in order to decrease the amount of layers used.

joakimr-axis:
Grouping some RUN and COPY commands.

Change-Id: I94f0c146248a54b86b49c3c13cbcf9ee41d16418

Resolves #6663


joakimr-axis:
See Ulysses' comment in #6663.

Issue #5010, opened 2017-07-09
Implement dns_opt in version 3


ylavoie:
Even with version 3, dns_opt is still required to pass specific options like `ndots` and get around dns problems

shin-:
Please submit v3 feature requests on the [docker/cli](https://github.com/docker/cli) repo.

maciejmatuszak:
Hi @shin- the docker cli already implements the --dns-opt or --dns-option. The docker-compose dns-opt is in V2 https://docs.docker.com/compose/compose-file/compose-file-v2/#dns_opt but is not V3. What would it take to implement/port it from V2 to V3?

shin-:
The v3 format is controlled and updated on the `docker/cli` repo (as the primary intention of the v3 format is to work with `docker stack deploy`), so it needs to be proposed there first before we can port the change back into the `docker-compose` codebase. For the change to be accepted in `docker/cli`, it would probably require an implementation of the feature for Swarm services (which [seems to exist at the API level](https://docs.docker.com/engine/api/v1.39/#operation/ServiceCreate), so shouldn't be too daunting to add).

ps-jay:
Hi @shin-,

I'm trying to understand what you mean by the "v3 format is controlled and updated on the `docker/cli` repo".

What file(s), should I look at in `docker/cli` to see what is (or should be) supported by compose?

From what I can see, `--dns-option` is already in the `docker run` CLI (see: https://github.com/docker/cli/commit/ebacbb50aec7281467daf59980ccb6d81724c050), but it's not available in compose.

Should we raise an issue in `docker/cli` that `dns_option` is not also available in the v3 format?

shin-:
> Hi @shin-,
> 
> I'm trying to understand what you mean by the "v3 format is controlled and updated on the `docker/cli` repo".
> 
> What file(s), should I look at in `docker/cli` to see what is (or should be) supported by compose?

https://github.com/docker/cli/tree/master/cli/compose/schema/data

> 
> From what I can see, `--dns-option` is already in the `docker run` CLI (see: [docker/cli@ebacbb5](https://github.com/docker/cli/commit/ebacbb50aec7281467daf59980ccb6d81724c050)), but it's not available in compose.
> 
> Should we raise an issue in `docker/cli` that `dns_option` is not also available in the v3 format?

Yes.

ps-jay:
Thanks @shin-,
I've opened an issue in [docker/cli](https://github.com/docker/cli).

[...]

PR #6642, opened 2019-04-11
[WIP] Sync compose-schemas with upstream (docker/cli)

thaJeztah:
I wanted to update the compose-schemas to make sure they're in sync with upstream, but noticed there's changes in the schema in this repository that are not upstream.

Also; some of these patches were applied to older version, but are not in the latest (3.7) and upcoming (3.8) version.

- https://github.com/docker/compose/pull/5364 Implement subnet config validation (fixes #4552) 
- https://github.com/docker/compose/pull/5585 Labels validation and basic type conversion (fixes https://github.com/docker/compose/issues/4904)

We should make sure that both `docker stack deploy` and `docker-compose up` validation is equal, otherwise files may work for one, but won't work for the other.

If there's differences because of the implementation (Python vs Go), we should look if we can still use the same schema (if needed, upstream changes to docker/cli).

I was looking at reproducing the original issue () earlier this week (but not sure where I ended up, as I had to finish some other things 😂); I see I have these files still on my machine, which was what I was testing with;

I think I was testing if the problem presented itself on compose-file schema `3.6` and `3.7` as it doesn't appear to have the fix from  https://github.com/docker/compose/issues/4904, thus "should be failing"; here's the diff between 3.6 and 3.7 (before this patch);

```bash
git diff --no-index config_schema_v3.5.json config_schema_v3.6.json
```

<details>
<summary>output:</summary>

```patch
diff --git a/config_schema_v3.5.json b/config_schema_v3.6.json
index e3bdecbc..95a552b3 100644
--- a/config_schema_v3.5.json
+++ b/config_schema_v3.6.json
@@ -1,6 +1,6 @@
 {
   "$schema": "http://json-schema.org/draft-04/schema#",
-  "id": "config_schema_v3.5.json",
+  "id": "config_schema_v3.6.json",
   "type": "object",
   "required": ["version"],
 
@@ -84,7 +84,7 @@
                 "context": {"type": "string"},
                 "dockerfile": {"type": "string"},
                 "args": {"$ref": "#/definitions/list_or_dict"},
-                "labels": {"$ref": "#/definitions/labels"},
+                "labels": {"$ref": "#/definitions/list_or_dict"},
                 "cache_from": {"$ref": "#/definitions/list_of_strings"},
                 "network": {"type": "string"},
                 "target": {"type": "string"},
@@ -156,7 +156,7 @@
         "image": {"type": "string"},
         "ipc": {"type": "string"},
         "isolation": {"type": "string"},
[...]

PR #6680, opened 2019-05-01
enable support for `dns_opt` in compose again

jeroenpeeters:
I found that in docker-compose v3 the dns_opt option was not enabled in the config_schema. The option however is correctly passed to the Docker CLI. A test even still exists for this, see `tests/integration/service_test.py` `test_dns_opt_value`.

This PR enables the dns_opt again. The schema is a copy from v3.7 with the addition of dns_opt on line 133.

Tested against Docker version 18.09.1, build 4c52b90.

Resolves
- #5010 
- https://github.com/docker/cli/issues/1557


ijc:
BTW see https://github.com/docker/compose/pull/6642 for an attempt at a resync (and the problems caused by editing things directly here).

phyrwork:
Looks like this has been done since 1.27.0.

ndeloof:
dns_opt indeed has been re-introduced in compose spec.
Closing this PR as "obsolete"

PR #6696, opened 2019-05-10
Offer docompo alias for docker-compose command

jpic:
Just a quick patch to provide a shorter alias without dash that's more friendly in the command line: docompo alias for docker-compose. docompo kind of sounds like docker-compose except it should autocomplete fine after typing docoTAB

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -95,7 +95,10 @@ setup(
     tests_require=tests_require,
     python_requires='>=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*',
     entry_points={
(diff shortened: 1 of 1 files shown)


ulyssessouza:
Hello @jpic! I completely understand your point, but I don't see how diverging in the command name after so many years. 
And then, this opens the field for questions like: Why `docompo` and not `dc` or any other small version?

IMHO, people that want shorter versions should include aliases in their shell configurations with their favorite flavor.

Obs: I'm another one that don't type `docker-compose` on my everyday usage ;)

jpic:
Nice, ok then let every people who install docker-compose make their own favorite alias ;)

PR #6700, opened 2019-05-12
Bump urllib3

ulyssessouza:
Signed-off-by: ulyssessouza <ulysses.souza@docker.com>

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -20,5 +20,5 @@ PyYAML==4.2b1
 requests==2.20.0
 six==1.10.0
 texttable==0.9.1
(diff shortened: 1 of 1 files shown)


requirements.txt
```diff
@@ -20,5 +20,5 @@ PyYAML==4.2b1
 requests==2.20.0
 six==1.10.0
 texttable==0.9.1
-urllib3==1.21.1; python_version == '3.3'
+urllib3==1.24.2; python_version == '3.3'
```
ijc: Why not 1.24.3, or even 1.25.2?

That said, this is clearly an improvement so if we want to take a smaller step fine, LGTM!

PR #6701, opened 2019-05-13
Make test and build scripts POSIX compatible

### LINKED ISSUE #6573: Make test scripts POSIX compatible

joakimr-axis:
There is no need for these scripts to be bash-specific. This patch
contains the bare minimum alterations to make them POSIX compatible.
The benefits of having portable scripts, when possible, are obvious.

Change-Id: I1bb78f77653e948fa301c1e77bee4ff55ca415c6

Resolves #6573 #6575

### DIFF
--- a/script/build/image
+++ b/script/build/image
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env sh
 
 set -e
 
--- a/script/build/linux
+++ b/script/build/linux
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env sh
 
 set -ex
 
--- a/script/build/linux-entrypoint
+++ b/script/build/linux-entrypoint
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env sh
 
 set -ex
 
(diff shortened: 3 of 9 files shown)


ulyssessouza:
Same as https://github.com/docker/compose/pull/6576#issuecomment-491804574

joakimr-axis:
I am really surprised that someone who works in the container field does _not_ see the value (or value at all) in portability.

joakimr-axis:
(And finds the non-portable `[[ "${?}" != "0" ]] `easier to read than the portable `[ $? -ne 0 ]`)

ulyssessouza:
Portability is not an issue here, by the very reason that we use containers.
If you enumerate the places where the scripts are called, you will see that they are called inside non final stages of multistage builds. Given that, it does not matter on what they are written, since they are not available to the user.

About the readability, I can also mention this https://github.com/docker/compose/pull/6576/files#diff-c5a5d13eae0db96a50b3e0b4be489b56R8 as less readable a classical `if [ ... ]; then`.
So since intermediate shell scripts are not the focus of the project, we care about readability and uniformity more than portability, since the last one is useless in this scope of the project.

PR #6744, opened 2019-06-08
[W.I.P.] Use yaml.safe_load rather than yaml.load

hartwork:
Mostly aiming at:
- addressing security warning that tools might show
- fixing the test suite for operating systems that path `yaml.load` to be disallowed


hartwork:
Closing as a duplicate, found #6623 just now

PR #6778, opened 2019-07-02
Strip up generic versions and bump requests

### LINKED ISSUE #6756: setup.py: Generic upper bounds on dependency versions are a problem to users and to security
Dear docker-compose team,

`setup.py` has upper bound limits for all dependencies of docker-compose as of today, e.g. `docopt >= 0.6.1, < 0.7` has an upper bound of `< 0.7`. While some bounds seem to follow semver logic (e.g. `six >= 1.3.0, < 2`), some don't and seem rather generic, e.g. `texttable >= 0.9.0, < 0.10`.  In a context where you install nothing but docker-compose into a dedicated virtualenv, that approach might work well.  Once there are *neighbors* involved, this approach becomes *challenging*.

During the livetime of docker-compose version X, dependencies can do new releases, even multiple of them.  If docker-compose keeps dependencies in the past artificially, the system around it can only be as up to date as docker-compose.  If the system around it is a Linux distribution, that means that users cannot install things like jsonschema 3.0.1 (see #6347), pyyaml 5.1 (see #6619), texttable 1.6.1 or requests 2.22.0 (in the past, see #6717) without uninstalling or patching docker-compose.  That has consequences regarding flexibility as well as security (see #6619).

Because those upper limits are a problem, major Linux distributions need to patch them away:
- [Debian does](https://sources.debian.org/patches/docker-compose/1.21.0-3/Relax-dependencies.patch/)
- [Fedora does](https://src.fedoraproject.org/rpms/docker-compose/blob/master/f/docker-compose.spec#_62)
- [Gentoo does](https://github.com/gentoo/gentoo/blob/da5c9b70bf178b68f852270718ef67dcbbfcb0c7/app-emulation/docker-compose/files/docker-compose-1.23.2-setup-py.patch)
- *Your favorite distro* probably does, too

In the interest of their users, distro packagers like me have two options:

- a) A static patch that needs manual adjustment for every version bump — compare [1](https://sources.debian.org/patches/docker-compose/1.8.0-2/Relax-dependencies.patch/) with [2](https://sources.debian.org/patches/docker-compose/1.21.0-3/Relax-dependencies.patch/) in Debian
- b) A mass-replacement a la `sed 's/, < [0-9.]\+//' -i setup.py` that is rather brittle

What I'd like to ask for is this:

Please drop all generic upper bounds and only re-add limits for dependencies that are _known_ broken, a blacklist approach, not a whitelist one.

The remaining list of upper bounds should be way shorter and expose actual bockers rather than artificial ones.  Linux distros could then install unpatched `setup.py` files and users would be more up to date and more secure with their python packages.

Thanks a lot in advance!

Best

Sebastian

ulyssessouza:
Replaces generic limitations with a next major value
Bump the minimal `requests` to 2.20.0

Resolves #6756

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -31,31 +31,31 @@ def find_version(*file_paths):
 
 install_requires = [
     'cached-property >= 1.2.0, < 2',
-    'docopt >= 0.6.1, < 0.7',
-    'PyYAML >= 3.10, < 4.3',
-    'requests >= 2.6.1, != 2.11.0, != 2.12.2, != 2.18.0, < 2.23',
-    'texttable >= 0.9.0, < 0.10',
-    'websocket-client >= 0.32.0, < 1.0',
-    'docker[ssh] >= 3.7.0, < 4.0.2',
-    'dockerpty >= 0.4.1, < 0.5',
+    'docopt >= 0.6.1, < 1',
+    'PyYAML >= 3.10, < 5',
+    'requests >= 2.20.0, < 3',
+    'texttable >= 0.9.0, < 1',
+    'websocket-client >= 0.32.0, < 1',
+    'docker[ssh] >= 3.7.0, < 5',
+    'dockerpty >= 0.4.1, < 1',
     'six >= 1.3.0, < 2',
     'jsonschema >= 2.5.1, < 3',
 ]
 
 
 tests_require = [
-    'pytest',
+    'pytest < 6',
 ]
 
 
 if sys.version_info[:2] < (3, 4):
-    tests_require.append('mock >= 1.0.1')
+    tests_require.append('mock >= 1.0.1, < 2')
 
 extras_require = {
     ':python_version < "3.4"': ['enum34 >= 1.0.4, < 2'],
-    ':python_version < "3.5"': ['backports.ssl_match_hostname >= 3.5'],
-    ':python_version < "3.3"': ['ipaddress >= 1.0.16'],
-    ':sys_platform == "win32"': ['colorama >= 0.4, < 0.5'],
+    ':python_version < "3.5"': ['backports.ssl_match_hostname >= 3.5, < 4'],
+    ':python_version < "3.3"': ['ipaddress >= 1.0.16, < 2'],
+    ':sys_platform == "win32"': ['colorama >= 0.4, < 1'],
     'socks': ['PySocks >= 1.5.6, != 1.5.7, < 2'],
 }
 


ulyssessouza:
Hello @hartwork ! 
Can you review this PR to check if that's a reasonable solution for your issue (#6756) ?

setup.py
```diff
@@ -31,31 +31,31 @@ def find_version(*file_paths):
...
 
 tests_require = [
-    'pytest',
+    'pytest < 6',
 ]
 
 
 if sys.version_info[:2] < (3, 4):
-    tests_require.append('mock >= 1.0.1')
+    tests_require.append('mock >= 1.0.1, < 2')
```
hartwork: That would downgrade people with mock 2.0.0 and upstream is at 3.0.5 by now, so that change makes things worse.

setup.py
```diff
@@ -31,31 +31,31 @@ def find_version(*file_paths):
...
-    'PyYAML >= 3.10, < 4.3',
-    'requests >= 2.6.1, != 2.11.0, != 2.12.2, != 2.18.0, < 2.23',
-    'texttable >= 0.9.0, < 0.10',
-    'websocket-client >= 0.32.0, < 1.0',
-    'docker[ssh] >= 3.7.0, < 4.0.2',
-    'dockerpty >= 0.4.1, < 0.5',
+    'docopt >= 0.6.1, < 1',
+    'PyYAML >= 3.10, < 5',
+    'requests >= 2.20.0, < 3',
+    'texttable >= 0.9.0, < 1',
```
hartwork: That still forces users to downgrade from texttable 1.6.1 and from PyYAML 5.1.

setup.py
ulyssessouza: The idea here was just to use the current minimal boundaries to set the upper ones as it's a guaranty (as a safe move).
Then we should to bump every requirement independently (including `requirements.txt` so that the binary can evolve too).

setup.py
ulyssessouza: Same as above

setup.py
hartwork: > The idea here was just to use the current minimal boundaries to set the upper ones as it's a guaranty (as a safe move).

Original `mock >= 1.0.1` pulled in 3.0.5 effectively so it should have been `mock <4` not `mock <2` to allow the same version as before.

> Then we should to bump every requirement independently (including `requirements.txt` so that the binary can evolve too).

That sounds right for texttable and pyyaml but not for mock.
Is there a guarantee that this branch is not ending up in a release? Is there a ticked or pull request about fixing the mock dependency regression now?

ulyssessouza: Sure! I'll do the respective pull requests today :smiley:

ulyssessouza:
Hello @hartwork 

As promissed, I did 2 PR's to bump texttable(https://github.com/docker/compose/pull/6792) and mock(https://github.com/docker/compose/pull/6791). 

For the case of PyYAML, it's more complicated. 
Please have a look at https://github.com/docker/compose/pull/6623

hartwork:
Thank you

PR #6535, opened 2019-02-23
Support for --env-file option for docker-compose

slowr:
Added additional argument (--env-file) for docker-compose to import environment variables from a given file.

Resolves #6170

PR #6800, opened 2019-07-12
Make '--env-file' option top-level only and fix failure with subcommands

### LINKED ISSUE #6746: '--env-file' option is ignored
## Description of the issue

The `--env-file` option added in PR #6535 doesn't appear to work.  The commands accept the argument, but the variables from the specified env file don't make it into the environment.

I tried the option with `run` and `up`, both before and after the subcommand.  For `up` it just gets ignored in either position.  When I put it after `run` I get the usage message, but if I add the option into [the options list in the `run` docstring](https://github.com/docker/compose/blob/1.25.0-rc1/compose/cli/main.py#L852-L870), it accepts the option but ignores it, just like when it's in the other position.

```
docker-compose version 1.25.0-rc1, build 8552e8e
```

```
Docker version 18.09.5, build e8ff056
```

This makes a super-simple compose environment using the busybox image, creates default and override `.env` files, and runs the container to show the bug.
```
cat <<\EOF > docker-compose.yml
version: '3.7'
services:
  test:
    image: busybox
    env_file: .env.conf
EOF

echo 'WHEREAMI=config' > .env.conf
echo 'WHEREAMI=override' > .env.override

docker-compose --env-file .env.override run --rm test env | grep WHEREAMI
```
### Observed result
`WHEREAMI=config`
### Expected result
`WHEREAMI=override`

## Additional information
If you remove the `env_file` line from the `docker-compose.yml` the variable disappears, so it's not a matter of preferring the config, it's just not managing to get the file from `--env-file` loaded into the environment.

Installation details: Ubuntu 18.04, using system Docker and installing `docker-compose` with `pipenv`, both as `docker-compose==1.25.0-rc1` and by installing the cloned repo at `master` (same result).

KlaasH:
Resolves #6746

The `--env-file` option added in PR #6535 makes sense as a top-level option but was added both to `TopLevelCommand` and to some subcommands.  At best, the handling in the subcommands is redundant and muddies the waters; at worst, it causes bugs, like the fact that, [as noted by @digitalist](https://github.com/docker/compose/issues/6746#issuecomment-510470371), `docker-compose --env-file=custom.env up` doesn't work, loading the `.env` file despite the attempt to override it.

So this removes the option from `build`, `up`, and `down`; moves the handling of the option into a method on `TopLevelCommand`; and changes the naming a bit to emphasize that this is a top-level option that controls the top-level environment (I've also been thinking of it as "project-level", as opposed to "service-level").

#### Testing

This does two main things:
- Removes the `--env-file` option from the `build`, `up`, and `down` subcommands. You can see that in their usage messages (which you can view by trying to put the option after the subcommand...)
- Resolves the issue whereby mishandling by some subcommands caused the top-level option to be ignored.  For that, you can do something like this, which creates a simple `docker-compose.yml` and a few env files and runs a command to show that the override is happening (do it in a new directory to avoid clobbering anything important):
```
cat <<\EOF > docker-compose.yml
version: '3.7'
services:
  test:
    image: busybox
    env_file: .env.conf
    entrypoint: env
EOF

echo 'WHEREAMI=default' > .env
echo "This won't parse as a variable" >> .env
echo 'WHEREAMI' > .env.conf
echo 'DEFAULT_CONF_LOADED=true' >> .env.conf
echo 'WHEREAMI=override' > .env.override

docker-compose --env-file .env.override up
```
The output will include `WHEREAMI=override` when it's working.  When it's not, i.e. on `master`, it'll produce an error saying there can't be whitespace in a variable name.

This doesn't add any cases to the test suite because I couldn't really think of good cases to test.  PR #6535 added tests for the "working override" case and there were already tests for the "working default" case, and this mostly removes the stuff that's not covered by those.  Adding a test that e.g. `up` works the same as `config` seems like it would take a bit of doing and wouldn't test anything very interesting under the circumstances.

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -248,7 +248,7 @@ class TopLevelCommand(object):
         return self.toplevel_options.get('--project-directory') or '.'
 
     @property
-    def environment(self):
+    def toplevel_environment(self):
         environment_file = self.toplevel_options.get('--env-file')
         return Environment.from_env_file(self.project_dir, environment_file)
 
@@ -281,7 +281,7 @@ class TopLevelCommand(object):
                     '--build-arg is only supported when services are specified for API version < 1.25.'
                     ' Please use a Compose file version > 2.2 or specify which services to build.'
                 )
-            build_args = resolve_build_args(build_args, self.environment)
+            build_args = resolve_build_args(build_args, self.toplevel_environment)
 
         self.project.build(
             service_names=options['SERVICE'],
@@ -433,7 +433,7 @@ class TopLevelCommand(object):
             -t, --timeout TIMEOUT   Specify a shutdown timeout in seconds.
                                     (default: 10)
         """
-        ignore_orphans = self.environment.get_boolean('COMPOSE_IGNORE_ORPHANS')
+        ignore_orphans = self.toplevel_environment.get_boolean('COMPOSE_IGNORE_ORPHANS')
 
         if ignore_orphans and options['--remove-orphans']:
             raise UserError("COMPOSE_IGNORE_ORPHANS and --remove-orphans cannot be combined.")
@@ -490,7 +490,7 @@ class TopLevelCommand(object):
                               not supported in API < 1.25)
             -w, --workdir DIR Path to workdir directory for this command.
         """
-        use_cli = not self.environment.get_boolean('COMPOSE_INTERACTIVE_NO_CLI')
+        use_cli = not self.toplevel_environment.get_boolean('COMPOSE_INTERACTIVE_NO_CLI')
         index = int(options.get('--index'))
         service = self.project.get_service(options['SERVICE'])
         detach = options.get('--detach')
@@ -513,7 +513,7 @@ class TopLevelCommand(object):
         if IS_WINDOWS_PLATFORM or use_cli and not detach:
             sys.exit(call_docker(
                 build_exec_command(options, container.id, command),
-                self.toplevel_options, environment)
+                self.toplevel_options, self.toplevel_environment)
             )
 
         create_exec_options = {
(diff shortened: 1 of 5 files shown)


KlaasH:
Rebased to resolve conflict with 9d2508cf.

compose/cli/main.py
```diff
@@ -1358,9 +1352,10 @@ def remove_container(force=False):
         if options['--rm']:
             project.client.remove_container(container.id, force=True, v=True)
 
-    environment_file = options.get('--env-file')
-    environment = Environment.from_env_file(project_dir, environment_file)
-    use_cli = not environment.get_boolean('COMPOSE_INTERACTIVE_NO_CLI')
+    environment_file = toplevel_options.get('--env-file')
+    toplevel_environment = Environment.from_env_file(project_dir, environment_file)
+    use_cli = not toplevel_environment.get_boolean('COMPOSE_INTERACTIVE_NO_CLI')
```
ulyssessouza: Why not using `use_cli = not self.toplevel_environment.get_boolean('COMPOSE_INTERACTIVE_NO_CLI')` here too?

KlaasH: This is a stand-alone function, not a method of `TopLevelCommand`.  Though it seems like it's only called in `TopLevelCommand.run`, so adding an argument to pass `toplevel_environment` in could work fine and remove this duplication.

KlaasH: Added a commit to do this--replaced the `project_dir` argument (which was only used for this) with a `toplevel_environment` argument.

ulyssessouza: Cool! I didn't have enough context in the browser based diff to see that.

rumpl:
Should we update the bash completion in this PR or in a followup?

KlaasH:
Added `--env-file` completions for `bash`, `fish`, and `zsh`.

KlaasH:
Rebased onto master to resolve conflict with #6813.

PR #6815, opened 2019-07-24
Add project_name top-level config key

### LINKED ISSUE #745: Proposal: make project-name persistent.

pipe01:
Resolves #745

I suppose the JSON schema change should go on a different version but I'm unsure about how to do that.


terjebraten-certua:
Duplicate of #3118 and #5369

ulyssessouza:
As commented in https://github.com/docker/compose/pull/6815#issuecomment-517624085, this is an already refused feature.

PR #6848, opened 2019-08-13
Use new style markers in setup.py

### LINKED ISSUE #6839: Update setup.py to be pipenv compatible.
## Feature Request

When using pipenv, [these lines](https://github.com/docker/compose/blob/f9d1075a5da575cf7fd58d058fd4f4b402aff688/setup.py#L63-L73) in `setup.py` don't have the appropriate markers after installing. E.g.: if I use `pipenv` in Linux and try to sync on Windows, colorama is not installed.

The [recommended](https://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-platform-specific-dependencies) way to declare platform specific dependencies has a set of markers different in `install_requires` from the ones in `extra_requires`.

## Related Problems

https://github.com/pypa/pipenv/issues/1229, specifically mentions [here](https://github.com/pypa/pipenv/issues/1279#issuecomment-363305721) that using `extra_requires` is an old (and unsupported) way 

## Desired Solution

If instead of just adding the dependencies to the `install_requires` list when that particular systems meets the requirements, instead add to the list using the appropriate `install_requires` markers.

Current:

```
extras_require = {
    ':python_version < "3.4"': ['enum34 >= 1.0.4, < 2'],
    ':python_version < "3.5"': ['backports.ssl_match_hostname >= 3.5, < 4'],
    ':python_version < "3.3"': ['ipaddress >= 1.0.16, < 2'],
    ':sys_platform == "win32"': ['colorama >= 0.4, < 1'],
}
```

Instead adds this to `install_requires`

```
[
    'enum34;python_version<"3.4"',
    'backports.ssl_match_hostname >= 3.5, < 4;python_version<"3.5"
    'ipaddress >= 1.0.16, < 2;python_version<"3.3"',
    'colorama >= 1.0;sys_platform=="win32"'
]
```

socks would stay in extra_requires, because it's an "extra" package. 

## Workaround

For now, I have to manually duplicate the dependencies colorama in my `Pipfile` for windows, adding

```
colorama = {version = ">=0.4, <1", markers = "sys_platform == 'win32'"}
```

## References

[PEP 0496](https://www.python.org/dev/peps/pep-0496/)

andyneff:
(I just used python2 as a way to check the python version < 3.3/3.4/3.5 stuff) 
- Tested under a python2 virtual env+pip, included addition packages
- Tested under a python3 virtual env+pip, did not include addition packages
- Tested under a python2 pipenv, included addition packages
- Tested under a python3 pipenv, did not include addition packages
- `python2 setup.py test`, installed mock
- `python3 setup.py test`, did not install mock

So it looked like tests_require respected the same markers 👍

I removed the `bdist_wheel` check, assuming it was not important with markers being used this way. I did not fully understand why it was there, so let me know if I need to put it back in.

@ulyssessouza


ulyssessouza:
Stale

Issue #6511, opened 2019-02-07
`.env` with `export` Lines No Longer Works


berwyn:
## Description of the issue
I recently upgraded Compose to 1.24.0-rc1 (build `0f3d4dda`) and now Compose's automatic parsing of my `.env` file fails. I keep `export` statements in my `.env` file so I can easily `source` it in addition to using it as a standard `.env`. In previous versions of Compose, this worked fine and didn't give me any issues, however with this new update I instead get an error about spaces inside a value.

## Context information (for bug reports)

**Output of `docker-compose version`**
```
docker-compose version 1.24.0-rc1, build 0f3d4dda
```

**Output of `docker version`**
```
Docker version 18.09.1, build 4c52b90
```

**Output of `docker-compose config`**
(Make sure to add the relevant `-f` and other flags)
```
services:
  db:
    image: mdillon/postgis:10-alpine
    ports:
    - 5432:5432/tcp
    volumes:
    - pgdata:/var/lib/postgresql/data:rw
  elasticsearch:
    environment:
      discovery.type: single-node
    image: docker.elastic.co/elasticsearch/elasticsearch:6.1.3
  memcached:
    image: memcached:1.5.10-alpine
  redis:
    image: redis:4.0.6-alpine
  web:
    build:
      context: /Users/berwyn/dev/<repo>
    depends_on:
    - db
    - elasticsearch
    - memcached
    - redis
    environment:
      DB_HOST: db
      DB_USERNAME: postgres
      FOUNDELASTICSEARCH_URL: elasticsearch
      REDIS_CACHE_HOST: redis
      # Several more removed variables here
    ports:
    - 3000:3000/tcp
    stdin_open: true
    tty: true
    volumes:
    - /Users/berwyn/dev/<repo>:/app:rw
version: '3.0'
volumes:
  pgdata: {}
```

## Steps to reproduce the issue

1. Create a `.env` file with something like `export FOO=1`
2. Attempt to run a one-off container `docker-compose run --rm service bash`
3. Notice the error

### Observed result
```
ERROR: In file ./.env: environment variable name `export FOO` may not contains whitespace.
```

### Expected result
The container runs and `FOO` is correctly set in it.

### Stacktrace / full error message
```
ERROR: In file ./.env: environment variable name `export FOO` may not contains whitespace.
```

## Additional information
macOS 10.14.3
Docker & Friends installed using Cask (`cask install docker-edge`)
[...]

Issue #6838, opened 2019-08-07
Variable name may not contain whitespace


alfredopalhares:
## Description of the issue

Docker compose does not recnognize a valid yaml file, even when validated with a yamlint.org. 

## Context information (for bug reports)

**Output of `docker-compose version`**
```
-> % docker-compose version
docker-compose version 1.24.1, build unknown
docker-py version: 4.0.2
CPython version: 3.7.4
OpenSSL version: OpenSSL 1.1.1c  28 May 2019
```

**Output of `docker version`**
```
Client:
 Version:           19.03.1-ce
 API version:       1.40
 Go version:        go1.12.7
 Git commit:        74b1e89e8a
 Built:             Sat Jul 27 21:08:50 2019
 OS/Arch:           linux/amd64
 Experimental:      false

Server:
 Engine:
  Version:          19.03.1-ce
  API version:      1.40 (minimum version 1.12)
  Go version:       go1.12.7
  Git commit:       74b1e89e8a
  Built:            Sat Jul 27 21:08:28 2019
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          v1.2.7.m
  GitCommit:        85f6aa58b8a3170aec9824568f7a31832878b603.m
 runc:
  Version:          1.0.0-rc8
  GitCommit:        425e105d5a03fabd737a126ad93d62a9eeede87f
 docker-init:
  Version:          0.18.0
  GitCommit:        fec3683

```

**Output of `docker-compose config`**
(Make sure to add the relevant `-f` and other flags)
```
ERROR: In file /home/masterkorp/Documents/prozis/repos/prozis-events/events-backend/docker/env/its.yaml: environment variable name 'host: hostname' may not contains whitespace.

```

## Steps to reproduce the issue
Just tun any docker command
### Observed result
Docker compose fails to run
### Expected result
Should run

## Additional information

**OS** Arch Linux Updated, but it happens accroos all OS.

ulyssessouza:
@alfredopalhares Could you please, provide the YAML file?

tjb-wiley:
i experienced this same issue, and after spending most of the afternoon, it looks like like maybe something changed with the way docker-compose processes `.env` files. and the only reason i hit this is because i upgraded docker on my mac laptop from `docker for mac` to `docker desktop`.

# current behavior on macos
```sh
docker version
Client: Docker Engine - Community
 Version:           19.03.1
 API version:       1.40
 Go version:        go1.12.5
[...]

PR #6850, opened 2019-08-14
add --skip-env-file option

### LINKED ISSUE #6741: Feature Request: option to disable reading from .env
Similar/repeat request to #4642. New/different reason now though. Related to https://github.com/docker/compose/issues/6511.

Other systems use the exact same file `.env` for parsing and processing environment variables. With the release of 1.24.0, the feature where `Compose will no longer accept whitespace in variable names sourced from environment files. This matches the Docker CLI behavior.` breaks compatibility with other .env utilities. Although my setup does not use the variables in .env for docker-compose, docker-compose now fails because the .env file does not meet docker-compose's format.

I either have to make the .env file fit docker-compose's format or remove the file.

I propose that you can specify an option to ignore the `.env` file or specify a different `.env` file (such as `.docker.env`) in the `docker-compose.yml` file so that we can work around projects that are already using the `.env` file for something else.

hirochachacha:
Fixes #6741
Updates #6511


alecgibson:
I'm not really sure this fixes https://github.com/docker/compose/issues/6741. It's not really that different to having a separate (empty) Docker `.env` file and using the existing `--env-file` argument.

I think the whole point of the issue is that Docker is being a bit too over-eager with its defaults (ie I'd expect the default to _not_ read anything, and if we have a `.env` file, that we'd have to specify which one to read). If you _do_ want to do automagical auto-reading, I'd expect it to only read an application-specific file (in the same way you might have a `.gitignore`, `.eslintrc`, `.bablerc`, etc.).

Obviously that's all a breaking change, so I'm not really sure what to suggest, apart from the original suggestion in the issue where we can specify `.env` from the `docker-compose.yml` file.  Perhaps another suggestion might be to attempt to parse the `.env` file, and only `WARN` if it's not valid (because then it _probably_ wasn't meant for Docker consumption?)

hirochachacha:
Hi, @jonmchan @alecgibson 

Thank you for your comments.

> I don't think it addresses the fundamental need that some projects need docker to not evaluate .env at all. Forcing devs to always type --skip-env-file as part of their regular workflow does not properly address the issue imho.

It's true that this CL doesn't implement the yaml part, but I have a reason. Let me explain.
```
Options:
  -f, --file FILE             Specify an alternate compose file
                              (default: docker-compose.yml)
  -p, --project-name NAME     Specify an alternate project name
                              (default: directory name)
  --log-level LEVEL           Set log level (DEBUG, INFO, WARNING, ERROR, CRITICAL)
  --no-ansi                   Do not print ANSI control characters
  -v, --version               Print version and exit
  -H, --host HOST             Daemon socket to connect to

  --tls                       Use TLS; implied by --tlsverify
  --tlscacert CA_PATH         Trust certs signed only by this CA
  --tlscert CLIENT_CERT_PATH  Path to TLS certificate file
  --tlskey TLS_KEY_PATH       Path to TLS key file
  --tlsverify                 Use TLS and verify the remote
  --skip-hostname-check       Don't check the daemon's hostname against the
                              name specified in the client certificate
  --project-directory PATH    Specify an alternate working directory
                              (default: the path of the Compose file)
  --compatibility             If set, Compose will attempt to convert keys
                              in v3 files to their non-Swarm equivalent
  --env-file PATH             Specify an alternate environment file
  --skip-env-file             Skip loading an environment file
```
This is an excerpt from the output of `docker-compose --help`.
You'll probably notice that your point can be applied almost all options aside from `--skip-env-file`.
In other words, we actually have two different problems.

1. There is no options to disable .env loading.
2. Putting a command with long options isn't user friendly.

I'd say it's better to focus the problem 1 on the issue, and we could make another issue or proposal for the problem 2. The problem 2 is general problem, which presumably require widely discussions. 

> It's not really that different to having a separate (empty) Docker .env file and using the existing --env-file argument.

The result is the same, but I think it's more understandable. At least, you don't have to create empty .env, also your colleagues don't have check the .env.

> I think the whole point of the issue is that Docker is being a bit too over-eager with its defaults (ie I'd expect the default to not read anything, and if we have a .env file, that we'd have to specify which one to read). If you do want to do automagical auto-reading, I'd expect it to only read an application-specific file (in the same way you might have a .gitignore, .eslintrc, .bablerc, etc.).

Maybe, but I don't think it's going to happen soon. I assume it requires a kind of major version upgrade.

> Perhaps another suggestion might be to attempt to parse the .env file, and only WARN if it's not valid (because then it probably wasn't meant for Docker consumption?)

Yes, it's a possible solution. I'd like to note that it can coexist with this one. I'd say it's nice there is a solution without `WARN`.

jonmchan:
I did start looking into checking docker-compose.yml for a `skip-env-file` flag, but I found that the 1st thing the docker-compose does is read the `.env` file before reading in the `docker-compose.yml` file. This makes a lot of sense since the .yml file uses variable substitution that can be taken from the `.env` file. If we were to implement a flag in the `docker-compose.yml` file, it would have to read the file 2x. Wasn't sure how favorable the maintainers would be to such a solution which is why I never put in a PR.

I think though that reading from the `docker-compose.yml` file is the only long term solution that meets all the requirements. For users that use `.env` for other programs, they would want a way to permanently disable `docker-compose` from using the file. A command line option would mean this command would have to be typed for every invocation every time (imagine how inconvenient - `docker-compose up --skip-env-file`, `docker-compose down --skip-env-file`, `docker-compose build --skip-env-file`) typing the flag over and over again doesn't seem like the proper solution. Thoughts?

jonmchan:
@hirochachacha btw, i want to let you know i did read your reply and i acknowledge your approach and what you're trying to do; the point I am trying to reiterate is that long term, it doesn't make sense for a project that wants to ignore `.env` file to use any command line option at all - short or long option. Having the project set to never use `.env` in something like `docker-compose.yml` makes the most sense to me.

esvm:
Some news about this?

Quintasan:
Can we not merge this and revert to docker-compose ignoring lines from `.env` it can't understand? I see no reason for this to be merged. The only thing it does is breaking workflows for people.

esvm:
> Can we not merge this and revert to docker-compose ignoring lines from `.env` it can't understand? I see no reason for this to be merged. The only thing it does is breaking workflows for people.

Actually we have a lot of people with problems with the support of `.env`, but we also have a lot of people enjoying this feature. Remove this feature isn't a good idea because it will cause more problems with the people that are already using this. I think that add an option in CLI is actually the best solution we have. And it needs to be merged as soon as possible, is really really so boring downgrade the compose version every time or exclude my .env files and add after

Quintasan:
@esvm I fail to understand why ignoring lines that do not conform to `VARIABLE=VALUE` would be problematic for anyone.

hirochachacha:
Previous behavior was a bug, that was out of spec. If you try to revert that behavior on purpose,  you don't just solve your problem, but also make that behavior be in spec. That strange spec could be problematic for people in future. Changing the spec just for working around your issue? I'm not sure that whether you deserve it or not.

hirochachacha:
Btw, I found that `pipenv` uses the clever approach. 

 *  https://github.com/pypa/pipenv/blob/master/docs/advanced.rst#-automatic-loading-of-env
 *  https://github.com/pallets/flask/issues/2722

They're using environmental variable instead of CLI parameters.
It maybe satisfy some people's requirements.
Once, you add `COMPOSE_SKIP_ENV_FILE=1`, you can forget everything.

Quintasan:
> Previous behavior was a bug, that was out of spec. If you try to revert that behavior on purpose, you don't just solve your problem, but also make that behavior be in spec. That strange spec could be problematic for people in future. Changing the spec just for working around your issue? I'm not sure that whether you deserve it or not.

Yes, `VAR=VAL` syntax appears to be part of the [spec](https://docs.docker.com/compose/env-file/). However the same spec does NOT say what will happen to lines that are not compliant. Hence my request to make docker-compose ignore them. This way you get what you wanted and most of people reporting the issues get to keep their `.env` files and deploy schemas. All you'd have to do is adjust the spec and revert to the previous behaviour. I don't think that *skipping* the lines that don't conform to the schema docker-compose prefers is that big of a cost. And it's not that rare in the software development industry for the spec to change.

I'm [hardly](https://github.com/docker/compose/issues/6511) [the](https://github.com/docker/compose/issues/6511#issuecomment-484036389) [only](https://github.com/docker/compose/issues/6511#issuecomment-485446237) [person](https://github.com/anttiviljami/wordpress-heroku-docker-project/issues/17) [that](https://github.com/docker/compose/issues/6511#issuecomment-509753742) [has](https://github.com/docker/compose/issues/6511#issuecomment-520605804) [problems](https://github.com/docker/compose/issues/6838) [with](https://github.com/docker/compose/issues/6511#issuecomment-528136681) [docker-compose](https://github.com/docker/compose/issues/6511#issuecomment-528873241) fixing this bug. Please note that docker-compose is not the only application in the entire world that is using this file as mentioned [here](https://github.com/docker/compose/issues/6511#issuecomment-498457183).

And to address the last part of your reply - I don't deserve anything. I don't think any person in the thread reporting the issue deserves anything. You are free to run your project in any way you want. Since you decided to let your users raise issues with the software I'm free to raise the issue of docker-compose breaking my workflow in many application (which, I unfortunately can't show you since they are not my IP) and to politely disagree with you. If you felt offended by any of my statements then I'd like to apologize - it was not my intention to offend anyone.

jonmchan:
@hirochachacha - using pipenv's approach sounds like a favorable compromise. Although users won't be able to have project specific settings, I suspect that most people will not be working with multiple projects that utilize docker env and other projects that do not on the same system.

Quintasan:
@hirochachacha Okay, let me put thing together:
1. This issue is very minor
2. The spec is incomplete with regards to expected behaviour when encountering "incompatible" lines
3. You refuse to change the spec to state the desired behaviour when encountering incompatible lines
4. You are adamant to implement the breaking change despite a better option being present

If this issue is minor (as you said it) then why the adamant refusal to adopt the behaviour that would result in the smallest amount of breakage and would **still** give you the result you wanted?

hirochachacha:
Actually, I'm fine with any solutions. To be honest, I'm not much interested in this issue.
I just said I don't think your option is better than this one.
However if you believe that your options is the best, then go ahead.
I'm not a stakeholder. You can try to convince docker-compose team instead of me.
Anyway, don't talk to me anymore.

CyrilHu:
Some news about this?

hirochachacha:
It looks like docker-compose team made a great decision. https://github.com/docker/compose/pull/7150
Since it changes some semantics, perhaps it comes with major version up?

willsheppard:
@jonmchan maybe reading the `docker-compose.yml` file twice would actually be okay.

Daghall:
Workaround: ```alias docker-compose='docker-compose --env-file /dev/null'```

glours:

PR #6858, opened 2019-08-20
Improve control over ANSI output

mikesep:
**Update 2019-10-27:** I added a new `--ansi (never|always|auto)` option to allow fine-grained control over when to give ANSI output.

-----

Improve control over ANSI output

- Disable parallel logger ANSI output if not attached to a tty.
  The console handler and progress stream already checked whether the
  output stream is a tty, but ParallelStreamWriter did not.

- Add a new --ansi option to allow clearer control over ANSI output.

Original description:

> The console handler and the progress stream already check whether the
output stream is a tty.
>
> Signed-off-by: Mike Seplowitz <mseplowitz@bloomberg.net>
>
> -----
>
> Looking for feedback on how to proceed from here.
> - Does this change seem reasonable?
> - Is the existing behavior (always ANSI output) important to people?
>   - Does this need to be behind a new CLI switch to avoid breaking people who rely on this behavior?
>   - Does there need to be an "always on" switch to let people get the existing behavior?
>
> This change breaks `test_parallel_execute_ansi` since the output stream is not a tty. I think we need to answer the questions above to know how to modify/replace that test.
>
> Thanks for reviewing!

-----

**Update 2020-12-16:** Reworked this PR after leaving it idle for almost a year. https://github.com/docker/compose/pull/6858#issuecomment-747059574

### DIFF
--- a/compose/cli/colors.py
+++ b/compose/cli/colors.py
@@ -1,3 +1,6 @@
+import enum
+import os
+
 from ..const import IS_WINDOWS_PLATFORM
 
 NAMES = [
@@ -10,6 +13,21 @@ NAMES = [
     'cyan',
     'white'
 ]
+
+
+@enum.unique
+class AnsiMode(enum.Enum):
+    """Enumeration for when to output ANSI colors."""
+    NEVER = "never"
+    ALWAYS = "always"
+    AUTO = "auto"
+
+    def use_ansi_codes(self, stream):
+        if self is AnsiMode.ALWAYS:
+            return True
+        if self is AnsiMode.NEVER or os.environ.get('CLICOLOR') == '0':
+            return False
+        return stream.isatty()
 
 
 def get_pairs():
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -62,7 +62,6 @@ if not IS_WINDOWS_PLATFORM:
     from dockerpty.pty import PseudoTerminal, RunOperation, ExecOperation
 
 log = logging.getLogger(__name__)
-console_handler = logging.StreamHandler(sys.stderr)
 
 
 def main():  # noqa: C901
@@ -139,7 +138,9 @@ def exit_with_metrics(command, log_msg=None, status=Status.SUCCESS, exit_code=1)
 
 
 def dispatch():
-    setup_logging()
+    console_stream = sys.stderr
+    console_handler = logging.StreamHandler(console_stream)
+    setup_logging(console_handler)
     dispatcher = DocoptDispatcher(
         TopLevelCommand,
         {'options_first': True, 'version': get_version_info('compose')})
@@ -172,7 +173,7 @@ def perform_command(options, handler, command_options):
         handler(command, command_options)
 
 
-def setup_logging():
+def setup_logging(console_handler):
     root_logger = logging.getLogger()
     root_logger.addHandler(console_handler)
     root_logger.setLevel(logging.DEBUG)
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -2,7 +2,6 @@ import contextlib
 import functools
 import json
 import logging
-import os
 import pipes
 import re
 import subprocess
@@ -27,6 +26,7 @@ from ..config.types import VolumeSpec
 from ..const import IS_WINDOWS_PLATFORM
 from ..errors import StreamParseError
 from ..metrics.decorator import metrics
+from ..parallel import ParallelStreamWriter
 from ..progress_stream import StreamOutputError
 from ..project import get_image_digests
 from ..project import MissingDigests
@@ -40,6 +40,7 @@ from ..service import ImageType
 from ..service import NeedsBuildError
 from ..service import OperationFailedError
 from ..utils import filter_attached_for_up
+from .colors import AnsiMode
 from .command import get_config_from_options
 from .command import get_project_dir
 from .command import project_from_options
@@ -146,12 +147,30 @@ def dispatch():
         {'options_first': True, 'version': get_version_info('compose')})
 
     options, handler, command_options = dispatcher.parse(sys.argv[1:])
+
+    ansi_mode = AnsiMode.AUTO
+    try:
+        if options.get("--ansi"):
+            ansi_mode = AnsiMode(options.get("--ansi"))
+    except ValueError:
+        raise UserError(
+            'Invalid value for --ansi: {}. Expected one of {}.'.format(
+                options.get("--ansi"),
+                ', '.join(m.value for m in AnsiMode)
+            )
+        )
+    if options.get("--no-ansi"):
+        if options.get("--ansi"):
+            raise UserError("--no-ansi and --ansi cannot be combined.")
+        log.warning('--no-ansi option is deprecated and will be removed in future versions.')
+        ansi_mode = AnsiMode.NEVER
+
     setup_console_handler(console_handler,
                           options.get('--verbose'),
-                          set_no_color_if_clicolor(options.get('--no-ansi')),
+                          ansi_mode.use_ansi_codes(console_handler.stream),
                           options.get("--log-level"))
-    setup_parallel_logger(set_no_color_if_clicolor(options.get('--no-ansi')))
-    if options.get('--no-ansi'):
+    setup_parallel_logger(ansi_mode)
+    if ansi_mode is AnsiMode.NEVER:
         command_options['--no-color'] = True
     return functools.partial(perform_command, options, handler, command_options)
 
@@ -184,14 +203,12 @@ def setup_logging(console_handler):
     logging.getLogger("docker").propagate = False
 
 
-def setup_parallel_logger(noansi):
-    if noansi:
-        import compose.parallel
-        compose.parallel.ParallelStreamWriter.set_noansi()
+def setup_parallel_logger(ansi_mode):
+    ParallelStreamWriter.set_default_ansi_mode(ansi_mode)
 
 
-def setup_console_handler(handler, verbose, noansi=False, level=None):
-    if handler.stream.isatty() and noansi is False:
+def setup_console_handler(handler, verbose, use_console_formatter=True, level=None):
+    if use_console_formatter:
         format_class = ConsoleWarningFormatter
     else:
         format_class = logging.Formatter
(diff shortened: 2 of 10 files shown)


mikesep:
@rumpl and @ulyssessouza, could you help me move forward with getting some feedback on this PR?

ndeloof:
Hi @mikesep , can you please explain the issue you're facing and how this change is expected to fix it? Pretty hard to offer guidance without better understanding your usage scenario.

mikesep:
> Hi @mikesep , can you please explain the issue you're facing and how this change is expected to fix it? Pretty hard to offer guidance without better understanding your usage scenario.

Absolutely -- sorry for the lack of context.

My team uses `docker-compose` to set up and tear down environments in which our tests run. We don't run `docker-compose` directly, though -- we wrap our tests in a library that calls `docker-compose`. When we run this in Jenkins, the output gets captured in Jenkins logs, but it looks somewhat mangled due to the ANSI escape characters. Rather than re-implementing tty detection in our wrapper library and running `docker-compose --no-ansi ...`, I think it would be appropriate to have `docker-compose` disable ANSI escapes automatically when not attached to a tty. There seems to already be code to do this in [progress_stream.py](https://github.com/docker/compose/blob/2ec7615ed63d9fc2d3c3cb69b3c4022ffe639652/compose/progress_stream.py#L20) and [main.py](https://github.com/docker/compose/blob/2199278b4448c1fbf0637efa42a1ab97dc9b8573/compose/cli/main.py#L146) (at least I *think* that's what that code does).

mikesep:
@ndeloof please let me know what you think (or if you have more questions). Thanks!

ndeloof:
sounds reasonable, but not being a Python guy I won't be the one to take decision here on this PR. Please also fix build failure
```
tests/unit/parallel_test.py:174: in test_parallel_execute_ansi


    assert "\x1b" in err


E   AssertionError: assert 'n 'Control characters something      ... 
\nControl characters something more ... 
\nControl characters something      ... done
\nControl characters something more ... done
\n'
```

mikesep:
@ndeloof right, I wanted to get some advice from project maintainers as to whether the test is just no longer relevant or if there was a good way to fake this out in a Jenkins test.

> This change breaks `test_parallel_execute_ansi` since the output stream is not a tty. I think we need to answer the questions above to know how to modify/replace that test.

I'm less looking for technical advice than for advice on whether we need backwards-compatible behavior and how to test tty detection.

mikesep:
I added a new `--ansi (never|always|auto)` option to allow fine-grained control over when to give ANSI output.

I'm happy to update documentation and shell completion files if maintainers agree this is the right way to go forward.

mikesep:
I rebased onto master to take #6973 into account (/cc @glours). All tests pass. I hope this is ready for review (/cc @ndeloof @rumpl).

ndeloof:
I think this is a good approach, as default behaviour is sane (only enable ansi when a tty is detected) but easy to override. We could even deprecate `-no-ansi`.

rumpl:
@mikesep could you add this to the [completion scripts](https://github.com/docker/compose/tree/master/contrib/completion) too please?

We would need to update the documentation too (don't bother about it, we will do it, I'm just saying it here so that we don't forget)

mikesep:
@rumpl sounds good. 👍

mikesep:
Added completions in 4d2fd74. I was only able to test bash myself, so if someone wants to double-check my work for fish and zsh, that might be a good idea. 😄

mikesep:
@ulyssessouza, I see that @rumpl might be away. Can you help merge this PR?

compose/cli/main.py
```diff
@@ -95,18 +94,25 @@ def main():
...
+
     setup_console_handler(console_handler,
                           options.get('--verbose'),
-                          set_no_color_if_clicolor(options.get('--no-ansi')),
+                          ansi,
                           options.get("--log-level"))
-    setup_parallel_logger(set_no_color_if_clicolor(options.get('--no-ansi')))
-    if options.get('--no-ansi'):
+    setup_parallel_logger(ansi)
+    if ansi == "never":
```
ulyssessouza: I think that using an enumeration for the values of  `--ansi` would be better

compose/cli/main.py
```diff
@@ -137,14 +143,13 @@ def setup_logging():
...
-        compose.parallel.ParallelStreamWriter.set_noansi()
+def setup_parallel_logger(ansi):
+    import compose.parallel
+    compose.parallel.ParallelStreamWriter.set_ansi(ansi)
 
 
-def setup_console_handler(handler, verbose, noansi=False, level=None):
-    if handler.stream.isatty() and noansi is False:
+def setup_console_handler(handler, verbose, ansi="auto", level=None):
+    if ansi == "always" or (ansi == "auto" and handler.stream.isatty()):
```
ulyssessouza: This expression could be an enumeration method

compose/parallel.py
```diff
@@ -314,10 +314,10 @@ def _write_noansi(self, msg, obj_index, status):
     def write(self, msg, obj_index, status, color_func):
         if msg is None:
             return
-        if self.noansi:
-            self._write_noansi(msg, obj_index, status)
-        else:
+        if self.ansi == "always" or (self.ansi == "auto" and self.stream.isatty()):
```
ulyssessouza: Same enum method could be used here

mikesep:
Will take a look. Thanks for reviewing!

compose/cli/colors.py
```diff
@@ -1,8 +1,19 @@
...
+
 from ..const import IS_WINDOWS_PLATFORM
 
+
+@enum.unique
+class AnsiMode(enum.Enum):
+    """Enumeration for when to output ANSI colors."""
+    never = "never"
+    always = "always"
+    auto = "auto"
```
ulyssessouza: IMHO Something like the definition below would be better:
```python
@enum.unique
class AnsiMode(enum.Enum):
    """Enumeration for when to output ANSI colour's."""
    AUTO = "auto"
    NEVER = "never"
    ALWAYS = "always"

    def __init__(self, mode):
        self.mode = mode

    def use_tty(self, stream):
        return self is AnsiMode.ALWAYS or (self is AnsiMode.AUTO and stream.isatty())
```

Note the method `use_tty` that can be used after in the code
(Maybe that's not the best method name, feel free to change)

compose/cli/main.py
```diff
@@ -137,14 +151,13 @@ def setup_logging():
...
-        compose.parallel.ParallelStreamWriter.set_noansi()
+def setup_parallel_logger(ansi):
+    import compose.parallel
+    compose.parallel.ParallelStreamWriter.set_ansi(ansi)
 
 
-def setup_console_handler(handler, verbose, noansi=False, level=None):
-    if handler.stream.isatty() and noansi is False:
+def setup_console_handler(handler, verbose, ansi=AnsiMode.auto, level=None):
+    if ansi == AnsiMode.always or (ansi == AnsiMode.auto and handler.stream.isatty()):
```
ulyssessouza:
```suggestion
    if ansi.use_tty(handler.stream):
```

Or other method name defined in the enum

compose/parallel.py
```diff
@@ -314,10 +315,10 @@ def _write_noansi(self, msg, obj_index, status):
     def write(self, msg, obj_index, status, color_func):
         if msg is None:
             return
-        if self.noansi:
-            self._write_noansi(msg, obj_index, status)
-        else:
+        if self.ansi == AnsiMode.always or (self.ansi == AnsiMode.auto and self.stream.isatty()):
```
ulyssessouza:
```suggestion
        if self.ansi.use_tty(self.stream):
```
Or other method name defined in the enum

ulyssessouza:
Also this feature is missing tests, could you please add that @mikesep ?

compose/cli/main.py
```diff
@@ -195,6 +208,7 @@ class TopLevelCommand(object):
       --verbose                   Show more output
       --log-level LEVEL           Set log level (DEBUG, INFO, WARNING, ERROR, CRITICAL)
       --no-ansi                   Do not print ANSI control characters
+      --ansi (never|always|auto)  Control when to print ANSI control characters
```
thaJeztah: should the new option deprecate the old (`--no-ansi`) one? (at least making them conflicting options, as (I think) `--no-ansi` is the equivalent of `--ansi=never`?)

For the `docker` cli, if there's an option that cannot be removed we usually _hide_ the option (to discourage use), and (depending on the case) print a deprecation warning if used (but keep it functional if needed).

Also looking if we should align the UX to the `--progress` option used on `docker build`;

```
      --progress string         Set type of progress output (auto, plain, tty). Use plain to show container output (default "auto")
```

mikesep:
> Also this feature is missing tests, could you please add that @mikesep ?

@ulyssessouza can you clarify a little about what kind of tests you'd like to see? There were already tests that I adjusted that check for ANSI codes in the output. Do you want tests of the command-line option parsing, or also end-to-end that include the ANSI output? I also assume there isn't a good way to test tty detection in a Jenkins test, but if you have a way to do that, let me know.

compose/cli/main.py
mikesep: Hi @thaJeztah, thanks for your feedback!

`--no-ansi` is indeed the same as `--ansi=never`. In this code, `--no-ansi` overrides `--ansi`, but I could certainly make `--ansi=MODE` and `--no-ansi` mutually exclusive. I could also hide `--no-ansi` and emit a warning.

Alignment across CLIs is good. If `--progress=tty` means "always print ANSI color codes", I dislike that name for it. If that ship has already sailed, and you'd prefer that I follow the auto/plain/tty convention, I can do that.

ulyssessouza:
The test I'm referring to is the one where `ansi == AnsiMode.auto and handler.stream.isatty()`. This one is the new behavior introduced by this PR. You can achieve this by mocking the handler.

gapipro:
+1 for completing this.

mikesep:
Rebased and rewrote the code I had from almost a year ago! Code from before the force-push can still be found on [my fork's `ansi.2020-01-08` branch](https://github.com/mikesep/compose/tree/ansi.2020-01-08). Added tests for `compose/cli/colors.py`.

mikesep:
Whoops, need to look into some failing tests.

mikesep:
@ulyssessouza could you help me get this PR re-reviewed? At the end of last year I reworked the code to address concerns mentioned above and bring it up to date. Thanks!

aiordache (approved):
LGTM. Thank you!

PR #6875, opened 2019-08-28
[TAR-995] ADDED a stage for executing License Scans

zelahi:
Signed-off-by: Zuhayr Elahi <elahi.zuhayr@gmail.com>

We are in the process of integrating our license scans to occur pre-commit.  In this PR, we are adding a Jenkinsfile and a Makefile which includes `fossa` commands.  This is to execute a license scan on every commit to check for license violations

The job was created here: 
https://ci.docker.com/teams-dsg/blue/organizations/jenkins/fossa-compose/detail/fossa-compose/2/pipeline

The job ran and a report was uploaded here: https://app.fossa.io/projects/custom+11%2Fgit@github.com:docker%2Fcompose/refs/branch/null/0d061d84d5babeb3403cf4c23fe3c0f79c05002f

Currently it won't run on the same pipeline as the Jenkins that is public facing but we could move it there if the compose team would like us to do so. 

The fossa scans could be flaky due to issues on their end which we are unable to fix, as a result separating them is a nice way for us to handle this as the scans won't block people if they fail

### DIFF
--- a/script/Jenkinsfile.fossa
+++ b/script/Jenkinsfile.fossa
@@ -1,0 +1,20 @@
+pipeline {
+    agent any
+    stages {
+        stage("License Scan") {
+            agent {
+                label 'ubuntu-1604-aufs-edge'
+            }
+
+            steps {
+                withCredentials([
+                    string(credentialsId: 'fossa-api-key', variable: 'FOSSA_API_KEY')
+                ]) {
+                    checkout scm
+                    sh "FOSSA_API_KEY='${FOSSA_API_KEY}' BRANCH_NAME='${env.BRANCH_NAME}' make -f script/fossa.mk fossa-analyze"
+                    sh "FOSSA_API_KEY='${FOSSA_API_KEY}' make -f script/fossa.mk fossa-test"
+                }
+            }
+        }
+    }
+}
--- a/script/fossa.mk
+++ b/script/fossa.mk
@@ -1,0 +1,16 @@
+# Variables for Fossa
+BUILD_ANALYZER?=docker/fossa-analyzer
+FOSSA_OPTS?=--option all-tags:true --option allow-unresolved:true
+
+fossa-analyze:
+	docker run --rm -e FOSSA_API_KEY=$(FOSSA_API_KEY) \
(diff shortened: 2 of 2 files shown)


script/Jenkinsfile.fossa
```diff
@@ -0,0 +1,30 @@
+hubCred = [
+    $class: 'UsernamePasswordMultiBinding',
```
ndeloof: $class is legacy syntax to be avoided. 
Also group credential binding within a single `withCredentials` definition

prefer: 
```
withCredentials([
    usernamePassword(credentialsId: 'orcaeng-hub.docker.com', usernameVariable: 'REGISTRY_USERNAME', passwordVariable: 'REGISTRY_PASSWORD'),
    string(credentialsId: 'fossa-api-key', variable: 'FOSSA_API_KEY')
]) { ... }

File: script/fossa.mk [hunk 2]
```diff
@@ -0,0 +1,17 @@
+# Variables for Fossa
+BUILD_ANALYZER?=dockereng/fossa-analyzer
+FOSSA_OPTS?=--option all-tags:true --option allow-unresolved:true
+
+fossa-analyze:
+	docker run --rm -e FOSSA_API_KEY=$(FOSSA_API_KEY) \
+		-e GO111MODULE=on \
```
thaJeztah: This repository doesn't use Golang, so we probably don't need this line

script/fossa.mk
```diff
@@ -0,0 +1,17 @@
+# Variables for Fossa
+BUILD_ANALYZER?=dockereng/fossa-analyzer
+FOSSA_OPTS?=--option all-tags:true --option allow-unresolved:true
```
thaJeztah: Does fossa need a "strategy" set for python? https://github.com/fossas/fossa-cli/blob/master/docs/integrations/python.md#python

Jenkinsfile
```diff
@@ -91,5 +91,4 @@ baseImages.each { baseImage ->
     }
   }
 }
-
```
thaJeztah: nit; unrelated change now

PR #6887, opened 2019-09-06
Add --target option to build command

### LINKED ISSUE #6886: Allow --target option in docker-compose build
I have some multi-stage builds in my project and I would like to be able to run some containers at a particular stage for testing. Currently, I can't use Compose for these - I have to run the tests for one of my services using plain docker.

The `docker build` option `--target` can be used with Compose.

I could use separate `docker-compose.yaml` files for testing, development, and production, but multi-stage builds are supposed to help with the problem of coordinating multiple files with the same information in them, so this isn't ideal.

samueljsb:
Allow a target to be passed to the build command for multi-stage builds.

Fixes: #6886

Signed-off-by: Samuel Searles-Bryant <samuel.searles-bryant@unipart.io>

Resolves #6886

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -275,6 +275,7 @@ class TopLevelCommand(object):
                                     To enable, run with COMPOSE_DOCKER_CLI_BUILD=1)
             --pull                  Always attempt to pull a newer version of the image.
             -q, --quiet             Don't print anything to STDOUT
+            --target string         Set the target build stage to build
         """
         service_names = options['SERVICE']
         build_args = options.get('--build-arg', None)
@@ -301,6 +302,7 @@ class TopLevelCommand(object):
             silent=options.get('--quiet', False),
             cli=native_builder,
             progress=options.get('--progress'),
+            target=options.get('--target'),
         )
 
     def bundle(self, options):
--- a/compose/project.py
+++ b/compose/project.py
@@ -356,7 +356,7 @@ class Project(object):
 
     def build(self, service_names=None, no_cache=False, pull=False, force_rm=False, memory=None,
               build_args=None, gzip=False, parallel_build=False, rm=True, silent=False, cli=False,
-              progress=None):
+              progress=None, target=None):
 
         services = []
         for service in self.get_services(service_names):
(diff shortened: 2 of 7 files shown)


ndeloof:
It seems to me for compose-file reproducibility, if we agree to support build targets then target definition should be part of the compose file `build` section, not a command line option like suggested. 
Do you maybe have a specific scenario in mind to explain this design decision?

samueljsb:
The specific scenario I had in mind was a multi-stage Dockerfile with `FROM npm as build` to build a service and `FROM nginx as deploy` to serve it. To run the tests as a one-off 'build then run' I would need to run a container from the first `build` stage, but in the `docker-compose.yml` I want the `deploy` target for actually deploying.

Essentially I want to use all of the other information from the `docker-compose.yml` but override this flag for a one-off build for testing.

samueljsb:
On reflection, I think the appropriate solution to my stated case was to have a separate compose file for testing. I'm going to close this PR, but feel free to reopen it if you feel it's something that should actually happen.

maverick1872:
To me this seems like a sensible addition and one I'd also personally like to see added. Given that the `--target` option is available via `docker build` it seems reasonable to me that a passthrough via the compose variant would be supported.

thaJeztah:
@maverick1872 the difference here is that `docker build` builds a single Dockerfile, but `docker compose build` can build multiple services, each of which may (and likely _will_) have their own target; the compose-file should be the source of truth for those (and define the options that need to be set to build the image for a service).

ndeloof:
Options available to the `docker build` command should have a matching attribute in the compose file format, but this doesn't mean a `docker compose` flag should be introduced. As @thaJeztah explained, such a flag is only relevant if it applies to all images OR it applies to docker compose logic. For sure, there's a few exception to this rule for legacy reasons

maverick1872:
@thaJeztah Ahh fair point. Hadn't considered the implications of multiple services defined and how that would be interpreted if a target flag were provided. I suppose I'm coming from the perspective where I'm leveraging the compose definition to easily rebuild and run a specific service out of the compose spec with a different target. In short the availability of the flag when specifying a specific service is my use case. That gets messy from a support/implementation perspective I'd imagine. Thanks for the additional insight!

maverick1872:
@thaJeztah Ahh fair point. Hadn't considered the implications of multiple services defined and how that would be interpreted if a target flag were provided. I suppose I'm coming from the perspective where I'm leveraging the compose definition to easily rebuild and run a specific service out of the compose spec with a different target. In short the availability of the flag when specifying a specific service is my use case. That gets messy from a support/implementation perspective I'd imagine. Thanks for the additional insight!

alliefitter:
Sorry for necroing this thread, but I found a pretty simple solution for this, and wanted to share it in case anyone else needs it.

```
$ cat docker-compose.yml | sed 's/target: production/target: development/' | docker compose -f - up --build
```

PR #6900, opened 2019-09-18
Include errors when raising ProjectError exception

### LINKED ISSUE #6899: Encountered errors while bringing up the project hides real errors
Apparently using docker compose python API hides essential error information from the (api) a user, making close to impossible to debug.

Usually user endsup with:
```
  File "/usr/lib/python2.7/site-packages/compose/project.py", line 559, in up
    'Encountered errors while bringing up the project.'
```

The bug is at https://github.com/docker/compose/blob/master/compose/project.py#L571-L581
where is can easily be seen that the `errors` information is not included in the error, nor is returned.

ssbarnea:
Avoids censoring of errors that cause the project failure by including
them with the raised exception.

Resolves #6899

### DIFF
--- a/compose/project.py
+++ b/compose/project.py
@@ -840,6 +840,6 @@ class ProjectError(Exception):
 
     def __str__(self):
         if self.value:
-            return self.msg + "\n" + "\n".join(self.value)
+            return "\n".join([self.msg] + self.value)
         else:
             return self.msg
--- a/compose/project.py
+++ b/compose/project.py
@@ -570,7 +570,8 @@ class Project(object):
         )
         if errors:
             raise ProjectError(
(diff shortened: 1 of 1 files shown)


compose/project.py
```diff
@@ -833,5 +834,12 @@ def __str__(self):
...
 
 class ProjectError(Exception):
-    def __init__(self, msg):
+    def __init__(self, msg, errors=None):
         self.msg = msg
+        self.value = errors
+
+    def __str__(self):
+        if self.value:
+            return self.msg + "\n" + "\n".join(self.value)
```
ulyssessouza:
```suggestion
            return "\n".join([self.msg] + self.value)
```

ssbarnea:
RIP!

PR #6905, opened 2019-09-20
Show uptime docker compose

### LINKED ISSUE #6867: docker-compose ps should show uptime

I would like to be able to see when the last time was that a docker container was restarted/updated. Similar to #4577, but my main concern is seeing the uptime of the container.

```
docker-compose ps
```

shows the time since the container was launched.

Right now I use:

```
docker ps -a
``` 

instead, which means I get a sea of output that may not be relevant, especially if I am running multiple docker-compose setups locally.

zelahi:
Resolves #6867 

**Problem Statement:**
In this PR, we are reporting the output when the user executes `docker-compose ps` after launching containers via `docker-compose up`

**Expected Output**
Screen Shot 2019-09-20 at 2 45 21 PM

**Implementation Details**
I currently used the `dateutils` library and retrieved the uptimes from the container properties.  After doing so, parsed them and displayed it in a human readable format.  I added one unit test, but had trouble writing an unit test with the code path using `datetime.now()`.  I didn't see any mocks, but I am open to suggestions and/or okay with reimplementing this without using `pyzar` and/or `dateutils`

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -742,6 +742,7 @@ class TopLevelCommand(object):
                 'Command',
                 'State',
                 'Ports',
+                'Started At',
             ]
             rows = []
             for container in containers:
@@ -753,6 +754,7 @@ class TopLevelCommand(object):
                     command,
                     container.human_readable_state,
                     container.human_readable_ports,
+                    container.human_readable_start_time,
                 ])
             print(Formatter().table(headers, rows))
 
@@ -883,10 +885,12 @@ class TopLevelCommand(object):
             )
 
         if options['COMMAND'] is not None:
+            print("ASDF - options['COMMAND']: ", options['COMMAND'])
             command = [options['COMMAND']] + options['ARGS']
         elif options['--entrypoint'] is not None:
             command = []
         else:
+            print("ASDF - getting command from service")
             command = service.options.get('command')
 
         options['stdin_open'] = service.options.get('stdin_open', True)
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -742,7 +742,7 @@ class TopLevelCommand(object):
                 'Command',
                 'State',
                 'Ports',
-                'Started At',
+                'Uptime',
             ]
             rows = []
             for container in containers:
@@ -754,7 +754,7 @@ class TopLevelCommand(object):
                     command,
                     container.human_readable_state,
                     container.human_readable_ports,
-                    container.human_readable_start_time,
+                    container.human_readable_uptime,
                 ])
             print(Formatter().table(headers, rows))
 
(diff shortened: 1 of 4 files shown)


compose/cli/main.py
```diff
@@ -753,6 +754,7 @@ def ps(self, options):
                     command,
                     container.human_readable_state,
                     container.human_readable_ports,
+                    container.human_readable_uptime,
```
thaJeztah: Wondering why you didn't use the `Status` field that's returned by the API (which is the field that's used by the `docker` CLI;

```bash
curl --unix-socket /var/run/docker.sock "http://localhost/containers/json" | jq .
```

```json

  {
    "Id": "82950c6535204a462c8a3c1f175408b456fbb91971d2d21a33ba04c8b91c74fd",
    "Names": [
      "/cranky_keldysh"
    ],
    "Image": "libnetworkbuild",
    "ImageID": "sha256:bc6bbc6a0032300d8818182eff0101ecb7af2fc1fb21da6f290943c286946a1e",
    "Command": "make unit-tests-local",
    "Created": 1573091888,
    "Ports": [],
    "Labels": {},
    "State": "running",
    "Status": "Up 5 minutes",
    "HostConfig": {
      "NetworkMode": "default"
    },
    "NetworkSettings": {
      "Networks": {
        "bridge": {
          "IPAMConfig": null,
          "Links": null,
          "Aliases": null,
          "NetworkID": "5a59d43d598f910579535ffb2cbbb4d0987807d7b5593c264c83337c4220ec1a",
          "EndpointID": "bf8b6b629b353988d047bcd0bbb897249d7a73a7811b2210b783338e6a975cd9",
          "Gateway": "172.17.0.1",
          "IPAddress": "172.17.0.5",
          "IPPrefixLen": 16,
          "IPv6Gateway": "",
          "GlobalIPv6Address": "",
          "GlobalIPv6PrefixLen": 0,
          "MacAddress": "02:42:ac:11:00:05",
          "DriverOpts": null
        }
      }
    },
    "Mounts": [
      {
        "Type": "bind",
        "Source": "/Users/sebastiaan/projects/libnetwork",
        "Destination": "/go/src/github.com/docker/libnetwork",
        "Mode": "",
        "RW": true,
        "Propagation": "rprivate"
      }
    ]
  }
]
```

ulyssessouza: I agree with this. Also, if it's not running, there is no `Uptime`.
Changing the `State` column to `Status` would do the job.

The only thing here is the people that could be parsing this output, but since it's not a guaranteed format, we should be fine.

zelahi: Ahh I believe I just missed it.  I didn't use the curl command and actually used the values I saw when I did a `docker inspect <container_name>`.  I'll go ahead and fix it so we use Status instead.  Thanks for the feedback!! =)

glours:

Issue #6831, opened 2019-08-02
Release docker/compose image for armv7 / arm64v8


maruel:
**Is your feature request related to a problem? Please describe.**

Using `curl -L "https://github.com/docker/compose/releases/download/1.24.1/docker-compose-$(uname -s)-$(uname -m)" > docker-compose`
fails on Raspbian as described in issue #6810.

The root causes are:

- https://github.com/docker/compose/releases only contain releases for amd64 / x64-64.
- https://hub.docker.com/r/docker/compose is not multi-arch enabled.

While issue #6810 was closed, I think this is a genuine feature request since instructions at https://docs.docker.com/compose/install/#install-as-a-container are not clear about which platforms are _officially_ supported to run docker-compose as a container.

Two recent events may now make this worth:

- The new RPi4 has real 1Gbit ethernet. It becomes a real platform for hobbyist with limited linux experience using Raspbian to host images. 
- Docker Hub has real native support for multi-arch images via [manifest v2](https://docs.docker.com/registry/spec/manifest-v2-2/).

Combined with the fact that docker-compose is a *core building block for orchestrating images*, the lack of support on arm v7 (32 bits) and v8 (64 bits) makes the bootstrapping process options more limited than on x86 equivalent and the documentation is not clear about which options work where.

As stated at https://github.com/docker/compose/issues/4733#issuecomment-294021065, there _used_ to be a Dockerfile.armhf. This file was deleted in #6666 by @joakimr-axis since the current one works just fine.

**Describe the solution you'd like**

Officially create armv7 and arm64v8 docker images to https://hub.docker.com/r/docker/compose as part of the release process.

This is essentially a process change, not a code change, since the Dockerfile is known to work on arm already. This would involve at the very least updating [.circleci/config.yml](https://github.com/docker/compose/blob/master/.circleci/config.yml).

**Describe alternatives you've considered**

Installing via `pip` works in the meantime but is trickier to run this process successfully for new comers with limited linux experience, especially that the documentation is misleading about what is supported and what is not. This leads to a poor user experience (UX).

webash:
As a new user to Docker, hoping to make my Raspberry Pi 4 the test ground for learning Docker, this stumbling block has been impossible to get past. Every tutorial uses docker-compose and the fact that it seems impossible to obtain easily has wasted hours of my time.

There is no clear reason why this isn't the case. It should at least be called out in the installation documentation in the meantime.

nemchik:
[...]

PR #6908, opened 2019-09-23
Add dependencies for ARM build

StefanScherer:
This PR adds missing dependencies if someone tries to build docker-compose on an ARM device like the Raspberry Pi. It seems these dependencies are implicit installed while building on amd64, but need to be installed on arm.

Prerequisites
- a Raspberry Pi
- have Git and Docker installed

How to verify
1. Login to your Raspberry Pi
2. `git clone https://github.com/docker/compose`
3. `cd compose`
4. `scripts/build/linux`
5. `dist/docker-compose-Linux-armv7l --version`

Helps #6831

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -30,15 +30,18 @@ RUN apk add --no-cache \
 ENV BUILD_BOOTLOADER=1
 
 FROM python:${PYTHON_VERSION}-${BUILD_DEBIAN_VERSION} AS build-debian
-RUN apt-get update && apt-get install -y \
+RUN apt-get update && apt-get install --no-install-recommends -y \
     curl \
     gcc \
     git \
     libc-dev \
+    libffi-dev \
     libgcc-6-dev \
+    libssl-dev \
     make \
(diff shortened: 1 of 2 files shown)


NerdsvilleCEO:
Will this mean there will be armv8 and armv7 images posted to dockerhub?

Dockerfile
```diff
@@ -35,10 +35,13 @@ RUN apt-get update && apt-get install -y \
     gcc \
```
thaJeztah: might be good to set the `--no-install-recommends` flag, and be specific about the exact dependencies to install (which _could_ explain the difference

ndeloof:
@StefanScherer can you please address Sebastiaan's comment? Can then be merged

Dockerfile
StefanScherer: Added the flag, thanks.

tiero:
Tried on Odroid H2 armv7 ubuntu 16. Worked flawlessly! 🎉

PR #6911, opened 2019-09-23
Error when Docker not found should be "Docker Desktop"

ChrisParkerGMSL:
Fixes a minor issue with an error message. "Docker for Windows" is now called "Docker Desktop".


ChrisParkerGMSL:
Give up start again. 🙂

PR #6917, opened 2019-09-25
Display successful message when dump validated config

### LINKED ISSUE #6916: [config subcmd] Display a message that configuration validation was successful
In version 1.24.1, executes config subcommand and then displays the configuration content of docker-compose.yml.
But the message doesn't include a message indicating validation was successful.
I don't know if configuration validation succeeded or failed.
So, it is better to display like "PASS: Validate the compose file".

Make a few changes on [compose/cli/main.py#L367](https://github.com/docker/compose/blob/1.24.1/compose/cli/main.py#L367):

```python
- print(serialize_config(compose_config, image_digests))

+ successful_msg = 'PASS: Validate the compose file'
+ print('{}\n{}'.format(successful_msg, serialize_config(compose_config, image_digests)))
```

Thanks

homoluctus:
Signed-off-by: homoluctus <w.slife18sy@gmail.com>

Resolves #6916

I implemented as following:

```
$ docker-compose config
PASS: Validate the compose file
services:
  app:
    build: .
    container_name: test
    image: test
    tty: true
version: '3.7'
```

### DIFF
--- a/tests/acceptance/cli_test.py
+++ b/tests/acceptance/cli_test.py
@@ -43,6 +43,7 @@ ProcessResult = namedtuple('ProcessResult', 'stdout stderr')
 
 BUILD_CACHE_TEXT = 'Using cache'
 BUILD_PULL_TEXT = 'Status: Image is up to date for busybox:1.27.2'
+PASS_VALIDATION_TEXT = 'PASS: Validate the compose file'
 
 
 def start_process(base_dir, options):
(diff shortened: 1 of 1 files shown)


ulyssessouza:
Hello @homoluctus thank you for the interest in docker-compose.
I think this is not a good approach, this is a breaking change that IMO doesn't add any feature. 
If you have the configuration yaml printed as output, that means the configuration is valid already, otherwise you will have an error and the error/exit code should be different of 0.

By adding what you suggest would break any script using this output.
Maybe if you share your use case I can help you to find an alternative solution.

PR #6918, opened 2019-09-25
Add env vars alternative substitution interpolation

zorn-v:
... | VAR is set and not null | VAR is set and null | VAR is unset
--- | ---------------------------- | ------------------------ | ----------------
`${VAR:+word}` | substitute word | substitute null | substitute null
`${VAR+word}` | substitute word | substitute word | substitute null

Maybe it not so much use cases for this but may be useful.

Consider nginx image with following config with ato redirect http to https if connect on 80 port
```nginx
    listen 80;
    listen 443 ssl http2;
    listen 8000;
...
    if ($server_port = 80) {
         return 301 https://$host$request_uri;
    }
```
On some containers you do not need that redirect, but don't want to use something like `HTTP_CONTAINER_PORT=8000` but `NO_HTTPS_REDIRECT=1`
Then you can use following port mapping
```yaml
ports:
  - 80:80${NO_HTTPS_REDIRECT+00}
```

### DIFF
--- a/compose/config/interpolation.py
+++ b/compose/config/interpolation.py
@@ -100,7 +100,7 @@ class TemplateWithDefaults(Template):
         """ % {
         'delim': re.escape('$'),
         'id': r'[_a-z][_a-z0-9]*',
-        'bid': r'[_a-z][_a-z0-9]*(?:(?P<sep>:?[-?])[^}]*)?',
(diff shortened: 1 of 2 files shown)


ndeloof:
Sounds to me this is a reasonable feature, I'm a bit concerned there's not realistic use-case driving it, and this will introduce more complexity to newcomers who might not be confident with bash-style parameter interpolation.

Also will need to be implemented in a consistent way in https://github.com/docker/cli/tree/master/cli/compose/interpolation

PR #6923, opened 2019-09-26
Added support for quoted env vars

### LINKED ISSUE #2854: Inconsistent env var parsing from docker-compose.yml
If I declare my env vars in bash, I get the following:

```
10:07 $ export TESTVAR=test
10:08 $ echo $TESTVAR
test
10:08 $ export TESTVAR="test"
10:08 $ echo $TESTVAR
test
```

However, if I create the following files:
Dockerfile:

```
FROM ubuntu:latest
COPY ./test.sh /test/test.sh
WORKDIR /test
ENTRYPOINT [ "/test/test.sh" ]
```

test.sh:

```
#!/bin/bash
echo "--------------TEST RUN---------------"
echo $TESTVAR

case $TESTVAR in
    test)
        echo "no quotes"
    ;;
esac
echo "------------END TEST RUN-------------"
```

docker-compose.yml

```
testthing:
    dockerfile: Dockerfile
    build: .
    environment:
        - TESTVAR="test"
```

I get the following output:

```
10:07 $ docker-compose up
Recreating compose_testthing_1
Attaching to compose_testthing_1
testthing_1 | --------------TEST RUN---------------
testthing_1 | "test"
testthing_1 | ------------END TEST RUN-------------
compose_testthing_1 exited with code 0
```

Changing the environment variable to `TESTVAR=test` produces:

```
10:07 $ docker-compose up
Creating compose_testthing_1
Attaching to compose_testthing_1
testthing_1 | --------------TEST RUN---------------
testthing_1 | test
testthing_1 | no quotes
testthing_1 | ------------END TEST RUN-------------
compose_testthing_1 exited with code 0
```

For the record, docker itself appears to handle the quotes correctly (`test:latest` is the same Dockerfile above being run directly):

```
10:11 $ docker run -it --rm -e TESTVAR="test" test:latest 
--------------TEST RUN---------------
test
no quotes
------------END TEST RUN-------------
```

Also, compose behaves correctly when using the map syntax, i.e. `TESTVAR: "test"` and `TESTVAR: test` behave the same way, and consistently with bash (`no quotes` is printed).

kamermans:
Signed-off-by: Steve Kamerman <stevekamerman@gmail.com>

Resolves #2854

This PR adds support for environment variables in `.env` files with quoted values.  The quotes can be single `'` or double `"`, but must be the same on both ends of the value, and must immediately follow the `=`.

Example:
```
MYVAR="this is a test"
```

It also unescapes escaped quotes, like `\"`:
```
MYVAR="this is a \"quoted\" test"
```
Value: `this is a "quoted" test`

All functionality is tested.

This problem has been thoroughly discussed in issue #2854 and there are many reasons for why docker-compose should support this.

### DIFF
--- a/compose/config/environment.py
+++ b/compose/config/environment.py
@@ -22,6 +22,12 @@ def split_env(env):
     key = value = None
     if '=' in env:
         key, value = env.split('=', 1)
+        # If the value is quoted, remove the quotes
+        if len(value) >= 2 and value[0] == value[-1] and value[0] in ['"', "'"]:
+            value = value[1:-1]
+            # unescape quoted values
+            quote = value[0]
+            value.replace("\\"+quote, quote)
     else:
         key = env
     if re.search(r'\s', key):
--- a/tests/unit/config/environment_test.py
+++ b/tests/unit/config/environment_test.py
@@ -62,3 +62,14 @@ class EnvironmentTest(unittest.TestCase):
         with pytest.raises(ConfigurationError) as exc:
             env_vars_from_file(str(tmpdir.join('whitespace.env')))
         assert 'environment variable' in exc.exconly()
+
+    def test_env_vars_from_file_quoted(self):
+        tmpdir = pytest.ensuretemp('env_file')
+        self.addCleanup(tmpdir.remove)
+        with codecs.open('{}/quoted.env'.format(str(tmpdir)), 'w', encoding='utf-8') as f:
+            f.write('DOUBLE_QUOTES="testing \"double\" quotes"\n')
+            f.write("SINGLE_QUOTES='testing \'single\' quotes'\n")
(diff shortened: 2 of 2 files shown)


kamermans:
I understand that this change will also [affect the documentation](https://docs.docker.com/compose/env-file/), so I'm happy to put in a PR for that as well.

kolaente:
Ping @ulyssessouza @ndeloof @rumpl Any chances this will be merged?

ulyssessouza:
The PR was closed in favor of the adoption of `python-dotenv` that solves the problem in a more general way. But anyways, thank you @kamermans !

feld:
I don't know how `python-dotenv` solves this issue that remains as seen in https://github.com/docker/compose/issues/2854#issuecomment-1287173270  ???

Can anyone clarify?

Issue #6936, opened 2019-10-03
Rebuild container if specified files has changed


kindritskyiMax:
Hi. The main idea of this feature request is the ability to specify a list of files on which image depends and rebuilds automatically.

It can be really helpful to just declaratively tell a service, when I want it to rebuild

How it can be used ?
For example we have a simple `nodejs` service.

```
webapp:
    image: webapp-dev
    build:
      dockerfile: ./docker/Dockerfile.webapp-dev
      context: .
    working_dir: /work
```

We can add something like `hash` or `checksum` or you name it.
 
```
webapp:
    image: webapp-dev
    build:
      dockerfile: ./docker/Dockerfile.webapp-dev
      checksum:
      - ./package.json
      - ./package-lock.json
      - ./docker/Dockerfile.webapp-dev
      context: .
    working_dir: /work
```

And now when one of these files in checksum changes, `docker-compose up webapp` will automatically rebuild an image.

This requires a place where docker-compose can store checksums for images and check those against newly calculated ones.

Also, it is a good idea, IMHO, to new command called checksum which calculates and prints the same checksum as it will be used to make the decision to rebuild an image

fabiosantoscode:
I really like this idea. However I find it unlikely that it will go in.

As a workaround, you can do some stuff around the npm `nodemon` package. It doesn't only run JS, as with the `--exec` argument you can do anything with it, including asking docker-compose to rebuild.

kindritskyiMax:
Well, yes it can be achieved by wrapping docker-compose in some kind of entry point script where the entry point is responsible for calculating checksums and rebuilding images but its more convenient to have such functionality built into docker-compose. I must say it's not only related to nodejs. Lets say puthon has its onw requirements, go and so on.

ndeloof:
Actually what you suggest could be better implemented relying on BuildKit as the build engine, as this one do offer very efficient caching and will be able to "rebuild" your image within a second if no change took place on source tree.
Now that we have option to use CLI builder with `COMPOSE_DOCKER_CLI_BUILD` this sounds feasible. But you would still have to stop/restart your compose stack. If you are looking for an shorter cycle, then you should better look into bind mounting project sources into container and configure your stack to live-reload when source get updated. This is something well supported by nodejs, might be harder to implement with other languages.

ndeloof:
Sounds like a resurection of https://github.com/docker/compose/issues/184

ndeloof:
[...]

PR #6938, opened 2019-10-04
Add checksum feature

kindritskyiMax:
This PR is intended to add a feature called `checksum` (may be changed).

It is a WIP (test will be written after we make sure the implementation is correct)

This is how it works.

1. add to `checksum` list of files

```
webapp:
    image: webapp-dev
    build:
      dockerfile: ./docker/Dockerfile.webapp-dev
      checksum:
      - ./package.json
      - ./package-lock.json
      - ./docker/Dockerfile.webapp-dev
      context: .
    working_dir: /work
```

2. If any of those files changed, rebuild the image

For now, I create file in `~/.docker/checksums.json`. This place is chosen because I did not found any better place. If you know where is the better place - I will be glad to hear.

I am calculating the checksum for image and write it to `checksums.json`. Then I recalculate checksum on every run, and if something changed, update checksum and rebuild the image.

UPD. also i've added new command called `checksum` which calculates and prints same checksum as it will be used to make the decision to rebuild an image

usage: 
`docker-compose checksum my_service`

### DIFF
--- a/compose/cli/command.py
+++ b/compose/cli/command.py
@@ -17,6 +17,7 @@ from ..project import Project
 from .docker_client import docker_client
 from .docker_client import get_tls_version
 from .docker_client import tls_config_from_options
+from .docker_client import get_checksums_path
 from .utils import get_version_info
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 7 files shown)


kindritskyiMax:
https://github.com/docker/compose/issues/6936 my issue

PR #6948, opened 2019-10-08
Log warning for docker-compose port binding not avaliable

### LINKED ISSUE #6708: Static check for duplicate port request in compose file
When a pod specification requests the same port twice(for example because of a change in a .env file), the "docker-compose run <service-name>" command responds with "Bind for 0.0.0.0:<Port> failed: port is already allocated" instead of responding that the problem is in the compose file.

zelahi:
Signed-off-by: Zuhayr Elahi <elahi.zuhayr@gmail.com>

Resolves #6708 

I added a warning at the bottom to let users know to check the compose file if this error occurs.  It seems to only happen when the user runs `docker-compose up` but I believe this could be potentially how we solve this issue above

### DIFF
--- a/compose/service.py
+++ b/compose/service.py
@@ -624,9 +624,7 @@ class Service(object):
             container.start()
         except APIError as ex:
             if "driver failed programming external connectivity" in ex.explanation:
-                log.warn(
-                    "Port is already in use, check the docker-compose file to see if the same" +
-                    " port was allocated to multiple services")
+                log.warn("Host is already in use by another container")
             raise OperationFailedError("Cannot start service %s: %s" % (self.name, ex.explanation))
         return container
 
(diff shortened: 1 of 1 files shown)


compose/service.py
```diff
@@ -623,6 +623,10 @@ def start_container(self, container, use_network_aliases=True):
         try:
             container.start()
         except APIError as ex:
+            if "driver failed programming external connectivity" in ex.explanation:
+                log.warn(
+                    "Port is already in use, check the docker-compose file to see if the same" +
```
ndeloof: I think that's a good think to check for this specific scenario, but the message here is incorrect : port collision may be the result of any other container running on same host, and not under control of this specific compose file.

zelahi: Ahh, okay, I can update the message.  I was worried a bit about the approach I was taking here

ndeloof: I don't know if this is the best approach, unfortunately I guess there's no other option

Issue #4611, opened 2017-03-12
Enable output of `depends_on` services


whitecolor:
When `docker-compose up my-service` is run it will also start services that `my-service` `depends_on`, but it won't show output of those services.

It would be useful to have option for `docker-compose` that would also turn on output for dependencies.

shin-:
You can use `docker-compose logs` to do exactly that, I think.

whitecolor:
you mean start for example detached `docker-compose up -d my-service` and then `docker-compose logs here i still need to enumerate all the services by name`

rawkode:
@whitecolor `docker-compose logs -f` will show logs of all services

For me, when I do `docker-compose up app` - I only want to see app logs, not dependant services.

I guess this is a pretty subjective issue :+1:

whitecolor:
well actually really `docker-compose logs -f service1 service2` may do the trick I need, I believe this request/question maybe closed.

gnarea:
I would've loved to see this feature, and the proposed workaround is a bit of a hassle.

I've just implemented this functionality in [docker-dev](https://github.com/2degrees/docker-dev#docker-dev-up2-services). So, @whitecolor's example becomes: `docker-dev up2 my-service`.

whitecolor:
I think any decent developer/team should have (or already has) their own meta-framework on top of used infrastructure tools like a docker/compose. It is hard to invent one-fit-all-solution to automate everyone's workflows.

sarink:
How would you add the necessary `docker-compose logs` command to something like this? `docker-compose -f ci.config.yml up --exit-code-from test-runner test-runner`

PR #6950, opened 2019-10-09
Add "--attach-dependencies" to command "up" for attaching to dependencies

benthorner:
When using the 'up' command, only services listed as arguments are
attached to, which can be very different to the 'no argument' case
if a service has many and deep dependencies:

   - It's not clear when dependencies have failed to start. Have to run
'compose ps' separately to find out.
   - It's not clear when dependencies are erroring. Have to run 'compose
logs' separately to find out.

With a simple setup, it's possible to work around theses issue by
using the 'up' command without arguments. But when there are lots of
'top-level' services, with common dependencies, in a single config,
using 'up' without arguments isn't practical due to resource limits
and the sheer volume of output from other services.

This introduces a new '--attach-dependencies' flag to optionally attach
dependent containers as part of the 'up' command. This makes their logs
visible in the output, alongside the listed services. It also means we
benefit from the '--abort-on-container-exit' behaviour when dependencies
fail to start, giving more visibility of the failure.

Resolves https://github.com/docker/compose/issues/4611

Extends https://github.com/docker/compose/pull/2130 to treat `up <service>`, which implicitly includes dependent 
containers, the same way as `up`, which refers to all services/containers.

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -1012,6 +1012,7 @@ class TopLevelCommand(object):
             --build                    Build images before starting containers.
             --abort-on-container-exit  Stops all containers if any container was
                                        stopped. Incompatible with -d.
+            --attach-dependencies      Attach to dependent containers
             -t, --timeout TIMEOUT      Use this timeout in seconds for container
                                        shutdown when attached or when containers are
                                        already running. (default: 10)
@@ -1033,16 +1034,18 @@ class TopLevelCommand(object):
         remove_orphans = options['--remove-orphans']
         detached = options.get('--detach')
         no_start = options.get('--no-start')
+        attach_dependencies = options.get('--attach-dependencies')
 
-        if detached and (cascade_stop or exit_value_from):
-            raise UserError("--abort-on-container-exit and -d cannot be combined.")
+        if detached and (cascade_stop or exit_value_from or attach_dependencies):
+            raise UserError(
+                "-d cannot be combined with --abort-on-container-exit or --attach-dependencies.")
 
         ignore_orphans = self.toplevel_environment.get_boolean('COMPOSE_IGNORE_ORPHANS')
 
         if ignore_orphans and remove_orphans:
             raise UserError("COMPOSE_IGNORE_ORPHANS and --remove-orphans cannot be combined.")
 
-        opts = ['--detach', '--abort-on-container-exit', '--exit-code-from']
+        opts = ['--detach', '--abort-on-container-exit', '--exit-code-from', '--attach-dependencies']
         for excluded in [x for x in opts if options.get(x) and no_start]:
             raise UserError('--no-start and {} cannot be combined.'.format(excluded))
 
@@ -1087,7 +1090,10 @@ class TopLevelCommand(object):
             if detached or no_start:
                 return
 
-            attached_containers = filter_containers_to_service_names(to_attach, service_names)
+            attached_containers = filter_attached_containers(
+                to_attach,
+                service_names,
+                attach_dependencies)
 
             log_printer = log_printer_from_project(
                 self.project,
@@ -1392,8 +1398,8 @@ def log_printer_from_project(
         log_args=log_args)
 
 
-def filter_containers_to_service_names(containers, service_names):
-    if not service_names:
+def filter_attached_containers(containers, service_names, attach_dependencies=False):
+    if attach_dependencies or not service_names:
         return containers
 
     return [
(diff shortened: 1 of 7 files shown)


benthorner:
Just wondering if I need to do something to get this reviewed/merged?

compose/cli/main.py
```diff
@@ -1065,9 +1066,11 @@ def up(self, options):
         remove_orphans = options['--remove-orphans']
         detached = options.get('--detach')
         no_start = options.get('--no-start')
+        attach_dependencies = options.get('--attach-dependencies')
 
         if detached and (cascade_stop or exit_value_from):
```
ulyssessouza: Here, shouldn't you test with `attach_dependencies` in the `or` section?

compose/cli/main.py
```diff
@@ -1065,9 +1066,11 @@ def up(self, options):
...
-            raise UserError("--abort-on-container-exit and -d cannot be combined.")
+            raise UserError(
+                "-d cannot be combined with --abort-on-container-exit or -attach-dependencies.")
```
ulyssessouza:
```suggestion
                "-d cannot be combined with --abort-on-container-exit or --attach-dependencies.")
```

compose/cli/main.py
```diff
@@ -1098,6 +1101,7 @@ def up(rebuild):
                     always_recreate_deps=always_recreate_deps,
                     reset_container_image=rebuild,
                     renew_anonymous_volumes=options.get('--renew-anon-volumes'),
+
```
ulyssessouza: nit: Remove new line

compose/cli/main.py
benthorner: 👍 - fixed. I've also added it in a subsequent section, as it's not compatible with `--no-start`.

compose/cli/main.py
benthorner: 👍 - updated.

compose/cli/main.py
```diff
@@ -1392,10 +1398,13 @@ def log_printer_from_project(
         log_args=log_args)
 
 
-def filter_containers_to_service_names(containers, service_names):
+def filter_attached_containers(containers, service_names, attach_dependencies):
     if not service_names:
         return containers
 
+    if attach_dependencies:
```
ulyssessouza: Since the content of this `if` is equal to the preceding one, please, just add an `or` to the first one

compose/cli/main.py
```diff
@@ -1392,10 +1398,13 @@ def log_printer_from_project(
         log_args=log_args)
 
 
-def filter_containers_to_service_names(containers, service_names):
+def filter_attached_containers(containers, service_names, attach_dependencies):
```
ulyssessouza:
```suggestion
def filter_attached_containers(containers, service_names, attach_dependencies=False):
```
By using a default value we can keep the same usage.
With this you can simplify the calls in the tests that are not related to the PR.

benthorner: Good suggestion - updated 👍 .

compose/cli/main.py
benthorner: Sure - I've done that.

ulyssessouza:
@benthorner  You also need to add the new option to the [shell auto completion scripts](https://github.com/docker/compose/tree/master/contrib/completion). After that I think it's complete

benthorner:
@ulyssessouza thanks for pointer - I completely missed that before. Should be good for another look now 👀 .

ulyssessouza:
Thank you for the contribution @benthorner !

thaJeztah:
LOL, was just about to comment: wondering why this should not be the default?

```bash
# starts all containers and attaches to them
docker-compose up

# start only service-a and service-b, including their dependencies, and attaches to them
docker-compose up <service-a> <service-b>
```

I see the original issue mentions some people only being interested in the output of the service, not their dependencies (which is inconsistent with `docker-compose up` (all services)).

For those, it would be possible to start the services _detached_

```bash
docker-compose up -d <service-a> <service-b>
```

And attach to the logs of those services that they want to view the logs for

```bash
docker-compose logs -f <service-a> <service-b>
```

benthorner:
@thaJeztah I think the previous behaviour was consistent: `docker-compose up [services...]` would only attach to the listed services; the only special is case is with no arguments, which is interpreted as "all services". So the new behaviour is different (and therefore not the default), since in the first case it will cause non-listed service output to be shown.

> Extends #2130 to treat up <service>, which implicitly includes dependent
containers, the same way as up, which refers to all services/containers.

I notice I'm sort of contradicting myself 😆.

PR #6960, opened 2019-10-11
avoid using Stty, rely on Posix compliant os.get_terminal_size

### LINKED ISSUE #122: when service fails, abort and show stderr/stdout
If one of my services can't execute their `command:`, the only error I get is from a depending service saying it `Cannot link to a non running container`. Example:

```bash
$ fig up
Recreating syzygy_sabnzbd_1...
Recreating syzygy_sickbeard_1...
Attaching to syzygy_sabnzbd_1, syzygy_sickbeard_1
start: Cannot start container 156b95ed57fddf8c4a504d63d07b6a2e0281c09f765a066f64da7fee99731b2e: Cannot link to a non running container: /syzygy_sabnzbd_1 AS /syzygy_sickbeard_1/sabnzbd_1
```

I think it would be very helpful while still setting up a cluster if the first service whose `fig.yml`'s `command:` exits with something non-zero, would fail hard & fast, and preferably showing it's stderr.

### LINKED ISSUE #6843: Docker-compose `ps`, `top` or `images` emits the error "The system cannot find the path specified." on Windows
## Description of the issue

Usage of `docker-compose ps`, `docker-compose images` or `docker-compose top` will emit an error on Windows like:

```
PS C:\users\puppet> docker-compose ps --all
The system cannot find the path specified.
```

```
docker-compose version 1.25.0-rc2, build 661ac20e
docker-py version: 4.0.1
CPython version: 3.7.4
OpenSSL version: OpenSSL 1.1.1c  28 May 2019
```

```
Client:
 Version:           master-dockerproject-2019-08-01
 API version:       1.40
 Go version:        go1.12.7
 Git commit:        e505a7c2
 Built:             Fri Aug  2 00:02:45 2019
 Experimental:      false

Server:
 Engine:
  Version:          master-dockerproject-2019-08-01
  API version:      1.41 (minimum version 1.24)
  Go version:       go1.12.7
  Git commit:       4fb5e9e
  Built:            Fri Aug  2 00:12:20 2019
  OS/Arch:          windows/amd64
  Experimental:     true
```

N/A

1. Run `docker-compose images` on Windows

### Observed result

```
PS C:\users\puppet> docker-compose ps --all
The system cannot find the path specified.
```

### Expected result

```
PS C:\users\puppet> docker-compose ps --all
```

I tracked this using `procmon` (roughly the equivalent of `strace` on Windows). Once I had the commands being executed it was easy to backtrack into the compose source:

Screen Shot 2019-08-12 at 3 34 24 PM

`top`, `images` and `ps` all make a call like this ot format tabular output https://github.com/docker/compose/blob/7a7c9ff67a9716831c4ec2fdfa099b91a6c1858f/compose/cli/main.py#L616

The implementation of `Formatter.table()` calls https://github.com/docker/compose/blob/7ae632a9ee7530fcf81e212baa3e588f477ea862/compose/cli/formatter.py#L14

The use of `stty` is completely invalid on Windows as there is no concept of `stty`

A fun way to get the problem to go away is to create the directory `c:\dev` on Windows. Then after running `docker-compose images` the file `c:\dev\null` will be created with the contents:

```
PS C:\users\puppet> type \dev\null
'stty' is not recognized as an internal or external command,
operable program or batch file.
```

## Additional information

Windows 10 Build 1903 (19H1)

ndeloof:
Resolves #6843

image

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,4 @@
+backports.shutil_get_terminal_size==1.0.0
 backports.ssl-match-hostname==3.5.0.1; python_version < '3'
 cached-property==1.3.0
 certifi==2017.4.17
--- a/compose/cli/formatter.py
+++ b/compose/cli/formatter.py
@@ -2,20 +2,25 @@ from __future__ import absolute_import
 from __future__ import unicode_literals
 
 import logging
-import os
+import shutil
 
 import six
 import texttable
 
 from compose.cli import colors
 
+if hasattr(shutil, "get_terminal_size"):
+    from shutil import get_terminal_size
+else:
+    from backports.shutil_get_terminal_size import get_terminal_size
+
 
 def get_tty_width():
-    tty_size = os.popen('stty size 2> /dev/null', 'r').read().split()
-    if len(tty_size) != 2:
+    try:
+        width, _ = get_terminal_size()
+        return int(width)
+    except OSError:
         return 0
-    _, width = tty_size
-    return int(width)
 
 
 class Formatter:
--- a/setup.py
+++ b/setup.py
@@ -55,7 +55,8 @@ extras_require = {
     ':python_version < "3.2"': ['subprocess32 >= 3.5.4, < 4'],
     ':python_version < "3.4"': ['enum34 >= 1.0.4, < 2'],
     ':python_version < "3.5"': ['backports.ssl_match_hostname >= 3.5, < 4'],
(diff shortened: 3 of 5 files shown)


requirements.txt
```diff
@@ -1,3 +1,4 @@
+backports.shutil_get_terminal_size==1.0.0
```
chris-crone: I think you'll need to add this to `setup.py` too so that it's included when people do a `pip install`

PR #6964, opened 2019-10-14
Add working dir, config files and env file in service labels

guillaumerose:
The original intent is to be able to build an UI on top of docker-compose. These new fields allow us to discover a compose application and be able to act on it on behalf of the user (start, stop, restart).

```
$ docker inspect container1
....
"Labels": {
    "com.docker.compose.config-hash": "19601b04ec0b9c40755e416940dff4373e89880bb5b86c111d32b927b9b1a991",
    "com.docker.compose.container-number": "1",
    "com.docker.compose.oneoff": "False",
    "com.docker.compose.project": "v2-simple",
    "com.docker.compose.project.config_files": "tests/fixtures/v2-simple/docker-compose.yml,tests/fixtures/entrypoint-composefile/docker-compose.yml",
    "com.docker.compose.project.environment_file": "tests/fixtures/env/one.env",
    "com.docker.compose.project.working_dir": "/Users/guillaumerose/go/src/github.com/docker/compose/tests/fixtures/v2-simple",
    "com.docker.compose.service": "another",
    "com.docker.compose.version": "1.25.0dev"
}
```


guillaumerose:
PTAL @jdrouet

and800:
Hello! I have been debugging some issues with my Elasticsearch logs, and it seems this change is the cause of failure.

Although this patch only adds new labels, it changes the type of `com.docker.compose.project` label from `string` to `string with some nested attributes`. So there are the tools which rely on some typing for the labels (like filebeat+elasticsearch), and they may fail at this point.

Maybe it would be better to not reuse `com.docker.compose.project.` prefix? Although I'm afraid it's too late, somebody may be already relying on these new labels

ndeloof:
@and800 I hardly understand the issue you have with those new labels. As you lookup container metadata for label `com.docker.compose.project` you'll only get a string, as before. To access other labels you would need to explicitly lookup label `com.docker.compose.project.working_dir` and others. 

Sounds like you ElasticSearch query do aggregate `com.docker.compose.project*` into a complex object, while it should only look for single label `com.docker.compose.project`. Could you please tell me more about it ?

and800:
I have:
+ a single Elasticsearch cluster
+ many docker containers, which are run on many hosts
+ some of that containers are managed by docker-compose
+ Filebeat daemons, which harvest docker logs, add docker labels to that logs, and insert into Elasticsearch

With previous docker-compose version, there is `com.docker.compose.project` field in Elasticsearch index of `string` type. Recently I have added more hosts to my infrastructure, installed the most recent docker-compose on them, and connected them to the Elasticsearch cluster. With the new labels, Filebeat tries to insert a `com.docker.compose.project` field of type `Object` (with keys `value`, `config_files`, `environment_file`, `working_dir`), therefore insert operation fails (as the index already has that value of different type).

So for me it is a breaking change. We can actually blame Filebeat/Elastic for it, but I think that the label which has both value and child labels, is an example of quite controversial design.

ndeloof:
I can understand the frustration but labels are by nature key=values, not expected to be interpreted as attributes for a more complex data structure.

This is a pretty unexpected side effect of Filebeat behaviour with dot notation.
Changing labels to adopt `com.docker.compose.*` is feasible with a transition period, need to check with Desktop team about possible timeframe for this.

PR #6932, opened 2019-09-30
Large re-engineering of the build

jcsirot:
Improve the release process from Jenkins

PR #6967, opened 2019-10-16
Update engine to 19.03.3, and Alpine 3.10.2

thaJeztah:
See individual commit messages for details

Some things I noticed;

Although the Dockerfile defaults to Python 3.7.4, the build-debian stage installs `python2.7-dev` - is that correct? https://github.com/docker/compose/blob/79f29dd/Dockerfile#L43

The Dockerfile still has a FIXME, and pins to (older?) versions of `virtualenv` and `tox`. What's the easiest way to find out if those issues have been resolved? Will there be an error when updating to later versions, or does the breaking stuff happen "silently"? https://github.com/docker/compose/blob/79f29dd/Dockerfile#L51-L53

The circleci configuration uses Xcode 9.4.1, and was last updated a year ago; with Catalina being released, should we try testing against newer versions? https://github.com/docker/compose/blob/79f29dd/.circleci/config.yml#L5


thaJeztah:
@chris-crone @ndeloof @ulyssessouza ptal

ndeloof:
I can at least reply on your question regarding circleci build: @jcsirot has been working on migrating the whole build into a Jenkisfile, see https://github.com/docker/compose/pull/6932

chris-crone:
> Although the Dockerfile defaults to Python 3.7.4, the build-debian stage installs python2.7-dev - is that correct? https://github.com/docker/compose/blob/79f29dd/Dockerfile#L43

Not sure about this, @ulyssessouza can answer when he's back

> The Dockerfile still has a FIXME, and pins to (older?) versions of virtualenv and tox. What's the easiest way to find out if those issues have been resolved? Will there be an error when updating to later versions, or does the breaking stuff happen "silently"? https://github.com/docker/compose/blob/79f29dd/Dockerfile#L51-L53

The issue was that it broke the macOS build in a way that was not obvious. I haven't tried newer versions recently to see if this is fixed. `virtualenv` and `tox` are used for testing and build purposes so we're not shipping them.

chris-crone:
I believe most of this PR is replaced by: https://github.com/docker/compose/pull/7068

PR #6973, opened 2019-10-18
Set no-colors to true if CLICOLOR env variable is set to 0

### LINKED ISSUE #4116: --no-color does not disable colors on ERROR and WARNING logs
`docker-compose up -d --no-color` still displays ERROR messages with the word "ERROR" in red and the word "WARNING" in yellow.

It would also be nice if compose respected the `CLICOLOR=0` env variable as a synonym for `--no-color`.

glours:
Resolves #4116

if `--no-colors` flag is not set and `CLICOLOR` is defined with the value 0, Compose CLI should print log without any colors


compose/cli/main.py
```diff
@@ -1602,3 +1603,8 @@ def warn_for_swarm_mode(client):
             "To deploy your application across the swarm, "
             "use `docker stack deploy`.\n"
         )
+
+
+def set_no_color_if_clicolor(no_color_flag):
+    return True if no_color_flag or ('CLICOLOR' in os.environ and os.environ['CLICOLOR'] == "0") \
```
rumpl: Can we just `return no_color_flag or ('CLICOLOR' in os.environ and os.environ['CLICOLOR'] == "0")` ?

glours: Of course, done ;-)

compose/cli/main.py
```diff
@@ -1602,3 +1604,7 @@ def warn_for_swarm_mode(client):
             "To deploy your application across the swarm, "
             "use `docker stack deploy`.\n"
         )
+
+
+def set_no_color_if_clicolor(no_color_flag):
+    return no_color_flag or ('CLICOLOR' in os.environ and os.environ['CLICOLOR'] == "0")
```
rumpl: Use ‘os.environ.get()’

compose/cli/main.py
```diff
@@ -148,6 +149,7 @@ def setup_console_handler(handler, verbose, noansi=False, level=None):
     else:
         format_class = logging.Formatter
 
+    print(format_class)
```
rumpl: Oups

glours: What are you talking about? 😇

PR #6974, opened 2019-10-19
Add `device_cgroup_rules` support to 3.7 schema

gh2o:
It was last present in the 2.4 schema. I could not find why it was removed for 3.0, so I am adding it back in.

This option is needed to selectively share devices between host and container. [See here.](https://docs.docker.com/engine/reference/commandline/create/#dealing-with-dynamically-created-devices---device-cgroup-rule)

### DIFF
--- a/compose/config/config_schema_v3.7.json
+++ b/compose/config/config_schema_v3.7.json
@@ -127,6 +127,7 @@
           "registry": {"type": "string"}
         }},
         "depends_on": {"$ref": "#/definitions/list_of_strings"},
+        "device_cgroup_rules": {"$ref": "#/definitions/list_of_strings"},
         "devices": {"type": "array", "items": {"type": "string"}, "uniqueItems": true},
         "dns": {"$ref": "#/definitions/string_or_list"},
         "dns_search": {"$ref": "#/definitions/string_or_list"},


agners:
We reverted back to Docker Compose file 2.4 for this feature, it would be nice we can use latest Docker Compose format again.

ndeloof:
Need to also update [ServiceConfig](https://github.com/docker/cli/blob/master/cli/compose/types/types.go#L156) and check support by `docker stack deploy`.

Schema v3.7 is not latest, changes should take place on https://github.com/docker/cli/blob/master/cli/compose/schema/data/ which is the canonical place for compose file schema.

agners:
@ndeloof oh I see those schemas exist in two places, in the `cli` repo in `cli/compose/schema/data/` as well as the one here under `compose/config/`. It seems that the `cli` repo has schemas up to 3.9, so I guess a resync is needed first? This is quite confusing, which one is considered the master?

From what I can tell `device_cgroup_rules` is not supported in  `docker stack deploy`, so it probably would need to go to the list of `UnsupportedProperties` for now.

ndeloof:
a resync is indeed needed, see https://github.com/docker/compose/issues/7047

agners:
The property now is part of the upcomming 3.9 version of the compose-spec: https://github.com/compose-spec/compose-spec/pull/63

Issue #6890, opened 2019-09-09
Deprecate Python2


ulyssessouza:
**Is your feature request related to a problem? Please describe.**
As January 1st approaches, support for "Python 2" gets to its end. By deprecating it we can free some machine/time on CI and maybe simplify our "dependency matrix"

Note that this issue is not meant to be executed before the January 1st 2020.

**Describe the solution you'd like**
- [ ] Remove py2 from the build matrix in Jenkins
- [ ] Remove Python 2 from setup.py
- [ ] Remove `six` from requirements.txt

**Describe alternatives you've considered**
NA

**Additional context**
NA

funkyfuture:
two quick remarks:

1) dropping support for Python 2 also allows to employ language features for better readability, like dict comprehensions, set literals; new-style classes should be used and might provide leverage for further improvements.

2) why would the support for Python 2.7 be dropped when Python 3.2 and 3.4 is still supported though they [reached their eol](https://devguide.python.org/#status-of-python-branches).

ulyssessouza:
@funkyfuture I agree with the first remark, and for the second, I don't see any problem on dropping that. 
The point here is just that dropping 3.2 and 3.4 will just impact the `setup.py` and not the compose code itself, like removing all the `six` conditions inside the code that would be a fairly big PR

agronholm:
> dropping support for Python 2 also allows to employ language features for better readability, like dict comprehensions, set literals; new-style classes should be used and might provide leverage for further improvements.

I'm not arguing against dropping Python 2 support, but all of these language features are already supported on Python 2.

ndeloof:
There's probably many places a pure python 3 codebase could offer improvements, but my main motivation here is to reduce the maintenance effort.

SuperSandro2000:
and Python 2.7 is EOL and is going to be removed from the next Debian and Ubuntu releases.

DeeDeeG:
Hi,

> Remove the announcement banner in README.md

I think it could be worthwhile to leave this as a deprecation notice for some time _after_ the change, so folks with broken python support who come here can know what's up (that this is intentional, etc.)

Otherwise all sounds reasonable to me. :+1:

PR #6982, opened 2019-10-25
Fixed non-ascii error when using COMPOSE_DOCKER_CLI_BUILD=1 for Buildkit

smamessier:
When using 
```
DOCKER_BUILDKIT=1 docker-compose build
```
on compose's master, with docker version 18.09.8
I'm getting 
```
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc2 in position 24: ordinal not in range(128)
```
which can be traced back to this line: https://github.com/docker/compose/blob/master/compose/service.py#L1809

The problem seems to be that the line read form `stdin` has non-ascii characters and `.startswith` fails when comparing a raw string with the unicode prefix `magic_word`. 

The culprit input line (which is part of the output of `docker build`) is
```
#1     duration: 26.788µs
```
which contains the utf-8 µ character: `\xc2\xb5`. `startswith` errors out as the prefix `magic_word` is a unicode string.

This fixes the problem by encoding magic_word before making the comparison.

### DIFF
--- a/compose/service.py
+++ b/compose/service.py
@@ -1806,7 +1806,10 @@ class _CLIBuilder(object):
                 line = p.stdout.readline()
                 if not line:
                     break
-                if line.startswith(str(magic_word)):
+                # Fix non ascii chars on Python2. To remove when #6890 is complete.
+                if six.PY2:
+                    magic_word = str(magic_word)
+                if line.startswith(magic_word):
                     appear = True
                 yield json.dumps({"stream": line})
 
--- a/compose/service.py
+++ b/compose/service.py
@@ -1806,7 +1806,7 @@ class _CLIBuilder(object):
                 line = p.stdout.readline()
                 if not line:
                     break
-                if line.startswith(magic_word):
+                if line.startswith(str(magic_word)):
                     appear = True
                 yield json.dumps({"stream": line})
 


thaJeztah:
thanks! 

ping @chris-crone PTAL

smamessier:
This seems to only work on Python 2.7 let me fix this.

thaJeztah:
Looks to be green now 👍. I'm not a maintainer in this repository; wondering if we can / should have a test for this (not sure if that's possible, but just brainstorming 😂)

@ndeloof @chris-crone ptal

smamessier:
You're right. Ideally we should add a test that breaks without this patch. I'm not a maintainer either so that would take me quite a bit of time to get into that.

smamessier:
I realized this patch may not be necessary. I was basically trying to use compose's master using
```
python setup.py develop
```
This probably assumes `python` is Python 3? Since string types and return type of stdin.readline has changed since python 2.7, it was failing. 

When I install `compose` using `python3 setup.py develop` I don't have this problem anymore. 
So unless we're supporting python 2.7 - which I doubt - we can dismiss this patch.

smamessier:
So should we merge this?

thaJeztah:
@ulyssessouza @chris-crone ptal

ulyssessouza:
Actually just adding an `if` before this line would be better to mark it as a Python2 issue, like:
```
if six.PY2:
    line = str(line)
```
So that's even easier to take this off when we execute #6890

I can do that now, and we are good to merge

smamessier:
Thanks a lot!

ulyssessouza:
Thank you @smamessier for the catch!

Issue #5905, opened 2018-04-25
Don't use red for CLI output unless there's an error


tedivm:
## Description of the issue

I'll admit this is kind of a silly request, but I think it would be nice if docker-compose avoided using the "red" cli text during normal operations. For most tools red means something bad happened, so I (and at least [eighteen other people](https://github.com/docker/compose/issues/4269#issuecomment-310674410)) think the CLI would be much better if it didn't use red.

shin-:
Running `git diff` must be an adventure for all 19 of you. :wink: 

Feel free to make a PR for it.

tedivm:
I'll definitely make a PR, I just wanted to make sure it would be appreciated first :-)

PR #6996, opened 2019-11-03
Make container service color deterministic, remove red from chosen colors

### LINKED ISSUE #4269: Allow colors for container log messages to be specified in docker-compose.yml
When debugging a particular group of containers run with `docker-compose`, one becomes very familiar with looking for messages reported from a certain color.  However, every once in a while these colors unpredictably change.  It would be nice if the RGB code could be specified for each container in the `docker-compose.yml` file, to ensure the colors are always what I expect.

Including @nweston, whose brought this idea to my attention.

ajlai:
Resolves #4269 (at least the deterministic color assignment part of it), and #5905

### DIFF
--- a/compose/cli/colors.py
+++ b/compose/cli/colors.py
@@ -41,9 +41,9 @@ for (name, code) in get_pairs():
 
 
 def rainbow():
-    cs = ['cyan', 'yellow', 'green', 'magenta', 'red', 'blue',
+    cs = ['cyan', 'yellow', 'green', 'magenta', 'blue',
           'intense_cyan', 'intense_yellow', 'intense_green',
-          'intense_magenta', 'intense_red', 'intense_blue']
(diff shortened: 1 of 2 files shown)


compose/cli/colors.py
```diff
@@ -41,9 +41,9 @@ def make_color_fn(code):
 
 
 def rainbow():
-    cs = ['cyan', 'yellow', 'green', 'magenta', 'red', 'blue',
+    cs = ['cyan', 'yellow', 'green', 'magenta', 'blue',
           'intense_cyan', 'intense_yellow', 'intense_green',
-          'intense_magenta', 'intense_red', 'intense_blue']
+          'intense_magenta', 'intense_blue']
```
loicwolff: Would be nice to remove `intense_blue` as well since it's not easily readable with a black background. Or considering replacing it with something else.

PR #7005, opened 2019-11-06
Pass in HOME env-var in container mode.

glance-:
To get ~/-paths to work as expected in contaier mode, env-var HOME must
be the same outside the container as inside the docker-compose
container, otherwise HOME inside the container points to /root which
might not be what the user expects.

### DIFF
--- a/script/run/run.sh
+++ b/script/run/run.sh
@@ -43,7 +43,7 @@ if [ -n "$compose_dir" ]; then
     VOLUMES="$VOLUMES -v $compose_dir:$compose_dir"
 fi
 if [ -n "$HOME" ]; then
-    VOLUMES="$VOLUMES -v $HOME:$HOME -v $HOME:/root" # mount $HOME in /root to share docker.config
+    VOLUMES="$VOLUMES -v $HOME:$HOME -e HOME" # Pass in HOME to share docker.config and allow ~/-relative paths to work.
 fi
 
 # Only allocate tty if we detect one


glance-:
Example:
```version: '3.7'
services:
  test:
    image: "ubuntu"
    command: >
      cat /run/secrets/BASH_RC
    secrets:
      - BASH_RC

secrets:
  BASH_RC:
    file: ~/.bashrc
```
Before this patch, this would print roots .bashrc when running in container-mode, and after this patch, this prints the .bashrc of the user running the container-mode script.

ndeloof:
Just to better understand your needs: can't you just use binary package installation so you don't get into this env variable mistmatch between container and host?

ndeloof:
> For a number of complicated reasons I can't.

That doesn't help me much :P
I can see this fixes the bug you describe, just trying to establish why this packaging is required / how much popular it is, so we can get a good governance on docker-compose project and focus on the right things. Just telling me "_my problem is too complex to explain, this is just a fix that works for me_" doesn't help :D

LGMT from a code point of view, just surprised we support such a packaging which comes with obvious environment variable limitations (not just `HOME`)

glance-:
One reason we use it is to abstract away the architecture specific binary need. We just have the compose wrapper along with the rest of the compose yaml files and thats all thats needed to start the service on a host.

I think the container wrapper is a valuable distribution method, and as long as it exists it should be kept as compatible with the "regular" distribution binaries as reasonably possible.

I don't do alot of compose magic, but this was the only thing I've found yet where they two don't behave the same way.

ndeloof:
"kept as compatible" is the main issue here : this packaging comes with some major constraint, docker-compose being used by most people with environment variable interpolation, that this packaging just block (due to inside/outside environment segregation by the nature of docker containers).

I'm fine with the specific change you suggest (already approved your pull-request) but I still consider this packaging format as just broken, and suggest it get deprecated.

glance-:
In which way is this packaging format broken? The only issue I've found was this, and it was a easy fix.

ndeloof:
compose container can't access you host environment variables to run interpolation. Might be considered a feature, not a bug :P
But at least this is pretty confusing for poeple getting a distinct behaviour depending on the way they installed docker-compose.

glance-:
It's not that hard to fix, if it's considered a bug, and I don't think it's enough reason to scrap this distribution mechanism.

ndeloof:
> It's not that hard to fix

Which env var should be propagated to compose container ? Would at least require to blacklist the most obvious `PATH`, maybe a few others ?

My point is that this packaging is not replicating a correct docker-compose user experience, and it comes with already two other supported way to install, so more maintenance and risk for specific bugs. I'd prefer we limit to just a reasonable set of supported packaging. Doesn't mean you can't just use official docker-compose binary in a container on your own if you like this approach.

script/run/run.sh
```diff
@@ -43,7 +43,7 @@ if [ -n "$compose_dir" ]; then
     VOLUMES="$VOLUMES -v $compose_dir:$compose_dir"
 fi
 if [ -n "$HOME" ]; then
-    VOLUMES="$VOLUMES -v $HOME:$HOME -v $HOME:/root" # mount $HOME in /root to share docker.config
+    VOLUMES="$VOLUMES -v $HOME:$HOME -e HOME" # Pass in HOME to share docker.config and allow ~/-relative paths to work.
```
ulyssessouza: You should be able to pass the `-e HOME` through the variable `COMPOSE_OPTIONS` (as described in line 11). If we remove this volume mount here, that can break peoples code that expects this mount.

glance-: Yes, it can break code for anyone who reads files from /root when running docker-compose wrapper script as a non root user. That sounds like "working by accident", and works differently than the stand-alone binary.

I'd say this makes the whole wrapper script behave more like the stand-alone binary.

ulyssessouza: Ok fair enough!

ulyssessouza:
Thank you @glance- !

PR #7009, opened 2019-11-07
Resolve shellcheck warnings in container-script

glance-:
This fixes a couple of small potential issues pointed out by shellcheck.

### DIFF
--- a/script/run/run.sh
+++ b/script/run/run.sh
@@ -56,7 +56,7 @@ DOCKER_RUN_OPTIONS="$DOCKER_RUN_OPTIONS -i"
 
 
 # Handle userns security
-if docker info 2>/dev/null | grep -q userns; then
+if docker info --format '{{json .SecurityOptions}}' 2>/dev/null | grep -q 'name=userns'; then
     DOCKER_RUN_OPTIONS="$DOCKER_RUN_OPTIONS --userns=host"
 fi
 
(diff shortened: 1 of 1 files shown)


script/run/run.sh
```diff
@@ -56,8 +56,9 @@ DOCKER_RUN_OPTIONS="$DOCKER_RUN_OPTIONS -i"
 
 
 # Handle userns security
-if [ ! -z "$(docker info 2>/dev/null | grep userns)" ]; then
+if docker info 2>/dev/null | grep -q userns; then
```
ulyssessouza: The absence of `[ ]` breaks the formatting compatibility with the others

glance-: Sorry, but I don't understand your statement here.

glance-: [ is not syntax, nor formating, its a semarate tool, also known as test.

if looks at the exit code from [ in the previous code, and in the change I've written it looks at the exit code from grep instead.

glance-: More about this can be read at https://github.com/koalaman/shellcheck/wiki/SC2143

glance-: https://github.com/koalaman/shellcheck/wiki/SC2236 is also related to previous code.

thaJeztah: A minor optimisation could be to use `--format` (so that only the Security options are output), and/or to match `name=userns` instead of just `userns` (as that's what the daemon will return; https://github.com/moby/moby/blob/b6684a403c99aaf6be5b8ce0bef3c6650fcdcd12/daemon/info.go#L180-L182

```suggestion
if docker info --format '{{json .SecurityOptions}}' 2>/dev/null | grep -q 'name=userns'; then
```

glance-:
@ulyssessouza Can we get this resolved?

PR #7010, opened 2019-11-08
Build OSX binary as a directory

### LINKED ISSUE #6956: docker-compose super slow on macOS Catalina
when i do docker-compose up -d, it takes a while to start the process, same when i do docker-compose restart. 

in other words, there's a delay of few seconds before it begins to restart containers or stopping containers.

ndeloof:
OSX Catalina do scan executables, and the singlefile distribution has side-effect that the bundled Pythin runtime, unpacked into a temporary folder, is scanned on each and every run.

With directory packaging, `docker-compose` executable will refer to side files in the distribution folder, only scanned once first time it is ran.

attempt to fix #6956

### DIFF
--- a/docker-compose_darwin.spec
+++ b/docker-compose_darwin.spec
@@ -1,0 +1,108 @@
+# -*- mode: python ; coding: utf-8 -*-
+
+block_cipher = None
+
+a = Analysis(['bin/docker-compose'],
+             pathex=['.'],
+             hiddenimports=[],
+             hookspath=[],
+             runtime_hooks=[],
+             cipher=block_cipher)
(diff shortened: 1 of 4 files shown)


jcsirot:
We need to backport this one to #6932

docker-compose_darwin.spec
```diff
@@ -0,0 +1,108 @@
+# -*- mode: python ; coding: utf-8 -*-
```
chris-crone: I'm guessing there isn't a way we can avoid duplicating the information here vs in `docker-compose.spec`?

PR #7020, opened 2019-11-14
Fix ``run`` action to respect COMPOSE_IGNORE_ORPHANS environment variable

vaab:
It seems that ``COMPOSE_IGNORE_ORPHANS`` is not checked when using ``compose run`` which is quite cumbersome : people are more likely to use ``compose run`` for run-once CLI scripts and would then care about the outputs (stdout and stderr), they might not want to see warning about orphans poping out.

As I was not sure this was something you might want, tell me if I need to write tests or documentation.

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -1343,7 +1343,8 @@ def run_one_off_container(container_options, project, service, options, toplevel
                 service_names=deps,
                 start_deps=True,
                 strategy=ConvergenceStrategy.never,
-                rescale=False
+                rescale=False,
+                ignore_orphans=toplevel_environment.get_boolean('COMPOSE_IGNORE_ORPHANS'),
             )
 
     project.initialize()


ndeloof:
Seems legitimate to me and not a risky change.

redcinelli:
> Hello @vaab ! Thank you for the contribution!
> Could you please add tests to your PR?

I am keen on taking on the testing, but @ulyssessouza could you please help me out.

What should we be testing ? Some test on the `project.up()` already exist with the `ignore_orphans` set to either values.
And I don't think you want us to test `toplevel_environment.get_boolean('COMPOSE_IGNORE_ORPHANS')`

So I feel this behavior is already covered.
Which makes me a bit confused. :sweat_smile:

Wirone:
Can't believe it's still open after 2 years 🤯

PR #7027, opened 2019-11-18
Allow compatibility option with `COMPOSE_COMPATIBILITY` environment variable

### LINKED ISSUE #7026: Allow compatibility mode to be setup via environment variables or .env
The `--compatibility` option is only available as an option and I find myself obliged to add it everywhere on command line or in shell scripts when I do `up` or `run` commands.

I would like to setup compatibility mode from my `.env` file.

The only alternative I've come up with is to use some script wrapper to docker-compose that prepend this option.

kiniou:
Resolves #7026

### DIFF
--- a/compose/cli/command.py
+++ b/compose/cli/command.py
@@ -59,7 +59,7 @@ def project_from_options(project_dir, options, additional_options={}):
         tls_config=tls_config_from_options(options, environment),
         environment=environment,
         override_dir=override_dir,
-        compatibility=options.get('--compatibility'),
+        compatibility=compatibility_from_options(project_dir, options, environment),
         interpolate=(not additional_options.get('--no-interpolate')),
         environment_file=environment_file
     )
@@ -90,7 +90,7 @@ def get_config_from_options(base_dir, options, additional_options={}):
     )
     return config.load(
         config.find(base_dir, config_path, environment, override_dir),
-        options.get('--compatibility'),
+        compatibility_from_options(config_path, options, environment),
         not additional_options.get('--no-interpolate')
     )
 
@@ -198,3 +198,13 @@ def get_project_name(working_dir, project_name=None, environment=None):
         return normalize_name(project)
 
     return 'default'
+
+
+def compatibility_from_options(working_dir, options=None, environment=None):
+    """Get compose v3 compatibility from --compatibility option
+       or from COMPOSE_COMPATIBILITY environment variable."""
+
+    compatibility_option = options.get('--compatibility')
+    compatibility_environment = environment.get_boolean('COMPOSE_COMPATIBILITY')
+
+    return compatibility_option or compatibility_environment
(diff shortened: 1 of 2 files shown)


tests/acceptance/cli_test.py
```diff
@@ -552,24 +570,15 @@ def test_config_compatibility_mode(self):
...
-                    'networks': {'bar': None},
-                }
-            },
-        }
+        assert yaml.load(result.stdout) == COMPOSE_COMPATIBILITY_DICT
+
+    @mock.patch.dict(os.environ)
+    def test_config_compatibility_mode_from_env(self):
+        self.base_dir = 'tests/fixtures/compatibility-mode'
+        result = self.dispatch(['--compatibility', 'config'])
```
ndeloof: I might misunderstand the testcase, but shouldn't you set env variable then run `config` command _without_ `--compatibility` ?

kiniou: Whoops and yes it's what I intended to do but I forgot to do it fully 🤦‍♂️
I'll fix the test as soon as i get a real computer in front of me.

kiniou: :heavy_check_mark: This should be fixed with commit 85730b3e9e4595488caf4c878f66a5dd80404af9

compose/cli/command.py
```diff
@@ -188,3 +188,12 @@ def normalize_name(name):
         return normalize_name(project)
 
     return 'default'
+
+
+def compatibility_from_options(working_dir, options=None, environment=None):
+    """Get compose v3 compatibility options from environment then from options."""
+
+    if 'COMPOSE_COMPATIBILITY' in environment:
+        return environment.get_boolean('COMPOSE_COMPATIBILITY')
```
ulyssessouza: In this case the environment takes the precedence that should be the opposite.
What if environment variable is set to `COMPOSE_COMPATIBILITY=no` and command line uses `--compatibility`?

Could you please, add a test using both?

kiniou: I think i have overlooked somehow the way the other functions handle precedence for options and environment variables. I will fix the precedence order :ok_hand:.

Now, if i have understood the precedence (:point_up:) correctly and we have both `COMPOSE_COMPATIBILITY=no` and `--compatibility`, `docker-compose` should results in a compatibility mode execution.
If that's the case, I can indeed add a test that cover this case.

ulyssessouza: Yep! The command line should have precedence over the environment variable

kiniou: :ok_hand: My last push (with squashing to make commit tidier) should do the job.

To be noted: `COMPOSE_COMPATIBILITY=no` would result in a True value since `Environment.get_boolean(key)` test negative value only and handle only `false` or `0`.

ulyssessouza:
Just one last little thing. 
Could you please mention the variable in the help so people can now the env var exists?

kiniou:
> Just one last little thing.
> Could you please mention the variable in the help so people can now the env var exists?

@ulyssessouza : 🤔 Do you mean the docstring or the docker help documentation ?
For the latter, I already did the changes in https://github.com/docker/docker.github.io/pull/9900 .

ulyssessouza:
The docstring

PR #7031, opened 2019-11-19
Removed Python2 support

### LINKED ISSUE #6890: Deprecate Python2

venthur:
Closes: #6890

Signed-off-by: Bastian Venthur <bastian.venthur@flixbus.com>

Resolves #6890


ndeloof:
build failure is expected as the Jenkinsfile being built is the one from target branch, not the one modified by PR.
@ulyssessouza are you ok I remove `py27` from Jenkinsfile (which we plan to replace by https://github.com/docker/compose/pull/6932 anyway) so this PR can be tested by CI?

venthur:
Yes please :)
I was desperately searching for any py27 occurrences in the code.

ndeloof:
I was expecting so, so my comment :) I had such issue myself in the past and started getting crazy.

ulyssessouza:
Hello @venthur ! Thank you for the PR!

@ndeloof Is right about the Jenkinsfile. 

My problem with this PR landing now is that Python2 is not in the EOL. Sorry, but it's too early to merge it. That would break people's code without having the Python2.7 EOL to back our move.

ndeloof:
Python 2 will be EOL in a few weeks, and this would only impact people who consider docker-compose as a library to build third party products, which we _know_ exists but never aknowledge we want to support. We should be free to refactor our codebase and drop python 2 support as long as the CLI tool is available to our end-user base.

ndeloof:
see https://github.com/docker/compose/issues/6890

ndeloof:
@venthur could you please reconfigure your PR to target `py2` branch ? 
I've disabled 2.7 as a target on this one, so we will get a real CI to prepare this move.

venthur:
@ndeloof Did I do something wrong?

ndeloof:
nope, I just missed the reference to python 2.7 by `.tox`.

ndeloof:
`No module named 'requests'` seems the build system goes way beyond my understanding...

venthur:
I don't know what's going on either, I double-checked the diff, `requests` does not even show up in them.

I did however remove `python2(-dev)` from the Dockerfile w/o replacing it with python3, I can try again with adding python3...

ndeloof:
please give me some time to get py2 branch fixed ;)

ndeloof:
I was able to fix the py2 branch (https://ci.docker.com/public/blue/organizations/jenkins/compose/detail/py2/3/pipeline/) and rebased your PR on top (I had to force push, so if you need to make some changes, be carreful to re-create your local branch from https://github.com/venthur/compose/tree/remove_python2)

venthur:
Nice!

With respect to the merge, I also tend to agree that it should not matter if we merge it now, as long as the CLI has a major version bump. If people consume directly from `master` and still depend on Python2, well that's dangerous anyways these days.

On the other hand, if the patch itself looks good to you and is likely to be merged at some point, I don't have a problem maintaining it for a while.

venthur:
Python2 reached its EOL, can we merge now?

compose/utils.py
```diff
@@ -22,9 +17,7 @@
 
 
 def get_output_stream(stream):
-    if six.PY3:
-        return stream
-    return codecs.getwriter('utf-8')(stream)
+    return stream
```
ulyssessouza: IMO, this function can be removed

tests/__init__.py
```diff
@@ -1,6 +1,3 @@
-from __future__ import absolute_import
-from __future__ import unicode_literals
-
 import sys
 
 if sys.version_info >= (2, 7):
```
ulyssessouza: This if will become always true

venthur: Fixed.

compose/utils.py
venthur: You're right, fixed.

tests/__init__.py
```diff
@@ -1,12 +1,4 @@
...
-import sys
-
-if sys.version_info >= (2, 7):
-    import unittest  # NOQA
-else:
-    import unittest2 as unittest  # NOQA
+import unittest  # NOQA
 
 try:
     from unittest import mock
```
ulyssessouza: Now that import doesn't need to be guarded by a try-catch anymore

venthur:
So, it's mid-Februrary. Any chance we can drop Py2 support by merging? :)

ulyssessouza:
@venthur Sorry for the late reply, but we got to the conclusion (based on pypi downloads data) that it would be more prudent to sync that with the next Ubuntu release. That new release will eliminate Python2.

alexrecuenco:
I was working on a branch before I noticed you had already done most of the same things I did @venthur . Is it ok if I add to it? In my branch I also changed dictionary comprehensions and other details that are cleaner in python3. (Like format strings)

graingert:
You might want to run and add pyupgrade as a pre-commit hook

alexrecuenco:
> You might want to run and add pyupgrade as a pre-commit hook
> […](#)

I noticed we are on this branch far behind master, can we rebase?

ulyssessouza:
@venthur Can you rebase this PR pls? We are finally ready to merge

venthur:
Done, re-based onto current master.

remram44:
This removed `release.py` script completely? Was that intended?

PR #7034, opened 2019-11-19
Prepare drop of python 2.x support

ndeloof:
Preparation step for #6890
As CI on PRs will run master's Jenkinsfile, so if we remove python 2.7 support, they will just fail.


ulyssessouza:
Dropping in favor of a branch in the py2 removal

Issue #6464, opened 2019-01-15
`docker-compose pull` fails if a service includes both image and build


justinmchase:
## Description of the issue
I have a `docker-compose.yml` file where I have multiple services built from the same file but which have different commands. I don't want to build the same image multiple times every time I do a `docker-compose build` so I am setting both an `image` and a `build` on the service, [per the documentation](https://docs.docker.com/compose/compose-file/#image):

> If the image does not exist, Compose attempts to pull it, unless you have also specified build, in which case it builds it using the specified options and tags it with the specified tag.

Related to:
- #2092
- #3853

**Example docker-compose.yml**
```yml
version: "3"
services:
  build:
    image: example
    build: .
    command: npx tsc
  test
    image: example
    command: npx mocha
```

**Example Dockerfile**
```docker
FROM node:8
WORKDIR /app
```

`build` works fine...
```sh
$ docker-compose build
test uses an image, skipping
Building build
Step 1/2 : FROM node:8
 ---> 62ba2b48bf28
Step 2/2 : WORKDIR /app
 ---> Using cache
 ---> c05a29354b96
Successfully built c05a29354b96
Successfully tagged example:latest
```

`up` works fine...
```sh
$ docker-compose up
Creating network "test_default" with the default driver
Creating test_test_1  ... done
Creating test_build_1 ... done
Attaching to test_test_1, test_build_1
test_1   | from test
build_1  | from build
test_test_1 exited with code 0
test_build_1 exited with code 0
```

`pull` fails, it erroneously tries to pull example from dockerhub
```sh
$ docker-compose pull
Pulling build ... error
Pulling test  ... error

ERROR: for test  pull access denied for example, repository does not exist or may require 'docker login'

ERROR: for build  pull access denied for example, repository does not exist or may require 'docker login'
ERROR: pull access denied for example, repository does not exist or may require 'docker login'
```

## Context information (for bug reports)

**Output of `docker-compose version`**
```
docker-compose version 1.23.2, build 1110ad01
docker-py version: 3.6.0
CPython version: 3.6.6
OpenSSL version: OpenSSL 1.1.0h  27 Mar 2018
```

**Output of `docker version`**
```
Client: Docker Engine - Community
 Version:           18.09.0
 API version:       1.39
 Go version:        go1.10.4
 Git commit:        4d60db4
 Built:             Wed Nov  7 00:47:43 2018
 OS/Arch:           darwin/amd64
 Experimental:      false

Server: Docker Engine - Community
 Engine:
[...]

Issue #3853, opened 2016-08-18
docker-compose pulls image if 'build' and 'image' are provided


thewilli:
I'm using docker-compose v1.8.0, build 94f7017.

Consider the following compose file where I am defining a service with an image name which exists on Docker Hub as well:

``` yml
version: "2"
services:
    myredis:
        build: .
        image: redis:2.8
```

The [compose file reference](https://docs.docker.com/compose/compose-file/) states that

> If you specify image as well as build, then Compose names the built image with the webapp and optional tag specified in image

for the example

``` yml
build: ./dir
image: webapp:tag
```

so this seems to be a valid usecase.

When I run `docker-compose pull` (because e.g. I have defined other services that I want to update), _docker-compose_ lets Docker pull the redis image 2.8 from Docker Hub, which really seems to be a bug to me. I would have expected that - because _build_ is specified - docker compose would not pull anything for this service.

PR #7039, opened 2019-11-21
Revert "only pull images that can't build"

### LINKED ISSUE #6934: Images are no longer loaded if there is a build entry in the file.
## Description of the issue

If I have build and image entries in the same docker-compose config file, then I cannot load my images and get no feedback on why this is happening. I did not understand the problem until I loaded the master and started the debugger. Perhaps it would be better if the program indicated that the images are not loaded. In addition, it also prints the reason for this.

I build my images on a CI server using a `docker-compose.yml` file and load these images into the registry. I use the same file to load the images from the registry on my server.

So we are talking about the following command sequence:

**CI-Server**
Step 1: `docker-compose build`
Step 2: `docker-compose push`

**Production Server**
Step 3: `docker-compose pull`
Step 4: `docker-compose up -d`

Because of commit [c6dd7da](https://github.com/docker/compose/pull/6494/commits/c6dd7da15eb3d85a1f7634e8ded9fe42c9035669), a docker-compose pull ignores any service that contains build instructions in the config.

If this behaviour is included in the release 1.25.0, then I (and anyone else who uses the same file for building and pulling) will have to use two separate config files in the future. 

E.g. `docker-compose.yml` + `docker-compose.build.yml` 

With this bug issue I would like to point out that the fix of issue #6464 breaks with the original behavior. I'm in favor of either reverting the commit or adding a flag that also loads image even though it contains build instructions in the config file.

```
docker-compose version 1.25.0dev, build bc57a1bd
docker-py version: 4.0.1
CPython version: 3.7.4
OpenSSL version: OpenSSL 1.1.1c  28 May 2019
```

```
 Version:           19.03.2
 API version:       1.40
 Go version:        go1.12.8
 Git commit:        6a30dfc
 Built:             Thu Aug 29 05:26:49 2019
 Experimental:      false

 Engine:
  Version:          19.03.2
  Go version:       go1.12.8
  Git commit:       6a30dfc
  Built:            Thu Aug 29 05:32:21 2019
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          v1.2.6
  GitCommit:        894b81a4b802e4eb2a91d1ce216b8817763c29fb
 runc:
  Version:          1.0.0-rc8
  GitCommit:        425e105d5a03fabd737a126ad93d62a9eeede87f
 docker-init:
  Version:          0.18.0
  GitCommit:        fec3683
```

```
services:
  web:
    build:
      context: /foo
    image: luuukas/nginx
    ports:
    - published: 80
      target: 80
version: '3.7'

```

```
1. docker-compose pull
// Nothing's happening
2. docker-compose up -d
// Wants to build the image
```
### Observed result
```
docker-compose pull                                
// Nothing's happening
```
### Expected result
```
docker-compose pull 
Pulling web ... done
```

## Additional information

```
Darwin macbook-pro-092.local 18.7.0 Darwin Kernel Version 18.7.0: Tue Aug 20 16:57:14 PDT 2019; root:xnu-4903.271.2~2/RELEASE_X86_64 x86_64
```

### LINKED ISSUE #7038: docker-compose pull is broken from the 1.25.0 release
## Description of the issue

in 1.24.x, we can call ```docker-compose -f xxx pull``` to download all images. After updating to 1.25.0, this function is not working.

```
docker-compose version 1.25.0, build 0a186604
docker-py version: 4.1.0
CPython version: 3.7.4
```

```
 Version:           19.03.5
 API version:       1.40
 Go version:        go1.12.12
 Git commit:        633a0ea
 Built:             Wed Nov 13 07:25:41 2019
 Experimental:      false

 Engine:
  Version:          19.03.5
  Go version:       go1.12.12
  Git commit:       633a0ea
  Built:            Wed Nov 13 07:24:18 2019
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.2.10
  GitCommit:        b34a5c8af56e510852c35414db4c1f4fa6172339
 runc:
  Version:          1.0.0-rc8+dev
  GitCommit:        3e425f80a8c931f88e6d94a8c831b9d5aa481657
 docker-init:
  Version:          0.18.0
  GitCommit:        fec3683
```

```
version: '3'

services:
  configurator:
    image: oharastream/configurator:${TAG:-0.9.0-SNAPSHOT}
    build:
      context: .
      dockerfile: configurator.dockerfile
      args:
        BRANCH: master
        REPO: https://github.com/oharastream/ohara.git

```

```docker-compose -f build.yml pull```

### Observed result
nothing

### Expected result
start to pull all images defined in build.yml

N/A

## Additional information
- OS version: Centos 7.7
- install method: download binary from github release page

ndeloof:
#6494 fix is incorect 

a service can define *both* `image` and `build`, and documentation gives some indications on how those are handled:

> If you specify image as well as build, then Compose names the built image with the webapp and optional tag specified in image

> If the image does not exist, Compose attempts to pull it, unless you have also specified build, in which case it builds it using the specified options and tags it with the specified tag.

unfortunately those indication are not command-specific and the `pull` command documentation doesn't cover this specific case.

Some use `image` to tag built services and later push images, which will be pulled by others. Some use `image` only as a "build cache" and then don't expect compose to pull those from registry. AFAICT this is the case for #6464 - can you please confirm @justinmchase ?

I also can see lack of clarification on the expectation when both `build` and `image` are set brings many ambiguities, see #3853

I propose we first revert change, so 1.25 will behave like 1.24 did, then work on better documentation on `docker-compose pull` and impacts of `build` and `image` combination for a service.

About #6464, `--ignore-pull-failures` should be enough to offer a reasonable workaround.

This PR reverts commit c6dd7da15eb3d85a1f7634e8ded9fe42c9035669.

Resolves #7038
Resolves #6934

### DIFF
--- a/compose/project.py
+++ b/compose/project.py
@@ -619,9 +619,6 @@ class Project(object):
     def pull(self, service_names=None, ignore_pull_failures=False, parallel_pull=False, silent=False,
              include_deps=False):
         services = self.get_services(service_names, include_deps)
-        images_to_build = {service.image_name for service in services if service.can_be_built()}
-        services_to_pull = [service for service in services if service.image_name not in images_to_build]
-
         msg = not silent and 'Pulling' or None
 
         if parallel_pull:
@@ -647,7 +644,7 @@ class Project(object):
                     )
 
             _, errors = parallel.parallel_execute(
-                services_to_pull,
+                services,
                 pull_service,
                 operator.attrgetter('name'),
                 msg,
@@ -660,7 +657,7 @@ class Project(object):
                 raise ProjectError(combined_errors)
 
         else:
-            for service in services_to_pull:
+            for service in services:
                 service.pull(ignore_pull_failures, silent=silent)
 
     def push(self, service_names=None, ignore_push_failures=False):
--- a/tests/acceptance/cli_test.py
+++ b/tests/acceptance/cli_test.py
@@ -661,13 +661,6 @@ class CLITestCase(DockerClientTestCase):
                 'image library/nonexisting-image:latest not found' in result.stderr or
                 'pull access denied for nonexisting-image' in result.stderr)
 
-    def test_pull_with_build(self):
-        result = self.dispatch(['-f', 'pull-with-build.yml', 'pull'])
-
-        assert 'Pulling simple' not in result.stderr
-        assert 'Pulling from_simple' not in result.stderr
-        assert 'Pulling another ...' in result.stderr
-
     def test_pull_with_quiet(self):
         assert self.dispatch(['pull', '--quiet']).stderr == ''
         assert self.dispatch(['pull', '--quiet']).stdout == ''


justinmchase:
> Some use image only as a "build cache" and then don't expect compose to pull those from registry.

Correct, in my case I'm using `image` and `build` such that the same image is shared between services but is built from local and never pulled from remote.

justinmchase:
So I don't understand this reversion, it looks like the bug was fixed and now you've just broken it again?

You said:

> unfortunately those indication are not command-specific and the pull command documentation doesn't cover this specific case.

But this change that was just reverted did cover that case, no? For those two bugs listed here it appears that they are "by design" and were previously explained by the documentation. They should then simply call "docker-compose build" and their problems are solved.

I guess I would not have recommended reversion as the fix here, perhaps I could see adding an automatic build step for images that aren't already built but I think this reversion was probably a mistake.

justinmchase:
I think I understand... the folks with the issue are using the same file to pull from prod for local work instead of building but they _can_ build it locally if they wanted to by explicitly using the build command.

One question about that, if you `pull` and then `build` will the build be skipped because you have the image already?

But if that's the case I would propose this, which I think would still work for these users but would also work for my use case:

1. If you run `docker-compose pull` and it fails to find the image on remote...
   a. If the service with a missing image also has a build, or any other service referencing that image has a build...
      i. A non-error message is printed: "example... not found, skipping local image"
   b. Else it has no local build
      i. An error is produced: "example... error"

I noticed one of the users had an image such as `image: oharastream/configurator:${TAG:-0.9.0-SNAPSHOT}` which is interesting because they may be working on a version which hasn't yet been deployed to production in which case they would (after this reversion) get the error I reported. However with the recommended change here they would get an "example... not found, skipping local image" non-error instead.

For bonus points a `-b|--build-images` (or whatever) flag could be introduced in this case to, instead of skipping, build local image which is missing from remote.

ndeloof:
Some people indeed use image as a "build cache" so that `docker-compose build` is efficient even with suboptimal Dockerfiles, as long as you use the same image tag - so the use of `image: foo/bar:${GIT_COMMIT}` and such things.

Others do use image as tag for the image being built and pushed to a registry, assuming the same compose file can later be used to pull back on production servers, fully ignoring the build section.

Once we get 1.25.1 released I'll discuss with maintainers so we create better documentation on this topic and/or introduce explicit flags so one get full control on whats happening.

justinmchase:
Sounds good thanks.

ndeloof:
See https://github.com/docker/compose/pull/7052 as my proposed "better" (?) fix for this issue

teohhanhui:
> Some use image to tag built services and later push images, which will be pulled by others.

Yes, this is very important for our use case. We build images in CI so that a new user can pull them (to provide the best OOTB experience), while the user should also be able to build the images locally.

Thanks for reverting.

PR #7046, opened 2019-11-22
Support buildkit build time secrets

antiBaconMachine:
Hi everyone. This is one possible solution to build time secret mounting. I've used the build section in the compose yaml, I think there's a good argument to be made for supporting `--build-secret` on the CLI as well / instead.

I'm new to this project and I don't typically write python so all feedback welcome, what do people think of using the build section like this?

```
    build:
      context: .
      secrets:
        - "id=secret,src=${FOO_SECRET:-secret_1}"
```

The use case I originally wanted this for is mounting .npmrc so I imagine I would check in something like `id=npmrc,src=${FOO_SECRET:-~/.npmrc}` and then anyone with their .npmrc in the "normal" place would never have to think about this again. That is is the motivation for doing it this way first before the CLI which puts more burden on user.

If this PR isn't to your liking that is fine but you should still consider the first commit which fixes an existing test which does not behave as I think it was intended.

Resolves ##6358


ndeloof:
Sounds a reasonable approach.
I'd prefer we abstract away from buildkit CLI syntax, and define build secrets as an id=>value map:
```
secrets:
        secret_id: "${FOO_SECRET:-secret_1}"
```
which possibly could be extended for later additional parameters buildKit could introduce:
```
secrets:
        secret_id: 
			src: "${FOO_SECRET:-secret_1}"
			x-to-be-debined: value
```

As a side-note, when you consider working on such an enhancement, please first report an issue so this can be discussed with maintainers, and you don't waste time working hard on code that might not be later approved.

note: should also be implemented in https://github.com/docker/buildx for consistency

Teyras:
Can this also be used to the pass SSH environment (e.g., `docker build --ssh default`) to the builder?

ndeloof:
@Teyras would need to look into details, but assuming BuildKit did define distinct build args for this purpose, we probably would better introduce `ssh` build option for this specific use-case. See https://github.com/docker/compose/issues/7025

antiBaconMachine:
Thanks @ndeloof for the feedback. I've had a go at an abstract buildkit syntax let me know what you think.

It looks a bit odd as it's a list of strings or dicts. I didn't want to force people to use the full `src` form if all they need is `id`. ideally I'd want something like 

Thanks for the hint about opening an issue. In this case I didn't really set out to write a PR it just kinda went that way once I started reading.

Can you please clarify your comment on buildx, this feature is already implemented there, I'm not sure what action you are recommending?

antiBaconMachine:
@Teyras No this is specifically to get parity with buildkit `--secret` flag but if the maintainers accept this solution or one like it, no reason the same approach couldn't be used for ssh as well. I've tried to keep the buildkit string builder generic to support such an eventuality.

ndeloof:
General approach looks good to me an could be used to add support for buildkit's `--ssh` option.
On schema update, please note the leading schema is hosted on [docker/cli](https://github.com/docker/cli/tree/master/cli/compose/schema/data) (see https://github.com/docker/compose/issues/7047). Should also avoid something generic as `dict` and define explicit `build_secret` type with single property `src` for the current buildkit implementation, which could evolve in future with more options. Also, for schema consistency, please mimic existing syntax in use, as services:volumes: https://github.com/docker/cli/blob/master/cli/compose/schema/data/config_schema_v3.9.json#L287 

I'll discuss with maintainers on way we'd like to proceed with such schema updates.

antiBaconMachine:
Thanks @ndeloof, have patched the last commit to be in line with volumes, thanks for pointing that out, I was looking for a good example to crib.

Shame about the schema replication issue, that does complicate things somewhat, but I'm sure there's a reasonable solution.

maxwellsmart84:
Do you guys know if there is a known workaround for doing ssh forwarding with docker-compose till then?  I have tried many suggested methods but can't seem to get around password protected RSA keys.

NinoFloris:
Would this also include a way to pass a secret as a CLI argument?

CpuID:
Can't wait to see this PR get over the line ;)

bilby91:
This change would allow using `mount=type=ssh` ? Looking forward on this one!

dsvictor94:
What is the behavior when building without `DOCKER_BUILDKIT=1` and `COMPOSE_DOCKER_CLI_BUILD=1` present?

psigen:
It appears that it should just ignore the secrets if `DOCKER_BUILDKIT` is not set:
https://github.com/docker/compose/pull/7046/files#diff-db70ffd20201b81c947f72fea35551d2R1812

tests/integration/service_test.py
```diff
@@ -973,12 +974,49 @@ def test_build_cli(self):
         self.addCleanup(shutil.rmtree, base_dir)
 
         with open(os.path.join(base_dir, 'Dockerfile'), 'w') as f:
-            f.write("FROM busybox\n")
+            # f.write("FROM busybox\n")
```
psigen: Remove commented code?
```suggestion
```

tests/integration/service_test.py
```diff
@@ -973,12 +974,49 @@ def test_build_cli(self):
...
+        assert self.client.inspect_image('composetest_web')
+
+    def test_build_cli_with_secrets(self):
+        base_dir = tempfile.mkdtemp()
+        self.addCleanup(shutil.rmtree, base_dir)
+        secret_path_1 = os.path.join(base_dir, 'secret_1')
+        secret_path_2 = os.path.join(base_dir, 'secret_2')
+
+        with open(os.path.join(base_dir, 'Dockerfile'), 'w') as f:
+            # f.write("FROM busybox\n")
```
psigen: Remove commented code?
```suggestion
```

compose/config/config_schema_v3.7.json
```diff
@@ -88,7 +88,8 @@
                 "cache_from": {"$ref": "#/definitions/list_of_strings"},
                 "network": {"type": "string"},
                 "target": {"type": "string"},
-                "shm_size": {"type": ["integer", "string"]}
+                "shm_size": {"type": ["integer", "string"]},
+                "secrets": {"$ref": "#/definitions/build_secrets"}
```
thaJeztah, line 92: Note that we can't update the 3.7 schema, as it's already been released, so to add this property to the schema, it probably has to be added to the upcoming 3.9 schema in https://github.com/docker/cli/blob/master/cli/compose/schema/data/config_schema_v3.9.json first

CpuID:
@antiBaconMachine @ndeloof next steps here...? :)

gnagel:
I'm very excited to be able to use `ssh` secrets in the docker compose builds.

greg-caffeine:
Just curious if anyone thinks using the existing secrets syntax (v3.1+) would be cleaner:

```
services:
  my-application:
    build:
      secrets:
        - my-secret

secrets:
  my-secret:
    file: /path/to/local/file
```

The secret name ("my-secret") would be passed in as the secret "id" and the path to the file would be passed in as the "src".

alechirsch:
Looking forward to seeing ssh flag for compose files as well. We just started to use ssh for git keys and this is a blocker right now. Any update on a timeline for this?

eliliam:
@antiBaconMachine could we get this finally merged in?

alechirsch:
It looks like this PR does not address the ssh flag, and only does it for the secrets flag

Billy-:
> could we get this finally merged in?

Agreed, what do we need to do to make it happen?!

> @antiBaconMachine @ndeloof next steps here...? :)

This is the question! What else needs to be done, anyone? @ndeloof ? Or @thaJeztah ?

Is it mainly this?
> thaJeztah (Sebastiaan van Stijn) on 1 Apr
Note that we can't update the 3.7 schema, as it's already been released, so to add this property to the schema, it probably has to be added to the upcoming 3.9 schema in docker/cli@master/cli/compose/schema/data/config_schema_v3.9.json first
https://github.com/docker/compose/pull/7046#discussion_r401650994

@antiBaconMachine any chance you're still around and willing if this PR needs updates? If not, anyone else willing to contribute? I am more than willing, although may need some hand-holding :sweat_smile: 

---

Although part of me wants to say nothing and advocate merging this ASAP, I can't help but think this is a great suggestion:

> Just curious if anyone thinks using the existing secrets syntax (v3.1+) would be cleaner:
> 
> ```
> services:
>   my-application:
>     build:
>       secrets:
>         - my-secret
> 
> secrets:
>   my-secret:
>     file: /path/to/local/file
> ```
> 
> The secret name ("my-secret") would be passed in as the secret "id" and the path to the file would be passed in as the "src".

---

RE

> It looks like this PR does not address the ssh flag, and only does it for the secrets flag

Imo that seems out of scope for this PR.. looks like it might need a separate issue/PR based on this comment:

> @Teyras would need to look into details, but assuming BuildKit did define distinct build args for this purpose, we probably would better introduce `ssh` build option for this specific use-case. See #7025

TechTeam12:
Is there any update on this? Anything else that needs to be done aside from handling merge conflicts? Happy to help if needed.

pikeas:
Would love to see support for build secrets in compose!

otherguy:
Any update on this?

jhagege:
Hi, any plans to support build time secrets ?

luvncoco:
UP, any plans ? It's a much appreciated feature

ndeloof:
@crazy-max custom extension allow to experiment without the need to get a long debate on the spec on some abstract solution, but this doesn't mean we can't discuss on the spec while bake is used as a concrete implementation example.
Do you plan to contribute an actual proposal to the spec?

crazy-max:
@ndeloof

> Do you plan to contribute an actual proposal to the spec?

I think the proposal https://github.com/compose-spec/compose-spec/issues/120 from @chris-crone is already aligned with what the `x-bake` already offers but I agree that it should be part of the spec in the future for consistency.

jheld:
Can someone make a comment about the overall process remaining (e.g. what steps and governing bodies are involved and where those decisions are currently) since it _seems_ like there is some positivity here but still feels like "nothing" has happened?

ndeloof:
1. changes on the compose schema MUST be debated under Compose Specification => https://github.com/compose-spec/compose-spec/issues/81 (staled, mostly for missing contributors on this topics)
2. implementation will probably only happen on Compose v2 branch (v1 is in "best-effort" maintenance mode, i.e will receive security fixes but not much)

Would be nice to get buildx contributors involved here, as they introduced custom extensions for this purpose in `bake`.

ni11c:
Guys, the issue has been submitted more than two years ago, still no concrete progress on this ? Unless there is a decent tool chain that enables docker users to apply best security practices, secrets will continue to be provided in an insecure way for a long time, which is in the interest of nobody. I don't understand how such a crucial feature can take so long to make its way through the intricacies of the Compose Specification process. Thank you to all those involved to make this happen,

dfr-exnaton:
Checkout `v2` of compose: https://github.com/docker/compose/pull/9386

ndeloof:
`build.secrets` support has been introduced in Compose v2, and we don't plan to have significant updates on the v1 branch (security fixes/maintenance), so better adopt the new compose generation 😇

Kludex:
Just for reference, merged on https://github.com/compose-spec/compose-spec/pull/238

PR #7049, opened 2019-11-22
Update maintainers and add CODEOWNERS for github

ulyssessouza:
Update maintainers and add CODEOWNERS for github

Willkommen @ndeloof !

### DIFF
--- a/MAINTAINERS
+++ b/MAINTAINERS
@@ -11,6 +11,7 @@
 [Org]
 	[Org."Core maintainers"]
 		people = [
+			"ndeloof",
 			"rumpl",
 			"ulyssessouza",
 		]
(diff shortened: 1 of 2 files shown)


MAINTAINERS
```diff
@@ -77,6 +78,11 @@
 	Email = "mazz@houseofmnowster.com"
 	GitHub = "mnowster"
 
+	[people.ndeloof]
+	Name = "Nicolas De Loof"
+	Email = "nicolas.deloof@docker.com"
```
ndeloof: Please use nicolas.deloof@gmail.com for everything OSS

ulyssessouza: ok

PR #7052, opened 2019-11-25
Report image we can't pull and must be built

### LINKED ISSUE #6464: `docker-compose pull` fails if a service includes both image and build

ndeloof:
As we pull images, on `NotFound` but service has build definition, report required command to build the missing images.

Resolves #6464 

cc @justinmchase

### DIFF
--- a/compose/project.py
+++ b/compose/project.py
@@ -11,6 +11,8 @@ from os import path
 import enum
 import six
 from docker.errors import APIError
+from docker.errors import ImageNotFound
+from docker.errors import NotFound
 from docker.utils import version_lt
 
 from . import parallel
@@ -25,6 +27,7 @@ from .container import Container
 from .network import build_networks
 from .network import get_networks
 from .network import ProjectNetworks
+from .progress_stream import read_status
 from .service import BuildAction
 from .service import ContainerNetworkMode
 from .service import ContainerPidMode
@@ -619,46 +622,68 @@ class Project(object):
     def pull(self, service_names=None, ignore_pull_failures=False, parallel_pull=False, silent=False,
              include_deps=False):
         services = self.get_services(service_names, include_deps)
-        msg = not silent and 'Pulling' or None
 
         if parallel_pull:
-            def pull_service(service):
-                strm = service.pull(ignore_pull_failures, True, stream=True)
-                if strm is None:  # Attempting to pull service with no `image` key is a no-op
-                    return
+            self.parallel_pull(services, silent=silent)
 
+        else:
+            must_build = []
+            for service in services:
+                try:
+                    service.pull(ignore_pull_failures, silent=silent)
+                except (ImageNotFound, NotFound):
+                    if service.can_be_built():
+                        must_build.append(service.name)
+                    else:
+                        raise
+
+            if len(must_build):
+                log.warning('Some service image(s) must be built from source by running:\n'
+                            '    docker-compose build {}'
+                            .format(' '.join(must_build)))
+
+    def parallel_pull(self, services, ignore_pull_failures=False, silent=False):
+        msg = 'Pulling' if not silent else None
+        must_build = []
+
+        def pull_service(service):
+            strm = service.pull(ignore_pull_failures, True, stream=True)
+
+            if strm is None:  # Attempting to pull service with no `image` key is a no-op
+                return
+
+            try:
                 writer = parallel.get_stream_writer()
-
                 for event in strm:
                     if 'status' not in event:
                         continue
-                    status = event['status'].lower()
-                    if 'progressDetail' in event:
-                        detail = event['progressDetail']
-                        if 'current' in detail and 'total' in detail:
-                            percentage = float(detail['current']) / float(detail['total'])
-                            status = '{} ({:.1%})'.format(status, percentage)
-
+                    status = read_status(event)
                     writer.write(
                         msg, service.name, truncate_string(status), lambda s: s
                     )
+            except (ImageNotFound, NotFound):
+                if service.can_be_built():
+                    must_build.append(service.name)
+                else:
+                    raise
 
-            _, errors = parallel.parallel_execute(
-                services,
-                pull_service,
-                operator.attrgetter('name'),
-                msg,
-                limit=5,
-            )
-            if len(errors):
-                combined_errors = '\n'.join([
-                    e.decode('utf-8') if isinstance(e, six.binary_type) else e for e in errors.values()
-                ])
-                raise ProjectError(combined_errors)
+        _, errors = parallel.parallel_execute(
+            services,
+            pull_service,
+            operator.attrgetter('name'),
+            msg,
+            limit=5,
+        )
 
-        else:
-            for service in services:
-                service.pull(ignore_pull_failures, silent=silent)
+        if len(must_build):
+            log.warning('Some service image(s) must be built from source by running:\n'
+                        '    docker-compose build {}'
+                        .format(' '.join(must_build)))
+        if len(errors):
+            combined_errors = '\n'.join([
+                e.decode('utf-8') if isinstance(e, six.binary_type) else e for e in errors.values()
+            ])
+            raise ProjectError(combined_errors)
 
     def push(self, service_names=None, ignore_push_failures=False):
         unique_images = set()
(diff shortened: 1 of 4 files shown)


compose/project.py
```diff
@@ -619,46 +621,68 @@ def _get_convergence_plans(self, services, strategy, always_recreate_deps=False)
...
-            def pull_service(service):
-                strm = service.pull(ignore_pull_failures, True, stream=True)
-                if strm is None:  # Attempting to pull service with no `image` key is a no-op
-                    return
+            self.parallel_pull(services, silent=silent)
 
-                writer = parallel.get_stream_writer()
+        else:
+            for service in services:
+                must_build = []
```
ulyssessouza: Shouldn't it be declared outside of the `for`?

compose/project.py
```diff
@@ -619,46 +621,68 @@ def _get_convergence_plans(self, services, strategy, always_recreate_deps=False)
...
+                            .format(' '.join(must_build)))
+
+    def parallel_pull(self, services, ignore_pull_failures=False, silent=False):
+        msg = not silent and 'Pulling' or None
+        must_build = []
+
+        def pull_service(service):
+            strm = service.pull(ignore_pull_failures, True, stream=True)
 
+            if strm is None:  # Attempting to pull service with no `image` key is a no-op
```
ulyssessouza:
```suggestion
            if not strm:  # Attempting to pull service with no `image` key is a no-op
```

compose/project.py
```diff
@@ -619,46 +621,68 @@ def _get_convergence_plans(self, services, strategy, always_recreate_deps=False)
...
+                    service.pull(ignore_pull_failures, silent=silent)
+                except NotFound:
+                    if service.can_be_built():
+                        must_build.append(service.name)
+                    else:
+                        raise
+
+            if len(must_build):
+                log.warning('Some service image(s) must be built from source by running:\n'
+                            '    docker compose build {}'
```
ulyssessouza:
```suggestion
                            '    docker-compose build {}'
```

compose/project.py
```diff
@@ -619,46 +621,68 @@ def _get_convergence_plans(self, services, strategy, always_recreate_deps=False)
...
+                            .format(' '.join(must_build)))
+
+    def parallel_pull(self, services, ignore_pull_failures=False, silent=False):
+        msg = not silent and 'Pulling' or None
```
ulyssessouza: Not sure if I understood this line, but the following equivalent would be more readable.

```suggestion
        msg = 'Pulling' if not silent else None
```

compose/project.py
```diff
@@ -619,46 +621,68 @@ def _get_convergence_plans(self, services, strategy, always_recreate_deps=False)
...
+            msg,
+            limit=5,
+        )
 
-        else:
-            for service in services:
-                service.pull(ignore_pull_failures, silent=silent)
+        if len(must_build):
+            log.warning('Some service image(s) must be built from source by running:\n'
+                        '    docker compose build {}'
```
ulyssessouza:
```suggestion
                        '    docker-compose build {}'
```

compose/project.py
ndeloof: was by initial code, but can change this indeed

compose/project.py
ndeloof: oups

compose/project.py
ndeloof: took me time indeed to understand this syntax used in original code.

justinmchase:
Hey, thanks for this!

I was thinking about it and I just wanted to add a little more context, I know previously I said that just not producing an error in the case of a service with image+build would solve my case (and it does) but I did just want to add that it is kind of a coincidence because my image was named something that doesn't happen to match an image in the docker repository.

For example mine is:
```
service:
  image: ambassador
  build: .
```

There is no image in docker hub simply called `ambassador` so it fails to pull and then goes down this route. However if I had happened to name the image `node` for example... then it would succeed to pull and probably would not do as I expect.

I'm not sure what the right thing to do there is, but I figured you'd want to consider it. My immediate thought would be to allow local image names to have characters that are illegal for remote images so it could just never pull in this case, such as `image: $node`... but I'm not sure how wise that is. Or maybe just a flag on the service: `local: true`. Or maybe its just always recommended they use one called `image: local` and that is special cased, I don't know.

Anyway thank you and I think this is a good improvement and compromise.

compose/project.py
```diff
@@ -619,46 +622,68 @@ def _get_convergence_plans(self, services, strategy, always_recreate_deps=False)
...
+            msg,
+            limit=5,
+        )
 
-        else:
-            for service in services:
-                service.pull(ignore_pull_failures, silent=silent)
+        if len(must_build):
+            log.warning('Some service image(s) must be built from source by running:\n'
+                        '    docker-compose build {}'
```
justinmchase: 👍

ndeloof:
there's not such thing as a "local" image. Images are by design built so they can be shared by registries. To avoid collision with a Dockerhub image, you should just always use your own registry names : justinmchase/ambassador of mycompany/ambassador

ulyssessouza:
LGTM

PR #7066, opened 2019-11-28
Use Python3 for macOS build environment

chris-crone:
With the deprecation of Python 2 coming soon, explicitly use
Python 3.


ndeloof:
LGTM
please note we have a py2 branch for incoming drop of pyhton2 support starting with 1.26

PR #7075, opened 2019-12-02
use stash to share baseImage between steps

ndeloof:
As we run build in a Jenkins pipeline, no need to push/pull to dockerhub using git SHA as unique tag, we can just stash the docker image

Also adopted declarative pipeline syntax where possible thanks to preleminary work on https://github.com/docker/compose/pull/6932


script/Jenkinsfile.fossa
```diff
@@ -1,20 +0,0 @@
-pipeline {
```
ulyssessouza: :heart:

PR #7079, opened 2019-12-04
handle the whole release process within Jenkinsfile

ndeloof:
Revisit the CI build process to use declarative syntax (mostly)

We unfortunately can't yet use `matrix` declarative directive which would just fit our needs, as 1.50 is not (yet) validated by CloudBees' beekeeper
(cc @bitwiseman)

image


thaJeztah:
> We unfortunately can't yet use matrix declarative directive

oh, that sounds good. gonna look at that one to see how we can use that in future 👍

PR #7085, opened 2019-12-06
Remove "bundle" subcommand and support for DAB files

thaJeztah:
> relates to https://github.com/docker/cli/issues/2215 "[epic] remove experimental "docker (stack) deploy" using "dab" files"

Deploying stacks using the "Docker Application Bundle" (`.dab`) file format was introduced as an experimental feature in Docker 1.13 / 17.03, but superseded by support for Docker Compose files in the CLI.

With no development being done on this feature, and no active use of the file format, support for the DAB file format and the top-level `docker deploy` command (hidden by default in 19.03), will be removed from the CLI, in favour of `docker stack deploy` using compose files.

This patch removes the `docker-compose bundle` subcommand from Docker Compose, which was used to convert compose files into DAB files (and given the above, will no longer be needed).


thaJeztah:
Fixed the linting issues, all green now 👍

thaJeztah:
@ulyssessouza ptal

PR #7099, opened 2019-12-18
Implement -f / --file argument in container-script

glance-:
When using -f / --file argument to container-script, and pointing it to
some path which isn't beneath cwd or home, the call would have failed
previously, like:
$ docker-compose -f /tmp/docker-compose.yml up
ERROR: .FileNotFoundError: [Errno 2] No such file or directory: '/tmp/docker-compose.yml'

After this patch:
$ docker-compose -f /tmp/docker-compose.yml up
Creating tmp_alpine_1 ... done
Attaching to tmp_alpine_1
tmp_alpine_1 exited with code 0

It also aborts the parsing at docker-compose verbs so it doesn't
confuse docker-compose args with anything thats ment for a command in
the container:
$ docker-compose -f /tmp/docker-compose.yml run alpine echo --file /tmp/kex.xml
--file /tmp/kex.xml

### DIFF
--- a/script/run/run.sh
+++ b/script/run/run.sh
@@ -34,11 +34,40 @@ fi
 if [ "$(pwd)" != '/' ]; then
     VOLUMES="-v $(pwd):$(pwd)"
 fi
+if [ -z "$COMPOSE_FILE" ] ; then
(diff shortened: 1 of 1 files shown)


glours:

Issue #7047, opened 2019-11-22
Reconcile docker-compose shcema on docker/cli vs docker/compose 


ndeloof:
Docker Compose schema on  https://github.com/docker/cli/tree/master/cli/compose/schema/data is considered by maintainers as the leading one, and should be the place we use to any proposal for schame enhancements.

We need to double check https://github.com/docker/compose/tree/master/compose/config didn't diverge, and enforce we don't make changes here that have not been already applied to the leading schema.

Idealy we should find a mechanism to avoid duplication, and define a canonical location for compose schemas.

PR #7124, opened 2020-01-06
WIP: Forward device requests

yoanisgil:
Add support for Nvidia GPUs (solves #6691)

### DIFF
--- a/compose/config/config_schema_v3.7.json
+++ b/compose/config/config_schema_v3.7.json
@@ -151,6 +151,8 @@
 
         "external_links": {"type": "array", "items": {"type": "string"}, "uniqueItems": true},
         "extra_hosts": {"$ref": "#/definitions/list_or_dict"},
+        "device_requests": {"$ref": "#/definitions/device_requests"},
+        "": {"$ref": "#/definitions/list_or_dict"},
         "healthcheck": {"$ref": "#/definitions/healthcheck"},
         "hostname": {"type": "string"},
         "image": {"type": "string"},
@@ -597,6 +599,21 @@
           }
         }
       }
-    }
+    },
+    "device_requests": {
+      "id": "#/definitions/device_requests",
+      "type": "array",
+      "items": {
+        "type": "object",
+        "properties": {
+          "driver": {"type": "string"},
+          "count": {"type": "number"},
+          "device_ids": {"$ref": "#/definitions/list_of_strings"},
+          "capabilities": {"$ref": "#/definitions/list_of_strings"},
+          "options": {"type": "object", "additionalProperties": {"type": "string"}}
+        },
+        "additionalProperties": false
+      }
+    }    
   }
 }
--- a/compose/config/config_schema_v3.7.json
+++ b/compose/config/config_schema_v3.7.json
@@ -614,6 +614,6 @@
         },
         "additionalProperties": false
       }
-    }    
+    }
   }
 }
--- a/compose/service.py
+++ b/compose/service.py
@@ -988,7 +988,7 @@ class Service(object):
         if 'cpus' in options:
             nano_cpus = int(options.get('cpus') * NANOCPUS_SCALE)
 
-        device_requests = options.get('device_requests')
+        device_requests = options.get('device_requests', [])
 
         for device_request in device_requests:
             if 'capabilities' not in device_request:
--- a/compose/service.py
+++ b/compose/service.py
@@ -114,6 +114,7 @@ HOST_CONFIG_KEYS = [
     'userns_mode',
     'volumes_from',
     'volume_driver',
+    'device_requests', 
 ]
 
 CONDITION_STARTED = 'service_started'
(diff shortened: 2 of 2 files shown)


yoanisgil:
This PR is pretty much WIP as I have tested only one use case and also depends on https://github.com/docker/docker-py/pull/2471

yoanisgil:
@rumpl @ndeloof @ulyssessouza I'm happy to make this PR more compliant (tests, documentation, etc) as long as you think it is heading int he right direction.

compose/config/config_schema_v3.7.json
```diff
@@ -597,6 +599,21 @@
           }
         }
       }
+    },
+    "device_requests": {
```
ndeloof: leading schema definition is https://github.com/docker/cli/tree/master/cli/compose/schema/data 
see https://github.com/docker/compose/issues/7047
I'm looking into this schame duplication issue to be resolved asap

compose/config/config_schema_v3.7.json
```diff
@@ -151,6 +151,8 @@
 
         "external_links": {"type": "array", "items": {"type": "string"}, "uniqueItems": true},
         "extra_hosts": {"$ref": "#/definitions/list_or_dict"},
+        "device_requests": {"$ref": "#/definitions/device_requests"},
```
ndeloof: Within schema v3, runtime constraints have been moved into `deployment.resources`. Sounds a better place for such device allocation request

compose/config/config_schema_v3.7.json
yoanisgil: That actually raises a very good point. Maybe the GPU allocation request should be based on [generic_resources](https://github.com/docker/cli/blob/master/cli/compose/schema/data/config_schema_v3.7.json#L439)?. Just sayin because of this: https://github.com/docker/cli/blob/master/cli/compose/loader/full-example.yml

But then that will require a translation to a [DeviceRequest](https://github.com/docker/docker-ce/blob/531930f3294c31db414f17f80fa8650d4ae66644/components/engine/api/types/container/host_config.go#L272)

compose/config/config_schema_v3.7.json
yoanisgil: I can move but seems to me like it's better to wait until you figure out how to reconcile schema definitions between docker and compose before making any other change ...

lshamis:
I thought a docker-compose file tries to match the docker command line.

If I want to set set `ipc`, `pid`, `privileged`, ... the docker cli would look like:
```
docker run --ipc=host --pid=host --privileged=true ...
```
and the docker-compose looks like:
```yaml
services:
    myservice:
         ipc: "host",
         pid: "host",
         privileged: true,
         ...
```

With newer versions of docker, the command line to use the nvidia runtime:
```
docker run --gpus=all --rm nvidia/cuda nvidia-smi
```

so I would assume that the docker-compose would look like:
```yaml
services:
    myservice:
         gpus: "all",
         ...
```

but with this PR (according to the comments https://github.com/docker/compose/issues/6691), gpu usage would look like:
```yaml
services:
    gpu:
        image: 'nvidia/cuda:9.0-base'
        command: 'nvidia-smi'
        device_requests:
            - capabilities:
               - "gpu"
```

which seems much more complicated to me.

amalic:
Hard to believe, that this is still an open issue!

hadim:
Is there anyone working on this PR still?

xkortex:
@lshamis I believe the motivation behind the `device_requests` syntax is the long-term ongoing goal of ensuring swarm mode is as generic as possible. `docker run --gpus=all` makes sense in the scope of `docker run` but not so much in `docker swarm`.  However I agree, the `requests` notation is more confusing. As awesome as the aspirations for `swarm` are, there is still a user base reliant on compose for single-machine orchestration.  

`network_mode: host` is a compose v3 field which is incompatible with swarm. Is it plausible to use this syntax for local-machine-only, and throw an error if used in swarm mode, in a similar manner to `network_mode`? 

```
services:
    myservice:
         gpus: "all",
         ...
```

I think it's more violating of law-of-least-surprisal to lack this syntax when `docker run` supports it. 

Nearly my entire department is still using 2.X syntax for `runtime: nvidia` because we use gpu enabled containers day-in and day-out and cannot always set `daemon.json` with a default runtime.

yoanisgil:
> Is there anyone working on this PR still?

I'm not (mostly because of the lack of feedback and/or interest).

hadim:
FYI: https://github.com/docker/docker-py/pull/2471 has been merged.

visheratin:
Can someone tell what stops the PR from being merged?

edurenye:
There are conflicting files, it needs to be rebased I guess.

PhotoTeeborChoka:
Guys, (@rumpl @ndeloof @ulyssessouza), would it be possible to give the necessary feedback to @yoanisgil so that he can finalize this PR or so that somebody can take over and finalize it so that it can be merged? ML applications, as well as GPU based ones cannot be used very well with `docker-compose` unless the compatibility with docker 19.03 is complete regarding the `--gpus` flag.

ysyyork:
Any progress on merging this into master?

alexandrecharpy:
Agree with PhotoTeerborChoka. If this is just a "rebase"/conflict issue, would it speed up things if I rebase a fork from @yoanisgil (with his consent) ?

yoanisgil:
@alexandrecharpy  please go ahead. At the time I submitted the PR I had the time and will to make this happen but unfortunately that's no longer the case.

glours:

PR #7145, opened 2020-01-14
Refactor mutable default values.

yukihira1992:
Replaced mutable default values with None.
https://docs.quantifiedcode.com/python-anti-patterns/correctness/mutable_default_value_as_argument.html

### DIFF
--- a/compose/cli/command.py
+++ b/compose/cli/command.py
@@ -40,7 +40,8 @@ SILENT_COMMANDS = {
 }
 
 
-def project_from_options(project_dir, options, additional_options={}):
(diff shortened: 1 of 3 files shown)


compose/cli/command.py
```diff
@@ -40,7 +40,9 @@
 }
 
 
-def project_from_options(project_dir, options, additional_options={}):
+def project_from_options(project_dir, options, additional_options=None):
+    if additional_options is None:
```
ulyssessouza:
```suggestion
    additional_options = additional_options or {}:
```

Same for the others

PR #7177, opened 2020-01-22
Tag as `x.y.z` without `v` prefix

### LINKED ISSUE #7168: Latest release tagged with 'v'
## Description of the issue

The tagging naming scheme changed with 1.24.2. Was this intended? This may break some automation (it did for me).

This isn't a bug, but I thought I'd report it since it seems that bootstraping via curl is fairly common. We had startup scripts like this:

```
export latest_stable_compose=`git ls-remote https://github.com/docker/compose | grep refs/tags | grep -oP "[0-9]+\.[0-9][0-9]+\.[0-9]+$" | tail -n 1`
sudo curl -L https://github.com/docker/compose/releases/download/${latest_stable_compose}/docker-compose-$(uname -s)-$(uname -m) -o /usr/local/bin/docker-compose
sudo chmod +x /usr/local/bin/docker-compos
```

That we probably snagged from some other example. In order to fix for our purposes, it was just a small change to the regex to optional consider leading `v` (`v*`):

```
export latest_stable_compose=`git ls-remote https://github.com/docker/compose | grep refs/tags | grep -oP "v*[0-9]+\.[0-9][0-9]+\.[0-9]+$" | tail -n 1`
```

```
docker-compose version 1.25.2, build 698e2846
docker-py version: 4.1.0
CPython version: 3.7.5
```

```
n/a
```

```
n/a
```

1. Automate docker-compose download with semver like regex, not considering `v` prefix

### Observed result

Automation fails - docker-compose not downloaded

### Expected result

docker-compose downloaded

ndeloof:
Resolves #7168

### DIFF
--- a/script/release/README.md
+++ b/script/release/README.md
@@ -5,5 +5,5 @@ The release process is fully automated by `Release.Jenkinsfile`.
 ## Usage
 
 1. edit `compose/__init__.py` to set release version number
-1. commit and tag as `v{major}.{minor}.{patch}`
+1. commit and tag as `{major}.{minor}.{patch}`
 1. edit `compose/__init__.py` again to set next development version number


thaJeztah:
Should we keep this PR for the _next_ release, and for the current one just manually push a tag using the same commit (698e2846a82dbc48c14a017890f826e3d63ab783) ?

ndeloof:
afaik adding a tag won't create an alias for the release download URL.

thaJeztah:
right, but after tagging, we can create a new release and upload the same binaries there?

Issue #3513, opened 2016-05-26
pull requires valid build path


tanmayambre:
```
docker-compose version
docker-compose version 1.7.0, build 0d7bf73
docker-py version: 1.8.0
CPython version: 2.7.9
OpenSSL version: OpenSSL 1.0.1e 11 Feb 2013

```

I have a docker-compose file 

```
version: '2'
services:
    xyzservice:
    image: xxx-image
    container_name: xxx_container
    build:
      context: ./somecontext
      dockerfile: Dockerfile.xxx
    networks:
      - some-network

```

When I want to build the image using `docker-compose build`, I understand that the **_Dockerfile.xxx**_ is absolutely required.

But if I just want to pull the image using `docker-compose pull` why is the **Dockerfile.xxx** required?

I get the following error on pull command

```
ERROR: build path ./somecontext either does not exist, is not accessible, or is not a valid URL.
```

shin-:
You don't need the `build` section if you're just going to pull the image. Your compose file should be

```yaml
version: '2'
services:
    xyzservice:
    image: xxx-image
    container_name: xxx_container
    networks:
      - some-network

```

tanmayambre:
Thanks,
My scenario is the developer requires the buildpath to build & push the container while if someone wants to just run the container without build just pulling mage using same compose file.

fvanderbiest:
This issue is annoying for our [use case](https://github.com/georchestra/georchestra/blob/16.12/docs/docker.md) where the compose file is used both to build and run the app. Valid build paths should **not** be required to run `docker pull` IMO.

jeserkin:
@shin- does what if I am getting it without `build` section?
My `docker-compose.yml` is as follows:
```yml
version: "3"
services:
  app:
	image: busybox
    networks:
      - backend
  db:
    #image: postgres
	image: busybox
    networks:
      - backend

networks:
  backend:
    # Use a custom driver which takes special options
    driver: bridge
```

Elyytscha:
this is annoying, just saying

zypA13510:
Are you sure this is `kind/question`? @docker/maintainer

As noted by @**fvanderbiest** and other people, there are valid use cases where pulling without building is required. Could you please consider this as an enhancement?

yamac-kurtulus:
This is seriously annoying and makes no sense. Could you please implement this?

nojvek:
Just hit this, this is really annoying indeed.

[...]

PR #7213, opened 2020-02-03
Add --pull arg to "up" command

GabrielBB:
Ability to do: 

 `docker-compose -f my-docker-compose.yml up --pull` 

Instead of :

`docker-compose -f my-docker-compose.yml pull && docker-compose -f my-docker-compose.test.yml up`. 

Why:

- It's shorter

- Less prone to errors since you don't have to type the yml file name twice (Which means double chance for typos)

- It has better performance because `&&` throws away the asynchrony

- Makes the API more consistent since there is already a `--build` flag and one expect to be a `--pull` one.

One example without this flag that looks weird to me (maybe is just me) is this:

`docker-compose pull && docker-compose up --build`

There it looks like `pull` and `build` are two features with different degrees of importance, when in fact, they don't, the condition to do one thing or another is even in the same function in the code:

```
def ensure_image_exists(self, do_build=BuildAction.none, silent=False, cli=False):
        if self.can_be_built() and do_build == BuildAction.force:
            self.build(cli=cli)
            return

        try:
            self.image()
            return
        except NoSuchImageError:
            pass

        if not self.can_be_built():
            self.pull(silent=silent)
            return

        if do_build == BuildAction.skip:
            raise NeedsBuildError(self)

        self.build(cli=cli)

```

This looks a better fit:

`docker-compose up --build --pull`

I added unit tests and acceptance tests


nojvek:
I really hope this still doesn't have the same problem as #3513.

It's really a pain to have to separate but almost identical docker-compose.yaml files.

Ideally I would love to see a single command `docker-compose restart/up --pull` that gets the latest built images and gracefully (re)starts all containers with the latest code.

GabrielBB:
> I really hope this still doesn't have the same problem as #3513.

The pull arg will have the same “issues” as docker-compose pull. It calls the same code.

> Ideally I would love to see a single command `docker-compose restart/up --pull` that gets the latest built images and gracefully (re)starts all containers with the latest code.

That will be possible with this PR

ndeloof:
I wonder it would make sense to introduce an equivalent for kubernetes `pull_strategy` so one can define on a per-service basis how to update images vs rebuild.

thaJeztah:
it should at least not be a boolean, but preferably more like https://github.com/docker/cli/pull/1498

There's also the `--resolve-image` option includes `docker stack deploy`, which is slightly related

tomaswarynyca:
Could you please review and approve this change that many of us need?
@ndeloof @rumpl @ulyssessouza

tomaswarynyca:
@GabrielBB could you sign the commits properly and fix the conflicts?
https://github.com/docker/compose/pull/7213#issuecomment-581676830

GabrielBB:
> @GabrielBB **[could you sign the commits properly](https://github.com/docker/compose/pull/7213#issuecomment-581676830) and fix the conflicts?**

@tomaswarynyca I messed by branch while trying to fix the commits signing. I had to open a new PR: https://github.com/docker/compose/pull/8328

PR #7217, opened 2020-02-06
Add v3.8 schema support

aiordache:
Add v3.8 schema support
- Update version mappings
- Limit service scale to 'max_replicas_per_node' 

```
$ cat docker-compose.yaml
version: "3.8"
services:
  test:
    image: nginx
    deploy:
      mode: replicated
      replicas: 3
      placement:
        max_replicas_per_node: 2
```
```
$ docker-compose --compatibility up -d
WARNING: Scale is limited to 2 ('max_replicas_per_node' field).
Creating network "schema_default" with the default driver
Creating schema_test_1 ... done
Creating schema_test_2 ... done
```
```
$ docker ps        
CONTAINER ID        IMAGE                 COMMAND                  CREATED              STATUS              PORTS                              NAMES
fc6510771d3a        nginx                 "nginx -g 'daemon of…"   41 seconds ago      Up 41 seconds       80/tcp                             schema_test_1
464e98b579f4        nginx                 "nginx -g 'daemon of…"   41 seconds ago      Up 41 seconds       80/tcp                             schema_test_2

```

Signed-off-by: Anca Iordache <anca.iordache@docker.com>

Resolves https://github.com/docker/compose/issues/6530


compose/const.py
```diff
@@ -74,4 +76,5 @@
     API_VERSIONS[COMPOSEFILE_V3_5]: '17.06.0',
     API_VERSIONS[COMPOSEFILE_V3_6]: '18.02.0',
     API_VERSIONS[COMPOSEFILE_V3_7]: '18.06.0',
+    API_VERSIONS[COMPOSEFILE_V3_8]: '19.03.0',
```
ndeloof: as both `COMPOSEFILE_V3_7` and `COMPOSEFILE_V3_8` == `'1.38'` doesn't this override `API_VERSIONS['1.28']` just being set to `'18.06.0'` ?

compose/config/config.py
```diff
@@ -990,12 +990,14 @@ def translate_deploy_keys_to_container_config(service_dict):
...
-        k for k in ['endpoint_mode', 'labels', 'update_config', 'rollback_config', 'placement']
+        k for k in ['endpoint_mode', 'labels', 'update_config', 'rollback_config']
         if k in deploy_dict
     ]
 
     if 'replicas' in deploy_dict and deploy_dict.get('mode', 'replicated') == 'replicated':
-        service_dict['scale'] = deploy_dict['replicas']
+        scale = deploy_dict.get('replicas', 1)
+        max_replicas = deploy_dict.get('placement', {}).get('max_replicas_per_node', scale)
+        service_dict['scale'] = min(scale, max_replicas)
```
ndeloof: will silently ignore constraint mismatch to provision N replicas on single node docker engine. Should at least warn user imho

compose/const.py
```diff
@@ -57,6 +58,7 @@
     COMPOSEFILE_V3_5: '1.30',
     COMPOSEFILE_V3_6: '1.36',
     COMPOSEFILE_V3_7: '1.38',
+    COMPOSEFILE_V3_8: '1.40',
```
ndeloof: Maybe I misunderstand this constant, but afaik 3.8 do not add any new field to require an API upgrade vs 3.7, does it?

aiordache: Yes, I set it back to 1.38. This should be the minimum API version we need for the v3.8 schema.

compose/const.py
aiordache: Yep, but I guess this mapping is more for us to easily view the minimum version needed for each compose file.

PR #7259, opened 2020-02-29
Fix exception causes in main.py

cool-RR:
In all the places I've touched, before the change, you would get a 2 tracebacks for the 2 exceptions, with this message between them:

    During handling of the above exception, another exception occurred:

This text is not appropriate here. It basically means "Your error-handling code has a bug in it" which is not the case here, your error-handling code is working well. After my change, the correct message will be displayed:

    The above exception was the direct cause of the following exception:

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -482,7 +482,7 @@ class TopLevelCommand(object):
         try:
             container = service.get_container(number=index)
         except ValueError as e:
-            raise UserError(str(e))
+            raise UserError(str(e)) from e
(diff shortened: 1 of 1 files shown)


alex-ubitec:
The issue you have right now is that in python2.7 `raise ... from...` [is not part of the syntax](https://docs.python.org/2.7/reference/simple_stmts.html#raise).

Your options are to either use `if six.PY2:...`, everywhere, or to instead wait for when python2 support is completely dropped (Not a very distant future)

cool-RR:
Let's wait.

glours:

PR #7266, opened 2020-03-03
Show ranges when you run docker-compose ps like docker ps.

### LINKED ISSUE #6910: docker-compose ps should show port ranges as docker ps does
`docker ps` gives a sensible output for ranges `1-5431/tcp, 5433-30000/tcp, 0.0.0.0:5432->5432/tcp` but frustratingly `docker-compose ps` gives `1/tcp, 10/tcp, 100/tcp, 1000/tcp, 10000/tcp, 10001/tcp, 10002/tcp, 10003/tcp, 10004/tcp, 10005/tcp, 10006/tcp, 10007/tcp`.

While large ranges like this are a bit nuts and that can be debated somewhere else, it would be helpful for docker-compose to match docker in the output format for `ps` in this case.

archibishop:
Signed-off-by: Wagubi Brian <wagubib@gmail.com>

#### Description
Show ports as ranges when you run docker-compose ps as docker ps does.

##### Before
Screen Shot 2020-03-03 at 10 53 06

##### After
Screen Shot 2020-03-03 at 10 52 10

Resolves #6910

### DIFF
--- a/compose/container.py
+++ b/compose/container.py
@@ -2,6 +2,8 @@ from __future__ import absolute_import
 from __future__ import unicode_literals
 
 from functools import reduce
+from itertools import groupby
+from operator import itemgetter
 
 import six
 from docker.errors import ImageNotFound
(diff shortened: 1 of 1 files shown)


glours:

PR #7283, opened 2020-03-11
support specifying a tar file as context

### LINKED ISSUE #1209: Add support for pointing to tarball contexts in the 'build' parameter.

MichaelKim0407:
Resolves #1209 

Implemented the support for both docker-py and cli build. Very minimal changes were required since both docker-py and cli already support the behavior.

I probably need to add tests and documentation somewhere, which I haven't done yet. If anyone can point me to the correct locations, I'll be happy to add them

### DIFF
--- a/compose/service.py
+++ b/compose/service.py
@@ -1083,33 +1083,50 @@ class Service(object):
             )
 
         builder = self.client if not cli else _CLIBuilder(progress)
-        build_output = builder.build(
-            path=path,
-            tag=self.image_name,
-            rm=rm,
-            forcerm=force_rm,
-            pull=pull,
-            nocache=no_cache,
-            dockerfile=build_opts.get('dockerfile', None),
-            cache_from=self.get_cache_from(build_opts),
-            labels=build_opts.get('labels', None),
-            buildargs=build_args,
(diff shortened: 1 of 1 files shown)


benmosher:
is there a proper way to "bump" this? would love this feature. and I see that #1209 was recently closed for staleness.

ndeloof:
Compose v1 is EoL, compose v2 codebase should be considered for new features

glours:

MichaelKim0407:
Lol this PR was opened more than 2 years ago.

I'm not sure if I want to spend the effort to create another one if this is how it works here.

PR #7293, opened 2020-03-14
Show diff on pre-commit failure

alexrecuenco:
This would provide a better CI experience, since `./scripts/tests/default` is executed within a docker and the `pre-commit` doesn't display the changes you need to make to your code if it fails inside the docker.

Signed-off-by: Alejandro G. Recuenco <alejandrogonzalezrecuenco@gmail.com>

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -32,7 +32,7 @@ deps =
     pre-commit
 commands =
     pre-commit install
-    pre-commit run --all-files
+    pre-commit run --all-files --show-diff-on-failure
 
 # Coverage configuration
 [run]


alexrecuenco:
Sorry for the multiple changes, I think the `CONTRIBUTING.md` is not up-to-date, maybe we can put an issue for that


NOTE: The CONTRIBUTORS.md leads directly to [moby `CONTRIBUTING.md`](https://github.com/moby/moby/blob/master/CONTRIBUTING.md#sign-your-work) which explicitly states 

> Use your real name (sorry, no pseudonyms or anonymous contributions.)

Issue #6358, opened 2018-11-21
Support for docker build --secret for build-time docker secrets.


mikhail-khodorovskiy:
**Is your feature request related to a problem? Please describe.**
https://medium.com/@tonistiigi/build-secrets-and-ssh-forwarding-in-docker-18-09-ae8161d066 described how to provide secret mounts available at build times.

> docker build --secret id=mysite.key,src=path/to/mysite.key .

Can be referenced during docker build:

> RUN --mount=type=secret,id=mysite.key,required <command-to-run>

docker compose build command: https://docs.docker.com/compose/compose-file/#build currently has no support for the secrets section

**Describe the solution you'd like**
build command should allow referencing the secrets defined in the same docker-compose file

**Describe alternatives you've considered**
The alternatives is to use secret-containing file contents as build arguments.  The solution is cumbersome since the secret has to be splatted so it can become available for Dockerfile commands which may cause it to become part of a docker layer, i.e it may leak the secret value.

**Additional context**

ciaranmcnulty:
This would be really useful - right now we're doing a `docker build` using the new option and matching the image name to the one that compose will generate, as part of a shell wrapper.

Native support would be great.

yinzara:
Agree! SUPER useful especially in a setting where you're building artifacts that have to access private repositories of some sort. Nearly impossible to do without this.

While I would support the idea of a 'flags' attribute, I think because of the specifics of this feature, a real language element would be better.  Maybe make a separate feature request specifically for this.

`
services:
  foo:
    build:
      context: .
      dockerfile: Dockerfile
      target: secure
      secrets:
        -  id: mysite.key
           src: path/to/mysite.key
`

or possibly assuming there will never be other attributes to the command:
`
services:
  foo:
    build:
      context: .
      dockerfile: Dockerfile
      target: secure
      secrets:
        'mysite.key': path/to/mysite.key
`

ciaranmcnulty:
@yinzara flags attribute was to address next time this happens - a specific syntax for secrets would be best

We should also support:

```
services:
  foo:
    build:
        # ...
        ssh: default
```

chris-crone:
Thanks for the suggestion! This requires that we support BuildKit in the Python SDK, see the issue [here](https://github.com/docker/docker-py/issues/2230).

mattsrobot:
Has this ticket moved on, is there a way to use ssh secrets in docker compose?

chris-crone:
[...]

PR #7296, opened 2020-03-15
Introduce a new env var "COMPOSE_DOCKER_CLI_BUILD_EXTRA_ARGS"

### LINKED ISSUE #7025: Enable ssh option as build parameter for buildkit
When building a container we need access to other Resources like ssh and dont want to expose a private key in the container. 

Support of `--ssh` in the build process that already works in buildkit.

**Additional context**
I tried `1.25.0-rc4`, but it does not support `--ssh` option. Tried the same as in bug 6440:

```
export COMPOSE_DOCKER_CLI_BUILD=1
export DOCKER_BUILDKIT=1
docker-compose build --ssh default my_image
```
This would be a major security improvement and save us so much pain.

Thanks,
bert

moriyoshi:
This patch allows one to pass arbitrary arguments in the docker command invocation through `COMPOSE_DOCKER_CLI_BUILD_EXTRA_ARGS`.

Resolves #7025.


Syntactical01:
Only note, to be conistent with `COMPOSE_DOCKER_CLI_BUILD` you should probably use `COMPOSE_DOCKER_CLI_BUILD_EXTRA_ARGS`. That being said can someone review this? Docker compose REALLY needs support for the `--ssh` argument.

mecampbellsoup:
Please merge!! Like many others we want to use `--ssh` when running builds thru docker compose. Thanks.

itsjef:
Any progress update on this PR? 🤔

mecampbellsoup:
Workaround:

```bash
COMPOSE_DOCKER_CLI_BUILD=1 docker-compose build {app-name-that-uses-build-kit}
```
Also here is my Docker daemon config:

```json
{
  "debug": true,
  "experimental": true,
  "features": {
    "buildkit": true
  }
}
```

ppg:
That workaround doesn't help with `--ssh` or any other buildkit options that aren't explicitly supported by compose right? That's just a way to save typing `DOCKER_BUILDKIT=1`, i.e. you could simply do:
```
COMPOSE_DOCKER_CLI_BUILD=1 DOCKER_BUILDKIT=1 docker-compose build {app-name-that-uses-build-kit}
```
But neither of them help with the fact the argument list is hardcoded (see https://github.com/docker/compose/blob/440c94ea7a7e62b3de50722120ca34c4e818205a/compose/service.py#L1790-L1803) and thus we will always be playing catchup to a buildkit build's options unless we have some pass through like this, specifically in this case the desire to pass `--ssh`.

thaJeztah:
I don't think we should implement this; this would be a global option, which means that in a compose-file that has multiple `build` instructions, this option would grant all builds access to the ssh-agent, which is not desirable.

ppg:
Is there an alternative plan? https://github.com/docker/compose/pull/7046 mentions https://github.com/docker/compose/pull/7046 but that is again hard-coding a new argument, so it will suffer the same issue of playing keep up, but also it's only adding `--secrets`; I didn't think you could implement the `--ssh` flag with `--secrets`?

thaJeztah:
https://github.com/docker/compose/pull/7046 might still be the way forward, but might need to be proposed in https://github.com/compose-spec/compose-spec

hannahkamundson:
Would love to see this merge request put in!

tommie:
I filed https://github.com/compose-spec/compose-spec/issues/81 to continue the compose-spec discussion there.

philomory:
I'd _very_ much prefer to see this in the `docker-compose.yaml` rather than adding more and more environment variables. In fact, I'd really prefer if `COMPOSE_DOCKER_CLI_BUILD` was moved to the `docker-compose.yaml` as well (but that probably belongs in a separate issue).

moriyoshi:
> I don't think we should implement this; this would be a global option, which means that in a compose-file that has multiple build instructions, this option would grant all builds access to the ssh-agent, which is not desirable.

Passing a different key to each build sounds pretty much unusual to me. As for the compatibility, I don't think the build will be broken if `SSH_AUTH_SOCK` is present because its Dockerfile should have explicitly seeded the required key somewhere in the container in the first place.

And in case the behavior is undesirable, one can always choose not to use the option at all.

eliliam:
Lets get this merged in guys, any devs know why the build it failed?

goetas:
I do not think that this is the best solution. `COMPOSE_DOCKER_CLI_BUILD_EXTRA_ARGS` would be a global option, but not all the services being built might need those extra args, or might need them different. https://github.com/docker/compose/issues/6358 (in particular https://github.com/docker/compose/issues/6358#issuecomment-443661538) proposes a much safer solution (still a bit hacky but IMO good enough)

Billy-:
> I don't think we should implement this; this would be a global option, which means that in a compose-file that has multiple `build` instructions, this option would grant all builds access to the ssh-agent, which is not desirable.

@thaJeztah how can we get a definitive decision? If we definitely don't want this, should we close this PR to avoid confusion? And people spending time on it?

> I do not think that this is the best solution. `COMPOSE_DOCKER_CLI_BUILD_EXTRA_ARGS` would be a global option, but not all the services being built might need those extra args, or might need them different. #6358 (in particular [#6358 (comment)](https://github.com/docker/compose/issues/6358#issuecomment-443661538)) proposes a much safer solution (still a bit hacky but IMO good enough)

I agree, this is a good suggestion. Looks like maybe we should either discuss further on that issue or file a new issue (as that issue is specifically about secrets, and this PR is about passing any flags you like).

Seros:
When can we expect to have this feature done or is there something we can support to speed up the process?

glours:

PR #7340, opened 2020-04-02
Change Service.get_container not found error to honor custom container_name.

### LINKED ISSUE #7337: Error message on exec-ing a container that doesn't exist should honor container_name
When a user `up`'s a docker-compose project with a service that has a `container_name` defined, then for some reason, that container dies, then the user attempts to `exec` into the service, they receive an error message with the default container name, rather than the name they have defined.

There have been some similar issues for `run` commands, but I did not see any discussion about the error message for `exec` commands.

When a user `exec`'s a service where the container has died or been killed, the error message should use the `container_name` defined name.

It might also be possible to give a more generic message referencing the fact the container for the requested service has died, without providing the misleading default container name.

**Additional context**
Simple docker-compose File:

```yaml
version: "2.0"

services:
  redis:
    image: redis:5.0.6
    container_name: "my_project_redis
```

When I did:

```
$ docker-compose up --detach
Creating my_project_redis ... done
$ docker-compose exec redis /bin/bash
root@e251d13265ee:/data# echo "ok"
ok
root@e251d13265ee:/data# exit
exit
$ docker kill my_project_redis
my_project_redis
$ docker-compose exec redis /bin/bash
ERROR: No container found for redis_1
```

What I expected:

```
$ docker-compose exec redis /bin/bash
ERROR: No container found for my_project_redis
```

There was never a container for `redis_1`, it always had a different name. I think the behavior is correct, but the error message is misleading.

RussellRollins:
Resolves #7337

If a service has a `container_name` set, and a user attempts to `exec` into it, but it does not exist, it will throw an error that references the default name instead of the custom name.

This PR changes the error logic to check for a custom name before raising the error.

### DIFF
--- a/compose/service.py
+++ b/compose/service.py
@@ -238,7 +238,11 @@ class Service(object):
         for container in self.containers(labels=['{0}={1}'.format(LABEL_CONTAINER_NUMBER, number)]):
             return container
 
-        raise ValueError("No container found for %s_%s" % (self.name, number))
+        container_name = "%s_%s" % (self.name, number)
+        if self.custom_container_name:
+            container_name = self.custom_container_name
+
+        raise ValueError("No container found for %s" % (container_name))
 
     def start(self, **options):
         containers = self.containers(stopped=True)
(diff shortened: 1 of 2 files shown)


glours:

PR #7328, opened 2020-03-28
Bug fix 7242 docker-compose with buildkit does not insert container labels

luHub:
Resolves #7242

This issue solves bug reported by issue [7242](https://github.com/docker/compose/issues/7242)
where labels are not present when using _CLIBuilder. 

This merge request contains:
- Fix
- Integration test that checks for labels when the _CLIBuilder is used

Issue #7242, opened 2020-02-19
docker-compose with buildkit does not insert container labels


sanderegg:
## Description of the issue
using the example from [https://docs.docker.com/compose/gettingstarted/](https://docs.docker.com/compose/gettingstarted/) with a modified compose file using in particular ```version: '3.7'``` and adding some labels in the ```build``` such as my compose file looks like this:
```
version: '3.7'
services:
  web:
    build: 
      context: ./
      labels:
        some_label: "I am the label"
    ports:
      - "5000:5000"    
  redis:
    image: "redis:alpine"
```
Basically in normal mode everything works well and the labels are added to the docker image.
In buildkit mode (with both DOCKER_BUILDKIT=1 and COMPOSE_DOCKER_CLI_BUILD=1) the labels are nowhere.

## Context information (for bug reports)

**Output of `docker-compose version`**
```
docker-compose version 1.25.3, build d4d1b42b
```

**Output of `docker version`**
```
Client: Docker Engine - Community
 Version:           19.03.5
 API version:       1.40
 Go version:        go1.12.12
 Git commit:        633a0ea838
 Built:             Wed Nov 13 07:29:52 2019
 OS/Arch:           linux/amd64
 Experimental:      false

Server: Docker Engine - Community
 Engine:
  Version:          19.03.5
  API version:      1.40 (minimum version 1.12)
  Go version:       go1.12.12
  Git commit:       633a0ea838
  Built:            Wed Nov 13 07:28:22 2019
  OS/Arch:          linux/amd64
  Experimental:     true
 containerd:
  Version:          1.2.10
  GitCommit:        b34a5c8af56e510852c35414db4c1f4fa6172339
 runc:
  Version:          1.0.0-rc8+dev
  GitCommit:        3e425f80a8c931f88e6d94a8c831b9d5aa481657
 docker-init:
  Version:          0.18.0
  GitCommit:        fec3683
```

**Output of `docker-compose config`**
(Make sure to add the relevant `-f` and other flags)
```
services:
  redis:
    image: redis:alpine
  web:
    build:
      context: /home/sanderegg/dev/github/cookiecutter-osparc-service/test
      labels:
        some_label: I am a label
    ports:
    - published: 5000
      target: 5000
version: '3.7'
```

## Steps to reproduce the issue

1. set-up he example as in [https://docs.docker.com/compose/gettingstarted/](https://docs.docker.com/compose/gettingstarted/)
2. modify the docker-compose file as above
[...]

PR #7344, opened 2020-04-06
Add Integration Test for Build CLI with buildargs, target and cache from parameters

kanyaparamita:
While we're working on resolving issue #7242 through PR #7328 we found out that it might be beneficial to add integration tests for other parameters used in `_CLIBuilder` as well. Here we choose `args`, `target` and `cache_from` parameters.

Signed-off-by: Kanya Paramita <kanyaparamita@yahoo.com>

Resolves #

### DIFF
--- a/tests/integration/service_test.py
+++ b/tests/integration/service_test.py
@@ -995,12 +995,76 @@ class ServiceTest(DockerClientTestCase):
         service = self.create_service('web',
                                       build={
                                           'context': base_dir,
-                                          'labels': {'com.docker.compose.test': 'true'}},
-                                      )
+                                          'labels': {'com.docker.compose.test': 'true'}})
+
         service.build(cli=True)
         self.addCleanup(self.client.remove_image, service.image_name)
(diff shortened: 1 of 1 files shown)


tests/integration/service_test.py
```diff
@@ -995,13 +995,77 @@ def test_build_cli_with_build_labels(self):
...
+    def test_build_cli_with_cachefrom(self):
+        base_dir = tempfile.mkdtemp()
+        self.addCleanup(shutil.rmtree, base_dir)
+
+        with open(os.path.join(base_dir, 'Dockerfile'), 'w') as f:
+            f.write("FROM busybox\n")
+
+        service = self.create_service('buildcachefrom',
+                                      build={'context': base_dir,
+                                             'cache_from': ['build1']})
```
n1ru4l, line 1061: Is there also a cache_to option?

n1ru4l: And does this also allow using a registry as the cache?

glours:

PR #7389, opened 2020-04-21
General bumps to prepare 1.26.0

ulyssessouza:


ulyssessouza:
@thaJeztah That's true. But I'm not sure if this would work because before 1.25.x we would not have the `latest` tag and the code of the `.ps1` is from 2015.

Anyways, I think that since we have it now, that should be used in the `run.sh` too and stop the bumping in the file.

thaJeztah:
Oh; was thinking possibly the reverse (have both the windows and linux variant contain a version), but both using "latest" could work if we only publish the stable release as latest

ulyssessouza:
@thaJeztah But the idea is that latest points to the latest GA.

PR #7390, opened 2020-04-21
Remove unused files

ulyssessouza:


.fossa.yml
```diff
@@ -1,14 +0,0 @@
...
-cli:
-  server: https://app.fossa.io
-  fetcher: custom
-  project: git@github.com:docker/compose
-analyze:
-  modules:
-  - name: .
-    type: pip
-    target: .
-    path: .
```
ulyssessouza, line 14: @chris-crone  Can you confirm that we are not using fossa anymore?

Dockerfile.s390x
```diff
@@ -1,15 +0,0 @@
...
-    python \
-    py-pip \
-  && pip install --no-cache-dir docker-compose==$COMPOSE_VERSION \
-  && rm -rf /var/cache/apk/*
-
-WORKDIR /data
-VOLUME /data
-
-
-ENTRYPOINT ["docker-compose"]
```
ulyssessouza, line 15: This file is not used by the project...

chris-crone: It would be better to adapt the other Dockerfile if we want s390x support than use this so I'm happy to remove

ulyssessouza:
> It would be better to adapt the other Dockerfile if we want s390x support than use this so I'm happy to remove

@chris-crone My problem with this file (Dockerfile.s390x ) is that it's completely useless in the code base.
Never referenced, never built, nothing... Looks like it's historically forgotten here... And if you see the content, it's just a `pip install docker-compose` with a specific `FROM`. So if someone is using this, it should be kept by the one using, not us.

PR #7395, opened 2020-04-22
parallel_pull is default behavior in new versions

pacoxu:
https://github.com/docker/compose/blob/13bacba2b9aecdf1f3d9a4aa9e01fbc1f9e293ce/compose/cli/main.py#L736
`            --parallel              Deprecated, pull multiple images in parallel (enabled by default).
`

parallel_pull should be enabled by default for performance.

### DIFF
--- a/compose/project.py
+++ b/compose/project.py
@@ -622,7 +622,7 @@ class Project(object):
 
         return plans
 
-    def pull(self, service_names=None, ignore_pull_failures=False, parallel_pull=False, silent=False,
+    def pull(self, service_names=None, ignore_pull_failures=False, parallel_pull=True, silent=False,
              include_deps=False):
         services = self.get_services(service_names, include_deps)
 


pacoxu:
@chris-crone @thaJeztah @ulyssessouza  
Could you review this pr ?

aiordache:
Hi @pacoxu! I would consider the removal of the parameter for a cleaner code. I marked this as a fix to get in for v1.28.0 release as it would give us some time to verify its impact.

thaJeztah:
> I would consider the removal of the parameter for a cleaner code.

There's a chance that removing the flag will break user scripts if they're using this flag. Not sure what the deprecation policy usually is on Docker Compose?

If possible, we could consider _hiding_ the flag, and printing a deprecation warning (or error-out) if it's used

pacoxu:
> Hi @pacoxu! I would consider the removal of the parameter for a cleaner code. I marked this as a fix to get in for v1.28.0 release as it would give us some time to verify its impact.

thanks for your reply.

pacoxu:
> > I would consider the removal of the parameter for a cleaner code.
> 
> 
> 
> There's a chance that removing the flag will break user scripts if they're using this flag. Not sure what the deprecation policy usually is on Docker Compose?
> 
> 
> 
> If possible, we could consider _hiding_ the flag, and printing a deprecation warning (or error-out) if it's used

as it has a different behavior with the cli， the warning is neccesery. Thanks for your comments.

PR #7413, opened 2020-04-29
GH-7397: Handle symlinked Dockerfiles

### LINKED ISSUE #7397: docker-compose up fails with symlinked Dockerfile
## Description of the issue

I have several microservices which share an identical `Dockerfile`. To simplify maintenance and reduce duplication I would like to replace all the Dockerfiles with a symlink to a single "common" `Dockerfile`.

Before symlinking, my directory structure looks like this:

```
- docker-compose.yaml
- microservice_a/
  - Dockerfile
- microservice_b/
  - Dockerfile
```

after, it looks like this:

```
- docker-compose.yaml
- common/
  - Dockerfile
- microservice_a/
  - Dockerfile  ->  ../common/Dockerfile
- microservice_b/
  - Dockerfile  ->  ../common/Dockerfile
```

Without symlinking, `docker-compose up` works as expected.

Once I switch to the symlink model, this no longer works, and I get errors related to `Dockerfile: no such file or directory`. See below for more detail:

```
```

```
Docker version 19.03.8, build afacb8b
```

(ommited docker-compose config because it contains confidential information like private keys which I'm configuring using env vars)

1. Create two services with idential dockerfiles and reference them both in a compose file.
2. Move their dockerfile to a subdirectory 'common' and symlink it in the services' Dockerfiles.

### Observed result

`docker-compose` can no longer find the Dockerfile, even though it's still in "the same place" because of the symlink.

### Expected result

`docker-compose` continues to function, and can read the Dockerfile through the new symlink.

```
[+] Building 0.0s (2/2) FINISHED
 => [internal] load .dockerignore                                                                                                  0.0s
 => => transferring context: 34B                                                                                                   0.0s
 => [internal] load build definition from Dockerfile                                                                               0.0s
 => => transferring dockerfile: 70B                                                                                                0.0s
failed to solve with frontend dockerfile.v0: failed to read dockerfile: open /var/lib/docker/tmp/buildkit-mount384971895/common/Dockerfile: no such file or directory
  File "docker-compose", line 6, in <module>
  File "compose/cli/main.py", line 72, in main
  File "compose/cli/main.py", line 128, in perform_command
  File "compose/cli/main.py", line 1077, in up
  File "compose/cli/main.py", line 1073, in up
  File "compose/project.py", line 548, in up
  File "compose/service.py", line 351, in ensure_image_exists
  File "compose/service.py", line 1110, in build
  File "compose/progress_stream.py", line 25, in stream_output
  File "compose/utils.py", line 61, in split_buffer
  File "compose/utils.py", line 37, in stream_as_text
  File "compose/service.py", line 1818, in build
FileNotFoundError: [Errno 2] No such file or directory: '/var/folders/7_/wk5b0_312sq4_9x59jhq0q_r0000gn/T/tmp0jhuoqwn'
[20348] Failed to execute script docker-compose
```

## Additional information

MacOS Catalina 10.15.4

Code0x58:
Resolves #7397

### DIFF
--- a/compose/service.py
+++ b/compose/service.py
@@ -1082,6 +1082,22 @@ class Service(object):
                 'Impossible to perform platform-targeted builds for API version < 1.35'
             )
 
+        dockerfile_opt = build_opts.get('dockerfile', None)
+        # logic based on https://github.com/docker/docker-py/blob/4.2.0/docker/api/build.py#L144
+        if path.startswith(('http://', 'https://',
+                            'git://', 'github.com/', 'git@')):
+            fileobj = None
+            dockerfile = dockerfile_opt
+        elif gzip:
+            if dockerfile_opt and os.path.islink(os.path.join(path, dockerfile_opt)):
+                # let the daemon do the complaining
+                log.warning('Dockerfile is a symlink and --compress is used')
+            fileobj = None
+            dockerfile = dockerfile_opt
+        else:
+            fileobj = open(os.path.join(path, dockerfile_opt or 'Dockerfile'), 'rb')
+            dockerfile = None
+
         builder = self.client if not cli else _CLIBuilder(progress)
         build_output = builder.build(
             path=path,
@@ -1090,7 +1106,8 @@ class Service(object):
             forcerm=force_rm,
             pull=pull,
             nocache=no_cache,
-            dockerfile=build_opts.get('dockerfile', None),
+            dockerfile=dockerfile,
+            fileobj=fileobj,
             cache_from=self.get_cache_from(build_opts),
             labels=build_opts.get('labels', None),
             buildargs=build_args,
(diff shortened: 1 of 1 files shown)


compose/service.py
```diff
@@ -1090,7 +1103,7 @@ def build(self, no_cache=False, pull=False, force_rm=False, memory=None, build_a
             forcerm=force_rm,
             pull=pull,
             nocache=no_cache,
-            dockerfile=build_opts.get('dockerfile', None),
```
Code0x58, line 1093: When the path is a URL or gzip is true, `APIClient` passes dockerfile through to the server, so this should probably still be provided in those cases

thaJeztah:
Thanks! Looking at my steps trying to reproduce the problem (see https://github.com/docker/compose/issues/7397#issuecomment-621513714), I'm actually wondering if this should be fixed in docker-compose; I'm thinking this should be fixed in the docker cli (or in buildkit), and docker-compose should pass the `dockerfile` path as-is (leaving the handling to the CLI)

Code0x58:
Nice drill-down! I like the idea of it being cleaned up and handled in the CLI, it sounds like it could touch a lot of complements - I'm certainly interested to see how it's managed.

I'm wondering if this PR would find it's way in as a workaround at the docker-compose level, until changes are made to the CLI (and python library?), then here would need updating again.

glours:

PR #7419, opened 2020-05-01
Speed up fish completions for subcommands

### LINKED ISSUE #7418: Loading fish shell completions is slow due to running `docker-compose --help`
## Description of the issue
Running `docker-compose --help` on my system (a 2018 MacBook Pro) routinely takes several hundred milliseconds:

```
$ hyperfine --warmup 3 'docker-compose --help'
Benchmark #1: docker-compose --help
  Time (mean ± σ):     351.6 ms ±  16.0 ms    [User: 295.5 ms, System: 44.3 ms]
  Range (min … max):   332.8 ms … 384.0 ms    10 runs
```

`docker-compose --help` is run in `contrib/completion/fish/docker-compose.fish` in order to dynamically generate completions which means that if fish is configured to automatically `source` that file, it will greatly increase start-up time.

```
docker-py version: 4.1.0
CPython version: 3.7.5
OpenSSL version: OpenSSL 1.1.1d  10 Sep 2019
```

```
 Version:           19.03.8
 API version:       1.40
 Go version:        go1.12.17
 Git commit:        afacb8b
 Built:             Wed Mar 11 01:21:11 2020
 Experimental:      false

 Engine:
  Version:          19.03.8
  Go version:       go1.12.17
  Git commit:       afacb8b
  Built:            Wed Mar 11 01:29:16 2020
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          v1.2.13
 runc:
  Version:          1.0.0-rc10
 docker-init:
  Version:          0.18.0
  GitCommit:        fec3683
```

1. Run `docker-compose --help`
2. Go brew yourself some coffee while you wait for it to finish

Ideally it shouldn't take very long to print out the help message but I'm not familiar enough with the codebase to fix it. In the meantime, I'm going to open a PR to remove the dependency on running `docker-compose --help` from the fish completions.

edit: That PR is here https://github.com/docker/compose/pull/7419

Update: I discovered that fish has a variable `fish_complete_path` which determines which directories it will search for completions in. By adding the vendor completions directory to that variable, I can make fish defer loading the completion until it's actually needed (i.e., only after typing `docker-compose <TAB>`). This speeds up launching new fish processes but there's still the issue that when I _do_ want to load the completions, it hangs for a good while before displaying them (not to mention the incorrect completions bug mentioned in the PR description). So I still think the PR is helpful.

cstyles:
The current method of generating completions for the fish shell relies
on running `docker-compose --help`. On my system (a 2018 MacBook Pro)
this routinely takes several hundred milliseconds:

```
$ hyperfine --warmup 3 'docker-compose --help'
Benchmark #1: docker-compose --help
  Time (mean ± σ):     351.6 ms ±  16.0 ms    [User: 295.5 ms, System: 44.3 ms]
  Range (min … max):   332.8 ms … 384.0 ms    10 runs
```

If I configure fish to automatically load these completions, it makes
launching a new fish process very sluggish.

In addition, the regular expression used to parse the output of
`docker-compose --help` incorrectly generates completions for `name` and
`in` due to the following lines:

https://github.com/docker/compose/blob/9c5351cf27d10a7b8078496fb21a37929b6687cb/compose/cli/main.py#L208
https://github.com/docker/compose/blob/9c5351cf27d10a7b8078496fb21a37929b6687cb/compose/cli/main.py#L212

This commit fixes these issues by explicitly listing subcommands instead
of generating them dynamically. This means that adding/removing
subcommands won't automatically update the completions but according to
the git blame for `compose/cli/main.py`, no one has touched the list of
subcommands in the help message in 3+ years so I hope that won't be an
issue.

Resolves #7418 

Update: I discovered that fish has a variable `fish_complete_path` which determines which directories it will search for completions in. By adding the vendor completions directory to that variable, I can make fish defer loading the completion until it's actually needed (i.e., only after typing `docker-compose <TAB>`). This speeds up launching new fish processes but there's still the issue that when I _do_ want to load the completions, it hangs for a good while before displaying them (not to mention the incorrect completions bug). So I still think this PR is helpful.

### DIFF
--- a/contrib/completion/fish/docker-compose.fish
+++ b/contrib/completion/fish/docker-compose.fish
@@ -3,12 +3,31 @@
 
 complete -e -c docker-compose
 
-for line in (docker-compose --help | \
-             string match -r '^\s+\w+\s+[^\n]+' | \
-             string trim)
-  set -l doc (string split -m 1 ' ' -- $line)
-  complete -c docker-compose -n '__fish_use_subcommand' -xa $doc[1] --description $doc[2]
-end
+complete -c docker-compose -n '__fish_use_subcommand' -xa 'build' --description 'Build or rebuild services'
+complete -c docker-compose -n '__fish_use_subcommand' -xa 'config' --description 'Validate and view the Compose file'
+complete -c docker-compose -n '__fish_use_subcommand' -xa 'create' --description 'Create services'
+complete -c docker-compose -n '__fish_use_subcommand' -xa 'down' --description 'Stop and remove containers, networks, images, and volumes'
+complete -c docker-compose -n '__fish_use_subcommand' -xa 'events' --description 'Receive real time events from containers'
+complete -c docker-compose -n '__fish_use_subcommand' -xa 'exec' --description 'Execute a command in a running container'
+complete -c docker-compose -n '__fish_use_subcommand' -xa 'help' --description 'Get help on a command'
+complete -c docker-compose -n '__fish_use_subcommand' -xa 'images' --description 'List images'
+complete -c docker-compose -n '__fish_use_subcommand' -xa 'kill' --description 'Kill containers'
+complete -c docker-compose -n '__fish_use_subcommand' -xa 'logs' --description 'View output from containers'
+complete -c docker-compose -n '__fish_use_subcommand' -xa 'pause' --description 'Pause services'
+complete -c docker-compose -n '__fish_use_subcommand' -xa 'port' --description 'Print the public port for a port binding'
+complete -c docker-compose -n '__fish_use_subcommand' -xa 'ps' --description 'List containers'
+complete -c docker-compose -n '__fish_use_subcommand' -xa 'pull' --description 'Pull service images'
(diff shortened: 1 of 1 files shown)


ulyssessouza:
Hello @cstyles ,

Well, for the wrong completions, I think that the path is to fix the regexp. But changing the approach to a fixed list I don't agree, since it will increase the time in maintainability and it will be for sure something to forget.

cstyles:
I agree that, at least in theory, it's not the cleanest solution. But in practice, the list of subcommands hasn't changed in 3 years so I don't think maintenance will take up that much time (especially since fish isn't exactly the most popular shell). On top of that, the completions for bash and zsh both hardcode the subcommands as well (plus all their options) so this approach is at least consistent.

IMO the best solution is to improve the performance of `docker-compose --help` but that would probably require a lot of effort. I figured just hardcoding the completions gets most of the benefits with little cost. Just my two cents.

glours:

Issue #7408, opened 2020-04-29
docker-compose logs no output if using log-driver


audunsolemdal:
## Description of the issue

docker-compose logs does not show output when using external log-driver. I would like to access the logs from both places.

## Context information (for bug reports)

**Output of `docker-compose version`**
```
docker-compose version 1.25.1, build a82fef07
docker-py version: 4.1.0
CPython version: 3.7.4
OpenSSL version: OpenSSL 1.1.0l  10 Sep 2019
```

**Output of `docker version`**
```
Client: Docker Engine - Community
 Version:           19.03.8
 API version:       1.39 (downgraded from 1.40)
 Go version:        go1.12.17
 Git commit:        afacb8b
 Built:             Wed Mar 11 01:27:04 2020
 OS/Arch:           linux/amd64
 Experimental:      false
Server: Docker Engine - Community
 Engine:
  Version:          18.09.1
  API version:      1.39 (minimum version 1.12)
  Go version:       go1.10.6
  Git commit:       4c52b90
  Built:            Wed Jan  9 19:06:30 2019
  OS/Arch:          linux/amd64
  Experimental:     false
```

**Output of `docker-compose config`**
(Make sure to add the relevant `-f` and other flags)
```
services:
  app:
    entrypoint: ping host.docker.internal
    image: alpine:latest
    logging:
      driver: loki
      options:
        loki-url: http://host.docker.internal:3100/loki/api/v1/push
  grafana:
    environment:
      GF_EXPLORE_ENABLED: "true"
    image: grafana/grafana:master
    ports:
    - published: 3000
      target: 3000
  loki:
    image: grafana/loki:master
    ports:
    - published: 3100
      target: 3100
version: '3.7'
```

## Steps to reproduce the issue

1. docker plugin install grafana/loki-docker-driver:latest --alias loki --grant-all-permissions
2. Set daemon.json config as described here https://github.com/grafana/loki/blob/v1.4.1/docs/clients/docker-driver/configuration.md#change-the-default-logging-driver
3. docker-compose up -d with the config above
4. docker-compose logs

### Observed result

From this issue https://github.com/grafana/loki/issues/1368
and this comment https://github.com/grafana/loki/issues/1368#issuecomment-606872659

`docker logs <containername>` works fine. `docker-compose logs` does not work.

### Expected result

I would expect `docker-compose logs` to be able to show output in the same fashion as `docker logs`

### Stacktrace / full error message

```
[...]

PR #7425, opened 2020-05-04
Add 'local' to list of logging drivers that support `docker-compose logs`

### LINKED ISSUE #7420: `docker-compose logs` doesn't work for local log driver

```
docker-compose version 1.25.5, build 8a1c60f6
docker-py version: 4.1.0
CPython version: 3.7.5

```

```
Client:
 Version:           18.09.9-ce
 API version:       1.39
 Go version:        go1.10.3
 Git commit:        039a7df
 Built:             Fri Nov  1 19:26:49 2019
 Experimental:      false

Server:
 Engine:
  Version:          18.09.9-ce
  Go version:       go1.10.3
  Git commit:       039a7df
  Built:            Fri Nov  1 19:28:24 2019
  OS/Arch:          linux/amd64
  Experimental:     false

```

```
services:
  rabbit:
    image: rabbitmq
    logging:
      driver: local
version: '3.0'
```

1. Create the above docker-compose.yaml file
2. Call `docker-compose up -d`
3. Call `docker-compose logs`

### Observed result
$ docker-compose logs 
Attaching to mike_rabbit_1
rabbit_1  | WARNING: no logs are available with the 'local' log driver

### Expected result
Get logs from rabbitmq

```
N/A
```

## Additional information
If I do `docker logs <container-id>` this works.

This seems like it comes from `compose/container.py`:
```
    @property
    def has_api_logs(self):
        log_type = self.log_driver
        return not log_type or log_type in ('json-file', 'journald')
```
That doesn't include 'local'.

I'll try to fix it and if it's as easy as adding 'local' to the list - will send a PR tomorrow.

svarogg:
Resolves #7420

### DIFF
--- a/compose/container.py
+++ b/compose/container.py
@@ -208,8 +208,8 @@ class Container(object):
         return status_string
 
     def attach_log_stream(self):
-        """A log stream can only be attached if the container uses a json-file
-        log driver.
+        """A log stream can only be attached if the container uses a
+        json-file, journald or local log driver.
         """
         if self.has_api_logs:
             self.log_stream = self.attach(stdout=True, stderr=True, stream=True)
--- a/compose/container.py
+++ b/compose/container.py
@@ -193,7 +193,7 @@ class Container(object):
     @property
     def has_api_logs(self):
         log_type = self.log_driver
-        return not log_type or log_type in ('json-file', 'journald')
+        return not log_type or log_type in ('json-file', 'journald', 'local')
 
     @property
     def human_readable_health_status(self):


compose/container.py
```diff
@@ -193,7 +193,7 @@ def log_driver(self):
     @property
     def has_api_logs(self):
         log_type = self.log_driver
-        return not log_type or log_type in ('json-file', 'journald')
+        return not log_type or log_type in ('json-file', 'journald', 'local')
```
thaJeztah, line 196: Slightly related; looks like the description here also needs updating; https://github.com/docker/compose/blob/4bc1cbc32accce59acddece7314eab6929eefe42/compose/container.py#L211

svarogg: Thanks for the tip. Updated the comment.

solomson:
Will this also fix #7408 ?

avkonst:
Hey, can this PR be merged and new compose version released with the fix?

avkonst:
Thanks for the merge! Is there any information when it is going to be released in a new version?

thaJeztah:
Not sure when the next release will be (@rumpl may know if something is planned already).


Just a heads up; the next Docker (20.xx) release will include a feature called "dual logging" (docs: https://docs.docker.com/config/containers/logging/dual-logging/) that was previously only in the Enterprise Edition of the Docker Engine. That feature has been upstreamed (through https://github.com/moby/moby/pull/40543). With that feature, the "local" logging driver can be always enabled, which means that the `logs` endpoint will always be able to read back logs from a buffer, irregardless of the logging driver that is used (which could be a logging plugin that sends logs to a remote logging aggregator)

PR #7430, opened 2020-05-05
Fix push command skips service

### LINKED ISSUE #7429: Docker-compose push skips second service if first has same image but no build section
## Description of the issue
When you have multiple services with same image, but only one (not first) has `build` field, `docker-compose push` ignores it.

```
docker-py version: 4.1.0
CPython version: 3.7.5
OpenSSL version: OpenSSL 1.1.1d  10 Sep 2019
```

```
 Version:           19.03.8
 API version:       1.40
 Go version:        go1.12.17
 Git commit:        afacb8b
 Built:             Wed Mar 11 01:21:11 2020
 Experimental:      false

 Engine:
  Version:          19.03.8
  Go version:       go1.12.17
  Git commit:       afacb8b
  Built:            Wed Mar 11 01:29:16 2020
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          v1.2.13
 runc:
  Version:          1.0.0-rc10
 docker-init:
  Version:          0.18.0
  GitCommit:        fec3683
```

```
services:
  first:
    image: my_app
  second:
    build:
      context: /Users/musikov/docker_push_issue
    image: my_app
version: '3.7'
```

1. Create dummy dir `mkdir ~/docker_push_issue`
2. Go to it `cd ~/docker_push_issue`
3. Create dummy Dockerfile `echo 'from python' > Dockerfile`
4. Create docker-compose.yml file as in issue:
```
echo "services:
  first:
    image: my_app
  second:
    build: .
    image: my_app
version: '3.7'
" > docker-compose.yml
```
5. Build project `docker-compose build`
6. Push project `docker-compose push`

### Observed result
Command runs, but nothing happened.
### Expected result
Image is pushed. (for current example we must receive `no permission` error)
```
# docker-compose -verbose push
```

## Additional information

macOS 10.15.4 (19E287), docker desktop

musikov:
Resolves #7429 

The first service is skipped because it has no build instruction - it's ok.
But his image still was added to `unique_images` set, so the second service (with same image, but with build instruction) was skipped.
I've fixed this by adding check is service is pushable. Only after this i add it to unique_images

### DIFF
--- a/compose/project.py
+++ b/compose/project.py
@@ -695,9 +695,7 @@ class Project(object):
             repo, tag, sep = parse_repository_tag(service.image_name)
             service_image_name = sep.join((repo, tag)) if tag else sep.join((repo, 'latest'))
 
-            # Add service to unique if it's pushable
-            if (service_image_name not in unique_images and
-                    service.can_be_pushed()):
+            if service.can_be_pushed() and service_image_name not in unique_images:
                 service.push(ignore_push_failures)
                 unique_images.add(service_image_name)
 
--- a/compose/project.py
+++ b/compose/project.py
@@ -695,7 +695,9 @@ class Project(object):
             repo, tag, sep = parse_repository_tag(service.image_name)
             service_image_name = sep.join((repo, tag)) if tag else sep.join((repo, 'latest'))
 
-            if service_image_name not in unique_images:
+            # Add service to unique if it's pushable
+            if (service_image_name not in unique_images and
+                    service.can_be_pushed()):
                 service.push(ignore_push_failures)
                 unique_images.add(service_image_name)
 
--- a/compose/service.py
+++ b/compose/service.py
@@ -1138,6 +1138,12 @@ class Service(object):
     def can_be_built(self):
         return 'build' in self.options
 
+    def can_be_pulled(self):
+        return 'image' in self.options
+
+    def can_be_pushed(self):
+        return 'build' in self.options and 'image' in self.options
+
     def labels(self, one_off=False, legacy=False):
         proj_name = self.project if not legacy else re.sub(r'[_-]', '', self.project)
         return [
@@ -1227,7 +1233,7 @@ class Service(object):
                 log.error(six.text_type(e))
 
     def pull(self, ignore_pull_failures=False, silent=False, stream=False):
-        if 'image' not in self.options:
+        if not self.can_be_pulled():
             return
 
         repo, tag, separator = parse_repository_tag(self.options['image'])
@@ -1250,7 +1256,7 @@ class Service(object):
         return progress_stream.get_digest_from_pull(event_stream)
 
     def push(self, ignore_push_failures=False):
-        if 'image' not in self.options or 'build' not in self.options:
+        if not self.can_be_pushed():
             return
 
         repo, tag, separator = parse_repository_tag(self.options['image'])
--- a/tests/unit/project_test.py
+++ b/tests/unit/project_test.py
@@ -12,6 +12,8 @@ from docker.errors import NotFound
 
 from .. import mock
 from .. import unittest
+from ..helpers import BUSYBOX_IMAGE_NAME
+from ..helpers import BUSYBOX_DEFAULT_TAG
 from ..helpers import BUSYBOX_IMAGE_WITH_TAG
 from compose.config import ConfigurationError
 from compose.config.config import Config
@@ -845,6 +847,23 @@ class ProjectTest(unittest.TestCase):
         with mock.patch('compose.service.Service.push') as fake_push:
             project.push()
             assert fake_push.call_count == 2
+
+    def test_push_duplicate_image(self):
+        svc_no_build = Service(
+            'busy1', image=BUSYBOX_IMAGE_WITH_TAG,
+            client=self.mock_client)
+        svc_with_build = Service(
+            'busy2', image=BUSYBOX_IMAGE_WITH_TAG,
+            build='.',
+            client=self.mock_client)
+        project = Project(
+            'composetest', [svc_no_build, svc_with_build], self.mock_client
+        )
+        project.push()
+        self.mock_client.push.assert_called_once_with(
+            BUSYBOX_IMAGE_NAME,
+            tag=BUSYBOX_DEFAULT_TAG,
+            stream=True)
 
     def test_get_secrets_no_secret_def(self):
         service = 'foo'
--- a/tests/unit/project_test.py
+++ b/tests/unit/project_test.py
@@ -12,8 +12,8 @@ from docker.errors import NotFound
 
 from .. import mock
 from .. import unittest
-from ..helpers import BUSYBOX_IMAGE_NAME
 from ..helpers import BUSYBOX_DEFAULT_TAG
+from ..helpers import BUSYBOX_IMAGE_NAME
 from ..helpers import BUSYBOX_IMAGE_WITH_TAG
 from compose.config import ConfigurationError
 from compose.config.config import Config


compose/project.py
```diff
@@ -695,7 +695,9 @@ def push(self, service_names=None, ignore_push_failures=False):
             repo, tag, sep = parse_repository_tag(service.image_name)
             service_image_name = sep.join((repo, tag)) if tag else sep.join((repo, 'latest'))
 
-            if service_image_name not in unique_images:
+            # Add service to unique if it's pushable
+            if (service_image_name not in unique_images and
+                    service.can_be_pushed()):
```
joseangel-sc, line 700: maybe remove the `(` I don't fell like it is needed here

compose/service.py
```diff
@@ -1138,6 +1138,12 @@ def get_cache_from(self, build_opts):
     def can_be_built(self):
         return 'build' in self.options
 
+    def can_be_pulled(self):
+        return 'image' in self.options
+
+    def can_be_pushed(self):
+        return 'build' in self.options and 'image' in self.options
```
thaJeztah, line 1145: Not sure if there's a validation for this already, but `image` can also have a _digest_ in it (e.g. `busybox@sha256:a8cf7ff6367c2afa2a90acd081b484cbded349a7076e7bdf37a05279f276bc12`)

While that's valid for _pulling_ an image, it's not valid for _building_ or _pushing_ an image; I think this function should check if `parse_repository_tag ` returns `digest_separator` (`@`) as separator, and if so, return `false`.

Probably something like this.

```python
def can_be_pushed(self):
    if 'build' not in self.options or 'image' not in self.options:
        return False
    _, _, separator = parse_repository_tag(self.options.get('image'))
    return separator != '@'
```

Can you also add a test for these?

thaJeztah: On a side-note: from my limited Python skills it looks like currently, `parse_repository_tag` does not support a reference to have _both_ a `tag` and a `digest`, for example:

```
busybox:latest@sha256:a8cf7ff6367c2afa2a90acd081b484cbded349a7076e7bdf37a05279f276bc12 
```

admittedly, that syntax is mostly is to make the reference "nicer" to read, because _digest_ is leading in that case, but it probably should be looked at 
https://github.com/docker/compose/blob/440c94ea7a7e62b3de50722120ca34c4e818205a/compose/service.py#L1443-L1445

/cc @ulyssessouza

compose/project.py
```diff
@@ -695,7 +695,9 @@ def push(self, service_names=None, ignore_push_failures=False):
             repo, tag, sep = parse_repository_tag(service.image_name)
             service_image_name = sep.join((repo, tag)) if tag else sep.join((repo, 'latest'))
```
thaJeztah, line 696: Perhaps `service.image_name` should be modified to do this (note: I'm not a Python developer, so perhaps there's reasons why that wasn't done 😅);  https://github.com/docker/compose/blob/440c94ea7a7e62b3de50722120ca34c4e818205a/compose/service.py#L386-L389

Which could look  something like (again, I'm not a python developer, so possibly there's a cleaner way):

```python
def image_name(self):
    image = self.options.get('image', '{project}_{s.name}'.format(
        s=self, project=self.project.lstrip('_-')
    ))
    if image == "":
        return
    repo, tag, separator = parse_repository_tag(image)
    tag = tag or 'latest'
    return repo+separator+tag
```

compose/project.py
```diff
@@ -695,7 +695,9 @@ def push(self, service_names=None, ignore_push_failures=False):
...
 
-            if service_image_name not in unique_images:
+            # Add service to unique if it's pushable
```
thaJeztah, line 698: This comment is redundant; it's an exact "copy" of the code below (only in a sentence)

compose/project.py
```diff
@@ -695,7 +695,9 @@ def push(self, service_names=None, ignore_push_failures=False):
...
+            if (service_image_name not in unique_images and
```
thaJeztah, line 699: With my suggestion to modify `service.image_name` above, this function could look something like;

```python
def push(self, service_names=None, ignore_push_failures=False):
    unique_images = set()
    for service in self.get_services(service_names, include_deps=False):
        if service.can_be_pushed() and service.image_name not in unique_images:
            service.push(ignore_push_failures)
            unique_images.add(service.image_name)
```

tests/unit/project_test.py
```diff
@@ -846,6 +848,23 @@ def test_avoid_multiple_push(self):
             project.push()
             assert fake_push.call_count == 2
 
+    def test_push_duplicate_image(self):
+        svc_no_build = Service(
+            'busy1', image=BUSYBOX_IMAGE_WITH_TAG,
+            client=self.mock_client)
+        svc_with_build = Service(
+            'busy2', image=BUSYBOX_IMAGE_WITH_TAG,
```
thaJeztah, line 856: See my comment about digests; can you add a test-case that has a service with a digest in the image reference?

compose/service.py
```diff
@@ -1250,7 +1256,7 @@ def pull(self, ignore_pull_failures=False, silent=False, stream=False):
         return progress_stream.get_digest_from_pull(event_stream)
 
     def push(self, ignore_push_failures=False):
-        if 'image' not in self.options or 'build' not in self.options:
+        if not self.can_be_pushed():
```
thaJeztah, line 1259: Wondering; should we log something if the service _does_ have an `image:` specified, but either doesn't have `build:` or image has a _digest_ set (to give some clue why the image for a service wasn't pushed)?

compose/project.py
thaJeztah: Actually, given that `service.push()` itself checks `can_be_pushed`, we could just skip that check here, and delegate it to `service.push()`;

```python
def push(self, service_names=None, ignore_push_failures=False):
    unique_images = set()
    for service in self.get_services(service_names, include_deps=False):
        if service.image_name not in unique_images:
            service.push(ignore_push_failures)
            unique_images.add(service.image_name)
```

compose/project.py
```diff
@@ -695,7 +695,9 @@ def push(self, service_names=None, ignore_push_failures=False):
...
+                    service.can_be_pushed()):
                 service.push(ignore_push_failures)
```
thaJeztah, line 701: Unrelated: actually never given much attention to `ignore_push_failures`, but I think `ignore_push_failures` is way too broad; it seems to ignore _any_ push failure (for whatever cause push failed). I think it'd make sense to either (by default?) ignore  _missing_ images (e.g., if I didn't build a service yet and the image isn't present), or add / rename this option to `--ignore-missing`.

I don't think other failures should be ignorable (e.g. authentication failed, registry unreachable, or pushing caused any other failure)

@ulyssessouza WDYT?

compose/project.py
musikov: Why do we need to do so? Just to ensure `tag` is not empty?

compose/project.py
musikov: > Actually, given that `service.push()` itself checks `can_be_pushed`, we could just skip that check here, and delegate it to `service.push()`;
That's what caused the original issue :)
Services that were not supposed to be pushed tried to push themselves, and added their `image` to `unique_images`

compose/service.py
musikov: It would be great to have logs!
For services without `build` and|or `image` just info log, that they were skipped.

compose/service.py
musikov: > Not sure if there's a validation for this already, but `image` can also have a _digest_ in it (e.g. `busybox@sha256:a8cf7ff6367c2afa2a90acd081b484cbded349a7076e7bdf37a05279f276bc12`)
> 
> While that's valid for _pulling_ an image, it's not valid for _building_ or _pushing_ an image; I think this function should check if `parse_repository_tag ` returns `digest_separator` (`@`) as separator, and if so, return `false`.
> 
> Probably something like this.
> 
> ```python
> def can_be_pushed(self):
>     if 'build' not in self.options or 'image' not in self.options:
>         return False
>     _, _, separator = parse_repository_tag(self.options.get('image'))
>     return separator != '@'
> ```
> 
> Can you also add a test for these?

Well, if someone stated, that service has `build` phase and `image` with digest, it's a mistake in config. `push` handler shouldn't be bothered by it, the `client` will fail on push with explanation.
The better way is to add such checks in config validation IMHO

compose/project.py
thaJeztah: Opened https://github.com/docker/compose/issues/7431 to propose adding an `--ignore-missing` or `--skip-missing` flag

compose/service.py
musikov: (update) suggestion to leave can_be_pushed same, but add exception if trying to push image with digest
```
    def can_be_pushed(self):
        return 'build' in self.options and 'image' in self.options

...

    def push(self, ignore_push_failures=False):
        if not self.can_be_pushed():
            return

        repo, tag, separator = parse_repository_tag(self.options['image'])
        if separator == '@':
            raise ConfigurationError('Can\'t push image with digest')
        tag = tag or 'latest'
        log.info('Pushing %s (%s%s%s)...' % (self.name, repo, separator, tag))
        output = self.client.push(repo, tag=tag, stream=True)
        ...
```

musikov: > On a side-note: from my limited Python skills it looks like currently, `parse_repository_tag` does not support a reference to have _both_ a `tag` and a `digest`, for example:
> 
> ```
> busybox:latest@sha256:a8cf7ff6367c2afa2a90acd081b484cbded349a7076e7bdf37a05279f276bc12 
> ```
> 
> admittedly, that syntax is mostly is to make the reference "nicer" to read, because _digest_ is leading in that case, but it probably should be looked at
> https://github.com/docker/compose/blob/440c94ea7a7e62b3de50722120ca34c4e818205a/compose/service.py#L1443-L1445
> 
> /cc @ulyssessouza

What is excepted behaviour? Jsut ignore tag, so
```python
>>> parse_repository_tag('user/repo:v1@sha256:digest')
('user/repo', 'sha256:digest', '@')
```

compose/project.py
musikov: But, we can make `service.push` return None if image can't be pushed. And rely on it:
```python
def push(self, service_names=None, ignore_push_failures=False):
    unique_images = set()
    for service in self.get_services(service_names, include_deps=False):
        if service.push(ignore_push_failures) and service.image_name not in unique_images:
            unique_images.add(service.image_name)
```

compose/service.py
thaJeztah: I think pulling with just the digest is ok (digest references can only be used for pulling / running, not for pushing/building). Unless we want this function to return all things separate (which may be more "correct", but would require more changes);

This is "ok" `image@sha256:deadbeef`

But more "correct" would be 


```
repo, tag, digest = parse_repository_tag('user/repo:v1@sha256:digest')
```

(return the tag (if any) and/or digest (if any)); perhaps it should return some struct having those properties.

But (as mentioned) for now, just using `image@sha256:deadbeef` should be ok

compose/project.py
thaJeztah: There's some old behavior where `docker push image<no :tag here>` pushes _all tags_ for `image`. In this case though, omitting `:tag` means `:latest`, so I think it would make sense to return that here (so that this logic doesn't have to be repeated everywhere); open to suggestions though

compose/project.py
thaJeztah: > Services that were not supposed to be pushed tried to push themselves, and added their image to unique_images

🤦 you're right, my mistake


> But, we can make service.push return None if it can't be pushed. (of course, document it and add test). And rely on it like this:

not sure if that's clearer; I think we should take the original approach (`if service.can_be_pushed() and service.image_name not in unique_images`) I think that properly describes the intent

compose/service.py
musikov: > But (as mentioned) for now, just using `image@sha256:deadbeef` should be ok

Added #7432 pull request for it, 'cause it's separate functionality

compose/project.py
musikov: If we look through whole code, we can see, that `latest` is added only after `parse_repository_tag` call. So maybe we should add it there?
Anyways, think we need separate issue/discussion for it (despite it can simplify solving current issue)

tests/unit/project_test.py
musikov: If we add config validation for "build + digest image", we won't need this check

domibarton:
Is there anybody working at this MR, because we're running into the same issue :)

musikov:
> Is there anybody working at this MR, because we're running into the same issue :)

I've changed service order in my compose file and forgot about this issue. I'll soon return to it, review requested changes

glours:

PR #7457, opened 2020-05-15
[Compose Spec] Make 'run' behave in the same way as 'up'

### LINKED ISSUE #7453: Respect service_healthy in docker-compose run
## Description of the issue
`docker-compose run` and `docker-compose up` demonstrate different behaviour when it comes to `condition: service_healthy`. `up` correctly waits for all the dependencies to be healthy while `run` only creates dependencies without waiting for them to be healthy (effectively disregarding the `condition`). Personally, I would consider this a bug since this is a rather unexpected inconsistent behaviour.

Relevant issues with examples of this:
- [docker-compose run: healthcheck not detecting if one of the services failed to start](https://github.com/docker/compose/issues/4420)
- [run dont do condition: service_healthy](https://github.com/docker/compose/issues/4952)
- [Healthcheck dependency not respected during one-off command (run)](https://github.com/docker/compose/issues/4369)

1. `docker-compose run` a service that declares a `condition: service_healthy` dependency

### Observed result
`docker-compose` only creates a container for the dependency and doesn't wait for the health-check to pass.

### Expected result
`docker-compose` creates a container for the dependency and waits for the health-check to pass - demonstrating the same behaviour as `docker-compose up`.

EricHripko:
# Overview
This PR updates `docker-compose run` to wait for `condition: service_healthy` dependencies to become healthy. This effectively reflects the behaviour that `docker-compose up` has.  
Resolves #7453. 

# Implementation Summary
- `Project.up` now takes care of creating the one-off container:
  - This is powered by new `one_off` `ConvergencePlan`
  - One-off containers also accept option overrides
- `Project.up` doesn't start the one-off container, this is still responsibility of `run_one_off_container`

# Pre-commit hooks
According to `script/test/default` pre-commit hooks pass:
```
Check for added large files..............................................Passed
Check docstring is first.................................................Passed
Check for merge conflicts................................................Passed
Check Yaml...............................................................Passed
Check JSON...............................................................Passed
Debug Statements (Python)................................................Passed
Fix End of Files.........................................................Passed
Flake8...................................................................Passed
Tests should end in _test.py.............................................Passed
Fix requirements.txt.....................................................Passed
Trim Trailing Whitespace.................................................Passed
Reorder python imports...................................................Passed
```

# Other Changes
- ~`service.py` and `main.py` have 2 additional changes I had to apply in order to get pre-commit hooks (namely, `flake8`) to pass~ (rebased to pick this up from main branch)
- Gotten rid of `force` parameter in `remove_container` since it doesn't actually get used

# User Experience
An extra `Creating <project>_<service>_run ... done` will be printed to the terminal in addition to dependency information that's already printed. Examples of happy paths:
```
$ docker-compose run web
Starting docker-compose-example_redis_1 ... done
Creating docker-compose-example_web_run ... done
 * Debugger is active!
 * Debugger PIN: 302-617-874

$ docker-compose run test bash
Starting docker-compose-example_db_1 ... done
Creating docker-compose-example_test_run ... done
root@008cf9d35de0:/# ls
bin   dev                  docker-entrypoint.sh  home  lib64  mnt  proc  run   srv  tmp  var
boot  docker-entrypoint.d  etc                   lib   media  opt  root  sbin  sys  usr
```

We're also now getting `up` checks for free. For example, if a dependency is unhealthy and `condition: service_healthy` was specified, the following will happen:
```
$ docker-compose run web bash
Starting docker-compose-example_redis_1 ... done

ERROR: for web  Container "830730e0db0c" is unhealthy.
```

# Testing
- An acceptance test was added; it passes
- Tested it manually to verify that `docker-compose run` indeed works as expected

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -1298,31 +1298,28 @@ def build_one_off_container_options(options, detach, command):
 
 def run_one_off_container(container_options, project, service, options, toplevel_options,
                           toplevel_environment):
-    if not options['--no-deps']:
-        deps = service.get_dependency_names()
-        if deps:
-            project.up(
-                service_names=deps,
-                start_deps=True,
-                strategy=ConvergenceStrategy.never,
-                rescale=False
-            )
-
-    project.initialize()
-
-    container = service.create_container(
-        quiet=True,
+    detach = options.get('--detach')
+    use_network_aliases = options.get('--use-aliases')
+    containers = project.up(
+        service_names=[service.name],
+        start_deps=not options['--no-deps'],
+        strategy=ConvergenceStrategy.never,
+        detached=detach,
+        rescale=False,
         one_off=True,
-        **container_options)
+        override_options=container_options,
+    )
+    try:
+        container = next(c for c in containers if c.service == service.name)
+    except StopIteration:
+        raise OperationFailedError('Could not bring up the requested service')
 
-    use_network_aliases = options['--use-aliases']
-
-    if options.get('--detach'):
+    if detach:
         service.start_container(container, use_network_aliases)
         print(container.name)
         return
 
-    def remove_container(force=False):
+    def remove_container():
         if options['--rm']:
             project.client.remove_container(container.id, force=True, v=True)
 
@@ -1355,7 +1352,7 @@ def run_one_off_container(container_options, project, service, options, toplevel
             exit_code = 1
     except (signals.ShutdownException, signals.HangUpException):
         project.client.kill(container.id)
-        remove_container(force=True)
+        remove_container()
         sys.exit(2)
 
     remove_container()
--- a/compose/project.py
+++ b/compose/project.py
@@ -565,6 +565,8 @@ class Project(object):
            renew_anonymous_volumes=False,
            silent=False,
            cli=False,
+           one_off=False,
+           override_options=None,
            ):
 
         if cli:
@@ -584,7 +586,11 @@ class Project(object):
         for svc in services:
             svc.ensure_image_exists(do_build=do_build, silent=silent, cli=cli)
         plans = self._get_convergence_plans(
-            services, strategy, always_recreate_deps=always_recreate_deps)
+            services,
+            strategy,
+            always_recreate_deps=always_recreate_deps,
+            one_off=service_names if one_off else [],
+        )
 
         def do(service):
 
@@ -597,6 +603,7 @@ class Project(object):
                 start=start,
                 reset_container_image=reset_container_image,
                 renew_anonymous_volumes=renew_anonymous_volumes,
+                override_options=override_options,
             )
 
         def get_deps(service):
@@ -628,7 +635,7 @@ class Project(object):
         self.networks.initialize()
         self.volumes.initialize()
 
-    def _get_convergence_plans(self, services, strategy, always_recreate_deps=False):
+    def _get_convergence_plans(self, services, strategy, always_recreate_deps=False, one_off=None):
         plans = {}
 
         for service in services:
@@ -638,6 +645,7 @@ class Project(object):
                 if name in plans and
                 plans[name].action in ('recreate', 'create')
             ]
+            is_one_off = one_off and service.name in one_off
 
             if updated_dependencies and strategy.allows_recreate:
                 log.debug('%s has upstream changes (%s)',
@@ -649,11 +657,11 @@ class Project(object):
                 container_has_links = any(c.get('HostConfig.Links') for c in service.containers())
                 should_recreate_for_links = service_has_links ^ container_has_links
                 if always_recreate_deps or containers_stopped or should_recreate_for_links:
-                    plan = service.convergence_plan(ConvergenceStrategy.always)
+                    plan = service.convergence_plan(ConvergenceStrategy.always, is_one_off)
                 else:
-                    plan = service.convergence_plan(strategy)
+                    plan = service.convergence_plan(strategy, is_one_off)
             else:
-                plan = service.convergence_plan(strategy)
+                plan = service.convergence_plan(strategy, is_one_off)
 
             plans[service.name] = plan
 
--- a/compose/service.py
+++ b/compose/service.py
@@ -388,8 +388,11 @@ class Service(object):
             platform = self.default_platform
         return platform
 
-    def convergence_plan(self, strategy=ConvergenceStrategy.changed):
+    def convergence_plan(self, strategy=ConvergenceStrategy.changed, one_off=False):
         containers = self.containers(stopped=True)
+
+        if one_off:
+            return ConvergencePlan('one_off', [])
 
         if not containers:
             return ConvergencePlan('create', [])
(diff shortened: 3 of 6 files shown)


EricHripko:
From user perspective, this is how this addition feels:
```
$ docker-compose run --rm mydfesvc
Starting myproj_mydb_1    ... done
ERROR: for mydb  Container "dfd548facba4" is unhealthy.
$
```

EricHripko:
I am not a huge fan of doing a `sleep` - so would appreciate any suggestions you have here 🙂

EricHripko:
Rebased PR on latest `master` to pick up the `flake8` fixes.

EricHripko:
Hey @rumpl @ndeloof @ulyssessouza - do you have any feedback on this?

EricHripko:
Hey @rumpl @ndeloof @ulyssessouza - could you take a look at this PR when you got a moment please?

EricHripko:
Now that Compose Spec [has been landed](https://github.com/docker/compose/pull/7588#event-3588894241), it make sense to merge this PR since it ensures the  `dokcer-compose` is [compliant with the spec](https://github.com/compose-spec/compose-spec/blob/master/spec.md#long-syntax-1).

aiordache:
Hi @EricHripko I can have a look at this but can you rebase it first please?

EricHripko:
@aiordache thanks for working with me on this 👍 I've rebased the PR and fixed the tests - looks all green and ready to review 🙂

aiordache:
Thank you @EricHripko! I had a look and tested the workflow and it works pretty well. 

I was thinking though why don't we let the `project.up` do the complete check(including the one_off container)  as it is anyway being called before the new method you added. That would require to change the way we create the one_off container which for now doesn't seem easy to follow.
What do you think about it?

EricHripko:
Thank you for taking a look @aiordache 👍 

I've re-implemented this using `project.up` as suggested. In order to make this possible, I've added `one_off` convergence strategy and `override_options` to customise the one-off container. This is what it looks like from user perspective:
```
# If dependency is unhealthy: now the same as up
$ docker-compose run web bash
Starting docker-compose-example_redis_1 ... done

ERROR: for web  Container "830730e0db0c" is unhealthy.

# If dependency is healthy: now follows new workflow more similar to up
$ docker-compose run web
Starting docker-compose-example_redis_1 ... done
Creating docker-compose-example_web_run ... done
 * Debugger is active!
 * Debugger PIN: 302-617-874
```

I still need to fix up a few tests, but would love to hear if I'm heading in the right direction 🙂

aiordache:
I tested the new changes and a small thing to fix is, on a successful run command, we need to press enter to get the shell prompt of the container.
Apart from this, it works great. Thank you @EricHripko! I like having the same workflow as `up`.

EricHripko:
Thanks for your feedback @aiordache 👍 That's interesting - I've not noticed this issue despite testing natively on Windows and inside WSL2. Will try on Mac or a Linux VM to see if I can reproduce.

As an aside, I've just updated `CLITestCase::test_run_handles_sig*` to swap `top` for `sleep infinity`. I assume that's fine since it looks like `top` is just used as a long-running process. The reason for this was that `top` failed the tests for some reason after my changes (and I couldn't reproduce the failures). This might be related to the issue you mentioned - so will investigate more and revert the change if that's the culprit.

EricHripko:
@aiordache, I've had a go at Mac as well and I am unable to reproduce this. `docker-compose run` with these changes drops me straight into the interactive shell - similar to what previous version (`1.25.5` on my Mac) does. What platform are you on? Is there a specific project I could try to reproduce this with?

aiordache:
@EricHripko I tested it on two different machines running Arch linux (bash shell).
The compose file I used:
```
services:
  db:
    image: redis
    healthcheck:
      test: "exit 0"
  test:
    image: nginx
    depends_on:
      db:
        condition: service_healthy
```

EricHripko:
Thank you for sharing this @aiordache 👍 I've tried this on Ubuntu (in bash) and had the following results:
```
$ docker-compose run test
Creating docker-compose-example_db_1 ... done
Creating docker-compose-example_test_run ... done

^C
$ docker-compose run test bash
Starting docker-compose-example_db_1 ... done
Creating docker-compose-example_test_run ... done
root@008cf9d35de0:/# ls
bin   dev                  docker-entrypoint.sh  home  lib64  mnt  proc  run   srv  tmp  var
boot  docker-entrypoint.d  etc                   lib   media  opt  root  sbin  sys  usr
root@008cf9d35de0:/# exit
```

Will install Arch to further eliminate the affecting factors. Could you please share the command that exhibits the bug too?

aiordache:
@EricHripko I tested this command:
```
$ docker-compose run --rm test sh
```

EricHripko:
Ah - great! I can indeed confirm that I see the same issue with `sh` (even on Ubuntu). Will dig in to see why this happens 👍

thaJeztah:
Curious; how does this interact with `--no-deps` ? Does it produce an error in that case, or just skips the checks?

EricHripko:
@thaJeztah, it'll skip the checks if `--no-deps` is specified 🙂

EricHripko:
@aiordache, found the root cause for the issue. It appears that my changes were trying to "start" the container twice. This caused weirdness in some shells. I've simplified the code now and will push once tests look good.

EricHripko:
@aiordache this should be all green and ready for review 🙌 I've also updated the PR [Overview](https://github.com/docker/compose/pull/7457#issue-418597891) to include the latest information about how it was implemented and what UX is like 🙂

compose/cli/main.py
```diff
@@ -1298,31 +1298,25 @@ def build_one_off_container_options(options, detach, command):
...
+        strategy=ConvergenceStrategy.never,
+        detached=detach,
+        rescale=False,
         one_off=True,
-        **container_options)
-
-    use_network_aliases = options['--use-aliases']
+        override_options=container_options,
+    )
+    container = [c for c in containers if c.service == service.name][0]
```
aiordache, line 1312: Could we do a check before accessing this please? Just to make sure we won't throw an exception to users.

aiordache:
LGTM! Thank you @EricHripko for the effort to get this done. I have tested the latest changes and it works great! If you check my last comment and stash the commits, I could merge this tomorrow unless anybody else sees an issue with it.

compose/cli/main.py
EricHripko: Of course 👍 I've modified this to raise an `OperationFailedError` in case target service isn't in the container list. This is what it'll look like to a user in case they somehow manage to get in this state:
```
$ docker-compose run --rm test sh
Creating docker-compose-example_db_1 ... done
Creating docker-compose-example_test_run ... done
ERROR: Could not bring up the requested service
```

EricHripko:
Awesome 🎉 Thank you for your review and mentoring throughout this PR - really appreciate it 🙂 I've addressed your feedback and squashed the PR down to a single commit.

Issue #7489, opened 2020-05-30
[compose-in-a-container] Setup volume mounts based on -f, --file argument for compose in a container.


EricsonMacedo:
When running docker-compose through compose in a container values given to -f, --file argument aren't evaluated, as such, sharing volumes in this situation doesn't work.

Evaluate values given to -f, --file argument when running compose in a container and setup volume mounts in order to make sharing volumes to work as expected in this situation.

FlyingOnion:
I'm having similar question when I use drone (docker runner) to build CI/CD pipelines. How do I update the cluster (not a real one but I know no more other words) using compose in a container?

EricsonMacedo:
> I'm having similar question when I use drone (docker runner) to build CI/CD pipelines. How do I update the cluster (not a real one but I know no more other words) using compose in a container?

@FlyingOnion, sorry for no previous reply.
After #7490 was merged, you can use `run.sh -f <value>`, `run.sh --file <value>` or `docker-compose -f <value>` or `docker-compose --file <value>` depending on how you "installed" and named `run.sh`

EricsonMacedo:
Merged in #7490

PR #7490, opened 2020-05-30
Setup environment variables for compose.

### LINKED ISSUE #7488: [compose-in-a-container] Setup host environment variables to compose in a container to run interpolation
When running docker-compose through compose in a container general environment variables from host isn't setup into container, as such, environment variable interpolation doesn't work like in the binary version.

Setup environment variables from host to work as expected, like the binary version, for compose config and context in compose in a container.

EricsonMacedo:
- Setup environment variables to work as expected for compose
  config and context in container mode.
- Setup volume mounts based on -f, --file argument for compose
  config and context.

Signed-off-by: Ericson Macedo <macedoericson@gmail.com>

Resolves #7488, #7489


nioncode:
This is a much needed improvement. I was debugging for hours why my container did not behave like I thought it would on my odroid (where I use docker-compose in a container), but everything worked fine on my desktop machine.
I then noticed that the environment variables are not applied when setting them through the env variables in the shell, but work just fine in the `.env` file. Initially, I thought there might be a bug in docker-compose, but after not being able to reproduce it locally I came to the conclusion that this is a problem of running docker-compose through a container.

I now applied your patch locally and everything works fine now. It would be great to have this merged soon, as otherwise docker-compose running in a container behaves completely differently than running docker-compose natively.

ulyssessouza (approved):
LGTM!

PR #7506, opened 2020-06-04
Pin all indirect dependencies

ulyssessouza:
This PR is due to recent problems regarding the version change of `wcwidth` that broke the build on `master`

The intent of the new file `requirements-indirect.txt` is to pin the version of all the indirect dependencies.
That was generated by creating a new virtualenv, installing all the dependencies in `requirements.txt`, `requirements-dev.txt` and `requirements-build.txt`, running `pip freeze` and then removing all the packages in the intersection with the union of the 3 others.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -52,6 +52,7 @@ WORKDIR /code/
 RUN pip install virtualenv==16.2.0
 RUN pip install tox==2.9.1
 
+COPY requirements-indirect.txt .
 COPY requirements.txt .
 COPY requirements-dev.txt .
 COPY .pre-commit-config.yaml .
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,5 +1,6 @@
 include Dockerfile
 include LICENSE
+include requirements-indirect.txt
 include requirements.txt
 include requirements-dev.txt
 include tox.ini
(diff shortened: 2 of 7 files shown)


thaJeztah:
> That was generated by creating a new virtualenv, installing all the dependencies in requirements.txt, requirements-dev.txt and requirements-build.txt, running pip freeze and then removing all the packages in the intersection with the union of the 3 others.

Would it be an idea to have a Makefile target for that (so that it's easy to update the list of indirect deps?)

ulyssessouza:
@thaJeztah Yep! But the problem I got for now was for python2 vs python3 stuff that has conditions. And this is manual

thaJeztah:
Gotcha! Also sorry for my drive-by comment, just thinking that it would be easy for these dependencies to get out of sync, so if we could make maintenance of that easier, that'd be great 😅 

> Hopefully after merging the removal of python2 that will get easier

Yeah, sounds like a good thing in any case 👍

ulyssessouza:
> sorry for my drive-by comment

Please! Always do that! (And if you do this as a PR I pay you a beer :smiley:)

There is also another thing that will play. That's the `dependabot` that should propose upgrades in this libs.

PR #7516, opened 2020-06-06
Update Docker Compose logo

StefanScherer:
This PR updates the Docker Compose logo to the one that is used at https://www.compose-spec.io

I only changed scaled it down with

```
convert ~/Desktop/compose_brandmark.png -resize 200 logo.png
```


ndeloof:
👎 This is the logo for the Compose specification, not intended to be used by a specific implementation, or replace the docker-compose logo. Also I personally enjoy the existing docker-compose logo ;)

StefanScherer:
OK, never mind. Thanks for the clarification. 😀

PR #7537, opened 2020-06-16
Upgrade Windows dependency

ofek:
See https://github.com/docker/docker-py/pull/2593

`pypiwin32` is completely deprecated as `pywin32` moved to GitHub and now ships binary wheels itself. Also, it seems the bug introduced in `222` has been fixed.

https://github.com/mhammond/pywin32
https://pypi.org/project/pywin32/#files
https://github.com/mhammond/pywin32/pull/1153


ofek:
@ulyssessouza

PR #7543, opened 2020-06-17
Add `--` as a seperator of flags and arguments.

### LINKED ISSUE #7540: docker-compose does not understand flag/argument separator
## Description of the issue

`docker-compose` does not handle `--` when passed as a separator of flags and arguments.

```
docker-compose version 1.26.0, build d4451659
```

```
 Version:           19.03.11
 API version:       1.40
 Go version:        go1.13.10
 Git commit:        42e35e61f3
 Built:             Mon Jun  1 09:12:23 2020
 Experimental:      false

 Engine:
  Version:          19.03.11
  Go version:       go1.13.10
  Git commit:       42e35e61f3
  Built:            Mon Jun  1 09:10:55 2020
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.2.13
 runc:
  Version:          1.0.0-rc10
 docker-init:
  Version:          0.18.0
  GitCommit:        fec3683
```

```
services:
  eg:
    image: busybox:1.31.1-musl
version: '3.2'
```

Notice that the following succeeds:

```
$ docker run -- busybox:1.31.1-musl echo -hello
-hello
```

(the `--` is being used, as is [common practice](https://linux.die.net/man/3/getopt), to clearly say where the flags for the `run` command stop, and the arguments start.)

However the following fails:

```
$ docker-compose run -- eg echo -hello
ERROR: No such service: --
```

### Observed result

As above.

### Expected result

The `docker-compose` command above should succeed in exactly the same way as the `docker run` command.

As above.

## Additional information

```
Ubuntu 20.04 LTS
```

Installed via https://docs.docker.com/compose/install/#install-compose-on-linux-systems

jarulsamy:
`--` is common practice to clearly say where the
flags stop and and arguments start.

Fixes #7540 

Given the following example:
```
version: "3"

services:
  test:
    container_name: test
    image: hello-world
```

Current behavior:
```
$ docker-compose run -- test
ERROR: No such service: --
```

Post fix:
```
$ docker-compose run -- test

Hello from docker!
. . .
```
This is just a series of small changes to the `docopt` argparser. 

Signed-off-by: Joshua Arulsamy <joshua.gf.arul@gmail.com>

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -180,7 +180,7 @@ class TopLevelCommand(object):
     """Define and run multi-container applications with Docker.
 
     Usage:
-      docker-compose [-f <arg>...] [options] [COMMAND] [ARGS...]
+      docker-compose [-f <arg>...] [options] [--] [COMMAND] [ARGS...]
       docker-compose -h|--help
 
     Options:
@@ -257,7 +257,7 @@ class TopLevelCommand(object):
         e.g. `composetest_db`. If you change a service's `Dockerfile` or the
         contents of its build directory, you can run `docker-compose build` to rebuild it.
 
-        Usage: build [options] [--build-arg key=val...] [SERVICE...]
+        Usage: build [options] [--build-arg key=val...] [--] [SERVICE...]
 
         Options:
             --build-arg key=val     Set build-time variables for services.
@@ -423,7 +423,7 @@ class TopLevelCommand(object):
         """
         Receive real time events from containers.
 
-        Usage: events [options] [SERVICE...]
+        Usage: events [options] [--] [SERVICE...]
 
         Options:
             --json      Output events as a stream of json objects
@@ -448,7 +448,7 @@ class TopLevelCommand(object):
         """
         Execute a command in a running container
 
-        Usage: exec [options] [-e KEY=VAL...] SERVICE COMMAND [ARGS...]
+        Usage: exec [options] [-e KEY=VAL...] [--] SERVICE COMMAND [ARGS...]
 
         Options:
             -d, --detach      Detached mode: Run command in the background.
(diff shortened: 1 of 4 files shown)


compose/cli/main.py
```diff
@@ -257,7 +257,7 @@ def build(self, options):
         e.g. `composetest_db`. If you change a service's `Dockerfile` or the
         contents of its build directory, you can run `docker-compose build` to rebuild it.
 
-        Usage: build [options] [--build-arg key=val...] [SERVICE...]
+        Usage: build [options] [--] [--build-arg key=val...] [SERVICE...]
```
thaJeztah, line 260: this doesn't look correct, as it would mean `--` can be before `--build-arg` and this means build args would not be processed as option?

jarulsamy: Maybe I misunderstood `--build-arg`, but as part of the POSIX standard, I believe, `--`  is intended to signify no flags are present. 

For example with `rm` it can be used on files with `-` in the name, e.g, `rm -- -foo`

jarulsamy:
I agree, including it in the help / usage may be unnecessary, I will investigate if that is possible. 

Also it appears this is not building, all the tests passed on my machine. I'll write some additional tests to see if I can pinpoint the issue. 

Thanks for the quick feedback!

compose/cli/main.py
thaJeztah: It means "stop processing" after,
so I could do (silly example);

```
docker-compose build --pull -- --pull
```

to build a service named `--pull` (but also have the `--pull` option set

jarulsamy:
I added tests and fixed the order to support the build flags we discussed earlier. I believe it is not possible to hide the `--` option from the help / usage, since `docopt` derives it's CLI functionality from the help menu itself.

Issue #6676, opened 2019-04-28
Flag to exclude services


montanaflynn:
**Is your feature request related to a problem? Please describe.**

For development we have several containers we want to use such as postgres, elasticsearch, kibana, etc... and the docker-compose.yml file has several of these as services such as postgres, elasticsearch and kibana. Sometimes we don't want to run kibana as it's mostly used for debugging. 

Another scenario is developing microservices where we want a full environment but also the ability to toggle one service off to develop it locally.

**Describe the solution you'd like**

Adding an `--exclude-service` flag would allow us to use our existing `docker-compose` files without having to comment out specific services before running `docker-compose` to remove specific services.

`docker-compose up --exclude-service kibana`

**Describe alternatives you've considered**

Commenting out the services we don't want to start, having multiple docker-compose files that are identical except for each combination of services (impractical when you have several services)

**Additional context**

I think this is a common request, I found some issues asking for optional services that were somewhat related. 

Having `--exclude-service` along with being able to specify which services to start like this `docker-compose up postgres elasticsearch widget-api --exclude-service user-api` would work well to allow users to have one master `docker-compose.yml` file and specify at runtime which services to start.

jcsirot:
Hi @montanaflynn Thank you for the ticket. We are going to evaluate the feature you are requesting.

I also think you may be interested into the https://github.com/docker/app project which gives the possibility to templatize the compose file. The project is still in early development so do not hesitate to open tickets if you want some features to be added.

montanaflynn:
@jcsirot thanks for the response! I'll check out app, as I've not heard of it until now.

I think this flag would alleviate the pain of https://github.com/docker/compose/issues/542, https://github.com/docker/compose/issues/1547, https://github.com/docker/compose/issues/1896, https://github.com/docker/compose/issues/6660 and many other similar issues.

Eshaan7:
I implemented this using environment variables. Solution on stackoverflow: https://stackoverflow.com/a/61107015/10534470

Issue #4650, opened 2017-03-21
Build-only flag


ekkis:
I've read through https://github.com/docker/compose/issues/942 but I think that thread primarily focuses on the automatic running of a container.  I think there are more states, actually, and for my use case, being able to build an image without creating and/or running a container would be very useful.

I think what makes the most sense is to have 3 flags (defaulted to `true`):

```
build: true
create: true
start: true
```

or perhaps an array of options could open up the choices for later distinctions:

```
process:
  - build
  - create
  - start
```

as context, I want to be able to build an image along with all my other stuff, but the creation of containers is delegated to a user-driven process and thus docker-compose should not create the container for me

shin-:
Hi!

This is essentially the same ask as #1896 - please refer to the discussion there.

ekkis:
yes, I had read that post before but it doesn't address specifically the issue of being able to build only

Issue #4096, opened 2016-10-31
Proposal: exec shortcuts


mathroc:
I'd like to propose a new feature to simplify using `docker-compose exec`.

In some case it could be an alternative #1896

docker-compose.yml example:

```yml
version: "2"
services:
  my_service:
    image: my/service
shortcuts:
  backup:
    service: my_service
    entrypoint: rsync /some/thing
    command: ssh://some.where/
    privileged: true|false
    user: <username>
```

and usage:

```bash
docker-compose sc backup
docker-compose sc backup ssh://some.where.else/
```

or maybe (if `@backup` is not a valid service name)

```bash
docker-compose exec @backup
docker-compose exec @backup ssh://some.where.else/
```

shin-:
Thanks for the suggestion!

We will have to take another good look at #1896 sooner or later, so any feedback is welcome there as well.

glensc:
I like the issue number here :D

vipseixas:
+1
Very useful idea.

mathroc:
I think this still ca be useful, profiles are great, but they don't cover all the same use cases

this could also be very useful on dev environment, , eg: for starting composer/phpcs/phpunit/etc/ (with php), npm/yarn/eslint/etc. (with js), etc.

Issue #3684, opened 2016-07-04
Specify unrunnable service


jylin:
Supposed I have a docker-compose.yml file like this:

```
version: '2'
services:
        base:
                build: ./base
                image: my-base
        service:
                container_name: my-service
                build: ./service
                image: my-service
                ports:
                        - 80:80
```

I mainly keep the "base" target in my compose file as a convenient way to build everything. In other words, I'm sort of using docker-compose.yml as a Makefile. This way I can just do `docker-compose up --build` to rebuild and re-run everything. However, the base image doesn't have an entrypoint, so if you run it, then it will quit right away. It would be nice if there were some way to specify that a service is not runnable so that compose doesn't even try to start it.

dnephin:
Duplicate of #1294

docker-compose is not designed to be a replacement for a Makefile.

dnephin:
If you're still interested in a "makefile" like tool that uses containers you should check out https://dnephin.github.io/dobi/. It's designed for exactly this use case.

PR #7548, opened 2020-06-20
Add auto_up option to exclude service from `docker-compose up` by default

### LINKED ISSUE #6742: Temporarily Disable/Enable containers in a docker-compose.yaml
Since it appears the approach is to keep blocking this feature request until it goes away, I'm again asking for the ability to enable/disable containers in a docker-compose.yaml file temporarily, with an 

enable:true or --disable <service>

kind of mechanism.

I get that the developers don't want to do this, and have suggested that we use multiple docker-compose.yaml files, but:

1) Valid uses cases have been presented over and over (since 2015 at least) on why this is a good feature, but I have one too...   I have a stack entirely used for local development purposes when we don't have AWS access.   If I intend to debug  only one of my applications in the stack, I don't need to bring up the other 12 applications that are also in my stack, or incur the cost of them starting up just to get my one app up.  I also don't want to remove the apps from the stack, because tomorrow, I might need them. I haven't heard a good reason why this use case isn't valid.

2) Using multiple docker-compose.yaml files generates warnings that we also then have to quiet because the containers in the other files are now considered "orphans" (which is STOOPID - I only have one stack!).  So, I had to turn off the warning about orphans, had to change my shutdown to force orphans to be removed, and then add code everywhere to run docker-compose multiple times based on flags I added in a startup script (which I wouldn't need if docker-compose supported this).  

3) docker-compose ps, logs, etc also don't show all stacks, so you have to also pass in the -f docker-file.yaml arg as well to all of those commands as well.  Now, so I don't spend all day typing, I have to have scripts to run what should be simple docker-compose commands.  This also means I have to run a whole bunch of docker-compose ps commands just to view the real status of my stack rather than a single command.

I get that this wasn't in the original design, but the developers simply putting their foot down on this and refusing to address it, even when presented with a fix, and blocking comments from users, just makes everyone else have to work harder to achieve the same goals, and puts a bad taste in your users mouth's.

I really hope someone new can take a fresh look into this.

acran:
This an updated version of #3047, now complete with added test cases.

This adds the new config option `auto_up` to services. When this is explicitly set to `false` the service will not be started when executing `docker-compose up` without service names (unless it is pulled in as a dependency of another service).

This allows to define _optional_ services in a `docker-compose.yml` which are only started when listed explicitly on the command line.

Resolves #6742, #1896 and many more reports requesting this feature or similar.

## Use case

For one example use case of the many listed in #1896 and linked issues consider the following `docker-compose.yml`:
~~~yml
version: '3.8'
services:
  app:
    image: php
    depends_on:
      - db

  db:
    image: mysql

  phpmyadmin:
    auto_up: false
    image: phpmyadmin/phpmyadmin
    depends_on:
      - db

  mailhog:
    auto_up: false
    image: mailhog/mailhog
~~~

This may be an example of a prototypical web app running on PHP and using a MySQL database. Most of the time one only wants to run the `app` and `db` service but for debugging/development purposes two additional, _optional_ services are defined here:
* `phpmyadmin`: simple web interface to access and manipulate the raw database
* `mailhog`: a mocked mail transfer agent to be able to test and debug e-mails sent by the application

This now allows to start the main services with a simple `docker-compose up` and the _optional_ services by listing them:
~~~sh
# start only app and db
# without the auto_up option this would also start phpmyadmin and mailhog
docker-compose up

docker-compose up phpmyadmin # starts phpmyadmin, as well as db if necessary
docker-compose up mailhog # starts only the mailhog service
~~~

## Implementation details

This feature is realized by slightly changing `get_services` in the `Project` module. If no service names are passed via the command line it falls back to _all_ services in the `docker-compose.yml` - which are [now filtered for services which have `auto_up` explicitly set to `false`](https://github.com/acran/compose/commit/3ef48f1bd3975ef1b1570b56eff3c2a0af7db53d#diff-c8c735e30c07f4c7494904b4570d1541R188-R194).

When `auto_up` is set to `true`or is omitted the services are included in the default set as before. Additionally this change only affects the `docker-compose up` and `docker-compose create` commands. Consider this `docker-compose.yml`:
~~~yml
version: '3.8'
services:
  service1:
    image: alpine
    command: echo "I'm service 1"

  service2:
    auto_up: false
    image: alpine
    command: echo "I'm service 2"

  service3:
    image: alpine
    depends_on:
      - service1
      - service2
    command: echo "I'm service 3"

  service4:
    auto_up: false
    image: alpine
    depends_on:
      - service2
    command: echo "I'm service 4"
~~~

With this `docker-compose` behaves as follows:
~~~sh
docker-compose up service1 # starts only service1
docker-compose up service2 # starts only service2

docker-compose up
# starts service1, service2, service3:
# service4 is skipped because of auto_up: false
# service2 is pulled in as dependency of service3

docker-compose up service4 # starts service4 and service2 as its dependency

docker-compose stop # stops all (4) services

docker-compose down # stops and removes all (4) services
# note that --remove-orphans is not needed here

docker-compose logs # shows logs of all 4 services
~~~

## Alternatives

Currently the only way to achive this is by splitting the services into multiple `docker-compose.yml` files and pass them via the `--file` parameter:
~~~sh
# our example from above...
docker-compose up phpmyadmin

# ..,would be instead
docker-compose -f docker-compose.tools.yml up phpmyadmin

# but since phpmyadmin depends_on db we actually need multiple files
docker-compose -f docker-compose.yml -f docker-compose.tools.yml up phpmyadmin
~~~

As can be seen this is quite cumbersome and especially when using `docker-compose` for local development this a **huge** inconvenience as discussed in #1896 extensively and is confirmed by the amount of feature request for many years.

This PR now adds the simple option `auto_up` to the schema from version 2.0 up.

## Documentation

The documentation for this option can be found - and later merged from - here: https://github.com/acran/docker.github.io/commit/03ce9ed4fe282a346da52b69a9187c6807ef739f

## References

This or a similar feature was requested in: #6742 (most recent discussion), #1896 (main discussion), #3047 (previous PR), #697, #912, #1294, #1439, #1547, #2803, #3027, #3289

also related: #1661, #1988, #2773, #3684, #4096, #4650, #6676

### DIFF
--- a/compose/config/config.py
+++ b/compose/config/config.py
@@ -124,6 +124,7 @@ DOCKER_CONFIG_KEYS = [
 ]
 
 ALLOWED_KEYS = DOCKER_CONFIG_KEYS + [
+    'auto_up',
     'blkio_config',
     'build',
     'container_name',
--- a/compose/config/config_schema_v2.0.json
+++ b/compose/config/config_schema_v2.0.json
@@ -51,6 +51,7 @@
       "type": "object",
 
       "properties": {
+        "auto_up": {"type": "boolean"},
         "blkio_config": {
           "type": "object",
           "properties": {
--- a/compose/config/config_schema_v2.1.json
+++ b/compose/config/config_schema_v2.1.json
@@ -51,6 +51,7 @@
       "type": "object",
 
       "properties": {
+        "auto_up": {"type": "boolean"},
         "blkio_config": {
           "type": "object",
           "properties": {
--- a/compose/config/config_schema_v2.2.json
+++ b/compose/config/config_schema_v2.2.json
@@ -51,6 +51,7 @@
       "type": "object",
 
       "properties": {
+        "auto_up": {"type": "boolean"},
         "blkio_config": {
           "type": "object",
           "properties": {
--- a/compose/config/config_schema_v2.3.json
+++ b/compose/config/config_schema_v2.3.json
@@ -51,6 +51,7 @@
       "type": "object",
 
       "properties": {
+        "auto_up": {"type": "boolean"},
         "blkio_config": {
           "type": "object",
           "properties": {
--- a/compose/config/config_schema_v2.4.json
+++ b/compose/config/config_schema_v2.4.json
@@ -51,6 +51,7 @@
       "type": "object",
 
       "properties": {
+        "auto_up": {"type": "boolean"},
         "blkio_config": {
           "type": "object",
           "properties": {
--- a/compose/config/config_schema_v3.0.json
+++ b/compose/config/config_schema_v3.0.json
@@ -51,6 +51,7 @@
       "type": "object",
 
       "properties": {
+        "auto_up": {"type": "boolean"},
         "deploy": {"$ref": "#/definitions/deployment"},
         "build": {
           "oneOf": [
--- a/compose/config/config_schema_v3.1.json
+++ b/compose/config/config_schema_v3.1.json
@@ -62,6 +62,7 @@
       "type": "object",
 
       "properties": {
+        "auto_up": {"type": "boolean"},
         "deploy": {"$ref": "#/definitions/deployment"},
         "build": {
           "oneOf": [
--- a/compose/config/config_schema_v3.2.json
+++ b/compose/config/config_schema_v3.2.json
@@ -62,6 +62,7 @@
       "type": "object",
 
       "properties": {
+        "auto_up": {"type": "boolean"},
         "deploy": {"$ref": "#/definitions/deployment"},
         "build": {
           "oneOf": [
--- a/compose/config/config_schema_v3.3.json
+++ b/compose/config/config_schema_v3.3.json
@@ -73,6 +73,7 @@
       "type": "object",
 
       "properties": {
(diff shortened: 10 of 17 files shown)


acran:
Although I now also added test cases (which just all passed green :smile:) and spend some extra time checking all code paths this change is affecting for any unintended side effects, I appreciate any feedback on this PR.
I'll gladly incorporate any changes required to make this PR ready for merge!

ulyssessouza:
Hello @acran .
Thank you for the contribution, but as this is a change on the specification, the change has to be discussed and validated in https://github.com/compose-spec/compose-spec
Also, changing the already released schemas are not allowed. That's like changing the past.

I'm very sorry, but I cannot integrate this change before it. I invite you to create an issue on https://github.com/compose-spec/compose-spec to discuss it for the next version.

acran:
Hi @ulyssessouza 

Thank you for your feedback. I wasn't aware of the dedicated repository for for the compose file specification. Since this feature definitely requires a schema change I now created an issue there to discuss the proposed change: https://github.com/compose-spec/compose-spec/issues/87

remort:
What's happened with this PR? Still can't see it functioning in the modern versions of docker-compose.

PR #7549, opened 2020-06-20
Fixing imprecise timing in timeparse module.

alexrecuenco:
When testing `tests/unit/config/config_test.py`, the result of the test
are very reliant on the exact floating point value inaccuracies
cancelling each other out.

The main issue is that

```python
MULTIPLIER['nano'] == 1.0 / 1000.0 / 1000.0 / 1000.0 == 9.999999999999999e-10
```

Therefore, on some serialization tests, when it goes through `parse_nanoseconds_int`,
it was calculating `int(1 / 9.99...)`, that is slightly above 1, and it
was succeeding because it was creating a number slightly above 1 and int
behaves similarly to floor, truncating the number. That doesn't mean that it works correctly serializing all possible configurations

This can be fixed, and by doing that allow more precise and more stable timing.

- We make `timeparse.timeparse` use Decimal internally.
- We also add an `INVERSE_MULTIPLIER` instead to provide better arithmetic
  precision. I want to ask if it is possible to remove the MULTIPLIER
  that are floating point, I am not sure if anything else from another
  codebase is using it
- To allow an easier access, we replace the use of those dictionaries
  with two helper function to and from seconds (They attempt to use the
  more precise floatinig poinnt arithmetic possible, but for exact
  measuring, they can be passed `Decimal` numbers)
- We provide on `timeparse.timeparse` a flag called `exact` to allow
  returning the Decimal result. However, for backwards compatibility, we
  set it to `False` by default
- We modified `parse_nanoseconds_int` to use the exact measuring
- We changed the regex to factor out `FLOAT_NUMBER`.
- We removed the very outdated doctest that wasn't working, and added
  tests for the new code.

Signed-off-by: alexrecuenco <alejandrogonzalezrecuenco@gmail.com>

### DIFF
--- a/compose/timeparse.py
+++ b/compose/timeparse.py
@@ -31,17 +31,20 @@ https://golang.org/pkg/time/#ParseDuration
 # CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
 import re
+from decimal import Decimal
 
-HOURS = r'(?P<hours>[\d.]+)h'
-MINS = r'(?P<mins>[\d.]+)m'
-SECS = r'(?P<secs>[\d.]+)s'
-MILLI = r'(?P<milli>[\d.]+)ms'
-MICRO = r'(?P<micro>[\d.]+)(?:us|µs)'
-NANO = r'(?P<nano>[\d.]+)ns'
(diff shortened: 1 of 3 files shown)


glours:

PR #7578, opened 2020-07-01
Improve version help message

bfirsh:
Just a little thing. I was looking at `docker-compose --help` and saw "Docker-Compose" and cringed a little. :)

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -233,7 +233,7 @@ class TopLevelCommand(object):
       top                Display the running processes
       unpause            Unpause services
       up                 Create and start containers
-      version            Show the Docker-Compose version information
+      version            Show version information and quit
     """
 
     def __init__(self, project, options=None):
@@ -1118,7 +1118,7 @@ class TopLevelCommand(object):
     @classmethod
     def version(cls, options):
         """
-        Show version information
+        Show version information and quit.
 
         Usage: version [--short]
 


compose/cli/main.py
```diff
@@ -233,7 +233,7 @@ class TopLevelCommand(object):
       top                Display the running processes
       unpause            Unpause services
       up                 Create and start containers
-      version            Show the Docker-Compose version information
+      version            Show version of Compose and its dependencies
```
ulyssessouza, line 236:
```suggestion
      version            Show version of Docker-Compose and its dependencies
```

bfirsh: In my day "Docker Compose" was the software, "Compose" was a short hand, and `docker-compose` was only ever used in documentation to refer to the command you were calling. Maybe that's changed? Either way, surely it shouldn't be "Docker-Compose" (capitalized with a dash)? :)

bfirsh: Maybe to side step this problem it could just be "Show version of this command and its dependencies", haha.

Or, just "Show version".

bfirsh: Actually, [the documentation says this is still the case](https://docs.docker.com/compose/), unless the documentation is wrong:

> Compose is a tool for defining and running multi-container Docker applications.

A tool, not a specification!

thaJeztah: `Docker-Compose` (both capital) is definitely incorrect (should either be `docker-compose` (name of the binary), or `Docker Compose`). Perhaps avoid the name altogether and use something similar as the `docker` cli uses (`Print version information and quit`)

```suggestion
      version            Show version information and quit
```

(quit/exit, not sure what's clearer)? `curl` uses `quit`;

```
 -V, --version       Show version number and quit
```

bfirsh: Sounds good! I could nitpick further but this is much better than what is already there. 😂

bfirsh: Pushed your suggested changes in both places.

ulyssessouza: Cool! LGTM

PR #7588, opened 2020-07-06
Merge V2 - V3 compose file formats (optional version field)

aiordache:
This update introduces an unified compose format that should be compatible with all v2.x and v3.x compose files. 
The version field is made optional (the format is automatically detected from the structure of the file). 
** version was required to distinguish with 1.x format
For each version, docker-compose validated against its specific schema. Now we have only 2 schemas:
-  v1 schema 
-  compose_spec schema (merged v2.x and v3.x)

Related to the conversion from v3 to v2 using the `--compatibility` flag: the deploy field was being ignored unless this flag was passed and only one container per service was created. 

We now apply the properties from deploy regardless of the --compatibility flag.

Test sample:
```
version: '3.8'
services:
  test:
    image: nginx
    deploy:
      replicas: 3
```
**BEFORE:**

One container per service was created unless we pass `--compatibility`. 
```
$ docker-compose up -d
WARNING: Some services (test) use the 'deploy' key, which will be ignored. Compose does not support 'deploy' configuration - use `docker stack deploy` to deploy to a swarm.
Creating network "hello_default" with the default driver
Creating hello_test_1 ... done
```

**NOW:**
Version field is optional, we can either remove it or leave it as it is, result is the same:
```
services:
  test:
    image: nginx
    deploy:
      replicas: 3
```
```
$ docker-compose up -d
WARNING: Some services (test) use the 'deploy' key, which will be ignored. Compose does not support 'deploy' configuration - use `docker stack deploy` to deploy to a swarm.
Creating network "hello_default" with the default driver
Creating hello_test_1 ... done
Creating hello_test_2 ... done
Creating hello_test_3 ... done
```
Same we get from a v2 file:
```
services:
  test:
    image: nginx
    scale: 3
```
v2.x and v3.x schema merging relates to https://github.com/docker/compose/issues/7201
The removal of compatibility mode relates to https://github.com/docker/dev-tooling-team/issues/121


compose/cli/command.py
```diff
@@ -120,14 +118,14 @@ def unicode_paths(paths):
 
 def get_project(project_dir, config_path=None, project_name=None, verbose=False,
                 context=None, environment=None, override_dir=None,
-                compatibility=False, interpolate=True, environment_file=None):
+                interpolate=True, environment_file=None):
```
ulyssessouza, line 121: If any part of the code using this function in a positional way. 
This would break that.

compose/config/config.py
```diff
@@ -12,18 +12,13 @@
 from cached_property import cached_property
 
 from . import types
-from .. import const
 from ..const import COMPOSEFILE_V1 as V1
-from ..const import COMPOSEFILE_V2_1 as V2_1
-from ..const import COMPOSEFILE_V2_3 as V2_3
-from ..const import COMPOSEFILE_V3_0 as V3_0
-from ..const import COMPOSEFILE_V3_4 as V3_4
+from ..const import COMPOSEFILE_V4 as VERSION
```
ndeloof, line 16: There's no such thing as a V4 composefile schema. Better use `COMPOSE_SPEC` here

compose/config/config_schema_v4.0.json
```diff
@@ -1,10 +1,7 @@
 {
   "$schema": "http://json-schema.org/draft-04/schema#",
-  "id": "config_schema_v3.8.json",
+  "id": "config_schema_v4.0.json",
```
ndeloof, line 3: compose-spec.json

ndeloof: in an ideal world, this whole file should be removed from docker/compose repo and replaced by a (build-time?) reference to https://github.com/compose-spec/compose-spec/blob/master/schema/config_schema.json

ndeloof: did you checked if the current status of the compose-spec schema already includes all those changes or if there's anything missing ?

compose/const.py
```diff
@@ -24,56 +24,16 @@
...
-    COMPOSEFILE_V3_0: '1.25',
-    COMPOSEFILE_V3_1: '1.25',
-    COMPOSEFILE_V3_2: '1.25',
-    COMPOSEFILE_V3_3: '1.30',
-    COMPOSEFILE_V3_4: '1.30',
-    COMPOSEFILE_V3_5: '1.30',
-    COMPOSEFILE_V3_6: '1.36',
-    COMPOSEFILE_V3_7: '1.38',
-    COMPOSEFILE_V3_8: '1.38',
+    COMPOSEFILE_V4: '1.38',
```
ndeloof, line 33: This whole mechanism to define required API version based on requested schema version should be replaced by a check that a specific field in actual compose file can't be implemented by the actual engine API. But I assume this is a complex change and we should get this implemented in a further PR

compose/cli/command.py
aiordache: Haven't found any in the compose code.

compose/config/config.py
aiordache: After a discussion with @ulyssessouza, we decided it was easier to name the merged schema v4. So far the compose_spec schema is the v3.9. I would find it confusing to name it as you suggested as long as we don't have these schemas aligned.

compose/config/config_schema_v4.0.json
aiordache: I started from the compose-spec schema and added properties from v2.x and v3.x.

compose/const.py
aiordache: Yes, this will be the next step.

compose/config/config.py
ndeloof: then better to start by getting schema aligned :P
I'm fine with any name you want to use here, but please don't use "V4". This is highly confusing as this suggest a compatibility break between v3 and compose-spec, that we definitively want to avoid. There's no such thing like a v4 schema

ndeloof: By the way, compose-spec schema to be 3.9 is obsolete now we decided to drop version field
=> https://github.com/compose-spec/compose-spec/pull/88

compose/cli/command.py
ulyssessouza: :+1:

jamshid:
This is great I was just asking for `--compatibility` to be the default in https://github.com/docker/compose/issues/7594, so glad to hear the frustrating version incompatibility is being removed.
Have `memswap_limit` and `mem_swappiness` equivalents been added to `deploy` `resources`?

aiordache:
> 
> Have `memswap_limit` and `mem_swappiness` equivalents been added to `deploy` `resources`?

@jamshid we don't have yet equivalent properties in `deploy` for these. With the unified schema we bring them back as service properties.

I opened an issue on the compose-spec to see if we can have them added/moved to `deploy`: https://github.com/compose-spec/compose-spec/issues/90

melicerte:
I don't clearly understand the consequences of this update.

I have single service extension with `extends` working with version: "3.5".
This means I can mix up features from 2.x and 3.x like `shm_size` with no problem ?

EricHripko:
Yes @melicerte - you're understanding this correctly 👍 Post-1.27.x, `version` attribute becomes optional and you can mix 2.x/3.x attributes in the same Compose file 🙂

melicerte:
Thanks @EricHripko this is amazing !

melicerte:
However shouldn't documentation be updated accordingly ?
For example, concerning `extends` here https://docs.docker.com/compose/extends/ it could be specified that it works from docker-compose 1.27 and up.

EricHripko:
Yep, sounds like a [PR to docker.github.io](https://github.com/docker/docker.github.io/pulls) is needed 🙂

PR #7598, opened 2020-07-10
 docker-compose .env file support dynamic values #3849

yamatoRT:
Hello,

For implement the issue #3849, i propose this PR.
I add the possibility to import the environment variable with bash interpreter, if you add `:bash` after the env file path.

```yaml
web:
  env_file:
    tests/fixtures/env/get_uuid.env:bash
```
with `tests/fixtures/env/get_uuid.env` env file :
```bash
CUSTOM_UUID=$(id -u)
CUSTOM_GUID=$(id -g)
```
If no interpreter is specify, we use dotenv format for retro compatibility.

### DIFF
--- a/compose/config/config.py
+++ b/compose/config/config.py
@@ -680,10 +680,21 @@ def resolve_environment(service_dict, environment=None, interpolate=True):
     """
     env = {}
     for env_file in service_dict.get('env_file', []):
-        env.update(env_vars_from_file(env_file, interpolate))
+        (path, interpreter) = split_path_and_interpreter(env_file)
+        env.update(env_vars_from_file(path, interpolate, interpreter=interpreter))
 
     env.update(parse_environment(service_dict.get('environment')))
     return dict(resolve_env_var(k, v, environment) for k, v in env.items())
+
+
+def split_path_and_interpreter(env_file_path):
+    drive, envfile_config = splitdrive(env_file_path)
+
+    if ':' in envfile_config:
+        (path, interpreter) = envfile_config.split(':', 1)
+        return (drive + path, interpreter)
+    else:
+        return (env_file_path, 'dotenv')
 
 
 def resolve_build_args(buildargs, environment):
--- a/compose/config/environment.py
+++ b/compose/config/environment.py
@@ -3,6 +3,7 @@ import os
 import re
 
 import dotenv
+from envbash import load_envbash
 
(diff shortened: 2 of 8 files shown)


ulyssessouza:
Hello @yamatoRT !

First of all, thank you for the interest on `docker-compose`!

We have 2 issues here. The first is that changes in the `compose format` should be proposed in https://github.com/compose-spec/compose-spec (the `:bash` part must be validated there).
Second issue is that I don't think it's a work for `docker-compose` to spawn a new process specified in the configuration file.
Specially when there are other "safer ways" to do that. Like in: 
```
TESTVAR=$(pwd) && echo $TESTVAR
```
In the example above the TESTVAR is properly fulfilled when `echo` tries to resolve it. The same can be done with `docker-compose`

yamatoRT:
Hello,
Thank you for your answer. I create a issue in compose-spec for resolve the first issue.

The solution `TESTVAR=$(pwd) && echo $TESTVAR` or `TESTVAR=$(pwd) docker-compose up ...` works but I find she isn't elegant because the user must modify variables outsite docker-compose files.
For me, it's have many problems :
  - if you many dynamic variables, the command looks like `VAR1=$(...) VAR2=$(...) VAR3=$(...) docker-compose up ...`
  - it is necessary explains at each user how set variables and it is necessary create documentation for it
  - by mistake, the user may set variables with bad value, it may be necessary to send error message for help the user, and it is additional dev

While with the dynamic env file :
  -  if you many dynamic variables, the command looks simply `docker-compose up ...`
  - there's not documentation to create
  - the user cannot mystake because he doesn't set any variables.

thaJeztah:
I'd definitely have concerns with this change. Think of an env file containing something like;

    FOO=$(rm -r ~/)

Or similar evil things (take your private keys and upload them?) would be really dangerous, and unexpected, and easy to trick users into running an example repository with a compose file

yamatoRT:
Hi,
I have had the answer of compose-spec in this issue https://github.com/compose-spec/compose-spec/issues/94. Them too, they don't agree with use subshell bash. I will close this PR.

michaelkrieger:
Sad.  This would be useful for a `.env` file containing
`COMPOSE_FILE: $(ls -m /compose-files/docker-*.yml | sed s%', '%':'%g)`
To finally allow a directory of compose files and make long compose files easier.

PR #7713, opened 2020-08-31
Fix bump of docker-py on `setup.py`

### LINKED ISSUE #7699: [1.27.0_rc2] Version bump missing in setup.py (but advertised as a bugfix in change log)
Hi!

I noticed that the change log of 1.27.0_rc2 advertises `Fix float to bytes conversion via docker-py bump to 4.3.1` (see 
45c6730e6464cb6e22439c558d3036647b301610) yet `setup.py` does not depend on that new version, see https://github.com/docker/compose/blob/45c6730e6464cb6e22439c558d3036647b301610/setup.py#L35

Is the bump missing in `setup.py`?

Best, Sebastian

ulyssessouza:
Resolves #7699


thaJeztah (approved):
SGTM

Issue #7742, opened 2020-09-09
docker-compose 1.27.0 does not run service using "run" when it's scale is 0


crossdot:
```
version: '2.4'
services:
  once:
    scale: 0
    image: alpine
```
`docker-compose run --rm once ps`
produces
`ERROR: Could not bring up the requested service`

I use it to single run tests inside the project, in docker-compose 1.26 it was working as expected.

#### Output of docker-compose version

```
docker-compose version 1.27.0, build unknown
docker-py version: 4.3.1
CPython version: 3.8.5
OpenSSL version: OpenSSL 1.1.1g  21 Apr 2020
```

#### Output of docker version                                                                                                                                                          
```
Client:
 Version:           19.03.12-ce
 API version:       1.40
 Go version:        go1.14.5
 Git commit:        48a66213fe
 Built:             Sat Jul 18 01:33:21 2020
 OS/Arch:           linux/amd64
 Experimental:      false

Server:
 Engine:
  Version:          19.03.12-ce
  API version:      1.40 (minimum version 1.12)
  Go version:       go1.14.5
  Git commit:       48a66213fe
  Built:            Sat Jul 18 01:32:59 2020
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          v1.4.0.m
  GitCommit:        09814d48d50816305a8e6c1a4ae3e2bcc4ba725a.m
 runc:
  Version:          1.0.0-rc92
  GitCommit:        ff819c7e9184c13b7c2607fe6c30ae19403a7aff
 docker-init:
  Version:          0.18.0
  GitCommit:        fec3683
```

PR #7746, opened 2020-09-10
Set service scale to 1 on `compose run`

aiordache:
Regression in 1.27.0

Set the service scale to 1 on a `compose run` otherwise it will start the number of containers set with the scale property.

Resolves https://github.com/docker/compose/issues/7742


ulyssessouza (approved):
LGTM

Issue #7741, opened 2020-09-09
docker-compose 1.27.0 hangs when running containers with lots of output


hrpatel:
## Context information

**Output of `docker-compose version`**
```
docker-compose --version
docker-compose version 1.27.0, build 980ec85b
```

**Output of `docker version`**
```
docker version
Client: Docker Engine - Community
 Version:           19.03.12
 API version:       1.40
 Go version:        go1.13.10
 Git commit:        48a66213fe
 Built:             Mon Jun 22 15:45:36 2020
 OS/Arch:           linux/amd64
 Experimental:      false

Server: Docker Engine - Community
 Engine:
  Version:          19.03.12
  API version:      1.40 (minimum version 1.12)
  Go version:       go1.13.10
  Git commit:       48a66213fe
  Built:            Mon Jun 22 15:44:07 2020
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.2.13
  GitCommit:        7ad184331fa3e55e52b890ea95e65ba581ae3429
 runc:
  Version:          1.0.0-rc10
  GitCommit:        dc9208a3303feef5b3839f4323d9beb36df0a9dd
 docker-init:
  Version:          0.18.0
  GitCommit:        fec3683
```

**Output of `docker-compose config`**
```
docker-compose config
services:
  compose_bug:
    build:
      context: /home/hpatel/Revisions/docker-compose-bug
      dockerfile: Dockerfile
    command: hexdump /dev/urandom
version: '3'
```

**Dockerfile**
```
FROM ubuntu:18.04

# Disable frontend dialogs
ENV DEBIAN_FRONTEND noninteractive

# Make sure unionfs fuse is installed
RUN apt-get update \
    && apt-get -y install bsdmainutils \
    && apt-get -y autoremove \
    && apt-get -y clean \
```

## Steps to reproduce the issue

1. `docker-compose build`
2. `docker-compose run compose_bug`

### Observed result
Once a certain amount of logs/output is generated, docker-compose AND the container will stall (block?) and nothing else happens on the container. I was able to `exec` into the container to determine this. 

### Expected result
The command/script should complete

### Stacktrace / full error message

```
No errors, just hangs
```

### Other versions:
Cannot reproduce with 1.26 or 1.25 - the code above runs and continues until interrupted with `Ctrl+c`. 

### Steps to kill container:
1. Issue `docker kill ...` command 
2. Use `Ctrl+c` to stop the container

```
uname -a
[...]

PR #7752, opened 2020-09-10
Use `detach=true` for oneoff containers

aiordache:
Use `detach=true` on `project.up` for a oneoff container. We start it after and attach to it.

Resolves https://github.com/docker/compose/issues/7741


ulyssessouza (approved):
LGTM

PR #7769, opened 2020-09-17
Update changelog for 1.27.3 release

aiordache:
Post-release update of changelog.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,21 @@
 Change log
 ==========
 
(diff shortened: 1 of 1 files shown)


ndeloof:
suggestion: update CHANGELOG.md with a link to https://github.com/docker/compose/releases where the changelog is actually generated

ndeloof (approved):
I expected we don't have to handle such manual update with the automated release process, seems we can still improve :P

PR #7784, opened 2020-09-22
Remove path check for bind mounts

aiordache:
Regression in 1.27.3
Skip the path check for bind mounts for backwards compatibility.
Fixes https://github.com/docker/compose/issues/7786

### DIFF
--- a/compose/config/config.py
+++ b/compose/config/config.py
@@ -450,9 +450,6 @@ def format_device_option(entity_type, config):
     device = config['driver_opts'].get('device')
     if o and o == 'bind' and device:
         fullpath = os.path.abspath(os.path.expanduser(device))
-        if not os.path.exists(fullpath):
-            raise ConfigurationError(
-                "Device path {} does not exist.".format(fullpath))
         return fullpath
 
 


compose/config/config.py
```diff
@@ -450,9 +450,10 @@ def format_device_option(entity_type, config):
     device = config['driver_opts'].get('device')
     if o and o == 'bind' and device:
         fullpath = os.path.abspath(os.path.expanduser(device))
-        if not os.path.exists(fullpath):
-            raise ConfigurationError(
-                "Device path {} does not exist.".format(fullpath))
+        # we may want to enforce the path check later
+        # if not os.path.exists(fullpath):
+        #    raise ConfigurationError(
+        #        "Device path {} does not exist.".format(fullpath))
```
ulyssessouza, line 456: Remove?

ulyssessouza (changes requested):

gtardif (approved):

compose/config/config.py
JakeDEvans: why not check if docker_host is local?

ulyssessouza (approved):
LGTM

compose/config/config.py
aiordache: We do a check for the `local` driver a few lines above that but apart from that I think it is better to leave the docker engine to manage it to keep the old behaviour.

Issue #7102, opened 2019-12-20
Support ${parameter:+word} style variable substitution


andrewodri:
Most continuous integration systems have a `$CI` variable to that acts as a flag that lets you know when you're in a CI environment. While `docker-compose.yml` allows for substitution when a variable **is not** defined/set (e.g. [`${parameter:-word}`](https://docs.docker.com/compose/compose-file/#variable-substitution)), it would be great to allow for substitution when a variable **is** defined/set.

Bash allows you to define this with [the following syntax](https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html):

> `${parameter:+word}`
> 
> If _parameter_ is null or unset, nothing is substituted, otherwise the expansion of _word_ is substituted.

Here are some examples of some CI systems/services that support the `$CI` flag:

* https://docs.travis-ci.com/user/environment-variables/#default-environment-variables
* https://circleci.com/docs/2.0/env-vars/#built-in-environment-variables
* https://docs.gitlab.com/ee/ci/variables/predefined_variables.html#variables-reference

thaJeztah:
I think the official specs (non-bash) can be found here; https://pubs.opengroup.org/onlinepubs/009695399/utilities/xcu_chap02.html

ndeloof:
Sounds reasonable to me, both because this would enable interesting use-cases like the one you describe, and also because this would make the syntax consistent with what any shell script author would expect.

ndeloof:
We probably could support most or all [bash parameter expansion](https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html).
I wonder we could use `os.Expand` directly ?

thaJeztah:
I'd generally prefer sticking to the POSIX set of options, otherwise we'd end up implementing regex matching and the whole shebang

bradynpoulsen:
Following POSIX, would the substitution value also be evaluated for values?

 My use-case is I'm wanting to conditionally pass a value in my `command` when a value is set
```yaml
services:
  my-service:
    command: |
      -abc
      --do-something
      ${MY_DEFAULT_PASSWORD+--with-default-credentials "${MY_DEFAULT_USERNAME:-user}:${MY_DEFAULT_PASSWORD}"}
```

My workarounds are currently limited to:
1. throw a substitution error
    ```
    --with-default-credenitals "${MY_DEFAULT_USERNAME:-user}:${MY_DEFAULT_PASSWORD?Password is required!}"
    ```
[...]

PR #7795, opened 2020-09-25
Support nested variable expansion

### LINKED ISSUE #7794: Support nested parameter expansion
I've come across several instances at work where I've wanted to use nested parameter expansion to provide an alternative value if an environment variable isn't set, but this isn't currently supported.

For example, hard coded default values for parameter expansion are supported (`${FOO:-bar}`) but I can't use another environment variable as the default (`${FOO:-$BAR}`)

I would like to use nested parameter expansion in docker-config.yml files:

```
${FOO:-$BAR}
```

For now, I'm hard coding default values.

MarilynFranklin:
Signed-off-by: MarilynFranklin <marilyn.j.franklin@gmail.com>

Resolves #7794

### DIFF
--- a/compose/config/interpolation.py
+++ b/compose/config/interpolation.py
@@ -95,26 +95,30 @@ class TemplateWithDefaults(Template):
         """.format(
         delim=re.escape('$'),
         id=r'[_a-z][_a-z0-9]*',
-        bid=r'[_a-z][_a-z0-9]*(?:(?P<sep>:?[-?])[^}]*)?',
+        bid=r'[_a-z][_a-z0-9]*(?:(?P<sep>:?[-?])(.*}|[^}]*))?',
     )
 
     @staticmethod
     def process_braced_group(braced, sep, mapping):
         if ':-' == sep:
             var, _, default = braced.partition(':-')
+            default = Interpolator(TemplateWithDefaults, mapping).interpolate(default)
             return mapping.get(var) or default
         elif '-' == sep:
             var, _, default = braced.partition('-')
+            default = Interpolator(TemplateWithDefaults, mapping).interpolate(default)
             return mapping.get(var, default)
 
         elif ':?' == sep:
             var, _, err = braced.partition(':?')
+            err = Interpolator(TemplateWithDefaults, mapping).interpolate(err)
             result = mapping.get(var)
             if not result:
                 raise UnsetRequiredSubstitution(err)
             return result
         elif '?' == sep:
             var, _, err = braced.partition('?')
+            err = Interpolator(TemplateWithDefaults, mapping).interpolate(err)
             if var in mapping:
                 return mapping.get(var)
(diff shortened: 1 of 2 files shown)


ndeloof:
This would follow bash parameter substitution
https://tldp.org/LDP/abs/html/parameter-substitution.html

so my +1
_but_ this should then also be documented on https://github.com/compose-spec/compose-spec/blob/master/spec.md#interpolation
and implemented by https://github.com/compose-spec/compose-go so we get the same behaviour by other go-based docker tools.

ndeloof:
Also see discussion on https://github.com/docker/compose/issues/7102 about posix vs bash-specific
Reading https://pubs.opengroup.org/onlinepubs/009695399/utilities/xcu_chap02.html I'm not sure posix do define support for such nested substitutions

MarilynFranklin:
Thank you for the quick review!

> but this should then also be documented on https://github.com/compose-spec/compose-spec/blob/master/spec.md#interpolation
and implemented by https://github.com/compose-spec/compose-go so we get the same behaviour by other go-based docker tools.

I'm happy to work on updating the documentation and compose-go if you want to move forward with this!

> Reading https://pubs.opengroup.org/onlinepubs/009695399/utilities/xcu_chap02.html I'm not sure posix do define support for such nested substitutions

It looks like there is an example of nested _command_ substitution: https://pubs.opengroup.org/onlinepubs/009695399/utilities/xcu_chap02.html#tag_02_06_02_01
```
${parameter:-word}
  In this example, ls is executed only if x is null or unset. (The $( ls) command substitution notation is explained in Command Substitution.)
    ${x:-$(ls)}
```

This makes it seem like "word" in `${parameter:-word}` may be [expanded](https://pubs.opengroup.org/onlinepubs/009695399/utilities/xcu_chap02.html#tag_02_06).

I'm not sure if I'm using the right terminology here, but that was my best interpretation :)

I'm interested in hearing your thoughts!

MarilynFranklin:
Here are links to the documentation and compose-go PRs:
https://github.com/compose-spec/compose-spec/pull/105
https://github.com/compose-spec/compose-go/pull/76

ndeloof (approved):

technorabilia:
@MarilynFranklin
Thank you for your work on this.
Will this also cover the following scenario or is this another ball game?

```
sh-4.4# echo ${A:-${B:-${C:-z}}}
z
sh-4.4# A=x
sh-4.4# echo ${A:-${B:-${C:-z}}}
x
sh-4.4# B=y
sh-4.4# echo ${A:-${B:-${C:-z}}}
x
sh-4.4# unset A
sh-4.4# echo ${A:-${B:-${C:-z}}}
y
sh-4.4# unset B
sh-4.4# echo ${A:-${B:-${C:-z}}}
z
sh-4.4#
```

tekumara (approved):

glours:

PR #7835, opened 2020-09-28
docker-compose pass arbitrary build flags

### LINKED ISSUE #6358: Support for docker build --secret for build-time docker secrets.

pareshmg:
added a flags option to the service.build section to pass arbitrary arguments to the docker build from the docker-compose file

Resolves #6358  (see https://github.com/docker/compose/issues/6358#issuecomment-443661538)

An example of how I pass aws credentials to my docker build (used for codeartifact in my case)

```version: "3.8"
services:
  worker:
    build:
      context: worker
      flags: '--secret id=aws_creds,src=${HOME}/.aws/credentials'```

### DIFF
--- a/compose/config/config.py
+++ b/compose/config/config.py
@@ -83,6 +83,7 @@ DOCKER_CONFIG_KEYS = [
     'env_file',
     'environment',
     'extra_hosts',
+    'flags',
     'group_add',
     'hostname',
     'healthcheck',
--- a/compose/config/config_schema_compose_spec.json
+++ b/compose/config/config_schema_compose_spec.json
@@ -74,6 +74,7 @@
               "type": "object",
               "properties": {
                 "context": {"type": "string"},
+                "flags": {"type": "string"},
                 "dockerfile": {"type": "string"},
                 "args": {"$ref": "#/definitions/list_or_dict"},
                 "labels": {"$ref": "#/definitions/list_or_dict"},
--- a/compose/service.py
+++ b/compose/service.py
@@ -82,6 +82,7 @@ HOST_CONFIG_KEYS = [
     'dns_opt',
     'env_file',
     'extra_hosts',
+    'flags',
     'group_add',
     'init',
     'ipc',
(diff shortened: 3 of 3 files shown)

### CODE REVIEW

**Comment by pareshmg** [comment]
@ciaranmcnulty : I think this implements your suggestion. 

Also, I'm new to contributing to docker-compose - I'm not quite sure how to write tests for this PR. Is there someone who can help me?

**Review by djmareewish — CHANGES_REQUESTED** [review]
@pareshmg #6510 #4235 
Duplicate of #

**Comment by pareshmg** [comment]
> @pareshmg #6510 #4235
> Duplicate of #

@djmareewish : I don't understand how this is a duplicate of the `--squash` option. This is a more generic method to pass arbitrary flags to docker from docker-compose to help people have a route for a workaround until these parameters become official docker-compose yml syntax options. Or am I misunderstanding what you are pointing out?

**Comment by kiorky** [comment]
In the meantime, i'm just using that to parse build compose args and rely on docker CLI to DRY my builds ...


```sh
# compute the docker-compose service build: args back to docker CLI build-args
buildargs=$(docker-compose config\
    |docker run -i --rm mikefarah/yq yq r - services.${service}.build.args -j\
    |docker run -i --rm imega/jq ".|to_entries[]|\" --build-arg \(.key)=\(.value)\"" -j)
# example: use squash which isnt possible now directly via compose
docker build --squash -t mytag . $buildargs
```

glours:

PR #7837, opened 2020-09-29
Support substitution for set variables

### LINKED ISSUE #7102: Support ${parameter:+word} style variable substitution

MarilynFranklin:
This adds support for ${parameter:+alternate} and ${parameter+alternate}
syntax to allow for substituting with alternate when the parameter is
set.

Signed-off-by: MarilynFranklin <marilyn.j.franklin@gmail.com>

Resolves #7102

### DIFF
--- a/compose/config/interpolation.py
+++ b/compose/config/interpolation.py
@@ -95,7 +95,7 @@ class TemplateWithDefaults(Template):
         """.format(
         delim=re.escape('$'),
         id=r'[_a-z][_a-z0-9]*',
-        bid=r'[_a-z][_a-z0-9]*(?:(?P<sep>:?[-?])[^}]*)?',
+        bid=r'[_a-z][_a-z0-9]*(?:(?P<sep>:?[-?+])[^}]*)?',
     )
 
     @staticmethod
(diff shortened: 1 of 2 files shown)


glours:

PR #7846, opened 2020-10-04
Update run.sh formatting

nemchik:
A few things here.

- Use `${BRACES}` on all variables (good formatting).
- Use `${PWD}` instead of `$(pwd)` to rely on the environment instead of a subshell.
- Allow users to define `DOCKER_COMPOSE_IMAGE_TAG` in their environment to select an image and version to run. Default to `docker/compose:1.27.4` if the user has not set a version. (`latest` seems to be 3 months old as of writing so I've changed this to `docker/compose:1.27.4`). This also allows users to use alternative images or registries such as ghcr.io if their preferred image is available on other registries.
- Uppercase `COMPOSE_DIR` to be in-line with all other variables.
- ~~Fix `shellcheck disable=SC2086` by quoting all variables in https://github.com/docker/compose/compare/master...nemchik:run.sh?expand=1#diff-7123de446b63ba53f37a8e092cafaaaeR67 which then requires adding `eval` to the beginning of the line so that it will be correctly expanded.~~ Reverted this as it caused other issues. It's fixable in bash, but I have not found a posix shell solution.
- Update the script to work with SELinux enforced (mostly RHEL) (from https://github.com/linuxserver/docker-docker-compose/commit/2658740a07816ed118a314c32b12c2a2a1edae25#diff-1b0c2b516b83393edb7200ad5ff12181R65-R69 )

I have tested these changes on Ubuntu 20.04

### DIFF
--- a/script/run/run.sh
+++ b/script/run/run.sh
@@ -69,4 +69,5 @@ if docker info --format '{{json .SecurityOptions}}' 2> /dev/null | grep -q 'name
     DOCKER_RUN_OPTIONS="${DOCKER_RUN_OPTIONS} --privileged"
 fi
 
-eval exec docker run --rm "${DOCKER_RUN_OPTIONS}" "${DOCKER_ADDR}" "${COMPOSE_OPTIONS}" "${VOLUMES}" -w "${PWD}" "${IMAGE}" "$@"
+# shellcheck disable=SC2086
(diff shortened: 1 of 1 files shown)


nemchik:
Unfortunately I had to remove my attempted fix for the shellcheck issue. Using eval prevented the ability to do things like `docker-compose exec ubuntu echo 'Ca$h Money'` since it would evaluate and expand `$h` in that example. I tried a couple other things, but in the end I can't find a way around it.

nemchik:
Updated `DOCKER_COMPOSE_IMAGE_TAG` to allow specifying the image and tag so that users can use alternate images or registries.

glours:

PR #7868, opened 2020-10-16
Add cpu_shares to conversion map

sartyukhov:
Resolves problem, when I tried to use .env variable in cpu_shares, like that:
```
    cpus: ${CPUS}
    cpu_shares: ${CPU_SHARES}
```
Result was:
`TypeError: Invalid type for cpu_shares param: expected int but found <class 'str'>`
But `${CPUS}` works well.


glours:

PR #7869, opened 2020-10-16
Cp command

### LINKED ISSUE #3593: Support `cp' sub command
For now, `docker-compose cp` is missing. :)

```
$ docker-compose -v
```

```
$ docker-compose -h
Define and run multi-container applications with Docker.

Usage:
  docker-compose [-f <arg>...] [options] [COMMAND] [ARGS...]
  docker-compose -h|--help

Options:
  -f, --file FILE             Specify an alternate compose file (default: docker-compose.yml)
  -p, --project-name NAME     Specify an alternate project name (default: directory name)
  -v, --version               Print version and exit
  -H, --host HOST             Daemon socket to connect to

  --tls                       Use TLS; implied by --tlsverify
  --tlscacert CA_PATH         Trust certs signed only by this CA
  --tlscert CLIENT_CERT_PATH  Path to TLS certificate file
  --tlskey TLS_KEY_PATH       Path to TLS key file
  --tlsverify                 Use TLS and verify the remote
  --skip-hostname-check       Don't check the daemon's hostname against the name specified
                              in the client certificate (for example if your docker host
                              is an IP address)

Commands:
  build              Build or rebuild services
  config             Validate and view the compose file
  create             Create services
  down               Stop and remove containers, networks, images, and volumes
  events             Receive real time events from containers
  exec               Execute a command in a running container
  help               Get help on a command
  kill               Kill containers
  logs               View output from containers
  pause              Pause services
  port               Print the public port for a port binding
  ps                 List containers
  pull               Pulls service images
  restart            Restart services
  rm                 Remove stopped containers
  run                Run a one-off command
  scale              Set number of containers for a service
```

alexkiro:
While this is only a basic implementation, after reading through the original issue (#3593), I believe it should handle all the described use cases there. That being said, I would be happy to contribute more to the PR if a more complete implementation is needed. One _(potentially)_ big limitation now is that the command can only work with the CLI enabled. 

_(This is also my first contribution to docker-compose, so I'm sure I missed at least a few obvious things. Quite happy to contribute and fix anything.)_

Resolves #3593

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -351,6 +351,45 @@ class TopLevelCommand:
 
         print(serialize_config(compose_config, image_digests, not options['--no-interpolate']))
 
+    def cp(self, options):
+        """
+        Copy files/folders between a container and the local filesystem
+
+        Usage: cp [options] [--] SERVICE:SRC_PATH DEST_PATH
+               cp [options] [--] SRC_PATH SERVICE:DEST_PATH
+
+        Options:
+            --index=index     index of the container if there are multiple
+                              instances of a service [default: 1]
+        """
+        if self.toplevel_environment.get_boolean("COMPOSE_INTERACTIVE_NO_CLI"):
+            raise UserError("Command cp is only available with CLI")
+
+        src_path = options.get("SERVICE:SRC_PATH")
+        dest_path = options.get("DEST_PATH")
+        if ":" in src_path and ":" in dest_path:
+            raise UserError("Cannot copy between containers")
+
+        from_service = ":" in src_path
+        if from_service:
+            service, src_path = src_path.split(":", 1)
+        else:
+            service, dest_path = dest_path.split(":", 1)
+
+        index = int(options.get("--index"))
+        service = self.project.get_service(service)
+        try:
+            container = service.get_container(number=index)
+        except ValueError as e:
+            raise UserError(str(e))
+
+        args = ["cp"]
+        if from_service:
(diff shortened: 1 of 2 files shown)


bryanpedini:
> One _(potentially)_ big limitation now is that the command can only work with the CLI enabled.

Just out of curiosity, may I ask why? Does it throw any error while being called from an unattended script or a crontab or something like that?

alexkiro:
> > One _(potentially)_ big limitation now is that the command can only work with the CLI enabled.
> 
> Just out of curiosity, may I ask why? Does it throw any error while being called from an unattended script or a crontab or something like that?

After re-reading this, it seems I worded this quite poorly, my bad. By CLI here I mean the [Docker CLI](https://docs.docker.com/compose/reference/envvars/#compose_interactive_no_cli). As far as I can tell docker-compose can either use `docker` via the CMD line by simply opening subprocesses to the executable OR it can use the [Docker API](https://docs.docker.com/engine/api/) in which case the Docker CLI is not necessary. 

With this PR the new `cp` command will only work if the `docker` executable is available, because, well... I just didn't implement a way to use it with the Docker API. :smile:

bryanpedini:
> With this PR the new `cp` command will only work if the `docker` executable is available

Is there anybody who runs Docker without the `docker` command installed? Is there any way to install Docker without getting the `docker` command installed? 😄

JohanMollevik:
What is the status of this? It looks interesting

PR #7872, opened 2020-10-20
Use ssh client by default

aiordache:
By default, shell out to the ssh client for an ssh connection as the paramiko connection is sometimes unreliable.

The `requirements.txt` must be updated once we have a release of docker-py with the ssh client support.

Depends on https://github.com/docker/docker-py/pull/2691

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,7 +4,7 @@ certifi==2020.6.20
 chardet==3.0.4
 colorama==0.4.3; sys_platform == 'win32'
 distro==1.5.0
-docker==4.3.1
+docker==4.4.0
 docker-pycreds==0.4.0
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -4,7 +4,7 @@ certifi==2020.6.20
 chardet==3.0.4
 colorama==0.4.3; sys_platform == 'win32'
 distro==1.5.0
-docker==4.3.1
+git+git://github.com/docker/docker-py@refs/pull/2691/head#egg=docker
```
chris-crone, line 7: I'm guessing this is a draft PR while we wait for the upstream change to be merged?

ulyssessouza (approved):
LGTM

PR #7875, opened 2020-10-20
Use BuildKit and add ability to build CentOS binaries

chris-crone:
Had this in my stash because I noticed we're using `docker cp` to get the binaries out of the images. What I did:
* Moved to using BuildKit so that we can use `docker build --output` instead of `docker container create && docker cp`
* Added build for CentOS (just in the Dockerfile)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,7 +5,7 @@ ARG BUILD_DEBIAN_VERSION=slim-stretch
 ARG RUNTIME_ALPINE_VERSION=3.11.5
 ARG RUNTIME_DEBIAN_VERSION=stretch-20200414-slim
 
-ARG BUILD_PLATFORM=alpine
+ARG DISTRO=alpine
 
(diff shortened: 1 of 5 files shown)


ndeloof (approved):
LGTM

I still miss ability to cp from image, not just containers. Using multi-stage as you did here is a valid workaround, but still a complex/verbose solution to a pretty common CI/CD problem :)

chris-crone:
> I still miss ability to cp from image, not just containers. Using multi-stage as you did here is a valid workaround, but still a complex/verbose solution to a pretty common CI/CD problem :)

I agree that having something more simple to get a single file would be good. Build does have a couple of advantages though: (1) you can define what you want out in the Dockerfile, and (2) you can get the output as a tarball.

ulyssessouza (approved):
LGTM

aiordache (approved):

PR #7926, opened 2020-11-13
Remove duplicate values check for build.cache_from

### LINKED ISSUE #7342: compose crashes on duplicate cache_from, while docker not
Hi,
why is it not possible to use 2 same images for cache_from? Standard docker has no problem with this command `TAG=latest docker build --cache-from=a:$(TAG) --cache-from=a:latest -t a:latest a`.
While when I have this compose file
```yml
version: '3.2'
services:
    a:
        build:
            cache_from:
                - a:${TAG}
                - a:latest
            context: a
            dockerfile: a/Dockerfile
         image: a:${TAG}
```
it gives me error about duplicate items in cache_from.

Can this behavior be changed?
Thanks.

maaarghk:
The `docker` command accepts duplicate values, so there is no benefit to performing this check.

Resolves #7342.

### DIFF
--- a/compose/config/compose_spec.json
+++ b/compose/config/compose_spec.json
@@ -87,7 +87,7 @@
                 "dockerfile": {"type": "string"},
                 "args": {"$ref": "#/definitions/list_or_dict"},
                 "labels": {"$ref": "#/definitions/list_or_dict"},
-                "cache_from": {"$ref": "#/definitions/list_of_strings"},
+                "cache_from": {"type": "array", "items": {"type": "string"}},
                 "network": {"type": "string"},
                 "target": {"type": "string"},
                 "shm_size": {"type": ["integer", "string"]},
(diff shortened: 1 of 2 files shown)


compose/config/compose_spec.json
```diff
@@ -87,7 +87,7 @@
                 "dockerfile": {"type": "string"},
                 "args": {"$ref": "#/definitions/list_or_dict"},
                 "labels": {"$ref": "#/definitions/list_or_dict"},
-                "cache_from": {"$ref": "#/definitions/list_of_strings"},
+                "cache_from": {"type": "array", "items": {"type": "string"}},
```
aiordache, line 90: We download this file from the compose specification repo. Please propose this change there first: https://github.com/compose-spec/compose-spec/blob/68a799e866c2a8ff5d553c3797564282347d2659/schema/compose-spec.json#L90

maaarghk: Thanks, the change has now been made upstream so this branch can be merged.

aiordache (approved):
LGTM. Thanks!

Issue #7539, opened 2020-06-17
Proposal: One-off tasks


bfirsh:
## Background & problem

Users quite often define maintenance scripts, test suites, debugging systems in their Compose files which they don't want to run when they do `docker-compose up`.

Personally, this is something I've wanted since almost the start. My daily paper cut is writing `docker-compose run web ./test.sh` instead of `docker-compose run test`.

I wrote a diplomatic aggregation of potential solutions and data points in #1896. This is an opinionated proposed solution we can use as a strawman. 

## Proposed solution

I propose we add a top-level `tasks` configuration option, which takes the same options as the `services` option. It's a bit like a service, but doesn't start when you run `docker-compose up`. It only works with `docker-compose run`. Service and task names must be unique so they don't collide.

As an example, let's add a task to the [Django getting started guide](https://docs.docker.com/compose/django/) app:

```yaml
services:
  db:
    image: postgres
    environment:
      - POSTGRES_DB=postgres
      - POSTGRES_USER=postgres
      - POSTGRES_PASSWORD=postgres
  web:
    build: .
    command: python manage.py runserver 0.0.0.0:8000
    volumes:
      - .:/code
    ports:
      - "8000:8000"
    depends_on:
      - db
tasks:
  test:
    build: .
    entrypoint: python manage.py test
    depends_on:
     - db
```

You would run the tests by running `docker-compose run test`.

Because we've specified `entrypoint` in the task definition, we can also pass arguments to the tests: `docker-compose run test myapp/test_views.py`

## Questions

- What am I not thinking about?
- `tasks` is a nice bike shed -- go at it
- How does this interact with Compose spec?
- Having to specify `entrypoint` instead of `command` is a bit clumsy. Users will automatically write `command` then wonder why their arguments aren't working as expected. Maybe there is a better solution here?

/cc @aanand @shin- in case there is any historical knowledge I am not aware of. (hi!)

ndeloof:
I like the idea compose file format can be used to define tasks/jobs that are expected to stop after completion, but imho this is orthogonal to the initial intent to only run some of them in a specific context. 

For this purpose, I'd prefer a "profile" approach like I used to do with Apache Maven on java projects: Just like docker-compose by default will load a docker-compose-override.yaml file if it exists aside the main yaml config, we can introduce support for `--profile` flag that will automatically pick an override file to be loaded.

i.e
`docker-compose up` => `docker-compose.yaml` + `docker-compose-override.yaml`
[...]

PR #7930, opened 2020-11-15
Implement service profiles

### LINKED ISSUE #7919: Implement service profiles in Compose
Now that [service profiles are part of the Compose spec](https://github.com/compose-spec/compose-spec/pull/110), we should implement them here.

acran:
Implement profiles as introduced in compose-spec/compose-spec#110

This is a first draft and probably needs some more discussion. Things I'm not quite sure about yet which may need further discussion:
## How to specify profiles to enable
* by parameter: I implemented the `--profile` flag for the _main_ command, multiple profiles may be passed
  * use another flag name?
  * move the flag to the affected subcommands only (see below)?
* by environment: I used `COMPOSE_PROFILE` with `:` as separator for multiple profiles
  * use another variable name?
  * use another separator(s)?
  * make the separator configurable?
* merging profiles: to be consistent with how other array options are handled already the `profiles` of a service from multiple configuration files are just added together
  * there is no way to remove profiles from a service or unset the `profiles` option all together with override configurations. Should there be?

## Affected commands
The main logic is implemented in `Project.get_services()` which was already used in the code paths for and therefor add profiles support to:
* `build`
* `create`
* `up`
* `start`
* `pull`
* `push`

The following functions use `Project.containers()` instead which is unchanged and therefor **ignore** profiles:
* `stop`
* `down`
  * using `remove_stopped` which already used `Project.containers()`
  * using `remove_images` which I changed to ignore profiles to be consistent with the rest
* `pause`
* `unpause`
* `kill`

I couldn't quite decide if the latter commands should be consider profiles, too, because the behavior expected by the user would most probably not consider them, most notably with `docker-compose down`:

Executing `docker-compose down` the user expects that all defined services will be removed. But services assigned to specific profiles only could be missed here when `docker-compose down` is called without enabling those profiles.
Then those services would need to be considered _orphans_. But this would be inconsistent with notion of _orphans_ since those services _are_ actually defined in the configuration, just not enabled.
On the other hand `docker-compose --profile foo down` might suggest that services with the `foo` profile only would be removed, but actually the spec would require _all services without profiles and services with the `foo` profile_.

That's why I thought it to be most consistent/intuitive for `docker-compose down` and the other commands to just ignore profiles and act on all defined services (and because changing the logic of `Project.containers()` would require the refactoring/verification of a whole lot more code paths and tests).

## Tests
Using the example `docker-compose.yml` from compose-spec/compose-spec#110 I implemented a test case for each of the examples in the spec as well as an extra test case for merging profiles from multiple configuration files.
Are there more tests needed here?

_Small disclaimer_: Since I'm neither familiar with the code base nor python even, please don't hold back with any feedback on the coding style/formatting/performance :)

Resolves #7919, #1896, #6742, #7539

### DIFF
--- a/compose/cli/command.py
+++ b/compose/cli/command.py
@@ -66,7 +66,8 @@ def project_from_options(project_dir, options, additional_options=None):
         environment=environment,
         override_dir=override_dir,
         interpolate=(not additional_options.get('--no-interpolate')),
-        environment_file=environment_file
+        environment_file=environment_file,
+        enabled_profiles=get_profiles_from_options(options, environment)
     )
 
 
@@ -115,9 +116,21 @@ def get_config_path_from_options(base_dir, options, environment):
     return None
 
 
+def get_profiles_from_options(options, environment):
+    profile_option = options.get('--profile')
+    if profile_option:
+        return profile_option
+
+    profiles = environment.get('COMPOSE_PROFILE')
+    if profiles:
+        return profiles.split(',')
+
+    return []
+
+
 def get_project(project_dir, config_path=None, project_name=None, verbose=False,
                 context=None, environment=None, override_dir=None,
-                interpolate=True, environment_file=None):
+                interpolate=True, environment_file=None, enabled_profiles=None):
     if not environment:
         environment = Environment.from_env_file(project_dir)
     config_details = config.find(project_dir, config_path, environment, override_dir)
@@ -139,6 +152,7 @@ def get_project(project_dir, config_path=None, project_name=None, verbose=False,
             client,
             environment.get('DOCKER_DEFAULT_PLATFORM'),
             execution_context_labels(config_details, environment_file),
+            enabled_profiles,
         )
 
 
--- a/compose/project.py
+++ b/compose/project.py
@@ -68,13 +68,15 @@ class Project:
     """
     A collection of services.
     """
-    def __init__(self, name, services, client, networks=None, volumes=None, config_version=None):
+    def __init__(self, name, services, client, networks=None, volumes=None, config_version=None,
+                 enabled_profiles=None):
         self.name = name
         self.services = services
         self.client = client
         self.volumes = volumes or ProjectVolumes({})
         self.networks = networks or ProjectNetworks({}, False)
         self.config_version = config_version
+        self.enabled_profiles = enabled_profiles or []
 
     def labels(self, one_off=OneOffFilter.exclude, legacy=False):
         name = self.name
@@ -86,7 +88,8 @@ class Project:
         return labels
 
     @classmethod
-    def from_config(cls, name, config_data, client, default_platform=None, extra_labels=None):
+    def from_config(cls, name, config_data, client, default_platform=None, extra_labels=None,
+                    enabled_profiles=None):
         """
         Construct a Project from a config.Config object.
         """
@@ -98,7 +101,7 @@ class Project:
             networks,
             use_networking)
         volumes = ProjectVolumes.from_config(name, config_data, client)
-        project = cls(name, [], client, project_networks, volumes, config_data.version)
+        project = cls(name, [], client, project_networks, volumes, config_data.version, enabled_profiles)
 
         for service_dict in config_data.services:
             service_dict = dict(service_dict)
@@ -186,7 +189,7 @@ class Project:
             if name not in valid_names:
                 raise NoSuchService(name)
 
-    def get_services(self, service_names=None, include_deps=False):
+    def get_services(self, service_names=None, include_deps=False, auto_enable_profiles=True):
         """
         Returns a list of this project's services filtered
         by the provided list of names, or all services if service_names is None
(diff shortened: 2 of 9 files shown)


compose/cli/command.py
```diff
@@ -115,9 +116,21 @@ def unicode_paths(paths):
...
 
 
+def get_profiles_from_options(options, environment):
+    profile_option = options.get('--profile')
+    if profile_option:
+        return profile_option
+
+    profiles = environment.get('COMPOSE_PROFILE')
+    if profiles:
+        return profiles.split(':')
```
acran, line 126: `:` as a separator is hard coded here
* should it be something else?
* should it be configurable?

compose/project.py
```diff
@@ -830,14 +860,29 @@ def _find():
...
+                    e for e in dep.options.get('profiles')
+                    if e in enabled_profiles
+                ]:
+                    raise ConfigurationError(
+                        'Service "{dep_name}" was pulled in as a dependency of '
+                        'service "{service_name}" but is not enabled by the '
+                        'active profiles. '
+                        'You may fix this by adding a common profile to '
+                        '"{dep_name}" and "{service_name}".'
+                        .format(dep_name=dep.name, service_name=service.name)
```
acran, line 884: Is this a suitable error message making it clear to the user what went wrong, why and how to fix it?

compose/project.py
```diff
@@ -412,10 +440,12 @@ def down(
         self.remove_images(remove_image_type)
 
     def remove_images(self, remove_image_type):
-        for service in self.get_services():
+        for service in self.services:
```
acran, line 443: The call to `self.get_services` was actually unnecessary before, since without any parameters it would just return `self.services`.
But now it would filter services by enabled profiles, so this was changed so `remove_images` and thus `down` would act on _all_ defined services.

compose/project.py
```diff
@@ -412,10 +440,12 @@ def down(
...
             service.remove_image(remove_image_type)
 
     def restart(self, service_names=None, **options):
+        # filter service_names by enabled profiles
+        service_names = [s.name for s in self.get_services(service_names)]
```
acran, line 448: `restart` should consider the enabled profiles so to not start services disabled by profile with `docker-compose restart`

acran (commented):
I comment on some code lines which probably require some attention

compose/cli/command.py
rumpl: Curious why wouldn't it be simply `,`?

acran: No specific reason for `:` here, I just oriented on the separator in [`COMPOSE_FILE`](https://docs.docker.com/compose/reference/envvars/#compose_file) (at least on Linux/macOS) but it could be `,`, whitespace or any of it equally well.

The [spec only defines the regex `[a-zA-Z0-9][a-zA-Z0-9_.-]+`](https://github.com/compose-spec/compose-spec/commit/36e14e20259425605016cabec75a6a28606d8dc8#diff-bc6661da34ecae62fbe724bb93fd69b91a7f81143f2683a81163231de7e3b545R1525) for valid profile names. So anything else could be used as separator.

aiordache: I would prefer `,` as separator, to make it similar to a list in python.

acran: Ok, then: I updated the PR to use a hard coded `,` as delimiter here

aiordache:
@acran Can you rebase please?

acran:
@aiordache yes, done

aiordache (approved):
Thank you! SGTM

ulyssessouza (changes requested):
LGTM in general, just missing changes

compose/project.py
```diff
@@ -198,15 +201,42 @@ def get_services(self, service_names=None, include_deps=False):
...
+            auto_enable_profiles = False
+            service_names = [
+                service.name
+                for service in self.services
+                # only include services without profiles or in enabled profiles
+                if 'profiles' not in service.options or [
+                    e for e in service.options.get('profiles')
+                    if e in enabled_profiles
+                ]
+            ]
```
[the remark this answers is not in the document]
acran: oh yes, there seems to be indentation missing/wrong.
How would I make this more _verbose_? extract into a separate function?

compose/cli/command.py
```diff
@@ -115,9 +116,21 @@ def unicode_paths(paths):
...
 
 
+def get_profiles_from_options(options, environment):
+    profile_option = options.get('--profile')
+    if profile_option:
+        return profile_option
+
+    profiles = environment.get('COMPOSE_PROFILE')
+    if profiles:
+        return profiles.split(',')
```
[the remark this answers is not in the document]
acran: Since the profile names are not paths or file names and there should be no difference in specifying profiles on different OS I feel this to be semantically incorrect.
But since I have no strong opinion on this I can implement this if it is the consensus.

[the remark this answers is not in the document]
aiordache: From my understanding, a profile is a service label that we use to filter services, not a file path. I wouldn't over-complicate things by enabling a custom separator. I think going with `,` should suffice.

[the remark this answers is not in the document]
ulyssessouza: OK. I agree with you both :smiley: 
My thoughts were more around the idea of using an homogeneous separator. 
Comma is fine for me.

compose/cli/command.py
```diff
@@ -115,9 +116,21 @@ def unicode_paths(paths):
...
+    if profiles:
+        return profiles.split(':')
+
+    return []
+
+
 def get_project(project_dir, config_path=None, project_name=None, verbose=False,
                 context=None, environment=None, override_dir=None,
-                interpolate=True, environment_file=None):
+                interpolate=True, environment_file=None, enabled_profiles=[]):
```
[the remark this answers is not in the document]
acran: Thanks for the link / the background on this; this was indeed unexpected behavior for me.

I updated the default values to `None`, but omitted the assignments to an empty array where the parameter wasn't actually used but only passed around.

compose/project.py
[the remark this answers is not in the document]
acran: I now moved this to [`service.enabled_for_profiles(enabled_profiles)`](https://github.com/docker/compose/commit/a5e5318f8c1eb9b5e3a7cafb154af8ce6b9aee20#diff-c0385cfc998f4ccdf336127a3d7cc3c934fa3067c8e04e67dac7da47846da6f0R1340-R1350) which should make it clearer and also re-used this in `_inject_deps`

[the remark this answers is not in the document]
acran: also, the indentation actually was correct - at least syntactically; and regarding the coding style matching with other list comprehensions already in the code.

acran:
I updated the PR again with the feedback from @ulyssessouza. For this I did a little refactoring and created [`service.enabled_for_profiles()`](https://github.com/docker/compose/commit/a5e5318f8c1eb9b5e3a7cafb154af8ce6b9aee20#diff-c0385cfc998f4ccdf336127a3d7cc3c934fa3067c8e04e67dac7da47846da6f0R1340-R1350) to make the code a bit more readable.

ulyssessouza (approved):
LGTM

Thanks @acran !

Issue #7931, opened 2020-11-16
Docker configs in non-swarm mode failing


jeffrson:
## Description of the issue

As discussed in https://github.com/docker/compose/issues/6338 docker-compose did output a warning when using configs in non-swarm mode.
However, somewhen in between 1.27.0-rc1 and 1.27.0-rc2 this warning disappeared.

## Context information (for bug reports)

**Output of `docker-compose version`**
```
1.27.4
```

**Output of `docker version`**
```
 19.03.13
```

PR #7946, opened 2020-11-25
Bring back warning for configs in non-swarm mode

aiordache:
Fixes https://github.com/docker/compose/issues/7931
```yaml
services:
  backend:
    image: nginx
    configs:
      - source: bar
configs:
  bar:
    external: true
    name: bar
```

```
$ docker-compose up
WARNING: Some services (backend) use the 'configs' key, which will be ignored. Compose does not support 'configs' configuration - use `docker stack deploy` to deploy to a swarm.
Creating test_backend_1 ... done
....

```

### DIFF
--- a/compose/config/config.py
+++ b/compose/config/config.py
@@ -373,6 +373,23 @@ def find_candidates_in_parent_dirs(filenames, path):
     return (candidates, path)
 
 
+def check_swarm_only_config(service_dicts):
+    warning_template = (
+        "Some services ({services}) use the '{key}' key, which will be ignored. "
+        "Compose does not support '{key}' configuration - use "
+        "`docker stack deploy` to deploy to a swarm."
+    )
+    key = 'configs'
+    services = [s for s in service_dicts if s.get(key)]
+    if services:
+        log.warning(
+            warning_template.format(
+                services=", ".join(sorted(s['name'] for s in services)),
+                key=key
+            )
+        )
+
+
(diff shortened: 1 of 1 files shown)


compose/config/config.py
```diff
@@ -373,6 +373,25 @@ def find_candidates_in_parent_dirs(filenames, path):
...
+    def check_swarm_only_key(service_dicts, key):
+        services = [s for s in service_dicts if s.get(key)]
+        if services:
+            log.warning(
+                warning_template.format(
+                    services=", ".join(sorted(s['name'] for s in services)),
+                    key=key
+                )
+            )
+    check_swarm_only_key(service_dicts, 'configs')
```
ulyssessouza, line 392: nit: IMHO the internal function only adds complexity to `check_swarm_only_key`. This could be "inlined"

ulyssessouza (approved):
LGTM

thaJeztah:
Weren't configs and secrets (at least partially) implemented using bind-mounts in https://github.com/docker/compose/pull/4896 ?

ulyssessouza:
@thaJeztah IIRC we only support `secrets` by now. At least we tried to run it with an older version and it does not support too.
Maybe that's just the implementation of the parsing, but not the full feature with bind-mounts.

Anyways, the feature is not there by now. We can implement that and remove the message in a follow up.

Issue #7945, opened 2020-11-25
Apple Silicon builds


chris-crone:
I've started looking at Apple Arm support and this is what I've found so far:
- [x] OpenSSL appears to build
- [x] Python will provide support [soon](https://github.com/pypa/packaging/pull/319#issuecomment-733290582)
- [x] Homebrew doesn't work on Arm
- [x] PyNaCl doesn't appear to build for macOS 11 (on Intel, unable to test on Arm)
- [ ] PyInstaller doesn't generate Arm binaries

Working towards Docker Desktop support on Apple Arm: https://github.com/docker/roadmap/issues/142

It appears that Rosetta allows the Intel version to launch:
```console
$ arch
arm64
$ ./docker-compose --version
docker-compose version 1.27.4, build 40524192
```

Ongoing work in https://github.com/docker/compose/pull/7951

PR #7951, opened 2020-11-26
[DO NOT MERGE] PoC macOS Arm build support

chris-crone:
* Uses local macOS Python to bootstrap instead of downloading Python from Homebrew
* Replaces GNU `realpath` with a bash script
* Patch OpenSSL to work on macOS Arm
* Do not use OpenSSL assembler optimization
* Bump to Python 3.9.1rc1 (which has Arm support)
* Fix typo in `LDFLAGS` (fixes PyNaCl build issue)

Contributes to #7945

### DIFF
--- a/openssl.patch
+++ b/openssl.patch
@@ -1,0 +1,19 @@
+diff --git a/Configurations/10-main.conf b/Configurations/10-main.conf
+index eb92c24..eaf2ab6 100644
+--- a/Configurations/10-main.conf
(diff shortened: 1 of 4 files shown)


justincormack (changes requested):
Not to be shipped without an official openssl release for darwin/arm64

chris-crone:
I've adjusted the title to make it more clear that this is discovery work for making `docker-compose` build on macOS Arm.

PR #7944, opened 2020-11-25
Bump dependencies

chris-crone:
* Python 3.7 to 3.9
* Linux: bump Debian base image from stretch to buster (required for Python 3.9)
* macOS: OpenSSL 1.1.1g to 1.1.1h, Python 3.7.7 to 3.9.0
* pyinstaller 3.6 to 4.1
* Loosen restriction on base images to latest minor

PR #7959, opened 2020-11-30
Bump Python version for release

chris-crone:
* Bumped Windows Python version in Jenkins file
* Reversed PATH order so that Jenkinsfile can be used to set Python version

Should've been included in #7944 but was missed.


StefanScherer (approved):
LGTM 
Makes sense!

aiordache (approved):

PR #7960, opened 2020-11-30
Simplify short version variable for CentOS in Dockerfile

chris-crone:
We need major.minor for configuring the default Python version. This is simpler than cut and handles minor releases after .9.


aiordache (approved):

ulyssessouza (approved):
LGTM

PR #7978, opened 2020-12-07
Make COMPOSE_DOCKER_CLI_BUILD=1 the default

thaJeztah:
This changes compose to use "native" build through the CLI by default. With this, docker-compose can take advantage of BuildKit (which is now enabled by default on Docker Desktop 2.5 and up).

Users that want to use the python client for building can opt-out of this feature by setting COMPOSE_DOCKER_CLI_BUILD=0

### DIFF
--- a/compose/project.py
+++ b/compose/project.py
@@ -489,7 +489,8 @@ class Project:
                 log.info('%s uses an image, skipping' % service.name)
 
         if cli:
-            log.warning("Native build is an experimental feature and could change at any time")
+            log.info("Building with native build. Learn about native build in Compose here: "
+                     "https://docs.docker.com/go/compose-native-build/")
             if parallel_build:
                 log.warning("Flag '--parallel' is ignored when building with "
                             "COMPOSE_DOCKER_CLI_BUILD=1")
@@ -649,7 +650,8 @@ class Project:
            ):
 
         if cli:
-            log.warning("Native build is an experimental feature and could change at any time")
+            log.info("Building with native build. Learn about native build in Compose here: "
+                     "https://docs.docker.com/go/compose-native-build/")
 
         self.initialize()
         if not ignore_orphans:
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -272,8 +272,6 @@ class TopLevelCommand:
             --no-rm                 Do not remove intermediate containers after a successful build.
             --parallel              Build images in parallel.
             --progress string       Set type of progress output (auto, plain, tty).
-                                    EXPERIMENTAL flag for native builder.
-                                    To enable, run with COMPOSE_DOCKER_CLI_BUILD=1)
             --pull                  Always attempt to pull a newer version of the image.
             -q, --quiet             Don't print anything to STDOUT
         """
@@ -287,7 +285,7 @@ class TopLevelCommand:
                 )
             build_args = resolve_build_args(build_args, self.toplevel_environment)
 
-        native_builder = self.toplevel_environment.get_boolean('COMPOSE_DOCKER_CLI_BUILD')
+        native_builder = self.toplevel_environment.get_boolean('COMPOSE_DOCKER_CLI_BUILD', True)
 
         self.project.build(
             service_names=options['SERVICE'],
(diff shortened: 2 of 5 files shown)


ulyssessouza (approved):
LGTM

thaJeztah:
Couple of failures; looks like some tests are using the build-log output of the classic builder

```
 =================================== FAILURES ===================================
 _______________________ CLITestCase.test_build_log_level _______________________
 tests/acceptance/cli_test.py:795: in test_build_log_level
     assert result.stderr == ''
 E   AssertionError: assert "The command '/bin/sh -c exit 1' returned a non-zero code: 1\n" == ''
 E     + The command '/bin/sh -c exit 1' returned a non-zero code: 1
 ----------------------------- Captured stdout call -----------------------------
 Running process: 34
 Running process: 42
 Running process: 50
 _________________ CLITestCase.test_build_shm_size_build_option _________________
 tests/acceptance/cli_test.py:852: in test_build_shm_size_build_option
     assert 'shm_size: 96' in result.stdout
 E   AssertionError: assert 'shm_size: 96' in 'Sending build context to Docker daemon  3.072kB\n\nStep 1/2 : FROM busybox\nlatest: Pulling from library/busybox\nea9...69a0\n ---> 48c73be27d98\nSuccessfully built 48c73be27d98\nSuccessfully tagged build-shm-size_custom_shm_size:latest\n'
 E    +  where 'Sending build context to Docker daemon  3.072kB\n\nStep 1/2 : FROM busybox\nlatest: Pulling from library/busybox\nea9...69a0\n ---> 48c73be27d98\nSuccessfully built 48c73be27d98\nSuccessfully tagged build-shm-size_custom_shm_size:latest\n' = ProcessResult(stdout='Sending build context to Docker daemon  3.072kB\n\nStep 1/2 : FROM busybox\nlatest: Pulling from...:latest\n', stderr='Building with native build; set COMPOSE_DOCKER_CLI_BUILD=0 to disable\nBuilding custom_shm_size\n').stdout
 ----------------------------- Captured stdout call -----------------------------
 Running process: 165
 ------------------------------ Captured log call -------------------------------
 WARNING  compose.network:network.py:289 Network build-shm-size_default not found.
 ________________ CLITestCase.test_run_up_with_docker_cli_build _________________
 tests/unit/cli_test.py:272: in test_run_up_with_docker_cli_build
     assert call_kwargs.get('cli')
 E   AssertionError: assert False
 E    +  where False = <built-in method get of dict object at 0x7f20564de140>('cli')
 E    +    where <built-in method get of dict object at 0x7f20564de140> = {'cli': False, 'detached': True, 'one_off': True, 'override_options': {'command': None, 'detach': True, 'ports': [], 'restart': None, ...}, ...}.get
 ----------------------------- Captured stdout call -----------------------------
```

compose/project.py
```diff
@@ -649,7 +649,7 @@ def up(self,
            ):
 
         if cli:
-            log.warning("Native build is an experimental feature and could change at any time")
+            log.info("Building with native build; set COMPOSE_DOCKER_CLI_BUILD=0 to disable")
```
rumpl, line 652: Reading this quickly and it seems as if compose is telling me I **should** disable native build. Could we turn it around a little bit and make it more explicit that they don't have to if they don't want?

rumpl (approved):
After the message change it should be good to merge 👍

compose/project.py
thaJeztah: Good one. My intent was to make it easy to discover how to switch, but (I guess) env-vars being UPPERCASE makes it less subtle and feel more like "you should do this".

Should we include a URL with details on how to switch?

```
Building with native build. Learn about native build in compose here: https://docs.docker.com/go/compose-native-build
```

chris-crone: +1 to adding a link to the docs!

One nit: `Building with native build. Learn about native build in Compose here`

thaJeztah:
Updated; also opened https://github.com/docker/docker.github.io/pull/11858

compose/project.py
nebuk89: I agree! Let's make sure we use re-directs again or something here so we can track the number of people this is impacting :)/who are using the link from the CLI

thaJeztah: I updated the message; https://github.com/docker/compose/pull/7978/files and a PR for the documentation is opened in https://github.com/docker/docker.github.io/pull/11858, but likely would have to wait being merged until / just before this patch is released and included in Docker Desktop (as it mentions the new default)

PR #7979, opened 2020-12-07
Use official docker:dind image instead of custom image

thaJeztah:
Similar to https://github.com/docker/docker-py/pull/2515

- replace the dockerswarm/dind image with the official docker image
- update versions of docker to test against (19.03.14 and 20.10.0-rc2)

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,6 +1,6 @@
 #!groovy
 
-def dockerVersions = ['19.03.13']
+def dockerVersions = ['20.10.0-rc2', '19.03.14']
 def baseImages = ['alpine', 'debian']
 def pythonVersions = ['py39']
 
--- a/script/test/all
+++ b/script/test/all
@@ -39,15 +39,13 @@ for version in $DOCKER_VERSIONS; do
 
   trap "on_exit" EXIT
 
-  repo="dockerswarm/dind"
-
   docker run \
     -d \
     --name "$daemon_container" \
     --privileged \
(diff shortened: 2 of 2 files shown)


script/test/all
```diff
@@ -39,15 +39,13 @@ for version in $DOCKER_VERSIONS; do
...
-
   docker run \
     -d \
     --name "$daemon_container" \
     --privileged \
     --volume="/var/lib/docker" \
-    "$repo:$version" \
-    dockerd -H tcp://0.0.0.0:2375 $DOCKER_DAEMON_ARGS \
+    "docker:${version}-dind" \
+    sh -c "apk add --no-cache git && dockerd -H tcp://0.0.0.0:2375 $DOCKER_DAEMON_ARGS" \
```
thaJeztah, line 48: This is a bit ugly; there are variants with `git` installed, but not for the `dind` variant. Looks like it's only needed for a single test 😞

Jenkinsfile
```diff
@@ -1,6 +1,6 @@
 #!groovy
 
-def dockerVersions = ['19.03.13']
+def dockerVersions = ['20.10.0-rc2', '19.03.14']
```
ulyssessouza, line 3: We typically don't use RCs here

thaJeztah: GA should now be available, so I can update. That said; I think it makes sense to test against upcoming releases (once we reach RC's) to make sure they work together.

PR #7989, opened 2020-12-10
Add metrics capturing

ulyssessouza:
Resolves https://github.com/docker/dev-tooling-team/issues/238

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -26,6 +26,7 @@ from ..config.serialize import serialize_config
 from ..config.types import VolumeSpec
 from ..const import IS_WINDOWS_PLATFORM
 from ..errors import StreamParseError
+from ..metrics.decorator import metrics
 from ..progress_stream import StreamOutputError
 from ..project import get_image_digests
 from ..project import MissingDigests
@@ -53,6 +54,8 @@ from .log_printer import LogPrinter
 from .utils import get_version_info
 from .utils import human_readable_file_size
 from .utils import yesno
+from compose.metrics.client import MetricsCommand
+from compose.metrics.client import Status
 
 
 if not IS_WINDOWS_PLATFORM:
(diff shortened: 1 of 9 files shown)


compose/cli/main.py
```diff
@@ -61,36 +64,77 @@
...
+        return ['version']
+
+
+def exit_with_metrics(command, log_msg=None, status=Status.SUCCESS, exit_code=1):
+    if log_msg:
+        if not exit_code:
+            log.info(log_msg)
+        else:
+            log.info(log_msg)
+
```
ndeloof, line 135: same log invocation in both cases

ndeloof (approved):
:+1:

compose/metrics/client.py
```diff
@@ -0,0 +1,48 @@
...
+    SUCCESS = "success"
+    FAILURE = "failure"
+    CANCELED = "canceled"
+
+
+class MetricsSource:
+    CLI = "docker-compose"
+
+
+METRICS_PORT = 9000
```
gtardif, line 17: We need to run over sockets, rather than ports, and named pipes on windows. 
Desktop has 2 defaults :
* `\\.\pipe\docker_cli`  on windows
* `/var/run/docker-cli.sock` on mac

e2e tests can have another socket in a test tmp folder, so we ensure this works over sockets rather than ports. Will need to validate on windows how this works

gtardif (changes requested):
LGTM, but needs to run over sockets / named pipes and include defaults from Desktop

ulyssessouza:
PTAL @gtardif 

Also need manual validation on MacOS and Windows

compose/cli/main.py
ulyssessouza: Thanks!

compose/metrics/client.py
ulyssessouza: Done. Thanks!

gtardif:
validated with Desktop mac & windows

gtardif (approved):

casdevs:
@ulyssessouza: is there some documentation available what this PR is about?
the given issue link https://github.com/docker/dev-tooling-team/issues/238 does not work...

chris-crone:
@casdevs The goal of this is us to collect usage data so that we can make Docker better. The code sends CLI events to Docker Desktop which then aggregates CLI daily usage (i.e.: x command was used y times). This data is only sent if you've opted in with Docker Desktop.

This data will help us, for example, implement Compose Go (see: https://github.com/docker/compose-cli) by giving us data about which commands are used more and thus should be higher priority.

casdevs:
@chris-crone: thanks for the clarification!
As some people may be sensitive about such data being sent out to external parties, maybe it would be a good idea to document the motiviation and options (opt-in etc.) about this somehwere. Or did I miss that anywhere?
Even with it being an opt-in, maybe some details about it can motivate people to enable it, because just writing `Added usage metrics` in the release notes may unnecessarily raise some data privacy / compliance questions...

chris-crone:
@casdevs I totally get that, I'm a developer too :)

We have a draft document internally that outlines what we're doing for telemetry but it needs a little more work before it's ready to share publicly. I'll see what I can do to get this sped up.

I think having a global statement will be more useful than answering this question on multiple repos.

casdevs:
@chris-crone great, thanks!

PR #7994, opened 2020-12-14
Add env var for base config file

laundmo:
The aim of this simple change is to allow a global base file to be configured. I made this for personal use but maybe there's interest in a proper PR in which case I would need help in adding tests.

Use case:
I personally have many docker-compose files which I would like to keep D.R.Y.
For this, the `COMPOSE_BASE_FILE` env var can be set to the path of a YAML file that contains global Extension fields which will then become available in all other compose files.

Signed-off-by: Laurin Schmidt <laurinschmidt2001@gmail.com>


laundmo:
it seems ive made a mistake in testing, and the Extension fields are not inherited.

PR #7997, opened 2020-12-15
Add support for SSH mounts by passing artbitrary arguments to the builder

### LINKED ISSUE #8015: Add build:cli_args to spec to remove strict coupling with CLI builds
Features and options are added to the CLI for `docker build` and then support must be manually added to `docker-compose` to allow configuring those new features and options. This minimally leads to some delay, but can also lead to very long delays; in addition some items might never make it into the compose spec since they are not popular enough, or don't have clear compatibility with the docker client build mode. As a concrete example `docker build` via buildkit has supported the `--ssh` flag for a while, which is extremely useful for anyone using private repositories for go, python, node, etc. packages. However, support for that has yet to make it into `docker-compose` multiple years later. See https://github.com/docker/compose/issues/7025 and the various other issues linked or related to that to get an appreciation for how much pain that causes and how long it has been an issue.

While it is nice to add first-class support for new features, especially ones that can be supported on docker client builds, those will always suffer from  playing catch up with `docker build`. Instead it seems like it would be very useful to have an option in the compose file to pass in arbitrary CLI options for `docker build` so a person could take advantage of new functionality as soon as it is added. The exact names and schema of that configuration seem like they could be tweaked, but I made a prototype implementation to prove to myself that it would work:
https://github.com/docker/compose/compare/master...TheJumpCloud:cli-args

This isn't handling non-CLI builds yet as I'm not sure the best way to split that given how the client vs. CLI build is handled as essentially an interface: https://github.com/docker/compose/blob/master/compose/service.py#L1106. However, as it is now I can simply add this snippet to a docker-compose file and solve the litany of woes people have had around build-time secrets, including `--ssh default`:
```
build:
  cli_args:
    - --ssh
    - default
```

This avoids, or perhaps rather allows, debate about how to add in first-class configuration for new options to progress at an appropriate speed, but still let's uses adopt the new features when they are read. I think this pattern would mean that anyone could adopt _any_ new `docker build` functionality on day 0 going forward.

There are various PRs that try to solve for a particular configuration shortcoming; for example the SSH issue already referenced has various approaches in  https://github.com/docker/compose/pull/7997, https://github.com/docker/compose/pull/7046 and https://github.com/docker/compose/pull/7296. I don't think this proposal is mutually exclusive from any of those: the compose spec can still have well defined configuration for things that are broadly adopted, but `build:cli_args` can act as an early adoption as well as escape valve for those that aren't useful or standardized enough to be promoted to a first class `build` configuration option, or are taking a while to get there.

**Additional context**
In terms of the impact, I'm not sure what other people have done, but we've essentially been forced to introduce a pre-build step that runs `docker build` directly with options such as `--ssh default` and then run `docker-compose build` after that. It breaks and confuses a bunch of stuff like multi-staged builds, as well as introduces a custom pattern outside the normal docker/docker-compose workflows. With something like `build:cli_args` we could have decided as an org to move to buildkit + that SSH option whenever we wanted, instead of having to construct this brittle workaround.

zoonage:
To solve #7025 this takes the idea from #8015 and allows passing arbitrary arguments through to the CLI builder.

It does this by adding `extra_cli_build_args` to the parameters available in the `build` section of a `service`.

As opposed to directly testing that `docker build --ssh default ....` works the tests confirm that we can pass arbitrary flags through to the builder.

Signed-off-by: Aaron George <aaron@ometria.com>

Resolves #8015 #7025


zoonage:
Run into a roadblock, I suspect that py-docker needs updating to add buildkit arguments. But this is only if we are using the Python client, in which scenarios is the CLI builder used vs the py-docker client?

evandam:
Any updates on this? It looks like a great approach to me with minimal risk, especially considering `COMPOSE_DOCKER_CLI_BUILD` is set by default now and keeps parity with `docker build`.

zoonage:
> Any updates on this? It looks like a great approach to me with minimal risk, especially considering `COMPOSE_DOCKER_CLI_BUILD` is set by default now and keeps parity with `docker build`.

Sorry I've been super busy lately, hopefully I'll get round to looking at this in the next few days

zoonage:
I've changed the way this is done based on what I was reading in #8015. Allowing arbitrary CLI arguments to be passed through means that when new arguments are added compose can "support" them immediately by using this work around.

This has the upside of easily being testable unlike the SSH stuff which would need SSH keys, potentially private repositories, and possibly other infrastructure.

compose/config/compose_spec.json
```diff
@@ -92,7 +92,8 @@
                 "target": {"type": "string"},
                 "shm_size": {"type": ["integer", "string"]},
                 "extra_hosts": {"$ref": "#/definitions/list_or_dict"},
-                "isolation": {"type": "string"}
+                "isolation": {"type": "string"},
+                "extra_cli_build_args": {"$ref": "list_of_strings"}
```
aiordache, line 96: Hi! Thank you for the contribution. Please propose a PR to add this new property in the Compose specification schema at https://github.com/compose-spec/compose-spec/blob/master/schema/compose-spec.json. 

This file is downloaded from there and we try to avoid having changes that are not coming from compose-spec.

zoonage: https://github.com/compose-spec/compose-spec/pull/145

zoonage: Is there a way to run the tests in CI against this branch of compose-spec?

zoonage:
Opened https://github.com/compose-spec/compose-spec/issues/157 for some discussion with the compose-spec team about how to implement this on their side

zoonage:
Closing this out as I don't have capacity to make any progress on this and haven't for a while

PR #8023, opened 2021-01-04
[1.28.0-rc2] Fix changelog typo

hartwork:
A quick typo fix for your consideration


aiordache:
Thanks for noticing the typo. We will fixing this as a part of the release process as we try to avoid having fixes going directly to the release branch. Closing this.

PR #8030, opened 2021-01-10
Allow dependabot to keep python packages up-to-date

jlosito:
This change allows dependabot to check any python dependency which this project uses on a weekly basis and submit PRs with version bumps in order to keep packages up-to-date.

https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/configuration-options-for-dependency-updates

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "pip"
+    directory: "/"
+    schedule:
+      interval: "weekly"


hartwork:
@jlosito any ideas how "dependabot-preview" is different? It seems there are *some* pull request done by some flavor ofm dependabot already at https://github.com/docker/compose/pulls?q=is%3Apr+is%3Aopen+label%3Adependencies . Is that security fixes only and not regular bumps? Would be interesting to know.

jlosito:
I'm not entirely sure myself what's the difference.

Surely all of those pull requests which you've linked can't all be security related. I could imagine there being one or even two which are security related, but all of them seems odd.

I found the post on the forums linked below which describes the difference between dependabot-preview and dependabot. It seems that the difference is one is only for security fixes and the other is for regular bumps regards of security fixes as you've stated.

https://github.community/t/dependabot-vs-dependabot-preview-detecting-security-vulnerabilities/2552

hartwork:
If it was up to me (it's not), I'd definitely add Dependabot for non-security updates :+1: Could you add the signed-of-by line to make the DCO check happy?

jlosito:
Signed and I see DCO is passing now.

jlosito:
I'm going to close this one since it seems dependabot was added after this purposal.

Issue #5730, opened 2018-02-28
add --timestamps option for "up" command


ggmod:
It would be great to have a `--timestamps` option for the "up" command, similarly to how the "logs" command handles it, displaying the timestamp for each line of log message. I've found the "up" command's output to be much more practical than the "logs" command for monitoring the app during development, and also for streaming the aggregated logs of the latest run into a file, but unfortunately the timestamps are missing.

shin-:
Thanks for the suggestion ; I think it's something we can consider.

essamhassan:
@shin- if possible, I can work on this one

jamesgoodhouse:
Just thought I'd bump this as it seems like a great feature.

MadDamDam:
Would be especially useful for docker-compose run.
If used with --rm logs are not available after the run so i often use it with output redirection.
OP's suggestion will allow to easily get timestamps on the output.

stephan-henningsen:
Still useful. Please implement. It's a nice little summer project ;)

urig:
I'd love to try and get this done if it's open to first time contributors. I've tried finding where the `--timestamps` option is parsed for the `logs` command but couldn't quite find it. Can someone point me in the right direction? TX

daevidvincent:
How is this still not implemented? Surely the data is there to display no? Isn't it literally just putting `date()` on the line as each one is output?

aaronsuns:
I really want this options as well! @Auther, isn't just add a data() for the log line?
docker-compose logs has --timestamps. there is no reason to miss it in "up"

mrajcok:
https://stackoverflow.com/a/40721348/215945:
Just in case some people are not aware... you can tail the logs, so this is the workaround I use:

    $ docker-compose up -d
    $ docker-compose logs -f -t [>> out.log]

Use Ctrl-z or Ctrl-c to stop tailing the logs.

jkeiser:
The thing that makes `docker up` different from `docker logs` is, when you stop `docker up` it will also stop running services. This makes it great for a dev environment, and I'd love to be able to see timestamps in it!

urig:
This is a small feature request both in "return" and in "investment" imho, but I'd rather not see it closed off. I've rummaged through the code a bit but could not get it done myself. I repeat my request from 2 years ago - if anyone can give me a pointer as to how to get this done, I'd love to create a PR for this. Thanks 🙏🏻

jkeiser:
@urig just looking through the code (first time here):

[...]

PR #8037, opened 2021-01-17
Add `--timestamps` to up command

bogdanbarna:
Enable `--timestamps` for `docker up` by expanding the log_args dictionary in L1168-1171.

Resolves https://github.com/docker/compose/issues/5730

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -1090,7 +1090,8 @@ class TopLevelCommand:
                                        container. Implies --abort-on-container-exit.
             --scale SERVICE=NUM        Scale SERVICE to NUM instances. Overrides the
                                        `scale` setting in the Compose file if present.
-            --no-log-prefix       Don't print prefix in logs.
+            --no-log-prefix            Don't print prefix in logs.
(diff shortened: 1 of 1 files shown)


aiordache (changes requested):
Hi! The LogPrinter needs to be updated too to print the timestamps.

jkeiser:
It looks like `docker-compose up` has been rewritten to the point that this PR will need to be completely redone. I put an [analysis](https://github.com/docker/compose/issues/5730#issuecomment-965577096) of the things that probably need to be done in the original bug. Grain of salt, I haven't looked in the Go code before today.

ndeloof:
Would better make sense indeed to reconsider this feature on the v2 branch

bogdanbarna:
Hello. I apologize for such a late reply, but somehow I got side-tracked each time I wanted to look at the feedback for this PR.
Given the rewrite from Python to Go (understandable, with all other container/cloud-native tools being built in the latter), I don't see this change being useful in any way.
Thank you for the time to look at this!

Issue #7773, opened 2020-09-17
docker-compose 1.27.3 config depends_on breaks stack deploy


dfresh613:
## Description of the issue
When we apply docker-compose config to a docker-compose file with depends_on, the end result is not deployable by docker stack..

## Steps to reproduce the issue
Example docker-compose file:
```
services:
  test:
    depends_on:
    - test2
    image: a/fakeimage:latest
  test2:
    image: a/fakeimage:latest
version: '3.4'
```

`Use docker-compose -f <example-config> config > new_config`

The result is:

```
services:
  test:
    depends_on:
      test2:
        condition: service_started
    image: a/fakeimage:latest
  test2:
    image: a/fakeimage:latest
version: '3.4'

```

Now, attempting to deploy that new compose config, docker stack deploy fails:
```
docker stack deploy -c configout.yml iservices
services.test.depends_on must be a list
```

This behavior does not occur in 1.26.x or 1.27.2

docker-ce == 19.03.5
docker-compose ==1.27.3

ulyssessouza:
Hello @dfresh613.

Actually the new output is compliant with the new `compose-spec`.
Could you please explain your usecase of using the output of `docker-compose config` as input for `docker stack deploy`?

dfresh613:
We have a docker-compose file to deploy an application that may be deployed either as a stack, or with docker-compose, easiest to maintain one config file for both, and consistently apply updates using other compose files (docker-compose config -f <file1> -f <file2> config

Using docker-compose config allows us to override and apply updates to the same environment file.

Regardless our use case, this is a change that doesn't seem like it should have been made in a patch (.3) release.

samueldeklund:
We are having a similar issue with piping `docker-compose config` output to `docker stack deploy` command.

We use several different environment files and select one based on which server we are deploying to, and use the `--env-file` parameter to inject those values into the yml file. But now that the presence of the `deploy` key throws an error instead of passing it through (even suppressing the error with `--log-level` doesn't re-enable output), our workflow is broken.

crossdot:
I think docker-compose should respect version specified in docker-compose.yml
For 2.* it should be object map with conditions
For 3.* it should be list

dfresh613:
What I find quite funny, and correct me if I'm wrong-- But it seems this change originally failed the existing unit tests which ensures depends_on matches type list: https://ci-next.docker.com/public/blue/organizations/jenkins/compose/detail/PR-7761/1/pipeline/#step-89-log-867

[...]

PR #8054, opened 2021-01-21
Fix docker-compose config depends_on to be compliant to v3 and v2

favipcj:
This change is based on docker-compose documentation https://docs.docker.com/compose/compose-file/compose-file-v3/#depends_on, which points out that version 3 no longer supports the condition form of **depends_on**, which was introduced on version 2.1 https://docs.docker.com/compose/compose-file/compose-file-v2/#depends_on

Resolves https://github.com/docker/compose/issues/7773

### DIFF
--- a/compose/config/serialize.py
+++ b/compose/config/serialize.py
@@ -4,8 +4,8 @@ from compose.config import types
 from compose.const import COMPOSE_SPEC as VERSION
 from compose.const import COMPOSEFILE_V1 as V1
 from compose.version import ComposeVersion
-VERSION30 = '3.0'
-VERSION21 = '2.1'
+VERSION30 = ComposeVersion('3.0')
+VERSION21 = ComposeVersion('2.1')
 
 
 def serialize_config_type(dumper, data):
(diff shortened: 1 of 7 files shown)


ndeloof:
my :-1 on this. 
Compose-spec (https://compose-spec.io/) was designed to embrace all the compose file format legacy without relying on an artificial `version` attribute which doesn't guarantee anything on supported attributes.

About #7773 :
>  When we apply docker-compose config to a docker-compose file with depends_on, the end result is not deployable by docker stack...

`docker-compose config` is not designed to filter a compose file to make it stack-compatible, but to render the actual compose model applying various overrides and interpolations. There's no reason it would silently remove `depends_on` without user being aware for the sole purpose `stack deploy` doesn't support it.

favipcj:
> my :-1 on this.
> Compose-spec (https://compose-spec.io/) was designed to embrace all the compose file format legacy without relying on an artificial `version` attribute which doesn't guarantee anything on supported attributes.
> 
> About #7773 :
> 
> > When we apply docker-compose config to a docker-compose file with depends_on, the end result is not deployable by docker stack...
> 
> `docker-compose config` is not designed to filter a compose file to make it stack-compatible, but to render the actual compose model applying various overrides and interpolations. There's no reason it would silently remove `depends_on` without user being aware for the sole purpose `stack deploy` doesn't support it.

Hi,
It shouldn't be removing `depends_on` but since version 3 and future versions; `depends_on` should be a list and not a dict as on versions 2.X. 
And the current docker-compose config behavior for all the versions is to create a dict for `depends_on` so even when not using stacks, docker-compose is not able to process the docker-compose with a depends_on dictionary on the services. i.e.
```
$docker-compose up
services.nginx.depends_on contains an invalid type, it should be an array
```
So @ndeloof , should be this the expected behavior? because this is not good in automated scenarios as would require reviewing and correcting the compose file generated by docker-compose config before be able to start the services.

glours:

PR #8058, opened 2021-01-22
Revert to Python 3.7 bump for Linux static builds

### LINKED ISSUE #8048: docker-compose 1.28.0 fails on Ubuntu 18.04: "version `GLIBC_2.28' not found"
## Description of the issue
On Ubuntu 18.04, docker-compose fails with 

>  docker-compose --version
[13] Error loading Python lib '/tmp/_MEIn9gFoA/libpython3.9.so.1.0': dlopen: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.28' not found (required by /tmp/_MEIn9gFoA/libpython3.9.so.1.0)

On the same machine, docker-compose 1.27.4 works as expected.

Build an image with both versions of docker-compose installed:

```Dockerfile
FROM ubuntu:18.04
RUN apt update
RUN apt install curl -y
RUN curl -L -o /usr/local/bin/docker-compose-1.28.0 \
    https://github.com/docker/compose/releases/download/1.28.0/docker-compose-Linux-x86_64
RUN curl -L -o /usr/local/bin/docker-compose-1.27.4 \
    https://github.com/docker/compose/releases/download/1.27.4/docker-compose-Linux-x86_64
RUN chmod +x /usr/local/bin/docker-compose-*
```

```bash
docker build --tag ubuntu:18.04-compose .
docker run --rm -it ubuntu:18.04-compose
```

Then, run both versions of docker-compose inside the image:
```bash
root@115848dbd647:/# docker-compose-1.27.4 --version
docker-compose version 1.27.4, build 40524192
root@115848dbd647:/# docker-compose-1.28.0 --version
[13] Error loading Python lib '/tmp/_MEIn9gFoA/libpython3.9.so.1.0': dlopen: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.28' not found (required by /tmp/_MEIn9gFoA/libpython3.9.so.1.0)
```
## Additional information

This issue was demonstrated in a Docker image for easier reproducability.
It also exists on a full desktop edition of Ubuntu 18.04 that was created with

```bash
vagrant init fasmat/ubuntu1804-desktop
vagrant up
```

chris-crone:
Resolves #8048 

This reverts the bump from Python 3.7 to 3.9 which allows us to revert the bump from Debian Stretch to Buster. This implicitly bumped the glibc requirements from 2.24 to 2.28.

**Changelog**
Revert to Python 3.7 for static Linux builds.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,13 +1,13 @@
 ARG DOCKER_VERSION=19.03
-ARG PYTHON_VERSION=3.9.0
+ARG PYTHON_VERSION=3.7.9
 
 ARG BUILD_ALPINE_VERSION=3.12
 ARG BUILD_CENTOS_VERSION=7
-ARG BUILD_DEBIAN_VERSION=slim-buster
+ARG BUILD_DEBIAN_VERSION=slim-stretch
 
 ARG RUNTIME_ALPINE_VERSION=3.12
 ARG RUNTIME_CENTOS_VERSION=7
-ARG RUNTIME_DEBIAN_VERSION=buster-slim
+ARG RUNTIME_DEBIAN_VERSION=stretch-slim
 
 ARG DISTRO=alpine
 
@@ -38,7 +38,7 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
     git \
     libc-dev \
     libffi-dev \
-    libgcc-8-dev \
+    libgcc-6-dev \
     libssl-dev \
     make \
     openssl \
@@ -68,8 +68,8 @@ WORKDIR /code/
 COPY docker-compose-entrypoint.sh /usr/local/bin/
 COPY --from=docker-cli /usr/local/bin/docker /usr/local/bin/docker
 RUN pip install \
-    virtualenv==20.2.2 \
-    tox==3.20.1
+    virtualenv==20.4.0 \
+    tox==3.21.2
 COPY requirements-dev.txt .
 COPY requirements-indirect.txt .
 COPY requirements.txt .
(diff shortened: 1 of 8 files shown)


Dockerfile
```diff
@@ -1,13 +1,13 @@
 ARG DOCKER_VERSION=19.03
-ARG PYTHON_VERSION=3.9.0
+ARG PYTHON_VERSION=3.7.9
```
gtardif, line 2: should we move back to 3.7.7 to avoid any surprise ? (that was in https://github.com/docker/compose/commit/a3e6e28eeb834435eca5a7e96a8016c7314c603e)

Dockerfile
```diff
@@ -1,13 +1,13 @@
...
 
 ARG BUILD_ALPINE_VERSION=3.12
 ARG BUILD_CENTOS_VERSION=7
-ARG BUILD_DEBIAN_VERSION=slim-buster
+ARG BUILD_DEBIAN_VERSION=slim-stretch
```
gtardif, line 6: not sure what was the reason previously to have `stretch-20200414-slim`

Dockerfile
chris-crone: We can do that if we see an issue but patch releases should be OK.

Dockerfile
chris-crone: The date just allows more granularity on exactly which image build you want. This is essentially `stretch-latest-slim` as opposed to the version that was build on 2020-04-14

aiordache (approved):

max-wittig:
Actually 3.6 would even be better, as Centos 7 still has Python 3.6 in it's package feed.

chris-crone:
@max-wittig This is just for the static builds, you may be able to `pip3 install docker-compose` (with Python 3.6) but we don't build static binaries for it.

max-wittig:
Okay that's sad, but understandable. We will use the pip install of docker-compose then.

PR #8075, opened 2021-01-26
Add warning for use of 'docker-compose down -v' in down cli usage #8033

### LINKED ISSUE #8033: Potential usability issue with "docker-compose down -v"
Hey everyone,

I tried searching the forums and issues for this issue and I couldn't find anything.  In most bash applications I use,   -v  is always short for "--version"    or "--verbose".   Two very safe commands if you accidentally type it.

Consider for a moment, a user who has used docker but maybe not familiar with all of the options. They attempt to update a system that uses docker-compose, and they want to find out the version or they want verbose output, and type "docker-compose down -v".   Now they have deleted the database that people may rely on.

I have never personally made that mistake, but I worry other people at my company might.  Does this sound like a valid concern?

grewy:
`docker-compose down -v` run by mistake will delete the database that people may rely on. This PR adds warning to user when they use this cli. The PR makes below two changes - 
1. Ask user for confirmation before executing down with -v flag
2. Add -f (--force) flag support to avoid prompt mentioned in point 1

Given the following `docker-compose.yml` file:

````
version: "3.9"
services:
  web:
    build: .
    ports:
      - "5000:5000"
  redis:
    image: "redis:alpine"
````

Pre-feature:
````
➜  compose git:(master) ✗ docker-compose down -v
Removing network compose_default
WARNING: Network compose_default not found.
➜  compose git:(master) ✗ docker-compose down -v -f
Stops containers and removes containers, networks, volumes, and images
created by `up`.

By default, the only things removed are:

- Containers for services defined in the Compose file
- Networks defined in the `networks` section of the Compose file
- The default network, if one is used

Networks and volumes defined as `external` are never removed.

Usage: down [options]

Options:
    --rmi type              Remove images. Type must be one of:
                              'all': Remove all images used by any service.
                              'local': Remove only images that don't have a
                              custom tag set by the `image` field.
    -v, --volumes           Remove named volumes declared in the `volumes`
                            section of the Compose file and anonymous volumes
                            attached to containers.
    --remove-orphans        Remove containers for services not defined in the
                            Compose file
    -t, --timeout TIMEOUT   Specify a shutdown timeout in seconds.
                            (default: 10)
````

Post-feature:

````
➜ compose git:(master) ✗ python3.7 bin/docker-compose down -v 
-v option will remove volumes. Are you sure? [yN] n
➜  compose git:(master) ✗ python3.7 bin/docker-compose down -v 
-v option will remove volumes. Are you sure? [yN] y
Removing network compose_default
WARNING: Network compose_default not found.
➜  compose git:(master) ✗ python3.7 bin/docker-compose down -v -f
Removing network compose_default
WARNING: Network compose_default not found.
````

Resolves #8033


grewy:
Hi @aiordache @ndeloof @rumpl @ulyssessouza,
Since this PR adds new flag and a warning prompt, this breaks one acceptance test. Please review. Please let me know if you want me to update acceptance test as part of this PR only or in a follow up PR. thanks!

grewy:
Hi @aiordache @ndeloof @rumpl @ulyssessouza - please review.

grewy:
Hi @aiordache @ndeloof @rumpl @ulyssessouza - please review.

yuuyins:
now it seems if you just `docker compose down` it will remove everything anyway, which happened to me.  `-v` flag now seems to stand only for version. This is terrible UX.

ndeloof:
Such a change implies a UX change: users relying on `docker-compose down -v` will get an interactive confirmation message, and this will break usage in scripts.
Also, Compose v1 is End of Life as Compose v2 has been announced G.A, so better consider changes in the v2 codebase

yuuyins:
ok then consider that for v2.

> Such a change implies a UX change: users relying on `docker-compose down -v` will get an interactive confirmation message, and this will break usage in scripts.

`--no-confirm` or `--yes` so that there's no need for interactive confirmation and people can add that to their scripts. I think this is justified enough for breaking backwards compatibility, otherwise we're back to `dd` times.

thaJeztah:
I'm a bit on the fence altogether; the `-v` / `--volumes` flag was specifically added to allow explicitly removing volumes (the default is not to), so this flag is already the ~equivalent of a `--force` / `--yes` option.

ndeloof:
> this flag is already the ~equivalent of a --force / --yes option

I agree. I can't see any reason a user would set this flag if they don't want volumes to be removed.

yuuyins:
@thaJeztah so I'm a bit confused as well, because I thought that was not the case in my experience since I could not find the volume anymore (i did not pass -v). I'll see if I can reproduce this.

yuuyins:
@thaJeztah @ndeloof BUT in any case, for `docker compose down" it should have confirmation even if not passing `-v`.

ndeloof:
if `docker compose down` removes volumes, then this is a bug. This should only happen when `-v` flag is set
(please check you're running latest release)

thaJeztah:
@yuuyins You're running on Linux? (because on macOS and Windows, the `/var/lib/docker/` directory is within Docker Desktop's VM, not on the host)

glours:

PR #8078, opened 2021-01-28
Support use of shell commands at env variable

### LINKED ISSUE #4081: Using shell command in docker-compose.yml
Hello, is there a way to use shell commands in `docker-compose.yml` file?
Here is my use case:

``` yml
 version: '2'
 services:
   ci:
     image: jenkins
     volumes:
       - ./data:/var/jenkins_home
       - $(command -v docker):/usr/bin/docker
     groupadd:
       - $(stat -c %g /var/run/docker.sock)
     ports:
       - "8080:8080"
       - "50000:50000"
```

Currently it's giving me this error: 

```
ERROR: Invalid interpolation format for "volumes" option in service "ci": "${command -v docker}:/usr/bin/docker"
```

GMouraP:
This adds support for the following syntax: ${var:\~(command)}, ${var:\~command}, ${var\~(command)} and ${var\~command}
Allowing the use a shell command to assign a value to a variable when the var is not already set at the host.

For example: "HUID_ENV: $(HUID_HOST:~(id -u))" would assign the value returned by "id -u" (UID of the current user)
to the environment variable "HUID_ENV", unless "HUID_HOST" was already set and not empty.

Signed-off-by: Gustavo Moura gusdev77@gmail.com

Resolves #4081


GMouraP:
This is a repost because I tried to correct the conflicts and realized I'd messed up committing to the master branch on my fork, so I just fixed it and the previous PR was closed automatically.

thaJeztah (changes requested):
Definitely not a fan of this; this looks like a huge security risk

```bash
${var:~(rm -rf /)}
```

`sudo docker-compose up` 💥

GMouraP:
> Definitely not a fan of this; this looks like a huge security risk
> 
> ```shell
> ${var:~(rm -rf /)}
> ```
> 
> `sudo docker-compose up`

Thanks for pointing that out, kind of obvious but I didn't think of that!
I'll work on a fix for this!

thaJeztah:
> I'll work on a fix for this!

I'm not sure that can be fixed. I think there's some other discussions similar to the above; effectively this allows executing arbitrary commands on the host, which is a risk, especially if a person would be "tricked" into running a compose-file (commands wouldn't necessarily have to be  "destructive" operations, but could also (e.g.) "steal your private keys, and send them somewhere")

chris-crone:
Thanks for the proposal @GMouraP !

I agree with @thaJeztah that this is too dangerous to merge so I will close the PR.

GMouraP:
Well, you guys are right!
I was stubborn enough to try to think about a solution to this and the best I came up with was whitelisting the commands one would like to use. The problem with this is that merely using Environment Variables in the host system containing the information needed would a lot be easier, quicker, less troublesome, and safer to do.

Thanks for your time reviewing my PR! Should I have any ideas (that would not cause any security risk) that could be useful to Docker or Compose, I'll be submitting them!

Again, thank you!!

thaJeztah:
Thanks for contributing!

PR #8081, opened 2021-01-28
Update dind setup for tests

aiordache:
Update Jenkins files and scripts:
- mount `.docker/config` for authenticating pulls in dind containers
- replace image `dockerswarm/dind` with `docker:<version>-dind`

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -62,7 +62,7 @@ pipeline {
 
 def buildImage(baseImage) {
     def scmvar = checkout(scm)
-    def imageName = "dockerbuildbot/compose:${baseImage}-${scmvar.GIT_COMMIT}"
+    def imageName = "dockerpinata/compose:${baseImage}-${scmvar.GIT_COMMIT}"
     image = docker.image(imageName)
 
     withDockerRegistry(credentialsId:'dockerbuildbot-index.docker.io') {
@@ -89,7 +89,7 @@ def runTests(dockerVersion, pythonVersion, baseImage) {
         stage("python=${pythonVersion} docker=${dockerVersion} ${baseImage}") {
             node("ubuntu && amd64 && !zfs") {
                 def scmvar = checkout(scm)
-                def imageName = "dockerbuildbot/compose:${baseImage}-${scmvar.GIT_COMMIT}"
+                def imageName = "dockerpinata/compose:${baseImage}-${scmvar.GIT_COMMIT}"
                 def storageDriver = sh(script: "docker info -f \'{{.Driver}}\'", returnStdout: true).trim()
                 echo "Using local system's storage driver: ${storageDriver}"
                 withDockerRegistry(credentialsId:'dockerbuildbot-index.docker.io') {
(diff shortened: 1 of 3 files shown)


Jenkinsfile
```diff
@@ -99,6 +99,8 @@ def runTests(dockerVersion, pythonVersion, baseImage) {
                       --privileged \\
                       --volume="\$(pwd)/.git:/code/.git" \\
                       --volume="/var/run/docker.sock:/var/run/docker.sock" \\
+                      --volume="\$(pwd)/.docker/config.json:/code/.docker/config.json" \\
```
StefanScherer, line 102: I think it needs to be in the home directory, so I assume it is `/root/.docker/config.json` in the container. On the host probably in $HOME.

script/test/all
```diff
@@ -38,14 +38,16 @@ for version in $DOCKER_VERSIONS; do
...
 
   docker run \
     -d \
     --name "$daemon_container" \
     --privileged \
     --volume="/var/lib/docker" \
-    "$repo:$version" \
+    -v ~/.docker/config.json:/code/.docker/config.json \
+    -e "DOCKER_TLS_CERTDIR=" \
+    "$repo:$version-dind" \
```
thaJeztah, line 50: The `$repo` variable is only used here; perhaps remove the variable above, and just inline it here;

```suggestion
    "docker:$version-dind" \
```

script/test/all
```diff
@@ -38,14 +38,16 @@ for version in $DOCKER_VERSIONS; do
 
   trap "on_exit" EXIT
 
-  repo="dockerswarm/dind"
+  repo="docker"
```
thaJeztah, line 42: (see my other comment)

```suggestion
```

Jenkinsfile
thaJeztah: Looks like `script/test/all ` uses `~/.docker/ ...`, which should also work

thaJeztah: (for the "host" side that is 😅)

Jenkinsfile
```diff
@@ -99,6 +99,8 @@ def runTests(dockerVersion, pythonVersion, baseImage) {
                       --privileged \\
```
thaJeztah, line 99: Do we actually need `--privileged` for this container? I see it bind-mounts `docker.sock`, which means that all `docker` commands will actually run against the docker daemon running on the host (not a `dockerd` daemon running inside the container (which _would_ require privileged)

thaJeztah: Actually wondering if we need a container here at all, because it's running a container, just to run a script, that uses the `docker` CLI inside the container to start new containers on the host 🤔

StefanScherer (approved):
LGTM
Wonder if it this works, I saw a `dockerbuildbot/compose` image 😅 
Maybe we should rerun CI?

thaJeztah:
Also opened https://github.com/docker-library/docker/pull/295 for discussion

If I find some time, I may have a look at https://github.com/docker/compose/pull/8081#discussion_r565989372

Issue #8060, opened 2021-01-22
Compatibility with dual-logging in docker engine 20.10


ByJacob:
**Is your feature request related to a problem? Please describe.**
In version 20.10 of the Docker engine, the functionality of double logging has been added (https://docs.docker.com/config/containers/logging/dual-logging/)

However, when I try to use this after using `docker-compose logs -f`, I get the message` WARNING: no logs are available with the 'gelf' log driver`

**Describe the solution you'd like**
Add a docker engine version check. If it supports double login add log display.

**Additional context**
- https://docs.docker.com/config/containers/logging/dual-logging/

PR #8082, opened 2021-01-28
Remove local check for log-driver read support

thaJeztah:
relates to https://github.com/moby/moby/pull/40543
relates to https://github.com/moby/moby/issues/30887
(hopefully) fixes https://github.com/docker/compose/issues/8060

Starting with Docker 20.10, the docker daemon has support for "dual logging", which allows reading back logs, irregardless of the logging-driver that is configured (except for "none" as logging driver).

This patch removes the local check, which used a hard-coded list of logging drivers that are expected to support reading logs.

When using an older version of Docker, the API should return an error that reading logs is not supported, so no local check should be needed.

### DIFF
--- a/compose/cli/log_printer.py
+++ b/compose/cli/log_printer.py
@@ -158,10 +158,8 @@ class QueueItem(namedtuple('_QueueItem', 'item is_stop exc')):
 
 
 def tail_container_logs(container, presenter, queue, log_args):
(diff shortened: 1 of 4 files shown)


thaJeztah:
@ulyssessouza @aiordache ptal (hope I did it right 😂)

aiordache (approved):
Thank you!

Issue #4133, opened 2016-11-09
Add --exclude to docker-compose up


philtay:
I have a Compose file with 10 services. I want to start only 8 of them. I can do `docker-compose up srv1 srv2 srv3 srv4 srv5 srv6 srv7 srv8` but it would be better to do `docker-compose up --exclude srv9 srv10`. Useful syntactic sugar IMO.

johnharris85:
Is it common that you'd want to start a regular subset? If so then perhaps using `extends` would help the use case. Compose is meant to model an application / common set of container deployments. Not sure how common wanting to start a project that is missing a couple of containers would be (although happy to be enlightened!).

philtay:
Judging by the fact that Compose already supports starting a subset, I would say that it's common enough. Many other commands, not only `up`, also allow to specify the services by name. The current syntax makes it easy to specify a limited subset of the services (2 or 3) as opposed to a larger one.

pmzajaczkowski:
For my team it's very common to start all services apart of particular one. That's the case where you are developing some microservices and want to run all of them on your machine (using docker-compose) apart of that single one that you are currently developing and running from your IDE (for example in the debug mode). So that's :+1: from me for --exclude. But that --exclude should also override dependencies - if I don't want to start something it should never start :)

yohanliyanage:
I'm facing the same situation as @pmzajaczkowski. Having this feature is quite useful when testing microservices locally.

DemianTinkiel:
+1 on suggestion. Definitely useful for my team

echo-xu:
+1

DutchKevv:
+1

rangerscience:
+1

My use case is that we have a handful of microservices, and I'm developing one of them. I want to `docker-compose up --exclude <service under development>`, which I then start myself; for clearer logging, custom environment, under hot-reload modes, etc.

er1z:
Second that.

cassiocardoso:
This would be super helpful.

Would appreciate if you guys consider this for a future release.

TVictor:
+1

serbaut:
Workaround:

     docker-compose up --scale foo=0

Flaque:
+1

evilosa:
+1

adithepnarula:
+1

sbz:
+1

ArcanjoQueiroz:
+1

stidiovip:
+1

[...]

PR #8102, opened 2021-02-05
Add --exclude flag to docker-compose up

HansAschenloher:
see: https://github.com/docker/compose/issues/4133

Adds an option to exclude specific services from docker-compose up.
With the --exclude flag all services, except those specified after the 'up' option, are going to be started.

Signed-off-by: Hans Aschenloher <hansi.aschenloher@gmail.com>

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -58,7 +58,6 @@ from .utils import yesno
 from compose.metrics.client import MetricsCommand
 from compose.metrics.client import Status
 
-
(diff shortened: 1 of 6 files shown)


ulyssessouza:
Hello @HansAschenloher. Thank you for the contribution, but this use case should be solved with profiles, as mentioned [here](https://github.com/docker/compose/issues/4133#issuecomment-811841908)

PR #8118, opened 2021-02-15
Use the host_binding_ipv4 value of the bridge network when creating the default network

### LINKED ISSUE #2999: docker-compose host-ip value should defaults to the --ip value of docker daemon
If we don't specify host ip address when declaring the ports in the compose file, like:

```
ports:
  - "8000:8000"
```

docker-compose will assume the host ip address to use is `0.0.0.0`, but if we run the docker daemon with the `--ip=<default-ip>` option, it should use `<default-ip>` instead.

See in the [docker daemon documentation](https://docs.docker.com/engine/reference/commandline/daemon/):
`--ip=0.0.0.0                           Default IP when binding container ports`

The `--ip=<default-ip>` option is correctly taken into account when running a container directly with `docker run`, but when I want to run it through `docker-compose`, it does not work and we have to force host ip value for all services in the docker-compose.yml file like:

```
ports:
  - "10.0.0.1:8000:8000"
```

t0rr3sp3dr0:
Resolves #2999

Currently, the default Docker Compose network is created without a value set for `com.docker.network.bridge.host_binding_ipv4` by default, thus making the exposed ports of containers on it bind to `0.0.0.0`.

This differs from the Docker Daemon behavior and it doesn't make possible to set a system-wide configuration. Instead, each Docker Compose manifest must set the value by itself.

```yaml
networks:
  default:
    driver_opts:
      com.docker.network.bridge.host_binding_ipv4: 127.0.0.1
```

This patch makes Docker Compose use the default binding IP set on the Docker Daemon when creating its default network.
Just like Docker, user-created networks are not affected and still bind to `0.0.0.0` if no other IP is specified in the driver options.

### DIFF
--- a/compose/network.py
+++ b/compose/network.py
@@ -71,6 +71,9 @@ class Network:
             data = self.inspect(legacy=self.legacy)
             check_remote_network_config(data, self)
         except NotFound:
+            if self.name == 'default':
+                self._set_host_binding_ipv4()
+
             driver_name = 'the default driver'
             if self.driver:
                 driver_name = 'driver "{}"'.format(self.driver)
@@ -145,6 +148,23 @@ class Network:
             self.legacy = data is not None
         except NotFound:
             self.legacy = False
+
+    def _set_host_binding_ipv4(self):
+        if self.driver and self.driver != 'bridge':
+            return
+        option = 'com.docker.network.bridge.host_binding_ipv4'
+        if self.driver_opts and option in self.driver_opts:
+            return
+        try:
+            data = self.client.inspect_network('bridge')
+        except NotFound:
+            return
+        host_binding_ipv4 = data.get('Options', {}).get(option, None)
+        if not host_binding_ipv4:
+            return
+        if not self.driver_opts:
+            self.driver_opts = {}
+        self.driver_opts[option] = host_binding_ipv4
 
 
 def create_ipam_config_from_dict(ipam_dict):


t0rr3sp3dr0:
@aiordache @ndeloof @rumpl @ulyssessouza could you review this PR?

aiordache:
@thaJeztah What is your opinion on this?

thaJeztah:
> This differs from the Docker Daemon behavior and it doesn't make possible to set a system-wide configuration. Instead, each Docker Compose manifest must set the value by itself.

@t0rr3sp3dr0 Can you elaborate on that one? The docker daemon itself binds on `0.0.0.0` (and `::`), or are you comparing with Docker Desktop? 🤔

t0rr3sp3dr0:
Sure!

The Docker Daemon binds container ports to `0.0.0.0` by default, but this can be changed by passing the `--ip` argument to dockerd.

https://docs.docker.com/engine/reference/commandline/dockerd/
```
      --ip ip                                 Default IP when binding container ports (default 0.0.0.0)
```

This only affects containers on the default network. If a user create a new network, its containers will continue to bind to `0.0.0.0` unless other IP is specified on network creation.

This patch makes Docker Compose follow the same behavior. The default network will bind to the IP specified on the Docker Daemon and user-created networks will bind to `0.0.0.0` by default.

thaJeztah:
> This patch makes Docker Compose follow the same behavior. The default network will bind to the IP specified on the Docker Daemon and user-created networks will bind to 0.0.0.0 by default.

Gotcha.

So the "default" network that docker compose creates is the "default" network within the compose stack, but it _is_ a custom network.

The "default" network mentioned in the daemon configuration is the default `bridge` network (on Linux). The daemon `--ip` setting only affects containers that are created on the default `bridge` network.

For example;

```yaml
services:
    web:
      image: nginx:alpine
      ports:
        - "80:80"
```

When starting that compose file, you'll see;

```
Creating network "myproject_default" with the default driver
Creating myproject_web_1 ... done
```

The equivalent of the above (with `docker run`) would be;

```bash
# create custom network named myproject_default
docker network create myproject_default

# run the container and attach it to that network
docker run -d --name myproject_web_1 --network=myproject_default -p 80:80 nginx:alpine
```

t0rr3sp3dr0:
Exactly!


This patch would make compose execute something like
```bash
IP=$(docker inspect bridge | jq -r '.[0].Options."com.docker.network.bridge.host_binding_ipv4" // "0.0.0.0"')
docker network create -o "com.docker.network.bridge.host_binding_ipv4=${IP}" myproject_default
```
instead.


This IP could still be overwritten from inside the `docker-compose.yaml` with something like
```yaml
networks:
  default:
    driver_opts:
      com.docker.network.bridge.host_binding_ipv4: 127.0.0.1
```
to make compose run
```bash
docker network create -o "com.docker.network.bridge.host_binding_ipv4=127.0.0.1" myproject_default
```
.


And a custom network like
```yaml
networks:
  mynetwork: {}
```
would still be equivalent to
``` bash
docker network create myproject_ mynetwork
```
.

thaJeztah:
The point is that it doesn't match the default behavior. The `--ip` option was created to configure the default (bridge) network only (at the time no custom networks were possible, so there was no way for users to define these options otherwise). I don't think we should conflate options that are meant for the bridge network to reuse it for other options.

In addition, there are many situations where the user running compose is not managing the daemon configuration. I think a better option would be to work on a design to allow setting default options from the client side. That ways these options could either be defined on a per-project base, or "globally" (i.e., whenever I run `docker network create`, or a compose-file creates a network from this client, it would set `127.0.0.1` or `<my preferred ip>` as a default)

glours:

PR #8122, opened 2021-02-16
Add init container support

### LINKED ISSUE #6855: Feature Request: Init Containers for Docker Compose

Kubernetes has a concept called [Init Containers](https://kubernetes.io/docs/concepts/workloads/pods/init-containers/) which currently has no direct parallel in Docker Compose. There are certain applications designed to take advantage of Init Containers which cannot be correctly tested locally using Docker Compose due to this limitation.

The Docker Compose docs includes an article on [Startup Order for Containers](https://docs.docker.com/compose/startup-order/) but this doesn't capture the essence of this problem. That article stresses, repeatedly, that there is no good way for Docker Compose to determine when a container is truly "up," so dependent applications should be designed with checks. However Init Containers are a totally different animal, as they are designed to start up **and terminate** before the next container runs. Termination is, in fact, very easy to check for and should be thought of as a different use case entirely.

In other words, there should be a solution in Docker Compose to declare that a container is dependent on a special class of "Init Containers" which will start up, run, and terminate before launching the rest. All you'd really have to do is add a flag in the configuration - something like `initContainer: true`

The only "alternatives" aren't really alternatives at all. You can override the default entrypoint / commands of the dependent containers to wait some fixed interval, but this is sloppy guesswork and does not represent good application design. You could write your application code in the dependent containers to check that other containers (with certain names, maybe?) at one time existed and are no longer running... but you see where this is going, no one does this, and indeed no one _should_ try to do this, because it's sloppy, unreliable, and bad design. On the other hand, it would be very easy to check, from a Docker Compose standpoint, that a container has terminated before launching the next.

**Additional context**

An example of an application that makes use of Kubernetes' Init Containers is `vault-env`, by Banzai Cloud. They have an in-depth article on it [here](https://banzaicloud.com/blog/inject-secrets-into-pods-vault/). Essentially, they have an Init Container application which copies itself to a shared volume as a means of injecting secrets from Hashicorp Vault, and then the default entrypoint / command of the dependent container is updated to call the binary in the shared volume. Currently with Docker Compose this is not possible, because it doesn't wait for the first container to finish executing before trying to launch the second, so the application doesn't copy to the shared volume in time. Again, you can hack around this by changing the entrypoint to a long string involving `bash -c sleep 3; /shared-volume/vault-env ...` but this is an absolute mess.

Kubernetes has Init Containers. Docker Compose needs them too!

ojab:
Fixes #6855

See https://github.com/compose-spec/compose-spec/pull/134

### DIFF
--- a/compose/config/compose_spec.json
+++ b/compose/config/compose_spec.json
@@ -188,7 +188,7 @@
                   "properties": {
                     "condition": {
                       "type": "string",
-                      "enum": ["service_started", "service_healthy"]
+                      "enum": ["service_started", "service_healthy", "service_completed_successfully"]
                     }
                   },
                   "required": ["condition"]
--- a/compose/errors.py
+++ b/compose/errors.py
@@ -27,3 +27,8 @@ class NoHealthCheckConfigured(HealthCheckException):
                 service_name
             )
         )
+
+
+class CompletedUnsuccessfully(Exception):
+    def __init__(self, container_id, exit_code):
+        self.msg = 'Container "{}" exited with code {}.'.format(container_id, exit_code)
--- a/compose/parallel.py
+++ b/compose/parallel.py
@@ -16,6 +16,7 @@ from compose.cli.colors import green
 from compose.cli.colors import red
 from compose.cli.signals import ShutdownException
 from compose.const import PARALLEL_LIMIT
+from compose.errors import CompletedUnsuccessfully
 from compose.errors import HealthCheckFailed
 from compose.errors import NoHealthCheckConfigured
 from compose.errors import OperationFailedError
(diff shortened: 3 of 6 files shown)


ojab:
Ci is failing with `Cannot start service svc1: cgroups: cgroup mountpoint does not exist: unknown`, test locally are failing due to `docker: Error response from daemon: manifest for dockerswarm/dind:20.10.3 not found: manifest unknown: manifest unknown.`, so let's send this to review and hope that CI will be fixed somehow.

ndeloof:
LGTM, but please amend your commits for sign-off or we can't accept contributions (git commit amend -s)

ojab:
But there is a signoff already, dunno what DCO expecting from me.

ojab:
Ok, it was expecting the same [real] name as a commit author, pushed with the changed name.

ojab:
Set up test environment locally and reworked the PR to make it actually work, `tests/integration/project_test.py` tests pass here (`test_up_with_nvidia_runtime` is skipped, `test_project_up_port_mappings_with_multiple_files` fails due to local docker setup). 
I've tried to add a test with dependency running `top`, but looks like `project.up()` wait for completion even with `detached=True`, so test hangs.

Ready for review, I guess?

ojab:
And this time CI on Alpine is green, so I suppose it works :tada:

ojab:
Rebased on master, CI is green now.

aiordache (approved):
Thank you

ndeloof:
=> https://github.com/docker/compose-cli/issues/1499

Issue #7686, opened 2020-08-18
OpenSSL version mismatch. Built against 1010106f, you have 101000cf


Legion2:
## Description of the issue

Error when using `docker-compose exec` with `DOCKER_HOST=ssh://...`

## Context information (for bug reports)

**Output of `docker-compose version`**
```
docker-compose version 1.26.2, build eefe0d31
docker-py version: 4.2.2
CPython version: 3.7.7
OpenSSL version: OpenSSL 1.1.0l  10 Sep 2019
```

**Output of `docker version`**
```
Client: Docker Engine - Community
 Version:           19.03.12
 API version:       1.40
 Go version:        go1.13.10
 Git commit:        48a66213fe
 Built:             Mon Jun 22 15:45:36 2020
 OS/Arch:           linux/amd64
 Experimental:      false

Server: Docker Engine - Community
 Engine:
  Version:          19.03.12
  API version:      1.40 (minimum version 1.12)
  Go version:       go1.13.10
  Git commit:       48a66213fe
  Built:            Mon Jun 22 15:49:27 2020
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          v1.2.13
  GitCommit:        7ad184331fa3e55e52b890ea95e65ba581ae3429
 runc:
  Version:          1.0.0-rc10
  GitCommit:        dc9208a3303feef5b3839f4323d9beb36df0a9dd
 docker-init:
  Version:          0.18.0
  GitCommit:        fec3683
```

**Output of `docker-compose config`**
```
version: '3'

services:
  portainer:
    image: portainer/portainer:1.24.1
    command: -H unix:///var/run/docker.sock
    volumes:
      - data:/data
      - /var/run/docker.sock:/var/run/docker.sock
    restart: unless-stopped

volumes:
  data:

```

## Steps to reproduce the issue

1. create a `.env` file and add `DOCKER_HOST=ssh://...` with a valid ssh host with docker engine
2. `docker-compose up -d`
3. `docker-compose exec portainer bash`

### Observed result
```
error during connect: Get http://docker/v1.40/containers/a1a823ab0e6b66087bf0fd609c34da5e70c23446cf19791f66d6bf277f14f89a/json: command [ssh -- example.org docker system dial-stdio] has exited with exit status 255, please make sure the URL is valid, and Docker 18.09 or later 
is installed on the remote host: stderr=OpenSSL version mismatch. Built against 1010106f, you have 101000cf
```
### Expected result

[...]

PR #8124, opened 2021-02-16
Install and pin openssl version to 1.1.1j

aiordache:
The Python cryptography package was built against a newer version of OpenSSL while the binary version  was older and this caused the mismatch error.

- updated also the openssl version for the MacOS binary

```
$ docker-compose version
docker-compose version 1.29.0dev, build fe51950d
docker-py version: 4.4.2
CPython version: 3.7.9
OpenSSL version: OpenSSL 1.1.1j  16 Feb 2021
```

Resolves https://github.com/docker/compose/issues/7686

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -41,8 +41,13 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
     libgcc-6-dev \
     libssl-dev \
     make \
-    openssl \
     zlib1g-dev
+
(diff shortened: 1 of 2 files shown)


chris-crone:
I'm not sure this is actually a fix. If the cryptography library in docker-compose is using the system's libssl/libcrypto, there's always a chance of a build vs run mismatch.

PR #8147, opened 2021-02-23
Set `platform` and `isolation` options for cli builder

aiordache:
Set CLI builder flags.


ndeloof (approved):

ulyssessouza (approved):
LGTM

ilatypov:
This correctly resets the image tag to the required architecture (if there are 2 images with the same tag for different architectures) when using `docker-compose build`.

But using `docker-compose run` against a compose file with a `platform: ${PLATFORM}` property picks one of the 2 images with the same tag indiscriminately (based on the architecture that was built last).

PR #8151, opened 2021-02-23
run docker-compose with remote's local socket

### LINKED ISSUE #8149: enable running containerized docker-compose in a remote host
Running containerized docker-compose with a DOCKER_HOST environment variable that is set to a remote host that is local only (an entry in /etc/hosts) *does not work*.

I would like for the docker-compose shell script to actually work.

To test:
1. use a machine (machineA) that runs docker daemon, with enabled TCP-socket
2. use another machine (machineB) that will use that external DOCKER_HOST for docker by:
 - setting the DOCKER_HOST ENV-variable to a name that you will use to reach the docker host
 - making sure the name resolves to an ip locally, for instance by mapping that name to an ip in /etc/hosts
3. now install the containerized docker-compose on machineB
4. use docker-compose and notice that it will not work because the docker-compose running inside the container on the remote cannot resolve the DOCKER_HOST that was transferred from machineB to the container running on machineA

imme-emosol:
This commit will use /var/run/docker.sock if it appears to be a socket within the docker container.
For local runs this would not add benefits but for remote runs of docker, this would allow using a docker host that is only known to the host that is running this shell script.

Also:
ignore some errors,
use long-options,
keep debugging commands as comments.

Resolves #8149

### DIFF
--- a/script/run/run.sh
+++ b/script/run/run.sh
@@ -87,7 +87,7 @@ done
 
 # Setup environment variables for compose config and context
 # shellcheck disable=SC1117
-ENV_OPTIONS=$(printenv | sed -E "/^PATH=.*/d; /^DOCKER_HOST=.*/d; s/^/-e /g; s/=.*//g; s/\n/ /g")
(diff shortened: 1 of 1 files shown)


glours:

PR #8153, opened 2021-02-24
Fix openssl mismatch bug via docker-py bump (fix in docker-py 4.4.4)

aiordache:
Bump docker-py version to 4.4.4

Fixes https://github.com/docker/compose/issues/7686


chris-crone (approved):
LGTM

Issue #8184, opened 2021-03-09
failed to solve with frontend dockerfile.v0: failed to read dockerfile: open /var/lib/docker/tmp/buildkit-mountXXXXXXXXX/<git_repo_url>#master/docker/Dockerfile.Local: no such file or directory


derekhubbard:
- [x] I have tried with the latest version of Docker Desktop
  - [x] I have tried disabling enabled experimental features
  - [x] I have uploaded Diagnostics
  - Diagnostics ID: 67D87832-0068-4291-B8D3-7BB49110586F/20210309004030

### Expected behavior

`docker-compose` supports the ability to configure a serivce to pull from a hosted git repository, using a dockerfile with a custom name and sitting in a non-default/custom location. 

### Actual behavior

`docker-compose` fails to build the service and runs into the following error: 

```
failed to solve with frontend dockerfile.v0: failed to read dockerfile: open /var/lib/docker/tmp/buildkit-mount428930559/<git_repo_url>#master/docker/Dockerfile.Local: no such file or directory
```

### Information

  - macOS Version: 10.14.6
  - Docker Desktop Version: 3.2.1
  - This appeared after updating Docker Desktop to version 3.2.1. I was previously running version 3.1.0.
  - Resetting my docker settings to the factory defaults did not resolve the issue
  - Everything works as expected when downgrading Docker Desktop to version 3.1.0.

Here's a pseudo-version of our docker-compose setup: 
```
  test_service:
    container_name: test_container
    build:
      context: <git_repo_url>#${GQL_BRANCH:-master}
      dockerfile: ./docker/Dockerfile.Local
```

### Steps to reproduce the behavior

This is a somewhat contrived example, but this should repro for you: 

  1. Create a new repo with a dockerfile in it. Probably any dockerfile will work. The dockerfile I used in my test is included below and is defined with a name different from the default. In my example, I named it `Dockerfile.Local`. 
  2. Also create a new docker-compose file and define a service to be created using the dockerfile defined in the previous step, but pull the file from the git repository via url. My docker-compose file is included below. 
  3. Run `docker-compose --verbose build` locally
  4. Get the folloiwng error: `failed to solve with frontend dockerfile.v0: failed to read dockerfile: open /var/lib/docker/tmp/buildkit-mount302704067/<git_repo_url>#main/Dockerfile.Local: no such file or directory`

Dockerfile.Local: 
```
FROM ubuntu:18.04
COPY . /app
RUN make /app
CMD python /app/app.py
```

[...]

PR #8188, opened 2021-03-09
Do not use context as base dir for Dockerfile when context is a git url

aiordache:
Resolves https://github.com/docker/compose/issues/8184


ulyssessouza (approved):
LGTM

Issue #8169, opened 2021-03-03
docker-compose exec: v1.28 started showing "ERROR: <retval>" on exec error


rfay:
## Description of the issue

If you have a running docker-compose project and do an exec to it and the exec fails, you get extraneous output "ERROR: <returnval>" that didn't use to be there. The return value of docker-compose is set correctly.

For example, use `docker-compose up` with this docker-compose.yaml:
```
version: "3.6"
services:
  mariadb:
    image: mariadb:10.3
    environment:
      - MYSQL_ROOT_PASSWORD=junker99
```

Now create a simple error with `docker-compose exec mariadb bash -c 'ls | grep nothing'`
```
$ docker-compose exec mariadb bash -c 'ls | grep nothing'
ERROR: 1
$ echo $?
1
```

That "ERROR: 1" looks like it's an inadvertent piece of debugging code left in perhaps? It wasn't there in docker-compose before v1.28 - I went back and tried the old versions.

## Context information (for bug reports)

**Output of `docker-compose version`**
```
$ docker-compose version
docker-compose version 1.28.5, build c4eb3a1f
docker-py version: 4.4.4
CPython version: 3.9.0
OpenSSL version: OpenSSL 1.1.1h  22 Sep 2020
```

**Output of `docker version`**
```
$ docker version
Client: Docker Engine - Community
 Cloud integration: 1.0.9
 Version:           20.10.3
 API version:       1.41
 Go version:        go1.13.15
 Git commit:        48d30b5
 Built:             Fri Jan 29 14:28:27 2021
 OS/Arch:           darwin/amd64
 Context:           default
 Experimental:      true

Server: Docker Engine - Community
 Engine:
  Version:          20.10.3
  API version:      1.41 (minimum version 1.12)
  Go version:       go1.13.15
  Git commit:       46229ca
  Built:            Fri Jan 29 14:31:38 2021
  OS/Arch:          linux/amd64
  Experimental:     true
 containerd:
  Version:          1.4.3
  GitCommit:        269548fa27e0089a8b8278fc4fc781d7f65a939b
 runc:
  Version:          1.0.0-rc92
  GitCommit:        ff819c7e9184c13b7c2607fe6c30ae19403a7aff
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

**Output of `docker-compose config`**
(Make sure to add the relevant `-f` and other flags)

```
services:
  mariadb:
    environment:
      MYSQL_ROOT_PASSWORD: junker99
[...]

PR #8213, opened 2021-03-17
Fix on removing error message on 'exec' error

ulyssessouza:
Resolves https://github.com/docker/compose/issues/8169

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -137,7 +137,7 @@ def get_filtered_args(args):
 
 
 def exit_with_metrics(command, log_msg=None, status=Status.SUCCESS, exit_code=1):
-    if log_msg:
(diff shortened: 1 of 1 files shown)


aiordache (approved):

compose/cli/main.py
```diff
@@ -137,7 +137,7 @@ def get_filtered_args(args):
 
 
 def exit_with_metrics(command, log_msg=None, status=Status.SUCCESS, exit_code=1):
-    if log_msg:
+    if log_msg and command != 'exec':
```
wadewilliams, line 140: Perhaps I'm driving by, but, this was a useful error? Now, when you execute commands that rely on docker containers, they exit 0 as if everything is fine. I'm sorry for being lazy, but, why is this an improvement? @ulyssessouza

PR #8216, opened 2021-03-18
Filter out ansi escape sequences from build output

aiordache:
The colored build output is not visible in windows powershell.

image

After removing ansi escape sequences:

image


ndeloof:
Can't we get this fixed by buildkit, so that we don't have to replicate this fix on various tools?

PR #8247, opened 2021-04-01
Add Snyk scan suggestion when building

ulyssessouza:
Resolves https://github.com/docker/dev-tooling-team/issues/286

### DIFF
--- a/compose/cli/scan_suggest.py
+++ b/compose/cli/scan_suggest.py
@@ -1,0 +1,85 @@
+import json
+import logging
+import os
+from distutils.util import strtobool
+
+from docker.constants import IS_WINDOWS_PLATFORM
+from docker.utils.config import find_config_file
+
+
+SCAN_BINARY_NAME = "docker-scan" + (".exe" if IS_WINDOWS_PLATFORM else "")
+
+log = logging.getLogger(__name__)
+
+
+class ScanConfig:
+    def __init__(self, d):
+        self.optin = False
+        vars(self).update(d)
+
+
+def display_scan_suggest_msg():
+    if environment_scan_avoid_suggest() or \
+            scan_available() is None or \
+            scan_already_invoked():
+        return
+    log.info("Use 'docker scan' to run Snyk tests against images to find vulnerabilities "
+             "and learn how to fix them")
+
+
+def environment_scan_avoid_suggest():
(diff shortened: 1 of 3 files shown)


compose/cli/scan_suggest.py
```diff
@@ -0,0 +1,59 @@
...
+    data = ''
+    with open(scan_config_file) as f:
+        data = f.read()
+    scan_config = json.loads(data, object_hook=ScanConfig)
+    return scan_config.optin if isinstance(scan_config.optin, bool) else strtobool(scan_config.optin)
+
+
+def scan_available():
+    docker_folder = docker_config_folder()
+    scan_config_file = os.path.join(docker_folder, 'cli-plugins', SCAN_BINARY_NAME)
```
thaJeztah, line 52: Note that this is not the only location where the plugin can be installed; see https://github.com/docker/cli/issues/1534

ulyssessouza: Fixed

compose/cli/scan_suggest.py
```diff
@@ -0,0 +1,80 @@
...
+
+    scan_config_file = os.path.join(docker_folder, 'scan', "config.json")
+    if not os.path.exists(scan_config_file):
+        return False
+
+    data = ''
+    with open(scan_config_file) as f:
+        data = f.read()
+    scan_config = json.loads(data, object_hook=ScanConfig)
+    return scan_config.optin if isinstance(scan_config.optin, bool) else strtobool(scan_config.optin)
```
gtardif, line 47: what happens here if anything goes wrong (can't read file, can't unmarshall json, ...)
We should return true by default, to not bother users if we cannot check if scan has been executed previously

gtardif (approved):

gtardif (changes requested):
Actually building & testing the branch, I don't manage to see the scan suggest message on mac or windows. 
I suppose this is because you expect the `optin` field to be defined in the config json file, by default after desktop install this config fileexist but does not have the field defined, it gets defined only after a first run of `docker scan`. This is when the scan suggest message should be displayed, where `optin` is not present in the config json

gtardif (approved):
Checked on windows, displays the message, and does not display it anymore after a `docker scan` invocation, LGTM

aiordache (approved):
Tested on osx. It displays the scan suggestion on build. After running `docker scan`, the message no longer appears.

PR #8261, opened 2021-04-07
Ignore command names on NoSuchCommand exceptions

ulyssessouza:
Resolves https://github.com/docker/dev-tooling-team/issues/298

### DIFF
--- a/compose/metrics/client.py
+++ b/compose/metrics/client.py
@@ -36,7 +36,7 @@ class MetricsCommand(requests.Session):
                  context_type=None, status=Status.SUCCESS,
                  source=MetricsSource.CLI, uri=None):
         super().__init__()
-        self.command = "compose " + command if command else "compose --help"
+        self.command = ("compose " + command).strip() if command else "compose --help"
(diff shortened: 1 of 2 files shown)


gtardif (commented):
Please add some test case to metrics_test.py for this bug.

compose/metrics/client.py
```diff
@@ -36,7 +36,7 @@ def __init__(self, command,
                  context_type=None, status=Status.SUCCESS,
                  source=MetricsSource.CLI, uri=None):
         super().__init__()
-        self.command = "compose " + command if command else "compose --help"
+        self.command = ("compose " + command).strip() if command else "compose --help"
```
gtardif, line 39: I suppose this is to leave `self.command = "compose"` value in case of NoSuchCommand exception (when command = "") ?

ulyssessouza: Yes. In the case of `NoSuchCommand` the variable will have an empty string leaving a trailing space on the final string. The `strip()` is to remove it in this case.

aiordache (approved):

PR #8313, opened 2021-05-07
Remove advertisement for `docker compose` from up

ulyssessouza:
Remove advertisement for docker compose V2 from docker-compose up


mat007 (approved):
Yes!

PR #8328, opened 2021-05-13
Add --pull arg to "up" command

GabrielBB:
Ability to do: 

 `docker-compose -f my-docker-compose.yml up --pull` 

Instead of :

`docker-compose -f my-docker-compose.yml pull && docker-compose -f my-docker-compose.test.yml up`. 

Why:

- It's shorter

- Less prone to errors since you don't have to type the yml file name twice (Which means double chance for typos)

- It has better performance because `&&` throws away the asynchrony

- Makes the API more consistent since there is already a `--build` flag and one expect to be a `--pull` one.

One example without this flag that looks weird to me (maybe is just me) is this:

`docker-compose pull && docker-compose up --build`

There it looks like `pull` and `build` are two features with different degrees of importance, when in fact, they don't, the condition to do one thing or another is even in the same function in the code.

This looks a better fit:

`docker-compose up --build --pull`

I added unit tests and acceptance tests

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -1111,6 +1111,7 @@ class TopLevelCommand:
                                        and image haven't changed.
             --always-recreate-deps     Recreate dependent containers.
                                        Incompatible with --no-recreate.
+            --pull                     Always attempt to pull a newer version of the image.
             --no-recreate              If containers already exist, don't recreate
                                        them. Incompatible with --force-recreate and -V.
             --no-build                 Don't build an image, even if it's missing.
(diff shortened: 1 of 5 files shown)


ndeloof:
a quick note: if this is approved, need to be also implemented on https://github.com/docker/compose-cli for UX parity

tomaswarynyca:
Any progress on the issue?

GabrielBB:
Should I close this? @aiordache @ndeloof @rumpl

ndeloof:
I indeed don't think compose v1 will evolve anymore, or only for security fixes.
Better get this implemented on the v2 branch

PR #8353, opened 2021-05-31
Add details to warning

### LINKED ISSUE #8343: [BUG?] docker-compose exec does not accept the container name
According to the [documentation](https://docs.docker.com/compose/reference/exec/)
```
docker-compose exec
```
is an equivalent of
```
docker exec
```
but this is not true, as docker-compose exec does not accept container's name, it only accepts the service name.
If you try to use container name, you get
```
ERROR: No such service: my-super-container-name
```

Add such functionality to docker-compose exec.

Add a corresponding warning to the documentation? And/Or make the error message more informative, e.g.
```
ERROR: No such service: my-super-container-name. Make sure you use service name, not container name.
```

bartier:
Resolves #8343

This gives details about the exec command using docker-compose as discussed in the #8343 issue.

### DIFF
--- a/compose/project.py
+++ b/compose/project.py
@@ -1153,7 +1153,9 @@ class NoSuchService(Exception):
         if isinstance(name, bytes):
             name = name.decode('utf-8')
         self.name = name
-        self.msg = "No such service: %s" % self.name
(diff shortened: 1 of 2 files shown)


glours:

PR #8471, opened 2021-08-09
Resolve seccomp path relative to docker-compose.yml

### LINKED ISSUE #8470: Relative seccomp paths are resolved relative to cwd instead of docker-compose.yml
This means that if you have a seccomp declaration for one of your services like:

```
    security_opt:
      - seccomp:./some_dir/example.json
```

You can only use `docker-compose` commands from the root dir and not from any subdir

carlbordum:
Resolves #8470


glours:

PR #8594, opened 2021-09-09
exclude .github from pre-commit

ndeloof:
This will prevent CI failure like https://ci-next.docker.com/public/blue/organizations/jenkins/compose/detail/PR-8589/3/pipeline/


ulyssessouza (approved):
LGTM

Issue #8595, opened 2021-09-09
Not able to add volume through cli command by using --volume parameter on docker-compose v2


sliwinski-milosz:
I am trying to add the volume to the docker-compose run command by passing `--volume` flag as described [here](https://docs.docker.com/compose/reference/run/) but I get:
```
unknown flag: --volume
```

Please note, that I have tried the short version of the parameter: `-v` and it works fine.

```
$ docker-compose --version
Docker Compose version v2.0.0-rc.2
$ docker --version
Docker version 20.10.8, build 3967b7d
```

I have tested on the docker-compose v1 and the `--volume` parameter (long one) works fine:
```
$ docker-compose --version
docker-compose version 1.29.2, build 5becea4c
```

EDIT: I realised it has been renamed from `--volume` to `--volumes`. Is that expected?
https://github.com/docker/compose/blob/v2.0.0-rc.3/cmd/compose/run.go#L48

ndeloof:
unexpected, this is a mistake we made replicating the legacy `docker-compose run` command.

mat007:
Fixed in #8596

PR #8596, opened 2021-09-09
fix run --volume flag (mistakely renamed --volumes)

ndeloof:
**What I did**
Renamed `--volumes` run flag as `--volume` to align with Compose V1
Created an alias for compatibility

**Related issue**
close https://github.com/docker/compose/issues/8595

### DIFF
--- a/cmd/compose/run.go
+++ b/cmd/compose/run.go
@@ -28,6 +28,7 @@ import (
 	"github.com/mattn/go-isatty"
 	"github.com/mattn/go-shellwords"
 	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
 
 	"github.com/docker/cli/cli"
 	"github.com/docker/compose/v2/pkg/api"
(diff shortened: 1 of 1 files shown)


ulyssessouza (approved):
LGTM

cmd/compose/run.go
```diff
@@ -148,15 +149,25 @@ func runCommand(p *projectOptions, backend api.Service) *cobra.Command {
...
 	flags.BoolVar(&opts.servicePorts, "service-ports", false, "Run command with the service's ports enabled and mapped to the host.")
 
+	flags.SetNormalizeFunc(normalizeRunFlags)
 	flags.SetInterspersed(false)
 	return cmd
 }
 
+func normalizeRunFlags(f *pflag.FlagSet, name string) pflag.NormalizedName {
+	switch name {
+	case "volumes":
```
mat007, line 164: Should we really introduce an alias given V2 isn’t GA yet?

ndeloof: we already have hundred thousands users, so: yes, we should.

mat007: Right, but there must be a way to handle deprecation for this then.
Or do we just want to keep this around forever?

ndeloof: sure, can use `Flags().MarkDeprecated()`

mat007: I don’t think that would help here as what this does is just add a «deprecated» mention in the help for a flag.
But here ` volumes `  is not a flag anymore and won’t show in the help anyway.

Never mind, not a big deal, and we can do something after GA if needed.

PR #8604, opened 2021-09-10
introduce `--output` option to `config` command

ndeloof:
This removes hack by compose-cli kube backend to expose additional `--output` flag by just making this an official `docker compose config` flag (defaults to stdout)


ulyssessouza (approved):
LGTM

PR #8611, opened 2021-09-13
run watchContainers with a background context

ndeloof:
When compose is ran attached by `compose up`, and user interrupt by `Ctrl+C`, the log printer is not notified about container termination and compose is stuck waiting.

This PR run `watchContainers` with a background context so it won't get canceled by the top level signal management (see `cmd/compose.go#AdaptCmd`). Doing so, the `logPrinter` will recieve `ContainerExit` events after stop sequence has started, and detect application is down to terminate with adequate exitcode

# How to test:
1. Run compose app with `docker compose up`
```yaml
services:
  web:
    image: nginx
````

2. Hit [Ctrl+C]
3. See notification for container being "Stopped", but still compose command doesn't exit

### DIFF
--- a/pkg/compose/logs.go
+++ b/pkg/compose/logs.go
@@ -37,7 +37,7 @@ func (s *composeService) Logs(ctx context.Context, projectName string, consumer 
 	if options.Follow {
 		eg.Go(func() error {
 			printer := newLogPrinter(consumer)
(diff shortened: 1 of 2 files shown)


mat007 (approved):
Makes sense!

lorenrh (approved):

PR #8617, opened 2021-09-14
compose/run: don't waitDependencies if NoDeps

### LINKED ISSUE #8577: run --no-deps hangs when using service_healthy
If no containers are already running, and `service_healthy` is used, `run --no-deps` will just hang.

In the python version, it would start the required container.

Example:
```yaml
version: "3.9"

services:

  container1:
    image: alpine
    command: tail -f /dev/null
    healthcheck:
      test: sh -c 'exit 0'
      interval: 5s
      timeout: 10s
      retries: 3

  container2:
    image: alpine
    depends_on:
      container1:
        condition: service_healthy
```

This command will hang:
```bash
docker compose --file=test.compose.yaml run --no-deps container2 echo test
```

In the python version, it would start:
```
docker-compose --file=test.compose.yaml run --no-deps container2 echo test
```

aviau:
Resolves #8577

### DIFF
--- a/cmd/compose/run.go
+++ b/cmd/compose/run.go
@@ -210,6 +210,7 @@ func runRun(ctx context.Context, backend api.Service, project *types.Project, op
 		Entrypoint:        opts.entrypointCmd,
 		Labels:            labels,
 		UseNetworkAliases: opts.useAliases,
+		NoDeps:            opts.noDeps,
 		Index:             0,
 	}
 	exitCode, err := backend.RunOneOffContainer(ctx, project, runOpts)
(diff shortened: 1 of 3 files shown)


ndeloof:
Compose model should not have any reference to disabled services, so code doesn't need to run checks like the one you suggest here.
A better (?) fix would be to skip https://github.com/docker/compose/blob/f7c360b721b1e2a0bb1eb78b60cd60bb1c1c6199/pkg/compose/run.go#L163 when `run` gets the `--no-deps` flag.

aviau:
@ndeloof Sure! I have updated the PR:

- Added a `NoDeps` field to api.RunOptions`
- Only `waitDependencies` if `NoDeps` is `true`.

Let me know if this looks like what you were thinking :)

ndeloof (approved):

lorenrh (approved):
LGTM 🎉

aviau:
thanks!

Issue #8619, opened 2021-09-15
docker-compose run --rm throws "You cannot remove a running container" on experimental V2


melicerte:
## Context information (for bug reports)

- [x] Using Compose V2 `docker compose ...`
- [ ] Using Compose V1 `docker-compose ...`

**Output of `docker(-)compose version`**
```
Docker Compose version v2.0.0-rc.3
```

**Output of `docker version`**
```
Client:
 Cloud integration: 1.0.17
 Version:           20.10.8
 API version:       1.41
 Go version:        go1.16.6
 Git commit:        3967b7d
 Built:             Fri Jul 30 19:55:20 2021
 OS/Arch:           darwin/amd64
 Context:           default
 Experimental:      true

Server: Docker Engine - Community
 Engine:
  Version:          20.10.8
  API version:      1.41 (minimum version 1.12)
  Go version:       go1.16.6
  Git commit:       75249d8
  Built:            Fri Jul 30 19:52:10 2021
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.4.9
  GitCommit:        e25210fe30a0a703442421b0f60afac609f950a3
 runc:
  Version:          1.0.1
  GitCommit:        v1.0.1-0-g4144b63
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

**Output of `docker-compose config`**
(Make sure to add the relevant `-f` and other flags)
```
services:
  back:
    build:
      context: /Users/user/workspace/myproject/docker/back
      dockerfile: /Users/user/workspace/myproject/docker/back/Dockerfile
    depends_on:
      php-fpm:
        condition: service_started
    environment:
      COMPOSE_PROJECT_NAME: myproject
      DB_DATABASE: myproject
      DB_HOST: mysql
      DB_PASSWORD: myproject
      DB_PORT: "3306"
      DB_ROOT_PASSWORD: myproject
      DB_USERNAME: myproject
    env_file:
    - .env_docker
    extends:
      file: docker-compose.base.yml
      service: back-base
    image: myprojectregistrydev.azurecr.io/myproject/back:latest
    networks:
      default: null
    ports:
    - mode: ingress
      target: 8080
      published: 80
      protocol: tcp
    volumes:
    - type: bind
      source: /Users/user/workspace/myproject
      target: /var/www
      bind:
        create_host_path: true
    working_dir: /var/www
  mysql:
    environment:
      COMPOSE_PROJECT_NAME: myproject
[...]

PR #8621, opened 2021-09-15
waitForContainer to ensure stopped before remove

ndeloof:
This is an attempt to fix https://github.com/docker/compose/issues/8619

I'm not sure why container is still running when we invoke `remove`. stdout stream has been closed, which I expected only happens once the container is terminated. Maybe a race condition?

Anyway, this PR relies on `containerWait` to both collect exitCode and enforce container is stopped before removal

### DIFF
--- a/pkg/compose/run.go
+++ b/pkg/compose/run.go
@@ -21,11 +21,11 @@ import (
 	"fmt"
 	"io"
 
-	"github.com/docker/compose/v2/pkg/api"
-
 	"github.com/compose-spec/compose-go/types"
 	"github.com/docker/cli/cli/streams"
+	"github.com/docker/compose/v2/pkg/api"
 	moby "github.com/docker/docker/api/types"
+	"github.com/docker/docker/api/types/container"
 	"github.com/docker/docker/pkg/ioutils"
 	"github.com/docker/docker/pkg/stdcopy"
 	"github.com/docker/docker/pkg/stringid"
@@ -100,7 +100,10 @@ func (s *composeService) runInteractive(ctx context.Context, containerID string,
 	for {
 		select {
 		case err := <-outputDone:
-			return s.terminateRun(ctx, containerID, opts, err)
+			if err != nil {
+				return 0, err
+			}
+			return s.terminateRun(ctx, containerID, opts)
 		case err := <-inputDone:
 			if _, ok := err.(term.EscapeError); ok {
 				return 0, nil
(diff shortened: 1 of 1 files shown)


pkg/compose/run.go
```diff
@@ -115,22 +118,18 @@ func (s *composeService) runInteractive(ctx context.Context, containerID string,
...
+	case exit := <-exitCh:
+		exitCode = int(exit.StatusCode)
+	case err = <-errCh:
+		return
 	}
 	if opts.AutoRemove {
 		err = s.apiClient.ContainerRemove(ctx, containerID, moby.ContainerRemoveOptions{})
 	}
-	return exitCode, err
+	return
```
lorenrh, line 132: I feel like I'm misreading, but `exitCode` being not being returned in this function?

ndeloof: function is using named return variables `(exitCode int, err error)`
so just need to set value then return

lorenrh: ah! Thanks!

lorenrh (approved):

pkg/compose/run.go
```diff
@@ -115,22 +118,18 @@ func (s *composeService) runInteractive(ctx context.Context, containerID string,
...
+	exitCh, errCh := s.apiClient.ContainerWait(ctx, containerID, container.WaitConditionNotRunning)
+	select {
+	case exit := <-exitCh:
+		exitCode = int(exit.StatusCode)
+	case err = <-errCh:
+		return
 	}
 	if opts.AutoRemove {
 		err = s.apiClient.ContainerRemove(ctx, containerID, moby.ContainerRemoveOptions{})
 	}
```
thaJeztah, line 131: Note that the `AutoRemove` option in API v1.30 and up (Docker 17.06 and up) is handled by the daemon (it will remove the container when it exits) (assuming the `AutoRemove` option is passed to the daemon when compose creates the container)

While it's of course possible for the client to also (manually) remove the container, there may be a race condition in that case (container no longer exists because it's already removed, or an error that the container is "already in the process of being removed").

See https://github.com/docker/cli/blob/135ffd205791001337d283310b286261242444e4/cli/command/container/run.go#L165-L183

And https://github.com/docker/cli/blob/135ffd205791001337d283310b286261242444e4/cli/command/container/utils.go#L16

ndeloof: sure, I'd like we can just get rid of all this stuff by using some of the docker/cli `runXX` functions - will need to make those public

PR #8624, opened 2021-09-15
(WIP)add --build option for run subcommand

### LINKED ISSUE #8623: Support `--build` option on `compose run` subcommand
When using `docker-compose run` command, it will use an already built image.
But in some cases, it will be frustrated.

For example, in running a code test with the container, the container image is already built and so I always need to run `docker-compose build` before runs `docker-compose run test`.

BTW, There is no `--build` option for `docker-compose run`.
But it seems no reason why the option not exists.

If the `--build` option exists, it is ok using `docker-compose --build run test`. run with a single command and two commands are largely different user experience.

Add `--build` and `--no-build` option for `docker-compose run` as same as `docker-compose up`.

- run `docker-compose up --build test`: this command assumes to run not single service so output log doesn't seem good and it can not use for interactive mode.
- run `docker-compose build ; docker-compose run test`: double commands. it doesn't compatible with other commands.

**Additional context**
- I already implemented this subcommand option: https://github.com/docker/compose/pull/8625

yuniruyuni:
Resolves #8623


ndeloof:
new features should better be implemented in Compose V2 (v2 branch)

yuniruyuni:
thank you...! I'll change branch

yuniruyuni:
branch changed and I need restart from initial because of language changed from python to golang. please wait to develop it. (It seems v2 have better codes so doesn't take much time.)

yuniruyuni:
I failed to change this branch's base branch and so I have created new PR for this: https://github.com/docker/compose/pull/8625 please continue to review on that branch :bow:

PR #8625, opened 2021-09-15
add createOptions for run command arguments

### LINKED ISSUE #8623: Support `--build` option on `compose run` subcommand

yuniruyuni:
Resolves #8623

it is base branch change for #8624

### DIFF
--- a/cmd/compose/run.go
+++ b/cmd/compose/run.go
@@ -103,56 +103,57 @@ func (opts runOptions) apply(project *types.Project) error {
 }
 
 func runCommand(p *projectOptions, backend api.Service) *cobra.Command {
-	opts := runOptions{
+	run := runOptions{
 		composeOptions: &composeOptions{
 			projectOptions: p,
 		},
 	}
+	create := createOptions{}
 	cmd := &cobra.Command{
 		Use:   "run [options] [-v VOLUME...] [-p PORT...] [-e KEY=VAL...] [-l KEY=VALUE...] SERVICE [COMMAND] [ARGS...]",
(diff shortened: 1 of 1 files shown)


ciaranmcnulty:
This would be great; currently when we `run` something in our CI we have to use a yml with the build steps removed, to prevent re-building when container pull fails

milas:
We have decided to not support a `--build` flag on `compose run` right now. I understand this might be frustrating for some use cases, but we try to consider both the ongoing maintenance impact of any new feature/option along with available alternatives/workarounds.

In this case, consider `docker compose build <service> && docker compose run <service>` or similar to trigger a build (if necessary) before the `run` invocation.

glours:
And in many case, especially the one describe in #8623, a simple `docker compose up --build <service> should provide the expecting behaviour

Issue #8518, opened 2021-07-12
docker-compose config unescapes $$ to $


puckey:
Say we have a docker-compose.yml contains the $ sign, and it should be escaped, like:

```
version: '3.8'
services:
  exporter:
    image: prom/node-exporter
    command:
      - -option_without_quote=$$
      - '-option_single_quote=$$'
      - "-option_double_quote=$$"
```

However, when using docker `compose config` to show this file, the $ sign are not escaped anymore:

```$ docker compose -f sample.yml config
services:
  exporter:
    command:
    - -option_without_quote=$
    - -option_single_quote=$
    - -option_double_quote=$
    image: prom/node-exporter
version: '3.8'
```
This will cause the problem when the output will be used in docker stack deploy case, as following error will be raised:

Invalid interpolation format for "command" option in services "exporter": "-option_without_quote=$". You may need to escape any $ with another $.

See the original issue where this behaviour was fixed in docker-compose here: https://github.com/docker/compose/issues/5031

vrestihnat:
you can try use 'sed' see:
`docker-compose -f  docker-compose.base.yml config | sed -e 's/\$\$/\$/g' >  docker-compose.yml`

ndeloof:
> This will cause the problem when the output will be used in docker stack deploy

Why do you use `config` to pre-process your compose file, and don't just pass it to `stack deploy`?

PR #8632, opened 2021-09-17
`compose config` to espace `$` signs so compose file is still valid

ndeloof:
Escape `$` sign used to define service command, entrypoint or environment variables so that the resulting compose file can still be parsed by docker-compose/stack deploy 

Resolves #https://github.com/docker/compose/issues/8518

### DIFF
--- a/pkg/compose/compose.go
+++ b/pkg/compose/compose.go
@@ -17,6 +17,7 @@
 package compose
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"fmt"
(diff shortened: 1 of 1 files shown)


pkg/compose/compose.go
```diff
@@ -67,9 +67,19 @@ func getContainerNameWithoutProject(c moby.Container) string {
...
+		out := strings.ReplaceAll(string(marshal), "$", "$$")
+		return []byte(out), err
 	case "yaml":
-		return yaml.Marshal(project)
+		marshal, err := yaml.Marshal(project)
+		if err != nil {
+			return nil, err
+		}
+		out := strings.ReplaceAll(string(marshal), "$", "$$")
+		return []byte(out), err
```
mat007, line 82:
```suggestion
		return []byte(out), nil
```

mat007, line 82: Probably wouldn’t optimize much, but this could just be
```suggestion
		out := bytes.ReplaceAll(marshal, []byte("$"), []byte("$$"))
		return out, nil
```
(and we could even extract to a function as it’s duplicated)

mat007 (approved):
Makes sense!

emersonf:
👋  We have a bunch of Compose files that we bring together using `docker compose config`. Those files contain variable references, e.g. `image: repo/foo:${FOO_VERSION}` and an `.env` file contains `FOO_VERSION=1.2`.

Before this change went live last night in Docker for Mac, the output of `docker compose config` would contain `image: repo/foo:1.2`. Now it contains `image: repo/foo:$${FOO_VERSION}`, breaking our tooling.

Would it be possible to preserve `$$` in the original files (which seems to be the linked bug), instead of escaping `$` to `$$` and breaking variable interpolation?

PR #8635, opened 2021-09-20
rebuild compose project from labels

### LINKED ISSUE #8552: Rebuild compose model from actual resources
This is a follow-up proposal for https://github.com/docker/compose-cli/pull/968#issuecomment-735207016

`compose up` creates/updates resources from compose.yaml definition, but all other commands might not have access to this file (or can't guarantee it has not been edited in the meantime). Resources should be created with metadata (labels) so we can rebuild a partial compose-go model by just inspecting actual resources.

ndeloof:
Introduce `com.docker.compose.depends_on` label and use it to rebuild project model from actual containers without the need for the original compose file (which might have been edited/removed) and environment (which is only partially captured by `com.docker.compose.project.environment_file`, we miss os.Env set by user)

=> `docker compose --project-name foo down` is able to stop container in reverse dependencies order without original compose file

To be debated if we still want `docker compose down` to parse the compose file and only stopped declared containers, or just compute project name

Resolves #8552

### DIFF
--- a/pkg/api/labels.go
+++ b/pkg/api/labels.go
@@ -49,6 +49,8 @@ const (
 	SlugLabel = "com.docker.compose.slug"
 	// ImageDigestLabel stores digest of the container image used to run service
 	ImageDigestLabel = "com.docker.compose.image"
+	// DependenciesLabel stores service dependencies
+	DependenciesLabel = "com.docker.compose.depends_on"
 	// VersionLabel stores the compose tool version used to run application
 	VersionLabel = "com.docker.compose.version"
 )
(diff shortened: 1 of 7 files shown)


mat007 (commented):
> To be debated if we still want docker compose down to parse the compose file and only stopped declared containers, or just compute project name

Wouldn’t it be weird if one worked one way and not the other?
I can’t think of a use case where we’d want `down` to use a compose file different than when `up` was done, but maybe I lack the experience with compose…

ndeloof:
> I can’t think of a use case where we’d want down to use a compose file different than when up was done

never under-estimate user creativity to adopt weird usages :D

I can imaging some may use this as an alternative to `docker-compose down SERVICE_A SERVICE_B`, but this indeed would be a corner case

ulyssessouza (approved):
LGTM

PR #8637, opened 2021-09-20
introduce --compatibility (noop) for backward ... compatibility

### LINKED ISSUE #8618: `docker-compose --compatibility up` gives *unknown flag: --compatibility*
## Description of the issue

After the latest update to Docker Desktop the docker-compose commands I use when integration testing no longer work. 

As there is limited memory on development/build machines there are memory limits on the containers defined in the compose files. These compose files use version 3.x of the compose file format (they are also used for deployment), and so the memory limits does not apply locally unless the `--compatibility` flag is used. Since the latest update this flag now causes the error `unknown flag: --compatibility`. I am using `docker-compose` and not `docker compose`

- [x] Using Compose V2 `docker compose ...`
- [ ] Using Compose V1 `docker-compose ...`

**Output of `docker(-)compose version`**
```
Docker Compose version v2.0.0-rc.2
```

```
Client:
 Cloud integration: 1.0.17
 Version:           20.10.8
 API version:       1.41
 Go version:        go1.16.6
 Git commit:        3967b7d
 Built:             Fri Jul 30 19:55:20 2021
 Context:           default
 Experimental:      true

 Engine:
  Version:          20.10.8
  API version:      1.41 (minimum version 1.12)
  Go version:       go1.16.6
  Git commit:       75249d8
  Built:            Fri Jul 30 19:52:10 2021
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.4.9
  GitCommit:        e25210fe30a0a703442421b0f60afac609f950a3
 runc:
  Version:          1.0.1
  GitCommit:        v1.0.1-0-g4144b63
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

```
It happens with all of my compose files
```

1. Attempt to `up` any compose file using `docker-compose --compatibility`

### Observed result

The command fails with `unknown flag: --compatibility`

### Expected result

The command succeeds, mapping the memory limits into the v2 compose format

```
unknown flag: --compatibility
```

## Additional information

macOS Big Sur 11.5.2, Docker Desktop 4.0.0

ndeloof:
Introduce no-op --compatibility flag

Resolves #8618

### DIFF
--- a/cmd/compose/compose.go
+++ b/cmd/compose/compose.go
@@ -148,7 +148,9 @@ func (o *projectOptions) addProjectFlags(f *pflag.FlagSet) {
 	f.StringVar(&o.EnvFile, "env-file", "", "Specify an alternate environment file.")
 	f.StringVar(&o.ProjectDir, "project-directory", "", "Specify an alternate working directory\n(default: the path of the Compose file)")
 	f.StringVar(&o.WorkDir, "workdir", "", "DEPRECATED! USE --project-directory INSTEAD.\nSpecify an alternate working directory\n(default: the path of the Compose file)")
+	f.Bool("compatibility", false, "DEPRECATED")
 	_ = f.MarkHidden("workdir")
+	_ = f.MarkHidden("compatibility")
 }
 
 func (o *projectOptions) toProjectName() (string, error) {


cmd/compose/compose.go
```diff
@@ -142,13 +142,16 @@ func (o *projectOptions) WithServices(fn ProjectServicesFunc) func(cmd *cobra.Co
...
 
 func (o *projectOptions) addProjectFlags(f *pflag.FlagSet) {
+	var compatibility bool
 	f.StringArrayVar(&o.Profiles, "profile", []string{}, "Specify a profile to enable")
 	f.StringVarP(&o.ProjectName, "project-name", "p", "", "Project name")
 	f.StringArrayVarP(&o.ConfigPaths, "file", "f", []string{}, "Compose configuration files")
 	f.StringVar(&o.EnvFile, "env-file", "", "Specify an alternate environment file.")
 	f.StringVar(&o.ProjectDir, "project-directory", "", "Specify an alternate working directory\n(default: the path of the Compose file)")
 	f.StringVar(&o.WorkDir, "workdir", "", "DEPRECATED! USE --project-directory INSTEAD.\nSpecify an alternate working directory\n(default: the path of the Compose file)")
+	f.BoolVar(&compatibility, "compatibility", false, "DEPRECATED")
```
mat007, line 152: I think it could just be
```suggestion
	_ = f.Bool("compatibility", false, "DEPRECATED")
```

mat007 (approved):
I wonder if this should not rather be in the compose switch? 🤔 
Anyway, it’s harmless and helps users so SGTM 😁

Issue #8540, opened 2021-05-28
Force stopping a single `docker compose up <service>` kills all dependent services


enisdenjo:
**Description**

Starting a single service with `docker compose up <service>` and then force stopping it with `Ctrl-C + Ctrl-C` kills **all dependent** running services. However, waiting for the graceful stop (with a single `Ctrl-C`) stops just that single service.

**Steps to reproduce the issue:**
1. `banana` depends on `chocolate` in the `docker-compose.yml`
1. 1st terminal `docker compose up chocolate`
1. _Running..._
1. 2nd terminal `docker compose up banana`
1. _Running..._
1. `Ctrl-C` in 2nd terminal (banana)
1. `Ctrl-C` (force stop) in 2nd terminal (banana)

**Describe the results you received:**

Both services `chocolate` and `banana` get killed.

**Describe the results you expected:**

Only `banana` should get killed.

**Additional information you deem important (e.g. issue happens only occasionally):**

This is **not** how the legacy `docker-compose` worked. With it, force stopping a single service, forcibly kills just that service regardless of the dependency tree.

**Output of `docker version`:**

```
Client:
 Cloud integration: 1.0.14
 Version:           20.10.6
 API version:       1.41
 Go version:        go1.16.3
 Git commit:        370c289
 Built:             Fri Apr  9 22:46:57 2021
 OS/Arch:           darwin/amd64
 Context:           default
 Experimental:      true

Server: Docker Engine - Community
 Engine:
  Version:          20.10.6
  API version:      1.41 (minimum version 1.12)
  Go version:       go1.13.15
  Git commit:       8728dd2
  Built:            Fri Apr  9 22:44:56 2021
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.4.4
  GitCommit:        05f951a3781f4f2c1911b05e61c160e9c30eaa8e
 runc:
  Version:          1.0.0-rc93
  GitCommit:        12644e614e25b05da6fd08a38ffa0cfe1903fdec
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

**Output of `docker info`:**

```
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  app: Docker App (Docker Inc., v0.9.1-beta3)
  buildx: Build with BuildKit (Docker Inc., v0.5.1-docker)
  compose: Docker Compose (Docker Inc., 2.0.0-beta.1)
[...]

PR #8638, opened 2021-09-20
Stop only the service ran by it's up command

ulyssessouza:
Resolves https://github.com/docker/compose/issues/8540


ndeloof (approved):

luxalpa:
This makes it so that hitting ctrl+c does not stop the containers that it booted up if they are in the dependsOn. Is that intended? Because it's very annoying...

PR #8655, opened 2021-09-22
define compose container names with a valid hostname

### LINKED ISSUE #229: Switch to using hyphens as a separator in hostnames

ndeloof:
This is a proposal to fix long running issue https://github.com/docker/compose/issues/229

tl;dr:
docker-compose creates containers as `project_service_number` which make them invalid hostnames. Most tools support this mistakes, but some don't.

As we move to Compose V2, we have a unique opportunity to fix this with a minor backward compatibility break. Maybe we could have this controlled by some COMPOSE_COMPATIBILITY env variable, and/or resurrect the --compatibility flag

Resolves #229

### DIFF
--- a/cmd/compose/compose.go
+++ b/cmd/compose/compose.go
@@ -91,12 +91,13 @@ func Adapt(fn Command) func(cmd *cobra.Command, args []string) error {
 var Warning string
 
 type projectOptions struct {
-	ProjectName string
-	Profiles    []string
-	ConfigPaths []string
-	WorkDir     string
-	ProjectDir  string
-	EnvFile     string
+	ProjectName   string
+	Profiles      []string
+	ConfigPaths   []string
+	WorkDir       string
+	ProjectDir    string
+	EnvFile       string
+	Compatibility bool
 }
 
 // ProjectFunc does stuff within a types.Project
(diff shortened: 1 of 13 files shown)


thaJeztah:
wondering; would it help with the transition if we would set the "old" name as an (network) alias?

Also possibly related if we'll be changing conventions; https://github.com/docker/swarmkit/issues/1242

ulyssessouza (commented):
Maybe just add a test with the `--compatibility`

ulyssessouza (approved):
LGTM

ndeloof:
> would it help with the transition if we would set the "old" name as an (network) alias?

I don't think this would be very useful, as recommended way to access sibling container is to use service name, not plain container name anyway.

PR #8656, opened 2021-09-22
"log" must stop watching on SIGTERM

ndeloof:
while "start" must not interupt the watch process, so need a background context,
"log" must stop on SIGTERM and as such watch operation must use the top-level cancelable context.
(otherwise `docker compose logs --follow` never stops)

this is a combination of https://github.com/docker/compose/commit/1458beea84fbd8b7aa9757bde025fc210cf545b5 and https://github.com/docker/compose/commit/f7c360b721b1e2a0bb1eb78b60cd60bb1c1c6199


mat007 (approved):
I think it makes sense!

lorenrh (approved):

PR #8669, opened 2021-09-24
parse security opts to process seccomp file as relative to project dir

ndeloof:
This address https://github.com/docker/compose/issues/8470 on Compose V2


ulyssessouza (approved):
LGTM

Maybe we could propose something like a 'p' -> 'P' change on `docker/cli`

Issue #8683, opened 2021-09-28
Checksum missing for 2.0.0 release


aminvakil:
**Description**

https://github.com/docker/compose/releases/download/2.0.0/docker-compose-Linux-x86_64.sha256 is missing.

This (`docker-compose-Linux-x86_64.sha256`) used to be the filename though, maybe this needs to be changed to `docker-compose-linux-amd64.sha256` from this release now on.

However that's missing too.

**Steps to reproduce the issue:**
1. Try getting checksums for 2.0.0 release.

**Describe the results you received:**
Missing :)

**Describe the results you expected:**
Available :)

ndeloof:
I fixed this manually for 2.0.0 release and proposed https://github.com/docker/compose/pull/8685 to get this generated automatically during release process

Vinfall:
And please do use consistent release version pattern, it should be `2.0.0` not `v2.0.0`. My script breaks today due to the change (along with the change from `*-Linux-x86_64` to `*-linux-amd64` ofc, but this is understandable).

> https://github.com/docker/compose/releases/download/2.0.0/docker-compose-Linux-x86_64.sha256
> https://github.com/docker/compose/releases/download/v2.0.0/docker-compose-linux-amd64.sha256

These two are quite different.

BTW, the current `sha256sum` is actually `sha1sum` XD.

```
$: cat docker-compose-linux-amd64.sha256
2316b5e34fd4e30cbf41ed1cdc0056e7f459446f  toto/docker-compose-linux-amd64
$: sha1sum docker-compose-linux-amd64
2316b5e34fd4e30cbf41ed1cdc0056e7f459446f  docker-compose-linux-amd64
```

ndeloof:
`v2.0.0` is a distinct, Golang codebase and so applies related conventions.
Installation is not the same as docker-compose V1 (V2 is a CLI plugin), so your script probably will need to be updated anyway.

I fixed the checksum oin v2.0.0 - this is what happen when not using an automated process :'(

PR #8685, opened 2021-09-28
compute sha256 checksums while releasing

ndeloof:
**Related issue**
https://github.com/docker/compose/issues/8683


lorenrh (approved):

aminvakil (approved):
Thanks for fixing this so quick!

PR #8689, opened 2021-09-28
Fix compose up in readme

mat007:
**What I did**

Replaced `docker-compose` with `docker compose` in readme for V2.

**Related issue**


ulyssessouza (approved):
LGTM

Issue #8700, opened 2021-09-29
Multiple `--status` options for `docker compose ps` do not work


matt9ucci:
**Description**

Multiple `--status` options for  `docker compose ps` do not work as I expected.

**Describe the results you received:**

All services that matches the last one of the `--status` options are shown.

For example, there are three services:

```
$ docker compose ps
NAME                      COMMAND                  SERVICE             STATUS              PORTS
test-redis-1-1   "docker-entrypoint.s…"   redis-1             exited (0)
test-redis-2-1   "docker-entrypoint.s…"   redis-2             running             6379/tcp
test-web-1       "flask run"              web                 running             0.0.0.0:5000->5000/tcp, :::5000->5000/tcp
```

If `--status running` is set, two running services are shown as expected:

```
$ docker compose ps --status running
NAME                      COMMAND                  SERVICE             STATUS              PORTS
test-redis-2-1   "docker-entrypoint.s…"   redis-2             running             6379/tcp
test-web-1       "flask run"              web                 running             0.0.0.0:5000->5000/tcp, :::5000->5000/tcp
```

But if `--status running --status exited` is set, one exited services is shown, and two running are hidden:

```
$ docker compose ps --status running --status exited
NAME                      COMMAND                  SERVICE             STATUS              PORTS
test-redis-1-1   "docker-entrypoint.s…"   redis-1             exited (0)
```

**Describe the results you expected:**

All services that matches one of the `--status` options are shown.

```
$ docker compose ps --status running --status exited
NAME                      COMMAND                  SERVICE             STATUS              PORTS
test-redis-1-1   "docker-entrypoint.s…"   redis-1             exited (0)
test-redis-2-1   "docker-entrypoint.s…"   redis-2             running             6379/tcp
test-web-1       "flask run"              web                 running             0.0.0.0:5000->5000/tcp, :::5000->5000/tcp
```

**Output of `docker compose version`:**

```
Docker Compose version v2.0.0
```

PR #8702, opened 2021-09-29
make --status a multi-value flag

ndeloof:
**What I did**
allow use of multiple --status filters on compose ps

**Related issue**
close https://github.com/docker/compose/issues/8700


ulyssessouza (approved):
LGTM

Issue #8704, opened 2021-09-29
Released 2.0.0 docker-compose-linux-* isn't docker-compose but the compose plugin


rfay:
**Description**

Yesterday's release of compose 2.0.0 (congratulations!) broke a lot of things for Linux users who dare to download it, but most especially for ArchLinux users who got it by default as docker-compose.

And of course on Arch it was immediately pushed out as v2.0.0 of docker-compose, which broke all docker-compose users, because it's not a standalone binary. 

Neither the release notes nor the [docs](https://github.com/docker/compose#linux) are very clear about what to do. 

1. You'll want to engage with the ArchLinux community to get something workable going for them.
2. Please make a complete doc that just says "How do I get docker-compose on Linux". Perhaps the Arch folks will then be able to implement that in their package?
3. Please use better naming for the binary. The current naming is `docker-compose-linux-amd64`, which implies that it's docker-compose for Linux, which it certainly is not. Maybe `compose-plugin-for-docker-linux-amd64` would be clearer?

siliconmeadow:
### Arch Linux workarounds:

1. Use the AUR `downgrade` [tool](https://aur.archlinux.org/packages/downgrade)
2. If you do stick with v2.0.0 in Arch, at the moment you could add a symlink to `docker-compose` as it's no longer in the common paths in Arch. Something like:

nantkowiak:
1. you can use `docker compose` instead of `docker-compose`
2. you can define an alias `alias docker-compose='docker compose'` to use it like before
3. downgrade with `pacman -U /var/cache/pacman/pkg/docker-compose-{your preffered version}`

ndeloof:
duplicates https://github.com/docker/compose/issues/8698

ndeloof:
My preferred option to workaround this issue is that we make compose CLI plugin act as a plain `docker-compose` executable, which seems feasible. cc @ulyssessouza @justincormack

Foxboron:
@ndeloof Can I get a positive or negative confirmation if that is going to be the plan or not? It affects the packaging decisions downstream and might complicate a few things.

ndeloof:
@Foxboron @ulyssessouza is working on it

ndeloof:
[...]

PR #8708, opened 2021-09-29
Add standalone capacility

ulyssessouza:
Signed-off-by: Ulysses Souza <ulyssessouza@gmail.com>

**What I did**
Add the capability of running as a standalone program

**Related issue**
Addresses https://github.com/docker/compose/issues/8704#issuecomment-930372585

**(not mandatory) A picture of a cute animal, if possible in relation with what you did*
goat
*

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -17,16 +17,15 @@
 package main
 
 import (
-	"fmt"
 	"os"
 
 	dockercli "github.com/docker/cli/cli"
 	"github.com/docker/cli/cli-plugins/manager"
 	"github.com/docker/cli/cli-plugins/plugin"
 	"github.com/docker/cli/cli/command"
+	"github.com/docker/compose-switch/redirect"
 	"github.com/spf13/cobra"
 
-	cliFlags "github.com/docker/cli/cli/flags"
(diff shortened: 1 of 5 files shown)


ulyssessouza:
The only difference so far is the output of `docker-compose --help` compared to `docker compose --help`.
Plugin manager changing it?

ndeloof:
I wonder you need to include `convert` from compose-switch to only pass docker cli flags when initializing the CLI and keep others for compose root command

cmd/main.go
```diff
@@ -34,7 +38,26 @@ func init() {
 		"To provide feedback or request new features please open issues at https://github.com/docker/compose"
 }
 
-func main() {
+func standaloneMain() int {
+	dockerCli, err := command.NewDockerCli()
```
ndeloof, line 42:
```console
$ docker-compose --host tcp://localhost:2345 up
Error: unknown flag: --host
```
need to first [Convert](https://github.com/docker/compose-switch/blob/master/redirect/convert.go#L27) the command line to be cli-plugin compliant

ndeloof (changes requested):

ndeloof (approved):

Issue #8695, opened 2021-09-28
New release filenames breaks some scripts and docs


CrimsonGlory:
This is not a bug in compose. But a note on the release filenames.

New release "v2.0.0" have a few changes in the format used for the release and names:
- "v2.0.0" includes a "v", so in the [official docs](https://docs.docker.com/compose/install/) where it says
```
To install a different version of Compose, substitute 1.29.2 with the version of Compose you want to use.
```

In this case ```1.29.2``` should be substituted for ```v2.0.0``` and not for ```2.0.0```.

Also the current name format ```docker-compose-linux-amd64``` differs from the previous one ```docker-compose-Linux-x86_64```.

The lowercase ```L``` in ```linux``` doesn't seem to be a problem because the link seems case insensitive. But the ```amd64``` instead of ```x86_64``` will break the script from the official documentation and [current CI scripts](https://github.com/CircleCI-Public/docker-orb/issues/94) which uses ```$(uname -m)```:

```
 sudo curl -L "https://github.com/docker/compose/releases/download/1.29.2/docker-compose-$(uname -s)-$(uname -m)" -o /usr/local/bin/docker-compose
```

Renaming the files and version name will be easier than fixing documentation, blogs, and scripts all over the place.

mat007:
Which official documentation are you referring to?
The installation instructions for Compose V2 differ from V1. V2 is not a standalone binary anymore and scripts will have to be adjusted anyway.
See https://github.com/docker/compose#where-to-get-docker-compose

Also compose V2 will be distributed with the docker CLI ultimately and won’t require to be installed separately.

ndeloof:
We used `v2.0.0` to comform with Golang requirements, as the new Compose V2 codebase is implemented in this language

dzmitrykliapkou:
Also there was a possibility to install the latest docker-compose with Ansible using the following download link:
https://github.com/docker/compose/releases/latest/download/docker-compose-{{ ansible_system }}-{{ ansible_architecture }}

But now it does not work due to replacement of `x86_64` with `amd64`

ndeloof:
This `amd64` string is indeed the suffix used by Golang for platform architecture, which isn't the same as the one reported by `uname -m`. That's a bit sad they didn't adopted something standard.

ndeloof:
we probably could adopt more "standard" suffixes on https://github.com/docker/compose/blob/v2/builder.Makefile#L49-L56
`amd64` -> `x86_64`
[...]

PR #8710, opened 2021-09-30
{Proposal} use `uname -m` for cross platform suffixes

ndeloof:
**What I did**
Adopted output for `uname -m` as cross-platform suffixes so it is easier to design download scripts

**Related issue**
https://github.com/docker/compose/issues/8695#issuecomment-930883155

### DIFF
--- a/builder.Makefile
+++ b/builder.Makefile
@@ -46,14 +46,14 @@ compose-plugin:
 
 .PHONY: cross
 cross:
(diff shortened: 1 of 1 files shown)


mat007:
Is this not going to make it more painful for us to integrate this in our build pipeline for the CLI and such?

ndeloof:
@mat007 the CLI system packages build from sources
For Desktop packaging we indeed rely on go arch const AFAIK. This unfortunately is a gopherism that isn't portable.

ulyssessouza (approved):
LGTM

Issue #8711, opened 2021-09-30
`--status` option combined with `--services` does not work


matt9ucci:
**Description**

`docker compose ps --status` works.
But if combined with `--services` option, it does not work.

**Describe the results you received:**

There are three containers:

```
$ docker compose ps
NAME                COMMAND                  SERVICE             STATUS              PORTS
test-redis-1-1      "docker-entrypoint.s…"   redis-1             exited (0)
test-redis-2-1      "docker-entrypoint.s…"   redis-2             running             6379/tcp
test-web-1          "flask run"              web                 running             0.0.0.0:5000->5000/tcp, :::5000->5000/tcp
```

`--status exited` works:

```
$ docker compose ps --status exited
NAME                COMMAND                  SERVICE             STATUS              PORTS
test-redis-1-1      "docker-entrypoint.s…"   redis-1             exited (0)
```

But if combined with `--services` option, it shows exited/running containers' service names:

```
$ docker compose ps --status exited --services
redis-1
redis-2
web
```

**Describe the results you expected:**

It shows only exited container's service name:

```
$ docker compose ps --status exited --services
redis-1
```

**Output of `docker compose version`:**

```
Docker Compose version v2.0.0
```

PR #8712, opened 2021-09-30
allow combination of --status and --services

ndeloof:
**What I did**
allow combination of --status and --services on compose ps command

**Related issue**
close https://github.com/docker/compose/issues/8711


ulyssessouza (approved):
LGTM

Issue #8715, opened 2021-09-30
Regression from v1.x: `docker compose logs -f <containername>` no longer exits when the container exits.


djpbessems:
**Description**
`docker compose logs -f <containername>` no longer exits when the container exits.

**Steps to reproduce the issue:**
1. `docker compose up -d` (with a container that is short-lived and is expected to exit)
2. `docker compose logs -f <containername>`
3.

**Describe the results you received:**
The entirety of the logs are shown, but you are not returned to the shell and have to press CTRL-C

**Describe the results you expected:**
The entirety of the logs are shown, and you are then dropped back to the shell

**Additional information you deem important (e.g. issue happens only occasionally):**
This issue prevents capturing an entire set of logs in a pipeline.

**Output of `docker compose version`:**

```
Docker Compose version v2.0.0
```

**Output of `docker info`:**

```
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  app: Docker App (Docker Inc., v0.9.1-beta3)
  buildx: Build with BuildKit (Docker Inc., v0.6.1-docker)
  compose: Docker Compose (Docker Inc., v2.0.0)
  scan: Docker Scan (Docker Inc., v0.8.0)

Server:
 Containers: 13
  Running: 7
  Paused: 0
  Stopped: 6
 Images: 12
 Server Version: 20.10.8
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 1
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runtime.v1.linux runc io.containerd.runc.v2
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: e25210fe30a0a703442421b0f60afac609f950a3
 runc version: v1.0.1-0-g4144b63
 init version: de40ad0
 Security Options:
  apparmor
  seccomp
   Profile: default
 Kernel Version: 5.4.0-88-generic
 Operating System: Ubuntu 20.04.3 LTS
 OSType: linux
 Architecture: x86_64
 CPUs: 2
 Total Memory: 3.844GiB
 Name: packer-template
[...]

PR #8726, opened 2021-10-01
`log --follow` must stop when container get killed

ndeloof:
**What I did**
`compose logs --follow` never complete if container gets killed
code missed running the logPrinter so it actually react on container events

#close https://github.com/docker/compose/issues/8715

image


thaJeztah:
hmf.. github is having issues (can't comment on the diff);

note that a `kill` event status doesn't necessarily means the container will exit (e.g. `docker kill --signal=SIGHUP <container>` also sends a `kill` event, but the container is still running

ndeloof:
@ulyssessouza ?

ulyssessouza (approved):
Lgtm

PR #8727, opened 2021-10-01
set EnvironmentFileLabel label after ServiceHash

### LINKED ISSUE #8725: All containers are recreated needlessly if project environment (`--env-file`) path changes
**Description**
We use the ``env-file`` option in docker compose (i.e. `docker compose --env-file foo.env up`).
In our use case, the path of the environment file changes on each deployment.

This worked without issue with the old python based `docker-compose`. With the new V2 all containers get recreated, even if there there are effectively no changes in the configuration.

1. Use any `docker-compose.yml` file, for example
```
version: "2"
services:
  wait:
    image: alpine
    command: [ "sleep", "100000" ]
```
2. Create some environment files
```
touch env1.env
touch env2.env
touch env3.env
```
3. ``docker-compose --env-file env1.env up``
4. ``docker-compose --env-file env2.env up``
5. ``docker-compose --env-file env3.env up``

In step 4 and 5 the container is recreated.

In step 4 and 5 the running container is untouched.

```
Docker Compose version v2.0.1
```
**Output of `docker info`:**

```
Client:
 Context:    default
 Debug Mode: false

Server:
 Containers: 24
  Running: 17
  Paused: 0
  Stopped: 7
 Images: 28
 Server Version: 20.10.7
 Storage Driver: zfs
  Zpool: rpool
  Zpool Health: ONLINE
  Parent Dataset: rpool/var/lib/docker
  Space Used By Parent: 24915320832
  Space Available: 335626928128
  Parent Quota: no
  Compression: lz4
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 1
 Plugins:
  Volume: local
 Swarm: inactive
 Default Runtime: runc
 Init Binary: docker-init
 containerd version:
 runc version:
 init version:
 Security Options:
  apparmor
  seccomp
   Profile: default
 Kernel Version: 5.4.0-80-generic
 Operating System: Ubuntu 20.04.3 LTS
 OSType: linux
 Architecture: x86_64
 CPUs: 4
 Total Memory: 31.22GiB
 Name: steven
 ID: QDX4:POZT:ZURM:DEMC:OTG5:3GBC:SCQK:AHFN:2JDJ:RORJ:X2VY:SI6X
 Debug Mode: false
 Labels:
 Experimental: false
 Insecure Registries:
  127.0.0.0/8
 Live Restore Enabled: false
```

joelbohnes:
**What I did**
In order to fix #8725, we need to not include the EnvironmentFileLabel in the hash computation.
Note that the other file system path labels (WorkingDirLabel and ConfigFilesLabel) are already set after hash computation, so path changes are not an issue there. This patch brings the behaviour of EnvironmentFileLabel inline with other file system paths.

The project environment is usually set directly inside `WithServices` of `cmd/compose/compse.go`. But this needs
to be delayed until `prepareLabels` in `pkg/compose/create.go`. The easiest (but perhaps not cleanest) way
was to pass down the variable inside an extension field. Both project or service extension would work, this patch
uses a service extension field.

**Related issue**
fixes #8725

### DIFF
--- a/cmd/compose/compose.go
+++ b/cmd/compose/compose.go
@@ -122,19 +122,20 @@ func (o *projectOptions) WithServices(fn ProjectServicesFunc) func(cmd *cobra.Co
 
 		if o.EnvFile != "" {
 			var services types.Services
-			for _, s := range project.Services {
-				ef := o.EnvFile
-				if ef != "" {
-					if !filepath.IsAbs(ef) {
-						ef = filepath.Join(project.WorkingDir, o.EnvFile)
-					}
-					if s.Labels == nil {
-						s.Labels = make(map[string]string)
-					}
-					s.Labels[api.EnvironmentFileLabel] = ef
-					services = append(services, s)
-				}
+
+			ef := o.EnvFile
+			if !filepath.IsAbs(ef) {
+				ef = filepath.Join(project.WorkingDir, o.EnvFile)
 			}
+
+			for _, s := range project.Services {
+				if s.Extensions == nil {
+					s.Extensions = make(map[string]interface{})
+				}
+				s.Extensions[compose.ServiceExtProjectEnvironmentFile] = ef
+				services = append(services, s)
+			}
+
 			project.Services = services
 		}
 
--- a/cmd/compose/compose.go
+++ b/cmd/compose/compose.go
@@ -121,22 +121,16 @@ func (o *projectOptions) WithServices(fn ProjectServicesFunc) func(cmd *cobra.Co
 		}
 
 		if o.EnvFile != "" {
-			var services types.Services
-
 			ef := o.EnvFile
 			if !filepath.IsAbs(ef) {
 				ef = filepath.Join(project.WorkingDir, o.EnvFile)
 			}
 
-			for _, s := range project.Services {
-				if s.Extensions == nil {
-					s.Extensions = make(map[string]interface{})
-				}
-				s.Extensions[compose.ServiceExtProjectEnvironmentFile] = ef
-				services = append(services, s)
+			if project.Extensions == nil {
+				project.Extensions = make(map[string]interface{})
 			}
 
-			project.Services = services
+			project.Extensions[compose.ExtProjectEnvironmentFile] = ef
 		}
 
 		return fn(ctx, project, args)
--- a/pkg/compose/create.go
+++ b/pkg/compose/create.go
@@ -46,6 +46,9 @@ import (
 	"github.com/docker/compose/v2/pkg/utils"
 )
 
+// ServiceExtProjectEnvironmentFile is name of service extension that is used to pass down the project environment file
+const ServiceExtProjectEnvironmentFile = "x-dockercompose-projectenvironmentfile"
+
 func (s *composeService) Create(ctx context.Context, project *types.Project, options api.CreateOptions) error {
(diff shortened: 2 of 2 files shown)


cmd/compose/compose.go
```diff
@@ -122,19 +122,20 @@ func (o *projectOptions) WithServices(fn ProjectServicesFunc) func(cmd *cobra.Co
...
-					}
-					if s.Labels == nil {
-						s.Labels = make(map[string]string)
-					}
-					s.Labels[api.EnvironmentFileLabel] = ef
-					services = append(services, s)
+				if s.Extensions == nil {
+					s.Extensions = make(map[string]interface{})
 				}
+				s.Extensions[compose.ServiceExtProjectEnvironmentFile] = ef
```
ndeloof, line 135: could be attached directly to project as this applies to all services

ndeloof (changes requested):
I'm not a big fan using a fake extension to pass hidden parameters
could use `context.WithValue()` for this purpose

OR you can just make `ServiceHash` exclude any `com.docker.compose.*` label from hash computation are those are not tied to configuration.

joelbohnes:
I don't like the fake extension either. I think `context.WithValue` is not a clean way to pass parameters down, at least with the extension method the value is directly associated with the compose project.

Adapting `ServiceHash` seems viable, would require some custom serialization of the labels array. It seems a bit excessive, we'd only require this functionality for a single label (all other `com.docker.compose.*` labels are already set after hash computation, there is no need to adapt `SeriveHash` for them).

Another option may be to add a new field in the Project structure (in https://github.com/compose-spec/compose-go/blob/master/types/project.go). Setting the field could be done in `ProjectFromOptions` or `toProject`. I would personally prefer this option, but it requires a change in another project.

ndeloof:
> Adapting ServiceHash seems viable, would require some custom serialization of the labels array

Nope, `ServiceHash` get a copy of the ServiceConfig struct, so you can just drop all labels that start with `com.docker.compose.`

joelbohnes:
> Nope, `ServiceHash` get a copy of the ServiceConfig struct, so you can just drop all labels that start with `com.docker.compose.`

Maps are never implicitly copied (same with arrays). The ServiceConfig struct is copied, but the `Labels` member is effectively a pointer which is copied, not the whole map.

What you can do safely is to assign a the `Labels` member to a new map, like so:
```
newlabels := make(map[string]string)
for k, v := range o.Labels {
  if strings.HasPrefix(k, "com.docker.compose.") {
    newlabels[k] = v
  }
}
o.Labels = newmap
```

The comment of the `ServiceHash` function indicates that this function should eventually be moved to `compose-go`. So I don't think it's a good idea to have vendor specific label-removal functionality inside the function.

ndeloof:
> The comment of the ServiceHash function indicates that this function should eventually be moved to compose-go

I indeed added this comment as "would be nice ..." but using hash to compare actual vs requested state is very specific to docker-compose, so that's not an issue.
Also note `com.docker.compose` labels are already referenced in the spec, so don't get worried about those.

ndeloof:
I created https://github.com/docker/compose/pull/8960 as an alternative approach to this issue, I welcome your review

ndeloof:
closing as https://github.com/docker/compose/pull/8960 has been merged

Issue #8535, opened 2021-06-16
Compose v2 fails with devices: "/dev/fuse", requires devices: "/dev/fuse:/dev/fuse"


jamshid:
**Description**

Slight incompatibility with devices in `docker compose` vs `docker-compose`.

**Steps to reproduce the issue:**
1. I have a service where I can run fuse in the container, so `rclone mount` works.
```
...
  client:
    restart: "${PROJECT_RESTART-no}"
    image: "${REGISTRY_URL}client:${CLIENT_IMAGE_TAG}"
    init: true
    deploy:
      resources:
        limits:
          memory: ${CLIENT_CONTAINER_MEM_LIMIT-2g}
    # Needed for mounting fuse filesystem (https://github.com/docker/docker/issues/16233)
    cap_add:
      - SYS_ADMIN
      - MKNOD
    devices:
      - "/dev/fuse"
    security_opt:
      - apparmor:unconfined
      # for strace, https://github.com/docker/docker/issues/7276#issuecomment-194137403
      - "seccomp:unconfined"
```
2. But fuse fails with `docker compose`, `rclone mount` reports `fuse: device not found, try 'modprobe fuse' first`. It has worked for years with `docker-compose`.
3. Change yml to this and it works.
```
    devices:
      - "/dev/fuse:/dev/fuse"
```

**Describe the results you received:**
above

**Describe the results you expected:**
above

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `docker version`:**

```
 Version:           20.10.7

% docker compose version
Docker Compose version 2.0.0-beta.3
```

**Additional environment details (AWS ECS, Azure ACI, local, etc.):**
latest docker for mac 3.4.0

zikaeroh:
A Sourcegraph search shows that this form isn't all that uncommon: https://sourcegraph.com/search?q=context:global+f:docker-compose.yml+devices:%5Cn%5Cs%2B-%5B%5E:%5D%2B%24+lang:yaml+&patternType=regexp

I hit this too, though with `/dev/net/tun` for a VPN proxy container.

PR #8732, opened 2021-10-02
fix support for devices

ndeloof:
**What I did**
copy/paste was missing a few lines https://github.com/docker/cli/blob/master/cli/command/container/opts.go#L934-L936

longer term fix would make https://github.com/docker/cli/blob/master/cli/command/container/opts.go#L900 public and just reuse

**Related issue**
https://github.com/docker/compose/issues/8535

image


mat007 (approved):
Fair enough!

ulyssessouza (approved):
Lgtm

mat007:
Does this entirely fix #8535?

PR #8735, opened 2021-10-02
:zap: Added a handy command to dynamically fetch the latest version of compose and install it on Linux.

Harmouch101:
Signed-off-by: Harmouch101 <mahmoudddharmouchhh@gmail.com>

**What I did**
Consider reading the title which is self-explanatory of what I did.

```bash
# create the docker plugins directory
mkdir -p ~/.docker/cli-plugins

# download the latest version of the CLI into the plugins directory and rename it to docker-compose with wget.
# the regex "https.*docker-compose-$(uname | awk '{print tolower($0)}')-$(uname -m)$" would match
#  the name of the binary that matches the one intended to be downloaded on the currently running distro.
# ex: on Ubuntu x86, the regex matches: docker-compose-linux-x86_64, and so on...
curl -SL https://api.github.com/repos/docker/compose/releases/latest | grep browser_download_url | cut -d '"' -f 4 | grep -wo https.*docker-compose-$(uname | awk '{print tolower($0)}')-$(uname -m)$ | wget -i - -O ~/.docker/cli-plugins/docker-compose

# make the CLI executable
chmod +x ~/.docker/cli-plugins/docker-compose
```

Combining the previous commands would result in the following:

```bash
mkdir -p ~/.docker/cli-plugins && curl -SL https://api.github.com/repos/docker/compose/releases/latest | grep browser_download_url | cut -d '"' -f 4 | grep -wo https.*docker-compose-$(uname | awk '{print tolower($0)}')-$(uname -m)$ | wget -i - -O ~/.docker/cli-plugins/docker-compose && chmod +x ~/.docker/cli-plugins/docker-compose
```

**Related issue**

An enhancement rather than a bug fix.
:cat2:

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,18 @@ Or copy it into one of these folders for installing it system-wide:
 
 (might require to make the downloaded file executable with `chmod +x`)
 
+Or you can use the following command, which has been tested on ubuntu 20.04, to dynamically pull the latest version of `docker compose` and install it on your machine:
+
+```bash
+mkdir -p ~/.docker/cli-plugins && curl -SL https://api.github.com/repos/docker/compose/releases/latest | grep browser_download_url | cut -d '"' -f 4 | grep -wo https.*docker-compose-$(uname | awk '{print tolower($0)}')-$(uname -m)$ | wget -i - -O ~/.docker/cli-plugins/docker-compose && chmod +x ~/.docker/cli-plugins/docker-compose
+```
+
+You can verify that `docker compose` has been installed by executing the following command on your terminal:
+
+```bash
+docker compose version
+> Docker Compose version v2.0.1
+```
 
 Quick Start
 -----------


page-down:
Since there is no version number in the release file name, you can just use the latest release link that GitHub officially supports.

[Linking to releases - GitHub Docs](https://docs.github.com/en/repositories/releasing-projects-on-github/linking-to-releases)
> To link directly to a download of your latest release asset that was manually uploaded, link to /owner/name/releases/latest/download/asset-name.zip

Note that the `latest` in GitHub Releases may not be what you want, for example when the latest patch revision of an older major or minor version has been released. It would be better to specify the version explicitly.

curl supports `--create-dirs` since 2003. And there is no need to use both curl and wget.

The following commands works for Linux and macOS. However it is not recommended to install docker compose blindly.

```shell
sudo curl -fSL "https://github.com/docker/compose/releases/latest/download/docker-compose-$(uname -s|tr DL dl)-$(uname -m)" --create-dirs -o /usr/local/lib/docker/cli-plugins/docker-compose && sudo chmod +x /usr/local/lib/docker/cli-plugins/docker-compose
```

Refer to Docker Docs:
https://github.com/docker/docker.github.io/blob/master/compose/install.md

Also, if you hardcode the installation to `~/.docker/cli-plugins` without checking the `DOCKER_CONFIG` environment variable, it may not work.

Harmouch101:
Hey @thaJeztah, I just want to let you know that I made the changes requested a while ago in [4163e73](https://github.com/docker/compose/pull/8735/commits/4163e73de4356fcdf818c3afb97d71ef34ba04d3). Kindly let me know if there is any blocker to resolve from my side.

ndeloof:
closing as obsolete

PR #8738, opened 2021-10-03
Restore missing version commands

### LINKED ISSUE #8729: docker-compose v2.0.1 Missing "version" in --help
Although `docker-compose version` outputs the version as expected, it is missing from the --help output.

In docker-compose v1:

```
[..]
  top                Display the running processes
```

In docker-compose v2.0.1:-

```
[..]
  start       Start services
  stop        Stop services
  top         Display the running processes
  unpause     unpause services
  up          Create and start containers

Run 'docker compose COMMAND --help' for more information on a command.
```

Shikachuu:
**What I did**
- Removed the hidden flag from the `version` command, since by default and for historical reasons aswell it shouldn't be hidden.
- Since the documentation for the v2 still constantly reffering to a `--version` flag on the root command, I added it aswell.
_(This is up for a discuisson, since I think this change shouldn't be exist and we have to correct the documentation.)
(~If we are still planning to keep this change, I think it could be done more elegantly.~
I did it aswell, since I forgot that the old python version also handle the `-v` flag next to `--version`)_

**Related issue**
Fix #8729

**A picture of a cute animal, if possible in relation with what you did**
_(A hiding cat, reffering to the hidden version command.)_
hidden_cat

### DIFF
--- a/cmd/compose/compose.go
+++ b/cmd/compose/compose.go
@@ -222,6 +222,7 @@ func RootCommand(backend api.Service) *cobra.Command {
 		ansi    string
 		noAnsi  bool
 		verbose bool
+		version bool
 	)
 	command := &cobra.Command{
 		Short:            "Docker Compose",
@@ -231,6 +232,9 @@ func RootCommand(backend api.Service) *cobra.Command {
 		RunE: func(cmd *cobra.Command, args []string) error {
 			if len(args) == 0 {
 				return cmd.Help()
+			}
+			if version {
+				return versionCommand().Execute()
 			}
 			_ = cmd.Help()
 			return dockercli.StatusError{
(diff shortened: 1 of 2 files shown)


ndeloof:
> Removed the hidden flag from the version command 

:+1, I didn't noticed we kept this one hidden, this was a mistake

> Since the documentation for the v2 still constantly reffering to a --version flag on the root command

I'm fine we introduce `--version` as this as low impact but documentation should be fixed as well.
Maybe this flag should be mark hidden so we encourage use of the explicit `version` command?

Shikachuu:
> > Removed the hidden flag from the version command
> 
> :+1, I didn't noticed we kept this one hidden, this was a mistake
> 
> > Since the documentation for the v2 still constantly reffering to a --version flag on the root command
> 
> I'm fine we introduce `--version` as this as low impact but documentation should be fixed as well. Maybe this flag should be mark hidden so we encourage use of the explicit `version` command?

I think hiding this would be a nice idea, I'll add the change soon.
Besides that, everything else looks okay?

cmd/compose/compose.go
```diff
@@ -232,6 +233,9 @@ func RootCommand(backend api.Service) *cobra.Command {
 			if len(args) == 0 {
 				return cmd.Help()
 			}
+			if version == true {
```
mat007, line 236:
```suggestion
			if version {
```

mat007 (changes requested):

ndeloof (approved):

ndeloof:
All good for me
Some of your commits are not signed-off. Maybe you could also squash all of them into a single one

mat007 (approved):
Thanks!

Shikachuu:
Thanks for the reviews!
Sorry I hade to do some acrobatic manovers with rebase, but now all commits squashed to one and signed-off. :laughing:

Issue #8505, opened 2021-08-31
panic: `runtime error: index out of range [1] with length 1` in compose.buildContainerMountOptions


jamshid:
**Description**

Bringing up an existing v1 environment with v2 results in below panic. One of my services has multiple replicas and has two local anonymous(?) volumes, that seems to be the trigger.
```
  foo:
    deploy:
      replicas: 3
    volumes:
      # directory is unique per container
      - /var/foo1
      - /var/foo2
```

I would be fine with having to `docker-compose down` the v1 environment, that seems to allow `docker-compose up -d` to then work, but that loses the unnamed per-container volume on my scaled `foo` service. So I'm still unable to migrate a running v1 environment to v2.0.0-rc.2. Also, I get this error even when starting with v2 and recreating containers -- it's not specific to a pre-existing v1 environment, so I'm still unable to use v2.

PS: contact me I'm "jamshid" in the Docker Community slack if you want access to the yml and registry creds needed to run this environment. Or, if you could explain how to run a test build of compose v2 in Docker for Mac maybe I could try that before the rc3 release to shake out any remaining bugs.

**Steps to reproduce the issue:**
1. Using docker-compose v2 bring up this docker-compose.yml using `docker-compose up -d`, it should come up fine:
```
services:
  test:
    deploy:
      replicas: 3
    image: centos
    volumes:
    - type: volume
      target: /var/perm1
      volume: {}
    - type: volume
      target: /var/perm2
      volume: {}
    command: sleep 300
```
2. Now bring it up again, recreating containers. It fails with the below panic:
```
$ docker-compose up -d --force-recreate
[+] Running 0/0
 ⠋ Container test_test_1  Recreate                                                                                 0.1s
 ⠋ Container test_test_3  Recreate                                                                                 0.1s
 ⠋ Container test_test_2  Recreate                                                                                 0.1s
...
[+] Running 0/3
 ⠸ Container test_test_1  Recreate                                                                                10.4s
 ⠸ Container test_test_3  Recreate                                                                                10.4s
 ⠸ Container test_test_2  Recreate                                                                                10.4s
panic: runtime error: index out of range [1] with length 1

goroutine 86 [running]:
github.com/docker/compose-cli/pkg/compose.buildContainerMountOptions(0xc0005b403d, 0x4, 0xc0005b4030, 0x11, 0xc0000c6a00, 0x1, 0x1, 0xc0002dd080, 0xc0002dd0b0, 0xc0002dd0e0, ...)
[...]

PR #8750, opened 2021-10-05
Fix index out of range on `compose.buildContainerMountOptions`

ulyssessouza:
**What I did**
Remove buggy code, since it seems to be incorrect even when not panic-ing

**Related issue**
Resolves https://github.com/docker/compose/issues/8505


pkg/compose/create.go
```diff
@@ -762,10 +762,6 @@ func buildContainerMountOptions(p types.Project, s types.ServiceConfig, img moby
 						Target:   m.Destination,
 						ReadOnly: !m.RW,
 					}
-					// Avoid mount to be later re-defined
-					l := len(s.Volumes) - 1
-					s.Volumes[i] = s.Volumes[l]
-					s.Volumes = s.Volumes[:l]
```
mat007, line 768: This is an attempt at deleting the i element, but clearly doing this while iterating is a recipe for disaster…

However there was a definite intent there, as the comment seems to imply.
Do we really not know what it’s supposed to do?

Do we have unit tests? 😁

pkg/compose/create.go
```diff
@@ -762,10 +762,6 @@ func buildContainerMountOptions(p types.Project, s types.ServiceConfig, img moby
 						Target:   m.Destination,
 						ReadOnly: !m.RW,
 					}
-					// Avoid mount to be later re-defined
```
ndeloof, line 765: I'm not sure I remember the detail, but deletion here is expected afaict.
Maybe better create a new `Volumes` before the loop, copy those that need to be kept while iterating, then override `s.Volumes`

mat007: Yes, either that or a good old `for i := 0; i < len(s.Volumes {`, incrementing i only when not deleting.

ulyssessouza:
I actually removed that cuz in the end in the `docker inpect` that was correct even without it.
But you are right... Maybe that breaks a special case.
I'll reimplement that just refactoring the removal

Issue #8722, opened 2021-10-01
compose v2 runtime issue


egyptianbman:
I'm trying to get a container that uses the `nvidia` runtime to work but I can't seem to be able to do so, using compose v2. I can run the following without issue:
```console
➜ docker run --rm --runtime nvidia nvidia/cuda:11.0-base nvidia-smi    
Fri Oct  1 07:24:26 2021       
+-----------------------------------------------------------------------------+
| NVIDIA-SMI 460.73.01    Driver Version: 460.73.01    CUDA Version: 11.2     |
|-------------------------------+----------------------+----------------------+
| GPU  Name        Persistence-M| Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp  Perf  Pwr:Usage/Cap|         Memory-Usage | GPU-Util  Compute M. |
|                               |                      |               MIG M. |
|===============================+======================+======================|
|   0  A100-SXM4-40GB      Off  | 00000000:00:04.0 Off |                    0 |
| N/A   32C    P0    50W / 350W |      0MiB / 40536MiB |      0%      Default |
|                               |                      |             Disabled |
+-------------------------------+----------------------+----------------------+
                                                                               
+-----------------------------------------------------------------------------+
| Processes:                                                                  |
|  GPU   GI   CI        PID   Type   Process name                  GPU Memory |
|        ID   ID                                                   Usage      |
|=============================================================================|
|  No running processes found                                                 |
+-----------------------------------------------------------------------------+
```
Running the following test compose:
```yaml
version: '3.9'
                                                                                                       
services:                                                                                              
  test:
    image: nvidia/cuda:11.0-base
    command: nvidia-smi
    runtime: nvidia  
```
Doesn't yield the expected output. I changed the command to bash and ran the container and noticed that upon running a docker inspect, the output isn't what I expected:
```console
➜ docker inspect test-test-1 | jq '.[0].HostConfig.Runtime'
"runc"
```

As expected, the runtime is configured in `/etc/docker/daemon.json`:
```console
➜ cat /etc/docker/daemon.json 
{
    "runtimes": {
        "nvidia": {
            "path": "nvidia-container-runtime",
            "runtimeArgs": []
        }
    }
}
```

egyptianbman:
Updated title as requested.

Looking at the documentation again, I noticed that `runtime` is marked as legacy. I tested the provided example utilizing `capabilities`:
```yaml
services:
  test:
    image: nvidia/cuda:10.2-base
[...]

PR #8783, opened 2021-10-12
pass runtime option to containerCreate

ndeloof:
**What I did**
Pass the (missing) `runtime` option to containerCreate HostConfig

**Related issue**
close https://github.com/docker/compose/issues/8722


ndeloof:
we have a few flaky tests... I wonder we have a way to mark them as such and get auto-retry on CI

PR #8793, opened 2021-10-14
[FIX] using --no-build with build but without an image section

### LINKED ISSUE #8785: Can't use --no-build with services with a build section
When using `docker compose up --no-build` a check is done to ensure either `image` or `build` exists in a service: https://github.com/docker/compose/blob/7365917244ea791d304d5284a706aff09965636e/pkg/compose/build.go#L98 called from https://github.com/docker/compose/blob/ef786f9245eef1759c1b6da2661a2703b243fb54/pkg/compose/create.go#L66

However as the `build` section is removed in `createOptions.Apply` when `--no-build` is specified https://github.com/docker/compose/blob/7c47673d4af41d79900e6c70bc1a3f9f17bdd387/cmd/compose/create.go#L117

This means that `docker compose up --no-build` always fails even if you've ensured these images have been built.

Full repro here: https://github.com/z0w13/docker-compose-no-build-repro

```
Docker Compose version 2.0.1
```

The behaviour in 1.x was (in my opinion) correct, failing if the image hasn't been preivously built and otherwise succeeding:

```
docker-compose up --no-build
ERROR: Service 'repro' needs to be built, but --no-build was passed.
```

```
docker-compose build
  Building repro
  [+] Building 0.0s (3/3) FINISHED
   => [internal] load build definition from Dockerfile
  0.0s => => transferring dockerfile: 31B
    0.0s => [internal] load .dockerignore
  0.0s => => transferring context: 2B
  0.0s => exporting to image
  0.0s => => writing image sha256:78f8bb0557f61538b4fbb173d0ca3a8f466d77cfeee0c62a1a6f969288eeff6b
  0.0s => => naming to docker.io/library/repro_repro
docker-compose up --no-build
  Creating repro_repro_1 ...
```

debdutdeb:
Signed-off-by: Debdut Chakraborty <debdut.chakraborty@rocket.chat>

### What I did

- Add the image name to the services if options.noBuild == true

### Related issue

closes #8785 

### Picture of a cute animal

76yfg5jaj9t71

### DIFF
--- a/cmd/compose/create.go
+++ b/cmd/compose/create.go
@@ -126,7 +126,9 @@ func (opts createOptions) Apply(project *types.Project) {
 	if opts.noBuild {
 		for i, service := range project.Services {
 			service.Build = nil
-			service.Image = project.Name + "_" + service.Name
+			if service.Image != "" {
+				service.Image = project.Name + "_" + service.Name
+			}
 			project.Services[i] = service
 		}
 	}
--- a/cmd/compose/create.go
+++ b/cmd/compose/create.go
@@ -126,6 +126,7 @@ func (opts createOptions) Apply(project *types.Project) {
 	if opts.noBuild {
 		for i, service := range project.Services {
 			service.Build = nil
+			service.Image = project.Name + "_" + service.Name
 			project.Services[i] = service
 		}
 	}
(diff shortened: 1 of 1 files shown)


debdutdeb:
@thaJeztah @ndeloof hi .. not sure why the e2e test is failing. Could you please let me know how this pr is related to that specific test?

ndeloof:
we suffer some flakiness on e2e tests

glours:
I just saw @ndeloof's PR which define a function to define the default image name logic and used it everywhere needed
I think we should rely to this one and close your PR, sorry

PR #8838, opened 2021-10-27
Update README.md: typographical edit of "About update..."

youssefeldakar:
Quick typographical edit.


youssefeldakar:
> Thanks, that looks good, but you need to sign your commit, see https://github.com/docker/compose/pull/8838/checks?check_run_id=4020167232

Sorry I missed that, done.

Issue #8832, opened 2021-10-25
Compose V2 - Problems when assigning IPv6 to container with macvlan


llitz:
Docker Compose version v2.0.1 - ipv6 assignment doesn't work

docker-compose version 1.29.2, build 5becea4c - ipv6 assignment works
docker-py version: 5.0.0
CPython version: 3.7.10
OpenSSL version: OpenSSL 1.1.0l  10 Sep 2019

When starting the container with "docker compose up -d" the container doesn't have ipv6 enabled at all, checking internal sysctl shows ipv6 is disabled

The same configuration, with "docker-compose up -d", starts without any issues.
The IPv6 being assigned to the container doesn't match the fixed-cidr-v6.
My RA advertises 2 prefixes, on top of the static IP provided here `fd12:4x:e237:6905::0506`, the container is also automatically assigned the following:
`2602:4x:e237:6905:42:c0ff:fea8:506/64` - gua
`fd12:4x:e237:6905:42:c0ff:fea8:506/64` - ula
`fe80::42:c0ff:fea8:506/64` - link local

Below the relevant configs

`/etc/docker/daemon.json`
```
{
	"log-opts": {
		"max-size": "100m",
		"max-file": "3",
		"compress": "true"
	},
	"experimental": true,
	"ipv6": true,
	"ip6tables": true,
	"fixed-cidr-v6": "fd27:f005:6ffb::/48"
}
```

`docker-compose.yml` for pihole
```
version: '2.4'
services:
    pihole:
      image: pihole/pihole:latest
      container_name: pihole
      mem_limit: 512M
      restart: unless-stopped
      environment:
        - TZ=America/Denver
      volumes:
        - /usr/share/zoneinfo/MST7MDT:/etc/localtime:ro
        - ./data/pihole/etc-pihole/:/etc/pihole/:z
        - ./data/pihole/etc-dnsmasq.d:/etc/dnsmasq.d/:z
      networks:
        priv_net:
          ipv4_address: 192.168.5.6
          ipv6_address: fd12:41:e237:6905::0506

networks:
  priv_net:
    enable_ipv6: true
    driver: macvlan
    driver_opts:
      parent: ens19
    ipam:
      config:
        - subnet: 192.168.5.0/24
          gateway: 192.168.5.1
          ip_range: 192.168.5.6/32
        - subnet: fd12:41:e237:6905::/64
          gateway: fd12:41:e237:6905::11
```

PR #8851, opened 2021-10-28
add support for EnableIPv6

ndeloof:
**What I did**
fix support for network IPv6

**Related issue**
close https://github.com/docker/compose/issues/8832


solidgoldbomb:
I've been using an identical change in my local environment for more than a week and can confirm that this works for me.

mat007 (approved):
Thanks!

Issue #8746, opened 2021-10-04
passing existing environment variable with docker compose exec -e SOMEVAR does not work anymore


thunder7553:
**Description**

Environment variables are not passed from the shell to the spawned process

**Steps to reproduce the issue:**
1. export SOMEVAR=1
2. docker compose exec -e SOMEVAR container sh 
2. printenv

**Describe the results you received:**

SOMEVAR is not in the environment

**Describe the results you expected:**
SOMEVAR=1 is listed in the environment. 

**Additional information you deem important (e.g. issue happens only occasionally):**

* Passing the variable value with `docker compose exec -e SOMEVAR=$SOMEVAR container sh` does work
* Passing the variable name without value worked in previous docker compose (and docker-compose) versions.

**Output of `docker compose version`:**

```
Docker Compose version v2.0.0
```

**Output of `docker info`:**

```
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Build with BuildKit (Docker Inc., v0.6.3)
  compose: Docker Compose (Docker Inc., v2.0.0)
  scan: Docker Scan (Docker Inc., v0.8.0)

Server:
 Containers: 95
  Running: 22
  Paused: 0
  Stopped: 73
 Images: 500
 Server Version: 20.10.8
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 1
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runtime.v1.linux runc io.containerd.runc.v2
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: e25210fe30a0a703442421b0f60afac609f950a3
 runc version: v1.0.1-0-g4144b63
 init version: de40ad0
 Security Options:
  seccomp
   Profile: default
 Kernel Version: 5.4.72-microsoft-standard-WSL2
 Operating System: Docker Desktop
 OSType: linux
 Architecture: x86_64
 CPUs: 6
 Total Memory: 11.7GiB
 Name: docker-desktop
 ID: ZP5T:J5XU:7BYY:3LXW:PNQA:KKM6:DEP3:RUOP:55XY:VETW:VEQQ:JHCT
[...]

PR #8857, opened 2021-10-31
Support inherited environment vars on exec

ulyssessouza:
**What I did**
Uses project's environment variables.

**Related issue**
Resolves https://github.com/docker/compose/issues/8852
Resolves https://github.com/docker/compose/issues/8746

### DIFF
--- a/cmd/compose/exec.go
+++ b/cmd/compose/exec.go
@@ -21,11 +21,12 @@ import (
 	"fmt"
 	"os"
 
+	"github.com/compose-spec/compose-go/types"
 	"github.com/containerd/console"
 	"github.com/docker/cli/cli"
-	"github.com/spf13/cobra"
-
 	"github.com/docker/compose/v2/pkg/api"
+	"github.com/docker/compose/v2/pkg/compose"
+	"github.com/spf13/cobra"
 )
 
 type execOpts struct {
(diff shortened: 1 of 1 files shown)


pkg/compose/exec.go
```diff
@@ -38,7 +39,7 @@ func (s *composeService) Exec(ctx context.Context, project string, opts api.RunO
 
 	exec, err := s.apiClient.ContainerExecCreate(ctx, container.ID, moby.ExecConfig{
 		Cmd:        opts.Command,
-		Env:        opts.Environment,
+		Env:        ToMobyEnv(types.NewMappingWithEquals(opts.Environment).Resolve(os.LookupEnv)),
```
ndeloof, line 42: Resolve should use project.Environment, not os.LookupEnv, so this also can resolve variables set by .env, and let us (in the future) get more control on the way environment is set

ndeloof (changes requested):
LGTM but os.env vs project.env

pkg/compose/exec.go
ulyssessouza: Hmmm... I think we need to change the interface for that. To construct the whole project we need to do that in the CLI side. So before arriving here.

cmd/compose/exec.go
```diff
@@ -77,7 +78,7 @@ func execCommand(p *projectOptions, backend api.Service) *cobra.Command {
 }
 
 func runExec(ctx context.Context, backend api.Service, opts execOpts) error {
-	project, err := opts.toProjectName()
+	project, err := opts.toProject([]string{opts.service}, composegocli.WithResolvedPaths(true))
```
mix3, line 81: #8658 の修正とコンフリクトするように見えます
Project を渡す interface の修正に同意しますが、compose ファイルなしに Project を生成できる修正も同時に必要かもしれません
また、その修正は compose-spec/compose-go にも及ぶかもしれません

(DeepL) Looks like it conflicts with fix #8658
I agree with the fix for the interface that passes the Project, but there may also need to be a fix to generate the Project without the compose file.
Also, the fix may extend to compose-spec/compose-go

ndeloof (changes requested):
that would be sad one can't `compose exec` without the original compose file (this is one of the major compose V2 improvements imho)
can't you pass environment to be applied to `Exec` by `api.RunOptions#Environment ` ?

ulyssessouza:
@ndeloof Wow! I totally bypassed the fact that exec doesn't need the file anymore (V1 state of mind). Sorry!
Yep! I'll change that to get it from the options.
But actually looking at the code I just saw a a call to `toProjectName` [here](https://github.com/docker/compose/blob/v2/cmd/compose/exec.go#L80) and from what I have in mind, that will read the compose file at some point.

ndeloof (approved):

PR #8873, opened 2021-11-03
Update golang to 1.17

rumpl:
**What I did**

* Updated golang to 1.17
* Ran `make go-mod-tidy`

image


mat007 (approved):
Sounds good!

ndeloof (approved):
weird this introduces so much changes in the go.mod file, bu .. that's go mod after all :trollface:

Issue #8884, opened 2021-11-05
Attaching to a new container, docker compose up misses the initial logs


sdt:
**Description**

With docker-compose up, attaching to a new container misses the initial log lines for that container.

While testing #8859, I noticed that when a container was stopped, removed, then recreated, the first few log lines were missing.

**Steps to reproduce the issue:**

```yaml
version: "3.8"
services:
  aaa:
    image: ubuntu:focal
    init: true
    command: bash -c 'echo Start aaa; while echo aaa; do sleep 5; done | cat -n'
  bbb:
    image: ubuntu:focal
    init: true
    command: bash -c 'echo Start bbb; while echo bbb; do sleep 5; done | cat -n'
```

1. Start the stack with the above compose file and`docker-compose up`
2. In another terminal, stop and remove one of the containers with `docker-compose rm -fs bbb`
3. Start a new container with `docker-compose up -d bbb` - the first few lines of the logs will not be shown

Steps 2 and 3 can be also replaced by `docker-compose up -d --scale bbb=1`

**Describe the results you received:**

The logs from `docker-compose up` will look something like this:

```
Attaching to test-aaa-1, test-bbb-1
test-bbb-1  | Start bbb
test-bbb-1  |      1    bbb
test-aaa-1  | Start aaa
test-aaa-1  |      1    aaa
test-bbb-1  |      2    bbb
test-aaa-1  |      2    aaa
test-bbb-1  |      3    bbb
test-aaa-1  |      3    aaa
test-bbb-1 exited with code 143
test-aaa-1  |      4    aaa
test-bbb-1  |      2    bbb
test-aaa-1  |      5    aaa
test-bbb-1  |      3    bbb
```

**Describe the results you expected:**

The logs should contain a second instance of "Start bbb" and "1 bbb".

Note that `docker-compose logs -f` shows the full logs with no missing lines.

```
Attaching to test-aaa-1, test-bbb-1
test-aaa-1  | Start aaa
test-aaa-1  |      1    aaa
test-bbb-1  | Start bbb
test-bbb-1  |      1    bbb
test-aaa-1  |      2    aaa
test-bbb-1  |      2    bbb
test-aaa-1  |      3    aaa
test-bbb-1  |      3    bbb
test-bbb-1 exited with code 143
test-bbb-1  | Start bbb       <--- missing
test-bbb-1  |      1    bbb   <--- missing
test-aaa-1  |      4    aaa
test-bbb-1  |      2    bbb
[...]

PR #8885, opened 2021-11-05
Fix missing logs when attaching new container

sdt:
**What I did**

When attaching to a container that has just been started, there may have already been some logs emitted.

Setting ContainerAttachOptions.Logs to true in getContainerStreams gives us the log lines that have already been emitted.

The docs for ContainerAttach suggest this is the right approach:

https://docs.docker.com/engine/api/v1.41/#operation/ContainerAttach

> Replay previous logs from the container.
>
> This is useful for attaching to a container that has started and you want to output everything since the container started.
>
> If stream is also enabled, once all the previous output has been returned, it will seamlessly transition into streaming current output.

I don't have any tests for this I'm afraid. Happy to create some, but I'd need a little guidance on the best way to attack it.

**Related issue**

Fixes: #8884 

image

Signed-off-by: Stephen Thirlwall <sdt@dr.com>

### DIFF
--- a/pkg/compose/attach.go
+++ b/pkg/compose/attach.go
@@ -142,7 +142,7 @@ func (s *composeService) getContainerStreams(ctx context.Context, container stri
 		Stdin:  true,
 		Stdout: true,
 		Stderr: true,
-		Logs:   false,
+		Logs:   true,
 	})
 	if err == nil {
 		stdout = ContainerStdout{HijackedResponse: cnx}


mat007 (approved):
I think that makes sense. I’m not sure how to write a reliable test either…

ndeloof:
I'm not sure why we miss the initial log, but this fix is wrong: doing so, we would capture logs from the creation of the container, and in many case (restart, compose logs on a running app, ...) this isn't expected.

we attach to container _before_ it get started, so there's no reason we miss some logs - `docker run` does the same, and doesn't have this issue. So the root cause for this is harder to tackle

ndeloof (changes requested):
unfortunately this would break other usages, and should not be required as we attach _before_  container is started. Need further investigations

sdt:
> I'm not sure why we miss the initial log, but this fix is wrong: doing so, we would capture logs from the creation of the container, and in many case (restart, compose logs on a running app, ...) this isn't expected.

You're right. I modified the compose file in #8884 to also print the pid. Running `docker-compose restart bbb` causes the duplicate logs case.

```
test-bbb-1  | Start bbb
test-bbb-1  |      1    bbb 8
test-bbb-1  |      2    bbb 8
test-bbb-1 exited with code 143  <-- docker-compose restart bbb
test-bbb-1  | Start bbb          <-- old logs repeated
test-bbb-1  |      1    bbb 8
test-bbb-1  |      2    bbb 8
test-bbb-1  | Start bbb          <-- new logs start here
test-bbb-1  |      1    bbb 7
test-bbb-1  |      2    bbb 7
test-bbb-1  |      3    bbb 7
test-bbb-1  |      4    bbb 7
```

As for the other case, `docker-compose logs -f` on a running app already shows similar behaviour, independent of this change. I'd noticed this earlier but wasn't sure if it was intentional. There's prior art for both cases. `dmesg -w` shows the full set of logs before watching for new ones. `tail -f` only shows the last 10 previous logs. For this case there's no duplicate log messages.

sdt:
> we attach to container before it get started, so there's no reason we miss some logs

The missing logs case comes from the scenario in #8884. `docker-compose up` is already running on the stack that it started. In another terminal we trigger a new container to be created. The original `docker-compose up` process receives a container started event, and ends up here, calling its `onStart` callback:

https://github.com/docker/compose/blob/125752c1272132737d76aac69bad5d430179d4e3/pkg/compose/start.go#L148-L154

Which in this case calls `s.attachContainer`

https://github.com/docker/compose/blob/125752c1272132737d76aac69bad5d430179d4e3/pkg/compose/start.go#L50-L52

This is the case where docker-compose hasn't already attached to a container. Instead the container has created by some other process, and it already running when we attach to it.

It looks to me that `ContainerAttachOptions.Logs` should vary depending on whether we're starting a container that we created ourselves and are already attached to, or if we're reacting to an external container start event. I'll have a play around with that and see what falls out.

sdt:
I've had a play around with this. I thought it might be enough to have a `wantOldLogs` parameter to `assignContainers`, and to hardcode that depending on the call site. In general that would be false, but in the `onStart` callback that would be true.

That turns out not to be enough. The `onStart` callback is triggered for externally created containers, where we want the old logs, but also for existing containers being restarted, where we don't.

The key predicate here seems to be "have we seen this container ID before?" For the restart case, the answer is yes, and for the new container case, no.

The `watched` map almost answers this for us, but not quite.

https://github.com/docker/compose/blob/125752c1272132737d76aac69bad5d430179d4e3/pkg/compose/start.go#L88-L91

It serves a double-duty. One: the integer values are used in the die event to determine if the container will be automatically restarted.

https://github.com/docker/compose/blob/125752c1272132737d76aac69bad5d430179d4e3/pkg/compose/start.go#L115-L118

Two: the presence of running containers is used to detect whether it's time to exit.

https://github.com/docker/compose/blob/125752c1272132737d76aac69bad5d430179d4e3/pkg/compose/start.go#L133-L136

With this logic, we can't answer the seen-this-container-before predicate because entries get deleted from `watched` during the die handler. When the container restarts and we get a start event, it has the same container id, but has already been removed from `watched` and so cannot be distinguished from a completely new container.

What might work is to store two values in that map: the start/restart count, and a running flag. The counter would work the same, and the running flags would be used to determine when to stop. In the die handler, rather than delete the entry, just clear the running flag. If all the running flags are false, it's time to stop.

In the start event handler, if there is already an entry for this container id, we know this is a container we've seen before, and we don't want old logs. Otherwise this is a completely new container, and we do want the old logs.

The destroy event can be used to remove entries from `watched` - at that point the container will not be coming back.

sdt:
As noted [here](https://github.com/docker/compose/issues/8884#issuecomment-962680584), the same missing logs happen with just `docker-compose restart $service`, so I think the "seen this container before" idea isn't right either.

sdt:
I'm going to close this for now. It's looking like #8897 may provide a better solution for this, or at least get us closer.

I'll re-evaluate things once that's sorted out.

Issue #8896, opened 2021-11-08
Compose should not exit when container is scheduled for restart 


ndeloof:
Compose should know container will restart, and wait for it.

```yaml
services:
  test:
    image: ubuntu
    command: bash -c 'for ((i=1; i <= 10; i++)); do echo $$i; done'
    restart: always
```

## expected behavior:
`docker compose up` to show 1..10 in a loop

## observed behavior:
`docker compose up` only show 1..10 once.

PR #8898, opened 2021-11-08
better detect container will restart

ndeloof:
**What I did**
detect container will restart based on inspected restart policy and number of previous restarts

**Related issue**
close https://github.com/docker/compose/issues/8896

image


ulyssessouza (approved):
LGTM

PR #8903, opened 2021-11-09
run build test with Both buildkit and classic builder

ndeloof:
**What I did**
Run build tests twice, once with builder classic and then with buildkit enabled

### DIFF
--- a/pkg/e2e/compose_build_test.go
+++ b/pkg/e2e/compose_build_test.go
@@ -18,6 +18,7 @@ package e2e
 
 import (
 	"net/http"
+	"os"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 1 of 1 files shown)


pkg/e2e/compose_build_test.go
```diff
@@ -26,7 +27,18 @@ import (
 	"gotest.tools/v3/icmd"
 )
 
-func TestLocalComposeBuild(t *testing.T) {
+func TestLocalComposeBuild_Classic(t *testing.T) {
+	os.Setenv("DOCKER_BUILDKIT", "0")    //nolint:errcheck
```
mat007, line 31: nit: here we could either do
```suggestion
	_ = os.Setenv("DOCKER_BUILDKIT", "0")
```
or check the error?

mat007 (commented):
Test is failing, it looks like a real error?
(@ulyssessouza ?)

pkg/e2e/compose_build_test.go
ndeloof: sorry, was not intended to be merged, just wanted to do a quick and dirty try to get both builder classic and buildkit CI run

mat007:
@ndeloof should we close this PR?

Issue #8880, opened 2021-11-04
CTRL-C does not end compose logs -f


logopk:
CTRL-C does not end compose logs -f

This started with compose 2.1.0

**Steps to reproduce the issue:**
1. docker-compose logs -f containername
2. CTRL-C 

whereas docker logs -f containername works as expected

**Describe the results you received:**
Command can only be killed

**Describe the results you expected:**
Return to shell

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `docker compose version`:**

```
Docker Compose version 2.1.0
```

**Output of `docker info`:**

```
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Build with BuildKit (Docker Inc., v0.6.0)
  compose: Docker Compose (Docker Inc., 2.1.0)

Server:
 Containers: 5
  Running: 5
  Paused: 0
  Stopped: 0
 Images: 18
 Server Version: 20.10.10
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 io.containerd.runtime.v1.linux runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 5b46e404f6b9f661a205e28d59c982d3634148f8
 runc version: v1.0.2-0-g52b36a2
 init version: de40ad0
 Security Options:
  apparmor
  seccomp
   Profile: default
  cgroupns
 Kernel Version: 5.10.0-9-amd64
 Operating System: Debian GNU/Linux 11 (bullseye)
 OSType: linux
 Architecture: x86_64
 CPUs: 2
 Total Memory: 1.936GiB
 Name: debian
 ID: FJ7Z:235F:BJJA:OYFZ:6SRB:6XWR:U3JO:SCBH:EQ4G:G55K:2KQL:5JK2
 Docker Root Dir: /var/lib/docker
 Debug Mode: false
 Registry: https://index.docker.io/v1/
 Labels:
 Experimental: true
 Insecure Registries:
  127.0.0.0/8
 Live Restore Enabled: false
 Default Address Pools:
[...]

PR #8926, opened 2021-11-15
interrupt printer when `compose log` is cancelled

ndeloof:
**What I did**
within `compose logs` stop the log printer when context is cancelled 
`compose up` is a distinct case as the interruption signal is actually used to stop the stack, and log printer waits for container to exit
(which makes me feel using `context` was not the best option to manage stop signal, but this is what we have)

**Related issue**
close https://github.com/docker/compose/issues/8880


ulyssessouza (approved):
LGTM

Issue #8894, opened 2021-11-06
Build from URL git repository no longer working in v2.1.0


ToshY:
**Description**

I just updated from `v2.0.1` to `v2.1.0` and tried to up my containers but build from URL for a git repo no longer works.

**Steps to reproduce the issue:**
```yaml
redis-overcommit:
    build: https://github.com/bkuhl/redis-overcommit-on-host.git
    container_name: redis-overcommit
    restart: 'no'
    privileged: true
    volumes:
        - /proc/sys/vm:/mnt/vm
```

```
docker compose up -d --build
```

FYI specifying the Dockerfile directly still works, `https://raw.githubusercontent.com/bkuhl/redis-overcommit-on-host/master/Dockerfile`.

**Describe the results you received:**
```
1 error occurred:
        * unable to prepare context: unable to evaluate symlinks in Dockerfile path: lstat /tmp/docker-build-git733896971/https:: no such file or directory
```

**Describe the results you expected:**

No error / container being up.

**Additional information you deem important:**

I'm not really comfortable in trying to debug this, but after looking a bit at the release notes, maybe it has something to do with [this change](https://github.com/docker/compose/blame/98b3353cbcc2cf01bbd5a4ca13928ccf6dc9a283/cmd/compose/up.go#L101)? I'm definitely not sure, but I thought maybe it helps.

**Output of `docker compose version`:**

```
Docker Compose version v2.1.0
```

ndeloof:
Could you try running with buildkit enabled?
`DOCKER_BUILDKIT=1 docker compose build`

ndeloof:
could you please also confirm a plain `docker build https://github.com/bkuhl/redis-overcommit-on-host.git` works as expected?

ToshY:
@ndeloof Excuse me for the delay.

For the `DOCKER_BUILDKIT=1 docker compose build`, it doesn't show anything related to the `redis-overcommit` container

```
[+] Building 0.3s (22/22) FINISHED
 => [xxx_mysql internal] load build definition from Dockerfile                                                                                                                                          0.0s
 => => transferring dockerfile: 84B                                                                                                                                                                     0.0s
 => [xxx_apache internal] load build definition from Dockerfile                                                                                                                                         0.0s
 => => transferring dockerfile: 722B                                                                                                                                                                    0.0s
 => [xxx_elasticsearch internal] load build definition from Dockerfile                                                                                                                                  0.0s
[...]

PR #8933, opened 2021-11-17
restore `compose build` to support context set as a git URL

ndeloof:
**What I did**
restore `compose build` to support context set as a git URL

**Related issue**
close https://github.com/docker/compose/issues/8894


glours (approved):
LGTM

Issue #8833, opened 2021-10-26
Compose breaks newlines of terminal when invoked without TTY attached


christianlupus:
**Description**

There is a problem on my machine with the newlines in the output of containers that are not attached to a TTY. The newline should (under linux) start the next chars on the next line at position 1. Instead, I get a _staircase of text lines_:
![grafik](https://user-images.githubusercontent.com/8202922/138915699-6f8a2f08-6f6a-4349-9513-51941a9fc986.png)

**Steps to reproduce the issue:**
To have a basic setup, I need 3 files. I created a [gist](https://gist.github.com/christianlupus/3f93f2bf6727ff1381c7f7a8600c9404) for simpler access. Make sure that `run.sh` is executable (not sure when using gist).

Now, you can try to run this in 3 ways (with a TTY attached):
- `docker run --rm -t docker-test_test` where the part `docker-test` must match the composer project name (folder name)
- `docker compose run --rm test`
- `docker-compose run --rm test`

Each one gives the same result:
```
$ docker run --rm docker-test_test # Or the others
First line
Second line
```

When you alter the commands to not attach a TTY (remove `-t` resp add `-T` for compose), only the `docker run` variant will output two lines. The other two variants output these type of stairs.

**Describe the results you received:**
I get a staircase output like
```
$ docker-compose run --rm -T test 
First line
          Second line
```

**Describe the results you expected:**
I expect the output of a container to be independent if the input is a TTY or not.

**Additional information you deem important (e.g. issue happens only occasionally):**

I tried piping the output through `hexdump -C`. There, two things were showing up:

1. The output of `docker run` is not equal with and without the `-t` flag. With the `-t` flag the newlines are printed as `\r\n` while the non-TTY version printed only `\n`.
2. When piping the output of the composed variants, the complete output is garbled:
```
$ docker-compose run --rm -T test | hexdump -C
00000000  46 69 72 73 74 20 6c 69  6e 65 0a 53 65 63 6f 6e  |First line.Secon|
                                                                              00000010  64 20 6c 69 6e 65 0a                              |d line.|
```
I suspect the problem is some post processing  or setting in the terminal I am using.

**Output of `docker compose version`:**

```
Docker Compose version 2.0.1
```

**Output of `docker info`:**

```
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Build with BuildKit (Docker Inc., v0.6.1-docker)
  compose: Docker Compose (Docker Inc., 2.0.1)

Server:
[...]

PR #8938, opened 2021-11-18
don't SetRawTerminal when run is ran with -T

ndeloof:
**What I did**
RawTerminal must not be set when container is ran explicitly with -T

**Related issue**
closes https://github.com/docker/compose/issues/8833


ulyssessouza (approved):
LGTM

Issue #8850, opened 2021-10-28
Compose V2 down --volumes does not remove named volumes created outside but referenced in docker-compose.yml


danw-mpl:
**Description**

Identical to https://github.com/docker/compose-cli/issues/1798 - looks like this bug is back or there's a difference because I'm creating the named volume outside of Compose.  Does not happen with Compose V1.

**Steps to reproduce the issue:**

**docker-compose.yml**
```yaml
services:
  mysql:
    ...
    volumes:
      - mysql_data:/var/lib/mysql

volumes:
  mysql_data:
    name: mysql_data
```

**Our 'data load' script**
```
docker container create --name temp_container -v mysql_data:/root hello-world
docker cp SRC DST temp_container:/root/
docker rm temp_container
```

```
docker-compose up
docker-compose down --volumes
```

**Describe the results you received:**
The volume was not removed.
```
$ docker-compose down --volumes
[+] Running 6/6
 ⠿ Container mysql            Removed
```

**Describe the results you expected:**
The volume should be removed.

**Additional information you deem important (e.g. issue happens only occasionally):**
Disabling V2 resolves the issue.

**Output of `docker compose version`:**

```
Docker Compose version v2.0.0
```

**Output of `docker info`:**

```
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  app: Docker App (Docker Inc., v0.9.1-beta3)
  buildx: Build with BuildKit (Docker Inc., v0.6.3)
  compose: Docker Compose (Docker Inc., v2.0.0)
  scan: Docker Scan (Docker Inc., v0.8.0)

Server:
 Containers: 6
  Running: 0
  Paused: 0
  Stopped: 6
 Images: 277
 Server Version: 20.10.8
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 1
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 io.containerd.runtime.v1.linux runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: e25210fe30a0a703442421b0f60afac609f950a3
 runc version: v1.0.1-0-g4144b63
 init version: de40ad0
[...]

PR #8941, opened 2021-11-19
detect volume we didn't created and ask user to explicitely mark them as external

ndeloof:
**What I did**
On `compose up` we search for existing volume by name, as for legacy and usability reasons this one is determined by project name. Doing so we risk to use an existing volume that was not created by Docker Compose or was created by another compose project.

This PR adds a check that volume doesn't match expected project, and tells user to make this explicitly `external: true` if this is really what he expects.

**Related issue**
see https://github.com/docker/compose/issues/8850

### DIFF
--- a/pkg/compose/create.go
+++ b/pkg/compose/create.go
@@ -210,7 +210,7 @@ func (s *composeService) ensureProjectVolumes(ctx context.Context, project *type
 		volume.Labels = volume.Labels.Add(api.VolumeLabel, k)
 		volume.Labels = volume.Labels.Add(api.ProjectLabel, project.Name)
 		volume.Labels = volume.Labels.Add(api.VersionLabel, api.ComposeVersion)
-		err := s.ensureVolume(ctx, volume)
+		err := s.ensureVolume(ctx, volume, project.Name)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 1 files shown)


pkg/compose/create.go
```diff
@@ -1086,27 +1086,42 @@ func (s *composeService) removeNetwork(ctx context.Context, networkID string, ne
...
+	}
+
+	// Volume exists with name, but let's double check this is the expected one
+	// (better safe than sorry when it comes to user's data)
+	p, ok := inspected.Labels[api.ProjectLabel]
+	if !ok {
+		return fmt.Errorf("volume %q alrady exists but was not created by Docker Compose. Use `external: true` to use an existing volume", volume.Name)
+	}
+	if p != project {
+		return fmt.Errorf("volume %q alrady exists but was not created for project %q. Use `external: true` to use an existing volume", volume.Name, p)
```
glours, line 1106:
```suggestion
		return fmt.Errorf("volume %q already exists but was not created for project %q. Use `external: true` to use an existing volume", volume.Name, p)
```

pkg/compose/create.go
```diff
@@ -1086,27 +1086,42 @@ func (s *composeService) removeNetwork(ctx context.Context, networkID string, ne
...
-		w.Event(progress.CreatedEvent(eventName))
+		err := s.createVolume(ctx, volume)
+		return err
+	}
+
+	// Volume exists with name, but let's double check this is the expected one
+	// (better safe than sorry when it comes to user's data)
+	p, ok := inspected.Labels[api.ProjectLabel]
+	if !ok {
+		return fmt.Errorf("volume %q alrady exists but was not created by Docker Compose. Use `external: true` to use an existing volume", volume.Name)
```
glours, line 1103:
```suggestion
		return fmt.Errorf("volume %q already exists but was not created by Docker Compose. Use `external: true` to use an existing volume", volume.Name)
```

glours (approved):
Only 2 typo to fix before merging

glours:
Ho  I didn't see that the CI was red 😬

ndeloof:
yes, CI suffers for flakiness, I use to re-run mostly on all pull-request to get a successful build
(I wonder we could enable auto-retry on tests :P )

glours:
Or we can spend some time to figure why the tests are flaky 😛

MillerCM:
A request: Since the users of docker are anywhere and everywhere on the gender spectrum, please make an effort to be inclusive with your phrasing. The phrase "if this is really what he expects" is unnecessarily gendering the user as male. It's more accurate to represent your entire user base by repeating "the user" or using a neutral pronoun (historically, "they").

Thank you for your work!

Issue #8914, opened 2021-11-10
Regression: `docker compose config` does not output portable config anymore


jpic:
Previously, `docker-compose config` would output a portable (reusable) config, without names for volumes and networks that are global to docker.

Now, it enforces outputting local volume and network names at the top level networks and volumes hashes of the yaml output:

```
networks:
  default:
    name: ansible5rt8w0cmbuild_default
volumes:
  syncthing:
    name: ansible5rt8w0cmbuild_syncthing
```

Which breaks portability of the configuration generated from `docker-compose config` from a project to another anymore:

- default should not even be there
- volumes that are local to services should probably stay there at the service level

Is there any way to generate a portable config with the new docker compose?

Thank you

jpic:
Worth mentioning another regression, getting `networks: { default: null }` in services as such:

```
services:
  some_service:
    networks:
      default: null
```

docker compose up won't complain about it, but the containers won't see each others.

jpic:
Maybe related, one of the first things I fixed, was that instead of:

```
    volumes:
      - ./syncthing:/var/syncthing
      - ./data/media:/app/media
      - ./spooler:/spooler
      - ./log:/app/log
```

I'm now getting:

```
    - target: /app/log
      type: volume
      volume: {}
    - target: /app/media
      type: volume
      volume: {}
    - target: /tmp
      type: volume
      volume: {}
    - source: syncthing
      target: /var/syncthing
      type: volume
      volume: {}
```

Works but very verbose... I thought it was a feature but now I'm not sure ...

ndeloof:
`docker-compose config` usage has never been clearly defined in compose V1.
For Compose V2, `config` command has been designed as a diagnostic solution to dump the actual, fully resolved model to be applied. This is why all "shot syntax" notation get extended, default values set, and implicit names resolved.

compose model portability is actually managed by the Compose Specification being adopted by various tools. What's your use case for this?

jpic:
> compose model portability is actually managed by the Compose Specification being adopted by various tools. What's your use case for this?

My use case is building a single compose file from several files, inspect the target host and eventually hack it a bit and patch the file, and then deploy that. I can show you if you're curious (it's all open source and documented), but I don't want to distract this issue from what matters which I think is the following:

[...]

PR #8943, opened 2021-11-19
don't normalize compose model in compatibility mode

ndeloof:
**What I did**
use WithNormalization (https://github.com/compose-spec/compose-go/pull/204) to keep the compose model as close as possible from the original yaml file when running `compose config` command. This is enabled by default in backward compatible mode (`compose --compatibility`) or can be triggered by `--no-normalize`. 
This doesn't make the `config` command strictly the same as the legacy V1 one due to the specific logic of compose-go to load individual yaml files then merge the go structs together, but get closer

**Related issue**
close https://github.com/docker/compose/issues/8914

### DIFF
--- a/cmd/compose/convert.go
+++ b/cmd/compose/convert.go
@@ -44,6 +44,7 @@ type convertOptions struct {
 	quiet               bool
 	resolveImageDigests bool
 	noInterpolate       bool
+	noNormalize         bool
 	services            bool
 	volumes             bool
 	profiles            bool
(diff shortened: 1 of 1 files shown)


ndeloof:
@jpic would you have an opportunity to give this a try and confirm this fixed your concerns with `config` command in compose v2?

glours (approved):
LGTM

mat007 (approved):
Makes sense!

Issue #8942, opened 2021-11-19
-–quiet-pull


DavidPerezIngeniero:
`docker-compose up` admits the option `-–quiet-pull`, whereas `docker-compose run` doesn't and it can also pull images.

Could this option also be supported?

PR #8947, opened 2021-11-21
introduce run —quiet-pull to align with up

ndeloof:
**What I did**
Introduce `compose run —quiet-pull` to align with same flag on `up` command

**Related issue**
close https://github.com/docker/compose/issues/8942


mat007 (approved):
Makes sense!

mat007:
Fixed the linting issue and force pushed, merging.

PR #8953, opened 2021-11-22
Add multiargs build e2e tests

ulyssessouza:


pkg/compose/build_classic.go
```diff
@@ -243,7 +243,7 @@ func imageBuildOptions(options buildx.Options) dockertypes.ImageBuildOptions {
 func toMapStringStringPtr(source map[string]string) map[string]*string {
 	dest := make(map[string]*string)
 	for k, v := range source {
-		v := v
+		// v := v
```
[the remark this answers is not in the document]
mat007: Without that line, what it does is assign the pointer to the last value in `source` to all values in `dest`.
Given that the test has two args with the same value, it explains why it doesn’t fail.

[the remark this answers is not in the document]
mat007: The space for `v` is allocated once, and then each iteration of the loop overwrites `v` with the new value.
But it’s always the same pointer value, as it points to the same memory location.
Therefore every time we assign the pointer, it’s actually the same pointer. And after the loop is finished, the variable pointed to by that pointer holds the value of the last element in the range.

pkg/e2e/compose_build_test.go
```diff
@@ -66,6 +66,21 @@ func TestLocalComposeBuild(t *testing.T) {
 		res.Assert(t, icmd.Expected{Out: `"FOO": "BAR"`})
 	})
 
+	t.Run("build with multiple build-args ", func(t *testing.T) {
+		for i := 0; i < 5; i++ {
```
mat007, line 70: I don’t think we need a loop here, it’s a deterministic bug.

pkg/e2e/compose_build_test.go
```diff
@@ -66,6 +66,21 @@ func TestLocalComposeBuild(t *testing.T) {
 		res.Assert(t, icmd.Expected{Out: `"FOO": "BAR"`})
 	})
 
+	t.Run("build with multiple build-args ", func(t *testing.T) {
```
mat007, line 69:
```suggestion
	t.Run("build with multiple build-args", func(t *testing.T) {
```

pkg/e2e/fixtures/build-test/multi-args/Dockerfile
```diff
@@ -0,0 +1,5 @@
+ARG IMAGE=666
+ARG TAG=666
+
+FROM ${IMAGE}:${TAG}
```
mat007, line 4: We need a way to input two different values for the args, and afterwards verify that they’re still two different values.

pkg/e2e/compose_build_test.go
ulyssessouza: That was just a desperate attempt to reproduce it 🥲

pkg/e2e/compose_build_test.go
```diff
@@ -66,6 +66,21 @@ func TestLocalComposeBuild(t *testing.T) {
...
+		for i := 0; i < 5; i++ {
+			// ensure local test run does not reuse previously build image
+			c.RunDockerOrExitError("rmi", "-f", "multiargs")
```
mat007, line 72: Your test works, it’s just that here it’s
```suggestion
		c.RunDockerOrExitError("rmi", "-f", "multi-args_multiargs")
```

mat007:
I updated the PR.

mat007 (approved):

Issue #5288, opened 2017-10-19
Feature Request: list all local images with docker-compose, even without containers


slavik57:
Currently `docker-compose images` lists only images with created containers.
```
> docker-compose images --help
List images used by the created containers.
Usage: images [options] [SERVICE...]

Options:
-q     Only display IDs
```

I need a way to list all images on local machine relevant to `docker-compose` file even if they don't have containers.

I found a workaround with these steps:
```
> docker-compose create
> docker-compose images
> docker-compose rm
```

I feel there should be a way of doing it with one command. Adding a flag to `images` would be nice, something like `-a` meaning `list all images`

Some info:
```
> docker version
Client:
 Version:      17.05.0-ce
 API version:  1.29
 Go version:   go1.7.5
 Git commit:   89658be
 Built:        Thu May  4 22:06:25 2017
 OS/Arch:      linux/amd64

Server:
 Version:      17.05.0-ce
 API version:  1.29 (minimum version 1.12)
 Go version:   go1.7.5
 Git commit:   89658be
 Built:        Thu May  4 22:06:25 2017
 OS/Arch:      linux/amd64
 Experimental: false
```

```
> docker-compose version
docker-compose version 1.15.0, build e12f3b9
docker-py version: 2.5.1
CPython version: 2.7.5
OpenSSL version: OpenSSL 1.0.1e-fips 11 Feb 2013
```

florianeichin:
If I'm getting the source code right, `docker-compose images `is not returning images but containers. Is that as planned or is it more kind of a bug?

shin-:
@florianeichin That's a bit of a wild assumption. Have you tried running the command?

florianeichin:
yes I did, but you getting the containers and the corresponding images, that I think is also true if I'm looking into the source code, where you get the information out of the "container.py". So therefore it's not simply possible to get the images, that are not associated to a container, no?

cosmopetrich:
I think what's being requested is a way to determine which image compose would use for a particular service without needing to run that service first.

This could be useful in CI where compose is used to provide the arguments to `docker-compose build` and there's a need to operate on a particular image even if its container hasn't been brought up.

```sh
# This will return no output
docker-compose build
docker-compose images

# This will return a list of containers and images
docker-compose build
docker-compose up -d
docker-compose images
```

[...]

PR #8955, opened 2021-11-22
introduce docker compose config --images

ndeloof:
**What I did**
To align with the few existing `config -xxx` pseudo-commands used to inspect the compose model, I suggest we introduce `compose config --images` to list service images. This is a distinct command vs `compose images` which list the _live_ images from running containers.

_Note:_ we might want to make the default image ID computed from project+service name something canonical by having this declared on compose-go as ServiceConfig#GetImageName()

**Related issue**
close https://github.com/docker/compose/issues/5288

### DIFF
--- a/cmd/compose/convert.go
+++ b/cmd/compose/convert.go
@@ -47,6 +47,7 @@ type convertOptions struct {
 	services            bool
 	volumes             bool
 	profiles            bool
+	images              bool
 	hash                string
 }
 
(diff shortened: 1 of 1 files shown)


cmd/compose/convert.go
```diff
@@ -95,6 +99,7 @@ func convertCommand(p *projectOptions, backend api.Service) *cobra.Command {
 	flags.BoolVar(&opts.services, "services", false, "Print the service names, one per line.")
 	flags.BoolVar(&opts.volumes, "volumes", false, "Print the volume names, one per line.")
 	flags.BoolVar(&opts.profiles, "profiles", false, "Print the profile names, one per line.")
+	flags.BoolVar(&opts.images, "images", false, "Print the images names, one per line.")
```
mat007, line 102:
```suggestion
	flags.BoolVar(&opts.images, "images", false, "Print the image names, one per line.")
```
I think

ulyssessouza (approved):
LGTM. Just the nit from @mat007

mat007 (approved):

Issue #8935, opened 2021-11-17
Critical: compose v2.1.1+ loses `links` affinity for hostname named in links directive


rfay:
**Description**

For all of the history of compose until v2.1.1, if you used `links` then you could use the defined name without ambiguity.

For example, if you had a docker-compose.yaml with 
```yaml
services:
  web: 
    image: busybox
    command: tail -f /dev/null
    links:
    - db:db
  db:
    image: busybox
    command: tail -f /dev/null
```

You could use the `db` hostname in the web container and you'd know that you were getting the `db` service in this project.

In v2.1.1 (and maybe 2.1.0, but it was too broken to test), this doesn't work. 

**Steps to reproduce the issue:**

Demonstrated in https://github.com/rfay/compose-links-bug-demo for ease, but here's the description:

1. Create two directories, `one` and `two`.
2. In each directory add the simple docker-compose.yaml https://github.com/rfay/compose-links-bug-demo/blob/master/one/docker-compose.yaml
3. `docker network create test`
4. In each directory `docker-compose up` - You'll now have two identical services on the same network.
5. Use `docker exec one-web-1 ping -c1 db` to see the name resolution going on inside `one-web-1`. On compose v1 and up through v2.0.1 you'll see consistent name lookup, always coming out to be the `db` service one-db-1. On compose v2.1.1, you'll see it flip back and forth between `one-db-1` and `two-db-1` - it's inconsistent round-robin behavior, so you may have to hit `docker exec one-web-1 ping -c1 db` a few times to see it.

**Describe the results you received:**

Inconsistent round-robin behavior, with different docker-compose projects interfering with each other

**Describe the results you expected:**

As above, the name given in the `links` directive should result in consistent name lookup of the named host.

**Additional information you deem important (e.g. issue happens only occasionally):**

This has been reported by ddev users on a few platforms. I don't think it's platform dependent.

**Output of `docker compose version`:**

As above, tested on 2.0.1 and 2.1.1, and it was always fine in 1.x

rfay:
This problem remains exactly the same in compose 2.2.0.  This is critical, having to tell people to go back to compose v1. 

I may have to do a ddev release that won't allow use of docker-compose 2.1.1+

rfay:
Awesome, look forward to trying it out, thanks.

PR #8967, opened 2021-11-26
Fix links resolution

ulyssessouza:
**What I did**
Provide proper link resolution to avoid inter project ambiguity

**Related issue**
Resolves https://github.com/docker/compose/issues/8935

### DIFF
--- a/pkg/compose/convergence.go
+++ b/pkg/compose/convergence.go
@@ -448,7 +448,7 @@ func (s *composeService) createMobyContainer(ctx context.Context, project *types
 			Networks: inspectedContainer.NetworkSettings.Networks,
 		},
 	}
-	links := append(service.Links, service.ExternalLinks...)
+	links := s.getLinks(ctx, project.Name, service, number)
 	for _, netName := range service.NetworksByPriority() {
 		netwrk := project.Networks[netName]
 		cfg := service.Networks[netName]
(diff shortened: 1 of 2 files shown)


pkg/compose/convergence.go
```diff
@@ -476,6 +476,60 @@ func (s *composeService) createMobyContainer(ctx context.Context, project *types
...
+		l := linkServiceName
+		if len(linkSplit) == 2 {
+			l = linkSplit[1] // linkName if informed like in: "serviceName:linkName"
+		}
+		cnts, err := s.getContainers(ctx, projectName, oneOffExclude, true, linkServiceName)
+		if err != nil {
+			return nil
+		}
+		for _, c := range cnts {
+			containerName := strings.TrimPrefix(c.Names[0], "/")
```
ndeloof, line 497: can't you use getCanonicalContainerName ?

pkg/compose/convergence.go
```diff
@@ -476,6 +476,60 @@ func (s *composeService) createMobyContainer(ctx context.Context, project *types
...
+			links = append(links,
+				format(containerName, l),
+				format(containerName, strings.Join([]string{linkServiceName, strconv.Itoa(number)}, Separator)),
+				format(containerName, strings.Join([]string{projectName, linkServiceName, strconv.Itoa(number)}, Separator)),
+			)
+		}
+	}
+
+	if service.Labels[api.OneoffLabel] == "True" {
+		cnts, err := s.getContainers(ctx, projectName, oneOffExclude, true, service.Name)
```
ndeloof, line 507: please refactor into a func to avoid code duplication

ndeloof (approved):
LGTM with potential minor improvements

pkg/compose/convergence.go
ulyssessouza: Cool!
I was looking for that function but I couldn't remember the name. 
Thanks!

pkg/compose/convergence.go
ulyssessouza: Done with a local function

ndeloof (approved):

pkg/compose/convergence.go
```diff
@@ -476,6 +476,64 @@ func (s *composeService) createMobyContainer(ctx context.Context, project *types
...
+	for _, rawLink := range service.Links {
+		linkSplit := strings.Split(rawLink, ":")
+		linkServiceName := linkSplit[0]
+		l := linkServiceName
+		if len(linkSplit) == 2 {
+			l = linkSplit[1] // linkName if informed like in: "serviceName:linkName"
+		}
+		cnts, err := getServiceContainers(linkServiceName)
+		if err != nil {
+			return nil
```
mat007, line 498: Is this intentional or should it be
```
		return err
```
?

pkg/compose/convergence.go
```diff
@@ -476,6 +476,64 @@ func (s *composeService) createMobyContainer(ctx context.Context, project *types
...
+				format(containerName, strings.Join([]string{linkServiceName, strconv.Itoa(number)}, Separator)),
+				format(containerName, strings.Join([]string{projectName, linkServiceName, strconv.Itoa(number)}, Separator)),
+			)
+		}
+	}
+
+	if service.Labels[api.OneoffLabel] == "True" {
+		cnts, err := getServiceContainers(service.Name)
+		if err != nil {
+			return nil
```
mat007, line 513: Same here?

pkg/compose/convergence.go
```diff
@@ -476,6 +476,64 @@ func (s *composeService) createMobyContainer(ctx context.Context, project *types
...
+			l = linkSplit[1] // linkName if informed like in: "serviceName:linkName"
+		}
+		cnts, err := getServiceContainers(linkServiceName)
+		if err != nil {
+			return nil
+		}
+		for _, c := range cnts {
+			containerName := getCanonicalContainerName(c)
+			links = append(links,
+				format(containerName, l),
```
mat007, line 503: nit: maybe `l` should be `linkName` or something a bit clearer, just like l.528?

pkg/compose/convergence.go
ulyssessouza: That's the right return. The return type is `[]string`

mat007: But shouldn’t the return type be `([]string, error)`?

Issue #8969, opened 2021-11-26
volumes external causes wrong error in docker-compose 2.2.0


sinbino:
**Description**

**Steps to reproduce the issue:**
1. Create a volume with "docker volume create blah"
2. Define a volume with external: true in docker-compose.yml
```docker-compose.yml 
# in docker-compose.yml 
volumes: 
  blah:
    external: true
```
3. Run the docker-compose run

**Describe the results you received:**
```
volume "blah" already exists but was not created by Docker Compose. Use `external: true` to use an existing volume
```

**Describe the results you expected:**
No error occurs and the container starts.

**Additional information you deem important (e.g. issue happens only occasionally):**
The problem occurs with docker-compose 2.2.0, but not with 2.1.1.

**Output of `docker compose version`:**
```
Docker Compose version v2.2.0
```

**Output of `docker info`:**

```
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  app: Docker App (Docker Inc., v0.9.1-beta3)
  buildx: Build with BuildKit (Docker Inc., v0.6.3-docker)
  scan: Docker Scan (Docker Inc., v0.9.0)

Server:
 Containers: 4
  Running: 4
  Paused: 0
  Stopped: 0
 Images: 25
 Server Version: 20.10.11
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 1
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 io.containerd.runtime.v1.linux runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 7b11cfaabd73bb80907dd23182b9347b4245eb5d
 runc version: v1.0.2-0-g52b36a2
 init version: de40ad0
 Security Options:
  apparmor
  seccomp
   Profile: default
 Kernel Version: 5.11.0-40-generic
 Operating System: Ubuntu 20.04.3 LTS
 OSType: linux
 Architecture: x86_64
 CPUs: 8
 Total Memory: 15.47GiB
 Name: bravo
[...]

PR #8970, opened 2021-11-27
Don't check compose labels on external volumes

ndeloof:
**What I did**
- Removed volume label check for external volumes
- Don't create volume if marked external:true, but report error

**Related issue**
close https://github.com/docker/compose/issues/8969


glours (approved):
LGTM

ulyssessouza (approved):
LGTM

PR #8974, opened 2021-11-29
Return an error when failing to list containers

ulyssessouza:
**What I did**
Some refactoring... As proposed by @mat007  on https://github.com/docker/compose/pull/8967/files#r757920773


mat007 (approved):
Looks good, thanks!

Issue #8976, opened 2021-11-29
v2.2.0: volume already exists but was not created by Docker Compose. Use `external: true` to use an existing volume'


rfay:
**Description**

Starting a project in docker-compose v2.2.0/2.2.1 fails all the time with various messages like

> volume "ddev-global-cache" already exists but was not created by Docker Compose. Use `external: true` to use an existing volume'

This is a simple `ddev start` and can be replicated on any project with docker-compose 2.2.1 (which seems to have already made its way to Arch Linux)

**Steps to reproduce the issue:**

Use this docker-compose file:

```yaml
services:
  ddev-router:
    container_name: ddev-router
    environment:
      DISABLE_HTTP2: "false"
    healthcheck:
      timeout: 2m0s
      interval: 1s
      retries: 120
      start_period: 2m0s
    image: drud/ddev-router:v1.18.0
    networks:
      default: null
    ports:
    - mode: ingress
      host_ip: 127.0.0.1
      target: 443
      published: 443
      protocol: tcp
    - mode: ingress
      host_ip: 127.0.0.1
      target: 80
      published: 80
      protocol: tcp
    - mode: ingress
      host_ip: 127.0.0.1
      target: 8025
      published: 8025
      protocol: tcp
    - mode: ingress
      host_ip: 127.0.0.1
      target: 8026
      published: 8026
      protocol: tcp
    - mode: ingress
      host_ip: 127.0.0.1
      target: 8036
      published: 8036
      protocol: tcp
    - mode: ingress
      host_ip: 127.0.0.1
      target: 8037
      published: 8037
      protocol: tcp
    restart: "no"
    volumes:
    - type: bind
      source: /var/run/docker.sock
      target: /tmp/docker.sock
      read_only: true
      bind:
        create_host_path: true
    - type: volume
      source: ddev-global-cache
      target: /mnt/ddev-global-cache
      volume: {}
networks:
  default:
    name: ddev_default
    external: true
volumes:
  ddev-global-cache:
    name: ddev-global-cache
```

You'll get 
> volume "ddev-global-cache" already exists but was not created by Docker Compose. Use `external: true` to use an existing volume

**Describe the results you received:**

See above

**Describe the results you expected:**

docker-compose has worked with this type of file through all versions until 2.2.1.

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `docker compose version`:**

2.2.1

**Output of `docker info`:**

```
 docker info
Client:
 Context:    default
 Debug Mode: false
[...]

PR #8983, opened 2021-12-02
Turn external volume usage into a warning instead of erroring

ulyssessouza:
This avoids a compatibility break since returning an error would avoid the project to be started.

This avoids breaking projects using volumes created by older versions of Docker Compose.

**What I did**
Turn errors into warnings

**Related issue**
Resolves https://github.com/docker/compose/issues/8976

### DIFF
--- a/pkg/compose/create.go
+++ b/pkg/compose/create.go
@@ -1103,14 +1103,13 @@ func (s *composeService) ensureVolume(ctx context.Context, volume types.VolumeCo
 		return nil
 	}
 
-	// Volume exists with name, but let's double check this is the expected one
-	// (better safe than sorry when it comes to user's data)
+	// Volume exists with name, but let's double-check this is the expected one
 	p, ok := inspected.Labels[api.ProjectLabel]
 	if !ok {
-		return fmt.Errorf("volume %q already exists but was not created by Docker Compose. Use `external: true` to use an existing volume", volume.Name)
+		logrus.Warnf("volume %q already exists but was not created by Docker Compose. Use `external: true` to use an existing volume", volume.Name)
 	}
-	if p != project {
-		return fmt.Errorf("volume %q already exists but was not created for project %q. Use `external: true` to use an existing volume", volume.Name, p)
+	if ok && p != project {
+		logrus.Warnf("volume %q already exists but was not created for project %q. Use `external: true` to use an existing volume", volume.Name, p)
 	}
 	return nil
 }


ndeloof (approved):

rfay:
Note that this still has a problem with volumes shared between projects. For example, `ddev-global-cache` is shared among many projects. So there will be a perpetual warning (and if you turn it into an error in the future there will be no way around it)

ndeloof:
if it's shared, can't you make it an `external` volume ?

rfay:
Yes, I can. The question is, if it's marked as `external` will it automatically be created if it doesn't exist? If not, there will be pain.

ulyssessouza:
@rfay The thing here is that creating an external volume is incoherent with the very definition of what external is. This resource's life cycle should not be managed by Docker Compose. So I see its creation as a bug.
So mainly it's relying on an incoherent behaviour.

An example would be that the volume should have a certain group of Labels that Docker Compose does not have a way to guess. And by creating it automatically, the volume would be incoherent.

Using this as a feature includes that kind of problem and this check before creating corrects this behaviour.

About how to solve the problem, I don't know from your perspective in `ddev`'s code, but looking from outside, that would be fixed with a simple line before Docker Compose's usage:
```sh
$ docker volume create <volume_name>
```

rfay:
So here you're talking about another massive change of behavior, volumes that didn't exist were created. I don't mind that you want to have new behavior, but do it in a new product. Or give people a long time to work it out, deprecate the old behavior.

ndeloof:
"_massive change of behavior_" is high exaggeration.
I consider this to just be a sanitization of the compose behavior. This is unrelated to v1 <-> v2, this could have happened on V1 as well. As an oss maintainer on ddev I'm sure there's many places you made some design fixes which might be (ab-)used by some user and considered by them a "feature". so please keep calm, I'm happy to give you a hand and consider issues you report, but stop giving lessons on change management.

rfay:
Again, just deprecate behaviors you want to get rid of and people will gradually work around them. Then remove them after a few releases. docker-compose v1 was really quite stable for years. We're having a lot more trouble with v2, as you've noted.

ndeloof:
docker-compose v1 was dormant for years, so for sure was "stable" ...

rfay:
Sorry... we can use more predictability around here. If that means dormant, we'll take dormant. No need to go back and forth here. I'm happy to help with automated testing for docker-compose, and happy to review PRs. But please aim for predictability in your development work, and use deprecation (warn for a few versions, announce for a few versions, then remove feature) to get rid of things you can't live with. 

Starting in the next release of ddev a known-good docker-compose will be bundled with it so we're less susceptible to these changes, and when it's time to get a new one it can be fully tested. I think v2.0.1 may be the best candidate for bundling right now.

Issue #8977, opened 2021-11-29
Regression on compose logs -f frozen wrt Ctrl-C


Rogdham:
**Description**

Hello, this bug is a regression with `docker compose logs -f`.

It follows other similar bugs that have now been fixed, and appears in more edge cases (see  #8715/#8726 #8880/#8926 and #8749).

@ndeloof you fixed those issues, could you please take a look at this one as well? :pray: 

---

**Steps to reproduce the issue:**

1. `docker-compose.yml`:

```yaml
version: '2.2'

services:
  foo:
    image: alpine
    command: sh -c 'while true; do date; sleep 1; done'
  bar:
    image: alpine
    command: sh -c 'while true; do date; sleep 1; done'
```

2. Run in first shell:

```sh
docker compose up -d
docker compose logs -f
```
3. Run in second shell:

```sh
docker compose stop bar
docker compose restart foo
```
4. Look at the `logs -f` in first shell, and do a Ctrl^C on it

**Describe the results you received:**

The `logs -f` command in first shell does not show the logs of the new _foo_ container, and Ctrl^C does nothing (the process hangs).

**Describe the results you expected:**

The `logs -f` command in first shell should show the logs of the new _foo_ container, and be able to exit properly on  Ctrl^C

---

**Output of `docker compose version`:**

```
Docker Compose version 2.2.0
```

**Output of `docker info`:**

<details><summary>Click</summary>
<p>

#### yes, even hidden code blocks!

```python
print("hello world!")
```

```
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Build with BuildKit (Docker Inc., v0.6.1-docker)
  compose: Docker Compose (Docker Inc., 2.2.0)

Server:
 Containers: 0
  Running: 0
  Paused: 0
  Stopped: 0
 Images: 270
 Server Version: 20.10.11
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: false
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 io.containerd.runtime.v1.linux runc
 Default Runtime: runc
 Init Binary: docker-init
[...]

PR #8984, opened 2021-12-02
compose logs to notify printer about container lifecycle events

ndeloof:
**What I did**
`compose logs` watch container for lifecycle events and need to forward them to logPrinter so this one knows when all containers are gone, and stop.
(just like `up` does with the attach API)

**Related issue**
partial fix for https://github.com/docker/compose/issues/8977


ulyssessouza (approved):
LGTM

Issue #8945, opened 2021-11-02
docker --context default compose reports unknown docker command: "compose compose" for 2.0.1 but not 2.0.0


parity3:
I'm going to ignore the instructions due to lack of time. Please close this issue immediately if this does not benefit the community. I wasn't able to find this in previous issues which I find very surprising.

After installing 2.0.1 over 2.0.0 (via overwriting ~/.docker/cli-plugins/docker-compose with the release binary), the command: `docker --context default compose` reports the usage / help text but the bottom says:

> unknown docker command: "compose compose"

Ubuntu 20.10 x86_64 5.8.0-63-generic virtualbox-from-windows10
Server version 20.10.8
runc version: v1.0.1-0-g4144b63
Experimental: true

This "unknown docker command" happens for every docker compose command that passes in `--context default`. It does not happen if I leave out that switch.

If it's user error, the error message needs to speak more truth than accusing me of typing compose twice.

ndeloof:
Right, the issue is with [standalone command detection](https://github.com/ndeloof/compose/blob/c4cfaeb12a30b7e645eca4a66f0ac499ce3a4b82/cmd/compose/compose.go#L218-L220) which doesn't manage flags passed to the root `docker` command :'(

larskuhtz:
Is there any known work around for this issue?

W1M0R:
You could drop the `--context` parameter, and then export the DOCKER_HOST environment variable explicitly.

solvaholic:
> You could drop the `--context` parameter, and then export the DOCKER_HOST environment variable explicitly.

:hat-tip: Thank you @W1M0R :bow:

While this syntax got the error:

```text
% docker --host "$_host" compose --file svc01-stack.yml ps
...
unknown docker command: "compose compose"
```

This syntax succeeded:

```text
% DOCKER_HOST="$_host" docker compose --file svc01-stack.yml ps   
NAME                COMMAND             SERVICE             STATUS              PORTS
svc01-svc01-1       "/init"             svc01               running             
```

hacker:
Actually, setting `DOCKER_CONTEXT` also works.

eriktelepovsky:
facepalm...

internalsystemerror:
Why was this closed? The issue still exists on v2.2.1.

`docker --context remote compose version`

```
Run 'docker compose COMMAND --help' for more information on a command.
unknown docker command: "compose compose"
```

parity3:
[...]

PR #8985, opened 2021-12-02
Merge and fix Convert function from `docker/compose-switch`

ulyssessouza:
**What I did**
Note that it fixes the problem of double "compose" on calling the compose plugin with "docker" root flags.
Like in "docker --context default compose version"

This also removes the vendoring of the repo

**Related issue**
Resolves https://github.com/docker/compose/issues/8945

### DIFF
--- a/cmd/convert.go
+++ b/cmd/convert.go
@@ -1,0 +1,96 @@
+/*
+   Copyright 2020 Docker Compose CLI authors
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package main
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/docker/compose/v2/cmd/compose"
+)
+
+func getBoolFlags() []string {
+	return []string{
+		"--debug", "-D",
+		"--verbose",
+		"--tls",
+		"--tlsverify",
+	}
+}
+
+func getStringFlags() []string {
+	return []string{
+		"--tlscacert",
+		"--tlscert",
+		"--tlskey",
+		"--host", "-H",
+		"--context",
+		"--log-level",
+	}
+}
+
+func convert(args []string) []string {
+	var rootFlags []string
+	command := []string{compose.PluginName}
+	l := len(args)
+	for i := 0; i < l; i++ {
+		arg := args[i]
+		if arg[0] != '-' {
+			// not a top-level flag anymore, keep the rest of the command unmodified
+			if arg == compose.PluginName {
+				i++
+			}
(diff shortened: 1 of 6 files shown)


cmd/convert.go
```diff
@@ -0,0 +1,74 @@
...
+		"--tlscert",
+		"--tlskey",
+		"--host", "-H",
+		"--context",
+	}
+)
+
+func convert(args []string) []string {
+	rootFlags := []string{}
+	command := []string{}
```
ulyssessouza, line 29: This differs from the original on `docker/compose-switch` that used to initialise with 
```
command := []string{"compose"}
```

ndeloof:
please also import https://github.com/docker/compose-switch/blob/master/redirect/convert_test.go

cmd/convert.go
ndeloof: doing so, this just breaks the standalone usage :'(

ndeloof: and, by the way, we should setup the CI so it run all tests with both `docker compose` and `docker-compose`

ndeloof (changes requested):
removal of the `compose` pre-command breaks standalone usage.

ulyssessouza:
@ndeloof Ops... Sorry. My environment was just restarted after the update of DD. I was testing with one thinking it was the other...

ulyssessouza:
@ndeloof PTAL

cmd/convert.go
```diff
@@ -0,0 +1,95 @@
...
+		"--tlscert",
+		"--tlskey",
+		"--host", "-H",
+		"--context",
+		"--log-level",
+	}
+}
+
+func convert(args []string) []string {
+	rootFlags := []string{}
```
mat007, line 46:
```suggestion
	var rootFlags []string
```
[`nil` is great](https://github.com/golang/go/wiki/CodeReviewComments#declaring-empty-slices) 😁

mat007 (commented):
So should we revert the `--compatibility` flag from compose-switch after merging this?

cmd/convert.go
ulyssessouza: I also like. But I have people marking exactly the opposite on other PR's :D 
Nice to know I have a `nil`'s lover friend then!

ndeloof: I never understood go let you use `nil` slices just like it gets magically lazy-allocated, but don't do the same with `map` 🤨

mat007: Well, there is no `append` for maps :)

ulyssessouza: Actually the allocation comes from the built-in `append` function

mat007:
> So should we revert the --compatibility flag from compose-switch after merging this?

Ah actually that PR doesn’t add that flag here, for some reason I thought it did, based on whether we detected if we’re running standalone or not…

Never mind, ignore me 😬

cmd/convert.go
ndeloof: > there is no append for maps :)

not being an explicit function doesn't mean go can't be clever. See `range` which supports nil maps for example. Could do something clever with

cmd/convert.go
```diff
@@ -0,0 +1,96 @@
...
+		"--tlskey",
+		"--host", "-H",
+		"--context",
+		"--log-level",
+	}
+}
+
+func convert(args []string) []string {
+	var rootFlags []string
+	command := []string{compose.PluginName}
```
ndeloof, line 48: 👍

ndeloof (approved):

Issue #8810, opened 2021-10-19
Compose v2 does not respect the group_add field


JonnieCache:
**Description**

The `group_add` field is not respected by docker compose v2, whereas it works fine in 1.29

**Steps to reproduce the issue:**
1. Create a `docker-compose.yml` like so:
```services:
  test:
    image: alpine
    user: "1000:1000"
    group_add:
      - 20
```
    
2. Run it with `sudo docker compose run test id`

**Describe the results you received:**

With compose v2:
`uid=1000 gid=1000`

With compose v1:
`uid=1000 gid=1000 groups=20(dialout)`

**Describe the results you expected:**
Both should function the same.

**Additional information you deem important (e.g. issue happens only occasionally):**

The `user` field is used here purely to simplify the output, the problem occurs with and without it.

**Output of `docker compose version`:**

```
Docker Compose version v2.0.1
```

**Output of `docker-compose --version`:**

```docker-compose version 1.29.2, build unknown```

NOTE that while this may imply that I have both v1 and v2 installed at the same time, the issue occurs even if this is not the case.
The v1 build is installed via pip, as there is no release provided for arm64.

**Output of `docker info`:**

```
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  app: Docker App (Docker Inc., v0.9.1-beta3)
  buildx: Build with BuildKit (Docker Inc., v0.6.3-docker)
  compose: Docker Compose (Docker Inc., v2.0.1)

Server:
 Containers: 24
  Running: 1
  Paused: 0
  Stopped: 23
 Images: 22
 Server Version: 20.10.9
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 1
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 io.containerd.runtime.v1.linux runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 5b46e404f6b9f661a205e28d59c982d3634148f8
 runc version: v1.0.2-0-g52b36a2
 init version: de40ad0
 Security Options:
  apparmor
  seccomp
   Profile: default
 Kernel Version: 5.4.0-1044-raspi
[...]

PR #8987, opened 2021-12-03
Propagate GroupAdd from ServiceConfig to HostConfig

lancechentw:
**What I did**
The `group_add` key is parsed correctly from a compose file, but it is not
passed into the `ContainerCreate` API call, thus the configuration does
not take effect. This commit fixes the issue by propagating the
configuration from Docker compose's ServiceConfig to Docker container's
HostConfig.

**Related issue**
I think #8810 was not completely fixed when it was closed. The PR https://github.com/compose-spec/compose-go/pull/201 that was considered the fix to #8810 only solved the first half part of the issue, which is parsing `group_add` correctly, this PR solves the other half.


ndeloof (approved):

ndeloof:
nice, thanks

PR #8988, opened 2021-12-03
use filepath instead of path to check if the dockerfile path is abolute or not

glours:
**What I did**
use `filepath.IsAbs` instead of `path.IsAbs` to fix the Dockerfile path check on windows which doesn't work because `filepath.IsAbs` don't have a windows specific implementation.

image


thaJeztah (approved):
LGTM

should we have a test for this?

ndeloof (approved):

PR #8989, opened 2021-12-03
upgrade version of opencontainers/image-spec (security issue)

glours:
**What I did**
Upgrade version of `opencontainers/image-spec` to `1.0.2`

image


ndeloof:
Can we setup dependabot to manage this?

ndeloof (approved):

PR #8990, opened 2021-12-03
`compose images` should list images of created containers

### LINKED ISSUE #8806: Compose v2 images only report if container is running
Sample compose file:
```yaml
services:
  test:
    build:
      context: .
```

Sample Dockerfile:
```dockerfile
FROM ubuntu:latest

CMD ["sleep", "100"]
```

While container is running:
```
❯ docker compose images
Container           Repository          Tag                 Image Id            Size
test-test-1         test_test           latest              86f4fed4a4ee        72.9MB
```

While container is stopped:
```
❯ docker compose images
Container           Repository          Tag                 Image Id            Size
```

kiniou:
Signed-off-by: Kevin Roy <kiniou@gmail.com>

**What I did**

Like in compose v1, the `images` sub-command should list images of **created** services instead of just the **running** ones.

I think this is due to the docker containers api is returning the list of running containers by default without status filters.
Therefore, we need to add `All: true` to get back the v1 behaviour according to the v1 python source code in [cli/main.py](https://github.com/docker/compose/blob/6eb51d3b5b8bcb35474e9ecd45972813c5680096/compose/cli/main.py#L642) where `stopped=True` and ends in `all=True` in [project.py](https://github.com/docker/compose/blob/6eb51d3b5b8bcb35474e9ecd45972813c5680096/compose/project.py#L835).

**Related issue**

closes #8806

### DIFF
--- a/pkg/compose/images.go
+++ b/pkg/compose/images.go
@@ -33,6 +33,7 @@ import (
 
 func (s *composeService) Images(ctx context.Context, projectName string, options api.ImagesOptions) ([]api.ImageSummary, error) {
 	allContainers, err := s.apiClient.ContainerList(ctx, moby.ContainerListOptions{
+		All:     true,
 		Filters: filters.NewArgs(projectFilter(projectName)),
 	})
 	if err != nil {


ndeloof (approved):

ndeloof:
nice catch!

Issue #8994, opened 2021-12-04
[Feature Request] "convert --images" list only built images


abdennour:
**Description**

`docker compose convert --images` list all images.
This feature request is about listing only built images not all of them

**Proposal**

- **Option 1**
  - Overview
  Adding another flag for the subcommand convert called "--built-images"
  - Sample(s)
 `docker compose convert --built-images`
  - Advantages over other options
      Dedicated flag for a unique purpose

- **Option 2**
  - Overview
   Adding another flag for the subcommand convert called "--only-built" to be used alongside with the current flag "--images"
  - Sample(s)
 `docker compose convert --images --only-built`
  - Advantages over other options
      the new flag "--only-built" becomes as filter for any convert subcommand ( not only images)

Please let me know which one do you prefer and i can open PR for that

abdennour:
Please let me know which one do you prefer and i can open PR
@mat007  @glours @ndeloof   PFYI

ndeloof:
My preference goes to 
1. introduce actual subcommand `compose config images`
2. have this new subcommand declare its own flags. In terms of UX I'd prefer a simpler `--built` but that's to be debated
3. deprecate `--images` and redirect to this new subcommand
4. (do the same for other pseudo-commands in the `config` command for UX homogeneity

abdennour:
I like it! 
As you mentioned for `--built`, it should be debated. indeed, what if i want to render the opposite : ready/pulled images only.
@ndeloof  So that option should cover:
- all images (binary opt value: 11)
- only built images (binary opt value : 01)
- only ready images (binary opt value: 10) ( all images - only built images)

abdennour:
can we go **agile** : releasing `config --built-images` , then think about refactoring/re-organizing? @ndeloof

abdennour:
@ndeloof  i am still thinking with you.
This is another design:
- `BUILT_SERVICES=docker compose config --built-services` ( new option here to be released)
- then: `docker compose config --images $BUILT_SERVICES` ( no change here )

abdennour:
Reminder @ndeloof  !

abdennour:
I just realized that you already use the terminology `imagesToBuild`
https://github.com/docker/compose/blob/381df200105f902db9d9e7f109c19bbed58302cd/pkg/compose/build.go#L51

I will rename `--built-images` to `--images-to-build` , and after that, i think it's acceptable and the related PR #8995 can be merged

@ndeloof  @glours FYI

[...]

PR #8995, opened 2021-12-04
introduce config --images-to-build

abdennour:
**What I did**
Ability to filter images by services which uses built images
implement proposal of #8994 , exactly, option 1

**Related issue**

#8994

### DIFF
--- a/cmd/compose/compose.go
+++ b/cmd/compose/compose.go
@@ -212,11 +212,11 @@ func (o *projectOptions) toProjectOptions(po ...cli.ProjectOptionsFn) (*cli.Proj
 			cli.WithName(o.ProjectName))...)
 }
 
-const pluginName = "compose"
+const PluginName = "compose"
 
 // RunningAsStandalone detects when running as a standalone program
 func RunningAsStandalone() bool {
-	return len(os.Args) < 2 || os.Args[1] != manager.MetadataSubcommandName && os.Args[1] != pluginName
+	return len(os.Args) < 2 || os.Args[1] != manager.MetadataSubcommandName && os.Args[1] != PluginName
 }
 
 // RootCommand returns the compose command with its child commands
@@ -230,7 +230,7 @@ func RootCommand(backend api.Service) *cobra.Command {
 	)
 	command := &cobra.Command{
 		Short:            "Docker Compose",
-		Use:              pluginName,
+		Use:              PluginName,
 		TraverseChildren: true,
 		// By default (no Run/RunE in parent command) for typos in subcommands, cobra displays the help of parent command but exit(0) !
 		RunE: func(cmd *cobra.Command, args []string) error {
--- a/cmd/compose/convert.go
+++ b/cmd/compose/convert.go
@@ -50,7 +50,7 @@ type convertOptions struct {
 	profiles            bool
 	images              bool
 	hash                string
-	onlyBuilt           bool
+	builtImages         bool
 }
 
 func convertCommand(p *projectOptions, backend api.Service) *cobra.Command {
@@ -87,7 +87,7 @@ func convertCommand(p *projectOptions, backend api.Service) *cobra.Command {
 			if opts.profiles {
 				return runProfiles(opts, args)
 			}
-			if opts.images {
+			if opts.images || opts.builtImages {
 				return runConfigImages(opts, args)
 			}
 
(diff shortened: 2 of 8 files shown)


abdennour:
@ndeloof  
> First-time contributors need a maintainer to approve running workflows.
please help to approve @ndeloof @mat007 @glours 

image

ndeloof:
code looks good, but I'm not fan for adding an option that only applies to the `--config` pseudo-command (which should be a subcommand), as this would be confusing and not apply to other (pseudo-)commands.
Maybe time to introduce `compose config images` as a **real** subcommand which could define it's own options?

glours (changes requested):
It doesn't work as defined in the scenario 2 of #8995 
```sh
> ./bin/docker-compose compose -f ~/sources/awesome-compose/nginx-golang/docker-compose.yml convert --only-built
services:
  backend:
    build:
      context: /Users/glours/sources/awesome-compose/nginx-golang/backend
      dockerfile: Dockerfile
    networks:
      default: null
  frontend:
    build:
      context: /Users/glours/sources/awesome-compose/nginx-golang/frontend
      dockerfile: Dockerfile
    depends_on:
      backend:
        condition: service_started
    networks:
      default: null
    ports:
    - mode: ingress
      target: 80
      published: 80
      protocol: tcp
networks:
  default:
    name: nginx-golang_default
> ./bin/docker-compose compose -f ~/sources/awesome-compose/nginx-golang/docker-compose.yml convert --volumes --only-built
~/sources/compose 8994-convert-built-images >
```

I was able to use the flag alone and I was also able to pass it to an unrelated flag such as `--volumes`

abdennour:
Scenario 2 requires a lot of development. 
- Should i finish the combination with other flags one by one ?
- Or just revert to Scenario 1 ?
- Or go with new scenario of @ndeloof  (`docker compose config images` ) ?

@glours @ndeloof  any team decision, so i can proceed?

ndeloof:
A possible issue with my proposed solution is that `compose config` accept a list of `services` as arguments, and introducing sub-commands would conflict with this usage or forbid use of some reserved names as services :'(

No strong opinion yet on the best approach, need to think twice about it

abdennour:
Yep! @ndeloof  , Think about `--only-built` as adjective for services , as global option.
that time, filtering will happen at the level of `project.Services` or `project.WithServices`
Does it make sense!

ndeloof:
Got it, make sense.

abdennour:
Also, @ndeloof we can go **agile** : releasing `config --built-images` , then think about refactoring/re-organizing.

abdennour:
@ndeloof  @glours , since adding new flag depends on `--images` flag requires redesign, i just pushed now the option (1) which is a new flag without dependency to others

`docker compose config --built-images`

I would suggest to accept this feature as incremental release , and with motivation to go AGILE.
Later on, we can think about refactoring this subcommand , indeed, you have already `--resolve-image-disgests`, which should belong to subsubcommand images since long time

glours:
@abdennour 

I don't think we should accept this incremental step and I'll explain why
Compose is a wildy used project, if we introduce a new command or a new flag this new feature will be quickly used to script some bots or used in CI flows and then we'll be block (or it'll a least been painful) to remove this "temporary" increment. 
This is typically the kind of behaviour which drive a project to feature overload.

I really prefer that we took few days to think to a proper solution instead of rushing.

ndeloof:
I tend to agree. Agile != "let's add random stuff and change our mind any time", especially when it comes to UX. Once we introduce a new flag, it will be there forever.

abdennour:
Totally agree @glours @ndeloof , this is INTERFACE design pattern & it should be stable by Design.
BTW, i am not blocked @glours , as i forked it and i released the feature in the fork: https://github.com/ElmCompany/compose/releases/tag/v2.2.3

abdennour:
Reminder @ndeloof @glours

abdennour:
@ndeloof  @glours  i made the change as per my comment here https://github.com/docker/compose/issues/8994#issuecomment-996905644
It should be acceptable now.
Let me know if you need any change

abdennour:
I really want to know if you need anything in order to approve/merge/release this PR.
@ndeloof @glours Kind reminder !!

abdennour:
@glours You are requested for review.
@ndeloof  who is the backup of @glours

abdennour:
I really want to know if you need anything in order to approve/merge/release this PR.
@ndeloof @glours  Kind reminder !!

glours:
Hello @abdennour 

Sorry for the delay of the response.
I still prefer the option of having a dedicated `compose config images` [option mentioned](https://github.com/docker/compose/issues/8994#issuecomment-986009381) by @ndeloof 
With the holidays period we didn't have time to talk about this option internally ATM.
cc @ulyssessouza FYI

Issue #8915, opened 2021-11-10
I miss the "building" output statement in compose v2


rfay:
**Description**

compose v2 told you when it was doing the "build" step in docker-compose.yaml - it output "building: web" or something. Just seeing it go by and seeing how long it took helped you know whether the build step was necessary (whether it had already been done) 

**Steps to reproduce the issue:**

With a docker-compose.yaml including a build step, do a `docker-compose up`, In compose v1 it told you when it was doing the build step, in compose v2 it doesn't.

PR #8998, opened 2021-12-04
add info about building service(s) and the warning to explain how to …

glours:
…rebuild those image service(s)

**What I did**
Add logs to inform that we start building service(s) image(s)
Add a warning log to explain how to rebuild those images

**Related issue**
#8915 

image

### DIFF
--- a/pkg/compose/build_buildkit.go
+++ b/pkg/compose/build_buildkit.go
@@ -18,7 +18,9 @@ package compose
 
 import (
 	"context"
+	"fmt"
 	"os"
+	"strings"
 
 	"github.com/compose-spec/compose-go/types"
 	"github.com/docker/buildx/build"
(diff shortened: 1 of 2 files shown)


ndeloof:
Seems you can avoid code duplication by moving this code to composeService#build just before as buildOptions are computed. This would also avoid hack-ish retrieval of the service name using `_` split which is incomplete:
- service can be configured with an explicit image 
- project could be a_b and service c_d, which results into image = a_b_c_d and you're stuck

Another option is to refactor `doBuild` so we pass a `map[service]build.Options` so we keep track of service name, I'm not sure how exactly buildx uses the map key, if this is just for login this would even have no impact

ndeloof (approved):

pkg/compose/build_buildkit.go
```diff
@@ -46,6 +48,12 @@ func (s *composeService) doBuildBuildkit(ctx context.Context, project *types.Pro
 	defer cancel()
 	w := xprogress.NewPrinter(progressCtx, os.Stdout, mode)
 
+	var servicesName []string
+	for name := range opts {
+		servicesName = append(servicesName, name[strings.LastIndex(name, "_")+1:])
```
mat007, line 53: Isn’t the separator `"-"` or `"_"` depending on `--compatibility`?

ndeloof: indeed, see my earlier comment on this being a fragile approach

Issue #184, opened 2014-04-06
Watch code and automatically rebuild when something changes


bfirsh:
Compose could watch your code and automatically kick off builds when something changes. This would mean:
- Code can be reloaded automatically regardless of language. This would also make code reloading work on compiled languages if that was part of the build process!
- Volumes are no longer needed in development
- Production server can be used in development (e.g. gunicorn can be used to serve Django apps in development)

It could either be an option to `up` (`docker-compose up --watch`), an option in the config file specifying directories to watch/ignore, or perhaps a separate command (`docker-compose watch`).

Thanks to @samalba for the suggestion!

txdavidtx:
It's been a few months, and I couldn't tell from the merge referenced above, was a "watch" ability integrated?

aanand:
No, `fig watch` has not been implemented.

aanand:
It's still on the table though - see [ROADMAP.md](https://github.com/docker/fig/blob/master/ROADMAP.md)

bfirsh:
An idea from @fxdgear: instead of rebuilding, it could also be useful to just restart the container when the code changes. E.g. when you've mounted your code with a volume.

aanand:
Now that we have a `--no-build` flag, perhaps the logic should be:

```
$ fig up                      # current behaviour
$ fig up --watch              # watch and rebuild + restart
$ fig up --watch --no-build   # watch and restart
```

That said, if we go down the #693 path, we'll have to revisit this.

amjibaly:
:+1: 

For this to be really awesome, there's got to be a way for docker build to accept or pass through some type of credentials without storing it in the image. I'm thinking private github modules but it really applies to any environment.

docteurklein:
what about the right tool for the right job ?

wouldn't `guard` fit perfectly here ?

``` ruby
guard :shell do
  watch(/Dockerfile$/) { `fig build` }
  watch(/fig.yml$/) { `fig up -d` }
end
```

ches:
@amjibaly:

> For this to be really awesome, there's got to be a way for docker build to accept or pass through some type of credentials without storing it in the image. I'm thinking private github modules but it really applies to any environment.
> ...
> I run fig up --watch and the first time npm install runs because the package.json hasn't been cached yet. This needs to fetch private modules from github so it passes those through (maybe this is setup in my ~/.figconfig or something similar?)

[...]

PR #9000, opened 2021-12-05
Introduce `compose watch` to rebuild/restart service on change to build context

ndeloof:
**What I did**
Introduced `compose watch` command.
This command watch filesystem (using inotify) to rebuild and restart container when build context is updated for a service

**Implementation details** 
uses `inotify` to get filesystem changes as an event stream
on first change detected, applies a quiet period as such events comes in a row (editing a file in vscode results in CHMOD+WRITE+CHMOD events)
refresh (rebuild+recreate) only run once in parallel using a 1-size bool channel (could have used sync.Once but this one can't be reset)

**Related issue**
close https://github.com/docker/compose/issues/184

image

### DIFF
--- a/cmd/compose/compose.go
+++ b/cmd/compose/compose.go
@@ -305,6 +305,7 @@ func RootCommand(backend api.Service) *cobra.Command {
 		pullCommand(&opts, backend),
 		createCommand(&opts, backend),
 		copyCommand(&opts, backend),
+		watchCommand(&opts, backend),
 	)
 	command.Flags().SetInterspersed(false)
 	opts.addProjectFlags(command.Flags())
--- a/cmd/compose/watch.go
+++ b/cmd/compose/watch.go
@@ -1,0 +1,57 @@
+/*
+   Copyright 2020 Docker Compose CLI authors
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package compose
+
+import (
+	"context"
+
+	"github.com/spf13/cobra"
+
+	"github.com/docker/compose/v2/pkg/api"
+)
+
+type watchOptions struct {
+	*projectOptions
+}
+
+func watchCommand(p *projectOptions, backend api.Service) *cobra.Command {
+	opts := watchOptions{
+		projectOptions: p,
+	}
+	cmd := &cobra.Command{
+		Use:   "watch [SERVICES...]",
+		Short: "Watch build context for service(s) and rebuild/refresh containers when files are updated",
+		PreRunE: Adapt(func(ctx context.Context, args []string) error {
+			return nil
+		}),
+		RunE: Adapt(func(ctx context.Context, args []string) error {
+			return runWatch(ctx, backend, opts, args)
+		}),
+		ValidArgsFunction: serviceCompletion(p),
+	}
+
+	return cmd
+}
+
+func runWatch(ctx context.Context, backend api.Service, opts watchOptions, services []string) error {
+	project, err := opts.toProject(services)
+	if err != nil {
+		return err
+	}
+
+	return backend.Watch(ctx, project, api.WatchOptions{})
+}
--- a/go.mod
+++ b/go.mod
@@ -16,6 +16,7 @@ require (
 	github.com/docker/docker v20.10.7+incompatible
 	github.com/docker/go-connections v0.4.0
 	github.com/docker/go-units v0.4.0
+	github.com/fsnotify/fsnotify v1.5.1
 	github.com/golang/mock v1.5.0
 	github.com/hashicorp/go-multierror v1.1.0
 	github.com/hashicorp/go-version v1.3.0
@@ -107,7 +108,7 @@ require (
 	golang.org/x/crypto v0.0.0-20210322153248-0c34fe9e7dc2 // indirect
 	golang.org/x/net v0.0.0-20210405180319-a5a99cb37ef4 // indirect
 	golang.org/x/oauth2 v0.0.0-20210402161424-2e8d93401602 // indirect
-	golang.org/x/sys v0.0.0-20210616094352-59db8d763f22 // indirect
+	golang.org/x/sys v0.0.0-20210630005230-0f9fa26af87c // indirect
 	golang.org/x/term v0.0.0-20210220032956-6a3ed077a48d // indirect
 	golang.org/x/text v0.3.5 // indirect
 	golang.org/x/time v0.0.0-20210220033141-f8bda1e9f3ba // indirect
--- a/go.sum
+++ b/go.sum
@@ -419,8 +419,9 @@ github.com/felixge/httpsnoop v1.0.1/go.mod h1:m8KPJKqk1gH5J9DgRY2ASl2lWCfGKXixSw
 github.com/form3tech-oss/jwt-go v3.2.2+incompatible/go.mod h1:pbq4aXjuKjdthFRnoDwaVPLA+WlJuPGy+QneDUgJi2k=
 github.com/frankban/quicktest v1.11.3/go.mod h1:wRf/ReqHper53s+kmmSZizM8NamnL3IM0I9ntUbOk+k=
 github.com/fsnotify/fsnotify v1.4.7/go.mod h1:jwhsz4b93w/PPRr/qN1Yymfu8t87LnFCMoQvtojpjFo=
-github.com/fsnotify/fsnotify v1.4.9 h1:hsms1Qyu0jgnwNXIxa+/V/PDsU6CfLf6CNO8H7IWoS4=
 github.com/fsnotify/fsnotify v1.4.9/go.mod h1:znqG4EE+3YCdAaPaxE2ZRY/06pZUdp0tY4IgpuI1SZQ=
+github.com/fsnotify/fsnotify v1.5.1 h1:mZcQUHVQUQWoPXXtuf9yuEXKudkV2sx1E06UadKWpgI=
+github.com/fsnotify/fsnotify v1.5.1/go.mod h1:T3375wBYaZdLLcVNkcVbzGHY7f1l/uK5T5Ai1i3InKU=
 github.com/fullsailor/pkcs7 v0.0.0-20190404230743-d7302db945fa/go.mod h1:KnogPXtdwXqoenmZCw6S+25EAm2MkxbG0deNDu4cbSA=
 github.com/fvbommel/sortorder v1.0.1 h1:dSnXLt4mJYH25uDDGa3biZNQsozaUWDSWeKJ0qqFfzE=
 github.com/fvbommel/sortorder v1.0.1/go.mod h1:uk88iVf1ovNn1iLfgUVU2F9o5eO30ui720w+kxuqRs0=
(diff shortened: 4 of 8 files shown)


thaJeztah:
I think we should look closely at doing a good design for this if we want to implement this; this can easily lead to opening a can of worms, e.g. consider a project with node_modules, or (subdirectories) with generated files that uses bind-mounts; we can easily end up in an infinite loop (file chance triggers `docker compose up`, triggers container to be started, causes files to be updated, causes `docker compose watch` to be triggered). Other things, like `.dockerignore` / `some/Dockerfile.dockerignore`, etc also will have to be taken into account.

I can see this being useful for some scenarios, but there are existing tools to trigger commands on filesystem changes, and I think we need to have a good design to prevent opening a can of worms.

ndeloof:
Agree there might be corner cases, so we could release this with an EXPERIMENTAL warning on command, so that users don't expect too much magic to be addressed, and also we collect real-world feedback and can investigate how to address concrete issues, vs trying to guess

Mahoney:
Hi - I put my work on https://github.com/Mahoney/docker-compose-watch on hold when I saw this PR start, but it looks a bit uncertain at the moment - could anyone give a view on whether picking my work up again would be a waste of time?

ndeloof:
I propose:
1. to mark this feature as `EXPERIMENTAL` so users don't have huge expectations and we can change our mind based on feedback from the trenches.
2. add an `--exlude` flag to this `watch` command to be added to the `.dockerignore` rules, to make it possible to workaround infinite loop scenario described on https://github.com/docker/compose/pull/9000#issuecomment-988612328

thaJeztah:
I'm somewhat "-1" on this; at least not without thinking it through. This can be a can of worms to get it right, and I don't think a `--exclude` flag will be sufficient to express most cases, bare the "very simple" ones, and before we know it, we'll end up with either another configuration file (e.g. `.watchconfig`), then requests to support exclude files that "are already in the repository" (`.gitignore`, `.dockerignore`), then to be able to define it within the compose file, or through an environment variable, etc., etc.

Having a quick glance over the ticket (https://github.com/docker/compose/issues/184), I see various similar concerns being raised. There's existing solutions for things like this (`inotifywait`, `entr`, `nodemon`, `gulp` watch), some of which are mentioned on https://github.com/docker/compose/issues/184 (so may already be used by users). So it's not "impossible" to do this currently, and if we want to add this to the core functionality of Docker Compose, it should be significantly better than the existing solutions, and/or provide detection/heuristics that only Compose itself can provide (and that are not possible through other means).

Mahoney:
FWIW The things I'm finding (or anticipate finding!) it a bit painful to reimplement in an external program, and which might justify building it into the docker compose codebase, are:

1. Establishing the build contexts of all the containers - requires merging all the yaml files in exactly the same way compose does, then reading the yaml to find the `build.context` path to make sure I'm watching all the right paths. 
2. Interpreting `.dockerignore` so that changes to files that would never get sent to the daemon don't trigger anything - this seems like the easiest way to minimise triggering a build unnecessarily. buildkit does a pretty good job of reducing unnecessary work, but just transferring the context to the daemon can be costly, and I don't _think_ that buildkit caches the context & checks client side if the context has changed since the last build. If it did then I could leave it to buildkit.
3. Maintaining the docker compose CLI
4. Queuing & debouncing changes to re-trigger any build in progress when they happened when it completes, so that they don't get mussed

In cases 1-3 I'm conscious that I'm repeating logic that must already exist in the docker codebase, and in some cases is trickier than it looks to get right.[1]

(4) has two facets to it - queuing in bash seems hard (I know, don't use bash...) but more fundamentally recognising the state of a service in a docker compose app to know if it is currently rebuilding/redeploying or not may be hard to do from the outside.

At the moment on change I do a `docker compose up -d --build <container(s) whose context changed>` in a background process whilst the foreground process is doing an attached `docker compose up`. This seems to work quite nicely from a debounce perspective, in that it seems if it's in the middle of a rebuild/redeploy of service X a background call to `docker compose up -d --build X` does not mean it starts again - though unless it's queued (not sure on that one, suspect not) this could mean you end up out of sync and need to kick it again.

I haven't thought through the infinite loop issue yet, need to work out the simplest possible test that demonstrates it.

[1] For instance:
* the CLI argument parsing has lots of little gotchas to reimplement - it allows `-fpathwithoutspacebeforeit` for instance, and `-f` and other flags that take arguments have to be before the command, so you need to know all the flags that take arguments before the command in order to recognise the command
* the precise format of `.dockerignore` I suspect has lots of ways you could get it wrong - is it go specific glob expressions?

thaJeztah:
Thanks for that! Yes, I think the complications you describe are at least part of what should be looked at, as those are the things that a (potential) `watch` command for compose should address. Without that, there may not be much value over other solutions.

- w.r.t. `.dockerignore`, yes, it might be difficult to handle this exactly as `docker build` itself handles those. Unfortunately the exact behavior of the ignore was somewhat underdefined when it was originally implemented; there's some efforts on making it align more with `.gitignore`, but it's difficult to do so without breaking existing users. So from that perspective, ideally there would be a way to know if `docker build` would consider a file to be included or excluded (which could be something similar to [`git check-ignore`](https://git-scm.com/docs/git-check-ignore) also see https://github.com/moby/moby/issues/40319).
- In addition to a `.dockerignore`, `docker build` (when backed by BuildKit) supports per-Dockerfile ignore (`some/dir/Dockerfile.dockerignore`), so a different ignore file may be in use for different services.
- Even without a `.dockerignore`, it's difficult to know which files will affect a build. The Dockerfile may depend on just a single file `COPY a-file.txt .` (or no files at all), which means that no rebuild should be triggered unless that one file changes. Things can get more complicated if the build depends on build-args or multi-arch conditions.

The above easily gets out of scope for Docker Compose itself, and may be better handled by BuildKit (and exposed through `docker build` / `buildx`). That said, Compose may be aware of dependencies _between_ services (if service A is being triggered to rebuild, service B should also be rebuilt or restarted, e.g.), so possibly a combination of both. I know there were some ideas to have a `docker build --watch` option at some point (but I don't think those ever got further than an internal "hack day" PoC).

There may be other questions, such as:

- Do I want `watch` to _rebuild_ my images, or only _restart_ (or _recreate_) my containers? In "inner loop" situations, it may be that I bind-mount my source (and only rebuild images periodically).
- Both `docker compose up` and `docker compose build` have many command-line options; when automating using `watch`, those would (likely) all be available as well.
- The alternative to the previous one could be to add a `--watch` flag to those commands, which would re-run the command with the same options on file-changes (but we're then back at "which files" to watch, and "which services")
- Some of the behavior may make sense to be defined inside the compose file as part of the service (whether or not it should be rebuilt, or restarted, and which files to watch for, which in some cases may be a subset of the files that are _actually_ used (e.g., don't bother rebuilding if I modify file-X, but _do_ trigger a rebuild if my JavaScript must be recompiled))

ndeloof:
ok, so no consensus on this feature, I'm giving up.

Mahoney:
Shoot - I was hoping a limited scope approach inside docker compose along the lines I was proposing would work.

ndeloof:
feel free to pick up my branch and iterate, I don't hav much time to spend debating an acceptable scope for a new feature

Mahoney:
I don't know quite why, but I can't bring myself to learn go! I'll return to hacking on my own little tool.

hholst80:
This is extremely wasteful. Rebuilding an entire container every time a single file is changed? Literally boiling the oceans. 

Besides that, there are much tooling that exists already that can produce this in conjunction with docker-compose if you really need to boil up the Atlantic ocean for your breakfast eggs.

The potential for feature creep is GIGANTIC.

If it is on-boarded someone will need to support it going forward.

This PR did however address the OLDEST STILL OPEN Issue on github for compose. And it has been liked hundred of times by users of compose. At least they were users at some point. If a tool wants to stay relevant it needs to cater to the needs of the users - or loose.

I think it is a tough judgement case. I hate the feature but I would probably accept it just because the dumb users really wants it.

ndeloof:
> Rebuilding an entire container every time a single file is changed...

Some cool technical stack do support hot-reload on source change, but some don't. Typically a `go` or `java` backend need to be re-built and restarted. Build might benefit from some caching, and `Dockerfile` should be optimized accordingly to avoid a full rebuild.

hholst80:
For us the best practice here has been to launch the container with `sleep inf` and then exec a watchdog process that monitors the mounted source code and reloads if it changed. Reloading the container would not work well at all with VS Code workflow, but it was also not a very good solution for my team back in 2017-18 when I tried that approach.

This works best if docker is running rootless for the desktop user.

```
dev:
  image: node:16
  command: sleep inf
  volumes:
    - $HOME:$HOME
  environment:
    - HOME
  working_dir: $PWD
```

```
user@host $ docker-compose up -d
user@host $ docker-compose exec dev bash
root@container $ cat start.sh
#!/bin/sh
exec npx nodemon --ignore data
root@container $ ./start.sh
```

patientplatypus6:
> > Rebuilding an entire container every time a single file is changed...
> 
> Some cool technical stack do support hot-reload on source change, but some don't. Typically a `go` or `java` backend need to be re-built and restarted. Build might benefit from some caching, and `Dockerfile` should be optimized accordingly to avoid a full rebuild.

Yeah - the criticism is completely unfounded. Some programs need to be compiled first before they can be run, so the entire container needs to be restarted so the program can be rerun. Obviously.

ndeloof:
@hholst80 I wonder: why don't you declare the `command` on this service to be `npx nodemon --ignore data` so you don't need to `exec` into container?

hholst80:
> @hholst80 I wonder: why don't you declare the `command` on this service to be `npx nodemon --ignore data` so you don't need to `exec` into container?
> 

That's leaving out npm install and other stuff that might be required to start the dev.

patientplatypus6:
Here's a right up of how I solved this problem for my application. Maybe it will be useful for somebody else.

Here's my `docker-compose.yml` file: 

```
version: "3.9"

volumes:
  volume1:

services:
  db:
    restart: always
    image: postgres
    volumes:
      - ./docker-entrypoint-initdb.d/init.sql:/docker-entrypoint-initdb.d/init.sql
      - ./data/db:/var/lib/postgresql/data
    environment:
      - POSTGRES_NAME=dev-postgres
      - POSTGRES_USER=pixel
      - POSTGRES_DATABASE=lightchan
      - POSTGRES_PASSWORD=stardust 
    ports: 
      - "5432:5432"
    healthcheck:
      test: ["CMD-SHELL", "pg_isready -U pixel"]
      interval: 5s
      timeout: 5s
      retries: 20
  backend:
    build: .
    restart: always
    volumes:
      - volume1:/code
    command: sh -c "./entrypoint.sh"
    ports:
      - "8000:8000"
    environment:
      - POSTGRES_NAME=dev-postgres
      - POSTGRES_USER=pixel
      - POSTGRES_DATABASE=lightchan
      - POSTGRES_PASSWORD=stardust
      - POSTGRES_HOST=db  
    depends_on:
      db:
        condition: service_healthy 
  watchvolume:
    build: ./watchvolume
    volumes: 
      - '/var/run/docker.sock:/var/run/docker.sock'
      - volume1:/code
    restart: always
    command: sh -c "./entrypoint.sh"
    depends_on:
      - 'backend'
  frontend:
    build: ../../
    restart: always
    volumes:
      - .:/app
    ports: 
      - "3000:3000"
    depends_on:
      - "backend"
  nginx:
    image: nginx
    ports:
      - "9000:80"
    volumes:
      - volume1:/usr/share/nginx/html
    links:
      - "backend"
      - "db"
 ```

The `watchvolume` container is copying the socket for docker so it can restart sibling containers from it's own dockerfile. The dockerfile for `watchvolume` is (from looking through stackoverflow posts I finally found it here: https://stackoverflow.com/questions/49713054/restart-container-from-inside): 

```
FROM ubuntu:latest

ENV DOCKER_VERSION='20.10.7'
RUN apt-get update 
RUN apt-get install curl -y
# RUN apt-get install systemd -y

RUN set -ex \
    && DOCKER_FILENAME=https://download.docker.com/linux/static/stable/x86_64/docker-${DOCKER_VERSION}.tgz \
    && curl -L ${DOCKER_FILENAME} | tar -C /usr/bin/ -xzf - --strip-components 1 docker/docker

```

with the `entrypoint.sh` file looks like this (it creates two daemon instances) : 

```
#!/bin/bash

./watchfront.sh &
./watchback.sh &
```

Which call 

```
#!/bin/bash

fswatch -or ../../lightone/*.py ../../lightchan/*.py ../../media ../../static | xargs -n1 -I{} docker restart lightchan-backend-1
```

and 

```
#!/bin/bash

fswatch -or ../../../src ../../../public | xargs -n1 -I{} docker restart lightchan-frontend-1
```

The repo is at github.com/patientplatypus6/lightchan.

hholst80:
> so the entire container needs to be restarted so the program can be rerun. Obviously.

No, that is not a valid conclusion. The application should not be running in the main thread and even if it would run like that you can wrap the restart in a watch-build-run script without destroying the container environment.

A container is an environment not an application packaging device. With the former view it is most reasonable to have a watch-and-rebuild process running without restarting (destroying / creating) the container environment. That process leaves lots of collateral damage say if you are using VS code remote container for your local dev workflow.

patientplatypus6:
Neat. Say I have a container that contains a C++ program running on the main thread that compiles, builds and runs. I change a file I would like it to build and rerun. Would it be nice to have some nifty utility that cc's into the container, checks which files are tagged to which processes and reruns only those processes? Yeah sure. It's also complicated. As you mentioned.

ndeloof:
Based on feedback on this PR, I've created https://github.com/compose-spec/compose-spec/pull/253 as a proposal to add support for descriptive policy to apply to services, so that compose can adopt the most relevant approach depending on stack capability to hot-reload on code change.

PR #9004, opened 2021-12-06
compose cp doesn't need a full project and can copy from stopped containers

ndeloof:
**What I did**
allow `compose cp` to use stopped container
doesn't require original compose file, just project name

**Related issue**
close #????
cc @carianmcnulty


ndeloof:
cc @ciaranmcnulty

ciaranmcnulty:
See #9005

ulyssessouza (approved):
LGTM

PR #9007, opened 2021-12-07
Add 2 modes test mechanism

ulyssessouza:
**What I did**
This should test on plugin and standalone mode

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,8 +42,16 @@ compose-plugin: ## Compile the compose cli-plugin
 	--output ./bin
 
 .PHONY: e2e-compose
-e2e-compose: ## Run End to end local tests. Set E2E_TEST=TestName to run a single test
-	gotestsum $(TEST_FLAGS) ./pkg/e2e -- -count=1
+e2e-compose: ## Run end to end local tests in plugin mode. Set E2E_TEST=TestName to run a single test
+	go test $(TEST_FLAGS) -count=1 ./pkg/e2e
+
+.PHONY: e2e-compose-standalone
+e2e-compose-standalone: ## Run End to end local tests in standalone mode. Set E2E_TEST=TestName to run a single test
+	go test $(TEST_FLAGS) -count=1 --tags=standalone ./pkg/e2e
+
+
+.PHONY: e2e
+e2e: e2e-compose e2e-compose-standalone ## Run end to end local tests in both modes. Set E2E_TEST=TestName to run a single test
 
 .PHONY: cross
 cross: ## Compile the CLI for linux, darwin and windows
--- a/pkg/e2e/framework.go
+++ b/pkg/e2e/framework.go
@@ -202,12 +202,9 @@ func (c *E2eCLI) RunDockerCmd(args ...string) *icmd.Result {
 	return res
 }
 
-const composeStandaloneEnvVar = "COMPOSE_STANDALONE_MODE"
-
 // RunDockerComposeCmd runs a docker compose command, expects no error and returns a result
 func (c *E2eCLI) RunDockerComposeCmd(args ...string) *icmd.Result {
-	composeStandaloneMode := os.Getenv(composeStandaloneEnvVar)
-	if composeStandaloneMode == "true" || composeStandaloneMode == "1" {
+	if composeStandaloneMode {
 		composeBinary, err := findExecutable(DockerComposeExecutableName, []string{"../../bin", "../../../bin"})
 		assert.NilError(c.test, err)
 		res := icmd.RunCmd(c.NewCmd(composeBinary, args...))
--- a/pkg/e2e/framework.go
+++ b/pkg/e2e/framework.go
@@ -29,6 +29,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/docker/compose/v2/cmd/compose"
 	"github.com/pkg/errors"
 	"gotest.tools/v3/assert"
 	is "gotest.tools/v3/assert/cmp"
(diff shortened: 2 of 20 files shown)


pkg/e2e/main_test.go
```diff
@@ -0,0 +1,37 @@
...
+import (
+	"os"
+	"testing"
+
+	"github.com/sirupsen/logrus"
+)
+
+func TestMain(m *testing.M) {
+	logrus.Info("Running tests on COMPOSE_STANDALONE_MODE=true")
+	_ = os.Setenv(composeStandaloneEnvVar, "true")
```
mat007, line 28: Can’t we do this using a command line flag and a (global) variable instead?

ndeloof (approved):

ndeloof:
🏆

pkg/e2e/main_test.go
ulyssessouza: My second option was to run the tests command line 2 times. One for each configuration of the envvar.
I'm not sure if I understood what you mean by using a global variable for this.

mat007: Why do you need an env var at all? Is this not the same process?

ulyssessouza: Ahhh... Finally got what you mean. The idea was to have it also configurable from outside of this `TestMain`.

With the envvar we can run just one specific test. And when running like this, `TestMain` doesn't run. So the envvar is a solution to be able to still pass this info to the test.

mat007: But can’t we use a command line flag instead?
Sorry it’s just I don’t like env vars much, they tend to magically configure (break) things, a bit like global vars on steroids 😀

ulyssessouza: The only alternative I see is to pass a build tag when running the test to choose in between tagged files. So we can choose which file to use, so we can use that conditionally depending on the tag.... But that really looks like over engineering and gets to the same problematic in the end...

pkg/e2e/framework.go
```diff
@@ -191,11 +194,32 @@ func (c *E2eCLI) RunCmd(args ...string) *icmd.Result {
...
 	res.Assert(c.test, icmd.Success)
 	return res
 }
 
+const composeStandaloneEnvVar = "COMPOSE_STANDALONE_MODE"
+
+// RunDockerComposeCmd runs a docker compose command, expects no error and returns a result
+func (c *E2eCLI) RunDockerComposeCmd(args ...string) *icmd.Result {
+	composeStandaloneMode := os.Getenv(composeStandaloneEnvVar)
+	if composeStandaloneMode == "true" || composeStandaloneMode == "1" {
```
mat007, line 210:
```suggestion
composeStandaloneMode := false

// RunDockerComposeCmd runs a docker compose command, expects no error and returns a result
func (c *E2eCLI) RunDockerComposeCmd(args ...string) *icmd.Result {
	if composeStandaloneMode {
```
then in the main:
```
func TestMain(m *testing.M) {
	logrus.Info("Running tests on COMPOSE_STANDALONE_MODE=true")
	composeStandaloneMode = true
	exitCode := m.Run()
	if exitCode != 0 {
		os.Exit(exitCode)
	}
	logrus.Info("Running tests on COMPOSE_STANDALONE_MODE=false")
	composeStandaloneMode = false
	os.Exit(m.Run())
}
```
We don’t really need an env var, do we?

ndeloof: would be nice this can be set as a `go test` argument so that we can parallelize run on CI

ulyssessouza: Not sure if parallelising this would be possible... 
The e2e tests open ports and maybe that would conflict with it's twin test

ulyssessouza: Also, I don't see a good way to run one specific test (by passing `E2E_TEST` environment variable) on one specific mode and not having to replicate Makefile targets or something like that...

ndeloof: this is what I had in mind: two targets, two parallel GHA workfows
never mind

mat007: What’s a Makefile? 😬 

It’s already possible to focus tests out of the box using `-run` with `go test`: https://pkg.go.dev/testing#hdr-Subtests_and_Sub_benchmarks
(I don’t have make on my machines)

ulyssessouza: @mat007 Yep. And that's what the Makefile does. The thing is, when targeting a specific test `TestMain` doesn't run. So that would execute just one of the modes (the default one).
Then the question is... Do we separate the execution of the 2 modes into different runs or leave it in the same (like it's implemented in this PR).
Now I'm leaning to having 2 runs in the GHA as @ndeloof said cuz that would be easier to run in local. Cuz I don't want to run the 2 things all the time I test locally. But for the CI that's important. And also that can be run in parallel...

ulyssessouza:
@mat007 PTAL

ndeloof (approved):

Makefile
```diff
@@ -42,8 +42,13 @@ compose-plugin: ## Compile the compose cli-plugin
...
 .PHONY: e2e-compose
-e2e-compose: ## Run End to end local tests. Set E2E_TEST=TestName to run a single test
-	gotestsum $(TEST_FLAGS) ./pkg/e2e -- -count=1
+e2e-compose: ## Run end to end local tests in plugin mode. Set E2E_TEST=TestName to run a single test
+	go test $(TEST_FLAGS) -count=1 ./pkg/e2e
+
+.PHONY: e2e-compose-standalone
+e2e-compose-standalone: ## Run End to end local tests in standalone mode. Set E2E_TEST=TestName to run a single test
+	go test $(TEST_FLAGS) -count=1 --tags=standalone ./pkg/e2e
+
```
mat007, line 52: Should we have a target to do both, e.g.
```
.PHONY: e2e
e2e: e2e-compose e2e-compose-standalone
```
?

ulyssessouza: I started like this... But I don't see who would use it...
In the CI it wouldn't be used neither. Cuz like this it cannot be parallelised.

ulyssessouza: Bu yes... I can add that

mat007 (approved):
Thanks!

PR #9012, opened 2021-12-07
Remove intermediate containers when build succeeded in classic build

notok:
**What I did**

Remove intermediate containers when build succeeded in classic build like default behavior of docker build.

**Related issue**

Intermediate containers remain even when build succeeded when building with classic way (i.e. not with buildkit).

The docker-compose cli [accepts the option `--no-rm`](https://github.com/docker/compose/blob/v2/cmd/compose/build.go#L92), but it's ignored because it's deprecated option.


ulyssessouza (approved):
LGTM

PR #9013, opened 2021-12-08
Docker compose bug for common parents

rogerhu:
**What I did**

- Put breakpoints on line 93 and 102: https://github.com/docker/compose/blob/8a9c4b52b4c56a2d448a39f544161169cb25c3ae/pkg/compose/dependencies.go#L93-L109
 
**Related issue**

Watch every so often the output vary between GOOD:

```
kochiku_redis
kochiku_mysql
kochiku_build
```

Versus BAD:

```
kochiku_redis
kochiku_mysql
kochiku_build
kochiku_build
```

### DIFF
--- a/pkg/compose/dependencies_test.go
+++ b/pkg/compose/dependencies_test.go
@@ -18,6 +18,7 @@ package compose
 
 import (
 	"context"
+	"fmt"
 	"testing"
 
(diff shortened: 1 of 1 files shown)


milas:
See #9878 for a fix. Thanks for the repro case!

Issue #8523, opened 2021-07-02
Docker-compose no longer gracefully stops


cchantep:
### Expected behavior

Successful graceful shutdown of `docker-compose`

### Actual behavior

All containers are stopped, but compose is hanging on forever until we are forced to kill it.

```
Gracefully stopping... (press Ctrl+C again to force)
[+] Running 7/7
 ⠿ Container PRJ-dev_spark-worker_1          Stopped                                                          11.3s
 ⠿ Container PRJ-dev_spark-master_1          Stopped                                                          12.2s
 ⠿ Container PRJ-dev_kafka-schemaregistry_1  Stopped                                                          11.3s
 ⠿ Container PRJ-dev_mongodb_1               Stopped                                                          12.3s
 ⠿ Container PRJ-dev_object-storage_1        Stopped                                                           1.7s
 ⠿ Container PRJ-dev_kafka_1                 Stopped                                                           4.3s
 ⠿ Container PRJ-dev_zookeeper_1             Stopped                                                          11.4s
```

(hanging forever ...)

```
^CERRO[3631] got 3 SIGTERM/SIGINTs, forcing shutdown      
[+] Running 0/0
 ⠿ Container PRJ-dev_mongodb_1               Error while Killing                                               0.0s
 ⠿ Container PRJ-dev_spark-worker_1          Error while Killing                                               0.0s
 ⠿ Container PRJ-dev_object-storage_1        Error while Killi...                                              0.0s
 ⠿ Container PRJ-dev_spark-master_1          Error while Killing                                               0.0s
 ⠿ Container PRJ-dev_kafka-schemaregistry_1  Error while...                                                    0.0s
 ⠿ Container PRJ-dev_kafka_1                 Error while Killing                                               0.0s
 ⠿ Container PRJ-dev_zookeeper_1             Error while Killing  
```

### Information

  - macOS Version: 10.15.7
  - Intel core i7
  - Docker Desktop version: 3.5.1 (3.5.1.7)
  - Docker Compose version: 2.0.0-beta.4

### Steps to reproduce the behavior

- On a fresh Docker Desktop (after data purge), `docker-compose -f any-conf.yml up`
- Once everything is up, `Ctrl+C` to gracefully stop

Radiergummi:
This also occurs on my machine if `restart: unless-stopped` is used; same setup as OP, latest Docker Desktop on Mac. This is pretty annoying.
[...]

PR #9018, opened 2021-12-08
Stop compose on CTRL+C and containers are stopped

ulyssessouza:
**What I did**
Stop compose on CTRL+C and containers are stopped

Even containers with "restart: always" should be stopped and leave docker-compose finish it's execution.

**Related issue**
Resolves: https://github.com/docker/compose/issues/8523

### DIFF
--- a/pkg/compose/start.go
+++ b/pkg/compose/start.go
@@ -115,7 +115,7 @@ func (s *composeService) watchContainers(ctx context.Context, projectName string
 				restarted := watched[container.ID]
 				watched[container.ID] = restarted + 1
 				// Container terminated.
-				willRestart := willContainerRestart(inspected, restarted)
+				willRestart := !s.stopping && willContainerRestart(inspected, restarted)
 
 				listener(api.ContainerEvent{
 					Type:       api.ContainerEventExit,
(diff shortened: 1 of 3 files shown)


pkg/compose/start.go
```diff
@@ -115,7 +115,7 @@ func (s *composeService) watchContainers(ctx context.Context, projectName string
 				restarted := watched[container.ID]
 				watched[container.ID] = restarted + 1
 				// Container terminated.
-				willRestart := willContainerRestart(inspected, restarted)
+				willRestart := !s.stopping && willContainerRestart(inspected, restarted)
```
ndeloof, line 118: This will for sure stop compose for watching, but container will still run!

ndeloof: Oh no, my bad. `willContainerRestart` tells us container is _configured_ to restart, but doesn't guarantee it will.
Anyway we can't just stop this way, as this just assumes container will stop as we asked him to, but doesn't check.

ndeloof:
I have an alternative proposal based on `stop` event: https://github.com/docker/compose/pull/9019

ulyssessouza:
Closing in favor of #9019

Issue #9003, opened 2021-12-06
(Regression) Pressing CTRL+C on `docker-compose up` hangs when using containers with restart: unless-stopped|always


joanbm:
**Description**

Since version 2.2.0, when I launch a `docker-compose.yml` file defining one or more containers with its restart policy set to `restart: unless-stopped` or `restart: always` with `docker-compose up`, then pressing CTRL+C after the container is up will stop the container as expected, but the docker-compose process will no longer successfully quit. Instead, it will hang, and the only way to get it to quit is to terminate it with something like `pkill -6 docker-compose`.

I have bisected the problem and it appears to have been introduced in commit ed2395819d81ac812a5eefeeccbf77432392626c. Testing a version previous to this commit or reverting it, avoids the problem when terminating with CTRL+C.

**Steps to reproduce the issue:**
1. Install Docker Compose >= 2.2.0  (tested up to latest v2 branch commit)
2. Create a new folder with a `docker-compose.yml` file with the following content:

```yml
version: "3.4"

services:
  db:
    image: alpine:3.15
    command: ["/bin/sleep", "9999"]
    restart: unless-stopped
```

3. Run `docker-compose up` on the folder (where `docker-compose` is the compiled `compose` binary, do *not* run `docker compose up`, as this seems to work around the problem).

4. Wait for the container to start up and then press CTRL+C and wait some seconds.

**Describe the results you received:**

The container stops, but docker-compose does never quit no matter how many times I press CTRL+C. I need to terminate the docker-compose process in order to quit it.

```console
$ docker-compose up
[+] Running 2/2
 ⠿ Network dc_default  Created                                                                                                                                                                                  0.0s
 ⠿ Container dc-db-1   Created                                                                                                                                                                                  0.1s
Attaching to dc-db-1
^CGracefully stopping... (press Ctrl+C again to force)
[+] Running 1/1
 ⠿ Container dc-db-1  Stopped                                                                                                                                                                                  10.4s
[+] Running 0/0
 ⠿ Container dc-db-1  Error while Killing                                                                                                                                                                       0.0s
^C^C^C^C^C^C^C^C^C
```

**Describe the results you expected:**

The container stops, then docker-compose quits gracefully.

```console
$ docker-compose up
[+] Running 2/2
 ⠿ Network dc_default  Created                                                                                                                                                                                  0.0s
 ⠿ Container dc-db-1   Created                                                                                                                                                                                  0.1s
Attaching to dc-db-1
[...]

PR #9019, opened 2021-12-08
handle "stop" event

ndeloof:
**What I did**
This is an alternative to https://github.com/docker/compose/pull/9018
catching the `stop` event, we can stop watching a container and consider it "done"

this include moving `convert` to a non-main package as this prevent debugger to start in Goland

**Related issue**
fixes: #8523

image

(he won't run anymore)

### DIFF
--- a/pkg/compose/start.go
+++ b/pkg/compose/start.go
@@ -111,6 +111,21 @@ func (s *composeService) watchContainers(ctx context.Context, projectName string
 			}
 			name := getContainerNameWithoutProject(container)
 
+			if event.Status == "stop" {
+				listener(api.ContainerEvent{
+					Type:      api.ContainerEventStopped,
+					Container: name,
+					Service:   container.Labels[api.ServiceLabel],
+				})
+
+				delete(watched, container.ID)
+				if len(watched) == 0 {
+					// all project containers stopped, we're done
+					stop()
+				}
+				return nil
+			}
+
 			if event.Status == "die" {
 				restarted := watched[container.ID]
 				watched[container.ID] = restarted + 1
--- a/cmd/compatibility/convert.go
+++ b/cmd/compatibility/convert.go
@@ -14,7 +14,7 @@
    limitations under the License.
 */
 
-package main
+package compatibility
 
 import (
 	"fmt"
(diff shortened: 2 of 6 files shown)


pkg/compose/start.go
```diff
@@ -111,7 +112,23 @@ func (s *composeService) watchContainers(ctx context.Context, projectName string
 			}
 			name := getContainerNameWithoutProject(container)
 
+			if event.Status == "stop" {
+				listener(api.ContainerEvent{
+					Type:       api.ContainerEventExit,
```
ulyssessouza, line 117: In the end it should make the same, but shouldn't it be stopped?
```suggestion
					Type:       api.ContainerEventStopped,
```

ndeloof: oh yes, forgot to update this

ulyssessouza:
@ndeloof gofmt

pkg/compose/start.go
```diff
@@ -111,7 +112,23 @@ func (s *composeService) watchContainers(ctx context.Context, projectName string
...
+				delete(watched, container.ID)
+				if len(watched) == 0 {
+					// all project containers stopped, we're done
+					stop()
+				}
+				return nil
+			}
+
 			if event.Status == "die" {
+				fmt.Println("container died")
```
ulyssessouza, line 131: Is that line intentional?
This would spam a line per container.

ulyssessouza (changes requested):
Just the spammer line

pkg/compose/start.go
ndeloof: oups, debug

ulyssessouza (approved):
LGTM. But I just noticed we have no tests for this function :(

1ace:
@ulyssessouza:
> But I just noticed we have no tests for this function :(

FYI, I posted a non-interactive (no manual step) reproducer for this bug here: https://github.com/docker/compose/issues/9003#issuecomment-987770048

I'm not familiar with your CI setup but I expect it should be easy enough to add that `docker-compose.yml` file somewhere and run the command I provided from your CI, avoiding this bug from being introduced again at some point in the future :slightly_smiling_face:

Issue #8864, opened 2021-11-02
v2 `docker compose kill` errors when the container is not running, where v1 was not


majkelcc:
**Description**

When I send a kill signal to a container that is stopped using `docker compose kill ...`, the command exits with an error. This was not the case with docker-compose v1.

**Steps to reproduce the issue:**
1. With the following example `docker-compose.yml` file:
```
---
services:
  hello:
    image: hello-world
```
2. Run `docker compose up -d hello`
3. Run `docker compose stop hello`
4. Run `docker compose kill hello`

**Describe the results you received:**

Sending a kill signal results in the following error with an exit status 1:

```
[+] Running 0/0
 ⠿ Container hello-1  Error while Killing                                                                                            0.0s
Error response from daemon: Cannot kill container: c966210d3f9f5267d8b60a4b6c77ed92e2bc4ce49cbb46bbe7db022f5a2baba7: Container c966210d3f9f5267d8b60a4b6c77ed92e2bc4ce49cbb46bbe7db022f5a2baba7 is not running
```

**Describe the results you expected:**

This is a breaking change, which I haven't seen documented anywhere. I would expect the same behaviour as with docker-compose v1 - no error message and exit code 0. Alternatively, a way to achieve the same result (with a `--force` flag for example).

**Output of `docker compose version`:**

```
Docker Compose version v2.0.0 (Docker Desktop on Mac)
Docker Compose version v2.0.1 (Linux)
```

**Output of `docker info`:**

```
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Build with BuildKit (Docker Inc., v0.6.3)
  compose: Docker Compose (Docker Inc., v2.0.0)
  scan: Docker Scan (Docker Inc., v0.8.0)

Server:
 Containers: 18
  Running: 3
  Paused: 0
  Stopped: 15
 Images: 16
 Server Version: 20.10.8
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 1
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: runc io.containerd.runc.v2 io.containerd.runtime.v1.linux
 Default Runtime: runc
[...]

PR #9020, opened 2021-12-09
Only kill running containers

ulyssessouza:
**What I did**
Only kill running containers

**Related issue**
Resolves https://github.com/docker/compose/issues/8864

### DIFF
--- a/cmd/compatibility/convert.go
+++ b/cmd/compatibility/convert.go
@@ -43,6 +43,7 @@ func getStringFlags() []string {
 	}
 }
 
+// Convert transforms standalone docker-compose args into CLI plugin compliant ones
 func Convert(args []string) []string {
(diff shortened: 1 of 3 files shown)


cmd/compatibility/convert.go
```diff
@@ -43,6 +43,7 @@ func getStringFlags() []string {
 	}
 }
 
+// Convert transforms args from a V1 to V2
```
ulyssessouza, line 46: Linter

ndeloof: "Convert transforms standalone docker-compose args into CLI plugin compliant ones"

ndeloof (approved):

jorgecolonconsulting:
When will this make it into a release version?

PR #9033, opened 2021-12-14
Add ddev's e2e test

ulyssessouza:
**What I did**
Add ddev's e2e test

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -51,7 +51,7 @@ e2e-compose-standalone: ## Run End to end local tests in standalone mode. Set E2
 	rm -f /usr/local/bin/docker-compose
 	cp bin/docker-compose /usr/local/bin
 	docker-compose version
-	go test $(TEST_FLAGS) -count=1 --tags=standalone ./pkg/e2e
+	go test $(TEST_FLAGS) -v -count=1 -parallel=1 --tags=standalone ./pkg/e2e
 
 .PHONY: mocks
 mocks:
--- a/Makefile
+++ b/Makefile
@@ -43,10 +43,14 @@ compose-plugin: ## Compile the compose cli-plugin
 
 .PHONY: e2e-compose
 e2e-compose: ## Run end to end local tests in plugin mode. Set E2E_TEST=TestName to run a single test
+	docker compose version
 	go test $(TEST_FLAGS) -count=1 ./pkg/e2e
 
 .PHONY: e2e-compose-standalone
 e2e-compose-standalone: ## Run End to end local tests in standalone mode. Set E2E_TEST=TestName to run a single test
+	rm -f /usr/local/bin/docker-compose
+	cp bin/docker-compose /usr/local/bin
+	docker-compose version
 	go test $(TEST_FLAGS) -count=1 --tags=standalone ./pkg/e2e
 
 .PHONY: mocks
--- a/pkg/e2e/ddev_test.go
+++ b/pkg/e2e/ddev_test.go
@@ -27,7 +27,7 @@ import (
 	"gotest.tools/v3/assert"
 )
 
-const ddevVersion = "v1.18.2"
+const ddevVersion = "v1.19.1"
 
 func TestComposeRunDdev(t *testing.T) {
 	if !composeStandaloneMode {
@@ -36,9 +36,14 @@ func TestComposeRunDdev(t *testing.T) {
 	if runtime.GOOS == "windows" {
 		t.Skip("Running on Windows. Skipping...")
 	}
+	_ = os.Setenv("DDEV_DEBUG", "true")
+
 	c := NewParallelE2eCLI(t, binDir)
 	dir, err := os.MkdirTemp("", t.Name()+"-")
 	assert.NilError(t, err)
+
+	// ddev needs to be able to find mkcert to figure out where certs are.
+	_ = os.Setenv("PATH", fmt.Sprintf("%s:%s", os.Getenv("PATH"), dir))
 
 	siteName := filepath.Base(dir)
 
@@ -60,31 +65,22 @@ func TestComposeRunDdev(t *testing.T) {
 			compressedFilename))
 
 	c.RunCmdInDir(dir, "tar", "-xzf", compressedFilename)
-	c.RunDockerCmd("pull", "drud/ddev-ssh-agent:v1.18.0")
-	c.RunDockerCmd("pull", "busybox:stable")
-	c.RunDockerCmd("pull", "phpmyadmin:5")
-
-	c.RunDockerCmd("pull", tagged("drud/ddev-router"))
-	c.RunDockerCmd("pull", tagged("drud/ddev-dbserver-mariadb-10.3"))
-	c.RunDockerCmd("pull", tagged("drud/ddev-webserver"))
 
 	// Create a simple index.php we can test against.
 	c.RunCmdInDir(dir, "sh", "-c", "echo '<?php\nprint \"ddev is working\";' >index.php")
 
 	c.RunCmdInDir(dir, "./ddev", "config", "--auto")
 	c.RunCmdInDir(dir, "./ddev", "config", "global", "--use-docker-compose-from-path")
+	vRes := c.RunCmdInDir(dir, "./ddev", "version")
+	out := vRes.Stdout()
+	fmt.Printf("ddev version: %s\n", out)
 
 	c.RunCmdInDir(dir, "./ddev", "poweroff")
 
-	startRes := c.RunCmdInDir(dir, "./ddev", "start", "-y")
-	assert.Equal(c.test, startRes.ExitCode, 0, "Could not start project")
+	c.RunCmdInDir(dir, "./ddev", "start", "-y")
 
 	curlRes := c.RunCmdInDir(dir, "curl", "-sSL", fmt.Sprintf("http://%s.ddev.site", siteName))
-	out := curlRes.Stdout()
+	out = curlRes.Stdout()
 	fmt.Println(out)
 	assert.Assert(c.test, strings.Contains(out, "ddev is working"), "Could not start project")
-}
-
-func tagged(img string) string {
-	return fmt.Sprintf("%s:%s", img, ddevVersion)
 }
--- a/pkg/e2e/ddev_test.go
+++ b/pkg/e2e/ddev_test.go
@@ -1,0 +1,90 @@
+/*
+   Copyright 2020 Docker Compose CLI authors
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package e2e
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"runtime"
+	"strings"
+	"testing"
+
+	"gotest.tools/v3/assert"
+)
+
+const ddevVersion = "v1.18.2"
+
+func TestComposeRunDdev(t *testing.T) {
+	if !composeStandaloneMode {
+		t.Skip("Not running on standalone mode.")
+	}
+	if runtime.GOOS == "windows" {
+		t.Skip("Running on Windows. Skipping...")
+	}
+	c := NewParallelE2eCLI(t, binDir)
+	dir, err := os.MkdirTemp("", t.Name()+"-")
+	assert.NilError(t, err)
+
+	siteName := filepath.Base(dir)
+
+	t.Cleanup(func() {
+		_ = c.RunCmdInDir(dir, "./ddev", "delete", "-Oy")
+		_ = c.RunCmdInDir(dir, "./ddev", "poweroff")
+		_ = os.RemoveAll(dir)
+	})
+
+	osName := "linux"
+	if runtime.GOOS == "darwin" {
+		osName = "macos"
+	}
+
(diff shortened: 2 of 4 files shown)


pkg/e2e/ddev_test.go
```diff
@@ -0,0 +1,89 @@
...
+	c := NewParallelE2eCLI(t, binDir)
+	dir, err := os.MkdirTemp("", t.Name()+"-")
+	if err != nil {
+		return
+	}
+	t.Cleanup(func() {
+		_ = os.RemoveAll(dir)
+	})
+
+	fmt.Println(runtime.GOARCH)
```
mat007, line 45: Is this a left over debug statement?

pkg/e2e/ddev_test.go
```diff
@@ -0,0 +1,89 @@
...
+
+func TestComposeRunDdev(t *testing.T) {
+	if runtime.GOOS == "windows" {
+		t.Skip("Running on Windows. Skipping...")
+	}
+	c := NewParallelE2eCLI(t, binDir)
+	dir, err := os.MkdirTemp("", t.Name()+"-")
+	if err != nil {
+		return
+	}
```
mat007, line 40: Should this not be
```suggestion
	assert.NotNil(err)
```
(or something like that, without completion I’m never sure which it is with the different testing libraries we use 😬)

pkg/e2e/ddev_test.go
```diff
@@ -0,0 +1,89 @@
...
+	t.Cleanup(func() {
+		_ = os.RemoveAll(dir)
+	})
```
mat007, line 43: I understand the `t.Cleanup` benefits when delegating to helper functions and such, be here is this really better than
```suggestion
	defer os.RemoveAll(dir)
```
?

pkg/e2e/ddev_test.go
```diff
@@ -0,0 +1,89 @@
...
+
+	compressedFilename := fmt.Sprintf("ddev_%s-%s.%s.tar.gz", osName, runtime.GOARCH, ddevVersion)
+
+	c.RunCmdInDir(dir, "curl", "-LO",
+		fmt.Sprintf("https://github.com/drud/ddev/releases/download/%s/%s",
+			ddevVersion,
+			compressedFilename))
+
+	c.RunCmdInDir(dir, "tar", "-xzf", compressedFilename)
+	c.RunDockerCmd("pull", "drud/ddev-ssh-agent:v1.18.0")
```
mat007, line 60: Is the version here not tied to the `ddevVersion` above?

pkg/e2e/ddev_test.go
ulyssessouza: Ops... debugging...

pkg/e2e/ddev_test.go
ulyssessouza: Sure! I think that was my automatic error checking kicking in!
Good catch!

pkg/e2e/ddev_test.go
ulyssessouza: Here I don't know. I think for this one it doesn't matter cuz there is just one run (the main).
I actually just used what the rest of the tests are using. But I will change this one for simplicity.

pkg/e2e/ddev_test.go
ulyssessouza: This one is strange. What I did was running `ddev start` directly without the pulls and see what it pulls automatically. And the version was that one. So yes... This one is intentional. I also thought everything was tied to the ddev version, but it seems that this one is released differently. Also, the version `v1.18.2` of this image doesn't exist

mat007 (approved):
Looks good!
@rfay what do you think?

pkg/e2e/ddev_test.go
```diff
@@ -0,0 +1,83 @@
...
+	defer os.RemoveAll(dir) // nolint:errcheck
+
+	osName := "linux"
+	if runtime.GOOS == "darwin" {
+		osName = "macos"
+	}
+
+	compressedFilename := fmt.Sprintf("ddev_%s-%s.%s.tar.gz", osName, runtime.GOARCH, ddevVersion)
+
+	c.RunCmdInDir(dir, "curl", "-LO",
```
rfay, line 48: It would probably be easier to use [install_ddev.sh](https://github.com/drud/ddev/blob/master/scripts/install_ddev.sh)

pkg/e2e/ddev_test.go
```diff
@@ -0,0 +1,83 @@
...
+			compressedFilename))
+
+	c.RunCmdInDir(dir, "tar", "-xzf", compressedFilename)
+	c.RunDockerCmd("pull", "drud/ddev-ssh-agent:v1.18.0")
+	c.RunDockerCmd("pull", "busybox:stable")
+	c.RunDockerCmd("pull", "phpmyadmin:5")
+
+	c.RunDockerCmd("pull", tagged("drud/ddev-router"))
+	c.RunDockerCmd("pull", tagged("drud/ddev-dbserver-mariadb-10.3"))
+	c.RunDockerCmd("pull", tagged("drud/ddev-webserver"))
```
rfay, line 60: All these pulls happen on `ddev start`, and you can't pull without a tag. `ddev start` will pull with the correct tag. There is no `latest` tag on any of them.

rfay (changes requested):
Thanks for this!

As is, this won't use your docker-compose I don't think. In ddev v1.18.2, ddev downloads a private docker-compose. So you'll either need to use `ddev config global --required-docker-compose-version=<currrent>` or `ddev config global --use-docker-compose-from-path` (those can also just be set in ~/.ddev/global_config.yaml). 

An actual end-to-end test will at least curl from an index.php. You could just make an index.php with 
```php
<?php
echo "hi there";
```
and then `curl http://<project>.ddev.php` and verify that you get that (and a 200 of course)

It looks like all this does is run a `ddev start`, which isn't "end to end". 

I recommend at least having a project with an index.php that has

pkg/e2e/ddev_test.go
ulyssessouza: `tagged` function will append that to the image name.

rfay: It's OK then, doesn't do any harm.

rfay:
I'll try running this in a bit and propose some code.

rfay:
Added a PR to your PR in https://github.com/ulyssessouza/compose/pull/23 that makes the test a little stronger, tests to see if it's actually working. cleans up project after it's done, etc.

This assumes that the docker-compose under test is the one that will be found in the $PATH.

rfay (approved):
Thanks, this is good for me **as long as the docker-compose under test first in the $PATH** - please just confirm that. If it's not, then this isn't relevant yet.

rfay:
I see `GOOS=linux GOARCH=amd64 CGO_ENABLED=0 go build -trimpath -ldflags="-s -w -X github.com/docker/compose/v2/internal.Version=" -tags e2e -o bin/docker-compose ./cmd` so it puts it into the ./bin directory from the build. Maybe just adding a build stanza in the .github workflow doing a `docker-compose --version` would help the confidence level here.

rfay:
It looks to me like all the tests running in parallel fight over each other's containers or networks. I'll do a PR to this PR that adds `-parallel=1` to the e2e-standalone test. It will also make the output a pile easier to read when something fails.

`make e2e-compose-standalone TEST_FLAGS="-run='^TestCompose.*Ddev' "` certainly works fine for me every time.

rfay:
Opened https://github.com/ulyssessouza/compose/pull/27  which has an explanation of fix.

glours (changes requested):
We should be able to run the e2e test locally, @ulyssessouza can move the copy of the binary to the CI script please?

Otherwise I didn't succeed to run the test on my Mac M1, I got the following issue

```console
 Network ddev-testcomposerunddev-2047219997_default  Created
        Error response from daemon: create .: volume name is too short, names should be at least two alphanumeric characters
        
        Stderr:   ssh-agent container is running: If you want to add authentication to the ssh-agent container, run 'ddev auth ssh' to enable your keys.
        mkcert may not be properly installed, we suggest installing it for trusted https support, `brew install mkcert nss`, `choco install -y mkcert`, etc. and then `mkcert -install`
        IsInternetActive DEBUG: err=<nil> ctx.Err()=<nil> addrs=[127.0.0.1] IsInternetactive==true, randomURL=yhygkjiacf.ddev.site internet_detection_timeout_ms=750ms
        
        Failed to start TestComposeRunDdev-2047219997: ComposeCmd failed to run 'COMPOSE_PROJECT_NAME=ddev-TestComposeRunDdev-2047219997 docker-compose -f /private/var/folders/rl/282crpp149x6xf_400p17fj40000gn/T/TestComposeRunDdev-2047219997/.ddev/.ddev-docker-compose-full.yaml up --build -d', action='[up --build -d]', err='exit status 1', stdout='#1 [drud/ddev-webserver:v1.19.0-1-TestComposeRunDdev-2047219997-built internal] load build definition from Dockerfile
        #1 transferring dockerfile: 580B done
        #1 DONE 0.0s
        
        #2 [drud/ddev-dbserver-mariadb-10.3:v1.19.0-TestComposeRunDdev-2047219997-built internal] load build definition from Dockerfile
        #2 transferring dockerfile: 408B done
        #2 DONE 0.0s
        
        #3 [drud/ddev-dbserver-mariadb-10.3:v1.19.0-TestComposeRunDdev-2047219997-built internal] load .dockerignore
        #3 transferring context: 2B done
        #3 DONE 0.0s
        
        #4 [drud/ddev-webserver:v1.19.0-1-TestComposeRunDdev-2047219997-built internal] load .dockerignore
        #4 transferring context: 2B done
        #4 DONE 0.0s
        
        #5 [drud/ddev-webserver:v1.19.0-1-TestComposeRunDdev-2047219997-built internal] load metadata for docker.io/drud/ddev-webserver:v1.19.0-1
        #5 DONE 0.0s
        
        #6 [drud/ddev-dbserver-mariadb-10.3:v1.19.0-TestComposeRunDdev-2047219997-built internal] load metadata for docker.io/drud/ddev-dbserver-mariadb-10.3:v1.19.0
        #6 DONE 0.0s
        
        #7 [drud/ddev-dbserver-mariadb-10.3:v1.19.0-TestComposeRunDdev-2047219997-built 1/2] FROM docker.io/drud/ddev-dbserver-mariadb-10.3:v1.19.0
        #7 DONE 0.0s
        
        #8 [drud/ddev-dbserver-mariadb-10.3:v1.19.0-TestComposeRunDdev-2047219997-built 2/2] RUN (groupadd --gid 20 "glours" || groupadd "glours" || true) && (useradd  -l -m -s "/bin/bash" --gid "glours" --comment '' --uid 501 "glours" || useradd  -l -m -s "/bin/bash" --gid "glours" --comment '' "glours" || useradd  -l -m -s "/bin/bash" --gid "20" --comment '' "glours")
        #8 CACHED
        
        #9 [drud/ddev-webserver:v1.19.0-1-TestComposeRunDdev-2047219997-built 1/4] FROM docker.io/drud/ddev-webserver:v1.19.0-1
        #9 DONE 0.0s
        
        #10 [drud/ddev-webserver:v1.19.0-1-TestComposeRunDdev-2047219997-built 2/4] RUN (groupadd --gid 20 "glours" || groupadd "glours" || true) && (useradd  -l -m -s "/bin/bash" --gid "glours" --comment '' --uid 501 "glours" || useradd  -l -m -s "/bin/bash" --gid "glours" --comment '' "glours" || useradd  -l -m -s "/bin/bash" --gid "20" --comment '' "glours")
        #10 CACHED
        
        #11 [drud/ddev-webserver:v1.19.0-1-TestComposeRunDdev-2047219997-built 3/4] RUN export XDEBUG_MODE=off && ( composer self-update --2 || composer self-update --2 || true )
        #11 CACHED
        
        #12 [drud/ddev-webserver:v1.19.0-1-TestComposeRunDdev-2047219997-built 4/4] RUN chmod 600 ~glours/.pgpass ~glours/.my.cnf
        #12 CACHED
        
        #13 [drud/ddev-dbserver-mariadb-10.3:v1.19.0-TestComposeRunDdev-2047219997-built] exporting to image
        #13 exporting layers done
        #13 writing image sha256:8479fcda75dc943532607f8255a0dae40982625130b2762a02eb285838d55959 done
        #13 naming to docker.io/drud/ddev-webserver:v1.19.0-1-TestComposeRunDdev-2047219997-built done
        #13 writing image sha256:72e0459aecebee2d72392a30f55003b65582d7e8381d25cf8365683db6594e96 done
        #13 naming to docker.io/drud/ddev-dbserver-mariadb-10.3:v1.19.0-TestComposeRunDdev-2047219997-built done
        #13 DONE 0.0s
        ', stderr='Network ddev-testcomposerunddev-2047219997_default  Creating
        Network ddev-testcomposerunddev-2047219997_default  Created
        Container ddev-TestComposeRunDdev-2047219997-db  Creating
        Container ddev-TestComposeRunDdev-2047219997-dba  Creating
        Container ddev-TestComposeRunDdev-2047219997-web  Creating
        Container ddev-TestComposeRunDdev-2047219997-dba  Created
        Error response from daemon: create .: volume name is too short, names should be at least two alphanumeric characters'
        
        
        Failures:
        ExitCode was 1 expected 0


```

rfay:
It looks to me like the failure you're seeing here is https://github.com/docker/compose/issues/9428 - DDEV hasn't been able to run against any version of compose since 2.3.3 due to a cascade of regressions.

ndeloof (approved):

Makefile
```diff
@@ -43,11 +43,15 @@ compose-plugin: ## Compile the compose cli-plugin
...
 .PHONY: e2e-compose
 e2e-compose: ## Run end to end local tests in plugin mode. Set E2E_TEST=TestName to run a single test
+	docker compose version
 	go test $(TEST_FLAGS) -count=1 ./pkg/e2e
 
 .PHONY: e2e-compose-standalone
 e2e-compose-standalone: ## Run End to end local tests in standalone mode. Set E2E_TEST=TestName to run a single test
-	go test $(TEST_FLAGS) -count=1 --tags=standalone ./pkg/e2e
+	rm -f /usr/local/bin/docker-compose
+	cp bin/docker-compose /usr/local/bin
```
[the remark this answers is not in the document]
glours: @ulyssessouza you didn't move the copy of the compose binary to the CI workflow

Issue #8908, opened 2021-11-09
`docker compose run` hangs indefinitely if stdout/err were duplicated to a file


chadwhitacre:
### Description

We have an [`onpremise`](https://github.com/getsentry/onpremise) repo where we provide a Dockerized version of our application for small self-hosted installs. Our [install script](https://github.com/getsentry/onpremise/blob/c16ae024dc3e8259b03679f950e80e1cbfb5dae8/install.sh) for this repo is fairly involved, and one of the things we do is [capture stdout/stderr in a file](https://github.com/getsentry/onpremise/blob/c16ae024dc3e8259b03679f950e80e1cbfb5dae8/install/_lib.sh#L4-L6) to aid in debugging installation issues:

```
exec &> >(tee -a "$log_file")
```

Straightforward enough, and it works fine with Docker Compose v1.

With Docker Compose v2, we're seeing that— following the logging capture—a `docker compose run` [invocation that prompts for input](https://github.com/getsentry/onpremise/blob/c16ae024dc3e8259b03679f950e80e1cbfb5dae8/install/set-up-and-migrate-database.sh#L12) will hang indefinitely, and can't be killed with Ctrl-C either. I've whittled down [a test case](https://github.com/chadwhitacre/docker-compose-v2-tty-bug/).

### Steps to reproduce the issue:
1. Clone https://github.com/chadwhitacre/docker-compose-v2-tty-bug/.
2. Ensure `which docker-compose` resolves to Docker Compose v1.
3. Ensure `which docker` resolves to a Docker with Compose v2.
2. [`./test.sh`](https://github.com/chadwhitacre/docker-compose-v2-tty-bug/blob/main/test.sh)

### Describe the results you received:

The test script accepts three prompts, then hangs.

```
$ ./test.sh
[+] Running 2/2
 ⠿ Container tty-test_testing_run_397b9b93f30e  Removed                                                   0.0s
 ⠿ Network tty-test_default                     Removed                                                   0.1s
[+] Building 0.1s (8/8) FINISHED                                                                               
 => [internal] load build definition from Dockerfile                                                      0.0s
 => => transferring dockerfile: 31B                                                                       0.0s
 => [internal] load .dockerignore                                                                         0.0s
 => => transferring context: 2B                                                                           0.0s
 => [internal] load metadata for docker.io/library/ubuntu:latest                                          0.0s
 => [1/3] FROM docker.io/library/ubuntu:latest                                                            0.0s
[...]

PR #9035, opened 2021-12-14
Do not try to guess when to allocate a TTY and keep it as default

ulyssessouza:
**What I did**
Do not try to guess when to allocate a TTY and keep it as default

**Related issue**
Resolves https://github.com/docker/compose/issues/8908

### DIFF
--- a/cmd/compose/run.go
+++ b/cmd/compose/run.go
@@ -25,7 +25,6 @@ import (
 	cgo "github.com/compose-spec/compose-go/cli"
 	"github.com/compose-spec/compose-go/loader"
 	"github.com/compose-spec/compose-go/types"
-	"github.com/mattn/go-isatty"
 	"github.com/mattn/go-shellwords"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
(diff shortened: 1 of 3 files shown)


ndeloof:
I can't find a better fix for this, but still sad we have to give up with tty detection, and wonder this will trigger more issues from users who expect compose to automatically adjust to environment :-/

cmd/compose/run.go
```diff
@@ -144,7 +144,7 @@ func runCommand(p *projectOptions, backend api.Service) *cobra.Command {
 	flags.StringArrayVarP(&opts.environment, "env", "e", []string{}, "Set environment variables")
 	flags.StringArrayVarP(&opts.labels, "label", "l", []string{}, "Add or override a label")
 	flags.BoolVar(&opts.Remove, "rm", false, "Automatically remove the container when it exits")
-	flags.BoolVarP(&opts.noTty, "no-TTY", "T", notAtTTY(), "Disable pseudo-noTty allocation. By default docker compose run allocates a TTY")
+	flags.BoolVarP(&opts.noTty, "no-TTY", "T", false, "Disable pseudo-noTty allocation. By default docker compose run allocates a TTY")
```
thaJeztah, line 147: Same logic is also used in `exec` - should it be removed there as well?

ulyssessouza: I think so... That would be strange to have a different behaviour.

ulyssessouza:
@ndeloof Yep. I also tried to save the automatic behaviour, but without success...
But at least the possible issues will have just a "-T" as "fix"

thaJeztah (approved):
SGTM

just to double-check; this brings it to the same behaviour / defaults as compose v1, correct?

ndeloof (approved):

rfay:
Since panic is a result of this I don't think it's probably satisfactory, see https://stackoverflow.com/questions/70855915/fix-panic-provided-file-is-not-a-console-from-docker-compose-in-github-action

ndeloof:
I don't like this either. I expect we get rid of this as we share more code with docker/cli, especially relying on the exact same `RunExec` function

rfay:
Right... but not preventing a panic? that doesn't seem like the right thing...

ndeloof:
We can burn some cycles trying to find a better way to manage this and avoid a `panic`, which is obviously the worst possible user experience to offer :'( 
but with https://github.com/docker/compose/pull/9198 and https://github.com/docker/cli/pull/3436 we are close to just drop all the custom `exec` code implemented by docker/compose to adopt the battle-tested `RunExec` code used by the docker CLI for years.

rfay:
Sounds good.  I had actually planned long ago and half-implemented a switch to using docker's exec for `ddev exec` instead of using `docker-compose exec` but didn't actually pull the trigger, just worried about unexpected results.

rostislav-simonik-plc:
Hello, i just tested with version 2.6.1 of docker compose and still having the issue. Once i remove exec with tee it starts working.

Issue #8895, opened 2021-11-07
docker compose (v2) down -v removes volumes not mentioned in the docker-compose file


ylilarry:
**docker compose down -v removes all volumes under the same PROJ_NAME**

I have multiple services running under the same `PROJ_NAME`, but they are launched by different docker-compose files. The effect of having them running under the same `PROJ_NAME` is that when I run `docker compose ps` on any docker-compose file, I see
        

        NAME                       COMMAND                  SERVICE             STATUS              PORTS
        kaleido-api-postgres       "docker-entrypoint.s…"   postgres            created             
        kaleido-api-sdk            "python3 /opt/bin/sd…"   sdk                 running             
        kaleido-api-server         "/bin/sh -c 'python3…"   server              created             
        kaleido-api_doc-server_1   "sh /entrypoint.bash"    doc-server          running             

Note that even though each service is defined separately in different compose file, they are all listed (not sure if this is the expected behavior? but that's not the bug focused by this report.)

When I run `docker compose up`, only the services defined in the current compose file are launched, this is the correct behavior.

When I run `docker compose down`, only the services defined in the current compose file are removed, this is the correct behavior.

When I run `docker compose down -v`, however, all volumes from the 4 services are removed. This is the wrong behavior. The correct behavior is only the volumes defined in the current compose file is removed.
        
Example, running `docker compose down -v` from the docker-compose file that lauched `doc-server`, I get:

        [+] Running 0/0
         ⠋ Volume kaleido-api-vscode-server  Removing                                                                                                                     0.0s
         ⠿ Network kaleido-api_default       Error                                                                                                                        0.0s
        Error response from daemon: remove kaleido-api-vscode-server: volume is in use - [c94633c92eda18ee851f559df5641752d56077262265d15fb528c43b85d0c705]

`Volume kaleido-api-vscode-server ` is defined in the docker-compose file that lauched sdk. It is not mentioned in the same docker-compose file. Therefore, it shouldn't be removed.

**Additional information you deem important (e.g. issue happens only occasionally):**

PROJ_NAME is set globally as `kaleido-api`. All services are under the same `PROJ_NAME`, which is probably why `docker compose` is confused.

**Output of `docker compose version`:**

```
Docker Compose version v2.1.0
```

**Output of `docker info`:**

```
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  app: Docker App (Docker Inc., v0.9.1-beta3)
[...]

PR #9074, opened 2022-01-03
only remove volumes set by compose file

ndeloof:
**What I did**
Only remove volumes set by compose file
As no compose file is set (down -p name) use volume labels to rebuild the project model

**Related issue**
close https://github.com/docker/compose/issues/8895


mr-karan:
+1 can this be prioritized? This looks an important critical bug that must be addressed ASAP.

ulyssessouza (approved):

PR #9092, opened 2022-01-07
Don't wait forever for unhealthy dependencies

benesch:
@ndeloof I've got one more `--wait`-related fix for you here. Does the test case look ok?

**What I did**

The previous code would wait for dependencies to become healthy forever,
even if they'd become unhealthy in the meantime. I can't find an issue
report for this bug, but it was described in a comment on the PR that
introduced the `--wait` flag [0].

[0]: https://github.com/docker/compose/pull/8777#issuecomment-965643839

### DIFF
--- a/pkg/compose/convergence.go
+++ b/pkg/compose/convergence.go
@@ -596,8 +596,15 @@ func (s *composeService) isServiceHealthy(ctx context.Context, project *types.Pr
 		if container.State == nil || container.State.Health == nil {
 			return false, fmt.Errorf("container for service %q has no healthcheck configured", service)
 		}
-		if container.State.Health.Status != moby.Healthy {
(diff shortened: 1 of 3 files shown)


ndeloof (approved):

ndeloof:
for legal reasons we require commits to be signed-off.
please `git commit --sign-off --amend` the force push to your branch

benesch:
Whoops, sorry! I'm not in the habit. Fixed now.

felipecrs:
What should I do to revert to the previous behavior? I have an Oracle container which can take since 2-10 minutes to become healthy, which includes some time during unhealthy state.

felipecrs:
Nevermind. I think the correct thing to do is to adjust the healthcheck so that it does not report as unhealthy.

PR #9096, opened 2022-01-10
Return only numbers in short version

guillaumerose:
**What I did**

Python version of docker compose removes the v prefix in the version.

```
$ docker-compose version --short
1.25.5
```

Some projects are using this command to check the version. I think it might be good to keep the old behavior.
Example: https://github.com/openmaptiles/openmaptiles/blob/master/quickstart.sh#L84

**Related issue**

image

### DIFF
--- a/cmd/compose/version.go
+++ b/cmd/compose/version.go
@@ -18,6 +18,7 @@ package compose
 
 import (
 	"fmt"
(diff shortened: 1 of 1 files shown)


glours (approved):
LGTM

Issue #9058, opened 2021-12-21
Docker compose does not remove intermediate containers


dobryak:
**Description**

`docker-compose build` or `docker-compose up` commands do not remove intermediate containers created during building of images.

**Steps to reproduce the issue:**

Structure of the compose project:
```text
.
├── docker-compose.yml
└── test
    └── Dockerfile
```

`docker-compose.yml` file:
```yaml
version: "3.4"

services:
  test:
    build:
      context: ./test
```
`test/Dockerfile` file:
```dockerfile
FROM php:5.6-fpm-alpine

RUN set -e && apk add --no-cache git
```

Run the command `docker-compose up -d`.

**Describe the results you received:**

```text
CONTAINER ID   IMAGE                      COMMAND                  CREATED          STATUS                      PORTS                                   NAMES
84c65de05c14   test_prj_test              "docker-php-entrypoi…"   12 minutes ago   Up 12 minutes               9000/tcp                                test_prj-test-1
344eacdf6b3b   6feac8f2ef42               "/bin/sh -c 'set -e …"   12 minutes ago   Exited (0) 12 minutes ago                                           sweet_newton

```

As you can see the intermediate container `344eacdf6b3b` was not removed after building.

**Describe the results you expected:**

```text
CONTAINER ID   IMAGE                      COMMAND                  CREATED          STATUS                      PORTS                                   NAMES
84c65de05c14   test_prj_test              "docker-php-entrypoi…"   12 minutes ago   Up 12 minutes               9000/tcp                                test_prj-test-1
```
I expect to have only the target container. 

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `docker compose version`:**

```
Docker Compose version 2.2.2
```

**Output of `docker info`:**

```
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Build with BuildKit (Docker Inc., v0.6.1-docker)
  compose: Docker Compose (Docker Inc., 2.2.2)

Server:
 Containers: 26
  Running: 4
  Paused: 0
  Stopped: 22
 Images: 330
 Server Version: 20.10.11
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: false
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
[...]

PR #9097, opened 2022-01-11
Always remove intermediate containers on building with the classical builder

ulyssessouza:
**What I did**
Set `ForceRemove` to `true` when running the classical builder.

**Related issue**
https://github.com/docker/compose/issues/9058

### DIFF
--- a/pkg/compose/build_classic.go
+++ b/pkg/compose/build_classic.go
@@ -232,6 +232,7 @@ func imageBuildOptions(options buildx.Options) dockertypes.ImageBuildOptions {
 		Tags:        options.Tags,
 		NoCache:     options.NoCache,
 		Remove:      true,
+		ForceRemove: true,
 		PullParent:  options.Pull,
 		BuildArgs:   toMapStringStringPtr(options.BuildArgs),
 		Labels:      options.Labels,


thaJeztah:
Note that this only happens if a build fails, and the intent here is to allow the failed container to be kept for debugging, which is the default behavior on the classic builder; 

```bash
DOCKER_BUILDKIT=0 docker build -t foo -<<'EOF'
FROM busybox
RUN echo foo
RUN echo bar && exit 1
EOF
```

After the above build (which fails in the second run), you'll see that the container for the first step was removed, but the second one left for debugging.


```bash
docker ps -a
CONTAINER ID   IMAGE          COMMAND                  CREATED             STATUS                     PORTS     NAMES
06adcc0e3268   7ca9e84195d6   "/bin/sh -c 'echo ba…"   4 seconds ago       Exited (1) 2 seconds ago             amazing_germain
```

Do we know what compose v1 did in this case? I know there's people that rely on this behavior, and use the classic builder (exactly for the above), as BuildKit does not have a full equivalent of this feature; did we receive more requests for this? (because it looks like there's no opt-out for this change)

ulyssessouza:
@thaJeztah The trick here is that on V2 we made this option hidden. So the user doesn't have the control over it anymore.

thaJeztah:
Yes, so BuildKit doesn't support the option. What was the default on v1 though?

ulyssessouza:
@thaJeztah The default for v1 is to not remove the container. So that would change the behaviour...
Another possibility is to add a (deprecated?) flag just for this case. But that sounds itchy for me to add a flag for something that we would want to remove in the future...

thaJeztah:
So my thinking is that;

- Compose V1 didn't do this, so changing this in v2 will be a change in behavior; with v2 (afaik) only supporting the classic builder for Windows images, it would only work for those; and as such won't resolve https://github.com/docker/compose/issues/9058); to resolve https://github.com/docker/compose/issues/9058, it would have to be implemented in compose v1.
- It's not consistent with `docker build` (which _does_ keep the containers in case of a failed build, when using the classic builder)
- For those using the classic builder, there are _known_ use-cases for keeping the build-containers for failed builds (to be able to debug the failed build)
- With the intent to deprecate the classic builder (for Linux images), I'm not sure we should introduce a new flag just for the legacy builder case.

In addition; given that Compose v1 has been in use for many years, and this is (may be?) the first time this feature has been requested, and only a few commenters, so I'd definitely not "hurry" to implement this. The ultimate goal is to use BuildKit for both Linux and Windows containers, and when that arrives, this feature would become obsolete.

ulyssessouza:
@thaJeztah Agreed!

PR #9108, opened 2022-01-13
Support COMPOSE_IGNORE_ORPHANS for `compose run`

PIG208:
**What I did**
Made `compose run` respect the env var COMPOSE_IGNORE_ORPHANS. E2E test cases added.

**Related issue**

This revives #7020 and resolves the issue mentioned in [#4992 (comment)](https://github.com/docker/compose/pull/4992#issuecomment-696569463).

mimi

### DIFF
--- a/cmd/compose/run.go
+++ b/cmd/compose/run.go
@@ -53,6 +53,7 @@ type runOptions struct {
 	servicePorts  bool
 	name          string
 	noDeps        bool
+	ignoreOrphans bool
 	quietPull     bool
 }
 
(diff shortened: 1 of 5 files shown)


PIG208:
@ndeloof @ulyssessouza any feedbacks will be greatly appreciated

juninhocruzg3:
+1
I need to use `COMPOSE_IGNORE_ORPHANS` on `docker-compose run`, but after #4992 it works only for `docker-compose up`.
Thank you for revive it!

ndeloof (changes requested):
LGTM from a code and feature point of view
reference doc need to be updated to reflect this new option: https://github.com/docker/compose/blob/v2/docs/reference/docker_compose_run.yaml

ndeloof (approved):
Oh yes indeed, env var only. But nice addition to the docs, thanks

PIG208:
The missing file that caused the CI to fail was added.

PR #9148, opened 2022-02-04
Using start, stop, restart from outside the working directory using --project-name (#9147)

### LINKED ISSUE #9147: Using start, stop, restart from outside the working directory using --project-name
Being able to start, stop and restart projects without changing directory could really simplify managing projects (at least for me).
I'm going to implement this feature and submit a pull request so that you can merge if you think it useful.

arhemd:
Being able to start, stop and restart projects without changing directory could really simplify managing projects (at least for me).
Merge if you think it useful too.

Fixes #9147


arhemd:
@glours @ndeloof PTAL

pkg/compose/compose.go
```diff
@@ -92,3 +92,34 @@ func escapeDollarSign(marshal []byte) []byte {
 	escDollar := []byte{'$', '$'}
 	return bytes.ReplaceAll(marshal, dollar, escDollar)
 }
+
+// projectFromName builds a types.Project based on actual resources with compose labels set
+func (s *composeService) projectFromName(containers Containers, projectName string) *types.Project {
```
ndeloof, line 97: This is basically a duplicate for `projectFromLabels`, better use/adapt this existing function

arhemd: Yes, that's right but `projectFromName` exists in the file `down.go ` and using that didn't seem right structure-wise.
I would handle this by removing it from `down.go` and modifying the code to use this one instead.
Does it sound good?

ndeloof: yes, move it to `compose.go`. Also, probably should be renamed `projectFromActualResources` or something comparable to be more explicit

arhemd: - Removed `projectFromLabels` from `down.go`
- Fixed an issue causing some tests to fail
I couldn't think of a short and non-misleading alternative for `projectFromName` tho.

ndeloof (approved):

arhemd:
@ndeloof I had to modify the label `com.docker.compose.depends_on` to the format `dependency:condition` to make `depends_on` work properly and I made sure my change is backward compatible. Is this change ok with you?

pkg/compose/create.go
```diff
@@ -433,7 +433,7 @@ func (s *composeService) prepareLabels(service types.ServiceConfig, number int)
 
 	var dependencies []string
 	for s := range service.DependsOn {
-		dependencies = append(dependencies, s)
+		dependencies = append(dependencies, s+":"+service.DependsOn[s].Condition)
```
ndeloof, line 436: can get this by iterator `for s, condition := range service.DependsOn`

pkg/compose/compose.go
```diff
@@ -92,3 +92,42 @@ func escapeDollarSign(marshal []byte) []byte {
...
+		}
+		dependencies := c.Labels[api.DependenciesLabel]
+		if len(dependencies) > 0 {
+			service.DependsOn = types.DependsOnConfig{}
+			for _, dc := range strings.Split(dependencies, ",") {
+				dcArr := strings.Split(dc, ":")
+				condition := ServiceConditionRunningOrHealthy
+				dependency := dcArr[0]
+
+				// backward compatibility
```
ndeloof, line 122: 👍

ndeloof (commented):
LGTM with some minor fixes

pkg/compose/create.go
arhemd: Done

ndeloof (approved):

pkg/compose/stop.go
```diff
@@ -21,29 +21,31 @@ import (
 
 	"github.com/docker/compose/v2/pkg/api"
 	"github.com/docker/compose/v2/pkg/progress"
-
-	"github.com/compose-spec/compose-go/types"
+	//"github.com/compose-spec/compose-go/types"
```
glours, line 24: nit : remove if not necessary

glours (changes requested):
Thank you so much @arhemd for you contribution 
Currently when you try one of this commands with an non-existent service you got the following error
```shell
> docker compose -f ~/sources/awesome-compose/nginx-golang-mysql/docker-compose.yaml restart frontend

no such service: frontend
```

Can you add the same behaviour when using the `--project-name` ?

pkg/compose/stop.go
arhemd: You're very welcome.

> Can you add the same behaviour when using the `--project-name` ?

Added this error check to `start`, `stop` and `restart`

> nit : remove if not necessary

Done.


Also added `service.Scale` to `projectFromName` to maintain compatibility with https://github.com/docker/compose/commit/8a9498c5710c8282dba42241b3217d8a0a366ff9

pkg/compose/start.go
```diff
@@ -28,15 +28,22 @@ import (
...
+		return s.start(ctx, projectName, options, nil)
 	})
 }
 
-func (s *composeService) start(ctx context.Context, project *types.Project, options api.StartOptions, listener api.ContainerEventListener) error {
-	if len(options.AttachTo) == 0 {
-		options.AttachTo = project.ServiceNames()
+func (s *composeService) start(ctx context.Context, projectName string, options api.StartOptions, listener api.ContainerEventListener) error {
+	var containers Containers
+	containers, err := s.getContainers(ctx, projectName, oneOffExclude, true, options.AttachTo...)
```
glours, line 39: Containers are used in the `projectFromName` function to detect and populate the services section of the project. 
If we pass the `options.AttachTo` array of services to `getContainers`, we'll only get the containers running services declared in the command line. 

If your compose stack declare services with dependencies you could get `no such service: XXX` error.
For example if you try the sample `nginx-golang-mysql` from the `awesome-compose` repository and try the following command `docker compose --project-name nginx-golang-mysql start proxy` you will get this error
`no such service: backend`

So we need to get all the all containers from a project and not only the ones linked to services passed as parameter of the command

```suggestion
	containers, err := s.getContainers(ctx, projectName, oneOffExclude, true)
```

arhemd: Thanks for pointing this out
I made the change and tested it with `nginx-golang-mysql` and it appears to be fixed.

Issue #8379, opened 2021-06-11
support --env-file in docker-compose run command


briceburg:
NOTE: this is different than the `.env` file that is read by docker-compose commands and used for substitutions in the composition file... first mentioned here: https://github.com/docker/compose/issues/6170#issuecomment-673914083

---

We generate an env file that we would like to include via the `docker-compose run` command. e.g.

```
echo "FOO=bar" > vars.env
docker-compose run --env-file=vars.env --rm test-service
```

`docker run` supports this flag and it would be great if we saw `docker-compose run` maintain parity. 

Is there planned support for this feature -- or will it be implemented in composev2 / "docker <space> compose" ? 

our generated vars file is cumbersome (with multiline values) and relatively complicated to convert to various `-e FOO=bar` flags...

thanks!

KoditkarVedant:
@chris-crone @ndeloof Hi guys, I would like to work on this issue. I am not new to Go lang. 

I have got a question. If I pass `--env-file` option to run command then should we append that to `env-file` provided as part of service config in docker-compose.yaml file or should we override them?

KoditkarVedant:
In meantime I've created a draft PR which assumes that if you pass --env-file from CLI it overrides the env-file specified in docker-compose.yaml file. I will update PR as per your suggestion if we should append them.

KoditkarVedant:
@chris-crone @ndeloof @ulyssessouza Friendly ping.

briceburg:
I would think would want to append the --env-file(s) passed via the cli to any included in the service's definition. appending would override any previously defined keys.

reggermont:
I'm surprised this issue hasn't more discussions.

Just had this case in a project and that's a deal breaker, will have to generate a huge `docker run` command just because of that issue

I think there's a misconception with the original `--env-file` option which should never have the same name than `docker`'s `--env-file`.

Now [the topic is stuck because it may conflict with the original `--env-file`](https://github.com/docker/compose/pull/9169#issuecomment-1065099236)

What about a `--env-from` instead?

ndeloof:
That's indeed sad this flag name was selected, but this is the legacy we have to deal with.
We could use `--service-env-file` or something comparable, even verbose, to workaround this ambiguity

KoditkarVedant:
@ndeloof I have updated the PR with changes suggested.

PR #9169, opened 2022-02-13
Add support to pass env-file to docker compose run

KoditkarVedant:
**What I did**
This pull request adds support to pass in `--env-file` to `docker compose run` command.

Behavior of the option:
If you pass env-file from CLI it overrides the env-files provided in `docker-compose.yaml` config.

**Related issue**
#8379 

image

### DIFF
--- a/cmd/compose/run.go
+++ b/cmd/compose/run.go
@@ -36,28 +36,28 @@ import (
 
 type runOptions struct {
 	*composeOptions
-	Service       string
-	Command       []string
-	environment   []string
-	Detach        bool
-	Remove        bool
-	noTty         bool
-	tty           bool
-	interactive   bool
-	user          string
-	workdir       string
-	entrypoint    string
-	entrypointCmd []string
-	labels        []string
-	volumes       []string
-	publish       []string
-	useAliases    bool
-	servicePorts  bool
-	name          string
-	noDeps        bool
-	ignoreOrphans bool
-	quietPull     bool
-	EnvFiles      []string
+	Service         string
+	Command         []string
+	environment     []string
+	Detach          bool
+	Remove          bool
+	noTty           bool
+	tty             bool
+	interactive     bool
+	user            string
+	workdir         string
+	entrypoint      string
+	entrypointCmd   []string
+	labels          []string
+	volumes         []string
+	publish         []string
+	useAliases      bool
+	servicePorts    bool
+	name            string
+	noDeps          bool
+	ignoreOrphans   bool
+	quietPull       bool
+	ServiceEnvFiles []string
 }
 
 func (opts runOptions) apply(project *types.Project) error {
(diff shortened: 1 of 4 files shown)


ndeloof (approved):

ndeloof:
command documentation also need to be updated to reflect this new flag: https://github.com/docker/compose/blob/v2/docs/reference/docker_compose_run.yaml

KoditkarVedant:
@ndeloof Looks like there is bug in the implementation. It isn't working as expected. I think when we parse the compose config files at that time only the `env_file` mentioned in it gets parsed. because when I see the `service.Environment` it was already populated with the key=value mentioned in the `env_file`. So my understanding that just replacing the `service.EnvFile` will work is not true. 

I'll just put back this PR in draft as it is not ready for merge.

ndeloof:
The main issue I see adding support for `compose run --env-file ... ` to set container environment from a file (which would make sense to align with `docker run ...`) is that this will introduce confusion with the top-level `--env-file` used to override `.env` file used to interpolate variables. i.e. 
`docker compose --env-file xx run` is a distinct thing vs `docker compose run --env-file xx` 
I'm not sure how we could address this UX issue. Any thoughts @glours @ulyssessouza @thaJeztah ?

ulyssessouza:
Same for me... The UX doesn't help... And specially because the new one is an array and the root command one is a single file.

cmd/compose/run.go
```diff
@@ -141,6 +142,7 @@ func runCommand(p *projectOptions, backend api.Service) *cobra.Command {
 	flags := cmd.Flags()
 	flags.BoolVarP(&opts.Detach, "detach", "d", false, "Run container in background and print container ID")
 	flags.StringArrayVarP(&opts.environment, "env", "e", []string{}, "Set environment variables")
+	flags.StringArrayVar(&opts.EnvFiles, "env-file", []string{}, "Read in a file of environment variables")
```
ndeloof, line 145: I suggest we use `--service-env-file` to avoid ambiguity with upper-level compose command --env-file flag

thaJeztah: If we want to retain a parallel between `docker run` and `docker compose run`, then I'm wondering if we should look at doing the reverse, and think of a new name for the flag for setting the `.env`  (we could keep the old flag, but hidden)

pkg/compose/run.go
```diff
@@ -127,4 +127,8 @@ func applyRunOptions(project *types.Project, service *types.ServiceConfig, opts
 	for k, v := range opts.Labels {
 		service.Labels = service.Labels.Add(k, v)
 	}
+
+	if opts.ServiceEnvFiles != nil {
+		service.EnvFile = append(service.EnvFile, opts.ServiceEnvFiles...)
```
ndeloof, line 132: need to make service envfile an absolute path

ndeloof (changes requested):
LGTM but path to service env file passed by command line flag could be relative to working directory

pkg/compose/run.go
KoditkarVedant: is there any sample code which I can refer in the repository?

ndeloof: You can just use:
```go
if !filepath.IsAbs(file) {
					file, err = filepath.Abs(file)
}
```

cmd/compose/run.go
```diff
@@ -156,6 +157,7 @@ func runCommand(p *ProjectOptions, streams api.Streams, backend api.Service) *co
 	flags := cmd.Flags()
 	flags.BoolVarP(&opts.Detach, "detach", "d", false, "Run container in background and print container ID")
 	flags.StringArrayVarP(&opts.environment, "env", "e", []string{}, "Set environment variables")
+	flags.StringArrayVar(&opts.ServiceEnvFiles, "service-env-file", []string{}, "Read in a file of environment variables")
```
ndeloof, line 160: Suggestion: rename flag `--env-from-file`, wdyt ?

ndeloof:
I cherry-picked your code into https://github.com/docker/compose/pull/12626

PR #9179, opened 2022-02-16
as --no-build is set, set service image to default image name

### LINKED ISSUE #8785: Can't use --no-build with services with a build section

ndeloof:
**What I did**
when `--no-build` is set, enforce service image to have a value, using default image name logic.

**Related issue**
close #8785

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -445,3 +445,12 @@ const (
 	// UserCancel user cancelled compose up, we are stopping containers
 	UserCancel
 )
+
+// GetImageNameOrDefault computes the default image name for a service, used to tag built images
+func GetImageNameOrDefault(service types.ServiceConfig, projectName string) string {
+	imageName := service.Image
+	if imageName == "" {
+		imageName = projectName + "_" + service.Name
+	}
+	return imageName
+}
--- a/cmd/compose/create.go
+++ b/cmd/compose/create.go
@@ -126,6 +126,9 @@ func (opts createOptions) Apply(project *types.Project) {
 	if opts.noBuild {
 		for i, service := range project.Services {
 			service.Build = nil
+			if service.Image == "" {
+				service.Image = api.GetImageNameOrDefault(service, project.Name)
+			}
 			project.Services[i] = service
 		}
 	}
--- a/pkg/compose/build.go
+++ b/pkg/compose/build.go
@@ -64,7 +64,7 @@ func (s *composeService) build(ctx context.Context, project *types.Project, opti
 		if service.Build == nil {
 			continue
 		}
-		imageName := getImageName(service, project.Name)
+		imageName := api.GetImageNameOrDefault(service, project.Name)
 		imagesToBuild = append(imagesToBuild, imageName)
 		buildOptions, err := s.toBuildOptions(project, service, imageName, options.SSHs)
 		if err != nil {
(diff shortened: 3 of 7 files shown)


pkg/api/api.go
```diff
@@ -445,3 +445,12 @@ const (
 	// UserCancel user cancelled compose up, we are stopping containers
 	UserCancel
 )
+
+// GetDefaultImageName computes the default image name for a service, used to tag built images
+func GetDefaultImageName(service types.ServiceConfig, projectName string) string {
+	imageName := service.Image
+	if imageName == "" {
+		imageName = projectName + "_" + service.Name
```
milas, line 453: @glours Should we make this `-` to align with other resources (compose-spec/compose-go#294)?

milas: Oh, I see this is actually not a change in this PR, as this logic existed perviously as `getImageName` and is just exported now, so don't need to address before merging this even if we want to change

pkg/api/api.go
```diff
@@ -445,3 +445,12 @@ const (
 	// UserCancel user cancelled compose up, we are stopping containers
 	UserCancel
 )
+
+// GetDefaultImageName computes the default image name for a service, used to tag built images
+func GetDefaultImageName(service types.ServiceConfig, projectName string) string {
```
milas, line 450: I'd propose renaming this to `GetImageNameOrDefault` since it reads `service.Image` if populated and then falls back to producing one. Particularly now that it's exported 😉

milas (changes requested):
I think the exported function should be renamed as it's a bit misleading right now, but otherwise LGTM

pkg/api/api.go
glours: It's less important for the image name, but I'll add a commit to use the compose separator 👍

glours: I'll do it in a PR to fix this issue #9618

milas (approved):
🚀

Issue #9188, opened 2022-02-18
lowercase dockerfile is broken since v2.1


exe-dealer:
**Description**

docker-compose cannot find lowercase dockerfile since v2.1, capitalized Dockerfile supported only

**Steps to reproduce the issue:**
```
# /tmp/proj1/dockerfile
FROM alpine
```

```yaml
# /tmp/proj1/compose.yml
version: '3.6'
services:
  hello:
    build: .
```

```sh
$ cd /tmp/proj1
$ docker compose build
```

**Describe the results you received:**

```
1 error occurred:
	* unable to prepare context: unable to evaluate symlinks in Dockerfile path: lstat /tmp/proj1/Dockerfile: no such file or directory
```

**Describe the results you expected:**

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `docker compose version`:**

```
Docker Compose version v2.2.3
```

**Output of `docker info`:**

```
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  app: Docker App (Docker Inc., v0.9.1-beta3)
  buildx: Build with BuildKit (Docker Inc., v0.6.3-docker)
  compose: Docker Compose (Docker Inc., v2.2.3)
  scan: Docker Scan (Docker Inc., v0.8.0)

Server:
 Containers: 9
  Running: 7
  Paused: 0
  Stopped: 2
 Images: 33
 Server Version: 20.10.9
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 1
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: active
  NodeID: y1mkd0l9bzb7lfbhp2448bqw5
  Is Manager: true
  ClusterID: 0ah4j2sti8t949x6hq3zodnyo
  Managers: 1
  Nodes: 1
  Default Address Pool: 10.0.0.0/8  
  SubnetSize: 24
  Data Path Port: 4789
  Orchestration:
   Task History Retention Limit: 5
  Raft:
   Snapshot Interval: 10000
   Number of Old Snapshots to Retain: 0
   Heartbeat Tick: 1
   Election Tick: 10
  Dispatcher:
   Heartbeat Period: 5 seconds
[...]

PR #9189, opened 2022-02-18
bump buildx to 0.7.1

ndeloof:
**What I did**
bump buildx to 0.7.1
used same `replace` directives used by buildx, which are required to align latest change on moby's `fsutil` with docker/cli to depend on some Moby's functions which had signature changed 🤯 

**Related issue**
close https://github.com/docker/compose/issues/9188


pkg/compose/build_buildkit.go
```diff
@@ -47,7 +47,7 @@ func (s *composeService) doBuildBuildkit(ctx context.Context, project *types.Pro
 	w := xprogress.NewPrinter(progressCtx, os.Stdout, mode)
 
 	// We rely on buildx "docker" builder integrated in docker engine, so don't need a DockerAPI here
-	response, err := build.Build(ctx, driverInfo, opts, nil, nil, w)
+	response, err := build.Build(ctx, driverInfo, opts, nil, path.Dir(s.configFile.Filename), w)
```
glours, line 50: Maybe we should use `filepath.Dir` to avoid issue with windows filesytem?

glours (commented):
LGTM if we sure we won't have issue on windows filesystem with `path.Dir`

glours (approved):
LGTM

PR #9198, opened 2022-02-23
composeService to use dockerCli's In/Out/Err streams

ndeloof:
**What I did**
composeService is created from dockerCli and uses the assigned In/Out/Err streams
this also will let us 
- use [BuildKitEnabled ](https://github.com/docker/cli/blob/master/cli/command/cli.go#L60) recently introduced in the docker/cli
- rely on more [code from the docker/cli](https://github.com/docker/cli/pull/3436), to avoid code duplication by copy/paste or re-implementation (with possible bugs)

**Related issue**


glours (approved):
LGTM

PR #9200, opened 2022-02-23
COMPOSE_REMOVE_ORPHANS can be set to always apply --remove-orphans

### LINKED ISSUE #9199: Feature request: Add COMPOSE_REMOVE_ORPHANS to let us default to --remove-orphans on a per project basis
Hi,

I find myself always using `docker-compose down --remove-orphans` because if I rename a service or happen to have `run` a 1 off task and Docker created a container with a random suffix I always want to clean those containers up.

I recently ran a poll on Twitter asking what other folks prefer https://twitter.com/nickjanetakis/status/1495458224912351242 and ~83% said they would like to have `--remove-orphans` set by default. I even set this as the 2nd option in the poll to eliminate a bias where someone might just choose the first item in the poll.

But since that's a breaking change @ndeloof suggested that maybe we can add a new `COMPOSE_REMOVE_ORPHANS` environment variable that when set would remove orphans. This is nice because we can set a single environment once and have it happen by default on a per project basis without making any backwards incompatible changes.

What do you think?

The proposed behavior would be things continue to work as they do today but if you set `COMPOSE_REMOVE_ORPHANS=1` or any value then it would remove orphans without needing to define `--remove-orphans` every time you run `docker-compose down`.

It would be nice if this were also applied to Docker Compose v1 because there's still issues with Docker Compose v2 where it doesn't parse YAML in the same way as Docker Compose v1 but that's a separate matter. I'm just trying to say that I still use v1.

Thanks!

ndeloof:
**What I did**
`--remove-orphans` default value is computed based on `COMPOSE_REMOVE_ORPHANS` env variable

**Related issue**
closes #9199

### DIFF
--- a/cmd/compose/down.go
+++ b/cmd/compose/down.go
@@ -19,6 +19,8 @@ package compose
 import (
 	"context"
 	"fmt"
+	"os"
+	"strings"
 	"time"
 
 	"github.com/compose-spec/compose-go/types"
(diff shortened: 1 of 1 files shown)


glours (approved):
LGTM

nickjj:
That was the fastest issue -> merged PR ever, thanks a lot!

PR #9214, opened 2022-02-28
exclude com.docker.compose.image label from service hash

### LINKED ISSUE #9211: `docker config --hash` output does not match `com.docker.compose.config-hash` container label
Mostly copied from my comment [here](https://github.com/docker/compose-cli/issues/935#issuecomment-1019066293).

While the hash scheme used by `--hash` subcommand for `config`/`convert` is the same, the result it produces never matches the hash stored in the `com.docker.compose.config-hash` container label in compose V2 - a regression from V1.

With `docker-compose`, it used to be possible to compare the result of `docker-compose config --hash=<service name>` to the config hash stored in the container label to determine if the config had been changed since the container had started, and needed recreating (to show as a status or warning).

As best as I can tell, this is due to `docker compose` changing the Service object in various ways before hashing it for the container label, which are not replicated before hashing it for the `docker compose config --hash` subcommand. For example, in some testing I found that just before the Service object is hashed either for storing in the container label ([here](https://github.com/docker/compose/blob/bfd7428619e7c0ca9aaedebaaa14ac0fc82e04b4/pkg/compose/create.go#L430)) or comparing against that label ([here](https://github.com/docker/compose/blob/bfd7428619e7c0ca9aaedebaaa14ac0fc82e04b4/pkg/compose/convergence.go#L196)), the digest of the available image has been added as a label in the Service config ([here](https://github.com/docker/compose/blob/bfd7428619e7c0ca9aaedebaaa14ac0fc82e04b4/pkg/compose/build.go#L144)) and any relative paths have been expanded. These changes have not been made when `config --hash` generates it's hash, and so the two never actually match any more.

The path expansion probably makes sense, and could be added in as a step in `config --hash`, but I'm not sure why adding the current image digest into the Service object is necessary for detecting convergence - this digest is already stored in the `com.docker.compose.image` container label, and can be directly compared.

Whether it is by allowing these hashes to match, or by implementing a completely new dry run system like [#8551](https://github.com/docker/compose/issues/8551) (though I suspect matching the hash mechanisms will be easier), providing a mechanism for external tools to determine if a container is running from an outdated config would be very useful.

ndeloof:
**What I did**
set `com.docker.compose.image label` as a custom label so it get excluded from config hash computation

**Related issue**
closes #9211

### DIFF
--- a/pkg/compose/build.go
+++ b/pkg/compose/build.go
@@ -141,7 +141,7 @@ func (s *composeService) ensureImagesExists(ctx context.Context, project *types.
 			if project.Services[i].Labels == nil {
 				project.Services[i].Labels = types.Labels{}
 			}
-			project.Services[i].Labels[api.ImageDigestLabel] = digest
+			project.Services[i].CustomLabels[api.ImageDigestLabel] = digest
 			project.Services[i].Image = image
 		}
 	}


glours (approved):
LGTM

Issue #9207, opened 2022-02-25
Proposal: add `-i` and `-t` flags on `docker compose run` and `exec`


thaJeztah:
The defaults for `-t` / `--tty` (allocate a pseudo-TTY) and `-i` / `--interactive` (Keep STDIN open even if not attached) for `docker compose exec` and `docker compose run` are exactly the reverse as the corresponding `docker (container) run` / `docker container exec` commands.

For `docker compose`, users can _disable_ these options using the `-T` / `--no-TTY` options; we'll unlikely be able to change these defaults on either side (also see https://github.com/docker/compose/issues/9202), but we could make the UX slightly more compatible by adding the same flags to `docker compose` subcommands.

Currently, trying to use the `-i` and/or `-t` options will produce an error;

```bash
docker compose run -it myservice
unknown shorthand flag: 'i' in -it

docker compose exec -it myservice
unknown shorthand flag: 'i' in -it
```

I propose to add these flags to (at least) the `run` and `exec` subcommands. Setting those options won't have an effect (as they're already set by default), but it allows users to use the same options when using compose (which can be just due to "muscle memory");

- Add `-t` / `--tty` flag, with a default of `true`
- Add `-i` / `--interactive` flag with a default of `true`
- Produce an error passing _both_ `-t` or `-i` _and_ `-T` (conflicting options)

I should note that there's other commands on the `docker` / `docker container` that also accept `-t` and/or `-i`, such as `docker (container) create`, and  a `--no-stdin` on `docker (container) attach` (but compose does not yet have an equivalent to that).

PR #9229, opened 2022-03-04
Add -i and -t to run and exec

ulyssessouza:
**What I did**
Add `-i` and `-t` to commands `run` and `exec` as hidden options, since they are not used.
This is just to avoid breakage on scripts or old usages from users.

**Related issue**
Resolves https://github.com/docker/compose/issues/9207


glours (approved):
LGTM

cmd/compose/exec.go
```diff
@@ -73,6 +73,11 @@ func execCommand(p *projectOptions, backend api.Service) *cobra.Command {
 	runCmd.Flags().BoolVarP(&opts.noTty, "no-TTY", "T", false, "Disable pseudo-TTY allocation. By default `docker compose exec` allocates a TTY.")
 	runCmd.Flags().StringVarP(&opts.workingDir, "workdir", "w", "", "Path to workdir directory for this command.")
 
+	runCmd.Flags().BoolP("interactive", "i", true, "Keep STDIN open even if not attached. DEPRECATED")
```
ndeloof, line 76: Those are **not** DEPRECATED, just on by default
(also, if we want to deprecate flag, should use flags.MarkDeprecated)

PR #9231, opened 2022-03-04
Remove DEPRECATED text, since it's just the default

ulyssessouza:
**What I did**
Remove DEPRECATED text, since it's just the default

**Related issue**
https://github.com/docker/compose/pull/9229#discussion_r819730788

### DIFF
--- a/cmd/compose/exec.go
+++ b/cmd/compose/exec.go
@@ -70,9 +70,9 @@ func execCommand(p *projectOptions, backend api.Service) *cobra.Command {
 	runCmd.Flags().BoolVarP(&opts.noTty, "no-TTY", "T", false, "Disable pseudo-TTY allocation. By default `docker compose exec` allocates a TTY.")
 	runCmd.Flags().StringVarP(&opts.workingDir, "workdir", "w", "", "Path to workdir directory for this command.")
 
-	runCmd.Flags().BoolP("interactive", "i", true, "Keep STDIN open even if not attached. DEPRECATED")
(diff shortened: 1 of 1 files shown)


ndeloof (commented):
👍 

strictly speaking, we _should_ handle `--interactive=false` and not just make this flag a No-Op, but I guess that's a corner case :D

thaJeztah:
Needs a rebase, @ulyssessouza

ndeloof (approved):

thaJeztah:
> strictly speaking, we should handle --interactive=false and not just make this flag a No-Op, but I guess that's a corner 

Yes, would be good to support that as well; also (from https://github.com/docker/compose/issues/9207#issue-1150294771) probably produce an error when passing _both_ `-t` or `-i`  _and_  `-T` (conflicting options)

ndeloof:
Already added support for `--interactive=false` on https://github.com/docker/compose/blob/v2/cmd/compose/run.go#L223
Support for combinations of `--tty=false` and `--no-tty` is a bit more complex, will look into this on my spare time :)

thaJeztah:
Yes, that one is more tricky; if the options do not "conflict", we could accept them, but if they "conflict", it's more tricky to define the behaviour (which one should take precedence??). In such cases, I usually make that an error-condition; the advantage with that is that you're sure nobody is able to use the combination (as it errors), which leaves the ability to "allow" the combination in future (if there's a decision on what the expected outcome is).

ndeloof:
indeed. I planned to rely on `flags.Changed` to check is `tty` is true because this is default _or_ user explicitly user `-t`, so that we can detect conflict with `-T`

PR #9238, opened 2022-03-07
don't failed when trying to stop or rm services with no containers ru…

glours:
…nning

**What I did**
Don't fail the process when trying to stop or rm services when no container are running. We check the error and failed only if not concerning number of containers running.

### DIFF
--- a/pkg/compose/stop.go
+++ b/pkg/compose/stop.go
@@ -44,7 +44,7 @@ func (s *composeService) stop(ctx context.Context, projectName string, options a
 	}
 
 	project, err := s.projectFromName(containers, projectName, services...)
-	if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/compose/stop.go
```diff
@@ -44,7 +44,7 @@ func (s *composeService) stop(ctx context.Context, projectName string, options a
 	}
 
 	project, err := s.projectFromName(containers, projectName, services...)
-	if err != nil {
+	if err != nil && !strings.HasPrefix(err.Error(), "no container found for project") {
```
ndeloof, line 47: use `api.IsNotFoundError(err)`

ndeloof (approved):

PR #9242, opened 2022-03-07
Fix json format for version command

### LINKED ISSUE #9236: Broken json format for version subcommad
**Description**

```sh
./bin/docker-compose version -f json
```

`{"version":"v2.3.0-2-gfcff3963"}\n`
The newline character was not escaped properly

`{"version":"v2.3.0-2-gfcff3963"}`
With a newline.

The json format is printed with:
```go
fmt.Printf(`{"version":%q}\n`, internal.Version)
```

I think we should either use double-quotes and escape the inner quotes, instead of backticks because the newline is not escaped properly.
```go
fmt.Printf("{\"version\":%q}\n", internal.Version)
```
Or add a newline character after the closing curly bracket like:
```go
fmt.Printf(`{"version":%q}
`, internal.Version)
```

I think the better option would be the first.

victor-timofei:
**What I did**
Fixed JSON formating for the version command.

**Related issue**
fixes #9236

### DIFF
--- a/cmd/compose/version.go
+++ b/cmd/compose/version.go
@@ -57,7 +57,7 @@ func runVersion(opts versionOptions) {
 		return
 	}
 	if opts.format == formatter.JSON {
-		fmt.Printf(`{"version":%q}\n`, internal.Version)
+		fmt.Printf("{\"version\":%q}\n", internal.Version)
 		return
 	}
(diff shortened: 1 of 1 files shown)


ndeloof (approved):

ndeloof:
thanks for catching this!

PR #9247, opened 2022-03-08
use plain text progress when ansi=never is set

ndeloof:
**What I did**
make progress mode configurable, and adapt to `--ansi` flag

usage: `docker compose --ansi=never up  ...` will render progress events as plain text


glours (approved):
LGTM

Issue #9249, opened 2022-03-09
`no such service` in Compose v2.3.2 when running/stopping services from multiple Compose files


antoineco:
**Description**

Compose v2.3.2 still seems to be struggling when juggling with services defined in multiple Compose files, and yields `no such service`: https://github.com/deviantony/docker-elk/runs/5473593690?check_suite_focus=true

**Steps to reproduce the issue:**

In the above example, taken from a CI run, I ran the following commands inside the repository https://github.com/deviantony/docker-elk:

```
docker compose up -d
docker compose -f docker-compose.yml -f extensions/logspout/logspout-compose.yml up -d logspout
docker compose -f docker-compose.yml -f extensions/logspout/logspout-compose.yml ps
docker compose -f docker-compose.yml -f extensions/logspout/logspout-compose.yml logs logspout
docker compose -f docker-compose.yml -f extensions/logspout/logspout-compose.yml stop logspout
docker compose stop logstash
```

The last `stop` caused the aforementioned error. (The `logstash` service is defined inside the main `docker-compose.yml` file)

**Describe the results you received:**

`no such service: logstash`

**Describe the results you expected:**

A success, like in v2.2 and below.

**Additional information you deem important (e.g. issue happens only occasionally):**

I started noticing this along with the error described in #9234.
The `no such project` error went away. However, the `no such service` error is still present in the latest version.

**Output of `docker compose version`:**

```
Docker Compose version v2.3.2
```

**Output of `docker info`:**

```
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Build with BuildKit (Docker Inc., 0.7.1+azure)

Server:
 Containers: 0
  Running: 0
  Paused: 0
  Stopped: 0
 Images: 16
 Server Version: 20.10.11+azure-3
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: false
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 1
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 io.containerd.runtime.v1.linux runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: d71fcd7d8303cbf684402823e425e9dd2e99285d
[...]

Issue #9241, opened 2022-03-07
Stopping container with dependencies leads to `no such service: dependent-service` (2.3.1 regression)


nocive:
**Description**

Attempting to stop a running container with dependencies leads to a `no such service: dependent-service` error.
Repeating the same will yield back different dependent service names.

This broke in `2.3.1` and worked fine before.

**Steps to reproduce the issue:**
1. `docker compose stop some-container`

**Describe the results you received:**
 `no such service: dependent-service`

**Describe the results you expected:**
Stopping the container successfully.

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `docker compose version`:**

```
Docker Compose version v2.3.1
```

**Output of `docker info`:**

```
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Build with BuildKit (Docker Inc., v0.6.1-docker)
  compose: Docker Compose (Docker Inc., v2.2.3)

Server:
 Containers: 82
  Running: 48
  Paused: 0
  Stopped: 34
 Images: 181
 Server Version: 20.10.11
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: false
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 io.containerd.runtime.v1.linux runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 1e5ef943eb76627a6d3b6de8cd1ef6537f393a71.m
 runc version: v1.0.3-0-gf46b6ba2
 init version: de40ad0
 Security Options:
  seccomp
   Profile: default
  cgroupns
 Kernel Version: 5.15.7-arch1-1
 Operating System: Antergos Linux
 OSType: linux
 Architecture: x86_64
 CPUs: 4
 Total Memory: 15.49GiB
 Name: jps-x1
 ID: TX6D:ALDF:OGX5:HVBY:YJNY:FLH2:GRJI:I52W:V3K3:L6ZT:S2YY:23UM
 Docker Root Dir: /var/lib/docker
 Debug Mode: false
 Username: nocive
 Registry: https://index.docker.io/v1/
 Labels:
 Experimental: true
[...]

PR #9250, opened 2022-03-09
build full compose model from resources, then filter by services

ndeloof:
**What I did**
introduced `actualState` (this reduces code duplication and risk to have same bug in other places) to list resources, rebuild compose project model, then filter by requested services

fix `docker compose stop SERVICE` with reference to other dependencies
bonus: `docker compose rm` doesn't need a compose file anymore

includes e2e test so we get rid of this bug and don't need yet another patch release🤞

**Related issue**
closes https://github.com/docker/compose/issues/9241
closes https://github.com/docker/compose/issues/9249

### DIFF
--- a/pkg/compose/compose.go
+++ b/pkg/compose/compose.go
@@ -149,3 +149,23 @@ SERVICES:
 
 	return project, nil
 }
+
+// actualState list resources labelled by projectName to rebuild compose project model
+func (s *composeService) actualState(ctx context.Context, projectName string, services []string) (Containers, *types.Project, error) {
+	var containers Containers
+	// don't filter containers by options.Services so projectFromName can rebuild project with all existing resources
+	containers, err := s.getContainers(ctx, projectName, oneOffInclude, true)
+	if err != nil {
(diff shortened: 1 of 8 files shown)


glours (approved):
LGTM

pkg/compose/compose.go
```diff
@@ -133,19 +176,41 @@ func (s *composeService) projectFromName(containers Containers, projectName stri
 		}
 		project.Services = append(project.Services, *service)
 	}
-SERVICES:
```
ndeloof, line 136: this is actually checked by `project.GetServices`

pkg/compose/compose.go
```diff
@@ -133,19 +176,41 @@ func (s *composeService) projectFromName(containers Containers, projectName stri
...
-	for _, qs := range services {
-		for _, es := range project.Services {
-			if es.Name == qs {
-				continue SERVICES
-			}
```
ndeloof, line 141: replaced by a sanity check call to `project.GetServices`

PR #9253, opened 2022-03-09
fix generated YAML missing an "examples" section, and update cli-docs-tool to v0.4.0

thaJeztah:
### docs: fix "source" path for YAML generator, and regenerate

Looks like the path was pointing to the project "root", therefore it not looking at the right path 🤔 

### docs: fix trailing whitespace from markdown and regenerate

Trailing whitespace in Markdown can force line-breaks, which doesn't seem to
be the intent on these;

    find . -type f -print0 | xargs -0 perl -pi -e 's/ +$//'

The trailing whitespace also can cause the YAML to go wonky (although the
cli-docs-tool now takes that into account), and caused the "examples" section
to be missed in the `docker compose pull` page (something we should fix in
the tool).

### docs: also generate "usage" in MarkDown files

doesn't affect the docs on docs.docker.com, but makes the pages on GitHub more "complete" for people reading it on GitHub.


ndeloof (approved):

fruityxsamsung (commented):
Nice

fruityxsamsung (commented):
Thanks

PR #9258, opened 2022-03-09
use docker/cli RunExec and RunStart to handle all the interactive/tty/* terminal logic

ndeloof:
**What I did**

dropped all the interactive run/exec logic by leveraging RunExec/RunStart from docker/cli
Using this battle-tested code will make compose way more robust than our attempts to re-implement all corner cases

tested to confirm we don't resurect https://github.com/docker/compose/issues/8908


rumpl (approved):
LGTM after cleanup of unused functions, conflict resolution and formatting :D

Issue #9259, opened 2022-03-09
`docker compose up` no longer recreates container if image is updated


csmith:
**Description**

As of v2.3.0, running `docker compose up -d` will no longer recreate containers whose images have been updated. Previously a `docker compose pull; docker compose up -d` would ensure all containers in the compose file were running the latest images.

**Steps to reproduce the issue:**
1. Create a simple compose file:
```docker-compose
version: '3.5'
services:
  test:
    image: some-registry/image
    restart: always
```
2. Start the container with `docker compose up -d`
3. Push a new image to the registry
4. `docker compose pull`
5. `docker compose up -d`

**Describe the results you received:**

The `pull` step pulls the updated image, but `up -d` does nothing.

**Describe the results you expected:**

The `pull` step pulls the updated image, then `up -d` recreates the container using the new image.

**Additional information you deem important (e.g. issue happens only occasionally):**

Bisect shows b5b8e7a116ccd8a9183c6661589b1c051fca17d9 as the first bad commit. In this example I am running a container with a `version=8` label, but have pushed a new image to the registry with `version=9`:

```
$ git checkout b5b8e7a116ccd8a9183c6661589b1c051fca17d9
Previous HEAD position was 16914e37 Bump compose-go 1.1.0
HEAD is now at b5b8e7a1 exclude com.docker.compose.image label from service hash
$ docker-compose pull; go run ./cmd/ up -d; docker inspect compose-test-1 | grep '"version"'
[+] Running 1/0
 ⠿ test Pulled                                                                                                                                                            0.1s
[+] Running 1/0
 ⠿ Container compose-test-1  Started                                                                                                                                      0.0s
                "version": "8"

$ git checkout b5b8e7a116ccd8a9183c6661589b1c051fca17d9^                                    
Previous HEAD position was b5b8e7a1 exclude com.docker.compose.image label from service hash
HEAD is now at f7a6c3bc Bump github.com/containerd/containerd from 1.5.8 to 1.6.0
$ docker-compose pull; go run ./cmd/ up -d; docker inspect compose-test-1 | grep '"version"'
[+] Running 1/0
 ⠿ test Pulled                                                                                                                                                            0.1s
[+] Running 1/1
 ⠿ Container compose-test-1  Started                                                                                                                                      0.4s
                "version": "9"
```

[...]

PR #9261, opened 2022-03-10
recreate container after image has been rebuilt/pulled

ndeloof:
**What I did**
Fix service being re-created after image is rebuilt (`docker compose up --build`)
This also includes an improvement to set com.docker.compose.image to image ID when image exists in local store, so we can detect image tag been pulled with a refreshed content

**Related issue**
closes https://github.com/docker/compose/issues/9259

### DIFF
--- a/pkg/compose/build.go
+++ b/pkg/compose/build.go
@@ -188,6 +188,15 @@ func (s *composeService) getLocalImagesDigests(ctx context.Context, project *typ
 	for name, info := range imgs {
 		images[name] = info.ID
 	}
+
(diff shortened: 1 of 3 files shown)


rouxn-bsi:
Thank you for the fix and your work!
We hope to see the v2.3.4 soon including this fix ;-)

Just FYI, [Plextrac](https://plextrac.com/) solution is using docker and we had to revert back docker-compose to v2.2.x to get the new instance management script working again. The update procedure was git pulling the updates but no longer recreating the containers to use the updates.

PR #9266, opened 2022-03-10
ps: un-deprecate --filter, and enhance docs

thaJeztah:
Compose currently only supports a single filter options on --filter,
for which reason the --status flag was added, which is more convenient
to use.

However, the `--filter` flag is common among various docker commands, and
it's possible that additional filters get added at some point (which may
be less "commonly" used, and not warrant a dedicated flag).

This PR removes the "deprecated" mention from the flag, to keep consistency
with other commands, but adds documentation to explain how they relate to
eachother.

Also added a short example for the `--format` flag.

**What I did**

**Related issue**

### DIFF
--- a/cmd/compose/ps.go
+++ b/cmd/compose/ps.go
@@ -81,12 +81,11 @@ func psCommand(p *projectOptions, backend api.Service) *cobra.Command {
 	}
 	flags := psCmd.Flags()
 	flags.StringVar(&opts.Format, "format", "pretty", "Format the output. Values: [pretty | json]")
-	flags.StringVar(&opts.Filter, "filter", "", "Filter services by a property. Deprecated, use --status instead")
+	flags.StringVar(&opts.Filter, "filter", "", "Filter services by a property (supported filters: status).")
 	flags.StringArrayVar(&opts.Status, "status", []string{}, "Filter services by status. Values: [paused | restarting | removing | running | dead | created | exited]")
 	flags.BoolVarP(&opts.Quiet, "quiet", "q", false, "Only display IDs")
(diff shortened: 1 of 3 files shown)


glours (approved):
Sounds good to me, thanks @thaJeztah

docs/reference/compose_ps.md
```diff
@@ -8,21 +8,109 @@ List containers
 | Name | Type | Default | Description |
 | --- | --- | --- | --- |
 | `-a`, `--all` |  |  | Show all stopped containers (including those created by the run command) |
-| `--format` | `string` | `pretty` | Format the output. Values: [pretty \| json] |
+| [`--format`](#format) | `string` | `pretty` | Format the output. Values: [pretty \| json] |
```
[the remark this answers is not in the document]
thaJeztah: Yes, that was it; changed that 😅

cmd/compose/ps.go
```diff
@@ -81,12 +81,11 @@ func psCommand(p *projectOptions, backend api.Service) *cobra.Command {
 	}
 	flags := psCmd.Flags()
 	flags.StringVar(&opts.Format, "format", "pretty", "Format the output. Values: [pretty | json]")
-	flags.StringVar(&opts.Filter, "filter", "", "Filter services by a property. Deprecated, use --status instead")
+	flags.StringVar(&opts.Filter, "filter", "", "Filter services by a property (currently supported filters: status).")
```
ndeloof, line 84:
```suggestion
	flags.StringVar(&opts.Filter, "filter", "", "Filter services by a property (supported filters: status).")
```

thaJeztah: fixed!

Issue #8550, opened 2021-02-22
Command-line completion?


sbellem:
I just wonder if there's a command completion like there is for the [docker-compose](https://docs.docker.com/compose/completion/) command.

SuperSandro2000:
There was some completion but it was removed in https://github.com/docker/compose/commit/c28aec2308a24f589814cf4ac55a225d144f557d

ndeloof:
AFAICT offering command line completion for `docker compose` (space) isn't trivial due to the CLI plugin architecture, but we could automatically produce completion for `docker-compose` (when used standalone) using Cobra command support for this feature. Wdyt @ulyssessouza @mat007 ?

mat007:
I was pretty sure this was already in place, but apparently it’s either not fully implemented or not working.
We’ll have a look: internally tracked as https://docker.atlassian.net/browse/IL-768

ndeloof:
IIRC @ulyssessouza experimented with this, but he was blocked by docker CLI plugin architecture (we can't extend the docker CLI completion script)
But for direct use of docker-compose binary, we could adopt https://gist.github.com/jsteenb2/0973c838b00940ee8ccdf958a79756af and also add support for completion functions to auto-complete flags / args values.

danger89:
This is very strange indeed. What happened with the bash auto completion. Otherwise I would have created a PR for upgrading docker-compose to v2.2.2 on https://github.com/docker/docker.github.io/

ndeloof:
Please note the existing `docker-compose` bash completion still works with compose v2 when used as a standalone binary. But it would beed some update to add support for additional commands and flags

stefanocudini:
This problem has been solved? in some version?

stefanocudini:
> Should be easy to generate one, as Cobra command library includes support for generating completion scripts https://github.com/spf13/cobra/blob/master/shell_completions.md
@ndeloof were you able to generate completion scripts? for compose v2?

PHLAK:
> This problem has been solved? in some version?

No, this doesn't seem to be solved yet.

rsteube:
Ok i think i missed something :thinking: : Is `docker compose` invoking `docker-compose` or is it compiled into the `docker` binary?

SuperSandro2000:
`docker compose` aka docker-compose v2 is a plugin for the ``docker`` command written in go and is not using any python stuff from docker-compose v1.

rsteube:
[...]

PR #9269, opened 2022-03-11
Add completion command and fix convert

ulyssessouza:
**What I did**
Add completion command and fix convert

To test, you can trigger the completions with the following example:

1. Create a compose file with the following content:
`compose.yaml`
```
services:
  foo:
    image: nginx
  bar:
    image: nginx
  baz:
    image: nginx
```

2. Then run the following command:
```
$ docker compose __complete up ""
```
This should print the 3 services as proposals

3. By running 
```
$ docker compose __complete up "b"
```
This should propose the 2 services starting with the letter "b"

For now, the `Docker CLI` is not triggering compose's completion yet. That should be done in a following PR.

### DIFF
--- a/cmd/compatibility/convert.go
+++ b/cmd/compatibility/convert.go
@@ -23,6 +23,13 @@ import (
 	"github.com/docker/compose/v2/cmd/compose"
 )
 
+func getCompletionCommands() []string {
+	return []string{
+		"__complete",
+		"__completeNoDesc",
+	}
+}
+
 func getBoolFlags() []string {
 	return []string{
 		"--debug", "-D",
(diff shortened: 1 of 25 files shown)


milas (approved):
LGTM! Can't wait to get it connected to the `docker` CLI 🎉

milas:
Ah project name completion isn't working, it looks like the `backend` instance isn't initialized?

I'm getting `not implemented` error and all the funcs on the `ServiceProxy` are `nil` at the point that `completeProjectNames` is called 😭

milas:
Oh, we lazily initialize the `ServiceProxy` via `PersistentPreRunE` which presumably does NOT run for the auto-completion 😬 

https://github.com/docker/compose/blob/db882416985a2d36ee82f869920ad9ce9d4c2caa/cmd/main.go#L41-L44

ulyssessouza:
@milas Good catch!
I changed to populate the serviceProxy on creation... We could actually get rid of this, but that can be useful on state tracing with interceptors.

milas (approved):
Nice! Project name completion looks good now. Can't wait 🥳

nicksieger:
Should this close #8550 as well?

PR #9274, opened 2022-03-11
Improve issue template

felipecrs:
**What I did**

I improved the issue template. What do you guys think?

**Related issue**


rfay:
It would be better to move to the new form version of the issue template, https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-issue-forms, and it's not hard at all.

ndeloof:
closing as obsolete

Issue #9226, opened 2022-03-03
[proposal] `docker compose config --parameters` to list compose file variables


ndeloof:
Archived project [Docker app](https://github.com/docker/app) offered an interesting feature regarding compose file variable interpolation. To avoid relying on uncontrolled environment variable substitution, it offered a migration path from a compose.yaml file to extract all variables as "parameters"

To address this use case, I suggest we introduce `docker compose config --parameters` to list variables declared in a compose file. In an ideal world, would also show default value and resolved value with source it was read from.

teror4uks:
Do you suggest showing runtime values for variables regarding this [interpolation spec](https://github.com/compose-spec/compose-spec/blob/master/spec.md#interpolation)?

Or re-implement the `parameters` feature from `Docker app` project?

ndeloof:
only show variables.

something like this (assuming this is feasible :P) :
```
$ docker compose config --parameters
NAME       DEFAULT        REQUIRED          ACTUAL         SOURCE
foo            bar                   false                    true                os.Env
zot                                     true                     qix                  .env   
`
``

That being said, we could _also_ add support for `docker compose` to only resolve variables based on some `parameters` file, vs using environment variables or `.env` file; but that's another orthogonal topic

teror4uks:
looks cool, I would like to participate, if you don't mind :)

PR #9279, opened 2022-03-13
Introduction compose config --parameters feature

teror4uks:
Introduced the feature to print variables from config files with additional info like default value, required flag, actual value, source.

example usage:

```
❯ ./docker-compose config --parameters
NAME           DEFAULT    REQUIRED   ACTUAL      SOURCE
LOGGIN_LEVEL   log_info   false      log_info    /home/user/projects/compose/bin/docker-compose.yml
BAR                       true       BAR_VALUE   .env
BUZZ           buzz       false      buzz        /home/user/projects/compose/bin/docker-compose.yml
FOO                       false                  /home/user/projects/compose/bin/docker-compose.yml

```

**Related issue**
[feature proposal](https://github.com/docker/compose/issues/9226)

**Cute animal**

image

### DIFF
--- a/cmd/compose/compose_test.go
+++ b/cmd/compose/compose_test.go
@@ -17,6 +17,7 @@
 package compose
 
 import (
+	"os"
 	"testing"
 
 	"github.com/compose-spec/compose-go/types"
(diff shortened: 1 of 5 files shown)


teror4uks:
@ndeloof hey, could you please help with the review?

cmd/compose/compose_test.go
```diff
@@ -51,3 +52,225 @@ func TestFilterServices(t *testing.T) {
...
+		},
+	}
+	assert.Equal(t, len(params), len(expectedMap))
+	for key, val := range expectedMap {
+		assert.Equal(t, params[key], val)
+	}
+}
+
+func TestPrintParametersWithDotFile(t *testing.T) {
+	dirName, err := os.MkdirTemp("", "docker_compose_dir")
```
laurazard, line 175:
```suggestion
        dirName := t.TempDir()
```

cmd/compose/compose_test.go
```diff
@@ -51,3 +52,225 @@ func TestFilterServices(t *testing.T) {
 	_, err = p.GetService("zot")
 	assert.NilError(t, err)
 }
+
+func TestPrintParametersWithEnvFile(t *testing.T) {
+	dirName, err := os.MkdirTemp("", "docker_compose_dir")
```
laurazard, line 57:
```suggestion
        dirName := t.TempDir()
```

laurazard (commented):
Overall LGTM! Just some small suggestions, but thank you for your contribution 🥳

ndeloof:
unfortunately obsolete, especially with the introduction of multi-values for `--env-file` so we can't track the origin of a value by file anymore. Could probably at least claim source="from env file"

Also, not very pleasant we have to replicate the model parsing logic here with `ParseYAML`, `ExtractVariables` etc. I wish we could configure a listener on compose-go `Load` to get more visibility on how variables get used, so we can build this parameters report

ndeloof:
implemented as docker compose config --variables

PR #9280, opened 2022-03-14
kill only need project name

ndeloof:
**What I did**
Removed need to get the original compose.yaml file to run `docker compose kill`

image


glours (approved):
LGTM and tests ok 👍

Issue #9172, opened 2022-02-14
docker-compose v2 down -v does not respect "external" declaration


MichaelKim0407:
**Description**

`docker-compose down -v` behavior not respecting documentation

**Steps to reproduce the issue:**
1. Create a compose file with an external volume
2. Create volume
3. Run `docker-compose down -v`

**Describe the results you received:**
Volume removed

**Describe the results you expected:**
Volume should not be removed, per documentation
> Networks and volumes defined as `external` are never removed.

**Additional information you deem important (e.g. issue happens only occasionally):**
1. Docker-compose v1 works correctly
2. May be related to #9145

**Output of `docker compose version`:**

```
Docker Compose version v2.2.3
```

**Output of `docker info`:**

```
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc., v0.7.1)
  compose: Docker Compose (Docker Inc., v2.2.3)
  scan: Docker Scan (Docker Inc., v0.16.0)

Server:
 Containers: 0
  Running: 0
  Paused: 0
  Stopped: 0
 Images: 54
 Server Version: 20.10.12
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 2
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 io.containerd.runtime.v1.linux runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 7b11cfaabd73bb80907dd23182b9347b4245eb5d
 runc version: v1.0.2-0-g52b36a2
 init version: de40ad0
 Security Options:
  seccomp
   Profile: default
  cgroupns
 Kernel Version: 5.10.76-linuxkit
 Operating System: Docker Desktop
 OSType: linux
 Architecture: x86_64
 CPUs: 2
 Total Memory: 1.94GiB
 Name: docker-desktop
 ID: VZLG:BCFO:BQRY:LOIT:UAIA:BS6H:EHTB:V3TM:YCLP:PI4B:FZ65:YBY3
 Docker Root Dir: /var/lib/docker
 Debug Mode: true
  File Descriptors: 46
[...]

PR #9281, opened 2022-03-14
don't remove external volumes/networks

ndeloof:
**What I did**
fix `down -v` so we do no remove `external` volumes and networks

**Related issue**
closes https://github.com/docker/compose/issues/9172


glours (approved):
LGTM

PR #9293, opened 2022-03-16
Validate doc on pr

glours:
**What I did**
To be sure we're always in sync between the code and the reference documentation, we added in the Makefile:
- a command to generate the doc from a container
- a command to check if the documentation is up to date or not

image

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -84,10 +84,14 @@ lint: ## run linter(s)
 
 .PHONY: docs
 docs: ## generate documentation
-	@docker build . \
-	--output type=local,dest=./docs/ \
+	$(eval $@_TMP_OUT := $(shell mktemp -d -t dockercli-output.XXXXXXXXXX))
+	docker build . \
+	--output type=local,dest=$($@_TMP_OUT) \
 	-f ./docs/docs.Dockerfile \
 	--target update
+	rm -rf ./docs/internal
+	cp -R "$($@_TMP_OUT)"/out/* ./docs/
+	rm -rf "$($@_TMP_OUT)"/*
 
 .PHONY: validate-docs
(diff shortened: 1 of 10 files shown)


docs/reference/docker_compose_run.yaml
```diff
@@ -124,7 +124,7 @@ options:
 - option: no-TTY
   shorthand: T
   value_type: bool
-  default_value: "false"
+  default_value: "true"
```
thaJeztah, line 127: This looks odd; is this because we're detecting if we're running with a TTY attached? Perhaps there should be no visible default (or the description needs to be updated to reflect that we're automatically detecting)

ndeloof: +1 for documenting this in description

glours: Doc updated, thank you folks

Makefile
```diff
@@ -82,6 +82,19 @@ lint: ## run linter(s)
 	--build-arg GIT_TAG=$(GIT_TAG) \
 	--target lint
 
+.PHONY: docs
+docs: ## generate documentation
+	@docker build . \
+	--output type=local,dest=./docs/ \
```
crazy-max, line 88: `output` will not rsync so if some files are not their anymore they will still be in working tree. I suggest to do smth like https://github.com/docker/cli/blob/11c76f1ddbb64e4d88b44b48c017b81ffe86e0e5/docker.Makefile#L84-L88

glours: done, thanks @crazy-max 🙏

crazy-max (approved):
LGTM

ndeloof (approved):

PR #9311, opened 2022-03-24
Don't use ansi even when deprecated option is requested

quite:
Signed-off-by: Daniel Lublin <daniel@lublin.se>

**I thought it insolent of docker-compose to emit ansi characters in the warning message, when it does in fact otherwise detect and honour --no-ansi, even though the option is deprecated**

**capybara_emoji**


milas (approved):
Thanks for the improvement!

PR #9317, opened 2022-03-25
go.mod: pin k8s version to fix build on go1.16, and update buildx to v0.8.1

thaJeztah:
- relates to https://github.com/docker/docker-ce-packaging/pull/652
- relates to https://github.com/docker/buildx/pull/948

### go.mod: pin k8s dependencies to the same version as in buildx

Buildx uses replace rules to pin these dependencies to the version
that's used in buildkit. Newer versions of these dependencies are
incompatible with some go versions, and we do not need the newer
versions for where it's used.

### go.mod: github.com/docker/buildx v0.8.1

Fix possible panic on handling build context scanning errors

**What I did**

**Related issue**


go.sum
```diff
@@ -2281,8 +2231,6 @@ rsc.io/sampler v1.3.0/go.mod h1:T1hPZKmBbMNahiBKFy5HrXp6adAjACjK9JXDnKaTXpA=
 sigs.k8s.io/apiserver-network-proxy/konnectivity-client v0.0.14/go.mod h1:LEScyzhFmoF5pso/YSeBstl57mOzx9xlU9n85RGrDQg=
 sigs.k8s.io/apiserver-network-proxy/konnectivity-client v0.0.15/go.mod h1:LEScyzhFmoF5pso/YSeBstl57mOzx9xlU9n85RGrDQg=
 sigs.k8s.io/apiserver-network-proxy/konnectivity-client v0.0.22/go.mod h1:LEScyzhFmoF5pso/YSeBstl57mOzx9xlU9n85RGrDQg=
-sigs.k8s.io/json v0.0.0-20211020170558-c049b76a60c6 h1:fD1pz4yfdADVNfFmcP2aBEtudwUQ1AlLnRBALr33v3s=
-sigs.k8s.io/json v0.0.0-20211020170558-c049b76a60c6/go.mod h1:p4QtZmO4uMYipTQNzagwnNoseA6OxSUutVw05NhYDRs=
```
thaJeztah, line 2285: This is the dependency that broke the build; it's a fork of Golang's `json` package, but "case-sensitive", and it's depending on some features that are not available in go1.16 (which is still in use in the docker 20.10 release pipeline)

thaJeztah:
@ndeloof @tonistiigi @crazy-max PTAL

crazy-max (approved):
LGTM

ndeloof (approved):
LGTM.
We should adopt same approach proposed on https://github.com/docker/buildx/pull/840 so that we don't get such side effect as packaging tries to build from source with a distinct runtime

Issue #9315, opened 2022-03-25
terminal handling on exec is completely unusable in 2.3.4


rfay:
**Description**

In 2.3.4, you can't `docker-compose exec -it` into a container... your terminal gets captured and you can't do anything.

**Steps to reproduce the issue:**

1. Use this docker-compose.yaml:
```yaml
services:
  deb:
    image: debian:bullseye
    command: tail -f /dev/null
```
2. `docker-compose up -d`
3. `docker-compose exec -it deb bash`

You'll get the initial prompt, but you can't interact with bash at all. It hangs. And you have to kill the terminal emulator to continue.

**Describe the results you expected:**

This has worked for years, of course, and was fine in 2.3.3.

**Output of `docker compose version`:**

Docker Compose version v2.3.4

**Output of `docker info`:**

```
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc., v0.8.1)
  compose: Docker Compose (Docker Inc., v2.3.3)
  scan: Docker Scan (Docker Inc., v0.17.0)

Server:
 Containers: 6
  Running: 1
  Paused: 0
  Stopped: 5
 Images: 10
 Server Version: 20.10.13
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 2
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 io.containerd.runtime.v1.linux runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 2a1d4dbdb2a1030dc5b01e96fb110a9d9f150ecc
 runc version: v1.0.3-0-gf46b6ba
 init version: de40ad0
 Security Options:
  seccomp
   Profile: default
  cgroupns
 Kernel Version: 5.10.104-linuxkit
 Operating System: Docker Desktop
 OSType: linux
 Architecture: aarch64
 CPUs: 4
 Total Memory: 2.922GiB
 Name: docker-desktop
 ID: TUH5:TMNE:DVL2:J3WB:RCCS:YIPF:PO3Z:DODW:C3QF:NELR:TSNC:TTZU
[...]

PR #9318, opened 2022-03-25
pass interactive flag to ExecOptions for the RunExec command

glours:
**What I did**
Pass the `interactive` flag from the Compose CLI to the Docker one to run exec command

**Related issue**
#9315 

image


ndeloof (approved):

rfay:
I don't understand why a critical feature and failure like this didn't get a test added.

Issue #9232, opened 2022-03-04
docker-compose build results in repository pull errors for local images


TrentonAdams:
**docker-compose build results in repository pull errors for local images**

Sample broken project at: [https://github.com/TrentonAdams/compose-bug](https://github.com/TrentonAdams/compose-bug)

According to the [docker compose documentation](https://docs.docker.com/compose/compose-file/compose-file-v3/)

> If you specify image as well as build, then Compose names the built image with the webapp and optional tag specified in image:

It used to do just that, but now it just breaks.

Running this just works...
```bash
./build.sh
```

Running this fails with repository pull errors...

```bash
docker-compose build
```

The project this sample project is based off of has been virtually untouched for a couple of years, but now it's not working the way that it used to.  Docker compose seems to have a regression.

**Steps to reproduce the issue:**
1. docker-compose build

**Describe the results you received:**

```
------
 > [compose-play_backend internal] load metadata for docker.io/library/prj-base:latest:
------
------
 > [compose-play_frontend internal] load metadata for docker.io/library/prj-build:latest:
------
failed to solve: rpc error: code = Unknown desc = failed to solve with frontend dockerfile.v0: failed to create LLB definition: pull access denied, repository does not exist or may require authorization: server message: insufficient_scope: authorization failed
```

**Describe the results you expected:**

Completion of the build

**Output of `docker compose version`:**

```
Docker Compose version v2.2.3
```

**Output of `docker info`:**

```
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc., v0.7.1)
  compose: Docker Compose (Docker Inc., v2.2.3)
  scan: Docker Scan (Docker Inc., v0.17.0)

Server:
 Containers: 3
  Running: 1
  Paused: 0
  Stopped: 2
 Images: 241
 Server Version: 20.10.12
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 2
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 io.containerd.runtime.v1.linux runc
 Default Runtime: runc
 Init Binary: docker-init
[...]

Issue #8538, opened 2021-06-06
Referencing local image name fails on "docker compose build"


neilyio:
**Description**

I'm getting an error from `docker compose build` with a setup that is consistently successful with `docker-compose build`.

I have a `docker-compose.yml` file with two services: `base` and `extended`. `base.image` gives a name to the image built by `base`, and I'd like to use that name as the `FROM` image in the Dockerfile for the `extended` service.

This works well with `docker-compose build`. It does not work with `docker compose build`.

**Steps to reproduce the issue:**
1. Re-create my `docker/` folder with these three files, plus an empty `README.md`. `cd` into `docker/`.
```yml
# docker/docker-compose.yml
services:
  base:
    image: neilyio/base
    build:
      context: .
      dockerfile: base.Dockerfile

  extended:
    build:
      context: .
      dockerfile: extended.Dockerfile
```
```dockerfile
# docker/base.Dockerfile
FROM scratch
COPY README.md /root/README.md
```
```dockerfile
# docker/extended.Dockerfile
FROM neilyio/base
CMD cat /root/README.md
```
2. `docker-compose build`, expect a successful run.
3. Clear your cache and delete these new images so we have a clean comparison for the next step. I used these commands:
```
docker image rm neilyio/base docker_extended
docker system prune -f
```
4. `docker compose build`, expect a failure.

**Describe the results you received:**

`docker compose build` produces:
```
[+] Building 0.6s (8/8) FINISHED                                                                                                                                                                            
 => [docker_extended internal] load build definition from extended.Dockerfile                                                                                                                          0.0s
 => => transferring dockerfile: 88B                                                                                                                                                                    0.0s
 => [neilyio/base internal] load build definition from base.Dockerfile                                                                                                                                 0.0s
 => => transferring dockerfile: 86B                                                                                                                                                                    0.0s
 => [docker_extended internal] load .dockerignore                                                                                                                                                      0.0s
 => => transferring context: 2B                                                                                                                                                                        0.0s
 => [neilyio/base internal] load .dockerignore                                                                                                                                                         0.0s
 => => transferring context: 2B                                                                                                                                                                        0.0s
 => ERROR [docker_extended internal] load metadata for docker.io/neilyio/base:latest                                                                                                                   0.4s
 => [neilyio/base internal] load build context                                                                                                                                                         0.0s
 => => transferring context: 3.65kB                                                                                                                                                                    0.0s
[...]

PR #9324, opened 2022-03-30
use NamedContext so buildkit is aware of cross service build dependencies

ndeloof:
**What I did**
Used `NamedContexts` so buildkit is aware of base image for a service being the actual result of another service build.

**Related issue**
closes https://github.com/docker/compose/issues/9232
closes https://github.com/docker/compose/issues/8538


ndeloof:
closing as obsolete

PR #9325, opened 2022-03-30
Add support of ssh authentications defined in compose file or via cli…

### LINKED ISSUE #7025: Enable ssh option as build parameter for buildkit

glours:
… flags

**What I did**

- Add support of an `ssh` flag for the `build` command
- Generate an SSH Agent Provider from both cli flag and Compose file configuration

**Related issue**
fix #7025

image


ndeloof (approved):

ndeloof (approved):

ndeloof:
🎉

levrik:
Is there any expected timeline when this will hit Docker on Linux and macOS?

aleksandrs-ledovskis:
It's possible to already use it on macOS and Linux if compiled from repo's source.

For those using Homebrew, the command is `$ brew install --head docker-compose`.

docs/reference/compose_build.md
```diff
@@ -12,6 +12,7 @@ Build or rebuild services
 | `--progress` | `string` | `auto` | Set type of progress output (auto, tty, plain, quiet) |
 | `--pull` |  |  | Always attempt to pull a newer version of the image. |
 | `-q`, `--quiet` |  |  | Don't print anything to STDOUT |
+| `--ssh` | `string` |  | Set SSH authentications used when building service images. (use 'default' for using you default SSH Agent) |
```
aleksandrs-ledovskis, line 15: `s/using you/using your/`

aleksandrs-ledovskis: Hivemind with https://github.com/docker/compose/pull/9334 🙂

nleroy917:
Hello,

I compiled from source (`brew install --head docker-compose`), but the `--ssh` flag still seems unknown:

```sh
docker-compose up --build --ssh default 
unknown flag: --ssh
```

Version seems up to date:

```
docker-compose -v                      
Docker Compose version HEAD-7969667
```

Any help would be appreciated!

glours:
@nleroy917 `--ssh` flag can be use with `docker compose build` but is not available with `docker compose up` command

nleroy917:
🤦🏻‍♂️ That makes sense. Thank you!

IT-Ideas:
Hello all!
If I am not mistaken this only take one pair key/value as argument of --ssh, right?
We are in a situation where we would have like to pass two different ssh keys.
Would there be any plans to manage this use case?

Thanks!

ndeloof:
@IT-Ideas ssh configuration should be set by your `compose.yaml` file inside the `build` section (https://docs.docker.com/compose/compose-file/build/#ssh). `--ssh` flag is just a helper for simplest use-cases, imho we should not expose such flags which are redundant with the yaml model

IT-Ideas:
Hi @ndeloof,
I agree with you and I actually tested it, but the path to the key was then not working although it worked when passing it through the command.
The path beeing ~/.ssh/keyname.

Thanks 🙂

IT-Ideas:
@ndeloof  Maybe I did a mistake, I'll have one more try!

ndeloof:
please open a dedicated issue if you can reproduce this error and attach a sample compose.yaml file, thanks

PR #9342, opened 2022-04-01
now we use directly the Docker CLI to run autoremove flag should be p…

### LINKED ISSUE #9314: docker compose run --rm no longer working on 2.3.x
**Description**

The `--rm` flag of `docker compose run` is not working with version 2.3.x
The container is created but then it remains in exit status. Up to v2.2.3 it was correctly removed

1. docker-compose.yml
```
version: "3.9"
services:
  alpine:
    image: alpine
```
2. `docker compose run --rm --name mytest alpine hostname`

The container remains in exit status

```
$ docker ps -a
CONTAINER ID   IMAGE     COMMAND      CREATED         STATUS                     PORTS     NAMES
b139d98c63f3   alpine    "hostname"   5 seconds ago   Exited (0) 3 seconds ago             mytest
```

I expect the container to be removed, like it was happening up to v2.2.3

```
Docker Compose version v2.3.4

```

**Output of `docker info`:**

```
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc., v0.8.1)
  compose: Docker Compose (Docker Inc., v2.3.4)

Server:
 Containers: 1
  Running: 0
  Paused: 0
  Stopped: 1
 Images: 32
 Server Version: 20.10.13
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 1
 Plugins:
  Volume: local
 Swarm: inactive
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 2a1d4dbdb2a1030dc5b01e96fb110a9d9f150ecc
 runc version: v1.0.3-0-gf46b6ba
 init version: de40ad0
 Security Options:
  apparmor
  seccomp
   Profile: default
 Kernel Version: 5.13.0-35-generic
 Operating System: Ubuntu 20.04.4 LTS
 OSType: linux
 Architecture: x86_64
 CPUs: 8
 Total Memory: 15.53GiB
 Name: xxxxxxxxxxxxxxxxx
 ID: CCLO:O7YG:FX77:W6HX:WRYV:6D6Z:4QPY:MUPX:HCK4:DMYR:LKSZ:5XLJ
 Debug Mode: false
 Labels:
 Experimental: false
 Insecure Registries:
  127.0.0.0/8

```

glours:
…ass as is

**What I did**
Pass the `rm` flag value as is to the Docker CLI when running a container with this flag

fix #9314 

image

### DIFF
--- a/pkg/compose/run.go
+++ b/pkg/compose/run.go
@@ -89,7 +89,7 @@ func (s *composeService) prepareRun(ctx context.Context, project *types.Project,
 	updateServices(&service, observedState)
 
 	created, err := s.createContainer(ctx, project, service, service.ContainerName, 1,
-		opts.Detach && opts.AutoRemove, opts.UseNetworkAliases, opts.Interactive)
+		opts.AutoRemove, opts.UseNetworkAliases, opts.Interactive)
 	if err != nil {
 		return "", err
 	}


pkg/compose/run.go
```diff
@@ -89,7 +89,7 @@ func (s *composeService) prepareRun(ctx context.Context, project *types.Project,
 	updateServices(&service, observedState)
 
 	created, err := s.createContainer(ctx, project, service, service.ContainerName, 1,
-		opts.Detach && opts.AutoRemove, opts.UseNetworkAliases, opts.Interactive)
```
thaJeztah, line 92: Was this because there was some code on the compose side to handle this case? (that has been removed?) 

Just double-checking if there's not also code on compose's side, causing a potential race on "who gets to remove it first" 😂

glours: Yep it was, previously the remove part was done by Compose, [see here](https://github.com/docker/compose/commit/1d4b4e3c8ebd60c2d3f71cb8f26801fee43b8b3f)

```
func (s *composeService) terminateRun(ctx context.Context, containerID string, opts api.RunOptions) (exitCode int, err error) {
	exitCh, errCh := s.apiClient().ContainerWait(ctx, containerID, container.WaitConditionNotRunning)
	select {
	case exit := <-exitCh:
		exitCode = int(exit.StatusCode)
	case err = <-errCh:
		return
	}
	if opts.AutoRemove {
		err = s.apiClient().ContainerRemove(ctx, containerID, moby.ContainerRemoveOptions{})
	}
	return
}
```

thaJeztah: Ah, great! Thanks for confirming 👍

thaJeztah (approved):
LGTM

ndeloof (approved):

Issue #9288, opened 2022-03-15
`tty: true` in docker-compose.yml no longer displays console output


khalwat:

tanzislam:
This behavior started with Docker Compose v2.3.0. v2.2.3 was fine. I've verified this behavior change in Linux.

nroi:
I can also confirm this issue on Linux. I ran git bisect, and it seems to be introduced by 65ed8cf4c29925c387d51906ebd260627d3bd20b.

glenpike:
This is a frustrating fix.  We use tty: true so we can attach to containers for interactive debugging, e.g. byebug in Ruby, but most of the time we also want to see container output without having to run a separate terminal window to follow the logs...

yoshiya0503:
hey guys, there are unwilling workaround...

```
1 set tty: true
2 docker compose up -d
3 docker compose logs -f
```

we can see logs (colorlized) same as before.

(BE CAREFUL TO CONFLICT CONTAINER PORT LIKE MYSQL, REDIS)

aviau:
On my end it seems that it causes more than just missing logs. I am building something with webpack in tty mode (so that I have colored logs) and it seems that webpack hangs when the build is completed and never starts the dev server.

chloerei:
esbuild also needs to set `tty: true`, otherwise it will exit watch mode. But in version 4.6.0 setting `tty: true` it doesn't show the logs of esbuild. Now I fallback to version 4.5.0.

ndeloof:
Can you please confirm this issue still applies to v2.4.0?
Main reason for me to ask is that, with https://github.com/docker/compose/pull/9258, compose now uses the same codebase as `docker run` command and should then offer the same user experience regarding tty management and ability to attach to container to collect console output.

alexandernst:
@ndeloof the issue doesn't seem to be fixed with 2.4.0 (tested on a MBP M1)

alexandernst:
Yes, 2.4.1 fixes this issue.

dedene:
I can also confirm, 2.4.1 fixes the issue! 🙌 Thanks for the fix!

aviau:
Not fixed for me on 2.4.1 :(

ell1e:
I seem to still have this problem with docker-compose v2.29.7, although perhaps I'm doing something wrong:

```
$ cat Dockerfile.build-x64 

FROM alpine:latest

[...]

PR #9348, opened 2022-04-04
get Tty from container to know adequate way to attach to

ndeloof:
**What I did**
inspect container so we know if we need to attach with a TTY

**Related issue**
closes #[9288](https://github.com/docker/compose/issues/9288)


glours (approved):
LGTM

PR #9362, opened 2022-04-07
Fix volumes_from cannot work with container: prefix

qnap-ericfan:
**What I did**
Here is my docker-compose-yaml
```
services:
  test:
    image: nginx
    volumes_from:
      - container:ct-1
```
I execute command `docker-compose up`, then I got error `no such service: ct-1`.
But actually container `ct-1` exists.
Because `getVolumesFrom` func return `dependServices` and store it on `DependsOn`, maybe don't need to double check on `prepareServicesDependsOn` func.

**Related issue**
n/a


glours:
replace by #9641 
`volumes_from` can also contains reference to services, so we can't remove the whole loop on the `volumes_from` slice

Issue #9301, opened 2022-03-21
Docker Compose 2.3.3 does not start subsidiary services called out in `links`


mshappe:
**Description**

`docker-compose` v2.3.3 appears not to fully implement the dependency-starting behaviour of `links`.

```
services:
  foo:
    links:
      - bar
 ```

will **create** `bar` but not start it.

```
services:
  foo:
    depends_on:
      - bar
    links:
      - bar
```

will ensure `bar gets started, but should be superfluous--`links` is explicitly documented to imply `depends_on`.

**Steps to reproduce the issue:**

See snippets above -- create a `docker-compose.yml` that uses `links` and not `depends_on`.

**Describe the results you received:**

Linked services that should be started are merely `created`.

**Describe the results you expected:**

Linked services are created and started, as prerequisites of the linking service, as if they had been called out in `depends_on`.

**Additional information you deem important (e.g. issue happens only occasionally):**

This behaviour was witnessed both on Arch Linux with `docker-compose` 2.3.3, and on macOS Docker Desktop 4.6.0

**Output of `docker compose version`:**

```
Docker Compose version 2.3.3
```

**Output of `docker info`:**

```
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc., v0.8.0-docker)
  compose: Docker Compose (Docker Inc., 2.3.3)

Server:
 Containers: 15
  Running: 5
  Paused: 0
  Stopped: 10
 Images: 61
 Server Version: 20.10.13
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: false
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runtime.v1.linux runc io.containerd.runc.v2
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 10f428dac7cec44c864e1b830a4623af27a9fc70.m
 runc version: v1.1.0-0-g067aaf85
 init version: de40ad0
 Security Options:
  seccomp
   Profile: default
  cgroupns
 Kernel Version: 5.16.15-arch1-1
 Operating System: Arch Linux
 OSType: linux
 Architecture: x86_64
[...]

PR #9364, opened 2022-04-07
Add Links in the labels for project reconstruction

ulyssessouza:
**What I did**
Add Links in the labels for project reconstruction

**Related issue**
Resolves https://github.com/docker/compose/issues/9301


ndeloof:
Have you checked `dependencies` label is already set to include linked services ?
also see https://github.com/docker/compose/pull/9365 which probably fixes the reported issue

ulyssessouza:
@ndeloof `dependencies` label does not include the `links`. https://github.com/docker/compose/pull/9365 Also fixes the issue.
But that makes me think about the project that is reconstructed with this missing part... But looks like the commands we reconstruct the project from its containers wouldn't use the `links`.

ndeloof:
Created https://github.com/docker/compose/pull/9368 for `links` not to be considered for dependencies labelling

PR #9368, opened 2022-04-08
include services declared by `links` as implicit dependencies

ndeloof:
**What I did**
include services declared by `links` as implicit dependencies

also fix linter error, as build.Compress never returns a nil interface value

**Related issue**
see https://github.com/docker/compose/issues/9301

image
(Hanneton "Melolontha Melolontha")


glours (approved):
LTGTM

Issue #9286, opened 2022-03-15
Services with profile in yml, always enable when execute command start, stop, restart, up


ani3llyon:
Since docker compose v 2.3.0 when ever I run docker compose with command `start`, `stop`, `restart`, and `up` all container for service with profile in docker-compose.yml will be always included even the profile is not enabled.

example command
```
# will create container for service without profile but will start all container
# even the container is for service with profile
docker compose up -d

# will start/stop/restart all container even the container is for service with profile
docker compose start
docker compose stop
docker compose restart
```

example docker-compose.yml
```
version: "3.9"
services:
  serviceA:
    ...
  serviceB:
    ...
    profiles: ["debug"]
```

step by step
```
# will create and start container for service A
docker compose up -d

# will start/stop/restart container for service A
docker compose start
docker compose stop
docker compose restart

# will stop and remove container for service A
docker compose down

# will create and start container for service B
docker compose up -d serviceB

# will stop container for service B
docker compose stop serviceB

# will create and start container for service A
# also start serviceB <-- why? profile 'debug' is not enabled
docker compose up -d

# will start/stop/restart container for service A
# also start/stop/restart serviceB <-- why? profile 'debug' is not enabled
docker compose start
docker compose stop
docker compose restart
```

JimDabell:
I’m experiencing this too. It started happening immediately after upgrading to Docker for Mac 3.6.0 which brought with it the new version of Compose.

**Output of `docker compose version`:**

```
Docker Compose version v2.3.3
```

**Output of `docker info`:**

```
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc., v0.8.0)
  compose: Docker Compose (Docker Inc., v2.3.3)
  scan: Docker Scan (Docker Inc., v0.17.0)

Server:
 Containers: 41
  Running: 40
  Paused: 0
  Stopped: 1
 Images: 34
 Server Version: 20.10.13
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 2
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
[...]

PR #9375, opened 2022-04-11
if command is ran with a compose file, apply the compose model

ndeloof:
**What I did**
pass project to start/restart/pause/unpause if user ran command with a compose.yaml file, not just project-name
backend will then apply model (active profiles) and not consider all resources with project label

**Related issue**
see https://github.com/docker/compose/issues/9286

Fixes https://github.com/docker/compose/issues/9705
Fixes https://github.com/docker/compose/issues/9671

image


laurazard:
I fixed the merge conflicts and added some of the tests I'd already written for the other PR, this one covers the other use case and applies the same logic to other commands 🥳

milas (approved):
🥳

laurazard:
Ahh! Need to fix some tests 😅

laurazard:
While looking into why this was failing when applying a similar logic to `compose ps` I noticed some inconsistencies regarding how the `-p` flag is handled vs the `COMPOSE_PROJECT_NAME` env var.

The reason why some tests were failing was that we are running `compose ps` with the `COMPOSE_PROJECT_NAME` env var set. Using `projectOptions.toProjectName(...)`, this worked because before attempting to load the project,  it checks for the env var and doesn't attempt to load the project otherwise. However, the introduction of `projectOptions.projectOrName(...)` causes some issues here because it doesn't attempt to check the env var, and instead always loads the project unless the passed `projectOptions` struct already has a name set. The inconsistencies here are due to the fact that the `projectOptions.ProjectName` field is set when running `compose -p [project-name]`, but not when running `COMPOSE_PROJECT_NAME=[project-name] compose`.

I think we should handle these in the same way, probably in the parent `compose` command, and set that value regardless of where the project name comes from. This would also make some other things more consistent, ie `compose -p [name] down` works even if there is no `compose.yaml` in the directory but `COMPOSE_PROJECT_NAME=[name] compose down` fails.

I made some changes to `projectOrName()` in this PR to fallback to the previous behavior so the tests can pass, but I think some of this logic should be cleaned up and moved up for more consistency.

ndeloof:
> compose -p [name] down works even if there is no compose.yaml in the directory but COMPOSE_PROJECT_NAME=[name] compose down fails

Why would you like the latter to fail? IMHO once `COMPOSE_PROJECT_NAME` is set, all compose command should be able to run based on existing resources, same as passing `-p` flag. But maybe I'm missing some corner cases...

ndeloof:
LGTM with @laurazard additions

pkg/e2e/fixtures/start-stop/other.yaml
```diff
@@ -0,0 +1,5 @@
+services:
+  a-different-one:
+    image:  nginx:alpine
+  and-another-one:
+    image:  nginx:alpine
```
glours, line 5: nit: add a new line

glours (approved):
LGTM 👍

pkg/e2e/fixtures/start-stop/other.yaml
laurazard: done :)

Issue #3660, opened 2016-06-29
Docker compose pull doesn't respect local images


antony:
During my searches I've seen this issue raised and closed a few times, but even using the latest build of docker-compose, this still doesn't behave the way I understand that it is expected to.

Scenario:
We're building a composed environment for our CI tests to run. We grab a database image, an api image, and a _local_ copy of the image containing application we are trying to test (it has been built in a previous CI step, and tagged with the build hash provided by our CI environment (CIRCLE_SHA1)

docker-compose.yml

``` yaml
api:
  image: mycompany/api

undertest:
  image: mycompany/undertest:${CIRCLE_SHA1}

db:
  image: mycompany/db
```

The commands we run then are as follows:

``` bash
docker-compose pull # Because we definitely want the latest db, and api for a true test.
docker-compose up # compose the environment
```

Actual Result:
No matter what I do, docker compose always tries to pull my CIRCLE_SHA1 tagged version from  docker hub. it doesn't exist there, I never want to push it (until it passes tests and is re-tagged as :latest and/or :release 

I have a unique tag CIRCLE_SHA1 which only exists inside the build environment, meaning no confusion for docker-compose when it tries to pull, and yet, it seems to try to fetch it anyway, and fail even though that exact tag exists locally.

Expected Result:
I'd expect the fact that there is no remote build tagged with CIRCLE_SHA1 to cause docker-compose to use the local copy it finds. I need to do pull, because I want everything else to be the latest.

I'd suggest that if there is confusion where `image:` refers to a remote repository, then perhaps we could use `local:` instead, to reference a local image?

decoomanj:
I'm facing similar issues. I use the --ignore-pull-failures option when I do a docker-compose pull. This basically ignores ALL errors. It is ok as a workaround. We do the same thing in our pipeline. I pull other production-images and start my freshly build release-candidate container locally. When the tests are ok, it's uploaded.

johnharris85:
I don't see this is a issue / bug with current functionality. `pull` is doing exactly that, how would it know in your original compose file that 2 images should be pulled remotely but one should be grabbed locally (seeing as they all probably exist locally right, you just want to 'refresh' the other two?).

Perhaps a --filter option, so something like:

    docker-compose.yml
    version: '2'
    services:
      api:
        image: mycompany/api
        labels:
          - "refresh"
      undertest:
        image: mycompany/undertest:${CIRCLE_SHA1}
      db:
      image: mycompany/db
      labels:
        - "refresh"

[...]

PR #9376, opened 2022-04-11
`pull` to respect pull_policy

ndeloof:
**What I did**
`pull` command to skip services configured as `pull_policy: never` and to ignore those with an existing image and `pull_policy: missing`

**Related issue**
fixes https://github.com/docker/compose/issues/3660


glours (approved):
LGTM

Issue #9378, opened 2022-04-11
Regression on `COMPOSE_PROJECT_NAME` since 2.4.0 (available on 2.4.1)


xunleii:
**Description**

Compose 2.4.0/2.4.1 does not allow uppercase characters. This is visible when we create a new project using `COMPOSE_PROJECT_NAME` with uppercase, the name visible with `docker-compose ls` will be lowercase, and no other commands will be used except `docker-compose down` (I can't explain why we have `strings.ToLower()` only on [this command](https://github.com/docker/compose/blob/v2/pkg/compose/down.go#L39)).

**Steps to reproduce the issue:**
1. `COMPOSE_PROJECT_NAME=AAAA-bbbb docker-compose up -d`
2. `COMPOSE_PROJECT_NAME=AAAA-bbbb docker-compose ps`
3. `COMPOSE_PROJECT_NAME=aaaa-bbbb docker-compose ps`

**Describe the results you received:**

Nothing displayed on step `2.` but yes in step `3.`

**Describe the results you expected:**

The step `2.` should return the container list but not `3.`

**Additional information you deem important (e.g. issue happens only occasionally):**

This issue is not available on [v2.3.4](https://github.com/docker/compose/releases/tag/v2.3.4)

**Output of `docker compose version`:**

```
Docker Compose version v2.4.1
```

**Output of `docker info`:**

```
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc., v0.8.2)
  compose: Docker Compose (Docker Inc., v2.4.1)
  sbom: View the packaged-based Software Bill Of Materials (SBOM) for an image (Anchore Inc., 0.6.0)
  scan: Docker Scan (Docker Inc., v0.17.0)

Server:
 Containers: 2
  Running: 2
  Paused: 0
  Stopped: 0
 Images: 3
 Server Version: 20.10.14
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 1
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 io.containerd.runtime.v1.linux runc
 Default Runtime: runc
[...]

PR #9385, opened 2022-04-13
project name MUST be lowercase

ndeloof:
**What I did**
enforce projectName to be lowercase
renamed API argument names to be homogeneous

**Related issue**
fixes https://github.com/docker/compose/issues/9378

image


glours (approved):
LGTM

shlomiken:
This is major breaking change - how can you push that the stream without notifying first.
we basically need to migrate hundreds of compose/gradle files now

PR #9422, opened 2022-04-28
Fix relative paths on envfile label

ulyssessouza:
**What I did**
This harmonizes the absolute path of project's working dir with the possible relative path of the env-file

**Related issue**
Resolves https://github.com/docker/for-mac/issues/6229

### DIFF
--- a/cmd/compose/compose.go
+++ b/cmd/compose/compose.go
@@ -169,7 +169,10 @@ func (o *projectOptions) toProject(services []string, po ...cli.ProjectOptionsFn
 
 	ef := o.EnvFile
 	if ef != "" && !filepath.IsAbs(ef) {
-		ef = filepath.Join(project.WorkingDir, o.EnvFile)
+		ef, err = filepath.Abs(ef)
+		if err != nil {
+			return nil, err
+		}
 	}
 	for i, s := range project.Services {
 		s.CustomLabels = map[string]string{


cmd/compose/compose.go
```diff
@@ -169,7 +169,12 @@ func (o *projectOptions) toProject(services []string, po ...cli.ProjectOptionsFn
 
 	ef := o.EnvFile
 	if ef != "" && !filepath.IsAbs(ef) {
-		ef = filepath.Join(project.WorkingDir, o.EnvFile)
```
ndeloof, line 172: I'm not sure I understand the intent here. afaict `filepath.Join(project.WorkingDir, o.EnvFile)` should already make this path absolute, but maybe I missed something?
could you please clarify this in your PR description as the Docker Desktop issue is not public

ulyssessouza: In the example where:
- `/some/path` is where we execute the command
- `/some/path/wrkdir` is the path passed as `--project-directory`
- `./wrkdir/envs/my.env` as the path passed in `--env-file`

in this we would end with `/some/path/wrkdir` + `./wrkdir/envs/my.env` which would repeat `wrkdir` resulting in `/some/path/wrkdir/wrkdir/envs/my.env` which is wrong.

ndeloof: ah ok, because we resolve envfile based on project dir (which is definitely a bug) and we should resolve relative to command working dir. Then `filepath.Abs` is relevant. Earlier version of the diff I reviewed was doing some voodoo magic with `WorkingDir` so my confusion

ndeloof (approved):

PR #9437, opened 2022-05-03
cp command: copy to all containers of a service as default behaviour

glours:
**What I did**
Change the default behaviour of the `cp` command when copying from host to a service. Now it will copy the source to all the containers of a service instead of the first one previously

Deprecation of the `--all` flag

image

### DIFF
--- a/cmd/compose/cp.go
+++ b/cmd/compose/cp.go
@@ -64,7 +64,7 @@ func copyCommand(p *projectOptions, backend api.Service) *cobra.Command {
 	}
 
 	flags := copyCmd.Flags()
-	flags.IntVar(&opts.index, "index", 0, "Index of the container if there are multiple instances of a service [default: 1 when copying from a container].")
+	flags.IntVar(&opts.index, "index", 0, "Index of the container if there are multiple instances of a service .")
 	flags.BoolVar(&opts.all, "all", false, "Copy to all the containers of the service.")
 	flags.MarkHidden("all")                                                                                                      //nolint:errcheck
 	flags.MarkDeprecated("all", "By default all the containers of the service will get the source file/directory to be copied.") //nolint:errcheck
--- a/cmd/compose/cp.go
+++ b/cmd/compose/cp.go
@@ -55,7 +55,7 @@ func copyCommand(p *projectOptions, backend api.Service) *cobra.Command {
 			}
 			return nil
 		}),
-		RunE: Adapt(func(ctx context.Context, args []string) error {
+		RunE: AdaptCmd(func(ctx context.Context, cmd *cobra.Command, args []string) error {
 			opts.source = args[0]
 			opts.destination = args[1]
 			return runCopy(ctx, backend, opts)
@@ -64,8 +64,10 @@ func copyCommand(p *projectOptions, backend api.Service) *cobra.Command {
 	}
 
 	flags := copyCmd.Flags()
-	flags.IntVar(&opts.index, "index", 1, "Index of the container if there are multiple instances of a service [default: 1].")
+	flags.IntVar(&opts.index, "index", 0, "Index of the container if there are multiple instances of a service [default: 1 when copying from a container].")
 	flags.BoolVar(&opts.all, "all", false, "Copy to all the containers of the service.")
+	flags.MarkHidden("all")                                                                                                      //nolint:errcheck
+	flags.MarkDeprecated("all", "By default all the containers of the service will get the source file/directory to be copied.") //nolint:errcheck
 	flags.BoolVarP(&opts.followLink, "follow-link", "L", false, "Always follow symbol link in SRC_PATH")
 	flags.BoolVarP(&opts.copyUIDGID, "archive", "a", false, "Archive mode (copy all uid/gid information)")
 
--- a/pkg/compose/cp.go
+++ b/pkg/compose/cp.go
@@ -66,18 +66,23 @@ func (s *composeService) Copy(ctx context.Context, projectName string, options a
 		direction |= toService
 		serviceName = destService
 	}
-
-	containers, err := s.getContainers(ctx, projectName, oneOffExclude, true, serviceName)
-	if err != nil {
-		return err
-	}
-
-	if len(containers) < 1 {
-		return fmt.Errorf("no container found for service %q", serviceName)
-	}
-
+	var containers Containers
+	var err error
 	if direction == fromService || (direction == toService && options.Index > 0) {
-		containers = containers.filter(indexed(options.Index))
+		container, err := s.getSpecifiedContainer(ctx, projectName, oneOffExclude, true, serviceName, options.Index)
+		if err != nil {
+			return err
+		}
+		containers = append(containers, container)
+	} else {
+		containers, err = s.getContainers(ctx, projectName, oneOffExclude, true, serviceName)
+		if err != nil {
+			return err
+		}
+
+		if len(containers) < 1 {
+			return fmt.Errorf("no container found for service %q", serviceName)
+		}
 	}
 
 	g := errgroup.Group{}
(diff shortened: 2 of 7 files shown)


cmd/compose/cp.go
```diff
@@ -64,8 +64,10 @@ func copyCommand(p *projectOptions, backend api.Service) *cobra.Command {
 	}
 
 	flags := copyCmd.Flags()
-	flags.IntVar(&opts.index, "index", 1, "Index of the container if there are multiple instances of a service [default: 1].")
+	flags.IntVar(&opts.index, "index", 0, "Index of the container if there are multiple instances of a service [default: 1 when copying from a container].")
```
glours, line 67: Not a huge fan of specifying a default value that we change depending of the context of the `cp` command. if you have better idea to do it, let me know

thaJeztah: Perhaps remove the `[default: 1 when copying from a container].` altogether? (note that Cobra already prints defaults, so in most cases, there's no need to manually put it in the description)

```suggestion
	flags.IntVar(&opts.index, "index", 0, "Index of the container if there are multiple instances of a service .")
```

As to the flag itself; are indexes guaranteed to start with `1`? (i.e., could a container be removed, and it starts with `2` ?)

The "when copying from a container" is a bit confusing, as compose works with "services", correct? (in other words; I don't think there's an option to specify a _container_ - for that you'd use `docker cp` or `docker container cp`.

Overall I'm wondering a bit if we should have this option; my train of thought here is that;

- by default, we should assume all instances of a service to be identical
- compose to interact with a compose "service" (so consider all to be identical)
- it there _IS_ a need to target a specific instance (which _should be_ an exception to the rule); perhaps users should use `docker cp` / `docker container cp` for the "I know what I'm doing" situation.

cmd/compose/cp.go
```diff
@@ -64,8 +64,10 @@ func copyCommand(p *projectOptions, backend api.Service) *cobra.Command {
...
 	flags.BoolVar(&opts.all, "all", false, "Copy to all the containers of the service.")
+	flags.MarkHidden("all")                                                                                                      //nolint:errcheck
+	flags.MarkDeprecated("all", "By default all the containers of the service will get the source file/directory to be copied.") //nolint:errcheck
```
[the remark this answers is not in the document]
ndeloof: I prefer we have noose by end of line with a nolint comment vs start of line, where my mind will focus reading, for useless assignment

cmd/compose/cp.go
ndeloof: we need to double check `index` is aligned wit `compose exec --index`, and that we didn't duplicated the logic to select a target container here so that we enforce an homogeneous behavior

when copying from a container indeed only make sense with a single container as target. Either we should REQUIRE `index` to be set when multiple container match, or set "first indexed in the available ones" as default (which might not be `1`) - but, again, should behave the same as `compose exec`

> it there IS a need to target a specific instance (which should be an exception to the rule); perhaps users should use docker cp / docker container cp for the "I know what I'm doing" situation.

While it obvious they can just run `docker exec <container>` as name is predicable with compose, for some reason (?) `docker-compose exec` was introduced and adopted by many, probably because users prefer to rely on the same tool for everything (when you get a hammer...). So, even with _containers_ as target, a compose command makes sense to them. Or maybe this helps writing scripts that do not depend on te project name (included in container name) set by project dir?

thaJeztah: (let me "unresolve" so that it's not hidden)

pkg/compose/cp.go
```diff
@@ -57,23 +57,32 @@ func (s *composeService) Copy(ctx context.Context, projectName string, options a
 		if options.All {
 			return errors.New("cannot use the --all flag when copying from a service")
 		}
+		// due to remove of the --all flag, restore the default value to 1 when copying from a service container to the host.
+		if options.Index == 0 {
+			options.Index = 1
```
ndeloof, line 62: I wonder: when user run `docker compose cp service:/path /tmp` (copy from container to local) without index, shall we force index=1, or just use first indexed container (which might not be `1`) ? This can be addressed in a follow-up PR, just noting

ndeloof (approved):

cmd/compose/cp.go
ndeloof: `docker compose exec  <containername>` doesn't make sense to me, the whole verbage used by compose is service-centric. Use of `--index` is unpleasant but I can't imagine a workaround. I'd just expect it's not required for service running 1 replica and user don't have to bother with this BUT when they try to diagnose a distributed-system issue with multiple replicas. But then, obviously, using plain container name with `docker exec` would be a natural option, and maybe we should deprecate `--index` support and suggest user to better use `docker exec` for this purpose. Or maybe this is fine we keep this as legacy/usability, as we - at least - share runExec implementation code!

pkg/compose/cp.go
thaJeztah: Yes, that's what I was wondering in my comment above;

> As to the flag itself; are indexes guaranteed to start with 1? (i.e., could a container be removed, and it starts with 2 ?)

glours: Ok, I'll address that point in this PR

pkg/compose/cp.go
```diff
@@ -105,6 +87,35 @@ func (s *composeService) Copy(ctx context.Context, projectName string, options a
 	return g.Wait()
 }
 
+func (s *composeService) listContainersTargetedForCopy(ctx context.Context, projectName string, index int, direction copyDirection, serviceName string) (Containers, error) {
+	var containers Containers
+	var err error
+	if direction == fromService || (direction == toService && index > 0) {
```
ndeloof, line 93: nested `if`s make this a bit hard to read, maybe just split this into simpler top-level `switch` statement?

```go
switch {
	case direction == fromService && index == 0: 
                containers = ...
	case index != 0:
                containers = ...
        default:
                containers = ...
	}
```

pkg/compose/cp.go
```diff
@@ -96,6 +87,32 @@ func (s *composeService) Copy(ctx context.Context, projectName string, options a
...
+		containers, err = s.getContainers(ctx, projectName, oneOffExclude, true, serviceName)
+		return containers[:0], err
+	case direction == fromService || (direction == toService && index > 0):
+		container, err := s.getSpecifiedContainer(ctx, projectName, oneOffExclude, true, serviceName, index)
+		if err != nil {
+			return nil, err
+		}
+		return append(containers, container), nil
+	default:
+		containers, err = s.getContainers(ctx, projectName, oneOffExclude, true, serviceName)
```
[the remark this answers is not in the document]
thaJeztah: Looks like the first and last branch are basically "the same", so this could be;


```go
	switch {
	case index > 0:
		container, err := s.getSpecifiedContainer(ctx, projectName, oneOffExclude, true, serviceName, index)
		if err != nil {
			return nil, err
		}
		return Containers{container}, nil
	default:
		containers, err := s.getContainers(ctx, projectName, oneOffExclude, true, serviceName)
		if err != nil {
			return nil, err
		}
		if len(containers) < 1 {
			return nil, fmt.Errorf("no container found for service %q", serviceName)
		}
		if direction == fromService {
			// copy from container defaults to the first container
			return containers[:1], nil
		}
		return containers, nil
	}
```

pkg/compose/cp.go
```diff
@@ -62,20 +62,11 @@ func (s *composeService) Copy(ctx context.Context, projectName string, options a
...
-		return fmt.Errorf("no container found for service %q", serviceName)
-	}
-
-	if !options.All {
-		containers = containers.filter(indexed(options.Index))
-	}
-
 	g := errgroup.Group{}
 	for _, container := range containers {
 		containerID := container.ID
```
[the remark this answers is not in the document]
glours: I addressed it 😉

pkg/compose/cp.go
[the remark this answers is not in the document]
glours: I prefer to keep the `s.getContainers()` as it because it's wildly used in the rest of the code. Having a bit more complexity in the `copyFrom` and `copyTo` seems to be a good tradeoff IHMO.

I updated the switch as you proposed 👍

PR #9439, opened 2022-05-04
Improve contributing instructions

ikedam:
**What I did**

* Minor improves for contributing instructions:
    * Link to BUILDING.md for testing instructions.
    * Not that you have to build the binary prior to `make e2e-compose`.

**Related issue**

None

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -43,6 +43,7 @@ make e2e-compose
 ```
 
 Note that this requires a local Docker Engine to be running.
+Make sure you have built the CLI (see "Building the CLI"). Otherwise it uses `docker compose` available in your system.
 
 ## Releases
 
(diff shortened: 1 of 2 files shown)


glours:
Hi @ikedam 
Thanks for your contribution, maybe we can improve the user experience by automatically build the plugin before running the e2e tests by default? 
I can open a PR in that way if it make more sense than updating the doc
cc @ndeloof @ulyssessouza WDYT?

ndeloof (approved):

ikedam:
> automatically build the plugin before running the e2e tests by default?

All targets in Makefile are marked `.PHONY`, and I believe that means the plugin will be built everytime we run e2e tests. I don't think that's so useful.
It might make sense to add a new target like `build-and-e2e`.

glours:
@ikedam I opened a new PR #9623 to propose new targets in the Makefile and update the documentation
If that's fine for you, I'll close your PR

glours:
Replace by #9623

ikedam:
Thanks for the update.
The update for `CONRTRIBUTING.md` is still valid and I created #9639 .

Issue #9426, opened 2022-04-29
Regression in `down` in 2.5.0 - exit code for "nothing needed" went from 0 (success) to 1 (failure)


1ace:
#9384 is causing the behaviour of `down` to go from:
```
$ docker compose --project-name computed-project-name down
[+] Running 1/0
 ⠿ computed-project-name  Warning: No resource found to remove 0.0s
$ echo $?
0
```
to:
```
$ docker compose --project-name computed-project-name down
no container found for project "computed-project-name": not found
$ echo $?
1
```

This is causing a regression in our CI as we always run `down` before running `up` (because [this GitLab bug](https://gitlab.com/gitlab-org/gitlab/-/issues/15603) means we can't rely on `down` having been properly called during the previous run).

I'll be adding a `|| true` after `down` to ignore the error, but I think "I didn't need to do anything" shouldn't be an error, so I'd consider this a docker compose regression :)

ilibar-zpt:
same, use it in CI too

PR #9440, opened 2022-05-04
compose down exit=0 if nothing to remove

ndeloof:
**What I did**
Exit with status=0 when nothing has been found to be removed
Just print a warning

**Related issue**
closes https://github.com/docker/compose/issues/9426

### DIFF
--- a/pkg/compose/down.go
+++ b/pkg/compose/down.go
@@ -90,7 +90,7 @@ func (s *composeService) down(ctx context.Context, projectName string, options a
 	}
 
 	if !resourceToRemove && len(ops) == 0 {
-		w.Event(progress.NewEvent(projectName, progress.Done, "Warning: No resource found to remove"))
+		fmt.Fprintf(s.stderr(), "Warning: No resource found to remove for project %q.\n", projectName)
 	}
 
 	eg, _ := errgroup.WithContext(ctx)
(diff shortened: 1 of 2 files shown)


pkg/compose/down.go
```diff
@@ -53,6 +53,10 @@ func (s *composeService) down(ctx context.Context, projectName string, options a
 	project := options.Project
 	if project == nil {
 		project, err = s.getProjectWithResources(ctx, containers, projectName)
+		if api.IsNotFoundError(err) {
```
ulyssessouza, line 56: Not sure that `s.getProjectWithResources` work for this verification.
From what I see, it returns in the first occurrence of a not found resource. 
So if you still have another type of resource to remove (let's say `volumes`) it would be ignored.

https://github.com/docker/compose/blob/v2/pkg/compose/down.go#L239-L258

ulyssessouza (changes requested):

pkg/compose/down.go
ndeloof: indeed, but that's another issue :P
with no container, we just quit with error not found

ndeloof: I had in mind many time to refactor `projectFromName` into a set of `actualServices`+`actualVolumes`+`actualNetworks`  so that we can better manage this. But this is a way bigger change

glours (approved):
LGTM when the CI will be fixed

PR #9449, opened 2022-05-09
Introduce `compose watch` to rebuild/restart service on change to build context

ndeloof:
**What I did**
This is a follow-up for https://github.com/docker/compose/pull/9000

This PR introduces the `docker compose watch` command, as a draft implementation while exploring https://github.com/compose-spec/compose-spec/pull/253

This address the "least common denominator" to offer a watch-and-update option to compose application developer, re-building the service image and restarting service. Better optimized workflow, relying on service stack being able to hot-reload, rebuild with cache, etc  is not addressed (yet)

close https://github.com/docker/compose/issues/184 (created on ... April 2014!)

image


Mi3-14159:
Are you still on this? Would love to use this ❤️

ydennisy:
This would be amazing!!

PR #9462, opened 2022-05-13
Add support for command line completion by docker/cli

ndeloof:
(see https://github.com/docker/cli/pull/3429)

**What I did**
Added command line completion support

**Related issue**
closes https://github.com/docker/compose/issues/8550

image

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -133,7 +133,7 @@ require (
 )
 
 replace (
-	github.com/docker/cli => github.com/docker/cli v20.10.3-0.20220309205733-2b52f62e9627+incompatible
+	github.com/docker/cli => github.com/ndeloof/cli v0.0.0-20220317134108-66942b8c2379
 	github.com/docker/docker => github.com/docker/docker v20.10.3-0.20220309172631-83b51522df43+incompatible
 
 	// For k8s dependencies, we use a replace directive, to prevent them being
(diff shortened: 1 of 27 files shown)


go.mod
```diff
@@ -133,7 +133,7 @@ require (
 )
 
 replace (
-	github.com/docker/cli => github.com/docker/cli v20.10.3-0.20220309205733-2b52f62e9627+incompatible
+	github.com/docker/cli => github.com/ndeloof/cli v0.0.0-20220317134108-66942b8c2379
```
ulyssessouza, line 136: Hmmm... This if I understood well is related to `manager.Metadata#Completion`. The thing is that this is not merged in `docker/cli` yet... 

I'm also not sure if the `SchemaVersion` should change too.

ulyssessouza: Also... If we are already changing the struct, that would be great if we could add a `Meta: map[string]string` to open for other applications.

ndeloof: this one is actually obsolete, in docker/cli we gave up with a new schema, and just run the `__complete` command on plugins. If they don't implement this command, it will just fail and offer no completion.
Also, the idea of `SchemaVersion` is to avoid an open `map[string]string` that would make it unpredictable about the possible values being set by plugin.

ulyssessouza:
@ndeloof Since the `docker/cli` PR was merged, is this PR ready to be changed in the replace directive in go.mod ?

Arkemlar:
Am I correct that only required thing remained is this one MR for completion to be working?

ulyssessouza:
Closing this PR in favor of https://github.com/docker/compose/pull/9269 that includes the changes of this one

Issue #8531, opened 2021-06-24
docker compose does not use current buildkit builder


strophy:
**Description**

`docker compose build` does not use the current buildkit builder. both `docker build` and `docker-compose build` use this builder, which I need because it contains cache mount data needed for a particularly slow `npm ci` command.

**Steps to reproduce the issue:**
1. `docker buildx create --name cacheable --use`
2. `docker buildx inspect --bootstrap`
3. `docker compose build <service_name>`

**Describe the results you received:**
The build runs and completes successfully, but does not use my currently selected builder. This can be identified by (a) the slow build and (b) if the builder is stopped beforehand with `docker stop buildx_buildkit_cacheable0` then it is still stopped after the build completes. 

**Describe the results you expected:**
I expect `docker compose build` to use my current buildkit builder, named `buildx_buildkit_cacheable0`

**Additional information you deem important (e.g. issue happens only occasionally):**
Both `docker build` and `docker-compose build` will start the currently selected builder container (if stopped) and use it to build, which leads me to believe this might be a bug in `docker compose`.

Environment contains these vars:
`COMPOSE_DOCKER_CLI_BUILD=1`
`DOCKER_BUILDKIT=1`

I have tried setting `BUILDKIT_HOST=docker-container://buildx_buildkit_cacheable0` without any change in the behaviour of `docker compose build`.

**Output of `docker version`:**

```
Client: Docker Engine - Community
 Cloud integration: 1.0.17
 Version:           20.10.6
 API version:       1.41
 Go version:        go1.13.15
 Git commit:        370c289
 Built:             Fri Apr  9 22:46:38 2021
 OS/Arch:           linux/amd64
 Context:           default
 Experimental:      true

Server: Docker Engine - Community
 Engine:
  Version:          20.10.6
  API version:      1.41 (minimum version 1.12)
  Go version:       go1.13.15
  Git commit:       8728dd2
  Built:            Fri Apr  9 22:44:49 2021
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.4.4
  GitCommit:        05f951a3781f4f2c1911b05e61c160e9c30eaa8e
 runc:
  Version:          1.0.0-rc93
  GitCommit:        12644e614e25b05da6fd08a38ffa0cfe1903fdec
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0

```

**Output of `docker context show`:**  
[...]

PR #9466, opened 2022-05-14
feat: use current buildx builder

johanneswuerbach:
**What I did**

Use the logic from buildx to select the current builder instance to support using different instances.

This is essentially copying https://github.com/docker/buildx/blob/v0.8.1/commands/build.go#L241 to compose as buildx currently doesn't expose this logic.

I'm not entirely sure how to test those changes (`make test` is still passing), but I can't see any dedicated buildx tests or a way to test using different builders.

**Related issue**

Fixes https://github.com/docker/compose/issues/8531

image

### DIFF
--- a/pkg/compose/build_buildkit.go
+++ b/pkg/compose/build_buildkit.go
@@ -18,26 +18,32 @@ package compose
 
 import (
 	"context"
+	"net/url"
 	"os"
-	"path/filepath"
+	"strings"
 
 	"github.com/compose-spec/compose-go/types"
 	"github.com/docker/buildx/build"
 	"github.com/docker/buildx/driver"
+	ctxkube "github.com/docker/buildx/driver/kubernetes/context"
+	"github.com/docker/buildx/store"
+	"github.com/docker/buildx/store/storeutil"
+	"github.com/docker/buildx/util/confutil"
 	xprogress "github.com/docker/buildx/util/progress"
+	"github.com/docker/cli/cli/command"
+	"github.com/docker/cli/cli/context/docker"
+	ctxstore "github.com/docker/cli/cli/context/store"
+	dockerclient "github.com/docker/docker/client"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+	"golang.org/x/sync/errgroup"
+	"k8s.io/client-go/tools/clientcmd"
 )
 
 func (s *composeService) doBuildBuildkit(ctx context.Context, project *types.Project, opts map[string]build.Options, mode string) (map[string]string, error) {
-	const drivername = "default"
-	d, err := driver.GetDriver(ctx, drivername, nil, s.apiClient(), s.configFile(), nil, nil, nil, nil, nil, project.WorkingDir)
+	driverInfo, err := getDefaultDrivers(ctx, s.dockerCli, false, project.WorkingDir)
 	if err != nil {
 		return nil, err
-	}
-	driverInfo := []build.DriverInfo{
-		{
-			Name:   drivername,
-			Driver: d,
-		},
 	}
 
 	// Progress needs its own context that lives longer than the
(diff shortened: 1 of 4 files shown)


glours:
I wonder if we can ask Buildx maintainers to export this method through an existing API or command. I prefer to avoid code duplication and get benefit from Buildx future releases without having to backport code.
WDYT @ndeloof @ulyssessouza @crazy-max @tonistiigi ?

ndeloof:
We copy/pasted code from buildx as compose v2 was not yet an established long term plan for compose but we needed support for `build`. So it would make perfect sense now we avoid this and use an exported higher level function so we only have to prepare build options, and give buildx a `dockercli` reference. Just like we did for docker/cli `exec` and `run`

pkg/compose/build_buildkit.go
```diff
@@ -71,3 +76,214 @@ func (s *composeService) doBuildBuildkit(ctx context.Context, project *types.Pro
 
 	return imagesBuilt, err
 }
+
+//
+// All code below is from https://github.com/docker/buildx/blob/v0.8.1/commands/util.go
+//
+
+// getDefaultDrivers returns drivers based on current cli config
+func getDefaultDrivers(ctx context.Context, dockerCli command.Cli, defaultOnly bool, contextPathHash string) ([]build.DriverInfo, error) {
```
ndeloof, line 85: @tonistiigi would it make sense to export this func so we don't have to copy/paste this code?

ndeloof (approved):
LGTM so we get a short terms fix, but I expect in a follow-up PR we can avoid duplicating such buildx code, with adequate functions being exported

pkg/compose/build_buildkit.go
```diff
@@ -71,3 +76,214 @@ func (s *composeService) doBuildBuildkit(ctx context.Context, project *types.Pro
...
+	if err := eg.Wait(); err != nil {
+		return nil, err
+	}
+
+	return dis, nil
+}
+
+func configFromContext(endpointName string, s ctxstore.Reader) (clientcmd.ClientConfig, error) {
+	if strings.HasPrefix(endpointName, "kubernetes://") {
+		u, _ := url.Parse(endpointName)
```
thaJeztah, line 220: Wondering if we should have this in Compose, now that support for kubernetes endpoints having been removed from the docker CLI

(same for registering the kubernetes driver)

pkg/compose/build_buildkit.go
ndeloof: vendoring is indeed actually a form of copy/paste, but at least is automated :P
About invoking `buildx bake` from compose, this assumes the latter will parse the compose file the same way we do with in-memory changes, that could be an option... but not something we considered so far.
Also, Compose V2 can be used standalone, and then should be able to support `build` command without external dependency (but this is something we could reconsider)

Is there a way to invoke `buildx bake` passing directly solver options?

pkg/compose/build_buildkit.go
thaJeztah: The `kubernetes` endpoints were added to `docker context` in support of compose-on-kubernetes, which was deprecated and removed (together with support for creating contexts that use kubernetes endpoints), so I wonder if support for kubernetes should be added here (knowing that at least the contexts can no longer be created, but also that the rest of the CLI has no kubernetes support).

The scope of buildx is somewhat different, as it can also be used  without docker (which is mostly where these things come in)

pkg/compose/build_buildkit.go
tonistiigi: >  same way we do with in-memory changes

I'm not sure what this means. We can probably start to track this in a separate issue.

> Is there a way to invoke buildx bake passing directly solver options?

The solver options come from parsing the compose files. Only the path may be needed to be passed if it is not default.

ndeloof: There's not just a compose file: user can set overrides, configure variables for interpolations, and pass flags to amend the model in many ways, this is what I mean by "in memory model" - which we could marshall back to yaml, but this assumes bake is up to date with the compose-spec, which we can't guarantee.

johanneswuerbach:
Any update here, anything I can do to move this forward?

johanneswuerbach:
Rebased, anything else I should be doing here? We would still love to use compose with gha build caching and without this change it makes this fairly complicated.

johanneswuerbach:
@ndeloof / @crazy-max any way I could help to move this forward? That would be 🤩

n-rodriguez:
Hi there! Any news?

rory-lamendola:
^^ seconding that, any news? this would be great to fix!

justinrosenthal:
+1 I also got bit by this and have the same motivation: enabling caching in GitHub Actions.

johanneswuerbach:
Replaced by https://github.com/docker/compose/pull/9729

PR #9467, opened 2022-05-14
Fix up with seccomp file and replicas

### LINKED ISSUE #9458: Docker compose not working with seccomp file and replicas together
**Description**
Docker compose does not work with a seccomp file AND replicas toghether.
It fails with an error message stating an invalid seccomp filename

1. Use this docker-compose.yaml and seccomp.json file from [here](https://github.com/NovelService/NovelWorkerNode/tree/3b2ab3f9351ec6925ea4781e75205c936ec5d6ec/deployment). Important bit being a replica count above 1 and using a seccomp file
2. Create an empty .env file
3. run `docker compose up -d`

This error [gist](https://gist.github.com/XiangRongLin/620b52c3fb41ce54e69bc2c803fc13a3) which states that the content of the seccomp.json file is used as the filename

Both containers start succesfully

```
Docker Compose version v2.3.3
```

**Output of `docker info`:**

```
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc., v0.8.0)
  compose: Docker Compose (Docker Inc., v2.3.3)

Server:
 Containers: 18
  Running: 0
  Paused: 0
  Stopped: 18
 Images: 41
 Server Version: 20.10.13
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 1
 Plugins:
  Volume: local
 Swarm: inactive
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 2a1d4dbdb2a1030dc5b01e96fb110a9d9f150ecc
 runc version: v1.0.3-0-gf46b6ba
 init version: de40ad0
 Security Options:
  seccomp
   Profile: default
 Kernel Version: 4.19.128-microsoft-standard
 OSType: linux
 Architecture: x86_64
 CPUs: 12
 Total Memory: 12.43GiB
 Name: docker-desktop
 ID: SZYV:GGTU:6OFD:WBYC:UQMX:K3QG:KOLK:4VNS:HXKO:SAJX:AN6O:HISJ
 Debug Mode: false
 Labels:
 Experimental: false
 Insecure Registries:
  hubproxy.docker.internal:5000
  127.0.0.0/8
 Live Restore Enabled: false
```

**My analysis:**
Note: I never worked with GO, but I was able to debug the application and verified the behavior below

https://github.com/docker/compose/blob/a603e271173f2d3cc09721db1767009a49f7a5ad/pkg/compose/create.go#L354 here the `service.SecurityOpt` is passed into `parseSecurityOpts`. In that function the values of `service.SecurityOpt` are overwritten [here] with the content of the file https://github.com/docker/compose/blob/a603e271173f2d3cc09721db1767009a49f7a5ad/pkg/compose/create.go#L414.
And it seems that GO passes a reference to the array to the function, meaning that that change also occured for all caller functions.
Because my docker-compose file is using replicas, the second time the function gets called, the function gets called with `service.SecurityOpt` containing the content of the file instead of the filename itself.

XiangRongLin:
**What I did**
Work on a copy of `securityOpts` when parsing them. This means that the input in not modified and the second time the function gets called for the replica `securityOpts` still contains the filename instead of the content of the file.
This is based on the practice to not modify input values, which may not be a good practice in go. But I don't know since it's my first time developing in go.

An alternative solution would be to do this copy in [the loop](https://github.com/docker/compose/blob/a603e271173f2d3cc09721db1767009a49f7a5ad/pkg/compose/convergence.go#L235) for the replicas. In this case it is also ensured that if other parts of the options parsing is modifying the input, they will not produce the same problem.

**Related issue**

fixes #9458

### DIFF
--- a/pkg/compose/create.go
+++ b/pkg/compose/create.go
@@ -361,7 +361,9 @@ func (s *composeService) getCreateOptions(ctx context.Context, p *types.Project,
 		return nil, nil, nil, err
 	}
 
-	securityOpts, err := parseSecurityOpts(p, service.SecurityOpt)
+	var securityOptDeepCopy = make([]string, len(service.SecurityOpt))
+	copy(securityOptDeepCopy, service.SecurityOpt)
+	securityOpts, err := parseSecurityOpts(p, securityOptDeepCopy)
 	if err != nil {
 		return nil, nil, nil, err
 	}
--- a/pkg/e2e/compose_test.go
+++ b/pkg/e2e/compose_test.go
@@ -117,6 +117,11 @@ func TestLocalComposeUp(t *testing.T) {
 		res := c.RunDockerCmd(t, "network", "ls")
 		assert.Assert(t, !strings.Contains(res.Combined(), projectName), res.Combined())
 	})
+
+	t.Run("up with security options file and replicas", func(t *testing.T) {
+		res := c.RunDockerComposeCmd("-f", "./fixtures/seccomp-and-replica/compose.yaml", "--project-name", projectName, "up", "-d")
+		res.Assert(t, icmd.Expected{ExitCode: 0})
+	})
 }
 
 func TestComposePull(t *testing.T) {
--- a/pkg/e2e/fixtures/seccomp-and-replica/compose.yaml
+++ b/pkg/e2e/fixtures/seccomp-and-replica/compose.yaml
@@ -1,0 +1,8 @@
+services:
+  seccomp-and-replica:
+    image: nginx:alpine
+    security_opt:
+      - "seccomp:seccomp.json" # From https://github.com/moby/moby/blob/master/profiles/seccomp/default.json
+    deploy:
+      mode: replicated
+      replicas: 2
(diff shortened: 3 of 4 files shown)


ndeloof:
related: https://github.com/docker/cli/pull/3616

XiangRongLin:
> IIUC you had thoughts for an alternate fix, then could you please investigate this option?

That would require doing a deep copy of `service` before passing in variable into this function.
https://github.com/docker/compose/blob/a603e271173f2d3cc09721db1767009a49f7a5ad/pkg/compose/convergence.go#L235

GO does not provide deep copy functionality for structs out of the box so a library or a lot of code would be needed for that. Additionally the operation would be very expensive for the whole struct.

---

Another way would be to lift out the changes out of the copied function. I force-pushed the branch with this variant. This way only the neccessary objects is deep copied.

milas:
Although the current behavior (mutating the input slice) is not ideal, we'd really prefer to make the changes upstream via docker/cli#3616 and rely on that having better behavior instead of working around that here, so I'm closing this PR, but I appreciate your contribution and apologize for the back and forth!

PR #9487, opened 2022-05-19
use AdaptCmd when executing run command to properly propagate syscall…

glours:
… to the container process

**What I did**
Use the AdaptCmd when executing `run` command and correctly intercept and propagate syscall to the container

**Related issue**
Internal support ticket

image

### DIFF
--- a/cmd/compose/run.go
+++ b/cmd/compose/run.go
@@ -134,7 +134,7 @@ func runCommand(p *projectOptions, dockerCli command.Cli, backend api.Service) *
 			}
 			return nil
 		}),
-		RunE: Adapt(func(ctx context.Context, args []string) error {
+		RunE: AdaptCmd(func(ctx context.Context, cmd *cobra.Command, args []string) error {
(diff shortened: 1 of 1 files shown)


cmd/compose/run.go
```diff
@@ -134,7 +134,7 @@ func runCommand(p *projectOptions, dockerCli command.Cli, backend api.Service) *
 			}
 			return nil
 		}),
-		RunE: Adapt(func(ctx context.Context, args []string) error {
+		RunE: AdaptCmd(func(ctx context.Context, cmd *cobra.Command, args []string) error {
```
ndeloof, line 137: I don't understand how this can offer the suggested fix, as `Adapt` just executes `AdaptCmd`, could you please elaborate?

glours: Yes sure, it was to late to fix some syscall issue 😅 
I didn't noticed I changed my test protocol just a little bit to make the container process behaving correctly and not crashing the internal stack
So we can close this PR, I'm able now explain that's not a Compose issue

PR #9488, opened 2022-05-20
attach _only_ to services declared by project applying profiles

ndeloof:
**What I did**
ensure we only attach to services defined by project with applied profiles
if we don't, we will attach to all containers created with project label, which might match disabled profiles

**Related issue**
close https://github.com/docker/compose/issues/9286


glours (approved):
LGTM and works as expected with the issue sample

PR #9495, opened 2022-05-23
bump compose-go to 1.2.6

maxcleme:
Signed-off-by: Maxime CLEMENT <maxime.clement@docker.com>

**What I did**

Bump compose-go to 1.2.6 to integrate latest fix : 
+ https://github.com/compose-spec/compose-go/pull/264

**Related issue**

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,7 +6,7 @@ require (
 	github.com/AlecAivazis/survey/v2 v2.3.2
 	github.com/buger/goterm v1.0.4
 	github.com/cnabio/cnab-to-oci v0.3.1-beta1
-	github.com/compose-spec/compose-go v1.2.5
+	github.com/compose-spec/compose-go v1.2.6
 	github.com/containerd/console v1.0.3
 	github.com/containerd/containerd v1.6.2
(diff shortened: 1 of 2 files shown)


glours (approved):
LGTM

madhavajay:
Does anyone know when this will hit Docker Desktop? This broke our environment for everyone on Monday and cost us a few hours of debugging and forcing a lot of uninstalls and re-installs of old versions.

maxcleme:
@madhavajay We are working on extensive test suite to be added as long with the fix on the next release.

madhavajay:
@maxcleme any chance of a quick point release for Docker Desktop because this breaks docker compose v2 for a lot of people and the new Upgrade policy for docker is very aggressive meaning that most people are being pushed to the new version rapidly, I would prefer not to have to explain to all our customers, devs and users that they need to uninstall docker, go back to some old version etc.

madhavajay:
@maxcleme thanks, although surely since this is a regression it would be worth releasing a docker desktop without compose 2.5.1 and yanking it instead or just yanking the new docker desktop version? I suspect we may have to update our code by early next week just in case. 😕

Issue #9427, opened 2022-04-29
2.5.0 regression: `--force-recreate` + `--no-deps` leads to "no such service" errors for dependent services


nocive:
**Description**

**Steps to reproduce the issue:**
1. docker compose up -d --force-recreate --no-deps SOME_SERVICE

**Describe the results you received:**
"no such service: some-dependent-service"

**Describe the results you expected:**
Container to be recreated and started without errors

**Additional information you deem important (e.g. issue happens only occasionally):**
Issue happens consistently with 2.5.0 but doesn't happen with 2.4.1.

**Output of `docker compose version`:**

```
Docker Compose version v2.4.1
```

nicksieger:
I think this might be a difference in behavior from v1 to v2. v1 seems to happily ignore dependencies when you tell it `--no-deps`. v2 does not. Instead, it puts all services not mentioned on the command line[ into a slice of disabled services](https://github.com/docker/compose/blob/v2/cmd/compose/up.go#L56-L67). Later when [v2 prepares dependencies](https://github.com/docker/compose/blob/v2/pkg/compose/create.go#L107) (always, regardless of whether `--no-deps` was supplied) is when the "no such service" error is produced.

#9368 added implicit deps for links since 2.4.1, are you dependencies declared as links? If not, I think the issue has been around longer than 2.4.1.

I am seeing this same issue with 2.5 with [tilt-dev/tilt-example-docker-compose](https://github.com/tilt-dev/tilt-example-docker-compose/blob/fbc2388fa267731fc4fc60ccf580aea94bb391b1/docker-compose.yml) due to the implicit `links` dependency. Removing the dependency allows the service to be upped.

nicksieger:
FWIW this patch fixes the issue for me. If this behavior makes sense, I can open a PR with these changes.

```diff
1 file changed, 5 insertions(+), 1 deletion(-)
pkg/compose/create.go | 6 +++++-

modified   pkg/compose/create.go
@@ -183,7 +183,11 @@ func prepareServicesDependsOn(p *types.Project) error {
 		if service.DependsOn == nil {
 			service.DependsOn = make(types.DependsOnConfig)
 		}
-		deps, err := p.GetServices(dependencies...)
+
+		// Verify dependencies exist in the project, whether disabled or not
+		projAllServices := types.Project{}
+		projAllServices.Services = append(p.Services, p.DisabledServices...)
+		deps, err := projAllServices.GetServices(dependencies...)
 		if err != nil {
 			return err
[...]

PR #9504, opened 2022-05-27
fix: bring up services with deps with --no-deps

nicksieger:
**What I did**

Don't fail on not finding dependent services because they were put in the disabled slice.

**Related issue**
#9427.

Photo by Artem Lysenko from Pexels

### DIFF
--- a/pkg/e2e/start_stop_test.go
+++ b/pkg/e2e/start_stop_test.go
@@ -131,10 +131,11 @@ func TestStartStopWithDependencies(t *testing.T) {
 		assert.Assert(t, strings.Contains(res.Combined(), "e2e-start-stop-with-dependencies-foo-1"), res.Combined())
 	})
 
-	t.Run("Up no-deps", func(t *testing.T) {
+	t.Run("Up no-deps links", func(t *testing.T) {
 		_ = c.RunDockerComposeCmd("--project-name", projectName, "down")
-		res := c.RunDockerComposeCmd("-f", "./fixtures/dependencies/compose.yaml", "--project-name", projectName, "up", "--no-deps", "-d", "foo")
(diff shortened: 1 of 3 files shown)


maxcleme:
LGTM

pkg/e2e/start_stop_test.go
```diff
@@ -131,6 +131,12 @@ func TestStartStopWithDependencies(t *testing.T) {
 		assert.Assert(t, strings.Contains(res.Combined(), "e2e-start-stop-with-dependencies-foo-1"), res.Combined())
 	})
 
+	t.Run("Up no-deps links", func(t *testing.T) {
+		_ = c.RunDockerComposeCmd("--project-name", projectName, "down")
+		res := c.RunDockerComposeCmd("-f", "./fixtures/links/compose.yaml", "--project-name", projectName, "up", "--no-deps", "-d", "foo")
+		assert.Assert(t, strings.Contains(res.Combined(), "Container e2e-start-stop-with-dependencies-foo-1  Started"), res.Combined())
+	})
```
glours, line 138: Does it make sens to add a `compose ps` command and check that there isn't any `links-bar-1` container running, just to be sure we don't start unexpected sevices?

glours (approved):
LTGM
It seems that you didn't  sign-off your commits (legally needed to be able to merge your contribution), click on the `DCO` check if you need instructions

pkg/e2e/start_stop_test.go
nicksieger: Sounds like a good additional check to put in the test!

glours:
Ahahaha my bad I enabled auto-merge 🤦‍♂️ 
Let's add the test in a follow up PR

PR #9507, opened 2022-05-28
TC: Use switch case to simplify flag conversion and avoid multiple if statements

TheodosiouTh:
**What I did**
I used a map to remove multiple if statements from the code and make it easier to add new argument conversions.
Also, I added missing tests for `-v` and `--verbose`.

**Related issue**
There are no related issues.

### DIFF
--- a/cmd/compatibility/convert.go
+++ b/cmd/compatibility/convert.go
@@ -43,7 +43,7 @@ func getStringFlags() []string {
 	}
 }
 
-var argToActualFlag = map[string]string{
+var argToActualArgument = map[string]string{
 	"--verbose": "--debug",
 	// docker cli has deprecated -h to avoid ambiguity with -H, while docker-compose still support it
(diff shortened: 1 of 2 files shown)


debdutdeb (commented):
Honestly i don't see the need to use a static one-use map for a simple action like this.  There is no benefits to this, no performance gain or anything. Would've made some sense in terms of maintainability if there were a lot of such supp args, but for just four, a couple if statements don't hurt, rather I'd say is better. Probably a better option would be using switch instead (even if the list grows to say 100, a switch statement is far better than a clever use of map. simplicity 😉)

TheodosiouTh:
@debdutdeb I understand your way of thinking. 
I changed the code to use a switch case statement and it looks just as good as the static map.

debdutdeb (approved):
looks good to me

glours (approved):
LGTM

Issue #9345, opened 2022-04-02
Feature: DOCKER_BUILDKIT=1 in .env


mathroc:
Hi,

I'd like to add `DOCKER_BUILDKIT=1` in my `.env` file to enable building with webkit instead of having to add it to the command line every time. 

I know I can enable it globally but I would also like to make it easier for other contributors to enable it (by putting it in a `.env.dist` that they would copy and edit when starting working on the project)

Is this a valid request or is it not taken from the .env file on purpose?

ndeloof:
Alternatively you can configure buildkit to be enabled by engine setting, see https://docs.docker.com/develop/develop-images/build_enhancements/#to-enable-buildkit-builds.

mathroc:
> Alternatively you can configure buildkit to be enabled by engine setting, see https://docs.docker.com/develop/develop-images/build_enhancements/#to-enable-buildkit-builds.

I know, but this needs to be done for everyone working on the project. That's not very complicated, but that's one more step, and it would be nice if it could be avoided. Also, if people have projects where they don't want to use buildkit, so this would be annoying for them. (I don't know if enabling buildkit could break something on some projects or anything, so maybe this less important than I imagine)

PR #9512, opened 2022-05-30
Import dotenv file to os environment

ulyssessouza:
**What I did**
Import dotenv file to os environment to enable setting variables like `DOCKER_BUILDKIT` and others through `.env` file

**Related issue**
Resolves https://github.com/docker/compose/issues/9345

### DIFF
--- a/cmd/compose/compose.go
+++ b/cmd/compose/compose.go
@@ -27,6 +27,7 @@ import (
 
 	"github.com/compose-spec/compose-go/cli"
 	"github.com/compose-spec/compose-go/types"
+	composegoutils "github.com/compose-spec/compose-go/utils"
 	dockercli "github.com/docker/cli/cli"
 	"github.com/docker/cli/cli-plugins/manager"
 	"github.com/docker/cli/cli/command"
@@ -257,6 +258,10 @@ func RootCommand(dockerCli command.Cli, backend api.Service) *cobra.Command {
 			}
 		},
 		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
+			err := setEnvWithDotEnv(&opts)
+			if err != nil {
+				return err
+			}
 			parent := cmd.Root()
 			if parent != nil {
 				parentPrerun := parent.PersistentPreRunE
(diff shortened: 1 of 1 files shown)


glours (approved):
LGTM

laurazard (approved):
Looks good

wbydc:
Hi, looks like this breaks `env_file` option a bit.

Previously I had `.env` with private values for global setup, and generated `service-name.env` for several services in compose, witch were overwriting some values like this:

```shell
# .env
KEY=DEFAULT_VALUE

# first-service.env
KEY=VALUE_FOR_FIRST_SERVICE

# second-service.env
KEY=VALUE_FOR_SECOND_SERVICE
```

```yml
services:
    first-service:
        ...
        env_file:
            - .env
            - first-service.env
        ...

    second-service:
        ...
        env_file:
            - .env
            - second-service.env
        ...
```

But now it always uses values specified in `.env`, no matter the order in `env_file`.

I guess such behaviour should be mentioned in the documentation.

digaoddc:
I agree with @wbydc, the env files passed with the `env_file` options are now ignored. I'm having problems with this new version of docker-compose in our project.

We use it like this:
```
app:
  env_file:
     - .docker.env
     - .docker.env.local
```

Before this PR, first, the `docker.env` was loaded, then the file `docker.env.local` merge any settings defined in the first env file.

Now, these env files declared in the `docker-compose.yml` are completely ignored. I don't think this is the expected behavior.

curry684:
They are not ignored, just processed in the wrong order as demonstrated in https://github.com/docker/compose/issues/9608#issuecomment-1176735103

Either way, this 'fix' is currently breaking production environments and should be rolled back or fixed asap.

crzdg:
Do we really want to have the `.env`-file loaded automatically in any case. Regardless of having it loaded before or after the defined files in the `docker-compose`. If so, there should at least by a function to disable this feature. And with this, we're back to the current feature set where you have to define the file in the `docker-compose` and the corresponding `env_file`-key. 

I really can't understand why we want to have this feature in any place? As far as I understand it should be no problem to achieve the requested behavior with features available `<2.6.0`. This breaks and changes a lot, not only conceptional but also concerning security, CI/CD, and dev-ops in general.

I would love to see to roll-back this and hardly think again about this "feature"

ulyssessouza:
This precedence issues should be fixed with https://github.com/docker/compose/pull/9636.

About the `.env` file load. That can be avoided by setting `--env-file /dev/null` for example. That would make Docker Compose avoid the loading of `.env`

PR #9514, opened 2022-05-31
Fix bind mounts when in project volumes definition

ulyssessouza:
**What I did**
Fix bind mounts when in project volumes definition

This fixes the type of a volume back to a bind mount

**Related issue**
Resolves https://github.com/docker/for-mac/issues/6317

### DIFF
--- a/pkg/compose/create.go
+++ b/pkg/compose/create.go
@@ -718,9 +718,15 @@ MOUNTS:
 			// so `Bind` API is used here with raw volume string
 			// see https://github.com/moby/moby/issues/43483
 			for _, v := range service.Volumes {
-				if v.Target == m.Target && v.Bind != nil && v.Bind.CreateHostPath {
-					binds = append(binds, v.String())
-					continue MOUNTS
+				if v.Target == m.Target {
+					switch {
+					case string(m.Type) != v.Type:
+						v.Source = m.Source
+						fallthrough
+					case v.Bind != nil && v.Bind.CreateHostPath:
+						binds = append(binds, v.String())
+						continue MOUNTS
+					}
 				}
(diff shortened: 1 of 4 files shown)


pkg/compose/create.go
```diff
@@ -892,9 +892,10 @@ func isUnixAbs(path string) bool {
 
 func buildMount(project types.Project, volume types.ServiceVolumeConfig) (mount.Mount, error) {
 	source := volume.Source
+	sourceTrimmed := strings.TrimPrefix(source, project.Name+"_")
```
ndeloof, line 895: you should not have to do this. compose model is expected to be consistent, so project prefix should be added to all items in `project.Volumes` and resource references.

cmd/compose/compose.go
```diff
@@ -206,6 +206,19 @@ func (o *projectOptions) toProject(services []string, po ...cli.ProjectOptionsFn
 
 	project.WithoutUnnecessaryResources()
 
+	for _, s := range project.Services {
+		for i, v := range s.Volumes {
+			if projectVolume, ok := project.Volumes[v.Source]; ok &&
+				projectVolume.DriverOpts["o"] == types.VolumeTypeBind {
+				s.Volumes[i].Source = project.Name + "_" + s.Volumes[i].Source
```
ndeloof, line 213: I'm not a big fan we hack the model this way. Maybe the "use bind vs mount" logic used in `create` could check the volume driver and use bind when necessary?

ulyssessouza: PTAL

milas (approved):
👍

glours (approved):
LGTM

PR #9545, opened 2022-06-08
e2e: add test for ps

nicksieger:
**What I did**

Add E2E test for `docker compose ps`. This PR will fail until #9539 is merged; I'll rebase this afterward.

**Related issue**


pkg/e2e/ps_test.go
```diff
@@ -0,0 +1,96 @@
...
+				count++
+			}
+		}
+		assert.Equal(t, 2, count, "Did not match both services:\n"+res.Combined())
+	})
+
+	t.Run("json", func(t *testing.T) {
+		res = c.RunDockerComposeCmd("-f", "./fixtures/ps-test/compose.yaml", "--project-name", projectName, "ps", "--format", "json")
+		var output []map[string]interface{}
+		err := json.Unmarshal([]byte(res.Combined()), &output)
```
milas, line 61: fwiw one thing I've done in situations like this before is to have an anonymous type like
```
type psOutput struct {
  Name string
  Publishers json.RawMessage
}
```
and then you can use `assert.JSONEq` on `.Publishers` and avoid type casting nuisances

(not suggesting you refactor this - it's totally readable as-is!)

milas (approved):
Yay for E2E tests 🎉

glours (approved):
LGTM

pkg/e2e/ps_test.go
nicksieger: Well, I suppose I could unmarshal it back into `api.ContainerSummary`! That might make this a little easier to look at. Fixed in ebb45b400cbf35febda5d969ece9892e60386c46.

PR #9547, opened 2022-06-08
e2e: unmarshal json into container summaries

nicksieger:
**What I did**

Oops! This didn't quite make it into #9545 before it was merged. I like how this reads even though it puts a dependency on api (or maybe that's not desired?). 

**Related issue**

### DIFF
--- a/pkg/e2e/ps_test.go
+++ b/pkg/e2e/ps_test.go
@@ -22,6 +22,8 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+
+	"github.com/docker/compose/v2/pkg/api"
 )
 
 func TestPs(t *testing.T) {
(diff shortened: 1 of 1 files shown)


pkg/e2e/ps_test.go
```diff
@@ -57,37 +59,34 @@ func TestPs(t *testing.T) {
...
 		err := json.Unmarshal([]byte(res.Combined()), &output)
 		assert.NoError(t, err)
 
 		count := 0
 		assert.Equal(t, 2, len(output))
 		for _, service := range output {
-			publishers := service["Publishers"].([]interface{})
-			if service["Name"] == "e2e-ps-busybox-1" {
+			publishers := service.Publishers
+			if service.Name == "e2e-ps-busybox-1" {
```
thaJeztah, line 70: Are the services expected to be returned in a consistent order? If so, this loop may not be needed, and possibly it could be a single compare between an expected slice (in the right order)?

thaJeztah: Ah, I guess there may be other fields that need to be ignored in the comparing in that case 🤔

nicksieger: Yeah, the order may be consistent, I'm not familiar enough with the way the summaries are generated to know for sure. I was a little surprised that they were reported in alpha order instead of declaration order in the compose file.

nicksieger: If there are any documented assertions about order then I'll update this to match.

thaJeztah: It was mostly a "drive-by" comment from my side; perhaps others more familiar with the codebase could provide more details 😅 👍

laurazard (approved):
LGTM :) much more readable

PR #9553, opened 2022-06-10
add support for setting secret from env variable

ndeloof:
**What I did**
add support for setting secret from env variable

**Related issue**
https://github.com/compose-spec/compose-spec/issues/251

image

### DIFF
--- a/pkg/compose/secrets.go
+++ b/pkg/compose/secrets.go
@@ -1,0 +1,88 @@
+/*
+   Copyright 2020 Docker Compose CLI authors
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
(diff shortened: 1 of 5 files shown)


jmb-altana:
Thanks for the impl @ndeloof

pkg/compose/secrets.go
```diff
@@ -0,0 +1,88 @@
...
+	b := bytes.Buffer{}
+	tarWriter := tar.NewWriter(&b)
+	mode := uint32(0400)
+	if config.Mode != nil {
+		mode = *config.Mode
+	}
+
+	target := config.Target
+	if config.Target == "" {
+		target = "/run/secrets/" + config.Source
```
glours, line 67: maybe we can use `path.Join` instead of plain string? (or `filepath.Join` if we need to target Windows containers)

glours (approved):
LGTM

laurazard (approved):
LGTM

PR #9561, opened 2022-06-15
e2e: fix subtests and block parallel unsafe tests

milas:
**What I did**
The big mechanical change here is to NOT store `t` as a field on
the `CLI` object (which has been renamed as well to fix the odd
capitalization). The way the tests are structured meant that the
"subtests" were using the _parent_ `*testing.T` instance, which
causes various oddities including confusing messages on failure
about a child test causing a panic in the parent.

Additionally, a few tests have been blocked from running in
parallel because they are sharing `compose.yaml` fixtures and
can fail as a result (e.g. due to a port conflict on the host).
I'll fix these in follow-up PRs but want to focus on correctness
for the tests before optimizing them.

baby turkey
(a baby turkey! 🦃)

### DIFF
--- a/pkg/e2e/cancel_test.go
+++ b/pkg/e2e/cancel_test.go
@@ -33,10 +33,10 @@ import (
 )
 
 func TestComposeCancel(t *testing.T) {
-	c := NewParallelE2eCLI(t, binDir)
+	c := NewParallelCLI(t)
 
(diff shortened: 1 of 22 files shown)


nicksieger (approved):
Sweet, nice cleanup

glours (approved):
LGTM

PR #9568, opened 2022-06-16
e2e: improve test output on failures

milas:
**What I did**
This is mostly marking a bunch of the run methods as helpers so
that the internal assertions they do will show the line number of
the calling test instead.

There's also some small tweaks around the plugin initialization to
help with the output in the event that it fails to make it easier
to debug what went wrong.

Catalina fox


nicksieger (approved):
Changes so far LGTM, do you have more changes to make to update calls to helper functions?

milas:
Apparently so! (I failed at rebasing this apparently 😓)

laurazard (approved):
LGTM!

glours (approved):

PR #9569, opened 2022-06-16
Add missing info from compose-v1 build reference docs

usha-mandya:
Signed-off-by: Usha Mandya <usha.mandya@docker.com>

**What I did**
Added info missing from the  Compose v2 reference. The detailed info is present in the v1 reference docs  - https://docs.docker.com/compose/reference/build/#native-build-using-the-docker-cli

### DIFF
--- a/docs/reference/compose_build.md
+++ b/docs/reference/compose_build.md
@@ -28,3 +28,15 @@ the image is tagged with that name, substituting any variables beforehand. See
 
 If you change a service's `Dockerfile` or the contents of its build directory,
 run `docker compose build` to rebuild it.
+
+### Native build using the docker CLI
+
+Compose by default uses the `docker` CLI to perform builds (also known as "native
+build"). By using the `docker` CLI, Compose can take advantage of features such
+as [BuildKit](../../develop/develop-images/build_enhancements.md), which are not
+supported by Compose itself. BuildKit is enabled by default on Docker Desktop,
+but requires the `DOCKER_BUILDKIT=1` environment variable to be set on other
+platforms.
+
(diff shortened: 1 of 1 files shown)


docs/reference/compose_build.md
```diff
@@ -28,3 +28,15 @@ the image is tagged with that name, substituting any variables beforehand. See
...
 run `docker compose build` to rebuild it.
+
+### Native build using the docker CLI
+
+Compose by default uses the `docker` CLI to perform builds (also known as "native
+build"). By using the `docker` CLI, Compose can take advantage of features such
+as [BuildKit](../../develop/develop-images/build_enhancements.md), which are not
+supported by Compose itself. BuildKit is enabled by default on Docker Desktop,
+but requires the `DOCKER_BUILDKIT=1` environment variable to be set on other
+platforms.
```
thaJeztah, line 39: Reading this part again, I think this may be (at least partially) outdated now;

- compose now uses a buildx client by default (currently it's built-in in compose, which _may_ change, but that should generally be an implementation detail for the user)
- the default is to use BuildKit (so there should no longer be a need to set `DOCKER_BUILDKIT=1`), _however_ I _think_ it's still possible to _disable_ BuildKit (through `DOCKER_BUILDKIT=0`) - not sure it's worth including that information here though.
- I'm not sure what the current behavior is on Windows (may be different between "windows client, linux engine" and "windows client and windows engine)


Perhaps @glours or @ulyssessouza is able to fill in the missing details on the above

milas:
Echoing what @thaJeztah said - Compose v2 does not have the "native build" / CLI functionality, that only applies to v1, so it's intentional that this section isn't in the docs anymore!

PR #9572, opened 2022-06-17
Use appropriate dependency condition for one-shot containers when running `compose up --wait`

### LINKED ISSUE #9273: `up --wait` should not wait for one-shot containers
**Description**

I guess the title says all.

```console
$ cat docker-compose.yaml
version: '3'
services:
  oneshot:
    image: ubuntu
    command: echo 'hello world'
  longrunning:
    image: ubuntu
    depends_on:
      oneshot:
        condition: service_completed_successfully
    command: sleep infinity

$ docker-compose version
Docker Compose version v2.3.3

# Hangs forever
$ docker-compose up --wait
[+] Running 2/2
 ⠿ Container jenkins-agent-dind-oneshot-1      Exited                             0.9s
 ⠿ Container jenkins-agent-dind-longrunning-1  Healthy                            1.4s
```

It hangs forever, most likely waiting for one-shot containers to be running.

It should stop waiting once all the long-running services are up (running or healthy).

How can docker-compose detect which services are one-shot and which ones are long-running?

Hm... not sure if it's the best way, but I would say that, if a service is found under any `depends_on` with condition `service_completed_successfully`, it can be inferred as one-shot. Otherwise, long-running.

Above.

**Output of `docker info`:**

Not meaningful, I guess.

Not meaningful, I guess.

laurazard:

**What I did**

When creating a dependency map to wait on during `compose up --wait`, we were previously applying the `running_or_healthy` condition to all containers indiscriminately. 

However, in compose files with one-shot containers used for setup (database migrations, etc.) that are configured as dependencies of others with the condition `service_completed_successfully`, this does not work and the command will get stuck as it is expected that these containers run and then exit, and so they never satisfy `running_or_healthy`.

This PR changes the dependency logic to use the appropriate condition for these containers so that `compose up --wait` exits properly when all services have reach the desired state (running/healthy for some, exited for one-shots).

Also add e2e tests to ensure `compose up --wait` does not get stuck forever waiting for one-shot containers.

**Related issue**

Fixes #9273

IMG_2944

### DIFF
--- a/pkg/e2e/fixtures/dependencies/deps-completed-successfully.yaml
+++ b/pkg/e2e/fixtures/dependencies/deps-completed-successfully.yaml
@@ -1,0 +1,11 @@
+version: '3'
+services:
+  oneshot:
+    image: ubuntu
+    command: echo 'hello world'
+  longrunning:
+    image: ubuntu
+    depends_on:
+      oneshot:
+        condition: service_completed_successfully
+    command: sleep infinity
--- a/pkg/e2e/fixtures/dependencies/deps-completed-successfully.yaml
+++ b/pkg/e2e/fixtures/dependencies/deps-completed-successfully.yaml
@@ -1,10 +1,9 @@
-version: '3'
 services:
   oneshot:
-    image: ubuntu
+    image: alpine
     command: echo 'hello world'
   longrunning:
-    image: ubuntu
+    image: alpine
     depends_on:
       oneshot:
         condition: service_completed_successfully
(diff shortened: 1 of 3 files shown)


milas (approved):
LGTM

ngwxw (approved):

ngwxw (approved):

pkg/e2e/fixtures/dependencies/deps-completed-successfully.yaml
```diff
@@ -0,0 +1,11 @@
+version: '3'
```
glours, line 1: Nit: No need to add the version 😉

glours (approved):
LGTM

felipecrs:
What animal is this?

dimaqq:
cc @ilyamonakov for your case of `--wait` waiting forever for a one-shot container in docker compose 2.15.x

PR #9579, opened 2022-06-21
build.go: initialize CustomLabels map if nil

paroque28:
This change fixes the` panic: assignment to entry in nil map` error.
I also accessed the map directly instead of the copy value, otherwise the label doesn't get set.

**What I did**

**Related issue**

Error that got me here:
```
panic: assignment to entry in nil map

goroutine 16 [running]:
github.com/docker/compose/v2/pkg/compose.(*composeService).getLocalImagesDigests(0x1c0006d4318?, {0x1579128?, 0xc00069f710?}, 0xc000419220)
        /home/prodrigu/go/pkg/mod/github.com/docker/compose/v2@v2.6.0/pkg/compose/build.go:195 +0x305
github.com/docker/compose/v2/pkg/compose.(*composeService).ensureImagesExists(0x0?, {0x1579128, 0xc00069f710}, 0xc000419220, 0x0)
        /home/prodrigu/go/pkg/mod/github.com/docker/compose/v2@v2.6.0/pkg/compose/build.go:105 +0x7b
github.com/docker/compose/v2/pkg/compose.(*composeService).create(0xc0003efc30?, {0x1579128, 0xc00069f710}, 0xc000419220, {{0xc0004284e0, 0x2, 0x2}, 0x1, 0x0, {0x0, ...}, ...})
        /home/prodrigu/go/pkg/mod/github.com/docker/compose/v2@v2.6.0/pkg/compose/create.go:66 +0x173
github.com/docker/compose/v2/pkg/compose.(*composeService).Up.func1({0x1579128, 0xc00069f710})
        /home/prodrigu/go/pkg/mod/github.com/docker/compose/v2@v2.6.0/pkg/compose/up.go:36 +0x8e
github.com/docker/compose/v2/pkg/progress.Run.func1({0x1579128?, 0xc00069f710?})
```

image

### DIFF
--- a/pkg/compose/build.go
+++ b/pkg/compose/build.go
@@ -189,11 +189,11 @@ func (s *composeService) getLocalImagesDigests(ctx context.Context, project *typ
 		images[name] = info.ID
 	}
 
-	for _, s := range project.Services {
-		imgName := getImageName(s, project.Name)
+	for i := range project.Services {
+		imgName := getImageName(project.Services[i], project.Name)
 		digest, ok := images[imgName]
 		if ok {
-			s.CustomLabels[api.ImageDigestLabel] = digest
+			project.Services[i].CustomLabels.Add(api.ImageDigestLabel, digest)
 		}
 	}
 


paroque28:
It's my first PR so any feedback is welcome

pkg/compose/build.go
```diff
@@ -189,11 +189,14 @@ func (s *composeService) getLocalImagesDigests(ctx context.Context, project *typ
 		images[name] = info.ID
 	}
 
-	for _, s := range project.Services {
+	for i, s := range project.Services {
 		imgName := getImageName(s, project.Name)
```
milas, line 193:
```suggestion
	for i := range project.Services {
		imgName := getImageName(project.Services[i], project.Name)
```

Minor suggestion - would prefer to avoid mixing and matching references within a given loop

milas (changes requested):
Good catch, and thanks for the PR!

CI failed because your commit didn't include a "sign off", you can read the details including how to fix it in the [DCO action log](https://github.com/docker/compose/pull/9579/checks?check_run_id=6978253030) - once you've done that and re-pushed, I'll get this merged 🙂

paroque28:
Done, thank you @milas :)

glours:
Hey @paroque28 
Thanks for your contribution.
Is this PR link to an existing issue? Is it possible to have more context on why this fix? Just want to be sure we address the root cause and not an symptom of the issue

paroque28:
@glours 
There is no linked issue, but I can create one if you need it.

The error was this:
```
panic: assignment to entry in nil map
```
I think it's a symptom since I am loading my project using the `loader.Load()` and the result contains an uninitialized `CustomLabels` map

I also notices that `"com.docker.compose.project"` was not set, and this is required for doing `docker compose up`

Here is my code: https://github.com/paroque28/container-chief/blob/master/pkg/manager/compose.go#L128
It's a small function

glours:
@paroque28 
We think that this issue should be fix directly in `compose-go`, so if that's fine for you we'll close this PR

paroque28:
@glours  yes, that's what I did in my code but it's not enough. The loader doesn't initialize the custom labels. That's why I created this PR

glours:
@paroque28 that's exactly my point, we should fix the behavior of `loader.Load`, I'll open a PR in `compose-go`

paroque28:
Oh yeah go it!

Sure, let's close this one

pkg/compose/build.go
```diff
@@ -189,11 +189,14 @@ func (s *composeService) getLocalImagesDigests(ctx context.Context, project *typ
...
 
-	for _, s := range project.Services {
-		imgName := getImageName(s, project.Name)
+	for i := range project.Services {
+		imgName := getImageName(project.Services[i], project.Name)
 		digest, ok := images[imgName]
 		if ok {
-			s.CustomLabels[api.ImageDigestLabel] = digest
+			if project.Services[i].CustomLabels == nil {
+				project.Services[i].CustomLabels = make(map[string]string)
```
glours, line 197: This is not needed, using the right method will let us remove this safely

pkg/compose/build.go
```diff
@@ -189,11 +189,14 @@ func (s *composeService) getLocalImagesDigests(ctx context.Context, project *typ
...
+			}
+			project.Services[i].CustomLabels[api.ImageDigestLabel] = digest
```
glours, line 199: Labels property propose an `add` method we should use to add the label

glours (changes requested):

paroque28:
@glours  done, sorry for the delay.

pkg/compose/build.go
paroque28: This assumes that the added labels didn't exist before, is that accurate?

glours:
@paroque28 you need to "sign off" your latest commit to make the CI pass

glours (approved):
LGTM

hmb (commented):
If the commit message is correct, this might not work. If `CustomLabels` is `nil` the code has no effect.

hmb:
> If the commit message is correct, this might not work. If `CustomLabels` is `nil` the code has no effect.

My bad, should have looked into the current code, it's already fixed

PR #9583, opened 2022-06-21
ci: workflow to dispatch event on release for docs reference

crazy-max:
closes https://github.com/docker/compose/pull/9296

follow-up https://github.com/docker/docker.github.io/pull/14976

**What I did**

creates workflow with a release published types event. This workflow will be triggered when a release is published. The step in this workflow will send a `compose-docs` event with a payload that contains the git context and release name from GitHub event as well as the dockerfile target that will be used upstream. here we use the official `actions/github-script` action instead of a community one to send the event.

cc @glours

**Related issue**


crazy-max:
Convert to draft for now, see https://github.com/docker/docker.github.io/pull/14976#issuecomment-1165726688

crazy-max:
As we can directly open a PR on docs repo from here there is no need to rely on dispatch events which avoid to creating a workflow on docs repo: https://github.com/docker/docker.github.io/pull/14976#issuecomment-1165737633

PTAL @glours

glours (approved):
LGTM 👍

PR #9585, opened 2022-06-22
networks: prevent issues due to duplicate names

milas:
**What I did**
Within the Docker API/engine, networks have unique IDs, and the
name is a friendly label/alias, which notably does NOT have any
guarantees around uniqueness (see moby/moby#18864 [^1]).

During day-to-day interactive/CLI Compose usage, this is rarely
an issue, as Compose itself isn't creating networks concurrently
across goroutines. However, if multiple Compose instances are
executed simultaneously (e.g. as part of a test suite that runs
in parallel), this can easily occur.

When it does happen, it's very confusing for users and resolving
it via the `docker` CLI is not straightforward either [^2].

There's two primary changes here:
 * Pass `CheckDuplicates: true` to the Docker API when creating
   networks to reduce the likelihood of Compose creating duplicates
   in the first place
 * On `down`, list networks using a name filter and then remove
   them all by ID, as the Docker API will return an error if the
   name alias is used and maps to more than one network

Hopefully, this provides a better UX, since the issue should be
less likely to occur, and if it does, it can now be resolved via
standard Compose workflow commands.

[^1]: https://github.com/moby/moby/issues/18864
[^2]: https://stackoverflow.com/questions/63776518/error-2-matches-found-based-on-name-network-nameofservice-default-is-ambiguo

**Related issue**
* moby/moby#18864

coyote puppies


milas:
Note on tests: I wrote an E2E test for this but backed it out of the PR because it's too flaky, as I was running 2x `docker network create` concurrently to trigger the condition, which is racy -- the CLI makes _some_ attempt to prevent duplicates.

To do it reliably, we'd need the Docker API client in the E2E tests, which I'm not sure if we want right now. (Docker API lets you specify `CheckDuplicates`, so by setting that to `false`, you can always create multiple networks with the same name.)

pkg/compose/down_test.go
```diff
@@ -59,8 +60,16 @@ func TestDown(t *testing.T) {
 	api.EXPECT().ContainerRemove(gomock.Any(), "456", moby.ContainerRemoveOptions{Force: true}).Return(nil)
 	api.EXPECT().ContainerRemove(gomock.Any(), "789", moby.ContainerRemoveOptions{Force: true}).Return(nil)
 
-	api.EXPECT().NetworkInspect(gomock.Any(), "myProject_default", moby.NetworkInspectOptions{}).Return(moby.NetworkResource{Name: "myProject_default"}, nil)
-	api.EXPECT().NetworkRemove(gomock.Any(), "myProject_default").Return(nil)
+	// network names are not guaranteed to be unique, ensure Compose handles
+	// cleanup properly if duplicates are inadvertently created
+	api.EXPECT().NetworkList(gomock.Any(), moby.NetworkListOptions{
```
nicksieger, line 65: Should this be reconciled with the other `NetworkList` mock from line 51?

nicksieger (approved):
LGTM

pkg/compose/down_test.go
milas: Good call - added in 5f89365c

glours (approved):
LGTM

ptrdom:
This issue can still be very easily reproduced on `Docker Desktop 4.19.0`, I think it uses version of compose that already has this change. With two processes starting the same compose it might sometimes work, but make it 3 and duplicates will happen every time, making compose fail to start. I reported the issue here initially - https://forums.docker.com/t/parallel-execution-of-docker-compose-up-commands-with-same-configuration/136142.

Any ideas what could be done?

ndeloof:
@ptrdom indeed easy to reproduce
```
Option CheckDuplicate is there to provide a best effort checking of any networks
which has the same name but it is not guaranteed to catch all name collisions.
```
Seems we don't have a better option here, as long as multiple compose executions run concurrently, all of those will try to create resources with the same name, and engine can't guarantee unicity (see https://github.com/moby/moby/issues/18864#issuecomment-167202589 for reason)

Regarding your use case, would need more details to better understand how you run multiple execution of compose for testing purpose with the same project name. Feel free to open a dedicated issue

thaJeztah:
I think @akerouanton was looking into this from the engine side. It always caused some head-scratching on my side, but was told in the past by the libnetwork maintainers "that's the way it is, and we can't promise unique names", but that may not actually be the case, and (if that's correct), we're considering `CheckDuplicate` to become redundant and network-names to always be unique.

(but @akerouanton may be able to fill me in on that one).

thaJeztah:
^^ we may need some migration path though for cases where there's existing duplicate names 😓

akerouanton:
>  "that's the way it is, and we can't promise unique names"

My educated guess is that it wasn't possible due to the eventually consistent nature of Classic Swarm's control plane. Since we don't support it anymore, I can't see any good reason we can't make sure a network name is not unique. 

> we may need some migration path though for cases where there's existing duplicate names sweat

Actually I don't think so: the last time I tried, I was unable to connect containers to networks with duplicate names, making these networks totally unusable. I think at some point the Engine broke support for this "networks with duplicate names" (dubious) feature.

For the record, kind has some far-fetched code to mitigate the exact issue described by @ptrdom: it first creates its network, and then check if there's a duplicate and remove it if needed.

I'll bump this in my to-do list to address it in the near future.

ptrdom:
Many thanks for hopping on this so eagerly, you guys rock! 🚀 ❤️

thaJeztah:
> it wasn't possible due to the eventually consistent nature of Classic Swarm's control plane

Ah! Yes that was it what you mentioned 👍 

> I'll bump this in my to-do list to address it in the near future.

Agreed, let's see if we can this fixed. We could already create a ticket in moby/moby for this (can be just an early placeholder, but if you have some notes on the work that needs to be done, we can paste it in there, in case someone wants to start working on that (doesn't have to be "you").

PR #9587, opened 2022-06-22
build: respect dependency order for classic builder

milas:
**What I did**
When using the "classic" (non-BuildKit) builder, ensure that
services are iterated in dependency order for a build so that
it's possible to guarantee the presence of a base image that's
been added as a dependency with `depends_on`. This is a very
common pattern when using base images with Compose.

A fix for BuildKit is blocked currently until we can rely on a
newer version of the engine (see docker/compose#9324).

**Related issue**
* docker/compose#9324
* docker/compose#8538
**⚠️ These are not being closed, as this doesn't fix things if you're using BuildKit**

American mink


glours (approved):
LGTM

You just need to add the license header to the 2 Dockerfile you added 😉

PR #9595, opened 2022-06-25
Replace deprecated ioutil pkg with os & io

abhinavnair:
**What I did**
As of Go 1.16, the same functionality of `ioutil` pkg is now provided by package `io` or
package `os`, and those implementations should be preferred in new code.

So replacing all usage of `ioutil` pkg with `io` & `os`.

**Related issue**
None

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -7,6 +7,7 @@ linters:
   disable-all: true
   enable:
     - deadcode
+    - depguard
     - errcheck
     - gocyclo
     - gofmt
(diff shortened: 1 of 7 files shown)


thaJeztah:
Thanks! Perhaps would be good to also configure golangci-lint to prevent this package from accidentally being used; similar to https://github.com/docker/cli/commit/8c3ae386cd263c5f4fb4c8f7519febedaf1e3281

abhinavnair:
> Thanks! Perhaps would be good to also configure golangci-lint to prevent this package from accidentally being used; similar to [docker/cli@8c3ae38](https://github.com/docker/cli/commit/8c3ae386cd263c5f4fb4c8f7519febedaf1e3281)

Thanks for the suggestion. I have added a commit for golangci-lint 👍🏻

nicksieger (approved):
LGTM

Issue #8097, opened 2021-02-03
'yes' is not boolean in YAML 1.2 spec


aleung:
## Description of the issue

Docker compose reports that the compose file is invalid because an environment variable contains true.

The value of the environment is `yes`. According to YAML 1.2, boolean is `true⎮True⎮TRUE⎮false⎮False⎮FALSE` and `yes` is no longer boolean type.

My docker-compose.yml file is generated by program (using the js module [yaml](https://eemeli.org/yaml)). The yaml library which comply with YAML 1.2 won't enclose `yes` in quotes because it doesn't enclose string in quotes.

I tried to add yaml version 1.2 directive to the head of docker-compose.yml but still got the same error.

## Context information (for bug reports)

**Output of `docker-compose version`**
```
docker-compose version 1.28.2, build unknown
docker-py version: 4.4.1
CPython version: 3.6.9
OpenSSL version: OpenSSL 1.1.1  11 Sep 2018
```

**Content of `docker-compose.yml`**
```yaml
%YAML 1.2
---
version: "3"
services:
  zookeeper:
    image: zookeeper
  kafka:
    image: kafka
    environment:
      KAFKA_CFG_ZOOKEEPER_CONNECT: zookeeper:2181
      ALLOW_PLAINTEXT_LISTENER: yes
    depends_on:
      - zookeeper
```

### Stacktrace / full error message

```
ERROR: The Compose file './docker-compose.yml' is invalid because:
services.kafka.environment.ALLOW_PLAINTEXT_LISTENER contains true, which is an invalid type, it should be a string, number, or a null
```

fernando-s97:
Any updates/workarounds on that?

CafeLungo:
I think there is something similar found here after a pre-commit yaml formatter has its way: https://github.com/macisamuele/language-formatters-pre-commit-hooks/issues/53#issuecomment-813623888

The `restart` policy is interpreting `no` as a boolean rather than a string.

PR #9596, opened 2022-06-25
(#8097) fix: [yaml 1.2.2] boolean (true, false)

seunggabi:
**What I did**
- https://yaml.org/spec/1.2.2/#10212-boolean
- yaml boolean type's values: `(true|false)`

**Related issue**

- #8097 


seunggabi:
https://github.com/docker/compose/pull/9598#issuecomment-1167305603

resolved in `v2`

@glours thx

PR #9609, opened 2022-06-28
e2e: split out `pause` tests and add more cases

milas:
**What I did**
Pause/unpause was being partially tested under the start/stop test.
This removes it from that test and adds dedicated pause + unpause
tests.

Note that the tests assert on current behavior, though it's been
noted where that is undesirable due to divergence from the Docker
CLI. Will change the behavior + update tests in a subsequent PR.

**Related Issues**
* EN-83


nicksieger (approved):
LGTM, like the markers for improving functionality later!

PR #9612, opened 2022-06-30
e2e: add more start/stop test cases

milas:
**What I did**
Added some `compose start` / `compose stop` test cases:
* Starting a service that's already running
* Stopping a service that's already stopped
* Starting/stopping multiple services (by name) at once

Also renamed a test that was about `up` behavior but was
misleadingly labeled start/stop.

**Related issue**
* EN-83

groundhog eating an apple


glours (approved):
LGTM

PR #9620, opened 2022-07-02
Recover: bash_completion/docker-compose

Zhou-Tx:
**What I did**
Recover: bash_completion/docker-compose

**Related issue**

### DIFF
--- a/contrib/completion/bash/docker-compose
+++ b/contrib/completion/bash/docker-compose
@@ -1,0 +1,665 @@
+#!/bin/bash
+#
+# bash completion for docker-compose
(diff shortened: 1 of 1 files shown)


milas:
Compose v2 uses Cobra for CLI flag management, so we're going to use its built-in support for advanced shell completion - see #9269.

Note: since v2 is now a Docker plugin (`docker compose` instead of `docker-compose`), there's still more work to be done here to connect shell completion.

DreamShaded:
@Zhou-Tx thx a  lot, it works for me!

Issue #9608, opened 2022-06-28
Compose 2.6.0+ reads .env files from unexpected locations


chrkr:
Docker compose v2.6.0+ has started reading additional .env files from other locations than the first one found. 2.5.1 and earlier read the file only from the first location found.

**Steps to reproduce the issue:**
1. Create a directory `project`
2. Create a subdirectory `project/sub`
3. In `sub` create a compose.yaml file (with any services you like)
4. In `sub` create a `.env` file, with `COMPOSE_FILE=compose.yaml`
5. In `project` create a `.env` file with `COMPOSE_FILE=sub/compose.yaml`
6. In `project` run `docker compose ps` (or any other compose command)

**Describe the results you received:**
I get the error:

stat /home/<my user account>/project/compose.yaml: no such file or directory

Compose tried to read the compose file from the `COMPOSE_FILE` variable as it is set in the `.env` file in the `sub` directory (i.e. `COMPOSE_FILE=compose.yaml`, even though it has previously read the `.env` file from the working directory (`project`), and initially it did set the `COMPOSE_FILE` variable as it was in that file (i.e. `COMPOSE_FILE=sub/compose.yaml`).

**Describe the results you expected:**
I expected that the `.env` file in the location of the first `COMPOSE_FILE` is not read at all, as it was in 2.5.1, especially since compose _has already read a `.env` file_. Thus compose reads the `.env` file in the current directory, sees that it needs to read `sub/compose.yaml` and correctly runs the given command with that file.

Compose should _not_ read more than one `.env` file.

**Additional information you deem important (e.g. issue happens only occasionally):**
As I wrote, this expected behaviour has been observed throughout the v2 series, until it broke in 2.6.0.

This does not just affect the `COMPOSE_FILE` variable, but all variables found in the `.env` files (e.g. `COMPOSE_PROJECT_NAME` etc).

The workaround is to run `docker compose --env-file .env <command>`. Then _only_ the given `.env` file is read, the extra `.env` file is _not_ read.

I use this in a project where there is an customized override file in a given directory, along with a `.env` file which specifies the 'main' compose file(s) as well as the extra one. My `.env` file looks like:
```
COMPOSE_PROJECT_NAME=tests-tds-$CI_PIPELINE_ID
COMPOSE_FILE=../../docker-compose.yml:../../docker-compose.dev.yml:docker-compose.yml
```
[...]

Issue #9578, opened 2022-06-20
docker-compose overrides values from .env with host value if present in host's environment variables


Tim-Roy:
**Description**

docker-compose env_file overwrites values from .env with host value if present in host's environment variables. Otherwise correctly applies the value from .env file

**Steps to reproduce the issue:**
1. Create a variable that has a set value on the host system
2. Create a .env file with the same key but different value
3. include in a service in the docker-compose.yml the following
env_file:
      - .env
4. Print container env variables after running docker-compose up --build -d

**Describe the results you received:**
 The value for the env var is the same as the host

**Describe the results you expected:**
 The value for the env var from the .env file

**Additional information you deem important (e.g. issue happens only occasionally):**
I downgraded docker-compose to v.2.5.1 and this issue is resolved

**Output of `docker compose version`:**

```
v2.6.0
```

**Output of `docker info`:**

```
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc., v0.8.2-docker)

Server:
 Containers: 0
  Running: 0
  Paused: 0
  Stopped: 0
 Images: 0
 Server Version: 20.10.16
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: false
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 io.containerd.runtime.v1.linux runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 96df0994faabc1944fc614e52b0b3c6feb609a57.m
 runc version: 
 init version: de40ad0
 Security Options:
  apparmor
  seccomp
   Profile: default
  cgroupns
 Kernel Version: 5.17.15-1-MANJARO
 Operating System: Manjaro Linux
 OSType: linux
 Architecture: x86_64
 CPUs: 24
 Total Memory: 31.29GiB
 ID: RU7W:L7AD:6ZWZ:GP6M:YANE:BFLX:ZYHD:TVET:TW4V:7CKW:DODE:6FZL
 Docker Root Dir: /var/lib/docker
 Debug Mode: false
 Registry: https://index.docker.io/v1/
 Labels:
[...]

Issue #9521, opened 2022-06-01
env_file variable values overwritten by .env values since v2.6.0


b3n3d1k7:
**Description**

Environment variable values specified via env_file in a docker_compose.yml are overwritten by the values in a .env file since v2.6.0.

**Steps to reproduce the issue:**
1. Create the following files
```
# .env
TEST=.env

# .env.test
TEST=.env.test

# docker-compose.yml
version: '3'
services:
  test:
    image: 'node:6-alpine'
    env_file:
      - .env.test
```
3. Run docker-compose config with docker-compose v2.6.0

**Describe the results you received:**
```
name: docker-compose-test
services:
  test:
    environment:
      TEST: .env
    image: node:6-alpine
    networks:
      default: null
networks:
  default:
    name: docker-compose-test_default
```

**Describe the results you expected:**
This is the result of running the docker-compose config command with e.g. docker-compose v2.5.1:
```
name: docker-compose-test
services:
  test:
    environment:
      TEST: .test.env
    image: node:6-alpine
    networks:
      default: null
networks:
  default:
    name: docker-compose-test_default
```

**Additional information you deem important (e.g. issue happens only occasionally):**
If the environment variables are defined directly in the docker-compose.yml instead of via an env_file, they are not overwritten by the values in the .env file.

**Output of `docker compose version`:**

```
Docker Compose version v2.6.0
```

**Output of `docker info`:**

```
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc., v0.8.2)
  compose: Docker Compose (Docker Inc., v2.6.0)

Server:
 Containers: 3
  Running: 2
  Paused: 0
  Stopped: 1
 Images: 375
 Server Version: 20.10.14
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 1
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 io.containerd.runtime.v1.linux runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 3df54a852345ae127d1fa3092b95168e4a88e2f8
[...]

Issue #9468, opened 2022-05-15
Required variables are not required except the last one in a string


rimelek:
**Description**

About two months ago I discovered a bug in Docker Compose when I wanted to use required environment variables. Everything works as expected when I have only one variable, but when I have multiple variables in a string, only the last one is validated. It is more complex, because it is actually validated and the error message is just overridden by the next non-empty variable.

I also fixed it in "compose-spec/compose-go": https://github.com/compose-spec/compose-go/pull/242

However it was not merged and now the latest development version of "compose-spec/compose-go" is not compatible with Docker Compose. I don't exactly know the relation between the compose and compose-go developer teams, so I don't know what the next step could be, but I would like Docker Compose v2 to work properly so I offer any help I can give you. In the end, people will see and remember bugs in Docker Compose and not in its modules.

I create this issue because even if my pull request were merged, considering the current branches and the process of the development, you could not apply that in Docker Compose until you make Docker Compose compatible with the new changes in compose-go in which ExtraHosts is not `[]string` anymore but `type HostsList map[string]string` . You might have a better idea how it could be fixed in Compose v2 as well without waiting for an other minor version or you can assure us that a new minor version with that fix could be released soon.

If I can see that bugfixes like these can happen, I am happy to continue to work on other bugs as well if it helps.

**Steps to reproduce the issue:**

Use a compose file like this:

```yaml
services:
  bash:
    image: bash:5.1
    command:
      - bash
      - -c
      - "echo VARIABLES: ${COMPOSE_TEST_A:?A is empty} ${COMPOSE_TEST_B:?B is empty} ${COMPOSE_TEST_C:?C is empty}"
```
with this .env file:
```
COMPOSE_TEST_A=A
COMPOSE_TEST_B=B
COMPOSE_TEST_C=C
```

Run

```
docker compose up
COMPOSE_TEST_B="" docker compose up
COMPOSE_TEST_C="" docker compose up
```

**Describe the results you received:**

The output of the above example is:

```text
[+] Running 1/0
 ⠿ Container compose-test-bash-1  Recreated                                                                                                                                                                                        0.0s
Attaching to compose-test-bash-1
compose-test-bash-1  | VARIABLES: A B C
compose-test-bash-1 exited with code 0

[+] Running 1/0
 ⠿ Container compose-test-bash-1  Created                                                                                                                                                                                          0.0s
Attaching to compose-test-bash-1
compose-test-bash-1  | VARIABLES: A C
[...]

PR #9636, opened 2022-07-08
Environment variables priority

ulyssessouza:
**What I did**
This PR is meant to define the precedence of the environment variables evaluation.

For the environment variables to be **available in the container**:
1. Command Line (`docker compose run --env <KEY[=VAL]>` https://docs.docker.com/engine/reference/commandline/compose_run/#options)
2. Compose File (`service::environment` section: https://docs.docker.com/compose/compose-file/#environment)
3. Compose File (`service::env_file` section file: https://docs.docker.com/compose/compose-file/#env_file)
4. Container Image `ENV` directive (https://docs.docker.com/engine/reference/builder/#env)

---

For the environment variables available for **`docker compose`'s runtime**:
1. Command Line parent command option `docker compose --env-file <FILE>` (or `.env` in the root of the project by default)
2. OS Environment Variables

Obs: Note that the variables available for runtime will not be available in the container unless it's imported by one of the methods on the list above

**Related issue**
Resolves https://github.com/docker/compose/issues/9521
Resolves https://github.com/docker/compose/issues/9638
Resolves https://github.com/docker/compose/issues/9608
Resolves https://github.com/docker/compose/issues/9578
Resolves https://github.com/docker/compose/issues/9468
Resolves https://github.com/docker/compose/issues/9683

Depends on https://github.com/compose-spec/compose-go/pull/280

Documentation update [PR](https://github.com/docker/docker.github.io/pull/15161)

beautiful-goat

### DIFF
--- a/cmd/compose/compose.go
+++ b/cmd/compose/compose.go
@@ -354,10 +354,8 @@ func setEnvWithDotEnv(prjOpts *projectOptions) error {
 		return err
 	}
 	for k, v := range envFromFile {
-		if _, ok := os.LookupEnv(k); !ok {
-			if err := os.Setenv(k, v); err != nil {
-				return err
-			}
+		if err := os.Setenv(k, v); err != nil { // overwrite the process env with merged OS + env file results
+			return err
 		}
 	}
 	return nil
--- a/go.mod
+++ b/go.mod
@@ -151,3 +151,5 @@ replace (
 	k8s.io/apimachinery => k8s.io/apimachinery v0.22.4
 	k8s.io/client-go => k8s.io/client-go v0.22.4
 )
+
+replace github.com/compose-spec/compose-go => github.com/compose-spec/compose-go v1.2.9-0.20220712021117-dcfe0889b601 // TODO Remove and bump require section when PR is merged
--- a/pkg/compose/run.go
+++ b/pkg/compose/run.go
@@ -114,15 +114,12 @@ func applyRunOptions(project *types.Project, service *types.ServiceConfig, opts 
 		service.Entrypoint = opts.Entrypoint
 	}
 	if len(opts.Environment) > 0 {
-		env := types.NewMappingWithEquals(opts.Environment)
-		projectEnv := env.Resolve(func(s string) (string, bool) {
-			if _, ok := service.Environment[s]; ok {
-				return "", false
-			}
+		cmdEnv := types.NewMappingWithEquals(opts.Environment)
+		serviceOverrideEnv := cmdEnv.Resolve(func(s string) (string, bool) {
 			v, ok := project.Environment[s]
 			return v, ok
 		}).RemoveEmpty()
-		service.Environment.OverrideBy(projectEnv)
+		service.Environment.OverrideBy(serviceOverrideEnv)
 	}
 	for k, v := range opts.Labels {
 		service.Labels = service.Labels.Add(k, v)
--- a/pkg/e2e/compose_environment_test.go
+++ b/pkg/e2e/compose_environment_test.go
@@ -84,6 +84,19 @@ func TestEnvPriority(t *testing.T) {
 	// 3. Environment file <-- Result expected
 	// 4. Dockerfile
 	// 5. Variable is not defined
+	t.Run("override env file from compose", func(t *testing.T) {
+		res := c.RunDockerComposeCmd("-f", "./fixtures/environment/env-priority/compose-with-env-file.yaml",
+			"--project-directory", projectDir,
+			"run", "--rm", "-e", "WHEREAMI", "env-compose-priority")
+		assert.Equal(t, strings.TrimSpace(res.Stdout()), "override")
+	})
+
+	//  No Compose file & no env variable but override env file
+	// 1. Compose file
+	// 2. Shell environment variables
+	// 3. Environment file <-- Result expected
+	// 4. Dockerfile
+	// 5. Variable is not defined
 	t.Run("override env file", func(t *testing.T) {
 		res := c.RunDockerComposeCmd(t, "-f", "./fixtures/environment/env-priority/compose.yaml", "--project-directory",
 			projectDir, "--env-file", "./fixtures/environment/env-priority/.env.override", "run", "--rm", "-e",
--- a/pkg/e2e/compose_environment_test.go
+++ b/pkg/e2e/compose_environment_test.go
@@ -27,150 +27,196 @@ import (
 func TestEnvPriority(t *testing.T) {
 	c := NewParallelCLI(t)
 
-	projectDir := "./fixtures/environment/env-priority"
-
 	t.Run("up", func(t *testing.T) {
 		c.RunDockerOrExitError(t, "rmi", "env-compose-priority")
 		c.RunDockerComposeCmd(t, "-f", "./fixtures/environment/env-priority/compose-with-env.yaml",
-			"--project-directory", projectDir, "up", "-d", "--build")
+			"up", "-d", "--build")
 	})
 
 	// Full options activated
-	// 1. Compose file <-- Result expected
-	// 2. Shell environment variables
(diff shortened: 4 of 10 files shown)


go.mod
```diff
@@ -151,3 +151,5 @@ replace (
 	k8s.io/apimachinery => k8s.io/apimachinery v0.22.4
 	k8s.io/client-go => k8s.io/client-go v0.22.4
 )
+
+replace github.com/compose-spec/compose-go => github.com/compose-spec/compose-go v1.2.9-0.20220712021117-dcfe0889b601 // TODO Remove and bump require section when PR is merged
```
ulyssessouza, line 155: Note that it points to https://github.com/compose-spec/compose-go/pull/280

pkg/compose/run.go
```diff
@@ -116,8 +116,13 @@ func applyRunOptions(project *types.Project, service *types.ServiceConfig, opts
...
 		projectEnv := env.Resolve(func(s string) (string, bool) {
-			if _, ok := service.Environment[s]; ok {
-				return "", false
+			cmdLineVar, ok := env[s]
+			if ok && cmdLineVar == nil {
+				v, ok := project.Environment[s]
+				return v, ok
+			}
+			if sEnv, ok := service.Environment[s]; ok && sEnv != nil {
+				return *sEnv, ok
```
milas, line 125: I'm having a hard time following this.

I _THINK_ it should be simplified/shortened to:
```go
v, ok := project.Environment[s]
return v, ok
```

My understanding is that `env` is the `cmdEnv`, and `Resolve` will call your `lookupFn` for any values that have `nil`, right?

So the immediate lookup from from `env` (aka `cmdEnv`) is guaranteed to be `nil`.
As a result, we'll always fall back to returning whatever is in the project (OS/host) env and never attempt to read from the service env here...which I think is what we want!
User did `compose run --env FOO` which explicitly means "read this value from my OS env".

If I'm missing some nuance here, some comments about the why/how you end up in each of these cases and how that relates to precedence would be super helpful.

TL;DR I think this code is _technically_ correct, but very roundabout?

pkg/e2e/compose_environment_test.go
```diff
@@ -27,12 +27,39 @@ import (
...
-			"--project-directory", projectDir, "up", "-d", "--build")
+			"up", "-d", "--build")
+	})
+
+	// Full options activated
+	// 1. Compose file
+	// 2. Shell environment variables <-- Result expected
+	// 3. Environment file
+	// 4. Dockerfile
+	// 5. Variable is not defined
```
milas, line 41: This is great 😻

cmd/compose/compose.go
```diff
@@ -354,10 +354,8 @@ func setEnvWithDotEnv(prjOpts *projectOptions) error {
...
 	}
 	for k, v := range envFromFile {
-		if _, ok := os.LookupEnv(k); !ok {
-			if err := os.Setenv(k, v); err != nil {
-				return err
-			}
+		if err := os.Setenv(k, v); err != nil {
+			return err
 		}
 	}
```
milas, line 360: Nit: Given the complexity/subtlety, I think a comment here is warranted, e.g. `// overwrite the process env with merged OS + env file results` or something

For context: I thought this was a regression at first since OS/shell has higher precedence than env file, but then I realized that `GetEnvFromFile` takes in the OS/shell env and handles the precedence/merging, which is why it's safe/necessary to overwrite everything here

milas (commented):
Wahoo! Looks good for approval once the `compose-go` side of the house is merged + tagged 🥳 

(Sorry for nits - as you're well aware, the env code is pretty complex, so I was trying to identify places I got confused and had to dig in more)

pkg/compose/run.go
ulyssessouza: Ahhh!!! Nice catch!!!
That was an old "bad good idea" of when using `-e UNSETVAR` and it couldn't be found in `project.Environment` it could check in the `service:environment` section of the Compose file.
But... As you said, that's bad! Specially when the user imperatively used the `-e` option to make its usage explicit.

cmd/compose/compose.go
ulyssessouza: Yep! Adding the comment.

pkg/compose/run.go
```diff
@@ -116,8 +116,13 @@ func applyRunOptions(project *types.Project, service *types.ServiceConfig, opts
 	if len(opts.Environment) > 0 {
 		env := types.NewMappingWithEquals(opts.Environment)
 		projectEnv := env.Resolve(func(s string) (string, bool) {
```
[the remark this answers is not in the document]
ulyssessouza: Sure! I felt the same while reading. Didn't change cuz it wasn't part of the PR.
Renaming in 1, 2, 3...

pkg/e2e/compose_environment_test.go
```diff
@@ -27,12 +27,39 @@ import (
...
+	t.Run("compose file priority", func(t *testing.T) {
+		cmd := c.NewDockerComposeCmd(t, "-f", "./fixtures/environment/env-priority/compose-with-env.yaml",
+			"--env-file", "./fixtures/environment/env-priority/.env.override",
+			"run", "--rm", "-e", "WHEREAMI", "env-compose-priority")
```
nicksieger, line 45: So there's a difference between `-e WHEREAMI` with the var defined, and `-e WHEREAMI=shell`? In this case shouldn't `3.` be the expected result?

pkg/e2e/compose_environment_test.go
```diff
@@ -43,121 +70,179 @@ func TestEnvPriority(t *testing.T) {
 	// 5. Variable is not defined
 	t.Run("compose file priority", func(t *testing.T) {
 		cmd := c.NewDockerComposeCmd(t, "-f", "./fixtures/environment/env-priority/compose-with-env.yaml",
-			"--project-directory", projectDir, "--env-file", "./fixtures/environment/env-priority/.env.override", "run",
+			"--env-file", "./fixtures/environment/env-priority/.env.override", "run",
 			"--rm", "-e", "WHEREAMI", "env-compose-priority")
 		cmd.Env = append(cmd.Env, "WHEREAMI=shell")
 		res := icmd.RunCmd(cmd)
-		assert.Equal(t, strings.TrimSpace(res.Stdout()), "Compose File")
+		assert.Equal(t, strings.TrimSpace(res.Stdout()), "override")
```
nicksieger, line 77: Update to `3. Environment file <-- Result expected` above?

nicksieger: Also, these tests so far seem to indicate that `1. Compose file` is not the highest priority. Am I reading this right?

nicksieger (commented):
Overall this looks good, just have some questions about the tests and what the new priority order is.

pkg/e2e/compose_environment_test.go
ulyssessouza: In the first, the variable value will be inherited from the environment. 
The second one will have the explicit value passed (`shell` in this case).

nicksieger (approved):

glours (approved):
LGTM

K0Te:
Hi,

Sorry if this is not the correct place to ask a question, I've checked this PR source code and documentation update and am still not sure what the expected behavior is.
I've used `.env` file to override `COMPOSE_FILE` for local development: `COMPOSE_FILE=docker/docker-compose.yml:docker/docker-compose.override.yml`. This allowed to have all docker stuff in a separate folder and provide some secrets in `docker-compose.override.yml`.
And on CI I've just used `COMPOSE_FILE=docker/docker-compose.yml` as an env variable in GH Actions.

This worked fine as of docker compose 2.7.0, but with docker compose 2.9.0 this workflow doesn't work anymore. Value from `.env` started to take precedence over the environment variable.

Is this a new expected behaviour or a bug?

PR #9637, opened 2022-07-08
Simplify e2e methods

ulyssessouza:
**What I did**
Add `testing.TB` to the struct `e2e.CLI` on e2e tests so it doesn't need to be passed on every method call.

This also let `e2e.CLI` to be passed where a `testing.TB` is expected.

### DIFF
--- a/pkg/e2e/framework.go
+++ b/pkg/e2e/framework.go
@@ -58,6 +58,8 @@ func init() {
 
 // CLI is used to wrap the CLI for end to end testing
 type CLI struct {
+	testing.TB
+
 	// ConfigDir for Docker configuration (set as DOCKER_CONFIG)
 	ConfigDir string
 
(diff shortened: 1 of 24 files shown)


pkg/e2e/framework.go
```diff
@@ -58,6 +58,8 @@ func init() {
 
 // CLI is used to wrap the CLI for end to end testing
 type CLI struct {
+	testing.TB
+
```
ulyssessouza, line 62: Object Orientation vibes here 😎

milas:
Ah so the reason I moved `t` to be an arg in #9561 is because of how we're (mis)using subtests currently:
```go
func TestFoo(t *testing.T) {
  cli := NewCLI()
  t.Run("set up things", func(t *testing.T) {
    cli.Run(t, ...)
            ⬆️
            // if we don't pass `t` here, it'll
            // use the outer closure's `t`, which
            // messes things up when tests fail
  }

  t.Run("more things", func(t* testing.T) { ... }
```

(Personally, I'd be happy if we moved away from using subtests like this, since it breaks other things, e.g. `-shuffle` to randomize test order, and is not particularly idiomatic Go from my experience. But that's a bigger change 😬)

PR #9639, opened 2022-07-09
Link to BUILDING.md for testing instructions

ikedam:
**What I did**

Link to BUILDING.md for testing instructions, which was linked to README.md.
(BUILDING.md wasn't linked from anywhere)

**Related issue**

#9439

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -124,9 +124,10 @@ Fork the repository and make changes on your fork in a feature branch:
     issue.
 
 Submit unit tests for your changes. Go has a great test framework built in; use
-it! Take a look at existing tests for inspiration. [Run the full test
-suite](BUILDING.md) on your branch before
-submitting a pull request.
+it! Take a look at existing tests for inspiration. Also end-to-end tests are
+available. Run the full test suite, both unit tests and e2e tests on your
+branch before submitting a pull request. See [BUILDING.md](BUILDING.md) for
+instructions to build and run tests.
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -125,7 +125,7 @@ Fork the repository and make changes on your fork in a feature branch:
 
 Submit unit tests for your changes. Go has a great test framework built in; use
 it! Take a look at existing tests for inspiration. [Run the full test
-suite](README.md) on your branch before
+suite](BUILDING.md) on your branch before
```
glours, line 128: Why not link directly to the right section?
```suggestion
suite](BUILDING.md#end-to-end-tests) on your branch before
```

ikedam: Contributors are expected to run both unit tests and e2e tests. Linking to the whole document makes sense for that reason.

Which is the best?:

* A: Link to `BUILDING.md#unit-tests` as this section describes "Submit unit tests".
* B: Rewrite this section. E.g. `Run the full test suite, both unit tests and e2e tests on your branch before submitting a pull request. See BUILDING.md for instructions to build and run tests.`

nicksieger: I'm fine with the change as-is, it's a small improvement. I do like the rewrite option B that @ikedam puts here too.

ikedam: Thanks for the comment.
I added a new commit updating this paragraph as option B.

glours (approved):

Issue #9630, opened 2022-07-07
Named network is not created anymore


davidspiess:
After upgrading Docker Desktop from 4.9.1 to 4.10.1 `docker compose up` fails with 

`Error response from daemon: network hgv not found`.
Before the upgrade the external network was created automatically if it was not found.

```yaml
services:
  api:
    container_name: awesome-api
    build:
      context: ./api
    networks:
      - hgv
      - default

networks:
  hgv:
    name: hgv
```

**Output of `docker compose version`:**

```
Docker Compose version v2.6.1
```

**Output of `docker info`:**

```
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc., v0.8.2)
  compose: Docker Compose (Docker Inc., v2.6.1)
  extension: Manages Docker extensions (Docker Inc., v0.2.7)
  sbom: View the packaged-based Software Bill Of Materials (SBOM) for an image (Anchore Inc., 0.6.0)
  scan: Docker Scan (Docker Inc., v0.17.0)

Server:
 Containers: 4
  Running: 2
  Paused: 0
  Stopped: 2
 Images: 13
 Server Version: 20.10.17
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 1
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 io.containerd.runtime.v1.linux runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 10c12954828e7c7c9b6e0ea9b0c02b01407d3ae1
 runc version: v1.1.2-0-ga916309
 init version: de40ad0
 Security Options:
  seccomp
   Profile: default
 Kernel Version: 5.4.72-microsoft-standard-WSL2
 Operating System: Docker Desktop
 OSType: linux
 Architecture: x86_64
 CPUs: 16
 Total Memory: 24.85GiB
 Name: docker-desktop
 ID: NNVH:TNPI:DJPP:VF6X:WZYT:QAUC:DLB5:GSA7:OGHR:7FKN:3RK3:U6JV
 Docker Root Dir: /var/lib/docker
[...]

PR #9647, opened 2022-07-13
add 2 named networks with close name in the networks test suite to gu…

glours:
…arantee the good distinction on network names

**What I did**
Add a 2 new networks with a close name to the network e2e test suite to detect any regression like issue #9630 

**Related issue**
issue #9630 

image


laurazard (approved):
tests! 🎉

PR #9158, opened 2022-02-09
down: Reject all arguments

Jille:
The down command silently ignored all arguments, which might cause
confusion and/or outages if someone expects `docker-compose down
$service` to be the opposite of `docker-compose up $service`, rather
than turning down everything.

**What I did**
Make docker-compose down reject any given arguments.

**Related issue**
closes #9151

PR #9660, opened 2022-07-16
Enforce consistent args rejection for no-args commands: ls, restart, version

piroux:
**What I did**
Make cobra to reject arguments for no-args commands in the same manner:
- `ls`
- `version`
- (`down` was done already) 

For the `restart` command, the documentation is updated to let the user know that services can be passed as arguments.

**Related issue**

Based on #9158 which enforced it for the `down` command.

---
<details>
<summary>According to the listing below, these 4 commands were the only ones to not handle arguments.</summary>

```
➜ for c in (docker compose --help | awk '/^  [a-z]+ +.*/ {print $1}');
      docker compose $c --help | grep 'Usage:';
  end | cat
Usage:  docker compose build [SERVICE...]
Usage:  docker compose convert SERVICES
Usage:  docker compose cp [OPTIONS] SERVICE:SRC_PATH DEST_PATH|-
Usage:  docker compose create [SERVICE...]
Usage:  docker compose down
Usage:  docker compose events [options] [--] [SERVICE...]
Usage:  docker compose exec [options] [-e KEY=VAL...] [--] SERVICE COMMAND [ARGS...]
Usage:  docker compose images [SERVICE...]
Usage:  docker compose kill [options] [SERVICE...]
Usage:  docker compose logs [SERVICE...]
Usage:  docker compose ls
Usage:  docker compose pause [SERVICE...]
Usage:  docker compose port [options] [--] SERVICE PRIVATE_PORT
Usage:  docker compose ps [SERVICE...]
Usage:  docker compose pull [SERVICE...]
Usage:  docker compose push [SERVICE...]
Usage:  docker compose restart
Usage:  docker compose rm [SERVICE...]
Usage:  docker compose run [options] [-v VOLUME...] [-p PORT...] [-e KEY=VAL...] [-l KEY=VALUE...] SERVICE [COMMAND] [ARGS...]
Usage:  docker compose start [SERVICE...]
Usage:  docker compose stop [SERVICE...]
Usage:  docker compose top [SERVICES...]
Usage:  docker compose unpause [SERVICE...]
Usage:  docker compose up [SERVICE...]
Usage:  docker compose version
```

<br>
</details>

### DIFF
--- a/cmd/compose/restart.go
+++ b/cmd/compose/restart.go
@@ -35,8 +35,8 @@ func restartCommand(p *projectOptions, backend api.Service) *cobra.Command {
 		projectOptions: p,
 	}
 	restartCmd := &cobra.Command{
-		Use:   "restart",
-		Short: "Restart containers",
+		Use:   "restart [SERVICE...]",
+		Short: "Restart service containers",
 		RunE: Adapt(func(ctx context.Context, args []string) error {
 			return runRestart(ctx, backend, opts, args)
 		}),
--- a/docs/reference/compose_restart.md
+++ b/docs/reference/compose_restart.md
@@ -1,7 +1,7 @@
 # docker compose restart
 
 <!---MARKER_GEN_START-->
-Restart containers
+Restart service containers
 
 ### Options
 
--- a/docs/reference/compose_restart.md
+++ b/docs/reference/compose_restart.md
@@ -14,7 +14,7 @@ Restart containers
 
 ## Description
 
-Restarts all stopped and running services.
+Restarts all stopped and running services, or the specified services only.
 
 If you make changes to your `compose.yml` configuration, these changes are not reflected
 after running this command. For example, changes to environment variables (which are added
--- a/docs/reference/docker_compose_restart.yaml
+++ b/docs/reference/docker_compose_restart.yaml
@@ -1,7 +1,7 @@
 command: docker compose restart
-short: Restart containers
+short: Restart service containers
 long: |-
-    Restarts all stopped and running services.
+    Restarts all stopped and running services, or the specified services only.
 
     If you make changes to your `compose.yml` configuration, these changes are not reflected
     after running this command. For example, changes to environment variables (which are added
@@ -11,7 +11,7 @@ long: |-
     If you are looking to configure a service's restart policy, please refer to
     [restart](https://github.com/compose-spec/compose-spec/blob/master/spec.md#restart)
     or [restart_policy](https://github.com/compose-spec/compose-spec/blob/master/deploy.md#restart_policy).
-usage: docker compose restart
+usage: docker compose restart [SERVICE...]
 pname: docker compose
 plink: docker_compose.yaml
 options:
--- a/cmd/compose/list.go
+++ b/cmd/compose/list.go
@@ -46,6 +46,7 @@ func listCommand(backend api.Service) *cobra.Command {
 		RunE: Adapt(func(ctx context.Context, args []string) error {
 			return runList(ctx, backend, lsOpts)
 		}),
+		Args:              cobra.NoArgs,
 		ValidArgsFunction: noCompletion(),
 	}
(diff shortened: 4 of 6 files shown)


ulyssessouza (approved):
LGTM

cmd/compose/restart.go
```diff
@@ -40,6 +40,7 @@ func restartCommand(p *projectOptions, backend api.Service) *cobra.Command {
 		RunE: Adapt(func(ctx context.Context, args []string) error {
 			return runRestart(ctx, backend, opts, args)
 		}),
+		Args:              cobra.NoArgs,
```
glours, line 43: Ok I think we should update the documentation! Restart allow you to choose a specific service to restart so you can pass an argument like that:
```shell
docker compose restart my-service
```

glours (changes requested):
I let you choose what you want to do:

- Update the restart documentation in your PR
- Remove your changes for the `restart` command

cmd/compose/restart.go
laurazard: We shouldn't change this, right? It implies losing the ability to just restart a specific service, which people might be relying on.

glours: We'll keep the actual behavior for sure, that's why I propose to update the doc or remove the modification for the `restart` command

piroux: Sorry, I had overlooked that `restart` could be used like that too.

So yes, let's update the docs.

glours:
You need to resolve your conflict in order to start the CI process

docs/reference/compose_restart.md
```diff
@@ -1,7 +1,7 @@
 # docker compose restart
 
 <!---MARKER_GEN_START-->
-Restart containers
+Restart service containers
```
glours, line 4: This change should be done directly [here](https://github.com/docker/compose/blob/def189fae1a74a9cac327781aabeaa214d6906a9/cmd/compose/restart.go#L39)

docs/reference/docker_compose_restart.yaml
```diff
@@ -1,7 +1,7 @@
 command: docker compose restart
-short: Restart containers
+short: Restart service containers
```
glours, line 2: this modification will auto generated

docs/reference/docker_compose_restart.yaml
```diff
@@ -1,7 +1,7 @@
...
 long: |-
-    Restarts all stopped and running services.
+    Restarts all stopped and running services, or the specified services only.
```
glours, line 4: this one also

docs/reference/docker_compose_restart.yaml
```diff
@@ -11,7 +11,7 @@ long: |-
     If you are looking to configure a service's restart policy, please refer to
     [restart](https://github.com/compose-spec/compose-spec/blob/master/spec.md#restart)
     or [restart_policy](https://github.com/compose-spec/compose-spec/blob/master/deploy.md#restart_policy).
-usage: docker compose restart
+usage: docker compose restart [SERVICE...]
```
glours, line 14: This change should be done [here](https://github.com/docker/compose/blob/def189fae1a74a9cac327781aabeaa214d6906a9/cmd/compose/restart.go#L38)

docs/reference/compose_restart.md
```diff
@@ -14,7 +14,7 @@ Restart containers
 
 ## Description
 
-Restarts all stopped and running services.
+Restarts all stopped and running services, or the specified services only.
```
glours, line 17: Keep this one 👍

glours (changes requested):
Doing modification in the documentation isn't trivial 😅
You should modify the command itself in the `restart.go` file and then use `make docs` to generate the yaml and md doc files.

glours (approved):
LGTM

milas (approved):

PR #9673, opened 2022-07-22
remove unused sessionConfig param for addSecretsConfig function

glours:
**What I did**
Fix a linter issue with an unused parameter


nicksieger (approved):
👍

PR #9675, opened 2022-07-25
use env variable for golang version and updates gh actions from v2 to v3

glours:
**What I did**
Update the GitHub actions workflows:
* use an env variable to define the version of Golang used
* update GitHub base actions to v3

image

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,8 @@ on:
         description: 'To run with tmate enter "debug_enabled"'
         required: false
         default: "false"
-
+env:
+  GO_VERSION: 1.18.4
 jobs:
   lint:
     name: Lint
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -11,22 +11,23 @@ on:
...
         default: "false"
-
+env:
+  GO_VERSION: 1.18.4
 jobs:
   lint:
     name: Lint
     runs-on: ubuntu-latest
     env:
       GO111MODULE: "on"
```
nicksieger, line 21: We can remove GO111MODULE everywhere now, right?

nicksieger (approved):
LGTM, nice cleanup with the newer actions!

.github/workflows/ci.yml
glours: Right!

StefanScherer (approved):
LGTM

PR #9677, opened 2022-07-26
use Google addlicense instead of kunalkushwaha/ltag

glours:
**What I did**
Use the same license header checker that the one use for `compose-cli` project, this one is much more active than the previous one.

**Related issue**

image


StefanScherer (approved):
LGTM

PR #9700, opened 2022-08-01
update to compose-go v1.4.0 as previous version introduced breaking changes

glours:
Signed-off-by: Guillaume Lours <guillaume.lours@docker.com>

**What I did**
Bump version of compose-go to `v1.4.0` which remove the breaking changes introduced in version `v1.3.0` by this [PR](https://github.com/compose-spec/compose-go/pull/294)

image

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,7 +6,7 @@ require (
 	github.com/AlecAivazis/survey/v2 v2.3.5
 	github.com/buger/goterm v1.0.4
 	github.com/cnabio/cnab-to-oci v0.3.5
-	github.com/compose-spec/compose-go v1.3.0
+	github.com/compose-spec/compose-go v1.4.0
 	github.com/containerd/console v1.0.3
 	github.com/containerd/containerd v1.6.6
 	github.com/distribution/distribution/v3 v3.0.0-20220725133111-4bf3547399eb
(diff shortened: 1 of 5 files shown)


ndeloof (approved):

nicksieger (approved):
LGTM

Jyrno42:
I am having some trouble with the updated compose version 2.9.0 where this fix should be applied.

Using the following example dockerfile to demonstrate the issue:

```Dockerfile
FROM docker:latest

RUN mkdir /app
WORKDIR /app

ENV BUILDKIT_PROGRESS=plain

RUN echo -e "\
services:\n\
  example:\n\
    build:\n\
      context: .\n\
      dockerfile: Dockerfile\n\
    command: -text="hello world"\n\
" > docker-compose.yml

RUN echo -e "\
FROM hashicorp/http-echo\n\
ENV FOO=BAR" > Dockerfile

RUN echo -e "\
docker-compose --version \n\
docker compose version \n\
docker compose build \n\
docker-compose build" > /cmd.sh && chmod +x /cmd.sh

CMD ["/cmd.sh"]
```

And building/running it with:

```
docker build -f Dockerfile . -t test-bug
docker run --rm -it --privileged -v /var/run/docker.sock:/var/run/docker.sock test-bug
```

I am observing the following (truncated) output:

```
Docker Compose version v2.9.0
Docker Compose version v2.9.0
#5 naming to docker.io/library/app-example done
#5 naming to docker.io/library/app-example done
```

E.g. the separator used is still `-`.

However when I run the same image with the `COMPOSE_COMPATIBILITY=1` env variable I am seeing:

```
Docker Compose version v2.9.0
Docker Compose version v2.9.0
#5 naming to docker.io/library/app_example done
#5 naming to docker.io/library/app_example done
```

Is there something I am doing wrong? 

Note: I don't think the issue is with the official docker image as I am observing a similar behavior with compose  v2.9.0 on my arch linux installation.

Jyrno42:
I can also open a new issue about this if needed.

glours:
Hi @Jyrno42 
The breaking change impacted resources such as networks, volumes, secrets and config. We keep the the `-` by default for the container names

Jyrno42:
Oh, OK. Thanks for the clarifications. Then I know its safe to update my scripts without fearing a revert there.

PR #9702, opened 2022-08-01
Generate SLSA provenance for released binaries

laurentsimon:
Hi,

I'm reaching out on behalf of the [Open Source Security Foundation (openssf.org)](https://openssf.org/). We work on improving the security of critical open source projects like yours. 

Together with [GitHub](https://github.blog/2022-04-07-slsa-3-compliance-with-github-actions/), we designed a free, easy-to-use method of code signing. It will help your users verify that your release binaries were built from your repository’s [workflow](https://github.com/docker/compose/blob/v2/.github/workflows/release.yaml) and not altered by anyone. It’s just a few lines of code, but it will make your project more secure against third-party tampering and attacks like [Codecov](https://about.codecov.io/apr-2021-post-mortem/) and [CTX](https://sockpuppets.medium.com/how-i-hacked-ctx-and-phpass-modules-656638c6ec5e).

This PR shows how to add this seamless code signing to your workflow. You don’t have to be a cryptography expert or learn complicated tools and [verification](https://github.com/slsa-framework/slsa-verifier/blob/main/README.md#verification-of-provenance) is simple for your users.  

You can read more on the [SLSA blog](https://slsa.dev/blog/2022/06/slsa-github-workflows). Please reach out if you have any questions!


.github/workflows/release.yaml
```diff
@@ -30,14 +32,50 @@ jobs:
         run: make GIT_TAG=${{ github.event.inputs.tag }} -f builder.Makefile cross
 
       - name: Compute checksums
-        run: cd bin; for f in *; do shasum --binary --algorithm 256 $f | tee -a checksums.txt > $f.sha256; done
+        working-directory: bin/
```
laurentsimon, line 35: note: this change is not needed. I added `working-directory` to avoid the `cd bin`, but I can revert.

laurentsimon:
friendly ping. Are there any question I could answer?

.github/workflows/release.yaml
```diff
@@ -9,7 +9,9 @@ on:
 env:
```
nicksieger, line 9: Would it make sense to have a "dry run" parameter that builds and signs the releases and verifies them, just so we can test the signing workflow?

```suggestion
      dry-run:
        description: "Build, sign, verify but do not release"
        required: false
        type: boolean
        default: true
env:
```

.github/workflows/release.yaml
```diff
@@ -30,14 +32,50 @@ jobs:
...
+    steps:
+      # Download binaries and license.
+      - uses: actions/download-artifact@v3
+        with:
+          name: "bin"
+      # Download provenance file.
+      - uses: actions/download-artifact@v3
+        with:
+          name: "${{ needs.provenance.outputs.attestation-name }}"
+      # Upload to release.
```
nicksieger, line 75:
```suggestion
      # Verify binaries if dry run
      - name: Verify binaries
        if: 'inputs.dry-run' 
        run: |
          echo TODO: verify with slsa-verifier
      # Upload to release.
```

.github/workflows/release.yaml
```diff
@@ -30,14 +32,50 @@ jobs:
...
       - uses: ncipollo/release-action@v1
         with:
```
nicksieger, line 77:
```suggestion
        if: '! inputs.dry-run'
        with:
```

nicksieger (approved):
LGTM, thanks for setting this up.

laurentsimon:
Thanks @nicksieger !

.github/workflows/release.yaml
```diff
@@ -30,14 +37,56 @@ jobs:
...
+      - uses: actions/download-artifact@v3
+        with:
+          name: "bin"
+      # Download provenance file.
+      - uses: actions/download-artifact@v3
+        with:
+          name: "${{ needs.provenance.outputs.attestation-name }}"
+      # Verify binaries if dry run
+      - name: Verify binaries
+        if: 'inputs.dry-run'
```
asraa, line 82: why only on dry-run?

.github/workflows/release.yaml
nicksieger: I guess we need to put in the actual steps here. Don't suppose you have a packaged action for verifying yet?

.github/workflows/release.yaml
asraa: it seems like SLSA provenance is generated either way

nicksieger: It doesn't only have to be on a dry run. On a real release it's theoretically extra work, but might still be nice to make sure the attestation loop is complete?

.github/workflows/release.yaml
laurentsimon: Not yet, but we are working on it, coming in the next release. 
Would you like me to add the steps in this PR or do that in a follow-up PR?
We have 2 options:
- `go install ...@vx.y`. This will build from source. Dependabot and removatebot don't parse commands, so you won't receive updates. This may be acceptable temporarily.
- pull the binary from our release. Can verify its hash as well, which are committed to the repo and reviews are enforced https://github.com/slsa-framework/slsa-verifier/blob/main/SHA256SUM.md. The verification will part of a different job (after the generation), and we can make this job have just `permissions: none` and not worry about verifying hashes - since a compromise would not affect the workflow / repository with no permissions

Let me know what you prefer.

laurentsimon: The cheapest option is to use the pre-compiled binary. But if you don't mind 2-3mn to build the verifier, it's the simplest option.

laurentsimon: I added the code for verification. PTAL

I also created a tracking issue https://github.com/slsa-framework/slsa-verifier/issues/206 on our repo as an AI to send a PR to update your workflow when we have the GHA released.

.github/workflows/release.yaml
```diff
@@ -30,14 +41,92 @@ jobs:
...
+          COMPUTED_HASH=$(sha256sum slsa-verifier-linux-amd64 | cut -d ' ' -f1)
+          EXPECTED_HASH="37db23392c7918bb4e243cdb097ed5f9d14b9b965dc1905b25bc2d1c0c91bf3d"
+          if [[ "$EXPECTED_HASH" != "$COMPUTED_HASH" ]];then
+              echo "error: expected $EXPECTED_HASH, computed $COMPUTED_HASH"
+              exit 1
+          fi
+          chmod ug+x slsa-verifier-linux-amd64
+
+          # Verify all files, except LICENSE, .sha256 and checksums.txt files.
+          rm LICENSE *.sha256 checksums.txt
```
laurentsimon, line 107: my bad, this should not be deleted.

.github/workflows/release.yaml
laurentsimon: I've made the changes so that the verification happens in any both cases.

laurentsimon:
friendly ping for feedback.

glours:
Hello @laurentsimon 

Regarding the current exchanges in your buildx [PR](https://github.com/docker/buildx/pull/1243), we're discussing internally to give you a common response quickly. 
If that's fine for you, could we use the buildx PR for all the discussions?

laurentsimon:
> Hello @laurentsimon
> 
> Regarding the current exchanges in your buildx [PR](https://github.com/docker/buildx/pull/1243), we're discussing internally to give you a common response quickly. If that's fine for you, could we use the buildx PR for all the discussions?

Absolutely. SGTM. Happy to answer further questions if you have any.

PR #9706, opened 2022-08-02
update usage strings for consistency

thaJeztah:
This updates the format of various usage strings to be more consistent
with other parts of the CLI.

- Use `[OPTIONS]` to indicate where command-specific options should be added
- Use `[SERVICE...]` to indicate zero-or-more services
- Remove some usage strings for specific options (e.g. `-e NAME=VAL`), as that
  option is part of the already mentioned `[OPTIONS]` and we don't provide usage
  for each possible option that can be passed.
- Remove `[--]`, which (I think) was needed for the Python implementation, but is
  a general feature to stop processing flag-options.

**What I did**

**Related issue**

### DIFF
--- a/cmd/compose/build.go
+++ b/cmd/compose/build.go
@@ -77,7 +77,7 @@ func buildCommand(p *projectOptions, backend api.Service) *cobra.Command {
 		projectOptions: p,
 	}
 	cmd := &cobra.Command{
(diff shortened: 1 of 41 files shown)


ulyssessouza (approved):
LGTM

glours (approved):
LGTM, thanks @thaJeztah

PR #9709, opened 2022-08-02
build: upgrade BuildKit & docker/distribution

milas:
**What I did**
Housekeeping - bumping dependencies to latest version
available.

chipmunk


glours (approved):
LGTM

PR #9711, opened 2022-08-03
Update dockerfile syntax to latest stable, and rename docs Dockerfile

thaJeztah:
**What I did**

### update dockerfiles to use latest stable syntax

Some Dockerfiles were pinned to a minor release, which meant they wouldn't be updated to get the latest stable syntax (and fixes), and one Dockerfile used the "labs" variant to use the HEREDOC syntax, which has now been promoted to the stable syntax.

### docs: rename Dockerfile

There's no other Dockerfiles in the same path, so the "docs" prefix was redundant.

**Related issue**


nicksieger (approved):
👍

milas (approved):

PR #9712, opened 2022-08-03
config: use correct YAML marshal func

milas:
**What I did**
When marshalling via CLI (`compose convert` / `compose config`), we
were using a _different_ YAML lib, which was a fork of `go-yaml`,
which is what `gopkg.in/yaml.v2` is based off of.

The `compose-spec/compose-go` lib is written with `gopkg.in/yaml.v2`
as a target, so I changed it to use that for marshalling the config.

The fork that was being used was [sanathkr/go-yaml](https://github.com/sanathkr/go-yaml).
Looking through `git log`, this was originally introduced transitively by
old ECS/CloudFormation code (that now lives in `compose-cli`), so its
usage was likely the result of IDE auto-complete and not intentional.

The only change it added compared to the upstream lib was custom tag
support, which is not something we use (and regardless, since the **unmarshal**
is done with a different lib, it wouldn't work regardless). See https://github.com/go-yaml/yaml/commit/ed9d249f429b3f5a69f80a7abef6bfce81fef894.

**Related issue**
I discovered this when working on compose-spec/compose-go#298.

a cat in a shark costume

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -27,13 +27,13 @@ require (
 	github.com/opencontainers/go-digest v1.0.0
 	github.com/opencontainers/image-spec v1.0.3-0.20211202183452-c5a74bcca799
 	github.com/pkg/errors v0.9.1
-	github.com/sanathkr/go-yaml v0.0.0-20170819195128-ed9d249f429b
 	github.com/sirupsen/logrus v1.9.0
 	github.com/spf13/cobra v1.5.0
 	github.com/spf13/pflag v1.0.5
 	github.com/stretchr/testify v1.8.0
 	github.com/theupdateframework/notary v0.7.0
 	golang.org/x/sync v0.0.0-20220722155255-886fb9371eb4
+	gopkg.in/yaml.v2 v2.4.0
 	gotest.tools v2.2.0+incompatible
 	gotest.tools/v3 v3.3.0
 )
@@ -120,7 +120,6 @@ require (
 	google.golang.org/grpc v1.45.0 // indirect
 	google.golang.org/protobuf v1.27.1 // indirect
 	gopkg.in/inf.v0 v0.9.1 // indirect
-	gopkg.in/yaml.v2 v2.4.0 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 	k8s.io/apimachinery v0.24.1 // indirect; see replace for the actual version used
 	k8s.io/client-go v0.24.1 // indirect; see replace for the actual version used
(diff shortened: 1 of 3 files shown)


milas:
As a bonus, this shrinks the executable size by ~100KB 🥳

nicksieger (approved):
Good find! 😎

thaJeztah:
Had a quick peek if there was a specific reason for using this library; I see this was added as part of https://github.com/docker/compose/commit/7f8bb030e66a4d8de4b44a61fc18fa60bf9b14ae, which originally came from https://github.com/docker/compose-cli/commit/7f8bb030e66a4d8de4b44a61fc18fa60bf9b14ae (https://github.com/docker/compose-cli/pull/544), but review on that PR doesn't mention a specific reason

/cc @ndeloof @rumpl (just in case)

thaJeztah (approved):
LGTM (not a maintainer)

go.mod
```diff
@@ -120,7 +119,7 @@ require (
 	google.golang.org/grpc v1.45.0 // indirect
 	google.golang.org/protobuf v1.27.1 // indirect
 	gopkg.in/inf.v0 v0.9.1 // indirect
-	gopkg.in/yaml.v2 v2.4.0 // indirect
+	gopkg.in/yaml.v2 v2.4.0
```
ulyssessouza, line 122: Maybe moving this entry to the "direct" section above this block, where the fork was for better visibility.

thaJeztah: good call, yes, that's better

milas: Ah yeah, good catch - moved it now. For some reason our `go.mod` has 2 indirect sections, so I think `go mod tidy` didn't move things as it normally would 🤔

milas:
> Had a quick peek if there was a specific reason for using this library; I see this was added as part of [7f8bb03](https://github.com/docker/compose/commit/7f8bb030e66a4d8de4b44a61fc18fa60bf9b14ae), which originally came from [docker/compose-cli@7f8bb03](https://github.com/docker/compose-cli/commit/7f8bb030e66a4d8de4b44a61fc18fa60bf9b14ae) ([docker/compose-cli#544](https://github.com/docker/compose-cli/pull/544)), but review on that PR doesn't mention a specific reason

Yup, I'm fairly certain it got auto-imported because it was already available as an indirect dependency at the time!

And even if we _had_ been attempting to utilize the custom YAML tag logic it added...it wouldn't work since we weren't ever using it for unmarshaling in the `compose-go` code

(Sidenote: I spent like 20 mins reading about YAML tags this morning and I still don't _really_ understand them 😵)

Issue #9655, opened 2022-07-15
Option `--rmi` doesn't remove images when used together with `-p`  


sylencecc:
**Description**
The command `docker-compose -p project down --rmi local` (or `--rmi all` ) won't remove unused images. This worked fine until compose 1.29.2 and fails with the current v2.6.1 as well as 2.1.0 (couldn't find a download to test with 2.0.0).

**Steps to reproduce the issue:**
1. Create minimal `Dockerfile` and `docker-compose.yml`:
```
$ cat Dockerfile
FROM alpine:3.16

$ cat docker-compose.yaml
services:

  api:
    build: .
    command: sleep infinity
```
3. Create the container with `docker-compose -p abc up -d`
4. After the container is running, remove it with `docker-compose -p abc down --rmi local`. Note that the image `abc_api` won't be removed.
```
 ⠿ Container abc-api-1  Removed
 ⠿ Network abc_default  Removed 
```
5. When running ` docker-compose -p abc down --rmi local`  (or `--rmi all`) again, the following warning appears. The image is still not removed.
```
Warning: No resource found to remove for project "abc".
```

**Describe the results you received:**
The docker images are kept despite telling docker-compose to remove them with ` --rmi local`, manual removal with ` docker rmi`  or similar works. Removal works fine if no custom project name is specified with ` -p <name>`.

**Describe the results you expected:**
(Local) images created during the build process should be removed when using `--rmi local`, even if a custom project name is specified with `-p`.

**Output of `docker compose version`:**
```
Docker Compose version v2.6.1
```
Reproducible with 2.1.0. Issue doesn't occur on 1.29.2 and earlier.

```
**Output of `docker info`:**
Client:
 Context:    default
 Debug Mode: false

Server:
 Containers: 15
  Running: 2
  Paused: 0
  Stopped: 13
 Images: 270
 Server Version: 20.10.12
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 1
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 io.containerd.runtime.v1.linux runc
 Default Runtime: runc
 Init Binary: docker-init
[...]

PR #9715, opened 2022-08-04
Fix `down` with `--rmi`

ulyssessouza:
**Related issue**
Resolves https://github.com/docker/compose/issues/9655

### DIFF
--- a/pkg/compose/build.go
+++ b/pkg/compose/build.go
@@ -139,6 +139,7 @@ func (s *composeService) ensureImagesExists(ctx context.Context, project *types.
 				project.Services[i].Labels = types.Labels{}
 			}
 			project.Services[i].CustomLabels[api.ImageDigestLabel] = digest
+			project.Services[i].CustomLabels[api.ImageNameLabel] = service.Image
 		}
 	}
 	return nil
@@ -191,6 +192,7 @@ func (s *composeService) getLocalImagesDigests(ctx context.Context, project *typ
 		digest, ok := images[imgName]
 		if ok {
 			project.Services[i].CustomLabels.Add(api.ImageDigestLabel, digest)
+			project.Services[i].CustomLabels.Add(api.ImageNameLabel, project.Services[i].Image)
 		}
 	}
 
--- a/pkg/compose/compose.go
+++ b/pkg/compose/compose.go
@@ -130,9 +130,13 @@ func (s *composeService) projectFromName(containers Containers, projectName stri
 		serviceLabel := c.Labels[api.ServiceLabel]
 		_, ok := set[serviceLabel]
 		if !ok {
+			serviceImage := c.Image
+			if serviceNameFromLabel, ok := c.Labels[api.ImageNameLabel]; ok {
+				serviceImage = serviceNameFromLabel
+			}
 			set[serviceLabel] = &types.ServiceConfig{
 				Name:   serviceLabel,
-				Image:  c.Image,
+				Image:  serviceImage,
 				Labels: c.Labels,
 			}
 		}
--- a/pkg/compose/down.go
+++ b/pkg/compose/down.go
@@ -120,7 +120,7 @@ func (s *composeService) ensureVolumesDown(ctx context.Context, project *types.P
 
 func (s *composeService) ensureImagesDown(ctx context.Context, project *types.Project, options api.DownOptions, w progress.Writer) []downOp {
 	var ops []downOp
-	for image := range s.getServiceImages(options, project) {
+	for image := range s.getServiceImagesToRemove(options, project) {
 		image := image
 		ops = append(ops, func() error {
 			return s.removeImage(ctx, image, w)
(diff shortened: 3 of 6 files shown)


milas:
I'm confused when this would help. The label seems duplicative. If the container exists, we can just read its actual image, right? Why would we need to read a value from a label instead?

pkg/compose/down.go
```diff
@@ -190,16 +190,14 @@ func (s *composeService) removeNetwork(ctx context.Context, name string, w progr
 	return nil
 }
 
-func (s *composeService) getServiceImages(options api.DownOptions, project *types.Project) map[string]struct{} {
+func (s *composeService) getServiceImagesToRemove(options api.DownOptions, project *types.Project) map[string]struct{} {
```
milas, line 193: Nice rename 👍

pkg/compose/build.go
```diff
@@ -139,6 +139,9 @@ func (s *composeService) ensureImagesExists(ctx context.Context, project *types.
 				project.Services[i].Labels = types.Labels{}
 			}
 			project.Services[i].CustomLabels[api.ImageDigestLabel] = digest
+			if service.Image != "" {
+				project.Services[i].CustomLabels[api.ImageNameLabel] = service.Image
+			}
```
milas, line 144: Shouldn't this _always_ add it (even if it's blank)?

pkg/compose/compose.go
```diff
@@ -130,9 +130,13 @@ func (s *composeService) projectFromName(containers Containers, projectName stri
 		serviceLabel := c.Labels[api.ServiceLabel]
 		_, ok := set[serviceLabel]
 		if !ok {
+			serviceImage := c.Image
+			if serviceNameFromLabel, ok := c.Labels[api.ImageNameLabel]; ok {
+				serviceImage = serviceNameFromLabel
+			}
```
milas, line 136: Nit: this is what tripped me up originally; I think a comment explaining behavior is warranted, i.e. that the label will be _present_ but empty for locally-built images and why we intentionally want that behavior 🙃

pkg/compose/down.go
```diff
@@ -190,16 +190,14 @@ func (s *composeService) removeNetwork(ctx context.Context, name string, w progr
...
 	for _, service := range project.Services {
-		image := service.Image
+		image := service.Labels[api.ImageNameLabel] // Information on the compose file at the creation of the container
 		if options.Images == "local" && image != "" {
 			continue
 		}
-		if image == "" {
-			image = api.GetImageNameOrDefault(service, project.Name)
-		}
+		image = api.GetImageNameOrDefault(service, project.Name)
```
milas, line 200: I think the changes in this function should be reverted -- the `projectFromName` now makes sure the project loads consistently, so `service.Image` should be `""` here, right?

Otherwise, as it stands right now, if you `compose down --rmi=local` on a project that was launched with previous version of Compose, this will behave like `--rmi=all`. (I just confirmed/tested this with an Awesome Compose sample, launched with v2.9.0, then ran `down --rmi=local` with this PR and it removed the `mariadb:10-focal` image 🙃)

milas (changes requested):
Thanks for explaining the logic to me! Definitely missed the subtlety around it 😅 

I did find a regression that'll remove non-locally built images from Compose projects that were launched before these changes though 😬

pkg/compose/down.go
ulyssessouza: `service.Image` will always have something in it, since containers need an image to be created. 
That's the reason why I needed to add a new label to express the content of the `service::image` field when building the image.

milas: So `container.Image` will always have a value, which is what the label is for, so the label (which should be `""`) will be used to overwrite `service.Image` @ https://github.com/docker/compose/pull/9715/files#diff-5dabf34848ad18b1b30364b2fd814b7678dbcc4d398227392578ef8266b0ac67R139

Repro example:
**`compose.yaml`**
```yaml
services:
  nginx:
    image: nginx
```

**Current Behavior**
```shell
$ docker-compose-2.9.0 up --wait
[+] Running 2/2
 ⠿ Network rmi_default    Created
 ⠿ Container rmi-nginx-1  Healthy

$ docker-compose-2.9.0 down --rmi=local
[+] Running 2/1
 ⠿ Container rmi-nginx-1  Removed
 ⠿ Network rmi_default    Removed
```

**`up` On 2.9.0, `down` on PR#9715**
```shell
$ docker-compose-2.9.0 up --wait
[+] Running 2/2
 ⠿ Network rmi_default    Created
 ⠿ Container rmi-nginx-1  Healthy

$ docker-compose-9715 down --rmi=local
[+] Running 3/1
 ⠿ Container rmi-nginx-1  Removed
 ⠿ Image nginx            Removed  # <-- 🚨🙀
 ⠿ Network rmi_default    Removed
```

pkg/compose/build.go
ulyssessouza: Good catch!

pkg/compose/down.go
ulyssessouza: Nice catch! 
Just fixed.

ulyssessouza:
@milas PTAL

milas (approved):

PR #9720, opened 2022-08-04
don't apply default pull policy from command line if one is define in service configuration

### LINKED ISSUE #9717: pull_policy: build pulls anyway before building
**Description**

If a service has `pull_policy: build`, I would expect that its image would never be pulled. However `docker-compose up` and `create` will try to pull it the image doesn't exist, and then build it and tag the built image on top of the pulled image.

1. Have a docker-compose.yml with a service with an image name, build context, and `pull_policy: build`
2. Ensure that the named image doesn't exist
3. Run `docker-compose create` or `up`

```
$ docker pull hello-world
$ docker images
...
hello-world            latest          feb5d9fea6a5   10 months ago    13.3kB
...
$ docker rmi hello-world
$ docker images | grep feb5d9fea6a5
$ # nothing, this is normal
$ cat > Dockerfile << EOF
FROM alpine
RUN echo
CMD ["echo", "I am not hello-world"]
EOF
$ cat > docker-compose.yml << EOF
services:
  hello-world:
    image: hello-world
    build: .
    pull_policy: build
EOF
$ docker-compose up
[+] Running 2/2
 ⠿ hello-world Pulled                                                                                                                                                                      2.7s
   ⠿ 2db29710123e Pull complete                                                                                                                                                            0.6s
[+] Building 0.1s (5/5) FINISHED                                                                                                                                                                
 => [internal] load build definition from Dockerfile                                                                                                                                       0.0s
 => => transferring dockerfile: 87B                                                                                                                                                        0.0s
 => [internal] load .dockerignore                                                                                                                                                          0.0s
 => => transferring context: 2B                                                                                                                                                            0.0s
 => [internal] load metadata for docker.io/library/alpine:latest                                                                                                                           0.0s
 => CACHED [1/2] FROM docker.io/library/alpine                                                                                                                                             0.0s
 => [2/2] RUN echo                                                                                                                                                                         0.5s
 => exporting to image                                                                                                                                                                     0.4s
 => => exporting layers                                                                                                                                                                    0.4s
 => => writing image sha256:9048e15af052267db498ed68a56cac4a6a0ee32810155b3cfb9829cf5819140a                                                                                               0.0s
 => => naming to docker.io/library/hello-world                                                                                                                                             0.0s
[+] Running 2/2
 ⠿ Network testcase_default          Created                                                                                                                                               0.0s
 ⠿ Container testcase-hello-world-1  Created                                                                                                                                               0.1s
Attaching to testcase-hello-world-1
testcase-hello-world-1  | I am not hello-world
testcase-hello-world-1 exited with code 0
$ docker images | grep hello-world
hello-world            latest          9048e15af052   About a minute ago   5.52MB                                                                                                                                              
hello-world            <none>          feb5d9fea6a5   10 months ago    13.3kB
$ # our built image is here and tagged latest, which is normal
$ # but the image from docker hub, that we deleted earlier, has been pulled again
```

The named image is downloaded from the repo before being built locally.

The repo is never contacted and the image is built locally.

```
Docker Compose version 2.9.0
```

**Output of `docker info`:**

```
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc., v0.8.2-docker)
  compose: Docker Compose (Docker Inc., 2.9.0)

Server:
 Containers: 3
  Running: 2
  Paused: 0
  Stopped: 1
 Images: 66
 Server Version: 20.10.17
 Storage Driver: btrfs
  Build Version: Btrfs v5.18.1
  Library Version: 102
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
 Swarm: inactive
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 10c12954828e7c7c9b6e0ea9b0c02b01407d3ae1.m
 runc version: 
 init version: de40ad0
 Security Options:
  seccomp
   Profile: default
  cgroupns
 Kernel Version: 5.15.55-2-lts
 Operating System: Arch Linux
 OSType: linux
 Architecture: x86_64
 CPUs: 8
 Total Memory: 7.456GiB
 Name: peanut.codl.fr
 ID: HTBR:M3DN:3IFA:VREG:7FMW:REU4:X23W:UUSQ:33P4:WQ74:CZTP:IHAT
 Debug Mode: false
 Labels:
 Experimental: false
 Insecure Registries:
  127.0.0.0/8
 Live Restore Enabled: false
```

glours:
**What I did**
Don't override service pull policy if the value from the command line is the default one

**Related issue**
fix #9717 

image

### DIFF
--- a/cmd/compose/create.go
+++ b/cmd/compose/create.go
@@ -31,6 +31,7 @@ type createOptions struct {
 	Build         bool
 	noBuild       bool
 	Pull          string
+	pullChanged   bool
 	removeOrphans bool
 	ignoreOrphans bool
 	forceRecreate bool
@@ -47,7 +48,8 @@ func createCommand(p *projectOptions, backend api.Service) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "create [OPTIONS] [SERVICE...]",
 		Short: "Creates containers for a service.",
-		PreRunE: Adapt(func(ctx context.Context, args []string) error {
+		PreRunE: AdaptCmd(func(ctx context.Context, cmd *cobra.Command, args []string) error {
+			opts.pullChanged = cmd.Flags().Changed("pull")
 			if opts.Build && opts.noBuild {
 				return fmt.Errorf("--build and --no-build are incompatible")
 			}
@@ -108,7 +110,7 @@ func (opts createOptions) GetTimeout() *time.Duration {
 }
 
 func (opts createOptions) Apply(project *types.Project) {
-	if opts.Pull != "" {
+	if opts.pullChanged {
 		for i, service := range project.Services {
 			service.PullPolicy = opts.Pull
 			project.Services[i] = service


cmd/compose/create.go
```diff
@@ -110,8 +109,10 @@ func (opts createOptions) GetTimeout() *time.Duration {
 func (opts createOptions) Apply(project *types.Project) {
 	if opts.Pull != "" {
 		for i, service := range project.Services {
-			service.PullPolicy = opts.Pull
-			project.Services[i] = service
+			if opts.Pull != "missing" || service.PullPolicy == "" {
```
thaJeztah, line 112: Instead of looking for the default _value_, it would be better to (for these);

- use the `flag.Changed()` to detect whether or not the flag was set by the user
- change `createOptions.Pull` to a pointer (`nil` means "no value set", in which case the default can be used)

```go
if flags.Changed("pull") {
	opts.Pull = ...
}
```

Doing that makes the option work as expected; setting the flag _even if it's the default_ value has a different meaning than "not setting the option"; because the user may _intentionally_ want to set the pull behavior back to the default (overriding the one in the compose file)

thaJeztah (commented):
Don't think this is the right fix (see my comment)

cmd/compose/create.go
glours: Good catch, I applied the same logic as for the other flags

cmd/compose/create.go
```diff
@@ -110,8 +110,10 @@ func (opts createOptions) GetTimeout() *time.Duration {
 func (opts createOptions) Apply(project *types.Project) {
 	if opts.Pull != "" {
```
thaJeztah, line 111: I guess this whole loop can be skipped if `opts.PullChanged` isn't set?

(that said; is there code elsewhere that sets the default if it's not defined in the service)?

glours: Right, it's too late, I should go to bed 😅

thaJeztah (approved):
LGTM (assuming green)

We should probably have a test for this (as a follow-up?)

thaJeztah:
```
Installed Go in 80ms
Requested golangci-lint 'latest', using 'v1.48.0', calculation took 119ms
Installing golangci-lint v1.48.0...
Downloading https://github.com/golangci/golangci-lint/releases/download/v1.48.0/golangci-lint-1.48.0-linux-amd64.tar.gz ...
/usr/bin/tar xz --overwrite --warning=no-unknown-keyword -C /home/runner -f /home/runner/work/_temp/7b3cfb22-e5a6-4959-a3f7-bbd4f340c7e4
Installed golangci-lint into /home/runner/golangci-lint-1.48.0-linux-amd64/golangci-lint in 647ms
```


Odd failures in CI during linting; I see we're not pinning golang-ci-lint to a version, which may mean it's not compatible with the go version we're using 🤔

thaJeztah:
unrelated; looks like CI is using a pretty old version of docker 20.10; https://github.com/docker/compose/blob/v2/.github/workflows/ci.yml#L104

laurazard (approved):
LGTM!

PR #9728, opened 2022-08-08
lint: run gofmt from Go 1.19

milas:
**What I did**
`gofmt -w -l -s .` with Go 1.19 😛 

**Related issue**
See #9722 

hamster with some broccoli

### DIFF
--- a/pkg/compose/build_classic.go
+++ b/pkg/compose/build_classic.go
@@ -65,7 +65,7 @@ func (s *composeService) doBuildClassic(ctx context.Context, project *types.Proj
 	return nameDigests, errs
 }
 
-//nolint: gocyclo
+// nolint: gocyclo
 func (s *composeService) doBuildClassicSimpleImage(ctx context.Context, options buildx.Options) (string, error) {
 	var (
 		buildCtx      io.ReadCloser


StefanScherer (approved):
LGTM

pkg/compose/build_classic.go
```diff
@@ -65,7 +65,7 @@ func (s *composeService) doBuildClassic(ctx context.Context, project *types.Proj
 	return nameDigests, errs
 }
 
-//nolint: gocyclo
+// nolint: gocyclo
```
thaJeztah, line 68: It took me some time (and a ticket in the Golang issue tracker https://github.com/golang/go/issues/53835) to learn, but the correct format for these "machine-readable" comments is to not have a leading space _AND_ no space after the colon; see https://github.com/moby/moby/pull/43804

So the valid format for these (and to be left alone by `gofmt`) is;

```go
//nolint:gocyclo
```

milas: Ahhh thanks! I was really lazy here because I just wanted to be able to run `golanglint-ci run` from the CLI without failures/downgrading it

Did a `nolint` cleanup in #9738 😊

PR #9729, opened 2022-08-08
Add platforms build

### LINKED ISSUE #8531: docker compose does not use current buildkit builder

glours:
**What I did**
Add support of the `platforms` attribut, [recently introduced](https://github.com/compose-spec/compose-spec/pull/267) in the Compose specification.

This feature will allow to build multi-arch images from the `build` command and push those imagines to a registry.

Huge thanks to @crazy-max for his help 🙏

Should fix #8531 

image


pkg/compose/build_buildkit.go
```diff
@@ -47,8 +56,10 @@ func (s *composeService) doBuildBuildkit(ctx context.Context, project *types.Pro
 	defer cancel()
 	w := xprogress.NewPrinter(progressCtx, s.stdout(), os.Stdout, mode)
 
-	// We rely on buildx "docker" builder integrated in docker engine, so don't need a DockerAPI here
-	response, err := build.Build(ctx, driverInfo, opts, nil, filepath.Dir(s.configFile().Filename), w)
+	// Get the DockerAPI if a "docker" export is defined (ie: up and run command), otherwise get nil and let use the default buildx builder
+	dockerAPI := getDockerAPI(s.dockerCli, opts)
+
+	response, err := build.Build(ctx, dis, opts, dockerAPI, filepath.Dir(s.configFile().Filename), w)
```
crazy-max, line 62: I think it's fine to always pass `dockerAPI` to `build.Build` as it's only used when loading the image in docker store anyway:

```suggestion
	response, err := build.Build(ctx, dis, opts, &internalAPI{s.dockerCli}, filepath.Dir(s.configFile().Filename), w)
```

pkg/compose/build_buildkit.go
```diff
@@ -47,8 +56,10 @@ func (s *composeService) doBuildBuildkit(ctx context.Context, project *types.Pro
 	defer cancel()
 	w := xprogress.NewPrinter(progressCtx, s.stdout(), os.Stdout, mode)
 
-	// We rely on buildx "docker" builder integrated in docker engine, so don't need a DockerAPI here
-	response, err := build.Build(ctx, driverInfo, opts, nil, filepath.Dir(s.configFile().Filename), w)
+	// Get the DockerAPI if a "docker" export is defined (ie: up and run command), otherwise get nil and let use the default buildx builder
+	dockerAPI := getDockerAPI(s.dockerCli, opts)
```
crazy-max, line 60: Can be removed (see previous comment)

pkg/compose/build_buildkit.go
```diff
@@ -71,3 +82,197 @@ func (s *composeService) doBuildBuildkit(ctx context.Context, project *types.Pro
...
+func getDockerAPI(cli command.Cli, opts map[string]build.Options) *internalAPI {
+	for _, opt := range opts {
+		for _, export := range opt.Exports {
+			if export.Type == "docker" {
+				return &internalAPI{dockerCli: cli}
+			}
+		}
+	}
+	return nil
+}
```
crazy-max, line 278: Can be removed (see previous comment)

pkg/compose/build.go
```diff
@@ -81,6 +81,14 @@ func (s *composeService) build(ctx context.Context, project *types.Project, opti
 				Attrs: map[string]string{"ref": image},
 			})
 		}
+		if len(buildOptions.Platforms) > 1 {
+			buildOptions.Exports = []bclient.ExportEntry{{
+				Type: "image",
+				Attrs: map[string]string{
+					"push": "true",
```
nicksieger, line 88: Does this mean that images are automatically pushed during a `docker compose build`?

pkg/compose/build_buildkit.go
```diff
@@ -71,3 +82,197 @@ func (s *composeService) doBuildBuildkit(ctx context.Context, project *types.Pro
...
+	var f driver.Factory
+	if ng.Driver != "" {
+		factories := driver.GetFactories()
+		for _, fac := range factories {
+			if fac.Name() == ng.Driver {
+				f = fac
+				continue
+			}
+		}
+		f = driver.GetFactory(ng.Driver, true)
```
nicksieger, line 108: Looks like this will overwrite any factory set in the loop, do you want an `if f == nil` guard?

pkg/compose/build.go
```diff
@@ -352,3 +358,26 @@ func addSecretsConfig(project *types.Project, service types.ServiceConfig) (sess
...
+	if service.Platform != "" && !utils.StringContains(service.Build.Platforms, service.Platform) {
+		return nil, fmt.Errorf("service.platform should be part of the service.build.platforms: %q", service.Platform)
+	}
+
+	for _, buildPlatform := range service.Build.Platforms {
+		p, err := platforms.Parse(buildPlatform)
+		if err != nil {
+			return nil, err
+		}
+		plats = append(plats, p)
```
milas, line 380: I think we need to make sure we don't add a duplicate if a platform is already present via `DOCKER_DEFAULT_PLATFORM` above:

```yaml
services:
  hello:
    build:
      context: .
      platforms:
        - linux/amd64
```

```shell
$ DOCKER_DEFAULT_PLATFORM="linux/amd64" docker compose build hello
...
 => ERROR exporting to image                                                                                                                           0.0s
------
 > exporting to image:
------
failed to solve: number of platforms does not match references 2 1
```

(Given how cryptic the BuildKit error is, it might also be beneficial [perhaps in `compose-go` loader/validate?] to error if duplicates exist in `service.build.platforms`, but I'm less concerned about that.)

nicksieger (commented):
LGTM! I have a couple questions.

crazy-max (commented):
Some minor changes.

Also looking at some logic copied over from buildx to this repo I'm afraid we are going to drift in the future.

Just my 2 cents but Compose should not push Docker images at all. Making it platform-neutral to build and run is a valid use-case but it should end here. If users want to build and push Docker images from their compose files they should use [Bake](https://docs.docker.com/build/bake/compose-file/).

As this is already [part of the specification](https://github.com/compose-spec/compose-spec/blob/master/build.md#publishing-built-images), I guess we have to move forward. So in a follow-up we should instead forward build requests to `docker buildx bake`. Or maybe by using [`bake.ReadTargets`](https://github.com/docker/buildx/blob/4fd3ec1a50376e587bb149088596d31d73fe5bb9/bake/bake.go#L87) and convert them to build opts like https://github.com/docker/buildx/blob/30d650862d26932ffa860465bcd61aa455595f60/commands/bake.go#L106-L120 but then you need extra logic to handle drivers and builders. Maybe there is room on our side to have better support for buildx as a library. cc @tonistiigi @jedevc

pkg/compose/build.go
```diff
@@ -352,3 +358,26 @@ func addSecretsConfig(project *types.Project, service types.ServiceConfig) (sess
 	}
 	return secretsprovider.NewSecretProvider(store), nil
 }
+
+func addPlatforms(project *types.Project, service types.ServiceConfig) ([]specs.Platform, error) {
+	var plats []specs.Platform
+	if platform, ok := project.Environment["DOCKER_DEFAULT_PLATFORM"]; ok {
```
milas, line 364: Should we respect this variable on `up` / `run` as well?

```shell
$ DOCKER_DEFAULT_PLATFORM="linux/amd64" docker compose up --build --quiet-pull --force-recreate hello

[+] Running 1/1
 ⠿ Container platform-hello-1  Recreated                                                                                                               0.1s
Attaching to platform-hello-1
platform-hello-1  | hello linux arm64
platform-hello-1 exited with code 0
```

pkg/compose/build.go
milas: Also, is it weird that you can "extend" the supported platforms via `DOCKER_DEFAULT_PLATFORM`?

e.g. in the same example above, `DOCKER_DEFAULT_PLATFORM=linux/arm64 docker compose build hello` would build both AMD64 + ARM64 images even though only `linux/amd64` is actually in `compose.yaml`.

pkg/compose/build.go
```diff
@@ -81,6 +81,14 @@ func (s *composeService) build(ctx context.Context, project *types.Project, opti
 				Attrs: map[string]string{"ref": image},
 			})
 		}
+		if len(buildOptions.Platforms) > 1 {
+			buildOptions.Exports = []bclient.ExportEntry{{
+				Type: "image",
+				Attrs: map[string]string{
+					"push": "true",
+				},
+			}}
```
milas, line 90: Shouldn't this be `type: "docker"` + `"load": "true"` like the other one?

I believe this is legitimately trying to push to registry:
```shell
$ docker compose build hello
...
 => ERROR exporting to image                                                                                                                           0.7s
 => => exporting layers                                                                                                                                0.1s
 => => exporting manifest sha256:90bf7923cd97b6dec8f53161c7df1e6d2587ec52ccc060273cdc6b6ae8bd6c9c                                                      0.0s
 => => exporting config sha256:ec0a66e1fe3ca40d1042b63f3f82a0ff10189f86c6b68c1dd2b6f3213ac7057a                                                        0.0s
 => => exporting manifest sha256:8a1a9bea74c44f39ab83dd4a831d74845c1f4eda2c3f2a5c230e0988fa554236                                                      0.0s
 => => exporting config sha256:296e0f515945635ccc81e78c428bdf2b341a725f8ad033afbff0db2d0615aec1                                                        0.0s
 => => exporting manifest list sha256:87e79b4bda828912aba9e22edd32010ab65b56da406c307bf60b61e79616b8cb                                                 0.0s
 => => pushing layers                                                                                                                                  0.5s
------
 > exporting to image:
------
failed to solve: server message: insufficient_scope: authorization failed
```

milas (commented):
Few questions from playing around with it but code LGTM at a high level

From a behavior standpoint, I think the most awkward thing is the necessity to create/activate a buildx builder out-of-band:
```
multiple platforms feature is currently not supported for docker driver. Please switch to a different driver (eg. "docker buildx create --use")
```

Personally, I'd expect Compose to create & use a project-scoped buildx context, similar to how it creates a network automatically.

pkg/compose/build.go
milas: I just realized it's doing this because you can't export multi-platform images with the `docker` export type

I wonder if we need a `--output` flag on `build` with a default of `type=docker` that behaves the same as the buildx flag, so it can error out if you try to do `docker compose build --output=docker` on a multi-platform image.

crazy-max:
> Few questions from playing around with it but code LGTM at a high level
> 
> From a behavior standpoint, I think the most awkward thing is the necessity to create/activate a buildx builder out-of-band:
> 
> ```
> multiple platforms feature is currently not supported for docker driver. Please switch to a different driver (eg. "docker buildx create --use")
> ```
> 
> Personally, I'd expect Compose to create & use a project-scoped buildx context, similar to how it creates a network automatically.

If compose can define its own buildx store that should be possible I think (using `BUILDX_CONFIG=$HOME/.docker/compose/buildx` env var maybe?). Then created builders would be scoped to compose.

pkg/compose/build.go
glours: Good catch, I added that to check e2e tests but as the default behaviour of Compose is to not push by default I will remove that and update the tests
Thanks @nicksieger

pkg/compose/build_buildkit.go
glours: yes we can

pkg/compose/build.go
milas: And TIL that `docker compose push` already exists 🤔 

But of course, it does not build:
```shell
$ docker compose push
An image does not exist locally with the tag: milas/hello
```

glours: Nope because `docker` exporter doesn't support multi-arch builds, I just need to remove the `push` option by default

pkg/compose/build.go
glours: yes! thanks @milas

pkg/compose/build.go
glours: Does it make sense for you if we manage the `DOCKER_DEFAULT_PLATFORM` like the `service.platform`, which means refuse to build if the `DOCKER_DEFAULT_PLATFORM` isn't defined in the `service.build.platforms` section?

milas:
Noticed today while using this build that I can't launch this project: https://github.com/dockersamples/compose-dev-env

```
❯ ~/dev/compose/bin/build/docker-compose up
[+] Building 1.6s (19/19) FINISHED
 => [compose-dev-env-backend internal] load build definition from Dockerfile                                                                           0.0s
 => => transferring dockerfile: 368B                                                                                                                   0.0s
 => [compose-dev-env-backend internal] load .dockerignore                                                                                              0.0s
 => => transferring context: 2B                                                                                                                        0.0s
 => [compose-dev-env-backend internal] load metadata for docker.io/library/alpine:3.12                                                                 1.0s
 => [compose-dev-env-backend internal] load metadata for docker.io/library/golang:1.16                                                                 0.3s
 => [compose-dev-env-proxy internal] load build definition from Dockerfile                                                                             0.0s
 => => transferring dockerfile: 147B                                                                                                                   0.0s
 => [compose-dev-env-proxy internal] load .dockerignore                                                                                                0.0s
 => => transferring context: 2B                                                                                                                        0.0s
 => [compose-dev-env-proxy internal] load metadata for docker.io/library/nginx:1.21                                                                    0.3s
 => [compose-dev-env-proxy 1/3] FROM docker.io/library/nginx:1.21@sha256:2bcabc23b45489fb0885d69a06ba1d648aeda973fae7bb981bafbb884165e514              0.0s
 => => resolve docker.io/library/nginx:1.21@sha256:2bcabc23b45489fb0885d69a06ba1d648aeda973fae7bb981bafbb884165e514                                    0.0s
 => [compose-dev-env-proxy internal] load build context                                                                                                0.0s
 => => transferring context: 157B                                                                                                                      0.0s
 => CACHED [compose-dev-env-proxy 2/3] RUN apt-get update     && apt-get install -y git                                                                0.0s
 => CACHED [compose-dev-env-proxy 3/3] COPY conf /etc/nginx/conf.d/default.conf                                                                        0.0s
 => [compose-dev-env-backend] exporting to image                                                                                                       0.2s
 => => exporting layers                                                                                                                                0.2s
 => => exporting manifest sha256:db821b139c4bd8c170ca26319bc2a57f68e7efe5763663828e8e77c7afa70366                                                      0.0s
 => => exporting config sha256:e5afd9082c8c57de530edec9b3dd2e890c1f1690652f84314aaa786178f26fa7                                                        0.0s
 => => exporting manifest sha256:c946c74d196d05d35421d351df959d6c8e5ca1bbdea83e8b0ae1d16d1714f875                                                      0.0s
 => => exporting config sha256:3fe4a410f3de7c181689ef16cc08595d3cff1f789b3fcde2dea49395514e9950                                                        0.0s
 => [compose-dev-env-backend build 1/4] FROM docker.io/library/golang:1.16@sha256:5f6a4662de3efc6d6bb812d02e9de3d8698eea16b8eb7281f03e6f3e8383018e     0.0s
 => => resolve docker.io/library/golang:1.16@sha256:5f6a4662de3efc6d6bb812d02e9de3d8698eea16b8eb7281f03e6f3e8383018e                                   0.0s
 => [compose-dev-env-backend stage-2 1/2] FROM docker.io/library/alpine:3.12@sha256:c75ac27b49326926b803b9ed43bf088bc220d22556de1bc5f72d742c91398f69   0.3s
 => => resolve docker.io/library/alpine:3.12@sha256:c75ac27b49326926b803b9ed43bf088bc220d22556de1bc5f72d742c91398f69                                   0.0s
 => => sha256:0eeab5c200691bd777e227c6eea27f7ca3c8232b67118a76edac2dcde3186aa1 2.72MB / 2.72MB                                                         0.2s
 => => extracting sha256:0eeab5c200691bd777e227c6eea27f7ca3c8232b67118a76edac2dcde3186aa1                                                              0.1s
 => [compose-dev-env-backend internal] load build context                                                                                              0.0s
 => => transferring context: 3.08kB                                                                                                                    0.0s
 => CACHED [compose-dev-env-backend build 2/4] WORKDIR /go/src/github.com/org/repo                                                                     0.0s
 => CACHED [compose-dev-env-backend build 3/4] COPY . .                                                                                                0.0s
 => CACHED [compose-dev-env-backend build 4/4] RUN go build -o server .                                                                                0.0s
 => [compose-dev-env-backend stage-2 2/2] COPY --from=build /go/src/github.com/org/repo/server /server                                                 0.0s
[+] Running 3/3
 ⠿ Network compose-dev-env_default      Created                                                                                                        0.0s
 ⠿ Volume "compose-dev-env_db-data"     Created                                                                                                        0.0s
 ⠿ Container compose-dev-env-db-1       Created                                                                                                        0.0s
 ⠋ Container compose-dev-env-backend-1  Creating                                                                                                       0.0s
Error: No such image: compose-dev-env-backend
```

(Not actually running via dev envs feature, using the `docker-compose.yaml` in the root of that repo)

pkg/utils/slices.go
```diff
@@ -0,0 +1,30 @@
...
+   limitations under the License.
+*/
+
+package utils
+
+import "reflect"
+
+// Contains helps to detect if a non-comparable struct is part of an array
+// only use this method if you can't rely on existing golang Contains function of slices (https://pkg.go.dev/golang.org/x/exp/slices#Contains)
+func Contains[T any](origin []T, element T) bool {
```
milas, line 23: image

milas (approved):
:shipit: !!

milas:
> @milas IHMO buildx is Compose defaut builder, so if users decided to use the legacy builder, they don't expect to some automagic behaviour during the build process, especially with an explicit discarded builder, right?

I think the awkwardness is for users who are using BuildKit but with the default Docker driver, so they will likely be confused if they don't know about buildx + builder contexts.

That said, I think the current approach is fine, as the error message does give you the command to create a buildx context, at which point things will work.

---

In the future, maybe a new flag, e.g. `docker compose build --current-platform` would be useful, so users who have multiple `platforms` defined can still build locally independently from `up` without needing to set up buildx + do extra builds for other platforms they don't need during development

lara-ec:
Hi,
We have a docker-compose.yml that specifies `platform: linux/amd64` for certain containers, since they don't properly run on arm64, so that the env also runs on M1 macs.
This change breaks our current config with error `service.platform should be part of the service.build.platforms: "linux/amd64"`, and I need to add a `platforms` entry to all the `build` sections to fix that. Is this intentional?
Thanks :)

nicksieger:
@lara-ec do your compose services that need to be constrained to amd64 have a build section at all? Would you mind including a sample snippet of your compose file? Thanks.

glours:
@lara-ec can you confirm me that it's only broken for M1 users? 
The error happens for `docker compose build` command? What is the result of a `docker compose up --build`?

glours:
@lara-ec can you test this [PR](https://github.com/docker/compose/pull/9847) and let me know if it fixes your issue for both `arm64` and `amd64` computers?

lara-ec:
@nicksieger Yes, we need a `build` section, the issue is that the underlying image doesn't work on arm64.
```yaml
service:
  platform: linux/amd64
  build:
    context: ./service/
    dockerfile: Dockerfile.dev
```

@glours I've only tested it on Intel, I can test it on M1 tomorrow. But at least on Intel, the PR fixes it :)

glours:
@lara-ec, @StefanScherer and I reproduced your issue, we confirm on our side that the PR resolve the problem. 
We'll do a `v2.11.1` release soon

PR #9731, opened 2022-08-08
fix formatting of "nolint" tags for go1.19

thaJeztah:
- follow-up to https://github.com/docker/compose/pull/9728 https://github.com/docker/compose/pull/9728#discussion_r940397357

The correct formatting for machine-readable comments is;

    //<namespace>:<options>[,<option>...][ // comment]

Which basically means:

- MUST NOT have a space before `<namespace>` (e.g. `nolint`)
- Namespace MUST be alphanumeric
- MUST be followed by a colon
- MUST be followed by at least one `<option>`
- Optionally additional `<options>` (comma-separated)
- Optionally followed by a comment

Any other format will not be considered a machine-readable comment by `gofmt`,
and thus formatted as a regular comment. Note that this also means that a
`//nolint` (without anything after it) is considered invalid, same for `//#nosec`
(starts with a `#`).

**What I did**

**Related issue**

### DIFF
--- a/pkg/compose/build_classic.go
+++ b/pkg/compose/build_classic.go
@@ -65,7 +65,7 @@ func (s *composeService) doBuildClassic(ctx context.Context, project *types.Proj
 	return nameDigests, errs
 }
 
-// nolint: gocyclo
+//nolint:gocyclo
 func (s *composeService) doBuildClassicSimpleImage(ctx context.Context, options buildx.Options) (string, error) {
 	var (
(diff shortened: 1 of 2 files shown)


milas:
Whoops, I saw your comment on the other PR and went right to fix it without looking for new PRs first, so I already merged these changes + some others in #9738  🤦

milas:
(but thanks for the tip & following up with the fix! sorry I'm bad at GitHub 😶)

PR #9738, opened 2022-08-09
lint: add `nolintlint` and clean up `nolint` directives

milas:
**What I did**
We had some incorrectly formatted and/or ineffective `nolint` directives hanging around. I added `nolintlint` to the list of linters, fixed up some violations it flagged, and did some creative finds to look for some other weird/pointless ones (but spent about 5 mins on this, so it's not exhaustive).

Thanks to @thaJeztah for pointing this out @ https://github.com/docker/compose/pull/9728#pullrequestreview-1065409215

**Related issue**
N/A

capybaras relaxing in the bath

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -18,6 +18,7 @@ linters:
     - lll
     - misspell
     - nakedret
+    - nolintlint
     - staticcheck
     - structcheck
     - typecheck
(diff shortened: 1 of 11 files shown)


nicksieger (approved):
👍

.golangci.yml
```diff
@@ -18,6 +18,7 @@ linters:
     - lll
     - misspell
     - nakedret
+    - nolintlint
```
thaJeztah, line 21: Oh! Didn't know about this linter; looks handy. Perhaps I should enable it one some other repositories

PR #9744, opened 2022-08-12
Better sandboxed workflow and enhanced cross compilation

crazy-max:
This PR brings various improvements and fix some issues linked to non sandboxed builds.

* `builder.Makefile` has been removed as it's called anyway only within the Dockerfile.
* `cross` target now effectively builds against any platforms (using `--platform` with buildx) without hardcoded targets so we avoid building all platforms sequentially.
* adds bake definition. `Makefile` now calls bake to reduce commands overhead. I have not changed goals names in the Makefile to avoid disturbing your habits but would be nice to make things more generic like `make vendor` makes more sense than `make go-mod-tidy` imo.
* `docs` Dockerfile merged with main Dockerfile to avoid deduplicated stages for better cache reuse.
* `scripts` folder not necessary anymore. Everything is handled inside the Dockerfile.
* license headers validation is now containerized so we are aligned with ci (also adds a license update stage if we want to update headers on all files if necessary).
* refactor GHA workflows
  * introduce concurrency check so we reduce queued runners
  * adds validate job in ci workflow calling each validation targets in a matrix to reduce build time.
  * cross stage has been improved and now runs against a matrix for all compatible platforms, take ~2m10s instead of ~20m previously. this way we can cross on PR without increasing build time and also be able to detect regressions.
  * use GitHub Cache backend (faster than local one)
  * mutualize and use matrix for e2e jobs
  * use commit sha for untrusted actions we use
  * merge release workflow in ci workflow to improve jobs, cache and artifacts reuse. This change how you release today. Now instead of invoking manually the workflow, the release is triggered when a tag is pushed.

This will also improve docker-ce-packaging for static bundles because atm `cross` target builds all platforms sequentially. cc @thaJeztah 


crazy-max:
Removed the step in ci that sends coverage report to Codecov (we can do that in a follow-up if you want)

crazy-max:
image

Would need to update the GitHub checks

glours (approved):
Awesome work! Thank you so much @crazy-max

glours:
I'll update the checks when the PR will be merged into `v2` branch 👌

nicksieger (approved):
Amazing, very much appreciated! 🎉

cpuguy83:
This made it a PITA to build compose without docker (or more appropriately with a different build toolchain).

crazy-max:
@cpuguy83 Sorry, https://github.com/docker/compose/pull/9776 should help.

cpuguy83:
Awesome, thanks!

Issue #9210, opened 2022-02-26
Support setting DOCKER_HOST in .env file


jmera:
**Description**

With Compose V1 we have the ability to set a project-specific `DOCKER_HOST` in .env, with V2 this is no longer possible.

**Steps to reproduce the issue:**
1. Add an `.env` file with a custom value for `DOCKER_HOST`
2. `docker compose` commands target the _default_ docker host

**Describe the results you received:**
`docker compose` commands target the _default_ docker host

**Describe the results you expected:**
`docker compose` commands target the docker host specified in .env

**Output of `docker compose version`:**

```
➜  ~ docker compose version
Docker Compose version v2.2.3
```

**Output of `docker info`:**

```
➜  ~ docker info
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc., v0.7.1)
  compose: Docker Compose (Docker Inc., v2.2.3)
  scan: Docker Scan (Docker Inc., v0.17.0)

Server:
 Containers: 0
  Running: 0
  Paused: 0
  Stopped: 0
 Images: 15
 Server Version: 20.10.12
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 2
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 io.containerd.runtime.v1.linux runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 7b11cfaabd73bb80907dd23182b9347b4245eb5d
 runc version: v1.0.2-0-g52b36a2
 init version: de40ad0
 Security Options:
  seccomp
   Profile: default
  cgroupns
 Kernel Version: 5.10.76-linuxkit
 Operating System: Docker Desktop
 OSType: linux
 Architecture: x86_64
 CPUs: 8
 Total Memory: 5.804GiB
 Name: docker-desktop
 ID: 6IUW:D7VH:NGSI:3VVB:WPYB:EQ3V:6JCD:IARV:6M6N:VRTI:MXNT:RCJ2
 Docker Root Dir: /var/lib/docker
 Debug Mode: false
[...]

PR #9745, opened 2022-08-12
Apply newly loaded envvars to "DockerCli" and "APIClient"

ulyssessouza:
**What I did**
Re-evaluate `DockerCli` and `APIClient` after reading the environment file.
I can contain `DOCKER_HOST` and/or `DOCKER_CONTEXT` so the `DockerCli` passed by `docker/cli` has to be re-evaluated.

**Related issue**
Resolves https://github.com/docker/compose/issues/9210

### DIFF
--- a/pkg/api/proxy.go
+++ b/pkg/api/proxy.go
@@ -22,6 +22,8 @@ import (
 	"github.com/compose-spec/compose-go/types"
 )
 
+var _ Service = &ServiceProxy{}
+
 // ServiceProxy implements Service by delegating to implementation functions. This allows lazy init and per-method overrides
 type ServiceProxy struct {
 	BuildFn              func(ctx context.Context, project *types.Project, options BuildOptions) error
@@ -58,8 +60,6 @@ func NewServiceProxy() *ServiceProxy {
 
 // Interceptor allow to customize the compose types.Project before the actual Service method is executed
 type Interceptor func(ctx context.Context, project *types.Project)
-
-var _ Service = &ServiceProxy{}
 
 // WithService configure proxy to use specified Service as delegate
 func (s *ServiceProxy) WithService(service Service) *ServiceProxy {
--- a/cmd/compose/compose.go
+++ b/cmd/compose/compose.go
@@ -32,6 +32,8 @@ import (
 	dockercli "github.com/docker/cli/cli"
 	"github.com/docker/cli/cli-plugins/manager"
 	"github.com/docker/cli/cli/command"
+	"github.com/docker/cli/cli/flags"
+	"github.com/docker/docker/client"
 	"github.com/morikuni/aec"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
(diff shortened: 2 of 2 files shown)


pkg/api/proxy.go
```diff
@@ -22,6 +22,8 @@ import (
 	"github.com/compose-spec/compose-go/types"
 )
 
+var _ Service = (*ServiceProxy)(nil)
```
milas, line 25: Just curious: any reason for this change? The old way (`var _ Service = &ServiceProxy{}`) is the way I've always seen it in Go code, and won't actually result in any allocations/memory usage at runtime.

milas (approved):
LGTM

pkg/api/proxy.go
ulyssessouza: Nothing special... Just that it visually avoids allocation. But it makes total sense that the compiler just avoids it in an optimisation phase.
I'll change it back then. The classical one looks prettier.

nicksieger (approved):

pkg/api/proxy.go
milas: Ahh yeah I def agree that the "idiomatic" way is 🪄 magic 🪄

markdascher:
Hate to be that guy 😄 But this might've broken something, at least on MacOS. I recently updated to docker-compose v2.10 and started getting this error for any `docker-compose` or `docker compose` command:

```
Error response from daemon: Client sent an HTTP request to an HTTPS server.
```

Using these env vars:

```
DOCKER_BUILDKIT="1"
DOCKER_TLS_VERIFY="1"
DOCKER_HOST="tcp://192.168…:2376"
DOCKER_CERT_PATH="…/certs"
```

Tried moving these env vars into an `.env` file but it didn't help. (Might've been doing it wrong though, since I've never used `.env` files with docker before.)

`git bisect` pointed to this commit, and reverting it solved the issue.

markdascher:
Found out a little more detail here. TLSOptions are only set in one spot, inside of [InstallFlags](https://github.com/docker/cli/blob/v20.10.17/cli/flags/common.go#L61-L65). That method isn't called by the `flags.NewClientOptions()` added in this PR, which results in `dockerCli.dockerEndpoint.EndpointMeta.TLSData` becoming nil.

InstallFlags seems like a strange place for that particular code to be tucked away. Also, that file calls `os.Getenv` in global initializers, which will always be incompatible with the runtime override mechanism added in this PR.

AFAICT there's no workaround, so I just have to use an older version. 2.7.0 is the newest one without a big warning on it, so that's probably the one I'll pick. But I don't know if the issue is really with this PR or if the code in `docker/cli` should be reworked?

nicksieger:
@markdascher thanks for the sleuthing, that's helpful. I'd like to see your issue addressed in some way, whether a kludge or rolling back this PR or some other intervention. I filed #9789 to track this.

trcoelho:
Hello, when this change (docker desktop version) will be available?

Thank you.

milas:
Copying https://github.com/docker/compose/issues/9210#issuecomment-1238223140 here:

> Re-opening because we had to revert #9745 in #9792 because it caused issues with TLS-enabled Docker hosts.
> 
> As a result, this is not currently available in the latest release of Docker Compose.

dmorrison:
I've been struggling for a long time with the "Client sent an HTTP request to an HTTPS server." error in a local dev setup, and @markdascher's suggestion to use an older version fixed my issue! Thank you!!!!!!

markdascher:
Glad to hear my comment was helpful! But the latest version (or anything after v2.10.2) already reverted this particular change. I'm currently running version 2.18.1 and there's no "Client sent an HTTP request to an HTTPS server" error. So that might be worth a shot, too.

PR #9770, opened 2022-08-19
ci: reduce noise from dependabot on Docker deps

milas:
**What I did**
There's a complex dependency situation with `docker/docker`,
`docker/cli`, and `docker/buildkit`. Upgrading them usually
needs to happen in unison to ensure compatible versions
between them, particularly because `docker/buildx` is not
1.0, so has no guarantees re: compatibility, and `docker/docker`
& `docker/cli` use CalVer rather than SemVer, so also have
different compatibility guarantees than necessarily expected
by Go tooling.

Patch versions are still considered for these to ensure we
don't miss important bugfixes.

**Related issue**
#9766 & #9758 😒 

sleeping cat covering its face with its paw

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,3 +4,15 @@ updates:
     directory: /
     schedule:
       interval: daily
+    ignore:
+      # docker/buildx + docker/cli + docker/docker require coordination to
+      # ensure compatibility between them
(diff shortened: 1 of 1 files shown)


laurazard (approved):
Less spam! 🎉

PR #9771, opened 2022-08-19
ci: restore builder.Makefile for docker-ce-packaging

milas:
**What I did**
Compose gets rebuilt for `docker-ce-packaging` as part of DEB &
RPM builds. These cannot use Docker within the build, so need a
native `go build` command.

We want to keep this here so that we have control of the build
flags etc independently of packaging.

NOTE: The output path is now `bin/build/docker-compose` to keep
      parity with the rest of the Compose build logic - previously
      it was `bin/docker-compose` (no intermediate `build/` dir)

**Related issue**
https://github.com/docker/docker-ce-packaging/pull/742

two very similar looking farm cats


nicksieger (approved):

nicksieger:
Checked out and ran `make -f builder.Makefile` and all is well ✅

thaJeztah:
We merged https://github.com/docker/docker-ce-packaging/pull/744, which fixes the issue in packaging; I guess we can decide to keep that, or to switch back to using the `builder.Makefile` from https://github.com/docker/compose/pull/9771.

Overall, the build for the compose plugin is fairly trivial, but of course the `-ldflags` is usually where the fun is (so there's still a small risk of diverging)

nicksieger:
#9776 should obsolete this

PR #9765, opened 2022-08-19
Add a local builder Makefile to easy local dev with compose-go replace

ulyssessouza:
**What I did**
Facilitate local development by allowing build with a `replace` clause in local file system for `compose-spec/compose-go` in `go.mod`

```console
make -f local.Makefile
```

PR #9776, opened 2022-08-21
Makefile: mutualize local and Dockerfile build opts

crazy-max:
follow-up https://github.com/docker/compose/pull/9771#issuecomment-1221299556

Made some changes to avoid drifting with build opts between local and Dockerfile.

@thaJeztah We can then change https://github.com/docker/docker-ce-packaging/blob/43635c4329d15c3639d6b8693b653b7ea9818487/deb/common/rules#L27-L34:

```
	&& mkdir -p /usr/libexec/docker/cli-plugins/ \
	&& GO111MODULE=on \
		CGO_ENABLED=0 \
			go build \
				-trimpath \
				-ldflags="-s -w -X github.com/docker/compose/v2/internal.Version=$(COMPOSE_VERSION)" \
				-o "/usr/libexec/docker/cli-plugins/docker-compose" \
				./cmd
```

With:

```
	&& make build DESTDIR=/usr/libexec/docker/cli-plugins
```

Also fixes the `cross` goal that was not building the right bake target.

closes #9771
closes #9765

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,15 +12,11 @@
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
 
-ifneq (, $(BUILDX_BIN))
-	export BUILDX_CMD = $(BUILDX_BIN)
-else ifneq (, $(shell docker buildx version))
-	export BUILDX_CMD = docker buildx
-else ifneq (, $(shell which buildx))
-	export BUILDX_CMD = $(which buildx)
-else
-	$(error "Buildx is required: https://github.com/docker/buildx#installing")
-endif
+PKG := github.com/docker/compose/v2
+VERSION ?= $(shell git describe --match 'v[0-9]*' --dirty='.m' --always --tags)
+
+GO_LDFLAGS ?= -s -w -X ${PKG}/internal.Version=${VERSION}
+GO_BUILDTAGS ?= e2e,kube
 
 UNAME_S := $(shell uname -s)
 ifeq ($(UNAME_S),Linux)
@@ -30,8 +26,6 @@ ifeq ($(UNAME_S),Darwin)
 	MOBY_DOCKER=/Applications/Docker.app/Contents/Resources/bin/docker
 endif
 
-BINARY_FOLDER=$(shell pwd)/bin
-GIT_TAG?=$(shell git describe --tags --match "v[0-9]*")
 TEST_FLAGS?=
 E2E_TEST?=
 ifeq ($(E2E_TEST),)
@@ -39,14 +33,21 @@ else
 	TEST_FLAGS=-run $(E2E_TEST)
 endif
 
-all: compose-plugin
+BUILDX_CMD ?= docker buildx
+DESTDIR ?= ./bin/build
 
-.PHONY: compose-plugin
-compose-plugin: ## Compile the compose cli-plugin
+all: build
+
+.PHONY: build ## Build the compose cli-plugin
+build:
+	CGO_ENABLED=0 GO111MODULE=on go build -trimpath -tags "$(GO_BUILDTAGS)" -ldflags "$(GO_LDFLAGS)" -o "$(DESTDIR)/docker-compose" ./cmd
+
+.PHONY: binary
+binary:
 	$(BUILDX_CMD) bake binary
 
 .PHONY: install
-install: compose-plugin
+install: binary
 	mkdir -p ~/.docker/cli-plugins
 	install bin/build/docker-compose ~/.docker/cli-plugins/docker-compose
 
@@ -61,10 +62,10 @@ e2e-compose-standalone: ## Run End to end local tests in standalone mode. Set E2
 	go test $(TEST_FLAGS) -v -count=1 -parallel=1 --tags=standalone ./pkg/e2e
 
 .PHONY: build-and-e2e-compose
-build-and-e2e-compose: compose-plugin e2e-compose ## Compile the compose cli-plugin and run end to end local tests in plugin mode. Set E2E_TEST=TestName to run a single test
+build-and-e2e-compose: build e2e-compose ## Compile the compose cli-plugin and run end to end local tests in plugin mode. Set E2E_TEST=TestName to run a single test
 
 .PHONY: build-and-e2e-compose-standalone
-build-and-e2e-compose-standalone: compose-plugin e2e-compose-standalone ## Compile the compose cli-plugin and run End to end local tests in standalone mode. Set E2E_TEST=TestName to run a single test
+build-and-e2e-compose-standalone: build e2e-compose-standalone ## Compile the compose cli-plugin and run End to end local tests in standalone mode. Set E2E_TEST=TestName to run a single test
 
 .PHONY: mocks
 mocks:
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -12,15 +12,12 @@
...
 
-ifneq (, $(BUILDX_BIN))
-	export BUILDX_CMD = $(BUILDX_BIN)
-else ifneq (, $(shell docker buildx version))
-	export BUILDX_CMD = docker buildx
-else ifneq (, $(shell which buildx))
-	export BUILDX_CMD = $(which buildx)
-else
-	$(error "Buildx is required: https://github.com/docker/buildx#installing")
-endif
```
nicksieger, line 23: Rather than using the funky `buildx-cmd` target with the eval function below, could we instead change the `else` block of the above code to be

```Makefile
else
    export BUILDX_CMD = echo "Buildx is required: https://github.com/docker/buildx#installing"; false
```

nicksieger (changes requested):

Makefile
```diff
@@ -76,55 +78,70 @@ mocks:
...
+.PHONY: buildx-cmd
+ifneq (, $(BUILDX_BIN))
+buildx-cmd:
+	$(eval BUILDX_CMD = $(BUILDX_BIN))
+else ifneq (, $(shell docker buildx version 2>/dev/null))
+buildx-cmd:
+	$(eval BUILDX_CMD = docker buildx)
+else ifneq (, $(shell which buildx 2>/dev/null))
+buildx-cmd:
+	$(eval BUILDX_CMD = $(shell which buildx))
```
thaJeztah, line 140: note that generally `command -v` is preferred over `which` (for portability). For the detection here, I think we would only need to know _IF_ it's available (and no need to run it using the full path, as we already use the `PATH` to discover)

Makefile
```diff
@@ -76,55 +78,70 @@ mocks:
...
+else
+buildx-cmd:
+	$(error "Buildx is required: https://github.com/docker/buildx#installing")
```
thaJeztah, line 143: But.. I have to admit that I kinda agree with https://github.com/docker/compose/pull/9776#discussion_r952832902, and wonder to what extend we need to have this complicated auto-detection.

I think common scenarios would be either;

- `docker` is installed (and `docker buildx` would be available as well for regular installs)
- it's run in GitHub actions with only `buildx` installed; in that case we can set `BUILDX_CMD=buildx`
- if neither is true, then I think a `/bin/sh: docker: not found` error may be "just fine" (after all, we're also not doing similar things to detect if `go` is installed, or `make`, or `git`)

If we agree with the above, just a;

```make
BUILDX_CMD ?= docker buildx
```

would cover that scenario

(feedback / thoughts welcome!)

crazy-max: Sure sgtm to remove this detection.

Makefile
```diff
@@ -12,15 +12,13 @@
...
-else ifneq (, $(shell which buildx))
-	export BUILDX_CMD = $(which buildx)
-else
-	$(error "Buildx is required: https://github.com/docker/buildx#installing")
-endif
+PKG := github.com/docker/compose/v2
+VERSION ?= $(shell git describe --match 'v[0-9]*' --dirty='.m' --always --tags)
+
+CGO_ENABLED ?= 0
+GO111MODULE ?= on
```
thaJeztah, line 19: - If this project cannot be built without go modules, we may not even need this variable (and just hard-code to `on` or `auto` below)
- Same for `CGO_ENABLED` (are there specific builds where it must be enabled? or should it _always_ be off?)

crazy-max: Yes might be fine for compose to keep them hardcoded. Don't see any reason to not to looking at current implementation.

Makefile
```diff
@@ -30,23 +26,29 @@ ifeq ($(UNAME_S),Darwin)
...
 
-.PHONY: compose-plugin
-compose-plugin: ## Compile the compose cli-plugin
+all: build
+
+.PHONY: build ## Build the compose cli-plugin
+build:
+	$(eval DESTDIR ?= $(shell pwd)/bin/build)
+	mkdir -p "$(DESTDIR)"
+	CGO_ENABLED=0 GO111MODULE=on go build -trimpath -tags "$(GO_BUILDTAGS)" -ldflags "$(GO_LDFLAGS)" -o "$(DESTDIR)/docker-compose" ./cmd
```
thaJeztah, line 44: Was there a specific reason to require an absolute path for `DESTDIR`?

I think the `mkdir` may be redundant as well; `go build` should create missing directories. I just gave that a try to be sure, but it looks to be doing that;

```bash
ls somedir
ls: somedir: No such file or directory

go build -o somedir/foo .

ls somedir/
foo*
```

Makefile
thaJeztah: Thanks! Sorry for being nit-picky there (I can see some value for auto-detection in other scenarios), just looking "can we simplify things (within reason)"? In the end, the repository would have a "how to build" with some prerequisites, so if things fail, users should just "read the manual" 😂

Makefile
```diff
@@ -30,23 +26,28 @@ ifeq ($(UNAME_S),Darwin)
...
-all: compose-plugin
+BUILDX_CMD ?= docker buildx
 
-.PHONY: compose-plugin
-compose-plugin: ## Compile the compose cli-plugin
+all: build
+
+.PHONY: build ## Build the compose cli-plugin
+build:
+	$(eval DESTDIR ?= ./bin/build)
```
thaJeztah, line 42: Sorry.. last question; is there an advantage to doing the `$(eval ..)` here, over having `DESTDIR ?= ./bin/build` at the top of the makefile? If there's no advantage, I think it'd be clearer to move it there at the top; makes it slightly more visible / transparent.

thaJeztah (commented):
LGTM

(one last question https://github.com/docker/compose/pull/9776#discussion_r953161860, but not a blocker)

do we need to try this one in docker-ce-packaging?

Makefile
crazy-max: `DESTDIR` would be propagated to the bake definition and overrides defaults in https://github.com/docker/compose/blob/23deefd20cc5dd5ebe11161df79ab7a60934ceda/docker-bake.hcl#L27-L34

Have to double-check though.

crazy-max:
> do we need to try this one in docker-ce-packaging?

Yes good idea

Makefile
crazy-max: Ah it's not actually, not an env var, so looks good to put it at the top.

thaJeztah:
> > do we need to try this one in docker-ce-packaging?
> 
> Yes good idea

I was about to start a branch, but let me know if you're working on that (otherwise we'll end up in a race condition 😂 )

crazy-max:
> > > do we need to try this one in docker-ce-packaging?
> > 
> > 
> > Yes good idea
> 
> I was about to start a branch, but let me know if you're working on that (otherwise we'll end up in a race condition 😂 )

Feel free to do it and @ me

thaJeztah:
There you go; https://github.com/docker/docker-ce-packaging/pull/748 (hope it works with using branch-name as reference 😂)

thaJeztah (approved):
CI on https://github.com/docker/docker-ce-packaging/pull/748 is happy 🎉 

LGTM, thank you!

thaJeztah:
Did a quick rebase, as this repo looks to have "PR is outdated" checks enabled, in case that's a blocker (rebase was clean).

@nicksieger I think your comment (https://github.com/docker/compose/pull/9776#discussion_r952832902) was addressed; see https://github.com/docker/compose/pull/9776#discussion_r953152104, PTAL

nicksieger (approved):
LGTM, looks cleaner too, thanks.

thaJeztah:
just realised this was not included in the v20.10.1 patch release 😂 - I was updating the build pipeline with the expectation this was fixed, and things failed, then realised it was not merged yet. 🙃 🙈

nicksieger:
Sorry about that. Wasn't sure how soon @crazy-max was going to get back to this so I didn't think to hold the release up for it. I have no problem creating a 2.10.2 release sooner than later.

thaJeztah:
No need to rush; we patched the release scripts for 20.10.0 to make it work. I was just "wait? what did I mess up?" ooooooooh.. 😂

PR #9783, opened 2022-08-23
Remove error message showing exit code when using --exit-code-from

### LINKED ISSUE #9782: Exit code printed as error when using --exit-code-from
**Description**

When using the `--exit-code-from` flag, an error message is output with the status code, giving the impression an error occurred even if the exit code was 0.

1. Use the following compose file
    ```yaml
    services:
      app:
        image: ubuntu
        command: sleep 1
    ```
1. Start using `docker compose up --exit-code-from app`

```
docker compose up --exit-code-from app
[+] Running 1/0
 ⠿ Container compose-app-1  Recreated                                      0.0s
Attaching to compose-app-1
compose-app-1 exited with code 0
Aborting on container exit...
[+] Running 1/0
 ⠿ Container compose-app-1  Stopped                                        0.0s
ERRO[0001] 0
```

The same output minus the `ERRO[0001] 0` message.

```
Docker Compose version v2.9.0
```

PR incoming (just making the issue first)

mikesir87:
**What I did**

Remove the unnecessary error message

**Related issue**
Fixes #9782

Panda chilling out

### DIFF
--- a/pkg/compose/printer.go
+++ b/pkg/compose/printer.go
@@ -21,8 +21,6 @@ import (
 	"fmt"
 
 	"github.com/docker/compose/v2/pkg/api"
-
-	"github.com/sirupsen/logrus"
 )
 
 // logPrinter watch application containers an collect their logs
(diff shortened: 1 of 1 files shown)


laurazard:
When I use `--exit-code-from` I expect to explicitly see the exit code, regardless of whether it's 0 or not. I wonder if we can instead use a different log type (not `logrus.Error`) to still output that without confusing the user

laurazard (approved):
LGTM

nicksieger (approved):
👍

PR #9787, opened 2022-08-24
Build and push docker/docker-compose images upon release

nicksieger:
**What I did**

Add release-image bake target and bin-image ci workflow for pushing docker/docker-compose image.

**Related issue**

butterfly emerging from cocoon

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -186,3 +186,9 @@ COPY --from=releaser /out/ /
 # see open-pr job in .github/workflows/docs.yml for more details
 FROM scratch AS docs-reference
 COPY docs/reference/*.yaml .
+
+FROM base AS release-image
+WORKDIR /root
+COPY --link --from=build /usr/bin/docker-compose /usr/bin/
+RUN mkdir -p /usr/local/lib/docker/cli-plugins
+RUN ln -s /usr/bin/docker-compose /usr/local/lib/docker/cli-plugins/
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,6 +22,7 @@ env:
   GO_VERSION: "1.18.5" # for non sandboxed e2e tests
   DESTDIR: "./bin"
   DOCKER_CLI_VERSION: "20.10.17"
+  REPO_SLUG: "docker/docker-compose"
 
 jobs:
   prepare:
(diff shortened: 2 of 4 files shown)


Dockerfile
```diff
@@ -186,3 +186,9 @@ COPY --from=releaser /out/ /
 # see open-pr job in .github/workflows/docs.yml for more details
 FROM scratch AS docs-reference
 COPY docs/reference/*.yaml .
+
+FROM base AS release-image
```
nicksieger, line 190: The base image is go/alpine, with docker and git installed in it. Anything else that we think should go in here?

crazy-max:
Don't think it makes sense to create a runnable image for compose. An artifact image with base scratch and the binary like buildx is enough imo. See also https://github.com/docker/buildx/pull/656. This way user can choose whatever base they want. Also update docs to show how to use it inside an image. See https://github.com/docker/buildx#dockerfile.

About hub slug might be better to have smth like buildx I think: `docker/compose-bin`. And makes clear in repo description that this is a non-runnable image: https://hub.docker.com/r/docker/buildx-bin.

thaJeztah:
> About hub slug might be better to have smth like buildx I think: docker/compose-bin. And makes clear in repo description that this is a non-runnable image: https://hub.docker.com/r/docker/buildx-bin.

👆 good point to use a different name for the image, as there's the existing "compose" image, which was a bit of a hack to run compose as container instead of installing (to be used in combination with this script: https://github.com/docker/compose/blob/1.29.2/script/run/run.sh)  and users may expect it to be a replacement for that.

laurazard:
We want to have a runnable `docker/docker-compose` image is so we can run it in a container for the dev-envs compose stack workflows, as opposed to the current scenario where code is being copied into the host and running compose from there.

An artifact image like you're suggesting could be useful for users to build upon, but I believe it still makes sense for us to have a runnable image (and we'd still want to have one regardless to move the above mentioned workflow from the host -> containers).

Dockerfile
thaJeztah: I notice we're installing docker from alpine's apk repositories, which are builds of docker built by alpine. Perhaps that's ok for testing, but we definitely should _not_ ship those as part of images for end-users.

Having a "runnable" compose image with docker installed basically replicates the official `docker:cli` image; see https://github.com/docker-library/docker/pull/369

thaJeztah:
compose is now part of the docker, so the best option for that is the official `docker:cli` image I mentioned above.

Dockerfile
nicksieger: Good point, I think basing compose off the official docker cli image would be a lot better.

nicksieger: Oh, nevermind! Compose is in the latest docker:*-cli images. Then I'll see if we can use that.

thaJeztah:
Let me also /cc @tianon for help with that image; I know the current images for "cli" contain the daemon as well; this should change soon for the next engine release, but perhaps we can already update for the current release.

tianon:
> I know the current images for "cli" contain the daemon as well; this should change soon for the next engine release, but perhaps we can already update for the current release.

This was fixed in https://github.com/docker-library/docker/pull/369! :smile:

(and Compose v2 has been part of the images since https://github.com/docker-library/docker/pull/361, both as `docker compose` and as `docker-compose`, and it should stay aggressively up-to-date with the latest releases :+1:)

nicksieger:
> (and Compose v2 has been part of the images since [docker-library/docker#361](https://github.com/docker-library/docker/pull/361), both as `docker compose` and as `docker-compose`, and it should stay aggressively up-to-date with the latest releases 👍)

Excellent, thanks @tianon

Issue #9778, opened 2022-08-23
the `--exit-code-from` is 0 from a service that `depends_on` when another container exited


Lenormju:
**Description**

I want to use Compose to test my services. I use `--exit-code-from test` to get the exit code of my `test` service, which is very convenient.
As mentioned in the docs, `--exit-code-from` implies `--abort-on-container-exit`, which is very useful in my case : if any service crashes, the test can be considered a failure.
But if my `test` service has a `depends_on` on any other service, then I don't get the exit code anymore, just 0.

**Steps to reproduce the issue:**

1. file: `docker-compose.yml`
   ```yaml
   ---
   name: mre__exit_code_from__when_other_container_aborts
   services:
     safe:
       image: 'ubuntu:20.04'
       command: ['/usr/bin/bash', '-c', 'sleep 30']
     failure:
       image: 'ubuntu:20.04'
       command: ['/usr/bin/bash', '-c', 'sleep 2 ; echo "exiting" ; exit 42']
     test:
       image: 'ubuntu:20.04'
       command: ['/usr/bin/bash', '-c', 'sleep 30']
       depends_on: [safe]
   ...
   ```
2. run (linux) `docker compose down ; docker compose up --abort-on-container-exit --exit-code-from test ; echo "EXIT_CODE=$?" ; docker ps -a --filter 'name=mre__exit_code_from__when_other_container_aborts'`

**Describe the results you received:**
```shell
[+] Running 4/3
 ⠿ Network mre__exit_code_from__when_other_container_aborts_default      Created                                                                                                                                                    0.1s
 ⠿ Container mre__exit_code_from__when_other_container_aborts-failure-1  Created                                                                                                                                                    0.1s
 ⠿ Container mre__exit_code_from__when_other_container_aborts-safe-1     Created                                                                                                                                                    0.1s
 ⠿ Container mre__exit_code_from__when_other_container_aborts-test-1     Created                                                                                                                                                    0.1s
Attaching to mre__exit_code_from__when_other_container_aborts-failure-1, mre__exit_code_from__when_other_container_aborts-safe-1, mre__exit_code_from__when_other_container_aborts-test-1
mre__exit_code_from__when_other_container_aborts-failure-1  | exiting
mre__exit_code_from__when_other_container_aborts-failure-1 exited with code 42
Aborting on container exit...
[+] Running 3/3
 ⠿ Container mre__exit_code_from__when_other_container_aborts-test-1     Stopped                                                                                                                                                   10.4s
[...]

PR #9794, opened 2022-08-28
Correctly capture exit code when service has dependencies

laurazard:

**What I did**

Only capture exit code from `exit` events.

Previously, we were capturing exit codes from both `exit` and `stop` events. Usually this is not a problem, as all containers get killed/stop around the same time and after we get the last `exit` event, we stop streaming events.

However, if the service we're trying to get an exit code for depends on another, long lived service, which doesn't respond to a `SIGTERM` and hence stays alive for another `10s` (the default timeout before `SIGKILL` is sent), we keep streaming events for longer. During that time, the daemon sends an extra `stop` event, even for containers which have already exited. With the way we were handling events previously, this event's exit code (0, because `stop` events don't carry exit codes) would override the correct exit code, resulting in an incorrect exit code being reported.

**Related issue**

resolves https://github.com/docker/compose/issues/9778

butterfly

### DIFF
--- a/pkg/compose/printer.go
+++ b/pkg/compose/printer.go
@@ -93,11 +93,13 @@ func (p *printer) Run(ctx context.Context, cascadeStop bool, exitCodeFrom string
 							return 0, err
 						}
 					}
-					if exitCodeFrom == "" {
-						exitCodeFrom = event.Service
-					}
-					if exitCodeFrom == event.Service {
-						exitCode = event.ExitCode
+					if event.Type == api.ContainerEventExit {
+						if exitCodeFrom == "" {
+							exitCodeFrom = event.Service
+						}
+						if exitCodeFrom == event.Service {
+							exitCode = event.ExitCode
+						}
 					}
 				}
 				if len(containers) == 0 {
--- a/pkg/e2e/compose_up_test.go
+++ b/pkg/e2e/compose_up_test.go
@@ -22,6 +22,7 @@ import (
 	"time"
 
 	"gotest.tools/v3/assert"
+	"gotest.tools/v3/icmd"
 )
 
 func TestUpWait(t *testing.T) {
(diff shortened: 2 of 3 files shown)


thaJeztah:
> During that time, the daemon sends an extra stop event, even for containers which have already exited. 

I'm curious about this part; does the daemon do that by itself? Or is it because compose does another "kill" / "stop" API call?

laurazard:
@thaJeztah I spent a lot of time trying to find a root cause (I assumed we were revisiting that node and trying to stop it again, causing the extra event, and spent a lot of time looking for that) but as far as I can tell that just happens by itself. I made a little writeup in Notion I can share later with more logs, and inspected the daemon logs as well for extra calls but didn't find any.

thaJeztah:
Hm, yes; definitely curious. The even is sent here; https://github.com/moby/moby/blob/b3675e18393e4d9c13d836da38c67b8dc9966ecc/daemon/stop.go#L65

And (at a quick glance) would only happen once the containerstop func exits, and the ContainerStop() function calls it once as well (that's the function called by the "stop" API endpoint; https://github.com/moby/moby/blob/cf8b057100fa1f513dccee700f0190462b93eb15/api/server/router/container/container_routes.go#L245

laurazard:
@thaJeztah we (hopefully) are calling the `stop` endpoint once, and getting one `stop` event, so that seems to be in line with what is expected. What was causing the issue with the way Compose is handling events is we are receiving the `stop` event after the container has already exited and we've already received a `die` event.

Maybe my wording was unclear -- we're not receiving an extra `stop` event, we're receiving an extra event after the `die` event which is a `stop` event, which Compose was handling incorrectly

thaJeztah:
Ah, thanks! Yes, the die even would be the one to look for in that case

laurazard:
Right! I initially assumed the `stop` after the `die` was incorrect (hence looking for if we were making any extra call to the API) but then realized this happens for all services, we just mostly don't get them because we close the event stream after getting `die` for all service containers, which tends to happen at the same time (unless a service depends on another long-lived service that doesn't respond to `SIGTERM`).

thaJeztah:
We used to have a diagram in the docs for the events; https://github.com/moby/moby/issues/12164, but I think it got removed (probably it wasn't kept up to date);

laurazard:
Ohhh, that's really useful! Shame it's not there anymore/kept up to date.

nicksieger (approved):
Nice find!

milas (approved):

Issue #9796, opened 2022-08-29
Failure to start single service


oseemann:
**Description**

Defining two or more services in a compose file and then creating and starting a single one fails with `service "$name" has no container to start`.

**Steps to reproduce the issue:**
1. Example compose file:
```
services:
  pgsql:
    image: postgres
  redis:
    image: redis
```
2. `docker-compose create pgsql`
3. `docker-compose start pgsql`

**Describe the results you received:**

Start fails with `service "redis" has no container to start`.

**Describe the results you expected:**

`pgsql` service/container should start individually, `redis` service should not run.

**Output of `docker compose version`:**

```
Docker Compose version 2.10.2
```

**Output of `docker info`:**

```
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc., v0.8.2)
  compose: Docker Compose (Docker Inc., v2.7.0)
  extension: Manages Docker extensions (Docker Inc., v0.2.8)
  sbom: View the packaged-based Software Bill Of Materials (SBOM) for an image (Anchore Inc., 0.6.0)
  scan: Docker Scan (Docker Inc., v0.17.0)

Server:
 Containers: 47
  Running: 10
  Paused: 0
  Stopped: 37
 Images: 49
 Server Version: 20.10.17
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 2
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: runc io.containerd.runc.v2 io.containerd.runtime.v1.linux
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 10c12954828e7c7c9b6e0ea9b0c02b01407d3ae1
 runc version: v1.1.2-0-ga916309
 init version: de40ad0
 Security Options:
  seccomp
   Profile: default
  cgroupns
 Kernel Version: 5.10.104-linuxkit
 Operating System: Docker Desktop
 OSType: linux
 Architecture: aarch64
 CPUs: 4
 Total Memory: 7.667GiB
 Name: docker-desktop
[...]

PR #9797, opened 2022-08-29
 Only attempt to start specified services on `compose start [services]`

laurazard:
**What I did**

Only attempt to start specified services on `compose start [services]`

Add E2E tests

**Related issue**

fixes https://github.com/docker/compose/issues/9796

IMG_4078


pkg/e2e/start_stop_test.go
```diff
@@ -229,22 +229,35 @@ func TestStartStopMultipleServices(t *testing.T) {
 		cli.RunDockerComposeCmd(t, "down", "--remove-orphans", "-v", "-t", "0")
 	})
 
-	cli.RunDockerComposeCmd(t, "up", "-d", "--wait")
+	t.Run("starts/stops multiple services", func(t *testing.T) {
```
laurazard, line 232: Easier to look at the diff here with whitespace off, I just wrapped up a previous test in a `t.Run(...` and added a separate one

pkg/compose/start.go
```diff
@@ -69,6 +70,9 @@ func (s *composeService) start(ctx context.Context, projectName string, options
 		if err != nil {
 			return err
 		}
+		if !utils.StringContains(options.AttachTo, name) {
```
nicksieger, line 73: If the attach to array is empty, that should be interpreted as starting all services right?

```suggestion
		if len(options.AttachTo) > 0 && !utils.StringContains(options.AttachTo, name) {
```

nicksieger (commented):
One question otherwise LGTM

pkg/compose/start.go
laurazard: Yes 😅 . I just noticed the red CI and realized I missed that

laurazard:
Oh you're right, I hadn't seen them there. I'll add some, and in the meantime there's some edge cases this still misbehaves with (although better than the previous behaviour 😅 ) so I might make a few more changes.

pkg/compose/start.go
```diff
@@ -64,6 +71,10 @@ func (s *composeService) start(ctx context.Context, projectName string, options
 		})
 	}
 
+	for _, s := range project.Services {
+		println("lmao:", s.Name)
+	}
```
milas, line 76:
```suggestion
```
😂

pkg/e2e/start_stop_test.go
```diff
@@ -222,29 +222,46 @@ func TestStopAlreadyStopped(t *testing.T) {
...
 
-	res = cli.RunDockerComposeCmd(t, "start", "simple", "another")
-	for _, svc := range services {
-		startMsg := fmt.Sprintf("Container e2e-start-stop-svc-multiple-%s-1  Started", svc)
-		assert.Assert(t, strings.Contains(res.Stderr(), startMsg),
-			fmt.Sprintf("Missing start message for %s\n%s", svc, res.Combined()))
-	}
+	t.Run("starts one service out of many, and it's dependencies", func(t *testing.T) {
+		cli.RunDockerComposeCmd(t, "down")
+		cli.RunDockerComposeCmd(t, "-f", "./fixtures/start-stop/start-stop-deps.yaml", "create", "desired")
```
milas, line 251: Nit: I'd probably make this its own test since it uses a different fixture (which would also help reduce some of the boilerplate of passing `-f ./fixtures/....` everywhere)

(Personally, I think sub-tests in Go are best used for _unit_ tests that have a bunch of cases/permutations - for E2E/more complicated tests, I think it ends up being more confusing/error-prone than helpful, but that's why this is a nit, it's a personal preference 😛)

milas (changes requested):
LGTM other than the errant debugging output!

pkg/e2e/start_stop_test.go
laurazard: No I agree. I think we could do with big refactors on the E2E side, it's especially confusing to me when subtests are used for setting stuff up/tearing them down and not as self contained tests. But I guess we'll be making changes to our E2E tests :)

pkg/compose/start.go
laurazard: woops 😅

pkg/compose/dependencies_test.go
```diff
@@ -69,3 +71,181 @@ func TestInDependencyReverseDownCommandOrder(t *testing.T) {
 	require.NoError(t, err, "Error during iteration")
 	require.Equal(t, []string{"test1", "test2", "test3"}, order)
 }
+
+func TestBuildGraph(t *testing.T) {
```
milas, line 75: Meant to comment on the last review but 🙌 for adding more test coverage here!

milas (approved):
LGTM!

Issue #9705, opened 2022-08-02
V2:docker compose ps command and docker-compose ps command get content not the same


LatreviousYI:
**Description**

**Steps to reproduce the issue:**
1.docker compose -f /home/kerwin/process_file/test_fastapi_v2_update/yaml/test_fastapi_v2_update.yaml ps
2.docker-compose -f /home/kerwin/process_file/test_fastapi_v2_update/yaml/test_fastapi_v2_update.yaml ps
3.

**Describe the results you received:**
output of `docker compose`
```
# docker compose -f test_fastapi_v2_update.yaml ps
NAME                COMMAND                  SERVICE             STATUS              PORTS
test_fastapi        "bash -c 'python tes…"   test_fastapi        running             0.0.0.0:6898->8000/tcp, :::6898->8000/tcp
test_fastapi_v1     "bash -c 'python tes…"   test_fastapi_v1     running             0.0.0.0:9898->8000/tcp, :::9898->8000/tcp
test_fastapi_v2     "bash -c 'python tes…"   test_fastapi_v2     running             0.0.0.0:9899->8000/tcp, :::9899->8000/tcp
```

**Describe the results you expected:**
output of `docker-compose`
```
# docker-compose -f test_fastapi_v2_update.yaml ps
     Name                Command           State                    Ports                  
-------------------------------------------------------------------------------------------
test_fastapi_v1   bash -c python test.py   Up      0.0.0.0:9898->8000/tcp,:::9898->8000/tcp
test_fastapi_v2   bash -c python test.py   Up      0.0.0.0:9899->8000/tcp,:::9899->8000/tcp
```

**Additional information you deem important (e.g. issue happens only occasionally):**

Compose file:
```
# cat test_fastapi_v2_update.yaml
services:
  test_fastapi_v1:
    command: bash -c "python test.py"
    container_name: test_fastapi_v1
    image: python:fastapi_comp
    logging:
      driver: json-file
      options:
        max-file: '3'
        max-size: 500mb
    ports:
    - 9898:8000
    restart: always
    volumes:
    - /home/kerwin/test_fastpi:/home/kerwin/test_fastpi
    working_dir: /home/kerwin/test_fastpi
  test_fastapi_v2:
    command: bash -c "python test.py"
    container_name: test_fastapi_v2
    image: python:fastapi_comp
    logging:
      driver: json-file
      options:
        max-file: '3'
        max-size: 500mb
    ports:
    - 9899:8000
    restart: always
    volumes:
[...]

PR #9811, opened 2022-09-06
logs: filter to services from current Compose file

### LINKED ISSUE #9801: Logs subcommand does not filter to services from current Compose file
docker compose -f ./ProcessFile/AIServer/yaml/AIServer.yaml logs command have same problem,and Compose v2 2.10.2 didn't fix it
https://github.com/docker/compose/issues/9705

milas:
**What I did**
When using the file model, only attach to services referenced in the active Compose file.

For example, let's say you have `compose-base.yaml` and `compose.yaml`, where the former only has a
subset of the services but are both run as part of the same named project.

Project based command:
```shell
docker compose -p myproj logs
```
This should return logs for active services based
on the project name, regardless of Compose file
state on disk.

File based command:
```shell
docker compose --file compose-base.yaml logs
```
This should return logs for ONLY services that are defined in `compose-base.yaml`. Any other services are considered 'orphaned' within the context of the command and should be ignored.

**Related issue**
See also #9705.

Fixes #9801.

photo of a quokka standing on its hind legs

### DIFF
--- a/cmd/compose/logs.go
+++ b/cmd/compose/logs.go
@@ -63,12 +63,13 @@ func logsCommand(p *projectOptions, backend api.Service) *cobra.Command {
 }
 
 func runLogs(ctx context.Context, backend api.Service, opts logsOptions, services []string) error {
-	projectName, err := opts.toProjectName()
+	project, name, err := opts.projectOrName()
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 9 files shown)


nicksieger (approved):
LGTM, agree that it would be nice to have project processing and filtering like this applied in the command, but this works and isn't too onerous.

PR #9812, opened 2022-09-06
ci: upgrade to Go 1.19.1

milas:
**What I did**
Go released 1.18.6 + 1.19.1 today which fix a couple CVEs. (`golang.org/x/net` also has a related fix.)

This jumps from 1.18.5 -> 1.19.1 since it was on the to-do list regardless :)

**Related issue**
https://groups.google.com/g/golang-announce/c/x49AQzIVX-s

Photo of a rusty-spotted cat moving stealthily along a log


.github/workflows/ci.yml
```diff
@@ -19,7 +19,6 @@ on:
         default: "false"
 
 env:
-  GO_VERSION: "1.18.5" # for non sandboxed e2e tests
```
thaJeztah, line 22: Note that the release pipeline is still on 1.18; does the CI matrix also test against that version?

glours (approved):

thaJeztah:
You'll have to update golangci-lint to v1.49 probably; note that there's some linters being deprecated (and/or replaced); https://github.com/docker/cli/pull/3768

```
> [lint 1/1] RUN --mount=type=bind,target=.     --mount=type=cache,target=/root/.cache     --mount=from=golangci-lint,source=/usr/bin/golangci-lint,target=/usr/bin/golangci-lint     golangci-lint run --build-tags "e2e,kube" ./...:
#21 9.836 panic: load embedded ruleguard rules: rules/rules.go:13: can't load fmt
#21 9.836 
#21 9.836 goroutine 1 [running]:
#21 9.836 github.com/go-critic/go-critic/checkers.init.22()
#21 9.836 	github.com/go-critic/go-critic@v0.6.3/checkers/embedded_rules.go:47 +0x4b4
```

.golangci.yml
```diff
@@ -21,13 +20,15 @@ linters:
...
     - typecheck
     - unconvert
     - unparam
     - unused
-    - varcheck
 linters-settings:
+  revive:
+    rules:
+      - name: package-comments
+        disabled: true
```
thaJeztah, line 31: Ah, you found the cleaner way to disable it 😂

.github/workflows/ci.yml
```diff
@@ -143,7 +142,8 @@ jobs:
         name: Set up Go
         uses: actions/setup-go@v3
         with:
-          go-version: ${{ env.GO_VERSION }}
+          go-version-file: 'go.mod'
```
ulyssessouza, line 145: Reading minds! I was about to make this change!

.golangci.yml
milas: I ❤️ the JSON schema support in IntelliJ for stuff like this!

(Also, I totally would have just copy-pasted from your PR but GitHub was erroring out on me at the time 🙈)

.github/workflows/ci.yml
milas: The release pipeline in `docker/docker-ce-packaging`, right? (IIRC we settled on it calls `make build` from the `Makefile` in this repo, which does a "local" `go build` inside of its container but not using `buildx bake` for the actual Compose build)

How do we generally handle that for other projects packaged by it? Something like this?
* Upgrade project CI to default to Go `n+1`
* Continue running project CI against Go `n` & `n+1`
* Once all projects on Go `n+1`, change packaging to use Go `n+1`
* Drop Go `n` from project CI

cc @glours if you have opinions

PR #9816, opened 2022-09-07
Fixed #9815 Forward ignore_pull_failures parameter to parallel_pull

### LINKED ISSUE #9815: ignore_pull_failures is ignored when using parallel_pull=True
We are using docker-compose from another Python-Skript. We know that it is not officially supported, still the error is quite simple to fix and may occur in other situations as well.

When using the Method `Project.pull` the Parameter `igore_pull_failures` will be ignored when using `parallel_pull=True` (which is default).
So `project.pull(ignore_pull_failures=True)` will still throw errors.

```
    def pull(self, service_names=None, ignore_pull_failures=False, parallel_pull=True, silent=False,
             include_deps=False):
        services = self.get_services(service_names, include_deps)

        if parallel_pull:
            self.parallel_pull(services, silent=silent)
[...]
    def parallel_pull(self, services, ignore_pull_failures=False, silent=False):
```
As `parallel_pull` supports the parameter too this should be easy to fix.

Berndone:
Signed-off-by: Victor Horvath <horvath.victor@web.de>

**What I did**

**Related issue**

### DIFF
--- a/compose/project.py
+++ b/compose/project.py
@@ -754,7 +754,10 @@ class Project:
         services = self.get_services(service_names, include_deps)
 
         if parallel_pull:
-            self.parallel_pull(services, silent=silent)
+            self.parallel_pull(
+                services, silent=silent,
+                ignore_pull_failures=ignore_pull_failures,
+            )
 
         else:
             must_build = []


glours:

PR #9792, opened 2022-08-26
Revert "Apply newly loaded envvars to `DockerCli` and `APIClient`"

milas:
**What I did**
Revert #9745 that introduced support for reading `DOCKER_HOST` & friends from `.env` files but broke TLS options.

**Related issue**
Fixes #9789.

PR #9817, opened 2022-09-07
Apply newly loaded envvars to "DockerCli" and "APIClient"

ulyssessouza:
**What I did**
Re-evaluate DockerCli and APIClient after reading the environment file. I can contain DOCKER_HOST and/or DOCKER_CONTEXT so the DockerCli passed by docker/cli has to be re-evaluated. Also checks for DOCKER_CERT_PATH and DOCKER_TLS_VERIFY.

**Related issue**
Resolves https://github.com/docker/compose/issues/9210

**Note**: This is a "retry" of https://github.com/docker/compose/pull/9745 that was reverted in https://github.com/docker/compose/pull/9792


go.mod
```diff
@@ -150,3 +152,5 @@ replace (
 	k8s.io/apimachinery => k8s.io/apimachinery v0.22.4
 	k8s.io/client-go => k8s.io/client-go v0.22.4
 )
+
+replace github.com/cnabio/cnab-go => github.com/ulyssessouza/cnab-go v0.0.0-20220907140501-5e7dd4d50442
```
ulyssessouza, line 156: Depends on https://github.com/cnabio/cnab-go/pull/293

milas (approved):
LGTM

Though I do think we should look to getting a fix into `docker/cli` itself; never a good feeling to be working around our own projects 🙈

glours (approved):
LGTM

PR #9819, opened 2022-09-07
build: label built images for reliable cleanup on `down`

milas:
**What I did**
When running `compose down`, the `--rmi` flag can be passed, which currently supports two values:
 * `local`: remove any _implicitly-named_ images that Compose built
 * `all`  : remove any named images (locally-built or fetched from a remote repo)

Removing images in the `local` case can be problematic, as it's historically been done via a fair amount of inference over the Compose model. Additionally, when using the "project-model" (by passing `--project-name` instead of using a Compose file), we're even more limited: if no containers for the project are running, there's nothing to derive state from to perform the inference on.

As a first pass, we started labeling _containers_ with the name of the locally-built image associated with it (if any) in #9715. Unfortunately, this still suffers from the aforementioned problems around using actual state (i.e. the containers might no longer exist) and meant that when operating in file mode (the default), things did not behave as expected: the label is not available in the project since it only exists at runtime.

Now, with these changes, Compose will label any images it builds with project metadata. Upon cleanup during `down`, the engine image API is queried for related images and matched up with the services for the project. As a fallback for images built with prior versions of Compose, the previous approach is still taken.

**Related issue**
 * https://github.com/docker/compose/issues/9655
 * https://github.com/docker/compose/pull/9715
 * [EN-224](https://docker.atlassian.net/browse/EN-224)

a seal swimming towards the camera


pkg/compose/down.go
```diff
@@ -190,17 +201,108 @@ func (s *composeService) removeNetwork(ctx context.Context, name string, w progr
 	return nil
 }
 
-func (s *composeService) getServiceImagesToRemove(options api.DownOptions, project *types.Project) map[string]struct{} {
-	images := map[string]struct{}{}
+//nolint:gocyclo
```
milas, line 204: 😬 😬 😬 

I've tried to overly comment this method, and it's actually quite procedural despite the high cyclomatic complexity, but I'm interested in suggestions and am happy to refactor, as I do realize it's quite long.

nicksieger, line 204: 😬

milas: Limits, like rules, are meant to be broken!!

Just kidding - good call re: `ImageRemover` helper struct, I'd contemplated something like that, will do some refactoring now

glours (approved):
LGTM

PR #9823, opened 2022-09-08
Add Codecov

ulyssessouza:
**What I did**
Add code coverage

**Related issue**


nicksieger (approved):
LGTM, I think. So we're just grabbing coverage for e2e tests at this point, yes? A little refactoring to make some of the tests run more smoothly with coverage turned on?

Makefile
```diff
@@ -26,7 +26,8 @@ ifeq ($(UNAME_S),Darwin)
 	MOBY_DOCKER=/Applications/Docker.app/Contents/Resources/bin/docker
 endif
 
-TEST_FLAGS?=
+TEST_COVERAGE_FLAGS = -race -coverprofile=coverage.out -covermode=atomic
```
ulyssessouza, line 29: @nicksieger 
Note that it adds "-race" to detect data races.
The changes in the tests are related to this.

glours (commented):
I would prefer to use the existing [`test-coverage` target of the Dockerfile](https://github.com/docker/compose/blob/61845dd7816cec9d42c715c8b157f88dfe78fd54/Dockerfile#L98) to pick up the result and just upload it like it's done in [buildx](https://github.com/docker/buildx/blob/25ceb9067893be764f9ab439722835eece2fe169/.github/workflows/build.yml#L45)

ulyssessouza:
@glours Hmmm... In the end, what happens is that `e2e` tests need to connect to the docker socket and I get an error:
`Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?`


The thing is that we cannot mount it during the build. The other target can use it because it only runs unit tests.
I even tried a `RUN --mount=target=/var/run/docker.sock,rw` but it doesn't work.

There is the solution of baking an e2e image and running it with `-v /var/run/docker.sock:/var/run/docker.sock` to be able to access it. But I think that would go around all the idea of having `buildx` set up.

glours:
@crazy-max any idea ☝️ Do you have the same kind of e2e tests on the buildx side and if yes how did you adress this issue?

crazy-max:
`test-coverage` in Dockerfile is just for unit test but yeah sandboxed unit test makes sense so [this step](https://github.com/docker/compose/blob/c4d79e60b6d27dd437cea8bc555086abceca3606/.github/workflows/ci.yml#L115-L122) could output coverage like we do in buildx: https://github.com/docker/buildx/blob/294421db9c3087561f44ff5a8dcdefde758e3c22/Dockerfile#L41-L45. See also bake target: https://github.com/docker/buildx/blob/294421db9c3087561f44ff5a8dcdefde758e3c22/docker-bake.hcl#L97-L101.

e2e tests need a runnable container so can't do this during build. Also mounting sockets is not supported atm on BuildKit afaik.

ulyssessouza:
@glours I think that in this case we can stay the way it is. WDYT?

laurazard:
We can either keep it as is, or change the `test-coverage` Dockerfile target to run some other step without the e2e, it doesn't make sense to run the end-to-end tests just for test coverage.

ulyssessouza:
@laurazard Actually `test-coverage` doesn't run e2e tests, only unit tests.
What we need to decide here is to what are we interested in mesuring.
- Only unit tests
- Only e2e tests
- Both tests (Codecov can [Merge coverage reports](https://docs.codecov.com/docs/merging-reports))

laurazard (approved):
Ahh okay! LGTM then

PR #9824, opened 2022-09-09
🥒 Cucumber PoC 🥒  

laurazard:
**What I did:**

I got a bit curious from what we discussed regarding a PoC using Cucumber, so I started playing around with it! It needs a ton of refactoring/reorganizing (it's currently really ugly), and a good look regarding the testing utils and parallelization.

So far, steps implemented are:
- `(Given) a compose file [xxx]`: sets the compose file to get piped into the compose command
- `(When) I run "compose [command]"`: runs the specified compose command with the given fixture piped in
- `(Then) output contains [xxx]`: checks that the output of the last command contains [xxx]
- `(...) exit code is [i]`: checks that the exit code of the last command is [i]
- `(...) [service] service is [status]`: runs `compose ps`, checks that the specific service is the desired status

This allows us to define tests such as:
```Gherkin
Feature: Simple service up

Background:
    Given a compose file
        """
        services:
          simple:
            image: alpine
            command: top
        """

Scenario: compose up
    When I run "compose up -d"
    Then the output contains "simple-1  Started"
    And service "simple" is "running"
```

Run the cucumber feature tests with `go test -v -run ^TestCucumberFeatures$ github.com/docker/compose/v2/pkg/e2e/cucumber`

Output should look something like:

image


laurazard:
For comparison:
```Go
t.Run("abort-on-container-exit", func(t *testing.T) {
		res := c.RunDockerComposeCmdNoCheck(t, "-f", "./fixtures/cascade-stop-test/compose.yaml", "--project-name", projectName, "up", "--abort-on-container-exit")
		res.Assert(t, icmd.Expected{ExitCode: 1, Out: `should_fail-1 exited with code 1`})
		res.Assert(t, icmd.Expected{ExitCode: 1, Out: `Aborting on container exit...`})
	})

	t.Run("exit-code-from", func(t *testing.T) {
		res := c.RunDockerComposeCmdNoCheck(t, "-f", "./fixtures/cascade-stop-test/compose.yaml", "--project-name", projectName, "up", "--exit-code-from=sleep")
		res.Assert(t, icmd.Expected{ExitCode: 137, Out: `should_fail-1 exited with code 1`})
		res.Assert(t, icmd.Expected{ExitCode: 137, Out: `Aborting on container exit...`})
	})

	t.Run("exit-code-from unknown", func(t *testing.T) {
		res := c.RunDockerComposeCmdNoCheck(t, "-f", "./fixtures/cascade-stop-test/compose.yaml", "--project-name", projectName, "up", "--exit-code-from=unknown")
		res.Assert(t, icmd.Expected{ExitCode: 1, Err: `no such service: unknown`})
	})

	t.Run("down", func(t *testing.T) {
		_ = c.RunDockerComposeCmd(t, "--project-name", projectName, "down")
	})
```

turns into:
```Gherkin
Feature: Stop

Background:
    Given a compose file
        """
        services:
          should_fail:
            image: alpine
            command: ls /does_not_exist
          sleep: # will be killed
            image: alpine
            command: ping localhost
        """

Scenario: Cascade stop
    When I run "compose up --abort-on-container-exit"
    Then the output contains "should_fail-1 exited with code 1"
    And the output contains "Aborting on container exit..."
    And the exit code is 1

Scenario: Exit code from
    When I run "compose up --exit-code-from sleep"
    Then the output contains "should_fail-1 exited with code 1"
    And the output contains "Aborting on container exit..."
    And the exit code is 137

Scenario: Exit code from unknown service
    When I run "compose up --exit-code-from unknown"
    Then the output contains "no such service: unknown"
    And the exit code is 1
```

nicksieger (commented):
Really like where this is going so far. I think it's worth trying out.

I'm wondering if deleting the corresponding existing tests that these cucumber tests replace would help with evaluating it? That way we can compare the current e2e test with a cucumber version in the same PR.

go.mod
```diff
@@ -2,6 +2,8 @@ module github.com/docker/compose/v2
 
 go 1.18
 
+replace github.com/cucumber/godog => github.com/laurazard/godog v0.0.0-20220922095256-4c4b17abdae7
```
nicksieger, line 5: What's behind the need for this replace, out of curiosity?

laurazard: `cucumber/godog` currently allows for passing in a `testing.T` instance to run scenarios as subtests with, but it does not make that `testing.T` instance available to the scenario initializer, which was needed for passing to our `CLI`. I'll make a PR to see if I can contribute that upstream soon :)

laurazard:
Thanks! :)

I removed `cascade_stop_test.go` which was replaced by `stop.feature` and `compose_down_test.go` which was replaced by `down.feature`, so the comparison is clearer in the PR.

go.mod
```diff
@@ -131,6 +133,7 @@ require (
 
 require (
 	github.com/cenkalti/backoff/v4 v4.1.2 // indirect
+	github.com/cucumber/godog v0.12.5
```
thaJeztah, line 136: Would it make sense to have a separate module for the e2e tests, so that these test dependencies don't become a dependency for the main module?

We took that approach in containerd, where we then replaced the main module  with a path, to make sure we use the code from the branch where it's needed ; https://github.com/containerd/containerd/blob/main/integration/client/go.mod#L79

laurazard: We can! I don't think that would be a large change, and it would allow us to separate test dependencies better. I wonder what the rest of the team thinks about that (cc @glours @ulyssessouza @milas @nicksieger).

Also whether we can/should leave splitting that out for another PR or we'd rather do it here.

I do wish Go had better tooling around that/test dependencies 😅

nicksieger: That occurred to me too, we already did that for dev-envs and it seems to work well. Also would make sense in light of https://github.com/compose-spec/conformance-tests if we decide to consolidate them.

laurazard: Sounds good! I went ahead and moved the new tests into an `e2e` module, and then separately we can move the others there too :)

laurazard:
From [here](https://github.com/docker/compose/pull/9824#discussion_r978066244), moved the new tests into a new `e2e` module to separate out test dependencies.

go.mod
thaJeztah: Thanks! So, yes, I think the "ideal" at some point would be to (e.g) have a docker image with the integration tests compiled in, which could be run with a compose binary and docker socket mounted; this would also allow for (e.g.) https://github.com/moby/moby to run the latest e2e/integration tests as part of CI.

e2e/go.mod
```diff
@@ -0,0 +1,168 @@
...
+	k8s.io/utils v0.0.0-20220210201930-3a6ce19ff2f9 // indirect
+	sigs.k8s.io/structured-merge-diff/v4 v4.2.1 // indirect
+	sigs.k8s.io/yaml v1.2.0 // indirect
+)
+
+replace (
+	github.com/docker/cli => github.com/docker/cli v20.10.3-0.20220309205733-2b52f62e9627+incompatible
+	github.com/docker/docker => github.com/docker/docker v20.10.3-0.20220309172631-83b51522df43+incompatible
+
+	github.com/opencontainers/runc => github.com/opencontainers/runc v1.1.2 // Can be removed on next bump of containerd to > 1.6.4
```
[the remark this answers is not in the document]
laurazard: It was throwing some errors without it, but newer versions might be fine 😅

e2e/go.mod
```diff
@@ -0,0 +1,168 @@
...
+	k8s.io/client-go v0.24.1 // indirect
+	k8s.io/klog/v2 v2.60.1 // indirect
+	k8s.io/utils v0.0.0-20220210201930-3a6ce19ff2f9 // indirect
+	sigs.k8s.io/structured-merge-diff/v4 v4.2.1 // indirect
+	sigs.k8s.io/yaml v1.2.0 // indirect
+)
+
+replace (
+	github.com/docker/cli => github.com/docker/cli v20.10.3-0.20220309205733-2b52f62e9627+incompatible
+	github.com/docker/docker => github.com/docker/docker v20.10.3-0.20220309172631-83b51522df43+incompatible
```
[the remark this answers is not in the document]
laurazard: Sounds good! That's definitely a concern now having two separate sets of dependencies to keep in sync. I'll take a look at it, thanks :)

go.mod
laurazard: Oooh, I like that idea. I'll keep that in mind for an improvement soon.

e2e/go.mod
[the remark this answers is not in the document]
laurazard: @thaJeztah took a look at it and added something [here](https://github.com/docker/compose/pull/9824/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R128-R131) to verify it, using the script from containerd :)

laurazard:
The CI isn't triggering 😢 Cursory google search seems to suggest I close and reopen the PR, so I'm going to try that.

verify-go-modules.sh
```diff
@@ -0,0 +1,112 @@
...
+    echo "${k} has an entry in root go.mod replace section, but is missing from" \
+      " replace section in $1/go.mod"
+    ERRORS=$(( ERRORS + 1 ))
+  fi
+done
+
+if [ "$ERRORS" -ne 0 ]; then
+  echo "Found $ERRORS error(s)."
+  exit 1
+fi
```
glours, line 112: nit: missing new line 😉

glours (approved):
Looks cool, we'll need to figure out how to import fixtures for tests using huge compose files

verify-go-modules.sh
```diff
@@ -0,0 +1,112 @@
...
+    fi
+  fi
+done
+
+# iterate through the second go.mod's replace section and ensure that all items
+# have the same values in the root go.mod's replace section. Except for the
+# containerd/containerd which we know will be different
+for k in "${!map_replaces_2[@]}"
+do
+  if [[ "${k}" == "github.com/containerd/containerd"* ]]; then
```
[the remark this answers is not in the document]
laurazard: Thanks :)

milas (approved):
🥒 🥒 🥒 🥒

laurazard:
🥳

PR #9836, opened 2022-09-13
clean service command if entrypoint is overrided in run command

### LINKED ISSUE #9622: `docker compose run --entrypoint=EXECUTABLE SERVICENAME` fails to clear CMD
---------------------------------------------------
BUG REPORT INFORMATION
---------------------------------------------------

**Description**

If I run a one-off container based on a service definition using the `run` subcommand and override the entrypoint, Compose 2.6.0 fails to clear the default command (default arguments to the default entrypoint) as well. As far as I can think of, every other Docker interface to override the entrypoint also clears CMD:
* `docker-compose run` in v1
*  the [`run` subcommand](https://docs.docker.com/engine/reference/run/#entrypoint-default-command-to-execute-at-runtime) of Docker itself:
> **Note:** Passing `--entrypoint` will clear out any default command set on the image (i.e. any `CMD` instruction in the Dockerfile used to build it).
* Compose YAML files according to the [Compose spec](https://github.com/compose-spec/compose-spec/blob/5260c1bc8ee5e93ce88412bb460924e9535d9080/spec.md#entrypoint):
> Compose implementations MUST clear out any default command on the Docker image - both `ENTRYPOINT` and `CMD` instruction in the Dockerfile - when `entrypoint` is configured by a Compose file.
* [Older versions](https://docs.docker.com/compose/compose-file/compose-file-v3/#entrypoint) of the Compose YAML file format:
>  **Note:** Setting `entrypoint` both overrides any default entrypoint set on the service’s image with the `ENTRYPOINT` Dockerfile instruction, _and_ clears out any default command on the image - meaning that if there’s a `CMD` instruction in the Dockerfile, it is ignored.

1. Define a service with both an entrypoint and a default command to pass to it, e.g.
```yaml
services:
  database:
    image: mysql:8.0
    command: ["--help"]
```
2. Run a docker compose run command that overrides the entrypoint with a different executable, e.g.
```console
$ docker compose run --rm --entrypoint=ls database
```

The new entrypoint, `ls`, responds as if it was passed the default arguments intended for the original entrypoint, namely the `--help`, flag, and prints a usage message.`

I expected the new entrypoint, `ls`, to be run without arguments and list the files in the container's working directory.

None.

```
Docker Compose version v2.6.0
```

**Output of `docker info`:**

Snippet:
```
Client:
 Context:    default
 Debug Mode: false
 Plugins
  buildx: Docker Buildx (Docker Inc., v0.8.2-docker)
  compose: Docker Compose (Docker Inc., v2.6.0)

Server:
 Containers: 0
  Running: 0
  Paused: 0
  Stopped: 0
 Images: 10
 Server Version: 20.10.17
[...]
 Kernel Version: 5.4.0-121-generic
 Operating System: Linux Mint 20.2
 OSType: linux
 Architecture: x86_64
[...]
```

—

glours:
**What I did**
re-init default service command when the `--entrypoint` is used with `docker compose run`

**Related issue**
fixes #9622 

image

### DIFF
--- a/pkg/compose/run.go
+++ b/pkg/compose/run.go
@@ -112,6 +112,9 @@ func applyRunOptions(project *types.Project, service *types.ServiceConfig, opts 
 	}
 	if opts.Entrypoint != nil {
 		service.Entrypoint = opts.Entrypoint
+		if len(opts.Command) == 0 {
+			service.Command = []string{}
+		}
 	}
 	if len(opts.Environment) > 0 {
 		cmdEnv := types.NewMappingWithEquals(opts.Environment)


milas (commented):
What do you think about doing this in `applyRunOptions` instead?

I'm unclear why this logic seems split between `runOptions::apply` and `applyRunOptions` in the first place...but since the "other" func is already messing with cmd/entrypoint, it seems the safer place to co-locate this code with

https://github.com/docker/compose/blob/1ed37ef7bdc52f21585ac74e9a89e5d5d3b9adfe/pkg/compose/run.go#L104-L106

https://github.com/docker/compose/blob/1ed37ef7bdc52f21585ac74e9a89e5d5d3b9adfe/pkg/compose/run.go#L113-L115

PR #9840, opened 2022-09-14
update compose-go version to v1.5.1

glours:
**What I did**
Bump `compose-go` version to `v1.5.1`

**Related issue**
N/A

image


StefanScherer (approved):
LGTM

PR #9844, opened 2022-09-14
Fix variable defenition

mo1ein:
Fix variable definition in create.go.


ndeloof:
This is not required. `var volumes []string` define a variable with `nil` value, this won't prevent use of `append`, `len` or `for`. We use this pattern in many places within compose codebase.

PR #9847, opened 2022-09-14
keep the platform defined at service level during build if no build patforms provided

glours:

**What I did**
Fix an issue when the compose file defines a `service.platform` and don't provide `services.build.platforms`

**Related issue**
https://github.com/docker/compose/pull/9729#issuecomment-1246748144

image


StefanScherer (approved):
LGTM

Issue #9843, opened 2022-09-14
`depends_on: condition: service_healthy` is not respected when using `volumes_from`


oalagtash:
**Description**

When running `docker  compose up -d` `test` service will be started without waiting for `db` to be healthy. However, if I remove the section `volumes_from`, the `depends_on` works as expected.

**Steps to reproduce the issue:**

In docker-compose.yml:
``` yaml
version: "3.9"
services:
  db:
    image: redis

    healthcheck:
      test: "exit 0"
  test:
    image: nginx
    volumes_from:
      - db
    depends_on:
      db:
        condition: service_healthy
``` 

 run: `docker compose up -d`

**Describe the results you received:**

```shell
❯ docker compose down --remove-orphans && docker compose up -d && docker compose ps                                                                                                                          ─╯
[+] Running 3/3
 ⠿ Network in-buildsupport_default   Created                                                                                                                                                               0.0s
 ⠿ Container in-buildsupport-db-1    Started                                                                                                                                                               0.3s
 ⠿ Container in-buildsupport-test-1  Started                                                                                                                                                               0.5s
NAME                     COMMAND                  SERVICE             STATUS               PORTS
in-buildsupport-db-1     "docker-entrypoint.s…"   db                  running (starting)   6379/tcp
in-buildsupport-test-1   "/docker-entrypoint.…"   test                running              80/tcp

```
As shown in the logs: `test` is already running while `db` is still starting.

**Describe the results you expected:**

`test` service should wait for `db` to be healthy before starting.
Should look like this:
``` shell
❯ docker compose down --remove-orphans && docker compose up -d && docker compose ps                                                                                                                          ─╯
[+] Running 2/3
 ⠿ Network in-buildsupport_default   Created                                                                                                                                                               0.0s
 ⠿ Container in-buildsupport-db-1    Waiting                                                                                                                                                               4.8s
 ⠿ Container in-buildsupport-test-1  Created                                                                                                                                                               0.0s


```

**Output of `docker compose version`:**

```
❯ docker compose version                                                                                                                                                                                     
Docker Compose version v2.10.2
```

**Output of `docker info`:**

```
❯ docker info                                                                                                                                                                                            
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc., v0.9.1)
  compose: Docker Compose (Docker Inc., v2.10.2)
  extension: Manages Docker extensions (Docker Inc., v0.2.9)
[...]

PR #9849, opened 2022-09-15
Keep `depends_on` condition when service has `volumes_from`

laurazard:
**What I did**

Check if service already has a dependency condition before overwriting

**Related issue**

fixes https://github.com/docker/compose/issues/9843

Poodle-Moth


nicksieger (approved):
👍 LGTM

PR #9854, opened 2022-09-16
keep the platform defined via DOCKER_DEFAULT_PLATFORM during build if no build platforms provided

### LINKED ISSUE #9853: [BUG] use of DOCKER_DEFAULT_PLATFORM in addition to explicit service "platform:" results in error
**Description**

Using a compose file that has a service which specifies `platform:` and also supplying `DOCKER_DEFAULT_PLATFORM` 
 will result in an error. The application appears to be requesting two platforms to be targeted despite the fact that an explicit platform should override (not be combined with) a default one.

**I believe this is a BUG introduce by #8848,** which provided support for `DOCKER_DEFAULT_PLATFORM`, but used `append` to add it to a list of platforms, instead of letting explicit platforms override the default platform.

> **NOTE:** I do _not_ believe this is an issue with buildkit because I'm unable to reproduce the issue with a direct `docker build`/`docker buildx build` command. It's only reproducible when using `docker compose build`. Hopefully I've created this issue in the right place.

1. Create these required files in a temporary testing directory:
```yaml
# docker-compose.yml
version: '3.7'
services:
  test:
    image: test
    build:
      context: ./
    platform: linux/amd64
```
```
# Dockerfile
FROM busybox:stable
```
2. Run the following command:
    ```sh
    DOCKER_DEFAULT_PLATFORM=linux/amd64 docker compose build test
    ```
4. You will see the error below.

```
multiple platforms feature is currently not supported for docker driver. Please switch to a different driver (eg. "docker buildx create --use")
```

`compose` should invoke a `docker build` command that ignores `DOCKER_DEFAULT_PLATFORM` because, as a "default" value, it is expected to be overridden by an explicitly declared `platform`.

```
Docker Compose version v2.10.2
```

**Output of `docker info`:**

```
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  compose: Docker Compose (Docker Inc., v2.10.2)
  extension: Manages Docker extensions (Docker Inc., v0.2.9)
  scan: Docker Scan (Docker Inc., v0.19.0)

Server:
 Containers: 68
  Running: 47
  Paused: 0
  Stopped: 21
 Images: 99
 Server Version: 20.10.17
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 2
 Plugins:
  Volume: local
 Swarm: inactive
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 9cd3357b7fd7218e4aec3eae239db1f68a5a6ec6
 init version: de40ad0
 Security Options:
  seccomp
   Profile: default
  cgroupns
 Kernel Version: 5.10.124-linuxkit
 OSType: linux
 Architecture: aarch64
 CPUs: 10
 Total Memory: 19.5GiB
 Name: docker-desktop
 ID: VIB7:SF2J:RBST:3P2A:JBNY:GFYK:VW6B:MFGG:TPCZ:3UON:U4FW:7THL
 Debug Mode: false
 Labels:
 Experimental: false
 Insecure Registries:
  hubproxy.docker.internal:5000
  127.0.0.0/8
 Live Restore Enabled: false
```

**Additional Details:**
My use case is that I have multiple services in my compose config file, one of which requires an explicit platform to be used. The others can be build for any target platform. We use, e.g., `DOCKER_DEFAULT_PLATFORM=linux/amd64` to specify the desired target platform for services that support it, while the services with explicit `platform:` keys are expected to ignore `DOCKER_DEFAULT_PLATFORM`.

glours:
**What I did**
Don't throw an error when DOCKER_DEFAULT_PLATFORM is set and `service.build.platforms`  isn't set

**Related issue**
fixes #9853 

image

### DIFF
--- a/pkg/compose/build.go
+++ b/pkg/compose/build.go
@@ -368,7 +368,10 @@ func addPlatforms(project *types.Project, service types.ServiceConfig) ([]specs.
 		}
 		// User defined a service platform and no build platforms, so we should keep the one define on the service level
 		p, err := platforms.Parse(service.Platform)
-		return append(plats, p), err
+		if !utils.Contains(plats, p) {
+			plats = append(plats, p)
+		}
+		return plats, err
 	}
 
 	for _, buildPlatform := range service.Build.Platforms {
@@ -400,7 +403,7 @@ func getImageBuildLabels(project *types.Project, service types.ServiceConfig) ty
 func useDockerDefaultPlatform(project *types.Project, platformList types.StringList) ([]specs.Platform, error) {
 	var plats []specs.Platform
 	if platform, ok := project.Environment["DOCKER_DEFAULT_PLATFORM"]; ok {
-		if !utils.StringContains(platformList, platform) {
+		if len(platformList) > 0 && !utils.StringContains(platformList, platform) {
 			return nil, fmt.Errorf("the DOCKER_DEFAULT_PLATFORM value should be part of the service.build.platforms: %q", platform)
 		}
 		p, err := platforms.Parse(platform)


nicksieger (approved):
LGTM!

laurazard (approved):

PR #9863, opened 2022-09-20
Add `merge` GitHub Actions workflow to run tests on Windows and macOS runners

laurazard:
**What I did**

Add GitHub Actions workflow `merge` to run unit/e2e tests on Windows and Mac runners, after PRs are merged to `v2` and when tags such as `v*` are pushed.

Removed `/rebase` action since this is no longer necessary.

IMG_4645


pkg/e2e/up_test.go
```diff
@@ -1,3 +1,6 @@
+//go:build !windows
+// +build !windows
```
laurazard, line 2: This test does not compile on Windows due to:
```go
cmd.SysProcAttr = &syscall.SysProcAttr{Setpgid: true}
```
and surrounding logic

pkg/e2e/compose_test.go
```diff
@@ -135,6 +135,9 @@ func TestDownComposefileInParentFolder(t *testing.T) {
 }
 
 func TestAttachRestart(t *testing.T) {
+	if _, ok := os.LookupEnv("CI"); ok {
+		t.Skip("Skipping test on CI... flaky")
```
laurazard, line 139: Same for this test, and same regarding always passing on the M1 mac but being flaky on the Intel mac/Windows runner.

Makefile
```diff
@@ -18,13 +18,20 @@ VERSION ?= $(shell git describe --match 'v[0-9]*' --dirty='.m' --always --tags)
 GO_LDFLAGS ?= -s -w -X ${PKG}/internal.Version=${VERSION}
 GO_BUILDTAGS ?= e2e,kube
 
-UNAME_S := $(shell uname -s)
-ifeq ($(UNAME_S),Linux)
```
laurazard, line 22: Running `$(shell uname -s)` on Windows breaks, so we only do it if we're not on Windows

pkg/e2e/build_test.go
```diff
@@ -85,6 +86,51 @@ func TestLocalComposeBuild(t *testing.T) {
...
+		c.RunDockerComposeCmd(t, "--project-directory", "fixtures/build-test", "down")
+		c.RunDockerCmd(t, "rmi", "build-test-nginx")
+		c.RunDockerCmd(t, "rmi", "custom-nginx")
+	})
+}
+
+func TestBuildSSH(t *testing.T) {
+	if runtime.GOOS == "windows" {
+		t.Skip("Running on Windows. Skipping...")
+	}
```
laurazard, line 131: I split out the SSH tests out so I could disable them on Windows -- they were failing with:
```
Invalid empty ssh agent socket: Windows OpenSSH agent not available at \\.\pipe\openssh-ssh-agent. Enable the SSH agent service or set SSH_AUTH_SOCK.
```
We can take a look at that later, maybe set up an SSH agent on the Windows runner and re-enable these

milas: Windows does have a built-in OpenSSH Agent nowadays, so it's fairly straightforward to enable in `services.msc` if we want

pkg/e2e/compose_test.go
milas: Oof yeah I've been 👀 this test for a while, might take a look tomorrow to see if I can figure out where the race is

pkg/e2e/up_test.go
milas: Ah yeah, we've dealt with this same thing before in Tilt 🙃 

What we can do is make a `func SetProcessPgid(cmd *exec.Cmd)` in `proc_windows.go` that's a no-op and `proc_unix.go` that does the code ⬆️ and then call the helper. Annoying but 🤷 

(IIRC Windows always has the desired behavior re: killing process groups)

pkg/e2e/compose_test.go
milas: (To be clear though, I'm in favor of merging this with the `Skip()` and deal with cleanup after!)

.github/workflows/merge.yml
```diff
@@ -0,0 +1,81 @@
...
+        os: [desktop-windows, desktop-macos, desktop-m1]
+        # mode: [plugin, standalone]
+        mode: [plugin]
+    env:
+      GO111MODULE: "on"
+    steps:
+      - name: test-action-unsafe-checkout
+        if: ${{ contains(matrix.os, 'desktop-windows') }}
+        run: |
+          git config --global --add safe.directory C:/actions-runner/_work/compose/compose
```
milas, line 32: Should we just do this on the runner manually instead of putting it in the job?

(Or maybe there's a way to avoid it all together...?)

.github/workflows/merge.yml
```diff
@@ -0,0 +1,81 @@
...
+        with:
+          go-version-file: go.mod
+          cache: true
+          check-latest: true
+
+      - name: Check arch of go binary
+        run: |
+          which go
+          file $(which go)
+        if: ${{ !contains(matrix.os, 'desktop-windows') }}
```
milas, line 45: I think it's safe to get rid of this -- it was a holdover from when we had to do Go M1 installs manually before it was available from the action I think.

pkg/e2e/build_test.go
```diff
@@ -130,47 +176,12 @@ func TestLocalComposeBuild(t *testing.T) {
...
-		c.RunDockerComposeCmd(t, "--project-directory", "fixtures/build-test", "down")
-		c.RunDockerCmd(t, "rmi", "build-test-nginx")
-		c.RunDockerCmd(t, "rmi", "custom-nginx")
-	})
 }
 
 func TestBuildSecrets(t *testing.T) {
+	if runtime.GOOS == "windows" {
+		t.Skip("skipping test on windows")
+	}
```
milas, line 184: I'm slightly surprised this test is failing on Windows! Should be a fun one to investigate 🙈

milas (approved):
wahoo 🎉

.github/workflows/merge.yml
laurazard: Ahh, I meant to remove it and see if it was still an issue. I think it should be fine now, given we’ve ran it before (a few times)

.github/workflows/merge.yml
laurazard: Sounds good!

Makefile
```diff
@@ -18,13 +18,20 @@ VERSION ?= $(shell git describe --match 'v[0-9]*' --dirty='.m' --always --tags)
...
+ifeq ($(OS),Windows_NT)
+    DETECTED_OS = Windows
+else
+    DETECTED_OS = $(shell uname)
```
nicksieger, line 24: I think you want `uname -s` here so that it will give `Linux`
```suggestion
    DETECTED_OS = $(shell uname -s)
```

nicksieger (changes requested):

Makefile
laurazard: Ahhh I didn't realize I lost that moving things around, I'll add it back! (Although I think by default `uname` will return the same thing as `uname -s` -- or so `man uname` says on Alpine and Ubuntu)

nicksieger (approved):

PR #9867, opened 2022-09-22
Refactor `StringContains` and `contains` function usages to use `go` ones

ulyssessouza:
**What I did**
Refactor contains functions

### DIFF
--- a/cmd/compatibility/convert.go
+++ b/cmd/compatibility/convert.go
@@ -20,6 +20,8 @@ import (
 	"fmt"
 	"os"
 
+	"golang.org/x/exp/slices"
+
 	"github.com/docker/compose/v2/cmd/compose"
 )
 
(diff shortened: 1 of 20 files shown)


glours (commented):
I'm wondering if we shouldn't wait the end of the experimental period before using those functions? 🤔 
>Warning: Packages here are experimental and unreliable. Some may one day be promoted to the main repository or other subrepository, or they may be modified arbitrarily or even disappear altogether.

ulyssessouza:
@glours I was thinking the same. But when I saw the implementation... I cannot think about something that would be change enough for our simple case. There is also a "k8s.io/utils/strings/slices" implementation for `Contains`.
For me that's the same, since the idea of the PR is just not to have "yet another" implementation of a so common function.

ndeloof:
isn't "yet another (copy/paste) implementation for common function" the basis of Golang design ?
kidding appart, shall we keep this PR open?

ulyssessouza:
I'll close it... It's still not extracted from the experimental package and all the files are conflicting... 😞

PR #9871, opened 2022-09-23
Remove support for `DOCKER_HOST` in `.env` files

milas:
**What I did**
This reverts commit 126cb988c6f0c00a2a9887b8a39dc0907daec289, reversing changes made to b80222fb077436f2d4fcc62b3765d3a493891632.

Upon further review, we’ve decided to not support dynamically setting `DOCKER_HOST` via `.env` files in Compose v2.

Compose v2 is a Docker CLI plugin, and we want to ensure consistent behavior across all CLI plugins, which means using the "inherited" Docker API connection provided to the plugin, respecting all options, such as the [active context](https://docs.docker.com/engine/context/working-with-contexts/).

If you were relying on this functionality, I’d recommend taking a look at projects such as [direnv](https://direnv.net/), which can automatically set environment variables such as `DOCKER_HOST` or `DOCKER_CONTEXT` based on the current directory.

**Related issue**
Reverts #9817.
JIRA: [EN-255](https://docker.atlassian.net/browse/EN-255)

polar bear face-palming


nicksieger (approved):

ulyssessouza (approved):
F

PR #9886, opened 2022-09-29
ci: update docs repo path

milas:
**What I did**
Updated a string because the Docker docs now live at `docker/docs` instead of `docker/docker.github.io`.

a dog in a squirrel costume


glours (approved):
LGTM

milas:
The robot's fork is still [docker-tools-robot/docker.github.io](https://github.com/docker-tools-robot/docker.github.io), but that should be fine:
https://github.com/docker/compose/blob/25c4bcef853d4e250b84931092e87e50ec32993c/.github/workflows/docs.yml#L48

VictorBersy (approved):

PR #9904, opened 2022-10-10
fix"convert --images" output to right format

Windforce17:
Signed-off-by: Windforce17 <42843391+Windforce17@users.noreply.github.com>

**What I did**
According to docker images name rule: https://docs.docker.com/engine/reference/commandline/tag/#description `docker compose convert --images` should not contain underscores.

closes https://github.com/docker/compose/issues/9920

### DIFF
--- a/cmd/compose/convert.go
+++ b/cmd/compose/convert.go
@@ -240,7 +240,7 @@ func runConfigImages(opts convertOptions, services []string) error {
 		if s.Image != "" {
 			fmt.Println(s.Image)
 		} else {
-			fmt.Printf("%s_%s\n", project.Name, s.Name)
+			fmt.Printf("%s%s%s\n", project.Name, api.Separator, s.Name)
 		}
 	}
 	return nil


cmd/compose/convert.go
```diff
@@ -237,7 +237,7 @@ func runConfigImages(opts convertOptions, services []string) error {
 		if s.Image != "" {
 			fmt.Println(s.Image)
 		} else {
-			fmt.Printf("%s_%s\n", project.Name, s.Name)
+			fmt.Printf("%s-%s\n", project.Name, s.Name)
```
milas, line 240: This should use `api.Separator` which will handle the compatibility case as well:
https://github.com/docker/compose/blob/32ae036fd013566cc45ceef35f5f5b7cb29725bc/pkg/api/api.go#L472-L473

milas (changes requested):
Thanks for the PR - good catch!

Left a comment re: the separator, as there is a global for this that gets set based on whether Compose is running in compatibility mode or not.

cmd/compose/convert.go
```diff
@@ -240,7 +240,7 @@ func runConfigImages(opts convertOptions, services []string) error {
 		if s.Image != "" {
 			fmt.Println(s.Image)
 		} else {
-			fmt.Printf("%s_%s\n", project.Name, s.Name)
+			fmt.Printf("%s\n", project.Name+api.Separator+s.Name)
```
ndeloof, line 243: as we already rely on `Printf`, better not use `+` to append strings

ndeloof (approved):

glours (approved):
LGTM

PR #9909, opened 2022-10-12
Misleading error message for port command.

tiagombsilva:
The error message is using V1 separator hardcoded, it should be using the configured separator value

Signed-off-by: Tiago Silva <Tiago.MB.Silva@edu.azores.gov.pt>

**What I did**
changed hardcoded underscore to configured separator in the api


milas (approved):
Good catch 🙂

PR #9927, opened 2022-10-18
vendor: github.com/moby/buildkit v0.10.5

thaJeztah:
https://github.com/moby/buildkit/releases/tag/v0.10.5

no changes to modules (hence no `go.sum` changes?)

full diff: https://github.com/moby/buildkit/compare/v0.10.4...v0.10.5

### DIFF
--- a/e2e/go.mod
+++ b/e2e/go.mod
@@ -79,7 +79,7 @@ require (
 	github.com/mgutz/ansi v0.0.0-20170206155736-9520e82c474b // indirect
 	github.com/miekg/pkcs11 v1.1.1 // indirect
 	github.com/mitchellh/mapstructure v1.5.0 // indirect
-	github.com/moby/buildkit v0.10.4 // indirect
+	github.com/moby/buildkit v0.10.5 // indirect
 	github.com/moby/locker v1.0.1 // indirect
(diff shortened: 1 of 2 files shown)


glours:
You should also update the `go.mod` file in the `e2e` directory 😉

thaJeztah:
oh! will do

thaJeztah:
done!

glours (approved):
LGTM

thaJeztah:
Actually; this doesn't look right; I think I know what's wrong

thaJeztah:
Yes, there's a replace rule to use BuildKit master, so this PR doesn't take any effect; need to have an updated version from master for that, not the v0.10.5 release

glours:
Don't forget we also have a replace instruction for buildkit to use the same version as buildx

thaJeztah:
Ah, yes; perhaps we should add comments, like https://github.com/moby/buildkit/blob/c717d6aa7543d4b83395e0552ef2eb311f563aab/go.mod#L29

```
	github.com/docker/docker v20.10.18+incompatible // v22.06.x - see "replace" for the actual version
```

Because those replace rules are really easy to miss.

glours:
Yes, I'll prepare this for the next PR 👍

PR #9932, opened 2022-10-19
use CGO_ENABLED to avoid full system crash for mac users with a IPV6 config

### LINKED ISSUE #9930: [BUG] docker compose hangs network stack on macOS, reboot required
### Description

Network stack on macOS hangs when building a Docker image using `compose`.

This is also filed here: https://github.com/docker/for-mac/issues/6449

It was identified and eventually resolved at Planetscale: https://github.com/planetscale/discussion/discussions/181

---

**Solution**: build `compose` with CGO enabled:

```bash
CGO_ENABLED=1 go build
```

### Steps To Reproduce

Please see: https://github.com/docker/for-mac/issues/6449

### Compose Version

```Text
Tested on both:

- Docker Compose version v2.12.0
- Docker Compose version v2.10.2
```

### Docker Environment

```Text
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  compose: Docker Compose (Docker Inc., v2.10.2)
  extension: Manages Docker extensions (Docker Inc., v0.2.9)
  scan: Docker Scan (Docker Inc., v0.19.0)

Server:
 Containers: 0
  Running: 0
  Paused: 0
  Stopped: 0
 Images: 0
 Server Version: 20.10.17
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 2
 Plugins:
  Volume: local
 Swarm: inactive
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 9cd3357b7fd7218e4aec3eae239db1f68a5a6ec6
 init version: de40ad0
 Security Options:
  seccomp
   Profile: default
  cgroupns
 Kernel Version: 5.10.124-linuxkit
 OSType: linux
 Architecture: aarch64
 CPUs: 5
 Total Memory: 7.653GiB
 Name: docker-desktop
 ID: 4A73:SLXI:2AUC:F5YB:CUNJ:AOZS:ADZV:XGVN:C6SH:SMAH:7UJ4:FHP5
 Debug Mode: false
 Labels:
 Experimental: false
 Insecure Registries:
  hubproxy.docker.internal:5000
  127.0.0.0/8
 Live Restore Enabled: fals
```

### Anything else?

_No response_

glours:
see this issue https://github.com/golang/go/issues/52839

**What I did**
Enable CGO by default to produce cross platform binaries to avoid mac users with an IPV6 config to crash their OS

**Related issue**
fixes #9930 

image


milas (commented):
I'm not super familiar with [tonistiigi/xx](https://github.com/tonistiigi/xx) that we use for the cross-compilation.

I'm _GUESSING_ for Linux it'll do `-static` etc and since we're using musl, we'll still produce a fully static binary. That said, I think we'll still need `apk add --no-cache build-base` to be able to build with CGO.

But for macOS, I think we need an SDK:
> C/C++/CGo builds are supported for MacOS targets when an external SDK image is provided.

nicksieger (approved):
😬 Sucks that this is the only workaround for now. I wonder if we should also use `--ldflags='-extldflags=-static'` during compilation to still produce a static binary? Or does `xx` take care of that already?

nicksieger:
Was just trying this and it looks like we override `CGO_ENABLED` in the makefile build target. I think we're going to run into what Milas was talking about on mac as well.

glours:
I pinged @crazy-max to get the best practices to solve this issue 😉
Let's move this one in draft until we get his feedback

Dockerfile
```diff
@@ -40,7 +40,9 @@ RUN apk add --no-cache \
       protoc \
       protobuf-dev
 WORKDIR /src
-ENV CGO_ENABLED=0
+# Switching CGO to true as there is a weird issue with the Golang DNS resolver and IPV6
+#https://github.com/golang/go/issues/52839
+ENV CGO_ENABLED=1
```
crazy-max, line 45: Setting the env var here will not be taken into account as Makefile already set it: https://github.com/docker/compose/blob/32ae036fd013566cc45ceef35f5f5b7cb29725bc/Makefile#L51

glours:
The issue will be addressed directly by Golang in the next [1.20 release](https://go-review.googlesource.com/c/go/+/446178)

ndeloof:
closing as obsolete

PR #9936, opened 2022-10-21
go.mod: update docker-credential-helpers v0.7.0

thaJeztah:
### go.mod: update docker-credential-helpers v0.7.0

to match the version used by the cli (the cli doesn't use go.mod, so go modules
doesn't automatically pick that up);

https://github.com/docker/cli/blob/1d6c6e2367e292950e483e3c8ace142792bca718/vendor.mod#L14

Used code doesn't change, but we want to keep the older github.com/danieljoos/wincred v1.1.0
out of the dependency tree :)

Other changes:

### go mod: tidy and group "require" blocks, update comments

The file had multiple "requires" blocks, which made it harder to find which
dependencies were used. Some direct modules also were in the "indirect" block.

### go.mod: remove replace for runc

BuildKit and Buildx no longer require this replace rule (it probably only was
needed in buildkit, which used this version to compile).

**What I did**

**Related issue**

- https://github.com/docker/docker-credential-helpers/pull/208
- https://github.com/danieljoos/wincred/pull/22
- https://github.com/moby/buildkit/issues/2313#issuecomment-902353224

### DIFF
--- a/e2e/go.mod
+++ b/e2e/go.mod
@@ -8,7 +8,7 @@ replace github.com/cucumber/godog => github.com/laurazard/godog v0.0.0-202209220
 
 require (
 	github.com/cucumber/godog v0.12.5
-	github.com/docker/compose/v2 v2.11.1
+	github.com/docker/compose/v2 v2.11.1 // replaced; see replace rule for actual version
 	gotest.tools/v3 v3.4.0
 )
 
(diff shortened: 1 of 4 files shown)


thaJeztah:
let me rebase

glours (approved):
LGTM

PR #9934, opened 2022-10-21
update docker engine API to apply fix of CVE-2022-39253

glours:
Signed-off-by: Guillaume Lours <705411+glours@users.noreply.github.com>

**What I did**
Bump version of Docker engine dependency

**(not mandatory) A picture of a cute animal, if possible in relation with what you did**
image

PR #9940, opened 2022-10-21
go.mod: docker 5aac513617f072b15322b147052cbda0d451d389 / v22.06-dev

thaJeztah:
This restores compatibility with go1.18, which was broken since commit; https://github.com/moby/moby/commit/c062238ea4274e9af98cb96dcfe31a0dfc763440

cmd.Environ() is new in go1.19, and not needed for this specific case. Without this, trying to use this package in code that uses go1.18 will fail;

    builder/remotecontext/git/gitutils.go:216:23: cmd.Environ undefined (type *exec.Cmd has no field or method Environ)

Changing to use `os.Environ()` instead restores compatibility with go1.18

Full diff: https://github.com/docker/docker/compare/f9cb47a052be6a5f73e3077773ffa6ccbde1285d...5aac513617f072b15322b147052cbda0d451d389

**Related issue**

- https://github.com/docker/compose/pull/9934
- https://github.com/docker/docker-ce-packaging/pull/775

### DIFF
--- a/e2e/go.mod
+++ b/e2e/go.mod
@@ -155,7 +155,7 @@ require (
 
 replace (
 	github.com/docker/cli => github.com/docker/cli v20.10.3-0.20221013132413-1d6c6e2367e2+incompatible // 22.06 master branch
-	github.com/docker/docker => github.com/docker/docker v20.10.3-0.20221021122114-f9cb47a052be+incompatible // 22.06 branch
+	github.com/docker/docker => github.com/docker/docker v20.10.3-0.20221021173910-5aac513617f0+incompatible // 22.06 branch
 	github.com/moby/buildkit => github.com/moby/buildkit v0.10.1-0.20220816171719-55ba9d14360a // same as buildx
 
 	github.com/opencontainers/runc => github.com/opencontainers/runc v1.1.2 // Can be removed on next bump of containerd to > 1.6.4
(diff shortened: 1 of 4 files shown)


thaJeztah:
@milas @StefanScherer @glours PTAL 🤗

glours (approved):
LGTM

thaJeztah:
I'll bring this one in, so that it's the first commit after v2.12.1, keeping our options open 👍

milas:
FWIW we maintain compatibility with `compose-go` for both actively supported Go versions (and have CI to that effect), but have explicitly not actively supported `docker/compose` usage as a library thus far

(Agreed that having the change available is good though, thanks for making the updates here)

thaJeztah:
@milas yeah, so in this case, it was because compose depends on (not yet released) v22.06 code (through buildx). v22.06 is intended for go1.19 and above, but the current 20.10 release is still on go1.18. The release pipeline uses that version of go to build the binaries (which currently does not allow for differentiation go versions for each individual binary), causing the build to fail; https://github.com/docker/docker-ce-packaging/pull/775

_Perhaps_  we can add some hacks to install multiple go versions in the container build environment (and some `PATH` tricks to switch versions), but we didn't have it yet, so having this change in gives use options to either patch the build pipeline, _or_ do a 2.12.2 patch release from the commit from this PR.

milas:
Ah right, `docker-ce-packaging` 🙈

thaJeztah:
I know.. I know. Having all tied together.. complicate "things".

Issue #10004, opened 2022-11-16
[BUG] `docker compose up` mysteriously errors `strconv.Atoi: parsing "": invalid syntax`


nicks:
### Description

When I run `docker compose up --build`, i see this error:

```
[+] Running 0/0
 ⠋ Container viva-extensionsbackend-1  Creating                                                                                                                                                                           0.0s
 ⠋ Container viva-mockbackend-1        Creating                                                                                                                                                                           0.0s
 ⠋ Container cool_chatelet             Recreate                                                                                                                                                                           0.0s
strconv.Atoi: parsing "": invalid syntax
```

### Steps To Reproduce

I have no idea how i got into this state, but I know what the problem is!

I have a container with these labels:

```
            "Labels": {
                "com.docker.compose.project": "viva",
                "com.docker.compose.service": "electron",
                "com.docker.compose.version": "2.12.2"
            }
```

The container does NOT have the label `com.docker.compose.container-number`

The convergence engine blows up trying to figure out how to deal with this.

Here's the first place it blows up:
https://github.com/docker/compose/blob/32ae036fd013566cc45ceef35f5f5b7cb29725bc/pkg/compose/convergence.go#L376

If I change that line to ignore the error, here's the second place it blows up:
https://github.com/docker/compose/blob/32ae036fd013566cc45ceef35f5f5b7cb29725bc/pkg/compose/convergence.go#L430

The fix that worked for me locally was to change change getContainers here:
https://github.com/docker/compose/blob/32ae036fd013566cc45ceef35f5f5b7cb29725bc/pkg/compose/containers.go#L44

to filter out containers that are missing this label.

### Compose Version

```Text
available on request
```

### Docker Environment

```Text
available on request
```

nicks:
Related issues:
https://github.com/docker/compose-cli/issues/1537
https://github.com/docker/compose/issues/9991

PR #10005, opened 2022-11-17
ignore error parsing container number label, just warn

ndeloof:
**What I did**
don't throw an error as we can't retrieve or parse container number label, just warn
While it's unclear how we end in such a bad state, this should not break compose usage as this has limited impact - we only use this to suffix container name with a unique number, nothing critical

**Related issue**
fixes https://github.com/docker/compose/issues/10004

image

### DIFF
--- a/pkg/compose/convergence.go
+++ b/pkg/compose/convergence.go
@@ -226,10 +226,7 @@ func (c *convergence) ensureService(ctx context.Context, project *types.Project,
 		updated[i] = container
 	}
 
-	next, err := nextContainerNumber(containers)
-	if err != nil {
-		return err
-	}
+	next := nextContainerNumber(containers)
 	for i := 0; i < expected-actual; i++ {
 		// Scale UP
 		number := next + i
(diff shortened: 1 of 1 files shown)


glours (approved):
LGTM

pkg/compose/convergence.go
```diff
@@ -370,18 +367,23 @@ func shouldWaitForDependency(serviceName string, dependencyConfig types.ServiceD
...
 }
 
-func nextContainerNumber(containers []moby.Container) (int, error) {
+func nextContainerNumber(containers []moby.Container) int {
 	max := 0
 	for _, c := range containers {
-		n, err := strconv.Atoi(c.Labels[api.ContainerNumberLabel])
+		s, ok := c.Labels[api.ContainerNumberLabel]
+		if !ok {
+			logrus.Warnf("container %q is missing %s label", c.ID, api.ContainerNumberLabel)
```
ulyssessouza, line 375: nit:
```suggestion
			logrus.Warnf("container %q is missing %q label", c.ID, api.ContainerNumberLabel)
```

ulyssessouza (approved):
LGTM!
Just nits on %s -> %q

nickform-ost:
I'm on macOS. The dashboard app shows version number v4.14.1 and I still hit the bug. Any idea how to estimate when this will make it all the way to users? Also, are you in a position to suggest any workaround that could help us out in the meantime?

glours:
@nickform-ost next release of Docker Desktop should be release soon with the `2.13.0` version of Compose
If you can, a `docker system prune -af` should resolve the issue

nickform-ost:
I really appreciate you getting back to me. That seems to have done the trick!

PR #10017, opened 2022-11-23
useDockerDefaultOrServicePlatform fct should return service.platform if defined

glours:
and present in the build.platforms list (or if the list is empty)

**What I did**
Move the check of the `service.platform` presence inside the `service.build.platforms` list with the test of the list size, to be sure we always return the `service.plaform` value when present in the build platforms list or when this list is empty

image


vvoland (approved):
LGTM

ndeloof (approved):

Issue #10013, opened 2022-11-21
[BUG] --ansi never doesn't do anything


elandorr:
### Description

We updated to the new compose. While it does look really fancy and matrix-y, frankly, it's useless. So I checked the flags by running `docker compose`. (Which are mismatched with the docs FYI!)

`--ansi never` sounds like it would get us back the useful full view. But it does not. It simply does absolutely nothing. The output is unchanged.

My workaround is writing it to a file, that disables it. It still looks different than before if I remember correctly, but at least the full log is readable.

PS: Is there a mailing list for only critical updates and changes like this? I merely randomly discovered that compose is now in your repo and v1 deprecated.

Thanks.

### Steps To Reproduce

`docker compose --ansi never build`

### Compose Version

```Text
Docker Compose version v2.12.2
```

### Docker Environment

```Text
Only the important bits, as I'm not keen on sending ids and potentially identifiable data randomly.

app: Docker App (Docker Inc., v0.9.1-beta3)
buildx: Docker Buildx (Docker Inc., v0.9.1-docker)
compose: Docker Compose (Docker Inc., v2.12.2)
scan: Docker Scan (Docker Inc., v0.21.0)
Server Version: 20.10.21
containerd version: 770bd0108c32f3fb5c73ae1264f7e503fe7b2661
```

ndeloof:
`build` output actually comes from `buildx` used under the hood 
You can control this output style using `compose build --progress=plain`

elandorr:
Thank you very much @ndeloof! That works.

Any clue about a potential mailing list? https://www.docker.com/newsletter-subscription/ is a bit much.

ndeloof:
we indeed don't have a mailing list, but you can join docker community slack #compose channel

elandorr:
Thanks. The Docker slack requires an @docker.com or an invite. But I'm not keen on another chat client eating resources in bg anyhow.

I'd like to suggest a plain list for critical issues, majors, and deprecations. Has been helpful in many projects for decades now (mail is universal, and needs minimal res).

robvdl:
When you use "docker compose help build" it says to use --ansi never, but then when you try it it says "--ansi doesn't exist". So it's own help is misleading.

arcanisgk:
same problem:

![image](https://github.com/docker/compose/assets/1207306/f9525b49-328b-4f77-a444-9f5783073b4a)

luciangabor:
[...]

PR #10020, opened 2022-11-24
configure buildx for plain output if --ansi=never has been set

ndeloof:
**What I did**
set builx progress=plain when compose is ran with ansi=never

**Related issue**
close https://github.com/docker/compose/issues/10013

image


glours (approved):
👍

PR #10026, opened 2022-11-29
move image digests resolution to backend

ndeloof:
**What I did**
introduced `ResolveImageDigests` in the API and moved image digest resolution to the backend
longer terms goal is to use `dockercli.RegistryClient()` to replace `cnab-to-oci` dependency

image


cmd/compose/convert.go
```diff
@@ -141,8 +141,9 @@ func runConvert(ctx context.Context, backend api.Service, opts convertOptions, s
 	}
```
milas, line 141: Delete lines 130-141 now?

milas (approved):
LGTM - small comment re: duplicated code

cmd/compose/convert.go
ndeloof: oh yes indeed, my bad

ndeloof:
@thaJeztah thanks, this is exactly what I was looking for! 🥰

thaJeztah:
Make sure to double check in the cli as well as at some point resolution was moved; check for "no-resolve-image" (I think) on the docker cli (I'm just away from my keyboard right now 😬😅)

milas (approved):
✨

pkg/compose/compose.go
```diff
@@ -93,13 +94,34 @@ func getContainerNameWithoutProject(c moby.Container) string {
...
+	}
+
 	switch options.Format {
 	case "json":
 		return json.MarshalIndent(project, "", "  ")
 	case "yaml":
 		return yaml.Marshal(project)
 	default:
-		return nil, fmt.Errorf("unsupported format %q", options)
+		return nil, fmt.Errorf("unsupported format %q", options.Format)
```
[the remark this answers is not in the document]
ndeloof: linter suddenly noticed this bug, while it existed before ... weird

Issue #10021, opened 2022-11-24
[BUG] Docker compose >=2.10 stuck on stopping containers started with --attach-dependencies option


sidz:
### Description

This issue was fixed in 2.6.0 and is not reproducible till 2.9.0 (includes 2.9.0) But since 2.10 was broken again.
Here my old issue: https://github.com/docker/compose/issues/9494

I assume (but not sure) that this https://github.com/docker/compose/pull/9488 PR fixes old issue. But as I can see fix is still in v2 branch.

### Steps To Reproduce

1. Define 3 services in the `docker-compose.yaml` with depends_on key. (example below)
2. Up only php container with `--attach-dependencies` eg. `docker compose up php --attach-dependencies`
3. Click on Ctrl + C or run docker compose kill in the separate tab
4. Docker stops only php container and freezes on this step and don't stop others.

```
version: "3.4"

services:
    php:
        image: php:8.1.12-fpm-alpine3.16
        depends_on:
            - mailhog

    memcache:
        image: memcached:1.6-alpine
        restart: on-failure

    mailhog:
        image: mailhog/mailhog
        ports:
            - "127.0.0.1:8025:8025"
```

### Compose Version

```Text
Docker Compose version v2.13.0
```

### Docker Environment

```Text
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  compose: Docker Compose (Docker Inc., v2.13.0)

Server:
 Containers: 11
  Running: 0
  Paused: 0
  Stopped: 11
 Images: 50
 Server Version: 20.10.12
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 1
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 io.containerd.runtime.v1.linux runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 
 runc version: 
 init version: 
 Security Options:
  apparmor
  seccomp
   Profile: default
 Kernel Version: 5.15.0-53-generic
 Operating System: Ubuntu 20.04.5 LTS
 OSType: linux
 Architecture: x86_64
 CPUs: 8
 Total Memory: 23.41GiB
[...]

PR #10029, opened 2022-11-29
fix deadlock waiting for attached-dependencies

ndeloof:
**What I did**
Fixed 2 race conditions as we can configure `up` to attach additional containers :
- we need to explicitly tell the printer to stop, as the additional services might still run
- we missed a way to stop watching containers for restart. This replaces use of a background context with a cancelable one

**Related issue**
closes https://github.com/docker/compose/issues/10021

image

### DIFF
--- a/pkg/compose/up.go
+++ b/pkg/compose/up.go
@@ -23,11 +23,10 @@ import (
 	"os/signal"
 	"syscall"
 
-	"github.com/docker/compose/v2/pkg/api"
-	"github.com/docker/compose/v2/pkg/progress"
-
 	"github.com/compose-spec/compose-go/types"
 	"github.com/docker/cli/cli"
+	"github.com/docker/compose/v2/pkg/api"
+	"github.com/docker/compose/v2/pkg/progress"
 	"golang.org/x/sync/errgroup"
 )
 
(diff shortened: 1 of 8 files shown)


pkg/compose/up.go
```diff
@@ -76,18 +77,21 @@ func (s *composeService) Up(ctx context.Context, project *types.Project, options
...
+		stopC <- struct{}{}
 		stopFunc() //nolint:errcheck
+		printer.Stop()
 	}()
 
 	var exitCode int
 	eg, ctx := errgroup.WithContext(ctx)
 	eg.Go(func() error {
 		code, err := printer.Run(context.Background(), options.Start.CascadeStop, options.Start.ExitCodeFrom, stopFunc)
+		fmt.Println("printer is done")
```
glours, line 89: Is this log needed?

nicksieger (approved):
Nice!

milas (approved):
LGTM other than the errant log that Guillaume already caught

pkg/compose/up.go
ndeloof: nope, reminds from diagnostic phase trying to understand the deadlock cause :)

milas:
I canceled auto-merge - it looks like the E2E tests are legitimately hanging now and timing out after 15m

ndeloof:
still buggy, switching back to draft to prevent merge

PR #10030, opened 2022-11-30
introduce --parallel to limit concurrent engine calls

ndeloof:
**What I did**
Introduce `--parallel=<int>` for user to limit the concurrency during image pulls
could be later applied to other operations, typically `build` with adequate buildx|kit support for this

**Related issue**

image

### DIFF
--- a/pkg/compose/compose.go
+++ b/pkg/compose/compose.go
@@ -41,12 +41,14 @@ import (
 // NewComposeService create a local implementation of the compose.Service API
 func NewComposeService(dockerCli command.Cli) api.Service {
 	return &composeService{
-		dockerCli: dockerCli,
+		dockerCli:      dockerCli,
+		maxConcurrency: -1,
 	}
 }
 
 type composeService struct {
-	dockerCli command.Cli
+	dockerCli      command.Cli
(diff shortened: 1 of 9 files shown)


pkg/compose/compose.go
```diff
@@ -45,7 +45,8 @@ func NewComposeService(dockerCli command.Cli) api.Service {
 }
 
 type composeService struct {
-	dockerCli command.Cli
+	dockerCli      command.Cli
+	maxConcurrency int
```
glours, line 49: shouldn't it be initialize with a `-1` value by default as the default value for an int will be `0`

glours (changes requested):
I think tests are failing due to the wrong default value of the `maxConcurrency` attribute.
Docs should also be updated 😉

laurazard:
Oh nice! This is something that people have asked for for _a while_, albeit mostly regarding build operations.

ndeloof:
Build max-parallelism can be controlled, but this takes place on the builder side: https://github.com/docker/buildx/blob/master/docs/guides/resource-limiting.md#max-parallelism

glours (approved):
LGTM

Issue #9873, opened 2022-09-23
Volumes of type `tmpfs` lacks support for the `mode` parameter


rawtaz:
Docker Compose currently does not support the `mode` parameter for `tmpfs` mounts. It only support the `size` parameter.

This has been verified both empirically with `Docker Compose version v2.6.0` with `Docker Engine - Community version 20.10.17` and by looking in the source code.

Here are some references showing that `mode` is currently not supported:
- The `buildTmpfsOptions` function is [missing it and even has a comment for it](https://github.com/docker/compose/blob/2977f4c89740809f20096673ab4e3ced543a905a/pkg/compose/create.go#L1026).
- The `ServiceVolumeTmpfs` struct is [missing it](https://github.com/compose-spec/compose-go/blob/5135b47ac4bb0c9b5c5bcb274e767b410ecef42e/types/types.go#L830).

Worth noting is that Docker does support it in the [`mount.TmpfsOptions` struct](https://github.com/moby/moby/blob/master/api/types/mount/mount.go#L115) that is [returned by the `buildTmpfsOptions` function](https://github.com/docker/compose/blob/2977f4c89740809f20096673ab4e3ced543a905a/pkg/compose/create.go#L1020) in Compose.

Also worth mentioning is that the official Docker Compose specification documentation for tmpfs volumes [clearly states that `mode` should be supported for `tmpfs`](https://docs.docker.com/compose/compose-file/#long-syntax-4). Despite this, however, [the `compose-go` library does not support the `mode` parameter](https://github.com/compose-spec/compose-go/issues/302).

Here's an example of a `tmpfs` specification for a service that does not work (Compose throws an error `services.foo.volumes.0.tmpfs Additional property mode is not allowed`):
~~~
volumes:
  - type: tmpfs
    target: /test
    tmpfs:
      size: 128k
      mode: 0777
~~~

I'm not sure if I got all of the relationships between specifications, libraries and other software right, I figured that given the above mentioned pieces of `mode` throughout the source code, it made sense to open an issue about it in this repository.

It looks like it's rather straight forward to add support for this option, since Docker itself supports it, or am I missing something?

Either way, it's a problem because it's currently impossible to mount and use `tmpfs` volumes with unprivileged processes/containers.

cavo789:
Same issue

I just met the same issue. I'm building a postgres image where I'm using a non root user and tmpfs storage for my PGDATA directory.

[...]

PR #10031, opened 2022-11-30
schema: add support for tmpfs.mode in mount definition

milas:
## ~⚠️ Draft until compose-spec/compose-go#325 is merged & tagged~

**What I did**
Add support for `tmpfs.mode` to set permissions on `tmpfs` mounts

**Related issue**
* #9873
* compose-spec/compose-go#325

prairie chicken


thaJeztah:
we may have to look if mode is fully functional as well; I recall this ticket, and I don't think we dug to the end of that one yet (at least I haven't found time to dig into it myself);

- https://github.com/moby/moby/issues/40881

nicksieger (approved):

milas:
> we may have to look if mode is fully functional as well

Thanks for sharing that! I think it's still fine to add here unless we're planning on deprecating `tmpfs` support entirely - once/if the upstream bug is fixed, nested permissions should start just working in Compose after upgrading engine. For now, this will at least give parity in that permissions on unnested `tmpfs` mounts are now supported

thaJeztah:
> Thanks for sharing that! I think it's still fine to add here unless we're planning on deprecating tmpfs support entirely - once/if the upstream bug is fixed, nested permissions should start just working in Compose after upgrading engine. For now, this will at least give parity in that permissions on unnested tmpfs mounts are now supported

Yup, change looks okay to me; mostly recalled that there was an issue around this option, and thought I'd mention it to set the expectations right 😄. Trying / testing /digging on that linked issue is always welcome as well (I created that ticket to have it at least written down somewhere as a reminder)


@milas looks like this needs a rebase now (conflict in `go.sum`)

milas:
Yeah, it's in draft as it needs the updated schema code from `compose-go` so will be fixing that when I remove my `replace`s

PR #10036, opened 2022-12-01
Fix replacing "service:x" with "container:y"

### LINKED ISSUE #9055: Unable to run service with network_mode "service:NAME"
**Description**

This ticket is a **duplicate** of https://github.com/docker/compose/issues/8766 given it was marked as "closed" but is not resolved.

Unable run to execute `compose run <service>` when a service is using `network_mode: service:nginx`

This is a docker-compose.yml file which can replicate the issue.

```
version: '3'

services:
  nginx:
    image: nginx

  cli:
    image: alpine
    network_mode: service:nginx
```

Commands which replicate the issue.

```
$ compose up -d
$ compose run cli sh
Error response from daemon: network service:nginx not found
```

```
Error response from daemon: network service:nginx not found
```

The service container to execute within the same network as the "nginx" service in this case.

This worked with Compose v1.

```
Docker Compose version v2.2.2
```

**Output of `docker info`:**

```
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc., v0.7.1-docker)
  compose: Docker Compose (Docker Inc., 2.0.0-beta.1)
  scan: Docker Scan (Docker Inc., v0.12.0)

Server:
 Containers: 129
  Running: 1
  Paused: 0
  Stopped: 128
 Images: 169
 Server Version: 20.10.12
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
 Swarm: inactive
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 7b11cfaabd73bb80907dd23182b9347b4245eb5d
 runc version: v1.0.2-0-g52b36a2
 init version: de40ad0
 Security Options:
  apparmor
  seccomp
   Profile: default
  cgroupns
 Kernel Version: 5.15.5-76051505-generic
 Operating System: Pop!_OS 21.10
 OSType: linux
 Architecture: x86_64
 CPUs: 12
 Total Memory: 31.32GiB
 Name: batman
 ID: QAXJ:HDRA:DGM3:DR75:OBYA:VOV7:EEOE:K5IG:QD57:54I7:SVH6:5EQ3
 Debug Mode: false
 Username: nickschuch
 Labels:
 Experimental: false
 Insecure Registries:
  127.0.0.0/8
 Live Restore Enabled: false
```

N/A

i-ky:
**What I did**
As mentioned in https://github.com/docker/compose/issues/9055#issuecomment-1333848036 and https://github.com/docker/compose/issues/9055#issuecomment-1333862905,  `updateServices()` gets called from `prepareRun()` with an array of _all_ containers belonging to the project and taking only first one of them is wrong in general case. For this use case we should construct a `map` from service name to ID of the first container belonging to that service. This `map` is then used ro replace `service:x` with `container:y` in `network_mode:`, `ipc:` and `pid:`. This change should not break the second use case of `updateServices()` when it is called from `updateProject()` with an array of containers already filtered by service (as pointed out in https://github.com/docker/compose/issues/9055#issuecomment-1333861996). In this case `map` will contain only one pair and code will behave exactly like before - only replacing mentions of one particular service and leaving all other `service:x` intact.

**Related issue**
Fixes #9055.

### DIFF
--- a/pkg/compose/convergence.go
+++ b/pkg/compose/convergence.go
@@ -126,22 +126,13 @@ func updateServices(service *types.ServiceConfig, cnts Containers) {
 	if len(cnts) == 0 {
 		return
 	}
-	serviceName2containerID := make(map[string]string)
-	for _, cnt := range cnts {
-		serviceName := cnt.Labels[api.ServiceLabel]
-		if _, exists := serviceName2containerID[serviceName]; !exists {
-			serviceName2containerID[serviceName] = cnt.ID
-		}
-	}
 
-	if id, found := serviceName2containerID[getDependentServiceFromMode(service.NetworkMode)]; found {
-		service.NetworkMode = types.NetworkModeContainerPrefix + id
-	}
-	if id, found := serviceName2containerID[getDependentServiceFromMode(service.Ipc)]; found {
-		service.Ipc = types.NetworkModeContainerPrefix + id
-	}
-	if id, found := serviceName2containerID[getDependentServiceFromMode(service.Pid)]; found {
-		service.Pid = types.NetworkModeContainerPrefix + id
+	for _, str := range []*string{&service.NetworkMode, &service.Ipc, &service.Pid} {
+		if d := getDependentServiceFromMode(*str); d != "" {
+			if serviceContainers := cnts.filter(isService(d)); len(serviceContainers) > 0 {
+				*str = types.NetworkModeContainerPrefix + serviceContainers[0].ID
+			}
+		}
 	}
 	var links []string
 	for _, serviceLink := range service.Links {
--- a/pkg/compose/convergence.go
+++ b/pkg/compose/convergence.go
@@ -126,17 +126,22 @@ func updateServices(service *types.ServiceConfig, cnts Containers) {
 	if len(cnts) == 0 {
 		return
 	}
-	cnt := cnts[0]
-	serviceName := cnt.Labels[api.ServiceLabel]
+	serviceName2containerID := make(map[string]string)
+	for _, cnt := range cnts {
+		serviceName := cnt.Labels[api.ServiceLabel]
+		if _, exists := serviceName2containerID[serviceName]; !exists {
+			serviceName2containerID[serviceName] = cnt.ID
+		}
+	}
 
-	if d := getDependentServiceFromMode(service.NetworkMode); d == serviceName {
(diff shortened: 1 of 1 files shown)


pkg/compose/convergence.go
```diff
@@ -126,17 +126,22 @@ func updateServices(service *types.ServiceConfig, cnts Containers) {
...
+	serviceName2containerID := make(map[string]string)
+	for _, cnt := range cnts {
+		serviceName := cnt.Labels[api.ServiceLabel]
+		if _, exists := serviceName2containerID[serviceName]; !exists {
+			serviceName2containerID[serviceName] = cnt.ID
+		}
+	}
 
-	if d := getDependentServiceFromMode(service.NetworkMode); d == serviceName {
-		service.NetworkMode = types.NetworkModeContainerPrefix + cnt.ID
```
ndeloof, line 133: you could use `cnts.filter(isService(d))` to retrieve project containers matching the requested service

ndeloof:
With this implemented in `updateService` you could remove https://github.com/docker/compose/blob/7cf5940f4a384a394f18b500b7817271c0f4351d/pkg/compose/convergence.go#L118

pkg/compose/convergence.go
i-ky: Thank you for the advice!

i-ky:
> With this implemented in `updateService` you could remove
> 
> https://github.com/docker/compose/blob/7cf5940f4a384a394f18b500b7817271c0f4351d/pkg/compose/convergence.go#L118

Do you mean removing filtering by `serviceName` there and passing all containers to `updateServices()`? Not sure about that as this would change behaviour. Currently when called from `updateProject()` `updateServices()` replaces only `service:x` (assuming `serviceName` equals `"x"`). If we pass all containers, it will replace not just `service:x`, but also `service:y` and `service:z`. Is this desirable? I guess there is a reason why `updateProject()` gets called with one `serviceName` at a time in very particular order:
https://github.com/docker/compose/blob/7cf5940f4a384a394f18b500b7817271c0f4351d/pkg/compose/convergence.go#L90

ndeloof (approved):

PR #10039, opened 2022-12-02
port: improve error-handling if port not found

### LINKED ISSUE #10010: [BUG] docker compose port app 80 returns `:0`
### Description

This week I noticed that the command `docker compose port app 80` sometimes returns `:0` instead of the actual port. See the screenshot here:

image

### Steps To Reproduce

_No response_

### Compose Version

```Text
Docker Compose version v2.12.2
```

### Docker Environment

```Text
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc., v0.9.1-docker)
  compose: Docker Compose (Docker Inc., v2.12.2)

Server:
 Containers: 4459
  Running: 19
  Paused: 0
  Stopped: 4440
 Images: 984
 Server Version: 20.10.21
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
 Swarm: inactive
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 770bd0108c32f3fb5c73ae1264f7e503fe7b2661
 init version: de40ad0
 Security Options:
  apparmor
  seccomp
   Profile: default
  cgroupns
 Kernel Version: 5.10.0-19-amd64
 Operating System: Debian GNU/Linux 11 (bullseye)
 OSType: linux
 Architecture: x86_64
 CPUs: 24
 Total Memory: 125.6GiB
 Name: shop
 ID: TZUB:ZIAT:RVG5:P27F:2HBN:2PSL:FHAO:KW7X:JLIY:CUGY:X3DL:PN2X
 Debug Mode: false
 Labels:
 Experimental: false
 Insecure Registries:
  127.0.0.0/8
 Live Restore Enabled: false
```

### Anything else?

_No response_

milas:
**What I did**
This method looked slightly incomplete. If the port wasn't found, it'd return `err`, but that was always `nil`, so we'd print out `:0`.

Now, we construct a nice error message with the targeted port and the ones we found.

The `--protocol` flag is also now case-insensitive to prevent any weirdness/confusion there.

**Related issue**
Fixes #10010.

a black cat with very large pupils

### DIFF
--- a/pkg/compose/port.go
+++ b/pkg/compose/port.go
@@ -27,7 +27,7 @@ import (
 	"github.com/docker/docker/api/types/filters"
 )
 
-func (s *composeService) Port(ctx context.Context, projectName string, service string, port int, options api.PortOptions) (string, int, error) {
+func (s *composeService) Port(ctx context.Context, projectName string, service string, port uint16, options api.PortOptions) (string, int, error) {
 	projectName = strings.ToLower(projectName)
 	list, err := s.apiClient().ContainerList(ctx, moby.ContainerListOptions{
 		Filters: filters.NewArgs(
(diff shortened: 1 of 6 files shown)


pkg/compose/port.go
```diff
@@ -44,9 +44,23 @@ func (s *composeService) Port(ctx context.Context, projectName string, service s
...
 			return p.IP, int(p.PublicPort), nil
 		}
 	}
-	return "", 0, err
+	return "", 0, portNotFoundError(options.Protocol, port, container)
+}
+
+func portNotFoundError(protocol string, port uint16, ctr moby.Container) error {
+	formatPort := func(protocol string, port uint16) string {
+		return fmt.Sprintf("%s/%d", protocol, port)
```
nicksieger, line 56: Don't these usually get represented as port/protocol (e.g., `80/tcp` or `53/udp`)?
```suggestion
		return fmt.Sprintf("%d/%s", port, protocol)
```

nicksieger (approved):

pkg/compose/port.go
thaJeztah: Yes, correct; also see https://github.com/docker/cli/pull/3892

ndeloof (approved):

pkg/compose/port.go
milas: You know what's sad is that I looked at the convention and tried to copy it and never noticed I had it reversed despite that 🤦

PR #10044, opened 2022-12-06
Add --include-deps to push command

gferon:
**What I did**

Add `--include-deps` to the `push` sub-command, similar to what `pull` can do. I believe the default behavior is a little bit weird since you can't explicitly push the built image of a service that has declared dependencies. That being said, this change will break backwards compatibility, so I feel like we could

_Cute animal tax_ - a photo of my containerized cat:

PXL_20201120_110811709 PORTRAIT(2)

### DIFF
--- a/cmd/compose/pull.go
+++ b/cmd/compose/pull.go
@@ -68,7 +68,7 @@ func pullCommand(p *projectOptions, backend api.Service) *cobra.Command {
 	return cmd
 }
 
-func FilterServices(project *types.Project, services []string) error {
+func withSelectedServicesOnly(project *types.Project, services []string) error {
 	enabled, err := project.GetServices(services...)
 	if err != nil {
 		return err
@@ -90,7 +90,10 @@ func runPull(ctx context.Context, backend api.Service, opts pullOptions, service
 	}
 
 	if !opts.includeDeps {
-		FilterServices(project, services)
+		err := withSelectedServicesOnly(project, services)
+		if err != nil {
+			return err
+		}
 	}
 
 	return backend.Pull(ctx, project, api.PullOptions{
--- a/cmd/compose/pull.go
+++ b/cmd/compose/pull.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"os"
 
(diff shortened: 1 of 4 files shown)


ndeloof:
While I don't expect many users to rely on this, I'm a bit concerned about another backward compatibility break.
Maybe better introduce `--no-deps` to mimic same option used by `up`?
Also, to avoid code duplication, could define a func to disable all services but selected ones

gferon:
> While I don't expect many users to rely on this, I'm a bit concerned about another backward compatibility break.

Fair point, I can definitely move it to a `--no-deps` flag instead, but this would be inconsistent with `docker-compose pull`. What about setting the value of this flag to `true` instead and let users pass `--include-deps=false` instead?

> Also, to avoid code duplication, could define a func

Sure, can you point me to a good place to do this? I'm not familiar with this codebase to select a proper location myself.

ndeloof:
you can just extract https://github.com/docker/compose/blob/v2/cmd/compose/up.go#L57-L66 into a func inside up.go and call it from both command.

I didn't noticed `--include-deps` flag existed for pull, makes more sense now. Both should be aligned. Minor compatibility break probably is acceptable from this point of view.

ndeloof:
also need to regenerate docs to reflect new flag using `make docs`

gferon:
@ndeloof not sure why you meant `up.go`, I extracted the function in `pull.go` - if you really wanted it to be in `up.go` I can change it.

Otherwise, this should be ready.

cmd/compose/pull.go
```diff
@@ -67,23 +68,29 @@ func pullCommand(p *projectOptions, backend api.Service) *cobra.Command {
 	return cmd
 }
 
+func FilterServices(project *types.Project, services []string) error {
```
ndeloof, line 71: Don't need to be exporte. Also name is bit confusing. Maybe `withSelectedServicesOnly` or something comparable?

gferon: Done!

ndeloof (approved):

PR #10047, opened 2022-12-06
Cleanup tips from output

thaJeztah:
The scan tip has been shown for two years, and most users will know about it by now. Presenting the message also involved checking if the plugin was installed, and wether or not the message was shown before, which also caused some overhead, so cleaning up the output a bit.

The corresponding DOCKER_SCAN_SUGGEST environment-variable is also removed with this.


glours (approved):
LGTM

milas (approved):

ndeloof (commented):
👍

Issue #9897, opened 2022-10-06
docker-compose up re-creates container when it shouldn't


g-lashuk:
**Description**

Hello, I have the following test.yml:

```
version: "3.2"
services:
  rabbitmq:
    image: rabbitmq:3-management-alpine
    container_name: 'rabbitmq'
    ports:
        - 5672:5672
        - 15672:15672
```
Running   `docker-compose -p test-container -f test.yaml up -d` leads to image pulling and creating container

![image](https://user-images.githubusercontent.com/84844620/194354348-cbe3cfac-1a96-4de4-aa7c-b4f08a195c9a.png)

However, the second run of the same `docker-compose -p test-container -f test.yaml up -d` command without any changes in test.yaml file leads to container re-creation:
![image](https://user-images.githubusercontent.com/84844620/194355682-0a672631-9ada-4e4e-9876-f192b20c950b.png)

It, seems to be unexpected because all subsequent runs of the same command do not recreate the container:
![image](https://user-images.githubusercontent.com/84844620/194356188-3e41e354-6ab6-4f67-85f5-1db1ebbeb3a3.png)

**Steps to reproduce the issue:**
1. Create docker-compose.yaml file with any image that is not presented locally and should be pulled first (rabbit-mq was used as example)
2. run docker-compose up for the first time (it pulls and creates the container)
3. run docker-compose up for the second time

**Describe the results you received:**
The second run recreates container, which was created

**Describe the results you expected:**
As far as I undestand the container should not be recreated because the configuration has not been changed plus all subsequent runs of the same command do not recreate the container 

**Additional information you deem important (e.g. issue happens only occasionally):**

 - The issue is reproduced exactly at the second run of `docker-compose up `command after pulling image that is not presented locally.
- The benavior does not depend on the image used in test.yaml file (as far as I noticed the only precondition is that the first run of docker-compose up pulls an image and as a result, only the next run recreates container )

**Output of `docker compose version`:**

```
Docker Compose version v2.10.2
```

**Output of `docker info`:**

```
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc., v0.9.1)
  compose: Docker Compose (Docker Inc., v2.10.2)
  extension: Manages Docker extensions (Docker Inc., v0.2.9)
  sbom: View the packaged-based Software Bill Of Materials (SBOM) for an image (Anchore Inc., 0.6.0)
  scan: Docker Scan (Docker Inc., v0.19.0)

[...]

PR #10051, opened 2022-12-07
fix race condition collecting pulled images IDs

ndeloof:
**What I did**
fix race condition: we only can set images ID _after_ all pulls completed and we collected the digests.

**Related issue**
fixes https://github.com/docker/compose/issues/9897

image

### DIFF
--- a/pkg/compose/pull.go
+++ b/pkg/compose/pull.go
@@ -294,14 +294,11 @@ func (s *composeService) pullRequiredImages(ctx context.Context, project *types.
 				return err
 			})
 		}
+		err := eg.Wait()
 		for i, service := range needPull {
 			if pulledImages[i] != "" {
 				images[service.Image] = pulledImages[i]
 			}
-		}
-		err := eg.Wait()
-		if err != nil {
-			return err
 		}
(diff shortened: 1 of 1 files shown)


pkg/compose/pull.go
```diff
@@ -286,23 +286,19 @@ func (s *composeService) pullRequiredImages(ctx context.Context, project *types.
 			i, service := i, service
 			eg.Go(func() error {
 				id, err := s.pullServiceImage(ctx, service, info, s.configFile(), w, quietPull, project.Environment["DOCKER_DEFAULT_PLATFORM"])
-				pulledImages[i] = id
```
milas, line 289: I think this should still be there - otherwise, `pulledImages` will always be full of empty strings, so we won't end up updating `images`

milas (changes requested):

pkg/compose/pull.go
ndeloof: Indeed, unexpected deletion
Seems my IDE is getting some fun with my nerves those days

milas (approved):

Issue #8724, opened 2021-10-01
`docker compose pull` fails if any image is not pullable


thespad:
**Description**

`docker compose pull` fails if any image is not pullable (such as being locally built).

**Steps to reproduce the issue:**
1. Have a docker-compose.yml with multiple services, at least one of which uses a locally built image and one of which is from an accessible registry.
2. Run `docker compose pull`

**Describe the results you received:**
When the locally built image fails to pull, all other pulls are aborted (`dcp` is an alias for `docker compose pull`)

![image](https://user-images.githubusercontent.com/8425502/135612235-fc220881-f7cc-457a-a0f6-e7265595eb36.png)

**Describe the results you expected:**
Pulls of other images should complete even if some are unavailable, as was the behaviour with docker-compose (v1.x)

**Additional information you deem important (e.g. issue happens only occasionally):**
Passing `--ignore-pull-failures` mitigates the issue - though still reports the failures - but this is still a break in behaviour from v1.x

![image](https://user-images.githubusercontent.com/8425502/135612388-320d5eb9-190f-4405-8b7a-6569a4dc08e9.png)

**Output of `docker compose version`:**

```
Docker Compose version v2.0.1
```

**Output of `docker info`:**

```
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  app: Docker App (Docker Inc., v0.9.1-beta3)
  buildx: Build with BuildKit (Docker Inc., v0.6.1-docker)
  compose: Docker Compose (Docker Inc., v2.0.1)
  scan: Docker Scan (Docker Inc., v0.8.0)

Server:
 Containers: 54
  Running: 54
  Paused: 0
  Stopped: 0
 Images: 65
 Server Version: 20.10.8
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 1
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 io.containerd.runtime.v1.linux runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: e25210fe30a0a703442421b0f60afac609f950a3
 runc version: v1.0.1-0-g4144b63
 init version: de40ad0
[...]

PR #10054, opened 2022-12-08
Don't stop pull for images that can be built

ndeloof:
**What I did**
Don't immediately stop pulling on failure for image that **can** be built: Wait for other pullable images to complete. 
Only stop with immediate error if some image can't be pulled and has no build config.

```
➜  docker compose pull
[+] Running 2/4
 ⠿ tata Pulled                                                                                                             7.8s
   ⠿ e96e057aae67 Pull complete                                                                                            5.1s
 ⠿ titi Warning                                                                                                            2.1s
 ⠿ toto Warning                                                                                                            2.1s
WARNING: Some service image(s) must be built from source by running:
    docker compose build toto titi
2 errors occurred:
	* Error response from daemon: pull access denied for tot, repository does not exist or may require 'docker login': denied: requested access to the resource is denied
	* Error response from daemon: pull access denied for tit, repository does not exist or may require 'docker login': denied: requested access to the resource is denied
```

**Note:** this still eventually fails with a !=0 exit code. Maybe we should introduce a new flag to ignore pull failures on buildable services: `docker compose pull --ignore-buildable`, wdyt ?

**Related issue**
close https://github.com/docker/compose/issues/8724

image

### DIFF
--- a/pkg/compose/pull.go
+++ b/pkg/compose/pull.go
@@ -31,6 +31,7 @@ import (
 	moby "github.com/docker/docker/api/types"
 	"github.com/docker/docker/pkg/jsonmessage"
 	"github.com/docker/docker/registry"
+	"github.com/hashicorp/go-multierror"
 	"golang.org/x/sync/errgroup"
 
 	"github.com/docker/compose/v2/pkg/api"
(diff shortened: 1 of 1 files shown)


glours (approved):

floriangeigl:
hi @ndeloof - any chance that this broke the "--ignore-pull-failures" flag? I've always used it to pull images and ignore the one I build my self... but since the last update, it fails with
```
1 error occurred:
        * Error response from daemon: pull access denied for __________, repository does not exist or may require 'docker login': denied: requested access to the resource is denied
```

Issue #8752, opened 2021-10-05
`up --scale` not working on 2.0.1


ghost:
**Description**

`docker compose up --scale name=num` always reports "unknown service: name"

**Steps to reproduce the issue:**
```shell
$ cat docker-compose.yml
version: '3.8'

services:
  helloworld:
    image: alpine:edge
    entrypoint: /bin/echo Hello world
  scalable:
    image: alpine:edge
    entrypoint: /bin/echo Hello scalable
$ docker-compose -v
Docker Compose version 2.0.1
$ docker-compose up --scale scalable=0 -d helloworld
unknown service "scalable"
```

With docker-compose version 1.29.2 it works as expected:

```shell
$ docker-compose -v
docker-compose version 1.29.2, build unknown
$  docker-compose up --scale scalable=0 -d helloworld
Creating network "docker-compose-playground_default" with the default driver
Creating docker-compose-playground_helloworld_1 ... done
```

**Describe the results you received:**

Error: `unknown service "scalable"`

**Describe the results you expected:**

scalable service will not be started, all others should

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `docker compose version`:**

```
Docker Compose version 2.0.1
```

**Output of `docker info`:**

```
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Build with BuildKit (Docker Inc., v0.6.1-docker)
  compose: Docker Compose (Docker Inc., 2.0.1)

Server:
 Containers: 62
  Running: 28
  Paused: 0
  Stopped: 34
 Images: 170
 Server Version: 20.10.9
 Storage Driver: btrfs
  Build Version: Btrfs v5.14.1
  Library Version: 102
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 1
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 io.containerd.runtime.v1.linux runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 8686ededfc90076914c5238eb96c883ea093a8ba.m
 runc version: v1.0.2-0-g52b36a2d
 init version: de40ad0
 Security Options:
  seccomp
   Profile: default
 Kernel Version: 5.14.8-arch1-1
 Operating System: Arch Linux (containerized)
[...]

PR #10058, opened 2022-12-08
Fix corner case when there's no container to attach to

ndeloof:
**What I did**
Fixed a corner case: 
if one run `docker compose up --scale helloworld=0` with a single-service compose file, no container is actually created (as expected) and `watch` will never detect application termination

**Related issue**
close https://github.com/docker/compose/issues/8752

image


milas (approved):
Looks good - does this actually address the linked issue, however?

AFAICT that issue isn't about "no container to attach to" cases but `<svc>=0` being a way to exclude a service from launching as part of `up` without having to list out every other service

ndeloof:
https://github.com/docker/compose/issues/8752 report an issue that doesn't behaves as reported with latest codebase, but still exists as a deadlock when `scale=0` is set (aka "regression").

Issue #9554, opened 2022-06-10
docker compose V2 stops all containers related to project


Couga54:
`Docker Compose version v2.6.0`

I have project with a few file, like:
```
01_db.yml
02_web.yml
```
I run them with separate commands:
```
docker compose -f 01_db.yml -p test up -d
docker compose -f 02_web.yml -p test up -d
```

When executing command
`docker compose -f 01_db.yml -p test stop`
It stops all services that are related to "-p test" (all services in 01_db.yml and 02_web.yml), meanwhile docker-compose (V1) stop only services related to file and project

danger89:
I experience the same issue. I guess we now need to explicitly provide the service name as last argument?

danger89:
Or... Use `down` instead of `stop`. Down will also remove the image btw.

Example: `docker compose -f ...yml down`

sueskind:
I have the same issue, but `stop` and `down` behaves the same for me. (Docker Compose version 2.6.1)

PhilippeAccorsi:
We seen the same issue :(

Very disapointed with this v2. I need to re-install v1 on server to not broken all service each time.

yokonao:
In my case, this is a fatal issue for our migration to V2.
I want to know whether it is likely to be resolved.

PR #10062, opened 2022-12-09
load project from files when explicitly set by user

ndeloof:
As a user run `docker compose --file foo.yaml --project-name bar ...`, parse the provided compose file as source of truth for the compose model the command will execute upon. Project should only be built from actual resources when there's no explicit compose file being set and we only rely on a project name.

**Related issue**
fixes https://github.com/docker/compose/issues/9554

image


glours (approved):
LGTM

milas (approved):
Definitely more predictable behavior-wise this way!

Issue #8530, opened 2021-06-24
docker-compose V2 fails for port range and replicas>1 where V1 would succeed


manuel-koch:
**Description**

Using multiple replicas of a container where each replica should use one of the ports selected by a port-range:

```
version: "3.4"
services:
  dummy:
    image: busybox
    command: ["sh", "-c", "while true ; do sleep 3 ; date ; done"]
    deploy:
      replicas: 2
    ports: 
      - "6005-6015:5000"
```

When starting up these containers with docker-compose **V1** the container instances use port-mapping where each container picks one host-port from the selected port range to map to internal container port.

```
$ docker-compose ps
            Name                          Command               State                    Ports
----------------------------------------------------------------------------------------------------------------
docker-compose-issue_dummy_1   sh -c while true ; do slee ...   Up      0.0.0.0:6007->5000/tcp,:::6007->5000/tcp
docker-compose-issue_dummy_2   sh -c while true ; do slee ...   Up      0.0.0.0:6008->5000/tcp,:::6008->5000/tcp
```

When starting up these containers with docker-compose **V2** the containers use all ports from the selected port range to map to internal container port, resulting in error like the following for the second instance of that service:

`Error response from daemon: driver failed programming external connectivity on endpoint docker-compose-issue_dummy_1 (6d9cff4937b3b6946e0bb949a36362cf91477f841be34ac9bd2693453b6e8837): Bind for 0.0.0.0:6009 failed: port is already allocated`

**Steps to reproduce the issue:**

1. Switch to docker-compose V1, `docker-compose disable-v2`
2. Startup containers, `docker-compose up -d`
3. List running containers and the port mapping they use, `docker-compose ps`
4. Shutdown service, `docker-compose down`
5. Switch to docker-compose V2, `docker-compose enable-v2` 
6. Startup containers, `docker-compose up -d`
7. docker-compose fails to startup because of port mapping clash

**Describe the results you received:**

docker-compose.yaml only seems to work for V1 and not V2.

**Describe the results you expected:**

I would expect that docker-compose.yaml can be run by V1 and V2 alike.

**Output of `docker version`:**

```
Client:
 Cloud integration: 1.0.17
 Version:           20.10.7
 API version:       1.41
 Go version:        go1.16.4
 Git commit:        f0df350
 Built:             Wed Jun  2 11:56:22 2021
 OS/Arch:           darwin/amd64
 Context:           default
 Experimental:      true

Server: Docker Engine - Community
 Engine:
[...]

PR #10067, opened 2022-12-12
ContainerStart must run sequentially for engine to assing distinct ports within configured range

ndeloof:
**What I did**
A service configured with `port: 6000-6010:80` will get a random port assigned within this range
As we invoke `ContainerStart`, engine select an available port within container's config port range.
If we invoke those concurrently for containers within same service, engine will fail to assign a distinct port. 
Arguably, engine _should_ manage concurrent access and prevent this to happen, but .. ¯\\_(ツ)_/

**Related issue**
fixes https://github.com/docker/compose/issues/8530

image

### DIFF
--- a/pkg/compose/convergence.go
+++ b/pkg/compose/convergence.go
@@ -721,21 +721,17 @@ func (s *composeService) startService(ctx context.Context, project *types.Projec
 	}
 
 	w := progress.ContextWriter(ctx)
-	eg, ctx := errgroup.WithContext(ctx)
 	for _, container := range containers {
 		if container.State == ContainerRunning {
 			continue
 		}
-		container := container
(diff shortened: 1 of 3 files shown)


glours (approved):
LGTM, I'm little bit sad to see this move but I definitely prefer lower perf than port collisions 🤷‍♂️

laurazard (approved):
LGTM

NicolaeBasso:
Is there a solution for compose V3 without adding containerized NGINX, for example? 
This has been an issue for years already looks like..

pribic:
I am still getting the port mapping error on windows. Here is the version:
```
 >docker-compose -v
Docker Compose version v2.24.6-desktop.1
```
is it fixed for above version?

PR #10082, opened 2022-12-15
don't assume os.Stdout and rely on dockerCLI.streams

ndeloof:
**What I did**
avoid use of os.Stdout and prefer dockerCli.Out() 

**Related issue**
-

image

### DIFF
--- a/cmd/compose/compose.go
+++ b/cmd/compose/compose.go
@@ -31,7 +31,6 @@ import (
 	"github.com/docker/buildx/util/logutil"
 	dockercli "github.com/docker/cli/cli"
 	"github.com/docker/cli/cli-plugins/manager"
-	"github.com/docker/cli/cli/command"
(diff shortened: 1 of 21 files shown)


cmd/compose/compose.go
```diff
@@ -243,7 +242,7 @@ func RunningAsStandalone() bool {
 }
 
 // RootCommand returns the compose command with its child commands
-func RootCommand(dockerCli command.Cli, backend api.Service) *cobra.Command { //nolint:gocyclo
+func RootCommand(dockerCli api.Streams, backend api.Service) *cobra.Command { //nolint:gocyclo
```
milas, line 245: Nit: `s/dockerCli/streams`

milas (approved):
🧹

laurazard (approved):
LGTM

Issue #9648, opened 2022-07-13
Secrets fail to set the `uid`, `gid` and `mode` specified in `docker-compose.yml`


sanchitbapat:
**Description**

Docker secrets specified using the [long syntax](https://docs.docker.com/compose/compose-file/#long-syntax-3) for the `docker-compose.yml` file fail to set the specified `uid`, `gid` and `mode`. 

Also, from the docs, the default value of the `uid` and `gid` fields should be the user that runs the container however the value remains whatever was set on the host machine.

**Steps to reproduce the issue:**

1. Create a `Dockerfile`
```
❯ cat Dockerfile
FROM ubuntu:20.04

RUN adduser tester -u 1005 --disabled-password
CMD ["/bin/bash", "-c", "ls -al /run/secrets/target_secret_file.txt"]
```

2. Create a `docker-compose.yml`
```
❯ cat docker-compose.yml
services:
  secrets-tester:
    build: .
    secrets:
      - source: some_secret_file
        target: target_secret_file.txt
        uid: "1005"
        gid: "1005"
        mode: 0440

secrets:
  some_secret_file:
    file: somefile.txt
```
3. Create a text file for secrets (`somefile.txt`)
```
❯ cat somefile.txt
Text from a secret file
```

4. Run the service `docker compose run secrets-tester`

**Describe the results you received:**
Received Output:
```
❯ docker compose run secrets-tester
-rw-r--r-- 1 1000 1000 24 Jul 13 16:47 /run/secrets/target_secret_file.txt
```

**Describe the results you expected:**
Expected Output:
```
❯ docker compose run secrets-tester
-r--r----- 1 1005 1005 24 Jul 13 16:47 /run/secrets/target_secret_file.txt
```

**Additional information you deem important (e.g. issue happens only occasionally):**
Same behavior is observed in these cases:
1. Without creating the `tester` user
2. With same `source` and `target` names for the file in `docker-compose.yml`
3. Using the `uid` and `gid` for `root`
4. Using random values for `uid` and `gid`
5. Different values for `mode`

**Output of `docker compose version`:**

```
❯ docker --version
Docker version 20.10.17, build 100c701

❯ docker compose version
Docker Compose version v2.6.0
```

**Output of `docker info`:**

```
❯ docker info
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  app: Docker App (Docker Inc., v0.9.1-beta3)
  buildx: Docker Buildx (Docker Inc., v0.8.2-docker)
[...]

PR #10084, opened 2022-12-15
apply uid/gid when creating secret from environment

ndeloof:
**What I did**

**Related issue**
relates to https://github.com/docker/compose/issues/9648 but only applies to secret set from environment variable

image


glours (approved):
LGTM

milas (approved):

Issue #10085, opened 2022-12-16
docker-compose -p projectname ps ; lisiting services which not  even are build


aayushave:
OS - centos7
command should list nothing if container doesn't exist of that project, but is listing containers.

((**docker-compose -p ${DOCKER_PROJECT_NAME}   ps --services**  ) | grep "$(docker-compose -p ${DOCKER_PROJECT_NAME}   ps --services --filter 'status=running')" )

ndeloof:
`docker compose ps` indeed lists containers just like it is ran with `--all` option.
I'll look into this today

PR #10086, opened 2022-12-16
only list running containers when --all=false

ndeloof:
**What I did**
List containers according to `docker compose ps --all=?` flag to exclude non-running ones by default

**Related issue**
closes https://github.com/docker/compose/issues/10085

image

### DIFF
--- a/pkg/compose/ps.go
+++ b/pkg/compose/ps.go
@@ -32,7 +32,7 @@ func (s *composeService) Ps(ctx context.Context, projectName string, options api
 	if options.All {
 		oneOff = oneOffInclude
 	}
-	containers, err := s.getContainers(ctx, projectName, oneOff, true, options.Services...)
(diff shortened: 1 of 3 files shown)


milas (approved):

sayhiben:
After this change, I have to explicitly pass `-a` even though [the documentation](https://docs.docker.com/engine/reference/commandline/compose_ps/) for `docker compose ps` states that stopped containers should be returned

> By default, both running and stopped containers are shown:
>    ```
>    $ docker compose ps
>    NAME           COMMAND                  SERVICE   STATUS       PORTS
>    example-bar-1  "/docker-entrypoint.…"   bar       exited (0)
>    example-foo-1  "/docker-entrypoint.…"   foo       running      0.0.0.0:8080->80/tcp
>    ```

ndeloof:
@sayhiben doc indeed is outdated, including the example output. `docker compose ps` UX has been aligned with `docker ps` for consistency

=> https://github.com/docker/compose/pull/10195

PR #10095, opened 2022-12-18
[Docs] Suggestion to add information about environment

benjaoming:
Signed-off-by: Benjamin Balder Bach <benjamin@overtag.dk>

**What I did**

I wondered about what happens to a container's environment when you launch commands via `exec`. Is it preserved or do I have to re-do it.

Turns out for me (Compose v2) it's nicely preserved.

Added a little general tip for debugging the environment on a running container.

### DIFF
--- a/docs/reference/compose_exec.md
+++ b/docs/reference/compose_exec.md
@@ -28,3 +28,6 @@ This is the equivalent of `docker exec` targeting a Compose service.
 
 With this subcommand, you can run arbitrary commands in your services. Commands allocate a TTY by default, so
 you can use a command such as `docker compose exec web sh` to get an interactive prompt.
+
+Environment from existing configuration is applied.
+You can for instance introspect your environment by running `docker-compose exec web "/usr/bin/env"`.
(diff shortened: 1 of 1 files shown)


ndeloof (approved):

laurazard:
I'm concerned this change will introduce more confusion than clarity.

Running `compose exec [...]` doesn't launch a new container, it executes a command in a running container, so `Environment from existing configuration is applied` isn't technically true as `compose exec` isn't applying any new environment config, that simply already exists in the running container. Maybe we can make that clear, and a new PR can be made for that, but I don't think this exact change should be made.

docs/reference/compose_exec.md
```diff
@@ -24,3 +24,6 @@ This is the equivalent of `docker exec` targeting a Compose service.
 
 With this subcommand, you can run arbitrary commands in your services. Commands allocate a TTY by default, so
 you can use a command such as `docker compose exec web sh` to get an interactive prompt.
+
+Environment from existing configuration is applied.
```
laurazard, line 28: Maybe something like
```suggestion
Environment from existing configuration is preserved.
```
would be clearer.

laurazard (changes requested):

benjaoming:
haha auto-merge vs. stale bot

benjaoming:
I think by using GitHub's interface, there's something wrong with the commit data. Please feel free to use this improvement as you please.

Issue #10073, opened 2022-12-13
[BUG] cpu limit


nexcode:
### Description

https://github.com/docker/compose/issues/9268?ysclid=lbm4e4boux299377343
https://github.com/docker/compose/pull/9552

### Steps To Reproduce

compose example file:
```
version: "2.4"
services:
  service:
    image: clickhouse/clickhouse-server:22.6.9.11
    container_name: clickhouse
    network_mode: host
    restart: on-failure
    cpus: 8
```

docker inspect clickhouse | grep -i cpu:
```
            "CpuShares": 0,
            "NanoCpus": 0,
            "CpuPeriod": 0,
            "CpuQuota": 0,
            "CpuRealtimePeriod": 0,
            "CpuRealtimeRuntime": 0,
            "CpusetCpus": "",
            "CpusetMems": "",
            "CpuCount": 0,
            "CpuPercent": 0,
```

But, if manual set from docker `docker update --cpus 8 clickhouse`:
```
            "CpuShares": 0,
            "NanoCpus": 8000000000,
            "CpuPeriod": 0,
            "CpuQuota": 0,
            "CpuRealtimePeriod": 0,
            "CpuRealtimeRuntime": 0,
            "CpusetCpus": "",
            "CpusetMems": "",
            "CpuCount": 0,
            "CpuPercent": 0,
```

### Compose Version

```Text
Docker Compose version v2.12.2
```

### Docker Environment

```Text
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  app: Docker App (Docker Inc., v0.9.1-beta3)
  buildx: Docker Buildx (Docker Inc., v0.9.1-docker)
  compose: Docker Compose (Docker Inc., v2.12.2)
  scan: Docker Scan (Docker Inc., v0.21.0)

Server:
 Containers: 29
  Running: 28
  Paused: 0
  Stopped: 1
 Images: 237
 Server Version: 20.10.21
 Storage Driver: overlay2
  Backing Filesystem: xfs
  Supports d_type: true
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 io.containerd.runtime.v1.linux runc
[...]

PR #10100, opened 2022-12-19
set CPU quota

ndeloof:
**What I did**
fix set CPU quota on created containers
`CPUPercent` we used so far only applies to ... windows containers!

**Related issue**
fixes https://github.com/docker/compose/issues/10073
(this doesn't fix compose-go to replicate deploy.resources... to cpus)

image


glours (approved):
LGTM

Issue #9030, opened 2021-12-13
Docker compose logs with logging driver: none halts log output with error


lvaneijden:
**Description**
When using docker compose logs -f (V2) or docker-compose logs -f (V1) on my composed services with logging driver set to "none", it outputs some log lines follow by: _Error response from daemon: configured logging driver does not support reading_.

**Steps to reproduce the issue:**
1. Create a small docker-compose.yaml file (See for a subset of our services https://gist.github.com/lvaneijden/0397b2cc2cbac4bf9c6f828a27ae759e. Our actual docker-compose file contains about 30-40 services)
2. Start the services with docker compose up -d
3. After the services have started, type the command docker compose logs -f

**Describe the results you received:**
lex.van.eijden@Lexs-MacBook-Pro test-compose % docker compose up -d
[+] Running 8/8
 ⠿ Network test-compose_default                 Created                                                                                                                      0.0s
 ⠿ Container test-compose-memcached-1           Started                                                                                                                      0.9s
 ⠿ Container test-compose-mongo-1               Started                                                                                                                      1.1s
 ⠿ Container test-compose-local-smtp-catcher-1  Started                                                                                                                      1.6s
 ⠿ Container test-compose-fake-s3-1             Started                                                                                                                      1.3s
 ⠿ Container test-compose-mariadb-1             Started                                                                                                                      1.6s
 ⠿ Container test-compose-redis-1               Started                                                                                                                      1.2s
 ⠿ Container test-compose-mysql-1               Started                                                                                                                      1.4s
lex.van.eijden@Lexs-MacBook-Pro test-compose % docker compose logs -f
Error response from daemon: configured logging driver does not support reading

**Describe the results you expected:**
Expect to view the logging for any service that does not have logging driver: none set. 
Expected to not see any log lines for the services with logging driver: none set.

**Additional information you deem important (e.g. issue happens only occasionally):**
- When I start the services with docker compose up, without the -d flag to sent it to the background, the logging output is correct.
- I tried adjusting the logging driver to any other keyword: None and in quotes "none", neither gave a different result (This idea came from another issue I found: https://github.com/docker/compose/issues/8136)
- Adjusting the docker compose version from version 3 to 3.9 did nothing

**Output of `docker compose version`:**
```
lex.van.eijden@Lexs-MacBook-Pro test-compose % docker compose version
Docker Compose version v2.2.1
```

**Output of `docker info`:**

```
Client:
 Context:    default
 Debug Mode: false
 Plugins:
[...]

PR #10107, opened 2022-12-20
don't fail `logs` when driver:none is set

ndeloof:
**What I did**
detect engine response requesting container logs as `ErrNotImplemented` when logging driver is configured to `none`. This should not fail the command, just show a warning

bonus:
- configure `logContainers` to collect logs _after_ last watch event, so we don't get previous logs repeated after a container restart.
- detect container will restart using container inspect's `Restarting` attribute (not sure why we didn't used this before?)
- thanks to ^ `logs` can now detect container will not restart and then exits
- introduced an e2e test to cover `--follow`

**Related issue**
fixes: https://github.com/docker/compose/issues/9030

image

### DIFF
--- a/pkg/e2e/fixtures/logs-test/restart.yaml
+++ b/pkg/e2e/fixtures/logs-test/restart.yaml
@@ -1,0 +1,5 @@
+services:
+  ping:
+    image: alpine
+    command: "sh -c 'ping -c 1 localhost && exit 1'"
+    restart: "on-failure:2"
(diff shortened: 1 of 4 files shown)


glours (approved):
LGTM

pkg/e2e/fixtures/logs-test/restart.yaml
```diff
@@ -0,0 +1,5 @@
+services:
+  ping:
+    image: alpine
+    command: "sh -c 'ping -c 1 localhost && exit 1'"
+    restart: "on-failure:2"
```
glours, line 5: Nit: new line

glours (approved):
LGTM

nicksieger (approved):
Nice!

PR #10118, opened 2022-12-22
Coverage unit tests

glours:
we can't mesure our test coverage with e2e tests as they're using a built Compose binary and not source code directly

**What I did**
Attach code coverage report to our unit tests as we're using a Compose binary and not source when running e2e tests.
We're just monitoring `assert.go` and `framework.go` files 😞 

Screenshot 2022-12-22 at 11 12 47

Changes have been done to be aligned with `buildx` CI

image

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -130,8 +130,6 @@ jobs:
 
   e2e:
     runs-on: ubuntu-latest
-    env:
-      DESTDIR: "./bin/build"
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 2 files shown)


milas:
We could consider making it possible to run the E2E tests without using an external binary as an option so that coverage would work. That's also really nice for working locally because it means you can e.g. use breakpoints throughout test & CLI code

ndeloof:
we already run e2e tests as "standalone" and "cli plugin". We could add one to set `os.Args` and invoke `pluginMain`, not running external binary

PR #10124, opened 2022-12-27
Fix Cucumber 🥒 tests

laurazard:

**What I did**

Update Cucumber tests expected `compose ps` output to match new changes.

(fixing these so I can make a follow up PR to add a GHA for 🥒 tests)

Run them with `go test -v ./e2e`

**Related issue**

maxresdefault

### DIFF
--- a/e2e/cucumber-features/simple.feature
+++ b/e2e/cucumber-features/simple.feature
@@ -12,4 +12,4 @@ Background:
 Scenario: compose up
     When I run "compose up -d"
     Then the output contains "simple-1  Started"
-    And service "simple" is "running"
+    And service "simple" is "Up"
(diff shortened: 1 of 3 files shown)


e2e/cucumber-features/simple.feature
```diff
@@ -12,4 +12,4 @@ Background:
 Scenario: compose up
     When I run "compose up -d"
     Then the output contains "simple-1  Started"
-    And service "simple" is "running"
+    And service "simple" is "Up Less than a second"
```
ndeloof, line 15: shall we avoid depending on time since created output here, and only consider `Up`?

laurazard: yes! I meant to revisit this before pushing, and then I forgot. That's a recipe for flaky tests 😅

ndeloof (approved):

Issue #10125, opened 2022-12-08
docker compose up --pull doesn't seem to work


gsong:
### Description

`docker compose up --pull=always --wait` brings up the containers as expected but doesn't pull down the latest images.

### Reproduce

1. `docker compose up --pull=always --wait`

### Expected behavior

Pull the latest images prior to starting the services.

### docker version

```bash
Client:
 Cloud integration: v1.0.29
 Version:           20.10.21
 API version:       1.41
 Go version:        go1.18.7
 Git commit:        baeda1f
 Built:             Tue Oct 25 18:01:18 2022
 OS/Arch:           darwin/arm64
 Context:           desktop-linux
 Experimental:      true

Server: Docker Desktop 4.15.0 (93002)
 Engine:
  Version:          20.10.21
  API version:      1.41 (minimum version 1.12)
  Go version:       go1.18.7
  Git commit:       3056208
  Built:            Tue Oct 25 17:59:41 2022
  OS/Arch:          linux/arm64
  Experimental:     false
 containerd:
  Version:          1.6.10
  GitCommit:        770bd0108c32f3fb5c73ae1264f7e503fe7b2661
 runc:
  Version:          1.1.4
  GitCommit:        v1.1.4-0-g5fd4c4d
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

### docker info

```bash
Client:
 Context:    desktop-linux
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc., v0.9.1)
  compose: Docker Compose (Docker Inc., v2.13.0)
  dev: Docker Dev Environments (Docker Inc., v0.0.5)
  extension: Manages Docker extensions (Docker Inc., v0.2.16)
  sbom: View the packaged-based Software Bill Of Materials (SBOM) for an image (Anchore Inc., 0.6.0)
  scan: Docker Scan (Docker Inc., v0.22.0)

Server:
 Containers: 2
  Running: 2
  Paused: 0
  Stopped: 0
 Images: 25
 Server Version: 20.10.21
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 2
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
[...]

PR #10128, opened 2022-12-29
Set `pullChanged` when setting `--pull` on `compose up`

laurazard:

**What I did**

Fixed issue where setting `--pull=always|never|missing` on `compose up` had no effect due to the changed flag not being set.

Also added e2e tests for this scenario.

**Related issue**

fixes https://github.com/docker/compose/issues/10125

360_F_203318225_LfINEv5hni0kUNGWeus9VBrxg4Eg21dD


ndeloof:
can look into this in a follow-up PR, but I wonder we have to rely on such a `pullChanged` bool in the API while we can just check `Pull` attribute value to be non-empty.

ndeloof (approved):

Issue #10121, opened 2022-12-24
[BUG] docker compose config --output results an empty file because of the not flushed bufio writer


rimelek:
### Description

The "config" (or convert) subcommand of Docker Compose should write the output to a file instead of the standard output when I use the `--output output.yml` parameter. It creates the file but doesnt write anything into it. The result is an empty file.

### Steps To Reproduce

1. Create any compose file
2. run `docker compose config --output output.yml`
3. Test the output: `cat output.yml`

### Compose Version

```Text
The original: `Docker Compose version v2.13.0`
The custom build for testing from the Compose v2 repo: `Docker Compose version v2.14.2-20-gfd353ffa.m`
```

### Docker Environment

```Text
Client:
 Context:    desktop-linux
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc., v0.9.1)
  compose: Docker Compose (Docker Inc., v2.13.0)
  dev: Docker Dev Environments (Docker Inc., v0.0.5)
  extension: Manages Docker extensions (Docker Inc., v0.2.16)
  sbom: View the packaged-based Software Bill Of Materials (SBOM) for an image (Anchore Inc., 0.6.0)
  scan: Docker Scan (Docker Inc., v0.22.0)

Server:
 Containers: 24
  Running: 20
  Paused: 0
  Stopped: 4
 Images: 68
 Server Version: 20.10.21
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 2
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 io.containerd.runtime.v1.linux runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 770bd0108c32f3fb5c73ae1264f7e503fe7b2661
 runc version: v1.1.4-0-g5fd4c4d
 init version: de40ad0
 Security Options:
  seccomp
   Profile: default
  cgroupns
 Kernel Version: 5.15.49-linuxkit
 Operating System: Docker Desktop
 OSType: linux
 Architecture: aarch64
 CPUs: 3
 Total Memory: 3.841GiB
 Name: docker-desktop
[...]

PR #10129, opened 2022-12-29
Fix empty file when using compose config in case of smaller source files

rimelek:
"docker compose config --output out.yml" resulted an empty file when the generated output was smaller than 4097 bytes. bufio.Writer doesn't seem necessary since only one write operation will happen. This way there is no need for a new bufio.Writer that could be flushed.

**What I did**

First I replaced `bufio.NewWriter(file)` with `io.Writer(file)`, then based on suggestion of @thaJeztah I replaced the whole conditional file writing with a simple `io.WriteFile`.  See the related issue for my previous ideas.

**Related issue**
closes https://github.com/docker/compose/issues/10121

### DIFF
--- a/cmd/compose/convert.go
+++ b/cmd/compose/convert.go
@@ -17,11 +17,9 @@
 package compose
 
 import (
-	"bufio"
 	"bytes"
 	"context"
 	"fmt"
-	"io"
 	"os"
 	"sort"
 	"strings"
@@ -139,15 +137,10 @@ func runConvert(ctx context.Context, streams api.Streams, backend api.Service, o
 		return nil
 	}
 
-	var out io.Writer = streams.Out()
 	if opts.Output != "" && len(content) > 0 {
-		file, err := os.Create(opts.Output)
-		if err != nil {
-			return err
-		}
-		out = bufio.NewWriter(file)
+		return os.WriteFile(opts.Output, content, 0o666)
 	}
-	_, err = fmt.Fprint(out, string(content))
+	_, err = fmt.Fprint(streams.Out(), string(content))
 	return err
 }
 


cmd/compose/convert.go
```diff
@@ -145,7 +144,7 @@ func runConvert(ctx context.Context, streams api.Streams, backend api.Service, o
 		if err != nil {
 			return err
 		}
-		out = bufio.NewWriter(file)
+		out = io.Writer(file)
```
thaJeztah, line 147: Isn't `file` already a `io.Writer`?

Wondering if we're missing a `defer file.Close()`

thaJeztah: I guess since it's all writing at once, could also be `os.WriteFile(opts.Output, content, 0o666)` if a file is specified instead of stdout, which would take care of both creating, truncating, and writing to the file

rimelek: @thaJeztah The default output is `streams.Out()` and the changed line saved a `bufio.Writer` into `out` declared as `io.Writer`. `out` didn't have a method called `Close`, but I also tried to close the file. It didn't help. `bufio.Writer` needs to be flushed unless the output is larger the the buffer size. Closing the file wouldn't have solved this, since the content was not passed to the underlying writer. In this case, the `file`. Since the end of the `runConvert` method is using `fmt.Fprint` to write the content to a file or to stdout depending on the value of `out`,  I thought it was the easiest way to fix it, but you are right. I tried with this much shorter version to handle both stdout and the output file and worked:

```go
	if opts.Output != "" && len(content) > 0 {
		return os.WriteFile(opts.Output, content, 0o666)
	}
	_, err = fmt.Fprint(streams.Out(), string(content))
```

Should I force push the above code to the source branch?

thaJeztah: Thanks! Yes, that looks like the correct approach to me (but I'm not a maintainer on this repository, so it's possible that there's perhaps something I'm overlooking 🙈  😅)

thaJeztah (approved):
LGTM

cmd/compose/convert.go
```diff
@@ -139,15 +137,10 @@ func runConvert(ctx context.Context, streams api.Streams, backend api.Service, o
...
 	if opts.Output != "" && len(content) > 0 {
-		file, err := os.Create(opts.Output)
-		if err != nil {
-			return err
-		}
-		out = bufio.NewWriter(file)
+		return os.WriteFile(opts.Output, content, 0o666)
 	}
-	_, err = fmt.Fprint(out, string(content))
+	_, err = fmt.Fprint(streams.Out(), string(content))
```
thaJeztah, line 143: This could even be;

```go
_, err = streams.Out().Write(content)
```

But no need to change; there's tons of places where that's not used (so should perhaps be looked at throughout the code)

cmd/compose/convert.go
```diff
@@ -139,15 +137,10 @@ func runConvert(ctx context.Context, streams api.Streams, backend api.Service, o
 		return nil
 	}
 
-	var out io.Writer = streams.Out()
 	if opts.Output != "" && len(content) > 0 {
```
thaJeztah, line 140: The only thing I'm looking at here (but this was already the case); should `let(content) == 0` be an error condition? In what situation would an empty content be created, and should that be an error? 

Currently, the code skips creating the file (which is good), but then continues to writing "nothing" to stdout; maybe it should be an error if there's no content created 🤔 

(Probably better for maintainers to have a look at, and likely better for a follow-up)

cmd/compose/convert.go
rimelek: Thank you for sharing this so I can learn from it. I appreciate it.

cmd/compose/convert.go
rimelek: Good point! The code would cleaner, although I don't think the content could be empty without throwing an error before that line. The content of the compose file is validated before that, so even if I have an empty compose file, that would be invalid, therefore it couldn't reach the line of that check.  If I have a valid compose file, even if it doesn't have any services only a mapping that could be used for anchors like this:

```yaml
x-test:
  hello: world
```
It would result a longer and not shorter output:
```yaml
name: phptest
services: {}
x-test:
  hello: world
```
Even if I use `--format json`, the project name and the empty services mapping would be generated.

If there is a way we don't know about to have an empty output without error messages, simply just removing `&& len(content) > 0` could be a good way to solve this, so you could have an empty file as you would expect it when you have a valid empty output.

Of course other parts of the code could be changed later to eventually generate an empty and valid output, but I agree, this is something that the maintainers should decide how they want to handle it.

Again, thanks for pointing it out :) I enjoy learning about the code of Docker Compose.

ndeloof (approved):

Issue #9091, opened 2022-01-07
Docker Compose V2 doesn't seem to allow you to stop running builds in parallel, stopping solution from building.


Jaydee123:
**Description**
I have a docker-compose file that runs 10 .Net Microservices, after upgrading to Docker Desktop 4.3.2, Docker Engine v20.10.11, when I try to do 'docker-compose up --build' whilst Docker Compose v2 is enabled, the following issue happens:
![image](https://user-images.githubusercontent.com/9397730/148548246-854a1b88-a438-4817-9142-f37df63526a7.png)

When I run 'docker-compose disable-v2', the issue is resolved and the builds run sequentially.

**Steps to reproduce the issue:**
1. Have Docker Compose v2 enabled
2. Have a docker-compose file with 10 .net microservices that reference Dockerfiles.
3. Run 'docker-compose up --build'

**Describe the results you received:**
Build run in parallel and error:
![image](https://user-images.githubusercontent.com/9397730/148548483-b81134be-7cc6-452c-a78f-0f241a00522c.png)

**Describe the results you expected:**
Build runs in either parallel or sequential and pass.

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `docker compose version`:**

```
Docker Compose version v2.2.1
```

**Output of `docker info`:**

```
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc., v0.7.1)
  compose: Docker Compose (Docker Inc., v2.2.1)
  scan: Docker Scan (Docker Inc., v0.14.0)

Server:
 Containers: 26
  Running: 24
  Paused: 0
  Stopped: 2
 Images: 92
 Server Version: 20.10.11
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 1
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 io.containerd.runtime.v1.linux runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 7b11cfaabd73bb80907dd23182b9347b4245eb5d
 runc version: v1.0.2-0-g52b36a2
 init version: de40ad0
[...]

PR #10133, opened 2023-01-03
limit build concurrency according to --parallel

ndeloof:
**What I did**
This run the image build process on a per-service basis with concurrency constrained according to `--parallel` flag
- this allows to limit concurrent builds (https://github.com/docker/compose/issues/9091)
- this allow to control build oder based on `depends_on` directive - could be interesting to investigate a new depends_on condition `build` to offer fine tuning for this use-case. Can be used to offer a workaround on https://github.com/docker/compose/issues/8538

**Related issue**
closes https://github.com/docker/compose/issues/9091

image

### DIFF
--- a/cmd/compose/compose.go
+++ b/cmd/compose/compose.go
@@ -22,6 +22,7 @@ import (
 	"os"
 	"os/signal"
 	"path/filepath"
+	"strconv"
 	"strings"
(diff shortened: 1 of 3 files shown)


glours (approved):
LGTM

PR #10135, opened 2023-01-03
fix: remove invalid cache-from registry entries

johanneswuerbach:
**What I did**

When specifying:

```
cache_from:
  - type=gha
```

docker compose currently translates this into `--cache-from type=gha --cache-from type=registry,ref=type=gha`, which causes a warning.

This PR removes the explicit addition of `type=registry`, which shouldn't be necessary as the cache flag parsing handles this case already https://github.com/docker/buildx/blob/25ceb9067893be764f9ab439722835eece2fe169/util/buildflags/cache.go#L22

This code was originally added in https://github.com/docker/compose/commit/70694e12a2f266d6122e90bea4299c08eec003be, but I couldn't find any mention why it would be necessary.

**Related issue**

image

### DIFF
--- a/pkg/compose/build.go
+++ b/pkg/compose/build.go
@@ -72,12 +72,6 @@ func (s *composeService) build(ctx context.Context, project *types.Project, opti
 		if err != nil {
 			return err
 		}
-		for _, image := range service.Build.CacheFrom {
-			buildOptions.CacheFrom = append(buildOptions.CacheFrom, bclient.CacheOptionsEntry{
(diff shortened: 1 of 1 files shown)


ndeloof (approved):

ndeloof:
@johanneswuerbach sorry for the laaaaaate review on this PR. Could you please rebase your branch to fix conflict ?

johanneswuerbach:
Looks like this was already fixed in https://github.com/docker/compose/pull/10374

Issue #3909, opened 2016-09-04
abort-on-container-exit not working if a service is given 


fchennouf:
Hi everybody,

Is it normal that abort-on-container-exit option of the up command is not working if a service is given for example: docker-compose up --abort-on-container-exit web doesn't stop the db service if the web service stops, whereas docker-compose up --abort-on-container-exit stops it

version: '2'

services:

```
db:
      image: postgres:9.3

web:
      build: ./
      links:
              - db
```

Client:
 Version:      1.12.1
 API version:  1.24
 Go version:   go1.6.3
 Git commit:   23cf638
 Built:        Thu Aug 18 05:22:43 2016
 OS/Arch:      linux/amd64

Server:
 Version:      1.12.1
 API version:  1.24
 Go version:   go1.6.3
 Git commit:   23cf638
 Built:        Thu Aug 18 05:22:43 2016
 OS/Arch:      linux/amd64

docker-compose version 1.8.0, build f3628c7
docker-py version: 1.9.0
CPython version: 2.7.9
OpenSSL version: OpenSSL 1.0.1e 11 Feb 2013

shin-:
Hi!

I believe that is the expected behavior, yes. The command you're running `docker-compose up [...] web` is explicit about handling the `web` service, and not any other - having it stop the `db` service, which as far as compose is concerned is running independently, would seem very counter-intuitive to me. Is there a specific reason why you need to do things that way?

fchennouf:
Hi Shin,
Thanks for the answer. In my view  --abort-on-container-exit should work as long as linked containers are started. If I give a service like web in param, by default linked services are also started and I think that  --abort-on-container-exit should by default stopped these services. What I wanna do is to put two différent configurations of a service in the compose file (web1 and web2 for example), and i can precise one of the service if i want to separately start it with its linked containers.

Zoltu:
We also are running into problems with this.  We are using `docker-compose` for integration testing and so we do `docker-compose up ... test app_under_test`.  This will allow us to see the console output for `app_under_test` and `test` but it won't show us the massive spam of output from our dependencies like MySQL, Zookeeper, Kafka, etc. which in almost all cases we don't care about, isn't under our control, and is very noisy.

However, after the `test` container exits only the `app_under_test` also exits, which leaves all of the dependencies running indefinitely even though they are no longer used.  They just sit there consuming resources and (in some cases) bound to ports.

[...]

PR #10140, opened 2023-01-05
introduce ConvergenceResult and use it to also stop dependencies

ndeloof:
**What I did**
introduced `ConvergenceResult` to track all operations we applied on containers during convergence
This allows to track services we started by dependency, and should be stopped when we abort the app

**Related issue**
closes https://github.com/docker/compose/issues/3909

image


ndeloof:
obsolete

Issue #9739, opened 2022-08-10
Removing a container with piped yes causes a panic runtime error


jgeiger-trane:
**Description**

Removing a container using `yes | docker-compose rm redis` causes the application to crash with a panic runtime error.

**Steps to reproduce the issue:**
1.  Add docker-compose.yml
```
version: "3.9"
services:
  redis:
    image: "redis:alpine"
```
2. `docker-compose up redis`
3. ctrl-c to stop container
4. `yes | docker-compose rm redis`

**Describe the results you received:**
```
(base) ➜ yes | docker-compose rm redis
? Going to remove compose_testing-redis-1 (y/N) ^[[69;254R^[[28;49Rpanic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xf66be0]

goroutine 1 [running]:
github.com/AlecAivazis/survey/v2/terminal.(*RuneReader).ReadLineWithDefault.func2(...)
        github.com/AlecAivazis/survey/v2@v2.3.5/terminal/runereader.go:65
github.com/AlecAivazis/survey/v2/terminal.(*RuneReader).ReadLineWithDefault(0xc0003d0d80, 0x12a000?, {0x1cda3b0, 0x0, 0x0}, {0x0?, 0x280?, 0x40d4a5?})
        github.com/AlecAivazis/survey/v2@v2.3.5/terminal/runereader.go:328 +0x880
github.com/AlecAivazis/survey/v2/terminal.(*RuneReader).ReadLine(...)
        github.com/AlecAivazis/survey/v2@v2.3.5/terminal/runereader.go:37
github.com/AlecAivazis/survey/v2.(*Confirm).getBool(0xc00078b130, 0x0, 0xc000176b48)
        github.com/AlecAivazis/survey/v2@v2.3.5/confirm.go:59 +0x186
github.com/AlecAivazis/survey/v2.(*Confirm).Prompt(0xc00078b130, 0xc000176b48)
        github.com/AlecAivazis/survey/v2@v2.3.5/confirm.go:136 +0xf6
github.com/AlecAivazis/survey/v2.Ask({0xc0008bf4a0, 0x1, 0xc0008bf488?}, {0x10326a0?, 0xc00048dbe0}, {0xc0008bf498, 0x1, 0xc0008bf4e0?})
        github.com/AlecAivazis/survey/v2@v2.3.5/survey.go:334 +0x38c
[...]

PR #10144, opened 2023-01-06
use a simpler prompt implementation when we lack a terminal

ndeloof:
**What I did**
When stdin is not a terminal, use a simpler Prompt implementation to collect and parse user confirmation

make the survey implementation to rely on dockercli.In() on Out() streams - unfortunately, not using `Fd()` but `FD()` as func name, so need an adapter

**Related issue**
closes https://github.com/docker/compose/issues/9739

image


nicksieger (commented):
LGTM

nicksieger (approved):

PR #10145, opened 2023-01-06
don't filter by services if no filter was set

ndeloof:
**What I did**
My assumption regarding https://github.com/docker/roadmap/issues/418 is that services have been created with some profiles enabled. Then as `compose ps` is executed, without profiles, we get a reduced compose model. As code used to assume no filter == filter by all project services, we actually filter out services launched previously

**Related issue**
closes https://github.com/docker/roadmap/issues/418

image


nicksieger (approved):
LGTM

glours (approved):

PR #10148, opened 2023-01-06
Support for docker compose build --push

maxcleme:
Signed-off-by: maxcleme <maxime.clement@docker.com>

**What I did**

Support `docker compose build --push` the same way as `docker buildx build --push` does.

Oddly, I was expecting it to be already supported. It occurs to me when dealing with a simple compose file locally using two platforms (`linux/arm64` & `linux/amd64`). 

However when doing `docker compose push` it only pushes the local image (which makes sense since my store can't really store the other). Going through the same as `docker buildx build` does seems logical here.

This PR is a bit opportunistic, all feedbacks are welcome 😄 

**Related issue**

### DIFF
--- a/pkg/compose/build.go
+++ b/pkg/compose/build.go
@@ -86,12 +86,15 @@ func (s *composeService) build(ctx context.Context, project *types.Project, opti
 			Type: "docker",
 			Attrs: map[string]string{
 				"load": "true",
+				"push": fmt.Sprint(options.Push),
 			},
 		}}
 		if len(buildOptions.Platforms) > 1 {
 			buildOptions.Exports = []bclient.ExportEntry{{
-				Type:  "image",
-				Attrs: map[string]string{},
+				Type: "image",
+				Attrs: map[string]string{
+					"push": fmt.Sprint(options.Push),
+				},
 			}}
 		}
 		opts := map[string]build.Options{imageName: buildOptions}
--- a/pkg/compose/build_classic.go
+++ b/pkg/compose/build_classic.go
@@ -59,6 +59,12 @@ func (s *composeService) doBuildClassic(ctx context.Context, project *types.Proj
 			errs = multierror.Append(errs, err).ErrorOrNil()
 		}
 		nameDigests[imageName] = digest
+		if errs != nil {
+			return nil
+		}
+		if len(o.Exports) != 0 && o.Exports[0].Attrs["push"] == "true" {
+			return s.push(ctx, project, api.PushOptions{})
+		}
 		return nil
 	})
 	if err != nil {
--- a/cmd/compose/build.go
+++ b/cmd/compose/build.go
@@ -38,6 +38,7 @@ type buildOptions struct {
 	composeOptions
 	quiet    bool
 	pull     bool
+	push     bool
 	progress string
 	args     []string
 	noCache  bool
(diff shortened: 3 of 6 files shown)


ndeloof (changes requested):
LGTM but for this option to be introduced, the "classic" builder also need to offer this feature

maxcleme:
@ndeloof What do you mean by "classic builder" ? Do you mind sharing more pointers? I'll be glad to update accordingly :)

ndeloof:
in this PR you only updated the options, but the actual build implementation is here:
https://github.com/docker/compose/blob/v2/pkg/compose/build_buildkit.go
https://github.com/docker/compose/blob/v2/pkg/compose/build_classic.go

first one uses buildkit (same as `docker buildx`), the other uses the classic build mechanism on docker engine

pkg/compose/build.go
```diff
@@ -48,7 +48,6 @@ func (s *composeService) Build(ctx context.Context, project *types.Project, opti
 }
 
 func (s *composeService) build(ctx context.Context, project *types.Project, options api.BuildOptions) error {
-	opts := map[string]build.Options{}
```
maxcleme, line 51: Move it to closure below otherwise it could end up with `concurrent map write` panic. It occurs on my end when building multi service with buildx.

maxcleme: Was spotted and merged in https://github.com/docker/compose/pull/10151, I have rebased accordingly.

maxcleme:
Any news on this? 🤔

glours:
I did some tests, it works well except for one use case. When you try to build and push an multi-arch image with the legacy builder we got the following error message :
```shell
~/sources/compose/pkg/e2e/fixtures/build-test/platforms feat/support_multiarch_push !1 > DOCKER_BUILDKIT=0 docker compose build --push                                                                 19:02:19
An image does not exist locally with the tag: gloursdocker/build-test-platform
```
instead of the message saying that the builder doesn't support the multi-arch builds
```shell
~/sources/compose/pkg/e2e/fixtures/build-test/platforms feat/support_multiarch_push !1 > DOCKER_BUILDKIT=0 docker compose build                                                                        19:02:46
1 error occurred:
	* this builder doesn't support multi-arch build, set DOCKER_BUILDKIT=1 to use multi-arch builder
```

maxcleme:
@glours Nice catch indeed! Should be fix now, do you mind giving it a try again? :)

glours:
Looks good 👍

glours (approved):
LGTM

pkg/compose/build_classic.go
```diff
@@ -59,6 +59,12 @@ func (s *composeService) doBuildClassic(ctx context.Context, project *types.Proj
 			errs = multierror.Append(errs, err).ErrorOrNil()
 		}
 		nameDigests[imageName] = digest
+		if errs != nil {
+			return nil
+		}
+		if len(o.Exports) != 0 && o.Exports[0].Attrs["push"] == "true" {
```
ndeloof, line 65: this really demonstrates we should not use `build.Options ` as input for classic builder, but that's fine for this PR

ndeloof (approved):

pkg/compose/build_classic.go
glours: Agree, I'll try to find time to review that

Issue #8797, opened 2021-10-14
docker compose up --build    or  docker compose build   and config.json for proxy


bb980701:
**Description**

The builder of the image do not use the proxy setting set in the config.json / in the daemon or as local env variables as runtime build argument  .

**Steps to reproduce the issue:**
0. Have a win10 coporate machine setup under windows where all internet traffic have to pass through a proxy
   where 
          Environment variables HTTP_PROXY , HTTPS_PROXY  are set at windows
          Docker Desktop settings > Resources > Proxies are ser
          .C:\Users\<winuser>\.docker\config.json    proxies property are set

1. 
docker-compose.yml 

```
version: "3.9"
services:
  agent:
    container_name: test_proxy
    build:
      context: .
      dockerfile: Dockerfile

```

start.sh
```
#!/bin/sh
set
sleep infinity
```

Dockerfile

```
FROM node:14-alpine
RUN set 
RUN apk add tzdata
RUN cp /usr/share/zoneinfo/UTC /etc/localtime
COPY ./start.sh .
RUN chmod +x start.sh
ENTRYPOINT [ "sh" , "./start.sh" ]
```

**Describe the results you received:**

a. docker compose up --build

![image](https://user-images.githubusercontent.com/71716559/137274369-9bc4f2b1-52d3-44a0-8ba6-c063ce85139f.png)
at the apk add step the build hang for ever , since impossible to fetch the package

it returns a fetch error since the build process do not have the proxy env variable

looking further why , with

b. docker compose build --progress plain --no-cache

the FROM use probably the Daemon Proxy setting from the Windows Docker Desktop to get the alpine image
BUT the RUN set command show clearly that the ENV variables related to proxy are no set, know at RUN command

![image](https://user-images.githubusercontent.com/71716559/137274645-7b854e6c-b937-42df-908a-7a5ab997d895.png)

**Describe the results you expected:**
For me it is a regression since with Docker Desktop 3.3.3  and at that time docker-compose 

![image](https://user-images.githubusercontent.com/71716559/137278151-54100fcb-d752-4e12-ba0f-6ca944dbfc6e.png)

```
docker-compose version 
docker-compose version 1.29.1, build c34c88b2
docker-py version: 5.0.0
CPython version: 3.9.0
OpenSSL version: OpenSSL 1.1.1g  21 Apr 2020

```

it was working without troubles, all command in the Dockerfile are aware of the proxy setting

docker-compose build --progress plain --no-cache

[...]

PR #10159, opened 2023-01-10
pass proxy config as build args - same as docker/buildx#959

ndeloof:
**What I did**
Apply same fix as https://github.com/docker/buildx/pull/959

**Related issue**
closes https://github.com/docker/compose/issues/8797


glours (approved):
LGTM

Issue #9718, opened 2022-08-04
docker-compose create suggests --remove-orphans, which doesn't work


codl:
**Description**

If it finds orphaned containers, `docker-compose create` will suggest running again with `--remove-orphans`, which is not a flag that `create` accepts.

**Steps to reproduce the issue:**
1. `docker-compose create`
2. change the name of a service
3. `docker-compose create` again, which will complain:
   ```
   WARN[0000] Found orphan containers ([foo, bar]) for this project. If you removed or renamed this service in your compose file, you can run this command with the --remove-orphans flag to clean it up.
   ```
4. `docker-compose create --remove-orphans`

**Describe the results you received:**
```
$ docker-compose create --remove-orphans
unknown flag: --remove-orphans
```

**Describe the results you expected:**
Orphaned containers are removed.

**Output of `docker compose version`:**

```
Docker Compose version 2.9.0
```

**Output of `docker info`:**

```
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc., v0.8.2-docker)
  compose: Docker Compose (Docker Inc., 2.9.0)

Server:
 Containers: 6
  Running: 4
  Paused: 0
  Stopped: 2
 Images: 66
 Server Version: 20.10.17
 Storage Driver: btrfs
  Build Version: Btrfs v5.18.1
  Library Version: 102
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runtime.v1.linux runc io.containerd.runc.v2
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 10c12954828e7c7c9b6e0ea9b0c02b01407d3ae1.m
 runc version: 
 init version: de40ad0
 Security Options:
  seccomp
   Profile: default
  cgroupns
 Kernel Version: 5.15.55-2-lts
 Operating System: Arch Linux
 OSType: linux
 Architecture: x86_64
 CPUs: 8
 Total Memory: 7.456GiB
 Name: peanut.codl.fr
 ID: HTBR:M3DN:3IFA:VREG:7FMW:REU4:X23W:UUSQ:33P4:WQ74:CZTP:IHAT
 Docker Root Dir: /var/lib/docker
[...]

PR #10160, opened 2023-01-10
introduce `--remove-orphans` in `compose create` command

ndeloof:
**What I did**
introduce `--remove-orphans` in `compose create` command to align with `compose up` and make orphan warning suggestion relevant

**Related issue**
closes https://github.com/docker/compose/issues/9718


milas (approved):
TIL We have `--remove-orphans` on `up` and not just `down` 😵

PR #10163, opened 2023-01-11
[ENV-44] introduce experimental watch command (skeletton)

ndeloof:
**What I did**
introduced `watch` command skeletton as an experimental command
watch is a subcommand for the new (and hidden) `alpha` command, so user can run `docker compose alpha watch`, knowing this is experimental and subject to changes. Once ready for announcement, command will be promoted to top-level compose command.

**Related issue**

### DIFF
--- a/docs/reference/docker_compose_alpha.yaml
+++ b/docs/reference/docker_compose_alpha.yaml
@@ -1,0 +1,15 @@
+command: docker compose alpha
+short: Experimental commands
+long: Experimental commands
+pname: docker compose
+plink: docker_compose.yaml
+cname:
+    - docker compose alpha watch
+clink:
+    - docker_compose_alpha_watch.yaml
+deprecated: false
+experimental: false
+experimentalcli: false
+kubernetes: false
+swarm: false
+
--- a/cmd/compose/alpha.go
+++ b/cmd/compose/alpha.go
@@ -1,0 +1,31 @@
+/*
+
+   Copyright 2020 Docker Compose CLI authors
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+       http://www.apache.org/licenses/LICENSE-2.0
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package compose
+
+import (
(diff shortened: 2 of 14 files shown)


milas (approved):
Looks like a good start!

---

FYI this won't actually watch recursively on the context directory - though not suggesting that be part of this PR as it's gonna be a loooot more code! 😬 

More file-watching brain dump:

For Linux, it's still necessary to walk the file tree and add watchers for each directory entry (on start and keep it up-to-date as directories are created/removed at runtime 🫣). Here's the Tilt watcher: https://github.com/tilt-dev/tilt/blob/master/internal/watch/watcher_naive.go (note: this uses [tilt-dev/fsnotify](https://github.com/fsnotify/fsnotify/compare/main...tilt-dev:fsnotify:main) fork!)

For Windows, recursive support should be available in the next `fsnotify/fsnotify` release: see https://github.com/fsnotify/fsnotify/pull/540 - note that this was inspired by the Tilt fork but uses the more Go-like `/...` path suffix instead of a method/option on the watcher itself

For macOS, `fsnotify` only supports `kqueue` (single files) rather than FSEvents (entire directories). [Apple has this to say about them](https://developer.apple.com/library/archive/documentation/Darwin/Conceptual/FSEvents_ProgGuide/KernelQueues/KernelQueues.html):
> File system events are intended to provide notification of changes with directory-level granularity. For most purposes, this is sufficient. In some cases, however, you may need to receive notifications with finer granularity. For example, you might need to monitor only changes made to a single file. For that purpose, the kernel queue (kqueue) notification system is more appropriate.
>
> If you are monitoring a large hierarchy of content, you should use file system events instead, however, because kernel queues are somewhat more complex than kernel events, and can be more resource intensive because of the additional user-kernel communication involved.

Again, to use Tilt as an example, here's how it uses FSEvents: https://github.com/tilt-dev/tilt/blob/master/internal/watch/watcher_darwin.go

laurazard (approved):
LGTM 🥳 🔄

glours (approved):

docs/reference/docker_compose_alpha.yaml
```diff
@@ -0,0 +1,15 @@
...
+long: Experimental commands
+pname: docker compose
+plink: docker_compose.yaml
+cname:
+    - docker compose alpha watch
+clink:
+    - docker_compose_alpha_watch.yaml
+deprecated: false
+experimental: false
+experimentalcli: false
```
thaJeztah, line 12: Perhaps we need to set the `experimentalcli` annotation for this. We removed the "gate" for disabling/hiding those commands, but I guess we could still use the annotation to be used on docs.docker.com. (happy to discuss / brainstorm)

PR #10166, opened 2023-01-11
Add remove-orphans functionality to run

matthewarmand:
**What I did**

Naively followed @ndeloof's example in https://github.com/docker/compose/pull/10160 to add support for the `--remove-orphans` flag on `docker compose run`, since that command recommends the flag on certain errors/warnings, but doesn't support it yet.

**Related issue**

Related to https://github.com/docker/compose/issues/9718#issuecomment-1209448445, though that issue has already been closed. I can open a new one specific to `run` if that's desired.

This is Czernobog, who was adopted off the street as an orphan. We will **not**, however, be supplying the `--remove-orphans` flag to remove her.

image

### DIFF
--- a/cmd/compose/run.go
+++ b/cmd/compose/run.go
@@ -162,7 +162,7 @@ func runCommand(p *ProjectOptions, streams api.Streams, backend api.Service) *co
 	flags.BoolVar(&opts.servicePorts, "service-ports", false, "Run command with the service's ports enabled and mapped to the host.")
 	flags.BoolVar(&opts.quietPull, "quiet-pull", false, "Pull without printing progress information.")
 	flags.BoolVar(&createOpts.Build, "build", false, "Build image before starting container.")
-	flags.BoolVar(&opts.removeOrphans, "remove-orphans", false, "Remove containers for services not defined in the Compose file.")
+	flags.BoolVar(&createOpts.removeOrphans, "remove-orphans", false, "Remove containers for services not defined in the Compose file.")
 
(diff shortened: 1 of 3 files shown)


ndeloof (approved):

laurazard:
awesome, thanks @matthewarmand! just run `make docs` and commit the changes to update the docs with the new argument.

matthewarmand:
Hi @laurazard, thanks! I noticed that message in the pipeline output but I'm having trouble running `make docs` locally. Am I missing a setup step or something?

```
$ make docs
docker buildx bake --set "*.output=type=local,dest=/tmp/compose-output.OlzO6qcrVj" docs-update
unknown flag: --set
See 'docker --help'.
...
```

Docker version: `Docker version 20.10.22, build 3a2c30b63a `

laurazard:
Oh, weird. It looks like maybe you don't have buildx installed? Could you check the output of `docker buildx ls`, `docker buildx version`? Some docs [here](https://docs.docker.com/build/install-buildx/), repo [here](https://github.com/docker/buildx).

matthewarmand:
Ah, here we go. I'm on Arch Linux and I was under the mistaken assumption that `buildx` was packaged with Docker on that OS when it's not. I've installed [`docker-buildx`](https://archlinux.org/packages/community/x86_64/docker-buildx/) and now successfully run `make docs`, see https://github.com/docker/compose/pull/10166/commits/dadad01e6351b723b6849dbc75d7bebd5dea389e

Issue #9962, opened 2022-10-27
[BUG] Output of "docker-compose pull" is unusable


williamdes:
### Description

I use Guake terminal and when I use `docker-compose pull` it's an infinie flow of updates in my terminal. If I scroll up it "freezes" the progress and continues to push content below.

The infinite flow of content pushed and never cleared is unusable.

### Steps To Reproduce

To  reproduce maybe clone https://github.com/wdesportes/phpmyadmin-local-setup and run `./dockerl pull`

### Compose Version

```Text
Docker Compose version v2.12.2
```

### Docker Environment

```Text
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  app: Docker App (Docker Inc., v0.9.1-beta3)
  buildx: Build with BuildKit (Docker Inc.)
  sbom: View the packaged-based Software Bill Of Materials (SBOM) for an image (Anchore Inc., 0.6.1)
  scan: Docker Scan (Docker Inc., v0.17.0)

Server:
 Containers: 0
  Running: 0
  Paused: 0
  Stopped: 0
 Images: 344
 Server Version: 20.10.14
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 io.containerd.runtime.v1.linux runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 3df54a852345ae127d1fa3092b95168e4a88e2f8
 runc version: v1.0.3-0-gf46b6ba
 init version: de40ad0
 Security Options:
  apparmor
  seccomp
   Profile: default
  cgroupns
 Kernel Version: 5.15.0-50-generic
 Operating System: Ubuntu 22.04.1 LTS
 OSType: linux
 Architecture: x86_64
 CPUs: 12
 Total Memory: 62.45GiB
 Name: williamdes
 ID: xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
 Docker Root Dir: /mnt/Docker-storage
 Debug Mode: false
 Username: williamdes
[...]

Issue #9500, opened 2022-05-24
Docker Compose pull don't clean terminal


JADC362:
**Description**

`docker compose` pull of images doesn't clean the terminal, making it really complicated to have a clean & useful output log of the progress of image download.

Despite the output log of `docker pull` and `docker-compose pull (or up)`, where the cmd cleans the terminal everytime will print out, `docker compose` don't clean the terminal, adding thousand of lines to the terminal stdout.

**Steps to reproduce the issue:**
For a given `docker-compose.yml `configuration file of services:
1. Remove the images from the working machine.
2. Call `docker compose up`. The process of pull will start, since the images were previously removed.
3. See the output (appreciated when the pull image has several layers).

**Describe the results you received:**

Thousand of lines are added to the stdout of the terminal, filling the entire window with repetitive information"
(output log was clean a little bit just for clarity)
```
[+] Running 22/109Waiting
 ⠸ service-compose Pulling
   ⠿ b282d58c3583 Already exists
   ⠿ a723e6be38fb Already exists
   ⠇ 4b701269f196 Downloading    [=========>                                         ]  242.2kB/1.318MB
   ⠇ d47090bab477 Downloading    [>                                                  ]  432.7kB/42.24MB
   ⠇ ad3dc22ab53c Downloading    [==>                                                ]  3.764MB/65.44MB
   ⠇ 67f5b059511e Waiting
   ⠇ a0eb0f94e1ac Waiting
   ⠇ e25f2676785a Waiting
   ⠇ f493a417a455 Waiting
   ⠇ 03c15bc55f17 Waiting
   ⠇ 35e6c5cef9ba Waiting
   ⠇ 48f6053856a5 Waiting
   ⠇ 6767a1cb8eab Pulling fs layer

[+] Running 22/109Waiting
 ⠸ service-compose Pulling
   ⠿ 4e524281c0e3 Already exists
   ⠿ 376f715ed127 Already exists
   ⠿ b282d58c3583 Already exists
   ⠿ a723e6be38fb Already exists
   ⠏ 4b701269f196 Downloading    [=========>                                         ]  259.6kB/1.318MB
   ⠏ d47090bab477 Downloading    [>                                                  ]  432.7kB/42.24MB
   ⠏ ad3dc22ab53c Downloading    [==>                                                ]  3.764MB/65.44MB
   ⠏ 67f5b059511e Waiting
   ⠏ 48f6053856a5 Waiting
   ⠏ 6767a1cb8eab Pulling fs layer
[...]

Issue #9377, opened 2022-04-11
Ugly CLI output when it exceeds terminal's height


MikeYermolayev:
**Description**

This issue is based on the discussion https://forums.docker.com/t/docker-compose-always-appends-to-console-output-instead-of-updating-it/122841/3

When the expected output exceeds my terminal's height then it's very hard to follow it since every update gets appended instead of replacing.

**Steps to reproduce the issue:**
1. Have docker-compose file with big amount of services (mine has around 30)
2. Run `docker-compose up` in terminal. 

**Describe the results you received:**
Every update to stdout gets appended instead of replacing current one

**Describe the results you expected:**
Update should replace current output so it's easy to understand the progress of operations

**Additional information you deem important (e.g. issue happens only occasionally):**
issue is happening only when terminal's height is lower than expected output height

**Output of `docker compose version`:**

```
Docker Compose version v2.3.3
```

**Output of `docker info`:**

```
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc., v0.8.1)
  compose: Docker Compose (Docker Inc., v2.3.3)
  scan: Docker Scan (Docker Inc., v0.17.0)

Server:
 Containers: 63
  Running: 44
  Paused: 0
  Stopped: 19
 Images: 166
 Server Version: 20.10.13
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 2
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runtime.v1.linux runc io.containerd.runc.v2
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 2a1d4dbdb2a1030dc5b01e96fb110a9d9f150ecc
 runc version: v1.0.3-0-gf46b6ba
 init version: de40ad0
 Security Options:
  seccomp
   Profile: default
  cgroupns
 Kernel Version: 5.10.104-linuxkit
 Operating System: Docker Desktop
 OSType: linux
 Architecture: x86_64
 CPUs: 4
 Total Memory: 3.842GiB
 Name: docker-desktop
[...]

PR #9173, opened 2022-02-14
Avoid pulling same images multiple times ⚡️

KoditkarVedant:
**What I did**
`docker compose pull` pulls images even if they are present locally and if there are multiple services using the same image it pulls them multiple times. This pull request avoid pulling image if it is already present and only fetch image once if required by many services.

**Related issue**
#8768 

**(not mandatory) A picture of a cute animal, if possible in relation with what you did**
image

PR #10168, opened 2023-01-12
Shorten tty output when terminal is too small

laurazard:

We've had lots of complaints about the new TUI breaking when output exceeds terminal height. Related discussion can be found in these issues:
- https://github.com/docker/compose/issues/9377
- https://github.com/docker/compose/issues/8800
- https://github.com/docker/compose/issues/8753
- https://github.com/docker/compose/issues/9500
- https://github.com/docker/compose/issues/9962

The issue was partly addressed by the changes to detect when multiple services use the same images in https://github.com/docker/compose/pull/9173, but the root issue persists. Whenever have more lines to write than the terminal height is, we're unable to scroll back far enough to rewrite the buffer, and leave dangling lines behind, filling up the terminal scrollback and resulting in bad UX.

**What I did**

Detect when the number of events we have to display is > than terminal height, and adjust the output to omit child events when that's the case. This isn't a perfect solution (if the number of services > terminal height we will still run into issues, and the shortened output isn't very explicit), but it only kicks in in cases where the output would otherwise be worse, so it's an improvement over the current situation.

**Before**:

https://user-images.githubusercontent.com/70572044/211946344-7a11e231-0087-4820-9caa-88c32a46d8b3.mov

After:

https://user-images.githubusercontent.com/70572044/211946808-780d8f8d-3194-459a-8c5b-3c45771c913f.mov

IMG_20221224_193236

### DIFF
--- a/pkg/progress/tty.go
+++ b/pkg/progress/tty.go
@@ -157,6 +157,10 @@ func (w *ttyWriter) print() {
 		}
 	}
 
+	skipChildEvents := false
+	if len(w.eventIDs) > goterm.Height()-1 {
+		skipChildEvents = true
+	}
 	numLines := 0
 	for _, v := range w.eventIDs {
 		event := w.events[v]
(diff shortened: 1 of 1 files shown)


ndeloof:
Nice !! 🥳

PR #10169, opened 2023-01-12
Adjust terminal height calc

laurazard:

**What I did**

Just adjust the number from https://github.com/docker/compose/pull/10168, to account for some edge cases

**Related issue**

### DIFF
--- a/pkg/progress/tty.go
+++ b/pkg/progress/tty.go
@@ -158,7 +158,7 @@ func (w *ttyWriter) print() {
 	}
 
 	skipChildEvents := false
-	if len(w.eventIDs) > goterm.Height()-1 {
+	if len(w.eventIDs) > goterm.Height()-2 {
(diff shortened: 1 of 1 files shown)


pkg/progress/tty.go
```diff
@@ -158,7 +158,7 @@ func (w *ttyWriter) print() {
 	}
 
 	skipChildEvents := false
-	if len(w.eventIDs) > goterm.Height()-1 {
+	if len(w.eventIDs) > goterm.Height()-2 {
```
glours, line 161: Can you use a const with an explicit name for the magic `-2` number?

laurazard: ahh sorry, I was too quick on the merge 😓. I'll change it to something like `terminalOffset`, it's just to account for statuslines/other edge cases.

PR #10170, opened 2023-01-12
mark alpha command as experimental

ndeloof:
Mark experimental CLI commands with annotation so docs will reflect it
based on @thaJeztah comment https://github.com/docker/compose/pull/10163#pullrequestreview-1245295414

### DIFF
--- a/cmd/compose/alpha.go
+++ b/cmd/compose/alpha.go
@@ -25,6 +25,9 @@ func alphaCommand(p *ProjectOptions, backend api.Service) *cobra.Command {
 		Short:  "Experimental commands",
 		Use:    "alpha [COMMAND]",
 		Hidden: true,
+		Annotations: map[string]string{
+			"experimentalCLI": "true",
+		},
 	}
 	cmd.AddCommand(watchCommand(p, backend))
 	return cmd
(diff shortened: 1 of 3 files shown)


glours (approved):
👌

cmd/compose/alpha.go
```diff
@@ -25,6 +25,10 @@ func alphaCommand(p *ProjectOptions, backend api.Service) *cobra.Command {
 		Short:  "Experimental commands",
 		Use:    "alpha [COMMAND]",
 		Hidden: true,
+		Annotations: map[string]string{
+			"experimental":    "true",
```
thaJeztah, line 29: I think we only need the `experimentalCLI` part, as this one may (not sure if that's implemented in compose) hide the command based on whether the daemon has experimental features enabled.

cmd/compose/watch.go
```diff
@@ -44,6 +44,10 @@ func watchCommand(p *ProjectOptions, backend api.Service) *cobra.Command {
 			return runWatch(ctx, backend, opts, args)
 		}),
 		ValidArgsFunction: completeServiceNames(p),
+		Annotations: map[string]string{
```
thaJeztah, line 47: If "watch" is a subcommand of "alpha", then _probably_ this isn't needed (but double check if I'm right); IIRC, we make the YAMLdocs check if parents are experimental, and make child-commands inherit that.

PR #10172, opened 2023-01-13
[ENV-40] debounce refresh requests with quietperiod

ndeloof:
**What I did**
As we capture fs events, collect them to list services which need to be updated, with debounce and a quiet period

### DIFF
--- a/pkg/compose/watch.go
+++ b/pkg/compose/watch.go
@@ -18,10 +18,14 @@ import (
 	"context"
 	"fmt"
 	"log"
+	"strings"
+	"time"
 
 	"github.com/compose-spec/compose-go/types"
 	"github.com/docker/compose/v2/pkg/api"
+	"github.com/docker/compose/v2/pkg/utils"
 	"github.com/fsnotify/fsnotify"
+	"github.com/jonboulle/clockwork"
 	"github.com/mitchellh/mapstructure"
 	"github.com/pkg/errors"
 	"golang.org/x/sync/errgroup"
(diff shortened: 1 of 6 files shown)


pkg/compose/watch.go
```diff
@@ -77,3 +91,23 @@ func (s *composeService) Watch(ctx context.Context, project *types.Project, serv
...
+
+func debounce(ctx context.Context, interval time.Duration, input chan string, fn func(services []string)) {
+	services := utils.Set[string]{}
+	timer := time.NewTimer(interval)
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case service := <-input:
+			timer.Reset(interval)
```
milas, line 103: Since we don't know if the timer has fired here, we need to stop it/drain first:
```go
if !timer.Stop() {
	<-timer.C
}
timer.Reset(interval)
```

https://pkg.go.dev/time#Timer.Reset

pkg/compose/watch.go
```diff
@@ -30,10 +33,20 @@ import (
...
+const quietPeriod = 2 * time.Second
+
 func (s *composeService) Watch(ctx context.Context, project *types.Project, services []string, options api.WatchOptions) error {
 	fmt.Fprintln(s.stderr(), "not implemented yet")
 
 	eg, ctx := errgroup.WithContext(ctx)
+	needRefresh := make(chan string)
+	eg.Go(func() error {
+		debounce(ctx, quietPeriod, needRefresh, func(services []string) {
+			fmt.Fprintf(s.stderr(), "Updating %s after changes were detected", strings.Join(services, ", "))
```
milas, line 45:
```suggestion
			fmt.Fprintf(s.stderr(), "Updating %s after changes were detected\n", strings.Join(services, ", "))
```

(Missing `\n`)

pkg/compose/watch_test.go
```diff
@@ -0,0 +1,29 @@
...
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"gotest.tools/v3/assert"
+)
+
+func Test_debounce(t *testing.T) {
```
milas, line 11: You might take a look at Go libs out there to help with controlling timers deterministically in tests, as this is likely to become a problematic/failing test in CI otherwise

Tilt uses this library fairly extensively: https://github.com/jonboulle/clockwork

It provides a clock, which is either the real one (i.e. pass calls through to `time.Time`) or a fake one where you can control when/how it advances

milas (approved):
🏃

pkg/utils/set.go
```diff
@@ -0,0 +1,39 @@
...
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package utils
+
+type Set[T comparable] map[T]struct{}
```
glours, line 17: 😍

glours (approved):
LGTM

pkg/utils/set.go
ndeloof: one baby step...

Issue #1203, opened 2015-03-27
docker-compose dry run


ghost:
Would like insight to what docker-compose generates/executes

aanand:
Agreed - for the time being, you can use `docker-compose --verbose` to see what it's doing.

paurullan:
This would be incredibly useful for debugging purposes.

jmkgreen:
So `config` just said my file was fine yet `up` told me a service was linked to a service not described.

Does `config` need to be improved or is `dry-run` intended for this? Gut tells me the former...

dnephin:
`config` should catch these errors, we don't currently for `links` (however we do for `depends_on`).

I created #2933

ekristen:
+1 this would be a great feature

atrauzzi:
Something like this would be extremely helpful in debugging situations like https://github.com/docker/compose/issues/3259.

Currently `--verbose` isn't showing me any information about volumes/bind-mounts.

hervenicol:
I am also looking for a `dry-run` feature, that would not change anything but would explain whether it would perform any action in a normal run.

xandreklein:
+1

lwm:
+1

yaanst:
+1

dsaydon90:
+1

mooperd:
+1

emersonf:
Any news on this? It's been 3 years. At a minimum knowing which services are going to be restarted would allow automated notifications of downtime.

sauraahu:
+1

zethis:
+1

kuroop:
+1

dugajean:
This would be great to have.

gniewomir:
+1

ivanlmj:
+1

ehrlichja:
+1

tnwhitwell:
+1

afonsosantos:
+1 for AzuraCast @ https://azuracast.com

anhtuan84:
+1

DanielBryars:
+1 --dry-run would enable diagnostics, and also provide a way to validate it's going to do what you expect before a production change.

hinorashi:
I'm still here waiting for this dry run feature :D

pravynandas:
+1 in 2020.

wfehrnstrom:
+1

biscaldis:
+1

lweberru:
+1

[...]

PR #10173, opened 2023-01-13
Skeleton for dry-run under alpha command

glours:
**What I did**
Add a `dry-run` command skeleton under alpha command to test and  incrementally implement `--dry-run` flag

**Related issue**
https://docker.atlassian.net/browse/ENV-1

start of work for #1203 

image

### DIFF
--- a/pkg/api/dryrunclient.go
+++ b/pkg/api/dryrunclient.go
@@ -39,255 +39,14 @@ var _ client.APIClient = &DryRunClient{}
 
 // DryRunClient implements APIClient by delegating to implementation functions. This allows lazy init and per-method overrides
 type DryRunClient struct {
-	CopyFromContainerFn    func(ctx context.Context, container, srcPath string) (io.ReadCloser, moby.ContainerPathStat, error)
-	CopyToContainerFn      func(ctx context.Context, container, path string, content io.Reader, options moby.CopyToContainerOptions) error
-	ContainersPruneFn      func(ctx context.Context, pruneFilters filters.Args) (moby.ContainersPruneReport, error)
-	ConfigListFn           func(ctx context.Context, options moby.ConfigListOptions) ([]swarm.Config, error)
-	ConfigCreateFn         func(ctx context.Context, config swarm.ConfigSpec) (moby.ConfigCreateResponse, error)
-	ConfigRemoveFn         func(ctx context.Context, id string) error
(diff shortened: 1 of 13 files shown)


pkg/compose/compose.go
```diff
@@ -63,6 +66,25 @@ func (s *composeService) MaxConcurrency(i int) {
...
+		}
+		err = cli.Initialize(flags.NewClientOptions(), command.WithInitializeClient(func(cli *command.DockerCli) (client.APIClient, error) {
+			dryRunClient := api.NewDryRunClient()
+			dryRunClient.WithAPIClient(s.apiClient())
+			return dryRunClient, nil
+		}))
+		if err != nil {
+			return err
+		}
+		s.dockerCli = cli
```
ndeloof, line 83: could make it simpler by making API client an attribute in `composeService`, initialize with `dockerCli.Client()` and override here with `NewDryRunClient`.

pkg/api/dryrunclient.go
```diff
@@ -0,0 +1,1081 @@
...
+	"github.com/docker/docker/api/types/volume"
+	"github.com/docker/docker/client"
+	specs "github.com/opencontainers/image-spec/specs-go/v1"
+)
+
+var _ client.APIClient = &DryRunClient{}
+
+// DryRunClient implements APIClient by delegating to implementation functions. This allows lazy init and per-method overrides
+type DryRunClient struct {
+	CopyFromContainerFn    func(ctx context.Context, container, srcPath string) (io.ReadCloser, moby.ContainerPathStat, error)
```
ndeloof, line 42: As this is not a generic Client proxy, but dedicated to dry-run usage, we could make it simpler and just have all API methods to directly invoke `client.XX` from dockerCli's client _BUT_ the few ones where we want to bypass actual actions on docker engine.
OR rename this into generic `ClientProxy`, then we can configure dry-run by setting individual func to be overriden

glours: 👍

pkg/compose/compose.go
glours: Ok after digging a little bit, the `dockerCli` attribute is used widely by the CLI APIs, so it seams safer to override the APIClient of the existing `command.Cli`

ndeloof (approved):

milas (approved):

PR #10192, opened 2023-01-22
Fix Goroutine leak in v2/command/formatter

### LINKED ISSUE #10157: [BUG] goroutine leak in `init()` of `v2/cmd/formatter` ?
### Description

Trying to build something that depends on `github.com/docker/compose/v2`, then ran into this issue, because all of our tests checks for goroutine leaks at the end of running.

### Steps To Reproduce

Run following test

```go
package main

import (
	"testing"

	_ "github.com/docker/compose/v2/cmd/formatter"
	"go.uber.org/goleak"
)

func TestLeak(t *testing.T) {
	defer goleak.VerifyNone(t)
}
```

Output:

```shell
$ go test
--- FAIL: TestLeak (0.45s)
    main_test.go:12: found unexpected goroutines:
        [Goroutine 19 in state chan send, with github.com/docker/compose/v2/cmd/formatter.init.0.func1 on top of the stack:
        goroutine 19 [chan send]:
        github.com/docker/compose/v2/cmd/formatter.init.0.func1()
                /home/wxh/go/pkg/mod/github.com/docker/compose/v2@v2.14.2/cmd/formatter/colors.go:120 +0x1de
        created by github.com/docker/compose/v2/cmd/formatter.init.0
                /home/wxh/go/pkg/mod/github.com/docker/compose/v2@v2.14.2/cmd/formatter/colors.go:104 +0x8b
        ]
FAIL
exit status 1
FAIL    tt      0.451s
```

This is because there is a non-terminating goroutine in the `init()`:

```go
	go func() {
		i := 0
		rainbow := []colorFunc{
			colors["cyan"],
			colors["yellow"],
			colors["green"],
			colors["magenta"],
			colors["blue"],
			colors["intense_cyan"],
			colors["intense_yellow"],
			colors["intense_green"],
			colors["intense_magenta"],
			colors["intense_blue"],
		}

		for {
			loop <- rainbow[i]
			i = (i + 1) % len(rainbow)
		}
	}()
```

### Workaround

Not a best practice to bypass questionable func, but...

```go
	defer goleak.VerifyNone(t, goleak.IgnoreTopFunction("github.com/docker/compose/v2/cmd/formatter.init.0.func1"))
```

### Compose Version

```Text
https://github.com/docker/compose/blob/v2.14.2/cmd/formatter/colors.go
```

### Docker Environment

_No response_

### Anything else?

_No response_

AhmedGrati:
**What I did**
In this PR, I fixed the Goroutine leak in the `v2/command/formatter` init function. The problem that we had previously, is known as The Forgotten Sender.
To fix the issue, we should simply quit the infinite loop in the Goroutine, by sending a single that indicates that the main thread finished its execution. It's done using `defer()` and the `quitChan` channel.
A test of the Goroutine leak was added to the unit tests to make sure that the fix is working properly.

**Related issue**

Fixes #10157.

### DIFF
--- a/cmd/formatter/colors.go
+++ b/cmd/formatter/colors.go
@@ -99,6 +99,7 @@ func cleanInfiniteGoroutine() {
 }
 
 func init() {
+	defer cleanInfiniteGoroutine()
 	colors := map[string]colorFunc{}
 	for i, name := range names {
 		colors[name] = makeColorFunc(strconv.Itoa(30 + i))
@@ -131,5 +132,4 @@ func init() {
 		}
 	}()
 
-	defer cleanInfiniteGoroutine()
 }
--- a/cmd/formatter/colors.go
+++ b/cmd/formatter/colors.go
@@ -92,6 +92,11 @@ func rainbowColor() colorFunc {
 }
 
 var loop = make(chan colorFunc)
+var quitChan = make(chan bool)
+
+func cleanInfiniteGoroutine() {
+	quitChan <- true
+}
 
 func init() {
 	colors := map[string]colorFunc{}
@@ -116,8 +121,15 @@ func init() {
 		}
 
 		for {
-			loop <- rainbow[i]
-			i = (i + 1) % len(rainbow)
+			select {
+			case <-quitChan:
+				return
+			default:
+				loop <- rainbow[i]
+				i = (i + 1) % len(rainbow)
+			}
 		}
 	}()
+
+	defer cleanInfiniteGoroutine()
 }
--- a/cmd/formatter/formatter_test.go
+++ b/cmd/formatter/formatter_test.go
@@ -75,5 +75,5 @@ func TestPrint(t *testing.T) {
 
 // Test the absence of unexpected goroutines.
 func TestColorsGoroutinesLeak(t *testing.T) {
-	defer goleak.VerifyNone(t)
+	goleak.VerifyNone(t)
 }
--- a/cmd/formatter/formatter_test.go
+++ b/cmd/formatter/formatter_test.go
@@ -22,6 +22,7 @@ import (
 	"io"
 	"testing"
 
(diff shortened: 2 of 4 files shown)


ndeloof:
AFAICT, doing so after all rainbow colors have been used once, next call will block - which is not what we expect, first color should be re-used - so the loop

AhmedGrati:
Oh makes sense, didn't know about this constraint! I will think of a different solution.

AhmedGrati:
Done @ndeloof.

ndeloof (approved):

glours (approved):
LGTM

AhmedGrati:
Cannot reproduce the failed e2e tests.
```
go test  -race -coverprofile=coverage.out -covermode=atomic -count=1 ./pkg/e2e -run ^TestCopy$
ok      github.com/docker/compose/v2/pkg/e2e    5.467s  coverage: 39.9% of statements

go test  -race -coverprofile=coverage.out -covermode=atomic -count=1 ./pkg/e2e -run ^TestLocalComposeRun$
ok      github.com/docker/compose/v2/pkg/e2e    13.272s coverage: 40.5% of statements
```
The error in the pipeline is really weird.
```
Stderr:   unknown shorthand flag: 'f' in -f
```
Any help would be much appreciated.

ndeloof:
I had to revert this PR, has this breaks compose up use of color assignment

AhmedGrati:
@ndeloof Can you give me a scenario that helps me to reproduce the error, please?

ndeloof:
plain good old `docker compose up` was broken: as `defer` runs after `init` completion, the color loop hangs and we never get logs.
cleanup should take place after `pluginMain` execution

AhmedGrati:
Thanks @ndeloof, I'll fix it an open a new PR.

PR #10197, opened 2023-01-24
Revert "Fix Goroutine leak in v2/command/formatter"

ndeloof:
Reverts docker/compose#10192
this break compose up, as `defer cleanInfiniteGoroutine()` takes place immediately after `init` execution and prevent use of colors


glours (approved):
LGTM

Issue #10204, opened 2023-01-25
`docker-compose create --scale`


mhoangvslev:
### Description

Use case:
Have the service A and I want to eventually deploy 10 containers of that service. I can do `docker-compose up --scale A=10`but it starts all 10 containers and use up all my ressources. I want to be able to create the 10 containers, then activate one/several of them at any point in time.

mhoangvslev:
Thank you for the quick reaction and saving my day! When could I expect the release that includes this?

PR #10208, opened 2023-01-26
Add `--scale`  to `compose create`

laurazard:

**What I did**

Added `--scale` argument to `compose create`, moved scale logic from `cmd/compose/up.go` into `cmd/compose/create.go`.

**Related issue**

fixes https://github.com/docker/compose/issues/10204

image

### DIFF
--- a/cmd/compose/create.go
+++ b/cmd/compose/create.go
@@ -19,6 +19,8 @@ package compose
 import (
 	"context"
 	"fmt"
+	"strconv"
+	"strings"
 	"time"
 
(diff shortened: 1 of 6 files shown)


ndeloof (approved):

ndeloof:
LGTM.
Regarding https://github.com/docker/compose/issues/10204, I guess resource consumption happens on start, not create. Then, use of --parallel is an alternative to avoid starting all containers at once.

milas (approved):

Issue #10200, opened 2023-01-24
[BUG] `--wait` shouldn't have any effect when using a single service


leonheld:
### Description

Pretty much the title, if you use "--wait" with a single service, it hangs indefinitely. This makes sense: if you `--wait` and you only have one service, it'll hang forever because it won't receive a Healthy or Running from anyone else (I think...).

### Steps To Reproduce

Run `docker compose up --wait` and `docker compose up` on the provided `docker-compose.yml`. 
Adding a `restart always` or adding another service that will emit a Healthy or Running status clears the problem, hence my conjecture.

### Compose Version

```Text
Docker Compose version v2.14.1
```

### Docker Environment

```Text
➜  test docker info
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  app: Docker App (Docker Inc., v0.9.1-beta3)
  buildx: Docker Buildx (Docker Inc., v0.9.1-docker)
  compose: Docker Compose (Docker Inc., v2.14.1)
  scan: Docker Scan (Docker Inc., v0.23.0)

Server:
 Containers: 20
  Running: 1
  Paused: 0
  Stopped: 19
 Images: 107
 Server Version: 20.10.22
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: false
  userxattr: true
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 io.containerd.runtime.v1.linux runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 5b842e528e99d4d4c1686467debf2bd4b88ecd86
 runc version: v1.1.4-0-g5fd4c4d
 init version: de40ad0
 Security Options:
  seccomp
   Profile: default
  rootless
  cgroupns
 Kernel Version: 5.15.0-58-generic
 Operating System: Ubuntu 22.04.1 LTS
 OSType: linux
 Architecture: x86_64
 CPUs: 20
 Total Memory: 15.37GiB
 Name: toradex
 ID: R7MP:XSGO:BLYX:7HHW:OFHR:U3S4:B4AN:B4AL:54BV:KRXB:MKXX:3MSF
[...]

PR #10209, opened 2023-01-26
use containers we expect to start for wait condition

ndeloof:
as we check dependency service is running/healthy, use containers we created for the service so that we can detect failure to start, and not wait with no end for a dead service to become healthy.

**Related issue**
closes https://github.com/docker/compose/issues/10200

image

### DIFF
--- a/pkg/compose/convergence.go
+++ b/pkg/compose/convergence.go
@@ -28,7 +28,6 @@ import (
 	"github.com/containerd/containerd/platforms"
 	moby "github.com/docker/docker/api/types"
 	containerType "github.com/docker/docker/api/types/container"
-	"github.com/docker/docker/api/types/filters"
 	"github.com/docker/docker/api/types/network"
 	specs "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/pkg/errors"
@@ -281,7 +280,7 @@ func containerEvents(containers Containers, eventFunc func(string) progress.Even
 // ServiceConditionRunningOrHealthy is a service condition on status running or healthy
 const ServiceConditionRunningOrHealthy = "running_or_healthy"
 
-func (s *composeService) waitDependencies(ctx context.Context, project *types.Project, dependencies types.DependsOnConfig) error {
+func (s *composeService) waitDependencies(ctx context.Context, project *types.Project, dependencies types.DependsOnConfig, containers Containers) error {
 	eg, _ := errgroup.WithContext(ctx)
 	w := progress.ContextWriter(ctx)
 	for dep, config := range dependencies {
@@ -291,10 +290,7 @@ func (s *composeService) waitDependencies(ctx context.Context, project *types.Pr
 			continue
 		}
 
-		containers, err := s.getContainers(ctx, project.Name, oneOffExclude, false, dep)
-		if err != nil {
-			return err
-		}
+		containers = containers.filter(isService(dep))
 		w.Events(containerEvents(containers, progress.Waiting))
 
 		dep, config := dep, config
@@ -305,7 +301,7 @@ func (s *composeService) waitDependencies(ctx context.Context, project *types.Pr
 				<-ticker.C
 				switch config.Condition {
 				case ServiceConditionRunningOrHealthy:
-					healthy, err := s.isServiceHealthy(ctx, project, dep, true)
+					healthy, err := s.isServiceHealthy(ctx, containers, dep, true)
 					if err != nil {
 						return err
 					}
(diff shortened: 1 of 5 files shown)


rborn-tx:
Hello @ndeloof. I've run some tests on your branch and I've found some possible issues that I wanted to let you know. If I should instead report this as part of the original issue ticket, please let me know, okay?

### Test 1

```yaml
version: "2.4"
services:
  # Long-lived container.
  hello1:
    image: "alpine"
    command: ["sh", "-c", "echo 'Long-lived container, no health check'; sleep 1h;"]

  # Short-lived container.
  hello2:
    image: "alpine"
    command: ["echo", "Short-lived container, no health check"]
```

Execution:

```
$ compose --file docker-compose-mixed-nohc.yml up --detach --remove-orphans --wait
[+] Running 2/3
 ⠿ Network aktualizr-tests_default     Created                                                    0.0s
 ⠿ Container aktualizr-tests-hello2-1  Waiting                                                   15.8s
 ⠿ Container aktualizr-tests-hello1-1  Started                                                    0.4s
Killed
```
Hitting ^C did not stop docker-compose which had to be killed.


### Test 2

```yaml
version: "2.4"
services:
  # Short-lived container.
  hello:
    image: "alpine"
    command: ["echo", "Short-lived container, no health check"]
```

Execution:

```
$ compose --file docker-compose-shortlife-nohc.yml up --detach --remove-orphans --wait
[+] Running 1/2
 ⠿ Network aktualizr-tests_default    Created                                                    0.0s
 ⠿ Container aktualizr-tests-hello-1  Waiting                                                    0.8s
container aktualizr-tests-hello-1 exited (0)

$ echo $?
1
```

Notice in this case the `--wait` didn't cause docker-compose to wait forever, but the exit code of the program was non-zero indicating something went wrong.

### Test 3

```yaml
# Run multiple short-lived containers.
version: "2.4"
services:
  hello1:
    image: "alpine"
    command: ["echo", "Short-lived container (1), no health check"]
  hello2:
    image: "alpine"
    command: ["echo", "Short-lived container (2), no health check"]
  hello3:
    image: "alpine"
    command: ["echo", "Short-lived container (3), no health check"]
```

Execution:

```
$ compose --file docker-compose-shortlife-nohc3.yml up --detach --remove-orphans --wait
[+] Running 3/4
 ⠿ Network aktualizr-tests_default     Created                                                          0.0s
 ⠿ Container aktualizr-tests-hello1-1  Waiting                                                         26.9s
 ⠿ Container aktualizr-tests-hello2-1  Started                                                          0.5s
 ⠿ Container aktualizr-tests-hello3-1  Started                                                          0.6s
Killed
```

Like with Test 1, hitting ^C did not stop docker-compose which had to be killed.

ndeloof:
@rborn-tx thanks, nice feedback. I'll try to convert those examples into e2e tests so we have better coverage for this feature

ndeloof:
I fixed reported corner cases.
For second one ("Test 2") service command immediately complete, `--wait` detect container exited and report failure to become "running". This sounds a reasonable behavior as service is expected to run until interrupted, not to complete this way (otherwise, --wait makes no sense)

pkg/compose/convergence.go
```diff
@@ -291,10 +290,7 @@ func (s *composeService) waitDependencies(ctx context.Context, project *types.Pr
 			continue
 		}
 
-		containers, err := s.getContainers(ctx, project.Name, oneOffExclude, false, dep)
-		if err != nil {
-			return err
-		}
+		containers = containers.filter(isService(dep))
```
glours, line 293: Shouldn't we use a dedicated variable to manage filtered containers? I wondering if we won't loose potential dependencies by re-assigning `containers`

glours (changes requested):
Not sure, but we may have a potential side effect in the dependencies check loop

pkg/compose/convergence.go
ndeloof: indeed, nice catch - I've been lazy here

kaos14:
> I fixed reported corner cases. For second one ("Test 2") service command immediately complete, `--wait` detect container exited and report failure to become "running". This sounds a reasonable behavior as service is expected to run until interrupted, not to complete this way (otherwise, --wait makes no sense)

I disagree. I have multiple services in my docker-compose.yml, that I do want compose to wait for. At the same time I have a service that does its job and exits with 0 code. This change broke my setup and I had to add artificial sleep at the end of the script. I think `--wait` should take into account the `restart` option and not error out in Test 2.

kaos14:
@ndeloof What did you mean "I fixed reported corner cases."? None of them works in 2.16. Is this as expected?

PR #10214, opened 2023-01-30
rename `convert` to `config` to align with compose v1 UX

ndeloof:
**What I did**
renamed "convert" command to "config" to be aligned with Compose V1 UX
`convert` was introduced by Cloud integration development as we translated the compose model into cloud provider custom format. It was later adopted as the basis for `compose config` implementation, but changing the command canonical name doesn't make sense for our users (thousands blogs, stackoverflow, ... mention use of `config` command)


glours (approved):
LGTM

laurazard (approved):

milas (approved):

PR #10218, opened 2023-01-30
Tilt watcher

ndeloof:
**What I did**
imported watcher code from [tilt](https://github.com/tilt-dev/tilt/tree) to support new `compose watch` command

image


go.mod
```diff
@@ -33,6 +33,8 @@ require (
 	github.com/spf13/pflag v1.0.5
 	github.com/stretchr/testify v1.8.1
 	github.com/theupdateframework/notary v0.7.0
+	github.com/tilt-dev/fsevents v0.0.0-20200515134857-2efe37af20de
+	github.com/tilt-dev/fsnotify v1.4.8-0.20220602155310-fff9c274a375
```
thaJeztah, line 37: Do we know what patches are in this fork, and if they were rejected upstream? I know `fsnotify` has had some time where maintenance was slow, but I think it improved in that respect (and I _think_ we have some maintainers on it that are also maintainers for moby/moby)

pkg/watch/pattern_matcher.go
```diff
@@ -0,0 +1,255 @@
...
+
+// PatternMatcher allows checking paths against a list of patterns
+type PatternMatcher struct {
+	patterns   []*Pattern
+	exclusions bool
+}
+
+// NewPatternMatcher creates a new matcher object for specific patterns that can
+// be used later to match against patterns against paths
+func NewPatternMatcher(patterns []string) (*PatternMatcher, error) {
```
thaJeztah, line 58: This looks like a fork of (previously) `github.com/docker/docker/pkg/fileutils.NewPatternMatcher()`. That code was moved to a separate https://github.com/moby/patternmatcher repository (see https://github.com/moby/moby/pull/44226), so possibly this could all use that module (unless there's changes that are not (yet) in that code)

milas: I believe this was forked off primarily so it could use `tilt-dev/dockerignore` which had some fixes to handle edge cases around wildcard globs: https://github.com/moby/moby/issues/41433

That got confirmation it's fixed in HEAD and I just double checked that we're referencing a new enough version from our `go.mod` to get the fix

(There was also https://github.com/moby/moby/pull/42961 which has also been pulled in upstream by Moby)

---

tl;dr I'd use `moby/patternmatcher`!

pkg/compose/watch.go
```diff
@@ -17,96 +17,89 @@ package compose
...
 					return nil
-				case event := <-watcher.Events:
-					log.Println("fs event :", event.String())
-					needRefresh <- service.Name
-				case err := <-watcher.Errors:
+				case event := <-watcher.Events():
+					fmt.Fprintf(s.stderr(), "change detected on %s\n", event.Path())
+
+					for src, dest := range config.Sync {
+						if strings.HasPrefix(event.Path(), src) {
```
milas, line 92: For portability, I'd look at using `filepath.Rel` to determine this:
* It'll call `Clean()` on both paths so separators match (e.g. if you're on Windows but defined your ignores `/a/b/c`)
* It has some (probably imperfect but better than nothing?) case-insensitivity logic for Windows

pkg/compose/watch.go
```diff
@@ -17,96 +17,89 @@ package compose
...
+							needSync <- api.CopyOptions{
+								Source:      event.Path(),
+								Destination: fmt.Sprintf("%s:%s/%s", service.Name, dest, event.Path()[len(src):]),
```
milas, line 95: Ugh, definitely not in-scope for this PR, but we should change `api.CopyOptions` to use full fields for this and do the parsing in the CLI, right now we're doing that all in the backend @ https://github.com/docker/compose/blob/f24d3458c6d05e19519e660e4faa04d47f6db103/pkg/compose/cp.go#L274-L289

pkg/watch/dockerignore.go
```diff
@@ -0,0 +1,146 @@
...
+	return &dockerPathMatcher{
+		repoRoot: absRoot,
+		matcher:  pm,
+	}, nil
+}
+
+func readDockerignorePatterns(repoRoot string) ([]string, error) {
+	var excludes []string
+
+	f, err := os.Open(filepath.Join(repoRoot, ".dockerignore"))
```
milas, line 127: Aren't the `.dockerignore` loading rules more complex than this? (For example, if you have `foo.dockerfile`, won't it auto-use `foo.dockerignore`?) 😬

pkg/watch/notify.go
```diff
@@ -0,0 +1,106 @@
...
+		if err == nil {
+			return size
+		}
+	}
+	return defaultBufferSize
+}
+
+func IsWindowsShortReadError(err error) bool {
+	return runtime.GOOS == "windows" && err != nil && strings.Contains(err.Error(), "short read")
+}
```
milas, line 106: FYI for once we switch off the Tilt fork...this has been superseded by `ErrEventOverflow ` in HEAD @ `fsnotify/fsnotify`

pkg/watch/notify.go
```diff
@@ -0,0 +1,106 @@
...
+func (EmptyMatcher) Matches(f string) (bool, error)          { return false, nil }
+func (EmptyMatcher) MatchesEntireDir(f string) (bool, error) { return false, nil }
+
+var _ PathMatcher = EmptyMatcher{}
+
+func NewWatcher(paths []string, ignore PathMatcher) (Notify, error) {
+	return newWatcher(paths, ignore)
+}
+
+const WindowsBufferSizeEnvVar = "TILT_WATCH_WINDOWS_BUFFER_SIZE"
```
milas, line 89: This sorta ties into below re: short reads, but Tilt would detect these errors and then provide a message back to the user letting them know they should try and raise this limit.

In practice, I'm not sure how often it actually got hit (maybe @nicks knows), so it might be safe to drop configurability entirely here

Otherwise...`s/TILT/COMPOSE/` 😛

pkg/compose/watch.go
milas: Or perhaps `IsChild` from `pkg/watch/paths.go` which is used to filter out watches on descendants of recursively watched paths

milas (approved):

go.mod
thaJeztah: Ah, thanks! Yes, saw the comment later on, and saw that (github indicated "10" commits in the fork);

https://github.com/fsnotify/fsnotify/compare/main...tilt-dev:fsnotify:main

Pinning to a commit from upstream SGTM (short term). I'm mostly trying to avoid having 2 forks of the same dependency (as I know we have fsnotify as dependency in our tree already).

If someone has some cycles to spare to look what patches are not (yet) in upstream, we could contribute them there.

@cpuguy83 were you a maintainer on that repo? Or do I misremember that? (Otherwise I _think_ there's some familiar people on it, that we may try to reach out to to ask for a (pre-)release).

cpuguy83: I think I have commit access in order to get some important changes in, but I wouldn't call myself a "maintainer".

pkg/compose/watch.go
nicks: ya +1 for using IsChild for this kind of comparison.

the things to watch out for are:

- case insensitive file systems (e.g., /path/to/myrepo vs /path/to/MyRepo)
- symlinks, like the /tmp -> /private/tmp symlink

so we tried to avoid string-based functions for doing this kind of check, because the path string we used as input isn't always the same as the path string that the OS attaches to the event.

Lizz and I spent some time trying to pull out a "canonical" version of paths for string comparison
https://github.com/tilt-dev/tilt/pull/5309/files
but got distracted

pkg/compose/watch.go
ndeloof: this indeed is something I had in mind for a while ...

pkg/watch/dockerignore.go
ndeloof: I don't think so. But I might be wrong. I expect users to reming me with new issues if this is the case :)

go.mod
ndeloof: watcher_naive relies on `SetRecursive` which doesn't exist in upstream

PR #10236, opened 2023-02-03
detect replacement container is created and inform printer so it attach and don't stop [ENV-87]

ndeloof:
change a bit the "recreate" logic:
- create replacement container with a temporary name
- stop obsolete container
- remove obsolete container
- rename replacement container to canonical name
- start

this allows to set a "replace" label on replacement container, and to detect the replaced container should _not_ trigger printer to stop printing logs on exit.

# how to test:
- run `docker compose up`
- on a distinct terminal, force container to be recreated using `docker compose up --force-recreate -d`

```
$ docker compose up
[+] Running 1/0
 ⠿ Container truc-test-1  Created                                                                                                                                                                                        0.0s
Attaching to truc-test-1
truc-test-1  | hello docker
truc-test-1  | hello docker
truc-test-1  | hello docker
truc-test-1  | hello docker
truc-test-1  | hello docker
truc-test-1 exited with code 137
truc-test-1 has been recreated
truc-test-1  | hello docker
truc-test-1  | hello docker
truc-test-1  | hello docker
```

same applies to `docker compose logs -f`

image

### DIFF
--- a/pkg/api/labels.go
+++ b/pkg/api/labels.go
@@ -55,6 +55,8 @@ const (
 	VersionLabel = "com.docker.compose.version"
 	// ImageBuilderLabel stores the builder (classic or BuildKit) used to produce the image.
 	ImageBuilderLabel = "com.docker.compose.image.builder"
+	// ContainerReplaceLabel is set when container is created to replace another container (recreated)
+	ContainerReplaceLabel = "com.docker.compose.replace"
 )
 
 // ComposeVersion is the compose tool version as declared by label VersionLabel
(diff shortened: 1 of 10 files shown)


pkg/api/labels.go
```diff
@@ -55,6 +55,8 @@ const (
 	VersionLabel = "com.docker.compose.version"
 	// ImageBuilderLabel stores the builder (classic or BuildKit) used to produce the image.
 	ImageBuilderLabel = "com.docker.compose.image.builder"
+	// ContainerReplaceLabel is set when container is created to replace another container (recreated)
+	ContainerReplaceLabel = "com.docker.compose.replace"
```
milas, line 59: Ah this is clever and I like that it lets us reliably associate it with the old one by ID!

milas (approved):
Nice! This is nifty

PR #10243, opened 2023-02-06
FIX CVE-2021-41092

pyama86:
**What I did**
Hi!
`github.com/docker/cli` have a vulnerbility and then I fixed it.

ref:  https://avd.aquasec.com/nvd/cve-2021-41092

**Related issue**


glours:
👋 @pyama86 
This CVE was fixed in `docker/cli` `20.10.9` and our initial import target a `20.10.20` so Compose isn't concern with this vulnerability. In fact we have a replace rule for the `docker/cli` import and we're using a recent version of the `master` branch of the CLI.
Anyway we could decide to bump the version to `23.0.0` but we should also sync the `docker/docker` and the `moby/buildki` dependencies to be the 3 keys components are compatible

PR #10249, opened 2023-02-07
use CGO to enable fsevent on OSX [ENV-37]

ndeloof:
**What I did**
use CGO so we can rely on fsevent for `compose watch` on MacOS

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,12 +27,16 @@ ARG LICENSE_FILES=".*\(Dockerfile\|Makefile\|\.go\|\.hcl\|\.sh\)"
 # xx is a helper for cross-compilation
 FROM --platform=${BUILDPLATFORM} tonistiigi/xx:${XX_VERSION} AS xx
 
+# osxcross contains the MacOSX cross toolchain for xx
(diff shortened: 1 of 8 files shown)


Dockerfile
```diff
@@ -73,10 +77,12 @@ EOT
 FROM build-base AS build
 ARG BUILD_TAGS
 ARG TARGETPLATFORM
-RUN xx-go --wrap
 RUN --mount=type=bind,target=. \
     --mount=type=cache,target=/root/.cache \
     --mount=type=cache,target=/go/pkg/mod \
+    --mount=type=bind,from=osxcross,src=/osxsdk,target=/xx-sdk \
+    xx-go --wrap && \
+    if [ "$(xx-info os)" == "darwin" ]; then export CGO_ENABLED=1; fi && \
```
nicksieger, line 85: Maybe default CGO_ENABLED to 0 on the other platforms? Not sure if it gets set elsewhere.

```suggestion
    export CGO_ENABLED=0 && \
    if [ "$(xx-info os)" == "darwin" ]; then CGO_ENABLED=1; fi && \
```

Issue #10244, opened 2023-02-06
[BUG] Proxy settings defined in config and no build args in compose file cause build commands to fail


komima:
### Description

When a compose file defines an image to be built, no other build args are provided, build process requires the proxy variables and proxies are set in `config.json`, build fails:

```
panic: assignment to entry in nil map

goroutine 25 [running]:
github.com/docker/compose/v2/pkg/compose.(*composeService).toBuildOptions(_, _, {{0xc0004e0d50, 0x9}, {0x0, 0x0, 0x0}, 0xc000234d80, 0x0, {0x0, ...}, ...}, ...)
        github.com/docker/compose/v2/pkg/compose/build.go:236 +0x31c
github.com/docker/compose/v2/pkg/compose.(*composeService).build.func1({0x2174f20, 0xc0005fa140}, {0xc0004e0d50?, 0x9?})
        github.com/docker/compose/v2/pkg/compose/build.go:68 +0x2b0
github.com/docker/compose/v2/pkg/compose.(*graphTraversal).run.func1()
        github.com/docker/compose/v2/pkg/compose/dependencies.go:127 +0x63
golang.org/x/sync/errgroup.(*Group).Go.func1()
        golang.org/x/sync@v0.1.0/errgroup/errgroup.go:75 +0x64
created by golang.org/x/sync/errgroup.(*Group).Go
        golang.org/x/sync@v0.1.0/errgroup/errgroup.go:72 +0xa5
```

Found this issue when I was testing docker-compose with podman machine, and I needed the results of #10159 to get the proxy vars into the build automatically. I'm not so familiar with Go, but it seems that the [assignment](https://github.com/docker/compose/blob/b3ec110612f36fd358b1fbd6a3caf2647bce7f9a/pkg/compose/build.go#L236) fails due to [flatten result](https://github.com/docker/compose/blob/b3ec110612f36fd358b1fbd6a3caf2647bce7f9a/pkg/compose/build.go#L307) being nil, when there are no build args from the compose file?

### Steps To Reproduce

1. Be behind corporate proxy
2. Add proxies to `~/.docker/config.json`
3. Use a compose file with an image build and no build args
4. `docker compose build` fails with `panic: assignment to entry in nil map` 
5. Add any build arg to the compose file
6. `docker compose build` succeeds

### Compose Version

[...]

PR #10250, opened 2023-02-07
prevent assignment to entry in nil map

ndeloof:
**What I did**
have `flatten` always return non-nil `Mapping` to prevent risk for assignment to entry in nil map.

**Related issue**
closes https://github.com/docker/compose/issues/10244


glours (approved):
LGTM

PR #10252, opened 2023-02-08
support dry-run for exec command

glours:
**What I did**
Add support of dry run mode for `exec` command

**Related issue**
https://docker.atlassian.net/browse/ENV-55

image

### DIFF
--- a/pkg/api/dryrunclient.go
+++ b/pkg/api/dryrunclient.go
@@ -23,7 +23,9 @@ import (
 	"net"
 	"net/http"
 	"strings"
+	"sync"
 
+	"github.com/distribution/distribution/v3/uuid"
 	moby "github.com/docker/docker/api/types"
 	containerType "github.com/docker/docker/api/types/container"
 	"github.com/docker/docker/api/types/events"
@@ -35,6 +37,7 @@ import (
 	"github.com/docker/docker/api/types/volume"
 	"github.com/docker/docker/client"
 	specs "github.com/opencontainers/image-spec/specs-go/v1"
+	"github.com/pkg/errors"
 )
 
 const (
@@ -48,12 +51,19 @@ type DryRunKey struct{}
 // DryRunClient implements APIClient by delegating to implementation functions. This allows lazy init and per-method overrides
 type DryRunClient struct {
 	apiClient client.APIClient
+	execs     sync.Map
+}
+
+type execDetails struct {
+	container string
+	command   []string
 }
 
 // NewDryRunClient produces a DryRunClient
 func NewDryRunClient(apiClient client.APIClient) *DryRunClient {
 	return &DryRunClient{
 		apiClient: apiClient,
+		execs:     sync.Map{},
 	}
 }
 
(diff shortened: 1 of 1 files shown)


pkg/api/dryrunclient.go
```diff
@@ -155,6 +155,17 @@ func (d *DryRunClient) VolumeRemove(ctx context.Context, volumeID string, force
...
 }
 
+func (d *DryRunClient) ContainerExecCreate(ctx context.Context, container string, config moby.ExecConfig) (moby.IDResponse, error) {
+	fmt.Printf("%sCreating Exec configuration for container %s with command '%s'\n", DRYRUN_PREFIX, container, strings.Join(config.Cmd, " "))
+	config.Cmd = []string{"true"}
+	return d.apiClient.ContainerExecCreate(ctx, container, config)
+}
+
+func (d *DryRunClient) ContainerExecStart(ctx context.Context, execID string, config moby.ExecStartCheck) error {
+	fmt.Printf("%sExecuting command in detach mode\n", DRYRUN_PREFIX)
```
ndeloof, line 165: why "in detach(ed) mode" ?
when starting exec command, you can't tell if client did/will attach or not

glours: When I tested all the potential cases, I identified that `ContainerExecStart` was only called when the detached mode is activated
https://github.com/docker/cli/blob/24b4924410dc1db788ce1ee59e2cc7fa8459f8b0/cli/command/container/exec.go#L139

pkg/api/dryrunclient.go
```diff
@@ -155,6 +155,17 @@ func (d *DryRunClient) VolumeRemove(ctx context.Context, volumeID string, force
 	return ErrNotImplemented
 }
 
+func (d *DryRunClient) ContainerExecCreate(ctx context.Context, container string, config moby.ExecConfig) (moby.IDResponse, error) {
+	fmt.Printf("%sCreating Exec configuration for container %s with command '%s'\n", DRYRUN_PREFIX, container, strings.Join(config.Cmd, " "))
+	config.Cmd = []string{"true"}
+	return d.apiClient.ContainerExecCreate(ctx, container, config)
```
milas, line 161: Should this be `return nil`?

I know we aren't _starting_ the command here, but not sure of any other possible implications here

pkg/api/dryrunclient.go
milas: Do you think it's worth having something like `map[string]ExecDetails` on the `DryRunClient`?

That way, `ExecCreate` could be silent, but here we could log out `DRY-RUN Running command "rm -rf /" in app-svc-1`.

(Don't feel strongly about this, also seems fine as-is!)

milas (approved):

pkg/api/dryrunclient.go
ndeloof: `ContainerExecAttach` in the client indeed calls the `/start` engine endpoint with `Detach` set but this can be used to distinguish detached mode.

milas (approved):

PR #10262, opened 2023-02-09
ci: don't use `-race` on Windows

milas:
I misunderstood the cause of the symptom in #10261 - thought that we'd explicitly turned off CGO for Windows with some of the build changes recently, but we don't even have `gcc` on the CI node, so it's actually just `-race` entirely that's the trouble.

For right now, going the easy route and disabling it. We can look at getting a C toolchain on the Windows machine later.


glours (approved):
👍

ndeloof (approved):

Issue #10263, opened 2023-02-09
restart behavior when using network_mode: service


nicks:
**Repro Steps**

- Use this Compose file:
```
services:
  pause:
    image: busybox
    command: "sleep infinity"
    ports:
    - 8800:8800
  bba:
    network_mode: "service:pause"
    image: busybox
    command: "sh -c 'echo ok > index.html && busybox httpd -f -p 8800'"
    depends_on:
    - pause
```

-  `docker compose up -d`
- `curl localhost:8800` -> returns "ok"
- `docker compose restart pause`
- `curl localhost:8800`

**Expected Result**

Server returns "ok"

**Actual Result**

Unable to connect

**Additional context**

Sometimes I work with services that need to speak to each other on localhost. So I use Pod-like networking so that all my containers can reach each other. There's a pause container that sets up the network namespace, and the other compose services join to that pause container. That's what this compose file is implementing.

But suppose I want to expose a new port on the network. I add it to the pause service and run `docker compose restart pause`.

Oops! everything is now broken, because bba's network got disconnected from the pause container.

If I restart bba, the container works again.

I'm not sure what feature I'm actually asking for here. Maybe we should auto-restart 'bba' when the pause container restarts? Not sure.

[edit - rewrote the ticket to be a bit more structured]

nicks:
(this is on compose 2.15.1)

(also, it took me quite a while to figure out why my containers were randomly breaking, though it made sense once i figured out the repro case)

systemmonkey42:
Hi,

I realise this is an old PR, but I wondered about this

> - `restart`: when `true` a Compose implementation MUST restart service after dependent service has been restarted.

vs

> - `restart`: when `true` a Compose implementation MUST restart this service after updates to the dependency service.

If a dependent service restarts because of failure, exit, termination or "autoheal" as mentioned above, the service which depends on it, should also be restarted.

I notice it currently only applies after a manual docker-compose-restart or as the result of a configuration change.

In my case, the "depends_on" is accompanied by a "network_mode: service:<dependent_service>", which means the
container _MUST_ be restarted after the dependent container restarts or all network connectivity is lost.

Is there a way to express this requirement in the current implementation?

Thanks

ndeloof:
@systemmonkey42 compose only can manage restart triggered by an explicit command, (auto)restart after container failure is handled by docker engine, which doesn't have such a "dependencies" logic.

systemmonkey42:
[...]

PR #10273, opened 2023-02-13
also restart dependent services after a service has been restarted

ndeloof:
**What I did**
As a service is restarted, restart dependent services to ensure they can access the restarted container, which is required when those are set to share namespaces.

Introduce `--no-deps` flag for those who prefer the legacy behavior, still disabled by default to enforce a predictable restart as side-effects of unshared namespaces is hard to diagnose.

**Related issue**
close https://github.com/docker/compose/issues/10263

### DIFF
--- a/cmd/compose/restart.go
+++ b/cmd/compose/restart.go
@@ -28,6 +28,7 @@ import (
 type restartOptions struct {
 	*ProjectOptions
 	timeout int
+	noDeps  bool
 }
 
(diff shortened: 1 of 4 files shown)


laurazard (approved):

ndeloof:
Also created https://github.com/compose-spec/compose-spec/pull/305 so that user can explicitly opt-in/out for service to be restarted after a dependency has been restarted

PR #10275, opened 2023-02-13
exclude unstable labels from config hash

ndeloof:
**What I did**
since 2.16.0 we introduced a `replace` label to track relationship between new container and the one it replaces
this label is transient, and MUST not be considered for config hash computation

**Related issue**
close https://github.com/docker/compose/issues/9600


laurazard (approved):

benlangfeld:
This is a fix for a different bug than the one reported in #9600 . That issue should be re-opened.

Issue #10269, opened 2023-02-11
Add timeout to --wait option


mitasov-ra:
### Description

Assume I have the `docker-compose.yml` file with single service `foo` which guaranteed to fail immediately after I start it.

If I run 
```sh
docker-compose up --wait
```
the command will wait forever for Healthy status of `foo`. Literally forever.

I think the lack of `--wait-timeout` or just `--wait <timeout>` option renders this behaviour utterly useless.

Yes, I foresee one can suggest wrap  `docker-compose up --wait` in some code with timeout or use some command (like `timeout` or so), but why then we have `healthcheck` with all the `interval`, `timeout` and `retries` configurations?

It's really disappointing to see such a good feature being so poorly implemented.

P.S. Also, it seems to be a good idea not only implement timeout configuration but also add possibility to specify list of services to be waited for. And I think there's more possible enhancements.

P.P.S. On second thought, if `--wait` fails on any configured `healthcheck` failure then it already can be called very useful. But again, what if there's no configured `healthcheck`s?

laurazard:
Hiya @mitasov-ra, thanks for the report.

Regarding what you mentioned (Compose waiting forever when using `--wait` with a service that immediately fails), https://github.com/docker/compose/pull/10209 should address it. I believe fixing that should address your issues with the usability of `--wait`.

ndeloof:
This indeed has been fixed for container to immediately exit on start.
 `--wait` is designed to keep `up` command running until the whole stack is started & healthy. If none of the selected service has `healthcheck` configured, it just check container successfully started, which is a bare minimum sanity check.

jamshid:
Compose v2.17 breaks scripts using `docker-compose up --timeout N`. Is that intentional? 
I guess scripts should use `-t` to work with both 2.15 and 2.17. Now `-t` means --wait-timeout.

jamshid:
Thanks no problem it happens. And I was happy when I realized I could change scripts to -t for compatibility. Maybe a release note can be added about that. 

Hopefully you can rename back to --timeout (and no one uses --waitTimeout) before the next compose release in Docker for Desktop.

ndeloof:
Sure, I added a note on https://github.com/docker/compose/releases/tag/v2.17.2

mgulyas42:
Could you tell me this `- -wait-timeout` when start the timer? Does it start the timer immediately or does it wait for pulling docker images and then start?

PR #10276, opened 2023-02-13
introduce --wait-timeout on `up`

ndeloof:
**What I did**
Introduce --timeout to prevent `compose up` to be stuck

**Related issue**
closes https://github.com/docker/compose/issues/10269


dennisvang:
also related to #8777

PR #10278, opened 2023-02-14
sort service --hash output by service name

ndeloof:
**What I did**
make `compose service --hash` output predictable by sorting services by name


laurazard (approved):
LGTM

PR #10281, opened 2023-02-14
compact TUI to monitor layers download progress

ndeloof:
**What I did**
As terminal height might not allow to display per-layers download progress status, render a compact TUI using 8-dots braille chars as progress status

see demo on https://asciinema.org/a/4jZWUZfhcXouJPd1p6h0yMxtC

### DIFF
--- a/pkg/compose/push.go
+++ b/pkg/compose/push.go
@@ -139,11 +139,15 @@ func toPushProgressEvent(prefix string, jm jsonmessage.JSONMessage, w progress.W
 		return
 	}
 	var (
-		text   string
-		status = progress.Working
+		text    string
+		status  = progress.Working
+		total   int64
+		current int64
+		percent int
 	)
-	if jm.Status == "Pull complete" || jm.Status == "Already exists" {
+	if jm.Status == "Pushed" || jm.Status == "Already exists" {
 		status = progress.Done
+		percent = 100
 	}
 	if jm.Error != nil {
 		status = progress.Error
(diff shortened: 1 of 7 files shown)


laurazard (approved):
Nice!! Looks really good 🥳

williamdes:
It's so cute!!
can we have a percentage right after the text?

ndeloof:
this unfortunately breaks the progress alignment, need to double check

pkg/compose/push.go
```diff
@@ -139,10 +139,12 @@ func toPushProgressEvent(prefix string, jm jsonmessage.JSONMessage, w progress.W
...
 	var (
-		text   string
-		status = progress.Working
+		text    string
+		status  = progress.Working
+		total   int64
+		current int64
 	)
-	if jm.Status == "Pull complete" || jm.Status == "Already exists" {
+	if jm.Status == "Pushed" || jm.Status == "Already exists" {
```
williamdes, line 147: can this strings not use some sort of enum or const ?
(I don't know golang)

williamdes (commented):
I tested it with my projects that had the initial scolling bug of docker-compose
Using artifact from: https://github.com/docker/compose/actions/runs/4182986999

The new UI you propose is comprehensive and clean !
It's easy to understand

Very good work !

pkg/compose/push.go
ndeloof: unfortunately golang doesn't have enums
ideally, const should be defined by the Docker Engine API

williamdes:
let me know if the recent changes need that I retry
this is a good test project https://github.com/wdesportes/phpmyadmin-local-setup/blob/main/docker-compose.yml

StefanScherer (approved):
I tried it on Windows in PowerShell/CMD/Windows terminal and looks good there as well.

Issue #9961, opened 2022-10-27
Add support for additional build contexts


chrmarti:
### Description

`docker buildx build` added `--build-context` for additional build contexts. Would be great to have a way to use this with Compose.

26tanishabanik:
What is the status of work on this issue? @chrmarti

PR #10288, opened 2023-02-16
Support `--build-context` in `compose build`

laurazard:
**What I did**

Add `--build-context` to `compose build` to support passing additional build contexts.

**Related issue**

fixes https://github.com/docker/compose/issues/9961

image

### DIFF
--- a/cmd/compose/build.go
+++ b/cmd/compose/build.go
@@ -36,14 +36,15 @@ import (
 type buildOptions struct {
 	*ProjectOptions
 	composeOptions
-	quiet    bool
-	pull     bool
-	push     bool
-	progress string
-	args     []string
-	noCache  bool
-	memory   string
-	ssh      string
+	quiet         bool
+	pull          bool
+	push          bool
+	progress      string
+	args          []string
+	noCache       bool
+	memory        string
+	ssh           string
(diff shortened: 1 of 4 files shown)


ndeloof:
makes sense to me we introduce this in the compose file specification, but does it make sense this is also exposed as a command line flag?

laurazard:
Right. I was thinking we could do both, but I'm open if you think it shouldn't be exposed as a CLI arg. cc @milas @glours thoughts?

ndeloof:
Generally speaking (while there's a few exception for legacy reasons) compose core concept is to get all the `docker build` and `docker run` flags stored in compose.yaml, so that user don't have long command lines anymore. If contexts value is not supposed to change on each and every command execution, or can be set from paths relative to project directory, (yet another) flag is not necessary

laurazard:
Sounds good. I'm happy to close this and open a PR on spec to address https://github.com/compose-spec/compose-spec/issues/306 (also if you could PTAL @ that), and I'll take care of the rest of the changes.

laurazard:
Closing in favor of https://github.com/compose-spec/compose-spec/pull/307, adding this to the Specification.

PR #10320, opened 2023-02-27
test: tweak pause test to try and prevent failures in Windows CI

milas:
This test keeps failing with a timeout in Windows. I don't actually think it should take that long to bring up an nginx container, so I'm guessing that there's something else going on that's causing trouble.

Increase the verbosity when running Compose commands: I think this will generally make E2E test failures easier to diagnose by always logging the full command that's going to be run and also capturing stdout.

Add a health check and use `--wait` when launching the fixture for the pause test. Combined with the verbosity increase, this should make it easier to understand what's going on here.

### DIFF
--- a/pkg/e2e/fixtures/pause/compose.yaml
+++ b/pkg/e2e/fixtures/pause/compose.yaml
@@ -2,8 +2,16 @@ services:
   a:
     image:  nginx:alpine
     ports: [80]
+    healthcheck:
+      test: wget --spider -S -T1 http://localhost:80
+      interval: 1s
+      timeout: 1s
   b:
     image:  nginx:alpine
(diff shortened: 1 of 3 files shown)


glours (approved):

pkg/e2e/fixtures/pause/compose.yaml
```diff
@@ -2,8 +2,16 @@ services:
   a:
     image:  nginx:alpine
     ports: [80]
+    healthcheck:
+      test: wget --spider -S -T1 http://localhost:80
```
ndeloof, line 6: I didn't know about this `--spider` flag :)

ndeloof (approved):

PR #10325, opened 2023-02-27
Preserve symbol table in binaries for static analysis (`govulncheck`)

tianon:
While this stripping does decrease the binary size by some amount, it also removes the ability for `govulncheck` (https://go.dev/blog/vuln) to scan the binary for actual uses of vulnerable functions, requiring the user to clone the code locally and hope they're testing against the same version of the stdlib, etc that the binary was built with.  If we stop passing `-s`, then we can then run `govulncheck` on the binary directly (making it easier to flag both false positives in CVE scans _and_ actual issues worth looking into).

Here's an example of the output on a freshly built binary with this change:

```console
$ govulncheck ./bin/build/docker-compose
govulncheck is an experimental tool. Share feedback at https://go.dev/s/govulncheck-feedback.

Using govulncheck@v0.0.0 with
vulnerability data from https://vuln.go.dev (last modified 27 Feb 23 16:29 UTC).

Scanning your binary for known vulnerabilities...
No vulnerabilities found.
```

Compared to the 1.16.0 release binary:

```console
$ govulncheck ./docker-compose
govulncheck is an experimental tool. Share feedback at https://go.dev/s/govulncheck-feedback.

Using govulncheck@v0.0.0 with
vulnerability data from https://vuln.go.dev (last modified 27 Feb 23 16:29 UTC).

Scanning your binary for known vulnerabilities...
govulncheck: vulncheck.Binary: reading go:func.*: no symbol "go:func.*"
```

It's not 100% apples-to-apples, but the size difference between these binaries is ~46MiB for the 1.16.0 release and ~52MiB for the binary I built from this commit.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,7 @@
 PKG := github.com/docker/compose/v2
 VERSION ?= $(shell git describe --match 'v[0-9]*' --dirty='.m' --always --tags)
 
-GO_LDFLAGS ?= -s -w -X ${PKG}/internal.Version=${VERSION}
+GO_LDFLAGS ?= -w -X ${PKG}/internal.Version=${VERSION}
 GO_BUILDTAGS ?= e2e
 
 ifeq ($(OS),Windows_NT)


ndeloof (commented):
I don't have a strong opinion on this topic as I don't know the actual impacts doing so

milas (approved):
LGTM - here's the full size comparisons across architectures from HEAD of v2

macOS is basically identical, Linux/Windows gain ~5-6MB as you noted

I'm ok with this as we're still removing the DWARF debug info (`-w`) which is much bigger [some of the binaries would be >80MB with that 😬]

(That said, I think we're probably due for a dependency review to try and shrink the binary some.)

STRIPPED
========
```
❯ du -k bin/build/*
53632	bin/build/darwin_amd64
51264	bin/build/darwin_arm64
47168	bin/build/linux_amd64
45120	bin/build/linux_arm64
44096	bin/build/linux_arm_v6
44096	bin/build/linux_arm_v7
46144	bin/build/linux_ppc64le
45120	bin/build/linux_riscv64
50240	bin/build/linux_s390x
48192	bin/build/windows_amd64
45120	bin/build/windows_arm64
```


UNSTRIPPED
==========
```
❯ du -k bin/build/*
53312	bin/build/darwin_amd64
51808	bin/build/darwin_arm64
53312	bin/build/linux_amd64
51264	bin/build/linux_arm64
50240	bin/build/linux_arm_v6
50240	bin/build/linux_arm_v7
52288	bin/build/linux_ppc64le
52288	bin/build/linux_riscv64
56384	bin/build/linux_s390x
53312	bin/build/windows_amd64
51264	bin/build/windows_arm64
```

laurazard (approved):

Issue #10313, opened 2023-02-25
[BUG] Can't build docker-compose.yml


Eronary:
### Description

When I run the `docker-compose up --build` command, I get an error: `Error response from daemon: no such image: -app: invalid reference format`

Later I realized that this error disappears when I remove `build` from the `app` service, for example if I replace the `build` with any `image` (just for testing purposes) everything works

### Steps To Reproduce

1. Run `docker-compose up --build` in this directory 
![directory](https://user-images.githubusercontent.com/63160741/221379085-423cca05-ec4e-4569-8819-0591c094926a.png)
2. An error occurs immediately
![9wU7Url](https://user-images.githubusercontent.com/63160741/221379140-04a01e44-2a18-4e3e-a7ba-ba2a40256aa1.png)

### Compose Version

```Text
Docker Compose version v2.15.1
```

### Docker Environment

```Text
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc., v0.10.0)
  compose: Docker Compose (Docker Inc., v2.15.1)
  dev: Docker Dev Environments (Docker Inc., v0.0.5)
  extension: Manages Docker extensions (Docker Inc., v0.2.17)
  sbom: View the packaged-based Software Bill Of Materials (SBOM) for an image (Anchore Inc., 0.6.0)
  scan: Docker Scan (Docker Inc., v0.23.0)

Server:
 Containers: 3
  Running: 0
  Paused: 0
  Stopped: 3
 Images: 3
 Server Version: 20.10.22
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 1
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 io.containerd.runtime.v1.linux runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 9ba4b250366a5ddde94bb7c9d1def331423aa323
 runc version: v1.1.4-0-g5fd4c4d
 init version: de40ad0
 Security Options:
  seccomp
   Profile: default
 Kernel Version: 5.10.16.3-microsoft-standard-WSL2
[...]

PR #10334, opened 2023-03-01
Report error if project name is empty after normalization

ndeloof:
**What I did**
Check project name is not empty after options have been applied.
This can happen as name is computed from parent directory

**Related issue**
see https://github.com/docker/compose/issues/10313

### DIFF
--- a/cmd/compose/compose.go
+++ b/cmd/compose/compose.go
@@ -189,6 +189,10 @@ func (o *ProjectOptions) ToProject(services []string, po ...cli.ProjectOptionsFn
 		return nil, compose.WrapComposeError(err)
 	}
 
+	if project.Name == "" {
+		return nil, errors.New("project name can't be empty. Use `--project-name` to set a valid name")
+	}
+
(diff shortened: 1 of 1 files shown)


cmd/compose/compose.go
```diff
@@ -189,6 +189,11 @@ func (o *ProjectOptions) ToProject(services []string, po ...cli.ProjectOptionsFn
 		return nil, compose.WrapComposeError(err)
 	}
 
+	if project.Name == "" {
+		return nil, errors.New("project name can't be empty. Use `--project-name` to set a valid nameZZZZ" +
```
glours, line 193: That's a lot of `ZZZZ`, no? 😄

glours (approved):
Remove unnecessary `Z`s before merging

cmd/compose/compose.go
ndeloof: 😓🤨

laurazard (approved):

Issue #8713, opened 2021-09-30
.env file referenced underneath env_file in a compose file is being parsed when file is not used.


beyondmeat:
**Description**

In Compose V2 -2.0.0.0, .env files that are underneath other folders are being picked up and parsed by compose. In my case these .env files are stored underneath compose-stack/.config/.myapp/.env and don't have anything to do with compose.

**Steps to reproduce the issue:**
1. use compose file (mine uses nested compose file and profiles)
2. make .env file in a deep nested folder
3. run compose up

**Describe the results you received:**

```
unexpected character "/" in variable name near "rclone/drive\nRCLONE_LOG_LEVEL=ERROR\nRCLONE_LOG_FILE=/home/user/drive.log\n\nRCLONE_DRIVE_PACER_MIN_SLEEP=10ms\nRCLONE_DRIVE_PACER_BURST=1000\nRCLONE_BWLIMIT_FILE=100M"
```

**Describe the results you expected:**
Not to include the .env file since it's not in the top working dir or in the same folder as the compose file.

**Additional information you deem important (e.g. issue happens only occasionally):**

I could rename the non-compose .env files but that feels like a workaround.

**Output of `docker compose version`:**

```
Docker Compose version v2.0.0
```

**Output of `docker info`:**

```
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  app: Docker App (Docker Inc., v0.9.1-beta3)
  buildx: Build with BuildKit (Docker Inc., v0.6.1-docker)
  compose: Docker Compose (Docker Inc., v2.0.0)
  scan: Docker Scan (Docker Inc., v0.8.0)

Server:
 Containers: 16
  Running: 16
  Paused: 0
  Stopped: 0
 Images: 35
 Server Version: 20.10.8
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 1
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: runc io.containerd.runc.v2 io.containerd.runtime.v1.linux
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: e25210fe30a0a703442421b0f60afac609f950a3
[...]

PR #10335, opened 2023-03-01
update compose-go

ndeloof:
**What I did**
bump compose-go 
- use updated `ForServices` with options to implement `--no-deps` flag
- pass selected profiles to loader to avoid loading disabled services environment

#closes https://github.com/docker/compose/issues/8713


glours (approved):
LGTM

PR #10341, opened 2023-03-02
add dry-run support for pull command

glours:
**What I did**
Add dry run support to `pull` command

**Related issue**
https://docker.atlassian.net/browse/ENV-62

image

### DIFF
--- a/pkg/compose/pull.go
+++ b/pkg/compose/pull.go
@@ -130,6 +130,13 @@ func (s *composeService) pull(ctx context.Context, project *types.Project, opts 
 					mustBuild = append(mustBuild, service.Name)
 				}
 				if !opts.IgnoreFailures && service.Build == nil {
+					if s.dryRun {
+						w.Event(progress.Event{
+							ID:     service.Name,
+							Status: progress.Error,
+							Text:   fmt.Sprintf(" - Pull error for image: %s", service.Image),
+						})
+					}
 					// fail fast if image can't be pulled nor built
 					return err
 				}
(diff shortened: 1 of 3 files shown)


ndeloof:
I wonder: should dry-run on pull check the image exists and is accessible with configured credentials, but not actually pull, or just do nothing as proposed here ?

glours:
Right! I'll add the check of the image existence and that the user is granted to pull it

ndeloof (approved):

pkg/compose/pull.go
```diff
@@ -130,6 +130,13 @@ func (s *composeService) pull(ctx context.Context, project *types.Project, opts
 					mustBuild = append(mustBuild, service.Name)
 				}
 				if !opts.IgnoreFailures && service.Build == nil {
+					if s.dryRun {
+						w.Event(progress.Event{
+							ID:     service.Name,
+							Status: progress.Error,
+							Text:   fmt.Sprintf(" - Pull error for image: %s", service.Image),
```
milas, line 137: Should we also show the contents of `err`?

(99% of the time I think it'll be everyone's favorite: `pull access denied, repository does not exist or may require authorization: server message: insufficient_scope: authorization failed`)

milas (approved):
LGTM

pkg/compose/pull.go
milas: Oh it returns the `err` and gets logged after 🤦🏻

PR #10350, opened 2023-03-06
update readme file images

### LINKED ISSUE #10349: Badges are not consistent
Modify the badge which is displaying the URL to the issue related the shields.io repo but not the message.

hemanthkotaprolu:
**Fixed a broken broken badge URL and aligned images to look nice**

These are the changes i made:
- Modified the badge which is displaying the URL to the issue related the shields.io repo. 
- Positioned the badges to make them look nice
- Aligned the image to the cener to look nice.

I am not sure what to send as the message so i sent the the message as action, since on clicking it is opening the actions tab in the compose repo. Please let me know if there is any change and can close #10349

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,15 +8,15 @@
 - [Contributing](#contributing)
 # Docker Compose v2
 <p align="center">
-    <a href="https://github.com/docker/compose/releases/latest"><img src="https://img.shields.io/github/release/docker/compose.svg?style=flat-square"</a>
-    <a href="https://pkg.go.dev/github.com/docker/compose/v2"><img src="https://img.shields.io/badge/go.dev-docs-007d9c?style=flat-square&logo=go&logoColor=white"</a>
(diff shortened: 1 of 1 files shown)


ndeloof:
I'd prefer we avoid using html in markdown files.
I wonder we could just introduce a line break before logo

hemanthkotaprolu:
- Please take a look at the picture. Markdown doesn't provide a way to style so this way by using html it is looking nice.
- We can add a horizontal line before logo as well. Let me add it as well.

docker_compose

ndeloof:
> Markdown doesn't provide a way to style

yes, this is the intent: it is not designed to offer a candy-looking web site, but offer some readable documentation while keeping the source easy to edit and maintain.

ndeloof:
Closing this as "won't fix" as README is not intended to be more than a plain markdown file

Issue #10354, opened 2023-03-08
[BUG] docker-compose run still creates containers with an underscore in the name


jamshid:
### Description

`docker compose run` creates a container name with underscores instead of dashes

### Steps To Reproduce

1. Using this docker-compose.yml in a directory named `detest`:
```
services:
  foo:
    image: centos
```

2. Running `docker compose up -d` correctly creates a container named `dctest-foo-1`. 

3. But `docker compose run -d foo` creates a container named `dctest_foo_run_e556189032da`.

I expected it to be named `dctest-foo-run-e556189032da` to avoid the problem where underscores are not a legal hostname.

### Compose Version

```Text
Docker Compose version v2.15.1
```

### Docker Environment

```Text
jafshar@JAFSHAR-MB1-17 dctest % docker info
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc., v0.10.3)
  compose: Docker Compose (Docker Inc., v2.15.1)
  dev: Docker Dev Environments (Docker Inc., v0.1.0)
  extension: Manages Docker extensions (Docker Inc., v0.2.18)
  sbom: View the packaged-based Software Bill Of Materials (SBOM) for an image (Anchore Inc., 0.6.0)
  scan: Docker Scan (Docker Inc., v0.25.0)
  scout: Command line tool for Docker Scout (Docker Inc., v0.6.0)
```

### Anything else?

This is a another inconsistency similar to the network and volume names in https://github.com/docker/compose/issues/9618 but it breaks the hostname so it's not just cosmetic. 

*Well actually the network name with an underscore is bad too, that is part of the fully qualified hostname. But that can be worked around by specifying the network name in the compose yml.

PR #10357, opened 2023-03-08
one off container name use configured Separator for naming

ndeloof:
**What I did**
use configured name separator to define oneoff container name

**Related issue**
closes https://github.com/docker/compose/issues/10354


glours (approved):
LGTM

PR #10371, opened 2023-03-14
build and push binaries images when a PR is merged or a tag pushed

glours:
**What I did**
Generate new binary images when a PR is merged into main branch or a tag is pushed, those images would be pushed to `docker/compose` Hub repo

**Related issue**
https://docker.atlassian.net/browse/ENV-119

image

### DIFF
--- a/.github/workflows/merge.yml
+++ b/.github/workflows/merge.yml
@@ -10,10 +10,12 @@ on:
       - 'v2'
     tags:
       - 'v*'
-  workflow_dispatch:
 
 permissions:
   contents: read # to fetch code (actions/checkout)
+
+env:
+  REPO_SLUG: "docker/compose-bin"
 
 jobs:
   e2e:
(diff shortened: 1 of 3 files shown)


.github/workflows/merge.yml
```diff
@@ -15,6 +15,11 @@ on:
 permissions:
   contents: read # to fetch code (actions/checkout)
 
+env:
+  REPO_SLUG: "docker/compose"
```
crazy-max, line 19:
```suggestion
  REPO_SLUG: "docker/compose-bin"
```

https://github.com/docker/compose/pull/9787#issuecomment-1226032270

.github/workflows/merge.yml
```diff
@@ -15,6 +15,11 @@ on:
...
+  BUILDX_VERSION: "latest"
+  BUILDKIT_IMAGE: "moby/buildkit:latest"
```
crazy-max, line 21: Don't think this is needed for compose. Just sticking to default ones should be enough.

.github/workflows/merge.yml
```diff
@@ -72,3 +77,51 @@ jobs:
...
+      -
+        name: Set up QEMU
+        uses: docker/setup-qemu-action@v2
+      -
+        name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+        with:
+          version: ${{ env.BUILDX_VERSION }}
+          driver-opts: image=${{ env.BUILDKIT_IMAGE }}
+          buildkitd-flags: --debug
```
crazy-max, line 95: Can be removed (see my previous comment)

.github/workflows/merge.yml
```diff
@@ -72,3 +77,51 @@ jobs:
...
+            type=ref,event=tag
+            type=edge
+          bake-target: meta-helper
+      -
+        name: Login to DockerHub
+        if: github.event_name != 'pull_request'
+        uses: docker/login-action@v2
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
```
crazy-max, line 113: We should use the org secrets put in place by @technicallyjosh 

Might need to create first `docker/compose-bin` repo and then add the bot user to it.

.github/workflows/merge.yml
glours: We imagined to re-use the existing `docker/compose` repository to store the new images as the users should now use the docker official image to run Compose in a container

crazy-max: I think it should be another repo. https://hub.docker.com/r/docker/compose is already a runnable image. It will be confusing for users imo.

Best would be to deprecate https://hub.docker.com/r/docker/compose and update hub repo description that would tell to switch to the official docker image with embedded compose or use the `compose-bin` one if user wants to use its own base image like:

```dockerfile
FROM debian
COPY --from=docker/compose-bin /docker-compose /usr/bin/compose
RUN compose version
```

(as compose can be standalone too)

glours: We'll use a dedicated repo ` https://hub.docker.com/r/docker/compose-bin` for the new images 👍

crazy-max (approved):
LGTM

ndeloof (approved):

PR #10372, opened 2023-03-14
collect built image IDs

ndeloof:
**What I did**
Fix collection of built image IDs. 

We used a map[string]string to be [overridden by subsequent calls to `s.doBuild`](https://github.com/docker/compose/blob/v2/pkg/compose/build.go#L101), so loosing earlier results (wonder this didn't resulted in some bugs...). 

Replaced by a pre-allocated `[]string` using service index from `project.Services` to record IDs collected by goroutines at a known position.


ndeloof:
I assume this used to work because buildx does not only return the built IDs for last call, but all the built IDs for the active session.

milas (approved):
Nice fix!

Issue #10056, opened 2022-12-08
[BUG] Build container isolation ignored


jhlgns:
### Description

I would like to build an windows-platform image that requires more than 1GB of RAM to build.
`docker build --isolation=process .` works, but setting `<service>.build.isolation: process` does not.

### Steps To Reproduce

Dockerfile:
```Dockerfile
FROM mcr.microsoft.com/windows/servercore:ltsc2022
SHELL ["powershell", "-Command", "$ErrorActionPreference = 'Stop'; $ProgressPreference = 'SilentlyContinue';"]
RUN Get-WmiObject Win32_PhysicalMemory |% Capacity
```
```powershell
docker build --no-cache .
```
Shows 1GB of memory (Hyper-V isolation default) - so like this it's not possible to build the image I want to build
```powershell
docker build --no-cache --process=isolation .
```
Shows 32GB of memory - so it's possible to build the image

Using the following docker-compose.yml:
```yml
version: '3'

services:
  cowabunga:
    build:
      context: .
```
```powershell
docker-compose build --no-cache
```
Shows 1GB of memory
Setting `<service>.build.isolation:`
```yml
version: '3'

services:
  cowabunga:
    build:
      context: .
      isolation: process
```
```powershell
docker-compose build --no-cache
```
**Still shows 1GB - impossible to build the image**

NOTE: The Docker daemon is configured for the windows platform!

### Compose Version

```Text
Both are `v2.12.2`
```

### Docker Environment

```Text
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc., v0.9.1)
  compose: Docker Compose (Docker Inc., v2.12.2)
  dev: Docker Dev Environments (Docker Inc., v0.0.3)
  extension: Manages Docker extensions (Docker Inc., v0.2.13)
  sbom: View the packaged-based Software Bill Of Materials (SBOM) for an image (Anchore Inc., 0.6.0)
  scan: Docker Scan (Docker Inc., v0.21.0)

Server:
 Containers: 65
  Running: 13
  Paused: 0
  Stopped: 52
 Images: 638
 Server Version: 20.10.21
 Storage Driver: windowsfilter
  Windows:
 Logging Driver: json-file
 Plugins:
  Volume: local
  Network: ics internal l2bridge l2tunnel nat null overlay private transparent
  Log: awslogs etwlogs fluentd gcplogs gelf json-file local logentries splunk syslog
 Swarm: inactive
[...]

PR #10374, opened 2023-03-15
use `build` as common API for build scenarios

ndeloof:
**What I did**
from various places we trigger builds (`build`,  `up` and `watch` commands), use `composeService#build` as the single entry point into the (complex) build logic. 

This allows to manage classic vs buildkit at early stage, and to pass the BuildConfig to classic builder, while we only get it converted into `buildx.Options` for buildkit builder

To get `up` to only build selected platform, `prepareProjectForBuild` mutate the `project` to remove unnecessary build config so that the exact same `build` function can be used both by `up` and `build`.
Added a test case to make it obvious about this mutation and prevent any regression next time we try to change this codebase :)

**Related issue**
see https://github.com/docker/compose/issues/10056 for illustration of the issue using `buildx.Options` to configure classic builder

image

### DIFF
--- a/pkg/compose/build.go
+++ b/pkg/compose/build.go
@@ -43,6 +43,10 @@ import (
 )
 
 func (s *composeService) Build(ctx context.Context, project *types.Project, options api.BuildOptions) error {
+	err := options.Apply(project)
+	if err != nil {
+		return err
+	}
 	return progress.Run(ctx, func(ctx context.Context) error {
 		_, err := s.build(ctx, project, options)
 		return err
@@ -50,10 +54,14 @@ func (s *composeService) Build(ctx context.Context, project *types.Project, opti
 }
 
 func (s *composeService) build(ctx context.Context, project *types.Project, options api.BuildOptions) (map[string]string, error) {
-	args := flatten(options.Args.Resolve(envResolver(project.Environment)))
+	args := options.Args.Resolve(envResolver(project.Environment))
 
+	buildkitEnabled, err := s.dockerCli.BuildKitEnabled()
+	if err != nil {
+		return nil, err
+	}
 	builtIDs := make([]string, len(project.Services))
-	err := InDependencyOrder(ctx, project, func(ctx context.Context, name string) error {
+	err = InDependencyOrder(ctx, project, func(ctx context.Context, name string) error {
 		if len(options.Services) > 0 && !utils.Contains(options.Services, name) {
 			return nil
 		}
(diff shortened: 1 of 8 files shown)


pkg/compose/build.go
```diff
@@ -61,55 +65,37 @@ func (s *composeService) build(ctx context.Context, project *types.Project, opti
...
+
 			if service.Build == nil {
 				return nil
 			}
-			imageName := api.GetImageNameOrDefault(service, project.Name)
-			buildOptions, err := s.toBuildOptions(project, service, imageName, options.SSHs)
-			if err != nil {
-				return err
+
+			if buildkitEnabled, err := s.dockerCli.BuildKitEnabled(); err != nil || !buildkitEnabled {
```
milas, line 73: Minor: it might be good to bring this out of the "loop" (`InDependencyOrder`)

Also, I think it's probably safe to fail and propagate the error:
https://github.com/docker/cli/blob/3c9e0073dd3ab31783a9d27d5fc26b70d59db367/cli/command/cli.go#L171 (tl;dr it can only fail if you do something like `DOCKER_BUILDKIT=not_a_bool`

Or, at the very least, a warning that we're falling back would be nice.

pkg/compose/build.go
```diff
@@ -172,37 +161,32 @@ func (s *composeService) ensureImagesExists(ctx context.Context, project *types.
 	return nil
 }
 
-func (s *composeService) getBuildOptions(project *types.Project, images map[string]string) (map[string]build.Options, error) {
-	opts := map[string]build.Options{}
-	for _, service := range project.Services {
-		if service.Image == "" && service.Build == nil {
-			return nil, fmt.Errorf("invalid service %q. Must specify either image or build", service.Name)
+func (s *composeService) prepareProjectForBuild(project *types.Project, images map[string]string) error {
+	err := api.BuildOptions{}.Apply(project)
```
milas, line 165: Do we need to actually pass the `api.BuildOptions` down?

In `Apply` we use `Pull` / `NoCache`, which I assume are for the global CLI `--pull` & `--no-cache` flags respectively

pkg/compose/build_classic.go
```diff
@@ -252,25 +211,18 @@ func isLocalDir(c string) bool {
 	return err == nil
 }
 
-func imageBuildOptions(options buildx.Options) dockertypes.ImageBuildOptions {
+func imageBuildOptions(config *types.BuildConfig) dockertypes.ImageBuildOptions {
```
milas, line 214: Perhaps rename `classicBuildOptions()`

pkg/compose/build_test.go
```diff
@@ -0,0 +1,121 @@
...
+package compose
+
+import (
+	"testing"
+
+	"github.com/compose-spec/compose-go/types"
+	"gotest.tools/v3/assert"
+)
+
+func TestPrepareProjectForBuild(t *testing.T) {
```
milas, line 26: 💯 for these tests! We definitely need more unit test coverage for some of these complex scenarios like platforms, so this is awesome

milas (approved):
✨ 🧹 👍🏻

glours (approved):
LGTM

pkg/compose/build.go
ndeloof: we only expose those flags for `build` command. `up --build` only applies options set by compose file

pkg/compose/build.go
```diff
@@ -298,20 +271,37 @@ func (s *composeService) toBuildOptions(project *types.Project, service types.Se
...
 			DockerfileInline: service.Build.DockerfileInline,
 			DockerfilePath:   dockerFilePath(service.Build.Context, service.Build.Dockerfile),
+			NamedContexts:    toBuildContexts(service.Build.AdditionalContexts),
 		},
 		CacheFrom:   cacheFrom,
 		CacheTo:     cacheTo,
-		NoCache:     service.Build.NoCache,
-		Pull:        service.Build.Pull,
+		NoCache:     service.Build.NoCache || options.NoCache,
+		Pull:        service.Build.Pull || options.Pull,
```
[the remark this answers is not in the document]
ndeloof: yes indeed, reminiscence for a previous attempts, before I moved it all to `Apply` :)

pkg/compose/build_classic.go
ndeloof: this produces the `options` for the `ImageBuild` engine API, so ...

PR #10386, opened 2023-03-20
watch: add file delete/rename handling

milas:
**What I did**
Add support for file deletes & renames while running `alpha watch`.

This approach mimics Tilt's behavior[^1]:
 1. At sync time, `stat` the path on host
 2. If the path does not exist -> `rm` from container
 3. If the path exists -> sync to container

By handling things this way, we're always syncing based on the true state, regardless of what's happened in the interim. For example, a common pattern in POSIX tools is to create a file and then rename it over an existing file. Based on timing, this could be a sync, delete, sync (every file gets seen & processed) OR a delete, sync (by the the time we process the event, the "temp" file is already gone, so we just delete it from the container, where it never existed, but that's fine since we deletes are idempotent thanks to the `-f` flag on `rm`).

Additionally, when syncing, if the `stat` call shows it's for a directory, we ignore it. Otherwise, duplicate, nested copies of the entire path could get synced in. (On some OSes, an event for the directory gets dispatched when a file inside of it is modified. In practice, I think we might want this pushed further down in the watching code, but since we're already `stat`ing the paths here now, it's a good place to handle it.)

Lastly, there's some very light changes to the text when it does a full rebuild that will list out the (merged) set of paths that triggered it. We can continue to improve the output, but this is really helpful for understanding why it's rebuilding.

**Related issue**
JIRA: ENV-139

two cats sleeping in a bed

[^1]: https://github.com/tilt-dev/tilt/blob/db7f887b0658ed042069dc0ff4cb266fe0596c23/internal/controllers/core/liveupdate/reconciler.go#L911

### DIFF
--- a/pkg/compose/watch.go
+++ b/pkg/compose/watch.go
@@ -17,6 +17,9 @@ package compose
 import (
 	"context"
 	"fmt"
+	"io/fs"
+	"os"
+	"path"
 	"path/filepath"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


ndeloof (approved):

glours (approved):
LGTM

PR #10399, opened 2023-03-24
watch involves up --build after change has been detected

ndeloof:
**What I did**
removed explicit request to rebuild images, then invoke up. Replaced by configuring services watched to have pull_policy=build so that images will always been rebuild. `watch` to rebuild image after changes has been detected is then strictly equivalent to `compose up --build SERVICE`

**Related issue**
As reported by @milas, after an initial rebuild, `watch` would run `up` but not rebuild anymore

image


milas (approved):
👍🏻

PR #10404, opened 2023-03-24
add dry-run support to run command

glours:
**What I did**
Add dry-run support for the `run` command

**Related issue**
https://docker.atlassian.net/browse/ENV-66

image

### DIFF
--- a/pkg/api/dryrunclient.go
+++ b/pkg/api/dryrunclient.go
@@ -87,12 +87,27 @@ func NewDryRunClient(apiClient client.APIClient, cli *command.DockerCli) (*DryRu
 // All methods and functions which need to be overridden for dry run.
 
 func (d *DryRunClient) ContainerAttach(ctx context.Context, container string, options moby.ContainerAttachOptions) (moby.HijackedResponse, error) {
-	return moby.HijackedResponse{}, ErrNotImplemented
+	return moby.HijackedResponse{}, errors.New("interactive run is not supported in dry-run mode")
 }
 
 func (d *DryRunClient) ContainerCreate(ctx context.Context, config *containerType.Config, hostConfig *containerType.HostConfig,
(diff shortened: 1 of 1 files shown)


pkg/api/dryrunclient.go
```diff
@@ -87,12 +87,40 @@ func NewDryRunClient(apiClient client.APIClient, cli *command.DockerCli) (*DryRu
...
 
 func (d *DryRunClient) ContainerAttach(ctx context.Context, container string, options moby.ContainerAttachOptions) (moby.HijackedResponse, error) {
-	return moby.HijackedResponse{}, ErrNotImplemented
+	return moby.HijackedResponse{}, errors.New("interactive run is not supported in dry-run mode")
 }
 
 func (d *DryRunClient) ContainerCreate(ctx context.Context, config *containerType.Config, hostConfig *containerType.HostConfig,
 	networkingConfig *network.NetworkingConfig, platform *specs.Platform, containerName string) (containerType.CreateResponse, error) {
-	return containerType.CreateResponse{}, ErrNotImplemented
+	projectLabel := fmt.Sprintf("com.docker.compose.project=%s", config.Labels["com.docker.compose.project"])
```
ndeloof, line 95: better use const from https://github.com/docker/compose/blob/v2/pkg/api/labels.go

ndeloof (commented):
I'm not sure to understand what you try to do here in `ContainerCreate`. Return existing container as a successful create if one existed ? But then this is not what would happen if user removes `--dry-run` flag as we will then create a second container for service

pkg/api/dryrunclient.go
glours: I did the implementation of this PR in 2 phases, first the ContainerCreate changes and then I add to tweak the ContainerInspect function. In fact the changes in the 2nd one fix issues I faced with no existing containers. 
I'll simplify the the `ContainerCreate` implementation, it would now fake the creation of a new container

ndeloof (approved):

milas (approved):

Issue #10439, opened 2023-04-05
[BUG] `--ansi always` does not have any effect


sueskind:
### Description

I want to use a Python script to start/stop docker-compose. However, when piping the stderr, the ANSI escape sequences are removed, even when using the `--ansi always` option.

### Steps To Reproduce

docker-compose.yml:

```yaml
version: "3.8"

services:
  nginx:
    image: nginx
```

main.py:

```python
from subprocess import Popen, PIPE, STDOUT

args = ["docker-compose", "--ansi", "always", "up", "-d"]

with Popen(args, stdout=PIPE, stderr=STDOUT) as process:
    for line in process.stdout:
        print(line.decode("utf-8"), end="")
```

This script invokes `docker-compose --ansi always up -d`, redirects stderr to stdin and collects the result which is then printed to the console.

However, when running `python main.py` there are no colors in the output.

I'm using Python 3.10.10.

### Compose Version

```shell
$ docker compose version
Docker Compose version 2.17.2

$ docker-compose version
Docker Compose version 2.17.2
```

### Docker Environment

```shell
$ docker info
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  compose: Docker Compose (Docker Inc.)
    Version:  2.17.2
    Path:     /usr/lib/docker/cli-plugins/docker-compose

Server:
 Containers: 19
  Running: 17
  Paused: 0
  Stopped: 2
 Images: 44
 Server Version: 23.0.2
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: true
  Native Overlay Diff: false
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 1fbd70374134b891f97ce19c70b6e50c7b9f4e0d.m
 runc version: 
 init version: de40ad0
 Security Options:
  seccomp
   Profile: builtin
  cgroupns
 Kernel Version: 6.2.9-arch1-1
 Operating System: Arch Linux
 OSType: linux
 Architecture: x86_64
 CPUs: 32
 Total Memory: 62.74GiB
 Name: dcnb-0057
[...]

PR #10442, opened 2023-04-05
ansi=auto|never|always

ndeloof:
**What I did**
`ansi` flag to accept legacy value `always`

**Related issue**
fixes https://github.com/docker/compose/issues/10439

### DIFF
--- a/cmd/compose/compose.go
+++ b/cmd/compose/compose.go
@@ -315,7 +315,7 @@ func RootCommand(streams command.Cli, backend api.Service) *cobra.Command { //no
 			switch ansi {
 			case "never":
 				progress.Mode = progress.ModePlain
-			case "tty":
+			case "always":
 				progress.Mode = progress.ModeTTY
 			}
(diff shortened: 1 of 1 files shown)


cmd/compose/compose.go
```diff
@@ -315,7 +315,7 @@ func RootCommand(streams command.Cli, backend api.Service) *cobra.Command { //no
 			switch ansi {
 			case "never":
 				progress.Mode = progress.ModePlain
-			case "tty":
+			case "always":
```
glours, line 318: Shouldn't we support both `tty` and `always`?
```suggestion
			case "tty", "always":
```

ndeloof: we could, but as this value never has been documented, and `SetANSIMode` doesn't manage this value, I considered we better should not pretend this even existed

glours (approved):

milas (approved):

Issue #10480, opened 2023-04-19
[BUG] Incomplete logs when "docker compose logs > filename &" is used and services are started one by one


ab-public:
### Description

When I redirect logs via `docker compose logs -f > filename &` to a file and start several services **one after the other** (this is needed in my setup), only the messages of the first two services appear in the file. The messages of the further started services are missing and the log file is incomplete. However, if I run "docker compose logs" directly (without redirecting), the messages are complete. See example...

Hint: With docker-compose 1.29.x this works as expected.

### Steps To Reproduce

Example for reproducing

**docker-compose.yml**

```
version: "3"
services:
  service-1:
    image: alpine
    hostname: service-1
    command: sh -c "while true; do echo hello from $(hostname); sleep 3; done"
  service-2:
    image: alpine
    hostname: service-2
    command: sh -c "while true; do echo hello from $(hostname); sleep 3; done"
  service-3:
    image: alpine
    hostname: service-3
    command: sh -c "while true; do echo hello from $(hostname); sleep 3; done"
  service-4:
    image: alpine
    hostname: service-4
    command: sh -c "while true; do echo hello from $(hostname); sleep 3; done"
```
**test.bash**

```
#!/usr/bin/env bash

docker compose up -d service-1

docker compose logs -t -f > example.log &
## start log capturing...

docker compose up -d service-2
## this service appears in file log

docker compose up -d service-3
## The messages of this and further started services do not appear in the log file
## But if you run "docker compose logs" they appear... 

docker compose up -d service-4
## ...

tail -f example.log
```

**My output**

The messages of example-service-3 and example-service-4 are missing in example.log !

```
❯ ./test.bash         
[+] Running 2/2
 ✔ Network example_default        Created                                                                         0.1s 
 ✔ Container example-service-1-1  Started                                                                         0.2s 
[+] Running 1/1
 ✔ Container example-service-2-1  Started                                                                         0.3s 
[+] Running 1/1
 ✔ Container example-service-3-1  Started                                                                         0.3s 
[+] Running 1/1
 ✔ Container example-service-4-1  Started                                                                         0.3s 
example-service-1-1  | 2023-04-19T11:47:13.601278918Z hello from service-1
example-service-1-1  | 2023-04-19T11:47:16.603528205Z hello from service-1
example-service-2-1  | 2023-04-19T11:47:17.003981961Z hello from service-2
example-service-1-1  | 2023-04-19T11:47:19.606058022Z hello from service-1
[...]

PR #10493, opened 2023-04-24
don't block events loop collecting logs

ndeloof:
**What I did**
as a new/restarted container is detected by `watchContainers`, the event handler must return ASAP otherwise the event loop is blocked. So `logContainers` must run in a goroutine in error group, with other log collectors

**Related issue**
fixes https://github.com/docker/compose/issues/10480
https://docker.atlassian.net/browse/ENV-175


milas (approved):
Tested locally and LGTM

---

I feel like we need to refactor some of this in the future -- I cannot actually follow the logic here, we've got multiple levels of `errgroup::Go()` calls happening across callback boundaries, for example

PR #10502, opened 2023-04-27
Dry run build command support

glours:
**What I did**
Add dry run support for `build` command

**Related issue**
https://docker.atlassian.net/browse/ENV-56

image

### DIFF
--- a/pkg/compose/build_buildkit.go
+++ b/pkg/compose/build_buildkit.go
@@ -18,6 +18,8 @@ package compose
 
 import (
 	"context"
+	"crypto/sha1"
+	"fmt"
 	"os"
 	"path/filepath"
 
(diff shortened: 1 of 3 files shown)


pkg/compose/build_buildkit.go
```diff
@@ -76,3 +84,23 @@ func (s *composeService) doBuildBuildkit(ctx context.Context, opts map[string]bu
 
 	return imagesBuilt, err
 }
+
+func (s composeService) dryRunBuildResponse(ctx context.Context, options map[string]build.Options) (resp map[string]*client.SolveResponse, err error) {
+	buildResponse := map[string]*client.SolveResponse{}
+	for name, option := range options {
+		dryRunUUID := fmt.Sprintf("dryRun-%x", sha1.Sum([]byte(name)))
+		args := []string{"buildx", "build", "--print=outline", "--output=type=cacheonly", "-f", option.Inputs.DockerfilePath, option.Inputs.ContextPath}
```
ulyssessouza, line 92: Sweet, but different from https://github.com/docker/compose/blob/e606701d9f7272e1858a4ce9d189e5347f337c70/pkg/compose/build_buildkit.go#L56 which is vendored. 

Not the same build options being passed.

pkg/compose/build_buildkit.go
```diff
@@ -76,3 +84,23 @@ func (s *composeService) doBuildBuildkit(ctx context.Context, opts map[string]bu
...
+		buildCMD := exec.CommandContext(ctx, "docker", args...)
```
ndeloof, line 93: we don't exec `docker buildx ...` for a standard build, but use the vendored buildx client, why not do the same here with `BuildOptions.PrintFunc`?

glours: As discuss internally, we'll wait for the buildx `--print` option to be marked as stable, this way we will only have to change the value in the `BuildOptions` and let buildx now when to build or not 👍 
I revert all the changes for a simplest solution for now

ndeloof (approved):

milas (approved):

milas (approved):
LGTM

jauderho:
It doesn't look like a harmful use of SHA1 here but generally, it would be a good idea to use SHA256 instead. Should this be changed?

PR #10507, opened 2023-04-28
make progress title configurable

glours:
**What I did**
Add a progress section title to tty writer so we're able to show a more accurate state than `Running` in the console output.
Screenshot 2023-04-28 at 12 15 39

This changes will also  improve dry run output.

image

### DIFF
--- a/pkg/compose/build.go
+++ b/pkg/compose/build.go
@@ -47,10 +47,10 @@ func (s *composeService) Build(ctx context.Context, project *types.Project, opti
 	if err != nil {
 		return err
 	}
-	return progress.Run(ctx, func(ctx context.Context) error {
+	return progress.RunWithTitle(ctx, func(ctx context.Context) error {
 		_, err := s.build(ctx, project, options)
 		return err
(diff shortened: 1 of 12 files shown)


ndeloof:
LGTM for a short terms fix

This is yet another occurence to demonstrate the progress mechanism is not adequate for our needs.
For example, on `pull` we don't distinguish the "download", "unpack" and "verify" phases, and then report 3 times a progression from 0 to 100% (hopefully, the 2 last steps are fast so nobody noticed :P).

Longer terms I think we should have components send structured events, not just strings, and define a real TUI to manage and format them. https://github.com/anchore/stereoscope has a nice UI, we should look into implementation for inspiration. Typically it fully parses the engine JSONMessages to convert them into structured events: https://github.com/anchore/stereoscope/tree/main/pkg/image/docker
(to be debated why moby doesn't send such events as an "API" and prefers to manage TUI from engine vs client, but ¯\\_(ツ)_/¯)

milas (approved):

PR #10508, opened 2023-04-28
Update 🥒 tests to use `dockerfile_inline`

laurazard:
**What I did**

Update Cucumber tests to use `dockerfile_inline` when appropriate, for clarity (and coverage).

image

### DIFF
--- a/e2e/cucumber-features/build-contexts.feature
+++ b/e2e/cucumber-features/build-contexts.feature
@@ -7,14 +7,12 @@ Background:
           a:
             build:
                 context: .
+                dockerfile_inline: |
+                    # syntax=docker/dockerfile:1
+                    FROM alpine:latest
+                    COPY --from=dep /etc/hostname /
                 additional_contexts:
                   - dep=docker-image://ubuntu:latest
-        """
-    And a dockerfile
-        """
-        # syntax=docker/dockerfile:1
-        FROM alpine:latest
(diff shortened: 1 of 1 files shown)


ndeloof:
LGTM. Maybe we could have both ?

ndeloof (approved):

glours (approved):

laurazard:
Yup! I changed both since the previous usecase is already thoroughly covered in the other tests, but we can add another cucumber scenario to make sure both are covered here. I'll TAL at it later.

milas (approved):

e2e/cucumber-features/ps.feature
```diff
@@ -8,14 +8,12 @@ Background:
             image: test:latest
             build:
                 context: ./
+                dockerfile_inline |
+                    FROM golang:1.19-alpine
```
milas, line 12:
```suggestion
                dockerfile_inline: |
                    FROM golang:1.19-alpine
```

milas (changes requested):
Oops I take-backsies my approval, there's a typo in the YAML so the CI is failing

e2e/cucumber-features/ps.feature
laurazard: thanks! :)

milas (approved):

PR #10526, opened 2023-05-03
trace: add OTEL initialization

milas:
**What I did**
This is a bunch of OTEL initialization code. It's all in `internal/` because there are re-usable parts here, but Compose isn't the right spot. Once we've stabilized the interfaces a bit and the need arises, we can move it to a separate module.

Currently, a single span is produced to wrap the root Compose command.

Compose will respect the standard OTEL environment variables as well as OTEL metadata from the Docker context. Both can be used simultaneously. The latter is intended for local system integration and is restricted to Unix sockets / named pipes.

None of this is enabled by default. It requires setting the `COMPOSE_EXPERIMENTAL_OTEL=1` environment variable to gate it during development.

grey fox

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -17,23 +17,33 @@
 package main
 
 import (
+	"context"
 	"os"
+	"time"
 
 	dockercli "github.com/docker/cli/cli"
 	"github.com/docker/cli/cli-plugins/manager"
 	"github.com/docker/cli/cli-plugins/plugin"
 	"github.com/docker/cli/cli/command"
+	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/codes"
+	"go.opentelemetry.io/otel/trace"
 
 	"github.com/docker/compose/v2/cmd/compatibility"
 	commands "github.com/docker/compose/v2/cmd/compose"
 	"github.com/docker/compose/v2/internal"
+	"github.com/docker/compose/v2/internal/tracing"
 	"github.com/docker/compose/v2/pkg/api"
 	"github.com/docker/compose/v2/pkg/compose"
 )
 
 func pluginMain() {
 	plugin.Run(func(dockerCli command.Cli) *cobra.Command {
+		var tracingShutdown tracing.ShutdownFunc
+		var cmdSpan trace.Span
+
 		serviceProxy := api.NewServiceProxy().WithService(compose.NewComposeService(dockerCli))
 		cmd := commands.RootCommand(dockerCli, serviceProxy)
 		originalPreRun := cmd.PersistentPreRunE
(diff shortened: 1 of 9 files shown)


internal/tracing/conn_unix.go
```diff
@@ -0,0 +1,26 @@
...
+	"context"
+	"fmt"
+	"net"
+	"strings"
+	"syscall"
+)
+
+const maxUnixSocketPathSize = len(syscall.RawSockaddrUnix{}.Path)
+
+func DialInMemory(ctx context.Context, addr string) (net.Conn, error) {
```
milas, line 13: @thaJeztah I was thinking of PRing this to `docker/go-connections` later. Thoughts? It's inspired by the `InmemSocket` there

ndeloof (approved):

internal/tracing/conn_unix.go
thaJeztah: @milas hmm. need to think about this; honest answer;

- there's an issue with the latest (main) version of `go-connections` that we still need to sort out (moby started failing when we tried to update it to latest)
- there's been some discussion about that repository (mainly around the TLS config specified in that repository, but also because it became a bit of a grab-bag of utilities that are not very related), and we need to re-think if we want to preserve it.


I wonder if we should have something like https://github.com/docker/go-metrics (either to provide abstractions, or to have a shared implementation for some of this).

(all just quick blurbs from my side! haven't dug in deep into all of this, but perhaps something we should discuss with more people)

internal/tracing/conn_windows.go
```diff
@@ -0,0 +1,14 @@
...
+	"context"
+	"net"
+	"time"
+
+	"github.com/Microsoft/go-winio"
+)
+
+func DialInMemory(ctx context.Context, addr string) (net.Conn, error) {
+	timeout := 1 * time.Second
+	return winio.DialPipeContext(ctx, addr, &timeout)
```
glours, line 13: Looks like CI complains about the timeout attribute (cf : https://github.com/docker/compose/actions/runs/4877301442/jobs/8737538726#step:6:322)

glours (commented):
Most of the new files need the Copyright header. 
e2e tests checking the console output seem to not be happy, maybe there is a trailing character added by the otel configuration? 🤔

internal/tracing/conn_unix.go
milas: Ahhh sorry, I should have clarified -- I was only referring to the `DialInMemory` function - the way that library is structured right now, you can:
 * create a listener type that works across OSes (impls behind build tags)
 * configure an HTTP transport across OSes
But it doesn't expose a function to create a platform-agnostic in-memory `net.Conn` directly


(That said, yes -- the rest of the observability pieces should be shared! I purposefully put this in `internal/tracing` so it _cannot_ become a dependency elsewhere & it doesn't depend on anything in the rest of the Compose code base. Will spin it out into its own repo/module later)

cmd/main.go
```diff
@@ -17,35 +17,88 @@
...
+		originalPersistentPostRunE := cmd.PersistentPostRunE
+		cmd.PersistentPostRunE = func(cmd *cobra.Command, args []string) (err error) {
+			defer func() {
+				if cmdSpan != nil {
+					if err != nil && !errors.Is(err, context.Canceled) {
+						cmdSpan.SetStatus(codes.Error, "CLI command returned error")
+						cmdSpan.RecordError(err)
+					}
+					cmdSpan.End()
+					fmt.Println("span end")
```
rumpl, line 89: We can safely remove this :)

glours (approved):
LGTM

Issue #10528, opened 2023-05-04
[BUG] build output is broken


ndeloof:
### Description

`docker compose build` output in recent releases is flacky and render buildkit output in random order:

bake build:
```
docker buildx bake -f compose.yaml
[+] Building 2.9s (9/9) FINISHED                                                                                                                                                                        
 => [test internal] load build definition from Dockerfile                                                                                                                                          0.0s
 => => transferring dockerfile: 109B                                                                                                                                                               0.0s
 => [test internal] load .dockerignore                                                                                                                                                             0.0s
 => => transferring context: 2B                                                                                                                                                                    0.0s
 => [hello internal] load .dockerignore                                                                                                                                                            0.0s
 => => transferring context: 2B                                                                                                                                                                    0.0s
 => [hello internal] load build definition from Dockerfile                                                                                                                                         0.0s
 => => transferring dockerfile: 109B                                                                                                                                                               0.0s
 => [test internal] load metadata for docker.io/library/ubuntu:latest                                                                                                                              2.9s
 => [test 1/3] FROM docker.io/library/ubuntu@sha256:dfd64a3b4296d8c9b62aa3309984f8620b98d87e47492599ee20739e8eb54fbf                                                                               0.0s
 => => resolve docker.io/library/ubuntu@sha256:dfd64a3b4296d8c9b62aa3309984f8620b98d87e47492599ee20739e8eb54fbf                                                                                    0.0s
 => CACHED [hello 2/3] RUN apt update && apt install -y git                                                                                                                                        0.0s
 => CACHED [hello 3/3] RUN echo "hello world"                                                                                                                                                      0.0s
 => [test] exporting to image                                                                                                                                                                      0.0s
 => => exporting layers                                                                                                                                                                            0.0s
 => => writing image sha256:58808bdbd3d8ca3e18bfcda8c73efc008d21bf6f61561d4ad05bc6b9dd14949e                                                                                                       0.0s
 => => naming to docker.io/library/alpine               
```

docker compose build:
```
[+] Building 9.1s (5/7)                                                                                                                                                                                 
 => [internal] load build definition from Dockerfile                                                                                                                                               0.0s 
 => => transferring dockerfile: 109B                                                                                                                                                               0.0s 
[+] Building 9.3s (5/7)                                                                                                                                                                                 
 => [internal] load .dockerignore                                                                                                                                                                  0.0s
 => => transferring context: 2B                                                                                                                                                                    0.0s
[...]

PR #10531, opened 2023-05-04
revisit the build UX by parsing buildkit solve events on our own

ndeloof:
**What I did**
`compose build` output is broken as we get multiple buildkit sessions concurrently trying to update the terminal.
This PR changes our build integration by passing a custom `progress.Writer` to collect buildkit's SolveStatus event, and format them for output with our own progress implementation. This isn't perfect either, but at least avoids console to flash with random lines

[demo](https://asciinema.org/a/lPPd5eqqVFLrnmWEB2FTZ6GII?autoplay=1)

**Related issue**
fixes https://github.com/docker/compose/issues/10528

### DIFF
--- a/pkg/compose/build_buildkit.go
+++ b/pkg/compose/build_buildkit.go
@@ -20,23 +20,120 @@ import (
 	"context"
 	"crypto/sha1"
 	"fmt"
-	"os"
 	"path/filepath"
+	"strings"
+	"unicode"
 
 	_ "github.com/docker/buildx/driver/docker"           //nolint:blank-imports
 	_ "github.com/docker/buildx/driver/docker-container" //nolint:blank-imports
 	_ "github.com/docker/buildx/driver/kubernetes"       //nolint:blank-imports
 	_ "github.com/docker/buildx/driver/remote"           //nolint:blank-imports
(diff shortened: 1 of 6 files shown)


pkg/compose/build_buildkit.go
```diff
@@ -20,23 +20,120 @@ import (
...
+	return true
+}
+
+func (b BuildOutput) ClearLogSource(i interface{}) {
+
+}
+
+func (b *BuildOutput) getVertexId(vertex digest.Digest) string {
+	return b.vertexes[vertex].Digest.String()
+}
```
glours, line 132: Can we move all this part in a dedicated file such as `pkg/compose/buildkitoutput`?

pkg/progress/event.go
```diff
@@ -41,10 +43,14 @@ const (
 	Working EventStatus = iota
 	// Done means that the current task is done
 	Done
+	// Done means that the current task is done
```
glours, line 46:
```suggestion
	// Success means that the current task is done and successful
```

glours (changes requested):
Some little changes! Display looks pretty neat otherwise 👏

PR #10533, opened 2023-05-05
move dry-run support from alpha to main command

### LINKED ISSUE #1203: docker-compose dry run

glours:
**What I did**
Move alpha `dry-run` command to a global `--dry-run` flag of the main Compose command

**Related issue**
closes #1203 

https://docker.atlassian.net/browse/ENV-183

image

### DIFF
--- a/docs/reference/compose.md
+++ b/docs/reference/compose.md
@@ -175,7 +175,7 @@ containers for the project.
 ### Use Dry Run mode to test your command
 
 Use `--dry-run` flag to test a command without changing your application stack state.
-Dry Run mode will show you all the steps Compose will apply by executing the command, for example:
+Dry Run mode shows you all the steps Compose applies when executing a command, for example:
 ```console
 $ docker compose --dry-run up --build -d
 [+] Pulling 1/1
@@ -192,10 +192,10 @@ $ docker compose --dry-run up --build -d
  ✔ DRY-RUN MODE -  Container nginx-golang-mysql-backend-1                                Started                                                                                                                                           0.0s
  ✔ DRY-RUN MODE -  Container nginx-golang-mysql-proxy-1                                  Started                                     Started
 ```
-You could see that the first step will be to pull the image defined by `db` service, then build the `backend` service.  
-After that, containers will be created, the `db` service started and the `backend` and `proxy` will wait until `db` service is healthy to start.
+From the example above, you can see that the first step is to pull the image defined by `db` service, then build the `backend` service.  
+Next, the containers are created. The `db` service is started, and the `backend` and `proxy` wait until the `db` service is healthy before starting.
 
-The Dry Run mode is not supported by all commands, especially by the command which doesn't change the state of a Compose stack
+Dry Run mode does not currently work with all commands. In particular, you cannot use Dry Run mode with a command that doesn't change the state of a Compose stack
 such as `ps`, `ls`, `logs` for example.  
 
 Here the list of commands supporting `--dry-run` flag:
--- a/docs/reference/compose.md
+++ b/docs/reference/compose.md
@@ -41,6 +41,7 @@ Docker Compose
 |:-----------------------|:--------------|:--------|:----------------------------------------------------------------------------------------------------|
 | `--ansi`               | `string`      | `auto`  | Control when to print ANSI control characters ("never"\|"always"\|"auto")                           |
 | `--compatibility`      |               |         | Run compose in backward compatibility mode                                                          |
+| `--dry-run`            |               |         | Execute command in dry run mode                                                                     |
 | `--env-file`           | `stringArray` |         | Specify an alternate environment file.                                                              |
 | `-f`, `--file`         | `stringArray` |         | Compose configuration files                                                                         |
 | `--parallel`           | `int`         | `-1`    | Control max parallelism, -1 for unlimited                                                           |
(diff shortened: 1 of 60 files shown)


glours:
@aevesdocker can you check the documentation part of this PR please 🙏 ?

thaJeztah:
I wonder if we really want this to be a global flag; even more so because it's not supported by all commands (which we now have to document as "this may, or may not work", depending on what command you run)


The global flags are also somewhat awkward, because they're global "to compose", but that means they go somewhere in the middle;


```
docker [DOCKER GLOBAL OPTIONS] compose [COMPOSE GLOBAL OPTIONS] command [OPTIONS]
```



And I _expect_  "dry run" would usually mean ;

- I want to run a command (`docker compose -f my-compose-file up --build --pull`)
- But let's dry-run first
- NoI have to edit the _top level_ options, instead of just "add `--dry-run` at the end"


/cc @jalonsogo

ndeloof:
> I wonder if we really want this to be a global flag

+1 for a global flag. For the very few commands where it's not relevant, this is just a no-op

> The global flags are also somewhat awkward, because they're global "to compose", but that means they go somewhere in the middle;

this is how CLI plugins works ¯\\_(ツ)_/¯ . But you also can run `docker-compose` directly :P

docs/reference/compose.md
```diff
@@ -169,3 +170,49 @@ If flags are explicitly set on the command line, the associated environment vari
 
 Setting the `COMPOSE_IGNORE_ORPHANS` environment variable to `true` will stop docker compose from detecting orphaned
 containers for the project.
+
+
+### Use Dry Run mode to test your command
+
+Use `--dry-run` flag to test a command without changing your application stack state.
+Dry Run mode will show you all the steps Compose will apply by executing the command, for example:
```
aevesdocker, line 178:
```suggestion
Dry Run mode shows you all the steps Compose applies when executing a command, for example:
```

docs/reference/compose.md
```diff
@@ -169,3 +170,49 @@ If flags are explicitly set on the command line, the associated environment vari
...
+ ✔ DRY-RUN MODE -  Container nginx-golang-mysql-proxy-1                                  Created                                                                                                                                           0.0s
+ ✔ DRY-RUN MODE -  Container nginx-golang-mysql-db-1                                     Healthy                                                                                                                                           0.5s
+ ✔ DRY-RUN MODE -  Container nginx-golang-mysql-backend-1                                Started                                                                                                                                           0.0s
+ ✔ DRY-RUN MODE -  Container nginx-golang-mysql-proxy-1                                  Started                                     Started
+```
+You could see that the first step will be to pull the image defined by `db` service, then build the `backend` service.
```
aevesdocker, line 195:
```suggestion
From the example above, you can see that the first step is to pull the image defined by `db` service, then build the `backend` service.  
```

docs/reference/compose.md
```diff
@@ -169,3 +170,49 @@ If flags are explicitly set on the command line, the associated environment vari
...
+ ✔ DRY-RUN MODE -  Container nginx-golang-mysql-proxy-1                                  Created                                                                                                                                           0.0s
+ ✔ DRY-RUN MODE -  Container nginx-golang-mysql-db-1                                     Healthy                                                                                                                                           0.5s
+ ✔ DRY-RUN MODE -  Container nginx-golang-mysql-backend-1                                Started                                                                                                                                           0.0s
+ ✔ DRY-RUN MODE -  Container nginx-golang-mysql-proxy-1                                  Started                                     Started
+```
+You could see that the first step will be to pull the image defined by `db` service, then build the `backend` service.  
+After that, containers will be created, the `db` service started and the `backend` and `proxy` will wait until `db` service is healthy to start.
```
aevesdocker, line 196:
```suggestion
Next, the containers are created. The `db` service is started, and the `backend` and `proxy` wait until the `db` service is healthy before starting.
```

docs/reference/compose.md
```diff
@@ -169,3 +170,49 @@ If flags are explicitly set on the command line, the associated environment vari
...
+
+The Dry Run mode is not supported by all commands, especially by the command which doesn't change the state of a Compose stack
```
aevesdocker, line 198:
```suggestion
Dry Run mode does not currently work with all commands. In particular, you cannot use Dry Run mode with a command that doesn't change the state of a Compose stack
```

docs/reference/docker_compose.yaml
```diff
@@ -117,6 +117,51 @@ long: |-
 
     Setting the `COMPOSE_IGNORE_ORPHANS` environment variable to `true` will stop docker compose from detecting orphaned
     containers for the project.
+
+
+    ### Use Dry Run mode to test your command
```
aevesdocker, line 122: Comments above apply here too of course

thaJeztah:
> +1 for a global flag. For the very few commands where it's not relevant, this is just a no-op

So how do you use it?


```bash
docker --context=remote compose -f=one.yaml -f two.yaml up --no-color --pull --build --wait
````

Oh! Let me first verify if that will do what I want

```bash
<left-arrow><left-arrow><left-arrow><left-arrow><left-arrow><left-arrow><left-arrow><left-arrow><left-arrow><left-arrow><left-arrow><left-arrow><left-arrow><left-arrow><left-arrow><left-arrow><left-arrow><left-arrow>

docker --context=remote compose -f=one.yaml -f two.yaml --dry-run up --no-color --pull --build --wait
<enter>
```

Yes, looks good

```bash
<left-arrow><left-arrow><left-arrow><left-arrow><left-arrow><left-arrow><left-arrow><left-arrow><left-arrow><left-arrow><left-arrow><left-arrow><left-arrow><left-arrow><left-arrow><left-arrow><left-arrow><left-arrow>

docker --context=remote compose -f=one.yaml -f two.yaml up --no-color --pull --build --wait
<enter>
```

docs/reference/docker_compose.yaml
aevesdocker: docs content on this page now doesn't match what's on `docs/reference/compose.md`. Can we make the same changes as we did above?

glours:
@thaJeztah did you know you can use shortcuts to move faster in you shell ALT+<left-arrow> or ALT+B? 😇 😉 
Joke aside, I got your point, we'll discuss this during our next team meeting

ndeloof:
> the whole v2 project was to get rid of that

the v2 project goal is to align compose codebase with other Docker tools, so using golang. Being a CLI plugin is an implementation detail, and we still 100% support running as standalone.

glours:
@thaJeztah using a `--dry-run` as a persistant flag will cover the both cases 🥳 
Thank you so much @milas for reminding us Cobra supported such option 🙏

aevesdocker (approved):

milas (approved):

ndeloof (approved):

PR #10535, opened 2023-05-05
fix builkit progressui integration

ndeloof:
**What I did**
share progressui for all images being built, isolate concurrent image builds using `buildx.WithPrefix`

this is a simpler alternative to https://github.com/docker/compose/pull/10531, which we could include in a patch release

**Related issue**
fix https://github.com/docker/compose/issues/10528

### DIFF
--- a/pkg/compose/build.go
+++ b/pkg/compose/build.go
@@ -19,6 +19,7 @@ package compose
 import (
 	"context"
 	"fmt"
+	"os"
 	"path/filepath"
 
 	"github.com/compose-spec/compose-go/types"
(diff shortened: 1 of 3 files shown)


pkg/compose/build.go
```diff
@@ -124,6 +139,18 @@ func (s *composeService) build(ctx context.Context, project *types.Project, opti
 	return imageIDs, err
 }
 
+func getServiceIndex(project *types.Project, name string) (types.ServiceConfig, int) {
+	var service types.ServiceConfig
+	var idx int
+	for i, s := range project.Services {
+		if s.Name != name {
```
milas, line 146:
```suggestion
		if s.Name == name {
```

milas (changes requested):
Looks good to me with the proposed change 🤪

milas (approved):
LGTM!

Issue #10068, opened 2022-12-12
container re-created when it shouldn't


swingbit:
### Description

I'm experiencing an issue similar to #9600 and #9897, i.e. a container gets recreated needlessly.

`docker-compose.yml contains` 3 services:
- `tools` (depends on `api`)
- `api` (depends on `db`)
- `db`

At least once every day, a `docker-compose run tools` recreates the `api` container, while nothing has changed. Trying the same again does not recreate the `api` container.

Container images are very much static. They haven't been updated for weeks.

For now I cannot easily replicate this with a simple example, but the following is to my understanding good evidence that there is a bug. 

I started saving the result of this every day:
```
TS=$(date +%s)
docker-compose config > config.${TS}
docker-compose config --hash api > hash.${TS}
docker inspect api -f '{{json .Config.Labels}}' | jq > labels.${TS}
```

The result of the first two commands does not change across two days (with the container being recreated in between).
The result of the third one (the labels) does change:

```
# diff config.1670434754 config.1670766565
# diff hash.1670434754 hash.1670766565
# diff labels.1670434754 labels.1670766565
2c2
<   "com.docker.compose.config-hash": "e4c45cf84fe52b2e910b49746c2c6102067609eea52ede56cbe7d52d3dc76545",
---
>   "com.docker.compose.config-hash": "e38f8b5bff668614848f485f577db900a56eb031386c32bc996b6187b357e161",
```

So, the label says the hash has changed, but the hash hasn't changed!

### Compose Version

```Text
# docker-compose version
Docker Compose version v2.12.0

# docker compose version
Docker Compose version v2.12.0
```

### Docker Environment

```Text
# docker info
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  app: Docker App (Docker Inc., v0.9.1-beta3)
  buildx: Docker Buildx (Docker Inc., v0.9.1-docker)
  compose: Docker Compose (Docker Inc., v2.12.0)
  scan: Docker Scan (Docker Inc., v0.17.0)

Server:
 Containers: 3
  Running: 3
  Paused: 0
  Stopped: 0
 Images: 16
 Server Version: 20.10.20
 Storage Driver: overlay2
  Backing Filesystem: xfs
  Supports d_type: true
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
[...]

PR #10540, opened 2023-05-09
fix container being recreated while config has not changed

ndeloof:
**What I did**
As we don't load project `WithDiscardEnvFile` the config hash includes `env_file` attribute. But when selecting service(s) to be ran, compose-go `EnableService` always applies the discard option (this could be considered a bug, but that's another discussion :P)

**Related issue**
fixes https://github.com/docker/compose/issues/10068


glours (approved):
LGTM

milas (approved):

Issue #10532, opened 2023-05-04
[BUG] docker compose up hangs with no output when resolving deps with --parallel limit


fho:
### Description

`docker compose --parallel=8 up -d --remove-orphans LONG-LIST-OF-SERVICES` sometimes hangs with no output at all. The compose plugin process will hang forever and never terminate.

I built a compose 2.17.3 binary with debug symbols and got the stacktraces from the goroutines.

It  is hanging when it tries to resolve the service dependencies.
8  go-routines ((**same than `--parallel` limit**) are hanging at [`eg.Go(func() error {` in `graphTraversal.run()`](https://github.com/docker/compose/blob/7b0ed13209894b5981608f7d9e329a3a58af957d/pkg/compose/dependencies.go#L126), waiting for another slot in the `errgroup.Group`. 
1 go-routine is waiting for all [`errgroup.Group` routines to terminate](https://github.com/docker/compose/blob/7b0ed13209894b5981608f7d9e329a3a58af957d/pkg/compose/dependencies.go#L108). In the same goroutine the Semaphore in the errgroup is at the limit:
```
Frame 6: github.com/docker/compose/v2/pkg/compose/dependencies.go:108 (PC: 2723605)
(dlv) print eg
*golang.org/x/sync/errgroup.Group {
	cancel: context.WithCancel.func1,
	wg: sync.WaitGroup {
[..]
	sem: chan golang.org/x/sync/errgroup.token {
		qcount: 8,
		dataqsiz: 8,
```

My theory is that there is a deadlock race in https://github.com/docker/compose/blob/7b0ed13209894b5981608f7d9e329a3a58af957d/pkg/compose/dependencies.go#L113-L134.

`run()` can start multiple go-routines in the errgroup.Group that invoke `run()` again.
It happens that `maxConcurrency` times `run` was  started via `errgroup.Group.Go()` and all of them are hanging when trying to call `errgroup.Group.Go()` again.
That causes the deadlock.

Please let me know if you need more information from the memory dump of the hanging docker-compose process.

Here are the stacktraces of all go-routines when it happened:
```
(dlv) goroutines -t
  Goroutine 1 - User: runtime/sema.go:62 sync.runtime_Semacquire (0x470dc7) [semacquire 288429366834]
	 0  0x000000000044261d in runtime.gopark
	     at runtime/proc.go:382
	 1  0x00000000004426aa in runtime.goparkunlock
	     at runtime/proc.go:387
[...]

PR #10544, opened 2023-05-09
fix race condition when --parallel is used with a large number of dependent services

ndeloof:
**What I did**
dependency graph traversal has been implemented with recursive calls, this prevent the error group max concurrency to be used as a parallel limit: a Node visitor, once completed, won't leave the routine concurrent slot for another use.

This PR changes the approach by introducing a "controller" routine, responsible to trigger visit of the adjacent nodes. Visitor notify abut completion by aa channel, doing so the routine can complete. This require limit to be set to concurrency + 1 for the controller. Controller exits after all nodes have been visited.

**Related issue**
fixes https://github.com/docker/compose/issues/10532


milas (approved):
LGTM, nice find

glours (approved):
LGTM

Issue #7346, opened 2020-04-07
Why does docker-compose log to STDERR for `Building ....` instead of STDOUT when executing docker-compose build


devedse:
## Description of the issue

I'm working on an Azure DevOps extension which needs to read the docker-compose build output. I was however running into the issue that the STDOUT had some missing log lines: (Lines with #### in front of it where missing from STDOUT)

```
##### Building sample1 <-- missing
Step 1/2 : FROM nginx:alpine
 ---> 377c0837328f
Step 2/2 : COPY . /usr/share/nginx/html
 ---> Using cache
 ---> 6cac8912ab73

Successfully built 6cac8912ab73
Successfully tagged sample1:latest
##### Building sample2 <-- missing
Step 1/2 : FROM nginx:alpine
 ---> 377c0837328f
Step 2/2 : COPY . /usr/share/nginx/html
 ---> Using cache
 ---> 6cac8912ab73

Successfully built 6cac8912ab73
Successfully tagged sample2:latest
```

I've been doing some investigation and have created a repository to reproduce this:
https://github.com/devedse/STDOUTMissingError/blob/master/README.md

As you can see there, apparently the lines `Building ...` are being written to STDERR instead of STDOUT.

Does anyone have any idea on why this is happening?

## Context information (for bug reports)

**Output of `docker-compose version`**
```
docker-compose version 1.25.4, build 8d51620a
```

**Output of `docker version`**
```
Client: Docker Engine - Community
 Version:           19.03.8
 API version:       1.40
 Go version:        go1.12.17
 Git commit:        afacb8b
 Built:             Wed Mar 11 01:23:10 2020
 OS/Arch:           windows/amd64
 Experimental:      false

Server: Docker Engine - Community
 Engine:
  Version:          19.03.8
  API version:      1.40 (minimum version 1.12)
  Go version:       go1.12.17
  Git commit:       afacb8b
  Built:            Wed Mar 11 01:29:16 2020
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          v1.2.13
  GitCommit:        7ad184331fa3e55e52b890ea95e65ba581ae3429
 runc:
  Version:          1.0.0-rc10
  GitCommit:        dc9208a3303feef5b3839f4323d9beb36df0a9dd
 docker-init:
  Version:          0.18.0
  GitCommit:        fec3683
```

**Output of `docker-compose config`**
See the repository linked above, I've added automated builds to that.

## Steps to reproduce the issue
[...]

PR #10549, opened 2023-05-10
introduce ability to write status messages on stdout

ndeloof:
**What I did**
introduce `stdinfo()` as output stream to dump status messages about compose progress applying requested operation.
this defaults to POSIX stderr (to easily filter out container output on stdout) but can be switched by setting `COMPOSE_STATUS_STDOUT=1`

**Related issue**
closes https://github.com/docker/compose/issues/7346


laurazard (approved):
Nice 🥳 I'm super happy to finally be able to configure this :)

milas (approved):

Issue #10230, opened 2023-02-02
compose down: add [SERVICE...] argument


Excpt0r:
### Description

Most compose subcommands support an optional list of services to operate with, instead of all compose services.
Is there a good reason why the `down` command behaves differently?

My use case:
Stop a single service of the compose stack, AND remove all corresponding named volumes with it.

A close workaround would be to use `compose stop && compose rm -v` - but this works only for anonymous volumes, not named volumes.
So in addition to the [SERVICE...] argument for `compose down`, maybe it would also be nice and consistent to have an additional flag for the `compose rm` to also include named volumes, when needed (not my use case).

ndeloof:
no specific reason, you're welcome to propose a pull-request if you feel confident with Go development :)

Excpt0r:
@ndeloof thanks for your response.
I had a look, but this change will probably require more than just adding some cmd flags, so no PR from my side, sorry.

The only workaround I see so far, is to use "docker volume rm" for the corresponding name volumes.
But that requires of course that the volume cleanup process knows the mapping between services and their volumes.
During runtime this could be retrieved by "docker container inspect", but deleting volumes from services that are currently not running/created would require to get the info from compose.yml.

PR #10552, opened 2023-05-10
introduce ability to select service to be stopped by `compose down`

ndeloof:
**What I did**
introduce ability to select service to be stopped by `compose down`

**Related issue**
closes https://github.com/docker/compose/issues/10230

require https://github.com/docker/compose/pull/10555

### DIFF
--- a/pkg/compose/down.go
+++ b/pkg/compose/down.go
@@ -44,7 +44,7 @@ func (s *composeService) Down(ctx context.Context, projectName string, options a
 	}, s.stdinfo())
 }
 
-func (s *composeService) down(ctx context.Context, projectName string, options api.DownOptions) error {
+func (s *composeService) down(ctx context.Context, projectName string, options api.DownOptions) error { //golint:nocyclo
 	w := progress.ContextWriter(ctx)
 	resourceToRemove := false
 
@@ -70,6 +70,9 @@ func (s *composeService) down(ctx context.Context, projectName string, options a
 	}
 
 	err = InReverseDependencyOrder(ctx, project, func(c context.Context, service string) error {
+		if len(options.Services) > 0 && !utils.StringContains(options.Services, service) {
+			return nil
+		}
 		serviceContainers := containers.filter(isService(service))
 		err := s.removeContainers(ctx, w, serviceContainers, options.Timeout, options.Volumes)
 		return err
--- a/pkg/compose/down.go
+++ b/pkg/compose/down.go
@@ -44,7 +44,7 @@ func (s *composeService) Down(ctx context.Context, projectName string, options a
 	}, s.stdinfo())
 }
 
-func (s *composeService) down(ctx context.Context, projectName string, options api.DownOptions) error { //golint:nocyclo
+func (s *composeService) down(ctx context.Context, projectName string, options api.DownOptions) error { //nolint:gocyclo
 	w := progress.ContextWriter(ctx)
 	resourceToRemove := false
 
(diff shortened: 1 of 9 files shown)


glours (approved):
LGTM, I think you need to update the doc also

milas (changes requested):
Current code looks like a good starting point, but I think we'll have some special cases to look at.

For example:
```yaml
services:
  a:
   image: nginx
  b:
   image: nginx
   depends_on: [a]
```

First, `docker compose down b` will fail because it'll try to remove the network:
```
❯ ~/dev/compose/bin/build/docker-compose up -d && ~/dev/compose/bin/build/docker-compose down b
[+] Building 0.0s (0/0)
[+] Running 2/2
 ✔ Container network-a-1  Running                                                                                                                     0.0s
 ✔ Container network-b-1  Started                                                                                                                     0.4s
[+] Running 2/1
 ✔ Container network-b-1    Removed                                                                                                                   0.1s
 ✘ Network network_default  Error                                                                                                                     0.0s
failed to remove network network_default: Error response from daemon: error while removing network: network network_default id 0a59b1a2a52def0193fd82617668d734ebfb33fc433cf80862e1db012c44ee1f has active endpoints
```

But also, what should happen if you `docker compose down a`? Currently, it'll just stop `a` even though `b` depends upon it 🤔

ndeloof:
> docker compose down b will fail because it'll try to remove the network:

yes, this is why this PR depends on https://github.com/docker/compose/pull/10555

> what should happen if you docker compose down a?

good point. For symmetry with `compose up` we could also stop dependent service and include support for `--no-deps` flag

ndeloof:
https://github.com/docker/compose/pull/10555 has been merged 🎉

implemented complimentary support for dependent services, walking the dependency graph from specified service nodes (actually walking the whole graph, but skipping nodes which aren't relevant for requested operation)

I haven't introduced `--no-deps` in this PR, this is something we could add if requested

pkg/compose/down.go
```diff
@@ -65,6 +65,21 @@ func (s *composeService) down(ctx context.Context, projectName string, options a
...
+		if err != nil {
+			if options.Project != nil {
+				// ran with an explicit compose.yaml file, so we should not ignore
+				return err
+			}
+		} else {
+			services = append(services, service)
+		}
+	}
+	options.Services = services
```
glours, line 81: Can we extract this as a function, this way we won't need the `//nolint:gocyclo` and will make this code reusable elsewhere? WDYT?

glours (changes requested):

pkg/compose/down.go
ndeloof: unfortunately, even with this extracted into a func, cyclomatic complexity still is 17

ndeloof: moved to `checkSelectedServices`

glours (approved):
LGTM

milas (commented):
No issues with the code as-is, and stopping a service is depended upon by other services properly brings those down as well.

Volumes and `--rmi` flag are problematic, however:
```yaml
name: pr10552

services:
  parent:
    image: nginx
    volumes:
      - aa:/a
  child:
    depends_on: [parent]
    image: nginx
    volumes:
      - aa:/a
      - bb:/b
  other:
    image: nginx
    volumes:
      - bb:/b

volumes:
 aa:
 bb:
```

Assume you've run `compose up --wait` first:

### Images
```shell
$ compose down other --rmi=all
[+] Running 2/1
 ✔ Container pr10552-other-1  Removed                                                                                                                                          0.1s
 ⠋ Image nginx:latest         Removing                                                                                                                                         0.0s
 ! Network pr10552_default    Resource is still in use                                                                                                                         0.0s
Error response from daemon: conflict: unable to remove repository reference "nginx:latest" (must force) - container 9e9979562577 is using its referenced image b005e88565d7
```

### Volumes
```shell
$ compose down child --volumes
[+] Running 3/1
 ✔ Container pr10552-child-1  Removed                                                                                                                                          0.1s
 ✔ Volume pr10552_bb          Removed                                                                                                                                          0.0s
 ⠋ Volume pr10552_aa          Removing                                                                                                                                         0.0s
 ! Network pr10552_default    Resource is still in use                                                                                                                         0.0s
Error response from daemon: remove pr10552_aa: volume is in use - [8199008ae17b920f984648b9d3f03bbd620c052007157413a0a7a4d8ae37ed2c]
```

---

For now, what do you think about blocking the usage of the `--rmi` and `--volumes` flags if `len(services) != 0`?

milas (approved):
🥳

Paraphraser:
I just tried out v2.19.0. This feature is *waaay* cool. I really like the way it takes account of container dependency chains and cleans up dangling networks.

👏👏👏👏👏

Issue #9970, opened 2022-11-03
[BUG] --no-interpolate broken for volumes path


lostcontrol:
### Description

`docker-compose config/convert --no-interpolate` is broken in releases > 2.0.1 when substitution occurs in volumes path definition. The issue is better seen in the example below:

```bash
jac@L-0267 /t/compose (v2)> cat .env
TEST_DIR="/my_test_dir"
```
```yaml
jac@L-0267 /t/compose (v2)> cat docker-compose.yml 
version: "3.7"
services:
  test:
    image: ubuntu:22.04
    volumes:
      - type: bind
        source: ${TEST_DIR}
        target: /data
```
```bash
jac@L-0267 /t/compose (v2)> docker-compose --version
Docker Compose version v2.12.2
```
```yaml
jac@L-0267 /t/compose (v2)> docker-compose -f docker-compose.yml config
name: compose
services:
  test:
    image: ubuntu:22.04
    networks:
      default: null
    volumes:
    - type: bind
      source: /my_test_dir
      target: /data
networks:
  default:
    name: compose_default
```
```yaml
jac@L-0267 /t/compose (v2)> docker-compose -f docker-compose.yml config --no-interpolate
name: compose
services:
  test:
    image: ubuntu:22.04
    networks:
      default: null
    volumes:
    - type: bind
      source: /tmp/compose/${TEST_DIR}
      target: /data
networks:
  default:
    name: compose_default
```
The expected output when using `--no-interpolate` in this case would be:

```yaml
name: compose
services:
  test:
    image: ubuntu:22.04
    networks:
      default: null
    volumes:
    - type: bind
      source: ${TEST_DIR}
      target: /data
networks:
  default:
    name: compose_default
```
This is how it was until v2.0.1. From v2.1.0 and up to the last release (v2.12.2 as of now), the path to the current context (here `/tmp/compose`) is wrongly added at the front of the variable. Somehow, docker-compose thinks that `${TEST_DIR}` is a relative path and decides to include the current context in front of it.

### Steps To Reproduce

* Install the latest docker-compose version
* Create `.env` and `docker-compose.yml` as above
* Run the above commands

### Compose Version

```Text
Docker Compose version v2.12.2
```

### Docker Environment

```Text
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Build with BuildKit (Docker Inc., v0.5.1)

Server:
 Containers: 4
  Running: 1
  Paused: 0
  Stopped: 3
 Images: 31
[...]

PR #10557, opened 2023-05-11
introduce --no-path-resolution to skip relative path to be resolved

ndeloof:
**What I did**
introduce --no-path-resolution to skip relative path to be resolved

**Related issue**
closes https://github.com/docker/compose/issues/9970


milas (approved):
LGTM

glours (approved):
LGTM, you just need to regenerate the documentation 😉

ndeloof:
also need to get compose-go released as this PR uses a custom `replace` rule

Issue #9914, opened 2022-10-13
`docker compose down --rmi local` doesn't remove images when specified with `--project-name`


Mukundan314:
**Description**

**Steps to reproduce the issue:**
1. Create minimal Dockerfile and docker-compose.yml:
 ```
# Dockerfile
FROM alpine

# docker-compose.yaml
version: "3.9"

services:
  test-service:
    build: .
    command: sleep inf
```
2. Bring up services with `docker-compose -p hello-world up -d`
3. Bring down services with `docker-compose -p hello-world down --rmi local`

**Describe the results you received:**
The image created for `test-service` was not deleted

**Describe the results you expected:**
The image created for `test-service` should have been deleted

**Additional information you deem important (e.g. issue happens only occasionally):**
  - Things work as expected if `--project-name` is not specified
  - `--rmi all` works as expected and removes the image created for `test-service`
  - This is slightly different from #9655 since this is specific to `--rmi local`

**Output of `docker compose version`:**

```
Docker Compose version v2.11.2
```

**Output of `docker info`:**

```
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc., 0.9.1)
  compose: Docker Compose (Docker Inc., v2.11.0)

Server:
 Containers: 1
  Running: 1
  Paused: 0
  Stopped: 0
 Images: 25
 Server Version: 20.10.18
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: false
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: runc io.containerd.runc.v2 io.containerd.runtime.v1.linux
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 9cd3357b7fd7218e4aec3eae239db1f68a5a6ec6.m
 runc version: 
 init version: de40ad0
 Security Options:
  seccomp
   Profile: default
  cgroupns
 Kernel Version: 5.19.13-arch1-1
 Operating System: Arch Linux
 OSType: linux
 Architecture: x86_64
 CPUs: 4
 Total Memory: 31.31GiB
 Name: sigma
[...]

PR #10558, opened 2023-05-11
fix local image removal when `compose down` is ran with `--project-name`

ndeloof:
**What I did**
As previously noted in comment, with built image being tagged to make listing more deterministic, we don't need anymore to check service.image being set. Running with `--project-name` this attribute is always set, whenever it might not have been in the initial compose.yaml file.

**Related issue**
closes https://github.com/docker/compose/issues/9914


milas (approved):
LGTM!

(Btw I have zero recollection of writing that `TODO`)

Issue #9604, opened 2022-06-27
Support an environment or user-/system-wide default for --ansi


rowdybeaver:
**I would like to have some method to force `--ansi never`, either by a `COMPOSE_ANSI=never` environment variable or some setting elsewhere.**

I have 'up'/'down' commands in some tools (Pycharm, for an example) where I cannot influence the command line, resulting in hundreds of lines of the continuing status for each service generating every 0.1s.

**Output of `docker compose version`:**

```
Docker Compose version v2.6.0
```

**Output of `docker info`:**

```
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  app: Docker App (Docker Inc., v0.9.1-beta3)
  buildx: Docker Buildx (Docker Inc., v0.8.2-docker)
  compose: Docker Compose (Docker Inc., v2.6.0)
  scan: Docker Scan (Docker Inc., v0.17.0)

Server:
 Containers: 9
  Running: 6
  Paused: 0
  Stopped: 3
 Images: 231
 Server Version: 20.10.17
 Storage Driver: overlay2
  Backing Filesystem: btrfs
  Supports d_type: true
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 1
 Plugins:
  Volume: local local-persist
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runtime.v1.linux runc io.containerd.runc.v2
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 10c12954828e7c7c9b6e0ea9b0c02b01407d3ae1
 runc version: v1.1.2-0-ga916309
 init version: de40ad0
 Security Options:
  apparmor
  seccomp
   Profile: default
 Kernel Version: 5.4.0-121-generic
 Operating System: Ubuntu 20.04.4 LTS
 OSType: linux
 Architecture: x86_64
 CPUs: 8
 Total Memory: 7.631GiB
 Name: laptop8
 ID: SWHF:XE3Q:4XZT:5P5D:EPIO:PMW7:KALX:CRLL:ZSYV:7SAO:22XE:A5HB
 Docker Root Dir: /var/lib/docker
 Debug Mode: false
 Username: rider1
 Registry: https://index.docker.io/v1/
 Labels:
 Experimental: true
 Insecure Registries:
[...]

PR #10559, opened 2023-05-12
introduce COMPOSE_ANSI to define --ansi default value

ndeloof:
**What I did**
introduce COMPOSE_ANSI to define --ansi default value

**Related issue**
closes https://github.com/docker/compose/issues/9604


glours (approved):
LGTM

Issue #10560, opened 2023-05-12
docker compose fail on windows


alexsiryi14:
### Description

docker compose on windows 10 failed with the stack trace: 
```
panic: interface conversion: *windowsconsole.ansiWriter is not console.File: missing method Close

goroutine 1 [running]:
github.com/docker/compose/v2/pkg/progress.NewWriter({0x2428180, 0xc0000d0960}, {0x240bca0?, 0xc00015d140})
        github.com/docker/compose/v2/pkg/progress/writer.go:115 +0xc9
github.com/docker/compose/v2/pkg/progress.RunWithStatus({0x2428180, 0xc0000d0960}, 0xc000659ef0, {0x240bca0, 0xc00015d140})
        github.com/docker/compose/v2/pkg/progress/writer.go:71 +0xa5
github.com/docker/compose/v2/pkg/progress.Run({0x2428180, 0xc0000d0960}, 0xc0003fa440, {0x240bca0, 0xc00015d140})
        github.com/docker/compose/v2/pkg/progress/writer.go:62 +0x85
github.com/docker/compose/v2/pkg/compose.(*composeService).Up(0xc0003fb160, {0x2428180, 0xc0000d0960}, 0xc0000c96b0, {{{0x33d4128, 0x0, 0x0}, 0x0, 0x0, {0x21057a8, ...}, ...}, ...})
        github.com/docker/compose/v2/pkg/compose/up.go:34 +0x172
github.com/docker/compose/v2/pkg/api.(*ServiceProxy).Up(0xc0003aa0f0, {0x2428180, 0xc0000d0960}, 0x0?, {{{0x33d4128, 0x0, 0x0}, 0x0, 0x0, {0x21057a8, ...}, ...}, ...})
        github.com/docker/compose/v2/pkg/api/proxy.go:181 +0x14a
github.com/docker/compose/v2/cmd/compose.runUp({0x2428180?, 0xc0000d0960}, {0x1b0336ddbd8, 0xc000186500}, {0x243f310, _}, {0x0, 0x0, {0x2104070, 0x7}, ...}, ...)
        github.com/docker/compose/v2/cmd/compose/up.go:196 +0x89d
github.com/docker/compose/v2/cmd/compose.upCommand.func2({0x2428180, 0xc0000d0960}, 0xc0000c96b0, {0x33d4128, 0x0, 0x0})
        github.com/docker/compose/v2/cmd/compose/up.go:89 +0x1e5
[...]

PR #10564, opened 2023-05-14
detect terminal is not a `console.File` to avoid a panic

ndeloof:
**What I did**
check conversion to console.File is successful, otherwise fallback to plain text output

**Related issue**
see https://github.com/docker/compose/issues/10560


glours (approved):
LGTM

Issue #10547, opened 2023-05-10
[BUG]  environment variable COMPOSE_PARALLEL_LIMIT has precedence over --parallel commandline parameter


fho:
### Description

When the `COMPOSE_PARALLEL_LIMIT` environment variable is set it takes precedence over the `--parallel` command line parameter.

I would expect that command line parameters have the highest precedence.
Because:
1. [`if  v, ok := os.LookupEnv("COMPOSE_PARALLEL_LIMIT"); ok && !cmd.Flags().Changed("parallel") {`](https://github.com/docker/compose/blob/0c1a691fa5b9b4a68ae00bf9d4b5120a1e2b87b4/cmd/compose/compose.go#L337) reads like the env var should be ignored if flag value was passed.
2. For the `-p` parameter it is also [documented](https://docs.docker.com/compose/environment-variables/envvars/#configure) that the flag has the highest precedence. I would assume the same precedence order is wanted for other parameters.
3. That parameters have the highest precedence is also the common order for other applications.

### Steps To Reproduce

1. compile compose with the following change:
```
diff --git a/cmd/compose/compose.go b/cmd/compose/compose.go
index 019213d8..025a1945 100644
--- a/cmd/compose/compose.go
+++ b/cmd/compose/compose.go
@@ -341,6 +341,7 @@ func RootCommand(streams command.Cli, backend api.Service) *cobra.Command { //no
                                }
                                parallel = i
                        }
+                       fmt.Printf("SET MAX CONCURRENCY to: %d\n", parallel)
                        if parallel > 0 {
                                backend.MaxConcurrency(parallel)
                        }
```
2. Run: `COMPOSE_PARALLEL_LIMIT=3 ~/git/compose/bin/build/docker-compose --parallel 1 config`.
You get the output: `SET MAX CONCURRENCY to: 3`
I would expect that parallel is set to `1`, the value passed to `--parallel`.

### Compose Version

```Text
v2.17.3
```

ndeloof:
that indeed unfortunate. Thanks for the detailed diagnostic
Proposed a fix in https://github.com/docker/compose/pull/10566

PR #10566, opened 2023-05-14
[fix] --parallel has precedence over COMPOSE_PARALLEL_LIMIT

ndeloof:
**What I did**
As we check `--parallel` flag has been set, we have to check the "compose" top-level command, not the current one. Surprisingly, cobra ignores request to check a changed flag for undefined flag, otherwise we would have detected this earlier

also groups `COMPOSE_*` definition as constants as boy scout "clean code" principle

**Related issue**
fixes https://github.com/docker/compose/issues/10547

### DIFF
--- a/cmd/compose/compose.go
+++ b/cmd/compose/compose.go
@@ -47,6 +47,19 @@ import (
 	"github.com/docker/compose/v2/pkg/utils"
 )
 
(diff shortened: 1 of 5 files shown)


glours (approved):
LGTM

PR #10599, opened 2023-05-23
only watch configured paths

ndeloof:
**What I did**
x-develop has got a bit of a wandering eye - It watches the whole build context instead of just the watch paths

**Related issue**
(https://docker.atlassian.net/browse/ENV-192)


milas (approved):
😅

PR #10602, opened 2023-05-23
[fix] `up` should not silently ignore missing depends_on service

ndeloof:
**What I did**
`docker compose up` used to silently ignore any dependency service that is missing in the project model as disabled service. This fix detect this and report a diagnostic message to the user

 ```
$ docker compose up
[+] Building 0.0s (0/0)                                                                                                                        
service mssql is required by app but is disabled. Can be enabled by profiles [dev]
```

**Related issue**
fixes https://github.com/docker/compose/issues/9575

require https://github.com/compose-spec/compose-go/pull/409

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,7 +5,7 @@ go 1.20
 require (
 	github.com/AlecAivazis/survey/v2 v2.3.6
 	github.com/buger/goterm v1.0.4
-	github.com/compose-spec/compose-go v1.13.5
+	github.com/compose-spec/compose-go v1.14.0
 	github.com/containerd/console v1.0.3
 	github.com/containerd/containerd v1.6.21
 	github.com/cucumber/godog v0.0.0-00010101000000-000000000000 // replaced; see replace for the actual version used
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Review by glours — APPROVED** [review]
👍

**Comment by glours** [comment]
You need to update some tests:
```
--- FAIL: TestStartStopWithDependencies/Up_no-deps_links (1.24s)
        start_stop_test.go:118: Running command: docker compose --project-name e2e-start-stop-with-dependencies down
        start_stop_test.go:119: Running command: docker compose -f ./fixtures/links/compose.yaml --project-name e2e-start-stop-with-dependencies up --no-deps -d foo
        start_stop_test.go:119: assertion failed: 
            Command:  docker compose -f ./fixtures/links/compose.yaml --project-name e2e-start-stop-with-dependencies up --no-deps -d foo
            ExitCode: 1
            Error:    exit status 1
            Stdout:   
            Stderr:   service bar is required by foo but is disabled. Can be enabled by profiles []
            
            
            Failures:
            ExitCode was 1 expected 0
            Expected no error
    start_stop_test.go:128: Running command: docker compose --project-name e2e-start-stop-with-dependencies rm -fsv
 ```

ndeloof:
What's weird here is that we had a test to check error is ignored 🤨

ndeloof:
this is not just about fixing a test, `--no-deps` makes it explicit we don't want this error to be a blocker. Will need to revisit my fix

milas (approved):

allanlewis:
This change appears to have caused #10751, which is breaking my CI workflows. I'm working around it by downgrading to v2.18.1.

PR #10604, opened 2023-05-23
update docs to reflect dry run mode is feature complete

aevesdocker:
**What I did**

Updated the docs to reflect the incoming change with DD4.20 where Dry Run mode is now feature complete


ndeloof (approved):

glours (approved):
LGTM

PR #10611, opened 2023-05-24
containerd-enabled engine return "Exists" status for already present layers

ndeloof:
**What I did**
Accept "Exists" as a pull JSONMessage.Status for already pulled layer. This is what we get with containerd-enabled docker engine

**How to test it**

enable container image store in Docker Desktop
then run `docker compose pull` (twice). Progress UI should show completion state


StefanScherer (approved):

ndeloof:
maybe obsolete after https://github.com/moby/moby/pull/45601

laurazard:
Looks like we can close this since the change has been merged/backported in moby/moby.

PR #10612, opened 2023-05-24
detect network conflict as name is not guaranteed to be unique

ndeloof:
**What I did**
this partially revisit https://github.com/docker/compose/pull/9585 to better detect misuse of network and name conflicts
- report error to user when ambiguous name is used with multiple networks found
- report error when network doesn't have the expected labels set
- manage creation conflict
- once target network has been identified, mutate the project to use ID as name, so that other API calls don't become ambiguous

also created dedicated func `resolveExternalNetwork` to keep code reasonably readable


pkg/compose/create.go
```diff
@@ -1088,87 +1093,128 @@ func (s *composeService) ensureNetwork(ctx context.Context, n types.NetworkConfi
...
-					IPRange:    pool.IPRange,
-					Gateway:    pool.Gateway,
-					AuxAddress: pool.AuxiliaryAddresses,
-				})
-			}
-			ipam = &network.IPAM{
-				Driver: n.Ipam.Driver,
-				Config: config,
-			}
+		if net.Labels[api.ProjectLabel] == n.Labels[api.ProjectLabel] &&
```
ndeloof, line 1103: we could invoke NetworkList with a projectFilter and networkFilter (same as `down`)  but not doing so allows to catch this scenario were a network with same name exists but doesn't have label, and `CheckDuplicate: true` will prevent us to create another one.

pkg/compose/convergence.go
```diff
@@ -513,6 +513,16 @@ func (s *composeService) createMobyContainer(ctx context.Context,
 		}
 		plat = &p
 	}
+
+	links, err := s.getLinks(ctx, project.Name, service, number)
+	if err != nil {
+		return created, err
+	}
+	for k, s := range networkingConfig.EndpointsConfig {
+		s.Links = links
```
ndeloof, line 522: using NetworkConnect with network name this wasn't required. Sounds like some engine weirdness, anyway this is consistent we define NetworkConfig the same way we later connect to network

pkg/compose/down.go
```diff
@@ -148,32 +148,42 @@ func (s *composeService) ensureImagesDown(ctx context.Context, project *types.Pr
...
 	})
 	if err != nil {
-		return errors.Wrapf(err, fmt.Sprintf("failed to inspect network %s", name))
+		return errors.Wrapf(err, "failed to list networks")
 	}
+
+	// NetworkList API doesn't return the exact name match, so we can retrieve more than one network with a request
+	networks = utils.Filter(networks, func(net moby.NetworkResource) bool {
+		return net.Name == name
+	})
```
milas, line 185: This is unnecessary - `NetworkList` with a name filter works reliably, it's `NetworkInspect` that does prefix matching

pkg/compose/down.go
```diff
@@ -148,32 +148,42 @@ func (s *composeService) ensureImagesDown(ctx context.Context, project *types.Pr
...
 	// an error along the lines of:
 	// 	failed to remove network test: Error response from daemon: network test is ambiguous (2 matches found based on name)
 	// as a workaround here, the delete is done by ID after doing a list using
 	// the name as a filter (99.9% of the time this will return a single result)
 	networks, err := s.apiClient().NetworkList(ctx, moby.NetworkListOptions{
-		Filters: filters.NewArgs(filters.Arg("name", name)),
+		Filters: filters.NewArgs(
+			projectFilter(projectName),
+			networkFilter(networkName),
+			filters.Arg("name", name)),
```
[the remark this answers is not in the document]
milas: Oh actually, shouldn't we just get rid of this filter entirely if we're using the `com.compose.network` label?

milas (commented):
A nice improvement!

Sorry to leave so many comments, let me know if anything doesn't make sense. I'm also not really adamant about anything here, just trying to be extra cautious!

pkg/compose/create.go
```diff
@@ -1088,87 +1093,131 @@ func (s *composeService) ensureNetwork(ctx context.Context, n types.NetworkConfi
...
+	if err != nil {
+		if errdefs.IsConflict(err) {
+			// Maybe another execution of `docker compose up|run` created same network
+			// let's retry
+			return s.ensureNetwork(ctx, n)
 		}
+		w.Event(progress.ErrorEvent(networkEventName))
+		return errors.Wrapf(err, "failed to create network %s", n.Name)
+	}
+	n.Name = created.ID
```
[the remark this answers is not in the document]
ndeloof: the `NetworkConnect` we rely on expect an `id` parameter to be ["Network ID or name"](see https://docs.docker.com/engine/api/v1.42/#tag/Network/operation/NetworkConnect) - there's actually many places where engine accepts both, so the [GetContainer](https://github.com/moby/moby/blob/master/daemon/container.go#L28-L35) and [FindNetwork](https://github.com/moby/moby/tree/master/daemon/network.go#L57-L62) having `idOrName` parameter and managing ambiguous references. 
I guess this design decision makes the CLI easier to implement, as there's no need to first resolve a name then invoke operation by ID, but this obviously leaks into compose model as `Name` - attribute should be `NameOfID` as well.

pkg/compose/down.go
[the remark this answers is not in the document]
ndeloof: I discovered moby overrides NetworkConfig set on `ContainerCreate` and ALWAYS use network.Name as key, which prevent `ContainerStart` to succeed when name is ambiguous. So using ID as Name won't help 😓

pkg/compose/create.go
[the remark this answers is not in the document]
ndeloof: I discovered moby overrides NetworkConfig set on `ContainerCreate` and ALWAYS use network.Name as key, which prevent `ContainerStart` to succeed when name is ambiguous. So using ID as Name won't help 😓

pkg/compose/create.go
```diff
@@ -1077,98 +1078,168 @@ func getAliases(s types.ServiceConfig, c *types.ServiceNetworkConfig) []string {
...
+
+	// First, try to find a unique network matching by name or ID
+	inspect, err := s.apiClient().NetworkInspect(ctx, n.Name, moby.NetworkInspectOptions{})
+	if err == nil {
+		// NetworkInspect will match on ID prefix, so double check we get the expected one
+		// as looking for network named `db` we could erroneously matched network ID `db9086999caf`
+		if inspect.Name == n.Name || inspect.ID == n.Name {
+			if inspect.Labels[api.ProjectLabel] != expectedProjectLabel {
+				logrus.Warnf("a network with name %s exists but was not created by compose.\n"+
+					"Set `external: true` to use a network you created", n.Name)
```
glours, line 1107:
```suggestion
					"Set `external: true` to use an already existing network", n.Name)
```
WDYT?

pkg/compose/create.go
```diff
@@ -1077,98 +1078,168 @@ func getAliases(s types.ServiceConfig, c *types.ServiceNetworkConfig) []string {
...
+	}
 
-		if n.Ipam.Driver != "" || len(n.Ipam.Config) > 0 {
-			createOpts.IPAM = &network.IPAM{}
+	// we could have set NetworkList with a projectFilter and networkFilter but not doing so allows to catch this
+	// scenario were a network with same name exists but doesn't have label, and use of `CheckDuplicate: true`
+	// prevents to create another one.
+	if len(networks) > 0 {
+		logrus.Warnf("a network with name %s exists but was not created by compose.\n"+
+			"Set `external: true` to use a network you created", n.Name)
```
glours, line 1142:
```suggestion
		logrus.Warnf("a network with name %s exists but was not created by compose.\n"+
			"Set `external: true` to use an already existing network", n.Name)
```

glours (approved):
I just added just few non blocking questions

milas (approved):

jauderho:
I'm not 100% sure if this is the patch that broke things but things were working on version 2.18.1

So I have a compose.yml file

```

services:
  dnscrypt-proxy:
    container_name: dnscrypt-proxy
    #build: .
    image: ghcr.io/jauderho/dnscrypt-proxy:2.1.4@sha256:1baca84db9c59a0f0d783f84b9f52a7ef4d08b8fddf9a0357cf319453285e1df
    ports:
      - "5400:5400/tcp"
      - "5400:5400/udp"
    # Volumes store your data between container upgrades
    volumes:
      - ./config/:/config/
    #dns:
    #  - 127.0.0.1
    #  - 192.168.1.10
    #cap_add:
      #- NET_ADMIN
    #  - NET_BIND_SERVICE
    restart: unless-stopped
    labels:
      - "com.centurylinklabs.watchtower.enable=true"

networks:
  default:
    name: dns
```

The `dns` network was created using `docker network create dns`. Prior to 2.19.0, this worked with no issue but upon switching to 2.19.0. This is what I get now. 

```
[+] Stopping 1/0
 ✔ Container dnscrypt-proxy  Stopped                                                                                                                                                         0.0s 
WARN[0000] a network with name dns exists but was not created by compose.
Set `external: true` to use an existing network 
network dns was found but has incorrect label com.docker.compose.network set to ""
```

This corresponds to line# 1110 of pkg/compose/create.go. Not quite sure how to fix this.

ndeloof:
isn't warning explicit enough?
To have compose use a network you created on your own, you _should_ use `external: true`

```yaml
networks:
  default:
    name: dns
    external: true
```

jauderho:
Thanks. This fixed it for me.

HannesJo0139:
> isn't warning explicit enough? To have compose use a network you created on your own, you _should_ use `external: true`
> 
> ```yaml
> networks:
>   default:
>     name: dns
>     external: true
> ```

To be honest, it's not. The message tells me the label was not expected. I don't know what label it was expecting nor why it was. I also don't know why an unexpected label would be caused by a missing `external: true`.

PR #10614, opened 2023-05-24
progress: remove errant import

milas:
**What I did**
Just write the warning to the passed in writer. The function being used was coming from `cfssl`'s log package, which was presumably the result of auto-import being slightly too aggressive.

(Note: `cfssl` is still an indirect dependency after this.)

**Related issue**
n/a

cat in a husky outfit next to two huskies

### DIFF
--- a/pkg/progress/writer.go
+++ b/pkg/progress/writer.go
@@ -21,12 +21,12 @@ import (
 	"io"
 	"sync"
 
-	"github.com/cloudflare/cfssl/log"
-	"github.com/docker/compose/v2/pkg/api"
-
 	"github.com/containerd/console"
 	"github.com/moby/term"
+	"github.com/sirupsen/logrus"
 	"golang.org/x/sync/errgroup"
+
+	"github.com/docker/compose/v2/pkg/api"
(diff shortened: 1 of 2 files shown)


glours (approved):

pkg/progress/writer.go
```diff
@@ -125,7 +125,7 @@ func NewWriter(ctx context.Context, out io.Writer, progressTitle string) (Writer
 	}
 	if Mode == ModeTTY {
 		if !isConsole {
-			log.Warning("Terminal is not a POSIX console")
+			_, _ = fmt.Fprintln(out, "Terminal is not a POSIX console")
```
ndeloof, line 128: we've been using `logrus.Warning` in other places, we should do the same here

thaJeztah:
cfssl is the worst; it had a _TON_ of dependencies; I managed to get rid of a bunch of them, but don't think they did a new release still; https://github.com/cloudflare/cfssl/pulls?q=is%3Apr+author%3AthaJeztah

glours (approved):
👍

Issue #10618, opened 2023-05-25
[BUG] docker compose -p <Project> logs -f not showing logs of later started containers


makon:
### Description

`docker compose -p <project> logs -f` shows only log entries of already running containers. 
It does not show any entries of later started containers in the same `--project-name`.

Doing it without the `--project-name` flag with `docker compose logs -f` it works as expected. 
It even works with the `name` attribute in the `docker-compose.yml`.
It worked with Docker Compose v1.

docker-compose.yml:
```yml
# name: test

services:
  hello:
    image: ubuntu
    command:
      [
        "/bin/bash",
        "-c",
        "while sleep 1; do echo 'Hello'; done"
      ]

  world:
    image: ubuntu
    command:
      [
        "/bin/bash",
        "-c",
        "while sleep 1; do echo 'World'; done"
      ]
```

### Steps To Reproduce

1. use the `docker-compose.yml`
2. run in a terminal `docker compose -p test up -d hello`
3. run in another terminal `docker compose -p test logs -f`
4. wait 10 seconds
5. run in a terminal `docker compose -p test up -d world`
6. the logging terminal shows only `hello` entries and no `world` entries

### Compose Version

```Text
Docker Compose version v2.18.1
```

### Docker Environment

```Text
Client: Docker Engine - Community
 Version:    24.0.1
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.10.4
    Path:     /usr/libexec/docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.18.1
    Path:     /usr/libexec/docker/cli-plugins/docker-compose

Server:
 Containers: 18
  Running: 7
  Paused: 0
  Stopped: 11
 Images: 58
 Server Version: 24.0.1
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: runc io.containerd.runc.v2
 Default Runtime: runc
 Init Binary: docker-init
[...]

PR #10622, opened 2023-05-25
fix `compose -p x logs -f` detect new services started after command

ndeloof:
when we run `compose --project-name logs` we SHOULD detect new services being started _after_ the initial model has been built from actual ressources. For this purpose, 
- options.Services filter is only set when project is not nil
- `create` events` adds container to the watched list

added an e2e test to check this behaves as expected

**Related issue**
fixes https://github.com/docker/compose/issues/10618

### DIFF
--- a/pkg/compose/events.go
+++ b/pkg/compose/events.go
@@ -25,7 +25,6 @@ import (
 	"github.com/docker/docker/api/types/filters"
 
 	"github.com/docker/compose/v2/pkg/api"
-
 	"github.com/docker/compose/v2/pkg/utils"
 )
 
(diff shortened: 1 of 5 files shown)


ndeloof:
e2e test is failing due to https://github.com/docker/compose/pull/10621 : scale up recreate container, which takes 10s

milas (approved):
LGTM!

I did a bunch of combinations of `up` & `down` across different terminals, targeting specific services/adjusting scale/etc with an open `compose -p $NAME logs -f` session and everything worked great ✨

glours (approved):
LGTM

PR #10623, opened 2023-05-25
Fix typo in warning about existing volume

jfly:
Previously, this was telling us "but was not created for project [project-it-was-created-for]", which is wrong. I opted to make the message super explicit and print both the actual project and the expected project.

**What I did**

**Related issue**
I couldn't find an existing issue for this.

This surprised koala has been my phone background ever since I got my first smartphone:

image

### DIFF
--- a/pkg/compose/create.go
+++ b/pkg/compose/create.go
@@ -1194,7 +1194,7 @@ func (s *composeService) ensureVolume(ctx context.Context, volume types.VolumeCo
 		logrus.Warnf("volume %q already exists but was not created by Docker Compose. Use `external: true` to use an existing volume", volume.Name)
 	}
 	if ok && p != project {
-		logrus.Warnf("volume %q already exists but was not created for project %q. Use `external: true` to use an existing volume", volume.Name, p)
+		logrus.Warnf("volume %q already exists but was created for project %q (expected %q). Use `external: true` to use an existing volume", volume.Name, p, project)
 	}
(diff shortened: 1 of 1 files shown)


pkg/compose/create.go
```diff
@@ -1194,7 +1194,7 @@ func (s *composeService) ensureVolume(ctx context.Context, volume types.VolumeCo
 		logrus.Warnf("volume %q already exists but was not created by Docker Compose. Use `external: true` to use an existing volume", volume.Name)
 	}
 	if ok && p != project {
-		logrus.Warnf("volume %q already exists but was not created for project %q. Use `external: true` to use an existing volume", volume.Name, p)
+		logrus.Warnf("volume %q already exists but was created for project %q (expected %q). Use `external: true` to use an existing volume", volume.Name, p, project)
```
jfly, line 1197: @milas, I could use some help here with the code coverage complaint. I checked, and none of the existing tests seem to execute this `ensureVolume` function at all, so there isn't anything easy for me to copy paste to start writing a test to cover this code. I don't know much of anything about `go` or this codebase, so I could use some pretty hand-holdy pointers if you want me to try to do that.

glours: @jfly our code coverage configuration doesn't work correctly ATM, so don't worry, your PR won't be blocked by this check

glours (approved):
LGTM

PR #10631, opened 2023-05-30
bump buildx and use confutil.ConfigDir

ndeloof:
**What I did**
bump buildx to latest release
use `confutil.ConfigDir` to better align with buildx codebase
only return digest for (single) image we have built (multi-value return is a legacy from the time we ran buildx for all images at once)


milas (approved):
LGTM!

laurazard (approved):

Issue #10586, opened 2023-05-19
[BUG] compose run doesn't forward signals


raylu:
### Description

as mentioned in https://github.com/docker/compose/issues/8549#issuecomment-1540937003, this doesn't seem to be working for `docker compose` (v2). the behavior works fine for `docker-compose` (v1), though

### Steps To Reproduce

I ran @ndeloof's demo (from https://github.com/docker/compose/issues/8549#issuecomment-1123261668). had to comment some lines first:
```diff
$ git diff
diff --git a/demo/Dockerfile b/demo/Dockerfile
index a4058f8..6b4998c 100644
--- a/demo/Dockerfile
+++ b/demo/Dockerfile
@@ -3,9 +3,9 @@ FROM golang:1.18-alpine
 
 RUN apk add --no-cache openssh-client git
 RUN mkdir -p -m 0700 ~/.ssh && ssh-keyscan github.com >> ~/.ssh/known_hosts
-RUN --mount=type=ssh git clone git@github.com:ndeloof/toolbox.git
+#RUN --mount=type=ssh git clone git@github.com:ndeloof/toolbox.git
 
-RUN --mount=type=secret,id=mysecret cat /run/secrets/mysecret
+#RUN --mount=type=secret,id=mysecret cat /run/secrets/mysecret
 
 WORKDIR /app
 EXPOSE 5050
```
```
$ docker compose run --rm demo
press Ctrl+C to stop
```
in another shell,
```
$ ps x | grep compose
 8584 s000  S+     0:00.04 docker compose run --rm demo
 8585 s000  S+     0:00.02 /Applications/Docker.app/Contents/Resources/bin/com.docker.cli compose run --rm demo
 8587 s000  S+     0:00.03 /Users/raylu/.docker/cli-plugins/docker-compose compose run --rm demo

$ ps o pid,ppid,pgid | grep 858
 8584  9824  8584
 8585  8584  8584
 8587  8585  8584
# we see that all 3 processes are in the 8584 process group and we can signal them all at once

$ kill -- -8584
# nothing happens in the first shell

$ kill -- -8584

$ kill -- -8584
# ERRO[0123] got 3 SIGTERM/SIGINTs, forcing shutdown

$ docker ps
CONTAINER ID   IMAGE       COMMAND                  CREATED          STATUS        PORTS    NAMES
[...]

PR #10633, opened 2023-05-30
forward signal to container

ndeloof:
**What I did**
as compose run a container, forward signals to container the same way `docker run` does
we have to because `RunStart` only doest it for container without a TTY (https://github.com/docker/cli/blob/master/cli/command/container/start.go#L91)

As container is ran interactive from a terminal (`--tty=true`), `RunStart` uses `setRawTerminal` and doing so user to hi `Ctrl+C` just directly send those keys to target process, not `compose` command itself.
But when signal is sent to compose by a `kill` command it has to handle this and forward signal explicitly, just like `RunStart` does when container has no TTY.

**Related issue**
fixes https://github.com/docker/compose/issues/10586

### DIFF
--- a/pkg/compose/run.go
+++ b/pkg/compose/run.go
@@ -19,6 +19,8 @@ package compose
 import (
 	"context"
 	"fmt"
+	"os"
+	"os/signal"
 
 	"github.com/compose-spec/compose-go/types"
 	"github.com/docker/cli/cli"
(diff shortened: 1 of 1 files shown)


glours (approved):
LGTM

g0t4:
Would this fix allow for `Ctrl+R` to be sent to a container started with `docker compose up`? I'm trying to find if there are any open issues with regards to `stdin` and `docker compose` managed containers.

ndeloof:
@g0t4 `docker compose up` is not interactive, it doesn't attach to a specific container, and as such won't forward signals to service(s). You can use `docker compose kill -s [SIGNAL] SERVICE` for this purpose

kuro337:
@ndeloof Hello sir, is this still true?

If a process is launched with  docker compose up  - there is no way for my image to be sent the shutdown signal?

Are there any ways to achieve it?

ndeloof:
Ctrl+C or `down` command will send a SIGINT signal to your container, you can tweak signal being sent and timeout waiting for termination in your compose.yaml file, see https://github.com/compose-spec/compose-spec/blob/master/05-services.md#stop_grace_period

Issue #9586, opened 2022-06-22
"Custom build outputs" support


tpict:
Are there any plans to support Docker's [custom build outputs](https://docs.docker.com/engine/reference/commandline/build/#custom-build-outputs), either via a flag to `docker-compose build` or in the Compose file? Thanks!

ndeloof:
How would  `docker compose up` handle this, if you configure build output to write image to a local tar file, to create a container for service after a successful build ?
At most, such an option could only be available for `docker compose build`. is this what you are asking for ?
Maybe `docker compose build --output type=tar,dest=some_path` could generate image files as some_path/<service>.tar`?

ndeloof:
created a draft PR to add support for custom build outputs, please give it a try and let me know if this covers your need
https://github.com/docker/compose/pull/10642

Aarbel:
It would be extremely useful 

<img width="884" alt="image" src="https://github.com/docker/compose/assets/25119847/f86bf3e3-2b49-43e8-938a-e6b18e498039">

Docker outputs on docker-compose doesn't show all necessary informations.
In this cas, i absolutely don't know which container is "eca8ccf1c84e"

ndeloof:
@Aarbel your comment is unrelated to custom build output, feel free to open a dedicated issue (and include `docker inspect eca...` for diagnostic

ndeloof:
Closing as obsolete. Requested feature is unclear how this would benefit docker compose usage

Ryanf55:
Here is my use case:

For [Custom Build Output](https://docs.docker.com/reference/cli/docker/image/build/#output) here, it is useful for me to take the build logs and test artifacts created during docker build and expose them back to the host so they can be uploaded into Gitlab as a CI artifact. Currently, I create these with `docker build` command. I would like to migrate everything to compose, but this lack of supports for outputs makes the migration impossible.

gegles:
Hello, I just stumbled upon this and yes, we would really need this as well... 
We leverage docker compose extensively for dev, test, build and prod... We especially use a complex multi stage build.

Having docker compose helps us encapsulating what would otherwise be pretty complex docker build or run commands..

One of the stage of the build is to to create official binaries (RPM/DEB packages). We can even build multiple archs (`amd64`, `arm64`, `s390x` and `ppc64le`) thanks to the `platform` field... 

But, sadly, custom build outputs (`--output` arg) is not supported by docker compose and thus we end up with a pretty complicated build command:
```
[...]

PR #10642, opened 2023-06-01
introduce support for custom build outputs

ndeloof:
**What I did**
introduce `--output` flag on `build` command to export built artifact with alternate exporter.
AFAICT having this as part of the compose.yaml service definition doesn't make much sense as `compose up` scenario require image to be exported to container runtime local store. On the other hand, maybe there's room for new use-cases?

**Related issue**
fixes https://github.com/docker/compose/issues/9586

### DIFF
--- a/cmd/compose/build.go
+++ b/cmd/compose/build.go
@@ -26,6 +26,7 @@ import (
 	"github.com/compose-spec/compose-go/loader"
 	"github.com/compose-spec/compose-go/types"
 	buildx "github.com/docker/buildx/util/progress"
+	"github.com/docker/cli/cli/command"
 	cliopts "github.com/docker/cli/opts"
 	"github.com/docker/compose/v2/pkg/progress"
 	"github.com/docker/compose/v2/pkg/utils"
@@ -45,6 +46,7 @@ type buildOptions struct {
 	noCache  bool
 	memory   cliopts.MemBytes
 	ssh      string
+	outputs  []string
 }
 
 func (opts buildOptions) toAPIBuildOptions(services []string) (api.BuildOptions, error) {
@@ -58,6 +60,7 @@ func (opts buildOptions) toAPIBuildOptions(services []string) (api.BuildOptions,
 	}
 
 	return api.BuildOptions{
+		Outputs:  opts.outputs,
 		Pull:     opts.pull,
 		Push:     opts.push,
 		Progress: opts.progress,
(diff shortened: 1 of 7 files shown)


Ryanf55:
What was the reason this was closed. I use build outputs to take unit tests results and report them back to CI. Without this support, looks like I cannot use compose in CI for build/test.

israelglar:
> **What I did** introduce `--output` flag on `build` command to export built artifact with alternate exporter. AFAICT having this as part of the compose.yaml service definition doesn't make much sense as `compose up` scenario require image to be exported to container runtime local store. On the other hand, maybe there's room for new use-cases?
> 
> **Related issue** fixes #9586
> 
> **(not mandatory) A picture of a cute animal, if possible in relation to what you did**

I believe this should be reopened and merged. Maybe your reason for closing was that there was no "new use-cases", but as per @Ryanf55  and now also me, there are use cases where this options would be really helpful.

On my case I want our developer team and CI to use cache. For that I'm story my cache in an ECR using [cache_to:type=registry](https://docs.docker.com/build/cache/backends/registry/)  to be able to do that I need to use the [docker-container driver](https://docs.docker.com/build/builders/drivers/) .

This works for services that don't depend on anything else, but breaks when in use with [additional_contexts](https://docs.docker.com/reference/compose-file/build/#additional_contexts) referencing another service in compose. 
Because even tho the additional services are built and are locally present but not pushed, when the main service wants to load their metadata it looks for it in the registry but since it wasn't pushed yet (this is related to an open issue in [docker-container where it doesn't load local images](https://github.com/moby/buildkit/issues/2343) )

To summarise: 
 - I need to user `docker-container` to be able to export cache
 - `docker-container has a bug where it doesnt load local images
 
The really easy way to fix this is to push images as soon as they are built. In my testing if I use the buildx option `--output type=registry`  [docs here](https://docs.docker.com/reference/cli/docker/buildx/build/#registry)
 it fixes this problem because it pushes after building.
 
 That together with the open issue #9586 where many other people also want this feature in, I think we should reopen it and merge.

israelglar:
Hi!
Thanks for answering so quickly, it always good to have project authors/contributors that are active and on top of stuff!

You're right. I was able to get this exact issue solved by using bake with the [x-bake key](https://docs.docker.com/build/bake/compose-file/#extension-field-with-x-bake)

In the future I can see how having this tight integration and delegating build to bake would be really good.

Thanks!

Issue #10647, opened 2023-06-02
[BUG] using "compose" as --project-name makes compose binary error


xAt0mZ:
### Description

When using `compose` as project name, with `--project-name` option, compose binary errors.

**Actual**
```
>$ ./docker-compose --project-name test down --remove-orphans
Warning: No resource found to remove for project "test".

>$ ./docker-compose --project-name compose down --remove-orphans
unknown flag: --remove-orphans
```

**Expected**
```
>$ ./docker-compose --project-name test down --remove-orphans
Warning: No resource found to remove for project "test".

>$ ./docker-compose --project-name compose down --remove-orphans
Warning: No resource found to remove for project "compose".
```

Tested with multiple "docker related" keywords and only `compose` produces this error.

### Compose Version

```Text
Tested all of the following versions

Docker Compose version 2.17.3
Docker Compose version 2.17.2
Docker Compose version 2.17.0
```

### Docker Environment

```Text
Multiple docker versions / environments setups tested so unrelated

Docker versions
23.0.5
19.03.12

Environments
Archlinux bare metal
boot2docker docker-machine
Ubuntu (stable) virtualbox VM
```

ndeloof:
Tried to reproduce without success with latest release:
```
$ docker compose --project-name compose down --remove-orphans
Warning: No resource found to remove for project "compose".
$ docker-compose --project-name compose down --remove-orphans
Warning: No resource found to remove for project "compose".
```

xAt0mZ:
`docker compose` works but it won't solve the issue in my usage scenario.

Also tested with the latest version and I'm still able to reproduce :arrow_down: 
* download the latest version with `wget -O /tmp/docker-compose https://github.com/docker/compose/releases/download/v2.18.1/docker-compose-linux-x86_64`
* make the binary executable `chmod +x  /tmp/docker-compose`
* try `/tmp/docker-compose --project-name compose down --remove-orphans`
  * fail with `unknown flag: --remove-orphans`
* try `/tmp/docker-compose --project-name test down --remove-orphans`
  * ok with `Warning: No resource found to remove for project "test".`

The TLDR of usage scenario: at [portainer/portainer](https://github.com/portainer/portainer) we download & ship compose with our containers (so we can guarantee it is always located somewhere we can use it)
All of our commands are generated to use our shipped binary
We can't rely on `docker compose` because we don't invoke docker binary directly ;we use docker API to communicate with the env (there is no API for compose actions)

[...]

PR #10657, opened 2023-06-05
don't skip `compose` used as project name

ndeloof:
**What I did**
fix support for converting `docker-compose --project-name compose` into a compose v2 command (`docker <space> compose`)
Tried to understand the fix in https://github.com/docker/compose/pull/8985 to skip `compose` argument but doesn't make any sense to me after 18 months. This code is only used if command is `docker-compose ..` but not when used as a CLI plugin with `docker compose ...`, maybe this wasn't the case earlier? @ulyssessouza do you remember?

**Related issue**
fixes https://github.com/docker/compose/issues/10647

### DIFF
--- a/cmd/compatibility/convert.go
+++ b/cmd/compatibility/convert.go
@@ -62,10 +62,6 @@ func Convert(args []string) []string {
 			continue
 		}
 		if len(arg) > 0 && arg[0] != '-' {
-			// not a top-level flag anymore, keep the rest of the command unmodified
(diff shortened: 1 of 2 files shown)


milas (approved):
LGTM, I'm guessing this was supposed to be `if arg[0] == compose.PluginName` originally? But clearly it's not doing anything useful so I am good with eliminating it

glours (approved):

Issue #10658, opened 2023-06-05
[BUG] drops last line of container log on exit


nicks:
### Description

If your container emits error logs without a newline at the end, Compose quietly drops them.

### Steps To Reproduce

compose.yaml:

```
services:
  sleep:
    image: alpine
    entrypoint: sh -c "echo -n exiting; exit 1"
```

run:

```
docker compose up
```

Expected behavior:
Compose to show "exiting" in the logs

Actual behavior:
No logs shown

```
[+] Building 0.0s (0/0)                                                                                                                                
[+] Running 2/2
 ✔ Network scratch_default    Created                                                                                                             0.0s 
 ✔ Container scratch-sleep-1  Created                                                                                                             0.1s 
Attaching to scratch-sleep-1
scratch-sleep-1 exited with code 1
```

### Compose Version

```Text
$ docker compose version
Docker Compose version v2.18.1
```

### Docker Environment

```Text
Client: Docker Engine - Community
 Version:    24.0.2
 Context:    desktop-linux
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.0.1-cloud-driver+005
    Path:     /home/nick/.docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.18.1
    Path:     /usr/lib/docker/cli-plugins/docker-compose
  dev: Docker Dev Environments (Docker Inc.)
    Version:  v0.1.0
    Path:     /usr/lib/docker/cli-plugins/docker-dev
  extension: Manages Docker extensions (Docker Inc.)
    Version:  v0.2.19
    Path:     /usr/lib/docker/cli-plugins/docker-extension
  init: Creates Docker-related starter files for your project (Docker Inc.)
    Version:  v0.1.0-beta.4
    Path:     /usr/lib/docker/cli-plugins/docker-init
  sbom: View the packaged-based Software Bill Of Materials (SBOM) for an image (Anchore Inc.)
    Version:  0.6.0
    Path:     /usr/lib/docker/cli-plugins/docker-sbom
  scan: Docker Scan (Docker Inc.)
    Version:  v0.26.0
    Path:     /usr/lib/docker/cli-plugins/docker-scan
  scout: Command line tool for Docker Scout (Docker Inc.)
    Version:  v0.12.0
    Path:     /usr/lib/docker/cli-plugins/docker-scout
[...]

Issue #8605, opened 2021-09-10
Escape sequences cannot be used inside quoted strings


milas:
## Description of the issue
Docker Compose v2 (`compose-go`) is stripping out escape characters when inside of a quoted string in a command (might happen in other circumstances).

If it's not in a quoted string, they are left as-is. For example, `command: printf 'Hello\tworld\n'` works while `command: sh -c "printf 'Hello\tworld\n'"` does not.

## Context information (for bug reports)
- [x] Using Compose V2 `docker compose ...`
- [ ] Using Compose V1 `docker-compose ...`

**Output of `docker(-)compose version`**
```
Docker Compose version v2.0.0-rc.2
```

**Output of `docker version`**
```
Client:
 Cloud integration: 1.0.17
 Version:           20.10.8
 API version:       1.41
 Go version:        go1.16.6
 Git commit:        3967b7d
 Built:             Fri Jul 30 19:55:20 2021
 OS/Arch:           darwin/amd64
 Context:           default
 Experimental:      true

Server: Docker Engine - Community
 Engine:
  Version:          20.10.8
  API version:      1.41 (minimum version 1.12)
  Go version:       go1.16.6
  Git commit:       75249d8
  Built:            Fri Jul 30 19:52:10 2021
  OS/Arch:          linux/amd64
  Experimental:     true
 containerd:
  Version:          1.4.9
  GitCommit:        e25210fe30a0a703442421b0f60afac609f950a3
 runc:
  Version:          1.0.1
  GitCommit:        v1.0.1-0-g4144b63
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

**Output of `docker-compose config`**
w/ v2 (v2.0.0-rc.2):
```yaml
services:
  hello-world:
    command:
    - sh
    - -c
    - printf 'Hellotworldn'
    image: busybox
    networks:
      default: null
networks:
  default:
    name: escape-repro_default
```
w/ v1 (1.29.2, build 5becea4c):
```yaml
services:
  hello-world:
    command: sh -c "printf 'Hello\tworld\n'"
    image: busybox
version: '3.9'
```

## Steps to reproduce the issue
1. Use an escape sequence (e.g. `\t` or `\n`) inside of a quoted string in `command`

### Observed result
* Docker Compose v2 (`compose-go`) strips out the `\` characters

### Expected result
* Docker Compose leaves the `\` characters as-is

### Stacktrace / full error message
n/a

## Additional information
n/a

[...]

PR #10660, opened 2023-06-06
assume we receive logs by lines and don't ignore those without EOL

ndeloof:
**What I did**
as we attach to containers we receive output as a low-level `Write([]byte)` that we split into lines to prepend container prefix
if container log doesn't include an EOL character we won't dump it to logs
This PR assumes we don't get container output per character but as full lines (which is the case AFAICT) so that we dump logs as we receive them

**Related issue**
fixes https://github.com/docker/compose/issues/10658


glours (approved):
👍

pkg/utils/writer.go
```diff
@@ -43,11 +43,17 @@ func (s *splitWriter) Write(b []byte) (int, error) {
 	for {
 		b = s.buffer.Bytes()
 		index := bytes.Index(b, []byte{'\n'})
-		if index < 0 {
+		if index > 0 {
```
laurazard, line 46: These aren't equivalent, right? I'm looking at this from my phone, so might be wrong (please disregard if so), but with the previous code if there's a `\n` at index `0` we handle it properly, but in the new code if we have a `\n` at index `0` we jump to the same branch as "no \n in this line".

pkg/utils/writer_test.go
```diff
@@ -28,13 +28,9 @@ func TestSplitWriter(t *testing.T) {
...
-	w.Write([]byte("h"))
-	w.Write([]byte("e"))
-	w.Write([]byte("l"))
-	w.Write([]byte("l"))
-	w.Write([]byte("o"))
-	w.Write([]byte("\n"))
-	w.Write([]byte("world!\n"))
-	assert.DeepEqual(t, lines, []string{"hello", "world!"})
+	w.Write([]byte("hello\n"))
+	w.Write([]byte("world\n"))
```
laurazard, line 32: Should we test the new behaviour too by writing a line without an EOL?

milas (approved):
LGTM!

Also 😂 I thought this sounded familiar: https://github.com/docker/compose/issues/8605#issuecomment-917089534

Issue #10686, opened 2023-06-09
[BUG] For `up` command, "pulling" and "Pulled" events should respect `quietPull` flag


Songyu-Wang:
### Description

The function `pullServiceImage` in https://github.com/docker/compose/blob/v2/pkg/compose/pull.go has `quietPull bool` as an argument.

However, the "pulling" and "Pulled" event logger are not wrapped under the argument.
It should follow the same logic as 
```
if !quietPull {
			toPullProgressEvent(service.Name, jm, w)
		}
```

### Steps To Reproduce

N/A

### Compose Version

```Text
v2 head
```

### Docker Environment

```Text
N/A
```

### Anything else?

![image](https://github.com/docker/compose/assets/36670132/86b69c79-58fa-4fa3-bc4c-ced37fe2a8f4)

ndeloof:
`quietPull` allows to skip individual pull progress messages to be displayed, but still we want the user to be aware about the current state of a service, which will be `Pulling` before it can become `Created` then `Started`.

Songyu-Wang:
@ndeloof 
Thanks for replying.
I think `[+] Running 1/15` would be more than enough to indicate `current state of a service`. Pulling and pulled events are juts redundant information.


I am gonna assume most people use quiet pull because they are using docker compose with some kinda SaaS CI and the interactive logs keep getting duplicates every seconds(10th of a second?), For example, https://github.com/docker/compose/actions/runs/5220651859/jobs/9423987875 is not the exact commands, but the symptom is the same.
<img width="939" alt="image" src="https://github.com/docker/compose/assets/36670132/c213f076-688c-4753-a6c6-f818658c954a">

Only remove part of the messages but keep others wont resolve the above pain point. 

I would really hope you can re-consider it

ndeloof:
> Only removing part of the messages but keeping others wont resolve the above pain point.

The only remaining parts are `Pulling` then `Pulled`, just 2 lines when ran in "non-TTY" mode, which are useful to understand the pull completed successfully.
```
$ docker compose --ansi=never up --quiet-pull
 test Pulling 
 test Pulled 
 Container truc-test-1  Creating
 Container truc-test-1  Created
Attaching to truc-test-1
```
I'm closing this as "not planned"

Songyu-Wang:
<img width="939" alt="Screenshot 2023-06-11 at 12 54 11" src="https://github.com/docker/compose/assets/36670132/56bcd923-830b-4373-bc1b-cd0ce2723011">

[...]

PR #10690, opened 2023-06-12
use `--progress` to configure progress UI style

ndeloof:
**What I did**
Introduce `--progress` top-level flag to configure progress UI as auto|tty|plain|quiet|none

This allows to have explicit control on the progress UI without using `--ansi` flag which also has impact on logs rendering

`build` already declared this flag. It is now redundant and kept for backward compatibility but marked hidden

**Related issue**
see https://github.com/docker/compose/issues/10686

### DIFF
--- a/cmd/compose/build.go
+++ b/cmd/compose/build.go
@@ -27,8 +27,7 @@ import (
 	"github.com/compose-spec/compose-go/types"
 	buildx "github.com/docker/buildx/util/progress"
 	cliopts "github.com/docker/cli/opts"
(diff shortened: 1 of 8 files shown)


ndeloof:
@milas made a last minute update : to avoid `--progress` for being documented on each and every command, I made `build --progress` a simple duplicate to set the same variable, and marked flag hidden so we keep backward compatibility but docs are relevant for recommended usage.

glours (approved):
LGTM

Issue #10695, opened 2023-06-12
add --no-attach option to the docker-compose.yml config 


iwm911:
### Description

When creating a docker-compose.yml with verbose containers (like rabbitmq) the screen is swamped with verbose logging from that service.

i am now using the --no-attach to disable that logging to stdout, however it would be wonderful if it would be possible to set it at the service level in the docker-compose.yml file instead of manually write it every run.

thanks!

CoenraadS:
Is there an example of how to use this?

0xMuTwo:
Hi, is there a way to do this now? Even looking at the PR it's not clear how to do this

iwm911:
Just attach this to the compose file

attach: false

0xMuTwo:
This worked perfectly, thanks so much!

Sent via Superhuman ( ***@***.*** )

PR #10700, opened 2023-06-13
support `attach`

ndeloof:
**What I did**
Add support for `attach` in service config, which can be overridden by command line `--attach --no-attach`

**Related issue**
closes https://github.com/docker/compose/issues/10695


cmd/compose/up.go
```diff
@@ -164,15 +163,29 @@ func runUp(ctx context.Context, streams api.Streams, backend api.Service, create
...
+	var attachTo []string
 	if len(upOptions.attach) > 0 {
 		attachTo = upOptions.attach
 	}
 	if upOptions.attachDependencies {
-		attachTo = project.ServiceNames()
+		if err := project.WithServices(attachTo, func(s types.ServiceConfig) error {
+			if s.Attach == nil || *s.Attach {
+				attachTo = append(attachTo, s.Name)
+			}
```
milas, line 174: If `upOptions.attach` is not empty, won't this add duplicates?

cmd/compose/up.go
```diff
@@ -164,15 +163,29 @@ func runUp(ctx context.Context, streams api.Streams, backend api.Service, create
...
+			return nil
+		}); err != nil {
+			return err
+		}
 	}
 	if len(attachTo) == 0 {
-		attachTo = project.ServiceNames()
+		if err := project.WithServices(services, func(s types.ServiceConfig) error {
```
milas, line 181: I think this should now be an `else if` and pass the `IgnoreDependencies` policy for the traversal

i.e.

* If you're attaching dependencies, iterate w/ dependencies
  * if `attachTo` is empty, `WithServices` will iterate all services
  * Otherwise, it'll just do the targeted services & their deps
* Otherwise, only iterate the `services` being up'd (w/o deps)

milas (commented):
I know this is still waiting on the `compose-go` side, will re-review once that's ready

cmd/compose/up.go
ndeloof: indeed. Updated to use a `set`, so we don't have to get into the possible execution paths to guess unicity is preserved :)

glours (approved):

Issue #10626, opened 2023-05-26
[BUG] docker compose alpha watch restart all my services


maxcleme:
### Description

Wanted to play with the new `docker compose alpha watch`, however it seems to restart all my services while I only added `x-watch` directive to one of them.

### Steps To Reproduce

1. Tried with the latest version installed manually
```
docker compose version
Docker Compose version v2.17.3
```

2. `compose.yaml`, `Dockerfile` & `entrypoint.sh`

```
services:
  a:
    build:
      context: .
    x-develop:
      watch:
      - action: rebuild
        path: hello.txt
  b:
    build:
      context: .
  c:
    build:
      context: .
```

```
FROM alpine:latest
WORKDIR /app
COPY entrypoint.sh .
COPY hello.txt .

STOPSIGNAL 9
ENTRYPOINT ["sh", "-c", "/app/entrypoint.sh"]
```
```
DATE=`date`
while true
do
  echo $DATE `cat hello.txt`
  sleep 1 
done
```

3. Starting up the stack & the watch

```
compose-watch-test-a-1  | Fri May 26 15:36:21 UTC 2023 hello
compose-watch-test-b-1  | Fri May 26 15:36:21 UTC 2023 hello
compose-watch-test-c-1  | Fri May 26 15:36:21 UTC 2023 hello
compose-watch-test-a-1  | Fri May 26 15:36:21 UTC 2023 hello
compose-watch-test-c-1  | Fri May 26 15:36:21 UTC 2023 hello
compose-watch-test-b-1  | Fri May 26 15:36:21 UTC 2023 hello
```

```
docker compose alpha watch
watch command is EXPERIMENTAL
watching /Users/maxcleme/workspace/maxcleme/compose-watch-test
watching /Users/maxcleme/workspace/maxcleme/compose-watch-test
watching /Users/maxcleme/workspace/maxcleme/compose-watch-test
```

4. Editing `hello.txt`

Rebuild is indeed triggered, but all my services got replaced

```
compose-watch-test-a-1  | Fri May 26 15:43:20 UTC 2023 world
compose-watch-test-c-1  | Fri May 26 15:43:20 UTC 2023 world
compose-watch-test-b-1  | Fri May 26 15:43:20 UTC 2023 world
compose-watch-test-a-1  | Fri May 26 15:43:20 UTC 2023 world
compose-watch-test-c-1  | Fri May 26 15:43:20 UTC 2023 world
```

### Compose Version

```Text
Docker Compose version v2.17.3
[...]

Issue #10570, opened 2023-05-15
[BUG] Alpha watch should not rebuild when no x-develop option set


simonmallet:
### Description

First of all, thank you for developing this amazing feature! It will be great to support our mac + windows dev workstations without using volumes!

I have a docker-compose file with 12 services in it.
Most of them will use the watch feature, but not all of them.

For instance I have 2 services that do not mention "x-develop -> watch" in the service, yet when some files get updated, they are rebuilt automatically.

```
Rebuilding indexer-typescript, azfunc-all-v4-5 after changes were detected:

/Users/simonmallet/Sites/cbc/CMS-clone/typescript/current/dist/libs/rcgraph/src/lib/interfaces/withShortStoryTileTemplate.d.ts
/Users/simonmallet/Sites/cbc/CMS-clone/typescript/current/dist/libs/rcgraph/src/lib/valueObjects/rightType.js
```

My service definition: no mention of x-develop -> watch
```
azfunc-all-v4-5:
    build:
      context: ../typescript/current
      dockerfile: apps/content-notification/docker-local/Dockerfile.azfunc-all
    env_file:
      - .env
      - ../typescript/current/.env.local
    environment:
      environment: ${Environment}
      DeployedEnvironmentWithIndex: ${EnvironmentWithIndex}
    depends_on:
      - "mongo"
      - "elasticsearch"
```

I'd like to be able to use `docker-compose alpha watch` command and that it does **NOT** watch apps that do not mention the `x-develop` option. As an alternative, if I could specify x-develop -> watch and use a pre-defined option : `none` so that it knows not to watch or rebuild anything for this service.

When I build an app it writes a lot of files to the dist folder then the docker-compose alpha watch command goes in frenzy rebuilding everything forever using all the disk space.

Thanks!

### Steps To Reproduce

1. Have a docker-compose.yaml file with 2 projects where 1 of them has a x-develop watch and not the other:
```
scoop-ui-v3:
    image: scoop-ui-v3
    build:
      context: ../typescript/scoop-ui-v3
      dockerfile: Dockerfile.scoop-ui-v3-dev
    x-develop:
      watch:
         - action: sync
           path: "../typescript/scoop-ui-v3/"
           target: "/app/"
           ignore:
            - node_modules/
    command: yarn start

azfunc-all-v4-5:
    build:
      context: ../typescript/current
      dockerfile: apps/content-notification/docker-local/Dockerfile.azfunc-all
    env_file:
      - .env
[...]

PR #10710, opened 2023-06-14
don't apply "rebuild" watch strategy by default

ndeloof:
**What I did**
stop automatically creating a "rebuild" watch config for services without an explicit develop section

**Related issue**
fixes https://github.com/docker/compose/issues/10626
fixes https://github.com/docker/compose/issues/10570
https://docker.atlassian.net/browse/ENV-213


glours (approved):
yes

milas (approved):
LGTM

PR #10717, opened 2023-06-15
otel: fix initialization / error-handling

milas:
**What I did**
* If there's no `otel` key (or the value is `null`) in the config, don't return an error
* Propagate error from the exporter instead of panicking

baby mountain lion cubs


glours (approved):
LGTM

ndeloof (approved):

Issue #10718, opened 2023-06-15
[BUG] Compose hangs indefinitely when build of a dependency dies


Regrau:
### Description

The docker compose output is stuck and can't be closed with Ctrl+C when the build of a dependency container fails.
Killing the docker compose process from an outside shell seems to be the only workaround

### Steps To Reproduce

For convenience sake, here's a repos with both files

https://github.com/Regrau/docker-compose-bug-report

```bash
git clone https://github.com/Regrau/docker-compose-bug-report && cd docker-compose-bug-report
docker compose up -d --build
```

### docker-compose.yml
```yaml
services:
  web:
    image: busybox
    command: tail -f /dev/null
    depends_on:
      - db
  db:
    container_name: test
    build:
      dockerfile: Dockerfile
      context: .
```

### Dockerfile
```Dockerfile
FROM alpine:3.17

RUN exit 1

CMD [ "echo", "hello" ]
```

### Compose Version

```Text
Docker Compose version 2.18.1
```

### Docker Environment

```Text
Client: Docker Engine - Community
 Version:    24.0.2
 Context:    colima
 Debug Mode: false
 Plugins:
  compose: Docker Compose (Docker Inc.)
    Version:  2.18.1
    Path:     /Users/regrau/.docker/cli-plugins/docker-compose

Server:
 Containers: 15
  Running: 10
  Paused: 0
  Stopped: 5
 Images: 15
 Server Version: 23.0.6
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 1
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: runc io.containerd.runc.v2
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 1fbd70374134b891f97ce19c70b6e50c7b9f4e0d
 runc version: 860f061b76bb4fc671f0f9e900f7d80ff93d4eb7
 init version: 
 Security Options:
  seccomp
   Profile: builtin
 Kernel Version: 6.1.29-0-virt
 Operating System: Alpine Linux v3.18
 OSType: linux
 Architecture: aarch64
 CPUs: 4
 Total Memory: 7.754GiB
[...]

Issue #10707, opened 2023-06-13
[BUG] Docker compose hangs forever on starting containers


mixmind:
### Description

I have a compose with number of services. If all services started as expected, then everything works fine. 
If one of services that other services depends on him failed, all process hangs forever and ctrl+c doesn't help

### Steps To Reproduce

Here is docker compose with problem..
using command:
docker compose up -d
```---
version: '3.7'

services:
  reverse-proxy:
    image: nginx
    container_name: reverse-proxy
    depends_on:
      flyway:
        condition: service_completed_successfully

  flyway:
    image: flyway/flyway
    container_name: flyway
    #command: "exit 1"`
```

### Compose Version

```Text
Docker Compose version v2.18.1
Docker Compose version v2.17.0
```

### Docker Environment

```Text
Client: Docker Engine - Community
 Version:    24.0.2
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.10.5
    Path:     /usr/libexec/docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.18.1
    Path:     /usr/libexec/docker/cli-plugins/docker-compose

Server:
 Containers: 13
  Running: 10
  Paused: 0
  Stopped: 3
 Images: 41
 Server Version: 24.0.2
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 3dce8eb055cbb6872793272b4f20ed16117344f8
 runc version: v1.1.7-0-g860f061
 init version: de40ad0
 Security Options:
  apparmor
  seccomp
   Profile: builtin
  cgroupns
 Kernel Version: 5.19.0-42-generic
 Operating System: Ubuntu 22.04.2 LTS
 OSType: linux
 Architecture: x86_64
 CPUs: 8
 Total Memory: 15.54GiB
 Name: home-pc
[...]

Issue #10648, opened 2023-06-02
[BUG] docker build process gets stuck on terminate when resolving deps


fho:
### Description

This is related to https://github.com/docker/compose/issues/10634 but I now have more information what is going wrong.

I have a large internal docker-compose project with a lot of dependencies. 
I run `docker build` and then press ctrl+c.
The terminal becomes broken (as described [here](https://github.com/docker/compose/issues/10634#issuecomment-1573388511)).
The `docker-build` plugin process does not terminate. It becomes a parent of PID 1. 
It does not react to any TERM signals. It seems that it will keep running forever and never terminate.

I compiled compose with debug simple and the stacktraces show that the process
always hangs while resolving dependencies, related to `nodeCh`.
I observed that it gets stuck in 2 different places, sometimes variant 1,
sometimes 2.

## Variant 1: stuck when receiving from `nodeCh`

No `graphTraversal.run` go routines are runnning.
1 goroutine executes `graphTraversal.visit`, it hangs at `for node := range nodeCh {`.
It will hang there forever because there are no go-routines that will send something to the channel.
My wild guess is:
Some run functions skipped sending something to `nodeCH` because of: https://github.com/docker/compose/blob/7c3fe359b776a2b26594dd97d2bc2bc257817aa1/pkg/compose/dependencies.go#L168-L171 matched.
It matches because cancelling the ctx, cancelled the start of children services.
No result for those services is send to `nodeCH`.
The `expect` counter in `visit()` then can not reach  0, `nodeCh` does not get closed and `visit()` gets stuck in the `for node := range nodeCh {` loop.
In a run were it happened `expect` had the value `99`.

Stacktraces of an occurence occurrence:

<details>

```
(dlv) goroutines
  Goroutine 1 - User: /usr/lib/go/src/runtime/sema.go:62 sync.runtime_Semacquire (0x472ae7) [semacquire 32698615744479]
  Goroutine 2 - User: /usr/lib/go/src/runtime/proc.go:382 runtime.gopark (0x44431d) [force gc (idle) 32698285413127]
  Goroutine 3 - User: /usr/lib/go/src/runtime/proc.go:382 runtime.gopark (0x44431d) [GC sweep wait]
  Goroutine 4 - User: /usr/lib/go/src/runtime/proc.go:382 runtime.gopark (0x44431d) [GC scavenge wait]
[...]

Issue #10661, opened 2023-06-06
[BUG] docker compose up not returning shell and erroneously detecting error


snspinn:
### Description

I want to get this issue down now before I forget some details, so apologies if it's a little light on detail. I will update as I reproduce this error for myself.
I'm running a fairly large project with docker compose and am running an issue where `docker compose up -d` and `docker compose up` are not returning the shell properly.

In the case of detached, one of the services is showing an error when I bring it up. When this happenss, the shell is not returned, even with a `CTRL+C`. The only way to exit is the kill the service directly w/ `kill -HUP <PID>`. When I get the shell back, I can attach to the logs and can confirm that the service that showed error has no errors (although this happens if service genuinely errors or now).

In the non-detached case, the services all come up without error and I attach to the output.log correctly. However, if I do a `CTRL-C` here, the services are stopped but I do not get my shell back automatically. I must then repeatedly do `CTRL-C` which results in
```
^Cno container to kill
^CERRO[0026] got 3 SIGTERM/SIGINTs, forcing shutdown  
```

I tried this in in zsh and bash shells, and each of these in xterm and gnome-terminal (xterm-256color).

Three things that might be causing it/ that are different from similar compose files:
- A lot of services in the compose (35)
- Some sections in the yaml are commented out
- Long file paths for volume mounts

### Steps To Reproduce

1. Environment: Ubuntu 20.04.2 LTS, xterm
     - Docker version 24.0.2, build cb74dfc
     - 
2. Run `docker compose up -d` (see psuedo-compose below)**

** I will add to this as I try reproduce with simpler compose files

```
version: '3'

services:

  myproject_pgsql:
    container_name: pgsql_myproject
    image: postgres:9.6-alpine
    restart: always
    environment:
      - POSTGRES_PASSWORD=root
    volumes:
      - ./subdir1/subdir2/configuration/postgresql/initdb:/docker-entrypoint-initdb.d
      - ./subdir1/subdir2/data/pgsql:/var/lib/postgresql/data
        #healthcheck:
        #test: ["CMD-SHELL", "pg_isready"]
        #interval: 10s
        #timeout: 5s
        #retries: 5
    networks:
      myproject_network:
        ipv4_address: 172.0.0.200


networks:
  myproject_network:
    external: true
```

### Compose Version

```Text
Docker Compose version v2.18.1
```

### Docker Environment

```Text
Client: Docker Engine - Community
 Version:    24.0.2
 Context:    default
 Debug Mode: false
 Plugins:
[...]

PR #10725, opened 2023-06-20
fix race condition, waiting for containers when one exit

ndeloof:
**What I did**
as a container exits with error, we must cancel but the goroutine waiting for expected container to be started used to get stuck waiting. Need to stop waiting when context is cancelled

**Related issue**
fixes https://github.com/docker/compose/issues/10707
fixes https://github.com/docker/compose/issues/10718

### DIFF
--- a/pkg/compose/dependencies.go
+++ b/pkg/compose/dependencies.go
@@ -140,24 +140,28 @@ func (t *graphTraversal) visit(ctx context.Context, g *Graph) error {
 	if t.maxConcurrency > 0 {
 		eg.SetLimit(t.maxConcurrency + 1)
 	}
-	nodeCh := make(chan *Vertex)
+	nodeCh := make(chan *Vertex, expect)
+	defer close(nodeCh)
+	// nodeCh need to allow n=expect writers while reader goroutine could have returner after ctx.Done
 	eg.Go(func() error {
-		for node := range nodeCh {
-			expect--
-			if expect == 0 {
-				close(nodeCh)
+		for {
+			select {
+			case <-ctx.Done():
 				return nil
+			case node := <-nodeCh:
+				expect--
+				if expect == 0 {
+					return nil
+				}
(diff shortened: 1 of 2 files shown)


glours (approved):
LGTM

pkg/compose/dependencies.go
```diff
@@ -140,24 +140,28 @@ func (t *graphTraversal) visit(ctx context.Context, g *Graph) error {
...
+	// nodeCh need to allow n=expect writers while reader goroutine could have returner after ctx.Done
 	eg.Go(func() error {
-		for node := range nodeCh {
-			expect--
-			if expect == 0 {
-				close(nodeCh)
+		for {
+			select {
+			case <-ctx.Done():
 				return nil
```
fho, line 149: if this code path is taken `nodeCh` is not closed

fho:
it also should fix: https://github.com/docker/compose/issues/10648

milas (approved):

ndeloof:
@fho would you have a chance to test this against  https://github.com/docker/compose/issues/10648 and https://github.com/docker/compose/issues/10661 to confirm those are all the same issue ?

fho:
> @fho would you have a chance to test this against #10648 and #10661 to confirm those are all the same issue ?

@ndeloof 
I already tested #10648 with this change, it fixes the issue.
With the described steps in #1661 I can not reproduce an issue.

jawys:
Hello there!

It took me 3 days of troubleshooting/purging/reinstalling docker-desktop to finally find this issue because compose also stopped throwing error messages which could have shortened my odyssey:

https://github.com/docker/compose/assets/22871352/1bf6fc4b-80b8-4bec-bc01-db659d75df3d

@ndeloof Could we please somehow trigger a release of Docker Desktop whenever a new Compose version is released so that always recent versions get shipped?? 
At our company compose is crucial for development, so the recent experience was real pain until I somehow found this related isse.

PR #10727, opened 2023-06-20
ci: upgrade to buildx v0.11

milas:
**What I did**
https://github.com/docker/buildx/releases/tag/v0.11.0

Several `replace` directives have been removed and dependencies aligned with buildx as needed.

baby beaver


glours (approved):
LGTM

pkg/compose/build.go
```diff
@@ -531,3 +533,18 @@ func useDockerDefaultOrServicePlatform(project *types.Project, service types.Ser
 	}
 	return plats, nil
 }
+
+func convertCacheOptions(in []*pb.CacheOptionsEntry) []bclient.CacheOptionsEntry {
```
ndeloof, line 537: AFAICT we could use `buildflags.ParseCacheEntry`

milas: It has to get converted from the buildx protobuf type to the BuildKit client type.

That said, I found the upstream method which I couldn't find earlier for some reason, so I switched to that: https://github.com/docker/buildx/blob/010e4c8d54dcaa5e661ba7bc6adbf26008c96d4c/controller/pb/cache.go

glours (approved):

ndeloof (approved):

Issue #10395, opened 2023-03-23
Add wait command to docker compose


chmely:
### Description

It would be nice to be able to start compose in detached mode and then be able to wait for a specific service(s) to finish, similar to what `docker wait` does.

Something like `docker compose [params] wait service [service] [--timeout xx]`
Return code - 0 if all awaited services completed successfully within the required time frame, error code otherwise

Why this could be useful?

We use compose in our CI pipeline where we want to start certain containers like LocalStack, Vault, etc. Then we have seeder containers that run as part of the compose and seed values or initialize infrastructure inside LocalStack.
After the seeding/initialization is done we run our integrations tests. But before we can run our integration tests we have to make sure the seeder containers have finished successfully.

So in our case it'd be great if we could do something like
```
docker compose [params] --profile infrastructure --profile localstack-seeder --profile vault-seeder up -d
docker compose [params] --profile infrastructure --profile localstack-seeder --profile vault-seeder wait localstack-seeder vault-seeder --timeout 60s

run integration tests here if $? is 0
```

kamilpi:
It could be useful for some scenarios.

You could always do the filter based on labels generated by ```docker compose``` service.

```
docker wait $(docker ps --filter label=com.docker.compose.service=foo -q)
```

chmely:
Yes, this is certainly one way, but not that intuitive, I didn't know that you could do it this way, so thanks for the tip, but I assume other people may also not know, so a wait command may be handy and more intuitive to use

ndeloof:
This is already available using `docker compose up --exit-code-from SERVICE`. Compose will wait for specified service to exit, then stop the stack and report exit code

chmely:
Well, but the documentation for `--exit-code-from` [says](https://docs.docker.com/engine/reference/commandline/compose_up/#options) that using it also implies `--abort-on-container-exit`, so in my scenario above it would not work well if waiting for multiple services to finish, as it would abort on the first of the multiple containers/services running. Also, there is no timeout option.

chmely:
Hi guys, any update on this?

chmely:
Hi guys, thanks for looking into this.
However, I am not sure the implementation is doing what I proposed.

It seems that `docker compose wait` waits for the first of any of the services to complete, what I proposed was something similar to what `docker container wait` does, to wait for one or more containers, in compose case one or more services.

PR #10742, opened 2023-06-26
Add `docker compose wait`

ulyssessouza:
**What I did**
Add `wait` command

**Related issue**
Resolves https://github.com/docker/compose/issues/10395

sloth

### DIFF
--- a/cmd/compose/wait.go
+++ b/cmd/compose/wait.go
@@ -1,0 +1,72 @@
+/*
+   Copyright 2023 Docker Compose CLI authors
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package compose
+
+import (
+	"context"
+	"os"
+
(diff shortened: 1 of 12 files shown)


nicksieger (approved):
LGTM!

nicksieger:
One more thing, what do you think about adding a e2e or cuke test?

pkg/compose/wait.go
```diff
@@ -0,0 +1,64 @@
...
+	containers, err := s.getContainers(ctx, project.Name, oneOffInclude, false, options.Services...)
+	if err != nil {
+		return 0, err
+	}
+
+	var returnErrorChan = make(chan error)
+	var statusCode int64
+	for _, c := range containers {
+		c := c
+		go func() {
```
ndeloof, line 37: could use an errorGroup (like we do in other places in compose) to capture first error and stop other API calls by cancelable context

ndeloof:
this new command deserves an e2e tests :)

ulyssessouza:
@ndeloof @nicksieger  The idea was to validate the concept first (that's the reason the draft). I'll write the tests now.

cmd/compose/wait.go
```diff
@@ -0,0 +1,64 @@
...
+			opts.services = services
+			statusCode, err = runWait(ctx, backend, project, &opts)
+			return err
+		}),
+		PostRun: func(cmd *cobra.Command, args []string) {
+			os.Exit(int(statusCode))
+		},
+	}
+
+	cmd.Flags().BoolVar(&opts.downProject, "down-project", false, "Drops project when the first container stops")
```
ulyssessouza, line 54: For now I just added this `--down-project`. But I can imagine having a `stop-project` or `stop-services`, only stopping the services we enumerated in the `wait`

ndeloof (approved):

Issue #10664, opened 2023-06-07
Feature: add the --builder flag support for build parameter for multi-arch builds


HarvyNBB:
### Description

* Brief:
Reference to the [documentation](https://docs.docker.com/compose/compose-file/build/)
Since it's possible to support multi-arch builds in docker-compose.yml, 

Additionally reference to the [documentation](https://docs.docker.com/build/architecture/)
```
In newer versions of Docker Desktop and Docker Engine, you’re using Buildx by default when you invoke the docker build command.
```

as a result, if the buildkit is enabled on the docker engine, then I assume docker compose build would use buildx

So during building simultaneous services with multi-arch there is a [buildkit issue](https://github.com/moby/buildkit/issues/3929) | [buildx issue](https://github.com/docker/buildx/issues/1868) faced while trying to build multi arch builds for multiple services

Note: I'm not sure whether this is buildx issue or buildkit issue, so I created an issue for both

so in order to get over this issue it would be so beneficial for multi-arch builds if the builder could be either passed via a flag or defined in the docker-compose.yml
for e.g docker-compose.yml

Multiple services:
```
version: '3.7'

services:
 
  service_1:
    container_name: service_1-${VERSION}
    image: service_1:${VERSION}    
    build:
      context: .
      dockerfile: Dockerfile
      builder: test1
      platforms: 
        - ${PLATFORM_1}
        - ${PLATFORM_2}
  
  service_2:
    container_name: service_2-${VERSION}
    image: service_2:${VERSION}
    build:
      context: .
      dockerfile: Dockerfile
      builder: test2
      platforms: 
        - ${PLATFORM_1}
        - ${PLATFORM_2}

  service_3:
    container_name: service_3-${VERSION}
    image: service_3:${VERSION}
    build:
      context: .
      dockerfile: Dockerfile
      builder: test3
      platforms: 
        - ${PLATFORM_1}
        - ${PLATFORM_2}
```

commands to run
```
docker buildx create --name test1
docker buildx create --name test2
docker buildx create --name test3
docker compose build service_1 service_2 service_3 
```
 Or for a single service
```
version: '3.7'

services:
 
  service_1:
    container_name: service_1-${VERSION}
    image: service_1:${VERSION}    
    build:
      context: .
      dockerfile: Dockerfile
      platforms: 
        - ${PLATFORM_1}
        - ${PLATFORM_2}
```

```
docker buildx create --name test1
docker compose build --builder test1 service_1 
```

[...]

PR #10745, opened 2023-06-27
add support of --builder and BUILDX_BUILDER

glours:
**What I did**
Add `--builder` flag to `build` command and support of the `BUILDX_BUILDER` env variable to support choose dedicated builder.

Should address #10664 until a proposal is discussed in the Compose specification

**Related issue**
N/A

image

### DIFF
--- a/cmd/compose/build.go
+++ b/cmd/compose/build.go
@@ -43,6 +43,7 @@ type buildOptions struct {
 	noCache bool
 	memory  cliopts.MemBytes
 	ssh     string
+	builder string
 }
 
 func (opts buildOptions) toAPIBuildOptions(services []string) (api.BuildOptions, error) {
@@ -54,6 +55,10 @@ func (opts buildOptions) toAPIBuildOptions(services []string) (api.BuildOptions,
 			return api.BuildOptions{}, err
 		}
 	}
+	builderName := opts.builder
+	if builderName == "" {
+		builderName = os.Getenv("BUILDX_BUILDER")
+	}
 
 	return api.BuildOptions{
 		Pull:     opts.pull,
@@ -64,6 +69,7 @@ func (opts buildOptions) toAPIBuildOptions(services []string) (api.BuildOptions,
 		Quiet:    opts.quiet,
 		Services: services,
 		SSHs:     SSHKeys,
+		Builder:  builderName,
 	}, nil
 }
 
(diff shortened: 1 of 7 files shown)


silvin-lubecki (approved):
LGTM 😻

cmd/compose/build.go
```diff
@@ -101,6 +103,7 @@ func buildCommand(p *ProjectOptions, progress *string, backend api.Service) *cob
 	cmd.Flags().BoolVar(&opts.pull, "pull", false, "Always attempt to pull a newer version of the image.")
 	cmd.Flags().StringArrayVar(&opts.args, "build-arg", []string{}, "Set build-time variables for services.")
 	cmd.Flags().StringVar(&opts.ssh, "ssh", "", "Set SSH authentications used when building service images. (use 'default' for using your default SSH Agent)")
+	cmd.Flags().StringVar(&opts.builder, "builder", "", "Set builder to use.")
 	cmd.Flags().Bool("parallel", true, "Build images in parallel. DEPRECATED")
 	cmd.Flags().MarkHidden("parallel") //nolint:errcheck
```
thaJeztah, line 108: Very unrelated, but if this flag is deprecated, should this use `MarkDeprecated()`? (are we still using the value anywhere?)

glours: I'll do a dedicated PR for this one 👍

ndeloof (approved):
LGTM but shall we also introduce a `builder` attribute in the compose specification?

ndeloof:
It would indeed be nice there's some way to describe a builder so that we can inspect the available ones and select a good match, but afaik we only can rely on a name. Also, I can imagine it will become quickly boring for a user to pass `--builder foo` every time he wants to build image, also would need this to be available on `compose up`... while this could just be set once in the compose.yaml file.
Last but not least, an image might require a specific builder to be used, but not the others

pkg/compose/build_buildkit.go
```diff
@@ -34,8 +35,11 @@ import (
 	"github.com/moby/buildkit/client"
 )
 
-func (s *composeService) doBuildBuildkit(ctx context.Context, service string, opts build.Options, p *buildx.Printer) (string, error) {
-	b, err := builder.New(s.dockerCli)
+func (s *composeService) doBuildBuildkit(ctx context.Context, service string, opts build.Options, p *buildx.Printer, builderName string) (string, error) {
+	if builderName == "" {
+		builderName = os.Getenv("BUILDX_BUILDER")
+	}
```
milas, line 41: Not critical, but I think it might be a bit cleaner/separation of concerns to do the fallback env lookup in the `cmd/` package parts and use the value as-is here

milas (approved):
For consistency, it'd be nice to have the flag available on `up` as well, but logic LGTM!

pkg/compose/build_buildkit.go
glours: Good catch, thanks Milas

glours:
> For consistency, it'd be nice to have the flag available on `up` as well, but logic LGTM!

I'm not sure we should add this flag to the `up`, the consistency should be to keep all the build flags to the build command only, IHMO, WDYT?

ndeloof:
Agree advanced build options should be limited to `build` command. This is also why I suggest we get such attributes introduce in compose.yaml

Issue #10748, opened 2023-06-28
[BUG] Random errors with docker compose up -d --force-recreate in version 2.19.0


jurgenhaas:
### Description

In our build and deploy pipelines we're using docker 24.0.2 and docker compose 2.19.0 where we see random errors since we've updated docker compose from 2.18.1 - before no errors at all, with hundreds of pipelines running every day.

```
$ docker compose up -d --force-recreate --quiet-pull --remove-orphans
 Container drupal-9_main-alerta-1  Recreate
 Container drupal-9_main-redis-1  Recreate
 Container drupal-9_main-mariadb-1  Recreate
 Container drupal-9_main-borgmatic-1  Recreate
 Container drupal-9_main-cron-1  Recreate
 Container drupal-9_main-php-1  Recreate
 Container drupal-9_main-cron-1  Recreated
 Container drupal-9_main-redis-1  Recreated
 Container drupal-9_main-php-1  Recreated
 Container drupal-9_main-apache-1  Recreate
Error response from daemon: container 70979001dd6e8a60ec6ecc7c83eaee8c87f551d1ea6efc92a771d37fd9283fed is not connected to the network traefik-public
```

When going to the host where this happened, we're seeing this status:

```
$ docker compose ps
NAME                                     COMMAND                  SERVICE             STATUS              PORTS
30f91d2adcbd_drupal-9_main-borgmatic-1   "/entry.sh"              borgmatic           created             
591391f7b262_drupal-9_main-mariadb-1     "/docker-entrypoint.…"   mariadb             created             
843a6249ea7e_drupal-9_main-apache-1      "/docker-entrypoint.…"   apache              created             
c63c713edb25_drupal-9_main-alerta-1      "/bin/sh -c 'python …"   alerta              created             
drupal-9_main-alerta-1                   "/bin/sh -c 'python …"   alerta              exited (137)        
drupal-9_main-apache-1                   "/docker-entrypoint.…"   apache              running             80/tcp
drupal-9_main-borgmatic-1                "/entry.sh"              borgmatic           exited (137)        
drupal-9_main-cron-1                     "/usr/bin/ofelia dae…"   cron                created             
drupal-9_main-mariadb-1                  "/docker-entrypoint.…"   mariadb             exited (0)          
drupal-9_main-php-1                      "/docker-entrypoint.…"   php                 created             
drupal-9_main-redis-1                    "/docker-entrypoint.…"   redis               created 
```

[...]

Issue #10668, opened 2023-06-07
[BUG] Intermittent "container <id> is not connected to the network <network>" errors on up


milas:
### Description

Since I've run the E2E suite a bunch today, I've noticed the ddev test fails intermittently when bringing up the project:
```
Failed to start 004: ComposeCmd failed to run 'COMPOSE_PROJECT_NAME=ddev-004 docker-compose -f /tmp/TestComposeRunDdev2386477416/004/.ddev/.ddev-docker-compose-full.yaml up --build -d', action='[up --build -d]', err='exit status 1', stdout='#1 [web internal] load build definition from Dockerfile
```

```
Network ddev-004_default  Creating
Network ddev-004_default  Created
Container ddev-004-web  Creating
Container ddev-004-dba  Creating
Container ddev-004-db  Creating
Error response from daemon: container 434c66ed5ce1c8b0a230b1ba25ac0b5cbedfc62710f3fbe01c0000d8fe2e2f80 is not connected to the network ddev_default'
```

That's coming from here: https://github.com/docker/compose/blob/5155d22e09313ddbf3de102a9870328e13d43af2/pkg/compose/convergence.go#L565

When I modified the test locally to leave things running, I could see that the container was in fact connected to the network, so I think this is another Engine race condition with networks given that we create the container just before this

### Steps To Reproduce

Run `TestComposeRunDdev` E2E test a bunch

### Compose Version

```Text
HEAD
```

### Docker Environment

```Text
Occurs on both Engine v20.10.x in CI and v24.x locally on my macOS Docker Desktop install
```

milas:
Given that we're about to reconnect the network (with the full set of options? which I guess we can't do on create), would it make more sense to do the initial container create with network: none and then connect so we don't have to do the pre-emptive disconnect?

ndeloof:
This was implemented to replicate the `docker run` API call payload, but maybe we indeed don't need to.

ToshY:
I got the same error when I upgraded compose from `2.18.1` to `2.19.0`. Reverted back in the meantime.

wintheriscomming:
We got the same error after upgrading to 2.19.0, after reverting to previous version things started working again.

rfay:
This is preventing [DDEV](https://github.com/ddev/ddev) from using v2.19.0, not sure what the fix is, fails on every platform
* https://github.com/ddev/ddev/pull/5006
[...]

PR #10756, opened 2023-06-29
up: fix race condition on network connect

milas:
**What I did**
Fix race conditions with network setup during `up`.

Engine API only allows at most one network to be connected as part of the `ContainerCreate` API request. Compose will pick the highest priority network.

Afterwards, the remaining networks (if any) are connected before the container is actually started.

The big change here is that, previously, the highest-priority network was connected in the create, and then disconnected and immediately reconnected along with all the others. This was racy because evidently connecting the container to the network as part of the create isn't synchronous, so sometimes when Compose tried to disconnect it, the API would return an error like:
```
container <id> is not connected to the network <network>
```

To avoid needing to disconnect and immediately reconnect, the network config logic has been refactored to ensure that it sets up the network config correctly the first time.

**Related issue**
* #10668 
* #10748

https://docker.atlassian.net/browse/ENV-241

a spotted piglet or mini-pig

### DIFF
--- a/pkg/compose/convergence.go
+++ b/pkg/compose/convergence.go
@@ -29,7 +29,6 @@ import (
 	"github.com/containerd/containerd/platforms"
 	moby "github.com/docker/docker/api/types"
 	containerType "github.com/docker/docker/api/types/container"
-	"github.com/docker/docker/api/types/network"
 	specs "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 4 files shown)


nicksieger (approved):
Nice fix and cleanup!

ndeloof (approved):

ndeloof:
Nice! 🎉

glours (approved):

ToshY:
@milas There are multiple reports in #10668 that this fix (released in `2.19.1`) solves the `not connected to network` problem, but subsequently causes other issues in return.

Issue #10766, opened 2023-07-03
[BUG] `--dry-run` flag causes Compose commands to to fail on Synology with official Docker package


Renerick:
### Description

When trying to run `docker compose --dry-run up` command I receive this error

```shell
$ sudo docker compose --dry-run up
invalid name , name needs to start with a letter and may not contain symbols, except ._-
```

Same file works normally without --dry-run

```shell
$ sudo docker compose up
[+] Running 0/1
 ⠹ db Pulling                  
 ```
 
<details>
<summary>The file</summary>
<pre>
services:
  db:
    image: postgres
    volumes:
      - ./tmp/db:/var/lib/postgresql/data
    environment:
      POSTGRES_PASSWORD: password
  web:
    build: .
    command: bash -c "rm -f tmp/pids/server.pid && bundle exec rails s -p 3000 -b '0.0.0.0'"
    volumes:
      - .:/myapp
    ports:
      - "3000:3000"
    depends_on:
      - db
 </pre>
</details>

The important nuance is that I run docker on Synology NAS from the official Container Manager package, so it's either the fact that Docker version is older than it should be, or some fundamental incompatibility of `--dry-run` with Synology DSM

The plugin is installed from release binary from this repo

### Steps To Reproduce

1. Acquire Synology NAS and install "Docker"/"Container Manager" package, as well as the Compose plugin
2. Create a valid docker-compose.yml
3. Try to run `docker compose up`, confirm that the file is valid
4. Try to run `docker compose --dry-run up`, receive an error message

### Compose Version

```Text
$ sudo docker compose version
Docker Compose version v2.19.1
```

### Docker Environment

```Text
$ sudo docker info
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  compose: Docker Compose (Docker Inc., v2.19.1)

Server:
 Containers: 5
  Running: 4
  Paused: 0
  Stopped: 1
 Images: 13
 Server Version: 20.10.23
 Storage Driver: btrfs
  Build Version: Btrfs v4.0
  Library Version: 101
 Logging Driver: db
 Cgroup Driver: cgroupfs
 Cgroup Version: 1
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs db fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 io.containerd.runtime.v1.linux runc
 Default Runtime: runc
[...]

PR #10770, opened 2023-07-04
don't use unitialized cli to setup DryRunClient

ndeloof:
**What I did**
use the actual dockerCli to setup DryRunClient, otherwise we get a chicked/egg problem with buildx trying to guess the active context while we are still intializing

**Related issue**
fixes https://github.com/docker/compose/issues/10766

image


glours (approved):
LGTM

milas (approved):

PR #10772, opened 2023-07-05
bump buildkit to 0.11.4

ndeloof:
**Related issue**
closes https://github.com/docker/compose/security/dependabot/21

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -24,7 +24,7 @@ require (
 	github.com/jonboulle/clockwork v0.4.0
 	github.com/mattn/go-shellwords v1.0.12
 	github.com/mitchellh/mapstructure v1.5.0
-	github.com/moby/buildkit v0.11.0-rc3.0.20230609092854-67a08623b95a
+	github.com/moby/buildkit v0.11.4
 	github.com/moby/patternmatcher v0.5.0
 	github.com/moby/term v0.5.0
 	github.com/morikuni/aec v1.0.0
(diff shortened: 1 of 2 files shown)


milas (approved):

glours:
I think we should use this [PR](https://github.com/docker/compose/pull/10794) instead to be sure the version of buildkit is align with the one defined in buildx

go.mod
```diff
@@ -24,7 +24,7 @@ require (
 	github.com/jonboulle/clockwork v0.4.0
 	github.com/mattn/go-shellwords v1.0.12
 	github.com/mitchellh/mapstructure v1.5.0
-	github.com/moby/buildkit v0.11.0-rc3.0.20230609092854-67a08623b95a
+	github.com/moby/buildkit v0.11.4
```
thaJeztah, line 27: ⚠️ v0.11.4 is **older** than `v0.11.0-rc3.0.20230609092854-67a08623b95a`. which is actually "v0.12-dev"


The `v0.11.0-rc3.0.20230609092854-67a08623b95a` pseudo version is based on the last tag that was cut from master (`v0.11.0-rc3`), which always makes it confusing. (sigh)

But if you look at the commit; https://github.com/moby/buildkit/commit/67a08623b95a, then you'll see it's a commit from master;
Screenshot 2023-07-17 at 01 51 03

thaJeztah (changes requested):
see comment below

go.mod
ndeloof: > v0.11.4 is older than v0.11.0-rc3

seriously? 😰

thaJeztah: Yeah, go modules tried to be smart and create versions for non-tagged commits, and it does so by;

- taking the commit
- find the last tag in the commit's history (in this case: master branch)
- if the tag cannot be parsed as SemVer, continue with the next one
- once a "SemVer" tag is found, take the tag and increment "Z" (`vX.Y.Z+1`)
- ...and that's what they use as "pseudo version" prefix


Which, erm, doesn't really work if a project uses release-branches, and the last tag was the pre-release before the release-branch was created.

Issue #10777, opened 2023-07-06
`Host name lookup failure` when using multiple networks


milas:
_Originally posted by @ToshY in https://github.com/docker/compose/issues/10668#issuecomment-1615217336_:

@milas  While this seems to fix the `not connected to network` issue, it gave me a `forward host lookup failed: Host name lookup failure` in return when running [`waisbrot/wait` ](https://hub.docker.com/r/waisbrot/wait) with `docker compose run`.

**Example**

```yaml
version: '3.9'
  
services:
  mysql:
    image: ${MYSQL_IMAGE_VERSION}
    ports:
      - "3308:3306"
    environment:
      MYSQL_USER: ${DATABASE_USER}
      MYSQL_DATABASE: ${DATABASE_NAME}
      MYSQL_ROOT_PASSWORD: ${DATABASE_ROOT_PASSWORD}
      MYSQL_PASSWORD: ${DATABASE_PASSWORD}
    volumes:
      - mysql_data:/var/lib/mysql
    networks:
      - webapp

  wait:
    image: waisbrot/wait
    environment:
      TARGETS: mysql:3306
      TIMEOUT: 300
    networks:
      - webapp

volumes:
  mysql_data:
    driver: local

networks:
  webapp:
    driver: bridge
```

```bash
$ docker compose up -d
[+] Running 4/4
 ✔ Container webapp-wait-1    Started                                                                                                                                                         3.0s
 ✔ Container webapp-mysql-1   Started                                                                                                                                                         2.9s

$ docker compose run --rm wait
Waiting for mysql:3306  .mysql: forward host lookup failed: Host name lookup failure
.mysql: forward host lookup failed: Host name lookup failure
.mysql: forward host lookup failed: Host name lookup failure

$ docker compose ps
NAME                            IMAGE                         COMMAND                  SERVICE             CREATED             STATUS              PORTS
webapp-mysql-1                  mysql:8.0                     "docker-entrypoint.s…"   mysql               8 minutes ago       Up 5 minutes        33060/tcp, 0.0.0.0:3308->3306/tcp, :::3308->3306/tcp
```

jshufro:
Heyo. We just ran into this. The container is only added to one network, so dns doesn't resolve. Appears to be a regression

jasonheffnerpsu:
Had to revert over the weekend to 2.18.1. One of the four containers replicated wouldn't join the postgres network, though all would connect to the main Traefik network fine.

oerp-odoo:
[...]

PR #10778, opened 2023-07-06
network: fix random missing network when service has more than one

milas:
**What I did**
As part of the fix for #10668, the logic was adjusted so that the default (highest-priority) network is used in the `ContainerCreate`, and then the remaining networks are connected via calls to `NetworkConnect` before starting the container.

Unfortunately, `ServiceConfig::NetworksByPriority` is neither deterministic nor stable when networks have the same priority.

It's non-deterministic because the order of networks from parsing YAML is random, since they are loaded into a Go map (which have random iteration order). Additionally, it's not using a `SortStable` in `compose-go`, so even if the load order was predictable, it still might produce different results.

While I look at improving `compose-go` here to prevent this from tripping us up in the future, this fix looks at _all_ networks for a service and ignores the "default" one now. Before, it would always skip the first one in the slice since that _should_ have been the "default".

**Related issue**
* #10777 
* #10668 

a baby panda on a tree branch

### DIFF
--- a/pkg/e2e/networks_test.go
+++ b/pkg/e2e/networks_test.go
@@ -29,47 +29,34 @@ import (
 
 func TestNetworks(t *testing.T) {
 	// fixture is shared with TestNetworkModes and is not safe to run concurrently
-	c := NewCLI(t)
-
 	const projectName = "network-e2e"
+	c := NewCLI(t, WithEnv(
+		"COMPOSE_PROJECT_NAME="+projectName,
+		"COMPOSE_FILE=./fixtures/network-test/compose.yaml",
+	))
 
-	t.Run("ensure we do not reuse previous networks", func(t *testing.T) {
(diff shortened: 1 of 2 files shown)


pkg/e2e/networks_test.go
```diff
@@ -29,47 +29,34 @@ import (
 
 func TestNetworks(t *testing.T) {
 	// fixture is shared with TestNetworkModes and is not safe to run concurrently
-	c := NewCLI(t)
-
 	const projectName = "network-e2e"
+	c := NewCLI(t, WithEnv(
```
milas, line 33: fyi this test is functionally the same, I just removed the "subtests" since it's really one test, used `WithEnv` to shorten the commands, and added a `-t0` on the `down` so that it's faster

glours (approved):

ndeloof (approved):

bf:
This issue cost me a lot of time. Thanks for fixing.

Issue #10779, opened 2023-07-07
[BUG] Compose output issues on 2.19.0/1


DawidJanczak:
### Description

Hello,

After upgrading from 2.18.1 to 2.19.1 the output in `docker compose up` command gets clobbered weirdly. Some instances:
1. When building I see messages like
```
container-webpack-1  | Compiling ...
Success!     pack-1  |
container-webpack-1  |
```
That "Success!" shouldn't be on the left hand side, should be after the pipe like everything else.

2. When I get warnings in one of the languages (Elm) they're being printed really oddly:
```
container-webpack-1  | The
container-webpack-1  |  2nd a
container-webpack-1  | r
container-webpack-1  | gum
container-webpack-1  | e
container-webpack-1  | nt
container-webpack-1  |
container-webpack-1  | t
container-webpack-1  | o
container-webpack-1  | `
container-webpack-1  | m
container-webpack-1  | a
container-webpack-1  | ybeRende
container-webpack-1  | r` i
container-webpack-1  | s
container-webpack-1  | not w
container-webpack-1  | hat
container-webpack-1  |  I
container-webpack-1  | e
container-webpack-1  | x
container-webpack-1  | p
container-webpack-1  | e
container-webpack-1  | c
container-webpack-1  | t
container-webpack-1  | :
container-webpack-1  |
container-webpack-1  |
container-webpack-1  |
container-webpack-1  |
container-webpack-1  | 659|                 , maybeRender (strike
container-webpack-1  | throughPrice per
container-webpack-1  | Perso
container-webpack-1  | nPri
container-webpack-1  | ce)
container-webpack-1  | 660|
container-webpack-1  |
container-webpack-1  |                    (Maybe.map .discount discountOffer)
container-webpack-1  |
container-webpack-1  |
```

On 2.18.1 they get printed normally.

3. When attaching to a running container with `docker attach` whatever I type in the session on the attached container gets mirrored in docker compose logs _one letter at a time (one letter per line)_, e.g.
```
container            | [1] pry(#<VenueShow>)>
container            | w
container            | h
container            | at
container            |
container            | i
container            | s
container            | t
container            | h
container            | h
container            | i
container            | s
container            |
container            | a
container            | b
container            | o
container            | t
container            |
container            | a
container            | d
container            | f
container            | s
container            | d
container            | f
```

On 2.18.1 nothing gets printed in docker logs output.

I checked the changelog and couldn't find anything obvious, hence this ticket.

### Compose Version

```Text
[...]

PR #10782, opened 2023-07-07
Flush splitWriter buffer on dettach to capture last line of logs

ndeloof:
**What I did**
This is a fix for https://github.com/docker/compose/issues/10658 which doesn't rely on assumption we receive container logs as a plain line. Some stacks don't buffer stdout and we actually receive logs char by char

`attach` event get a new callback attribute so that, as we detach on exit, we can flush buffers to collect the remaining logs

**Related issue**
fixes https://github.com/docker/compose/issues/10779
https://docker.atlassian.net/browse/ENV-246

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -524,6 +524,9 @@ type LogConsumer interface {
 // ContainerEventListener is a callback to process ContainerEvent from services
 type ContainerEventListener func(event ContainerEvent)
 
+// ContainerDetachFn detach from container stream and return the remaining logs from buffers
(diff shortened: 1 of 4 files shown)


milas (approved):
Fix makes sense.

TBH I'm not a huge fan of the `Detach` field as it's pretty specialized/feels like a workaround, but given the way event -> print logic works, I don't really have a better suggestion.

ndeloof:
Closing this. This implementation fails to capture first line of "hello" service in TestLocalComposeLogsFollow and architecture is very unpleasant

PR #10784, opened 2023-07-07
fix(secrets): file permission value does not comply with spec

### LINKED ISSUE #10783: [BUG] Secrets permission mode value does not comply with current Compose Specification
### Description

According to the [Long Syntax Description for Secrets in the Compose Specification][1] Docs:

> mode: The [permissions](https://wintelguy.com/permissions-calc.pl) for the file to be mounted in /run/secrets/ in the service's task containers, in octal notation. Default value is world-readable permissions (mode 0444). The writable bit must be ignored if set. The executable bit may be set.

However, based on the implementation for [`pkg/compose/secrets.go`][2] the value is set to `0o400`.

[1]: https://github.com/compose-spec/compose-spec/blob/master/05-services.md#long-syntax-4
[2]: https://github.com/docker/compose/blob/v2/pkg/compose/secrets.go#L61

### Steps To Reproduce

For the following Compose File `docker-compose.alpine.yml`:

```yaml
services:
  demo:
    image: alpine
    secrets:
      - foo

secrets:
    foo:
      environment: FOO
```

mounting a secret via an environment variable via CLI:

```bash
$ FOO=hello docker compose -f docker-compose.alpine.yml run demo stat -c "%a" /run/secrets/foo
```

provides the following output:

```bash
[+] Building 0.0s (0/0)                                                                                                                                                                                            
[+] Building 0.0s (0/0)                                                                                                                                                                                            
400
```

### Compose Version

```Text
v2.18.1 / 2.19.1
```

### Docker Environment

```Text
Client:
 Version:    24.0.0
 Context:    default
 Debug Mode: false
 Plugins:
    Version:  0.10.4
    Path:     /usr/lib/docker/cli-plugins/docker-buildx
    Version:  2.18.1
```

has also been tested on __Docker Engine v24.0.3__

### Anything else?

_No response_

shantanoo-desai:
Compose Spec mentions that default values for secrets is `0444` aka. world-readable permissions. However, the value was previously set to `0400`.

**What I did**
Fix the file permission mode value from previously set `0400` to `0444` in order to comply with the Compose Specifications where Secrets should be defaulted to world-readable permissions unless not explicitly mentioned via Long Syntax.

**Related issue**

closes #10783

### DIFF
--- a/pkg/compose/secrets.go
+++ b/pkg/compose/secrets.go
@@ -58,7 +58,7 @@ func createTar(env string, config types.ServiceSecretConfig) (bytes.Buffer, erro
 	value := []byte(env)
 	b := bytes.Buffer{}
 	tarWriter := tar.NewWriter(&b)
-	mode := uint32(0o400)
+	mode := uint32(0o444)
 	if config.Mode != nil {
 		mode = *config.Mode
 	}


milas (approved):
LGTM, thanks for the PR!

glours (approved):

ndeloof (approved):

Issue #10751, opened 2023-06-28
[BUG] service is required by service but is disabled


nmguse-bighealth:
### Description

v2.19.0 introduced a change that broke a docker compose run command that was working previously (v2.18.1)

The change, I believe, would have come from this PR: https://github.com/docker/compose/pull/10602

`docker compose -p unique_name run --rm --no-deps test black --line-length 150 --check app/ tests/`

Error: `service postgres is required by test but is disabled. Can be enabled by profiles []`

I believe this is roughly all that would be required to reproduce, but I haven't directly tested this as docker mac hasn't updated docker compose yet:
```
version: "3"
services:
  test:
    profiles:
      - test
    build:
      context: .
    depends_on:
      postgres

  postgres:
    image: postgres:14-alpine
```

### Compose Version

```Text
v2.19.0
```

allanlewis:
I'm trying to work around this by adding these steps to my failing job:

```yaml
    - name: Log Docker Compose version
      run: docker compose version
    - name: Ensure Docker CLI plugins directory exists
      run: mkdir -pv "${HOME}"/.docker/cli-plugins
    - name: Download working version of Docker Compose
      run: wget -nv -O "${HOME}"/.docker/cli-plugins/docker-compose "${DOCKER_COMPOSE_URL}"
      env:
        DOCKER_COMPOSE_URL: https://github.com/docker/compose/releases/download/v2.18.1/docker-compose-linux-x86_64
    - name: Make Docker Compose binary executable
      run: chmod -v +x "${HOME}"/.docker/cli-plugins/docker-compose
    - name: Assert that the version of Docker Compose is as expected
      run: set -o pipefail && docker compose version | tee /dev/stderr | grep -q 'v2\.18\.1$'
```

jvalecillos:
I can confirm that the version `v2.18.1` works.

milas:
I'm not able to reproduce this with the example given. (Tweaked slightly to fix syntax errors and not use a local build, I'm launching an container that just does nothing.)

```yaml
services:
  test:
    profiles:
      - test
    image: busybox
    init: true
    command: ["sleep", "infinity"]
    depends_on:
      - postgres

  postgres:
    image: postgres:14-alpine

```

If I run `test`:
```
$ docker compose run --rm test
[+] Creating 2/2
 ✔ Network dep_default       Created                                                                                                                  0.0s
 ✔ Container dep-postgres-1  Created                                                                                                                  0.0s
[+] Running 1/1
[...]

PR #10789, opened 2023-07-10
Apply no-deps before we select and mutate target service

ndeloof:
**What I did**
`project.ForServices` mutates the project and services to remove depends_on relations, so we have to run it first before we select the run target to mutate it (again) otherwise we override `project.ForServices` tweaks

**Related issue**
fixes https://github.com/docker/compose/issues/10751


glours (approved):
LGTM

milas (approved):
A test case might be nice here to help improve coverage around `depends_on` 😉

PR #10790, opened 2023-07-10
test: fix process leak in wait e2e test

milas:
**What I did**
* Run `down` before and after test to not leave around containers
* Kill the `wait` process that's waiting on `infinity`
  * NOTE: If the test is actually working, this should exit once the `down` happens, but this ensures that we kill everything we start

I'd like to generalize more of this into the framework, but this is a quick fix to prevent filling up CI machines with tons of processes over time.

two rabbits coming out of a burrow

### DIFF
--- a/pkg/e2e/wait_test.go
+++ b/pkg/e2e/wait_test.go
@@ -21,12 +21,20 @@ import (
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


glours (approved):
LGTM

PR #10791, opened 2023-07-10
watch: move sync logic into separate package

milas:
**What I did**
Just moving some code around in preparation for an alternative sync implementation that can do bulk transfers by using `tar`.

**Related issue**
https://docker.atlassian.net/browse/ENV-191

beaver


glours (approved):
LGTM

PR #10792, opened 2023-07-10
add support of depends_on.required attribute

glours:
**What I did**
Add support of `depends_on.required` attribute

**Related issue**
N/A

image


pkg/compose/compose.go
```diff
@@ -211,8 +212,11 @@ func (s *composeService) projectFromName(containers Containers, projectName stri
 					if len(dcArr) > 2 {
 						restart, _ = strconv.ParseBool(dcArr[2])
 					}
+					if len(dcArr) > 3 {
```
ndeloof, line 215: this block is for backward compatibility only, I don't think we should extend syntax to support a 3rd argument

pkg/compose/convergence.go
```diff
@@ -355,6 +355,9 @@ func shouldWaitForDependency(serviceName string, dependencyConfig types.ServiceD
 		// already managed by InDependencyOrder
 		return false, nil
 	}
+	if !dependencyConfig.Required {
```
ndeloof, line 358: it SHOULD wait IF dependency has been detected. Should only ignore dependency was not found, but here you just bypass waiting for dependency whenever it's been detected or not

pkg/compose/create.go
```diff
@@ -413,7 +414,7 @@ func (s *composeService) prepareLabels(labels types.Labels, service types.Servic
 
 	var dependencies []string
 	for s, d := range service.DependsOn {
-		dependencies = append(dependencies, fmt.Sprintf("%s:%s:%t", s, d.Condition, d.Restart))
+		dependencies = append(dependencies, fmt.Sprintf("%s:%s:%t:%t", s, d.Condition, d.Restart, d.Required))
```
ndeloof, line 417: not a fan extending syntax here. I'd prefer we drop dependency if we can't find it.

pkg/compose/compose.go
milas: I'm definitely not a big fan of stuffing comma&colon-delimited data into the label, but without it here, what does that mean? When in project mode, we'd consider all dependencies required? 🤔

pkg/compose/restart.go
```diff
@@ -48,10 +48,10 @@ func (s *composeService) restart(ctx context.Context, projectName string, option
 		}
 	}
 
-	// ignore depends_on relations which are not impacted by restarting service
+	// ignore depends_on relations which are not impacted by restarting service or not required
 	for i, service := range project.Services {
 		for name, r := range service.DependsOn {
-			if !r.Restart {
+			if !r.Restart || !r.Required {
```
milas, line 54: Similar to @ndeloof's [Q around wait](https://github.com/docker/compose/pull/10792/files#r1266706274), should this not filter them out, but ignore any failures attempting to restart IF it was not required?

pkg/e2e/up_test.go
```diff
@@ -153,3 +153,16 @@ func TestScaleDoesntRecreate(t *testing.T) {
...
+	c := NewCLI(t)
+	const projectName = "compose-e2e-dependency-not-required"
+	t.Cleanup(func() {
+		c.RunDockerComposeCmd(t, "--project-name", projectName, "down")
+	})
+
+	res := c.RunDockerComposeCmd(t, "-f", "./fixtures/dependencies/deps-not-required.yaml", "--project-name", projectName,
+		"up", "-d")
+	assert.Assert(t, strings.Contains(res.Combined(), "foo"), res.Combined())
+	assert.Assert(t, !strings.Contains(res.Combined(), "bar"), res.Combined())
```
milas, line 167: From compose-spec/compose-spec#382:
> Compose only warns you when the dependency service isn't started or available.

```
❯ docker compose up
[+] Running 2/0
 ✔ Network required_default  Created                                                                                                                  0.0s
 ✔ Container required-foo-1  Created                                                                                                                  0.0s
Attaching to required-foo-1
required-foo-1  | foo
required-foo-1 exited with code 0
```

Shouldn't there be a warning "`foo` has optional dependency on `bar` which was not selected by active profiles" or something?

milas (commented):
Overall LGTM. I realize I missed the boat on https://github.com/compose-spec/compose-spec/pull/382 😅 but I do wonder if `optional: true` would be better semantically?

It's a bit weird to have the default/absence of a boolean field be `true` IMO 😬

pkg/e2e/up_test.go
glours: Right!

milas (approved):
LGTM!

---

Just a note: I do think we could do some future CLI work on the output to make startup around dependencies in particular clearer, e.g. I'd still love it if here we could convey "`foo` launched without `bar` because `not-required` profile is not active".

PR #10793, opened 2023-07-10
ci: speed up a couple Dockerfile targets w/ cache mount

milas:
**What I did**
The local Go package module path was missing from a couple of jobs, which made them slower than needed since they were re-downloading a bunch of dependencies.

In particular, this makes `make lint` waaaay faster!

jumping ferret: https://www.flickr.com/photos/usfwsmtnprairie/5244105513


glours (approved):
Alt Text

glours:
@milas is you DCO issue back? 😅

PR #10800, opened 2023-07-11
watch: fix missing events from symlinked path (e.g. tmp on macOS)

### LINKED ISSUE #10513: compose watch can't run from a /tmp on OSX
### Description

running `compose watch` from `/tmp` on OSX won't detect file updates.
the root cause is that this is actually a symlink to /private/tmp

see https://twitter.com/aurelien_lair/status/1652309539188875264

### Steps To Reproduce

_No response_

### Compose Version

_No response_

### Docker Environment

_No response_

### Anything else?

_No response_

milas:
**What I did**
Fix for watching for files in paths that have symlinks in them.

Resolve the project directory before joining it to the trigger path. Additionally, resolve symlinks in any absolute trigger paths.

There's probably (definitely) still some edge cases in here, but this should fix the most common case, which is temporary directories on macOS, which are symlinked to their real location under `/private`.

**Related issue**
* Fixes #10513 

camouflaged owl

### DIFF
--- a/pkg/compose/watch.go
+++ b/pkg/compose/watch.go
@@ -249,9 +249,19 @@ func loadDevelopmentConfig(service types.ServiceConfig, project *types.Project) 
 	if err != nil {
 		return nil, err
 	}
+	// TODO: compose-go should return the canonical path with all symlinks resolved
+	baseDir, err := filepath.EvalSymlinks(project.WorkingDir)
+	if err != nil {
+		return nil, fmt.Errorf("resolving symlink for %q: %w", project.WorkingDir, err)
+	}
 	for i, trigger := range config.Watch {
-		if !filepath.IsAbs(trigger.Path) {
-			trigger.Path = filepath.Join(project.WorkingDir, trigger.Path)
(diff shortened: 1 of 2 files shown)


pkg/compose/watch.go
```diff
@@ -249,9 +249,19 @@ func loadDevelopmentConfig(service types.ServiceConfig, project *types.Project)
 	if err != nil {
 		return nil, err
 	}
+	// TODO: compose-go should return the canonical path with all symlinks resolved
```
ndeloof, line 252: indeed.
As we can release compose-go any time (we use to just depend on HEAD, but _some_ requested we adopt versions) would it make better sense to fix this in `loader.absPath` and by the way make this method public so we can use it in watch ?

ndeloof (approved):
LGTM but I'd prefer we fix this in compose-go with a patch release

milas:
Switched this to draft - going to make the changes in `compose-go` as suggested and will likely end up closing this PR in favor of that

milas:
Closing in favor of:
 * compose-spec/compose-go#436

PR #10801, opened 2023-07-11
watch: add end-to-end test

milas:
**What I did**
Add an end-to-end test that covers the core watch functionality, i.e. CRUD on files & directories.

**Related issue**
This PR will need to be rebased after this is merged or the test won't pass on macOS, as it relies on using a temporary directory:
* #10800 

a snow leopard peeking over a branch


ndeloof (approved):
🤟

PR #10811, opened 2023-07-17
add support for git remote resources

ndeloof:
**What I did**
This demonstrates use of https://github.com/compose-spec/compose-go/pull/435

```yaml
include:
  - git@github.com:ndeloof/test.git
```

**Related issue**
https://docker.atlassian.net/browse/ENV-251


cmd/compose/compose.go
```diff
@@ -256,7 +263,7 @@ func RunningAsStandalone() bool {
 }
 
 // RootCommand returns the compose command with its child commands
-func RootCommand(streams command.Cli, backend api.Service) *cobra.Command { //nolint:gocyclo
+func RootCommand(steams command.Cli, backend api.Service) *cobra.Command { //nolint:gocyclo
```
milas, line 266: nit: `streams` got renamed to `steams` 😅

pkg/remote/git.go
```diff
@@ -0,0 +1,92 @@
...
+	"github.com/compose-spec/compose-go/cli"
+	"github.com/compose-spec/compose-go/loader"
+	"github.com/docker/cli/cli/config"
+	"github.com/docker/compose/v2/pkg/api"
+	"github.com/moby/buildkit/util/gitutil"
+	"github.com/opencontainers/go-digest"
+)
+
+func NewGitRemoteLoader() (loader.ResourceLoader, error) {
+	cache := filepath.Join(config.Dir(), "compose", "remotes")
```
milas, line 35: Should we consider using something like `XDG_CACHE_HOME`?

See also: https://pkg.go.dev/github.com/adrg/xdg#pkg-variables

pkg/remote/git.go
```diff
@@ -0,0 +1,92 @@
...
+}
+
+func (g gitRemoteLoader) Load(ctx context.Context, path string) (string, error) {
+	ref, err := gitutil.ParseGitRef(path)
+	if err != nil {
+		return "", err
+	}
+	path = fmt.Sprintf("%s#%s", ref.Remote, ref.Commit)
+	local := filepath.Join(g.cache, digest.FromString(path).Hex())
+	if _, err := os.Stat(local); os.IsNotExist(err) {
```
milas, line 58: Have we thought about the UX for updates here? If my `ref` is a branch (`main`), it's never going to get updated.

`--remote-update-policy=always`? 😅

milas:
Btw you can consider my review a LGTM, I just didn't approve (yet) since this is PR is blocked on waiting for new `compose-go` to get tagged with compose-spec/compose-go#435

ndeloof:
This really is just a prototype implementation used as demonstration purpose. questions like https://github.com/docker/compose/pull/10811#issuecomment-1640320237 are to be addressed - makes me wonder, how does builkit manage this scenario?

pkg/remote/git.go
ndeloof: I changed the implementation so it relies on `git ls-remote` to check remote for commit matching branch name, and checkout selected commit if not already known

pkg/remote/git.go
```diff
@@ -0,0 +1,145 @@
...
+}
+
+func (g gitRemoteLoader) Accept(path string) bool {
+	_, err := gitutil.ParseGitRef(path)
+	return err == nil
+}
+
+var commitSHA = regexp.MustCompile(`^[a-f0-9]{40}$`)
+
+func (g gitRemoteLoader) Load(ctx context.Context, path string) (string, error) {
```
milas, line 63: Doing background Git fetches is always tricky because of credential helpers.

Take a look at https://github.com/tilt-dev/go-get/blob/295ccc6550921a9766a7294360a77a4d8704d826/get.go#L18-L55

I'd clone `os.Environ()` and then conditionally add `GIT_TERMINAL_PROMPT` / `GIT_SSH_COMMAND` to the env passed to the exec'd processes.

milas (changes requested):
A few robustness suggestions but nothing major.

My "blocking" concern right now is gating this somehow since it's not formally part of the spec and we're still experimenting somewhat here.

pkg/remote/git.go
ndeloof: If Env is nil, the new process uses the current process's environment, so we don't have to

milas (approved):

crazy-max:
There should be entitlements for this feature and I don't think there is any in this implementation. An attacker could potentially read any files from the host if remote repo is compromised such as:

```yaml
include:
  - git@github.com:hacked/test.git
```

Compromised compose file on remote repo:

```yaml
services:
  db:
    image: haxxor-mariadb:latest
    volumes:
      - "db:/var/lib/mysql"
      - "$HOME/.aws/credentials:/aws_credentials"
    restart: always

volumes:
  db:
```

`haxxor-mariadb:latest` would then send `/aws_credentials` to a CNC server.

ndeloof:
@crazy-max sure, but AFAICT this is the case for any form of "sharing compose file" scenario. Any suggestion ?

Issue #10813, opened 2023-07-17
[BUG] `build --push` attempts to push images for services without `image` defined


milas:
### Description

If a service does not have an `image` field, Compose generates a default image name in the format `$project-$service`. This should never be attempted to be pushed, as it'll always fail, since it implicitly resolves to `docker.io/library/$project-$service`.

### Steps To Reproduce

**compose.yaml**
```yaml
services:
  app:
    build: .
```

**Dockerfile**
```dockerfile
FROM alpine
```

1. Run `docker compose build --push`

```
 => [app] exporting to image                                                                                                                          0.0s
 => => exporting layers                                                                                                                               0.0s
 => => writing image sha256:b247771ad3b311c4fc85551871712e872e34c5319c2c23b955b200375f994ec0                                                          0.0s
 => => naming to docker.io/library/build-app                                                                                                          0.0s
 => ERROR [app] pushing build-app with docker                                                                                                         1.2s
 => => pushing layer 0a13d2aaa54c                                                                                                                     0.9s
 => => pushing layer 45437bbd87f2                                                                                                                     0.9s
 => => pushing layer 7cd1e5cbf124                                                                                                                     0.9s
 => => pushing layer ad6517b0c914                                                                                                                     0.9s
 => => pushing layer 4e6bef96e37e                                                                                                                     0.9s
 => => pushing layer c58d5a26ffa8                                                                                                                     1.0s
 => => pushing layer efd1965f1684                                                                                                                     1.0s
------
 > [app] pushing build-app with docker:
------
denied: requested access to the resource is denied
```

### Compose Version

```Text
v2.20.0
```

### Docker Environment

```Text
Occurs with all Engine versions
```

PR #10814, opened 2023-07-17
build: do not attempt to push unnamed service images

milas:
**What I did**
When building, if images are being pushed, ensure that only named images (i.e. services with a populated `image` field) are attempted to be pushed.

Services without `image` get an auto-generated name, which will be a "Docker library" reference since they're in the format `$project-$service`, which is implicitly the same as `docker.io/library/$project-$service`. A push for that is never desirable / will always fail.

The key here is that we cannot overwrite the `<svc>.image` field when doing builds, as we need to be able to check for its presence to determine whether a push makes sense.

**Related issue**
* #10813

otters on a water slide

### DIFF
--- a/cmd/compose/config.go
+++ b/cmd/compose/config.go
@@ -233,11 +233,7 @@ func runConfigImages(streams api.Streams, opts configOptions, services []string)
 		return err
 	}
 	for _, s := range project.Services {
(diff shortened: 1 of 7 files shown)


glours (approved):
LGTM

PR #10815, opened 2023-07-17
introduce publish (alpha) command

ndeloof:
**What I did**
introduce `publish` command - this is just command skeleton, backend is missing an implementation

**Related issue**
this is skeleton code toward a `publish` command to implement https://github.com/compose-spec/compose-spec/issues/288
https://docker.atlassian.net/browse/ENV-249

image

### DIFF
--- a/cmd/compose/publish.go
+++ b/cmd/compose/publish.go
@@ -1,0 +1,49 @@
+/*
+   Copyright 2020 Docker Compose CLI authors
+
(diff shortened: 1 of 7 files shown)


milas (approved):

ndeloof:
Pushed to https://github.com/docker/compose/tree/publish so we can iterate on implementing this feature

PR #10816, opened 2023-07-17
test: speed up the e2e test suite

milas:
**What I did**
Lots of our phony Compose files launch pointless long-lived processes so we can assert on state. However, this means they often don't respond well to signals on their own, requiring Compose to timeout and kill them when doing a `down`.

Add in lots of `init: true` where appropriate so that we don't block for no reason while running E2E tests all over the place.

Additionally, a couple tests have gotten a cleanup so they don't leave behind containers. I still want to build this into the framework in the future, but this is easier for the moment and won't cause any trouble in the future.

flying squirrel mid-air

### DIFF
--- a/pkg/e2e/compose_exec_test.go
+++ b/pkg/e2e/compose_exec_test.go
@@ -35,6 +35,12 @@ func TestLocalComposeExec(t *testing.T) {
 		return ret
 	}
 
+	cleanup := func() {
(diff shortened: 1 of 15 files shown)


milas:
the flex tape slap meme where it's applying `init: true` to a Compose service

ndeloof (approved):

laurazard (approved):

Issue #10817, opened 2023-07-18
[BUG] Secrets on Windows Container mounted through Linux path


dev-anton-ko:
### Description

It seems secrets feature is not working on Windows Containers.

### Steps To Reproduce

Here's docker-compose.yml file.
Secrets > target has been added to illustrate the problem.
```
version: "3.8"

secrets:
  db_connection_string:
    file: db_connection_string.txt

services:
  test:
    image: mcr.microsoft.com/dotnet/aspnet:6.0
    environment:
      - DB_CONNECTION_STRING=C:\ProgramData\docker\secrets\db_connection_string
    secrets:
        - source: db_connection_string
          target: C:\ProgramData\docker\secrets\db_connection_string
```
docker compose output:
`Error response from daemon: invalid mount config for type "bind": invalid mount path: '/run/secrets/C:\ProgramData\docker\secrets\db_connection_string'`

### Compose Version

```Text
Docker Compose version v2.20.0
```

### Docker Environment

```Text
Client:
 Version:           24.0.4
 API version:       1.43
 Go version:        go1.20.5
 Git commit:        3713ee1
 Built:             Fri Jul  7 14:52:09 2023
 OS/Arch:           windows/amd64
 Context:           default

Server: Docker Engine - Community
 Engine:
  Version:          24.0.4
  API version:      1.43 (minimum version 1.24)
  Go version:       go1.20.5
  Git commit:       4ffc614
  Built:            Fri Jul  7 14:51:02 2023
  OS/Arch:          windows/amd64
  Experimental:     false
```

PR #10819, opened 2023-07-18
check secret target is an absolute windows path

ndeloof:
**What I did**
As we check secret target is absolute, we need to consider both unix and windows absolute path syntaxes
we _could_ be more deterministic relying on image platform to only run the relevant code, but doesn't seem there are major risks for confusion

**Related issue**
fixes https://github.com/docker/compose/issues/10817

### DIFF
--- a/pkg/compose/create.go
+++ b/pkg/compose/create.go
@@ -863,7 +863,7 @@ func buildContainerConfigMounts(p types.Project, s types.ServiceConfig) ([]mount
 		target := config.Target
 		if config.Target == "" {
 			target = configsBaseDir + config.Source
-		} else if !isUnixAbs(config.Target) {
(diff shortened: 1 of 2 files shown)


milas (approved):
We do have a version of this that looks like it might be slightly more robust in `compose-go` already: https://github.com/compose-spec/compose-go/blob/95ac1be8bfdc4dc21cdfb14e37b52c31275046a3/loader/windows_path.go#L32-L43

(and it has test coverage 😉)

I think we should export & clean up the naming (`s/isAbs/IsWindowsAbs/`) from `compose-go` and switch to them here, but this should unblock all but the most pathological cases so LGTM 👍🏻

glours (approved):

PR #10826, opened 2023-07-18
cmd: add project load instrumentation

milas:
**What I did**
Add a new `project-load` span which includes high-level details based on parsing YAML.

Most of this PR is plumbing `context.Context` through to where it's needed.


milas:
Decided I didn't like this approach, much happier with #10831

PR #10828, opened 2023-07-18
pkg/compose: RunOneOffContainer: don't use NewStartOptions()

thaJeztah:
It's no longer used in docker/cli, and doesn't do anything other than creating an empty struct, so replacing it (as we're planning to deprecate that function)

**What I did**

**Related issue**


milas (approved):
LGTM, thanks!

glours (approved):

PR #10843, opened 2023-07-24
Improve buildkit node creation

silvin-lubecki:
**What I did**
Move builder and nodes initialization code up, avoiding to recreate/load them for every service build.
Added support of `BUILDX_BUILDER` env var.

**Related issue**

image


pkg/compose/build.go
```diff
@@ -68,6 +69,27 @@ func (s *composeService) build(ctx context.Context, project *types.Project, opti
...
+	// Initialize buildkit nodes
+	var (
+		b     *builder.Builder
+		nodes []builder.Node
+	)
+	if buildkitEnabled {
+		builderName := options.Builder
+		if builderName == "" {
+			builderName = os.Getenv("BUILDX_BUILDER")
+		}
```
milas, line 81: Hmm, so we already do a look up in `cmd/build.go`:
https://github.com/docker/compose/blob/8318f66330358988a058bf611a953f464ab7973f/cmd/compose/build.go#L58-L61

But that won't apply for `up --build` (while this will).

@glours Do you think that `--builder` should become a global flag on Compose, and let the CLI/cmd part handle the env var lookup?

[I don't think we need to resolve this as part of this PR]

milas (approved):
LGTM

Note to Compose team: we should make sure to mention that `BUILDX_BUILDER` is now supported for `up --build` as well in the release notes (see also my non-blocking review comment)

PR #10852, opened 2023-07-28
cmd/compose: use reflect.Type to check Cancel context

gocurr:
The original implemention uses hardcoded string to check whether ctx is not a Cancel context. I think it can be avoid
by using reflect.Type to compare. The context package doesn't provide funcs such as 'IsCancelCtx', but I noticed
it uses 'switch ctx := c.(type)' to distinguish ctxs. Maybe this PR is not worthless.

Please take a look, thanks.

### DIFF
--- a/cmd/compose/compose.go
+++ b/cmd/compose/compose.go
@@ -22,6 +22,7 @@ import (
 	"os"
 	"os/signal"
 	"path/filepath"
+	"reflect"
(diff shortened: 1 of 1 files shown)


gocurr:
Any comments? If this PR is worthless, it will be closed.

gocurr:
Thanks.

PR #10860, opened 2023-08-01
watch: support multiple containers for tar implementation

milas:
**What I did**
Support services with scale > 1 for the tar watch sync.

Add a "lossy" multi-writer specific to pipes that writes the tar data to each `io.PipeWriter`, which is connected to `stdin` for the `tar` process being exec'd in the container.

The data is written serially to each writer. This could be adjusted to do concurrent writes but that will rapidly increase the I/O load, so is not done here - in general, 99% of the time you'll be developing (and thus using watch/sync) with a single replica of a service.

If a write fails, the corresponding `io.PipeWriter` is removed from the active set and closed with an error.

This means that a single container copy failing won't stop writes to the others that are succeeding. Of course, they will be in an inconsistent state afterwards still, but that's a different problem.

4 cats that all look very similar

### DIFF
--- a/internal/sync/tar.go
+++ b/internal/sync/tar.go
@@ -80,9 +80,6 @@ func (t *Tar) Sync(ctx context.Context, service types.ServiceConfig, paths []Pat
 		}
 	}
 
-	// TODO: this can't be read from multiple times
-	tarReader := tarArchive(pathsToCopy)
-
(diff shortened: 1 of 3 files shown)


glours (approved):
LGTM, thanks for the test 👍

Issue #10851, opened 2023-07-28
[BUG] depends_on.required: false triggers exit status warning


ShahimEssaid:
### Description

I updated to v2.20.2 to try out the required property. It is working but it's also triggering a warning that I can't explain. See the minimum example.

### Steps To Reproduce

**1.** Run the following example in Ubuntu 22.04 and Compose 2.20.2
2. Example:
```
services:
  first:
    image: ubuntu
    command:
      - echo
      - hello from first
  second:
    image: ubuntu
    depends_on:
      first:
        condition: service_completed_successfully
        required: false
    command:
      - echo
      - hello from second
```
3. Output
```
06:58 $ docker compose up
[+] Running 2/0
 ✔ Container tmp-first-1   Created                                                                                                                                                                                                                                      0.0s 
 ✔ Container tmp-second-1  Created                                                                                                                                                                                                                                      0.0s 
Attaching to tmp-first-1, tmp-second-1
tmp-first-1   | hello from first
tmp-first-1 exited with code 0
WARN[0000] optional dependency "first" didn't complete successfully: exit 0 
tmp-second-1  | hello from second
tmp-second-1 exited with code 0
```
4. I do not understand the output line: `WARN[0000] optional dependency "first" didn't complete successfully: exit 0 `

### Compose Version

```Text
06:54 $ docker compose version
Docker Compose version v2.20.2
```


### Docker Environment

```Text
06:58 $ docker info
Client: Docker Engine - Community
 Version:    24.0.5
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.11.2
    Path:     /usr/libexec/docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.20.2
    Path:     /usr/libexec/docker/cli-plugins/docker-compose

Server:
 Containers: 5
  Running: 1
  Paused: 0
  Stopped: 4
 Images: 82
 Server Version: 24.0.5
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
[...]

PR #10870, opened 2023-08-03
up: do not warn on successful optional dependency complete

milas:
**What I did**
If an optional dependency exits successfully (exit code of 0), with a service condition of `service_completed_successfully`, don't log a warning.

**Related issue**
 * #10851

frog floating in water


glours (approved):
Nice 👍

ndeloof (approved):

PR #10877, opened 2023-08-03
watch: enable tar-based syncer by default

milas:
**What I did**
Swap the default implementation now that batching is merged. Keeping the `docker cp` based implementation around for the moment, but it needs to be _explicitly_ disabled now by setting `COMPOSE_EXPERIMENTAL_WATCH_TAR=0`.

After the next release, we should remove the `docker cp` implementation entirely.

capybara in a hot tub


glours (approved):
👍

PR #10881, opened 2023-08-04
Display builder's name on the first build line.

silvin-lubecki:
**What I did**
Added on the top right side the builder name, to get the same UX as buildx.

Screenshot 2023-08-04 at 10 32 19

**Related issue**

image


glours (approved):
LGTM

Issue #10803, opened 2023-07-12
[BUG] Proxy config not applied as build args with classic builder


komima:
### Description

Proxy options from `config.json` are not set as build args when using the classic builder (`DOCKER_BUILTKIT=0`).

### Steps To Reproduce

1. Setup a simple compose project with a to-be-built image, which would need proxy environment variables at build-time

![image](https://github.com/docker/compose/assets/58747243/74fd2a4a-e49b-4186-a976-246597aaa5b8)

2. Have proxies set in `config.json`
3. Run `docker-compose build`
4. Observe that proxy environment variables are unset

2.16.0 still has args set:

![image](https://github.com/docker/compose/assets/58747243/05d8e72d-1beb-4aa1-888a-590c789fd777)

2.17.0 (and 2.20.0) do not:

![image](https://github.com/docker/compose/assets/58747243/d3d64d19-8e4d-4490-892e-1f1fce354804)

### Compose Version

```Text
2.16.0 works, 2.17.0 & 2.20.0 do not work
```

### Docker Environment

```Text
podman 4.4.1 with `DOCKER_BUILDKIT=0` and `DOCKER_HOST=<podman rootless socket>` set for compose to access the compat api
```

### Anything else?

Support for `config.json` proxies to build args added in https://github.com/docker/compose/pull/10159

Since https://github.com/docker/compose/pull/10374 (released in 2.17.0) `toBuildOptions` seems to be [no longer called](https://github.com/docker/compose/blob/852e19282094e74e81734ca0c25bf6a72db632f2/pkg/compose/build.go#L91) in the `if !buildkitEnabled` code, and proxies are populated from config [in that code](https://github.com/docker/compose/blob/852e19282094e74e81734ca0c25bf6a72db632f2/pkg/compose/build.go#L299).

milas:
Thanks for the report! This is fixed in Compose v2.21.

PR #10887, opened 2023-08-08
build: fix missing proxy build args for classic builder

milas:
**What I did**
Refactor to use a consistent code path for determining the build args for a service image regardless of whether BuildKit or the classic builder is being used.

After recent changes, these code paths had diverged, so the classic builder was missing the proxy variables from the Docker client config.

**Related issue**
* #10803
* #10876 

a very cute weasel on a tree branch looking at the camera

### DIFF
--- a/pkg/compose/build.go
+++ b/pkg/compose/build.go
@@ -22,18 +22,19 @@ import (
 	"os"
 	"path/filepath"
 
-	"github.com/docker/buildx/builder"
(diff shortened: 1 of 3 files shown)


milas:
GitHub broke on #10876, it wasn't seeing the rebase with the fixed merge conflict, so I closed it, and now I can't re-open it (because NOW it could see the rebase) 🤦🏻

glours (approved):

PR #10890, opened 2023-08-10
upgrade Golang to 1.21

glours:
**What I did**
Bump Golang to version `1.21`

**Related issue**
N/A

image


ndeloof (approved):

Dockerfile
```diff
@@ -15,7 +15,7 @@
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
 
-ARG GO_VERSION=1.20.7
+ARG GO_VERSION=1.21
```
milas, line 18:
```suggestion
ARG GO_VERSION=1.21.0
```

We've generally been pinning to the patch as well, right? 

That said...I'm fine with using the rolling tag (`1.21`) if you want, despite the `Host` issues recently, Go patch upgrades are generally very safe IMO, so up to you!

milas (approved):

cpuguy83:
Why?
Please add reasons _why_ this is changed not just that it's changed, that's plain to see from the code that there's a change.

ndeloof:
@cpuguy83 why would we want to keep an older version ?

cpuguy83:
Bumping the go.mod means everyone who compiles it has to bump the go version.
What would you not support a still in-support go version?

Anyway, the point is not to say "you shouldn't have done this" but "what's the reason for the decision?".

ndeloof:
> everyone who compiles it has to bump the go version

obviously! But if you exclude a few Linux distribution maintainers who have to compile on their own, the many others just consume Docker Compose as a pre-built binary and aren't affected.

> what's the reason for the decision?

because we want to rely on latest available and can benefits potential optimizations or new constructs it offers. We don't wait to see a coding opportunity to bump version.

Issue #10902, opened 2023-08-15
[BUG] docker-compose 2.20.3 rewriting environment variables


grahamb:
### Description

This sounds very strange, but it appears that docker-compose v2.20.3 is rewriting environment variables, specifically http_proxy and https_proxy (and their upper-case variants).

I have http(s)_proxy and no_proxy variables set in my shell, in ~/.docker/config.json, and in /etc/systemd/system/docker.service.d/http-proxy.conf

My shell variables:

```
❯ env | grep -i proxy                                          
https_proxy=http://bby-vcontrol-proxy.its.sfu.ca:8080
http_proxy=http://bby-vcontrol-proxy.its.sfu.ca:8080
HTTPS_PROXY=http://bby-vcontrol-proxy.its.sfu.ca:8080
HTTP_PROXY=http://bby-vcontrol-proxy.its.sfu.ca:8080
NO_PROXY=localhost,127.0.0.1,10.28.130.135,sfu.ca,www.sfu.ca,*.sfu.ca,*.its.sfu.ca,.sfu.ca,.its.sfu.ca,its.sfu.ca,github.sfu.ca,sentry.its.sfu.ca,twilio.com
no_proxy=localhost,127.0.0.1,10.28.130.135,sfu.ca,www.sfu.ca,*.sfu.ca,*.its.sfu.ca,.sfu.ca,.its.sfu.ca,its.sfu.ca,github.sfu.ca,sentry.its.sfu.ca,twilio.com
```

I have a very simple test Dockerfile:

```Dockerfile
FROM alpine:latest
RUN env | grep -i proxy
```

If I build the image with Docker (not docker-compose), the proxy variables output by the build are what I would expect:

```
❯ docker buildx build --tag local/dev:latest --progress plain --no-cache .
#0 building with "default" instance using docker driver

#1 [internal] load .dockerignore
#1 transferring context: 210B done
#1 DONE 0.0s

#2 [internal] load build definition from Dockerfile
#2 transferring dockerfile: 1.05kB done
#2 DONE 0.0s

#3 [internal] load metadata for docker.io/library/alpine:latest
#3 DONE 0.3s

#4 [1/2] FROM docker.io/library/alpine:latest@sha256:7144f7bab3d4c2648d7e59409f15ec52a18006a128c733fcff20d3a4a54ba44a
#4 CACHED

#5 [2/2] RUN env | grep -i proxy
[...]

PR #10908, opened 2023-08-17
clone variable berfore we capture a pointer

ndeloof:
**What I did**
fix a for loop variable collision issue

**Related issue**
fixes https://github.com/docker/compose/issues/10902


milas (approved):
😭

Issue #10874, opened 2023-08-03
`docker compose ps --format` should be configurable by `~/.docker/config.json`.


FeryET:
### Description

Hi.

The current ps format is too verbose for my taste. I usually only care about service names, status and (sometimes) binded ports.

I have changed the default `dockerPs` format in `config.json`, but compose does not respect this. While I can live without this feature, having this will save me from aliasing my compose ps command in every VM shell config that I create.

PR #10918, opened 2023-08-22
align `docker compose ps` with `docker ps`

ndeloof:
**What I did**
Use the docker/cli formatter to implement `docker compose ps` so that `--format` offers the same templating options
Note: this is a breaking change as `--format "json"` used to produce a json array, while docker/cli produces Newline-separated json stream.

**Related issue**
closes https://github.com/docker/compose/issues/10874
https://docker.atlassian.net/browse/ENV-282


milas (approved):
LGTM! I only ran into one issue, which is there's no truncation on `COMMAND`.

Current:
```
❯ docker compose ps -a
NAME                IMAGE               COMMAND                  SERVICE             CREATED             STATUS              PORTS
pstest-fib-1        debian              "bash -c 'function f…"   fib                 3 seconds ago       Up 3 seconds        0.0.0.0:57185->34123/tcp
pstest-sleep-1      debian              "sleep infinity"         sleep               2 minutes ago       Up 2 minutes
```

This PR:
```
❯ ~/dev/compose/bin/build/docker-compose ps -a
NAME           IMAGE     COMMAND   SERVICE   CREATED   STATUS    PORTS
pstest-fib-1   debian    bash -c 'function fib(){
    if [ $1 -le 0 ]; then
        echo 0
    elif [ $1 -eq 1 ]; then
        echo 1
    else
        echo $[`fib $[$1-2]` + `fib $[$1 - 1]` ]
    fi

}

fib $THIS_IS_THE_FIBONACCI_NUMBER
'                fib       13 seconds ago   Up 12 seconds   0.0.0.0:57185->34123/tcp
pstest-sleep-1   debian    sleep infinity   sleep           3 minutes ago   Up 3 minutes
```
(yes I copy pasted a shell script to calculate the fibonacci sequence as an inline `command` in Compose YAML)

`docker ps`:
```
❯ docker ps
CONTAINER ID   IMAGE         COMMAND                  CREATED          STATUS          PORTS                      NAMES
0657d4b4fcd6   debian        "bash -c 'function f…"   37 seconds ago   Up 36 seconds   0.0.0.0:57185->34123/tcp   pstest-fib-1
1b9380a354f1   debian        "sleep infinity"         3 minutes ago    Up 3 minutes                               pstest-sleep-1
```

```yaml
name: pstest

services:
  fib:
    image: debian
    init: true
    ports:
      - 34123
    command:
      - bash
      - -c
      - |
        function fib(){
            if [ $$1 -le 0 ]; then
                echo 0
            elif [ $$1 -eq 1 ]; then
                echo 1
            else
                echo $$[`fib $$[$$1-2]` + `fib $$[$$1 - 1]` ]
            fi

        }

        fib $$THIS_IS_THE_FIBONACCI_NUMBER
    environment:
      - THIS_IS_THE_FIBONACCI_NUMBER=${THIS_IS_THE_FIBONACCI_NUMBER:-100}
  sleep:
    image: debian
    init: true
    command: [sleep, infinity]
```

ndeloof:
@milas indeed, missed a copy/paste from docker/cli :P
Also added support for other attributes exposed by `docker ps` which we didn't included in the `backend.Ps` API (yet another legacy from ECS integration...)

ndeloof:
> Does Docker compose follow semver?

nope, at least not in the sense "no single backward incompatible change will happen without a major version bump"

ndeloof:
> JSON Statham

😅


> Now, apparently, it is:
> docker compose ps --no-trunc --format=json

indeed.
This just aligns with `docker ps --format json`, so this is not a new thing json output truncates IDs, whenever this is not designed for human usage and long ID would be a reasonable default.

thaJeztah:
Related to the above;

- https://github.com/docker/cli/issues/674

PR #10949, opened 2023-08-30
introduce publish (alpha) command

ndeloof:
This introduce new `docker compose (alpha) publish REPOSITORY` command to publish a compose.yaml file as an OCI artifact, and support for `docker:<REPOSITORY>` remote resource to be used with `include`

illustration example:
```yaml
include:
  - docker:ndeloof/test:demo
```

```
$ COMPOSE_EXPERIMENTAL_OCI_REMOTE=true docker compose config --no-normalize
name: truc
services:
  test:
    image: nginx
```

note: most of the changes in this PR is about passing `dockerCli` to the project loader

### DIFF
--- a/cmd/compose/compose.go
+++ b/cmd/compose/compose.go
@@ -26,18 +26,16 @@ import (
 	"strings"
 	"syscall"
 
-	buildx "github.com/docker/buildx/util/progress"
-
-	"github.com/compose-spec/compose-go/dotenv"
-	"github.com/docker/cli/cli/command"
-	"github.com/docker/compose/v2/pkg/remote"
-
 	"github.com/compose-spec/compose-go/cli"
+	"github.com/compose-spec/compose-go/dotenv"
 	"github.com/compose-spec/compose-go/types"
 	composegoutils "github.com/compose-spec/compose-go/utils"
 	"github.com/docker/buildx/util/logutil"
+	buildx "github.com/docker/buildx/util/progress"
 	dockercli "github.com/docker/cli/cli"
 	"github.com/docker/cli/cli-plugins/manager"
+	"github.com/docker/cli/cli/command"
+	"github.com/docker/compose/v2/pkg/remote"
 	"github.com/morikuni/aec"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
@@ -119,6 +117,7 @@ type ProjectOptions struct {
 	EnvFiles      []string
 	Compatibility bool
 	Progress      string
+	Offline       bool
 }
 
 // ProjectFunc does stuff within a types.Project
@@ -128,14 +127,14 @@ type ProjectFunc func(ctx context.Context, project *types.Project) error
 type ProjectServicesFunc func(ctx context.Context, project *types.Project, services []string) error
 
 // WithProject creates a cobra run command from a ProjectFunc based on configured project options and selected services
-func (o *ProjectOptions) WithProject(fn ProjectFunc) func(cmd *cobra.Command, args []string) error {
-	return o.WithServices(func(ctx context.Context, project *types.Project, services []string) error {
+func (o *ProjectOptions) WithProject(fn ProjectFunc, dockerCli command.Cli) func(cmd *cobra.Command, args []string) error {
+	return o.WithServices(dockerCli, func(ctx context.Context, project *types.Project, services []string) error {
 		return fn(ctx, project)
 	})
 }
 
 // WithServices creates a cobra run command from a ProjectFunc based on configured project options and selected services
-func (o *ProjectOptions) WithServices(fn ProjectServicesFunc) func(cmd *cobra.Command, args []string) error {
+func (o *ProjectOptions) WithServices(dockerCli command.Cli, fn ProjectServicesFunc) func(cmd *cobra.Command, args []string) error {
 	return Adapt(func(ctx context.Context, args []string) error {
 		options := []cli.ProjectOptionsFn{
 			cli.WithResolvedPaths(true),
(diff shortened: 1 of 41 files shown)


cmd/compose/compose.go
```diff
@@ -201,14 +188,38 @@ func (o *ProjectOptions) toProjectName() (string, error) {
...
-	project, err := o.ToProject(nil)
+	project, err := o.ToProject(dockerCli, nil)
 	if err != nil {
 		return "", err
 	}
 	return project.Name, nil
 }
 
-func (o *ProjectOptions) ToProject(services []string, po ...cli.ProjectOptionsFn) (*types.Project, error) {
+func (o *ProjectOptions) ToProject(dockerCli command.Cli, services []string, po ...cli.ProjectOptionsFn) (*types.Project, error) {
```
milas, line 198: This gets used as part of shell completion, which ~probably~ definitely shouldn't try to clone Git repos to handle auto-complete.

What do you think about having a `ToProjectOffline` (which does _not_ take the `dockerCli` / attempt to create loaders), which could be used by `ValidArgsFunction`?

(If it returns an error since the project requires remote resources, shell completion will not return any results, which seems fine)

pkg/remote/oci.go
```diff
@@ -0,0 +1,135 @@
...
+	}, err
+}
+
+type ociRemoteLoader struct {
+	cache     string
+	dockerCli command.Cli
+}
+
+func (g ociRemoteLoader) Accept(path string) bool {
+	return strings.HasPrefix(path, "docker:")
```
milas, line 72: I believe elsewhere (e.g. build) we've used `docker-image://` as the format

milas: I guess this isn't technically an image, but a new artifact type, at which point I wonder if `oci://` would be better?

(I know flux2 uses that format for example: https://fluxcd.io/flux/cheatsheets/oci-artifacts/#consuming-artifacts)

milas (commented):
Logic LGTM, just a couple higher-level things

pkg/remote/oci.go
ndeloof: `oci` would be a better terminology from a pure implementation point of view, but I don't expect more than 1% Docker users know about Open Container Initiative, and just have a basic understanding of Docker images

cmd/compose/compose.go
milas: Yeah, if the "offline" mode could use cache, that'd be pretty neat, but for simplicity, it might be fine to just ignore for the moment. My real concern is hitting `<tab>` and then having Compose start cloning a huge monorepo in the background by surprise 😬

pkg/remote/oci.go
milas: Definitely agree that most people don't know what OCI is 😛 

I would still say that:
1. We should at least use a URL scheme, i.e. `foo://bar` instead of `foo:bar`, being able to do a cheap check for `://` to distinguish between local & remote (without awareness of loader) is a helpful property
2. `docker` isn't really descriptive and could mean many different things/is un-Googleable, but "Docker Compose OCI include" will [hopefully] get results

milas (approved):

milas:
Oh no, I caused a bunch of conflicts with the build changes...sorry 😭

pkg/remote/oci.go
BretFisher: helm also uses `oci://` for modern helm push/pull

glours (approved):
Sounds good to me

go.mod
```diff
@@ -50,6 +50,8 @@ require (
 	gotest.tools/v3 v3.5.0
 )
 
+require github.com/distribution/distribution/v3 v3.0.0-20230214150026-36d8c594d7aa
```
milas, line 53: See #10954 - will need to swap things to `github.com/distribution/reference` to keep this from reappearing.

pkg/api/api.go
```diff
@@ -74,6 +74,8 @@ type Service interface {
 	Events(ctx context.Context, projectName string, options EventsOptions) error
 	// Port executes the equivalent to a `compose port`
 	Port(ctx context.Context, projectName string, service string, port uint16, options PortOptions) (string, int, error)
+	// Publish executes the equivalent to a `compose publish`
+	Publish(ctx context.Context, project *types.Project, repository string) error
```
milas, line 78: Minor: we often end up refactoring the API methods to add an options object later...maybe worth having `PublishOptions` with a `Repository string` field upfront

milas: e.g. I'm guessing this will need `Quiet bool` at some point

pkg/remote/oci.go
```diff
@@ -0,0 +1,145 @@
...
+	}, err
+}
+
+type ociRemoteLoader struct {
+	cache     string
+	dockerCli command.Cli
+	offline   bool
+}
+
+const prefix = "oci:"
```
milas, line 73: Shouldn't this be `oci://`?

milas (commented):
LGTM, few last things but mostly the `distribution/distribution/v3` dep that snuck back 😅

pkg/api/api.go
ndeloof: Repository being mandatory parameter should not be part of the `Options` struct imho

pkg/remote/oci.go
ndeloof: This is not an URL, just a prefix we decide on our own. There's no canonical URL style for OCI artifacts, as this ends being transported over http

jaydrogers:
This feature seems very exciting! I found the documentation on this feature (https://docs.docker.com/engine/reference/commandline/compose_alpha_publish/) but are there more resources to learn how to use it and play around with it? 

**My questions:**
1. Do I need to do anything special to create a template?
2. When I publish it, where does it go?
3. How can I have someone else use it?
4. Are there plans for it to work with Swarm?

ndeloof:
`docker compose -f <your compose file> alpha publish <registry>/<repo>:<tag>`

then you can use this published model from another compose file using `include: - oci:<registry>/<repo>:<tag>`

> Are there plans for it to work with Swarm?

swarm doesn't use the compose specification nor compose-go to parse compose files.
.. at least until we get https://github.com/docker/cli/pull/4863/ approved

Issue #10948, opened 2023-08-29
[BUG] restarting single service that has a profile fails with 'no such service'


ngfeldman:
### Description

When I run `docker compose restart <service>` on a service that has `profiles` set to a non-empty list, the command fails with `no such service: <service>`. I would expect the single service to restart successfully in this case.

By contrast, the following all work just fine:

1. `docker compose up <service>` when the service belongs to a profile
2. `docker compose down <service>` when the service belongs to a profile
3. `docker compose restart <service>` when the service does not belong to a profile.

### Steps To Reproduce

1. On MacOS Ventura 13.5.1 with Docker Desktop for Mac 4.21.1 (Compose v2.19.1)
2. With this docker-compose.yml:
    ```yaml
    version: "3.9"
    services:
      foo:
        profiles:
          - a
        image: "bash"
        command: ["sleep", "inf"]
      bar:
        image: "bash"
        command: ["sleep", "inf"]
    ```
3. Run `docker compose up -d` and then `docker restart foo`
4. See error `no such service: foo`

### Compose Version

```Text
Docker Compose version v2.19.1
```

### Docker Environment

```Text
Client:
 Version:    24.0.2
 Context:    desktop-linux
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.11.0
    Path:     /Users/nfeldman/.docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.19.1
    Path:     /Users/nfeldman/.docker/cli-plugins/docker-compose
  dev: Docker Dev Environments (Docker Inc.)
    Version:  v0.1.0
    Path:     /Users/nfeldman/.docker/cli-plugins/docker-dev
  extension: Manages Docker extensions (Docker Inc.)
    Version:  v0.2.20
    Path:     /Users/nfeldman/.docker/cli-plugins/docker-extension
  init: Creates Docker-related starter files for your project (Docker Inc.)
    Version:  v0.1.0-beta.6
    Path:     /Users/nfeldman/.docker/cli-plugins/docker-init
  sbom: View the packaged-based Software Bill Of Materials (SBOM) for an image (Anchore Inc.)
    Version:  0.6.0
    Path:     /Users/nfeldman/.docker/cli-plugins/docker-sbom
  scan: Docker Scan (Docker Inc.)
    Version:  v0.26.0
    Path:     /Users/nfeldman/.docker/cli-plugins/docker-scan
[...]

PR #10952, opened 2023-08-31
Enable service explicitly requested to be restarted

ndeloof:
**What I did**
loading local compose.yaml file, if service(s) are explicitly set by command args, automagically enable profiles
also moved "no deps" logic to backend code otherwise running without a project cause a nil pointer panic.

**Related issue**
fixes https://github.com/docker/compose/issues/10948

### DIFF
--- a/cmd/compose/restart.go
+++ b/cmd/compose/restart.go
@@ -20,7 +20,6 @@ import (
 	"context"
 	"time"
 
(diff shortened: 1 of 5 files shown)


milas (approved):
LGTM - would be nice to update `restart` e2e test to cover this 😉

PR #10954, opened 2023-08-31
migrate to github.com/distribution/reference

thaJeztah:
- depends on https://github.com/docker/compose/pull/10953
- relates to https://github.com/distribution/distribution/pull/4031

**What I did**

The "reference" package was moved to a separate module, which was extracted
from https://github.com/distribution/distribution/commit/b9b19409cf458dcb9e1253ff44ba75bd0620faa6

**Related issue**


thaJeztah:
Looks like the distribution-types made their way into compose's signatures, so we need to wait for distribution to have that PR merged, and I'll also open a PR in compose-go

```
#21 142.4 cmd/main.go:33:2: could not import github.com/docker/compose/v2/pkg/compose (-: # github.com/docker/compose/v2/pkg/compose
#21 142.4 pkg/compose/compose.go:154:32: cannot use func(named reference.Named) (digest.Digest, error) {…} (value of type func(named "github.com/distribution/reference".Named) (digest.Digest, error)) as func(named "github.com/distribution/distribution/v3/reference".Named) (digest.Digest, error) value in argument to project.ResolveImages) (typecheck)
#21 142.4 	"github.com/docker/compose/v2/pkg/compose"
#21 142.4 	^
```

thaJeztah:
@ndeloof I kept it in draft because there was no tagged version of `compose-go` with this patch; do you want to have a tagged version first?

(:warning: not sure if that should be `1.18.x` or if we want to be on the safe side, and to `1.19.0` so that we could still do `1.18.x` patch releases _without_ this switch)

glours (approved):

PR #10957, opened 2023-08-31
watch: build & launch the project at start

milas:
**What I did**
The `alpha watch` command current "attaches" to an already-running Compose project, so it's necessary to run something like `docker compose up --wait` first.

Now, we'll do the equivalent of an `up --build` before starting the watch, so that we know the project is up-to-date and running.

Additionally, unlike an interactive `up`, the services are not stopped when `watch` exits (e.g. via `Ctrl-C`). This prevents the need to start from scratch each time the command is run - if some services are already running and up-to-date, they can be used as-is. A `down` can always be used to destroy everything, and we can consider introducing a flag like `--down-on-exit` to `watch` or changing the default.

NOTE: This requires #10956, so it won't even build until that's merged 😬 

**Related issue**
https://docker.atlassian.net/browse/ENV-286

lynx in the snow

### DIFF
--- a/cmd/compose/watch.go
+++ b/cmd/compose/watch.go
@@ -31,10 +31,14 @@ import (
 type watchOptions struct {
 	*ProjectOptions
 	quiet bool
+	noUp  bool
 }
 
 func watchCommand(p *ProjectOptions, dockerCli command.Cli, backend api.Service) *cobra.Command {
-	opts := watchOptions{
+	watchOpts := watchOptions{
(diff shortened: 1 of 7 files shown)


milas:
This is rebased and good to go!

(Sidenote: IntelliJ's local history feature saved me because I accidentally deleted most of this when splitting it from #10956 😬)

ndeloof (approved):

PR #10974, opened 2023-09-05
OTEL: adding flags to cli traces

rvigus:
**What I did**
Adds cli args and flags (e.g. `--timeout`, `--dry-run`, etc)  as attributes to cli trace data. These attributes are then picked up by the collector and made available  to the end user.

Related issue
https://docker.atlassian.net/browse/ENV-276


cmd/cmdtrace/cmd_span.go
```diff
@@ -53,6 +54,9 @@ func Setup(cmd *cobra.Command, dockerCli command.Cli) error {
 		ctx,
 		"cli/"+strings.Join(commandName(cmd), "-"),
 	)
+	cmdSpan.SetAttributes(attribute.StringSlice("cliArgs", args))
+	cmdSpan.SetAttributes(attribute.StringSlice("cliFlags", getFlags(cmd.Flags())))
```
milas, line 58: Since span attributes are a single-level map, so far I've been separating logical groups with `.`, e.g. `cli.args` / `cli.flags`.

(I'm realizing I should add a `TRACING.md` to the repo here with some info...serious 🚌 factor right now)

milas (changes requested):
Nice! LGTM, just a pedantic naming style request that's not your fault since I never documented it anywhere 😛

milas:
Ah, for your failures:

* The DCO failure has the explanation, but basically you have to `git commit --amend -s` to sign-off and then force-push your PR
* You need to add the license to the top of the test file, you can copy+paste it from another. (There's also a way to configure GoLand to insert this automatically, but I always do it manually 🙈)
  * Run `make validate-headers` to ensure it's good
* Can run lint locally with `golangci-lint run` if you have it installed or `make lint` to let it do it in a container

Issue #10965, opened 2023-09-04
[Feature Request] `docker compose up --scale` without side effects


lonix1:
### Description

I originally asked this on [SO](https://stackoverflow.com/q/77009955/9971404) without resolution, so I assume it's not possible, and hence this feature request.

### Description

I have a number of services in a docker stack, where some are started manually only.

`docker-compose.yml`:
```yml
services:

  other:
    # ...

  app:
    # ...
    depends_on:
     - postgres
    deploy:
      replicas: 0

  postgres:
    # ...
    deploy:
      replicas: 0

  pgadmin:
    # ...
    deploy:
      replicas: 0
```

I can start `app` and `postgres` like this:
```console
docker compose up -d --scale app=1 --scale postgres=1
```

If I then start `pgadmin`:
```console
docker compose up -d --scale pgadmin=1
```

...it will stop `app` and `postgres` and then start `pgadmin`. I don't want to stop them.

I assume I can't scale a service without impacting existing services. The only reason I can imagine is the swarm reasoning: you tell compose what state you want, and it scales services to get to that state. BUT: 1) I'm not using swarm, and 2) why should that be the rule?

*This is a very common use case*, and the only workaround is complex hard-to-maintain scripts.

Please allow us to scale services without side effects. For example via a `--scale-independently` switch:
```console
docker compose up -d --scale pgadmin=1 --scale-independently
```
Docker would realise that other services are already running, so that command would be equivalent to:
```console
docker compose up -d --scale app=1 --scale postgres=1 --scale pgadmin=1
```

glours:
Hey @lonix1 
I have a question for you, did you consider using [`profiles`](https://github.com/compose-spec/compose-spec/blob/master/15-profiles.md) instead of setup all your services with `replicas: 0` ?
Profiles were typically design for this kind of use cases, being able to start services only when you explicitly need/want

glours:
To avoid `compose up` to stop your previous started services you can explicitly ask for the `pgadmin` service to start, like this : 
```
> docker compose up -d --scale app=1 --scale postgres=1
[+] Building 0.0s (0/0)                                                                                                                                                                                                   docker:desktop-linux
[+] Running 4/4
 ✔ Network issue-10965_default       Created                                                                                                                                                                                              0.0s
 ✔ Container issue-10965-postgres-1  Started                                                                                                                                                                                              0.0s
 ✔ Container issue-10965-other-1     Started                                                                                                                                                                                              0.0s
[...]

PR #10979, opened 2023-09-06
add scale command

glours:
**What I did**
Add a `scale` command

**Related issue**
https://docker.atlassian.net/browse/ENV-310
https://github.com/docker/compose/issues/10965

image

### DIFF
--- a/cmd/compose/scale.go
+++ b/cmd/compose/scale.go
@@ -1,0 +1,108 @@
+/*
+   Copyright 2020 Docker Compose CLI authors
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package compose
+
+import (
+	"context"
+	"strconv"
+	"strings"
+
+	"github.com/docker/cli/cli/command"
+
+	"github.com/compose-spec/compose-go/types"
+	"github.com/pkg/errors"
+	"golang.org/x/exp/maps"
+
+	"github.com/docker/compose/v2/pkg/api"
+	"github.com/spf13/cobra"
+)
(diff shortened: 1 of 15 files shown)


ndeloof:
`scale` used to be a docker-compose v1 command, so I don't think we need this to be introduced as an alpha command, just resurrect the legacy one
```
$ docker-compose-v1 scale --help
Set number of containers to run for a service.

Numbers are specified in the form `service=num` as arguments.
For example:

    $ docker-compose scale web=2 worker=3
```

milas:
Noting that we should update docs after merging this: https://docs.docker.com/compose/migrate/#unsupported-in-v2

cmd/compose/scale.go
```diff
@@ -0,0 +1,108 @@
...
+
+	return backend.Scale(ctx, project, api.ScaleOptions{Services: services})
+}
+
+func parseServicesReplicasArgs(args []string) (map[string]int, error) {
+	serviceReplicaTuples := map[string]int{}
+	for _, arg := range args {
+		key, val, ok := strings.Cut(arg, "=")
+		if !ok || key == "" || val == "" {
+			return nil, errors.Errorf("Invalide scale specifier %q.", arg)
```
milas, line 98:
```suggestion
			return nil, errors.Errorf("invalid scale specifier: %s", arg)
```

cmd/compose/scale.go
```diff
@@ -0,0 +1,108 @@
...
+		}
+		intValue, err := strconv.Atoi(val)
+
+		if err != nil {
+			return nil, errors.Errorf("Invalide scale specifier, can't parse replicate value to int %q.", arg)
```
milas, line 103:
```suggestion
			return nil, errors.Errorf("invalid scale specifier: can't parse replica value as int: %v", arg)
```

milas (approved):
LGTM

glours:
Ok my test assertions are fragile as we can't bet which container will be removed 😬 
```
scale_test.go:54: Running command: docker compose --project-directory fixtures/scale scale dbadmin=1
    scale_test.go:55: assertion failed: expression is false: strings.Contains(res.Combined(), "Container scale-basic-tests-dbadmin-1  Running"):  Container scale-basic-tests-dbadmin-2  Running
```
I'll improve this tomorrow

glours:
@milas I think the 2 replicas were created at the same time 🤷‍♂️ 
Anyway it's fix now and assertions are way more robust!

ndeloof (approved):

PR #10980, opened 2023-09-06
test: add a fully containerized (via Compose) e2e setup

milas:
**What I did**
`dind.compose.yaml` has a Docker-in-Docker setup, which is then referenced via `include` in `e2e.compose.yaml`.

Not all tests currently pass, as some rely on bind mounts, for example.

falcon


glours (approved):
Nice 😊 🎉

ndeloof:
obsolete

Issue #10977, opened 2023-09-06
docker compose pull has no options for pull policy


FFdhorkin:
### Description

I was having issues with my docker compose project when I tried to run it from within a Github Action. Seems like for some reason, on Github, I get rate limited for pulls, but don't locally.

This led to me coming up with the workaround of running `docker compose --parallel 1 pull` before the `docker compose up`

I noticed that even though the [docker cache github action](https://github.com/marketplace/actions/docker-cache) was showing corresponding cached images, the `docker compose --parallel 1 pull` was resulting in images being pulled even though they were already in the cache (I was still getting occasional rate limit pull failures on cached images)

In other words, there's no way to say `docker compose pull --missing-only` the way you can do `docker compose up --pull missing`

This led to me attempting the following even hackier workaround:
```
docker compose --parallel 1 up --pull missing --no-start
# the above line breaks docker compose up; I guess it's expecting docker compose start, but I need to use my makefile task
docker compose rm --force --volumes
# the following line, among other things, runs docker compose up
make setup
```

I'm not happy with this solution, but `COMPOSE_PARALLEL_LIMIT=1 make setup` would cause performance issues. Doing `docker compose --parallel 1 pull --ignore-pull-failures` also is not an ideal solution, since it will attempt unnecessary pulls.

## TL;DR
there should be some sort of flag for `docker compose pull` that corresponds to `docker compose up --pull ____` so that you can pull only missing images

glours:
Hey @FFdhorkin 
Did you try add a step in your workflow to be logged and avoid to be rate limited? You can try to use our official action for this https://github.com/docker/login-action

thaJeztah:
@FFdhorkin do you have more details about your issue? 

- Perhaps more information about your compose file as well (for example, does it have pull-policies set in the compose file?)


> Seems like for some reason, on Github, I get rate limited for pulls, but don't locally.

Do you have more information about how the "rate limit" exposed itself? Do you have the error message (if any) that you got? (want to be sure these are Docker Hub rate limits, or other rate-limits: I've seen cases where GitHub itself was producing `429` statuses).

w.r.t.

>  the `docker compose --parallel 1 pull` was resulting in images being pulled even though they were already in the cache (I was still getting occasional rate limit pull failures on cached images)

and

> ```bash
> docker compose --parallel 1 up --pull missing --no-start
> # the above line breaks docker compose up; I guess it's expecting docker compose start, but I need to use my makefile task
> ```


[...]

PR #10981, opened 2023-09-07
introduce pull --missing flag to only pull images not present in cache

ndeloof:
**What I did**
introduce `--missing` flag on `docker compose pull` to only pull images not present in cache. 
Note: "latest" is always pulled

**Related issue**
closes https://github.com/docker/compose/issues/10977


thaJeztah:
> Note: "latest" is always pulled

Is there a reason for that?


Overall, I'm still a  bit curious about the  general issue that led to opening https://github.com/docker/compose/issues/10977, and if it's an isolated event / one specific scenario that doesn't affect the wider group of users.

`docker compose up` should already be able to pull what's needed (based on pull policy defined), so for most (if not _all_) scenarios "it should just work".

To some extent, I would expect `docker compose pull` to follow the pull-policy that's defined in the compose-file, and a `--force` option to _force_ it to download  / refresh all images (even if present locally)

ndeloof:
@thaJeztah afaict this is inherited from docker compose v1
`docker compose pull` applies the pull policy, this flags just allows to globally override

ndeloof:
we also could use `--policy=missing` so that we don't need additional flags

FFdhorkin:
> @thaJeztah afaict this is inherited from docker compose v1 `docker compose pull` applies the pull policy, this flags just allows to globally override

sounds like there must be a bug, then.

> compose v1 of course didn't have pull-policies, so I'm a bit curious what the reporter's compose-file looks like;
> 
>     * if the compose file has `missing` as pull policy, then `docker compose --parallel=1 pull` would pull duplicate images once (first service using the image would pull it, second one would (should) see it's there, so skip pulling)
> 
>     * ^^ without `--parallel=1`, ideally, compose would still de-duplicate images (2 services using the same image would not be doing two separate pulls)

I replied on the original issue, but the [docs](https://docs.docker.com/compose/compose-file/05-services/#pull_policy) say `pull_policy` defaults to `missing`. I don't have any explicit pull_policy set.

> I'd like to understand the cause of the issue; perhaps we're adding a new feature, whereas there's s bug somewhere, and adding a `--missing`, would lead to adding an `--always` and a `--never` and `--unless-built` (and there's already a `--ignore-pull-failures` and .... an explosion of new options

Yeah, if the intended behavior is to respect `pull_policy`, I don't really see much reason to add a `--missing` option, as opposed to figuring out why it either isn't respecting the `pull_policy` or the defaults aren't being applied, etc.

> we also could use `--policy=missing` so that we don't need additional flags

That seems a lot more useful, IMHO, than `--missing`, since it'd be much more flexible

glours (approved):

cmd/compose/pull.go
```diff
@@ -83,6 +85,13 @@ func runPull(ctx context.Context, dockerCli command.Cli, backend api.Service, op
 		}
 	}
 
+	if opts.policy != "" {
+		for i, service := range project.Services {
+			service.PullPolicy = opts.policy
```
milas, line 90: I think this should ignore services that either have `pull_policy: build` already set in YAML or have a `<svc>.build` section and no `<svc>.image` section.

e.g. imagine this project
```yaml
services:
  explicit-build:
    image: registry.example.com/myservice
    pull_policy: build
    build: .
  no-image:
    build: .
  nginx:
    image: nginx
```

Without args:
```
❯ docker compose pull
[+] Pulling 3/3
 ✔ explicit-build Skipped                                                                                                                             0.0s
 ✔ no-image Skipped - No image to be pulled                                                                                                           0.0s
 ✔ nginx Pulled
```

With `--policy=missing` (or `--policy=always`):
```
❯ ~/dev/compose/bin/build/docker-compose pull --policy=missing
[+] Pulling 3/3
 ✔ no-image Skipped - No image to be pulled                                                                                                           0.0s
 ✔ nginx Pulled                                                                                                                                       0.5s
 ! explicit-build Warning                                                                                                                             0.2s
WARNING: Some service image(s) must be built from source by running:
    docker compose build explicit-build
1 error occurred:
	* Error response from daemon: Get "https://registry.example.com/v2/": dialing registry.example.com:443 with direct connection: resolving host registry.example.com: lookup registry.example.com: no such host
```

So it is properly skipping the one without `<svc>.image`, but is unusable if you have any `pull_policy: build` services that do specify an image name.

cmd/compose/create.go
```diff
@@ -80,8 +80,8 @@ func createCommand(p *ProjectOptions, dockerCli command.Cli, backend api.Service
 	}
 	flags := cmd.Flags()
 	flags.BoolVar(&opts.Build, "build", false, "Build images before starting containers.")
-	flags.BoolVar(&opts.noBuild, "no-build", false, "Don't build an image, even if it's missing.")
-	flags.StringVar(&opts.Pull, "pull", "missing", `Pull image before running ("always"|"missing"|"never")`)
+	flags.BoolVar(&opts.noBuild, "no-build", false, "Don't build an image, even if it's policy.")
+	flags.StringVar(&opts.Pull, "pull", "policy", `Pull image before running ("always"|"policy"|"never")`)
```
milas, line 84: We have these args/descriptions across several commands, and I think we need to improve the copy here, especially since we're making changes, as they're pretty vague and unclear as-is.

cc @aevesdocker 

My understanding of desired behaviors:
 * `--build`: Re-build images for all services being launched that have a `build` section. Images with `pull_policy: build` are always re-built unless `--no-build` is specified.
 * `--no-build`: Don't build ANY images, which might mean the project can't start if local versions are not already available in the engine.
 * `--policy`
   * `always`
     * for services with `build` section, fetch the latest base image (`FROM`), similar to `docker build --pull .`
     * for services without a `build` section, fetch the `image` even if it's already present locally (e.g. useful for `:latest`)
   * `never`
      * for services with a `build` section, no-op (AFAIK there's no builder option to say "don't automatically pull missing base images" nor would it make much sense)
      * for services without a `build` section, don't pull the `image` even if it doesn't exist in the engine, which might mean the project can't launch.

milas (commented):
No real concerns with the code, but I think there's more to think through here around `pull_policy: build`.

---

Also...I'd love some form of tests here as a result, even if it's for the `createOptions::apply` so that it can be a cheap unit test to cover some of the different possibilities

cmd/compose/pull.go
ndeloof: Agree about service without an `image` MUST be excluded
disagree on `pull_policy: build`: we should only require a local build if image is not available on registry under configured image

cmd/compose/create.go
ndeloof: > `always` for services with build section, fetch the latest base image (FROM), 

Nope, fetch the configured tagged image, which is the result of the build. We don't parse the Dockerfile to know about the base image. No relation with `docker build --pull`

thaJeztah:
How  was this merged with pending questions, and force-pushes after the last review? (Note that the commit message and PR title also don't match the implementation.

ndeloof:
@thaJeztah AFAICT all comments have been addressed, did I missed something
force push is used as we prefer rebase over merge commits

thaJeztah:
> force push is used as we prefer rebase over merge commits

I'm familiar with rebase over merge commits. (although not sure if a rebase is _always_ needed (of course in case there's a conflict)); if the concern is that merging an outdated branch _could_ in some cases make the main branch fail, then perhaps the merge-queue feature would work.

Rebases are ok (if they're needed), but at least the last force push mad actual code changes, and those changes were _not_ reviewed, and were _not_ called out;
https://github.com/docker/compose/compare/0c2a6c8ce74644a746e46299b4748fe4b41923f6..4e9ff18d62fbf6fe538a8ea61c7db283894fbfce

Screenshot 2023-09-19 at 13 33 04


Those changes look fine, but I'm not a fan of doing so if not needed (force-push and self-merge), especially not if it's not a critical PR to get merged. If such a fix gets included in a force-push-and-merge, what's keeping me from including _other_  changes? (without calling them out or having them reviewed).

> AFAICT all comments have been addressed, did I missed something

There were still questions pending, and your answers were not acknowledged by the persons asking them.


My other issue is that we picked "low-hanging-fruit" here, _without fully triaging the request_. This PR was created for a _single_ report for a _single_ user, and after further questioning may have been a valid _bug_ report, written as a "feature request", as acknowledged by the original reporter; https://github.com/docker/compose/pull/10981#issuecomment-1710690110 

- the original issue reported something what _looks like_ a potential _bug_
- but did so in the form of an "xy problem" (https://xyproblem.info)
   - something didn't work (the bug)
   - I tried X to work around the problem (the bug)
   - But X didn't work, so can you add Y to make my workaround X for a bug work?

And what we did was "implement Y to make the workaround for a bug work".

It was not clear yet if there was an actual need for the new flag, and we didn't fix the bug, only more flags to workaround a bug.

ndeloof:
That's just the way we use to manage PRs and issues triage here, I'm not saying this is perfect or can't introduce regressions (this happened to us) but that's our way. Up to you to adopt a stricter problem solving protocol on the repo you're responsible for

cmd/compose/up.go
```diff
@@ -101,8 +101,8 @@ func upCommand(p *ProjectOptions, dockerCli command.Cli, backend api.Service) *c
 	flags := upCmd.Flags()
 	flags.BoolVarP(&up.Detach, "detach", "d", false, "Detached mode: Run containers in the background")
 	flags.BoolVar(&create.Build, "build", false, "Build images before starting containers.")
-	flags.BoolVar(&create.noBuild, "no-build", false, "Don't build an image, even if it's missing.")
-	flags.StringVar(&create.Pull, "pull", "missing", `Pull image before running ("always"|"missing"|"never")`)
+	flags.BoolVar(&create.noBuild, "no-build", false, "Don't build an image, even if it's policy.")
+	flags.StringVar(&create.Pull, "pull", "policy", `Pull image before running ("always"|"policy"|"never")`)
```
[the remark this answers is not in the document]
thaJeztah: We should look at that by the way, because if `build` uses a custom builder (which may be remote), we would now potentially be;

- pulling an image locally
- run a build (which won't use the image we pulled)
- load the image from the builder (replacing the image we just pulled)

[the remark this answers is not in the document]
ndeloof: `docker compose build --pull` just pass the pull flag to the builder, same as `docker build --pull`

PR #10982, opened 2023-09-07
cli: fix `--build` flag for `create`

milas:
**What I did**
I missed this during a refactor and there wasn't test coverage. Instead of adding more heavy-weight integration tests, I tried to use `gomock` here to assert on the options objects after CLI flag parsing. I think with a few more helpers, this could be a good way to get a lot more combinations covered without adding a ton of slow E2E tests.

**Related issue**
Introduced here:
* #10956

skunk


glours (approved):
LGTM

PR #10992, opened 2023-09-11
ci: tweak restricted imports in linter

milas:
**What I did**
* Eliminate direct dependency on `gopkg.in/yaml.v2`
* Add `gopkg.in/yaml.v2` as a restricted import
* Add `github.com/distribution/distribution` as a restricted dependency in favor of `distribution/reference` which is the subset of functionality that Compose needs
* Remove an unused exclusion

**Related issue**
n/a

a small lizard


glours (approved):
👍

ndeloof (approved):

Issue #10997, opened 2023-09-13
[BUG] network not found when using network_mode:service with --no-deps


nicks:
### Description

network not found when using network_mode:service with --no-deps

### Steps To Reproduce

1. docker-compose.yaml

```
services:
  app-x:
    image: busybox
    init: true
    command:
    - "sleep"
    - "1000"
    network_mode: service:app-y
      
  app-y:
    image: busybox
    init: true
    command:
    - "sleep"
    - "1000"
 ```

2. Run `docker compose up -d`
3. Run `docker compose up --no-deps app-x`

Expected behavior:
app-x recreates successfully

Actual behavior:
```
Attaching to scratch-app-x-1
Error response from daemon: network service:app-y not found
```

### Compose Version

```Text
docker compose version
Docker Compose version v2.21.0-desktop.1
```

### Docker Environment

```Text
Client: Docker Engine - Community
 Version:    24.0.5
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.11.0-cloud-driver+002
    Path:     /home/nick/.docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.21.0-desktop.1
    Path:     /usr/lib/docker/cli-plugins/docker-compose
  dev: Docker Dev Environments (Docker Inc.)
    Version:  v0.1.0
    Path:     /usr/lib/docker/cli-plugins/docker-dev
  extension: Manages Docker extensions (Docker Inc.)
    Version:  v0.2.20
    Path:     /usr/lib/docker/cli-plugins/docker-extension
  harmonia: A brief description of Harmonia (Docker Inc.)
    Version:  1014447
    Path:     /home/nick/.docker/cli-plugins/docker-harmonia
  init: Creates Docker-related starter files for your project (Docker Inc.)
    Version:  v0.1.0-beta.7
    Path:     /usr/lib/docker/cli-plugins/docker-init
  sbom: View the packaged-based Software Bill Of Materials (SBOM) for an image (Anchore Inc.)
    Version:  0.6.0
    Path:     /usr/lib/docker/cli-plugins/docker-sbom
  scan: Docker Scan (Docker Inc.)
    Version:  v0.26.0
    Path:     /usr/lib/docker/cli-plugins/docker-scan
  scout: Command line tool for Docker Scout (Docker Inc.)
    Version:  0.24.1
[...]

Issue #10996, opened 2023-09-13
[BUG] unexpected 'no such service' error when --no-deps and volumes_from used together


nicks:
### Description

unexpected 'no such service' error when --no-deps and volumes_from used together

### Steps To Reproduce

1. Create the following `docker-compose.yaml`:

```

services:
  app:
    image: busybox
    command:
    - "sleep"
    - "1000"
    volumes_from:
    - db

  db:
    image: mysql:5.7.37
    environment:
      MYSQL_ROOT_PASSWORD: root
      MYSQL_DATABASE: DOCKERDB
    volumes:
      - db-data:/var/lib/mysql

volumes:
  db-data:

```

2. Run `docker compose up -d`
3. Run `docker compose up --no-deps app`

Expected Result:
The app service starts successfully, without restarting db.

Actual result:
```
docker compose up --no-deps app
no such service: db
```

### Compose Version

```Text
docker compose version
Docker Compose version v2.21.0-desktop.1
```

### Docker Environment

```Text
Client: Docker Engine - Community
 Version:    24.0.5
 Context:    desktop-linux
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.11.0-cloud-driver+002
    Path:     /home/nick/.docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.21.0-desktop.1
    Path:     /usr/lib/docker/cli-plugins/docker-compose
  dev: Docker Dev Environments (Docker Inc.)
    Version:  v0.1.0
    Path:     /usr/lib/docker/cli-plugins/docker-dev
  extension: Manages Docker extensions (Docker Inc.)
    Version:  v0.2.20
    Path:     /usr/lib/docker/cli-plugins/docker-extension
  harmonia: A brief description of Harmonia (Docker Inc.)
    Version:  1014447
    Path:     /home/nick/.docker/cli-plugins/docker-harmonia
  init: Creates Docker-related starter files for your project (Docker Inc.)
    Version:  v0.1.0-beta.7
    Path:     /usr/lib/docker/cli-plugins/docker-init
  sbom: View the packaged-based Software Bill Of Materials (SBOM) for an image (Anchore Inc.)
    Version:  0.6.0
    Path:     /usr/lib/docker/cli-plugins/docker-sbom
  scan: Docker Scan (Docker Inc.)
    Version:  v0.26.0
[...]

PR #10999, opened 2023-09-14
don't rely on depends_on to resolve volume_from, better use observed state

ndeloof:
**What I did**
remove preemptive parsing of `volumes_from` doing convergence based on known dependencies, better rely on observed state to select target container. This also removes the assumption we will be able to use container number `1`

**Related issue**
closes https://github.com/docker/compose/issues/10996
closes https://github.com/docker/compose/issues/10997


glours (approved):
LGTM

glours (approved):

milas (approved):

PR #11008, opened 2023-09-15
implement publish

ndeloof:
**What I did**
implement `publish` command, to publish compose.yaml files as pseudo-image "layers" with a custom artifact type

### DIFF
--- a/pkg/compose/publish.go
+++ b/pkg/compose/publish.go
@@ -18,24 +18,136 @@ package compose
 
 import (
 	"context"
+	"encoding/json"
+	"os"
 
 	"github.com/compose-spec/compose-go/types"
 	"github.com/distribution/reference"
+	"github.com/docker/buildx/util/imagetools"
 	"github.com/docker/compose/v2/pkg/api"
+	"github.com/docker/compose/v2/pkg/progress"
+	"github.com/opencontainers/go-digest"
+	"github.com/opencontainers/image-spec/specs-go"
+	v1 "github.com/opencontainers/image-spec/specs-go/v1"
 )
 
 func (s *composeService) Publish(ctx context.Context, project *types.Project, repository string, options api.PublishOptions) error {
+	return progress.RunWithTitle(ctx, func(ctx context.Context) error {
+		return s.publish(ctx, project, repository, options)
+	}, s.stdinfo(), "Publishing")
+}
+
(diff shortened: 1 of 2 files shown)


pkg/compose/publish.go
```diff
@@ -18,24 +18,138 @@ package compose
...
+		}
+
+		w.Event(progress.Event{
+			ID:     file,
+			Text:   "publishing",
+			Status: progress.Working,
+		})
+		layer := v1.Descriptor{
+			MediaType: v1.MediaTypeImageLayer,
+			Digest:    digest.FromString(string(f)),
```
milas, line 71:
```suggestion
			Digest:    digest.FromBytes(f),
```

Internally, `digest.FromString` will just convert back to bytes so might as well pass it directly

pkg/compose/publish.go
```diff
@@ -18,24 +18,138 @@ package compose
...
+			Size:      int64(len(f)),
+			Annotations: map[string]string{
+				"com.docker.compose": api.ComposeVersion,
+			},
+			ArtifactType: "application/vnd.docker.compose.file",
```
milas, line 76: I wonder if this should have a `+yaml`

pkg/compose/publish.go
```diff
@@ -18,24 +18,138 @@ package compose
...
 
-	return api.ErrNotImplemented
+	err = resolver.Push(ctx, named, v1.Descriptor{
+		MediaType: v1.MediaTypeImageManifest,
+		Digest:    digest.FromString(string(imageManifest)),
+		Size:      int64(len(imageManifest)),
+		Annotations: map[string]string{
+			"com.docker.compose": api.ComposeVersion,
+		},
+		ArtifactType: "application/vnd.docker.compose",
```
milas, line 139: `.project` maybe?

milas (approved):
LGTM, I bikeshed all the names but they also seem fine as-is.

One thing worth mentioning is that OCI projects won't work (with `include`) if they have external file dependencies like `<svc>.build` or `<svc>.env_file`, since it'll look in the temp/cache download directory when trying to load. I don't think that needs to block this, though

ndeloof:
also adopted `oci://` as a prefix to align with [helm](https://helm.sh/docs/topics/registries/)

glours (approved):
LGTM

PR #11010, opened 2023-09-15
Make hashes between `up` and `configure` consistent

### LINKED ISSUE #10907: [BUG] config --hash return incorrect hash value for containers with depends_on
### Description

The hash computed by `docker compose config --hash service` does not match the hash that is actually used when creating the container. In our setup we compare the hash generated by `config --hash` with the hash set on the container to detect if its config is stale. This used to work fine until compose 2.19. `docker compose up -d` does not recreate the container, but the value of the label does not match the value computed by `config --hash`.

### Steps To Reproduce

Use the following `docker-compose.yml`:
```
version: '2'
services:
  test-container:
    container_name: "test-container"
    image: "redis:alpine"
  test-container-2:
    depends_on: ["test-container"]
    container_name: "test-container-2"
    image: "redis:alpine"
```

Run the following commands and observe the different values:
```
docker compose up -d
docker compose config --hash test-container-2
docker inspect test-container-2 -f '{{index .Config.Labels "com.docker.compose.config-hash"}}'
```

### Compose Version

```Text
Docker Compose version v2.19.1
```

### Docker Environment

```Text
 Version:    24.0.5
 Context:    default
 Debug Mode: false
 Plugins:
    Version:  v0.11.2
    Version:  v2.19.1
    Path:     /usr/local/libexec/docker/cli-plugins/docker-compose

Server:
 Containers: 9
  Running: 9
  Paused: 0
  Stopped: 0
 Images: 9
 Server Version: 24.0.5
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 1
 Plugins:
  Volume: local
 Swarm: inactive
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 8165feabfdfe38c65b599c4993d227328c231fca
 runc version: v1.1.8-0-g82f18fe
 init version: de40ad0
 Security Options:
  seccomp
   Profile: builtin
 Kernel Version: 4.18.0-477.21.1.el8_8.x86_64
 Operating System: AlmaLinux 8.8 (Sapphire Caracal)
 OSType: linux
 Architecture: x86_64
 CPUs: 2
 Total Memory: 6.783GiB
 Name: tkh-server1.keyhub-test-nw
 ID: da84eca3-73fd-42ea-9e3c-d56aa94d8c4f
 Debug Mode: false
 Experimental: false
 Insecure Registries:
  127.0.0.0/8
 Live Restore Enabled: true
```

### Anything else?

_No response_

mattwalo32:
**What I did**
Currently, when `docker compose config --hash [service]` computes a hash for the service, it deletes all other services before computing the hash. That means if a service depends on another service, the `DependsOn` field gets deleted, causing the hash to be different than when `docker up` creates the labels.

I prevented the dependencies from getting deleted before hash computation to fix this.

I also noticed that `docker up` does some modification to the platform of the image **before** doing the hash (I.e. if no platform was specified in the yaml file, the default one is set). `docker config` doesn't do this at all. So I modified `docker config` to apply these modifications before computing the hash. 

It could be argued that `config` shouldn't call `applyPlatform`. Take the scenario where a compose file contains services that don't have a platform specified. If that same file is run on two different machines with different default platforms, the hashes will be different. Personally, I think the hashes for the identical files should be different because they'll run on different platforms, but would love to get some feedback on that.

**Related issue**
Fixes #10907 

image

### DIFF
--- a/cmd/compose/config.go
+++ b/cmd/compose/config.go
@@ -186,6 +186,10 @@ func runHash(ctx context.Context, dockerCli command.Cli, opts configOptions) err
 		return err
 	}
 
+	if err := applyPlatforms(project, true); err != nil {
+		return err
+	}
+
 	if len(services) > 0 {
 		err = project.ForServices(services, types.IgnoreDependencies)
 		if err != nil {
--- a/cmd/compose/config.go
+++ b/cmd/compose/config.go
@@ -190,20 +190,19 @@ func runHash(ctx context.Context, dockerCli command.Cli, opts configOptions) err
 		return err
 	}
 
-	if len(services) > 0 {
-		err = project.ForServices(services, types.IgnoreDependencies)
+	sorted := services
+	sort.Slice(sorted, func(i, j int) bool {
+		return sorted[i] < sorted[j]
+	})
+
+	for _, name := range sorted {
+		s, err := project.GetService(name)
 		if err != nil {
 			return err
 		}
-	}
 
-	sorted := project.Services
-	sort.Slice(sorted, func(i, j int) bool {
-		return sorted[i].Name < sorted[j].Name
-	})
-
-	for _, s := range sorted {
 		hash, err := compose.ServiceHash(s)
+
 		if err != nil {
 			return err
 		}


ndeloof (approved):

ndeloof:
nice catch! and thanks for investigating this issue

AkihiroSuda:
Commit a345515f91c7bc6a7822fc8668cda65734584890 broke `docker compose config --hash=*`
- #11145

PR #11021, opened 2023-09-19
move watch from alpha to main command

glours:
**What I did**
Move `watch` command away from the `alpha` cmd umbrella

**Related issue**
N/A

image


milas (approved):

cmd/compose/alpha.go
```diff
@@ -31,7 +31,6 @@ func alphaCommand(p *ProjectOptions, dockerCli command.Cli, backend api.Service)
 		},
 	}
 	cmd.AddCommand(
-		watchCommand(p, dockerCli, backend),
```
ndeloof, line 34: shall we (temporary) keep this alpha command with a warning for user to adopt the official one ?

ndeloof (approved):

cmd/compose/alpha.go
glours: Yes sure 👍

ndeloof (approved):

PR #11026, opened 2023-09-20
use official `develop` section to configure watch command

ndeloof:
**What I did**
use https://github.com/compose-spec/compose-go/pull/461 to configure watch command with `develop``
user is warned about deprecated `x-develop` field. A future Docker Compose release will drop support for this

**Related issue**

image


glours:
You also need to update `pkg/compose/watch_test.go` with the new types from `compose-go`

PR #11042, opened 2023-09-25
deps: remove deprecated github.com/pkg/errors

mmorel-35:
**What I did**

Remove github.com/pkg/errors and replace it with errors or fmt.
It also updates golangci-lint with errorlint and gomodguard configuration to make sure it complies with the best practices.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -7,6 +7,7 @@ linters:
   enable:
     - depguard
     - errcheck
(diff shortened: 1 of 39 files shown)


ndeloof (approved):
LGTM
I will miss `errors.Wrap` as for me `%w` in message is not strictly the same semantic, but I can live with that :)

Issue #10993, opened 2023-09-12
[BUG] `--profile` breaking changes


maciejpankanin:
### Description

Imagine a stack like this:
```yml
services:
  db:
    image: postgres:13.3
    environment:
      POSTGRES_PASSWORD: example
      POSTGRES_USER: example
      POSTGRES_DB: example
    depends_on:
      hello-world:
        condition: service_completed_successfully

  hello-world:
    image: hello-world
    profiles:
      - hello
```

In previous versions of `docker compose` I could run `docker compose up` and it would start the `db` container without any problems. Recently I upgraded `docker` and was really surprised that it didn't work anymore. Here is the output of `docker compose up':
```text
service hello-world is required by db, but is disabled. Can be enabled using profiles [hello]
```

This is a devastating change for many compose stacks, and wasn't even mentioned in the changelog :(

There is also another surprising behaviour when running `docker compose logs db` reports:
```text
no such service: hello-world
```

### Steps To Reproduce

1. Run `docker compose up` with the config above.

### Compose Version

```Text
Any > 2.17.3(?)
```

### Docker Environment

```Text
Client: Docker Engine - Community
 Version:    24.0.5
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.11.2
    Path:     /usr/libexec/docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.20.2
    Path:     /usr/libexec/docker/cli-plugins/docker-compose

Server:
 Containers: 16
  Running: 0
  Paused: 0
  Stopped: 16
 Images: 266
 Server Version: 24.0.5
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 8165feabfdfe38c65b599c4993d227328c231fca
 runc version: v1.1.8-0-g82f18fe
 init version: de40ad0
 Security Options:
  apparmor
  seccomp
[...]

PR #11050, opened 2023-09-29
some commands don't need Dependencies being checked

ndeloof:
**What I did**
some commands like `docker compose logs SERVICE` don't have to validate dependencies are well defined in project model

**Related issue**
fixes https://github.com/docker/compose/issues/10993

### DIFF
--- a/cmd/compose/compose.go
+++ b/cmd/compose/compose.go
@@ -118,6 +118,7 @@ type ProjectOptions struct {
 	Compatibility bool
 	Progress      string
 	Offline       bool
+	Dependencies  types.DependencyOption
 }
 
 // ProjectFunc does stuff within a types.Project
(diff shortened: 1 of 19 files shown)


cmd/compose/compose.go
```diff
@@ -311,7 +312,9 @@ func RootCommand(dockerCli command.Cli, backend api.Service) *cobra.Command { //
 		"commandConn.CloseRead:",
 	))
 
-	opts := ProjectOptions{}
+	opts := ProjectOptions{
+		Dependencies: types.IncludeDependencies,
```
ndeloof, line 316: I wonder `IgnoreDependencies` should be the default value, and we opt-in for dependencies in up/start/stop/down commands

PR #11051, opened 2023-09-29
warn user remote resource is disabled

ndeloof:
**What I did**
as user forgot (?) to enable git/oci remote resource support (still experimental) compose tries to resolve a local path from URL. This is obvious something is wrong, but the way to get this fixed isn't.

this PR let the remote resource loader detect it _should_ load a remote resource, but report an explicit error and way to fix it.

### DIFF
--- a/pkg/remote/oci.go
+++ b/pkg/remote/oci.go
@@ -114,51 +114,51 @@ func (g ociRemoteLoader) Load(ctx context.Context, path string) (string, error) 
 			return "", err
 		}
 
-		s, err2 := g.pullComposeFiles(ctx, local, composeFile, manifest, ref, resolver)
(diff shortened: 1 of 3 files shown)


glours (approved):
LGTM when the lint issue will be fixed

pkg/remote/oci.go
```diff
@@ -114,51 +114,51 @@ func (g ociRemoteLoader) Load(ctx context.Context, path string) (string, error)
 			return "", err
 		}
 
-		s, err2 := g.pullComposeFiles(ctx, local, composeFile, manifest, ref, resolver)
```
ndeloof: could also rename variable `err`

glours:
Sorry I see your comment too late 😬

PR #11061, opened 2023-10-02
add support of COMPOSE_ENV_FILES env variable to pass a list of env files

glours:
**What I did**
Add a new `COMPOSE_ENV_FILES` to allow users to define once the env files they want to associate to their Compose configuration
This PR will need a follow up PR in the doc repo to add the new env var cc @aevesdocker  

**Related issue**
Internal discussion with Docker Captains

image

### DIFF
--- a/cmd/compose/compose.go
+++ b/cmd/compose/compose.go
@@ -60,6 +60,8 @@ const (
 	ComposeRemoveOrphans = "COMPOSE_REMOVE_ORPHANS"
 	// ComposeIgnoreOrphans ignore "orphaned" containers
 	ComposeIgnoreOrphans = "COMPOSE_IGNORE_ORPHANS"
+	// ComposeEnvFiles defines the env files to use if --env-file isn't used
+	ComposeEnvFiles = "COMPOSE_ENV_FILES"
 )
 
 // Command defines a compose CLI command as a func with args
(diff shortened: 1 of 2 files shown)


ndeloof (approved):

nickjj:
Hi,

I'm not sure if this handles the use case described in the Docker Captains channel.

```yml
# docker-compose.yml
services:
  debian:
    image: "debian:bookworm-slim"
    restart: "${DOCKER_RESTART_POLICY:-unless-stopped}"
```

```sh
# .env
COMPOSE_ENV_FILES=.env,.env.local
DOCKER_RESTART_POLICY=yes
```

```sh
# .env.local
DOCKER_RESTART_POLICY=no
```

If you run `docker compose config` the value of restart is `yes` but it would be expected that it's `no` through the logic of:

- `docker compose` was called without explicitly defining any `--env-file` flags which means it will look for `.env` by default
- The new `COMPOSE_ENV_FILES` env var will be used which says to interpolate the `.env` file and then `.env.local`
  - This is the part that is failing the use case expectation
- `no` should be set

If you run `COMPOSE_ENV_FILES=.env,.env.local docker compose config` then everything works as expected but this requires altering your shell or creating a custom alias to run docker compose with that set. The original use case was to allow for running `docker compose` commands without modification.

Was this an accidental oversight or am I missing something?

glours:
Hey @nickjj 
I totally forgot to add the support via the `.env` files 🤦 
Can you open a new issue to track it, please?

nickjj:
No problem, here it is: https://github.com/docker/compose/issues/11122

PR #11076, opened 2023-10-10
remove cucumber tests as we haven't added new ones for a while

glours:
**What I did**
Remove cucumber tests as we only added 2 new ones since we introduced them, it costs us CI time and we should give a huge engineering effort to provide a good coverage using them

**Related issue**
N/A

image


ndeloof (approved):

laurazard (approved):
😢 👋

Issue #11091, opened 2023-10-12
Inconsistent/undefined behavior between default merges and those with profiles


ericdbarry:
### Description

When passing in -f <filename> a merge happens and resources sharing a name are brought together into a single resource.  It's not entirely clear at what level profiles affect this, and the behavior between the up and down commands is inconsistent, so I am not even sure I know what I should be logging.

Expected:
Running compose with two files using -f and no profile flag, docker would first filter out the services with a profile and only merge the default services, or error saying that it could not find the service to launch. (depending on when the merge happened)

Actual:
Compose launches the services that are assigned to a profile when the user has not passed in the profile.

### Steps To Reproduce

**With these config files:**
`compose.yml`

```
version: "3.7"

services:
  mongo:
    image: "mongo:4.0.23-xenial"
    container_name: "mongo"
    ports:
      - "27017:27017"
```

`compose-enterprise.yml`
```
version: "3.7"

services:
  mongo:
    image: "mongo:4.4-focal"
    container_name: "mongo"
    profiles:
      - "cluster"
    ports:
      - "27017:27017"
    restart: always
    command: mongos -f /etc/mongos.conf
    depends_on:
      mongo-shard-1:
        condition: "service_started"
      mongo-shard-2:
        condition: "service_started"
      mongo-shard-3:
        condition: "service_started"
  mongo-shard-1:
    image: "mongo:4.4-focal"
    container_name: "mongo-shard-1"
    profiles:
      - "cluster"
    ports:
      - "27111:27018"
    restart: always
    command: -f /etc/mongod.conf
  mongo-shard-2:
    image: "mongo:4.4-focal"
    container_name: "mongo-shard-2"
    profiles:
      - "cluster"
    ports:
      - "27112:27018"
    restart: always
    command: -f /etc/mongod.conf
  mongo-shard-3:
    image: "mongo:4.4-focal"
    container_name: "mongo-shard-3"
    profiles:
      - "cluster"
    ports:
      - "27113:27018"
    restart: always
    command: -f /etc/mongod.conf
```
**Running these commands:** 
When running just the compose.yml file, I get the expected behavior from both up/down
```
dockertest % docker-compose -f docker/compose.yml up -d mongo 
[+] Building 0.0s (0/0)                                                                                                                                                                                                 docker:desktop-linux
[+] Running 2/2
 ✔ Network docker_default  Created                                                                                                                                                                                                      0.0s 
[...]

PR #11108, opened 2023-10-18
enable profile when down is ran with explicit service names

ndeloof:
**What I did**
enable services explicitly selected by user running `down` command

**Related issue**
closes https://github.com/docker/compose/issues/11091


glours (approved):
LGTM

Issue #11107, opened 2023-10-18
[BUG] Endless start of containers


madcrinkle:
### Description

There are two containers, one in task mode, executed and exited, the second depends on the first
If for some reason you delete it after executing the first container, then after stopping the containers the launch will freeze and hang without errors.
In version 1.29.2, such containers failed and the launch continued

### Steps To Reproduce

Create docker-compose.yml

```
version: '3.7'

services:
  busybox:
    image: busybox:latest
    command: sleep 600
    depends_on:
      busybox-init:
        condition: service_completed_successfully
  
  busybox-init:
    image: busybox:latest
    command: sh -c "exit 0"
```

Up containers

```
% docker-compose -f docker-compose-depends.yml up -d
[+] Building 0.0s (0/0)                                                                                                                                                                    docker:default
[+] Running 3/3
 ✔ Network test_default           Created                                                                                                                                                            0.1s 
 ✔ Container test-busybox-init-1  Exited                                                                                                                                                             0.0s 
 ✔ Container test-busybox-1       Started
```

Remove container

```
% docker rm -f test-busybox-init-1
test-busybox-init-1
```

Stop containers

```
% docker-compose -f docker-compose-depends.yml stop 
[+] Stopping 1/1
 ✔ Container test-busybox-1  Stopped
```

Start containers, and nothing happens

```
% docker-compose -f docker-compose-depends.yml start
[+] Running 0/0
 ⠋ Container test-busybox-1  Starting                                                                                                                                                                0.0s 
canceled
```

### Compose Version

```Text
% docker compose version
Docker Compose version 2.22.0
% docker-compose version
Docker Compose version 2.22.0
```

### Docker Environment

```Text
% docker info
Client:
 Version:    24.0.6
 Context:    default
 Debug Mode: false
 Plugins:
  compose: Docker Compose (Docker Inc.)
    Version:  2.22.0
    Path:     /usr/lib/docker/cli-plugins/docker-compose

Server:
 Containers: 1
  Running: 0
  Paused: 0
  Stopped: 1
 Images: 756
 Server Version: 24.0.6
 Storage Driver: overlay2
  Backing Filesystem: xfs
  Supports d_type: true
  Using metacopy: true
  Native Overlay Diff: false
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
[...]

PR #11110, opened 2023-10-19
fail start if depependency is missing

ndeloof:
**What I did**
as we wait for dependencies, if no container is found fail or warn user. This should never happen using `up` as convergence created the dependencies in order.

**Related issue**
closes https://github.com/docker/compose/issues/11107


glours (approved):
LGTM

PR #11113, opened 2023-10-20
add docs upstream validation workflow

dvdksn:
Signed-off-by: David Karlsson <35727626+dvdksn@users.noreply.github.com>

**What I did**

added a gha workflow that verifies that the reference docs don't contain broken links etc

**Related issue**

🐥

### DIFF
--- a/.github/workflows/docs-upstream.yml
+++ b/.github/workflows/docs-upstream.yml
@@ -1,0 +1,42 @@
+# this workflow runs the remote validate bake target from docker/docs
+# to check if yaml reference docs used in this repo are valid
+name: docs-upstream
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+on:
+  push:
+    branches:
+      - 'main'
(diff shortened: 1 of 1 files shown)


dvdksn:
hey @glours and @ndeloof, I added this new workflow that gives you some validation (check for broken links etc) in your CI. We have the same in Buildx and BuildKit, and since we vendor docs from this repo too I thought it would be good to add here

.github/workflows/docs-upstream.yml
```diff
@@ -0,0 +1,42 @@
...
+name: docs-upstream
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+on:
+  push:
+    branches:
+      - 'master'
```
glours, line 12: `master`? 🤔 
```suggestion
      - 'main'
```

dvdksn: Oops :)

glours (approved):
LGTM

PR #11120, opened 2023-10-23
align with OCI artifact best practices

ndeloof:
**What I did**
better align with OCI artifact best practices (https://oras.land/docs/concepts/artifact/):
- artifact blob layer is pushed with mediatype `application/vnd.oci.image.layer.v1.tar`
- `org.opencontainers.image.created` annotation is set

### DIFF
--- a/pkg/compose/publish.go
+++ b/pkg/compose/publish.go
@@ -20,6 +20,8 @@ import (
 	"context"
 	"encoding/json"
 	"os"
+	"path/filepath"
+	"time"
 
 	"github.com/compose-spec/compose-go/types"
 	"github.com/distribution/reference"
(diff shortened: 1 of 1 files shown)


glours (approved):

neersighted:
I don't think ORAS represents best practices; in particular there is no need to use a layer tar mediaType when that is not actually the format of the data.

ndeloof:
@neersighted right, would you recommend a better "source of truth" ? I hardly can find documentation on OCI artifacts, Google 1st result being archived https://github.com/opencontainers/artifacts 😅

neersighted:
https://github.com/opencontainers/image-spec/blob/main/manifest.md#guidelines-for-artifact-usage

neersighted:
Oh! And if I wasn't clear -- I think it's still worth carrying the annotation change; just not the change to the layer mediaType.

ndeloof:
sure, will just prepare another PR with a few other fixes

ndeloof:
moved to draft waiting for more fixes

Issue #11123, opened 2023-10-23
[BUG] CTRL+C No longer force kills processes


jaredlockhart:
### Description

Previously in compose, pressing CTRL+C would begin the graceful shutdown of all compose processes, and then a second CTRL+C would immediately force kill all processes.  On the latest version `Docker Compose version v2.22.0-desktop.2` no amount of CTRL+C seems to force kill the processes.

### Steps To Reproduce

1. Have a compose file that starts some long lived processes
2. docker compose up
3. Press CTRL+C
4. Graceful stop of all processes begins
5. Press CTRL+C again
6. Processes continue to attempt to gracefully stop

### Compose Version

```Text
Docker Compose version v2.22.0-desktop.2
```

### Docker Environment

```Text
Client:
 Version:    24.0.6
 Context:    desktop-linux
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.11.2-desktop.5
    Path:     /Users/jaredlockhart/.docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.22.0-desktop.2
    Path:     /Users/jaredlockhart/.docker/cli-plugins/docker-compose
  dev: Docker Dev Environments (Docker Inc.)
    Version:  v0.1.0
    Path:     /Users/jaredlockhart/.docker/cli-plugins/docker-dev
  extension: Manages Docker extensions (Docker Inc.)
    Version:  v0.2.20
    Path:     /Users/jaredlockhart/.docker/cli-plugins/docker-extension
  init: Creates Docker-related starter files for your project (Docker Inc.)
    Version:  v0.1.0-beta.8
    Path:     /Users/jaredlockhart/.docker/cli-plugins/docker-init
  sbom: View the packaged-based Software Bill Of Materials (SBOM) for an image (Anchore Inc.)
    Version:  0.6.0
    Path:     /Users/jaredlockhart/.docker/cli-plugins/docker-sbom
  scan: Docker Scan (Docker Inc.)
    Version:  v0.26.0
    Path:     /Users/jaredlockhart/.docker/cli-plugins/docker-scan
  scout: Docker Scout (Docker Inc.)
    Version:  v1.0.7
    Path:     /Users/jaredlockhart/.docker/cli-plugins/docker-scout

Server:
 Containers: 14
  Running: 0
  Paused: 0
  Stopped: 14
 Images: 19
 Server Version: 24.0.6
[...]

PR #11124, opened 2023-10-23
fix support for 2nd Ctrl+C to kill containers

ndeloof:
**What I did**
Don't close signal channel before we return from `up` func otherwise we don't manage second Ctrl+C and don't kill container which didn't reacted to graceful stop signal.

**Related issue**
closes https://github.com/docker/compose/issues/11123


glours (approved):
LGTM.
Tests ok I can't reproduce the error and didn't notice side effect with `watch` command 👍

PR #11127, opened 2023-10-24
fix SIGTERM support to stop/kill stack

ndeloof:
**What I did**
this fix signal management:

A dedicated goroutine is set to handle signal and trigger stop/kill commands (first signal will stop, the next one kills). 
`doneCh` is set to stop this goroutine when no sigterm signal is caught. It's not actually used as a channel, just as a way for `select` to react as channel is closed and stop the goroutine
`isTerminated` boolean let us know stack was terminated and `doneCh` doesn't need to be closed. It is set once `stop` succeed, so that it in the meantime we can capture subsequent SIGTERM and kill the stack.

Note: this works well when ran by `~/.docker/cli-plugins/docker-compose up` or `com.docker.cli compose up` but when used with cloud integrations which forwards signals (😓) we receive duplicated SIGTERM signals and go runtime kills compose plugin after 3rd one

**Related issue**
closes https://github.com/docker/compose/issues/11123

### DIFF
--- a/cmd/compose/compose.go
+++ b/cmd/compose/compose.go
@@ -83,6 +83,8 @@ func AdaptCmd(fn CobraCommand) func(cmd *cobra.Command, args []string) error {
 			go func() {
 				<-s
 				cancel()
+				signal.Stop(s)
+				close(s)
 			}()
 		}
 		err := fn(ctx, cmd, args)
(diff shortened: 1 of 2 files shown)


glours (approved):
LGTM

Issue #11126, opened 2023-10-24
[BUG] v2.23.0 breaks if can't create cache for any reason


mritalian:
### Description

Starting v2.23.0 something broke and doing basic `docker-compose pull` fails when run from systemd unit, I assume this is due to lack of env var as the error seen is:

`initializing git cache: could not create any of the following paths: /.cache/docker-compose`

I can't imagine writing in root is desired but again is probably side effect of $HOME not being set.  We should no longer assume / is writeable which more and more is not, considering OS with ro root fs like coreos.  We can also not assume $HOME is set, some stronger mechanism should be used.

Rolling back to v2.22.x works fine

### Steps To Reproduce

1. create a oneshot systemd unit that runs docker-compose pull
2. start the unit
3. journalctl will show the error

Noticed on fedora coreos

### Compose Version

```Text
Docker Compose version v2.22.0

no plugin
```

### Docker Environment

```Text
Client:
 Context:    default
 Debug Mode: false

Server:
 Containers: 7
  Running: 7
  Paused: 0
  Stopped: 0
 Images: 7
 Server Version: 20.10.23
 Storage Driver: overlay2
  Backing Filesystem: xfs
  Supports d_type: true
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: journald
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 io.containerd.runtime.v1.linux runc
 Default Runtime: runc
 Init Binary: /usr/libexec/docker/docker-init
 containerd version: 
 runc version: 
 init version: 
 Security Options:
  seccomp
   Profile: default
  selinux
  cgroupns
 Kernel Version: 6.4.15-200.fc38.x86_64
 Operating System: Fedora CoreOS 38.20230902.3.0
 OSType: linux
 Architecture: x86_64
 CPUs: 4
 Total Memory: 1.907GiB
 Name: bastion01-test-5grcs-com
 ID: J7JL:ULEZ:XK6Z:APFN:4HTL:P3H3:CISE:2KTK:IVDM:P7O3:5R46:YMZY
 Docker Root Dir: /var/lib/docker
 Debug Mode: false
 Username: ericbsummit
 Registry: https://index.docker.io/v1/
 Labels:
[...]

PR #11130, opened 2023-10-25
re-implement cache folder detection

ndeloof:
**What I did**
replace `xdg.CacheFile` with our own implementation (basically, a copy/paste) so that we detect `HOME` is not set and don't assume we can use `/` as a replacement. Unclear to me if this is legitimate from a Posix point of view, but we should offer users a better error message for diagnostic than "can't write /.cache/docker-compose"

**Related issue**
closes https://github.com/docker/compose/issues/11126
closes https://github.com/docker/compose/issues/11137


thaJeztah:
> basically, a copy/paste

If this implementation is copying the original code, we probably must include the copyright information from the original;
https://github.com/adrg/xdg/blob/693ef18888dfbd619bc164da09317f0df8496a1a/LICENSE#L12-L13

glours (approved):
LGTM

PR #11146, opened 2023-10-27
Fix --hash regression

### LINKED ISSUE #11145: [BUG] regression in v2.23.0: `docker compose config --hash=*` prints empty result
### Description

Starting with v2.23.0, `docker compose config --hash=*` prints empty result.

This is a regression in commit a345515f91c7bc6a7822fc8668cda65734584890 (#11010)

### Steps To Reproduce

Create a `docker-compose.yaml` like this:
```yaml
services:
  hello1:
    image: alpine:3.13
```

v2.22.0 works as expected:
```console
$ docker compose config --hash=*             
hello1 5bc8bc96d82232fd5546825b6e64c03b2074467edf89dc48be706283b8e4b4d6
```

v2.23.0 shows no result:
```
$ docker compose config --hash=*
(no output)
```

### Compose Version

```Text
Docker Compose version v2.23.0
```

### Docker Environment

```Text
(No dependency on the daemon)
```

### Anything else?

_No response_

mattwalo32:
**What I did**
When I previously changed the hashing behavior in #11010, I neglected to consider the case where someone provides `--hash=*` to `docker configure`. My code assumed that a service name would always be passed in explicitly. This change uses all project service names in the absence of any provided name .

**Related issue**
Fixes #11145, thanks for pointing this out!

### DIFF
--- a/cmd/compose/config.go
+++ b/cmd/compose/config.go
@@ -190,6 +190,10 @@ func runHash(ctx context.Context, dockerCli command.Cli, opts configOptions) err
 		return err
 	}
 
+	if len(services) == 0 {
+		services = project.ServiceNames()
+	}
+
 	sorted := services
 	sort.Slice(sorted, func(i, j int) bool {
 		return sorted[i] < sorted[j]


glours (approved):
LTGM, tests ok with `--hash=*` or `--hash=serviceA,serviceB`

ndeloof (approved):

Issue #11176, opened 2023-11-09
[BUG] The --quiet option to the docker compose ls command seems to ignore the --filter option


JMCanning78:
### Description

When running a `docker compose ls` operation, I can use either the `--filter` option or the `--quiet` option but not both together. When `--quiet` is present, the output is not filtered.

Here's a sample transcript to illustrate:

```
$ docker compose ls
NAME                STATUS              CONFIG FILES
repo                running(1)          /home/example/repo/compose.yaml
test-repo           running(2)          /home/example/test-repo/compose.yaml
$ docker compose ls --filter "name=^report$"
NAME                STATUS              CONFIG FILES
$ docker compose ls --quiet
repo
test-repo
$ docker compose ls --filter "name=^report$" --quiet
repo
test-repo
```

I would expect the last command to have no output, since none of the composer projects match the regular expression for the name.

### Compose Version

```Text
$ docker compose version
Docker Compose version v2.21.0
$ docker-compose version
Command 'docker-compose' not found, but can be installed with:
sudo snap install docker          # version 20.10.24, or
sudo apt  install docker-compose  # version 1.29.2-1
See 'snap info docker' for additional versions.
```

### Docker Environment

```Text
$ docker info
Client: Docker Engine - Community
 Version:    24.0.6
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.11.2
    Path:     /usr/libexec/docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.21.0
    Path:     /usr/libexec/docker/cli-plugins/docker-compose

Server:
 Containers: 5
  Running: 3
  Paused: 0
  Stopped: 2
 Images: 9
 Server Version: 24.0.6
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: false
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 61f9fd88f79f081d64d6fa3bb1a0dc71ec870523
[...]

PR #11177, opened 2023-11-10
render quiet after filtering applied

ndeloof:
**What I did**
render in quiet mode after applying filter

**Related issue**
closes https://github.com/docker/compose/issues/11176


glours (approved):
Sounds good to me

PR #11181, opened 2023-11-11
Implement `docker compose attach`

g0t4:
**What I did**

I implemented `docker compose attach` by hooking into the underlying `docker container attach` using it's `NewAttachCommand` with some overrides to support lookup of service name/index => container ID. 

Usage:
```bash
docker compose attach foo
# STDIN/STDOUT/ERR connected
```

In my case, I had a development environment that uses `dotnet watch` inside of a container (started with compose). And then I needed to send `Ctrl+R` to reload my server (as needed). And thus the need to attach to the container's STDIN.

### DIFF
--- a/cmd/compose/attach.go
+++ b/cmd/compose/attach.go
@@ -31,11 +31,8 @@ type attachOpts struct {
 	index   int
 
 	detachKeys string
-
-	// todo docker container attach also has:
-	// --no-stdin  // whole point of attach is for STDIN (can already attach to STDOUT and STDERR via docker compose up)
-	// --sig-proxy
-
+	noStdin    bool
+	proxy      bool
 }
 
 func attachCommand(p *ProjectOptions, dockerCli command.Cli, backend api.Service) *cobra.Command {
@@ -61,6 +58,8 @@ func attachCommand(p *ProjectOptions, dockerCli command.Cli, backend api.Service
 	runCmd.Flags().IntVar(&opts.index, "index", 0, "index of the container if service has multiple replicas.")
 	runCmd.Flags().StringVarP(&opts.detachKeys, "detach-keys", "", "", "Override the key sequence for detaching from a container.")
 
+	runCmd.Flags().BoolVar(&opts.noStdin, "no-stdin", false, "Do not attach STDIN")
+	runCmd.Flags().BoolVar(&opts.proxy, "sig-proxy", true, "Proxy all received signals to the process")
 	return runCmd
 }
 
@@ -74,6 +73,8 @@ func runAttach(ctx context.Context, dockerCli command.Cli, backend api.Service, 
 		Service:    opts.service,
 		Index:      opts.index,
 		DetachKeys: opts.detachKeys,
+		NoStdin:    opts.noStdin,
+		Proxy:      opts.proxy,
 	}
 	return backend.Attach(ctx, projectName, attachOpts)
 }
--- a/cmd/compose/attach.go
+++ b/cmd/compose/attach.go
@@ -1,0 +1,79 @@
+/*
+   Copyright 2020 Docker Compose CLI authors
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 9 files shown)


g0t4:
@ndeloof

glours:
Hello @g0t4 
Isn't the `--attach` from the `up` command enough for what you want to do? I mean did you try with `docker compose up --attach foo` ? 
By the way there is already code to find and attach to a container see [here](https://github.com/docker/compose/blob/main/pkg/compose/attach.go#L36) for example, it should be better to re-use this existing code
Same for your Lookup function, everything is already available with [the `getSpecifiedContainer` function](https://github.com/docker/compose/blob/b0af2deb2b3a1ac8dbf47d668b9fd9e1db5b51ed/pkg/compose/containers.go#L74)

ndeloof:
`--attach` require you know in advance which service you want to attach to, while `docker compose attach` allows to run "diagnostic" on an existing stack. This is basically for parity with `docker attach``
by the way, my PR to export `RunAttach` on docker/cli has been merged

cmd/compose/attach.go
```diff
@@ -0,0 +1,82 @@
...
+		projectName, err := p.toProjectName(dockerCli)
+		if err != nil {
+			return err
+		}
+
+		if len(args) > 0 {
+			opts.Service = args[0]
+		}
+
+		containerID, err := backend.LookupContainer(ctx, projectName, opts)
```
ndeloof, line 71: please mimic the `exec` commands, and have api.Backend to declare a new `Attach` command, then rely on docker/cli  for implementation if relevant.
This for sure requires to duplicate flags definition, but those don't change much and this is a general UX-homogeneity issue we should address in a global way

g0t4:
> Isn't the `--attach` from the `up` command enough for what you want to do? I mean did you try with `docker compose up --attach foo` ? 

`docker compose up --attach` doesn't attach STDIN, in a perfect world I would prefer that it did but given the nature of multiple containers, I went the route of mirroring the underlying `docker container attach` command for a single service's container.

cmd/compose/attach.go
g0t4: Ok so you would prefer that I not reuse `NewAttachCommand` and instead hook into `RunAttach` like `exec` does with `RunExec`?

g0t4: And btw, if I hook into `RunAttach` I can easily add back the ability to `docker compose attach` (w/o service name) and have it grab the first container in the project. I removed that when using `NewAttachCommand` due to some friction with overriding validation.

ndeloof: Yes indeed, I'd prefer not to rely on `NewAttachCommand` here (sharing command and flags definition with docker CLI is a larger effort we need to investigate)

> I can easily add back the ability to docker compose attach (w/o service name) and have it grab the first container in the project

please don't, containers in a project are not ordered, so this won't be deterministic. Better rely on the same selection mechanism used by `compose exec` to enforce an homogeneous UX

cmd/compose/attach.go
```diff
@@ -0,0 +1,82 @@
...
+		if err != nil {
+			return err
+		}
+
+		// TODO? --dry-run? would it be helpful to print the matching container name/ID and not attach?
```
glours, line 76: `--dry-run` apply to commands which could mutate the state of your stack

g0t4:
@ndeloof ok I finally got a chance to rework things to use RunAttach, I had to update to `docker/cli` and `docker/docker` v1.25 beta 1

Let me know what you think.

ndeloof:
nice! please amend your commit to sign them off, this is required for any contribution by legals.
I guess we could have a dedicated PR to bump docker/cli to v1.25 beta.1

g0t4:
@ndeloof rebased

ndeloof:
run `make mocks` to regenerate mocks according to API updates

g0t4:
FYI I rebased and ran `make mocks` (no changes after rebasing, previously I did find changes)... anything else need to be done?

g0t4:
nvm, I just noticed this is closed, apologies!

g0t4:
Ok I believe I blindly clicked close, reopening...

g0t4:
Ok, fixed

g0t4:
signoff done too

ndeloof:
Please rebase and run `make docs` to update command line reference docs, then we can merge

g0t4:
done!

glours (approved):
LGTM

glours:
We should add an e2e test as a follow up PR

g0t4:
@glours is there a similar e2e test in mind that you think would work well with `attach`?

glours:
@g0t4 we should release a new version today

Issue #11179, opened 2023-11-10
[BUG] template_driver being ignored in config


jedahan:
### Description

I am trying to use secrets in a config template to work around [zitadel not supporting file-based secrets management](https://github.com/zitadel/zitadel/issues/6860).

It seems the config is correctly identified and mounted, but not executing the golang template wrapper. I would expect the output to be `Hello, 007` instead of `Hello, {{ secret "codename" }}`.

Otherwise, I would expect some sort of error message and early exit if `template_driver` is incorrect.

```
$ docker compose up
[+] Building 0.0s (0/0)                                                                   docker:orbstack
[+] Running 1/0
 ✔ Container compose-scratch-app-1  Recreated                                                        0.0s 
Attaching to compose-scratch-app-1
compose-scratch-app-1  | hello, {{ secret "codename" }}
compose-scratch-app-1 exited with code 0
```

**compose.yaml**
```yaml
version: "3.8"

secrets:
  codename:
    file: codename

configs:
  greeting:
    file: greeting
    template_driver: golang

services:
  app:
    image: alpine
    secrets: [ codename ]
    configs: [ greeting ]
    command: [ "sh", "-c", "/greeting" ]
```

**greeting**
```bash
echo Hello, {{ secret "codename" }}
```

**codename**
```text
007
```

### Steps To Reproduce

Use the above files, then run `docker compose up`, and observe that the configs are not being run through the specified template_driver.

### Compose Version

```bash
$ docker compose version
Docker Compose version v2.22.0

$ docker-compose version
Docker Compose version 2.23.0
```

### Docker Environment

```yaml
Client:
 Version:    24.0.6
 Context:    orbstack
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.11.1
    Path:     /Users/micro/.docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.22.0
    Path:     /Users/micro/.docker/cli-plugins/docker-compose

Server:
 Containers: 4
  Running: 0
  Paused: 0
  Stopped: 4
 Images: 8
 Server Version: 24.0.6
 Storage Driver: overlay2
  Backing Filesystem: btrfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 2
 Plugins:
  Volume: local
[...]

PR #11185, opened 2023-11-13
reject compose file using `secrets|configs.driver or template_driver`

ndeloof:
**What I did**
reject compose file using `secrets|configs.driver or template_driver`
those are implemented by docker swarm configs/secrets engine-drivers, which have no equivalent component in Docker Compose, and we have no "extensibility" mechanism that would allow to run such a "driver" client-side (assuming we want to)

**Related issue**
see https://github.com/docker/compose/issues/11179

### DIFF
--- a/pkg/compose/create.go
+++ b/pkg/compose/create.go
@@ -20,6 +20,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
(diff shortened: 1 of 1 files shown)


glours (approved):
LGTM

PR #11187, opened 2023-11-13
bump compose-go to v1.20.1

ndeloof:
**What I did**
bump compose-go to v1.20.1
implemented support for build.ulimits
implemented support for config.content (inlined config)


glours (approved):
Sounds good to me 👍 
Just need to fix the lint issue

Issue #11188, opened 2023-11-14
Strip project prefix from docker-compose up output in non-compatibility mode


gyorokpeter:
### Description

Without compatibility mode, the prefix looks like `project-container-1`. With compatibility mode (`export COMPOSE_COMPATIBILITY=true`), it's `container_1`, lacking the prefix. In my case the project name has to be very long, and printing it on every line makes it hard to read and increases the size of the log file by a significant amount.  Considering that compatibility mode should only be a temporary measure, I would like to be able to strip the project prefix from the log lines so the prefix is just `container-1` without the project name even when not using compatibility mode. The closest option seems to be `--no-log-prefix` but that strips the entire prefix, not just the project name, making it hard to tell which container is outputting which line.

ndeloof:
compose v2 indeed uses plain container name for prefix, we could just use service + replica index (similar to compose v1) to offer a shorter prefix.

ndeloof:
Which version are you using ?
Can't reproduce with latest release:
```
$ cat compose.yaml 
services:
  test:
    image: alpine
    command: ping localhost
$  docker compose up
[+] Running 1/0
 ✔ Container bidule_test_1  Created                                        0.0s 
Attaching to test_1
test_1  | PING localhost (127.0.0.1): 56 data bytes
test_1  | 64 bytes from 127.0.0.1: seq=0 ttl=64 time=0.058 ms
test_1  | 64 bytes from 127.0.0.1: seq=1 ttl=64 time=0.205 ms
```

gyorokpeter:
```
$ docker compose version
Docker Compose version v2.17.3
$ docker compose -p bidule up
[+] Running 2/2
 ✔ Network bidule_default   Created                                                                              0.0s
 ✔ Container bidule-test-1  Created                                                                              0.1s
Attaching to bidule-test-1
bidule-test-1  | PING localhost (127.0.0.1): 56 data bytes
bidule-test-1  | 64 bytes from 127.0.0.1: seq=0 ttl=64 time=0.675 ms
bidule-test-1  | 64 bytes from 127.0.0.1: seq=1 ttl=64 time=0.089 ms
^CGracefully stopping... (press Ctrl+C again to force)
```

PR #11190, opened 2023-11-15
Strip project prefix from docker-compose up output

ndeloof:
**What I did**
fix `getContainerNameWithoutProject` to correctly remove project name prefix from container name 

**Related issue**
fixes https://github.com/docker/compose/issues/11188


glours (approved):
LGTM

PR #11192, opened 2023-11-15
Make it `context aware` and add test skipping options through environment variables

ulyssessouza:
**What I did**
- Add skipping capability through an environment variable (`EXCLUDE_E2E_TESTS`).
- Make e2e context aware

**Related issue**

25259092_goat-on-the-grass-scaled-e1618776775566-1662x2048

### DIFF
--- a/pkg/e2e/framework.go
+++ b/pkg/e2e/framework.go
@@ -30,6 +30,7 @@ import (
 	"testing"
 	"time"
 
+	cp "github.com/otiai10/copy"
 	"github.com/stretchr/testify/require"
 	"gotest.tools/v3/assert"
 	"gotest.tools/v3/icmd"
@@ -95,6 +96,7 @@ func NewCLI(t testing.TB, opts ...CLIOption) *CLI {
 
 	configDir := t.TempDir()
 	initializePlugins(t, configDir)
+	initializeContextDir(t, configDir)
 
 	c := &CLI{
 		ConfigDir: configDir,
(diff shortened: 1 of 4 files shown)


pkg/e2e/framework.go
```diff
@@ -30,6 +30,7 @@ import (
 	"testing"
 	"time"
 
+	cp "github.com/otiai10/copy"
```
ndeloof, line 33: I wonder you could use https://github.com/moby/moby/blob/master/daemon/graphdriver/copy/copy.go#L124 as we already depend on moby and this would about an additional dependency

ulyssessouza: I tried but kept on getting a very strange error:
```
# github.com/docker/docker/pkg/chrootarchive
../../../../pkg/mod/github.com/docker/docker@v24.0.6+incompatible/pkg/chrootarchive/archive_unix.go:31:8: undefined: goInChroot
../../../../pkg/mod/github.com/docker/docker@v24.0.6+incompatible/pkg/chrootarchive/archive_unix.go:53:8: undefined: goInChroot
../../../../pkg/mod/github.com/docker/docker@v24.0.6+incompatible/pkg/chrootarchive/diff_unix.go:45:8: undefined: goInChroot
FAIL    github.com/docker/compose/v2/pkg/e2e [build failed]
```

thaJeztah: I'd avoid the graphdriver packages as they are really intended for the daemon (and don't have macos implementations)

Perhaps (but typing from my phone) the copy package in fsutils could work? Is this just to copy a directory? https://github.com/tonistiigi/fsutil/tree/master/copy

ndeloof: right, I just wondered, but that's ok
(I just miss the ability to define test-only dependencies in `go.mod`. I miss you `pom.xml` :)

ndeloof (approved):

PR #11206, opened 2023-11-20
introduce RuntimeVersion for code to check container runtime support

ndeloof:
**What I did**
introduce `RuntimeVersion` so compose components can check container runtime supports required API to apply compose.yaml config - better than expect engine to return an error

**Related issue**


glours (approved):
LGTM

PR #11207, opened 2023-11-20
adopt compose-go/v2

ndeloof:
**What I did**
adopt compose-go/v2

**Related issue**

image


milas (commented):
LGTM - just a couple small places where I think it's using a value where it needs a pointer

cmd/compose/options.go
```diff
@@ -19,32 +19,29 @@ package compose
...
 	"github.com/docker/compose/v2/pkg/utils"
 )
 
 func applyPlatforms(project *types.Project, buildForSinglePlatform bool) error {
 	defaultPlatform := project.Environment["DOCKER_DEFAULT_PLATFORM"]
-	for i := range project.Services {
-		// mutable reference so platform fields can be updated
-		service := &project.Services[i]
-
+	for name, service := range project.Services {
```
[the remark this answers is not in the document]
ndeloof: indeed, this is missing a final `project.Services[name] = service` to update the source after config has been mutated. Just worked by chance as this only makes changes to `build` sub-section.
The range semantic is the same for slices and maps, so I didn't checked

pkg/compose/compose.go
```diff
@@ -216,9 +219,11 @@ func (s *composeService) projectFromName(containers Containers, projectName stri
 				}
 				service.DependsOn[dependency] = types.ServiceDependency{Condition: condition, Restart: restart, Required: required}
 			}
+			set[name] = service
```
laurazard, line 222: this might not be practically problematic, but in general modifying a map during it's iteration is bad practice and can produce unpredictable results

laurazard: an alternative could be to do this inside the other `for` (where we build `set`), which might also be slightly more efficient?

pkg/compose/pull.go
```diff
@@ -63,11 +63,11 @@ func (s *composeService) pull(ctx context.Context, project *types.Project, opts
 		imagesBeingPulled = map[string]string{}
 	)
 
-	for i, service := range project.Services {
-		i, service := i, service
+	i := 0
```
laurazard, line 66: could remove this if we turned `pullErrors` from a `[]error` to a `map[string]error`, where the key is the service name?

laurazard (commented):
Overall LGTM, just a suggestion and a small callout

pkg/compose/compose.go
ndeloof: > in general modifying a map during it's iteration is bad practice and can produce unpredictable results

any reference ?
I was looking for official reference regarding this and only found https://github.com/golang/go/issues/9926

laurazard: Ah, that's fair. I probably got mixed up with adding key/value pairs, but modifying is safe (which makes sense). It was lots of code to comb through 😅

pkg/compose/pull.go
ndeloof: we can't, as `pullServiceImage` is ran in goroutines and then would required concurrent access to such a map. Here we use a pre-allocated slice with index-based writes.

anyway, I'd like to revisit in a future PR the pattern we use to access services, typically `InDependencyOrder` so that this one collects results from goroutines and return a `map[string]T`, so that we only have to manage go-concurrency in a single place

laurazard (approved):

milas (approved):

PR #11208, opened 2023-11-20
Add support for endpoint-specific MAC address

akerouanton:
Related to:

- https://github.com/compose-spec/compose-spec/pull/435
- https://github.com/moby/moby/pull/45905

Since API v1.44, Moby supports a per-endpoint MAC address and returns a warning when the container-wide mac_address field is set.

A corresponding field has been added to compose-spec and compose-go, so we need to leverage it to set the right API field.

This commit is backward-compatible with compose files that still set the container-wide mac_address field, and older API versions that don't know about the endpoint-specific MAC address field.

**What I did**

**Related issue**

### DIFF
--- a/pkg/compose/create.go
+++ b/pkg/compose/create.go
@@ -35,6 +35,7 @@ import (
 	"github.com/docker/docker/api/types/mount"
 	"github.com/docker/docker/api/types/network"
 	"github.com/docker/docker/api/types/strslice"
+	"github.com/docker/docker/api/types/versions"
 	volume_api "github.com/docker/docker/api/types/volume"
 	"github.com/docker/docker/errdefs"
 	"github.com/docker/go-connections/nat"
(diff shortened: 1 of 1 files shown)


pkg/compose/create.go
```diff
@@ -182,6 +183,56 @@ func (s *composeService) getCreateConfigs(ctx context.Context,
...
+		var withMacAddress []string
+		for nwName, nw := range service.Networks {
+			if nw != nil && nw.MacAddress != "" {
+				withMacAddress = append(withMacAddress, nwName)
+			}
+		}
+
+		// TODO(aker): maybe show a warning and reset all mac_address fields except for the main endpoint?
+		if len(withMacAddress) > 1 {
+			return createConfigs{}, fmt.Errorf("a MAC address is specified for multiple networks (%s), but the target Docker Engine is too old to support that", strings.Join(withMacAddress, ", "))
```
ndeloof, line 228: better let user know the exact version needed, see https://github.com/docker/compose/pull/10939/files#diff-afca7d69550488d990f7db87d3d979a6980f6029c8c86e6b24e26858e87aa70cR77

ndeloof (approved):
LGTM, please rebase so we can merge

ndeloof (approved):

glours (commented):
LGTM

PR #11211, opened 2023-11-21
avoir use of []types.ServiceConfig

ndeloof:
**What I did**

**Related issue**


glours (approved):
LGTM

PR #11213, opened 2023-11-21
in watch mode force pull policy to build for services with both build and develop attributes

glours:
This default behaviour will force a rebuild of the service images at watch process startup and be sure containers will be in sync with the local source code

**What I did**
For each service target by `watch` process (with a `develop` configuration) force the pull policy to build to be sure they will be rebuild at watch startup

**Related issue**

image

### DIFF
--- a/cmd/compose/watch.go
+++ b/cmd/compose/watch.go
@@ -20,6 +20,8 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/compose-spec/compose-go/types"
+
 	"github.com/docker/cli/cli/command"
 	"github.com/docker/compose/v2/internal/locker"
 	"github.com/docker/compose/v2/pkg/api"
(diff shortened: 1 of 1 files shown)


ndeloof:
LGTM
A forced-(re)build might take time even with cache (baking a cache-efficient Dockerfile is an art), so I wonder we should guard this by comparison between last modified time for watched files and image's creation date.

ndeloof (approved):

glours:
> LGTM A forced-(re)build might take time even with cache (baking a cache-efficient Dockerfile is an art), so I wonder we should guard this by comparison between last modified time for watched files and image's creation date.

We have the `--no-op` flag to avoid building images for this, you think this may not be enough @ndeloof?

milas (approved):

milas:
I agree with @ndeloof - that's something I've seen first-hand with Tilt, which has similar behavior. Some larger projects, even with efficient layer caching, could take a while just to run through and do context transfers etc.

That said, I do think that, even imperfect, this is a beneficial change, but we should look at using `mtime` (also imperfect) or other approaches to continue to refine the behavior

PR #11219, opened 2023-11-23
let contributor know we might close unanswered issues

ndeloof:
**What I did**
some issues never get updated by reporter as we request more details to reproduce
clarify such issues will be closed so we can focus on active ones


glours (approved):
LGTM

Issue #10138, opened 2023-01-04
Log that `docker-compose run` does not enable port-forwarding by default


fred84:
### Description

Hello docker team!

Yesterday my colleague and I spent significant time figuring out why `docker-compose run` does not map ports. Eventually, we found the `--service-ports` flag and that the [docs](https://docs.docker.com/engine/reference/commandline/compose_run/) mention this behavior. However, we are not the first (https://github.com/docker/compose/issues/1259, and likely not the last) to encounter this.

I think the following changes can improve overall user experience in similar cases: 

1. add the `verbose` flag to the `docker-compose run`
2. when a callee provides the "verbose" flag, emit a message "port mapping is turned off to avoid conflicts, please pass `--service-ports` to enable it".

ndeloof:
I'm worried having such a warning on every `compose run` command would be very annoying for much users who don't depend on this for their usage. I understand the frustration spending time before you understood this specific behavior of the `compose run` command, but as you noticed this is already well documented.

fred84:
@ndeloof I agree printing it everytime is overkill, but doing it together with the "verbose" flag should be ok

ndeloof:
sure, but - as you were investigating before you understand this issue - would you have guessed such a `--verbose` flag can be of any help? This assumes you noticed this flag, and if you did by typing `docker compose run --help` I guess `--service-ports` description would then ring a bell.

fred84:
@ndeloof, when something is not working the way I expect, one of the first things I try is to run it in the most chatty way (usually a `verbose` flag). Unfortunately, I didn't spot `service-ports` early on, so it may be worth adding `expose-service-ports` alias to have more similarity with the `expose` configuration block and the `expose` Dockerfile command.

ndeloof:
Right.

We are considering introduction for a "dry run" mode, as many user requested to review what changes `compose up` would apply before actually changing state. Maybe we could consider this as well, as part of a global "give more visibility" feature.

PR #11227, opened 2023-11-28
log we don't expose service ports when --verbose

ndeloof:
**What I did**
when ran in `--verbose` mode, log we don't expose service ports until requested by options
also make `-P` an alias to align with docker/cli

**Related issue**
closes https://github.com/docker/compose/issues/10138


glours (approved):
👌

PR #11230, opened 2023-11-28
send out a cancel event on SIGINT/SIGTERM signal for `compose up`

vyneer:
**What I did**
Re-added the `printer.Cancel()` call to the SIGINT/SIGTERM signal handling for `docker compose up`. It being missing caused the output on Ctrl-C to be slightly messed up, due to `%container% exited with code %code%` messages interrupting the otherwise neat status info.

<details>
  <summary>Before and after</summary>

comp
</details>

cool doggies


ndeloof (approved):

ndeloof:
nice!

PR #11232, opened 2023-11-29
Fix configs are mounted under /<id>

ndeloof:
**What I did**
Fix inlined and environment-defined configs to be mounted under `/<id>` until an explicit target is set
includes a test case to avoid such a regression to appear again

**Related issue**
see https://github.com/compose-spec/compose-spec/issues/346#issuecomment-1831861104


glours (approved):
👍

PR #11233, opened 2023-11-29
Update README.md to use standard `compose.yaml` file name

johnthagen:
**What I did**

Update the README to recommend using the standard Compose file name, `compose.yaml`

- https://github.com/compose-spec/compose-spec/blob/master/spec.md#compose-file
- https://docs.docker.com/compose/gettingstarted/#step-3-define-services-in-a-compose-file

image


ndeloof (approved):

ndeloof:
nice catch, thanks!

PR #11239, opened 2023-12-01
fix(publish): add OCI 1.0 fallback support for AWS ECR

milas:
**What I did**
Currently, we publish Compose artifacts following the OCI 1.1 specification, which is still in the RC state.

As a result, not all registries support it yet. Most notably, AWS ECR will reject certain OCI 1.1-compliant requests with `405 Method Not Supported` with cryptic `Invalid JSON` errors.

This adds initial support for Compose to generate either an OCI 1.0 or OCI 1.1 compatible manifest. Notably, the OCI 1.0 manifest will be missing the `application/vnd.docker.compose.project` artifact type, as that does not exist in that version of the spec. (Less importantly, it uses an empty `ImageConfig` instead of the newer `application/vnd.oci.empty.v1+json` media type for the config.)

Currently, this is not exposed as an option (via CLI flags or env vars). By default, OCI 1.1 is used unless the registry domain is `amazonaws.com`, which indicates an ECR registry, so Compose will instead use OCI 1.0.

Moving forward, we should decide how much we want to expose/ support different OCI versions and investigate if there's a more generic way to feature probe the registry to avoid maintaining a hardcoded list of domains, which is both tedious and insufficient.

**Related issue**
https://docker.atlassian.net/browse/ENV-344

family of meerkats


milas:
fyi I want to add an E2E test for this but need to get an ECR registry set up with permissions for GHA to use so will likely do that as a follow-up PR

ndeloof:
would it make sense we:
- offer a flag to select OCI spec version to be used
- if not set, first try to publish as OCI 1.1, then on failure warn user and have a second try as OCI 1.0

pkg/compose/publish.go
```diff
@@ -76,58 +70,25 @@ func (s *composeService) publish(ctx context.Context, project *types.Project, re
...
-		imageManifest, err = json.Marshal(v1.Manifest{
-			Versioned:    specs.Versioned{SchemaVersion: 2},
-			MediaType:    v1.MediaTypeImageManifest,
-			ArtifactType: "application/vnd.docker.compose.project",
-			Config:       configDescriptor,
-			Layers:       layers,
-			Annotations: map[string]string{
-				"org.opencontainers.image.created": time.Now().Format(time.RFC3339),
-			},
+		layerDescriptor := ocipush.DescriptorForComposeFile("image-diegests.yaml", yaml)
```
glours, line 73: Typo ?
```suggestion
		layerDescriptor := ocipush.DescriptorForComposeFile("image-digests.yaml", yaml)
```

glours (approved):
Sounds good to me, maybe a typo to fix first

docs/reference/compose_alpha_publish.md
```diff
@@ -5,10 +5,11 @@ Publish compose application
...
 ### Options
 
-| Name                      | Type | Default | Description                     |
-|:--------------------------|:-----|:--------|:--------------------------------|
-| `--dry-run`               |      |         | Execute command in dry run mode |
-| `--resolve-image-digests` |      |         | Pin image tags to digests.      |
+| Name                      | Type     | Default | Description                                                           |
+|:--------------------------|:---------|:--------|:----------------------------------------------------------------------|
+| `--dry-run`               |          |         | Execute command in dry run mode                                       |
+| `--oci-version`           | `string` |         | OCI Image specification version (automatically determined by default) |
```
ndeloof, line 11: OCI artifact specification

ndeloof (approved):
LGTM but helper message should refer to OCI artifact spec, not image spec

docs/reference/compose_alpha_publish.md
milas: OCI artifacts are dead, long live OCI images: https://github.com/opencontainers/image-spec/blob/main/manifest.md#guidelines-for-artifact-usage

😬

ndeloof: really? So this is an non-secret plan to make all this think damn confusing as an "image" can now either be an image or an arbitrary artifact 🥸

pkg/compose/publish.go
milas: I blame my cat - fixed in [3b9572b](https://github.com/docker/compose/pull/11239/commits/3b9572b63b0cfa9afe1109f13cc66c0eb14cc232)

milas:
fyi my latest push also includes a slight change: https://github.com/docker/compose/pull/11239/commits/4264d536b17b14d3442e9b323e64867247a58885

there's comments and references to the OCI spec there, but now using a custom media type for the config descriptor in OCI 1.0 style pushes so that the manifest is still identifiable as a Compose project - that's the suggested fall back in the spec

ndeloof (approved):

PR #11246, opened 2023-12-04
bump docker/cli to v1.25 beta.1

g0t4:
**What I did**

Updated `docker/cli` to get access to `RunAttach`. Also updated `docker/docker` and added dependencies + code changes to get build working.

I could use some guidance as I am not familiar with these dependencies! Thanks

**Related issue**

This PR is needed for `docker compose attach` PR: https://github.com/docker/compose/pull/11181

PR #11247, opened 2023-12-05
Bump docker cli v1.25 beta1

ndeloof:
**What I did**
same as https://github.com/docker/compose/pull/11246 with go.sum fixed

### DIFF
--- a/pkg/compose/create.go
+++ b/pkg/compose/create.go
@@ -449,15 +449,15 @@ func mapRestartPolicyCondition(condition string) container.RestartPolicyMode {
 	// map definitions of deploy.restart_policy to engine definitions
 	switch condition {
 	case "none", "no":
-		return types.RestartPolicyNo
+		return container.RestartPolicyDisabled
 	case "on-failure":
-		return types.RestartPolicyOnFailure
+		return container.RestartPolicyOnFailure
 	case "unless-stopped":
-		return types.RestartPolicyUnlessStopped
+		return container.RestartPolicyUnlessStopped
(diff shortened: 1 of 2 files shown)


pkg/compose/create.go
```diff
@@ -445,17 +445,19 @@ func getRestartPolicy(service types.ServiceConfig) container.RestartPolicy {
...
+	case "on-failure":
+		return types.RestartPolicyOnFailure
+	case "unless-stopped":
+		return types.RestartPolicyUnlessStopped
 	case "any", "always":
-		return "always"
+		return types.RestartPolicyAlways
 	default:
-		return condition
+		return types.RestartPolicyNo
```
milas, line 460: Previously we passed it through as-is, this is now silently coercing everything unknown to no restart. Should probably just do a type conversion and let Engine error out if it's unknown:
```suggestion
		return container.RestartPolicyMode(condition)
```

pkg/compose/logs_test.go
```diff
@@ -172,7 +172,7 @@ func TestComposeService_Logs_ServiceFiltering(t *testing.T) {
 
 type testLogConsumer struct {
 	mu sync.Mutex
-	// logs is keyed container; values are log lines
+	// logs is keyed containerType; values are log lines
```
milas, line 175: nit: over-aggressive find/replace

```suggestion
	// logs is keyed by container ID; values are log lines
```

milas (approved):
LGTM, just a couple of small comments

Issue #11245, opened 2023-12-04
[BUG] Bash completion doesn't work using profiles


taha-yassine:
### Description

When a profile is added to a service, the bash completion stops working for that service.

### Steps To Reproduce

1. Create `compose.yaml` and add:
    ```yaml
    services:
        hello_world:
            image: hello-world
            profiles:
                - dev
    ```
2. Try `docker composer up <tab>`

### Compose Version

```Text
Docker Compose version v2.21.0
```

### Docker Environment

```Text
Client: Docker Engine - Community
 Version:    24.0.7
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.11.2
    Path:     /usr/libexec/docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.21.0
    Path:     /usr/libexec/docker/cli-plugins/docker-compose

Server:
 Containers: 5
  Running: 3
  Paused: 0
  Stopped: 2
 Images: 4
 Server Version: 24.0.7
 Storage Driver: fuse-overlayfs
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: runc io.containerd.runc.v2
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: d8f198a4ed8892c764191ef7b3b06d8a2eeb5c7f
 runc version: v1.1.10-0-g18a0cb0
 init version: de40ad0
 Security Options:
  seccomp
   Profile: builtin
  cgroupns
 Kernel Version: 5.4.203-1-pve
 Operating System: Debian GNU/Linux 11 (bullseye)
 OSType: linux
 Architecture: x86_64
 CPUs: 2
 Total Memory: 4GiB
 Name: lm-ops
 ID: bf4eacc8-0a67-459c-845d-4da1133a68d6
 Docker Root Dir: /var/lib/docker
 Debug Mode: false
 Experimental: false
 Insecure Registries:
  127.0.0.0/8
 Live Restore Enabled: false

WARNING: No memory limit support
WARNING: No swap limit support
WARNING: No cpu cfs quota support
WARNING: No cpu cfs period support
WARNING: No cpu shares support
[...]

PR #11251, opened 2023-12-06
include disabled services for shell completion

ndeloof:
**What I did**
As invoking a service name explicitly does enable it without the need to pass a profile, those should be offered as candidates by shell completion

**Related issue**
close https://github.com/docker/compose/issues/11245


glours (approved):
👍

PR #11252, opened 2023-12-06
dependency graph visitor to receive ServiceConfig

ndeloof:
**What I did**
refactored dependency graph so that `InDependencyOrder` visitor get immediate access to the visited `ServiceConfig` (as in most cases, first action is to get this from project)

### DIFF
--- a/pkg/compose/dependencies.go
+++ b/pkg/compose/dependencies.go
@@ -38,6 +38,8 @@ const (
 	ServiceStarted
 )
 
+type visitorFn func(context.Context, string, types.ServiceConfig) error
+
 type graphTraversal struct {
 	mu      sync.Mutex
 	seen    map[string]struct{}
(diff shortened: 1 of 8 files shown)


glours (approved):
LGTM

pkg/compose/dependencies.go
```diff
@@ -180,11 +168,11 @@ func (t *graphTraversal) run(ctx context.Context, graph *Graph, eg *errgroup.Gro
...
 		eg.Go(func() error {
 			var err error
-			if _, ignore := t.ignored[node.Service]; !ignore {
-				err = t.visitorFn(ctx, node.Service)
+			if _, ignore := t.ignored[node.Key]; !ignore {
+				err = t.visitorFn(ctx, node.Key, *node.Service)
 			}
 			if err == nil {
-				graph.UpdateStatus(node.Key, t.targetServiceStatus)
+				graph.UpdateStatus(node.Key)
```
milas, line 175: nit: should we rename this something like `graph.SetVisited(node.Key)`?

milas (approved):
Nice refactor - particularly happy with the `ServiceStatus` stuff going away in the graph traversal, I never understood that code properly 🙈

ndeloof:
eventually considered we could get this moved to compose-go as a general-purpose utility to walk the service graph:
https://github.com/compose-spec/compose-go/pull/506

PR #11272, opened 2023-12-13
github.com/golang/mock is deprecated

ndeloof:
**What I did**
mockgen is deprecated and suggest to use its fork https://github.com/uber-go/mock


g0t4:
FYI, if I have the deprecated version installed then `make mocks` fails until I install the uber version

Would it help to modify the check to ensure the right version is installed?

i.e.

```bash
[ "$(mockgen --version)" = "v0.3.0" ] || go install go.uber.org/mock/mockgen@v0.3.0
```

thaJeztah:
We should probably consider adding a stage to the Dockerfile to generate them, so that it's not needed to install locally (similar t what we have for generating docs etc)

Issue #11275, opened 2023-12-14
Add option to build dependencies to `docker compose build`


richin13:
### Description

Similar to how `--build --always-recreate-deps` work when using `docker compose up`.

I understand I can run:

```
docker compose up --build --always-recreate-deps --no-start my-service
```

And have a very similar functionality, however the above still creates the service containers which for my use case is a bit of an inconvenience.

richin13:
I'll re-open this on https://github.com/compose-spec/compose-spec as it seems the most appropriate place for it

richin13:
> I'll re-open this on [compose-spec/compose-spec](https://github.com/compose-spec/compose-spec) as it seems the most appropriate place for it

I'm honestly not so sure about this anymore. I'll keep this open for visibility and if anyone knows if this should go some other place happy to forward it :)

PR #11290, opened 2023-12-20
`--with-dependencies` let `docker compose build` build dependencies transitively

ndeloof:
**What I did**
introduce `--with-dependencies` flag so `docker compose build` will also build dependencies transitively

**Related issue**
closes https://github.com/docker/compose/issues/11275


glours (approved):
LGTM

PR #11292, opened 2023-12-21
Up: teardown when command context is cancelled

laurazard:
**What I did**

Previously, if a long-lived plugin process (such as an execution of `compose up`) was running and then detached from a terminal, signalling the parent CLI process to exit would leave the plugin process behind.

To address this, changes were introduced on the CLI side (see: https://github.com/docker/cli/pull/4599) to enable the CLI to notify a running plugin process that it should exit. This makes it so that, when the parent CLI process is going to exit, the command context of the plugin command being executed is cancelled.

This PR takes advantage of that change by using that channel on an `up` to know when to teardown.

Also fixed the "cancellable context" detection in `cmd/compose/compose.go`, as the string suffix matching was missing contexts such as:
```go
context.Background.WithCancel.WithValue(type trace.traceContextKeyType, val <not Stringer>).WithValue(type api.DryRunKey, val <not Stringer>)
```

*Before*:

https://github.com/docker/compose/assets/70572044/e1ffa1f5-469b-410a-a281-61aa970867bd

*After*:

https://github.com/docker/compose/assets/70572044/9e27f4da-4f02-4aef-addb-9c72e605a98a

**Related issue**
https://github.com/docker/cli/pull/4599

Screenshot 2023-12-21 at 12 15 46

### DIFF
--- a/cmd/compose/compose.go
+++ b/cmd/compose/compose.go
@@ -42,6 +42,7 @@ import (
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 
+	"github.com/docker/cli/cli-plugins/plugin"
 	"github.com/docker/compose/v2/cmd/formatter"
 	"github.com/docker/compose/v2/pkg/api"
 	"github.com/docker/compose/v2/pkg/compose"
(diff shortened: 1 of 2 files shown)


laurazard:
Looks like `TestComposeCancel/metrics_on_cancel_Compose_build` is failing, I'll take a look.

cmd/compose/compose.go
```diff
@@ -277,7 +278,12 @@ const PluginName = "compose"
 
 // RunningAsStandalone detects when running as a standalone program
 func RunningAsStandalone() bool {
```
ndeloof, line 280: Should have been replaced by https://github.com/docker/cli/blob/ae41fd913eea86eb4be153782846c33935a1039e/cli-plugins/plugin/plugin.go#L223

laurazard: Agree! 😅

thaJeztah (approved):
LGTM

milas (approved):
LGTM, and you're my hero for fixing this!

I tested standalone, with Compose CLI trampoline, and with Docker CLI and all looks good. (Well, the standalone case does return exit code 130 but that's not a new regression; I'll open a quick PR for that later.)

PR #11298, opened 2023-12-22
remove ServiceProxy which was introduced for archived compose-cli

ndeloof:
`ServiceProxy` was introduced to allow "Cloud integration" to tweak the compose UX (see https://github.com/docker/compose-cli/blob/c156ce6da4c2b317174d42daf1b019efa87e9f92/cli/main.go#L277)
as https://github.com/docker/compose-cli has been archived we don't need to support this anymore


milas (approved):

glours (approved):
LGTM

PR #11325, opened 2024-01-08
Add source policies for build

cpuguy83:
Build{x,kit} support passing in source policies via an (experimental) env var.
This change adds those policies to the build request.

Equivalent code in `buildx build`
https://github.com/docker/buildx/blob/7c590ecb9a8005d7b9e8f4ef7217130d79310200/commands/build.go#L144-L147

`ReadSourcePolicy` code:
https://github.com/docker/buildx/blob/7c590ecb9a8005d7b9e8f4ef7217130d79310200/build/build.go#L1502-L1525

### DIFF
--- a/pkg/compose/build.go
+++ b/pkg/compose/build.go
@@ -410,6 +410,11 @@ func (s *composeService) toBuildOptions(project *types.Project, service types.Se
 		}}
 	}
 
+	sp, err := build.ReadSourcePolicy()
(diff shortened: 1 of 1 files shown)


joshspicer:
+1 My team is hoping to take advantage of buildkit source policies + compose

glours (approved):
LGTM

PR #11330, opened 2024-01-09
watch: remove requirements for tar binary and for sync target to be rw

ndeloof:
**What I did*
Replaced execution of `tar` within an exec to untar files by a PutContainerArchive API call (untar is managed by engine)
This removes requirement for container to have (gnu) `tar` command available + target path doesn't have to be RW by (unprivileged) container user

**Related issue**

### DIFF
--- a/pkg/compose/watch.go
+++ b/pkg/compose/watch.go
@@ -419,6 +419,12 @@ func (t tarDockerClient) Exec(ctx context.Context, containerID string, cmd []str
 	return nil
 }
 
+func (t tarDockerClient) Untar(ctx context.Context, id string, archive io.ReadCloser) error {
+	return t.s.apiClient().CopyToContainer(ctx, id, "/", archive, moby.CopyToContainerOptions{
+		CopyUIDGID: true,
+	})
+}
+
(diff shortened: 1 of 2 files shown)


pkg/compose/watch.go
```diff
@@ -419,6 +419,10 @@ func (t tarDockerClient) Exec(ctx context.Context, containerID string, cmd []str
 	return nil
 }
 
+func (t tarDockerClient) Untar(ctx context.Context, id string, archive io.ReadCloser) error {
+	return t.s.apiClient().CopyToContainer(ctx, id, "/", archive, moby.CopyToContainerOptions{})
```
milas, line 423: Should we set `CopyUIDGID: true`?

Engine uses that to make the files owned by the container user if set:
https://github.com/moby/moby/blob/4aa1bea04af697a4b6b6dbe5b6bef90747a3596d/daemon/archive_tarcopyoptions_unix.go

milas (approved):
LGTM, I did some quick tests on a simple project and didn't see any issues

glours (approved):
LGTM, we also need to update the doc repo to remove the requirement, but this can wait for the next release

cyberloof:
Yes, I had in mind we wait for this to be released with docker desktop before we update docs

Issue #8901, opened 2021-11-08
compose v2 not sending meaningful user agent string


thespad:
**Description**

When sending a pull request to a registry docker-compose 1.x will send a user agent which includes something like `UpstreamClient(docker-compose/1.29.2 docker-py/5.0.0 Linux/5.4.0-88-generic)` clearly identifying that the client is docker-compose and what its version is. Compose 2.x appears to send a generic user agent in the form `UpstreamClient(Docker-Client/unknown-version \(linux\))`.

**Steps to reproduce the issue:**
1. User docker compose v2 to pull from a registry you control
2. Observe the user agent sent with the request

**Describe the results you received:**
A user agent string of `docker/20.10.9 go/go1.16.8 git-commit/79ea9d3 kernel/5.4.0-77-generic os/linux arch/amd64 UpstreamClient(Docker-Client/unknown-version \(linux\))`

**Describe the results you expected:**
A user agent string with specific client and version information in keeping with the 1.x behaviour

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `docker compose version`:**

```
Docker Compose version v2.1.0
```

**Output of `docker info`:**

```
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  app: Docker App (Docker Inc., v0.9.1-beta3)
  buildx: Build with BuildKit (Docker Inc., v0.6.3-docker)
  compose: Docker Compose (Docker Inc., v2.1.0)
  scan: Docker Scan (Docker Inc., v0.8.0)

Server:
 Containers: 56
  Running: 56
  Paused: 0
  Stopped: 0
 Images: 59
 Server Version: 20.10.9
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 1
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 io.containerd.runtime.v1.linux runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 5b46e404f6b9f661a205e28d59c982d3634148f8
 runc version: v1.0.2-0-g52b36a2
 init version: de40ad0
 Security Options:
  apparmor
  seccomp
   Profile: default
[...]

Issue #4574, opened 2017-03-05
unexpected behavior with interactive tty and command history


jeffrom:
When I `docker-compose exec service` bash, psql, and other interactive programs, i've noticed that `control-P` and `control-N` do not have the same behavior as the up and down arrows when navigating history. Up and down work as expected, but ctrl-P doesn't show the previous command. When that happens, if i use a left or right arrow to move the cursor, the command will show up. But it's pretty annoying. Am I the only one with this problem? :)

nodakai:
Take a look at moby/moby#1446 & moby/moby#3405

PR #11333, opened 2024-01-10
feat(cli): report more useful User-Agent on engine API requests

milas:
**What I did**
When using the Moby/Docker Engine API client, we do not have a useful user agent value being reported. Ideally, in the future, the Docker CLI will set this appropriately for plugins when it initializes the client.

For now, manually set it, which is a bit hacky because it requires some casting & manually invoking an option function that's technically meant for initialization. In practice, this is pretty safe - the cast is checked defensively and we ignore any errors (which shouldn't be possible anyway).

two rats snuggling

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -24,6 +24,7 @@ import (
 	"github.com/docker/cli/cli-plugins/plugin"
 	"github.com/docker/cli/cli/command"
 	"github.com/docker/compose/v2/cmd/cmdtrace"
+	"github.com/docker/docker/client"
 	"github.com/spf13/cobra"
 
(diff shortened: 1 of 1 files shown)


glours (approved):
LGTM

ndeloof (approved):
LGTM, but I wonder we are in a hurry to get this and have to introduce such a hack. Can't we just wait for #4574 to be approved ?

glours:
This will allow us to remove hacks on the Docker Desktop side, and we don't know when the next release of Docker CLI with the Milas' PR will be done

thaJeztah:
Looks like this fixes;

- https://github.com/docker/compose/issues/8901

Issue #11326, opened 2024-01-09
[BUG] The initiated health check process is missing


yangjinheng:
### Description

* new version
> Docker Compose version v2.23.3
Created -> Started -> Healthy                                      -> 0.0s

* old version
> Docker Compose version v2.20.2
Created -> Starting -> Waiting -> Healthy                   -> 0.1s  ~ 24.8s

### Steps To Reproduce

# services with healthcheck

```yaml
services:
  mysql:
    image: mysql:8.2.0
    environment:
      - MYSQL_DATABASE=testing
      - MYSQL_ROOT_PASSWORD=password
    command: ["--character-set-server=utf8mb4", "--collation-server=utf8mb4_bin"]
    restart: always
    healthcheck:
      test: ["CMD-SHELL", 'mysqladmin ping -h 127.0.0.1 --password="password" --silent']
      interval: 10s
      timeout: 1m30s
      retries: 3
      start_period: 10s
    networks: ["quicknet"]
    volumes:
      - mysql:/var/lib/mysql
```

### Compose Version

```Text
Docker Compose version v2.23.3
```

### Docker Environment

```Text
Client:
 Version:    24.0.6
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.10.4
    Path:     /usr/libexec/docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.23.3
    Path:     /usr/libexec/docker/cli-plugins/docker-compose

Server:
 Containers: 9
  Running: 8
  Paused: 0
  Stopped: 1
 Images: 102
 Server Version: 24.0.6
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: runc io.containerd.runc.v2
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 7880925980b188f4c97b462f709d0db8e8962aff
 runc version: v1.1.9-0-gccaecfc
 init version: de40ad0
 Security Options:
  apparmor
  seccomp
[...]

PR #11357, opened 2024-01-16
Stop the resource timer after last expected event

ndeloof:
**What I did**
`Done` event status doesn't mean the resource reached the expected status, it typically is first `Created` then `Started` and eventually `Healthy`. We need the TUI to be aware more events are expected and not stop the timer.

**Related issue**
fixes https://github.com/docker/compose/issues/11326


jhrotko (approved):
LGTM

milas (approved):
lgtm

K2ouMais:
This seems to be broken again, after updating to Docker Desktop 4.29.0

PR #11361, opened 2024-01-17
signals/utils: always handle received signals

laurazard:
The changes in https://github.com/docker/compose/commit/dcbf005fe4fa2b360b8dd013d6dedc4796c37602 fixed the "cancellable context" detection, and made it so that Compose would conditionally set up signal handling when the context was already not cancellable/when the plugin was running through the CLI, as we'd introduced a mechanism into the CLI to signal plugins to exit through a socket instead of handling signals themselves.

This had some (not noticed at the time) issues when running through the CLI as, due to sharing a process group id with the parent CLI process, when a user CTRL-Cs the CLI will notify the plugin via the socket but the plugin process itself will also be signalled if attached to the TTY. This impacted some Compose commands that don't set up signal handling - so not `compose up`, but other commands would immediately quit instead of getting some "graceful" cancelled output.

We initially attempted to address this "double notification" issue in the CLI by executing plugins under a new pgid so that they wouldn't be signalled, but that posed an issue with Buildx reading from the TTY, (see: https://github.com/moby/moby/issues/47073) so we reverted the process group id changes and ended at a temporary solution in https://github.com/docker/cli/pull/4792 where the CLI will only notify plugins via the socket when they are not already going to be signalled (when attached to a TTY).

Due to this, plugins should always set up some signal handling, which this commit implements.

**What I did**

Make Compose always handle signals – you can verify the broken/fixed behaviour by building Compose from this PR and the CLI from master, and then trying:

`docker compose pull` and hitting CTRL-C before it finishes

Pre-changes:
```shell
 ../../cli/build/docker compose pull
[+] Pulling 0/1
 ⠹ a Pulling                                                                                          0.3s
^C%
```

after:
```shell
 ../../cli/build/docker compose pull
[+] Pulling 1/1
 ✘ a Error                                                                                            0.2s
canceled
```

**Related issue**

### DIFF
--- a/cmd/compose/compose.go
+++ b/cmd/compose/compose.go
@@ -42,7 +42,6 @@ import (
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 
-	"github.com/docker/cli/cli-plugins/plugin"
 	"github.com/docker/compose/v2/cmd/formatter"
 	"github.com/docker/compose/v2/pkg/api"
 	"github.com/docker/compose/v2/pkg/compose"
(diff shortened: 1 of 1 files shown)


cmd/compose/compose.go
```diff
@@ -74,20 +73,17 @@ type CobraCommand func(context.Context, *cobra.Command, []string) error
...
-				<-s
-				cancel()
-				signal.Stop(s)
-				close(s)
-			}()
-		}
+		ctx, cancel := context.WithCancel(cmd.Context())
+
+		s := make(chan os.Signal, 1)
+		signal.Notify(s, syscall.SIGTERM, syscall.SIGINT)
```
milas, line 79: fwiw I think this could be further simplified w/ `signal.NotifyContext`: https://pkg.go.dev/os/signal#NotifyContext

milas (approved):
lgtm!

milas (approved):
lgtm!

glours (approved):

laurazard:
Also sorry about all of these changes/reverts/fixes, CLI signal handling stuff is hard and sockets are not equal on different OSs (did you know macOS does not support abstract sockets?) 🥲😭

laurazard:
Ahh don't merge this, it looks like it's breaking something in `run`

PR #11382, opened 2024-01-19
chore(e2e): fix flaky test & standalone behavior

milas:
**What I did**
This test was flaking in CI even though it was succeeding; I think there's a race with us killing and getting the output. Regardless, refactored so that it watches the process, which should make the test more reliable and faster.

Additionally, I noticed that the standalone config was incorrect, so I (reset) the bool to the correct value to ensure we test things in plugin mode for E2E.

**Related issue**
n/a

fox in the snow


glours (approved):
LGTM

laurazard (approved):
❤️

Issue #11392, opened 2024-01-23
[BUG] .env file at the same level of $COMPOSE_FILE not loaded


eathtespagheti:
### Description

Up to version `2.23.3` docker compose would end up loading both the `.env` file of the CWD, and the `.env` file at the same level of the `$COMPOSE_FILE` in use

Since version `2.24.0` only the first `.env` il loaded,  but the other `.env` file at the same level of the `$COMPOSE_FILE` is not sourced, causing those env variables to be empty strings

#### Example on 2.24.0 running `docker compose build`
```
WARN[0000] The "IMAGE_NAME" variable is not set. Defaulting to a blank string. 
WARN[0000] The "IMAGE_TAG" variable is not set. Defaulting to a blank string. 
[+] Building 0.0s (0/0)                                                                                                             docker:default
invalid tag "/:": invalid reference format
```

### Steps To Reproduce

I have a setup with the following tree

```
.
├── .env
└── stack
    ├── development
    │   ├── docker-compose.yaml
    │   ├── .env
    │   └── node.env
    ├── Dockerfile
    └── local
        └── node.env
```

The main `.env` file is automatically loaded if I run `docker compose` in that directory, and this `.env` file set-up the `$COMPOSE_FILE` variable to specify which docker compose file to use

#### `.env`
``` sh
#!/usr/bin/env sh

export COMPOSE_FILE="${COMPOSE_FILE:-stack/development/docker-compose.yaml}"
```

while `stack/development/.env` contains some ENV variables used inside the `stack/development/docker-compose.yaml` file

#### `stack/development/.env`
``` sh
#!/usr/bin/env sh

# Docker
export COMPOSE_PROJECT_NAME=projectname

# Image
export IMAGE_NAME="${IMAGE_NAME:-backend}"
export IMAGE_TAG="${IMAGE_TAG:-latest}"
```

#### `stack/development/docker-compose.yaml`
``` yaml
version: "3.8"

services:

  backend:
    image: $IMAGE_NAME:$IMAGE_TAG
    build:
      context: ../../
      dockerfile: stack/Dockerfile
    restart: unless-stopped
```

### Compose Version

```Text
Docker Compose version 2.24.2
```

### Docker Environment

```Text
Client:
 Version:    24.0.7
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  0.12.1
    Path:     /usr/lib/docker/cli-plugins/docker-buildx
[...]

PR #11405, opened 2024-01-25
Fix load .env from project directory when project file is set by COMPOSE_FILE

ndeloof:
**What I did**
get location of the compose file (maybe set by env) before loading default .env file. Doing so, if COMPOSE_FILE is set to a distinct directory, we will load sibling `.env` file as user would expect

**Related issue**
fixes https://github.com/docker/compose/issues/11392

### DIFF
--- a/cmd/compose/compose.go
+++ b/cmd/compose/compose.go
@@ -261,10 +261,10 @@ func (o *ProjectOptions) toProjectOptions(po ...cli.ProjectOptionsFn) (*cli.Proj
 		append(po,
 			cli.WithWorkingDirectory(o.ProjectDir),
 			cli.WithOsEnv,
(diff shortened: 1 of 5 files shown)


jhrotko (approved):

milas (approved):

MikaelElkiaer:
Hey @ndeloof I just dropped a comment on a related issue: https://github.com/docker/compose/issues/11531#issuecomment-1978064657

It seems there is another regression related to this.
At first this fix did not help me.
Even though I have a `.env` right next to my `docker-compose.yaml` - which sets `COMPOSE_FILE=subdir/docker-compose.yaml`, it still traverses parent directories to resolve a different `docker-compose.yaml` before adhering to the `.env` in CWD.

PR #11429, opened 2024-01-30
Include all networks in ContainerCreate call if API >= 1.44

laurazard:
Previously, we included the container's primary network in the `ContainerCreate` API call, and then connected the created container to each extra network one-by-one, by calling `NetworkConnect`.

However, starting API version 1.44 (https://github.com/moby/moby/pull/45906), the `ContainerCreate` endpoint now takes multiple networks, allowing us to just include all the network configurations there and skip connecting the container to each extra network separately.

**What I did**

Change `composeService.createMobyContainer` to include all the network configurations for the container in the `ContainerCreate` call and skip connecting the container to networks individually, when API version >= 1.44.

Also added some tests to make sure we don't break this for <1.44.

(a good callout here is that currently, afaik, Compose isn't being tested against different Engine versions in CI, so it may be worth it to do that soon/wait on that before merging this PR)

Screenshot 2024-01-30 at 17 54 10


pkg/compose/convergence.go
```diff
@@ -600,19 +601,27 @@ func (s *composeService) createMobyContainer(ctx context.Context,
...
-		epSettings := createEndpointSettings(project, service, number, networkKey, cfgs.Links, opts.UseNetworkAliases)
-		if err := s.apiClient().NetworkConnect(ctx, mobyNetworkName, created.ID, epSettings); err != nil {
-			return created, err
+	versionInfo, err := s.apiClient().ServerVersion(ctx)
+	if err != nil {
+		return created, err
+	}
+	// Starting API version 1.44, the ContainerCreate API call takes multiple networks
+	// so we include all the configurations there and can skip the one-by-one calls here
+	if versions.LessThan(versionInfo.APIVersion, "1.44") {
```
laurazard, line 610: Oh, I was thinking about whether to put this version (1.44) in a const somewhere

ndeloof: You can use `RuntimeVersion` here so we only check engine version once

ndeloof (commented):
LGTM as a general feature but I'm not a big fan we cover such an improvement with unit tests relying on API mocks. As we start relying on 1.25 features we need to run real tests on multiple engine versions

laurazard:
Yup! I wrote in the PR test:
> (a good callout here is that currently, afaik, Compose isn't being tested against different Engine versions in CI, so it may be worth it to do that soon/wait on that before merging this PR)

Probably a good change regardless :')

milas (commented):
LGTM! Want to rebase to pick up #11459? and then we can mergeeee

laurazard:
Sorry, just saw this @milas! Rebased :)

laurazard:
cc @glours if you want to TAL

glours (approved):
Sounds good to me 👍

ndeloof (approved):

laurazard:
It's failing on a ddev test:
```
Failed to start 004: ddev-router failed to become ready; debug with 'docker logs ddev-router'; logOutput=, err=health check timed out: labels map[com.docker.compose.service:ddev-router] timed out without becoming healthy, status=
```
which I think might be flaky, but I don't have permissions to restart the test to check.
@ndeloof @glours can you TAL?

glours:
CI was stuck, I restarted the whole workflow

glours:
Ok we forgot to change the checks since we introduced the tests against 2 Docker engine versions @ndeloof 😅

Issue #11464, opened 2024-02-05
[BUG] `docker compose ps --status` behavior differs from documentation


adam-hintz-olo:
### Description

When looking for exited containers by status, containers do not show up unless I also pass the `--all` flag.

```
$ docker compose ps --status=exited | wc -l
1

$ docker compose ps --all --status=exited | wc -l
7
```

This is in conflict with what I see in the docs:

```
docker compose ps --status=exited
NAME            IMAGE     COMMAND           SERVICE    CREATED         STATUS          PORTS
example-bar-1   alpine    "/entrypoint.â¦"   bar        4 seconds ago   exited (0)
```

### Steps To Reproduce

Given the following `docker-compose.yml`:
```
version: "3"
services:
  ls:
    image: alpine:latest
    command: "ls"
```

Run the following commands and observe the output:

```
$ docker-compose up -d
## output truncated

$ docker compose ps
NAME      IMAGE     COMMAND   SERVICE   CREATED   STATUS    PORTS

$ docker compose ps --all
NAME                   IMAGE           COMMAND   SERVICE   CREATED          STATUS                      PORTS
docker-ps-repro-ls-1   ubuntu:latest   "ls"      ls        15 seconds ago   Exited (0) 13 seconds ago

$ docker compose ps --status=exited
NAME      IMAGE     COMMAND   SERVICE   CREATED   STATUS    PORTS

$ docker compose ps --status=exited --all
NAME                   IMAGE           COMMAND   SERVICE   CREATED          STATUS                      PORTS
docker-ps-repro-ls-1   ubuntu:latest   "ls"      ls        36 seconds ago   Exited (0) 34 seconds ago
```

### Compose Version

```Text
Docker Compose version v2.24.3-desktop.1
```

### Docker Environment

```Text
Client:
 Version:    25.0.2
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.12.1-desktop.4
    Path:     /usr/local/lib/docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.24.3-desktop.1
    Path:     /usr/local/lib/docker/cli-plugins/docker-compose
  debug: Get a shell into any image or container. (Docker Inc.)
    Version:  0.0.22
    Path:     /usr/local/lib/docker/cli-plugins/docker-debug
  dev: Docker Dev Environments (Docker Inc.)
    Version:  v0.1.0
[...]

PR #11476, opened 2024-02-08
list all containers as we filter by status

ndeloof:
**What I did**
As `compose ps` is set to filter by status, list all containers otherwise we will miss those with non-running status

**Related issue**
fixes https://github.com/docker/compose/issues/11464


1arp:
Since this forces the all option only in parse filter function, this will only work for `--filter status=exited` and not for `--status=exited`.

ndeloof:
indeed

PR #11478, opened 2024-02-08
Add JSON stream progress writer

### LINKED ISSUE #10872: docker-compose - dry-run - json output
### Description

It would be great if the docker-compose dry-run outputs had a flag to output a json blob at the end of their run with the information about proposed changes (if any).

This would allow for easier wrapping with tooling eg ansible/salt/python-on-whales 

https://github.com/gabrieldemarmiesse/python-on-whales/issues/448

felixfontein:
**What I did**
I added a JSON stream progress writer. This allows to write machine readable output to stderr, which makes it a lot easier to use Docker Compose from other tools, like Ansible.

**Related issue**
Fixes #10872.

A cute cat with rather sharp claws, watch out when petting her

### DIFF
--- a/cmd/compose/build.go
+++ b/cmd/compose/build.go
@@ -67,10 +67,9 @@ func (opts buildOptions) toAPIBuildOptions(services []string) (api.BuildOptions,
 		builderName = os.Getenv("BUILDX_BUILDER")
 	}
 
-	// The build backends do not have a "json" progress; use "plain" instead
 	uiMode := ui.Mode
 	if uiMode == ui.ModeJSON {
-		uiMode = ui.ModePlain
+		uiMode = "rawjson"
 	}
 	return api.BuildOptions{
 		Pull:     opts.pull,
--- a/cmd/compose/build.go
+++ b/cmd/compose/build.go
@@ -67,10 +67,15 @@ func (opts buildOptions) toAPIBuildOptions(services []string) (api.BuildOptions,
 		builderName = os.Getenv("BUILDX_BUILDER")
 	}
 
+	// The build backends do not have a "json" progress; use "plain" instead
+	uiMode := ui.Mode
+	if uiMode == ui.ModeJSON {
+		uiMode = ui.ModePlain
+	}
 	return api.BuildOptions{
 		Pull:     opts.pull,
 		Push:     opts.push,
-		Progress: ui.Mode,
+		Progress: uiMode,
 		Args:     types.NewMappingWithEquals(opts.args),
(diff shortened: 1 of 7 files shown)


felixfontein:
I still need to run `make docs` and commit the changes. Unfortunately it never comes remotely near to completion with the not so great internet connection I have on this train, so that has to wait until later...

felixfontein:
Docs are updated, CI passes (checked in https://github.com/felixfontein/compose/pull/1), but while testing this (now that I was actually able to build this) I noticed that Docker Compose apparently expects Docker Buildx and Docker CLI itself to also have a JSON progress writer. I'm getting `invalid progress mode json` errors when it tries to pull images, and building also doesn't seem to work.

felixfontein:
I eventually noticed that in case of errors, the error is always printed regularly and not via the progress API. I added code in d3b85f9e388787e06fe514957231e1b9696e051a that in case `--progress json` it also formats the error as JSON.

felixfontein:
This is now ready for review.

felixfontein:
It probably makes sense to set logging to JSON (`logrus.SetFormatter(&logrus.JSONFormatter{})`, see https://github.com/sirupsen/logrus/blob/master/README.md) to avoid random non-JSON logging messages to show up on stderr.

cmd/compose/build.go
```diff
@@ -67,10 +67,15 @@ func (opts buildOptions) toAPIBuildOptions(services []string) (api.BuildOptions,
 		builderName = os.Getenv("BUILDX_BUILDER")
 	}
 
+	// The build backends do not have a "json" progress; use "plain" instead
```
ndeloof, line 70: you can use `rawjson` so buildkit will render raw json solver events and you get a clean json stream for all events

felixfontein: I created https://github.com/moby/buildkit/issues/4769 since I think it would be better if the `rawjson` output would write one JSON object per line, instead of using multiple lines to show a more human-readable JSON record (that's not really that human readable anyway).

felixfontein:
@ndeloof can you take another look at this? Thanks!

ndeloof (approved):

ndeloof:
Can you please squash your commits and rebase, I'll enable auto-merge on this PR

felixfontein:
@ndeloof thanks a lot for reviewing and merging! Since you rebase-merged I assume I don't need to rebase/squash this anymore :)

ndeloof:
auto-merge applied before squash, never mind

PR #11492, opened 2024-02-12
Use listener for file metadata

jhrotko:
**What I did**

**Related issue**


internal/tracing/attributes.go
```diff
@@ -56,31 +60,37 @@ func (s SpanOptions) EventOptions() []trace.EventOption {
...
 		attribute.StringSlice("project.services.capabilities", capabilities),
 		attribute.StringSlice("project.services.capabilities.gpu", gpu),
 		attribute.StringSlice("project.services.capabilities.tpu", tpu),
 	}
+	countExtends := 0
+	metrics, ok := ctx.Value(Metrics{}).(Metrics)
+	if ok {
+		countExtends = metrics.CountExtends
+	}
+	attrs = append(attrs, attribute.Int("project.services.extends", countExtends))
```
ndeloof, line 92: shall we skip this attribute if we don't have a value? Using `0` might corrupt the metric

ndeloof (approved):

internal/tracing/attributes.go
glours: We could only append if the `Metrics{}` is present inside the context, but I'm sure the 0 default value we'll be used on the other side

glours (approved):

Issue #11493, opened 2024-02-13
[BUG] docker compose up --abort-on-container-exit not exiting when printing long log


LennartC:
### Description

When running `docker compose up --abort-on-container-exit`, we expect `docker compose` to abort as soon as the container exits.

However, when it has to print a lot of logs to stdout, `docker compose` will hang and wait forever.

### Steps To Reproduce

With the given compose file (the image doesn't matter):
```
services:
    test:
        image: debian:bookworm-slim
        entrypoint: /bin/bash
        command: -c 'cat /testfile'
        volumes:
            - ./large_text_file.txt:/testfile
```
With `large_text_file.txt` being a large file, e.g. 300.000 lines / 10 MB

Run:
```
docker compose up --abort-on-container-exit
```
The container will exit, but docker compose will not:
```
test-1 exited with code 0
Aborting on container exit...
[+] Stopping 1/0
 ✔ Container test-test-1  Stopped            
```

### Compose Version

```Text
Docker Compose version v2.24.5
```

### Docker Environment

```Text
Client: Docker Engine - Community
 Version:    25.0.3
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.12.1
    Path:     /usr/libexec/docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.24.5
    Path:     /usr/libexec/docker/cli-plugins/docker-compose

Server:
 Containers: 3
  Running: 0
  Paused: 0
  Stopped: 3
 Images: 249
 Server Version: 25.0.3
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 1
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: ae07eda36dd25f8a1b98dfbf587313b99c0190bb
 runc version: v1.1.12-0-g51d5e94
 init version: de40ad0
 Security Options:
  apparmor
  seccomp
   Profile: builtin
  userns
[...]

PR #11496, opened 2024-02-13
fix deadlock collecting large logs

ndeloof:
**What I did**
re-implement log consumer concurrency using a stop-signal channel to prevent more events write after printer stopped.

The original issue is that we receive the container `die` event while we still are processing a huge log, line by line. As a result the receiver already stopped processing events after `ContainerEventExit`, but the log writer still tries to write to channel,  and has acquired the lock so is blocked on channel write

inspired by https://go101.org/article/channel-closing.html

**Related issue**
fixes https://github.com/docker/compose/issues/11493#issuecomment-1940639535

### DIFF
--- a/pkg/compose/printer.go
+++ b/pkg/compose/printer.go
@@ -32,18 +32,19 @@ type logPrinter interface {
 }
 
 type printer struct {
-	sync.Mutex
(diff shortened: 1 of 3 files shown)


milas (approved):
👍🏻

PR #11506, opened 2024-02-15
Fix flaky scale test

jhrotko:
## Problem
In compose there is one flaky test related to scale command. We are removing the latest created container when scaling down https://github.com/docker/compose/blob/07bda5960e07cb6ebee5beaa0c2ba0e42b788b13/pkg/compose/convergence.go#L129-L131

And in the test, might not necessarily be the 3rd service (`front-3`)
First we scale up two `front` services, and we do not know for sure which of `front-2` or `front-3` is going to be created first.
https://github.com/docker/compose/blob/b1c06770d49da1dbcaacf77fb347ae0e6f577f37/pkg/e2e/scale_test.go#L46-L52
https://github.com/docker/compose/blob/b1c06770d49da1dbcaacf77fb347ae0e6f577f37/pkg/e2e/scale_test.go#L63-L67

Running locally the commands this is a possible state:
image

Running `scale_test.go` several times it's possible to get failures locally.
The proposed fix, keeps track of the containers creation order, and asserts if the last created container is the one removed, instead of expecting `scale-basic-tests-front-3` to be the one removed when scaling down.
Note:
in Go there is no lookahead forward/behind, so the regex rule had to be more hacky.

**Related issue**
https://github.com/docker/compose/pull/11473

**Work Item**
https://docker.atlassian.net/browse/COMP-435

image

### DIFF
--- a/pkg/e2e/scale_test.go
+++ b/pkg/e2e/scale_test.go
@@ -17,6 +17,7 @@ package e2e
 
 import (
 	"fmt"
+	"regexp"
 	"strings"
 	"testing"
 
@@ -36,6 +37,7 @@ func TestScaleBasicCases(t *testing.T) {
 	}
 	t.Cleanup(reset)
 	res := c.RunDockerComposeCmd(t, "--project-directory", "fixtures/scale", "up", "-d")
+	frontOrder := getContainerCreationOrder(t, res.Combined(), "front", "Created")
 	res.Assert(t, icmd.Success)
 
 	t.Log("scale up one service")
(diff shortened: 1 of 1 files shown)


ndeloof:
might relate to https://github.com/docker/compose/pull/11473

laurazard:
So I don't have full context, but I was looking at https://github.com/docker/compose/blob/d714a2ebee08a697a40efacc93272b89b8fcad49/pkg/e2e/scale_test.go#L102 and https://github.com/docker/compose/pull/11473 and I got the idea that when scaling down, under normal circumstances we shouldn't fall back to using created time and should be using the container number, so this would be deterministic.

Did I get this wrong, or did we do something wrong with https://github.com/docker/compose/pull/11473 and we're falling back to creation date?

ndeloof (approved):

milas (commented):
No concern with the actual code changes but same question as @laurazard - shouldn't the fix from #11473 mean this is not necessary anymore?

jhrotko:
@laurazard  @milas I retested with @ndeloof and it seems it is not flaky anymore

Issue #11089, opened 2023-10-12
[BUG] Watch: docker compose watch does not show service output / container logs


Axxxx0n:
### Description

`docker compose watch` does not show service output (logs), only syncing info. Running `docker compose up` with same docker-compose.yml config shows service outputs normally.

### Steps To Reproduce

1. run docker compose watch
2. see that there are no service output logs

### Compose Version

```Text
Docker Compose version v2.22.0-desktop.2
```

### Docker Environment

```Text
Client:
 Version:    24.0.6
 Context:    desktop-linux
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.11.2-desktop.5
    Path:     /Users/ax/.docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.22.0-desktop.2
    Path:     /Users/ax/.docker/cli-plugins/docker-compose
  dev: Docker Dev Environments (Docker Inc.)
    Version:  v0.1.0
    Path:     /Users/ax/.docker/cli-plugins/docker-dev
  extension: Manages Docker extensions (Docker Inc.)
    Version:  v0.2.20
    Path:     /Users/ax/.docker/cli-plugins/docker-extension
  init: Creates Docker-related starter files for your project (Docker Inc.)
    Version:  v0.1.0-beta.8
    Path:     /Users/ax/.docker/cli-plugins/docker-init
  sbom: View the packaged-based Software Bill Of Materials (SBOM) for an image (Anchore Inc.)
    Version:  0.6.0
    Path:     /Users/ax/.docker/cli-plugins/docker-sbom
  scan: Docker Scan (Docker Inc.)
    Version:  v0.26.0
    Path:     /Users/ax/.docker/cli-plugins/docker-scan
  scout: Docker Scout (Docker Inc.)
    Version:  v1.0.7
    Path:     /Users/ax/.docker/cli-plugins/docker-scout

Server:
 Containers: 0
  Running: 0
  Paused: 0
  Stopped: 0
 Images: 19
 Server Version: 24.0.6
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 2
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
[...]

PR #11511, opened 2024-02-16
show logs on watch

gabilaza:
**What I did**
Show logs for containers on `compose watch` with the `logs` flag, and include related flags for logs such as color, prefix, and timestamp.

**Related issue**
#11089 

**Why I did**
I believe that the 'compose watch' command is not complete without logs.

Cute-Baby-Owl-Bird-Watercolor-Clip-art-Graphics-82135327-1-1-580x435

### DIFF
--- a/cmd/compose/watch.go
+++ b/cmd/compose/watch.go
@@ -23,6 +23,7 @@ import (
 	"github.com/compose-spec/compose-go/v2/types"
 
 	"github.com/docker/cli/cli/command"
+	"github.com/docker/compose/v2/cmd/formatter"
 	"github.com/docker/compose/v2/internal/locker"
 	"github.com/docker/compose/v2/pkg/api"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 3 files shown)


ndeloof:
Thanks for looking into this.
I've been working on this same topic from another angle: adding `watch` feature inside `up` command: https://github.com/docker/compose/pull/11525
Could you please give it a try and let me know if this would cover your needs ?

gabilaza:
LGTM with the `--watch` option for `up`. I'm not sure if you noticed, but there are overlapping logs when recreating the container after rebuilding in your demo. I made a case where the container just builds something and exits with the build status and I noticed a problem: the following containers didn't show the logs. Perhaps the last issue will be fixed if the tests pass. I didn't look that deeply into the failed test to see if that's the problem.

ndeloof:
> there are overlapping logs when recreating the container after rebuilding in your demo

yes, the recreate progress UI interacts with the logs, need to fix this

PR #11549, opened 2024-02-27
Remove docker cli step in ci.yml

jhrotko:
**What I did**
Removed `Setup docker CLI` step because it was overriding docker version to `25.0.1` thus ignoring the matrix.engine versions previously installed. 

After looking at this [PR](https://github.com/docker/compose/pull/11535), I noticed something was odd, since this network issue has been fixed not long ago. 

image


glours (approved):
LGTM

PR #11555, opened 2024-02-27
Pull: include error message in warnings/errors

felixfontein:
**What I did**
When warnings or errors happen during pulling when running `docker compose up`, simply "<service_name> Warning" resp. "<service_name> Error" is printed without any information on what went wrong.

This PR changes this by adding the error message.

Before:
```
[+] Running 3/3
 ! dummy3 Warning                                                                                                                                                                        2.4s 
 ! dummy2 Warning                                                                                                                                                                        2.4s 
 ✘ dummy Error                                                                                                                                                                           2.4s 
```
After:
```
[+] Running 3/3
 ! dummy2 Warning Error response from daemon: pull access denied for dummy2, repository does not exist or may require 'docker login': denied: requested access to the resource is denied2.2s 
 ✘ dummy Error    Error response from daemon: pull access denied for dummy2, repository does not exist or may require 'docker login': denied: requested access to the resource is denied2.2s 
 ! dummy3 Warning Error response from daemon: pull access denied for dummy, repository does not exist or may require 'docker login': denied: requested access to the resource is denied2.2s 
```

**Related issue**
Ref: https://github.com/ansible-collections/community.docker/issues/807

A cute cat with rather sharp claws, watch out when petting her

### DIFF
--- a/pkg/compose/pull.go
+++ b/pkg/compose/pull.go
@@ -200,18 +200,20 @@ func (s *composeService) pullServiceImage(ctx context.Context, service types.Ser
 	// then the status should be warning instead of error
 	if err != nil && service.Build != nil {
 		w.Event(progress.Event{
-			ID:     service.Name,
-			Status: progress.Warning,
-			Text:   "Warning",
+			ID:         service.Name,
+			Status:     progress.Warning,
+			Text:       "Warning",
+			StatusText: err.Error(),
 		})
 		return "", WrapCategorisedComposeError(err, PullFailure)
 	}
 
 	if err != nil {
 		w.Event(progress.Event{
-			ID:     service.Name,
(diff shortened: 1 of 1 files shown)


ndeloof:
I wonder we could use `Unwrap` to get down to the interesting diagnostic here, and avoid generic daemon message, as such a long status text will probably get truncated

felixfontein:
@ndeloof I added some code for the unwrapping. This specific message needs to be unpacked at least twice, only unpacking once yields exactly the same output.

felixfontein:
@ndeloof is there a chance to get this actually merged?

felixfontein:
@ndeloof thanks a lot for reviewing and merging this!

Issue #7964, opened 2020-12-02
Compose config with --no-interpolate fails as it tries to validate environment variables as ports


padraic-padraic:
## Description of the issue

When trying combine compose and override files with `docker-compose config --no-interpolate`, I get an validation error in  the `ports` block for a service that uses environment variables.

## Context information (for bug reports)

**Output of `docker-compose version`**
```
docker-compose version 1.25.0, build unknown
```

**Output of `docker version`**
```
Docker version 19.03.8, build afacb8b7f0
```

**Output of `docker-compose config`**
(Make sure to add the relevant `-f` and other flags)
The issue relates to config!

## Steps to reproduce the issue

1. Small example file
    ```YAML
     #compose.yml
     version: '3'
     services:
        test:
          image: hello-world
          ports:
              - "${DUMMY_PORT}:${DUMMY_PORT}"
    ```
2. `docker-compose -f compose.yml config --no-interpolate`

### Observed result
`services.test.ports is invalid: Invalid port "${DUMMY_PORT}:${DUMMY_PORT}", should be [[remote_ip:]remote_port[-remote_port]:]port[/protocol]`
### Expected result
```
version: '3'
services:
  test:
    image: hello-world
    ports:
      - "${DUMMY_PORT}:${DUMMY_PORT}"
```
### Stacktrace / full error message

```
compose.config.config.find: Using configuration files: ./compose.yml
ERROR: compose.cli.main.main: The Compose file './compose.yml' is invalid because:
services.test.ports is invalid: Invalid port "${DUMMY_PORT}:${DUMMY_PORT}", should be [[remote_ip:]remote_port[-remote_port]:]port[/protocol]

```

## Additional information

OS version / distribution, `docker-compose` install method, etc.
Pop! OS 20.04, using system packages for docker / docker-compose.

o-sokolova:
How detailed should docker-compose validate variables in port field?
The whole pattern is `[[remote_ip:]remote_port[-remote_port]:]port[-port][/protocol]`
I think it would be redundant to validate case when variables are in every possible place. How you think?

emersonf:
Any bandwidth to make progress on this, e.g. PR feedback?

pseusys:
We really need this bug fixed! What's the progress?

ndeloof:
Compose v1 development is now paused as Compose v2 has been release. I don't expect some significant progress made for this issue regarding the python codebase.

pseusys:
[...]

PR #11556, opened 2024-02-28
restore support for `config --no-interpolate`

ndeloof:
**What I did**
implement `config` using compose-go `LoadModel` so we can actually skip interpolation

this PR uses a direct ref to compose-go HEAD waiting for a new release

**Related issue**
fixes https://github.com/docker/compose/issues/7964

### DIFF
--- a/cmd/compose/config.go
+++ b/cmd/compose/config.go
@@ -136,55 +136,19 @@ func configCommand(p *ProjectOptions, dockerCli command.Cli) *cobra.Command {
 }
 
 func runConfig(ctx context.Context, dockerCli command.Cli, opts configOptions, services []string) error {
-	var content []byte
 	model, err := opts.ToModel(ctx, dockerCli, services)
 	if err != nil {
 		return err
 	}
 
 	if opts.resolveImageDigests {
-		// create a pseudo-project so we can rely on WithImagesResolved to resolve images
-		p := &types.Project{
-			Services: types.Services{},
-		}
-		services := model["services"].(map[string]any)
-		for name, s := range services {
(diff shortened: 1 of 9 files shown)


cmd/compose/config.go
```diff
@@ -123,18 +135,56 @@ func configCommand(p *ProjectOptions, dockerCli command.Cli, backend api.Service
...
+		if err != nil {
+			return err
+		}
+
+		for name, s := range services {
+			service := s.(map[string]any)
+			config := p.Services[name]
+			if config.Image != "" {
+				service["image"] = config.Image
+			}
```
jhrotko, line 170: we could add this in the previous loop

cmd/compose/config.go
```diff
@@ -123,18 +135,56 @@ func configCommand(p *ProjectOptions, dockerCli command.Cli, backend api.Service
...
+		content, err = json.MarshalIndent(model, "", "  ")
+	case "yaml":
+		buf := bytes.NewBuffer([]byte{})
+		encoder := yaml.NewEncoder(buf)
+		encoder.SetIndent(2)
+		err = encoder.Encode(model)
+		content = buf.Bytes()
+	default:
+		return fmt.Errorf("unsupported format %q", opts.Format)
+	}
```
jhrotko, line 187: nit: this could be extracted to a private function

cmd/compose/config.go
ndeloof: no, because this is computed by `p.WithImagesResolved`: value isn't set to the expected value (image id with digest) before we call this func

jhrotko (approved):

PR #11592, opened 2024-03-08
chore(log): fix debug/verbose logging flags

milas:
**What I did**
Re-claim the `--debug` flag for debug logging.

Keep the `--verbose` flag hidden (for compat), but now as a no-op.

In the future, we should use `--verbose` for more detailed user-facing output.

**Related issue**
n/a

a lil inch worm


ndeloof:
Why not use the existing `docker --log-level=debug compose ...` feature ?

PR #11593, opened 2024-03-08
feat(desktop): add Docker Desktop detection and client skeleton

milas:
**What I did**
Use the system info Engine API to auto-discover the Desktop integration API endpoint (a local `AF_UNIX` socket or named pipe).

If present and responding to a `/ping`, the client is populated on the Compose service for use. Otherwise, it's left `nil`, and Compose will not try to use any functionality that relies on Docker Desktop (e.g. opening `docker-desktop://` deep links to the GUI).

This refactors some of the in-memory socket dialer code, so we can share it easier (time to bring back `docker/go-connections`? 🤔) and make it slightly more robust.

**Related issue**
JIRA: [COMP-457](https://docker.atlassian.net/browse/COMP-457)

badger

[COMP-457]: https://docker.atlassian.net/browse/COMP-457?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cmd/compose/compose.go
+++ b/cmd/compose/compose.go
@@ -37,6 +37,7 @@ import (
 	"github.com/docker/cli/cli-plugins/manager"
 	"github.com/docker/cli/cli/command"
 	"github.com/docker/compose/v2/cmd/formatter"
+	"github.com/docker/compose/v2/internal/desktop"
 	"github.com/docker/compose/v2/internal/tracing"
 	"github.com/docker/compose/v2/pkg/api"
 	"github.com/docker/compose/v2/pkg/compose"
@@ -365,11 +366,17 @@ func RootCommand(dockerCli command.Cli, backend api.Service) *cobra.Command { //
 			}
 		},
 		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
+			ctx := cmd.Context()
+
+			// (1) process env vars
 			err := setEnvWithDotEnv(&opts)
 			if err != nil {
 				return err
 			}
 			parent := cmd.Root()
+
+			// (2) call parent pre-run
+			// TODO(milas): this seems incorrect, remove or document
 			if parent != nil {
 				parentPrerun := parent.PersistentPreRunE
 				if parentPrerun != nil {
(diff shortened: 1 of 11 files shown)


jhrotko:
Could we also add an e2e test for this cases? unix, windows and darwin?

pkg/compose/compose.go
```diff
@@ -60,12 +62,29 @@ func NewComposeService(dockerCli command.Cli) api.Service {
...
+// In practice, this service has the same lifetime as the process, so everything
+// will get cleaned up at about the same time regardless even if not invoked.
+func (s *composeService) Close() error {
+	var errs []error
+	if s.dockerCli != nil {
+		errs = append(errs, s.dockerCli.Client().Close())
+	}
+	if s.desktopCli != nil {
+		errs = append(errs, s.desktopCli.Close())
+	}
```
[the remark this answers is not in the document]
ndeloof: not the same CLI here

cmd/compose/compose.go
```diff
@@ -466,13 +474,26 @@ func RootCommand(dockerCli command.Cli, backend api.Service) *cobra.Command { //
...
+
+			// (5) dry run detection
+			ctx, err = backend.DryRunMode(ctx, dryRun)
 			if err != nil {
 				return err
 			}
 			cmd.SetContext(ctx)
+
+			// (6) Desktop integration
+			if db, ok := backend.(desktop.IntegrationService); ok {
```
ndeloof, line 489: I wish we refactor this init sequence so we don't mutate `backend` but use setup builder according to options:
```
backend := compose.NewComposeService(dockerCli, WithDryRun(xx), WithDesktopIntegration(xx))
```

ndeloof (approved):

glours (approved):

cmd/compose/compose.go
milas: Agreed 😞 the init is overly complicated, especially because of the split between Docker/CLI/plugins, Compose itself, and Cobra 🐍 

This was already kind of hacky - I didn't want to add this to `api.Service` but the concrete `composeService` implementation is unexported

I'll see if I can refactor this in another PR to swap to options an consolidate some things, e.g. we have several lazy/memoized components but they all happen in different ways currently

ndeloof: > I didn't want to add this to api.Service

why not ?

milas:
> Could we also add an e2e test for this cases? unix, windows and darwin?

Yes, this is a great call out! This will require me to sort out a couple things first - need to get updated Docker Desktop builds on our CI runners & service accounts on Hub for them to log in as, so I made a ticket in JIRA.

I'll go ahead and get started on the test and `t.Skip()` it for the moment at least

jhrotko (approved):

Issue #11598, opened 2024-03-11
[BUG] config no longer builds environment attribute in 2.24.7


er0k:
### Description

`docker compose config` no longer populates the `environment` attribute with values from an env file.

[This change](https://github.com/compose-spec/compose-go/pull/582/files) broke a few of our deployment pipelines, since we prefer to ship only the compose file generated by the `config` command, rather than shipping the compose file and the env file.

Is there any way to get this old behavior back?

### Steps To Reproduce

given these two files:

`foo.yml`:
```
services:
  foo_serv:
    image: debian:latest
    env_file:
      - foo.env
```

and `foo.env`:
```
FOO=bar
BAR=baz
```

prior to v2.24.7 we were able to run `docker compose --file foo.yml config` and get the following output:
```
name: foo                                                                                                                                      
services:                                                                                                                                             
  foo_serv:                                                                                                                                           
    environment:                                                                                                                                      
      BAR: baz                                                                                                                                        
      FOO: bar                                                                                                                                        
    image: debian:latest                                                                                                                              
    networks:                                                                                                                                         
      default: null                                                                                                                                   
networks:                                                                                                                                             
  default:                                                                                                                                            
    name: foo_default 
```
notice the `environment` attribute is populated with our environment variables from the env file. This allowed us to ship a single compose file with all of our env vars defined.

However, since compose 2.24.7, we get this output:
```
name: foo                                                                                                                                      
networks:                                                                                                                                             
  default:                                                                                                                                            
    name: foo_default                                                                                                                          
services:                                                                                                                                             
  foo_serv:                                                                                                                                           
    env_file:                                                                                                                                         
      - path: /home/er0k/tmp/foo/foo.env                                                                                                       
        required: true                                                                                                                                
    image: debian:latest                                                                                                                              
    networks:                                                                                                                                         
      default: null    
```
it no longer populates the `environment` attribute at all, instead we have the `env_file` attribute that points to the env file.

### Compose Version

```Text
Docker Compose version v2.24.7
```

### Docker Environment

```Text
Client: Docker Engine - Community
 Version:    26.0.0-rc2
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.13.0
    Path:     /usr/libexec/docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.24.7
    Path:     /usr/libexec/docker/cli-plugins/docker-compose

Server:
[...]

PR #11604, opened 2024-03-12
only use ToModel when --no-interpolate is set

ndeloof:
**What I did**
https://github.com/docker/compose/pull/11556 has impact on `compose config` which disturbs our users. This PR allows to preserve the previous behavior as long as `--no-interpolate` isn't set, and only switch to raw model parsing when set

We can then work in parallel to get `ToModel` closer to `ToProject` until we can eventually remove duplicated code path

**Related issue**
https://github.com/docker/compose/issues/11598


jhrotko (approved):
LGTM

kamilchm:
Hey @ndeloof, great to see a fix. When are you planning to release it?

PR #11605, opened 2024-03-12
Add navigation menu with in compose up (attached)

jhrotko:
**What I did**
We want to open Docker Desktop through a shortcut in `docker compose up`.

To use this feature run
```
docker compose up
```
... yes it comes as default for attached mode :). it will render the navigation menu at the bottom.
You can disable this feature while using `up` run:
```
docker compose up --menu=[false|true]
```
OR set  environment var `COMPOSE_MENU=[false|true]`
OR use Docker Desktop experimental features and enable/disable - `Compose: TUI nav bar`

https://github.com/docker/compose/assets/22371565/ef8a6b37-1dde-4e27-a7af-4f55feca4c58

When an error occurs, it will be displayed at the top of the navigation menu for 10s.
We are also sending telemetry to understand if users' usage of this new feature. PR merged in pinata: https://github.com/docker/pinata/pull/27207

This PR introduces a keyboard listener that is currently listening to the following events:
- CTRL+C 
- ENTER
- w -> Start watch mode (alternatively you can also start watch mode with `docker compose up --watch`)
- v -> Open compose project in Docker Desktop through a deeplink `docker-desktop://dashboard/apps/NAME`
Screenshot 2024-03-12 at 13 04 57

**Related issue**

https://docker.atlassian.net/jira/software/c/projects/COMP/boards/576?selectedIssue=COMP-417

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -101,6 +101,7 @@ RUN --mount=type=bind,target=. \
 FROM build-base AS test
 ARG CGO_ENABLED=0
 ARG BUILD_TAGS
+ENV COMPOSE_MENU=FALSE
 RUN --mount=type=bind,target=. \
     --mount=type=cache,target=/root/.cache \
     --mount=type=cache,target=/go/pkg/mod \
--- a/cmd/compose/up.go
+++ b/cmd/compose/up.go
@@ -42,20 +42,22 @@ type composeOptions struct {
 
 type upOptions struct {
 	*composeOptions
-	Detach             bool
-	noStart            bool
-	noDeps             bool
-	cascadeStop        bool
-	exitCodeFrom       string
-	noColor            bool
-	noPrefix           bool
-	attachDependencies bool
-	attach             []string
-	noAttach           []string
-	timestamp          bool
-	wait               bool
-	waitTimeout        int
-	watch              bool
+	Detach                bool
+	noStart               bool
+	noDeps                bool
+	cascadeStop           bool
+	exitCodeFrom          string
+	noColor               bool
+	noPrefix              bool
+	attachDependencies    bool
+	attach                []string
+	noAttach              []string
+	timestamp             bool
+	wait                  bool
+	waitTimeout           int
+	watch                 bool
+	navigationMenu        bool
+	navigationMenuChanged bool
 }
 
 func (opts upOptions) apply(project *types.Project, services []string) (*types.Project, error) {
@@ -87,6 +89,7 @@ func upCommand(p *ProjectOptions, dockerCli command.Cli, backend api.Service, ex
 		PreRunE: AdaptCmd(func(ctx context.Context, cmd *cobra.Command, args []string) error {
 			create.pullChanged = cmd.Flags().Changed("pull")
 			create.timeChanged = cmd.Flags().Changed("timeout")
+			up.navigationMenuChanged = cmd.Flags().Changed("menu")
 			return validateFlags(&up, &create)
 		}),
 		RunE: p.WithServices(dockerCli, func(ctx context.Context, project *types.Project, services []string) error {
@@ -128,6 +131,8 @@ func upCommand(p *ProjectOptions, dockerCli command.Cli, backend api.Service, ex
 	flags.BoolVar(&up.wait, "wait", false, "Wait for services to be running|healthy. Implies detached mode.")
 	flags.IntVar(&up.waitTimeout, "wait-timeout", 0, "Maximum duration to wait for the project to be running|healthy")
 	flags.BoolVarP(&up.watch, "watch", "w", false, "Watch source code and rebuild/refresh containers when files are updated.")
+	flags.BoolVar(&up.navigationMenu, "menu", false, "Enable interactive shortcuts when running attached (Experimental). Incompatible with --detach.")
+	flags.MarkHidden("menu") //nolint:errcheck
 
 	return upCmd
 }
@@ -161,7 +166,7 @@ func runUp(
 	ctx context.Context,
 	dockerCli command.Cli,
 	backend api.Service,
-	_ *experimental.State,
+	experimentals *experimental.State,
 	createOptions createOptions,
 	upOptions upOptions,
 	buildOptions buildOptions,
@@ -180,6 +185,9 @@ func runUp(
 	project, err = upOptions.apply(project, services)
 	if err != nil {
 		return err
+	}
+	if !upOptions.navigationMenuChanged {
+		upOptions.navigationMenu = SetUnchangedOption(ComposeMenu, experimentals.NavBar())
 	}
 
 	var build *api.BuildOptions
@@ -253,15 +261,16 @@ func runUp(
 	return backend.Up(ctx, project, api.UpOptions{
 		Create: create,
 		Start: api.StartOptions{
-			Project:      project,
-			Attach:       consumer,
-			AttachTo:     attach,
-			ExitCodeFrom: upOptions.exitCodeFrom,
-			CascadeStop:  upOptions.cascadeStop,
-			Wait:         upOptions.wait,
-			WaitTimeout:  timeout,
-			Watch:        upOptions.watch,
-			Services:     services,
+			Project:        project,
+			Attach:         consumer,
+			AttachTo:       attach,
+			ExitCodeFrom:   upOptions.exitCodeFrom,
+			CascadeStop:    upOptions.cascadeStop,
+			Wait:           upOptions.wait,
+			WaitTimeout:    timeout,
+			Watch:          upOptions.watch,
+			Services:       services,
+			NavigationMenu: upOptions.navigationMenu,
 		},
 	})
 }
--- a/cmd/formatter/colors.go
+++ b/cmd/formatter/colors.go
@@ -36,6 +36,18 @@ var names = []string{
 }
 
 const (
+	BOLD      = "1"
+	FAINT     = "2"
+	ITALIC    = "3"
+	UNDERLINE = "4"
+)
+
+const (
+	RESET = "0"
+	CYAN  = "36"
+)
+
+const (
 	// Never use ANSI codes
 	Never = "never"
 
@@ -72,12 +84,17 @@ var monochrome = func(s string) string {
 	return s
 }
 
-func ansiColor(code, s string) string {
-	return fmt.Sprintf("%s%s%s", ansi(code), s, ansi("0"))
+func ansiColor(code, s string, formatOpts ...string) string {
+	return fmt.Sprintf("%s%s%s", ansiColorCode(code, formatOpts...), s, ansiColorCode("0"))
 }
 
-func ansi(code string) string {
-	return fmt.Sprintf("\033[%sm", code)
+// Everything about ansiColorCode color https://hyperskill.org/learn/step/18193
+func ansiColorCode(code string, formatOpts ...string) string {
+	res := "\033["
+	for _, c := range formatOpts {
+		res = fmt.Sprintf("%s%s;", res, c)
+	}
+	return fmt.Sprintf("%s%sm", res, code)
 }
 
 func makeColorFunc(code string) colorFunc {
--- a/cmd/formatter/shortcut.go
+++ b/cmd/formatter/shortcut.go
@@ -1,0 +1,321 @@
+/*
+   Copyright 2024 Docker Compose CLI authors
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package formatter
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"math"
+	"os"
+	"syscall"
+	"time"
+
+	"github.com/buger/goterm"
+	"github.com/compose-spec/compose-go/v2/types"
+	"github.com/docker/compose/v2/internal/tracing"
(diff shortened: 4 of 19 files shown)


pkg/compose/up.go
```diff
@@ -88,6 +96,27 @@ func (s *composeService) Up(ctx context.Context, project *types.Project, options
 		}
 		for {
 			select {
+			case event := <-kEvents:
+				switch key := event.Key; key {
+				case keyboard.KeyCtrlC:
+					keyboard.Close()
+					formatter.KeyboardInfo.ClearInfo()
+					gracefulTeardown()
```
ndeloof, line 104: I wonder about the impact doing so as context is not cancelled.

go.mod
```diff
@@ -81,6 +83,7 @@ require (
 	github.com/docker/docker-credential-helpers v0.8.0 // indirect
 	github.com/docker/go v1.5.1-1.0.20160303222718-d30aec9fd63c // indirect
 	github.com/docker/go-metrics v0.0.1 // indirect
+	github.com/eiannone/keyboard v0.0.0-20220611211555-0d226195f203
```
ndeloof, line 86: Probably does what we expect, by I wonder we want to rely on a dependency which hasn't been updated for 2 years

pkg/compose/up.go
ndeloof: I'm not sure. We rely on `<-ctx.Done()` in a few place to stop goroutines after Ctrl+C, typically to stop collecting logs, watch container events, etc. I assume after the application stopped those will be released anyway, but this might bring some new race conditions.

laurazard: I think it should be okay here (although it makes me think the graceful termination code/`doneCh`/`ctx.Done()` handling could use a refactor).

I wonder if cancelling the context here would be better.

Also I wonder, since we're already capturing `SIGINT` above (when we do `signal.Notify(signalChan, syscall.SIGINT, syscall.SIGTERM)`, if the new `keyboard` stuff is going to interfere with that.

cmd/formatter/shortcut.go
```diff
@@ -0,0 +1,73 @@
...
+	height := goterm.Height()
+	fmt.Print("\0337") // save cursor position
+	if lk.err != nil {
+		fmt.Printf("\033[%d;0H", height-1)
+		fmt.Print("\033[2K") // clear line
+	}
+	fmt.Printf("\033[%d;0H", height) // Move to last line
+	fmt.Print("\033[2K")             // clear line
+	fmt.Print("\0338")               // restore cursor position
+}
```
laurazard, line 68: Just thinking/not a blocker, but I'm starting to wonder if we should think about migrating over to an actual TUI lib as opposed to rolling our own. `bubbletea` is pretty nice for this kind of stuff.

cmd/formatter/shortcut.go
```diff
@@ -0,0 +1,73 @@
...
+
+func (lk *LogKeyboard) ClearInfo() {
+	height := goterm.Height()
+	fmt.Print("\0337") // save cursor position
+	if lk.err != nil {
+		fmt.Printf("\033[%d;0H", height-1)
+		fmt.Print("\033[2K") // clear line
+	}
+	fmt.Printf("\033[%d;0H", height) // Move to last line
+	fmt.Print("\033[2K")             // clear line
```
laurazard, line 66: Might be more readable to put all these hieroglyphs in `const`s

laurazard:
Looks really good overall! Happy with the "TUI" improvements 🥳 

I think that's the plan from what @jhrotko has mentioned, but I wanted to confirm that the idea is to use https://github.com/docker/compose/pull/11593 and only enable this when running this on a Docker Desktop environment (cc @milas), so that people running this headless/on bare Linux w/o DD don't get confused.

cmd/compose/up.go
```diff
@@ -127,6 +128,7 @@ func upCommand(p *ProjectOptions, dockerCli command.Cli, backend api.Service) *c
 	flags.BoolVar(&up.wait, "wait", false, "Wait for services to be running|healthy. Implies detached mode.")
 	flags.IntVar(&up.waitTimeout, "wait-timeout", 0, "Maximum duration to wait for the project to be running|healthy")
 	flags.BoolVarP(&up.watch, "watch", "w", false, "Watch source code and rebuild/refresh containers when files are updated.")
+	flags.BoolVar(&up.navigationBar, "navigation-menu", true, "While running in attach mode, enable shortcuts and shortcuts info bar.")
```
ndeloof, line 131: "shortcuts and shortcuts" ?

jhrotko: Did a fix. Let me know what do you think

cmd/formatter/shortcut.go
jhrotko: Did huge refactor on this file. Was still work in progress at the time. It should be more readable now :)

cmd/formatter/shortcut.go
jhrotko: agreed, we should consider this lib for a further interaction if this feature gets traction

pkg/compose/up.go
jhrotko: Changed the way we handle the tear down. I am sending a SIGTERM signal to signalChan

docs/reference/docker_compose_up.yaml
```diff
@@ -108,6 +108,16 @@ options:
       experimentalcli: false
       kubernetes: false
       swarm: false
+    - option: menu
+      value_type: bool
+      default_value: "false"
+      description: While running in attach mode, enable helpful shortcuts.
```
laurazard, line 114:
```suggestion
      description: Enable interactive shortcuts when running attached. Incompatible with --detach.
```

Dockerfile
```diff
@@ -101,6 +101,7 @@ RUN --mount=type=bind,target=. \
 FROM build-base AS test
 ARG CGO_ENABLED=0
 ARG BUILD_TAGS
+ENV COMPOSE_MENU=FALSE
```
laurazard, line 104: Is this necessary for building?

cmd/formatter/colors.go
```diff
@@ -73,10 +73,11 @@ var monochrome = func(s string) string {
...
 func ansiColor(code, s string) string {
-	return fmt.Sprintf("%s%s%s", ansi(code), s, ansi("0"))
+	return fmt.Sprintf("%s%s%s", ansiColorCode(code), s, ansiColorCode("0"))
 }
 
-func ansi(code string) string {
+// Everything about ansiColorCode color https://hyperskill.org/learn/step/18193
+func ansiColorCode(code string) string {
 	return fmt.Sprintf("\033[%sm", code)
 }
```
laurazard, line 82: Didn't realize we had this – isn't the stuff in `aec` enough? I think we use it sometimes https://github.com/docker/compose/blob/9b0e3d53eb4191fa031fafe65f74ef377ba5c99d/pkg/progress/colors.go#L30-L36

laurazard: then it's used like https://github.com/docker/compose/blob/9b0e3d53eb4191fa031fafe65f74ef377ba5c99d/pkg/progress/tty.go#L258

Dockerfile
jhrotko: yes, because by default it should be true, in order not to break tests this is easier to set the COMPOSE_MENU false in all containers

cmd/formatter/colors.go
jhrotko: I need to add the bold to the prefix color :( `\033[36m` -> cyan vs` \033[1;36m` -> bold + cyan

cmd/compose/up.go
```diff
@@ -181,6 +186,9 @@ func runUp(
 	if err != nil {
 		return err
 	}
+	if !upOptions.navigationMenuChanged {
```
dock-tor, line 189: seems bit complicated to manage value set by env var, is there any drawback doing the same as https://github.com/docker/compose/blob/main/cmd/compose/up.go#L109-L110 ?

dock-tor (approved):

cmd/compose/up.go
ndeloof: (this comment is mine, I was connected with by @docker.com email while commenting :P)

jhrotko: we also want to have into consideration the value of the Docker Desktop experimental feature value 
Screenshot 2024-03-21 at 16 05 49

ndeloof: oh indeed

cmd/formatter/shortcut.go
```diff
@@ -0,0 +1,324 @@
...
+)
+
+const DISPLAY_ERROR_TIME = 10
+
+type KeyboardError struct {
+	err       error
+	timeStart time.Time
+}
+
+func (ke *KeyboardError) shoudlDisplay() bool {
```
milas, line 45: s/shoudlDisplay/shouldDisplay

cmd/formatter/shortcut.go
```diff
@@ -0,0 +1,324 @@
...
+		menu := lk.navigationMenu()
+
+		MoveCursorX(0)
+		SaveCursor()
+
+		lk.kError.printError(height, menu)
+
+		MoveCursor(height-linesOffset(menu), 0)
+		ClearLine()
+		fmt.Print(menu)
```
milas, line 181: (for follow-up)

You can make it easier to test this by storing stdout/`io.Writer` on the type and writing to that.

cmd/formatter/shortcut.go
```diff
@@ -0,0 +1,324 @@
...
+
+		MoveCursorX(0)
+		RestoreCursor()
+	}
+}
+
+func (lk *LogKeyboard) navigationMenu() string {
```
milas, line 188: fyi `strings.Builder` can sometimes be handy for these types of finicky formatting formatting type functions

(this is readable/fine as-is! no need to change, just sharing)

https://pkg.go.dev/strings#Builder

cmd/formatter/shortcut.go
```diff
@@ -0,0 +1,324 @@
...
+				err := fmt.Errorf("Watch is not yet configured. Learn more: %s", ansiColor(CYAN, "https://docs.docker.com/compose/file-watch/"))
+				lk.kError.addError("Watch", err)
+				return err
+			}))
+		return
+	}
+	lk.Watch.switchWatching()
+	if !lk.Watch.isWatching() && lk.Watch.Cancel != nil {
+		lk.Watch.Cancel()
+	} else {
```
milas, line 260: should this be 

```go
	lk.Watch.switchWatching()
	if !lk.Watch.isWatching() {
		if lk.Watch.Cancel != nil {
			lk.Watch.Cancel()
		}
	} else {
```

(that is, even if the cancel function is nil, we only want to start watch if `isWatching == true`, right?)

pkg/compose/compose.go
```diff
@@ -320,3 +320,7 @@ func (s *composeService) RuntimeVersion(ctx context.Context) (string, error) {
 	return runtimeVersion.val, runtimeVersion.err
 
 }
+
+func (s *composeService) isDesktopIntegrationActive() bool {
```
milas, line 324: 👍🏻

milas (approved):
LGTM - few random comments, biggest question is over that one conditional but otherwise no major concerns from me

cmd/formatter/shortcut.go
jhrotko: you mean, instead of using fmt.Print first pipe all the characters to stdout and then print?

cmd/formatter/shortcut.go
jhrotko: yes, I think this was an old bug that was fixed, I think it was because the configuration validation was being done after and not before at some point. Both expressions are equivalent

cmd/formatter/shortcut.go
ndeloof: add an `io.Writer` field in `KeyboardError` type, and use `fmt.Fprint`. This will allow unit tests to pass a `bytes.Buffer` as writer and check output
For actual use, should be initialized to `dockercli.Out()` to respect docker CLI plugin architecture

cmd/formatter/shortcut.go
ndeloof: nice catch

pkg/compose/up.go
```diff
@@ -124,7 +146,7 @@ func (s *composeService) Up(ctx context.Context, project *types.Project, options
 		return err
 	})
 
-	if options.Start.Watch {
+	if options.Start.Watch && !options.Start.NavigationMenu {
```
ndeloof, line 149: IIUC `--watch` would then **not** start watch mode until menu is disabled. Why ?

pkg/compose/watch.go
```diff
@@ -159,17 +170,15 @@ func (s *composeService) Watch(ctx context.Context, project *types.Project, serv
...
 			defer watcher.Close() //nolint:errcheck
 			return s.watch(ctx, project, service.Name, options, watcher, syncer, config.Watch)
 		})
 	}
-
 	if !watching {
 		return fmt.Errorf("none of the selected services is configured for watch, consider setting an 'develop' section")
 	}
-	options.LogTo.Log(api.WatchLogger, "watch enabled")
+	options.LogTo.Log(api.WatchLogger, "Watch started")
```
ndeloof, line 181: I personally preferred "enabled", but that's just me

pkg/compose/up.go
jhrotko: If the --watch is passed and the menu is active, this will start the watch command here https://github.com/docker/compose/pull/11605/files#diff-809980e7b57a811e276a5d7951e759d3071d157ebad94f0a52370476a1fbfcedR103-R105

pkg/compose/watch.go
jhrotko: I agree, even the navigation menu message switches between `watch enabled/disabled`

pkg/compose/up.go
ndeloof: right. yet more spaghetti code to our already confusing "architecture" 😅

pkg/compose/compose.go
jhrotko: I refactored other places in the code that also did this validation

PR #11614, opened 2024-03-13
feat(desktop): synchronized file share integration

milas:
**What I did**
Initial integration between Docker Desktop's "Synchronized File Shares" functionality and Compose.

When running under Docker Desktop, File Shares will be created/used for any bind mount paths. (There's a missing feature check that still needs to be added, but ALL desktop functionality is currently gated behind the `COMPOSE_EXPERIMENTAL_DESKTOP` env var.)

Compose will wait for the File Share to reach steady state, i.e. having completed initial synchronization. The progress is shown, similar to how layer pulls for images are displayed.

**Related issue**
JIRA:
* https://docker.atlassian.net/browse/COMP-465
* https://docker.atlassian.net/browse/COMP-466

happy looking platypus on someone's hand

### DIFF
--- a/internal/desktop/client.go
+++ b/internal/desktop/client.go
@@ -17,14 +17,18 @@
 package desktop
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
+	"io"
 	"net"
 	"net/http"
 	"strings"
 
 	"github.com/docker/compose/v2/internal/memnet"
+	"github.com/r3labs/sse"
 	"go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp"
 )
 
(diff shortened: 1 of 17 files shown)


ndeloof (approved):
LGTM
My understanding was we wanted file share to be used only for watch sync, not for all bind mounts, maybe I missed some discussion :)

internal/desktop/file_shares.go
```diff
@@ -0,0 +1,387 @@
...
+		w.TailMsgf("Synchronized File Shares not removed due to error: %v", err)
+		return
+	}
+	// filter the list first, so we can early return and not show the event if
+	// there's no sessions to clean up
+	var toRemove []FileShareSession
+	for _, share := range existing {
+		if share.Labels["com.docker.compose.project"] == projectName {
+			toRemove = append(toRemove, share)
+		}
```
xenoscopic, line 55: We could extend the file shares API with a mechanism to remove these using a label selector.  I'm not sure if that would make things more robust; it might make status updates more difficult.

internal/desktop/client.go
```diff
@@ -84,6 +88,175 @@ func (c *Client) Ping(ctx context.Context) (*PingResponse, error) {
...
+type FileShareSession struct {
+	SessionID string            `json:"identifier"`
+	Alpha     FileShareEndpoint `json:"alpha"`
+	Beta      FileShareEndpoint `json:"beta"`
+	Labels    map[string]string `json:"labels"`
+	Status    string            `json:"status"`
+}
+
+func (c *Client) ListFileShares(ctx context.Context) ([]FileShareSession, error) {
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, backendURL("/mutagen/file-shares"), http.NoBody)
```
xenoscopic, line 145: Maybe we should add some comments in the backend that these URL patterns need to stay fixed.  I guess we could add some sort of service discovery mechanism, but that seems like overkill.

internal/desktop/file_shares.go
```diff
@@ -0,0 +1,387 @@
...
+			fileShareID = fss.Beta.Path
+			logrus.Debugf("Found existing suitable file share %s for path %q [%s]", fileShareID, path, fss.Alpha.Path)
+			wait.addShare(fileShareID)
+			handler(*fss)
+			continue
+		} else {
+			req := CreateFileShareRequest{
+				HostPath: path,
+				Labels: map[string]string{
+					"com.docker.compose.project": m.projectName,
```
xenoscopic, line 225: We may need to sanitize `m.projectName` to comply with Mutagen's (i.e. Kubernetes') label value syntax (maybe it already is, I'm not sure).  I don't think it will affect the loop-based approach you have above, but it will break `LabelSelector`-based selection in the Mutagen synchronization manager (possibly/probably for other sessions too).

xenoscopic (commented):
LGTM, my only real concern is the project label values, but maybe Compose's normalization is sufficient?  If we can add label validation to the backend, then it's a little less worrying that we might ingest an invalid value.

pkg/compose/up.go
```diff
@@ -34,8 +34,7 @@ import (
 func (s *composeService) Up(ctx context.Context, project *types.Project, options api.UpOptions) error { //nolint:gocyclo
 	err := progress.Run(ctx, tracing.SpanWrapFunc("project/up", tracing.ProjectOptions(ctx, project), func(ctx context.Context) error {
 		w := progress.ContextWriter(ctx)
-		w.HasMore(options.Start.Attach == nil)
```
jhrotko, line 37: not necessary anymore?

jhrotko (approved):

PR #11633, opened 2024-03-19
feat(experiments): add experimental feature state

milas:
**What I did**
Use environment variable for global opt-out and Docker Desktop (if
available) to determine specific experiment states.

In the future, we'll allow per-feature opt-in/opt-out via env vars
as well, but currently there is a single `COMPOSE_EXPERIMENTAL` env
var that can be used to opt-out of all experimental features
independently of Docker Desktop configuration.

**Related issue**
* https://docker.atlassian.net/browse/COMP-456
* https://docker.atlassian.net/browse/COMP-487


jhrotko (approved):
LGTM

ndeloof (approved):

PR #11641, opened 2024-03-20
feat(otel): initialize meter provider from Docker context

milas:
**What I did**
In addition to tracing, add support for metering (counters/gauges/etc) via OpenTelemetry with similar OTLP/gRPC auto-enablement. User-defined endpoints not currently supported; this will be added as a follow-up.

**Related issue**
JIRA: [COMP-489]

[COMP-489]: https://docker.atlassian.net/browse/COMP-489?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


nicksieger (approved):
Good stuff!

ndeloof (approved):

PR #11643, opened 2024-03-21
introduce config --variables to list compose model variables

ndeloof:
**What I did**
introduce `config --variables` pseudo-command to list compose model variables


ndeloof:
set as draft waiting for https://github.com/compose-spec/compose-go/pull/608

Issue #10596, opened 2023-05-22
[BUG] compose up --wait exits 1 on init containers successfully completing


nmguse-bighealth:
### Description

When using `docker compose up --wait`, where I have an init container that populates some data and then exits (0), docker compose will exit 1.

If any other container depends on the init container finishing, it will exit properly, as described/exampled in this PR: https://github.com/docker/compose/pull/9572. In my situation however, there is no container that runs a service that depends on the init container finishing.

### Steps To Reproduce

docker-compose.yml example:
```
version: '3'

services:
  postgres:
    image: postgres:14-alpine
    environment:
      POSTGRES_DB: postgres
      POSTGRES_USER: postgres
      POSTGRES_PASSWORD: postgres
    healthcheck:
      test: ['CMD', 'pg_isready']

  postgres_setup:
    image: alpine
    depends_on:
      postgres:
        condition: service_healthy
    restart: "no"
    command: pwd
```

Run docker compose/print exit code:
```
$ docker compose up --wait
[+] Running 2/3
 ✔ Network test_default             Created                                                                                                                                                    0.1s
 ✔ Container test-postgres-1        Healthy                                                                                                                                                   31.9s
 ⠿ Container test-postgres_setup-1  Waiting                                                                                                                                                   31.9s
container test-postgres_setup-1 exited (0)

$ echo $?
1
```

### Compose Version

```Text
Docker Compose version v2.17.3
```

### Docker Environment

```Text
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.10.4
    Path:     /Users/nathan/.docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.17.3
    Path:     /Users/nathan/.docker/cli-plugins/docker-compose
  dev: Docker Dev Environments (Docker Inc.)
    Version:  v0.1.0
    Path:     /Users/nathan/.docker/cli-plugins/docker-dev
  extension: Manages Docker extensions (Docker Inc.)
    Version:  v0.2.19
    Path:     /Users/nathan/.docker/cli-plugins/docker-extension
  init: Creates Docker-related starter files for your project (Docker Inc.)
    Version:  v0.1.0-beta.4
    Path:     /Users/nathan/.docker/cli-plugins/docker-init
  sbom: View the packaged-based Software Bill Of Materials (SBOM) for an image (Anchore Inc.)
    Version:  0.6.0
[...]

Issue #11638, opened 2024-03-20
[BUG] docker compose up -d --wait option with restart polilcy 'no'


skul87:
### Description

Hello, I have run `docker compose up -d --wait`
In docker-compose.yml there is a init service with restart police - 'no'

`docker compose up -d --wait` ends with return code 1 and my ci will fail.

may be if service return 0 code, then assume that everything is fine ?

```
[runner@partners-dev partners]$ cat docker/docker-compose.yml
services:
  db:
    build:
      args:
        MYSQL_TAG: ${MYSQL_TAG}
      context: ../
      dockerfile: docker/mysql/Dockerfile
    restart: always
    environment:
      TZ: ${TZ}
    healthcheck:
      test:
        [
          "CMD-SHELL",
          "sh -c 'mysqladmin ping -h 127.0.0.1 -u ${DB_USERNAME} --password=${DB_PASSWORD}'"
        ]
      interval: 5s
      timeout: 4s
      retries: 40
    volumes:
      - db-storage:/var/lib/mysql
      - db-logs:/var/log/mysql
  app:
    build:
      args:
        uid: ${UID}
        gid: ${GID}
        php_tag: ${PHP_TAG}
      context: ../
      dockerfile: docker/php-fpm/dev/Dockerfile
    entrypoint: [ "sleep" ]
    command: [ "999" ]
    restart: always
    environment:
      TZ: ${TZ}
    volumes:
      - ../app:/app
  init:
    build:
      args:
        uid: ${UID}
        gid: ${GID}
        php_tag: ${PHP_CLI_TAG}
      context: ../
      dockerfile: docker/php-fpm/dev/Dockerfile
    entrypoint: [ "date" ]
    restart: "no"
    environment:
      TZ: ${TZ}
    volumes:
      - ../app:/app

volumes:
  db-storage:
  db-logs:


networks:
  default:
```



```
[runner@partners-dev partners]$ docker compose version
Docker Compose version v2.25.0
[runner@partners-dev partners]$ docker compose up -d --wait
[+] Running 2/6
 ⠹ Network example-compose98_default      Created                                                                                                                                                                                                                             139.3s
 ⠹ Volume "example-compose98_db-storage"  Created                                                                                                                                                                                                                             139.2s
 ⠹ Volume "example-compose98_db-logs"     Created                                                                                                                                                                                                                             139.2s
 ✔ Container example-compose98-app-1      Healthy                                                                                                                                                                                                                               1.2s
[...]

PR #11649, opened 2024-03-22
Support disabling error for services exiting with code 0 in `wait` mode

pkoutsovasilis:
**What I did**
This PR introduces a flag for `compose up`, namely `--wait-allow-exit`, that targets wait mode (`--wait`) and enables treating services exited with a code of 0 not as an error

**Related issue**

closes https://github.com/docker/compose/issues/10596 and https://github.com/docker/compose/issues/11638


ndeloof:
I don't think adding an additional flag is the right solution, we miss a way to declare a more explicit init container and have compose manage it accordingly. `service_completed_successfully` unlocked a few usages but is only a partial solution

pkoutsovasilis:
> I don't think adding an additional flag is the right solution, we miss a way to declare a more explicit init container and have compose manage it accordingly. `service_completed_successfully` unlocked a few usages but is only a partial solution

I mean sure having a more explicit way to declare init containers, I agree, sounds like the way to go. But until that's implemented, a minor change like adding an additional flag could still be beneficial for managing init containers more effectively, right?! I think that when such a way becomes a thing, the code of this PR can be repurposed/removed easily.

ndeloof:
such a "minor change" we would have to support and maintain for years, so I prefer to say "no" and take time to find a better way, even this takes bit more time, than making the code ever more complicated by introducing some more corner cases

PR #11654, opened 2024-03-22
Remove COMPOSE_MENU env from e2e tests

jhrotko:
**What I did**
Removed COMPOSE_MENU from env in Dockerfile and Makefile
**Related issue**

image


milas (approved):
LGTM but tbh the more I think about it, it seems like we shouldn't need to specify `COMPOSE_MENU` / `--menu` for tests at all, I'd assume everything would be detected as NOT a TTY and thus not try to show the menu 🤔

glours (approved):
I agree with @milas we should not have the `--menu` in e2e test, Compose should detect TTY isn't available when running e2e in CI

ndeloof:
if removal for COMPOSE_MENU require `--menu=false` to be added to all our e2e test I'd prefer we keep the variable!

PR #11664, opened 2024-03-26
Handle --no-build and --watch args

jhrotko:
**What I did**
Fix case when running
```
docker compose up --watch --no-build
```
and 
```
docker compose up --no-build --menu (and then hit w - triggering watch)
```
It was throwing a panic since we were trying to access `options.Create.Build` which was nil.

**Related issue**

fixes https://docker.atlassian.net/browse/COMP-501

### DIFF
--- a/cmd/compose/up.go
+++ b/cmd/compose/up.go
@@ -137,6 +137,7 @@ func upCommand(p *ProjectOptions, dockerCli command.Cli, backend api.Service, ex
 	return upCmd
 }
 
+//nolint:gocyclo
 func validateFlags(up *upOptions, create *createOptions) error {
 	if up.exitCodeFrom != "" {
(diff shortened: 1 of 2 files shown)


milas (approved):
LGTM! I think this makes the most sense for now given how Compose mutates the `<svc>.build` entries in the project.

In the future, it'd be nice if we could get to the point where it does not need to `nil` those out, so that they're still available for _future_ rebuilds (via `watch.action: rebuild` for example), but that's a substantial change

ndeloof (approved):

ndeloof:
LGTM, while `--watch` does not require build until _some_ service use the `rebuild` action, so we could be clever here and detect this as develop config is being parsed. Anyway still better than a panic 😅

PR #11671, opened 2024-03-27
Change menu information text to dim

jhrotko:
**What I did**
Change hard coded gray color to use dim code `\033[2m` instead and let the terminal handle the dimming of the default color text.

**Related issue**

fixes https://docker.atlassian.net/jira/software/c/projects/COMP/boards/576?selectedIssue=COMP-502

image

### DIFF
--- a/cmd/formatter/logs.go
+++ b/cmd/formatter/logs.go
@@ -102,8 +102,6 @@ func (l *logConsumer) Err(container, message string) {
 	l.write(l.stderr, container, message)
 }
 
-var navColor = makeColorFunc("90")
-
 func (l *logConsumer) write(w io.Writer, container, message string) {
 	if l.ctx.Err() != nil {
 		return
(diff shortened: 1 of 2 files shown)


glours (approved):
👍

cmd/formatter/colors.go
```diff
@@ -37,7 +37,7 @@ var names = []string{
 
 const (
 	BOLD      = "1"
-	FAINT     = "2"
+	DIM       = "2"
```
ndeloof, line 40: I always seen this one referred as "FAINT", as listed on https://en.wikipedia.org/wiki/ANSI_escape_code

ndeloof (approved):

cmd/formatter/colors.go
jhrotko: reverted the change

PR #11672, opened 2024-03-28
Does not start menu manager if there is no tty on up

jhrotko:
**What I did**
Handle when there is no tty available and navigation menu is enabled. Also gracefully handle key events when an error occurs instead of panic.
Also did small refactor in `cmd/up.go` related with `NavigationMenu` validation.

This can be tested locally by running:
```
(/opt/homebrew/opt/util-linux/bin/setsid ~/workspace/compose/bin/build/docker-compose up --menu=true) </dev/null |& cat
```

make sure you have `util-linux` install
```
brew install util-linux
```

**Related issue**

fixes https://docker.atlassian.net/browse/DDB-25

image

### DIFF
--- a/cmd/compose/up.go
+++ b/cmd/compose/up.go
@@ -79,6 +79,16 @@ func (opts upOptions) apply(project *types.Project, services []string) (*types.P
 	return project, nil
 }
 
+func (opts *upOptions) validateNavigationMenu(dockerCli command.Cli, experimentals *experimental.State) {
+	if !dockerCli.Out().IsTerminal() {
+		opts.navigationMenu = false
+		return
+	}
+	if !opts.navigationMenuChanged {
+		opts.navigationMenu = SetUnchangedOption(ComposeMenu, experimentals.NavBar())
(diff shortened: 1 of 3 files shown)


cmd/compose/up.go
```diff
@@ -79,6 +80,16 @@ func (opts upOptions) apply(project *types.Project, services []string) (*types.P
 	return project, nil
 }
 
+func (opts *upOptions) validateNavigationMenu(experimentals *experimental.State) {
+	if !opts.navigationMenuChanged {
+		opts.navigationMenu = SetUnchangedOption(ComposeMenu, experimentals.NavBar())
+	}
+
+	if opts.navigationMenu && !streams.NewIn(os.Stdout).IsTerminal() {
```
ndeloof, line 88: you should access dockerCli.Out() directly here

jhrotko: changed to `dockerCli.Out().isTerminal`

glours (approved):

Issue #10225, opened 2023-02-01
Fail the docker-compose command if any one of the conatiner failed


Gobi-Ganesan:
### Description

I'm using the following docker-compose file

``````
version: '3.4'
services:
    one:
        image: busybox
        command: 'true'

    two:
        image: busybox
        command: 'false'

    three:
        image: busybox
        command: 'true'
``````
   
 I want to report an error  If any one of the containers failed.  `--exit-code-from` Only works for selecting a single container. 
` --abort-on-container-exit` will give exit code 0 if service one or three is completed first.

glours:
What are you trying to do exactly? Can you explain your use case? 
`--exit-code-from` return the exit code of the specified service container
`--abort-on-container-exit` will stop as soon as one container exit and you will get the exit code of this container.

Usually services are designed to run long process such as backend servers, databases... so using `--abort-on-container-exit` will work in this kind of cases. In your example all the services run commands which stop the containers immediately,  so you get the exit code of the first one to stop

Gobi-Ganesan:
@glours I'm running tests using compose. I want to report an error if any one of the tests gets failed.
Currently, it shows exit code 0.

sureshsgith:
Hello @Gobi-Ganesan

You can use the **' docker-compose up '** command with the **' --exit-code-from '** option to report an error if any one of the containers failed. The **' --exit-code-from '** option allows you to specify a single container for which you want to receive an error exit code in case of failure. To achieve your desired behavior, you can use a script that loops through the containers and checks their exit codes, and exits with a non-zero code if any of the containers have failed.

Here is an example shell script that accomplishes this:

`#!/bin/bash

services=(one two three)

for service in "${services[@]}"; do
  docker-compose up --exit-code-from $service
  if [ $? -ne 0 ]; then
    exit 1
  fi
done
`
This script starts the services one by one, and if any of the services exit with a non-zero code, the script will exit with a non-zero code as well, indicating that one of the services failed.


Suresh

Gobi-Ganesan:
@sureshsgith Thanks for the info. But here it will wait for service `one` to gets complete its test before starting the next service `two`. 
I need the option to capture exit status from multiple services.  Convert  `--exit-code-from` to a list or something.

Bessonov:
[...]

PR #11680, opened 2024-04-03
Introduce --abort-on-container-failure

ndeloof:
**What I did**

Introduce `--abort-on-container-failure` as a custom flavor for cascade stop, which only stops on container failure (exit != 0). This allows to run "tasks" defined by services (sorry for the conceptual mismatch) and complete successfully if all succeeded, but fail if any failed, and kill the others (aka "fail fast")

**Related issue**
closes https://github.com/docker/compose/issues/10225

image

### DIFF
--- a/cmd/compose/up.go
+++ b/cmd/compose/up.go
@@ -46,6 +46,7 @@ type upOptions struct {
 	noStart               bool
 	noDeps                bool
 	cascadeStop           bool
+	cascadeFail           bool
 	exitCodeFrom          string
 	noColor               bool
 	noPrefix              bool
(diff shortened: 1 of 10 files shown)


cmd/compose/up.go
```diff
@@ -131,6 +143,7 @@ func upCommand(p *ProjectOptions, dockerCli command.Cli, backend api.Service, ex
 	flags.BoolVar(&create.noRecreate, "no-recreate", false, "If containers already exist, don't recreate them. Incompatible with --force-recreate.")
 	flags.BoolVar(&up.noStart, "no-start", false, "Don't start the services after creating them")
 	flags.BoolVar(&up.cascadeStop, "abort-on-container-exit", false, "Stops all containers if any container was stopped. Incompatible with -d")
+	flags.BoolVar(&up.cascadeFail, "abort-on-container-failure", false, "Stops all containers if any container exited with failure. Incompatible with -d")
```
glours, line 146: Incompatible with `-d` or `cascadeStop` or both? 
If not compatible with `-d`, I think you forgot to add the associated check in `validateFlags` function

glours (commented):
Some questions about the command flags compatibility before approving

glours (approved):
LGTM

Issue #11681, opened 2024-04-03
[BUG] systemd-journald recognizes logs as blob


lms-jk:
### Description

I am using a systemd service wrapper to start my containers with docker compose. Up until now, the logs in journald looked as expected:
```
Mär 30 13:07:54 lmsman-rnddh01 docker[600]: application-1  | 2024/03/30 13:07:54 ...
Mär 30 13:07:54 lmsman-rnddh01 docker[600]: application-1  | 2024/03/30 13:07:54 ...
Mär 30 13:07:54 lmsman-rnddh01 docker[600]: application-1  | 2024/03/30 13:07:54 ...
```
After updating from 2.24.7 -> 2.26.1 the logs in journald are recognized as binary data:
```
Apr 03 09:19:00 lmsman-rnddh01 docker[612]: [238B blob data]
Apr 03 09:19:03 lmsman-rnddh01 docker[612]: [232B blob data]
Apr 03 09:19:04 lmsman-rnddh01 docker[612]: [252B blob data]
```

### Steps To Reproduce

Create a systemd unit `/etc/systemd/system/compose.service` with the following content:

```
[Unit]
Requires=docker.service
After=docker.service

[Service]
WorkingDirectory=<path_to_docker-compose.yaml>
ExecStart=/usr/bin/docker compose up
ExecStop=/usr/bin/docker compose down

[Install]
WantedBy=multi-user.target
```
Start the service and view the logs:

`systemctl daemon-reload && systemctl start compose.service && journalctl -u compose.service -f`

### Compose Version

```Text
Docker Compose version 2.26.1
```

### Docker Environment

```Text
Client:
 Version:    26.0.0
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  0.13.1
    Path:     /usr/lib/docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  2.26.1
    Path:     /usr/lib/docker/cli-plugins/docker-compose

Server:
 Containers: 6
  Running: 6
  Paused: 0
  Stopped: 0
 Images: 8
 Server Version: 26.0.0
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: true
  Native Overlay Diff: false
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
[...]

PR #11682, opened 2024-04-03
don't use ansi espace sequence when disabled

ndeloof:
**What I did**
when ansi is disabled or output is not a terminal, fully disable any ANSI magic

**Related issue**
closes https://github.com/docker/compose/issues/11681

### DIFF
--- a/cmd/formatter/ansi.go
+++ b/cmd/formatter/ansi.go
@@ -22,36 +22,65 @@ import (
 	"github.com/acarl005/stripansi"
 )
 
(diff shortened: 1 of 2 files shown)


glours (approved):
👍

lms-jk:
Can confirm that this patch will resolve #11681. Thank you!

ndeloof:
Let's merge this with this quick and dirty fix, will need to revisit the `formatter` package at some point as this becomes spaghetti code

Issue #11690, opened 2024-04-05
Filter out empty DNS entries


IvanRibakov:
### Description

Previously this DockerCompose configuration would successfully start a container:
```
# ANOTHER_SERVICE_IP is not set and defaults to blank string
dns: ${ANOTHER_SERVICE_IP}
```

But using latest Docker+DockerCompose on Linux
```
Docker version 26.0.0, build 2ae903e
Docker Compose version v2.25.0
```
container fails to start:
```
WARN[0000] The "ANOTHER_SERVICE_IP" variable is not set. Defaulting to a blank string. 
Error response from daemon: bad nameserver address : ParseAddr(""): unable to parse IP
```
https://github.com/moby/moby/blob/d25b0bd7ea6ce17ca085c54d5965eeeb66417e52/libnetwork/sandbox_dns_unix.go#L259-L269

I can see how it makes sense from Docker perspective to be strict about supplied configuration options. I'm wondering whether it would be possible for the DockerCompose to filter out empty DNS entries before passing them to Docker and logging a warning. As far as I understand there is no other way to have a conditional behaviour that either uses external DNS service or the default host one without creating 2 different service configurations.

enmanuelmoreira:
Same problem here

jhrotko:
@IvanRibakov Could you provide an example?

IvanRibakov:
I went as far as to provide a [pull request](https://github.com/docker/compose/pull/11691) which unfortunately was rejected. 
@jhrotko what exactly are you looking for from the example that is missing from the current issue description?

IvanRibakov:
@ndeloof Than you for the https://github.com/compose-spec/compose-go/pull/700 . I see in the PR that `services.*.dns` is a hardcoded field. Am I right that this feature can be applied to any other service field using YAML tags similarly to [`!reset` ](https://docs.docker.com/reference/compose-file/merge/#reset-value)or [`!override`](https://docs.docker.com/reference/compose-file/merge/#replace-value)? Or is this not available at the moment?

ndeloof:
@IvanRibakov this is a generic mechanims, but I only configured it for `dns` with the intent we can add more step by step

IvanRibakov:
Ok, so no `!omitempty` YAML tag at the moment, right?

ndeloof:
yes indeed, this is hard-coded support in compose-go

PR #11691, opened 2024-04-05
Strip empty DNS entries to avoid Moby parse errors when creating sandbox DNS

IvanRibakov:
**What I did**

Ensured that each service's DNS settings are stripped from empty entries that could result from using ENV variables that defaulted to blank strings.

For example:
```
  service:
    # ANOTHER_SERVICE_IP is not set and defaults to blank string
    dns: ${ANOTHER_SERVICE_IP}
    ...
```

**Related issue**
https://github.com/docker/compose/issues/11690

### DIFF
--- a/pkg/compose/create.go
+++ b/pkg/compose/create.go
@@ -99,6 +99,8 @@ func (s *composeService) create(ctx context.Context, project *types.Project, opt
 		return err
 	}
 
+	prepareServiceDNS(project)
+
 	if err := s.ensureProjectVolumes(ctx, project); err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


IvanRibakov:
Is there anything I could do to help move this PR forward?

glours (approved):
LGTM

ndeloof:
Implementation is correct but IMHO this issue will apply to way more places, not just `dns`. We miss a way to mark an entry in `compose.yaml` as "to be ignored if value is empty". I wonder it would be better we introduce a global mechanism for this purpose, maybe comparable to [`!override`](https://docs.docker.com/compose/compose-file/13-merge/#replace-value):

```yaml
 service:
    # ANOTHER_SERVICE_IP is not set and defaults to blank string
    dns: !omitempty ${ANOTHER_SERVICE_IP} 
    ...
```

Issue #10930, opened 2023-08-24
Query About Creating Volumes Without Services in Docker Compose v2


fti-rgurav:
### Description

In Docker Compose v1, I was able to create volumes independently using the following syntax:
```
version: '3.2'
volumes:
  volume1:
    driver: local
    name: "volume1"
```
However, upon migrating to Docker Compose v2, I attempted to use the same approach. Unfortunately, I encountered an error message  **"no service selected."** It seems that the behavior has changed in v2 regarding how volumes are define

I'm looking for advice on how to create volumes in Docker Compose v2  without associating them with any specific service in Docker Compose v2. This was possible in v1, and I'm curious whether there's a workaround or new syntax in v2 that achieves the same result.

### Steps To Reproduce

1. Up Docker Compose with the YAML file using v1 syntax to define volumes.
Use the following YAML code:
```
version: '3.2'
volumes:
  volume1:
    driver: local
    name: "volume1"
```

2. Attempt to use the same file in Docker Compose v2 environment.
3. Observe the "no service selected" error.

### Compose Version

```Text
Docker Compose version v2.20.3
```

### Docker Environment

```Text
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.11.2
    Path:     /usr/libexec/docker/cli-plugins/docker-buildx

Server:
 Containers: 16
  Running: 16
  Paused: 0
  Stopped: 0
 Images: 63
 Server Version: 23.0.1
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: journald
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 2456e983eb9e37e47538f59ea18f2043c9a73640
 runc version: v1.1.4-0-g5fd4c4d
 init version: de40ad0
 Security Options:
  apparmor
  seccomp
   Profile: builtin
  cgroupns
 Kernel Version: 6.1.0-11-amd64
 Operating System: Debian GNU/Linux 12 (bookworm)
[...]

PR #11702, opened 2024-04-09
introduce `--all-resources` to _not_ exclude resources not used by services

ndeloof:
**What I did**
introduce `--all-resources` so that compose model don't get resources not used by services excluded
This allows to use `docker compose create` to manage non-services resources, when such a resource is shared by multiple compose application.

**Related issue**
closes https://github.com/docker/compose/issues/10930

image

### DIFF
--- a/cmd/compose/compose.go
+++ b/cmd/compose/compose.go
@@ -132,6 +132,7 @@ type ProjectOptions struct {
 	Compatibility bool
 	Progress      string
 	Offline       bool
+	All           bool
 }
 
 // ProjectFunc does stuff within a types.Project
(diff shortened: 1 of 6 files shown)


glours (approved):
Sounds good to me

milas (changes requested):
Blocking: Can we get some sort of test coverage please? 🙏🏻 

---

Non-blocking / my opinion: Compose Up is pretty service-oriented, and this feels like a band-aid for a niche use case. If declaratively managing volumes/networks/secrets/configs independent of containers is something we really want to support, I feel like it's something that'd be better to approach more generically/as a feature.

ndeloof:
added e2e test for this new flag

milas (approved):

PR #11713, opened 2024-04-11
check container_name is not in use by another service we will create

ndeloof:
**What I did**
check container_name is not in use by another service we will create

**Related issue**
see https://github.com/compose-spec/compose-go/pull/619

image

### DIFF
--- a/cmd/compose/config.go
+++ b/cmd/compose/config.go
@@ -169,6 +169,13 @@ func runConfig(ctx context.Context, dockerCli command.Cli, opts configOptions, s
 			return err
 		}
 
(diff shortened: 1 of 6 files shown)


glours (approved):
LGTM

glours:
Ho can we add a test to be sure we won't have regression in future as this check is managed here now and not in `compose-go` anymore?

ndeloof:
@glours I have a test case ready but need to wait for https://github.com/compose-spec/compose-go/pull/619 to be merged otherwise test will fail as compose-go will prevent loading compose.yaml file with same container_name used by services

Issue #11712, opened 2024-04-11
[BUG] `docker compose --exit-code-from` does not return correct exit code.


jeremiahrose:
### Steps To Reproduce

`docker-compose.yml`:
```
services:
  long-running:
    image: debian:bookworm-slim
    command: bash -c "sleep 20; exit 1"

  short-running:
    image: debian:bookworm-slim
    command: bash -c "sleep 5; exit 0"
```

In this example, the `long-running` service should never return an exit code of `0`. Either it completes and returns an exit code of `1`, or it is interrupted in which case it should return an exit code to indicate that it was interrupted.

However, when using `--exit-code-from long-running` an exit code of `0` is returned:
```
jez@mac % docker compose up --exit-code-from long-running
[+] Running 3/3
 ✔ Network dockercomposetest_default            Created                                                            0.0s 
 ✔ Container dockercomposetest-long-running-1   Created                                                            0.1s 
 ✔ Container dockercomposetest-short-running-1  Created                                                            0.1s 
Attaching to long-running-1, short-running-1
short-running-1 exited with code 0
Aborting on container exit...
[+] Stopping 2/2
 ✔ Container dockercomposetest-short-running-1  Stopped                                                            0.0s 
 ✔ Container dockercomposetest-long-running-1   Stopped                                                           10.3s 
jez@mac % echo $?                                        
0
```

### Expected outcome
`docker compose up --exit-code-from long-running` should return a non-zero exit code because `long-running` was interrupted and did not complete successfully.

### Observed outcome
`docker compose up --exit-code-from long-running` returned an error code of `0`.

### Why this matters
When using `docker-compose up` to run integration tests in CI, it is crucial that `--exit-code-from` does what it says, otherwise a test may produce a false positive result.

### Compose Version

```Text
Docker Compose version v2.24.6-desktop.1
```

### Docker Environment

```Text
Client:
 Version:    25.0.3
 Context:    desktop-linux
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.12.1-desktop.4
    Path:     /Users/jez/.docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.24.6-desktop.1
    Path:     /Users/jez/.docker/cli-plugins/docker-compose
  debug: Get a shell into any image or container. (Docker Inc.)
    Version:  0.0.24
    Path:     /Users/jez/.docker/cli-plugins/docker-debug
[...]

PR #11715, opened 2024-04-11
capture exit code for container being stopped, and return according to `--exit-code-from`

ndeloof:
**What I did**
capture container's exitCode on stop event, as container has been interrupted

**Related issue**
fixes https://github.com/docker/compose/issues/11712

image


glours (approved):
LGTM

milas (approved):

PR #11718, opened 2024-04-11
Fix #10661: docker compose up always kills the containers on second Ctrl-C

### LINKED ISSUE #10661: [BUG] docker compose up not returning shell and erroneously detecting error

jsoriano:
**What I did**

Hitting Ctrl-C twice in docker compose up is supposed to immediately kill the containers. 
Kill executed on the second Ctrl-C was filtering containers depending on its state. In some
cases the containers can reach an state for what these filters don't get any container, and
the command hangs while printing `no container to kill`.

Remove this filtering, so it tries to kill any container it finds, independently of their state.

**Related issue**

Fixes https://github.com/docker/compose/issues/10661

### DIFF
--- a/pkg/compose/up.go
+++ b/pkg/compose/up.go
@@ -119,11 +119,14 @@ func (s *composeService) Up(ctx context.Context, project *types.Project, options
 					gracefulTeardown()
 				} else {
 					eg.Go(func() error {
-						return s.Kill(context.Background(), project.Name, api.KillOptions{
+						// Intentionally ignore errors, for cases where some
+						// of the containers are already stopped.
+						s.kill(context.Background(), project.Name, api.KillOptions{
 							Services: options.Create.Services,
 							Project:  project,
 							All:      true,
 						})
+						return nil
 					})
 					return nil
 				}
--- a/pkg/compose/up.go
+++ b/pkg/compose/up.go
@@ -122,6 +122,7 @@ func (s *composeService) Up(ctx context.Context, project *types.Project, options
 						return s.Kill(context.Background(), project.Name, api.KillOptions{
 							Services: options.Create.Services,
 							Project:  project,
+							All:      true,
 						})
 					})
 					return nil
--- a/pkg/compose/up.go
+++ b/pkg/compose/up.go
@@ -21,7 +21,6 @@ import (
 	"fmt"
 	"os"
 	"os/signal"
-	"sync/atomic"
 	"syscall"
 
 	"github.com/compose-spec/compose-go/v2/types"
@@ -109,8 +108,6 @@ func (s *composeService) Up(ctx context.Context, project *types.Project, options
 			}
 		}
 
-		// killRunning is used to control that we don't run more than one kill if signals are spammed.
-		var killRunning atomic.Bool
 		for {
 			select {
 			case <-doneCh:
@@ -124,13 +121,7 @@ func (s *composeService) Up(ctx context.Context, project *types.Project, options
 					gracefulTeardown()
 					break
 				}
-				if !killRunning.CompareAndSwap(false, true) {
-					break
-				}
 				eg.Go(func() error {
-					defer killRunning.Store(false)
-					// Intentionally ignore errors, for cases where some
-					// of the containers are already stopped.
 					err := s.kill(context.Background(), project.Name, api.KillOptions{
 						Services: options.Create.Services,
 						Project:  project,
@@ -143,6 +134,7 @@ func (s *composeService) Up(ctx context.Context, project *types.Project, options
 
 					return err
 				})
+				return nil
 			case event := <-kEvents:
 				formatter.KeyboardManager.HandleKeyEvents(event, ctx, project, options)
 			}
--- a/pkg/compose/up.go
+++ b/pkg/compose/up.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"os"
 	"os/signal"
+	"sync/atomic"
 	"syscall"
 
 	"github.com/compose-spec/compose-go/v2/types"
@@ -29,6 +30,7 @@ import (
 	"github.com/docker/compose/v2/internal/tracing"
 	"github.com/docker/compose/v2/pkg/api"
 	"github.com/docker/compose/v2/pkg/progress"
+	"github.com/docker/docker/errdefs"
 	"github.com/eiannone/keyboard"
 	"github.com/hashicorp/go-multierror"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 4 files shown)


ndeloof:
AFAICT `kill` on a stopped container will fail
```
Error response from daemon: cannot kill container: dafcbcd023c1: container dafcbcd023c1c1bbed6b0a35a7e1e2420129672c327519698ef8cbd644fefd06 is not running
```
would need to ignore such an error, assuming this is the right fix for this race condition

jsoriano:
> AFAICT `kill` on a stopped container will fail
> 
> ```
> Error response from daemon: cannot kill container: dafcbcd023c1: container dafcbcd023c1c1bbed6b0a35a7e1e2420129672c327519698ef8cbd644fefd06 is not running
> ```

You are right, this was briefly showing an error. I have changed the code to use the internal kill method, that doesn't show the progress. And to ignore the returned error.

pkg/compose/up.go
```diff
@@ -119,10 +119,14 @@ func (s *composeService) Up(ctx context.Context, project *types.Project, options
 					gracefulTeardown()
 				} else {
 					eg.Go(func() error {
-						return s.Kill(context.Background(), project.Name, api.KillOptions{
+						// Intentionally ignore errors, for cases where some
+						// of the containers are already stopped.
+						s.kill(context.Background(), project.Name, api.KillOptions{
```
ndeloof, line 124: this should not ignore _all_ errors, but only the "not found" one

jsoriano: I don't find a reliable way to distinguish errors here, as they don't seem to be typed. Should I add the check based on the text of the error? I guess it would be to check if the error message contains `is not running`.

Also, with the current code, even if it returns the error, it doesn't seem to be doing anything about it. If the kill fails, the loop is still finished and the command hangs if there are still containers running. I could try to fix this too if I add error handling.

ndeloof: have you tried using `errdefs.IsNotFound(err)` ?

jsoriano: Thanks! This worked. I also checked `errdefs.IsConflict`, which happens if one of the containers is already stopped.

pkg/compose/up.go
```diff
@@ -117,15 +122,27 @@ func (s *composeService) Up(ctx context.Context, project *types.Project, options
...
 				if first {
 					gracefulTeardown()
-				} else {
-					eg.Go(func() error {
-						return s.Kill(context.Background(), project.Name, api.KillOptions{
-							Services: options.Create.Services,
-							Project:  project,
-						})
-					})
-					return nil
```
jsoriano, line 127: I am removing this return, so it is possible now to try to kill the containers again after one kill failed.

ndeloof: This should not be necessary, I hardly can imagine process could survive a SIGKILL
Also I'm not sure about the potential side-effects doing so, so if you don't mind I'd prefer to be conservative here.

jsoriano: Well, indeed hopefully no process is going to survive a SIGKILL :slightly_smiling_face: But this is sending requests to kill containers to the docker daemon through HTTP, so there are other things that could fail in the way.

In any case I agree with being conservative and keep previous behaviour, I will revert this.

ndeloof (approved):

jsoriano:
@ndeloof thanks for the reviews and the help on this change. Could you also take a look to https://github.com/docker/compose/pull/11719? Thanks!

Issue #11717, opened 2024-04-11
Healthcheck console timer is broken.


K2ouMais:
### Description

After updating Docker Desktop to 4.29.0 this seems to be broken again.

https://github.com/docker/compose/issues/11086
https://github.com/docker/compose/pull/11357

### Compose Version

```Text
Docker Compose version v2.26.1-desktop.1
```

### Docker Environment

```Text
Client:
 Version:    26.0.0
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.13.1-desktop.1
    Path:     /usr/local/lib/docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.26.1-desktop.1
    Path:     /usr/local/lib/docker/cli-plugins/docker-compose
  debug: Get a shell into any image or container. (Docker Inc.)
    Version:  0.0.27
    Path:     /usr/local/lib/docker/cli-plugins/docker-debug
  dev: Docker Dev Environments (Docker Inc.)
    Version:  v0.1.2
    Path:     /usr/local/lib/docker/cli-plugins/docker-dev
  extension: Manages Docker extensions (Docker Inc.)
    Version:  v0.2.23
    Path:     /usr/local/lib/docker/cli-plugins/docker-extension
  feedback: Provide feedback, right in your terminal! (Docker Inc.)
    Version:  v1.0.4
    Path:     /usr/local/lib/docker/cli-plugins/docker-feedback
  init: Creates Docker-related starter files for your project (Docker Inc.)
    Version:  v1.1.0
    Path:     /usr/local/lib/docker/cli-plugins/docker-init
  sbom: View the packaged-based Software Bill Of Materials (SBOM) for an image (Anchore Inc.)
    Version:  0.6.0
    Path:     /usr/local/lib/docker/cli-plugins/docker-sbom
  scout: Docker Scout (Docker Inc.)
    Version:  v1.6.4
    Path:     /home/alves/.docker/cli-plugins/docker-scout
WARNING: Plugin "/usr/local/lib/docker/cli-plugins/docker-scan" is not valid: failed to fetch metadata: fork/exec /usr/local/lib/docker/cli-plugins/docker-scan: no such file or directory

Server:
 Containers: 0
  Running: 0
  Paused: 0
  Stopped: 0
 Images: 0
 Server Version: 26.0.0
 Storage Driver: overlay2
  Backing Filesystem: extfs
[...]

PR #11725, opened 2024-04-15
Fix progress timer and spinner

ndeloof:
**What I did**
https://github.com/docker/compose/pull/11357 was way too optimistic: we can't guess the last event expected for a resource. Better allow progress to manage "more work coming in" and restart the timer/spinner.

progress will report "done" state for some milliseconds before a new task starts and switch it back to "working", so there might be some flicker in UI but at least the end result is correct.

**Related issue**
fixes https://github.com/docker/compose/issues/11717

image


glours (approved):
LGTM, tested and I confirm it works way better this way 👍

PR #11738, opened 2024-04-17
fix: do not try to create file shares for non-directories

milas:
**What I did**
When creating automatic file shares, ignore any non-directory bind mounts, e.g. for an individual (normal) file or a special type like a Unix socket (`/var/run/docker.sock`).

Additionally, there's no need to create a directory if one does not exist, the API will handle that. However, the check for existence remains so that `create_host_path: false` mounts can be ignored.

**Related issue**
https://docker.atlassian.net/browse/COMP-515

black squirrel


jhrotko (approved):
LGTM

milas:
CI failures are because `TestCascadeFail` is flaky right now, fix @ #11739

PR #11739, opened 2024-04-17
chore(e2e): fix flaky cascade failure test

milas:
**What I did**
This was racy with the sleep, so the Compose file has been tweaked to make it pass reliably.

Now, there's 3 services:
 * `running` - sleeps forever
 * `exit` - exits _successfully_ immediately
   * depends on `running` started
 * `fail` - exits _with error_ immediately
   * depends on `exit` succeeding

Now, the test can ensure that the containers are all run/ started in the expected order the assertions will be reliable.

Before, it was possible for `fail` to run & exit before `exit`, for example. The `running` service also ensures there's always at least one other "running" container when we do an abort.

**Related issue**
n/a, just my failing CI jobs

baby cheetah


jhrotko (approved):
LGTM

jhrotko:
> ```
> cascade_test.go:50: Running command: /home/runner/work/compose/compose/bin/build/docker-compose -f ./fixtures/cascade/compose.yaml --project-name compose-e2e-cascade-fail up --abort-on-container-failure
> Attaching to exit-1, fail-1, running-1
> exit-1 exited with code 0
> fail-1 exited with code 111
>     cascade_test.go:55: assertion failed: 143 (res.ExitCode int) != 111 (int)
> ```
> 
> 🤔

Screenshot 2024-04-18 at 16 59 01
🤔

glours (approved):

PR #11748, opened 2024-04-19
fix(desktop): remove overly-aggressive feature flag check

milas:
**What I did**
Synchronized file share integration between Desktop & Compose is still experimental, so there's a feature flag. However, it's slightly more complex than that - not all OS/hypervisor combos support Synchronized file shares and it's an entitled feature, so the config API from Desktop covers all this logic.

I accidentally left the old feature flag check in, but it's possible to opt-in via settings independent of the feature flag, which wasn't being respected. (That is, it's possible to enable the feature w/o the feature flag being on because of weirdness about how feature flags & settings work in Desktop currently.)

**Related issue**
https://docker.atlassian.net/browse/COMP-533

tortoise shell cat camouflaged against a similar patterned wall

### DIFF
--- a/internal/experimental/experimental.go
+++ b/internal/experimental/experimental.go
@@ -71,10 +71,6 @@ func (s *State) NavBar() bool {
 	return s.determineFeatureState("ComposeNav")
 }
 
-func (s *State) AutoFileShares() bool {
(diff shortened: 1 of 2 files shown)


glours (approved):
LGTM

Issue #11749, opened 2024-04-19
[BUG] Inconsistent behavior with `--context` (and likely other shared Docker CLI) flags


milas:
### Description

When running Compose in standalone mode (invoking `docker-compose` directly), the `--context` flag only works if specified as `--context foo` but not `--context=foo`, which is inconsistent with our general CLI args behavior, and returns a confusing `unknown flag: --context` error messsage

### Steps To Reproduce

Run Compose with `--context` flag directly, e.g. `~/.docker/cli-plugins/docker-compose --context=foo ls` on macOS+DD

Must use the `key=val` variant, e.g. `--context=foo` NOT `--context foo`, which does work as expected.

### Compose Version

```Text
Docker Compose version v2.26.1-desktop.1

(also repro'd back as far as v2.23.3 and probably older)
```

### Docker Environment

```Text
Client:
 Version:    26.0.1
 Context:    dev-mode
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.13.1-desktop.1
    Path:     /Users/milas/.pinata/downloads/docker-plugins/v0.13.1-desktop.1/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.26.1-desktop.1
    Path:     /Users/milas/.docker/cli-plugins/docker-compose
  debug: Get a shell into any image or container (Docker Inc.)
    Version:  0.0.29
    Path:     /Users/milas/.docker/cli-plugins/docker-debug
  dev: Docker Dev Environments (Docker Inc.)
    Version:  v0.1.2
    Path:     /Users/milas/.docker/cli-plugins/docker-dev
  extension: Manages Docker extensions (Docker Inc.)
    Version:  v0.2.23
    Path:     /Users/milas/.docker/cli-plugins/docker-extension
  feedback: Provide feedback, right in your terminal! (Docker Inc.)
    Version:  v1.0.4
    Path:     /Users/milas/.docker/cli-plugins/docker-feedback
  init: Creates Docker-related starter files for your project (Docker Inc.)
    Version:  v1.1.0
    Path:     /Users/milas/.docker/cli-plugins/docker-init
  sbom: View the packaged-based Software Bill Of Materials (SBOM) for an image (Anchore Inc.)
    Version:  0.6.0
    Path:     /Users/milas/.docker/cli-plugins/docker-sbom
  scout: Docker Scout (Docker Inc.)
    Version:  v1.7.0
[...]

PR #11752, opened 2024-04-22
fix support for `--flag=value` syntax in compatibility mode

ndeloof:
**What I did**
fix support for `--flag=value` syntax in compatibility mode

**Related issue**
fixes https://github.com/docker/compose/issues/11749

image


glours (approved):
LGTM

Issue #11760, opened 2024-04-23
Release with all vendor dependencies


catap:
### Description

Good day,

Is it possible to add to release a distirbutive which contains all vendor dependencies?

It should make life of package managers much easier.

Thanks!

ndeloof:
Can't you just rely on `go.mod` for this purpose?
What's the challenge you have to address as a package manager?

catap:
Yes, I can. Right now I need to pack vedors folder as dedicated archive to make an offline build. Without it build of `docker-compose` needs an internet to download everything.

The issue that usual workflow for build a pachage is:
 - fetch everything
 - extract
 - patch when needed
 - configure
 - build

`go mod` nice solution but it should be run after extract that means fetch stage isn't complete, and if other stages goes without internet connection, build fails.

Just an observation: `docker-buildx` and `docker-cli` provides such archive.

ndeloof:
both buildx and docker cli projects adopted a `vendor` folder inside repository, which we don't. This is a long debate, but compose maintainers don't consider this to be a good practice. the source zip archive produced by github during release from tag only contains a git shapshot, so is obviously missing such a `vendor` folder. If you have a requirement to build offline, then indeed you'll need your pipeline to include an additional `do mod download` step after extraction

catap:
Right now I host such artifacts on my host for OpenBSD artifacts for example, see: https://github.com/openbsd/ports/blob/master/sysutils/docker-compose/Makefile

Maybe it isn't so bad idea to prepare one more release artifcat via githib actions? I don't ask to include vendor into sources, just make one more artifact which I may use as source for future updates.

ndeloof:
right, feel free to offer us a pull request for this purpose :)

catap:
Seems that it will be won't fix

PR #11764, opened 2024-04-24
Build vendor artifact

catap:
Closes: https://github.com/docker/compose/issues/11760

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -249,6 +249,34 @@ jobs:
         with:
           files: ./coverage.txt
 
+  vendor:
+    if: github.event_name != 'pull_request'
+    runs-on: ubuntu-latest
+    steps:
+      -
+        name: Checkout
+        uses: actions/checkout@v3
+      -
+        name: Set up Go
+        uses: actions/setup-go@v3
+        with:
+          go-version-file: 'go.mod'
+          check-latest: true
+          cache: true
+      -
+        name: Build vendor
+        run: |
+          go mod vendor
+          mkdir -p bin/release
+          tar czf bin/release/docker-compose-vendor.tar.gz vendor
+      -
+        name: Upload artifacts
+        uses: actions/upload-artifact@v3
+        with:
+          name: compose
+          path: ./bin/release/*
+          if-no-files-found: error
+
   release:
     permissions:
       contents: write # to create a release (ncipollo/release-action)


catap:
@ndeloof here it is

.github/workflows/ci.yml
```diff
@@ -249,6 +249,34 @@ jobs:
...
+        name: Set up Go
+        uses: actions/setup-go@v3
+        with:
+          go-version-file: 'go.mod'
+          check-latest: true
+          cache: true
+      -
+        name: Build vendor
+        run: |
+          go mod tidy
```
ndeloof, line 268: `tidy` is not relevant here, as we enforce go.sum to be up-to-date as part of the PR workflow

catap: Fixed

catap:
> LGTM but should be declared in the `merge` workflow so this run for commits on main branch and tags

`ci` workflow is triggered for by push into `main` brnach and tags as well. Also, this workflow makes all artifacts and it seems for me logical that it should be used to produce this one.

Have I missed something?

ndeloof:
`ci` workflow is triggered by push to `main` but also by pull_requests, and this doesn't make sense to build this src archive for PRs
`merge` workflow is used for stuff to only happen on main branch and releases, like building bin image

catap:
@thaJeztah this is one of root cause why I'm doing this.

Right now OpenBSD port for `docker-compose` depends on `.tar.gz` which I build and host on my machine. Everyone should trust that I haven't been compromised and doesn't inject something very bad into this  `.tar.gz` and has we seen such injection can be quite complicated.

So, the right way to do it is download from the same source where it sources is produced becuase this should be trusted source by defenition.

Regarding the way: as soon as I know `go mod vendor` respects checksums inside `go.mod` that should prevent from packing bad things. Or, let me say it that way: it should pack the same things which user should get if he run it locally.

catap:
> `ci` workflow is triggered by push to `main` but also by pull_requests, and this doesn't make sense to build this src archive for PRs `merge` workflow is used for stuff to only happen on main branch and releases, like building bin image

To keep things simpler I've added if inside this step.

Otherwise I need to move and duplicate code inside `merge` workflow which seems wrong from my point of view.

ndeloof (approved):

thaJeztah:
> Right now OpenBSD port for docker-compose depends on .tar.gz which I build and host on my machine. Everyone should trust that I haven't been compromised and doesn't inject something very bad into this  .tar.gz and has we seen such injection can be quite complicated.

Yes, anyone producing binaries / artifacts for others to consume must have provenance. I think that goes beyond just these tars, and there would be many other ways for that build-environment / infra to be comprimised.

> Regarding the way: as soon as I know go mod vendor respects checksums inside go.mod that should prevent from packing bad things. Or, let me say it that way: it should pack the same things which user should get if he run it locally.

Correct; it won't protect against a compromised tarball that has `go.mod` and `go.sum` tampered with though; so I'm looking if there's any way to have provenance for the uploaded tarball to verify if it's produced by the GitHub action, and not replaced in other ways.

The default tarballs produced by GitHub always match the git-commit of the release, but for other artifacts that's not the case, so some validation may be necessary for that.

catap:
> > Regarding the way: as soon as I know go mod vendor respects checksums inside go.mod that should prevent from packing bad things. Or, let me say it that way: it should pack the same things which user should get if he run it locally.
> 
> Correct; it won't protect against a compromised tarball that has `go.mod` and `go.sum` tampered with though; so I'm looking if there's any way to have provenance for the uploaded tarball to verify if it's produced by the GitHub action, and not replaced in other ways.
> 
> The default tarballs produced by GitHub always match the git-commit of the release, but for other artifacts that's not the case, so some validation may be necessary for that.

I haven't see any differences with producing `.tar.gz` by this job with producing binary artifacts by the same github actions.

Second point: GitHub and other git hosting sometimes *changes* produced `tar.gz` from tag and they officialy confirm it, see: https://github.blog/2023-02-21-update-on-the-future-stability-of-source-code-archives-and-hashes/ and to avoid such mess they suggested to attach artifacts to releases.

catap:
just a ping

.github/workflows/ci.yml
```diff
@@ -249,6 +249,34 @@ jobs:
...
+        name: Build vendor
+        run: |
+          go mod vendor
+          mkdir -p bin/release
+          tar czf bin/release/docker-compose-vendor.tar.gz vendor
+      -
+        name: Upload artifacts
+        uses: actions/upload-artifact@v3
+        with:
+          name: compose
```
gabriellavengeo, line 276: I see that the artifact always has the same name. The behaviour of this depends on the `overwrite` value. 

```
    # If true, an artifact with a matching name will be deleted before a new one is uploaded.
    # If false, the action will fail if an artifact for the given name already exists.
    # Does not fail if the artifact does not exist.
    # Optional. Default is 'false'
    overwrite:
```

This will have the default `overwrite` value of `false` so it will always fail after the first time, if I understand this correctly.

.github/workflows/ci.yml
```diff
@@ -249,6 +249,34 @@ jobs:
...
+          cache: true
+      -
+        name: Build vendor
+        run: |
+          go mod vendor
+          mkdir -p bin/release
+          tar czf bin/release/docker-compose-vendor.tar.gz vendor
+      -
+        name: Upload artifacts
+        uses: actions/upload-artifact@v3
```
gabriellavengeo, line 274: Is there a reason that we use `actions/upload-artifact@v3` instead of `v4`? According to the [action README](https://github.com/actions/upload-artifact?tab=readme-ov-file#actionsupload-artifact), `v3` will be deprecated in November.

gabriellavengeo: Speaking of `v4`, some of the breaking changed in it relate to my other comment bellow about the aritfact name.

```
2. Uploading to the same named Artifact multiple times.

Due to how Artifacts are created in this new version, it is no longer possible to upload to the same named Artifact multiple times. You must either split the uploads into multiple Artifacts with different names, or only upload once. Otherwise you will encounter an error.
```

ndeloof: aritfact name must be unique for a built commit, I don't think this has any impact for this workflow.

We still use upload-artifact@v3 on this repo as this breaking change require the whole workflow to be revisited to support the new behavior

ndeloof:
Can you look into https://docs.github.com/en/actions/security-guides/using-artifact-attestations-to-establish-provenance-for-builds#generating-build-provenance-for-binaries so this additional artifact gets provenance attestation and you can safely rely on this for an offline build ?

.github/workflows/ci.yml
gabriellavengeo: > aritfact name must be unique for a built commit, I don't think this has any impact for this workflow.

Ah ok, then this should be fine.

catap:
On Mon, 06 May 2024 13:54:21 +0100,
Nicolas De loof ***@***.***> wrote:
> 
> Can you look into https://docs.github.com/en/actions/security-guides/using-artifact-attestations-to-establish-provenance-for-builds#generating-build-provenance-for-binaries so this additional artifact gets provenance attestation and you can safely rely on this for an offline build ?
> 

Thanks for the link.

I've read it and I doesn't see how it can help me.

OpenBSD ports, for example, ships its own checksum of artifacts which should
be used to build port.

It is checked by build servers which downloads artifacts from github.com or
other plases and confirms that checksum holds.

-- 
wbr, Kirill

PR #11765, opened 2024-04-24
Fix menu print when logs/menu number of lines change

jhrotko:
**What I did**
The is related with the clean up of the navigation menu which can be tricky when the number of lines change.
Removed all ClearLines and kept the cleaning of the menu inside shortcut.go. Please check the video.

https://github.com/docker/compose/assets/22371565/8744d6d1-c9a8-44a4-ac9b-517e8feae22a

**Related issue**

fixes [#14021](https://github.com/docker/for-win/issues/14021)
https://docker.atlassian.net/jira/software/c/projects/COMP/boards/576?selectedIssue=COMP-536

image

### DIFF
--- a/cmd/compose/up.go
+++ b/cmd/compose/up.go
@@ -32,6 +32,7 @@ import (
 	"github.com/spf13/cobra"
 
 	"github.com/docker/compose/v2/pkg/api"
+	ui "github.com/docker/compose/v2/pkg/progress"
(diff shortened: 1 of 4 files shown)


milas (approved):
Nice! This looks great, I can't repro any weird issues with it overlapping or appearing when it shouldn't after this.

PR #11800, opened 2024-05-07
Add documentation for --menu up option and COMPOSE_MENU environment var

jhrotko:
**What I did**
Added documentation related to recent added option in `up` command to enable/disable the helper menu in detach mode.
Seems like this information was lacking in our documentation and some users were confused on how to enable/disable the menu.

**Related issue**
relates to https://github.com/docker/for-win/issues/14021
https://docker.atlassian.net/browse/COMP-540

image

### DIFF
--- a/docs/reference/docker_compose.yaml
+++ b/docs/reference/docker_compose.yaml
@@ -118,6 +118,9 @@ long: |-
     Setting the `COMPOSE_IGNORE_ORPHANS` environment variable to `true` stops docker compose from detecting orphaned
     containers for the project.
 
+    Setting the `COMPOSE_MENU` environment variable to `false` disables the helper menu when running `docker compose up`
+    in attached mode. Alternatively, you can also run `docker compose up --menu=false` to disable the helper menu.
+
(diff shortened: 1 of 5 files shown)


docs/reference/docker_compose.yaml
```diff
@@ -118,6 +118,9 @@ long: |-
     Setting the `COMPOSE_IGNORE_ORPHANS` environment variable to `true` stops docker compose from detecting orphaned
     containers for the project.
 
+    Setting the `COMPOSE_MENU` environment variable to `false` disabled the helper menu when running `docker compose up`
+    in detached mode. Alternatively, you can also run `docker compose up --menu=false` to disable the helper menu.
```
milas, line 122:
```suggestion
Setting the `COMPOSE_MENU` environment variable to `false` disables the helper menu when running `docker compose up`
in attached mode. Alternatively, you can also run `docker compose up --menu=false` to disable the helper menu.
```

milas (approved):
Tiny copy fixes but LGTM

PR #11815, opened 2024-05-14
Per endpoint driver opts

robmry:
**What I did**

It's possible to set driver options when creating a network using compose ... [compose spec](https://github.com/robmry/compose-go/blob/b08e2159a8b1fd78444c2ef434bd4913d74697d1/schema/compose-spec.json#L637-L641) / [engine API](https://github.com/moby/moby/blob/9d07820b221db010bf1bdc26ca904468804ca712/api/swagger.yaml#L10134-L10138).

But not when connecting an endpoint to a network ... [compose spec](https://github.com/robmry/compose-go/blob/b08e2159a8b1fd78444c2ef434bd4913d74697d1/schema/compose-spec.json#L293-L300) / [engine API](https://github.com/moby/moby/blob/9d07820b221db010bf1bdc26ca904468804ca712/api/swagger.yaml#L2541-L2544).

The API field isn't new, but upcoming change https://github.com/moby/moby/pull/47686 means it'll be useful for setting per-interface sysctls.

https://github.com/compose-spec/compose-go/pull/627 adds the types.
https://github.com/compose-spec/compose-spec/pull/498 updates the spec to match.
This PR pulls in the `go-compose` changes, and copies the driver options to `EndpointSettings` for the API.

_I noticed a couple of fields in `EndpointSettings` that perhaps shouldn't be set ... the struct is an unfortunate mix of config and operational-data (used to report running state). Those fields will be cleared by the daemon, so the only potential problem is a bit of confusion - in particular, the container's IPv6 address is provided in the IPv6 gateway. But, probably best to deal with that separately, so I just left a FIXME comment in the new test._ 

**Related issue**

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,7 +7,7 @@ require (
 	github.com/Microsoft/go-winio v0.6.2
 	github.com/acarl005/stripansi v0.0.0-20180116102854-5a71ef0e047d
 	github.com/buger/goterm v1.0.4
-	github.com/compose-spec/compose-go/v2 v2.1.1-0.20240516145816-197e6352c64a
+	github.com/compose-spec/compose-go/v2 v2.1.1-0.20240514125539-94bd7356c641
 	github.com/containerd/console v1.0.4
 	github.com/containerd/containerd v1.7.16
 	github.com/davecgh/go-spew v1.1.1
--- a/go.mod
+++ b/go.mod
@@ -7,7 +7,7 @@ require (
 	github.com/Microsoft/go-winio v0.6.2
 	github.com/acarl005/stripansi v0.0.0-20180116102854-5a71ef0e047d
 	github.com/buger/goterm v1.0.4
-	github.com/compose-spec/compose-go/v2 v2.1.1-0.20240514125539-94bd7356c641
+	github.com/compose-spec/compose-go/v2 v2.1.1
 	github.com/containerd/console v1.0.4
 	github.com/containerd/containerd v1.7.16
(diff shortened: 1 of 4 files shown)


robmry:
I don't think the failing test looks related, is it just something that happens? ...

```
FAIL pkg/e2e.TestCascadeFail (2.19s)
      cascade_test.go:44: WARNING: docker-buildx cli-plugin not found, using default buildx installation.
      cascade_test.go:44: Running command: docker compose version
  Docker Compose version 0eef47d
      cascade_test.go:50: Running command: docker compose -f ./fixtures/cascade/compose.yaml --project-name compose-e2e-cascade-fail up --abort-on-container-failure
  Attaching to exit-1, fail-1, running-1
  
  exit-1 exited with code 0
  
  fail-1 exited with code 111
      cascade_test.go:55: assertion failed: 143 (res.ExitCode int) != 111 (int)
      cascade_test.go:47: Running command: docker compose --project-name compose-e2e-cascade-fail down
      framework.go:130: Contents of config dir:
      framework.go:132:   - /tmp/TestCascadeFail1299123561/001
      framework.go:132:   - /tmp/TestCascadeFail1299123561/001/cli-plugins
      framework.go:132:   - /tmp/TestCascadeFail1299123561/001/cli-plugins/docker-compose
      framework.go:132:   - /tmp/TestCascadeFail1299123561/001/cli-plugins/docker-scan
```

ndeloof (approved):

Issue #11823, opened 2024-05-15
[BUG] Regressions in .env loading


jpetazzo:
### Description

The way `.env` files are loaded has changed twice recently.

Let's assume the following project structure:
```
$ tree -a
.
├── .env
├── compose.yaml
└── sub 
    └── .env
```

When running `docker compose` in the `sub/` directory, ...
- up to Compose 2.23.3, Compose merges both `.env` files, with the top-level `.env` file taking precedence over the sub-level one
- from Compose 2.24.0 to 2.24.5, Compose only uses the `.env` file in the current directory, and ignore the one at the top level
- from Compose 2.24.6 to at least 2.27.0, Compose only uses the `.env` file at the top-level directory, and ignores the one in the sub directory

Both changes broke some folks' workflows in different ways (see #11531 and #11575). The second change might be related to #11405 but I haven't looked at the code, so I might be wrong.

 

### Steps To Reproduce

```
cd /tmp
mkdir -p envtest envtest/sub

cat >envtest/compose.yaml << "EOF"
services:
  echo:
    image: alpine
    command: echo $ROOT $SUB win=$WIN
EOF

cat >envtest/.env << EOF
ROOT=root
WIN=root
EOF

cat >envtest/sub/.env << EOF
SUB=sub
WIN=sub
EOF

cd envtest/sub
docker compose run echo
```

Here is the output with various versions of Compose...

Up to Compose 2.23.3:
```
root sub win=root
```

From Compose 2.24 to 2.24.5:
```
WARN[0000] The "ROOT" variable is not set. Defaulting to a blank string.
sub win=sub
```

From Compose 2.24.6:
```
WARN[0000] The "SUB" variable is not set. Defaulting to a blank string.
root win=root
```

### Compose Version

```Text
N/A
```

### Docker Environment

```Text
N/A
```

### Anything else?

The ability of loading a `.env` file from the current directory (instead, or in addition to, the top-level `.env` file) is very powerful.

Example user story: I develop an application stack made of 2 services, `service1` and `service2`. I sell that stack as a SAAS to multiple customers. Each customer gets their own stack, with their own instance of each service. I want to deploy each stack with Compose and follow "DRY" principles: I want to avoid repeating or duplicating code as much as possible.

I can achieve that with the following structure:
```
.
├── compose.yaml
├── customers
│   ├── customer1
│   │   └── .env
│   ├── customer1
│   │   └── .env
│   └── customer3
│       └── .env
├── service1
│   └── Dockerfile
[...]

Issue #11575, opened 2024-03-05
[BUG] docker compose config regression - traverses parent dirs before applying `.env`


MikaelElkiaer:
### Description

Since v2.24.4 (possibly before) some behaviour has changed in how `docker compose` resolves configuration.
It used to apply `.env` located in `cwd`, before traversing parent dirs to look for a `docker-compose.yaml`.
This has changed, as it will now traverse parent dirs and if it finds a `docker-compose.yaml` it will instead look for a `.env` in that directory.

### Steps To Reproduce

1. With files:

```yaml
# ./docker-compose.yaml
services:
  parent-service:
    image: alpine
```

```yaml
# ./root/subdir/docker-compose.yaml
services:
  subdir-service:
    image: alpine
```

```sh
# ./root/.env
COMPOSE_FILE=./subdir/docker-compose.yaml
```

2. with cwd `./root`, run `docker compose config` - it used to be the config based on the subdir, but now it is based on the parent dir.

### Compose Version

```Text
❯ docker compose version
Docker Compose version 2.24.6
❯ docker-compose version
Docker Compose version 2.24.6
```

### Docker Environment

```Text
❯ docker info
Client:
 Version:    25.0.3
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  0.12.1
    Path:     /usr/lib/docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  2.24.6
    Path:     /usr/lib/docker/cli-plugins/docker-compose

Server:
 Containers: 1
  Running: 0
  Paused: 0
  Stopped: 1
 Images: 168
 Server Version: 25.0.3
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: true
  Native Overlay Diff: false
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 7c3aca7a610df76212171d200ca3811ff6096eb8.m
 runc version:
 init version: de40ad0
 Security Options:
  seccomp
   Profile: builtin
[...]

PR #11824, opened 2024-05-16
allow a local .env file to override compose.yaml sibling .env

ndeloof:
**What I did**
Allow local .env file both to set COMPOSE_* variables and override .env file aside the target compose.yaml file

This restores behavior seen in compose v2 before v2.24.x, while undocumented. Was initially introduced by https://github.com/docker/compose/pull/9512 while not designed for this usage, but users started then to rely on this "feature"

A doc PR will follow so this is well documented

**Related issue**
fixes https://github.com/docker/compose/issues/11823
fixes https://github.com/docker/compose/issues/11575

https://docker.atlassian.net/browse/COMP-570


jhrotko (approved):
LGTM

jhrotko:
> This restores behavior seen in compose v2 before v2.24.x, while undocumented.
Did you change the documentation to make it more clear? I remember you have mentioned it before just asking the current status

ndeloof:
@jhrotko I'll do once this PR is merged

PR #11832, opened 2024-05-20
Backport OpenBSD patches

catap:
**What I did**

Backport a trivial patch whcih I've used inside OpenBSD port to build it

**Related issue**

Note that I aware of


glours (approved):
LGTM

PR #11845, opened 2024-05-23
fix opentelemetry

ndeloof:
**What I did**
Do **not** capture `Tracer` in a var as `TracerProvider` get updated (twice) during the command PreRun

**Related issue**
https://docker.atlassian.net/browse/COMP-584


glours (approved):
LGTM

PR #11846, opened 2024-05-23
WIP fix tracing

laurazard:
**What I did**

**Related issue**

### DIFF
--- a/cmd/cmdtrace/cmd_span.go
+++ b/cmd/cmdtrace/cmd_span.go
@@ -20,7 +20,9 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"os"
 	"sort"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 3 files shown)


cmd/cmdtrace/cmd_span.go
```diff
@@ -44,13 +46,17 @@ import (
...
+	}
+
 	tracingShutdown, err := tracing.InitTracing(dockerCli)
 	if err != nil {
 		return fmt.Errorf("initializing tracing: %w", err)
 	}
 
 	ctx := cmd.Context()
-	ctx, cmdSpan := tracing.Tracer.Start(
+	ctx, cmdSpan := tracing.TracingProvider.Tracer("compose").Start(
```
ndeloof, line 59: must not set Tracer name to "compose"

> 	// The passed name needs to uniquely identify instrumentation code.
>	// Therefore, it is recommended that name is the Go package name of the
>	// library providing instrumentation (note: not the code being
>	// instrumented)

yes, this is crazy and terribly confusing, instrumented code is identified by `service.name` attribute, ... which we set in compose to track compose service name 😓, need to be fixed

laurazard: 😭

Issue #11857, opened 2024-05-28
[BUG] `Watch` crashes when sync folders overlap


korverdev:
### Description

`docker compose watch` crashes when `sync` actions for multiple containers use the same `path`. This is the config I use for three separate containers which all need access to the same source code.

```
watch:
        - action: sync
          path: ./src
          target: /home/user/adapt/src/
```

### Steps To Reproduce

I haven't been able to consistently reproduce this behavior, but by running `docker compose watch` when the conditions described above are met and working as normally, it will crash basically every day from a full work day. In this example, I hit an error immediately, but that's uncommon. The error only occurs on file sync, so repeatedly changing a file should cause it to happen quickly.

### Compose Version

```Text
Docker Compose version v2.27.0-desktop.2
```

### Docker Environment

```Text
Client:
 Version:    26.1.1
 Context:    desktop-linux
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.14.0-desktop.1
    Path:     /Users/casey/.docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.27.0-desktop.2
    Path:     /Users/casey/.docker/cli-plugins/docker-compose
  debug: Get a shell into any image or container (Docker Inc.)
    Version:  0.0.29
    Path:     /Users/casey/.docker/cli-plugins/docker-debug
  dev: Docker Dev Environments (Docker Inc.)
    Version:  v0.1.2
    Path:     /Users/casey/.docker/cli-plugins/docker-dev
  extension: Manages Docker extensions (Docker Inc.)
    Version:  v0.2.23
    Path:     /Users/casey/.docker/cli-plugins/docker-extension
  feedback: Provide feedback, right in your terminal! (Docker Inc.)
    Version:  v1.0.4
    Path:     /Users/casey/.docker/cli-plugins/docker-feedback
  init: Creates Docker-related starter files for your project (Docker Inc.)
    Version:  v1.1.0
    Path:     /Users/casey/.docker/cli-plugins/docker-init
  sbom: View the packaged-based Software Bill Of Materials (SBOM) for an image (Anchore Inc.)
    Version:  0.6.0
    Path:     /Users/casey/.docker/cli-plugins/docker-sbom
  scout: Docker Scout (Docker Inc.)
    Version:  v1.8.0
    Path:     /Users/casey/.docker/cli-plugins/docker-scout

Server:
 Containers: 7
[...]

PR #11860, opened 2024-05-29
prevent concurrent map write relying on project immutability

ndeloof:
**What I did**
Use `WithServicesTransform` to transform project, in respect with immutability to prevent concurrent map write

**Related issue**
fixes https://github.com/docker/compose/issues/11857
https://docker.atlassian.net/browse/COMP-590

image


glours (approved):
LGTM

PR #11862, opened 2024-05-29
Fix: change append to use slice index in ps

### LINKED ISSUE #11858: [BUG] Using append instead of slice index
### Description

is ps.go we have this piece of code:
`var publishers []api.PortPublisher
	sort.Slice(container.Ports, func(i, j int) bool {
		return container.Ports[i].PrivatePort < container.Ports[j].PrivatePort
	})
	for _, p := range container.Ports {
		publishers = append(publishers, api.PortPublisher{
			URL:           p.IP,
			TargetPort:    int(p.PrivatePort),
			PublishedPort: int(p.PublicPort),
			Protocol:      p.Type,
		})
	}`

which should probably use the length of the slice we are taking values from to reduce need to create backing arrays when slices become too long.  Should also use the index for the for loop to avoid a small overhead from the append function.  I think it would be something like this: 

`publishers := make([]api.PortPublisher, len(container.Ports)
	sort.Slice(container.Ports, func(i, j int) bool {
		return container.Ports[i].PrivatePort < container.Ports[j].PrivatePort
	})
	for i, p := range container.Ports {
		publishers[i] = publishers, api.PortPublisher{
			URL:           p.IP,
			TargetPort:    int(p.PrivatePort),
			PublishedPort: int(p.PublicPort),
			Protocol:      p.Type,
		}
	}`

### Steps To Reproduce

_No response_

### Compose Version

_No response_

### Docker Environment

_No response_

### Anything else?

_No response_

IDOMATH:
**What I did**
Changed slice appending to use index to remove overhead from using build in append function.
Also made the slice using the length of slice we are copying values from to remove overhead of making new backing arrays.

**Related issue**
Fixes #11858 
https://github.com/docker/compose/issues/11858

### DIFF
--- a/pkg/compose/ps.go
+++ b/pkg/compose/ps.go
@@ -45,17 +45,17 @@ func (s *composeService) Ps(ctx context.Context, projectName string, options api
 	for i, container := range containers {
 		i, container := i, container
 		eg.Go(func() error {
-			var publishers []api.PortPublisher
+			publishers := make([]api.PortPublisher, len(container.Ports))
 			sort.Slice(container.Ports, func(i, j int) bool {
 				return container.Ports[i].PrivatePort < container.Ports[j].PrivatePort
 			})
-			for _, p := range container.Ports {
-				publishers = append(publishers, api.PortPublisher{
+			for i, p := range container.Ports {
+				publishers[i] = api.PortPublisher{
 					URL:           p.IP,
 					TargetPort:    int(p.PrivatePort),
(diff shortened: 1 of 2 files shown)


ndeloof:
For legal reasons we require all contributor to sign off commits. Please amend your commit with `git amend --signoff` and force-push your branch

ndeloof (approved):

IDOMATH:
I have added my sign off.

Issue #11867, opened 2024-05-30
Support remote DOCKER_HOST by copying configs and secrets files instead of bind mounting them


andoks:
### Description

Per title, currently docker compose uses bind mounts to put configs and secrets in containers, but this does not work when using either docker context or DOCKER_HOST variable to remotely deploy to a different system.

This would help some of the use cases described in these issues:
- #7857 
- #11141
- #5523
- #7110

It builds upon the config content and environment support added in https://github.com/compose-spec/compose-spec/issues/346

andoks:
A suggestion for how to fix this, see the following attached patch that should apply to main branch
[0001-Enable-config.file-s-on-remote-docker-hosts.patch.gz](https://github.com/user-attachments/files/15506646/0001-Enable-config.file-s-on-remote-docker-hosts.patch.gz) (had to gzip the file, as it would not upload as a plain .patch file)

andoks:
Two alternatives if this change in existing behaviour is undesireable, could be to make the config.driver = "copy" change to this new behaviour, or support a new config.copy property for copying the file into the container/service.

@ndeloof hope it is ok I am tagging you here, as I saw you were involved in the https://github.com/compose-spec/compose-spec/issues/346 issue

ndeloof:
I fully agree with this direction.
Regarding change in behavior, I don't feel like this is an issue as long as this is documented in release note.
A trivial workaround for those who require dynamic sync with container would be to declare an additional bind mount or to rely on a dedicated `watch` trigger to sync changes

ndeloof:
> see the following attached patch

Please open a pull-request

ndeloof:
related: https://github.com/docker/compose/issues/9648#issuecomment-1557068598

andoks:
> > see the following attached patch
> 
> Please open a pull-request

@ndeloof: done -> https://github.com/docker/compose/pull/11871

LawrenceWarren:
This is a good feature that lends itself to a GitOps approach to remote Docker Compose deployments. However I think this should be a toggle, especially with secrets. 

For example, in my workflow I have various `docker-compose.yml`s each in different directories, with each directory corresponding to a remote server. As a very basic example:

```tree
infrastructure-as-code
├── dev
│   └── instances
│       ├── dev-server-1
│       │   └── docker-compose.yml
│       └── dev-server-2
│           └── docker-compose.yml
└── prod
    └── instances
        ├── prod-server-1
        │   └── docker-compose.yml
[...]

PR #11871, opened 2024-05-31
Enable configs.file's on remote docker hosts

andoks:
**What I did**
Copy configs.file's instead of bind-mounting them to make it possible to use file configs when working with remote docker hosts (like setting DOCKER_HOST to a ssh address or setting docker context)

**Related issue**

implements: #11867

image


pkg/compose/secrets.go
```diff
@@ -31,7 +32,25 @@ import (
 func (s *composeService) injectSecrets(ctx context.Context, project *types.Project, service types.ServiceConfig, id string) error {
 	for _, config := range service.Secrets {
 		file := project.Secrets[config.Source]
-		if file.Environment == "" {
+		var content = file.Content
+
+		switch {
+		case file.File != "":
+			buf, err := os.ReadFile(file.File)
```
ndeloof, line 39: `secret.file` could also be a folder, not just a file

pkg/compose/secrets.go
```diff
@@ -64,13 +79,24 @@ func (s *composeService) injectConfigs(ctx context.Context, project *types.Proje
 	for _, config := range service.Configs {
 		file := project.Configs[config.Source]
 		content := file.Content
-		if file.Environment != "" {
+		switch {
+		case file.File != "":
+			fmt.Sprintf("test")
```
andoks, line 84: remove stray test output

pkg/compose/secrets.go
andoks: config.file too?

ndeloof: indeed

pkg/compose/secrets.go
```diff
@@ -30,27 +34,36 @@ import (
...
 			config.Target = "/run/secrets/" + config.Target
 		}
 
-		env, ok := project.Environment[file.Environment]
-		if !ok {
-			return fmt.Errorf("environment variable %q required by file %q is not set", file.Environment, file.Name)
+		file := project.Secrets[config.Source]
+		var tarArchive bytes.Buffer
+		var err error
+		switch {
```
andoks, line 46: Note: not handling file.Content as this seems to be rejected some other place, causing the printout "validating docker-compose.yml: secrets.content_secret Additional property content is not allowed"

pkg/compose/secrets.go
andoks: Implemented in new version

pkg/compose/secrets.go
```diff
@@ -94,47 +114,135 @@ func (s *composeService) injectConfigs(ctx context.Context, project *types.Proje
...
+	err = tarWriter.Close()
+	if err != nil {
+		return bytes.Buffer{}, fmt.Errorf("failed closing tar archive after writing, error was: %w", err)
+	}
+
+	return b, err
+}
+
+func makeTarFileEntryParams(config types.FileReferenceConfig) (mode int64, uid, gid int, err error) {
+	mode = 0o444
```
andoks, line 227: :question: should the default mode be different for secrets compared to configs?

pkg/compose/secrets.go
```diff
@@ -94,47 +114,135 @@ func (s *composeService) injectConfigs(ctx context.Context, project *types.Proje
...
+
+	subdir := os.DirFS(path)
+	b := bytes.Buffer{}
+	tarWriter := tar.NewWriter(&b)
+
+	// build the tar by walking instead of using archive/tar.Writer.AddFS to be able to adjust mode, gid and uid
+	err = fs.WalkDir(subdir, ".", func(filePath string, d fs.DirEntry, err error) error {
+		header := &tar.Header{
+			Name:    filepath.Join(config.Target, filePath),
+			Mode:    mode,
```
andoks, line 174: :question: Should directories have the same mode as files? Or should they perhaps have the exec bit set for the owner/group/other to access them?

andoks:
@ndeloof left you a few questions in the review

andoks:
@alexey-sh 

> any ETA?

See discussion in #11867 - it might take some time to decide how to proceed.

pkg/compose/secrets.go
schaubl: I think it comes from the [compose-spec](https://github.com/compose-spec/compose-spec/blob/0d029ec8be0894970d08607972242163224fd28e/schema/compose-spec.json#L733-L782). 
There is no `content` property in `secrets`, only in `configs`. 
This correspond to [the doc](https://docs.docker.com/compose/compose-file/09-secrets/) where secrets can only use `file` or `environment`. 
I assume the reason is to prevent writing secrets value directly in the compose file.

pkg/compose/secrets.go
schaubl: I think it would make sense to remove at least public read (`0o440`)

pkg/compose/secrets.go
schaubl: > secret.file could also be a folder, not just a file

I'm not sure about that.
To keep the behaviour consistent with docker swarm/stack configs&secrets, I think only a file should be allowed.
This will also allow for a smooth transition if/when docker config&secrets will be available even when swarm mode is disabled.

schaubl:
Yesterday I made a [PR](https://github.com/docker/compose/pull/11984) which I then realized it was a duplicate of this one.

I think this PR could benefit from the changes I made in the file [pkg/compose/create.go](https://github.com/docker/compose/pull/11984/files#diff-1f4d596600c620b3df55dbfbc63e50a6f43b356865fae411c553d0937fbb6d91).

It basically keeps some preexisting checks and adds a preflight check to ensure the file used as the source for config/secret can be accessed so that if it's not the case it will fail before creating the container.

Please also see the following comment I made [here](https://github.com/docker/compose/pull/11871#discussion_r1680942735):

> > secret.file could also be a folder, not just a file
> 
> I'm not sure about that. To keep the behaviour consistent with docker swarm/stack configs&secrets, I think only a file should be allowed. This will also allow for a smooth transition if/when docker config&secrets will be available even when swarm mode is disabled.

andoks:
Replaced with #12251 that has been updated with rebase on v2.30.0 to get fix for #12186

PR #11874, opened 2024-06-01
Make port arg optional in `docker compose port`

### LINKED ISSUE #11859: Make PRIVATE_PORT optional for `docker compose port`
### Description

"Normal" docker already does this. See the usage for `docker port`:

```
$ docker port --help

Usage: docker port CONTAINER [PRIVATE_PORT[/PROTO]]
```

However `docker compose` differs:
```
$ docker compose port --help

Usage:  docker compose port [OPTIONS] SERVICE PRIVATE_PORT
```

I can look into making a PR if you agree this should be optional.

crazybolillo:
**What I did**

This comand no longer needs a mandatory port. If no port is provided it will instead list all published ports. This behavior falls in line with the behavior provided by `docker port`.

**Related issue**
Closes #11859.

### DIFF
--- a/cmd/compose/port.go
+++ b/cmd/compose/port.go
@@ -40,10 +40,15 @@ func portCommand(p *ProjectOptions, dockerCli command.Cli, backend api.Service) 
 		ProjectOptions: p,
 	}
 	cmd := &cobra.Command{
-		Use:   "port [OPTIONS] SERVICE PRIVATE_PORT",
-		Short: "Print the public port for a port binding",
-		Args:  cobra.MinimumNArgs(2),
+		Use:   "port [OPTIONS] SERVICE [PRIVATE_PORT]",
+		Short: "List port mappings or print the public port of a specific mapping for the service",
+		Args:  cobra.RangeArgs(1, 2),
 		PreRunE: Adapt(func(ctx context.Context, args []string) error {
+			if len(args) == 1 {
+				opts.port = 0
+				return nil
+			}
+
 			port, err := strconv.ParseUint(args[1], 10, 16)
 			if err != nil {
 				return err
@@ -67,7 +72,8 @@ func runPort(ctx context.Context, dockerCli command.Cli, backend api.Service, op
 	if err != nil {
 		return err
 	}
-	ip, port, err := backend.Port(ctx, projectName, service, opts.port, api.PortOptions{
+
+	ports, err := backend.Port(ctx, projectName, service, opts.port, api.PortOptions{
 		Protocol: opts.protocol,
 		Index:    opts.index,
 	})
@@ -75,6 +81,9 @@ func runPort(ctx context.Context, dockerCli command.Cli, backend api.Service, op
 		return err
 	}
 
-	fmt.Fprintf(dockerCli.Out(), "%s:%d\n", ip, port)
+	for _, port := range ports {
+		fmt.Fprintf(dockerCli.Out(), "%s\n", port.String())
+	}
+
 	return nil
 }
--- a/pkg/compose/port.go
+++ b/pkg/compose/port.go
@@ -26,18 +26,40 @@ import (
 	moby "github.com/docker/docker/api/types"
 )
 
-func (s *composeService) Port(ctx context.Context, projectName string, service string, port uint16, options api.PortOptions) (string, int, error) {
+func (s *composeService) Port(ctx context.Context, projectName string, service string, port uint16, options api.PortOptions) (api.PortPublishers, error) {
 	projectName = strings.ToLower(projectName)
 	container, err := s.getSpecifiedContainer(ctx, projectName, oneOffInclude, false, service, options.Index)
 	if err != nil {
-		return "", 0, err
+		return nil, err
 	}
-	for _, p := range container.Ports {
-		if p.PrivatePort == port && p.Type == options.Protocol {
-			return p.IP, int(p.PublicPort), nil
+
+	if port != 0 {
+		for _, p := range container.Ports {
+			if p.PrivatePort == port && p.Type == options.Protocol {
+				return api.PortPublishers{
+					api.PortPublisher{
+						URL:           p.IP,
+						TargetPort:    p.PrivatePort,
+						PublishedPort: p.PublicPort,
+						Protocol:      p.Type,
+					},
+				}, nil
+			}
+		}
+		return nil, portNotFoundError(options.Protocol, port, container)
+	}
+
+	res := make(api.PortPublishers, len(container.Ports))
+	for idx, p := range container.Ports {
+		res[idx] = api.PortPublisher{
+			URL:           p.IP,
+			TargetPort:    p.PrivatePort,
+			PublishedPort: p.PublicPort,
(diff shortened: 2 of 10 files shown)


cmd/compose/port.go
```diff
@@ -40,10 +41,15 @@ func portCommand(p *ProjectOptions, dockerCli command.Cli, backend api.Service)
 		ProjectOptions: p,
 	}
 	cmd := &cobra.Command{
-		Use:   "port [OPTIONS] SERVICE PRIVATE_PORT",
-		Short: "Print the public port for a port binding",
-		Args:  cobra.MinimumNArgs(2),
+		Use:   "port [OPTIONS] SERVICE [PRIVATE_PORT]",
+		Short: "List port mappings or a specific mapping for the service",
```
glours, line 45:
```suggestion
		Short: "List port mappings or print the public port for a specific mapping for the service",
```

cmd/compose/port.go
```diff
@@ -62,19 +68,45 @@ func portCommand(p *ProjectOptions, dockerCli command.Cli, backend api.Service)
...
-	}
 
-	fmt.Fprintf(dockerCli.Out(), "%s:%d\n", ip, port)
-	return nil
+	if opts.list {
+		summaries, err := backend.Ps(ctx, projectName, api.PsOptions{Services: []string{service}})
+		if err != nil {
+			return err
+		}
+		printPorts(dockerCli, summaries[opts.index].Publishers, opts)
```
glours, line 98: We need to test the size of the `summaries` slice otherwise we would potentially trigger `slice bounds out of range` error

cmd/compose/port.go
```diff
@@ -62,19 +68,45 @@ func portCommand(p *ProjectOptions, dockerCli command.Cli, backend api.Service)
...
+		return nil
+	} else {
```
glours, line 100: the `else` isn't needed as the following statement is the last one and the `if` section finish with a `return`

glours (changes requested):
Thanks for your contribution, I have some small suggestions to improve your changes

crazybolillo:
Thanks for the review, I have taken care of your observations.

glours (approved):
LGTM

cmd/compose/port.go
ndeloof: this is just the same help used in docker/cli

cmd/compose/port.go
```diff
@@ -40,10 +41,15 @@ func portCommand(p *ProjectOptions, dockerCli command.Cli, backend api.Service)
...
 	cmd := &cobra.Command{
-		Use:   "port [OPTIONS] SERVICE PRIVATE_PORT",
-		Short: "Print the public port for a port binding",
-		Args:  cobra.MinimumNArgs(2),
+		Use:   "port [OPTIONS] SERVICE [PRIVATE_PORT]",
+		Short: "List port mappings or print the public port of a specific mapping for the service",
+		Args:  cobra.RangeArgs(1, 2),
 		PreRunE: Adapt(func(ctx context.Context, args []string) error {
+			if len(args) == 1 {
+				opts.list = true
```
ndeloof, line 49: incompatibility with `--index` should be tested

cmd/compose/port.go
```diff
@@ -62,11 +68,40 @@ func portCommand(p *ProjectOptions, dockerCli command.Cli, backend api.Service)
...
+}
+
 func runPort(ctx context.Context, dockerCli command.Cli, backend api.Service, opts portOptions, service string) error {
 	projectName, err := opts.toProjectName(ctx, dockerCli)
 	if err != nil {
 		return err
 	}
+
+	if opts.list {
+		summaries, err := backend.Ps(ctx, projectName, api.PsOptions{Services: []string{service}})
```
ndeloof, line 94: better update `backend.Port` so it can manage multiple services

cmd/compose/port.go
```diff
@@ -62,11 +68,40 @@ func portCommand(p *ProjectOptions, dockerCli command.Cli, backend api.Service)
...
+		if err != nil {
+			return err
+		}
+		if len(summaries) <= opts.index {
```
ndeloof, line 98: []ContainerSummary index is unrelated to replica index in service set by `--index`.

ndeloof (changes requested):

cmd/compose/port.go
```diff
@@ -62,11 +68,40 @@ func portCommand(p *ProjectOptions, dockerCli command.Cli, backend api.Service)
...
 }
 
+func printPorts(dockerCli command.Cli, publishers api.PortPublishers, opts portOptions) {
+	for _, publisher := range publishers {
+		if publisher.Protocol != opts.protocol {
+			continue
+		}
+
+		var format string
+		if strings.Contains(publisher.URL, ":") {
```
ndeloof, line 78: would make sense to me `PortPublisher` has a `String()` method to handle this, aligned with docker/cli

ndeloof:
while not strictly required, I think we should use this opportunity while revisiting this command to add support for app_protocol and port name (https://github.com/compose-spec/compose-spec/blob/master/05-services.md#long-syntax-3)

crazybolillo:
> while not strictly required, I think we should use this opportunity while revisiting this command to add support for app_protocol and port name (https://github.com/compose-spec/compose-spec/blob/master/05-services.md#long-syntax-3)

Wow, did not even know those existed. How do you envision it? Allow filtering? Or just displaying the name and app_protocol if they exist for the given port?

cmd/compose/port.go
crazybolillo: Should `--index` be prohibited if no specific port was provided?

crazybolillo:
Went ahead and changed `backend.Port `so it returns a slice of publishers. That should fix other issues like dealing with `--index`. Since I was modifying `PortPublisher` I went ahead and also changed its port members to be uint16 so it falls in line with `types.Port`

I think the only thing left is the question about `--index` incompatibility, and support for `app_protocol` and port name,

pkg/compose/port.go
```diff
@@ -26,18 +26,40 @@ import (
 	moby "github.com/docker/docker/api/types"
 )
 
-func (s *composeService) Port(ctx context.Context, projectName string, service string, port uint16, options api.PortOptions) (string, int, error) {
+func (s *composeService) Port(ctx context.Context, projectName string, service string, port uint16, options api.PortOptions) (api.PortPublishers, error) {
```
ndeloof, line 29: as signature changes, and breaks compatibility, makes sense to _also_ rename `Ports`(plural)

ndeloof (approved):

ndeloof:
`--index` is used to select service replica, already implemented in `backend.Port` calling `getSpecifiedContainer`
Adding support for app_protocol would not be that simple, as such metadata only exists in compose model, not in actual resources. Let's keep this for a future improvement

pkg/compose/port.go
crazybolillo: @ndeloof Sorry to bother but do you have any thoughts on my comments? Should we still perform the renaming?

PR #11888, opened 2024-06-06
Chore(publish): remove unreachable code

Lunal98:
if statement is preceded by another one that has the same condition and ends with a return.

**What I did**

Removed the statement

50073150871_f0ba0ee441_c


pkg/compose/publish.go
```diff
@@ -87,15 +87,6 @@ func (s *composeService) publish(ctx context.Context, project *types.Project, re
 		if err != nil {
 			return err
 		}
```
glours, line 89: IHMO if we need to remove one of the 2 identical if, we should keep the one sending the event and remove this one.

Lunal98: Fair enough, swapped it around.

glours (approved):
LGTM

ndeloof (approved):

Issue #11887, opened 2024-06-06
[BUG] option --env-file not working after updating from 2.27.0 to 2.27.1


fichte:
### Description

In version 2.27.1, variables loaded through an .env file called with --env-file are not interpreted.

## working version 2.27.0
```
user@stage:~# sudo apt-get install docker-compose-plugin=2.27.0-1~debian.12~bookworm
Preparing to unpack .../docker-compose-plugin_2.27.0-1~debian.12~bookworm_amd64.deb ...
Unpacking docker-compose-plugin (2.27.0-1~debian.12~bookworm) over (2.27.1-1~debian.12~bookworm) ...
Setting up docker-compose-plugin (2.27.0-1~debian.12~bookworm) ...
user@stage:~# docker compose --env-file /srv/docker/project/stage/.env ps -a
NAME                             IMAGE                                                           COMMAND                  SERVICE      CREATED        STATUS        PORTS
xx-xx-part1-1                    registry.example.com/xx/xx/xx/part1:f362a748                    "/docker-entrypoint.…"   part1        20 hours ago   Up 20 hours   80/tcp
xx-xx-part2-1                    registry.example.com/xx/xx/xx/part2:f362a748                    "/docker-entrypoint.…"   part2        20 hours ago   Up 20 hours   80/tcp
user@stage:~#
```

## none working version 2.27.1
```
user@stage:~# apt-get install docker-compose-plugin=2.27.1-1~debian.12~bookworm
Preparing to unpack .../docker-compose-plugin_2.27.1-1~debian.12~bookworm_amd64.deb ...
Unpacking docker-compose-plugin (2.27.1-1~debian.12~bookworm) over (2.27.0-1~debian.12~bookworm) ...
Setting up docker-compose-plugin (2.27.1-1~debian.12~bookworm) ...
user@stage:~# docker compose --env-file /srv/docker/project/stage/.env ps -a
no configuration file provided: not found
user@stage:~#
```

## contents of /srv/docker/project/stage/.env
```
user@stage:~# cat /srv/docker/project/stage/.env
COMPOSE_ENV=stage
COMPOSE_PROJECT_NAME=project
COMPOSE_FILE=/srv/docker/project/stage/compose/stack-common.yml:/srv/docker/project/stage/compose/stack.yml
COMPOSE_STACK_SHA=4211ca832c58733cc895dfddb8966f86c2a0df6f228576557bfc974349dfdd60
COMPOSE_COMMIT_SHA=d284ede3
[...]

Issue #11880, opened 2024-06-03
[BUG] Incompatible `.env` file behavior


zlondrej:
### Description

Before `2.27.1`, `.env` file was automatically loaded from the **project directory**. However in `2.27.1`, the `.env` file is loaded only from the **working directory** and the **project directory's** `.env` file is ignore regardless whether there's `.env` in the **working directory**.

This affects both `--project-directory` and `--file` options and it seems that `--env-file` is also ignored.

### Steps To Reproduce

Behavior can be easily verified using `docker compose config` command, but affect other commands such as `docker compose up`.

Let's assume scenario with following project structure:

```
.   #  current working directory with no .env file
├── sub
│   ├── docker-compose.yaml
│   └── .env
└── test-environment -> sub/.env
```

```sh
# file: sub/.env
COMPOSE_PROFILES=sub
```

```yaml
# file: sub/docker-compose.yaml
services:
  dummy:
    image: placeholder/dummy
    profiles:
      - sub
```

---

## `2.27.0` and prior at least down to `2.24.5`

All 3 of the commands below resulted in the same output and all these commands correctly load the `sub` profile.
```sh
docker compose --project-directory sub config
docker compose --file sub/docker-compose.yaml config
(cd sub && docker compose config)
```

```
name: sub
services:
  dummy:
    profiles:
      - sub
    image: placeholder/dummy
    networks:
      default: null
networks:
  default:
    name: sub_default
```

## `2.27.1`

When `docker compose` specified either `--project-directory sub` or `--file sub/docker-compose.yaml`, the `./.env` file was loaded instead of `./sub/.env`.

So as expected, command that uses `cd` still behaves the same way.

```sh
(cd sub && docker compose config)
```

```
name: sub
services:
  dummy:
    profiles:
      - sub
    image: placeholder/dummy
    networks:
      default: null
networks:
  default:
    name: sub_default
```

However 2 of the commands below resulted in the different output compared to prior versions as no `.env` profile is loaded and `COMPOSE_PROFILES` is not set.
```sh
docker compose --project-directory sub config
docker compose --file sub/docker-compose.yaml config
```

```
name: sub
services: {}
```

This then results in behavior such as getting `no service selected` when using `docker compose --project-directory <dir> up`. Command which used to work in past suddenly stopped working.

What makes it even worse is that `--env-file` should have a priority over implicit default, it seems to be ignored regardless of file's location.

```sh
docker compose --env-file sub/.env --project-directory sub config
[...]

PR #11889, opened 2024-06-07
Fix dot env file to define COMPOSE_* variables

ndeloof:
**What I did**
Revisit support for setting COMPOSE_* variable using a dot env file
The actual logic is now implemented and documented all within `toProjectOptions`

**Related issue**
closes https://github.com/docker/compose/issues/11887
closes https://github.com/docker/compose/issues/11880

image

### DIFF
--- a/cmd/compose/compose.go
+++ b/cmd/compose/compose.go
@@ -28,10 +28,8 @@ import (
 	"syscall"
 
 	"github.com/compose-spec/compose-go/v2/cli"
-	"github.com/compose-spec/compose-go/v2/dotenv"
 	"github.com/compose-spec/compose-go/v2/loader"
 	"github.com/compose-spec/compose-go/v2/types"
-	composegoutils "github.com/compose-spec/compose-go/v2/utils"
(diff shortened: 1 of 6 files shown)


cmd/compose/compose.go
```diff
@@ -601,42 +599,6 @@ func RootCommand(dockerCli command.Cli, backend Backend) *cobra.Command { //noli
 	return c
 }
 
-// If user has a local .env file, load it as os.environment so it can be used to set COMPOSE_ variables
-// This also allows to override values set by the default .env in a compose project when ran from a distinct folder
-func setEnvWithLocalDotEnv(prjOpts *ProjectOptions) error {
```
jhrotko, line 606: Question: why are we removing this?

jhrotko:
LGTM: only left a question which is non-blocking

cmd/compose/compose.go
ndeloof: this is the way we used to load PWD/.env before we know the actual project directory. This logic is now covered by toProjectOptions

PR #11892, opened 2024-06-07
go.mod: github.com/docker/docker v27.0-dev, github.com/docker/cli v27.0-dev

akerouanton:
full diffs:

- https://github.com/moby/moby/compare/v26.1.3...a24af26aba4abaf3c1d553a08d431b15b3b602fb
- https://github.com/docker/cli/compare/v26.1.3...64206aef767d23420803badbd205d1c5a661f097


go.mod
```diff
@@ -59,6 +59,8 @@ require (
 	gotest.tools/v3 v3.5.1
 )
 
+replace github.com/docker/cli v26.1.3+incompatible => github.com/docker/cli v26.1.1-0.20240607143903-7ae9f2738c6b+incompatible
```
akerouanton, line 62: Not sure why, but `go mod tidy` kept reverting my change when trying to do that in the `require` block above 🤔 @thaJeztah @glours @ndeloof Any ideas?

thaJeztah: Because 26.1.1 is lower than 26.1.3. Thank Go for inventing pseudo versions and not understanding release branches

akerouanton: So we'll need to make a rc1 tag to make it possible to update github.com/docker/cli here?

thaJeztah:
Did a quick rebase and update to latest in https://github.com/docker/compose/pull/11926 to test in other repositories.

Issue #11869, opened 2024-05-31
[BUG] compose captures CTRL+C (MacOS, Iterm2, zsh)


ErjanGavalji:
### Description

Running `docker compose up` when the program in the container exits successfully leaves the CTRL+C combination captured and the key combination does not work anymore.

### Steps To Reproduce

On a MacOS (Sonoma 14.5), using Iterm2 (v3.5.0) and zsh (5.9),

1. Add the following two files:

*docker-compose.yml*:
```
name: simple-test
services:
  app:
    image: node:20.14.0-alpine3.20
    command:
      - node
      - /app.mjs
    volumes:
      - ./app.mjs:/app.mjs
```

and *app.mjs*:
```
console.log("Hello, world!");
```

2. Run `docker compose up`
3. Type "anything"
4. Press CTRL+C

### Compose Version

```Text
v2.27.0-desktop.2
```

### Docker Environment

```Text
Client:
 Version:    26.1.1
 Context:    desktop-linux
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.14.0-desktop.1
    Path:     /Users/testuser/.docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.27.0-desktop.2
    Path:     /Users/testuser/.docker/cli-plugins/docker-compose
  debug: Get a shell into any image or container (Docker Inc.)
    Version:  0.0.29
    Path:     /Users/testuser/.docker/cli-plugins/docker-debug
  dev: Docker Dev Environments (Docker Inc.)
    Version:  v0.1.2
    Path:     /Users/testuser/.docker/cli-plugins/docker-dev
  extension: Manages Docker extensions (Docker Inc.)
    Version:  v0.2.23
    Path:     /Users/testuser/.docker/cli-plugins/docker-extension
  feedback: Provide feedback, right in your terminal! (Docker Inc.)
    Version:  v1.0.4
    Path:     /Users/testuser/.docker/cli-plugins/docker-feedback
  init: Creates Docker-related starter files for your project (Docker Inc.)
    Version:  v1.1.0
    Path:     /Users/testuser/.docker/cli-plugins/docker-init
  sbom: View the packaged-based Software Bill Of Materials (SBOM) for an image (Anchore Inc.)
    Version:  0.6.0
    Path:     /Users/testuser/.docker/cli-plugins/docker-sbom
  scout: Docker Scout (Docker Inc.)
[...]

PR #11896, opened 2024-06-10
enforce keyboard.Close is always executed to restore terminal

ndeloof:
**What I did**
ensure `keyboard.Close()` is called as `up` command complete. `Close()` is re-entrant so this is not an issue it get called even after an interruption (Ctrl+C) and clean shutdown, where codebase manages this already

**Related issue**
fixes https://github.com/docker/compose/issues/11869


glours (approved):
LGTM

PR #11893, opened 2024-06-07
go.mod: github.com/compose-spec/compose-go v2.1.3

akerouanton:
- Depends on https://github.com/docker/compose/pull/11892

full diff:

- https://github.com/compose-spec/compose-go/compare/7dfa54c9658b...v2.1.3

PR #11912, opened 2024-06-17
bump compose-go to version v2.1.3

glours:
**What I did**
Bump compose-go to version `v2.1.3`

**Related issue**
N/A


glours:
Removed as we just need to bump the version in @akerouanton PR https://github.com/docker/compose/pull/11893

PR #11919, opened 2024-06-18
using as flag of the up command, watch was blocking process shutdown

glours:
This happened when sunsetting the application from docker compose down command

### DIFF
--- a/pkg/compose/watch.go
+++ b/pkg/compose/watch.go
@@ -183,15 +183,13 @@ func (s *composeService) watch(ctx context.Context, syncChannel chan bool, proje
 	}
 	options.LogTo.Log(api.WatchLogger, "Watch enabled")
 
-	err = eg.Wait()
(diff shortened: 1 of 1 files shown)


pkg/compose/watch.go
```diff
@@ -183,15 +183,14 @@ func (s *composeService) watch(ctx context.Context, syncChannel chan bool, proje
...
 
-	err = eg.Wait()
 	for {
 		select {
 		case <-ctx.Done():
-			return err
+			return eg.Wait()
 		case <-syncChannel:
 			options.LogTo.Log(api.WatchLogger, "Watch disabled")
 			ctx.Done()
```
ndeloof, line 192: ctx.Done() here is weird, this does nothing (return the done channel, this doesn't stop context)

ndeloof (approved):

Issue #11900, opened 2024-06-12
[BUG] docker inline config when changed is not being taken into account


0xbase12:
### Description

On change of inline config in yaml, I'm not getting the new content on rerun. I need an explicit option `--force-recreate` or to do a `docker compose down` to get the new content. I expect the behavior to be similar to an env variable change that is taken into account on change and make container restart on `docker compose up`

### Steps To Reproduce

How to reproduce:
1. create following docker-compose.yaml
```yaml
name: config-inline
services:
  service:
    image: ubuntu
    entrypoint: cat
    command: /path/to/config.txt
    configs:
      - source: config.txt
        target: /path/to/config.txt
configs:
  config.txt:
    content: |
      foo bar 1
```
2. `docker compose up`
4. change config.txt content in the yaml to foo bar 2
5. rerun `docker compose up`

Output command 5:
```
# docker compose up
[+] Running 1/0
 ✔ Container config-inline-service-1  Created                                                                                                                                                                                                                                                                                                                          0.0s
Attaching to service-1
service-1  | foo bar 1
service-1 exited with code 0
``` 

Expected output:
```
d1c6f75f97dd:/khaled# docker compose up
[+] Running 1/0
 ✔ Container config-inline-service-1  Created                                                                                                                                                                                                                                                                                                                          0.0s
Attaching to service-1
service-1  | foo bar 2
service-1 exited with code 0
```


### Compose Version

```Text
Docker Compose version v2.23.3
```


### Docker Environment

```Text
Client:
 Version:    25.0.5
 Context:    default
 Debug Mode: false
 Plugins:
  compose: Docker Compose (Docker Inc.)
    Version:  v2.23.3
    Path:     /usr/libexec/docker/cli-plugins/docker-compose

Server:
 Containers: 97
  Running: 14
  Paused: 0
  Stopped: 83
 Images: 64
 Server Version: 23.0.6
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: active
  NodeID: cz1gqp5o4uw6ppyics6xn6a2k
  Is Manager: true
[...]

PR #11931, opened 2024-06-23
fix(hash): recreate container on project config content change

idsulik:
**What I did**
Fixed `hash.ServiceHash()` to support config content change

**Related issue**
https://github.com/docker/compose/issues/11900

image

### DIFF
--- a/pkg/compose/hash.go
+++ b/pkg/compose/hash.go
@@ -17,9 +17,13 @@
 package compose
 
 import (
+	"bytes"
 	"encoding/json"
+	"fmt"
+	"time"
 
 	"github.com/compose-spec/compose-go/v2/types"
+	"github.com/docker/compose/v2/pkg/utils"
 	"github.com/opencontainers/go-digest"
 )
 
@@ -35,20 +39,84 @@ func ServiceHash(o types.ServiceConfig) (string, error) {
 	o.DependsOn = nil
 	o.Profiles = nil
 
-	bytes, err := json.Marshal(o)
+	data, err := json.Marshal(o)
 	if err != nil {
 		return "", err
 	}
-	return digest.SHA256.FromBytes(bytes).Encoded(), nil
+	return digest.SHA256.FromBytes(data).Encoded(), nil
+}
+
+// ServiceConfigsHash computes the configuration hash for service configs.
+func ServiceConfigsHash(project *types.Project, serviceConfig types.ServiceConfig) (map[string]string, error) {
+	serviceNameToHash := make(map[string]string)
+	for _, config := range serviceConfig.Configs {
+		file := project.Configs[config.Source]
+		b, err := createTarForConfig(project, types.FileReferenceConfig(config), types.FileObjectConfig(file))
+		if err != nil {
+			return nil, err
+		}
+
+		serviceNameToHash[config.Target] = digest.SHA256.FromBytes(b.Bytes()).Encoded()
+	}
+
+	return serviceNameToHash, nil
+}
+
+// ServiceSecretsHash computes the configuration hash for service secrets.
+func ServiceSecretsHash(project *types.Project, serviceConfig types.ServiceConfig) (map[string]string, error) {
+	serviceNameToHash := make(map[string]string)
+	for _, secret := range serviceConfig.Secrets {
+		file := project.Secrets[secret.Source]
+		b, err := createTarForConfig(project, types.FileReferenceConfig(secret), types.FileObjectConfig(file))
+		if err != nil {
+			return nil, err
+		}
+
+		serviceNameToHash[secret.Target] = digest.SHA256.FromBytes(b.Bytes()).Encoded()
+	}
+
+	return serviceNameToHash, nil
+}
+
+func createTarForConfig(
+	project *types.Project,
+	serviceConfig types.FileReferenceConfig,
+	file types.FileObjectConfig,
+) (*bytes.Buffer, error) {
+	// fixed time to ensure the tarball is deterministic
+	modTime := time.Unix(0, 0)
+
+	if serviceConfig.Target == "" {
+		serviceConfig.Target = "/" + serviceConfig.Source
+	}
+
+	switch {
+	case file.Content != "":
+		return bytes.NewBuffer([]byte(file.Content)), nil
+	case file.Environment != "":
+		env, ok := project.Environment[file.Environment]
+		if !ok {
+			return nil, fmt.Errorf(
+				"environment variable %q required by file %q is not set",
+				file.Environment,
+				file.Name,
+			)
+		}
+		return bytes.NewBuffer([]byte(env)), nil
+	case file.File != "":
+		return utils.CreateTarByPath(file.File, modTime)
+	}
+
+	return nil, fmt.Errorf("config %q is empty", file.Name)
 }
 
 // NetworkHash computes the configuration hash for a network.
 func NetworkHash(o *types.NetworkConfig) (string, error) {
-	bytes, err := json.Marshal(o)
+	data, err := json.Marshal(o)
 	if err != nil {
 		return "", err
 	}
-	return digest.SHA256.FromBytes(bytes).Encoded(), nil
+	return digest.SHA256.FromBytes(data).Encoded(), nil
 }
 
 // VolumeHash computes the configuration hash for a volume.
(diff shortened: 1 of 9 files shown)


ndeloof:
While I understand the intent, I don't like we get the config content added into the service hash. This also only makes sense as the config content is inlined. 
I wonder we could rely on a label to track the config state by the time it was created : `com.docker.compose.config.name=<config hash>`.

pkg/compose/hash.go
```diff
@@ -37,5 +37,12 @@ func ServiceHash(o types.ServiceConfig) (string, error) {
 	if err != nil {
 		return "", err
 	}
+
+	for _, serviceConfig := range o.Configs {
+		if projectConfig, ok := project.Configs[serviceConfig.Source]; ok {
+			bytes = append(bytes, []byte(projectConfig.Content)...)
```
ndeloof, line 43: This won't detect changes for config set by environment of file

idsulik: added support for env and config file

idsulik:
> I don't like we get the config content added into the service hash

why? 



> I wonder we could rely on a label to track the config state by the time it was created : `com.docker.compose.config.name=<config hash>`.

I don't get the idea, _time it was created_ - do you mean config file? or docker-compose.yaml? but docker-compose can refer to external config file

ndeloof:
time _container_ was created, so we can check it needs to be recreated if current config doesn't match

pkg/compose/hash.go
```diff
@@ -37,5 +38,22 @@ func ServiceHash(o types.ServiceConfig) (string, error) {
 	if err != nil {
 		return "", err
 	}
+
+	for _, serviceConfig := range o.Configs {
+		if projectConfig, ok := project.Configs[serviceConfig.Source]; ok {
```
jhrotko, line 43: nit: I would invert the if statements so that we do not get the code so nested

idsulik: @jhrotko done

ndeloof:
Some poins:
1. At some point we will have to do the same for secrets (same constraints) and networks/volumes (which can get labeled with a has). Solution we adopt here should be extensible to allow this in the future
2. Configs set by file are implemented by a bind mount but this may change in near future (see https://github.com/docker/compose/pull/11871) so we also need to consider those
3. There's no technical reason docker engine can't offer secrets/configs natively, just this is guarded by Swarm mode, but AFAIK some discussion happened to get them _also_ available in standalone mode. I can't tell if/when this would take place, but preferably the logic here should consider this may happen
4. Last but not least, I'd prefer we don't mix service config hash with resources it depends on, so my suggestion to introduce an additional label `com.docker.compose.config.xx=hash` that we can use to track this relation, and need to recreate container, without the need to change the service hash computation (which as impact on existing installations)

idsulik:
@ndeloof thanks for the details. pushed changes:
1. reverted old changes
2. added new func `ServiceDependenciesHash` and label 	
```
// ConfigHashDependenciesLabel stores configuration hash for a compose service dependencies
ConfigHashDependenciesLabel = "com.docker.compose.config-hash-dependencies"
```

pkg/compose/hash.go
```diff
@@ -39,3 +40,28 @@ func ServiceHash(o types.ServiceConfig) (string, error) {
...
+	for _, serviceConfig := range o.Configs {
+		projectConfig, ok := project.Configs[serviceConfig.Source]
+		if !ok {
+			continue
+		}
+
+		if projectConfig.Content != "" {
+			bytes = append(bytes, []byte(projectConfig.Content)...)
+		} else if projectConfig.File != "" {
+			content, err := os.ReadFile(projectConfig.File)
```
ndeloof, line 56: `File` could also be a folder. Maybe an option would be to read content into a tar (code exists to inject secrets) and compute hash from it

While the general approach looks good, I'd prefer we have a more explicit hash and label _per resource_, and a dedicated `hash` function for configs (and secrets), so that it make it easier to diagnose use of those labels

idsulik:
Up)

idsulik:
Hi @ndeloof  ,

I noticed that this pull request has been pending for a while. Is there anything more needed from my side? Please let me know if there are any updates or additional changes required.

Thanks!

idsulik:
Hi @ndeloof ,

I noticed that this pull request has been pending for a while. Is there anything more needed from my side? Please let me know if there are any updates or additional changes required.

Thanks!

idsulik:
Hi @ndeloof ,

I noticed that this pull request has been pending for a while. Is there anything more needed from my side? Please let me know if there are any updates or additional changes required.

Thanks!

idsulik:
Hi @ndeloof ,

I noticed that this pull request has been pending for a while. Is there anything more needed from my side? Please let me know if there are any updates or additional changes required.

Thanks!

idsulik:
Hi @ndeloof , what do you think about think about the PR?

pkg/compose/hash.go
```diff
@@ -40,3 +46,122 @@ func ServiceHash(o types.ServiceConfig) (string, error) {
...
+	}
+
+	if info.IsDir() {
+		// If it's a directory, read all files and concatenate their contents
+		err = filepath.Walk(path, func(path string, info os.FileInfo, err error) error {
+			if err != nil {
+				return err
+			}
+			if !info.IsDir() {
+				fileContent, err := os.ReadFile(path)
```
ndeloof, line 146: this will not detect file being rename. Maybe better to create a tar stream from file/dir content

idsulik: @ndeloof it creates a tar from the content here pkg/compose/hash.go:122. 

`b, err := utils.CreateTar(content, serviceConfig, modTime)`

idsulik: @ndeloof is there any reason to create tar inside a tar?   the `createTarForConfig` func calls this `readPathContent` func and creates a tar with the path content

ndeloof: here you are creating a tar with all content appended as []byte, this doesn't reflect the filesystem structure for intput. You have to create one tar entry per file with file/folder name.

idsulik: @ndeloof you're right, thank you! Updated the code, now it creates a tar file for a given path and keeps file structure

pkg/compose/hash.go
```diff
@@ -34,9 +38,75 @@ func ServiceHash(o types.ServiceConfig) (string, error) {
...
+		b, err := createTarForConfig(project, types.FileReferenceConfig(config), types.FileObjectConfig(file))
+
+		if err != nil {
+			return "", err
+		}
+
+		data = append(data, b.Bytes()...)
+	}
+
+	return digest.SHA256.FromBytes(data).Encoded(), nil
```
ndeloof, line 62: I'd prefer we have one label per config/secret mount, so it makes it easier to track|debug changes and container being recreated.
Also need to consider config can be mounted from docker host, i.e. file is not available for compose to compute hash, and then must be excluded from label / no label created

idsulik: do you mean something like that:
```
com.docker.compose.service.configs-hash-{configName}={hash}
com.docker.compose.service.configs-hash-{serviceName}={hash}
```
?

ndeloof: indeed, or maybe, to follow the dot-notation style used for labels, `com.docker.compose.service.configs.{configName}.hash`

idsulik: It seems to me that this will complicate the logic, first you need to generate a hash for each item separately, then you need to go through all labels whose names start with “com.docker.compose.service.configs.” to check if the hash has changed.

ndeloof: Doesn't look such a pain to me, as this would allow to trace reason we recreate a container, and make it easier to diagnose potential regressions (this sometimes happened :P)

```go
for c := range service.Configs {
  hash := labels["com.docker.compose.configs."+c+".hash
  expected := ConfigHash(project.Configs[c]
  if hash := expected {
    log.Debug("container has to be recreated after config %s has been updated", c)
    return DIVERGED
  }
}
```

idsulik: @ndeloof, but we don't have a config/service name that we can use to create the label name. I pushed changes to create a hash of the config/secret of each service so it'll be easy to figure out what service's config/secret caused the change

apollo13:
This would be a really great addition, anything I can help out with?

idsulik:
> This would be a really great addition, anything I can help out with?

no, thank you, waiting for @ndeloof response

apollo13:
@idsulik Maybe it might make sense to fix the failing linting checks before another round of reviews?

idsulik:
@ndeloof could you please check it?

idsulik:
@glours hi! Maybe you can check it? otherwise, it can be stuck here for a long time

idsulik:
@ndeloof @glours hi! Rebased the main branch and resolved conflicts, is there anything else I should do ?

idsulik:
I have to resolve conflicts several times, could you please check it? @ndeloof

ianhinder:
Would it be possible to get this in? I just ran into the problem it would fix.

Issue #11928, opened 2024-06-22
[BUG] (v2.28.0) Fancy (ANSI) progress output not available


leo9800:
### Description

Docker compose commands, e.g. `docker compose up -d`, used to produce a fancy interactive output on terminals. But when compose is upgraded to v2.28.0, the fancy output disappears, left only a plain output.

A sample output (fancy, ANSI) of `docker compose up -d` from compose v2.27.3

![Screenshot from 2024-06-22 14-41-14](https://github.com/docker/compose/assets/13928879/239fa8a2-199b-47b1-b753-63e92d8db795)

A sample output (plain) of `docker compose up -d` from compose v2.28.0

![Screenshot from 2024-06-22 14-42-39](https://github.com/docker/compose/assets/13928879/51aa3953-5dd8-46d8-9a3c-6ef08b5ece6a)

### Steps To Reproduce

1. install docker
2. install docker-compose
3. run `docker compose up -d` on any stack directory\*

\*: stack directory: a directory contains a docker compose stack, i.e. a `docker-compose.yml` file at least

### Compose Version

```Text
Docker Compose version 2.28.0
```

### Docker Environment

```Text
Client:
 Version:    26.1.4
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  0.15.1
    Path:     /usr/lib/docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  2.28.0
    Path:     /usr/lib/docker/cli-plugins/docker-compose

Server:
 Containers: 0
  Running: 0
  Paused: 0
  Stopped: 0
 Images: 0
 Server Version: 26.1.4
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: true
  Native Overlay Diff: false
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: ae71819c4f5e67bb4d5ae76a6b735f29cc25774e.m
 runc version: 
 init version: de40ad0
 Security Options:
  seccomp
   Profile: builtin
[...]

PR #11926, opened 2024-06-20
bump docker/docker and docker/cli v27.0.1-rc.1, compose-go v2.1.3

thaJeztah:
rebased of / updated version of:

- https://github.com/docker/compose/pull/11892
- https://github.com/docker/compose/pull/11893
- supersedes / closes https://github.com/docker/compose/pull/11892
- supersedes / closes https://github.com/docker/compose/pull/11893
- supersedes / closes https://github.com/docker/compose/pull/11916

### go.mod: github.com/compose-spec/compose-go v2.1.3

- related: https://github.com/compose-spec/compose-go/pull/637

diff:

- https://github.com/compose-spec/compose-go/compare/v2.1.2...v2.1.3

### go.mod: docker/docker and docker/cli v27.0.1-rc.1

diffs:

- https://github.com/containerd/ttrpc/compare/v1.2.4...v1.2.5
- https://github.com/docker/cli/compare/v26.1.4..v27.0.1-rc.1
- https://github.com/docker/docker/compare/v26.1.4..v27.0.1-rc.1

**What I did**

**Related issue**

**(not mandatory) A picture of a cute animal, if possible in relation to what you did**

PR #11933, opened 2024-06-24
fix terminal console check

glours:
this was broken by cli v27 release and usage of stream writer for the stderr the new type encapsulates the io.Writer and don't pass the console.File type assertion

**What I did**
Introduce the check of the console in Compose code (for windows terminal)

**Related issue**
fixes #11928 

**(not mandatory) A picture of a cute animal, if possible in relation to what you did**
image

PR #11934, opened 2024-06-24
Remove `console.Terminal` check and use `IsTerminal` from `streams.Out`

vvoland:
- follow up to: https://github.com/docker/compose/pull/11926

**What I did**
Remove `console.Terminal` check and use `IsTerminal` from `streams.Out`

docker/cli v27 changed the return value of `Err()` to `streams.Out`
which made the typecheck for `console.File` fail.

The check is no longer needed due to the `IsTerminal` method present in
`streams.Out` which also has a special handling for Windows console.

**Related issue**
fixes https://github.com/docker/compose/issues/11928


vvoland:
Oh, I didn't notice, there's already https://github.com/docker/compose/pull/11933

glours (approved):
LGTM

PR #11965, opened 2024-07-04
empty env variable with no value must be unset in container

ndeloof:
**What I did**

**Related issue**
see https://github.com/compose-spec/compose-go/pull/654


pkg/e2e/fixtures/environment/empty-variable/Dockerfile
```diff
@@ -0,0 +1,17 @@
+#   Copyright 2020 Docker Compose CLI authors
```
jhrotko, line 1: 2024?

jhrotko (approved):

glours (approved):
LGTM

Issue #11971, opened 2024-07-05
compose-files include not support pattern


k8scat:
### Description

When I want to include many compose files, I found that it does not support pattern `include`

<img width="518" alt="image" src="https://github.com/docker/compose/assets/55872984/71e401f8-4b1a-45a7-b14a-f64f4bf6d43d">

glours:
Hey @k8scat 
I have some concerns with this kind of behaviour, especially regarding the merging phase. 
In which order will the files be load? In alphabetical order? 
As the order could completely change the final generation, I think keeping the loading order explicit by using multiple `-f` will reduce confusion about the generated configuration.

idsulik:
I think we can ignore the order or use alphabetical order and if anyone needs to use a different order, they can specify all files manually in the needed order. I think that in most cases the order will not be important

k8scat:
> I think we can ignore the order or use alphabetical order and if anyone needs to use a different order, they can specify all files manually in the needed order. I think that in most cases the order will not be important

yeah, in my case the order will be ignored

ShadowLNC:
We also have about 20 `*.yml` files in a directory, each one being listed in the main `docker-compose.yml` file. In our case, the order is unimportant since they don't conflict with each other; this would simplify our main Compose file a lot.

ndeloof:
order is unimportant for you but it is for many others, and allowing this would bring potential misuses and confusion to our user base.
It's pretty easy to write a makefile target to list compose files is your directory and set `COMPOSE_FILES` variables so you don't have to maintain a dedicated aggregation compose file.

PR #11973, opened 2024-07-06
feat: add support globs pattern for compose files include

idsulik:
**What I did**
Added glob pattern support for compose-files include, for example, `docker compose up -f compose.*.yaml` will use all files:
```
compose.1.yaml
compose.2.yaml
compose.local.yaml
compose.{anything}.yaml
```

**Related issue**
https://github.com/docker/compose/issues/11971

image


glours:
see my comment https://github.com/docker/compose/issues/11971#issuecomment-2211708498 , I'm not sure this is a good idea to add this kind of behaviour.

idsulik:
The issue was closed https://github.com/docker/compose/issues/11971#issuecomment-2213224295

Issue #11967, opened 2024-07-05
[BUG] 2.28.1 multiple COMPOSE_* vars no longer read from .env file


ShadowLNC:
### Description

It seems like Compose is inconsistently reading the .env file for `COMPOSE_*` variables.

In 2.27.1 we had #11856 and #11955 where `COMPOSE_PROFILES` was ignored.

According to [the docs](https://docs.docker.com/compose/environment-variables/envvars/), I would expect all of the `COMPOSE_*` variables to be respected when placed in `.env` (or other file as specified by `--env-file`).

Variable|2.27.0|2.27.1|2.28.1
-|-|-|-
`COMPOSE_PROFILES`|:white_check_mark:|:x:|:white_check_mark:
`COMPOSE_ANSI`|:white_check_mark:|?|:x:
`COMPOSE_REMOVE_ORPHANS`|:information_source:|?|:x:
`COMPOSE_IGNORE_ORPHANS`|?|?|:white_check_mark:

:information_source: In 2.27.0, `COMPOSE_REMOVE_ORPHANS` in an `.env` file only worked during `compose up` and not during `compose down`.

### Steps To Reproduce

`docker-compose.yml`

```yaml
services:
    test1:
        image: nginx:alpine
        healthcheck:
            test: "wget -q -O - http://localhost:80/ || exit 1"
        profiles:
            - one
    test2:
        image: nginx:alpine
        healthcheck:
            test: "wget -q -O - http://localhost:80/ || exit 1"
        profiles:
            - two
```

`.env`

```config
COMPOSE_REMOVE_ORPHANS='1'
COMPOSE_ANSI='never'
COMPOSE_PROFILES='*'
```

For `COMPOSE_ANSI`, simply run `compose up` and see it's back to using the dynamic formatting.

For `COMPOSE_REMOVE_ORPHANS`, run `compose up` and then comment out the `test2` service. Running `compose down` wouldn't destroy `test2` in 2.27.0 but `compose up` did.

In 2.28.1 you get the warning about orphans on `compose up` (which can be suppressed by specifying `COMPOSE_IGNORE_ORPHANS=1`), but it doesn't remove them. (There is also no warning on `compose down` except for the network remaining in use, and perhaps there should be an explicit note about orphans.)

### Compose Version

`docker-compose` is no longer present on my system (perhaps this should be removed from the template prompt?)

```Text
$ docker compose version
Docker Compose version v2.28.1-desktop.1
```

### Docker Environment

<details>
[...]

PR #11974, opened 2024-07-08
restore setEnvWithDotEnv

ndeloof:
**What I did**
restore `setEnvWithDotEnv` to load .env into os.Env so the many places we rely on os.LookupEnv aren't borken
this also restore support for configuring buildx using env var set in .env

includes e2e test + a tiny fix to include (stopped) one-off containers as "orphans"

**Related issue**
closes https://github.com/docker/compose/issues/11967

image


pkg/compose/containers.go
```diff
@@ -118,17 +118,16 @@ func isRunning() containerPredicate {
...
-		return !utils.StringContains(services, service)
-	}
-}
-
 // isOrphaned is a predicate to select containers without a matching service definition in compose project
 func isOrphaned(project *types.Project) containerPredicate {
 	services := append(project.ServiceNames(), project.DisabledServiceNames()...)
 	return func(c moby.Container) bool {
+		// One-off container
+		v, ok := c.Labels[api.OneoffLabel]
```
jhrotko, line 126: question: what is the oneOffLabel?

ndeloof: this is used to distinguish containers created by  `docker compose run` vs actual services created by `up`

pkg/e2e/compose_run_test.go
```diff
@@ -37,7 +37,7 @@ func TestLocalComposeRun(t *testing.T) {
 			"Hello one more time")
 		lines = Lines(res.Stdout())
 		assert.Equal(t, lines[len(lines)-1], "Hello one more time", res.Stdout())
-		assert.Assert(t, !strings.Contains(res.Combined(), "orphan"))
+		assert.Assert(t, strings.Contains(res.Combined(), "orphan"))
```
jhrotko, line 40: question: was this a bug? :o

jhrotko (approved):
LGTM

pkg/e2e/compose_run_test.go
ndeloof: more or less. We only considered "orphans" for service container not declared in current compose model. This PR also consider stopped `run` containers as orphaned

glours (approved):

Issue #11976, opened 2024-07-08
[BUG] docker compose up service fails because of required network from another service


RicardoM17:
### Description

Based on the `compose.yml` file below, we should be able to start `test_service` without the external network `test_network` being present as `test_service` does not depend on it.

- If I remove `other_service` from the compose file then this is possible.
- If I disable `other_service` with for example a `docker profile` this is also possible.

### Steps To Reproduce

Given the following compose file:

```
services:

  test_service:
    image: python:3
    container_name: test_service
    command: sleep infinity

  other_service:
    image: python:3
    container_name: other_service
    command: sleep infinity
    networks:
      test_network:
        ipv4_address: 8.8.8.8

networks:
  test_network:
    external: true
    name: bee_rf  
```

```
docker compose -f compose.yml up -d test_service
```

returns
```
Error response from daemon: network test_network not found
```

### Compose Version

```Text
Docker Compose version v2.26.0
```

### Docker Environment

```Text
Client: Docker Engine - Community
 Version:    25.0.4
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.13.0
    Path:     /usr/libexec/docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.26.0
    Path:     /home/ricardom/.docker/cli-plugins/docker-compose
WARNING: Plugin "/home/ricardom/.docker/cli-plugins/docker-compose-v2.15.1" is not valid: plugin candidate "compose-v2.15.1" did not match "^[a-z][a-z0-9]*$"
WARNING: Plugin "/home/ricardom/.docker/cli-plugins/docker-compose-v2.26.0" is not valid: plugin candidate "compose-v2.26.0" did not match "^[a-z][a-z0-9]*$"

Server:
 Containers: 17
  Running: 2
  Paused: 0
  Stopped: 15
 Images: 151
 Server Version: 25.0.4
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 1
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local splunk syslog
 Swarm: inactive
[...]

PR #11978, opened 2024-07-09
exclude unecessary resources after services have been selected

ndeloof:
**What I did**
remove definition for unused resources _after_  services have been selected
includes e2e test to avoid a future regression with this feature

**Related issue**
fixes https://github.com/docker/compose/issues/11976


glours (approved):
LGTM

PR #11984, opened 2024-07-16
Update config/secret handling: Copy files into containers instead of bind mounting

schaubl:
Currently, Docker Compose copies a config or secret into the container if the source is an environment variable or inlined content. However, if the source is a file, the file is bind mounted into the container. This approach fails when Docker Compose is invoked on a different filesystem than the Docker daemon host filesystem (e.g., when called from within a container or remotely using the -H option).

This PR changes the behavior for file-based config/secret sources to match the handling of environment variables and inlined content. Specifically, it reads the file from the filesystem where Docker Compose is executed (like Docker Stack) and copies it into the container.

This update enhances the functionality of Docker Compose when run "remotely," providing the ability to push configuration files, a feature previously available only with Docker Stack/Swarm.

What I did:

* Updated the behavior for handling configs and secrets sourced from files.
* Ensured consistency with the documentation and the behavior of Docker Stack.
* Enabled Docker Compose to push configuration files when run remotely, aligning with Docker Stack/Swarm capabilities.

### DIFF
--- a/pkg/compose/secrets.go
+++ b/pkg/compose/secrets.go
@@ -21,6 +21,7 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"os"
 	"strconv"
 	"time"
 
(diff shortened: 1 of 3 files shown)


pkg/compose/secrets.go
```diff
@@ -31,7 +32,23 @@ import (
...
+		content := file.Content
+
+		if file.Environment != "" {
+			env, ok := project.Environment[file.Environment]
+			if !ok {
+				return fmt.Errorf("environment variable %q required by file %q is not set", file.Environment, file.Name)
+			}
+			content = env
+		} else if file.File != "" {
+			data, err := os.ReadFile(file.File)
```
ndeloof, line 44: `file` can actually be a folder. Need to change logic here to produce tar, not just extract content

ndeloof (commented):
One challenge we need to address here is that we have _some_ users who rely on bind mounts being used. While a simple workaround is for them to declare an explicit bind mount, this would still be a breaking change.

PR #11996, opened 2024-07-20
Use logrus instead of direct output to stderr

felixfontein:
**What I did**
There are two warnings on stderr that `--progress json` (#11478) still emits as plain text. This is because they are printed there directly, without using logrus or the events framework.

This PR changes these places to use `logrus.Warnf(...)` instead of `fmt.Fprintf(s.stderr(), ...)`.

**Related issue**
Ref: #11478 (not an issue, but a PR)

### DIFF
--- a/pkg/compose/build_classic.go
+++ b/pkg/compose/build_classic.go
@@ -44,6 +44,8 @@ import (
 	"github.com/docker/docker/pkg/streamformatter"
 
 	"github.com/docker/compose/v2/pkg/api"
+
+	"github.com/sirupsen/logrus"
 )
 
 //nolint:gocyclo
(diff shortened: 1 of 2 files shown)


felixfontein:
@cmrigney @ndeloof you might also be interested in this PR since it avoids more free-form stderr text in case the JSON stream progress writer is used.

pkg/compose/build_classic.go
```diff
@@ -180,7 +182,7 @@ func (s *composeService) doBuildClassic(ctx context.Context, project *types.Proj
 	aux := func(msg jsonmessage.JSONMessage) {
 		var result dockertypes.BuildResult
 		if err := json.Unmarshal(*msg.Aux, &result); err != nil {
-			fmt.Fprintf(s.stderr(), "Failed to parse aux message: %s", err)
+			logrus.Warnf("Failed to parse aux message: %s", err)
```
ndeloof, line 185: should be Errorf

felixfontein: Thanks, adjusted it.

felixfontein:
@ndeloof any chance of getting this merged anytime soon?

ndeloof (approved):

felixfontein:
@ndeloof thanks for reviewing and merging!

PR #11998, opened 2024-07-22
gha: add docker 27.1.0

thaJeztah:
**What I did**

**Related issue**


.github/workflows/ci.yml
```diff
@@ -145,6 +145,7 @@ jobs:
           - 25.0.5
           - 26.1.4
           - 27.0.3
+          - 27.1.0
```
glours, line 148: Can we just keep the last version for each major release to limit resources and CI time consumption?

thaJeztah: Yup, makes sense! Let me do so. Also considering to swap 24.x for 23.x (as 24 is no longer maintained, but 23 is maintained by Mirantis for their LTS release); will do that separately though (maybe there was a reason we didn't do 23.0)

thaJeztah: updated; PTAL 🫶

glours (approved):
LGTM

PR #12032, opened 2024-07-30
run: add support --no-build, --pull flags

legobeat:
Allows setting build- and pull policies for `docker-compose run` like for `docker run` and `docker-compose up`.

### DIFF
--- a/cmd/compose/run.go
+++ b/cmd/compose/run.go
@@ -47,6 +47,7 @@ type runOptions struct {
 	Detach        bool
 	Remove        bool
 	noTty         bool
+	noBuild       bool
 	tty           bool
 	interactive   bool
 	user          string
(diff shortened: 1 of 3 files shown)


cmd/compose/run.go
```diff
@@ -153,6 +153,13 @@ func runCommand(p *ProjectOptions, dockerCli command.Cli, backend api.Service) *
 					options.noTty = !options.tty
 				}
 			}
+			if cmd.Flags().Changed("build") {
+				if cmd.Flags().Changed("no-build") {
+					return fmt.Errorf("--build and --no-build are incompatible")
+				} else {
+					options.noTty = !options.tty
```
ndeloof, line 160: why have this here ?

cmd/compose/run.go
```diff
@@ -81,6 +82,12 @@ func (options runOptions) apply(project *types.Project) (*types.Project, error)
 	}
 
 	target.Tty = !options.noTty
+	if options.noBuild {
```
ndeloof, line 85: why configure build is no-build was requested?

Issue #11122, opened 2023-10-23
Add .env file support to COMPOSE_ENV_FILES


nickjj:
### Description

This is in relation to the new feature introduced in this PR: https://github.com/docker/compose/pull/11061

When `COMPOSE_ENV_FILES` is defined in an `.env` file it has no effect. This is a potential oversight / bug.

Here's a reproduceable example:

```yml
# docker-compose.yml
services:
  debian:
    image: "debian:bookworm-slim"
    restart: "${DOCKER_RESTART_POLICY:-unless-stopped}"
```

```sh
# .env
COMPOSE_ENV_FILES=.env,.env.local
DOCKER_RESTART_POLICY=yes
```

```sh
# .env.local
DOCKER_RESTART_POLICY=no
```

If you run `docker compose config` the value of restart is `yes` but it would be expected that it's `no` through the logic of:

- `docker compose` was called without explicitly defining any `--env-file` flags which means it will look for `.env` by default
- The new `COMPOSE_ENV_FILES` env var will be used which says to interpolate the `.env` file and then `.env.local`
  - This is the part that is failing the use case expectation
- `no` should be set

If you run `COMPOSE_ENV_FILES=.env,.env.local docker compose config` then everything works as expected but this requires altering your shell or creating a custom alias to run docker compose with that set. The original use case was to allow for running `docker compose` commands without modification.

ndeloof:
I'm :-1 on this.
`COMPOSE_ENV_FILES` let you define the env files you want to include, as an alternative to a (potentially long) list of .env files to be included. From this point of view, this is an improvement.
With your proposed change, we are just back to defaulting to `.env`. If `COMPOSE_ENV_FILES` in this file doesn't include `.env` then resulting environment would just be inconsistent. And another env file might override this variable and include other files, what would then be the expected result ?
Basically, I'd say "this is going too far"

nickjj:
As it currently stands (with `COMPOSE_ENV_FILES` as it is), we can do:

- `docker compose --env-file .env --env-file .env.custom build`
- `COMPOSE_ENV_FILES=.env,.env.custom docker compose build`

In both cases if you want `docker compose` to interpolate `.env.custom` you must modify how you use Docker Compose. You can no longer just call `docker compose`. You can do 1 of the following:

- Run either of the 2 options in the first set of bullets above
- Manually `export COMPOSE_ENV_FILES=.env,.env.custom` in your current shell
[...]

PR #12034, opened 2024-07-30
Updated COMPOSE_ENV_FILES in env files

ls-aron-kyle:
Updated the docker compose command to recursively search through any specified .env files for the COMPOSE_ENV_FILES parameter.

When found, the paths are followed recursively to extend the list of .env files used in the command.
A cache is kept that prevents circular dependencies.
Recursion is depth first.

**Related issue**

Example fix for #11122

Recursive guinea pigs

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ bin/
 coverage.out
 covdatafiles/
 .DS_Store
+/.idea
--- a/cmd/compose/compose.go
+++ b/cmd/compose/compose.go
@@ -647,12 +647,12 @@ func RootCommand(dockerCli command.Cli, backend Backend) *cobra.Command { //noli
 	c.Flags().MarkHidden("verbose") //nolint:errcheck
 	return c
 }
-
 func setEnvWithDotEnv(opts ProjectOptions) error {
 	options, err := cli.NewProjectOptions(opts.ConfigPaths,
 		cli.WithWorkingDirectory(opts.ProjectDir),
 		cli.WithOsEnv,
 		cli.WithEnvFiles(opts.EnvFiles...),
+		WithExtendedEnvFiles,
 		cli.WithDotEnv,
 	)
 	if err != nil {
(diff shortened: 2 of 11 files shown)


ndeloof:
👎  IMHO having an env file to declare COMPOSE_ENV_FILES demonstrates a mistake organizing config
It seems what you're looking for is some kind f support for `include` in dot env file. We have no plan yet for such a feature

ls-aron-kyle:
> 👎 IMHO having an env file to declare COMPOSE_ENV_FILES demonstrates a mistake organizing config It seems what you're looking for is some kind f support for `include` in dot env file. We have no plan yet for such a feature

The situation we're trying to solve is to standardize our env files while still providing a place to put local values. i.e. we'd like to commit the .env file but optionally reference a second file that devs can use for their personal config values, something like `~/.env.local`. By standardizing the locations it's easier to add automation to provision the `~/.env.local` file.

Judging by the comments in thread in https://github.com/docker/compose/issues/11122 there is interest in this feature.

I'm open to discussing an `include` feature, in particular how it might differ from this solution. At the end of the day I think we all just want the best user experience possible.

ShadowLNC:
There seems to be a common issue that some config is tracked/dynamically pulled, and some should not be tracked. Right now, Compose has no way to manage this config without changing the CLI invocation.

We can't set `COMPOSE_ENV_FILES` globally in the shell configuration, because there might be other projects which don't manage `.env` files the same way. We also can't expect users to remember to specify the "magic" `--env-file ...` arguments (in many cases, the users don't care about these files at all, and may not even know they exist if they're working on code in a container and not the Compose configuration itself).

For now, I can't use `COMPOSE_ENV_FILES` at all, so I've got a wrapper script around `docker compose` to add in the `--env-files` argument. This has attracted additional scrutiny on my code, where I'm being asked "why aren't we just letting users directly invoke `docker compose`?".

I wonder if we can settle on a "halfway" point with the following restrictions (which may help alleviate @ndeloof's concerns both here and on the linked issue):

- If Compose is not given the flag `--env-file` or the shell env var `COMPOSE_ENV_FILES`, it will first parse `.env` and see if `COMPOSE_ENV_FILES` exists
- If `.env` does have `COMPOSE_ENV_FILES`:
    - If it doesn't list `.env`, the remainder of `.env` is ignored
    - For any file that is listed (and parsed), including `.env`, no further attempt is made to read `COMPOSE_ENV_FILES`
- If `.env` does not have `COMPOSE_ENV_FILES`, then we follow the current main behaviour

ndeloof:
> I'm being asked "why aren't we just letting users directly invoke docker compose?".

Compose comes with many flags to support customization and various usages, it is expected `docker compose up` can't cover all needs out of the box. It's 100% legitimate to have a launcher script, typically a `Makefile` to select adequate flags depending on your use-case: run in development, run with debugger enabled, run for tests, etc.

Issue #11102, opened 2023-10-17
[BUG] `watch` `sync` does not pass project files to container / does not reflect file changes between container restarts


edvordo:
### Description

I've wanted to try the newest `docker compose watch` (released with latest desktop), but I have two issue with it.
The [announcement blog post](https://www.docker.com/blog/announcing-docker-compose-watch-ga-release/) seems to suggest, that adding a config such as [this one in an example repository](https://github.com/edvordo/docker-compose-watch-test/blob/main/docker-compose.yml) and running `docker compose watch` should result in a container named `app` and the folder `/var/www/html` should have the contents of the current directory.

Thing is, the directory is empty and only new changes to files are copied over. 

This lead me to believe, that I would need to add `COPY . /var/www/html` to `Dockerfile` so files are copied in on image build, but this creates a second problem, as any changes I made to project files since image build are not reflected to the container between container restarts (say across multiple days of development).

Anyway, here is a repository with an example setup, including steps to reproduce
https://github.com/edvordo/docker-compose-watch-test

I'm more than willing to concede I'm reading the docs wrong, but I've tried support channels like the (unofficial?) discord server and we could not figure out what could be wrong.

---
The issue is the same using the official examples provided here:
https://github.com/dockersamples/avatars

While yes, the containers will have content after first build (thanks to the `COPY` instruction), they will not have the changes made to the project files between container restarts:
- while watch is running, change `<title>` in `web/index.html` to whatever
- changes are reflected on `localhost:5735` and in container
- kill `watch`, run `docker compose down` and run `docker compose watch` again
- `<title>` in browser / container will ***NOT*** have your changes

### Compose Version

```Text
compose version
Docker Compose version v2.22.0-desktop.2

docker-compose version
Docker Compose version v2.22.0-desktop.2
```

### Docker Environment

```Text
Client:
 Version:    24.0.6
 Context:    desktop-linux
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.11.2-desktop.5
    Path:     /Users/edvordo/.docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.22.0-desktop.2
    Path:     /Users/edvordo/.docker/cli-plugins/docker-compose
  dev: Docker Dev Environments (Docker Inc.)
    Version:  v0.1.0
    Path:     /Users/edvordo/.docker/cli-plugins/docker-dev
[...]

PR #12047, opened 2024-08-06
Watch initial sync

jhrotko:
**What I did**
Introduce initial sync for `sync` and `sync+restart` actions in `develop.watch` configuration to solve issue when restarting watch - after a `docker compose down` - where the container no longer reflects previous changes. This behaviour is triggered when adding `x-initialSync: true` to a watch configuration. In order to make this work, it was needed to introduce `Extensions` to `develop.watch.trigger` in `compose-go` (check related PR below). 
 
The initial sync will evaluate if any file in `develop.watch.path` was modified after the image creation time. It also ignores files from `develop.watch.ignore` paths and binded volumes.
 
**Related issue**

compose-go
- https://github.com/compose-spec/compose-go/pull/672

fixes
- https://github.com/docker/compose/issues/11102

image

### DIFF
--- a/pkg/compose/watch.go
+++ b/pkg/compose/watch.go
@@ -19,6 +19,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"io/fs"
 	"os"
 	"path"
 	"path/filepath"
@@ -157,6 +158,13 @@ func (s *composeService) watch(ctx context.Context, syncChannel chan bool, proje
 			if checkIfPathAlreadyBindMounted(trigger.Path, service.Volumes) {
 				logrus.Warnf("path '%s' also declared by a bind mount volume, this path won't be monitored!\n", trigger.Path)
 				continue
+			} else {
+				var initialSync bool
+				success, err := trigger.Extensions.Get("x-initialSync", &initialSync)
+				if err == nil && success && initialSync && (trigger.Action == types.WatchActionSync || trigger.Action == types.WatchActionSyncRestart) {
+					// Need to check initial files are in container that are meant to be synched from watch action
+					s.initialSync(ctx, service, trigger, ignore, syncer)
+				}
 			}
 			paths = append(paths, trigger.Path)
 			pathLogs = append(pathLogs, fmt.Sprintf("Action %s for path %q", trigger.Action, trigger.Path))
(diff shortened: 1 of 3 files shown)


pkg/compose/watch.go
```diff
@@ -574,3 +585,114 @@ func (s *composeService) pruneDanglingImagesOnRebuild(ctx context.Context, proje
...
+			info, err := d.Info()
+			if err != nil {
+				// handle possible path err, just in case...
+				return err
+			}
+			if !d.IsDir() && info.ModTime().Before(timeImageCreated) {
+				// skip file if it was modified before image creation
+				return nil
+			}
+			pathsToCopy = append(pathsToCopy, sync.PathMapping{
```
jhrotko, line 650: maybe I should only copy files and not dirs?

pkg/compose/watch.go
```diff
@@ -574,3 +585,114 @@ func (s *composeService) pruneDanglingImagesOnRebuild(ctx context.Context, proje
 		}
 	}
 }
+
+func (s *composeService) initialSync(ctx context.Context, project *types.Project, service types.ServiceConfig, trigger types.Trigger, ignore watch.PathMatcher, syncer sync.Syncer) error {
+	dockerFileIgnore, err := watch.NewDockerPatternMatcher("/", []string{"Dockerfile", "*compose*.y*ml"})
```
jhrotko, line 590: is there a better way to handle the `compose.yaml` pattern? For instance, tried `compose.y[a]?ml` but it did not work :(

pkg/compose/watch.go
```diff
@@ -574,3 +585,123 @@ func (s *composeService) pruneDanglingImagesOnRebuild(ctx context.Context, proje
...
+}
+
+func shouldIgnore(name string, ignore watch.PathMatcher) bool {
+	shouldIgnore, _ := ignore.Matches(name)
+	// ignore files that match any ignore pattern
+	return shouldIgnore
+}
+
+// gets the image creation time for a service
+func (s *composeService) imageCreatedTime(ctx context.Context, project *types.Project, serviceName string) (time.Time, error) {
```
jhrotko, line 683: possibly can be moved somewhere else more generic

ndeloof (approved):

glours (approved):
LGTM

Issue #12063, opened 2024-08-16
[BUG] `docker compose wait` waits for ALL containers, not just the first


Raphtor:
### Description

In the [documentation](https://docs.docker.com/reference/cli/docker/compose/wait/), it is said that `docker compose wait SERVICE [SERVICE...]` waits for the first service to exit. However, it seems to wait for all the listed services to exit.

### Steps To Reproduce

1. With the compose file:
```yaml
services:
  service1:
    image: ubuntu:latest
    command: tail -f /dev/null
  service2:
    image: ubuntu:latest
    command: tail -f /dev/null
```
2. Run `docker compose up -d`
3. Run `docker compose wait service1 service2`
4. In another terminal, show the running containers `docker ps`:
```
CONTAINER ID   IMAGE           COMMAND               CREATED          STATUS         PORTS     NAMES
7d71dc7baa3a   ubuntu:latest   "tail -f /dev/null"   14 seconds ago   Up 1 second              service1-1
357917766e1e   ubuntu:latest   "tail -f /dev/null"   14 seconds ago   Up 2 seconds             service2-1
```
5. Kill one of the containers `docker kill 7d`
6. The wait command doesn't exit, but prints that the container was killed.
7. Kill the other container `docker kill 35`
8. The wait command prints that the second container was killed, and exits.

### Compose Version

```Text
Docker Compose version v2.29.1
```

### Docker Environment

```Text
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  0.11.2+azure-1
    Path:     /usr/libexec/docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.29.1
    Path:     /usr/libexec/docker/cli-plugins/docker-compose

Server:
 Containers: 8
  Running: 0
  Paused: 0
  Stopped: 8
 Images: 26
 Server Version: 23.0.6+azure-2
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: false
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 1
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
[...]

PR #12064, opened 2024-08-17
fix(wait): Wait only until first container exit

idsulik:
**What I did**
The doc says https://docs.docker.com/reference/cli/docker/compose/wait/ `Block until the first service container stops` but in reality it waits for all containers, changed logic to wait only for first container exit

**Related issue**
https://github.com/docker/compose/issues/12063

image

### DIFF
--- a/cmd/compose/wait.go
+++ b/cmd/compose/wait.go
@@ -43,7 +43,7 @@ func waitCommand(p *ProjectOptions, dockerCli command.Cli, backend api.Service) 
 	var err error
 	cmd := &cobra.Command{
 		Use:   "wait SERVICE [SERVICE...] [OPTIONS]",
-		Short: "Block until the first service container stops",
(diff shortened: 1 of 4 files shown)


ndeloof:
or just fix the doc

idsulik:
> or just fix the doc

but which one is correct? should we wait for all the containers all only first of the container?

ndeloof:
> but which one is correct?

to preserve backward compatibility we can't significantly change the existing behavior. 
Also one point to clarify is: would we want to wait for first container (in service) to stop / all containers in service to stop / one of the container in services to stop, but some service might not have completed, etc
If we want to cover those scenario we need to offer more control to the user, with additional flags
short terms, easier to just fix the doc

ndeloof:
👍 
please squash and sign off commits

idsulik:
done

ndeloof (approved):

Issue #12023, opened 2024-07-26
[BUG] docker-compose does not close unix socket with dockerd after container exits and restarts


Yinette:
### Description

While troubleshooting an issue that resulted in a dockerd crash, I have found that docker-compose does not close a unix socket with the docker daemon when a container exits and restarts.

This means that exponentially as containers restart more sockets are left open with dockerd, which spins up a new thread for each time a new one is opened, eventually leading dockerd to hit `max-threads` limit in the Kernel and crash.

In our case, the production workload has about 10 containers that are in an on-demand deployment style, so not all 10 need to be up and running depending on the pool of "devices" needing data processed upstream of them. Ones that aren't used will restart and re-query the database for a new endpoint to process input from.

While this approach of restarting containers is probably not the greatest, at the very least the unix sockets should be closed when finished.

From looking into this on the dockerd side, it appears that dockerd sends a `notifyClosed` via the socket when it exits, does docker compose handle that?

### Steps To Reproduce

1. create `docker-compose.yaml`
```yaml
---
services:
  occupied:
    image: container-used:latest
    logging:
       driver: "json-file"
       options:
         max-file: "5"
         max-size: "30m"
  
  surplus:
    restart: unless-stopped
    image: container-surplus:latest
    logging:
      driver: "json-file"
      options:
        max-file: "5"
        max-size: "30m"
```
2. create Dockerfile
```Dockerfile
FROM alpine:latest AS occupied

CMD ["sleep", "3600"]

FROM alpine:latest AS surplus

CMD ["sleep", "10"]
```

3. build images
```shell
docker build --target occupied -t container-used:latest .
docker build --target surplus -t container-surplus:latest .
```

4. start project
```shell
docker compose -f docker-compose.yaml --scale occupied=8 --scale surplus=2 up
```

5. concurrently with 4, observe total number of tasks/threads under dockerd:
```shell
watch -n1 "ls /proc/`pgrep dockerd`/task | wc -l"
```

6. concurrently with 4, observe unix socket fds under the `docker-compose` pid:
```shell
watch -n1 "lsof -p `pgrep docker-compose` | grep unix"
```

You will observe the following:
 - Docker compose will restart the exiting containers as they gracefully close.
 - In step 5, you will see the total number of threads under dockerd will begin to rise.
 - in step 6, you will see more unix sockets being added to the fd list for docker-compose process.

### Compose Version

```Text
Docker Compose version v2.29.1
```

### Docker Environment

```Text
[...]

PR #12112, opened 2024-09-10
attach: close streams when done

laurazard:
**What I did**

When Compose is watching a project/reattaching streams on container start, it will make new API `ContainerAttach()` calls every time a container it's watching is started. However, it only closes the stream when the context used to start the attach is canceled.

This means that if a user has a project with multiple containers where containers keep restarting, Compose will attach to the new containers but never close the previous streams, causing fds to pile up and goroutines on the engine to get stuck.

You can verify this by using a Dockerfile such as:
```Dockerfile
FROM alpine:latest AS occupied

CMD ["sleep", "3600"]

FROM alpine:latest AS surplus

CMD ["sleep", "10"]
```

And the following Compose file:
```yaml
services:
  occupied:
    image: container-used:latest
  
  surplus:
    restart: unless-stopped
    image: container-surplus:latest
```

Run
```console
docker compose up --scale occupied=8 --scale surplus=2`.
```

In another terminal, run 
```console
watch -n1 "lsof -p `pgrep docker-compose` | grep unix"
```

And gradually watch the connections pile up.

(There might also be some bug engine-side here, I haven't looked into it yet.)

**Related issue**

fixes https://github.com/docker/compose/issues/12023

### DIFF
--- a/pkg/compose/attach.go
+++ b/pkg/compose/attach.go
@@ -102,9 +102,7 @@ func (s *composeService) attachContainer(ctx context.Context, container moby.Con
 
 func (s *composeService) attachContainerStreams(ctx context.Context, container string, tty bool, stdin io.ReadCloser, stdout, stderr io.WriteCloser) (func(), chan bool, error) {
 	detached := make(chan bool)
-	var (
-		restore = func() { /* noop */ }
-	)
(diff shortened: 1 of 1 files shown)


glours (approved):
LGTM

PR #12115, opened 2024-09-11
Bump engine cli v27.2.1

glours:
**What I did**
Bump dependency versions:
- Engine and Cli v27.2.1
- containerd v1.7.22
- buildx v0.17.0
- buildkit v0.16.0
- Golang 1.22.7

**Related issue**


go.mod
```diff
@@ -1,23 +1,23 @@
 module github.com/docker/compose/v2
 
-go 1.21.0
+go 1.22.0
```
thaJeztah, line 3: Did it automatically update this one, or did it still allow go 1.21.0 here?

I generally try to treat this one the same as other dependencies; list the minimum required version, and only update if it's _impossible_ to use with older versions; see https://github.com/containerd/containerd/pull/10596#discussion_r1721294997

go.mod
```diff
@@ -1,23 +1,23 @@
...
 
-toolchain go1.22.2
+toolchain go1.22.7
```
thaJeztah, line 5: mostly curious; have you tried if it allows removing this one? I know it's sometimes added automatically, but if you remove it, it doesn't add it back

ndeloof (approved):
LGTM. Wonder go version bump won't bring issue to release Linux system packages and need to align with engine

thaJeztah:
> Wonder go version bump won't bring issue to release Linux system packages and need to align with engine

Engine moved to go1.22 as well, so no problem there; but left the two comments above to not aggressively _enforce_ (if possible)


(LGTM otherwise)

Dockerfile
```diff
@@ -15,7 +15,7 @@
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
 
-ARG GO_VERSION=1.21.12
+ARG GO_VERSION=1.22.7
 ARG XX_VERSION=1.2.1
 ARG GOLANGCI_LINT_VERSION=v1.55.2
```
thaJeztah, line 20: Oh! You probably want to update GOLANGCI_LINT_VERSION as well for compatibility with newer go versions (CI still seemed happy, so maybe it still works); I think 1.60.2 is the current version, but you may run into linters complaining, so if it works, you could do it separately; see https://github.com/moby/moby/pull/48359

ndeloof:
If cli/engine moved to 1.22 then I'd prefer we do the same, even code technically would be happy with 1.21

Dockerfile
thaJeztah: The `poll.Continue(r.Combined())` ones will be addressed by https://github.com/gotestyourself/gotest.tools/pull/279, but in the meantime you can use;

```go
poll.Continue("%v", r.Combined())
```

thaJeztah:
We still keep the version in go.mod / vendor.mod at go1.21.0; https://github.com/moby/moby/blob/34a2f953b13dd06cf09a457980f58fc30976e0ba/vendor.mod#L8

Same for BuildKit and Buildx; 
https://github.com/moby/buildkit/blob/3a7055008a5e58a2abbe0e0c21c919d9e014e062/go.mod#L3
https://github.com/docker/buildx/blob/40f444f4b80f1c5e76f99517f01dbd6cfeec7ad8/go.mod#L3

thaJeztah:
@glours I opened https://github.com/glours/compose/pull/44 against your PR branch (feel free to squash or change order of commits)

thaJeztah (approved):
LGTM (if CI is happy)

Issue #10774, opened 2023-07-05
[BUG] Renew Anon Volumes doesn't work as expected compared to v1


ryanneufeld:
### Description

`-V, --renew-anon-volumes        Recreate anonymous volumes instead of retrieving data from the previous containers.`

This functions differently than in V1

### Steps To Reproduce

With docker-compose v1 it was possible to run `docker-compose up -dV` and have all the containers restarted with fresh anonymous volumes. 

Currently this no longer works expected and `up` simply updates any containers with new versions fetched via `pull`

```yaml
version: "3"
services:
  mysql:
    restart: unless-stopped
    image: mysql:5.7
    ports:
      - "3306:3306"
    expose:
      - "3306"
    environment:
      MYSQL_ALLOW_EMPTY_PASSWORD: "yes"
```

### Compose Version

```Text
Docker Compose version v2.18.1
```

### Docker Environment
MacOS Ventura 13.4.1 (22F82)
```Text
Client:
 Version:    24.0.2
 Context:    desktop-linux
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.10.5
    Path:     /Users/rneufeld/.docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.18.1
    Path:     /Users/rneufeld/.docker/cli-plugins/docker-compose
  dev: Docker Dev Environments (Docker Inc.)
    Version:  v0.1.0
    Path:     /Users/rneufeld/.docker/cli-plugins/docker-dev
  extension: Manages Docker extensions (Docker Inc.)
    Version:  v0.2.19
    Path:     /Users/rneufeld/.docker/cli-plugins/docker-extension
  init: Creates Docker-related starter files for your project (Docker Inc.)
    Version:  v0.1.0-beta.4
    Path:     /Users/rneufeld/.docker/cli-plugins/docker-init
  sbom: View the packaged-based Software Bill Of Materials (SBOM) for an image (Anchore Inc.)
    Version:  0.6.0
    Path:     /Users/rneufeld/.docker/cli-plugins/docker-sbom
  scan: Docker Scan (Docker Inc.)
    Version:  v0.26.0
    Path:     /Users/rneufeld/.docker/cli-plugins/docker-scan
  scout: Command line tool for Docker Scout (Docker Inc.)
    Version:  v0.12.0
    Path:     /Users/rneufeld/.docker/cli-plugins/docker-scout

Server:
[...]

PR #12116, opened 2024-09-11
Restore compose v1 behavior to recreate containers when ran with -V

ndeloof:
**What I did**
Compose v1 made it implicit for `--renew-anon-volumes` to imply `--force-recreate`. Indeed if we don't do,existing container that doesn't need to be re-created for other reasons won't get a fresh new volume attached?

**Related issue**
closes https://github.com/docker/compose/issues/10774


glours (approved):
LGTM

PR #12118, opened 2024-09-11
chore(watch): Add changed file paths/count to log

idsulik:
**What I did**
Reverted changed file paths/count to the watch logs.  For some reason @ndeloof deleted it from the log, but I think it's pretty useful to see what's going on

image


ndeloof:
I removed this from logs as in many case this brings a huge amount of noise in the application logs

idsulik:
> I removed this from logs as in many case this brings a huge amount of noise in the application logs
Suggest to show files to sync only `if logrus.IsLevelEnabled(logrus.DebugLevel)`. 



I use such feature in skaffold and it's pretty useful to see what's going on, but I can control when to show it(if log level is "info")

idsulik:
pushed commit

ndeloof (approved):

glours (approved):
LGTM 👍

Issue #12103, opened 2024-09-05
[BUG] Deadlock on recreation of logging container


smoehrle:
### Description

See example below. We have a fluentbit container and an app container which uses the fluentbit container in its logging configuration. If there is anything which triggers a recreation of the fluentbit container, the following happens:

- Fluentbit is recreated, waits with status created, is not running
- Depends-on triggers a recreation of the app which tries to terminate, but cannot since the log driver is blocking the termination (can only terminate when fluentbit is available)

**Workaround**: run `docker start [project-name]-fluentbit-1` in a second terminal

**Expected behaviour**: No deadlock

### Steps To Reproduce

```
# compose.yml
services:
  fluentbit:
    image: fluent/fluent-bit:3.1.7-debug
    ports:
      - "24224:24224"
      - "24224:24224/udp"
    environment:
      FOO: ${BAR}

  app:
    image: nginx
    depends_on:
      - fluentbit
    logging:
      driver: fluentd
      options:
        fluentd-address: 127.0.0.1:24224
```
```
# test.txt
BAR=test
```

1. docker compose up -d
2. docker compose --env-file test.txt up -d
    - Second command can be anything that triggers a recreate of the fluentbit container. 

### Compose Version

```Text
Docker Compose version v2.24.5
```

### Docker Environment

```Text
Client: Docker Engine - Community
 Version:    25.0.2
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.12.1
    Path:     /usr/libexec/docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.24.5
    Path:     /usr/libexec/docker/cli-plugins/docker-compose

Server:
 Containers: 373
  Running: 3
  Paused: 0
  Stopped: 370
 Images: 770
 Server Version: 25.0.2
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 runc runsc
 Default Runtime: runc
 Init Binary: docker-init
[...]

PR #12122, opened 2024-09-13
stop dependent containers before recreating diverged service

ndeloof:
**What I did**
As we detect service container has to be recreated, stop dependent containers so they don't miss dependency.
This can be a blocker when service is actually a logging driver, as stopping it for recreation just lock the docker engine, as running dependent service still tries to publish logs.

We also used to fully recreate such dependent containers, while they just need to restart

**Related issue**
fixes https://github.com/docker/compose/issues/12103

https://docker.atlassian.net/browse/COMP-750

image

### DIFF
--- a/pkg/compose/convergence.go
+++ b/pkg/compose/convergence.go
@@ -45,9 +45,6 @@ import (
 )
 
 const (
-	extLifecycle  = "x-lifecycle"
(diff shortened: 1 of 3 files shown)


glours (approved):
LGTM

Issue #12140, opened 2024-09-20
[BUG] Source directory not auto-created with docker-compose v2.29.3 volumes in Bitbucket Pipelines


iamvinov-atlassian:
### Description

When using `volumes` to bind mount non-existing source directories, the directories are not automatically created in docker-compose v2.29.3 in Bitbucket Pipelines. According to the docker volumes [documentation](https://docs.docker.com/engine/storage/bind-mounts/#differences-between--v-and---mount-behavior), when using the `--volumes` flag, the directories should be created (I can confirm this was the case with the `v2.29.2`). Though the documentation is about Docker CLI, I imagine this behaviour should extend to docker-compose files since, as mentioned before, this was the behaviour with `v2.29.2`. 

### Steps To Reproduce

Consider this docker-compose file:
```
services:
  localstack:
    image: alpine
    command: [ "echo", "hi" ]
    ports:
      - "4566"
    volumes:
      - "./to-be-created-directory:/dot"
```

This `bitbucket.yml` file: 
```
pipelines:
  default:
    - step:
        name: Bind mount test
        services:
          - docker
        script:
          - echo "Installing Docker Compose v2"
          - source ./bin/install-docker-compose-v2.sh
          - docker info
          - docker compose version
          - docker compose up
```

The error you get will be:
```
Error response from daemon: invalid mount config for type "bind": bind source path does not exist: /opt/atlassian/pipelines/agent/build/to-be-created-directory
```

### Compose Version
```
Docker Compose version v2.29.3
```

### Docker Environment

```
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  compose: Docker Compose (Docker Inc., v2.29.3)
Server:
 Containers: 0
  Running: 0
  Paused: 0
  Stopped: 0
 Images: 0
 Server Version: 25.0.3
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 1
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 7c3aca7a610df76212171d200ca3811ff6096eb8
[...]

PR #12141, opened 2024-09-20
revert commits link to mount API over bind changes

### LINKED ISSUE #12139: [BUG] Unable to bind mount when mount source is within the daemon root
### Description

After upgrading Compose v2.29.6 from v2.29.2, the following service began failing:

```
services:
  cadvisor:
    image: gcr.io/cadvisor/cadvisor:latest
    privileged: true
    volumes:
      - /:/rootfs:ro
      - /var/run:/var/run:ro
      - /sys:/sys:ro
      - /var/lib/docker/:/var/lib/docker:ro
      - /dev/disk/:/dev/disk:ro
```

Running `docker compose up -d` gives the following error message:

```
Error response from daemon: invalid mount config: must use either propagation mode "rslave" or "rshared" when mount source is within the daemon root, daemon root: "/var/lib/docker", bind mount source: "/", propagation: "rprivate"
```

### Steps To Reproduce

_No response_

### Compose Version

```Text
Docker Compose version v2.29.6
```

### Docker Environment

```Text
 Version:    27.3.0
 Context:    default
 Debug Mode: false
 Plugins:
    Version:  v0.17.1
    Version:  v2.29.6

Server:
 Containers: 0
  Running: 0
  Paused: 0
  Stopped: 0
 Images: 1
 Server Version: 26.1.3
 Storage Driver: vfs
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
 Swarm: inactive
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 8b3b7ca2e5ce38e8f31a34f35b2b68ceb8470d89
 runc version: v1.1.14-0-g2c9f560
 init version: de40ad0
 Security Options:
  seccomp
   Profile: builtin
  cgroupns
 Kernel Version: 6.6.16-linuxkit
 Operating System: Ubuntu 20.04.6 LTS
 OSType: linux
 Architecture: aarch64
 CPUs: 12
 Total Memory: 7.657GiB
 Name: d4d8c60ee1fe
 ID: 186ebd80-8c59-49b8-92ae-007c1c620318
 Debug Mode: false
 Experimental: false
 Insecure Registries:
  127.0.0.0/8
 Live Restore Enabled: false
```

### Anything else?

_No response_

glours:
**What I did**
I reverted the following commits:

- c16df17e1f2022c907a68b010f380bce53ac3e36
- cb00aaad288ff83ed98835a1edea635800f5b083
- f9c7a0cc085eba94d6140dcae18f16d7ef5a7691

**Related issue**

fixes #12139 #12140

image

### DIFF
--- a/pkg/compose/compose.go
+++ b/pkg/compose/compose.go
@@ -321,23 +321,6 @@ func (s *composeService) RuntimeVersion(ctx context.Context) (string, error) {
 
 }
 
-var windowsContainer = struct {
-	once sync.Once
-	val  bool
-	err  error
-}{}
-
-func (s *composeService) isWindowsContainer(ctx context.Context) (bool, error) {
-	windowsContainer.once.Do(func() {
-		info, err := s.apiClient().Info(ctx)
-		if err != nil {
-			windowsContainer.err = err
-		}
-		windowsContainer.val = info.OSType == "windows"
-	})
-	return windowsContainer.val, windowsContainer.err
-}
-
 func (s *composeService) isDesktopIntegrationActive() bool {
 	return s.desktopCli != nil
 }
--- a/pkg/compose/convergence_test.go
+++ b/pkg/compose/convergence_test.go
@@ -28,6 +28,7 @@ import (
 	containerType "github.com/docker/docker/api/types/container"
 	"github.com/docker/docker/api/types/filters"
 	"github.com/docker/docker/api/types/network"
+	"github.com/docker/go-connections/nat"
 	"go.uber.org/mock/gomock"
 	"gotest.tools/v3/assert"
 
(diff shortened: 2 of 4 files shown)


ndeloof (approved):
👍 
This experiment demonstrate `Mount` and `Bind` API are not equivalent, as mount require client to know about engine constraints (typically "_must use either propagation mode "rslave" or "rshared" when mount source is within the daemon root_").
compse-go makes the assumption a bind can be safely expanded in to a mount, so the "long syntax", but actually this is wrong. We will have to manage this in compose-go in a better way, to really distinguish user intent to use mount-specific options. Maybe we should introduce a `mount` section and deprecate use of advanced volume attributes (like `subpath`) which are only supported by the `mount` API, to fully align with the docker CLI

thaJeztah:
@ndeloof IIUC this worked on a bare Linux machine, but failed with desktop? Do we know what Desktop did that made it fail?

prashant-shahi:
The previous release failed in my Linux machine.

thaJeztah:
@prashant-shahi Yes; the long story was that;

1. https://github.com/docker/compose/commit/f9c7a0cc085eba94d6140dcae18f16d7ef5a7691 switched to using a different API (to replace a legacy API)
2. This caused an issue with Docker Desktop, which had some code-path related to mapping host <--> VM mounts that got triggered if no bind-propagation was set https://github.com/docker/compose/commit/cb00aaad288ff83ed98835a1edea635800f5b083
3. Which caused an issue for native Windows containers, because windows has no concept of "bind propagation"; so it was made conditional https://github.com/docker/compose/commit/c16df17e1f2022c907a68b010f380bce53ac3e36

Step `2.` chose to use `rprivate`, as it's the default that would be set by the daemon, so the assumption was that setting the same default on the client-side (compose) would be OK, but that default was not always applicable; a special exception existed on the daemon side to use `rshared` as a default when bind-mounting Docker's own storage location (`/var/lib/docker` by default). That special case is _NOT_ handled when the user ("compose" in this case) explicitly sets a propagation mode. It's also quite likely that exception was never documented (other than "in code"), and bind-mounting `/var/lib/docker` is generally not recommended and more a niche use-case (at least compose did not have tests for this). But for those that _did_ use this, an error was now produced by the daemon, which has a safeguard in place if an incorrect propagation is set;

```
Error response from daemon: invalid mount config: must use either propagation mode "rslave" or "rshared" when mount source is within the daemon root, daemon root: "/var/lib/docker", bind mount source: "/", propagation: "rprivate"
```

This PR reverts all of the above, pending changes in Docker Desktop to allow this again.

Issue #8305, opened 2021-04-30
Windows: missing secrets file creates file-named directory


philliphoff:
## Description of the issue

On Windows, when starting a composition that refers to a secret file which does not exist, an "invalid mount config for type" error is generated and a directory with the same name as the file is created.

This behavior is *not* seen on Mac, with the same version of Docker Desktop and Docker Compose.

This may be related to #5377 but seems only to affect Windows.

## Context information (for bug reports)

**Output of `docker-compose version`**
```
docker-compose version 1.29.0, build 07737305
docker-py version: 5.0.0
CPython version: 3.9.0
OpenSSL version: OpenSSL 1.1.1g  21 Apr 2020
```

**Output of `docker version`**
```
Client: Docker Engine - Community
 Cloud integration: 1.0.12
 Version:           20.10.5
 API version:       1.41
 Go version:        go1.13.15
 Git commit:        55c4c88
 Built:             Tue Mar  2 20:14:53 2021
 OS/Arch:           windows/amd64
 Context:           default
 Experimental:      true

Server: Docker Engine - Community
 Engine:
  Version:          20.10.5
  API version:      1.41 (minimum version 1.12)
  Go version:       go1.13.15
  Git commit:       363e9a8
  Built:            Tue Mar  2 20:15:47 2021
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.4.4
  GitCommit:        05f951a3781f4f2c1911b05e61c160e9c30eaa8e
 runc:
  Version:          1.0.0-rc93
  GitCommit:        12644e614e25b05da6fd08a38ffa0cfe1903fdec
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

**Output of `docker-compose config`**
(Make sure to add the relevant `-f` and other flags)
```
D:\Repos\Scratch\compose-test>docker-compose config
secrets:
  postgres_db:
    file: D:\Repos\Scratch\compose-test\postgres_db.txt
  postgres_password:
    file: D:\Repos\Scratch\compose-test\postgres_password.txt
  postgres_user:
    file: D:\Repos\Scratch\compose-test\postgres_user.txt
services:
  db:
    environment:
      POSTGRES_DB_FILE: /run/secrets/postgres_db
      POSTGRES_INITDB_ARGS: --auth=md5
      POSTGRES_PASSWORD_FILE: /run/secrets/postgres_password
      POSTGRES_USER_FILE: /run/secrets/postgres_user
[...]

PR #12151, opened 2024-09-23
check secret source exists, as bind mount would create target

ndeloof:
**What I did**
Check secret.file exists otherwise target will be silently created as a folder by bind mount

**Related issue**
fixes https://github.com/docker/compose/issues/8305

### DIFF
--- a/pkg/compose/create.go
+++ b/pkg/compose/create.go
@@ -1003,11 +1003,18 @@ func buildContainerSecretMounts(p types.Project, s types.ServiceConfig) ([]mount
 			continue
 		}
 
+		if _, err := os.Stat(definedSecret.File); os.IsNotExist(err) {
+			logrus.Warnf("secret file %s does not exist", definedSecret.Name)
+		}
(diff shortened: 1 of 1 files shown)


thaJeztah:
This will break uses of compose against a remote daemon (where the secret is stored remotely)

ndeloof:
@thaJeztah good point. I would expect anyway `secrets` and `config` are only used when data is set by client, otherwise why not just declare a plain bind mount ?
Also see https://github.com/docker/compose/issues/11867 as a request to set those from client

In the meantime, adopted here an alternate approach to just warn user, and configure bind with `CreateHostPath: false`

thaJeztah:
Ah! Yes I wanted to suggest the `CreateHostPath`, but thought this was using the old API (and we couldn't use the new one).

Yup! That's better

thaJeztah (approved):
LGTM

glours (approved):
LGTM

Issue #12134, opened 2024-09-17
[BUG] --wait-timeout option not working with 'service_completed_successfully' dependency


norbert-balazs:
### Description

Using the --wait-timeout option with --wait is supposed to ensure that the command doesn't hang indefinitely in case of issues. However the command still hangs when the compose file contains a dependency with a `service_completed_successfully` condition.

The --timeout option doesn't seem to be working either in this scenario.

This bug was found in a real compose configuration when one of our setup containers got stuck. Below is just a minimal reproduction.

The expected behavior would be that the `docker compose` command exits after the specified `wait-timeout`.

### Steps To Reproduce

1. Use the following compose file:
```
services:
  long_setup:
    image: alpine:latest
    #command: "sh -c 'while true; do sleep 1; done'"
    command: "sh -c 'sleep 10; exit 0'"
  server:
    image: alpine:latest
    command: "sh -c 'while true; do sleep 1; echo 'running'; done'"
    depends_on:
      long_setup:
        condition: service_completed_successfully
```
2. Run it with `docker compose up -d --wait --wait-timeout 1`, or any timeout less than 10.
3. Observe the command not failing in the defined timeout.
4. Change to the commented out command containing the infinite loop
5. Run `docker compose up -d --wait --wait-timeout 1` again.
6. Observe the command running indefinitely.

Run the above with `docker compose up -d --timeout 1` or `docker compose up --timeout 1`, and observe that the --timeout option doesn't work either.

### Compose Version

```Text
Docker Compose version v2.29.2
```

### Docker Environment

```Text
Client: Docker Engine - Community
 Version:    27.2.1
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.16.2
    Path:     /usr/libexec/docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.29.2
    Path:     /usr/libexec/docker/cli-plugins/docker-compose

Server:
 Containers: 2
  Running: 1
  Paused: 0
  Stopped: 1
 Images: 153
 Server Version: 27.2.1
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
[...]

PR #12156, opened 2024-09-23
wait for dependent service up to delay set by --wait-timeout

ndeloof:
**What I did**
Apply --wait-timeout delay while waiting for dependencies, so that process don't get stuck waiting.

**Related issue**
fixes https://github.com/docker/compose/issues/12134

image


glours (approved):
LGTM

Issue #12163, opened 2024-09-25
[BUG] Inconsistency when running $ docker compose -p <project_name> down <list_services>


jefersoonaf:
### Description

When running the command `$ docker compose -p <project_name> down <list_services>` , it removes all services created with the same <project_name> when it does not find any service that was informed in <list_services>.

### Steps To Reproduce

Assuming I have two docker-compose.yml files in different directories:
`/app1/docker-compose.yml` with services: `app1`, `nginx_app1`:
```yaml
services:
  app1:
    image: app1
    expose:
      - "80"
  nginx_app1:
    image: nginx
    ports:
      - "8080:8080"
```
`/app2/docker-compose.yml` with services: `app2`, `nginx_app2`:
```yaml
services:
  app2:
    image: app2
    expose:
      - "80"
  nginx_app2:
    image: nginx
    ports:
      - "8081:8081"
```

And I performed the up of them separately but using the same project_name:
```bash
$ cd /app1
$ docker compose -p myproject up -d app1 nginx_app1
```
```bash
$ cd /app2
$ docker compose -p myproject up -d app2 nginx_app2
```

After that, if I run: 
```bash
$ docker compose -p myproject ps
```
I will see all 4 services running.

And if I want to stop only the services of app1, I can naturally do:
```bash
$ cd /app1
$ docker compose -p myproject down app1 nginx_app1
```

However, this works at first because there were services running with the names `app1 `and `nginx_app1`, but if I run this same command a second time, docker ends the services of `app2`. Why? Since I explicitly informed the services of `app1` in the command?

I only noticed that this happens when the command no longer finds any services related to app1 running. But in this case, the expected behavior would be for docker to inform that it did not find any services with the names informed for them to be stopped.

### Compose Version

```Text
Docker Compose version v2.29.7
```

### Docker Environment

```Text
Client: Docker Engine - Community
 Version:    25.0.2
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.12.1
    Path:     /usr/libexec/docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.29.7
    Path:     /usr/libexec/docker/cli-plugins/docker-compose

Server:
 Containers: 12
  Running: 12
  Paused: 0
  Stopped: 0
 Images: 7
 Server Version: 26.1.4
 Storage Driver: overlay2
  Backing Filesystem: extfs
[...]

PR #12164, opened 2024-09-26
fix(down): Fix down command if specified services are not running

idsulik:
**What I did**
Fixed down command, it passes empty services if specified services are not running, in this case we stops all services despite command line argument where we specify list of services we want to stop

**Related issue**
[](https://github.com/docker/compose/issues/12163)

image

### DIFF
--- a/pkg/compose/down.go
+++ b/pkg/compose/down.go
@@ -67,10 +67,17 @@ func (s *composeService) down(ctx context.Context, projectName string, options a
 	}
 
 	// Check requested services exists in model
-	options.Services, err = checkSelectedServices(options, project)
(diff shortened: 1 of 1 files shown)


glours (approved):
LGTM, thanks @idsulik 🙏

ndeloof (commented):
We try to avoid mock-based tests vs e2e tests, as those based on mocks are a pain to maintain, while e2e are way more explicit and can be easily ran with plain command line for experiment - (even they are slower to execute)

ndeloof (approved):

idsulik:
> We try to avoid mock-based tests vs e2e tests,

I'll keep it in mind, thank you

PR #12166, opened 2024-09-26
introduce service hooks

ndeloof:
**What I did**

Added support for `post_start` and `pre_stop` lifecycle hooks to services

demo: https://asciinema.org/a/Oh1ZFu2IERd3Fo4nK6lcSAhwS

**Related issue**

image

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,7 +7,7 @@ require (
 	github.com/Microsoft/go-winio v0.6.2
 	github.com/acarl005/stripansi v0.0.0-20180116102854-5a71ef0e047d
 	github.com/buger/goterm v1.0.4
-	github.com/compose-spec/compose-go/v2 v2.2.1-0.20241003145835-48d3a5bbf4ea
+	github.com/compose-spec/compose-go/v2 v2.2.1-0.20241007090213-a59035ad2bf4
 	github.com/containerd/containerd v1.7.22
 	github.com/containerd/platforms v0.2.1
 	github.com/davecgh/go-spew v1.1.1
--- a/pkg/compose/down.go
+++ b/pkg/compose/down.go
@@ -85,7 +85,8 @@ func (s *composeService) down(ctx context.Context, projectName string, options a
 
 	err = InReverseDependencyOrder(ctx, project, func(c context.Context, service string) error {
 		serviceContainers := containers.filter(isService(service))
-		err := s.removeContainers(ctx, serviceContainers, options.Timeout, options.Volumes)
+		serv := project.Services[service]
+		err := s.removeContainers(ctx, serviceContainers, &serv, options.Timeout, options.Volumes)
 		return err
 	}, WithRootNodesAndDown(options.Services))
 	if err != nil {
@@ -94,7 +95,7 @@ func (s *composeService) down(ctx context.Context, projectName string, options a
 
 	orphans := containers.filter(isOrphaned(project))
 	if options.RemoveOrphans && len(orphans) > 0 {
-		err := s.removeContainers(ctx, orphans, options.Timeout, false)
+		err := s.removeContainers(ctx, orphans, nil, options.Timeout, false)
 		if err != nil {
 			return err
 		}
@@ -296,9 +297,19 @@ func (s *composeService) removeVolume(ctx context.Context, id string, w progress
 	return err
 }
 
-func (s *composeService) stopContainer(ctx context.Context, w progress.Writer, container moby.Container, timeout *time.Duration) error {
+func (s *composeService) stopContainer(ctx context.Context, w progress.Writer, service *types.ServiceConfig, container moby.Container, timeout *time.Duration) error {
 	eventName := getContainerProgressName(container)
 	w.Event(progress.StoppingEvent(eventName))
+
+	if service != nil {
+		for _, hook := range service.PreStop {
+			err := s.runHook(ctx, container, *service, hook, nil)
+			if err != nil {
+				return err
+			}
+		}
+	}
+
 	timeoutInSecond := utils.DurationSecondToInt(timeout)
 	err := s.apiClient().ContainerStop(ctx, container.ID, containerType.StopOptions{Timeout: timeoutInSecond})
 	if err != nil {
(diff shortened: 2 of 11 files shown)


go.mod
```diff
@@ -193,4 +193,4 @@ require (
 	tags.cncf.io/container-device-interface v0.8.0 // indirect
 )
 
-replace github.com/compose-spec/compose-go/v2 => github.com/compose-spec/compose-go/v2 v2.2.1-0.20240926141145-d3fd7d94aa31
+replace github.com/compose-spec/compose-go/v2 => github.com/ndeloof/compose-go/v2 v2.0.1-0.20240924090029-29a100a62f5f
```
jhrotko, line 196: leaving this comment here not to forget to delete this after compose-spec is bumped

pkg/compose/down.go
```diff
@@ -296,9 +297,19 @@ func (s *composeService) removeVolume(ctx context.Context, id string, w progress
...
 	eventName := getContainerProgressName(container)
 	w.Event(progress.StoppingEvent(eventName))
+
+	if service != nil {
+		for _, hook := range service.PreStop {
+			err := s.runHook(ctx, container, *service, hook, nil)
+			if err != nil {
+				return err
+			}
+		}
```
jhrotko, line 310: We are always iterating over all hooks for PreStop, PostStart, etc.
I think it would make sense to add a function that encapsulates this in `hook.go`.
```
func (s composeService) ExecuteHooks(ctx context.Context, container moby.Container, service types.ServiceConfig, listener api.ContainerEventListener, hooks []types.ServiceHook) error {
	for _, hook := range hooks {
		err := s.runHook(ctx, container, service, hook, listener)
		if err != nil {
			return err
		}
	}
	return nil
}
```

pkg/compose/down.go
```diff
@@ -85,7 +85,8 @@ func (s *composeService) down(ctx context.Context, projectName string, options a
 
 	err = InReverseDependencyOrder(ctx, project, func(c context.Context, service string) error {
 		serviceContainers := containers.filter(isService(service))
-		err := s.removeContainers(ctx, serviceContainers, options.Timeout, options.Volumes)
+		serv := project.Services[service]
```
jhrotko, line 88: In the end we use the service and not the pointer (just like in `startContainer`) if the only reason to do this is just to make sure the service is defined, might as well validate here 
``` 
serv, ok := project.Services[service] 
if !ok {
  return fmt.Error() ....
}
```

ndeloof: as we run inside `InReverseDependencyOrder` service is guaranteed to exist (actually, `InReverseDependencyOrder` could pass a `*ServiceConfig`, but this could have impact on immutability)

`service` is passed as a pointer down to https://github.com/docker/compose/blob/0aacddfb1d21740e941db2881a58cfcd39c16067/pkg/compose/down.go#L304, as this same method is used to also stop orphaned container, where we don't have a defined service so need the ability to pass `nil`

pkg/compose/stop.go
```diff
@@ -54,6 +54,7 @@ func (s *composeService) stop(ctx context.Context, projectName string, options a
 		if !utils.StringContains(options.Services, service) {
 			return nil
 		}
-		return s.stopContainers(ctx, w, containers.filter(isService(service)).filter(isNotOneOff), options.Timeout)
+		serv := project.Services[service]
```
jhrotko, line 57: (repeat)
We could add a validation that the service exists here. Since this is the place where we get its value becoming easier later on for debugging.

serv, ok := project.Services[service] 
if !ok {
  return fmt.Error() ....
}

pkg/compose/hook.go
```diff
@@ -0,0 +1,122 @@
...
+	// We miss a ContainerExecWait API
+	tick := time.NewTicker(100 * time.Millisecond)
+	for {
+		select {
+		case <-ctx.Done():
+			return nil
+		case <-tick.C:
+			inspect, err := s.apiClient().ContainerExecInspect(ctx, exec.ID)
+			if err != nil {
+				return nil
```
jhrotko, line 112: Might have missed something, but shouldn't we return the error in this case?

pkg/compose/down.go
```diff
@@ -309,32 +320,32 @@ func (s *composeService) stopContainer(ctx context.Context, w progress.Writer, c
 	return nil
 }
 
-func (s *composeService) stopContainers(ctx context.Context, w progress.Writer, containers []moby.Container, timeout *time.Duration) error {
+func (s *composeService) stopContainers(ctx context.Context, w progress.Writer, serv *types.ServiceConfig, containers []moby.Container, timeout *time.Duration) error {
```
jhrotko, line 323: nit: rename to `serc` -> `service` for consistency

pkg/compose/hook.go
```diff
@@ -0,0 +1,122 @@
...
+	moby "github.com/docker/docker/api/types"
+	containerType "github.com/docker/docker/api/types/container"
+	"github.com/docker/docker/pkg/stdcopy"
+)
+
+func (s composeService) runHook(ctx context.Context, container moby.Container, service types.ServiceConfig, hook types.ServiceHook, listener api.ContainerEventListener) error {
+	wOut := utils.GetWriter(func(line string) {
+		listener(api.ContainerEvent{
+			Type:      api.HookEventLog,
+			Container: getContainerNameWithoutProject(container) + " ->",
```
jhrotko, line 37: I would propose to change the `->` to the name of the hook + index? we could pass it as a variable. If all hooks have the "signature" `->` might be complicated to understand from which hook the log comes from

jhrotko:
Really cool stuff @ndeloof 😊

pkg/compose/hook.go
ndeloof: hooks have no name so far (to be debated)
I wanted here to distinguish hook output from main logs without making the prefix a looooooong line, as this will remain for the whole compose execution after hook completed

jhrotko: how about we choose an emoji for each hook? 😆

ndeloof: My first implementation was using 🪝 but this won't run well on a Windows terminal

glours (approved):
Sounds good to me 👍 
Glad we finally have the first implementation of hooks mechanism 🥳

Issue #12167, opened 2024-09-27
[BUG] failed to read expected number of bytes: unexpected EOF


marverix:
### Description

Publishing to docker.io does not work, I'm getting:
```
❯ docker compose -f src/redis.yaml alpha publish docker.io/corporatedataleague/compose-redis:1.0.0
[+] Pushing 1/0
 ✔ redis Skipped                                                                                                                                                                                                                                        0.0s 
WARN[0000] reference for unknown type: application/vnd.docker.compose.file+yaml 
[+] Publishing 0/1
[+] Publishing 1/1ratedataleague/compose-redis:1.0.0 publishing                                                                                                                                                                                         1.1s 
 ✘ docker.io/corporatedataleague/compose-redis:1.0.0 publishing                                                                                                                                                                                         1.5s 
failed to read expected number of bytes: unexpected EOF
```

### Steps To Reproduce

1. Create `src/redis.yaml` file:

```
services:
  redis:
    image: redis:6.2
    ports:
      - 6379:6379
```

2. Login to Docker Hub
3. Try to publish:

```
docker compose -f src/redis.yaml alpha publish docker.io/corporatedataleague/compose-redis:1.0.0
```

### Compose Version

```Text
Docker Compose version v2.29.2
```


### Docker Environment

```Text
Client: Docker Engine - Community
 Version:    27.2.1
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.16.2
    Path:     /usr/libexec/docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.29.2
    Path:     /usr/libexec/docker/cli-plugins/docker-compose

Server:
 Containers: 8
  Running: 4
  Paused: 0
  Stopped: 4
 Images: 56
 Server Version: 27.2.1
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 runc
 Default Runtime: runc
 Init Binary: docker-init
[...]

PR #12169, opened 2024-09-27
fix(push): Fix unexpected EOF on alpha publish

idsulik:
**What I did**
Initialized `Data` slice for `v1.DescriptorEmptyJSON` to avoid `unexpected EOF`.  
Here `docker/buildx/util/imagetools/create.go:210` we pass `bytes.NewReader(dt)`, but the `dt` is nil(`internal/ocipush/push.go:182`), because of that it can't copy data `docker/buildx/util/imagetools/create.go:210` into the empty slice.   

**Related issue**
#12167 

image


glours (approved):
LGTM

PR #12172, opened 2024-09-29
feat(up): Add --wait-services to up command

idsulik:
**What I did**
Add `--wait-services` flag into the `up` command so the user can specify what services they want to wait, it'll be useful in cases like this one https://github.com/docker/compose/issues/10596

**Example**:
`docker compose up --wait-services my-service1,my-service2`

**Related issue**
https://github.com/docker/compose/issues/10596


ndeloof:
IMHO https://github.com/docker/compose/issues/10596 should better be fixed by _actual_ support for init containers, as the proposed UX is highly confusing between services to wait for and services selected as `up` target.

glours:
I agree with @ndeloof , with this UX it could be very easy to make mistakes and notice them
```shell
$ docker compose up --wait-services serviceA,serviceB serviceC
```

idsulik:
okay, thank you! I'll close the PR then

Issue #12174, opened 2024-09-30
Env files parsing


present-g:
We still have a problem with different parsing of env files in `docker compose` and `docker run`

Files:

`.env.test`
```bash
TEST_VAR='{"key": "value"}'
```

`test.Dockerfile`
```dockerfile
FROM ubuntu

CMD echo "$TEST_VAR"
```

`test.docker-compose.yaml`
```yaml
services:
  service:
    build:
      dockerfile: test.Dockerfile
    env_file: .env.test
    image: test-image
```

Execution:

`docker compose -f test.docker-compose.yaml up --build` => `{"key": "value"}`
`docker run --env-file .env.test test-image` => `'{"key": "value"}'`

We need unified results, otherwise, apps may not work correctly

Docker version 27.1.1, build 6312585
Docker Compose version v2.29.1-desktop.1

_Originally posted by @present-g in https://github.com/docker/compose/issues/2854#issuecomment-2383417633_

ndeloof:
Docker Compose .env file parser is not strictly equivalent to the one used by `docker run`, because the former manages variable interpolation while the latter just consider values as raw string, including quotes.

There's no workaround for this. A possible fix would be to give you an option to explicitly request env file to be loaded without compose to manage values:
```yaml
services:
  service:
    env_file: 
      path: .env.test
      format: raw # default is to parse with interpolation
```

present-g:
> Docker Compose .env file parser is not strictly equivalent to the one used by docker run, because the former manages variable interpolation while the latter just consider values as raw string, including quotes.

And that's the issue. Docker's primary goal is to ensure that everything runs consistently across all environments.

`format: raw` - is it working now? I've never seen that before

ndeloof:
introducing `format` is my proposal to offer a fix for this issue, it has to be implemented

PR #12179, opened 2024-10-02
add support for raw env_file format

ndeloof:
**What I did**
Add support for env_file `format` with `raw` to rely on docker/cli parser

requires 

**Related issue**
fixes https://github.com/docker/compose/issues/12174
requires https://github.com/docker/cli/pull/5491
requires https://github.com/compose-spec/compose-go/pull/690

image

### DIFF
--- a/pkg/e2e/fixtures/dotenv/raw.yaml
+++ b/pkg/e2e/fixtures/dotenv/raw.yaml
@@ -1,0 +1,7 @@
+services:
+  test:
+    image: alpine
+    command: sh -c "echo $$TEST_VAR"
+    env_file:
+      - path: .env.raw
+        format: raw # parse without interpolation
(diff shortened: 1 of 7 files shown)


pkg/e2e/fixtures/dotenv/raw.yaml
```diff
@@ -0,0 +1,7 @@
+services:
+  test:
+    image: alpine
+    command: sh -c "echo $$TEST_VAR"
+    env_file:
+      - path: .env.raw
+        format: raw # parse without interpolation
```
jhrotko, line 7: Do you think there can be other formats? I was wondering if it could be a boolean instead `raw:true` instead

ndeloof: my goal is to allow reusing docker/cli parser, which this repository can't refer to.
Also, while we don't have another format yet, I could imaging one might want to use java-style properties files, or maybe xml 🧌

jhrotko (approved):
LGTM! Should we update `compose-go` version?

glours (approved):

PR #12184, opened 2024-10-04
add support for CDI device request using `devices`

ndeloof:
**What I did**

mimic docker/cli behavior to detect `devices` path is set to a CDI string and create a device request accordingly
TBD this should be addressed in compose-go, converting such `devices` item into an actual `resources.reservations.devices` as canonical format

**Related issue**

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -57,6 +57,7 @@ require (
 	google.golang.org/grpc v1.62.0
 	gopkg.in/yaml.v3 v3.0.1
 	gotest.tools/v3 v3.5.1
+	tags.cncf.io/container-device-interface v0.8.0
 )
 
 require (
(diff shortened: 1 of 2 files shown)


glours (approved):
LGTM

akvadrako:
This isn't documented, right? Does it mean I can do this?

```yaml
    deploy:
      resources:
        reservations:
          devices:
            - device_ids: [ nvidia.com/gpu=all ]
```

hhansen-bdai:
I would also really appreciate some documentation on this feature, looks very useful!

ndeloof:
indeed, see https://github.com/compose-spec/compose-spec/pull/574/files

PR #12192, opened 2024-10-07
Remove feature flag integration with Docker Desktop for ComposeUI and ComposeNav

jhrotko:
**What I did**
Remove integration with feature flags ComposeUI and ComposeNav from Docker Desktop

image


glours (approved):
LGTM

Issue #12191, opened 2024-10-07
[BUG] docker compose up --watch does not respect Dockerfile.dockerignore


danrot:
### Description

I am having a project in which we have a `.docker` folder in which we keep the `Dockerfile` for everything related to this project. In there we would like to use the `Dockerfile.dockerignore` notation to have a separate `.dockerignore` file for all `Dockerfile`s. While this file seems to be respected during the initial `docker compose up --watch` step, it does not seem to respect it when syncing files using the `develop.watch` configuration in `docker-compose.yaml`. I think it should be this way, because I don't want to specify those files in multiple places.

I've tested with a `.dockerignore` file in the root folder, in which it worked as expected, i.e. the files would not be synced into the container as described in https://docs.docker.com/compose/how-tos/file-watch/. I'd like that exact behavior, just working for the `Dockerfile.dockerignore` notation as well.

### Steps To Reproduce

1. Create a project with a `docker-compose.yaml` file similar to this one:
```
services:
    php-fpm:
        build:
            context: .
            dockerfile: .docker/php-fpm/Dockerfile
        develop:
            watch:
                - action: sync
                  path: .
                  target: /var/www
```
2. Create a file `.docker/php-fpm/Dockerfile.dockerignore` which contains some files that should not be copied to the image/container
3. Execute `docker compose up --watch`
4. See that the files listed in `.docker/php-fpm/Dockerfile.dockerignore` are actually not in the started container (they won't be there)
5. Edit one of the files listed in `.docker/php-fpm/Dockerfile.dockerignore`
6. See that the file that has just been edited turns up in the started container (which it IMO shouldn't, because it is listed in `.docker/php-fpm/Dockerfile.dockerignore`)

### Compose Version

```Text
Docker Compose version v2.28.1-desktop.1
```

### Docker Environment

```Text
Client:
 Version:    27.0.3
 Context:    desktop-linux
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.15.1-desktop.1
    Path:     /Users/danielrotter/.docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.28.1-desktop.1
    Path:     /Users/danielrotter/.docker/cli-plugins/docker-compose
  debug: Get a shell into any image or container (Docker Inc.)
    Version:  0.0.32
    Path:     /Users/danielrotter/.docker/cli-plugins/docker-debug
[...]

PR #12193, opened 2024-10-08
Support Dockerfile-specific ignore-file with watch

ndeloof:
**What I did**
Follow https://docs.docker.com/reference/dockerfile/#dockerignore-file and first consider a Dockerfile-specific ignore file when computing exclude patterns for watch. 

**Related issue**
closes https://github.com/docker/compose/issues/12191

### DIFF
--- a/pkg/watch/dockerignore.go
+++ b/pkg/watch/dockerignore.go
@@ -18,10 +18,12 @@ package watch
 
 import (
 	"fmt"
+	"io"
 	"os"
 	"path/filepath"
 	"strings"
 
+	"github.com/compose-spec/compose-go/v2/types"
 	"github.com/docker/compose/v2/internal/paths"
 	"github.com/moby/patternmatcher"
 	"github.com/moby/patternmatcher/ignorefile"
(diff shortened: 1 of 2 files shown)


glours (approved):
👌 LGTM

pkg/watch/dockerignore.go
```diff
@@ -61,13 +63,28 @@ func (i dockerPathMatcher) MatchesEntireDir(f string) (bool, error) {
...
+func LoadDockerIgnore(build *types.BuildConfig) (*dockerPathMatcher, error) {
+	repoRoot := build.Context
 	absRoot, err := filepath.Abs(repoRoot)
 	if err != nil {
 		return nil, err
 	}
 
-	patterns, err := readDockerignorePatterns(absRoot)
+	// first try Dockerfile-specific ignore-file
+	f, err := os.Open(filepath.Join(repoRoot, build.Dockerfile+".dockerignore"))
```
thaJeztah, line 74: Is `build.Dockerfile` always just the file name, or can it be an absolute path? (in that case, no need to join if with the repo-root)

ndeloof: I'm actually running some tests as I wondered watch do abspath resolution, while compose-go should already have done

thaJeztah:
Quick blurb; not sure if "watch" is supported for windows container (and the classic builder), but I think the classic (non-buildkit) doesn't support per-dockerfile ignore-files; just in case that's relevant for this use 🤗

pkg/watch/dockerignore.go
ndeloof: `Dockerfile` is always just the file name within the context

PR #12197, opened 2024-10-09
convert gpus to DeviceRequests with implicit "gpu" capability

ndeloof:
**What I did**

This adds support for `gpus` to align with docker CLI
Coucou Guillaume 🥰

**Related issue**
https://github.com/compose-spec/compose-go/pull/693

image

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,7 +7,7 @@ require (
 	github.com/Microsoft/go-winio v0.6.2
 	github.com/acarl005/stripansi v0.0.0-20180116102854-5a71ef0e047d
 	github.com/buger/goterm v1.0.4
-	github.com/compose-spec/compose-go/v2 v2.2.1-0.20241007090213-a59035ad2bf4
+	github.com/compose-spec/compose-go/v2 v2.2.1-0.20241009124803-7218685580f7
 	github.com/containerd/containerd v1.7.22
 	github.com/containerd/platforms v0.2.1
 	github.com/davecgh/go-spew v1.1.1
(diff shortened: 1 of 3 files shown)


ndeloof:
Hello Guillaume, did you receive my notification ?

polarathene:
Was this meant to have documentation at https://docs.docker.com/reference/compose-file/services ?

I assume `gpus: all` is valid? (_to match [`--gpus all`](https://docs.docker.com/reference/cli/docker/container/run/#gpus)_), not able to test that any time soon.

The CLI docs also specifically refer to `--gpus` as only for nvidia, I'm not sure if that's outdated or a limitation (_I'm aware of [AMD just relying on `devices` field](https://github.com/open-webui/open-webui/blob/main/docker-compose.amdgpu.yaml), rather than `deploy.resources.reservations.devices`, but maybe that also works with `--gpus all` these days_ 🤷‍♂️).

polarathene:
> we could introduce such a shortcut notation

I assumed that was one of the main conveniences in supporting this alternative 😅 (_to have parity with the docker CLI `--gpus all` example that is commonly documented_)

---

> AFAIK docker engine only supports nvidia GPUs, but from compose point of view this is vendor-neutral: we just forward device capabilities request to engine, which has to find adequate device to match.

From what I've seen with AMD (and presumably Intel), those two are stuck with the more verbose `deploy` config, however CDI with `--device` / `device:` appears to be the preferred way going forward once the vendors better support it (_I've seen third-party tools to generate the CDI configs needed on the host_).

Presently it might only be nvidia with proper support (_I've not tried the CDI support yet, but have read there might be some extra changes required, and Docker Desktop / WSL2 may also affect compatibility_). So at the same time I'm not sure how beneficial `gpus: all` is to users of Compose (_Podman and k8s both adopted CDI_).

I imagine for consistency with docker CLI that `gpus: all` should be supported, but docs should likewise direct users towards CDI (_I have no clue how long that will remain behind a daemon feature flag to use though, which adds some friction_).

PR #12199, opened 2024-10-09
fix: Update down command to validate service names

### LINKED ISSUE #12196: Unexpected behavior with `docker compose down`
I have a compose file with 3 services:

```yaml
services:
    foo-api:
    bar-api:
    baz-api:
```

If I accidentally type `fooapi` with `docker compose down`, like this:

```shell
docker compose down fooapi
```

Then all services go down.

It must error instead. Or is this an intended behavior? Please enlighten me.

ganeshdipdumbare:
**What I did**
- I added validation to check if the service name(args to down command) is valid or not
**Related issue**
fixes #12196 

### DIFF
--- a/cmd/compose/down.go
+++ b/cmd/compose/down.go
@@ -83,6 +83,15 @@ func runDown(ctx context.Context, dockerCli command.Cli, backend api.Service, op
 		return err
 	}
 
+	// Validate service names
+	if project != nil {
+		for _, service := range services {
+			if _, exists := project.Services[service]; !exists {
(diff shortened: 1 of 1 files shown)


glours:
Hey @ganeshdipdumbare 
Thanks for your contribution! 
Unfortunately this code is needed as this is already covered by `compose-go` [here](https://github.com/compose-spec/compose-go/blob/7218685580f78b86b605669c0c6171674bbb2e9a/types/project.go#L208) and [this recent PR in Compose](https://github.com/docker/compose/pull/12164)

PR #12206, opened 2024-10-13
add kill process method when docker-compose down command

dohyunKim12:
**What I did**

Add kill process stage when docker-compose down command

**Related issue**
fixes 12205

### DIFF
--- a/cmd/compose/port_cleanup.go
+++ b/cmd/compose/port_cleanup.go
@@ -1,0 +1,44 @@
+package compose
+
+import (
+	"fmt"
+	"os/exec"
+	"strings"
+)
+
+func killProcessesUsingExposedPorts(project *types.Project) error {
+	for _, service := range project.Services {
+		for _, port := range service.Ports {
+			exposedPort := port.Published
+			if exposedPort != "" {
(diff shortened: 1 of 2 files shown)


cmd/compose/port_cleanup.go
```diff
@@ -0,0 +1,44 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to find process using port %s: %v", port, err)
+	}
+
+	pid := strings.TrimSpace(string(output))
+	if pid == "" {
+		return nil
+	}
+
+	killCmd := exec.Command("kill", "-9", pid)
```
glours, line 36: This part of the code is not portable and won't work on Windows environment for example.
The fix proposed should work for Linux, Mac and Windows

ndeloof: Also Compose is a client library, while docker engine may run on a remote machine, so this isn't the relevant place to propose such a fix. We already invoke `ContainerStop` API which is responsible to stop (or kill after timeout) the target container and release resources, if you found a bug with this, better report on github.com/moby/moby

ndeloof (changes requested):

Issue #11345, opened 2024-01-15
[BUG] "bind source path does not exist" on v2.24.0


pjsmith404:
### Description

Hello,

On version 2.24.0 of docker compose, I get the following bind error when trying to run `docker compose up`, if the source directory does not already exist:
```
Error response from daemon: invalid mount config for type "bind": bind source path does not exist: /tmp/logs
```

I expect that if the source directory does not exist, it should be automatically created. If I revert back to version 2.23.3, the error does not occur and the directory is created automatically as expected.

I didn't see anything in the change logs that indicated an expected change to this behaviour. Apologies if I overlooked something.

Cheers,

Pat

### Steps To Reproduce

1. Create a compose file located at `/tmp/test.yml` with the following content:
```
services:
  test:
    image: alpine
    volumes:
      - ./logs/:/tmp/logs/
    command: ls /tmp/logs/
```

2. Run `docker compose --file /tmp/test.yml up`
3. You should see the error `Error response from daemon: invalid mount config for type "bind": bind source path does not exist: /tmp/logs`
4. Create the folder `/tmp/logs/`
5. Run `docker compose --file /tmp/test.yml up`.
6. The container should launch as expected.

### Compose Version

```Text
Docker Compose version v2.24.0
```

### Docker Environment

```Text
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc., v0.11.2)
  compose: Docker Compose (Docker Inc., v2.24.0)

Server:
 Containers: 0
  Running: 0
  Paused: 0
  Stopped: 0
 Images: 0
 Server Version: 20.10.25
 Storage Driver: overlay2
  Backing Filesystem: xfs
  Supports d_type: true
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 1
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 io.containerd.runtime.v1.linux runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 0cae528dd6cb557f7201036e9f43420650207b58
 runc version: f19387a6bec4944c770f7668ab51c4348d9c2f38
 init version: de40ad0
 Security Options:
  seccomp
   Profile: default
[...]

PR #12208, opened 2024-10-14
compose-go clean volume target to avoid ambiguous comparisons

ndeloof:
**What I did**
use cleaned paths as volume target so we can't safely retrieve volume by target

**Related issue**
requires https://github.com/compose-spec/compose-go/pull/697
fix https://github.com/docker/compose/issues/11345

### DIFF
--- a/pkg/compose/create.go
+++ b/pkg/compose/create.go
@@ -24,7 +24,6 @@ import (
 	"fmt"
 	"io/fs"
 	"os"
-	"path"
 	"path/filepath"
 	"sort"
 	"strconv"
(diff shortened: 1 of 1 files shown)


pkg/compose/create.go
```diff
@@ -1095,8 +1093,6 @@ func buildMount(project types.Project, volume types.ServiceVolumeConfig) (mount.
 
 	bind, vol, tmpfs := buildMountOptions(volume)
 
-	volume.Target = path.Clean(volume.Target)
```
ndeloof, line 1098: this was the cause for https://github.com/docker/compose/issues/11345
volume being a func parameter, original struct isn't updated here and we eventually compare `/logs/` with `/logs`

jhrotko (approved):
LGTM

glours (approved):

PR #12209, opened 2024-10-14
introduce generate command as alpha command

glours:
**What I did**

Add a command to generate a compose file from a containers list

`docker compose alpha generate CONTAINER_NAME or CONTAINER_ID`

**Related issue**

image


pkg/compose/generate.go
```diff
@@ -0,0 +1,61 @@
...
+	if err != nil {
+		return nil, err
+	}
+	for _, container := range containersByIds {
+		if !utils.Contains(containers, container) {
+			containers = append(containers, container)
+		}
+	}
+
+	return s.projectFromName(containers, options.Project.Name)
```
ndeloof, line 60: `projectFromName` will search for compose labels and as such can't build a model from containers which have been explicitly created for this command, sounds to me this is a huge restriction for usability.
IMHO this command should create a service for each container passed as argument, and include volumes they refer to (+ a default network if there's no explicit one)

ndeloof: in this func we also only collect image and labels, resource limits isn't converted into compose model

pkg/compose/convert.go
```diff
@@ -97,3 +101,97 @@ func ToSeconds(d *compose.Duration) *int {
...
+	}
+	return volumeConfigs, serviceVolumeConfigs, secretConfigs, serviceSecretConfigs
+}
+
+func (s *composeService) toComposeNetwork(networks map[string]*network.EndpointSettings) (map[string]compose.NetworkConfig, map[string]*compose.ServiceNetworkConfig) {
+	networkConfigs := make(map[string]compose.NetworkConfig)
+	serviceNetworkConfigs := make(map[string]*compose.ServiceNetworkConfig)
+
+	for name, net := range networks {
+		if strings.HasSuffix(name, "_default") {
```
ndeloof, line 180: why skip this one?
while compose creates a xx_default network, I don't expect user to do the same just to help our tool generate a nice compose project :)

pkg/compose/generate.go
```diff
@@ -0,0 +1,244 @@
...
+			Aliases: net.Aliases,
+		}
+	}
+	return networkConfigs, serviceNetworkConfigs
+}
+
+func cleanDockerPreviousLabels(labels types.Labels) types.Labels {
+	cleanedLabels := types.Labels{}
+	for key, value := range labels {
+		if !strings.HasPrefix(key, "com.docker.compose.") && !strings.HasPrefix(key, "desktop.docker.io") {
```
lizzthabet, line 239: suggestion: are there other label paths that make sense to ignore, like `org.opencontainers.image`?

pkg/compose/generate.go
```diff
@@ -0,0 +1,244 @@
...
+	project.Services = services
+	project.Networks = networks
+	project.Volumes = volumes
+	project.Secrets = secrets
+	project.Configs = configs
+	return project, nil
+}
+
+func (s *composeService) extractComposeConfiguration(service *types.ServiceConfig, inspect moby.ContainerJSON, volumes types.Volumes, secrets types.Secrets, networks types.Networks) {
+	service.Environment = types.NewMappingWithEquals(inspect.Config.Env)
```
lizzthabet, line 115: do you know how container runs with env files get handled? do those values just get mapped to the `Config.Env`? i had this as a question / todo in my own work.

pkg/compose/generate.go
```diff
@@ -0,0 +1,244 @@
...
+
+	project.Services = services
+	project.Networks = networks
+	project.Volumes = volumes
+	project.Secrets = secrets
+	project.Configs = configs
+	return project, nil
+}
+
+func (s *composeService) extractComposeConfiguration(service *types.ServiceConfig, inspect moby.ContainerJSON, volumes types.Volumes, secrets types.Secrets, networks types.Networks) {
```
lizzthabet, line 114: wdyt about adding support for `cmd`, `entrypoint`, and `expose` blocks, too? they should be pretty straightforward. i'm also happy to put up a pr for those too!

lizzthabet (commented):
this is awesome to see! i left a few suggestions and pointed out a couple areas where i think the behavior of `generate` could be non-obvious to users, but those areas aren't blockers here. i'm not an expert on compose and the spec, so you're welcome to take my feedback with a grain of salt. there are definitely areas (like mapping networks and volumes) that i don't know well enough to review precisely.

after this pr, there are two areas of work that i think would improve the value of this command's output:
* filtering out redundant/verbose information in the service block that's inherited from images
* adding more precise image-handling logic so it can define `build` blocks where possible

curious what your thoughts are, and i'd be happy to work on those additions next!

pkg/compose/generate.go
glours: Yes the CLI read the file and populate the Config.Env attributes with the content of the file
https://github.com/docker/cli/blob/61baf2a3d984a1e59fcadb38e28d77b571434b88/cli/command/container/opts.go#L496

pkg/compose/generate.go
glours: Sure, that should be part of follow up PRs

pkg/compose/generate.go
```diff
@@ -0,0 +1,244 @@
...
+	filtersListNames := filters.NewArgs()
+	filtersListIDs := filters.NewArgs()
+	for _, containerName := range options.Containers {
+		filtersListNames.Add("name", containerName)
+		filtersListIDs.Add("id", containerName)
+	}
+	containers, err := s.apiClient().ContainerList(ctx, containerType.ListOptions{
+		Filters: filtersListNames,
+		All:     true,
+	})
```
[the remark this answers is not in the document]
ndeloof: indeed, filter by name may return more than expected. With an explicit container list better iterate over names and use `ContainerInspect`

pkg/compose/generate.go
```diff
@@ -0,0 +1,244 @@
...
+	return s.createProjectFromContainers(containers)
+}
+
+func (s *composeService) createProjectFromContainers(containers []moby.Container) (*types.Project, error) {
+	project := &types.Project{}
+	services := types.Services{}
+	networks := types.Networks{}
+	volumes := types.Volumes{}
+	secrets := types.Secrets{}
+	configs := types.Configs{}
```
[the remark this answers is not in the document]
glours: We just check and there isn't any way to detect a config configuration from a regular bind mount, so we just need to remove it

pkg/compose/generate.go
```diff
@@ -0,0 +1,244 @@
...
+
+	for _, c := range containers {
+		// if the container is from a previous Compose application, use the existing service name
+		serviceLabel, ok := c.Labels[api.ServiceLabel]
```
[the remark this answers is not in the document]
ndeloof: IMHO this is anticipating a need that isn't obvious. If we pretend we can recreate compose model from a (running) compose application, we actually don't (we're close, but not 100%) so better not try to.

pkg/compose/generate.go
```diff
@@ -0,0 +1,244 @@
...
+		if !ok {
+			serviceLabel = getCanonicalContainerName(c)
+		}
+		service, ok := services[serviceLabel]
+		if !ok {
```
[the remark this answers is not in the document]
ndeloof: That's a typical ambiguity we should not try to solve by black magic. Create a compose model from live resources should only cover the obvious model, and not try to guess how users will abuse it :P

pkg/compose/generate.go
ndeloof: would need to compare image CMD and ENTRYPOINT with actual container to detect those have been overridden (same for healthcheck)

pkg/compose/generate.go
ndeloof: this would need to exclude environment variables set by image

ndeloof (commented):
LGTM as a first iteration, just some notes
1. this tries to detect use of compose labels, which I don't expect a user would use for live resources, or .. he already used compose to run those and then, why run this command ? (and/or why would we encourage this practice)
2. we need to distinguish config set by image vs live config when guessing the compose model, i.e `docker run mysql` has ENV,  CMD and ENTRYPOINT set by image, I don't want those to be reflected in generated compose.yaml until those have been overridden intentionally

glours:
I'll remove the detections done via Compose specific labels  (maybe just keep the service name one 😇 )

pkg/compose/generate.go
```diff
@@ -0,0 +1,247 @@
...
+	}
+	if healthConfig.Retries != 0 {
+		retries := uint64(healthConfig.Retries)
+		healthCheck.Retries = &retries
+	}
+	return &healthCheck
+}
+
+func (s *composeService) toComposeVolumes(volumes []moby.MountPoint) (map[string]types.VolumeConfig,
+	[]types.ServiceVolumeConfig, map[string]types.SecretConfig, []types.ServiceSecretConfig) {
```
jhrotko, line 175: Might have missed me, but what about volume labels? Be aware we need to cleanup these labels for volumes https://github.com/docker/compose/blob/fafaa9c5b8e2c3d2f2fbdc5ad0dc9b13f2bf388a/pkg/compose/create.go#L143-L146

glours: Currently `labels` are only added to `services` not to `volumes`, `networks` ...

jhrotko (approved):

PR #12212, opened 2024-10-15
desktop: allow this client to be identified via user-agent

djs55:
**What I did**

Set the `User-Agent` when calling the Docker Desktop API, so it's possible to trace the server logs back to this client.

Previously the HTTP requests were sent with a generic Go-http-client user-agent which made it hard to determine where the requests are coming from. It's important that we can find clients to help debug issues (or to ensure we update clients if APIs change in future)

**Related issue**

Let's make it easier to see the owl:

image

### DIFF
--- a/internal/desktop/client.go
+++ b/internal/desktop/client.go
@@ -27,10 +27,14 @@ import (
 	"net/http"
 	"strings"
 
+	"github.com/docker/compose/v2/internal"
 	"github.com/docker/compose/v2/internal/memnet"
 	"github.com/r3labs/sse"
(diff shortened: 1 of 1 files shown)


glours (approved):
LGTM

glours:
@djs55 you need to sign-off your commits 😉

djs55:
@glours sorry about that! I've force pushed a`git commit --amend -s` 😅

ndeloof (approved):

djs55:
Thanks!

PR #12223, opened 2024-10-21
Recreate on volume change

### LINKED ISSUE #10060: [BUG] docker compose up does not recreate containers when named volume configuration is changed
### Description

Docker compose up is supposed to observe the state of running containers, compare this to the service definition, and recreate containers that need to be converged to the definition.

The options `--force-recreate` and `--no-recreate` are meant to override this behavior, so they are not specified when you want containers only to be recreated if a change is needed to bring them in line with the service definition.

When a service definition references a named volume defined in the top level element, currently a change to that volume configuration does not result in existing containers using that named volume being recreated.

### Steps To Reproduce

This is tested on Docker Desktop Mac (Apple Silicon)

1. With a compose.yml like the following
```yaml
services:
  changer:
    image: ealen/echo-server
    container_name: changer-container
    volumes:
      - changer-vol:/changervol

volumes:
  changer-vol:
    name: first
```
Run `docker compose up`

2. Then ctrl-c to stop the running container. Observe that a volume named `first` is mounted to the container
3. Edit the compose file to change the last line to `name: second`
4. Run `docker compose up` again

Expected behavior is that the `changer` container is recreated so that it has a volume named `second` mounted.

Actual behavior is that the container is not recreated, and still has the `first` volume mounted.

### Compose Version

```Text
Docker Compose version v2.12.2
```

### Docker Environment

```Text
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  compose: Docker Compose (Docker Inc., v2.12.2)
  dev: Docker Dev Environments (Docker Inc., v0.0.3)
  extension: Manages Docker extensions (Docker Inc., v0.2.13)
  scan: Docker Scan (Docker Inc., v0.21.0)

Server:
 Containers: 12
  Running: 2
  Paused: 0
  Stopped: 10
 Images: 33
 Server Version: 20.10.21
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 2
 Plugins:
  Volume: local
 Swarm: inactive
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 1c90a442489720eec95342e1789ee8a5e1b9536f
 init version: de40ad0
 Security Options:
  seccomp
   Profile: default
  cgroupns
 Kernel Version: 5.15.49-linuxkit
 OSType: linux
 Architecture: aarch64
 CPUs: 4
 Total Memory: 3.841GiB
 Name: docker-desktop
 ID: RM2G:OFRZ:AO72:RUPG:EMGW:6DOJ:UVW5:OEVK:5B3B:DF2L:AXD7:5VIA
 Debug Mode: false
 Labels:
 Experimental: false
 Insecure Registries:
  hubproxy.docker.internal:5000
  127.0.0.0/8
 Live Restore Enabled: false
```

### Anything else?

_No response_

jhrotko:
**What I did**
Add volumes on configuration hash in order to recreate container when `volumes` changes

**Related issue**

- fixes #10060 

image

### DIFF
--- a/cmd/compose/up.go
+++ b/cmd/compose/up.go
@@ -148,6 +148,7 @@ func upCommand(p *ProjectOptions, dockerCli command.Cli, backend api.Service) *c
 	flags.BoolVar(&create.noBuild, "no-build", false, "Don't build an image, even if it's policy")
 	flags.StringVar(&create.Pull, "pull", "policy", `Pull image before running ("always"|"missing"|"never")`)
 	flags.BoolVar(&create.removeOrphans, "remove-orphans", false, "Remove containers for services not defined in the Compose file")
+	flags.BoolVar(&create.recreateVolumes, "recreate-volumes", false, "Recreate volumes when volume configuration in the Compose file changes.")
 	flags.StringArrayVar(&create.scale, "scale", []string{}, "Scale SERVICE to NUM instances. Overrides the `scale` setting in the Compose file if present.")
 	flags.BoolVar(&up.noColor, "no-color", false, "Produce monochrome output")
 	flags.BoolVar(&up.noPrefix, "no-log-prefix", false, "Don't print prefix in logs")
@@ -255,6 +256,7 @@ func runUp(
 		Inherit:              !createOptions.noInherit,
 		Timeout:              createOptions.GetTimeout(),
 		QuietPull:            createOptions.quietPull,
+		RecreateVolumes:      createOptions.recreateVolumes,
 	}
 
 	if upOptions.noStart {
--- a/pkg/api/labels.go
+++ b/pkg/api/labels.go
@@ -29,8 +29,10 @@ const (
 	ProjectLabel = "com.docker.compose.project"
 	// ServiceLabel allow to track resource related to a compose service
 	ServiceLabel = "com.docker.compose.service"
-	// ConfigHashLabel stores configuration hash for a compose service
+	// ConfigHashLabel stores configuration hash for a compose service or volume
 	ConfigHashLabel = "com.docker.compose.config-hash"
+	// ExternalVolumeHashLabel stores volume external configuration hash for a compose service
+	ExternalVolumeHashLabel = "com.docker.compose.external-volume-hash"
 	// ContainerNumberLabel stores the container index of a replicated service
 	ContainerNumberLabel = "com.docker.compose.container-number"
 	// VolumeLabel allow to track resource related to a compose volume
--- a/pkg/compose/convergence.go
+++ b/pkg/compose/convergence.go
@@ -123,11 +123,11 @@ func (c *convergence) ensureService(ctx context.Context, project *types.Project,
 
 	sort.Slice(containers, func(i, j int) bool {
 		// select obsolete containers first, so they get removed as we scale down
-		if obsolete, _ := mustRecreate(service, containers[i], recreate); obsolete {
+		if obsolete, _ := mustRecreate(project, service, containers[i], recreate); obsolete {
 			// i is obsolete, so must be first in the list
 			return true
 		}
-		if obsolete, _ := mustRecreate(service, containers[j], recreate); obsolete {
+		if obsolete, _ := mustRecreate(project, service, containers[j], recreate); obsolete {
 			// j is obsolete, so must be first in the list
 			return false
 		}
@@ -154,7 +154,7 @@ func (c *convergence) ensureService(ctx context.Context, project *types.Project,
 			continue
 		}
 
-		mustRecreate, err := mustRecreate(service, container, recreate)
+		mustRecreate, err := mustRecreate(project, service, container, recreate)
 		if err != nil {
 			return err
 		}
@@ -312,7 +312,7 @@ func (c *convergence) resolveSharedNamespaces(service *types.ServiceConfig) erro
 	return nil
 }
 
-func mustRecreate(expected types.ServiceConfig, actual moby.Container, policy string) (bool, error) {
+func mustRecreate(project *types.Project, expected types.ServiceConfig, actual moby.Container, policy string) (bool, error) {
 	if policy == api.RecreateNever {
 		return false, nil
 	}
@@ -325,7 +325,28 @@ func mustRecreate(expected types.ServiceConfig, actual moby.Container, policy st
 	}
 	configChanged := actual.Labels[api.ConfigHashLabel] != configHash
 	imageUpdated := actual.Labels[api.ImageDigestLabel] != expected.CustomLabels[api.ImageDigestLabel]
-	return configChanged || imageUpdated, nil
+
+	// it is needed recreate the mount if the external volume name changes
+	externalChanged := externalVolumeNameChanged(expected, project, actual)
+	return configChanged || imageUpdated || externalChanged, nil
+}
+
+func externalVolumeNameChanged(expected types.ServiceConfig, project *types.Project, actual moby.Container) (externalChanged bool) {
+	for _, v := range expected.Volumes {
+		if vol, ok := project.Volumes[v.Source]; ok && bool(vol.External) {
+
+			externalFound := false
+			for _, mount := range actual.Mounts {
+				if mount.Name == vol.Name {
+					externalFound = true
+				}
+			}
+			if !externalFound {
+				externalChanged = true
+			}
+		}
+	}
+	return externalChanged
 }
 
 func getContainerName(projectName string, service types.ServiceConfig, number int) string {
--- a/pkg/compose/create.go
+++ b/pkg/compose/create.go
@@ -64,6 +64,8 @@ type createConfigs struct {
 	Links     []string
 }
 
+type ComposeVolumes map[string]*volumetypes.Volume
+
 func (s *composeService) Create(ctx context.Context, project *types.Project, createOpts api.CreateOptions) error {
 	return progress.RunWithTitle(ctx, func(ctx context.Context) error {
 		return s.create(ctx, project, createOpts)
@@ -80,12 +82,6 @@ func (s *composeService) create(ctx context.Context, project *types.Project, opt
 		return err
 	}
 
-	var observedState Containers
-	observedState, err = s.getContainers(ctx, project.Name, oneOffInclude, true)
-	if err != nil {
-		return err
-	}
-
 	err = s.ensureImagesExists(ctx, project, options.Build, options.QuietPull)
 	if err != nil {
 		return err
@@ -97,10 +93,15 @@ func (s *composeService) create(ctx context.Context, project *types.Project, opt
 		return err
 	}
 
-	if err := s.ensureProjectVolumes(ctx, project); err != nil {
+	if err := s.ensureProjectVolumes(ctx, project, options.RecreateVolumes); err != nil {
 		return err
 	}
 
+	var observedState Containers
+	observedState, err = s.getContainers(ctx, project.Name, oneOffInclude, true)
+	if err != nil {
+		return err
+	}
 	orphans := observedState.filter(isOrphaned(project))
 	if len(orphans) > 0 && !options.IgnoreOrphans {
 		if options.RemoveOrphans {
@@ -116,6 +117,25 @@ func (s *composeService) create(ctx context.Context, project *types.Project, opt
 		}
 	}
 	return newConvergence(options.Services, observedState, s).apply(ctx, project, options)
+}
+
+func (s *composeService) getVolumes(ctx context.Context, projectName string) (ComposeVolumes, error) {
+	volumeList, err := s.apiClient().VolumeList(ctx, volumetypes.ListOptions{
+		Filters: filters.NewArgs(projectFilter(projectName)),
+	})
+	if err != nil {
+		return nil, err
+	}
+	if volumeList.Volumes == nil {
+		return nil, nil
+	}
+
+	volumes := make(ComposeVolumes, len(volumeList.Volumes))
+	for _, v := range volumeList.Volumes {
+		name := v.Labels[api.VolumeLabel]
+		volumes[name] = v
+	}
+	return volumes, nil
 }
 
 func prepareNetworks(project *types.Project) {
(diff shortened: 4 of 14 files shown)


jhrotko:
I wonder in this case if we should remove the old container?

ndeloof:
IMHO the right way to address this issue is for volumes/networks to have a config-hash label just like services, so we detect those need to be recreated, and then in cascade we re-create the attached containers

pkg/compose/hash.go
```diff
@@ -40,3 +41,23 @@ func ServiceHash(o types.ServiceConfig) (string, error) {
...
+// DriverOpts,  External and Extensions
+func VolumeHash(o types.VolumeConfig) (string, error) {
+	// We can't have these attributes into account, because when converting
+	// from a volume inspect there is no way to deduce these attributes back
+	// to a VolumeConfig
+	if o.Driver == "" { // (TODO: jhrotko) This probably should be fixed in compose-go
+		o.Driver = "local"
+	}
+	o.External = false // the name can change. Need to think about this case
+	o.Labels = nil
```
ndeloof, line 55: why drop labels ?

pkg/compose/hash.go
```diff
@@ -40,3 +41,23 @@ func ServiceHash(o types.ServiceConfig) (string, error) {
...
+// From a top-level Volume Configuration, creates a unique hash ignoring
+// DriverOpts,  External and Extensions
+func VolumeHash(o types.VolumeConfig) (string, error) {
+	// We can't have these attributes into account, because when converting
+	// from a volume inspect there is no way to deduce these attributes back
+	// to a VolumeConfig
+	if o.Driver == "" { // (TODO: jhrotko) This probably should be fixed in compose-go
+		o.Driver = "local"
+	}
+	o.External = false // the name can change. Need to think about this case
```
ndeloof, line 54: if volume is external you should just never enter this func

pkg/compose/hash.go
jhrotko: can you add labels in a compose configuration?

ndeloof: sure: https://github.com/compose-spec/compose-spec/blob/main/07-volumes.md#labels
Remember compose `volumes` is just a 1:1 mapping for `docker volume create`.

jhrotko:
Added `--recreate-volumes` in `up` command. Missing e2e tests @ndeloof @glours

pkg/compose/create.go
```diff
@@ -139,15 +161,42 @@ func (s *composeService) ensureNetworks(ctx context.Context, networks types.Netw
...
+			// Check if it's necessary to recreate the volume by comparing
+			// with the current volume config
+			ok, err = shouldUpdateVolume(volume, actualVolume)
+			if err != nil {
+				return err
+			}
+			if !ok {
+				continue
+			}
+			logrus.Warnf("Warning: Running with --recreate-volumes will delete data for volume %s", k)
```
jhrotko, line 194: Added a warning message just to emphasize the risk that you will lose the data

cmd/compose/up.go
```diff
@@ -148,6 +148,7 @@ func upCommand(p *ProjectOptions, dockerCli command.Cli, backend api.Service) *c
 	flags.BoolVar(&create.noBuild, "no-build", false, "Don't build an image, even if it's policy")
 	flags.StringVar(&create.Pull, "pull", "policy", `Pull image before running ("always"|"missing"|"never")`)
 	flags.BoolVar(&create.removeOrphans, "remove-orphans", false, "Remove containers for services not defined in the Compose file")
+	flags.BoolVar(&create.recreateVolumes, "recreate-volumes", false, "Recreate volumes when volume configuration in the Compose file changes.")
```
jhrotko, line 151: we can rename it :P

pkg/compose/convergence.go
```diff
@@ -215,6 +239,7 @@ func (c *convergence) ensureService(ctx context.Context, project *types.Project,
 
 	err = eg.Wait()
 	c.setObservedState(service.Name, updated)
+	// TODO update obeserved state for volumes
```
jhrotko, line 242: should not forget about this 😄

pkg/compose/create.go
```diff
@@ -97,11 +91,20 @@ func (s *composeService) create(ctx context.Context, project *types.Project, opt
...
+	// create observed state for volumes
+	observedVolumesState, err := s.getVolumes(ctx, project.Name)
+	if err != nil {
+		return err
+	}
+	if err := s.ensureProjectVolumes(ctx, project, observedVolumesState, options.RecreateVolumes); err != nil {
 		return err
 	}
 
+	var observedState Containers
```
jhrotko, line 103: If the optional flag is enabled, we remove the container, so the observedState needs to be moved after `ensureProjectVolumes`

pkg/compose/create.go
```diff
@@ -116,7 +119,26 @@ func (s *composeService) create(ctx context.Context, project *types.Project, opt
 				"--remove-orphans flag to clean it up.", orphans.names())
 		}
 	}
-	return newConvergence(options.Services, observedState, s).apply(ctx, project, options)
+	return newConvergence(options.Services, observedState, s, observedVolumesState).apply(ctx, project, options)
+}
+
+func (s *composeService) getVolumes(ctx context.Context, projectName string) (map[string]*volumetypes.Volume, error) {
```
jhrotko, line 125: I think I should name this type to something prettier

pkg/compose/convergence.go
```diff
@@ -325,7 +350,35 @@ func mustRecreate(expected types.ServiceConfig, actual moby.Container, policy st
...
+	var needsUpdate = false
+	for _, volume := range service.Volumes {
+		name := volume.Source
+		// we need to check hash volumes (from inspect that should be inside the convergence) labels
+		if v, ok := actualVolumes[name]; ok {
+			// get volume hash from inspected volumes
+			volumeConfig := project.Volumes[name]
+
+			// get current hash for service volume
+			currentHash, err := VolumeHash(volumeConfig)
```
jhrotko, line 373: Note to self: add optional flag here

pkg/api/labels.go
```diff
@@ -31,6 +31,8 @@ const (
 	ServiceLabel = "com.docker.compose.service"
 	// ConfigHashLabel stores configuration hash for a compose service
 	ConfigHashLabel = "com.docker.compose.config-hash"
+	// ConfigHashLabel stores configuration hash for a compose service
+	VolumeConfigHashLabel = "com.docker.compose.volume-config-hash"
```
ndeloof, line 35: you could re-use ConfigHashLabel as this label is used to own the config for resource, regardless resource type

pkg/compose/create.go
ndeloof: warning comes too late, as when users see this we already recreated volume.
We should have a warning for user to know volume doesn't match config, and must use `--recreate-volumes` to force compose to recreate accordingly

pkg/compose/create.go
```diff
@@ -206,6 +254,46 @@ func (s *composeService) ensureProjectVolumes(ctx context.Context, project *type
...
+		for _, v := range s.Volumes {
+			if v.Source == volumeID {
+				selectedService = append(selectedService, s.Name)
+			}
+		}
+	}
+	err := s.Remove(ctx, project.Name, api.RemoveOptions{
+		Project:  project,
+		Services: selectedService,
+		Volumes:  true, // this does not work in the apiClient???
```
ndeloof: this also removes anonymous volumes, which is not required/desirable here

pkg/compose/convergence.go
```diff
@@ -123,11 +147,11 @@ func (c *convergence) ensureService(ctx context.Context, project *types.Project,
 
 	sort.Slice(containers, func(i, j int) bool {
 		// select obsolete containers first, so they get removed as we scale down
-		if obsolete, _ := mustRecreate(service, containers[i], recreate); obsolete {
+		if obsolete, _ := mustRecreate(project, service, containers[i], recreate, volumes); obsolete {
```
ndeloof, line 150: you don't need to check volumes here: if any was not matching config, it has just been recreated in `ensureVolumes` and service was removed, so `ensureService` will recreate service from scratch

cmd/compose/up.go
ndeloof: I also wonder this could be ambiguous, considering we have `--renew-anon-volumes`. Need to noun that make it clear we recreate volumes _when_ config has diverged

ndeloof (commented):
ensureService doesn't need to be updated for this feature AFAICT, as we remove containers mounting a volume before we recreate volume, so `ensureService` will start from scratch - possibly just requires `observedState` to be updated after `Remove`?

pkg/compose/convergence.go
jhrotko: If you go back to the first issue, changing the name of the volume won't update the mount in the container. This piece of code is necessary for it

ndeloof: you don't get my point: As you detect volume must be recreated within `ensureVolume`, you remove all containers with a mount it before you can delete and recreate volume. Later when you enter `ensureService`, as those containers have been removed, this code will never run and is useless.

pkg/compose/create.go
```diff
@@ -139,71 +163,139 @@ func (s *composeService) ensureNetworks(ctx context.Context, networks types.Netw
...
+	}
+
+	// create the volume
+	return s.createVolume(ctx, volumeConfig)
+}
 
+func shouldUpdateVolume(config types.VolumeConfig, volume *volumetypes.Volume) (bool, error) {
+	previousHash, ok := volume.Labels[api.VolumeConfigHashLabel]
+	if !ok {
+		return false, fmt.Errorf("could not get hash label for volume %s", volume.Name)
```
ndeloof, line 292: this should not return an error but just ignore, as user who bump compose version will get networks without labels, and will suddenly get all application recreated without a notice -> prepare for a bunch of angry users :P

pkg/api/labels.go
jhrotko: I would prefer to create a new label with a new function. It is easier for us ti later find the reference to the volume configuration label later in the code.

ndeloof: :-1 we only use this in a few places

cmd/compose/up.go
jhrotko: how about `update-volumes`?

pkg/compose/hash.go
jhrotko: there is an edge case where you can change the `name` to of an external volume and we need to update the service. The current code does not comtemplate this case. I need to implement this

ndeloof: by nature, external resources are not managed by compose. There's nothing we can and should do to cover this scenario

ndeloof: I better understand your point, the issue you refer to is a special case for a feature we don't support (yet): recreate services when referred resource is reconfigured. I agree we could kill 2 birds with one stone here, but this makes many changes at once.
Maybe split this into 2 separate PRs ?
- 1 to recreate volumes on config changes
- 1 to detect named resource changed (ensureService to inspect mounted volumes for container and check name matches expectation)

cmd/compose/up.go
jhrotko: @ndeloof the `recreate-volumes` makes somewhat sense to me, also given the 'pattern' for `force-recreate` flag. I don't think it is ambiguous with the flag `--renew-anon-volumes`

ndeloof: it is, as we only recreate diverged volumes. Not all volumes (and not anonymous ones)

pkg/compose/convergence.go
jhrotko: Applied changes, still need to account for the external volume name change

pkg/compose/create.go
```diff
@@ -138,71 +158,152 @@ func (s *composeService) ensureNetworks(ctx context.Context, networks types.Netw
...
+		// Check if it's necessary to recreate the volume by comparing
+		// with the current volume config hash
+		shouldRecreateVolume, err := shouldUpdateVolume(volume, actualVolume)
 		if err != nil {
 			return err
 		}
+		if !shouldRecreateVolume {
+			continue
+		}
+		if !recreateVolumes && shouldRecreateVolume {
```
ndeloof, line 195: we could rely on user interaction to confirm recreation, the way we do on https://github.com/docker/compose/blob/main/pkg/compose/remove.go#L88

pkg/compose/hash.go
jhrotko: I applied both solutions to this PR

ndeloof: sure, I mean it could be simpler to review if we split this effort into two separate PRs

pkg/compose/hash.go
```diff
@@ -40,3 +41,32 @@ func ServiceHash(o types.ServiceConfig) (string, error) {
...
+func VolumeHash(v types.VolumeConfig) (string, error) {
+	o, err := DeepCopy(&v)
+	if err != nil {
+		return "", err
+	}
+
+	if o.Driver == "" { // (TODO: jhrotko) This probably should be fixed in compose-go
+		o.Driver = "local"
+	}
+	delete(o.Labels, api.ConfigHashLabel)
```
ndeloof, line 66: also need to ignore a few more labels, like ComposeVersionLabel (don't wan't to recreate all volumes after you updated compose, do you?)
See https://github.com/compose-spec/compose-go/pull/708 and https://github.com/docker/compose/pull/12267 for example usage

pkg/compose/create.go
jhrotko: looks really nice, but I will defer it for a follow-up PR

cmd/compose/up.go
jhrotko: @ndeloof I can add that at the flag description: that this will ignore anonymous volumes. As for the names maybe? If you have any other ideas please feel free to suggest

--recreate-volumes-if-changed
--force-recreate-volumes
--sync-volumes
--update-volumes
--refresh-named-volumes
--rebuild-volumes

jhrotko: Or do you prefer to make this the default behaviour with a prompt?

Issue #12010, opened 2024-07-23
[BUG] Incorrect warning about secrets from environment variable being ignored


qhaas:
### Description

While it is known that [file secrets](https://github.com/docker/compose/issues/9648#issuecomment-1824340917) cannot have their ownership / permissions set via [service level secrets](https://docs.docker.com/compose/compose-file/05-services/#long-syntax-4), it appears that [environment secrets](https://docs.docker.com/compose/compose-file/09-secrets/#example-2) can, yet the warning about said ownership / permissions being ignored is displayed.

### Steps To Reproduce

1. Set a secret from an environment variable and attempt to modify its uid, gid, and mode via service-level secrets
```
$ cat compose.yml 
services:
  my-service:
    image: busybox:stable
    command: ls -ln /run/secrets/my-secret
    secrets:
      - source: my-secret
        uid: "10"
        gid: "20"
        mode: 0754
secrets:
  my-secret:
    environment: MY_SECRET
```
2. Run docker compose and display the uid, gid, and mode of the secret, noting the incorrect warning about them not being set
```
$ MY_SECRET=success docker compose run --rm my-service
WARN[0000] secrets `uid`, `gid` and `mode` are not supported, they will be ignored 
-rwxr-xr--    1 10       20               7 Jul 23 20:40 /run/secrets/my-secret
```

### Compose Version

```Text
Docker Compose version v2.29.0
```

### Docker Environment

```Text
Client: Docker Engine - Community
 Version:    27.1.0
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.16.1
    Path:     /usr/libexec/docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.29.0
    Path:     /usr/libexec/docker/cli-plugins/docker-compose

Server:
 Containers: 0
  Running: 0
  Paused: 0
  Stopped: 0
 Images: 42
 Server Version: 27.1.0
 Storage Driver: overlay2
  Backing Filesystem: xfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local splunk syslog
 Swarm: inactive
[...]

PR #12232, opened 2024-10-24
don't warn about uid/gid not being supported while ... they are

ndeloof:
**What I did**
this warning only applies to bind mount. As we skip creating a bind mount when source is environment or inlined content, must be skipped.

**Related issue**
fixes https://github.com/docker/compose/issues/12010


glours (approved):
LGTM

Issue #11524, opened 2024-02-19
[BUG] Container created by Docker Compose 2.24.6 fails to reconnect to default NAT network on Windows restart


michal-korniak-kornit:
### Description

**Description:**
I am experiencing an issue with Docker Compose 2.24.6 on Windows, using Docker version 25.0.3. The problem revolves around how these versions handle network connections to a pre-made NAT network after a system restart. This behaviour is inconsistent with that observed in Docker Compose 1.29.2 and when using Docker commands directly (docker run), which manage the scenario as expected.

To observe the behaviour, follow the steps described in the "Steps To Reproduce" section for both versions 2.24.6 and 1.29.2.

**Steps To Reproduce**

1. Use the following docker-compose.yml configuration:

```
services:
  mongo:
    image: mongo
    restart: unless-stopped
    networks:
      - nat
networks:
  nat:
    external: true
```

2. Start the container with Docker Compose using docker-compose up -d
3. Inspect the network mode of the container with docker inspect --format='{{.HostConfig.NetworkMode}}' mongo
4. Restart the Windows operating system.
5. Check if the container is up and running. If not, investigate the failure reason.

**Findings:**

* For Docker Compose 1.29.2:

1. The container starts correctly
2. The command docker inspect returns nat, indicating a connection by network name.
3. After a system restart, the container continues to operate correctly.

* For Docker Compose 2.24.6:

1. The container initially starts correctly. 
3. However, docker inspect returns a network ID (e.g., 96d6b0255d82eb1f904a9a502f333957a5260bee69fd16b2ecf5f337a23f72de), showing a connection by network ID rather than by name.
4. After a system restart, the container fails to start with an error indicating that the network with the specified ID could not be found:
Error: "could not find a network matching network mode 96d6b0255d82eb1f904a9a502f333957a5260bee69fd16b2ecf5f337a23f72de: network 96d6b0255d82eb1f904a9a502f333957a5260bee69fd16b2ecf5f337a23f72de not found."

**Core Issue:**
The difference in the output of docker inspect between versions 1.29.2 and 2.24.6 is crucial. It reveals that version 1.29.2 connects to the network using its name, as anticipated, while version 2.24.6 uses the network ID for connection. This approach poses a problem after a system restart because the NAT network, though persistently present, is regenerated with a new ID. Consequently, this leads to a failure for containers expecting the network to retain the ID it had before the restart.

**Temporary Workaround:**
[...]

PR #12234, opened 2024-10-25
connect to external networks by name

ndeloof:
**What I did**

Network ID changes after a system restart, so better **not** resolve external network name into ID
This was introduced by https://github.com/docker/compose/commit/1bd8a773a783512b93de6b938173959a4886a081 to prevent risk of collision with network name not being guaranteed unique. the code still detect such a collision during container creation, I would assume the race condition to get duplicated network name won't take place on a system restart.

**Related issue**
fixes https://github.com/docker/compose/issues/11524


glours (approved):
LGTM

PR #12238, opened 2024-10-26
[Addresses 12236]: docker compose --watch doesn't properly handle .dockerignore whitelists

mvhatch:
1. `make build`
2. `cd bug-12236`
3. `../bin/build/docker-compose -D up --watch alpine`

## Expected Output:
1. Debug logs show that docker compose copies 1 file for initial sync. 
2. Docker compose copies`./dir1` from the host machine to `/app/dir1` in the created container.
3. The `ls -la` output shows `dir1`.

## Actual Output:
```
✗ ../bin/build/docker-compose -D up --watch alpine
[+] Running 1/0
 ✔ Container bug-12236-alpine-1  Created                                                                                                                                       0.0s 
Attaching to alpine-1
DEBU[0000] Using .dockerignore file: /Users/michaelhatch/compose/bug-12236/.dockerignore 
DEBU[0000] ignoring file /Users/michaelhatch/compose/bug-12236/dir1/foo.txt 
DEBU[0000] copying 0 files for initial sync             
alpine-1  | total 8
alpine-1  | drwxr-xr-x    2 root     root          4096 Oct 26 17:11 .
alpine-1  | drwxr-xr-x    1 root     root          4096 Oct 26 17:34 ..
DEBU[0000] Watch configuration for service "alpine":
  - Action sync+restart for path "/Users/michaelhatch/compose/bug-12236/dir1" 
          ⦿ Watch enabled
alpine-1 exited with code 0
          ⦿ Watch disabled
```

## Related Issue
See issue [here](https://github.com/docker/compose/issues/12236).

Issue #12236, opened 2024-10-25
[BUG] compose watch x-initialSync does not properly handle override options in *.ignore files


mvhatch:
### Description

I have a `.dockerignore` file that ignores everything by default, and then subsequently overrides this behavior w/ exceptions. For example:
```
*
!dir1
```

From what I can tell, it seems `docker compose up <service> --watch` will **not** perform an initial sync of files, presumably because the `*` in the `.dockerignore` file. When I comment out the `*`, the initial sync syncs files.

### Compose Version

```
✗ docker compose version
Docker Compose version 2.29.3
```

```
✗ docker-compose version
Docker Compose version 2.29.3
```

### Docker Environment

```
✗ docker info
Client: Docker Engine - Community
 Version:    27.2.1
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.15.1
    Path:     /opt/homebrew/lib/docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  2.29.3
    Path:     /opt/homebrew/lib/docker/cli-plugins/docker-compose

Server:
 Containers: 5
  Running: 5
  Paused: 0
  Stopped: 0
 Images: 22
 Server Version: 26.1.1
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 2
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: e377cd56a71523140ca6ae87e30244719194a521
 runc version: v1.1.12-0-g51d5e94
 init version: de40ad0
 Security Options:
  apparmor
  seccomp
   Profile: builtin
  cgroupns
 Kernel Version: 6.8.0-31-generic
 Operating System: Ubuntu 24.04 LTS
 OSType: linux
 Architecture: aarch64
 CPUs: 8
 Total Memory: 39.09GiB
 Name: colima
 ID: 07cdc830-1c03-4414-9428-2a2dbab39027
 Docker Root Dir: /var/lib/docker
[...]

PR #12239, opened 2024-10-27
fix(dockerignore): Add wildcard support to dockerignore.go

idsulik:
**What I did**
Added wildcard support to `dockerignore::NewDockerPatternMatcher`, because if the .dockerignore contains a wildcard to ignore all files + lines with an exclusion, it ignores all files while it shouldn't ignore the excluded files

**Related issue**
https://github.com/docker/compose/issues/12236

PR to reproduce the bug https://github.com/docker/compose/pull/12238

image

### DIFF
--- a/pkg/watch/dockerignore_test.go
+++ b/pkg/watch/dockerignore_test.go
@@ -1,3 +1,19 @@
+/*
+   Copyright 2020 Docker Compose CLI authors
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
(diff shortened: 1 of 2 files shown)


pkg/watch/dockerignore_test.go
```diff
@@ -0,0 +1,92 @@
+package watch
```
glours, line 1: You need to add the license header to this new file

glours (approved):
Sounds good to me, you just need to add the license header to the new created file

pkg/watch/dockerignore_test.go
idsulik: thank you, pushed fix

PR #12251, opened 2024-10-30
Add support for configs.file's and secrets.file's on remote docker hosts

andoks:
**What I did**
Copy configs.file's and secrets.file's instead of bind-mounting them to make it possible to use file configs when working with remote docker hosts (like setting DOCKER_HOST to a ssh address or setting docker context)

Includes support for config.files and secrets.files as directories.

Note that file.Content as source of secrets is denied elsewhere with the error "validating docker-compose.yml: secrets.content_secret Additional property content is not allowed", but it is implemented here in case this restriction is liften in the future.

Configs and secrets from environment is also handled as plain content inserted into target file.

**Related issue**

implements: #11867


andoks:
Replaces #11871

Issue #12256, opened 2024-10-31
[BUG] docker compose up --watch restarts services that shouldn't


RasmusN:
### Description

**The issue**
It is my understanding that only the service that has watch configured should rebuild/restart/sync when changes are made to the specified code. However, in the example below, both services are restarted even though the second one does not have watch configured. I'm suspecting that this is because the container has exited. 

**Background**
I encountered this when trying to have a migration service that is only ran once when starting my service+db, but not every time I make code changes in my service. 

### Steps To Reproduce

1. In this environment
```
project/
├── docker-compose.yml
├── service_a/
│   ├── Dockerfile
│   └── app.py
└── service_b/
    ├── Dockerfile
    └── app.py
```
docker-compose.yml
```yml
services:
  service_a:
    build:
      context: ./service_a
    command: python app.py
    develop:
      watch:
        - action: rebuild
          path: ./service_a
   
  service_b:
    build:
      context: ./service_b
    command: python app.py
```
service_a/Dockerfile & service_b/Dockerfile
```
FROM python:3.9-slim
WORKDIR /app
COPY app.py .
CMD ["python", "app.py"]
```
service_a/app.py
```python
import time
print("Hello from A.")
while True:
    time.sleep(1)
```
service_b/app.py
```python
print("Hello from B")
```

2. Run this command
`docker compose build && docker compose up --watch`

3. Make a change to service_a/app.py (insert blank line or whatever).

4. Check logs
```
$ docker compose logs -t service_b
service_b-1  | 2024-10-31T10:10:49.879451621Z Hello from B
service_b-1  | 2024-10-31T10:12:13.803864357Z Hello from B
```

### Compose Version

$ docker compose version
Docker Compose version v2.29.2-desktop.2

$ docker-compose --version
Docker Compose version v2.29.1-desktop.1

### Docker Environment

$ docker info
Client:
 Version:    27.2.0
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.16.2-desktop.1
    Path:     C:\Program Files\Docker\cli-plugins\docker-buildx.exe
  compose: Docker Compose (Docker Inc.)
    Version:  v2.29.2-desktop.2
    Path:     C:\Program Files\Docker\cli-plugins\docker-compose.exe
  debug: Get a shell into any image or container (Docker Inc.)
    Version:  0.0.34
    Path:     C:\Program Files\Docker\cli-plugins\docker-debug.exe
[...]

PR #12258, opened 2024-10-31
Avoid starting all services on rebuild

jhrotko:
**What I did**
When a service is rebuild during watch, it will trigger to start all services since the graph to start the project is built based on the selected services the project has

**Related issue**

- fixes https://github.com/docker/compose/issues/12256

silly ball

### DIFF
--- a/pkg/compose/start.go
+++ b/pkg/compose/start.go
@@ -214,13 +214,13 @@ func (s *composeService) watchContainers(ctx context.Context, //nolint:gocyclo
 	}
 
 	var (
-		expected []string
+		expected = utils.NewSet[string]()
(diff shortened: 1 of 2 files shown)


ndeloof (approved):
LGTM. I wonder anyway why other services would be restarted, `start` should just ignore those if they are already up

RasmusN:
> LGTM. I wonder anyway why other services would be restarted, `start` should just ignore those if they are already up

Only services that has exited is restarted

Rizalibrahim (commented):
Want to change

ndeloof:
@Rizalibrahim can you please clarify your comment ?

PR #12267, opened 2024-11-05
detect network config changes and recreate if needed

ndeloof:
**What I did**
use a config hash label on networks to detect need to be recreated.
attached services will be removed so we can recreate network, then convergence will recreate them

**Related issue**
fixes https://github.com/docker/compose/issues/9203


pkg/compose/hash.go
```diff
@@ -40,3 +40,11 @@ func ServiceHash(o types.ServiceConfig) (string, error) {
 	}
 	return digest.SHA256.FromBytes(bytes).Encoded(), nil
 }
+
+func NetworkHash(o *types.NetworkConfig) (string, error) {
+	bytes, err := json.Marshal(o)
+	if err != nil {
+		return "", err
+	}
```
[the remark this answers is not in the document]
ndeloof: nope, this is why I introduced CustomLabel in compose-go (same mechanism used for services)
https://github.com/docker/compose/pull/12267/files#diff-1f4d596600c620b3df55dbfbc63e50a6f43b356865fae411c553d0937fbb6d91R122

terev:
Bless you for this change! Was just working on a workaround for our internal compose wrapper but it was proving difficult to be sure it works sanely. 

One question, is there a reason we can't do the network recreation more gracefully? Like detach all containers connected to the diverged network then remove the network only. Is this because the disconnected containers won't be reattached in the current reconciliation strategy?

ndeloof:
@terev I also considered this approach, but this indeed require more changes on the convergence side to re-attach + this involves the container will manage network being disconnected/reconnected (IP address will change) and in many cases this is not well supported. Recreating is "safer" from this point of view.

What's your use-case for changing network definition and requirement for app to keep running ?

terev:
@ndeloof Ok understood, that makes sense. It's not a super hard requirement, as my use case only applies to local development where I'd like to gracefully update the default binding address. 

I was just weighing network detachment against the possibility of accidentally removing container logs and data. But even if container data is lost it probably should have been kept in a volume in the first place.

ndeloof:
Right. This makes me wonder we should preserve anonymous volumes on recreation, to avoid such a data loss.

terev:
That would be neat if possible, though that sounds like it'd be a totally new mechanism given that anonymous volumes are currently disassociated with a service if its container is recreated. I do think that'd be a neat feature that would make the Dockerfile volume instruction much more useful.

ndeloof:
@terev when a service recreated, anonymous volumes are preserve already. We can reuse this mechanism

terev:
Oh! Right sorry I'm mistaken, that's perfect.

pkg/compose/convergence.go
```diff
@@ -328,7 +330,33 @@ func mustRecreate(expected types.ServiceConfig, actual moby.Container, policy st
...
+			}
+			found := false
+			for _, settings := range actual.NetworkSettings.Networks {
+				if settings.NetworkID == id {
+					found = true
+					break
+				}
+			}
+			if !found {
+				// config is up-t-date but container is not connected to network - maybe recreated ?
```
terev, line 353: Fwiw this makes a lot of sense to me. Would even say it makes sense to recreate if it is connected to undesired networks.

pkg/compose/convergence.go
```diff
@@ -328,7 +330,33 @@ func mustRecreate(expected types.ServiceConfig, actual moby.Container, policy st
 	}
 	configChanged := actual.Labels[api.ConfigHashLabel] != configHash
 	imageUpdated := actual.Labels[api.ImageDigestLabel] != expected.CustomLabels[api.ImageDigestLabel]
-	return configChanged || imageUpdated, nil
+	if configChanged || imageUpdated {
+		return true, nil
+	}
+
+	if c.networks != nil {
```
terev, line 337: Should we recreate if networks is nil but expected  has networks?

ndeloof: networks is nil only if caller didn't created convergence with actual network state - this is the case running `compose run` command, need to investigate why we don't manage networks in this command, will do in a follow-up PR.

glours (approved):
Sounds good to me

PR #12268, opened 2024-11-05
Add commit command

jarqvi:
What I did
I noticed that we have this command in the Docker CLI, but not in Docker Compose. So I thought of adding it to make it useful for others.


jarqvi:
ping @ndeloof :)

ndeloof (approved):

jarqvi:
@ndeloof 

Thank you for reviewing this PR.
I fixed the failed tests.

ndeloof (approved):

Issue #12015, opened 2024-07-24
[BUG] Weird behavior between orphaned containers and `--remove-orphans` flag


MatTheCat:
### Description

After using `run` on an updated Compose, I’ve been greeted by a

> WARN[0000] Found orphan containers […] for this project. If you removed or renamed this service in your compose file, you can run this command with the --remove-orphans flag to clean it up.

The thing is, I didn’t remove or rename the corresponding service, so I don’t know how to reproduce.

Adding the `--remove-orphans` flag on `run` didn’t change anything, but it did work with `up`.

### Compose Version

```Text
$ docker compose version
Docker Compose version v2.29.0
```
```


### Docker Environment

```Text
$ docker info
docker info
Client: Docker Engine - Community
 Version:    27.1.0
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.16.1
    Path:     /usr/libexec/docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.29.0
    Path:     /usr/libexec/docker/cli-plugins/docker-compose

Server:
 Containers: 11
  Running: 4
  Paused: 0
  Stopped: 7
 Images: 30
 Server Version: 27.1.0
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 1
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 2bf793ef6dc9a18e00cb12efb64355c2c9d5eb41
 runc version: v1.1.13-0-g58aa920
 init version: de40ad0
 Security Options:
  apparmor
  seccomp
   Profile: builtin
 Kernel Version: 6.5.0-44-generic
 Operating System: Ubuntu 22.04.4 LTS
 OSType: linux
 Architecture: x86_64
 CPUs: 8
 Total Memory: 30.74GiB
 Name: P-ASN-TINY
 ID: 35d93767-e18b-4cc4-8eeb-d7fd382366f2
[...]

PR #12288, opened 2024-11-12
fix support for --remove-orphans on `docker compose run`

ndeloof:
**What I did**
Fix: `remove-orphans` flag was assigned to createOptions we don't use to run convergence

**Related issue**
fixes https://github.com/docker/compose/issues/12015

### DIFF
--- a/cmd/compose/run.go
+++ b/cmd/compose/run.go
@@ -63,6 +63,7 @@ type runOptions struct {
 	name          string
 	noDeps        bool
 	ignoreOrphans bool
+	removeOrphans bool
 	quietPull     bool
 }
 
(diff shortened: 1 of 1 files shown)


glours (approved):
LGTM

volcan01010:
Adding this comment for information in case anyone else got the same problem as I did following this update:  With two docker-compose files in the same directory, calling `docker compose run --remove-orphans`  on one file will now remove running containers belonging to the other file unless you have explicitly defined a top level `name` in each file.

--- 

I had two docker compose files in a single directory.  The first, `docker-compose-datasette.yml`, defined a long-running web service (Datasette).  The second, `docker-compose-data-update.yml`, defined configuration options for a container with many Python dependencies.  It allowed me to run the Python script within the container using `docker compose -f docker-compose-data-update.yml run ...`.

I had seen the error message about orphaned containers and was running the command equivalent to

```bash
docker compose -f docker-compose-data-update.yml run --rm --remove-orphans update_data python myscript.py
```

This worked fine until this fix landed in an update in mid-December 2024.

From then, running the update script command would terminate the running datasette process.  This was because both `docker-compose.yml` files were in the same directory.  The update file sees the running datasette as an orphan of the same project and terminates it.

To fix my issue, I set a [top level `name`](https://docs.docker.com/reference/compose-file/version-and-name/#name-top-level-element) for each project.  The running commands from the different compose files no longer affect each other.

PR #12300, opened 2024-11-15
delegate build to buildx bake

ndeloof:
**What I did**

introduce support for COMPOSE_BAKE to opt-in for build to be delegated to buildx bake. Compose generates a bake json config from application model and executes bake, which (we assume)  better knows how to orchestrate build steps for best efficiency.

**Related issue**

### DIFF
--- a/pkg/compose/build.go
+++ b/pkg/compose/build.go
@@ -21,7 +21,6 @@ import (
 	"errors"
 	"fmt"
 	"os"
-	"path/filepath"
 
 	"github.com/compose-spec/compose-go/v2/types"
 	"github.com/containerd/platforms"
@@ -38,7 +37,6 @@ import (
 	"github.com/docker/compose/v2/pkg/progress"
 	"github.com/docker/compose/v2/pkg/utils"
 	"github.com/docker/docker/api/types/container"
-	"github.com/docker/docker/builder/remotecontext/urlutil"
 	bclient "github.com/moby/buildkit/client"
 	"github.com/moby/buildkit/session"
 	"github.com/moby/buildkit/session/auth/authprovider"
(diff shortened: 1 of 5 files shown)


pkg/compose/build.go
```diff
@@ -92,13 +86,23 @@ func (s *composeService) build(ctx context.Context, project *types.Project, opti
...
-		serviceToBeBuild[serviceName] = serviceToBuild{name: serviceName, service: *service}
+		serviceToBuild[serviceName] = *service
 		return nil
 	}, policy)
-	if err != nil || len(serviceToBeBuild) == 0 {
+	if err != nil || len(serviceToBuild) == 0 {
 		return imageIDs, err
 	}
 
+	if b, ok := os.LookupEnv("COMPOSE_BAKE"); ok {
```
thaJeztah, line 96: Perhaps could also be worth (if compose reads the cli config) to consider either an option in `features`, or `plugins` (plugins allows plugin-specific options to be set), which would allow opt-in/opt-out of this without having to use an env-var;
https://github.com/docker/cli/blob/9861ce90fd6b8ddca19db5f803dcbef9a583e9e1/cli/config/configfile/file.go#L42-L44

```go
	Plugins              map[string]map[string]string `json:"plugins,omitempty"`
	Aliases              map[string]string            `json:"aliases,omitempty"`
	Features             map[string]string            `json:"features,omitempty"`
```

ndeloof: we never used this mechanism in the past for optional/experimental docker compose features

thaJeztah: Yeah, my thinking here was that the cli-config would more easily allow this to be set as a default, which could also allow (e.g.) it to be set through docker desktop "settings".

glours (approved):
LGTM, tests ok
We should open follow up PRs to add bake flavour of existing e2e build tests

ndeloof:
I was indeed looking for configuration on CI to run build-related tests with this new option enabled.

thaJeztah:
Just some quick blurbs;

- Happy to see this being worked on again (thank you!)
- I asked @laurazzard and others to have a peek at the cli-config part (we've been pretty bad at documenting conventions to use, so I asked them to have a peek)
- One thing I'm wondering (not for this stage!) is if there'd be cases where we could even take out compose "in the middle", i.e., if we (in the CLI) could alias `docker compose build` to `docker buildx bake` (with the right options). Haven't wrapped my head around that option, but mostly considering that spinning up multiple binaries isn't always ideal, so if there's more optimizations to be made.

ndeloof:
> One thing I'm wondering (not for this stage!) is if there'd be cases where we could even take out compose "in the middle", i.e., if we (in the CLI) could alias docker compose build to docker buildx bake (with the right options). Haven't wrapped my head around that option, but mostly considering that spinning up multiple binaries isn't always ideal, so if there's more optimizations to be made.

I don't think this is a good solution. 1st, this would break separation of concerns, and force the docker CLI to be aware about compose sub-commands and UX (only a subset of service might be requested for build). Also, this would interfere with other commands which involve building images but also some compose actions, typically `up --build` or `watch`.

Issue #12310, opened 2024-11-21
[BUG] Docker Compose up with build option stops services one by one slowly before restarting them


joaomcarlos:
### Description

Hi,

This issue only started to come up recently (only noticed it yesterday but before yesterdays update as well)

`docker compose up -d --build` will:
- rebuild all services
- [BUG] stop each service one by one (rather than stopping them in parallel)
- Recreate + Start the stopped services all together

I have a project with 10 microservices. This means that instead of about 10 seconds building + 4/5 seconds restarting. I get 10 seconds building

### Steps To Reproduce

1. Have a config with N services (DB or RabbitMQ or Kafka + actual services that depend on those, as in, have health checks and start conditions)
2. Have everything running already
3. Do a `docker system up -d --build` (with build will do it regardless if theres actual changes)
4. What it SLOWLY stop each service one by one instead of stopping all at once and taking a long time

Note: As previously mentioned, we have `depend_on` conditions on our services, example :

```
    depends_on:
      migrations:
        condition: service_completed_successfully
      kafka:
        condition: service_healthy
      rabbit:
        condition: service_healthy
```

### Compose Version

```Text
Docker Compose version v2.30.3-desktop.1

I am on a MacBook M3 Pro running Sequoia 15.1.1
```

### Docker Environment

```Text
➜ docker info
Client:
 Version:    27.3.1
 Context:    desktop-linux
 Debug Mode: false
 Plugins:
  ai: Ask Gordon - Docker Agent (Docker Inc.)
    Version:  v0.1.0
    Path:     /Users/joaoc/.docker/cli-plugins/docker-ai
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.18.0-desktop.2
    Path:     /Users/joaoc/.docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.30.3-desktop.1
    Path:     /Users/joaoc/.docker/cli-plugins/docker-compose
  debug: Get a shell into any image or container (Docker Inc.)
    Version:  0.0.37
    Path:     /Users/joaoc/.docker/cli-plugins/docker-debug
  desktop: Docker Desktop commands (Alpha) (Docker Inc.)
    Version:  v0.0.15
    Path:     /Users/joaoc/.docker/cli-plugins/docker-desktop
  dev: Docker Dev Environments (Docker Inc.)
    Version:  v0.1.2
    Path:     /Users/joaoc/.docker/cli-plugins/docker-dev
[...]

PR #12322, opened 2024-11-25
use service.stop to stop dependent containers

ndeloof:
**What I did**
as we detect a service must be recreated, use `composeService.stop` to stop them following dependency tree and with concurrent operations where possible. 

**Related issue**
fixes https://github.com/docker/compose/issues/12310


glours (approved):
👌

glours (commented):
CI did the job 😅

PR #12325, opened 2024-11-26
run build tests against bake

ndeloof:
image


glours (approved):
🥳 👍

PR #12348, opened 2024-12-04
Make e2e tests pass locally

glours:
**What I did**
Make sure the existing docker local config is passed to the e2e config to avoid pull rate limitation
Be sure all resources have been cleanup after each test

image


ndeloof (approved):

ndeloof:
🥳

PR #12349, opened 2024-12-04
disable failing TestBuildSSH test

ndeloof:
**What I did**

**Related issue**


glours (approved):
LGTM, buildx `v0.18.1` including the fix for the `--no-cache` option has just been released

glours:
https://github.com/docker/compose/pull/12350 should fix the problem without having to bump to latest buildkit version (OTel deps version differs)

PR #12363, opened 2024-12-09
Recreate container on volume configuration change

ndeloof:
**What I did**
supersede https://github.com/docker/compose/pull/12223
- detect actual volume doesn't match configuration
- request user to confirm dropping volume before recreation, until ran with `--yes`
- recreate container if the mounted volume is not the expected one

drawbacks:
- anonymous volumes from previous container can't be re-attached to recreated one, as we have to remove container (we miss a `VolumeRename` API)

**Related issue**
fixes https://github.com/docker/compose/issues/10060

### DIFF
--- a/cmd/compose/create.go
+++ b/cmd/compose/create.go
@@ -46,6 +46,7 @@ type createOptions struct {
 	timeout       int
 	quietPull     bool
 	scale         []string
+	AssumeYes     bool
 }
 
 func createCommand(p *ProjectOptions, dockerCli command.Cli, backend api.Service) *cobra.Command {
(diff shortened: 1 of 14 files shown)


glours (approved):
Sounds good to me and tests ok 👌

jhrotko:
Nice!

jonasgeiler:
Why is the long form of the `-y` flag just `--y`? Is that a typo or intended?

jonasgeiler:
@ndeloof Why is the long form of the `-y` flag just `--y`? Is that a typo or intended?
Shouldn't it be `--yes`??

PR #12365, opened 2024-12-10
format code with gofumpt

thaJeztah:
Format the code  with gofumpt to prevent my IDE from reformatting every time I open a file. gofumpt provides a superset of gofmt, so should not impact users that are not using gofumpt.

**What I did**

**Related issue**


ndeloof:
Could you add a make target so any contributor can run gofumpt and get the same formatting applied ?

thaJeztah:
closing; this was superseded by https://github.com/docker/compose/pull/12392

Thanks for carrying @ndeloof !

Issue #12364, opened 2024-12-10
[BUG] MAC address changed with the update


LordMilutin:
### Description

Hello.
I have a docker-compose file that has fixed MAC address.
Everything worked flawlessly until recently which I suspect is the new docker update.

Now, when I redeploy the container (docker-compose down && docker-compose up), my fixed MAC address is changed.

I can still see the old one in one part of docker inspect command, but under Networks, it is another one. This used to be the fixed address, however that is not the case anymore.

Is there anything that can be done to preserve that fixed address again under Network, since our licences are based on MAC addresses.

My docker-compse:

```
services:
  serx:
    image: registry.my.ap/app:release
    container_name: serx
    restart: always
    ports:
      - "9090:9090"
    deploy:
      resources:
        limits:
          memory: 4G
    mac_address: 00:e0:84:35:d0:e8
```

Running docker inspect serx outputs this, and you can clearly see two MacAddress values, one that I set and one that is randomly generated:

```
[
    {
        "Id": "1d320278517429b8380ab19522908c3f6d8f0752658c26dca7802128cf028277",
        "Created": "2024-12-10T08:05:14.097145494Z",
        "Path": "/bin/sh",
        "Args": [
            "-c",
            "/bin/bash \"/home/serx/server/serx.sh\" \"console\""
        ],
        "State": {
            "Status": "running",
            "Running": true,
            "Paused": false,
            "Restarting": false,
            "OOMKilled": false,
            "Dead": false,
            "Pid": 31982,
            "ExitCode": 0,
            "Error": "",
            "StartedAt": "2024-12-10T08:10:50.905014863Z",
            "FinishedAt": "2024-12-10T08:09:50.818190984Z"
        },
        "Image": "sha256:e37110e899db5af9c6edea669262f2e9d4a5914fb51edf59b6b19bd5b1b604f3",
        "ResolvConfPath": "/var/lib/docker/containers/1d320278517429b8380ab19522908c3f6d8f0752658c26dca7802128cf028277/resolv.conf",
        "HostnamePath": "/var/lib/docker/containers/1d320278517429b8380ab19522908c3f6d8f0752658c26dca7802128cf028277/hostname",
        "HostsPath": "/var/lib/docker/containers/1d320278517429b8380ab19522908c3f6d8f0752658c26dca7802128cf028277/hosts",
[...]

PR #12371, opened 2024-12-10
fix support for service.mac_address

ndeloof:
**What I did**

**Related issue**
fixes https://github.com/docker/compose/issues/12364


glours (approved):
LGTM

Issue #12383, opened 2024-12-14
[BUG] 2.32.0 may have a volume persistence issue


sayhiben:
### Description

Unfortunately, I haven't yet had time to come up with a solid explanation or understanding of what's happening, so I don't have good repro steps. That said, the issue is severe enough for my org. that I wanted to report the problem here before I stopped work for the weekend:

My team manages a Django application backed by MySQL. Each service runs in a separate container orchestrated by Docker Compose on GitHub Codespaces using their Docker-in-Docker devcontainer feature. This devcontainer feature defaults to the latest version of Docker Compose when creating new Codespaces. As of this morning, we have been unable to persist data in MySQL's attached volume, and the container seems to restart and/or be recreated unnecessarily

Here's the MySQL Dockerfile we pull: https://github.com/docker-library/mysql/blob/090eb25ac69bca920fc5320484bc35aac92a8143/8.0/Dockerfile.debian

Note the volume in the above Dockerfile, which is where MySQL databases are persisted

Our in-house reproduction steps have been to: 
1. Get a shell_plus session in our Django app inside the container
2. Create instances of our models and save them
3. Exit the shell_plus session and container
4. Get another shell_plus session
5. Try to load the previously created instance from MySQL
6. Data miss

After manually downgrading back to Docker Compose 2.31.0, I can no longer reproduce the data persistence issue and see fewer recreates/restarts when running `docker compose` commands.

I apologize for the lack of specifics in this report, but again, given the instability, I figured that perhaps Docker Compose maintainers would want to hear from a canary sooner than later

### Steps To Reproduce

I'm unsure how reproducible this is outside my specific set of Dockerfiles and Docker Compose configurations, but I'd test with using a MySQL 8 image based on `mysql:8-debian` and see if you can persist data in the volume it creates using Docker Compose 2.32.0

### Compose Version

```Text
2.32.0
```

### Docker Environment

```Text
Client:
 Version:    27.3.1-1
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.19.2
    Path:     /usr/libexec/docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.31.0
    Path:     /usr/libexec/docker/cli-plugins/docker-compose

Server:
 Containers: 11
  Running: 11
  Paused: 0
  Stopped: 0
 Images: 15
 Server Version: 27.3.1-1
[...]

PR #12386, opened 2024-12-16
only check volume mounts for updated config

ndeloof:
**What I did**
Compose uses a config hash to detect service need to be recreated, but as compose file uses reference names for volumes (which may be != volume name), we have to check if service volume is actually used to mount the desired volume.
This only has to apply to `volume` mounts, where this reference ambiguity exisrts - otherwise we detect a false positive and recreate container while it's up-to-date.

**Related issue**
fixes https://github.com/docker/compose/issues/12383

468959073_1221260192696349_5555675514319708470_n


glours (approved):
LGTM

PR #12391, opened 2024-12-16
Condense output of `compose top`

### LINKED ISSUE #12381: Condense output of `docker compose top`
### Description

In a current compose project, `docker compose top` reports the following:

```console
$ docker compose top
myproject-converters-1
UID    PID      PPID     C    STIME   TTY   TIME       CMD
root   336052   335954   0    11:18   ?     00:00:01   ruby bin/server

myproject-redis-ephemeral-1
UID        PID      PPID     C    STIME   TTY   TIME       CMD
systemd+   336001   335887   0    11:18   ?     00:00:05   valkey-server *:6379

myproject-redis-persisted-1
UID    PID      PPID     C    STIME   TTY   TIME       CMD
root   336038   335921   0    11:18   ?     00:00:08   valkey-server *:6379

myproject-sidekiq-default-1
UID      PID      PPID     C    STIME   TTY   TIME       CMD
appuser   338516   338496   19   12:00   ?     00:00:09   sidekiq 7.3.6 default [0 of 3 busy]

myproject-watchtower-1
UID    PID      PPID     C    STIME   TTY   TIME       CMD
root   335931   335850   0    11:18   ?     00:00:00   /watchtower --label-enable --include-stopped --revive-stopped --cleanup --rolling-restart --stop-timeout=80s --http-api-update

myproject-web-1
UID      PID      PPID     C    STIME   TTY   TIME       CMD
appuser   338619   338597   0    12:00   ?     00:00:00   /bin/bash -e /rails/bin/docker-entrypoint.sh rails server
appuser   338653   338619   0    12:00   ?     00:00:00   flock -x /rails/data/migration-check.lock bundle exec rake db:prepare
appuser   338665   338653   46   12:00   ?     00:00:01   /usr/local/bundle/ruby/3.2.0/bin/rake db:prepare
```

I find this a bit wasteful, in regards to the screen estate. The output is also hard to grasp because the columns are not aligned across the various services.

Would you accept a PR to format the output to something like the following?

```console
$ docker compose top
SERVICE                       UID        PID      PPID     C    STIME   TTY   TIME       CMD
myproject-converters-1        root       336052   335954   0    11:18   ?     00:00:01   ruby bin/server
myproject-redis-ephemeral-1   systemd+   336001   335887   0    11:18   ?     00:00:05   valkey-server *:6379
myproject-redis-persisted-1   root       336038   335921   0    11:18   ?     00:00:08   valkey-server *:6379
myproject-sidekiq-default-1   appuser    338516   338496   19   12:00   ?     00:00:09   sidekiq 7.3.6 default [0 of 3 busy]
myproject-watchtower-1        root       335931   335850   0    11:18   ?     00:00:00   /watchtower --label-enable --include-stopped --revive-stopped --cleanup --rolling-restart --stop-timeout=80s --http-api-update
myproject-web-1               appuser    338619   338597   0    12:00   ?     00:00:00   /bin/bash -e /rails/bin/docker-entrypoint.sh rails server
myproject-web-1               appuser    338653   338619   0    12:00   ?     00:00:00   flock -x /rails/data/migration-check.lock bundle exec rake db:prepare
myproject-web-1               appuser    338665   338653   46   12:00   ?     00:00:01   /usr/local/bundle/ruby/3.2.0/bin/rake db:prepare
```

dmke:
**What I did**

This changes the output format of `compose top` and inlines the service container name into the table.

Previously, `compose top` had printed something like:

    <service name>
    UID    PID   ...
    root   1     ...

Now, the output looks more like this:

    SERVICE   UID    PID   ...
    <name>    root   1     ...

**Related issue**

Closes #12381 

image

*Baby mountain goat on rock, CC-by-sa [Shaundd](https://commons.wikimedia.org/wiki/User:Shaundd), [WP:Commons](https://commons.wikimedia.org/wiki/File:Baby_mountain_goat_on_rock.JPG)*

### DIFF
--- a/cmd/compose/top.go
+++ b/cmd/compose/top.go
@@ -49,6 +49,9 @@ func topCommand(p *ProjectOptions, dockerCli command.Cli, backend api.Service) *
 	return topCmd
 }
 
+type topHeader map[string]int // maps a proc title to its output index
+type topEntries map[string]string
+
 func runTop(ctx context.Context, dockerCli command.Cli, backend api.Service, opts topOptions, services []string) error {
 	projectName, err := opts.toProjectName(ctx, dockerCli)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


dmke:
This is a first draft. The tests could be a bit more exhaustive (i.e. mocking `(api.Service).Top()`).

Looking at the test cases now, I also think it should ensure to reorder the CMD column to the end.

cmd/compose/top.go
```diff
@@ -63,30 +66,60 @@ func runTop(ctx context.Context, dockerCli command.Cli, backend api.Service, opt
...
-		_, _ = fmt.Fprintf(dockerCli.Out(), "%s\n", container.Name)
-		err := psPrinter(dockerCli.Out(), func(w io.Writer) {
-			for _, proc := range container.Processes {
-				info := []interface{}{}
-				for _, p := range proc {
-					info = append(info, p)
-				}
-				_, _ = fmt.Fprintf(w, strings.Repeat("%s\t", len(info))+"\n", info...)
+		for _, proc := range container.Processes {
+			entry := topEntries{"SERVICE": container.Name}
```
ndeloof, line 83: container name is not service name - you should use container label `com.docker.compose.service` to retrieve service name (and make display shorter)

ndeloof (commented):
LGTM, just a comment regarding service vs container name

cmd/compose/top.go
dmke: But then I should probably also include `com.docker.compose.container-number`, no?

Should I recombine them (service + "-" + container-number), or should this be a separate column (titled "#" to keep it short)?

ndeloof: a separate column for replica # makes sense

dmke: Hm. Apart from `(api.Service).Ps()`, no other method returns the labels.

AFAICT, I have these options:

- Just call `Ps()` and discard all but the labels.
- Extend `api.ContainerProcSummary` with `ServiceName` and `ReplicaID` fields (and adjust `(*composeService).Top()` to fill these).
- Extend `api.ContainerProcSummary` with a `Labels` field (and adjust `(*composeService).Top()` to fill it).
- Keep `container.Name` as-is.

Any preference?

dmke: I'm tending to prefer option (2) or (3), as `(*composeService).Top()` already has the labels available.

Option (1) would cause additional API calls (looks like one for each container, not sure).

dmke: There's also a `getContainerNameWithoutProject()` helper function, which is used in various other places.

Maybe this is a better choice (in terms of consistency) for `(ContainerProcSummary).Name`, as it defaults to `getCanonicalContainerName()` (the current value for `(ContainerProcSummary).Name`)?

ndeloof: extend `ContainerProcSummary` to match your needs. `getContainerNameWithoutProject` is not 100% safe as service may declare a `container_name` that can confuse this function.

dmke: I'm now passing the container labels unchanged to the commandlet.

- It doesn't make much sense filtering the labels in `(*composeService).Top()` (any other consumer would wonder where the other labels are).
- Promoting the service and container-number labels to proper fields on `ContainerProcSummary` also feels arbitrary. Also, the container-number ought to be of type `int`, which implies `strconv` both from and to string...

PR #12392, opened 2024-12-17
Gofumpt all the things

ndeloof:
same as https://github.com/docker/compose/pull/12365 + make target

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -10,7 +10,7 @@ linters:
     - errorlint
     - gocritic
     - gocyclo
-    - gofmt
+    - gofumpt
     - goimports
     - gomodguard
     - revive
--- a/Makefile
+++ b/Makefile
@@ -116,6 +116,11 @@ cache-clear: ## Clear the builder cache
 lint: ## run linter(s)
 	$(BUILDX_CMD) bake lint
 
+.PHONY: fmt
+fmt:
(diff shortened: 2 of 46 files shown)


glours (approved):
LGTM
Open question, should not we add a check of the formatting during the validation step in the CI? 🤔

.github/workflows/ci.yml
```diff
@@ -50,6 +50,7 @@ jobs:
           - validate-go-mod
           - validate-headers
           - validate-docs
+          - validate-fmt
```
thaJeztah, line 53: FWIW; there's a golangci-lint linter for `gofumpt`; https://golangci-lint.run/usage/linters/#gofumpt

thaJeztah:
Thanks! I started to add the same logic as we had in docker/cli in the Makefile, which does a "optimistic" try to see if `gofumpt` is installed, but then wasn't sure if that's what you wanted, or if you preferred a Dockerfile-based / bake target instead https://github.com/docker/cli/blob/8454e012cadeca7719f170012c5331655ed5e5ec/Makefile#L52-L58

.github/workflows/ci.yml
ndeloof: nice, makes things even simpler

thaJeztah (approved):
LGTM

Issue #12400, opened 2024-12-18
[BUG] Compose Watch `restart` action does not monitor bind mounts


bsuth:
### Description

When using `action: restart`, specifying a `path` that is a bind mount results in the following error:

```
WARN[0000] path '/home/...' also declared by a bind mount volume, this path won't be monitored! 
```

### Steps To Reproduce

1. `compose.yaml`

```yaml
services:
  test:
    image: bash
    command: sleep 1m
    volumes:
      - .:/test
    develop:
      watch:
        - path: .
          action: restart
```

### Compose Version

```Text
Docker Compose version 2.32.1
```

### Docker Environment

```Text
Client:
 Version:    27.3.1
 Context:    default
 Debug Mode: false
 Plugins:
  compose: Docker Compose (Docker Inc.)
    Version:  2.32.1
    Path:     /usr/lib/docker/cli-plugins/docker-compose

Server:
 Containers: 4
  Running: 0
  Paused: 0
  Stopped: 4
 Images: 12
 Server Version: 27.3.1
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: true
  Native Overlay Diff: false
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 88aa2f531d6c2922003cc7929e51daf1c14caa0a.m
 runc version: 
 init version: de40ad0
 Security Options:
  seccomp
   Profile: builtin
  cgroupns
 Kernel Version: 6.12.4-arch1-1
 Operating System: Arch Linux
 OSType: linux
 Architecture: x86_64
 CPUs: 12
 Total Memory: 15.23GiB
 Name: bsuth
 ID: 0d2ac96b-cd23-401e-be0c-c394f4021f3b
 Docker Root Dir: /var/lib/docker
 Debug Mode: false
 Experimental: false
 Insecure Registries:
  127.0.0.0/8
 Live Restore Enabled: false

WARNING: bridge-nf-call-iptables is disabled
WARNING: bridge-nf-call-ip6tables is disabled
```

### Anything else?

[...]

PR #12409, opened 2024-12-19
only check bind mount conflict if sync action is involved

ndeloof:
**What I did**

**Related issue**
fixes https://github.com/docker/compose/issues/12400


glours (approved):
LGTM 👍

PR #12439, opened 2025-01-07
Only override service mac if set on the main network.

apollo13:
**What I did**

I am using compose against podman which exposes API 1.41 and as such does not yet support mac per network. But the existing code unconditionally overrides the mac address set on the service level with the mac from the main network even if none is defined -- hence only override if it is actually set.

**Related issue**


apollo13:
Thank you for the quasi instant turnaround @ndeloof. Much appreciated. I guess asking for a new point release is a bit much given there was one an hour ago? 😀 Do you have a rough idea when you want to do the next release? I am asking because then I'd know whether I will just put my podman PRs on hold for now or look for a workaround in the meantime. No pressure though, if you can give me an educated guess that would be more than helpful!

Issue #12440, opened 2025-01-07
[BUG] run for service with depends_on: service_healthy checks status of another run instance


chrkr:
### Description

When I try `run` on a service which has a `depends_on: condition: service_healthy`, compose also checks the health status of `run` instances of the dependency service. The health check is likely failing on the dependency one-off `run` instance, which is then preventing the start of the `run`.

I expected compose to ignore any `run` instances and only check the 'regular' service, e.g. started with `up`.

The `run` in the dependency is _not_ preventing the dependent service itself from starting with e.g. `up`, this only happens for `run`. Having _only_ a `run` in the dependency is also not sufficient for compose, as it still starts the 'regular' dependency service if not running.

### Steps To Reproduce

1. Using docker desktop v4.37.1 on Windows 11, (same effect if used from an Ubuntu 22.04 WSL2 session on this machine). I have a postgres container which has a health check using `pg_isready` with `-h localhost` to ensure it is checking that the _actual_ database is running. Using `pg_isready` on the default unix socket will succeed also during the `initdb` stage, which would start dependent services while the database is not actually available yet.
4. Configuration:
```yaml
services:
  db:
    image: postgres:17
    environment:
      POSTGRES_PASSWORD: mypassword
    healthcheck:
      # Healthcheck verifies that postgres is ready on the TCP port
      test: ["CMD-SHELL", "pg_isready -q -h localhost -d postgres"]
      interval: 1s
      timeout: 3s
      retries: 3
      start_period: 30s
  
  bash:
    image: bash:latest
    depends_on:
      db:
        condition: service_healthy
```
3. Run commands:
```
docker compose run --rm -d db bash -c "while true; do sleep 1; done"
docker compose run --rm bash
```
4. Error:
```
docker compose run --rm bash
[+] Creating 1/0
 ✔ Container depends-db-1  Created                                                                                 0.1s
[+] Running 1/1
 ✔ Container depends-db-1  Started                                                                                 0.4s
dependency failed to start: container depends-db-run-39415395b84c is unhealthy

<...>

docker compose run --rm bash
[+] Creating 1/0
 ✔ Container depends-db-1  Running                                                                                                                                                                          0.0s
dependency failed to start: container depends-db-run-9bb91edee49f is unhealthy
```

PS status:
```
docker compose ps
NAME           IMAGE         COMMAND                  SERVICE   CREATED         STATUS                   PORTS
[...]

PR #12441, opened 2025-01-08
exclude one-off container running convergence

ndeloof:
**What I did**

**Related issue**
fixes https://github.com/docker/compose/issues/12440


glours (approved):
LGTM

Issue #12370, opened 2024-12-10
[BUG] Overlapping rows


MattiasOz:
### Description

When docker-compose is building images the printout seems to be confused about which rows to use. It prints out, e.g., both a Dockerfile instruction, like COPY, as well as the current container status on the same row. This results in a constant switching between what is written. 

### Steps To Reproduce

1. Enter a directory with a `docker-compose.yml` file
2. Run `docker-compose build --no-cache`

### Compose Version

```Text
`docker compose version`:
`Docker Compose version 2.31.0`

`docker-compose version`:
`Docker Compose version 2.31.0`
```

### Docker Environment

```Text
Client:
 Version:    27.3.1
 Context:    default
 Debug Mode: false
 Plugins:
  compose: Docker Compose (Docker Inc.)
    Version:  2.31.0
    Path:     /usr/lib/docker/cli-plugins/docker-compose

Server:
 Containers: 4
  Running: 3
  Paused: 0
  Stopped: 1
 Images: 978
 Server Version: 27.3.1
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: true
  Native Overlay Diff: false
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 207ad711eabd375a01713109a8a197d197ff6542.m
 runc version: 
 init version: de40ad0
 Security Options:
  seccomp
   Profile: builtin
  cgroupns
 Kernel Version: 6.6.63-1-MANJARO
 Operating System: Manjaro Linux
 OSType: linux
 Architecture: x86_64
 CPUs: 16
 Total Memory: 13.34GiB
 Name: Saturn
 ID: e1c30eb4-300f-4d4d-97d4-2cdbd78aeca8
 Docker Root Dir: /var/lib/docker
 Debug Mode: false
 Experimental: false
 Insecure Registries:
  127.0.0.0/8
 Live Restore Enabled: false
```

### Anything else?

I don't really know what caused it to happen. I know the steps to reproduce aren't satisfactory but I'm not sure what else to write. Any insight regarding the printout is appreciated. 
[...]

PR #12442, opened 2025-01-08
can't render progress concurrently with buildkit

ndeloof:
**What I did**
removed "building" events are the progress UX collides with buildkit on terminal
only keep "built" events after all images have been built as a general status after build succeeded

**Related issue**
fixes https://github.com/docker/compose/issues/12370


glours (approved):
LGTM and tests ok 👍

Issue #12443, opened 2025-01-08
[BUG] 2.31.0 broke referencing local image IDs


gzm0:
### Description

Using an image ID (SHA 256) as `image` worked until compose 2.29.7.

```yaml
services:
  test:
    image: f691df868290275974103c307537784a0e457ecfd8f188e06e52b8de8d1439f8
```

On 2.27.9:

```
$ docker compose up -d
[+] Running 2/2
 ✔ Network hex-string-test_default   Created                                                                                                                     0.1s 
 ✔ Container hex-string-test-test-1  Started    
```

On 2.31.0 (and later):

```
$ docker compose up -d
invalid repository name (f691df868290275974103c307537784a0e457ecfd8f188e06e52b8de8d1439f8), cannot specify 64-byte hexadecimal strings
```

I suspect
- #12278 

as the culprit, or more specifically, the now unconditional call to `reference.ParseDockerRef`:

https://github.com/docker/compose/blob/a8469db83f514a5abe4681c7fee773061f1941c6/pkg/compose/images.go#L97

Related:
- #923

### Steps To Reproduce

1. Run `docker image ls --digests`, pick the full digest of any image (without the `sha256:` prefix).
2. Create a `docker-compose.yml` like above (but with the digest you picked in the previous step).
3. Run `docker compose up`

Outcome:

```
$ docker compose up -d
invalid repository name (f691df868290275974103c307537784a0e457ecfd8f188e06e52b8de8d1439f8), cannot specify 64-byte hexadecimal strings
```

Expected outcome: Compose starts a container with the image having this ID.

### Compose Version

```Text
Docker Compose version v2.32.1
```

### Docker Environment

```Text
Client: Docker Engine - Community
 Version:    27.4.1
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.19.3
    Path:     /usr/libexec/docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.32.1
    Path:     /usr/libexec/docker/cli-plugins/docker-compose

Server:
 Containers: 11
  Running: 1
  Paused: 0
  Stopped: 10
 Images: 118
 Server Version: 27.4.1
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: systemd
[...]

PR #12446, opened 2025-01-09
image can be set to a local ID, that isn't a valid docker ref

ndeloof:
**What I did**
`ParseDockerRef` rejects a 64-digits local image ID, whenever one can execute `docker run <ID>` without such a reference being rejected

**Related issue**
fixes https://github.com/docker/compose/issues/12443


glours (approved):
LGTM

PR #12448, opened 2025-01-09
Update config/secret handling: Copy files into containers instead of bind mounting

### LINKED ISSUE #11867: Support remote DOCKER_HOST by copying configs and secrets files instead of bind mounting them

schaubl:
Currently, Docker Compose copies a config or secret into the container if the source is an environment variable or inlined content. However, if the source is a file, the file is bind mounted into the container. This approach fails when Docker Compose is invoked on a different filesystem than the Docker daemon host filesystem (e.g., when called from within a container or remotely using the -H option).

This PR changes the behavior for file-based config/secret sources to match the handling of environment variables and inlined content. Specifically, it reads the file from the filesystem where Docker Compose is executed (like Docker Stack) and copies it into the container.

This update enhances the functionality of Docker Compose when run "remotely," providing the ability to push configuration files, a feature previously available only with Docker Stack/Swarm.

What I did:
- Updated the behavior for handling configs and secrets sourced from files.
- Ensured consistency with the documentation and the behavior of Docker Stack.
- Enabled Docker Compose to push configuration files when run remotely, aligning with Docker Stack/Swarm capabilities.

Notes:
1. It fixes #11867
2. It is a reopening of #11984 based on the latest code base
3. It is an alternative to #12251
Differences it that this PR checks the validity of the config/secret parameters in `getCreateConfigs` before the container is created where the other one does it after.
Also this PR doesn't support directories which correspond to the doc and the behaviour of Docker Configs&Secrets used by Swarm/Stack.


polarathene:
> This PR changes the behavior for file-based config/secret sources to match the handling of environment variables and inlined content.

Question, will this allow file sources to leverage [`uid`, `gid`, and `mode`](https://docs.docker.com/reference/compose-file/services/#long-syntax) long options for `config` like already supported for `content` and `environment`? That's been a long standing caveat with `file` for secrets in compose when not using swarm IIRC.

ndeloof:
This PR doesn't address comments on https://github.com/docker/compose/issues/11867
tl;dr: 
config/secret file _MAY_ be mounted from (remote) docker host. Many users rely on this, some do even require changes to the original file to be propagated inside container. So we can't just switch to a distinct behavior which would introduce a breaking change. This would _at least_ require an explicit opt-in flag in compose.yaml

ndeloof:
@polarathene yes indeed, this is the main reason for this approach to be experimented

schaubl:
> This PR doesn't address comments on #11867 tl;dr: config/secret file _MAY_ be mounted from (remote) docker host. Many users rely on this, some do even require changes to the original file to be propagated inside container. So we can't just switch to a distinct behavior which would introduce a breaking change. This would _at least_ require an explicit opt-in flag in compose.yaml

@ndeloof I think if the goal of a user is to mount files from the (remote) docker host, he should use bind mounts.
I explained my rationale more in detail in the following comment on the original GH issue (https://github.com/docker/compose/issues/11867#issuecomment-2232745553) and on my first PR (https://github.com/docker/compose/pull/11984#issue-2410767140) and the [first comment on the current PR](https://github.com/docker/compose/pull/12448#issue-2778344884).
As mentioned in my comment on #11867, I think this `is a breaking change that "fix" the initial breaking change`.
I share your view on [your first comment on the GH issue](https://github.com/docker/compose/issues/11867#issuecomment-2141310118), which says that this change of behaviour should be mentioned in the release notes.

@andoks @LaXiS96 @domsew I'm mentioning you here because you showed interest into this and you might want to test this branch and then maybe provide feedback.

pkg/compose/create.go
```diff
@@ -952,135 +958,81 @@ func fillBindMounts(p types.Project, s types.ServiceConfig, m map[string]mount.M
...
-			return nil, err
-		}
-		mounts[target] = mnt
 	}
-	values := make([]mount.Mount, 0, len(mounts))
-	for _, v := range mounts {
-		values = append(values, v)
-	}
-	return values, nil
+	return nil
```
vahan-sahakyan-op, line 1035: just curious - why not keeping the convention of `value, error` instead of `value or error`

schaubl: Hello,
It's not `value or error` but just `error` if any, `nil` otherwise.
This function now only do checks, it doesn't return any value (its name&definition has been changed accordingly).
See point "_9. Return nil for Success_" of [Best Practices For Error Handling in Go](https://www.geeksforgeeks.org/best-practices-for-error-handling-in-go/).

PR #12452, opened 2025-01-10
ci: update bake-action to v6

crazy-max:
Update back-action to v6 which defaults to Git context so need some changes in our workflows.


glours (approved):
LGTM

Issue #12457, opened 2025-01-13
[BUG] Compose does not observe the BUILDKIT_PROGRESS environement variable


freesteph:
### Description

Buildkit provides [a `--progress` option which can also be set via the `BUILDKIT_PROGRESS` environment variable.](https://docs.docker.com/reference/cli/docker/buildx/build/#progress). 

The `--progress` flag is effectively observed by Compose, but not the environment variable. For consistency and ease of configuration it would be nice if it did.

### Steps To Reproduce

1. Run `docker compose build --progress=plain`
2. Enjoy the simpler output in environments where TERM=dumb;
3. Try and configure it for good by setting `BUILDKIT_PROGRESS=plain`;
4. But `docker compose build` is back to fancy output.

### Compose Version

```Text
Docker Compose version v2.31.0-desktop.2
```

### Docker Environment

```Text
Client:
 Version:    27.4.0
 Context:    desktop-linux
 Debug Mode: false
 Plugins:
  ai: Ask Gordon - Docker Agent (Docker Inc.)
    Version:  v0.5.1
    Path:     /Users/steph/.docker/cli-plugins/docker-ai
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.19.2-desktop.1
    Path:     /Users/steph/.docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.31.0-desktop.2
    Path:     /Users/steph/.docker/cli-plugins/docker-compose
  debug: Get a shell into any image or container (Docker Inc.)
    Version:  0.0.37
    Path:     /Users/steph/.docker/cli-plugins/docker-debug
  desktop: Docker Desktop commands (Beta) (Docker Inc.)
    Version:  v0.1.0
    Path:     /Users/steph/.docker/cli-plugins/docker-desktop
  dev: Docker Dev Environments (Docker Inc.)
    Version:  v0.1.2
    Path:     /Users/steph/.docker/cli-plugins/docker-dev
  extension: Manages Docker extensions (Docker Inc.)
    Version:  v0.2.27
    Path:     /Users/steph/.docker/cli-plugins/docker-extension
  feedback: Provide feedback, right in your terminal! (Docker Inc.)
    Version:  v1.0.5
    Path:     /Users/steph/.docker/cli-plugins/docker-feedback
  init: Creates Docker-related starter files for your project (Docker Inc.)
    Version:  v1.4.0
[...]

PR #12458, opened 2025-01-13
add support for BUILDKIT_PROGRESS

ndeloof:
fixes https://github.com/docker/compose/issues/12457

### DIFF
--- a/pkg/compose/build.go
+++ b/pkg/compose/build.go
@@ -129,6 +129,9 @@ func (s *composeService) build(ctx context.Context, project *types.Project, opti
 		if options.Quiet {
 			options.Progress = progress.ModeQuiet
 		}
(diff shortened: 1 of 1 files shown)


glours (approved):
LGTM

pkg/compose/build.go
```diff
@@ -129,6 +129,9 @@ func (s *composeService) build(ctx context.Context, project *types.Project, opti
 		if options.Quiet {
 			options.Progress = progress.ModeQuiet
 		}
+		if options.Progress == "" {
```
willnode, line 132: I believe this doesn't work since the default value is `string(buildkit.AutoMode)`.

My docker compose is v2.33.1 and this envar doesn't work.

Issue #12427, opened 2025-01-02
[BUG] Docker Compose watch fails to replace new containers if change affects multiple containers


joaomcarlos:
### Description

Hi,

I am having an issue where I have setup watch rules for changes in a common dependency should trigger a container rebuild on all services that depend upon it.

However, once I do this and I make a simple change on "lib common", the rebuild will eventually fail with the message `WARN[0171] Error handling changed files for service <service1>: Error response from daemon: Error when allocating new name: Conflict. The container name "/<proj>-<service1>-1" is already in use by container "862f5bff1233aebc766fe966b06cc0e491eedbf4f475cd169fa7c6af10ff752b". You have to remove (or rename) that container to be able to reuse that name`

### Steps To Reproduce

__The stage__
Have a project with a `packages/common` and several `project/service1/2/3/4/etc` projects.

At the root, there is a single `docker-compose.yml` with a list of included `docker-compose.yml` entries, one for each service.

Each of the services docker compose has a watch rule for `rebuild` when lib-common code changes.

```
    develop:
      watch:
        - action: sync+restart
          path: api
          target: /app/api
        - action: rebuild
          path: pdm.lock
          target: /app/pdm.lock
        - action: rebuild
          path: ../../packages/common/common
```

__Steps__

1- `docker compose watch` to build and watch all services
2- Change anything in the `common` package, so that it triggers a multi-service rebuild
3- Watch and wait until all services finish build
4- Watch as multiple error messages appear, as it fails to replace the running containers

__The error__
When building finishes, I get multiple: 

`WARN[0171] Error handling changed files for service <service1>: Error response from daemon: Error when allocating new name: Conflict. The container name "/<proj>-<service1>-1" is already in use by container "862f5bff1233aebc766fe966b06cc0e491eedbf4f475cd169fa7c6af10ff752b". You have to remove (or rename) that container to be able to reuse that name`

And none of those services that error will now be running. There will be multiple stopped containers with "uuid" prefixes and one or none of the services that build will be running.

### Compose Version

```Text
➜ docker compose version
Docker Compose version 2.32.0



➜ docker-compose version
Docker Compose version v2.31.0-desktop.2
```

### Docker Environment

```Text
➜ docker info
Client:
 Version:    27.4.0
 Context:    desktop-linux
 Debug Mode: false
 Plugins:
  ai: Ask Gordon - Docker Agent (Docker Inc.)
    Version:  v0.5.1
[...]

PR #12469, opened 2025-01-15
manage `watch` applied to mulitple services

ndeloof:
**What I did**
`watch` used to be ran _per service_ which brings race condition when a file event has impacts on multiple services.
This PR changes the implementation so we collect file events, then check all services for watch rules matching update, and eventually apply required actions.

Updated logic is :
- create a single `watcher` for the paths declared by all services in project
- debounce raw filesystem event for batch processing
- select watch rule(s) to match updated paths; apply ignore filter _by rule_ (as distinct service may declare distinct ignores)
- apply changes to services and paths impacted

**Related issue**
closes https://github.com/docker/compose/issues/12427

image

### DIFF
--- a/internal/sync/shared.go
+++ b/internal/sync/shared.go
@@ -16,8 +16,6 @@ package sync
 
 import (
 	"context"
-
-	"github.com/compose-spec/compose-go/v2/types"
 )
 
 // PathMapping contains the Compose service and modified host system path.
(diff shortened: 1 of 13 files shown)


glours (approved):
Sounds good to me, I didn't detect any regression or behavior changes during my tests 👍 
I really like those improvements, thanks @ndeloof 🙏

PR #12482, opened 2025-01-21
add --with-env flag to publish command

glours:
this flag allow publishing env variables in the Compose OCI artifact

**What I did**
Add `--with-env` flag to let user explicitly optin the publication of env variables in the OCI artifact

**Related issue**
https://docker.atlassian.net/browse/COMP-873

image


glours:
@aevesdocker can you check if the new error message is ok for you? 😇 🙏

glours:
@ndeloof I wonder if we should not just set all env variables to empty strings when publishing? 🤔 
This way we'll push the env variables needed to run the application without sending sensitive data and we'll be able to ask consumers to set values when doing an `docker compose -f oci://... up`, wdyt?

pkg/compose/publish.go
```diff
@@ -120,3 +125,22 @@ func (s *composeService) generateImageDigestsOverride(ctx context.Context, proje
...
+func preChecks(project *types.Project, options api.PublishOptions) error {
+	if !options.WithEnvironment {
+		for _, service := range project.Services {
+			if len(service.EnvFiles) > 0 {
+				return fmt.Errorf("service %q has env_file declared. To avoid leaking sensitive data, "+
+					"you must either explicitly allow the sending of environment variables by using the --with-env flag,"+
+					" or remove sensitive data from your Compose configuration", service.Name)
+			}
+			if len(service.Environment) > 0 {
+				return fmt.Errorf("service %q has environment variable(s) declared. To avoid leaking sensitive data, "+
```
ndeloof, line 138: service.environment may be set with a fixed value, not relying on any interpolation. Typically:
```
db:
    image: mysql
    environment:
      MYSQL_DATABASE: avatar
      MYSQL_ROOT_PASSWORD_FILE: /run/secrets/db-password
```
those should not prevent compose file to be published, right ?

glours: In fact yes that should prevent from publishing by default because you can also have `MYSQL_ROOT_PASSWORD: mySuperSecretPassword` defined the same way

glours:
After internal discussions with @ndeloof , we have decided to maintain the defensive approach for now. If, based on usage and user feedback, we find this behavior too restrictive, we will keep the option of filtering only interpolated environment variables.

ndeloof (approved):

PR #12485, opened 2025-01-22
support additional_context reference to another service

ndeloof:
- classic builder doesn't support additional context
- buildkit builder detects `service:` prefix in contexts and builds images in order (in addition to runtime dependencies, to preserve backward compatibility)
- bake builder delegates to `buildx bake` with a named context for service's target

```
$ COMPOSE_BAKE=false docker compose build
INFO[0000] additional_context with "service:"" is better supported when delegating build go bake. Set COMPOSE_BAKE=true 
[+] Building 2.0s (16/16) FINISHED                                                                                                       docker:desktop-linux
 => [a internal] load build definition from Dockerfile.a                                                                                                 0.0s
...
 => => naming to docker.io/library/b:latest                                                                                                              0.0s
 => => unpacking to docker.io/library/b:latest                                                                                                           0.0s
 => [b] resolving provenance for metadata file                                                                                                           0.0s
[+] Building 2/2
 ✔ a  Built                                                                                                                                              0.0s 
 ✔ b  Built  
```

### DIFF
--- a/cmd/compose/build.go
+++ b/cmd/compose/build.go
@@ -68,17 +68,16 @@ func (opts buildOptions) toAPIBuildOptions(services []string) (api.BuildOptions,
 		uiMode = "rawjson"
 	}
 	return api.BuildOptions{
-		Pull:          opts.pull,
-		Push:          opts.push,
-		Progress:      uiMode,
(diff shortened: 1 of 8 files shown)


glours (approved):
LGTM

PR #12486, opened 2025-01-22
Publish warn display env vars

glours:
**What I did**
Add a warning message listing the env variables and asking for confirmation before publishing a Compose configuration with environment variables

Also added a `--force`/`-f` flag to force the publish without asking for confirmation

**Related issue**

https://docker.atlassian.net/browse/COMP-875

image

### DIFF
--- a/cmd/compose/publish.go
+++ b/cmd/compose/publish.go
@@ -30,6 +30,7 @@ type publishOptions struct {
 	resolveImageDigests bool
 	ociVersion          string
 	withEnvironment     bool
+	assumeYes           bool
 }
 
 func publishCommand(p *ProjectOptions, dockerCli command.Cli, backend api.Service) *cobra.Command {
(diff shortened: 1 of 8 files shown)


cmd/compose/publish.go
```diff
@@ -48,6 +49,7 @@ func publishCommand(p *ProjectOptions, dockerCli command.Cli, backend api.Servic
 	flags.BoolVar(&opts.resolveImageDigests, "resolve-image-digests", false, "Pin image tags to digests")
 	flags.StringVar(&opts.ociVersion, "oci-version", "", "OCI image/artifact specification version (automatically determined by default)")
 	flags.BoolVar(&opts.withEnvironment, "with-env", false, "Include environment variables in the published OCI artifact")
+	flags.BoolVarP(&opts.force, "force", "f", false, "Force publish without asking for confirmation")
```
ndeloof, line 52: I'd suggest to use `-y` to align with https://github.com/docker/compose/blob/main/cmd/compose/create.go#L84

pkg/compose/publish.go
```diff
@@ -130,31 +137,80 @@ func (s *composeService) generateImageDigestsOverride(ctx context.Context, proje
...
+	return envVarList, nil
+}
 
-	return nil
+func acceptPublishEnvVariables(cli command.Cli) (bool, error) {
+	_, err := fmt.Fprint(cli.Out(), "Are you ok to publish these environment variables? [y/N]: ")
+	if err != nil {
+		return false, err
+	}
+	reader := bufio.NewReader(cli.In())
```
ndeloof, line 200: could use `UI#Confirm` from pkg/prompt package

cmd/compose/publish.go
glours: Agree will update my PR 👍

pkg/compose/publish.go
glours: TIL 😉

PR #12498, opened 2025-01-25
bump docker/docker and docker/cli to get API 1.48

ndeloof:
this PR is set to test compose with next Docker engine release and moby API 1.48


thaJeztah:
Looks like this one's hitting an issue I was afraid would happen; buildkit and buildx depend on docker 27.x, and buildx has a function that's imported by compose (there was some discussion around this 😔);

```
#22 96.51 pkg/compose/build_buildkit.go:31:2: could not import github.com/docker/buildx/util/dockerutil (-: # github.com/docker/buildx/util/dockerutil

#22 96.51 /go/pkg/mod/github.com/docker/buildx@v0.20.1/util/dockerutil/client.go:55:41: cannot use false (untyped bool constant) as image.LoadOptions value in argument to dapi.ImageLoad) (typecheck)

#22 96.51 	"github.com/docker/buildx/util/dockerutil"

#22 96.51 	^
```

PR #12502, opened 2025-01-27
remove exit code per error type used by legacy metrics system

ndeloof:
**What I did**
removed custom exit code used to distinguish error cause and collect metrics by the legacy compose-cli

**Related issue**


Benehiko (commented):
Wouldn't some scripts rely on the custom exit code?

On another note: I see compose is using the `cli.StatusError` type directly which won't be possible after this is merged https://github.com/docker/cli/pull/5666

ndeloof:
@Benehiko https://github.com/docker/cli/pull/5666 will require some adjustment effort as we bump dependency
I'm not sure why we use `StatusError` directly, probably we can get rid of this as well.

glours (approved):

PR #12506, opened 2025-01-29
watch should ony build selected services

ndeloof:
.. otherwise `docker compose watch` will build all images in compose project, not just those required by `docker compose watch SERVICE`


glours (approved):
LGTM

PR #12507, opened 2025-01-29
Fix bake

ndeloof:
**What I did**
fix build delegation to bake for corner-cases detected on https://github.com/docker/compose/pull/12501

ability to build service in dependency order is preserved when `--compatibility` is set, as it would be better imho to have explicit support for this by https://github.com/docker/compose/pull/12485

see tests running on https://github.com/docker/compose/pull/12501 (some will need to be fixed as output isn't the same vs buildkit)


glours (approved):
Sounds good to me 👍

Issue #12511, opened 2025-01-31
[BUG] Severe performance loss since 4.38.0 docker dekstop upgrade on M1 MacOS


Headscracher:
### Description

Since the docker upgrade to 4.38.0 ( docker compose version v2.32.4 ) on my Macbook with a M1 chip I got a huge performance loss for any docker compose command with running containers. Just running docker compose ps takes for a single service 3 seconds
❯ time docker compose ps

```
NAME              IMAGE          COMMAND                  SERVICE   CREATED          STATUS          PORTS
coolify-redis-1   redis:latest   "docker-entrypoint.s…"   redis     34 seconds ago   Up 30 seconds   0.0.0.0:6379->6379/tcp
docker compose ps  0.06s user 0.03s system 2% cpu 3.101 total
```

while with a previous version (v2.31.0-desktop.2) it's only taking 0.09 seconds

```
NAME              IMAGE          COMMAND                  SERVICE   CREATED         STATUS         PORTS
coolify-redis-1   redis:latest   "docker-entrypoint.s…"   redis     4 minutes ago   Up 5 seconds   0.0.0.0:6379->6379/tcp
docker compose ps  0.05s user 0.03s system 94% cpu 0.090 total
```

docker ps itself isn't affected by it, it's just happening with docker compose

### Steps To Reproduce

- Run docker desktop version 4.38.0 on a M1 Mac
- Create a simple docker-compose.yml. Only one service is enough to reproduce:
```yaml
services:
  redis:
    image: redis:latest
    pull_policy: always
    ports:
      - "${FORWARD_REDIS_PORT:-6379}:6379"
    env_file:
      - .env
```
- Run a docker compose command that involves those containers like `docker compose ps `or `docker compose stop`

### Compose Version

```Text
Docker Compose version v2.32.4-desktop.1
```

### Docker Environment

```Text
Client:
 Version:    27.5.1
 Context:    desktop-linux
 Debug Mode: false
 Plugins:
  ai: Ask Gordon - Docker Agent (Docker Inc.)
    Version:  v0.7.3
    Path:     /Users/tobi/.docker/cli-plugins/docker-ai
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.20.1-desktop.2
    Path:     /Users/tobi/.docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.32.4-desktop.1
    Path:     /Users/tobi/.docker/cli-plugins/docker-compose
[...]

PR #12516, opened 2025-02-01
fix: Revert user-agent for docker cli client

idsulik:
**What I did**
Removed custom user-agent for docker cli client(that was added within this [PR](https://github.com/docker/compose/pull/11333)), because docker docker desktop's does specific logic depending on the user agent. The latest docker compose takes 3 seconds to do a request to the docker daemon.  
More details here https://github.com/docker/for-mac/issues/7575#issuecomment-2629044992

I also tried the ping method(internal/desktop/client.go:78) with the new user agent and it works fine, so decided to left it as is.

**Related issue**
https://github.com/docker/compose/issues/12511

p.s. I'm not sure should we change logic here or the docker desktop should start accepting the new user-agent, feel free to close this PR in case of the second case

@milas fyi

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -24,7 +24,6 @@ import (
 	"github.com/docker/cli/cli-plugins/plugin"
 	"github.com/docker/cli/cli/command"
 	"github.com/docker/compose/v2/cmd/cmdtrace"
-	"github.com/docker/docker/client"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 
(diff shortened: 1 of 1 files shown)


milas:
fyi I am no longer a Docker Compose maintainer / Docker Inc employee, so I can't really help here

idsulik:
closing it because of this comment https://github.com/docker/for-mac/issues/7575#issuecomment-2629071576

> Thanks for testing! No, docker compose shouldn't update; or at least, shouldn't have to because this looks like a regression in the docker desktop API proxy.
> 
> I recall that the fix at the time was fixing the immediate issue, but that there was a tracking ticket to look at ways to remove that conditional code altogether (having conditional code based on user-agent definitely isn't ideal! I recall the alternative was a much more involved change, so as things go, that was left as a backlog item).
> 
> I already posted this issue internally, but I'll try to find the right person to look into this after the weekend.

PR #12521, opened 2025-02-03
add global flag to define download directory for remote configuration (oci & git)

glours:
**What I did**
Add a `--download-dir` flag to choose the directory were remote configurations (OCI or git) will be downloaded
This flag throws an error if there is no `-f` or `--file` mentioning `oci://` or `git://` resources. 

The flag is hidden for now

**Related issue**
https://docker.atlassian.net/browse/APCLI-881

image

### DIFF
--- a/cmd/compose/compose.go
+++ b/cmd/compose/compose.go
@@ -150,6 +150,7 @@ type ProjectOptions struct {
 	Progress      string
 	Offline       bool
 	All           bool
+	DownloadDir   string
 }
 
 // ProjectFunc does stuff within a types.Project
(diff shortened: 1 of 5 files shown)


ndeloof (commented):
LGTM regarding the codebase, I just wonder this will just bring issues once we will try to offer remote loader cache management (cleanup)

glours:
After discussion with @ndeloof we think a better option, to avoid extra checks and potential side effect of adding this flag, will be to display the path of the download directory.

closed in favor of https://github.com/docker/compose/pull/12595

PR #12524, opened 2025-02-04
let user know bake is now supported

ndeloof:
**What I did**
Print a hint as we build image to let user know Compose can now use Bake

### DIFF
--- a/pkg/compose/build.go
+++ b/pkg/compose/build.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"os"
 	"strings"
+	"sync"
 
 	"github.com/compose-spec/compose-go/v2/types"
 	"github.com/containerd/platforms"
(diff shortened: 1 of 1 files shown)


pkg/compose/build.go
```diff
@@ -63,6 +65,10 @@ func (s *composeService) Build(ctx context.Context, project *types.Project, opti
 	}, s.stdinfo(), "Building")
 }
 
+const bakeSuggest = "Compose now can delegate build to bake for better performances\nJust set COMPOSE_BAKE=true"
```
ndeloof, line 68: @colinhemmings any thoughts on this wording ?

colinhemmings: Something like
`Boost build performance: enable Bake with COMPOSE_BAKE=true`

ndeloof: also, this is a hint, not advertisement we want to print on console. Just my 2 cents

PR #12530, opened 2025-02-05
don't send raw os.Args to opentelemetry but a pseudo command line

ndeloof:
**What I did**

removed `cli.args` telemetry attribute as raw os.Args is hardly comparable between executions. We already provide compose command as `cli/command` and active flags as `cli.flags`

**Related issue**


glours (approved):
LGTM

Issue #12031, opened 2024-07-29
[BUG] Unable to attach config file to read only filesystem


s4s0l:
### Description

I'm unable to attach config file to container when container has read-only filesystem.

### Steps To Reproduce

```yml
networks:
  proxy-network:
    name: proxy-network

services:
  caddy:
    image: caddy:latest
    restart: unless-stopped
    container_name: caddy
    read_only: true
    networks:
      - proxy-network
    configs:
      - source: caddy_caddyfile
        target: /etc/caddy/Caddyfile

configs:
    caddy_caddyfile:
      content: |
        does not matter

```

docker compose up ends with:

```
[+] Running 1/2
 ✔ Network proxy-network  Created                                                                                                                                                                                                                                                                                0.1s 
 ⠋ Container caddy        Creating                                                                                                                                                                                                                                                                               0.0s 
Error response from daemon: container rootfs is marked read-only

```

### Compose Version

```Text
v2.29.1
```


### Docker Environment

```Text
Client: Docker Engine - Community
 Version:    27.1.1
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.16.1
    Path:     /usr/libexec/docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.29.1
    Path:     /usr/libexec/docker/cli-plugins/docker-compose

Server:
 Containers: 1
  Running: 0
  Paused: 0
  Stopped: 1
 Images: 1
 Server Version: 27.1.1
 Storage Driver: overlay2
  Backing Filesystem: zfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 2bf793ef6dc9a18e00cb12efb64355c2c9d5eb41
 runc version: v1.1.13-0-g58aa920
 init version: de40ad0
 Security Options:
  apparmor
  seccomp
   Profile: builtin
  cgroupns
 Kernel Version: 6.8.0-38-generic
[...]

PR #12531, opened 2025-02-06
report error using non-file secret|config with read-only service

ndeloof:
**What I did**
Let user know non-file based secret/config isn't compatible with a read-only service (until engine adds support for those in non-swarm mode ?)

**Related issue**
closes https://github.com/docker/compose/issues/12031


glours (approved):
LGTM

PR #12535, opened 2025-02-07
Test commandName subcommand order

maxproske:
**What I did**

Previously, when the root CLI span `cli/<cmd>` was named, the subcommand path was returned in **reverse alphabetical order** instead of **CLI order**, contradicting the comment above the function.

**edit:** Reverted to the original sorting and updated test and function comment to clarify intended behaviour.

w460

### DIFF
--- a/cmd/cmdtrace/cmd_span.go
+++ b/cmd/cmdtrace/cmd_span.go
@@ -114,13 +114,14 @@ func wrapRunE(c *cobra.Command, cmdSpan trace.Span, tracingShutdown tracing.Shut
 	}
 }
 
-// commandName returns the path components for a given command.
+// commandName returns the path components for a given command,
+// in reverse alphabetical order for consistent usage metrics.
 //
 // The root Compose command and anything before (i.e. "docker")
 // are not included.
 //
 // For example:
-//   - docker compose alpha watch -> [alpha, watch]
+//   - docker compose alpha watch -> [watch, alpha]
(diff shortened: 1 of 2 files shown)


glours:
Hio @maxproske 
Thanks for this proposal, I think that will potentially affect the metrics history we have, @ndeloof can you double check please?

maxproske:
No worries, let me know if the original sorting is desired, and I'll update the comment/test accordingly.

cmd/cmdtrace/cmd_span.go
```diff
@@ -130,7 +129,9 @@ func commandName(cmd *cobra.Command) []string {
 		}
 		name = append(name, c.Name())
 	}
-	sort.Sort(sort.Reverse(sort.StringSlice(name)))
+	for i, j := 0, len(name)-1; i < j; i, j = i+1, j-1 {
```
ndeloof, line 132: can use `slices.Reverse` - which basically does the same, but will make the intent obvious

ndeloof:
I indeed need to check the impact on usage metrics

maxproske:
Reverted to the original sorting, and updated the tests/comment to match intended behaviour.

maxproske:
@glours Is auto-merge ready to be enabled? I morphed the PR into a test, so it won't affect the metrics history.

ndeloof (approved):

PR #12541, opened 2025-02-10
Update ls --quiet help description

maxproske:
**What I did**

`docker compose ls --help` claimed that `--quiet` displays IDs, but actually prints project names. Updated the help text to accurately reflect the output.

owl-fly-silently-5

### DIFF
--- a/cmd/compose/list.go
+++ b/cmd/compose/list.go
@@ -50,7 +50,7 @@ func listCommand(dockerCli command.Cli, backend api.Service) *cobra.Command {
 		ValidArgsFunction: noCompletion(),
 	}
 	lsCmd.Flags().StringVar(&lsOpts.Format, "format", "table", "Format the output. Values: [table | json]")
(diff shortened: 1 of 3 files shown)


cmd/compose/list.go
```diff
@@ -50,7 +50,7 @@ func listCommand(dockerCli command.Cli, backend api.Service) *cobra.Command {
 		ValidArgsFunction: noCompletion(),
 	}
 	lsCmd.Flags().StringVar(&lsOpts.Format, "format", "table", "Format the output. Values: [table | json]")
-	lsCmd.Flags().BoolVarP(&lsOpts.Quiet, "quiet", "q", false, "Only display IDs")
+	lsCmd.Flags().BoolVarP(&lsOpts.Quiet, "quiet", "q", false, "Only display names")
```
ndeloof, line 53: A compose application identified is actually project name. Using "name" in the help is still ambiguous imho. Maybe "project names" would be better ?

PR #12542, opened 2025-02-10
golangci-lint: enable copyloopvar linter

thaJeztah:
capturing loop variables is no longer needed in go1.22 and higher; https://go.dev/blog/loopvar-preview

This path enables the copyloopvar linter, which finds places where capturing is no longer needed, and removes locations where they could be removed.

Also made some minor changes, and renamed some vars in places where we could use a shorter name that's less likely to conflict with imports.

**What I did**

**Related issue**

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -5,6 +5,7 @@ linters:
   enable-all: false
   disable-all: true
   enable:
+    - copyloopvar
     - depguard
     - errcheck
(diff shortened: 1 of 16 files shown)


ndeloof (approved):

thaJeztah:
Missed one that was deliberately copied; should be fixed now 😅

PR #12545, opened 2025-02-10
go.mod: update to docker v28.0.0

thaJeztah:
### go.mod: docker/docker, docker/cli v28.0.0, buildx v0.21.0

full diff:

- https://github.com/docker/docker/compare/v27.5.1...v28.0.0
- https://github.com/docker/cli/compare/v27.5.1...v28.0.0
- https://github.com/docker/buildx/compare/v0.20.1...v0.21.1

**What I did**

**Related issue**


go.mod
```diff
@@ -1,9 +1,12 @@
 module github.com/docker/compose/v2
 
-go 1.22.10
+go 1.23.0
```
thaJeztah, line 3: Looks like the minimum version was bumped in this PR; I need to check if that was intentional

- https://github.com/docker/buildx/pull/2965

pkg/compose/create.go
```diff
@@ -274,7 +274,7 @@ func (s *composeService) getCreateConfigs(ctx context.Context,
 		WorkingDir:      service.WorkingDir,
 		Entrypoint:      entrypoint,
 		NetworkDisabled: service.NetworkMode == "disabled",
-		MacAddress:      macAddress,
+		MacAddress:      macAddress, // FIXME(thaJeztah) this should be moved to networksettings -> EndpointSettings.MacAddress
```
thaJeztah, line 277: Note to self; looks like we need to update this to use the new field

ndeloof:
I can't push to your fork so created https://github.com/docker/compose/pull/12564

thaJeztah:
Updated to v28.0.0 and included your fix from https://github.com/docker/compose/pull/12564/commits/152490f6fd150d8dac8a7f6246f622ed593c2b59 @ndeloof

thaJeztah:
> I can't push to your fork

Also not sure why that didn't work 🤔 something broken on GitHub?

Screenshot 2025-02-20 at 11 45 08

pkg/compose/create.go
thaJeztah: @robmry @akerouanton could one of you help doing a follow-up to adjust the code in compose for the new approach? (I wasn't 100% sure if "only new" should be set, or "both old and new to same value")

glours (approved):
Sounds good to me, thanks @thaJeztah and @ndeloof 👍

pkg/compose/create.go
thaJeztah: Thanks! I wasn't sure what would happen if we did _NOT_ set the old one, if compose would connect to an older version of the engine, but perhaps we take care of that in the client

robmry: Yes, I guess we'll have to ... the change was in API 1.44.

ndeloof: this doesn't need a FIXME, we just support older API and migrate to new API when relevant (see `prepareContainerMACAddress`)

thaJeztah: Ah! I missed that one; in that case, I may add a `//nolint:` and a comment to prevent confusion; that sound good?

thaJeztah: Added; let me know if that looks good (thanks!)

thaJeztah:
Derp; I must've had a different version of the linter running locally that did complain :joy:

```
#22 114.9 pkg/compose/create.go:277:32: directive `//nolint:staticcheck // Deprecated since API v1.44, but kept for compatibility with older API versions.` is unused for linter "staticcheck" (nolintlint)
```

thaJeztah:
Rebased to get a fresh run of CI

thaJeztah:
oh! buildx 0.21.1 is there; I can update :joy: https://github.com/docker/docker-ce-packaging/pull/1167

thaJeztah:
> oh! buildx 0.21.1 is there; I can update 😂 [docker/docker-ce-packaging#1167](https://github.com/docker/docker-ce-packaging/pull/1167)

Done!


https://github.com/docker/buildx/compare/v0.21.0...v0.21.1

ndeloof (approved):

Issue #12544, opened 2025-02-10
Watch: Support Path Pattern Matching


davidgfitzgerald:
# Request

It would be nice if the [watch](https://docs.docker.com/reference/compose-file/develop/#watch) `path` attribute behaved like the `ignore` attribute. For example, `path` could accept wildcards such as `*`.

At the very least it would help to have some more detail in the API documentation adding that (at least for the time being) pattern matching does not work on `path`.

Reference: https://docs.docker.com/reference/compose-file/develop/

# Example

I would like the convenience of being able to do this:

```yaml
services:
  frontend:
    build:
      context: ./frontend
      dockerfile: Dockerfile
    ports:
      - "5173:5173"
    develop:
        watch:
        - action: rebuild
          path: frontend/*.json
        - action: rebuild
          path: frontend/*.js
```

instead of the verbose (yet explicit):

```yaml
services:
  frontend:
    build:
      context: ./frontend
      dockerfile: Dockerfile
    ports:
      - "5173:5173"
    develop:
        watch:
        - action: rebuild
          path: frontend/jsconfig.json
        - action: rebuild
          path: frontend/package-lock.json
        - action: rebuild
          path: frontend/package.json
        - action: rebuild
          path: frontend/eslint.config.js
        - action: rebuild
          path: frontend/svelte.config.js
        - action: rebuild
          path: frontend/tailwind.config.js
        - action: rebuild
          path: frontend/vite.config.js
```

## Nice To Have

Bonus points for allowing multiple `path` entries in a single `rebuild` action:

```yaml
services:
  frontend:
    build:
      context: ./frontend
      dockerfile: Dockerfile
    ports:
      - "5173:5173"
    develop:
        watch:
        - action: rebuild
          path: frontend/*.json frontend/*.js
```

Currently, passing a space separated list is silently interpreted as a single path. So, the below example (misleadingly) does not match `frontend/package.json` nor `frontend/vite.config.js`

```yaml
        - action: rebuild
          path: frontend/package.json frontend/vite.config.js
```

jhrotko:
I created a POC in these branches:
- [compose](https://github.com/jhrotko/compose/tree/feature-wild-star-watch-path)
- [compose-go](https://github.com/jhrotko/compose-go/tree/feature-wild-star-watch-path)

in these branches the implementation is under the `path` which can now receive a string or a list of strings and also validates the `*` pattern matching. 

[...]

PR #12557, opened 2025-02-17
feat: Handle glob pattern in watch configuration path

jhrotko:
**What I did**
This PR enables watch configuration to handle glob patterns in watch configuration `path` keyword.

Example:

```yaml
    develop:
      watch:
        - path: api/*.txt
          action: rebuild
```
In order to watch files under some glob pattern, during watch initialisation we register the **anchor** folder for the glob pattern (ex.: `api`). When a file event is received, is going to check if any of the watch rules matches the pattern. In this case, an additional rule pattern was added to `watchRule` -- [`globPattern`](https://github.com/jhrotko/compose/blob/4396577c6d05a0378e0766c9c8eaf9aeb9d74fa6/pkg/compose/watch.go#L252-L267) -- in order to fulfil this verification. 

Note: Some of these changes require changes made in https://github.com/compose-spec/compose-go/pull/740. For more details please check this PR.

**Related issue**

This PoC is based on the request
- #12544 

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -200,4 +200,4 @@ require (
 	sigs.k8s.io/yaml v1.4.0 // indirect
 )
 
-replace github.com/compose-spec/compose-go/v2 v2.4.8 => ../compose-go
+replace github.com/compose-spec/compose-go/v2 v2.4.8 => github.com/jhrotko/compose-go/v2 v2.0.0-20250217114813-b0bb892d9de8
--- a/go.mod
+++ b/go.mod
@@ -200,4 +200,4 @@ require (
 	sigs.k8s.io/yaml v1.4.0 // indirect
 )
 
-replace github.com/compose-spec/compose-go/v2 v2.4.8 => github.com/jhrotko/compose-go/v2 v2.0.0-20250217114813-b0bb892d9de8
+replace github.com/compose-spec/compose-go/v2 v2.4.8 => github.com/jhrotko/compose-go/v2 v2.0.0-20250225153415-9ce61ad83a27
--- a/go.mod
+++ b/go.mod
@@ -168,6 +168,7 @@ require (
 	github.com/xeipuuv/gojsonpointer v0.0.0-20190905194746-02993c407bfb // indirect
 	github.com/xeipuuv/gojsonreference v0.0.0-20180127040603-bd5ef7bd5415 // indirect
 	github.com/xeipuuv/gojsonschema v1.2.0 // indirect
+	github.com/xhit/go-str2duration/v2 v2.1.0 // indirect
 	github.com/zclconf/go-cty v1.16.0 // indirect
 	go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc v0.56.0 // indirect
 	go.opentelemetry.io/contrib/instrumentation/net/http/httptrace/otelhttptrace v0.56.0 // indirect
@@ -198,3 +199,5 @@ require (
 	sigs.k8s.io/structured-merge-diff/v4 v4.4.1 // indirect
 	sigs.k8s.io/yaml v1.4.0 // indirect
 )
+
+replace github.com/compose-spec/compose-go/v2 v2.4.8 => ../compose-go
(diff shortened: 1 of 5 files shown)


go.mod
```diff
@@ -198,3 +198,5 @@ require (
 	sigs.k8s.io/structured-merge-diff/v4 v4.4.1 // indirect
 	sigs.k8s.io/yaml v1.4.0 // indirect
 )
+
+replace github.com/compose-spec/compose-go/v2 v2.4.8 => ../compose-go
```
glours, line 202: Hey @jhrotko 👋 
You need to replace `compose-go` with a version of your fork 😉

jhrotko: Coucou @glours ! 😉
Done

ndeloof:
LGTM

just as a side not, support for multiple paths in watch definition could also be implemented as some syntaxic sugar in compose-go, i.e. while parsing we translate multiple `path` values into a set of watch rules, with same action definition, for each declared paths. Somehow comparable to the way we translate port ranges into a set of simpler port declarations.

This would avoid changes to compose and keep logic simpler (especially to later introduce support for match patterns). wdyt?

jhrotko:
> LGTM
> 
> just as a side not, support for multiple paths in watch definition could also be implemented as some syntaxic sugar in compose-go, i.e. while parsing we translate multiple `path` values into a set of watch rules, with same action definition, for each declared paths. Somehow comparable to the way we translate port ranges into a set of simpler port declarations.
> 
> This would avoid changes to compose and keep logic simpler (especially to later introduce support for match patterns). wdyt?

Hey @ndeloof  :) Could you confirm that when you say sugar syntax do you mean to represent the `path` like this? (keeping the `string` type in compose-spec, etc?)

```
...
   develop:
      watch:
        - path: api;other;*.txt;web/some/really/super/long/path; ### assuming ; as a delimiter (por example) for each path
          action: rebuild
```

ndeloof:
Nope, I mean `path` can be a _string or list_ in yaml as you suggested, but compose-go while loading would actually convert this into a single `path` string per watch rule, so that you don't need this PR.

Actually I see more benefits in pattern support vs multiple paths, which sounds a corder case to me

jhrotko:
> Nope, I mean `path` can be a _string or list_ in yaml as you suggested, but compose-go while loading would actually convert this into a single `path` string per watch rule, so that you don't need this PR.
> 
> i.e.
> 
> ```yaml
>    develop:
>       watch:
>         - path: [api, other]
>            action: rebuild
> ```
> 
> would be cannonical-ized into:
> 
> ```yaml
>    develop:
>       watch:
>         - path: api
>            action: rebuild
>         - path: other
>            action: rebuild
> ```
> 
> Actually I see more benefits in pattern support vs multiple paths, which sounds a corder case to me

I see, I will make this changes. The pattern matching can translate into multiple files though so these features are actually related. And can be annoying to define multiple watch rules for the same action (rebuild and restart)

ndeloof:
> The pattern matching can translate into multiple files though so these features are actually related

I don't get this. A basic path matches multiple files (anything under folder matches). A pattern is just more restrictive. This is unrelated to multiple values in `path` (which we can support, but this is another feature then)

My initial thoughts to implement this feature was to introduce an `includes` attribute to define pattern filtering for files under `path`, but it makes sense for a simpler approach to make `path` a pattern. The fixed part of the pattern would then declare the source path, and the match expression the filter
For illustration:
```
path: src/*.txt # watch src/ but only consider .txt files
target: /app
```
File `src/foo.txt` would be copied into container as `app/foo.txt`
This would somehow be equivalent to a negative `ignore`:
```
path: src/
ignore: !*.txt # AFAIK this isn't supported
target: /app
```
.. but with a more comprehensible syntax :P

jhrotko:
so imagine the case where the target is a file

```
  path: app/*.txt
  target: app/some.txt
  action: sync
```

in this case, we need to be careful and through some error. During validation, when target is a file, compose-go needs to assert the relation 1 to 1 is true or else it will have a weird behaviour in this case. That is why I was assuming path could be treated as a []string to be easier to validate this case.

The presented example I would assume it should throw an error or show a warning at least as multiple files can overwrite one file in the container

ndeloof:
target can't be a file, MUST be a folder. If you want to sync with file `app/some.txt` inside container, source MUST be `some.txt`. What you describe is way more complex use-case but just support for pattern

jhrotko:
@ndeloof you should make it explicit in the [documentation page](https://docs.docker.com/compose/how-tos/file-watch/#path-and-target). The given example for `target` is actually a file 😓

PR #12559, opened 2025-02-17
Link to configuration file docs

### LINKED ISSUE #12526: Compose file reference seems very hidden in help file and not to be found from starting page
### Description

Perhaps I'm just too tired to navigate the docs pages, but it occurred to me perhaps beginners might be confused too:

When reading the compose starting page, you see a "compose configuration file" mentioned here. It is **not clickable.** This means the natural place to find out what that is, is probably "compose config" to the left:

Image

However, this seems to talk about something entirely different, but then mentions the "compose files" again but now by a **different name ("compose file" rather than "compose config file") that still isn't clickable.** That name also doesn't have any equivalent in the alphabetic listing to the left, where there is no F letter to be found.

Image

I happen to know what a compose file is and have used them for a while and was merely searching the reference without using a search engine, but I failed. Perhaps it's just me though, but I figured if it isn't, this could use some more links and more unified naming.

andrew-kramer:
**What I did**
Added a link from the [compose command docs](https://docs.docker.com/reference/cli/docker/compose/#examples) to the [config file docs](https://docs.docker.com/reference/compose-file/) to improve clarity and ease of finding related docs.

**Related issue**
Closes #12526 

A lynx:
220px-Lynx_lynx_poing-1084623192

### DIFF
--- a/docs/reference/compose.md
+++ b/docs/reference/compose.md
@@ -67,7 +67,7 @@ Define and run multi-container applications with Docker
 ## Examples
 
 ### Use `-f` to specify the name and path of one or more Compose files
-Use the `-f` flag to specify the location of a Compose configuration file.
+Use the `-f` flag to specify the location of a Compose [configuration file](/reference/compose-file/).
 
 #### Specifying multiple Compose files
 You can supply multiple `-f` configuration files. When you supply multiple files, Compose combines them into a single
--- a/docs/reference/docker_compose.yaml
+++ b/docs/reference/docker_compose.yaml
@@ -229,7 +229,7 @@ options:
       swarm: false
 examples: |-
     ### Use `-f` to specify the name and path of one or more Compose files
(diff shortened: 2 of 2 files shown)


ndeloof:
Please run `make docs`

ndeloof (approved):

glours:
Thanks for the PR, @aevesdocker Compose Technical writer is looking to make a global PR to address all such potential inconsistency in the command flags.

andrew-kramer:
> Please apply such change to the flag definition in go code (cmd/compose/compose.go) then run `make docs`

I'm happy to do that, but the Go code already has all the text. This PR was just to add a link, which I think the Go code wouldn't include.

Sorry, I might well be misunderstanding the request or missing something in the code. This is my first PR here, which is why I chose something minuscule.

ndeloof (approved):

PR #12584, opened 2025-02-25
Introduced `include` to filter files considered by `watch`

ndeloof:
**What I did**

Introduce `include` to only select a subset of files in source path for watch command. 
Illustration example:
```yaml
services:
  base:
    build:
      context: .
      dockerfile_inline: |
        FROM nginx
    develop:
      watch:
        - path: .
          include: "*.txt" # can also be an array
          action: rebuild
```

**Related issue**
closes https://github.com/docker/compose/issues/12544
see https://github.com/compose-spec/compose-go/pull/747

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,7 +7,7 @@ require (
 	github.com/Microsoft/go-winio v0.6.2
 	github.com/acarl005/stripansi v0.0.0-20180116102854-5a71ef0e047d
 	github.com/buger/goterm v1.0.4
-	github.com/compose-spec/compose-go/v2 v2.4.9-0.20250225151507-331db8fefcb7
+	github.com/compose-spec/compose-go/v2 v2.4.9-0.20250302154753-e508c724a35f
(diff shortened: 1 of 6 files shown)


ndeloof:
cc @jhrotko

jhrotko:
@ndeloof it can be a pattern, check my PR. This can be a more restrictive, can you handle in this PR glob pattern for folders? `app/*/sub`?

ndeloof:
@jhrotko no it can't - this is intentional at this stage, as we need a base path to start collecting filesystem events, I'm investigating if we could get rid of this for better UX

About pattern for folder:
```yaml
    develop:
      watch:
        - path: app
          include: "*/sub"
          action: rebuild
```
(see https://docs.docker.com/build/concepts/context/#syntax about actual pattern syntax)

ndeloof:
@glours added a dedicated e2e test

glours (approved):
LGTM

PR #12595, opened 2025-03-03
display the location of OCI or GIT Compose stack download

glours:
**What I did**
Display a message with the location where the OCI artifact or GIT Compose stack were downloaded

**Related issue**
https://docker.atlassian.net/browse/APCLI-881

image

### DIFF
--- a/cmd/compose/up.go
+++ b/cmd/compose/up.go
@@ -301,6 +301,7 @@ func runUp(
 		attachSet.RemoveAll(upOptions.noAttach...)
 		attach = attachSet.Elements()
 	}
+	displayLocationRemoteStack(dockerCli, project, buildOptions)
 
 	timeout := time.Duration(upOptions.waitTimeout) * time.Second
 	return backend.Up(ctx, project, api.UpOptions{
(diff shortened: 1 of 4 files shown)


pkg/remote/git.go
```diff
@@ -116,6 +121,9 @@ func (g gitRemoteLoader) Load(ctx context.Context, path string) (string, error)
 	if stat.IsDir() {
 		local, err = findFile(cli.DefaultFileNames, local)
 	}
+	if progress.Mode != progress.ModeQuiet && progress.Mode != progress.ModeJSON {
+		_, _ = fmt.Fprintf(g.dockerCli.Out(), "Your compose stack %q is stored in %q\n", strings.TrimPrefix(path, prefix), local)
```
ndeloof, line 125: this message will print for every Git remote loaded by compose, including those involved by `includes`
Maybe better load project, then check if `project.ComposeFiles[0]` is remote _then_ print result of `Dir(".")` as location where remote stack has been stored on disk

cmd/compose/up.go
```diff
@@ -329,3 +331,13 @@ func setServiceScale(project *types.Project, name string, replicas int) error {
 	project.Services[name] = service
 	return nil
 }
+
+func displayLocationRemoteStack(dockerCli command.Cli, project *types.Project, options buildOptions) {
+	mainComposeFile := options.ProjectOptions.ConfigPaths[0]
+	if ui.Mode != ui.ModeQuiet && ui.Mode != ui.ModeJSON {
+		if strings.HasPrefix(mainComposeFile, remote.OciPrefix) || strings.HasPrefix(mainComposeFile, "git://") {
```
ndeloof, line 338: Better use `remote.Accept` as Git support has a bunch of corner cases

ndeloof (approved):

PR #12601, opened 2025-03-04
`publish`reject compose file with bind mounts

ndeloof:
**What I did**
`publish` reject compose file with bind mounts
Compose file author should rely on volumes to prevent consumer to expose personal data to a potentially untrusted compose stack

**Related issue**
https://docker.atlassian.net/browse/APCLI-878

### DIFF
--- a/pkg/compose/publish.go
+++ b/pkg/compose/publish.go
@@ -155,6 +155,15 @@ func (s *composeService) preChecks(project *types.Project, options api.PublishOp
 		}
 		return acceptPublishEnvVariables(s.dockerCli)
 	}
+
+	for name, config := range project.Services {
+		for _, volume := range config.Volumes {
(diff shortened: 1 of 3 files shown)


pkg/compose/publish.go
```diff
@@ -155,6 +155,15 @@ func (s *composeService) preChecks(project *types.Project, options api.PublishOp
 		}
 		return acceptPublishEnvVariables(s.dockerCli)
 	}
+
+	for _, config := range project.Services {
+		for _, volume := range config.Volumes {
+			if volume.Type == types.VolumeTypeBind {
+				return false, errors.New("cannot publish a compose file relying on bind-mount. You should use volumes")
```
glours, line 162: Maybe we could let the user know which service has a bind mount declared, no?

glours (approved):

PR #12606, opened 2025-03-06
Publish compose file with required siblings used by `extends`

ndeloof:
**What I did**

Include sibling compose files used by `extends` when publishing a compose application as OCI artifact.
These sibling files are named based on hash so that we don't have to manage complex scenarios with relative paths.
Annotation is added to OCI layer to distinguish extends files vs the main compose files to define the compose application.

**Related issue**
https://docker.atlassian.net/browse/APCLI-1016


pkg/compose/publish.go
```diff
@@ -116,6 +151,94 @@ func (s *composeService) publish(ctx context.Context, project *types.Project, re
...
+		hash := fmt.Sprintf("%x.yaml", sha256.Sum256([]byte(xf)))
+		extFiles[xf] = hash
+
+		// FIXME implement without relying on yqlib so we don't get extra dependencies
+		exp := fmt.Sprintf(".services.%s.extends.file = %q", name, hash)
+		encoder := yqlib.NewYamlEncoder(yqlib.YamlPreferences{
+			ColorsEnabled: false,
+		})
+		decoder := yqlib.NewYamlDecoder(yqlib.YamlPreferences{})
+		out, err := yqlib.NewStringEvaluator().Evaluate(exp, string(f), encoder, decoder)
```
ndeloof, line 232: TODO: I'm investigating if we could avoid ylib as a dependency and "just" reproduce this single search-and-replace feature

glours: Do you want to make the changes in this PR or in a follow up one?

glours (approved):
Sounds good to me 👍

pkg/compose/publish.go
ndeloof: I'll look into this as a follow-up

PR #12611, opened 2025-03-06
only load env_file after services have been selected

ndeloof:
**What I did**
resolve `env_file` _after_  active services have been selected, so we don't read (potentially missing) env_file for inactive ones

**Related issue**
fixes https://github.com/docker/compose/issues/8713
requires https://github.com/compose-spec/compose-go/pull/751

### DIFF
--- a/cmd/compose/compose.go
+++ b/cmd/compose/compose.go
@@ -170,7 +170,7 @@ func (o *ProjectOptions) WithServices(dockerCli command.Cli, fn ProjectServicesF
 	return Adapt(func(ctx context.Context, args []string) error {
 		options := []cli.ProjectOptionsFn{
 			cli.WithResolvedPaths(true),
-			cli.WithDiscardEnvFile,
+			cli.WithoutEnvironmentResolution,
 		}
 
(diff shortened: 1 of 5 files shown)


glours (approved):
LGTM 👍

masalim2:
@ndeloof is it possible this change would cause `env_file` not to be read when a `docker compose run` command is invoked on a service without specifying its profile explicitly?

Concretely:

```yaml
service-foo:
  profiles: ["foo"]
  env_file: settings.env
```

```bash
# stuff from settings.env is missing!
$ docker compose run --entrypoint /bin/bash service-foo -c 'printenv'
```

Apologies if a closed PR is not a good thread for this type of question.  I've noticed a regression in my CI suite in moving from Compose 2.33.1 to 2.34.0, and I'm realizing it may be that I was depending on the behavior prior to this change.  It doesn't feel exactly like a bug and I wasn't able to put together a quick reproducer.

ndeloof:
@masalim2 see https://github.com/docker/compose/issues/12644

masalim2:
Ah I was only looking at open issues and missed this.  Thank you!!

PR #12615, opened 2025-03-07
implement extends.file replace without yqlib

ndeloof:
**What I did**
implemented replacement of `services.xx.extends.file` without relying on yqlib

**Related issue**


glours (approved):
Noice 👍

PR #12620, opened 2025-03-10
use Defang secret-detector to identify potential secret leaks before publishing OCI artifacts

glours:
**What I did**
Use Defang Labs `secret-detector` library to detect potential secret leaks before publishing OCI artifacts and ask user if they really want to publish their Compose stacks with those data

**Related issue**
https://docker.atlassian.net/browse/APCLI-876

image


glours:
I had to load compose files without interpolation and env resolution, then marshall them back to yaml because `secret-detector` doesn't handle properly `environment` attribute as a list and don't detect potential leaks

pkg/compose/publish.go
```diff
@@ -334,3 +360,99 @@ func (s *composeService) checkOnlyBuildSection(project *types.Project) (bool, er
...
+			findings, err := scan.ScanFile(envFile.Path)
+			if err != nil {
+				return nil, fmt.Errorf("failed to scan env file %s: %w", envFile.Path, err)
+			}
+			allFindings = append(allFindings, findings...)
+		}
+	}
+
+	// Check configs defined by files
+	for _, config := range project.Configs {
```
ndeloof, line 403: AFAICT we do **not** publish the configs and secrets files (yet) as part of the `publish` command

ndeloof (approved):

Issue #12622, opened 2025-03-10
[BUG] Unknown flag: `--yes`


jonasgeiler:
### Description

Not sure if this is a bug or a request for change, but the long form of the `-y` flag on the `docker compose up`, `docker compose create` and `docker compose publish` command is currently just `--y`, which seems wrong to me and also created some confusion, as seen here: ansible-collections/community.docker#1054

Shouldn't the long form of this flag be `--yes`, as is usual? This `--yes` long form was even used in the description of the original pull request, #12363, which introduced this flag. But then, in the code, `"y"` was used instead. This is why I assume this was an oversight/typo from @ndeloof.

### Steps To Reproduce

1. Run `docker compose up --yes`
2. An error is printed that this flag is unknown. Of course it is correctly listed as `--y` in the help message, but I just assume this flag was named incorrectly on accident.

### Compose Version

```Text
Docker Compose version 2.33.1
```

### Docker Environment

```Text
Client:
 Version:    28.0.1
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  0.21.2
    Path:     /usr/lib/docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  2.33.1
    Path:     /usr/lib/docker/cli-plugins/docker-compose

Server:
 Containers: 0
  Running: 0
  Paused: 0
  Stopped: 0
 Images: 5
 Server Version: 27.3.1
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: local
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: c507a0257ea6462fbd6f5ba4f5c74facb04021f4.m
 runc version:
 init version: de40ad0
 Security Options:
  seccomp
   Profile: builtin
  cgroupns
 Kernel Version: 6.12.10-arch1-1
 Operating System: EndeavourOS
 OSType: linux
 Architecture: x86_64
 CPUs: 4
 Total Memory: 15.51GiB
 Name: redacted
[...]

PR #12623, opened 2025-03-11
deprecate --y, prefer --yes

ndeloof:
**What I did**

**Related issue**
fixes https://github.com/docker/compose/issues/12622


glours (approved):
LGTM

PR #12626, opened 2025-03-11
Introduce --env-from-file in `docker compose run` command

ndeloof:
**What I did**
Allow to set `run` environment from env file

**Related issue**
fix https://github.com/docker/compose/issues/8379
based on https://github.com/docker/compose/pull/9169


cmd/compose/run.go
```diff
@@ -65,6 +67,7 @@ type runOptions struct {
 	ignoreOrphans bool
 	removeOrphans bool
 	quietPull     bool
+	EnvFiles      []string
```
glours, line 70: doest it need to be access somewhere else?
```suggestion
	envFiles      []string
```

glours (approved):
Sounds good to me

glours:
@ndeloof you forgot to regenerate the doc 😉

PR #12629, opened 2025-03-12
make publish a regular command of Compose

glours:
**What I did**
Move `publish` away from the `alpha` command umbrella

**Related issue**
N/A

image

### DIFF
--- a/cmd/compose/publish.go
+++ b/cmd/compose/publish.go
@@ -20,12 +20,12 @@ import (
 	"context"
 	"errors"
 
+	"github.com/docker/cli/cli"
 	"github.com/docker/cli/cli/command"
+	"github.com/docker/compose/v2/pkg/api"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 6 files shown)


cmd/compose/publish.go
```diff
@@ -46,7 +46,7 @@ func publishCommand(p *ProjectOptions, dockerCli command.Cli, backend api.Servic
 		RunE: Adapt(func(ctx context.Context, args []string) error {
 			return runPublish(ctx, dockerCli, backend, opts, args[0])
 		}),
-		Args: cobra.ExactArgs(1),
+		Args: cli.RequiresMinArgs(1),
```
ndeloof, line 49: why this change ? Don't we only support a single (required) arg for this command ?

ndeloof (approved):

cmd/compose/publish.go
ndeloof, line 49:
```suggestion
		Args: cli.ExactArgs(1),
```

PR #12415, opened 2024-12-19
Allow `up --watch --prune`

remcokranenburg:
**What I did**
Just like #11932, which added `watch --prune`, but now for the up command.

**Related issue**
#11073

**(not mandatory) A picture of a cute animal, if possible in relation to what you did**
image

PR #11932, opened 2024-06-24
feat(watch): Add --prune option to docker-compose watch command

idsulik:
**What I did**
Added the `--prune` option to the `docker-compose watch` command. This option ensures that dangling images are pruned automatically when rebuilding, helping to keep the Docker environment clean and efficient.

**Related issue**
https://github.com/docker/compose/issues/11073

image

Issue #11073, opened 2023-10-08
Idea: docker compose watch --prune


arjasco:
### Description

Having just started using `watch` to rebuild a Go application on code changes, I'm quite happy! It's a solid feature that means I no longer need to rely on other tools to watch for file changes.

I've done some searching around but haven't found any discussion on tidying up dangling images after the numerous if not hundreds of rebuilds that could occur throughout the day. 

Adding a `--prune` option to the command that does this automatically would be a nice next addition. 

At present you only have the following options (as far as I'm aware):

- Run `docker system prune` manually once in a while. 
- Set up a cronjob to run the above command to remove the manual element.
- Or a little more programatic would be to pipe the output of `docker system events` to another process, pulling out the label for `com.docker.compose.service` and providing that as a filter to the prune command. 

All are undesirable, you might not want this running all the time on a schedule and anything manual is going against the point of having a watcher in the first instance to build / sync.

nivpenso:
I have the exact same issue.

I thought of solving it by adding an additional field `keep_dangled_image` to the configuration of the `rebuild` action section.

by default `keep_dangled_image` will be `true`, and everything will act as today. but if it is `false`, the compose will delete the image that became dangled after rebuilding the image

```
services:
  nginx:
    build:
      context: .
    develop:
      watch:
        - action: rebuild
          path: ./deps/nginx/nginx.conf
          keep_dangled_image: false
```

g0t4:
This could work for any compose project with a build.

```
services:
  nginx:
    build:
      context: .
      prune: true
```

idsulik:
hi everyone! It was [implemented](https://github.com/docker/compose/pull/11932) and [released](https://github.com/docker/compose/releases/tag/v2.29.0)

PR #12642, opened 2025-03-17
Add `--prune` for up command

remcokranenburg:
*Note: this PR is reopened unchanged from [one](https://github.com/docker/compose/pull/12415) that was closed due to inactivity.*

**What I did**
Just like #11932, which added `watch --prune`, but now for the up command.

**Related issue**
#11073

image

### DIFF
--- a/cmd/compose/up.go
+++ b/cmd/compose/up.go
@@ -57,6 +57,7 @@ type upOptions struct {
 	wait                  bool
 	waitTimeout           int
 	watch                 bool
+	prune                 bool
 	navigationMenu        bool
 	navigationMenuChanged bool
 }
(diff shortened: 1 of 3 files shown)


cmd/compose/up.go
```diff
@@ -310,6 +315,7 @@ func runUp(
 			Wait:           upOptions.wait,
 			WaitTimeout:    timeout,
 			Watch:          upOptions.watch,
+			Prune:          upOptions.prune,
```
glours, line 318: I wonder if we should not pass directly [`WatchOptions`](https://github.com/docker/compose/blob/f70209cf15efbcc009531879cc4cf9c9082cfb81/pkg/api/api.go#L128)

glours (commented):
@ndeloof what do you think, shouldn't we use `WatchOptions` directly here? I don't have a strong option

ndeloof (approved):

cmd/compose/up.go
```diff
@@ -170,6 +171,7 @@ func upCommand(p *ProjectOptions, dockerCli command.Cli, backend api.Service) *c
 	flags.BoolVar(&up.wait, "wait", false, "Wait for services to be running|healthy. Implies detached mode.")
 	flags.IntVar(&up.waitTimeout, "wait-timeout", 0, "Maximum duration in seconds to wait for the project to be running|healthy")
 	flags.BoolVarP(&up.watch, "watch", "w", false, "Watch source code and rebuild/refresh containers when files are updated.")
+	flags.BoolVar(&up.prune, "prune", false, "Prune dangling images on rebuild")
```
ndeloof, line 174: I wonder: is there any benefits _not_ pruning those ? Shall we just make this the default ? Then do we need a flag ?

glours: Yep, I agree
@remcokranenburg do you mind adapting your PR to prune by default? We just need to keep it for the `watch` command itself to let users who want it to choose the option of keeping dangling images

ndeloof: up already has so much flags we'd prefer not to have more. For those who really want to keep dangling images (?) `watch --prune=false` could be a reasonable workaround

glours: Yes I'm aligned with @ndeloof on this one, just keep the `--prune` flag for the `watch` command
BTW thank you for doing it 🙏

ndeloof:
you have to run `make docs` as a new flag has been introduced

PR #12643, opened 2025-03-17
PWD

ndeloof:
**What I did**
set `PWD` variable as all Unix shell will offer this variable _but_ windows powershell.
Using `${PWD}` is such a common practice for compose users we can't pretend this isn't supported :P 

**Related issue**


glours (approved):
LGTM

Issue #12477, opened 2025-01-17
[BUG] the restart command ignores the depends_on constraint


yangjinheng:
### Description

the restart command ignores the depends_on constraint

Is this normal?

### Compose Version

```Text
Docker Compose version v2.32.4
```

### Docker Environment

```Text
Client:
 Version:           27.4.1
 API version:       1.47
 Go version:        go1.22.10
 Git commit:        b9d17ea
 Built:             Tue Dec 17 15:44:45 2024
 OS/Arch:           linux/amd64
 Context:           default

Server: Docker Engine - Community
 Engine:
  Version:          27.4.1
  API version:      1.47 (minimum version 1.24)
  Go version:       go1.22.10
  Git commit:       c710b88
  Built:            Tue Dec 17 15:46:06 2024
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          v1.7.24
  GitCommit:        88bf19b2105c8b17560993bee28a01ddc2f97182
 runc:
  Version:          1.2.3
  GitCommit:        v1.2.3-0-g0d37cfd
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

idsulik:
Please, fill the **Steps To Reproduce** section

ndeloof:
`depends_on` is only applied as container is created, until you explicitly declare a dependent service to require being restarted after it's dependency was updated/restarted, see https://docs.docker.com/reference/compose-file/services/#long-syntax-1

yangjinheng:
> `depends_on` is only applied as container is created, until you explicitly declare a dependent service to require being restarted after it's dependency was updated/restarted, see https://docs.docker.com/reference/compose-file/services/#long-syntax-1

I read this document, but it doesn't solve my problem. restart = true is not what I want.

I need the dependent service to be restarted after my service

Ensure data is saved correctly

yangjinheng:
> 请填写**重现步骤**部分

all containers restart at the same time, and those with dependencies will fail

```
root@palyground:~/deploy# docker compose restart
[+] Restarting 7/7
 ✔ Container prod-mysql-1         Restarting    0.5s
 ✔ Container prod-prometheus-1    Restarting    0.5s
 ✔ Container prod-quickrest-1     Restarting    0.5s
 ✔ Container prod-nginx-1         Restarting    0.5s
 ✔ Container prod-quickproxy-1    Restarting    0.5s
[...]

Issue #10308, opened 2023-02-23
[BUG] Compose v2 is recreating ALL containers where there is a change to one of the "depends_on" in the hierarchy (bring back compose v1 behaviour)


captainfalcon23:
### Description

Separating out an issue described on https://github.com/docker/compose/issues/9600

Basically, this comment (https://github.com/docker/compose/issues/9600#issuecomment-1403124191) encompasses this issue. Sometimes containers are completely recreated when we do not expect them to, and when using "depends_on", and a dependant container changes in some way, all of the children are restarted. In compose v1.X, this is NOT the case, and ONLY the changed container is recreated and all others are untouched.

Either this is a bug, or a feature request to bring back compose v1 behaviour (at least with a flag).

### Steps To Reproduce

Refer https://github.com/docker/compose/issues/9600#issuecomment-1403124191

### Compose Version

```Text
Docker Compose version v2.16.0
```

### Docker Environment

```Text
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  app: Docker App (Docker Inc., v0.9.1-beta3)
  buildx: Docker Buildx (Docker Inc., v0.9.1-docker)
  compose: Docker Compose (Docker Inc., v2.16.0)
  scan: Docker Scan (Docker Inc., v0.21.0)

Server:
 Containers: 4
  Running: 4
  Paused: 0
  Stopped: 0
 Images: 6
 Server Version: 20.10.21
 Storage Driver: overlay2
  Backing Filesystem: xfs
  Supports d_type: true
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 1
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 io.containerd.runtime.v1.linux runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 78f51771157abb6c9ed224c22013cdf09962315d
 runc version: v1.1.4-0-g5fd4c4d
 init version: de40ad0
 Security Options:
  seccomp
   Profile: default
 Kernel Version: 3.10.0-1160.80.1.el7.x86_64
 Operating System: CentOS Linux 7 (Core)
 OSType: linux
 Architecture: x86_64
 CPUs: 8
 Total Memory: 15.06GiB
[...]

PR #12649, opened 2025-03-19
Fixed support for `depends_on.restart` in `up` and `restart` commands 

ndeloof:
**What I did**

- `up` when it detects diverged container will only restart dependent services with `restart: true` set, and keep the others running
- `restart` already was running in dependency order, but was not waiting for healthy state

**Related issue**
closes https://github.com/docker/compose/issues/10308
closes https://github.com/docker/compose/issues/12477


glours (approved):
👍

PR #12650, opened 2025-03-19
Set watch option --prune=true as default

remcokranenburg:
**What I did**
Prune images by default when watching. This also has the effect of enabling pruning when using `up --watch`. It's still possible to disable pruning with `watch --prune=false`.

Replaces [previous approach](https://github.com/docker/compose/pull/12642).

**Related issue**
#11073

image


remcokranenburg:
Sorry for the force-push, I forgot to add the sign-off line.

ndeloof (approved):

glours (approved):

Issue #12659, opened 2025-03-22
[BUG] env not being applied to specific service profile


SayidHosseini:
### Description

docker compose run <PROFILE_NAME> does not use .env.

### Steps To Reproduce

Here is a sample `docker-compose.yml`:
```bash
services:
  migration:
    image: nginx
    env_file: [.env]
    profiles: [migration]
    command: env
```

And a sample .env:
```bash
TEST=True
```

Here is the output of `docker compose config migration`:
```bash
services:
  migration:
    profiles:
      - migration
    command:
      - env
    env_file:
      - /root/.env
    image: nginx
    networks:
      default: null
networks:
  default:
    name: example_default
```

Here is the output of `docker compose config migration` on `Docker Compose version v2.33.1`:
```bash
services:
  migration:
    profiles:
      - migration
    command:
      - env
    environment:
      TEST: "True"
    image: nginx
    networks:
      default: null
networks:
  default:
    name: example_default
```

### Compose Version

```Text
Docker Compose version v2.34.0
```

### Docker Environment

```Text
Client: Docker Engine - Community
 Version:           28.0.2
 API version:       1.48
 Go version:        go1.23.7
 Git commit:        0442a73
 Built:             Wed Mar 19 14:36:49 2025
 OS/Arch:           linux/amd64
 Context:           default

Server: Docker Engine - Community
 Engine:
  Version:          28.0.2
  API version:      1.48 (minimum version 1.24)
  Go version:       go1.23.7
  Git commit:       bea4de2
  Built:            Wed Mar 19 14:36:49 2025
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.7.25
  GitCommit:        bcc810d6b9066471b0b6fa75f557a15a1cbf31bb
 runc:
  Version:          1.2.4
  GitCommit:        v1.2.4-0-g6c52b3f
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
root@prod1:~# docker info
Client: Docker Engine - Community
 Version:    28.0.2
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.22.0
    Path:     /usr/libexec/docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.34.0
    Path:     /usr/libexec/docker/cli-plugins/docker-compose

Server:
 Containers: 8
  Running: 8
  Paused: 0
[...]

PR #12665, opened 2025-03-24
introduce config --no-env-resolution

ndeloof:
**What I did**
introduce `config --no-env-resolution` but resolve env_file by default
also fix a potential nil map assignment (can't reproduce, but better safe than sorry)

**Related issue**
closes https://github.com/docker/compose/issues/12659

### DIFF
--- a/cmd/compose/config.go
+++ b/cmd/compose/config.go
@@ -47,6 +47,7 @@ type configOptions struct {
 	noInterpolate       bool
 	noNormalize         bool
 	noResolvePath       bool
+	noResolveEnv        bool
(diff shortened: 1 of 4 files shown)


glours:
@ndeloof you need to run the doc generation as you added a new flag 😉

dxdc:
> also fix a potential nil map assignment (can't reproduce, but better safe than sorry)

@ndeloof @glours did you test without any `.env` file present (even, no empty file), but still manually specifying env flags on the cli? That is how I am getting the nil map assignment, though haven't tested this patch.

Issue #12658, opened 2025-03-22
[BUG] injected secrets are no longer world readable by default


r3mi:
### Description

after upgrading to docker compose v2.34.0, file secrets are not long world readable by default.
According to [documentation](https://docs.docker.com/reference/compose-file/services/#secrets), should be `0444` by default.

### Steps To Reproduce

1. create `docker-compose.yml`
```
services:
  bug:
    image: bash
    secrets:
      - USER
    command: "ls -la /run/secrets"
secrets:
  USER:
    environment: "USER"
```
2. `docker compose run bug`
3. using docker compose v2.33.1
```
-r--r--r--    1 root     root             4 Mar 22 13:25 USER
```
4. using docker compose v2.34.0
```
-r--r-----    1 root     root             4 Mar 22 13:26 USER
```

### Compose Version

```Text
Docker Compose version v2.34.0
```

### Docker Environment

```Text
Client: Docker Engine - Community
 Version:    28.0.2
 Context:    rootless
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.22.0
    Path:     /usr/libexec/docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.34.0
    Path:     /usr/libexec/docker/cli-plugins/docker-compose

Server:
 Containers: 62
  Running: 59
  Paused: 0
  Stopped: 3
 Images: 81
 Server Version: 28.0.2
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: true
  userxattr: true
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: bcc810d6b9066471b0b6fa75f557a15a1cbf31bb
 runc version: v1.2.4-0-g6c52b3f
 init version: de40ad0
 Security Options:
  seccomp
   Profile: builtin
  rootless
  cgroupns
 Kernel Version: 6.8.0-52-generic
 Operating System: Ubuntu 22.04.5 LTS
 OSType: linux
 Architecture: x86_64
[...]

PR #12633, opened 2025-03-13
bump compose-go to version v2.4.9

glours:
**What I did**
Bump `compose-go` to the latest version `v2.4.9`

**Related issue**
N/A

**(not mandatory) A picture of a cute animal, if possible in relation to what you did**
image

PR #12666, opened 2025-03-25
fix(secrets): Reverted secrets file mode 440 -> 444

idsulik:
**What I did**
Reverted secrets file mode from 440 to 444. it looks like it was changed by mistake in this PR https://github.com/docker/compose/pull/12633

**Related issue**
https://github.com/docker/compose/issues/12658


ndeloof (approved):

shaoye:
2.35.0 is still not available in docker's APT repository

could you add it? right now we have to downgrade to 2.33.1 as a workaround

https://docs.docker.com/engine/install/ubuntu/#install-using-the-repository

sin-ack:
Yes, this should have been released as 2.34.1 given that it's a regression.

PR #12699, opened 2025-04-02
use bake by default

ndeloof:
user can still opt-out by `COMPOSE_BAKE=false`
a future change will deprecate this option, so that we can remove the buildkit builder and fully rely on buildx


glours (approved):
LGTM

PR #12715, opened 2025-04-07
pkg/compose: implement Export using atomicwriter

thaJeztah:
Replaces the use of cli/command.CopyToFile with an atomicwriter, as cli/command.CopyToFile will be deprecated in the next release.

**What I did**

**Related issue**


pkg/compose/export.go
```diff
@@ -41,11 +42,11 @@ func (s *composeService) export(ctx context.Context, projectName string, options
...
-	if options.Output == "" && s.dockerCli.Out().IsTerminal() {
-		return fmt.Errorf("output option is required when exporting to terminal")
-	}
-
-	if err := command.ValidateOutputPath(options.Output); err != nil {
+	if options.Output == "" {
+		if s.dockerCli.Out().IsTerminal() {
+			return fmt.Errorf("output option is required when exporting to terminal")
+		}
+	} else if err := command.ValidateOutputPath(options.Output); err != nil {
```
thaJeztah, line 49: FWIW; `atomicwriter.New()` also performs this validation, so we could have a look if we still need this check before we run that code 🤔 

I should also have a look at; the `copy` commands;
https://github.com/docker/compose/blob/a1f673dcf54d8031eeae952630eb71b3f6f3f3b1/pkg/compose/cp.go#L174-L177

https://github.com/docker/compose/blob/a1f673dcf54d8031eeae952630eb71b3f6f3f3b1/pkg/compose/cp.go#L255-L257

ndeloof (approved):

Issue #12121, opened 2024-09-13
[BUG] Bind mounts no longer work in Bitbucket Pipelines in v2.29.3


acarapetis:
### Description

First of all, apologies for the framing here - I don't know enough about the Docker API to understand what the real problem is here, and calling this a bug is probably wrong.

Since upgrading our pipelines runner image to use compose v2.29.3, attempting to run a compose configuration that uses bind mounts in Bitbucket Pipelines is producing this error:

```Error response from daemon: authorization denied by plugin pipelines: --mounts is not allowed```

I'm guessing the relevant change here is #12078, and the problem is that compose has switched to using a new API for mounts that Bitbucket does not support.

Is there a way to revert to using the old API for bind mounts? Or do we just need to pin v2.29.2 for the foreseeable future?  

### Steps To Reproduce

```# docker-compose.yml
services:
    blah:
        image: alpine
        command: ["echo", "hi"]
        volumes:
            - .:/dot
```

```# bitbucket-pipelines.yml
pipelines:
  default:
    - step:
      name: Test compose mount
      services:
        - docker
      script:
        - docker info
        - docker compose run blah
```

Push to a Bitbucket Cloud repository.

### Compose Version

```Text
Docker Compose version v2.29.3
```

### Docker Environment

```Text
+ docker info
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  compose: Docker Compose (Docker Inc., v2.29.3)
Server:
 Containers: 0
  Running: 0
  Paused: 0
  Stopped: 0
 Images: 0
 Server Version: 25.0.2
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
WARNING: API is accessible on http://0.0.0.0:2375 without encryption.
         Access to the remote API is equivalent to root access on the host. Refer
         to the 'Docker daemon attack surface' section in the documentation for
         more information: https://docs.docker.com/go/attack-surface/
  Using metacopy: false
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 1
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Authorization: pipelines
  Log: awslogs fluentd gcplogs gelf journald json-file local splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 runc
 Default Runtime: runc
 Init Binary: docker-init
[...]

PR #12734, opened 2025-04-11
test to cover preference for bind API

ndeloof:
**What I did**
added a test case to check we always prefer the `bind` API when `mount` is not required

**Related issue**
test to prevent regression reported on https://github.com/docker/compose/issues/12121


pkg/compose/create.go
```diff
@@ -874,8 +866,8 @@ func (s *composeService) buildContainerVolumes(
 			v := findVolumeByTarget(service.Volumes, m.Target)
 			vol := findVolumeByName(p.Volumes, m.Source)
 			if v != nil && vol != nil {
-				if _, ok := vol.DriverOpts["device"]; ok && vol.Driver == "local" && vol.DriverOpts["o"] == "bind" {
```
ndeloof, line 877: removed this earlier fix, as with `volumeRequiresMountAPI` we will prefer `bind` for all volumes without advanced mount options, not just those created with `o=bind`

pkg/compose/create.go
```diff
@@ -838,13 +837,7 @@ func (s *composeService) buildContainerVolumes(
 	var mounts []mount.Mount
 	var binds []string
 
-	img := api.GetImageNameOrDefault(service, p.Name)
```
ndeloof, line 841: moved to `buildContainerMountOptions` as this is only used when we manage anonymous volumes inheritance (allows to write test without a useless API call/mock)

glours (approved):
LGTM

vg22:
@ndeloof  Will this fix be available in 2.36.0 release?

glours:
@vg22 the issue should be fixed in the latest `v2.35.1` release

PR #12746, opened 2025-04-15
volumes: support for api-socket volume type

stevvooe:
**What I did**

The `api-socket` volume type allows an easy path to run containers that need access to the Docker API socket with a simple configuration on the service definition:

    service:
      foo:
        ...
        volumes:
          - type: api-socket

The bind mount and credentials are automatically injected using existing service infrastructure. Some work is still needed to make this a bit more robust.

**Related issue**

https://github.com/docker/cli/pull/5858
https://github.com/compose-spec/compose-go/pull/770
https://github.com/compose-spec/compose-spec/pull/587

### DIFF
--- a/pkg/compose/create.go
+++ b/pkg/compose/create.go
@@ -50,6 +50,11 @@ import (
 	cdi "tags.cncf.io/container-device-interface/pkg/parser"
 )
 
+const (
+	secretsDir                  = "/run/secrets/"
+	dockerConfigPathInContainer = secretsDir + "/docker/config.json"
+)
+
 type createOptions struct {
 	AutoRemove        bool
 	AttachStdin       bool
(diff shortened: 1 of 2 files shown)


ndeloof:
For better alignment with docker CLI UX, we should b introduce a `use_api_socket` attribute in the compose spec.
I also wonder this is limited to unix socket. As implementation relies on `DockerEndpoint().Host`, running on Docker Desktop Windows the `npipe://` URL will prevent use of this feature, while the actual engine in VM can expose a docker.sock to containers. Same applies to a remote docker engine accessed over ssh or http. I agree with https://github.com/docker/cli/pull/5858#issuecomment-2751892688 this should be a feature exposed by docker engine, not a hack by client code. At least this should be considered a temporary solution until engine has support for it

stevvooe:
I agree that we should look to the engine side long term but there are some difficulties in making that work. I'll move this to be `use_api_socket` and work from there.

I'll look into the `npipe` issue. That might require a rewrite to the unix socket to do correctly but I'm not sure if we have access to that mapping at the cli-level.

ndeloof:
see alternative implementation https://github.com/docker/compose/pull/12908

Issue #12747, opened 2025-04-15
[BUG] fatal error: concurrent map writes when pulling


samifruit514:
### Description

From time to time, docker-compose up trigger errors when pulling (before bringing up the containers)
here's the error/exception:
```
fatal error: concurrent map writes

goroutine 116 [running]:
github.com/docker/compose/v2/pkg/compose.(*composeService).pullRequiredImages.func1.1()
        github.com/docker/compose/v2/pkg/compose/pull.go:328 +0x198
golang.org/x/sync/errgroup.(*Group).Go.func1()
        golang.org/x/sync@v0.13.0/errgroup/errgroup.go:79 +0x54
created by golang.org/x/sync/errgroup.(*Group).Go in goroutine 114
        golang.org/x/sync@v0.13.0/errgroup/errgroup.go:76 +0x98

goroutine 1 [semacquire]:
sync.runtime_Semacquire(0x400056a058?)
        runtime/sema.go:71 +0x2c
sync.(*WaitGroup).Wait(0x40005b6d88)
        sync/waitgroup.go:118 +0x74
golang.org/x/sync/errgroup.(*Group).Wait(0x40005b6d80)
        golang.org/x/sync@v0.13.0/errgroup/errgroup.go:56 +0x2c
github.com/docker/compose/v2/pkg/progress.RunWithStatus({0x26435c0, 0x4000834510}, 0x4000694520, 0x400070d920, {0x22653ce, 0x7})
        github.com/docker/compose/v2/pkg/progress/writer.go:97 +0x1dc
github.com/docker/compose/v2/pkg/progress.Run({0x26435c0, 0x4000834510}, 0x40005b6d40, 0x400070d920)
        github.com/docker/compose/v2/pkg/progress/writer.go:61 +0x84
github.com/docker/compose/v2/pkg/compose.(*composeService).Up(0x40004a1e80, {0x26435c0, 0x4000834510}, _, {{0x40002a1c80, {0x400049e500, 0x0, 0x2}, 0x0, 0x0, ...}, ...})
        github.com/docker/compose/v2/pkg/compose/up.go:40 +0x198
```

### Steps To Reproduce

I dont know how to reproduce but it seems that it happens when several docker-compose on the same node pulls the same image?

### Compose Version

```Text
v2.35.0
```

### Docker Environment

```Text

```

glours:
Hey @samifruit514 👋 

Can you provide a minimal Compose file to reproducible the issue, please

[...]

PR #12752, opened 2025-04-17
fix: concurrent map writes when pulling

skanehira:
**What I did**
fixes https://github.com/docker/compose/issues/12747

**Related issue**


samifruit514:
just want to let you know that I've tested it and everything is ok so far. No failures!
Thank you for your fast response

ndeloof (approved):

PR #12769, opened 2025-04-22
Add support for COMPOSE_PROGRESS env variable

### LINKED ISSUE #12751: [FEATURE REQUEST] Allow the `--quiet` and `--quiet-build` options from the `run` command to be provided from environment variables
### Description

I need to use the behavior provided by these options in all the `docker compose run` commands executed in my CI environment.

I think it could be nice to be able to configure these options globally through environment variables, similar to how we do with other options, like [`BUILDKIT_PROGRESS`](https://docs.docker.com/build/building/variables/#buildkit_progress).

Relates to #12685.

AnvarU:
**What I did**
Added support for the new env variable COMPOSE_PROGRESS to suppress output.

Behaviour is similar to `docker compose --progress=<mode> ...`

**Related issue**

closes #12751 

### DIFF
--- a/pkg/progress/writer.go
+++ b/pkg/progress/writer.go
@@ -19,6 +19,7 @@ package progress
 import (
 	"context"
 	"io"
+	"os"
 	"sync"
 
 	"github.com/docker/cli/cli/streams"
@@ -120,6 +121,9 @@ func NewWriter(ctx context.Context, out *streams.Out, progressTitle string) (Wri
 	dryRun, ok := ctx.Value(api.DryRunKey{}).(bool)
 	if !ok {
 		dryRun = false
+	}
+	if v, ok := os.LookupEnv("COMPOSE_PROGRESS"); ok && Mode == ModeAuto {
+		Mode = v
 	}
 	if Mode == ModeQuiet {
 		return quiet{}, nil
--- a/pkg/progress/writer.go
+++ b/pkg/progress/writer.go
@@ -19,7 +19,6 @@ package progress
 import (
 	"context"
 	"io"
-	"os"
 	"sync"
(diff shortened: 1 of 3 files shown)


ndeloof:
IMHO it would make more sense to add support for `COMPOSE_PROGRESS` and apply this to all commands, vs having a specific one for `run` case

AnvarU:
> IMHO it would make more sense to add support for `COMPOSE_PROGRESS` and apply this to all commands, vs having a specific one for `run` case

@ndeloof I definitely agree with this, and I was thinking about the `COMPOSE_PROGRESS`. This approach will require more changes and will take more time for me. If you're fine with this, I will put this PR to draft and request a review once I'm done with implementation. And I guess I would need to do the PR to the docs repository [here](https://github.com/docker/docs/tree/main), am I right?

pkg/progress/writer.go
```diff
@@ -121,6 +122,9 @@ func NewWriter(ctx context.Context, out *streams.Out, progressTitle string) (Wri
 	if !ok {
 		dryRun = false
 	}
+	if v, ok := os.LookupEnv("COMPOSE_PROGRESS"); ok && Mode == ModeAuto {
```
ndeloof, line 125: a better place to manage this is `addProjectFlags` in cmd/compose.go, as you can set flag default value from env value (the same way we do for `--env-file`)

AnvarU: Done [ccfd133](https://github.com/docker/compose/pull/12769/commits/ccfd13384871f9d46ae9bbe90b06f07aa6a73602)

ndeloof (approved):

PR #12777, opened 2025-04-25
document extensibility using service.provider

ndeloof:
**What I did**

Document extensibility mechanism using `service.provider` to configure third-party command responsible to manage service lifecycle
Minor change to the "API": pass service name, so that same provider can be used to configure multiple resources and tag them with service name for later removal, as `compose down` _may_ be ran with only a subset of the model

### DIFF
--- a/docs/extension.md
+++ b/docs/extension.md
@@ -26,13 +26,13 @@ the resource(s) needed to run a service.
 - Another Docker CLI plugin (typically, `model` to run `docker-model`)
 - An executable in user's `PATH`
 
-To be a valid Compose extension, provider command *MUST* accept subcommand `compose` (which can be hidden)
+To be a valid Compose extension, provider command *MUST* accept a `compose` command (which can be hidden)
 with subcommands `up` and `down`.
 
 ## Up lifecycle
 
 To execute an application's `up` lifecycle, Compose executes the provider's `compose up` command, passing 
-the project name, service name and additional options. The `provider.options` are translated 
+the project name, service name, and additional options. The `provider.options` are translated 
 into command line flags. For example:
 ```console
 awesomecloud compose --project-name <NAME> up --type=mysql --size=256 "database"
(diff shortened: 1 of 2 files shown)


pkg/compose/plugins.go
```diff
@@ -156,24 +164,3 @@ func (s *composeService) setupPluginCommand(ctx context.Context, project *types.
...
-			var exitErr *exec.ExitError
-			if errors.As(err, &exitErr) && exitErr.ExitCode() == 1 {
-				return fmt.Errorf("you should enable model runner to use %q provider services: %s", plugin.Name, err.Error())
-			}
-		}
-	} else {
-		return fmt.Errorf("unsupported provider %q", plugin.Name)
-	}
-	return nil
-}
```
glours, line 179: Maybe we could keep it for Model Runner and just remove the `unsupported provider` error, wdyt?

ndeloof: If plugin is installed, but disabled in DD, won't it report an adequate error message ?

ndeloof:
```
$ docker compose up
[+] Running 1/1
 ✘ node  error                                                                                                                                                0.0s 
Failed to pull model: service unavailable
```
(model-cli could be improved to send a more user-friendly error)

docs/extension.md
```diff
@@ -0,0 +1,102 @@
...
+      options:
+        type: mysql
+        size: 256
+```
+
+`provider.type` tells Compose the binary to run, which can be either:
+- another Docker CLI plugin (typically, `model` to run `docker-model`)
+- an executable in user's `PATH`
+
+To be a valid Compose extension, provider command *MUST* accept subcommand `compose` (which can be hidden)
```
glours, line 29: nit:
```suggestion
To be a valid Compose extension, provider command *MUST* accept a `compose` command (which can be hidden)
```

glours (commented):
Almost good to go for me 👍

glours (approved):
LGTM

PR #12781, opened 2025-04-25
add support for external binary with provider services

glours:
**What I did**
Added the capability of running command from external binaries from provider services definition.

**Related issue**
N/A

### DIFF
--- a/pkg/compose/plugins.go
+++ b/pkg/compose/plugins.go
@@ -52,9 +52,19 @@ func (s *composeService) runPlugin(ctx context.Context, project *types.Project, 
 
 	plugin, err := s.getPluginBinaryPath(provider.Type)
 	if err != nil {
-		return err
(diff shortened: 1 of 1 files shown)


ndeloof:
already implemented in https://github.com/docker/compose/pull/12777 :)

pkg/compose/plugins.go
```diff
@@ -52,9 +52,19 @@ func (s *composeService) runPlugin(ctx context.Context, project *types.Project,
 
 	plugin, err := s.getPluginBinaryPath(provider.Type)
 	if err != nil {
-		return err
+		if !strings.Contains(err.Error(), "Error: No such CLI plugin") {
```
ndeloof, line 55: manager.IsNotFound

glours:
Close in favor of #12777

Issue #12774, opened 2025-04-24
[BUG] docker compose does not start newly built image when `COMPOSE_BAKE` is true


Paraphraser:
### Description

Running `docker compose up -d --build «container»` in the presence of `COMPOSE_BAKE=true` does not re-create the container with the newly-built image. Removing `COMPOSE_BAKE` from the environment returns to the original behaviour.

## Scenario

I am using an instance of a customised MariaDB container intended to be the back-end for a Gitea container. Although I haven't tested with different base images, I don't believe that the base image is relevant. MariaDB is simply where I noticed this problem and am able to demonstrate it easily.

### Instrumentation

Instrumentation used to demonstrate problem:

* Health-check script includes:

	``` sh
	FORCE_CHANGE=1
	echo "Forced change marker = $FORCE_CHANGE"
	```

* Dockerfile includes:

	``` Dockerfile
	RUN date >/builddate.txt
	```
	
	This is positioned **after** the health-check script is copied into the image.

### Runtime environment

``` console
$ echo $COMPOSE_BAKE
true
$
```

### Test 1

1. Show container not running:

	``` console
	$ docker top gitea_db
	Error response from daemon: No such container: gitea_db
	```

2. Build container:

	``` console
	$ docker compose up -d --build gitea_db
	[+] Building 0.5s (11/11) FINISHED                                                                                                                              
	 => [internal] load local bake definitions                                                                                                                 0.0s
	 => => reading from stdin 382B                                                                                                                             0.0s
	 => [internal] load build definition from Dockerfile                                                                                                       0.0s
	 => => transferring dockerfile: 1.16kB                                                                                                                     0.0s
	 => [internal] load metadata for ghcr.io/linuxserver/mariadb:latest                                                                                        0.0s
	 => [internal] load .dockerignore                                                                                                                          0.0s
	 => => transferring context: 2B                                                                                                                            0.0s
	 => [1/4] FROM ghcr.io/linuxserver/mariadb:latest                                                                                                          0.0s
	 => [internal] load build context                                                                                                                          0.0s
	 => => transferring context: 45B                                                                                                                           0.0s
	 => CACHED [2/4] RUN for CNF in /defaults/my.cnf /defaults/custom.cnf ; do [ -f ${CNF} ] && break ; done ;     sed -i.bak         -e "s/^thread_cache_siz  0.0s
[...]

PR #12784, opened 2025-04-28
fix collect image digests for service images built by bake

ndeloof:
**What I did**
Fix: collect build images from bake output metdata indexed by expected image tag, not service name
Note: bake returns `image.name` but we can't rely on this as this is set with fully qualified name (`docker.io/library/foo:latest`) and user could have configured a "familiar name" (`foo` or `foo:latest`). So I preferred to use the list of expected built targets and check those were found in metadata.

**Related issue**
fixes https://github.com/docker/compose/issues/12774


glours (approved):
LGTM

Issue #12783, opened 2025-04-27
Usage of additional_contexts with the service: directive should imply build time dependencies


davireis:
### Description

The new service: directive for additional_contexts almost fulfills the dream of letting docker/docker-compose become highly usable in a monorepo scenario, where images depend on other images built on demand in the same repository. See https://stackoverflow.com/a/79487072/95653 for an 8-year-old question in that direction finally getting an answer.

However, the build time dependency created by the additional context is not represented anywhere, and as a result, most commands break. Please review the example below, which is adapted from the aforementioned Stack Overflow answer.

```services:
  a:
     image: service_a
     build: 
       dockerfile_inline: "FROM alpine"
  b:
     image: service_b
     build:
       args:
         BASE_IMAGE: service_a
       dockerfile_inline: "FROM service_a"
       additional_contexts:
         # instructs resolving `FROM service_a` to "service a" image
         service_a: "service:a"
``` 

If you run `docker compose build`, everything works as expected. However, if I do `docker compose build b`, I get `failed to find target a`. The problem can be easily understood by running `docker compose config b`, which will show that service a does not exist in the configuration when you specify b explicitly.

There is an ugly workaround, which is using a depends_on for creating a runtime dependency, which will force a into the config when b is specified explicitly. This will let me do `docker compose run --build b`, but it will startup both service b and a, and I although I think it is reasonable to want b, I usually do not want a. One can start to play with more workarounds, like adding a parallel tree of build dependencies with a no-op entrypoint, but everything starts to get cumbersome and you start to hit other bugs. 

Everything would work out of the box if `docker compose config b` brought a into the config, but somehow managed to mark it as not to start.

Going further down, once this is fixed, the build time dependency graph should also be exposed to docker compose watch, but I guess that goes as a different bug later.

ndeloof:
requalified as a bug, my [earlier fix](https://github.com/docker/compose/pull/12683) for this was incomplete

davireis:
Hi @ndeloof ,

Thanks a lot for the fix. Running `docker compose build b` now works fine, but `docker compose run --build b` and `docker compose up --build b` still fail, each one with a different error.

```
pkgx +docker-compose@2.36.0 docker-compose up  --build b
Compose can now delegate builds to bake for better performance.
 To do so, set COMPOSE_BAKE=true.
[+] Building 0.0s (1/1) FINISHED                                              docker:desktop-linux
 => [b internal] load build definition from Dockerfile                                        0.0s
[...]

PR #12785, opened 2025-04-28
enable services implicitly declared by a service:xx build dependency

ndeloof:
**What I did**
As we detect some services are declared as implicit dependencies by `additional_context: service:xx`, enable those services in project so they also get built

**Related issue**
fix https://github.com/docker/compose/issues/12783


glours (approved):
LGTM

PR #12788, opened 2025-04-28
display proper event message for provider services on up and down

glours:
**What I did**
Display proper event message when running `up` and `down` plugin commands
Fix cyclomatic alert for `runPlugin` function

**Related issue**
N/A

image

### DIFF
--- a/pkg/compose/plugins.go
+++ b/pkg/compose/plugins.go
@@ -61,15 +61,33 @@ func (s *composeService) runPlugin(ctx context.Context, project *types.Project, 
 
 	cmd := s.setupPluginCommand(ctx, project, provider, plugin.Path, command)
 
-	eg := errgroup.Group{}
-	stdout, err := cmd.StdoutPipe()
+	variables, err := s.executePlugin(ctx, cmd, command, service)
 	if err != nil {
 		return err
 	}
 
+	for name, s := range project.Services {
+		if _, ok := s.DependsOn[service.Name]; ok {
+			prefix := strings.ToUpper(service.Name) + "_"
+			for key, val := range variables {
+				s.Environment[prefix+key] = &val
+			}
+			project.Services[name] = s
+		}
+	}
+	return nil
+}
+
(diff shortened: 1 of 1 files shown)


pkg/compose/plugins.go
```diff
@@ -61,15 +61,33 @@ func (s *composeService) runPlugin(ctx context.Context, project *types.Project,
...
 
+	variables, err := s.executePlugin(ctx, cmd, command, service)
+	if err != nil {
+		return err
+	}
+
+	prefix := strings.ToUpper(service.Name) + "_"
+	for name, s := range project.Services {
+		if _, ok := s.DependsOn[service.Name]; ok {
+			for key, val := range variables {
```
thaJeztah, line 72: micro-nit 😅 (definitely not blocking); looks like the `prefix` is only used when `depends_on` matches, so could be put inside that branch; 

```suggestion
	for name, s := range project.Services {
		if _, ok := s.DependsOn[service.Name]; ok {
			prefix := strings.ToUpper(service.Name) + "_"
			for key, val := range variables {
```

pkg/compose/plugins.go
```diff
@@ -79,50 +97,55 @@ func (s *composeService) runPlugin(ctx context.Context, project *types.Project,
...
-	pw.Event(progress.CreatingEvent(service.Name))
+	var action string
+	switch command {
+	case "up":
+		pw.Event(progress.CreatingEvent(service.Name))
+		action = "create"
+	case "down":
+		pw.Event(progress.RemovingEvent(service.Name))
+		action = "remove"
+	}
```
thaJeztah, line 108: Should this switch also have a default? Looks like if `command` is not `up` or `down`, then the switch at the end of the function returns an error (but after a bunch of work is done). Maybe that's intentional, but I just noticed it.

If it's intentional, it would be good to add a `default:` with a comment that explains that it is intentional (but will produce an error further down).

ndeloof (approved):

pkg/compose/plugins.go
glours: Yep, I added the `default` to the wrong switch 🤦

Issue #12280, opened 2024-11-11
[BUG] Inconsistency between license information


CarstenLeue:
### Description

The license information of the docker [compose source](https://github.com/docker/compose/blob/main/LICENSE) code is the Apache License. The information on the releases page for the [downloadable binaries](https://github.com/docker/compose/releases/download/v2.30.3/LICENSE) however is the docker license.

Is this on purpose? If so does this imply that one is allowed download the source code and build the binaries on their own using the Apache license? 

Purpose is to use the docker compose binaries outside of the scope of docker desktop.

### Steps To Reproduce

Consult

https://github.com/docker/compose/releases/download/v2.30.3/LICENSE
vs
https://github.com/docker/compose/blob/main/LICENSE

### Compose Version

```Text

```

### Docker Environment

```Text

```

ndeloof:
> Is this on purpose? 

Yes. The source code is under Apache License - you can reuse it and build your own compose binaries - but the pre-built binaries as distributed under Docker End User License Agreement. While arguably confusing, this has no impact on using those regardless your usage of Docker Desktop.

ndeloof:
https://github.com/docker/compose/blob/main/packaging/LICENSE should offer a link to "Docker End User License Agreement" for clarification

ndeloof:
> are you suggesting that they are equivalent

nope, as Docker EULA only applies to binaries and is not an OSS license. I just mean using Compose binaries standalone doesn't count as a "seat" regarding Docker subscription licensing.

PR #12829, opened 2025-05-14
remove Docker EULA licensing which isn't relevant

ndeloof:
There's no Docker EULA anymore since Docker switched to a subscription model

fix https://github.com/docker/compose/issues/12280

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -307,9 +307,6 @@ jobs:
             fi
           done
       -
-        name: License
-        run: cp packaging/* ./bin/release/
-      -
         name: List artifacts
         run: |
           tree -nh ./bin/release
--- a/packaging/LICENSE
+++ b/packaging/LICENSE
@@ -1,2 +1,0 @@
-The Docker End User License Agreement (https://www.docker.com/legal/docker-software-end-user-license-agreement) describes Docker's Terms for this software.
-By downloading, accessing, or using this software you expressly accept and agree to the Terms set out in the Docker End User License Agreement.


ndeloof:
cc @rickredondo

glours (approved):

LICENSE
```diff
@@ -187,7 +187,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright [yyyy] [name of copyright owner]
+   Copyright 2025 Docker, Inc.
```
thaJeztah, line 190: This should not be filled in in the license; this section is an example how to apply the license (which is done by adding the headers to the source code and/or a NOTICE file).

We also don't own the copyright, so this would be incorrect in either case

ndeloof: same on https://github.com/moby/moby/blob/master/LICENSE#L179 😅

thaJeztah: I know! And it's been wrong since forever. Still want to change that (and apply proper license headers), but I may need approval from legal

thaJeztah: The TL;DR is that (for Apache 2)

- Adding a LICENSE file to the repository does NOT apply the license (the LICENSE file is only there for convenience, and not even strictly neeeded; it's just so that those using the software don't have to download it from the apache website)
- The "copyright" section in the Apache 2 LICENSE file is basically part of the documentation; the section above it literally describes "how to apply the license", which mentions "to apply the license, take the snippet below, fill in the copyright lines, and add it to each and every file that the license must be applied to". Also add a NOTICE file to document other legal constraints or licenses (such as patents, licenses of dependencies used (go.mod / vendor))
- Given that we DON'T have a CLA, no copyright is transferred, which means that the copyright is owned by the "AUTHOR" of each line of code.
- :point_up: and because mentioning each individual author ... for each file .. would be a bit lengthy, it's common to use "The Authors" (or "The `<insert project name>` Authors") as copyright holder.
- :point_up: and the copyright "year" is implied from year of publication for each line of code, which basically is "date it was pushed to GitHub", so the copyright-years will be different for each file
- :point_up: given that source is in source-control, it's therefore usually OK to omit copyright year(s) altogether, as it can be deducted from git-history. But alternatively, it's not uncommon to add a "first year of publication of the code (as a whole)"

Also see;

- https://github.com/containerd/containerd/commit/b12c3215a0af92b9244b14c8555fc028b0d1d8c1
- https://github.com/kubernetes/kubernetes/commit/d30db1f9a915aa95402e1190461469a1889d92be
    - related discussion; https://github.com/kubernetes/kubernetes-template-project/issues/3

thaJeztah: Related to the above; this is also why there's an `AUTHORS` file in the Moby (and other) repositories, as that effectively lists the copyright-holders; https://github.com/moby/moby/blob/master/AUTHORS

.github/workflows/ci.yml
```diff
@@ -306,9 +306,6 @@ jobs:
               echo "$sum $file" > ${file#\*}.sha256
             fi
           done
-      -
-        name: License
-        run: cp packaging/* ./bin/release/
```
thaJeztah, line 311: Should this one be swapped for the top-level `LICENSE`, or was a license already added as part of one of the steps?

ndeloof: AFAIK compose is the only docker artifact where we include a LICENSE file with built artifact (for legacy reasons)
Licensing is already set by github project and source code archive

LICENSE
ndeloof: dropped second commit

.github/workflows/ci.yml
thaJeztah: Ah, right; I wasn't sure if folks would expect it to be there.

I know / recall it's included in the `.deb` and/or `.rpm` packages but that's probably separate from this?

Issue #12836, opened 2025-05-15
[FEAT] Sub-command for locking/pinning images to digests with lock-file support for the compose command in general


andoks:
### Description

Currently, if one wants to ensure supply-chain security with regards to use of docker images in docker-compose files, users have to specify the digest manually in the compose file.

Previously there was a project ([docker-lock](https://pkg.go.dev/github.com/safe-waters/docker-lock#section-readme)) that could automate this, but this project has been deleted.

It would be great if docker-compose supported pinning images to digest, somewhat like [npm lock-files](https://docs.npmjs.com/cli/v7/configuring-npm/package-lock-json), [python lock-files](https://pip.pypa.io/en/stable/cli/pip_lock/) and [go sum-file](https://go.dev/ref/mod#go-sum-files).

If docker-compose could also support updating said images digest upon request, that would be great too, but that would not be necessary for a Minimal Viable Product version of the feature IMO, as long as i can request a "re-lock" from compose that updates the digests in the assumed new lock-file, when I have updated the tag in the compose-file

ndeloof:
You can rely on `docker compose config --resolve-image-digests` to fix image digest.
Then som yq trickery can be used to generate an override file that will only contain the image references

```
$ docker compose config --resolve-image-digests | yq '{"services": (.services|to_entries|.[]|{.key: {"image": .value.image}})}'

services:
  test:
    image: docker.io/library/alpine:latest@sha256:a8560b36e8b8210634f77d9f7f9efd7ffa463e380b75e2e74aff4511df3ef88c
```

This feature is actually already implemented in Docker Compose for the benefits of the `publish` command, so maybe we could  expose it as `docker compose config --image-digests`

andoks:
Relates to https://github.com/docker/compose/issues/5818

PR #12843, opened 2025-05-16
introduce config --lock-image-digests

ndeloof:
**What I did**
introduce `config --lock-image-digests`. This produces an override file with image digest locked, comparable to the one generated by `publish`

**Related issue**
fixes https://github.com/docker/compose/issues/12836


glours (approved):
LGTM

PR #12847, opened 2025-05-19
docs: add an example of extension

LinuxSuRen:
**What I did**

As mentioned in the title.

**Related issue**


ndeloof:
That's a good idea having example implementation, anyway I'd prefer we keep such an example within docker/compose project so it is easier for us to maintain, see https://github.com/docker/compose/pull/12848

LinuxSuRen:
@ndeloof Agree with you about the ideas of examples. How about adding the tool (which mentioned in this PR) into an existing tool list?

PR #12848, opened 2025-05-19
example provider implementation

ndeloof:
**What I did**

**Related issue**
see https://github.com/docker/compose/pull/12847

### DIFF
--- a/docs/examples/provider.go
+++ b/docs/examples/provider.go
@@ -1,0 +1,74 @@
+/*
+   Copyright 2020 Docker Compose CLI authors
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


docs/examples/provider.go
```diff
@@ -0,0 +1,70 @@
...
+	servicename := args[0]
+	for i := 0; i < 100; i += 10 {
+		time.Sleep(1 * time.Second)
+		fmt.Printf(`{ "type": "info", "message": "Processing ... %d%%" }\n`, i)
+	}
+	fmt.Printf(`{ "type": "setenv", "message": "URL=https://magic.cloud/%s" }\n`, servicename)
+}
+
+func down(_ *cobra.Command, _ []string) {
+	fmt.Printf(`{ "type": "error", "message": "Permission error" }\n`)
```
glours, line 69: Removing `\n`, the parsing fails on the Compose side with the following error `invalid character '\\' looking for beginning of object key string`

```suggestion
		fmt.Printf(`{ "type": "info", "message": "Processing ... %d%%" }`, i)
	}
	fmt.Printf(`{ "type": "setenv", "message": "URL=https://magic.cloud/%s" }`, servicename)
}

func down(_ *cobra.Command, _ []string) {
	fmt.Printf(`{ "type": "error", "message": "Permission error" }`)
```

glours (approved):
just few nits to fix before merging

Issue #12846, opened 2025-05-18
[BUG] docker engine and desktop expose random ports differently


alexanderankin:
### Description

exposing ports with a random mapping should be able to be described using docker compose language portably across docker engine and docker desktop environments. however, different options need to be supplied in order to achieve the goal.

```yaml
services:
  alpine:
    image: nginx:alpine-slim
    init: true
    ports:
      - '81'
      - '82'
      - target: 80
        host_ip: 127.0.0.1
        protocol: tcp
    command:
      - sh
      - -c
      - 'd=/etc/nginx/conf.d; echo "server { listen 81; location / { return 202; } }" > $$d/81.conf && echo "server { listen 82; location / { return 204; } }" > $$d/82.conf && nginx -g "daemon off;"'

  alpine2:
    image: nginx:alpine-slim
    init: true
    ports:
      - target: 80
        host_ip: 127.0.0.1
        protocol: tcp
    command:
      - sh
      - -c
      - 'd=/etc/nginx/conf.d; echo "server { listen 81; location / { return 202; } }" > $$d/81.conf && echo "server { listen 82; location / { return 204; } }" > $$d/82.conf && nginx -g "daemon off;"'
```

e.g. this works on mac but on linux i also have to add (next to both `host_ip` fields) a `published` field and give it a range.

### Steps To Reproduce

1. copy this `compose.yaml` file onto two computers, one linux (ubuntu 24/Docker version 28.0.4, build b8034c0), one mac (Docker version 27.5.1, build 9f9e405)
2. cd into the respective directories and `docker compose up -d` in both
3. observe linux not works (until adding `published`)
4. observe mac works

### Compose Version

```Text
Docker Compose version v2.32.4-desktop.1
Docker Compose version v2.34.0
```

### Docker Environment

```Text
$ docker info | tr '\n' '|' && echo
[...]

PR #12851, opened 2025-05-20
run ContainerStart sequentially

ndeloof:
**What I did**
As compose calls ContainerStart concurrently, Docker engine sometimes fails to allocate a port in a range. This should be fixed in engine, but this preventive workaround has low impact on Compose

**Related issue**
closes https://github.com/docker/compose/issues/12846

### DIFF
--- a/pkg/compose/convergence.go
+++ b/pkg/compose/convergence.go
@@ -650,9 +650,14 @@ func (s *composeService) recreateContainer(ctx context.Context, project *types.P
 	return created, err
 }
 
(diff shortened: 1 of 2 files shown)


thaJeztah:
@ndeloof was this on Docker Desktop, or were you also able to reproduce on Linux?

Wondering if it's the port mapping issues we've seen related to docker desktop.

cc @robmry @akerouanton

ndeloof:
@thaJeztah only applies to Linux, not DD

glours (approved):

PR #12853, opened 2025-05-20
only use attestation when building image outside the development inner loop

glours:
When building a image, by default attestation are generated and modify the image ID which trigger a container recreation on up and run command even if there isn't any changes on the image content itself

**What I did**
* Added the `provenance` attribute to the build options, which is set to true only when build command is used
* Remove attestation if build is coming from inner loop process or if user set `none` mode from the `build` command line

**Related issue**

image

### DIFF
--- a/cmd/compose/build.go
+++ b/cmd/compose/build.go
@@ -46,7 +46,7 @@ type buildOptions struct {
 	deps       bool
 	print      bool
 	check      bool
-	provenance string
+	provenance bool
 }
 
 func (opts buildOptions) toAPIBuildOptions(services []string) (api.BuildOptions, error) {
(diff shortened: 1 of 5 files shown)


crazy-max (commented):
From build tool standpoint, `provenance` flag is a shorthand that can either be a `bool` (to enable/disable) or provenance attributes (atm only mode attribute exists): https://docs.docker.com/build/metadata/attestations/slsa-provenance/#create-provenance-attestations

By default provenance flag is empty: https://github.com/docker/buildx/blob/ea2b7020a4645bff395eb49e4e87ef08ba24eb93/commands/build.go#L648

And it defaults to min: https://github.com/docker/buildx/blob/ea2b7020a4645bff395eb49e4e87ef08ba24eb93/build/opt.go#L153

If compose wants to use this shorthand then I think it should have the same behavior as build provenance flag.

So to summarize `provenance` field should be bool/string type:

* `--provenance=true`: enable provenance
* `--provenance=false`: disable provenance
* `--provenance=mode=xxx`: set mode (xxx can be `min` or `max`)

And defaults to empty so `build.Build` handles the default behavior.

___

In the case of your fix to disable provenance in the inner loop you would just need to set `Attests["provenance"] = nil` to disable provenance.

___

Maybe it would be better to handle that first to fix the issue and implement the `provenance` field in another PR as ti might need more thoughts?

ndeloof (approved):

crazy-max (approved):

PR #12859, opened 2025-05-21
report error (re)creating container

ndeloof:
**What I did**

As an error occurs (re)creating container, update progress with cause, but only for the culprit container - others will get context cancelled and report the same error as cause

```
$ docker compose up
[+] Running 2/2
 ✔ hello-arm64 Pulled                                                                                                                                                              1.1s 
 ✔ hello-amd64 Pulled                                                                                                                                                              1.1s 
[+] Running 1/2
 ✘ Container truc-hello-arm64-1               Error response from daemon: image with reference debian was found but its platform (linux/amd64) d...                                0.0s 
 ⠋ Container 445577bb3fe3_truc-hello-amd64-1  Recreate                                                                                                                             0.0s 
Error response from daemon: image with reference debian was found but its platform (linux/amd64) does not match the specified platform (linux/arm64)

```


glours (approved):
LGTM

PR #12877, opened 2025-05-27
refactor: use slices.Contains to simplify code

tongjicoder:
**What I did**

There is a [new function](https://pkg.go.dev/slices@go1.21.0#Contains) added in the go1.21 standard library, which can make the code more concise and easy to read.

**Related issue**

image


ndeloof:
you could go one step further and remove the StringContains function, replacing with this newer func from stdlib

PR #12903, opened 2025-06-03
add support of metadata subcommand for provider services

glours:
This command will let Compose and external tooling know about which parameters should be passed to the Compose plugin

**What I did**
Propose to add a `metadata` subcommand to provider implementation allowing Compose and LSP tools to get more informations about parameters to be passed.
This will allow LSP tool to provide code completion or hints when setupping a provider service and Compose to validate mandatory property have been correctly define before run the provider for examples

**Related issue**
N/A

image

### DIFF
--- a/docs/examples/provider.go
+++ b/docs/examples/provider.go
@@ -17,11 +17,13 @@
 package main
 
 import (
+	"encoding/json"
 	"fmt"
 	"os"
 	"time"
 
 	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
 )
 
 func main() {
@@ -43,16 +45,27 @@ func composeCommand() *cobra.Command {
 		TraverseChildren: true,
 	}
 	c.PersistentFlags().String("project-name", "", "compose project name") // unused
-	c.AddCommand(&cobra.Command{
+	upCmd := &cobra.Command{
 		Use:  "up",
 		Run:  up,
 		Args: cobra.ExactArgs(1),
-	})
-	c.AddCommand(&cobra.Command{
+	}
+	upCmd.Flags().String("type", "", "Database type (mysql, postgres, etc.)")
+	_ = upCmd.MarkFlagRequired("type")
+	upCmd.Flags().Int("size", 10, "Database size in GB")
+	upCmd.Flags().String("name", "", "Name of the database to be created")
+	_ = upCmd.MarkFlagRequired("name")
+
+	downCmd := &cobra.Command{
 		Use:  "down",
 		Run:  down,
 		Args: cobra.ExactArgs(1),
-	})
+	}
+	downCmd.Flags().String("name", "", "Name of the database to be deleted")
+	_ = downCmd.MarkFlagRequired("name")
+
+	c.AddCommand(upCmd, downCmd)
+	c.AddCommand(metadataCommand(upCmd, downCmd))
 	return c
 }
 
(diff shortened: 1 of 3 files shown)


docs/examples/provider.go
```diff
@@ -72,3 +88,42 @@ func up(_ *cobra.Command, args []string) {
 func down(_ *cobra.Command, _ []string) {
 	fmt.Printf(`{ "type": "error", "message": "Permission error" }%s`, lineSeparator)
 }
+
+func metadata(metadataCmd *cobra.Command, _ []string) {
+	composeCmd := metadataCmd.Parent()
+	metadata := map[string]CommandMetadata{}
+	for _, cmd := range composeCmd.Commands() {
```
ndeloof: you can use `composeCmd.Find("up")`, or could pass `upCmd` to the metadata command

pkg/compose/plugins.go
```diff
@@ -198,3 +204,51 @@ func (s *composeService) setupPluginCommand(ctx context.Context, project *types.
 	cmd.Env = append(cmd.Env, types.Mapping(carrier).Values()...)
 	return cmd
 }
+
+func (s *composeService) getPluginMetadata(path string) map[string]CommandMetadata {
+	cmd := exec.Command(path, "compose", "metadata")
+	stdout := &bytes.Buffer{}
+	cmd.Stdout = stdout
+
+	if err := cmd.Start(); err != nil {
```
ndeloof: can just call `cmd.Run()` as we don't try to decode a long running json stream here

pkg/compose/plugins.go
```diff
@@ -198,3 +204,51 @@ func (s *composeService) setupPluginCommand(ctx context.Context, project *types.
...
+		var exiterr *exec.ExitError
+		if errors.As(err, &exiterr) {
+			logrus.Debugf("plugin metadata command failed with exit code: %d\n", exiterr.ExitCode())
+		}
+		return nil
+	}
+
+	decoder := json.NewDecoder(stdout)
+	var metadata map[string]CommandMetadata
+	if err := decoder.Decode(&metadata); err != nil {
```
ndeloof:
```suggestion
	var metadata map[string]CommandMetadata
	if err := json.Unmarshal(stdout, &metadata); err != nil {
```

pkg/compose/plugins.go
```diff
@@ -198,3 +210,49 @@ func (s *composeService) setupPluginCommand(ctx context.Context, project *types.
...
+	Description string          `json:"description"`
+	Up          CommandMetadata `json:"up"`
+	Down        CommandMetadata `json:"down"`
+}
+
+type CommandMetadata struct {
+	Parameters []ParametersMetadata `json:"parameters"`
+}
+
+type ParametersMetadata struct {
```
rcjsuen: Since this metadata is for a _single_ parameter, should we name this `ParameterMetadata` instead of `ParametersMetadata`?

pkg/compose/plugins.go
```diff
@@ -198,3 +213,59 @@ func (s *composeService) setupPluginCommand(ctx context.Context, project *types.
...
+	// Save metadata into docker home directory to be used by Docker LSP tool
+	// Just log the error as it's not a critical error for the main flow
+	metadataDir := filepath.Join(config.Dir(), providerMetadataDirectory)
+	if err := os.MkdirAll(metadataDir, 0o700); err != nil {
+		logrus.Debugf("failed to create plugin metadata directory: %v", err)
+	}
+	metadataFilePath := filepath.Join(metadataDir, command+".json")
+	if err := os.WriteFile(metadataFilePath, stdout.Bytes(), 0o600); err != nil {
+		logrus.Debugf("failed to save plugin metadata: %v", err)
+	}
```
rcjsuen: Should we only try to save if `MkdirAll` succeeds?

ndeloof (approved):

Issue #12081, opened 2024-08-25
[BUG] Docker Compose pull fails when a required env_file is missing


KostaGorod:
### Description

I Believe this is part of previous requests https://github.com/docker/compose/issues/1973 & https://github.com/docker/compose/issues/9181.

which defines that some commands don't need to validate full config.

my issue that when I pull images, using  env_file with required=true.
the pull fails due to missing env file.

My expectation is that the `required`  is part of running the compose (and validating), not pulling.
If there is situation that env_file is expected as part of pull (which I'm not aware of such situations), I would expect option to declare that its required only for commands such `up` but ignored for `pull`.
for example `required: "runtime"`

### Steps To Reproduce

1. add to docker-compose.yml

```
env_file:
  - path: .myservice.env
    required: true
 ```
 
 2. run `docker compose pull`
 3. command fails due to missing file. - **undesired behavior**
 4. edit to `required: false`
 5. command succeeded.
 6. but now `docker compose up` ignores missing file as well 

### Compose Version

```Text
Docker Compose version v2.25.0
```

ndeloof:
Docker Compose indeed fully parse and resolve your compose file before it execute any command, and doing so will detect missing env file whenever this isn't used by the `pull` command.
This is comparable to https://github.com/docker/compose/issues/10993

We miss a more flexible mechanism to only load and validate required parts of a compose model depending on command being ran.

PR #12904, opened 2025-06-04
pull does not require env_file being resolved

ndeloof:
**What I did**

**Related issue**
fix https://github.com/docker/compose/issues/12081


glours (approved):
LGTM

PR #12906, opened 2025-06-05
(refactoring) introduce monitor to manage containers events and application termination

ndeloof:
**What I did**
refactoring:
- introduced `monitor` to dispatch container events and manage application termination
- re-implemented cascade stop as a separate component, relying on event dispatch
- re-implemented `watchcontainer` relying on event dispatch, removed from `start` and `log`
- `printer` only focus on rendering application status, not trying to detect termination - uses explicit termination
- removed useless pre-allocation of `presenter` (color) in `printer`. Will be set on first use.

improvement (2nd commit):
collecting start event, capture container's StartTime. This is used to collect logs using log API (we are too late to attach) and retrieve the very first logs. Tested using:
```go
func main() {
        fmt.Println("Hello, World!")
        time.Sleep(10*time.Second)
}
```

**Related issue**

image


glours (approved):
LGTM

PR #12908, opened 2025-06-06
move `run` logic inside backend

ndeloof:
**What I did**
refactored `run` so that all the container management logic is owned by backend
added support for `use_api_socket` as a syntaxic sugar, adding bind/config resources to compose model, the same way it is implemented as experimental by https://github.com/docker/cli/pull/5858 (waiting for actual support by Docker Engine)

**Related issue**

image

### DIFF
--- a/pkg/compose/apiSocket.go
+++ b/pkg/compose/apiSocket.go
@@ -1,0 +1,90 @@
+/*
+   Copyright 2020 Docker Compose CLI authors
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


max-wittig:
Apparently now `docker compose run` does no longer consume environment variable so this broke our setup.

We normally inject proxy variables into a build process and docker compose used to automatically used those. Now they are not even consumed, if explicitly supplied via `docker compose run -e`

max-wittig:
Downgrading to `2.37.1` made it work again.

ndeloof:
@max-wittig I added an e2e test to check `docker compose run -e` works as expected and didn't detected any regression. Can you please give more details ?
(https://github.com/docker/compose/pull/12967)

max-wittig:
Thank you for your quick response!

Maybe its relevant that we need to set `DOCKER_BUILDKIT: 0` and that we use this command to maintain Sentry? 

See: https://github.com/getsentry/self-hosted/blob/494051b8df0c6e18badb701daf85b3818c170888/install/set-up-and-migrate-database.sh#L14

`docker compose --profile feature-complete run --rm web upgrade --noinput --create-kafka-topics`

And maybe its actually another change in `2.37.2` that causes this, but I'm not sure. I was really thinking that this could be the cause.

PR #12929, opened 2025-06-10
track only images for services to be built

### LINKED ISSUE #12928: [BUG] build failing to check already built dependencies
### Description

I'm running the code that's on `main` . Because of #12925 it now [produces bake targets for all services](https://github.com/docker/compose/commit/213c03f99afc516b76bdf46f60769d3fa7015359)

This now causes all depedencies to be in the expected built images and causes an error.

### Steps To Reproduce

```yaml
services:
  main:
    build:
      dockerfile_inline: |
        FROM alpine
    entrypoint: ["echo", "Hello from main"]
    depends_on:
      - dep1

  dep1:
    build:
      dockerfile_inline: |
        FROM alpine
    entrypoint: ["echo", "Hello from dep1"]
```
first run `docker compose dep1` 
then run  `docker compose main` - this will build the image but exit with an error `build result not found in Bake metadata for service dep1`

### Compose Version

```Text
Commit 213c03f
```

### Docker Environment

```Text

```

### Anything else?

_No response_

israelglar:
**What I did**
Added a conditional check to verify if a service exists in `serviceToBeBuild` before adding its image to the `expectedImages` map. This prevents unintended services from being included.

**Related issue**

fixes #12928 

### DIFF
--- a/pkg/compose/build_bake.go
+++ b/pkg/compose/build_bake.go
@@ -174,7 +174,10 @@ func (s *composeService) doBuildBake(ctx context.Context, project *types.Project
 		}
 
 		image := api.GetImageNameOrDefault(service, project.Name)
-		expectedImages[serviceName] = image
+
+		if _, ok := serviceToBeBuild[serviceName]; ok {
+			expectedImages[serviceName] = image
+		}
 
 		entitlements := build.Entitlements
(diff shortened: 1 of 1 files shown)


ndeloof:
Already fixed in another PR (my bad, I didn't reviewed yours before)

PR #12930, opened 2025-06-11
only look for required image in bake metadata

ndeloof:
**What I did**

**Related issue**
closes https://github.com/docker/compose/issues/12928


glours (commented):
it seams `image` should be added to the tags list 😅

ndeloof:
yes, I removed the wrong line 🥸

PR #12931, opened 2025-06-11
don't create metadatafile, just generate a random name

ndeloof:
**What I did**
only generate a random file name for bake metadata, don't actually create/open file
(this prevent conflict with atomicwritter)

**Related issue**
issue detected on Docker Desktop Windows


glours (approved):
LGTM

PR #12933, opened 2025-06-11
Fix the generated manifest for compose artifacts

jcarter3:
**What I did**
The generated manifests from `docker compose alpha publish` include setting the config as a layer -> [example](https://oci.dag.dev/?image=registry-1-stage.docker.io%2Fhubregistrystagetest%2Fflask-redis-compose-stack%3A1.0.0)

This PR updates this so that we will still push the empty descriptor blob (so the manifest push succeeds), but we don't set it as a layer.

**Related issue**


ndeloof:
for legal reasons 🤪 all contributors must sign-off commits so we can accept them in this repository
please amend your PR
thanks

PR #12942, opened 2025-06-15
replace "pipes.quote" with equivalent "shlex.quote"

asareynolds:
**What I did**
Replaced `pipes.quote` with equivalent `shlex.quote` to maintain functionality in Python 3.13 after [removal](https://peps.python.org/pep-0594) of `pipes`.

See:
- https://github.com/cylc/cylc-flow/issues/2864
- https://stackoverflow.com/questions/26790916/python-3-backward-compatability-shlex-quote-vs-pipes-quote
- https://goldeneye2.videolan.org/Astrom/vlc/-/blob/master/extras/tools/ninja-1.11.1-replace-pipes-quote-with-shlex-quote.patch?ref_type=heads
- https://github.com/Tanguy-Boisset/bloodhound-automation/issues/20

### DIFF
--- a/compose/cli/main.py
+++ b/compose/cli/main.py
@@ -2,7 +2,7 @@ import contextlib
 import functools
 import json
 import logging
-import pipes
+import shlex
 import re
 import subprocess
 import sys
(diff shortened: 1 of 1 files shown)


ndeloof:
Docker Compose v1 is EoL and unmaintained. Please stop relying on this python codebase

Issue #12940, opened 2025-06-14
[BUG] `docker compose images --format json` is missing the` ContainerName` field compared to 2.36.x


felixfontein:
### Description

2.36.x's `docker compose images --format json` contained the `ContainerName` field; since 2.37.1 it does not contain this field anymore, but it does contain `Platform` which wasn't there before.

(I apparently mislooked in https://github.com/docker/compose/issues/12916#issuecomment-2952954425; while #12917 restored the list output, it did not restore the `ContainerName` field.)

The `ContainerName` information has been removed in #12856 (which caused the breaking change for 2.37.0 that got reverted in #12917).

While the field was actively removed in #12856, there has been no announcement in the changelog of this breaking change, so I'm reporting this as a bug. 

### Steps To Reproduce

Run `docker compose images --format json` with 2.36.x and 2.37.1 or `main`.

### Compose Version

```Text
2.37.1 / `main` branch
```

### Docker Environment

```Text

```

PR #12917, opened 2025-06-07
Revert docker compose images JSON output to array format

x0rw:
In case #12916 was decided on.
Changed  `docker compose images --format json` behaviour from json map to a json list.

```
{
    "lab_postgres": {
        "ID": "sha256:XXXXXXXX",
        "Repository": "postgres",
        ...
    "redis_ins": {
        "ID": "sha256:YYYYYYY",
        "Repository": "redis",
        ...
```
Back to:
```
{
{"ID":"sha256:XXXXXX","Repository":"research-labs-redis-ins",...},
{"ID":"sha256:YYYYY","Repository":"research-labs-test-postgres",...}
```

**Related issue**
Resolves #12916.

PR #12943, opened 2025-06-16
restore ContainerName in images --json

ndeloof:
**What I did**
restore ContainerName which is missing in json output since https://github.com/docker/compose/pull/12917

**Related issue**
fixes https://github.com/docker/compose/issues/12940


glours (approved):

felixfontein:
@ndeloof @glours thanks for fixing this!

Issue #12966, opened 2025-06-23
[BUG] docker compose run triggers an unwanted rebuild (regression in v2.37.2)


Tycale:
## Affected versions

| Component        | Works (v2.37.1) | Fails (v2.37.2) |
|------------------|-----------------|-----------------|
| Docker Compose   | ✅              | ❌              |
---

Since Docker Compose v2.37.2, docker compose run seems to always rebuilds a service that has both an image: tag and a build: section, even when the tagged image is already present locally. Earlier versions (≤ v2.37.1) correctly skipped the build. This regression breaks CI pipelines that pre-build images with secret build args and expect the run phase to start the container without rebuilding.

## Steps to reproduce

```bash
# 1 – Build and load the image locally
docker buildx build \
  --load \
  -f src/path/Dockerfile.api-core \
  --target base \
  -t X-api .

# 2 – docker-compose.yml
cat <<'YML' > docker-compose.yml
services:
  core-test:
    image: X-api          # Matches the tag built above
    build:
      context: .
      dockerfile: src/path/Dockerfile.api-core
      target: base
YML

# 3 – Run the service
docker compose run --rm core-test echo OK
```

## Expected behavior (v2.37.1)

docker compose run detects the local image X-api and starts the container without any build.

## Actual behavior (v2.37.2)

Compose always launches a new build for core-test.

## Why is it an issue ?
In my CI, it fails because secret build-args (e.g. Y_SECRET) are only available during the initial build step.

### Compose Version

```Text
/
```

### Docker Environment

```Text
## Env
Docker version

/usr/bin/docker version
  Client:
   Version:           25.0.4
   API version:       1.44
   Go version:        go1.21.8
   Git commit:        1a576c5
   Built:             Wed Mar  6 16:32:02 2024
   OS/Arch:           linux/amd64
   Context:           default
  Server: Docker Engine - Community
   Engine:
    Version:          28.2.2
    API version:      1.50 (minimum version 1.24)
    Go version:       go1.24.3
    Git commit:       45873be
    Built:            Fri May 30 11:52:20 2025
    OS/Arch:          linux/amd64
    Experimental:     false
   containerd:
    Version:          v1.7.27
    GitCommit:        05044ec0a9a75232cad458027ca83437aae3f4da
   runc:
    Version:          1.2.6
    GitCommit:        v1.2.6-0-ge89a[29](https://github.com/PII/PII/actions/runs/15829363608/job/44617836286?pr=26309#step:7:31)9
   docker-init:
[...]

PR #12968, opened 2025-06-23
don't create from run command during dependencies creation process

### LINKED ISSUE #12965: [BUG] compose run --rm does not remove the container
### Description

`docker compose run --rm <service>` doesn't remove the container any more.

### Steps To Reproduce

With this yaml file

```yaml
services:
  web:
    image: nginx
```

Run `docker compose run --rm web`, wait for the container to start and then ctrl+c

Then run `docker ps -a` and you should see something like

```console
; docker ps -a
CONTAINER ID   IMAGE                    COMMAND                  CREATED         STATUS       PORTS                                                                                          NAMES
66066872be26   nginx                    "/docker-entrypoint.…"   4 seconds ago   Created                                                                                                     test-web-1
```

With 3.27.1 the container is removed as it should.

### Compose Version

```Text
3.27.2
```

### Docker Environment

```Text
DD on a mac
```

### Anything else?

_No response_

glours:
**What I did**
Remove services passed as parameters of the `docker compose run` command from being created during the dependencies starting process

**Related issue**
Fix #12965 #12966 

image


pkg/e2e/compose_run_test.go
```diff
@@ -196,4 +196,18 @@ func TestLocalComposeRun(t *testing.T) {
...
+		projectName := "run-test"
+		res := c.RunDockerComposeCmd(t, "--project-name", projectName, "-f", "./fixtures/ps-test/compose.yaml", "run", "--rm", "-d", "nginx")
+		res.Assert(t, icmd.Success)
+
+		res = c.RunDockerCmd(t, "ps", "--quiet", "--filter", "name=run-test-nginx")
+		containerID := strings.TrimSpace(res.Stdout())
+
+		res = c.RunDockerCmd(t, "stop", containerID)
+		res.Assert(t, icmd.Success)
+		res = c.RunDockerCmd(t, "ps", "--all", "--filter", "name=run-test-nginx")
```
ndeloof: suggestion:  use `--format '{{.Names}}'` in command so output only shows container(s) name

PR #12971, opened 2025-06-24
test with Docker Engine nightly builds

ndeloof:
**What I did**

**Related issue**


pkg/compose/compose.go
```diff
@@ -314,6 +314,8 @@ var runtimeVersion runtimeVersionCache
 func (s *composeService) RuntimeVersion(ctx context.Context) (string, error) {
 	runtimeVersion.once.Do(func() {
 		version, err := s.dockerCli.Client().ServerVersion(ctx)
+		fmt.Println("APIVersion == " + version.APIVersion)
+		fmt.Println("Version == " + version.Version)
```
vvoland, line 318: Looks like for some reason it's using a different Docker socket?
```
        APIVersion == 1.48
        Version == 28.0.4
```

The Docker installed by the setup action seems correct:
image

PR #12983, opened 2025-06-26
exclude provider services from the list of dependencies that Compose should wait for

### LINKED ISSUE #12953: [BUG] docker compose up with --wait and a provider exits as soon as provider is done
### Description

I would expect all the services to be up and running before exiting.

### Steps To Reproduce

Use the following compose file:
```yaml
services:
  docker-model-runner:
    provider:
      type: model
      options:
        model: ai/smollm2:135M-Q4_K_M

  backend:
    image: nginx
    depends_on:
      - docker-model-runner
```
and then `docker compose up --wait`: the compose command will exit almost immediately, before the actual backend service is up and running. 

### Compose Version

```Text

```

### Docker Environment

```Text

```

### Anything else?

_No response_

glours:
**What I did**
Explicitly exclude provider services from the list of dependencies to check when `--wait` is used

**Related issue**
Fix #12953 

image

### DIFF
--- a/pkg/compose/convergence.go
+++ b/pkg/compose/convergence.go
@@ -563,6 +563,9 @@ func shouldWaitForDependency(serviceName string, dependencyConfig types.ServiceD
 	} else if service.GetScale() == 0 {
 		// don't wait for the dependency which configured to have 0 containers running
 		return false, nil
+	} else if service.Provider != nil {
+		// don't wait for provider services
+		return false, nil
 	}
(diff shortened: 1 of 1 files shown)


glours:
ping @silvin-lubecki if you want to test it

silvin-lubecki (approved):
It works, thanks a lot @glours 🙏 LGTM ✅

ndeloof (approved):

PR #12995, opened 2025-06-30
mount /var/run/docker.sock for --use-api-socket

ndeloof:
**What I did**

`s.dockerCli.DockerEndpoint().Host` returns the engine URL as seen by client, which is only relevant an Unix/MacOS with a local docker engine. Waiting for engine to expose this information in some API, or implement this feature on its own, hard-coding `/var/run/docker.sock` offer a more reliable path for engine socket to be mounted in container, while not 100% guaranteed to succeed.

**Related issue**
see https://github.com/docker/cli/pull/6157


fiam (approved):

mikesir87:
Built the modified Compose and tested on a Windows machine (arm64 on WSL) and confirmed the socket mount is working and that the credentials are injected into the container as expected.

PR #12997, opened 2025-06-30
remove publish limitation on bind mount

glours:
list all bind mounts and ask user validation before publishing

**What I did**
Remove the bind mount restriction on `docker compose publish` command
List all the bind mounts and ask for confirmation before publishing the OCI artifacts

⚠️  Important note, only the bind mount declarations will be present in the OCI artifact, no data/content will be added

**Related issue**

image


pkg/compose/publish.go
```diff
@@ -236,8 +236,20 @@ func (s *composeService) preChecks(project *types.Project, options api.PublishOp
 	if ok, err := s.checkOnlyBuildSection(project); !ok || err != nil {
 		return false, err
 	}
-	if ok, err := s.checkForBindMount(project); !ok || err != nil {
-		return false, err
+	bindMounts := s.checkForBindMount(project)
+	if len(bindMounts) > 0 {
+		fmt.Println("you are about to publish bind mounts declaration within your OCI artifact.\n" +
+			"only the bind mount declarations will be added to the OCI artifact\n" +
+			"please double check that you are not mounting potential sensitive directories or data")
```
ndeloof:
```suggestion
		fmt.Println("you are about to publish bind mounts declaration within your OCI artifact.\n" +
			"only the bind mount declarations will be added to the OCI artifact (not content)\n" +
			"please double check that you are not mounting potential user's sensitive directories or data")
```

PR #12999, opened 2025-06-30
fix provider concurrent environment map accesses

### LINKED ISSUE #12937: concurrent map write by provider
### Description

happens when a service depends on 2 or more services with providers
https://github.com/docker/compose/blob/ea48480d809a03e4abe0a81f69be66bf7b85bcd9/pkg/compose/plugins.go#L77

### Steps To Reproduce

_No response_

### Compose Version

```Text

```

### Docker Environment

```Text

```

### Anything else?

_No response_

glours:
**What I did**
Add mutex to fix concurrent map access to environment configuration when a service depends on multiple provider services

**Related issue**
fix #12937 

image

### DIFF
--- a/pkg/compose/plugins.go
+++ b/pkg/compose/plugins.go
@@ -27,16 +27,18 @@ import (
 	"os/exec"
 	"path/filepath"
 	"strings"
+	"sync"
 
-	"github.com/compose-spec/compose-go/v2/types"
-	"github.com/docker/cli/cli-plugins/manager"
-	"github.com/docker/cli/cli-plugins/socket"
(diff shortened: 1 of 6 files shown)


pkg/compose/plugins.go
```diff
@@ -70,6 +74,8 @@ func (s *composeService) runPlugin(ctx context.Context, project *types.Project,
 		return err
 	}
 
+	mux.Lock()
+	defer mux.Unlock()
```
ndeloof, line 78: We don't want providers to run sequentially, only the project mutation must be guarded by a mutex.

ndeloof (approved):

Issue #12918, opened 2025-06-08
Allow compose to show container shutdown logs. With an alternative UI.


robinhickmann:
### Description

Containers printing logs on shutdown don't get displayed.

## Example code

```go
package main

import (
    "fmt"
    "os"
    "os/signal"
    "syscall"
)

func main() {
    sigChan := make(chan os.Signal, 1)
    signal.Notify(sigChan, syscall.SIGINT, syscall.SIGTERM)

    fmt.Println("Hello, World!")

    <-sigChan

    fmt.Println("Goodbye, World!")
}
```

## Example Dockerfile

```dockerfile
FROM golang:alpine

WORKDIR /app

COPY . .

RUN go build -o hello .

CMD ["./hello"]
```

## Example compose.yml

```yaml
services:
  hello:
    build: .
```

1. Run `docker compose up --build`.
 
2. Press CTRL + C to gracefully shutdown all the services.

3. Notice how you won't see the "Goodbye, World!" log but if you run `docker compose logs hello`, you'll see the log.

### Compose Version

```Text
Docker Compose version v2.36.2
```

### Docker Environment

```Text
Client: Docker Engine - Community
 Version:    28.2.2
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.24.0
    Path:     /usr/libexec/docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.36.2
    Path:     /usr/libexec/docker/cli-plugins/docker-compose

Server:
 Containers: 25
  Running: 0
  Paused: 0
  Stopped: 25
 Images: 73
 Server Version: 28.2.2
 Storage Driver: overlay2
  Backing Filesystem: btrfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local splunk syslog
 CDI spec directories:
  /etc/cdi
  /var/run/cdi
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 05044ec0a9a75232cad458027ca83437aae3f4da
[...]

PR #13010, opened 2025-07-02
keep containers attached on stop to capture termination logs

ndeloof:
**What I did**

https://asciinema.org/a/dUq4g3wYLXsEQMn6okzvYGcIt

**Related issue**
fixes https://github.com/docker/compose/issues/12918

### DIFF
--- a/cmd/formatter/logs.go
+++ b/cmd/formatter/logs.go
@@ -73,9 +73,12 @@ func (l *logConsumer) register(name string) *presenter {
 	} else {
 		cf := monochrome
 		if l.color {
-			if name == api.WatchLogger {
+			switch name {
+			case "":
+				cf = monochrome
+			case api.WatchLogger:
 				cf = makeColorFunc("92")
-			} else {
+			default:
 				cf = nextColor()
 			}
 		}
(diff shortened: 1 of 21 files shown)


cmd/formatter/logs.go
```diff
@@ -167,15 +170,15 @@ type logDecorator struct {
...
-	l.decorated.Log(containerName, message)
+	l.decorated.Err(containerName, message)
 	l.After()
 }
 
-func (l logDecorator) Err(containerName, message string) {
+func (l logDecorator) Log(containerName, message string) {
 	l.Before()
-	l.decorated.Err(containerName, message)
+	l.decorated.Log(containerName, message)
```
glours: 🤔

gomock_reflect_1911265598/prog.go
glours, line 1: Not sure we should have this file commited

glours (approved):
Nicely done!! 
Few nit comments 😉

cmd/formatter/logs.go
ndeloof: seems I used goland shortcut to move func up and down in file by mistake and didn't even noticed ;)

PR #13021, opened 2025-07-06
Add --networks flag to config command

jarqvi:
**What I did**
Add a new flag to config command for print the network names, one per line.

### DIFF
--- a/cmd/compose/config.go
+++ b/cmd/compose/config.go
@@ -50,7 +50,7 @@ type configOptions struct {
 	noResolveEnv        bool
 	services            bool
 	volumes             bool
-	networks			bool
+	networks            bool
 	profiles            bool
 	images              bool
 	hash                string


ndeloof (approved):

ndeloof:
I'm ok we introduce more config "subcommands" (as flags 🤪) but I'd like this to be aligned with docker cli `docker network ls`: render as a table, or optional format. We can look into this as a follow-up PR

jarqvi:
> I'm ok we introduce more config "subcommands" (as flags 🤪) but I'd like this to be aligned with docker cli `docker network ls`: render as a table, or optional format. We can look into this as a follow-up PR

I agree, and I'm ready for any follow-up PR:))

jarqvi:
I fixed the lint errors
@ndeloof

glours (approved):

PR #13022, opened 2025-07-07
Add --models flag to config command

jarqvi:
**What I did**
Add a new flag to config command for print the model names, one per line.

### DIFF
--- a/cmd/compose/config.go
+++ b/cmd/compose/config.go
@@ -393,8 +393,10 @@ func runModels(ctx context.Context, dockerCli command.Cli, opts configOptions) e
 	if err != nil {
 		return err
 	}
-	for n := range project.Models {
-		_, _ = fmt.Fprintln(dockerCli.Out(), n)
+	for _, model := range project.Models {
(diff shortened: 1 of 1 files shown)


cmd/compose/config.go
```diff
@@ -383,6 +388,17 @@ func runNetworks(ctx context.Context, dockerCli command.Cli, opts configOptions)
 	return nil
 }
 
+func runModels(ctx context.Context, dockerCli command.Cli, opts configOptions) error {
+	project, err := opts.ToProject(ctx, dockerCli, nil, cli.WithoutEnvironmentResolution)
+	if err != nil {
+		return err
+	}
+	for n := range project.Models {
+		_, _ = fmt.Fprintln(dockerCli.Out(), n)
```
ndeloof: this will dump model name used in compose file, that can't be used in any command. I guess dumping `Model` attribute (OCI artifact ID) would make more sense

glours (approved):
LGTM

Issue #13020, opened 2025-07-04
[BUG] Build secret files not mounted in /run/secrets when using docker compose build with profiles


kmfarley11:
### Description

If I set up a Dockerfile for build secrets, docker build --secret ... works just fine [per the official docs](https://docs.docker.com/build/building/secrets/).

If I set up a docker compose project [per the official docs](https://docs.docker.com/reference/compose-file/secrets/) ([this one too](https://docs.docker.com/compose/how-tos/use-secrets/#build-secrets)) secrets do still seem to work just fine.

However, **if I add a profile to my service that has build secrets, and try to `up` or `build` the service in isolation, I run into issues**!
Further, if i do a direct `up --build` on the command behind the profile, that still works too...

Note: this is highly related to other issues such as https://github.com/docker/compose/issues/9822 & https://github.com/docker/compose/issues/13019 (sorry for the confusion on that latter one). But it describes a different issue with profiles and `build` vs. `up` behavior on the latest docker engine.

### Steps To Reproduce

1. Create a simple local test file: `echo "SECRET TEST" > test-secret.txt`
2. Create a Dockerfile with this content
```dockerfile
#Dockerfile
FROM debian:stable-slim
RUN --mount=type=secret,id=test-secret ls -la /run/secrets/; cp /run/secrets/test-secret /tmp

CMD ["cat", "/tmp/test-secret"]
```

3. Create a docker-compose file with this content:
```yaml
#docker-compose.yaml
secrets:
    test-secret:
        file: test-secret.txt

services:
    secret-build-test:
        profiles: ["test"]
        build:
            context: .
            dockerfile: Dockerfile
            secrets:
                - test-secret
```

4. Verify that a docker compose build with profiles activated works as expected 
```bash
docker builder prune -fa
docker compose --progress=plain --profile test build secret-build-test
```
should output something like the following:
```
...
#6 [stage-0 2/2] RUN --mount=type=secret,id=test-secret ls -la /run/secrets/; cp /run/secrets/test-secret /tmp
#6 0.156 total 12
#6 0.156 drwxr-xr-x 2 root root 4096 Jul  4 15:42 .
#6 0.156 drwxr-xr-x 1 root root 4096 Jul  4 15:42 ..
#6 0.156 -r-------- 1 root root   12 Jul  4 15:42 test-secret
#6 DONE 0.2s
...
```
5. Demonstrate that a direct docker compose build fails!
```bash
docker builder prune -fa
[...]

PR #13023, opened 2025-07-07
build: fix service with profile missing secrets

ndeloof:
**What I did**

As build resolves enabled service after project has been loaded, we must not drop "unused resources" as we actually don't know they are.

**Related issue**
fixes https://github.com/docker/compose/issues/13020


glours (approved):
LGTM

PR #13029, opened 2025-07-08
introduce development mode

ndeloof:
**What I did**
`docker compose up --dev` enable development mode, which allows to publish additional ports (typically: a remote debugger, or administration UI) and override the service command (to switch to `yarn dev`)

While less powerful, this offers an alternative to a sibling compose-dev.yaml file and need to specify multiple compose files on command line

example:
```yaml
services:
  app:
     build: .
     develop:
        watch:
           - path: src
             target: /app
        command: yarn dev --port 4001
        ports:
           - 4001:4001
```

**Related issue**

### DIFF
--- a/cmd/compose/create.go
+++ b/cmd/compose/create.go
@@ -49,6 +49,7 @@ type createOptions struct {
 	quietPull     bool
 	scale         []string
 	AssumeYes     bool
+	Development   bool
 }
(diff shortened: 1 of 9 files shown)


zoriya:
Is there a reason this was abandoned?

PR #13036, opened 2025-07-11
feat(os): add FreeBSD support

atagtm:
**What I did**
Added FreeBSD to the "unix list"

**Related issue**
None (?) This way it is possible to build it on FreeBSD without any issues.

cute animal


glours (approved):

ndeloof:
I wonder we could add Open|FreeBSD nodes to our CI

PR #13038, opened 2025-07-11
Add a space character to separate the timestamp from the log message

xduugu:
**What I did**

**Related issue**


ndeloof:
LGMT as this is better aligned with `docker logs --timestamps`

Issue #13048, opened 2025-07-15
[BUG] Labels no longer created on build


kevin-vitro:
### Description

I was using these labels that docker compose automatically generated at least in compose v2.35.1
            "com.docker.compose.project": "deploy",
            "com.docker.compose.service": "deploy",
            "com.docker.compose.version": "2.35.1",

As of 2.38.2, these labels no longer show up in Config.Labels section of our docker image. This negatively affects us because vs code docker management tools group based on the "com.docker.compose.project" so that is how I initially noticed this difference. 

### Steps To Reproduce

Using compose v2.35.1: 
1. Build a container using `docker compose build`
2. Observe the labels using `docker inspect {built_image_name}` 
     1. Specifically look at Config.Labels section
     2. Confirm the existence of:
          1. com.docker.compose.project
          2. com.docker.compose.service
          3. com.docker.compose.version

Repeat using compose v2.38.2 (I haven't tested on versions between these)

#### Minimal Reproducible Example: 
Using the following files: 
`docker-compose.yml`
```
services:
  test:
    build: .
    ports:
      - "8000:5000"
```

`Dockerfile`
```
FROM python:3.9-slim
CMD ["/bin/bash"]
```

run these commands: 
```
docker compose -f docker-compose.yml build
docker inspect nbs-test
```

The name of the docker image may be different from nbs-test. nbs was the name of the parent folder where I was running this test. 

### Compose Version

```Text
2.38.2
```

### Docker Environment

```Text
Client: Docker Engine - Community
 Version:    28.3.2
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.25.0
    Path:     /usr/libexec/docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.35.1
    Path:     /usr/libexec/docker/cli-plugins/docker-compose
  model: Docker Model Runner (EXPERIMENTAL) (Docker Inc.)
    Version:  v0.1.33
    Path:     /usr/libexec/docker/cli-plugins/docker-model
WARNING: Plugin "/usr/libexec/docker/cli-plugins/docker-compose_2_38_2" is not valid: plugin candidate "compose_2_38_2" did not match "^[a-z][a-z0-9]*$" 

Server:
 Containers: 35
  Running: 21
  Paused: 0
  Stopped: 14
 Images: 376
 Server Version: 28.3.2-1
 Storage Driver: overlay2
[...]

PR #13049, opened 2025-07-15
add default compose labels to images built from bake

glours:
**What I did**
Add default Compose labels to images build from bake via Compose

**Related issue**
#13048 

image

### DIFF
--- a/pkg/compose/build_bake.go
+++ b/pkg/compose/build_bake.go
@@ -164,6 +164,7 @@ func (s *composeService) doBuildBake(ctx context.Context, project *types.Project
 			continue
 		}
 		build := *service.Build
+		labels := getImageBuildLabels(project, service)
 
 		args := types.Mapping{}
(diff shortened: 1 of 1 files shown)


kevin-vitro:
For what it's worth, I tested the [compose-linux-amd64](https://github.com/docker/compose/actions/runs/16304428445/artifacts/3540038499) artifact and it seems to show the labels I was expecting to see.  The "com.docker.compose.version" label is blank, but I'm guessing that is to be expected since this is not a tag-labeled version.  

I'm not sure if this matters, but the "Config" section of the docker inspect for compose v2.35.1 was higher on the list (between Author and Architecture). It shows up at the end of the list now (after "Metadata") but it does have the Labels section. Just wanted to point that out because I did a diff and noticed.  

Thank you for looking into the issue! It looks like a nice fix even though I don't know enough go to provide helpful feedback :)

ndeloof (approved):

PR #13051, opened 2025-07-16
abstract model-cli commands execution with a model (pseudo) API

ndeloof:
**What I did**
created a pseudo-API for model management, waiting for https://github.com/docker/model-cli/issues/113 so we can actually use `modelAPI.PullImage` in compose pull

**Related issue**


glours (approved):
LGTM and tests ok 👍

PR #13052, opened 2025-07-16
pkg/watch: remove unused IsWindowsShortReadError

thaJeztah:
This function was added in b3615d64e2ce7c5496e71fc51e965a6368da2ba2 but appears to be unused.

**What I did**

**Related issue**

### DIFF
--- a/pkg/watch/notify.go
+++ b/pkg/watch/notify.go
@@ -17,15 +17,11 @@
 package watch
 
 import (
-	"errors"
 	"expvar"
 	"fmt"
 	"os"
 	"path/filepath"
-	"runtime"
 	"strconv"
-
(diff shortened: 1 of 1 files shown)


pkg/watch/notify.go
```diff
@@ -17,15 +17,11 @@
...
 import (
-	"errors"
 	"expvar"
 	"fmt"
 	"os"
 	"path/filepath"
-	"runtime"
 	"strconv"
-
-	"github.com/tilt-dev/fsnotify"
```
thaJeztah, line 28: I think I started looking at this to see if we still needed this fork; some of the reasons it was needed for were no longer there (upstream merged some fixes), but I don't know where I left off 😂

Issue #13044, opened 2025-07-14
[BUG] "docker compose up --watch" no longer displays build log when 'rebuild' action triggers


coredumperror:
### Description

I've been using `docker compose up --watch` for a while now, and I love it. But ever since some time last week (maybe earlier), the build log for my image builds no longer displays when a `rebuild` action triggers. I just get this:

```
                    ⦿ Rebuilding service(s) ["multitenant" "multitenant_www"] after changes were detected...
[+] Building 0.0s (0/0)
                    ⦿ service(s) ["multitenant" "multitenant_www"] successfully built
```

### Steps To Reproduce

With this docker-compose.yml (truncated significantly):

``` yml
services:
  multitenant:
    image: image_name
    ...
    build:
      context: .
      tags:
        - multitenant:6.8.3
        - multitenant:latest
    develop:
      watch:
        # Sync the working directory with the /multitenant directory in the container.
        - action: sync
          path: .
          target: /multitenant
          ignore:
            # Ignore all migrations directories, because those are controlled by the volume mounts below.
            - "**/migrations"
        ...
        - action: rebuild
          path: ./Dockerfile

  multitenant_www:
    image: image_name
    ...
    build:
      context: .
      tags:
        - multitenant:6.8.3
        - multitenant:latest
    develop:
      watch:
        # Sync the working directory with the /multitenant directory in the container.
        - action: sync
          path: .
          target: /multitenant
          ignore:
            # Ignore all migrations directories, because those are controlled by the volume mounts below.
            - "**/migrations"
        ...
        - action: rebuild
          path: ./Dockerfile
```

I get the following output when I update the `Dockerfile`:

```
                    ⦿ Rebuilding service(s) ["multitenant" "multitenant_www"] after changes were detected...
[+] Building 0.0s (0/0)
                    ⦿ service(s) ["multitenant" "multitenant_www"] successfully built

```

I expect that `Building 0.0s (0/0)` to instead be an entire build log for the image.

### Compose Version

```Text
$ docker compose version
Docker Compose version v2.38.2-desktop.1



$ docker-compose version
Docker Compose version v2.38.1-desktop.1
```

Huh... weird that one is a slightly different version.

### Docker Environment

```Text
$ docker info
Client:
 Version:    28.3.0
 Context:    desktop-linux
 Debug Mode: false
 Plugins:
  ai: Docker AI Agent - Ask Gordon (Docker Inc.)
    Version:  v1.8.0
[...]

PR #13059, opened 2025-07-17
show build progress during watch rebuild

ndeloof:
**What I did**
During a watch rebuild, show build logs (plain mode) mixed with running services logs
Introduced `--quiet-build` to support the previous behavior for watch to hide build logs
(makes me wonder we should deprecate `watch` command)

https://asciinema.org/a/2LfcAifSGq8FWmfUnbShuNn9g

**Related issue**
fixes https://github.com/docker/compose/issues/13044

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -19,6 +19,7 @@ package api
 import (
 	"context"
 	"fmt"
+	"io"
 	"slices"
 	"strings"
 	"time"
@@ -176,6 +177,8 @@ type BuildOptions struct {
 	Provenance string
 	// SBOM generate a SBOM attestation
 	SBOM string
+	// Out is the stream to write build progress
+	Out io.Writer
 }
 
 // Apply mutates project according to build options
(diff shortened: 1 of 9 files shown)


glours:
IMHO we definitely should add a flag to activate this behaviour, this mix of 2 kinds of logs is very confusing and I think this should be display only on demand

cmd/compose/create_test.go
```diff
@@ -19,6 +19,7 @@ package compose
 import (
 	"context"
 	"fmt"
+	"os"
```
glours: 🤔 Why do we have a new import if there isn't nothing else added?

pkg/api/api.go
```diff
@@ -172,6 +173,8 @@ type BuildOptions struct {
 	Check bool
 	// Provenance
 	Provenance bool
+	// Ou is the stream to write build progress
```
glours:
```suggestion
	// Out is the stream to write build progress
```

glours (commented):
I'm waiting for the version of this PR with the flag to merge it 😉

glours (approved):
LGTM

PR #13065, opened 2025-07-17
warn user COMPOSE_BAKE=false is deprecated

ndeloof:
**What I did**

**Related issue**


glours (approved):
LGTM

PR #13067, opened 2025-07-17
introduce build.provenance and sbom support

ndeloof:
**What I did**

introduce build `provenance` and `sbom` support

**Related issue**
this re-implements https://github.com/docker/compose/pull/12853, so provenance/sbom attestations are only generated by an explicit `docker compose build` command, but non for inner-loop builds. Also, provenance is **not** enabled by default (same as `docker build`)

requires https://github.com/compose-spec/compose-go/pull/809


glours (approved):
LGTM

PR #13086, opened 2025-07-23
bump buildx to v0.26.1

glours:
**What I did**
bump `buildx` to v0.26.1 and `buildkit` with the same version as the one defined in `buildx` release

**Related issue**
N/A

image


crazy-max (approved):
LGTM

PR #13094, opened 2025-07-24
add info about models usage to OpenTelemetry spans

glours:
**What I did**
Add `models` data to OpenTelemetry spans, so we could monitor usage of this new feature

**Related issue**

image


internal/tracing/attributes.go
```diff
@@ -110,6 +112,7 @@ func ServiceOptions(service types.ServiceConfig) SpanOptions {
 		attribute.String("service.name", service.Name),
 		attribute.String("service.image", service.Image),
 		attribute.StringSlice("service.networks", keys(service.Networks)),
+		attribute.StringSlice("service.models", keys(service.Models)),
```
glours, line 115: I wonder if you should collect names of the `models` configuration or the real model names defined with the `model` attribute

ndeloof: all names we collect are abstract name, used inside compose.yaml as references, and do not reflect the actual resource. I wonder any of those is of any use. Maybe we should better collect a **count** of those

ndeloof (approved):

Issue #13099, opened 2025-07-25
[BUG] Updating `2.38.2` to `2.39.x` changes `docker compose build` output progress to plain


ToshY:
### Description

When updating docker compose from `v2.38.2` to `v2.39.0`, the output of `docker compose build` / `docker compose up --build` in the terminal has changed to `plain`.

I am running these commands in the terminal (WSL2), so I think this is a bug released in `2.39.0` caused by https://github.com/docker/compose/pull/13074.

When explicitely trying `--progress=tty` in `v2.38.2` it works and shows the build output.

```
$ docker compose --progress=tty up -d --build
[+] Building 1.1s (8/8) FINISHED
 => [internal] load local bake definitions                                                                                                                                                                  0.0s
 => => reading from stdin 394B                                                                                                                                                                              0.0s
 => [internal] load build definition from Dockerfile                                                                                                                                                        0.0s
 => => transferring dockerfile: 87B                                                                                                                                                                         0.0s
 => [internal] load metadata for docker.io/library/debian:bookworm                                                                                                                                          0.5s
 => [internal] load .dockerignore                                                                                                                                                                           0.0s
 => => transferring context: 2B                                                                                                                                                                             0.0s
 => CACHED [1/1] FROM docker.io/library/debian:bookworm@sha256:b6507e340c43553136f5078284c8c68d86ec8262b1724dde73c325e8d3dcdeba                                                                             0.0s
 => => resolve docker.io/library/debian:bookworm@sha256:b6507e340c43553136f5078284c8c68d86ec8262b1724dde73c325e8d3dcdeba                                                                                    0.0s
 => exporting to docker image format                                                                                                                                                                        0.3s
 => => exporting layers                                                                                                                                                                                     0.0s
 => => exporting manifest sha256:a1f4177034ca86ae0dd77c3a0f3b7051dcd2e2c6c1563edc63a39c8b7712d61a                                                                                                           0.0s
 => => exporting config sha256:466a98363892fcc85bbab0a92d292999b291e34f24ca80ab7394e7f7c2dce937                                                                                                             0.0s
 => => sending tarball                                                                                                                                                                                      0.3s
 => importing to docker                                                                                                                                                                                     0.0s
 => resolving provenance for metadata file                                                                                                                                                                  0.0s
[+] Running 2/2
 ✔ hello-world                          Built                                                                                                                                                               0.0s
 ✔ Container hello-world-hello-world-1  Started
```

[...]

PR #13107, opened 2025-07-28
buildkit require os.Stdout to access the raw terminal

ndeloof:
**What I did**
buildkit (actually, containerd/console) requires a `*File` to access the raw terminal and render build progress UI
it actually can't just be based on a file descriptor (we can get for `dockerCli.Out().FD()`) as windows support requires an actual file 🥲
so .. we have to assume dockerCli.Out == os.Stdout, which basically makes dockerCli.Out abstraction useless 🥹

**Related issue**
fixes https://github.com/docker/compose/issues/13099

### DIFF
--- a/pkg/compose/build_bake.go
+++ b/pkg/compose/build_bake.go
@@ -133,11 +133,10 @@ func (s *composeService) doBuildBake(ctx context.Context, project *types.Project
 	displayMode := progressui.DisplayMode(options.Progress)
 	out := options.Out
 	if out == nil {
-		cout := s.dockerCli.Out()
-		if !cout.IsTerminal() {
+		if !s.dockerCli.Out().IsTerminal() {
 			displayMode = progressui.PlainMode
 		}
-		out = cout
+		out = os.Stdout // should be s.dockerCli.Out(), but NewDisplay require access to the underlying *File
 	}
 	display, err := progressui.NewDisplay(out, displayMode)
 	if err != nil {


thaJeztah:
Hm... interesting; I saw this PR and recalled I stumbled on something similar. While it's for `In()` in this case, I guess this may partially answer the question I dropped in a TODO here, but I had no Windows machine to verify those bits; https://github.com/docker/cli/blob/a1035b0796241b3bbd0f41940525373001fc146c/cli/command/registry.go#L128-L139

```go
func PromptUserForCredentials(ctx context.Context, cli Cli, argUser, argPassword, defaultUsername, serverAddress string) (registrytypes.AuthConfig, error) {
	// On Windows, force the use of the regular OS stdin stream.
	//
	// See:
	// - https://github.com/moby/moby/issues/14336
	// - https://github.com/moby/moby/issues/14210
	// - https://github.com/moby/moby/pull/17738
	//
	// TODO(thaJeztah): we need to confirm if this special handling is still needed, as we may not be doing this in other places.
		cli.SetIn(streams.NewIn(os.Stdin))
	}
```


Similar code in another place; https://github.com/docker/cli/blob/a1035b0796241b3bbd0f41940525373001fc146c/internal/prompt/prompt.go#L90-L93


Definitely looks like something to look into. The `Streams` types are convenient, but it's definitely possible that _either_ someone implemented them with the wrong assumptions, or that these subtleties got lost during some refactor.

PR #13114, opened 2025-07-29
only monitor attached services on up command

### LINKED ISSUE #13113: [BUG] Faulty service is attached despite it's value
### Description

Starting docker compose version v2.39.0, the service `attach` key seems to be misfunctional. In case of a service that fails (with `restart: on-failure`) with `attach: false` value, the docker compose command is "stuck" - due to that faulty service, despite the attach value being false. The expected behaviour (that was until v2.39.0) is that the docker compose command will not wait for a service to (successfully) complete. I suspect the reason for that buggy behavior is https://github.com/docker/compose/pull/13010 which was introduced in v2.39.0

### Steps To Reproduce

1. Install docker with docker compose plugin v2.39.0
2. create the below dummy docker-compose.yaml:
```
services:
  svc-1:
    image: ubuntu
    restart: on-failure
    entrypoint: "hostname"
  svc-2:
    image: ubuntu
    restart: on-failure
    command: ["ls", "/nil"]
    attach: false
```
3. run `docker compose up`
4. Expected: the above command will completed successfully (as the svc-2 is with attach:false)
5. Actual: the above command is stuck as svc-2 fails 

### Compose Version

```Text
Docker Compose version v2.39.0
```

### Docker Environment

```Text

```

### Anything else?

_No response_

glours:
**What I did**
Only pass the list of attached services to the logs monitor

**Related issue**
Fixes #13113 

image

### DIFF
--- a/pkg/compose/up.go
+++ b/pkg/compose/up.go
@@ -168,7 +168,8 @@ func (s *composeService) Up(ctx context.Context, project *types.Project, options
 	if len(options.Start.Services) > 0 {
 		monitor.withServices(options.Start.Services)
 	} else {
-		monitor.withServices(project.ServiceNames())
+		// Start.AttachTo have been already curated with only the services to monitor
+		monitor.withServices(options.Start.AttachTo)
 	}
 	monitor.withListener(printer.HandleEvent)
 


mohammedahmadi-1995:
👍✔

PR #13130, opened 2025-08-04
use cli-plugins/metadata package

thaJeztah:
The metadata types and consts where moved to a separate package, so update the code to use the new location instead of the aliases provided.

**What I did**

**Related issue**


glours (approved):
LGTM

PR #13132, opened 2025-08-05
add sigint and sigterm support for logs when follow flag is used

### LINKED ISSUE #13118: [BUG] `logs -f` is not terminated by Ctrl-C / SIGINT
### Description

This behavior is new with Docker Compose shipped with Docker 28.3.3. It is not seen in the Compose shipped with 28.3.2

Run `docker compose logs -f service` and hit Ctrl-C. The log display stops, but does not terminate. A second Ctrl-C terminates it, with exit code `255`

`docker compose logs service` does not show this: If Ctrl-C is hit during log display, Compose exits as expected.

Expected behavior: Ctrl-C / SIGINT terminates a `logs -f` display

### Steps To Reproduce

See above

### Compose Version

```Text
`Docker Compose version v2.39.1`
```

### Docker Environment

```Text
 Version:    28.3.3
 Context:    default
 Debug Mode: false
 Plugins:
    Version:  v0.26.1
    Version:  v2.39.1

Server:
 Containers: 26
  Running: 25
  Paused: 0
  Stopped: 1
 Images: 42
 Server Version: 28.3.3
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
 CDI spec directories:
  /etc/cdi
  /var/run/cdi
 Swarm: inactive
 Default Runtime: runc
 Init Binary: docker-init
 init version: de40ad0
 Security Options:
  apparmor
  seccomp
   Profile: builtin
  cgroupns
 Kernel Version: 6.1.0-37-amd64
 OSType: linux
 Architecture: x86_64
 CPUs: 12
 Total Memory: 125.4GiB
 Name: rpc3-a
 ID: 37efb9cd-cf8d-4491-acb9-7173e9f77d13
 Debug Mode: false
 Experimental: false
 Insecure Registries:
  ::1/128
  127.0.0.0/8
 Live Restore Enabled: false
```

### Anything else?

_No response_

glours:
**What I did**
Add a channel to monitor SIGINT and SIGTERM and exit immediately the `logs` command when `-f` or `--follow` flag is used

**Related issue**
Fix #13118 

image

### DIFF
--- a/pkg/compose/logs.go
+++ b/pkg/compose/logs.go
@@ -19,6 +19,9 @@ package compose
 import (
 	"context"
 	"io"
+	"os"
+	"os/signal"
+	"syscall"
 
 	"github.com/containerd/errdefs"
 	"github.com/docker/docker/api/types/container"
@@ -71,6 +74,11 @@ func (s *composeService) Logs(
 	}
 
 	if options.Follow {
+		signalChan := make(chan os.Signal, 2)
+		defer close(signalChan)
+		signal.Notify(signalChan, syscall.SIGINT, syscall.SIGTERM)
+		defer signal.Stop(signalChan)
+
 		printer := newLogPrinter(consumer)
 
 		monitor := newMonitor(s.apiClient(), projectName)
@@ -106,6 +114,10 @@ func (s *composeService) Logs(
 		eg.Go(func() error {
 			return monitor.Start(ctx)
 		})
+		for {
+			<-signalChan
+			return nil
+		}
 	}
 
 	return eg.Wait()


pkg/compose/logs.go
```diff
@@ -106,6 +114,10 @@ func (s *composeService) Logs(
 		eg.Go(func() error {
 			return monitor.Start(ctx)
 		})
```
alessio-perugini, line 116:
```suggestion
		eg.Go(func() error {
			return monitor.Start(ctx)
		})
```

Sorry if I jump in, but I got bitten by the recent change done by #13062.
With the fix you propose in this PR I'm afraid we'll have a goroutine leak. I think the monitor.Start will keep running.

Considering only this codebase's use case for the Logs function, the leak won't be a big deal because you always quit the process. However, this will become a problem for other users who use this method (`Logs`) and do not want to terminate the process.

I think it would be great if we could pass to the `monitor.Start`/`newMonitor` something to make it go back the old behaviour (quitting the monitor upon context cancellation)

glours:
replaced by #13193

PR #13140, opened 2025-08-08
chore: use pinned dependencies for github-actions

mmorel-35:
**What I did**

Scorecard asks for dependencies to be pinned (github-actions as Docker images), this focus on github-actions.
It provides the commits number to the used actions.
It also updates dependabot so both commit and semver are checked weekly for update.

[OpenSSF Scorecard](https://securityscorecards.dev/viewer/?uri=github.com/docker/compose)

**Related issue**

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,5 +1,9 @@
 version: 2
 updates:
+  - package-ecosystem: github-actions
+    directory: /
+    schedule:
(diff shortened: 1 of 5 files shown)


ndeloof (dismissed):

mmorel-35:
@ndeloof ,

I rebased and updated pinned version to match the new one

PR #13141, opened 2025-08-08
chore: enable extra-rules from gofumpt and paramTypeCombine from go-critic

mmorel-35:
**What I did**

Enable extra-rules from gofumpt and paramTypeCombine from go-critic

**Related issue**

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -40,7 +40,6 @@ linters:
               desc: compose-go uses yaml.v3
     gocritic:
       disabled-checks:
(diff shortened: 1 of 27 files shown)


ndeloof (commented):
IHMO combined parameters types make sense when they are related, like `prefix, suffix string` but are more confusing when parameters have same type by change but unrelated roles, like `registry, credentials string`. So I'd prefer we don't enforce such a rule.
Any reason you want those set ?

PR #13155, opened 2025-08-15
Add `provenance: max` shorthand for per-service build attestations

maxproske:
**What I did**

Added a more readable `provenance: max` syntax as an alternative to the existing `provenance: "mode=max"`. Less verbose to encourage its use, since max is the default mode for docker/build-push-action on public GitHub repos. Follows the same pattern as other boolean/string configs (eg. pull: true/always), so hopefully it isn't against the spec.

Also added tests.

## Example Usage

```yaml
# compose.yaml
services:
  app:
    build:
      provenance: max # shorthand for "mode=max"
```
```sh
./bin/build/docker-compose build --print

...

"attest": [
    "type=provenance,mode=max"
]
```

### DIFF
--- a/pkg/compose/build_bake.go
+++ b/pkg/compose/build_bake.go
@@ -473,6 +473,8 @@ func toBakeAttest(build types.BuildConfig) []string {
 	if build.Provenance != "" {
 		if build.Provenance == "true" {
 			attests = append(attests, "type=provenance")
+		} else if build.Provenance == "max" {
+			attests = append(attests, "type=provenance,mode=max")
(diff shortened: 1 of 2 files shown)


ndeloof:
We try to align with docker CLI for compose file format and syntax, so as long as `docker build` does not support `--provenance=max`, we won't offer such a syntaxic sugar in compose. Anyway I agree this would help with adoption, maybe you can propose this on github.com/docker/cli

glours:
Hello @maxproske 
Is it ok if we close this PR?

PR #13163, opened 2025-08-20
Fix build slowdown when OpenTelemetry Exporter env variables are present

### LINKED ISSUE #13157: [BUG] Delays in build time when environment files have OTEL-related values and built with bake
### Description

When building an image via compose on its own, a fully cached build takes about one second, as expected.  When built in the context of a test automation, the same fully cached build was consistently taking about thirty seconds (all expected output was displayed within a second, but just 'hung' after that).  I finally determined the difference was an environment variable file.

```sh
docker compose build  // very fast
```

```sh
export COMPOSE_ENV_FILES=.env.it

docker compose build  // now very slow, despite being fully cached
docker compose up --wait -d  // intended usage of the environment file
// a bunch of other up/run executions, which is why COMPOSE_ENV_FILES is used
// rather than having --env-file all over the place
```

I would not expect an environment file to influence the build when it's not used/applicable to the build.

### Steps To Reproduce

This is a simple reproduction, though convoluted.  My actual scenario entails an app with a number of instrumentation services (including an OTEL collector).

Compose file (`compose.yml`):
```yaml
name: foo
services:
  bar:
    build:
      dockerfile_inline: 'FROM busybox'
```
Environment file (`.env2`):
```env 
OTEL_TRACES_EXPORTER=otlp
```

JSON format makes it easier to see where the delay is; the standalone build (no environment file) takes about one second:
```sh
$ docker compose --progress json build && date
{"vertexes":[{"digest":"sha256:d2899eaf0a248c34a3019eab848de181281357af11a26783aa8f3d0e77c88f8e","name":"[internal] load local bake definitions","started":"2025-08-15T23:43:44.043025811-06:00"}]}
<snip>
{"vertexes":[{"digest":"sha256:8fd38d818d8dfc6e3d5102ad680794cedb05705a6f6da902f81c20e1eb158929","name":"resolving provenance for metadata file","started":"2025-08-15T23:43:44.546957079-06:00","completed":"2025-08-15T23:43:44.549499075-06:00"}]}
{"id":"foo-bar","status":"Built"}
Fri Aug 15 11:43:44 PM MDT 2025
```

Specifying an environment file containing an OTEL-related value (intended for containers) takes ten seconds (.3 seconds to actually build, the remainder spent being 'stuck'):
```sh
$ COMPOSE_ENV_FILES=.env2 docker compose --progress json build && date
{"vertexes":[{"digest":"sha256:886408019599d4ef02e885797cec086043cedc306bc1209b2ef6bd07a8e556c3","name":"[internal] load local bake definitions","started":"2025-08-15T23:45:12.488612617-06:00"}]}
<snip>
{"vertexes":[{"digest":"sha256:3bcbe203111b9e1c54c504aa3da8ee1d380334bb3c6c8d69705a820355273d12","name":"resolving provenance for metadata file","started":"2025-08-15T23:45:12.710200999-06:00","completed":"2025-08-15T23:45:12.712882571-06:00"}]}
{"id":"foo-bar","status":"Built"}
Fri Aug 15 11:45:22 PM MDT 2025
```

In my summary I said I was witnessing delays around thirty seconds; I suspect `OTEL_EXPORTER_OTLP_ENDPOINT` was largely responsible, but the lag was not consistent (I'm sure related to whether that endpoint is active).  But the one I used above was very consistent, no doubt due to a ten second timeout somewhere.

Also, this used a non-standard file to make it easier to demonstrate (and also reflects my actual scenario), but the behavior is the same whether it's the default `.env` or custom (like my example) or if it's specified via `--env-file` or `COMPOSE_ENV_FILES`.

### Compose Version

```Text
Docker Compose version v2.39.1 (plugin)
Docker Compose version v2.15.1 (standalone; not used)
```

### Docker Environment

```Text
 Version:    28.3.3
 Context:    default
 Debug Mode: false
 Plugins:
    Version:  v1.9.3
    Path:     /home/robertovillarreal/.docker/cli-plugins/docker-ai
    Version:  v0.26.1
    Version:  v2.39.1
  model: Docker Model Runner (EXPERIMENTAL) (Docker Inc.)
    Version:  v0.1.36
    Path:     /usr/libexec/docker/cli-plugins/docker-model
  scan: Docker Scan (Docker Inc.)
    Version:  v0.23.0
    Path:     /usr/libexec/docker/cli-plugins/docker-scan

Server:
 Containers: 6
  Running: 6
  Paused: 0
  Stopped: 0
 Images: 76
 Server Version: 28.3.3
 Storage Driver: overlayfs
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
 CDI spec directories:
  /etc/cdi
  /var/run/cdi
 Swarm: inactive
 Runtimes: runc sysbox-runc io.containerd.runc.v2
 Default Runtime: runc
 Init Binary: docker-init
 init version: de40ad0
 Security Options:
  apparmor
  seccomp
   Profile: builtin
  cgroupns
 Kernel Version: 6.8.0-60-generic
 Operating System: Ubuntu 24.04.3 LTS
 OSType: linux
 Architecture: x86_64
 CPUs: 16
 Total Memory: 30.56GiB
 Name: l-9jylpn3
 ID: B6K2:2BOW:BSIE:WIGE:RODV:GC2B:JMYF:6XP4:25AT:3S4Q:6634:3OII
 Debug Mode: true
  File Descriptors: 106
  Goroutines: 295
  System Time: 2025-08-16T00:01:27.812705522-06:00
  EventsListeners: 1
 Experimental: true
 Insecure Registries:
<snip>
  192.168.1.0/24
  ::1/128
  127.0.0.0/8
 Registry Mirrors:
  http://localhost:5005/
  http://localhost:5006/
  http://localhost:5007/
 Live Restore Enabled: false
 Default Address Pools:
   Base: 172.25.0.0/16, Size: 24
```

### Anything else?

This has similarities to #12998 (but may be just that... similar).

This is specific to bake (i.e., this does not happen with `COMPOSE_BAKE=false`), but I don't think it is bake's fault.  I'm pretty sure I know where the problem is so I didn't try older versions, but I did try with main with the same results.

It looks like anything passed via an environment file is added to the environment of compose itself
https://github.com/docker/compose/blob/f266715dd0310f1020b195443bf6cafe1922f313/cmd/compose/compose.go#L687-L693

and though some additions and removals are done, it's more-or-less kept intact
https://github.com/docker/compose/blob/f266715dd0310f1020b195443bf6cafe1922f313/pkg/compose/shellout.go#L33-L59

before passing it to the buildx plugin/process.

To complicate things, that near-verbatim passing of environment file to buildx (in #11974) was intentional:
> also restore support for configuring buildx using env var set in .env

which I assume means the overriding of bake variables, not configuring buildx behavior itself.

This seems more like an oversight or unintended consequence rather than a true bug.  But it goes against expectations... I wouldn't expect an environment variable intended to control how my OTEL container is used to be (indirectly) used to build the images itself.

Note: if the environment itself had those OTEL values, I would expect (or at least not be surprised) if they trickled down to buildx.  But they are not in my environment, but in a file that the docs say are used only to replace values in a compose file or other environment files.

maxproske:
Closes #13157

Fixes a 10s slowdown on `docker compose build`, including cached builds, when `OTEL_EXPORTER_*` or `OTEL_*_EXPORTER` env variables are present. Scrubs OpenTelemetry env variables when shelling out to bake.

Bake was called with the full project environment. When a user's .env file include OpenTelemetry exporter env variables (eg. `OTEL_EXPORTER_OTLP_ENDPOINT=http://otel-collector:4318`), these were passed to buildx, which attempted to init its OTLP exporter and connect to an unreachable collector, making the process wait for a network timeout. For the user, this was during the "resolving provenance for metadata file" step visually.

This is a short-term hotfix for the immediate issue, before more projects opt-out of Compose bake with `COMPOSE_BAKE=false`. A better long term solution might be to sanitize the bake env with a minimal allowlist (eg. DOCKER_, BUILDX_, etc.) or update hasOtelEndpointInEnv detection in internal/tracing.

## Steps to reproduce

```bash
cat > Dockerfile <<EOF
FROM alpine
EOF

cat > compose.yaml <<EOF
services:
  app:
    build: .
EOF

cat > .env <<EOF
OTEL_EXPORTER_OTLP_ENDPOINT=http://otel-collector:4318
EOF

echo "--- Slow ---"
time docker compose build
# 11 seconds

echo "--- Fast ---"
time COMPOSE_BAKE=false docker compose build
# 0.5 seconds

echo "--- Fast ---"
rm .env
time docker compose build
# 0.5 seconds
```

### DIFF
--- a/pkg/compose/shellout.go
+++ b/pkg/compose/shellout.go
@@ -19,6 +19,7 @@ package compose
 import (
 	"context"
 	"os/exec"
+	"strings"
 
 	"github.com/compose-spec/compose-go/v2/types"
 	"github.com/docker/cli/cli-plugins/metadata"
@@ -33,6 +34,13 @@ func (s *composeService) prepareShellOut(gctx context.Context, env types.Mapping
 	env = env.Clone()
 	// remove DOCKER_CLI_PLUGIN... variable so a docker-cli plugin will detect it run standalone
 	delete(env, metadata.ReexecEnvvar)
+
+	// Remove OpenTelemetry env variables so buildx doesn't attempt to init exporter and timeout
+	for k := range env {
+		if strings.HasPrefix(k, "OTEL_") {
+			delete(env, k)
+		}
+	}
 
 	// propagate opentelemetry context to child process, see https://github.com/open-telemetry/oteps/blob/main/text/0258-env-context-baggage-carriers.md
 	carrier := propagation.MapCarrier{}


ndeloof:
:-1 This fix would disable Otel tracing for all users, just to prevent a delay for a corner case (`OTEL_*` variable set in `.env` with an unreachable target).

maxproske:
Ah, thank you @ndeloof, I was hoping to only catch `.env`/`--env-file` sourced variables, but this was more aggressive?

The core issue, as I see it, is although Compose has always merged .env into its own environment, the timeout only occurs when Bake is enabled -- That degraded change is what I want to solve.

Because without a way to scope env variables to services only, users are unknowingly penalized for using Compose as an orchestrator for a CNCF project (OTel) just because Docker happens to use it too, and end up disabling Bake https://github.com/mozilla/bedrock/pull/16464

What do you suggest, or can this PR be closed as superseded by #13180?

glours:
superseded by https://github.com/docker/compose/pull/13180

PR #13176, opened 2025-08-25
replace most uses of hashicorp/go-multierror for stdlib

thaJeztah:
### internal/sync: replace go-multierror.Group with golang.org/x/sync/errgroup

The go-multierror Group is just a shallow wrapper around sync.WaitGroup;
https://github.com/hashicorp/go-multierror/blob/v1.1.1/group.go#L5-L38

It does not limit concurrency, but handles synchronisation to collect
all errors (if any) in a go-multierror.

This patch replaces the go-multierror.Group for a sync.ErrGroup (which
is slightly easier to use, and does allow for limiting concurrency if
wanted), and a basic slice with mutex to collect the errors and to produce
a stdlib multi-error through errors.Join

### internal/tracing: replace go-multierror.Group with sync.WaitGroup

The go-multierror Group is just a shallow wrapper around sync.WaitGroup;
https://github.com/hashicorp/go-multierror/blob/v1.1.1/group.go#L5-L38

This patch replaces the go-multierror.Group for a sync.WaitGroup (we
probably don't need to limit concurrency for this one) and stdlib multi-
errors.

### pkg/compose: pull: use native multi-errors

**What I did**

**Related issue**


internal/sync/tar.go
```diff
@@ -84,25 +84,38 @@ func (t *Tar) Sync(ctx context.Context, service string, paths []*PathMapping) er
...
 	}
-	var eg multierror.Group
+
+	var (
+		eg    errgroup.Group
+		errMu sync.Mutex
+		errs  = make([]error, 0, len(containers)*2) // max 2 errs per container
+	)
+
+	eg.SetLimit(16) // arbitrary limit, adjust to taste :D
```
thaJeztah, line 94: For this one, I used an `errgroup` so that we have the option to limit concurrency, but we could use a regular waitgroup as well

internal/tracing/mux.go
```diff
@@ -28,23 +29,45 @@ type MuxExporter struct {
...
+		wg    sync.WaitGroup
+		errMu sync.Mutex
+		errs  = make([]error, 0, len(m.exporters))
+	)
+
+	for _, exporter := range m.exporters {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			if err := exporter.ExportSpans(ctx, spans); err != nil {
```
thaJeztah, line 42: I didn't use a `errgroup` for these, but we could if we want the slightly easier UX. But once we have go1.25 as minimum version, these can be simpified using `WaitGroup.Go`; https://pkg.go.dev/sync#WaitGroup.Go

thaJeztah:
I have one more remaining after this to update `composeService.Up()` (in `pkg/compose`), but that one was a bit more involved, so I'll push that one as a follow-up as it may need more eyes.

PR #13177, opened 2025-08-25
pkg/compose: composeService.Up: rewrite without go-multierror

thaJeztah:
- follow-up to https://github.com/docker/compose/pull/13176

---

- Use a errgroup.Group and add a appendErr utility to not fail-fast, but collect errors.
- replace doneCh for a global context to cancel goroutines
- Commented out attachCtx code, as it didn't appear to be functional (as it wouldn't be cancelled).

**What I did**

**Related issue**

### DIFF
--- a/pkg/compose/up.go
+++ b/pkg/compose/up.go
@@ -18,10 +18,12 @@ package compose
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"os"
 	"os/signal"
 	"slices"
+	"sync"
 	"sync/atomic"
 	"syscall"
 
@@ -33,7 +35,6 @@ import (
 	"github.com/docker/compose/v2/pkg/api"
 	"github.com/docker/compose/v2/pkg/progress"
 	"github.com/eiannone/keyboard"
-	"github.com/hashicorp/go-multierror"
 	"github.com/sirupsen/logrus"
 	"golang.org/x/sync/errgroup"
 )
@@ -61,14 +62,11 @@ func (s *composeService) Up(ctx context.Context, project *types.Project, options
 		return err
 	}
 
-	var eg multierror.Group
-
 	// if we get a second signal during shutdown, we kill the services
 	// immediately, so the channel needs to have sufficient capacity or
 	// we might miss a signal while setting up the second channel read
 	// (this is also why signal.Notify is used vs signal.NotifyContext)
 	signalChan := make(chan os.Signal, 2)
-	defer close(signalChan)
 	signal.Notify(signalChan, syscall.SIGINT, syscall.SIGTERM)
 	defer signal.Stop(signalChan)
 	var isTerminated atomic.Bool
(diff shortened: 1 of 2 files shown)


pkg/compose/up.go
```diff
@@ -61,14 +62,11 @@ func (s *composeService) Up(ctx context.Context, project *types.Project, options
...
-	var eg multierror.Group
-
 	// if we get a second signal during shutdown, we kill the services
 	// immediately, so the channel needs to have sufficient capacity or
 	// we might miss a signal while setting up the second channel read
 	// (this is also why signal.Notify is used vs signal.NotifyContext)
 	signalChan := make(chan os.Signal, 2)
-	defer close(signalChan)
 	signal.Notify(signalChan, syscall.SIGINT, syscall.SIGTERM)
 	defer signal.Stop(signalChan)
```
thaJeztah, line 71: Closing the `signalChan` isn't needed for these; `signal.Stop` should be enough.

pkg/compose/up.go
```diff
@@ -103,26 +101,45 @@ func (s *composeService) Up(ctx context.Context, project *types.Project, options
...
+				return nil
 			})
 			isTerminated.Store(true)
 		}
 
 		for {
 			select {
-			case <-doneCh:
+			case <-globalCtx.Done():
 				if watcher != nil {
```
thaJeztah, line 143: I had a bit of input from my AI friends to see if we could have everything handled through the context (there were some paths that were a bit involved, and I saw some paths were possibly `errgroup.Wait` were not handled).

So pay close attention if I did it right, as I'm not so familiar with this code ❤️ 🙈

pkg/compose/up.go
```diff
@@ -212,9 +234,14 @@ func (s *composeService) Up(ctx context.Context, project *types.Project, options
 	// the primary context is still used for other operations
 	// this means that once any attach operation fails, all other attaches are cancelled,
 	// but an attach failing won't interfere with the rest of the start
-	_, attachCtx := errgroup.WithContext(ctx)
-	containers, err := s.attach(attachCtx, project, printer.HandleEvent, options.Start.AttachTo)
+	//
+	// FIXME(thaJeztah): this context was never cancelled, so didn't add anything? See https://github.com/docker/compose/commit/2c12ad19db5003cdf4cc0de8711970dbb6914b17
+	// _, attachCtx := errgroup.WithContext(ctx)
+	// containers, err := s.attach(attachCtx, project, printer.HandleEvent, options.Start.AttachTo)
+	containers, err := s.attach(globalCtx, project, printer.HandleEvent, options.Start.AttachTo)
```
thaJeztah: This part seemed like it was not yet functional, or at least, the derived context was never cancelled, and the new error-group was discarded, so effectively it was just a copy of the context that was never cancelled (added in https://github.com/docker/compose/commit/2c12ad19db5003cdf4cc0de8711970dbb6914b17)

I didn't remove the code yet, because it probably needs to be looked at, and either removed, or perhaps the "attach" ones being in their own errgroup? I wasn't sure, so could use input here.

ndeloof: seems to me this is legacy code that wasn't removed during recent refactoring

thaJeztah: OK, let me remove the comment and commented-out code

thaJeztah:
CI is green again; ptal @ndeloof 🤗

ndeloof (approved):

PR #13180, opened 2025-08-26
only propagate os.Env to bake, not the whole project.Environment

ndeloof:
**What I did**

**Related issue**


ndeloof:
cc @maxproske

Issue #13134, opened 2025-08-07
[BUG] Impossible to build on remote since 2.38.2


sprietNathanael:
### Description

We use docker compose to build images on remote servers.
It worked fine until `docker-compose-plugin 2.38.2`.
Since then, the docker images are now built on local machine.

### Steps To Reproduce

*What works*
1. On Ubuntu 25.04, with the package `docker-compose-plugin=2.38.1-1~ubuntu.25.04~plucky`
2. execute `docker -H ssh://<remoteServer> compose build`
3. The image is present in the remote server and not in the local machine

*What does not work*
1. `sudo apt remove docker-compose-plugin`
2. `sudo apt install docker-compose-plugin=2.38.2-1~ubuntu.25.04~plucky`
3. execute `docker -H ssh://<remoteServer> compose build`
4. The image is now present in locale machine and not in the remote server

### Compose Version

```Text
docker compose version
Docker Compose version v2.38.2



docker-compose version
command not found: docker-compose
```

### Docker Environment

```Text
Client: Docker Engine - Community
 Version:    28.3.3
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.26.1
    Path:     /usr/libexec/docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.38.2
    Path:     /usr/libexec/docker/cli-plugins/docker-compose

Server:
 Containers: 0
  Running: 0
  Paused: 0
  Stopped: 0
 Images: 1
 Server Version: 28.3.3
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local splunk syslog
 CDI spec directories:
  /etc/cdi
  /var/run/cdi
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 05044ec0a9a75232cad458027ca83437aae3f4da
 runc version: v1.2.5-0-g59923ef
 init version: de40ad0
 Security Options:
  apparmor
  seccomp
   Profile: builtin
[...]

PR #13182, opened 2025-08-26
detect compose run wit --host and set DOCKER_HOST accordingly running bake

ndeloof:
**What I did**

enforce child process uses the same docker host as compose by setting DOCKER_* variables

** note **: after docker/cli processed flags we can't get the original source for TLS keys, so create a temporary cert folder.

**Related issue**
fix https://github.com/docker/compose/issues/13134

### DIFF
--- a/pkg/compose/build_bake.go
+++ b/pkg/compose/build_bake.go
@@ -335,6 +335,12 @@ func (s *composeService) doBuildBake(ctx context.Context, project *types.Project
 	if err != nil {
 		return nil, err
 	}
+	endpoint, cleanup, err := s.propagateDockerEndpoint()
+	if err != nil {
+		return nil, err
+	}
+	cmd.Env = append(cmd.Env, endpoint...)
+	defer cleanup()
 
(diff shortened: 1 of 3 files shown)


pkg/compose/build_bake.go
```diff
@@ -335,6 +335,14 @@ func (s *composeService) doBuildBake(ctx context.Context, project *types.Project
...
 		return nil, err
 	}
+	endpoint, cleanup, err := s.propagateDockerEndpoint()
+	if err != nil {
+		return nil, err
+	}
+	cmd.Env = append(cmd.Env, endpoint...)
+	defer cleanup()
+
+	fmt.Println(strings.Join(cmd.Env, "\n"))
```
glours: Not sure we want to print the env variables of the command by default, WDYT?
```suggestion
	logrus.Debugf((strings.Join(cmd.Env, "\n"))
```

ndeloof: my bad, some debugging stuff left in this PR

PR #13197, opened 2025-09-01
go.mod: bump buildx v0.28.0-rc2, buildkit v0.24.0-rc2

thaJeztah:
- supersedes / closes https://github.com/docker/compose/pull/13196
- relates to https://github.com/moby/buildkit/pull/6172

**What I did**

**Related issue**

### DIFF
--- a/pkg/compose/build_bake.go
+++ b/pkg/compose/build_bake.go
@@ -218,7 +218,7 @@ func (s *composeService) doBuildBake(ctx context.Context, project *types.Project
 
 		read = append(read, build.Context)
 		for _, path := range build.AdditionalContexts {
-			_, err := gitutil.ParseGitRef(path)
+			_, _, err := gitutil.ParseGitRef(path)
 			if !strings.Contains(path, "://") && err != nil {
 				read = append(read, path)
 			}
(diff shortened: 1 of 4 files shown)


pkg/compose/build_bake.go
```diff
@@ -218,7 +218,7 @@ func (s *composeService) doBuildBake(ctx context.Context, project *types.Project
 
 		read = append(read, build.Context)
 		for _, path := range build.AdditionalContexts {
-			_, err := gitutil.ParseGitRef(path)
+			_, _, err := gitutil.ParseGitRef(path)
```
thaJeztah, line 221: @tonistiigi @crazy-max was the boolean that was added needed to be handled anywhere (see other uses in this PR)? Not exactly sure what the boolean was for.

tonistiigi: You can use the boolean to distinguish invalid input and invalid git ref errors. So `false` if you do `ParseGitRef("foobar")` and `true` if input is a git URL but it can't be parsed into ref type because the querystring options are incorrect (eg. you set both branch and tag in the same URL).

ndeloof (approved):

PR #13201, opened 2025-09-04
document (hidden) --tty --interactive flags

ndeloof:
**What I did**

Those flags are hidden as they are no-op flags introduced to align with docker/cli. Still, there are a few relevant usages, typically when it comes to use Compose in a shell script

**Related issue**
close https://github.com/docker/docs/issues/23352

### DIFF
--- a/docs/reference/compose_exec.md
+++ b/docs/reference/compose_exec.md
@@ -6,6 +6,12 @@ This is the equivalent of `docker exec` targeting a Compose service.
 With this subcommand, you can run arbitrary commands in your services. Commands allocate a TTY by default, so
 you can use a command such as `docker compose exec web sh` to get an interactive prompt.
 
+By default, Compose will enter container in interactive mode and allocate a TTY, while the equivalent `docker exec`
+command requires passing `--interactive --tty` flags to get the same behavior. Compose also support those two flags
+to offer a smooth migration between commands, whenever they are no-op by default. Still, `interactive` can be used to
+force disabling interactive mode (`--interactive=false`), typically when `docker compose exec` command is used inside
+a script.
+
 ### Options
 
 | Name              | Type          | Default | Description                                                                      |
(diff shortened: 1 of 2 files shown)


docs/reference/compose_exec.md
```diff
@@ -6,6 +6,12 @@ This is the equivalent of `docker exec` targeting a Compose service.
 With this subcommand, you can run arbitrary commands in your services. Commands allocate a TTY by default, so
 you can use a command such as `docker compose exec web sh` to get an interactive prompt.
 
+By default, Compose will enter container in interactive mode and allocate a TTY, while the equivalent `docker exec`
```
aevesdocker, line 9:
```suggestion
By default, Compose will enter a container in interactive mode and allocate a TTY, while the equivalent `docker exec`
```

docs/reference/compose_exec.md
```diff
@@ -6,6 +6,12 @@ This is the equivalent of `docker exec` targeting a Compose service.
...
+command requires passing `--interactive --tty` flags to get the same behavior. Compose also support those two flags
```
aevesdocker, line 10:
```suggestion
command requires passing `--interactive --tty` flags to get the same behavior. Compose also supports these two flags
```

docs/reference/compose_exec.md
```diff
@@ -6,6 +6,12 @@ This is the equivalent of `docker exec` targeting a Compose service.
...
+to offer a smooth migration between commands, whenever they are no-op by default. Still, `interactive` can be used to
```
aevesdocker, line 11:
```suggestion
to allow a smooth migration between commands, even when they are no-op by default. `interactive` can still be used to
```

docs/reference/compose_exec.md
```diff
@@ -6,6 +6,12 @@ This is the equivalent of `docker exec` targeting a Compose service.
...
+force disabling interactive mode (`--interactive=false`), typically when `docker compose exec` command is used inside
```
aevesdocker, line 12:
```suggestion
disable interactive mode (`--interactive=false`), typically when `docker compose exec` command is used inside
```

aevesdocker:
sorry @ndeloof, I should've been clearer - my wording suggestions will need to also apply to line 38 onwards on the .md file, and also the .yaml

glours (approved):

Issue #13161, opened 2025-08-19
[BUG] Docker compose up --watch stopped working


tamis-laan:
### Description

I upgraded to the latest version of docker compose and now I get the following:

```
              ⦿ Syncing service "app" after 1 changes were detected
app-1         | INFO:uvicorn.error: Shutting down
app-1         | INFO:uvicorn.error: Waiting for application shutdown.
app-1         | INFO:uvicorn.error: Application shutdown complete.
app-1         | INFO:uvicorn.error: Finished server process [1]
              ⦿ service(s) ["app"] restarted
app-1 exited with code 0
```
I have `sync+restart` in my watch configuration for app. On file change the app is properly restarted but the logs and not printed any more. I know this because `docker compose logs -f app` will show logs properly.

Now this problem has come up in the past and was fixed, but I think someone reverted things or reimplemented the same mistake.

I'm on macos 14.7.6 (23H626) using the following versions through homebrew:

```
colima 0.8.4
docker 28.3.3
docker-buildx 0.26.1
docker-compose 2.39.2
```

mateuspiresl:
I've been experiencing similar behavior since last week.

Since then, restarting a container stops displaying the service logs in the existing `compose up` process output.

I've heard colleagues experiencing the same and also seeing duplicated log lines. I couldn't reproduce that last behavior, though.

coredumperror:
Same thing happens to me on Docker Desktop 4.45.0 with Docker Compose v2.39.2-desktop.1. 

It does not happen with Docker Desktop 4.44.3 and Docker Compose v2.39.1-desktop.1, so it seems to have been introduced in the latest patch release of Compose (which is the version that fixed the log duplication... so I bet that fix is probably the cause).

As an additional note, pressing Ctrl-C on a `docker compose up --watch` with v2.39.2-desktop.1 immediately kills the app, rather than starting the container shutdown sequence, which is what it did before. Possibly related?

ulyssessouza:
@tamis-laan Could you please provide a minimalistic version of your `compose.yaml` so that I can reproduce the issue?

ndeloof:
I was not able to reproduce with a minimal example:
```yaml
services:
  app:
    build:
      dockerfile_inline: |
        FROM alpine:latest
        CMD ping localhost
    develop:
      watch:
        - path: .
          target: /tmp
          action: sync+restart
```

@coredumperror can't reproduce either:
```
$ docker compose up --watch
[+] Running 1/1
[...]

PR #13210, opened 2025-09-05
detect container is restarted

ndeloof:
**What I did**
detect a container is restarted. `stop` event doesn't tell us anything about container state (just, engine received a stop request) so better rely on `die` event which demonstrates container termination, then we can inspect and know container is restarting

** note ** I wish `die` event includes a `restarting` attribute so we don't even need to inspect container. I will check I can offer a contribution to moby for this purpose

**Related issue**
fixes https://github.com/docker/compose/issues/13161
supersede https://github.com/docker/compose/pull/13199


ndeloof:
Proposed https://github.com/moby/moby/pull/50908 so we could make this even simpler when engine is running API > x.y

glours (approved):
LGTM + tests ok 👌

PR #13216, opened 2025-09-06
pkg/compose: use state consts from moby API

thaJeztah:
**What I did**

**Related issue**

### DIFF
--- a/pkg/compose/container.go
+++ b/pkg/compose/container.go
@@ -20,23 +20,17 @@ import (
 	"io"
 
 	moby "github.com/docker/docker/api/types"
+	"github.com/docker/docker/api/types/container"
 )
 
 const (
-	// ContainerCreated created status
-	ContainerCreated = "created"
-	// ContainerRestarting restarting status
-	ContainerRestarting = "restarting"
-	// ContainerRunning running status
-	ContainerRunning = "running"
-	// ContainerRemoving removing status
-	ContainerRemoving = "removing"
-	// ContainerPaused paused status
-	ContainerPaused = "paused"
-	// ContainerExited exited status
-	ContainerExited = "exited"
-	// ContainerDead dead status
-	ContainerDead = "dead"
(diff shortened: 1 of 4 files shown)


pkg/compose/container.go
```diff
@@ -20,23 +20,17 @@ import (
...
-	ContainerExited = "exited"
-	// ContainerDead dead status
-	ContainerDead = "dead"
+	ContainerCreated    = container.StateCreated    // StateCreated indicates the container is created, but not (yet) started.
+	ContainerRunning    = container.StateRunning    // StateRunning indicates that the container is running.
+	ContainerPaused     = container.StatePaused     // StatePaused indicates that the container's current state is paused.
+	ContainerRestarting = container.StateRestarting // StateRestarting indicates that the container is currently restarting.
+	ContainerRemoving   = container.StateRemoving   // StateRemoving indicates that the container is being removed.
+	ContainerExited     = container.StateExited     // StateExited indicates that the container exited.
+	ContainerDead       = container.StateDead       // StateDead indicates that the container failed to be deleted. Containers in this state are attempted to be cleaned up when the daemon restarts.
```
ndeloof: why keep those declaration as we won't use them anymore ?

thaJeztah: They are exported, so we technically can't remove them until compose v3

ndeloof: We never considered Compose a "library" and we also don't strictly apply semver. While we try not to break backward compatibility in the (misnamed) `api` package, stuff in `pkg/compose` should be considered internal implementation details

thaJeztah: Thx! Yeah, I went looking, but found various imports, including in testcontainers-go, so it felt a bit risky in case users would have it as (indirect) import; https://grep.app/search?q=%22github.com%2Fdocker%2Fcompose%2Fv2%2Fpkg%2Fcompose

But I have not checked if those importers use these consts though.

thaJeztah:
@ndeloof rebased, and added a commit to remove them 👍

PR #13233, opened 2025-09-18
Disable Tty if run command started from a piped command

### LINKED ISSUE #13228: [BUG] Error when stdin is not a TTY but stdout is a TTY
### Description

`docker-compose` will error with a misleading message when stdin is not a TTY but stdout is a TTY, and `stdin_open: true`. `docker` doesn't have this problem.

### Steps To Reproduce

1. `compose.yaml`:
```yaml
services:
  main:
    image: debian:trixie
    command: ['cat']
    stdin_open: true
```
2. `echo a | docker-compose run --rm main | cat`
    You will get an `a` as output.
3. `echo a | docker-compose run --rm main`
    I expect to get an `a` as output, because `echo a | docker run -i --rm debian:trixie cat` also outputs an `a`. But I got `the input device is not a TTY`, which is misleading --- I did not ask for a TTY, why is `docker-compose` complaining the input device not being a TTY? And it's very hard to guess that the actual problem is related to stdout (being a TTY), not stdin.

### Compose Version

```Text
# docker compose version
Docker Compose version 2.26.1-4
# docker-compose version
Docker Compose version 2.26.1-4
```

### Docker Environment

```Text
Client:
 Version:    26.1.5+dfsg1
 Context:    default
 Debug Mode: false
 Plugins:
    Version:  0.13.1+ds1
    Version:  2.26.1-4
...
 Kernel Version: 6.12.43+deb13-amd64
 Operating System: Debian GNU/Linux 13 (trixie)
...
```

### Anything else?

_No response_

glours:
**What I did**
Checked if `run` is started from a piped command such as `echo test | docker compose run my-service`, if this is the case and the user don't change any tty related flags then we automatically pass in `no-Tty` mode.

**Related issue**
Fixes #13228 

image

### DIFF
--- a/cmd/compose/run.go
+++ b/cmd/compose/run.go
@@ -184,7 +184,10 @@ func runCommand(p *ProjectOptions, dockerCli command.Cli, backend api.Service) *
 					options.noTty = !ttyFlag
 				}
 			} else if !cmd.Flags().Changed("no-TTY") && !cmd.Flags().Changed("interactive") && !dockerCli.In().IsTerminal() {
-				// Check if the command was piped or not, if so, force noTty to tru
+				// while `docker run` requires explicit `-it` flags, Compose enables interactive mode and TTY by default
+				// but when compose is used from a scripr has stdin piped from another command, we just can't
+				// Here, we detect we run "by default" (user didn't passed explicit flags) and disable TTY allocation if
+				// we don't have an actual terminal to attach to for interactive mode
 				options.noTty = true
 			}
 


cmd/compose/run.go
```diff
@@ -178,9 +181,16 @@ func runCommand(p *ProjectOptions, dockerCli command.Cli, backend api.Service) *
 				if cmd.Flags().Changed("no-TTY") {
 					return fmt.Errorf("--tty and --no-TTY can't be used together")
 				} else {
-					options.noTty = !options.tty
+					options.noTty = !ttyFlag
 				}
 			}
+			//Check if the command was piped or not
+			stdinInfo, err := os.Stdin.Stat()
```
ndeloof: why don't you use dockerCli.stdin().IsTerminal ?

glours: Fixed!

glours: But unfortunately don't behave like the `os.Stdin.Stat()` check, it still have the `the input device is not a TTY` error when checking the cli stdin with `IsTerminal`

cmd/compose/run.go
```diff
@@ -178,9 +181,13 @@ func runCommand(p *ProjectOptions, dockerCli command.Cli, backend api.Service) *
 				if cmd.Flags().Changed("no-TTY") {
 					return fmt.Errorf("--tty and --no-TTY can't be used together")
 				} else {
-					options.noTty = !options.tty
+					options.noTty = !ttyFlag
 				}
+			} else if !cmd.Flags().Changed("no-TTY") && !cmd.Flags().Changed("interactive") && !dockerCli.In().IsTerminal() {
+				// Check if the command was piped or not, if so, force noTty to tru
```
ndeloof:
```suggestion
				// while `docker run` requires explicit `-it` flags, Compose enables interactive mode and TTY by default
				// but when compose is used from a scripr has stdin piped from another command, we just can't
				// Here, we detect we run "by default" (user didn't passed explicit flags) and disable TTY allocation if
				// we don't have an actual terminal to attach to for interactive mode
```

PR #13078, opened 2025-07-21
migrate to moby modules

thaJeztah:
- [x] depends on https://github.com/docker/compose/pull/13278
- testing https://github.com/moby/moby/pull/50280
- testing https://github.com/docker/cli/pull/6196
- testing https://github.com/docker/buildx/pull/3326

**What I did**

**Related issue**

**(not mandatory) A picture of a cute animal, if possible in relation to what you did**

PR #13236, opened 2025-09-19
pkg/compose: build: remove permissions warning on Windows

thaJeztah:
- relates to https://github.com/docker/cli/pull/6432
- relates to https://github.com/docker/compose/pull/13078

This warning was added in [moby@4a8b3ca] to print a warning when building Linux images from a Windows client. Window's filesystem does not have an "executable" bit, which mean that, for example, copying a shell script to an image during build would lose the executable bit. So for Windows clients, the executable bit would be set on all files, unconditionally.

Originally this was detected in the client, which had direct access to the API response headers, but when refactoring the client to use a common library in [moby@535c4c9], this was refactored into a `ImageBuildResponse` wrapper, deconstructing the API response into an `io.Reader` and a string field containing only the `OSType` header.

This was the only use and only purpose of the `OSType` field, and now that BuildKit is the default builder for Linux images, this warning didn't get printed unless BuildKit was explicitly disabled.

This patch removes the warning, so that we can potentially remove the field, or the `ImageBuildResponse` type altogether.

[moby@4a8b3ca]: https://github.com/moby/moby/commit/4a8b3cad6096854027151dfbcfb4b2cd8841ad95
[moby@535c4c9]: https://github.com/moby/moby/commit/535c4c9a59b1e58c897677d6948a595cb3d28639

**What I did**

**Related issue**

### DIFF
--- a/pkg/api/dryrunclient.go
+++ b/pkg/api/dryrunclient.go
@@ -218,8 +218,7 @@ func (d *DryRunClient) ImageBuild(ctx context.Context, reader io.Reader, options
 	rc := io.NopCloser(bytes.NewReader(jsonMessage))
 
 	return build.ImageBuildResponse{
-		Body:   rc,
-		OSType: "",
+		Body: rc,
 	}, nil
 }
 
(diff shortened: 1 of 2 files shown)


thaJeztah:
@ndeloof this one should be ready to review as well 👍

ndeloof (approved):

Issue #13235, opened 2025-09-19
[BUG] 2.39.3 prevents building with secrets values read from .env file - update doc?


paul-ri:
### Description

Since 2.39.3, docker-compose will not set the value of a secret from the `.env` file during build.

## Reproduction steps

3 files needed:
1. `.env`
```
BAR="success!"
```
2. `docker-compose.yaml`
```yaml
secrets:
  FOO:
    environment: BAR

services:
  test:
    build:
      secrets:
        - FOO
      dockerfile: Dockerfile
```
3. `Dockerfile`
```dockerfile
FROM debian

RUN --mount=type=secret,id=FOO,env=FOO_BUILDTIME_ENV \
    printenv | sort \
    && [ -n "${FOO_BUILDTIME_ENV}" ]
```
`docker compose build --no-cache` fails in 2.39.3, passes on 2.39.2.

## Investigation

This PR #13180, and specifically the discussion thread in the related #13157 issue, suggest this behaviour was not intended to be supported. Maybe I'm wrong!

> one person's bug is another person's feature

That may be me here...

## Suggestion

Maybe the documentation [on secrets](https://docs.docker.com/reference/compose-file/secrets/) and [on `.env`](https://docs.docker.com/compose/how-tos/environment-variables/variable-interpolation/#env-file) could be updated to reflect secret values will never be read from `.env`? The compose doc on using secrets whose value comes from an env-var for building is explicitly [shown in here](https://docs.docker.com/compose/how-tos/use-secrets/#build-secrets). Here I'm just expecting this to work with the `.env` trick too

In any case, this ticket may help someone knowing this was not meant to be a supported use case. The release notes weren't super clear about this fix.

## My use case

That was very useful to set a secret token value fetched from a `.env` file to pull private Python packages when building. Git would ignore the `.env` file and all looks clean.
That `.env` has a bunch of local secrets used in different services, set in 1 place (a Makefile) but spread by docker-compose to each service for different purposes.

I'll switch to use Bake file to build. Otherwise I'd now have to set the env var in the build process. Or create a secret file per secret and use `file: .bar.env` in the secret instead of `environment: BAR`.

### Docker Environment

```Text
Arch Linux
$ docker --version
Docker version 28.4.0, build d8eb465f86
```

ndeloof:
100%  a bug, https://github.com/docker/compose/pull/13180 is about running the child `bake` process with unexpected environment variable set, still build definition should be resolved by compose to forward secrets.

[...]

PR #13237, opened 2025-09-22
resolve secrets based on env var before executing bake

ndeloof:
**What I did**
Use a temp folder to store all secrets set from environment variable before running bake child process. This allows to run bake without having to set environment with all values from `.env` which may collide with bake own configuration variables

**Related issue**
fix https://github.com/docker/compose/issues/13235


glours (approved):
LGTM

PR #13240, opened 2025-09-24
pluginMain: remove uses of DockerCLI.Apply

thaJeztah:
relates to:

- https://github.com/docker/cli/pull/4574
- https://github.com/docker/cli/pull/6467

### pluginMain: remove uses of DockerCLI.Apply

The Apply method was added when CLI options for constructing the CLI were rewritten into functional options in [cli@7f207f3]. There was no mention in the pull request of this method specifically, and we want to remove or reduce functions that mutate the CLI configuration after initialization if possible (and likely remove the `Apply` method).

This patch removes the use of the `Apply` function as an intermediate step; improvements will be made in the CLI itself for a more solid implementation.

[cli@7f207f3]: https://github.com/docker/cli/commit/7f207f3f957ed3f5129aeb22bef2a429c14caf22

**What I did**

**Related issue**

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -47,8 +47,15 @@ func pluginMain() {
 			if err := plugin.PersistentPreRunE(cmd, args); err != nil {
 				return err
 			}
-			// compose-specific initialization
-			dockerCliPostInitialize(dockerCli)
+			// compose-specific initialization. This must be called AFTER
+			// plugin.PersistentPreRunE.
+			//
+			// FIXME(milas): remove once https://github.com/docker/cli/pull/4574 is merged; for now,
+			//  set it in a rather roundabout way by grabbing the underlying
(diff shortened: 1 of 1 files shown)


thaJeztah:
FWIW, manually verified it still applies as expected, as I wasn't sure there were tests for this;

```bash
docker run -d --rm --name debugger -p 127.0.0.1:2375:8080 mendhak/http-https-echo

./bin/build/docker-compose -H tcp://localhost:2375 ps
no configuration file provided: not found

docker logs debugger 
...
    "headers": {
        "host": "localhost:2375",
        "user-agent": "compose/v2.39.4-4-g4761fd88b.m",
        "traceparent": "00-27c84494c639321cc7268e62945a22e2-d028e97a1b306252-01",
        "accept-encoding": "gzip"
    },
```

PR #13056, opened 2025-07-16
drop support for internal buildkit builder

ndeloof:
**What I did**
Removed internal buildkit-based builder, as Compose can delegate build to Bake and this is set as default since a few releases already.

**Related issue**

**(not mandatory) A picture of a cute animal, if possible in relation to what you did**

PR #13245, opened 2025-09-25
use containerd registry client

ndeloof:
**What I did**
replace use of buildx's imagetool with direct use of containerd's registry client
this is related to https://github.com/docker/compose/pull/13056, removing compose use of buildx codebase

**Related issue**

### DIFF
--- a/internal/ocipush/push.go
+++ b/internal/ocipush/push.go
@@ -26,9 +26,10 @@ import (
 	"slices"
 	"time"
 
+	"github.com/containerd/containerd/v2/core/remotes"
 	pusherrors "github.com/containerd/containerd/v2/core/remotes/errors"
+	"github.com/containerd/errdefs"
 	"github.com/distribution/reference"
-	"github.com/docker/buildx/util/imagetools"
 	"github.com/docker/compose/v2/pkg/api"
 	"github.com/opencontainers/go-digest"
 	"github.com/opencontainers/image-spec/specs-go"
@@ -67,11 +68,6 @@ var clientAuthStatusCodes = []int{
 	http.StatusProxyAuthRequired,
 }
 
-type Pushable struct {
-	Descriptor v1.Descriptor
-	Data       []byte
-}
-
 func DescriptorForComposeFile(path string, content []byte) v1.Descriptor {
 	return v1.Descriptor{
 		MediaType: ComposeYAMLMediaType,
@@ -81,6 +77,7 @@ func DescriptorForComposeFile(path string, content []byte) v1.Descriptor {
 			"com.docker.compose.version": api.ComposeVersion,
 			"com.docker.compose.file":    filepath.Base(path),
 		},
+		Data: content,
 	}
 }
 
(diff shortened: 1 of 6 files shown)


thaJeztah:
Looks like you missed a test-file;

```
116.9 pkg/compose/apiSocket.go:1: : # github.com/docker/compose/v2/pkg/compose [github.com/docker/compose/v2/pkg/compose.test]
116.9 pkg/compose/publish_test.go:61:32: undefined: ocipush.Pushable (typecheck)
```

pkg/compose/publish.go
```diff
@@ -64,11 +64,29 @@ func (s *composeService) publish(ctx context.Context, project *types.Project, re
...
+	config := s.dockerCli.ConfigFile()
+
+	resolver := docker.NewResolver(docker.ResolverOptions{
+		Hosts: docker.ConfigureDefaultRegistries(
+			docker.WithAuthorizer(docker.NewDockerAuthorizer(
+				docker.WithAuthCreds(func(host string) (string, string, error) {
+					if host == "registry-1.docker.io" {
+						host = "https://index.docker.io/v1/"
+					}
+					auth, err := config.GetAuthConfig(host)
```
thaJeztah: Maybe this should use the internal utility to keep it central? Or is the logic different here? https://github.com/docker/compose/blob/032e0309ee09055d9bc563c63e19b197ed405c1c/internal/registry/registry.go#L30-L38


That one doesn't use `registry-1.docker.io` though; ISTR we had that at some point, but some users depended on `registry-1.docker.io` not to be normalised to `docker.io` (and thus "default auth")


(😭 I need to cry every time I run into these)

ndeloof: 😅

internal/ocipush/push.go
```diff
@@ -135,19 +128,32 @@ func PushManifest(
...
+		return nil
+	}
+	if err != nil {
+		return err
+	}
+	_, err = push.Write(descriptor.Data)
+	if err != nil {
+		return err
+	}
+	return nil
```
glours:
```suggestion
	_, err = push.Write(descriptor.Data)
	return err
```

pkg/compose/publish_test.go
```diff
@@ -23,7 +23,6 @@ import (
 
 	"github.com/compose-spec/compose-go/v2/loader"
 	"github.com/compose-spec/compose-go/v2/types"
-	"github.com/docker/compose/v2/internal/ocipush"
 	"github.com/docker/compose/v2/pkg/api"
 	v1 "github.com/opencontainers/image-spec/specs-go/v1"
```
thaJeztah, line 27: (very unrelated) In our other repositories, we alias this import as either `ocispec` (or `ocispecs`); I can open a follow-up PR to update the aliases in this repo, which may help reading the code.

internal/registry/registry.go
```diff
@@ -16,22 +16,21 @@
...
-import "github.com/distribution/reference"
-
 const (
 	// IndexHostname is the index hostname, used for authentication and image search.
 	IndexHostname = "index.docker.io"
 	// IndexServer is used for user auth and image search
 	IndexServer = "https://index.docker.io/v1/"
 	// IndexName is the name of the index
 	IndexName = "docker.io"
+	// LegacyRegistry is the legacy name of the inde
```
thaJeztah: This is actually the canonical, only correct  hostname of the registry 🫠 
https://github.com/moby/moby/blob/2670796a010406b6b35aaff1cdd02f4309e79060/daemon/pkg/registry/config.go#L40-L46

glours (approved):
Just 1 small nit, tests ok on my side 👍

internal/registry/registry.go
```diff
@@ -16,23 +16,13 @@
...
-
-const (
-	// IndexHostname is the index hostname, used for authentication and image search.
-	IndexHostname = "index.docker.io"
-	// IndexServer is used for user auth and image search
-	IndexServer = "https://index.docker.io/v1/"
-	// IndexName is the name of the index
-	IndexName = "docker.io"
-)
+import "github.com/docker/docker/registry"
```
thaJeztah: OH! Don't import this one; it is removed in the next release / moved internal. It's the package we want to get rid of.

ndeloof: 😮‍💨

ndeloof: so, while those are used everywhere there's no canonical place where such const are declared ?

thaJeztah: I don't even think Docker Hub publicly documents them

thaJeztah: > I don't even think Docker Hub publicly documents them

- OH, actually it does now; https://github.com/docker/docs/pull/22497


But yeah, that package was my attempt at having a somewhat documented place for these, and at the time there were hopes that we would finally have the registry running at `docker.io`, but that effort didn't go anywhere, and that package was problematic, and didn't fit in "client" or "api" for Moby.

PR #13251, opened 2025-09-29
use containerd client for OCI operations

ndeloof:
**What I did**

replace use of buildx/util/imagetools with containerd/v2/core/remotes

**Related issue**

see https://github.com/docker/compose/pull/13056

### DIFF
--- a/internal/oci/resolver.go
+++ b/internal/oci/resolver.go
@@ -1,0 +1,72 @@
+/*
+   Copyright 2023 Docker Compose CLI authors
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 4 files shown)


internal/oci/resolver.go
```diff
@@ -0,0 +1,72 @@
...
+import (
+	"context"
+	"io"
+
+	"github.com/containerd/containerd/v2/core/remotes"
+	"github.com/containerd/containerd/v2/core/remotes/docker"
+	"github.com/distribution/reference"
+	"github.com/docker/cli/cli/config/configfile"
+	"github.com/docker/compose/v2/internal/registry"
+	"github.com/opencontainers/image-spec/specs-go/v1"
```
thaJeztah: This probably requires an alias to be used, because of the `v1` package-name;


```
121.1 internal/oci/resolver.go:28:1: File is not properly formatted (goimports)
121.1 	"github.com/opencontainers/image-spec/specs-go/v1"
121.1 ^
```



```suggestion
	v1 "github.com/opencontainers/image-spec/specs-go/v1"
```

PR #13254, opened 2025-09-29
gha: update test-matrix: remove docker 26.x

thaJeztah:
- relates to https://github.com/moby/moby/pull/51018

---

### gha: update test-matrix: remove docker 26.x

- Mirantis Container Runtime (MCR) 23.0 reached EOL, and the next LTS
  version of MCR is 25.x, but download.docker.com does not have 25.x
  packages for the latest Ubuntu release.
- Docker 26.x reached EOL and is no longer maintained

**What I did**

**Related issue**

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -156,9 +156,8 @@ jobs:
           - plugin
           - standalone
         engine:
-          - 26
-          - 27
-          - 28
+          - 27 # old stable (latest major - 1)
+          - 28 # current stable
     steps:
       - name: Prepare
         run: |


thaJeztah:
Ah! I thought something like this would be the case; download.docker.com doesn't have 25.x packages for Ubuntu Noble, so installation fails.


```
# Executing docker install script, commit: 1ebc10ccd106153669a1e7b5984497e0fcf9d3b1
+ sh -c apt-get -qq update >/dev/null
+ sh -c DEBIAN_FRONTEND=noninteractive apt-get -y -qq install ca-certificates curl >/dev/null
+ sh -c install -m 0755 -d /etc/apt/keyrings
+ sh -c curl -fsSL "https://download.docker.com/linux/ubuntu/gpg" -o /etc/apt/keyrings/docker.asc
+ sh -c chmod a+r /etc/apt/keyrings/docker.asc
+ sh -c echo "deb [arch=amd64 signed-by=/etc/apt/keyrings/docker.asc] https://download.docker.com/linux/ubuntu noble test" > /etc/apt/sources.list.d/docker.list
+ sh -c apt-get -qq update >/dev/null
INFO: Searching repository for VERSION '25'
INFO: apt-cache madison docker-ce | grep '25' | head -1 | awk '{$1=$1};1' | cut -d' ' -f 3

ERROR: '25' not found amongst apt-cache madison results
```


Guess we should just remove 26.x, and not add 25.x

thaJeztah:
Ah, this requires changes to the repository config, as the 26 check is marked as "required", so GitHub actions will wait for that check to complete (which it won't)

thaJeztah:
> Ah, this requires changes to the repository config, as the 26 check is marked as "required", so GitHub actions will wait for that check to complete (which it won't)

I don't have permissions to change those rules, but perhaps @ndeloof or @glours can update the settings if this PR looks good otherwise 😅

glours:
@thaJeztah settings updated 😉

thaJeztah:
Thanks!

PR #13268, opened 2025-10-08
Fix: set PWD only if not set

### LINKED ISSUE #13265: [BUG] The PWD environment variable will always be overwritten
### Description

With https://github.com/docker/compose/commit/f38f3f754ce32d6a640c39dcd41af126392986cd the environment variable `PWD` will always be overwritten with the parent working directory using the Go function https://pkg.go.dev/os#Getwd.

This might be OK in 99% of all use cases, but this change breaks a special use case:

I need to set the `PWD` variable differently to match another directory structure within a VM that actually runs the Docker daemon. e.g. a Windows host system and a Debian guest system. The CLI tools `docker` with the `docker compose` plugin are installed on the Windows host and access the Docker daemon on the Debian guest. Directories are synced from the Windows host to the Debian host and therefore have another parent working directory in the context of the VM

**Suggestion:** Please check if a `PWD` environment variable is present and only if it is not set it to the result of `os.Getwd()`. That is a defensive programming technique and will satisfy both use cases.

This defect has been introduced with https://github.com/docker/compose/releases/tag/v2.35.0 with v2.34.0 it is still possible to set a specific `PWD` variable.

### Steps To Reproduce

_No response_

### Compose Version

```Text

```

### Docker Environment

```Text

```

### Anything else?

_No response_

kianelbo:
**What I did**
Check whether a `PWD` env variable is present and only if it isn't, set it to the result of `os.Getwd()`.

**Related issue**
Fixes #13265

### DIFF
--- a/cmd/compose/compose.go
+++ b/cmd/compose/compose.go
@@ -383,32 +383,38 @@ func (o *ProjectOptions) remoteLoaders(dockerCli command.Cli) []loader.ResourceL
 }
 
 func (o *ProjectOptions) toProjectOptions(po ...cli.ProjectOptionsFn) (*cli.ProjectOptions, error) {
-	pwd, err := os.Getwd()
-	if err != nil {
-		return nil, err
+	opts := []cli.ProjectOptionsFn{
+		cli.WithWorkingDirectory(o.ProjectDir),
+		// First apply os.Environment, always win
+		cli.WithOsEnv,
 	}
 
-	return cli.NewProjectOptions(o.ConfigPaths,
-		append(po,
-			cli.WithWorkingDirectory(o.ProjectDir),
-			// First apply os.Environment, always win
-			cli.WithOsEnv,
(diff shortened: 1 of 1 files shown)


cmd/compose/compose.go
```diff
@@ -383,32 +383,38 @@ func (o *ProjectOptions) remoteLoaders(dockerCli command.Cli) []loader.ResourceL
 }
 
 func (o *ProjectOptions) toProjectOptions(po ...cli.ProjectOptionsFn) (*cli.ProjectOptions, error) {
-	pwd, err := os.Getwd()
-	if err != nil {
-		return nil, err
+	envOptions := []cli.ProjectOptionsFn{
```
ndeloof: `envOptions` is a bad name, as those many options are not only set to manage env, but many aspected of project setup. A basic `opts` would be better

ndeloof (approved):

Issue #13269, opened 2025-10-08
[BUG] dockerfile_inline no longer supports ARG


alexleach:
### Description

Some of my existing `compose.yaml` files have recently stopped working, with respect to using `build.args` and `build.dockerfile_inline` together.

An `ARG` is no longer expanded within the `dockerfile_inline` block.

**Current Behaviour**

Here is a minimal non-working example of a `compose.yaml` file:

```yaml
services:
  hello:
    build:
      context: .

      args:
        ARCH: "x86_64"

      dockerfile_inline: |
        FROM alpine:latest
        ARG ARCH
        RUN wget "https://github.com/docker/compose/releases/download/v2.40.0/docker-compose-linux-$ARCH"
```

**Expected Behaviour**

If instead, I write a separate `Dockerfile` and comment out the `dockerfile_inline` block, then it builds fine:

```Dockerfile
FROM alpine:latest
ARG ARCH
RUN wget "https://github.com/docker/compose/releases/download/v2.40.0/docker-compose-linux-$ARCH"
```

### Steps To Reproduce

1. `docker compose build`

### Compose Version

```Text
Docker Compose version 2.40.0
```

### Docker Environment

```Text
Client:
 Version:    28.5.0
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  0.29.1
    Path:     /usr/lib/docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  2.40.0
    Path:     /usr/lib/docker/cli-plugins/docker-compose
  mcp: Docker MCP Plugin (Docker Inc.)
    Version:  v0.20.0
    Path:     /usr/lib/docker/cli-plugins/docker-mcp

Server:
 Containers: 42
  Running: 12
  Paused: 0
  Stopped: 30
 Images: 42
 Server Version: 28.5.0
 Storage Driver: btrfs
  Btrfs:
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local splunk syslog
 CDI spec directories:
  /etc/cdi
  /var/run/cdi
 Discovered Devices:
  cdi: nvidia.com/gpu=0
  cdi: nvidia.com/gpu=GPU-87a03c29-583b-822c-7a8a-fccd4b75cc9c
[...]

PR #13270, opened 2025-10-08
bake only interpolates ${*}

ndeloof:
**What I did**

bake variable use syntax `${VAR}`, not the simpler `$VAR` we also interpolate in docker compose

**Related issue**
fixes https://github.com/docker/compose/issues/13269


pkg/e2e/fixtures/build-test/escaped/compose.yaml
```diff
@@ -4,3 +4,20 @@ services:
...
+        EOF
+
+  arg:
+    build:
+      args:
+        BOOL: "true"
+      dockerfile_inline: |
+        FROM alpine:latest
+        ARG BOOL
+        RUN /bin/$${BOOL}
```
thaJeztah: nit: missing newline at end of file

Issue #13243, opened 2025-09-25
[BUG] <title>failed to execute bake: read |0: file already closed


Zhanbingli:
### Description

Environment
macOS 15.3.x (Apple Silicon – MacBook Air)
Docker Desktop 28.4.0 (context desktop-linux)
Docker Compose v2.39.2-desktop.1
Docker Engine client API 1.51
BuildKit enabled (default)
Steps to Reproduce
Clone the repository and cd medical_net
Run docker compose -f infrastructure/docker-compose.yaml up --build
The build phase fails immediately with:
failed to execute bake: read |0: file already closed
There is also a warning about the version field being obsolete; removing that field does not change the outcome.
The compose file defines several services that build from local folders (backend, etl, frontend) plus a Postgres service.
Expected Behavior
All services should build and start normally.
Actual Behavior
The command stops during the build stage with failed to execute bake: read |0: file already closed. No containers are started.

Additional Info / Workaround
Running DOCKER_BUILDKIT=0 COMPOSE_DOCKER_CLI_BUILD=0 docker compose -f infrastructure/docker-compose.yaml up --build avoids the error.
The same happens if I run docker compose -f infrastructure/docker-compose.yaml build.
Happy to provide docker compose --verbose logs or other diagnostics if needed.

### Compose Version

```Text

```

### Docker Environment

```Text

```

ndeloof:
Please run `docker compose --verbose build` to capture the executed bake config and command line
You also ca try using `docker compose build --print` to capture the bake.json build definition, then run this one with `docker buildx bake -f bake.json`

Zhanbingli:
> Please run `docker compose --verbose build` to capture the executed bake config and command line You also ca try using `docker compose build --print` to capture the bake.json build definition, then run this one with `docker buildx bake -f bake.json`

Thanks for the suggestion!

belimawr:
I'm having the same problem since a while ago, usually I just re-run docker compose and it works, but today I kept getting the same error over and over again.

I followed @ndeloof suggestion and when I ran `docker buildx bake -f bake.json`, it worked and I could run the `docker compose...` again.

Versions
```
% docker compose version
Docker Compose version 2.39.4

% docker-compose version
Docker Compose version 2.39.4
```

Docker Environment
```
docker info
Client:
 Version:    28.4.0
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  0.28.0
[...]

Issue #12998, opened 2025-06-30
[BUG] Building with bake hangs indefinitely


razzeee:
### Description

Our CI started to hang on deployment somewhen last month, as we're not pinning docker compose.

```

❯ docker compose --verbose build
TRAC[0000] Docker Desktop integration not enabled       
WARN[0000] The "NODE_ENV" variable is not set. Defaulting to a blank string. 
WARN[0000] The "SENTRY_AUTH_TOKEN" variable is not set. Defaulting to a blank string. 
WARN[0000] The "NODE_ENV" variable is not set. Defaulting to a blank string. 
WARN[0000] The "SENTRY_AUTH_TOKEN" variable is not set. Defaulting to a blank string. 
WARN[0000] The "NODE_ENV" variable is not set. Defaulting to a blank string. 
WARN[0000] The "SENTRY_AUTH_TOKEN" variable is not set. Defaulting to a blank string. 
DEBU[0000] bake build config:
{
  "group": {
    "default": {
      "targets": [
        "web",
        "worker",
        "nginx",
        "api"
      ]
    }
  },
  "target": {
    "api": {
      "context": "/home/user/dev/project",
      "dockerfile": "/home/user/dev/project/Dockerfile",
      "args": {
        "BASIC_AUTH_TOKEN": "",
        "BASIC_AUTH_TOKEN_USER_IMAGE_EXPORT": "",
        "CI_COMMIT_SHA": "",
      },
      "tags": [
        "project-api"
      ],
      "target": "serve_api",
      "output": [
        "type=docker,load=true,push=false"
      ]
    },
    "nginx": {
      "context": "/home/user/dev/project/nginx",
      "dockerfile": "/home/user/dev/project/nginx/Dockerfile",
      "tags": [
        "project-nginx"
      ],
      "output": [
        "type=docker,load=true,push=false"
      ]
    },
    "web": {
      "context": "/home/user/dev/project",
      "dockerfile": "/home/user/dev/project/Dockerfile",
      "args": {
        "BASE_URL": "https://test-prog.domain.local/",
        "CI_COMMIT_SHA": "",
      },
      "tags": [
        "project-web"
      ],
      "target": "serve_web",
      "output": [
        "type=docker,load=true,push=false"
      ]
    },
    "worker": {
      "context": "/home/user/dev/project",
      "dockerfile": "/home/user/dev/project/Dockerfile",
      "args": {
        "BASE_URL": "https://test-proj.domain.local/",
[...]

PR #13275, opened 2025-10-10
Fix failure to delegate build with bake

ndeloof:
**What I did**

Fix failure to delegate build with bake:
- failure to create temporary metadata file
- bake stderr stream being closed

**Related issue**
fixes https://github.com/docker/compose/issues/12998
fixes https://github.com/docker/compose/issues/13243

### DIFF
--- a/pkg/compose/build_bake.go
+++ b/pkg/compose/build_bake.go
@@ -25,7 +25,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"math/rand"
+	"io/fs"
 	"os"
 	"os/exec"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


pkg/compose/build_bake.go
```diff
@@ -39,6 +38,7 @@ import (
 	"github.com/docker/cli/cli/command"
 	"github.com/docker/compose/v2/pkg/api"
 	"github.com/docker/compose/v2/pkg/progress"
+	"github.com/docker/distribution/uuid"
```
thaJeztah: don't use this one; docker/distribution is deprecated;

- https://github.com/docker/compose/pull/10953

thaJeztah: I think we already have google uuid in most places; which is the mostly commonly used.

ndeloof: this is just a diagnostic PR, not intended to be merged

glours (approved):
LGTM

Issue #13276, opened 2025-10-10
[BUG] Publish does not push all compose files when using "extends" with "profiles".


ogoulpeau-ledger:
### Description

It seems that `docker compose publish` misses pushing the associated "extends" files when using profiles.

### Steps To Reproduce

1. With the compose files:

 - `docker-compose.yml`:
```
services:
  etcd:
    extends:
      file: etcd/docker-compose.yml
      service: etcd
    environment:
      ETCD_DEBUG: true
    profiles: [ default ]
```
 - `etcd/docker-compose.yml`:
```
services:
  etcd:
    image: quay.io/coreos/etcd:v3.3.27
```

2. Publishing the app as an OCI artifact:
```
docker compose --profile '*' publish oci://ghcr.io/my-org/my-compose-config:latest
```

3. Results in failure when using that artifact:

  - Expected result:
```
$ docker compose -f oci://ghcr.io/my-org/my-compose-config:latest --profile default config
name: compose-issue
services:
  etcd:
    profiles:
      - default
    environment:
      ETCD_DEBUG: "true"
    image: quay.io/coreos/etcd:v3.3.27
    networks:
      default: null
networks:
  default:
    name: compose-issue_default
```

 - Actual results:
```
$ docker compose -f oci://ghcr.io/my-org/my-compose-config:latest --profile default config
INFO[0000] fetch failed after status: 404 Not Found      host=*** spanID=db89337f59236507 traceID=34db7b71f1b9c951ca3b337db583a63c
open ~/.cache/docker-compose/ff3638119f2b259733ec2bc62ace1501edf2f729e80cae8c625c71cf34d724f5/etcd/docker-compose.yml: no such file or directory
```

### Compose Version

```Text
Docker Compose version v2.40.0
```

### Docker Environment

```Text

```

### Anything else?

Not using "profiles" in the main `docker-compose.yml` file works:
 - `docker-compose.yml`:
```
services:
  etcd:
    extends:
      file: etcd/docker-compose.yml
      service: etcd
    environment:
      ETCD_DEBUG: true
```
 - `etcd/docker-compose.yml`:
```
services:
  etcd:
    image: quay.io/coreos/etcd:v3.3.27
```

PR #13277, opened 2025-10-10
fix: make "publish" push all compose files addressed in "extends" statements when using "profiles".

ogoulpeau-ledger:
**What I did**
 - in `processFile()`, load the compose file selecting the profiles from `project.Profiles` in the `loader.Options`.
 - make the command fail if publishing a compose app with no service:
   - Not sure about this one:
     - Let me know if this is better to remove it
     - I thought also about enforcing all profiles by default in `cmd/compose/publish.go` as if `--profile '*'` was passed but this is not consistent with the other commands, I guess.

**Related issue**
 - Fix for #13276 

### DIFF
--- a/pkg/compose/publish.go
+++ b/pkg/compose/publish.go
@@ -50,6 +50,10 @@ func (s *composeService) Publish(ctx context.Context, project *types.Project, re
 
 //nolint:gocyclo
 func (s *composeService) publish(ctx context.Context, project *types.Project, repository string, options api.PublishOptions) error {
+	project, err := project.WithProfiles([]string{"*"})
+	if err != nil {
+		return err
+	}
 	accept, err := s.preChecks(project, options)
 	if err != nil {
 		return err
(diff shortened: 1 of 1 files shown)


ndeloof:
Profiles indeed are not considered by the publish command. Once published, as the raw compose.yaml has been published, disabled profiles will be available for consumer but not the related images or other references (like `extends` your issue refers to)

IMHO `publish` command _should_ implicitly enable all profiles, so that the published compose model is fully functional whatever profile the consumer eventually enables

ogoulpeau-ledger:
> Profiles indeed are not considered by the publish command. Once published, as the raw compose.yaml has been published, disabled profiles will be available for consumer but not the related images or other references (like `extends` your issue refers to)
> 
> IMHO `publish` command _should_ implicitly enable all profiles, so that the published compose model is fully functional whatever profile the consumer eventually enables

Okay I rewrote the second commit that way.

ogoulpeau-ledger:
Thanks for this quick merge! I enjoy working with Compose and I enjoyed discovering and working on clean code: keep it up guys!

PR #13280, opened 2025-10-13
pkg/compose: build with bake: drop support for buildx v0.16 and lower

thaJeztah:
[buildx v0.17][1] was released a Year ago, so any version this conditional code was accounting for would be versions before that; the latest of which being [buildx v0.16.2][2] (July 2024).

Given that those versions are long EOL and no longer supported, we can probably remove the conditional code.

[1]: https://github.com/docker/buildx/releases/tag/v0.17.0
[2]: https://github.com/docker/buildx/releases/tag/v0.16.2

**What I did**

**Related issue**

### DIFF
--- a/pkg/compose/build_bake.go
+++ b/pkg/compose/build_bake.go
@@ -40,7 +40,6 @@ import (
 	"github.com/docker/cli/cli/command/image/build"
 	"github.com/docker/compose/v2/pkg/api"
 	"github.com/docker/compose/v2/pkg/progress"
-	"github.com/docker/docker/api/types/versions"
 	"github.com/moby/buildkit/client"
 	gitutil "github.com/moby/buildkit/frontend/dockerfile/dfgitutil"
 	"github.com/moby/buildkit/util/progress/progressui"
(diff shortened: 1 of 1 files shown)


ndeloof:
LGTM, but I wonder we could check version number and warn user running an obsolete release of buildx

thaJeztah:
That's always gonna  be tricky;  that applies to all binaries (cli, compose itself, buildx, ... ); we can somewhat fix it in packaging and set a minimum version, https://github.com/docker/packaging/blob/66807ae5e8fbbed4dd498bfe2891c504b78cf38b/pkg/docker-cli/deb/control#L19

ndeloof:
I had in mind Mac/Windows installation with some users relying on homebrew or manual installation, or maybe an older version was selected. We could just check by `versions.GreaterThanOrEqualTo(buildx.Version, xxx)` if some unsupported version is selected

thaJeztah:
Yeah, it's a tricky one; it's good to shell out to `buildx` because the alternative wasn't great, but it also means that the `buildx bake` command-line now becomes a formal contract / API for compose.

I wonder if we should consider moving the options-struct-to-command-line-flags conversion to a utility package in buildx; that way changes made in the buildx command-line will stay in lock-step with the code to set those flags (e.g. if a flag is added / removed / renamed / changed, the buildx maintainers can make sure the code that compose uses to call `buildx bake` is updated as well and set the right version-switch.

_Perhaps_ it would even be an idea to have a (hidden) flag that acccepts a JSON-struct with _all the options_ - that would have a more structured way to pass options to buildx when invoked; i.e., something like

```bash
docker buildx bake --___bake_options='{"version":"v0.29.1", "progress": "rawjson", "provenance": true .....}'
```

- Some version in that struct could tell `buildx bake` what version was expected to be used by compose, which could allow buildx bake to adjust things where needed (or detect "unknown fields" and produce an error that a newer verrsion is needed).
- Some of those options could even be implicit (e.g. when using `__bake_options`, then always use `rawjson`)



@crazy-max any thoughts? Would something like that make sense to do? (i.e. a flag or subcommand to execute buildx programmatically?)

PR #13284, opened 2025-10-14
use fixed version of compose bridge transformer images

glours:
to avoid CI issue on Compose when a new version is released and change the outputs

**What I did**
Use explicit version of Compose Bridge Transformer images for transformation e2e tests 

**Related issue**
N/A

image

### DIFF
--- a/pkg/e2e/bridge_test.go
+++ b/pkg/e2e/bridge_test.go
@@ -17,6 +17,7 @@
 package e2e
 
 import (
+	"fmt"
 	"path/filepath"
 	"strings"
 	"testing"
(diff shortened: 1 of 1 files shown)


.claude/settings.local.json
```diff
@@ -0,0 +1,9 @@
+{
+  "permissions": {
+    "allow": [
+      "Bash(E2E_TEST=TestConvertAndTransformList make build-and-e2e-compose)"
+    ],
+    "deny": [],
+    "ask": []
+  }
+}
```
ndeloof: should be added to ~/.gitignore :)

glours: 🤦

PR #13291, opened 2025-10-16
fix race-condition bug in publish command

### LINKED ISSUE #13290: [BUG] publish race-condition introduced in #13245
### Description

When publishing with `docker compose publish` (without --app) it sometimes fails as the PUT request is not yet finished when the command returns. This can be seen as `request canceled` errors on the server side.
The publish command returns with code 0, even though the publishing did not complete.
This was introduced with #13245 

I think it is because `.Commit()` is not called in https://github.com/docker/compose/blob/main/internal/oci/resolver.go#L121-L133

It should be something like this:
```go
	push, err := pusher.Push(ctx, descriptor)
	if errdefs.IsAlreadyExists(err) {
		return nil
	}
	if err != nil {
		return err
	}

	_, err = push.Write(descriptor.Data)
	if err != nil {
		// Close the writer on error since Commit won't be called
		_ = push.Close()
		return err
	}
	// Commit will close the writer
	return push.Commit(ctx, int64(len(descriptor.Data)), descriptor.Digest)
```

What do you think @ndeloof ?

### Compose Version

```Text
Docker Compose version v2.40.0
```

It can also be seen when verbose logging is turned on:

`docker compose --verbose --progress=plain publish registry.internal/stack`

with the bug no "fetch response received" is logged:
```
DEBU[0000] do request                                    request.header.content-type=application/vnd.oci.image.manifest.v1+json request.header.user-agent=containerd/2.1.4+unknown request.method=PUT spanID=0b114615dc6374b3 traceID=7a4166b557bc5b196ce9e26801563407 url="https://registry.internal/v2/stack/manifests/8978c10"
 registry.internal/stack:8978c10 published 
```

correct behavior is:
```
DEBU[0000] do request                                    request.header.content-type=application/vnd.oci.image.manifest.v1+json request.header.user-agent=containerd/2.1.4+unknown request.method=PUT spanID=4a4fbf9bae9419ca traceID=b4833963b40128fe4c297536f9c21122 url="https://registry.internal/v2/stack/manifests/032e030"
DEBU[0000] fetch response received                       response.header.alt-svc="h3=\":443\"; ma=2592000" response.header.content-length=0 response.header.date="Wed, 15 Oct 2025 15:35:21 GMT" response.header.docker-content-digest="sha256:e70545097296a2c91e9f0a06093a23a811e527b1ce6728396b73ba3550a8b0ab" response.header.docker-distribution-api-version=registry/2.0 response.header.location="https://registry.internal/v2/stack/manifests/sha256:e70545097296a2c91e9f0a06093a23a811e527b1ce6728396b73ba3550a8b0ab" response.header.via="1.1 Caddy" response.header.x-content-type-options=nosniff response.status="201 Created" spanID=4a4fbf9bae9419ca traceID=b4833963b40128fe4c297536f9c21122 url="https://registry.internal/v2/stack/manifests/032e030"
 registry.internal/stack:032e030 published 
```

paul-kinexon:
**What I did**
Add missing `push.Commit` to `resolver.go` to fix race-condition, where the program would end before the manifest was fully pushed to the registry.

**Related issue**
fixes #13290

### DIFF
--- a/internal/oci/resolver.go
+++ b/internal/oci/resolver.go
@@ -125,10 +125,13 @@ func Push(ctx context.Context, resolver remotes.Resolver, ref reference.Named, d
 	if err != nil {
 		return err
 	}
-	defer func() {
-		_ = push.Close()
-	}()
 
 	_, err = push.Write(descriptor.Data)
-	return err
+	if err != nil {
+		// Close the writer on error since Commit won't be called
+		_ = push.Close()
+		return err
+	}
+	// Commit will close the writer
+	return push.Commit(ctx, int64(len(descriptor.Data)), descriptor.Digest)
 }


ndeloof:
For legal reasons we require contributors to sign-off all commits. Please amend your commit with `git commit --amend --signoff ` and force-push your branch to update this PR

PR #13304, opened 2025-10-20
Refactor to remove dependencies on github.com/docker/docker/pkg/streamformatter and

austinvazquez:
github.com/docker/docker/pkg/progress

**What I did**

This changes copies the streamformatter and progress packages from github.com/docker/docker/api. The upcoming Docker 29 release is introducing Go modules: moby/api, moby/client, and moby/moby. Previously part of the top level pkg package, the streamformatter and progress packages are being made internal daemon packages.

**Related issue**

Moby issue: https://github.com/moby/moby/issues/50575
Moby PR: https://github.com/moby/moby/pull/51153


ndeloof:
> Copied from https://github.com/moby/moby

Why?
We should just follow code being relocated in api module: https://github.com/moby/moby/pull/50565

austinvazquez:
> > Copied from https://github.com/moby/moby
> 
> Why? We should just follow code being relocated in api module: [moby/moby#50565](https://github.com/moby/moby/pull/50565)

👍🏼 , @ndeloof, I added a writeup to the description to clarify. PTAL.

ndeloof:
is the docker/cli also supposed to use a local copy of this code then?
IMHO https://github.com/moby/moby/pull/51153 is just wrong. `/imageBuild` API to render build events is non-trivial to manage and this deserves a utility package to handle rendering.

austinvazquez:
I have opened https://github.com/docker/cli/pull/6570 but no review yet from CLI folks. IIUC yes there probably can be a shared place between CLI and compose. Just not sure where that needs to be.

ndeloof:
I'm fine this code is relocated in docker/cli, I just wonder how it will be used if made internal in moby, as this is basically API-consumer code

ndeloof:
we already have docker/cli as a dependency, so if it eventually own (a copy of) this code, would be simpler we just us it from there vs have our own copy

austinvazquez:
I've pushed a new revision to https://github.com/moby/moby/pull/51153 to move the packages out of the api module up into the moby client. Probably long term we can investigate having these be in a better location perhaps CLI or some other Go SDK. For now this unblocks the API module release. Thanks!

PR #13305, opened 2025-10-21
Fix ANSI color preservation across multiple log lines

### LINKED ISSUE #12789: [BUG] ANSI escape color formatting lost on multiple lines in logs
### Description

When a service outputs text with ANSI escape codes for color formatting across multiple lines, `docker compose logs` only preserves the color for the first line.
Subsequent lines are displayed without the indended color.

```yaml
# compose.yaml
services:
  app:
    image: bash
    command: echo -e "\\e[31m This line is RED.\\n This line is also RED.\\n This span is ... \\e[34m BLUE."
```

Expected|Actual
-|-
Image|Image

### Steps To Reproduce

_No response_

### Compose Version

```Text

```

### Docker Environment

```Text

```

### Anything else?

_No response_

chojs23:
Track ANSI color state across log lines so container output with color codes spanning multiple lines preserves formatting on all lines, not just the first.

This fix adds an ansiState field to the presenter struct to track active ANSI SGR (color/style) codes per container and restores them at the start of each line. 
Only handles SGR codes (colors, bold, underline).

Fixes #12789

### DIFF
--- a/cmd/formatter/logs.go
+++ b/cmd/formatter/logs.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"regexp"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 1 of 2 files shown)


ndeloof:
I don't think this is a feature we should support in Compose: While _some_ process may only rely on coloring, ANSI-enabled output is used by many tools with cursor-position ANSI sequence (typically to show computation progress, same as Compose does) which will break Compose logs formatting.

ndeloof:
Closing as "not planned"

PR #13315, opened 2025-10-23
Code Cleanup

ndeloof:
**What I did**

Removed unused function
Avoid access to s.dockerCli to access APIClient and ConfigFile as we have dedicated methods for this purpose

**Related issue**


glours (approved):
LGTM

Issue #13323, opened 2025-10-28
[BUG] runing a service builds its depends-ons twice when pull policy is build


luciangabor:
### Description

Please see the steps to reproduce

### Steps To Reproduce

With a `compose.yaml` like this:
``` yaml
services:
  nginx:
    pull_policy: build
    build:
      dockerfile_inline: |
          FROM nginx
          # this is used as "marker" of time spent during builds
          ADD --link --keep-git-dir https://github.com/kubernetes/kubernetes.git /big-repo
    healthcheck:
      test: curl -fs localhost
      start_interval: 1s
      start_period: 10s

  curl:
    image: alpine/curl
    depends_on:
      nginx:
        condition: service_healthy
    command: curl -fsI nginx
```

`docker compose run curl` will produce the following output:
```console
[+] Building 5.7s (10/10) FINISHED                                                                                                                                                            
 => [internal] load local bake definitions                                                                                                                                               0.0s
 => => reading from stdin 605B                                                                                                                                                           0.0s
 => [internal] load build definition from Dockerfile                                                                                                                                     0.0s
 => => transferring dockerfile: 186B                                                                                                                                                     0.0s
 => [internal] load metadata for docker.io/library/nginx:latest                                                                                                                          1.4s
 => [auth] library/nginx:pull token for registry-1.docker.io                                                                                                                             0.0s
 => [internal] load .dockerignore                                                                                                                                                        0.0s
 => => transferring context: 2B                                                                                                                                                          0.0s
 => [2/2] ADD --link --keep-git-dir https://github.com/kubernetes/kubernetes.git /big-repo                                                                                               3.9s
 => [1/2] FROM docker.io/library/nginx:latest@sha256:029d4461bd98f124e531380505ceea2072418fdf28752aa73b7b273ba3048903                                                                    0.0s
 => CACHED [2/2] ADD --link --keep-git-dir https://github.com/kubernetes/kubernetes.git /big-repo                                                                                        0.0s
 => exporting to image                                                                                                                                                                   0.0s
 => => exporting layers                                                                                                                                                                  0.0s
 => => writing image sha256:94163b15c51df0859e7f02f29c07aa61618af9de589b579b1c73c951c9772ccd                                                                                             0.0s
 => => naming to docker.io/library/run-nginx                                                                                                                                             0.0s
 => resolving provenance for metadata file                                                                                                                                               0.0s
[+] Creating 3/3
 ✔ run-nginx              Built                                                                                                                                                          0.0s 
[...]

PR #13325, opened 2025-10-28
fix(run): Ensure images exist only for the target service in run command

idsulik:
**What I did**
Fixed a bug where services with `pull_policy: build` were built twice during docker compose run when they were dependencies of the target service. The issue occurred because ensureImagesExists was called twice - once for dependencies in `startDependencies` and again for the full project in `prepareRun`. Modified `prepareRun` to restrict the second `ensureImagesExists` call to only build the target service by filtering `BuildOptions.Services`.

**Related issue**
https://github.com/docker/compose/issues/13323

image

### DIFF
--- a/pkg/e2e/compose_run_build_once_test.go
+++ b/pkg/e2e/compose_run_build_once_test.go
@@ -32,10 +32,11 @@ func TestRunBuildOnce(t *testing.T) {
 	c := NewParallelCLI(t)
 
 	t.Run("dependency with pull_policy build is built only once", func(t *testing.T) {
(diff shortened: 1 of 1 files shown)


idsulik:
It's strange... the tests work fine locally

idsulik:
Looks like it was a parallel issue. Should I return the parallel execution and try to fix the issue, or is it ok to leave non-parallel execution?

glours:
Yes fix it please, use `c := NewParallelCLI(t)` instead in your test

idsulik:
done

glours (approved):
LGTM! Thanks @idsulik 🙏

Issue #13330, opened 2025-10-29
[BUG] remote-Git subdir include leads to loading local files when resolving remote git includes


0pepsi:
### Description

When using a remote Git -f URL containing a subdirectory traversal component (e.g. repo#branch:../../../../..../somepath), docker compose resolves and loads a local file outside the intended checkout/cached area. I observed docker compose config loading a compose file stored under /tmp/* on the host when passed a remote git URL with :.. path traversal.

```go
func (g gitRemoteLoader) Load(ctx context.Context, path string) (string, error) {
    // ... parsing and validation ...
    
    local, ok := g.known[path]
    if !ok {
        if ref.Ref == "" {
            ref.Ref = "HEAD" // default branch
        }

        err = g.resolveGitRef(ctx, path, ref)  // Resolves ref to commit SHA
        if err != nil {
            return "", err
        }

        cache, err := cacheDir()
        if err != nil {
            return "", fmt.Errorf("initializing remote resource cache: %w", err)
        }

        local = filepath.Join(cache, ref.Ref)  // ref.Ref is now a commit SHA!
        if _, err := os.Stat(local); os.IsNotExist(err) {
            if g.offline {
                return "", nil
            }
            err = g.checkout(ctx, local, ref)  // Clones repo to cache/SHA
            if err != nil {
                return "", err
            }
        }
        g.known[path] = local
    }
    if ref.SubDir != "" {
        local = filepath.Join(local, ref.SubDir)  // vulnerable line
    }
    stat, err := os.Stat(local)
    if err != nil {
        return "", err
    }
    if stat.IsDir() {
        local, err = findFile(cli.DefaultFileNames, local)  // Looks for compose.yaml
    }
    return local, err
}
```

### Steps To Reproduce

1. create a directory `/tmp/pwned` and place a valid `compose.yaml` inside it. For example:
```
name: pwning
services:
  exploited:
    image: alpine
    command: ["echo", "LOADED FROM /tmp/pwned NOT FROM GIT REPO"]
```
2. `export COMPOSE_EXPERIMENTAL_GIT_REMOTE=1`
3.  run `docker compose -f 'https://github.com/docker/compose.git#main:../../../../../../../tmp/pwned' config`

### Compose Version

```Text
└─$ docker compose version                                                                                    

Docker Compose version v2.40.2
```

### Docker Environment

```Text
a vm



sudo docker info
Client:
 Version:    26.1.5+dfsg1
 Context:    default
 Debug Mode: false
[...]

PR #13331, opened 2025-10-30
fix(git): Fix path traversal vulnerability in git remote loader

idsulik:
**What I did**
This PR fixes a path traversal vulnerability in the Git remote loader where subdirectory components containing .. sequences could escape the cached repository directory and load arbitrary files from the host filesystem (e.g., `https://github.com/docker/compose.git#main:../../../../../../../tmp/pwned`). The fix adds a `validateGitSubDir` function that validates subdirectory paths before joining them with the cache directory, using multiple defense-in-depth checks: rejecting absolute paths, detecting `..` traversal patterns after path normalization, and verifying the final resolved path remains within the base directory 

**Related issue**
[](https://github.com/docker/compose/issues/13330)

image

### DIFF
--- a/pkg/e2e/compose_run_build_once_test.go
+++ b/pkg/e2e/compose_run_build_once_test.go
@@ -17,6 +17,9 @@
 package e2e
 
 import (
+	"crypto/rand"
+	"encoding/hex"
+	"fmt"
 	"strings"
 	"testing"
 
@@ -32,11 +35,11 @@ func TestRunBuildOnce(t *testing.T) {
 	c := NewParallelCLI(t)
 
 	t.Run("dependency with pull_policy build is built only once", func(t *testing.T) {
-		projectName := "e2e-run-build-once-single"
+		projectName := randomProjectName("build-once")
 		res := c.RunDockerComposeCmd(t, "-p", projectName, "-f", "./fixtures/run-test/build-once.yaml", "down", "--rmi", "local", "--remove-orphans")
 		res.Assert(t, icmd.Success)
 
-		res = c.RunDockerComposeCmd(t, "-p", projectName, "-f", "./fixtures/run-test/build-once.yaml", "run", "--rm", "curl")
+		res = c.RunDockerComposeCmd(t, "-p", projectName, "-f", "./fixtures/run-test/build-once.yaml", "run", "--build", "--rm", "curl")
 		res.Assert(t, icmd.Success)
 
 		// Count how many times nginx was built by looking for its unique RUN command output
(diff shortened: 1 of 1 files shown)


glours (approved):
Sounds good, thanks for the quick fix @idsulik

glours:
@idsulik can you create a dedicated PR for the e2e flaky test fix please? I need to merge this one

idsulik:
> @idsulik can you create a dedicated PR for the e2e flaky test fix please? I need to merge this one

Yes, working on it, can't find why it fails on GitHub while locally it works stably

0pepsi:
Great work guys, can i assign it as a cve ?

glours:
@0pepsi No, please, you should have followed the recommendations in [our security policy](https://github.com/docker/compose/blob/main/.github/SECURITY.md)
 to avoid disclosing the vulnerability before a fix was available.

I’m currently checking with our internal security team to determine the best process to follow.
We’ll get back to you shortly.

0pepsi:
Sounds good, mb i thought the whole thing was over (: (👍

temenuzhka-thede:
@0pepsi because this vulnerability was disclosed publicly, via a github issue, we had to urgently patch and release a new version, and bypass the proper workflow via a private Github Security Advisory and a CVE publication. Hence, no CVE will be issued for this.

Our security policy asks that researchers disclose secuity issues privately via security@docker.com and follow a responsible disclosure policy which allows us to privately review a submission within 72 hours, and work on a patch before it is disclosed publicly (we ask for up to 90 days of keeping an issue confidential to allow us to properly patch depending on complexity).

We will now proceed to delete this issue. Please reach out directly to security@docker.com with any follow-up questions or additional findings.

Thank you

j-baines:
> Hence, no CVE will be issued for this.

So you aren't going to tell the world about the vulnerability because you didn't like the way it was disclosed?

temenuzhka-thede:
We've taken a closer look at this issue and appreciated the feedback from community members. While our earlier statement about the CVE disclosure process was inaccurate, we want to clarify that, after further analysis, this is not a vulnerability in Docker Compose and does not warrant a CVE ID. We apologize for the confusion and are sharing the following technical details to provide more context.

**Understanding the Reported Behavior**
In order to trigger this specific issue and load a chosen file from the host via the git loader a potential attacker would need to manipulate the -f argument of the Compose command line. Using the Docker Compose CLI requires the same permissions as running any arbitrary workload on Docker Compose and also requires that the chosen file is accessible by the user, therefore there would be no confidentiality, integrity or availability impact. Other places where the git loader can be used are the "include" top level element or "extends" attribute in compose files but those also allow to specify any file the user has access to and setting them requires write access to the Compose file.

Considering a hypothetical scenario in which a third party application allows untrusted and unsanitized input to be injected into a part of the -f CLI option or in a part of an include or extends field of an otherwise trusted Compose file, that would be a vulnerability of the hypothetical third party application. That said, this PR fixes an actual bug which is the fact that the git loader used to accept an invalid input.

PR #13334, opened 2025-10-30
flaky compose_run_build_once_test: cleanup and more logs when test fa…

glours:
**What I did**
Use directly buildx logs to check the number of time each image was built
I wonder if there wasn't some log format changes between the local version we were using and the one used by the CI (I honestly didn't check)

This is a quick win attempt to fix the CI and we should do a follow up PR with a more robust solution for the long run

The CI was green 3 times in a row with those changes

**Related issue**
N/A


idsulik (approved):

idsulik:
I made several attempts to figure out the cause, but haven’t succeeded yet. I’ll try again over the weekend. I think these changes will help me understand what’s going on, thank you

Issue #13341, opened 2025-11-01
[BUG] Compose Bridge nil pointer error when using models in docker-compose.yml


Vilatsut:
### Description

I encountered a `nil pointer evaluating interface {}.endpoint_var` error when using Docker Compose Bridge to convert my `docker-compose.yml` file into Kubernetes manifests. The issue occurs when the `models` attribute is defined for a service without explicitly setting ``endpoint_var`` and ``model_var`` attributes, even though the documentation suggests these should be automatically handled.

### Steps To Reproduce

1. Create the following ``docker-compose.yml`` file, with the DMR "short" syntax:
```yaml
services:
  app:
    build: .
    ports:
      - "8000:8000"
    models:
      - ai_runner

models:
  ai_runner:
    model: ai/smollm2
```
2. Run the following command to convert the Compose file:
```bash
docker compose bridge convert 
```
3. Observe the error:
```bash
cannot execute template /templates/base/deployment.tmpl: template: deployment.tmpl:60:32: executing "deployment.tmpl" at <$value.endpoint_var>: nil pointer evaluating interface {}.endpoint_var
```

### Compose Version

```Text
Docker Compose version v2.40.3-desktop.1
Docker Compose version v2.40.2-desktop.1
```

### Docker Environment

```Text
lient:
 Version:    28.5.1
 Context:    desktop-linux
 Debug Mode: false
 Plugins:
  ai: Docker AI Agent - Ask Gordon (Docker Inc.)
    Version:  v1.9.11
    Path:     C:\Program Files\Docker\cli-plugins\docker-ai.exe
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.29.1-desktop.1
    Path:     C:\Program Files\Docker\cli-plugins\docker-buildx.exe
  compose: Docker Compose (Docker Inc.)
    Version:  v2.40.3-desktop.1
    Path:     C:\Users\vilatsut\.docker\cli-plugins\docker-compose.exe
  debug: Get a shell into any image or container (Docker Inc.)
    Version:  0.0.45
    Path:     C:\Program Files\Docker\cli-plugins\docker-debug.exe
  desktop: Docker Desktop commands (Docker Inc.)
    Version:  v0.2.0
    Path:     C:\Program Files\Docker\cli-plugins\docker-desktop.exe
  extension: Manages Docker extensions (Docker Inc.)
    Version:  v0.2.31
    Path:     C:\Program Files\Docker\cli-plugins\docker-extension.exe
[...]

PR #13342, opened 2025-11-01
feat(convert): set default model and endpoint variables for services in convert.go

idsulik:
**What I did**
This PR resolves issue #13341 by automatically setting default `model_var` and `endpoint_var` values for services that use models but don't explicitly define these variables

**Related issue**
https://github.com/docker/compose/issues/13341

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ bin/
 coverage.out
 covdatafiles/
 .DS_Store
+vendor


ndeloof (commented):
default values should be set by compose-go as we load a project IMHO

idsulik:
@ndeloof but I'm not sure if it ok to make these changes for all places, because the documentation says only about environment values https://docs.docker.com/reference/compose-file/services/#models  

> If either is omitted, Compose automatically generates the environment variable names based on the model key using the following rules...


but in this PR I made the changes only for the converter, because it doesn't work even if you specify the environment values

ndeloof:
closing as https://github.com/docker/compose/issues/13341 was fixed

Issue #13230, opened 2025-09-17
[epic] Allow use of Docker Compose as a library


ndeloof:
We have recurrent request for third-party software developers to use docker/compose as a go dependency, embedding Compose in their own application. Docker Compose was not initially designed with this use-case in mind.

This epic captures work to complete to actually support such an usage, which implies API stability and well defined boundaries between exposed APIs and internal implementation details subject to changes

- Refine requirements for `composeService` regarding `command.Cli` - we only need a subset
- remove use of `docker/cli/streams` which is highly specific to docker/cli integration
- Move any "implementation details" subject to changes, into `internal` package
- Refactor `RunWithStatus` so that Writer mode is part of the API, not a global variable
- Move `progress` package to `cmd` as a command line component
- Move `prompt` package to `cmd` as a command line component
- Document use of docker/compose as a library

PR #13357, opened 2025-11-05
Move progress package to cmd as a command line component

ndeloof:
**What I did**
Moved `progress` package to cmd as a command line component and renamed
Removed uses of `fmt.Printxx` in pkg/compose to fully delegate UI rendering to display component

**Related issue**
https://github.com/docker/compose/issues/13230

### DIFF
--- a/cmd/display/dryrun.go
+++ b/cmd/display/dryrun.go
@@ -16,41 +16,6 @@
 
 package display
 
-import (
-	"context"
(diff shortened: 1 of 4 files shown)


Sainan:
> (not mandatory) A picture of a cute animal, if possible in relation to what you did

Don't worry, I got you covered

computer cat

PR #13366, opened 2025-11-10
Making the American/British spellings consistent to the error messages

JLesDev:
**What I did**
In the error messages + most of the code base it uses American English spelling of "canceled". However, in some parts of the code, seemingly random, the British English spelling of "cancelled" is used. As this is used in client facing code I just made it more consistent.

**Related issue**

image


JLesDev (commented):
Self explanatory

PR #13370, opened 2025-11-11
Dockerfile: update golangci-lint to v2.6.2

thaJeztah:
**What I did**

**Related issue**


cmd/compose/compose.go
```diff
@@ -387,8 +387,8 @@ func (o *ProjectOptions) toProjectOptions(po ...cli.ProjectOptionsFn) (*cli.Proj
 		// if none was selected, get default compose.yaml file from current dir or parent folder
 		cli.WithDefaultConfigPath,
 		// .. and then, a project directory != PWD maybe has been set so let's load .env file
-		cli.WithEnvFiles(o.EnvFiles...),
-		cli.WithDotEnv,
+		cli.WithEnvFiles(o.EnvFiles...), //nolint:gocritic // intentionally applying cli.WithEnvFiles twice.
+		cli.WithDotEnv,                  //nolint:gocritic // intentionally applying cli.WithDotEnv twice.
```
thaJeztah, line 397: From the comments, I _THINK_ it's intentional to apply these options multiple times, but if not, I can update and remove these

pkg/compose/loader.go
```diff
@@ -95,12 +95,18 @@ func (s *composeService) buildProjectOptions(options api.ProjectLoadOptions, rem
...
 		cli.WithDotEnv,
+		// get compose file path set by COMPOSE_FILE
 		cli.WithConfigFileEnv,
+		// if none was selected, get default compose.yaml file from current dir or parent folder
 		cli.WithDefaultConfigPath,
-		cli.WithEnvFiles(options.EnvFiles...),
-		cli.WithDotEnv,
+		// .. and then, a project directory != PWD maybe has been set so let's load .env file
+		cli.WithEnvFiles(options.EnvFiles...), //nolint:gocritic // intentionally applying cli.WithEnvFiles twice.
+		cli.WithDotEnv,                        //nolint:gocritic // intentionally applying cli.WithDotEnv twice.
```
thaJeztah, line 108: Same here; I also copied the comments from the other code.

cmd/compose/compose.go
ndeloof: this is indeed intentional

pkg/compose/loader.go
ndeloof: I need to check why we have this code duplicated

ndeloof (dismissed):

thaJeztah:
Doh! The nolint tags don't work because the current version used in CI doesn't use that linter; let me have a look at updating it; I can drop those for the time being.

thaJeztah:
@ndeloof @glours Updated; I updated golangci-lint to v2.6.2, PTAL

PR #13380, opened 2025-11-16
gha: test against docker v29, v28, and rename checks to use `stable`, `oldstable`

thaJeztah:
### gha: use custom names for matrix

Manually enumerate the combinations ((plugin|standalone), (version))
so that we can assign a predictable name ("stable", "oldstable") and
prevent having to update the branch-protection rules for each update
to mark the tests as required.

**What I did**

**Related issue**

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -149,15 +149,26 @@ jobs:
         if: always()
   e2e:
     runs-on: ubuntu-latest
+    name: e2e (${{ matrix.mode }}, ${{ matrix.channel }})
     strategy:
       fail-fast: false
       matrix:
-        mode:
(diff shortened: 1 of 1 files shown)


thaJeztah:
@ndeloof @glours PTAL;  I updated the versions, and added an extra commit to use a stable name for the "stable", "oldstable" checks.

Before this, these were named, e.g.;


Screenshot 2025-11-16 at 11 53 35


With this PR, those will now be named;

Screenshot 2025-11-16 at 12 10 58



This will require changes to the branch config, as 27.x, 28.x are still marked as required, but will no longer be there, but after this PR it will no longer be needed to change settings on GitHub whenever the versions are updated 🎉

Issue #13396, opened 2025-11-25
[BUG] Long notation of `volumes` `bind` mount doesn't apply selinux labels


TheRealBecks:
### Description

I already created containers/podman#27600 as I use Podman with docker-compose.

There are short and long notations available for `volumes` under `services`:

Short:

```
services:
  mariadb:
    volumes:
      # private
      - ./mariadbtest1:/var/lib/mysql:Z
```

Long:

```
services:
  mariadb:
    volumes:
      # private
      - type: bind
        source: "mariadbtest2"
        target: "/var/lib/mysql"
        bind:
          selinux: Z
```

The long notation does not apply the already generated selinux label and therefore the volume is not working due to a `Permission denied` error int he container.

The long notation only works when running `setenforce 0` to deactivate selinux entirely.

### Steps To Reproduce

There are two tests that can be run. I named the containers `mariadbtest1` and `mariadbtest2`. Both containers will mount a directory that has the same name as the containers. Each container has it's own `compose.yaml` file:

`mariadbtest1` with short notation:

```
---
name: mariadbtest1
services:
  mariadb:
    image: docker.io/mariadb:latest
    container_name: mariadbtest1
    user: "50000:50000"
    userns_mode: "keep-id"
    volumes:
      # private
      - ./mariadbtest1:/var/lib/mysql:Z
    environment:
      - MARIADB_DATABASE=exmple-database
      - MARIADB_USER=example-user
      - MARIADB_PASSWORD=my_cool_secret
      - MARIADB_ROOT_PASSWORD=my-secret-pw
```

`mariadbtest2` with long notation:

```
---
name: mariadbtest2
services:
  mariadb:
    image: docker.io/mariadb:latest
    container_name: mariadbtest2
    user: "50000:50000"
    userns_mode: "keep-id"
    volumes:
      # private
      - type: bind
        source: "mariadbtest2"
        target: "/var/lib/mysql"
        bind:
          selinux: Z
    environment:
      - MARIADB_DATABASE=exmple-database
      - MARIADB_USER=example-user
      - MARIADB_PASSWORD=my_cool_secret
      - MARIADB_ROOT_PASSWORD=my-secret-pw
```

The to be mounted directory can be checked with `ls -lZ` before and after each test. The selinux labels for `mariadbtest1` will be applied to the directory, but that's not the case for `mariadbtest2`. The `mariadbtest2` directory has teh same selinux labels as before.

[...]

PR #13397, opened 2025-11-26
Long notation of volumes bind mount handle the selinux labels

zhangguanzhang:
Fixes: https://github.com/docker/compose/issues/13396

**What I did**

**Related issue**

### DIFF
--- a/pkg/compose/create.go
+++ b/pkg/compose/create.go
@@ -912,6 +912,8 @@ func bindRequiresMountAPI(bind *types.ServiceVolumeBind) bool {
 	switch {
 	case bind == nil:
 		return false
+	case bind.SELinux != "":
+		return false
 	case !bind.CreateHostPath:
 		return true
 	case bind.Propagation != "":
--- a/pkg/compose/create_test.go
+++ b/pkg/compose/create_test.go
@@ -434,6 +434,21 @@ volumes:
 				},
 			},
 		},
+		{
+			name: "mount bind with selinux",
+			yaml: `
+services:
+  test:
+    volumes:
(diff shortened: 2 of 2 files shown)


pkg/compose/create.go
```diff
@@ -912,6 +912,8 @@ func bindRequiresMountAPI(bind *types.ServiceVolumeBind) bool {
 	switch {
 	case bind == nil:
 		return false
+	case bind.SELinux != "":
```
ndeloof, line 915: this has no effect as return false is the default

zhangguanzhang:
> your test case is ok but the fix isn't. The actual root cause is that `CreateHostPath` isn't set to `true` by default as it is supposed to be by the spec.
> 
> also see [compose-spec/compose-go#836](https://github.com/compose-spec/compose-go/pull/836)

I think I understand. Does it mean this issue should be fixed in the repository https://github.com/compose-spec/compose-go?

ndeloof:
~The root cause is https://github.com/compose-spec/compose-go/commit/978e4cff0f03a29c0a4607ac88f2fd6bd05d41a8 was missing an update to ServiceVolumeBind.String() to include SELinux~

my bad, need one more coffee or read the code more carrefuly :)

ndeloof:
the actual root cause is that, when using the Mount API, Compose can't set SELinux flag as there's no such option on the mount API (see https://github.com/moby/moby/blob/18d2a08fcf1e02aaccc25482267b714f994ba014/api/types/mount/mount.go#L27)

you can report this to github.com/moby/moby

Issue #13394, opened 2025-11-25
[BUG] Inconsistent `compose publish` behavior with variables


apollo13:
### Description

It is not exactly clear from the docs how `compose publish` is supposed to handle variables. There is an `--with-env` flag which somehow seems to suggest that env variables are included in the OCI manifest. When using it docker compose asks whether sensitive variables are included, but the final manifest doesn't seem to include them anyways?

### Steps To Reproduce

Using this compose file:
```
services:
  test1:
    image: nginx:latest
    command: ["${SECRET}"]
    environment:
      TEST: "${SECRET}"
```
and publishing it via:
```
SECRET=test docker compose publish localhost:5000/test:latest --insecure-registry
```
results in an error:
```
service "test1" has environment variable(s) declared.
To avoid leaking sensitive data, you must either explicitly allow the sending of environment variables by using the --with-env flag,
or remove sensitive data from your Compose configuration
```
So far so good, but publishing with `--with-env` doesn't really seem to result in leakage either:
```
SECRET=test docker compose publish localhost:5000/test:latest --insecure-registry --with-env                                                                                                                                                                                                                                                              1 ✘  14:40:46  
? you are about to publish environment variables within your OCI artifact.
please double check that you are not leaking sensitive data
Service/Config  test1
TEST=test
Are you ok to publish these environment variables? Yes
[+] push 1/1
 ✔ test1 Skipped                                                                                                                                                                                                                                                                                                                                                                       0.0s 
[+]  2/2t:5000/test:latest publishing 
```

When checking the published data:
```
regctl --host reg=localhost:5000,tls=disabled blob get localhost:5000/test:latest sha256:5868bab8f786242ba43d6604ea68ac5f4509457d6e8bfb3222ee7e85b4b9575f
```
it does not include the env variables:
```
services:
  test1:
    image: nginx:latest
    command: ["${SECRET}"]
    environment:
      TEST: "${SECRET}"
```

The result seems plausible to me, one usually does not want to publish with environment variables interpolated to allow for customization, but what is `--with-env` warning about then?

### Compose Version

```Text
Docker Compose version v5.0.0-rc.2
```

ndeloof:
The result indeed is the expected one, i.e the original compose file is published to support full customization by consumer.

[...]

PR #13402, opened 2025-11-27
only check for env_file

ndeloof:
**What I did**

as we publish the raw compose.yaml, there's no need to reject use of environment: the actual value from published environment won't be published. The only env var we _could_ publish are those from env_file(s)

**Related issue**
fixes https://github.com/docker/compose/issues/13394

### DIFF
--- a/pkg/compose/publish.go
+++ b/pkg/compose/publish.go
@@ -296,7 +296,6 @@ func (s *composeService) generateImageDigestsOverride(ctx context.Context, proje
 	return override.MarshalYAML()
 }
 
-//nolint:gocyclo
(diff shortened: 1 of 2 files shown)


ndeloof:
@glours I'm not sure why we introduced those checks in the first place, as we don't actually publish interpolated values, and we use scanner to detect hard-coded sensitive data.

glours:
@ndeloof IIRC, the env check was done before we introduced the secret scanner, and I forgot to remove the unnecessary controls.
If there is any issue with this PR, it will be easy to revert it, so let's merge it

PR #13404, opened 2025-11-28
feat(model): ignore runtime flags in model configuration

ilopezluna:
This pull request updates the model configuration logic to improve user feedback and clarify runtime flag handling.


pkg/compose/model.go
```diff
@@ -152,6 +152,9 @@ func (m *modelAPI) PullModel(ctx context.Context, model types.ModelConfig, quiet
 }
 
 func (m *modelAPI) ConfigureModel(ctx context.Context, config types.ModelConfig, events api.EventProcessor) error {
+	if len(config.RuntimeFlags) != 0 {
+		return fmt.Errorf("runtime flags are not supported for model configuration")
+	}
```
glours, line 159: Why not simply warn the user, clear the RuntimeFlags attribute, and continue the flow?
Right now, Compose stops entirely, which isn’t a great UX when we could instead just notify the user that their runtime flags won’t be applied.
This approach would avoid breaking existing Compose applications.

ndeloof: agree

ndeloof (approved):

PR #13406, opened 2025-11-28
golangci-lint: use gci formatter instead of goimports

thaJeztah:
Most files already grouped imports into "stdlib -> other -> local", but some files didn't. The gci formatter is similar to goimports, but has better options to make sure imports are grouped in the expected order (and to make sure no additional groups are present).

This formatter has a 'fix' function, so code can be re-formatted auto- matically;

    golangci-lint run -v --fix

**What I did**

**Related issue**

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -79,11 +79,18 @@ issues:
   max-same-issues: 0
 formatters:
   enable:
+    - gci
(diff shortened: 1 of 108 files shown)


thaJeztah:
rebased

Issue #13414, opened 2025-12-03
[BUG] docker compose build panic


yangfeiyu20102011:
### Description

<img width="834" height="633" alt="Image" src="https://github.com/user-attachments/assets/388a0571-37c3-46df-b418-3829b248c489" />

### Compose Version

```Text

```

### Docker Environment

```Text

```

ndeloof:
fixed by https://github.com/docker/compose/pull/13415

PR #13415, opened 2025-12-03
check buildx version before comparing it

yangfeiyu20102011:
**What I did**

**Related issue**

https://github.com/docker/compose/issues/13414

### DIFF
--- a/pkg/compose/build_bake.go
+++ b/pkg/compose/build_bake.go
@@ -289,13 +289,9 @@ func (s *composeService) doBuildBake(ctx context.Context, project *types.Project
 		_ = os.Remove(metadataFile)
 	}()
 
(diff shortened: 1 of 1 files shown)


ndeloof (changes requested):
I had a look at the docker/cli pluginManager code:
Actually, `getPlugin` may return a plugin instance and err=nil, but `plugin.Err` gives more details on some missing plugin expectation (typically: failure to get plugin metadata). This `Err` should be checked as well

yangfeiyu20102011:
> I had a look at the docker/cli pluginManager code: Actually, `getPlugin` may return a plugin instance and err=nil, but `plugin.Err` gives more details on some missing plugin expectation (typically: failure to get plugin metadata). This `Err` should be checked as well
OK. How about get and check it in `getBuildxPlugin`. @ndeloof

Issue #13422, opened 2025-12-04
[BUG] COMPOSE_COMPATIBILITY=true ENV not respected anymore (--compatibility is)


systemofapwne:
### Description

I rely on certain scripts like below, that use `COMPOSE_COMPATIBILITY` ENV to fallback to v1 container naming schemes. The scripts may looked likt this

```bash
#!/bin/bash

# Container names as folder_service_instance instead of folder-service-instance
export COMPOSE_COMPATIBILITY=true

#Pull latest images as described in compose.yml
docker compose pull

#Rebuild everything
docker compose up --build -d
```

This used to work like 2 days ago but now seems to be broken with compose v5.

Using `docker compose --compatibility up --build -d` however still produces the v1 names, that I rely on.

### Steps To Reproduce

1. Set ENV: `export COMPOSE_COMPATIBILITY=true`
2. Run `docker compose up --build -d`
3. The container will now be named according to v2 naming scheme (folder-container-1) instead of (folder_container_1)

### Compose Version

```Text
Docker Compose version v5.0.0
```

### Docker Environment

```Text
Client: Docker Engine - Community
 Version:    29.1.2
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.30.1
    Path:     /usr/libexec/docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v5.0.0
    Path:     /usr/libexec/docker/cli-plugins/docker-compose

Server:
 Containers: 74
  Running: 74
  Paused: 0
  Stopped: 0
 Images: 90
 Server Version: 29.1.2
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: syslog
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local splunk syslog
 CDI spec directories:
  /etc/cdi
  /var/run/cdi
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 nvidia runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 1c4457e00facac03ce1d75f7b6777a7a851e5c41
 runc version: v1.3.4-0-gd6d73eb8
[...]

PR #13424, opened 2025-12-05
Restored support for `COMPOSE_COMPATIBILITY`

ndeloof:
**What I did**

**Related issue**
fixes https://github.com/docker/compose/issues/13422

### DIFF
--- a/pkg/compose/loader.go
+++ b/pkg/compose/loader.go
@@ -28,6 +28,7 @@ import (
 
 	"github.com/docker/compose/v5/pkg/api"
 	"github.com/docker/compose/v5/pkg/remote"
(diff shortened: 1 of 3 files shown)


pkg/compose/loader.go
```diff
@@ -48,7 +50,7 @@ func (s *composeService) LoadProject(ctx context.Context, options api.ProjectLoa
 		}
 	}
 
-	if options.Compatibility {
+	if options.Compatibility || utils.StringToBool(projectOptions.Environment["COMPOSE_COMPATIBILITY"]) {
```
ndeloof: this should use ComposeCompatibility const but this would create an import cycle

Issue #13420, opened 2025-12-04
[BUG] Docker compose v5 pull breaks terminal (makes cursor vanish)


scotttag:
### Description

Since upgrading to the new docker compose v5, whenever I run a docker compose pull and it pulls images, it seems to remove the terminal cursor which never comes back until restarting the terminal. Previous versions never did this

Seems to be consistent on multiple servers (both Ubuntu and Debian), remote connecting with Putty over SSH

Additionally, the output text appears to do something weird with the cursor text when combined with echos in a script and makes a mess of the final output (also not the case on v2). If my script echos after successive pulls, the echo overwrites the previous docker compose line

### Steps To Reproduce

Run docker compose pull and cursor will vanish

### Compose Version

```Text
Docker Compose version v5.0.0
```

### Docker Environment

```Text
Client: Docker Engine - Community
 Version:    29.1.2
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.30.1
    Path:     /usr/libexec/docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v5.0.0
    Path:     /usr/libexec/docker/cli-plugins/docker-compose

Server:
 Containers: 13
  Running: 13
  Paused: 0
  Stopped: 0
 Images: 21
 Server Version: 29.1.2
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local splunk syslog
 CDI spec directories:
  /etc/cdi
  /var/run/cdi
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 1c4457e00facac03ce1d75f7b6777a7a851e5c41
 runc version: v1.3.4-0-gd6d73eb8
 init version: de40ad0
 Security Options:
  apparmor
  seccomp
   Profile: builtin
  cgroupns
 Kernel Version: 6.1.0-41-amd64
 Operating System: Debian GNU/Linux 12 (bookworm)
 OSType: linux
 Architecture: x86_64
 CPUs: 4
[...]

PR #13439, opened 2025-12-10
Fix progress UI not restoring terminal once operation completes.

ndeloof:
**What I did**
Run TTY display finalization synchronously on operation end, to enforce terminal is fully restored

**Related issue**
fixes https://github.com/docker/compose/issues/13420


glours (approved):
LGTM

PR #13444, opened 2025-12-11
Add the `--no-build` option to `run`

Cyberax:
**What I did**

`docker compose up` supports `--no-build` but `run` doesn't expose it, even though it has the `--build` option. Fix that omission.


ndeloof (dismissed):

ndeloof (dismissed):

ndeloof:
I'm fine you set a `nolint:gocyclo` annotation on function to workaround linter issue.

ndeloof (approved):

Issue #13448, opened 2025-12-13
SDK packaging: avoid forcing CGO on Darwin for watch-only functionality


rmvangun:
I’m consuming the new Compose Go SDK from `github.com/docker/compose/v5`. My tool does not use file watching, but building on macOS currently forces `CGO_ENABLED=1` due to the watch implementation pulling in an FSEvents-backed watcher.

Would you consider restructuring so that watch is opt-in for SDK consumers?

PR #13452, opened 2025-12-15
Introduced a build tag to select watcher implementation.

ndeloof:
**What I did**

Watcher implementation is selected based on a build tag. This one used to be set by OS, but this brings a constraining requirement to use CGO for users of the compose SDK, even if they don't rely on this feature.
=> Switched to a specific build tag, set by makefile when running on MacOS
Also fixed `GO_BUILDTAGS` not being passed to bake

**Related issue**
closes https://github.com/docker/compose/issues/13448

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,6 +29,10 @@ ifeq ($(DETECTED_OS),Windows)
 	BINARY_EXT=.exe
 endif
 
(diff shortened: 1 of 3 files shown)


glours (approved):
LGTM

PR #13460, opened 2025-12-16
gets back runtime flags when configuring models

ilopezluna:
**What I did**
This pull reverts the changes of the model configuration that disabled `--runtime-args`

**Related issue**
https://github.com/docker/model-runner/issues/515

### DIFF
--- a/pkg/compose/model.go
+++ b/pkg/compose/model.go
@@ -72,7 +72,6 @@ func (s *composeService) ensureModels(ctx context.Context, project *types.Projec
 
 type modelAPI struct {
 	path    string
-	version string // cached plugin version
 	env     []string
 	prepare func(ctx context.Context, cmd *exec.Cmd) error
 	cleanup func()


doringeman (approved):
Thanks!
FWIW this reverts https://github.com/docker/compose/pull/13404.

ndeloof (changes requested):
https://github.com/docker/compose/pull/13404 was introduced for security reasons, restoring support for those flags should only be allowed when the docker-model CLI matches a version that comes with security fix.

pkg/compose/model.go
```diff
@@ -272,3 +273,114 @@ func (m *modelAPI) ListModels(ctx context.Context) ([]string, error) {
 	}
 	return availableModels, nil
 }
+
+// getModelVersion retrieves the docker model CLI version
+func (m *modelAPI) getModelVersion(ctx context.Context) (string, error) {
```
ndeloof: as docker-model is a CLI plugin, better store plugin version obtained by pluginManager in newModelAPI

ndeloof: see https://github.com/docker/compose/pull/13464

pkg/compose/model.go
```diff
@@ -272,3 +273,114 @@ func (m *modelAPI) ListModels(ctx context.Context) ([]string, error) {
...
+	if err != nil {
+		return false
+	}
+
+	return !currentVersion.LessThan(minVersion)
+}
+
+// parseVersion parses a semantic version string
+// Strips build metadata and prerelease suffixes (e.g., "1.0.6-dirty" or "1.0.6+build")
+func parseVersion(versionStr string) (*semVersion, error) {
```
ndeloof: we already use github.com/docker/docker/api/types/versions for comparable purpose checking buildx minimal version

pkg/compose/model.go
ilopezluna: thanks for the example!

pkg/compose/model.go
```diff
@@ -273,3 +276,24 @@ func (m *modelAPI) ListModels(ctx context.Context) ([]string, error) {
...
+	// If version is not cached, don't append runtime flags to be safe
+	if m.version == "" {
+		return false
+	}
+
+	// Strip 'v' prefix if present (e.g., "v1.0.6" -> "1.0.6")
+	versionStr := strings.TrimPrefix(m.version, "v")
+
+	// Strip build metadata or prerelease suffix after "-" or "+"
+	// This is necessary because versions.LessThan treats "1.0.6-dirty" < "1.0.6" per semver rules
```
ndeloof: model-runner 1.0.6 has been released, so is this still relevant ?

ilopezluna: This is helpful for anyone using the Docker Model CLI built locally. I can remove it if you prefer, I just thought it might be useful in some cases. Do you see any downside?

ndeloof: latest tag is v1.0.10 so it doesn't seem this would be useful, even in development mode.

ilopezluna: DMR tag is `v1.0.10`
Docker Model CLI is `v1.0.6` https://github.com/docker/model-runner/releases/tag/cmd%2Fcli%2Fv1.0.6

Anyway, I remove it.

pkg/compose/model.go
```diff
@@ -72,6 +72,7 @@ func (s *composeService) ensureModels(ctx context.Context, project *types.Projec
 
 type modelAPI struct {
 	path    string
+	version string // cached plugin version
```
doringeman:
```suggestion
```

doringeman: To fix https://github.com/docker/compose/actions/runs/20334963688/job/58419170527.
It looks like you can `make lint` locally.

PR #13463, opened 2025-12-17
Add logging while waiting for services

mjagyo:
Introduce a --log flag for docker compose wait and docker compose up --wait to emit progress output while services are starting.

### DIFF
--- a/cmd/compose/up.go
+++ b/cmd/compose/up.go
@@ -58,7 +58,6 @@ type upOptions struct {
 	noAttach              []string
 	timestamp             bool
 	wait                  bool
-	log                   bool
 	waitTimeout           int
 	watch                 bool
 	navigationMenu        bool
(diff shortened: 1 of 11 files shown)


ndeloof (commented):
I don't think we need a new `--log` flag, which would only apply in combination with `--wait`. We can just attach to containers when `--wait` is set; users who prefer to keep it quiet can redirect output, rely on `start --wait` or maybe use `--attach=`.
Doing so you don't need a new flag not new AP attributes, the challenge becomes getting `composeService.up` to stop early, after container all get healthy (which is managed in `start`

mjagyo:
> I don't think we need a new `--log` flag, which would only apply in combination with `--wait`. We can just attach to containers when `--wait` is set; users who prefer to keep it quiet can redirect output, rely on `start --wait` or maybe use `--attach=`. Doing so you don't need a new flag not new AP attributes, the challenge becomes getting `composeService.up` to stop early, after container all get healthy (which is managed in `start`

okay noted, will work on it.

mjagyo:
@ndeloof I think its working now on `up --wait` but was wondering on `wait` command if it should change anything.

mjagyo:
bumping this up @ndeloof again just incase you missed it.

cc: @glours

mbrevda:
hey - any way we can get this reopened? Would really help with our usecase

jclaveau:
It's definitely a pain point in docker compose use in CI

PR #13478, opened 2025-12-25
feat: conditionally warn when no services to build

ismdeep:
- Add COMPOSE_BUILDKIT_WARN_NOBUILD environment variable support
- Only show "No services to build" warning when variable is set to true
- Prevent unnecessary warning output in default behavior


ndeloof:
closing after https://github.com/docker/compose/pull/13493 has been merged

Issue #13484, opened 2025-12-29
[BUG] Unhelpful warning "No services to build" with `docker compose up`


snowdroppe:
### Description

Commit [c89b8a2](https://github.com/docker/compose/commit/c89b8a2d6b4495986b8e8227463895d7a0af680a) added a warning log "No services to build" when `build` is run without any services to build.

This also affects `docker compose up` even if no `build` directive exists within `docker-compose.yml`. Running `up` without the image present in the platform cache (and `pull_policy: missing`, which is default if unset), results in a pull from the `image` directive (if possible) and starts the container(s).

Subsequent runs of `up` with a cached image result in the unhelpful (albeit accurate) warning message "WARN[0000] No services to build".

I imagine this behaviour is due to compose running `build` as a fallback to `pull`, even if the image is cached and `--build` **not** specified. 

Running `docker compose up --no-build` results in no warning as expected, but I would expect this to be default behaviour if `--build` is **not** specified and no `build` directive exists.

Obviously, pre 5.0 behaviour had no such warning (even if the internal calls to `build` were identical).

### Steps To Reproduce

docker-compose.yml:
~~~yaml
services:
  app:
    image: hello-world:latest
~~~

1. `docker compose up` (assuming the image is not cached)
~~~bash
root@host /srv/myapp # docker compose up
[+] up 0/2
[+] up 4/4ello-world:latest [⠀] Pulling                                                                                                                                                                 2.0s
 ✔ Image hello-world:latest Pulled                                                                                                                                                                      2.1s
   ✔ 17eec7bbc9d7           Pull complete                                                                                                                                                               0.5s
 ✔ Network myapp_default    Created                                                                                                                                                                     0.0s
 ✔ Container myapp-app-1    Created                                                                                                                                                                     0.0s
Attaching to app-1
app-1  |
app-1  | Hello from Docker!
app-1  | This message shows that your installation appears to be working correctly.
...
~~~
2. `docker compose up` (assuming the image is cached)
~~~bash
root@host /srv/myapp # docker compose up
WARN[0000] No services to build
Attaching to app-1
app-1  |
app-1  | Hello from Docker!
app-1  | This message shows that your installation appears to be working correctly.
...
~~~

### Compose Version

```Text
Docker Compose version 5.0.1
```

### Docker Environment

```Text
Client:
 Version:    29.1.3
 Context:    default
 Debug Mode: false
 Plugins:
  compose: Docker Compose (Docker Inc.)
    Version:  5.0.1
[...]

PR #13485, opened 2025-12-30
bugfix; change log level of "No services to build" warning log

leo9800:
**What I did**

suppress docker compose log "No services to build" by changing its level from `Warn` to `Debug`.

this situation is rather common when a stack is deployed on production systems with all images pulled from registry and no presense of `build:` block in `docker-compose.yml`. it is not something that needs to be 'warned'.

**Related issue**

#13484


ndeloof:
Warning is legitimate when compose command is `build`. See https://github.com/docker/compose/issues/13484

PR #13488, opened 2025-12-31
refactor(attach): improve attach error handling and cleanup

htoyoda18:
**What I did**
- Improved error handling around `attach` output and stream/tty operations while keeping existing attach behavior.
- `attach`: stopped ignoring the result of `fmt.Fprintf` and log a debug message when writing the "Attaching to ..." line fails.
- `doAttachContainer`: ensured the restore function returned by `attachContainerStreams` is always called (defer restore) to avoid leaving the terminal in a broken state.
- `attachContainerStreams`: stopped ignoring errors from terminal restore and stream closes, and added debug/warn logs for unexpected failures.
- Stream copy goroutines now capture and log non-EOF copy errors.
- Simplified `getContainerStreams` by removing unused variables and returning explicit nils.

**Related issue**
N/A

image

### DIFF
--- a/cmd/compose/compose.go
+++ b/cmd/compose/compose.go
@@ -32,6 +32,7 @@ import (
 	"github.com/compose-spec/compose-go/v2/cli"
 	"github.com/compose-spec/compose-go/v2/dotenv"
 	"github.com/compose-spec/compose-go/v2/loader"
+	composepaths "github.com/compose-spec/compose-go/v2/paths"
 	"github.com/compose-spec/compose-go/v2/types"
 	composegoutils "github.com/compose-spec/compose-go/v2/utils"
 	"github.com/docker/buildx/util/logutil"
@@ -550,11 +551,14 @@ func RootCommand(dockerCli command.Cli, backendOptions *BackendOptions) *cobra.C
 				fmt.Fprint(os.Stderr, aec.Apply("option '--workdir' is DEPRECATED at root level! Please use '--project-directory' instead.\n", aec.RedF))
 			}
 			for i, file := range opts.EnvFiles {
+				file = composepaths.ExpandUser(file)
 				if !filepath.IsAbs(file) {
 					file, err := filepath.Abs(file)
 					if err != nil {
 						return err
 					}
+					opts.EnvFiles[i] = file
+				} else {
 					opts.EnvFiles[i] = file
 				}
 			}
--- a/go.mod
+++ b/go.mod
@@ -37,7 +37,7 @@ require (
 	github.com/opencontainers/go-digest v1.0.0
 	github.com/opencontainers/image-spec v1.1.1
 	github.com/otiai10/copy v1.14.1
-	github.com/sirupsen/logrus v1.9.3
+	github.com/sirupsen/logrus v1.9.4
 	github.com/skratchdot/open-golang v0.0.0-20200116055534-eef842397966
 	github.com/spf13/cobra v1.10.2
 	github.com/spf13/pflag v1.0.10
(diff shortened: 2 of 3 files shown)


pkg/compose/attach.go
```diff
@@ -91,10 +95,16 @@ func (s *composeService) doAttachContainer(ctx context.Context, service, id, nam
 		})
 	})
 
-	_, _, err = s.attachContainerStreams(ctx, id, inspect.Config.Tty, nil, wOut, wErr)
-	return err
+	restore, _, err := s.attachContainerStreams(ctx, id, inspect.Config.Tty, nil, wOut, wErr)
```
ndeloof: as `attachContainerStreams` is only used by `doAttachContainer` it could be refactored to avoid a restore function being returned (this function signature is a legacy from https://github.com/docker/compose/commit/d4a8e92f462fd21446352cd29841c4e9e16257bc).

htoyoda18: Thanks for the review! You're absolutely right that the restore function is legacy and no longer needed.

https://github.com/docker/compose/pull/13488/changes/aad4071032a0cc1e255a34979d1ccf001e264c95

I've refactored `attachContainerStreams` to remove the unused signature elements:
- Removed `stdin` parameter (always passed as `nil` by `doAttachContainer`)
- Changed return type from `(func(), chan bool, error)` to `error`
- Removed the restore function and detached channel logic
- Removed unused imports (`streams`, `term`)

Since `stdin` was always `nil`, the restore function was always a no-op and the detached channel was never used, so this change doesn't affect behavior—it just simplifies the function to match its actual usage.

pkg/compose/attach.go
```diff
@@ -91,66 +93,49 @@ func (s *composeService) doAttachContainer(ctx context.Context, service, id, nam
...
-	if streamIn != nil && stdin != nil {
-		go func() {
-			_, err := io.Copy(streamIn, stdin)
-			var escapeErr term.EscapeError
-			if errors.As(err, &escapeErr) {
-				close(detached)
-			}
-		}()
+func (s *composeService) attachContainerStreams(ctx context.Context, container string, tty bool, stdout, stderr io.WriteCloser) error {
+	_, streamOut, err := s.getContainerStreams(ctx, container)
```
ndeloof: it seems to me getContainerStreams is only used from here, so there's another opportunity to simplify/cleanup code by removing useless attachment to stdin

htoyoda18: Thanks! Simplified `getContainerStreams` to remove the unused stdin:

https://github.com/docker/compose/pull/13488/changes/1036fbbed6bf0d07364acb411590a058bc92f1d9

- Changed signature to return only the output stream: `(io.ReadCloser, error)`
- Set `Stdin: false` in `ContainerAttach` (was `true` but discarded by caller)
- Removed `DetachKeys` (not needed without stdin)

The function now matches its actual read-only usage.

htoyoda18:
@ndeloof 
Would you mind re-running or approving the CI workflow when you have a moment?
I found a small diff in go.mod and pushed the tidy result.

htoyoda18:
Hi @ndeloof, just following up on this.

I previously pushed a small go.mod tidy fix, so the CI should be clean now.
Whenever you have a moment, could you please re-run or approve the CI workflow?

CC @glours in case you have time to take a look as well.
Thanks!

pkg/compose/attach.go
```diff
@@ -91,78 +93,59 @@ func (s *composeService) doAttachContainer(ctx context.Context, service, id, nam
...
-	var stdout io.ReadCloser
-	var stdin io.WriteCloser
+func (s *composeService) getContainerStreams(ctx context.Context, container string) (io.ReadCloser, error) {
 	cnx, err := s.apiClient().ContainerAttach(ctx, container, containerType.AttachOptions{
-		Stream:     true,
-		Stdin:      true,
-		Stdout:     true,
-		Stderr:     true,
-		Logs:       false,
-		DetachKeys: s.configFile().DetachKeys,
```
thaJeztah, line 160: Curious; wouldn't we need the detach-keys? i.e., does this mean it's no longer possible to detach from a container, or is this not used for interactive use?

ndeloof: only used to get output logs. For interactive mode, we rely on docker/cli RunExec and RunAttach

pkg/compose/attach.go
```diff
@@ -49,15 +48,18 @@ func (s *composeService) attach(ctx context.Context, project *types.Project, lis
 		names = append(names, getContainerNameWithoutProject(c))
 	}
 
-	_, _ = fmt.Fprintf(s.stdout(), "Attaching to %s\n", strings.Join(names, ", "))
+	_, err = fmt.Fprintf(s.stdout(), "Attaching to %s\n", strings.Join(names, ", "))
+	if err != nil {
+		logrus.Debugf("failed to write attach message: %v", err)
```
thaJeztah, line 53: Slightly wondering how useful these are, other than "please the linter"; logging here would try to write to `stderr`, which may also fail

PR #13504, opened 2026-01-12
Add unit tests for getDefaultFilters in pkg/compose

htoyoda18:
**What I did**
Added unit tests for the `getDefaultFilters()` function in `pkg/compose/containers.go`.

This function builds Docker API filter combinations and has conditional logic for handling oneOff modes and service selection.

**Related issue**

N/A

### DIFF
--- a/pkg/compose/api_versions.go
+++ b/pkg/compose/api_versions.go
@@ -1,0 +1,73 @@
+/*
+   Copyright 2020 Docker Compose CLI authors
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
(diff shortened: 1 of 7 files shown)


ndeloof:
compose test suite is mostly based on end-to-end tests, so we capture actual result when using the Moby API.

htoyoda18:
> compose test suite is mostly based on end-to-end tests, so we capture actual result when using the Moby API.

Thanks for the clarification! I now understand that the project prefers E2E tests over unit tests for internal helpers to ensure actual behavior via the Moby API.

I'll close this PR for now. I appreciate the guidance and look forward to contributing in a way that better aligns with the project's goals in the future!

Issue #13508, opened 2026-01-13
[BUG] docker compose down --env does not translate ~/.env to .env file in home directory


niemsoen:
### Description

If I docker compose down with an env file as arguments using the tilde notation for the home directory as such:
```
username@hostname:~/some-folder-in-home$ docker compose -f docker-compose.yml --env-file=~/.env down
couldn't find env file: /home/username/some-folder-in-home/~/.env
```
It will print the error `couldn't find env file: /home/username/some-folder-in-home/~/.env`, which indicates that it doesn't translate the tilde symbol to the home directory.

- what it should use: `/home/username/.env`
- what it translates to instead: `/home/username/some-folder-in-home/~/.env`

### Steps To Reproduce

1-2. Start your container with docker compose
3. docker compose it down with an env file that exists in ~/.env like this: 
```
username@hostname:~/some-folder-in-home$ docker compose -f docker-compose.yml --env-file=~/.env down
```
4. See error `couldn't find env file: /home/username/some-folder-in-home/~/.env`

### Compose Version

```Text
docker compose version: Docker Compose version v2.40.0
docker-compose version: bash: docker-compose: command not found
```

### Docker Environment

```Text
Client: Docker Engine - Community
 Version:    28.5.1
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.29.1
    Path:     /usr/libexec/docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.40.0
    Path:     /usr/libexec/docker/cli-plugins/docker-compose

Server:
 Containers: 7
  Running: 7
  Paused: 0
  Stopped: 0
 Images: 25
 Server Version: 27.5.1
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 nvidia runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: b98a3aace656320842a23f4a392a33f46af97866
 runc version: v1.3.0-0-g4ca628d1
[...]

PR #13510, opened 2026-01-14
fix: expand tilde in --env-file paths to user home directory

tensorworkerr:
When using `--env-file=/~.env`, the tilde was not expanded to the user's home directory. Instead, it was treated as a literal character and resolved relative to the current working directory, resulting in errors like "couldn't find env `file: /current/dir/~/.env`".

This adds an ExpandUser function that expands ~ to the home directory before converting relative paths to absolute paths.
Fixes https://github.com/docker/compose/issues/13508

What I did
-Added ExpandUser() function in internal/paths/paths.go that expands a leading ~ to the user's home directory using os.UserHomeDir()
-Updated the env file path processing in cmd/compose/compose.go to call ExpandUser() before checking if paths are absolute
-Added unit tests for the new function covering various edge cases (empty string, ~ only, ~/.env, ~/subdir/.env, absolute paths, relative paths, tilde in middle of path, ~otheruser/... syntax)

Related issue
Fixes https://github.com/docker/compose/issues/13508

Contribution by Gittensor, see my contribution statistics at https://gittensor.io/miners/details?githubId=178221229


internal/paths/paths.go
```diff
@@ -22,6 +22,30 @@ import (
 	"strings"
 )
 
+// ExpandUser expands a leading tilde (~) in a path to the user's home directory.
+// If the path doesn't start with ~, it is returned unchanged.
+// If the home directory cannot be determined, the original path is returned.
+func ExpandUser(path string) string {
```
ndeloof: you can use `paths.ExpandUser` from compose-go

ndeloof:
A better place to fix this is `GetEnvFromFile` in compose-go: [relativePathsResolver](https://github.com/compose-spec/compose-go/blob/9642b85460f2de66e6295c8b1f2e968fc9c3cbae/paths/resolve.go#L100) already manages this for usage within a compose file as file attributes, would make sense to have this logic all in one place

internal/paths/paths.go
tensorworkerr: Updated as you suggested.

thaJeztah:
@ndeloof I don't think this was a bug, and actually the correct behavior; expansion is handled by the shell, and does not happen if the value is prefixed by a `=`; you'll see the same for any tool, e.g. trying with `git`;

```bash
git --git-dir=~/go/src/github.com/docker/compose/.git describe
fatal: not a git repository: '~/go/src/github.com/docker/compose/.git'

git --git-dir ~/go/src/github.com/docker/compose/.git describe
v2.26.1-891-gb760afaf9
```

Bit concerned that this could be a sliding slope.

ndeloof:
@thaJeztah yes I know shell is in charge for this, but as long as we support both space and equal flag values, need to get an homogeneous behavior. compose.yaml already offers this feature

thaJeztah:
Yeah, the issue here would be that it's not homogeneous across the CLI; other commands will _not_ expand these, so now `docker <some command> --some-flag=~/foo` won't work, but `docker compose --some-flag=~/foo` will.

PR #13512, opened 2026-01-14
Fix gw priority network order

### LINKED ISSUE #13511: [BUG] gw_priority is ignored on ipv4 network if service also joins an ipv6 network without priority
### Description

Not sure if this is "as intended", if so, it might worth documenting it.

As mentioned in comments like: 
https://github.com/docker/roadmap/issues/658#issuecomment-2864025303

### Steps To Reproduce

```yaml
networks:
  lan_access:
    name: lan_access
    driver: macvlan
    enable_ipv6: false
    enable_ipv4: true
    driver_opts:
      parent: eth0
    ipam:
      driver: default
      config:
        - subnet: 192.168.10.0/24
          gateway: 192.168.10.1

  databases:
    name: databases
    driver: bridge
    enable_ipv6: true
    enable_ipv4: true

services:
  postgres:
    image: postgres:18.1@sha256:bfe50b2b0ddd9b55eadedd066fe24c7c6fe06626185b73358c480ea37868024d
    networks:
      databases: {}
  home-assistant:
    image: ghcr.io/home-operations/home-assistant:2025.12.5@sha256:6be0843b06ad82233639e56f261b2377f78df140e3d16d2d08b2c29c8889b697
    networks:
      databases: {}
      lan_access:
        ipv4_address: 192.168.10.18
        gw_priority: 1000 
```

`databases` will be the default gw.
Forcing `databases` to disable ipv6 it will make `lan_access` the default gw.

### Compose Version

```Text

```

### Docker Environment

```Text
 Version:    28.3.1
 Context:    default
 Debug Mode: false
 Plugins:
    Version:  v0.25.0
    Version:  v2.38.1

Server:
 Containers: 29
  Running: 29
  Paused: 0
  Stopped: 0
 Images: 32
 Server Version: 28.3.1
 Storage Driver: overlay2
  Backing Filesystem: zfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 2
 Plugins:
  Volume: local
 CDI spec directories:
  /etc/cdi
  /var/run/cdi
 Swarm: inactive
 Runtimes: runc io.containerd.runc.v2 nvidia
 Default Runtime: nvidia
 Init Binary: docker-init
 init version: de40ad0
 Security Options:
  apparmor
  seccomp
   Profile: builtin
  cgroupns
 Kernel Version: 6.12.33-production+truenas
 OSType: linux
 Architecture: x86_64
 CPUs: 16
 Total Memory: 62.63GiB
 Name: truenas
 ID: af416a29-e97e-48c4-a294-35253e26efb6
 Docker Root Dir: /mnt/.ix-apps/docker
 Debug Mode: false
 Experimental: false
 Insecure Registries:
  ::1/128
  127.0.0.0/8
 Live Restore Enabled: false
 Default Address Pools:
   Base: 172.16.0.0/12, Size: 24
```

### Anything else?

_No response_

Pnkcaht:
## What I did

Adjusted how `NetworkingConfig.EndpointsConfig` is built so that networks are ordered by `gw_priority` before being sent to the Docker Engine.

Previously, even when gw_priority was defined in the Compose file, the order of endpoints passed to the Engine was non-deterministic due to map iteration.
This change ensures endpoints are explicitly ordered by descending gw_priority, matching the expected Compose semantics and Docker Engine behavior.

A helper function was introduced in `defaultNetworkSettings` to sort endpoint settings before constructing `NetworkingConfig`, and a unit test was added to validate the ordering.

## Related issue

Fixes #13511

🐳 Carefully routing containers through the right gateway

### Notes
- No behavior change for single-network services
- Only affects multi-network services with `gw_priority` configured
- Covered by unit test

### DIFF
--- a/pkg/compose/convergence_test.go
+++ b/pkg/compose/convergence_test.go
@@ -33,6 +33,7 @@ import (
 	"github.com/stretchr/testify/require"
 	"go.uber.org/mock/gomock"
 	"gotest.tools/v3/assert"
+	"gotest.tools/v3/assert/cmp"
 
 	"github.com/docker/compose/v5/pkg/api"
 	"github.com/docker/compose/v5/pkg/mocks"
@@ -427,7 +428,7 @@ func TestCreateMobyContainer(t *testing.T) {
 	})
 }
 
-func TestDefaultNetworkSettingsOrdersEndpointsByGwPriority(t *testing.T) {
+func TestDefaultNetworkSettingsGwPrioritySelectsPrimaryNetwork(t *testing.T) {
 	project := &types.Project{
 		Name: "test",
 		Networks: map[string]types.NetworkConfig{
@@ -459,16 +460,9 @@ func TestDefaultNetworkSettingsOrdersEndpointsByGwPriority(t *testing.T) {
 	endpoints := netConfig.EndpointsConfig
 	require.Len(t, endpoints, 2)
 
-	keys := make([]string, 0, len(endpoints))
-	for k := range endpoints {
-		keys = append(keys, k)
-	}
+	assert.Assert(t, cmp.Contains(endpoints, "net1"))
+	assert.Assert(t, cmp.Contains(endpoints, "net2"))
 
-	// Highest gw_priority must come first in EndpointsConfig
-	assert.Equal(t, keys[0], "net2")
-	assert.Equal(t, keys[1], "net1")
-
-	// NetworkMode must remain the primary network
-	assert.Equal(t, mode, container.NetworkMode("net1"))
-
+	// gw_priority must only affect primary network selection
+	assert.Assert(t, cmp.Equal(container.NetworkMode("net1"), mode))
 }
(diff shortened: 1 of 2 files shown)


ghost:
## What changed

Hi again @ndeloof , I made the changes based on your request. 👍
This PR removes any reliance on endpoint ordering when handling `gw_priority`.

`EndpointsConfig` is a map and does not guarantee ordering, so using its order to infer gateway priority was incorrect. The logic was updated to use `gw_priority` **only** to select the primary network (`NetworkMode`), without assuming any deterministic order for network endpoints.

The networking configuration now treats endpoint ordering as undefined and only asserts the intended semantic behavior.

## Tests

- Added coverage to ensure `gw_priority` affects **only** primary network selection
- Assert that all expected endpoints are present, without relying on order
- Validate that the selected `NetworkMode` matches the network with the lowest `gw_priority`


image

ndeloof:
so this PR adds a test, changes formatting, but doesn't update the code in any way that would help fix https://github.com/docker/compose/issues/13511 (assuming a fix is needed)
Closing

PR #13516, opened 2026-01-15
Add support for --label flag to docker compose up

### LINKED ISSUE #13436: Add label to all services when starting Compose stack
### Description

The `docker compose run` command has the ability to add a label to the service. It would be nice to have that same ability when launching the entire stack.

For example:

1. I run `docker compose up --label some-key=some-value` (the flag can be used multiple times)
2. All default services are started
3. While inspecting the containers, I see that the label `some-key=some-value` are on all services, networks, and volumes created by the Compose stack.

## Why this is useful

While a Compose stack adds labels for the project name, it would be nice to be able to add arbitrary metadata for those stacks, especially for tooling that might managing multiple stacks (as we are now doing with Labspaces). With this metadata, we can better annotate the stacks to know which stack is which, etc.

Pnkcaht:
### What I did

Added support for the `--label` flag to `docker compose up`, allowing users to add or override labels on all services when starting a Compose stack.

This brings `compose up` in line with existing behavior from other commands (such as `compose run` and `compose create`), making it possible to attach arbitrary metadata to a full Compose stack at runtime.

The provided labels are applied before container creation, ensuring they are correctly propagated to all created containers.

### Related issue

Fixes #13436

🐜 An ant carefully tagging everything in its path — just like Compose now labels all services when bringing a stack up.

### Notes

- Adds `--label` flag support to `docker compose up`
- Allows labeling all services in a stack at startup
- Reuses existing label parsing and application logic
- No behavior change when the flag is not used
- Fully backward compatible

## Screenshot (Test) 

Image

### DIFF
--- a/cmd/compose/create.go
+++ b/cmd/compose/create.go
@@ -49,6 +49,7 @@ type createOptions struct {
 	timeout       int
 	quietPull     bool
 	scale         []string
+	labels        []string // 👈 ADICIONADO
 	AssumeYes     bool
 }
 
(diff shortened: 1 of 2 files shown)


ndeloof:
👎 I'm not favorable to this feature request.

labels in the compose file can be declared with value interpolated. Doing so the compose.yaml has a full definition of services, with all expected labels, just the value is dynamic and can be tweaked by user at runtime.

ghost:
> 👎 I'm not favorable to this feature request.
> 
> labels in the compose file can be declared with value interpolated. Doing so the compose.yaml has a full definition of services, with all expected labels, just the value is dynamic and can be tweaked by user at runtime.

Thank you for your feedback @ndeloof .
I can adjust or discard this if you feel it's not the right direction.

Issue #13149, opened 2025-08-12
[BUG] docker compose ignores update_config


alexey-sh:
### Description

**Docker Compose ignores `failure_action: rollback` in `update_config` and restarts container instead**

### **Description**
When a container fails during an update, Docker Compose ignores the `failure_action: rollback` setting in `update_config` and continuously restarts the failing container instead of performing a rollback to the previous version.

### **Expected Behavior:**
According to the [Compose Deploy Specification](https://docs.docker.com/reference/compose-file/deploy/#update_config), when `failure_action: rollback` is set, Docker Compose should roll back to the previous container version if the update fails.

### **Actual Behavior:**
Docker Compose ignores the rollback configuration and shows the container status as "Restarting" indefinitely, never attempting a rollback.

**Configuration**
```
services:
  app:
    # ... other config
    deploy:
      update_config:
        parallelism: 1
        delay: 10s
        failure_action: rollback  # Expected to rollback on failure, but ignored
```

### **Steps to Reproduce**

1. Clone the example repository: https://github.com/alexey-sh/docker-rollback-update
2. Edit `docker-entrypoint.sh`:
    - Uncomment the line: `if ! npm run success; then`
    - Comment out the line: `if ! npm run test; then`
3. Run: `docker compose up --build -d`
4. Revert the changes made in step 2
5. Run: `docker compose up --build -d` (this should trigger the failure scenario)
6. Check container status: `docker ps | grep roll`

**Result**: Container shows "Restarting" status instead of performing a rollback

### **Environment:**

Docker version 28.2.2, build e6534b4
OS Ubuntu 24.04.2 LTS

### **Additional Context:**

The [Docker Compose 1.22.0 release notes](https://docs.docker.com/compose/releases/release-notes/#1220) mention support for update_config parameters
The failure_action parameter is documented in the official Compose file reference
This suggests the feature should be supported but may not be implemented correctly

### **Questions**

Is `failure_action: rollback` actually supported in Docker Compose, or is this a Swarm-only feature?
If supported, is this a known issue with a planned fix?
Are there any workarounds to achieve rollback behavior in Docker Compose?

### Compose Version

```Text
Docker Compose version v2.36.2
```

### Docker Environment

```Text
Client: Docker Engine - Community
 Version:    28.2.2
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.24.0
[...]

PR #13517, opened 2026-01-15
Warn when unsupported deploy attribute is used in docker compose

### LINKED ISSUE #13150: [BUG] Compose don't warn user for unsupported attributes
### Description

As mentioned in Compose Specification the runtime is in charge of warn the users about the unsupported attributes. 

We should also reference all those attributes in the official documentation

### Steps To Reproduce

See #13149 for example

### Compose Version

```Text
N/A
```

### Docker Environment

```Text
N/A
```

### Anything else?

_No response_

Pnkcaht:
## What I did

Added a warning when unsupported Compose attributes are detected during project
loading. Currently this warns when the `deploy` section is used, which is
ignored by docker compose outside of Swarm mode.

This avoids silent misconfiguration and makes it clear to users why `deploy`
settings are not applied when running docker compose outside of Swarm mode.

## Related issue

Fixes #13150  
Related to #13149

🐶 A vigilant watchdog barking loudly when something unsupported sneaks in — just like Compose now warns users when `deploy` is defined outside of Swarm mode.

###  Notes

- Emits a warning when the `deploy` attribute is present in a service definition
- Does not change runtime behavior; only adds user-facing feedback
- Keeps backward compatibility
- Introduces a small, dedicated `warnings.go` helper to keep the logic isolated
- Verified with local build and tests:
  - `go build ./cmd/compose`
  - `go test ./cmd/compose/...`

## Screenshot

_Build and tests passing locally_

image

### DIFF
--- a/cmd/compose/warnings.go
+++ b/cmd/compose/warnings.go
@@ -1,0 +1,86 @@
+package compose
+
+import (
+	"sort"
+	"strings"
+
+	"github.com/compose-spec/compose-go/v2/types"
+	"github.com/sirupsen/logrus"
+)
+
+// warnIgnoredDeployAttributes emits informational warnings for deploy
+// attributes that are ignored by Docker Compose when running in
+// standalone (non-Swarm) mode.
+//
+// Note that some deploy attributes are still considered in standalone
+// mode, most notably `replicas` and device reservations (e.g. GPU requests).
+func warnIgnoredDeployAttributes(project *types.Project) {
+	for _, svc := range project.Services {
+		if svc.Deploy == nil {
+			continue
+		}
+
+		ignored := collectIgnoredDeployAttributes(svc.Deploy)
+		if len(ignored) == 0 {
+			continue
+		}
+
+		sort.Strings(ignored)
+
+		logrus.Warnf(
+			"Service %q uses deploy attributes that are ignored by Docker Compose "+
+				"in standalone (non-Swarm) mode. Some deploy attributes are still considered "+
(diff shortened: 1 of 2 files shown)


cmd/compose/warnings.go
```diff
@@ -0,0 +1,17 @@
...
+import (
+	"github.com/compose-spec/compose-go/v2/types"
+	"github.com/sirupsen/logrus"
+)
+
+func warnUnsupportedAttributes(project *types.Project) {
+	for _, service := range project.Services {
+		if service.Deploy != nil {
+			logrus.Warnf(
+				"Service %q uses 'deploy' which is not supported by docker compose and will be ignored",
```
ndeloof: that's wrong. _most_ of deploy section is ignored by compose, but some attributes are considered. Typically replica and some device requests (used for gpu support)

ghost: ok

ghost:
## Summary
@ndeloof 
Add a targeted warning for deploy attributes that are consistently ignored by Docker Compose in standalone mode.

## Details

- Detects ignored deploy subfields (placement, update/rollback config, endpoint mode, mode)

- Avoids struct comparison issues by checking meaningful subfields only

- Explicitly excludes supported or fallback attributes (replicas, device reservations, restart_policy)

- Emits a single, clear warning per service with a sorted attribute list

## Impact

- Improves user feedback and clarity

- No behavior or runtime changes

- Warning-only, non-breaking improvement

Issue #12637, opened 2025-03-14
[BUG] False "variable is not set" warning when using the ${A:+} interpolation syntax


SpecLad:
### Description

When a variable interpolation is nested inside a `${A:+}`-style interpolation construct for the same variable, `compose` emits a "variable is not set" warning, even though the nested evaluation isn't supposed to be evaluated.

### Steps To Reproduce

1. Create `compose.yml` as follows:

   ```yaml
   services:
     test:
       environment:
         A: ${A:+${A},}B
       image: busybox
       command: ['sh', '-c', 'echo "$$A"']
   ```

2. Run `docker compose up`. It prints the following:

   ```
   WARN[0000] The "A" variable is not set. Defaulting to a blank string.
   [+] Running 1/1
    ✔ Container docker-compose-repro-test-1  Recreated                                                                0.1s
   Attaching to test-1
   test-1  | B
   test-1 exited with code 0
   ```

The warning is a false positive: the nested `${A},` expression should only be evaluated if `A` is defined, which it isn't.

### Compose Version

```Text
Docker Compose version v2.33.1
```

### Docker Environment

```Text
Client: Docker Engine - Community
 Version:    28.0.1
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.21.1
    Path:     /usr/libexec/docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.33.1
    Path:     /usr/libexec/docker/cli-plugins/docker-compose
  scout: Docker Scout (Docker Inc.)
    Version:  v1.14.0
    Path:     /home/dpb/.docker/cli-plugins/docker-scout

Server:
 Containers: 3
  Running: 1
  Paused: 0
  Stopped: 2
 Images: 242
 Server Version: 28.0.1
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 1
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 runc
 Default Runtime: runc
 Init Binary: docker-init
[...]

PR #13520, opened 2026-01-15
fix: avoid eager interpolation warnings

seekskyworld:
## Issue
- https://github.com/docker/compose/issues/12637

## Summary
- use compose-go merged fix for lazy interpolation by updating dependency
- drop replace to forked compose-go

## Motivation
- keep compose file loading/processing in compose-go per maintainer guidance

## Validation
- go test ./cmd/compose -run TestApplyPlatforms_InferFromRuntime -count=1

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,7 +8,7 @@ require (
 	github.com/Microsoft/go-winio v0.6.2
 	github.com/acarl005/stripansi v0.0.0-20180116102854-5a71ef0e047d
 	github.com/buger/goterm v1.0.4
-	github.com/compose-spec/compose-go/v2 v2.10.0
+	github.com/compose-spec/compose-go/v2 v2.10.1-0.20260116073713-ddb94f10f3a0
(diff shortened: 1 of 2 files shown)


ndeloof (changes requested):
compose file loading and processing should be implemented in compose-go

seekskyworld:
@ndeloof Updated the PR to align with your guidance: removed the compose-local workaround and now rely on the compose-go fix (compose-spec/compose-go#847) via a temporary go.mod replace. Once the compose-go PR lands, I can switch this to a clean version bump. Tests: go test ./... .

seekskyworld:
@ndeloof Thanks! I’ve replaced the temporary fork/replace with the updated compose-go merge-commit version and removed the replace.

ndeloof:
please squash commits then we can merge

seekskyworld:
@ndeloof Squashed into a single signed-off commit and force-pushed. Thanks!

PR #13529, opened 2026-01-19
emit container status events after network reconnection (fixes #13524)

### LINKED ISSUE #13524: [BUG] misleading status for non-host-mode containers
### Description

I've noticed something odd but I don't know whether it's actually a bug or the intended behaviour.

Consider this `diff`:

```
$ diff docker-compose.override.yml.save docker-compose.override.yml
4a5,14
>   default:
>     ipam:
>       config:
>         - subnet: 172.30.0.0/22
> 
>   nextcloud:
>     ipam:
>       config:
>         - subnet: 172.30.4.0/22
> 
```

In other words, `docker-compose.override.yml` defines specific subnets for two internal networks whereas the `.save` file does not.

The situation with those subnet definitions in force:

```
$ docker ps
CONTAINER ID   IMAGE                       COMMAND                  CREATED         STATUS                   PORTS                    NAMES
a3d379eb6ba7   koenkk/zigbee2mqtt:latest   "docker-entrypoint.s…"   2 minutes ago   Up 2 minutes             0.0.0.0:8080->8080/tcp   zigbee2mqtt
c4e47e0f236d   iotstack-nodered            "./entrypoint.sh"        2 minutes ago   Up 2 minutes (healthy)   0.0.0.0:1880->1880/tcp   nodered
c484a9c1b5e9   grafana/grafana             "/run.sh"                2 minutes ago   Up 2 minutes (healthy)   0.0.0.0:3000->3000/tcp   grafana
ff51c96674a3   esphome/esphome             "/entrypoint.sh dash…"   2 minutes ago   Up 2 minutes (healthy)                            esphome
f127b0a10b85   iotstack-mosquitto          "/docker-entrypoint.…"   2 minutes ago   Up 2 minutes (healthy)   0.0.0.0:1883->1883/tcp   mosquitto
d8d58efc1c9d   influxdb:1.12               "/entrypoint.sh infl…"   2 minutes ago   Up 2 minutes (healthy)   0.0.0.0:8086->8086/tcp   influxdb
```

Now let's change the override file to be the one **without** the subnet definitions:

```
$ mv docker-compose.override.yml.save docker-compose.override.yml
```

Let's apply that:

```
$ docker compose up -d
WARN[0000] No services to build                         
[+] up 7/7
 ✔ Container nodered        Stopped 0.7s 
 ✔ Container zigbee2mqtt    Stopped 9.6s 
 ✔ Container grafana        Stopped 0.7s 
 ✔ Container influxdb       Stopped 1.0s 
 ✔ Container mosquitto      Stopped 0.4s 
 ✔ Network iotstack_default Created 0.1s 
 ✔ Container esphome        Running 0.0s
```

Compose says it has stopped five containers. The true situation:

``` 
$ docker ps
CONTAINER ID   IMAGE                       COMMAND                  CREATED         STATUS                             PORTS                    NAMES
a3d379eb6ba7   koenkk/zigbee2mqtt:latest   "docker-entrypoint.s…"   3 minutes ago   Up 12 seconds                      0.0.0.0:8080->8080/tcp   zigbee2mqtt
c4e47e0f236d   iotstack-nodered            "./entrypoint.sh"        3 minutes ago   Up 12 seconds (health: starting)   0.0.0.0:1880->1880/tcp   nodered
c484a9c1b5e9   grafana/grafana             "/run.sh"                3 minutes ago   Up 12 seconds (healthy)            0.0.0.0:3000->3000/tcp   grafana
ff51c96674a3   esphome/esphome             "/entrypoint.sh dash…"   3 minutes ago   Up 3 minutes (healthy)                                      esphome
f127b0a10b85   iotstack-mosquitto          "/docker-entrypoint.…"   3 minutes ago   Up 12 seconds (healthy)            0.0.0.0:1883->1883/tcp   mosquitto
d8d58efc1c9d   influxdb:1.12               "/entrypoint.sh infl…"   3 minutes ago   Up 12 seconds (health: starting)   0.0.0.0:8086->8086/tcp   influxdb
```

To summarise:

1. We started with 6 containers running (esphome, grafana, influxdb, mosquitto, nodered and zigbee2mqtt).
2. We finished with the same six containers running.
3. After changing the override file, all bar esphome are reported as having been stopped. esphome being the exception is expected because that container runs in host mode so it isn't affected by internal network changes.
4. What's odd is that docker compose hasn't reported **starting** the stopped containers again, yet they're all clearly running.

Compare and contrast what happens if I pick a container at random (eg grafana) and add some dummy environment variable to its service definition. On the subsequent "up", I'll get:

```
 ✔ Container grafana     Recreated 0.5s 
```

In short, the expected behaviour is that all five non-host-mode containers would be reported as **either** "Recreated" **or** "Stopped" then "Created", rather than just "Stopped" which isn't actually true.

For the record, this pattern is the same in each direction. Going from "no subnets defined" to "subnets defined" has the same behaviour. It also doesn't matter whether the subnet definition occurs in the override file or the compose file. I just happen to be using the override file.

I hope that makes sense.

### Steps To Reproduce

Implied in the above. In a compose file with a `networks:` section plus at least one non-host-mode container:

* either:

	1. Compose or override file **without** defined subnets.
	2. Down stack
	3. Up stack.
	3. Add subnet definitions to compose or override file.
	4. Up the stack.

* or:

	1. Compose or override file **with** defined subnets.
	2. Down stack
	3. Up stack.
	3. Remove subnet definitions from compose or override file.
	4. Up the stack.

### Compose Version

```Text
$ docker compose version
Docker Compose version v5.0.1

$ docker-compose version
Docker Compose version v5.0.1
```

### Docker Environment

```Text
 Version:    29.1.3
 Context:    default
 Debug Mode: false
 Plugins:
    Version:  v0.30.1
    Version:  v5.0.1

Server:
 Containers: 6
  Running: 6
  Paused: 0
  Stopped: 0
 Images: 7
 Server Version: 29.1.3
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: local
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
 CDI spec directories:
  /etc/cdi
  /var/run/cdi
 Swarm: inactive
 Default Runtime: runc
 Init Binary: docker-init
 init version: de40ad0
 Security Options:
  seccomp
   Profile: builtin
  cgroupns
 Kernel Version: 6.1.21-v8+
 OSType: linux
 Architecture: aarch64
 CPUs: 4
 Total Memory: 3.704GiB
 Name: sec-dev
 Debug Mode: false
 Experimental: false
 Insecure Registries:
  127.0.0.0/8
  ::1/128
 Live Restore Enabled: false
 Firewall Backend: iptables
```

### Anything else?

_No response_

MaheshThakur9152:
Summary
- When a network configuration (for example IPAM/subnet) changes, Compose stops containers and recreates the network.
- The Docker engine can restart those containers automatically (e.g., due to a restart policy). Compose previously did not emit any event representing the containers' resumed state after reconnect, so the TTY UI could be left showing only "Stopped" even though the containers were actually running.
- This PR makes Compose inspect containers after reconnecting them and emits the correct container status (Running / Created / Stopped), so the UI reflects the real state.

Root cause
- The code path that handles diverged networks stops and disconnects containers, removes the network, recreates it and then reconnects containers.
- If the engine restarts those containers between disconnection and reconnection, Compose had no further event for them, so the TTY progress UI only displayed the earlier `Stopped`.

What I changed
- After reconnecting previously-dangled containers, inspect each container and emit one of the following events depending on the container state:
  - Running -> `Running` event
  - Created -> `Created` event
  - Exited -> `Stopped` event
- Ignore inspect errors to avoid failing network creation due to races where a container disappears between disconnect and reconnect.
- Add unit test `TestResolveOrCreateNetworkEmitsRunningEvent` to assert that a running event is emitted for a container restarted by the engine.

Files changed
- create.go — emits container status events after reconnecting containers
- create_test.go — new unit test simulating the diverged network flow and the engine restarting a container

Testing done
- Unit test: go test ./pkg/compose -run TestResolveOrCreateNetworkEmitsRunningEvent -v — passes locally.
- Full test suite: will run in CI (needs Docker daemon access and runs in their environments).

Notes for reviewers
- Small, focused change to event emission — no new event types or UI changes.
- Emitted events use existing statuses (Running, Created, Stopped) so the current UI picks them up.
- If maintainers prefer a different message (e.g., “Reconnected”), I’m happy to adjust.
- Signed-off-by: Mahesh Thakur <maheshthakur9152@gmail.com>

Fixes: docker/compose#13524

### DIFF
--- a/cmd/compose/compose.go
+++ b/cmd/compose/compose.go
@@ -505,6 +505,7 @@ func RootCommand(dockerCli command.Cli, backendOptions *BackendOptions) *cobra.C
 				display.Mode = display.ModeTTY
 			}
 
+			detached, _ := cmd.Flags().GetBool("detach")
 			var ep api.EventProcessor
 			switch opts.Progress {
 			case "", display.ModeAuto:
@@ -513,7 +514,7 @@ func RootCommand(dockerCli command.Cli, backendOptions *BackendOptions) *cobra.C
 					display.Mode = display.ModePlain
 					ep = display.Plain(dockerCli.Err())
 				case dockerCli.Out().IsTerminal():
-					ep = display.Full(dockerCli.Err(), stdinfo(dockerCli))
+					ep = display.Full(dockerCli.Err(), stdinfo(dockerCli), detached)
 				default:
 					ep = display.Plain(dockerCli.Err())
 				}
@@ -522,7 +523,7 @@ func RootCommand(dockerCli command.Cli, backendOptions *BackendOptions) *cobra.C
 					return fmt.Errorf("can't use --progress tty while ANSI support is disabled")
 				}
 				display.Mode = display.ModeTTY
-				ep = display.Full(dockerCli.Err(), stdinfo(dockerCli))
+				ep = display.Full(dockerCli.Err(), stdinfo(dockerCli), detached)
 
 			case display.ModePlain:
 				if ansi == "always" {
--- a/cmd/display/tty.go
+++ b/cmd/display/tty.go
@@ -37,13 +37,14 @@ import (
 
 // Full creates an EventProcessor that render advanced UI within a terminal.
 // On Start, TUI lists task with a progress timer
-func Full(out io.Writer, info io.Writer) api.EventProcessor {
+func Full(out io.Writer, info io.Writer, detached bool) api.EventProcessor {
 	return &ttyWriter{
-		out:   out,
-		info:  info,
-		tasks: map[string]*task{},
-		done:  make(chan bool),
-		mtx:   &sync.Mutex{},
+		out:      out,
+		info:     info,
+		tasks:    map[string]*task{},
+		done:     make(chan bool),
+		mtx:      &sync.Mutex{},
+		detached: detached,
 	}
 }
 
@@ -60,6 +61,7 @@ type ttyWriter struct {
 	ticker    *time.Ticker
 	suspended bool
 	info      io.Writer
+	detached  bool
 }
 
 type task struct {
@@ -190,7 +192,7 @@ func (w *ttyWriter) On(events ...api.Resource) {
 			continue
 		}
 
-		if w.operation != "start" && (e.Text == api.StatusStarted || e.Text == api.StatusStarting) {
+		if w.operation != "start" && (e.Text == api.StatusStarted || e.Text == api.StatusStarting) && !w.detached {
 			// skip those events to avoid mix with container logs
 			continue
 		}
--- a/pkg/e2e/networks_test.go
+++ b/pkg/e2e/networks_test.go
@@ -212,11 +212,9 @@ func TestNetworkRecreate(t *testing.T) {
 	res := c.RunDockerComposeCmd(t, "-f", "./fixtures/network-recreate/compose.yaml", "--project-name", projectName, "--progress=plain", "up", "-d")
 	err := res.Stderr()
 	fmt.Println(err)
-	res.Assert(t, icmd.Expected{Err: `
- Container network_recreate-web-1 Stopped 
- Network network_recreate_test Removed 
- Network network_recreate_test Creating 
- Network network_recreate_test Created 
- Container network_recreate-web-1 Starting 
- Container network_recreate-web-1 Started`})
+	hasStopped := strings.Contains(err, "Stopped")
+	hasResumed := strings.Contains(err, "Started") || strings.Contains(err, "Recreated")
+	if !hasStopped || !hasResumed {
+		t.Fatalf("unexpected output, missing expected events, stderr: %s", err)
+	}
 }


pkg/compose/create.go
```diff
@@ -1420,6 +1420,23 @@ func (s *composeService) resolveOrCreateNetwork(ctx context.Context, project *ty
...
 	}
 
+	// Emit container status events for containers that may have been restarted by the engine
+	for _, c := range dangledContainers {
+		inspected, err := s.apiClient().ContainerInspect(ctx, c.ID)
+		if err != nil {
+			// ignore inspect errors
+			continue
+		}
+		name := "Container " + strings.TrimPrefix(inspected.Name, "/")
```
ndeloof: use getCanonicalContainerName for consistency with other places we produce events

pkg/compose/create_test.go
```diff
@@ -455,3 +455,75 @@ volumes:
 		})
 	}
 }
+
+func TestResolveOrCreateNetworkEmitsRunningEvent(t *testing.T) {
```
ndeloof: We prefer use of e2e tests as mocks are a pain to maintain with many API calls, also in most case we don't get the actual engine behavior demonstrated by such tests.

MaheshThakur9152: Hi — thanks for the review and the suggestions!

I implemented both fixes you pointed out:
- Switched to using the canonical container name (now using `getContainerProgressName(tmp)`) so emitted event names match other code paths.
- Fixed the state check to consistently use the `Status` field (`inspected.State.Status == container.StateRunning`), consistent with other code.

Per your preference for e2e tests, I added an e2e test:
- `TestNetworkIPAMChangeReportsStarted` changes the subnet and runs `docker compose up --progress=plain`, then asserts the output contains the expected container state transition (Stopped then Started or Recreated).

I left the focused unit test as an additional guard; if you prefer to rely only on e2e tests I can remove the unit test in a follow-up. Happy to tweak wording or add more coverage if you prefer.

— Mahesh

ndeloof:
docker engine to automatically restart container is a corner case. The actual issue on https://github.com/docker/compose/issues/13524 is Compose not to log an event after a container is (re)started.

MaheshThakur9152:
> docker engine to automatically restart container is a corner case. The actual issue on #13524 is Compose not to log an event after a container is (re)started.

Ack, understood. The auto-restart is just the trigger, but the missing event log is the gap in Compose. 

This PR addresses that root cause: it inspects the container state after network reconciliation and emits the missing status event.

I've updated the logic and added the E2E test to cover this. Ready for review.

ndeloof:
> This PR addresses that root cause

it doesn't. It just emits an additional event, while this one should be managed by `start`

MaheshThakur9152:
> > This PR addresses that root cause
> 
> it doesn't. It just emits an additional event, while this one should be managed by `start`

Got it. So instead of manually emitting the event post-reconnect, we should ensure the standard start lifecycle handles this transition?

Could you point me to the specific start method/flow you are referring to? I want to make sure I hook into the correct lifecycle stage rather than patching it here.

ndeloof:
this is intentional: https://github.com/docker/compose/blob/c8d687599a425502116dac1de9c55d8ecc52bb77/cmd/display/tty.go#L193

if we don't, and compose run attached, the progress UI will collides with first logs
A fix could be to let the `Full` (tty.go) display be aware we run detached, and not hide events.

MaheshThakur9152:
> this is intentional:
> 
> https://github.com/docker/compose/blob/c8d687599a425502116dac1de9c55d8ecc52bb77/cmd/display/tty.go#L193
> 
> if we don't, and compose run attached, the progress UI will collides with first logs A fix could be to let the `Full` (tty.go) display be aware we run detached, and not hide events.


> Updated! I've reverted the changes in `create.go` and moved the logic to `tty.go` as suggested. The display now allows `StatusStarted` events specifically when running in detached mode.
> Validated with the E2E test. Ready for re-review.

ndeloof (approved):

cmd/compose/compose.go
```diff
@@ -505,6 +505,7 @@ func RootCommand(dockerCli command.Cli, backendOptions *BackendOptions) *cobra.C
 				display.Mode = display.ModeTTY
 			}
 
+			detached, _ := cmd.Flags().GetBool("detach")
```
ndeloof, line 508: I'm surprised this works, as `detach` flag is declared on child command

Makefile
```diff
@@ -29,6 +29,10 @@ ifeq ($(DETECTED_OS),Windows)
 	BINARY_EXT=.exe
 endif
 
+ifeq ($(DETECTED_OS),Darwin)
```
ndeloof: this was intentionnaly removed from main branch, now managed within Dockerfile

MaheshThakur9152: > this was intentionnaly removed from main branch, now managed within Dockerfile

My apologies, that was a remnant from my local dev environment on macOS. I've reverted the Makefile changes. Thanks for catching that

ndeloof (approved):

PR #13532, opened 2026-01-20
set fsnotify build tag when building for OSX

ndeloof:
**What I did**

follow-up https://github.com/docker/compose/pull/13452
during release, we use cross-compilation with GHA bake action. This updates the Dockerfile to enable fsnotify build tag when we target OSX binaries

**Related issue**
fixes https://github.com/docker/for-mac/issues/7832


ndeloof:
```
$ strings bin/build/darwin_arm64/docker-compose |grep -c fsevent
128
```

PR #13538, opened 2026-01-21
fix: warn on Windows when Git Bash mangles volume paths

MaheshThakur9152:
## Summary
This PR adds a heuristic to detect a common issue for Windows users running Docker Compose via Git Bash (MinGW). MinGW often applies "POSIX-to-Windows" path conversion to the `-v` volume arguments, breaking the mount syntax (e.g., converting the container path `/etc/...` to `C:/Program Files/Git/etc/...`).

## The Fix
I added a check in `pkg/compose/create.go` inside `buildMount`.
If `runtime.GOOS == "windows"` and the mount looks mangled (Source contains `;` OR Target points to `Program Files/Git`), we emit a `logrus.Warn` with actionable advice:
> "Warning: It looks like Git Bash mangled your volume path. Please use '//' at the start of the container path (e.g., //etc/...) or set MSYS_NO_PATHCONV=1 to avoid this."

## Testing
* Added `pkg/compose/create_mingw_test_windows.go` (Build tagged for Windows only).
* Validated that the warning triggers when the heuristic conditions are met.

### DIFF
--- a/pkg/compose/create.go
+++ b/pkg/compose/create.go
@@ -26,6 +26,7 @@ import (
 	"path/filepath"
 	"slices"
 	"strconv"
+	"runtime"
 	"strings"
 
 	"github.com/compose-spec/compose-go/v2/paths"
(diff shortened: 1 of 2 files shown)


ndeloof:
There's no active issue reporting this issue
IMHO this isn't Compose role to manage this. AFAICT docker/cli has the same issue

PR #13548, opened 2026-01-23
feat: Add --skip-validation flag to docker compose alpha publish

### LINKED ISSUE #13539: Option to skip validation during publishing
### Description

There are certain Compose files that I would like to publish that are intended to overlay another set of Compose files. However, I want to publish each of them separately to allow the end user to pick and choose versions, etc.

However, when publishing one of these overlay files, I end up with validation failing because the Compose file isn't complete (since it's merely an overlay/override). I get something like this:

```
service "workspace" has neither an image nor a build context specified: invalid compose project
```

It would be nice if I could publish a partial file by informing Compose I want to skip validation. Something like...

```
docker compose publish --skip-validation my-namespace/my-repo
```

For the context/use case - we have the main Labspace Compose files and are creating various extensions to them to add other services. We'd like to publish each of them independently.

amyssnippet:
## What I did

Added a `--skip-validation` flag to the `docker compose alpha publish` command to allow publishing partial Compose files (overlays) that don't contain complete service definitions.

### Changes Made:
- Added `skipValidation` option to `publishOptions` struct in `cmd/compose/publish.go`
- Added `--skip-validation` command-line flag with appropriate help text
- Extended `api.ProjectLoadOptions` and `api.PublishOptions` to support skip validation
- Modified project loading to use `cli.WithSkipValidation()` when flag is enabled
- Updated publish logic in `pkg/compose/publish.go` to skip image push for services without images when validation is disabled
- Enhanced compose-go library (vendor) to support `WithSkipValidation` functionality

### Key Features:
- Skips compose project validation during loading when `--skip-validation` is used
- Allows publishing services that don't have `image` or `build` specifications
- Maintains backward compatibility (validation enabled by default)
- Provides clear error messages when validation is required but missing

## Related issue

Resolves #13539: Option to skip validation during publishing

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -56,6 +56,7 @@ require (
 	golang.org/x/sync v0.19.0
 	golang.org/x/sys v0.40.0
 	google.golang.org/grpc v1.78.0
+	gopkg.in/yaml.v3 v3.0.1
 	gotest.tools/v3 v3.5.2
 	tags.cncf.io/container-device-interface v1.1.0
 )
@@ -153,5 +154,4 @@ require (
 	google.golang.org/genproto/googleapis/rpc v0.0.0-20251029180050-ab9386a59fda // indirect
 	google.golang.org/protobuf v1.36.10 // indirect
 	gopkg.in/ini.v1 v1.67.0 // indirect
-	gopkg.in/yaml.v3 v3.0.1 // indirect
 )


amyssnippet:
The changes allow the user to skip the validation tests of the compose file structure.

[Screencast from 2026-01-23 12-58-45.webm](https://github.com/user-attachments/assets/d0d408e8-cd8b-41b2-a7ac-88691fc4f708)

amyssnippet:
the `test_overlay` folder is just made for test with some sample empty compose file for testing the command after the new build of the compose, ie the skip validation flag.

For safe testing the skip validation, it is mandatory to use the dry run mode to prevent accidental pushes to real registries during development.

ndeloof:
see https://github.com/docker/compose/issues/13539#issuecomment-3788820504

tl;dr: dont just on all possible feature request to prove your AI agent can deal with compose codebase, many of those never will get implemented. Better confirm with maintainers which one deserve some time for investigation and design discussions

PR #13549, opened 2026-01-23
compose: recreate container when mounted image digest changes

### LINKED ISSUE #13547: [BUG] Image mount holds on to stale image even once image is removed
### Description

Mounting an image which has been built in the same stack does not work as expected: rebuilds of the image do not cause containers which mount it to be recreated.

Since https://github.com/docker/compose/pull/12725 and https://github.com/moby/moby/pull/48798 it's possible to mount files from images. Docker warns it's still an experimental feature and there are open issues (e.g. https://github.com/moby/moby/issues/50999). So some problems are to be expected.

In my test case I expect the output from both containers to be the same, both the first time and the second time they run. However, the second time the output from container b is the stale "foo" content. For some reason container b still mounts the old image, which shouldn't exist any more: I explicitly pruned builder and images and the images list only outputs bash and the new a image.

Once you add `--force-recreate` to the second `docker compose up` call, the output of both containers is correct: they now both output "bar". However I don't want to add `--force-recreate` as it impacts all other services in the stack which may be unrelated to image `a`.

### Steps To Reproduce

```yaml
services:
  a:
    container_name: ca
    build:
      context: .
      dockerfile: Dockerfile
    image: a
    command: ["cat", "/a/test.txt"]
  b:
    container_name: cb
    image: bash
    depends_on:
      - a
    command: ["cat", "/a/test.txt"]
    volumes:
      - type: image
        source: a
        target: /a
        image:
          subpath: a
```

```sh
docker builder prune -a -f > /dev/null
docker image prune -a -f > /dev/null
echo "FROM bash\nWORKDIR /a\nRUN echo foo > /a/test.txt" > Dockerfile
docker compose --progress=plain up --quiet-build --quiet-pull --pull=always --build
docker builder prune -a -f > /dev/null
docker image prune -a -f > /dev/null
echo "FROM bash\nWORKDIR /a\nRUN echo bar > /a/test.txt" > Dockerfile
docker compose --progress=plain up --quiet-build --quiet-pull --pull=always --build
docker compose down
docker images -a --digests
```

Output:

```sh
 Image a Pulling 
 Image bash Pulling 
 Image a Error pull access denied for a, repository does not exist or may require 'docker login'
 Image bash Pulled 
 Image a Building 
 Image a Built 
 Network test_default Creating 
 Network test_default Created 
 Container ca Creating 
 Container ca Created 
 Container cb Creating 
 b Image mount is an experimental feature 
 Container cb Created 
Attaching to ca, cb
 Container ca Starting 
ca  | foo
 Container ca Started 
ca exited with code 0
 Container cb Starting 
cb  | foo
cb exited with code 0
 Container cb Started 
 Image bash Pulling 
 Image bash Pulled 
 Image a Building 
 Image a Built 
 Container ca Recreate 
 Container ca Recreated 
Attaching to ca, cb
 Container ca Starting 
ca  | bar
 Container ca Started 
 Container cb Starting 
ca exited with code 0
cb  | foo
cb exited with code 0
 Container cb Started 
[+] down 3/3
 ✔ Container cb            Removed                                                                                                                                                                                                                                                                          0.0s 
 ✔ Container ca            Removed                                                                                                                                                                                                                                                                          0.0s 
 ✔ Network test_default Removed                                                                                                                                                                                                                                                                          0.2s 
REPOSITORY   TAG       DIGEST                                                                    IMAGE ID       CREATED        SIZE
a            latest    sha256:e431715ba01e4d236e411af2f5ba4c690d6b6b40e172f3d94c5b15dbcd3ce540   e431715ba01e   1 second ago   26.6MB
bash         latest    sha256:32ca1653c68dd414bbdcdb9d5dc811010022b04e37bfcc886a5d052fc0c56656   32ca1653c68d   5 weeks ago    26.7MB
```

### Compose Version

```Text
Docker Compose version v5.0.1
```

### Docker Environment

```Text
Client:
 Version:    29.1.3
 Context:    desktop-linux
 Debug Mode: false
 Plugins:
    Version:  v1.17.1
    Path:     /Users/docker/.docker/cli-plugins/docker-ai
    Version:  v0.30.1-desktop.1
    Path:     /Users/docker/.docker/cli-plugins/docker-buildx
    Version:  v5.0.1
    Path:     /Users/docker/.docker/cli-plugins/docker-compose
    Version:  0.0.47
    Path:     /Users/docker/.docker/cli-plugins/docker-debug
    Version:  v0.2.0
    Path:     /Users/docker/.docker/cli-plugins/docker-desktop
    Version:  v0.2.31
    Path:     /Users/docker/.docker/cli-plugins/docker-extension
    Version:  v1.4.0
    Path:     /Users/docker/.docker/cli-plugins/docker-init
    Version:  v0.35.0
    Path:     /Users/docker/.docker/cli-plugins/docker-mcp
    Version:  v1.0.6
    Path:     /Users/docker/.docker/cli-plugins/docker-model
    Version:  v0.5.40
    Path:     /Users/docker/.docker/cli-plugins/docker-offload
    Version:  v0.0.22
    Path:     /Users/docker/.docker/cli-plugins/docker-pass
  sandbox: Docker Sandbox (Docker Inc.)
    Version:  v0.6.0
    Path:     /Users/docker/.docker/cli-plugins/docker-sandbox
    Version:  0.6.0
    Path:     /Users/docker/.docker/cli-plugins/docker-sbom
    Version:  v1.19.0
    Path:     /Users/docker/.docker/cli-plugins/docker-scout

Server:
 Containers: 0
  Running: 0
  Paused: 0
  Stopped: 0
 Images: 2
 Server Version: 29.1.3
 Storage Driver: overlayfs
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 2
 Plugins:
  Volume: local
 CDI spec directories:
  /etc/cdi
  /var/run/cdi
 Swarm: inactive
 Runtimes: runc io.containerd.runc.v2
 Default Runtime: runc
 Init Binary: docker-init
 init version: de40ad0
 Security Options:
  seccomp
   Profile: builtin
  cgroupns
 Kernel Version: 6.12.54-linuxkit
 OSType: linux
 Architecture: aarch64
 CPUs: 12
 Total Memory: 54.82GiB
 Name: docker-desktop
 ID: 04401225-34f9-4d32-b681-5c63e65391db
 Debug Mode: false
 Labels:
  com.docker.desktop.address=unix:///Users/docker/Library/Containers/com.docker.docker/Data/docker-cli.sock
 Experimental: false
 Insecure Registries:
  hubproxy.docker.internal:5555
  ::1/128
  127.0.0.0/8
 Live Restore Enabled: false
```

### Anything else?

_No response_

ibrahimypr:
**What I did**
Until now, `mustRecreate` logic only checked for divergence in `TypeVolume` mounts but ignored `TypeImage` mounts. This inconsistency caused containers to erroneously retain stale images even after the source image was rebuilt.

This commit introduces a new label `com.docker.compose.image.mounts` that tracks the digests of images used as mounts. The convergence logic is updated to verify this label, ensuring that containers are correctly recreated whenever the underlying image mount digest changes.
**Related issue**
Fixes #13547

### DIFF
--- a/pkg/compose/build.go
+++ b/pkg/compose/build.go
@@ -158,9 +158,35 @@ func (s *composeService) ensureImagesExists(ctx context.Context, project *types.
 		if ok {
 			service.CustomLabels.Add(api.ImageDigestLabel, img.ID)
 		}
+
+		resolveImageVolumes(&service, images, project.Name)
+
 		project.Services[name] = service
 	}
 	return nil
+}
+
+func resolveImageVolumes(service *types.ServiceConfig, images map[string]api.ImageSummary, projectName string) {
+	for i, vol := range service.Volumes {
+		if vol.Type == types.VolumeTypeImage {
+			imgName := vol.Source
+			if _, ok := images[vol.Source]; !ok {
+				// check if source is another service in the project
+				imgName = api.GetImageNameOrDefault(types.ServiceConfig{Name: vol.Source}, projectName)
+				// If we still can't find it, it might be an external image that wasn't pulled yet or doesn't exist
+				if _, ok := images[imgName]; !ok {
+					continue
+				}
+			}
+			if img, ok := images[imgName]; ok {
+				// Use Image ID directly as source.
+				// Using name@digest format (via reference.WithDigest) fails for local-only images
+				// that don't have RepoDigests (e.g. built locally in CI).
+				// Image ID (sha256:...) is always valid and ensures ServiceHash changes on rebuild.
+				service.Volumes[i].Source = img.ID
+			}
+		}
+	}
 }
 
 func (s *composeService) getLocalImagesDigests(ctx context.Context, project *types.Project) (map[string]api.ImageSummary, error) {
--- a/pkg/e2e/fixtures/image-volume-recreate/Dockerfile
+++ b/pkg/e2e/fixtures/image-volume-recreate/Dockerfile
@@ -1,0 +1,21 @@
+# syntax=docker/dockerfile:1
+#
+#    Copyright 2020 Docker Compose CLI authors
+#
+#    Licensed under the Apache License, Version 2.0 (the "License");
+#    you may not use this file except in compliance with the License.
+#    You may obtain a copy of the License at
+#
+#        http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS,
+#    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#    See the License for the specific language governing permissions and
+#    limitations under the License.
+#
+
+FROM alpine
+WORKDIR /app
+ARG CONTENT=initial
+RUN echo "$CONTENT" > /app/content.txt
--- a/pkg/e2e/fixtures/image-volume-recreate/compose.yaml
+++ b/pkg/e2e/fixtures/image-volume-recreate/compose.yaml
@@ -1,0 +1,18 @@
+services:
+  source:
+    build:
+      context: .
+      dockerfile: Dockerfile
+    image: image-volume-source
+
+  consumer:
+    image: alpine
+    depends_on:
+      - source
+    command: ["cat", "/data/content.txt"]
+    volumes:
+      - type: image
+        source: image-volume-source
+        target: /data
+        image:
+          subpath: app
--- a/pkg/e2e/volumes_test.go
+++ b/pkg/e2e/volumes_test.go
@@ -190,3 +190,47 @@ func TestImageVolume(t *testing.T) {
 	out := res.Combined()
 	assert.Check(t, strings.Contains(out, "index.html"))
 }
+
+func TestImageVolumeRecreateOnRebuild(t *testing.T) {
+	c := NewCLI(t)
+	const projectName = "compose-e2e-image-volume-recreate"
+	t.Cleanup(func() {
+		c.cleanupWithDown(t, projectName)
+		c.RunDockerOrExitError(t, "rmi", "-f", "image-volume-source")
+	})
+
+	version := c.RunDockerCmd(t, "version", "-f", "{{.Server.Version}}")
+	major, _, found := strings.Cut(version.Combined(), ".")
+	assert.Assert(t, found)
+	if major == "26" || major == "27" {
+		t.Skip("Skipping test due to docker version < 28")
+	}
+
+	// First build and run with initial content
+	c.RunDockerComposeCmd(t, "-f", "./fixtures/image-volume-recreate/compose.yaml",
+		"--project-name", projectName, "build", "--build-arg", "CONTENT=foo")
+	res := c.RunDockerComposeCmd(t, "-f", "./fixtures/image-volume-recreate/compose.yaml",
+		"--project-name", projectName, "up", "-d")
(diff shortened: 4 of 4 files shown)


ibrahimypr:
Thanks for the feedback! I've updated the implementation to resolve the image digest directly in the volume source.
This way, ServiceHash naturally detects the change without needing an extra label.
Updated the tests as well.

pkg/compose/build.go
```diff
@@ -158,6 +158,21 @@ func (s *composeService) ensureImagesExists(ctx context.Context, project *types.
...
+		// Resolve image volume sources to include digest, so ServiceHash detects changes
+		for i, vol := range service.Volumes {
+			if vol.Type == types.VolumeTypeImage {
+				imgName := vol.Source
+				if target, ok := project.Services[vol.Source]; ok {
+					imgName = api.GetImageNameOrDefault(target, project.Name)
+				}
+				if img, ok := images[imgName]; ok {
+					// Append digest to source so config hash changes when image is rebuilt
+					service.Volumes[i].Source = fmt.Sprintf("%s@%s", vol.Source, img.ID)
```
ndeloof: better use of `reference.ParseDockerRef()` then `reference.WithDigest()` to enforce format follows expectations

pkg/compose/convergence_image_mount_test.go
```diff
@@ -0,0 +1,63 @@
...
+import (
+	"testing"
+
+	"github.com/compose-spec/compose-go/v2/types"
+	"github.com/docker/compose/v5/pkg/api"
+	"github.com/docker/docker/api/types/container"
+	"gotest.tools/v3/assert"
+)
+
+func TestMustRecreateImageMounts(t *testing.T) {
```
ndeloof: we prefer use of e2e test on docker/compose as those allows to detect actual issues with the engine API.

ibrahimypr:
Sorry for the amateur mistakes, this is my first open source contribution!
I've implemented the requested changes: used `reference.WithDigest` for proper formatting and replaced the unit test with an E2E test case. Thanks for guiding me through it!

ibrahimypr:
Fixed subpath in e2e test fixture (must be relative path) causing daemon error. Also addressed linter feedback.

pkg/compose/build.go
```diff
@@ -158,11 +160,43 @@ func (s *composeService) ensureImagesExists(ctx context.Context, project *types.
...
 		}
+
+		resolveImageVolumes(service, images, project.Name)
+
 		project.Services[name] = service
 	}
 	return nil
 }
 
+func resolveImageVolumes(service types.ServiceConfig, images map[string]api.ImageSummary, projectName string) {
```
ndeloof: I wonder this works as expected as you don't pass a `*types.ServiceConfig` so struct is copied in function scope, leaving original unmodified.

ibrahimypr:
Thanks again for the review!
1. **Fixed Struct Copy:** You were spot on about the struct copy issue—I've updated [resolveImageVolumes](compose/pkg/compose/build.go:168:0-189:1) to accept `*types.ServiceConfig` so modifications apply correctly.
2. **Local Image & CI:** I noticed the E2E tests for local builds were failing with `No such image` when using the `name@digest` format (via `reference.WithDigest`), likely because locally built images in CI don't always have valid repo references. To fix this, I switched to using the Image ID (`sha256:...`) directly as the volume source. This reliably triggers the `ServiceHash` change on rebuilds and works for both local and remote images.
Please let me know if this implementation looks good or if you'd prefer a different approach for handling local image references!

Issue #13561, opened 2026-01-28
[BUG] Panic when stopping Compose Watch


mikesir87:
### Description

When closing out Compose Watch, I'm seeing a panic output:

```
panic: close of closed channel

goroutine 28 [running]:
github.com/docker/compose/v5/pkg/watch.(*fseventNotify).Close(0x140003ad650)
	github.com/docker/compose/v5/pkg/watch/watcher_darwin.go:98 +0x70
github.com/docker/compose/v5/pkg/compose.(*composeService).watch.func2()
	github.com/docker/compose/v5/pkg/compose/watch.go:289 +0x48
github.com/docker/compose/v5/pkg/compose.(*Watcher).Start.func2()
	github.com/docker/compose/v5/pkg/compose/watch.go:98 +0x30
created by github.com/docker/compose/v5/pkg/compose.(*Watcher).Start in goroutine 1
	github.com/docker/compose/v5/pkg/compose/watch.go:97 +0x248
```

### Steps To Reproduce

1. Create a Compose file with Watch:
    
    ```yaml
    services:
      app:
        image: nginx
        develop:
          watch:
            - path: ./src
              action: sync
              target: /usr/share/nginx/html
              initial_sync: true
    ```

2. Add a simple index.html file (so it has something to watch):

    ```shell
    mkdir src && echo "HI THERE" > src/index.html
    ```

3. Start watch:
    
    ```shell
    docker compose up --watch
    ```

4. Stop it with a Ctrl+C. See the error

### Compose Version

```Text
Docker Compose version v5.0.2
```

### Docker Environment

```Text
Client:
 Version:    29.2.0
 Context:    desktop-linux
 Debug Mode: false
 Plugins:
  ai: Docker AI Agent - Ask Gordon (Docker Inc.)
    Version:  v1.17.2
    Path:     /Users/mikesir87/.docker/cli-plugins/docker-ai
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.31.0-desktop.1
    Path:     /Users/mikesir87/.docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v5.0.2
    Path:     /Users/mikesir87/.docker/cli-plugins/docker-compose
  debug: Get a shell into any image or container (Docker Inc.)
    Version:  0.0.47
    Path:     /Users/mikesir87/.docker/cli-plugins/docker-debug
  desktop: Docker Desktop commands (Docker Inc.)
[...]

PR #13562, opened 2026-01-28
fix panic

ndeloof:
**What I did**

**Related issue**
fixes https://github.com/docker/compose/issues/13561

### DIFF
--- a/pkg/watch/watcher_darwin.go
+++ b/pkg/watch/watcher_darwin.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"sync"
 	"time"
 
 	"github.com/fsnotify/fsevents"
@@ -38,6 +39,7 @@ type fseventNotify struct {
 	stop   chan struct{}
 
 	pathsWereWatching map[string]any
+	closeOnce         sync.Once
 }
 
 func (d *fseventNotify) loop() {
@@ -81,6 +83,8 @@ func (d *fseventNotify) Start() error {
 		return nil
 	}
 
+	d.closeOnce = sync.Once{}
+
 	numberOfWatches.Add(int64(len(d.stream.Paths)))
 
 	err := d.stream.Start()
@@ -92,11 +96,13 @@ func (d *fseventNotify) Start() error {
 }
 
 func (d *fseventNotify) Close() error {
-	numberOfWatches.Add(int64(-len(d.stream.Paths)))
+	d.closeOnce.Do(func() {
+		numberOfWatches.Add(int64(-len(d.stream.Paths)))
 
-	d.stream.Stop()
-	close(d.errors)
-	close(d.stop)
+		d.stream.Stop()
+		close(d.errors)
+		close(d.stop)
+	})
 
 	return nil
 }
--- a/pkg/watch/watcher_darwin_test.go
+++ b/pkg/watch/watcher_darwin_test.go
@@ -1,0 +1,48 @@
+//go:build fsnotify
+
+/*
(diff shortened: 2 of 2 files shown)


glours (dismissed):

pkg/watch/watcher_darwin.go
```diff
@@ -92,11 +94,13 @@ func (d *fseventNotify) Start() error {
 }
 
 func (d *fseventNotify) Close() error {
-	numberOfWatches.Add(int64(-len(d.stream.Paths)))
+	d.closeOnce.Do(func() {
```
thaJeztah: I see there's also a `Start()` - should the `closeOnce` be reset in `Start()` (otherwise it won't be run again after restarting?)


(not sure if that scenario is possible)

pkg/watch/watcher_darwin.go
```diff
@@ -81,6 +83,8 @@ func (d *fseventNotify) Start() error {
 		return nil
 	}
 
+	d.closeOnce = sync.Once{}
```
thaJeztah: LOL, not a new problem, but now starting to wonder if either `Start()` should be protected too only run once 🙈 - calling start multiple times wouldn't work well

That's not a new problem though, but perhaps something to look into.


That made me wonder if there's cases where "create watcher" and "start watcher" should be separate; looks like it's used here, and calls `Start` immediately after `NewWatcher` (wondering if `Start` should be done as part of the constructor); https://github.com/docker/compose/blob/bf86ba0e95b626a9e38c25b1dfd4ce232dfd8390/pkg/compose/watch.go#L271-L279

Although I guess it's nice to be able to construct a new watching without immediately starting it in some cases.

pkg/watch/watcher_darwin_test.go
```diff
@@ -0,0 +1,48 @@
...
+	err = watcher.Close()
+	assert.NilError(t, err)
+
+	// Close should be idempotent - calling it again should not panic
+	err = watcher.Close()
+	assert.NilError(t, err)
+
+	// Even a third time should be safe
+	err = watcher.Close()
+	assert.NilError(t, err)
```
thaJeztah, line 47: Perhaps the test should also  run `Start`  after this, and try `Close` again?

thaJeztah (approved):
LGTM

but left two comments / suggestions (perhaps adding the test would make sense for this PR).

pkg/watch/watcher_darwin.go
ndeloof: `Start` could indeed take place in constructor.
`Start` is not expected to be ran after watched was closed (channels are closed)

thaJeztah: Yea, I was looking because otherwise calling start multiple times would append a new listener to the global var, and start a new goroutine
https://github.com/docker/compose/blob/e22426443e6bc187f41ef0e6755ad8f121bf6d86/pkg/watch/watcher_darwin.go#L84
https://github.com/docker/compose/blob/e22426443e6bc187f41ef0e6755ad8f121bf6d86/pkg/watch/watcher_darwin.go#L90


So yeah, probably could be merged into `NewWatcher` then; looks like that requires an interface change 🤔 https://github.com/docker/compose/blob/e22426443e6bc187f41ef0e6755ad8f121bf6d86/pkg/watch/notify.go#L38-L40

For a more "organic" approach an alternative could also be to remove the `Close` function, and make `Start` return a "cancel" / "close" function, so that it's clear to the caller that it must be closed after, i.e. something like;

```go
close, err := watcher.Start()
// ....
defer close()
```


Either way, that's an existing issue, and not something that's hit by the current code

glours (approved):

PR #13563, opened 2026-01-29
pkg/compose: remove dependency on github.com/docker/buildx/driver

thaJeztah:
The driver.Auth interface was describing the configfile.GetAuthConfig implementation; define a local interface instead of using buildx's definition as an intermediate.

**What I did**

**Related issue**

### DIFF
--- a/pkg/compose/pull.go
+++ b/pkg/compose/pull.go
@@ -29,8 +29,8 @@ import (
 
 	"github.com/compose-spec/compose-go/v2/types"
 	"github.com/distribution/reference"
-	"github.com/docker/buildx/driver"
 	"github.com/docker/cli/cli/config/configfile"
+	clitypes "github.com/docker/cli/cli/config/types"
 	"github.com/docker/docker/api/types/image"
 	"github.com/docker/docker/client"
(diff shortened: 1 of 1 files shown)


pkg/compose/pull.go
```diff
@@ -260,7 +260,11 @@ func ImageDigestResolver(ctx context.Context, file *configfile.ConfigFile, apiCl
 	}
 }
 
-func encodedAuth(ref reference.Named, configFile driver.Auth) (string, error) {
+type authProvider interface {
+	GetAuthConfig(registryHostname string) (clitypes.AuthConfig, error)
+}
+
+func encodedAuth(ref reference.Named, configFile authProvider) (string, error) {
```
thaJeztah, line 267: Alternatively, this could just use the configfile itself as argument, but perhaps it was meant to be able to stub out / use different implementations;


```suggestion
func encodedAuth(ref reference.Named, configFile *configfile.ConfigFile) (string, error) {
```

thaJeztah:
test-failure is probably a flakey

ndeloof (approved):

PR #13568, opened 2026-02-03
use Docker GitHub Builder to build and sign binaries and bin image

crazy-max:
**What I did**

Similar to what we did in Buildx repo:
* https://github.com/docker/buildx/pull/3520
* https://github.com/docker/buildx/pull/3586
* https://github.com/docker/buildx/pull/3635

Switch to our [Docker GitHub Builder](https://github.com/docker/github-builder) to build and sign binaries and bin image.

For binaries, the sigstore bundle will be upload as asset to the GitHub Release. For the bin image the signature is an OCI artifact.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -109,6 +109,24 @@ jobs:
           path: ./bin/release/*
           if-no-files-found: error
 
+  bin-image-test:
+    if: github.event_name == 'pull_request'
+    uses: docker/github-builder/.github/workflows/bake.yml@v1
+    with:
+      runner: amd64
+      target: image-cross
+      cache: true
+      cache-scope: bin-image-test
+      output: image
+      push: false
+      sbom: true
+      set-meta-labels: true
+      meta-images: |
+        compose-bin
+      meta-tags: |
+        type=ref,event=pr
+      meta-bake-target: meta-helper
+
   test:
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/merge.yml
+++ b/.github/workflows/merge.yml
@@ -74,63 +74,39 @@ jobs:
         run: |
           make e2e-compose-standalone
 
-  bin-image:
-    runs-on: ubuntu-22.04
+  bin-image-prepare:
+    runs-on: ubuntu-24.04
     outputs:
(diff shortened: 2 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -63,63 +45,88 @@ jobs:
...
+      cache-scope: bin-image-test
+      output: image
+      push: false
+      sbom: true
+      set-meta-labels: true
+      meta-images: |
+        compose-bin
+      meta-tags: |
+        type=ref,event=pr
+      meta-bake-target: meta-helper
```
crazy-max, line 128: This job is just for testing bin image job works properly on pull request event as the merge workflow doesn't have any pull request trigger atm.

.github/workflows/merge.yml
```diff
@@ -74,63 +74,39 @@ jobs:
...
+      output: image
+      push: ${{ github.event_name != 'pull_request' }}
+      sbom: true
+      set-meta-labels: true
+      meta-images: |
+        ${{ needs.bin-image-prepare.outputs.repo-slug }}
+      meta-tags: |
+        type=ref,event=tag
+        type=edge
+      meta-bake-target: meta-helper
```
crazy-max, line 104: With our reusable workflow, the build is distributed across runners using a matrix based on target platforms to be built, reducing the build time quite a lot.

Before ~14m:

image

Now ~3m:

image

crazy-max:
Looking at new ci run, would need to update branch protection rule to remove the following jobs from required check:

image

.github/workflows/ci.yml
```diff
@@ -63,63 +45,88 @@ jobs:
...
+    with:
+      runner: amd64
+      artifact-name: compose
+      artifact-upload: true
+      cache: true
+      cache-scope: binary
+      target: release
+      output: local
+      sbom: true
+      sign: ${{ github.event_name != 'pull_request' }}
```
ndeloof, line 61: nice! I was looking for a way to get this into docker/compose and eventually kill compose-private, seems we are getting closer

ndeloof (approved):
🪄

.github/workflows/ci.yml
crazy-max: Yes for now we should keep compose-private for notarizing darwin and windows binaries.

PR #13569, opened 2026-02-03
ci: fix bin-image job

crazy-max:
relates to https://github.com/docker/compose/actions/runs/21631806154/job/62346301683#step:8:540

```
Error: meta-images is required when output is image
```

Forgot to put the dependency on bin-image :persevere:

Also fixes the bin-image edge tag being used for e2e tests.


crazy-max:
Oh I also need to fix the resulted digest in https://github.com/docker/compose/blob/9dc7f1e70ca11468ea9c6d12b26f117fe44f1726/.github/workflows/merge.yml#L137

PR #13576, opened 2026-02-07
fix(progress): restore TTY output on Windows consoles

### LINKED ISSUE #13570: [BUG] TTY Progress No Longer Works
### Description

Ever since upgrading to Docker Desktop 4.56.0, the output of `docker compose build` or `docker compose up --build` only outputs plain progress. For years, by default, it had that nice, colorized, simple output. Now, no matter what I try to do, I'm stuck with the verbose plain output.

### Steps To Reproduce

Run `docker compose -f Web/docker-compose-corewebs.yml -f Web/docker-compose-corewebs.override.yml up -d --build`, get plain progress.

Run `$env:BUILDKIT_PROGRESS="tty"` then the above command, and you get the error: `failed to get console: creating a console from a file is not supported on windows`

I've tried both Microsoft Terminal and a plain PowerShell window.

### Compose Version

```Text
PS C:\> docker version
Client:
 Version:           29.2.0
 API version:       1.53
 Go version:        go1.25.6
 Git commit:        0b9d198
 Built:             Mon Jan 26 19:28:56 2026
 Context:           desktop-linux

Server: Docker Desktop 4.59.0 (217644)
 Engine:
  Version:          29.2.0
  API version:      1.53 (minimum version 1.44)
  Go version:       go1.25.6
  Git commit:       9c62384
  Built:            Mon Jan 26 19:26:07 2026
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          v2.2.1
  GitCommit:        dea7da592f5d1d2b7755e3a161be07f43fad8f75
 runc:
  Version:          1.3.4
  GitCommit:        v1.3.4-0-gd6d73eb8
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
PS C:\> docker compose version
Docker Compose version v5.0.2
PS C:\> docker-compose version
Docker Compose version v5.0.2
```

### Docker Environment

```Text
PS C:\> docker info
Client:
 Version:    29.2.0
 Context:    desktop-linux
 Debug Mode: false
 Plugins:
    Version:  v1.17.2
    Path:     C:\Program Files\Docker\cli-plugins\docker-ai.exe
    Version:  v0.31.1-desktop.1
    Path:     C:\Program Files\Docker\cli-plugins\docker-buildx.exe
    Version:  v5.0.2
    Path:     C:\Program Files\Docker\cli-plugins\docker-compose.exe
    Version:  0.0.47
    Path:     C:\Program Files\Docker\cli-plugins\docker-debug.exe
    Version:  v0.2.0
    Path:     C:\Program Files\Docker\cli-plugins\docker-desktop.exe
  dev: Docker Dev Environments (Docker Inc.)
    Version:  v0.1.2
    Path:     C:\Program Files\Docker\cli-plugins\docker-dev.exe
    Version:  v0.2.31
    Path:     C:\Program Files\Docker\cli-plugins\docker-extension.exe
  feedback: Provide feedback, right in your terminal! (Docker Inc.)
    Version:  v1.0.5
    Path:     C:\Program Files\Docker\cli-plugins\docker-feedback.exe
    Version:  v1.4.0
    Path:     C:\Program Files\Docker\cli-plugins\docker-init.exe
    Version:  v0.37.0
    Path:     C:\Program Files\Docker\cli-plugins\docker-mcp.exe
    Version:  v1.0.8
    Path:     C:\Program Files\Docker\cli-plugins\docker-model.exe
    Version:  v0.5.41
    Path:     C:\Program Files\Docker\cli-plugins\docker-offload.exe
    Version:  v0.0.24
    Path:     C:\Program Files\Docker\cli-plugins\docker-pass.exe
  sandbox: Docker Sandbox (Docker Inc.)
    Version:  v0.10.1
    Path:     C:\Program Files\Docker\cli-plugins\docker-sandbox.exe
    Version:  0.6.0
    Path:     C:\Program Files\Docker\cli-plugins\docker-sbom.exe
    Version:  v1.19.0
    Path:     C:\Program Files\Docker\cli-plugins\docker-scout.exe

Server:
 Containers: 9
  Running: 9
  Paused: 0
  Stopped: 0
 Images: 21
 Server Version: 29.2.0
 Storage Driver: overlayfs
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 2
 Plugins:
  Volume: local
 CDI spec directories:
  /etc/cdi
  /var/run/cdi
 Discovered Devices:
  cdi: docker.com/gpu=webgpu
 Swarm: inactive
 Default Runtime: runc
 Init Binary: docker-init
 init version: de40ad0
 Security Options:
  seccomp
   Profile: builtin
  cgroupns
 OSType: linux
 Architecture: x86_64
 CPUs: 8
 Total Memory: 7.756GiB
 Name: docker-desktop
 ID: ac452699-68fc-46bd-9803-4280f0763067
 Debug Mode: false
 Labels:
  com.docker.desktop.address=npipe://\\.\pipe\docker_cli
 Experimental: false
 Insecure Registries:
  hubproxy.docker.internal:5555
  127.0.0.0/8
  ::1/128
 Live Restore Enabled: false
 Firewall Backend: iptables
```

### Anything else?

_No response_

maishivamhoo123:
## Description

Fixes a regression where TTY progress output (colors/bars) was lost on Windows consoles.
The issue was caused by [explain why the console detection failed].

## Related Issue
Fixes #13570

## Type of change
- [x] Bug fix (non-breaking change which fixes an issue)

##
## Testing
- [x] Verified on PowerShell that progress bars appear with `docker compose up --build`

### DIFF
--- a/pkg/compose/build_bake.go
+++ b/pkg/compose/build_bake.go
@@ -435,6 +435,9 @@ func (s *composeService) getBuildxPlugin() (*manager.Plugin, error) {
 // buildkit's NewDisplay doesn't actually require a [io.Reader], it only uses the [containerd.Console] type to
 // benefits from ANSI capabilities, but only does writes.
 func makeConsole(out io.Writer) io.Writer {
+	if s, ok := out.(*streams.Out); ok {
+		return &_console{s}
+	}
 	return out
 }
 
@@ -444,20 +447,24 @@ type _console struct {
 	*streams.Out
 }
 
-func (c _console) Read(p []byte) (n int, err error) {
+func (c *_console) Read(p []byte) (n int, err error) {
 	return 0, errors.New("not implemented")
 }
 
-func (c _console) Close() error {
+func (c *_console) Close() error {
 	return nil
 }
 
-func (c _console) Fd() uintptr {
+func (c *_console) Fd() uintptr {
 	return c.FD()
 }
 
-func (c _console) Name() string {
+func (c *_console) Name() string {
 	return "compose"
+}
+func (c *_console) File() *os.File {
+	// streams.Out wraps the actual file. We attempt to unwrap it.
+	return os.NewFile(c.Out.FD(), "compose")
 }
 
 func toBakeExtraHosts(hosts types.HostsList) map[string]string {
--- a/pkg/compose/build_bake.go
+++ b/pkg/compose/build_bake.go
@@ -435,9 +435,6 @@ func (s *composeService) getBuildxPlugin() (*manager.Plugin, error) {
 // buildkit's NewDisplay doesn't actually require a [io.Reader], it only uses the [containerd.Console] type to
 // benefits from ANSI capabilities, but only does writes.
 func makeConsole(out io.Writer) io.Writer {
-	if s, ok := out.(*streams.Out); ok {
-		return &_console{s}
-	}
 	return out
 }
 


ndeloof (changes requested):
don't you think this `_console` wrapper was introduced for a reason?

amyssnippet:
you r right @ndeloof , i looked into history (commit bf50c99 by u) and looks like _console acts as an adapter to force BuildKit to treat the output as a containerd.Console, which enables the ANSI progress bars and colors. but here is the twist, as this works great on linux and mac, but on windows this wrapper seems to interfere with how the console handle is passed down, causing the TTY output to disappear/break. since we don't want to lose progress bars on linux and mac, I suggest to modify makeConsole to strictly return the original out on windows, while keeping the wrapper for everyone else. do you understand @maishivamhoo123

pkg/compose/build_bake.go
```diff
@@ -435,9 +435,6 @@ func (s *composeService) getBuildxPlugin() (*manager.Plugin, error) {
 // buildkit's NewDisplay doesn't actually require a [io.Reader], it only uses the [containerd.Console] type to
 // benefits from ANSI capabilities, but only does writes.
 func makeConsole(out io.Writer) io.Writer {
-	if s, ok := out.(*streams.Out); ok {
-		return &_console{s}
-	}
 	return out
```
amyssnippet, line 441:
```go
func makeConsole(out io.Writer) io.Writer {
    // On windows, the wrapper breaks TTY output, so we return the raw writer.
        return out
    }
    // ... existing logic for linux and mac
}
```

amyssnippet (commented):
i think this should be the logic

pkg/compose/build_bake.go
ndeloof: this proposal maybe is a valid workaround but doesn't address the root issue. For some reason (?) the wrapped `streams.Out` can't be used by buildkit progressui - probably `checkConsole`, which differs between OSes, can't handle this. Then we need to fix this bug, not just hide it

maishivamhoo123:
> you r right @ndeloof , i looked into history (commit [bf50c99](https://github.com/docker/compose/commit/bf50c99193a61f09c0be7d6b6a06c014d825521f) by u) and looks like _console acts as an adapter to force BuildKit to treat the output as a containerd.Console, which enables the ANSI progress bars and colors. but here is the twist, as this works great on linux and mac, but on windows this wrapper seems to interfere with how the console handle is passed down, causing the TTY output to disappear/break. since we don't want to lose progress bars on linux and mac, I suggest to modify makeConsole to strictly return the original out on windows, while keeping the wrapper for everyone else. do you understand @maishivamhoo123

Yes @amyssnippet Sir , understood. I have updated the code as suggested. Thank you sir.

amyssnippet:
look @maishivamhoo123 we are here to fix bugs which lie under the hood, what i suggested was just a workaround. rn as @ndeloof suggested, that this should work on every os, and we should consider on fixing it under the hood of buildkit tht why it fails. we should fix that. 

we should look at how _console implements Fd(). On Windows, we might need to ensure the wrapper creates a bridge to the underlying file handle correctly so isTerminal checks pass.

Can you check how streams.Out implements FD() vs how our wrapper passes it through? We might need a Windows-specific implementation of the wrapper that exposes the handle correctly.

amyssnippet:
and also make sure to do a signoff commit

pkg/compose/build_bake.go
```diff
@@ -447,21 +447,25 @@ type _console struct {
...
 	return c.FD()
 }
 
-func (c _console) Name() string {
+func (c *_console) Name() string {
 	return "compose"
 }
+func (c *_console) File() *os.File {
+	// streams.Out wraps the actual file. We attempt to unwrap it.
+	return os.NewFile(c.Out.FD(), "compose")
```
amyssnippet, line 467: this could lead to resource leaks or unexpected behavior if the file descriptor is closed prematurely. The comment acknowledges the wrapping, but this assumes streams.Out.FD() always points to a valid console file.

amyssnippet (commented):
this is still just a workaround, we have to dig deeper into buildkit and docker cli, they are different maintained projects in the docker and containerd ecosystem.

amyssnippet:
tl;dr: dont just on any possible feature request to prove your AI agent can deal with compose codebase, as this can never  get implemented in the compose repo. Better confirm with maintainers like @ndeloof which one deserve some time for investigation and design discussions in issue which posted this #13570

amyssnippet:
@ndeloof kindly proceed to close this pr, it no longer solves the issue and also no response from the issue author

PR #13577, opened 2026-02-08
PRIVATE_PORT optional for `docker compose port`

### LINKED ISSUE #11859: Make PRIVATE_PORT optional for `docker compose port`

maxproske:
**Related issue**

Closes #11859, and continues #11874 which was approved but went stale.

AI Disclaimer: Claude Code helped sharpen after I incorporated all requested changes from the stale PR.

**What I did**

Align `docker compose port` closer with `docker port`. When no port given, list all mappings.

Future improvement: allow inline protocol like docker port does (eg. 80/udp)?

```bash
# --- Before ---
docker compose port web
# "port" requires at least 2 arguments

# --- After ---
docker compose port web
# 80/tcp -> 0.0.0.0:8080
# 443/tcp -> 0.0.0.0:8443
# 53/udp -> 0.0.0.0:5353
```

### DIFF
--- a/cmd/compose/port.go
+++ b/cmd/compose/port.go
@@ -19,6 +19,8 @@ package compose
 import (
 	"context"
 	"fmt"
+	"net"
+	"sort"
 	"strconv"
 	"strings"
 
@@ -41,16 +43,20 @@ func portCommand(p *ProjectOptions, dockerCli command.Cli, backendOptions *Backe
 		ProjectOptions: p,
 	}
 	cmd := &cobra.Command{
-		Use:   "port [OPTIONS] SERVICE PRIVATE_PORT",
-		Short: "Print the public port for a port binding",
-		Args:  cobra.MinimumNArgs(2),
+		Use:   "port [OPTIONS] SERVICE [PRIVATE_PORT]",
+		Short: "List port mappings or print the public port of a specific mapping for the service",
+		Args:  cobra.RangeArgs(1, 2),
 		PreRunE: Adapt(func(ctx context.Context, args []string) error {
-			port, err := strconv.ParseUint(args[1], 10, 16)
-			if err != nil {
-				return err
-			}
-			opts.port = uint16(port)
 			opts.protocol = strings.ToLower(opts.protocol)
+			if len(args) > 1 {
+				port, err := strconv.ParseUint(args[1], 10, 16)
+				if err != nil {
+					return err
+				}
+				opts.port = uint16(port)
+			} else {
+				opts.protocol = ""
+			}
 			return nil
 		}),
 		RunE: Adapt(func(ctx context.Context, args []string) error {
@@ -73,7 +79,7 @@ func runPort(ctx context.Context, dockerCli command.Cli, backendOptions *Backend
 	if err != nil {
 		return err
 	}
-	ip, port, err := backend.Port(ctx, projectName, service, opts.port, api.PortOptions{
+	publishers, err := backend.Ports(ctx, projectName, service, opts.port, api.PortOptions{
 		Protocol: opts.protocol,
 		Index:    opts.index,
 	})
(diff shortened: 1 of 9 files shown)


maxproske:
@glours Resolved conflicts :)

pkg/compose/port.go
```diff
@@ -26,18 +26,40 @@ import (
...
 	if err != nil {
-		return "", 0, err
+		return nil, err
 	}
+
+	if port != 0 {
+		for _, p := range ctr.Ports {
+			if p.PrivatePort == port && p.Type == options.Protocol {
+				return api.PortPublishers{{
+					URL:           p.IP.String(),
```
glours: `p.IP.String()` on an unpublished port (invalid IP) literally returns `"invalid IP"`. `containerPublishers()` in `ps.go` already guards this, this function doesn't.
```suggestion
				var url string
                  if p.IP.IsValid() {
                          url = p.IP.String()
                  }
				return api.PortPublishers{{
					URL:           p.IP.String(),
```

pkg/compose/port.go
```diff
@@ -26,18 +26,40 @@ import (
...
+
+	var publishers api.PortPublishers
 	for _, p := range ctr.Ports {
-		if p.PrivatePort == port && p.Type == options.Protocol {
-			return p.IP.String(), int(p.PublicPort), nil
+		if options.Protocol != "" && p.Type != options.Protocol {
+			continue
 		}
+		publishers = append(publishers, api.PortPublisher{
+			URL:           p.IP.String(),
```
glours: Same issue for list mode (docker compose port web with no PRIVATE_PORT).

```suggestion
		var url string
        if p.IP.IsValid() {
                url = p.IP.String()
        }
		publishers = append(publishers, api.PortPublisher{
			URL:           p.IP.String(),
```

cmd/compose/port.go
```diff
@@ -41,16 +43,20 @@ func portCommand(p *ProjectOptions, dockerCli command.Cli, backendOptions *Backe
 		ProjectOptions: p,
 	}
 	cmd := &cobra.Command{
-		Use:   "port [OPTIONS] SERVICE PRIVATE_PORT",
-		Short: "Print the public port for a port binding",
-		Args:  cobra.MinimumNArgs(2),
+		Use:   "port [OPTIONS] SERVICE [PRIVATE_PORT]",
+		Short: "List port mappings or print the public port of a specific mapping for the service",
+		Args:  cobra.RangeArgs(1, 2),
 		PreRunE: Adapt(func(ctx context.Context, args []string) error {
```
glours: `PreRunE` needs `cmd` to check whether `--protocol` was explicitly set.
```suggestion
		PreRunE: AdaptCmd(func(ctx context.Context, cmd *cobra.Command, args []string) error {
```

cmd/compose/port.go
```diff
@@ -41,16 +43,20 @@ func portCommand(p *ProjectOptions, dockerCli command.Cli, backendOptions *Backe
...
-			}
-			opts.port = uint16(port)
 			opts.protocol = strings.ToLower(opts.protocol)
+			if len(args) > 1 {
+				port, err := strconv.ParseUint(args[1], 10, 16)
+				if err != nil {
+					return err
+				}
+				opts.port = uint16(port)
+			} else {
```
glours: Without a `PRIVATE_PORT`, `opts.protocol` is reset unconditionally, an explicit `--protocol udp` gets ignored.
```suggestion
			} else if !cmd.Flags().Changed("protocol") {
```

glours (changes requested):
So sorry for the delay @maxproske 
There is still things to fix before we can merge

pkg/compose/port.go
maxproske: Thank you, I did not even consider an `expose:` only port'd be invalid when I ported (haha) this over

cmd/compose/port.go
maxproske: You're right I get tcp rows, fff-

maxproske:
Thanks @glours for taking the time to explain each gotcha. Updated!

glours (approved):
Thanks @maxproske 🙏

PR #13580, opened 2026-02-10
pkg/compose: defaultNetworkSettings: refactor

thaJeztah:
- [x] follow-up to / stacked on https://github.com/docker/compose/pull/13078

### pkg/compose: defaultNetworkSettings: slight refactor

- use an intermediate serviceNetworks slice so that we don't have to call service.NetworksByPriority multiple times.
- shift the primary network from the slice (if any), so that we can drop some checks for "additional networks"
- group code related to setting up the primary network as first step, then append remaining networks.

**What I did**

**Related issue**

### DIFF
--- a/pkg/compose/create.go
+++ b/pkg/compose/create.go
@@ -525,7 +525,7 @@ func defaultNetworkSettings(project *types.Project,
 	}
 
 	if len(project.Networks) == 0 {
-		return "none", nil, nil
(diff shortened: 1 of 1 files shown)


thaJeztah:
Thx! I rebased https://github.com/docker/compose/pull/13584 and moved it out of draft

PR #13589, opened 2026-02-13
Fix SIGINT handling: re-raise signal instead of exit(130) 

### LINKED ISSUE #13586: [BUG] "pull" doesn't exit on SIGINT correctly
### Description

Since `pull` (and possibly) other commands catch SIGINT they have to take care how to exit properly and communicate this to parent process (e.g., a shell script) so it can react appropriately and exit too.

### Steps To Reproduce

This can be demostrated by running a script consisting of a loop over all sub-directories (compose projects) in order to `pull` them.

Assume `/opt/docker/make` is:

```
#!/usr/bin/env bash

set -eu

cd "${BASH_SOURCE%/*}"

for dir in [^.]*/; do
    cd "$dir"
    printf '\n> %s\n' "$PWD"
    if (( $# )); then
        docker compose "$@" || true
    fi
    cd -
done
echo
```

1) `cd /opt/docker`
2) `./make pull`
3) `Ctrl-C` somewhere during pull of one of projects
4) observe that the parent script won't break the loop but continue with next project directory

### Compose Version

```Text
Docker Compose version 5.0.2
```

### Docker Environment

```Text
Client:
 Version:    29.2.1
 Context:    default
 Debug Mode: false
 Plugins:
    Version:  5.0.2

Server:
 Containers: 18
  Running: 18
  Paused: 0
  Stopped: 0
 Images: 24
 Server Version: 29.2.1
 Storage Driver: overlay2
  Backing Filesystem: cramfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
 CDI spec directories:
  /etc/cdi
  /var/run/cdi
 Swarm: inactive
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: dea7da592f5d1d2b7755e3a161be07f43fad8f75.m
 runc version: 
 init version: de40ad0
 Security Options:
  seccomp
   Profile: builtin
  cgroupns
 Kernel Version: 6.12.59
 Operating System: Arch Linux
 OSType: linux
 Architecture: x86_64
 CPUs: 8
 Total Memory: 4GiB
 Name: baobab
 ID: 2802ebac-9027-4cad-ace9-fb64970a9604
 Debug Mode: false
 Experimental: false
 Insecure Registries:
  ::1/128
  127.0.0.0/8
 Live Restore Enabled: false
 Default Address Pools:
   Base: 172.17.0.0/16, Size: 24
 Firewall Backend: iptables

WARNING: No io.weight support
WARNING: No io.weight (per device) support
WARNING: No io.max (rbps) support
WARNING: No io.max (wbps) support
WARNING: No io.max (riops) support
WARNING: No io.max (wiops) support
```

### Anything else?

For reference see:

* https://www.cons.org/cracauer/sigint.html
* https://pubs.opengroup.org/onlinepubs/9699919799/functions/wait.html

4RH1T3CT0R7:
- Fix docker compose pull (and all other commands) to properly re-raise SIGINT/SIGTERM after cleanup instead of exiting with code 130. This ensures parent processes see WIFSIGNALED=true via waitpid(), enabling correct WCE (Wait and Cooperative Exit) shell propagation.
  - Propagate context cancellation error from pullServiceImage instead of swallowing it, so concurrent pulls stop immediately on Ctrl+C.

  Problem

  When a user presses Ctrl+C during docker compose pull, the process exits via os.Exit(130). The parent shell's waitpid() sees WIFEXITED=true (normal exit), not WIFSIGNALED=true (signal death). This breaks shell script interrupt handling — the parent script continues executing instead of stopping.

Fixes #13586


pkg/e2e/up_test.go
```diff
@@ -94,11 +94,13 @@ func TestUpDependenciesNotStopped(t *testing.T) {
...
-		errors.As(err, &exitErr)
-		if exitErr.ExitCode() == -1 {
-			t.Fatalf("`compose up` was killed: %v", err)
+		if !errors.As(err, &exitErr) {
+			t.Fatalf("`compose up` failed with non-exit error: %v", err)
 		}
-		require.Equal(t, 130, exitErr.ExitCode())
+		// Process is expected to die from re-raised SIGINT signal (exit code -1).
+		// If signal re-raise doesn't terminate the process, the fallback path exits with code 130.
+		assert.Assert(t, exitErr.ExitCode() == -1 || exitErr.ExitCode() == 130,
```
ndeloof:
```suggestion
		if runtime.GOOS == "windows" {
      assert.Equal(t, 130, exitErr.ExitCode())
  } else {
      assert.Equal(t, -1, exitErr.ExitCode())
  }
```

ndeloof (dismissed):

4RH1T3CT0R7:
@ndeloof Good point! Since up_test.go already has a //go:build !windows constraint, the process will always die from the re-raised signal on this platform. Simplified the assertion to expect exactly -1 instead of == -1 || == 130. Thanks for the review!

ndeloof:
test failure is related to the proposed fix.

4RH1T3CT0R7:
> test failure is related to the proposed fix.

fixed

Issue #13602, opened 2026-02-21
[BUG] bind.create_host_path: false is ignored and missing host path is still auto-created


kyoh86:
### Description

When using bind mounts in Compose long syntax, I think that `bind.create_host_path: false` should prevent automatic creation of missing host paths.

https://github.com/compose-spec/compose-spec/blob/main/spec.md#short-syntax-5

> **Note**
> 
> For bind mounts, short syntax creates a directory at the source path on the host if it doesn't exist. This is for backward compatibility with `docker-compose` legacy. It can be prevented by using long syntax and setting `create_host_path` to `false`.

In my environment (`docker compose v5.0.2` from Docker Desktop on WSL2), the missing host directory is still created automatically, and the service starts successfully.

Expected behavior:
- `docker compose run` should fail if the bind source path does not exist and `create_host_path: false` is set.

Actual behavior:
- The missing host path is created anyway.
- The container starts.
- The created directory is owned by `root:root` on host.

This breaks workflows that rely on strict pre-created runtime directories and ownership control.

### Steps To Reproduce

1. Create this `compose.yml`:

```yaml
services:
  should_fail:
    image: alpine:3.21
    command: ["sh", "-lc", "echo unexpected-success"]
    volumes:
      - type: bind
        source: ./tmp/missing-dir
        target: /mnt/missing
        bind:
          create_host_path: false
```

2. Run:

```console
$ sudo rm -rf ./tmp
$ docker compose -f compose.yml run --rm should_fail
$ ls -ld ./tmp ./tmp/*
```

3. Observe:

- should_fail unexpectedly succeeds.
- `./tmp/missing-dir` is created.

### Compose Version

```Text
$ docker compose version
Docker Compose version v5.0.2
$ docker-compose version
Docker Compose version v5.0.2
```

### Docker Environment

```Text
Client:
 Version:    29.2.1
 Context:    default
 Debug Mode: false
 Plugins:
  ai: Docker AI Agent - Ask Gordon (Docker Inc.)
    Version:  v1.18.0
    Path:     /usr/local/lib/docker/cli-plugins/docker-ai
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.31.1-desktop.1
    Path:     /usr/local/lib/docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v5.0.2
    Path:     /usr/local/lib/docker/cli-plugins/docker-compose
[...]

PR #13605, opened 2026-02-22
fix: check Docker Engine version for create_host_path: false support

Aliexe-code:
Added an API version check for the `CreateMountpoint` bind mount option, which was introduced in Docker Engine API v1.42 (Engine v23.0).

Previously, when using `create_host_path: false` on an older Docker Engine (< v23.0), the setting was silently ignored and the host path was created anyway. This caused unexpected behavior for users who relied on strict pre-created runtime directories.

Now, when a user attempts to use `create_host_path: false` on an unsupported Docker Engine, they receive a clear error message: bind mount create_host_path: false requires Docker Engine v23 or later

   **Changes:**
   - Added `apiVersion142` constant for Docker Engine 23.0
   - Added `dockerEngineV23` version string for error messages
   - Added API version check in `buildContainerVolumes()`
   - Updated tests to use mocks and added test case for the new error path
   - Bonus: removed duplicate `RuntimeVersion()` call in `TypeImage` case

   ## Related issue

See: #13602

### DIFF
--- a/pkg/compose/api_versions.go
+++ b/pkg/compose/api_versions.go
@@ -19,6 +19,15 @@ package compose
 // Docker Engine API version constants.
 // These versions correspond to specific Docker Engine releases and their features.
 const (
+	// apiVersion142 represents Docker Engine API version 1.42 (Engine v23.0).
+	//
+	// New features in this version:
+	//  - CreateMountpoint option for bind mounts (allows create_host_path: false)
+	//
+	// Before this version:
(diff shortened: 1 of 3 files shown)


ndeloof (commented):
this won't fix https://github.com/docker/compose/issues/13602 as reported is running docker engine v29

ndeloof (approved):
LGTM, while reported issue is unrelated

PR #13607, opened 2026-02-23
fix: execute post_start hooks in docker compose run

### LINKED ISSUE #13593: [BUG] "docker compose run" does not run post_start hook
### Description

Reading the documentation, I think post-start hooks should be run when using "docker compose run", but they don't appear to be.

For example the documentation for "run" (https://docs.docker.com/reference/cli/docker/compose/run/) states: "The following command starts the web service and runs bash as its command: $ docker compose run web bash"
While the documentation for the hooks (https://docs.docker.com/compose/how-tos/lifecycle/) states "Post-start hooks are commands that run after the container has started". I would assume that in order to start a service, compose would first need to start the container, and thus run post-start hooks.

### Steps To Reproduce

Trying a small example, using the following docker-compose.yml file:
```services:
    test_service:
        image: alpine
        command: /bin/sh -c "echo this is command; sleep 3600"
        post_start:
          - command: /bin/sh -c "echo this is post_start > /hook_output.txt && cat /hook_output.txt"
```
"docker compose up" runs the hook:
```
$ docker compose up
[+] Running 2/2
 ✔ test_service Pulled                                                                                                                                                                                        4.5s 
   ✔ 589002ba0eae Pull complete                                                                                                                                                                               2.4s 
[+] Running 2/2
 ✔ Network docker_hook_test_default           Created                                                                                                                                                         0.1s 
 ✔ Container docker_hook_test-test_service-1  Created                                                                                                                                                         0.3s 
Attaching to test_service-1
test_service-1  | this is command
test_service-1 ->  | this is post_start
```

While "docker compose run test_service sh" apparently does not:
```
$ docker compose run test_service sh
[+] Creating 1/1
 ✔ Network docker_hook_test_default  Created                                                                                                                                                                  0.1s 
/ # cat /hook_output.txt
cat: can't open '/hook_output.txt': No such file or directory
/ # 
```

Note that the "test_service-1 ->  | this is post_start" line is not printed when using "run", and the file created by the post-start hook command is also missing. 

### Compose Version

```Text
Docker Compose version 2.37.1+ds1-0ubuntu2~22.04.1
```

### Docker Environment

```Text
Client:
 Version:    27.5.1
 Context:    default
 Debug Mode: false
 Plugins:
    Version:  0.21.3
    Version:  2.37.1+ds1-0ubuntu2~22.04.1

Server:
 Containers: 136
  Running: 3
  Paused: 0
  Stopped: 133
 Images: 119
 Server Version: 27.5.1
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
 Swarm: inactive
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 
 runc version: 
 init version: 
 Security Options:
  apparmor
  seccomp
   Profile: builtin
  cgroupns
 Kernel Version: 6.8.0-94-generic
 Operating System: Ubuntu 22.04.5 LTS
 OSType: linux
 Architecture: x86_64
 CPUs: 8
 Total Memory: 31.14GiB
 Name: t14
 ID: a211b7cc-ec1b-4310-9f47-cb1d414e4b9f
 Debug Mode: false
 Experimental: false
 Insecure Registries:
  127.0.0.0/8
 Live Restore Enabled: false
```

### Anything else?

_No response_

veeceey:
`docker compose run` wasn't executing `post_start` lifecycle hooks — they only ran via `docker compose up` (in `startService`) and `docker compose restart`. This adds hook execution to `RunOneOffContainer` so the behavior is consistent across commands.

The tricky part is that `cmd.RunStart` handles both starting and attaching to the container in one call, so we can't just run hooks sequentially after start. Instead, a goroutine listens for the container's start event via the Docker Events API and kicks off the hooks once the container is running. This runs concurrently with the attach in interactive mode, and the hook result is collected after `RunStart` returns.

Closes #13593

### DIFF
--- a/pkg/compose/run.go
+++ b/pkg/compose/run.go
@@ -27,14 +27,22 @@ import (
 	"github.com/compose-spec/compose-go/v2/types"
 	"github.com/docker/cli/cli"
 	cmd "github.com/docker/cli/cli/command/container"
+	"github.com/moby/moby/api/types/container"
+	"github.com/moby/moby/api/types/events"
 	"github.com/moby/moby/client"
 	"github.com/moby/moby/client/pkg/stringid"
 
 	"github.com/docker/compose/v5/pkg/api"
 )
 
+type prepareRunResult struct {
+	containerID string
+	service     types.ServiceConfig
+	created     container.Summary
+}
+
 func (s *composeService) RunOneOffContainer(ctx context.Context, project *types.Project, opts api.RunOptions) (int, error) {
-	containerID, err := s.prepareRun(ctx, project, opts)
+	result, err := s.prepareRun(ctx, project, opts)
 	if err != nil {
 		return 0, err
 	}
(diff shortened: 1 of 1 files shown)


veeceey:
### Test results

Unit tests pass:
```
$ go test ./pkg/compose/...
ok  	github.com/docker/compose/v5/pkg/compose	0.555s
ok  	github.com/docker/compose/v5/pkg/compose/transform	(cached)
```

Build succeeds with no issues:
```
$ go build ./pkg/compose/...
(no errors)
```

PR #13629, opened 2026-03-09
fix: restore post-connect fallback for multi-network stacks on API < 1.44

### LINKED ISSUE #13628: [bug] regression from 5.02 - multi-network stacks broken on Docker API < 1.44 (regression in v5.1.0)
### Summary

Docker Compose 5.1.0 broke multi-network stacks on Docker daemons older than API 1.44 (Docker Engine < 23.0), including Synology DSM 7.1 (API 1.41) and DSM 7.2 (API 1.43).

I maintain [Dockhand](https://github.com/Finsys/dockhand), a Docker management application. Many of our users run Dockhand on Synology NAS devices (DSM 7.1/7.2), which ship with Docker 20.10 (API 1.41). This regression broke multi-network stack deployments for all of them when we updated docker-compose to 5.1.0. We had to cap docker-compose at 5.0.2 as a workaround while investigating the root cause.

### Error Message

```
Container cannot be connected to network endpoints
```

This error comes from the Docker daemon when `ContainerCreate` receives multiple `EndpointsConfig` entries, which old daemons do not support.

### Root Cause

Two changes in 5.1.0 combined to remove backward compatibility:

**1. `defaultNetworkSettings()` refactor (commit `fefdc9522`)**

In 5.0.2, extra networks were only added to `EndpointsConfig` for API >= 1.44:

```go
// 5.0.2 — guarded by version check
if versions.GreaterThanOrEqualTo(version, APIVersion144) {
    for _, networkKey := range serviceNetworks {
        endpointsConfig[mobyNetworkName] = epSettings
    }
}
```

In 5.1.0, the guard was removed. Now ALL networks are unconditionally added to `EndpointsConfig` in `ContainerCreate`, even on old daemons:

```go
// 5.1.0 — no guard, always adds all networks
for _, networkKey := range serviceNetworks {
    endpointsConfig[mobyNetworkName] = epSettings
}
```

**2. The post-connect fallback was removed from `convergence.go`**

In 5.0.2, `createMobyContainer()` had a fallback that connected extra networks one-by-one via `NetworkConnect()` for daemons older than API 1.44:

```go
// 5.0.2 convergence.go — this was removed in 5.1.0
if versions.LessThan(apiVersion, APIVersion144) {
    for _, networkKey := range serviceNetworks {
        // skip primary (already in ContainerCreate)
        // NetworkConnect() for each extra network
    }
}
```

### Affected Users

- Synology DSM 7.1 (Docker API 1.41 / Engine 20.10)
- Synology DSM 7.2 (Docker API 1.43 / Engine 23.0-pre)
- Any Docker Engine < 23.0 (API < 1.44)

### Reproduction

```yaml
# docker-compose.yml
services:
  web:
    image: nginx:alpine
    networks:
      - frontend
      - backend
networks:
  frontend:
  backend:
```

Run `docker compose up -d` against a Docker 20.10 daemon:

```
Error response from daemon: Container cannot be connected to network endpoints
```

### Expected Behavior

Same behavior as 5.0.2: multi-network stacks work on all supported Docker daemon versions.

### Fix

Restore the two behaviors that existed in 5.0.2:

1. In `defaultNetworkSettings()` (`create.go`): only include extra networks in `EndpointsConfig` when `apiVersion >= 1.44`
2. In `createMobyContainer()` (`convergence.go`): add back the post-connect fallback that calls `NetworkConnect()` for each extra network when `apiVersion < 1.44`

A fix is proposed in the linked PR.

### Workaround

Downgrade to Docker Compose 5.0.2 or earlier.

jotka:
## Problem

Docker Compose 5.1.0 broke multi-network stacks on Docker daemons older than API 1.44 (Docker Engine < 23.0), including Synology DSM 7.1 (API 1.41) and DSM 7.2 (API 1.43). The error is:

```
Container cannot be connected to network endpoints
```

This is a regression from 5.0.2 which worked correctly on these daemons.

Closes #13628

## Context

I maintain [Dockhand](https://github.com/Finsys/dockhand), a Docker management application. Many of our users run Dockhand on Synology NAS devices (DSM 7.1/7.2), which ship with Docker 20.10 (API 1.41). This regression broke multi-network stack deployments for all of them when we updated docker-compose to 5.1.0. We had to cap docker-compose at 5.0.2 as a workaround while investigating the root cause.

## Root Cause

Both regressions were introduced in commit bfb5511d0 (`go.mod: bump github.com/moby/moby/api v1.53.0, moby/client v0.2.2`):

**1. `defaultNetworkSettings()` in `create.go`**

The `versions.GreaterThanOrEqualTo(version, APIVersion144)` guard was removed during the moby API bump, making `defaultNetworkSettings` unconditionally include ALL networks in `EndpointsConfig` for `ContainerCreate`. Old daemons (< API 1.44) only accept a single entry and reject the request with the error above.

**2. The post-connect fallback was removed from `convergence.go`**

The same commit removed the `NetworkConnect` fallback loop from `createMobyContainer()`. Previously (since aaa7ef6de), after `ContainerCreate`, the code checked `versions.LessThan(apiVersion, "1.44")` and connected extra networks one-by-one via `NetworkConnect()`. This entire block was dropped during the API migration.

### First commit (`4f14ab8`): Restore backward compat

Restores the two behaviors from 5.0.2:

1. **`create.go` — `defaultNetworkSettings()`**: Only include extra networks in `EndpointsConfig` when `apiVersion >= 1.44` (guarded by `!versions.LessThan(version, apiVersion144)`).

2. **`convergence.go` — `createMobyContainer()`**: Restore the post-connect fallback. After container creation, when `apiVersion < 1.44`, iterate the service's extra networks and call `NetworkConnect()` for each, skipping the primary network (already configured via `NetworkMode` in `ContainerCreate`).

### Second commit (`7895a5a`)

1. **Pass `APIVersion` through `createConfigs`**: Added `APIVersion` field to the `createConfigs` struct, populated from the existing `apiVersion` local in `getCreateConfigs()`. This eliminates the redundant `s.RuntimeVersion(ctx)` call in `createMobyContainer`.

2. **Reorder `ContainerInspect` after `NetworkConnect`**: Moved the inspect call after the `NetworkConnect` loop so the returned `container.Summary` includes all attached networks.

3. **Unit tests for the < 1.44 code path**:
   - `TestDefaultNetworkSettings` sub-test with version `"1.43"` — asserts `EndpointsConfig` contains only the primary network.
   - `TestCreateMobyContainerLegacyAPI` — asserts `ContainerCreate` gets only the primary network, `NetworkConnect` is called for the secondary, and `ContainerInspect` runs after `NetworkConnect`.

## Testing

- Unit tests pass (the `TestRunHook_ConsoleSize` failure is a pre-existing PTY crash on macOS/arm64, unrelated to this change and present on `main` before this fix)
- Tested manually against Docker 20.10 (API 1.41) — multi-network stacks now deploy correctly
- No regression expected on modern daemons (API >= 1.44): they continue to use the multi-endpoint `ContainerCreate` path

### DIFF
--- a/pkg/compose/convergence.go
+++ b/pkg/compose/convergence.go
@@ -33,6 +33,7 @@ import (
 	"github.com/moby/moby/api/types/container"
 	mmount "github.com/moby/moby/api/types/mount"
 	"github.com/moby/moby/client"
+	"github.com/moby/moby/client/pkg/versions"
 	specs "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/sirupsen/logrus"
 	"go.opentelemetry.io/otel/attribute"
@@ -732,6 +733,37 @@ func (s *composeService) createMobyContainer(ctx context.Context, project *types
 			Text:   warning,
 		})
 	}
+	// Starting API version 1.44, the ContainerCreate API call takes multiple networks
+	// so we include all configurations there and can skip the one-by-one calls here.
+	// For older API versions (e.g. Docker 20.10/API 1.41, Synology DSM 7.1/7.2),
+	// extra networks must be connected individually after creation via NetworkConnect.
+	apiVersion, err := s.RuntimeAPIVersion(ctx)
+	if err != nil {
+		return created, err
+	}
+	if versions.LessThan(apiVersion, apiVersion144) {
+		serviceNetworks := service.NetworksByPriority()
+		for _, networkKey := range serviceNetworks {
+			mobyNetworkName := project.Networks[networkKey].Name
+			if string(cfgs.Host.NetworkMode) == mobyNetworkName {
+				// primary network already configured as part of ContainerCreate
+				continue
+			}
+			epSettings, err := createEndpointSettings(project, service, number, networkKey, cfgs.Links, opts.UseNetworkAliases)
+			if err != nil {
+				_, _ = s.apiClient().ContainerRemove(ctx, response.ID, client.ContainerRemoveOptions{Force: true})
+				return created, err
+			}
+			if _, err := s.apiClient().NetworkConnect(ctx, mobyNetworkName, client.NetworkConnectOptions{
+				Container:      response.ID,
+				EndpointConfig: epSettings,
+			}); err != nil {
+				_, _ = s.apiClient().ContainerRemove(ctx, response.ID, client.ContainerRemoveOptions{Force: true})
+				return created, err
+			}
+		}
+	}
+
 	res, err := s.apiClient().ContainerInspect(ctx, response.ID, client.ContainerInspectOptions{})
 	if err != nil {
 		return created, err
--- a/pkg/compose/convergence_test.go
+++ b/pkg/compose/convergence_test.go
@@ -498,6 +498,152 @@ func TestCreateMobyContainer(t *testing.T) {
 	assert.NilError(t, err)
 }
 
+func TestCreateMobyContainerLegacyAPI(t *testing.T) {
+	mockCtrl := gomock.NewController(t)
+	defer mockCtrl.Finish()
+	apiClient := mocks.NewMockAPIClient(mockCtrl)
+	cli := mocks.NewMockCli(mockCtrl)
+	tested, err := NewComposeService(cli)
+	assert.NilError(t, err)
+	cli.EXPECT().Client().Return(apiClient).AnyTimes()
+	cli.EXPECT().ConfigFile().Return(&configfile.ConfigFile{}).AnyTimes()
+	apiClient.EXPECT().DaemonHost().Return("").AnyTimes()
+	apiClient.EXPECT().ImageInspect(anyCancellableContext(), gomock.Any()).
+		Return(client.ImageInspectResult{}, nil).AnyTimes()
+
+	apiClient.EXPECT().Ping(gomock.Any(), client.PingOptions{NegotiateAPIVersion: true}).
+		Return(client.PingResult{APIVersion: "1.43"}, nil).AnyTimes()
+	apiClient.EXPECT().ClientVersion().Return("1.43").AnyTimes()
+
+	service := types.ServiceConfig{
+		Name: "test",
+		Networks: map[string]*types.ServiceNetworkConfig{
+			"a": {Priority: 10},
+			"b": {Priority: 100},
+		},
+	}
+	project := types.Project{
+		Name: "bork",
+		Services: types.Services{
+			"test": service,
+		},
+		Networks: types.Networks{
+			"a": types.NetworkConfig{Name: "a-moby-name"},
+			"b": types.NetworkConfig{Name: "b-moby-name"},
+		},
+	}
+
+	var gotCreate client.ContainerCreateOptions
+	apiClient.EXPECT().ContainerCreate(gomock.Any(), gomock.Any()).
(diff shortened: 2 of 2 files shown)


jotka:
Good catch to flag — I had the same instinct initially. But after scanning the codebase, this comparison is safe due to a structural guarantee in compose-go:

**`network_mode` and `networks` are mutually exclusive** (`loader/validate.go:67-68`). When `network_mode` is explicitly set (`host`, `none`, `container:xxx`, `service:xxx` → resolved to `container:<id>`), `defaultNetworkSettings()` returns early with `nil` NetworkingConfig, so the loop body never executes — there are zero networks to iterate.

The only case where the loop runs is the standard `networks:` path, where `defaultNetworkSettings()` sets `NetworkMode` to exactly `container.NetworkMode(project.Networks[primaryKey].Name)` — the same moby name string we compare against.

That said, the codebase itself identifies primary by position (`serviceNetworks[0]` from `NetworksByPriority()`), not by string match. I can switch to a positional skip if you'd prefer:

```go
for i, networkKey := range serviceNetworks {
    if i == 0 {
        continue // primary, already in ContainerCreate
    }
```

Functionally equivalent here, but arguably more explicit about intent. Let me know.

pkg/compose/convergence.go
```diff
@@ -732,6 +733,35 @@ func (s *composeService) createMobyContainer(ctx context.Context, project *types
...
+			}
+			epSettings, err := createEndpointSettings(project, service, number, networkKey, cfgs.Links, opts.UseNetworkAliases)
+			if err != nil {
+				return created, err
+			}
+			if _, err := s.apiClient().NetworkConnect(ctx, mobyNetworkName, client.NetworkConnectOptions{
+				Container:      response.ID,
+				EndpointConfig: epSettings,
+			}); err != nil {
+				return created, err
```
glours, line 762: For `API < 1.44`, this PR now creates the container first and then attaches secondary networks one by one here. If one of these NetworkConnect() calls fails, the function returns immediately and leaves the just-created container behind on the primary network. That means the legacy fallback is no longer atomic on failure, and a later up / recreate can hit a name conflict against the orphaned container.

I think we should clean up the created container on NetworkConnect failure before returning.

jotka: Good catch, thanks! Added ContainerRemove with Force:true on both error paths (createEndpointSettings and NetworkConnect). Also added a test for the failure case. Re the API version negotiation — noted, will rebase on top of #13690 once it lands.

glours: FYI PR #13690 has been merged 
ps: don't forgot to sign-off your commits 😉

jotka:
Rebased on main (post go1.25.8, moby/client v1.54.0, etc.) and fixed lint issues (line length in test closures). All tests pass locally.

jotka:
Rebased on main (post #13690 merge). Changes:

- Dropped `APIVersion` from `createConfigs` struct — no longer needed since `createMobyContainer` now calls `s.RuntimeAPIVersion(ctx)` directly
- Updated all test mocking to use the new `Ping` + `ClientVersion` pattern (no more `ServerVersion` or package-level `runtimeVersion` reset)
- Fixed lint issues (line length in test closures)

All tests pass, linter clean.

jotka:
awesome, thanks!

PR #13637, opened 2026-03-15
Filter services in config --no-interpolate to respect [SERVICE...] argument

### LINKED ISSUE #13614: [BUG] `--no-interpolate` flag on `docker compose config` has unexpected side effects
### Description

Given the example compose file under "Steps To Reproduce", the result of running:

``` console
$ docker compose config --no-normalize --no-path-resolution example1
```

is:

``` yaml
name: demo
services:
  example1:
    container_name: example1
    environment:
      TZ: Australia/Sydney
    image: example/example
    ports:
      - mode: ingress
        target: 3000
        published: "3001"
        protocol: tcp
    restart: unless-stopped
    volumes:
      - type: bind
        source: ./volumes/example1/data
        target: /var/lib/data
        bind: {}
```

I have chosen:

* `--no-normalize` because it removes references to networks; and
* `--no-path-resolution` because it maintains project-relative external paths in `volumes:` clauses.

The output shows `TZ` with the value `Australia/Sydney` (from the `.env` file). My goal is that that should be displayed in its original form as `TZ: ${TZ:-Etc/UTC}`.

The `--no-interpolate` flag is [defined](https://docs.docker.com/reference/cli/docker/compose/config/) as "Don't interpolate environment variables" and seems like a reasonable candidate for achieving this goal. The result of adding that flag is shown in the screen shot in the next section.

The `--no-interpolate` flag has achieved the expected result with respect to `TZ` but it has also produced a bunch of unexpected side effects, ranging from relatively unimportant through problematic, none of which would seem to have anything to do with environment variables.

For what I am trying to achieve, the most problematic difference is that the service definition for `example2` has appeared in the output. This behaviour suggests that the presence of `--no-interpolate` causes `config` to ignore the `[SERVICE...]` argument.

I assume the `name: demo` (first line on the left hand side of the screen shot) is the project name. That line disappearing is actually beneficial for what I am trying to achieve but it isn't immediately clear what this has to do with environment variables.
 
The reverse-ordering of the last three lines of the `ports:` clause doesn't matter for my purposes. The same applies to the `volumes:` clause where the ordering and structural differences are just … *strange*.

### Steps To Reproduce

1. Example compose file:

	``` yaml
	---
	
	services:
	
	  example1:
	    container_name: example1
	    image: example/example
	    restart: unless-stopped
	    ports:
	      - "3001:3000"
	    environment:
	      TZ: ${TZ:-Etc/UTC}
	    volumes:
	      - ./volumes/example1/data:/var/lib/data
	
	  example2:
	    container_name: example2
	    image: example/example
	    restart: unless-stopped
	    ports:
	      - "3002:3000"
	    environment:
	      TZ: ${TZ:-Etc/UTC}
	    volumes:
	      - ./volumes/example2/data:/var/lib/data
	
	...
	```

2. Run:

	``` console
	$ diff -y \
	  <(docker compose config --no-normalize --no-path-resolution                  example1) \
	  <(docker compose config --no-normalize --no-path-resolution --no-interpolate example1)
	```
	
3. The result is:

	Image

### Compose Version

```Text
$ docker compose version
Docker Compose version v5.1.0

$ docker-compose version
Docker Compose version v5.1.0
```

### Docker Environment

```Text
$ docker info
 Version:    29.2.1
 Context:    default
 Debug Mode: false
 Plugins:
    Version:  v0.31.1
    Version:  v5.1.0

Server:
 Containers: 6
  Running: 6
  Paused: 0
  Stopped: 0
 Images: 6
 Server Version: 29.2.1
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: local
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
 CDI spec directories:
  /etc/cdi
  /var/run/cdi
 Swarm: inactive
 Default Runtime: runc
 Init Binary: docker-init
 init version: de40ad0
 Security Options:
  seccomp
   Profile: builtin
  cgroupns
 Kernel Version: 6.1.21-v8+
 OSType: linux
 Architecture: aarch64
 CPUs: 4
 Total Memory: 3.704GiB
 Name: sec-dev
 Debug Mode: false
 Experimental: false
 Insecure Registries:
  ::1/128
  127.0.0.0/8
 Live Restore Enabled: false
 Firewall Backend: iptables
```

### Anything else?

Older cases raising issues with `--no-interpolate` (eg #13327, #12292, #12079, #11547) all seem to use examples where the compose file only contains a single service definition. While that scenario does show the unexpected re-ordering of clauses that would not seem to be related to environment variables, it won't reveal the issue with the `[SERVICE...]` argument being ignored.

odlev:
## What I did

When using `docker compose config --no-interpolate` with specific service names
(e.g., `docker compose config --no-interpolate example1`), the `[SERVICE...]`
argument was being ignored and all services were output.

The root cause is that `runConfigNoInterpolate()` calls `ToModel()` which
returns the full model without filtering by service names. Added service
filtering after model loading to match the behavior of the interpolated
code path.

## Related issue

Fixes #13614

## How I tested

- Added an e2e test with a multi-service compose file that verifies only
  the requested service is output when using `--no-interpolate` with a
  service name filter
- Verified the build compiles and existing unit tests pass

### DIFF
--- a/cmd/compose/config.go
+++ b/cmd/compose/config.go
@@ -271,6 +271,20 @@ func runConfigNoInterpolate(ctx context.Context, dockerCli command.Cli, opts con
 		return nil, err
 	}
 
+	if len(services) > 0 {
+		if svcs, ok := model["services"].(map[string]any); ok {
+			filtered := make(map[string]any, len(services))
+			for _, name := range services {
+				if svc, exists := svcs[name]; exists {
+					filtered[name] = svc
+				} else {
+					return nil, fmt.Errorf("no such service: %s", name)
+				}
+			}
+			model["services"] = filtered
+		}
+	}
+
 	if opts.resolveImageDigests {
 		err = resolveImageDigests(ctx, dockerCli, model)
 		if err != nil {
--- a/pkg/e2e/config_test.go
+++ b/pkg/e2e/config_test.go
@@ -17,8 +17,10 @@
 package e2e
 
 import (
+	"strings"
 	"testing"
 
+	"gotest.tools/v3/assert"
 	"gotest.tools/v3/icmd"
 )
 
@@ -45,6 +47,13 @@ func TestLocalComposeConfig(t *testing.T) {
 	t.Run("--no-interpolate", func(t *testing.T) {
 		res := c.RunDockerComposeCmd(t, "-f", "./fixtures/config/compose.yaml", "--project-name", projectName, "config", "--no-interpolate")
 		res.Assert(t, icmd.Expected{Out: `- ${PORT:-8080}:80`})
+	})
+
+	t.Run("--no-interpolate with service filter", func(t *testing.T) {
+		res := c.RunDockerComposeCmd(t, "-f", "./fixtures/config-filter/compose.yaml", "--project-name", projectName, "config", "--no-interpolate", "example1")
+		res.Assert(t, icmd.Expected{Out: `example1`})
+		// example2 should NOT appear in the output when only example1 is requested
+		assert.Assert(t, !strings.Contains(res.Stdout(), "example2"), "expected example2 to be filtered out")
 	})
 
 	t.Run("--variables --format json", func(t *testing.T) {
--- a/pkg/e2e/fixtures/config-filter/compose.yaml
+++ b/pkg/e2e/fixtures/config-filter/compose.yaml
@@ -1,0 +1,18 @@
+services:
+  example1:
+    container_name: example1
+    image: example/example
+    restart: unless-stopped
+    ports:
+      - "3001:3000"
+    environment:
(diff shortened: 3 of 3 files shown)


ndeloof:
that's incorrect. A compose command with a SERVICE set will enable target service **and dependencies**, which is challenging to implement as the dependency logic relies on go types in compose-go

themavik (commented):
nit: if `[SERVICE...]` is non-empty but `model["services"]` is not a `map[string]any`, filtering is skipped with no error — easy to misread as "filter worked" on odd loader output.

PR #13645, opened 2026-03-18
Fix docker compose ps to honor psFormat setting in .docker/config.json

### LINKED ISSUE #13643: 'docker compose ps' not honoring 'psFormat' setting in '.docker/config.json'
Hi, 

Running 

```bash
```

```json
{
}
```

```bash
```
results in the correct table. 

This means

```bash
docker compose ps --format ""
```

results in the same output as 

```bash
```

but **not** 

```bash
docker compose ps
```

```go
format := options.format
if len(format) == 0 {
}
```

maks2134:
- Change format flag default from "table" to empty string to allow config override
- Add logic to use PsFormat from config when no format argument provided
- Fall back to "table" format when neither PsFormat nor format arg is set
- Add warning when both --format and --quiet flags are used
- Add test to verify format flag default value

Fixes #13643

**What I did**

**Related issue**


cmd/compose/ps.go
```diff
@@ -152,9 +152,14 @@ func runPs(ctx context.Context, dockerCli command.Cli, backendOptions *BackendOp
...
 
-	if opts.Format == "" {
+	if len(opts.Format) == 0 {
 		opts.Format = dockerCli.ConfigFile().PsFormat
 	}
+	if len(opts.Format) == 0 {
+		opts.Format = "table"
+	} else if opts.Quiet {
+		_, _ = dockerCli.Err().Write([]byte("WARNING: Ignoring custom format, because both --format and --quiet are set.\n"))
+	}
```
glours, line 163: With this changes users with psFormat in their config who runs `docker compose ps -q` would get a spurious warning.
We should only trigger this check when `cmd.Flags().Changed("format")` is true

glours: Also I would prefer we use `fmt.Fprintln` instead of `Write` like this
`fmt.Fprintln(dockerCli.Err(), "WARNING: ...")`

glours (changes requested):
Thanks for the contribution 🙏 
This one need some changes to properly and transparently support `psFormat` setting
Can you also add more tests regarding the comment I made when changing or not the `--format` flag please?

cmd/compose/ps.go
maks2134: Thanks for the review, I've made some changes.

cmd/compose/ps.go
```diff
@@ -69,19 +69,20 @@ func psCommand(p *ProjectOptions, dockerCli command.Cli, backendOptions *Backend
...
+	var psCmd *cobra.Command
+	psCmd = &cobra.Command{
 		Use:   "ps [OPTIONS] [SERVICE...]",
 		Short: "List containers",
 		PreRunE: func(cmd *cobra.Command, args []string) error {
 			return opts.parseFilter()
 		},
 		RunE: Adapt(func(ctx context.Context, args []string) error {
-			return runPs(ctx, dockerCli, backendOptions, args, opts)
+			return runPs(ctx, dockerCli, backendOptions, args, opts, psCmd)
```
glours, line 80:
```suggestion
			return runPs(ctx, dockerCli, backendOptions, args, opts, cmd.Flags().Changed("format"))
```

maks2134:
Please take a look at the changes when you have the opportunity. @glours

cmd/compose/ps.go
glours, line 142: We already leave the function here if the `quiet` flag is used by the user. So the code checking both `quiet` and `format` should be done in `RunE` function before we call `runPs`

cmd/compose/ps.go
```diff
@@ -152,9 +153,14 @@ func runPs(ctx context.Context, dockerCli command.Cli, backendOptions *BackendOp
 		return nil
 	}
 
-	if opts.Format == "" {
+	if len(opts.Format) == 0 {
```
glours, line 156: `opts.Format` is a string
```suggestion
	if opts.Format == "" {
```

cmd/compose/ps.go
```diff
@@ -152,9 +153,14 @@ func runPs(ctx context.Context, dockerCli command.Cli, backendOptions *BackendOp
...
 		opts.Format = dockerCli.ConfigFile().PsFormat
 	}
+	if len(opts.Format) == 0 {
```
glours, line 159: `opts.Format` is a string
```suggestion
	if opts.Format == "" {
```

cmd/compose/ps_format_test.go
```diff
@@ -0,0 +1,161 @@
...
+
+	backendOptions := &BackendOptions{}
+	cmd := psCommand(projectOpts, cli, backendOptions)
+
+	// Check default value of format flag
+	formatFlag := cmd.Flags().Lookup("format")
+	assert.Equal(t, formatFlag.DefValue, "")
+}
+
+func TestPsCommandUsesConfigFormat(t *testing.T) {
```
glours, line 49: As we don't check `runPs` behaviour, this test don't provide any additional value, this is already covered by `TestPsCommandDefaultFormat`

cmd/compose/ps_format_test.go
```diff
@@ -0,0 +1,161 @@
...
+	// Simulate only quiet flag change
+	cmd.SetArgs([]string{"--quiet"})
+	cmd.ParseFlags([]string{"--quiet"})
+	
+	// Only quiet flag should be changed, not format
+	assert.Assert(t, !cmd.Flags().Changed("format"))
+	assert.Assert(t, cmd.Flags().Changed("quiet"))
+}
+
+func TestPsCommandFormatFallback(t *testing.T) {
```
glours, line 137: This test only check the flag was untouched not it defaulted to the `table` format

glours (changes requested):
Thanks for this work 🙏 
I did a more deep dive review, and there is still fixes to apply
Tests are only checking the flags combinatory but none of them really check the `runPs` logic
Maybe we should add a e2e test for that

PR #13646, opened 2026-03-18
Add `docker compose deploy` command

ericcurtin:
Implements a production-oriented deploy command that builds, pushes, and applies a Compose project to a Docker server with force-recreate semantics and optional health-check-based wait for zero-downtime deployments.


ndeloof:
I'm not convinced this requires a dedicated command. `compose up --build` basically does the same (without pushing). Many production system would not build locally but better pull from  a trusted registry. I also don't see `--force-recreate` to be a "production" requirement, what makes you feel so?

IMHO many user would have a distinct vision about what a "production" deployment looks like, and as such would just write their own minimal script to automate this compose command sequence

ndeloof:
I would also suggest you open a feature request issue / talk to maintainers on slack when you have such a proposal, so we can align before you volunteer to implement the feature

PR #13652, opened 2026-03-20
Fix: respect required: false for env_file in publish command

### LINKED ISSUE #13648: [BUG] docker compose publish ignores `required: false` on `env_file`
### Description

`env_file` with `required: false` is honoured at runtime (`up`, `run`, etc.)
but `docker compose publish` scans every `env_file` entry unconditionally,
failing if the file is absent:

```yaml
env_file:
  - path: ./.env
    required: false
```

```text
failed to scan env file ./.env: open ./.env: no such file or directory
```

### Steps To Reproduce

1. Create the file:

```bash
cat > compose.yaml <<EOF
services:
  whoami:
    env_file:
      - path: ./.env
        required: false
EOF
```

2. Publish (no `.env` present):

```bash
```

### Compose Version

```Text
$ docker-compose version

$ docker compose version
Docker Compose version v5.1.0
```

### Docker Environment

```Text
$ docker info
 Version:    29.3.0
 Context:    default
 Debug Mode: false
 Plugins:
    Version:  v0.31.1
    Version:  v5.1.0

Server:
 Containers: 0
  Running: 0
  Paused: 0
  Stopped: 0
 Images: 119
 Server Version: 4.9.3
 Storage Driver: overlay
  Supports d_type: true
  Native Overlay Diff: true
  Using metacopy: false
  Supports shifting: false
  Supports volatile: true
  Backing Filesystem: extfs
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
 Swarm: inactive
 Default Runtime: crun
 Init Binary: 
 containerd version: 
 runc version: 
 init version: 
 Security Options:
  seccomp
   Profile: default
  rootless
 Operating System: ubuntu
 OSType: linux
 Architecture: amd64
 CPUs: 8
 Total Memory: 7.557GiB
 Name: DENEC1
 Debug Mode: false
 Experimental: true
 Live Restore Enabled: false
 Product License: Apache-2.0
```

### Anything else?

I don't think it change anything but for transparency.

maks2134:
**What I did**
Fixed docker compose publish command to respect `required: false` setting on env_file entries. Previously, the command would fail when optional env files were missing, even when explicitly marked as not required.

Changes made:
- Modified [checkForSensitiveData()](cci:1://file:///Users/makskozlov/GolandProjects/temp-scud/compose/pkg/compose/publish.go:427:0-480:1) to skip env files with `required: false`
- Updated [processFile()](cci:1://file:///Users/makskozlov/GolandProjects/temp-scud/compose/pkg/compose/publish.go:225:0-285:1) to only process required env files into publish layers  
- Fixed [checkEnvironmentVariables()](cci:1://file:///Users/makskozlov/GolandProjects/temp-scud/compose/pkg/compose/publish.go:353:0-379:1) to only consider required env files for environment warnings
- Added comprehensive tests to verify the fix works correctly

**Related issue**
fixes #13648

### DIFF
--- a/pkg/compose/publish.go
+++ b/pkg/compose/publish.go
@@ -251,6 +251,9 @@ func processFile(ctx context.Context, file string, project *types.Project, extFi
 	}
 	for name, service := range base.Services {
 		for i, envFile := range service.EnvFiles {
+			if !envFile.Required {
+				continue
+			}
 			hash := fmt.Sprintf("%x.env", sha256.Sum256([]byte(envFile.Path)))
 			envFiles[envFile.Path] = hash
 			f, err = transform.ReplaceEnvFile(f, name, i, hash)
@@ -351,8 +354,11 @@ func (s *composeService) checkEnvironmentVariables(project *types.Project, optio
 	errorList := map[string][]string{}
 
 	for _, service := range project.Services {
-		if len(service.EnvFiles) > 0 {
-			errorList[service.Name] = append(errorList[service.Name], fmt.Sprintf("service %q has env_file declared.", service.Name))
+		for _, envFile := range service.EnvFiles {
+			if envFile.Required {
+				errorList[service.Name] = append(errorList[service.Name], fmt.Sprintf("service %q has env_file declared.", service.Name))
+				break
+			}
 		}
 	}
 
@@ -438,6 +444,9 @@ func (s *composeService) checkForSensitiveData(project *types.Project) ([]secret
 	for _, service := range project.Services {
 		// Check env files
 		for _, envFile := range service.EnvFiles {
+			if !envFile.Required {
+				continue
+			}
 			findings, err := scan.ScanFile(envFile.Path)
 			if err != nil {
 				return nil, fmt.Errorf("failed to scan env file %s: %w", envFile.Path, err)
--- a/pkg/compose/publish.go
+++ b/pkg/compose/publish.go
@@ -252,7 +252,9 @@ func processFile(ctx context.Context, file string, project *types.Project, extFi
 	for name, service := range base.Services {
 		for i, envFile := range service.EnvFiles {
 			if !envFile.Required {
-				continue
+				if _, err := os.Stat(envFile.Path); os.IsNotExist(err) {
+					continue
+				}
 			}
 			hash := fmt.Sprintf("%x.env", sha256.Sum256([]byte(envFile.Path)))
 			envFiles[envFile.Path] = hash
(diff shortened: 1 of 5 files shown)


ndeloof (dismissed):

themavik (commented):
nit: `checkEnvironmentVariables` still emits the same string (`has env_file declared`) even though the gate is now “any required env_file”; minor, but users with a mix of optional+required files might read it as “any env_file”.

pkg/compose/publish.go
```diff
@@ -251,6 +251,9 @@ func processFile(ctx context.Context, file string, project *types.Project, extFi
 	}
 	for name, service := range base.Services {
 		for i, envFile := range service.EnvFiles {
+			if !envFile.Required {
+				continue
+			}
```
glours, line 258: If the file does exist on the user's host, it should still be published as a layer and skipped by default

pkg/compose/publish.go
```diff
@@ -438,6 +444,9 @@ func (s *composeService) checkForSensitiveData(project *types.Project) ([]secret
 	for _, service := range project.Services {
 		// Check env files
 		for _, envFile := range service.EnvFiles {
+			if !envFile.Required {
+				continue
+			}
```
glours, line 454: Same here, we should check env variables if the file exist on user's host

glours (changes requested):
IMHO, we should allow non-required env files, check the sensitive data and add them as a layer when present on the user’s host.

pkg/compose/publish.go
maks2134: Done

pkg/compose/publish.go
maks2134: Done

maks2134:
Please take a look at the changes when you have the opportunity. @glours

pkg/compose/publish.go
```diff
@@ -351,8 +356,11 @@ func (s *composeService) checkEnvironmentVariables(project *types.Project, optio
 	errorList := map[string][]string{}
 
 	for _, service := range project.Services {
-		if len(service.EnvFiles) > 0 {
-			errorList[service.Name] = append(errorList[service.Name], fmt.Sprintf("service %q has env_file declared.", service.Name))
+		for _, envFile := range service.EnvFiles {
+			if envFile.Required {
+				errorList[service.Name] = append(errorList[service.Name], fmt.Sprintf("service %q has required env_file declared.", service.Name))
+				break
+			}
```
glours, line 363: We should also return an error when the file is not required but present on the user disk, because it will be upload as part of the Compose artifact

pkg/compose/publish_test.go
```diff
@@ -100,3 +101,75 @@ services:
...
+				},
+			},
+		},
+	}
+
+	service := &composeService{}
+
+	err := service.checkEnvironmentVariables(project, api.PublishOptions{})
+	assert.Assert(t, err != nil)
+	assert.Assert(t, strings.Contains(err.Error(), `service "test" has required env_file declared.`))
```
glours, line 173: nit 
```suggestion
	assert.ErrorContains(t, err, `service "test" has required env_file declared.`)
```

glours (changes requested):
This is still not the right expected behaviour, we should check the env files content if the files exist even if there are not mandatory

PR #13662, opened 2026-03-24
chore: pin GitHub Actions to commit SHA, remove pr-review workflow

glours:
**What I did**
- Pin all action references to full commit SHA instead of mutable version tags. Tag retained as inline comment for readability.
- Remove pr-review.yml workflow.

**Related issue**
N/A

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,10 +35,10 @@ jobs:
     steps:
       -
         name: Checkout
-        uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -35,17 +35,17 @@ jobs:
     steps:
       -
         name: Checkout
-        uses: actions/checkout@v4
+        uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
```
thaJeztah: Should we take the opportunity to update to current versions?  (v6 or v6.0.2)? I _think_ dependabot then also will update them.

(Not sure if the older versions use obsolete Node versions, so may become non-functional)

glours: Done @thaJeztah

PR #13674, opened 2026-03-26
publish: return api.ErrCanceled when user declines interactive prompts

ishwar170695:
**What I did**

Previously, declining the bind mount or sensitive data warning prompt caused publish() to return `nil`, which is indistinguishable from success for callers and automation scripts (exit code 0).

Change publish() to return the existing `api.ErrCanceled` sentinel so callers can detect user-initiated cancellation via `errors.Is()`. Add unit tests covering both the `preChecks` decline path and the full `publish()` path.

Signed-off-by: Ishwar chand meena <ishwarcm@iitbhilai.ac.in>

### DIFF
--- a/pkg/compose/publish_test.go
+++ b/pkg/compose/publish_test.go
@@ -104,7 +104,6 @@ services:
 }
 
 func Test_preChecks_sensitive_data_detected_decline(t *testing.T) {
-	
 	dir := t.TempDir()
 	envPath := dir + "/secrets.env"
 	secretData := `AWS_SECRET_ACCESS_KEY="wJalrXUtnFEMI/K7MDENG/bPxRfiCYEXAMPLEKEY"`
@@ -112,7 +111,6 @@ func Test_preChecks_sensitive_data_detected_decline(t *testing.T) {
 	assert.NilError(t, err)
 
 	project := &types.Project{
-
 		Services: types.Services{
 			"web": {
 				Name:  "web",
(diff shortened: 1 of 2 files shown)


ishwar170695:
Updated PR description to reflect the use of the existing api.ErrCanceled instead of creating a new custom error.

pkg/compose/publish_test.go
```diff
@@ -100,3 +101,62 @@ services:
 		return !slices.Contains([]string{".Data", ".Digest", ".Size"}, path.String())
 	}, cmp.Ignore()))
 }
+
+func Test_preChecks_decline_returns_ErrPublishAborted(t *testing.T) {
```
glours: There isn’t any ErrPublishedArboted error in the codebase, so the test name is misleading.
Second, this test is effectively a subset of Test_publish_decline_returns_ErrCanceled.

We can keep it, but I’d suggest repurposing it to cover the sensitive data detection path instead. That would make it exercise an uncovered code path and give it a clearer purpose.

ishwar170695:
Thanks for the feedback! I've repurposed the test to cover the checkForSensitiveData + decline path using a temp .env file with a dummy AWS secret key. The test name is updated to reflect what it actually tests.

glours:
@ishwar170695 can you sing-off your commits please

glours:
@ishwar170695   CI detected tests which need to be changed to check properly the new 130 exit code
And there is some lint issues to fix also

ishwar170695:
Sorry for disabling auto-merge, the force-push after rebase triggered that.
Addressed the lint issue (gofumpt on publish_test.go), updated e2e tests to expect exit code 130 for the decline paths, and rebased cleanly onto main. Ready for re-review when you get a chance.

Issue #13027, opened 2025-07-08
`--remove-orphans` not-working if compose profile changed


skul87:
When `--remove-orphans` is activated, the orphaned service is not deleted if I change profiles

```
Docker Compose version v2.36.2
```

```
$ cat compose.yml
services:

  sleep1:
    image: alpine:latest
    entrypoint: ["sleep"]
    command: ["9999"]
    profiles:
      - A

  sleep2:
    image: alpine:latest
    entrypoint: ["sleep"]
    command: ["9999"]
    profiles:
      - B

$ cat .env
COMPOSE_FILE=compose.yml
COMPOSE_PROJECT_NAME=profile-test
```

start compose with A,B profiles
```
$ COMPOSE_PROFILES=A,B docker compose up -d
[+] Running 3/3
 ✔️ Network profile-test_default     Created                                              0.1s
 ✔️ Container profile-test-sleep2-1  Started                                              7.6s
 ✔️ Container profile-test-sleep1-1  Started                                              7.7s
```
stop compose with A profile and --remove-orphans
```
$ COMPOSE_PROFILES=A docker compose down --remove-orphans
[+] Running 2/2
 ✔️ Container profile-test-sleep1-1  Removed                                              16.5s
 ! Network profile-test_default     Resource is still in use                               0.0s
```

same thing with up -d
```
$ COMPOSE_PROFILES=A,B docker compose up -d
[+] Running 3/3
 ✔️ Network profile-test_default     Created                                               0.1s
 ✔️ Container profile-test-sleep2-1  Started                                               1.9s
 ✔️ Container profile-test-sleep1-1  Started                                               1.9s
$ COMPOSE_PROFILES=A docker compose up -d --remove-orphans
[+] Running 1/1
 ✔️ Container profile-test-sleep1-1  Running                                               0.0s
```

Is this the correct behavior?

ndeloof:
An "orphan" is a container created by Compose but not declared in the compose file. When profile is disabled, service still is known by compose and as such not considered orphaned. Some Compose users rely on this to do partial `down` with only a subset of the services

skul87:
Сhecked it again. I removed the service with profile B from compose.yml and it works as expected - service was deleted.
Thanks for the explanation👍

I wanted control services via ansible + COMPOSE_PROFILES ENV + `up -d --remove-orphans` to delete unnecessary services)
I'll look for another solution.

skul87:
Found an ugly workaround, but it works)

```
$ cat .env
[...]

PR #13685, opened 2026-03-29
compose: add --all flag to down for full cleanup

### LINKED ISSUE #13647: [Feature] "down --all" to include all containers regardless of active profiles.
### Description

I would like an option to have a single command removing all containers, volumes, orphans, network for a given compose file. I thought `docker compose down -v -t 0 --remove-orphans` does this but as experienced it does not work with profiles. https://github.com/docker/compose/issues/13027

Example in which this would be useful:
1. Setup:
```yml
services:
  foo:
    image: alpine
    command: tail -f /dev/null
    stop_grace_period: 1s

  bar:
    profiles:
    - manual
    depends_on:
      - foo
    image: alpine
    command: tail -f /dev/null
    stop_grace_period: 1s
```

2. commands:
- `docker compose up -d bar`
- `docker compose down -t 0 -v --remove-orphans`

3. I expected that all containers for this compose project are wiped. But container bar is still there, also blocking network removal and could also block volumes.

What is the issue with this?
There is always additional knowledge required to wipe a running compose project. E.g. for a random project maybe having different profiles in it you would need something like `COMPOSE_PROFILES=$(docker compose config --profiles | tr '\n' ',') docker compose down -t 0 -v --remove-orphans` which is the only way of truly resetting it in a generic way.

Proposing `--all` for down which including all containers for the compose project, like: `docker compose down -t 0 -v --all`, all should include orphans by default. This would be consistent with other commands which provide `--all` for docker / docker compose e.g. ps or prune, so I already expected it to exist for down.

sarika-03:
## What I did

Added a new `--all` flag to `docker compose down` so users can fully clean up a Compose project, including resources created under inactive profiles.

### Changes included

- added `--all` to `docker compose down`
- made `down --all` load the project with all profiles enabled
- made `down --all` include orphan containers automatically
- ensured associated networks and volumes are also cleaned up
- added unit and e2e coverage for inactive-profile cleanup
- updated `down` command help and docs

### Manual verification

- reproduced the bug with a profiled service (`bar`)
- `docker compose down -v -t 0 --remove-orphans` left the profiled container, network, and volume behind
- `docker compose down --all -v -t 0` removed all remaining project resources

## Related issue

Fixes #13647  
Related to #13027

Screenshot from 2026-03-29 14-50-26

## Cute animal 🐳
Screenshot from 2026-03-29 14-58-51

### DIFF
--- a/cmd/compose/down.go
+++ b/cmd/compose/down.go
@@ -22,6 +22,8 @@ import (
 	"os"
 	"time"
 
+	composecli "github.com/compose-spec/compose-go/v2/cli"
+	"github.com/compose-spec/compose-go/v2/types"
 	"github.com/docker/cli/cli/command"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
@@ -34,6 +36,7 @@ import (
 
 type downOptions struct {
 	*ProjectOptions
+	all           bool
 	removeOrphans bool
 	timeChanged   bool
 	timeout       int
@@ -48,12 +51,33 @@ func downCommand(p *ProjectOptions, dockerCli command.Cli, backendOptions *Backe
 	downCmd := &cobra.Command{
 		Use:   "down [OPTIONS] [SERVICES]",
 		Short: "Stop and remove containers, networks",
+		Long: `Stops containers and removes containers, networks, volumes, and images created by up.
+
+By default, the only things removed are:
+
+- Containers for services defined in the Compose file.
+- Networks defined in the networks section of the Compose file.
+- The default network, if one is used.
+
+Networks and volumes defined as external are never removed.
+
+Anonymous volumes are not removed by default. However, as they don't have a stable name, they are not automatically
+mounted by a subsequent up. For data that needs to persist between updates, use explicit paths as bind mounts or
+named volumes.
+
+Use --all to remove every resource for the project, including services from inactive profiles and orphan containers.`,
+		Example: `docker compose down
+docker compose down -v --remove-orphans
+docker compose down --all -v`,
 		PreRunE: AdaptCmd(func(ctx context.Context, cmd *cobra.Command, args []string) error {
 			opts.timeChanged = cmd.Flags().Changed("timeout")
 			if opts.images != "" {
 				if opts.images != "all" && opts.images != "local" {
 					return fmt.Errorf("invalid value for --rmi: %q", opts.images)
 				}
+			}
+			if opts.all && len(args) > 0 {
+				return fmt.Errorf("cannot combine --all with service arguments")
 			}
 			return nil
 		}),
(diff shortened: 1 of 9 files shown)


sarika-03:
@ndeloof @glours 
please review my PR

ndeloof:
using `--profile=*` already offers this feature

sarika-03:
@ndeloof  , Thanks for pointing that out.

Using --profile="*" does allow including all profiles, but it’s not very discoverable and requires users to know this internal detail.

The goal of --all is to provide a simple and consistent way to fully clean up a Compose project, similar to other Docker commands that expose an explicit --all flag (e.g., docker system prune --all).

It avoids the need for users to reason about profiles when they just want a full reset of the project.

glours (commented):
Thanks for the contribution! The underlying problem (inactive profiles surviving down) is real and worth solving.                                                                                                                                                                                              
                                                                                                                                                                                                                                                                                                                 
 However, we have concerns about the `--all` flag as proposed:                                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                                                                 
Naming ambiguity: `--all` in existing Compose commands (`ps --all`, `ls --all`) means "include containers/projects in states normally filtered out." This PR redefines it to mean "activate all profiles + force orphan removal" which is a different semantic. Users will reasonably expect down `--all` to also remove volumes and images, but it doesn't (you still need `-v` and `--rmi`). This will lead to confusion: "I used `--all` but my volumes are still there."                                                                                                                                                          
                                                                                                                                                                                                                                                                                                                 
`--profile=*` already works: As @ndeloof pointed out, `docker compose --profile=* down -v --remove-orphans` already achieves the same result today. The gap is discoverability, not functionality.                                                                                                                 
   
 Hidden side-effects: `--all` silently forces `--remove-orphans` even if the user didn't ask for it. These should remain orthogonal flags the user composes explicitly.                                                                                                                                             
                                                            
 Suggestions:                                                                                                                                                                                                                                                                                                   
  1. Preferred: Improve discoverability of `--profile=*` by adding examples to `down --help` (e.g., `docker compose --profile='*' down -v --remove-orphans` for full cleanup)
  2. If a dedicated flag is needed: Consider `--all-profiles`, it says exactly what it does, stays orthogonal to `-v`/`--rmi`/`--remove-orphans`, and avoids the naming collision with existing `--all` semantics

sarika-03:
@glours Thanks for the detailed feedback this makes a lot of sense.

I agree that overloading `--all` introduces ambiguity, especially given its existing meaning in other Compose commands, and that it could lead to incorrect expectations around volumes/images cleanup.

Also agreed that `--profile="*"` already provides the underlying functionality, and the main gap here is discoverability and usability.

Based on your suggestion, I can update the PR to:

1. replace `--all` with a more explicit `--all-profiles` flag
2. keep it orthogonal to other flags (i.e, not implicitly enabling `--remove-orphans`)
3. ensure behavior is limited strictly to profile expansion
4 . improve help text with examples including `--profile="*"` usage

Let me know if this direction aligns better with the intended design.

glours:
@Sarika-03, as I said, I’d prefer that we improve discoverability first. We already have everything in place, we just need to make it more explicit to users. It’s a low-cost change and something we should be able to merge quickly.

We’ll still need to discuss with @ndeloof the pros and cons of adding a new flag to the command, so I’d rather not have you spend time on something that may ultimately not be merged, for now.

sarika-03:
@glours I agree 
@ndeloof please share your opinion

PR #13686, opened 2026-03-29
pkg/compose: fix TestRunHook_ConsoleSize on macOS

thaJeztah:
relates to:

- https://github.com/containerd/console/issues/79
- https://github.com/containerd/console/pull/80

containerd/console is broken on macOS, and panics; use creack/pty instead for this test.

**What I did**

**Related issue**


thaJeztah:
Thx! We should look if we can fix containerd/console, but I ran tests on my machine, and thought I broke something, then found it was already like this 😂

PR #13689, opened 2026-03-30
fix mixed assertion libraries in tests

thaJeztah:
Before this, assertion libraries were mixed, sometimes even in the same file.

    git grep -l '"gotest.tools/v3/' | wc -l
    75
    git grep -l '"github.com/stretchr/testify' | wc -l
    24

**What I did**

**Related issue**

### DIFF
--- a/pkg/utils/safebuffer.go
+++ b/pkg/utils/safebuffer.go
@@ -18,12 +18,13 @@ package utils
 
 import (
 	"bytes"
+	"fmt"
 	"strings"
 	"sync"
 	"testing"
 	"time"
 
-	"github.com/stretchr/testify/require"
+	"gotest.tools/v3/poll"
 )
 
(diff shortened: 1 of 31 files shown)


pkg/utils/safebuffer.go
```diff
@@ -64,15 +65,21 @@ func (b *SafeBuffer) Bytes() []byte {
...
+		if !strings.Contains(bufContents.String(), v) {
+			return poll.Continue(
+				"buffer does not contain %q\n============\n%s\n============",
+				v, &bufContents)
+		}
+		return poll.Success()
+	},
+		poll.WithTimeout(2*time.Second),
+		poll.WithDelay(20*time.Millisecond),
+	)
```
thaJeztah, line 84: CodeCov is complaining about a test-utility, but probably doesn't ignore it, because it's not in a `_test.go`

PR #13690, opened 2026-03-30
pkg/compose: use negotiated API version for network setup

glours:
**What I did**
This branch changes the API-version plumbing more generally, even if the immediate consumer is the container create path. A better short PR description would be:

This PR makes Compose distinguish between the raw daemon API version and the negotiated client API version.

It adds per-service caching for both values on composeService, introduces CurrentAPIVersion() for callers that need the effective negotiated version, and keeps RuntimeVersion() for raw daemon capability checks.

This avoids using ServerVersion().APIVersion in places where request shaping should follow the client’s negotiated API level.

Targeted tests were added to cover the create path and to verify negotiated API lookup is cached per service instance.

**Related issue**
N/A

image

### DIFF
--- a/pkg/compose/compose.go
+++ b/pkg/compose/compose.go
@@ -216,8 +216,7 @@ type composeService struct {
 	maxConcurrency int
 	dryRun         bool
 
-	runtimeVersion    runtimeVersionCache
-	currentAPIVersion runtimeVersionCache
+	runtimeAPIVersion runtimeVersionCache
 }
 
 // Close releases any connections/resources held by the underlying clients.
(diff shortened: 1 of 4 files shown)


pkg/compose/create.go
```diff
@@ -897,6 +897,8 @@ func (s *composeService) buildContainerVolumes(
 				}
 			}
 		case mount.TypeImage:
+			// This is a daemon capability check for mount type=image support, so
+			// use the raw server version rather than the negotiated request version.
 			version, err := s.RuntimeVersion(ctx)
```
thaJeztah: It's probably OK for a follow-up, but I suspect this should also be using the API version; a current daemon can provide older API versions, in which case it will refuse image mounts when using an older API version; https://github.com/moby/moby/blob/afb0cac1c5c0819277312b4cbccebc98220da5c4/daemon/server/router/container/container_routes.go#L657-L668

glours: Done!

pkg/compose/images.go
thaJeztah: Oh! I overlooked this one; probably same here as it's also comparing API version 🙈 🫶

glours: No worries, I didn't check neither on my side
It's fixed now!

thaJeztah:
Damn you, tests! 😂 

```console
#18 54.92 === RUN   TestImages
#18 54.92     compose.go:536: Unexpected call to *mocks.MockAPIClient.Ping([context.Background.WithCancel {true false}]) at /src/pkg/compose/compose.go:536 because: there are no expected calls of the method "Ping" for that receiver
#18 54.92     controller.go:251: missing call(s) to *mocks.MockAPIClient.ImageInspect(is assignable to , is equal to bar:2 (string)) /src/pkg/compose/images_test.go:52
#18 54.92     controller.go:251: aborting test due to missing call(s)
#18 54.92 --- FAIL: TestImages (0.00s)
```

pkg/compose/compose.go
```diff
@@ -491,22 +494,62 @@ func (s *composeService) isSwarmEnabled(ctx context.Context) (bool, error) {
...
+	mu  sync.Mutex
+	val string
 }
 
-var runtimeVersion runtimeVersionCache
-
+// RuntimeVersion returns the raw API version reported by the daemon.
+// Callers that need the negotiated/effective client API version should use
+// CurrentAPIVersion instead.
 func (s *composeService) RuntimeVersion(ctx context.Context) (string, error) {
```
ndeloof: We should only rely on CurrentAPIVersion, as after API has been negociated, if for some reason uses a version < runtime version, compose should never rely on features or request attributes not defined by this API version.
I suggest you merge both functions into `RuntimeAPIVersion` and use it everywhere we depend on Engine's features

glours: Indeed, I'll do it right now

PR #13713, opened 2026-04-04
fix: ensure deterministic variable interpolation (#13712)

yunus25jmi1:
## Description

This fixes the non-deterministic behavior when multiple required variables (e.g., ${TIMEZONE:?}, ${TRAEFIK_ACME_PATH:?}) are missing in docker-compose.yaml.

## Root Cause

Go's map iteration order is randomized. When iterating over service environment variables or volumes during interpolation, different keys can be processed in different orders on each run. This causes the error message for missing required variables to vary between runs.

## Fix

The fix adds sort.Strings(keys) before map iteration in the compose-go library's interpolation.go file to ensure deterministic processing order.

### Changes:

1. **compose-go fix** (already pushed to https://github.com/yunus25jmi1/compose-go):
   - Added sort import to interpolation.go
   - Sort map keys before iteration in Interpolate() function
   - Sort map keys before iteration in recursiveInterpolate() function

2. **docker/compose** (this PR):
   - Added replace directive to use the fixed compose-go fork

## Testing

- go build: passes
- go vet: passes  
- go fmt: passes
- Unit tests: pass

## Related

- Fixes https://github.com/docker/compose/issues/13712


yunus25jmi1:
Thank you for the review! You raise valid concerns about the replace directive approach.

**Plan:**
1. The compose-go fix is in PR https://github.com/compose-spec/compose-go/pull/860 - once merged, we will update docker/compose to use the official compose-go version
2. This replace directive is a temporary measure to validate the fix works
3. After compose-go PR merges, I will create a follow-up PR to:
   - Remove the replace directive
   - Update the compose-go dependency to the version with the fix

The current approach allows testing the fix while waiting for upstream compose-go merge. I will add a TODO comment with a tracking link as suggested.

yunus25jmi1:
Hi @ndeloof @glours and docker/compose-maintainers,

I've updated this PR to address the supply-chain risk concern raised in the review. The replace directive pointing to the personal fork (github.com/yunus25jmi1/compose-go/v2) has been removed, and we're now using the official compose-go v2.10.2 from upstream.

This eliminates the supply-chain risk of depending on a personal fork.

Note: The deterministic interpolation fix from compose-spec/compose-go#860 is not yet merged upstream. Once that PR is merged and released, we'll need to update to the version that includes the fix.

Please review the updated changes.

glours:
@yunus25jmi1 What do you want us to review? PR is empty

PR #13721, opened 2026-04-08
feat: add Docker Desktop Logs view hints and navigation shortcut

glours:
**What I did**

- Add CLI hooks handler to show "What's next:" hints pointing to the Docker Desktop Logs view after `docker logs`, `docker compose logs`, and `docker compose up -d`.

- Add `l` keyboard shortcut in the `compose up` navigation menu to open the Logs view, gated on Docker Desktop feature flag and settings.

**Related issue**
N/A

image

### DIFF
--- a/cmd/compose/hooks.go
+++ b/cmd/compose/hooks.go
@@ -1,0 +1,101 @@
+/*
+   Copyright 2020 Docker Compose CLI authors
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package compose
+
+import (
+	"encoding/json"
+	"io"
+
+	"github.com/docker/cli/cli-plugins/hooks"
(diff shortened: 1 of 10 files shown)


pkg/compose/up.go
```diff
@@ -88,8 +89,9 @@ func (s *composeService) Up(ctx context.Context, project *types.Project, options
 			if err != nil {
 				return err
 			}
-			tracing.KeyboardMetrics(ctx, options.Start.NavigationMenu, isDockerDesktopActive)
-			navigationMenu = formatter.NewKeyboardManager(isDockerDesktopActive, signalChan)
+			isLogsViewEnabled := s.isDesktopFeatureActive(ctx, desktop.FeatureLogsTab)
```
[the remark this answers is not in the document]
glours: It's ok, calls are almost instant to the local socket

cmd/compose/hooks.go
```diff
@@ -0,0 +1,101 @@
...
+// When checkFlags is set, the hint is only returned if the check passes.
+type hookHint struct {
+	template   string
+	checkFlags func(flags map[string]string) bool
+}
+
+// hooksHints maps hook root commands to their hint definitions.
+var hooksHints = map[string]hookHint{
+	// standalone "docker logs" (not a compose subcommand)
+	"logs":         {template: dockerLogsHint},
```
ndeloof, line 45: unclear to me why we provide `docker logs` hint inside compose. Maybe _just_ because we have no place in docker/cli for this and compose used to be installed by 99% users?

ndeloof (approved):

cmd/compose/hooks.go
glours: It's because Compose is in charge of this hint globally, it could trigger the hint display to any other cli plugins too.

PR #13728, opened 2026-04-11
Fix initial_sync for existing watch files

### LINKED ISSUE #13725: [BUG] develop.watch initial_sync: true does not fire at container start
### Description

### Steps To Reproduce

`docker-compose.yml`:
```yaml
services:
  test:
    image: python:3.14-slim
    command: sleep infinity
    develop:
      watch:
        - path: ./src
          action: sync
          target: /app/src
          initial_sync: true
```

```bash
docker compose up --watch -d
sleep 30

touch src/test.txt
sleep 10
```

### Observed Behavior

### Expected Behavior

### Environment

- macOS 15.4 (Darwin 25.4.0)

### Notes

trinhchien:
## Summary
- make `initial_sync` copy existing files regardless of image creation time
- add tests covering both directory and single-file watch paths

## Repro
Fixes #13725. Before this change, `initial_sync: true` could skip files that already existed on the host if their mtime was older than the service image creation time. Change-triggered syncs still worked, but the initial startup sync could be empty.

## Testing
- `go test ./pkg/compose -run TestInitialSyncFiles_ -v`

## AI usage
Used AI assistance to help analyze the repro, shape the initial patch, and draft the tests. I reviewed and validated the final code and test results manually.

### DIFF
--- a/pkg/compose/watch.go
+++ b/pkg/compose/watch.go
@@ -248,7 +248,7 @@ func (s *composeService) watch(ctx context.Context, project *types.Project, opti
 
 				if shouldInitialSync && isSync(trigger) {
 					// Need to check initial files are in container that are meant to be synced from watch action
-					err := s.initialSync(ctx, project, service, trigger, syncer)
+					err := s.initialSync(ctx, service, trigger, syncer)
 					if err != nil {
 						return nil, err
 					}
@@ -730,7 +730,7 @@ func (s *composeService) pruneDanglingImagesOnRebuild(ctx context.Context, proje
 
 // Walks develop.watch.path and checks which files should be copied inside the container
 // ignores develop.watch.ignore, Dockerfile, compose files, bind mounted paths and .git
-func (s *composeService) initialSync(ctx context.Context, project *types.Project, service types.ServiceConfig, trigger types.Trigger, syncer sync.Syncer) error {
+func (s *composeService) initialSync(ctx context.Context, service types.ServiceConfig, trigger types.Trigger, syncer sync.Syncer) error {
 	dockerIgnores, err := watch.LoadDockerIgnore(service.Build)
 	if err != nil {
 		return err
@@ -752,7 +752,7 @@ func (s *composeService) initialSync(ctx context.Context, project *types.Project
 		dotGitIgnore,
 		triggerIgnore)
 
-	pathsToCopy, err := s.initialSyncFiles(ctx, project, service, trigger, ignoreInitialSync)
+	pathsToCopy, err := s.initialSyncFiles(service, trigger, ignoreInitialSync)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


pkg/compose/watch.go
glours: With the change this function is now deadcode and should be removed

glours: With the changes this is no longer accurate

glours: `ctx` and `project` aren't used anymore
```suggestion
func (s *composeService) initialSyncFiles(service types.ServiceConfig, trigger types.Trigger, ignore watch.PathMatcher) ([]*sync.PathMapping, error) {
```

pkg/compose/watch_test.go
```diff
@@ -194,3 +195,71 @@ func (f *fakeSyncer) Sync(ctx context.Context, service string, paths []*sync.Pat
 	f.synced <- paths
 	return nil
 }
+
+func TestInitialSyncFiles_DirectoryIncludesExistingFilesEvenIfOlderThanImage(t *testing.T) {
+	mockCtrl := gomock.NewController(t)
+	cli := mocks.NewMockCli(mockCtrl)
+	cli.EXPECT().Err().Return(streams.NewOut(os.Stderr)).AnyTimes()
+	apiClient := mocks.NewMockAPIClient(mockCtrl)
+	cli.EXPECT().Client().Return(apiClient).AnyTimes()
```
glours: No more reference to the api client, so we don't need the mock

pkg/compose/watch_test.go
```diff
@@ -194,3 +195,71 @@ func (f *fakeSyncer) Sync(ctx context.Context, service string, paths []*sync.Pat
...
+	}}
+	assert.DeepEqual(t, expected, got)
+}
+
+func TestInitialSyncFiles_FileIncludesExistingFileEvenIfOlderThanImage(t *testing.T) {
+	mockCtrl := gomock.NewController(t)
+	cli := mocks.NewMockCli(mockCtrl)
+	cli.EXPECT().Err().Return(streams.NewOut(os.Stderr)).AnyTimes()
+	apiClient := mocks.NewMockAPIClient(mockCtrl)
+	cli.EXPECT().Client().Return(apiClient).AnyTimes()
```
glours: same here

pkg/compose/watch_test.go
```diff
@@ -194,3 +195,71 @@ func (f *fakeSyncer) Sync(ctx context.Context, service string, paths []*sync.Pat
...
+
+	dir := t.TempDir()
+	hostFile := filepath.Join(dir, "test.txt")
+	assert.NilError(t, os.WriteFile(hostFile, []byte("hello"), 0o644))
+
+	fileTime := time.Date(2026, 4, 1, 10, 0, 0, 0, time.UTC)
```
glours:
```suggestion
	fileTime := time.Now().Add(-24*time.Hour)
```

pkg/compose/watch_test.go
```diff
@@ -194,3 +195,71 @@ func (f *fakeSyncer) Sync(ctx context.Context, service string, paths []*sync.Pat
...
+	apiClient := mocks.NewMockAPIClient(mockCtrl)
+	cli.EXPECT().Client().Return(apiClient).AnyTimes()
+
+	dir := t.TempDir()
+	hostDir := filepath.Join(dir, "src")
+	assert.NilError(t, os.MkdirAll(hostDir, 0o755))
+	hostFile := filepath.Join(hostDir, "test.txt")
+	assert.NilError(t, os.WriteFile(hostFile, []byte("hello"), 0o644))
+
+	fileTime := time.Date(2026, 4, 1, 10, 0, 0, 0, time.UTC)
```
glours:
```suggestion
	fileTime := time.Now().Add(-24*time.Hour)
```

glours (changes requested):
Hey @trinhchien 
Thanks for the contribution, there is few details to adapt in your PR. Can you take a look?

trinhchien:
Thanks for the review — I addressed the cleanup comments, simplified the tests as suggested, and updated the commit with a proper DCO sign-off.

pkg/compose/watch.go
```diff
@@ -752,26 +752,22 @@ func (s *composeService) initialSync(ctx context.Context, project *types.Project
...
 		return err
 	}
 
 	return syncer.Sync(ctx, service.Name, pathsToCopy)
 }
 
-// Syncs files from develop.watch.path if thy have been modified after the image has been created
+// Syncs files from develop.watch.path, ignoring bind-mounted and excluded paths.
 //
 //nolint:gocyclo
```
glours: nit: not needed anymore

glours (commented):
You need to sign-off your commit for legal reasons otherwise I won't be able to merge your contribution

trinhchien:
Thanks — I added a small regression test to cover the initialSync path as well, and updated the branch.

PR #13734, opened 2026-04-14
feat: make hook hint deep links clickable using OSC 8 terminal hyperlinks

glours:
**What I did**

Wrap the docker-desktop://dashboard/logs URL in OSC 8 escape sequences with underline styling so it appears as a clickable link in supported terminals. Respects NO_COLOR and COMPOSE_ANSI=never to suppress escapes.

**Related issue**
N/A

image

### DIFF
--- a/cmd/compose/hooks.go
+++ b/cmd/compose/hooks.go
@@ -19,23 +19,58 @@ package compose
 import (
 	"encoding/json"
 	"io"
+	"os"
 
 	"github.com/docker/cli/cli-plugins/hooks"
 	"github.com/docker/cli/cli-plugins/metadata"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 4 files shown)


afurm:
The `template string` field on `hookHint` was changed to `template func() string`, and callers switched to `hint.template()` — this shifts from a static string to a lazy evaluation. The original design appeared to call `enc.Encode` once per hook invocation, so lazy evaluation should be fine here. However, if `hookHint` is ever used in a way where the template is inspected rather than rendered (e.g., log output, JSON marshaling of the struct itself), the function would not serialize correctly. Does any code path marshal `hookHint` directly to JSON?

ndeloof (approved):

PR #13742, opened 2026-04-16
Add rawsetenv message type for provider plugins

### LINKED ISSUE #13727: Provider services: cannot inject environment variables without service name prefix
## Description

I'm building a Compose provider plugin that injects secrets from external vaults into dependent services. The goal is to replace wrapper scripts (e.g., `chamber exec`, `doppler run --`, `op run --`) with a native Compose integration:

```yaml
services:
  secrets:
    provider:
      type: my-vault-provider
      options:
        backend: aws-ssm
        param:
          - AUTH0_CLIENT_SECRET=/app/stg/auth0_client_secret

  app:
    depends_on:
      - secrets
```

### Problem

`setenv` variables are always prefixed with the service name, so `AUTH0_CLIENT_SECRET` becomes `SECRETS_AUTH0_CLIENT_SECRET`. Many applications and frameworks require exact variable names (e.g., `AUTH0_CLIENT_SECRET` for Auth0, `NEXT_PUBLIC_*` for Next.js), and there is no way to inject them as-is.

`${SECRETS_AUTH0_CLIENT_SECRET}` in the `environment` section doesn't work either, since it resolves at parse time before the provider runs.

### Proposal

Add a `rawsetenv` message type to the provider protocol that injects variables without the service name prefix:

```json
{"type": "rawsetenv", "message": "AUTH0_CLIENT_SECRET=xxx"}
```

The provider binary decides whether each variable should be prefixed (`setenv`) or injected as-is (`rawsetenv`).

Possible implementation in `pkg/compose/plugins.go`:

```go
type envVar struct {
    value    string
    prefixed bool
}

// During plugin output parsing:
case SetEnvType:
    variables[key] = envVar{val, true}
case RawSetEnvType:
    variables[key] = envVar{val, false}

// When injecting into dependent services:
prefix := strings.ToUpper(service.Name) + "_"
for key, v := range variables {
    if v.prefixed {
        s.Environment[prefix+key] = &v.value
    } else {
        s.Environment[key] = &v.value
    }
}
```

rajyan:
## What I did

Added a new `rawsetenv` message type to the provider plugin protocol. This allows providers to inject environment variables into dependent services **without** the automatic service name prefix.

Currently, `setenv` always prefixes variables with the service name (e.g., `URL` becomes `DATABASE_URL`). This works well for connection strings, but some applications require exact variable names that cannot be altered. There is no way to inject these as-is today.

With this change, providers can choose per-variable whether to use the prefixed (`setenv`) or unprefixed (`rawsetenv`) behavior:

```json
{"type": "setenv", "message": "URL=https://example.com"}
{"type": "rawsetenv", "message": "SECRET_KEY=xxx"}
```

### Changes

- `pkg/compose/plugins.go` — Add `RawSetEnvType` constant and `pluginVariables` struct to separate prefixed/raw variables
- `docs/extension.md` — Document `rawsetenv` in the protocol specification
- `docs/examples/provider.go` — Add `rawsetenv` usage to the example provider
- `pkg/e2e/providers_test.go` — Test for single-provider rawsetenv and multi-provider rawsetenv
- `pkg/e2e/fixtures/providers/rawsetenv.yaml` — Test fixture

## Related issue

resolves #13727

### DIFF
--- a/docs/extension.md
+++ b/docs/extension.md
@@ -72,6 +72,8 @@ sequenceDiagram
     Compose-)Shell: pulling 75%
     Provider--)Compose: json { "setenv": "URL=http://cloud.com/abcd:1234" }
     Compose-)Compose: set DATABASE_URL
+    Provider--)Compose: json { "rawsetenv": "SECRET_KEY=xxx" }
+    Compose-)Compose: set SECRET_KEY (as-is)
     Provider-)Compose: EOF (command complete) exit 0
     Compose-)Shell: service started
 ```
@@ -128,7 +130,7 @@ The provider is responsible for releasing all resources associated with the serv
 When the user runs `docker compose stop`, Compose invokes `<provider> compose --project-name <NAME> stop <SERVICE>` for each
 provider-backed service in reverse dependency order. The provider should pause the resource without releasing it, so a later
 `docker compose up` can resume it (note that `docker compose start` only restarts existing containers and does not invoke
-provider hooks). Any `setenv` JSON message returned during `stop` is ignored, since dependent services are also stopping.
+provider hooks). Any `setenv` or `rawsetenv` JSON message returned during `stop` is ignored, since dependent services are also stopping.
 
 The `stop` hook is opt-in: Compose invokes it only when the provider declares a `stop` block in its `metadata` subcommand
 output. Providers that do not advertise `stop` in metadata (or do not implement the `metadata` subcommand at all) are
--- a/docs/extension.md
+++ b/docs/extension.md
@@ -106,9 +106,14 @@ When the provider command sends a `rawsetenv` JSON message, Compose injects the 
 ```
 The `app` service will receive `SECRET_KEY` exactly as specified, regardless of the provider service name.
 This is useful when injecting secrets or configuration values that must match exact variable names expected by
-applications or frameworks. Unlike `setenv`, which avoids collisions through automatic prefixing, `rawsetenv` keys
-are the provider's responsibility to keep unique. If multiple providers emit the same `rawsetenv` key, the last one
-to run will overwrite previous values.
+applications or frameworks.
+
+Unlike `setenv`, which avoids collisions through automatic prefixing, `rawsetenv` keys are the provider's
+responsibility to keep unique. If a `rawsetenv` key collides with a variable already set on the dependent service,
+the existing value is overwritten and Compose logs a warning. This includes variables declared by the user in the
+service `environment` section as well as values emitted by other providers. Providers that are not linked by a
+`depends_on` relationship may run concurrently, so when several of them emit the same `rawsetenv` key the resulting
+value is not deterministic.
 
 > __Note:__  The `compose up` provider command _MUST_ be idempotent. If resource is already running, the command _MUST_ set
 > the same environment variables to ensure consistent configuration of dependent services.
--- a/docs/extension.md
+++ b/docs/extension.md
@@ -56,7 +56,8 @@ JSON messages MUST include a `type` and a `message` attribute.
 `type` can be either:
 - `info`: Reports status updates to the user. Compose will render message as the service state in the progress UI
 - `error`: Lets the user know something went wrong with details about the error. Compose will render the message as the reason for the service failure.
-- `setenv`: Lets the plugin tell Compose how dependent services can access the created resource. See next section for further details.
+- `setenv`: Lets the plugin tell Compose how dependent services can access the created resource. The variable is automatically prefixed with the service name. See next section for further details.
+- `rawsetenv`: Same as `setenv`, but the variable is injected as-is without the service name prefix. Useful when applications require exact variable names that cannot be altered.
 - `debug`: Those messages could help debugging the provider, but are not rendered to the user by default. They are rendered when Compose is started with `--verbose` flag.
 
 ```mermaid
@@ -98,6 +99,16 @@ automatically prefixing it with the service name. For example, if `awesomecloud 
 ```
 Then the `app` service, which depends on the service managed by the provider, will receive a `DATABASE_URL` environment variable injected
 into its runtime environment.
+
+When the provider command sends a `rawsetenv` JSON message, Compose injects the variable as-is without any prefix:
+```json
+{"type": "rawsetenv", "message": "SECRET_KEY=xxx"}
+```
+The `app` service will receive `SECRET_KEY` exactly as specified, regardless of the provider service name.
+This is useful when injecting secrets or configuration values that must match exact variable names expected by
+applications or frameworks. Unlike `setenv`, which avoids collisions through automatic prefixing, `rawsetenv` keys
+are the provider's responsibility to keep unique. If multiple providers emit the same `rawsetenv` key, the last one
+to run will overwrite previous values.
 
 > __Note:__  The `compose up` provider command _MUST_ be idempotent. If resource is already running, the command _MUST_ set
 > the same environment variables to ensure consistent configuration of dependent services.
--- a/pkg/compose/plugins.go
+++ b/pkg/compose/plugins.go
@@ -94,7 +94,7 @@ func (s *composeService) runPlugin(ctx context.Context, project *types.Project, 
 				s.Environment[prefix+key] = &val
 			}
 			for key, val := range variables.raw {
-				if existing, ok := s.Environment[key]; ok && existing != nil && *existing != val {
+				if existing, ok := s.Environment[key]; ok && (existing == nil || *existing != val) {
 					logrus.Warnf("provider %q overrides environment variable %q in service %q", service.Name, key, name)
 				}
 				s.Environment[key] = &val
--- a/pkg/compose/plugins.go
+++ b/pkg/compose/plugins.go
@@ -76,7 +76,7 @@ func (s *composeService) runPlugin(ctx context.Context, project *types.Project, 
 		return nil
 	}
 
-	vars, err := s.executePlugin(cmd, command, service)
+	variables, err := s.executePlugin(cmd, command, service)
 	if err != nil {
 		return err
 	}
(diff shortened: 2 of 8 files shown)


phillias:
this patch worked for my issue with mariadb requiring MYSQL_ROOT_PASSWORD with no override, not LOCKET_MYSQL_ROOT_PASSWORD as the compose provider was passing.

pkg/compose/plugins.go
```diff
@@ -70,7 +76,7 @@ func (s *composeService) runPlugin(ctx context.Context, project *types.Project,
 		return nil
 	}
 
-	variables, err := s.executePlugin(cmd, command, service)
+	vars, err := s.executePlugin(cmd, command, service)
```
glours: Why this rename?

pkg/compose/plugins.go
```diff
@@ -84,16 +90,19 @@ func (s *composeService) runPlugin(ctx context.Context, project *types.Project,
 	for name, s := range project.Services {
 		if _, ok := s.DependsOn[service.Name]; ok {
 			prefix := strings.ToUpper(service.Name) + "_"
-			for key, val := range variables {
+			for key, val := range vars.prefixed {
 				s.Environment[prefix+key] = &val
 			}
+			for key, val := range vars.raw {
+				s.Environment[key] = &val
```
glours, line 100: `rawsetenv` silently overwrites user-defined env vars                                                                                                                                                                                                                                                

If a user sets `environment: { CLOUD_REGION: eu-west-1 }` and a provider emits `rawsetenv CLOUD_REGION=us-east-1`, the user value is silently clobbered. `setenv` was immune thanks to the prefix. 
Please either skip writes when the key already exists, or we need to document this precedence explicitly. 

@ndeloof what is you preference here? I'm in favor of the overwrite + warning message in logs, and you?

docs/extension.md
```diff
@@ -99,6 +100,16 @@ automatically prefixing it with the service name. For example, if `awesomecloud
...
 
+When the provider command sends a `rawsetenv` JSON message, Compose injects the variable as-is without any prefix:
+```json
+{"type": "rawsetenv", "message": "SECRET_KEY=xxx"}
+```
+The `app` service will receive `SECRET_KEY` exactly as specified, regardless of the provider service name.
+This is useful when injecting secrets or configuration values that must match exact variable names expected by
+applications or frameworks. Unlike `setenv`, which avoids collisions through automatic prefixing, `rawsetenv` keys
+are the provider's responsibility to keep unique. If multiple providers emit the same `rawsetenv` key, the last one
+to run will overwrite previous values.
```
glours: `executor.go` runs plan nodes concurrently via `errgroup`. Two providers without a mutual dependency run in parallel; `mux` serializes the writes but not their order. The doc says "the last one to run will overwrite previous values", literally true, but readers will assume declaration order wins.  
Either tighten the doc, or detect cross-provider `rawsetenv` collisions and fail.

pkg/e2e/providers_test.go
```diff
@@ -76,6 +76,27 @@ func TestDependsOnMultipleProviders(t *testing.T) {
 	env := getEnv(res.Combined(), false)
 	assert.Check(t, slices.Contains(env, "PROVIDER1_URL=https://magic.cloud/provider1"), env)
 	assert.Check(t, slices.Contains(env, "PROVIDER2_URL=https://magic.cloud/provider2"), env)
+	assert.Check(t, slices.Contains(env, "CLOUD_REGION=us-east-1"), env)
```
glours: both providers emit the same `CLOUD_REGION` value, so the conflict path is never exercised. Please add a test where a `rawsetenv` key collides with a user-defined env var

glours (changes requested):
Few things to address before we can merge. 
Also we need to add a sibling fixture to `pkg/e2e/fixtures/providers/rawsetenv.yaml` that sets environment: { CLOUD_REGION: user-value } on the test service, and assert that the resulting value matches the documented behavior (skipped write, warning, or overwrite, depending on which resolution is chosen).

pkg/compose/plugins.go
rajyan: There's no good reason for it. Reverted to variables.

pkg/compose/plugins.go
rajyan: Went with overwrite + warning as suggested. One clarification on the framing though: setenv normally doesn't overwrite anything, since the prefix gives each provider its own namespace and collisions are rare. But the write has no existence check either, so if a service does define the exact prefixed key (e.g. a user-set SECRETS_URL and a provider named "secret" set URL), setenv silently overwrites it the same way. I kept the warning on rawsetenv only since prefixed collisions are unlikely, but can extend it to setenv if you'd prefer consistency.

docs/extension.md
rajyan: Reworded. Dropped the "last one to run will overwrite" phrasing and now state that providers without a depends_on link may run concurrently, so the resulting value is non-deterministic on collision, and that an override logs a warning.

pkg/e2e/providers_test.go
rajyan: Removed the CLOUD_REGION assertion from TestDependsOnMultipleProviders since, as you noted, both providers emit the same value so it never exercised a conflict, and a real cross-provider conflict can't be asserted deterministically (independent providers run concurrently, so it would be flaky). The deterministic conflict is now covered by TestProviderRawSetEnvOverridesUserEnv, which asserts the provider value wins over a user-defined one and that the override is logged as a warning rather than happening silently. The non-deterministic cross-provider case is documented in extension.md.

rajyan:
Hi @glours, I've addressed all the review comments. Would appreciate another look when you get a chance. Thanks!

pkg/compose/plugins.go
```diff
@@ -84,16 +90,22 @@ func (s *composeService) runPlugin(ctx context.Context, project *types.Project,
 	for name, s := range project.Services {
 		if _, ok := s.DependsOn[service.Name]; ok {
 			prefix := strings.ToUpper(service.Name) + "_"
-			for key, val := range variables {
+			for key, val := range variables.prefixed {
 				s.Environment[prefix+key] = &val
 			}
+			for key, val := range variables.raw {
+				if existing, ok := s.Environment[key]; ok && existing != nil && *existing != val {
```
glours: Follow-up on the previous collision fix: the `existing != nil` guard short-circuits when the user declares the env var without a value, e.g.:
                                                                                                                                                                                                                                                            
  ```yaml
  environment:                                                                                                                                                                                                                                              
    - CLOUD_REGION                                          
  # or
  environment:
    CLOUD_REGION:
  ```                                                                                                                                                                                                                                                       
   
  `MappingWithEquals` stores that as `*string == nil`, meaning "inherit from the shell environment at runtime". A `rawsetenv` for the same key then overwrites it silently — no warning fires. Same class of bug as the case I flagged before, just exposed 
  by a different YAML form (and arguably the more common form when users want the host shell value to flow through).
                                                                                                                                                                                                                                                            
Could you extend the condition to cover the nil case as well?                                                                                                                                                                                             
```suggestion
				if existing, ok := s.Environment[key]; ok && (existing != nil || *existing != val) {
```
                                                                                                                                                                                                                                                            
An additional e2e fixture that declares `environment: [- CLOUD_REGION]` and asserts the warning fires would lock the behaviour in.

docs/extension.md
glours: Now that `rawsetenv` is a defined message type, this sentence should mention it too. The code at `plugins.go:84-86` ignores both `setenv` and `rawsetenv` when `command == "stop"`, but the doc reads as if only `setenv` is dropped.
                                                                                                                                                                                                                                                            
Suggested wording: `Any `setenv` or `rawsetenv` JSON message returned during `stop` is ignored, since dependent services are also stopping.`

glours, line 185: Could you extend the sequence diagram to show `rawsetenv` as well? Right now plugin authors reading the diagram alone would not know it exists.

glours (changes requested):
Almost there, thanks for the iteration on the collision handling. A couple of follow-ups before we can merge:                                                                                                                                            
                                                                                                                                                                                                                                                            
  1. The collision fix covers `KEY: value` but not the `- KEY` / `KEY:` (inherit-from-shell) form — see inline comment on `plugins.go`. Same class of bug as the one I flagged previously, just exposed by a different YAML syntax.                         
  2. `extension.md` needs a couple of doc tweaks so the protocol contract is consistent — the `stop` section still mentions only `setenv`, and the mermaid sequence diagram doesn't show `rawsetenv`.

pkg/compose/plugins.go
rajyan: Thanks for catching the nil-pointer case — I hadn't considered the - KEY / KEY: form. Fixed the condition to existing == nil || *existing != val so the warning fires for shell-inherited env vars as well.

Added e2e tests for both list-style (- KEY) and map-style (KEY:) declarations that assert the warning message, which fail against the previous code and pass with the fix.

rajyan:
@glours Thanks for the review! Addressed the follow-ups:

1. Fixed the nil-pointer guard for - KEY / KEY: form and added e2e tests.
2. Updated the stop section to mention rawsetenv and added it to the mermaid sequence diagram.

glours (dismissed):
We're good to go! Thank you very much for this contribution @rajyan 🙏 

edit: It seems the CI validation step disagree with me 😬

rajyan:
@glours The lint CI is failing with `unparam` on the `getEnv` function in `pkg/e2e/providers_test.go`:

```
pkg/e2e/providers_test.go:158:25: getEnv - run always receives false (unparam)
```

This `getEnv(out string, run bool)` function already existed on `main` with the `run` parameter always being `false`, but lint passes on `main` — likely because there's only one call site, so `unparam` doesn't flag it. My PR added more test cases calling `getEnv(..., false)`, which seems to have triggered the detection.

Since this is a pre-existing issue, how would you prefer I handle it? Should I remove the `run` parameter and the dead code branch in this PR, or would you prefer a separate fix?

glours:
@rajyan I did a PR to fix it, I'll let you know when merged so you could rebase your branch

rajyan:
@glours Thanks for the quick fix! I'll rebase once it's merged.

I also noticed the Codecov check is failing. It looks like it might be due to not having the latest `main` changes, so rebasing should fix that as well.

glours:
@rajyan it's merged, you can rebase now

rajyan:
@glours 
Thank you! Rebased with the latest main.

glours (approved):
LGTM, thanks @rajyan

PR #13746, opened 2026-04-17
fix: don't hang forever in Pipe.Confirm when stdin has no data

### LINKED ISSUE #13722: [BUG] --yes or -y flag not being respected on docker compose publish on Azure DevOps pipeline
### Description

The same as #13148 
The `docker compose publish` command is not respecting the `-y` or `--yes` flag as indicated in the documentation, when used in an Azure DevOps pipeline running on MS-hosted ubuntu-24.04 agent.

### Steps To Reproduce

1. Create a yml file with:

```yml
services:
  bindwarn:
    image: busybox:latest
    command: ["sh", "-c", "sleep 5"]
    volumes:
      - .:/data:ro # bind mount to trigger the prompt
```

2. Run `docker compose -f compose.yml publish -y "some-upstream-url"`
(the same happens if you use --yes)

3. You will see the output:

```sh
you are about to publish bind mounts declaration within your OCI artifact.
only the bind mount declarations will be added to the OCI artifact (not content)
please double check that you are not mounting potential user's sensitive directories or data
bindwarn
/Users/dkloumar/work/docker-bug:/data:ro
```

and it waits forever.
If you cancel the pipeline running job, other lines appear on the output

```sh
##[error]The Operation will be canceled. The next steps may not contain expected logs.
Are you ok to publish these bind mount declarations? [y/N]: 
##[error]Bash exited with code 'null'.
##[error]The operation was canceled
```

Indicating that the -y flag was not respected.

### Compose Version

```Text
Docker Compose version v2.40.3
```

### Docker Environment

```Text
 Version:           28.0.4
 API version:       1.48
 Go version:        go1.23.7
 Git commit:        b8034c0
 Built:             Tue Mar 25 15:07:16 2025
 Context:           default

 Engine:
  Version:          28.0.4
  API version:      1.48 (minimum version 1.24)
  Go version:       go1.23.7
  Git commit:       6430e49
  Built:            Tue Mar 25 15:07:16 2025
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          v2.2.2
  GitCommit:        301b2dac98f15c27117da5c8af12118a041a31d9
 runc:
  Version:          1.3.4
  GitCommit:        v1.3.4-0-gd6d73eb8
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
Docker compose version:
Docker Compose version v2.40.3
```

### Anything else?

I can run the pipeline if I run
`yes | docker compose -f compose.yml publish -y "some-upstream-url"`

ssam18:
When running `docker compose publish` in a CI environment like Azure DevOps, stdin is often an open pipe that never sends data, causing `fmt.Fscanln` to block indefinitely even when `-y` is passed. This fix makes `Pipe.Confirm` return the default value immediately when stdin returns an error (e.g. EOF or closed pipe) instead of hanging. Added a test to cover the `--yes` path so a prompt with bind mounts doesn't block. Fixes #13722


glours (changes requested):
Hi @ssam18, thanks for the contribution. I took a close look and I have some concerns before we can move forward.                                                                                                                                                                                            
                                                                                                                                                                                                                                                                                                               
  ### 1. The change doesn't match the reported bug                                                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                                                               
  Issue #13722 is about `-y` not being respected. But in `cmd/compose/publish.go`:                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                                                               
  ```go                                                                                                                                                                                                                                                                                                        
  if opts.assumeYes {
      backendOptions.Options = append(backendOptions.Options, compose.WithPrompt(compose.AlwaysOkPrompt()))
  }
  ```                                                                                                                                                                                                                                                                                                          
   
  When `-y` is passed, the prompt is replaced with `AlwaysOkPrompt`, which returns `(true, nil)` without ever reading stdin. `Pipe.Confirm` is not reached. So the code you changed isn't on the `-y` path, and your PR description saying *"causing `fmt.Fscanln` to block indefinitely even when `-y` is     
  passed"* doesn't line up with what the code actually does.
                                                                                                                                                                                                                                                                                                               
  Could you share a stack trace or reproducer showing `Pipe.Confirm` being hit with `-y`? That would help us confirm whether the real bug is somewhere else (for example, option ordering, a missed prompt path, or a different confirmation like `confirmRemoteIncludes`).                                    
   
  ### 2. The fix may not actually prevent the hang                                                                                                                                                                                                                                                             
                  
  Your description says *"stdin is often an open pipe that never sends data"*. In that case `fmt.Fscanln` blocks waiting for input and never returns an error, so the new `if err != nil` branch is never reached. The change only alters behavior when `Fscanln` returns an error (EOF or read error), and in
   that case the previous code already returned `(false, nil)` via `StringToBool("")`. For the `preChecks` callers (which pass `defaultValue=false`), the runtime behavior is unchanged.
                                                                                                                                                                                                                                                                                                               
  If `Pipe.Confirm` genuinely needs to not hang on an idle pipe, the fix likely needs to be non-blocking — e.g., a deadlined read on the underlying fd, or detecting upfront that stdin is `/dev/null`/closed, not an error check after `Fscanln` returns.                                                    
   
  ### 3. The test doesn't exercise the changed code                                                                                                                                                                                                                                                            
                  
  `Test_preChecks_bind_mount_skipped_with_assume_yes` uses `AlwaysOkPrompt()`, which doesn't touch `Pipe.Confirm`. It lives in a different package than your fix and would pass identically with or without the change to `prompt.go`. It's not a regression test for the behavior you're claiming to fix.     
   
  ### Suggested path forward                                                                                                                                                                                                                                                                                   
                  
  Before we change `Pipe.Confirm`, could you:

  - Confirm with the issue reporter whether Azure DevOps closes stdin (EOF) or leaves it open — that determines where the hang actually is.                                                                                                                                                                    
  - Add logs/trace showing which code path is reached when `-y` is passed.
  - If the root cause really is in `Pipe.Confirm`, provide a test that drives `Pipe.Confirm` directly (with a blocking/EOF reader) and demonstrates the hang and its fix.

ssam18:
@glours  You are right. I was wrong to change Pipe.Confirm. With -y set, AlwaysOkPrompt replaces s.prompt entirely so Pipe.Confirm is never in the picture, and even if it were, an open idle pipe never returns an error so the if err != nil branch wouldn't trigger.

I will revert that change. The only thing left in the PR is the test, which verifies that AlwaysOkPrompt (the prompt wired up by --yes) lets preChecks pass through the bind-mount check without blocking, a regression guard rather than a fix.

I don't have access to reproduce the issue on an Azure DevOps agent, so I cannot confirm the real hang path. Based on your hints, the next step is probably to ask the reporter whether stdin is actually closed (EOF) or just an open idle pipe, and whether there's any stack trace from a debug build. Happy to dig further once there's more info from the reporter.

escherstair:
@ssam18 
> the next step is probably to ask the reporter whether stdin is actually closed (EOF) or just an open idle pipe, and whether there's any stack trace from a debug build

If you can tell me how to collect the info you need I'll report here.
Thanks

ssam18:
> @ssam18
> 
> > the next step is probably to ask the reporter whether stdin is actually closed (EOF) or just an open idle pipe, and whether there's any stack trace from a debug build
> 
> If you can tell me how to collect the info you need I'll report here. Thanks

Here's a compact set of things to ask the reporter to run. The goal is to (a) confirm what stdin actually is in their environment, and (b) capture a goroutine dump showing where compose publish --yes is parked.

1. What is stdin in their environment?
Ask them to run the exact same invocation they're using, but first inspect fd 0:

# Show what stdin is wired to in the same shell context
ls -la /proc/self/fd/0
# or, for the actual compose invocation, prefix it:
ls -la /proc/self/fd/0 && docker compose publish --yes ...

What to look for:

a. /dev/null --> closed/EOF on read
b. pipe:[…] --> a pipe (could be open-idle if the writer hasn't closed it)
c. /dev/pts/N --> an interactive TTY
d. socket:[…] --> SSH/CI runner pseudo-stream

Also useful: what is the exact CI/runner (GitHub Actions step, GitLab, Jenkins, cron, ssh -T, nohup, systemd unit, docker exec without -i, etc.). The runner often dictates whether stdin is /dev/null (EOF) vs an idle pipe.

2. Is the read actually blocking on stdin?
While the hang is happening, in another shell:

PID=$(pgrep -nf "compose publish")
sudo cat /proc/$PID/stack          # kernel-side: should show pipe_wait / pipe_read if blocked on stdin
sudo strace -f -p $PID 2>&1 | head -50   # should show read(0, ...) parked
ls -la /proc/$PID/fd/0             # confirm what fd 0 is on the running process

If strace shows it parked in read(0, ...) and /proc/$PID/fd/0 is a pipe, that's the smoking gun for "open idle pipe, not EOF."

3. Stack trace from the Go binary
Compose is Go, so the cheapest "debug build" trick is just to send SIGQUIT — the Go runtime dumps all goroutine stacks to stderr and exits:
kill -QUIT $(pgrep -nf "compose publish")

Ask them to capture stderr (2> compose.stacks) and paste the dump. That will show the exact goroutine blocked on stdin and the call chain that led there — usually more useful than a gdb stack for Go.

If they want a non-fatal dump (keep the process alive), they can attach delve:
dlv attach $PID
(dlv) goroutines -with running   # or: goroutines -t

4. Minimal repro hint to include
Ask them to also confirm whether the hang reproduces with:

docker compose publish --yes ... < /dev/null
docker compose publish --yes ... < /dev/zero      # never EOFs; if this also hangs, your fix is the right shape

If < /dev/null does not hang but their real invocation does, then their stdin is an open-idle pipe and your --yes short-circuit is the correct fix; the contrast is good evidence to put in the PR.

That's the package: fd inspection + strace + SIGQUIT goroutine dump + the /dev/null vs /dev/zero contrast.

PR #13754, opened 2026-04-21
refactor to simplify handling json streams

thaJeztah:
**What I did**

**Related issue**

### DIFF
--- a/pkg/bridge/convert.go
+++ b/pkg/bridge/convert.go
@@ -178,7 +178,7 @@ func LoadAdditionalResources(ctx context.Context, dockerCLI command.Cli, project
 			if err != nil {
 				return nil, err
 			}
-			exposed.Add(strconv.Itoa(int(p.Num())))
+			exposed.Add(p.Port())
 		}
 		for _, port := range service.Ports {
 			exposed.Add(strconv.Itoa(int(port.Target)))
--- a/pkg/bridge/convert.go
+++ b/pkg/bridge/convert.go
@@ -232,9 +232,7 @@ func inspectWithPull(ctx context.Context, dockerCli command.Cli, imageName strin
 		}
 		defer func() { _ = stream.Close() }()
 
-		out := dockerCli.Out()
-		err = jsonmessage.DisplayJSONMessagesStream(stream, out, out.FD(), out.IsTerminal(), nil)
-		if err != nil {
+		if err := jsonmessage.DisplayStream(stream, dockerCli.Out()); err != nil {
 			return image.InspectResponse{}, err
 		}
 		if inspect, err = dockerCli.Client().ImageInspect(ctx, imageName); err != nil {
--- a/pkg/compose/build_classic.go
+++ b/pkg/compose/build_classic.go
@@ -268,24 +268,21 @@ func (s *composeService) doBuildImage(ctx context.Context, project *types.Projec
 	defer response.Body.Close() //nolint:errcheck
 
 	imageID := ""
-	aux := func(msg jsonstream.Message) {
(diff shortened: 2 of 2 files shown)


pkg/compose/build_classic.go
```diff
@@ -268,24 +268,22 @@ func (s *composeService) doBuildImage(ctx context.Context, project *types.Projec
...
+	err = jsonmessage.DisplayStream(response.Body, buildBuff, jsonmessage.WithAuxCallback(func(msg jsonstream.Message) {
 		var result buildtypes.Result
 		if err := json.Unmarshal(*msg.Aux, &result); err != nil {
 			logrus.Errorf("Failed to parse aux message: %s", err)
 		} else {
 			imageID = result.ID
 		}
-	}
-
-	err = jsonmessage.DisplayJSONMessagesStream(response.Body, buildBuff, progBuff.FD(), true, aux)
```
thaJeztah, line 280: Not sure if this one works; the old code was unconditionally setting "terminal" to "true".

pkg/bridge/convert.go
```diff
@@ -178,7 +178,7 @@ func LoadAdditionalResources(ctx context.Context, dockerCLI command.Cli, project
 			if err != nil {
 				return nil, err
 			}
-			exposed.Add(strconv.Itoa(int(p.Num())))
+			exposed.Add(p.Port())
 		}
 		for _, port := range service.Ports {
 			exposed.Add(strconv.Itoa(int(port.Target)))
```
thaJeztah, line 184: Not introduced in my changes, but wondering if this code is correct, or if it should preserve the protocol (tcp, udp, ...) https://docs.docker.com/reference/compose-file/services/#expose

pkg/compose/build_classic.go
```diff
@@ -268,24 +268,22 @@ func (s *composeService) doBuildImage(ctx context.Context, project *types.Projec
...
 		var jerr *jsonstream.Error
 		if errors.As(err, &jerr) {
 			// If no error code is set, default to 1
-			if jerr.Code == 0 {
-				jerr.Code = 1
-			}
-			return "", cli.StatusError{Status: jerr.Message, StatusCode: jerr.Code}
+			//
+			// TODO(thaJeztah): DisplayStream should return a errdefs Error corresponding with the status-code.
+			errCode := min(1, jerr.Code)
```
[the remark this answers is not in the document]
thaJeztah: Doh! Thought I fixed this one, but it ended in the wrong branch; should be fixed now.

pkg/bridge/convert.go
```diff
@@ -232,9 +232,7 @@ func inspectWithPull(ctx context.Context, dockerCli command.Cli, imageName strin
 		}
 		defer func() { _ = stream.Close() }()
 
-		out := dockerCli.Out()
-		err = jsonmessage.DisplayJSONMessagesStream(stream, out, out.FD(), out.IsTerminal(), nil)
-		if err != nil {
+		if err := jsonmessage.DisplayStream(stream, dockerCli.Out()); err != nil {
```
glours, line 235: I wonder if we won't loose the TTY detection here, the underling `term.GetFdInfo(out)` does a hard type assertion on `in.(*os.File)` and `dockerCli.out()` returns a `*cli/streams.Out` which won't pass the previous assertion, WDYT @thaJeztah ?

thaJeztah: OH! Hm... you may be right; I recall looking at that part, and maybe it was why I still had it in draft. There were other inconsistencies in moby/term related to that as well; e.g. this code being inconsistent in how it detected things (on windows); https://github.com/moby/term/blob/4a719361cea2e649e13aedf0a51965f90b7ef9f5/windows/console.go#L12-L31

PR #13760, opened 2026-04-28
fix(publish): flag literal inline environment values

glours:
**What I did**
PR #13402 removed the inline-env check entirely, so literal hardcoded values like `MYSQL_ROOT_PASSWORD: toto` now silently leak into the OCI artifact. Restore the check, scoped to literal values only, `${VAR}` and `$VAR` remain symbolic in the published YAML and still pass without `--with-env`, preserving the legitimate fix from #13402.

The walk follows the same files processFile actually serializes (top-level compose files plus local extends parents), so inherited literals and unrelated services in extends parents are covered too. Errors aggregate per service. With `--with-env` the walk is skipped entirely since the user has opted in.

**Related issue**

image

### DIFF
--- a/pkg/compose/publish.go
+++ b/pkg/compose/publish.go
@@ -25,8 +25,10 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"slices"
 	"strings"
 
+	"github.com/DefangLabs/secret-detector/pkg/detectors/keyword"
 	"github.com/DefangLabs/secret-detector/pkg/scanner"
 	"github.com/DefangLabs/secret-detector/pkg/secrets"
 	"github.com/compose-spec/compose-go/v2/loader"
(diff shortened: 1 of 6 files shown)


ndeloof:
I wonder: doing so we only warn on literal environment variable definition for services, not relying on interpolation. While this will prevent leaking `AWS_API_KEY=abc123` (which we have to) this also will warn user about the many `DB_USER=xx`, `LOG_LEVEL=info` or comparable variables which are used for 99% usages. I wonder we could distinguish those to avoid many false positive to preserve a sane UX. As we run secret detection on the compose files, is this check really relevant ?

glours:
@ndeloof I addressed your comments 😉

ndeloof (approved):

PR #13781, opened 2026-05-07
Add Compose-time variables: extension

WarrDoge:
## Summary

Adds a Compose-time `variables:` extension to the CLI plugin: YAML-native, scoped, include-aware variables that feed interpolation. Distinct from container runtime env (`environment:` / `env_file`).

- Top-level `variables:` (flat scalars) and `variables_file:` (list of YAML files)
- Include-local `variables:` and `variables_file:` per include entry
- Root-persistent CLI flags `--var KEY=VALUE` and `--var-file PATH` (both repeatable)
- Cross-variable references via `${OTHER}` between declarations, with cycle detection
- Precedence: shell env > `--var` > `--var-file` > include-local inline > include-local file > root inline > root file > included-file own
- `compose config --variables` extended with `RESOLVED VALUE` and `SOURCE` columns; existing JSON/YAML output schema preserved via `omitempty` fields

Implemented entirely in this CLI plugin. A preprocessor strips the extension keys, performs interpolation against the resolved scope, and hands rendered files to compose-go with `SkipInterpolation` set. No upstream `compose-go` change is required.

A design doc (`SPEC.md`) is available in the author's fork (intentionally not included in this commit) and can be added to the PR if maintainers prefer it co-located.

> Note: per CONTRIBUTING.md an enhancement issue is normally filed first to announce intent. I'm opening this PR directly to make the proposal concrete; happy to file a tracking issue if maintainers prefer.

## Test plan

- [x] `go test ./pkg/variables/...` — unit tests for scope precedence, cross-refs, cycles, coercion, no-leakage
- [x] `go test -tags e2e ./pkg/e2e/ -run TestComposeVariables` — 11/11 e2e tests pass against local Docker
- [x] `go test -tags e2e ./pkg/e2e/ -run 'TestLocalComposeConfig|TestSecretFromInclude|TestEnvPriority|TestEnvFile'` — adjacent existing e2e still pass (no regression)
- [x] `make build` — binary builds; smoke-tested with `compose config`, `--variables`, `--no-interpolate`, `--var`, `--var-file`, cross-ref, include-local fixtures

### DIFF
--- a/cmd/compose/compose.go
+++ b/cmd/compose/compose.go
@@ -51,6 +51,7 @@ import (
 	"github.com/docker/compose/v5/pkg/compose"
 	"github.com/docker/compose/v5/pkg/remote"
 	"github.com/docker/compose/v5/pkg/utils"
+	"github.com/docker/compose/v5/pkg/variables"
 )
 
 const (
@@ -144,6 +145,8 @@ type ProjectOptions struct {
 	WorkDir            string
 	ProjectDir         string
 	EnvFiles           []string
+	Vars               []string
+	VarFiles           []string
 	Compatibility      bool
 	Progress           string
 	Offline            bool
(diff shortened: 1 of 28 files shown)


glours:
@WarrDoge  thanks for the proposal, but I don't think this is something we'd want to merge.

 Process: As you note yourself, enhancement proposals should start with an issue. A feature this large, new top-level Compose keys, new CLI flags, a custom preprocessor, needs upstream design discussion before implementation, and ideally alignment with the compose-spec working group.

The problem is already solved. Looking at the examples in the PR, every use case is covered by existing mechanisms:

  - variables: APP_VERSION: "1.4.2" → just put APP_VERSION=1.4.2 in your .env file. ${VAR} interpolation is already Compose-time, not container runtime — a .env var only ends up in the
  container if you also declare it under environment:.
  - Scoped values per include → include: already supports env_file: for exactly this.
  - Node reuse within a single file → YAML anchors.
  - Layered overrides → multiple -f files.

 The one thing neither handles elegantly is cross-variable references (IMAGE: "${BASE}/api"), but that's a shell/Makefile one-liner, not a gap that warrants a new YAML extension.

 Architecture concern: The implementation preprocesses files into a tempdir and sets SkipInterpolation = true globally on compose-go's loader. This silently disables `compose-go`'s own interpolation for all variables in the project the moment any variables: key appears, a breaking behavioral change for existing configs.

I'd suggest closing this and opening a lightweight issue, in the [compose-spec](https://github.com/compose-spec/compose-spec) repo, if there's a specific gap you've hit that the existing mechanisms genuinely don't cover.

PR #13784, opened 2026-05-08
fix: do not add service name as alias on external networks

### LINKED ISSUE #8223: Add option to remove service name as default alias on networks
(Copied from https://forums.docker.com/t/add-option-to-remove-service-name-as-default-alias-on-networks/106172 after I realized this was a compose-specific request)

I am facing an issue where two separate docker compose projects sharing a common externally defined docker network would encounter DNS name collisions when both projects have a service with the same name.

**Example Setup:**

project1/docker-compose.yml:
```yaml
version: '3'

networks:
  app_internal:
    driver: bridge
  shared:
    external:
      name: shared_network

services:
  database:
    ...
    networks:
      app_internal:
      shared:
        aliases:
          - project1-database
```

project2/docker-compose.yml:
```yaml
version: '3'

networks:
  app_internal:
    driver: bridge
  shared:
    external:
      name: shared_network

services:
  database:
    ...
    networks:
      app_internal:
      shared:
        aliases:
          - project2-database
```
**Problem:**

`shared_network` is an externally created docker network common to both `project1` and `project2`. When I spin both projects up, both the `project1_database_1` and `project2_database_1` containers are aliased to `database` on `shared_network`, in addition to the custom aliases provided. 

Making things worse, when I try to communicate with `database` on other containers in `project1` or `project2` who are also connected to `shared_network`, it's a tossup whether I'll be talking with `project1_database_1` or `project2_database_1` .

**Desired Outcome for Example:**

* The only aliases `project1_database_1` and `project2_database_1` should be known by on `shared_network` are the custom aliases provided.
* From within `project1`:
    * The DNS for `database` should always point to `project1_database_1` over the `project1_app_internal` network
    * The DNS for `project1-database` should always point to `project1_database_1` over the `shared_network` network
    * The DNS for `project2-database`  should always point to `project2_database_1` over the `shared_network` network
* Ditto for `project2`

**Feature Request:**

Currently, the problematic behavior occurs because of this interaction in the Docker-compose documentation:

`Other containers on the same network can use *either* the service name or this alias to connect to one of the service’s containers.` *(emphasis mine)*

I would like to have an option at either a service level or a network level to tell Docker Compose *not* to add the service name as an alias when attaching a network.

stasadev:
## Summary

- Fixes #8223
- Stop adding the service name as an alias on external networks (`pkg/compose/create.go`)
- On `up`, log a `WARN` when this alias is skipped, so users notice the change and learn about the workaround
- Add unit tests and an e2e test

### What changed

Before this fix, `getAliases()` always added the service name as an alias when `useNetworkAliases` was true. This put internal service names onto external networks (networks Compose does not manage). On a shared external network, two projects with a service named `db` would resolve to each other - which is the bug.

The fix skips the service name on external networks. Aliases that the user sets under `services.<name>.networks.<net>.aliases` still go through.

### Discoverability

Users who relied on cross-project resolution over a shared external network would lose it silently. To make the change visible, `up` prints one warning per external network used by the project:

```
WARN[0000] service names [db, web] not registered as aliases on external network "shared-net"; add them under each service's networks.shared-net.aliases if needed
```

One warning per network keeps the output short, even on stacks with many services on the same external network. A service that already lists its own name under `networks.<net>.aliases` is left out of the warning, so the warning disappears once the user adopts the workaround.

Only `up` prints the warning. `create` and `run --use-aliases` do not:

- `create` is rarely used alone - it almost always runs as part of `up`
- `run --use-aliases` creates one-off containers, where the warning would just be noise

### Workaround

Add the alias per service, under that service's network entry. The alias namespace lives on the service, not on the network itself:

```yaml
networks:
  shared-net:
    external: true

services:
  web:
    networks:
      shared-net:
        aliases: [web]
  db:
    networks:
      shared-net:
        aliases: [db]
```

### Related

https://github.com/ddev/ddev/issues/8390 - DDEV projects share one external network (`ddev_default`). A `db` in one project resolved to another project's `db` over that shared network, because the service name was registered as an alias on it.

## Before / After

<details>
<summary>Setup</summary>

Two independent Compose projects and a standalone router share one external network. Project1 has `web` and `db`. Project2 has only `web` - no `db` of its own, so any `db` resolution from it would cross into project1. The router is a separate project on `shared-net` only, simulating a reverse proxy like ddev-router.

```sh
docker network create shared-net

cat <<'EOF' > project1.yaml
networks:
  default: {}
  shared-net:
    external: true

services:
  web:
    image: busybox
    command: sleep infinity
    networks:
      - default
      - shared-net
  db:
    image: busybox
    command: sleep infinity
    networks:
      - default
      - shared-net
EOF

cat <<'EOF' > project2.yaml
networks:
  default: {}
  shared-net:
    external: true

services:
  web:
    image: busybox
    command: sleep infinity
    networks:
      - default
      - shared-net
EOF

cat <<'EOF' > router.yaml
networks:
  shared-net:
    external: true

services:
  router:
    image: busybox
    command: sleep infinity
    networks:
      - shared-net
EOF

docker-compose -p project1 -f project1.yaml up -d
docker-compose -p project2 -f project2.yaml up -d
docker-compose -p router -f router.yaml up -d
```

Cleanup:

```bash
docker-compose -p project1 -f project1.yaml down
docker-compose -p project2 -f project2.yaml down
docker-compose -p router -f router.yaml down
docker network rm shared-net
```

</details>

<details>
<summary>Before (buggy behavior)</summary>

Service names appear as aliases on the external network:

```
$ docker inspect $(docker-compose -p project1 -f project1.yaml ps -q web) \
    --format '{{json .NetworkSettings.Networks}}' | jq 'to_entries[] | {network: .key, aliases: .value.Aliases}'
{"network": "project1_default", "aliases": ["project1-web-1", "web"]}
{"network": "shared-net",       "aliases": ["project1-web-1", "web"]}
#                                                               ^^^
#                                          service name leaks onto external network
```

Within project1, `web` can ping its own `db` by service name (expected):

```
$ docker-compose -p project1 -f project1.yaml exec web ping -c1 db
PING db (172.18.0.3): 56 data bytes
64 bytes from 172.18.0.3: seq=0 ttl=64 time=0.121 ms
```

Project2's `web` has no `db` of its own, but can reach project1's `db` via `shared-net` (bug):

```
$ docker-compose -p project2 -f project2.yaml exec web ping -c1 db
PING db (172.20.0.3): 56 data bytes
64 bytes from 172.20.0.3: seq=0 ttl=64 time=0.112 ms
```

`router` in project2 can also resolve project1's services via `shared-net` (bug):

```
$ docker-compose -p router -f router.yaml exec router ping -c1 web
PING web (172.20.0.2): 56 data bytes
64 bytes from 172.20.0.2: seq=0 ttl=64 time=0.098 ms

$ docker-compose -p router -f router.yaml exec router ping -c1 db
PING db (172.20.0.3): 56 data bytes
64 bytes from 172.20.0.3: seq=0 ttl=64 time=0.104 ms
```

</details>

<details>
<summary>After (fixed behavior)</summary>

Service names are no longer registered as aliases on the external network:

```
$ docker inspect $(docker-compose -p project1 -f project1.yaml ps -q web) \
    --format '{{json .NetworkSettings.Networks}}' | jq 'to_entries[] | {network: .key, aliases: .value.Aliases}'
{"network": "project1_default", "aliases": ["project1-web-1", "web"]}
{"network": "shared-net",       "aliases": ["project1-web-1"]}
#                                                        no service name
```

Within project1, `web` can still ping its own `db` by service name via the internal `default` network (unchanged):

```
$ docker-compose -p project1 -f project1.yaml exec web ping -c1 db
PING db (172.18.0.3): 56 data bytes
64 bytes from 172.18.0.3: seq=0 ttl=64 time=0.121 ms
```

Project2's `web` can no longer resolve project1's `db` (fixed):

```
$ docker-compose -p project2 -f project2.yaml exec web ping -c1 db
ping: bad address 'db'
```

`router` can no longer resolve services from project1 either (fixed):

```
$ docker-compose -p router -f router.yaml exec router ping -c1 web
ping: bad address 'web'

$ docker-compose -p router -f router.yaml exec router ping -c1 db
ping: bad address 'db'
```

Containers are still reachable on `shared-net` by their fully qualified container name, which remains as an alias on the external network:

```
$ docker-compose -p router -f router.yaml exec router ping -c1 project1-web-1
PING project1-web-1 (172.20.0.2): 56 data bytes
64 bytes from 172.20.0.2: seq=0 ttl=64 time=0.094 ms

$ docker-compose -p router -f router.yaml exec router ping -c1 project2-web-1
PING project2-web-1 (172.20.0.4): 56 data bytes
64 bytes from 172.20.0.4: seq=0 ttl=64 time=0.091 ms
```

</details>

## Test plan

- [x] `TestCreateEndpointSettings` - non-external network case still produces `containerName`, `serviceName`, and configured aliases
- [x] `TestCreateEndpointSettings_ExternalNetwork` - external network case produces only `containerName` and configured aliases (no service name)
- [x] `TestWarnExternalNetworkAliases` - covers the warning helper:
  - internal-only networks emit no warning
  - external network emits one warning per network, with services listed in sorted order
  - services that already list their own name under `networks.<net>.aliases` are excluded
  - if all services are explicitly self-aliased, no warning is emitted
  - multiple external networks each get their own warning
- [x] `TestExternalNetworkAliases` (e2e) - mixed internal + external network fixture, where `web` has no explicit alias on the external network and `db` declares `networks.external-net.aliases: [db]`:
  - the warning is logged on `up`
  - the warning lists `web` and excludes `db` (since `db` is self-aliased)
  - `web`'s service name is in `Aliases` on the internal network
  - `web`'s service name is not in `Aliases` on the external network
  - `db`'s explicit alias `[db]` is registered on the external network (workaround works end-to-end)
  - service-name DNS resolution still works between services over the internal network (`web` -> `db`)
- [x] `golangci-lint run --build-tags "e2e" ./pkg/compose/... ./pkg/e2e/...` - 0 issues

### DIFF
--- a/pkg/compose/create.go
+++ b/pkg/compose/create.go
@@ -368,10 +368,13 @@ func (s *composeService) prepareContainerMACAddress(service types.ServiceConfig,
 	return nil
 }
 
-func getAliases(project *types.Project, service types.ServiceConfig, serviceIndex int, cfg *types.ServiceNetworkConfig, useNetworkAliases bool) []string {
+func getAliases(project *types.Project, service types.ServiceConfig, serviceIndex int, networkKey string, cfg *types.ServiceNetworkConfig, useNetworkAliases bool) []string {
 	aliases := []string{getContainerName(project.Name, service, serviceIndex)}
 	if useNetworkAliases {
-		aliases = append(aliases, service.Name)
+		// service name is not registered as an alias on external networks; see warnExternalNetworkAliases
+		if n := project.Networks[networkKey]; !n.External {
+			aliases = append(aliases, service.Name)
+		}
 		if cfg != nil {
 			aliases = append(aliases, cfg.Aliases...)
 		}
@@ -445,7 +448,7 @@ func createEndpointSettings(p *types.Project, service types.ServiceConfig, servi
 	}
 
 	return &network.EndpointSettings{
-		Aliases:     getAliases(p, service, serviceIndex, config, useNetworkAliases),
+		Aliases:     getAliases(p, service, serviceIndex, networkKey, config, useNetworkAliases),
 		Links:       links,
 		IPAddress:   ipv4Address,
 		IPv6Gateway: ipv6Address,
--- a/pkg/compose/create_test.go
+++ b/pkg/compose/create_test.go
@@ -365,6 +365,26 @@ func TestCreateEndpointSettings(t *testing.T) {
 	}, cmpopts.EquateComparable(netip.Addr{})))
 }
 
+func TestCreateEndpointSettings_ExternalNetwork(t *testing.T) {
+	eps, err := createEndpointSettings(&composetypes.Project{
+		Name: "projName",
+		Networks: composetypes.Networks{
+			"netName": {External: true},
+		},
+	}, composetypes.ServiceConfig{
+		Name:          "serviceName",
+		ContainerName: "containerName",
+		Networks: map[string]*composetypes.ServiceNetworkConfig{
+			"netName": {
+				Aliases: []string{"alias1"},
+			},
+		},
+	}, 0, "netName", nil, true)
+	assert.NilError(t, err)
+	assert.Check(t, cmp.DeepEqual(eps.Aliases, []string{"containerName", "alias1"}),
+		"service name must not be added as alias on external networks")
+}
+
 func Test_buildContainerVolumes(t *testing.T) {
 	pwd, err := os.Getwd()
 	assert.NilError(t, err)
--- a/pkg/compose/up.go
+++ b/pkg/compose/up.go
@@ -23,6 +23,7 @@ import (
 	"os"
 	"os/signal"
 	"slices"
+	"strings"
 	"sync"
 	"sync/atomic"
 	"syscall"
@@ -42,6 +43,7 @@ import (
 )
 
 func (s *composeService) Up(ctx context.Context, project *types.Project, options api.UpOptions) error { //nolint:gocyclo
+	warnExternalNetworkAliases(project)
 	err := Run(ctx, tracing.SpanWrapFunc("project/up", tracing.ProjectOptions(ctx, project), func(ctx context.Context) error {
 		err := s.create(ctx, project, options.Create)
 		if err != nil {
@@ -313,4 +315,36 @@ func shouldFollowStartEvent(event api.ContainerEvent, attached []string, attachT
 		return false
 	}
 	return true
+}
+
+// warnExternalNetworkAliases emits one warning per external network the project is connected to, listing the services
+// whose service-name alias is intentionally not registered (see getAliases). Services that already include their own
+// name under the network's explicit aliases are excluded, so the warning disappears once the user adopts the
+// workaround. Must be called before Run() to avoid interleaving with the TUI progress display.
+func warnExternalNetworkAliases(project *types.Project) {
+	byNetwork := map[string][]string{}
+	for _, service := range project.Services {
+		for networkKey, cfg := range service.Networks {
+			n := project.Networks[networkKey]
+			if !n.External {
+				continue
+			}
+			if cfg != nil && slices.Contains(cfg.Aliases, service.Name) {
+				continue
+			}
+			byNetwork[networkKey] = append(byNetwork[networkKey], service.Name)
+		}
+	}
+	networks := make([]string, 0, len(byNetwork))
+	for k := range byNetwork {
+		networks = append(networks, k)
+	}
+	slices.Sort(networks)
+	for _, networkKey := range networks {
+		services := byNetwork[networkKey]
+		slices.Sort(services)
+		logrus.Warnf("service names [%s] not registered as aliases on external "+
+			"network %q; add them under each service's networks.%s.aliases if needed",
+			strings.Join(services, ", "), networkKey, networkKey)
+	}
 }
--- a/pkg/compose/up_test.go
+++ b/pkg/compose/up_test.go
@@ -17,8 +17,12 @@
 package compose
 
 import (
+	"bytes"
+	"strings"
 	"testing"
 
+	composetypes "github.com/compose-spec/compose-go/v2/types"
+	"github.com/sirupsen/logrus"
 	"gotest.tools/v3/assert"
 
 	"github.com/docker/compose/v5/pkg/api"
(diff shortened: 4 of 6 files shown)


glours:
Thanks for the fix and for the detailed write-up. The collision scenario is a real pain point, and the proposed approach makes sense.

Before merging, though, I think this should be called out as a silent breaking change. Users who intentionally rely on service-name resolution across an external network will suddenly lose DNS resolution, with no indication from Compose as to why. While a workaround exists through explicit aliases, it is not discoverable.

Could you add a log notice when the service-name alias is skipped on an external network? Something along the lines of:
`service "web": not registering service name as an alias on external network "shared-net"; add it explicitly under networks.shared-net.aliases if needed`

A `logrus.Warnf` level message would feel appropriate here.

An end-to-end test covering a mixed internal + external network scenario would also be very helpful. The current unit test covers the single-network case, but a test asserting that the service name still resolves on the internal network while not being registered on the external one would provide a stronger regression guard.

@ndeloof what do you think?

stasadev:
> this should be called out as a silent breaking change.

Makes sense, I added a `logrus.Warnf` and the e2e test. PR description updated with the full details.

One note: the warning lives in a separate helper `warnExternalNetworkAliases(project)` called once at the top of `Up()`, not inside `getAliases()` - putting it there would print the message many times and mix it with the TUI progress output.

thaJeztah:
I wonder if we need to have a closer look at different scenarios.

This is a known problem for some use-cases, but disabling the alias may also be covering over other issues;

Networks in docker also form a security boundary; services connected to the same network are granted access to other services connected to the same network. Disabling DNS resolution won't prevent other services from connecting; in the example given (multiple projects defining a `db` service); that may mean that unrelated projects might be able to connect to a database service from another project.

If the shared network is used for some reverse proxy (traefik, caddy, nginx-proxy), that could even mean the database may become exposed publicly.

The reverse can also be true; perhaps a less common use-case, but I've seen users work with "composable" compose projects; a network that's set up, then multiple compose projects that define part of the stack, but attached to the common network; for those, DNS resolution between those may be intentional.

Wondering of some options should be configurable on the network, to more clearly define the purpose of the network, which could be an ingress network, disallowing communication between services attached to the  network.

TIRTAGT:
Honestly what I think this PR needs are to change the behavior from `on external networks` to `when opt-in via Docker's daemon.json`, that way this isn't a breaking change but is an opt-in change.

----

Personally, I would instantly opt-in as soon as this feature exist because it will prevent a LOT of misconfigurations like these:

Scenario:
> You have mailpit (a mail capture service) as `app1_mailpit` and `app2_mailpit` (swarm stacks)

Currently:
> Someone used `mailpit` to connect their app, it connected as usual. Testing shows the mail feature randomly stopped working (which it didn't, the mails are just going where they didn't expect).

With this PR:
> Someone used `mailpit` to connect their app, DNS resolve fails hard and misconfiguration is identified

Scenario source: 2 of our team members that have gone 3 hours deep into debugging "smtp feature isn't reliable"

stasadev:
I can't reopen it, there's no button.

stasadev:
@glours and @thaJeztah,

I'd be happy to work on this in whatever way you prefer.

I just need a bit of direction on how you'd like to proceed or what you'd like me to do next.

PR #13803, opened 2026-05-21
docs: compose logs update

### LINKED ISSUE #20712: Document what format `docker compose logs --since` takes
### Is this a docs issue?

- [X] My issue is about the documentation content or website

### Type of issue

I can't find what I'm looking for

### Description

The `docker compose logs` reference doc does not document what datetime formats I can pass to `--since` or `--until`. 

### Location

https://docs.docker.com/reference/cli/docker/compose/logs/

### Suggestion

Presumably valid formats are the same as `docker container logs`, but I'm left to guess. (If that's the case, a link to https://docs.docker.com/reference/cli/docker/container/logs/ would probably suffice - no need to duplicate all this text.)

Source file: https://github.com/docker/docs/blob/main/content/reference/cli/docker/compose/logs.md

aevesdocker:
**What I did**

Minor docs update. Closes https://github.com/docker/docs/issues/20712

**Related issue**


glours (approved):
LTGM 👍

cmd/compose/logs.go
```diff
@@ -62,8 +62,8 @@ func logsCommand(p *ProjectOptions, dockerCli command.Cli, backendOptions *Backe
 	flags := logsCmd.Flags()
 	flags.BoolVarP(&opts.follow, "follow", "f", false, "Follow log output")
 	flags.IntVar(&opts.index, "index", 0, "index of the container if service has multiple replicas")
-	flags.StringVar(&opts.since, "since", "", "Show logs since timestamp (e.g. 2013-01-02T13:23:37Z) or relative (e.g. 42m for 42 minutes)")
-	flags.StringVar(&opts.until, "until", "", "Show logs before a timestamp (e.g. 2013-01-02T13:23:37Z) or relative (e.g. 42m for 42 minutes)")
+	flags.StringVar(&opts.since, "since", "", "Show logs since timestamp (e.g. 2013-01-02T13:23:37Z) or relative (e.g. 42m for 42 minutes). Accepts the same formats as `docker container logs --since`.")
+	flags.StringVar(&opts.until, "until", "", "Show logs before a timestamp (e.g. 2013-01-02T13:23:37Z) or relative (e.g. 42m for 42 minutes). Accepts the same formats as `docker container logs --until`.")
```
thaJeztah, line 66: Do we want to add this all in the flag description? I think we could have some alternatives here; if we add an `Examples` header in the markdown, and a header that has the right anchor, the docs should create a link to the flag description (similar to https://github.com/docker/cli/blob/v29.5.2/docs/reference/commandline/container_logs.md)


An alternative is to set the `ExternalURL` annotation; that would allow the docs to link to an alternative URL, which can be pointing to the corresponding flag for `docker container logs`;
https://github.com/docker/cli-docs-tool/blob/7543983fe39fcedd9182dfbee8824375999f8add/annotation/annotation.go#L19-L20

For an example, see https://github.com/docker/cli/blob/ad776d1e107b0d1cb3559a9c6b6391e11c3c51de/cli/command/image/build.go#L118-L119

```go
	flags.Var(&options.buildArgs, "build-arg", "Set build-time variables")
	flags.SetAnnotation("build-arg", annotation.ExternalURL, []string{"https://docs.docker.com/reference/cli/docker/buildx/build/#build-arg"})
```

thaJeztah: Looks like we still have work to do on the CLI documentation for the accepted formats though https://docs.docker.com/reference/cli/docker/container/logs/#until 

(I recall there were some formats that worked, but not _intentional_, so we need to a tread carefully which formats we want to document).

thaJeztah:
Gave it a quick go with my suggestions; I'm also opening a PR in docker/cli to allow linking to the other flags (but this will require those changes to be published to docs.docker.com first);

- https://github.com/docker/compose/pull/13806

aevesdocker:
Thanks @thaJeztah, your PR makes sense. Will close this one

thaJeztah:
Thanks! Yeah, we should have a look if we can make the flag-descriptions more complete, but it's slightly difficult; add too much, and the output becomes unreadable; add too little, and the flag descriptions aren't very informative.

PR #13808, opened 2026-05-22
feat: add confirmation label when volume spec changes

### LINKED ISSUE #13807: Add optional auto-confirm label on volumes
### Description

In CI/CD workflows, changes to volumes may trigger interactive confirmation prompts, which can break automation and require manual intervention.

For example, when an existing volume does not match the Compose specification, Docker may prompt:

`Volume "... " exists but doesn't match the Compose specification in the Compose file. Recreate (data will be lost)?`

To support non-interactive environments, it would be useful to provide an option to automatically confirm volume recreation in such cases, while keeping the default interactive behavior unchanged.

This behavior is controlled via a per-volume label, allowing precise control when multiple volumes are defined in a single Compose file. The label does not apply globally.

### Example

```diff
services:
  app:
    image: alpine
    volumes:
      - my_vol:/my_vol
      - no_recreate_label:/no_recreate_label

volumes:
  my_vol:
    driver: local
    driver_opts:
      device: tmpfs
      o: size=1k,uid=65530,gid=0,mode=0700,noexec
      type: tmpfs
    labels:
+      com.docker.compose.volume.recreate-when-spec-updated: "true"
      foo: bar
  nfs:
    driver: local
    driver_opts:
      type: nfs
      o: addr=XXX,nfsvers=4.2,soft,timeo=600,retrans=2,rsize=1048576,wsize=1048576
      device: :/storage2/sonarr/media/
    labels:
+      com.docker.compose.volume.recreate-when-spec-updated: "true"
  no_recreate_label:
    labels:
      foo: bar

```

### use case 

1. Shared tmpfs mounts between services
2. Volume mounts that can be safely deleted without removing the underlying data (e.g., NFS volumes)

https://github.com/kimdre/doco-cd/issues/1130 https://github.com/kimdre/doco-cd/issues/1267

qianlongzt:
**What I did**

Add support for label: `com.docker.compose.volume.recreate-when-spec-updated`

```yaml
volumes:
  my_data:
    labels:
      com.docker.compose.volume.recreate-when-spec-updated: "true"
````

This allows automatic handling of volume recreation when specs change, improving compatibility with CD workflows by avoiding interactive prompts.

**Related issue**

closes: https://github.com/docker/compose/issues/13807

### DIFF
--- a/pkg/api/labels.go
+++ b/pkg/api/labels.go
@@ -33,6 +33,9 @@ const (
 	ContainerNumberLabel = "com.docker.compose.container-number"
 	// VolumeLabel allow to track resource related to a compose volume
 	VolumeLabel = "com.docker.compose.volume"
+	// VolumeRecreateWhenSpecUpdatedLabel, when set to true, causes the volume to be recreated when its specification is updated. Otherwise, the volume is not recreated.
+	// WARNING: Recreating a volume may result in data loss, depending on the volume driver.
+	VolumeRecreateWhenSpecUpdatedLabel = "com.docker.compose.volume.recreate-when-spec-updated"
 	// NetworkLabel allow to track resource related to a compose network
 	NetworkLabel = "com.docker.compose.network"
 	// WorkingDirLabel stores absolute path to compose project working directory
--- a/pkg/compose/create.go
+++ b/pkg/compose/create.go
@@ -43,6 +43,7 @@ import (
 	cdi "tags.cncf.io/container-device-interface/pkg/parser"
 
 	"github.com/docker/compose/v5/pkg/api"
+	"github.com/docker/compose/v5/pkg/utils"
 )
 
 type createOptions struct {
@@ -1623,8 +1624,8 @@ func (s *composeService) ensureVolume(ctx context.Context, name string, volume t
 	}
 	actual, ok := inspected.Volume.Labels[api.ConfigHashLabel]
 	if ok && actual != expected {
-		msg := fmt.Sprintf("Volume %q exists but doesn't match configuration in compose file. Recreate (data will be lost)?", volume.Name)
-		confirm, err := s.prompt(msg, false)
+		// use the volume label from the compose file.
+		confirm, err := confirmVolumeRecreate(volume.Labels, s.prompt, volume.Name)
 		if err != nil {
 			return "", err
 		}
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

**Comment by Quantum0uasar** [comment]
This is a useful feature for CD pipeline automation — the interactive volume recreation prompt is a common friction point in CI/CD environments. A few thoughts on the implementation:

## Label naming convention

`com.docker.compose.volume.recreate-when-spec-updated` follows the existing Docker label namespace convention well. However, consider whether a shorter, more ergonomic alias like `com.docker.compose.recreate=on-spec-change` would be easier to remember. Looking at the existing Compose label space, verbose names are the norm, so the proposed name is consistent.

## Scope: volumes only, or also other resources?

The same "recreate on spec change" problem applies to networks (e.g., network subnet changes) and potentially configs/secrets. While it's fine to scope this PR to volumes only, it would be worth noting in the PR description whether this label pattern is intended to be extensible to other resource types in the future, or if it's intentionally volume-specific.

## Safety consideration: data loss risk

Reproducing a named volume **destroys existing data**. The label effectively silences a prompt that exists specifically to prevent accidental data loss. The implementation should clearly document this in:
1. The label's inline comment in the code
2. The Compose spec documentation (reference docs for labels)
3. Potentially a warning logged to stderr even when the label is set (e.g., `WARN: Volume 'mydata' is being recreated due to spec change (label: com.docker.compose.volume.recreate-when-spec-updated)`) so operators have an audit trail

## CI workflow example

A concrete usage example in the PR description would help reviewers understand the target use case:
```yaml
volumes:
  my_data:
    labels:
      com.docker.compose.volume.recreate-when-spec-updated: "true"
```

This makes it immediately clear whether the label value matters (boolean string) or just the presence of the label key is sufficient.

Good feature for automation-first workflows. The main thing I'd want to see before merge is explicit data-loss documentation.

qianlongzt:
I checked the code and couldn't find any other resource types that currently support this kind of recreation behavior. Because of that, I'd prefer to keep the scope limited to volumes for now.

I'm also not a big fan of a value-based label such as `com.docker.compose.recreate=on-spec-change`.

At the moment, the behavior is essentially a yes/no decision, similar to how the `--yes` flag works. A boolean-style label seems simpler and easier to understand. If we introduce additional recreation policies in the future, we can revisit the label format then.

qianlongzt:
@Quantum0uasar I updated the code comments and added warning logs to show what will happen when the volume spec changes. But I couldn't find where to add the documentation. Should I add it to `docs/reference/compose_up.md`?

vaguul:
Thanks for adding the PR and tests. One edge case I would check before this moves forward: the current patch calls `confirmVolumeRecreate(inspected.Volume.Labels, ...)`, so the decision is based on the labels already stored on the existing Docker volume, not the labels from the Compose file being applied.

That means these two cases can behave unexpectedly:

- old file has no recreate label, new file adds it: Compose still prompts because the existing volume does not have the label yet
- old file has the label, new file removes it or sets it false: Compose may still auto-recreate because the existing volume still has the old label

A useful regression test would be `old.yml` without `com.docker.compose.volume.recreate-when-spec-updated`, then `new.yml` with that label set to true. If the intended policy is controlled by the desired Compose spec, the recreate decision probably needs to use the desired `volume.Labels` rather than `inspected.Volume.Labels`.

Small secondary point: if the label is present with an invalid or false-ish value, it currently skips the prompt and returns false. If the label is only meant to opt in to auto-confirm, false/invalid values may be safer as normal prompt behavior instead of silently changing the decision.

qianlongzt:
@vaguul I updated it to use a Docker Compose volume label and prompt the user for confirmation when the label is set to false.

vaguul:
Thanks for updating it. The current version addresses the edge case I was worried about: confirmVolumeRecreate(volume.Labels, ...) now uses the desired Compose spec labels, and the new e2e case covers the old-file-without-label -> new-file-with-label path.\n\nI also like that false/no-label now falls back to the existing prompt behavior instead of silently changing the decision. From my side, this looks much safer.

PR #13822, opened 2026-06-06
feat: resolve ref+ secret references in environment values

### LINKED ISSUE #13821: Support ref+<backend>:// secret resolution in environment values (vals/helmfile URI syntax)
### Description

When running services that need secrets (database passwords, API keys, etc.), there's no native way to resolve them from external secret stores like OpenBao, HashiCorp Vault, AWS Secrets Manager, or others. Users currently rely on external preprocessors:

```bash
vals eval -f compose.yaml | docker compose -f - up
```

or shell variable substitution, which adds friction and breaks Docker Compose UX (`docker compose ps`, `logs`, etc. with stdin-based flows).

## Describe the solution you'd like

Support the `ref+<backend>://path#/key` URI scheme in environment values, consistent with the convention already established by:

- [helmfile/vals](https://github.com/helmfile/vals) — the reference implementation of this URI scheme
- [ArgoCD Vault Plugin](https://argocd-vault-plugin.readthedocs.io/)
- [helmfile](https://github.com/helmfile/helmfile)

Example `compose.yaml`:

```yaml
services:
  postgres:
    image: postgres:16
    environment:
      POSTGRES_USER: "ref+openbao://secret/data/prod/postgres#/username"
      POSTGRES_PASSWORD: "ref+openbao://secret/data/prod/postgres#/password"
      POSTGRES_DB: "ref+openbao://secret/data/prod/postgres#/database"

  myapp:
    image: myapp:latest
    environment:
      DATABASE_URL: "ref+vault://secret/data/prod/postgres#/connection_string"
      API_KEY: "ref+awssecrets://prod/api-keys#/stripe"
    depends_on:
      - postgres
```

The URI format follows the vals convention:

```text
ref+<backend>://<path>#/<key>
```

| Component | Description |
|-----------|-------------|
| `ref+` | Prefix indicating a secret reference (distinguishes from regular values) |
| `<backend>` | Secret store backend (`openbao`, `vault`, `awssecrets`, `gcpsecrets`, `azurekeyvault`, etc.) |
| `<path>` | Path to the secret in the backend |
| `#/<key>` | JSON pointer to extract a specific field from the secret |

## Correlation with helmfile/vals

The [vals](https://github.com/helmfile/vals) project defines this URI scheme and supports 20+ backends. Docker Compose could either:

1. Use vals as a library (`github.com/helmfile/vals`) — immediate support for all backends with minimal code
2. Implement the same URI scheme natively — starting with a few common backends (OpenBao, Vault, AWS Secrets Manager) using a pluggable resolver interface, keeping compatibility with the vals syntax so users have a consistent experience across tools

Either way, adopting the same `ref+<backend>://` syntax means users who already use helmfile, ArgoCD Vault Plugin, or vals get a familiar, zero-learning-curve experience in Docker Compose.

## Proposed implementation

Resolution would happen at project load time (after interpolation, before container creation), in `pkg/compose/loader.go`:

- Values prefixed with `ref+` are detected and resolved against the appropriate backend
- Backend authentication uses standard environment variables (`OPENBAO_ADDR`/`OPENBAO_TOKEN`, `VAULT_ADDR`/`VAULT_TOKEN`, `AWS_REGION`, etc.)
- Non-`ref+` values are untouched (fully backward compatible)
- A resolver interface makes adding new backends straightforward

## Describe alternatives you've considered

- **vals eval as preprocessor**: Works but breaks normal Docker Compose workflows (can't use named projects, `docker compose down`, etc. properly with stdin)
- **Shell variable substitution with wrapper scripts**: Exposes secrets in process environment and shell history
- **Docker secrets / external secret operators**: Only works with Swarm or Kubernetes, not standalone Compose

## Additional context

- The `ref+` prefix makes this opt-in and backward compatible — existing compose files are unaffected
- OpenBao and HashiCorp Vault use nearly identical APIs (OpenBao is a fork), so supporting both is trivial
- This would make Docker Compose a first-class citizen in secret-store-backed workflows alongside Helm and ArgoCD

johan-nilsson-82:
## What I did

Add support for resolving `ref+openbao://` secret references in service environment values.

The implementation resolves references at project load time using the OpenBao API client.

Currently this supports OpenBao KV v2-style secret paths.

## Related issue

Closes docker/compose#13821


glours:
Hi @johan-nilsson-82
Thanks for this PR. We've been looking at adding external secret store support to Compose and your contribution is actually good evidence that this is a real need — we're citing it internally.

The direction we're taking is native support for [Docker Secret Engine](https://github.com/docker/secrets-engine) (se://), a backend-agnostic daemon that proxies to multiple vaults (1Password today, Vault/OpenBao/Bitwarden on the roadmap). That way Compose only ever needs one integration, rather than a separate SDK dependency for each backend.

Given that, we're unlikely to merge this approach, but this isn't a no to the goal, just a different path to it. The design is in progress and your input would be welcome.

johan-nilsson-82:
Thanks, that direction makes sense.

One clarification from my side: in #13821 I mentioned [vals](https://github.com/helmfile/vals) as one possible implementation option, but I intentionally did not use it in this PR.

The main reasons were:

* I did not want to force a Go/toolchain version bump in Compose.
* Pulling in `vals` directly also brings a large dependency graph, because it supports many backends.
* That seemed too heavy for Compose itself, especially for an initial proof of concept.

So the PR took the smaller native route: implement only the minimal OpenBao/Vault-style resolver path directly, while keeping the user-facing syntax compatible with the `vals` / Helmfile convention.

Given the Docker Secret Engine direction, I agree that the backend integrations probably belong outside Compose. In that model, `vals` may still be useful as prior art, and possibly even as a good starting point for early compatibility work, but probably not as a direct Compose dependency.

One thing I think is worth considering in the `se://` design is how backend-specific paths and mixed backends would be represented.

For example, with the vals-style syntax the backend and path semantics are explicit:

```yaml
environment:
  POSTGRES_PASSWORD: "ref+openbao://secret/data/prod/postgres#/password"
  API_KEY: "ref+awssecrets://prod/api-keys#/stripe"
```

Different secret stores do not always have the same path model, field selector model, or authentication context. Some projects may also need to mix secret stores, for example OpenBao for internal app secrets and AWS Secrets Manager for cloud-provider-managed secrets.

A generic `se://...` reference may be cleaner from the Compose side, but it also seems like it would hide those backend details behind separate Docker Secret Engine configuration. That may be the right abstraction, but it would be useful to understand how users will:

* choose which backend a given `se://` reference resolves against
* represent backend-specific path conventions
* mix multiple vaults/backends in the same Compose project
* migrate from existing `ref+<backend>://...` references

So I think the important question is not whether Compose should depend on `vals` directly, but whether Docker Secret Engine could learn from the vals URI model or provide a compatibility/import/migration path for users already using Helmfile/vals-style references.

glours:
@johan-nilsson-82 thanks for your feedback, I'll close the PR and keep the issue open

PR #13825, opened 2026-06-07
fix(publish): bypass Docker Desktop proxy for loopback registries

### LINKED ISSUE #13824: docker compose publish ignores insecure/localhost registry and routes through proxy (proxyconnect ./pipe/dockerHttpProxy)
### Description

`docker compose publish` to a **localhost** (insecure, plain-HTTP) registry fails on Docker Desktop because the publish client routes the registry request through Docker Desktop's transparent HTTP proxy (named pipe `dockerHttpProxy`) instead of connecting directly.

`docker push` / `docker pull` to the same registry work fine — the registry is correctly recognized as insecure (`docker info` lists `127.0.0.0/8` under *Insecure Registries*). Only `docker compose publish` is affected.

**Current behavior** — with `--insecure-registry`, publish switches to `http://` but then fails:

```
failed to do request: Head "http://localhost:5000/v2/spike-app/blobs/sha256:...":
proxyconnect tcp: open ./pipe/dockerHttpProxy: The system cannot find the path specified.
```

Two things are wrong:

1. **The proxy is not bypassed for localhost.** The daemon proxy is `http.docker.internal:3128` with `No Proxy: hubproxy.docker.internal` — localhost/`127.0.0.1` are not excluded, and `compose publish` applies the proxy to the localhost target. It does **not** honor `NO_PROXY`/`HTTP_PROXY` env vars, nor Docker Desktop's proxy-exclude list (verified below).
2. **The proxy named-pipe path is malformed.** The error shows a relative `./pipe/dockerHttpProxy`, not the real pipe `\\.\pipe\dockerHTTPProxy`, so the `npipe://` proxy URL appears to be mis-parsed and dialed as a file path.

**Expected behavior** — `docker compose publish` connects directly to the localhost/insecure registry (as `docker push` does) and publishes the compose artifact, without going through the proxy.

## Steps To Reproduce

Both shells below were run on the **same Windows 11 + Docker Desktop** host and produce identical results — so this is not shell- or WSL-specific. `docker push` succeeds; `docker compose publish` fails either way. (I have not tested on a native Linux/macOS host.)

- **Git Bash** — MSYS2 / MINGW64, GNU bash 5.2.26
- **PowerShell** — Windows PowerShell 5.1

In both, `docker` is the Windows Docker Desktop CLI (`C:\Program Files\Docker\Docker\resources\bin\docker`).

**Git Bash (MSYS2 / MINGW64):**

```bash
# 1. Plain-HTTP registry on localhost
docker run -d -p 5000:5000 --name registry registry:3

# 2. Seed an image (proves registry + insecure-HTTP path are fine)
docker pull alpine:3.20
docker tag alpine:3.20 localhost:5000/spike-alpine:1
docker push localhost:5000/spike-alpine:1          # ✅ succeeds

# 3. Minimal compose file referencing that image
cat > compose.yaml <<'EOF'
name: spike
services:
  app:
    image: localhost:5000/spike-alpine:1
    command: ["sleep", "3600"]
EOF

# 4. Publish the compose application to the same registry
docker compose -f compose.yaml publish localhost:5000/spike-app:1 --insecure-registry -y
# ❌ proxyconnect tcp: open ./pipe/dockerHttpProxy: The system cannot find the path specified.
```

**PowerShell (Windows / Docker Desktop):**

```powershell
# 1. Plain-HTTP registry on localhost
docker run -d -p 5000:5000 --name registry registry:3

# 2. Seed an image (proves registry + insecure-HTTP path are fine)
docker pull alpine:3.20
docker tag alpine:3.20 localhost:5000/spike-alpine:1
docker push localhost:5000/spike-alpine:1          # ✅ succeeds

# 3. Minimal compose file referencing that image
@'
name: spike
services:
  app:
    image: localhost:5000/spike-alpine:1
    command: ["sleep", "3600"]
'@ | Set-Content -Encoding utf8 compose.yaml

# 4. Publish the compose application to the same registry
docker compose -f compose.yaml publish localhost:5000/spike-app:1 --insecure-registry -y
# ❌ proxyconnect tcp: open ./pipe/dockerHttpProxy: The system cannot find the path specified.
```

What I observed across variants:

| Command | Result |
|---|---|
| `docker push localhost:5000/...` | ✅ succeeds (direct, insecure-HTTP auto-trusted) |
| `docker compose publish ...` (no flag) | ❌ tries `https://localhost:5000` (publish defaults to HTTPS) |
| `docker compose publish ... --insecure-registry` | ❌ switches to `http://`, then `proxyconnect ... open ./pipe/dockerHttpProxy` |
| same + `HTTP_PROXY=` / `NO_PROXY=localhost,127.0.0.1` | ❌ identical error — env ignored |
| same + bogus `HTTP_PROXY=http://127.0.0.1:9` | ❌ still uses `./pipe/dockerHttpProxy` — confirms env not consulted |
| after Docker Desktop → Proxies → exclude `localhost,127.0.0.1` + restart | ❌ unchanged; `docker info` still shows `No Proxy: hubproxy.docker.internal` |

### Compose Version

```Text
$ docker compose version
Docker Compose version v5.1.4

$ docker-compose version
Docker Compose version v5.1.4
```

### Docker Environment

```Text
Client:
 Version:    29.5.2
 Context:    desktop-linux
 Plugins:
  buildx: Docker Buildx (Docker Inc.)  Version: v0.34.0-desktop.1
  compose: Docker Compose (Docker Inc.)  Version: v5.1.4

Server:
 Server Version: 29.5.2
 Storage Driver: overlayfs (io.containerd.snapshotter.v1)
 Cgroup Version: 2
 Kernel Version: 6.6.114.1-microsoft-standard-WSL2
 OSType: linux
 Architecture: x86_64
 Name: docker-desktop
 Labels:
  com.docker.desktop.address=npipe://\\.\pipe\docker_cli
 Insecure Registries:
  hubproxy.docker.internal:5555
  ::1/128
  127.0.0.0/8
 Live Restore Enabled: false


Host: Windows 11. Docker Desktop with WSL2 backend.
```

### Anything else?

- The named pipe actually exists as `\\.\pipe\dockerHTTPProxy`, but the dialer reports the relative path `./pipe/dockerHttpProxy` — so this looks like an `npipe://` proxy-URL parsing/dialing bug in the publish transport, in addition to the missing localhost bypass.
- Related: #13352 (insecure-registry support for `publish`) added `--insecure-registry` / `__TEST__INSECURE__REGISTRY__`. That part works (it switches the scheme to HTTP); this issue is the remaining **proxy bypass / named-pipe** behavior on top of it.
- No env-var or flag workaround found. `docker push` is a working alternative; only `docker compose publish` hits the proxy.

Please let me know if this is a https://github.com/docker/for-win issue and should be reported there, or elsewhere altogether, instead.

ptrdom:
## What I did

`docker compose publish` routed **all** registry traffic through Docker Desktop's HTTP proxy. Publishing to a registry on `localhost` therefore failed on Windows with:

```
proxyconnect tcp: open ./pipe/dockerHttpProxy: The system cannot find the path specified.
```

even though `docker push`/`docker pull` worked against the same registry.

Two bugs in `internal/desktop/proxy.go`:

1. **No loopback bypass.** `ProxyTransport` forced every request through the DD proxy and its `DialContext` always dialed the proxy socket, so loopback targets could never connect directly. Proxy selection now bypasses the proxy **only** for loopback targets (`localhost`, `127.0.0.0/8`, `::1`); all other registry traffic stays routed through Docker Desktop's PAC-aware proxy, so Desktop keeps ownership of proxy decisions (e.g. enterprise-managed proxies). The local process `NO_PROXY`/`no_proxy` is deliberately **not** honored, so a broad value such as `*` or `.corp` cannot bypass centrally managed proxy policy. `DialContext` routes only the sentinel proxy address to the DD socket and dials loopback targets directly.
2. **Malformed Windows pipe path.** The proxy named-pipe endpoint was hardcoded as `npipe://./pipe/...`, yielding the relative path `./pipe/dockerHttpProxy`. It is now derived from the engine endpoint, preserving its namespace. Docker Desktop reports the endpoint in the backslash form `npipe://\\.\pipe\docker_cli`, so the derivation uses `LastIndexAny` to handle both backslash and forward-slash forms.

Result: publishing to `localhost:5000` connects directly like `docker push`, while every non-loopback registry still goes through the Docker Desktop proxy (now reachable on Windows).

## Related issue

Fixes #13824

## How I tested

Validated end-to-end on Docker Desktop 29.5 / Windows 11 (the issue's environment) by building binaries from both the parent commit and this branch and running the issue's reproduction steps:

```console
# start a plain-HTTP localhost registry
docker run -d -p 5000:5000 --name registry registry:3

# minimal compose.yaml referencing any pushable image, then publish
# (spike-app is the placeholder image name from #13824's repro steps;
#  --insecure-registry tells the OCI resolver to use plain HTTP)
docker compose publish localhost:5000/spike-app:1 --insecure-registry -y
```

- **Parent commit:** reproduces the issue verbatim — `proxyconnect tcp: open ./pipe/dockerHttpProxy: The system cannot find the path specified.` (exit 1).
- **This branch:** publishes successfully (exit 0); the artifact lands in the localhost registry.

Automated coverage in `internal/desktop/proxy_test.go`:
- `TestDDProxyFunc_BypassesLoopbackOnly` — loopback names/IPs bypass the proxy; all other hosts (including ones a local `NO_PROXY` would match) stay routed through Docker Desktop.
- `TestHTTPProxySocketEndpoint_WindowsNamedPipe` — proxy pipe path keeps the engine endpoint's namespace for both backslash and forward-slash forms.

---

### DIFF
--- a/internal/desktop/proxy.go
+++ b/internal/desktop/proxy.go
@@ -31,6 +31,13 @@ import (
 	"github.com/docker/compose/v5/internal/memnet"
 )
 
+// ddProxyHost is a sentinel hostname stamped into the proxy URL handed to the
+// stdlib transport. It is never resolved on the network: the transport's
+// DialContext recognizes it and routes the connection to Docker Desktop's
+// proxy socket instead. The .invalid TLD is reserved (RFC 6761) so it can
+// never collide with a real registry host.
+const ddProxyHost = "docker-desktop-http-proxy.invalid"
+
 // Endpoint returns the Docker Desktop API socket endpoint advertised via the
 // engine info labels, or "" when the active engine is not Docker Desktop.
 func Endpoint(ctx context.Context, apiClient client.APIClient) (string, error) {
@@ -51,8 +58,8 @@ func Endpoint(ctx context.Context, apiClient client.APIClient) (string, error) {
 // when the input is not a recognized form or when the derived unix socket
 // does not exist (older DD versions or non-DD installs).
 //
-// On macOS/Linux: unix:///path/to/Data/docker-cli.sock      → unix:///path/to/Data/httpproxy.sock
-// On Windows:    npipe://./pipe/dockerDesktopLinuxEngine    → npipe://./pipe/dockerHttpProxy
+// On macOS/Linux: unix:///path/to/Data/docker-cli.sock     → unix:///path/to/Data/httpproxy.sock
+// On Windows:    npipe://\\.\pipe\docker_cli               → npipe://\\.\pipe\dockerHttpProxy
 func httpProxySocketEndpoint(endpoint string) string {
 	if sockPath, ok := strings.CutPrefix(endpoint, "unix://"); ok {
 		proxyPath := filepath.Join(filepath.Dir(sockPath), "httpproxy.sock")
@@ -62,7 +69,20 @@ func httpProxySocketEndpoint(endpoint string) string {
 		return "unix://" + proxyPath
 	}
 	if strings.HasPrefix(endpoint, "npipe://") {
-		return "npipe://./pipe/dockerHttpProxy"
+		// Named pipes all live in the same `\\.\pipe\` namespace, so only the
+		// trailing pipe name differs. Swap it in place rather than rebuilding
+		// the endpoint string: this preserves the engine endpoint's exact
+		// prefix (Docker Desktop reports the backslash form
+		// `npipe://\\.\pipe\docker_cli`), which winio can dial. Hardcoding
+		// `npipe://./pipe/...` instead would yield the relative path
+		// `./pipe/dockerHttpProxy`, which fails with "open
+		// ./pipe/dockerHttpProxy: The system cannot find the path specified."
+		// (docker/compose#13824). LastIndexAny handles both the backslash form
+		// above and the forward-slash form `npipe:////./pipe/...`.
+		if idx := strings.LastIndexAny(endpoint, `/\`); idx >= 0 {
+			return endpoint[:idx+1] + "dockerHttpProxy"
+		}
+		return ""
 	}
 	return ""
 }
@@ -73,6 +93,13 @@ func httpProxySocketEndpoint(endpoint string) string {
 // transport in that case — for the OCI resolver this means containerd's
 // built-in transport). Pass "" for endpoint when DD is not the active
 // engine.
+//
+// Loopback targets (localhost, 127.0.0.0/8, ::1) bypass the proxy and connect
+// directly, so `compose publish` to a local/insecure registry behaves like
+// `docker push` instead of failing inside the proxy (docker/compose#13824).
+// All other registry traffic continues through Docker Desktop's proxy so
+// Desktop stays in control of proxy decisions (e.g. enterprise-managed
+// proxies); the local process NO_PROXY is deliberately not honored.
 //
 // When DD is available, the returned transport is a clone of
 // http.DefaultTransport with only Proxy and DialContext overridden, so it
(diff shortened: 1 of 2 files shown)


ptrdom:
I have also manually tested this on MacOS (M1 Mac) with Docker Desktop, managed to reproduce the issue and confirmed that this PR fixes it.

glours (changes requested):
Hello @ptrdom 
Thanks for this contribution. 

I’m concerned about the `NO_PROXY` part of this change.

The loopback bypass makes sense and seems like the right fix for the localhost registry case. However, this PR also makes the Compose OCI transport honor the local process `NO_PROXY` / `no_proxy` environment before routing through Docker Desktop’s proxy.

That changes the behavior introduced in [66c21c3](https://github.com/docker/compose/commit/66c21c3b6c23b66d8fa88c3ff78b7becfa1bc3bb), where OCI traffic was routed through Docker Desktop’s PAC-aware proxy so Desktop can own proxy decisions, especially for enterprise environments with managed or non-standard proxy configurations.

With this change, a broad local `NO_PROXY` value such as `*`, `.corp`, or a specific internal registry host could bypass Docker Desktop’s proxy entirely and connect directly. That could break enterprise proxy setups or bypass centrally managed proxy policy.

Can we keep the direct bypass limited to loopback targets only, and continue routing all other OCI registry traffic through Docker Desktop’s proxy?

ptrdom:
@glours Thank you for a prompt review, I think your concern makes sense, I have pushed changes to address it.

glours (approved):
I don’t see a blocker in the code after this update. One thing I’d still like to adjust before merging is the PR description/title: they still mention insecure registries broadly and the body still says `NO_PROXY` is honored, which is no longer the intended behavior. Could you update the wording to make it clear that only loopback targets bypass the Desktop proxy, while all other registry traffic stays routed through Docker Desktop?

ptrdom:
Makes sense, updated, let me know if any further tweaks are needed.

glours (approved):
LGTM

glours:
@ptrdom can you do a `git commit --amend -s --no-edit` and then a `git push --force-with-lease ...` , there is a bug in GitHub that blocks me to merge the PR right now

ptrdom:
@glours Done, does it work now?

glours:
@ptrdom yes thank you, we just need to wait for the CI now

glours:
@ptrdom I was able to ask for auto-merge, but it was disabled automatically at the end of the CI process...
Screenshot 2026-06-15 at 16 58 59

Can you try to push an empty commit? Don't forget the sign-off

ptrdom:
@glours Just pushed, hopefully it is okay now. A bit of a flaky situation 😅

glours:
Ok we're using `rebase and merge` to integrate PR into main branch and there is a conflict with your 1st commit and the `golang.org/x/net v0.51.0` dependency.
I propose you to run `git rebase -i upstream/main` and squash everything in 1 commit (it will remove the empty commit), is it ok for you?

ptrdom:
@glours Is it okay now?

PR #13848, opened 2026-06-17
fix(publish): honor env_file required: false for missing files

### LINKED ISSUE #13648: [BUG] docker compose publish ignores `required: false` on `env_file`

Ijtihed:
**What I did**

When env_file declares required false and the file is missing then docker compose publish now skips the file instead of failing. If the optional file exists it is still scanned for sensitive data and included as a layer which preserves existing behavior.

Two locations needed guards:
- processFile: where env files are hashed and added as OCI layers
- checkForSensitiveData: where env files are scanned for secrets

Non-ENOENT errors (permissions, etc.) still propagate regardless of the required flag.

**Related issue**

Fixes #13648

Supersedes #13650, #13651, #13652 with a diff implementing what the maintainers request in #13652 review

image

(unrelated photo of my fiance's cute cat :3)

### DIFF
--- a/pkg/compose/publish.go
+++ b/pkg/compose/publish.go
@@ -255,8 +255,21 @@ func processFile(ctx context.Context, file string, project *types.Project, extFi
 	}
 	for name, service := range base.Services {
 		for i, envFile := range service.EnvFiles {
+			// A real stat failure (e.g. permissions) is fatal, but a missing file is not:
+			// the project loader already rejects missing required env files before we get
+			// here, so an absent file at this point is an optional one.
+			_, statErr := os.Stat(envFile.Path)
+			if statErr != nil && !os.IsNotExist(statErr) {
+				return nil, fmt.Errorf("failed to access env file %s: %w", envFile.Path, statErr)
+			}
+			// The hash is derived from the path string alone, so the env_file is always
+			// rewritten to its opaque <hash>.env placeholder, even for a missing optional
+			// file, so the published artifact never leaks the local path. Only files that
+			// exist are registered for upload, mirroring the extends handling below.
 			hash := fmt.Sprintf("%x.env", sha256.Sum256([]byte(envFile.Path)))
-			envFiles[envFile.Path] = hash
+			if statErr == nil {
+				envFiles[envFile.Path] = hash
+			}
 			f, err = transform.ReplaceEnvFile(f, name, i, hash)
 			if err != nil {
 				return nil, err
@@ -690,6 +703,15 @@ func (s *composeService) checkForSensitiveData(ctx context.Context, project *typ
 	for _, service := range project.Services {
 		// Check env files
 		for _, envFile := range service.EnvFiles {
+			if _, statErr := os.Stat(envFile.Path); statErr != nil {
+				if !os.IsNotExist(statErr) {
+					return nil, fmt.Errorf("failed to access env file %s: %w", envFile.Path, statErr)
+				}
+				if envFile.Required {
+					return nil, fmt.Errorf("env file %s not found", envFile.Path)
+				}
+				continue
+			}
 			findings, err := scan.ScanFile(envFile.Path)
 			if err != nil {
 				return nil, fmt.Errorf("failed to scan env file %s: %w", envFile.Path, err)
--- a/pkg/compose/publish_test.go
+++ b/pkg/compose/publish_test.go
@@ -17,7 +17,9 @@
 package compose
 
 import (
+	"crypto/sha256"
 	"errors"
+	"fmt"
 	"os"
 	"path/filepath"
 	"slices"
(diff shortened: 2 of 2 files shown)


pkg/compose/publish.go
```diff
@@ -255,6 +255,15 @@ func processFile(ctx context.Context, file string, project *types.Project, extFi
 	}
 	for name, service := range base.Services {
 		for i, envFile := range service.EnvFiles {
+			if _, statErr := os.Stat(envFile.Path); statErr != nil {
+				if !os.IsNotExist(statErr) {
+					return nil, fmt.Errorf("failed to access env file %s: %w", envFile.Path, statErr)
+				}
+				if envFile.Required {
+					return nil, fmt.Errorf("env file %s not found", envFile.Path)
+				}
```
glours: Nit: this `if envFile.Required` branch is unreachable here, the project loader (called from `runPublish` without `SkipResolveEnvironment`) already rejects missing required env files via compose-go's `loadEnvFile`. 
The test only reaches it by forcing `SkipResolveEnvironment: true`. 
Could you either drop the branch (and `Test_processFile_required_env_file_missing`) or add a comment marking it as belt-and-suspenders? 
The guard in `checkForSensitiveData` is fine, that one is reachable because `loadUnresolvedFile` does set `SkipResolveEnvironment`.

pkg/compose/publish.go
```diff
@@ -255,6 +255,15 @@ func processFile(ctx context.Context, file string, project *types.Project, extFi
...
+				continue
```
glours: Could you still call `transform.ReplaceEnvFile` with the hash in this skip branch? The hash is derived from `envFile.Path` only (not file contents), so it's deterministic whether the file exists or not. 
Without the rewrite, the published YAML keeps the publisher's local (resolved-to-absolute) path instead of the opaque `<hash>.env` placeholder, small filesystem-layout leak, and inconsistent with the file-present case.

glours (changes requested):
Thanks for picking this up and aligning with the #13652 review, the fix is in the right shape. Two small asks before merging:
                                                                                                                                                                                                                                                            
  1. **Dead branch in `processFile`** (`publish.go:262-264`): the `if envFile.Required` path is unreachable in production because the project loader already enforces `required:` at load time. Suggest dropping it (and its test) or commenting it as defensive only.                                                                                                                                                                                                                                           
  2. **Path leak on optional missing** (`publish.go:265`): when the file is absent we skip `ReplaceEnvFile`, so the published YAML keeps the local path instead of the `<hash>.env` placeholder. Since the hash is computed from the path string alone, we can still rewrite it — keeps the artifact filesystem-agnostic and consistent with the file-present case.

pkg/compose/publish.go
Ijtihed: Dropped them, thanks!

pkg/compose/publish.go
Ijtihed: Good catch and thanks! Done. path no longer leaks now :)

glours (approved):
LGTM, thanks for the contribution 🙏

PR #13849, opened 2026-06-17
fix(publish): accept short-form port mappings (fixes #13672)

### LINKED ISSUE #13672: [BUG] docker compose publish rejects short-form port mapping

Ijtihed:
**What I did**

`docker compose publish` failed with `'services[whoami].ports[0]' expected a map or struct, got "string"` when a service used short-form port syntax like `${DASHBOARD_PORT:-3000}:3000`.

`composeFileAsByteReader` used `LoadWithContext` which decodes ports into typed Go structs which rejects short-form string entries. Switching to `LoadModelWithContext` returns a raw map instead which in ths case avoids the struct decode

**Related issue**

Fixes #13672

### DIFF
--- a/pkg/compose/publish.go
+++ b/pkg/compose/publish.go
@@ -38,6 +38,7 @@ import (
 	"github.com/opencontainers/image-spec/specs-go"
 	v1 "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/sirupsen/logrus"
+	"go.yaml.in/yaml/v4"
 
 	"github.com/docker/compose/v5/internal/desktop"
 	"github.com/docker/compose/v5/internal/oci"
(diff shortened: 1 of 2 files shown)


Mohataseem89:
i have one question regarding coverage:
does this approach also handle other valid short-form port declarations beyond the reproducer, for example:

```yaml
ports:
  - "8080:80"
  - "127.0.0.1:8080:80"
  - "${PORT:-8080}:80"
```
It may be worth adding or extending tests to cover a few representative short-form variants to ensure the publish path remains aligned with Compose port syntax support

Ijtihed:
> i have one question regarding coverage: does this approach also handle other valid short-form port declarations beyond the reproducer, for example:
> 
> ```yaml
> ports:
>   - "8080:80"
>   - "127.0.0.1:8080:80"
>   - "${PORT:-8080}:80"
> ```
> 
> It may be worth adding or extending tests to cover a few representative short-form variants to ensure the publish path remains aligned with Compose port syntax support

Agree, done

Ijtihed:
ping :) @glours @ndeloof

pkg/compose/publish.go
```diff
@@ -724,11 +725,24 @@ func (s *composeService) checkForSensitiveData(ctx context.Context, project *typ
 }
 
 func composeFileAsByteReader(ctx context.Context, filePath string, project *types.Project) (io.Reader, error) {
```
ndeloof: I wonder we actually need to parse the compose file here. This func is only used by `checkForSensitiveData` and disables all compose-go processing to get raw file, so we could just open compose file as a raw text file and let `ScanReader` process. A user to use a non-yaml file here would anyway get a parsing error from other steps in the publish pipeline

Ijtihed: Thanks for the review, updated the PR

PR #13854, opened 2026-06-19
fix(watch): don't rebuild dependencies when a service changes

### LINKED ISSUE #13853: watch: rebuild also rebuilds depends_on services
## Description

In `compose watch`, editing a service's code rebuilds not only that service but also the services it `depends_on`. Those dependencies are then rebuilt but not restarted.

## Steps to reproduce

git clone https://github.com/dgageot/bug-compose

```yaml
services:
  backend:
    build: ./backend
  frontend:
    build: ./frontend
    depends_on:
      - backend
    develop:
      watch:
        - action: rebuild
          path: ./frontend
```

```
docker compose up --build --watch
# edit ./frontend/main.go
```

## Expected

Only `frontend` is rebuilt and restarted. `backend` is left untouched.

## Actual

`backend` is rebuilt too (but not restarted).

dgageot:
## What this does

When a watched service is rebuilt, Compose also rebuilt the services it `depends_on` (which were then left not restarted).

`up` builds its `BuildOptions` with `Deps = true`. The same options are reused by watch's `rebuild()`, which narrows `Build.Services` to the changed service but never cleared `Build.Deps`, so `build()` still included dependencies via `IncludeDependencies`.

This clears `Build.Deps` in `rebuild()` so only the changed services are built.

## Testing

Added e2e `TestWatchRebuildIgnoresDependencies` (plus a `rebuild-deps.yaml` fixture): a `frontend` that `depends_on: backend`, watched only on `frontend`. It asserts the frontend picks up the edit while the backend container ID is unchanged.

Fixes #13853

### DIFF
--- a/pkg/compose/watch.go
+++ b/pkg/compose/watch.go
@@ -138,6 +138,10 @@ func (s *composeService) Watch(ctx context.Context, project *types.Project, opti
 	return wait()
 }
 
+func selectWatchServices(project *types.Project, services []string) (*types.Project, error) {
+	return project.WithSelectedServices(services, types.IgnoreDependencies)
+}
+
 type watchRule struct {
 	types.Trigger
 	include watch.PathMatcher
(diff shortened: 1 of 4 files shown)


dgageot:
closing in favour of https://github.com/docker/compose/pull/13856

PR #13856, opened 2026-06-19
watch: do not rebuild depends_on services on file change

### LINKED ISSUE #13853: watch: rebuild also rebuilds depends_on services

ndeloof:
## Summary

- Fixes #13853: `compose up --build --watch` rebuilt every `depends_on` dependency whenever a watched service changed.
- Root cause: `up --build` sets `BuildOptions.Deps=true` to build dep images at startup. The watch `rebuild()` path mutated the shared `*BuildOptions` to restrict `Services` to the watched service but never reset `Deps`, so `s.build()` switched back to `IncludeDependencies` and walked up the graph.
- Fix: `rebuild()` now works on a local copy of `BuildOptions` and explicitly sets `Deps=false`. The local copy also removes the data race on the shared pointer when concurrent file events fire.
- Bonus: `doBuildBake` was emitting `Image X Building` and tracking `expectedImages` for every service in the project, not just the ones in `serviceToBeBuild`. Filter those side-effects to the actual build set — `cfg.Targets` is still populated for every service so `additional_contexts: service:xxx` references keep resolving.

### Why not the approach in #13854

That PR also forced `IgnoreDependencies` in `selectWatchServices`, which strips dependency configs from the selected project. `create()` and `start()` need those configs (network membership, env interpolation) so the side-effects bleed into the restart path. Resetting `Build.Deps` and filtering the bake event leak is enough.

## Test plan

- [x] New e2e `TestWatchRebuildIgnoresDependencies` (fixture `pkg/e2e/fixtures/watch/rebuild-deps.yaml`) launches `up --build --watch` on a `frontend` → `backend` stack, modifies a frontend-watched file, and asserts the rebuild log between `Rebuilding service(s) ["frontend"]` and `successfully built` never mentions `backend Building` / `backend Built`.
- [x] `TestWatchMultiServices` and `TestWatchExec` still pass.
- [x] `go test ./pkg/compose/...` passes.
- [x] `golangci-lint run --build-tags "e2e" ./pkg/compose/... ./pkg/e2e/...` clean.

### DIFF
--- a/pkg/compose/build_bake.go
+++ b/pkg/compose/build_bake.go
@@ -206,9 +206,14 @@ func (s *composeService) doBuildBake(ctx context.Context, project *types.Project
 		}
 
 		image := api.GetImageNameOrDefault(service, project.Name)
-		s.events.On(buildingEvent(image))
-
-		expectedImages[serviceName] = image
+		// project.Services lists every service (we still need their bake
+		// targets defined so additional_contexts: service:xxx references can
+		// resolve), but only emit "Building" progress and track expected
+		// images for services we actually plan to build.
+		if _, ok := serviceToBeBuild[serviceName]; ok {
+			s.events.On(buildingEvent(image))
+			expectedImages[serviceName] = image
+		}
 
 		pull := service.Build.Pull || options.Pull
 		noCache := service.Build.NoCache || options.NoCache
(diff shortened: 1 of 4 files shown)


glours (approved):
LGTM

PR #13862, opened 2026-06-22
Pre start init containers

glours:
**What I did**
Add support for the `pre_start` service lifecycle hook from the compose spec.                                                                                                                                                                             
                                                                                                                                                                                                                                                            
  Hooks run as ephemeral containers before a service's containers start, sharing the first replica's volumes via `VolumesFrom` and attached to the same networks. A non-zero hook exit gates service start.                                                 
                                                            
  Semantics:                                                                                                                                                                                                                                                
  - Runs **once per service**, only when no replica is already running (initial `up`, `--force-recreate`, spec change).
  - Not re-triggered on `scale up`.                                                                                                                                                                                                                         
  - `per_replica: true` is rejected up front; only `per_replica: false` is supported for now.                                                                                                                                                               
  - Hook image defaults to the service image (including `build:`); overridable per hook.
                                                                                                                                                                                                                                                            
  Covered by 8 unit tests and 10 E2E tests (success, failure gating, build inheritance, idempotent re-up, spec change, force-recreate, mid-sequence failure, ordering, scale-up, scaled service).                                                           
                                                                                                                                                                                                                                                            
  **Note**: the first commit adds a temporary `replace` to a `compose-go` fork exposing `PreStart` on `ServiceConfig`. To be reverted once a proper `compose-go` release lands.

**Related issue**
https://github.com/compose-spec/compose-spec/issues/11 & https://github.com/compose-spec/compose-spec/issues/291

<img width="2400" height="1920" alt="image" src="https://github.com/user-attachments/assets/f9c41984-cd0f-437e-9ba9-0674e0f1b1ea" />

### DIFF
--- a/pkg/compose/pre_start.go
+++ b/pkg/compose/pre_start.go
@@ -1,0 +1,301 @@
+/*
+   Copyright 2020 Docker Compose CLI authors
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package compose
+
+import (
+	"context"
+	"fmt"
+	"strconv"
+
+	"github.com/compose-spec/compose-go/v2/types"
+	"github.com/moby/moby/api/pkg/stdcopy"
+	"github.com/moby/moby/api/types/container"
+	"github.com/moby/moby/client"
+	"github.com/moby/moby/client/pkg/versions"
+	"github.com/sirupsen/logrus"
+
+	"github.com/docker/compose/v5/pkg/api"
+	"github.com/docker/compose/v5/pkg/utils"
+)
+
+// lowestNumberedContainer returns the container with the lowest
+// com.docker.compose.container-number label, so pre_start always targets the
+// same replica regardless of the order the daemon returned them in.
+// Panics on an empty slice; callers must guard.
+func lowestNumberedContainer(containers Containers) container.Summary {
+	pick := containers[0]
+	pickNum, _ := strconv.Atoi(pick.Labels[api.ContainerNumberLabel])
+	for _, ctr := range containers[1:] {
+		num, _ := strconv.Atoi(ctr.Labels[api.ContainerNumberLabel])
+		if num < pickNum {
+			pick, pickNum = ctr, num
+		}
+	}
+	return pick
+}
+
+// runPreStart executes the service's pre_start hooks sequentially, in declared
+// order. Each hook runs as an ephemeral container that shares the service
+// container's volumes via VolumesFrom and is attached to the same networks.
+// A non-zero exit gates service start.
+//
+// With per_replica: false (the only currently supported mode), the hook sees
+// the volumes of the first non-running replica only — anonymous volumes and
+// tmpfs mounts are per-replica and not shared. Use named volumes or bind
+// mounts for data the hook produces.
+func (s *composeService) runPreStart(ctx context.Context, project *types.Project, service types.ServiceConfig, ctr container.Summary, listener api.ContainerEventListener) error {
+	// Validate every hook up front so an unsupported entry never triggers any I/O.
+	for i, hook := range service.PreStart {
+		if hook.PerReplica {
+			return fmt.Errorf("service %q pre_start[%d]: per_replica is not yet supported; remove per_replica or set it to false", service.Name, i)
+		}
+	}
+	for i, hook := range service.PreStart {
+		if err := s.runPreStartHook(ctx, project, service, ctr, i, hook, listener); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (s *composeService) runPreStartHook(
+	ctx context.Context, project *types.Project, service types.ServiceConfig,
+	ctr container.Summary, index int, hook types.ServiceHook, listener api.ContainerEventListener,
+) error {
+	created, err := s.createPreStartContainer(ctx, project, service, ctr, hook)
+	if err != nil {
+		return err
+	}
+
+	// Subscribe to wait before start to avoid missing the exit event for short-lived hooks.
+	// WaitConditionNotRunning would match immediately because the container is still in
+	// "created" state, so use WaitConditionNextExit to block until the run actually finishes.
+	waitRes := s.apiClient().ContainerWait(ctx, created.ID, client.ContainerWaitOptions{
+		Condition: container.WaitConditionNextExit,
+	})
+
+	// Open the log stream before ContainerStart so AutoRemove cannot race us
+	// to a 404 on a fast-exiting hook. The dedicated logCtx lets us force the
+	// follow stream closed once the hook has exited, so a daemon that keeps
+	// the connection open cannot deadlock `<-logsDone`.
+	logCtx, cancelLogs := context.WithCancel(ctx)
+	defer cancelLogs()
+	logsDone := s.streamPreStartLogs(logCtx, created.ID, service, index, listener)
+
+	if _, err := s.apiClient().ContainerStart(ctx, created.ID, client.ContainerStartOptions{}); err != nil {
+		// AutoRemove only fires after a successful start, so the never-started
+		// container has to be dropped explicitly. A failed removal is logged
+		// at warn level — without that hint the orphan is only discoverable
+		// via the project/service labels.
+		if _, removeErr := s.apiClient().ContainerRemove(ctx, created.ID, client.ContainerRemoveOptions{Force: true}); removeErr != nil {
+			logrus.Warnf("service %q pre_start[%d]: failed to remove orphan hook container %s: %v", service.Name, index, created.ID, removeErr)
+		}
+		// Drain waitRes so the client's wait goroutine exits without having to
+		// wait for the parent context to be canceled.
+		select {
+		case <-waitRes.Error:
+		case <-waitRes.Result:
+		case <-ctx.Done():
+		}
+		cancelLogs()
+		<-logsDone
+		return err
+	}
+
+	waitErr := waitPreStart(ctx, service.Name, index, waitRes)
+	cancelLogs()
+	<-logsDone
+	return waitErr
+}
+
+func (s *composeService) createPreStartContainer(
+	ctx context.Context, project *types.Project, service types.ServiceConfig,
+	ctr container.Summary, hook types.ServiceHook,
+) (client.ContainerCreateResult, error) {
+	image := hook.Image
+	if image == "" {
+		image = api.GetImageNameOrDefault(service, project.Name)
+	}
+
+	cfg := &container.Config{
+		Image:      image,
+		Cmd:        hook.Command,
+		User:       hook.User,
+		WorkingDir: hook.WorkingDir,
+		Env:        append(ToMobyEnv(service.Environment), ToMobyEnv(hook.Environment)...),
+		// Tag the ephemeral hook container with the project/service it belongs
+		// to so a failed AutoRemove leaves something that `compose down` (and
+		// other label-scoped tooling) can still find.
+		Labels: map[string]string{
+			api.ProjectLabel: project.Name,
+			api.ServiceLabel: service.Name,
+			api.VersionLabel: api.ComposeVersion,
+		},
+	}
+	hostCfg := &container.HostConfig{
+		AutoRemove:  true,
+		Privileged:  hook.Privileged,
+		VolumesFrom: []string{ctr.ID},
+	}
+
+	apiVersion, err := s.RuntimeAPIVersion(ctx)
+	if err != nil {
+		return client.ContainerCreateResult{}, err
+	}
+
+	networkMode, networkingConfig, err := defaultNetworkSettings(project, service, 0, nil, true, apiVersion)
+	if err != nil {
+		return client.ContainerCreateResult{}, err
+	}
+	hostCfg.NetworkMode = networkMode
+
+	created, err := s.apiClient().ContainerCreate(ctx, client.ContainerCreateOptions{
+		Config:           cfg,
+		HostConfig:       hostCfg,
+		NetworkingConfig: networkingConfig,
+	})
+	if err != nil {
+		return client.ContainerCreateResult{}, err
+	}
+
+	if versions.LessThan(apiVersion, apiVersion144) {
+		if err := s.connectPreStartExtraNetworks(ctx, project, service, created.ID, networkMode); err != nil {
+			// Same reason as the ContainerStart-failure cleanup: AutoRemove never
+			// fires on a container that was created but not started. Surface
+			// any cleanup failure so the orphan is at least visible in logs.
+			if _, removeErr := s.apiClient().ContainerRemove(ctx, created.ID, client.ContainerRemoveOptions{Force: true}); removeErr != nil {
+				logrus.Warnf("service %q pre_start: failed to remove orphan hook container %s: %v", service.Name, created.ID, removeErr)
+			}
+			return client.ContainerCreateResult{}, err
+		}
+	}
+	return created, nil
+}
+
+// connectPreStartExtraNetworks mirrors the createMobyContainer fallback path for
+// older API versions: ContainerCreate only accepts one EndpointsConfig, so extra
+// networks have to be attached via NetworkConnect after creation.
+func (s *composeService) connectPreStartExtraNetworks(ctx context.Context, project *types.Project, service types.ServiceConfig, containerID string, primary container.NetworkMode) error {
+	for _, networkKey := range service.NetworksByPriority() {
+		mobyNetworkName := project.Networks[networkKey].Name
+		if string(primary) == mobyNetworkName {
+			continue
+		}
+		eps, err := createEndpointSettings(project, service, 0, networkKey, nil, true)
+		if err != nil {
+			return err
+		}
+		if _, err := s.apiClient().NetworkConnect(ctx, mobyNetworkName, client.NetworkConnectOptions{
+			Container:      containerID,
+			EndpointConfig: eps,
+		}); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func waitPreStart(ctx context.Context, serviceName string, index int, waitRes client.ContainerWaitResult) error {
+	// ContainerWait can deliver on Result and Error at the same instant. Two
+	// races have to be closed deterministically here:
+	//   1. The daemon closing a successful stream cleanly sends nil on Error
+	//      AND the exit code on Result — a plain 3-case select would let Go
+	//      pick the Error branch and report a spurious "wait ended" failure.
+	//   2. A real transport error on Error can race with a stale Result — if
+	//      the scheduler picks Result, we would silently drop the error and
+	//      let the service start.
+	// Loop until Result is delivered, nil-ing the Error channel after a nil
+	// receive so a closed channel cannot busy-loop. After Result lands, do a
+	// non-blocking check on Error so a real error still wins over Result.
+	errCh := waitRes.Error
+	for {
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case res := <-waitRes.Result:
+			select {
+			case err := <-errCh:
+				if err != nil {
+					return err
+				}
+			default:
+			}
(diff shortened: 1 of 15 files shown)


ndeloof (dismissed):
LGTM. I wonder we could get this managed by reconciliation, so orchestrating init container is part of the reconciliation plan, this can come later in a follow-up PR

ndeloof (dismissed):

pkg/compose/pre_start.go
```diff
@@ -0,0 +1,240 @@
...
+		Config:           cfg,
+		HostConfig:       hostCfg,
+		NetworkingConfig: networkingConfig,
+	})
+	if err != nil {
+		return client.ContainerCreateResult{}, err
+	}
+
+	if versions.LessThan(apiVersion, apiVersion144) {
+		if err := s.connectPreStartExtraNetworks(ctx, project, service, created.ID, networkMode); err != nil {
```
docker-agent, line 176: **[MEDIUM] Orphaned container leaked when `connectPreStartExtraNetworks` fails**

When `connectPreStartExtraNetworks` returns an error (e.g. the network does not exist, or the daemon rejects the `NetworkConnect` call on older API versions), `createPreStartContainer` returns the error immediately without removing the container it already created:

```go
if versions.LessThan(apiVersion, apiVersion144) {
    if err := s.connectPreStartExtraNetworks(ctx, project, service, created.ID, networkMode); err != nil {
        return client.ContainerCreateResult{}, err   // ← created container is abandoned
    }
}
```

`AutoRemove: true` only fires **after** a container exits following a successful start; a container that was created but never started is not auto-removed. The caller `runPreStartHook` only calls `ContainerRemove` on the `ContainerStart` failure path, so this container silently persists in "created" state until the user runs `docker compose down`.

**Suggested fix** — remove the container before propagating the error:

```go
if versions.LessThan(apiVersion, apiVersion144) {
    if err := s.connectPreStartExtraNetworks(ctx, project, service, created.ID, networkMode); err != nil {
        _, _ = s.apiClient().ContainerRemove(ctx, created.ID,
            client.ContainerRemoveOptions{Force: true})
        return client.ContainerCreateResult{}, err
    }
}
```

pkg/compose/pre_start.go
```diff
@@ -0,0 +1,240 @@
...
+	// Open the log stream before ContainerStart so AutoRemove cannot race us
+	// to a 404 on a fast-exiting hook.
+	logsDone := s.streamPreStartLogs(ctx, created.ID, service, index, listener)
+
+	if _, err := s.apiClient().ContainerStart(ctx, created.ID, client.ContainerStartOptions{}); err != nil {
+		// AutoRemove only fires after a successful start, so an aborted hook would
+		// otherwise leave a ghost container and keep the follow log stream and the
+		// pending wait request open until the parent context is canceled.
+		_, _ = s.apiClient().ContainerRemove(ctx, created.ID, client.ContainerRemoveOptions{Force: true})
+		<-logsDone
```
docker-agent, line 115: **[LOW] Log goroutine drain after `ContainerStart` failure depends on Docker daemon closing the stream**

When `ContainerStart` fails, the code removes the container and then blocks on `<-logsDone` to drain the log goroutine:

```go
_, _ = s.apiClient().ContainerRemove(ctx, created.ID, client.ContainerRemoveOptions{Force: true})
<-logsDone
return err
```

The log goroutine is blocked inside `stdcopy.StdCopy(w, w, logs)` on the container's log stream (opened with `Follow: true`). For `<-logsDone` to unblock promptly, Docker must close the log stream when the container is removed. This is standard daemon behaviour, so it works in practice.

However, if the caller's context `ctx` is already cancelled at this point (e.g. the user pressed Ctrl+C just before `ContainerStart` was issued), `ContainerRemove` receives the cancelled context and may itself return immediately with a context error — without actually removing the container. The goroutine then has no path to exit until the underlying HTTP connection is torn down by the client. Because `ContainerLogs` was opened with the same `ctx`, the cancelled context will indeed close that connection, so the goroutine does eventually exit. The current behaviour is safe, but a small comment noting this dependency would help future readers.

docker-agent (commented):

This PR adds well-structured `pre_start` lifecycle hook support with good test coverage (8 unit tests + 10 E2E tests). One confirmed resource leak on older API versions and one low-severity observation about log goroutine cleanup are noted below.

pkg/compose/pre_start.go
```diff
@@ -0,0 +1,240 @@
...
+		image = api.GetImageNameOrDefault(service, project.Name)
+	}
+
+	cfg := &container.Config{
+		Image:      image,
+		Cmd:        hook.Command,
+		User:       hook.User,
+		WorkingDir: hook.WorkingDir,
+		Env:        append(ToMobyEnv(service.Environment), ToMobyEnv(hook.Environment)...),
+	}
```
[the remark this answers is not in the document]
glours: This is one of the purpose of the hook, to be able to use service image definition and changes some properties such as the command or to override env variables...

pkg/compose/pre_start.go
```diff
@@ -0,0 +1,243 @@
...
+		}
+	}
+	return nil
+}
+
+func waitPreStart(ctx context.Context, serviceName string, index int, waitRes client.ContainerWaitResult) error {
+	select {
+	case <-ctx.Done():
+		return ctx.Err()
+	case err := <-waitRes.Error:
```
docker-agent: **[MEDIUM] `waitPreStart` returns nil when the error channel fires with a nil value, bypassing exit-code check**

In the `select` statement, `case err := <-waitRes.Error: return err` will return `nil` if the Docker client sends a nil value on the error channel. This bypasses the `waitRes.Result` case that checks `res.StatusCode != 0`, potentially allowing a non-zero hook exit to go undetected and the dependent service to start despite the hook failing.

The Docker client's `ContainerWait` API uses two channels: `Error` for transport/connection-level errors and `Result` for the container exit event. In practice the error channel should only fire with non-nil values, but edge cases (unexpected stream close, context handling in older client versions) may result in a nil send on this channel. Adding a nil guard would make this robust:

```go
case err := <-waitRes.Error:
    if err != nil {
        return err
    }
    // fall through or select again — nil on error channel isn't a success signal
```

pkg/compose/pre_start.go
```diff
@@ -0,0 +1,243 @@
...
+	// "created" state, so use WaitConditionNextExit to block until the run actually finishes.
+	waitRes := s.apiClient().ContainerWait(ctx, created.ID, client.ContainerWaitOptions{
+		Condition: container.WaitConditionNextExit,
+	})
+
+	// Open the log stream before ContainerStart so AutoRemove cannot race us
+	// to a 404 on a fast-exiting hook.
+	logsDone := s.streamPreStartLogs(ctx, created.ID, service, index, listener)
+
+	if _, err := s.apiClient().ContainerStart(ctx, created.ID, client.ContainerStartOptions{}); err != nil {
```
docker-agent, line 99: **[MEDIUM] `ContainerWait` goroutine left unjoined when `ContainerStart` fails**

When `ContainerStart` returns an error, the code force-removes the container and drains `logsDone`, but never drains the `waitRes` channels. The `ContainerWait` goroutine inside the Docker client is left running until the parent context is canceled.

The comment acknowledges this ("keep the follow log stream and the pending wait request open until the parent context is canceled"), but whether force-removing the container causes the daemon to close the wait stream promptly is not guaranteed. If it does not, the goroutine leaks for the lifetime of the `ctx`, which in a `docker compose up` is typically the full session.

Consider draining `waitRes` after `ContainerRemove` to ensure the goroutine terminates before returning:

```go
_, _ = s.apiClient().ContainerRemove(ctx, created.ID, client.ContainerRemoveOptions{Force: true})
// Drain waitRes so the Docker client's goroutine can exit
select {
case <-waitRes.Error:
case <-waitRes.Result:
case <-ctx.Done():
}
<-logsDone
return err
```

docker-agent (commented):

Two medium-severity issues found in `pkg/compose/pre_start.go`, both in the new hook execution path. The overall design is solid — cleanup on error paths is well-handled, the deterministic replica selection for VolumesFrom is correct, and the test coverage is thorough. The two items below are edge-case robustness concerns in the wait/cleanup logic.

pkg/compose/pre_start.go
```diff
@@ -0,0 +1,243 @@
...
+	pick := containers[0]
+	pickNum, _ := strconv.Atoi(pick.Labels[api.ContainerNumberLabel])
+	for _, ctr := range containers[1:] {
+		num, _ := strconv.Atoi(ctr.Labels[api.ContainerNumberLabel])
+		if num < pickNum {
+			pick, pickNum = ctr, num
+		}
+	}
+	return pick
+}
```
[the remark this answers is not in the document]
glours: The label is added by Compose itself, so no need to add defensive guard here

pkg/compose/pre_start.go
```diff
@@ -0,0 +1,255 @@
...
+			EndpointConfig: eps,
+		}); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func waitPreStart(ctx context.Context, serviceName string, index int, waitRes client.ContainerWaitResult) error {
+	select {
```
docker-agent: **[HIGH] `waitPreStart`: nil-error from closed channel may race with `Result`, causing spurious hook-failure errors**

`waitPreStart` selects over two simultaneously-ready channels and Go's scheduler picks one at random. The Docker daemon's `ContainerWait` stream, when it closes cleanly without an explicit error payload, sends a zero-value nil to the `Error` channel while *also* delivering the exit code on the `Result` channel. When both are ready at the same time, Go's `select` can nondeterministically pick the `Error` case, executing the `err == nil` branch and returning:

```
service "web" pre_start[0]: wait ended without an exit status
```

…even though the hook exited successfully with code 0. The effect is that `compose up` fails intermittently on healthy hooks — the failure rate depends purely on goroutine scheduling. The unit tests avoid this path by only pre-populating `Result` and leaving `Error` empty, so the race is not covered.

**Suggested fix**: drain `Result` first and only fall back to `Error` on a separate select, or treat a nil error as a no-op (allow the second `select` iteration to pick up the result):

```go
select {
case <-ctx.Done():
    return ctx.Err()
case res := <-waitRes.Result:
    if res.Error != nil {
        return fmt.Errorf(...)
    }
    if res.StatusCode != 0 {
        return fmt.Errorf(...)
    }
    return nil
case err := <-waitRes.Error:
    if err != nil {
        return err
    }
    // nil means the stream closed without a result — treat as unknown exit
    return fmt.Errorf("service %q pre_start[%d]: wait ended without an exit status", serviceName, index)
}
```

Alternatively, add a test case that delivers nil to the `Error` channel simultaneously with a valid `Result` to verify the scheduler never picks the error path for a successful run.

pkg/compose/pre_start.go
```diff
@@ -0,0 +1,255 @@
...
+			Type:    api.HookEventLog,
+			Source:  source,
+			ID:      containerID,
+			Service: service.Name,
+			Line:    fmt.Sprintf("warning: could not attach pre_start log stream: %s", err),
+		})
+		close(done)
+		return done
+	}
+	go func() {
```
docker-agent, line 285: **[MEDIUM] `streamPreStartLogs`: goroutine may block indefinitely if daemon never closes the log stream**

The goroutine calls `stdcopy.StdCopy(w, w, logs)`, which reads from the HTTP response body until EOF. Because `ContainerLogs` is called with `ctx`, context cancellation propagates through the HTTP transport and should terminate the read — but only when `ctx` is cancelled. In the normal success path, `waitPreStart` returns as soon as the container exits, but `<-logsDone` is called unconditionally *after* that:

```go
waitErr := waitPreStart(ctx, service.Name, index, waitRes)
<-logsDone   // blocks here until the log goroutine finishes
```

If the daemon keeps the log stream open after the container exits (e.g., due to a daemon bug or a hang), `stdcopy.StdCopy` never returns, `done` is never closed, and the call to `<-logsDone` stalls the entire `compose up` indefinitely. The caller has no timeout or escape hatch beyond waiting for `ctx` to be cancelled externally.

**Suggested mitigation**: after `waitPreStart` returns, cancel a child context dedicated to the log stream, or call `logs.Close()` from the caller side to force `stdcopy.StdCopy` to unblock:

```go
logCtx, cancelLogs := context.WithCancel(ctx)
defer cancelLogs()
logsDone := s.streamPreStartLogs(logCtx, created.ID, service, index, listener)
// ...
waitErr := waitPreStart(ctx, service.Name, index, waitRes)
cancelLogs() // signal the log goroutine to stop
<-logsDone
```

pkg/compose/pre_start.go
```diff
@@ -0,0 +1,255 @@
...
+		return client.ContainerCreateResult{}, err
+	}
+
+	networkMode, networkingConfig, err := defaultNetworkSettings(project, service, 0, nil, true, apiVersion)
+	if err != nil {
+		return client.ContainerCreateResult{}, err
+	}
+	hostCfg.NetworkMode = networkMode
+
+	created, err := s.apiClient().ContainerCreate(ctx, client.ContainerCreateOptions{
```
docker-agent, line 166: **[MEDIUM] Hook containers created without Compose labels — orphaned containers invisible to `compose down`**

`createPreStartContainer` calls `ContainerCreate` with a `Config` that has no `Labels` set. All regular service containers carry `com.docker.compose.project`, `com.docker.compose.service`, and related labels so that `compose down` can discover and remove them. Hook containers omit these labels entirely.

Because `AutoRemove: true` is set, the container normally removes itself on exit, so no label is needed for the happy path. However, if the explicit `ContainerRemove` call in the `ContainerStart`-failure path silently fails (its error is discarded with `_, _`), the orphaned container has no name and no Compose labels — it is completely invisible to `compose down` and can only be found by inspecting all containers on the host.

**Suggested fix**: add at minimum the project/service labels (reusing the existing `api.ProjectLabel`, `api.ServiceLabel` constants) to `cfg.Labels` when creating the hook container. This allows future cleanup tooling to identify and remove stale hook containers:

```go
cfg := &container.Config{
    // ...
    Labels: map[string]string{
        api.ProjectLabel: project.Name,
        api.ServiceLabel: service.Name,
        api.VersionLabel: api.ComposeVersion,
    },
}
```

pkg/compose/pre_start.go
```diff
@@ -0,0 +1,255 @@
...
+	})
+
+	// Open the log stream before ContainerStart so AutoRemove cannot race us
+	// to a 404 on a fast-exiting hook.
+	logsDone := s.streamPreStartLogs(ctx, created.ID, service, index, listener)
+
+	if _, err := s.apiClient().ContainerStart(ctx, created.ID, client.ContainerStartOptions{}); err != nil {
+		// AutoRemove only fires after a successful start, so the never-started
+		// container has to be dropped explicitly.
+		_, _ = s.apiClient().ContainerRemove(ctx, created.ID, client.ContainerRemoveOptions{Force: true})
```
docker-agent: **[LOW] `ContainerRemove` error silently discarded after `ContainerStart` failure**

When `ContainerStart` fails, the code attempts to clean up the never-started container with `ContainerRemove`, but the result is discarded:

```go
_, _ = s.apiClient().ContainerRemove(ctx, created.ID, client.ContainerRemoveOptions{Force: true})
```

If `ContainerRemove` itself fails (e.g., the daemon is unavailable, or the container has already been removed by another process), the container is silently orphaned. The primary `ContainerStart` error is returned to the caller, but there is no log message or warning that the cleanup failed.

This pattern is common in Docker Compose cleanup code, but given that hook containers have no Compose labels (see the related finding above), a failed remove is particularly unrecoverable. Logging the error at debug/warning level would help operators diagnose lingering containers:

```go
if _, removeErr := s.apiClient().ContainerRemove(ctx, created.ID, ...); removeErr != nil {
    // log or emit a warning event
}
```

docker-agent (commented):
### Assessment: 🔴 CRITICAL

This PR adds well-structured `pre_start` lifecycle hook support with solid test coverage (8 unit tests + 10 E2E tests). One high-severity race condition in `waitPreStart` can cause intermittent spurious failures for healthy hooks, two medium findings relate to goroutine blocking and missing Compose labels on hook containers, and one low finding is a silent discard of a cleanup error. All findings are in new code introduced by this PR.

docker-agent (commented):
### Assessment: 🟢 APPROVE

The `pre_start` lifecycle hook implementation is well-structured and carefully handles the tricky aspects of ephemeral container management:

- **Correct lifecycle ordering**: `ContainerWait` is subscribed before `ContainerStart` to avoid missing exit events for fast-exiting hooks, and the log stream is opened before start to race-safely avoid AutoRemove-vs-404.
- **Orphan cleanup on failure**: If `ContainerStart` fails, the never-started hook container is explicitly force-removed (since `AutoRemove` only fires after a successful start), with a warn-level log if removal itself fails.
- **`waitPreStart` race handling**: The two-channel ContainerWait race (nil-on-Error coinciding with Result) is handled correctly. The `errCh = nil` pattern after a clean nil receive correctly prevents a busy-loop on a closed channel; since Docker's Error channel delivers at most one event, no real error can arrive after the nil. The non-blocking post-Result check still catches concurrent real errors when errCh has not yet been nil'd.
- **Log stream cancellation**: A dedicated `logCtx` prevents a daemon that holds the log stream open from deadlocking `<-logsDone` after the hook exits.
- **Skip condition**: `len(serviceContainers) == len(toStart)` correctly gates the hook to initial start (no replicas running), and is skipped on scale-up as specified.

No bugs introduced by this PR were confirmed. The implementation is solid.

ndeloof (approved):

StefanWetterActiware:
Hey @glours, i really really like this. **Thanks a lot**! 🥇 
(I already used "service-containers" with "depends_on" "service_completed.." for it with completely doubled (anchored etc.) configuration)
But I miss the log of the pre_start-Container somewhere persisted.
While the container is running one can see the docker logs (if you are fast enough if it is a short task), but it seems they are not persisted anywhere. Correct?

glours:
Hi @StefanWetterActiware, 
Thanks a lot for the feedback, you're right: hook containers are auto-removed on exit, so their logs are only streamed live during `up` and aren't retrievable afterwards (and in `up -d` they currently aren't shown at all, which we're treating as a bug on its own).

  We're exploring a few ways to persist them, and your answers would really help us pick the right one:

  1. **When do you need the logs?** Mostly to debug a *failed* `pre_start` (why didn't my service start?), or also for *successful* runs (e.g. auditing what a migration did)?
  2. **How do you usually run Compose?** `up` attached or `up -d`? If the full hook output were reliably printed during `up` (including detached mode, and in the error when a hook fails), would that cover your need, or do you need to retrieve logs *after* the fact?
  3. **Where would you expect to find them?** Via `docker compose logs`, in a file on disk, or through your existing logging setup (do you use a centralized logging driver like `syslog`/`fluentd`/`journald`, or the default `json-file`)?
  4. **How much history matters?** Only the latest run per hook, or several runs back?
  5. **What's the context?** Local dev, CI, or a server-like environment? (In CI the `up` output is usually already captured by the runner.)
  6. **Would a kept container bother you?** One option is to keep the hook container *only when it fails* (removed on success, at most one per hook, cleaned up on the next `up`/`down`) so you can `docker logs` / `docker inspect` it, given you moved away from the `depends_on` hack partly because of `ps` pollution, would that be acceptable, or would you rather have zero leftover containers and logs written elsewhere?

 Thanks again — this kind of real-world input is exactly what helps us shape the feature.

StefanWetterActiware:
Hi @glours,

>     1. **When do you need the logs?** Mostly to debug a _failed_ `pre_start` (why didn't my service start?), or also for _successful_ runs (e.g. auditing what a migration did)?
I would like to persist them. Even on success. But in my case an additional init-Volume would do it, i could write a logfile from the application running as init.

>     2. **How do you usually run Compose?** `up` attached or `up -d`? If the full hook output were reliably printed during `up` (including detached mode, and in the error when a hook fails), would that cover your need, or do you need to retrieve logs _after_ the fact?
I always go `-d`

>     3. **Where would you expect to find them?** Via `docker compose logs`, in a file on disk, or through your existing logging setup (do you use a centralized logging driver like `syslog`/`fluentd`/`journald`, or the default `json-file`)?
My personal choice would be IN the "main" containers log, but of course it is a very different container...

>     4. **How much history matters?** Only the latest run per hook, or several runs back?
Only latest.

>     5. **What's the context?** Local dev, CI, or a server-like environment? (In CI the `up` output is usually already captured by the runner.)
My context is a server-like environment.

>     6. **Would a kept container bother you?** One option is to keep the hook container _only when it fails_ (removed on success, at most one per hook, cleaned up on the next `up`/`down`) so you can `docker logs` / `docker inspect` it, given you moved away from the `depends_on` hack partly because of `ps` pollution, would that be acceptable, or would you rather have zero leftover containers and logs written elsewhere?
Yes, that is wat we have now before 5.3 as own container with startup-dependency "service_completed_successfully".

Thanks a lot for investing your time here.

Regards
Stefan

PR #13873, opened 2026-06-24
fix(oci): route authorizer token fetches through provided transport

glours:
**What I did**
The transport passed to NewResolver was applied only via docker.WithClient. Pass it via docker.WithAuthClient too so the authorizer's OAuth token fetches use it instead of http.DefaultClient.

**Related issue**
https://docker.atlassian.net/browse/DDB-498

### DIFF
--- a/internal/oci/resolver_test.go
+++ b/internal/oci/resolver_test.go
@@ -19,6 +19,7 @@ package oci
 import (
 	"net/http"
 	"net/http/httptest"
+	"strings"
 	"sync/atomic"
 	"testing"
 
@@ -27,14 +28,20 @@ import (
 )
 
 // recordingRoundTripper counts RoundTrip invocations on a delegate so tests
-// can verify a supplied transport is actually used by the resolver.
+// can verify a supplied transport is actually used by the resolver. It also
+// tracks token-endpoint calls separately so tests can assert the authorizer's
+// token fetch goes through the same transport.
 type recordingRoundTripper struct {
-	delegate http.RoundTripper
-	calls    atomic.Int32
+	delegate  http.RoundTripper
+	calls     atomic.Int32
+	authCalls atomic.Int32
 }
 
 func (r *recordingRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
 	r.calls.Add(1)
+	if strings.HasSuffix(req.URL.Path, "/token") {
+		r.authCalls.Add(1)
+	}
 	return r.delegate.RoundTrip(req)
 }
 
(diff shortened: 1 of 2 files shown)


internal/oci/resolver_test.go
```diff
@@ -64,6 +71,37 @@ func TestNewResolver_UsesProvidedTransport(t *testing.T) {
...
+// http.DefaultClient.
+func TestNewResolver_AuthorizerUsesProvidedTransport(t *testing.T) {
+	var server *httptest.Server
+	server = httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if strings.HasSuffix(r.URL.Path, "/token") {
+			_, _ = w.Write([]byte(`{"token":"fake","access_token":"fake","expires_in":300}`))
+			return
+		}
+		if r.Header.Get("Authorization") == "" {
+			w.Header().Set("Www-Authenticate", `Bearer realm="`+server.URL+`/token",service="test"`)
```
docker-agent, line 85: **[LOW] Non-standard `Www-Authenticate` header casing in test**

The test sets `w.Header().Set("Www-Authenticate", ...)`. Go's `net/http` package canonicalises header names via `textproto.CanonicalMIMEHeaderKey`, so the wire value is `Www-Authenticate`. HTTP is case-insensitive, and the containerd docker authorizer reads challenges via `http.Header.Get` (also case-insensitive), so this works today.

However, the conventional and unambiguous spelling is `WWW-Authenticate` (as used in RFC 6750 and throughout Docker/containerd documentation). Using the non-standard casing is a latent fragility: any challenge parser that performs a direct map lookup (rather than going through `http.Header.Get`) would miss the challenge entirely, leaving `authCalls` at 0 and causing the test to fail in a non-obvious way. Consider changing to `w.Header().Set("WWW-Authenticate", ...)` to match the canonical form.

internal/oci/resolver_test.go
```diff
@@ -64,6 +71,37 @@ func TestNewResolver_UsesProvidedTransport(t *testing.T) {
 		"resolver did not invoke the supplied transport — wiring is broken")
 }
 
+// TestNewResolver_AuthorizerUsesProvidedTransport guards that the docker
+// authorizer's token fetch goes through the supplied transport, not
+// http.DefaultClient.
+func TestNewResolver_AuthorizerUsesProvidedTransport(t *testing.T) {
+	var server *httptest.Server
```
docker-agent, line 78: **[LOW] Split `var server; server = httptest.NewServer(...)` can be collapsed**

The two-step declaration (`var server *httptest.Server` on one line, then `server = httptest.NewServer(...)` on the next) is necessary here only because the handler closure references `server` before the assignment expression is complete — a classic self-referential server setup. The pattern is safe (the handler is never invoked before `httptest.NewServer` returns), but the split form can mislead readers into thinking there's a nil-pointer risk between declaration and assignment.

A clarifying comment alongside it (e.g. `// server is referenced by its own handler; two-step init is intentional`) would make the intent explicit and prevent future editors from "simplifying" it incorrectly.

docker-agent (commented):
### Assessment: 🟢 APPROVE

The production fix is correct and well-scoped: extracting authorizer options into `authOpts` and conditionally appending `docker.WithAuthClient(&http.Client{Transport: transport})` ensures OAuth token fetches use the caller-supplied transport rather than `http.DefaultClient`. Sharing the same `http.RoundTripper` between two `http.Client` instances is safe — `http.Transport` is designed to be shared. The new test correctly exercises the auth dance through the recording transport. Two minor test style observations are left as inline comments.

internal/oci/resolver_test.go
glours: ignored

internal/oci/resolver_test.go
glours: ignored

internal/oci/resolver_test.go
docker-agent: Understood, thanks for the feedback.

internal/oci/resolver_test.go
docker-agent: Understood — happy to defer to your judgment on this one.

ndeloof (approved):

PR #13889, opened 2026-06-30
Validate bind source for create_host_path false

scarab-systems:
**What I did**

Added Compose-side validation for bind mounts that set `bind.create_host_path: false`.

Compose now returns a missing bind source error before container creation when the resolved bind source path does not exist. This keeps the `create_host_path: false` contract enforced without relying on daemon or Desktop path handling.

I also added unit and end-to-end coverage for `docker compose run --rm` with a missing long-syntax bind source.

**Related issue**

Related to #13602

**Validation**

- `go test ./pkg/compose -run 'TestBuildContainerVolumesReturnsErrorWhenBindSourceMissingAndCreateHostPathFalse|Test_buildContainerVolumes' -count=1`
- `go build -trimpath -tags e2e -o bin/build/docker-compose ./cmd`
- `go test -v -count=1 ./pkg/e2e -run TestRunDoesNotCreateMissingBindSourceWhenCreateHostPathIsFalse`
- `docker buildx bake lint`

### DIFF
--- a/pkg/compose/create.go
+++ b/pkg/compose/create.go
@@ -876,22 +876,13 @@ func (s *composeService) buildContainerVolumes(
 	for _, m := range mountOptions {
 		switch m.Type {
 		case mount.TypeBind:
-			// `Mount` is preferred but does not offer option to created host path if missing
-			// so `Bind` API is used here with raw volume string
-			// see https://github.com/moby/moby/issues/43483
-			v := findVolumeByTarget(service.Volumes, m.Target)
-			if v != nil {
-				if v.Type != types.VolumeTypeBind {
-					v.Source = m.Source
-				}
-				if !bindRequiresMountAPI(v.Bind) {
-					source := m.Source
-					if vol := findVolumeByName(p.Volumes, m.Source); vol != nil {
-						source = m.Source
-					}
-					binds = append(binds, toBindString(source, v))
-					continue
-				}
+			bind, ok, err := buildContainerBindMount(p, service, m)
+			if err != nil {
+				return nil, nil, err
+			}
+			if ok {
+				binds = append(binds, bind)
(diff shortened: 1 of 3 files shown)


scarab-systems:
Gentle follow-up on this PR. The visible checks are not showing failures on my side, and I’m happy to make adjustments if there is anything maintainers would like changed.

If this is not a direction the project wants to take, I’m also happy to close it rather than leave noise in the queue.

pkg/compose/create.go
```diff
@@ -955,6 +971,23 @@ func findVolumeByTarget(volumes []types.ServiceVolumeConfig, target string) *typ
...
 
+func validateBindSource(volume *types.ServiceVolumeConfig, source string) error {
+	if volume.Type != types.VolumeTypeBind || volume.Bind == nil || bool(volume.Bind.CreateHostPath) {
+		return nil
+	}
+	return validateBindSourceExists(source)
+}
+
+func validateBindSourceExists(source string) error {
+	if _, err := os.Stat(source); err != nil {
```
ndeloof, line 982: This assumes source is LOCAL, but docker engine may be remote. There's no way to check bind source exists from compose CLI

ndeloof (changes requested):
This won't work with a remote docker engine, and we can't offer such a feature

pkg/compose/create.go
scarab-systems: Thanks for the review, that makes sense. I was validating from the Compose CLI side, but I agree that this gives the wrong answer when the Docker engine is remote because the bind source is resolved on the engine host rather than necessarily on the CLI host.

I’ll close this PR rather than keep pushing a CLI-side check that can’t be correct for remote contexts. Thanks for taking a look.

PR #13896, opened 2026-07-03
ci: harden GitHub Actions workflows

glours:
**What I did**
The pinned codeql-action/upload-sarif v2 (v2.28.1) falls in the vulnerable range of CVE-2025-24362 and the v2 line has no patched release, so bump to v3.36.3. Enable Dependabot for github-actions to keep action pins from going stale. Scope the release job to tag refs so its contents:write token is only minted when a release is actually created. In merge.yml, drop a dead conditional (workflow only triggers on push) and pass DOCKERDESKTOP_REPO to github-script via env rather than inline interpolation, as recommended against script injection.

**Related issue**
N/A

image

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -304,6 +304,7 @@ jobs:
           files: ./coverage.txt
 
   release:
+    if: startsWith(github.ref, 'refs/tags/v')
     permissions:
       contents: write # to create a release (ncipollo/release-action)
     runs-on: ubuntu-latest
@@ -329,7 +330,6 @@ jobs:
           find bin/release -type f -exec file -e ascii -- {} +
       -
         name: GitHub Release
-        if: startsWith(github.ref, 'refs/tags/v')
         uses: ncipollo/release-action@339a81892b84b4eeb0f6e744e4574d79d0d9b8dd # v1.21.0
         with:
           artifacts: ./bin/release/*
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -102,6 +102,14 @@ jobs:
             fi
           done
       -
+        name: List artifacts
+        run: |
+          tree -nh ./bin/release
+      -
+        name: Check artifacts
+        run: |
+          find bin/release -type f -exec file -e ascii -- {} +
+      -
         name: Upload artifacts
         uses: actions/upload-artifact@bbbca2ddaa5d8feaa63e36b76fdaad77386f024f # v7.0.0
         with:
@@ -306,7 +314,7 @@ jobs:
   release:
     if: startsWith(github.ref, 'refs/tags/v')
     permissions:
-      contents: write # to create a release (ncipollo/release-action)
+      contents: write # to create a release (softprops/action-gh-release)
     runs-on: ubuntu-latest
     needs:
       - binary-finalize
@@ -321,6 +329,8 @@ jobs:
           path: ./bin/release
           name: release
       -
+        # inspect what was actually downloaded from the artifact store —
+        # this is what gets attached to the release
         name: List artifacts
         run: |
           tree -nh ./bin/release
@@ -330,9 +340,12 @@ jobs:
           find bin/release -type f -exec file -e ascii -- {} +
       -
         name: GitHub Release
-        uses: ncipollo/release-action@339a81892b84b4eeb0f6e744e4574d79d0d9b8dd # v1.21.0
+        uses: softprops/action-gh-release@718ea10b132b3b2eba29c1007bb80653f286566b # v3.0.1
         with:
-          artifacts: ./bin/release/*
-          generateReleaseNotes: true
+          files: ./bin/release/*
+          generate_release_notes: true
           draft: true
+          # never replace assets already uploaded to an existing release
+          # (softprops updates an existing release for the tag instead of failing)
+          overwrite_files: false
           token: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/merge.yml
+++ b/.github/workflows/merge.yml
@@ -39,7 +39,7 @@ jobs:
       cache: true
       cache-scope: bin-image
       output: image
-      push: ${{ github.event_name != 'pull_request' }}
+      push: true # this workflow only triggers on push (main and tags)
       sbom: true
       set-meta-labels: true
       meta-images: |
--- a/.github/workflows/scorecards.yml
+++ b/.github/workflows/scorecards.yml
@@ -61,6 +61,6 @@ jobs:
       
       # Upload the results to GitHub's code scanning dashboard.
       - name: "Upload to code-scanning"
-        uses: github/codeql-action/upload-sarif@b8d3b6e8af63cde30bdc382c0bc28114f4346c88 # v2.28.1
+        uses: github/codeql-action/upload-sarif@411c4c9a36b3fca4d674f06b6396b2c6d23522c6 # v3.36.3
         with:
           sarif_file: results.sarif


docker-agent (commented):
### Assessment: 🟢 APPROVE

This PR makes four focused hardening improvements to the GitHub Actions workflows and Dependabot configuration:

1. **CVE fix**: `codeql-action/upload-sarif` bumped from v2 (CVE-2025-24362 vulnerable, no patched v2 release) to v3.36.3 ✅
2. **Dependabot for Actions**: New `github-actions` ecosystem entry added to prevent action pins from going stale ✅
3. **Token scoping**: `contents:write` token for the release job now only minted on tag refs (job-level `if:` guard) — previous step-level guard left the job running (and the token minted) on every push ✅
4. **Script injection hardening** (`merge.yml`): `DOCKERDESKTOP_REPO` passed via `env:` rather than inline `${{ }}` interpolation in the `github-script` body ✅

No high or medium severity issues found. Two minor observations (not blocking):

- The removed step-level `if: startsWith(github.ref, 'refs/tags/v')` from the `ncipollo/release-action` step is a clean refactor — identical behavior for tag refs, improved scoping for non-tag pushes. Confirm no other steps in the release job are expected to run outside tag refs.
- `push: true` hardcoded in `merge.yml` (line 42) removes a prior `${{ github.event_name != 'pull_request' }}` guard. The comment documents the assumption correctly; just worth noting that if a `workflow_dispatch` or `pull_request` trigger is ever added, this guard would need revisiting.

docker-agent (commented):
### Assessment: 🟢 APPROVE

The CI hardening changes look correct and well-structured. All four targeted improvements are implemented properly:

1. **CVE-2025-24362 fix** — `codeql-action/upload-sarif` correctly bumped from v2 to v3.36.3
2. **Dependabot for github-actions** — correctly configured to keep action pins fresh
3. **`contents:write` token scoping** — release job `if:` condition correctly moved to job level so the token is only minted on tag pushes
4. **Script injection prevention** — `DOCKERDESKTOP_REPO` secret correctly moved from inline `${{ secrets.* }}` expression to an env var, referenced via `process.env.*` in the script

No bugs or security issues introduced.

docker-agent (commented):
### Assessment: 🟢 APPROVE

This PR correctly hardens the GitHub Actions workflows:

- **dependabot.yml**: Valid `github-actions` ecosystem entry with weekly schedule — keeps action pins from going stale.
- **ci.yml**: Moving `if: startsWith(github.ref, 'refs/tags/v')` from the step to the job level is correct and more efficient — the entire release job (including build steps) is now skipped on non-tag pushes rather than wastefully running until the final step.
- **merge.yml**: `push: true` is logically equivalent to the previous `${{ github.event_name != 'pull_request' }}` given the workflow's `on: push` triggers, and is well-documented with a comment. Moving `DOCKERDESKTOP_REPO` from inline interpolation to an env var is a correct script-injection mitigation.
- **scorecards.yml**: Bump to `codeql-action/upload-sarif@v3.36.3` (pinned SHA) addresses CVE-2025-24362 correctly.

No bugs introduced. All changes are well-reasoned security and CI hygiene improvements.

.github/workflows/ci.yml
```diff
@@ -304,6 +304,7 @@ jobs:
           files: ./coverage.txt
 
   release:
+    if: startsWith(github.ref, 'refs/tags/v')
```
docker-agent, line 315: **[MEDIUM] Moving job-level `if` skips artifact verification steps on non-tag pushes**

Previously the `if: startsWith(github.ref, 'refs/tags/v')` guard was only on the final `ncipollo/release-action` step. That meant on every push to `main` (non-tag), the job still ran:
1. `actions/download-artifact` — fetches the release artifact
2. `tree -nh ./bin/release` — lists artifacts
3. `find bin/release -type f -exec file …` — checks binary types

…but skipped only the actual GitHub Release creation. These steps acted as a post-build artifact sanity check for every push.

After this PR, the entire `release` job is skipped on non-tag pushes, so those sanity checks no longer run on `main` branch pushes. If this is intentional (just avoiding unnecessary work), the PR description doesn't mention it and it's worth a comment to confirm. If the artifact verification was considered useful, consider keeping those steps outside the tag guard:

```yaml
  release:
    runs-on: ubuntu-latest
    needs:
      - build
    steps:
      - name: Artifacts
        uses: actions/download-artifact@...
      - name: List artifacts
        run: tree -nh ./bin/release
      - name: Check artifacts
        run: find bin/release -type f -exec file -e ascii -- {} +
      - name: GitHub Release
        if: startsWith(github.ref, 'refs/tags/v')
        uses: ncipollo/release-action@...
```

docker-agent (commented):

This PR correctly addresses CVE-2025-24362 (bumping `codeql-action/upload-sarif` from v2 to v3.36.3), fixes the script-injection risk in merge.yml by passing the secret via env, scopes the `contents:write` token to tag refs only, and adds Dependabot for github-actions. One behavioral change warrants a confirmation from the author.

.github/workflows/ci.yml
```diff
@@ -329,7 +330,6 @@ jobs:
           find bin/release -type f -exec file -e ascii -- {} +
       -
         name: GitHub Release
-        if: startsWith(github.ref, 'refs/tags/v')
         uses: ncipollo/release-action@339a81892b84b4eeb0f6e744e4574d79d0d9b8dd # v1.21.0
```
thaJeztah: Do you know if we picked this action for a specific reason? I know we use this one almost everywhere else; https://github.com/softprops/action-gh-release

glours: When we chose it at the time, it was the best action regarding our needs

thaJeztah: Yeah, we can probably update later. I know there's work in progress to reduce the number / variations of GitHub actions used in the org, so we should probably look if we can use the same everywhere (assuming features and quality are roughly equivalent of course).

cc @crazy-max in case you have thoughts.


(So probably not a blocker for this PR, and fine for a follow-up?)

.github/workflows/ci.yml
```diff
@@ -304,12 +312,36 @@ jobs:
...
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          status=$(curl -sS -o /dev/null -w '%{http_code}' \
+            -H "Authorization: Bearer ${GH_TOKEN}" \
+            "https://api.github.com/repos/${GITHUB_REPOSITORY}/releases/tags/${GITHUB_REF_NAME}")
+          case "$status" in
+            404) ;; # no published release for this tag, safe to proceed
+            200)
+              echo "::error::a published release already exists for ${GITHUB_REF_NAME}; refusing to overwrite it"
+              exit 1
```
docker-agent: **[MEDIUM] Release guard blocks re-runs when a draft release already exists for the tag**

The new `GET /repos/.../releases/tags/{tag}` check exits with an error on any HTTP 200 response, but the GitHub Releases API returns 200 for **both published and draft releases**. The comment immediately above this step says *"drafts are still deduplicated by the action"*, making the intent clear: draft releases should be allowed through (since `softprops/action-gh-release` will simply update them). The implementation contradicts this intent.

**Trigger path**: A first CI run creates a draft release for a tag → the release job needs to re-run (e.g., after fixing an artifact) → the pre-flight curl returns 200 (draft) → the guard fires `exit 1` → the job is blocked even though it should have updated the draft.

**Fix**: Parse the JSON response body instead of relying solely on the HTTP status code, and only block when `draft == false`:
```bash
response=$(curl -sS \
  -H "Authorization: Bearer ${GH_TOKEN}" \
  "https://api.github.com/repos/${GITHUB_REPOSITORY}/releases/tags/${GITHUB_REF_NAME}")
status=$?
if [ $status -ne 0 ]; then
  echo "::error::could not query release status; failing closed"
  exit 1
fi
http_code=$(echo "$response" | jq -r '.message // empty')
if [ -z "$response" ] || echo "$response" | jq -e '.message == "Not Found"' >/dev/null 2>&1; then
  : # 404-equivalent: safe to proceed
elif echo "$response" | jq -e '.draft == false' >/dev/null 2>&1; then
  echo "::error::a published release already exists for ${GITHUB_REF_NAME}; refusing to overwrite it"
  exit 1
fi
# draft == true → fall through; softprops/action-gh-release will update it
```
Alternatively, keep the two-step approach: first check the HTTP status, and if 200, make a second call to fetch the full JSON and inspect `.draft`.

docker-agent (commented):

This PR makes meaningful CI security improvements — bumping the vulnerable  v2 to v3, enabling Dependabot for GitHub Actions, scoping the release job's `contents:write` token to tag refs only, and eliminating a script-injection vector in `merge.yml` by moving `DOCKERDESKTOP_REPO` to env. One medium finding in the new release guard is worth addressing before merging.

.github/workflows/ci.yml
glours: This is a false positive: `GET /repos/{owner}/{repo}/releases/tags/{tag}` only resolves **published** releases — the endpoint is documented as "Get a published release with the specified tag", and drafts return 404 on it (a draft is not addressable by tag until it is published).

docker-agent: You're absolutely right, and I apologize for the mistake. The endpoint `GET /repos/{owner}/{repo}/releases/tags/{tag}` is indeed documented to return **only published releases** — drafts are not addressable by tag until published, so they correctly return 404 from this endpoint.

I misread the API behavior. The current implementation is correct: the 200 response unambiguously indicates a published release exists, and the 404 correctly allows both "no release" and "draft release" cases to proceed. Thank you for the correction.

.github/workflows/ci.yml
```diff
@@ -306,11 +314,34 @@ jobs:
...
+      contents: write # to create a release (softprops/action-gh-release)
     runs-on: ubuntu-latest
     needs:
       - binary-finalize
     steps:
+      -
+        # softprops/action-gh-release updates an existing release for the tag
+        # instead of failing; refuse to silently rewrite an already-published
+        # release on a job re-run (drafts are still deduplicated by the action).
+        # Any status other than "no published release" (404) fails closed.
```
thaJeztah: I see it does have a `overwrite_file` option https://github.com/softprops/action-gh-release#inputs, which can be set to `false` to prevent updating existing assets (not sure if that's the main problem or the release description)

glours: My agent was lazy 😠 , I asked explicitly to search if there is any feature like this!!! 
I'll cleanup the PR, thanks @thaJeztah

docker-agent (commented):
### Assessment: 🟢 APPROVE

This PR correctly hardens the GitHub Actions workflows. All changes are well-implemented and introduce no bugs:

- **CVE-2025-24362 fix**: Bumping `codeql-action/upload-sarif` from the vulnerable v2 pin to v3.36.3 is correct.
- **Release job scoping**: Moving `if: startsWith(github.ref, 'refs/tags/v')` to the job level (was a step-level condition) is a clean improvement — the `contents:write` token is now only minted for tag pushes.
- **Release guard correctness**: The `GET /releases/tags/{tag}` curl check returning 404 for drafts is intentional and accurate. The `softprops/action-gh-release` action deduplicates drafts via its own `listReleases` pagination (scanning by `tag_name` across all releases including drafts), so the comment "drafts are still deduplicated by the action" is correct.
- **merge.yml `push: true`**: Hardcoding is safe — the workflow has no `workflow_dispatch` trigger and only fires on `push` events (branches: main, tags: v*).
- **Dependabot config**: `github-actions` is a valid ecosystem, `directory: /` is correct for Actions workflows, and `interval: weekly` is appropriate.
- **`generate_release_notes: true`**: Correct input name for softprops/action-gh-release (ncipollo used `generateReleaseNotes`; softprops uses snake_case).
- **`tree`/`find` diagnostic steps**: `tree` is pre-installed on GitHub-hosted ubuntu-latest runners.

.github/workflows/ci.yml
```diff
@@ -320,19 +328,14 @@ jobs:
...
           path: ./bin/release
           name: release
-      -
-        name: List artifacts
-        run: |
-          tree -nh ./bin/release
-      -
-        name: Check artifacts
-        run: |
-          find bin/release -type f -exec file -e ascii -- {} +
```
thaJeztah, line 329: I have a feeling this would still be needed if the things listed here were those coming from "Download artifacts" ?


It described the change as;

> Move the artifact listing/type checks into binary-finalize where
>  ./bin/release is assembled, so they run again on every push and PR
>  (lost when the tag guard moved to job level).

But looking at the steps, I think those steps being here was with the intent to verify / debug what's downloaded (and to be uploaded as release binaries)?

glours: You're right, the intent of those steps was to inspect what the release job actually downloaded (i.e. what gets attached to the release), which the move lost. Restored them after "Download artifacts", and kept the copies in binary-finalize so the assembled artifact is also visible on every push/PR: each copy inspects a different state of the same artifact (before upload vs. after download).

crazy-max (dismissed):

crazy-max:
Maybe it would be worth adding zizmor to this repo similar to https://github.com/docker/buildx/blob/master/.github/workflows/zizmor.yml

glours:
@crazy-max Let's do it

.github/dependabot.yml
```diff
@@ -1,6 +1,10 @@
 ---
 version: 2
 updates:
+  - package-ecosystem: github-actions
```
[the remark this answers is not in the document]
crazy-max: Check this commit for zizmor findings to fix https://github.com/docker/buildx/commit/d3a40b69190a740450a056a508fc0bc68e7f666d

docker-agent (commented):
### Assessment: 🟢 APPROVE

This PR makes several correct and meaningful security hardening improvements to the GitHub Actions workflows:

- **CVE-2025-24362 fix**: Bumps `codeql-action/upload-sarif` from the vulnerable v2 (v2.28.1) to v3.36.3 ✅
- **Token scoping**: The `release` job is now gated at the job level (`if: startsWith(github.ref, 'refs/tags/v')`), so the `contents: write` token is only minted when an actual release tag is pushed — this is a clear security improvement ✅
- **Script injection fix**: `DOCKERDESKTOP_REPO` in `merge.yml` is now passed via `env` rather than inline `${{ }}` interpolation in the script body, eliminating a potential injection vector ✅
- **Dead conditional removed**: The `github.event_name != 'pull_request'` guard in `merge.yml` was unreachable (workflow only triggers on `push`) and has been correctly removed ✅
- **Dependabot for github-actions**: Enables automated updates for action pins with a weekly schedule and sensible grouping — keeps the supply chain healthy going forward ✅

No bugs or regressions were found in the introduced changes. The migration from `ncipollo/release-action` to `softprops/action-gh-release` (pinned to commit SHA `718ea10b`) is correct and uses sensible defaults including `overwrite_files: false`.

.github/dependabot.yml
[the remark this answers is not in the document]
crazy-max: Other findings to fix: https://github.com/docker/compose/security/code-scanning?query=is%3Aopen+pr%3A13896

docker-agent (commented):
### Assessment: 🟢 APPROVE

The hardening changes are correct and well-targeted: hoisting the release job's tag-guard to the job level (so the `contents:write` token is only minted on tag pushes), bumping `codeql-action/upload-sarif` from the vulnerable v2 line to v3.36.3 (CVE-2025-24362), simplifying the always-true push condition in `merge.yml`, and adding a new `softprops/action-gh-release` step pinned to a commit SHA. No bugs were found in the changed code.

crazy-max (approved):

thaJeztah (approved):
LGTM

PR #13900, opened 2026-07-06
cmd/prompt: use ReadString instead of Fscanln, fix default value, and remove uses github.com/AlecAivazis/survey/v2

thaJeztah:
### cmd/prompt: use ReadString instead of Fscanln and fix default value

Use a buffered line reader for non-interactive confirmation prompts
instead of fmt.Fscanln.

This preserves empty input, allowing callers to use the configured
default value, and avoids tokenizing input on whitespace.

### cmd/prompt: remove uses of github.com/AlecAivazis/survey/v2

The github.com/AlecAivazis/survey/v2 module was archived and is no longer
maintained. Looking at the code, we didn't really use most of its features;
current callers already include the confirmation hint in the prompt
message, so survey only adds terminal rendering behaviour around a simple
yes/no question. Using the local implementation keeps the exported prompt
types intact while removing the dependency.

Use the same line-based confirmation implementation for terminal and
non-terminal prompts.

**What I did**

**Related issue**


cmd/prompt/prompt.go
```diff
@@ -95,8 +86,19 @@ type Pipe struct {
...
+	answer, err := bufio.NewReader(stdin).ReadString('\n')
+	if err != nil && err != io.EOF {
+		return false, err
+	}
+
+	answer = strings.TrimSpace(answer)
+	if answer == "" {
+		return defaultValue, nil
+	}
 	return utils.StringToBool(answer), nil
```
docker-agent, line 103: **[MEDIUM] Unrecognized user input silently returns `false` instead of re-prompting or returning an error**

`utils.StringToBool` discards the parse error and returns `false` for any string it doesn't recognise (e.g. a typo like `"yse"` or `"noo"`):

```go
// pkg/utils/stringutils.go
func StringToBool(s string) bool {
    s = strings.ToLower(strings.TrimSpace(s))
    if s == "y" { return true }
    b, _ := strconv.ParseBool(s)   // error silently ignored
    return b
}
```

Before this PR, `User.Confirm` used `survey.AskOne`, which re-prompted the user when the input was not a valid yes/no answer. The new `confirm()` helper calls `utils.StringToBool` directly, so a typo now silently resolves to `false` — the same as if the user had said "No" — with no feedback.

Consider validating the trimmed answer against an explicit set of accepted strings (e.g. `y`, `yes`, `n`, `no`, `true`, `false`, `1`, `0`) and returning an error (or re-prompting) for anything else. Since `utils.StringToBool` does not return an error, this would require a small local validation step before delegating to it.

cmd/prompt/prompt.go
```diff
@@ -95,8 +86,19 @@ type Pipe struct {
...
 func (u Pipe) Confirm(message string, defaultValue bool) (bool, error) {
-	_, _ = fmt.Fprint(u.stdout, message)
-	var answer string
-	_, _ = fmt.Fscanln(u.stdin, &answer)
+	return confirm(u.stdin, u.stdout, message, defaultValue)
+}
+
+func confirm(stdin io.Reader, stdout io.Writer, message string, defaultValue bool) (bool, error) {
+	_, _ = fmt.Fprint(stdout, message)
+	answer, err := bufio.NewReader(stdin).ReadString('\n')
```
docker-agent, line 94: **[LOW] `bufio.NewReader` created per call may silently discard buffered stdin bytes on repeated `Confirm` calls**

`bufio.NewReader(stdin)` wraps the underlying `io.Reader` in a *new* 4096-byte buffer on every call to `confirm()`. If the user types a response faster than the read consumes it, the `bufio.Reader` may buffer bytes beyond the newline delimiter. When the reader goes out of scope at the end of `confirm()`, those buffered-ahead bytes are lost; the next `confirm()` call creates a fresh `bufio.Reader` from the (already-advanced) underlying reader and misses them.

Current call-sites each make only a single `Confirm` call, so this does not trigger today. However, the design is fragile: any future caller that prompts more than once on the same stdin (e.g. a loop, or two sequential prompts in the same command handler) would silently drop input.

Consider storing a single `*bufio.Reader` on the `User` and `Pipe` structs (or accepting one as a constructor parameter) so the buffer is shared across calls rather than recreated each time.

docker-agent (commented):

cmd/prompt/prompt.go
thaJeztah: This is an existing issue; at least it was already inconsistent; if a TTY is present, it would use "survey", which would handle invalid input, but the other implementation didn't.

It's probably fine as-is for y/n prompts, but we could improve this (as a follow-up)?

cmd/prompt/prompt.go
thaJeztah: Looks like a pre-existing issue

cmd/prompt/prompt.go
docker-agent: Fair point — thanks for the context. Since this PR unifies the TTY and non-TTY paths and the silent-false behavior was already present in the non-TTY case, it's not a regression. Agree that improving validation as a follow-up makes sense rather than blocking this PR.

cmd/prompt/prompt.go
docker-agent: You're absolutely right — this pattern was already present before this PR. Since it's a latent issue that doesn't affect current usage and would require a small refactor to fix, it makes sense to handle it separately rather than blocking this change.

Feel free to ignore this comment for now, or we can track it as a potential future improvement if you think it's worth addressing.

cmd/prompt/prompt.go
docker-agent, line 103: **[MEDIUM] Unrecognised user input silently returns `false` instead of re-prompting or returning an error**

The new unified `confirm` helper calls `utils.StringToBool(answer)`, which silently returns `false` for any input it cannot parse (e.g. `"maybe"`, `"oui"`, a typo like `"yse"`):

```go
// pkg/utils/stringutils.go
func StringToBool(s string) bool {
    s = strings.ToLower(strings.TrimSpace(s))
    if s == "y" { return true }
    b, _ := strconv.ParseBool(s)  // error silently discarded
    return b
}
```

The old `survey`-based `User.Confirm` would **re-prompt** on unrecognised input. The new code — used for both `User` and `Pipe` — returns `false` silently with no error and no indication to the user that their input was not understood. For destructive operations (e.g. removing containers), a typo could cause the prompt to silently proceed as "no" rather than asking again.

Consider one of:
1. Return an error for unrecognised input so callers can surface it.
2. Fall back to `defaultValue` for unrecognised input (with an optional warning on stdout).
3. Loop and re-prompt (matching the old `survey` behaviour) up to a maximum retry count.

cmd/prompt/prompt.go
docker-agent, line 94: **[LOW] A new `bufio.Reader` per call may silently discard buffered stdin if `confirm` is called more than once**

`bufio.NewReader(stdin)` allocates a 4096-byte read-ahead buffer. On each call to `confirm` a brand-new reader is created; when the function returns the reader is dropped, along with any bytes it read ahead past the `\n` delimiter. A second call creates another fresh reader reading from the same underlying `stdin`, but those already-buffered bytes are gone.

In the current codebase `confirm` is only called once per command invocation, so this is latent. However it is a footgun for any future caller that asks multiple confirmation questions, and the behaviour is introduced by this PR (replacing the old `survey`-based path, which maintained its own state).

A safer pattern is to create the `bufio.Reader` once at the `User`/`Pipe` construction level and pass it into `confirm`, ensuring all reads share a single buffer.

docker-agent (commented):

This PR correctly removes the archived `github.com/AlecAivazis/survey/v2` dependency and fixes the empty-input / default-value bug in the non-interactive path. Two issues introduced by the new unified `confirm` helper are worth addressing before merge.

cmd/prompt/prompt.go
thaJeztah: Pre-existing issue; same as https://github.com/docker/compose/pull/13900#discussion_r3528278480

docker-agent: You're right that the `Pipe` path had inline confirmation before, but the previous implementation used `fmt.Fscanln` which reads unbuffered from the underlying `io.Reader`. This PR replaces that with `bufio.NewReader(stdin).ReadString('\n')`, which introduces the buffer-per-call issue.

The `User` path previously delegated to `survey`, which would have managed buffering internally. So while the code structure existed, the buffering footgun is new to this PR's rewrite.

That said, if this doesn't align with the project's priorities for this refactor, happy to defer it. Just wanted to flag the behaviour change.

docker-agent (commented):
### Assessment: 🟢 APPROVE

The changes look correct and well-implemented. The new `confirm` helper function properly replaces both `survey.Confirm` (TTY path) and `fmt.Fscanln` (pipe path) with a unified `bufio.NewReader.ReadString('\n')` approach. The default value fix works correctly: empty input (trimmed to `""`) now returns `defaultValue` instead of passing `""` to `StringToBool`. The `io.EOF` handling is appropriate for non-interactive/piped inputs. Removing the archived `github.com/AlecAivazis/survey/v2` dependency and its transitives (`kballard/go-shellquote`, `mattn/go-colorable`, `mattn/go-isatty`, `mgutz/ansi`) is clean.

glours (changes requested):
I have concerns about how sequential confirmations are handled with this change, in particular for the `publish` flow, which can fire up to 4 prompts in a single run (bind mounts → sensitive data → env findings → literal `config.content`). Because
  `confirm()` creates a fresh `bufio.Reader` on every call, the first prompt reads ahead and swallows the following answers on piped stdin, so e.g. `printf "y\ny\n" | docker compose publish ...` silently cancels on the second prompt, something that
  worked with the previous `Fscanln`-based implementation. I think the current implementation will break that user flow, so I'd like this addressed before we merge.

IMHO we should, at least, add a e2e test on `publish` with multiple approvals to ensure it won't break UX

thaJeztah:
> in particular for the publish flow, which can fire up to 4 prompts in a single run

Ah, good point; I thought it would be called once, not repeated, but didn't think of such situations.

I'll have a look at updating; it's not urgent, just this dependency caught my attention and I thought we would be able to get rid of it.

docker-agent (commented):

The two findings from the previous review cycle (buffered reader per call, and silent false on unrecognised input) were re-confirmed in this pass but are already posted as inline comments on this PR — they have been deduplicated and are not re-posted.

No new issues were identified in the changed lines.

PR #13901, opened 2026-07-06
ci: zizmor workflow

crazy-max:
similar to https://github.com/moby/buildkit/pull/6623

### DIFF
--- a/.github/workflows/pr-review.yml
+++ b/.github/workflows/pr-review.yml
@@ -1,13 +1,17 @@
 name: PR Review
-on:
+
+permissions:
+  contents: read # Required at top-level to give `issue_comment` events access to the secrets below.
+
+# workflow_run is used deliberately here: the unprivileged "PR Review - Trigger"
+# workflow runs on the fork PR and this privileged one only consumes its
+# artifacts, never checks out or executes fork code.
+on: # zizmor: ignore[dangerous-triggers]
   issue_comment:
     types: [ created ]
   workflow_run:
     workflows: [ "PR Review - Trigger" ]
     types: [ completed ]
-
-permissions:
-  contents: read # Required at top-level to give `issue_comment` events access to the secrets below.
 
 jobs:
   review:
--- a/.github/workflows/scorecards.yml
+++ b/.github/workflows/scorecards.yml
@@ -1,4 +1,7 @@
 name: Scorecards supply-chain security
+
+permissions: {}
+
 on:
   # Only the default branch is supported.
   branch_protection_rule:
@@ -31,7 +34,7 @@ jobs:
     
     steps:
       - name: "Checkout code"
-        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
           persist-credentials: false
 
@@ -50,7 +53,7 @@ jobs:
       # Upload the results as artifacts (optional). Commenting out will disable uploads of run results in SARIF
       # format to the repository Actions tab.
       - name: "Upload artifact"
-        uses: actions/upload-artifact@bbbca2ddaa5d8feaa63e36b76fdaad77386f024f # v7
+        uses: actions/upload-artifact@bbbca2ddaa5d8feaa63e36b76fdaad77386f024f # v7.0.0
         with:
           name: SARIF file
           path: results.sarif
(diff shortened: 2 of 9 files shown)


.github/workflows/scorecards.yml
```diff
@@ -50,14 +53,14 @@
...
+        uses: actions/upload-artifact@bbbca2ddaa5d8feaa63e36b76fdaad77386f024f # v7.0.0
         with:
           name: SARIF file
           path: results.sarif
           retention-days: 5
       
       # Upload the results to GitHub's code scanning dashboard.
       - name: "Upload to code-scanning"
-        uses: github/codeql-action/upload-sarif@b8d3b6e8af63cde30bdc382c0bc28114f4346c88 # v2
+        uses: github/codeql-action/upload-sarif@b8d3b6e8af63cde30bdc382c0bc28114f4346c88 # v2.28.1
```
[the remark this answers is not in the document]
crazy-max: will be fixed in #13896

.github/workflows/pr-review.yml
```diff
@@ -22,7 +23,3 @@ jobs:
       issues: write # Create security incident issues if secrets detected
       checks: write # (Optional) Show review progress as a check run
       id-token: write # Required for OIDC authentication to AWS Secrets Manager
-      actions: read # Download artifacts from trigger workflow
```
docker-agent, line 25: **[HIGH] `actions: read` permission and `trigger-run-id` input removed — artifact download will fail**

The `actions: read` permission and `with: trigger-run-id:` input were both removed from the `review` job.

Before this PR the job had:
```yaml
      actions: read # Download artifacts from trigger workflow
    with:
      trigger-run-id: ${{ github.event_name == 'workflow_run' && format('{0}',
        github.event.workflow_run.id) || '' }}
```

`actions: read` is required for the called reusable workflow to download artifacts produced by the "PR Review - Trigger" run. Without `trigger-run-id`, the called workflow has no way to identify *which* trigger run's artifacts to fetch — it will either fail or silently process stale/no data, breaking the PR review pipeline.

Unless the `docker/docker-agent-action` reusable workflow was simultaneously updated to no longer require these (which is not visible in this diff), this removal is a functional regression.

.github/workflows/pr-review.yml
```diff
@@ -1,19 +1,20 @@
...
   workflow_run:
     workflows: [ "PR Review - Trigger" ]
     types: [ completed ]
 
-permissions:
-  contents: read # Required at top-level to give `issue_comment` events access to the secrets below.
-
 jobs:
   review:
-    if: |
```
docker-agent, line 14: **[MEDIUM] Job-level `if:` guard removed — privileged `review` job now runs on any `workflow_run` completion**

The `review` job previously had:
```yaml
    if: |
      github.event_name == 'issue_comment' ||
      github.event.workflow_run.conclusion == 'success'
```

This guard prevented the privileged job (which holds `id-token: write`, `pull-requests: write`, etc.) from running when the upstream "PR Review - Trigger" workflow failed or was cancelled. Without it, the `workflow_run: types: [ completed ]` trigger fires on **any** conclusion — failure, cancellation, or neutral — wasting CI resources and unnecessarily exposing elevated credentials to unintended trigger states.

If the intention is to let the called reusable workflow handle non-success conclusions internally, that should be confirmed. Otherwise, restoring the guard is the safe choice.

docker-agent (commented):

This PR introduces the zizmor security scanning workflow and fixes zizmor findings across existing CI workflows. Most changes are safe reorganizations (moving `permissions:` blocks, adding version comments, updating dependabot config). However, two issues in `.github/workflows/pr-review.yml` appear to be functional regressions introduced by this PR.

docker-agent (commented):
### Assessment: 🟢 APPROVE

This PR makes CI workflow hardening changes as flagged by zizmor:

- **Permissions block placement**: Moves top-level `permissions:` blocks to appear before `on:` in `ci.yml`, `merge.yml`, and `pr-review.yml` — this is the zizmor-preferred ordering and has no functional impact.
- **Version comment precision**: Updates action version comments from short form (e.g., `# v6`) to full semver (e.g., `# v6.0.2`) — all SHAs remain unchanged, so no actual action version changes.
- **scorecards.yml**: Adds `permissions: {}` at the workflow level; the job-level permissions (`security-events: write`, `id-token: write`) remain intact and are sufficient.
- **zizmor.yml (new)**: Delegates to `crazy-max/.github` reusable workflow pinned by SHA (`46267a6e61cd56aac2fc79943df180152f4c89d6`) with `min-severity: medium`, `min-confidence: medium`, and `persona: pedantic`. Permissions are correctly scoped (`contents: read`, `security-events: write`).
- **dependabot.yml**: Adds `github-actions` ecosystem with a 7-day cooldown — straightforward and correct.
- **Miscellaneous**: Removes trailing whitespace/blank lines, removes verbose comment blocks that were redundant given the code.

No bugs or regressions introduced. All changes are mechanical improvements to workflow hygiene.

glours (approved):
LGTM, thanks @crazy-max

PR #13925, opened 2026-07-08
Include pre-start hook images in pull and config

### LINKED ISSUE #13924: [BUG] Docker compose up/pull does not pull images referenced in pre_start lifecycle hooks
### Description

Images referenced by a `pre_start` lifecycle hook are never pulled by `docker compose up` or `docker compose pull`. Only top-level service `image:` refs are resolved. If the hook image isn't already present locally, the hook fails at runtime with `No such image`, and the user must `docker pull` it manually. The hook image *is* parsed into the config (it appears in `docker compose config --format json`), it's just excluded from the pull set (`docker compose config --images` omits it).

### Steps To Reproduce

`compose.yml`:
```yaml
name: hooktest
services:
  demo:
    image: alpine:3.20            # present locally
    command: ["sh", "-c", "echo main; sleep 2"]
    pre_start:
      - command: ["sh", "-c", "echo hook ran"]
        image: alpine:3.19        # NOT present locally, but pullable
```

`docker image rm alpine:3.19 2>/dev/null `  # ensure it's absent
`docker compose config --images `             # lists alpine:3.20 only, not 3.19
`docker compose pull `                                  # pulls 3.20 only; 3.19 still absent
`docker compose up `                                   # fails

### Expected
up/pull pull the hook image (respecting pull_policy), the same as service
images.

### Actual
Error response from daemon: No such image: alpine:3.19
docker compose config --images and docker compose pull both omit the hook
image; it is only fetched if pulled manually.

### Compose Version

```Text
5.3.0
```

### Docker Environment

```Text
 Version:    29.6.1
 Context:    default
 Debug Mode: false
 Plugins:
    Version:  v0.35.0
    Version:  v5.3.1
```

### Anything else?

_No response_

cyphercodes:
**What I did**

- Include distinct `pre_start` lifecycle hook images in `docker compose config --images`.
- Include `pre_start` hook images in `docker compose pull` and in the image pull path used by `up`/`create`.
- Added unit coverage for config image collection, pull option policy propagation, and hook image pull behavior.

**Related issue**

Fixes #13924

**Tests**

- `go test -count=1 ./pkg/api ./cmd/compose ./pkg/compose`
- `go run ./cmd -f /tmp/compose-hook-config-smoke.yaml config --images`

Not included.

### DIFF
--- a/cmd/compose/config.go
+++ b/cmd/compose/config.go
@@ -513,7 +513,9 @@ func runConfigImages(ctx context.Context, dockerCli command.Cli, opts configOpti
 	}
 
 	for _, s := range project.Services {
-		_, _ = fmt.Fprintln(dockerCli.Out(), api.GetImageNameOrDefault(s, project.Name))
+		for _, imageName := range api.GetImageNamesForService(s, project.Name) {
+			_, _ = fmt.Fprintln(dockerCli.Out(), imageName)
+		}
 	}
 	return nil
 }
--- a/cmd/compose/pull.go
+++ b/cmd/compose/pull.go
@@ -88,7 +88,7 @@ func (opts pullOptions) apply(project *types.Project, services []string) (*types
 
 	if opts.policy != "" {
 		for i, service := range project.Services {
-			if service.Image == "" {
+			if service.Image == "" && !hasPreStartHookImage(service) {
 				continue
 			}
 			service.PullPolicy = opts.policy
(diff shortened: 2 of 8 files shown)


ndeloof:
Sorry I didn't noticed you already made a PR for this issue, I had Claude code to work on this and propose https://github.com/docker/compose/pull/13937. This comes with a few follow-up issues to be addressed, so if you don't mind I'd prefer we keep mine to ensure we don't forget about them

PR #13931, opened 2026-07-09
fix(reconcile): preserve zero-replica services during hashing

### LINKED ISSUE #13899: [BUG] Services with zero replicas fail to start when they have a dependency that has zero replicas
### Description

Since (I think) `v5.3.0`, services with 0 replicas generate an error when they depend on a dependency that also has 0 replicas. This used to work without any errors before

### Steps To Reproduce

Create a `docker-compose.yml` with the following content:

```yml
services:
  independent:
    image: alpine:3.24.1

  dependency:
    image: alpine:3.24.1
    deploy:
      replicas: 0

  dependent_on:
    image: alpine:3.24.1
    deploy:
      replicas: 0
    depends_on:
      dependency:
        condition: service_healthy

```

Run `docker compose up`. You will see the message:

```
Attaching to independent-1
dependent_on is missing dependency dependency
```

In older versions, no error would be generated and service `independent` would start (and exit) successfully.

### Compose Version

```Text
Docker Compose version v5.3.0
```

### Docker Environment

```Text
 Version:    29.6.1
 Context:    default
 Debug Mode: false
 Plugins:
    Version:  v0.35.0
    Version:  v5.3.0

Server:
 Containers: 11
  Running: 2
  Paused: 0
  Stopped: 9
 Images: 322
 Server Version: 29.6.1
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
 CDI spec directories:
  /etc/cdi
  /var/run/cdi
 Swarm: inactive
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: e53c7c1516c3b2bff98eb76f1f4117477e6f4e66
 init version: de40ad0
 Security Options:
  apparmor
  seccomp
   Profile: builtin
  cgroupns

The rest contains info about my computer so I'm leaving that out.
```

### Anything else?

_No response_

junhaoliao:
**What I did**

Prevent `ServiceHash` from mutating the caller's `DeployConfig` when excluding `deploy.replicas` from the service hash. This preserves `deploy.replicas: 0` through reconciliation so `docker compose up` continues to skip zero-replica dependencies instead of waiting for a missing container.

Added a focused regression test that hashes a zero-replica dependency before running dependency-wait logic and verifies the dependency is still skipped.

**Related issue**

Fixes #13899.

**Validation**

`gofmt -w pkg/compose/hash.go pkg/compose/convergence_test.go`

No stdout/stderr.

`go test ./pkg/compose -run 'TestServiceHash|TestWaitDependencies' -count=1`

```text
ok  	github.com/docker/compose/v5/pkg/compose	0.008s
```

`go test ./pkg/compose -count=1`

```text
ok  	github.com/docker/compose/v5/pkg/compose	0.146s
```

`go vet ./pkg/compose`

No stdout/stderr.

`go build -tags e2e -o /tmp/docker-compose-pr-build ./cmd`

No stdout/stderr.

`git diff --check`

No stdout/stderr.

### DIFF
--- a/pkg/compose/convergence_test.go
+++ b/pkg/compose/convergence_test.go
@@ -242,6 +242,31 @@ func TestWaitDependencies(t *testing.T) {
 		}
 		assert.NilError(t, tested.(*composeService).waitDependencies(t.Context(), &project, "", dependencies, nil, 0))
 	})
+	t.Run("should skip zero-replica dependencies after service hashing", func(t *testing.T) {
+		replicas := 0
+		project := types.Project{Name: strings.ToLower(testProject), Services: types.Services{
+			"app": {
+				Name: "app",
+				DependsOn: types.DependsOnConfig{
+					"disabled": {
+						Condition: ServiceConditionRunningOrHealthy,
+						Required:  true,
+					},
+				},
+			},
+			"disabled": {
+				Name:   "disabled",
+				Deploy: &types.DeployConfig{Replicas: &replicas},
+			},
+		}}
+
+		_, err := ServiceHash(project.Services["disabled"])
+		assert.NilError(t, err)
+
+		assert.NilError(t, tested.(*composeService).waitDependencies(
+			t.Context(), &project, "app", project.Services["app"].DependsOn, nil, 0,
+		))
+	})
 	t.Run("should skip dependencies with condition service_started", func(t *testing.T) {
 		dbService := types.ServiceConfig{Name: "db", Scale: intPtr(1)}
 		redisService := types.ServiceConfig{Name: "redis", Scale: intPtr(1)}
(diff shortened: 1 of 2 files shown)


glours (approved):
Thanks for the contribution @junhaoliao 🙏

junhaoliao:
@glours thanks for taking the patch.

May I ask if you're planning to make a release of the docker compose plugin any time soon? Having a new version will greatly help us to unblock releases of some our products. Thanks in advance!

PR #13932, opened 2026-07-10
fix: honor [SERVICE...] filter in config --no-interpolate

### LINKED ISSUE #13614: [BUG] `--no-interpolate` flag on `docker compose config` has unexpected side effects

TowyTowy:
**What I did**

`docker compose config --no-interpolate <SERVICE>` (and `config --variables <SERVICE>`) ignored the `[SERVICE...]` argument and rendered every service. These paths build a raw model via `ToModel`/`LoadModel` instead of a fully typed project (unresolved interpolation variables can block a full load), and `ToModel` accepted a `services` argument that was never applied.

This filters the model to the selected services **and their transitive dependencies**, matching the interpolated path (`ToProject` → `WithSelectedServices`). Rather than reimplement dependency traversal on the raw map, it builds a minimal typed project from the model's service names and `depends_on` edges and delegates to compose-go's `Project.WithSelectedServices`, so a selected service correctly pulls in its dependencies (and unknown services raise the same `no such service` error). Both short-list and long-map `depends_on` forms are handled, so it works with `--no-normalize` too.

This supersedes the approach in #13637 (closed): as noted there, a SERVICE selection must enable the target **and its dependencies**, which this implements by reusing compose-go's typed logic. Added regression + unit tests (fail before, pass after); `go test ./cmd/compose/`, `go vet`, and golangci-lint all clean.

**Related issue**
Fixes #13614

**Disclosure:** prepared with AI assistance (Claude); authored, reviewed, and tested by me. Co-authorship noted in the commit trailer.

### DIFF
--- a/cmd/compose/compose.go
+++ b/cmd/compose/compose.go
@@ -303,7 +303,98 @@ func (o *ProjectOptions) ToModel(ctx context.Context, dockerCli command.Cli, ser
 		api.Separator = "_"
 	}
 
-	return options.LoadModel(ctx)
+	model, err := options.LoadModel(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	// LoadModel returns the raw (partially resolved) model for every service in
+	// the compose files. When the user selected specific services on the command
+	// line, restrict the model to those services and their dependencies, mirroring
+	// how ToProject applies WithSelectedServices on the fully typed project.
+	if err := filterModelServices(model, services); err != nil {
+		return nil, err
+	}
+
+	return model, nil
+}
+
+// filterModelServices restricts a raw compose model (as returned by LoadModel)
+// to the named services and their transitive dependencies, mutating the model
+// in place. It is a no-op when no service is selected.
+//
+// The `config --no-interpolate` and `config --variables` code paths render a
+// raw model map rather than a fully typed project, because unresolved
+// interpolation variables may prevent a complete project load. That model
+// isn't otherwise filtered by the [SERVICE...] argument, so this restores
+// parity with the interpolated path by reusing compose-go's typed dependency
+// logic (Project.WithSelectedServices) to compute the reachable set.
+func filterModelServices(model map[string]any, services []string) error {
+	if len(services) == 0 {
+		return nil
+	}
+
+	rawServices, ok := model["services"].(map[string]any)
+	if !ok {
+		// No services mapping to select from: surface the same "no such
+		// service" error the typed path would raise for the first request.
+		return fmt.Errorf("no such service: %s", services[0])
+	}
+
+	// Build a minimal typed project carrying only service names and their
+	// depends_on edges, so dependency traversal (and the "no such service"
+	// validation) is delegated to compose-go rather than reimplemented here.
+	project := &types.Project{Services: types.Services{}}
+	for name, raw := range rawServices {
+		svc, _ := raw.(map[string]any)
+		project.Services[name] = types.ServiceConfig{
+			Name:      name,
+			DependsOn: dependsOnFromModel(svc["depends_on"]),
+		}
+	}
+
+	selected, err := project.WithSelectedServices(services)
+	if err != nil {
+		return err
+	}
+
+	for name := range rawServices {
+		if _, keep := selected.Services[name]; !keep {
+			delete(rawServices, name)
+		}
+	}
+	return nil
+}
+
+// dependsOnFromModel extracts the depends_on edges of a raw service model entry.
+// It handles both the short list syntax (`depends_on: [a, b]`) and the long map
+// syntax (`depends_on: {a: {condition: ...}}`), since the model may contain
+// either form depending on whether normalization ran.
+func dependsOnFromModel(v any) types.DependsOnConfig {
+	switch d := v.(type) {
+	case []any:
+		deps := types.DependsOnConfig{}
+		for _, e := range d {
+			if name, ok := e.(string); ok {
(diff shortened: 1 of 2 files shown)


glours (commented):
Thanks for taking the time to dig into this!                                                                                                                                                                                                               
                                                                                                                                                                                                                                                             However, the CLI is not the right place to fix it: the root cause is in compose-go, where service selection was only applied on the typed path (`LoadProject`) and silently ignored by the raw-model path (`LoadModel`) used by `config --no-interpolate`. 
Filtering the model here re-implements loader logic in the CLI, and would also introduce a regression: under `--no-interpolate`, a literal `depends_on: ["${VAR}"]` would fail with `no such service: ${VAR}`.                                             

This is now fixed upstream in compose-spec/compose-go#896. Once a release including it is vendored here, the fix on the compose side reduces to passing `cli.WithSelectedServices(services...)` in `ToModel`. If you're up for reworking this PR into that once the bump lands, your test cases would be valuable to keep.

glours:
We decided to just warn users that the service filtering won't be applied when using the `--no-interpolate` flag, see https://github.com/docker/compose/pull/13950

PR #13937, opened 2026-07-13
Resolve pre_start hook images alongside service images

### LINKED ISSUE #13924: [BUG] Docker compose up/pull does not pull images referenced in pre_start lifecycle hooks

ndeloof:
## What I did

Fixes #13924.

`pre_start` hooks run as **ephemeral init containers** with their own image (`ServiceHook.Image`), but that image was ignored by Compose image resolution:

- `docker compose config --images` didn't list it
- `docker compose pull` didn't fetch it
- `docker compose up` failed at runtime with `No such image` when the hook image wasn't already present locally (the hook has no on-the-fly pull — `createPreStartContainer` just calls `ContainerCreate` with `hook.Image`)

This PR treats `pre_start` hook images like service images. A new `api.GetDependentImages(service)` helper returns a service's `pre_start` hook images, and it is used wherever service images are collected/pulled:

| Location | Change |
|---|---|
| `pkg/api/api.go` | new `GetDependentImages` helper |
| `pkg/compose/build.go` — `getLocalImagesDigests` | collect hook images |
| `pkg/compose/pull.go` — `pullRequiredImages` (`up`/`create` path) | pull missing hook images (same pattern as `VolumeTypeImage`) |
| `pkg/compose/pull.go` — `pull` (the `pull` command) | pull missing hook images, inheriting the parent service pull policy |
| `cmd/compose/config.go` — `runConfigImages` | list hook images |

## Scope

Only `pre_start` is in scope. `post_start` / `pre_stop` hooks run via `ExecCreate` **inside** the service container (`pkg/compose/hook.go`) and never use `hook.Image`, so there is nothing to pull for them. An empty hook image falls back to the service image, which is already resolved.

## Known limitations / follow-ups

Digest resolution and locking are **not** covered by this PR, deliberately:

- **`config --resolve-image-digests`**: `compose-go`'s `Project.WithImagesResolved` only resolves `service.Image`, never hook images. Covering hooks cleanly requires extending `WithImagesResolved` upstream — done in compose-spec/compose-go#894, to be wired here once it is released and vendored.
- **`config --lock-image-digests`**: this produces an *override* file that is re-merged, but `pre_start` has no special merge rule and its list is merged by **concatenation** (`override/merge.go`), so a hook override would **duplicate** the hook instead of pinning it. To be addressed together with / after the compose-go change.

## Tests

Unit tests only:

- `TestGetDependentImages` (`pkg/api`) — with/without hook image, and confirms `post_start`/`pre_stop` are not collected.
- `TestGetLocalImagesDigests_PreStartHook` (`pkg/compose`) — the hook image (`alpine:3.19`) is inspected alongside the service image (`alpine:3.20`).

## How to test manually

```yaml
services:
  sample:
    image: alpine:3.20
    command: sh -c 'sleep 5'
    pre_start:
      - image: alpine:3.19
        command: sh -c 'echo init'
```

```console
$ docker rmi alpine:3.19
$ docker compose config --images   # now lists alpine:3.19 AND alpine:3.20
$ docker compose pull              # now pulls alpine:3.19
```

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -760,3 +760,17 @@ func GetImageNameOrDefault(service types.ServiceConfig, projectName string) stri
 	}
 	return imageName
 }
+
+// GetDependentImages returns the additional images a service depends on beyond
+// its main image. Currently this is the set of pre_start hook images, which run
+// as ephemeral init containers with their own image (an empty hook image falls
+// back to the service image, which is already accounted for elsewhere).
+func GetDependentImages(service types.ServiceConfig) []string {
+	var images []string
+	for _, hook := range service.PreStart {
+		if hook.Image != "" {
+			images = append(images, hook.Image)
+		}
+	}
+	return images
+}
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -763,12 +763,14 @@ func GetImageNameOrDefault(service types.ServiceConfig, projectName string) stri
 
 // GetDependentImages returns the additional images a service depends on beyond
 // its main image. Currently this is the set of pre_start hook images, which run
-// as ephemeral init containers with their own image (an empty hook image falls
-// back to the service image, which is already accounted for elsewhere).
-func GetDependentImages(service types.ServiceConfig) []string {
+// as ephemeral init containers with their own image. A hook without an explicit
+// image, or one reusing the service image, is skipped as that image is already
+// accounted for as the service image.
+func GetDependentImages(service types.ServiceConfig, projectName string) []string {
+	serviceImage := GetImageNameOrDefault(service, projectName)
 	var images []string
 	for _, hook := range service.PreStart {
-		if hook.Image != "" {
+		if hook.Image != "" && hook.Image != serviceImage {
 			images = append(images, hook.Image)
 		}
 	}
--- a/pkg/compose/pull.go
+++ b/pkg/compose/pull.go
@@ -135,6 +135,42 @@ func (s *composeService) pull(ctx context.Context, project *types.Project, opts 
 		i++
 	}
 
+	// pre_start hook images run as ephemeral init containers with their own
+	// image, so they must be pulled too. They have no pull policy of their own,
+	// so we inherit the parent service's policy for skip decisions.
+	for name, service := range project.Services {
+		if service.PullPolicy == types.PullPolicyNever || service.PullPolicy == types.PullPolicyBuild {
+			continue
+		}
+		for _, img := range api.GetDependentImages(service) {
+			switch service.PullPolicy {
+			case types.PullPolicyMissing, types.PullPolicyIfNotPresent:
+				if imageAlreadyPresent(img, images) {
+					s.events.On(api.Resource{
+						ID:      "Image " + img,
+						Status:  api.Done,
+						Text:    "Skipped",
+						Details: "Image is already present locally",
+					})
+					continue
+				}
+			}
+			if _, ok := imagesBeingPulled[img]; ok {
+				continue
+			}
+			imagesBeingPulled[img] = name
+			hookService := types.ServiceConfig{Name: name, Image: img}
+			eg.Go(func() error {
+				_, err := s.pullServiceImage(ctx, hookService, opts.Quiet, project.Environment["DOCKER_DEFAULT_PLATFORM"])
+				if err != nil && !opts.IgnoreFailures {
+					// fail fast: a hook image can't be built as a fallback
+					return err
+				}
+				return nil
+			})
+		}
+	}
+
 	err = eg.Wait()
 
 	if len(mustBuild) > 0 {
@@ -313,6 +349,16 @@ func (s *composeService) pullRequiredImages(ctx context.Context, project *types.
 			}
 		}
 
+		for i, img := range api.GetDependentImages(service) {
+			if _, ok := images[img]; !ok {
+				// Hack: create a fake ServiceConfig so we pull missing pre_start hook image
+				n := fmt.Sprintf("%s:pre_start %d", name, i)
+				needPull[n] = types.ServiceConfig{
+					Name:  n,
+					Image: img,
+				}
+			}
+		}
 	}
 	if len(needPull) == 0 {
 		return nil
--- a/pkg/compose/pull.go
+++ b/pkg/compose/pull.go
@@ -136,15 +136,17 @@ func (s *composeService) pull(ctx context.Context, project *types.Project, opts 
 	}
 
 	// pre_start hook images run as ephemeral init containers with their own
-	// image, so they must be pulled too. They have no pull policy of their own,
-	// so we inherit the parent service's policy for skip decisions.
+	// registry image. They have no pull policy of their own, so we inherit the
+	// parent service's policy for skip decisions. Unlike the service image, a
+	// hook image can't be built, so `build` does not exempt it from pulling —
+	// only `never` does (consistent with the `up`/create path).
 	for name, service := range project.Services {
-		if service.PullPolicy == types.PullPolicyNever || service.PullPolicy == types.PullPolicyBuild {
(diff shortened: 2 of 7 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The PR correctly surfaces `pre_start` hook images for resolution, pulling, and listing. The core logic in `GetDependentImages`, the goroutine closures in `pull()`, and error propagation are all sound. No high or medium severity bugs were confirmed.

**Summary of analysis:**
- `GetDependentImages` correctly filters empty images and only includes `pre_start` hooks (not `post_start`/`pre_stop`, which have no image to pull).
- `pull()` correctly inherits the parent service's pull policy for hook images, including skipping on `PullPolicyNever`/`PullPolicyBuild`, checking local presence for `PullPolicyMissing`/`PullPolicyIfNotPresent`, and always pulling for `PullPolicyAlways`.
- `pullRequiredImages()` follows the same "pull if missing" pattern as the existing volume-image block — consistent with the function's existing design.
- Goroutine closures are safe: `hookService` is a local copy per iteration.
- Error propagation from hook-image pulls goes through `eg.Wait()` and is correctly returned to callers.

pkg/compose/pull.go
```diff
@@ -313,6 +349,16 @@ func (s *composeService) pullRequiredImages(ctx context.Context, project *types.
 			}
 		}
 
+		for i, img := range api.GetDependentImages(service) {
+			if _, ok := images[img]; !ok {
+				// Hack: create a fake ServiceConfig so we pull missing pre_start hook image
```
ndeloof: Note to myself: we should extract an `ImageRef` struct so `pullServiceImage` doesn't relies on `ServiceConfig` and we don't need such a hack

pkg/api/api.go
```diff
@@ -760,3 +760,17 @@ func GetImageNameOrDefault(service types.ServiceConfig, projectName string) stri
...
+// back to the service image, which is already accounted for elsewhere).
+func GetDependentImages(service types.ServiceConfig) []string {
+	var images []string
+	for _, hook := range service.PreStart {
+		if hook.Image != "" {
+			images = append(images, hook.Image)
+		}
+	}
+	return images
+}
```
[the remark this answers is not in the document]
ndeloof: Fixed in dd2ff0fb8: `GetDependentImages` now skips a hook image equal to the service image (resolved via `GetImageNameOrDefault`), so `config --images` no longer prints a duplicate and `pullRequiredImages` no longer schedules a redundant pull for it.

pkg/compose/pull.go
```diff
@@ -313,6 +349,16 @@ func (s *composeService) pullRequiredImages(ctx context.Context, project *types.
...
+				n := fmt.Sprintf("%s:pre_start %d", name, i)
+				needPull[n] = types.ServiceConfig{
+					Name:  n,
+					Image: img,
+				}
+			}
+		}
```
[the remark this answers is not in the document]
ndeloof: Fixed in dd2ff0fb8: `pullRequiredImages` now dedups dependent images by reference through a project-wide `scheduled` set, so several hooks/services sharing the same missing image no longer schedule concurrent redundant pulls. The hook pass was extracted into `addPreStartHookPulls` (run as a second pass so the dedup is independent of service iteration order). Same commit also makes the `pull` command pull hook images under `pull_policy: build` (only `never` skips them), for consistency with this path.

pkg/compose/pull.go
```diff
@@ -348,6 +393,35 @@ func (s *composeService) pullRequiredImages(ctx context.Context, project *types.
...
+// policy explicitly forbids any pull (never). Running as a second pass over the
+// services keeps the dedup against service/volume images (via scheduled)
+// independent of service iteration order.
+func addPreStartHookPulls(project *types.Project, images map[string]api.ImageSummary, needPull map[string]types.ServiceConfig, scheduled map[string]bool) {
+	for name, service := range project.Services {
+		if service.PullPolicy == types.PullPolicyNever {
+			continue
+		}
+		for i, img := range api.GetDependentImages(service, project.Name) {
+			if _, ok := images[img]; ok {
```
docker-agent: **[medium] `addPreStartHookPulls` silently skips hook images when service has `pull_policy: always`**

`addPreStartHookPulls` guards against pulling already-present images with a plain existence check:

```go
if _, ok := images[img]; ok {
    continue  // skips regardless of pull policy
}
```

This means that when a service has `pull_policy: always`, its own image is **force-pulled** every time (because `mustPull` returns `true` for `PullPolicyAlways`), but its `pre_start` hook images are silently **skipped** if they are already present locally — they never get the same "always re-pull" treatment.

A user who sets `pull_policy: always` to ensure freshness on every `docker compose up` will see the service image updated while stale hook images are left behind, which can lead to subtle divergence between the service container and its init hooks.

The fix is to also respect `PullPolicyAlways` in `addPreStartHookPulls`, analogous to how `mustPull` handles it for service images:

```suggestion
	for name, service := range project.Services {
		if service.PullPolicy == types.PullPolicyNever {
			continue
		}
		for i, img := range api.GetDependentImages(service, project.Name) {
			if service.PullPolicy != types.PullPolicyAlways {
				if _, ok := images[img]; ok {
					continue
				}
			}
			if scheduled[img] {
				continue
			}
```

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 67/100 |

docker-agent (commented):

This PR correctly wires `pre_start` hook images into `docker compose config --images`, `docker compose pull`, and the `up`/`create` pull path. The overall approach is sound. One inconsistency was found in the `up`/`create` path: hook images are not force-pulled when the parent service has `pull_policy: always`, while service images are. See the inline comment for details.

pkg/compose/pull.go
ndeloof: Good catch — fixed in 9d2dad5. addPreStartHookPulls now skips the already-present shortcut when the parent service is `pull_policy: always`, so hook images are force-pulled like the service image (and consistent with the `pull` command path). Covered by unit tests for the always / missing-present / missing-absent / never / dedup cases.

docker-agent: Thanks for the quick fix! The implementation looks spot-on — wrapping the existence check in a guard for the always policy exactly mirrors how mustPull handles service images, and the comprehensive test coverage gives me confidence this behavior will not regress.

ndeloof:
`pull_policy` inconsistency will be addressed in a follow-up PR

glours (approved):
LGTM

PR #13946, opened 2026-07-16
fix(build): use image content digest, not attested index digest

### LINKED ISSUE #13636: [BUG] `docker compose build` produces different image IDs on each rebuild with containerd image store, but works correctly with overlay2
### Description

When Docker Engine uses the containerd image store (default for new installations since
Docker 28), `docker compose build` produces a new image ID on every rebuild even when no
files have changed. This causes `docker compose up` to unnecessarily recreate all containers
on every deploy.

Disabling the containerd image store on the same server fixes the issue — the second build
reuses the cache completely and the image ID stays the same.

### Steps To Reproduce

Use Docker CE on Linux with the containerd image store enabled (default for fresh installs):

```
$ docker info -f '{{.Driver}}'
overlayfs
```

Create these three files:

**Dockerfile:**
```dockerfile
FROM alpine:3.21
RUN echo "hello" > /hello.txt
COPY config.txt /config.txt
CMD ["cat", "/hello.txt", "/config.txt"]
```

**config.txt:**
```
static config file
```

**docker-compose.yml:**
```yaml
services:
  app:
    build: .
    container_name: test-cache
```

Then run:

```bash
# Build twice with nothing changed in between:
docker compose build
docker images --format '{{.ID}}' --filter reference='*-app'

docker compose build
docker images --format '{{.ID}}' --filter reference='*-app'
```

## Expected behavior

Second build reuses cache. Image ID is the same. `docker compose up` does not recreate the
container.

## Actual behavior

Second build produces a different image ID. `docker compose up` sees a different image and
recreates the container.

Note: individual build steps show `CACHED` for RUN/COPY layers, but the final image still
gets a new ID. The exporting steps (`exporting layers`, `writing image`, `naming`) take
0.1–0.2s instead of 0.0s, suggesting they actually re-export rather than reuse.

### Compose Version

```Text
Docker Compose version v5.1.0
```

### Docker Environment

```Text
 Version:    29.3.0
 Context:    default
 Debug Mode: false
 Plugins:
    Version:  v0.31.1
    Version:  v5.1.0

Server:
 Containers: 8
  Running: 7
  Paused: 0
  Stopped: 1
 Images: 8
 Server Version: 29.3.0
 Storage Driver: overlayfs
 Logging Driver: local
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
 CDI spec directories:
  /etc/cdi
  /var/run/cdi
 Swarm: inactive
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 301b2dac98f15c27117da5c8af12118a041a31d9
 init version: de40ad0
 Security Options:
  apparmor
  seccomp
   Profile: builtin
  cgroupns
 Kernel Version: 6.8.0-106-generic
 Operating System: Ubuntu 24.04.4 LTS
 OSType: linux
 Architecture: x86_64
 CPUs: 4
 Total Memory: 7.57GiB
 Name: secondary
 ID: 091526cd-489b-4f47-acad-33efcea821cb
 Debug Mode: false
 Experimental: false
 Insecure Registries:
  ::1/128
  127.0.0.0/8
 Live Restore Enabled: false
 Firewall Backend: iptables
```

## Related issues

- docker/for-mac#7341 — same symptom on macOS Docker Desktop
- docker/for-mac#7359 — layer not cached under containerd image store (fixed by increasing
  `defaultKeepStorage`, but that does not help here — layers are small)

## Workaround

Disable the containerd image store:

```json
{
  "features": {
    "containerd-snapshotter": false
  }
}
```

After changing `daemon.json` and restarting Docker, existing images are no longer available
(format is incompatible) and need to be rebuilt/re-pulled. Volumes are preserved.

nabsei:
**What I did**

`docker compose build` (via Bake) recorded BuildKit's top-level `containerimage.digest` metadata field as the built image's identity, used to decide whether a container needs to be recreated. With provenance attestations enabled (the default since recent Buildx/BuildKit) and the containerd image store, this digest is for the whole attested index (image manifest + attestation manifest), and it changes on every build — even a fully cached one — because the attestation manifest embeds per-build metadata (timestamps, build refs). The image content itself never changes.

This made `docker compose up --build` recreate containers unnecessarily on every run, even when nothing actually changed.

The fix inspects the built (or already-local) image via the Engine API with `manifests=true` (API 1.48+ / Engine 28.0+) and picks the digest of the manifest with `Kind == "image"`, ignoring the `"attestation"` manifest. That digest only reflects the actual image content (config + layers), so it's stable across rebuilds with unchanged content, and still changes correctly when the content does change. I applied the same fix to `getImageSummaries` (used for already-local images) so both sides of the staleness comparison use the same, consistent digest. Images that only exist in a registry (push / multi-platform builds) aren't locally inspectable, so they keep the Bake-reported digest as before — this only affects the common local build case where the bug was reported.

**Verification**

I reproduced the bug against a real Docker Desktop engine (containerd snapshotter, default provenance attestations on arm64), confirmed the root cause by inspecting BuildKit's raw Bake metadata and the daemon's `manifests=true` image inspect response directly, then verified the fix end-to-end:
- Before the fix: `docker buildx bake` on an unchanged Dockerfile produces a different `containerimage.digest`/image ID on every run despite `CACHED` build steps.
- With the fix: two consecutive `docker compose up -d --build` runs with no changes keep the exact same container (same ID, same creation timestamp) — no unnecessary recreate.
- Sanity check: changing the Dockerfile content still correctly triggers `Recreate`/`Recreated`.
- `go build ./...`, `go vet ./...`, and `golangci-lint run` are all clean.
- Full `go test ./...` passes; the only pre-existing failures (in `pkg/watch` and `pkg/e2e`) are environmental/flaky and reproduce identically on unmodified `main`, unrelated to this change.
- Added unit tests (`TestContentDigest`, `TestGetImageSummariesUsesContentDigestNotAttestedIndexDigest`) covering the digest-selection logic directly.

**Related issue**

Fixes #13636

---
Disclosure: this PR was prepared with the assistance of Claude Code (Anthropic). All investigation, the root-cause diagnosis, the fix, and the verification steps above were done and checked by me; I'm disclosing the tool's involvement for transparency.

### DIFF
--- a/pkg/compose/build_bake.go
+++ b/pkg/compose/build_bake.go
@@ -404,6 +404,7 @@ func (s *composeService) doBuildBake(ctx context.Context, project *types.Project
 	}
 
 	results := map[string]string{}
+	var builtImages []string
 	for name := range serviceToBeBuild {
 		image := expectedImages[name]
 		target := targets[name]
@@ -412,8 +413,28 @@ func (s *composeService) doBuildBake(ctx context.Context, project *types.Project
 			return nil, fmt.Errorf("build result not found in Bake metadata for service %s", name)
 		}
 		results[image] = built.Digest
+		builtImages = append(builtImages, image)
 		s.events.On(builtEvent(image))
 	}
+
+	// Bake's "containerimage.digest" metadata is the top-level attested
+	// image/index digest. When BuildKit provenance attestations are enabled
+	// (the default), that digest also covers the attestation manifest and
+	// changes on every build even when the runnable image content is
+	// unchanged, causing compose to recreate containers unnecessarily (see
+	// https://github.com/docker/compose/issues/13636). Images that were
+	// loaded into the local engine (the common non-push case) are inspected
+	// to substitute a digest that only reflects the actual image content.
+	// Images that only exist in a registry (push/multi-platform builds)
+	// keep the Bake-reported digest.
+	summaries, err := s.getImageSummaries(ctx, builtImages)
+	if err != nil {
+		return nil, err
+	}
+	for image, summary := range summaries {
+		results[image] = summary.ID
+	}
+
 	return results, nil
 }
 
--- a/pkg/compose/images.go
+++ b/pkg/compose/images.go
@@ -28,6 +28,7 @@ import (
 	"github.com/containerd/platforms"
 	"github.com/distribution/reference"
 	"github.com/moby/moby/api/types/container"
+	"github.com/moby/moby/api/types/image"
 	"github.com/moby/moby/client"
 	"github.com/moby/moby/client/pkg/versions"
 	"golang.org/x/sync/errgroup"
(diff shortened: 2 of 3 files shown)


nabsei:
Closing this — looks like @glours landed an equivalent fix directly on main a few days ago (fd794ea, fixes #13636), so this is now redundant. Thanks for looking at the issue!

PR #13949, opened 2026-07-17
fix(build): use platform image-manifest digest, not attested index

### LINKED ISSUE #13636: [BUG] `docker compose build` produces different image IDs on each rebuild with containerd image store, but works correctly with overlay2

glours:
**What I did**
With the containerd image store and BuildKit provenance attestations (the default), a built image is stored as an attested index whose top-level digest also covers the attestation manifest. That digest churns on every build even when the runnable content is unchanged, so compose recreated containers on every `up --build`.

Compare the digest of the "image" kind manifest instead, selected for the target platform and restricted to locally available manifests, so it is deterministic and reflects only config + layers. Both the build and up sides of the staleness check go through the same selection, and registry-only images keep the Bake-reported digest.

**Related issue**
Fixes #13636

### DIFF
--- a/pkg/compose/build.go
+++ b/pkg/compose/build.go
@@ -215,7 +215,10 @@ func (s *composeService) getLocalImagesDigests(ctx context.Context, project *typ
 			if err != nil {
 				return nil, err
 			}
-			inspect, err := s.apiClient().ImageInspect(ctx, img.ID)
+			// inspect by name, not img.ID: img.ID now holds a content-manifest
+			// digest (see contentDigest) which is not necessarily inspectable,
+			// whereas the image name always resolves.
+			inspect, err := s.apiClient().ImageInspect(ctx, imgName)
 			if err != nil {
 				return nil, err
 			}
--- a/pkg/compose/build_bake.go
+++ b/pkg/compose/build_bake.go
@@ -404,6 +404,7 @@ func (s *composeService) doBuildBake(ctx context.Context, project *types.Project
 	}
 
 	results := map[string]string{}
+	var builtImages []string
 	for name := range serviceToBeBuild {
 		image := expectedImages[name]
 		target := targets[name]
@@ -412,8 +413,28 @@ func (s *composeService) doBuildBake(ctx context.Context, project *types.Project
 			return nil, fmt.Errorf("build result not found in Bake metadata for service %s", name)
 		}
 		results[image] = built.Digest
+		builtImages = append(builtImages, image)
 		s.events.On(builtEvent(image))
 	}
+
+	// Bake reports the top-level attested image/index digest, which changes on
+	// every build when provenance attestations are enabled — even for a fully
+	// cached build (see https://github.com/docker/compose/issues/13636). For
+	// images loaded into the local engine (the common non-push case), substitute
+	// the content digest so unchanged rebuilds don't recreate containers.
+	// Registry-only images (push/multi-platform) aren't inspectable locally, so
+	// they keep the Bake-reported digest.
+	// Best effort: if the built images can't be inspected, keep the
+	// Bake-reported digests rather than failing an already-successful build.
+	summaries, err := s.getImageSummaries(ctx, builtImages)
+	if err != nil {
+		logrus.Debugf("unable to inspect built images for content digest, keeping bake digests: %v", err)
+	} else {
+		for image, summary := range summaries {
+			results[image] = summary.ID
+		}
+	}
+
 	return results, nil
 }
 
--- a/pkg/compose/images.go
+++ b/pkg/compose/images.go
@@ -28,6 +28,7 @@ import (
 	"github.com/containerd/platforms"
 	"github.com/distribution/reference"
 	"github.com/moby/moby/api/types/container"
+	"github.com/moby/moby/api/types/image"
 	"github.com/moby/moby/client"
 	"github.com/moby/moby/client/pkg/versions"
 	"golang.org/x/sync/errgroup"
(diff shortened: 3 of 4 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The fix correctly replaces the attested-index top-level digest with the platform-specific content-manifest digest to prevent spurious container recreation on repeated builds.

**What was verified:**
- The `contentDigest` function's `len(available)==1` fallback is intentional and safe: when only one platform manifest is available locally but doesn't match the service platform, `getLocalImagesDigests` detects the mismatch via a secondary `ImageInspect` call and removes the entry — so the digest is never written to `ImageDigestLabel`.
- The use of `platforms.Default()` in `getImageSummaries` is correct for all realistic scenarios: single-platform cross-platform builds return the right manifest anyway (via the lone-available fallback), and multi-platform mismatches are caught by the downstream platform guard.
- `manifestsSupported` calling `RuntimeAPIVersion` is not a new failure mode: the result is cached after the first call (which happens in many earlier code paths during `compose up`), and daemon unavailability would fail the subsequent `ImageInspect` calls regardless.
- The `contentDigest` function and its six test cases comprehensively cover the important paths (no manifests, attested index, single-platform non-matching, multi-platform with match, unavailable manifests, and ambiguous no-match fallback).

docker-agent (commented):
### Assessment: 🟢 APPROVE

The PR correctly fixes the attested-index digest churn problem. The new `contentDigest` function is well-designed and handles all key scenarios:

- **Single-platform attested images** (primary fix): filters for `ManifestKindImage` + `Available` manifests, then platform-matches → returns stable content digest instead of churning index digest.
- **Cross-platform services** (`service.platform: linux/arm64` on amd64 host): single-manifest fallback (`len(available) == 1`) returns the correct platform digest regardless of the host-platform matcher.
- **Multi-platform images** (multiple available manifests, no host-platform match): falls back to `inspect.ID`, which is intentional — the PR comment explicitly documents this case.
- **Legacy engines** (API < 1.48): `manifestsSupported` gates the `WithManifests` option; `contentDigest` with empty `Manifests` returns `inspect.ID` (the plain image ID, correct for old stores).
- **Registry-only/push images**: `errdefs.IsNotFound` in `getImageSummaries` skips them, preserving the Bake-reported digest.

The `build.go` fix (inspect by `imgName` instead of `img.ID`) correctly avoids inspecting a manifest digest directly, which may not be addressable the same way as a config ID on the containerd store.

Test coverage in `images_test.go` is thorough, covering the normal attested case, single-manifest fallback, multi-platform matching, unavailable manifests, attestation-only images, and the no-manifests legacy fallback.

ndeloof (approved):

PR #13956, opened 2026-07-21
fix(config): pin type:image volume sources and pre_start hook images

### LINKED ISSUE #13827: [BUG] image version pinning isn't applied to image mounts
### Description

docker compose provides a way to pin/resolve services.image digest using either `--resolve-image-digests` or `--lock-image-digests`. However, it doesn't seem to pin image digests for image mounts. I would expect that image pinning would also apply there (services.volumes.source for type "image").

### Steps To Reproduce

With following compose.yaml:
```yaml
services:
  app:
    image: busybox:latest
    volumes:
      - type: image
        source: busybox:latest
        target: /test_mount
```

Resolving image digests returns:
```bash
docker compose config --resolve-image-digests
name: test_compose
services:
  app:
    image: docker.io/library/busybox:latest@sha256:fd8d9aa63ba2f0982b5304e1ee8d3b90a210bc1ffb5314d980eb6962f1a9715d
    networks:
      default: null
    volumes:
      - type: image
        source: busybox:latest
        target: /test_mount
networks:
  default:
    name: test_compose_default`
```

instead of 
```bash
docker compose config --resolve-image-digests
name: test_compose
services:
  app:
    image: docker.io/library/busybox:latest@sha256:fd8d9aa63ba2f0982b5304e1ee8d3b90a210bc1ffb5314d980eb6962f1a9715d
    networks:
      default: null
    volumes:
      - type: image
        source: busybox:latest@sha256:fd8d9aa63ba2f0982b5304e1ee8d3b90a210bc1ffb5314d980eb6962f1a9715d
        target: /test_mount
networks:
  default:
    name: test_compose_default`
```

I suppose same behavior is expected with `--lock-image-digests`.

### Compose Version

```Text
docker compose version
Docker Compose version v5.1.3


docker-compose version
Docker Compose version v5.1.3
```

### Docker Environment

```Text
docker info
Client:
 Version:    26.1.5+dfsg1
 Context:    default
 Debug Mode: false
 Plugins:
  agent: Docker AI Agent Runner (Docker Inc.)
    Version:  v1.57.0
    Path:     /usr/local/lib/docker/cli-plugins/docker-agent
    Version:  v1.20.2
    Path:     /usr/local/lib/docker/cli-plugins/docker-ai
    Version:  v0.34.0-desktop.1
    Path:     /usr/local/lib/docker/cli-plugins/docker-buildx
    Version:  v5.1.3
    Path:     /usr/local/lib/docker/cli-plugins/docker-compose
    Version:  0.0.47
    Path:     /usr/local/lib/docker/cli-plugins/docker-debug
    Version:  v0.3.0
    Path:     /usr/local/lib/docker/cli-plugins/docker-desktop
  dhi: CLI for managing Docker Hardened Images (Docker Inc.)
    Version:  v0.0.3
    Path:     /usr/local/lib/docker/cli-plugins/docker-dhi
    Version:  v0.2.31
    Path:     /usr/local/lib/docker/cli-plugins/docker-extension
    Version:  v1.4.0
    Path:     /usr/local/lib/docker/cli-plugins/docker-init
    Version:  v0.42.1
    Path:     /usr/local/lib/docker/cli-plugins/docker-mcp
    Version:  v1.1.37
    Path:     /usr/local/lib/docker/cli-plugins/docker-model
    Version:  v0.5.92
    Path:     /usr/local/lib/docker/cli-plugins/docker-offload
    Version:  v0.0.27
    Path:     /usr/local/lib/docker/cli-plugins/docker-pass
  sandbox:  (Docker Inc.)
    Version:  v0.12.0
    Path:     /usr/local/lib/docker/cli-plugins/docker-sandbox
    Version:  0.6.0
    Path:     /usr/local/lib/docker/cli-plugins/docker-sbom
    Version:  v1.20.4
    Path:     /usr/local/lib/docker/cli-plugins/docker-scout

Server:
 Containers: 0
  Running: 0
  Paused: 0
  Stopped: 0
 Images: 38
 Server Version: 29.5.2
 Storage Driver: overlayfs
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 2
 Plugins:
  Volume: local
 CDI spec directories:
  /etc/cdi
  /var/run/cdi
 Swarm: inactive
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 77c84241c7cbdd9b4eca2591793e3d4f4317c590 (expected: )
 runc version: v1.3.5-0-g488fc13e (expected: )
 init version: de40ad0 (expected: )
 Security Options:
  seccomp
   Profile: builtin
  cgroupns
 OSType: linux
 Architecture: x86_64
 CPUs: 12
 Total Memory: 15.52GiB
 Name: docker-desktop
 ID: 274db751-fa66-4f3a-81e0-7ac45e16bcb3
 Debug Mode: false
 Labels:
  com.docker.desktop.address=unix:///var/run/docker-cli.sock
 Experimental: true
 Insecure Registries:
  hubproxy.docker.internal:5555
  ::1/128
  127.0.0.0/8
 Live Restore Enabled: false
```

### Anything else?

_No response_

glours:
**What I did**
Rely on compose-go WithImagesResolved, which now resolves dependent images — `type: image` volume sources and pre_start hook images — with its already-digested guard, per-call memoization and sibling-service detection (compose-spec/compose-go#894, #899), rather than duplicating resolution logic CLI-side. The interpolated path gets this for free; --no-interpolate maps the raw model onto a pseudo-project keyed by service names to reuse the same resolution, and --lock-image-digests keeps `type: image` volumes in its output. As a side effect, `compose publish` now fails fast on unresolvable dependent images.

**Related issue**

Fixes #13827

image

### DIFF
--- a/cmd/compose/config.go
+++ b/cmd/compose/config.go
@@ -236,6 +236,7 @@ func runConfigInterpolate(ctx context.Context, dockerCli command.Cli, opts confi
 	}
 
 	if opts.lockImageDigests {
+		warnHooksNotLockable(project)
 		project = imagesOnly(project)
 	}
 
@@ -254,13 +255,19 @@ func runConfigInterpolate(ctx context.Context, dockerCli command.Cli, opts confi
 	return content, nil
 }
 
-// imagesOnly return project with all attributes removed but service.images
+// imagesOnly return project with all attributes removed but service.images and `type: image` volumes
 func imagesOnly(project *types.Project) *types.Project {
 	digests := types.Services{}
 	for name, config := range project.Services {
-		digests[name] = types.ServiceConfig{
+		service := types.ServiceConfig{
 			Image: config.Image,
 		}
+		for _, vol := range config.Volumes {
+			if vol.Type == types.VolumeTypeImage {
+				service.Volumes = append(service.Volumes, vol)
+			}
+		}
+		digests[name] = service
 	}
 	project = &types.Project{Services: digests}
 	return project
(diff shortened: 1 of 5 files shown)


cmd/compose/config.go
```diff
@@ -284,58 +290,124 @@ func runConfigNoInterpolate(ctx context.Context, dockerCli command.Cli, opts con
...
-func resolveImageDigests(ctx context.Context, dockerCli command.Cli, model map[string]any) (err error) {
-	// create a pseudo-project so we can rely on WithImagesResolved to resolve images
+func resolveImageDigests(ctx context.Context, dockerCli command.Cli, model map[string]any) error {
+	// create a pseudo-project so we can rely on WithImagesResolved to resolve images,
+	// pre_start hook images and `type: image` volume sources, keyed by actual service
+	// names so sources referencing another service are detected as such and kept unresolved
 	p := &types.Project{
 		Services: types.Services{},
 	}
 	services := model["services"].(map[string]any)
```
docker-agent: **[medium] Unguarded type assertion on `model["services"]` panics on empty/malformed Compose file**

`resolveImageDigests` unconditionally casts `model["services"]` to `map[string]any` at line 332:

```go
services := model["services"].(map[string]any)
```

If the raw Compose file has no `services` key (e.g. a file with only `networks:` or an empty file), `model["services"]` is `nil` and the bare type assertion panics at runtime instead of returning a descriptive error.

Compare with the guarded pattern already used elsewhere in the same file (line ~434):
```go
if _, ok := data["services"]; ok {
    for serviceName := range data["services"].(map[string]any) { ... }
}
```

A safe fix mirrors that guard — return early or skip silently when `services` is absent:
```go
services, ok := model["services"].(map[string]any)
if !ok {
    return nil
}
```

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 57/100 |

docker-agent (commented):

One medium-confidence finding in the new `--no-interpolate` code path.

cmd/compose/config.go
```diff
@@ -284,58 +290,128 @@ func runConfigNoInterpolate(ctx context.Context, dockerCli command.Cli, opts con
...
+			continue
+		}
+		for _, s := range e.(map[string]any) {
+			service := s.(map[string]any)
+			for key := range service {
+				switch key {
+				case "image":
+				case "volumes":
+					if volumes := imageVolumes(service); len(volumes) > 0 {
+						service["volumes"] = volumes
```
docker-agent: **[medium] `lockModel` mutates `service["volumes"]` type from `[]any` to `[]map[string]any`, risking downstream type-assertion panics**

`lockModel` calls `imageVolumes(service)` (which asserts `service["volumes"].([]any)` to read the volumes), then immediately writes the filtered result back via `service["volumes"] = volumes` where `volumes` is `[]map[string]any`. This permanently changes the runtime type of the `service["volumes"]` map entry from `[]any` to `[]map[string]any`.

Any subsequent call that does `service["volumes"].([]any)` — exactly what `imageVolumes` does on its first line — will **panic** with a type assertion failure. In the current `runConfigNoInterpolate` call order, `resolveImageDigests` (which calls `imageVolumes`) runs *before* `lockModel`, so the present code is safe. But this is an order-dependent invariant with no guard: if `lockModel` is ever called before another `imageVolumes` call on the same model map, or if the call order in `runConfigNoInterpolate` is changed without noticing this constraint, the panic will be triggered.

**Fix**: Preserve the original slice type, or call `imageVolumes` inside `lockModel` only once and avoid writing a typed slice back to the raw model. Alternatively, write `[]any` back after building the filtered list:
```go
if volumes := imageVolumes(service); len(volumes) > 0 {
    result := make([]any, len(volumes))
    for i, v := range volumes {
        result[i] = v
    }
    service["volumes"] = result
} else {
    delete(service, "volumes")
}
```

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 75/100 |

cmd/compose/config.go
```diff
@@ -284,58 +290,128 @@ func runConfigNoInterpolate(ctx context.Context, dockerCli command.Cli, opts con
...
+	// fields absent from the resolved pseudo-project (empty Image / Source) are left untouched
 	for name, s := range services {
 		service := s.(map[string]any)
 		config := p.Services[name]
 		if config.Image != "" {
 			service["image"] = config.Image
 		}
-		services[name] = service
+		for i, hook := range preStartHooks(service) {
+			if image := config.PreStart[i].Image; image != "" {
```
docker-agent, line 413: **[medium] `resolveImageDigests` write-back loop may panic with index-out-of-bounds if `WithImagesResolved` alters the `PreStart` or `Volumes` slice length**

`resolveImageDigests` builds a pseudo-project by appending one `types.ServiceHook` to `config.PreStart` for each entry returned by `preStartHooks(service)`, then calls `p.WithImagesResolved(...)`. The write-back loop then iterates `preStartHooks(service)` again (same filter) and accesses `config.PreStart[i]` by index.

This alignment holds only if `WithImagesResolved` leaves `p.Services[name].PreStart` with exactly the same number of entries as were appended during the build phase. If `WithImagesResolved` deduplicates, reorders, filters, or canonicalizes hooks (e.g., by normalising empty-image hooks or merging entries), the index will diverge and `config.PreStart[i]` will either panic (index out of bounds) or silently assign the wrong digest to the wrong hook. The same structural risk applies to the `config.Volumes[i]` write-back for `type: image` volumes.

**Fix**: Instead of using positional indexing across two independent loops, build a lookup map from the resolved pseudo-project before the write-back, or assert `i < len(config.PreStart)` (and `i < len(config.Volumes)`) with a bounds check before each indexed access. A safer pattern:
```go
for i, hook := range preStartHooks(service) {
    if i >= len(config.PreStart) {
        break // WithImagesResolved changed the count; stop rather than panic
    }
    if image := config.PreStart[i].Image; image != "" {
        hook["image"] = image
    }
}
```

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 60/100 |

docker-agent (commented):

Two medium-confidence findings in the new `--no-interpolate` image-digest resolution path. Both are in the newly added `lockModel` and `resolveImageDigests` functions in `cmd/compose/config.go`.

cmd/compose/config.go
glours: Not taking this one. The index alignment is guaranteed by `WithServicesTransform`'s contract: `WithImagesResolved` mutates `PreStart[i].Image`/`Volumes[i].Source` in place on a deep copy and never adds, removes or reorders entries, so both loops iterate the same filtered list with unchanged length. Adding a silent `break` would be worse than the hypothetical it guards against: if a future compose-go change ever broke this invariant, it would silently emit unpinned refs in a "locked" output instead of failing loudly in our tests (which cover hooks + image volumes across multiple services).

docker-agent: I appreciate the clarification—I misunderstood the mutation contract. You're right that if `WithServicesTransform` guarantees in-place mutation without reordering or filtering, the index alignment is safe and a silent `break` would mask a serious regression rather than prevent a runtime panic.

That said, the contract you're relying on isn't documented in the `WithServicesResolved` or `WithServicesTransform` signature or godoc (I checked `compose-go`), so future maintainers have no programmatic way to know this invariant exists. Would it make sense to add a short comment above the write-back loop explaining the alignment assumption, or alternatively add an assertion in your test suite that explicitly validates the pre/post lengths match after `WithImagesResolved`? That way the failure you're counting on to "fail loudly" is guaranteed to happen in the right place, rather than depending on downstream locked-output tests to catch it indirectly.

Thanks for the pushback—this is a clearer false positive on my part (the alignment *is* safe given the contract), but I think the visibility concern stands independently.

docker-agent (commented):
### Assessment: 🟢 APPROVE

The PR cleanly delegates image resolution for `type: image` volume sources and pre_start hook images to `WithImagesResolved` rather than duplicating logic CLI-side. The new `resolveImageDigests`, `preStartHooks`, and `imageVolumes` helpers in the `--no-interpolate` path are well-structured: both population and writeback use the same deterministic filtering functions, keeping slice indices aligned. The `lockModel` function correctly limits `--lock-image-digests` output to image/type-image-volume fields. Test coverage in `config_test.go` and `publish_test.go` covers the main scenarios.

docker-agent (commented):
### Assessment: 🟢 APPROVE

The diff refactors `cmd/compose/config.go` to delegate image digest resolution for `type: image` volume sources and `pre_start` hook images to compose-go's `WithImagesResolved`, rather than duplicating that logic CLI-side. Key correctness properties verified:

- **Index alignment** between pseudo-project construction and write-back loops is guaranteed — both use the same raw service map and helper functions with unchanged length.
- **Map reference semantics** are correctly used; the removed `services[name] = service` reassignment was redundant.
- **Hook filtering** by `preStartHooks` is consistent between construction and write-back, preserving index alignment even for non-map entries.
- **Tests** are comprehensive and cover the new code paths.

ndeloof (approved):

PR #13958, opened 2026-07-21
Push custom project config api at start of Up

nicksieger:
**What I did**

`POST /v{version}/compose/project` is a new custom api to be sent over the engine connection at the start of `compose up` that sends the full project configuration to the server to allow pre-populating caches and service maps. An engine marks that it can receive the config api by putting a  `com.docker.compose.coordinator` key in the context metadata with value bool `true` or string `"true"`.

- version-gated behind new apiVersionComposeProjectConfig constant
- skipped in dry-run
- non-fatal: warn and continue on request or engine error

**Related issue**

IMG_0074


docker-agent (commented):
🟢 **No issues found** — LGTM! [View logs](https://github.com/docker/compose/actions/runs/29866594683).

glours (commented):
The project-config push is coordinator-specific integration code, and `pkg/compose` isn't the right home for it. `pkg/` is Compose's reusable, general-purpose surface, tool-specific integrations should stay under `internal/`, the way Docker Desktop integration lives entirely in `internal/desktop` (endpoint detection, HTTP client, feature gating) with `pkg/compose` only calling into it.

Could we move this into a dedicated `internal/coordinator` package following that same pattern? The detection (the `com.docker.compose.project.config` context-metadata check), the HTTP client over the engine dialer, the versioned URL and the outcome handling would all live there, and `pkg/compose` would keep just a thin call-site in `up.go`.

Besides keeping the core library clean, this gives us a natural home as more coordinator-specific features land, rather than scattering them across `pkg/compose`.

**Separate concern - the request needs a timeout.** The `http.Client` has no `Timeout` and the `Up` context has no deadline; since this runs before `s.create`, a coordinator that accepts the connection but never responds would hang `up` indefinitely, so the "non-fatal, warn and continue" guarantee doesn't hold. This should be bounded (I'll leave the exact approach up to you).

internal/coordinator/coordinator.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+	req.Header.Set("Content-Type", "application/json")
+
+	resp, err := httpClient.Do(req)
+	if err != nil {
+		return err
+	}
+	defer func() {
+		_ = resp.Body.Close()
+	}()
```
docker-agent, line 178: **[medium] `http.Transport` created per-call is never cleaned up — idle-connection goroutines leak until timeout**

A new `*http.Transport` is allocated on every call to `PushProjectConfig`. Go's `http.Transport` spawns internal `persistConn` read/write goroutines per connection, and those goroutines stay alive until the transport's `IdleConnTimeout` fires (default 90 s) — the garbage collector does **not** stop them. Because `CloseIdleConnections()` is never called and the transport is immediately discarded, each invocation leaves an orphaned goroutine pair for up to 90 s.

For the current usage (one call per `compose up`) the practical impact is minimal. It becomes more visible in automated tests or tooling that invokes `compose up` in a loop, where goroutines pile up until they self-expire.

**Fix:** call `httpClient.CloseIdleConnections()` inside (or just after) the deferred close, so the transport is torn down eagerly once the response is handled:

```suggestion
	defer func() {
		_ = resp.Body.Close()
		httpClient.CloseIdleConnections()
	}()
```

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 62/100 |

docker-agent (commented):

- [medium] `internal/coordinator/coordinator.go`:80 — `PushEnabled` type-switch lacks `*map[string]any` arm — contexts stored via testStoreCfg factory may silently return false (confidence: weak 52/100)

  The `testStoreCfg` uses a factory returning `&map[string]any{}`. If the Docker CLI context store calls this factory to allocate a deserialization target, round-tripped metadata of type `map[string]any` would arrive as `*map[string]any`. The type-switch in `PushEnabled` only handles `command.DockerContext` and `map[string]any` — a `*map[string]any` falls to the `default` branch and returns `false`, silently disabling the project-config push for any such context. If this matters in production (contexts backed by generic map metadata), adding a `case *map[string]any:` arm is the fix.

pkg/compose/up.go
```diff
@@ -36,13 +36,23 @@ import (
...
 )
 
 func (s *composeService) Up(ctx context.Context, project *types.Project, options api.UpOptions) error { //nolint:gocyclo
 	err := Run(ctx, tracing.SpanWrapFunc("project/up", tracing.ProjectOptions(ctx, project), func(ctx context.Context) error {
+		// When the current Docker context opts into the project-config push, send
+		// the project configuration to the coordinator before any other Docker API
+		// call. Failures are non-fatal: warn and continue bringing the project up.
+		if !s.dryRun && coordinator.PushEnabled(s.dockerCli) {
+			if err := s.pushProjectConfig(ctx, project); err != nil {
+				logrus.Warnf("project config push to coordinator failed, continuing: %v", err)
```
glours: To ensure the warning is well displayed by ttyWriter (don't forget to remove the logrus import 😉 )
```suggestion
				 s.events.On(newEvent(api.ResourceCompose, api.Warning, "project config push to coordinator failed, continuing", err.Error()))
```

internal/coordinator/coordinator.go
```diff
@@ -0,0 +1,165 @@
...
+// POST /v{version}/compose/project for the Compose project-config push. It
+// documents the minimum coordinator version and is the single place the push
+// gates on.
+const MinAPIVersion = "1.51"
+
+// defaultTimeout bounds a single project-config push. The push runs before any
+// other Docker API call in "compose up", so without a deadline a coordinator
+// that accepts the connection but never responds would hang "up" indefinitely,
+// defeating the non-fatal "warn and continue" guarantee.
+const defaultTimeout = 30 * time.Second
```
glours: Isn't 10 second enough?
```suggestion
const defaultTimeout = 10 * time.Second
```

glours (changes requested):
1 last small change to ensure the warning message will be properly displayed by the `ttyWriter`

internal/coordinator/coordinator.go
```diff
@@ -0,0 +1,165 @@
...
+const responseBodyLimit = 2048
+
+// PushEnabled reports whether the current Docker context opts into the
+// project-config push via the MetadataKey metadata key. A metadata read
+// failure is treated as "not enabled" so that "compose up" is never blocked on
+// context inspection. The value is accepted as either a JSON boolean true or
+// the string "true" (case-insensitive), mirroring how custom context metadata
+// may be decoded (see ConfigFromDockerContext in
+// internal/tracing/docker_context.go).
+func PushEnabled(dockerCli command.Cli) bool {
```
glours: One last question: does `PushEnabled` only check whether the context used by the coordinator supports uploading the Compose stack, or is it simply meant to detect whether we’re using a coordinator as an Engine?

If it’s the latter, shouldn’t we rename the function to just Enabled?
```suggestion
func Enabled(dockerCli command.Cli) bool {
```

nicksieger: I think it initially meant to check if the custom coordinator /compose/project API is supported, but I think we can generalize to `Enabled` so that code using this function will read like `coordinator.Enabled`.

pkg/compose/up.go
```diff
@@ -36,13 +36,30 @@ import (
...
+		// the project configuration to the coordinator before any other Docker API
+		// call. Failures are non-fatal: warn and continue bringing the project up.
+		if !s.dryRun && coordinator.Enabled(s.dockerCli) {
+			// The push is complete only when "up" resolved the whole project:
+			// an empty service selection AND no profiles applied. A non-empty
+			// selection narrows the project to a subset (+ dependency closure),
+			// and an applied profile disables the services outside it; either
+			// way the payload is partial and the coordinator must merge it
+			// rather than treat it as the authoritative project.
+			complete := len(options.Start.Services) == 0 && !hasActiveProfiles(project.Profiles)
```
glours: A service under `profiles:` with no active profile ends up in `DisabledServices` (`json:"-"` → dropped by `MarshalJSON`), but `hasActiveProfiles` returns false → we'd send a partial payload with `complete=true` and the coordinator would prune the disabled service. `DisabledServices` is the signal we want: both `up <svc...>` and profile narrowing populate it.

```suggestion
                      // Any narrowing ("up <svc...>", profiles) moves the excluded
                      // services to DisabledServices, which MarshalJSON omits: if it's
                      // not empty the payload is partial and the coordinator must merge
                      // it rather than treat it as authoritative.
                      complete := len(project.DisabledServices) == 0
```

Could you add a test for it? A service under `profiles: [debug]`, no active profile → header must be `false`.

pkg/compose/up.go
```diff
@@ -302,6 +319,31 @@ func (s *composeService) Up(ctx context.Context, project *types.Project, options
...
+// COMPOSE_PROFILES), so blank entries are ignored.
+func hasActiveProfiles(profiles []string) bool {
+	for _, p := range profiles {
+		if p != "" {
+			return true
+		}
+	}
+	return false
+}
+
```
glours: With the new predicate to test if the model is complete or not, we don't need this function anymore

internal/coordinator/coordinator.go
```diff
@@ -0,0 +1,182 @@
...
+	}()
+
+	if resp.StatusCode >= http.StatusBadRequest {
+		body, _ := io.ReadAll(io.LimitReader(resp.Body, responseBodyLimit))
+		msg := strings.TrimSpace(string(body))
+		if msg == "" {
+			return fmt.Errorf("coordinator returned status %d", resp.StatusCode)
+		}
+		return fmt.Errorf("coordinator returned status %d: %s", resp.StatusCode, msg)
+	}
```
glours: What do you think of doing this?

```suggestion
	switch {
      case resp.StatusCode == http.StatusNotFound || resp.StatusCode == http.StatusNotImplemented:
              // The engine answered but does not expose the coordinator endpoint
              // (e.g. a plain moby daemon behind a context that carries the
              // metadata flag). Callers treat this as "feature unsupported"
              // rather than as an error worth surfacing.
              return fmt.Errorf("%w (status %d)", ErrNotSupported, resp.StatusCode)
      case resp.StatusCode >= http.StatusBadRequest:
              body, _ := io.ReadAll(io.LimitReader(resp.Body, responseBodyLimit))
              msg := strings.TrimSpace(string(body))
              if msg == "" {
                      return fmt.Errorf("coordinator returned status %d", resp.StatusCode)
              }
              return fmt.Errorf("coordinator returned status %d: %s", resp.StatusCode, msg)
      }
```

glours (changes requested):
I think there is still an issue with the `profiles` support in the proposed implementation

pkg/compose/up.go
```diff
@@ -302,6 +319,14 @@ func (s *composeService) Up(ctx context.Context, project *types.Project, options
 	return err
 }
 
+func (s *composeService) pushProjectConfig(ctx context.Context, project *types.Project, complete bool) error {
+	version, err := s.RuntimeAPIVersion(ctx)
+	if err != nil {
+		return fmt.Errorf("negotiating API version: %w", err)
+	}
+	return coordinator.NewClient(s.apiClient().Dialer()).PushProjectConfig(ctx, version, project, complete)
```
ndeloof, line 327: `Dialer` unfortunately is the only exported component we can get from Moby client, after it has been moved to a module and `HTTPClient()` has been removed. I wonder we should open a PR on moby to get some `Client.Do(ctx, method, path, query, body, headers)` method exposed to allow extensibility (see https://github.com/moby/moby/pull/53021).

ndeloof: https://github.com/moby/moby/pull/53345

ndeloof (changes requested):
My main concern here is this introduces a new Endpoint on top of moby API, while https://github.com/moby/moby/pull/53021 is already discussed to offer an official extensibility mechanism. I wonder it would be safer (in terms of long-terms maintenance) to align with the latter, and define a general purpose "compose" extension endpoint (that is considered anyway) that could benefit both coordinator + general purpose compose usage (typically : jobs support)

See a PoC here: https://github.com/ndeloof/docker/pull/4

PR #13966, opened 2026-07-23
feat(reconcile): model network lifecycle in the plan

ndeloof:
Follow-up to #13962 (volumes), applying the same treatment to networks.

**What I did**

Move network divergence detection and recreation out of the imperative
pre-reconcile path (`ensureNetwork`/`resolveOrCreateNetwork`/`removeDivergedNetwork`)
into the reconciliation plan.

- `reconcileNetworks` owns creation of missing networks and, for a diverged
  network, an explicit recreation sequence — **no user confirmation**, since
  recreating a network is not destructive:
  `stop containers → disconnect → remove network → create network → reconnect`.
  Attached containers keep their identity (reconnected, not recreated), matching
  the previous behavior. If a container is independently recreated by
  `reconcileContainers`, its removal is ordered after the reconnect so they don't race.
- `collectObservedState` discovers legacy/unlabeled networks by name and records
  them as unmanaged matches (empty config-hash) so the reconciler reuses them
  untouched; ownership warnings move to `warnUnmanagedNetworks`.
  `checkExternalNetworks` keeps external-network validation/resolution.
- `execCreateNetwork` now issues a plain `createNetwork`; the imperative
  `ensureNetwork`/`resolveOrCreateNetwork`/`removeDivergedNetwork` and the
  connect/disconnect helpers are removed.

Reconcile, observed-state and executor tests cover network create/diverge/rename,
the entangled diverge+recreate case, legacy by-name discovery and the ownership warnings.

**Network rename behavior (intentional change)**

A network rename is handled as a recreation: the old network is removed, the new
one created, and attached containers are **migrated** onto it. This differs
slightly from previous Compose releases, which created the new network and left
existing containers attached to the old one (they'd only move once recreated for
another reason). We chose migration for the more logical outcome — after the next
`up`, the service runs on the network it declares.

Because networks carry no data (unlike volumes, where the old volume is kept),
removing the previous network is safe and, importantly, keeps subsequent runs
deterministic: it avoids leaving two networks sharing the same compose label.
The volume/network asymmetry (volume rename keeps the old volume, network rename
removes it) is intentional and reflects that only volumes hold data worth preserving.

### DIFF
--- a/pkg/compose/create.go
+++ b/pkg/compose/create.go
@@ -1453,6 +1453,14 @@ func (s *composeService) createNetwork(ctx context.Context, n *types.NetworkConf
 	s.events.On(creatingEvent(networkEventName))
 
 	if _, err := s.apiClient().NetworkCreate(ctx, n.Name, createOpts); err != nil {
+		// A concurrent `docker compose up|run` may have created the same network
+		// between the observed-state snapshot and now. Treat the resulting
+		// conflict as success rather than failing hard, mirroring the retry the
+		// previous ensureNetwork performed.
+		if errdefs.IsConflict(err) {
+			s.events.On(createdEvent(networkEventName))
+			return nil
+		}
 		s.events.On(errorEvent(networkEventName, err.Error()))
 		return fmt.Errorf("failed to create network %s: %w", n.Name, err)
 	}
--- a/pkg/compose/create.go
+++ b/pkg/compose/create.go
@@ -86,8 +86,7 @@ func (s *composeService) create(ctx context.Context, project *types.Project, opt
 	}
 
 	prepareNetworks(project)
-
-	networks, err := s.ensureNetworks(ctx, project)
+	externalNetworks, err := s.checkExternalNetworks(ctx, project)
 	if err != nil {
 		return err
 	}
@@ -108,8 +107,9 @@ func (s *composeService) create(ctx context.Context, project *types.Project, opt
 	if err != nil {
 		return err
 	}
-	observed.setResolvedNetworks(networks, project)
+	observed.setResolvedNetworks(externalNetworks, project)
 	observed.setResolvedVolumes(externalVolumes)
+	warnUnmanagedNetworks(project, observed)
 	warnUnmanagedVolumes(project, observed)
 
 	if len(observed.Orphans) > 0 && !options.IgnoreOrphans && !options.RemoveOrphans {
@@ -141,17 +141,51 @@ func prepareNetworks(project *types.Project) {
 	}
 }
 
-func (s *composeService) ensureNetworks(ctx context.Context, project *types.Project) (map[string]string, error) {
-	networks := map[string]string{}
-	for name, nw := range project.Networks {
-		id, err := s.ensureNetwork(ctx, project, name, &nw)
+// checkExternalNetworks validates that every external network exists and returns
+// their resolved IDs. External networks carry no compose label and are therefore
+// absent from the label-scoped observed state, so the reconciler needs them
+// injected via setResolvedNetworks.
+//
+// Managed and legacy (unlabeled, name-matched) networks are discovered by
+// collectObservedState; their lifecycle is owned by the reconciliation plan, so
+// this function performs no mutation on them.
+func (s *composeService) checkExternalNetworks(ctx context.Context, project *types.Project) (map[string]string, error) {
+	external := map[string]string{}
+	for k, nw := range project.Networks {
+		if !nw.External {
+			continue
+		}
+		id, err := s.resolveExternalNetwork(ctx, &nw)
 		if err != nil {
 			return nil, err
 		}
-		networks[name] = id
-		project.Networks[name] = nw
+		external[k] = id
 	}
-	return networks, nil
+	return external, nil
+}
+
+// warnUnmanagedNetworks warns about declared networks backed by a live network
+// this project does not own — either created outside Compose (no project label)
+// or by another project. Such networks are matched by name and reused untouched
+// (see discoverUnmanagedNetworks); the warning tells the user to set
+// `external: true` to make the intent explicit.
+func warnUnmanagedNetworks(project *types.Project, observed *ObservedState) {
+	for k, nw := range project.Networks {
+		if nw.External {
+			continue
+		}
+		obs, ok := observed.Networks[k]
+		if !ok || obs.ProjectName == project.Name {
+			continue
+		}
+		if obs.ProjectName == "" {
+			logrus.Warnf("a network with name %s exists but was not created by compose.\n"+
+				"Set `external: true` to use an existing network", nw.Name)
+		} else {
+			logrus.Warnf("a network with name %s exists but was not created for project %q.\n"+
+				"Set `external: true` to use an existing network", nw.Name, project.Name)
+		}
+	}
 }
 
 // prepareVolumes injects the compose-managed labels onto every project volume so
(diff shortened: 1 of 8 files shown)


pkg/compose/create.go
```diff
@@ -1494,99 +1444,19 @@ func (s *composeService) resolveOrCreateNetwork(ctx context.Context, project *ty
...
 		}
 		createOpts.IPAM.Config = append(createOpts.IPAM.Config, c)
 	}
 
 	networkEventName := fmt.Sprintf("Network %s", n.Name)
 	s.events.On(creatingEvent(networkEventName))
 
-	resp, err := s.apiClient().NetworkCreate(ctx, n.Name, createOpts)
-	if err != nil {
+	if _, err := s.apiClient().NetworkCreate(ctx, n.Name, createOpts); err != nil {
```
docker-agent, line 1455: **[medium] `createNetwork` no longer retries on conflict — concurrent `docker compose up` will fail**

The deleted `ensureNetwork` contained an explicit retry on `errdefs.IsConflict`:

```go
// Removed by this PR:
id, err := s.resolveOrCreateNetwork(ctx, project, name, n)
if errdefs.IsConflict(err) {
    // Maybe another execution of `docker compose up|run` created same network
    // let's retry once
    return s.resolveOrCreateNetwork(ctx, project, name, n)
}
```

The new `createNetwork` propagates the error without retrying. There is a TOCTOU window between `collectObservedState` (which skips scheduling a `CreateNetwork` node if the network already exists) and the actual `NetworkCreate` API call: if a second `docker compose up` or `docker compose run` process creates the same network in that window, the call here returns a conflict error that surfaces to the user as a hard failure.

The fix would be to detect `errdefs.IsConflict` after `NetworkCreate` and, in that case, verify the existing network is compatible (or simply treat it as success), mirroring what the old `resolveOrCreateNetwork` did.

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 97/100 |

pkg/compose/observed_state.go
```diff
@@ -156,13 +156,55 @@ func (s *composeService) collectObservedState(ctx context.Context, project *type
...
+		// NetworkInspect matches on ID prefix, so guard against a partial match
+		// (e.g. a network whose ID starts with the requested name).
+		if inspected.Network.Name != nw.Name && inspected.Network.ID != nw.Name {
+			continue
+		}
+		state.Networks[key] = ObservedNetwork{
+			ID:          inspected.Network.ID,
+			Name:        inspected.Network.Name,
+			ProjectName: inspected.Network.Labels[api.ProjectLabel],
+			// ConfigHash intentionally left empty: the network is not owned by
```
docker-agent: **[low] `discoverUnmanagedNetworks` records project-owned networks with an empty `ConfigHash`, silently skipping divergence detection**

When a network carries `Labels[api.ProjectLabel] == project.Name` but lacks `Labels[api.NetworkLabel]`, it is invisible to the label-filtered `NetworkList` call in `collectObservedState` (that loop skips entries with `key == ""`). `discoverUnmanagedNetworks` then picks it up and stores it with `ConfigHash: ""`.

Down in `reconcileNetworks`:

```go
if observed.ConfigHash == "" || observed.ConfigHash == expected {
    continue  // treated as "up-to-date / unmanaged"
}
```

…the empty hash causes the reconciler to skip divergence detection for this network, even if it has genuinely drifted from the desired config. The same network would have been recreated correctly by the old `resolveOrCreateNetwork` path (which compared hashes regardless of label completeness).

In practice this only affects networks created by an older Compose version that wrote the project label but not the network label — a narrow edge case — but it is a silent correctness regression for those deployments. If `discoverUnmanagedNetworks` finds a network whose `Labels[api.ConfigHashLabel]` is non-empty, it could preserve that hash rather than clearing it, letting the reconciler catch genuine divergence.

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 57/100 |

docker-agent (commented):

This PR cleanly migrates network lifecycle into the reconciliation plan. The plan ordering (stop → disconnect → remove → create → reconnect) and the anti-race dependency edges (`connectNodes`) look correct and are well-tested. Two findings worth addressing:

1. **`createNetwork` drops the conflict-retry** that `ensureNetwork` had for concurrent invocations — concurrent `docker compose up` or `docker compose run` processes can now fail with a hard error instead of retrying.
2. **Networks with an incomplete label set** (project label present, network label absent) are stored with an empty `ConfigHash` by `discoverUnmanagedNetworks`, silently bypassing divergence detection for a narrow class of legacy deployments.

pkg/compose/create.go
ndeloof: Fixed in e9bbe7d. createNetwork now treats a NetworkCreate conflict as success (emits the created event, returns nil), so a concurrent up/run creating the same network in the TOCTOU window no longer fails hard — mirroring the old ensureNetwork retry. Covered by TestExecutePlanCreateNetworkConflictIsSuccess.

pkg/compose/observed_state.go
ndeloof: Fixed in e9bbe7d. discoverUnmanagedNetworks (and its volume counterpart) now preserve the config-hash when the resource is owned by this project (project label matches) — so a network/volume written by an older Compose with the project label but no key label still gets divergence detection. Resources we don't own keep an empty hash and are reused untouched. Covered by TestCollectObservedState_OwnedNetworkMissingKeyLabelKeepsHash and its volume twin.

docker-agent: Perfect — the new `ownedConfigHash` helper cleanly separates the "owned but unlabeled" case (preserve the hash) from the "foreign resource" case (empty hash, reuse untouched). The test coverage for both networks and volumes confirms the fix handles the edge case correctly while leaving unowned resources alone. Thank you for addressing this!

pkg/compose/reconcile.go
```diff
@@ -133,101 +140,159 @@ func reconcile(_ context.Context, project *types.Project, observed *ObservedStat
...
+			resID := fmt.Sprintf("service:%s:%d", oc.Summary.Labels[api.ServiceLabel], oc.Number)
+			deps := []*PlanNode{createNode}
+			if rename {
+				deps = append(deps, disconnectNodes[i])
+			}
+			connectNode := r.plan.addNode(Operation{
+				Type:       OpConnectNetwork,
+				ResourceID: resID,
+				Cause:      fmt.Sprintf("network %s recreate", key),
+				Container:  &oc.Summary,
```
docker-agent, line 334: **[medium] OpConnectNetwork targets the old container ID when the container is also being independently recreated**

In `planRecreateNetworks`, the reconnect node is built with `Container: &oc.Summary`, which holds the **observed** (old) container's ID. When `reconcileContainers` later decides to recreate that same container for an unrelated reason (e.g. its config hash changed), the plan's dependency ordering is:

```
StopContainer(old)
   ↓ (parallel with)
OpConnectNetwork(old_id → new_network)   ← this targets the old container
   ↓
RemoveContainer(old)          ← waits for OpConnectNetwork via connectNodes
   ↓
RenameContainer(new)
```

The `connectNodes` guard correctly prevents `RemoveContainer` from racing `OpConnectNetwork`, so the removal ordering is sound. However, `StopContainer` is **not** ordered relative to `OpConnectNetwork`: the old container may already be stopped (or stopping) when `NetworkConnect(old_id)` is called. Whether Docker accepts `NetworkConnect` on a stopped container is implementation-defined; if it rejects it, the reconnect step errors and the entire reconcile operation fails even though the new container is already being created with the new network via `infrastructureDeps`.

The reconnect of the old container is also redundant when the container is being recreated: the new container is gated on `networkNodes[net]` (the `CreateNetwork` node) through `infrastructureDeps`, so it is created already attached to the new network.

**Suggested fix:** In `planRecreateNetworks`, skip emitting `OpConnectNetwork` for any container that is already scheduled for recreation (`r.stoppedByPlan` is set from an earlier volume-recreate stage, or the container will be detected diverged by `reconcileContainers`). Since the check cannot be done at plan-build time for `reconcileContainers`-driven recreations (they run after `planRecreateNetworks`), a simpler guard is: if the container's service has a config-hash mismatch, skip the reconnect — the new container will come up on the correct network. Alternatively, record the fact that the container's reconnect node was already provided by a network recreate, and let `planRecreateContainer` skip the `connectNodes` for it while still honoring the `createNetwork` ordering via `infrastructureDeps`.

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 57/100 |

docker-agent (commented):

This PR cleanly ports network lifecycle management into the declarative reconciliation plan, mirroring the prior volumes change. The dependency-ordering machinery (`stoppedByPlan`, `connectNodes`, `infrastructureDeps`) is well-structured and the test coverage is solid.

One moderate-confidence finding: in the entangled scenario where a network diverges **and** an attached container is independently scheduled for recreation, `planRecreateNetworks` emits an `OpConnectNetwork` targeting the old container's ID. The ordering guarantee prevents the reconnect from racing the removal, but `StopContainer` is not ordered relative to `OpConnectNetwork`, so the reconnect may be called against a stopped container — Docker's behavior in that case is implementation-defined. The new container already attaches to the new network via `infrastructureDeps`, so the reconnect is also redundant in this scenario.

pkg/compose/reconcile.go
```diff
@@ -133,101 +140,159 @@ func reconcile(_ context.Context, project *types.Project, observed *ObservedStat
...
+					Type:       OpStopContainer,
+					ResourceID: resID,
+					Cause:      fmt.Sprintf("network %s config changed", key),
+					Container:  &oc.Summary,
+					Timeout:    r.options.Timeout,
+				}, "")
+				r.stoppedByPlan[oc.ID] = stopNode
+			}
+			disconnectNodes = append(disconnectNodes, r.plan.addNode(Operation{
+				Type:       OpDisconnectNetwork,
```
glours, line 275: After a rename where the old network survives (the exact case `BestEffort` was added for), it keeps its compose labels: two live networks now share the same project+network labels. The discovery loop then picks a daemon-order-dependent winner via
  map overwrite (`observed_state.go:126-137`).

  On the next `up`, if the **old** network wins, the rename path re-fires and emits `OpDisconnectNetwork` for containers that are only attached to the **new** network (`reconcile.go:229-236`). Verified against a live daemon: `NetworkDisconnect` on a non-attached container errors even with `Force: true` → the errgroup cancels the whole plan **after** the stop node ran → `up` hard-fails with containers left stopped. And nothing ever removes the old network, so every subsequent `up` is a coin flip between a no-op and this failure.

  **Suggested fix** (first half alone prevents the failure):
  1. In `planRecreateNetworks`, only emit the disconnect for containers actually attached to the observed network — `oc.ConnectedNetworks` already has the data.
  2. On label-key collision in the discovery loop, prefer the network whose `Name` matches `project.Networks[key].Name` (and/or warn), so runs become deterministic.

glours (changes requested):
Rename doesn't converge when the best-effort removal is skipped

pkg/compose/reconcile.go
ndeloof: this is the exact purpose for https://github.com/docker/compose/pull/13967

glours (dismissed):

glours (approved):
LGTM

PR #13967, opened 2026-07-24
refactor(reconcile): aggregate observed resources, resolve conflicts in reconcile

ndeloof:
> [!IMPORTANT]
> **Draft — stacked on #13966** (networks in the reconciliation plan), itself a follow-up to #13962 (volumes). Review after #13966; until it merges, the diff here includes its commits. The last commit is the only one specific to this PR.

**What I did**

`collectObservedState` indexed networks and volumes by their compose label into a
**single-valued** map, so when two live resources share a label — typically a
leftover after a rename (volume rename keeps the old volume for its data) — they
collided and the "winner" depended on the daemon's list order. The result was a
nondeterministic `up`: on a subsequent, unchanged `up` the reconciler could pick
the stale resource and emit spurious create/recreate events (churn).

This makes collection lossless and moves the selection into the reconciler:

- `ObservedState.Networks`/`Volumes` become `map[string][]Observed*`: collection
  records **every** label-sharing resource and makes no premature choice.
- `selectNetwork`/`selectVolume` deterministically pick the resource matching the
  desired name (else the lexicographically smallest name), returning the others
  as orphans.
- `reconcile` resolves the observed state once (`resolveObserved`) into
  single-valued `resolvedNetworks`/`resolvedVolumes` used everywhere, and **warns**
  about orphans instead of acting on them — they are left untouched because
  removing them could drop data (volumes) or break unrelated workloads.

Selection unit tests, a collector aggregation test and a reconcile conflict test
(deterministic no-op + orphan warning across both list orders) are included.

**Note**

Orphans are only reported, never removed — the user decides whether to clean them
up. This resolves the nondeterminism the two previous PRs left open (a rename can
transiently leave two resources under the same compose label).

### DIFF
--- a/pkg/compose/create.go
+++ b/pkg/compose/create.go
@@ -86,8 +86,7 @@ func (s *composeService) create(ctx context.Context, project *types.Project, opt
 	}
 
 	prepareNetworks(project)
-
-	networks, err := s.ensureNetworks(ctx, project)
+	externalNetworks, err := s.checkExternalNetworks(ctx, project)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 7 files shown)


docker-agent (commented):
🟢 **No issues found** — LGTM! [View logs](https://github.com/docker/compose/actions/runs/30078405200).

ndeloof:
Folded into #13966: the observed-state aggregation and deterministic conflict resolution (resources sharing the same compose label) now live directly in that PR, on top of the network reconcile work and its review fixes. Closing this stacked draft to avoid a duplicate.

PR #13971, opened 2026-07-25
Fix broken MAINTAINERS link in CONTRIBUTING.md

ZayanKhan-12:
`CONTRIBUTING.md` links to the MAINTAINERS file as a repo-relative `[MAINTAINERS](MAINTAINERS)`, but that file no longer exists in this repository, so the link 404s.

Point it at the canonical MAINTAINERS file in `docker/opensource`, which this same document already links to a few lines later. Verified the target returns HTTP 200. Docs-only change; commit is DCO signed-off.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -193,7 +193,7 @@ indicate acceptance.
 A change requires at least 2 LGTMs from the maintainers of each
 component affected.
 
-For more details, see the [MAINTAINERS](MAINTAINERS) page.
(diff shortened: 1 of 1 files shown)


maxproske:
@ZayanKhan-12 Nice catch. Given it appears to have been unintentionally removed in #13231, I'd suggest restoring the file, rather than pointing to a Maintainerslist that was last updated 4 years ago.

ndeloof:
removal was intentional, we rely on .github/CODEOWNERS

maxproske:
Makes sense!

Small potatoes: CODEOWNERS team members are private to anyone not in the docker org

PR #13603, opened 2026-02-21
Tolerate missing env file on more runtime commands

maxproske:
**What I did**

Align toProjectName (used by exec, image, etc.) with projectOrName (used by ps, logs, etc.), which already had these flags since #13156

I know, this skips compose.yaml parsing. Being pragmatic for now, as I haven't touched compose-go yet, but can probably add a function to avoid the need to call ToProject at all.

**Related issue**

Closes #13601

PR #13156, opened 2025-08-15
Fix runtime operations failing when env file is missing

maxproske:
**What I did**

Closes #13135

Tolerate missing env file on runtime commands that should only depend on container labels like down, ps, logs, etc.

When an env file is deleted while a container is running, or an unused env file is declared by another service, most Compose commands fail, including `docker compose down`. Containers become unmanageable by Compose, and you must use `docker down` directly or recreate env files to bring containers down.

Note: This pattern likely affects other file dependencies like `secrets.file` and `configs.file`.

## Steps to Reproduce

```bash
cat > compose.yaml << 'EOF'
services:
  app:
    image: alpine
    env_file: deleteme.env
EOF

touch deleteme.env

docker compose up -d

rm deleteme.env

docker compose ps
# env file /Users/max/dev/repro/compose-13135/deleteme.env not found: stat /Users/max/dev/repro/compose-13135/deleteme.env: no such file or directory

docker compose down
# env file /Users/max/dev/repro/compose-13135/deleteme.env not found: stat /Users/max/dev/repro/compose-13135/deleteme.env: no such file or directory
```

PR #13973, opened 2026-07-27
fix: tolerate missing env file on scale, watch and shell completion

glours:
**What I did**
Follow-up to #13603: scale, watch and shell completion loaded the project without any tolerance option, so a missing env_file on a service not involved in the operation aborted the command, while up/exec/ps already tolerate this since #13156 and #13603.

Mirror the WithServices pattern: load with WithoutEnvironmentResolution and resolve the environment once the project has been reduced to the selected services, so targeted services still get their env_file validated. Completion only needs names and never resolves. This also aligns the config hash of scale-created containers with up-created ones.
**Related issue**

N/A

image

### DIFF
--- a/pkg/e2e/env_file_test.go
+++ b/pkg/e2e/env_file_test.go
@@ -42,6 +42,25 @@ func TestUnusedMissingEnvFile(t *testing.T) {
 	c.RunDockerComposeCmd(t, "-f", "./fixtures/env_file/compose.yaml", "ps")
 	c.RunDockerComposeCmd(t, "-f", "./fixtures/env_file/compose.yaml", "logs")
 	c.RunDockerComposeCmd(t, "-f", "./fixtures/env_file/compose.yaml", "exec", "serviceA", "echo", "hello")
+
+	// scale should work even with missing env file on a service not being scaled
+	c.RunDockerComposeCmd(t, "-f", "./fixtures/env_file/compose.yaml", "scale", "serviceA=2")
+
+	// but scaling the service with the missing env file must still fail
+	res := c.RunDockerComposeCmdNoCheck(t, "-f", "./fixtures/env_file/compose.yaml", "scale", "serviceB=1")
+	res.Assert(t, icmd.Expected{ExitCode: 1, Err: "env file /doesnotexist/.env not found"})
+
+	// shell completion should list services even with missing env file.
+	// ComposeStandalonePath fails the test outside standalone mode, so only the
+	// plugin form can be the default here.
+	completeCmd := []string{DockerExecutableName, "__complete", "compose"}
+	if composeStandaloneMode {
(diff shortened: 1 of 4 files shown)


pkg/e2e/env_file_test.go
```diff
@@ -42,6 +42,25 @@ func TestUnusedMissingEnvFile(t *testing.T) {
...
+	// plugin form can be the default here.
+	completeCmd := []string{DockerExecutableName, "__complete", "compose"}
+	if composeStandaloneMode {
+		completeCmd = []string{ComposeStandalonePath(t), "__complete"}
+	}
+	res = c.RunCmd(t, append(completeCmd, "-f", "./fixtures/env_file/compose.yaml", "exec", "")...)
+	res.Assert(t, icmd.Expected{Out: "serviceA"})
+	res.Assert(t, icmd.Expected{Out: "serviceB"})
+
 	c.RunDockerComposeCmd(t, "-f", "./fixtures/env_file/compose.yaml", "down")
```
docker-agent, line 64: **[low] No e2e test for watch tolerating missing env_file on unrelated service**

The PR title and description explicitly name `watch` as one of the three commands fixed, and `cmd/compose/watch.go` gets the same `WithoutEnvironmentResolution` + `WithServicesEnvironmentResolved` treatment as `scale`. However, `TestUnusedMissingEnvFile` adds coverage for `scale` and shell completion but not for `watch`.

A minimal watch-specific check would be something like:

```go
// watch should work even with missing env file on a service not being watched
// (requires --no-up to avoid a blocking daemon; or check exit code directly)
res = c.RunDockerComposeCmdNoCheck(t, "-f", "./fixtures/env_file/compose.yaml", "watch", "--no-up", "serviceA")
// watch starts a long-running daemon, so just assert it doesn't immediately fail
// with the env_file error
res.Assert(t, icmd.Expected{ExitCode: 0})
```

Without this, a future regression (e.g. removing `WithoutEnvironmentResolution` from `watch.go`) would not be caught by the test suite.

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 100/100 |

docker-agent (commented):
### Assessment: 🟢 APPROVE

The fix correctly mirrors the `WithServices` pattern already used in `cmd/compose/compose.go`, `cmd/compose/run.go`, and `cmd/compose/config.go`: load with `cli.WithoutEnvironmentResolution` first (so the project is filtered to the selected services), then call `project.WithServicesEnvironmentResolved(true)` on the reduced set. This ensures only the env_files of targeted services are validated. Shell completion correctly skips environment resolution entirely since only service names are needed. The `completeScaleArgs` rename from `cli` → `dockerCli` avoids a name collision with the newly imported package. One minor test-coverage gap noted inline.

_None._

ndeloof (approved):

PR #13975, opened 2026-07-28
chore(deps): github.com/docker/buildx v0.36.0, buildkit v0.32.0

glours:
**What I did**
Test the upcoming buildx and BuildKit releases ahead of GA. buildx v0.36.0 requires Go >= 1.26.3, so the go directive moves from 1.25.9 to 1.26.3.

**Related issue**
N/A


docker-agent (commented):
### Assessment: 🟢 APPROVE

This PR bumps `github.com/docker/buildx` to v0.36.0-rc2, `github.com/moby/buildkit` to v0.32.0-rc2, and advances the `go` directive from 1.25.9 to 1.26.3. The changes to `go.mod` and `go.sum` are mechanically consistent with the dependency upgrades — updated direct deps, new/updated transitives, and matching hash entries in `go.sum`. No bugs were identified in the introduced changes.

thaJeztah:
Failure looks consistent; could be changes in containerd code? Or is it a test-registry not accepting some content?

```
url="http://localhost:32768/v2/test/manifests/sha256:492f68f4767a11b76825698c289e3519ba52bfeb04b0c95fe7e9f12422ce957d"
        time="2026-07-28T08:22:08Z" level=debug msg="fetch response received" host="localhost:32768" method=HEAD response.header.content-length=122 response.header.content-type=application/json response.header.date="Tue, 28 Jul 2026 08:22:08 GMT" response.header.docker-distribution-api-version=registry/2.0 response.status="500 Internal Server Error" spanID=b3a91e5b2268295f traceID=e3137962f4fbc1ba30eb878a2c86e732 url="http://localhost:32768/v2/test/manifests/sha256:492f68f4767a11b76825698c289e3519ba52bfeb04b0c95fe7e9f12422ce957d"
        time="2026-07-28T08:22:08Z" level=info msg="fetch failed after status: 500 Internal Server Error" host="localhost:32768" method=HEAD spanID=b3a91e5b2268295f traceID=e3137962f4fbc1ba30eb878a2c86e732 url="http://localhost:32768/v2/test/manifests/sha256:492f68f4767a11b76825698c289e3519ba52bfeb04b0c95fe7e9f12422ce957d"
        unexpected status from HEAD request to http://localhost:32768/v2/test/manifests/sha256:492f68f4767a11b76825698c289e3519ba52bfeb04b0c95fe7e9f12422ce957d: 500 Internal Server Error
        
        
        Failures:
        ExitCode was 1 expected 0
        Expected no error
    publish_test.go:166: 	[TestPublish] docker rm --force compose-e2e-publish-registry
```

internal/oci/resolver.go
```diff
@@ -95,6 +96,27 @@ func Get(ctx context.Context, resolver remotes.Resolver, ref reference.Named) (s
...
+	fetcher, err := resolver.Fetcher(ctx, ref.String())
+	if err != nil {
+		return nil, fmt.Errorf("creating fetcher for %s: %w", ref, err)
+	}
+	fetch, err := fetcher.Fetch(ctx, descriptor)
+	if err != nil {
+		return nil, fmt.Errorf("fetching blob %s: %w", descriptor.Digest, err)
+	}
+	defer func() { _ = fetch.Close() }()
+	content, err := io.ReadAll(fetch)
```
docker-agent: **[medium] `io.ReadAll` in `GetBlob` has no size bound — a rogue registry can cause unbounded memory allocation**

`GetBlob` reads the blob body with a bare `io.ReadAll(fetch)` without enforcing an upper bound against `descriptor.Size`. A malicious or misconfigured registry could stream far more bytes than the declared size, causing unbounded heap growth and potential OOM.

The same pattern exists in the pre-existing `Get` function (unchanged), but `GetBlob` is new public API that is called in a loop over all layers — amplifying the exposure.

Consider wrapping the reader:
```suggestion
	content, err := io.ReadAll(io.LimitReader(fetch, descriptor.Size+1))
```
After reading, verify the length: if `int64(len(content)) != descriptor.Size`, return an error.

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 57/100 |

docker-agent: **[medium] Blob content is not verified against `descriptor.Digest` after download**

`GetBlob` returns the raw bytes from the registry without computing and comparing the content hash against `descriptor.Digest`. If the registry (or a MITM) serves corrupted or substituted blob content, the mismatch goes undetected, and the caller (`pullComposeFiles`) writes the untrusted bytes directly to disk as a Compose file.

The same omission exists in the pre-existing `Get` function, but `GetBlob` is new code that is called for every layer in a loop. Whether containerd's `fetcher.Fetch` performs digest verification internally depends on the concrete resolver implementation — it is not guaranteed by the `remotes.Fetcher` interface contract.

Consider adding an explicit check after `io.ReadAll`:

```go
if got := digest.FromBytes(content); got != descriptor.Digest {
    return nil, fmt.Errorf("blob %s digest mismatch: expected %s, got %s", descriptor.Digest, descriptor.Digest, got)
}
```

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 57/100 |

docker-agent (commented):

Both findings are in the newly introduced `GetBlob` function in `internal/oci/resolver.go`. The dependency bumps and refactoring in `pullComposeFiles` are otherwise correct and well-structured. The blob-endpoint approach is the right fix for the manifests-vs-blobs endpoint confusion described in the PR.

internal/oci/resolver.go
```diff
@@ -95,6 +96,40 @@ func Get(ctx context.Context, resolver remotes.Resolver, ref reference.Named) (s
...
+	}
+	if int64(len(content)) != descriptor.Size {
+		return nil, fmt.Errorf("blob %s size mismatch: expected %d bytes, got %d", descriptor.Digest, descriptor.Size, len(content))
+	}
+	// GetBlob bypasses containerd's content store, so integrity must be
+	// checked here before callers write the bytes to disk.
+	if err := descriptor.Digest.Validate(); err != nil {
+		return nil, fmt.Errorf("invalid digest %s: %w", descriptor.Digest, err)
+	}
+	if actual := descriptor.Digest.Algorithm().FromBytes(content); actual != descriptor.Digest {
```
docker-agent, line 127: **[medium] Potential panic in digest verification if the algorithm is unregistered**

`descriptor.Digest.Validate()` (line 124) only checks that the digest string is syntactically well-formed (`<algorithm>:<hex>`); it does **not** verify that the algorithm is registered with the Go runtime. When `Algorithm().FromBytes(content)` is then called on an unrecognised algorithm, `opencontainers/go-digest` calls `Algorithm.Hash()` → `crypto.Hash.New()` on an unavailable hash, which **panics** with `"requested hash function is unavailable"`.

A rogue registry could craft a descriptor whose digest field contains a syntactically valid but unsupported algorithm name (e.g. `sha999:abcdef…`) that passes `Validate()` but triggers the panic here, crashing the Compose process.

**Suggested fix:** guard `FromBytes` with an availability check:

```go
if !descriptor.Digest.Algorithm().Available() {
    return nil, fmt.Errorf("unsupported digest algorithm %s", descriptor.Digest.Algorithm())
}
if actual := descriptor.Digest.Algorithm().FromBytes(content); actual != descriptor.Digest {
    return nil, fmt.Errorf("blob digest mismatch: expected %s, got %s", descriptor.Digest, actual)
}
```

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 57/100 |

docker-agent (commented):

#### Dismissed security findings (review manually)
- internal/oci/resolver.go:115 — Integer overflow in `io.LimitReader` when `descriptor.Size` is `math.MaxInt64` (verifier mitigation: overflow produces 0-byte read, and the size-mismatch check at line 119 immediately returns an error — no silent bypass; impact is at most an unexpected error, not data corruption)

internal/oci/resolver.go
glours: Verified manually: if a registry advertises `descriptor.Size == math.MaxInt64`, the `Size+1` overflow yields a negative `io.LimitReader` limit, so the read returns 0 bytes and the size-mismatch check right below returns an error — the request fails cleanly, no silent bypass or corruption, so no change needed.

internal/oci/resolver.go
```diff
@@ -95,6 +96,40 @@ func Get(ctx context.Context, resolver remotes.Resolver, ref reference.Named) (s
...
+		return nil, fmt.Errorf("creating fetcher for %s: %w", ref, err)
+	}
+	fetch, err := fetcher.Fetch(ctx, descriptor)
+	if err != nil {
+		return nil, fmt.Errorf("fetching blob %s: %w", descriptor.Digest, err)
+	}
+	defer func() { _ = fetch.Close() }()
+	// bound the read by the declared size so a rogue registry can't cause
+	// unbounded allocation; the extra byte detects oversized responses.
+	content, err := io.ReadAll(io.LimitReader(fetch, descriptor.Size+1))
```
docker-agent, line 115: **[low] Integer overflow in `LimitReader` bound when `descriptor.Size == math.MaxInt64`**

`descriptor.Size+1` overflows to `math.MinInt64` when `descriptor.Size` is `math.MaxInt64`. `io.LimitReader` with a negative limit reads 0 bytes immediately, so the size mismatch check fires and the error message says "got 0 bytes" instead of OOM — but the OOM guard intent is defeated for any descriptor claiming exactly `math.MaxInt64` bytes. A safe fix caps the limit before adding 1:

```suggestion
	limit := descriptor.Size
	if limit < math.MaxInt64 {
		limit++
	}
	content, err := io.ReadAll(io.LimitReader(fetch, limit))
```

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 100/100 |

docker-agent (commented):
### Assessment: 🟢 APPROVE

The PR is a routine dependency bump (buildx v0.36.0-rc2, BuildKit v0.32.0-rc2, Go 1.25.9→1.26.3) plus a new `GetBlob` helper in `internal/oci/resolver.go` that fetches artifact layers directly from the blobs endpoint. The dependency changes look correct and the new function is well-structured with size-bounding and digest verification. One low-severity edge case was found.

ndeloof (dismissed):

crazy-max (approved):

PR #13978, opened 2026-07-28
build: refactor Dockerfile

masontikhonov:
This refactors Dockerfile to inline Docker version: this is required to be compatible with Dependabot.

This also enables Dependabot and supportive workflows for service version bump.


masontikhonov:
Closing: open by mistake.

PR #13983, opened 2026-07-30
bump compose-go to version v2.14.0

glours:
**What I did**
Bump `compose-go` to version `v2.14.0`
 
**Related issue**
N/A

image

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,7 +8,7 @@ require (
 	github.com/Microsoft/go-winio v0.6.2
 	github.com/acarl005/stripansi v0.0.0-20180116102854-5a71ef0e047d
 	github.com/buger/goterm v1.0.4
-	github.com/compose-spec/compose-go/v2 v2.13.1-0.20260721083329-e70e961c3a3a
+	github.com/compose-spec/compose-go/v2 v2.14.0
 	github.com/containerd/console v1.0.5
 	github.com/containerd/containerd/v2 v2.2.5
 	github.com/containerd/errdefs v1.0.0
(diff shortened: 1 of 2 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

thaJeztah:
diff; https://github.com/compose-spec/compose-go/compare/e70e961c3a3a...v2.14.0

thaJeztah:
Not a blocker, but perhaps that `joinErrors` should match `errors.Join`; it currently ignores empty slices, but does not check for nil-errors (which `errors.Join` does), so it could panic if the slice of errors would contain a nil error;

```go
func joinErrors(errs []error) error {
	if len(errs) == 0 {
		return nil
	}
	slices.SortStableFunc(errs, func(a, b error) int {
		return slices.Compare(errorPath(a), errorPath(b))
	})
	return errors.Join(errs...)
}
```

This would panic (`err.Error()`);

```go
func errorPath(err error) []string {
	var pe pathError
	if errors.As(err, &pe) {
		return pe.path.Parts()
	}
	return []string{err.Error()}
}
```

(possibly the `pathError` error-type could also provide a `Cmp` helper for the sorting); wondering though if we should look at an exported error-type (I think BuildKit has something for the Dockerfile parsing errors), that could point to the path and line the error occurred in the compose-file (or are these different)?

thaJeztah (approved):
LGTM

thaJeztah:
> (possibly the pathError error-type could also provide a Cmp helper for the sorting)

Mostly thinking here to keep sorting and joining separate. The `joinErrors` makes it feel too much like a generic utility, but it should only be used for these specific cases, so making it transparent "sort path errors before joining" also makes it  more transparent that we're doing something special;

```go
func sortPathErrors(errs []error) {
	slices.SortStableFunc(errs, func(a, b error) int {
		return slices.Compare(errorPath(a), errorPath(b))
	})
}
```

:point_up: should have some guards for empty slices and nil errors.

Then sort before joining;

```go
sortPathErrors(errs)
return out, errors.Join(errs...)
```

ndeloof (approved):

PR #13988, opened 2026-07-31
Handle destroy events in up monitor

### LINKED ISSUE #13985: [BUG] Attached `up` never exits after external `stop`+`down` when a container is in restart backoff (regression in v2.39.3)
### Description

Since Compose v2.39.3, an attached `docker compose up` never exits when the project is stopped externally (`docker compose stop` + `docker compose down` from another process) while one of the containers is in restart backoff (`Status=restarting` with `restart: unless-stopped`).

Up to v2.39.2 the attached `up` exited a few seconds after `down` completed. From v2.39.3 on it hangs forever, even though all containers are stopped and removed. The behavior is still present in v5.3.1 (latest at the time of writing).

The apparent mechanism, observed via `docker events` during the stop window:

- a container that is stopped while it is in restart backoff emits only `stop` → `destroy` events — there is no `die` event, because no process is running at that moment;
- containers that are running normally when stopped emit the full `kill` → `stop` → `die` → `destroy` sequence;
- since v2.39.3 (#13210), the attached `up` appears to wait for a `die` event per container to detect termination, so it waits forever for the container that never emits one.

This matters for supervisors that spawn an attached `up` as a child process and stop the project externally: the leftover `up` process keeps holding resources it inherited (in our case, `flock` file locks on the compose files taken by a wrapper), long after the project is gone.

### Steps To Reproduce

`docker-compose.yml`:

```yaml
services:
  crasher:
    image: busybox:latest
    command: ["false"]
    restart: unless-stopped
  sleeper:
    image: busybox:latest
    command: ["sleep", "3600"]
```

1. `docker compose -p repro up --remove-orphans --force-recreate &` (attached, in background)
2. Wait until the crasher is in restart backoff: `docker inspect repro-crasher-1 --format '{{.State.Status}} {{.RestartCount}}'` shows `restarting` with RestartCount >= 5
3. From another shell: `docker compose -p repro stop -t 5 && docker compose -p repro down -t 5`
4. Observe the backgrounded `up` process:
   - v2.39.2 and earlier: exits within a few seconds after `down`
   - v2.39.3 and later (incl. v5.0.2, v5.3.1): never exits (waited 120+ s)

Bisect over official release binaries (same engine, same procedure, PASS = up exits / FAIL = up hangs):

| version | result |
|---|---|
| v2.15.1 | PASS |
| v2.26.0 | PASS |
| v2.33.1 | PASS |
| v2.37.3 | PASS |
| v2.38.2 | PASS |
| v2.39.0 | PASS |
| v2.39.1 | PASS |
| v2.39.2 | PASS |
| v2.39.3 | **FAIL** (first bad release) |
| v2.39.4 | FAIL |
| v5.0.2 | FAIL |
| v5.3.1 | FAIL |

We also verified on a separate machine (same engine v29.3.0) that swapping only the compose CLI plugin between v5.0.2 and v2.38.2 toggles the hang with an identical `docker events` sequence, so the regression is in compose, not in the engine.

### Compose Version

```Text
Docker Compose version v5.0.2
(also reproduced with v2.39.3, v2.39.4, v5.3.1; not reproducible with v2.39.2 and earlier)
```

### Docker Environment

```Text
Reproduced on:
- Docker Engine 29.6.2 (amd64, WSL2) — bisect environment
- Docker Engine 29.3.0 (amd64, embedded Linux device) — production environment
Client: docker compose CLI plugin, official release binaries
```

### Anything else?

The first bad release v2.39.3 contains #13210 ("use events to detect container exit"), which matches the observed mechanism (no `die` event is emitted for a container stopped during restart backoff).

Possibly related: moby/moby#50908 (open) adds a `restarting` attribute to `die` events for #13210's event-based detection. However, in this scenario the container emits no `die` event at all (only `stop` → `destroy`), so the attribute alone would not cover this case.

ousamabenyounes:
**What I did**
Handle container `destroy` events in the attached `up` monitor so `docker compose up` can finish when an externally stopped project removes a container that never emitted a `die` event.

The monitor still uses `die` for exit status and restart detection; `destroy` only removes the container from the tracked set so the monitor can terminate when the project is gone.

**Related issue**
Fixes #13985

## Test verification (RED -> GREEN)

RED on `origin/main` with the new regression test only:

```text
=== RUN   TestMonitorStopsAfterDestroyWithoutDie
    monitor_test.go:73: monitor did not stop after all containers were destroyed
--- FAIL: TestMonitorStopsAfterDestroyWithoutDie (1.00s)
FAIL
FAIL	github.com/docker/compose/v5/pkg/compose	1.026s
```

GREEN on this branch:

```text
=== RUN   TestMonitorStopsAfterDestroyWithoutDie
--- PASS: TestMonitorStopsAfterDestroyWithoutDie (0.00s)
PASS
ok  	github.com/docker/compose/v5/pkg/compose	0.023s
```

Full local validation:

```text
./run-ci.sh
DONE 574 tests, 2 skipped
make lint: golangci-lint run --build-tags "e2e" ./... -> 0 issues
```

### DIFF
--- a/pkg/compose/monitor.go
+++ b/pkg/compose/monitor.go
@@ -159,6 +159,10 @@ func (c *monitor) Start(ctx context.Context) error {
 					}
 					containers.Remove(ctr.ID)
 				}
+			case events.ActionDestroy:
+				logrus.Debugf("container %s destroyed", ctr.Name)
+				restarting.Remove(ctr.ID)
+				containers.Remove(ctr.ID)
 			}
 		}
 	}
--- a/pkg/compose/monitor_test.go
+++ b/pkg/compose/monitor_test.go
@@ -1,0 +1,87 @@
+/*
+   Copyright 2026 Docker Compose CLI authors
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package compose
+
+import (
+	"strings"
+	"testing"
+	"time"
+
+	containerType "github.com/moby/moby/api/types/container"
+	"github.com/moby/moby/api/types/events"
+	"github.com/moby/moby/client"
+	"go.uber.org/mock/gomock"
+	"gotest.tools/v3/assert"
+
+	compose "github.com/docker/compose/v5/pkg/api"
+)
+
+const monitorCompletionTimeout = time.Second
+
+func TestMonitorStopsAfterDestroyWithoutDie(t *testing.T) {
+	mockCtrl := gomock.NewController(t)
+	defer mockCtrl.Finish()
+
+	apiClient, _ := prepareMocks(mockCtrl)
+	projectName := strings.ToLower(testProject)
+	eventsCh := make(chan events.Message)
+	errCh := make(chan error)
+
+	apiClient.EXPECT().ContainerList(t.Context(), client.ContainerListOptions{
+		All: true,
+		Filters: projectFilter(projectName).Add("label",
+			oneOffFilter(false),
+			compose.ConfigHashLabel,
+		),
+	}).Return(client.ContainerListResult{
+		Items: []containerType.Summary{
+			testContainer("crasher", "crasher-container", false),
+			testContainer("sleeper", "sleeper-container", false),
+		},
+	}, nil)
+	apiClient.EXPECT().Events(gomock.Any(), client.EventsListOptions{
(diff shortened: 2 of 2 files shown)


ndeloof:
Thanks a lot for digging into #13985 and for the clean regression test! Closing this one in favor of #13990 though — here is why.

Handling only `destroy` fixes the literal repro from the issue (`stop` **then** `down`), but not the underlying contract: an attached `up` must exit once all containers are **stopped**, not only once they are removed. I verified with the pre-regression v2.39.2 binary that an external `docker compose stop` alone (no `down`) makes the attached `up` exit within about a second. With the destroy-only approach, a container in restart backoff that is externally stopped emits only a `stop` event — no `die`, no `destroy` — so it never leaves the tracked set and `up` still hangs forever. I confirmed this empirically: the `stop`-alone e2e test in #14023 fails on this branch, and passes with #13990.

Unfortunately covering `stop` can't be a simple `containers.Remove(...)` either: during `watch` sync+restart (#13161) the engine emits `stop` then `start` on `ContainerRestart`, so a `stop` event has to be inspected to distinguish a transient stop from a terminal one — which is exactly what #13990 does. Its `destroy` handling matches yours, by the way.

Thanks again for the contribution!

PR #13990, opened 2026-07-31
fix: detect externally stopped and removed containers in up monitor

### LINKED ISSUE #13985: [BUG] Attached `up` never exits after external `stop`+`down` when a container is in restart backoff (regression in v2.39.3)

glours:
**What I did**
Since 2.39.3 the monitor relies solely on die events to stop tracking containers, but a container stopped while in restart backoff emits only stop and destroy events, so an attached `up` hangs forever after an external `stop`/`down`. Handle destroy as terminal, and on stop inspect the container to distinguish a definitive stop from the transient one emitted during a ContainerRestart, which keeps watch sync+restart (#13161) working.

**Related issue**
Fixes #13985

image

### DIFF
--- a/pkg/compose/monitor.go
+++ b/pkg/compose/monitor.go
@@ -133,19 +133,40 @@ func (c *monitor) Start(ctx context.Context) error {
 					listener(newContainerEvent(event.TimeNano, ctr, api.ContainerEventRestarted))
 				}
 				logrus.Debugf("container %s restarted", ctr.Name)
-			case events.ActionDie:
-				logrus.Debugf("container %s exited with code %d", ctr.Name, ctr.ExitCode)
-				inspect, err := c.apiClient.ContainerInspect(ctx, event.Actor.ID, client.ContainerInspectOptions{})
-				if errdefs.IsNotFound(err) {
-					// Source is already removed
-				} else if err != nil {
+			case events.ActionStop:
+				// a stop event with no following start means the container won't come back:
+				// either it has no restart policy, or an external `stop`/`down` canceled the
+				// restart loop of a container in backoff (https://github.com/docker/compose/issues/13985).
+				// The event alone can't tell us: during a ContainerRestart (watch sync+restart,
+				// https://github.com/docker/compose/issues/13161) the engine also emits `stop`
+				// before `start`.
+				willRestart, err := c.isRestarting(ctx, ctr.ID)
+				if err != nil {
 					return err
 				}
-
-				if inspect.Container.State != nil && (inspect.Container.State.Restarting || inspect.Container.State.Running) {
-					// State.Restarting is set by engine when container is configured to restart on exit
-					// on ContainerRestart it doesn't (see https://github.com/moby/moby/issues/45538)
-					// container state still is reported as "running"
+				if willRestart {
+					logrus.Debugf("container %s stopped, restart in progress", ctr.Name)
+					restarting.Add(ctr.ID)
+				} else {
+					// definitive stop: the exit was already reported to listeners by the
+					// preceding die event, just stop tracking the container
+					logrus.Debugf("container %s stopped", ctr.Name)
+					restarting.Remove(ctr.ID)
+					containers.Remove(ctr.ID)
+				}
+			case events.ActionDestroy:
+				// container removed (e.g. by an external `docker compose down`): terminal
+				// state, there is nothing left to inspect
+				logrus.Debugf("container %s destroyed", ctr.Name)
+				restarting.Remove(ctr.ID)
+				containers.Remove(ctr.ID)
+			case events.ActionDie:
+				logrus.Debugf("container %s exited with code %d", ctr.Name, ctr.ExitCode)
+				willRestart, err := c.isRestarting(ctx, ctr.ID)
+				if err != nil {
+					return err
+				}
+				if willRestart {
 					logrus.Debugf("container %s is restarting", ctr.Name)
 					restarting.Add(ctr.ID)
 					for _, listener := range c.listeners {
@@ -162,6 +183,23 @@ func (c *monitor) Start(ctx context.Context) error {
 			}
 		}
 	}
+}
+
+// isRestarting tells whether a container which just stopped is expected to come back.
+// State.Restarting is set by the engine when the container is configured to restart on
+// exit, but not on a ContainerRestart, where state still is reported as "running"
+// (see https://github.com/moby/moby/issues/45538). A container already removed won't
+// come back.
+func (c *monitor) isRestarting(ctx context.Context, containerID string) (bool, error) {
+	inspect, err := c.apiClient.ContainerInspect(ctx, containerID, client.ContainerInspectOptions{})
+	if errdefs.IsNotFound(err) {
+		return false, nil
+	}
+	if err != nil {
+		return false, err
+	}
+	state := inspect.Container.State
+	return state != nil && (state.Restarting || state.Running), nil
 }
 
 func newContainerEvent(timeNano int64, ctr *api.ContainerSummary, eventType int, opts ...func(e *api.ContainerEvent)) api.ContainerEvent {
--- a/pkg/compose/monitor_test.go
+++ b/pkg/compose/monitor_test.go
@@ -1,0 +1,209 @@
+/*
+   Copyright 2026 Docker Compose CLI authors
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package compose
+
+import (
+	"strconv"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/containerd/errdefs"
+	"github.com/moby/moby/api/types/container"
+	"github.com/moby/moby/api/types/events"
+	"github.com/moby/moby/client"
+	"go.uber.org/goleak"
+	"go.uber.org/mock/gomock"
+	"gotest.tools/v3/assert"
+
+	"github.com/docker/compose/v5/pkg/api"
+	"github.com/docker/compose/v5/pkg/mocks"
+)
+
+// monitorEvent builds an engine event for container "123"/service1, with the
+// Actor.Attributes shape reported by the engine: compose labels plus the
+// container name.
+func monitorEvent(action events.Action) events.Message {
+	attrs := containerLabels("service1", false)
+	attrs["name"] = "testproject-service1-1"
+	return events.Message{
+		Type:   events.ContainerEventType,
+		Action: action,
+		Actor:  events.Actor{ID: "123", Attributes: attrs},
+	}
+}
+
+// monitorDieEvent builds a die event, which the engine reports with an exit code.
+func monitorDieEvent(exitCode int) events.Message {
+	event := monitorEvent(events.ActionDie)
+	event.Actor.Attributes["exitCode"] = strconv.Itoa(exitCode)
+	return event
+}
+
+// newMonitorTestFixture wires a monitor against a mocked API client, with the
+// goroutine-leak guard and the standard initial ContainerList expectation.
+func newMonitorTestFixture(t *testing.T) (*monitor, *mocks.MockAPIClient) {
+	t.Helper()
+	ignoreExisting := goleak.IgnoreCurrent()
+	t.Cleanup(func() {
+		goleak.VerifyNone(t, ignoreExisting)
+	})
+	mockCtrl := gomock.NewController(t)
+	t.Cleanup(mockCtrl.Finish)
+	apiMock := mocks.NewMockAPIClient(mockCtrl)
+
+	apiMock.EXPECT().ContainerList(gomock.Any(), gomock.Any()).
+		Return(client.ContainerListResult{Items: []container.Summary{testContainer("service1", "123", false)}}, nil)
+
+	m := newMonitor(apiMock, strings.ToLower(testProject))
+	return m, apiMock
+}
+
+// expectEvents makes the mocked engine deliver the given events, in order.
+func expectEvents(apiMock *mocks.MockAPIClient, msgs ...events.Message) {
+	ch := make(chan events.Message, len(msgs))
+	for _, msg := range msgs {
+		ch <- msg
+	}
+	apiMock.EXPECT().Events(gomock.Any(), gomock.Any()).
+		Return(client.EventsResult{Messages: ch, Err: make(chan error)})
+}
+
+// expectInspects makes successive inspections of container "123" report the
+// given states, in order.
+func expectInspects(apiMock *mocks.MockAPIClient, states ...container.State) {
+	calls := make([]any, 0, len(states))
+	for _, state := range states {
+		calls = append(calls, apiMock.EXPECT().
+			ContainerInspect(gomock.Any(), "123", client.ContainerInspectOptions{}).
+			Return(client.ContainerInspectResult{Container: container.InspectResponse{State: &state}}, nil))
+	}
+	gomock.InOrder(calls...)
+}
+
+// runMonitor starts the monitor under test in a goroutine and waits (with a
+// timeout) for it to return, reporting the events it published. It fails the
+// test if the monitor doesn't stop on its own, which is how an un-fixed
+// monitor.Start reacts to stop/destroy events it doesn't know how to process:
+// the tracked containers set never empties, so the loop blocks forever on the
+// events channel.
+func runMonitor(t *testing.T, m *monitor) ([]api.ContainerEvent, error) {
+	t.Helper()
+	var got []api.ContainerEvent
+	m.withListener(func(e api.ContainerEvent) {
+		got = append(got, e)
+	})
+
+	done := make(chan error, 1)
+	go func() {
+		done <- m.Start(t.Context())
+	}()
+	select {
+	case err := <-done:
+		return got, err
+	case <-time.After(10 * time.Second):
+		t.Fatal("monitor did not stop")
+		return nil, nil
+	}
+}
+
+// TestMonitorExitsOnDestroy pins the expectation that a destroy event (e.g. a
+// container removed by `docker rm` or `docker compose rm` outside of a
+// tracked lifecycle transition) drops the container from the tracked set
+// without requiring any inspection, so the monitor loop terminates.
+func TestMonitorExitsOnDestroy(t *testing.T) {
+	m, apiMock := newMonitorTestFixture(t)
+	expectEvents(apiMock, monitorEvent(events.ActionDestroy))
+
+	got, err := runMonitor(t, m)
+	assert.NilError(t, err)
+	assert.Equal(t, len(got), 0)
+}
+
+// TestMonitorExitsWhenRestartingContainerStopped is the #13985 repro: a
+// container configured to restart on failure dies (engine reports it as
+// still "restarting"), then is explicitly stopped (e.g. `docker stop`)
+// before the restart happens. The monitor must inspect on stop, observe the
+// container is no longer restarting/running, and evict it so the loop
+// terminates instead of waiting forever for a start event that never comes.
+func TestMonitorExitsWhenRestartingContainerStopped(t *testing.T) {
+	m, apiMock := newMonitorTestFixture(t)
+	expectEvents(apiMock, monitorDieEvent(1), monitorEvent(events.ActionStop))
+	expectInspects(apiMock,
+		// on die: waiting for the restart policy to kick in
+		container.State{Status: container.StateRestarting, Restarting: true, ExitCode: 1},
+		// on stop: the restart loop got canceled
+		container.State{Status: container.StateExited, ExitCode: 1},
+	)
+
+	got, err := runMonitor(t, m)
+	assert.NilError(t, err)
+	assert.Equal(t, len(got), 1)
+	assert.Equal(t, got[0].Type, api.ContainerEventExited)
+	assert.Equal(t, got[0].Restarting, true)
+	assert.Equal(t, got[0].ExitCode, 1)
+}
+
+// TestMonitorKeepsRunningOnRestart is the #13161 guard: a container that
+// dies and is restarted by the engine (watch/sync workflows trigger this via
+// `docker restart`) must not be evicted by an intervening stop event that is
+// merely part of the moby#45538 restart sequence (State reports
+// Running=true while mid-ContainerRestart). The monitor must keep tracking
+// it and still observe the subsequent start.
+func TestMonitorKeepsRunningOnRestart(t *testing.T) {
+	m, apiMock := newMonitorTestFixture(t)
+	expectEvents(apiMock,
+		monitorDieEvent(0),
+		monitorEvent(events.ActionStop),
+		monitorEvent(events.ActionStart),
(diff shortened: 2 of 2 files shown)


pkg/compose/monitor.go
```diff
@@ -133,19 +133,40 @@ func (c *monitor) Start(ctx context.Context) error {
...
+				}
+				if willRestart {
+					logrus.Debugf("container %s stopped, restart in progress", ctr.Name)
+					restarting.Add(ctr.ID)
+				} else {
+					// definitive stop: the exit was already reported to listeners by the
+					// preceding die event, just stop tracking the container
+					logrus.Debugf("container %s stopped", ctr.Name)
+					restarting.Remove(ctr.ID)
+					containers.Remove(ctr.ID)
```
docker-agent, line 155: **[medium] ActionStop "definitive stop" branch never fires `ContainerEventExited` to listeners**

The comment at line 151 says *"the exit was already reported to listeners by the preceding `die` event"*, but this is an assumed invariant rather than an enforced one. Docker Engine normally emits `die` → `stop` for clean exits, but edge cases exist where a `stop` event arrives without a prior `die`:

- A container that is OOM-killed during startup and immediately removed may emit only `stop` (no `die`).
- `docker compose down --no-stop` destroys a paused/created container that was never fully running.
- The new `TestMonitorStopInspectNotFound` test itself exercises exactly this path — a raw `stop` with no preceding `die` — and correctly expects zero listener events, but that may itself be a silent failure mode: the `up` watch loop never learns the container is gone.

When this happens, `containers.Remove(ctr.ID)` causes the monitor to exit (loop terminates on `len(containers) == 0`), but **no** `ContainerEventExited` is fired. Dependent service health-checkers or progress reporters in the `up` command may never receive the terminal event.

Consider firing `ContainerEventExited` unconditionally in the definitive-stop branch, or adding a guard that only skips the event when a prior `die` already recorded the exit (e.g. by checking `restarting.Has(ctr.ID)` as a proxy, since die-with-willRestart adds to `restarting`).

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 57/100 |

pkg/compose/monitor.go
```diff
@@ -133,19 +133,40 @@ func (c *monitor) Start(ctx context.Context) error {
...
 				}
 				logrus.Debugf("container %s restarted", ctr.Name)
+			case events.ActionStop:
+				// a stop event with no following start means the container won't come back:
+				// either it has no restart policy, or an external `stop`/`down` canceled the
+				// restart loop of a container in backoff (https://github.com/docker/compose/issues/13985).
+				// The event alone can't tell us: during a ContainerRestart (watch sync+restart,
+				// https://github.com/docker/compose/issues/13161) the engine also emits `stop`
+				// before `start`.
+				willRestart, err := c.isRestarting(ctx, ctr.ID)
```
docker-agent, line 143: **[medium] Any transient `ContainerInspect` error during `ActionStop` fatally kills the monitor**

The new `ActionStop` handler calls `isRestarting` (which makes a live Docker API call) and propagates any non-`NotFound` error via `return err`, tearing down the entire monitor goroutine:

```go
willRestart, err := c.isRestarting(ctx, ctr.ID)
if err != nil {
    return err   // kills compose up entirely
}
```

A transient daemon issue (momentary hiccup, daemon restart, network blip) during `ContainerInspect` will permanently terminate `docker compose up` rather than recovering. The same pattern in `ActionDie` is more defensible (the container has already exited and something is clearly wrong), but for a `stop` event — which may arrive while the container is mid-transition — the inspect is best-effort context gathering, not a hard requirement.

**Suggested approach**: On inspect error in the stop handler, log a warning and treat the container as definitively stopped (the conservative safe default), instead of aborting the monitor:

```suggestion
			willRestart, err := c.isRestarting(ctx, ctr.ID)
			if err != nil {
				logrus.Warnf("container %s: inspect on stop failed, treating as stopped: %v", ctr.Name, err)
				willRestart = false
			}
```

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 75/100 |

docker-agent (commented):

Two medium-confidence findings introduced by this PR's new `ActionStop` handling — both relate to how the monitor responds when `ContainerInspect` does or doesn't find the container.

- [low] pkg/compose/monitor.go:122 — `ActionStart` handler does not clear the `restarting` set (confidence: weak 10/100) — not a real bug given the current code flow; noted for future maintainability

pkg/compose/monitor.go
glours: Checked against a live daemon (29.5.3): none of these sequences exist — OOM emits `oom → die → destroy`, `docker stop` on a created-never-started container emits no event at all, and `down --no-stop` isn't a flag. Any container that ever ran got its exit reported to listeners by its `die` event, so there is no path where this branch silently drops a terminal event.

docker-agent: You're absolutely right, and thank you for testing this empirically. I incorrectly assumed edge cases that don't exist in practice — Docker Engine's event contract is more reliable than I gave it credit for. The OOM scenario I described emits `die` before `destroy`, created-never-started containers emit no events, and `--no-stop` isn't even a real flag.

The test I referenced (`TestMonitorStopInspectNotFound`) models a race condition where the container is already gone by the time `isRestarting` inspects it, not a case where `stop` arrives without a prior `die`. The zero-event expectation is correct: by the time the monitor sees the `stop` event, the container no longer exists, so `isRestarting` returns `false, nil` and the branch correctly evicts it. No listener event is needed because the container was already removed externally.

The existing code is sound — the comment at line 151 reflects the actual daemon behavior, not an assumption. Apologies for the false alarm.

pkg/compose/monitor.go
glours: This mirrors the exact error semantics of the `ActionDie` case just below, so a transient daemon error is no more fatal than it already was. Defaulting to `willRestart = false` on error would be the risky option: a blip during a watch-triggered restart would evict the container and silently terminate `up` mid-restart (a quieter #13161).

docker-agent: You're right — I missed that both handlers use identical error semantics, and asymmetry would be worse than consistency. The key distinction I overlooked: in both cases, the restart status directly determines whether to evict the container from the tracked set, so guessing wrong on error is risky.

That said, the current approach still means any transient inspect failure during a stop or die event kills the entire monitor. Is there telemetry showing how often daemon hiccups cause spurious `compose up` terminations in practice? If this trade-off is intentional (fail-stop beats silent mistrack), it might be worth documenting in a comment near `isRestarting` to explain why errors aren't swallowed.

docker-agent (commented):
### Assessment: 🟢 APPROVE

The fix is well-structured and correct. The new `ActionStop` handler correctly distinguishes a definitive stop from a transient one by inspecting the container state, and `ActionDestroy` unconditionally terminates tracking without requiring an inspect (appropriate since there is nothing left to inspect). The `isRestarting` helper cleanly handles the NotFound case (container already removed) by returning `false, nil`, which is the right terminal disposition. The test suite covers the four key scenarios: destroy-only, die+stop canceling a restart backoff (#13985 repro), die+stop+start across a ContainerRestart cycle (#13161 guard), and stop racing a removal (inspect returns NotFound).

ndeloof (dismissed):

ndeloof:
Considering the impact event management had in the past 😓 I'd like this new corner case is covered by an e2e test, based on reproduction scenario documented by https://github.com/docker/compose/issues/13985

pkg/compose/monitor.go
```diff
@@ -262,6 +263,55 @@ func (c *monitor) onContainerDie(ctx context.Context, event events.Message, ctr
...
+	}
+	if willRestart {
+		logrus.Debugf("container %s stopped, restart in progress", ctr.Name)
+		restarting.Add(ctr.ID)
+	} else {
+		// definitive stop: the exit was already reported to listeners by the
+		// preceding die event, just stop tracking the container
+		logrus.Debugf("container %s stopped", ctr.Name)
+		restarting.Remove(ctr.ID)
+		containers.Remove(ctr.ID)
```
docker-agent, line 285: **[medium] Final event consumers receive after external stop-of-backoff has `Restarting: true`, with no corrective follow-up**

In the `#13985` fix scenario (container in restart backoff, externally stopped), the event sequence is:

1. **`die`** → `onContainerDie`: `isRestarting` returns `true` → emits `ContainerEventExited{Restarting: true}`, keeps container tracked
2. **`stop`** → `onContainerStop`: `isRestarting` returns `false` → calls `restarting.Remove` + `containers.Remove`, **emits no event**, monitor loop ends

The only `ContainerEventExited` consumers ever see for this scenario carries `Restarting: true`. No follow-up event corrects it to `Restarting: false`. The code comment ("the exit was already reported to listeners by the preceding die event") is accurate, but that prior event's flag was set optimistically at die-time and is now stale.

The test `TestMonitorExitsWhenRestartingContainerStopped` encodes this shape (`got[0].Restarting == true`), which means callers that check the flag to decide whether to wait for a restart may behave incorrectly — they see "exiting, restart expected" and then nothing. For callers that rely solely on monitor termination (not on the event flag) to detect the definitive stop this is harmless, but it is a semantic mismatch worth noting.

**Fix**: pass the stop-event timestamp into `onContainerStop` and emit a corrective `ContainerEventExited{Restarting: false}` from the definitive-stop branch before evicting the container:

```go
// in the else branch of onContainerStop:
c.notify(newContainerEvent(stopTimeNano, ctr, api.ContainerEventExited))
restarting.Remove(ctr.ID)
containers.Remove(ctr.ID)
```

This also means the test assertion should flip to `got[1].Restarting == false` (two events: the die-time `Restarting: true` and the stop-time `Restarting: false`).

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 67/100 |

pkg/compose/monitor.go
```diff
@@ -238,17 +245,11 @@ func (c *monitor) onContainerRestart(event events.Message, ctr *api.ContainerSum
...
+	willRestart, err := c.isRestarting(ctx, ctr.ID)
+	if err != nil {
 		return err
 	}
-
-	if inspect.Container.State != nil && (inspect.Container.State.Restarting || inspect.Container.State.Running) {
-		// State.Restarting is set by engine when container is configured to restart on exit
-		// on ContainerRestart it doesn't (see https://github.com/moby/moby/issues/45538)
-		// container state still is reported as "running"
+	if willRestart {
```
docker-agent, line 252: **[low] `onContainerDie`'s non-restart path leaves a stale `restarting` entry when `onContainerStop` ran first**

The new `onContainerStop` handler (added by this PR) can add a container to `restarting` when `isRestarting` returns `true` (i.e., `State.Running == true`, mid-exit during a normal `docker stop`). If this `stop` event arrives before the corresponding `die` event:

1. **`stop`** → `onContainerStop`: `Running=true` seen → `restarting.Add(ctr.ID)`
2. **`die`** → `onContainerDie`: `Exited` seen → `willRestart=false` → calls `containers.Remove(ctr.ID)` **but NOT `restarting.Remove(ctr.ID)`**

The container is removed from `containers` (so the monitor eventually terminates), but the `restarting` set retains a stale entry. In the current session this is harmless — once the container leaves `containers`, no future `start` event for it is expected. However if the container is externally re-created with the same ID before the monitor session ends, `onContainerStart` would see `restarting.Has(ctr.ID) == true` and emit `ContainerEventStarted{Restarting: true}` incorrectly (suggesting it is recovering from a monitored crash rather than being a fresh start).

**Fix**: add `restarting.Remove(ctr.ID)` to `onContainerDie`'s non-restart branch (alongside the existing `containers.Remove`):

```go
// in onContainerDie, willRestart=false branch:
c.notify(newContainerEvent(event.TimeNano, ctr, api.ContainerEventExited))
restarting.Remove(ctr.ID)   // clean up any entry added by a preceding stop event
containers.Remove(ctr.ID)
return nil
```

This is a defensive cleanup that makes the two paths (`stop`-first and `die`-first) symmetric.

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 57/100 |

docker-agent (commented):

This PR correctly fixes the core `#13985` issue — a container in restart backoff that is externally stopped no longer hangs `up` forever. Two semantic edge cases in the new event-handling logic are worth reviewing:

1. The final event consumers receive in the stop-of-backoff scenario carries `Restarting: true` with no corrective follow-up
2. A `stop`-before-`die` race on a normal container stop can leave a stale entry in the `restarting` set

ndeloof (commented):
The event-sequence map in monitor.go's header (daemon-verified orderings) makes this design genuinely auditable, the #13985 repro test with the backoff-window de-risking in e2e is careful work, and `-race` is green locally. One ordering hole blocks approval, and it converges with the two docker-agent threads still open:

**[medium] stop-before-die on the last container loses the final `Exited` event — and with it the exit code.** The header itself notes that `stop` and `die` have no guaranteed relative order for an external `docker stop` of a running container. When `stop` lands first: `onContainerStop` inspects (state exited) → definitive branch → `containers.Remove`; if that was the last container, the loop's `len(containers) == 0 → return nil` runs **before draining** the `die` already queued in `res.Messages` (the emptiness check precedes the channel read). Consequence: `--abort-on-container-exit` and `--exit-code-from` never see the `Exited` — concretely, `up --exit-code-from app` on a single service with an external `docker stop` exits 0 instead of the container's code whenever the race lands stop-first. Two workable shapes: only evict on a definitive stop when a `die` was already observed for that ID (the #13985 scenario had one — this fixes the hang without opening the race), or emit the `Exited` from the definitive branch (the inspect carries the code) with dedup — which would also close docker-agent's open [medium] about the stale `Restarting: true` on the last event.

The two open docker-agent threads (stale `Restarting` flag; `restarting`-set ghost entry on stop-before-die of a normal shutdown, a one-line symmetry fix) both look legitimate and unanswered — worth folding into the same revision.

**[nit]** `onContainerDestroy`'s doc says "external `docker compose down`" — it's equally `docker rm -f`, which is the interesting case during a restart backoff (the header documents that ordering; the handler comment could point back at it).

pkg/compose/monitor.go
```diff
@@ -258,10 +259,82 @@ func (c *monitor) onContainerDie(ctx context.Context, event events.Message, ctr
...
+		logrus.Debugf("container %s stopped, restart in progress", ctr.Name)
+		restarting.Add(ctr.ID)
+	case restarting.Has(ctr.ID):
+		// definitive stop, cancelling a restart backoff: correct the
+		// optimistic Restarting: true the earlier die reported, an external
+		// stop just cancelled that restart for good. Skip the correction
+		// when the container is also already gone (found is false): there's
+		// no exit code left to report.
+		logrus.Debugf("container %s stopped", ctr.Name)
+		if found {
```
docker-agent, line 298: **[medium] No corrective `ContainerEventExited` emitted when restarting container is removed before the stop-inspect completes**

When a container in restart backoff is stopped and externally removed (e.g. `docker compose down` completing its rm before this inspect runs), the `case restarting.Has(ctr.ID):` branch correctly calls `restarting.Remove`/`containers.Remove` so the monitor terminates — but it skips emitting the corrective `ContainerEventExited{Restarting:false}` notification because `found == false`. Listeners therefore only ever see the earlier `ContainerEventExited{Restarting:true}` from `onContainerDie`, with no final correction to `Restarting:false`.

This means features that consume the notification stream (e.g. `--abort-on-container-exit`, `--exit-code-from`) are left with the container's last known state being `Restarting:true`, even though it was definitively stopped. The `up` process itself exits cleanly (the `containers` set empties), so there is no hang. The impact is limited to how the terminal exit status is surfaced.

A simple mitigation is to emit the corrective event even when `found == false`, using the exit code from the prior die event (it is available in `ctr.ExitCode` populated from the die event's attributes):

```suggestion
		if found {
			ctr.ExitCode = exitCode
		}
		c.notify(newContainerEvent(event.TimeNano, ctr, api.ContainerEventExited))
```

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 70/100 |

pkg/compose/monitor.go
```diff
@@ -258,10 +259,82 @@ func (c *monitor) onContainerDie(ctx context.Context, event events.Message, ctr
...
+// container reaching a stop event has necessarily run, so it always gets a
+// die (see the Start doc's event catalogue); only restarting.Has(ctr.ID)
+// proves that die was already processed here, so it's the sole eviction
+// signal — even a NotFound inspect (the container's removal outran this
+// inspect, racing the still-unread die right behind this stop) defers to
+// that pending die rather than evicting on `!found` alone.
+func (c *monitor) onContainerStop(ctx context.Context, event events.Message, ctr *api.ContainerSummary, containers, restarting utils.Set[string]) error {
+	willRestart, found, exitCode, err := c.isRestarting(ctx, ctr.ID)
+	if err != nil {
+		return err
```
docker-agent, line 285: **[medium] A transient `ContainerInspect` error in the new `onContainerStop` path fatally terminates the monitor**

`onContainerStop` now calls `isRestarting` (a `ContainerInspect` RPC) and propagates any non-NotFound error back to the `Start` loop, which returns it and aborts the entire `up` session. This pattern already existed in `onContainerDie`, so the PR is internally consistent — but it adds one more event type (every `stop` event) where a transient daemon HTTP-500 or timeout would kill an otherwise healthy `up`.

In practice the daemon is rarely flaky during a `stop`, and the equivalent risk already exists for `die` events. Still, a simple guard could limit the blast radius: treat an inspect error during `stop` as "state unknown — defer to the pending `die`" (i.e. return `nil` instead of the error), matching the NotFound semantics already applied in `isRestarting`. This would be consistent with the documented guarantee that every container that has run will always emit a die event, so the definitive state will be resolved there regardless.

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 62/100 |

docker-agent (commented):
### Assessment: 🟢 APPROVE

The fix correctly handles the two missing terminal-state events (`stop` during restart backoff and bare `destroy`) that caused `up` to hang forever on external stop/remove. The logic is well-documented and the test suite thoroughly covers the regression scenarios, including the subtle ordering cases (stop-before-die, die-then-stop, ContainerRestart interleave). Two low-impact observations are noted inline.

pkg/compose/monitor.go
glours: Thanks for flagging this — the gap itself is real (no notification when `restarting.Has(ctr.ID)` and `!found`), but I don't think the impact or the suggested fix hold up:

- `captureExitCodeFrom` and `stopOnFirstExit` (up.go) don't read `event.Restarting` at all — they act on the *first* `ContainerEventExited` they see, which is the one `onContainerDie` already emitted with `Restarting: true`. A second, corrective event wouldn't change their behavior either way (`once` is already false by then).
- The only consumer of `Restarting` is `printer.HandleEvent`, purely for the log line text ("exited (restarting)" vs "exited"). So the actual gap is a missing status line in this narrow race (external stop racing an rm ahead of our inspect), not a wrong terminal exit status.
- The suggested fix (`ctr.ExitCode = exitCode` unconditionally) wouldn't produce the prior die event's exit code: `ctr` is rebuilt fresh from the `stop` event's own attributes (`getContainerSummary`), which carries no `exitCode`, and `isRestarting`'s `exitCode` return is only meaningful when `found` is true. Emitting anyway would report a fabricated `0`, not the real one.

Given the impact is cosmetic, I'd rather leave this as-is than emit a misleading exit code — let me know if I'm missing another consumer that does care about `Restarting`.

pkg/compose/monitor.go
glours: Fair point on the exposure, but `onContainerDie` already propagates any non-NotFound inspect error the same way (line 248-250) for the exact same risk, and that's been in place unaddressed. Treating it only in `onContainerStop` would be asymmetric for a failure mode that's already accepted elsewhere in this file, without new evidence that stop-time inspects are meaningfully flakier than die-time ones. I'd rather keep the two paths consistent than add a one-sided guard here.

docker-agent: You're absolutely right — I was proposing a targeted fix for `stop` when the same propagation already exists for `die` (line 248-250), which is inconsistent. If the project has accepted this failure mode for `die` events in production without it being a practical problem, applying a different policy just to `stop` doesn't make sense.

Keeping the two code paths symmetric is the better choice here. Withdrawn — thanks for the context on the existing behavior.

ndeloof (approved):
Independent pass confirms the fix: verified build, `go vet`, `golangci-lint`, and the full `pkg/compose` unit suite with `-race` all pass on the branch. The remaining known gap (no corrective `Restarting:false` when `onContainerStop`'s inspect races a removal) has already been discussed and is a reasonable, narrowly-scoped trade-off — cosmetic log-line impact only, no consumer relies on it. LGTM.

PR #13998, opened 2026-08-03
Use content digest for pulled service images

benjick:
**What I did**

`pullServiceImage` returned the pulled image's raw inspect ID, while `getImageSummaries` resolves already-local images through `contentDigest` (the platform image-manifest digest, introduced for #13636). Both values feed the `com.docker.compose.image` label that `mustRecreate` compares to detect image changes - so the two paths disagreeing made the first `up` after the pulling `up` see a phantom image change and recreate every container once, with no change to the project, the image, or the compose file.

Under the containerd image store a `tag@digest` image reference triggers this reliably: the pull path records the **index digest** (the raw inspect ID) on the container, and the next `up` resolves the same image to its **platform manifest digest** via `contentDigest`. The strings differ, `mustRecreate` fires, and the replacement container gets the platform digest - so it happens exactly once per container, which makes it easy to miss but breaks any workflow that relies on `up` being idempotent (we hit it because our orchestrator asserts unchanged services are never recreated across reconciles).

The fix makes `pullServiceImage` resolve the pulled image through the same manifests-aware inspect + `contentDigest` call `getImageSummaries` uses, so both sides of the staleness comparison speak the same scheme. Engines without manifest support (API < 1.48) keep the previous behavior via `contentDigest`'s plain-ID fallback.

**Repro**

```console
$ docker run -d --privileged --name repro -e DOCKER_TLS_CERTDIR= docker:dind   # 29.7.0, containerd store
$ cat repro.yml
services:
  app:
    image: alpine:3.19@sha256:6baf43584bcb78f2e5847d1de515f23499913ac9f12bdf834811a3145eb11ca1
# (compose v5.4.0 as the CLI plugin inside the dind)
$ docker compose -f repro.yml -p repro up -d   # pulls, creates container A
$ docker compose -f repro.yml -p repro up -d   # => "Recreated" - container replaced, nothing changed
$ docker compose -f repro.yml -p repro up -d   # stable from here on
```

The two containers' `com.docker.compose.image` labels show the mismatch: container A carries `sha256:6baf...` (the pinned index digest), its replacement carries the platform manifest digest. v5.3.1 does not exhibit this; it appeared when the local-inspect side moved to `contentDigest`.

**Verification**

- New unit test `TestPullServiceImageUsesContentDigest` fails against the old return value and passes with the fix.
- `go build ./...`, `go vet ./pkg/compose/`, `gofmt -s -l` clean; `go test ./pkg/compose/` passes.
- End-to-end against a fresh `docker:dind` (29.7.0): with this fix the container from the pulling `up` survives repeated `up` runs (same container ID); unpatched v5.4.0 recreates it on the second `up`.

Related: #13636 (the fix that introduced `contentDigest` for the local-inspect path; this aligns the pull path with it).

---
Disclosure: this PR was prepared by Claude Code (Anthropic) - the diagnosis, fix, tests, and the verification runs above - in an agentic session I directed. I'm the submitter and point of contact.

### DIFF
--- a/pkg/compose/images.go
+++ b/pkg/compose/images.go
@@ -185,6 +185,27 @@ func (s *composeService) manifestsSupported(ctx context.Context) (bool, error) {
 	return versions.GreaterThanOrEqualTo(version, apiVersion148), nil
 }
 
+// inspectContentDigest inspects ref, requesting per-manifest data on engines
+// that support it, and returns the digest identifying the image's runnable
+// content for the default platform. Callers that record an image identity
+// compose later compares for staleness must go through this, so every such
+// identity is computed the same way — see contentDigest.
+func (s *composeService) inspectContentDigest(ctx context.Context, ref string) (string, error) {
+	withManifests, err := s.manifestsSupported(ctx)
+	if err != nil {
+		return "", err
+	}
+	var opts []client.ImageInspectOption
+	if withManifests {
+		opts = append(opts, client.ImageInspectWithManifests(true))
+	}
+	inspected, err := s.apiClient().ImageInspect(ctx, ref, opts...)
+	if err != nil {
+		return "", err
+	}
+	return contentDigest(inspected.InspectResponse, platforms.Default()), nil
+}
+
 // contentDigest returns the digest identifying an image's runnable content
 // (config + layers) for the given platform. With BuildKit provenance
 // attestations enabled (the default since recent Buildx/BuildKit), the image is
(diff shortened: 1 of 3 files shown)


glours:
Hi @benjick, thanks for this fix and for the detailed analysis, the pull-path/local-path digest disagreement you identified was exactly right.

While reviewing it we realized it's one instance of a broader structural problem: compose records image identities through several independent code paths (pull, bake, classic builder, already-local inspect), and they don't all produce the same kind of digest for the same image. The same mismatch class caused #13636 and #14005, plus a few more latent cases we found along the way.

We've addressed the whole class in #14011: a single canonical content-digest producer shared by every path, plus a containerd-store e2e job in CI so this can't regress silently again. Your commit is included there with your authorship preserved (squashed with your lint follow-up), and your `up` idempotency scenario is now locked by a dedicated e2e test (`TestUpIdempotentContainerdStore`) that runs in CI against the containerd image store.

If that works for you, we'll close this PR in favor of #14011 , and feel free to review it there, your feedback would be very welcome. Thanks again!

benjick:
Yeah 100%!

Issue #12648, opened 2025-03-19
[BUG] `docker compose --profile <XYZ> down ` shut down all services


Fabccc:
### Description

According to the documentation found here https://docs.docker.com/compose/how-tos/profiles/#stop-specific-profiles, I should be able to down only specific services by using their profile. If they do not depends on anything, it should only down these services.

When i'm doing so, it shut down all services in the docker compose file.

### Steps To Reproduce

1. With this simple docker compose
```yaml
services:
  first:
    image: nginx
  second:
    image: nginx
    profiles:
      - my-profile
```
2. Execute the following commands
```shell
❯ docker compose up -d
  [+] Running 1/1
   ✔ Container docker-compose-issue-services-first-1  Started
❯ docker compose --profile my-profile up -d
  [+] Running 2/2
   ✔ Container docker-compose-issue-services-first-1   Running
   ✔ Container docker-compose-issue-services-second-1  Started
❯ docker compose --profile my-profile down
  [+] Running 3/3
   ✔ Container docker-compose-issue-services-second-1  Removed
   ✔ Container docker-compose-issue-services-first-1   Removed
   ✔ Network docker-compose-issue-services_default     Removed
```

It should only remove the service with the profile `my-profile` according to documentation.

### Compose Version

```Text
Docker Compose version v2.33.1-desktop.1
```

### Docker Environment

```Text
Client:
 Version:    28.0.1
 Context:    default
 Debug Mode: false
 Plugins:
  ai: Docker AI Agent - Ask Gordon (Docker Inc.)
    Version:  v0.9.8
    Path:     /home/fabcc/.docker/cli-plugins/docker-ai
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.21.1-desktop.2
    Path:     /usr/local/lib/docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.33.1-desktop.1
    Path:     /usr/local/lib/docker/cli-plugins/docker-compose
  debug: Get a shell into any image or container (Docker Inc.)
    Version:  0.0.38
    Path:     /usr/local/lib/docker/cli-plugins/docker-debug
  desktop: Docker Desktop commands (Beta) (Docker Inc.)
    Version:  v0.1.5
    Path:     /usr/local/lib/docker/cli-plugins/docker-desktop
  dev: Docker Dev Environments (Docker Inc.)
    Version:  v0.1.2
[...]

PR #14000, opened 2026-08-04
feat: add --profiles-only flag to stop and restart commands

### LINKED ISSUE #13993: direct command to compose down services a specific profile without down default services
## problem description

there’s no proper way to compose down only a profile’s services without specifying specific service names or getting default services down along as well.

i understand the current ideology of having default(core) services getting up along with any profile(s). Though now I want to keep my core services running but `compose down` my other profiles. the only way is to:
```bash
docker compose down <profile_service_name_1> <profile_service_name_2> …
```
as if you do 
```bash
docker compose --profile=xyz down
```
removes the default services along, which is expected as per docs: https://docs.docker.com/compose/how-tos/profiles/?uuid=46ec72e8-3598-42b7-876d-6ce34666bad9#stop-application-and-services-with-specific-profiles

however that’s inconvenient. we might nest command to get services names dynamically from config maybe. but hard to get around such complex command.

is there some ideal way. Or can we introduce something for this?

## proposed approaches & community discussion

after discussing on community forum:
https://forums.docker.com/t/direct-command-to-compose-up-down-a-profile-without-up-down-default/152172

Metin Y suggested an argument addition  `--profile-only` in the `compose down` action. 

i assume that's a little difficult as `--profile` argument is for `compose` and not of `up/down/logs/stop`. 
adding a new arg might be a heavy task.

another approach I can think of is firstly
allow multiple comma separated profile:
```bash
docker compose --profile="hi,bye" up
```

then we can introduce something like a tilde at the beginning to indicate only this profile(s):
```bash
docker compose --profile="~hi,bye" down
```
this may seem to be a bit confusing as well, there's a requirement for a convenient way.

## additional proposal: Container labels for profiles

additionally there should be proper and easier way to filter out the containers based on profile, this might help to at least get more convenient way to inspect and get information without having to have the knowledge of source path of config files.

`docker inspect <id>` already contains `com.docker.compose.project`, `com.docker.compose.project.config_files`, `com.docker.compose.project.working_dir` in `[0].Config.Labels` which I'm assuming are there for the similar reasons I'm proposing. having `com.docker.compose.profile` in the labels would be a great addition to filter out containers based on profile name like `docker (compose) ps --filter "label=com.docker.compose.profile=xyz"`.

> might be worth delegating to a sub issue/task and/or good first issue. lmk.

## current workarounds vs ideal commands

current way of getting down the services of a project for specified profiles without removing default is:
```bash
export PROJECT="myproject" PROFILE="abc,xyz"

export CFG=$(docker compose ls --format json | jq -r --arg p "$PROJECT" '.[]|select(.Name==$p)|.ConfigFiles' | tr ',' ':')

COMPOSE_FILE=$CFG COMPOSE_PROFILES=$PROFILE docker compose -p "$PROJECT" config --format json | jq -r --arg p "$PROFILE" '($p|split(",")) as $r | .services|to_entries[]|select(.value.profiles and any(.value.profiles[]; IN($r[])))|.key' | xargs -r docker compose -p "$PROJECT" down
```

if `docker ps --filter "label=com.docker.compose.profile=abc,xyz"` worked that could become:
```bash
PROJECT="myproject" PROFILE="HIII,THAI"; docker compose -p "$PROJECT" down $(docker ps -a --filter "label=com.docker.compose.project=$PROJECT" $(printf ' --filter label=com.docker.compose.profile=%s' ${PROFILE//,/ }) --format '{{.Label "com.docker.compose.service"}}' | sort -u)
```

and if `docker compose ps --filter "label=com.docker.compose.profile=abc,xyz"` worked that could become:
```bash
PROJECT="myproject" PROFILE="HIII,THAI"; docker compose -p "$PROJECT" down $(docker compose -p "$PROJECT" ps -a $(printf ' --filter label=com.docker.compose.profile=%s' ${PROFILE//,/ }) --format '{{.Label "com.docker.compose.service"}}' | sort -u)
```
even after shortening the command is not short enough

ideal or expecting something like:
```bash
docker compose -p <project> --profile-only=profile1,profile2 down
# or
docker compose -p <project> --profile=~profile1,profile2 down
```
clean and intuitive.

> ps note: the above commands are big cuz I run many projects on my pc(10+ projects, 6-7 services each) hard not to lose track, most with restarts and clashing ports. while default services are what i want to keep for my teammates, i want to down the other profile ones without actually going to that folder/location. vscode-containers extensions is no help cuz it contains no profile info. though i've customized the extension for me, but doesn't help much my team. and using ai agent is no help, cuz agents don't even know we can compose down specific services.

---

info on what i have:

```bash
h3110fr13nd@rog:~$ docker version
 Version:           29.7.1
 API version:       1.55
 Go version:        go1.26.5
 Git commit:        e9452d6
 Built:             Fri Jul 31 17:06:16 2026
 Context:           default

 Engine:
  Version:          29.7.1
  API version:      1.55 (minimum version 1.40)
  Go version:       go1.26.5
  Git commit:       c5b8ce9
  Built:            Fri Jul 31 17:06:16 2026
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          v2.2.6
  GitCommit:        11ce9d5f3c68c941867e82890e93e815c1304f1b
 runc:
  Version:          1.3.6
  GitCommit:        v1.3.6-0-g491b69ba
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

```bash
h3110fr13nd@rog:~$ docker context show
default
```

```bash
docker info
 Version:    29.7.1
 Context:    default
 Debug Mode: false
 Plugins:
    Version:  v0.36.0
    Version:  v5.3.1
    Version:  v1.2.6
    Path:     /usr/libexec/docker/cli-plugins/docker-model
    Version:  v1.10.0
    Path:     /home/h3110fr13nd/.docker/cli-plugins/docker-scout

Server:
 Containers: 8
  Running: 0
  Paused: 0
  Stopped: 8
 Images: 30
 Server Version: 29.7.1
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
 CDI spec directories:
  /etc/cdi
  /var/run/cdi
 Swarm: inactive
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 11ce9d5f3c68c941867e82890e93e815c1304f1b
 init version: de40ad0
 Security Options:
  apparmor
  seccomp
   Profile: builtin
  cgroupns
 Kernel Version: 7.0.0-28-generic
 Operating System: Ubuntu 26.04 LTS
 OSType: linux
 Architecture: x86_64
 CPUs: 16
 Total Memory: 14.38GiB
 Name: rog
 ID: 1fea1183-a226-49b9-8eb0-8e060ffa3fe7
 Debug Mode: false
 Experimental: false
 Insecure Registries:
  ::1/128
  127.0.0.0/8
 Live Restore Enabled: false
 Firewall Backend: iptables
  EnableUserlandProxy: true
```
---

glours:
**What I did**
Acting on only the services of a profile currently requires naming each service explicitly, as `--profile x stop` or `--profile x restart` also target services without profiles (#13993, previously #12648).

The new flag restricts both commands to services enabled by a profile: the active profiles when some are set via --profile or COMPOSE_PROFILES, otherwise services from all profiles. Services without profiles are never affected. `down` intentionally keeps its whole-application semantics.

**Related issue**
fixes #13993 

image

### DIFF
--- a/cmd/compose/profiles_only.go
+++ b/cmd/compose/profiles_only.go
@@ -1,0 +1,78 @@
+/*
+   Copyright 2020 Docker Compose CLI authors
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package compose
+
+import (
+	"errors"
+	"fmt"
+	"io"
+	"slices"
+	"strings"
+
+	"github.com/compose-spec/compose-go/v2/types"
+)
+
+// profilesOnlyServices resolves the service names targeted by the --profiles-only
+// flag: services enabled by one of the active profiles, or services belonging to
+// any profile when no profile is active. It returns the project to run the
+// command with, which has all profiles enabled when none was active. An empty
+// service list with a nil error means there is nothing to do.
+func profilesOnlyServices(project *types.Project, services []string, action string, w io.Writer) (*types.Project, []string, error) {
+	if len(services) > 0 {
+		return nil, nil, errors.New("--profiles-only cannot be combined with service names, naming a service already activates its profiles")
+	}
+	if project == nil {
+		return nil, nil, errors.New("--profiles-only requires the project's compose file(s), pass --file or run the command from the project directory")
+	}
+
+	// COMPOSE_PROFILES being unset yields a single blank profile name, filter
+	// such entries out so it is treated as "no active profile"
+	activeProfiles := slices.DeleteFunc(slices.Clone(project.Profiles), func(p string) bool {
+		return p == ""
+	})
+	if len(activeProfiles) == 0 {
+		var err error
+		project, err = project.WithProfiles([]string{"*"})
+		if err != nil {
+			return nil, nil, err
+		}
+	}
+
+	var names, allProfiles []string
+	for name, service := range project.Services {
+		if len(service.Profiles) == 0 {
+			continue
+		}
+		names = append(names, name)
+		allProfiles = append(allProfiles, service.Profiles...)
+	}
+	slices.Sort(names)
+	slices.Sort(allProfiles)
+	allProfiles = slices.Compact(allProfiles)
+
+	switch {
+	case len(names) == 0 && len(activeProfiles) > 0:
+		_, _ = fmt.Fprintf(w, "no services matched the active profiles [%s]\n", strings.Join(activeProfiles, " "))
+	case len(names) == 0:
+		_, _ = fmt.Fprintln(w, "no service in this project uses profiles")
+	case len(activeProfiles) > 0:
+		_, _ = fmt.Fprintf(w, "%s services in profiles [%s]\n", action, strings.Join(activeProfiles, " "))
+	default:
+		_, _ = fmt.Fprintf(w, "%s services from all profiles [%s] as none is active\n", action, strings.Join(allProfiles, " "))
+	}
+	return project, names, nil
+}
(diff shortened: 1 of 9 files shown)


glours:
Closing in favor of #14046 approach

PR #14003, opened 2026-08-05
fix(bridge): validate arguments of bridge subcommands

glours:
**What I did**
`docker compose bridge transformations create` panicked with an index out of range when invoked without its PATH argument, as RunE indexed args[0] without any validator. Add cli.ExactArgs(1) so the CLI reports a proper usage error instead.

Also add cobra.NoArgs to `bridge convert` and `bridge transformations list`, which silently ignored stray arguments.

**Related issue**
https://docker.atlassian.net/browse/DDB-612

image

### DIFF
--- a/cmd/compose/bridge.go
+++ b/cmd/compose/bridge.go
@@ -22,6 +22,7 @@ import (
 	"io"
 
 	"github.com/distribution/reference"
+	"github.com/docker/cli/cli"
 	"github.com/docker/cli/cli/command"
 	"github.com/docker/go-units"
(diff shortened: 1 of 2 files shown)


maxproske:
That was fast! :)

```sh
./bin/docker-compose bridge transformations create
# docker: 'docker compose bridge transformations create' requires 1 argument
# Usage:  docker compose bridge transformations create [OPTION] PATH
```

ndeloof (approved):

PR #14004, opened 2026-08-05
ci: grant actions:write to PR review workflow

glours:
**What I did**
docker-agent-action v2.0.3 raised the review job's required permission from actions:read to actions:write (cache delete for review-lock release, feedback artifact management). GitHub refuses to start a reusable workflow requesting more permissions than the caller grants, so every PR Review run since the v2.0.3 bump ended in startup_failure and docker-agent stopped launching automatically.

**Related issue**

Screenshot 2026-08-05 at 12 30 08

image

### DIFF
--- a/.github/workflows/pr-review.yml
+++ b/.github/workflows/pr-review.yml
@@ -26,7 +26,7 @@ jobs:
       issues: write # Create security incident issues if secrets detected
       checks: write # (Optional) Show review progress as a check run
       id-token: write # Required for OIDC authentication to AWS Secrets Manager
-      actions: read # Download artifacts from trigger workflow
+      actions: write # Download artifacts from trigger workflow; cache delete for review-lock release (required since docker-agent-action v2.0.3)
     with:
       trigger-run-id: ${{ github.event_name == 'workflow_run' && format('{0}',
         github.event.workflow_run.id) || '' }}


.github/workflows/pr-review.yml
```diff
@@ -26,7 +26,7 @@ jobs:
       issues: write # Create security incident issues if secrets detected
       checks: write # (Optional) Show review progress as a check run
       id-token: write # Required for OIDC authentication to AWS Secrets Manager
-      actions: read # Download artifacts from trigger workflow
+      actions: write # Download artifacts from trigger workflow; cache delete for review-lock release (required since docker-agent-action v2.0.3)
```
crazy-max, line 29: Looks good but reusable workflows should never define permissions: https://github.com/docker/docker-agent-action/blob/5396207026ca7d22d0a06a706260b7eb671c9346/.github/workflows/review-pr.yml#L86-L91

It should be only the caller to define them. More info:
* https://github.com/docker/github-builder/issues/24
* https://github.com/docker/packaging/pull/398#issue-4220380875
* https://github.com/docker/github-builder/pull/49

cc @derekmisler @Sayt-0

crazy-max (approved):

.github/workflows/pr-review.yml
derekmisler: > reusable workflows should never define permissions

I think reusable workflows should define the minimum permissions they need, following least-privilege.

derekmisler: It also causes consumers to fail fast/loud if they're missing required permissions. We've run into silent/annoying errors in the past.

crazy-max: > following least-privilege.

It doesn't work like that unfortunately: https://docs.github.com/en/actions/reference/workflows-and-actions/reusing-workflow-configurations#supported-keywords-for-jobs-that-call-a-reusable-workflow

> ... permissions passed from the caller workflow can be only downgraded (not elevated) by the called workflow.

For example in our reusable workflow we don't need `packages: write` for registries other than GHCR and if we define `packages: read` it cannot be elevated:

```
The nested job 'build' is requesting 'packages: write', but is only allowed 'packages: read'.
```

Not sure about docker/docker-agent-action pr-review workflow need all these write access?

So defining elevated permissions even if not needed feels over-scope

derekmisler: > Not sure about docker/docker-agent-action pr-review workflow need all these write access?

Yeah, we got to the permissions we have now incrementally as we ran into failures by consumers (I have like half a dozen of my own PRs just bumping individual permissions, similar to this PR).

> It doesn't work like that unfortunately: https://docs.github.com/en/actions/reference/workflows-and-actions/reusing-workflow-configurations#supported-keywords-for-jobs-that-call-a-reusable-workflow

That's interesting, I understood that doc as implying reusable workflows SHOULD define their own permissions but only to _downgrade,_ rather than _elevate._ Thank you for sharing that!

Regardless, all of this is moot once our AI reviewer is moved off of GHA and into our own infra, which is currently in progress.

PR #14010, opened 2026-08-06
fix(bridge): skip pulling default image references for build-only ser…

ericwyles:
This fixes LoadAdditionalResources so "build-only services" (those with a `build:` definition but no explicit `image:` reference) are not pulled from a registry or local Docker daemon.

I encountered this while developing a Compose Bridge transformation that converts Compose files into [Zarf](https://zarf.dev/) packages using [imageArchives](https://docs.zarf.dev/ref/components/#container-image-archives). The transformation uses Buildx to build images directly into OCI archive files, which Zarf then consumes; no image needs to be available in a registry or local Docker image store.

Previously, Compose derived a default image name for the build-only service and attempted to pull it before the transformation ran. This caused conversion to fail even though the image was going to be produced directly as an archive and there was nothing to pull.

image
Build-only services now cross the bridge without trying to pull a nonexistent image.

### DIFF
--- a/pkg/bridge/convert.go
+++ b/pkg/bridge/convert.go
@@ -169,8 +169,11 @@ func LoadAdditionalResources(ctx context.Context, dockerCLI command.Cli, project
 		var inspect image.InspectResponse
 		if service.Build != nil && service.Image == "" {
 			result, err := dockerCLI.Client().ImageInspect(ctx, imageName)
-			if err != nil && !errdefs.IsNotFound(err) {
-				return nil, err
+			if err != nil {
+				if !errdefs.IsNotFound(err) {
+					return nil, err
+				}
+				logrus.Warnf("image %s for service %s not found locally; Dockerfile-exposed ports will not be included — run `docker compose build` first to include them", imageName, name)
 			}
 			inspect = result.InspectResponse
 		} else {
--- a/pkg/bridge/convert.go
+++ b/pkg/bridge/convert.go
@@ -166,19 +166,31 @@ func LoadAdditionalResources(ctx context.Context, dockerCLI command.Cli, project
 	for name, service := range project.Services {
 		imageName := api.GetImageNameOrDefault(service, project.Name)
 
-		inspect, err := inspectWithPull(ctx, dockerCLI, imageName)
-		if err != nil {
-			return nil, err
+		var inspect image.InspectResponse
+		if service.Build != nil && service.Image == "" {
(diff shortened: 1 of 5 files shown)


pkg/bridge/convert.go
```diff
@@ -166,19 +166,31 @@ func LoadAdditionalResources(ctx context.Context, dockerCLI command.Cli, project
...
 
-		inspect, err := inspectWithPull(ctx, dockerCLI, imageName)
-		if err != nil {
-			return nil, err
+		var inspect image.InspectResponse
+		if service.Build != nil && service.Image == "" {
+			result, err := dockerCLI.Client().ImageInspect(ctx, imageName)
+			if err != nil && !errdefs.IsNotFound(err) {
+				return nil, err
+			}
```
glours:
```suggestion
			if err != nil {
				if !errdefs.IsNotFound(err) {
					return nil, err
				}
				logrus.Warnf("image %s for service %s not found locally; Dockerfile-exposed ports will not be included — run `docker compose build` first to include them", imageName, name)
			}
```

pkg/e2e/bridge_test.go
```diff
@@ -59,3 +60,17 @@ func TestConvertAndTransformList(t *testing.T) {
 		assert.Assert(t, strings.Contains(res.Stdout(), "docker/compose-bridge-kubernetes"), res.Combined())
 	})
 }
+
+func TestConvertBuildOnlyService(t *testing.T) {
+	c := NewParallelCLI(t)
+	outDir := t.TempDir()
+
+	res := c.RunDockerComposeCmd(t, "-f", "./fixtures/bridge-build-only/compose.yaml", "--project-name", "bridge-build-only", "bridge", "convert",
+		"--output", outDir, "--transformation", "docker/compose-bridge-kubernetes:v0.0.3")
```
glours: Maybe we could reuse `const bridgeImageVersion = "v0.0.3"` from line 33 and expose it at the package level?

glours (commented):
Hey @ericwyles 
Thanks for the contribution, maybe 2 nits to address before we merged it

pkg/e2e/bridge_test.go
ericwyles: Thanks @glours , good catch! This should be done now and I applied your other suggestion too.

glours (approved):
LGTM, thanks for the contribution 🙏

PR #14006, opened 2026-08-05
fix(build): resolve image volumes to a mountable name, not a manifest digest

ricardobranco777:
**What I did**

ImageSummary.ID holds the digest of the platform-specific manifest so ServiceHash stays stable across attested rebuilds (see contentDigest). resolveImageVolumes reused that same value as the `type: image` mount Source, but the daemon only resolves a mount Source by name/tag or top-level image ID, not by manifest digest — so `compose up` failed with "No such image" whenever the volume's source image was already present locally (always for a built image; on a second run for a pulled one).

Keep Source as the resolved image name, and track the digest separately via a new com.docker.compose.image-volume-digest label so mustRecreate can still detect a rebuilt/updated source image independently of Source.

**Related issue**

Fixes #14005

**(not mandatory) A picture of a cute animal, if possible in relation to what you did**

PR #14011, opened 2026-08-06
Fix image digest reconciliation

### LINKED ISSUE #14005: [BUG] type=image volumes fail with "No such image" when the source image is already present locally
### Description

`docker compose up` fails to create a container that mounts a `type: image` volume with `Error response from daemon: No such image: sha256:<digest>` even though the referenced image exists locally.

This was caught by openQA when running your tests and these failed: `TestImageVolume` & `TestImageVolumeRecreateOnRebuild`.

### Steps To Reproduce

Scenario 1: built image

```bash
mkdir -p /tmp/imgvol-repro && cd /tmp/imgvol-repro

cat > Dockerfile <<'EOF'
FROM alpine
RUN echo hello > /content.txt
EOF

cat > compose.yaml <<'EOF'
services:
  source:
    build: .
    image: imgvol-source

  consumer:
    image: alpine
    depends_on:
      - source
    command: ["cat", "/data/content.txt"]
    volumes:
      - type: image
        source: imgvol-source
        target: /data
EOF

docker compose -p imgvol-repro up --build
```

  Result:

```
 ✘ service:consumer:1                Error response from daemon: No such image: sha256:8d32cebc29f3dd77e043c002d09576aef8a85f58f0265541b1c082d65b3ec799                                  0.0s
Error response from daemon: No such image: sha256:8d32cebc29f3dd77e043c002d09576aef8a85f58f0265541b1c082d65b3ec799
```

Expected: consumer is created and prints hello.

Scenario 2: pulled image, second run

```bash
cat > compose-pull.yaml <<'EOF'
services:
  consumer:
    image: alpine
    command: ["ls", "-la", "/data"]
    volumes:
      - type: image
        source: nginx:alpine
        target: /data
        image:
          subpath: usr/share/nginx/html/
EOF
````

First run: nginx:alpine is not yet present locally
`docker compose -f compose-pull.yaml -p imgvol-repro up`

Result: succeeds, consumer lists the mounted files.

Second run: nginx:alpine is now present locally from the first run
  
`docker compose -f compose-pull.yaml -p imgvol-repro up`

Result:

```
  ✘ service:consumer:1 Error response from daemon: No such image: sha256:1d40e3eb3bf4f138de1d67193f2aa5309fcaf343eb5ffadbf5e9439de1eb1ebb
```

Expected: identical successful result to the first run.

### Compose Version

```Text
Docker Compose version 5.4.0
```

### Docker Environment

```Text
Client:
 Version:    29.4.0-ce
 Context:    default
 Debug Mode: false
 Plugins:
    Version:  0.33.0
    Path:     /usr/lib/docker/cli-plugins/docker-buildx
    Version:  5.4.0

Server:
 Containers: 1
  Running: 0
  Paused: 0
  Stopped: 1
 Images: 2
 Server Version: 29.4.0-ce
 Storage Driver: overlayfs
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
 CDI spec directories:
  /etc/cdi
  /var/run/cdi
 Swarm: inactive
 Runtimes: oci runc io.containerd.runc.v2
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: e8b1a9bc270f9952197c470b8bad573b03a3a608
 runc version: v1.4.3-0-gbb14dabe
 init version: 
 Security Options:
  seccomp
   Profile: builtin
  selinux
  cgroupns
 Kernel Version: 7.1.5-1-default
 Operating System: openSUSE Tumbleweed
 OSType: linux
 Architecture: x86_64
 CPUs: 4
 Total Memory: 15.55GiB
 Name: susetest
 ID: 9fd56017-09a8-4357-a23a-159e628b2959
 Debug Mode: false
 Experimental: false
 Insecure Registries:
  ::1/128
  127.0.0.0/8
 Live Restore Enabled: false
 Firewall Backend: iptables+firewalld
  ReloadedAt: 2026-08-05T12:22:02-04:00
```

### Anything else?

For some reason this was not caught by your CI.  This was caught in openQA when testing the latest opensuse Tumbleweed snapshot 20260804:

http://openqa-assets.opensuse.org/tests/6146470/file/docker_compose-e2e-compose.txt

```
# Test messages # TestImageVolume
# failure: 

Failed
=== RUN   TestImageVolume
    framework.go:170: WARNING: docker-model cli-plugin not found
    volumes_test.go:176: Running command: docker compose version
Docker Compose version 5.4.0
    volumes_test.go:182: 	[TestImageVolume] docker version -f {{.Server.Version}}
    volumes_test.go:189: Running command: docker compose -f ./fixtures/volumes/compose.yaml --project-name compose-e2e-image-volume up with_image
    volumes_test.go:189: assertion failed: 
        Command:  docker compose -f ./fixtures/volumes/compose.yaml --project-name compose-e2e-image-volume up with_image
        ExitCode: 1
        Error:    exit status 1
        Stdout:   
        Stderr:    Network compose-e2e-image-volume_default Creating 
         Network compose-e2e-image-volume_default Creating 
         Network compose-e2e-image-volume_default Created 
         Network compose-e2e-image-volume_default Created 
         Container compose-e2e-image-volume-with_image-1 Creating 
         service:with_image:1 Error response from daemon: No such image: sha256:1d40e3eb3bf4f138de1d67193f2aa5309fcaf343eb5ffadbf5e9439de1eb1ebb 
        Error response from daemon: No such image: sha256:1d40e3eb3bf4f138de1d67193f2aa5309fcaf343eb5ffadbf5e9439de1eb1ebb
        
        
        Failures:
        ExitCode was 1 expected 0
        Expected no error
    volumes_test.go:179: Running command: docker compose -p compose-e2e-image-volume down -v --remove-orphans
    framework.go:143: Contents of config dir:
    framework.go:145:   - /tmp/TestImageVolume790676610/001
    framework.go:145:   - /tmp/TestImageVolume790676610/001/cli-plugins
    framework.go:145:   - /tmp/TestImageVolume790676610/001/cli-plugins/docker-buildx
    framework.go:145:   - /tmp/TestImageVolume790676610/001/cli-plugins/docker-compose
    framework.go:145:   - /tmp/TestImageVolume790676610/001/cli-plugins/docker-scan
--- FAIL: TestImageVolume (0.59s)
```

```
# Test messages # TestImageVolumeRecreateOnRebuild
# failure: 

Failed
=== RUN   TestImageVolumeRecreateOnRebuild
    framework.go:170: WARNING: docker-model cli-plugin not found
    volumes_test.go:195: Running command: docker compose version
Docker Compose version 5.4.0
    volumes_test.go:202: 	[TestImageVolumeRecreateOnRebuild] docker version -f {{.Server.Version}}
    volumes_test.go:210: Running command: docker compose -f ./fixtures/image-volume-recreate/compose.yaml --project-name compose-e2e-image-volume-recreate build --build-arg CONTENT=foo
#1 [internal] load local bake definitions
#1 reading from stdin 643B done
#1 DONE 0.0s

#2 [internal] load build definition from Dockerfile
#2 transferring dockerfile: 835B done
#2 DONE 0.0s

#3 resolve image config for docker-image://docker.io/docker/dockerfile:1
#3 DONE 0.4s

#4 docker-image://docker.io/docker/dockerfile:1@sha256:87999aa3d42bdc6bea60565083ee17e86d1f3339802f543c0d03998580f9cb89
#4 resolve docker.io/docker/dockerfile:1@sha256:87999aa3d42bdc6bea60565083ee17e86d1f3339802f543c0d03998580f9cb89 0.0s done
#4 sha256:50ba52cd6a2c01eaf1a9efbedc7c75b5da5e3965c1586001c722980487a73fd7 14.36MB / 14.36MB 0.1s done
#4 extracting sha256:50ba52cd6a2c01eaf1a9efbedc7c75b5da5e3965c1586001c722980487a73fd7 0.1s done
#4 DONE 0.3s

#5 [internal] load metadata for docker.io/library/alpine:latest
#5 DONE 0.3s

#6 [internal] load .dockerignore
#6 transferring context: 2B done
#6 DONE 0.0s

#7 [1/3] FROM docker.io/library/alpine:latest@sha256:28bd5fe8b56d1bd048e5babf5b10710ebe0bae67db86916198a6eec434943f8b
#7 resolve docker.io/library/alpine:latest@sha256:28bd5fe8b56d1bd048e5babf5b10710ebe0bae67db86916198a6eec434943f8b 0.0s done
#7 CACHED

#8 [2/3] WORKDIR /app
#8 DONE 0.0s

#9 [3/3] RUN echo "foo" > /app/content.txt
#9 DONE 0.1s

#10 exporting to image
#10 exporting layers 0.1s done
#10 exporting manifest sha256:a86fb08624c3c1f065a96b693c6e3c8ec7595c906437cef55e82613012142bbe done
#10 exporting config sha256:771cc3ddf4f6bebce42c15337aab835cedca2b351f33e71bf3692e5fc2210ec4 done
#10 exporting attestation manifest sha256:312941b01ca2a2128b69eaf8559b8735b92201440f79ff8d9e4cf741444d42f1 0.0s done
#10 exporting manifest list sha256:0e97c48557001c59e388d5e41108df0d1901b27042a893a5abc30030071cab6a done
#10 naming to docker.io/library/image-volume-source:latest
#10 naming to docker.io/library/image-volume-source:latest done
#10 unpacking to docker.io/library/image-volume-source:latest 0.0s done
#10 DONE 0.1s

#11 resolving provenance for metadata file
#11 DONE 0.0s
    volumes_test.go:212: Running command: docker compose -f ./fixtures/image-volume-recreate/compose.yaml --project-name compose-e2e-image-volume-recreate up -d
    volumes_test.go:212: assertion failed: 
        Command:  docker compose -f ./fixtures/image-volume-recreate/compose.yaml --project-name compose-e2e-image-volume-recreate up -d
        ExitCode: 1
        Error:    exit status 1
        Stdout:   
        Stderr:    Network compose-e2e-image-volume-recreate_default Creating 
         Network compose-e2e-image-volume-recreate_default Creating 
         Network compose-e2e-image-volume-recreate_default Created 
         Network compose-e2e-image-volume-recreate_default Created 
         Container compose-e2e-image-volume-recreate-source-1 Creating 
         Container compose-e2e-image-volume-recreate-source-1 Created 
         Container compose-e2e-image-volume-recreate-consumer-1 Creating 
         service:consumer:1 Error response from daemon: No such image: sha256:a86fb08624c3c1f065a96b693c6e3c8ec7595c906437cef55e82613012142bbe 
        Error response from daemon: No such image: sha256:a86fb08624c3c1f065a96b693c6e3c8ec7595c906437cef55e82613012142bbe
        
        
        Failures:
        ExitCode was 1 expected 0
        Expected no error
    volumes_test.go:198: Running command: docker compose -p compose-e2e-image-volume-recreate down -v --remove-orphans
    volumes_test.go:199: 	[TestImageVolumeRecreateOnRebuild] docker rmi -f image-volume-source
    framework.go:143: Contents of config dir:
    framework.go:145:   - /tmp/TestImageVolumeRecreateOnRebuild1125399722/001
    framework.go:145:   - /tmp/TestImageVolumeRecreateOnRebuild1125399722/001/buildx
    framework.go:145:   - /tmp/TestImageVolumeRecreateOnRebuild1125399722/001/buildx/.buildNodeID
    framework.go:145:   - /tmp/TestImageVolumeRecreateOnRebuild1125399722/001/buildx/.lock
    framework.go:145:   - /tmp/TestImageVolumeRecreateOnRebuild1125399722/001/buildx/activity
    framework.go:145:   - /tmp/TestImageVolumeRecreateOnRebuild1125399722/001/buildx/activity/default
    framework.go:145:   - /tmp/TestImageVolumeRecreateOnRebuild1125399722/001/buildx/current
    framework.go:145:   - /tmp/TestImageVolumeRecreateOnRebuild1125399722/001/buildx/defaults
    framework.go:145:   - /tmp/TestImageVolumeRecreateOnRebuild1125399722/001/buildx/instances
    framework.go:145:   - /tmp/TestImageVolumeRecreateOnRebuild1125399722/001/buildx/refs
    framework.go:145:   - /tmp/TestImageVolumeRecreateOnRebuild1125399722/001/buildx/refs/__group__
    framework.go:145:   - /tmp/TestImageVolumeRecreateOnRebuild1125399722/001/buildx/refs/__group__/uxp25pzplg7vmlg36bg81tt88
    framework.go:145:   - /tmp/TestImageVolumeRecreateOnRebuild1125399722/001/buildx/refs/default
    framework.go:145:   - /tmp/TestImageVolumeRecreateOnRebuild1125399722/001/buildx/refs/default/default
    framework.go:145:   - /tmp/TestImageVolumeRecreateOnRebuild1125399722/001/buildx/refs/default/default/fo2yu6atj4qaovhgvox33mm9o
    framework.go:145:   - /tmp/TestImageVolumeRecreateOnRebuild1125399722/001/buildx/refs/version
    framework.go:145:   - /tmp/TestImageVolumeRecreateOnRebuild1125399722/001/cli-plugins
    framework.go:145:   - /tmp/TestImageVolumeRecreateOnRebuild1125399722/001/cli-plugins/docker-buildx
    framework.go:145:   - /tmp/TestImageVolumeRecreateOnRebuild1125399722/001/cli-plugins/docker-compose
    framework.go:145:   - /tmp/TestImageVolumeRecreateOnRebuild1125399722/001/cli-plugins/docker-scan
--- FAIL: TestImageVolumeRecreateOnRebuild (2.39s)
```

### LINKED ISSUE #14007: [BUG] docker compose ignores local non-native variant from a multi-platform image (pulls instead)
### Description

*Disclaimer:* This bug report was created with AI (due to me having very little time for these kind of things). The problem, however, is very real. I have been trying to figure it out for the last couple of day and this bug report is the result of the analysis.

When Docker's containerd image store is enabled and a locally tagged multi-platform image contains both the host platform and a non-native platform, Docker Compose incorrectly treats the image as unavailable for the non-native platform and attempts to pull it from a registry.

In this reproduction, the Docker host is `linux/arm64`, while the Compose service requests `linux/amd64`. The unpublished local image tag contains runnable variants for both platforms.

Current behavior:

- `docker image inspect --platform linux/amd64` finds the requested local variant.
- `docker run --platform linux/amd64` runs the local variant successfully.
- `docker compose up` reports that the local image does not match `linux/amd64`, treats it as missing, and attempts to pull the unpublished tag (which fails).
- `docker compose up --pull never` runs the same local `linux/amd64` variant successfully.

Expected behavior:

With the default `missing` pull policy, Compose should recognize that the requested `linux/amd64` variant is available in the local multi-platform image and start it without contacting a registry.

### Steps To Reproduce

*Disclaimer:* These instructions were generated by AI. I have manually verified them.

These steps use an ARM64 Docker host (macOS in my case) and Docker's containerd image store. Docker Desktop provides the emulation needed to run the non-native AMD64 variant. On an AMD64 host, the requested and host platforms can be reversed if ARM64 emulation is available.

1. Confirm that the containerd image store is enabled:

   ```console
   docker info --format '{{json .DriverStatus}}'
   ```

   The output should contain:

   ```text
   io.containerd.snapshotter.v1
   ```

1. Pull the ARM64 and AMD64 variants of Alpine into the local multi-platform image store, then give the image a tag that does not exist in the registry:

   ```console
   docker pull --platform linux/arm64 alpine:3.22
   docker pull --platform linux/amd64 alpine:3.22
   docker tag alpine:3.22 docker.io/library/compose-multiplatform-repro-codex:local-only
   ```

1. Confirm that both variants are locally available:

   ```console
   docker image ls --tree docker.io/library/compose-multiplatform-repro-codex:local-only
   ```

   The output includes locally stored content for both requested platforms:

   ```text
   IMAGE                                          ID             DISK USAGE   CONTENT SIZE   EXTRA
   compose-multiplatform-repro-codex:local-only   14358309a308       26.2MB         8.08MB
   ├─ linux/amd64                                 7c8cb692ae09       12.7MB         3.79MB
   ├─ linux/arm64/v8                              2c9d26f410d0       13.3MB         4.12MB
   ```

1. Create `compose.yaml`:

   ```yaml
   services:
     repro:
       image: docker.io/library/compose-multiplatform-repro-codex:local-only
       platform: linux/amd64
       command: ["uname", "-m"]
   ```

1. Run the requested local variant directly with Docker:

   ```console
   docker run --rm --platform linux/amd64 docker.io/library/compose-multiplatform-repro-codex:local-only uname -m
   ```

   This succeeds:

   ```text
   x86_64
   ```

1. Run the equivalent service with Compose and enable debug logging:

   ```console
   docker --debug compose up --abort-on-container-exit --exit-code-from repro
   ```

   Compose incorrectly rejects the local image and attempts to pull it:

   ```text
   level=debug msg="local image docker.io/library/compose-multiplatform-repro-codex:local-only doesn't match expected platform linux/amd64"
   Image docker.io/library/compose-multiplatform-repro-codex:local-only Pulling
   Image docker.io/library/compose-multiplatform-repro-codex:local-only Error pull access denied for compose-multiplatform-repro-codex, repository does not exist or may require 'docker login'
   Error response from daemon: pull access denied for compose-multiplatform-repro-codex, repository does not exist or may require 'docker login'
   ```

1. Disable pulling and run the same Compose service:

   ```console
   docker compose up --pull never --abort-on-container-exit --exit-code-from repro
   ```

   This succeeds using the requested local variant:

   ```text
   repro-1  | x86_64
   repro-1 exited with code 0
   ```

1. Clean up the reproduction resources:

   ```console
   docker compose down --remove-orphans
   docker image rm docker.io/library/compose-multiplatform-repro-codex:local-only
   ```

### Compose Version

```Text
$ docker compose version
Docker Compose version v5.3.1

$ docker-compose version
Docker Compose version v5.3.1
```

### Docker Environment

```Text
Client:
 Version:    29.6.2
 Context:    desktop-linux
 Debug Mode: false
 Plugins:
  agent: Docker AI Agent Runner (Docker Inc.)
    Version:  v1.115.0
    Path:     /Users/manski/.docker/cli-plugins/docker-agent
    Version:  v1.27.0
    Path:     /Users/manski/.docker/cli-plugins/docker-ai
    Version:  v0.35.0-desktop.2
    Path:     /Users/manski/.docker/cli-plugins/docker-buildx
    Version:  v5.3.1
    Path:     /Users/manski/.docker/cli-plugins/docker-compose
    Version:  0.0.47
    Path:     /Users/manski/.docker/cli-plugins/docker-debug
    Version:  v0.4.3
    Path:     /Users/manski/.docker/cli-plugins/docker-desktop
  dhi: CLI for managing Docker Hardened Images (Docker Inc.)
    Version:  v0.0.7
    Path:     /Users/manski/.docker/cli-plugins/docker-dhi
    Version:  v0.2.31
    Path:     /Users/manski/.docker/cli-plugins/docker-extension
    Version:  v1.4.0
    Path:     /Users/manski/.docker/cli-plugins/docker-init
    Version:  v0.43.3
    Path:     /Users/manski/.docker/cli-plugins/docker-mcp
    Version:  v1.2.6
    Path:     /Users/manski/.docker/cli-plugins/docker-model
    Version:  v0.6.9
    Path:     /Users/manski/.docker/cli-plugins/docker-offload
    Version:  v0.2.0
    Path:     /Users/manski/.docker/cli-plugins/docker-pass
  sandbox: "docker sandbox" is deprecated, use Docker Sandboxes instead (Docker Inc.)
    Version:  v0.13.0
    Path:     /Users/manski/.docker/cli-plugins/docker-sandbox
    Version:  v1.24.0
    Path:     /Users/manski/.docker/cli-plugins/docker-scout

Server:
 Containers: 1
  Running: 0
  Paused: 0
  Stopped: 1
 Images: 2
 Server Version: 29.6.2
 Storage Driver: overlayfs
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 2
 Plugins:
  Volume: local
 CDI spec directories:
  /etc/cdi
  /var/run/cdi
 Discovered Devices:
  cdi: docker.com/gpu=webgpu
 Swarm: inactive
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: e53c7c1516c3b2bff98eb76f1f4117477e6f4e66
 init version: de40ad0
 Security Options:
  seccomp
   Profile: builtin
  cgroupns
 Kernel Version: 6.12.76-linuxkit
 OSType: linux
 Architecture: aarch64
 CPUs: 8
 Total Memory: 3.826GiB
 Name: docker-desktop
 ID: 3eff3bdc-fd0e-4464-bc5d-1cfad3f2db4e
 Debug Mode: false
 Labels:
  com.docker.desktop.address=unix:///Users/manski/Library/Containers/com.docker.docker/Data/docker-cli.sock
 Experimental: false
 Insecure Registries:
  hubproxy.docker.internal:5555
  127.0.0.0/8
  ::1/128
 Live Restore Enabled: false
 Firewall Backend: iptables
```

### Anything else?

*Disclaimer:* This analysis has been done by the AI. I haven't verified it (lack of knowledge), so take it with a grain of salt.

The issue only occurs with Compose's default pull behavior. Both of the following avoid the incorrect registry request and successfully run the requested local variant:

```yaml
pull_policy: never
```

```console
docker compose up --pull never
```

Compose's debug output shows the incorrect local-image decision directly:

```text
local image docker.io/library/compose-multiplatform-repro-codex:local-only doesn't match expected platform linux/amd64
```

The likely cause is the platform validation in [`getLocalImagesDigests`](https://github.com/docker/compose/blob/v5.3.1/pkg/compose/build.go#L192-L240). It inspects the local image without requesting `service.Platform`, compares the returned platform with the requested platform, and deletes the image from the local-image map on a mismatch.

With a multi-platform containerd image store, a platformless inspection can return the host variant even though the requested non-native variant is also locally available. Docker supports selecting the variant during inspection with [`docker image inspect --platform`](https://docs.docker.com/reference/cli/docker/image/inspect/).

The same local-platform check remains in [Compose v5.4.0](https://github.com/docker/compose/blob/v5.4.0/pkg/compose/build.go#L192-L239).

Related reports provide historical context but do not appear to cover this exact containerd multi-platform case:

- [#9740](https://github.com/docker/compose/issues/9740) led to the local-platform validation, but concerned the classic single-platform image store.
- [#11842](https://github.com/docker/compose/issues/11842) concerned a genuinely wrong-platform local image and diagnostic output.
- [#12854](https://github.com/docker/compose/issues/12854) concerned multiple platforms under one tag with the classic store; enabling the containerd image store was the recommended solution.

The bug was originally observed in this [GitHub Actions job](https://github.com/skrysm/homelab-containers/actions/runs/30977143074/job/92215341562?pr=40), then reproduced independently with the minimal Alpine example above.

### LINKED ISSUE #14016: [BUG] `docker compose up --pull missing` sometimes recreates container without pulling
### Description

In the CI for the Ansible collection community.docker, I noticed anohter problem with the latest Compose version. This one happens in a Ubuntu 22.04 VM and also in a RHEL 10.2 VM (probably also RHEL 9.8, didn't test that), but does not happen for example on my Arch Linux box.

`docker-compose.yml`:
```
services:
  foo:
          image: ghcr.io/ansible-collections/simple-1:tag
          command: 10m
          stop_grace_period: 1s
```

### Steps To Reproduce

1. Make sure image is not present: `docker rmi ghcr.io/ansible-collections/simple-1:tag`
2. Bring up compose with pulling: `docker compose --ansi never --progress json up --detach --quiet-pull --pull missing --`
3. Bring up compose again, should not pull:` docker compose --ansi never --progress json up --detach --quiet-pull --pull missing --`
3. For good measure, another time: `docker compose --ansi never --progress json up --detach --quiet-pull --pull missing --`

Note that 2. as expected pulls the image, creates the network and container, and starts the container:
```
{"id":"Image ghcr.io/ansible-collections/simple-1:tag","status":"Working","text":"Pulling"}
{"id":"Image ghcr.io/ansible-collections/simple-1:tag","status":"Done","text":"Pulled"}
{"id":"Network t_default","status":"Working","text":"Creating"}
{"id":"Network t_default","status":"Working","text":"Creating"}
{"id":"Network t_default","status":"Done","text":"Created"}
{"id":"Network t_default","status":"Done","text":"Created"}
{"id":"Container t-foo-1","status":"Working","text":"Creating"}
{"id":"Container t-foo-1","status":"Done","text":"Created"}
{"id":"Container t-foo-1","status":"Working","text":"Starting"}
{"id":"Container t-foo-1","status":"Done","text":"Started"}
```
Then 3. re-creates the container, even though the image is already there!
```
{"id":"Container t-foo-1","status":"Working","text":"Recreate"}
{"id":"Container t-foo-1","status":"Done","text":"Recreated"}
{"id":"Container t-foo-1","status":"Working","text":"Starting"}
{"id":"Container t-foo-1","status":"Done","text":"Started"}
```
Interestingly, when running the command again (4.), everything's fine:
```
{"id":"Container t-foo-1","status":"Done","text":"Running"}
```
Every further run results in the same.

### Compose Version

```Text
Docker Compose version v5.4.0
```

### Docker Environment

```Text
 Version:    29.7.2
 Context:    default
 Debug Mode: false
 Plugins:
    Version:  v0.36.1
    Version:  v5.4.0

Server:
 Containers: 1
  Running: 1
  Paused: 0
  Stopped: 0
 Images: 1
 Server Version: 29.7.2
 Storage Driver: overlayfs
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
 CDI spec directories:
  /etc/cdi
  /var/run/cdi
 Swarm: inactive
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: aad11006b869517fcd3009450b6f82da282e1a9b
 runc version: v1.4.3-0-gbb14dabe
 init version: de40ad0
 Security Options:
  apparmor
  seccomp
   Profile: builtin
  cgroupns
 Kernel Version: 6.8.0-1061-aws
 Operating System: Ubuntu 22.04.5 LTS
 OSType: linux
 Architecture: x86_64
 CPUs: 2
 Total Memory: 3.727GiB
 Name: ip-192-168-5-65
 ID: 14c2997c-3861-4156-97f9-84463d814926
 Debug Mode: false
 Experimental: false
 Insecure Registries:
  ::1/128
  127.0.0.0/8
 Live Restore Enabled: false
 Firewall Backend: iptables
  EnableUserlandProxy: true
```

### Anything else?

This also can be reproduced in the same environment as in #14014. It does not happen on my Arch Linux box with the same Docker Compose and Docker deamon versions, or in the GHA based tests (which use Docker daemon from GHA's `ubuntu:latest`).

### LINKED ISSUE #14036: [BUG] Second up recreates image pulled by first up with containerd store
*This issue was filed by an AI agent on a human's behalf. The human submitter may not have independently verified the report.*

### Description

On an Engine using the containerd image store, Compose 5.4.0 can recreate a container on the second identical `docker compose up` when the first `up` had to pull a multi-manifest image.

Current behavior:

- The first `up` pulls the missing image and creates the container.
- The second identical `up`, without another pull or any configuration/image change, reports the service as `Recreate` and replaces the container.

Expected behavior:

- The second `up` recognizes that the image and configuration are unchanged and leaves the container running.

This can destroy state held in the container writable layer. We encountered it with a stateful infrastructure container initialized between the two Compose invocations.

### Steps To Reproduce

Use Docker Engine with the containerd image store and API >= 1.48. Start with the selected multi-platform image absent locally.

```yaml
services:
  app:
    image: alpine:3.22.2
```

```console
docker image rm alpine:3.22.2 # only to make the first Compose invocation pull

docker compose up -d
first_id=$(docker compose ps -q app)
docker inspect -f '{{index .Config.Labels "com.docker.compose.image"}}' "$first_id"

docker compose up -d
second_id=$(docker compose ps -q app)

test "$first_id" = "$second_id"
```

On the affected setup, the second `up` prints `Recreate` and the final comparison fails.

### Compose Version

```Text
Docker Compose version v5.4.0
```

### Docker Environment

```Text
Docker Engine 29.7.2
API version 1.55
Storage Driver: overlayfs
containerd 2.3.3
linux/amd64
```

### Anything else?

This appears to be a digest-type inconsistency introduced when `getImageSummaries` was changed by fd794ea8421812cd0604d5ebae632eb4a098513c to select the platform runnable-manifest digest:

- During the initial pull, `pullRequiredImages` stores the value returned by `pullServiceImage` in the project image map.
- `pullServiceImage` currently returns plain `ImageInspect.ID`. With the containerd image store, this can be the top-level image-index digest. That value becomes the new container's `com.docker.compose.image` label.
- During the next `up`, the image is already local. `getImageSummaries` requests manifests and `contentDigest` returns the platform runnable-manifest digest.
- Reconciliation compares the runnable-manifest digest with the existing container's index-digest label and treats the unchanged image as stale.

In the observed failure, the existing container label held `sha256:46302bcb91a7e8008e6394be8afafdbfa40fb77a54d4046a38be35992042d5de`, while the later local-image lookup selected a different platform manifest digest for the same unchanged tag. The Compose configuration hash and mounts were identical, and no pull occurred during the second invocation.

Compose 5.1.4 does not reproduce this because both paths use plain `inspect.ID`. A patch that resolves the post-pull image through the same `getImageSummaries` path fixes the mismatch.

This is related to the containerd multi-manifest handling in #14007, but it is a different path and symptom: #14007 rejects a locally available non-native platform and attempts a pull; this issue successfully pulls a same-platform image and then unnecessarily recreates it on the following `up`.

glours:
**What I did**
Compose records image identities for staleness detection (the `com.docker.compose.image` label compared by `mustRecreate`) through several independent code paths — pull, bake, classic builder, already-local inspect — and they didn't all produce the same *kind* of digest for the same image: top-level index digest, per-platform manifest digest, or config digest, depending on the engine (graphdriver vs containerd image store), the API version, the builder, and how the image arrived locally. Whenever two runs resolved the same image through different paths, the comparison failed and containers were recreated with no actual change. This mismatch class is behind #13636, #14005, and the phantom-recreate-after-pull fixed by #13998 — plus several latent cases (platform-pinned services resolved with the host platform, wrong-platform digests leaking into labels, bake push-only builds recreating on every `up`).

This PR fixes the class, not just the instances:

- **One canonical content-digest producer** (`matchLocalManifest` / `localContentDigest`): every path — pull, bake, classic builder, local inspect — resolves the recorded identity the same way. Digest selection and the platform-mismatch check now share the same manifest resolution, so the digest recorded and the platform validated always refer to the same manifest.
- **Single label writer**: `ensureImagesExists` is the only writer of `com.docker.compose.image`. Platform-pinned services get the digest of *their* platform's manifest, resolved in-process from the already-fetched inspects (zero extra API calls in steady state).
- **`type: image` volumes** mount by resolvable name and track their source-image digest in a dedicated `com.docker.compose.image-volume-digest` label (fixes the #14005 regression).
- **CI now runs e2e against the containerd image store** — the configuration where digest kinds actually diverge, and the blind spot that let this whole series ship. A new `TestUpIdempotentContainerdStore` locks the invariant: two consecutive `up` runs with no change must not recreate anything.
- **Command alignment**: `scale` and `run` now resolve `DOCKER_DEFAULT_PLATFORM` like `up` does (they recreated containers on every invocation otherwise), and `compose pull` interprets `pull_policy` through the same interpreter as `up` (two deliberate, documented differences remain: an unset policy always refreshes, and a present `latest` tag is still refreshed under `missing`).

This PR supersedes and includes #13998 (@benjick) and #14006 (@ricardobranco777) — both commits are integrated with their original authorship preserved. Thanks to both for the investigations that narrowed this down.

**User-visible changes**

- `compose pull` now honors `daily`/`weekly`/`every_N` refresh windows instead of always re-pulling.
- One-time recreate on first `up` after upgrading: label values change kind (and image-volume users transition from digest-as-mount-source to the new label). Subsequent runs are stable.

Each commit is independently CI-green and reviewable on its own: image-volume mount fix → containerd-store CI job (test skipped) → pulled-image content digest (test enabled) → canonical producer/single writer → scale/run platform resolution → pull policy alignment.

**Related issue**
Closes #14005
Closes #14007 (verified: reproduced on main, fixed by this PR — e2e regression test in #14025)
Closes #14016 (verified: phantom recreate after pull, locked by `TestUpIdempotentContainerdStore`)
Closes #14036 (same phantom-recreate-after-pull scenario, reported independently — supersedes #14037)
Supersedes #13998, #14006

image

### DIFF
--- a/cmd/compose/scale.go
+++ b/cmd/compose/scale.go
@@ -80,6 +80,13 @@ func runScale(ctx context.Context, dockerCli command.Cli, backendOptions *Backen
 		return err
 	}
 
+	// resolve DOCKER_DEFAULT_PLATFORM into service.Platform exactly like
+	// `up`/`create` do: Platform feeds the service config-hash, so scale
+	// hashing a different value would recreate every container
+	if err := applyPlatforms(project, true); err != nil {
+		return err
+	}
+
 	if opts.noDeps {
 		if project, err = project.WithSelectedServices(services, types.IgnoreDependencies); err != nil {
 			return err
--- a/pkg/compose/build.go
+++ b/pkg/compose/build.go
@@ -19,6 +19,7 @@ package compose
 import (
 	"context"
 	"fmt"
+	"sort"
 	"strings"
 	"time"
 
@@ -167,25 +168,36 @@ func (s *composeService) ensureImagesExists(ctx context.Context, project *types.
 }
 
 func resolveImageVolumes(service *types.ServiceConfig, images map[string]api.ImageSummary, projectName string) {
+	var digests []string
 	for i, vol := range service.Volumes {
-		if vol.Type == types.VolumeTypeImage {
-			imgName := vol.Source
-			if _, ok := images[vol.Source]; !ok {
-				// check if source is another service in the project
-				imgName = api.GetImageNameOrDefault(types.ServiceConfig{Name: vol.Source}, projectName)
-				// If we still can't find it, it might be an external image that wasn't pulled yet or doesn't exist
-				if _, ok := images[imgName]; !ok {
-					continue
-				}
-			}
-			if img, ok := images[imgName]; ok {
-				// Use Image ID directly as source.
-				// Using name@digest format (via reference.WithDigest) fails for local-only images
-				// that don't have RepoDigests (e.g. built locally in CI).
-				// Image ID (sha256:...) is always valid and ensures ServiceHash changes on rebuild.
-				service.Volumes[i].Source = img.ID
+		if vol.Type != types.VolumeTypeImage {
+			continue
+		}
+		imgName := vol.Source
+		if _, ok := images[vol.Source]; !ok {
+			// check if source is another service in the project
+			imgName = api.GetImageNameOrDefault(types.ServiceConfig{Name: vol.Source}, projectName)
+			// If we still can't find it, it might be an external image that wasn't pulled yet or doesn't exist
+			if _, ok := images[imgName]; !ok {
+				continue
 			}
 		}
+		img, ok := images[imgName]
+		if !ok {
+			continue
+		}
+		// The daemon only resolves a `type=image` mount Source that is a name/tag
+		// or a top-level image ID, not a per-platform manifest digest (which is
+		// what ImageSummary.ID holds to stay stable across attested rebuilds, see
+		// contentDigest()). Keep Source as the resolved name so mounting always
+		// works, and track the digest separately so mustRecreate can still detect
+		// a changed source image.
+		service.Volumes[i].Source = imgName
+		digests = append(digests, vol.Target+"="+img.ID)
+	}
+	if len(digests) > 0 {
+		sort.Strings(digests)
+		service.CustomLabels.Add(api.ImageVolumeDigestLabel, strings.Join(digests, ","))
 	}
 }
 
--- a/pkg/compose/build.go
+++ b/pkg/compose/build.go
@@ -24,8 +24,6 @@ import (
 	"time"
 
 	"github.com/compose-spec/compose-go/v2/types"
-	"github.com/containerd/platforms"
-	specs "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/sirupsen/logrus"
 
 	"github.com/docker/compose/v5/internal/tracing"
@@ -115,7 +113,7 @@ func (s *composeService) ensureImagesExists(ctx context.Context, project *types.
 		}
 	}
 
-	images, err := s.getLocalImagesDigests(ctx, project)
+	images, pinnedDigests, err := s.getLocalImagesDigests(ctx, project)
 	if err != nil {
 		return err
 	}
@@ -152,12 +150,17 @@ func (s *composeService) ensureImagesExists(ctx context.Context, project *types.
 		}
 	}
 
-	// set digest as com.docker.compose.image label so we can detect outdated containers
+	// set digest as com.docker.compose.image label so we can detect outdated
+	// containers — the single writer of that label, so the platform-pinned
+	// resolution below can't be overwritten by another code path
 	for name, service := range project.Services {
 		image := api.GetImageNameOrDefault(service, project.Name)
 		img, ok := images[image]
 		if ok {
-			service.CustomLabels.Add(api.ImageDigestLabel, img.ID)
+			// a platform-pinned service is labelled with the digest of ITS
+			// platform's manifest — see serviceImageDigest
+			digest := s.serviceImageDigest(ctx, service, image, img, pinnedDigests)
+			service.CustomLabels = service.CustomLabels.Add(api.ImageDigestLabel, digest)
 		}
 
 		resolveImageVolumes(&service, images, project.Name)
@@ -174,22 +177,19 @@ func resolveImageVolumes(service *types.ServiceConfig, images map[string]api.Ima
 			continue
 		}
 		imgName := vol.Source
-		if _, ok := images[vol.Source]; !ok {
+		img, ok := images[imgName]
+		if !ok {
 			// check if source is another service in the project
 			imgName = api.GetImageNameOrDefault(types.ServiceConfig{Name: vol.Source}, projectName)
 			// If we still can't find it, it might be an external image that wasn't pulled yet or doesn't exist
-			if _, ok := images[imgName]; !ok {
+			if img, ok = images[imgName]; !ok {
 				continue
 			}
-		}
-		img, ok := images[imgName]
-		if !ok {
-			continue
 		}
 		// The daemon only resolves a `type=image` mount Source that is a name/tag
 		// or a top-level image ID, not a per-platform manifest digest (which is
 		// what ImageSummary.ID holds to stay stable across attested rebuilds, see
-		// contentDigest()). Keep Source as the resolved name so mounting always
+		// localContentDigest). Keep Source as the resolved name so mounting always
 		// works, and track the digest separately so mustRecreate can still detect
 		// a changed source image.
 		service.Volumes[i].Source = imgName
@@ -197,11 +197,22 @@ func resolveImageVolumes(service *types.ServiceConfig, images map[string]api.Ima
 	}
 	if len(digests) > 0 {
 		sort.Strings(digests)
-		service.CustomLabels.Add(api.ImageVolumeDigestLabel, strings.Join(digests, ","))
+		service.CustomLabels = service.CustomLabels.Add(api.ImageVolumeDigestLabel, strings.Join(digests, ","))
 	}
 }
 
-func (s *composeService) getLocalImagesDigests(ctx context.Context, project *types.Project) (map[string]api.ImageSummary, error) {
+// pinnedImageDigest is the content digest of a platform-pinned service's
+// image, resolved for the service's platform rather than the host's.
+type pinnedImageDigest struct {
+	digest string
+	// from is the shared summary digest at resolution time: once a pull or
+	// build refreshed the summary, this resolution is stale and
+	// serviceImageDigest re-resolves the pinned platform, keeping the
+	// refreshed shared value only as fallback.
+	from string
+}
+
+func (s *composeService) getLocalImagesDigests(ctx context.Context, project *types.Project) (map[string]api.ImageSummary, map[string]pinnedImageDigest, error) {
 	imageNames := utils.Set[string]{}
 	for _, s := range project.Services {
 		imageNames.Add(api.GetImageNameOrDefault(s, project.Name))
@@ -214,48 +225,45 @@ func (s *composeService) getLocalImagesDigests(ctx context.Context, project *typ
 			imageNames.Add(img)
 		}
 	}
-	imgs, err := s.getImageSummaries(ctx, imageNames.Elements())
+	inspections, err := s.inspectLocalImages(ctx, imageNames.Elements())
 	if err != nil {
-		return nil, err
+		return nil, nil, err
+	}
+	imgs := make(map[string]api.ImageSummary, len(inspections))
+	for repoTag, inspect := range inspections {
+		imgs[repoTag] = imageSummary(repoTag, inspect)
 	}
 
-	for i, service := range project.Services {
+	pinnedDigests := map[string]pinnedImageDigest{}
+	for name, service := range project.Services {
+		if service.Platform == "" {
+			continue
+		}
 		imgName := api.GetImageNameOrDefault(service, project.Name)
 		img, ok := imgs[imgName]
 		if !ok {
 			continue
 		}
-		if service.Platform != "" {
-			platform, err := platforms.Parse(service.Platform)
-			if err != nil {
-				return nil, err
-			}
-			// inspect by name, not img.ID: img.ID now holds a content-manifest
-			// digest (see contentDigest) which is not necessarily inspectable,
-			// whereas the image name always resolves.
-			inspect, err := s.apiClient().ImageInspect(ctx, imgName)
-			if err != nil {
-				return nil, err
-			}
-			actual := specs.Platform{
-				Architecture: inspect.Architecture,
-				OS:           inspect.Os,
-				Variant:      inspect.Variant,
-			}
-			if !platforms.NewMatcher(platform).Match(actual) {
-				logrus.Debugf("local image %s doesn't match expected platform %s", service.Image, service.Platform)
-				// there is a local image, but it's for the wrong platform, so
-				// pretend it doesn't exist so that we can pull/build an image
-				// for the correct platform instead
-				delete(imgs, imgName)
-			}
+		// digest selection and platform validation share the same manifest
+		// resolution (matchLocalManifest) on the inspect we already hold,
+		// otherwise the digest recorded and the platform validated could
+		// refer to different manifests of the same image
+		digest, satisfied, err := localContentDigest(inspections[imgName], service.Platform)
+		if err != nil {
+			return nil, nil, err
 		}
-
-		project.Services[i].CustomLabels.Add(api.ImageDigestLabel, img.ID)
-
+		if !satisfied {
+			logrus.Debugf("local image %s doesn't match expected platform %s", service.Image, service.Platform)
+			// there is a local image, but it's for the wrong platform, so
+			// pretend it doesn't exist so that we can pull/build an image
+			// for the correct platform instead
+			delete(imgs, imgName)
+			continue
+		}
+		pinnedDigests[name] = pinnedImageDigest{digest: digest, from: img.ID}
 	}
 
-	return imgs, nil
+	return imgs, pinnedDigests, nil
 }
 
 // resolveAndMergeBuildArgs returns the final set of build arguments to use for the service image build.
--- a/pkg/compose/images.go
+++ b/pkg/compose/images.go
@@ -185,6 +185,27 @@ func (s *composeService) manifestsSupported(ctx context.Context) (bool, error) {
 	return versions.GreaterThanOrEqualTo(version, apiVersion148), nil
 }
 
+// inspectContentDigest inspects ref, requesting per-manifest data on engines
+// that support it, and returns the digest identifying the image's runnable
+// content for the default platform. Callers that record an image identity
+// compose later compares for staleness must go through this, so every such
+// identity is computed the same way — see contentDigest.
+func (s *composeService) inspectContentDigest(ctx context.Context, ref string) (string, error) {
+	withManifests, err := s.manifestsSupported(ctx)
+	if err != nil {
+		return "", err
+	}
+	var opts []client.ImageInspectOption
+	if withManifests {
+		opts = append(opts, client.ImageInspectWithManifests(true))
+	}
+	inspected, err := s.apiClient().ImageInspect(ctx, ref, opts...)
+	if err != nil {
+		return "", err
+	}
+	return contentDigest(inspected.InspectResponse, platforms.Default()), nil
+}
+
 // contentDigest returns the digest identifying an image's runnable content
 // (config + layers) for the given platform. With BuildKit provenance
 // attestations enabled (the default since recent Buildx/BuildKit), the image is
--- a/pkg/compose/images.go
+++ b/pkg/compose/images.go
@@ -24,6 +24,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/compose-spec/compose-go/v2/types"
 	"github.com/containerd/errdefs"
 	"github.com/containerd/platforms"
 	"github.com/distribution/reference"
@@ -31,6 +32,8 @@ import (
 	"github.com/moby/moby/api/types/image"
 	"github.com/moby/moby/client"
 	"github.com/moby/moby/client/pkg/versions"
+	specs "github.com/opencontainers/image-spec/specs-go/v1"
+	"github.com/sirupsen/logrus"
 	"golang.org/x/sync/errgroup"
 
 	"github.com/docker/compose/v5/pkg/api"
@@ -126,22 +129,19 @@ func (s *composeService) Images(ctx context.Context, projectName string, options
 	return summary, err
 }
 
-func (s *composeService) getImageSummaries(ctx context.Context, repoTags []string) (map[string]api.ImageSummary, error) {
-	summary := map[string]api.ImageSummary{}
-	l := sync.Mutex{}
-
-	withManifests, err := s.manifestsSupported(ctx)
+// inspectLocalImages inspects the given references in parallel, requesting
+// per-manifest data on engines that support it. References not found locally
+// are simply absent from the result.
+func (s *composeService) inspectLocalImages(ctx context.Context, repoTags []string) (map[string]client.ImageInspectResult, error) {
+	opts, err := s.imageInspectOptions(ctx)
 	if err != nil {
 		return nil, err
 	}
-
+	inspections := map[string]client.ImageInspectResult{}
+	l := sync.Mutex{}
 	eg, ctx := errgroup.WithContext(ctx)
 	for _, repoTag := range repoTags {
 		eg.Go(func() error {
-			var opts []client.ImageInspectOption
-			if withManifests {
-				opts = append(opts, client.ImageInspectWithManifests(true))
-			}
 			inspect, err := s.apiClient().ImageInspect(ctx, repoTag, opts...)
 			if err != nil {
 				if errdefs.IsNotFound(err) {
@@ -149,29 +149,47 @@ func (s *composeService) getImageSummaries(ctx context.Context, repoTags []strin
 				}
 				return fmt.Errorf("unable to get image '%s': %w", repoTag, err)
 			}
-			tag := ""
-			repository := ""
-			ref, err := reference.ParseDockerRef(repoTag)
-			if err == nil {
-				// ParseDockerRef will reject a local image ID
-				repository = reference.FamiliarName(ref)
-				if tagged, ok := ref.(reference.Tagged); ok {
-					tag = tagged.Tag()
-				}
-			}
 			l.Lock()
-			summary[repoTag] = api.ImageSummary{
-				ID:          contentDigest(inspect.InspectResponse, platforms.Default()),
-				Repository:  repository,
-				Tag:         tag,
-				Size:        inspect.Size,
-				LastTagTime: inspect.Metadata.LastTagTime,
-			}
+			inspections[repoTag] = inspect
 			l.Unlock()
 			return nil
 		})
 	}
-	return summary, eg.Wait()
+	return inspections, eg.Wait()
+}
+
+// imageInspectOptions requests per-manifest data when the engine supports it
+// (see manifestsSupported).
+func (s *composeService) imageInspectOptions(ctx context.Context) ([]client.ImageInspectOption, error) {
+	withManifests, err := s.manifestsSupported(ctx)
+	if err != nil {
+		return nil, err
+	}
+	if !withManifests {
+		return nil, nil
+	}
+	return []client.ImageInspectOption{client.ImageInspectWithManifests(true)}, nil
+}
+
+func imageSummary(repoTag string, inspect client.ImageInspectResult) api.ImageSummary {
+	tag := ""
+	repository := ""
+	ref, err := reference.ParseDockerRef(repoTag)
+	if err == nil {
+		// ParseDockerRef will reject a local image ID
+		repository = reference.FamiliarName(ref)
+		if tagged, ok := ref.(reference.Tagged); ok {
+			tag = tagged.Tag()
+		}
+	}
+	id, _, _ := localContentDigest(inspect, "")
+	return api.ImageSummary{
+		ID:          id,
+		Repository:  repository,
+		Tag:         tag,
+		Size:        inspect.Size,
+		LastTagTime: inspect.Metadata.LastTagTime,
+	}
 }
 
 // manifestsSupported reports whether the engine can return per-manifest data on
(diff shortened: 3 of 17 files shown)


benjick:
Thanks for picking this up and generalizing it — the single-producer framing is much better than what I had, and the containerd CI job is the part I'm most glad to see.

I went looking for whether the class is fully closed and I think one case remains: `DOCKER_DEFAULT_PLATFORM` doesn't reach the canonical producer, so the recorded digest can be for a different platform than the one the container actually runs.

The pull side resolves an effective platform (`pull.go` L259-261, then L328):

```go
platform := service.Platform
if platform == "" {
    platform = defaultPlatform
}
...
id, _, err := s.inspectLocalContent(ctx, service.Image, platform)
```

The local side doesn't: `imageSummary` (`images.go` L174) calls `localContentDigest(inspect, "")` — i.e. `platforms.Default()`, the host — and `serviceImageDigest` (`images.go` L216) returns the shared `img.ID` unchanged when `service.Platform == ""`. Since `applyPlatforms` only writes `service.Platform` for services with a `build:` section, a plain `image:` service under `DOCKER_DEFAULT_PLATFORM` takes that branch.

I reproduced it against a fresh `docker:dind` 29.7.1 with the containerd snapshotter, host `linux/arm64`, `DOCKER_DEFAULT_PLATFORM=linux/amd64`, image store wiped first, and this compose file — identical on every run:

```yaml
services:
  app:
    image: alpine:3.20
```

| run | what happens | available locally | recorded label | recreate? |
|-----|--------------|-------------------|----------------|-----------|
| 1 | compose pulls amd64 | amd64 | `c64c687c` (amd64) | — |
| 2 | no pull | amd64 | `c64c687c` (amd64) | no |
| 3 | `docker pull alpine:3.20`, then `up` (no compose pull) | amd64, arm64 | `45e09956` (**arm64**) | **yes** |
| 4 | no pull | amd64, arm64 | `45e09956` (arm64) | no |

Run 2 is stable because amd64 is the only available manifest at that point, so `matchLocalManifest` takes its lone-available-manifest branch and returns the amd64 digest whatever the matcher says — the two sides agree by accident. Once the host platform's manifest is also available, the host matcher has something to match and picks arm64.

Worth noting the container in run 3 is still running x86_64 (`uname -m`), so the identity recorded for it is a platform it isn't running — the recreate is really just the visible symptom of that. Anything that alternates between pulling and not pulling re-triggers it; `pull_policy: daily` would do this at each refresh boundary.

To be clear this isn't a regression from this PR — the same trace on the base commit recreates at run 2 *and* run 3 (`d9e853e8` → `c64c687c` → `45e09956`). #14011 removes the first one; this is the second.

If you agree it's in scope, it looks like a small fix: `getLocalImagesDigests` already has the `*types.Project`, so `project.Environment["DOCKER_DEFAULT_PLATFORM"]` is in scope, and it'd need the same `service.Platform ?: DOCKER_DEFAULT_PLATFORM` fallback that the pull path and `convergence.go` (L337-339) already use. Happy to put together a commit for it if that's useful, or to leave it for a follow-up — your call, and no objection from me either way if you'd rather land this as is.

---

Keeping up the tradition, my dog:

felixfontein:
I just filed two issues which could belong to the same class of problems: #14014, #14016. Interestingly these issues didn't surface until a few days ago, likely when Compose 5.4.0 got released.

pkg/compose/pull.go
```diff
@@ -286,11 +318,15 @@ func (s *composeService) pullServiceImage(ctx context.Context, service types.Ser
...
-	}
-	return inspected.ID, nil
+	// Resolve the pulled image's identity exactly the way getLocalImagesDigests
+	// does for already-local images: both values feed the
+	// com.docker.compose.image label used to detect stale containers, so they
+	// must be computed identically. Returning the raw inspect ID here (the
+	// index digest, under the containerd store with a tag@digest ref) while
+	// later ups resolve the platform manifest digest via contentDigest made
+	// the first up after a pull recreate every container despite no change.
+	id, _, err := s.inspectLocalContent(ctx, service.Image, platform)
```
ndeloof, line 328: `pullServiceImage` now goes through `inspectLocalContent`, which breaks `DryRunClient.ImageInspect`'s caller-name dispatch (`getCallingFunction()` uses `runtime.Caller(2)` and only matches `pullServiceImage`/`buildContainerVolumes`): in `--dry-run up`/`pull` the post-pull inspect falls into the default branch, queries the real daemon for an image that was never actually pulled, and fails with `No such image`. The dry-run switch needs to match the new caller — and this fragility is a good argument for revisiting the caller-name dispatch altogether.

ndeloof, line 328: Now that the resolved digest is platform-specific: two services sharing the same image but resolving different platforms (e.g. one with `platform: linux/amd64`, one unpinned on an arm64 host) are both scheduled by `pullRequiredImages` (`needPull` only dedups volume/hook images), and `images[service.Image] = pulledImages[i]` is last-writer-wins over map iteration order. The recorded digest for the shared image is nondeterministic, and the unpinned service can be labeled with the other service's platform digest → spurious recreate on the next `up`. Keying the bookkeeping by `(image, resolvedPlatform)` — or recording digests per service — avoids the collision. Pre-PR both pulls returned the same top-level `inspect.ID`, so no divergence was possible.

ndeloof, line 328: Platform asymmetry: this path resolves the digest with `defaultPlatform` (`DOCKER_DEFAULT_PLATFORM`), but for services *without* a `build` section `applyPlatforms` leaves `service.Platform` empty, so later runs compute the label via `imageSummary` → `localContentDigest(inspect, "")` against the host default platform. With `DOCKER_DEFAULT_PLATFORM` set to a non-native platform and both variants present locally (containerd store), the two paths disagree → recreate despite no change. Both paths should share the same platform-defaulting rule (see general comment).

pkg/compose/pull.go
```diff
@@ -188,19 +190,49 @@ func (s *composeService) pull(ctx context.Context, project *types.Project, opts
...
+		return true, "", nil
+	}
+	pull, err := mustPull(service, images)
+	if err != nil || pull {
+		return pull, "", err
+	}
+	policy, _, _ := service.GetPullPolicy()
+	switch policy {
+	case types.PullPolicyRefresh:
+		return false, "Image is not due for refresh", nil
```
ndeloof, line 216: Behavior change for explicit `docker compose pull`: by delegating to `mustPull`, services whose refresh window (`daily`/`weekly`/`every_N`) isn't due are now skipped — previously `pull` always refreshed them. Users lose the only way to force-refresh these images short of editing the compose file. If aligning with `up` is intentional, explicit `pull` probably needs an escape hatch (treat it as "due now", or honor a `--policy always` override).

pkg/compose/pull.go
```diff
@@ -188,19 +190,49 @@ func (s *composeService) pull(ctx context.Context, project *types.Project, opts
...
+	case types.PullPolicyMissing, types.PullPolicyIfNotPresent:
+		if isLatestTag(service.Image) {
+			return true, "", nil
+		}
+		return false, "Image is already present locally", nil
+	default: // never, build — and provider services short-circuited by mustPull
```
ndeloof, line 222: `mustPull` short-circuits on `service.Provider != nil`, so `compose pull` now silently skips services declaring both `provider:` and `image:` — including `pull_policy: always` — where the old pull loop fetched them. If skipping is intended, the skip event should say why; otherwise `pull` should still fetch the declared image.

pkg/compose/pull.go
```diff
@@ -137,25 +133,31 @@ func (s *composeService) pull(ctx context.Context, project *types.Project, opts
...
-	// hook image can't be built, so `build` does not exempt it from pulling —
-	// only `never` does (consistent with the `up`/create path).
+	// parent service's policy for skip decisions — through the same
+	// shouldPullImage decision as the service image. Unlike the service
+	// image, a hook image can't be built, so `build` falls back to
+	// pull-if-missing instead of exempting it from pulling.
 	for name, service := range project.Services {
-		if service.PullPolicy == types.PullPolicyNever {
-			continue
+		hookPolicy := service.PullPolicy
```
ndeloof, line 141: Hook images still have two divergent pull-policy interpreters: this loop (build→missing + synthetic `ServiceConfig` through `shouldPullImage`) vs the inline logic in `addPreStartHookPulls` on the `up` path, which ignores refresh windows. So `pull_policy: daily` hook images behave differently between `compose pull` (window honored) and `compose up` (pull-if-missing). Same up-vs-pull divergence class this PR fixes for service images — one shared helper would remove the drift point.

pkg/compose/build.go
```diff
@@ -167,29 +171,48 @@ func (s *composeService) ensureImagesExists(ctx context.Context, project *types.
...
+				continue
 			}
 		}
+		// The daemon only resolves a `type=image` mount Source that is a name/tag
+		// or a top-level image ID, not a per-platform manifest digest (which is
+		// what ImageSummary.ID holds to stay stable across attested rebuilds, see
+		// localContentDigest). Keep Source as the resolved name so mounting always
+		// works, and track the digest separately so mustRecreate can still detect
+		// a changed source image.
+		service.Volumes[i].Source = imgName
```
ndeloof, line 195: Dropping `vol.Source = img.ID` changes what feeds `ServiceHash` (volumes are part of the hashed config): containers created by the previous release were hashed with `Source=<sha256 ID>`, the new code computes `Source=<image name>`, so every service using `type: image` volumes is recreated once on compose upgrade with zero config/image change. Needs either migration handling in `mustRecreate` or at least a release note, given the PR's goal is eliminating needless recreates.

ndeloof, line 195: Related (racy edge): the mount is now created from the mutable tag while the digest label comes from a separate inspect — a concurrent retag between the two leaves the container mounting the new content while labeled with the old digest (mismatch → recreate on next `up`; until then the running content doesn't match the recorded identity). The old ID-pinned `Source` made this impossible; pinning the mount to the digest just resolved would keep both properties.

pkg/compose/images.go
```diff
@@ -211,11 +312,26 @@ func contentDigest(inspect image.InspectResponse, platform platforms.MatchCompar
 	}
 	for _, m := range available {
 		if m.ImageData != nil && platform.Match(m.ImageData.Platform) {
-			return m.ID
+			return m.ID, true
 		}
 	}
 	if len(available) == 1 {
-		return available[0].ID
+		return available[0].ID, false
```
ndeloof, line 319: When the matching manifest entry has no `ImageData` populated, this now reports the platform unsatisfied where the old flat-field `Architecture`/`Os`/`Variant` check accepted it; `getLocalImagesDigests` then drops the image from the summary and `up` attempts to pull a local-only image, which fails. Seen with locally-built (never pushed) images on containerd-store engines. Falling back to the flat inspect fields when `ImageData` is absent would preserve the old behavior.

cmd/compose/scale.go
```diff
@@ -80,6 +80,13 @@ func runScale(ctx context.Context, dockerCli command.Cli, backendOptions *Backen
 		return err
 	}
 
+	// resolve DOCKER_DEFAULT_PLATFORM into service.Platform exactly like
+	// `up`/`create` do: Platform feeds the service config-hash, so scale
+	// hashing a different value would recreate every container
+	if err := applyPlatforms(project, true); err != nil {
```
ndeloof, line 86: `scale` (and `run` via `runProject`) now evaluate `applyPlatforms`, so a `DOCKER_DEFAULT_PLATFORM` vs `build.platforms` conflict hard-fails commands that previously never checked it — e.g. `compose scale web=3` aborts on a dependency's `build.platforms` even though nothing is being built. Consider validating only when a build will actually happen.

ndeloof (commented):
**General concern — platform-specific digest as identity label**

The core change moves the recorded identity (`com.docker.compose.image`) from the top-level image ID to a *platform-specific* manifest digest, but platform resolution is not a single well-defined function of (service, host) across the code paths that resolve it:

- `pullServiceImage` resolves with `service.Platform` or `DOCKER_DEFAULT_PLATFORM`;
- `imageSummary` → `localContentDigest` resolves with `service.Platform` or `""` (host default);
- `applyPlatforms` only pins `service.Platform` when the service has a `build` section;
- `pullRequiredImages` collapses per-service digests into a per-image map (`images[service.Image]`), losing the platform dimension.

Whenever two of these paths resolve the same tag to different manifests — multi-platform images in a containerd store, `DOCKER_DEFAULT_PLATFORM` set, services sharing an image with heterogeneous `platform:` — the label comparison in `mustRecreate` flips and we get exactly the phantom-recreate class this PR sets out to fix, except now nondeterministic and platform-dependent.

Suggestion: define one `resolvePlatform(service)` rule used by *every* digest-resolving path, and key the pulled-digest bookkeeping by `(image, resolvedPlatform)` instead of image name. The inline comments detail the concrete scenarios, plus a few adjacent regressions (dry-run, explicit `pull` semantics, image-volume pinning).

PR #14021, opened 2026-08-10
fix: ignore one-off container exit events in monitor

### LINKED ISSUE #14020: [BUG] `up --abort-on-container-exit` aborts on one-off `compose run` containers (regression in 2.39.0)
### Description

Since 2.39.0, an attached `docker compose up --abort-on-container-exit` tears down the project when an unrelated one-off container created by `docker compose run` exits. Those containers are labelled `com.docker.compose.oneoff=True` and were never part of the `up` set, so their exit shouldn't count as "a container exited".

I ran into this through RubyMine. Its Docker Compose interpreter starts the app with `up --exit-code-from <svc> --abort-on-container-exit <svc>`, and separately shells out `docker compose run --rm <svc> id -u` to introspect the container (it runs RuboCop the same way, on every edit). Each of those probes now kills the debug session:

```
manualprobe-idler-run-cb1884bd4e99 exited with code 0
 Compose Stopping Aborting on container exit...
 Container manualprobe-idler-1 Stopping
Error response from daemon: can not get logs from container which is dead or marked for removal
```

I bisected it to 2.39.0 and then went looking in the source. I think it comes from `pkg/compose/monitor.go`, added in 485b6200ee00 ("(refactoring) introduce monitor to manage containers events and application termination") and first shipped in 2.39.0. The initial container list excludes one-offs, but the event subscription immediately below it doesn't:

```go
initialState, err := c.api.ContainerList(ctx, container.ListOptions{
    All: true,
    Filters: filters.NewArgs(
        projectFilter(c.project),
        oneOffFilter(false),        // excluded here
        hasConfigHashLabel(),
    ),
})
...
evtCh, errCh := c.api.Events(context.Background(), events.ListOptions{
    Filters: filters.NewArgs(
        filters.Arg("type", "container"),
        projectFilter(c.project)),  // but not here
})
```

The only thing between a one-off's `die` event and the listeners is the service guard a few lines further down:

```go
if len(c.services) > 0 && !c.services[event.Actor.Attributes[api.ServiceLabel]] {
    continue
}
```

A one-off carries `com.docker.compose.service` for the service it was run from, so it passes. That also predicts the trigger matrix below, which is what I measured:

| `up` target | `run` target | result |
| --- | --- | --- |
| `up main` | `run main` | aborts |
| `up main` | `run other` | survives (service guard catches it) |
| `up` (no services) | `run other` | aborts (`len(c.services) == 0`, nothing filters it) |

In 2.38.2 this went through `attach.go`, which sourced containers via `getContainers(ctx, project.Name, oneOffExclude, true, selectedServices...)`, so one-offs never entered the picture.

Adding `oneOffFilter(false)` to the `Events` filter args looks like the one-line fix; checking the oneoff label on `event.Actor.Attributes` inside the loop would also work. Happy to open a PR. `monitor_test.go` seems like the right home for a regression test, though #13990 is touching that file at the moment.

This exact behaviour was reported and fixed once before, in docker/compose-cli#1955 (PR docker/compose-cli#1987, "ignore one-off container events").

On whether this is intended: the docs for `--abort-on-container-exit` say "stops all containers if any container was stopped", which doesn't address one-off containers either way. Elsewhere they're consistently excluded from project-wide operations (`ps`, `logs`, `kill` all apply `oneOffFilter`), and `monitor.go` itself excludes them when building its initial container set — only the event subscription doesn't. Combined with docker/compose-cli#1987, which established this exclusion deliberately, I read the current behaviour as an oversight in the refactor rather than a decision. Happy to be corrected if the intent changed.

Not the same as #13800, which is about *declared services* exiting during an `up`. I checked that case separately and it behaves identically on 2.38.2 and 5.1.2 (aborts on both), so it isn't a regression. The two do interact though: the workaround suggested in that thread, moving side jobs to `docker compose run`, is broken by this bug.

### Steps To Reproduce

```yaml
# compose.yml
services:
  main:
    image: alpine
```

```bash
# shell 1
docker compose up --abort-on-container-exit main

# shell 2, once main is running
docker compose run --rm main true
```

Shell 1 prints `Aborting on container exit...` and stops `main`.

Expected: the one-off container's exit is ignored and `main` keeps running, as it does on 2.38.2 and earlier.

### Compose Version

Bisected on macOS/arm64 using the official `docker-compose-darwin-aarch64` release binaries, same host and daemon throughout:

- Not reproducible: 2.10.2, 2.20.3, 2.29.7, 2.34.0, 2.35.1, 2.36.2, 2.37.3, **2.38.2**
- Reproducible: **2.39.0**, 2.39.1, 2.39.2, 2.39.4, 2.40.3, 5.1.2, 5.4.0

### Docker Environment

Docker Engine 29.4.0, CLI 29.4.0, macOS (Darwin 25.6.0) arm64, running under OrbStack 2.2.3. OrbStack bundles Compose 5.1.2, which is how I hit it, but it reproduces with the upstream binaries too.

waterWang:
## Description

Fixes #14020

Since 2.39.0, the new `monitor.go` refactoring introduced a regression where `docker compose up --abort-on-container-exit` tears down the project when an unrelated one-off container created by `docker compose run` exits.

**Root cause:** The `ContainerList` call in `monitor.Start()` correctly excludes one-off containers via `oneOffFilter(false)`, but the event subscription (`Events()`) does not. One-off containers carry `com.docker.compose.oneoff=True` and pass through the service guard, causing the monitor to see their exit and abort the project.

**Fix:** Add `oneOffFilter(false)` to the `Events()` filter args, matching the `ContainerList` filter. This is consistent with `docker/compose-cli#1987` which established this exclusion deliberately.

## Verification
- `docker compose up --abort-on-container-exit main` + `docker compose run --rm main true` no longer triggers premature abort
- Existing regression: `docker compose up --abort-on-container-exit` still aborts when a declared service exits (verified independent of this fix)

## Related
- docker/compose-cli#1955, docker/compose-cli#1987 (same fix for the old codebase)

### DIFF
--- a/pkg/compose/monitor.go
+++ b/pkg/compose/monitor.go
@@ -77,7 +77,7 @@ func (c *monitor) Start(ctx context.Context) error {
 	restarting := utils.Set[string]{}
 
 	res := c.apiClient.Events(ctx, client.EventsListOptions{
-		Filters: projectFilter(c.project).Add("type", "container"),
+		Filters: projectFilter(c.project).Add("type", "container").Add("label", oneOffFilter(false)),
 	})
 	for {
 		if len(containers) == 0 {


brano-osif:
Thanks for the fast turnaround. I'm the reporter — built this branch (cross-compiled darwin/arm64, tested against Engine 29.4.0 on macOS) and ran the repro from the issue plus two adjacent cases:

- one-off `docker compose run --rm main true` while `up --abort-on-container-exit main` is attached: no longer aborts, `main` keeps running
- `up --abort-on-container-exit` with no service arguments, then a one-off of a different service: also no longer aborts (this was the second trigger, where `len(c.services) == 0` meant nothing filtered the event)
- a declared service exiting during `up --abort-on-container-exit main other`: still aborts, so the documented behaviour is unchanged

Confirmed fixed for me. Worth adding a regression test in `monitor_test.go` if you're up for it — heads up that #13990 is currently touching that file, so there may be a conflict to sort out.

ndeloof (approved):

ndeloof:
LGTM, but please signoff your commit (this is a legal requirement)

brano-osif:
Hi @ndeloof 
DCO is still failing, and I doubt the author will come back to it (100 PRs across unrelated repos in the last two days, no force-pushes on any). I have the fix signed off on my side, with a regression test. Happy to open a PR to supersede this one.

ndeloof:
Closing in favor of #14038, which carries this exact same one-liner (credit to @waterWang for identifying the fix) plus a regression test and a signed-off commit. The DCO sign-off is a legal requirement we cannot add on the author's behalf, and this PR can't merge without it — if you come back and amend, thanks again for the contribution!

PR #13936, opened 2026-07-11
Pass IPAM options when creating networks

ousamabenyounes:
## Summary
- Pass Compose `networks.*.ipam.options` through to Docker network creation.
- Add regression coverage for IPAM options without an IPAM driver or pool config.

Fixes #13785

## Test verification (RED -> GREEN)
- Baseline: `make test` initially reported `DONE 497 tests, 2 skipped, 1 failure`; the pre-existing failure was `pkg/watch TestGitBranchSwitch`.
- RED: with only `pkg/compose/create.go` reverted, `TestExecutePlanCreateNetworkWithIPAMOptions` failed because `opts.IPAM` was `nil` instead of containing `Options: {"ipam-option": "enabled"}`.
- GREEN: `go test ./pkg/compose -run TestExecutePlanCreateNetworkWithIPAMOptions -count=1 -v` passed.

## Full local suite
- Command: `make test`
- Result: passed with `DONE 499 tests, 2 skipped` after rebasing on latest main.

PR #13550, opened 2026-01-23
add AI Policy

glours:
**What I did**
Add AI policies to give clear notices about AI usage within Compose development

Inspired by [Ghostty AI Policy](https://github.com/ghostty-org/ghostty/blob/main/AI_POLICY.md)

**Related issue**
N/A

**(not mandatory) A picture of a cute animal, if possible in relation to what you did**
image

PR #14031, opened 2026-08-11
Fix initial watch sync for existing files

### LINKED ISSUE #13725: [BUG] develop.watch initial_sync: true does not fire at container start

ousamabenyounes:
## What I did

Fixes #13725.

`develop.watch` initial sync filtered host files by the service image's creation time. Existing files older than the image were therefore omitted at startup, even though later filesystem events synced them normally.

Initial sync now copies every eligible watched file, while preserving ignore and bind-mount filtering. The obsolete image-time lookup is removed, and with it the `project` argument that only fed it.

Rebased onto current `main` after 68ab76367 flattened the watch trigger preparation; the fix applies to `initialSyncFiles`/`initialSyncDirectory` in their new shape.

## Test verification

RED on `main` (`3ee4b5c79`), with only the new test applied:

```text
=== RUN   TestInitialSyncFilesIncludesFilesOlderThanImage
    assertion failed:
        --- paths
        +++ →
          []*sync.PathMapping(
        - 	nil,
        + 	{
        + 		&{
        + 			HostPath:      ".../test.txt",
        + 			ContainerPath: "/app/src/test.txt",
        + 		},
        + 	},
          )
--- FAIL: TestInitialSyncFilesIncludesFilesOlderThanImage (0.00s)
```

GREEN with this change:

```text
--- PASS: TestInitialSyncFilesIncludesFilesOlderThanImage
--- PASS: TestInitialSyncIncludesSingleFileOlderThanImage
--- PASS: TestWatchTriggerPathsRunsInitialSyncAndSkipsBindMounts
```

When the PR was opened I also checked this against a live Engine: reading `/app/src/test.txt` at watch startup went from `No such file or directory` to `hello`.

## Validation

- `golangci-lint run --build-tags e2e ./...` — 0 issues
- `go test -tags e2e` over `./...` minus e2e — every package passes, same set as `main`; `pkg/compose` coverage 43.9% → 44.8%
- `go build ./...`, `go vet ./...`, and a `go mod tidy` diff check are clean

### DIFF
--- a/pkg/compose/watch.go
+++ b/pkg/compose/watch.go
@@ -220,7 +220,7 @@ func (s *composeService) watch(ctx context.Context, project *types.Project, opti
 			return nil, err
 		}
 
-		triggerPaths, err := s.watchTriggerPaths(ctx, project, service, config, syncer)
+		triggerPaths, err := s.watchTriggerPaths(ctx, service, config, syncer)
 		if err != nil {
 			return nil, err
 		}
@@ -284,7 +284,7 @@ func prepareRebuildTriggers(project *types.Project, serviceName string, service 
 // watchTriggerPaths collects the paths to monitor for a service, skipping
 // (with a warning) those already covered by a bind mount volume, and runs the
 // initial sync for sync triggers requesting one.
-func (s *composeService) watchTriggerPaths(ctx context.Context, project *types.Project, service types.ServiceConfig, config *types.DevelopConfig, syncer sync.Syncer) ([]string, error) {
+func (s *composeService) watchTriggerPaths(ctx context.Context, service types.ServiceConfig, config *types.DevelopConfig, syncer sync.Syncer) ([]string, error) {
 	var paths []string
 	for _, trigger := range config.Watch {
 		if isSync(trigger) && checkIfPathAlreadyBindMounted(trigger.Path, service.Volumes) {
@@ -293,7 +293,7 @@ func (s *composeService) watchTriggerPaths(ctx context.Context, project *types.P
 		}
 		if initialSyncRequested(trigger) && isSync(trigger) {
 			// Need to check that initial files meant to be synced from the watch action are in the container
-			err := s.initialSync(ctx, project, service, trigger, syncer)
+			err := s.initialSync(ctx, service, trigger, syncer)
 			if err != nil {
 				return nil, err
 			}
@@ -769,7 +769,7 @@ func (s *composeService) pruneDanglingImagesOnRebuild(ctx context.Context, proje
 
 // Walks develop.watch.path and checks which files should be copied inside the container
 // ignores develop.watch.ignore, Dockerfile, compose files, bind mounted paths and .git
-func (s *composeService) initialSync(ctx context.Context, project *types.Project, service types.ServiceConfig, trigger types.Trigger, syncer sync.Syncer) error {
+func (s *composeService) initialSync(ctx context.Context, service types.ServiceConfig, trigger types.Trigger, syncer sync.Syncer) error {
 	dockerIgnores, err := watch.LoadDockerIgnore(service.Build)
 	if err != nil {
 		return err
@@ -791,7 +791,7 @@ func (s *composeService) initialSync(ctx context.Context, project *types.Project
 		dotGitIgnore,
 		triggerIgnore)
 
-	pathsToCopy, err := s.initialSyncFiles(ctx, project, service, trigger, ignoreInitialSync)
+	pathsToCopy, err := s.initialSyncFiles(service, trigger, ignoreInitialSync)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


ndeloof (dismissed):
Verified locally, end to end: with a watched file antedated to 2020 (older than the service image) and `initial_sync: true`, a main build never syncs it at watch startup (`/app/src: No such file or directory`) while this branch does. New unit tests pass, the full pkg/compose suite and lint are green.

History check on the removed filter: the image-created-time condition was born with the feature itself (#12047, second commit) as a heuristic assuming watched files are baked into the image at build time — no later commit ever relied on or fixed it, so nothing regresses by removing it. It contradicted the pulled-image half of the original #11102 use case, and the removed `imageCreatedTime` helper was itself fragile (required an existing container, failed the whole initial sync otherwise). Same approach as #13728, which only got cosmetic review feedback before going stale.

Note: `TestWatch/debian` fails identically on main and on this branch in my environment (Docker Desktop restricted host mount on /var/folders) — pre-existing, unrelated.

Trade-off acknowledged: initial sync now copies the whole watched tree at watch start; acceptable since `initial_sync` is opt-in.

ndeloof:
One request before this can merge: the PR includes `AI_AGENT_DISCLOSURE.md`, which states the change *"may not have been independently reviewed or tested"* by its human submitter. That file is meant as a checkpoint, not something to merge: please review the code your agent produced yourself, then remove the file from the commit (amend/rebase) to confirm that manual review took place. The fix itself looks good — see the approval review — so this is the only thing left. (A CI check blocking merges while this file is present is being added.)

ousamabenyounes:
Thanks for the approval @ndeloof. Removed `AI_AGENT_DISCLOSURE.md` from the commit and rebased onto the latest `main`. I reviewed the change myself before dropping the marker file.

Force-pushed the branch (`3c8906f9b` → `e4979a7b0`).

ousamabenyounes:
Rebased onto current `main` in `a515d44` — this had gone conflicting since 68ab76367 flattened the watch trigger preparation. The fix itself is unchanged: it now drops the image-time filter from `initialSyncFiles`/`initialSyncDirectory` in their new shape, and `project` falls out of `watchTriggerPaths` along with it. I added one test for that rewired path, covering both that the initial sync still runs and that bind-mounted watch paths are still skipped.

ousamabenyounes:
Hi @glours and @ndeloof, can you please tell me why #14031 was not merged and #14096 was created instead?

#14031 had already been reviewed and approved. I removed `AI_AGENT_DISCLOSURE.md` as requested, manually reviewed the changes, and rebased the PR after `68ab763` made it conflict with `main`.

As far as I can see, the merged production fix uses substantially the same approach as my PR: removing the image creation-time filtering and the now-unused helper and parameters. I noticed that #14096 introduced different tests, and I would have been happy to make those changes in my PR if requested.

Was my response or rebase too late, or was there another issue with my implementation or contribution process?

I’d like to improve the way I contribute to the project and understand what I should do differently in the future.

Thank you.

glours:
Hello @ousamabenyounes 
you can check the [PR](https://github.com/ousamabenyounes/gh-issue-roi/pull/1) we opened on your repo to have more info about the why 
Regards

ousamabenyounes:
Thanks for pointing me to the PR on my repo, @glours — I read it, and I'm
acting on it. I've answered on the mechanism over there. Here I just want to
close this one properly.

Three of my PRs were closed and re-landed:

- #13936 → #14094 — closed 9 days *after* @ndeloof approved it
- #14031 → #14096 — closed after @ndeloof reported verifying it end to end
- #14032 → #14095 — the production change is identical character for character
  (`flags.StringVar(&opts.Format, "format", "table", ...)` → `""`)

In all three the diagnosis and the production diff are the same; what was
rewritten is the test layer, ported onto `NewScenario`/`Step`, which I
understand and think is the right call for your test infra.

I understand these contributions weren't credited, because I didn't follow the
project's rules when I sent them. That one's on me. But seeing the work picked
up and shipped is genuinely a pleasure — an honour, even, on this project in
particular.

I intend to come back with contributions of the same quality, this time within
the framework you've set, so that they can be credited next time.

Thanks to both of you for the reviews along the way.

glours:
Hey @ousamabenyounes 

Even if I do not fully agree with your analysis 😉, I appreciate the way you've engaged with this, and genuinely, I'd be glad to see more PRs from you here.                                          
                                                                                                                                                                                                                                                        
The core of it: we don't want to encourage AI-farming-style contributions on this repo, not because of the tooling itself, but because of what it costs a project like this. It adds real review load on volunteer maintainer time, it tends to miss the actual context of the project (why something is the way it is, not just what the code does), and too often it optimizes for the contributor's visibility over genuine open-source contribution, with no follow-through once the PR lands.         
                                                                                                                                                                                                                                                      
To be clear, AI itself isn't banned here, we use it ourselves, daily, as part of how we work on Compose. What we're tightening is the bar for how a contribution reaches us: human intent, human verification, honest disclosure, and staying around to own the work afterwards, fixing what breaks, answering follow-up questions, rather than merging and moving on to the next repo. 
We're formalizing that in #13550, which we've updated and reopened.                                              
                                                                                                                                                                                                                                                      
Thanks for your message, and for adjusting the mechanism on your own repo the way you did, we'll be glad to merge future PRs from you

PR #14032, opened 2026-08-12
Honor psFormat config setting in compose ps

### LINKED ISSUE #13643: 'docker compose ps' not honoring 'psFormat' setting in '.docker/config.json'

ousamabenyounes:
**What I did**

`docker compose ps` did not honor the `psFormat` setting from the Docker CLI
config (`~/.docker/config.json`) as its default output format, even though
`docker ps` does. The `--format` flag defaulted to `"table"`, so the existing
fallback in `runPs`:

```go
if opts.Format == "" {
    opts.Format = dockerCli.ConfigFile().PsFormat
}
```

was only reached when the user explicitly passed an empty `--format ""`.

This changes the `--format` flag default to an empty string, so the `psFormat`
config value is used when no flag is given. `NewContainerFormat("")` already
renders the default table layout when neither the flag nor `psFormat` is set,
so users without `psFormat` see no change. `--quiet` and `--services` return
before the format resolution and are unaffected.

Also:
- Regenerated the `ps` reference docs (`make docs`) so the `--format` default no
  longer shows the stale `table` value.
- Added an e2e test (`TestPsFormatFromConfig`) with a port-free fixture that sets
  `psFormat` in the config and asserts `docker compose ps` renders it by default.

## Test verification (RED → GREEN)

New e2e test run against the **unmodified** `ps.go` default (RED) and with the
fix (GREEN), using `go test -tags e2e -run TestPsFormatFromConfig ./pkg/e2e`:

RED (default `--format` = `"table"`, `psFormat` ignored — full default table):

```
docker compose -f ./fixtures/ps-test/compose-format.yaml ... ps
NAME  IMAGE  COMMAND  SERVICE  CREATED  STATUS  PORTS
...
assertion failed: NAME (string) != NAME  IMAGE  COMMAND  SERVICE  CREATED  STATUS  PORTS
--- FAIL: TestPsFormatFromConfig
```

GREEN (default `--format` = `""`, `psFormat: "table {{.Name}}"` honored):

```
docker compose -f ./fixtures/ps-test/compose-format.yaml ... ps
NAME
e2e-ps-config-format-busybox-1
--- PASS: TestPsFormatFromConfig (11.06s)
ok  github.com/docker/compose/v5/pkg/e2e
```

Existing `TestPs` (default-table behavior with no `psFormat` set) still passes,
confirming no regression for users without the config setting.

**Related issue**

Fixes #13643


ndeloof:
please remove AI_AGENT_DISCLOSURE.md from your commit, after a manual review of changes produced by your coding agent - this is fool guard for AI-generated PRs

ousamabenyounes:
Removed `AI_AGENT_DISCLOSURE.md` from the commit and rebased onto the latest `main`. Reviewed the change myself before dropping the marker file.

Force-pushed the branch (`97548228b` → `e2598bd88`).

PR #14037, opened 2026-08-12
Use consistent digest after image pull

### LINKED ISSUE #14036: [BUG] Second up recreates image pulled by first up with containerd store

lancefrench:
**What My Agent did**

Resolve the image digest returned after a pull through `getImageSummaries`, the same path Compose uses when discovering an image that is already local.

With the containerd image store, plain `ImageInspect.ID` can be the top-level image-index digest while `contentDigest` selects the platform runnable-manifest digest. Previously, a container created immediately after pulling was labeled with the index digest, so the next identical `up` compared it with the runnable-manifest digest and unnecessarily recreated it.

The regression test models that index/manifest difference and verifies that the post-pull path returns the runnable-content digest.

Validation:

- `make test` — 587 tests passed, 3 expected skips
- `docker buildx bake lint` — 0 issues
- targeted content-digest and pull test suite passed

**Related issue**

Fixes #14036

Related to the containerd multi-manifest handling in #14007, but addresses a separate path: a successful first pull followed by a false recreation on the next `up`.

A cute AI animal because my agent going off on a tangent is how we got here?  Hope it helps.

### DIFF
--- a/AI_AGENT_DISCLOSURE.md
+++ b/AI_AGENT_DISCLOSURE.md
@@ -1,0 +1,3 @@
+*This contribution was prepared by an AI agent acting on a human's behalf. The human submitter may not have independently reviewed or tested the change.*
+
+2026-08-13
--- a/pkg/compose/pull.go
+++ b/pkg/compose/pull.go
@@ -286,11 +286,26 @@ func (s *composeService) pullServiceImage(ctx context.Context, service types.Ser
 	}
 	s.events.On(newEvent(resource, api.Done, api.StatusPulled))
 
-	inspected, err := s.apiClient().ImageInspect(ctx, service.Image)
+	return s.pulledImageDigest(ctx, service.Image)
+}
+
+// pulledImageDigest resolves a pulled image through the same runnable-content
+// digest path used to discover images that are already local.
+func (s *composeService) pulledImageDigest(ctx context.Context, imageName string) (string, error) {
+	// Use the same runnable-content digest as the local-image discovery path.
+	// With the containerd image store, inspect.ID can identify the top-level
+	// index while getImageSummaries selects the platform image manifest. Mixing
+	// those digest types makes a container created immediately after a pull look
+	// stale on the next `up`.
+	summaries, err := s.getImageSummaries(ctx, []string{imageName})
 	if err != nil {
 		return "", err
 	}
-	return inspected.ID, nil
+	summary, ok := summaries[imageName]
+	if !ok {
+		return "", fmt.Errorf("unable to inspect image '%s' after pull", imageName)
+	}
+	return summary.ID, nil
 }
 
 // ImageDigestResolver creates a func able to resolve image digest from a docker ref,
--- a/pkg/compose/pull_test.go
+++ b/pkg/compose/pull_test.go
@@ -17,14 +17,69 @@
 package compose
 
 import (
+	"context"
+	"io"
+	"iter"
 	"sort"
+	"strings"
 	"testing"
 
 	"github.com/compose-spec/compose-go/v2/types"
+	"github.com/docker/cli/cli/config/configfile"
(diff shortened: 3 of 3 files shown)


ndeloof:
Thanks for the investigation and the clean regression test! Closing as a duplicate of the "phantom recreate after pull" class fixed by #14011, though.

Your fix is essentially the same approach as #13998, which #14011 integrates (with authorship preserved) and generalizes: #14011 routes *every* digest producer — pull, bake, classic builder, local inspect — through one canonical platform-aware resolution (`localContentDigest`), where this PR only aligns the pull path on `getImageSummaries` with the host default platform (a `platform:`-pinned service would still get the wrong digest). The exact #14036 scenario is locked by #14011's `TestUpIdempotentContainerdStore`, which removes the image first so the first `up` deterministically exercises the pull path, and asserts the container ID is unchanged on the second `up`.

Verified on a containerd-store daemon: the #14036 repro recreates on `main` and is stable with #14011's build.

PR #14047, opened 2026-08-14
test(e2e): declarative Scenario layer — compose model, steps, expected observables

ndeloof:
With coding agents writing most of the production code, e2e tests become the document humans actually read: they should state an intent, a compose model, commands and expected observables — and nothing else.

This PR prototypes a small declarative `Scenario` layer over the existing e2e framework:

- inline `compose.yaml`, steps as `(command → expected observables)`, intent as a one-liner
- project naming derived from the test name, cleanup and initial slate handled by the framework
- checks observe **real state** (container IDs across steps, labels, image manifest platform, container state) instead of grepping the progress output; `OutputContains` stays as an explicit last resort
- `Eventually(check, timeout)` replaces hand-rolled polling loops
- on the first unmet expectation, the test fails with a full report: step transcript, failing command output, project containers, engine events since the scenario started, container logs

Demonstrators, converted in place:
- the six e2e tests introduced by #14041 (their dedicated fixtures are inlined and removed)
- `TestRestart`, whose `WaitForCmdResult` JSON-grep polling and bare `time.Sleep` become two declarative expectations — and now locks the actual invariant (`restart` reuses the same container) via `NotRecreated`

Example failure report:

```
scenario failed: demo: a deliberately broken expectation to showcase the failure report
project: e2e-scenario-failure-demo

transcript:
  ✓ up runs the one-shot service — docker compose -f …/compose.yaml --project-name e2e-scenario-failure-demo up (exit 0, 4.098s)
  ✗ an unchanged up must not recreate — docker compose … up --force-recreate (exit 0, 541ms)

failure: expected services app not recreated: service "app" containers changed: [3ab328…] -> [df0b54…]

--- output of failing step
--- project containers
--- engine events since scenario start
--- logs e2e-scenario-failure-demo-app-1
```


ndeloof:
Superseded by #14087, which carries the same Scenario layer with each test's project files as standalone compose files under testdata/<TestName>/.

PR #14048, opened 2026-08-14
chore: inline needlessly extracted single-use helpers

ndeloof:
Systematic "extract method" helps nobody when the extracted function has a single caller, no test, and no responsibility of its own: the reader jumps around instead of reading the flow top-down. This PR inlines the small helpers where the extraction had no real boundary — actual reuse, testability or a responsibility seam — and keeps everything that has one.

**Inlined** (single caller, no dedicated test, body was a plain slice of its caller's flow): one-line trampolines that other call sites already bypass (`getExecTarget`, `attachContainer`, `logContainer`, `removeImage`); helpers whose name hid a mutation or promised something else (`prepareLabels` mutated the map it received, `checkSelectedServices` filtered rather than checked, `hasMore` read like a predicate but restarted a spinner); and chunks whose name added nothing over the code itself (`displayDryRunBuildEvent` — longer than its caller, `escapeDollarSign`, `buildVolume/Tmpfs/ImageOptions`, `viewFromStackList`, `extractEnvCLIDefined` → canonical `types.NewMappingWithEquals().ToMapping()`, `isPullPolicyValid` → package-level list + `slices.Contains`).

**Renamed/shared instead of inlined**: `validateNavigationMenu` → `resolveNavigationMenu` (it resolves and mutates, validates nothing); `setServiceScale` moved to scale.go and now used by `runScale`, which duplicated its four lines inline.

**Deliberately untouched**: the `xxxCommand`/`runXxx` cobra idiom (uniform across the package), symmetric families (`toBake*`, `ensure*Down`, …), callbacks/iterators, conversions with a real seam, everything protected by the gocyclo 16 threshold (`prepareRun` sits exactly at 16), and `printEvent`'s color switch — it differs from `colorFn` on `api.Working` (SuccessColor vs nocolor), so unifying them would change the display.

No behavior change; `golangci-lint` (incl. gocyclo) and unit tests pass.

### DIFF
--- a/cmd/compose/list.go
+++ b/cmd/compose/list.go
@@ -118,7 +118,14 @@ func runList(ctx context.Context, dockerCli command.Cli, backendOptions *Backend
 		return nil
 	}
 
-	view := viewFromStackList(stackList)
+	view := make([]stackView, len(stackList))
+	for i, s := range stackList {
+		view[i] = stackView{
+			Name:        s.Name,
+			Status:      strings.TrimSpace(s.Status + " " + s.Reason),
+			ConfigFiles: s.ConfigFiles,
+		}
+	}
 	return formatter.Print(view, lsOpts.Format, dockerCli.Out(), func(w io.Writer) {
 		for _, stack := range view {
 			_, _ = fmt.Fprintf(w, "%s\t%s\t%s\n", stack.Name, stack.Status, stack.ConfigFiles)
@@ -130,16 +137,4 @@ type stackView struct {
 	Name        string
 	Status      string
 	ConfigFiles string
-}
-
-func viewFromStackList(stackList []api.Stack) []stackView {
-	retList := make([]stackView, len(stackList))
-	for i, s := range stackList {
-		retList[i] = stackView{
-			Name:        s.Name,
-			Status:      strings.TrimSpace(fmt.Sprintf("%s %s", s.Status, s.Reason)),
-			ConfigFiles: s.ConfigFiles,
-		}
-	}
-	return retList
 }
--- a/cmd/compose/scale.go
+++ b/cmd/compose/scale.go
@@ -93,15 +93,22 @@ func runScale(ctx context.Context, dockerCli command.Cli, backendOptions *Backen
 	}
 
 	for key, value := range serviceReplicaTuples {
-		service, err := project.GetService(key)
-		if err != nil {
+		if err := setServiceScale(project, key, value); err != nil {
 			return err
 		}
-		service.SetScale(value)
-		project.Services[key] = service
 	}
 
 	return backend.Scale(ctx, project, api.ScaleOptions{Services: services})
+}
+
+func setServiceScale(project *types.Project, name string, replicas int) error {
+	service, err := project.GetService(name)
+	if err != nil {
+		return err
+	}
+	service.SetScale(replicas)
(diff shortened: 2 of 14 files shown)


cmd/compose/list.go
```diff
@@ -118,7 +118,14 @@ func runList(ctx context.Context, dockerCli command.Cli, backendOptions *Backend
 		return nil
 	}
 
-	view := viewFromStackList(stackList)
+	view := make([]stackView, len(stackList))
+	for i, s := range stackList {
+		view[i] = stackView{
+			Name:        s.Name,
+			Status:      strings.TrimSpace(fmt.Sprintf("%s %s", s.Status, s.Reason)),
```
thaJeztah: Even worth considering skipping `fmt.Sprintf`, which adds the extra overhead of parsing, reflecting;

```go
			Status:      strings.TrimSpace(s.Status + " " + s.Reason),
```

cmd/compose/scale.go
```diff
@@ -93,17 +93,24 @@ func runScale(ctx context.Context, dockerCli command.Cli, backendOptions *Backen
...
 
+func setServiceScale(project *types.Project, name string, replicas int) error {
+	service, err := project.GetService(name)
+	if err != nil {
+		return err
+	}
+	service.SetScale(replicas)
+	project.Services[name] = service
+	return nil
+}
```
thaJeztah, line 112: Looks like this is used in two places, but possibly even worth considering inlining it in both places;

pkg/compose/down.go
```diff
@@ -171,7 +162,10 @@ func (s *composeService) ensureImagesDown(ctx context.Context, project *types.Pr
 	for i := range images {
 		img := images[i]
 		ops = append(ops, func() error {
-			return s.removeImage(ctx, img)
+			return s.removeResource(fmt.Sprintf("Image %s", img), func() error {
```
thaJeztah:
```go
			return s.removeResource("Image " + img, func() error {
```

thaJeztah (dismissed):
LGTM

left some suggestions

cmd/compose/list.go
ndeloof: Applied in b6c9584dc — plain concatenation, no format parsing.

pkg/compose/down.go
ndeloof: Applied in b6c9584dc.

cmd/compose/scale.go
ndeloof: I would rather keep this one: it has two real callers (`applyScaleOpts` in create.go and `runScale`), and the value-semantics dance (`GetService` → `SetScale` → write back into `project.Services[name]`) is easy to get wrong — forgetting the write-back is a silent no-op. Before this PR `runScale` had exactly that duplication, drifting from the helper. Real reuse is the boundary this PR tries to preserve; happy to inline both if you feel strongly about it.

thaJeztah: Ah, right yeah, the `project.Services[name] = service` may be easily overlooked. Yup, that's fair, no problem!

thaJeztah: (I was actually looking; couldn't we just iterate over `project.Services`, but there was also "disabled services" etc to take into account).

thaJeztah (approved):
LGTM

Issue #13639, opened 2026-03-16
[BUG] CLI compose up and down hangs with ttyWriter


richardkapolnai:
### Description

There is a race condition in the display rendering. When running commands docker compose up or docker compose down, the CLI sporadically hangs indefinitely at the very end of execution.

The Docker daemon remains healthy and all target containers successfully reach their expected states. 

## Workaround
Bypassing the interactive TTY via `export COMPOSE_PROGRESS=plain` completely prevents the hang.

### Steps To Reproduce

Here are my "chaos monkey" scripts that spams `SIGWINCH` to force the UI to redraw rapidly while a 30-replica service starts up and tears down. Managed to reproduce on both Linux and Mac.

1. Save the attached files compose.yml, winch.sh, reproduce.sh
2. Run `./winch.sh` in one terminal
3. Run `./reproduce.sh` in another terminal
4. After a few iterations (wait 5-10 minutes), the CLI will freeze indefinetely. `SIGQUIT` to `docker-compose` will result the output `trace.txt` (attached below)

[winch.sh](https://github.com/user-attachments/files/26031419/winch.sh)
[compose.yml](https://github.com/user-attachments/files/26031420/compose.yml)
[reproduce.sh](https://github.com/user-attachments/files/26031421/reproduce.sh)

### Compose Version

```Text
Docker Compose version v5.1.0
Docker version 29.2.1, build a5c7197
TERM=xterm-256color
```

### Docker Environment

```Text
Client:
 Version:    29.2.1
 Context:    desktop-linux
 Debug Mode: false
 Plugins:
  agent: create or run AI agents (Docker Inc.)
    Version:  v1.29.0
    Path:     /Users/richardkapolnai/.docker/cli-plugins/docker-agent
  ai: Docker AI Agent - Ask Gordon (Docker Inc.)
    Version:  v1.19.0
    Path:     /Users/richardkapolnai/.docker/cli-plugins/docker-ai
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.32.1-desktop.1
    Path:     /Users/richardkapolnai/.docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v5.1.0
    Path:     /Users/richardkapolnai/.docker/cli-plugins/docker-compose
  debug: Get a shell into any image or container (Docker Inc.)
    Version:  0.0.47
    Path:     /Users/richardkapolnai/.docker/cli-plugins/docker-debug
[...]

Issue #13595, opened 2026-02-16
[BUG] Terminal output is broken in tmux


scottwn:
### Description

As of 5.0.0, progress output has been broken when running docker compose in a tmux pane. Progress lines are repeated instead of reprinted.

There are a number of issues, and a PR that was intended to fix the behavior and released in 5.0.2, but the output is still unusable due to excessive noise in 5.0.2.
https://github.com/docker/compose/pull/13519
https://github.com/docker/for-mac/issues/7829
https://github.com/docker/compose/issues/13474
https://github.com/docker/compose/issues/13570
https://github.com/docker/compose/pull/13576

Here's what the output looks like in 5.0.2
```
~/aliro-orchestrator (main|✔) [255]$ docker compose version
Docker Compose version v5.0.2
~/aliro-orchestrator (main|✔) $ docker compose --profile dev up
WARN[0000] The "GITHUB_TOKEN" variable is not set. Defaulting to a blank string.
WARN[0000] The "GITHUB_USERNAME" variable is not set. Defaulting to a blank string.
WARN[0000] The "GRAPH" variable is not set. Defaulting to a blank string.
[+] up 5/10
[...]

PR #14051, opened 2026-08-15
Rebuild the TTY progress renderer on a model/layout/screen split

ndeloof:
**What I did**

Replaced the TTY progress renderer with a three-layer design that makes the recurring display corruptions impossible by construction:

- `tty_model.go` — pure event reducer (injected clock); preserves first-parent-wins updates, monotonic progress and header counters
- `tty_layout.go` — pure `(model, size, now) → lines` function; widths measured in terminal cells (go-runewidth, CJK-correct) on plain text before coloring; every line — status text included — is clipped to the terminal width, so lines can never wrap and desync cursor arithmetic
- `tty_screen.go` — diff-based repaint: unchanged rows are skipped, identical frames write nothing, one `Write` per frame; terminal shrink abandons the block instead of repainting over reflowed rows

The writer coordinates them behind a mutex; the refresh goroutine stops through context cancellation, so `Done()` can no longer deadlock when the operation context was cancelled first (Ctrl-C during pull). The spinner frame derives from the clock instead of advancing per call, and truncation can no longer split multi-byte runes.

Visual output is unchanged — the snapshot test reproduces the previous renderer's golden output character for character. Net −401 lines.

**Related issue**

Structural follow-up to the line-overflow / cursor-desync family (#13595) and the `Done()` deadlock (#13639).


cmd/display/tty_layout.go
```diff
@@ -0,0 +1,354 @@
...
+)
+
+// layoutFrame renders the whole progress block, one string per terminal row.
+func layoutFrame(t *taskTree, operation string, o layoutOpts) []string {
+	done, total := t.counts()
+	header := fmt.Sprintf("[+] %s %d/%d", operation, done, total)
+	lines := []string{renderSegs([]seg{{text: header}}, o.width)}
+
+	rows := buildRows(t, o)
+	maxRows := max(o.height-2, 1)
```
glours, line 72: maxRows floors at 1, but for height==1 header+"more" still yields 2 lines, overflowing the terminal. Fix: clip lines to o.height before returning.

cmd/display/tty.go
```diff
@@ -20,716 +20,204 @@ import (
...
-	case api.Error:
-		return ErrorColor(spinnerError)
-	default:
-		return CountColor(t.spinner.String())
-	}
-}
-
-func colorFn(s api.EventStatus) colorFunc {
+func plainEventColor(s api.EventStatus) colorFunc {
 	switch s {
```
glours: duplicate status→color switch with `cmd/display/tty_layout.go:317`, differ only in default case. Fix: merge into one eventColor(s, def) helper.

cmd/display/tty.go
```diff
@@ -20,716 +20,204 @@ import (
...
-			}
-		}
+func (w *termWriter) handle(e api.Resource) {
+	// Buildkit paints its own UI on the same stream: stay silent while a
+	// build is in flight, and once it's over start a fresh block below
+	// whatever buildkit wrote instead of repainting over it.
+	if e.Text == api.StatusBuilding {
+		w.suspended = true
+	} else if w.suspended {
+		w.suspended = false
```
glours: cmd/display/tty.go:108-109 - same 2-line "clear suspension" sequence duplicated. Fix: extract a small unsuspend() method.

cmd/display/tty.go
```diff
@@ -20,716 +20,204 @@ import (
...
-	text      string
-	details   string
-	status    api.EventStatus
-	current   int64
-	percent   int
-	total     int64
-	spinner   *Spinner
+	// injected for tests
+	size func() (width, height int)
+	now  func() time.Time
```
glours: Use `clockwork.Clock` here instead of a raw `now func() time.Time`,  it's already a direct dependency and used the same way in `pkg/watch/debounce.go`. This would let `tty_test.go:157` advance a fake clock instead of `time.Sleep(20ms)`, removing a timing-dependent test.

cmd/display/tty_model.go
```diff
@@ -0,0 +1,132 @@
...
+	n.total = max(n.total, e.Total)
+	n.current = max(n.current, e.Current)
+	n.percent = max(n.percent, e.Percent)
+	if n.completed() && !wasCompleted {
+		n.endedAt = now
+	}
+}
+
+// roots returns the top-level tasks in arrival order.
+func (t *taskTree) roots() []*node {
```
glours: **`cmd/display/tty_model.go:101`, `:113`, `:125`** — `roots()`/`children()`/`counts()` each rescan every task, called on every 100ms tick. 
Maintain these incrementally in `apply()` (line 65) instead, e.g. a `map[parent][]*node` and running done/total counters, so a frame with no tree change costs O(1) here instead of O(n).

cmd/display/tty_layout.go
```diff
@@ -0,0 +1,354 @@
...
+	switch n.status {
+	case api.Done:
+		return seg{text: spinnerDone, color: SuccessColor}
+	case api.Warning:
+		return seg{text: spinnerWarning, color: WarningColor}
+	case api.Error:
+		return seg{text: spinnerError, color: ErrorColor}
+	default:
+		frame := int(now.Sub(n.startedAt)/tickInterval) % len(termSpinnerFrames)
+		if frame < 0 {
```
glours: `if frame < 0` patches a symptom of a clock that isn't guaranteed monotonic relative to `n.startedAt`. Enforce the invariant at the source instead — assert/guarantee in `apply()` that `now >= n.startedAt`, then drop the clamp here.

cmd/display/tty.go
```diff
@@ -20,716 +20,204 @@ import (
...
-	// matching Done, e.g. nested Start/Done misuse), close it out first so
-	// its render goroutine exits instead of leaking until ctx is done.
-	if w.done != nil {
-		w.done.close()
+func (w *termWriter) Start(ctx context.Context, operation string) {
+	w.mu.Lock()
+	defer w.mu.Unlock()
+	// A Start before the matching Done (nested brackets are a misuse, but
+	// Full is public API): retire the previous cycle so its refresh
+	// goroutine doesn't outlive it.
```
glours: Update this comment: `stopTicks()` only guarantees the old goroutine *will* exit (via context cancellation), not that it already has by the time `Start` returns. Worth spelling out so a future reader doesn't assume synchronous cleanup.

glours (commented):
Some potential fixes to apply before we merge it

cmd/display/tty_layout.go
ndeloof: Done in 6ba9fadbc — layoutFrame now clips its output to o.height as a final guarantee, with a regression test sweeping heights 1..4.

cmd/display/tty.go
ndeloof: Done in 6ba9fadbc — merged into eventColor(status, def); the two call sites pass their own default (nocolor for rows, SuccessColor for plain lines).

cmd/display/tty.go
ndeloof: Done in 6ba9fadbc — extracted unsuspend(), both call sites use it.

cmd/display/tty.go
ndeloof: Done in 6ba9fadbc — termWriter now carries a clockwork.Clock; tests use NewFakeClockAt and Advance. The cancel test's sleep is simply gone: Done waits for the refresh goroutine to exit, so no settling delay is needed.

cmd/display/tty_model.go
ndeloof: Done in 6ba9fadbc — apply() now maintains roots, children (map[parent][]*node) and a running done counter incrementally; counts/roots/children reads are O(1) per tick. Added a bookkeeping regression test (root demotion, completed counter both ways).

cmd/display/tty_layout.go
ndeloof: Done in 6ba9fadbc — clamp dropped; the single-clock invariant (time never regresses between apply and layout) is documented on taskTree, and the clockwork migration makes it structural.

cmd/display/tty.go
ndeloof: Done in 6ba9fadbc — comment now states cancellation only guarantees the old goroutine will exit, and why its possible last repaint is harmless.

cmd/display/tty.go
```diff
@@ -20,716 +20,199 @@ import (
...
-	return func(yield func(*task) bool) {
-		for _, id := range w.ids { // iterate on ids to enforce a consistent order
-			t := w.tasks[id]
-			if t.parents.Has(parent) {
-				yield(t)
-			}
-		}
+	if w.operation == "" {
+		// outside any operation: degrade to one plain line per event
+		_, _ = fmt.Fprintf(w.out, "%s %s %s\n", e.ID, eventColor(e.Status, SuccessColor)(e.Text), e.Details)
```
glours, line 195:
```suggestion
		// outside any operation: degrade to one plain line per event. This                                                                                                                                                                     
        // prints outside the tracked block, so the next repaint must not                                                                                                                                                                       
        // assume the block is still directly above the cursor (same reason                                                                                                                                                                     
        // unsuspend resets after buildkit output).                                                                                                                                                                                             
         _, _ = fmt.Fprintf(w.out, "%s %s %s\n", e.ID, eventColor(e.Status, SuccessColor)(e.Text), e.Details)                                                                                                                                    
         w.scr.reset()  
```

glours (commented):
Still one potential issue to fix before merging

cmd/display/tty.go
ndeloof: Applied in ddbba1c45 (rebased on main along the way — clockwork had been removed there in favor of `testing/synctest`, so the branch also drops its clockwork usage for the plain injected `now func() time.Time`).

glours (approved):

PR #14054, opened 2026-08-17
test(e2e): migration to the Scenario DSL

ndeloof:
Continues the migration started in #14047 (stacked on it): the bulk of the e2e suite moves to the declarative Scenario DSL — pause, no-deps, start/stop, orphans, expose, container_name, ipc, wait, cascade, profiles, healthcheck, recreate, scale, secrets, env_file, configs, commit, export, logs, pull, networks, exec, hooks, cp, environment, config, publish checks, run, up, volumes, build-once, providers, the 'Built' build family, image identity/images and model. Compose models are inlined (txtar for multi-file projects) and ~45 fixture directories/files are removed.

The check vocabulary grows accordingly (all in checks.go): `ExitCode`, `ServiceNotCreated`, one-off awareness (`OneOffState`, `OneOffsUntouched`, `OneOffsRemoved` — service checks now ignore run containers), `Recreated`, `ServiceHealthy`, `ServiceScale`, `ReplicaNumbers`, `ImageExists`, `FileExists`, `FileContains`, `StdoutContains`, `StderrContains`, `OutputMatches`, `OutputMatchesCount`. Action modifiers: `Within` (timeout), `WithStdin` (prompt answers); `Scenario.Dir()` exposes the project directory for host-file interactions.

Migration surfaced a few things reviewers should know:
- `TestPauseServiceDoesNotExist` now locks the current behavior (unknown service rejected), resolving the legacy TODO.
- `TestRemoveOrphans` states its real intent: the removal came from the fixture's `.env` declaring `COMPOSE_REMOVE_ORPHANS=true`, invisible in the legacy test.
- `TestScaleDownNoRecreate`'s final assertion now actually runs (the legacy test built the command without executing it).
- `TestNetworkConfigChanged` locks that a subnet change replaces the network but reconnects the same container.
- commit/export now verify the produced image/archive exists instead of only checking exit 0; several fixed image/container/volume names become project-scoped, removing parallel-run collisions.

Kept legacy by design: tests driving long-running/background processes (watch, logs -f, cancel/SIGINT, attach-restart, wait infinity), format-verification tests (ps, config --hash cross-check), invocation-without`-f` tests (COMPOSE_FILE-from-.env family), and infra-heavy tests (registry publish, buildx builders, dind TLS, sentences fixture, bridge tree diffs).


glours:
Why not using [`embed` directives](https://pkg.go.dev/embed) and keep Compose and Dockerfile fixtures in their own files?

ndeloof:
> Why not using [embed directives](https://pkg.go.dev/embed) and keep Compose and Dockerfile fixtures in their own files?

The intent is to get the whole test scenario in a single file, so a human get all the input visible at once.
Also, using fixtures for our e2e tests, we used to share them between tests making them harder to express the intended scenario

ndeloof:
Superseded by #14087, which includes this migration converted to the testdata/<TestName>/ layout.

PR #14055, opened 2026-08-17
lint: replace gocyclo with gocognit

ndeloof:
**What I did**

Replaced the `gocyclo` linter with `gocognit` (cognitive complexity, threshold 30 — the linter default). Cognitive complexity barely charges guard clauses and early returns, and penalizes nesting instead, which matches how this codebase is written: 8 of the 21 `//nolint:gocyclo` suppressions become unnecessary, while deeply nested functions gocyclo never flagged are now caught. The 19 functions still above the threshold keep a suppression, each marked `FIXME` so they can be restructured over time to complete the migration.

includes refactoring of doBuildBake (hot spot after migration to gocognit, never identified by gocyclo)

**Related issue**

n/a

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #14057, opened 2026-08-17
lint: replace gocyclo with gocognit

ndeloof:
**What I did**

Replaced the `gocyclo` linter with `gocognit` (cognitive complexity, threshold 30 — the linter default). Cognitive complexity barely charges guard clauses and early returns, and penalizes nesting instead, which matches how this codebase is written: 8 of the 21 `//nolint:gocyclo` suppressions become unnecessary, while deeply nested functions gocyclo never flagged are now caught. Remaining functions above the threshold keep a suppression, each marked `FIXME` so they can be restructured over time; `doBuildBake` (the worst offender, cognitive complexity 88) is restructured in this PR as a demonstrator, and `gocyclo` is explicitly disabled so it doesn't sneak back in.

includes refactoring of doBuildBake (hot spot after migration to gocognit, never identified by gocyclo)

Supersedes #14055 (recreated from an in-repo branch to allow a stacked follow-up PR).

**Related issue**

n/a

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -3,6 +3,13 @@ run:
   concurrency: 2
 linters:
   default: none
+  disable:
+    # gocognit deliberately replaced gocyclo — do not re-enable. Cyclomatic
+    # complexity charges a flat guard clause the same as a nested branch, so
+    # our fail-fast style accumulated //nolint suppressions on functions that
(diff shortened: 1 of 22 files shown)


glours (approved):
Cool thanks!

PR #14073, opened 2026-08-17
e2e: fix test resource isolation and make standalone parallelism configurable

glours:
**What I did**

Several e2e tests used fixed host ports (8070, 8080, 8090, etc.), global buildx
builders registered with `--use`, and a hardcoded IPC container name. Under
parallel execution these daemon-level resources caused port conflicts,
buildkitd container-name collisions, and intermittent failures. Separately,
`make e2e-compose-standalone` hardcoded `-parallel=1`, forcing the entire
standalone suite to run serially even though the same tests run in parallel in
plugin mode.

Changes:

- Convert fixed host ports in four fixtures (`network-test`, `sentences`,
  `build-test`, `volume-test`) to ephemeral bindings. Add a `ServicePublishedPort`
  helper to `framework.go` to resolve the actual mapped port at test runtime.
- Give each buildx-builder test a unique daemon-scoped name via a `BuilderName`
  helper, preventing container-name collisions between parallel goroutines.
  Covers `TestBuildPlatformsWithCorrectBuildxConfig`, `TestBuildPrivileged`,
  `TestBuildBuilder`, `TestBuildEntitlements`, `TestBuildTLS`.
- Replace `-parallel=1` in `make e2e-compose-standalone` with
  `E2E_STANDALONE_PARALLEL?=4`, aligning standalone with plugin mode on
  4-core GitHub Actions runners. `E2E_PARALLEL_PLUGIN?=4` makes the
  plugin-mode knob explicit too. Both are overridable (`E2E_STANDALONE_PARALLEL=1`
  restores serial local debugging).
- Raise the poll delay in `watch_test.go` long-running polls (90 s / 120 s
  timeout) from the default 100 ms to 1 s; bump `cascade_test.go` and
  `publish_test.go` from 100 ms to 500 ms. Timeouts are unchanged.

The `logging-driver` fixture keeps its fixed port 24224: the Docker fluentd
log driver encodes `fluentd-address` at service-start time, so ephemeral
resolution would require a multi-step `compose up`. `TestLoggingDriver` is
serial, so there is no collision risk in practice.

**Related issue**

No related issue — this is a test-infrastructure improvement.

*A panda who has learned not to share its bamboo with parallel neighbours* 🐼

panda

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,6 +32,8 @@ endif
 BUILD_FLAGS?=
 TEST_FLAGS?=
 E2E_TEST?=
+E2E_PARALLEL_PLUGIN?=4
+E2E_STANDALONE_PARALLEL?=4
 ifneq ($(E2E_TEST),)
 	TEST_FLAGS:=$(TEST_FLAGS) -run '$(E2E_TEST)'
 endif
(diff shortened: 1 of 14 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

This PR correctly implements two waves of e2e CI optimisations. The diff was reviewed end-to-end across all 17 changed files.

**Key changes verified:**

- **`ServicePublishedPort`** — Correctly omits `-f` in its `compose port` call. `docker compose port` resolves the project purely from the daemon by the `com.docker.compose.project` label when `-p` is provided (`toProjectName` returns early, `backend.Port()` never reads a compose file). The `net.SplitHostPort` parse handles both IPv4 and IPv6 address formats produced by Compose. `t.Fatalf` calls `runtime.Goexit()` so the zero-value return after it is unreachable — no silent zero-port risk.

- **`BuilderName`** — Sanitisation and 60-char truncation are correct. All five call sites pass a top-level `t` (`t.Name()` = function name, e.g. `TestBuildPlatformsWithCorrectBuildxConfig` = 40 chars): the longest composed name is 56 chars, well below the truncation threshold. The resulting names are globally unique across parallel tests, compatible with Docker container-name and context-name character rules.

- **IPC fixture** — `ipc: "container:ipc_e2e-src"` matches exactly what the test creates (`projectName+"-src"` = `"ipc_e2e-src"`). The `--rm` flag on the source container and the final `rm -f` in cleanup coexist safely because the latter's result is discarded.

- **Makefile** — `-parallel=$(E2E_PARALLEL_PLUGIN)` and `-parallel=$(E2E_STANDALONE_PARALLEL)` are correctly positioned before `./pkg/e2e` (test flags before package). `--tags=standalone` is also before the package path. `TEST_FLAGS` is only appended with `-run`/`-skip`, so no double `-parallel` issue in practice.

- **Ephemeral ports** — All removed fixed-port bindings (8070, 8090, 9090, 90, 95, 8080 in fixtures) have corresponding call-site updates to `ServicePublishedPort` or relaxed assertions (`0.0.0.0:` vs `0.0.0.0:8080`). No residual hardcoded port numbers remain for the changed fixtures.

- **Poll-delay calibrations** — Raising delays from 100 ms → 500 ms–1 s reduces `compose exec` storm during long-running watch polls without changing test semantics.

docker-agent (commented):
🟢 **No issues found** — LGTM! [View logs](https://github.com/docker/compose/actions/runs/32035542877).

ndeloof (approved):

PR #14076, opened 2026-08-17
fix(events): validate service names before subscribing to events

glours:
**What I did**
`docker compose events <name>` was passing service arguments straight to the event filter without checking they exist in the project, so a typo would produce an event stream matching nothing and block forever with no output and no error.

Switch to `projectOrName` (which validates args via ToProject) before calling `backend.Events`, consistent with logs, kill, and restart.

**Related issue**
JIRA DDB-637

image

### DIFF
--- a/pkg/e2e/events_test.go
+++ b/pkg/e2e/events_test.go
@@ -1,0 +1,33 @@
+/*
+   Copyright 2020 Docker Compose CLI authors
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package e2e
+
+import (
+	"testing"
+	"time"
+
+	"gotest.tools/v3/icmd"
+)
+
+func TestEventsUnknownService(t *testing.T) {
+	c := NewParallelCLI(t)
+
+	cmd := c.NewDockerComposeCmd(t, "-f", "./fixtures/simple-composefile/compose.yaml", "events", "not-a-service")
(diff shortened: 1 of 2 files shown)


pkg/e2e/events_test.go
```diff
@@ -0,0 +1,30 @@
...
+import (
+	"testing"
+
+	"gotest.tools/v3/icmd"
+)
+
+func TestEventsUnknownService(t *testing.T) {
+	c := NewParallelCLI(t)
+
+	res := c.RunDockerComposeCmdNoCheck(t, "-f", "./fixtures/simple-composefile/compose.yaml", "events", "not-a-service")
```
docker-agent: **[medium] E2E test may block indefinitely if `events` does not exit quickly on unknown service**

`RunDockerComposeCmdNoCheck` calls `icmd.RunCmd` from `gotest.tools/v3/icmd`, which blocks until the subprocess exits — there is no built-in timeout. When the fix is working correctly the `events` command exits immediately with code 1, so the test passes quickly. But if the fix ever regresses (or is not triggered for some reason), the command streams events indefinitely and the test hangs forever instead of failing with a clear timeout error. A hanging test is worse than a clean test failure: it can exhaust the CI parallelism pool and leave the worker stuck.

Consider setting an explicit timeout on the command, for example:

```go
cmd := c.NewDockerComposeCmd(t, "-f", "./fixtures/simple-composefile/compose.yaml", "events", "not-a-service")
cmd.Timeout = 10 * time.Second
res := icmd.RunCmd(cmd)
```

Or use `RunDockerComposeCmdNoCheck` with a deadline-aware approach consistent with other E2E tests.

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 67/100 |

docker-agent (commented):

One medium-severity finding was identified in the new E2E test.

#### Dismissed finding (pre-existing, not introduced by this PR)
- `cmd/compose/events.go:60` — Validation bypass when `--project-name` is set without `-f`: both the old `toProjectName` and new `projectOrName` skip `ToProject` in this case, so this is pre-existing behaviour unchanged by the PR.

maxproske:
Verified it validates the service now :) 

```sh
$ ./bin/build/docker-compose events typo
# no such service: typo
```

ndeloof (approved):

Issue #14074, opened 2026-08-17
Epic: make the codebase agent-legible — fix misleading self-description, ambiguous contracts, and legacy leftovers


ndeloof:
*"This issue was filed by an AI agent on a human's behalf. The human submitter may not have independently verified the report."*

## Goal

Coding agents (and new human contributors) navigate this codebase by reading it locally: file names, doc comments, option structs, docs, and error messages are their map. Today several of those signposts are **inaccurate, ambiguous, or point at code that no longer exists**, so a reader who trusts them lands on wrong conclusions and produces plausible-looking regressions.

This epic tracks making the codebase *self-describing and truthful*: fix documentation and comments that lie, make implicit invariants explicit, remove legacy leftovers, and add guardrails so the map stays accurate.

The findings below come from a systematic legibility review (5 parallel deep dives: lifecycle backend, CLI layer, `pkg/api` contract, build/watch, cross-cutting), with each claim spot-verified against the current code. Every item lists file references. Items marked 🐛 are arguably user-facing bugs discovered along the way and may deserve extraction into standalone issues.

## A. The code misdescribes its own structure — split into focused PRs (supersede #14075)

- [x] ~~`pkg/compose/convergence.go` is a false friend: the `convergence` type was removed (fbea647b9), but the file keeps its name and its doc comment ("convergence manages service's container lifecycle") now sits on top of the unrelated `getScale()` (convergence.go:47-52). The file is a grab-bag (naming, dependency waiting, container creation, start helpers). Rename/split it.~~ → PR #14129 (merged)
- [x] ~~8 comments reference deleted code or stale line numbers: `reconcile.go:195`, `:245`, `:349`, `:378`, `:719` ("matching the previous `ensureNetwork`/`ensureVolume`/convergence.go behavior"), `reconcile.go:1012` ("same way as convergence.go:138-160" — those lines now hold unrelated code), `observed_state.go:154-155`, `:369`. Replace historical references with actual behavioral statements.~~ → PR #14130 (merged)
- [x] ~~`docs/sdk.md` documents a `progress` package that does not exist (`progress.NewTTYWriter/NewPlainWriter/NewJSONWriter/NewQuietWriter`, sdk.md:148-155; `WithEventProcessor(progress.EventProcessor)` at :110 — actual type is `api.EventProcessor`). The real renderers live in `cmd/display`, which an SDK consumer cannot reasonably import.~~ → PR #14131 (merged)
[...]

PR #14077, opened 2026-08-17
api: ScaleOptions carries the replica counts

ndeloof:
`Scale`'s interface doc promises it 'manages numbers of container instances running per service', but `ScaleOptions` had no replica count: the CLI mutated the project model (`SetScale`) before calling the backend, and an SDK consumer following the doc got a no-op convergence.

`ScaleOptions` gains `Replicas map[string]int`, applied to the model by the backend itself; `Services` is documented for what it really does (recreation-policy selection, defaulting to the `Replicas` keys). The CLI is reduced to parsing `SERVICE=REPLICAS` tuples — user-facing formatting stays in `cmd/`, the logic moves to the backend. Behavior is unchanged: a pre-mutated project with an empty `Replicas` map works as before.

Covered by a unit test on the replica application and the existing e2e scale suite (run locally, green).

Part of #14074 (section B). One PR per finding; more to follow.

### DIFF
--- a/cmd/compose/scale.go
+++ b/cmd/compose/scale.go
@@ -92,13 +92,7 @@ func runScale(ctx context.Context, dockerCli command.Cli, backendOptions *Backen
 		}
 	}
 
-	for key, value := range serviceReplicaTuples {
-		if err := setServiceScale(project, key, value); err != nil {
-			return err
-		}
-	}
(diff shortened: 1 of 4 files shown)


glours (approved):
LGTM

PR #14078, opened 2026-08-17
api: option structs document what the implementation actually honors

ndeloof:
Doc-only. Three commits, one per finding of #14074 section B:

1. **The scoping invariant** (`Services` × 3 semantics): the interface doc now states that methods act on the whole `*types.Project` they receive and that scoping is done by narrowing the project — the `Services` fields of downstream option structs are intent markers, not filters. Each of the three fields (`ProjectLoadOptions`, `CreateOptions`, `StartOptions`) documents its actual effect.
2. **`StartOptions`**: the five fields only Up's foreground session reads (`Attach`, `OnExit`, `ExitCodeFrom`, `Watch`, `NavigationMenu`) are marked as ignored by `Start`; two implicit behaviors become explicit — `Attach == nil` is the detached/interactive mode switch, and `AttachTo` has two unrelated meanings.
3. **`RunOptions`**: the embedded `CreateOptions` only propagates 4 fields to dependency creation, and the same struct doubles as Exec's options, which honors only the exec-relevant subset.

No code change — comments only. Structural reshaping of these structs is deliberately out of scope.

Part of #14074 (section B).

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -52,7 +52,10 @@ type ProjectLoadOptions struct {
 	EnvFiles []string
 	// Profiles to activate
 	Profiles []string
-	// Services to select (empty = all)
+	// Services narrows the loaded project to the named services and their
+	// dependencies, and enables the profiles they require (empty = all).
+	// This is where an operation's scope is decided: the Services fields of
+	// downstream option structs do not filter.
 	Services []string
 	// Offline mode disables remote resource loading
 	Offline bool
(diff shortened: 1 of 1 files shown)


glours (changes requested):
Per-field docs all check out, I verified them against the code. 👍

Just the interface-level comment:

> The Services fields found on some option structs are not filters

That's not true for most of the API. Example: `StopOptions.Services` filters for
real at stop.go:51-53, so `compose stop web` doesn't stop `db`. Same story for
ps, logs, kill, images, build, watch, down, restart and others — 15 of the 18
`Services` fields in api.go. Same claim on `ProjectLoadOptions.Services`
("downstream option structs do not filter").

The per-struct docs are exactly what the epic asked for, can you just drop the
interface-level generalisation?

glours (approved):

PR #14080, opened 2026-08-17
ci: pre-pull and cache heavy e2e images via GitHub Actions cache

glours:
**What I did**

Added a GHA cache layer for the passive Docker images used by the e2e suite.

A new allowlist `.github/e2e-images.txt` lists the infrastructure images that every test cell downloads: `docker:dind`, `mariadb`, `golang:alpine`, `fluent/fluent-bit:3.1.7-debug`, `registry:3`, and the `gtardif/sentences-*` / `nginx:alpine` helper images.

Before each test run the workflow restores a pre-built tar from the GHA cache (key = Docker engine version + hash of the image list). On a cache hit the images are loaded instantly with `docker load`; on a miss they are pulled in parallel (`xargs -P 4 -n 1 docker pull`) and saved for the next run. Any change to `.github/e2e-images.txt` automatically invalidates the cache.

Images used by pull-policy and image-identity assertions (`alpine:*` and several fixture-specific tags) are intentionally absent from the list so those tests remain unaffected. No change to local `make e2e-compose` or `make e2e-compose-standalone` runs.

> **Note:** this PR is stacked on #14073. Until that PR merges into `main` the diff includes both sets of changes.
>
> The `ci.yml` changes cannot be pushed directly from the sandbox environment (requires `workflow` OAuth scope). They are provided as `.github/ci-workflow-cache-only.patch`. To apply:
> ```bash
> git apply .github/ci-workflow-cache-only.patch
> git rm .github/ci-workflow-cache-only.patch
> git commit --amend --no-edit --signoff
> git push --force-with-lease
> ```

**Related issue**

No related issue — this is a CI infrastructure improvement.

*A panda that found its bamboo already pre-stocked instead of having to go fetch it* 🐼

Giant Panda


docker-agent (commented):
🟢 **No issues found** — LGTM! [View logs](https://github.com/docker/compose/actions/runs/32040182310).

pkg/e2e/framework.go
```diff
@@ -517,3 +519,42 @@ func (c *CLI) cleanupWithDown(t testing.TB, project string, args ...string) {
...
 }
+
+// ServicePublishedPort returns the ephemeral host port mapped to targetPort
+// on the named service in the given compose project. It requires the service
+// to be already running. The test fails immediately if the mapping cannot be
+// resolved.
+func (c *CLI) ServicePublishedPort(t testing.TB, project, service string, targetPort int) int {
+	t.Helper()
+	res := c.RunDockerComposeCmd(t, "-p", project, "port", service, strconv.Itoa(targetPort))
+	addr := strings.TrimSpace(res.Stdout())
```
docker-agent: **[medium] `ServicePublishedPort` breaks for scaled services — multi-line `compose port` output causes `SplitHostPort` to fail**

`docker compose port` outputs one `host:port` line per replica when a service is scaled. `strings.TrimSpace` only strips outer whitespace; embedded newlines remain. `net.SplitHostPort` then sees a string like `"0.0.0.0:32768\n0.0.0.0:32769"` and returns an error, causing `t.Fatalf` to fire immediately.

Taking only the first line is the idiomatic fix and is consistent with how other helpers in this file handle `compose` output:

```suggestion
	addr := strings.SplitN(res.Stdout(), "\n", 2)[0]
```

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 62/100 |

docker-agent (commented):

Two findings in the newly added code — one in `pkg/e2e/framework.go` and one in the `Makefile` — affect correctness of local developer runs and the new test helper.

Makefile
```diff
@@ -32,6 +32,8 @@ endif
 BUILD_FLAGS?=
 TEST_FLAGS?=
 E2E_TEST?=
+E2E_PARALLEL_PLUGIN?=4
+E2E_STANDALONE_PARALLEL?=4
```
docker-agent, line 36: **[medium] `E2E_STANDALONE_PARALLEL` defaults to 4, silently removing the serial guarantee for local `make e2e-compose-standalone` runs**

The previous command hardcoded `-parallel=1`; standalone tests have always been run serially. This PR changes the default to 4, meaning any developer who runs `make e2e-compose-standalone` without explicitly setting `E2E_STANDALONE_PARALLEL=1` will now get 4 parallel workers.

The PR description says "make e2e-compose-standalone unchanged for local runs", but this is a behaviour change. Standalone tests that rely on unique ports, container names, or Docker-daemon-global state (e.g., build caches, named builders) can fail intermittently when run 4-way in parallel. The recent history of this file also shows the standalone runner was previously kept serial for stability.

If the intent is to let CI run at higher parallelism while keeping local runs safe, the default should stay at `1` and only CI should override via the environment:

```suggestion
E2E_STANDALONE_PARALLEL?=1
```

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 67/100 |

glours:
**Re: `ServicePublishedPort` / [discussion_r3802243422](https://github.com/docker/compose/pull/14080#discussion_r3802243422)**

Fixed in 9045bd46 — switched to `strings.SplitN(res.Stdout(), "\n", 2)[0]` so only the first line is consumed, making the helper robust when a service is scaled.

---

**Re: `E2E_STANDALONE_PARALLEL?=4` / [discussion_r3802243440](https://github.com/docker/compose/pull/14080#discussion_r3802243440)**

The default of 4 was intentional: the wave-1 isolation work (ephemeral ports, per-test builder names, unique container names) removed the resource-sharing that previously forced standalone to run serially. CI does not explicitly override this value, so reverting to 1 would regress CI performance back to fully serial. Developers who need deterministic serial runs locally can still pass `E2E_STANDALONE_PARALLEL=1 make e2e-compose-standalone` explicitly. Happy to accept the suggestion to revert to 1 and instead add an explicit `E2E_STANDALONE_PARALLEL=4` in `ci.yml` if that is preferred, but that would require a separate commit to the workflow file.

glours:
The measurements didn’t show any meaningful benefit from introducing the cache, so I’m closing the PR.

Issue #14081, opened 2026-08-17
Epic: converge the start phase into the plan engine


ndeloof:
*"This issue was filed by an AI agent on a human's behalf. The human submitter may not have independently verified the report."*

## Context

Two lifecycle engines coexist today (#14074, section C). The plan-based reconciler (`reconcile.go` → `executor.go`, single entry point `create.go`) is pure, deterministic and golden-tested — but **create-only**: containers leave the plan in `created` state, and `OpStartContainer` is only ever emitted for exotic states (paused/dead). Everything that makes an application actually start lives in the second, imperative engine (`InDependencyOrder`): `waitDependencies` (health/completion polling), secret/config injection, `pre_start`/`post_start` hooks. `up` chains the two with **two different daemon snapshots**, no canonical `project` object across the phases, two event-emission systems, and latent bugs at the seam (`startMx` not held on the path actually exercised; the start-phase `ContainerList` skips the `config-hash` filter; dependency-wait timeouts swallowed by `ctx.Done() → nil`).

This epic tracks converging the **start phase** into the plan engine, one reviewable PR at a time.

## Target architecture

Guiding split: **decision vs execution** — not "plan vs imperative". The `Plan` is the pivot execution format of Compose; the reconciler is just one producer of plans.

- **One snapshot, one project, one plan with two phases.** `ReconcileOptions` gains a scope (Create, Start, or both); `PlanNode` gains a `Phase` field. `up -d` builds a single Create+Start plan; `start`/`scale`/watch-rebuild use scope Start (never recreating); `compose create` keeps scope Create unchanged — **the existing golden tests do not change**.
- **New operations** (explicit numbering, 40+):
  - `OpWaitCondition` — one node per (awaited service, condition ∈ healthy / completed_successfully / running_or_healthy), **deduplicated** across dependents (a `waitNodes` map, like `networkNodes`). `required: false` is absorbed locally: Skipped event, node succeeds — same pattern as the existing `BestEffort`. `condition: service_started` needs **no node**: a plain DAG edge expresses it. Health is **re-observed at execution time** (the node runs today's `waitDependency` polling); `ObservedState` deliberately does not grow a `Health` field — it would be stale by construction.
  - `OpRunPreStart` — per service. Emitted at plan time only when no replica was running at observation (today's rule in `startService`), targeting the lowest-numbered replica (`lowestNumberedContainer`).
  - `OpRunPostStart` — per container, after its start.
[...]

PR #14082, opened 2026-08-17
plan/start groundwork: defined wait-timeout semantics, soft-failing operations

ndeloof:
Phase 0 of #14081 (converge the start phase into the plan engine) — the two prerequisites, each useful on its own:

1. **`waitDependencies` no longer swallows its timeout** (🐛 from #14074 C): every polling goroutine returned `nil` on `ctx.Done()`, making the 'timeout waiting for dependencies' path unreachable except by race — `up --wait --wait-timeout` could succeed silently with nothing healthy. A deadline now surfaces as the failure it is; a plain user cancellation still returns nil so Ctrl-C doesn't masquerade as a dependency failure. Both paths gain their first unit tests.
2. **Walker-level `Operation.Optional`**: a failing Optional node emits a Skipped event + warning and lets dependents run, instead of aborting the plan — the capability `required: false` dependency waits will need once they become plan nodes. Documented against the narrower pre-existing `BestEffort` flag so the two can't be confused. Unit-tested.

Part of #14081; also closes one 🐛 item of #14074 (section C).

### DIFF
--- a/pkg/compose/convergence.go
+++ b/pkg/compose/convergence.go
@@ -187,6 +187,14 @@ func (s *composeService) waitDependencies(ctx context.Context, project *types.Pr
 				select {
 				case <-ticker.C:
 				case <-ctx.Done():
+					// A deadline IS the failure this function must detect:
+					// surface it so --wait/--wait-timeout callers report it
+					// instead of silently succeeding. A plain cancellation
+					// (user interruption) is not a wait failure: return nil
+					// and let the caller's own context handling decide.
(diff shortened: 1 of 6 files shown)


ndeloof:
Superseded by the redesigned plan in #14081: the swallowed-timeout fix and the soft-fail (`Optional`) idea are carried over as dedicated PRs (lot 0/lot 1) rebuilt on current main, where `waitDependencies` has since been split into per-condition functions. Closing in favor of that series.

PR #14084, opened 2026-08-18
fix(watch): sync into a symlinked directory instead of failing

### LINKED ISSUE #13795: [BUG] watch says: Error response from daemon: cannot overwrite non-directory "xxx" with directory "yyy"
### Description

When I do a `docker compose up --watch` I get the error:

> copying files to dff6....: Error response from daemon: cannot overwrite non-directory "/some/path/in/image" with directory "/"

In this case `/some/path/in/image` is a relative symlink.

The error occurs only with `---watch` which probably means that this is an issue with the initial sync of compose watch.

### Steps To Reproduce

Unfortunately I have not managed to reproduce the issue, but I have seen it multiple times.

- It always occurs after switching to a Git commit where folder structure is entirely different.
- The first time I was able to fix it by running `docker system prune`. The second time by `docker compose build --no-cache`.
- The error occurs only with `--watch`. I can start containers successfully without watch.

### Compose Version

```Text
$ docker compose version
Docker Compose version v5.1.3
```

### Docker Environment

```Text
$ docker info
 Version:    29.5.0
 Context:    default
 Debug Mode: false
 Plugins:
    Version:  v0.34.0
    Version:  v5.1.3

Server:
 Containers: 7
  Running: 3
  Paused: 0
  Stopped: 4
 Images: 10
 Server Version: 29.5.0
 Storage Driver: overlayfs
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
 CDI spec directories:
  /etc/cdi
  /var/run/cdi
 Swarm: inactive
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 77c84241c7cbdd9b4eca2591793e3d4f4317c590
 runc version: v1.3.5-0-g488fc13e
 init version: de40ad0
 Security Options:
  apparmor
  seccomp
   Profile: builtin
  cgroupns
 Kernel Version: 6.1.0-44-amd64
 Operating System: Neptune 8.2
 OSType: linux
 Architecture: x86_64
 CPUs: 12
 Total Memory: 33.64GiB
 Name: longy0201
 ID: 2b992073-edb2-48f1-b2c1-81e969c69a58
 Debug Mode: false
 Experimental: false
 Insecure Registries:
  ::1/128
  127.0.0.0/8
 Live Restore Enabled: false
 Firewall Backend: iptables
  EnableUserlandProxy: true
```

### Anything else?

_No response_

Endika:
**What I did**

`compose watch` aborted the whole sync batch when a synced directory maps onto a path the
image exposes as a symlink:

```
Error handling changed files: copying files to <id>: Error response from daemon:
cannot overwrite non-directory "/app/data/sub" with directory "/"
```

The archive built for a sync carries a header for every directory it walks, and extraction
runs with `noOverwriteDirNonDir`, so a directory header for a path the container resolves
through a symlink makes the engine reject the entire copy — not a single file of that batch
lands.

A sync sends the archive unchanged, exactly as it does today. Only when the engine rejects the
copy does compose rebuild it without the headers the archive's own entries already imply — the
engine creates those directories while extracting (`createImpliedDirectories`), so the file
entries resolve through the symlink, which is what a `sync` action is expected to do. A
directory nothing else would create — an empty one — keeps its header. The retried archive is a
strict subset of the one that just failed, so replaying it is safe, and it costs an extra round
trip only in the case that was already broken.

Reproducer:

```yaml
services:
  app:
    image: alpine
    command: sh -c 'mkdir -p /app/data /var/sub && ln -s /var/sub /app/data/sub && sleep infinity'
    develop:
      watch:
        - action: sync
          path: ./data
          target: /app/data
```

`docker compose up --watch`, then create `./data/sub/hello.txt` on the host. On main the sync
fails with the error above; with this change the file lands in `/var/sub/hello.txt` and
`/app/data/sub` is still a symlink.

Two things worth flagging:

- Because the retry only fires on a copy that already failed, an ordinary sync keeps sending
  the directory headers and the host's mode and ownership with them. In the fallback the
  engine creates those directories with `ImpliedDirectoryMode` (0755) and its own root
  uid/gid; empty directories still carry their own header either way.
- Creating an *empty* directory right on top of a symlinked path still fails. There the
  header is the only thing that would create the directory, and the client has no way to know
  the container resolves that path through a symlink.

**Tests**

- `internal/sync`, all three reading back the archives handed to `Untar`:
  `TestSync_PopulatedDirectoryKeepsItsHeader` asserts a directory that arrives already
  populated still carries its own header, with the host's mode;
  `TestSync_RetriesWithoutImpliedDirectoriesWhenCopyFails` asserts the retried archive drops
  that header while an empty nested directory keeps its own; and
  `TestSync_ReportsTheCopyErrorWhenTheRetryFailsToo` asserts the engine's original error is
  what surfaces when the fallback fails too.
- `pkg/e2e`: `TestWatchSyncIntoSymlinkedDirectory` moves an already populated directory into
  the watched tree, the way a branch switch does, and checks the file shows up in the symlink
  target with the symlink intact. It fails on main with the exact error from the issue.

**Related issue**

fixes #13795

### DIFF
--- a/internal/sync/tar.go
+++ b/internal/sync/tar.go
@@ -96,7 +96,7 @@ func (t *Tar) Sync(ctx context.Context, service string, paths []*PathMapping) er
 	eg.SetLimit(16) // arbitrary limit, adjust to taste :D
 	for i := range containers {
 		containerID := containers[i].ID
-		tarReader := tarArchive(pathsToCopy)
+		tarReader := tarArchive(pathsToCopy, keepImpliedDirectories)
 
 		eg.Go(func() error {
 			if len(deleteCmd) != 0 {
@@ -108,9 +108,24 @@ func (t *Tar) Sync(ctx context.Context, service string, paths []*PathMapping) er
 			}
 
 			if err := t.client.Untar(ctx, containerID, tarReader); err != nil {
-				errMu.Lock()
-				errs = append(errs, fmt.Errorf("copying files to %s: %w", containerID, err))
-				errMu.Unlock()
+				// The engine answers with a plain 500, so the message is the only discriminator.
+				if !strings.Contains(err.Error(), "cannot overwrite non-directory") {
+					errMu.Lock()
+					errs = append(errs, fmt.Errorf("copying files to %s: %w", containerID, err))
+					errMu.Unlock()
+					return nil
+				}
+
+				// A directory header for a path the container resolves through a symlink makes
+				// the engine reject the whole copy, so retry without the headers the archive's
+				// own entries already imply. That archive is a strict subset of the one that
+				// just failed, so sending it again is safe.
+				retry := tarArchive(pathsToCopy, dropImpliedDirectories)
+				if retryErr := t.client.Untar(ctx, containerID, retry); retryErr != nil {
+					errMu.Lock()
+					errs = append(errs, fmt.Errorf("copying files to %s: %w", containerID, errors.Join(err, retryErr)))
+					errMu.Unlock()
+				}
 			}
 			return nil // don't fail-fast; collect all errors
 		})
@@ -139,7 +154,7 @@ func (a *ArchiveBuilder) Close() error {
 }
 
 // ArchivePathsIfExist creates a tar archive of all local files in `paths`. It quietly skips any paths that don't exist.
-func (a *ArchiveBuilder) ArchivePathsIfExist(paths []PathMapping) error {
+func (a *ArchiveBuilder) ArchivePathsIfExist(paths []PathMapping, dropImpliedDirs bool) error {
 	// In order to handle overlapping syncs, we
 	// 1) collect all the entries,
 	// 2) de-dupe them, with last-one-wins semantics
@@ -161,6 +176,9 @@ func (a *ArchiveBuilder) ArchivePathsIfExist(paths []PathMapping) error {
 	}
 
 	entries = dedupeEntries(entries)
+	if dropImpliedDirs {
+		entries = withoutImpliedDirectories(entries)
+	}
 	for _, entry := range entries {
 		err := a.writeEntry(entry)
 		if err != nil {
@@ -318,11 +336,17 @@ func (a *ArchiveBuilder) entriesForPath(localPath, containerPath string) ([]arch
 	return result, nil
 }
 
-func tarArchive(ops []PathMapping) io.ReadCloser {
+// Whether tarArchive keeps the header of a directory the archive itself fills with entries.
+const (
+	keepImpliedDirectories = false
+	dropImpliedDirectories = true
+)
+
+func tarArchive(ops []PathMapping, dropImpliedDirs bool) io.ReadCloser {
 	pr, pw := io.Pipe()
 	go func() {
 		ab := NewArchiveBuilder(pw)
-		err := ab.ArchivePathsIfExist(ops)
+		err := ab.ArchivePathsIfExist(ops, dropImpliedDirs)
 		if err != nil {
 			_ = pw.CloseWithError(fmt.Errorf("adding files to tar: %w", err))
 		} else {
@@ -336,6 +360,28 @@ func tarArchive(ops []PathMapping) io.ReadCloser {
 		}
 	}()
 	return pr
+}
+
+// withoutImpliedDirectories removes the header of every directory the archive already fills with
+// entries: the engine creates those while extracting, while an explicit header makes it reject the
+// whole copy when the container resolves that path through a symlink. A directory nothing else
+// would create - an empty one - keeps its header.
+//
+// The engine creates them with ImpliedDirectoryMode and its own root uid/gid rather than the host's,
+// so this is a fallback for a copy that already failed, never the first attempt.
+func withoutImpliedDirectories(entries []archiveEntry) []archiveEntry {
+	holdsEntries := make(map[string]bool, len(entries))
+	for _, entry := range entries {
+		holdsEntries[path.Dir(path.Clean(entry.header.Name))] = true
+	}
+	result := make([]archiveEntry, 0, len(entries))
+	for _, entry := range entries {
+		if entry.header.Typeflag == tar.TypeDir && holdsEntries[path.Clean(entry.header.Name)] {
+			continue
+		}
+		result = append(result, entry)
+	}
+	return result
 }
 
 // Dedupe the entries with last-entry-wins semantics.
--- a/internal/sync/tar_test.go
+++ b/internal/sync/tar_test.go
@@ -15,7 +15,9 @@
 package sync
 
 import (
+	"archive/tar"
 	"context"
+	"errors"
 	"io"
 	"os"
 	"path/filepath"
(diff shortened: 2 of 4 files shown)


ndeloof (dismissed):

glours (changes requested):
Reviewing and testing this PR surfaced a regression on a very common case.

  Dropping the directory headers means directories that arrive **already populated** in a single sync batch are no longer created from the header (host uid/gid) but by the daemon via `createImpliedDirectories`, i.e. `root:root 0755`. On a non-root service the container can no longer write into them — and the next sync's `rm -rf` (which runs as the container user) fails too, so the container tree diverges from the host and never reconverges, with only a warning logged. No symlink is needed to trigger this: it's the default watch path.

Minimal reproducer:

  ```yaml
  services:
    app:
      image: alpine
      user: "1000:1000"      # any non-root service
      command: sleep infinity
      develop:
        watch:
          - action: sync
            path: ./data
            target: /app/data
  ```

  ```bash
  mkdir -p data staged && echo hello > staged/hello.txt
  docker compose up --watch          # wait for "Watch enabled"
  cp -a staged data/sub              # directory arrives already populated (the git checkout / mv shape)

  docker compose exec app stat -c '%U:%G %a' /app/data/sub
  docker compose exec app touch /app/data/sub/x
  rm data/sub/hello.txt              # then check inside the container
  ```

  |              | `stat /app/data/sub` | `touch`             | delete on the host                       |
  | ------------ | -------------------- | ------------------- | ---------------------------------------- |
  | before       | `1000:1000 700`      | OK                  | file removed                             |
  | with this PR | `root:root 755`      | `Permission denied` | `rm: Permission denied`, file still there |
Smallest fix direction: keep the current archive for the first attempt and only rebuild with `dropImpliedDirectories` and retry if `Untar` fails, the reduced archive is a strict subset, so the retry is idempotent and only costs an extra round trip in the already-broken case.

Endika:
> Reviewing and testing this PR surfaced a regression on a very common case.
> 
> Dropping the directory headers means directories that arrive **already populated** in a single sync batch are no longer created from the header (host uid/gid) but by the daemon via `createImpliedDirectories`, i.e. `root:root 0755`. On a non-root service the container can no longer write into them — and the next sync's `rm -rf` (which runs as the container user) fails too, so the container tree diverges from the host and never reconverges, with only a warning logged. No symlink is needed to trigger this: it's the default watch path.
> 
> Minimal reproducer:
> 
> ```yaml
> services:
>   app:
>     image: alpine
>     user: "1000:1000"      # any non-root service
>     command: sleep infinity
>     develop:
>       watch:
>         - action: sync
>           path: ./data
>           target: /app/data
> ```
> 
> ```shell
> mkdir -p data staged && echo hello > staged/hello.txt
> docker compose up --watch          # wait for "Watch enabled"
> cp -a staged data/sub              # directory arrives already populated (the git checkout / mv shape)
> 
> docker compose exec app stat -c '%U:%G %a' /app/data/sub
> docker compose exec app touch /app/data/sub/x
> rm data/sub/hello.txt              # then check inside the container
> ```
> 
> `stat /app/data/sub`	`touch`	delete on the host
> before	`1000:1000 700`	OK	file removed
> with this PR	`root:root 755`	`Permission denied`	`rm: Permission denied`, file still there
> Smallest fix direction: keep the current archive for the first attempt and only rebuild with `dropImpliedDirectories` and retry if `Untar` fails, the reduced archive is a strict subset, so the retry is idempotent and only costs an extra round trip in the already-broken case.

Reproduced on both revisions with your compose file, host uid/gid 1000:

| `cp -a staged data/sub` | previous revision | with the retry |
| --- | --- | --- |
| `stat -c '%u:%g %a' /app/data/sub` | `0:0 755` | `1000:1000 755` |
| `touch /app/data/sub/x` | `Permission denied` | OK |
| `rm data/sub/hello.txt` on the host | `rm: can't remove ...: Permission denied`, file still in the container | removed |

Implemented the direction you suggested. The first attempt sends the archive untouched, so an
ordinary sync is byte-for-byte what main sends today and the directory headers keep carrying the
host's mode and ownership. Only a copy the engine rejects is rebuilt without the headers its own
entries already imply and sent again. If that retry fails too, the engine's original error is
what surfaces — the fallback shouldn't rename the failure.

Three unit tests cover it, each reading back the archives handed to `Untar`: a directory that
arrives already populated keeps its own header on the first attempt, with the host's mode; the
retried archive drops it while an empty nested directory keeps its own; and the original error
is what comes out when both attempts fail. All three fail without the change.
`TestWatchSyncIntoSymlinkedDirectory` still passes, so #13795 stays fixed.

glours (changes requested):
The shape is right and the ownership regression is gone.

One change: the retry fires on any `Untar` error, not just the symlink one.

Failing the first `PUT /containers/{id}/archive` with an unrelated 500:
```
Sync err: <nil>; untar calls: 2
HOST:      sub 700 1000:1000
CONTAINER: /app/newdir/sub 755 0:0
```

So a transient blip silently re-owns new directories to root:root 0755. Could you, please gate it,  the daemon returns a plain 500, so the message is the only
discriminator, and it's stable across go-archive v0.1.0..v0.3.3:

```go
if !strings.Contains(err.Error(), "cannot overwrite non-directory") {
      // report and return
}

Also errors.Join(err, retryErr), so a retry failing for another reason isn't
reported as the symlink error.

**Comment by Endika** [comment]
> The shape is right and the ownership regression is gone.
> 
> One change: the retry fires on any `Untar` error, not just the symlink one.
> 
> Failing the first `PUT /containers/{id}/archive` with an unrelated 500:
> 
> ```
> Sync err: <nil>; untar calls: 2
> HOST:      sub 700 1000:1000
> CONTAINER: /app/newdir/sub 755 0:0
> ```
> 
> So a transient blip silently re-owns new directories to root:root 0755. Could you, please gate it, the daemon returns a plain 500, so the message is the only discriminator, and it's stable across go-archive v0.1.0..v0.3.3:
> 
> ```go
> if !strings.Contains(err.Error(), "cannot overwrite non-directory") {
>       // report and return
> }
> 
> Also errors.Join(err, retryErr), so a retry failing for another reason isn't
> reported as the symlink error.
> ```

Both done in f095c22.

The retry is gated on the message now, so only the failure the directory headers
themselves cause gets retried:

```go
// The engine answers with a plain 500, so the message is the only discriminator.
if !strings.Contains(err.Error(), "cannot overwrite non-directory") {
      errMu.Lock()
      errs = append(errs, fmt.Errorf("copying files to %s: %w", containerID, err))
      errMu.Unlock()
      return nil
}
```

and the report is errors.Join(err, retryErr), so a retry that fails for an
unrelated reason no longer surfaces as the symlink error.

TestSync_DoesNotRetryACopyThatFailedForAnotherReason covers the case you hit:
an unrelated error on the first PUT leaves untar calls: 1, the original error
is what comes back, and the archive that was sent still carries the directory
header. TestSync_ReportsTheCopyErrorWhenTheRetryFailsToo now also asserts both
errors are in the reported one.

glours (approved):
Sounds good to me, all tests are green now, thanks for the contribution @Endika

PR #14085, opened 2026-08-18
Clean up legacy/experimental environment-variable leftovers

ndeloof:
Cleans up the legacy/experimental environment-variable leftovers identified in #14074 (section D, item 4), one commit per point:

- **`COMPOSE_EXPERIMENTAL_WATCH_TAR=0` no longer breaks `compose watch`** 🐛 — since the docker-cp syncer removal (d20340299) the tar syncer is the only implementation, so opting out failed hard with "no available sync implementation". The variable is now ignored, with a warning when explicitly set to false.
- **Remove dead `internal/experimental` package** — zero importers; its `COMPOSE_EXPERIMENTAL` global opt-out and Desktop feature-flag loading are not wired anywhere.
- **Drop stale `COMPOSE_BAKE=1` from the build e2e env** — no production code reads it since the internal buildkit builder was dropped.
- **Document `COMPOSE_EXPERIMENTAL_GIT_REMOTE` / `COMPOSE_EXPERIMENTAL_OCI_REMOTE` as opt-outs enabled by default** — the EXPERIMENTAL name suggests an opt-in, but both default to true; the semantics are now stated at the declaration and read sites. No behavior change (renaming the variables would break existing opt-outs).

Part of #14074.


ndeloof:
I wonder we should deprecate `COMPOSE_EXPERIMENTAL_*_REMOTE` variables, wdyt @glours ?

PR #14087, opened 2026-08-18
test(e2e): declarative Scenario layer, project files under testdata/<TestName>/

ndeloof:
Alternative to #14047 + #14054, as a single PR: the declarative Scenario layer plus the full migration of the e2e suite to it, with each scenario's project files under `testdata/<TestName>/` as standalone compose files instead of inline YAML/txtar in the test source.

- Files are in their native, directly-runnable format: `docker compose -f pkg/e2e/testdata/TestRestart/compose.yaml up` works as-is. Multi-file projects (Dockerfile, .env, includes) need no special format.
- The scenario resolves `testdata/<t.Name()>/` by convention (subtests map to nested directories) and copies it to a temp dir, so committed files are never mutated.
- Ownership is strictly one test per directory, enforced by a check that fails on any unowned `testdata` directory — no shared fixtures, no leftovers after a rename, no catch-all. Models previously shared between tests through Go constants are materialized per test.
- The migration commits mirror #14054 one-for-one (~35 test files, `fixtures/` shrinks from 70 to 28 entries; the remainder belongs to tests still in legacy style).

Only one of #14047/#14054 and this PR should be merged.


pkg/e2e/scenario.go
```diff
@@ -0,0 +1,611 @@
...
+		switch {
+		case unicode.IsUpper(r):
+			if i > 0 && !unicode.IsUpper(prev) {
+				b.WriteRune('-')
+			}
+			b.WriteRune(unicode.ToLower(r))
+		case unicode.IsLower(r) || unicode.IsDigit(r):
+			b.WriteRune(r)
+		default:
+			b.WriteRune('-')
```
glours: `projectNameFor` maps both `/` (subtest separator) and CamelCase word boundaries to a single `-`. This makes `TestFoo/bar` and `TestFooBar` resolve to the same project name (`e2e-foo-bar`), so two such tests running in parallel would share a Compose project and interfere.                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                            
Using `--` for the slash and `-` for CamelCase boundaries makes the two kinds of names non-colliding: `TestFoo/bar` → `e2e-foo--bar`, `TestFooBar` → `e2e-foo-bar`.                                                                                       
                                                                                                                                                                       
```suggestion
				if r == '/' {
					b.WriteString("--")
				} else {
					b.WriteRune('-')
				}
```

pkg/e2e/scenario.go
```diff
@@ -0,0 +1,611 @@
...
+		}
+		target := filepath.Join(dst, rel)
+		if d.IsDir() {
+			return os.MkdirAll(target, 0o755)
+		}
+		data, err := os.ReadFile(path)
+		if err != nil {
+			return err
+		}
+		return os.WriteFile(target, data, 0o644)
```
glours: `copyDir` unconditionally writes every file with mode `0o644`, stripping execute bits. Nothing in the current testdata requires an execute bit on the host (scripts are invoked via `ash /script.sh`), so no test breaks today, but any future testdata  script that needs to be directly executable will fail silently with `permission denied` at runtime, not at copy time.                                                                                                                                     
                                                                                                                                                                                                                                                            
Propagating the source file's mode (`info` is already available from the `WalkDir` entry via `os.Lstat(path)`) costs nothing and prevents the trap.

```suggestion
		return os.WriteFile(target, data, info.Mode())
```

pkg/e2e/checks.go
```diff
@@ -0,0 +1,572 @@
...
+	return Check{
+		name: fmt.Sprintf("output matches %q", pattern),
+		fn: func(ctx *CheckContext) error {
+			if !regexp.MustCompile(pattern).MatchString(ctx.result.Stdout()) {
+				return fmt.Errorf("no match in stdout")
+			}
+			return nil
+		},
+	}
+}
```
glours, line 105: `regexp.MustCompile(pattern)` is called inside the `fn` closure, so it runs on every invocation of the check. For a bare `Step` that is a one-shot cost; inside an `Eventually` loop (500 ms tick, 30 s timeout) the pattern is recompiled ~60 times per check. It also means a malformed pattern panics inside the closure rather than at `Check` construction time, making the error harder to attribute.                                                                                                        
                                                                                                                                                                                                                                                            
  Compile once at construction and close over the result:

```suggestion
func OutputMatches(pattern string) Check {
	re := regexp.MustCompile(pattern) // panics here, at construction time                                                                                                                                                                                
	return Check{
		name: fmt.Sprintf("output matches %q", pattern),
		fn: func(ctx *CheckContext) error {
			if !re.MatchString(ctx.result.Stdout()) {
				return fmt.Errorf("no match in stdout")
			}
			return nil
		},
	}
}
```

pkg/e2e/checks.go
```diff
@@ -0,0 +1,572 @@
...
+
+// OutputMatchesCount expects the regular expression to match the command's
+// stdout exactly n times, e.g. counting how many times a service was built.
+func OutputMatchesCount(pattern string, n int) Check {
+	return Check{
+		name: fmt.Sprintf("output matches %q %d time(s)", pattern, n),
+		fn: func(ctx *CheckContext) error {
+			matches := regexp.MustCompile(pattern).FindAllString(ctx.result.Stdout(), -1)
```
glours: Same as above ☝️

glours (changes requested):
Few changes to apply but overall look pretty good 👍

pkg/e2e/scenario.go
ndeloof: Fixed in 4bcb1d0c6: the subtest separator now maps to a double dash, so TestFoo/bar -> e2e-foo--bar and TestFooBar -> e2e-foo-bar no longer collide.

pkg/e2e/scenario.go
ndeloof: Fixed in 4bcb1d0c6: copyDir now propagates the source file mode.

pkg/e2e/checks.go
ndeloof: Fixed in 4bcb1d0c6: pattern compiled once at construction, panic surfaces where the check is declared.

pkg/e2e/checks.go
ndeloof: Fixed in 4bcb1d0c6, same treatment for OutputMatchesCount.

glours (approved):
LGTM

PR #14091, opened 2026-08-19
feat(hooks): capture and surface lifecycle hook output

glours:
## What this PR delivers

Builds on #14088 (merged), which added the hook output tail and error reporting foundation.

**Context cancellation fix:** `runHook` previously blocked indefinitely on the exec-attach reader. The copy goroutine held the connection open, so the first Ctrl+C had no effect — only the OS kill on the second signal interrupted the process, skipping cleanup. Fix: run the copy in a background goroutine and `select` on `ctx.Done()` vs completion; on cancellation, close the connection to unblock the goroutine and return `ctx.Err()` directly (no `ExecInspect` on the cancel path). Restores the behaviour of the old `runWaitExec`.

**outputTail robustness:**
- `Write()` rewritten to block-scan with `bytes.IndexByte` (O(n) vs O(n²) byte-by-byte loop); partial line byte-capped via `appendToPartial`.
- `strings.ToValidUTF8` applied at both truncation sites so a cut multi-byte rune does not corrupt the error message.
- `pushLine` handles mid-line `\r` (progress bars from curl/apt/pip): strip trailing `\r`, then keep the segment after the last bare `\r`.

**pre_start parity (from #14088 base):** log stream always opened even without a listener so the tail is populated in detached mode; stderr-biased error reporting.

## What was intentionally excluded

The client-side NDJSON hook-log store and `compose logs --hooks` flag were prototyped in an earlier revision of this branch but have been dropped. Persisting hook output belongs in the engine (see moby/moby#32047, moby/moby#9527), not as a client-side workaround. That work is deferred until an engine-native solution exists.

## Testing

- `TestRunHook_ContextCancellation`: net.Pipe that never writes; asserts `context.Canceled` within 5 s of `cancel()`.
- `TestRunHook_FailureIncludesOutput/with_listener`: asserts listener receives all output lines on failure.
- `TestRunHook_StderrBias`: uses a capturing listener; asserts both stream lines are delivered.
- `TestOutputTail`: CRLF stripping, progress-bar `\r`, UTF-8-safe byte truncation.

### DIFF
--- a/pkg/compose/down.go
+++ b/pkg/compose/down.go
@@ -107,6 +107,10 @@ func (s *composeService) down(ctx context.Context, projectName string, options a
 		}
 	}
 
+	if err := s.removePreStartHookContainers(ctx, projectName, options.Services); err != nil {
+		return err
+	}
+
 	ops := s.ensureNetworksDown(ctx, project)
 
 	if options.Images != "" {
@@ -384,4 +388,44 @@ func (s *composeService) getProjectWithResources(ctx context.Context, containers
 	project.Networks = networks
 
 	return project, nil
+}
+
+// removePreStartHookContainers force-removes any pre_start hook containers that
+// were retained after a failed hook run. These containers are created without a
+// ConfigHashLabel, so getContainers and the normal teardown path never see them;
+// without this step they would survive compose down. When services is non-empty
+// the cleanup is scoped to those services; otherwise the whole project is swept.
+// Individual removal failures are logged at warn level and do not abort teardown.
+func (s *composeService) removePreStartHookContainers(ctx context.Context, projectName string, services []string) error {
+	var filters []client.Filters
+	if len(services) == 0 {
+		f := projectFilter(projectName)
+		f.Add("label", hookFilter(preStartHookType))
+		filters = []client.Filters{f}
+	} else {
+		for _, service := range services {
+			f := projectFilter(projectName)
+			f.Add("label", serviceFilter(service))
+			f.Add("label", hookFilter(preStartHookType))
+			filters = append(filters, f)
+		}
+	}
+	for _, f := range filters {
+		res, err := s.apiClient().ContainerList(ctx, client.ContainerListOptions{
+			All:     true,
+			Filters: f,
+		})
+		if err != nil {
+			return err
+		}
+		for _, ctr := range res.Items {
+			if _, removeErr := s.apiClient().ContainerRemove(ctx, ctr.ID, client.ContainerRemoveOptions{
+				Force:         true,
+				RemoveVolumes: true,
+			}); removeErr != nil {
+				logrus.Warnf("failed to remove retained pre_start hook container %s: %v", ctr.ID, removeErr)
+			}
+		}
+	}
+	return nil
 }
--- a/pkg/compose/hook.go
+++ b/pkg/compose/hook.go
@@ -216,7 +216,7 @@ func (t *outputTail) pushLine(line string) {
 
 // String returns the kept output, newest content last, trimmed to maxBytes.
 func (t *outputTail) String() string {
-	lines := t.lines
+	lines := append([]string(nil), t.lines...)
 	if partial := strings.TrimSpace(strings.ToValidUTF8(t.partial.String(), "")); partial != "" {
 		lines = append(lines, partial)
 	}
--- a/pkg/compose/hook.go
+++ b/pkg/compose/hook.go
@@ -17,6 +17,7 @@
 package compose
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"io"
@@ -39,27 +40,33 @@ const (
 )
 
 func (s *composeService) runHook(ctx context.Context, ctr container.Summary, service types.ServiceConfig, hook types.ServiceHook, listener api.ContainerEventListener) error {
-	// The output of a hook is the only explanation of why it failed, so keep the
-	// tail of it even when no listener is attached: the exec used to run
-	// unattached in that case, the daemon discarded both streams, and the caller
-	// was left with a bare exit code. Only surfaced on failure, and bounded.
-	tail := newOutputTail(hookOutputTailLines, hookOutputTailBytes)
+	// Keep stdout and stderr in separate tails so the error message can prefer
+	// stderr (which usually holds the real error) over stdout progress noise.
+	tailOut := newOutputTail(hookOutputTailLines, hookOutputTailBytes)
+	tailErr := newOutputTail(hookOutputTailLines, hookOutputTailBytes)
 
-	var out io.Writer = tail
+	// wOut/wErr receive the demultiplexed stdout/stderr from StdCopy. When a
+	// listener is attached they also forward each line to it.
+	var wOut, wErr io.Writer
+	wOut = tailOut
+	wErr = tailErr
 
 	if listener != nil {
-		wOut := utils.GetWriter(func(line string) {
+		source := getContainerNameWithoutProject(ctr) + " ->"
+		// stdout and stderr share one listener writer: ContainerEvent has no stream
+		// field, so both appear identically in the live display.
+		lw := utils.GetWriter(func(line string) {
 			listener(api.ContainerEvent{
 				Type:    api.HookEventLog,
-				Source:  getContainerNameWithoutProject(ctr) + " ->",
+				Source:  source,
 				ID:      ctr.ID,
 				Service: service.Name,
 				Line:    line,
 			})
 		})
-		defer wOut.Close() //nolint:errcheck
-
-		out = io.MultiWriter(wOut, tail)
+		defer lw.Close() //nolint:errcheck
+		wOut = io.MultiWriter(lw, tailOut)
+		wErr = io.MultiWriter(lw, tailErr)
 	}
 
 	exec, err := s.apiClient().ExecCreate(ctx, ctr.ID, client.ExecCreateOptions{
@@ -92,13 +99,31 @@ func (s *composeService) runHook(ctx context.Context, ctr container.Summary, ser
 	}
 	defer attach.Close()
 
-	if service.Tty {
-		_, err = io.Copy(out, attach.Reader)
-	} else {
-		_, err = stdcopy.StdCopy(out, out, attach.Reader)
-	}
-	if err != nil {
-		return err
+	// Run the copy in a goroutine so that a context cancellation (Ctrl+C) can
+	// interrupt it. Without this the blocked Read inside StdCopy/io.Copy would
+	// prevent the first Ctrl+C from taking effect.
+	copyDone := make(chan error, 1)
+	go func() {
+		var cErr error
+		if service.Tty {
+			_, cErr = io.Copy(wOut, attach.Reader)
+		} else {
+			_, cErr = stdcopy.StdCopy(wOut, wErr, attach.Reader)
+		}
+		copyDone <- cErr
+	}()
+
+	select {
+	case err = <-copyDone:
+		if err != nil {
+			return err
+		}
+	case <-ctx.Done():
+		// Close the connection to unblock the blocked Read in the copy goroutine,
+		// then wait for it to exit before returning so there is no goroutine leak.
+		attach.Close()
+		<-copyDone
+		return ctx.Err()
 	}
 
 	inspected, err := s.apiClient().ExecInspect(ctx, exec.ID, client.ExecInspectOptions{})
@@ -106,13 +131,23 @@ func (s *composeService) runHook(ctx context.Context, ctr container.Summary, ser
 		return err
 	}
 	if inspected.ExitCode != 0 {
-		if output := tail.String(); output != "" {
-			return fmt.Errorf("%s hook exited with status %d: %s", service.Name, inspected.ExitCode, output)
-		}
-
-		return fmt.Errorf("%s hook exited with status %d", service.Name, inspected.ExitCode)
+		return hookExitError(service.Name, inspected.ExitCode, tailOut, tailErr)
 	}
 	return nil
+}
+
+// hookExitError builds a hook failure error. Stderr content is preferred over
+// stdout because it usually holds the real failure reason rather than progress
+// noise. Falls back to stdout when stderr is empty.
+func hookExitError(serviceName string, code int, tailOut, tailErr *outputTail) error {
+	output := tailErr.String()
+	if output == "" {
+		output = tailOut.String()
+	}
+	if output == "" {
+		return fmt.Errorf("%s hook exited with status %d", serviceName, code)
+	}
+	return fmt.Errorf("%s hook exited with status %d: %s", serviceName, code, output)
 }
 
 // outputTail keeps the last lines written to it, bounded by a line and a byte
@@ -131,27 +166,44 @@ func newOutputTail(maxLines, maxBytes int) *outputTail {
 }
 
 func (t *outputTail) Write(p []byte) (int, error) {
-	for _, b := range p {
-		if b != '\n' {
-			// Cap the line being assembled, so a hook printing megabytes without
-			// a newline cannot grow this buffer without bound.
-			if t.partial.Len() < t.maxBytes {
-				t.partial.WriteByte(b)
-			}
-
-			continue
+	n := len(p)
+	for len(p) > 0 {
+		i := bytes.IndexByte(p, '\n')
+		if i < 0 {
+			t.appendToPartial(p)
+			break
 		}
-
-		t.pushLine(t.partial.String())
+		t.appendToPartial(p[:i])
+		// ToValidUTF8 ensures a byte-truncated multi-byte rune does not corrupt
+		// the error message.
+		t.pushLine(strings.ToValidUTF8(t.partial.String(), ""))
 		t.partial.Reset()
+		p = p[i+1:]
 	}
+	return n, nil
+}
 
-	return len(p), nil
+// appendToPartial appends b to the in-progress partial line, capped at maxBytes
+// so a hook printing megabytes without a newline cannot grow the buffer unbounded.
+func (t *outputTail) appendToPartial(b []byte) {
+	if remaining := t.maxBytes - t.partial.Len(); remaining > 0 {
+		if len(b) > remaining {
+			b = b[:remaining]
+		}
+		_, _ = t.partial.Write(b)
+	}
 }
 
 // pushLine appends a line and drops the oldest one when the limit is reached.
 func (t *outputTail) pushLine(line string) {
+	// Strip trailing \r (from \r\n sequences). For progress-bar output that uses
+	// bare \r to rewind the cursor (curl, apt, pip), keep only the last segment
+	// after the final \r so the visible content is preserved rather than the
+	// intermediate states.
 	line = strings.TrimRight(line, "\r")
+	if idx := strings.LastIndex(line, "\r"); idx >= 0 {
+		line = line[idx+1:]
+	}
 	if strings.TrimSpace(line) == "" {
 		return
 	}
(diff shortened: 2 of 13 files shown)


pkg/compose/hook_logs.go
```diff
@@ -0,0 +1,285 @@
...
+func newSinkWriter(sink *hookLogSink, stream string) *sinkWriter {
+	return &sinkWriter{sink: sink, stream: stream}
+}
+
+// Write implements io.Writer. Never returns an error.
+func (w *sinkWriter) Write(p []byte) (int, error) {
+	if w.sink == nil || w.full {
+		return len(p), nil
+	}
+	w.buf = append(w.buf, p...)
```
docker-agent: **[medium] `sinkWriter.buf` grows without bound for newline-free output**

`sinkWriter.Write` appends all incoming bytes to `w.buf` and only flushes when a `'\n'` is found — there is no cap on how large `w.buf` can grow. A hook that emits megabytes of output without a newline character (e.g. a binary payload, a long SQL statement, or a stalled process printing dots) holds all of it in memory until either a newline appears or the connection closes.

By contrast, `outputTail.Write` — which also collects hook output — caps its partial-line buffer at `maxBytes`. The `hookLogSink` enforces a 1 MiB cap on *committed* data, but that check only runs once a complete line is flushed; the intermediate accumulation in `w.buf` is unprotected.

Fix: add a cap immediately after the `append`, mirroring `outputTail`:
```go
w.buf = append(w.buf, p...)
if len(w.buf) > sinkMaxBytes {
    w.buf = w.buf[len(w.buf)-sinkMaxBytes:]
}
```

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 100/100 |

docker-agent (commented):

Two real bugs introduced by this PR — one memory-management issue and one output-correctness issue — plus a usability gap in `--hooks --follow`. See inline comments for details.

pkg/compose/logs.go
```diff
@@ -49,6 +49,12 @@ func (s *composeService) Logs(
 		})
 	}
 
+	if options.Hooks {
```
docker-agent: **[low] `--hooks --follow` shows a snapshot, not live hook output**

`logHooks` reads existing NDJSON files once and returns. When `--follow` is also active, the monitor goroutine picks up new *container* log streams for restarted containers, but there is no equivalent watcher for new hook-log files. Hook runs that occur after `compose logs --hooks --follow` starts — such as hooks fired by a concurrent `compose restart` or a new service being scaled up — are silently omitted.

This creates an asymmetry: container logs follow live, hook logs are a point-in-time snapshot. A user combining `--hooks --follow` will reasonably expect both to be live, but only the container side is.

Consider either:
- Documenting `--hooks` as a post-mortem flag (not compatible with `--follow`), or
- Implementing a periodic re-scan of the hook-log directory when `options.Follow` is true.

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 67/100 |

pkg/compose/hook.go
```diff
@@ -90,38 +133,85 @@ func (s *composeService) runHook(ctx context.Context, ctr container.Summary, ser
...
+// String returns the kept output, newest content last, trimmed to maxBytes.
+func (t *outputTail) String() string {
+	lines := t.lines
+	if partial := strings.TrimSpace(t.partial.String()); partial != "" {
+		lines = append(append([]string{}, lines...), partial)
+	}
+
+	output := strings.TrimSpace(strings.Join(lines, "\n"))
+	if len(output) > t.maxBytes {
+		output = output[len(output)-t.maxBytes:]
```
docker-agent: **[low] Byte-slice truncation in `outputTail.String()` can produce invalid UTF-8**

The tail is trimmed by byte count:

```go
output = output[len(output)-t.maxBytes:]
```

If the hook's output contains multi-byte UTF-8 characters (e.g. accented letters, CJK, or emoji) and the truncation boundary falls inside one of them, the resulting string has an invalid byte sequence. That string is then embedded directly in the error message returned to the caller.

```suggestion
		output = strings.ToValidUTF8(output[len(output)-t.maxBytes:], "")
```

(`strings.ToValidUTF8` is available since Go 1.13 and is already imported via the `strings` package used elsewhere in this file.)

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 100/100 |

ndeloof:
Some notes: 

the first two commits are the right fix, and they're @dennislapchenko's work from #14088: I'd rather we merge his PR as-is and rebase the rest on top, so authorship lands where it belongs.

On the third commit, I'm going to push back on the NDJSON store.

First, scoping the actual problem: in attached mode hook output already streams to the console (`HookEventLog`), and with the tail-in-error fix a failing hook is now diagnosable in every mode — detached, restart, run, and library usage through Start. What remains uncovered is only post-hoc inspection of hook output, and it's worth being precise about why that gap exists: post_start/pre_stop are execs, and the engine's logging driver never sees exec output. There is genuinely nothing for compose logs to read today.

**But** a client-side store is the wrong place to fill that gap. Compose's contract is that project state lives in the engine: any client, from any machine, can operate a project it didn't create. With logs under the client's config dir, `logs --hooks` shows nothing from another workstation or CI, a remote docker context writes them on the wrong machine, cleanup depends on down running (a crash leaves orphans), and we now own a second log pipeline — rotation, concurrency, platform paths — duplicating what the engine's logging driver is for.

I'd rather we go engine-native:

- pre_start hooks are containers — their output does go through the logging driver, and we only lose it because we run them with `AutoRemove: true`. Keeping the container around on failure (removed by the next up/down, like any other project container, and/or after successful completion) makes compose logs and docker logs work with zero new machinery. That covers the case where post-hoc inspection matters most.
- For post_start/pre_stop, the structural fix is upstream: an exec-create option to tee output into the container's logging driver. That would give us logs --hooks for free, stateless, for every client. I'll open that discussion on moby; until then the bounded tail on failure covers the diagnosis need, which is what #14088 set out to solve.

So: merge #14088, add failed-pre_start retention here, drop the store. WDYT?

ndeloof:
see https://github.com/moby/moby/issues/32047 / https://github.com/moby/moby/issues/9527

ndeloof (commented):
Scope is exactly right: pure client-side, bounded tail in the error + retained container for post-mortem, with persistence deferred to the engine — and the test coverage is outstanding (31 new tests including every cleanup-failure path). Nice work on the Ctrl-C fix and the `\r`/UTF-8 handling too.

Inline comments below. I'd consider the first three worth fixing in this PR (anonymous-volume leak in the orphan sweep, retention on user cancellation, stale AutoRemove comments); the rest is fine as follow-ups:

- `down <svc>` cleans only the named services' hook containers while the traversal removes the service *and its dependents* — the dependents' retained hooks survive a targeted down.
- a `ContainerList` failure aborts the teardown mid-way, while the doc comment promises removal failures don't; warn-and-continue would match.
- the two near-identical sweep helpers (`removeOrphanPreStartContainers` / `removePreStartHookContainers`) could merge into one, which would structurally prevent the RemoveVolumes divergence.
- ordering: retained hooks are removed *after* the service containers, yet they reference the service's volumes via `VolumesFrom` — worth checking `down -v` still reclaims the service's anonymous volumes in that order (sweeping hooks first would be safer).
- "intentionally excluded" could also point at the concrete engine-native path in flight: moby/moby#53406 / moby/moby#53407 (opt-in exec output capture through the logging driver) — that's the API a future `logs --hooks` would consume.

pkg/compose/pre_start.go
```diff
@@ -209,6 +242,29 @@ func (s *composeService) connectPreStartExtraNetworks(ctx context.Context, proje
...
+	f.Add("label", hookFilter(preStartHookType))
+	res, err := s.apiClient().ContainerList(ctx, client.ContainerListOptions{
+		All:     true,
+		Filters: f,
+	})
+	if err != nil {
+		return err
+	}
+	for _, ctr := range res.Items {
+		if _, removeErr := s.apiClient().ContainerRemove(ctx, ctr.ID, client.ContainerRemoveOptions{Force: true}); removeErr != nil {
```
ndeloof: This sweep is missing `RemoveVolumes: true`: the success path and the `down` sweep both pass it, so a hook whose image declares `VOLUME`s leaks its anonymous volumes when *this* path does the cleanup. (Merging this helper with `removePreStartHookContainers` in down.go would prevent the divergence structurally.)

pkg/compose/pre_start.go
```diff
@@ -119,7 +133,21 @@ func (s *composeService) runPreStartHook(
 	waitErr := waitPreStart(ctx, service.Name, index, waitRes)
 	cancelLogs()
 	<-logsDone
-	return waitErr
+	if waitErr != nil {
```
ndeloof, line 134: Two things on the retention branch:

1. `waitPreStart` returns `ctx.Err()` on cancellation, so a plain Ctrl-C lands here too: the container is retained and the tail gets wrapped around `context.Canceled`. A user cancellation isn't a post-mortem — at minimum don't decorate `context.Canceled` with the tail, and arguably don't retain in that case.
2. The docs say the operator can run `docker logs <id>`, but neither the ID nor a name appears in the error (hook containers are created unnamed). Including the short ID here — or better, naming the container (`<project>-<service>-pre_start-<i>`) — would make the retention actually actionable.

pkg/compose/pre_start.go
```diff
@@ -94,7 +108,7 @@ func (s *composeService) runPreStartHook(
 	// the connection open cannot deadlock `<-logsDone`.
 	logCtx, cancelLogs := context.WithCancel(ctx)
 	defer cancelLogs()
-	logsDone := s.streamPreStartLogs(logCtx, created.ID, service, index, listener)
+	logsDone, getTail := s.streamPreStartLogs(logCtx, created.ID, service, index, listener)
 
 	if _, err := s.apiClient().ContainerStart(ctx, created.ID, client.ContainerStartOptions{}); err != nil {
 		// AutoRemove only fires after a successful start, so the never-started
```
ndeloof: This comment (and its sibling in the `connectPreStartExtraNetworks` failure path) still justifies the explicit removal with "AutoRemove only fires after a successful start" — but AutoRemove is now `false` unconditionally. The behavior is right, the rationale is stale.

pkg/compose/down.go
```diff
@@ -385,3 +389,43 @@ func (s *composeService) getProjectWithResources(ctx context.Context, containers
...
+// the cleanup is scoped to those services; otherwise the whole project is swept.
+// Individual removal failures are logged at warn level and do not abort teardown.
+func (s *composeService) removePreStartHookContainers(ctx context.Context, projectName string, services []string) error {
+	var filters []client.Filters
+	if len(services) == 0 {
+		f := projectFilter(projectName)
+		f.Add("label", hookFilter(preStartHookType))
+		filters = []client.Filters{f}
+	} else {
+		for _, service := range services {
```
ndeloof, line 406: Follow-up: scoping by `options.Services` diverges from what the teardown actually removes — `WithRootNodesAndDown` takes down the named services *plus their dependents*, so a targeted `down web` leaves the dependents' retained hook containers behind.

pkg/compose/down.go
```diff
@@ -385,3 +389,43 @@ func (s *composeService) getProjectWithResources(ctx context.Context, containers
...
+			filters = append(filters, f)
+		}
+	}
+	for _, f := range filters {
+		res, err := s.apiClient().ContainerList(ctx, client.ContainerListOptions{
+			All:     true,
+			Filters: f,
+		})
+		if err != nil {
+			return err
```
ndeloof, line 419: Follow-up: this `return err` aborts the teardown (networks/volumes/images not yet processed) on a listing failure, while the doc comment above promises removal failures don't abort. Warn-and-continue would be consistent.

pkg/compose/down.go
```diff
@@ -107,6 +107,10 @@ func (s *composeService) down(ctx context.Context, projectName string, options a
 		}
 	}
 
+	if err := s.removePreStartHookContainers(ctx, projectName, options.Services); err != nil {
```
ndeloof, line 110: Follow-up / to verify: the retained hook containers reference the service containers' volumes via `VolumesFrom`, and they are removed *after* the service containers — so the service's `rm -v` may fail to reclaim anonymous volumes still referenced by a retained hook (they'd only go away with the hook's own `RemoveVolumes` afterwards). Sweeping hooks *before* the service containers would be the safer order.

pkg/compose/hook.go
```diff
@@ -165,13 +217,15 @@ func (t *outputTail) pushLine(line string) {
 // String returns the kept output, newest content last, trimmed to maxBytes.
 func (t *outputTail) String() string {
 	lines := t.lines
-	if partial := strings.TrimSpace(t.partial.String()); partial != "" {
-		lines = append(append([]string{}, lines...), partial)
+	if partial := strings.TrimSpace(strings.ToValidUTF8(t.partial.String(), "")); partial != "" {
+		lines = append(lines, partial)
```
ndeloof, line 221: Nit: this dropped the defensive copy (`append(append([]string{}, lines...), partial)`) — `append(lines, partial)` can write into `t.lines`' backing array. Harmless with the current single-goroutine write / read-after-done usage, but fragile for the one line it saves.

glours:
> 🤖 **Automated implementer agent** — *this comment was posted by the implementer bot from Docker Agentic Platform, not by a human developer*

Addressed @ndeloof's three must-fix items in commit **60fce351**:

**1. RemoveVolumes:true in orphan sweep** (discussion_r3821978833)
Added `RemoveVolumes:true` to `removeOrphanPreStartContainers` (now matches the `down.go` sweep and the success-path explicit remove). Same flag applied to the ContainerStart-failure and connectPreStartExtraNetworks-failure cleanup paths.

**2. Cancellation does not retain; short ID in failure error** (discussion_r3821978843)
- Added `ctx.Err() != nil` guard: on Ctrl-C, removes the container (Force+RemoveVolumes) and returns the raw context error without tail decoration. New test: `TestPreStart_CancellationRemovesContainer`.
- Genuine hook failures now include the short container ID: `exited with code 1: ... (hook container abc123 retained for inspection)`. `TestPreStart_FailureRetainsContainer` asserts it.

**3. Stale AutoRemove comments** (discussion_r3821978846)
Rewrote the three stale comments (ContainerStart failure path, connectPreStartExtraNetworks failure path, logCtx comment) to describe the current explicit-removal behaviour without the old AutoRemove rationale.

All pre_start tests pass; lint is clean.

glours:
> 🤖 **Automated implementer agent** — *this comment was posted by the implementer bot from Docker Agentic Platform, not by a human developer*

Follow-up issues opened for the non-blocking ndeloof items:
- #14107 — `compose down <svc>` skips dependent hook containers (down.go:406)
- #14108 — ContainerList failure aborts teardown instead of warn-and-continue (down.go:419)
- #14109 — retained hook containers removed before service containers on `down -v` (down.go:110)
- #14110 — defensive copy missing in outputTail.String() append (hook.go:221)

.github/workflows/ci.yml
```diff
@@ -32,7 +32,6 @@ jobs:
           - validate-go-mod
           - validate-headers
           - validate-docs
```
ndeloof, line 34: This commit removes `validate-mocks` from the validate matrix — the check just introduced on main (#14102), and nothing in this PR justifies it (the commit message doesn't mention it either). Looks like a rebase leftover; please restore it. If the check is red on this branch, that's a mocks drift to fix with `make mocks`, not a check to drop.

.github/workflows/scorecards.yml
```diff
@@ -61,6 +61,6 @@ jobs:
       
       # Upload the results to GitHub's code scanning dashboard.
       - name: "Upload to code-scanning"
-        uses: github/codeql-action/upload-sarif@ff2f1c621b7f889edc0d3c761ac2e6a3f8cdb0dd # v4.37.7
+        uses: github/codeql-action/upload-sarif@5595ccaf912efad79be6eef63a5619ff05969be3 # v4.37.6
```
ndeloof: Same commit downgrades codeql-action v4.37.7 → v4.37.6, reverting the dependabot bump merged on main. Rebase artifact — please drop this hunk.

pkg/compose/pre_start.go
```diff
@@ -119,7 +131,35 @@ func (s *composeService) runPreStartHook(
 	waitErr := waitPreStart(ctx, service.Name, index, waitRes)
 	cancelLogs()
 	<-logsDone
-	return waitErr
+	if waitErr != nil {
+		// Ctrl-C is a user cancellation, not a hook failure: remove the container
+		// and return the raw context error without decorating it with the tail or
+		// retaining the container for post-mortem inspection.
+		if ctx.Err() != nil {
```
ndeloof, line 138: This `ContainerRemove` runs on the already-cancelled `ctx`, so in production it fails immediately with `context.Canceled` and lands in the Warnf below — the container is retained despite this branch's intent (the mock in `TestPreStart_CancellationRemovesContainer` doesn't honor ctx, which is why the test passes). Use `context.WithoutCancel(ctx)` for the cleanup call — same pattern `up` uses for its teardown.

ndeloof (commented):
Thanks — the three must-fix items from my previous review are all properly addressed: `RemoveVolumes: true` on every sweep/cleanup path (with the mocks updated to pin it), the cancellation branch no longer retains nor decorates `context.Canceled`, the retained-container short ID makes the error actionable, and the stale AutoRemove comments now state the actual behavior. The new `TestPreStart_CancellationRemovesContainer` and the short-ID assertion are exactly the locks I hoped for, and the start_test.go adaptation is correct and well-explained.

Two **unrelated regressions slipped into the last commit** (rebase leftovers, I assume) that must be dropped before merge — see inline: the `validate-mocks` CI check removal and the codeql-action downgrade.

One real bug remains in the new cancellation path (inline): the cleanup `ContainerRemove` runs on the already-cancelled context, so in production it fails immediately and the container is left behind anyway (the orphan sweep saves the day on the *next* run, but the code's promise is dead on arrival). `context.WithoutCancel(ctx)` fixes it.

The agreed follow-ups (down scoping vs dependents, list-failure aborting teardown, merging the two sweep helpers, hook-before-service removal order, moby#53406/#53407 pointer) remain open — fine by me to land them separately.

ndeloof (commented):
The rebase regressions are gone (`validate-mocks` back in the matrix, codeql-action back on v4.37.7) and the `outputTail.String()` defensive copy is restored — thanks, and the squashed history reads well.

One item from my previous review survived the rewrite though (the inline thread got marked outdated by the force-push, so re-anchoring it): the cancellation-path cleanup still runs `ContainerRemove` on the already-cancelled context, so in production it can't succeed. One-line fix, and with it I have nothing else blocking — LGTM once addressed.

pkg/compose/pre_start.go
```diff
@@ -119,7 +131,35 @@ func (s *composeService) runPreStartHook(
...
+			if _, removeErr := s.apiClient().ContainerRemove(ctx, created.ID, client.ContainerRemoveOptions{Force: true, RemoveVolumes: true}); removeErr != nil {
```
ndeloof: Still passing the already-cancelled `ctx` here: the moby client aborts the request immediately with `context.Canceled`, so this removal always lands in the Warnf and the container is left behind despite the branch's intent (the gomock in `TestPreStart_CancellationRemovesContainer` doesn't honor ctx, which is why the test stays green). `context.WithoutCancel(ctx)` on this call fixes it.

pkg/compose/pre_start.go
```diff
@@ -119,7 +131,35 @@ func (s *composeService) runPreStartHook(
 	waitErr := waitPreStart(ctx, service.Name, index, waitRes)
 	cancelLogs()
 	<-logsDone
-	return waitErr
+	if waitErr != nil {
+		// Ctrl-C is a user cancellation, not a hook failure: remove the container
+		// and return the raw context error without decorating it with the tail or
+		// retaining the container for post-mortem inspection.
+		if ctx.Err() != nil {
+			if _, removeErr := s.apiClient().ContainerRemove(context.Background(), created.ID, client.ContainerRemoveOptions{Force: true, RemoveVolumes: true}); removeErr != nil {
```
ndeloof, line 139: Non-blocking:
```suggestion
			if _, removeErr := s.apiClient().ContainerRemove(context.WithoutCancel(ctx), created.ID, client.ContainerRemoveOptions{Force: true, RemoveVolumes: true}); removeErr != nil {
```
`WithoutCancel` keeps the tracing/values of the parent context while dropping its cancellation — same effect as `Background()` for the API call, better trace correlation, and consistent with `runInteractiveUp`'s teardown.

ndeloof (approved):
The cancellation-path removal now survives the cancelled context — that was my last blocking point. LGTM.

One optional polish inline (fine as-is or in a follow-up): `context.WithoutCancel(ctx)` instead of `context.Background()` keeps the context values (tracing spans notably) while still dropping cancellation, and matches the idiom `runInteractiveUp` already uses for its teardown.

PR #14093, opened 2026-08-19
adopt compose-go jobs and container-spec layering

### LINKED ISSUE #13934: pre_start/init-Containers should support own/overwritten Volumes
### Description

Hi,

i appreciate the feature of pre_start/Init-Containers. That is really cool and solves many things, where we built own solutions with startup-scripts and stuff...

But i miss a thing in it:
The configuration as far as i can see currently does not support volumes for the pre_start-Container. It should.
And: I have volumes mounted readonly to my service (https-Certificates and stuff), which the init-Container should be able to write. It would by nice to

* be able to define ADDITONAL volumes
* be able to overwrite/change "base-service"-volumes

I am sorry, i am not able to do a pullrequest, "go" is not my language and i am not into compose-codewise.

Thanks a lot in advance!

Best regards
Stefan

PS: I partly mentioned that in my reply to the pre_start-pull-request here: https://github.com/docker/compose/pull/13862#issuecomment-4887343263

### LINKED ISSUE #13939: pre_start containers don't inherit/support extra_hosts
### Description

Hi, thanks for the awesome pre_start feature!

We're running into the problem that we 're not seeing `extra_hosts` in the `pre_start` containers being applied. For example:

```yaml
services:
  test:
    image: busybox
    extra_hosts:
      - "somehost:162.242.195.82"
    pre_start:
      - command: cat /etc/hosts
    command: cat /etc/hosts
```

Then, running `docker compose up` outputs:
```
laughing_hertz  | 127.0.0.1	localhost
laughing_hertz  | ::1	localhost ip6-localhost ip6-loopback
laughing_hertz  | fe00::	ip6-localnet
laughing_hertz  | ff00::	ip6-mcastprefix
laughing_hertz  | ff02::1	ip6-allnodes
laughing_hertz  | ff02::2	ip6-allrouters
laughing_hertz  | 172.19.0.2	731eafcc9089
laughing_hertz exited with code 0
test-1          | 127.0.0.1	localhost
test-1          | ::1	localhost ip6-localhost ip6-loopback
test-1          | fe00::	ip6-localnet
test-1          | ff00::	ip6-mcastprefix
test-1          | ff02::1	ip6-allnodes
test-1          | ff02::2	ip6-allrouters
test-1          | 162.242.195.82	somehost
test-1          | 172.19.0.2	876dcd9db560
test-1 exited with code 0
```

The `somehost` is not visible in the `laughing_hertz` `pre_start` container.

ndeloof:
Adopts compose-spec/compose-go#866 (jobs top-level element, container specification layered as ContainerSpec/WorkloadSpec, pre_start hooks carrying the full container specification).

This first step adapts the types and gives jobs their first runtime entry point:

- composite literals setting moved fields are wrapped into the embedded `ContainerSpec`/`WorkloadSpec` (no behavior change; promoted field access was already source-compatible)
- `pre_start` handling is typed against `PreStartHook`; exec hooks (`post_start`/`pre_stop`) lose the image field they never honored — now enforced by construction
- `up` warns about declared jobs (pointing at `docker compose run` to trigger them) and fails on active (profile-enabled) **scheduled** jobs with `scheduled jobs are not supported in this version`: silently not scheduling would break expectations, while manual jobs just wait for an explicit trigger
- **`docker compose run <job>` executes a job exactly like a service**: the job — a `ContainerSpec`+`WorkloadSpec`, the same layers a service is made of — is materialized as a service for the one-off machinery; its profile is activated, its dependencies start as they would for a service, the exit code flows back. Per the spec, manual execution is always available — scheduled jobs included — unless the job explicitly opts out with `triggers.manual: false` (tri-state `*bool` in compose-go), which run rejects. All locked by e2e scenarios (scheduled rejection by `up` before any resource is created; manual run starting its `depends_on` services first; `manual: false` refusal).

- **`pre_start` hooks inherit the whole container specification, resolved at load time**: compose-go completes each hook with the service's container-spec attributes while loading (idempotently across `config` round-trips, thanks to the strictly-identical-entry merge dedup), so the model itself carries the spec a hook runs with and the runtime consumes it as-is through the standard create path — every `ContainerSpec` attribute (resources, capabilities, dns, sysctls, ...) inherits with zero attribute-specific code, following the exact multi-file merge rules (command replaces, environment merges per key, extra_hosts accumulate). The only deliberate exception is volumes, inherited at runtime through `volumes_from` (the only way to share anonymous and image volumes), with hook-declared mounts taking precedence per target. This closes the two standing asks against init containers: inheriting/declaring `extra_hosts` (fixes #13939) and giving the hook read-write access to a volume the service mounts read-only (fixes #13934) — both locked by e2e scenarios.

compose-spec/compose-go#866 is merged: `go.mod` now points at the merged compose-go head (`v2.15.1-0.20260910154416-11feead015fc`), no replace directive involved — this PR is self-contained and mergeable. Jobs are hereby exposed end to end: declared in the model, warned about and guarded by `up` (scheduled jobs rejected until a scheduler exists), and runnable through `docker compose run <job>`. A compose-go tagged release can still substitute the pseudo-version before merge if preferred.

---

**Update (independent review follow-ups).** An adversarial review pass surfaced real gaps, all addressed in the rebuilt stack:

- **`volumes_from` (and `service:`-scoped `network_mode`/`ipc`/`pid`) inherited into a `pre_start` hook now resolve to live container IDs** like the service create path — the daemon used to receive raw service names and fail `up` for any service combining `volumes_from` with a hook (e2e-locked). Hook `labels` merge into the container's labels (the runtime identification set wins), and the <API 1.44 network fallback joins the hook's own networks.
- **A job's `env_file`/`label_file` resolve through `run`**: materialization now happens during loading, before environment resolution — and the retry is keyed on "the target is a declared job", never on matching an error message, so a plain typo keeps the precise selection error and unrelated services' env_file still don't need to exist.
- **Job→job `depends_on` executes**: the whole job closure materializes, each dependency running to completion through the machinery a service dependency uses (e2e-locked). Materialized jobs carry the standard compose labels — without them their containers were invisible to every label-driven path — and their `x-*` extensions.
- **`create` and `start` refuse active scheduled jobs like `up` does**, keeping the "before any resource is created" promise on every entry point.
- **`publish` safeguards and image pinning cover jobs**: sensitive-env/env_file/bind-mount/build-only checks, digest override, application index, and `config --images`.

**Note on config hashes**: the container-spec layering reorders `ServiceConfig` fields, which would have changed every service's config-hash (mass recreation on upgrade) — #14215 pins the hash byte layout to its historical form — fully backward compatible, zero recreation — and must land first; this PR then deletes its transient continuity test in the compose-go bump commit, as that test documents.


glours (commented):
A few things worth fixing before this lands, even given the interim/client-side nature of this jobs support:                                                                                                                                          
                                                                                                                                                                                                                                                        
  **cmd/compose/up.go:362** — `warnIgnoredJobs` iterates `project.AllJobs()` (includes profile-disabled jobs), while `rejectScheduledJobs` (up.go:379, also called from create.go:104 and start.go:63) correctly uses `project.Jobs` (profile-enabled   
  only). Result: `up` can warn about jobs that aren't even part of this invocation's profile selection. Should use `project.Jobs` here too.                                                                                                             
                                                                                                                                                                                                                                                        
  **cmd/compose/run.go:287-292** — the job-detection fallback now retries the full project load on *any* narrowed-load error, not just "no such service". Any unrelated load failure (bad `include:`, interpolation error) now triggers a second full   
  load — duplicating remote `include:` fetches and unsupported-attribute warnings — before falling through to the (still correct) original error. Worth narrowing the retry condition back to the selection-failure case, since job detection doesn't   
  need the broader retry.                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                        
  **pkg/compose/pre_start_test.go** — still no test exercising the security-sensitive fields (`Privileged`, `CapAdd`/`CapDrop`, `SecurityOpt`, `ReadOnly`, `Sysctls`) that hooks now inherit from the service by default. This is pre-existing/permanent
  hook behavior, not something the interim jobs work will replace — worth covering given what's at stake if inheritance breaks silently.                                                                                                                
                                                                                                                                                                                                                                                        
  **pkg/compose/publish.go** — the new job branches (`pushApplicationIndex`, `generateImageDigestsOverride`, `collectEnvCheckFindings`, `checkOnlyBuildSection`, `checkForBindMount`, `checkForSensitiveData`) are correct on inspection but have zero  
  dedicated test coverage — `publish_test.go`'s only changes are mechanical `ContainerSpec` adaptations. This is the secret-scanning gate for `publish`; a future regression here would go undetected by CI.                                            
                                                                                                                                                                                                                                                        
  Nits, non-blocking:                                                                                                                                                                                                                                   
  - pre_start.go: `resolveHookServiceReferences` rebuilds the per-service container map inline instead of reusing `getContainersByService`.                                                                                                             
  - publish.go: findings for a job are still labeled `"service %q: ..."` in the merged prompt message.

ndeloof:
Addressed in d04a2453e, 447c8674b, dff0e0e92:
- `warnIgnoredJobs` now uses `project.Jobs` (profile-enabled only), matching `rejectScheduledJobs`.
- `runProject`'s job-detection fallback now only retries the full/unselected load on the specific "no such service" selection failure (via a new `isNoSuchServiceErr` helper), not on any load error.
- Added `TestPreStart_SecuritySensitiveFieldsHonored` covering hook inheritance of `Privileged`/`CapAdd`/`CapDrop`/`SecurityOpt`/`ReadOnly`/`Sysctls`.
- Added dedicated job-path coverage for `pushApplicationIndex`, `generateImageDigestsOverride`, `collectEnvCheckFindings`, `checkOnlyBuildSection`, and `checkForBindMount` (only `checkForSensitiveData` had a service-level test to mirror before; `pushApplicationIndex` had none at all — the job case is proven via an invalid job image surfacing its own reference error, since exercising the real push path needs a live registry).
- Nits: `resolveHookServiceReferences` now reuses `getContainersByService`; job findings in the publish prompt are no longer mislabeled "service %q".

All new/changed tests fail without their respective fix, confirmed locally. build/vet/lint/tests all green.

ndeloof:
Rebased onto main (07f3de8b7) — root cause of the lint/test failure: the relay work merged into main after this branch was cut added tests using the pre-layering ServiceConfig shape (DependsOn/Networks set directly, PreStart as []ServiceHook), which don't compile against the ContainerSpec/WorkloadSpec split this branch adopts. GitHub tests the hypothetical merge of the PR into main, which is why this only showed up in CI and not locally against the branch's own tip. Adapted both files to the new shape; not a real regression in the reviewed commits. Full build/vet/lint/test green locally afterward (pkg/watch's fsevents flakes aside, unrelated and pre-existing).

ndeloof:
Réécrit l'historique (13 → 9 commits) pour que chaque commit reste un enchaînement logique plutôt qu'une suite de fixups de revue empilés :
- les 2 correctifs trouvés par la revue (`warnIgnoredJobs`/`isNoSuchServiceErr`) sont repliés dans le commit qui a introduit le code concerné (`run: execute jobs exactly like services`), puisqu'ils touchent exactement le même code ;
- le test de couverture des champs de sécurité pre_start est replié dans `pre_start: consume the load-time-resolved hook specification` ;
- le test de couverture publish + le nit de label + son propre correctif e2e sont repliés dans `publish, config --images: jobs are scanned and pinned like services` ;
- seul le correctif d'adaptation post-rebase (`relay_test.go`/`start_test.go` face au découpage ContainerSpec) reste séparé, car il n'est pas un fixup de revue mais une consequence mécanique du rebase sur main.

build/vet/lint/test complets vérifiés après la réécriture, force-push effectué (`d28ad4ce4`).

ndeloof:
Addressed the follow-up review:

- f761e08e7 — fix 1: the fallback reload in `run.go` now uses `skipUnsupportedAttributesWarning` instead of `warnUnsupportedAttributes`, matching the existing convention for secondary loads.
- 2bc29e38c — fix 2: `--images`, `--lock-image-digests` and `--resolve-image-digests` now scan jobs the same way they scan services. Added `TestResolveImageDigestsWithJobs`, which caught a real bug in my first attempt: `WithImagesResolved` only walks a project's Services, so job images silently stayed unpinned until dressed as their own pseudo-project the same way `generateImageDigestsOverride` in publish.go already does. Split `resolveImageDigests` into `resolveServiceImageDigests` (unchanged) + `resolveJobImageDigests` to keep cognitive complexity under the lint threshold.
- c1163975d — fix 3: `create`/`start` now give `job %q can only be triggered with "docker compose run"` instead of the raw compose-go selection error, via a small shared helper (`jobTargetErr`) mirroring run's own job-detection fallback. A genuine typo still gets the original, precise error.
- 8fc491831 — open question: decided `manual: false` should also block a job pulled in transitively via `depends_on`. depends_on doesn't change who caused the execution or when — pulling the job in to satisfy a dependency is still the run command causing that out-of-schedule execution, one hop removed, and manual: false exists specifically to prevent that. Added the check in `materializeJobClosure`, a unit test, and a locking e2e scenario (`TestRunJobDependsOnManualFalseJob`).

All four verified locally: build/vet/lint/test green (`pkg/watch`'s pre-existing fsevents flakes aside), plus manual runs of `create`/`start` against a job target.

ndeloof:
Addressed a third-party review of the previous commit (c1163975d) plus an independent fresh-eyes pass I ran separately, both converging on the same two real gaps in `jobTargetErr`/`create`/`start`:

- `06f51dd17` — `jobTargetErr` matched *any* name in the `[SERVICE...]` argument list against the job set, not the one name compose-go's selection actually failed on. `docker compose create totally-bogus migrate` (a real typo plus a legitimate job) reported the job error instead of the real `no such service: totally-bogus`, masking the typo. Fixed by correlating against the specific failing name; extracted `unselectedJobs` out of `jobTargetErr` for reuse below.
- `2e5a6876c` — two things from the same review round:
  - `create.go`/`start.go` had zero test coverage for their job-target-error wiring (no `create_test.go`/`start_test.go`, and `pkg/e2e/jobs_test.go` only exercised `up`/`run`). Added `TestJobTargetErr` (unit) and `TestCreateRefusesJob`/`TestStartRefusesJob` (e2e).
  - A real bug this gap let through: when `COMPOSE_PROJECT_NAME` is set, `start`'s underlying `projectOrName` helper falls back to a label-driven, file-less project on *any* load failure — including "no such service" for a job — discarding the error entirely. Both `jobTargetErr` (gated on `err != nil`) and `rejectScheduledJobs` (gated on `project != nil`) were skipped on that path, so `docker compose start migrate` silently exited 0 having done nothing (no container exists for a job that was never run). Fixed, with `TestStartRefusesJobWithProjectNameEnv` locking in the repro.

All verified locally: build/vet/lint/test green, including a manual regression check on the correlation fix (reverting it back-to-back against the new unit test confirmed it catches the bug).

glours (approved):
Sounds like a good 1st iteration

PR #14095, opened 2026-08-19
fix(ps): honor psFormat from Docker CLI config when no --format flag given

### LINKED ISSUE #13643: 'docker compose ps' not honoring 'psFormat' setting in '.docker/config.json'

glours:
**What I did**
The `--format` flag defaulted to "table" instead of "", so the `opts.Format == ""` guard in `runPs` never fired for a bare compose ps. 
Setting the default to "" lets the existing config-fallback path work, matching the behaviour of docker ps.

**Related issue**
Fixes #13643
Supersedes #14032

image


docker-agent (commented):
### Assessment: 🟢 APPROVE

The fix is correct and minimal. Changing the `--format` default from `"table"` to `""` allows the existing `opts.Format == ""` guard in `runPs` to fire for a bare `compose ps`, which then reads `dockerCli.ConfigFile().PsFormat` — exactly matching `docker ps` behaviour.

Key paths are all safe:
- `NewContainerFormat` treats both `""` and `"table"` identically (same `case` branch), so a user without `psFormat` in their config gets the same default table output as before.
- Existing `TestPs` subtests use a fresh per-test temp config directory (via `NewParallelCLI`) with no `psFormat` set, so they are unaffected.
- The new `TestPsHonorsPsFormat` E2E test correctly anchors `DOCKER_CONFIG` to the testdata subdirectory containing the custom `psFormat`, uses the declarative Scenario DSL, and asserts the right positive/negative column presence.

ndeloof (approved):

PR #14096, opened 2026-08-19
fix(watch): sync all files on initial_sync, regardless of mtime

### LINKED ISSUE #13725: [BUG] develop.watch initial_sync: true does not fire at container start

glours:
**What I did**
`initialSyncFiles` and `initialSyncDirectory` filtered out any host file whose mtime predated the image creation time, which is exactly the files that `initial_sync` is meant to copy. 
The `imageCreatedTime` lookup (`ContainerList` + `ImageInspect`) was also a silent failure point when the container list returned nothing. 
Remove the filter and the helper entirely: sync all files that are not ignored or bind-mounted.

**Related issue**
Fixes #13725
Supersedes #14031

image

### DIFF
--- a/pkg/compose/watch.go
+++ b/pkg/compose/watch.go
@@ -220,7 +220,7 @@ func (s *composeService) watch(ctx context.Context, project *types.Project, opti
 			return nil, err
 		}
 
-		triggerPaths, err := s.watchTriggerPaths(ctx, project, service, config, syncer)
+		triggerPaths, err := s.watchTriggerPaths(ctx, service, config, syncer)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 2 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The changes are correct and clean. This PR fixes a genuine bug ([#13725](https://github.com/docker/compose/issues/13725)) where `initial_sync` silently skipped pre-existing host files whose mtime predated the image creation time — exactly the files it was designed to copy.

**What was removed and why it's safe:**
- `imageCreatedTime` / `ContainerList` + `ImageInspect` lookup: was a silent failure point (empty container list returned an error, blocking sync entirely). Correctly eliminated.
- `mtime` filter in `initialSyncDirectory` (the `d.Info()` + `ModTime().Before()` block): was the root-cause of #13725. Correctly removed.
- `mtime` guard on the regular-file path in `initialSyncFiles`: same fix applied symmetrically.

**Remaining ignore/exclusion logic is intact:** dockerignore patterns, ephemeral paths, `.git/`, trigger-level ignore lists, and bind-mount checks all still apply — only the mtime filter is gone.

**Tests:** Two new unit tests (`TestInitialSyncFilesDirectory`, `TestInitialSyncFilesRegularFile`) back-date files with `os.Chtimes` and verify they are now included, directly covering the regression.

ndeloof (approved):

PR #14102, opened 2026-08-20
ci: validate generated mocks are in sync

ndeloof:
`pkg/mocks` was only kept in sync by discipline: a change to a mocked interface with no regeneration went unnoticed until a test broke. This adds a `mocks-validate` CI check on the vendor-validate model (Dockerfile stage + bake target + `make validate-mocks` + validate matrix entry), and fixes the stale `mocks` target on the way: the mockgen invocation still passed the pre-rename `Service` argument and used whatever mockgen was on PATH; it now runs a pinned version via `go run` (v0.4.0 — package mode introduced in v0.5 copies moby's `_`-named variadic parameters into invalid Go code).

Verified locally: `make validate-mocks` is green on the branch, and fails with `ERROR: Generated mocks differ` on an artificial drift.

Part of #14074 (section G).


glours (approved):
LGTM

PR #14105, opened 2026-08-20
fix: dependency wait timeout surfaced only by luck

ndeloof:
`waitDependency` returned `nil` whenever its context ended, so an expired wait deadline surfaced only when the 500ms poll ticker won the select race and the condition check then failed on the expired context — whether `up --wait --wait-timeout` failed or silently succeeded after the timeout was effectively a coin toss.

The deadline now returns its error deterministically; a plain user cancellation (Ctrl-C) is still not reported as a dependency failure. Unit test pins both sides (a deadline shorter than the first tick used to be swallowed with certainty); a new e2e scenario documents the user-visible contract (`application not healthy after 3s`, non-zero exit, container left running).

Part of #14074 (section C) — lot 0 of #14081.

### DIFF
--- a/pkg/compose/convergence.go
+++ b/pkg/compose/convergence.go
@@ -198,6 +198,11 @@ func (s *composeService) waitDependency(ctx context.Context, dep string, config 
 		select {
 		case <-ticker.C:
 		case <-ctx.Done():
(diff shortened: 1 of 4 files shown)


glours (approved):
Nice catch 👍

PR #14112, opened 2026-08-20
fix(build): honor provenance/sbom false in per-service bake attest

### LINKED ISSUE #14111: build.provenance: false is ignored with the containerd image store; up --build recreates unchanged containers
### Description

With the containerd image store enabled, `build.provenance: false` in the compose file does not stop the default provenance attestation from being attached. The attestation carries build timestamps and a unique invocation id, so every `docker compose up --build` produces a new manifest-list digest for an unchanged build, and Compose recreates the container even though nothing changed.

Setting `BUILDX_NO_DEFAULT_ATTESTATIONS=1` in the environment suppresses the attestation and the container is left running, so the compose-file setting appears to not reach BuildKit in this path.

#13146 describes the same behavior and is closed as completed, but the issue is still reproducible on the versions below.

### Steps To Reproduce

`Dockerfile`:

```dockerfile
FROM alpine:3.20
CMD ["sleep","infinity"]
```

`compose.yml`:

```yaml
services:
  app:
    build:
      context: .
      provenance: false
    init: true
```

1. `docker compose up --build -d` → container Created, Started.
2. `docker compose up --build -d` again, no file changed → container **Recreated** (new container id).
3. `BUILDX_NO_DEFAULT_ATTESTATIONS=1 docker compose up --build -d` twice → second run prints `Running` and keeps the same container id.

### Compose Version

```text
Docker Compose version v5.1.1
```

### Docker Environment

```text
Docker Engine 29.3.1 (Docker Desktop, macOS)
Image store: containerd (io.containerd.snapshotter.v1, overlayfs)
```

### Anything else?

Expected: with `build.provenance: false`, an unchanged `up --build` leaves the container running, the same as with `BUILDX_NO_DEFAULT_ATTESTATIONS=1`.

glours:
**What I did**
When a service set `build.provenance: false` (or `sbom: false`), the bake attest entry was silently omitted (the field is `omitempty`), so BuildKit applied its defaults and attached a provenance attestation anyway. The attestation changes the manifest-list digest on every build, which caused containers to be needlessly recreated on `up --build`.

Emit `"type=<t>,disabled=true"` instead, which BuildKit's `Attest.UnmarshalText` parses as `Disabled:true` and forwards to BuildKit as an explicit disable signal.

Extract `appendAttest` to avoid duplicating the three-branch switch for both provenance and sbom.

**Related issue**

Fixes #14111

### DIFF
--- a/pkg/compose/build_bake.go
+++ b/pkg/compose/build_bake.go
@@ -599,25 +599,21 @@ func toBakeSecrets(project *types.Project, secrets []types.ServiceSecretConfig) 
 
 func toBakeAttest(buildConfig types.BuildConfig) []string {
 	var attests []string
+	attests = appendAttest(attests, "provenance", buildConfig.Provenance)
+	attests = appendAttest(attests, "sbom", buildConfig.SBOM)
+	return attests
+}
 
-	// Handle per-service provenance configuration (only from build config, not global options)
-	if buildConfig.Provenance != "" {
-		if buildConfig.Provenance == "true" {
-			attests = append(attests, "type=provenance")
-		} else if buildConfig.Provenance != "false" {
-			attests = append(attests, fmt.Sprintf("type=provenance,%s", buildConfig.Provenance))
-		}
+func appendAttest(attests []string, attestType, value string) []string {
+	switch value {
+	case "":
+	case "true":
+		attests = append(attests, "type="+attestType)
+	case "false":
+		attests = append(attests, "type="+attestType+",disabled=true")
+	default:
+		attests = append(attests, fmt.Sprintf("type=%s,%s", attestType, value))
 	}
-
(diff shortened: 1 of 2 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The refactor is correct and complete. The new `appendAttest` helper cleanly handles all four value branches (`""`, `"true"`, `"false"`, and user-supplied strings), and the `"type=<t>,disabled=true"` format is the correct signal for BuildKit's `Attest.UnmarshalText` to set `Disabled: true`. No behavioral regressions from the extraction — both provenance and sbom code paths are now symmetric and the test suite covers the key cases including the previously-broken `false` branch.

ndeloof (approved):

PR #14115, opened 2026-08-22
Close the ttyWriter done channel after context cancel

### LINKED ISSUE #14114: [BUG] ttyWriter.Done still hangs forever when the context was canceled before Done runs
### Description

The #13639 fix (25b29d77) makes `ttyWriter.Done` send on the unbuffered `done` channel before taking the render mutex, which fixes the completion-path deadlock. A second wedge path survives it.

The render goroutine started by `ttyWriter.Start` has three exits: `ctx.Done()`, receiving from `w.done`, and the 100ms tick. The `done` channel has exactly two users in the codebase: that receive, and the send in `Done`. `AdaptCmd` cancels the command context on the first SIGTERM/SIGINT. When that happens, the render goroutine exits through the `ctx.Done()` branch and never receives. `pkg/compose.Run` then calls `bus.Done` unconditionally, and its send blocks forever: the only receiver is gone. There is also a narrower window when the cancel and the send race: the select can pick the `ctx.Done()` branch while the sender is already blocked.

Interactive use masks this: the signal handler unregisters itself after the first signal, so a second Ctrl-C kills the process by default disposition. An unattended caller that sends a single TERM to a TTY-attached compose command gets a process that never exits.

Expected: `Done` returns after a canceled operation and the process exits with status 130.

### Steps To Reproduce

The mechanism reduces to a deterministic unit test. On current main (f729794), add this to `cmd/display/tty_cancel_test.go`:

```go
// A single SIGTERM/SIGINT cancels the command context (AdaptCmd). The render
// goroutine started by Start then exits through its ctx.Done() branch and never
// receives from the unbuffered done channel. pkg/compose.Run still calls
// bus.Done unconditionally afterward, and its send blocks forever: the only
// receiver is gone.
func TestDoneAfterContextCancelDoesNotHang(t *testing.T) {
	w, _ := newTestWriter()
	ctx, cancel := context.WithCancel(context.Background())
	w.Start(ctx, "down")
	cancel()
	// Let the render goroutine observe the cancellation and exit.
	time.Sleep(100 * time.Millisecond)

	finished := make(chan struct{})
	go func() {
		w.Done("down", false)
		close(finished)
	}()
	select {
	case <-finished:
	case <-time.After(2 * time.Second):
		t.Fatal("ttyWriter.Done blocked forever after context cancellation")
	}
}
```

```
--- FAIL: TestDoneAfterContextCancelDoesNotHang (2.10s)
    tty_cancel_test.go:46: ttyWriter.Done blocked forever after context cancellation
FAIL	github.com/docker/compose/v5/cmd/display
```

Operationally this corresponds to: `docker compose down` on a TTY with a slow-stopping service, one SIGTERM while the progress UI renders, and the process never exits. We found it while root-causing #13639 on an embedded device fleet, where we captured on-target goroutine dumps of that deadlock.

### Compose Version

```Text
Built from main at e8c21434 plus a backport of 25b29d77; the involved code paths are unchanged at f729794.
```

### Docker Environment

```Text
Embedded Linux devices, docker 29.x; the analysis is environment-independent.
```

### Anything else?

Replacing the send in `Done` with `close(w.done)` fixes it: close never blocks, and the select exits on the closed channel. With that one-line change, the test above and the whole `cmd/display` suite pass, race detector included.

Hashim1999164:
## Summary
Fixes #14114

`ttyWriter.Done` sent on an unbuffered channel. After SIGTERM the render goroutine exits through `ctx.Done()` and never receives, so `Done` blocked forever. Closing the channel returns immediately even when that goroutine is already gone.

## Test plan
* `go test ./cmd/display/`
* `TestDoneAfterContextCancelDoesNotHang` covers cancel then `Done`

### DIFF
--- a/cmd/display/tty.go
+++ b/cmd/display/tty.go
@@ -172,7 +172,8 @@ func (w *ttyWriter) Start(ctx context.Context, operation string) {
 
 func (w *ttyWriter) Done(operation string, success bool) {
 	w.print()
-	w.done <- true
+	// close never blocks if the render goroutine already exited via ctx.Done()
+	close(w.done)
 	w.mtx.Lock()
 	defer w.mtx.Unlock()
 	if w.ticker != nil {
--- a/cmd/display/tty_test.go
+++ b/cmd/display/tty_test.go
@@ -569,6 +569,26 @@ func TestDoneDeadlockFix(t *testing.T) {
 	}
 }
 
+func TestDoneAfterContextCancelDoesNotHang(t *testing.T) {
+	w, _ := newTestWriter()
+	ctx, cancel := context.WithCancel(context.Background())
+	w.Start(ctx, "down")
+	cancel()
+	// Let the render goroutine observe the cancellation and exit.
+	time.Sleep(100 * time.Millisecond)
+
+	finished := make(chan struct{})
+	go func() {
+		w.Done("down", false)
+		close(finished)
+	}()
+	select {
+	case <-finished:
+	case <-time.After(2 * time.Second):
+		t.Fatal("ttyWriter.Done blocked forever after context cancellation")
+	}
+}
+
 // TestAdjustLineWidth_WideProgressForcesSizeInfoDrop is the unit-level
 // regression test for docker/compose#13595. When progress contains the
 // " X.XMB / Y.YMB" size suffix and the bar makes beforeStatus large enough


glours:
Thanks for opening this — the underlying bug (`ttyWriter.Done` hanging forever after the command context is cancelled) is real, and we're addressing it. This particular fix isn't safe to merge though.                                              
                                                                                                                                                                                                                                                        
`close(w.done)` never blocks, which does fix the reported hang, but it panics with `close of closed channel` the moment the shared `ttyWriter` bus goes through more than one `Start`/`Done` cycle in the same process, and that already happens     
 
today:                                                                                                                                                                                                                                                
                                                                                                                                                                                                                                                        
  - `docker compose run` nests a `create`/`start` cycle inside its own `run` cycle (`startDependencies` calling the public `Create`/`Start`).                                                                                                           
  - `docker compose publish` nests a `push` cycle inside its own `publish` cycle.                                                                                                                                                                       
  - `docker compose rm --stop` runs a `stop` cycle sequentially, then its own `remove` cycle.                                                                                                                                                           
                                                                                                                                                                                                                                                        
We reproduced the panic against this PR's diff on all three paths. A correct fix needs to avoid opening a second bracket on the shared bus in the first place (the unexported methods should be called directly there, as `up`/`scale` already do), and make the channel itself safe to close more than once for the paths that legitimately run sequential cycles. 
#14119 fixes this properly, closing in favor of that one.                                                                            
                                                                                                                                                                                                                                                        
Separately: this account's recent activity, near-simultaneous, template-identical PRs opened across dozens of unrelated, high-profile repositories over the last few days, matches the pattern of low-effort, automated AI-generated contribution farming. We don't accept that kind of submission on docker/compose.

Hashim1999164:
Okay @glours

PR #14117, opened 2026-08-23
fix(watch): exclude Dockerfile and compose files from initial sync

htoyoda18:
**What I did**

`initialSync`'s doc comment promises that the Dockerfile and compose files are never copied into the container during `docker compose watch`'s initial sync — but the matcher enforcing that was dropped without replacement in a January 2025 refactor ([#12469](https://github.com/docker/compose/pull/12469)), leaving only a `// FIXME .dockerignore` comment where it used to sit. Neither `.dockerignore`-derived ignores nor `EphemeralPathMatcher()` (editor swap/temp files) cover this, so a `develop.watch` trigger whose path includes the project root would copy `Dockerfile`/`compose.yaml` into the running container on initial sync.

Reintroduced a dedicated `dockerFileIgnore` matcher (`Dockerfile`, `*compose*.y*ml`) inside `initialSync`, matching the pre-refactor behavior.

Scoped to `initialSync` only: the continuous watch loop's ignore (`getWatchRules`) never carried this exclusion, and — because it matches against full paths rather than basenames — would need a differently-anchored matcher to work there; that's a separate change if wanted.

**Related issue**
N/A

### DIFF
--- a/pkg/compose/watch.go
+++ b/pkg/compose/watch.go
@@ -784,11 +784,19 @@ func (s *composeService) initialSync(ctx context.Context, service types.ServiceC
 	if err != nil {
 		return err
 	}
-	// FIXME .dockerignore
+
+	// the Dockerfile and compose files drive the build/orchestration, not the
+	// application: never copy them into the container on initial sync
+	dockerFileIgnore, err := watch.NewDockerPatternMatcher("/", []string{"Dockerfile", "*compose*.y*ml"})
+	if err != nil {
+		return err
+	}
+
 	ignoreInitialSync := watch.NewCompositeMatcher(
 		dockerIgnores,
 		watch.EphemeralPathMatcher(),
 		dotGitIgnore,
+		dockerFileIgnore,
 		triggerIgnore)
 
 	pathsToCopy, err := s.initialSyncFiles(service, trigger, ignoreInitialSync)
--- a/pkg/compose/watch.go
+++ b/pkg/compose/watch.go
@@ -29,6 +29,7 @@ import (
 	gsync "sync"
 	"time"
 
+	"github.com/compose-spec/compose-go/v2/cli"
 	"github.com/compose-spec/compose-go/v2/types"
 	"github.com/compose-spec/compose-go/v2/utils"
 	ccli "github.com/docker/cli/cli/command/container"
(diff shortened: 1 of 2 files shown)


pkg/compose/watch.go
```diff
@@ -784,11 +784,19 @@ func (s *composeService) initialSync(ctx context.Context, service types.ServiceC
 	if err != nil {
 		return err
 	}
-	// FIXME .dockerignore
+
+	// the Dockerfile and compose files drive the build/orchestration, not the
+	// application: never copy them into the container on initial sync
+	dockerFileIgnore, err := watch.NewDockerPatternMatcher("/", []string{"Dockerfile", "*compose*.y*ml"})
```
glours: `*compose*.y*ml` matches any basename containing "compose" plus a `.y…ml` suffix (e.g. `my-compose-notes.yaml`, `recompose.yml`), silently dropping unrelated files from the sync. Consider matching literal names from: 
`compose-go/v2/cli.DefaultFileNames`/`DefaultOverrideFileNames`.

pkg/compose/watch.go
```diff
@@ -784,11 +784,19 @@ func (s *composeService) initialSync(ctx context.Context, service types.ServiceC
 	if err != nil {
 		return err
 	}
-	// FIXME .dockerignore
+
+	// the Dockerfile and compose files drive the build/orchestration, not the
+	// application: never copy them into the container on initial sync
```
glours: This restates the existing doc comment on `initialSync` a few lines above; consider dropping it or shortening it to just the new detail.

pkg/compose/watch_test.go
```diff
@@ -241,6 +241,33 @@ func TestInitialSyncFilesRegularFile(t *testing.T) {
 	}})
 }
 
+// initialSync's doc comment promises the Dockerfile and compose files are
+// never copied into the container, but a refactor (ed10804e0) dropped the
+// matcher enforcing it without replacement — neither the .dockerignore-derived
+// matcher nor EphemeralPathMatcher cover this.
+func TestInitialSync_ExcludesDockerfileAndComposeFiles(t *testing.T) {
+	hostDir := t.TempDir()
+	for _, name := range []string{"Dockerfile", "compose.yaml", "compose.override.yml", "app.go"} {
```
glours: `compose.yaml` and `compose.override.yml` both hit the same glob branch with an empty prefix, so the second file adds no coverage. 
Swap it for `docker-compose.yml` instead, it exercises the wildcard matching a non-empty prefix, which thecurrent fixtures don't test.

pkg/compose/watch.go
glours: This only excludes the literal `Dockerfile`, so a custom-named one (`build.dockerfile: Dockerfile.prod`) still gets copied into the container,  the same bug this PR fixes, just for non-default names. 
Add `service.Build.Dockerfile` (nil-checked)  to the pattern list.

glours (changes requested):
Hey @htoyoda18 
Thanks for this contribution, can you address the few comments I made so we can merge it?

htoyoda18:
@glours 
Thanks for the review! Addressed all four points:

- Literal names from `cli.DefaultFileNames`/`DefaultOverrideFileNames`
  instead of the `*compose*.y*ml` glob.
- Trimmed the comment duplicating `initialSync`'s doc comment.
- Added `docker-compose.yml` to the test fixtures to cover the
  non-empty-prefix wildcard case.
- Excluded `service.Build.Dockerfile` from the sync, with a test.

While testing the last one, I found it didn't cover a Dockerfile in a
subdirectory (`docker/Dockerfile.prod`) — the matcher only ever compares
basenames. Fixed with `filepath.Base(service.Build.Dockerfile)` and added
a regression test.

glours (approved):
Thanks for your contribution @htoyoda18

PR #14119, opened 2026-08-24
fix(display): stop ttyWriter.Done from hanging after context cancel

### LINKED ISSUE #14114: [BUG] ttyWriter.Done still hangs forever when the context was canceled before Done runs

glours:
**What I did**
Fixes #14114: Done() blocked forever once the render goroutine had already exited via ctx.Done() (e.g. after SIGTERM), since it sent on an unbuffered channel with nobody left to receive.

Closing the channel instead of sending never blocks, but a bare close() panics the moment ttyWriter runs a second Start/Done cycle on the same shared bus - which already happens today (run's nested create/start, publish's nested push, rm --stop's sequential stop then remove). Start now hands out a fresh, once-only-closable signal each cycle instead of a single shared channel, and run/publish stop opening a redundant nested bracket around calls already inside one.

**Related issue**
closes #14114 
Supersedes #14115

image

### DIFF
--- a/cmd/display/tty.go
+++ b/cmd/display/tty.go
@@ -42,10 +42,27 @@ func Full(out io.Writer, info io.Writer, detached bool) api.EventProcessor {
 		out:      out,
 		info:     info,
 		tasks:    map[string]*task{},
-		done:     make(chan bool),
+		done:     newDoneSignal(),
 		mtx:      &sync.Mutex{},
 		detached: detached,
 	}
+}
+
+// doneSignal is a channel that's safe to close more than once: a shared bus
+// can go through more than one sequential Start/Done cycle in its lifetime
+// (e.g. `docker compose rm --stop` runs a "stop" cycle before its own
+// "remove" cycle), and a bare channel can only ever be closed once.
+type doneSignal struct {
+	ch   chan bool
+	once sync.Once
+}
+
+func newDoneSignal() *doneSignal {
+	return &doneSignal{ch: make(chan bool)}
+}
+
+func (d *doneSignal) close() {
+	d.once.Do(func() { close(d.ch) })
 }
 
 type ttyWriter struct {
@@ -54,7 +71,7 @@ type ttyWriter struct {
 	tasks     map[string]*task
 	repeated  bool
 	numLines  int
-	done      chan bool
+	done      *doneSignal // (re)created by Start for each Start/Done cycle
 	mtx       *sync.Mutex
 	dryRun    bool // FIXME(ndeloof) (re)implement support for dry-run
 	operation string
@@ -152,18 +169,36 @@ func (t *task) Completed() bool {
 }
 
 func (w *ttyWriter) Start(ctx context.Context, operation string) {
-	w.ticker = time.NewTicker(100 * time.Millisecond)
+	w.mtx.Lock()
+	// If a previous cycle is still open (Start called again before its
+	// matching Done, e.g. nested Start/Done misuse), close it out first so
+	// its render goroutine exits instead of leaking until ctx is done.
+	if w.done != nil {
+		w.done.close()
+	}
+	if w.ticker != nil {
+		w.ticker.Stop()
+	}
+	done := newDoneSignal()
+	ticker := time.NewTicker(100 * time.Millisecond)
+	w.done = done
+	w.ticker = ticker
 	w.operation = operation
+	w.mtx.Unlock()
+
+	// done and ticker are this cycle's own, captured locally: a later Start
+	// (a new Start/Done cycle on the same writer) reassigns w.done/w.ticker,
+	// but that never affects this goroutine's own wait.
 	go func() {
 		for {
 			select {
 			case <-ctx.Done():
 				// interrupted
-				w.ticker.Stop()
+				ticker.Stop()
 				return
-			case <-w.done:
+			case <-done.ch:
 				return
-			case <-w.ticker.C:
+			case <-ticker.C:
 				w.print()
 			}
 		}
@@ -172,11 +207,19 @@ func (w *ttyWriter) Start(ctx context.Context, operation string) {
 
 func (w *ttyWriter) Done(operation string, success bool) {
 	w.print()
-	w.done <- true
+	w.mtx.Lock()
+	done := w.done
+	ticker := w.ticker
+	w.mtx.Unlock()
+
+	// close never blocks, unlike a send, so this returns even if the render
+	// goroutine already exited via ctx.Done().
+	done.close()
+
 	w.mtx.Lock()
 	defer w.mtx.Unlock()
-	if w.ticker != nil {
-		w.ticker.Stop()
+	if ticker != nil {
+		ticker.Stop()
 	}
 	w.operation = ""
 }
--- a/cmd/display/tty_test.go
+++ b/cmd/display/tty_test.go
@@ -20,12 +20,14 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"io"
 	"strings"
 	"sync"
 	"testing"
 	"time"
 	"unicode/utf8"
 
+	"go.uber.org/goleak"
 	"gotest.tools/v3/assert"
 
 	"github.com/docker/compose/v5/pkg/api"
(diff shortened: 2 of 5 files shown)


cmd/display/tty.go
```diff
@@ -172,11 +197,19 @@ func (w *ttyWriter) Start(ctx context.Context, operation string) {
...
 	w.print()
-	w.done <- true
+	w.mtx.Lock()
+	done := w.done
+	ticker := w.ticker
+	w.mtx.Unlock()
+
+	// close never blocks, unlike a send, so this returns even if the render
+	// goroutine already exited via ctx.Done().
+	done.close()
```
docker-agent, line 217: **[medium] `Done()` will nil-panic if called before `Start()` — `Full()` no longer initialises `w.done`**

Before this PR `Full()` initialised `w.done` with `make(chan bool)`, so `Done()` could always dereference it safely. This PR removes that initialisation:

```go
// Full() now returns a ttyWriter with w.done == nil
return &ttyWriter{
    tasks: map[string]*task{},
    mtx:   &sync.Mutex{},
    ...
    // w.done is zero-value (*doneSignal)(nil)
}
```

`Done()` then does:

```go
w.mtx.Lock()
done := w.done   // nil if Start() was never called
w.mtx.Unlock()
done.close()     // nil pointer dereference → panic
```

The `Run()` wrapper always calls `Start()` before `Done()`, so the normal path is safe. But the `api.EventProcessor` interface is public, and the old code's defensive initialisation guarded against misuse and future code paths. Consider either re-adding the initialisation in `Full()`:

```go
done: newDoneSignal(),
```

or adding a nil guard in `Done()`:

```go
if done == nil {
    return
}
```

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 62/100 |

cmd/display/tty.go
```diff
@@ -152,18 +168,27 @@ func (t *task) Completed() bool {
...
+	ticker := time.NewTicker(100 * time.Millisecond)
+	w.done = done
+	w.ticker = ticker
 	w.operation = operation
+	w.mtx.Unlock()
+
+	// done and ticker are this cycle's own, captured locally: a later Start
+	// (a new Start/Done cycle on the same writer) reassigns w.done/w.ticker,
+	// but that never affects this goroutine's own wait.
 	go func() {
```
docker-agent, line 192: **[medium] Goroutine leak when `Start()` is called a second time before `Done()` — the first goroutine's `doneSignal` is never closed**

Each `Start()` call creates a fresh `doneSignal` and overwrites `w.done`. The goroutine launched by the first `Start()` captures its own local `done` (the first signal) and blocks on `done.ch`. When a second `Start()` runs:

- `w.done` → replaced with second signal
- First goroutine is still waiting on the first `done.ch`

When `Done()` eventually runs it snapshots `w.done` (the **second** signal) and closes it, signalling the second goroutine. The first goroutine's channel is never closed and the goroutine leaks until `ctx` is cancelled.

`TestNestedStartDoneDoesNotPanic` exercises exactly this pattern:

```go
w.Start(ctx, "outer")   // goroutine-1 waits on signal-1
w.Start(ctx, "inner")   // goroutine-2 waits on signal-2; w.done = signal-2
w.Done("inner", false)  // closes signal-2; goroutine-2 exits
w.Done("outer", false)  // closes signal-2 again (no-op); goroutine-1 leaks
```

The test passes because `sync.Once` prevents the double-close panic, but goroutine-1 is leaked. It only exits when `ctx` is cancelled (which `t.Context()` does after the test completes in Go 1.21+), masking the leak in tests.

If nested `Start/Done` is not a supported use case, a `panic` or assertion on re-entry would be clearer. If it is supported, the previous signal's channel must also be closed when `Start()` replaces it.

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 67/100 |

docker-agent (commented):

The core fix is correct and well-structured: replacing the blocking channel send with a `close()`-based `doneSignal` (using `sync.Once` for idempotency) solves the deadlock, the per-cycle fresh signal design prevents double-close panics, and the switch to unexported `create`/`start`/`push` in `run.go`/`publish.go` correctly avoids double-bracketing.

Two medium-severity issues were found in the changed code.

- [low] cmd/display/tty.go:199 — TOCTOU race: `w.done` can be replaced between `w.print()` and the mutex snapshot in `Done()` (confidence: 🟠 weak 52/100). Requires concurrent `Start()`+`Done()` on the same writer outside the sequential `Run()` wrapper — unlikely in practice but structurally present.

docker-agent (commented):
### Assessment: 🟢 APPROVE

The fix is well-designed and correct. The core changes — replacing the shared unbuffered `done` channel with a per-cycle `doneSignal` (wrapping `sync.Once`) and switching from a blocking send to a non-blocking `close()` — soundly fix the #14114 hang without introducing any new races or panics.

**Key design points verified:**
- `sync.Once` in `doneSignal.close()` prevents double-close panics across sequential and nested `Start/Done` cycles.
- `Start()` holds `w.mtx` while replacing `w.done`/`w.ticker` and captures them as local variables in the goroutine, so a later `Start()` reassigning the struct fields never affects the running goroutine's wait.
- `Done()` reads `w.done` and `w.ticker` under lock into locals before calling `close()`, so it always operates on the correct cycle's signal even if a concurrent `Start()` races in.
- Switching `run.go` and `publish.go` to the unexported `create`/`start`/`push` avoids the nested `Start/Done` bracket on the shared bus, with `strings.ToLower` preserved for parity with the public methods.
- The new tests (`TestDoneAfterContextCancelDoesNotHang`, `TestNestedStartDoneDoesNotPanic`, `TestSequentialStartDoneEachGetFreshChannel`, `TestDoneBeforeStartDoesNotPanic`) provide solid regression coverage, including goroutine-leak detection via `goleak.VerifyNone`.

ndeloof (approved):

PR #14123, opened 2026-08-25
remove buildx as a Go dependency

ndeloof:
buildx was imported for two narrow purposes, both now covered by docker/cli machinery compose already uses elsewhere:

- `storeutil.GetProxyConfig` → `configfile.ParseProxyConfig`, the same lookup `docker build` uses. Behavioral note: build args now also pick up `ALL_PROXY` and the lowercase proxy variants, matching `docker build`.
- the dry-run image resolver (`buildx/builder` + `imagetools`) → a `DistributionInspect` through the daemon with credentials from the CLI config file — the same path the real pull/push would take. The encoded-auth helper moves to `internal/registry`, shared with `pull.go`.

bake is unaffected: it already execs the buildx CLI plugin rather than importing it. Dropping the module removes 23 buildx packages from the binary and the whole transitive `aws-sdk-go-v2` footprint (−328 lines of go.sum).

Follows up on the intent recorded in `build_bake.go`: *"long term plan is to remove buildx as a dependency"*.


internal/registry/registry.go
```diff
@@ -42,3 +50,25 @@ func GetAuthConfigKey(indexName string) string {
...
+// AuthProvider provides registry credentials for a registry hostname, as
+// implemented by the docker CLI's configfile.
+type AuthProvider interface {
+	GetAuthConfig(registryHostname string) (clitypes.AuthConfig, error)
+}
+
+// EncodedAuth returns the credentials for the registry hosting the given
+// image reference, base64-encoded as expected by the Docker API's
+// X-Registry-Auth header.
+func EncodedAuth(ref reference.Named, cfg AuthProvider) (string, error) {
```
thaJeztah, line 68: This could probably use https://github.com/moby/moby/blob/78739e07a44c2448cf7704e54f836463633531c5/api/pkg/authconfig/authconfig.go#L14-L29

internal/registry/registry.go
```diff
@@ -59,16 +63,12 @@ type AuthProvider interface {
...
-		return "", err
-	}
-
-	buf, err := json.Marshal(authConfig)
+	auth, err := cfg.GetAuthConfig(reference.Domain(ref))
 	if err != nil {
 		return "", err
 	}
-	return base64.URLEncoding.EncodeToString(buf), nil
+	return authconfig.Encode(registrytypes.AuthConfig(auth))
```
thaJeztah: I would recommend not depending on the type being exactly identical (and to do a staight type-cast); better to be explicit and assign the fields; https://github.com/docker/cli/blob/f81816ef882769f880d63deaec6220e86327bf50/cli/command/registry/search.go#L102-L115

ndeloof: Done in 5a6baced9 — switched to explicit field assignment.

thaJeztah (dismissed):
LGTM

pkg/compose/compose.go
```diff
@@ -248,11 +247,18 @@ func (s *composeService) getContextInfo() api.ContextInfo {
 }
 
 // getProxyConfig returns the proxy config - either custom override or environment-based
-func (s *composeService) getProxyConfig() map[string]string {
+func (s *composeService) getProxyConfig() map[string]*string {
 	if s.proxyConfig != nil {
-		return s.proxyConfig
+		m := make(map[string]*string, len(s.proxyConfig))
+		for k, v := range s.proxyConfig {
```
thaJeztah, line 253: Ugh; no idea TBH why we picked a map with pointers for this in the CLI (otherwise this could've been a `maps.Clone` 😞

thaJeztah: Oh, or was this perhaps for the "`nil` means "take from current environment"?

glours (approved):
🥳 LGTM!!! SHIP!!!!

PR #14124, opened 2026-08-25
refactor: NewGraph stops mutating the project

ndeloof:
Building the dependency graph silently deleted unresolvable optional `depends_on` references from the caller's project — and since `DependsOn` is a map, even from projects sharing it. What later readers observed depended on how many graphs had been built before them.

The pruning becomes an explicit step (`withoutUnresolvedOptionalDependencies`), applied once where the project consumed by the engines is established — `create` and `start` — before the graph, the dependency waits and the `com.docker.compose.depends_on` label read the model. `NewGraph` now just skips the edge; required dangling references still flow to its existing errors (naming the profiles that would enable the dependency). No behavior change.

Epic #14081, Lot 0 step 4 — precondition for a canonical project object across plan phases. Uses `Project.WithoutUnresolvedOptionalDependencies` from compose-spec/compose-go#922 (merged); go.mod is bumped to the post-merge commit — to be re-pinned on the next compose-go release.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,7 +8,7 @@ require (
 	github.com/Microsoft/go-winio v0.6.3-0.20251027160822-ad3df93bed29
 	github.com/acarl005/stripansi v0.0.0-20180116102854-5a71ef0e047d
 	github.com/buger/goterm v1.0.4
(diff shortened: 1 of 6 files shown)


glours (approved):
LGTM

PR #14075, opened 2026-08-17
chore: make the code describe itself truthfully (epic #14074, part A)

ndeloof:
First installment of #14074: fixes the places where the codebase misdescribes its own structure — documentation, comments, file names and error messages that point at code that no longer exists or promise behavior that isn't there. One commit per item of section A:

1. `convergence.go` → `service_containers.go`: the convergence engine the file was named after was removed; its doc comment sat on the unrelated `getScale()`.
2. Eleven comments in the reconciler justified decisions by referencing deleted code ('matching the previous ensureNetwork behavior', stale line numbers); each is rewritten as a self-contained behavioral statement — including that starting containers is not the plan's job, and that `sortContainers`' ordering *is* the scale-down policy.
3. `docs/sdk.md` documented a nonexistent `progress` package; it now points at `api.EventProcessor` and the real renderers in `cmd/display`.
4. The e2e suite is now genuinely gated behind the `e2e` build tag AGENTS.md already documented: `go test ./pkg/...` no longer launches e2e (previously it required a daemon + built binary for up to 20 minutes), and `-tags e2e` stops being a no-op. Makefile targets pass the tag; behavior in CI is unchanged.
5. The classic-builder rejections advertised `DOCKER_BUILDKIT=1`, a remedy tied to the internal BuildKit builder removed long ago; they now name the actual requirement (buildx plugin installed, `DOCKER_BUILDKIT` not forced to 0), and the silent buildx-missing fallback warning states its consequences.
6. AGENTS.md gains a ~35-line architecture map of the facts local reading cannot reveal (two lifecycle engines, `Run()`/events convention, stdout/stderr split, env-resolution idioms, config-hash membership invariant, Desktop integration points).

No behavior change except the wording of five build error messages (e2e assertions updated accordingly) and one log warning.

Part of #14074.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #14129, opened 2026-08-27
chore(compose): rename convergence.go, whose convergence engine no longer exists

ndeloof:
The `convergence` type was removed long ago but the file kept its name, and its doc comment ("convergence manages service's container lifecycle") sat on top of the unrelated `getScale()`. Renamed to `service_containers.go` to match its actual content (naming, dependency waiting, container creation, start helpers).


docker-agent (commented):
### Assessment: 🟢 APPROVE

Pure rename/refactor — no logic changes. The updated filename reference in `create.go`, the new file-level doc comment, and the accurate `getScale()` doc comment all look correct. No issues found.

glours (commented):
The convergence.go → service_containers.go rename is missing 3 references to the old filename

- pkg/compose/reconcile.go:719 — comment // Nothing to do (exited containers are left as-is, matching convergence.go behavior)
- pkg/compose/reconcile.go:1012 — comment // sortContainers sorts containers the same way as convergence.go:138-160:
- pkg/compose/reconcile_test.go:1245 — comment // Exited containers are left as-is, matching convergence.go:199 behavior

Since the whole point of this PR (epic #14074, section A) is to fix misleading pointers to this file, it'd be worth folding these into the same change rather than leaving them for a follow-up.

ndeloof:
Good catch — folded into the rename commit (1bc908e43): the three references now point at `service_containers.go` (and the test comment is rewritten as a behavioral statement, dropping the stale `:199` line number).

Note on the two `reconcile.go` sites: #14130 (the comments-state-behavior sibling split from the same #14075) rewrites both comments entirely as self-contained behavioral statements — whichever merges second will need a trivial rebase, I'll handle it.

glours (approved):
LGTM

PR #14130, opened 2026-08-27
chore(compose): comments state behavior, not history

ndeloof:
Eleven comments in the reconciler justified decisions by referencing deleted code ("matching the previous ensureNetwork behavior", stale line numbers to code that moved). Each is rewritten as a self-contained behavioral statement — including that starting containers is not the plan's job, and that `sortContainers`' ordering *is* the scale-down policy.

### DIFF
--- a/pkg/compose/create.go
+++ b/pkg/compose/create.go
@@ -124,8 +124,8 @@ func (s *composeService) create(ctx context.Context, project *types.Project, opt
 		return err
 	}
 
-	// Emit "Running" events for containers that are already up-to-date,
-	// matching the previous convergence behavior for progress display.
+	// Emit "Running" events for containers that are already up-to-date, so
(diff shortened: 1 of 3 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

All eleven comment rewrites accurately describe the behavior of the code they annotate. No logic changes, no misleading descriptions, and no omitted critical caveats were found across `create.go`, `observed_state.go`, and `reconcile.go`.

glours (dismissed):
Nit: `pkg/compose/reconcile_test.go:1245` still has `// Exited containers are left as-is, matching convergence.go:199 behavior` same stale-pointer pattern this PR fixes elsewhere, just left out of scope here (probably because it's a test file rather than "the reconciler" itself). Not blocking, just flagging for a follow-up pass.

glours (approved):

PR #14132, opened 2026-08-27
test(e2e): gate the e2e suite behind the build tag the docs already promise

ndeloof:
`go test ./pkg/...` used to launch the whole e2e suite (requiring a daemon and a built binary for up to 20 minutes) because no file in `pkg/e2e` carried the `e2e` build tag AGENTS.md documents — `-tags e2e` was a no-op. Every test file is now gated; non-test helpers stay untagged so the package always compiles. Makefile targets pass the tag; CI behavior is unchanged.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -4,9 +4,11 @@
 
 - Build: `make build`
 - Test all: `make test`
-- Test unit: `go test ./pkg/...`
+- Test unit: `go test ./pkg/...` — needs no Docker daemon; the e2e suite is
+  gated behind the `e2e` build tag and is not picked up
 - Test single: `go test ./pkg/compose/ -run TestFunctionName`
-- E2E tests: `go test -tags e2e ./pkg/e2e/ -run TestName`
+- E2E tests: `go test -tags e2e ./pkg/e2e/ -run TestName` — requires a Docker
+  daemon and the locally built binary (`make build`)
 
 ## E2E tests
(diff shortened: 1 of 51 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

All 51 changed files were reviewed. The PR mechanically adds `//go:build e2e` to every `_test.go` file in `pkg/e2e/` and updates the two Makefile targets (`e2e-compose` and `e2e-compose-standalone`) to pass the tag at test-run time.

**Makefile changes are correct:**
- `e2e-compose`: adds `-tags e2e` — tests now compile and run only when the tag is present.
- `e2e-compose-standalone`: changes `--tags=standalone` → `--tags=e2e,standalone` — both tags are required for that mode, correctly ordered.

**CI is unaffected:** `.github/workflows/ci.yml` calls `make e2e-compose` and `make e2e-compose-standalone`, which are both updated. No direct `go test` invocations of `pkg/e2e` exist in CI that would need updating.

**Non-test helpers are untagged by design:** `framework.go`, `scenario.go`, `assert.go`, `checks.go` carry no build tag so the package always compiles; `e2e_config_plugin.go` (`!standalone`) and `e2e_config_standalone.go` (`standalone`) keep their existing mode-selection logic untouched.

No bugs found.

glours (approved):
LGTM

PR #14139, opened 2026-08-27
fix: COMPOSE_REMOVE_ORPHANS resolves identically for every command carrying the flag

ndeloof:
Epic #14074, D.1 (and G.3). `COMPOSE_REMOVE_ORPHANS` had three resolution mechanisms: `up` read it after the project's `.env` was injected into the process environment; `down`/`kill` read it at cobra-tree construction time — before that injection, so a value in the project's `.env` worked for `up` but not for them; `create`/`run` never read it.

All five commands now resolve it identically from `PreRunE` through one helper: explicit flag > process env > local `.env`. `create` also gains `COMPOSE_IGNORE_ORPHANS` handling (its `ignoreOrphans` field was sent to the API without ever being assigned), with `up`'s conflict guard.

Behavioral notes:
- `down`/`kill` now honor the variable from the project's `.env`; the shell-exported case behaves as before (the `--help` output no longer reflects it as the flag default, since resolution moved out of the flag definition).
- the remote-config (OCI/Git) exclusion is kept and now documented as a product decision on `setEnvWithDotEnv`: COMPOSE_* variables are the local user's choice; a remote model must not steer the CLI consuming it.
- unit tests pin the `.env` → process env → flag resolution order and the remote exclusion.

### DIFF
--- a/cmd/compose/run.go
+++ b/cmd/compose/run.go
@@ -195,6 +195,7 @@ func runCommand(p *ProjectOptions, dockerCli command.Cli, backendOptions *Backen
 				display.Mode = display.ModeQuiet
 				backendOptions.Add(compose.WithEventProcessor(display.Quiet()))
 			}
+			options.removeOrphans = removeOrphansFromEnv(cmd.Flags(), options.removeOrphans)
 			createOpts.pullChanged = cmd.Flags().Changed("pull")
 			return nil
 		}),
--- a/cmd/compose/run.go
+++ b/cmd/compose/run.go
@@ -214,6 +214,13 @@ func runCommand(p *ProjectOptions, dockerCli command.Cli, backendOptions *Backen
 				buildOpts.Progress = string(xprogress.QuietMode)
 			}
 
+			// Deliberate source asymmetry with removeOrphans: the destructive
+			// variable (COMPOSE_REMOVE_ORPHANS) resolves through the process
+			// environment, which setEnvWithDotEnv completes from the local .env
+			// only — a remote model cannot enable container removal. The benign
+			// COMPOSE_IGNORE_ORPHANS reads project.Environment, remote configs
+			// included: the worst a remote model can do there is suppress a
+			// warning.
 			options.ignoreOrphans = utils.StringToBool(project.Environment[ComposeIgnoreOrphans])
 			if options.ignoreOrphans && options.removeOrphans {
 				return fmt.Errorf("cannot combine %s and --remove-orphans", ComposeIgnoreOrphans)
--- a/cmd/compose/run.go
+++ b/cmd/compose/run.go
@@ -215,6 +215,9 @@ func runCommand(p *ProjectOptions, dockerCli command.Cli, backendOptions *Backen
 			}
 
 			options.ignoreOrphans = utils.StringToBool(project.Environment[ComposeIgnoreOrphans])
+			if options.ignoreOrphans && options.removeOrphans {
+				return fmt.Errorf("cannot combine %s and --remove-orphans", ComposeIgnoreOrphans)
+			}
 			return runRun(ctx, backend, project, options, createOpts, buildOpts, dockerCli)
 		}),
 		ValidArgsFunction: completeServiceNames(dockerCli, p),
--- a/cmd/compose/compose.go
+++ b/cmd/compose/compose.go
@@ -73,6 +73,21 @@ const (
 )
 
 // rawEnv load a dot env file using docker/cli key=value parser, without attempt to interpolate or evaluate values
+// removeOrphansFromEnv resolves the effective --remove-orphans value for a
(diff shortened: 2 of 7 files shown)


cmd/compose/run.go
```diff
@@ -195,6 +195,7 @@ func runCommand(p *ProjectOptions, dockerCli command.Cli, backendOptions *Backen
 				display.Mode = display.ModeQuiet
 				backendOptions.Add(compose.WithEventProcessor(display.Quiet()))
 			}
+			options.removeOrphans = removeOrphansFromEnv(cmd.Flags(), options.removeOrphans)
```
docker-agent, line 198: **[medium] Missing `ignoreOrphans` vs `removeOrphans` conflict guard in `run` subcommand**

This PR correctly adds `removeOrphansFromEnv` to `run`'s `PreRunE` so `COMPOSE_REMOVE_ORPHANS` is now resolved for `run` — but the corresponding conflict guard that both `up` and `create` have was not added to `run`'s `RunE`.

In `up.go` (line 129) and `create.go` (line 103):
```go
if create.ignoreOrphans && create.removeOrphans {
    return fmt.Errorf("cannot combine %s and --remove-orphans", ComposeIgnoreOrphans)
}
```

In `run.go`'s `RunE`, after `options.ignoreOrphans = utils.StringToBool(project.Environment[ComposeIgnoreOrphans])` (line 217), there is no equivalent check. Now that `removeOrphans` is properly resolved for `run`, a user with both `COMPOSE_IGNORE_ORPHANS=true` in their project `.env` and `COMPOSE_REMOVE_ORPHANS=true` (or `--remove-orphans`) will get conflicting options silently accepted for `run`, while `up` and `create` would return an error.

**Fix**: Add the same guard in `run.go`'s `RunE` after `options.ignoreOrphans` is assigned:
```go
options.ignoreOrphans = utils.StringToBool(project.Environment[ComposeIgnoreOrphans])
if options.ignoreOrphans && options.removeOrphans {
    return fmt.Errorf("cannot combine %s and --remove-orphans", ComposeIgnoreOrphans)
}
```

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 97/100 |

docker-agent (commented):

- [medium→low] cmd/compose/compose.go:757 — Remote-config exclusion guard in `setEnvWithDotEnv` only works when `ConfigPaths` is pre-populated; silently skipped when paths are discovered dynamically (confidence: weak 52/100)
- [medium→low] cmd/compose/create.go:100 — Asymmetric env sources: `removeOrphans` reads from `os.Getenv` (process env) while `ignoreOrphans` reads from `project.Environment` (loader-resolved); conflict guard may behave differently for remote configs (confidence: weak 52/100)

cmd/compose/run.go
glours: Agreed and confirmed: run is the only command missing this guard; the fix above resolves it.

docker-agent: Thank you for confirming! Adding the conflict guard to `run` will ensure consistent validation across all commands that support orphan handling.

ndeloof: Done in ea0f04e1d — run now rejects the combination exactly like up and create.

ndeloof:
On the lower-confidence finding about the remote-config guard in `setEnvWithDotEnv` (`COMPOSE_FILE` discovering paths dynamically): the observation is technically right — `COMPOSE_FILE` is resolved later by compose-go (`cli.WithConfigFileEnv`), so `opts.ConfigPaths` only carries `--file` and the early return doesn't fire for `COMPOSE_FILE=oci://…`. It has no product impact though: `setEnvWithDotEnv` only ever reads the **local** `.env` (cwd / `--env-file`) — a remote artifact's environment cannot flow through it under any path. The early return exists so `NewProjectOptions` isn't asked to process a remote path string; with an empty `ConfigPaths` that call is harmless and what gets injected is the local user's own `.env`, which is exactly the product decision (COMPOSE_* variables are the local user's choice). The residual inconsistency — `-f oci://…` skips the local-.env injection while `COMPOSE_FILE=oci://…` applies it — is benign and worth harmonizing separately if desired.

cmd/compose/run.go
docker-agent: Perfect! Confirmed the guard is now in place at lines 218-220, matching the exact pattern from `up` and `create`. The conflict between `COMPOSE_IGNORE_ORPHANS` and `--remove-orphans` will now be properly caught for `run` just as it is for the other commands. Thank you for implementing this!

ndeloof:
Follow-up on the second lower-confidence finding (asymmetric env sources for `removeOrphans`/`ignoreOrphans`): the asymmetry is deliberate, and 0aa14bdb3 now states it at the three read sites — the destructive variable resolves through the process environment (local `.env` only, a remote model cannot enable container removal), the benign one through `project.Environment` (remote included, worst case a suppressed warning). Future reviewers won't have to rediscover the decision.

PR #14140, opened 2026-08-27
fix(logs): restart re-attach loses a fast run's output — found by deflaking e2e

ndeloof:
Three e2e tests failed 9 CI runs across 5 unrelated branches this week. Deflaking them (first commit) turned `TestAttachRestart` from a flake into a reliable detector — and its first CI run proved the loss is real: with the hardened test *waiting* a full minute, the third restart's log line still never arrived (`2× world` for `3× exit notices`, the re-attached `logs?since=…` request hanging empty).

**The bug** (second commit): re-attaching with `since=StartedAt` drops a fast run's first lines forever — the daemon starts copying stdout before it records `StartedAt`, so a process that prints and exits within milliseconds gets its output timestamped just before the anchor. Both re-attach sites (attached `up`, `logs --follow`) now anchor on the previous run's `FinishedAt`: nothing can be logged between one run's end and the next one's start, so the window captures the whole new run without replaying the previous one (zero `FinishedAt` ⇒ no lower bound, exact for a fresh container).

**The deflakes** (first commit):
- `RequireServiceState` polls until the daemon-reported state converges instead of asserting one `compose ps` snapshot (`TestUpDependenciesNotStopped`, ×5);
- `TestAttachRestart` awaits the log-line count like it already awaited exit notices — events monitor and re-attached logs stream are unordered channels; a genuinely lost line still fails, by timeout (that is how the bug above was caught);
- `TestUpExitCodeFromContainerKilled`'s 60s ceiling → 120s (one timeout on a loaded oldstable runner).

### DIFF
--- a/pkg/compose/logs.go
+++ b/pkg/compose/logs.go
@@ -19,6 +19,8 @@ package compose
 import (
 	"context"
 	"io"
+	"sync"
+	"time"
 
 	"github.com/containerd/errdefs"
 	"github.com/moby/moby/api/pkg/stdcopy"
@@ -110,19 +112,29 @@ func (s *composeService) logContainer(ctx context.Context, consumer api.LogConsu
 // while following, ignoring those whose logging driver doesn't support
 // reading logs
 func (s *composeService) followStartedContainersLogs(ctx context.Context, eg *errgroup.Group, consumer api.LogConsumer, options api.LogOptions) api.ContainerEventListener {
+	runEnds := newRunEndTracker()
 	return func(event api.ContainerEvent) {
+		runEnds.Observe(event)
 		if event.Type != api.ContainerEventStarted {
 			return
 		}
+		// Captured synchronously: the monitor delivers events in order, so
+		// the recorded end cannot yet include THIS run's own exit — reading
+		// it inside the goroutine below could (fast run), and the window
+		// would drop the whole run.
+		since := runEnds.Since(event.ID)
 		eg.Go(func() error {
 			res, err := s.apiClient().ContainerInspect(ctx, event.ID, client.ContainerInspectOptions{})
 			if err != nil {
 				return err
 			}
+			if since == "" {
+				since = logsSinceLastRun(res.Container)
+			}
 
 			err = s.doLogContainer(ctx, consumer, event.Source, res.Container, api.LogOptions{
 				Follow:     options.Follow,
-				Since:      res.Container.State.StartedAt,
+				Since:      since,
 				Until:      options.Until,
 				Tail:       options.Tail,
 				Timestamps: options.Timestamps,
@@ -134,6 +146,67 @@ func (s *composeService) followStartedContainersLogs(ctx context.Context, eg *er
 			return err
 		})
 	}
+}
+
+// runEndTracker remembers, per container, when the session last saw it exit —
+// the re-attach anchor that stays correct even when the NEW run is already
+// over: the event stream is ordered, so at start-event time the recorded
+// value is necessarily the PREVIOUS run's end. The inspected FinishedAt
+// (logsSinceLastRun) cannot give that guarantee — by the time we inspect, a
+// fast run's own FinishedAt has overwritten it and the window would exclude
+// everything the run printed.
+type runEndTracker struct {
+	mu   sync.Mutex
+	ends map[string]int64 // container ID → TimeNano of the last observed exit
+}
+
+func newRunEndTracker() *runEndTracker {
+	return &runEndTracker{ends: map[string]int64{}}
+}
+
+// Observe records exit events (other event types are ignored).
+func (t *runEndTracker) Observe(e api.ContainerEvent) {
+	if e.Type != api.ContainerEventExited || e.Time == 0 {
+		return
+	}
+	t.mu.Lock()
+	t.ends[e.ID] = e.Time
+	t.mu.Unlock()
+}
+
+// Since returns the log-window anchor for a container being re-attached: the
+// recorded end of its previous run in RFC3339Nano — the same format the
+// FinishedAt fallback feeds the logs API — or "" when the session never saw
+// it exit (first start).
+func (t *runEndTracker) Since(containerID string) string {
+	t.mu.Lock()
+	nano, ok := t.ends[containerID]
+	t.mu.Unlock()
+	if !ok {
+		return ""
+	}
+	return time.Unix(0, nano).UTC().Format(time.RFC3339Nano)
+}
+
+// logsSinceLastRun returns the FALLBACK log window anchor for a container
+// (re)started while we follow the project, used when the session has not
+// observed a previous exit (runEndTracker): the previous run's FinishedAt.
+// The new run's StartedAt looks like the natural anchor but loses output —
+// the daemon starts copying stdout before it records StartedAt, so a fast
+// process can get its first lines timestamped just before it, and
+// `since=StartedAt` then drops them forever. Nothing can be logged between
+// the previous run's end and the new run's start, so FinishedAt captures the
+// entire new run without replaying the previous one — UNLESS the new run
+// already finished by inspection time (its own FinishedAt shadows the
+// previous run's), which is exactly what the tracker protects against. A
+// container with no previous run has a zero FinishedAt, which means "no
+// lower bound" — equally exact for a fresh container.
+func logsSinceLastRun(ctr container.InspectResponse) string {
+	finished := ctr.State.FinishedAt
+	if t, err := time.Parse(time.RFC3339Nano, finished); err != nil || t.Unix() <= 0 {
+		return ""
+	}
+	return finished
 }
 
 func (s *composeService) doLogContainer(ctx context.Context, consumer api.LogConsumer, name string, ctr container.InspectResponse, options api.LogOptions) error {
--- a/pkg/compose/logs_test.go
+++ b/pkg/compose/logs_test.go
@@ -245,3 +245,34 @@ func (l *testLogConsumer) LogsForContainer(containerName string) []string {
 	defer l.mu.Unlock()
 	return l.logs[containerName]
 }
+
+// TestRunEndTrackerAnchorsOnPreviousRun pins the re-attach anchor against the
+// fast-run race seen in CI: with events delivered in order, the anchor
+// captured at start-event time is the PREVIOUS run's end — even when the new
+// run exits (and is observed) before the log stream is actually opened.
+func TestRunEndTrackerAnchorsOnPreviousRun(t *testing.T) {
+	tr := newRunEndTracker()
+
+	// First start: no previous exit observed → no anchor (caller falls back
+	// to the inspected FinishedAt).
+	assert.Equal(t, tr.Since("c1"), "")
+
+	// Run N exits at t=1_000_000_001ns, run N+1 starts: the anchor captured
+	// at start-event time is run N's end, nanosecond-precise.
+	tr.Observe(compose.ContainerEvent{Type: compose.ContainerEventExited, ID: "c1", Time: 1_000_000_001})
+	anchor := tr.Since("c1")
+	assert.Equal(t, anchor, "1970-01-01T00:00:01.000000001Z")
+
+	// Run N+1 is fast: its own exit is observed before the log stream opens.
+	// The anchor captured above must NOT move — reading it after this point
+	// would exclude everything run N+1 printed.
+	tr.Observe(compose.ContainerEvent{Type: compose.ContainerEventExited, ID: "c1", Time: 2_000_000_002})
+	assert.Equal(t, anchor, "1970-01-01T00:00:01.000000001Z")
+	// The NEXT start anchors on run N+1's end.
+	assert.Equal(t, tr.Since("c1"), "1970-01-01T00:00:02.000000002Z")
+
+	// Non-exit events and other containers do not pollute the anchor.
+	tr.Observe(compose.ContainerEvent{Type: compose.ContainerEventStarted, ID: "c1", Time: 9_000_000_000})
+	tr.Observe(compose.ContainerEvent{Type: compose.ContainerEventExited, ID: "c2", Time: 3_000_000_003})
+	assert.Equal(t, tr.Since("c1"), "1970-01-01T00:00:02.000000002Z")
+}
--- a/pkg/e2e/assert.go
+++ b/pkg/e2e/assert.go
@@ -18,26 +18,62 @@ package e2e
 
 import (
 	"encoding/json"
+	"fmt"
 	"strings"
 	"testing"
+	"time"
 
-	"gotest.tools/v3/assert"
-	is "gotest.tools/v3/assert/cmp"
+	"gotest.tools/v3/poll"
 )
 
-// RequireServiceState ensures that the container is in the expected state
-// (running or exited).
+// RequireServiceState ensures that the container reaches the expected state
+// (running or exited). The daemon reports state transitions asynchronously
+// from everything else a test can observe (a container whose logs already
+// flowed may still be listed under its previous state for a moment), so the
+// check polls `compose ps` until the state converges instead of asserting on
+// a single snapshot.
 func RequireServiceState(t testing.TB, cli *CLI, service string, state string) {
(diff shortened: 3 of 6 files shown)


pkg/e2e/assert.go
```diff
@@ -18,26 +18,35 @@ package e2e
...
+	poll.WaitOn(t, func(poll.LogT) poll.Result {
+		psRes := cli.RunDockerComposeCmd(t, "ps", "--all", "--format=json", service)
+		var serviceState map[string]any
+		if err := json.Unmarshal([]byte(psRes.Stdout()), &serviceState); err != nil {
+			return poll.Error(fmt.Errorf("invalid `compose ps` JSON: %w: command output: %s", err, psRes.Combined()))
+		}
+		if service != serviceState["Service"] {
+			return poll.Error(fmt.Errorf("found ps output for unexpected service %q", serviceState["Service"]))
+		}
+		if !strings.EqualFold(state, serviceState["State"].(string)) {
```
docker-agent: **[high] Unsafe type assertion `serviceState["State"].(string)` panics when key is absent or null**

The assertion on line 46 has no comma-ok guard:

```go
if !strings.EqualFold(state, serviceState["State"].(string)) {
```

If `compose ps --all --format=json <service>` returns an object where the `"State"` key is absent or its value is JSON `null` (both possible when the service isn't registered yet), then `serviceState["State"]` is `nil` (`interface{}`). The direct `.(string)` assertion panics at runtime. Because this is inside a `poll.WaitOn` closure, the panic propagates out and crashes the test goroutine rather than returning `poll.Continue` or `poll.Error`, defeating the entire purpose of the polling loop.

Use the comma-ok idiom instead:

```go
stateVal, ok := serviceState["State"].(string)
if !ok {
    return poll.Continue("state field missing or not a string for service %q", service)
}
if !strings.EqualFold(state, stateVal) {
    return poll.Continue("service %q (%s) is %q, expected %q",
        service, serviceState["Name"], stateVal, state)
}
```

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 100/100 |

pkg/e2e/assert.go
```diff
@@ -18,26 +18,35 @@ package e2e
...
-		service, serviceState["Name"],
-	)
+	poll.WaitOn(t, func(poll.LogT) poll.Result {
+		psRes := cli.RunDockerComposeCmd(t, "ps", "--all", "--format=json", service)
+		var serviceState map[string]any
+		if err := json.Unmarshal([]byte(psRes.Stdout()), &serviceState); err != nil {
+			return poll.Error(fmt.Errorf("invalid `compose ps` JSON: %w: command output: %s", err, psRes.Combined()))
+		}
+		if service != serviceState["Service"] {
+			return poll.Error(fmt.Errorf("found ps output for unexpected service %q", serviceState["Service"]))
```
docker-agent: **[medium] `poll.Error` on missing/unexpected service name stops polling on a transient startup condition**

Line 43 returns `poll.Error` when `serviceState["Service"]` doesn't match the expected service name:

```go
if service != serviceState["Service"] {
    return poll.Error(fmt.Errorf("found ps output for unexpected service %q", serviceState["Service"]))
}
```

`poll.Error` terminates the poll immediately (like a fatal assertion). However, during a startup race, `compose ps` can temporarily return `{}` (empty object) or an object without a `"Service"` key — in which case `serviceState["Service"]` is `nil`, and the comparison with the non-nil `service` string fails. This converts a transient "not yet visible" condition into a hard test failure, exactly the kind of flakiness this PR is trying to fix.

Consider using `poll.Continue` here so the poller keeps retrying until the service appears:

```suggestion
		return poll.Continue("service %q not yet visible in compose ps output (got %q)", service, serviceState["Service"])
```

(`poll.Error` is still appropriate for a genuinely wrong service name if you can distinguish the two cases — but given the async context, `poll.Continue` is safer.)

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 75/100 |

docker-agent (commented):
### Assessment: 🔴 CRITICAL

Two issues in the new polling closure introduced by this PR:

1. **Unsafe type assertion (panics)** — `serviceState["State"].(string)` at line 46 has no nil/type guard. If `compose ps` returns `{}` or `{"State": null}` (possible when a service is not yet registered), the assertion panics inside the `poll.WaitOn` closure and crashes the test goroutine instead of retrying.

2. **Wrong poll result on missing service** — Line 43 returns `poll.Error` (terminates polling) when the service key is absent/mismatched. During a startup race this converts a transient "not yet visible" condition into a hard failure — the same class of race this PR aims to fix.

ndeloof:
CI caught the loss again *with* the FinishedAt anchor (still 2× world for 3× exit notices after a full minute, oldstable runner) — so there is a second mechanism. Local experiments can't reproduce it: 15 bare-daemon runs of a millisecond-lived `--restart=on-failure:2` container always land 3 worlds in `docker logs`, and the e2e passes consistently on this machine with the fix.

Added instrumentation instead of speculation: on timeout the test now dumps the daemon's own `docker logs` view. Next CI occurrence will discriminate: line present there but absent from compose's output → compose still fails to relay; absent from both → the daemon's copier loses ultra-short-lived output (moby issue). The FinishedAt anchor stays — the StartedAt race it closes is real regardless.

pkg/e2e/assert.go
ndeloof: Fixed in 4a853e3c4 (squashed into the commit introducing the helper): both assertions now use the comma-ok form — an absent or null `State`/`Service` can no longer panic the poll closure; it degrades to a `poll.Continue` retry.

pkg/e2e/assert.go
ndeloof: Fixed in 4a853e3c4. Every transient shape now retries instead of hard-failing: empty `ps` output (container not registered yet), a foreign/incomplete entry, and — one more case in the same family — multi-line output during a recreate transition (old container being removed while its replacement is created): the output is parsed line by line and the poll succeeds as soon as one entry of the target service reaches the expected state. Only malformed JSON remains a hard `poll.Error`.

docker-agent: Perfect! The fix handles all the transient shapes elegantly: empty output, foreign/incomplete entries, and the multi-line recreate transition I hadn't considered. Treating only malformed JSON as `poll.Error` is exactly the right boundary — it distinguishes parse failures (unrecoverable) from all the async state variations (expected and transient).

The line-by-line parsing with early success on any matching entry is a nice touch for the recreate case. Thanks for the thorough fix!

pkg/compose/logs.go
```diff
@@ -136,6 +149,67 @@ func (s *composeService) followStartedContainersLogs(ctx context.Context, eg *er
...
+// to the inspected FinishedAt).
+func (t *runEndTracker) Since(containerID string) string {
+	t.mu.Lock()
+	nano, ok := t.ends[containerID]
+	t.mu.Unlock()
+	if !ok {
+		return ""
+	}
+	return fmt.Sprintf("%d.%09d", nano/1e9, nano%1e9)
+}
```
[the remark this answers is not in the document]
ndeloof: Done in 037947a65: `Since()` now returns RFC3339Nano (`time.Unix(0, nano).UTC().Format(time.RFC3339Nano)`) — same format as the FinishedAt fallback. The fractional-unix form is accepted by the daemon (`GetTimestamp` parses it), but consistency and debug readability win.

pkg/compose/logs_test.go
```diff
@@ -245,3 +245,34 @@ func (l *testLogConsumer) LogsForContainer(containerName string) []string {
...
+	// would exclude everything run N+1 printed.
+	tr.Observe(compose.ContainerEvent{Type: compose.ContainerEventExited, ID: "c1", Time: 2_000_000_002})
+	assert.Equal(t, anchor, "1.000000001")
+	// The NEXT start anchors on run N+1's end.
+	assert.Equal(t, tr.Since("c1"), "2.000000002")
+
+	// Non-exit events and other containers do not pollute the anchor.
+	tr.Observe(compose.ContainerEvent{Type: compose.ContainerEventStarted, ID: "c1", Time: 9_000_000_000})
+	tr.Observe(compose.ContainerEvent{Type: compose.ContainerEventExited, ID: "c2", Time: 3_000_000_003})
+	assert.Equal(t, tr.Since("c1"), "2.000000002")
```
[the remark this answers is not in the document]
ndeloof: Done in 037947a65 — expected strings updated to the RFC3339Nano form.

pkg/e2e/assert.go
```diff
@@ -18,26 +18,54 @@ package e2e
...
-		"Invalid `compose ps` JSON: command output: %s",
-		psRes.Combined())
-
-	assert.Assert(t, is.Equal(service, serviceState["Service"]), "Found ps output for unexpected service")
-	assert.Assert(t, is.Equal(strings.ToLower(state), strings.ToLower(serviceState["State"].(string))),
-		"Service %q (%s) not in expected state",
-		service, serviceState["Name"],
-	)
+	poll.WaitOn(t, func(poll.LogT) poll.Result {
+		psRes := cli.RunDockerComposeCmd(t, "ps", "--all", "--format=json", service)
```
docker-agent: **[medium] `RunDockerComposeCmd` calls `t.FailNow()` inside `poll.WaitOn` goroutine, obscuring real errors as timeout**

`poll.WaitOn` executes each check in a freshly spawned goroutine (`go func() { chResult <- check(t) }()`). `RunDockerComposeCmd` unconditionally calls `res.Assert(t, icmd.Success)`, which in turn calls `t.FailNow()` when `compose ps` exits non-zero.

Per Go testing rules (and gotest.tools' own docs), `t.FailNow()` must only be called from the test goroutine. When called from the poll goroutine it invokes `runtime.Goexit()` in that goroutine, which terminates it without ever sending a value to `chResult`. The `poll.WaitOn` select then blocks until the 15-second timeout fires, producing an opaque "timeout hit after 15s: first check never completed" error rather than the actual `compose ps` failure message.

`compose ps` does exit non-zero in transient/error conditions — for example when the project is not yet registered (`opts.projectOrName` fails) or the service name is not found in the loaded project file. These are exactly the states this poll helper is meant to absorb.

```suggestion
		psRes := cli.RunDockerComposeCmdNoCheck(t, "ps", "--all", "--format=json", service)
```

Use `RunDockerComposeCmdNoCheck` instead. If `compose ps` returns a non-zero exit, treat it as a transient state with `poll.Continue` (or `poll.Error` for unrecoverable failures) rather than letting `t.FailNow()` escape the goroutine boundary.

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 100/100 |

docker-agent (commented):

One CONFIRMED medium-severity finding in the new `RequireServiceState` poll helper.

pkg/e2e/assert.go
ndeloof: Fixed in 42a7b35f4 (squashed into the deflake commit): the poll closure now uses `RunDockerComposeCmdNoCheck` and maps a non-zero `compose ps` to `poll.Continue` with the exit code and output — transient states keep polling, and no `t.FailNow()` can escape the poll goroutine anymore (capture still works: icmd tees the pre-set Stdout with its result buffer).

docker-agent: Excellent fix — verified the implementation. `RunDockerComposeCmdNoCheck` now correctly handles non-zero exits as transient states with `poll.Continue`, and the goroutine boundary issue is resolved. The detailed error messages (exit code + output in the poll continuation string) will make debugging much clearer than the previous opaque timeout.

The empty-output case is a nice touch for early-stage creation races, and the multi-line JSON handling accounts for mid-transition states where both containers are briefly visible. Thanks for the thorough resolution.

docker-agent (commented):

The anchor-on-FinishedAt approach is sound, and the `runEndTracker` design correctly captures the previous run's exit time synchronously before the re-attach goroutine starts. One edge case in the new `Observe` guard may reintroduce the original log-loss bug in rare daemon environments.

pkg/compose/logs.go
```diff
@@ -136,6 +148,67 @@ func (s *composeService) followStartedContainersLogs(ctx context.Context, eg *er
...
+	ends map[string]int64 // container ID → TimeNano of the last observed exit
+}
+
+func newRunEndTracker() *runEndTracker {
+	return &runEndTracker{ends: map[string]int64{}}
+}
+
+// Observe records exit events (other event types are ignored).
+func (t *runEndTracker) Observe(e api.ContainerEvent) {
+	if e.Type != api.ContainerEventExited || e.Time == 0 {
```
docker-agent, line 169: **[medium] `Observe` silently discards exit events with `Time == 0`, falling back to `logsSinceLastRun` which may read a stale `FinishedAt`**

`runEndTracker.Observe` skips any exit event whose `Time` field is zero:

```go
if e.Type != api.ContainerEventExited || e.Time == 0 {
    return
}
```

If the Docker daemon delivers a `die` event with `TimeNano == 0` (a known edge-case on some daemon builds / containerised environments where the event clock is unreliable), the tracker never records that exit. When the container restarts and `ContainerEventStarted` arrives, `runEnds.Since(id)` returns `""`, triggering the `logsSinceLastRun` fallback. On a fast run the goroutine may not reach `ContainerInspect` until after the new run has already finished — at which point `FinishedAt` has been overwritten with the new run's end time, and `since` is anchored past all of the new run's output: the original log-loss bug reappears.

Consider either storing `0` (letting the downstream call decide what "no time" means) and only filtering in `Since`, or treating a `Time == 0` exit event like a first-start (returning `""` from `Since`) so the fallback path is never reached for this case. Alternatively, a comment explaining why `Time == 0` is safe to skip would help reviewers understand the invariant.

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 57/100 |

ndeloof: Addressed in 516230b6f by documenting the invariant (the third option you offered). The two code alternatives are behaviorally identical to the current skip: storing 0 and filtering in Since, or returning empty from Since, both land the container on the same logsSinceLastRun fallback. The only substantive change would be patching the missing timestamp with the local clock, and that is a bad trade: the anchor is evaluated by the DAEMON against its own log timestamps, so local-clock substitution introduces real skew mis-anchoring to cover a hypothetical daemon quirk. Dropping the event degrades exactly to the pre-tracker behavior, imperfect only for a run fast enough to have finished again by inspection time.

glours (dismissed):
`runEndTracker`/drain/`stopLateStarter` are well-reasoned and well-tested.

One gap worth a follow-up: the race `stopLateStarter` fixes for `stopOnFirstExit` (on-exit cascade) still seems open for `gracefulTeardown` (Ctrl+C/SIGINT/SIGTERM).

`monitor.Start` only returns once its `containers` set is empty, and `onContainerStart` re-adds any container that starts, regardless of why. `stopApplication()`'s stop sweep is a one-shot listing — a container that starts *after* it keeps `monitor.Start` blocked until it exits on its own. This PR wires `stopLateStarter` to catch that, but only inside `stopOnFirstExit`'s own closure. `gracefulTeardown` calls the same `stopApplication()` with no equivalent listener.

Worse, the window is *wider* here: `stopOnFirstExit` only fires once something has already exited (startup is well underway), while Ctrl+C can land at `t≈0`, before `monitor.Start` even starts and while `s.start()` (on its deliberately uncancelable context) is still climbing the dependency graph — most services may still be `created`, not `running`. The only recovery today is a second manual Ctrl+C.

Not a regression from this PR, so I wouldn't block on it — but same root cause, same file, worth a fast follow-up (e.g. arm the late-starter watch once `u.isTerminated` is set, not just inside `stopOnFirstExit`).

ndeloof:
Addressed in 39bf0d654: extracted the late-starter watch out of `stopOnFirstExit` into its own `stopLateStarters` listener, armed on `u.isTerminated` (already set by `gracefulTeardown`, just not consulted for this) instead of the local `once` flag, and registered unconditionally so it also covers Ctrl+C/SIGTERM regardless of the on-exit policy. Added `TestIsLateStarter` -- the original fix (ca867482a) shipped with no dedicated regression test at all, only a flaky test that happened to hit the on-exit case.

glours (approved):
LGTM

PR #14142, opened 2026-08-27
chore(compose): orphan semantics stated where up and down diverge

ndeloof:
**What I did**

Epic #14074, item C.4. The orphan-container asymmetry between `up` and `down` is intentional but was stated nowhere — this PR writes it down and pins it with tests, with **no behavior change**:

- a container is **orphaned** when it is a one-off (`compose run`) that **finished its task**, or when it carries the project labels but its service is **not defined by the compose model** (the typical leftover after the compose file was edited);
- a still-**running** one-off is somebody's live session: `up --remove-orphans` cleans up leftovers and never kills it (it is deliberately absent from the observed state);
- `down --remove-orphans` is the explicit stop-the-application action: it **does** take running one-offs down, through the per-service removal loop — now said explicitly in `down.go` instead of being an unexplained side effect of the container listing.

Tests pin the predicate matrix, the observed-state classification (a running one-off is neither a service replica nor an orphan) and down's single stop+remove path for a running one-off of a declared service.

**Related issue**
Item C.4 of #14074

### DIFF
--- a/pkg/compose/containers.go
+++ b/pkg/compose/containers.go
@@ -142,18 +142,26 @@ func isService(services ...string) containerPredicate {
 	}
 }
 
-// isOrphaned is a predicate to select containers without a matching service definition in compose project
+// isOrphaned selects the containers `--remove-orphans` cleans up — and that
+// `up` warns about otherwise. A container is orphaned when:
+//
+//   - it is a one-off (`compose run`) that FINISHED its task (exited/dead):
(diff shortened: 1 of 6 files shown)


docker-agent (commented):
🟢 **No issues found** — LGTM! [View logs](https://github.com/docker/compose/actions/runs/33072235733).

glours (approved):
LGTM

PR #14143, opened 2026-08-27
fix(compose): reject unknown subcommands under bridge/transformations

glours:
**What I did**
`bridgeCommand` and `transformersCommand` had no `RunE`, so an unknown subcommand (e.g. "bridge zzz") fell through cobra's default non-Runnable path: help printed, exit 0. Scripts couldn't distinguish a typo from success. Mirrors the same fix already shipped on the compose root command.

**Related issue**
https://docker.atlassian.net/browse/DDB-666

image

### DIFF
--- a/cmd/compose/bridge.go
+++ b/cmd/compose/bridge.go
@@ -39,12 +39,27 @@ func bridgeCommand(p *ProjectOptions, dockerCli command.Cli) *cobra.Command {
 		Use:              "bridge CMD [OPTIONS]",
 		Short:            "Convert compose files into another model",
 		TraverseChildren: true,
+		RunE:             rejectUnknownSubcommand,
 	}
 	cmd.AddCommand(
 		convertCommand(p, dockerCli),
 		transformersCommand(dockerCli),
 	)
 	return cmd
+}
+
+// rejectUnknownSubcommand is the RunE for a parent command that only groups
+// subcommands: by default (no Run/RunE), cobra shows help for an unknown
+// subcommand but exits 0.
+func rejectUnknownSubcommand(cmd *cobra.Command, args []string) error {
+	if len(args) == 0 {
+		return cmd.Help()
+	}
+	_ = cmd.Help()
+	return cli.StatusError{
+		StatusCode: 1,
+		Status:     fmt.Sprintf("unknown docker command: %q", cmd.CommandPath()+" "+args[0]),
+	}
 }
 
 func convertCommand(p *ProjectOptions, dockerCli command.Cli) *cobra.Command {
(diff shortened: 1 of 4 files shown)


cmd/compose/bridge.go
```diff
@@ -47,6 +48,20 @@ func bridgeCommand(p *ProjectOptions, dockerCli command.Cli) *cobra.Command {
...
 }
 
+// rejectUnknownSubcommand is the RunE for a parent command that only groups
+// subcommands: by default (no Run/RunE), cobra shows help for an unknown
+// subcommand but exits 0.
+func rejectUnknownSubcommand(cmd *cobra.Command, args []string) error {
+	if len(args) == 0 {
+		return cmd.Help()
+	}
+	_ = cmd.Help()
```
docker-agent, line 58: **[low] Error from `cmd.Help()` silently discarded before returning StatusError**

When `args` is non-empty, `rejectUnknownSubcommand` calls `_ = cmd.Help()` and discards the error before returning the `cli.StatusError`. If the cobra output writer returns an I/O error (e.g. a broken pipe), the help text won't render and the error will be silently swallowed — the caller only sees the `StatusError`, with no indication that help rendering failed and no help context displayed.

The conventional Go pattern is to propagate the error:

```suggestion
	if err := cmd.Help(); err != nil {
		return err
	}
```

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 100/100 |

docker-agent (commented):
### Assessment: 🟢 APPROVE

The PR correctly wires `rejectUnknownSubcommand` as `RunE` on the `bridgeCommand` and `transformersCommand` parent commands so that unknown subcommands exit 1 instead of 0. The core logic is sound: cobra dispatches unknown tokens to the parent's `RunE`, known subcommands are dispatched directly so the new guard is never incorrectly invoked, and the tests cover all the key scenarios (unknown subcommand, no subcommand, nested transformations).

One low-severity finding below: a discarded `cmd.Help()` error in the non-zero-args path.

cmd/compose/bridge.go
glours: Not changing this, propagating the error here would drop the unknown docker command StatusError in favor of a raw I/O error on the one edge case (broken pipe) it could ever trigger

docker-agent: You're right — preserving the semantic StatusError that tells the user "unknown docker command" is more valuable than catching an I/O error on help rendering. The broken pipe case is rare (user pipes to head/less and quits early), and when it happens, they've likely already seen the help text anyway before the pipe broke.

Thanks for the context on prioritizing the domain error over the transport error.

ndeloof (approved):

PR #14144, opened 2026-08-27
feat(run): --no-recreate reuses existing dependencies as-is

### LINKED ISSUE #13769: container re-created when it shouldn't, if using `docker-compose run`
> I confirm strange recreation when I do `docker-compose run --rm my_service_with_oneoff_tools`. The service defines `depends_on` the db service and the db doesn't have the expected users after recreation.
> 
> The issue is not reproduced with older version of docker-compose (docker-compose version 1.29.2, build 5becea4c) 

 _Originally posted by @DXist in [#9897](https://github.com/docker/compose/issues/9897#issuecomment-1453704821)_

ndeloof:
`run` starts its dependencies through the regular create path and its config-hash divergence check, with no way to opt out: any resolution difference with the command that created them recreates the dependency — destroying state a previous run built up (#13769: database losing its users). `up` already offers `--no-recreate` for exactly this; `run` now aligns.

The flag maps to the existing "never" recreate policy, propagated through the `RunOptions.CreateOptions` fields `startDependencies` previously dropped. Default behavior is unchanged (empty policy = "diverged"). E2E test pins both sides: default recreates a diverged dependency, `--no-recreate` reuses it as-is.

Closes #13769

### DIFF
--- a/cmd/compose/run.go
+++ b/cmd/compose/run.go
@@ -241,6 +241,7 @@ func runCommand(p *ProjectOptions, dockerCli command.Cli, backendOptions *Backen
 	flags.BoolVar(&buildOpts.quiet, "quiet-build", false, "Suppress progress output from the build process")
 	flags.BoolVar(&options.quietPull, "quiet-pull", false, "Pull without printing progress information")
 	flags.BoolVar(&createOpts.Build, "build", false, "Build image before starting container")
+	flags.BoolVar(&createOpts.noRecreate, "no-recreate", false, "If dependent containers already exist, don't recreate them, even if their configuration diverged")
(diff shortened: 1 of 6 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The  flag implementation is clean and correct. The flag wires properly from the CLI option through `createOptions.noRecreate` → `recreateStrategy()` / `dependenciesRecreateStrategy()` (both already handle `noRecreate` → `RecreateNever`), and the propagation into `startDependencies` via `options.Recreate` / `options.RecreateDependencies` is complete. Default behavior is unchanged. No logic errors, nil-pointer risks, or missing error handling found in the changed code.

PR #14145, opened 2026-08-27
docs: register every environment variable the binary reads

ndeloof:
Epic #14074, section D (env-var registry): ~25 recognized variables scattered across five constant files plus the compose-go-handled ones, 6 of them documented. Whether the project `.env` can set a given variable depends on an unwritten fact — the moment it is read (flag-default time, after `setEnvWithDotEnv` injection, or from `project.Environment`).

`docs/envvars.md` is now the single registry: name → read site → moment → default, including the known divergences (`COMPOSE_REMOVE_ORPHANS` per-command inconsistency → #14139). A guardrail test sweeps production sources for `COMPOSE_*` literals and fails when one has no row — the registry stays exhaustive or the build says so.

**Stacked on #14085** (base `env-legacy-leftovers`): the registry documents the post-cleanup state — `COMPOSE_EXPERIMENTAL`/`COMPOSE_BAKE` listed as gone, `COMPOSE_EXPERIMENTAL_WATCH_TAR` as recognized-but-ignored. Will be retargeted to `main` once #14085 merges.

Documentation + test only, no runtime change.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -26,6 +26,15 @@
 - In tests: use `t.Context()` instead of `context.Background()` or `context.TODO()`
 - Prefer `fmt.Fprintf` over `WriteString(fmt.Sprintf(...))`
 
+## Environment variables
+
+- Every environment variable the binary reads is registered in
+  `docs/envvars.md` (name, read site, moment, default) — consult it before
+  adding a read, and update it in the same commit when you do
+  (`docs/envvars_registry_test.go` fails otherwise). The moment a variable is
+  read decides whether the project `.env` can set it; the registry explains
+  the three moments.
+
 ## Git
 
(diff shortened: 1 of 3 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The env-var registry and its guardrail test look correct. The test walks `../cmd`, `../pkg`, and `../internal`, correctly filters out `_test.go` files and `e2e` directories, and asserts every double-quoted `COMPOSE_*` literal has a row in `docs/envvars.md`. The registry itself is internally consistent and the new AGENTS.md convention is clear.

One potential gap was identified (regex only matches double-quoted string literals, not backtick raw-string literals) but the verifier confirmed no production code currently uses backtick-quoted `COMPOSE_*` literals — the risk is theoretical and the test reliably guards the current and foreseeable codebase patterns.

PR #14149, opened 2026-08-28
api: DownOptions.Images becomes a typed ImagePruneMode, validated upfront

ndeloof:
Epic #14074, B.6. The legal values of `DownOptions.Images` lived in `pkg/compose` while the field was a bare string in `pkg/api`, and the only validation fired inside `ImagesToPrune` — after `down` had already removed the project's containers, leaving the teardown half done on a typo (SDK callers only; the CLI validated its `--rmi` flag separately).

- `ImagePruneMode` and its three values move to `pkg/api`, next to the field they constrain (type-aliased in `pkg/compose` for existing consumers — untyped string literals still compile);
- `down()` rejects any other value **before touching a single resource**;
- the CLI `--rmi` check reuses the same definition;
- a unit test runs `Down` with expectation-free mocks: one daemon call would fail it. Mocks verified in sync (`make mocks`, no drift).


docker-agent (commented):
### Assessment: 🟢 APPROVE

pkg/compose/image_pruner.go
glours, line 81: This branch is dead code now: `down()` (pkg/compose/down.go) already validates `options.Images.Valid()` before ever reaching `ensureImagesDown`/`ImagesToPrune`, so `opts.Mode` here can only be `ImagePruneLocal` or `ImagePruneAll` by the time it's checked again. Re-deriving the same set of legal values in two places means they can silently drift if a fourth mode is ever added.

```suggestion
      if opts.Mode == ImagePruneNone {
              return nil, nil
      } else if !opts.Mode.Valid() {
              return nil, fmt.Errorf("unsupported image prune mode: %s", opts.Mode)
      }
```

glours (commented):
Something we need to check before we merge IMHO

PR #14150, opened 2026-08-28
fix(dry-run): commit no longer creates a real image; the interception set becomes declarative

ndeloof:
Epic #14074, B.7. `compose commit --dry-run` delegated `ContainerCommit` to the real client and actually created the image; it now returns a dry-run ID without touching the engine (e2e step added, with a new `ImageAbsent` check).

Auditing the rest of the surface: **39 more mutating operations were delegated** — none that compose itself issues today (verified caller by caller), but any future code path reaching one would mutate the engine under `--dry-run` silently. They now refuse with an explicit `<op> is not allowed in dry-run mode`.

The interception set was only implied by a method's position under a comment; it is now **declared** in `dryrunclient_test.go` (read-only delegated / faked / refused), with an AST-based test that fails on any method missing, misclassified, or new and unclassified — extending `APIClient` requires an explicit dry-run decision.

### DIFF
--- a/pkg/dryrun/dryrunclient.go
+++ b/pkg/dryrun/dryrunclient.go
@@ -100,7 +100,17 @@ func getCallingFunction() string {
 	return fullName[strings.LastIndex(fullName, ".")+1:]
 }
 
-// All methods and functions which need to be overridden for dry run.
+// errDryRunForbidden refuses a mutating operation compose itself never
+// issues: delegating it would let future code mutate the engine under
+// --dry-run silently. The full classification of every APIClient method
+// (read-only delegated / faked / refused) is pinned by
+// TestDryRunClientClassifiesEveryMethod.
+func errDryRunForbidden(operation string) error {
+	return fmt.Errorf("%s is not allowed in dry-run mode", operation)
+}
+
+// Methods below either fake the operation for dry-run or delegate the
+// read-only ones to the real client.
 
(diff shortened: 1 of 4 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The changes are mechanically correct and well-structured. The 39 new refused stubs follow a consistent, uniform pattern. The AST-based classification test (`TestDryRunClientClassifiesEveryMethod`) provides strong bidirectional coverage — it catches both unclassified new methods and classified names that no longer exist on `DryRunClient`. `ContainerCommit` now returns a synthetic dry-run ID instead of delegating to the real engine (defense-in-depth, since `pkg/compose/commit.go` already short-circuits on `s.dryRun`). The new `ImageAbsent` check and the e2e dry-run commit step are logically sound.

glours (commented):
LGTM

Note: I verified the premise couldn't actually reproduce, built the parent commit (727fa0e67) and ran `docker compose --dry-run commit` against a live daemon; no image was created. `pkg/compose/commit.go`'s pre-existing `if s.dryRun { return nil }` guard already short-circuits before ContainerCommit is ever reached, so this wasn't an active bug.

Still worth merging as defense-in-depth (same rationale as the other 39 refused methods, for a future caller that might not have that guard). Just consider tweaking the commit message so it doesn't read as fixing a live regression.

glours (approved):

glours (approved):

PR #14151, opened 2026-08-28
cli: one project-resolution story — single name precedence, no swallowed load error, strict service validation

ndeloof:
Epic #14074, F.4, per the agreed behavior:

**One name precedence** (Q1), documented on both resolvers and applied identically by compose-go while loading: `--project-name` > `COMPOSE_PROJECT_NAME` > the model's name. `projectOrName` and `toProjectName` used to disagree.

**Explicit failure policy** (Q2.c) — `projectOrName` used to swallow *any* load error when `COMPOSE_PROJECT_NAME` was set, silently sending `stop`/`down`/`ps`… into label-based reconstruction, even for an explicit `--file`:
- unreadable explicit `--file` → hard error;
- no compose file around + `COMPOSE_PROJECT_NAME` → the normal file-less workflow, silent;
- present-but-broken implicit file + `COMPOSE_PROJECT_NAME` → label-based fallback **with a warning**.

**Strict service validation whenever a model is available** (Q3): `restart` and `wait` no longer silently no-op on a typo (profile-disabled services remain legitimate targets); the hand-rolled checks in `ps`/`volumes` are removed as dead code — load-time selection already rejects unknown names (pinned by test).

Behavioral changes: `restart`/`wait` on an unknown service now error; a broken explicit `--file` now errors instead of silently falling back; a broken implicit file now warns. Unit tests cover the full resolution matrix.

### DIFF
--- a/cmd/compose/compose.go
+++ b/cmd/compose/compose.go
@@ -31,6 +31,7 @@ import (
 
 	"github.com/compose-spec/compose-go/v2/cli"
 	"github.com/compose-spec/compose-go/v2/dotenv"
+	"github.com/compose-spec/compose-go/v2/errdefs"
 	"github.com/compose-spec/compose-go/v2/loader"
 	composepaths "github.com/compose-spec/compose-go/v2/paths"
 	"github.com/compose-spec/compose-go/v2/types"
@@ -243,6 +244,20 @@ func defaultStringArrayVar(env string) []string {
 	})
 }
 
+// projectOrName resolves the target project for commands that exploit the
+// compose model when one is available and fall back to container labels
+// otherwise. The project name follows one precedence everywhere, shared with
+// toProjectName and applied identically by compose-go while loading:
+// --project-name, then COMPOSE_PROJECT_NAME, then the model's name.
+//
+// When the model cannot be loaded:
+//   - an explicit --file is a hard error: the user named the file, failing
+//     to read it cannot be ignored;
+//   - no compose file around and a name available from COMPOSE_PROJECT_NAME
+//     is the normal file-less workflow: label-based mode, silently;
+//   - a compose file present but broken, with COMPOSE_PROJECT_NAME set,
+//     falls back to label-based mode with an explicit warning (this used to
+//     happen silently).
 func (o *ProjectOptions) projectOrName(ctx context.Context, dockerCli command.Cli, services ...string) (*types.Project, string, error) {
 	name := o.ProjectName
 	var project *types.Project
@@ -254,8 +269,14 @@ func (o *ProjectOptions) projectOrName(ctx context.Context, dockerCli command.Cl
 
 		p, _, err := o.ToProject(ctx, dockerCli, backend, services, cli.WithDiscardEnvFile, cli.WithoutEnvironmentResolution)
 		if err != nil {
+			if len(o.ConfigPaths) > 0 {
+				return nil, "", err
+			}
 			envProjectName := os.Getenv(ComposeProjectName)
 			if envProjectName != "" {
+				if !errdefs.IsNotFoundError(err) {
+					logrus.Warnf("compose file found but could not be loaded (%s) — falling back to label-based mode for project %q", err, envProjectName)
+				}
 				return nil, envProjectName, nil
 			}
 			return nil, "", err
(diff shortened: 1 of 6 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The PR is well-structured and implements the described behavior correctly:

- **Name precedence** (`--project-name` > `COMPOSE_PROJECT_NAME` > model name) is applied consistently through the refactored `projectOrName` and `toProjectName`.
- **Error handling** for explicit `--file` failures (hard error), missing files with env name (silent), and broken implicit files with env name (warning + fallback) is correct and clearly documented.
- **`validateServiceNames`** correctly handles the nil-project case (label-based mode has no manifest to validate against) and covers profile-disabled services as legitimate targets.
- The removal of hand-rolled service checks in `ps.go` and `volumes.go` is safe: `ToProject` service-selection already rejects unknowns at load time when a model is available.
- The new test file covers the full resolution matrix with `t.Context()` and clearly-named subtests.

ndeloof:
Re-checked this PR's legitimacy given everything that's landed since it was opened (Aug 28) — it's still exactly the right fix (Q1/Q2.c/Q3 from #14074 are unaffected by the jobs work), but it needed reconciling with `projectOrName`'s new job-target detection (#14093, #14234, both merged since):

- **`jobTargetErr` restored ahead of the new explicit-file hard error**: this PR's rewrite replaced `projectOrName`'s whole error-handling body, dropping the job check entirely. Without it, `docker compose stop migrate -f compose.yaml` (a declared job, explicit file) would have regressed from the clear `job "migrate" can only be triggered with "docker compose run"` back to a raw `no such service: migrate` — the file loads fine, only the target's selection fails, so the job check must run before the new explicit-file/implicit-file branching, not be skipped by it.
- **`validateServiceNames` now also checks `project.AllJobs()`**: `restart`/`wait` don't route their service arguments through `projectOrName`'s own `ToProject` selection (they call it with no services, precisely so `validateServiceNames` validates the full name afterward) — a separate code path from every other `projectOrName` caller, so `restart migrate`/`wait migrate` would otherwise have fallen through to the generic `no such service` instead of the job-specific message.
- Added `TestRestartRefusesJob`/`TestWaitRefusesJob` (e2e) and a job case in `TestValidateServiceNames`, since this is genuinely distinct wiring from the other commands' job-refusal coverage.

Rebased onto current `main`, build/vet/lint clean, full unit suite green, all job-refusal e2e tests (create/start/stop/down/restart/wait) pass.

cmd/compose/compose.go
```diff
@@ -260,12 +279,20 @@ func (o *ProjectOptions) projectOrName(ctx context.Context, dockerCli command.Cl
...
 			if jobErr, replaced := jobTargetErr(ctx, dockerCli, o, services, err); replaced {
 				return nil, "", jobErr
 			}
+			if len(o.ConfigPaths) > 0 {
+				return nil, "", err
+			}
 			envProjectName := os.Getenv(ComposeProjectName)
 			if envProjectName != "" {
+				if !errdefs.IsNotFoundError(err) {
+					logrus.Warnf("compose file found but could not be loaded (%s) — falling back to label-based mode for project %q", err, envProjectName)
```
docker-agent, line 294: **[medium] Spurious "file found but could not be loaded" warning in the no-file workflow**

When `COMPOSE_PROJECT_NAME` is set and no compose file exists in the working directory, `ToProject` returns an error (file not found). The new guard at line 293 calls `errdefs.IsNotFoundError(err)` — using `github.com/compose-spec/compose-go/v2/errdefs` — to suppress the `logrus.Warnf` in that case. But if compose-go wraps the OS-level `os.ErrNotExist` in a way that `errdefs.IsNotFoundError` does not recognise (e.g. returns a raw syscall error or wraps it with a non-compose-go type), the guard evaluates to `false` and the warning fires:

> `compose file found but could not be loaded (…) — falling back to label-based mode for project "X"`

That message is factually wrong when no file exists at all. The function still returns the correct `nil, envProjectName, nil`, so behaviour is not broken — only the user-facing message is misleading.

The test `"no file at all with COMPOSE_PROJECT_NAME is the file-less workflow"` validates that no error is returned but does not assert that no warning is printed, so this path would be undetected by the test suite if the guard misses.

**Suggested mitigation**: add `|| errors.Is(err, os.ErrNotExist)` as a fallback, or confirm that compose-go's `errdefs.IsNotFoundError` correctly matches `os.ErrNotExist`-wrapped errors from its own loader:
```suggestion
				if !errdefs.IsNotFoundError(err) && !errors.Is(err, os.ErrNotExist) {
```

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 65/100 |

docker-agent (commented):

cmd/compose/compose.go
ndeloof: Good catch — fixed in 065922e1d: added `errors.Is(err, os.ErrNotExist)` as a fallback to the guard, and added assertions on the warning's presence/absence in both directions in `TestProjectOrNameResolution` (the existing compose-go path already wraps `errdefs.ErrNotFound` correctly for the no-file case, but this protects against a raw os-level not-found, e.g. a nonexistent --project-directory).

PR #14152, opened 2026-08-28
feat(tracing): surface otel shutdown errors via --debug

htoyoda18:
**What I did**

The tracer shutdown/flush error in `wrapRunE` was always discarded, making it impossible to diagnose tracing issues (e.g. a misconfigured or unreachable OTLP endpoint) without instrumenting the code.

~~Originally fixed this with a new `COMPOSE_OTEL_DEBUG` env var~~ — per review feedback, routed it through the existing `--debug`/`-D` mechanism instead:
- the previously-discarded `tracingShutdown(ctx)` error now goes through `logrus.WithError(err).Debug(...)`, which is already gated by `--debug`/`-D` via logrus's own log level — no explicit check needed
- also removed `internal/tracing`'s custom `otel.ErrorHandler` and the `COMPOSE_OTEL_DEBUG` env var entirely: `docker/cli`'s own plugin framework (`plugin.Run()`) already installs `otel.SetErrorHandler(debug.OTELErrorHandler)` at startup, which runs *after* our package's `init()` and was silently clobbering our handler — so it had never actually been in effect

**Related issue**
N/A

image


cmd/cmdtrace/cmd_span.go
```diff
@@ -62,7 +63,7 @@ func Setup(cmd *cobra.Command, dockerCli command.Cli, args []string) error {
 	)
 
 	cmd.SetContext(ctx)
-	wrapRunE(cmd, cmdSpan, tracingShutdown)
+	wrapRunE(cmd, cmdSpan, tracingShutdown, dockerCli.Err())
```
glours: No needed , we should use logrus to write the error which is already configured to write in the right IO


```suggestion
	wrapRunE(cmd, cmdSpan, tracingShutdown)
```

cmd/cmdtrace/cmd_span.go
```diff
@@ -72,7 +73,7 @@ func Setup(cmd *cobra.Command, dockerCli command.Cli, args []string) error {
 //
 // Unfortunately, PersistentPostRun(E) can't be used for this purpose because it
 // only runs if RunE does _not_ return an error, but this should run unconditionally.
-func wrapRunE(c *cobra.Command, cmdSpan trace.Span, tracingShutdown tracing.ShutdownFunc) {
+func wrapRunE(c *cobra.Command, cmdSpan trace.Span, tracingShutdown tracing.ShutdownFunc, errOut io.Writer) {
```
glours:
```suggestion
func wrapRunE(c *cobra.Command, cmdSpan trace.Span, tracingShutdown tracing.ShutdownFunc) {
```

cmd/cmdtrace/cmd_span.go
```diff
@@ -110,9 +111,9 @@ func wrapRunE(c *cobra.Command, cmdSpan trace.Span, tracingShutdown tracing.Shut
 			// been canceled already
 			ctx, cancel := context.WithTimeout(context.Background(), 100*time.Millisecond)
 			defer cancel()
-			// TODO(milas): add an env var to enable logging from the
-			// OTel components for debugging purposes
-			_ = tracingShutdown(ctx)
+			if err := tracingShutdown(ctx); err != nil && tracing.DebugEnabled() {
+				fmt.Fprintln(errOut, "otel: shutdown:", err)
```
glours: Use logrus

```suggestion
				logrus.Debugf("otel: shutdown: %v", err)
```

internal/tracing/errors.go
```diff
@@ -17,13 +17,31 @@
...
-
-var _ otel.ErrorHandler = skipErrors{}
+// DebugEnabled reports whether OTel SDK/exporter internals should print
+// their diagnostics, controlled by the COMPOSE_OTEL_DEBUG environment
+// variable. It defaults to false so tracing plumbing never leaks into
+// ordinary CLI output.
+func DebugEnabled() bool {
+	enabled, _ := strconv.ParseBool(os.Getenv("COMPOSE_OTEL_DEBUG"))
+	return enabled
+}
```
glours: no need for a dedicated env var; `logrus.Debugf` is already a no-op unless `--debug` is set, so the check is redundant.

internal/tracing/errors.go
```diff
@@ -17,13 +17,31 @@
...
+
+// errorHandler is the otel.ErrorHandler installed for the CLI: it discards
+// errors unless DebugEnabled reports true, in which case it prints them to
+// stderr.
+type errorHandler struct{}
+
+func (errorHandler) Handle(err error) {
+	if DebugEnabled() {
+		fmt.Fprintln(os.Stderr, "otel:", err)
+	}
```
glours: reuses the same debug mechanism cmd/main.go:55 already uses for tracing setup failures, instead of a second, parallel gate for the same diagnostic category.


```suggestion
	logrus.Debugf("otel: %v", err)
```

internal/tracing/tracing.go
```diff
@@ -40,8 +40,8 @@ import (
 
 func init() {
 	detect.ServiceName = "compose"
-	// do not log tracing errors to stdio
-	otel.SetErrorHandler(skipErrors{})
+	// do not log tracing errors to stdio, unless COMPOSE_OTEL_DEBUG is set
```
glours:
```suggestion
	// do not log tracing errors to stdio, unless `--debug` is set
```

glours (changes requested):
Thanks for the PR,  the diagnosability goal makes sense, but we don't think a new environment variable is needed here: Compose already has a debug mechanism with `--debug` that surfaces this exact kind of tracing diagnostic elsewhere in the codebase. 
  
Could you route the OTel error printing through that instead of introducing `COMPOSE_OTEL_DEBUG`? Should end up simpler, with a single debug switch instead of two overlapping ones.

htoyoda18:
Thanks for the review, @glours — you're right, and it turned out to go even deeper than the env var: `docker/cli`'s own plugin framework already installs `otel.SetErrorHandler(debug.OTELErrorHandler)`at startup, which runs after our package's `init()` and was silently overriding our custom handler the whole time, so it had never actually been in effect.

Dropped the custom `otel.ErrorHandler` and `COMPOSE_OTEL_DEBUG` entirely, and routed the one genuinely-discarded error (the tracer shutdown/flush error) through `logrus.WithError(err).Debug(...)`, which is already gated by `--debug/-D` via logrus's own log level. Updated the PR description with the details. Would appreciate another look when you have a moment 🙏

glours (approved):
Sounds good to me! Thanks

PR #14153, opened 2026-08-29
fix(formatter): handle nil JSON values

zloglevel:
**What I did**

Prevent `formatter.Print` from panicking when JSON output receives an untyped nil value. `reflect.TypeOf(nil)` returns nil, so calling `Kind()` on it panics; `reflect.ValueOf(nil).Kind()` safely returns `Invalid` and lets the existing standard JSON path emit `null`.

Added regression coverage for both `json` and legacy `{{json .}}` formatting. Both should produce `null\n`.

**Related issue**

No issue; this is a small, self-contained nil-input fix.

0502d80a9e7cd8cff98f188e59b65b13

### DIFF
--- a/cmd/formatter/formatter.go
+++ b/cmd/formatter/formatter.go
@@ -31,7 +31,7 @@ func Print(toJSON any, format string, outWriter io.Writer, writerFn func(w io.Wr
 	case TABLE, PRETTY, "":
 		return PrintPrettySection(outWriter, writerFn, headers...)
 	case TemplateLegacyJSON:
-		switch reflect.TypeOf(toJSON).Kind() {
+		switch reflect.ValueOf(toJSON).Kind() {
 		case reflect.Slice:
 			s := reflect.ValueOf(toJSON)
 			for i := 0; i < s.Len(); i++ {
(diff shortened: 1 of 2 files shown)


glours:
Thanks for the PR, but I'm closing this one.

No reproducible issue. There's no linked issue, and we tested and verified this directly: none of Compose's actual call paths can put us in this situation. All four call sites of formatter.Print (bridge.go, config.go, images.go, list.go) pass concretely-typed values (slices/maps), even when empty — never a literal untyped nil. reflect.TypeOf(x).Kind() only panics on a genuinely nil any, and none of our call sites can produce that. We couldn't construct any real Compose command that hits this panic.

The PR doesn't even pass its own test. go test ./cmd/formatter/... fails on this branch — TestPrintNilAsJSON expects "null\n" but gets "null\n\n".

We've also noticed a pattern consistent with automated/low-effort PR farming on this account (near-identical PRs opened across many unrelated repos within the same short window). That's not something we want to encourage on this repo, even when a submission happens to be well-formatted.

If you hit this in a real scenario, feel free to reopen with a repro tied to an actual Compose command.

PR #14154, opened 2026-08-30
fix(down): spare dangling images of orphaned services, like their tagged image

htoyoda18:
**What I did**

`down --rmi` never cleaned up dangling (untagged) images left behind by earlier local builds of the project — historically skipped because printing raw image IDs in CLI progress output is confusing on its own.

Per review feedback, the first version of this fix had a gap: it removed every dangling image for the project unconditionally, while the tagged-image counterpart (`ImagesToPrune`) already skips a service no longer in the project unless `--remove-orphans` is set. That meant `down --rmi local` on a project with an orphaned service kept the tagged image but deleted its dangling ones — an inconsistent result for one command.

Fixed by extracting the "list a project's dangling images, remove what a `keep` predicate doesn't spare, tolerate individual failures" logic — which `down.go` and `watch.go`'s `pruneDanglingImagesOnRebuild` had each partially and divergently implemented — into one shared `composeService.removeDanglingImages`, parallelized via `errgroup` like the rest of the codebase's bulk removals:
- `down.go`'s `keep` now mirrors `ImagesToPrune`'s own `RemoveOrphans`/`GetService` check, so dangling and tagged images are spared/removed consistently
- `watch.go`'s `pruneDanglingImagesOnRebuild` now calls the same helper with its existing built-IDs check as `keep` — behavior unchanged, just de-duplicated
- removal is reported as a single grouped "Dangling images" event for `down`, instead of one row per meaningless raw image ID

**Related issue**
N/A

image


glours (changes requested):
Thanks for the PR — the intent and the modeling on `pruneDanglingImagesOnRebuild` make sense, but I'd like to ask for a slightly different shape before merging.                                                                                      
                                                                                                                                                                                                                                                        
Right now `danglingLocalImages`/`pruneDanglingProjectImages` introduce a second, parallel path next to the one `ImagePruner` already owns for tagged images, and it diverges from it in a way that's user-visible: `ImagesToPrune` skips images belonging to a service no longer in the compose file unless `RemoveOrphans` is set, but the new dangling-image path has no such check, it removes every dangling image for the project unconditionally. 
Concretely, `down --rmi local` on a project with an orphaned service would leave that service's tagged image on disk while deleting its dangling image, which is an inconsistent result for the same removal command.                                                                             
                                                                                                                                                                                                                                                        
  Rather than patching that check in as a one-off, could we fold this into a small shared helper that both `down.go` and `watch.go`'s `pruneDanglingImagesOnRebuild` call, since they're solving the same problem (list dangling images for a project,  
  remove the ones that should go, tolerate individual failures) with only the "which ones should go" and "should this be parallelized" parts differing? Something like:                                                                                 
                                                                                                                                                                                                                                                        
  ```go                                                                                                                                                                                                                                                 
  func (s *composeService) removeDanglingImages(ctx context.Context, projectName string, keep func(image.Summary) bool) ([]string, error)                                                                                                               
  ```                                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                        
  - `down.go` would pass a `keep` that mirrors `ImagesToPrune`'s own `GetService`/`RemoveOrphans` check, and wrap the call in the existing `removeResource` helper for event reporting instead of hand-rolling the Working/Done pair.                   
  - `watch.go` would pass its existing `builtIDs` exclusion as `keep`, with no event wrapping, preserving today's silent behavior.                                                                                                                      
  - The removal loop inside the helper can fan out per-image via `errgroup`, matching how every other bulk-removal loop in this codebase already works, instead of removing dangling images one at a time sequentially.                                 
                                                                                                                                                                                                                                                        
That would fix the `RemoveOrphans` gap, drop the duplicated event-handling code, and keep both call sites consistent going forward instead of two separately-maintained copies of the same logic. 
Happy to discuss if you see a reason to keep them separate.

htoyoda18:
Thanks for the thorough review, @glours! Addressed both points: extracted a shared `composeService.removeDanglingImages(ctx, projectName, keep)` used by both `down.go` and `watch.go`'s `pruneDanglingImagesOnRebuild`, with `down.go`'s `keep` now mirroring `ImagesToPrune`'s `RemoveOrphans`/`GetService` check (fixing the inconsistency), and removal now fans out via `errgroup` like the rest of the codebase's bulk removals. Added tests for the orphan-sparing behavior and the shared helper.

One trade-off worth flagging: since individual failures are swallowed inside the helper by design, the "Dangling images" event always reports `Removed` even if every removal actually failed. Happy to add failure-count plumbing if you'd rather it surface a `Warning` there.

Would appreciate another look 🙏

pkg/compose/image_pruner.go
```diff
@@ -164,6 +159,40 @@ func (p *ImagePruner) labeledLocalImages(ctx context.Context) ([]image.Summary,
...
+	res, err := s.apiClient().ImageList(ctx, client.ImageListOptions{
+		Filters: projectFilter(projectName).Add("dangling", "true"),
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	var mu sync.Mutex
+	var removed []string
+	eg, ctx := errgroup.WithContext(ctx)
```
glours, line 176: Unbounded `errgroup` fan-out — one goroutine per dangling image with no `eg.SetLimit(...)`, unlike every other bulk operation in this package (`pull.go:60/423`, `push.go:50`, `dependencies.go:140`) which caps concurrency at `s.maxConcurrency` to honor the user's `--parallel` flag.


```suggestion
	eg, ctx := errgroup.WithContext(ctx)
	eg.SetLimit(s.maxConcurrency)
```

glours (changes requested):
One more things, we should honor user's parallel computation configuration

htoyoda18:
Thanks, @glours — fixed: `removeDanglingImages` now calls `eg.SetLimit(s.maxConcurrency)`, matching `push.go`/`pull.go`, so `--parallel` is honored instead of always running unbounded.

As a side effect this caught a test fixture bug: `TestWatch_Sync` built `composeService{}` directly with a zero-value `maxConcurrency`, which `errgroup.SetLimit(0)` treats as "allow no goroutines" rather than "unlimited" — fixed by setting `maxConcurrency: -1` there, matching `NewComposeService's` own default.

Would appreciate another look 🙏

glours (approved):
LGTM

PR #14156, opened 2026-09-01
plan engine: the start phase (epic #14081, Lot 1)

ndeloof:
Lot 1 of #14081, built as three commits — each one CI-validated before the next lands on the branch (the PR stays draft until all three are green):

1. **`feat: reconciler plans the start phase`** (this push) — the plan vocabulary: `OpWaitCondition` (deduplicated per awaited service+condition, best-effort when every dependent is optional), `OpRunPreStart` (plan-time gating: only when no replica was running at observation), `OpRunPostStart`, `PlanNode.Phase`, replica chains preserving today's sequential start order, and `ReconcileOptions.Scope` (zero value = today's create-only plans, byte-identical — pinned by the existing 42 golden tests). Golden tests only, no consumer.
2. **`feat: executor runs start-phase operations`** — to follow once 1 is green.
3. **`refactor: split create() into preparePlan + execute`** — to follow once 2 is green.

### DIFF
--- a/pkg/compose/executor_ops.go
+++ b/pkg/compose/executor_ops.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"slices"
 
+	"github.com/compose-spec/compose-go/v2/types"
 	"github.com/containerd/errdefs"
 	"github.com/moby/moby/api/types/container"
 	"github.com/moby/moby/client"
@@ -127,11 +128,97 @@ func (exec *planExecutor) execCreateContainer(ctx context.Context, node *PlanNod
 	return nil
 }
 
-func (exec *planExecutor) execStartContainer(ctx context.Context, op Operation) error {
+func (exec *planExecutor) execStartContainer(ctx context.Context, node *PlanNode) error {
+	op := node.Operation
+	if node.Phase != PhaseStart {
+		// create-phase start (exceptional states: paused, dead, ...): a bare
+		// engine start, exactly the historical behavior
+		startMx.Lock()
+		defer startMx.Unlock()
+		_, err := exec.compose.apiClient().ContainerStart(ctx, op.Container.ID, client.ContainerStartOptions{})
+		return err
+	}
+
+	ctr, err := exec.startTarget(node)
+	if err != nil {
+		return err
+	}
+	// secrets and configs always inject as a pair right before the start —
+	// they are part of starting a container, not a separate decision
+	if err := exec.compose.injectSecrets(ctx, exec.project, *op.Service, ctr.ID); err != nil {
+		return err
+	}
+	if err := exec.compose.injectConfigs(ctx, exec.project, *op.Service, ctr.ID); err != nil {
+		return err
+	}
 	startMx.Lock()
 	defer startMx.Unlock()
-	_, err := exec.compose.apiClient().ContainerStart(ctx, op.Container.ID, client.ContainerStartOptions{})
+	_, err = exec.compose.apiClient().ContainerStart(ctx, ctr.ID, client.ContainerStartOptions{})
 	return err
+}
+
+// startTarget resolves the container a start-phase node acts on: the
+// observed summary carried by the plan, or the result of the create node
+// that materialized the replica earlier in the same plan.
+func (exec *planExecutor) startTarget(node *PlanNode) (container.Summary, error) {
+	op := node.Operation
+	if op.Container != nil {
+		return *op.Container, nil
+	}
+	res := exec.pctx.get(op.CreateNodeID)
+	if res.ContainerID == "" {
+		return container.Summary{}, fmt.Errorf("internal: %s has no materialized container to act on", op.ResourceID)
+	}
+	return container.Summary{ID: res.ContainerID, Names: []string{"/" + res.ContainerName}}, nil
+}
+
+// execWaitCondition re-observes the depends_on condition at execution time,
+// delegating the polling to the same waitDependency primitive the imperative
+// engine uses — Waiting, Healthy, Exited and Skipped events included. A
+// best-effort node (every dependent optional) absorbs a missing dependency
+// as a Skipped event instead of failing.
+func (exec *planExecutor) execWaitCondition(ctx context.Context, node *PlanNode) error {
+	op := node.Operation
+	exec.containersMu.Lock()
+	waitingFor := exec.containersByService[op.Name].filter(isNotOneOff)
+	exec.containersMu.Unlock()
+	if len(waitingFor) == 0 {
+		if op.BestEffort {
+			exec.compose.events.On(skippedEvent("Service "+op.Name, "no container to wait for"))
+			return nil
+		}
+		return fmt.Errorf("required dependency %q has no container to wait for", op.Name)
+	}
+	exec.compose.events.On(containerEvents(waitingFor, waiting)...)
+	return exec.compose.waitDependency(ctx, op.Name, types.ServiceDependency{
+		Condition: op.Condition,
+		Required:  !op.BestEffort,
+	}, waitingFor)
+}
+
+// execRunPreStart runs the service's pre_start hooks against the designated
+// replica, through the same primitive as the imperative engine.
+func (exec *planExecutor) execRunPreStart(ctx context.Context, node *PlanNode) error {
+	ctr, err := exec.startTarget(node)
+	if err != nil {
+		return err
+	}
+	return exec.compose.runPreStart(ctx, exec.project, *node.Operation.Service, ctr, exec.listener)
+}
+
+// execRunPostStart runs the service's post_start hooks inside the started
+// container, through the same primitive as the imperative engine.
+func (exec *planExecutor) execRunPostStart(ctx context.Context, node *PlanNode) error {
+	ctr, err := exec.startTarget(node)
+	if err != nil {
+		return err
+	}
+	for _, hook := range node.Operation.Service.PostStart {
+		if err := exec.compose.runHook(ctx, ctr, *node.Operation.Service, hook, exec.listener); err != nil {
+			return err
+		}
+	}
+	return nil
 }
 
 func (exec *planExecutor) execStopContainer(ctx context.Context, op Operation) error {
--- a/pkg/compose/reconcile.go
+++ b/pkg/compose/reconcile.go
@@ -944,7 +944,8 @@ func (r *reconciler) planStartPhase() error {
 // materialized by a create-phase node (create) or already observed (container).
 type startReplica struct {
 	resID  string
-	number int // replica number, the start order within the service
+	number int    // replica number, the start order within the service
+	name   string // container name, for progress events
 	// container is the observed container to start; nil when the create
 	// phase materializes it.
 	container *container.Summary
@@ -964,7 +965,7 @@ type startReplica struct {
 // create node), and an exceptional-state restart (paused, dead, ...) plans
 // no create at all — the observed container sits on the node itself.
 func plannedReplica(resID string, number int, node *PlanNode) startReplica {
-	rep := startReplica{resID: resID, number: number, after: node}
+	rep := startReplica{resID: resID, number: number, name: node.Operation.Name, after: node}
 	switch node.Operation.Type {
 	case OpCreateContainer:
 		rep.createNodeID = node.ID
@@ -974,6 +975,7 @@ func plannedReplica(resID string, number int, node *PlanNode) startReplica {
 		// exceptional-state restart (paused, dead, ...): that registration
 		// always carries the observed container on the node itself
 		rep.container = node.Operation.Container
+		rep.name = getCanonicalContainerName(*node.Operation.Container)
 	default:
 		// no other node type is registered in containerNodes today; leave
 		// the target unresolved so execution fails with a clean "no
@@ -1009,7 +1011,7 @@ func (r *reconciler) startPhaseReplicas(service types.ServiceConfig) (replicas [
 			continue
 		}
 		seen[resID] = true
-		replicas = append(replicas, startReplica{resID: resID, number: oc.Number, container: &oc.Summary})
+		replicas = append(replicas, startReplica{resID: resID, number: oc.Number, name: getCanonicalContainerName(oc.Summary), container: &oc.Summary})
 	}
 	for resID, node := range r.containerNodes {
 		if !seen[resID] && strings.HasPrefix(resID, fmt.Sprintf("service:%s:", service.Name)) {
@@ -1113,6 +1115,7 @@ func (r *reconciler) planServiceStart(service types.ServiceConfig) error {
 			Cause:        "pre_start hooks",
 			Service:      &serviceCopy,
 			Container:    first.container,
+			Name:         first.name,
 			CreateNodeID: first.createNodeID,
 		}
 		deps := prev
@@ -1136,6 +1139,7 @@ func (r *reconciler) planServiceStart(service types.ServiceConfig) error {
 			Cause:        "start",
 			Service:      &serviceCopy,
 			Container:    rep.container,
+			Name:         rep.name,
 			CreateNodeID: rep.createNodeID,
 		}
 		deps := slices.Clone(prev)
@@ -1152,6 +1156,7 @@ func (r *reconciler) planServiceStart(service types.ServiceConfig) error {
 				Cause:        "post_start hooks",
 				Service:      &serviceCopy,
 				Container:    rep.container,
+				Name:         rep.name,
 				CreateNodeID: op.CreateNodeID,
 			}, group, start)
 			post.Phase = PhaseStart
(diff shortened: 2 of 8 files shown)


pkg/compose/reconcile.go
```diff
@@ -877,6 +918,209 @@ func (r *reconciler) hasVolumeMismatch(expected types.ServiceConfig, oc Observed
...
+		}
+		seen[resID] = true
+		replicas = append(replicas, startReplica{resID: resID, container: &oc.Summary})
+	}
+	for resID, node := range r.containerNodes {
+		if !seen[resID] && strings.HasPrefix(resID, fmt.Sprintf("service:%s:", service.Name)) {
+			replicas = append(replicas, startReplica{resID: resID, create: node})
+		}
+	}
+	sort.Slice(replicas, func(i, j int) bool { return replicas[i].resID < replicas[j].resID })
```
docker-agent: **[medium] Lexicographic sort of replica resIDs produces wrong start order for services with 10+ replicas**

`startPhaseReplicas` sorts replicas by `resID` string using `<`, which is a lexicographic comparison. Because the resID format is `"service:<name>:<number>"`, the numeric suffix is compared as a string. For a service with 11 replicas this means:

```
"service:app:1" < "service:app:10" < "service:app:11" < "service:app:2" < ... < "service:app:9"
```

So replica 10 starts before replicas 2–9, breaking the sequential start-order guarantee the PR explicitly documents. Docker Compose supports `scale: 10` (and `deploy.replicas`) so this affects real workloads.

Fix: extract the integer suffix from `resID` and compare numerically:

```suggestion
	sort.Slice(replicas, func(i, j int) bool {
		ni := replicas[i].resID[strings.LastIndexByte(replicas[i].resID, ':')+1:]
		nj := replicas[j].resID[strings.LastIndexByte(replicas[j].resID, ':')+1:]
		if len(ni) != len(nj) {
			return len(ni) < len(nj)
		}
		return ni < nj
	})
```

Alternatively, store the numeric replica number in `startReplica` directly and compare that field.

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 100/100 |

docker-agent (commented):

pkg/compose/reconcile.go
ndeloof: Good catch — fixed by sorting on the numeric replica number (new `number` field on `startReplica`, populated from the observed container number or parsed from the resID), with a regression test starting 11 replicas (`TestPlanStart_ReplicaOrderIsNumeric`). Folded into the first commit, now d0056878f.

docker-agent: Perfect fix — the `number` field on `startReplica` gives you clean numeric sorting without repeated string parsing. The test covering 11 replicas is exactly right; it would have caught the lexicographic bug immediately (replica 10 starting before 2–9).

Thanks for the quick turnaround!

pkg/compose/reconcile.go
```diff
@@ -877,6 +919,221 @@ func (r *reconciler) hasVolumeMismatch(expected types.ServiceConfig, oc Observed
...
+// phase acts on. anyRunning reports whether a replica was running at
+// observation and left untouched by the plan, which gates pre_start.
+func (r *reconciler) startPhaseReplicas(service types.ServiceConfig) (replicas []startReplica, anyRunning bool) {
+	seen := map[string]bool{}
+	for i := range r.observed.Containers[service.Name] {
+		oc := &r.observed.Containers[service.Name][i]
+		resID := fmt.Sprintf("service:%s:%d", service.Name, oc.Number)
+		if node, planned := r.containerNodes[resID]; planned {
+			seen[resID] = true
+			replicas = append(replicas, startReplica{resID: resID, number: oc.Number, name: node.Operation.Name, create: node})
```
docker-agent: **[HIGH] `startPhaseReplicas` treats `OpStartContainer` nodes as create nodes, causing `startTarget` to fail for paused/dead containers**

When `reconcileService` encounters a container in an exceptional state (`paused`, `dead`, etc.), it plans an `OpStartContainer` node for it and stores that node in `r.containerNodes`:

```go
// reconcileService default case (around line 770)
lastNode = r.plan.addNode(Operation{
    Type:      OpStartContainer,
    Container: &containers[i].Summary, // Container IS set on the node
    // Name is NOT set
}, "", infraDeps...)
r.containerNodes[...] = lastNode  // stored as if it were a create node
```

`startPhaseReplicas` then finds this node via the `containerNodes` lookup and creates a `startReplica` with `create = <OpStartContainer node>`:

```go
if node, planned := r.containerNodes[resID]; planned {
    replicas = append(replicas, startReplica{
        ...,
        name:   node.Operation.Name,  // empty — Name not set for this case
        create: node,                 // points to OpStartContainer, NOT OpCreateContainer
    })
}
```

`planServiceStart` then sets `op.CreateNodeID = first.create.ID` (the `OpStartContainer` node's ID) and `op.Container = first.container` (nil, since `startReplica.container` is not copied from `node.Operation.Container`). At execution time, `startTarget` finds `op.Container == nil` and calls `exec.pctx.get(op.CreateNodeID)` — but `pctx` is only populated by `execCreateContainer`, never by `execStartContainer`. The result has `ContainerID == ""`, so:

```
internal: service:app:1 has no materialized container to act on
```

**Impact**: Any `ScopeCreateStart` or `ScopeStart` plan that includes a paused or dead container will fail at runtime with this internal error. This is a regression — the imperative engine (start.go) did not suffer this because it re-queried the daemon for container state rather than relying on `pctx`.

**Fix**: In `startPhaseReplicas`, distinguish `OpStartContainer` nodes (which carry the container in `node.Operation.Container`) from create-type nodes. When the planned node is an `OpStartContainer`, treat the replica as an observed container (copy `container: node.Operation.Container`) rather than as a create-path replica:

```suggestion
		if node, planned := r.containerNodes[resID]; planned {
			seen[resID] = true
			if node.Operation.Type == OpStartContainer {
				// Exceptional-state start (paused/dead): the node carries the
				// observed container; treat it as an observed (non-create) replica.
				replicas = append(replicas, startReplica{resID: resID, number: oc.Number, name: getCanonicalContainerName(oc.Summary), container: &oc.Summary})
			} else {
				replicas = append(replicas, startReplica{resID: resID, number: oc.Number, name: node.Operation.Name, create: node})
			}
			continue
		}
```

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 100/100 |

docker-agent (commented):
### Assessment: 🔴 CRITICAL

One high-severity bug was found in the start-phase planning code introduced by this PR.

pkg/compose/reconcile.go
ndeloof: Confirmed — and the same class of bug also hit recreated replicas: containerNodes registers the recreate chain's *rename* node, whose ID never lands in pctx either (only execCreateContainer stores a result). Fixed by restructuring startReplica: `after` carries the ordering edge to the registered node, and `createNodeID` resolves per node type (create → its own ID, rename → its CreateNodeID, exceptional-state start → none, the observed container is used directly). Regression tests: TestPlanStart_RecreatedReplicaTargetsCreateNode and TestPlanStart_ExceptionalStateReplicaKeepsObservedTarget. Folded into the planning commit, now 6bf3baf0f.

docker-agent (commented):
### Assessment: 🟢 APPROVE

The plan engine start phase implementation looks solid. The DAG-based approach to `OpWaitCondition`, `OpRunPreStart`, `OpRunPostStart`, and the replica chain ordering is well-structured. The zero-value `ReconcileOptions.Scope` preserves byte-identical behavior for existing create-only plans (confirmed by the 42 golden tests). No high or medium confidence bugs found in the changed code.

- [medium] pkg/compose/reconcile.go:1119 — `pre_start` node placed in replica-1's start group causes `Starting` event to fire during pre_start hook, not during actual container start (confidence: weak 52/100)

  The `preStart` node is assigned the same group name as replica-1's container start node (`fmt.Sprintf("start%s", strings.TrimPrefix(first.resID, "service"))`). In `buildGroupTracker`, `onNodeStart` fires `Starting` when the first node in a group starts (`gs.started == 1`). With `pre_start` in the same group as `start` and `post_start`, the `Starting` progress event fires when the pre_start hook begins execution — potentially seconds before the container actually starts. The `doneText = "Started"` comment in the new code says this is intentional ("Started is emitted after the hooks ran — word-for-word the imperative sequence"), but the `Starting` signal fires earlier than users might expect. Worth a conscious decision on whether the group boundary should begin at the hook or at the container start.

docker-agent (commented):
### Assessment: 🟢 APPROVE

The plan engine changes (OpWaitCondition, OpRunPreStart, OpRunPostStart, PlanNode.Phase, ReconcileOptions.Scope) are well-structured. The drafter identified one potential slice alias hazard in `planServiceStart` around the pre_start dep construction, but verification confirmed it is self-contained and harmless: the original `prev` slice is overwritten with a fresh single-element slice on line 1121 immediately after, so any backing-array mutation from the conditional `append` is never observable by subsequent code. The replica loop correctly uses `slices.Clone(prev)` on the updated slice. No real bugs introduced by this PR.

docker-agent (commented):
This review covers only the commits pushed since `ba3c16642ed0fd78dba6dee06ce687523c02c857`.

### Assessment: 🟢 APPROVE

The refactor cleanly splits `create()` into `preparePlan()` and a thin caller. All error paths in `preparePlan` return `nil, nil, nil, err` consistently, the relative ordering of every step is preserved, and `emitRunningEvents` is correctly moved to the caller in `create()` — same position in the call sequence as before. No bugs introduced by this incremental change.

pkg/compose/reconcile.go
```diff
@@ -877,6 +919,249 @@ func (r *reconciler) hasVolumeMismatch(expected types.ServiceConfig, oc Observed
...
+func plannedReplica(resID string, number int, node *PlanNode) startReplica {
+	rep := startReplica{resID: resID, number: number, name: node.Operation.Name, after: node}
+	switch node.Operation.Type {
+	case OpCreateContainer:
+		rep.createNodeID = node.ID
+	case OpRenameContainer:
+		rep.createNodeID = node.Operation.CreateNodeID
+	default:
+		rep.container = node.Operation.Container
+		rep.name = getCanonicalContainerName(*node.Operation.Container)
```
docker-agent, line 978: **[medium] Nil pointer dereference in `plannedReplica` default branch — `Container` not nil-checked before dereference**

The `default` branch of `plannedReplica` dereferences `node.Operation.Container` unconditionally:

```go
default:
    rep.container = node.Operation.Container
    rep.name = getCanonicalContainerName(*node.Operation.Container)  // panics if Container == nil
```

The comment on the function documents the invariant ("the observed container sits on the node itself"), but nothing enforces it at runtime. Currently the only operation type reaching this branch is `OpStartContainer` (registered in `containerNodes` at line 776 with a non-nil `Container`), so this is safe today. However, as new operation types are added to the plan engine, any future node registered in `containerNodes` with a nil `Container` and an unrecognised type will panic here.

**Suggested fix:** add a nil guard before the dereference, or an explicit `panic`/`log` assertion to make the invariant visible:

```go
default:
    if node.Operation.Container == nil {
        panic(fmt.Sprintf("plannedReplica: unexpected nil Container for op %s on %s", node.Operation.Type, resID))
    }
    rep.container = node.Operation.Container
    rep.name = getCanonicalContainerName(*node.Operation.Container)
```

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 57/100 |

docker-agent (commented):

One medium-severity finding: a latent nil pointer dereference in newly introduced code. All other drafter hypotheses (slice aliasing in pre-start deps, missing OpRunPreStart/OpRunPostStart in event switch statements) were verified to be false positives — either the code flow is safe by construction or the cases are unreachable.

_(none — all below-threshold findings were dismissed)_

pkg/compose/reconcile.go
ndeloof: Fair point. Rather than a runtime panic, the switch is now exhaustive over the three node types containerNodes registers (create, rename, exceptional-state start — the dereference lives in the OpStartContainer case, whose registration always carries the container), and the default branch leaves the target unresolved with a comment: a future unregistered type then fails at execution with the clean internal "no materialized container" error instead of panicking during planning. Folded into the planning commit, now db22082be.

docker-agent: Excellent solution — making the switch exhaustive and deferring the error to execution is cleaner than a runtime panic. Leaving the target unresolved in the default case preserves the internal invariant check while keeping the plan-time logic defensive. Thanks for the refinement!

pkg/compose/executor_ops.go
```diff
@@ -127,13 +128,99 @@ func (exec *planExecutor) execCreateContainer(ctx context.Context, node *PlanNod
...
+	op := node.Operation
+	exec.containersMu.Lock()
+	waitingFor := exec.containersByService[op.Name].filter(isNotOneOff)
+	exec.containersMu.Unlock()
+	if len(waitingFor) == 0 {
+		if op.BestEffort {
+			exec.compose.events.On(skippedEvent("Service "+op.Name, "no container to wait for"))
+			return nil
+		}
+		return fmt.Errorf("required dependency %q has no container to wait for", op.Name)
```
docker-agent, line 190: **[medium] `execWaitCondition` error path emits internal resource ID instead of user-friendly service name**

When a required `OpWaitCondition` node fails, the error propagates to `groups.onNodeError` → `emitErrorEvent`, which uses `op.ResourceID` (`"wait:db:service_healthy"`) as the progress-UI event ID. The best-effort skip path two lines above correctly uses `"Service " + op.Name`:

```go
// best-effort skip: user sees "Service db: Skipped: no container to wait for"
exec.compose.events.On(skippedEvent("Service "+op.Name, "no container to wait for"))
return nil

// required fail: error propagates; emitErrorEvent fires with ID="wait:db:service_healthy"
return fmt.Errorf("required dependency %q has no container to wait for", op.Name)
```

Because `OpWaitCondition` nodes are always ungrouped (empty `Group`), `onNodeError` delegates to `emitErrorEvent`. The default branch there (`id = op.ResourceID`) produces `"wait:db:service_healthy: required dependency "db" has no container to wait for"` in the UI — an internal key, not a display name.

**Fix**: emit a named error event before returning, mirroring the skip path:
```go
exec.compose.events.On(api.Resource{ID: "Service " + op.Name, Status: api.Error, Text: "no container to wait for"})
return fmt.Errorf("required dependency %q has no container to wait for", op.Name)
```

Or extend `emitErrorEvent`'s switch with a case for `op.Name != ""` (when `Container` is nil but `Name` is set).

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 75/100 |

docker-agent (commented):

This PR introduces the start-phase plan vocabulary and executor operations (OpWaitCondition, OpRunPreStart, OpRunPostStart) along with the ScopeCreateStart/ScopeStart reconcile scopes. The implementation is well-structured, with comprehensive golden tests confirming byte-identical output for the existing create-only scope. One medium-confidence issue was found in the new wait-condition error path.

docker-agent (commented):
### Assessment: 🟢 APPROVE

pkg/compose/executor_ops.go
ndeloof: Fixed (folded into the executor commit, now 86f5d4235): emitErrorEvent names an OpWaitCondition failure "Service <dep>", matching the Waiting/Skipped vocabulary the wait itself emits — never the internal wait:<dep>:<condition> ID.

docker-agent (commented):
### Assessment: 🟢 APPROVE

The start-phase plan engine (Lot 1 of #14081) is implemented correctly. The core invariants hold:

- **OpWaitCondition deduplication**: shared wait nodes per *(dep, condition)* pair, with correct BestEffort→required upgrade path (single-threaded planning, safe mutation).
- **Pre_start gating**: `anyRunning` correctly reflects observed running replicas left untouched by the plan, consistent with the imperative engine's `len(serviceContainers)==len(toStart)` guard.
- **Replica chain ordering**: numeric sort on the `number` field (not string/lexicographic), regression-tested with 11 replicas (TestPlanStart_ReplicaOrderIsNumeric).
- **startTarget resolution**: cleanly routes observed containers (op.Container), recreated containers (pctx.get(op.CreateNodeID) via rename node's CreateNodeID), and exceptional-state restarts (observed container passed through plannedReplica's OpStartContainer case).
- **Thread safety**: reconciliationContext (pctx) and containersByService are both mutex-protected; execWaitCondition's filter() returns an independent copy before unlocking, so waitDependency holds no reference into the shared map.
- **Scope zero-value**: ScopeCreate (iota=0) skips planStartPhase entirely, preserving byte-identical plan output for all 42 golden tests.
- **Event routing**: OpWaitCondition nodes are ungrouped and self-emitting (Waiting/Healthy via waitDependency); OpRunPreStart/OpRunPostStart share the replica's start group so "Started" fires only after post_start hooks complete — matching the imperative sequence word-for-word.
- **Switch exhaustiveness**: the default case in plannedReplica and executeNode both fail cleanly without panicking on unknown types.

PR #14159, opened 2026-09-01
fix(plugins): honor PATHEXT for provider lookup on Windows

### LINKED ISSUE #14157: Compose provider extensions/plugins can only execute ".exe" on windows
The Windows plugin discovery currently appears to restrict plugins to .exe files. I'm not sure this is desirable behavior.

https://github.com/docker/compose/blob/b3284911a304910035d1b9a5203f7b50190e1474/pkg/compose/plugins_windows.go#L22

On Windows, executable command resolution normally respects PATHEXT, which allows extensions such as .CMD, .BAT, and potentially .PY (depending on the environment) to be invoked without explicitly specifying the extension.

Restricting Compose plugins to .exe therefore prevents lightweight script-based plugins. For example, a plugin could be implemented as a Python, batch, or other script rather than requiring a compiled executable.

I think plugin discovery should probably respect PATHEXT (or otherwise use the same executable-resolution semantics as Windows), rather than hard-coding .exe.

This would also make Compose plugin behavior more consistent with how commands are normally resolved on Windows.

ndeloof:
The PATH fallback of provider discovery appended a hardcoded `.exe` on Windows, restricting providers to compiled binaries. `exec.LookPath` already implements Windows executable resolution (PATHEXT), which finds `.exe` but also `.com`/`.bat`/`.cmd` scripts — the hardcoded suffix only narrowed the lookup. This drops the `executable()` indirection and looks up the bare provider name.

Only the PATH fallback changes: the docker CLI plugin channel (`docker-<name>`) keeps requiring `.exe` on Windows, per docker/cli plugin candidate rules. Note that scripts relying on file associations (e.g. `.py`) still won't launch: `CreateProcess` does not honor them — PATHEXT support effectively covers what Windows can execute directly.

Argument-injection concerns for batch files (BatBadBut, CVE-2024-3566) are mitigated by the Go runtime, which refuses to run `.bat`/`.cmd` with arguments it cannot safely escape.

Fixes #14157


docker-agent (commented):

The PR correctly drops the `executable()` helper and lets `exec.LookPath` use Windows PATHEXT semantics. Unix behavior is unchanged. One medium-severity security finding: PATHEXT resolution now allows `.bat`/`.cmd` scripts to match before `.exe` binaries when an earlier PATH directory contains a same-named script — the old `.exe`-only lookup blocked this. The Go runtime's BatBadBut mitigation covers argument-injection escaping, not script discovery.

pkg/compose/plugins.go
```diff
@@ -198,7 +198,11 @@ func (s *composeService) getPluginBinaryPath(provider string) (path string, err
 		path = plugin.Path
 	}
 	if errdefs.IsNotFound(err) {
-		path, err = exec.LookPath(executable(provider))
+		// A plain LookPath honors Windows executable-resolution semantics
+		// (PATHEXT: .exe, but also .com/.bat/.cmd), so provider executables
+		// need not be compiled binaries. Callers must not append a hardcoded
+		// ".exe": it restricts the lookup instead of helping it.
+		path, err = exec.LookPath(provider)
```
docker-agent, line 205: **[medium/security] PATHEXT-resolved .bat/.cmd scripts can be executed on Windows instead of binaries**

The change from `exec.LookPath(executable(provider))` (which appended `.exe` on Windows) to `exec.LookPath(provider)` means Windows PATHEXT resolution now applies. On Windows, `exec.LookPath` searches PATH directories against the full PATHEXT list (`.COM;.EXE;.BAT;.CMD`). If an attacker — or an inadvertent script — places a `provider.bat` or `provider.cmd` in a PATH directory that appears *before* the directory containing `provider.exe`, the batch file will be executed instead of the intended binary.

The PR description notes that the Go runtime mitigates BatBadBut/CVE-2024-3566 (argument-injection in `.bat`/`.cmd` files). That mitigation covers *argument escaping*, not *discovery*: the batch file is still found and launched; the Go runtime simply refuses the call if it cannot safely escape the arguments. In practice, `getPluginBinaryPath` is called with no arguments to the provider itself at LookPath time, so the escape guard may not trigger at all — leaving the wrong binary executing silently.

The old `.exe`-only lookup fully prevented `.bat`/`.cmd` execution. If the intent is to support `.bat`/`.cmd` providers intentionally, consider documenting the accepted risk and the PATHEXT ordering that a well-configured system would need. If only compiled binaries should run, filter by extension after `LookPath`:

```go
if ext := strings.ToLower(filepath.Ext(path)); ext == ".bat" || ext == ".cmd" {
    return "", fmt.Errorf("provider %q resolved to a script (%s); only .exe providers are supported", provider, ext)
}
```

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 97/100 |

glours (approved):

PR #14161, opened 2026-09-01
cmd/prompt: remove uses of github.com/AlecAivazis/survey/v2

thaJeztah:
- taken from https://github.com/docker/compose/pull/13900

### cmd/prompt: remove uses of github.com/AlecAivazis/survey/v2

The github.com/AlecAivazis/survey/v2 module was archived and is no longer
maintained. Looking at the code, we didn't really use most of its features;
survey mostly provided terminal input handling around a simple yes/no
question.

Replace the interactive confirmation implementation with a small local
prompt reader while keeping the exported prompt types and existing
non-terminal prompt behavior unchanged. Preserve the confirmation hint,
default handling, retries for invalid input, and Ctrl+C handling.

Interactive prompts continue to use raw terminal input so Ctrl+C is handled
by the prompt instead of requiring a second interrupt to terminate the
process.

**What I did**

**Related issue**


cmd/prompt/prompt.go
```diff
@@ -100,3 +65,18 @@ func (u Pipe) Confirm(message string, defaultValue bool) (bool, error) {
 	_, _ = fmt.Fscanln(u.stdin, &answer)
 	return utils.StringToBool(answer), nil
 }
+
+func confirm(stdin *bufio.Reader, stdout io.Writer, message string, defaultValue bool) (bool, error) {
+	_, _ = fmt.Fprint(stdout, message)
```
docker-agent: **[medium] confirm() does not print [y/N] hint; behavioral regression from survey.Confirm**

The new `confirm()` function writes `message` verbatim to stdout with no indication of valid responses or the current default:

```go
_, _ = fmt.Fprint(stdout, message)
```

`survey.Confirm` automatically appended `(y/N)` or `(Y/n)` (depending on `defaultValue`) to every prompt. The new implementation omits this entirely. Unless every caller already embeds the hint in its own message string, users running in a terminal will see a bare prompt with no cue that `y`/`n` are the expected inputs or which value is the default.

Consider appending the hint in `confirm()` itself so callers don't have to know about this detail:

```suggestion
	if defaultValue {
		_, _ = fmt.Fprintf(stdout, "%s [Y/n] ", message)
	} else {
		_, _ = fmt.Fprintf(stdout, "%s [y/N] ", message)
	}
```

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 70/100 |

cmd/prompt/prompt.go
```diff
@@ -100,3 +65,18 @@ func (u Pipe) Confirm(message string, defaultValue bool) (bool, error) {
...
+	answer, err := stdin.ReadString('\n')
+	if err != nil && err != io.EOF {
+		return false, err
+	}
+
+	answer = strings.TrimSpace(answer)
+	if answer == "" {
+		return defaultValue, nil
+	}
+	return utils.StringToBool(answer), nil
```
docker-agent: **[medium] Invalid user input silently returns false instead of re-prompting**

When the user types an unrecognised string (e.g. `"nno"`, `"yyes"`, `"maybe"`), `utils.StringToBool` returns `false` and the function returns without any error or re-prompt:

```go
return utils.StringToBool(answer), nil
```

`StringToBool` accepts `"y"`, `"yes"`, `"true"`, `"1"` as true and maps everything else to false. `survey.Confirm` looped until it received a valid answer. The new single-pass read means a mistype on a destructive confirmation (e.g. "Remove all volumes?") silently becomes a denial (`false`), giving the user no feedback that their input was not understood.

Consider validating the answer and returning an error or looping on unrecognised input. A minimal fix:

```go
result, ok := parseBool(answer)
if !ok {
    return false, fmt.Errorf("invalid input %q: expected y/yes/n/no", answer)
}
return result, nil
```

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 100/100 |

docker-agent (commented):

This PR cleanly removes the archived `survey/v2` dependency and replaces it with a minimal local line-based implementation. The overall approach is sound. Two behavioral regressions in the new `confirm()` function are worth addressing before merge.

cmd/prompt/prompt.go
```diff
@@ -100,3 +65,18 @@ func (u Pipe) Confirm(message string, defaultValue bool) (bool, error) {
 	_, _ = fmt.Fscanln(u.stdin, &answer)
 	return utils.StringToBool(answer), nil
 }
+
+func confirm(stdin *bufio.Reader, stdout io.Writer, message string, defaultValue bool) (bool, error) {
+	_, _ = fmt.Fprint(stdout, message)
+
+	answer, err := stdin.ReadString('\n')
```
glours: This blocking read has no way to observe cancellation. Since `AdaptCmd`'s SIGINT handler (`cmd/compose/compose.go:112`) now intercepts Ctrl+C to cancel the context instead of letting the OS kill the process, a single Ctrl+C at a prompt no longer aborts it,  the read stays blocked here and a second Ctrl+C is needed to actually terminate. 
`survey`'s raw-mode reader handled this internally (ISIG disabled, Ctrl+C caught as a keypress).

glours (changes requested):
I confirm the 2 catches made by @docker-agent and found a 3rd one

thaJeztah:
Ah, yes, I need to look at this one; I thought I'd move this separate, but didn't look too closely yet 😂

docker-agent (commented):
### Assessment: 🟢 APPROVE

The PR correctly removes the archived `github.com/AlecAivazis/survey/v2` dependency and replaces it with `golang.org/x/term`. The core `Confirm` implementation is structurally sound: `SetRawTerminal` / `defer RestoreTerminal` is safe, `term.NewTerminal` usage is idiomatic, the error propagation path is correct, and `go.mod` properly promotes `golang.org/x/term` from indirect to direct. The new tests cover the happy path, invalid-input retry, and Ctrl+C interruption.

- [medium → verified low] `cmd/prompt/prompt.go:94` — Ctrl+C handling relies on `term.ReadLine` returning `io.EOF` on byte `0x03`; if `golang.org/x/term` changes this behaviour in a future release the loop would spin indefinitely until the TTY is closed. Currently working as intended in v0.45.0 — the verifier assessed this as speculative/low, score 40/100 (🟠 weak). No action required now; worth a comment in the code or a version note in `go.mod` if the team wants to guard against it long-term.

cmd/prompt/prompt.go
thaJeztah: That's already handled by callers, which include it, so we just print verbatim; e.g.;

```
msg := "\nDo you want to proceed with these variables? [Y/n]: "
```

docker-agent: You're right, thanks for the clarification. I see that both call sites in options.go already include the hint in their message strings (line 166 uses [Y/n] and line 290 uses [y/N]), so the verbatim printing is intentional and there's no regression here.

I should have checked all the callers before flagging this. Apologies for the noise!

thaJeztah:
> if golang.org/x/term changes this behaviour in a future release the loop would spin indefinitely until the TTY is closed. Currently working as intended in v0.45.0 

Yeah, not fixing that; we have tests, and they shouldn't break behavior.

thaJeztah:
Rebased; probably should be OK for review now 😅

ndeloof (commented):
Legitimate change, worth landing: survey/v2 is archived, compose only used a yes/no out of it, and this drops five transitive dependencies while promoting an already-present `x/term`. Exported API intact, non-terminal `Pipe` path untouched, the pty-backed tests cover the right level, deferred `RestoreTerminal` covers error paths, and the retry/Enter-default semantics match survey.

One real compatibility regression and two smaller notes, inline. On top of those, one thing CI cannot tell us: the interactive path now goes through streams `SetRawTerminal` + `x/term.Terminal` on a Windows console, a different plumbing than survey's (`go-colorable`); the pty tests only exercise unix. Worth one manual check on Windows before merge, or a note in the description.

cmd/prompt/prompt.go
```diff
@@ -74,17 +75,34 @@ func (s streamsFileReader) Fd() uintptr {
 
 // Confirm asks for yes or no input
 func (u User) Confirm(message string, defaultValue bool) (bool, error) {
```
ndeloof, line 56: **The "(y/N)" hint survey auto-appended is lost, and three call sites relied on it** — the description's "current callers already include the confirmation hint" holds for the two prompts in `cmd/compose/options.go` only. These don't carry any hint in their message:

- `pkg/compose/publish.go` — "Are you ok to publish these bind mount declarations?" and "…these sensitive data?" (via `confirmOrCancel`);
- `pkg/bridge/convert.go` — "Output directory … will be permanently deleted. Continue?".

After this change those render as a bare question: the user no longer sees the expected input format nor the default — on *destructive* confirmations. Minimal fix that preserves every caller at once: have `Confirm` append `" [y/N]: "` / `" [Y/n]: "` (per `defaultValue`) when the message doesn't already end with a hint; alternatively, fix the three messages in this same PR.

cmd/prompt/prompt.go
```diff
@@ -74,17 +75,34 @@ func (s streamsFileReader) Fd() uintptr {
...
+		Writer: u.stdout,
+	}, message)
+
+	for {
+		answer, err := terminal.ReadLine()
+		if err != nil {
+			return false, err
+		}
+
+		switch strings.ToLower(answer) {
```
ndeloof: nit: `strings.ToLower` without `strings.TrimSpace` — a ` y ` (stray whitespace) silently loops back to the prompt, where survey tolerated it. One-liner: `switch strings.ToLower(strings.TrimSpace(answer))`.

cmd/prompt/prompt_test.go
```diff
@@ -0,0 +1,137 @@
...
+	}()
+
+	readUntil(t, ptmx, "Continue? ")
+
+	_, err = ptmx.Write([]byte{3}) // Ctrl+C
+	assert.NilError(t, err)
+
+	select {
+	case err := <-done:
+		assert.ErrorIs(t, err, io.EOF)
```
ndeloof: Behavior note this test locks in: Ctrl+C now surfaces as `io.EOF` (that's `x/term`'s mapping), where survey returned `InterruptErr` ("interrupt"). No compose caller special-cased `InterruptErr` (they all `return err`), so nothing breaks logically — but the user-facing failure message degrades to "EOF", and ^C becomes indistinguishable from ^D by construction. Non-blocking suggestion: intercept `0x03` in the wrapping Reader and return a dedicated error ("prompt interrupted by user"), which also keeps the door open for a proper exit 130 later.

cmd/prompt/prompt.go
glours: If you go with option 2 (fixing individual messages rather than the centralized approach), don't miss two more call sites through the same `Confirm` with the same bare-question issue:

- `pkg/compose/publish.go:619` — `buildEnvPromptMessage`: "...Are you ok to publish these env declarations?"
- `pkg/compose/publish.go:629` — `buildConfigContentPromptMessage`: "...Are you ok to publish these config contents?"

thaJeztah:
🤔 Hm... 

```
(Reading database ... 201598 files and directories currently installed.)
Purging configuration files for docker-ce (5:28.0.4-1~ubuntu.24.04~noble) ...
dpkg: warning: while removing docker-ce, directory '/usr/libexec/docker' not empty so not removed
Purging configuration files for containerd.io (2.3.4-2~ubuntu.24.04~noble) ...
Reading package lists...
Building dependency tree...
Reading state information...
curl is already the newest version (8.5.0-2ubuntu10.13).
0 upgraded, 0 newly installed, 0 to remove and 1 not upgraded.
# Executing docker install script, commit: 4e16be805b3af595aa448d1a4def857841f1c62c
+ sh -c apt-get -qq update >/dev/null
E: Failed to fetch https://dl.google.com/linux/chrome-stable/deb/dists/stable/main/binary-amd64/Packages.gz  Hash Sum mismatch
   Hashes of expected file:
    - Filesize:1405 [weak]
    - SHA256:233e56de019b57db89238fa7bcc3647718dbbea3a40c2dc1c633a8c8952aa9e9
    - SHA1:f25f722f552dc3fd510081e6b65a0e7174ec9321 [weak]
    - MD5Sum:2fc39e32c7bdf718a907c390338392ea [weak]
   Hashes of received file:
    - SHA256:bc1428ab27c6d76ee9bb76de07f1ded0ddb4aaabd958fc72855634ef5894a4b3
    - SHA1:f601c49edc3ea059583213f0be1589cb0fa6e2e5 [weak]
    - MD5Sum:2ce0a26fc7a49a488341fc13c6d05b52 [weak]
    - Filesize:1405 [weak]
   Last modification reported: Wed, 09 Sep 2026 09:41:12 +0000
   Release file created at: Wed, 09 Sep 2026 17:16:59 +0000
E: Some index files failed to download. They have been ignored, or old ones used instead.
Error: Process completed with exit code 100.
```

glours (approved):
LGTM

thaJeztah:
> LGTM

Three time's a charm 😂

glours:
>Three time's a charm 😂

Ho who knows, maybe @ndeloof will find something I missed 😂

PR #14164, opened 2026-09-02
build: bump golangci-lint to v2.13.2

ndeloof:
Bumps the pinned golangci-lint from v2.11.3 to the latest release, v2.13.2.

- `gomodguard` is deprecated since v2.13 (warns on every run): migrated to `gomodguard_v2`, same blocked modules in the new flat-list schema.
- The newer staticcheck flags SA4023 in `watcher_naive.go`: tilt-dev/fsnotify's `SetRecursive` only succeeds on Windows, so on any single GOOS the `err == nil` comparison is provably constant. The cross-platform code is correct — silenced with a documented `//nolint`.


docker-agent (commented):
### Assessment: 🟢 APPROVE

This PR is a routine toolchain bump (golangci-lint v2.11.3 → v2.13.2) with a config migration and a nolint suppression. No bugs were found in the changed code with sufficient confidence for an inline comment.

- [medium] `.golangci.yml`:83 — gomodguard_v2 flat schema may silently drop or misinterpret version constraint for gotest.tools (confidence: 🟠 weak 31/100)
  The `gotest.tools` entry carries `version: < 3.0.0` in the new flat blocked list. Whether `gomodguard_v2` honours the `version` field as a constraint (banning pre-3.0.0 imports only) or treats it as an unrecognised key and blocks all imports of `gotest.tools` regardless of version could not be confirmed from the diff alone. Similarly, `github.com/distribution/distribution` moved from the old `versions:` sub-section to a plain entry with only `reason:` — the intent looks right but is worth a quick check against the gomodguard_v2 schema docs. Low evidence strength because the linter schema docs were not available for verification.

glours (approved):
LGTM

pkg/watch/watcher_naive.go
```diff
@@ -309,8 +309,11 @@ func newWatcher(paths []string) (Notify, error) {
 	}
 	MaybeIncreaseBufferSize(fsw)
 
-	err = fsw.SetRecursive()
-	isWatcherRecursive := err == nil
+	// SetRecursive only succeeds on Windows; every other platform returns
+	// "Not supported" and the non-recursive fallback below is taken. On a
+	// single GOOS staticcheck proves the comparison constant (SA4023).
+	err = fsw.SetRecursive()         //nolint:staticcheck
+	isWatcherRecursive := err == nil //nolint:staticcheck
```
thaJeztah, line 316: How about;

```go
	var isWatcherRecursive bool
		// Only supported on Windows.
		if err := fsw.SetRecursive(); err != nil {
			return nil, err
		}
		isWatcherRecursive = true
	}
```


- on non-Windows; it's not supported, so don't attempt
- on Windows, it should always pass, but _technically_ could fail if the watcher was already used; in that case return an error (we could even `panic`);

```go
func (w *Watcher) SetRecursive() error {
	if len(w.watches) > 0 {
		return fmt.Errorf("Cannot call SetRecursive after watches are set")
	}
	w.recursive = true
	return nil
}
```

glours:
@thaJeztah I just opened a follow up [PR](https://github.com/docker/compose/pull/14165)

PR #14169, opened 2026-09-02
fix: skip unreadable default .env instead of failing

### LINKED ISSUE #14166: [BUG] permission denied on .env file in current directory lead to error
### Description

my current use case is using a sudo command to execute a tool that execute some docker commands...

to simplify the problem, running the following command fail with a permission denied (and btw the .env file did not exists)
```
[root@myhost ~]# sudo -u myuser docker compose -p myproject ps -a --format "{{.ID}}|{{title .Status}}|{{.Ports}}"
stat /root/.env: permission denied
[root@myhost ~]# sudo -u myuser docker compose ps -a --format "{{.ID}}|{{title .Status}}|{{.Ports}}"
stat /root/.env: permission denied
```
With current working directory changes it work as expected (no .env file in the /tmp directory)
```
[root@myhost ~]# cd /tmp
[root@myhost tmp]# sudo -u myuser docker compose ps -a --format "{{.ID}}|{{title .Status}}|{{.Ports}}"
no configuration file provided: not found
[root@myhost tmp]# sudo -u myuser docker compose -p myproject ps -a --format "{{.ID}}|{{title .Status}}|{{.Ports}}"
29ddbf36662e|Exited (0) 4 Minutes Ago|
05ed8afdf90d|Exited (0) 4 Minutes Ago|
797d87bcc51e|Exited (0) 4 Minutes Ago|
6be65b4385c5|Exited (0) 4 Minutes Ago|
```

Why do not just print a warning for this permission denied ?
Is there an acceptable workarround for this ?

Thanks fj

### Compose Version

```Text
docker-compose-plugin-5.5.0-1.el9.x86_64
```

### Docker Environment

```Text
RedHat 9
container-selinux-2.235.0-1.el9_6.noarch
docker-ce-cli-29.7.2-1.el9.x86_64
containerd.io-2.3.4-1.el9.x86_64
docker-ce-rootless-extras-29.7.2-1.el9.x86_64
docker-ce-29.7.2-1.el9.x86_64
```

### Anything else?

i didn't test in different system, i guess it working like this

locker95:
`docker compose` currently dies if it can't `stat` the default `.env` in the working directory, even when that file isn't required.

that shows up as `stat /root/.env: permission denied` when you `sudo -u otheruser docker compose ...` from a directory the target user can't read (the file doesn't even have to exist).

skip the default `.env` in that case and keep going. explicit `--env-file` is unchanged.

Fixes #14166

### DIFF
--- a/cmd/compose/compose.go
+++ b/cmd/compose/compose.go
@@ -399,7 +399,7 @@ func (o *ProjectOptions) toProjectOptions(po ...cli.ProjectOptionsFn) (*cli.Proj
 
 	opts = append(opts,
 		// Load PWD/.env if present and no explicit --env-file has been set
-		cli.WithEnvFiles(o.EnvFiles...),
+		utils.WithEnvFiles(o.EnvFiles...),
 		// read dot env file to populate project environment
 		cli.WithDotEnv,
 		// get compose file path set by COMPOSE_FILE
@@ -407,7 +407,7 @@ func (o *ProjectOptions) toProjectOptions(po ...cli.ProjectOptionsFn) (*cli.Proj
 		// if none was selected, get default compose.yaml file from current dir or parent folder
 		cli.WithDefaultConfigPath,
 		// .. and then, a project directory != PWD maybe has been set so let's load .env file
-		cli.WithEnvFiles(o.EnvFiles...), //nolint:gocritic // intentionally applying cli.WithEnvFiles twice.
+		utils.WithEnvFiles(o.EnvFiles...), //nolint:gocritic // intentionally applying cli.WithEnvFiles twice.
 		cli.WithDotEnv,                  //nolint:gocritic // intentionally applying cli.WithDotEnv twice.
 		// eventually COMPOSE_PROFILES should have been set
 		cli.WithDefaultProfiles(o.Profiles...),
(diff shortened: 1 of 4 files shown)


thaJeztah:
I'm not sure this is correct; compose _by default_ uses `.env` if present. If no `.env` file is present, that's ok to skip, but if it's present but cannot be read, that's a genuine error situation, and shouldn't be ignored (or reduced to a warning).

For these cases, either set the permissions on the `.env` to be accessible by the user that's used, or set the `COMPOSE_DISABLE_ENV_FILE=1` env-var to explicitly disable using `.env` (see https://docs.docker.com/compose/how-tos/environment-variables/envvars/#compose_disable_env_file)

ndeloof:
I created https://github.com/compose-spec/compose-go/pull/925 as a fix on the compose-go library

glours:
Separately, this account's public activity shows a very high-volume, high-velocity pattern of PRs opened across a large number of unrelated repositories in a very short period, with no prior history of contribution to the Docker/Compose ecosystem before this PR.

AI-assisted contributions are welcome here, but this pattern is more consistent with automated or bulk PR farming than with genuine project engagement, and that is not something we want to encourage. Regardless of whether any individual PR happens to be technically sound, we expect contributions to be individually considered, validated against the project, and followed through by the contributor.

PR #14173, opened 2026-09-02
build(deps): golang.org/x/crypto v0.56.0

thaJeztah:
full diff: https://github.com/golang/crypto/compare/v0.55.0...v0.56.0

We have tagged version v0.56.0 of golang.org/x/crypto in order to address the following security issues:
- ssh: prevent DoS on deadlocked established channel Previously, after a channel has been established, a malicious peer could send crafted messages that would deadlock the entire connection. Now, we handle all RFC 4254 channel messages; global requests are handled explicitly. Then, treat all other messages as a protocol error and tear the connection down instead of buffering and blocking. Thanks to Will Mortensen for reporting this issue. This is CVE-2026-56855 and Go issue https://go.dev/issue/81317.
- ssh: prevent DoS on deadlocked undecided channel Previously, a channel registered in the mux's chanList is not usable until it is established. A malicious peer was able flood the channel's incomingRequests, deadlocking the entire connection. Now, we add an atomic established state, set when a channel becomes usable. Until such a time, handlePacket drops every packet other than the open confirmation/failure, without blocking and without tearing down the connection. Thanks to Will Mortensen for reporting this issue. This is CVE-2026-78662 and Go issue https://go.dev/issue/81316.

**What I did**

**Related issue**


docker-agent (commented):
### Assessment: 🟢 APPROVE

glours (approved):

PR #14174, opened 2026-09-03
fix(bridge): confirm before wiping non-empty convert output dir

glours:
**What I did**
`bridge convert -o <dir>` unconditionally ran `os.RemoveAll` on the output directory and discarded any error, so a typo (`-o .`, `-o $PWD`, `-o ~`) silently destroyed unrelated user data.

An empty or missing output directory is still created silently, but a non-empty one now requires explicit confirmation (prompted interactively, or via the new `--yes` flag for scripts/CI) before being wiped.

**Related issue**
https://docker.atlassian.net/browse/DDB-683

image

### DIFF
--- a/cmd/compose/bridge.go
+++ b/cmd/compose/bridge.go
@@ -30,6 +30,7 @@ import (
 	"github.com/spf13/cobra"
 
 	"github.com/docker/compose/v5/cmd/formatter"
+	"github.com/docker/compose/v5/cmd/prompt"
 	"github.com/docker/compose/v5/pkg/bridge"
 	"github.com/docker/compose/v5/pkg/compose"
 )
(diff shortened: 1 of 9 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The safety guard is correctly implemented. `prepareOutputDir` checks for a non-empty directory, prompts for confirmation (or auto-confirms via `--yes`), and only then calls `os.RemoveAll` + `os.MkdirAll` — all with proper error propagation. The `--yes` flag is wired correctly through the CLI layer. The new `FileAbsent` / `FileContains` check helpers and the two Scenario-based e2e tests cover both the confirmed and declined paths cleanly.

Two hypotheses were raised and investigated during review:

- **Non-TTY stdin behaviour without `--yes`**: `NewPrompt` already branches on `stdin.IsTerminal()` and returns a `Pipe` implementation for non-TTY contexts. `Pipe.Confirm` reads via `fmt.Fscanln`; on EOF it silently gets an empty string, `StringToBool("")` → `false`, and the convert is declined with an explicit error — no hang, no silent data destruction. ✅ Dismissed.
- **E2E declined-path test hanging**: The test intentionally exercises the non-TTY decline path. Because `Pipe.Confirm` returns `false` on EOF stdin, the command exits with code 1 and the guard file is untouched — exactly what `ExitCode(1)` + `FileContains` assert. ✅ Dismissed.

ndeloof (approved):

PR #14175, opened 2026-09-03
feat(provider): get-service-config provider request

ndeloof:
Providers could not see the definition of the service they manage: options had to be duplicated between the compose file and the provider, or re-resolved outside the running process.

A provider may now emit `{"type": "get-service-config"}` on stdout; compose answers on the provider's stdin with one JSON line holding the resolved canonical configuration of the provider's own service, from the in-memory model. A compose that predates the message aborts on it and never writes to stdin, so a provider treats EOF as "unsupported, upgrade compose".

Scoped down from the original version: `addhost` and the links-style variables convention are superseded by the relay approach (#14193).

### DIFF
--- a/docs/examples/provider.go
+++ b/docs/examples/provider.go
@@ -91,6 +91,30 @@ func up(options options, args []string) {
 	servicename := args[0]
 	fmt.Printf(`{ "type": "debug", "message": "Starting %s" }%s`, servicename, lineSeparator)
 
+	// Ask the running Compose process for the resolved definition of the
+	// service this provider manages. A Compose that predates the message
+	// aborts on it, so only providers that require the configuration
+	// should send it. The decoder must be created once and reused across
+	// requests: it reads ahead, so a fresh decoder per request would discard
+	// buffered bytes and hang on the next answer.
+	responses := json.NewDecoder(os.Stdin)
+	fmt.Printf(`{ "type": "get-service-config" }%s`, lineSeparator)
+	var config struct {
+		Provider struct {
+			Type string `json:"type"`
+		} `json:"provider"`
+	}
+	if err := responses.Decode(&config); err != nil {
+		// error text is not JSON-safe either: encode, don't interpolate
+		msg, _ := json.Marshal(map[string]string{"type": "error", "message": fmt.Sprintf("get-service-config failed: %v", err)})
+		fmt.Println(string(msg))
+		return
+	}
+	// values read from the configuration are not necessarily JSON-safe:
+	// encode the message instead of interpolating it into a JSON literal
+	setenv, _ := json.Marshal(map[string]string{"type": "setenv", "message": "CONFIG_TYPE=" + config.Provider.Type})
+	fmt.Println(string(setenv))
+
 	for i := 0; i < options.size; i += 10 {
 		time.Sleep(1 * time.Second)
 		fmt.Printf(`{ "type": "info", "message": "Processing ... %d%%" }%s`, i*100/options.size, lineSeparator)
--- a/pkg/compose/plugins.go
+++ b/pkg/compose/plugins.go
@@ -41,7 +41,7 @@ import (
 
 type JsonMessage struct {
 	Type    string `json:"type"`
-	Message string `json:"message"`
+	Message string `json:"message,omitempty"`
 }
 
 const (
@@ -51,6 +51,11 @@ const (
 	RawSetEnvType             = "rawsetenv"
 	DebugType                 = "debug"
 	providerMetadataDirectory = "compose/providers"
+
+	// GetServiceConfigType is a message the provider sends to receive, on
+	// its stdin, one JSON line holding the resolved canonical configuration
+	// of the service it manages — answered from the in-memory model.
+	GetServiceConfigType = "get-service-config"
 )
 
 type pluginVariables struct {
@@ -125,11 +130,51 @@ func (s *composeService) executePlugin(cmd *exec.Cmd, command string, service ty
 	if err != nil {
 		return pluginVariables{}, err
 	}
+	stdin, err := cmd.StdinPipe()
+	if err != nil {
+		return pluginVariables{}, err
+	}
+	// Answers are written from their own goroutine: writing a config larger
+	// than the OS pipe buffer from the read loop would deadlock against a
+	// provider that emits stdout before draining its stdin. Every answer is
+	// the same serialized service, so completion order is irrelevant; the
+	// mutex keeps individual writes atomic. Write failures are not reported
+	// from here — a provider missing its answer reads EOF once stdin closes.
+	var stdinMu sync.Mutex
+	var answers sync.WaitGroup
+	processExited := false
+	// Closing stdin on exit unblocks a provider waiting for a response the
+	// loop will never produce (e.g. a request emitted after an error). An
+	// answer dispatched but not yet written must land before the close —
+	// but only once the process has exited can a write not block forever
+	// (a dead peer turns it into EPIPE); on error paths the provider may
+	// still be alive and not reading, so close first to error the write
+	// out instead of hanging the wait.
+	defer func() {
+		if processExited {
+			answers.Wait()
+			_ = stdin.Close()
+		} else {
+			_ = stdin.Close()
+			answers.Wait()
+		}
+	}()
 
 	err = cmd.Start()
 	if err != nil {
 		return pluginVariables{}, err
 	}
+	// Error paths return before the normal cmd.Wait below and would leave
+	// the provider as a zombie (and possibly running): reap it — kill
+	// first, as it may be misbehaving or blocked, which also errors out
+	// any in-flight answer write. Runs before the stdin/answers defer
+	// above (LIFO).
+	defer func() {
+		if !processExited {
+			_ = cmd.Process.Kill()
+			_ = cmd.Wait()
+		}
+	}()
 
 	decoder := json.NewDecoder(stdout)
 	defer func() { _ = stdout.Close() }()
@@ -166,6 +211,19 @@ func (s *composeService) executePlugin(cmd *exec.Cmd, command string, service ty
 				return pluginVariables{}, fmt.Errorf("invalid response from plugin: %s", msg.Message)
 			}
 			variables.raw[key] = val
+		case GetServiceConfigType:
+			payload, err := json.Marshal(service)
+			if err != nil {
+				return pluginVariables{}, fmt.Errorf("failed to answer get-service-config: %w", err)
+			}
+			payload = append(payload, '\n')
+			answers.Add(1)
+			go func() {
+				defer answers.Done()
+				stdinMu.Lock()
+				defer stdinMu.Unlock()
+				_, _ = stdin.Write(payload)
+			}()
 		case DebugType:
 			logrus.Debugf("%s: %s", service.Name, msg.Message)
 		default:
@@ -174,6 +232,7 @@ func (s *composeService) executePlugin(cmd *exec.Cmd, command string, service ty
 	}
 
 	err = cmd.Wait()
+	processExited = true
 	if err != nil {
 		s.events.On(errorEvent(service.Name, err.Error()))
 		return pluginVariables{}, fmt.Errorf("failed to %s service provider: %s", action, err.Error())
--- a/pkg/compose/plugins_control_test.go
+++ b/pkg/compose/plugins_control_test.go
@@ -1,0 +1,96 @@
+/*
+   Copyright 2020 Docker Compose CLI authors
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package compose
+
+import (
+	"encoding/json"
+	"fmt"
+	"os"
+	"os/exec"
+	"testing"
+
+	"github.com/compose-spec/compose-go/v2/types"
+	"go.uber.org/mock/gomock"
+	"gotest.tools/v3/assert"
+
+	"github.com/docker/compose/v5/pkg/mocks"
+)
+
+// TestExecutePlugin_GetServiceConfig runs executePlugin against a fake
+// provider (this test binary re-executed, see TestHelperProviderConfig): each
+// get-service-config message must be answered on the provider's stdin with
+// one JSON line holding the in-memory service's canonical configuration.
+func TestExecutePlugin_GetServiceConfig(t *testing.T) {
+	mockCtrl := gomock.NewController(t)
+	cli := mocks.NewMockCli(mockCtrl)
+	cli.EXPECT().Client().Return(mocks.NewMockAPIClient(mockCtrl)).AnyTimes()
+	svc, err := NewComposeService(cli, WithEventProcessor(noopEventProcessor{}))
+	assert.NilError(t, err)
+
+	cmd := exec.Command(os.Args[0], "-test.run=TestHelperProviderConfig")
+	cmd.Env = append(os.Environ(), "GO_WANT_HELPER_PROCESS=1")
+
+	service := types.ServiceConfig{
+		Name: "db",
+		Provider: &types.ServiceProviderConfig{
+			Type:    "sbx",
+			Options: types.MultiOptions{"template": {"agent:latest"}},
+		},
+	}
+	variables, err := svc.(*composeService).executePlugin(cmd, "up", service)
+	assert.NilError(t, err)
+	assert.Equal(t, variables.prefixed["TEMPLATE"], "agent:latest")
+	// the channel stays usable for more than one request
+	assert.Equal(t, variables.prefixed["TEMPLATE_AGAIN"], "agent:latest")
+}
+
+// TestHelperProviderConfig is not a test: it is the fake provider process
+// spawned by TestExecutePlugin_GetServiceConfig.
+func TestHelperProviderConfig(t *testing.T) {
+	if os.Getenv("GO_WANT_HELPER_PROCESS") != "1" {
+		t.Skip("helper process for TestExecutePlugin_GetServiceConfig")
+	}
+	responses := json.NewDecoder(os.Stdin)
+	emit := func(msg JsonMessage) {
+		if err := json.NewEncoder(os.Stdout).Encode(msg); err != nil {
+			os.Exit(1)
+		}
+	}
+	getServiceConfig := func() (template string, ok bool) {
+		emit(JsonMessage{Type: GetServiceConfigType})
+		var config struct {
+			Provider struct {
+				Options map[string][]string `json:"options"`
+			} `json:"provider"`
+		}
+		if err := responses.Decode(&config); err != nil {
+			emit(JsonMessage{Type: ErrorType, Message: fmt.Sprintf("reading service config: %v", err)})
+			return "", false
+		}
+		return config.Provider.Options["template"][0], true
+	}
+
+	template, ok := getServiceConfig()
+	if !ok {
+		os.Exit(0)
+	}
+	emit(JsonMessage{Type: SetEnvType, Message: "TEMPLATE=" + template})
+	if template, ok = getServiceConfig(); ok {
+		emit(JsonMessage{Type: SetEnvType, Message: "TEMPLATE_AGAIN=" + template})
+	}
+	os.Exit(0)
+}
--- a/docs/extension.md
+++ b/docs/extension.md
@@ -59,6 +59,30 @@ JSON messages MUST include a `type` and a `message` attribute.
 - `setenv`: Lets the plugin tell Compose how dependent services can access the created resource. The variable is automatically prefixed with the service name. See next section for further details.
 - `rawsetenv`: Same as `setenv`, but the variable is injected as-is without the service name prefix. Useful when applications require exact variable names that cannot be altered.
 - `debug`: Those messages could help debugging the provider, but are not rendered to the user by default. They are rendered when Compose is started with `--verbose` flag.
+- `get-service-config`: Asks Compose for the resolved configuration of the service the provider manages. See next section.
+
+## Requesting the service configuration
+
+A provider can ask the running Compose process for the resolved definition of the service it manages —
+the exact model Compose is executing, not a re-resolution. The request is a regular JSON line on `stdout`:
+```json
+{ "type": "get-service-config" }
+```
+
+Compose answers on the provider's `stdin` with one JSON line: the resolved, canonical JSON of the service —
+the same shape as this service's entry in `docker compose config --format json`, after interpolation and
+normalization:
+```json
+{ "image": "mysql:8", "environment": { "...": "..." } }
+```
+
+There is no parameter: a provider can only obtain the definition of its own service. The message can be sent
+several times; each occurrence is answered with one line.
+
+Compose versions that predate this message treat it as a protocol error and abort the command, and never
+write anything to the provider's `stdin` (the provider reads EOF). A provider that requires the service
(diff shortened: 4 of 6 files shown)


docs/examples/provider.go
```diff
@@ -91,6 +91,35 @@ func up(options options, args []string) {
...
+		var config struct {
+			Provider struct {
+				Type string `json:"type"`
+			} `json:"provider"`
+		}
+		if err := json.Unmarshal(response.Result, &config); err != nil {
+			fmt.Printf(`{ "type": "error", "message": "invalid service config: %v" }%s`, err, lineSeparator)
+			return
+		}
+		fmt.Printf(`{ "type": "setenv", "message": "CONFIG_TYPE=%s" }%s`, config.Provider.Type, lineSeparator)
```
docker-agent: **[low] JSON injection: unescaped `config.Provider.Type` in hand-crafted JSON string**

The example emits a `setenv` message by interpolating `config.Provider.Type` directly into a raw string literal:

```go
fmt.Printf(`{ "type": "setenv", "message": "CONFIG_TYPE=%s" }%s`, config.Provider.Type, lineSeparator)
```

If `config.Provider.Type` contains a `"` (double-quote), `\` (backslash), or a newline, the emitted line is not valid JSON. Compose reads it with `json.NewDecoder(stdout).Decode(&msg)` — a decode error causes `executePlugin` to return with an error, aborting the provider interaction entirely.

While provider type values are typically simple identifiers like `"sbx"`, the example code is what users copy when writing their own providers. Following this pattern with user-controlled or URL-like values propagates the bug into real implementations.

Use a struct and `json.Marshal` (or `json.NewEncoder`) to produce the message safely:

```go
type msg struct {
    Type    string `json:"type"`
    Message string `json:"message"`
}
b, _ := json.Marshal(msg{Type: "setenv", Message: "CONFIG_TYPE=" + config.Provider.Type})
fmt.Printf("%s%s", b, lineSeparator)
```

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 75/100 |

docker-agent (commented):
### Assessment: 🟢 APPROVE

The control-channel implementation is well-structured. The synchronous request/response loop is simple and correct for the typical case. One finding in the example code, and one lower-confidence note below.

- [low] pkg/compose/plugins.go:208 — Potential deadlock when large service config response exceeds OS pipe buffer (confidence: weak 52/100). The synchronous `responses.Encode()` write to the provider's stdin happens in the same goroutine that reads from stdout. If the marshalled `types.ServiceConfig` ever exceeds the OS pipe buffer (~64 KB on Linux), and the provider is blocked waiting for this response before writing more output, neither end makes progress. Typical configs are well under 1 KB so this is unlikely in practice, but a goroutine for the write would eliminate the risk entirely.

docker-agent (commented):

One medium-severity correctness issue found in the example code added by this PR.

docs/examples/provider.go
```diff
@@ -91,6 +91,22 @@ func up(options options, args []string) {
...
+	// service this provider manages. A Compose that predates the message
+	// aborts on it, so only providers that require the configuration
+	// should send it.
+	fmt.Printf(`{ "type": "get-service-config" }%s`, lineSeparator)
+	var config struct {
+		Provider struct {
+			Type string `json:"type"`
+		} `json:"provider"`
+	}
+	if err := json.NewDecoder(os.Stdin).Decode(&config); err != nil {
```
docker-agent: **[medium] `json.NewDecoder(os.Stdin)` re-created per call — buffered data silently discarded on repeated requests**

`json.Decoder` maintains an internal read-ahead buffer: when it decodes one JSON token it may have already consumed additional bytes from the underlying reader into that buffer. Discarding the decoder also discards those bytes.

The example creates a fresh `json.NewDecoder(os.Stdin)` on every call to decode a service-config response:

```go
if err := json.NewDecoder(os.Stdin).Decode(&config); err != nil { … }
```

The extension docs explicitly state "The message can be sent several times; each occurrence is answered with one line." A provider that follows this example pattern and sends `get-service-config` a second time will hang forever waiting for data Compose already wrote — it was silently consumed by the first decoder's internal buffer and then thrown away when that decoder was discarded.

The `TestHelperProviderConfig` test helper (added in the same PR) correctly avoids this by reusing a single `bufio.NewReader(os.Stdin)`. The example should follow the same pattern — create the decoder (or reader) once and reuse it:

```suggestion
	stdin := json.NewDecoder(os.Stdin)
	if err := stdin.Decode(&config); err != nil {
```

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 97/100 |

docs/examples/provider.go
ndeloof: Fixed in 500f4e0e8: the setenv message is now json.Marshal-ed instead of interpolated into a JSON literal, with a comment stating why (config values are not JSON-safe).

docs/examples/provider.go
ndeloof: Fixed in 500f4e0e8: the stdin decoder is created once and reused, with a comment explaining the read-ahead pitfall.

docker-agent: Perfect, thank you! The fix in 500f4e0e8 correctly addresses the issue — the decoder is now created once as `responses` and reused, and the comment clearly explains the read-ahead pitfall for future readers.

The example now matches the safe pattern used in the test helper. Appreciate the quick turnaround on this!

docs/examples/provider.go
```diff
@@ -91,6 +91,28 @@ func up(options options, args []string) {
...
+	// buffered bytes and hang on the next answer.
+	responses := json.NewDecoder(os.Stdin)
+	fmt.Printf(`{ "type": "get-service-config" }%s`, lineSeparator)
+	var config struct {
+		Provider struct {
+			Type string `json:"type"`
+		} `json:"provider"`
+	}
+	if err := responses.Decode(&config); err != nil {
+		fmt.Printf(`{ "type": "error", "message": "get-service-config failed: %v" }%s`, err, lineSeparator)
```
docker-agent: **[medium] Error message interpolated directly into JSON literal — invalid JSON if `err` contains special characters**

On the error path after a failed `Decode`, the error value is interpolated directly into a hand-rolled JSON string:

```go
fmt.Printf(`{ "type": "error", "message": "get-service-config failed: %v" }%s`, err, lineSeparator)
```

If `err.Error()` contains a `"`, `\`, or newline (e.g. the Go JSON decoder error `invalid character '"' after object key`) the resulting line is not valid JSON. The Compose parent process then receives a malformed message and will either misparse it or log a confusing error.

The very next block in the same function correctly avoids this problem by using `json.Marshal`:

```go
// values read from the configuration are not necessarily JSON-safe:
// encode the message instead of interpolating it into a JSON literal
setenv, _ := json.Marshal(map[string]string{"type": "setenv", "message": "CONFIG_TYPE=" + config.Provider.Type})
```

The same treatment should be applied to the error path:

```suggestion
		errMsg, _ := json.Marshal(map[string]string{"type": "error", "message": fmt.Sprintf("get-service-config failed: %v", err)})
		fmt.Println(string(errMsg))
```

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 97/100 |

pkg/compose/plugins.go
```diff
@@ -166,6 +179,10 @@ func (s *composeService) executePlugin(cmd *exec.Cmd, command string, service ty
 				return pluginVariables{}, fmt.Errorf("invalid response from plugin: %s", msg.Message)
 			}
 			variables.raw[key] = val
+		case GetServiceConfigType:
+			if err := responses.Encode(service); err != nil {
```
docker-agent: **[medium] Potential pipe deadlock when `types.ServiceConfig` JSON exceeds the OS pipe buffer (~64 KB)**

The `GetServiceConfigType` handler writes the serialized service config to the provider's stdin synchronously, in the same goroutine that is also reading from the provider's stdout:

```go
case GetServiceConfigType:
    if err := responses.Encode(service); err != nil {   // blocks until written
        return pluginVariables{}, fmt.Errorf("failed to answer get-service-config: %w", err)
    }
```

On the provider side, the same sequential pattern applies: emit the `get-service-config` request to stdout, then immediately block reading from stdin for the answer.

This creates a classic pipe-buffer deadlock:
1. Compose reads the `get-service-config` request from the provider's stdout → OK so far.
2. Compose calls `responses.Encode(service)` and starts writing the config to the provider's stdin pipe.
3. If the encoded JSON exceeds ~64 KB, the kernel pipe buffer fills. Compose's write blocks, waiting for the provider to drain the buffer.
4. Meanwhile the provider is also blocked in its read call — it is waiting for Compose to finish writing before it can return from `Decode`/`ReadBytes`.
5. Neither side can proceed. **Deadlock.**

A `types.ServiceConfig` can realistically exceed 64 KB for services with many environment variables, a large `configs` or `secrets` tree, multiple volume mounts, or deeply nested `extensions`. The risk is not theoretical on complex projects.

**Fix**: write the config to the provider's stdin from a separate goroutine, so Compose's read loop is never blocked behind its own write:

```go
case GetServiceConfigType:
    go func() {
        _ = responses.Encode(service)
    }()
```

(Error handling can be added to the goroutine if needed; the provider will get an EOF or partial JSON and should report an error via its normal `ErrorType` message path.)

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 57/100 |

docker-agent (commented):

Two findings in the newly introduced `get-service-config` protocol code:
1. **docs/examples/provider.go:108** — error path emits hand-rolled JSON that breaks on special characters in the error string, contradicting the comment on the very next line.
2. **pkg/compose/plugins.go:183** — synchronous write to the provider's stdin pipe inside the stdout-reading loop can deadlock when the serialized `ServiceConfig` JSON exceeds the OS pipe buffer (~64 KB).

pkg/compose/plugins_control_test.go
```diff
@@ -0,0 +1,102 @@
...
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package compose
+
+import (
+	"bufio"
+	"encoding/json"
```
glours:
```suggestion
	"encoding/json"
```

pkg/compose/plugins_control_test.go
```diff
@@ -0,0 +1,102 @@
...
+	assert.Equal(t, variables.prefixed["TEMPLATE_AGAIN"], "agent:latest")
+}
+
+// TestHelperProviderConfig is not a test: it is the fake provider process
+// spawned by TestExecutePlugin_GetServiceConfig.
+func TestHelperProviderConfig(t *testing.T) {
+	if os.Getenv("GO_WANT_HELPER_PROCESS") != "1" {
+		t.Skip("helper process for TestExecutePlugin_GetServiceConfig")
+	}
+	stdin := bufio.NewReader(os.Stdin)
```
glours:
```suggestion
	 responses := json.NewDecoder(os.Stdin)
```

pkg/compose/plugins_control_test.go
```diff
@@ -0,0 +1,102 @@
...
+			os.Exit(1)
+		}
+	}
+	getServiceConfig := func() (template string, ok bool) {
+		emit(JsonMessage{Type: GetServiceConfigType})
+		line, err := stdin.ReadBytes('\n')
+		if err != nil {
+			emit(JsonMessage{Type: ErrorType, Message: fmt.Sprintf("reading service config: %v", err)})
+			return "", false
+		}
```
glours:
```suggestion
	
```

pkg/compose/plugins_control_test.go
```diff
@@ -0,0 +1,102 @@
...
+		var config struct {
+			Provider struct {
+				Options map[string][]string `json:"options"`
+			} `json:"provider"`
+		}
+		if err := json.Unmarshal(line, &config); err != nil {
+			emit(JsonMessage{Type: ErrorType, Message: fmt.Sprintf("bad service config %s: %v", line, err)})
```
glours:
```suggestion
		if err := responses.Decode(&config); err != nil {
			emit(JsonMessage{Type: ErrorType, Message: fmt.Sprintf("reading service config: %v", err)})
```

glours (commented):
nit: this hand-rolls what `json.Decoder` already does — `bufio.Reader.ReadBytes('\n')` + `json.Unmarshal` assumes each message is exactly one newline-terminated line, which happens to hold today but is a more fragile assumption than just decoding from the stream directly. 
`docs/examples/provider.go`, added by this same PR, already uses the simpler `json.NewDecoder(os.Stdin)` / `.Decode(&config)` pattern, worth mirroring it here for consistency (and it lets you drop the `bufio` import too).

docs/examples/provider.go
ndeloof: Fixed in 1dcf08b79: the error message is json.Marshal-ed like the setenv one.

pkg/compose/plugins.go
ndeloof: Fixed in 1dcf08b79: answers are written from their own goroutine (mutex-serialized), so the read loop never blocks on a config larger than the pipe buffer against a provider still emitting stdout. Every answer is the same serialized service, so completion order is irrelevant; a provider missing its answer reads EOF once stdin closes on exit.

pkg/compose/plugins_control_test.go
ndeloof: Applied in 1dcf08b79 (bufio dropped, shared json.Decoder).

pkg/compose/plugins_control_test.go
ndeloof: Applied in 1dcf08b79.

pkg/compose/plugins_control_test.go
ndeloof: Applied in 1dcf08b79.

pkg/compose/plugins_control_test.go
ndeloof: Applied in 1dcf08b79.

pkg/compose/plugins.go
```diff
@@ -166,6 +186,17 @@ func (s *composeService) executePlugin(cmd *exec.Cmd, command string, service ty
 				return pluginVariables{}, fmt.Errorf("invalid response from plugin: %s", msg.Message)
 			}
 			variables.raw[key] = val
+		case GetServiceConfigType:
+			payload, err := json.Marshal(service)
+			if err != nil {
+				return pluginVariables{}, fmt.Errorf("failed to answer get-service-config: %w", err)
+			}
+			payload = append(payload, '\n')
+			go func() {
```
docker-agent, line 221: **[medium] Goroutine dispatched for `get-service-config` may write to stdin after `stdin.Close()`, silently dropping the reply**

The goroutine spawned here (`go func() { stdinMu.Lock(); …; stdin.Write(payload) }()`) is not tracked by any `sync.WaitGroup`. When `executePlugin` returns, `defer func() { _ = stdin.Close() }()` fires immediately — but Go's scheduler may not have run the goroutine yet. The resulting sequence:

1. Read loop decodes the last `get-service-config` message, dispatches goroutine.
2. Provider closes its stdout → decoder hits EOF → loop exits.
3. `defer stdin.Close()` fires.
4. Goroutine wakes up, acquires `stdinMu`, writes to the now-closed pipe → `EPIPE` / broken pipe, silently discarded via `_, _ = stdin.Write(...)`.
5. Provider's next `os.Stdin` read returns EOF instead of the expected config JSON.

This is most likely to occur when `get-service-config` is the provider's last stdout message before exiting — exactly the natural usage pattern where a provider requests its config and then acts on it. The author's comment acknowledges "a provider missing its answer reads EOF once stdin closes", but in that case the feature silently fails: the provider receives no error from compose, no config data arrives, and `responses.Decode(&config)` returns `io.EOF`.

**Fix**: add a `sync.WaitGroup` to track in-flight goroutines and call `wg.Wait()` before returning (before the defer closes stdin), so every dispatched write completes or is aborted cleanly:

```suggestion
			payload = append(payload, '\n')
			wg.Add(1)
			go func() {
				defer wg.Done()
				stdinMu.Lock()
				defer stdinMu.Unlock()
				_, _ = stdin.Write(payload)
			}()
```

(Declare `var wg sync.WaitGroup` alongside `var stdinMu sync.Mutex`, and call `wg.Wait()` after the read loop exits and before `cmd.Wait()`.)

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 100/100 |

docker-agent (commented):

Two medium/low-confidence correctness findings in the new `get-service-config` control-channel handler. The core issue is that goroutines dispatched to write the config reply to stdin are not tracked by a `sync.WaitGroup`, so they can write to a closed pipe when the read loop exits — silently dropping the reply to the provider.

pkg/compose/plugins.go
```diff
@@ -166,6 +186,17 @@ func (s *composeService) executePlugin(cmd *exec.Cmd, command string, service ty
 				return pluginVariables{}, fmt.Errorf("invalid response from plugin: %s", msg.Message)
 			}
 			variables.raw[key] = val
+		case GetServiceConfigType:
+			payload, err := json.Marshal(service)
+			if err != nil {
+				return pluginVariables{}, fmt.Errorf("failed to answer get-service-config: %w", err)
```
docker-agent, line 217: **[low] Early return on `json.Marshal` error fires `defer stdin.Close()` while previously-dispatched goroutines may still be in-flight**

If `json.Marshal(service)` fails at line 191 (after earlier `get-service-config` goroutines have already been dispatched), `executePlugin` returns early and `defer stdin.Close()` fires immediately. Any goroutines that were queued but not yet scheduled will attempt to write to the now-closed pipe, silently failing. In practice `json.Marshal` on a `types.ServiceConfig` almost never fails (no un-marshallable fields), so this is a theoretical edge case — but it shares the same root cause as the missing `WaitGroup` noted in the goroutine dispatch above. Fixing the `WaitGroup` there would also resolve this path.

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 100/100 |

pkg/compose/plugins.go
ndeloof: Fixed in e1ce900df: dispatched answers join a WaitGroup. On the normal path the wait runs after cmd.Wait (the peer is gone, a pending write cannot block) and before stdin.Close, so the last answer always lands; on error paths stdin closes first — the provider may still be alive and not reading, so a blocked write errors out instead of hanging the wait. Race-tested (-race, repeated runs).

pkg/compose/plugins.go
ndeloof: Fixed in e1ce900df by the same WaitGroup: the marshal-error return waits for in-flight answers too (close-then-wait on that path).

pkg/compose/plugins.go
```diff
@@ -166,6 +200,19 @@ func (s *composeService) executePlugin(cmd *exec.Cmd, command string, service ty
...
```
docker-agent, line 217: **[medium] Early return on `json.Marshal` failure skips `cmd.Wait()`, leaving a zombie child process**

When `json.Marshal(service)` fails inside the `GetServiceConfigType` case, `executePlugin` returns immediately here — bypassing the `cmd.Wait()` call at the end of the function. Since `cmd.Start()` has already been called, the child provider process is running. The deferred closure does call `answers.Wait()` and `stdin.Close()`, so goroutine resources are cleaned up; however, the process itself is not reaped: Go's runtime does not automatically reap child processes started with `cmd.Start()`, so the OS process-table entry remains as a zombie until the parent `compose` process exits.

`json.Marshal` rarely fails for a well-typed struct, but the error path is new code introduced by this PR. Consider adding a `defer func() { _ = cmd.Wait() }()` just after `cmd.Start()` succeeds (guarded to run only once), or calling `cmd.Process.Kill(); cmd.Wait()` before returning on this path.

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 62/100 |

docker-agent (commented):

One medium-severity finding introduced by this PR.

pkg/compose/plugins.go
```diff
@@ -125,6 +130,35 @@ func (s *composeService) executePlugin(cmd *exec.Cmd, command string, service ty
...
 		return pluginVariables{}, err
 	}
+	stdin, err := cmd.StdinPipe()
+	if err != nil {
+		return pluginVariables{}, err
+	}
+	// Answers are written from their own goroutine: writing a config larger
+	// than the OS pipe buffer from the read loop would deadlock against a
+	// provider that emits stdout before draining its stdin. Every answer is
+	// the same serialized service, so completion order is irrelevant; the
```
docker-agent, line 140: **[medium] Race between `defer stdin.Close()` and concurrent goroutine `stdin.Write()`**

`defer func() { _ = stdin.Close() }()` at line 140 runs without holding `stdinMu`. A goroutine spawned for a prior `get-service-config` response may be about to call — or mid-way through — `stdin.Write(payload)` while `stdin.Close()` executes concurrently. `Close()` is not serialised against the mutex, so both the pipe's file-descriptor state and the internal buffering inside `exec.Cmd`'s pipe implementation are accessed from two goroutines simultaneously. Go's race detector would flag this.

The fix is straightforward: acquire `stdinMu` in the defer before closing, so no in-flight writer can overlap with the close:

```suggestion
	defer func() { stdinMu.Lock(); _ = stdin.Close(); stdinMu.Unlock() }()
```

This ensures `Close()` only runs after any current writer has released the mutex, and any subsequent goroutine that wakes up finds the pipe already closed (gets a harmless broken-pipe error), rather than racing mid-write.

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 100/100 |

docker-agent (commented):

One CONFIRMED medium-severity race condition found in the new `get-service-config` control channel implementation.

docker-agent (commented):
### Assessment: 🟢 APPROVE

The `get-service-config` protocol implementation is well-structured. The concurrency design in `plugins.go` correctly handles the bidirectional stdin/stdout channel:

- **Goroutine coordination**: `answers.WaitGroup` + `stdinMu` mutex properly serialize writes to the provider's stdin pipe, preventing interleaving and buffering deadlocks.
- **Defer ordering**: The two-branch defer (close-then-wait on error paths, wait-then-close on clean exit) is the right design — it prevents hanging when the provider is still alive and reading, while ensuring in-flight writes complete on normal exit.
- **Closure safety**: `payload` is declared locally in the `GetServiceConfigType` case block before `answers.Add(1)`, so each goroutine captures its own copy — no shared-variable capture issue.
- **`processExited` access**: Read and written exclusively in the main goroutine (the defer runs in the same goroutine after `cmd.Wait()`) — no data race.
- **Decoder reuse**: The example provider correctly creates `json.NewDecoder(os.Stdin)` once and reuses it across requests, avoiding the read-ahead pitfall.
- **JSON safety**: Config values written to setenv messages go through `json.Marshal` rather than string interpolation — no injection risk.

pkg/compose/plugins.go
ndeloof: Fixed in 6a4b68457 — and generalized: every error path (not only the new marshal one — decode errors, provider error messages, invalid messages all predate this PR and skipped cmd.Wait too) now reaps the provider via a deferred kill+Wait guarded by processExited. Kill first: the provider may be misbehaving or blocked, and it also errors out any in-flight answer write.

pkg/compose/plugins.go
ndeloof: Not a data race: StdinPipe hands back an *os.File (wrapped in os/exec closeOnce, whose Close is sync.Once-guarded), and os.File methods are documented safe for concurrent use — poll.FD serializes Close against in-flight Writes internally. Closing WITHOUT taking stdinMu is deliberate: serializing Close behind the mutex would deadlock against a write blocked on a full pipe, which the close exists to unblock. Verified with -race over repeated multi-request runs.

docker-agent (commented):
### Assessment: 🟢 APPROVE

The `get-service-config` control-channel implementation is well-structured. The goroutine-based stdin write pattern correctly avoids deadlocks (large configs won't block the stdout read loop), the LIFO defer ordering is sound for both normal and error exit paths, and the `processExited` flag correctly gates the two cleanup strategies. The `omitempty` addition on `JsonMessage.Message` is safe: Compose only decodes `JsonMessage` (reads provider stdout) and never encodes it, so omitting the empty field from the test helper's output has no impact on real providers. The provider example correctly reuses a single `json.Decoder` for stdin across multiple requests, avoiding read-ahead data loss.

glours (approved):
LGTM

PR #14177, opened 2026-09-03
fix: honor --parallel across all bulk engine-call fan-outs

glours:
**What I did**
errgroup.SetLimit(0) means "allow zero goroutines", not "unlimited", and maxConcurrency's Go zero-value is 0 — only NewComposeService sets it to -1 explicitly. Any composeService{} literal built without it (common in tests) silently deadlocked at every call site that called SetLimit unconditionally.

Separately, --parallel/COMPOSE_PARALLEL_LIMIT was only ever wired into pull, push, and the dependency-graph traversal, despite the docs promising a generic bound on "concurrent engine calls". Every other bulk operation (kill, pause, down, the up/create plan executor, logs, ps, top, wait, restart, remove, images, model pulls, watch) launched one goroutine per container/image/DAG-node with no cap at all.

Fix both via a shared newLimitedErrgroup helper applied at every fan-out site, threading maxConcurrency through forEachContainerConcurrent and ImagePruner, which had no access to composeService. Add a regression test for the zero-value case.

service_containers.go's waitDependencies is intentionally left unguarded: it's a per-dependency ticker poll, not a burst of engine calls.

**Related issue**


docker-agent (commented):
### Assessment: 🟢 APPROVE

The `newLimitedErrgroup` helper is correctly implemented: it only calls `SetLimit` when `maxConcurrency > 0`, properly guarding against the zero-value deadlock. The fan-out at every bulk operation site follows the standard pattern safely. The regression test in `compose_test.go` correctly exercises the zero-value case. All hypothesized issues were dismissed after verification.

ndeloof (dismissed):

ndeloof (commented):
Thorough pass over every converted fan-out (result ordering, shared-state races, cancellation, error semantics) plus a local `-race` run: the mechanics are solid — `ps`/`top` fill indexed slices, `images`/`pull` guard their maps, `pullRequiredImages` keeps the caller ctx for post-pull resolution, and the executor's no-deadlock argument in the comment actually holds (topological `plan.Nodes` by construction + bounded dispatch ⇒ the oldest unfinished node always has finished dependencies). The zero-value test targets exactly the original bug.

One blocker, then two notes:

**[high] `Logs` in `--follow` starves everything past the cap — including the monitor.** Every `logContainer` call in follow mode streams indefinitely and never frees its errgroup slot, and `monitor.Start` is itself submitted through the same bounded group *after* the container loop. With `--parallel 2` and 5 services, only two log streams ever open and the other three never appear; with `--parallel 1`, one stream opens and the monitor never starts at all — silently, for a documented knob. The PR already excludes `waitDependencies` for exactly this reason (long waits, not bursts of engine calls): follow-mode `Logs` belongs in the same exclusion. Simplest fix: keep the bounded group for the non-follow case only (or bound just the attach phase).

**[low]** `restart` bounds its inner per-service group, but it runs inside `InDependencyOrder` whose traversal has its own `SetLimit(maxConcurrency+1)`: up to N×N concurrent engine calls. Strictly better than the unbounded inner loop it replaces, but the "bounded engine calls" promise isn't global — a shared semaphore would be the real fix, fine as a follow-up.

**[nit]** `NewImagePruner` gains a parameter — exported API, external consumers of pkg/compose break at compile. Consistent with the repo's API posture, just flagging it.

pkg/compose/logs.go
```diff
@@ -146,6 +179,7 @@ func (s *composeService) doLogContainer(ctx context.Context, consumer api.LogCon
 		Tail:       options.Tail,
 		Timestamps: options.Timestamps,
 	})
+	releaseSlot(limiter)
```
docker-agent, line 182: **[medium] `releaseSlot` in `doLogContainer` is not panic-safe — a panic inside `ContainerLogs` leaks the semaphore slot**

`doLogContainer` relies on the caller (`logContainer`) having acquired a semaphore slot via `acquireSlot`, and releases it unconditionally on line 182 right after `ContainerLogs` returns:

```go
r, err := s.apiClient().ContainerLogs(...)
releaseSlot(limiter)   // released only if ContainerLogs returns normally
```

If `ContainerLogs` panics (e.g. due to a nil dereference inside the Docker client), the `releaseSlot` call on line 182 is never reached. The slot is never returned to the semaphore, so the effective concurrency limit drops by 1 permanently for the lifetime of the process. With a small `--parallel` value this can eventually exhaust all slots and cause subsequent containers to block in `acquireSlot` indefinitely.

The design intentionally avoids `defer releaseSlot(limiter)` here because the slot must be freed right after the `ContainerLogs` call opens the stream (not at function exit), so that other containers can begin opening their streams while this goroutine copies its indefinite log stream. A panic guard that re-panics after releasing the slot would preserve both the intent and panic safety:

```go
func panicSafeReleaseSlot(limiter *semaphore.Weighted) {
    if p := recover(); p != nil {
        releaseSlot(limiter)
        panic(p)
    }
}
// at the top of doLogContainer, before ContainerLogs:
defer panicSafeReleaseSlot(limiter)
```

In practice, panics in the Docker client's HTTP call path are extremely rare, so the real-world impact is low. Still, given that the `acquireSlot`/`releaseSlot` split-ownership pattern is new and non-obvious, documenting the panic-safety gap and adding the guard is worth considering.

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 62/100 |

docker-agent (commented):
### Assessment: 🟢 APPROVE

The PR correctly fixes the `errgroup.SetLimit(0)` footgun via `newLimitedErrgroup` and applies the concurrency cap consistently across all bulk fan-out sites. The helper and its callers are well-structured; the follow-mode logs path correctly uses a separate semaphore (rather than a bounded errgroup) to avoid starving the monitor goroutine.

One minor concern surfaced by the review:

pkg/compose/wait.go
```diff
@@ -48,7 +47,7 @@ func (s *composeService) Wait(ctx context.Context, projectName string, options a
 		return 0, fmt.Errorf("no containers for project %q", projectName)
 	}
 
-	eg, waitCtx := errgroup.WithContext(ctx)
+	eg, waitCtx := newLimitedErrgroup(ctx, s.maxConcurrency)
```
docker-agent: **[medium] `newLimitedErrgroup` applied to `Wait` serialises container-exit polling when `--parallel` is set**

`compose wait` is designed to watch **all** target containers concurrently and return once they have all exited. Before this PR, `errgroup.WithContext(ctx)` gave unlimited concurrency, so every container had its own goroutine blocking on `ContainerWait` simultaneously — total elapsed time was `max(individual_lifetimes)`.

After this change, `newLimitedErrgroup(ctx, s.maxConcurrency)` caps the number of in-flight goroutines. `ContainerWait` is an indefinitely-blocking RPC: a goroutine that calls it occupies its concurrency slot until that specific container exits. With `--parallel=1`:

1. Goroutine for container[0] starts, calls `ContainerWait`, blocks until container[0] exits, releases its slot.
2. Goroutine for container[1] can only start after step 1 completes.
3. …

Total time becomes `sum(individual_lifetimes)` instead of `max(individual_lifetimes)`. In a worst-case fleet of long-running containers, this is a severe regression.

The `--parallel` flag is documented as bounding concurrent **engine calls**, not concurrent observation/monitoring loops. Indefinitely-blocking watch operations (like `ContainerWait` and log-follow streams) are best excluded from this bound, or handled with the `acquireSlot`/`releaseSlot` semaphore pattern already used in `logs.go` so the slot is released right after the RPC initiates, before it blocks.

```suggestion
	eg, waitCtx := errgroup.WithContext(ctx)
```

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 100/100 |

docker-agent (commented):

docker-agent (commented):
> **Incremental review** — this review covers only the commits pushed since `06b2184b4c8d`. Earlier commits were reviewed in the previous cycle.

### Assessment: 🟢 APPROVE

The incremental diff is clean. The shared-limiter approach via `forEachContainerWithLimiter` is correct: the errgroup-derived context is consistently threaded to both `acquireSlot` and `fn`; `acquireSlot`/`releaseSlot` are nil-safe for the unlimited case; all call sites of `stopContainers` and `removeContainers` have been updated with the new `limiter` parameter; the `inspectWithSlot` slot-transfer contract is properly maintained with `releaseSlot` on error; and the `wait.go` switch to `errgroup.WithContext` correctly removes the unnecessary cap on indefinitely-blocking `ContainerWait` calls (using `waitCtx` consistently in goroutines). The `start.go` traversal option sets `traversal.maxConcurrency` on a non-nil `graphTraversal`. Regression tests for the bounded concurrency pattern have been added across `down`, `stop`, and `start`.

The prior resolved thread on `wait.go` (serialization of `ContainerWait` under `--parallel`) is correctly addressed in this diff.

pkg/compose/dependencies.go
```diff
@@ -49,7 +49,13 @@ type graphTraversal struct {
...
 
-	visitorFn      func(context.Context, string) error
+	visitorFn func(context.Context, string) error
+	// maxConcurrency bounds concurrent node (service) visits, not concurrent
+	// engine calls: it's only a correct proxy for --parallel when visitorFn
+	// makes exactly one engine call per node (e.g. build_classic.go). A
+	// visitor that fans out multiple engine calls per node — like restart's,
+	// one per container — needs its own call-level limiter shared across
+	// nodes instead (see restart.go), or this bound is too coarse to help.
 	maxConcurrency int
```
ndeloof, line 59: Nitpick: this is the only remaining direct `eg.SetLimit(t.maxConcurrency + 1)` (line ~146) not routed through the new `newLimitedErrgroup` helper. It was already correctly guarded (`if t.maxConcurrency > 0`) before this PR, so not a bug -- just an inconsistency now that every other call site shares one helper. Could fold into `newLimitedErrgroup` too (with the `+1` applied by the caller) for a single source of truth.

ndeloof (commented):
Reviewed the diff myself and cross-checked with an independent local Copilot CLI pass. Core fix is sound: verified every claimed fan-out site (kill, pause, down, executor, logs, ps, top, images, model, watch, restart, stop, start) is now correctly routed through `newLimitedErrgroup`/a shared limiter, no unconditional `SetLimit` left. The zero-value regression test genuinely fails without the `>0` guard (checked by reverting it locally), and `TestExecutePlanRespectsMaxConcurrencyAcrossDependencyChain` correctly locks in the topological-order invariant `executor.go`'s bounded errgroup now relies on — traced it by hand, no deadlock risk.

One real gap, not inline-commentable since the file has no diff here: **`pkg/compose/cp.go` (`docker compose cp`)** still fans out via a bare, unbounded `errgroup.Group{}` (around line 82) — one `copyFunc` engine call per matched container, with no `--parallel`/`maxConcurrency` involvement at all. Same class of bug this PR fixes everywhere else; worth closing before merge or explicitly noting as a follow-up.

One nitpick inline below (`dependencies.go`).

docker-agent (commented):
### Assessment: 🟢 APPROVE

pkg/compose/model.go
```diff
@@ -49,7 +48,7 @@ func (s *composeService) ensureModels(ctx context.Context, project *types.Projec
 	defer mdlAPI.Close()
 	availableModels, err := mdlAPI.ListModels(ctx)
```
ndeloof: Two pre-existing issues here, not introduced by this PR (which only swaps the errgroup constructor a few lines below):
1. The error from `ListModels` on this line is never checked before `availableModels` is used — if it fails, every model is treated as unavailable and `PullModel` is attempted for all of them, masking the real error.
2. A few lines down, each model's goroutine does `err = mdlAPI.PullModel(...)`, reassigning this same outer `err` instead of a goroutine-local one — unlike the other fan-outs this PR bounds (ps.go, top.go, images.go all use a local `:=`). With 2+ models pulled concurrently that's a real data race: one goroutine's `nil` can be read by another right after its own failing `PullModel`, letting it proceed to `ConfigureModel` on a model that was never pulled.

ndeloof (approved):
LGTM overall — the bounded-concurrency machinery (semaphore/errgroup helpers, panic-safe slot release, shared limiters across dependency-ordered fan-outs) is solid, and build/full test suite (`-race`)/lint all pass cleanly. Not blocking, but two pre-existing issues in `ensureModels` (predate this PR, untouched by it beyond swapping the errgroup constructor) are worth a follow-up since this PR touches that exact function — left as an inline comment. One more speculative note on `executor.go`: bounding the plan executor's errgroup now makes deadlock-freedom depend on `plan.Nodes` staying strictly topologically sorted (each node's goroutine now holds its concurrency slot for the whole dependency wait, not just its own work) — already true by construction today, just flagging the coupling in case a future planner change tests it.

PR #14180, opened 2026-09-04
chore: remove github.com/buger/goterm dependency

thaJeztah:
Replace goterm's terminal size detection with golang.org/x/term, which provides the same functionality and is already an indirect dependency.

Use the existing github.com/morikuni/aec dependency for clearing the current terminal line.

**What I did**

**Related issue**

### DIFF
--- a/cmd/display/tty.go
+++ b/cmd/display/tty.go
@@ -20,10 +20,11 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"os"
 	"sync"
 	"time"
 
-	"github.com/buger/goterm"
+	"golang.org/x/term"
 
 	"github.com/docker/compose/v5/pkg/api"
 )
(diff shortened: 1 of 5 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

This PR cleanly removes the `github.com/buger/goterm` dependency by substituting:
- `golang.org/x/term` for terminal size detection (`term.GetSize`)
- `github.com/morikuni/aec` for line erasure (`aec.EraseLine(aec.EraseModes.Tail)`)

The replacement is functionally equivalent:
- `aec.EraseLine(aec.EraseModes.Tail)` emits `\033[K` which, combined with the explicit `\r`, produces the same `\r\033[K` sequence as `goterm.RESET_LINE`
- The combined fallback to `80×24` when `term.GetSize` fails is a reasonable policy (previously width and height were guarded independently, but both falling back was already the expected behaviour)
- Division-by-zero in `extraLines()` is guarded by the `width <= 0` check in `getTermSize()`
- `golang.org/x/term` is correctly promoted from indirect to direct dependency in `go.mod`

No bugs were introduced.

PR #14181, opened 2026-09-04
fix(exec): don't steal stdin when stdout is not a TTY

### LINKED ISSUE #13129: [BUG] Keypresses are swallowed when piping `docker compose exec` output to `less`
### Description

When running commands like `docker compose exec containername mysql... | less` you've got to press keys twice (or more) for them to take effect.

### Steps To Reproduce

1. `docker compose exec db find / | less` (use any running container name instead of db)
2. try to navigate with arrow keys
3. only if you manage to get to the end of the output, the keys start working normally

### Compose Version

```Text
Docker Compose version v2.39.1
```

### Docker Environment

```Text
 Version:    28.3.3
 Context:    default
 Debug Mode: false
 Plugins:
    Version:  v0.26.1
    Version:  v2.39.1

Server:
 Containers: 13
  Running: 13
  Paused: 0
  Stopped: 0
 Images: 149
 Server Version: 28.3.3
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: local
 Cgroup Driver: cgroupfs
 Cgroup Version: 2
 Plugins:
  Volume: local
 CDI spec directories:
  /etc/cdi
  /var/run/cdi
 Swarm: inactive
 Default Runtime: runc
 Init Binary: docker-init
 init version: de40ad0
 Security Options:
  apparmor
  seccomp
   Profile: builtin
  cgroupns
 Kernel Version: 6.15.8-custom+
 Operating System: Debian GNU/Linux 13 (trixie)
 OSType: linux
 Architecture: x86_64
 CPUs: 24
 Total Memory: 125.6GiB
 Name: homecomp
 ID: 351707c3-7887-42ad-af47-d492581b85d5
 Debug Mode: false
 Experimental: false
 Insecure Registries:
  ::1/128
  127.0.0.0/8
 Live Restore Enabled: false
```

### Anything else?

Works fine if I run the same command with just `docker exec` and use an unscoped container name. This is a repost of issue #11775 (closed as stale) which is a repost of issue #8428 which has been closed without any comments.

Please, exclude it from the stale bot, there's nothing to add and this issue is trivially reproduced. Is there a real need to bump it every month so it doesn't get closed?

locker95:
`docker compose exec db find / | less` keeps stdin by default, so compose and less both grab the keyboard.

`-T` already defaults off when stdout isn't a tty. do the same for `-i`.

Fixes #13129

### DIFF
--- a/cmd/compose/exec.go
+++ b/cmd/compose/exec.go
@@ -83,10 +83,11 @@ func execCommand(p *ProjectOptions, dockerCli command.Cli, backendOptions *Backe
 	runCmd.Flags().IntVar(&opts.index, "index", 0, "Index of the container if service has multiple replicas")
 	runCmd.Flags().BoolVarP(&opts.privileged, "privileged", "", false, "Give extended privileges to the process")
 	runCmd.Flags().StringVarP(&opts.user, "user", "u", "", "Run the command as this user")
-	runCmd.Flags().BoolVarP(&opts.noTty, "no-tty", "T", !dockerCli.Out().IsTerminal(), "Disable pseudo-TTY allocation. By default 'docker compose exec' allocates a TTY.")
+	isTTY := dockerCli.Out().IsTerminal()
+	runCmd.Flags().BoolVarP(&opts.noTty, "no-tty", "T", !isTTY, "Disable pseudo-TTY allocation. By default 'docker compose exec' allocates a TTY.")
 	runCmd.Flags().StringVarP(&opts.workingDir, "workdir", "w", "", "Path to workdir directory for this command")
 
-	runCmd.Flags().BoolVarP(&opts.interactive, "interactive", "i", true, "Keep STDIN open even if not attached")
+	runCmd.Flags().BoolVarP(&opts.interactive, "interactive", "i", isTTY, "Keep STDIN open even if not attached")
 	runCmd.Flags().MarkHidden("interactive") //nolint:errcheck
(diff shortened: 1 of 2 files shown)


ndeloof (commented):
`-i` (Interactive) only controls attaching the **local stdin** to the exec'd process (see `docker/cli` `parseExec`/`interactiveExec`: `Interactive` → `AttachStdin` → `in = dockerCli.In()`). `dockerCli.Out()` (stdout) plays no role in that decision — that's the signal used for `-T`/TTY, an independent concern.

This fix reuses `isTTY := dockerCli.Out().IsTerminal()` as the default for `-i`. It "fixes" the #13129 repro as a side effect, but breaks a legitimate case:

```
docker compose exec db psql | tee session.log
```
stdin = terminal (the user is actually typing), stdout is not a terminal (duplicated via `tee`) → with this patch `-i` defaults to `false`, and keystrokes silently stop being forwarded to the container.

The repo already has the right precedent for this kind of detection, based on **stdin**: `cmd/compose/run.go:186` (`!dockerCli.In().IsTerminal()` to disable auto-TTY/interactive when stdin comes from a script/pipe). It would be more consistent with `-i`'s semantics to follow that same logic here instead of reusing `-T`'s signal.

The added test only checks the flag's default value when stdout is a pipe; it's missing a case covering stdin=terminal / stdout=pipe to catch the regression above.

PR #14184, opened 2026-09-05
fix(build): don't fail bake progress when stdout is not a console

### LINKED ISSUE #14182: [BUG] failed to get console: provided file is not a console
### Description

since 5.5.1 docker compose build >/dev/null fails with the error from the title
downgrading to 5.5.0 works

### Steps To Reproduce

_No response_

### Compose Version

```Text
Docker Compose version v5.5.1
```

### Docker Environment

```Text

```

### Anything else?

_No response_

locker95:
#14090 unwraps the real stdout file so Windows TTY progress works. that also feeds a redirected stdout (`compose build >/dev/null`) into ConsoleFromFile, which then dies with `provided file is not a console`.

only unwrap when stdout actually is a terminal, and fall back to plain progress otherwise.

Fixes #14182

### DIFF
--- a/pkg/compose/build_bake.go
+++ b/pkg/compose/build_bake.go
@@ -140,7 +140,7 @@ func (s *composeService) doBuildBake(ctx context.Context, project *types.Project
 	if out == nil {
 		out = s.stdout()
 	}
-	display, err := progressui.NewDisplay(makeConsole(out), displayMode)
+	display, err := newBakeDisplay(out, displayMode)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


ndeloof:
Thanks for digging into #14182 — the analysis of the failure is right, but the fix lands in the wrong place, so we'll go with an alternative.

The root cause is a mode/stream mismatch: `display.Mode` resolves `auto` against **stderr** (deliberately — Compose's own progress renders there, and probing stdout would force plain on `up | tee`), while bake's display writes to **stdout**. `compose build >/dev/null` therefore hands bake a `tty` mode computed for a stream it doesn't write to.

Patching it in `build_bake` has two problems:
- `newBakeDisplay` silently downgrades an **explicit** `--progress=tty` to plain when stdout is not a console, where `buildx bake` (and compose's own `--progress tty` handling) fails loudly — auto should degrade, an explicit tty request should error.
- the `makeConsole` change is redundant with that fallback (progressui already falls back to plain in auto mode when the stream is not a console), and its second `IsTerminal` branch is unreachable in practice.

#14194 fixes the root cause instead: pass the unresolved `auto` to bake unless the user explicitly asked for tty, and let buildkit's progressui probe the actual output stream. That also revives the `BUILDKIT_PROGRESS` override, which could no longer trigger.

Closing in favor of #14194 — thanks again for the report and the repro.

PR #14186, opened 2026-09-05
Support file-based secrets in provider extensions

### LINKED ISSUE #14163: Support Mounts and Secrets in Compose Provider Extensions
Compose provider extensions are useful for provisioning or configuring resources that are not managed directly by Compose. They can also inject information about those resources into the application, commonly through environment variables—for example, providing managed database credentials.

However, credentials and other sensitive values are often better distributed as mounted files rather than environment variables. Environment variables can unintentionally leak through process inspection, debugging output, logs, crash reports, or other tooling.

At present, Compose providers appear to only have the ability to add or override environment variables. There does not seem to be a way for a provider to inject a mount, secret, or other file-based credential into the resulting Compose service.

## Proposed Feature

Allow Compose provider extensions to contribute file-based (secrets) resources such as, and overall align them better with existing capabilities like cdi extensions which in principle can mutate containers at runtime with additional volume and/or bind mounts.

For example, a database provider could expose credentials through a secret:

```
services:
  app:
    providers:
      database:
        type: managed-db
    secrets:
      - db_password

secrets:
  db_password:
    # Provided by the provider
```

The application could then consume the credential from the mounted secret file rather than receiving it through an environment variable.

## Motivation

This would allow providers to follow more secure secret-distribution practices while retaining the convenience of provider extensions. In particular, it would: better align the feature with security recommendations/best practices, and align the feature for images that exclusively be used with such password files.

NAVEENKUMARKR777:
## Summary
- Provider extensions can currently only inject environment variables into dependent services (`setenv`/`rawsetenv`). Credentials are often better delivered as mounted files than env vars, which can leak through process inspection, debugging output, logs and crash reports.
- Adds two new provider protocol messages, `setsecret` and `rawsetsecret`, mirroring `setenv`/`rawsetenv`: a provider can hand a dependent service a secret whose content is mounted at `/run/secrets/<name>`, without the compose file declaring it up front.
- Fixes a latent bug in `buildContainerSecretMounts`: a secret with `Content` set (no `File`) wasn't skipped before building a bind mount, so it would bind-mount the process's working directory into the container. Provider-contributed secrets are content-based, so this fix is required for the new feature to behave correctly.
- Documents the new message types in `docs/extension.md` and extends the example provider (`docs/examples/provider.go`) to emit them.

## Test plan
- [x] `go test ./pkg/compose/...` — new unit tests for message parsing (`TestExecutePlugin_ParsesSecretMessages`) and the merge/dedup logic (`TestUpsertProviderSecret`)
- [ ] New e2e scenarios (`TestProviderSetSecret`, `TestProviderRawSetSecretOverridesUserSecret`) added but not run in this environment — no Docker daemon available. Please run `go test -tags e2e ./pkg/e2e/ -run TestProviderSetSecret` / `TestProviderRawSetSecretOverridesUserSecret` before merging.

Closes #14163

### DIFF
--- a/AI_AGENT_DISCLOSURE.md
+++ b/AI_AGENT_DISCLOSURE.md
@@ -1,0 +1,4 @@
+> *"This contribution was prepared by an AI agent acting on a human's behalf.
+> The human submitter may not have independently reviewed or tested the change."*
+
+2026-09-05
--- a/docs/examples/provider.go
+++ b/docs/examples/provider.go
@@ -97,6 +97,8 @@ func up(options options, args []string) {
 	}
 	fmt.Printf(`{ "type": "setenv", "message": "URL=https://magic.cloud/%s" }%s`, servicename, lineSeparator)
 	fmt.Printf(`{ "type": "rawsetenv", "message": "CLOUD_REGION=us-east-1" }%s`, lineSeparator)
+	fmt.Printf(`{ "type": "setsecret", "message": "db_password=hunter2" }%s`, lineSeparator)
+	fmt.Printf(`{ "type": "rawsetsecret", "message": "shared_api_key=raw-secret-value" }%s`, lineSeparator)
 }
 
 func down(_ *cobra.Command, _ []string) {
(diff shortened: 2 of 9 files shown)


glours:

PR #14188, opened 2026-09-05
Support per-replica volumes for scaled services (x-per-replica)

### LINKED ISSUE #9026: docker-compose dynamic volume names for replicas
moved over from https://github.com/moby/moby/issues/43079

Tried some of the "solutions" from this thread and stackoverflow. No swarm, node or other hosts at all. Just trying to get it to work locally first without starting containers around docker-compose. Main idea is to get rid of these copies and just use replicas:

```yml
  configsvr0:
    image: mongo
    command: mongod --configsvr --replSet configsvr --port 27017 --dbpath /data/db --keyFile /data/keyfile
    volumes:
      - configsvr0:/data/db
      - ./keyfile:/data/keyfile:ro
  configsvr1:
    image: mongo
    command: mongod --configsvr --replSet configsvr --port 27017 --dbpath /data/db --keyFile /data/keyfile
    volumes:
      - configsvr1:/data/db
      - ./keyfile:/data/keyfile:ro
  configsvr2:
    image: mongo
    command: mongod --configsvr --replSet configsvr --port 27017 --dbpath /data/db --keyFile /data/keyfile
    volumes:
      - configsvr2:/data/db
      - ./keyfile:/data/keyfile:ro

volumes:
  configsvr0:
  configsvr1:
  configsvr2:
```
See the difference? It's just counting up on the volume name...

It would be great to just have something like this:

```yml
  configsvr0:
    image: mongo
    command: mongod --configsvr --replSet configsvr --port 27017 --dbpath /data/db --keyFile /data/keyfile
    volumes:
      - configsvr:/data/db
      - ./keyfile:/data/keyfile:ro

volumes:
  configsvr:
    name: 'configsvr-{{.Task.Slot}}'
```
Tried this (see below) but it didn't work.

Note: Below I only copied small junks from my yml that I thought are important for this.

### test 1 - volume name

would require to add entries under volumes so that's not really dynamic - but I gave it a try:

```yml
    volumes:
      - "configsvr{{.Task.Slot}}:/data/db"
```

```text
➜  msc git:(develop) ✗ docker-compose up configsvr0
service "configsvr0" refers to undefined volume configsvr{{.Task.Slot}}: invalid compose project
```

### test2 - dynamic name with task slot

that would be dynamic but still doesn't work though it was accepted in stackoverflow. That's actually my sample from top.

```yml
    volumes:
      - configsvr:/data/db

volumes:
  configsvr:
    name: 'configsvr-{{.Task.Slot}}'
```

```text
➜  msc git:(develop) ✗ docker-compose up configsvr0
[+] Running 0/0
 ⠿ Volume "configsvr-{{.Task.Slot}}"  Error                                                                                                                                                          0.0s
Error response from daemon: create configsvr-{{.Task.Slot}}: "configsvr-{{.Task.Slot}}" includes invalid characters for a local volume name, only "[a-zA-Z0-9][a-zA-Z0-9_.-]" are allowed. If you intended to pass a host directory, use absolute path
```

### my env

mac monteray with docker desktop.

```text
➜  msc git:(develop) ✗ docker -v
Docker version 20.10.11, build dea9396
➜  msc git:(develop) ✗ docker-compose -v
Docker Compose version v2.2.1

docker-compose.yml version: 3.9
```

NAVEENKUMARKR777:
## Summary
- Today, when a service is scaled to several replicas, every replica shares the exact same named volume — which corrupts any service that expects each replica to keep its own independent state (e.g. members of a database replica set, the original use case in the issue). The only workaround has been to hand-declare one service per replica, each with its own named volume, defeating the purpose of `replicas`/`--scale`.
- Adds an opt-in `x-per-replica: true` extension on a long-syntax service volume entry: when set, replica `N` of the service gets its own distinct volume (`<project>_<volume>-N`) instead of sharing the one declared under the top-level `volumes:` key, created on demand with the same driver, driver options and labels as the declared volume.
- `down --volumes` removes every per-replica volume actually backing the project's observed containers (derived from each container's replica-number label), in addition to the base volume name.
- This is intentionally opt-in via an `x-` extension rather than a compose-spec change or automatic behavior: changing the default for scaled named volumes would be a breaking change for anyone currently relying on replicas sharing state through one volume.
- New doc page: `docs/per-replica-volumes.md`.

## Test plan
- [x] `go test ./pkg/compose/...` — new unit tests for `buildMount`'s per-replica naming, `ensurePerReplicaVolumes` (including external-volume and bind-mount rejection), and `down --volumes` cleanup (`TestDownRemovesPerReplicaVolumes`)
- [ ] New e2e scenario (`TestPerReplicaVolumes`) added but not run in this environment — no Docker daemon available. Please run `go test -tags e2e ./pkg/e2e/ -run TestPerReplicaVolumes` before merging.

Closes #9026

### DIFF
--- a/AI_AGENT_DISCLOSURE.md
+++ b/AI_AGENT_DISCLOSURE.md
@@ -1,0 +1,4 @@
+> *"This contribution was prepared by an AI agent acting on a human's behalf.
+> The human submitter may not have independently reviewed or tested the change."*
+
+2026-09-05
--- a/docs/per-replica-volumes.md
+++ b/docs/per-replica-volumes.md
@@ -1,0 +1,62 @@
+# Per-replica volumes
+
+## Problem
+
+When a service is scaled to several replicas (`deploy.replicas` or `--scale`), every replica shares the exact
+same named volume:
+
+```yaml
+services:
+  configsvr:
+    image: mongo
+    command: mongod --configsvr --replSet configsvr
+    deploy:
+      replicas: 3
+    volumes:
+      - configsvr-data:/data/db
+
+volumes:
+  configsvr-data:
+```
+
+All three `configsvr` replicas above mount the very same `configsvr-data` volume, which corrupts any service
+that expects each replica to keep its own independent state (e.g. members of a database replica set). The only
+workaround has been to hand-declare one service per replica, each with its own named volume — defeating the
+purpose of `replicas`.
+
+## `x-per-replica`
+
+Setting `x-per-replica: true` on a long-syntax volume entry gives every replica of that service its own,
+distinct volume instead of sharing the one declared under the top-level `volumes:` key:
+
+```yaml
+services:
+  configsvr:
+    image: mongo
+    command: mongod --configsvr --replSet configsvr
+    deploy:
+      replicas: 3
+    volumes:
+      - type: volume
+        source: configsvr-data
+        target: /data/db
+        x-per-replica: true
+
+volumes:
+  configsvr-data:
+```
+
+Replica `N` of `configsvr` gets a volume named `<project>_configsvr-data-N` (`myproject_configsvr-data-1`,
+`myproject_configsvr-data-2`, `myproject_configsvr-data-3`), created on demand with the same driver, driver
+options and labels as the declared `configsvr-data` volume. Replicas keep the same volume across restarts and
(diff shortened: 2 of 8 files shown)


glours:

PR #14189, opened 2026-09-05
Add secrets/configs driver: copy, for use with a remote Docker host

### LINKED ISSUE #11867: Support remote DOCKER_HOST by copying configs and secrets files instead of bind mounting them

NAVEENKUMARKR777:
## Summary
This has been open since May 2024 with 50+ comments and three separate community PRs (#11871/#12251 by @andoks, #11984/#12448 by @schaubl, and the narrower #12625) all closed without merging. The problem: `secrets.<name>.file`/`configs.<name>.file` are interpreted as a path on the **Docker host**, so Compose bind-mounts it into the container — which breaks entirely once `DOCKER_HOST`/a docker context points at a remote engine, since Compose is a client-side tool and cannot bind-mount a path that only exists on the client.

Every prior attempt switched the behavior **unconditionally** based on `file:` being set, and was rejected for the same reason each time. From @ndeloof's review on #12448:

> This PR doesn't address comments on #11867 tl;dr: config/secret file MAY be mounted from (remote) docker host. Many users rely on this, some do even require changes to the original file to be propagated inside container. So we can't just switch to a distinct behavior which would introduce a breaking change. This would at least require an explicit opt-in flag in compose.yaml

This PR adds exactly that: **`driver: copy`** as an explicit opt-in.

```yaml
secrets:
  db_password:
    file: ./db_password.txt
    driver: copy
```

When set, the file is read from the **client's** local filesystem and copied into the container once, at creation — reusing the content-injection path Compose already has for `content:`-based secrets/configs (`copyFileToContainer`/`createTar`) — instead of bind-mounted. Without `driver: copy`, behavior is completely unchanged: existing users who rely on the bind mount (editing the file on the Docker host directly, or `watch`-driven sync) are unaffected. Any other `driver` value is still rejected, as before.

New doc page: `docs/secrets-and-configs-driver-copy.md`.

## Test plan
- [x] `go test ./pkg/compose/...` — new unit tests for `resolveFileContent`'s copy-driver path and precedence (`Content`/`Environment` still win), and for the bind-mount-skip/validation logic in `buildContainerSecretMounts`/`buildContainerConfigMounts`
- [ ] New e2e scenario (`TestSecretCopyDriver`) added but not run in this environment — no Docker daemon available. It specifically verifies the copy is a one-time snapshot (editing the client's file after `up` does not reach the running container), the key behavioral difference from a bind mount. Please run `go test -tags e2e ./pkg/e2e/ -run TestSecretCopyDriver` before merging.

Closes #11867

### DIFF
--- a/AI_AGENT_DISCLOSURE.md
+++ b/AI_AGENT_DISCLOSURE.md
@@ -1,0 +1,4 @@
+> *"This contribution was prepared by an AI agent acting on a human's behalf.
+> The human submitter may not have independently reviewed or tested the change."*
+
+2026-09-05
--- a/docs/secrets-and-configs-driver-copy.md
+++ b/docs/secrets-and-configs-driver-copy.md
@@ -1,0 +1,48 @@
+# `secrets`/`configs`: `driver: copy`
+
+## Problem
+
+`secrets.<name>.file` and `configs.<name>.file` are, by default, interpreted as a path on the **Docker host**:
+Compose creates a read-only bind mount from that host path into the container. This works fine when the
+daemon is local, but breaks entirely once `DOCKER_HOST` or a Docker context points at a remote engine —
+Compose is a client-side tool and cannot bind-mount a path that only exists on the client's machine into a
+container running on a different host.
+
+```yaml
+services:
+  app:
+    secrets:
+      - db_password
+
+secrets:
+  db_password:
+    file: ./db_password.txt   # only works if the daemon is local
+```
+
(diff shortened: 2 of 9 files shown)


glours:

PR #14190, opened 2026-09-06
refactor(compatibility): hoist flag/command lists to package vars

htoyoda18:
**What I did**

`getCompletionCommands()` / `getBoolFlags()` / `getStringFlags()` in
`cmd/compatibility/convert.go` rebuilt their slice literals on every
call, and `Convert` calls them once per parsed argument in its main
loop. This moves the three lists to package-level `var`s so they are
allocated once instead of repeatedly. No behavior change: call sites
now reference the vars directly with `slices.Contains`/range instead
of calling the getter functions.

**Related issue**

image

### DIFF
--- a/cmd/compatibility/convert.go
+++ b/cmd/compatibility/convert.go
@@ -25,31 +25,25 @@ import (
 	"github.com/docker/compose/v5/cmd/compose"
 )
 
-func getCompletionCommands() []string {
-	return []string{
-		"__complete",
-		"__completeNoDesc",
-	}
+var completionCommands = []string{
+	"__complete",
+	"__completeNoDesc",
 }
 
-func getBoolFlags() []string {
-	return []string{
-		"--debug", "-D",
-		"--verbose",
-		"--tls",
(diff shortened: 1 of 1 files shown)


glours (changes requested):
This introduces an unconditional cost where none is needed: `compatibility` is imported
unconditionally in `cmd/main.go`, so these package-level slices now get allocated at process init for **every** invocation, but `Convert` (and these lists) only ever runs when `plugin.RunningStandalone()` is true. For every normal `docker compose <cmd>` invocation.
That's pure waste with zero payoff, whereas before, the getters simply weren't called on that path.

Not sure this micro-optimization is useful honestly

htoyoda18:
Thanks for pointing this out. I overlooked the fact that `compatibility` is always imported while `Convert` is only used in standalone mode.

Given that, this change shifts a small allocation cost from the standalone path to every invocation, so I agree that the optimization isn't worthwhile.

I'll close this PR. Thanks for the review!

PR #14193, opened 2026-09-06
feat(provider): publish-endpoint deploys a network relay for provider services

ndeloof:
A provider's resource lives outside the compose network: consumers could only reach it through injected variables carrying a host-published address. This adds the transparent path:

- **`publish-endpoint` provider message**: `{"type": "publish-endpoint", "message": "80=localhost:49152"}` — container port consumers know on the left, real location on the right. Providers express that location naturally from the **host's perspective**; compose owns the translation to the relay's vantage point, rewriting host-relative addresses (`localhost`, any loopback IP, unspecified or empty host) to `host.docker.internal` — provisioned through ExtraHosts (`host-gateway`) so it also works on a plain Linux engine. LAN IPs and DNS names pass verbatim.
- **Relay container**: when endpoints are published, compose deploys `docker/compose-relay` (new `relay/` directory: static Go TCP forwarder, `FROM scratch`, bake target `relay-image`, `COMPOSE_RELAY_IMAGE` override for internal registries) **in place of the service** — canonical `<project>-<service>-1` name, service alias on the networks of the depending services. Consumers use `http://<service>:<port>`, no injected variables involved.
- **First-class project container, full lifecycle**: standard compose labels incl. config-hash, so label-driven commands run without the compose file (`ps`, `logs`, `stop`, `down`) keep seeing the service — plus `com.docker.compose.relay` (value: identity hash of image+routes) declaring its role:
  - `up` idempotent: relay kept when routes unchanged, recreated otherwise (reconciler already leaves provider services' containers alone);
  - `down` treats the relay as part of the provider service's deprovisioning: its containers are removed **before** the plugin removes the provider's resource — mirroring `up`, which provisions the resource before deploying the relay — so `down -v` releases the project network instead of failing on "resource is still in use";
  - `exec`/`cp` refuse a relay (no service process to act on);
  - the attached-`up` monitor excludes relays from the containers whose termination ends the command — they are long-lived infrastructure.

This benefits every provider: the example provider demonstrates the flow (detached HTTP helper + publish-endpoint), covered by an e2e scenario asserting `http://db` works from a consumer, `exec db` is refused, and `down -v` leaves no relay behind, plus unit tests (message parsing, upstream rewriting, route identity, network selection, relay guard, image override). A provider message compose cannot decode is now reported as an "invalid message" rather than an "invalid response" — providers push messages, they don't answer requests.

Standalone — no dependency on #14175 (the two protocol additions are orthogonal).

Publishing note for maintainers: the CI wiring is included — `merge.yml` publishes `docker/compose-relay` through the same github-builder pipeline as `docker/compose-bin` (`edge` from main; the `v1` tag consumers pull moves only on compose release tags, since v1 is the image contract major, not a compose version), and PRs validate the cross build (`relay-image-test`). The one remaining step is org-side: create the `docker/compose-relay` Docker Hub repository and allow the existing registry identity to push to it.

### DIFF
--- a/docs/examples/provider.go
+++ b/docs/examples/provider.go
@@ -193,7 +193,10 @@ func up(options options, args []string) {
 		// through it only after up has returned, so reaping it here would
 		// tear the endpoint down before anyone reached it. Its lifetime is
 		// its own: it exits by itself after three minutes (serve-demo), the
-		// way a real provider's resource outlives the provider CLI run.
+		// way a real provider's resource outlives the provider CLI run. No
+		// Wait() and no zombie either: this process exits within seconds, so
+		// the subprocess is long re-parented to init — which reaps it — when
+		// its three minutes are up.
 		//
 		// the endpoint is announced as seen from THIS process's host —
 		// the relay translates loopback into the container-visible name
--- a/docs/examples/provider.go
+++ b/docs/examples/provider.go
@@ -203,7 +203,12 @@ func up(options options, args []string) {
 }
 
 func down(_ *cobra.Command, _ []string) {
-	fmt.Printf(`{ "type": "error", "message": "Permission error" }%s`, lineSeparator)
+	// A failing down can be simulated for tests and demos.
+	if os.Getenv("PROVIDER_DOWN_FAILURE") != "" {
+		fmt.Printf(`{ "type": "error", "message": "Permission error" }%s`, lineSeparator)
+		return
+	}
+	fmt.Printf(`{ "type": "info", "message": "Resource removed" }%s`, lineSeparator)
 }
 
 func stop(_ *cobra.Command, _ []string) {
--- a/docs/examples/provider.go
+++ b/docs/examples/provider.go
@@ -188,6 +188,13 @@ func up(options options, args []string) {
 			fmt.Printf(`{ "type": "error", "message": "demo endpoint did not come up: %v" }%s`, err, lineSeparator)
 			return
 		}
+		// The subprocess deliberately outlives this invocation — it IS the
+		// provisioned resource the relay forwards to, and consumers connect
+		// through it only after up has returned, so reaping it here would
+		// tear the endpoint down before anyone reached it. Its lifetime is
+		// its own: it exits by itself after three minutes (serve-demo), the
+		// way a real provider's resource outlives the provider CLI run.
+		//
 		// the endpoint is announced as seen from THIS process's host —
 		// the relay translates loopback into the container-visible name
 		_, port, _ := net.SplitHostPort(strings.TrimSpace(addr))
--- a/docs/examples/provider.go
+++ b/docs/examples/provider.go
@@ -17,9 +17,13 @@
 package main
 
 import (
+	"bufio"
 	"encoding/json"
 	"fmt"
+	"net"
 	"os"
+	"os/exec"
+	"strings"
 	"time"
 
 	"github.com/spf13/cobra"
@@ -32,6 +36,7 @@ func main() {
 		Use:   "demo",
 	}
 	cmd.AddCommand(composeCommand())
+	cmd.AddCommand(serveDemoCommand())
 	err := cmd.Execute()
 	if err != nil {
 		fmt.Fprintln(os.Stderr, err)
@@ -85,6 +90,44 @@ func composeCommand() *cobra.Command {
 	return c
 }
 
+// serveDemoCommand is the detached helper process behind the
+// publish-endpoint demonstration: a TCP server on the given address
+// answering every connection with a fixed HTTP response, exiting on its own
+// after three minutes. It owns the port from bind to exit: the bound address
+// is reported on stdout once listening, so the parent never has to probe or
+// pre-reserve the port (no TOCTOU window, and works on Windows where handing
+// a socket over ExtraFiles is not supported).
+func serveDemoCommand() *cobra.Command {
+	return &cobra.Command{
+		Use:    "serve-demo ADDR",
+		Hidden: true,
+		Args:   cobra.ExactArgs(1),
+		RunE: func(_ *cobra.Command, args []string) error {
+			listener, err := net.Listen("tcp", args[0])
+			if err != nil {
+				return err
+			}
+			fmt.Println(listener.Addr().String())
+			go func() {
+				time.Sleep(3 * time.Minute)
+				os.Exit(0)
+			}()
+			for {
+				conn, err := listener.Accept()
+				if err != nil {
+					return err
+				}
+				go func() {
+					defer func() { _ = conn.Close() }()
+					buf := make([]byte, 1024)
+					_, _ = conn.Read(buf)
+					_, _ = conn.Write([]byte("HTTP/1.1 200 OK\r\nContent-Length: 19\r\nConnection: close\r\n\r\nhello from provider"))
+				}()
+			}
+		},
+	}
+}
+
 const lineSeparator = "\n"
 
 func up(options options, args []string) {
@@ -114,6 +157,42 @@ func up(options options, args []string) {
 	// encode the message instead of interpolating it into a JSON literal
 	setenv, _ := json.Marshal(map[string]string{"type": "setenv", "message": "CONFIG_TYPE=" + config.Provider.Type})
 	fmt.Println(string(setenv))
+
+	// When asked to, stand up a real endpoint on the host and publish it, so
+	// compose deploys a relay and consumers reach it as http://<service>:80.
+	if os.Getenv("PROVIDER_DEMO_ENDPOINT") != "" {
+		// The subprocess binds the port itself and reports the resulting
+		// address on its stdout; only then is the endpoint published. This
+		// avoids the two races of a pre-reserved port: another process
+		// grabbing it between release and re-bind, and publish-endpoint
+		// pointing at a server that is not listening yet.
+		// All interfaces, not loopback: on a plain Linux engine host-gateway
+		// is the bridge IP, which cannot reach a host loopback bind.
+		server := exec.Command(os.Args[0], "serve-demo", "0.0.0.0:0")
+		stdout, err := server.StdoutPipe()
+		if err != nil {
+			fmt.Printf(`{ "type": "error", "message": "demo endpoint: %v" }%s`, err, lineSeparator)
+			return
+		}
+		if err := server.Start(); err != nil {
+			fmt.Printf(`{ "type": "error", "message": "demo endpoint: %v" }%s`, err, lineSeparator)
+			return
+		}
+		// A crashed subprocess closes the pipe (EOF below); a hung one would
+		// block the read forever, so kill it after a deadline — the read then
+		// fails with EOF and lands on the same error path.
+		watchdog := time.AfterFunc(30*time.Second, func() { _ = server.Process.Kill() })
+		addr, err := bufio.NewReader(stdout).ReadString('\n')
+		watchdog.Stop()
+		if err != nil {
+			fmt.Printf(`{ "type": "error", "message": "demo endpoint did not come up: %v" }%s`, err, lineSeparator)
+			return
+		}
+		// the endpoint is announced as seen from THIS process's host —
+		// the relay translates loopback into the container-visible name
+		_, port, _ := net.SplitHostPort(strings.TrimSpace(addr))
+		fmt.Printf(`{ "type": "publish-endpoint", "message": "80=localhost:%s" }%s`, port, lineSeparator)
+	}
 
 	for i := 0; i < options.size; i += 10 {
 		time.Sleep(1 * time.Second)
--- a/pkg/compose/cp.go
+++ b/pkg/compose/cp.go
@@ -121,6 +121,9 @@ func (s *composeService) listContainersTargetedForCopy(ctx context.Context, proj
 		if err != nil {
 			return nil, err
 		}
+		if err := checkRelayTarget(ctr, serviceName, "cp"); err != nil {
+			return nil, err
+		}
 		return append(containers, ctr), nil
 	default:
 		withOneOff := oneOffExclude
@@ -130,6 +133,11 @@ func (s *composeService) listContainersTargetedForCopy(ctx context.Context, proj
 		containers, err = s.getContainers(ctx, projectName, withOneOff, true, serviceName)
 		if err != nil {
 			return nil, err
+		}
+		for _, ctr := range containers {
+			if err := checkRelayTarget(ctr, serviceName, "cp"); err != nil {
+				return nil, err
+			}
 		}
 
 		if len(containers) < 1 {
--- a/pkg/compose/plugins.go
+++ b/pkg/compose/plugins.go
@@ -23,9 +23,11 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"net"
 	"os"
 	"os/exec"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"sync"
 
@@ -50,6 +52,7 @@ const (
 	SetEnvType                = "setenv"
 	RawSetEnvType             = "rawsetenv"
 	DebugType                 = "debug"
+	PublishEndpointType       = "publish-endpoint"
 	providerMetadataDirectory = "compose/providers"
 
 	// GetServiceConfigType is a message the provider sends to receive, on
@@ -61,6 +64,11 @@ const (
 type pluginVariables struct {
 	prefixed types.Mapping
 	raw      types.Mapping
+	// endpoints are "port=host:port" publish-endpoint messages: container
+	// port the consumers know, mapped to where the provider's resource
+	// actually listens. When present, compose deploys a relay container
+	// under the service's name on the consumers' networks.
+	endpoints map[int]string
 }
 
 var mux sync.Mutex
@@ -107,7 +115,28 @@ func (s *composeService) runPlugin(ctx context.Context, project *types.Project, 
 			project.Services[name] = s
 		}
 	}
+	if command == "up" && len(variables.endpoints) > 0 {
+		if err := s.ensureServiceRelay(ctx, project, service, variables.endpoints); err != nil {
+			return err
+		}
+	}
 	return nil
+}
+
+// parseEndpointMessage decodes a publish-endpoint payload: "80=host:port".
+func parseEndpointMessage(message string) (int, string, error) {
+	portPart, upstream, found := strings.Cut(message, "=")
+	if !found {
+		return 0, "", fmt.Errorf("publish-endpoint %q: want port=host:port", message)
+	}
+	port, err := strconv.Atoi(portPart)
+	if err != nil || port < 1 || port > 65535 {
+		return 0, "", fmt.Errorf("publish-endpoint %q: invalid port", message)
+	}
+	if _, _, err := net.SplitHostPort(upstream); err != nil {
+		return 0, "", fmt.Errorf("publish-endpoint %q: invalid endpoint: %w", message, err)
+	}
+	return port, upstream, nil
 }
 
 func (s *composeService) executePlugin(cmd *exec.Cmd, command string, service types.ServiceConfig) (pluginVariables, error) {
@@ -180,8 +209,9 @@ func (s *composeService) executePlugin(cmd *exec.Cmd, command string, service ty
 	defer func() { _ = stdout.Close() }()
 
 	variables := pluginVariables{
-		prefixed: types.Mapping{},
-		raw:      types.Mapping{},
+		prefixed:  types.Mapping{},
+		raw:       types.Mapping{},
+		endpoints: map[int]string{},
 	}
 
 	for {
@@ -224,6 +254,12 @@ func (s *composeService) executePlugin(cmd *exec.Cmd, command string, service ty
 				defer stdinMu.Unlock()
 				_, _ = stdin.Write(payload)
 			}()
+		case PublishEndpointType:
+			port, upstream, err := parseEndpointMessage(msg.Message)
+			if err != nil {
+				return pluginVariables{}, fmt.Errorf("invalid response from plugin: %w", err)
+			}
+			variables.endpoints[port] = upstream
 		case DebugType:
 			logrus.Debugf("%s: %s", service.Name, msg.Message)
 		default:
--- a/pkg/compose/plugins.go
+++ b/pkg/compose/plugins.go
@@ -232,13 +232,13 @@ func (s *composeService) executePlugin(cmd *exec.Cmd, command string, service ty
 		case SetEnvType:
 			key, val, found := strings.Cut(msg.Message, "=")
 			if !found {
-				return pluginVariables{}, fmt.Errorf("invalid response from plugin: %s", msg.Message)
+				return pluginVariables{}, fmt.Errorf("invalid message from plugin: %s", msg.Message)
 			}
 			variables.prefixed[key] = val
 		case RawSetEnvType:
 			key, val, found := strings.Cut(msg.Message, "=")
 			if !found {
-				return pluginVariables{}, fmt.Errorf("invalid response from plugin: %s", msg.Message)
+				return pluginVariables{}, fmt.Errorf("invalid message from plugin: %s", msg.Message)
 			}
 			variables.raw[key] = val
 		case GetServiceConfigType:
@@ -257,13 +257,13 @@ func (s *composeService) executePlugin(cmd *exec.Cmd, command string, service ty
 		case PublishEndpointType:
 			port, upstream, err := parseEndpointMessage(msg.Message)
 			if err != nil {
-				return pluginVariables{}, fmt.Errorf("invalid response from plugin: %w", err)
+				return pluginVariables{}, fmt.Errorf("invalid message from plugin: %w", err)
 			}
 			variables.endpoints[port] = upstream
 		case DebugType:
 			logrus.Debugf("%s: %s", service.Name, msg.Message)
 		default:
-			return pluginVariables{}, fmt.Errorf("invalid response from plugin: %s", msg.Type)
+			return pluginVariables{}, fmt.Errorf("invalid message from plugin: %s", msg.Type)
 		}
 	}
 
--- a/pkg/compose/plugins.go
+++ b/pkg/compose/plugins.go
@@ -98,8 +98,15 @@ func (s *composeService) runPlugin(ctx context.Context, project *types.Project, 
 		return nil
 	}
 
+	deployRelay := command == "up" && len(variables.endpoints) > 0
+
+	// project.Services is shared state mutated by every concurrent provider
+	// run: the env-var injection below writes it, and the relay's network
+	// selection reads it — both belong under the mutex. The Docker API work
+	// in ensureServiceRelay does not: holding the lock across it would make
+	// every concurrent provider wait on the slowest one (image pull
+	// included), so the relay is deployed after the lock is released.
 	mux.Lock()
-	defer mux.Unlock()
 	for name, s := range project.Services {
 		if _, ok := s.DependsOn[service.Name]; ok {
 			prefix := strings.ToUpper(service.Name) + "_"
@@ -115,10 +122,14 @@ func (s *composeService) runPlugin(ctx context.Context, project *types.Project, 
 			project.Services[name] = s
 		}
 	}
-	if command == "up" && len(variables.endpoints) > 0 {
-		if err := s.ensureServiceRelay(ctx, project, service, variables.endpoints); err != nil {
-			return err
-		}
+	var networkKeys []string
+	if deployRelay {
+		networkKeys = relayNetworks(project, service)
+	}
+	mux.Unlock()
+
+	if deployRelay {
+		return s.ensureServiceRelay(ctx, project, service, variables.endpoints, networkKeys)
 	}
 	return nil
 }
--- a/pkg/compose/relay.go
+++ b/pkg/compose/relay.go
@@ -21,6 +21,7 @@ import (
 	"crypto/sha256"
 	"encoding/hex"
 	"fmt"
+	"net"
 	"os"
 	"sort"
 	"strings"
@@ -50,6 +51,8 @@ func relayImage() string {
 
 // relayRoutesSpec renders endpoints as the relay's RELAY_ROUTES value,
 // canonically ordered so it doubles as the identity the relay label hashes.
+// Upstreams are rewritten for the relay's vantage point (relayUpstream)
+// before rendering, so the identity follows what the relay actually dials.
 func relayRoutesSpec(endpoints map[int]string) string {
 	ports := make([]int, 0, len(endpoints))
 	for port := range endpoints {
@@ -58,9 +61,33 @@ func relayRoutesSpec(endpoints map[int]string) string {
 	sort.Ints(ports)
 	routes := make([]string, 0, len(ports))
 	for _, port := range ports {
-		routes = append(routes, fmt.Sprintf("%d=%s", port, endpoints[port]))
+		routes = append(routes, fmt.Sprintf("%d=%s", port, relayUpstream(endpoints[port])))
 	}
 	return strings.Join(routes, ",")
+}
+
+// relayUpstream rewrites a host-relative upstream for the relay's vantage
+// point. Providers express endpoints from the host's perspective — a loopback
+// or unspecified address names the machine compose runs on — but the relay
+// dials from its own network namespace, where those addresses name the relay
+// container itself. host.docker.internal resolves natively on Docker Desktop
+// and is provisioned through ExtraHosts (host-gateway) on plain Linux
+// engines. Anything else (a LAN IP, a DNS name) is reachable as-is from the
+// relay and passes verbatim.
+func relayUpstream(endpoint string) string {
+	host, port, err := net.SplitHostPort(endpoint)
+	if err != nil {
+		// validated at parse time (parseEndpointMessage); keep verbatim
+		return endpoint
+	}
+	hostRelative := host == "" || strings.EqualFold(host, "localhost")
+	if ip := net.ParseIP(host); ip != nil && (ip.IsLoopback() || ip.IsUnspecified()) {
+		hostRelative = true
+	}
+	if !hostRelative {
+		return endpoint
+	}
+	return net.JoinHostPort("host.docker.internal", port)
 }
 
 func relayIdentity(routes string) string {
--- a/pkg/compose/relay.go
+++ b/pkg/compose/relay.go
@@ -133,7 +133,11 @@ func relayNetworks(project *types.Project, service types.ServiceConfig) []string
 // treat it as the service, plus the RelayLabel identifying its role — the
 // reconciler leaves provider services' containers alone, and process-level
 // commands (exec) refuse it.
-func (s *composeService) ensureServiceRelay(ctx context.Context, project *types.Project, service types.ServiceConfig, endpoints map[int]string) error {
+//
+// networkKeys (see relayNetworks) is computed by the caller under the shared
+// project mutex: this function performs only Docker API work and must not
+// touch project.Services, which concurrent provider runs mutate.
+func (s *composeService) ensureServiceRelay(ctx context.Context, project *types.Project, service types.ServiceConfig, endpoints map[int]string, networkKeys []string) error {
 	routes := relayRoutesSpec(endpoints)
 	identity := relayIdentity(routes)
 	name := getContainerName(project.Name, service, 1)
@@ -175,7 +179,6 @@ func (s *composeService) ensureServiceRelay(ctx context.Context, project *types.
 		}
 	}
 
-	networkKeys := relayNetworks(project, service)
 	if len(networkKeys) == 0 {
 		logrus.Warnf("service %q published endpoints but no service depends on it and the project has no default network; skipping relay", service.Name)
 		return nil
--- a/pkg/compose/relay.go
+++ b/pkg/compose/relay.go
@@ -1,0 +1,315 @@
+/*
+   Copyright 2020 Docker Compose CLI authors
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package compose
+
+import (
+	"context"
+	"crypto/sha256"
+	"encoding/hex"
+	"fmt"
+	"os"
+	"sort"
+	"strings"
+	"time"
+
+	"github.com/compose-spec/compose-go/v2/types"
+	"github.com/containerd/errdefs"
+	"github.com/moby/moby/api/types/container"
+	"github.com/moby/moby/api/types/network"
+	"github.com/moby/moby/client"
+	"github.com/sirupsen/logrus"
+
+	"github.com/docker/compose/v5/pkg/api"
+)
+
+// defaultRelayImage is the published network-relay image deployed in place of
+// a provider service that published endpoints. Overridable for development
+// and air-gapped setups with COMPOSE_RELAY_IMAGE.
+const defaultRelayImage = "docker/compose-relay:v1"
+
+func relayImage() string {
+	if img := os.Getenv("COMPOSE_RELAY_IMAGE"); img != "" {
+		return img
+	}
+	return defaultRelayImage
+}
+
+// relayRoutesSpec renders endpoints as the relay's RELAY_ROUTES value,
+// canonically ordered so it doubles as the identity the relay label hashes.
+func relayRoutesSpec(endpoints map[int]string) string {
+	ports := make([]int, 0, len(endpoints))
+	for port := range endpoints {
+		ports = append(ports, port)
+	}
+	sort.Ints(ports)
+	routes := make([]string, 0, len(ports))
+	for _, port := range ports {
+		routes = append(routes, fmt.Sprintf("%d=%s", port, endpoints[port]))
+	}
+	return strings.Join(routes, ",")
+}
+
+func relayIdentity(routes string) string {
+	digest := sha256.Sum256([]byte(relayImage() + "|" + routes))
+	return hex.EncodeToString(digest[:])[:12]
+}
+
+// relayNetworks returns the compose network keys the relay must join: the
+// union of the networks of every service depending on the provider service —
+// the consumers the relay exists for — falling back to the project default.
+func relayNetworks(project *types.Project, service types.ServiceConfig) []string {
+	set := map[string]bool{}
+	for _, s := range project.Services {
+		if _, ok := s.DependsOn[service.Name]; !ok {
+			continue
+		}
+		for key := range s.Networks {
+			// a resolved project declares every service network, but guard
+			// anyway: an unknown key would yield an empty network name later
+			if _, ok := project.Networks[key]; ok {
+				set[key] = true
+			}
+		}
+	}
+	if len(set) == 0 {
+		if _, ok := project.Networks["default"]; ok {
+			set["default"] = true
+		}
+	}
+	keys := make([]string, 0, len(set))
+	for key := range set {
+		keys = append(keys, key)
+	}
+	sort.Strings(keys)
+	return keys
+}
+
+// ensureServiceRelay converges the relay container standing in for a provider
+// service that published endpoints: consumers reach the provider's resource
+// at the compose-native address (http://<service>:<port>) through it. The
+// relay is a regular project container (standard compose labels, canonical
+// name, service alias on the consumers' networks) so label-driven commands
+// treat it as the service, plus the RelayLabel identifying its role — the
+// reconciler leaves provider services' containers alone, and process-level
+// commands (exec) refuse it.
+func (s *composeService) ensureServiceRelay(ctx context.Context, project *types.Project, service types.ServiceConfig, endpoints map[int]string) error {
+	routes := relayRoutesSpec(endpoints)
+	identity := relayIdentity(routes)
+	name := getContainerName(project.Name, service, 1)
+
+	existing, err := s.findRelayContainer(ctx, project.Name, service.Name)
+	if err != nil {
+		return err
+	}
+	if existing != nil {
+		if existing.Labels[api.RelayLabel] == identity {
+			switch existing.State {
+			case container.StateRunning, container.StateRestarting:
+				// Up to date; restarting means Docker is already recovering
+				// it — recreating would tear down in-flight connections.
+				return nil
+			case container.StateCreated, container.StateExited:
+				// Routes unchanged: start the existing container rather than
+				// recreating it.
+				if _, err := s.apiClient().ContainerStart(ctx, existing.ID, client.ContainerStartOptions{}); err != nil {
+					return fmt.Errorf("start relay for service %s: %w", service.Name, err)
+				}
+				return nil
+			case container.StatePaused:
+				if _, err := s.apiClient().ContainerUnpause(ctx, existing.ID, client.ContainerUnpauseOptions{}); err != nil {
+					return fmt.Errorf("unpause relay for service %s: %w", service.Name, err)
+				}
+				return nil
+			}
+		}
+		if existing.State == container.StateRemoving {
+			// the daemon is already removing it: a concurrent ContainerRemove
+			// fails with "removal already in progress", so wait for the name
+			// to free up instead
+			if err := s.waitRelayRemoved(ctx, project.Name, service.Name); err != nil {
+				return err
+			}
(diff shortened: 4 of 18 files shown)


docs/examples/provider.go
```diff
@@ -85,12 +88,81 @@ func composeCommand() *cobra.Command {
...
+	// When asked to, stand up a real endpoint on the host and publish it, so
+	// compose deploys a relay and consumers reach it as http://<service>:80.
+	if os.Getenv("PROVIDER_DEMO_ENDPOINT") != "" {
+		listener, err := net.Listen("tcp", "127.0.0.1:0")
+		if err != nil {
+			fmt.Printf(`{ "type": "error", "message": "demo endpoint: %v" }%s`, err, lineSeparator)
+			return
+		}
+		addr := listener.Addr().String()
+		_ = listener.Close()
```
docker-agent: **[low] TOCTOU race: port released before subprocess can bind it**

In the `up` function, `net.Listen("tcp", "127.0.0.1:0")` is used to discover a free ephemeral port. The listener is immediately closed and the port address is handed to the `serve-demo` subprocess. Two independent problems arise:

1. **Port stolen between close and subprocess bind**: after `listener.Close()` releases the OS port, any other process on the host may grab it before the `serve-demo` subprocess reaches its own `net.Listen("tcp", args[0])` call.
2. **`publish-endpoint` sent before subprocess is ready**: `server.Start()` only confirms the subprocess was spawned — it does *not* wait for the subprocess to successfully bind the port. The `publish-endpoint` message (line 163) is sent immediately after `Start()` returns, so the relay is wired to an address the demo server may not yet be listening on.

If either race is lost, the relay points to a dead endpoint and consumer requests fail silently.

**Fix**: keep the listener open and pass the file descriptor to the subprocess, letting it call `net.FileListener` instead of `net.Listen`:
```go
// in up():
f, _ := listener.(*net.TCPListener).File()
server := exec.Command(os.Args[0], "serve-demo")
server.ExtraFiles = []*os.File{f}
_ = listener.Close()   // subprocess now owns the FD

// in serveDemoCommand RunE:
listener, err := net.FileListener(os.NewFile(3, "tcp"))
```
This transfers the already-bound socket atomically — no window for another process to steal the port, and the socket is ready before the parent sends `publish-endpoint`.

Note: this is example/demo code in `docs/examples/`, so production compose users are unaffected.

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 100/100 |

docker-agent (commented):
### Assessment: 🟢 APPROVE

This PR introduces a well-structured relay mechanism for provider services — the core logic in `relay.go`, `plugins.go`, and `relay/main.go` is sound. One low-severity issue was found in the example/demo code only.

docs/examples/provider.go
ndeloof: Fixed in 893cb95f0: the subprocess now binds the port itself ("serve-demo 0.0.0.0:0") and reports the bound address on its stdout; the parent reads it before emitting publish-endpoint. This removes both races (no release/re-bind window, and the endpoint is guaranteed listening when published). Went with self-bind + readiness-report rather than ExtraFiles, which is not supported on Windows.

docker-agent: Perfect fix — the readiness-report pattern eliminates both races cleanly. The subprocess owns the port from bind to exit, and the parent blocks on reading the actual bound address before publishing, so the endpoint is guaranteed listening when the relay is wired up. Great call on avoiding `ExtraFiles` for Windows portability; the stdout handshake is much more robust across platforms.

Thanks for the fast turnaround on this!

docker-agent (commented):

Two findings in the new relay code introduced by this PR.

pkg/compose/relay.go
```diff
@@ -0,0 +1,257 @@
...
+	routes := relayRoutesSpec(endpoints)
+	identity := relayIdentity(routes)
+	name := getContainerName(project.Name, service, 1)
+
+	existing, err := s.findRelayContainer(ctx, project.Name, service.Name)
+	if err != nil {
+		return err
+	}
+	if existing != nil {
+		if existing.Labels[api.RelayLabel] == identity && existing.State == container.StateRunning {
```
docker-agent: **[medium] Idempotency check accepts only `StateRunning`; relay in `"restarting"` state is spuriously force-removed**

The relay container is created with `RestartPolicyUnlessStopped`. If `compose up` is re-run while the relay is momentarily restarting (e.g., it briefly crashed), the idempotency guard `existing.State == container.StateRunning` evaluates to `false` even though the routes are identical. The code falls through to `ContainerRemove(..., Force: true)` followed by a full recreate, tearing down all in-progress TCP sessions unnecessarily.

Valid non-stale states that should be treated as up-to-date (routes unchanged):
- `container.StateRestarting` — Docker is already recovering the relay
- `container.StateCreated` — just created, not yet started (e.g., a previous `up` call interrupted between create and start)

**Suggested fix** — expand the keep condition:

```suggestion
		if existing.Labels[api.RelayLabel] == identity &&
			(existing.State == container.StateRunning || existing.State == container.StateRestarting || existing.State == container.StateCreated) {
```

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 97/100 |

relay/main.go
```diff
@@ -0,0 +1,136 @@
...
+
+func serve(ctx context.Context, listener net.Listener, upstream string) {
+	for {
+		conn, err := listener.Accept()
+		if err != nil {
+			if ctx.Err() != nil {
+				return
+			}
+			log.Printf("accept on %s: %v", listener.Addr(), err)
+			continue
```
docker-agent, line 141: **[low] `serve()` busy-loops and floods logs when `Accept()` returns a permanent error**

On any `Accept()` error where `ctx.Err()` is not set (e.g., `EMFILE` — "too many open files" — when the container FD limit is hit), the loop immediately continues: no sleep, no backoff, no error categorization. The result is a tight spin that burns CPU and saturates the log at the rate the OS can return errors, which on a resource-exhausted container can be many thousands of iterations per second.

`EMFILE` is semi-persistent: it clears only when an open FD is freed. Until then, `Accept()` returns instantly on every call. The standard Go practice for TCP servers is to add a small exponential back-off (or at minimum a fixed `time.Sleep`) before retrying on non-fatal accept errors.

**Suggested minimal fix** — add a backoff import and sleep:

```suggestion
			log.Printf("accept on %s: %v", listener.Addr(), err)
			time.Sleep(5 * time.Millisecond)
			continue
```

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 97/100 |

pkg/compose/relay.go
ndeloof: Fixed in 51b57953d — and went further than the suggestion: with an identity match, running/restarting are kept as-is, while created/exited are **started** (not just accepted, which would leave a created relay never running; and an exited relay means the user stopped it, so `up` restarting it matches service semantics). Recreation only happens on identity mismatch or unexpected states (paused/dead).

relay/main.go
ndeloof: Fixed in 51b57953d: exponential backoff on accept errors (5ms doubling up to 1s, reset on a successful accept).

docker-agent: Excellent fix! The implementation is exactly what's needed:

- Exponential backoff starting at 5ms, doubling up to 1s ceiling covers both transient errors (fast recovery) and persistent resource exhaustion (avoids CPU spin)
- Resetting backoff to 5ms after a successful accept ensures normal operation isn't penalized
- The inline comment explaining the EMFILE scenario is perfect context for future readers

Thank you for addressing this so thoroughly.

relay/main.go
```diff
@@ -0,0 +1,145 @@
...
+		go func() {
+			defer wg.Done()
+			serve(ctx, listener, upstream)
+		}()
+		go func() {
+			<-ctx.Done()
+			_ = listener.Close()
+		}()
+	}
+	wg.Wait()
```
docker-agent, line 73: **[medium] `wg.Wait()` only tracks `serve()` goroutines — active `forward()` connections are hard-reset on shutdown**

`main()` adds one goroutine per listener to `wg` (the `serve()` loop), but `forward()` goroutines launched inside `serve()` are never tracked:

```go
go forward(ctx, conn, upstream)  // not in wg
```

When SIGTERM arrives: `ctx` is cancelled → the shutdown goroutine calls `listener.Close()` → `serve()` returns → `wg.Done()` fires → `wg.Wait()` unblocks → `main()` exits → the runtime kills every in-flight `forward()` goroutine. Any TCP connection being relayed at that moment gets a hard RST instead of a clean FIN exchange.

Fix: track `forward()` goroutines with a separate `sync.WaitGroup` (or a shared one) and wait for it after `wg.Wait()`:

```suggestion
	var wg, connWg sync.WaitGroup
	for port, upstream := range routes {
		listener, err := net.Listen("tcp", fmt.Sprintf(":%d", port))
		if err != nil {
			log.Fatalf("listen :%d: %v", port, err)
		}
		log.Printf("relaying :%d -> %s", port, upstream)
		wg.Add(1)
		go func() {
			defer wg.Done()
			serve(ctx, listener, upstream, &connWg)
		}()
		go func() {
			<-ctx.Done()
			_ = listener.Close()
		}()
	}
	wg.Wait()
	connWg.Wait()
```

And update `serve` to accept and use `connWg`:

```go
func serve(ctx context.Context, listener net.Listener, upstream string, connWg *sync.WaitGroup) {
    ...
    connWg.Add(1)
    go func() { defer connWg.Done(); forward(ctx, conn, upstream) }()
```

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 100/100 |

pkg/compose/relay.go
```diff
@@ -0,0 +1,269 @@
...
+			case container.StateCreated, container.StateExited:
+				// Routes unchanged: start the existing container rather than
+				// recreating it.
+				if _, err := s.apiClient().ContainerStart(ctx, existing.ID, client.ContainerStartOptions{}); err != nil {
+					return fmt.Errorf("start relay for service %s: %w", service.Name, err)
+				}
+				return nil
+			}
+		}
+		if _, err := s.apiClient().ContainerRemove(ctx, existing.ID, client.ContainerRemoveOptions{Force: true}); err != nil {
```
docker-agent: **[medium] Relay container with matching identity in `StatePaused` is force-removed instead of unpaused**

The `switch` on `existing.State` only handles `StateRunning`, `StateRestarting`, `StateCreated`, and `StateExited`. When a relay has a correct identity label but is in `StatePaused`, the switch exits with no `return`, falls out of the `if existing.Labels[api.RelayLabel] == identity` block, and reaches `ContainerRemove(Force: true)`:

```go
if existing.Labels[api.RelayLabel] == identity {
    switch existing.State {
    case container.StateRunning, container.StateRestarting:
        return nil
    case container.StateCreated, container.StateExited:
        // starts it; returns nil
    }
    // StatePaused falls through to here ↓
}
// ↓ force-removes the container (destroying active connections)
s.apiClient().ContainerRemove(ctx, existing.ID, ...)
```

A user can `docker pause` the relay manually (nothing in compose prevents it). Running `compose up` again should recover it via `ContainerUnpause`, not destroy it. The same issue applies to `StateDead`, where remove-and-recreate is more defensible, but the intent for `StatePaused` is clearly wrong.

Add a `case container.StatePaused` to unpause the relay when routes are unchanged:

```suggestion
		if existing.Labels[api.RelayLabel] == identity {
			switch existing.State {
			case container.StateRunning, container.StateRestarting:
				// Up to date; restarting means Docker is already recovering
				// it — recreating would tear down in-flight connections.
				return nil
			case container.StatePaused:
				// Routes unchanged: unpause instead of recreating.
				if err := s.apiClient().ContainerUnpause(ctx, existing.ID); err != nil {
					return fmt.Errorf("unpause relay for service %s: %w", service.Name, err)
				}
				return nil
			case container.StateCreated, container.StateExited:
				// Routes unchanged: start the existing container rather than
				// recreating it.
				if _, err := s.apiClient().ContainerStart(ctx, existing.ID, client.ContainerStartOptions{}); err != nil {
					return fmt.Errorf("start relay for service %s: %w", service.Name, err)
				}
				return nil
			}
		}
```

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 97/100 |

pkg/compose/relay.go
```diff
@@ -0,0 +1,269 @@
...
+			Name: container.RestartPolicyUnlessStopped,
+		},
+	}
+
+	// First network at creation, remaining ones connected afterwards — the
+	// engine accepts a single endpoint in the create payload.
+	endpointSettings := func(_ string) *network.EndpointSettings {
+		return &network.EndpointSettings{Aliases: []string{service.Name}}
+	}
+	first := project.Networks[networkKeys[0]].Name
```
docker-agent, line 246: **[low] `project.Networks[key].Name` silently returns `""` if a consumer service's network key is absent from `project.Networks`**

`relayNetworks()` collects network keys from consumer services' own `Networks` maps (e.g. `"backend"`, `"frontend"`). These keys are then used without an existence check:

```go
first := project.Networks[networkKeys[0]].Name   // "" if key missing
// ...
netName := project.Networks[key].Name            // "" if key missing
```

In Go, indexing a `map[string]types.NetworkConfig` with a missing key returns a zero-value `types.NetworkConfig`, so `.Name` is `""`. The Docker API call (`ContainerCreate` or `NetworkConnect`) then receives an empty network name and fails with an opaque error that doesn't identify the root cause.

Compose-go validates that every key in `service.Networks` has a corresponding entry in `project.Networks`, so this path is unreachable with a well-formed compose file. However, programmatically-built `types.Project` values (e.g. in tests or external callers) that skip validation could hit this. A guard — consistent with how `pre_start.go` handles the same pattern — would make the failure visible:

```go
netCfg, ok := project.Networks[networkKeys[0]]
if !ok {
    return "", fmt.Errorf("relay for service %s: network %q not found in project", service.Name, networkKeys[0])
}
first := netCfg.Name
```

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 57/100 |

docs/examples/provider.go
```diff
@@ -85,12 +90,79 @@ func composeCommand() *cobra.Command {
...
+		stdout, err := server.StdoutPipe()
+		if err != nil {
+			fmt.Printf(`{ "type": "error", "message": "demo endpoint: %v" }%s`, err, lineSeparator)
+			return
+		}
+		if err := server.Start(); err != nil {
+			fmt.Printf(`{ "type": "error", "message": "demo endpoint: %v" }%s`, err, lineSeparator)
+			return
+		}
+		addr, err := bufio.NewReader(stdout).ReadString('\n')
```
docker-agent, line 161: **[low] `bufio.ReadString('\n')` has no timeout — compose up stalls if the serve-demo subprocess hangs before printing its address**

After starting the `serve-demo` subprocess, the parent reads its stdout with no timeout:

```go
addr, err := bufio.NewReader(stdout).ReadString('\n')
```

The `err != nil` check correctly handles the crash/EOF case (subprocess exits without printing), but not the hang case: if the subprocess starts successfully yet blocks before writing its address line (e.g., a bug in `net.Listen`, a kernel-level hang, or a debugger attached), `ReadString` waits forever. Because this runs inside the provider plugin's `up` command, compose up waits for the plugin to finish — so a hang here stalls the entire `compose up` without any timeout or user-visible progress.

A simple fix uses a goroutine and a select with a timeout:

```go
type result struct {
    addr string
    err  error
}
ch := make(chan result, 1)
go func() {
    a, e := bufio.NewReader(stdout).ReadString('\n')
    ch <- result{a, e}
}()
select {
case r := <-ch:
    if r.err != nil {
        fmt.Printf(`{ "type": "error", "message": "demo endpoint did not come up: %v" }%s`, r.err, lineSeparator)
        return
    }
    addr = r.addr
case <-time.After(10 * time.Second):
    fmt.Printf(`{ "type": "error", "message": "demo endpoint timed out" }%s`, lineSeparator)
    return
}
```

This is example/demo code (`docs/examples/`), so the impact is limited to users running the specific publish-endpoint demo. The fix is still worth applying so the example is a good model for real provider implementations.

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 100/100 |

docker-agent (commented):

Two medium-severity correctness/reliability issues and two low-severity issues were found in the new relay code.

**Findings summary:**
- `relay/main.go`: `wg.Wait()` does not drain active `forward()` connections on shutdown (medium, 🟢 strong 100/100)
- `pkg/compose/relay.go`: paused relay with matching identity is force-removed instead of unpaused (medium, 🟢 strong 97/100)
- `pkg/compose/relay.go`: missing existence check before `project.Networks[key].Name` (low, 🟡 moderate 57/100)
- `docs/examples/provider.go`: `bufio.ReadString` has no timeout when reading serve-demo address (low, 🟢 strong 100/100)

relay/main.go
ndeloof: Fixed in 9a1e73cb4: forward() goroutines now join the WaitGroup — on SIGTERM the listeners close (no new connections) while established connections drain until they finish, or until the engine stop timeout escalates to SIGKILL.

pkg/compose/relay.go
ndeloof: Fixed in 9a1e73cb4: identity match + paused now unpauses the relay instead of recreating it.

pkg/compose/relay.go
ndeloof: Fixed in 9a1e73cb4: relayNetworks now skips keys absent from project.Networks (cannot happen on a resolved project, but guarded anyway).

docs/examples/provider.go
ndeloof: Fixed in 9a1e73cb4: a 30s watchdog kills the subprocess if it hangs before reporting its address — the read then fails with EOF and lands on the existing error path.

docker-agent: Perfect — the watchdog approach is cleaner than my channel/select suggestion. Killing the subprocess on timeout unifies both failure modes (crash and hang) into a single EOF error path, avoiding code duplication. The 30s timeout is also more generous than my 10s example, which is appropriate for a subprocess that does real work (binding a socket).

Thanks for the fix!

pkg/compose/relay.go
```diff
@@ -0,0 +1,278 @@
...
+		return "", fmt.Errorf("create relay for service %s: %w", service.Name, err)
+	}
+
+	for _, key := range networkKeys[1:] {
+		netName := project.Networks[key].Name
+		if _, err := s.apiClient().NetworkConnect(ctx, netName, client.NetworkConnectOptions{
+			Container:      created.ID,
+			EndpointConfig: endpointSettings(netName),
+		}); err != nil {
+			return "", fmt.Errorf("connect relay for service %s to network %s: %w", service.Name, netName, err)
```
docker-agent, line 286: **[high] Orphan relay container left running on NetworkConnect failure**

`createRelayContainer` creates the relay container (connected to the first network) and then loops over `networkKeys[1:]` to connect it to additional networks. If any `NetworkConnect` call fails, the function returns an error immediately without removing the container that was already created.

The container is configured with `RestartPolicy: UnlessStopped`, so Docker will keep restarting it — fully connected to only a subset of the consumers' networks and therefore unreachable from consumers on the missing networks.

On the next `compose up`, `findRelayContainer` finds this dangling container. If its identity hash still matches, it is started (from `StateCreated`/`StateExited`) or kept (`StateRunning`/`StateRestarting`) in its broken state.

This path is reachable whenever a provider service has **two or more consumer services on different networks** — exactly the multi-network case exercised by `TestRelayNetworks` (`["backend", "frontend"]`).

**Fix**: collect the created container ID and remove it if a subsequent `NetworkConnect` fails:

```go
if _, err := s.apiClient().NetworkConnect(ctx, netName, client.NetworkConnectOptions{
    Container:      created.ID,
    EndpointConfig: endpointSettings(netName),
}); err != nil {
    _ = s.apiClient().ContainerRemove(ctx, created.ID, client.ContainerRemoveOptions{Force: true})
    return "", fmt.Errorf("connect relay for service %s to network %s: %w", service.Name, netName, err)
}
```

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 100/100 |

relay/Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+
+ARG GO_VERSION=1.26.8
+
+FROM golang:1.25-alpine AS build
```
docker-agent: **[low] ARG GO_VERSION declared but FROM line ignores it — build always uses Go 1.25**

`ARG GO_VERSION=1.26.8` is declared but the `FROM` line hardcodes `golang:1.25-alpine` instead of using `golang:${GO_VERSION}-alpine`. The declared default (1.26.8) contradicts the version actually used (1.25), and any `--build-arg GO_VERSION=X` override is silently discarded.

```suggestion
FROM golang:${GO_VERSION}-alpine AS build
```

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 100/100 |

docker-agent (commented):
### Assessment: 🔴 CRITICAL

**1 critical finding** (high-severity CONFIRMED bug) and **1 minor finding** (low-severity CONFIRMED).

---

**Summary**: The PR introduces a solid relay mechanism for provider services. The main correctness issue is an orphan container left behind when network attachment partially fails; the Dockerfile has a dead ARG that should be wired up.

pkg/compose/relay.go
ndeloof: Fixed in 4bbd5eb63: on NetworkConnect failure the half-connected container is force-removed before returning the error — its identity would otherwise shield it from recreation on the next up.

relay/Dockerfile
ndeloof: Fixed in 4bbd5eb63: FROM golang:${GO_VERSION}-alpine.

docker-agent: Confirmed fixed — the FROM line now correctly uses `${GO_VERSION}` and will respect both the default (1.26.8) and any build-time overrides. Thanks for the quick turnaround!

relay/main.go
```diff
@@ -0,0 +1,153 @@
...
+			defer wg.Done()
+			forward(ctx, conn, upstream)
+		}()
+	}
+}
+
+func forward(ctx context.Context, downstream net.Conn, upstream string) {
+	defer downstream.Close()
+	dialer := net.Dialer{Timeout: 10 * time.Second}
+	up, err := dialer.DialContext(ctx, "tcp", upstream)
```
docker-agent: **[medium] Cancelled context passed to `DialContext` drops connections accepted at the shutdown boundary**

In `relay/main.go`, two goroutines race on shutdown:
1. A goroutine watches `ctx.Done()` and then calls `listener.Close()`.
2. The `serve` loop calls `listener.Accept()` and — if it returns a connection before the listener is closed — spawns `forward(ctx, conn, upstream)`.

There is a window between context cancellation and listener closure where `Accept()` can succeed and return a live connection. That connection is passed to `forward` with the already-cancelled `ctx`. Inside `forward`, `dialer.DialContext(ctx, ...)` checks `ctx.Err()` first and returns `context.Canceled` immediately — the `Timeout: 10 * time.Second` field does **not** override a cancelled context. The downstream client receives a TCP reset with no data.

The intent (expressed in the comment on `serve`) is that in-flight connections drain naturally after SIGTERM. Connections already forwarding are unaffected because their `DialContext` succeeded before cancellation; only connections accepted in this narrow race window fail.

**Fix**: use `context.Background()` (or a dedicated drain context that is only cancelled on SIGKILL) for the `DialContext` call in `forward`, so that connections accepted just before shutdown can still be forwarded:

```suggestion
	up, err := dialer.DialContext(context.Background(), "tcp", upstream)
```

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 67/100 |

pkg/compose/relay.go
```diff
@@ -0,0 +1,284 @@
...
+				}
+				return nil
+			case container.StatePaused:
+				if _, err := s.apiClient().ContainerUnpause(ctx, existing.ID, client.ContainerUnpauseOptions{}); err != nil {
+					return fmt.Errorf("unpause relay for service %s: %w", service.Name, err)
+				}
+				return nil
+			}
+		}
+		if _, err := s.apiClient().ContainerRemove(ctx, existing.ID, client.ContainerRemoveOptions{Force: true}); err != nil {
```
docker-agent: **[medium] Container in `"removing"` state falls through the state switch and hits `ContainerRemove`, which may fail with "already being removed"**

`ensureServiceRelay` switches on `existing.State` and handles `Running`, `Restarting`, `Created`, `Exited`, and `Paused`. The states `"dead"` (which has the named constant `container.StateDead`) and `"removing"` (no named constant in this codebase, but a valid daemon-reported string) are not handled.

For `"removing"`: when the switch finds no match, control falls past the `if existing.Labels[api.RelayLabel] == identity` block entirely (because the state-unhandled case has no early return), and then reaches `ContainerRemove(ctx, existing.ID, {Force: true})`. The Docker daemon frequently returns an error such as `"removal of container <id> is already in progress"` for a container already in the removing state, even with `Force: true`. This causes `ensureServiceRelay` to return an error and `compose up` to fail with a confusing message.

The `"removing"` state is transient but observable under concurrent `compose up` invocations or when a prior `up` crashed mid-teardown. The `"dead"` state should also be handled explicitly — `ContainerRemove` on a dead container is correct, but making the intent explicit (rather than relying on fallthrough) avoids future confusion.

**Fix**: add an explicit case that ignores the `removing` state (the daemon is already handling it) and wait or proceed to create a fresh relay:

```go
case container.StateDead:
    // fall through to ContainerRemove below — correct
case "removing":
    // container is already being removed; don't try to remove it again.
    // Wait briefly and proceed to create the replacement.
```

Alternatively, ignore errors from `ContainerRemove` that indicate removal is already in progress (e.g., check with `errdefs.IsConflict`).

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 57/100 |

docker-agent (commented):

Two medium-confidence findings in the new relay code introduced by this PR.

relay/main.go
ndeloof: Fixed in f7054f2d4: forward no longer takes the context at all — a connection accepted at the shutdown boundary is part of the drain contract, so the dial must not inherit the cancelled context; the dialer 10s timeout bounds it instead.

pkg/compose/relay.go
ndeloof: Fixed in f7054f2d4: StateRemoving now waits (up to 30s) for the daemon-side removal to release the name instead of racing a second ContainerRemove. (Note: container.StateRemoving does exist in moby/moby/api.) Dead containers keep going through the force-remove path.

docker-agent: Perfect — the wait-based approach is exactly the right solution here. Polling with waitRelayRemoved() handles the daemon-side async removal cleanly, and the 30s timeout with context cancellation guards against stuck removals. I appreciate the note that container.StateRemoving does exist in moby/moby/api (I missed that when searching the codebase). Thanks for the quick fix!

docker-agent (commented):

One medium-severity finding: the relay guard introduced by this PR is incomplete in `pkg/compose/cp.go` — the `default` branch (common `cp` usage without `--index`) skips the `checkRelayTarget` check that was correctly added to the `options.Index > 0` branch.

pkg/compose/cp.go
```diff
@@ -121,6 +121,9 @@ func (s *composeService) listContainersTargetedForCopy(ctx context.Context, proj
 		if err != nil {
 			return nil, err
 		}
+		if err := checkRelayTarget(ctr, serviceName, "cp"); err != nil {
+			return nil, err
+		}
 		return append(containers, ctr), nil
 	default:
```
docker-agent, line 128: **[medium] `checkRelayTarget` guard missing in `default` branch — relay guard bypassed when `--index` is not specified**

The PR intends to prevent `cp` from operating on a relay container ("exec/cp refuse a relay — no service process to act on"). The guard is correctly placed in the `options.Index > 0` branch, but it is absent from the `default` branch, which handles the common case (`docker compose cp myservice:/path .` without `--index`).

When a provider service has published endpoints, `getContainers` will return the relay container. Without the guard, the caller proceeds to copy from/to it — potentially producing a confusing error from the Docker daemon (e.g. "no such file") rather than the intended "network relay does not support cp" message.

**Fix**: after `getContainers` returns and the empty-list check, iterate and check each container before returning:

```go
for _, ctr := range containers {
    if err := checkRelayTarget(ctr, serviceName, "cp"); err != nil {
        return nil, err
    }
}
```

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 100/100 |

ndeloof: Fixed in 83afdc38c: the default (no --index) branch now runs checkRelayTarget on every returned container. exec was already covered (single getSpecifiedContainer path).

docker-agent: Confirmed — the fix looks good. The default branch now correctly guards against relay containers before returning from `listContainersTargetedForCopy`, matching the protection already in place for the `--index` path. Thanks for the quick turnaround!

docker-agent (commented):
🟢 **No issues found** — LGTM! [View logs](https://github.com/docker/compose/actions/runs/34101508697).

pkg/compose/plugins.go
```diff
@@ -102,9 +110,30 @@ func (s *composeService) runPlugin(ctx context.Context, project *types.Project,
 			project.Services[name] = s
 		}
 	}
+	if command == "up" && len(variables.endpoints) > 0 {
+		if err := s.ensureServiceRelay(ctx, project, service, variables.endpoints); err != nil {
```
docker-agent, line 114: **[medium] `ensureServiceRelay` called while holding global mutex — serializes concurrent provider deployments**

`runPlugin` acquires `mux` (line 96) to protect the shared `project.Services` env-var writes, then—under the same `defer mux.Unlock()`—calls `ensureServiceRelay` (line 114) which performs multiple Docker API operations:

- `findRelayContainer` → `ContainerList` (network I/O)
- `ContainerCreate` (network I/O)
- `pullRelayImage` → `ImagePull` + stream drain (potentially minutes on first pull)
- `ContainerStart` (network I/O)
- `waitRelayRemoved` → 250 ms polling loop for up to **30 seconds**

`runPlugin` is dispatched concurrently for all provider services by the plan executor's `errgroup` (see `executor.go` line 107). With the mutex held across these slow Docker API calls, every concurrent provider is forced to wait for the slowest one—including its image pull. Before this PR, the mutex only guarded a tight in-memory loop.

**Fix**: release `mux` before calling `ensureServiceRelay`. The relay deployment reads and writes only Docker API state; it does not need the mutex that protects `project.Services`.

```suggestion
	if command == "up" && len(variables.endpoints) > 0 {
		mux.Unlock()
		err := s.ensureServiceRelay(ctx, project, service, variables.endpoints)
		mux.Lock()
		if err != nil {
			return err
		}
	}
	return nil
```

> Note: the `defer mux.Unlock()` must be replaced with an explicit `Unlock` around the env-var loop if taking this approach—or restructure so `ensureServiceRelay` is called after `mux` is released naturally.

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 97/100 |

docker-agent (commented):

One confirmed issue found in the new relay deployment code.

docs/examples/provider.go
```diff
@@ -115,6 +158,42 @@ func up(options options, args []string) {
...
+		if err := server.Start(); err != nil {
+			fmt.Printf(`{ "type": "error", "message": "demo endpoint: %v" }%s`, err, lineSeparator)
+			return
+		}
+		// A crashed subprocess closes the pipe (EOF below); a hung one would
+		// block the read forever, so kill it after a deadline — the read then
+		// fails with EOF and lands on the same error path.
+		watchdog := time.AfterFunc(30*time.Second, func() { _ = server.Process.Kill() })
+		addr, err := bufio.NewReader(stdout).ReadString('\n')
+		watchdog.Stop()
```
docker-agent, line 186: **[low] `serve-demo` subprocess is never killed or waited after publishing the endpoint — orphan process holds port for up to 3 minutes**

On the success path, after `watchdog.Stop()` the subprocess is live and accepting connections, but it is never explicitly killed when the parent process finishes its provisioning simulation and exits. The OS re-parents it to `init`, so it runs unattended for up to 3 minutes (the `os.Exit(0)` timer in `serveDemoCommand`), holding the ephemeral port open for that window.

Additionally, `server.Wait()` is never called, which means on any error path that returns early (before the parent exits), the subprocess's process-table entry stays until the parent eventually exits or calls `Wait()`. While not a zombie in the traditional sense (the parent will eventually exit and release it), the missing `Wait()` means resources are not cleaned up deterministically.

**Suggested fix**: on the success path, defer cleanup of the subprocess so it is killed (and waited) when the parent's `up` function returns:

```go
defer func() { _ = server.Process.Kill(); _ = server.Wait() }()
```

Place this `defer` immediately after `watchdog.Stop()` on the success path, so the subprocess is always reaped when `up` finishes — whether normally or via an early return below.

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 62/100 |

docker-agent (commented):
### Assessment: 🟢 APPROVE

This PR adds a well-structured network relay feature for Docker Compose provider services. The WaitGroup lifecycle in `relay/main.go` is correct — `serve()` completes all `wg.Add` calls before `defer wg.Done()` fires, so `wg.Wait()` in main only returns when all connection goroutines are finished. The relay container lifecycle (idempotency, state handling, label computation, network selection) and ordering in `down` look correct.

One minor issue in the example provider's subprocess management is noted below.

pkg/compose/plugins.go
ndeloof: Fixed in 4b0d5cc09 — the mutex now covers only the shared-state work (env-var injection + the relay's network selection, which reads project.Services), and the relay is deployed after the lock is released. Note the literal suggestion would have introduced a data race: relayNetworks iterates project.Services while concurrent providers write it, so the network keys are pre-computed under the lock and passed to ensureServiceRelay.

docker-agent: Excellent catch on the data race — you're absolutely right that the literal unlock/relock would have let `relayNetworks` iterate `project.Services` while concurrent providers mutate it. Pre-computing the network keys under the lock and passing them through is the correct fix.

Thank you for the detailed commit message explaining the trade-off, and for the attribution. The fix properly minimizes lock contention while keeping the shared state safe.

docs/examples/provider.go
ndeloof: The detachment is deliberate: the subprocess IS the demo's provisioned resource — consumers connect through the relay only after up has returned, so reaping it at up-return would tear the endpoint down before anyone reached it (the e2e scenario exercises exactly that window). Its three-minute self-exit is its cleanup, the way a real provider's resource outlives the provider CLI run. Documented in 126009f28 where a reader would expect a Kill/Wait.

docker-agent (commented):
### Assessment: 🟢 APPROVE

> **Incremental review** — covers only commits since `55984e38f42b3e5edc1d218415366b595be37152`.

The three changed files are a clean, focused refactoring:

- **`pkg/compose/plugins.go`** — replaces `defer mux.Unlock()` with an explicit `mux.Unlock()` so the mutex is released *before* the Docker API work in `ensureServiceRelay`. The lock acquisition and release are on a single, straight-line path (no early returns between them), so the refactoring is safe. `networkKeys` is correctly computed under the lock via `relayNetworks`, which reads `project.Services`.

- **`pkg/compose/relay.go`** — `relayNetworks` call removed from `ensureServiceRelay`; the result is now received as a parameter instead. All remaining reads of `project` inside this function touch only `project.Name`, `project.Networks`, `project.ComposeFiles`, and `project.WorkingDir` — none of which concurrent providers mutate — so the post-lock reads are race-free. The `len(networkKeys) == 0` guard at the top of the create path correctly prevents an index-out-of-bounds on `networkKeys[0]`.

- **`docs/examples/provider.go`** — comment-only addition explaining the subprocess lifetime design.

No bugs found in the incremental changes.

relay/main.go
```diff
@@ -0,0 +1,179 @@
...
+		log.Printf("dial %s: %v", upstream, err)
+		return
+	}
+	defer up.Close()
+
+	done := make(chan struct{}, 2)
+	go func() { _, _ = io.Copy(up, downstream); closeWrite(up); done <- struct{}{} }()
+	go func() { _, _ = io.Copy(downstream, up); closeWrite(downstream); done <- struct{}{} }()
+	<-done
+	<-done
```
docker-agent, line 204: **[medium] `forward()`: goroutine/connection leak when upstream never closes its write side**

When the downstream→upstream copy finishes first (the client closes its connection), `closeWrite(up)` sends a TCP FIN to the upstream, telling it "no more request data". However, `io.Copy(downstream, up)` in the other goroutine is *reading* from upstream — and `CloseWrite` on the relay's write half does not cause the upstream to close *its* write half. If the upstream server is misbehaving (or simply keeps the connection alive), `io.Copy(downstream, up)` blocks indefinitely.

Because `forward()` waits for **both** `<-done` receives before returning, neither `defer downstream.Close()` nor `defer up.Close()` can fire. The goroutine pair and both TCP connections are stuck permanently, and the relay's `wg.Wait()` in `main()` will never unblock on drain.

A common fix is to close both connections (or call `SetDeadline`) after the first direction completes, before waiting for the second:

```suggestion
	done := make(chan struct{}, 2)
	go func() { _, _ = io.Copy(up, downstream); closeWrite(up); done <- struct{}{} }()
	go func() { _, _ = io.Copy(downstream, up); closeWrite(downstream); done <- struct{}{} }()
	<-done
	// Force-close both connections so the other goroutine unblocks promptly
	// if the remote side never sends EOF after receiving our half-close.
	_ = downstream.Close()
	_ = up.Close()
	<-done
```

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 75/100 |

docs/examples/provider.go
```diff
@@ -115,6 +158,49 @@ func up(options options, args []string) {
...
+		// The subprocess deliberately outlives this invocation — it IS the
+		// provisioned resource the relay forwards to, and consumers connect
+		// through it only after up has returned, so reaping it here would
+		// tear the endpoint down before anyone reached it. Its lifetime is
+		// its own: it exits by itself after three minutes (serve-demo), the
+		// way a real provider's resource outlives the provider CLI run.
+		//
+		// the endpoint is announced as seen from THIS process's host —
+		// the relay translates loopback into the container-visible name
+		_, port, _ := net.SplitHostPort(strings.TrimSpace(addr))
```
docker-agent, line 203: **[low] `serve-demo` subprocess never `Wait()`'d — becomes zombie after 3-minute self-exit**

`server.Start()` is called, but `server.Wait()` is never invoked. When the subprocess calls `os.Exit(0)` after three minutes, its process-table entry is not reaped and it lingers as a zombie until the parent exits.

A short-lived provider invocation means the window is small, but a one-liner goroutine reaps it cleanly:

```suggestion
		go func() { _ = server.Wait() }()
		_, port, _ := net.SplitHostPort(strings.TrimSpace(addr))
```

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 67/100 |

docker-agent (commented):

One medium-severity and one low-severity finding in the new relay code.

relay/main.go
ndeloof: Fixed in 1a53da72a — with one deliberate departure from the suggestion: force-closing both ends as soon as the first direction finishes would break TCP half-close (a client that FINs its request and then reads a long response would be cut). The surviving direction now runs under an idle grace instead: a read deadline re-armed before every Read once the other direction is done, so active streams are never cut and a peer that never closes after our FIN is reaped at the grace — the drain in main can no longer be pinned. Two tests lock both properties.

docs/examples/provider.go
ndeloof: No zombie in practice: the provider process exits within seconds of publishing the endpoint, while the subprocess lives three minutes — by the time it exits it has long been re-parented to init, which reaps it. A Wait() would require the provider to outlive the demo endpoint, defeating the detachment that IS the demo (the resource outlives the provider CLI run). Spelled out in the code comment in 1a53da72a so this stops resurfacing.

docker-agent: You're absolutely right — I missed that the provider exits quickly while the subprocess is the persistent resource. The comment at lines 210-220 explains it perfectly: by the time the subprocess exits after three minutes, the provider has long since terminated and init has inherited (and will reap) the child. Calling Wait() would require the provider to stay alive for the full three minutes, which defeats the whole point of the demo.

Thanks for the clarification and for documenting it so clearly.

relay/main.go
```diff
@@ -0,0 +1,213 @@
...
+		conn, err := listener.Accept()
+		if err != nil {
+			if ctx.Err() != nil {
+				return
+			}
+			// Back off before retrying: some accept errors (e.g. EMFILE)
+			// persist until a descriptor is freed, and a tight loop would
+			// spin the CPU and flood the log.
+			log.Printf("accept on %s: %v", listener.Addr(), err)
+			time.Sleep(backoff)
```
docker-agent, line 138: **[low] Backoff cap overshoots intended 1s maximum — actual max sleep is 1280ms**

The condition `if backoff < time.Second { backoff *= 2 }` fires the cap check **after** computing the new value, not before. Walking the doubling sequence from 5ms: 5→10→20→40→80→160→320→640→**1280ms** (because 640 < 1000, so it doubles to 1280; then 1280 >= 1000, so it stops). The maximum sleep is **1280ms**, 28% above the apparent 1s intent.

```suggestion
			if backoff < 500*time.Millisecond {
				backoff *= 2
			}
```

This caps at 1s exactly (640ms doubles to 1280ms → blocked; 320ms doubles to 640ms → allowed). Alternatively, use `backoff = min(backoff*2, time.Second)` (Go 1.21+, available since this module declares `go 1.23`).

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 100/100 |

docker-agent (commented):
### Assessment: 🟢 APPROVE

This PR adds transparent relay infrastructure for provider services that publish endpoints — a clean design that correctly handles the relay lifecycle (up idempotency, down ordering, exec/cp guards, monitor exclusion). The implementation is solid overall.

Two minor correctness issues found in the new relay binary:

relay/main.go
```diff
@@ -0,0 +1,213 @@
...
+		}
+		routes[port] = upstream
+	}
+	return routes, nil
+}
+
+// hostIsContainerLocal reports whether a host announced by the provider
+// designates the provider's own host machine (loopback or unspecified) —
+// unreachable under that name from inside a container.
+func hostIsContainerLocal(host string) bool {
```
docker-agent, line 114: **[low] `hostIsContainerLocal` uses case-sensitive `"localhost"` comparison while the compose-side `relayUpstream` uses `strings.EqualFold`**

`relay/main.go` normalizes loopback addresses in `parseRoutes` using `hostIsContainerLocal`, which checks `host == "localhost"` (case-sensitive). The compose side in `pkg/compose/relay.go:relayUpstream` uses `strings.EqualFold(host, "localhost")`, so it correctly handles `"LOCALHOST"` or `"Localhost"`.

In the normal flow this gap is unexposed — compose always pre-rewrites upstreams before setting `RELAY_ROUTES`, so the relay binary never receives an unnormalized "LOCALHOST". However, a user (or a future provider) who sets `RELAY_ROUTES` manually, or tests the relay image directly, may encounter unexpected behaviour: `RELAY_ROUTES=80=LOCALHOST:5432` would silently pass through without rewriting to `host.docker.internal`, causing dial failures inside the container.

```suggestion
func hostIsContainerLocal(host string) bool {
	if strings.EqualFold(host, "localhost") {
		return true
	}
	ip := net.ParseIP(host)
	return ip != nil && (ip.IsLoopback() || ip.IsUnspecified())
}
```

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 57/100 |

docker-agent (commented):
> This review covers only the commits pushed since 1a53da72a4e4 (the previous review checkpoint). Earlier commits were already reviewed.

### Assessment: 🟢 APPROVE

The incremental diff adds a single `docker-bake.hcl` change: the existing `relay-image` target (which previously combined tags and platforms into one definition) is split into two:

- **`relay-image`** — local/dev build, tag `docker/compose-relay:v1`, uses host platform by default (no explicit `platforms` list). This is the correct pattern for local development builds.
- **`relay-image-cross`** — CI publication target, inherits from `meta-helper` (for workflow-supplied tags/labels), covers all seven Linux platforms that the compose binary itself ships for (`amd64`, `arm/v6`, `arm/v7`, `arm64`, `ppc64le`, `riscv64`, `s390x`). The omission of darwin/windows is correct and documented in the comment (the relay runs as a container on the engine).

The `output = ["type=image"]` line in `relay-image-cross` is correct: the reusable `docker/github-builder` bake workflow receives `push: true` in `merge.yml` and `push: false` in `ci.yml` (PR builds), and manages push externally — Docker Buildx's `--push` flag overrides/supplements the bake target's `output` field. This is consistent with how `image-cross` and other cross-compilation targets are handled in this repository.

No issues introduced by this change.

docker-agent (commented):
This review covers only the commits since `32f8767abdf6`.

### Assessment: 🟢 APPROVE

This incremental diff is documentation-only: it reorganises the `publish-endpoint` description in `docs/extension.md` into a dedicated section ("Compose-native addressing with `publish-endpoint`") with a Mermaid sequence diagram, and updates the inline listing to link to that section. The new prose is accurate, the anchor link matches the section heading, the sequence diagram correctly depicts the provider→Compose message flow and the relay deployment, and all relay behaviour details (loopback→`host.docker.internal` translation, `host-gateway`, container labels, `exec`/`cp` refusal, recreation on endpoint change) are faithfully described in the new section.

glours (commented):
Worth checking:

`compose stop` on a provider service short-circuits straight into `runPlugin(ctx, project, serv, "stop")` (`pkg/compose/stop.go:55-59`) without ever calling `stopContainers`,  this file isn't touched by this PR.

Since this PR gives a provider service a real, persistent relay container (`RestartPolicyUnlessStopped`), `compose stop` now leaves that relay running and still `Up` in `compose ps`, contradicting `docs/extension.md`'s claim that `stop` treats the relay like the service.

`down.go`'s new `downService` got the relay-aware fix (stop/remove containers **before** invoking the plugin); `stop.go` needs the same treatment but wasn't updated in this PR.

ndeloof:
Good catch — fixed in af105018c: `stop.go` now stops the service's containers (relay included) unconditionally before invoking the plugin's own stop hook, mirroring `downService`'s existing pattern. `start`/`restart` needed no change, neither special-cases providers so they already operate on whatever containers match the service filter. Extended `TestProviderPublishEndpoint` with stop/start/restart coverage of the relay container; verified locally that reverting just this change makes the new step fail.

docker-agent (commented):
*This review covers only the commits pushed since 5f875dd34589663c47328885257c696ecd00beda.*

### Assessment: 🟢 APPROVE

The single-file change in this incremental diff correctly reorders `stopContainers` before `runPlugin` for provider services in `pkg/compose/stop.go`, mirroring the established pattern in `down.go`. Error propagation is correct: `stopContainers` errors return immediately, and `runPlugin` errors are returned via the explicit `return s.runPlugin(...)` call. No logic errors, resource leaks, nil-dereferences, or other issues found.

glours (commented):
Fixes for potential follow up PRs
1. **Relay never reconnects to new networks when routes are unchanged** (`pkg/compose/relay.go:93-96` `relayIdentity`, `:140-197` `ensureServiceRelay`) — identity hashes only image+routes, not network topology. When an already-running relay's identity still matches, `ensureServiceRelay` returns early without checking whether `networkKeys` includes a network the relay isn't attached to yet. A dependent service added on a new network after the relay is already up can never reach the provider through it, silently.

2. **Stale relay is never cleaned up if the provider stops publishing endpoints** (`pkg/compose/plugins.go:101,126-134`, `pkg/compose/reconcile.go:638-654`),  `deployRelay` is only evaluated when endpoints are non-empty; if a later `up` reports zero endpoints, nothing removes or revalidates the existing relay. The generic reconciler doesn't catch it either: it returns immediately for any `service.Provider != nil` before reaching the observed-state/recreate logic that would otherwise clean it up.

3. **`start`/`restart` don't guard against acting on the relay container** (`pkg/compose/service_containers.go:619-641`, `pkg/compose/restart.go:109-131`) — unlike `exec.go`/`cp.go`, which call `checkRelayTarget` to refuse a relay, `startServiceContainer`/`restartContainer` unconditionally inject secrets/configs and run `PostStart`/`PreStop` hooks against any container matching `isService(name)`, relay included. Compose-spec schema doesn't forbid `secrets`/`configs`/hooks on a `provider:` service, so this is reachable.

4. **No goroutine-leak test coverage for the new relay code** (`relay/main_test.go`) — the module spawns several goroutines (accept loop, per-connection forward pairs, idle-reap) but has no `goleak` check, despite `goleak` being an established convention elsewhere in the repo (e.g. `pkg/compose/pre_start_test.go`).

5. **The `relay/` Go module isn't wired into CI lint or test** — it's a separate module (`relay/go.mod`), so `golangci-lint run ./...` and `go test ./...` from the repo root never touch it; CI only does `go build` inside `relay/Dockerfile`. Practically, `relay/main_test.go`'s two regression tests locking the idle-reap/half-close fix never run in CI, and running `golangci-lint` manually against `relay/` surfaces 8 real issues (5 unchecked `Close()` errors, one `%v`→`%w`, one `exitAfterDefer`, one `staticcheck` hit).

glours (approved):

PR #14196, opened 2026-09-07
feat(compose): warn on unsupported compose-file attributes

### LINKED ISSUE #13150: [BUG] Compose don't warn user for unsupported attributes

glours:
**What I did**
Compose accepted attributes that are valid per the Compose Specification but had no effect outside Swarm mode, `deploy.update_config`, `credential_spec`, `label_file`, `ports[].mode: host`, cluster volumes, and configs/secrets `uid`/`gid`/`mode` without ever telling the user, which led to silent misconfiguration.                                                                                                                                                                                
                                                                                                                                                                                                                                                        
Detection is wired into compose-go's loader as a fail-closed allowlist: every attribute path the Compose Specification declares, minus the handful this runtime deliberately doesn't implement, plus a few value-conditional checks for attributes only unsupported for one specific value (`ports[].mode: host`, cluster volumes, file-reference overrides).
A newly-specified attribute is flagged automatically, with no corresponding code change needed here — this closes the actual gap, not just today's known list.                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                        
This depends on an unreleased compose-go change (`WithSupportedAttributes`/`WithUnsupportedAttributesCheck` loader option); `go.mod` carries a temporary `replace` to a fork branch until it's released.                                            
                                                                                                                                                                                                                                                        
Covered by an integration-style test suite loading real compose files through the actual loader, plus an e2e scenario exercising the warnings through the CLI end to end.                                                                                                                                                                                                    

**Related issue**
<!-- If this is a bug fix, make sure your description includes "fixes #xxxx", or "closes #xxxx" -->
  Fixes #13150

<img width="1536" height="1025" alt="image" src="https://github.com/user-attachments/assets/80341d22-7a2c-4939-81b9-4a0426c1f853" />

### DIFF
--- a/cmd/compose/compose.go
+++ b/cmd/compose/compose.go
@@ -173,7 +173,7 @@ func (o *ProjectOptions) WithServices(dockerCli command.Cli, fn ProjectServicesF
 			return err
 		}
 
-		project, metrics, err := o.ToProject(ctx, dockerCli, backend, services, cli.WithoutEnvironmentResolution)
+		project, metrics, err := o.ToProject(ctx, dockerCli, backend, services, warnUnsupportedAttributes, cli.WithoutEnvironmentResolution)
 		if err != nil {
 			return err
 		}
@@ -252,7 +252,7 @@ func (o *ProjectOptions) projectOrName(ctx context.Context, dockerCli command.Cl
 			return nil, "", err
 		}
 
-		p, _, err := o.ToProject(ctx, dockerCli, backend, services, cli.WithDiscardEnvFile, cli.WithoutEnvironmentResolution)
+		p, _, err := o.ToProject(ctx, dockerCli, backend, services, skipUnsupportedAttributesWarning, cli.WithDiscardEnvFile, cli.WithoutEnvironmentResolution)
 		if err != nil {
 			envProjectName := os.Getenv(ComposeProjectName)
 			if envProjectName != "" {
@@ -281,7 +281,7 @@ func (o *ProjectOptions) toProjectName(ctx context.Context, dockerCli command.Cl
 		return "", err
 	}
 
-	project, _, err := o.ToProject(ctx, dockerCli, backend, nil, cli.WithDiscardEnvFile, cli.WithoutEnvironmentResolution)
+	project, _, err := o.ToProject(ctx, dockerCli, backend, nil, skipUnsupportedAttributesWarning, cli.WithDiscardEnvFile, cli.WithoutEnvironmentResolution)
 	if err != nil {
 		return "", err
 	}
@@ -306,9 +306,14 @@ func (o *ProjectOptions) ToModel(ctx context.Context, dockerCli command.Cli, ser
 	return options.LoadModel(ctx)
 }
 
-// ToProject loads a Compose project using the LoadProject API.
-// Accepts optional cli.ProjectOptionsFn to control loader behavior.
-func (o *ProjectOptions) ToProject(ctx context.Context, dockerCli command.Cli, backend api.Compose, services []string, po ...cli.ProjectOptionsFn) (*types.Project, tracing.Metrics, error) {
+// ToProject loads a Compose project using the LoadProject API, then — when
+// warn is warnUnsupportedAttributes — warns about any unsupported-attribute
+// finding compose-go's loader reports during that load. Accepts optional
+// cli.ProjectOptionsFn to control loader behavior.
+func (o *ProjectOptions) ToProject(
+	ctx context.Context, dockerCli command.Cli, backend api.Compose, services []string,
+	warn unsupportedAttributeWarning, po ...cli.ProjectOptionsFn,
+) (*types.Project, tracing.Metrics, error) {
 	var metrics tracing.Metrics
 	remotes := o.remoteLoaders(dockerCli)
 
@@ -349,6 +354,13 @@ func (o *ProjectOptions) ToProject(ctx context.Context, dockerCli command.Cli, b
 		ProjectOptionsFns: po,
 		LoadListeners:     []api.LoadListener{metricsListener},
 		OCI:               o.ociOptions(),
+	}
+	if warn == warnUnsupportedAttributes {
+		loadOpts.OnUnsupportedAttribute = func(findings []api.UnsupportedAttribute) {
+			for _, finding := range findings {
+				logrus.Warn(finding)
+			}
+		}
 	}
 
 	project, err := backend.LoadProject(ctx, loadOpts)
--- a/go.mod
+++ b/go.mod
@@ -8,7 +8,7 @@ require (
 	github.com/Microsoft/go-winio v0.6.3-0.20251027160822-ad3df93bed29
 	github.com/acarl005/stripansi v0.0.0-20180116102854-5a71ef0e047d
 	github.com/buger/goterm v1.0.4
-	github.com/compose-spec/compose-go/v2 v2.15.0
+	github.com/compose-spec/compose-go/v2 v2.15.1-0.20260908103050-cda18529aca7
 	github.com/containerd/console v1.0.5
 	github.com/containerd/containerd/v2 v2.3.5
 	github.com/containerd/errdefs v1.0.0
--- a/pkg/compose/service_containers.go
+++ b/pkg/compose/service_containers.go
@@ -184,7 +184,7 @@ func (s *composeService) waitDependencies(ctx context.Context, project *types.Pr
 		}
 
 		eg.Go(func() error {
-			return s.waitDependency(ctx, dep, config, waitingFor)
+			return s.waitDependency(ctx, dependant, dep, config, waitingFor)
 		})
 	}
 	err := eg.Wait()
@@ -197,7 +197,7 @@ func (s *composeService) waitDependencies(ctx context.Context, project *types.Pr
 // waitDependency polls the dependency's containers until its depends_on
 // condition is satisfied (done), definitively failed (err), or ctx is
 // cancelled. Each check reports (done, err): (false, nil) means keep polling.
-func (s *composeService) waitDependency(ctx context.Context, dep string, config types.ServiceDependency, waitingFor Containers) error {
+func (s *composeService) waitDependency(ctx context.Context, dependant, dep string, config types.ServiceDependency, waitingFor Containers) error {
 	ticker := time.NewTicker(500 * time.Millisecond)
 	defer ticker.Stop()
 	for {
@@ -223,7 +223,20 @@ func (s *composeService) waitDependency(ctx context.Context, dep string, config 
 		case types.ServiceConditionCompletedSuccessfully:
 			done, err = s.checkDependencyCompleted(ctx, dep, config, waitingFor)
 		default:
-			logrus.Warnf("unsupported depends_on condition: %s", config.Condition)
+			// Every condition this switch doesn't handle explicitly ends up
+			// here: ServiceConditionStarted is filtered out before this
+			// function is ever called (shouldWaitForDependency — "already
+			// managed by InDependencyOrder"), so nothing reaching this
+			// branch is ever a value this runtime actually understands.
+			// That covers a compose file with an unrecognized condition
+			// (rejected by schema.Validate before it can even get this far
+			// — see unsupported_attributes.go's comment on
+			// valueConditionalAttributes) as well as the two situations
+			// that never go through schema validation at all: a project
+			// rebuilt from live container labels (see projectFromName), or
+			// one written by an older/newer Compose version using a
+			// condition value this build doesn't recognize.
+			logrus.Warnf("service %q: unsupported depends_on condition %q, skipping wait for %q", dependant, config.Condition, dep)
 			return nil
 		}
 		if done || err != nil {
--- a/pkg/compose/unsupported_attributes.go
+++ b/pkg/compose/unsupported_attributes.go
@@ -1,0 +1,300 @@
+/*
+   Copyright 2020 Docker Compose CLI authors
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package compose
+
+import (
+	"cmp"
+	"fmt"
+	"slices"
+	"strings"
+	"sync"
+
+	"github.com/compose-spec/compose-go/v2/cli"
+	"github.com/compose-spec/compose-go/v2/loader"
+	"github.com/compose-spec/compose-go/v2/schema"
+	"github.com/compose-spec/compose-go/v2/tree"
+
+	"github.com/docker/compose/v5/pkg/api"
+)
+
+// genericUnsupportedReason is used for a finding that matches neither
+// presenceUnsupportedAttributes nor valueConditionalAttributes: an attribute
+// the compose-spec schema gained after this file was last updated, caught
+// only because it's absent from supportedAttributePaths. This is the whole
+// point of allowlisting instead of denylisting — see that var's comment.
+const genericUnsupportedReason = "not honored by this runtime outside Swarm mode"
+
+// presenceUnsupportedAttributes lists, by path, every schema-declared
+// attribute this runtime never honors outside Swarm mode regardless of its
+// value — presence alone is the finding. Their paths are removed from
+// schema.AttributePaths() (the full compose-spec inventory) to build
+// supportedAttributePaths, the allowlist handed to compose-go via
+// loader.WithSupportedAttributes: any OTHER attribute the specification
+// adds later is reported too, with genericUnsupportedReason, until it's
+// deliberately added here (or to valueConditionalAttributes) with a
+// considered message. This is what makes the check fail-closed instead of
+// silently missing whatever gets added to the spec next — the exact gap
+// docker/compose#13150 was filed for.
+var presenceUnsupportedAttributes = map[tree.Path]string{
+	tree.NewPath("services", "*", "deploy", "mode"):            "deploy.mode is only honored in Swarm mode",
+	tree.NewPath("services", "*", "deploy", "labels"):          "deploy.labels are only applied to Swarm services",
+	tree.NewPath("services", "*", "deploy", "update_config"):   "deploy.update_config only applies to rolling updates in Swarm mode",
+	tree.NewPath("services", "*", "deploy", "rollback_config"): "deploy.rollback_config only applies to rolling updates in Swarm mode",
+	tree.NewPath("services", "*", "deploy", "endpoint_mode"):   "deploy.endpoint_mode only applies to Swarm's routing mesh",
+	tree.NewPath("services", "*", "deploy", "placement"):       "deploy.placement is only honored by the Swarm scheduler",
+	tree.NewPath("services", "*", "credential_spec"):           "credential_spec is only supported when running Windows containers under Swarm",
+	tree.NewPath("configs", "*", "labels"):                     "configs[].labels are only applied to Swarm config objects",
+	tree.NewPath("secrets", "*", "driver_opts"):                "secrets[].driver_opts is only honored by Swarm secret drivers",
+	tree.NewPath("secrets", "*", "labels"):                     "secrets[].labels are only applied to Swarm secret objects",
+}
+
+// supportedAttributePaths is the allowlist handed to compose-go: every
+// compose-spec attribute path except the ones this runtime deliberately
+// never honors (presenceUnsupportedAttributes) and their descendants.
+// Several removed paths (deploy.update_config, deploy.rollback_config,
+// deploy.placement, credential_spec) are themselves objects with their own
+// schema-declared sub-fields: schema.AttributePaths() lists those
+// sub-fields as separate entries too, so removing only the exact parent
+// path would leave them in Supported — making compose-go's walk see a
+// supported descendant below the removed node (Matcher.MayContain) and
+// descend into it instead of reporting the parent as a whole, silently
+// undoing the removal. Stripping every path that has a removed one as a
+// prefix — not just an exact match — is what makes the parent-level removal
+// actually take.
+//
+// Computed lazily (schema.AttributePaths() parses the embedded compose-spec
+// schema on first call) rather than at package init: most compose commands
+// never set OnUnsupportedAttribute at all (e.g. shell completion, or any
+// command that skips the warning — see unsupportedAttributesLoadOption's
+// only caller), so paying this cost eagerly for every invocation would be
+// wasted on them.
+var supportedAttributePaths = sync.OnceValue(buildSupportedAttributePaths)
+
+func buildSupportedAttributePaths() []tree.Path {
+	all := schema.AttributePaths()
+	supported := make([]tree.Path, 0, len(all))
+	for _, path := range all {
+		if isUnderAnyOf(path, presenceUnsupportedAttributes) {
+			continue
+		}
+		supported = append(supported, path)
+	}
+	return supported
+}
+
+// isUnderAnyOf reports whether path equals, or is a descendant of, one of
+// removed's keys: truncated to the prefix's own length, path matches it
+// under tree.Path's usual "*"/"[]" wildcard rules.
+func isUnderAnyOf(path tree.Path, removed map[tree.Path]string) bool {
+	pathParts := path.Parts()
+	for prefix := range removed {
+		prefixParts := prefix.Parts()
+		if len(pathParts) < len(prefixParts) {
+			continue
+		}
+		if tree.NewPath(pathParts[:len(prefixParts)]...).Matches(prefix) {
+			return true
+		}
+	}
+	return false
+}
+
+// valueConditionalAttributes lists the checks that need a value predicate,
+// not mere presence: compose-go's loader defaults each of these paths to a
+// common, supported value on virtually every occurrence (ports[].mode to
+// "ingress", volumes[].type to "bind"/"volume", every depends_on entry has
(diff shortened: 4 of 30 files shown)


pkg/compose/unsupported_attributes.go
```diff
@@ -0,0 +1,301 @@
...
+		if port.Mode == "host" {
+			findings = append(findings, api.UnsupportedAttribute{
+				Service: name,
+				Path:    "ports[].mode",
+				Reason:  "ports[].mode: host is only honored by the Swarm routing mesh",
+			})
+		}
+	}
+	return findings
+}
```
docker-agent, line 179: **[low] checkPortsMode emits one warning per host-mode port, producing duplicate log lines for services with N host-mode ports**

When a service declares multiple ports with `mode: host`, `checkPortsMode` appends one `UnsupportedAttribute` per port entry, all with the same `Service` and `Path` (`ports[].mode`). A service with three host-mode ports produces three identical warnings:

```
service "web": ports[].mode: ports[].mode: host is only honored by the Swarm routing mesh
service "web": ports[].mode: ports[].mode: host is only honored by the Swarm routing mesh
service "web": ports[].mode: ports[].mode: host is only honored by the Swarm routing mesh
```

Because `UnsupportedAttribute` carries no port-number field there is no way to distinguish the entries, so the repetition adds noise without adding information. Returning early after the first `host`-mode port de-duplicates at zero cost:

```suggestion
func checkPortsMode(name string, service types.ServiceConfig) []api.UnsupportedAttribute {
	for _, port := range service.Ports {
		if port.Mode == "host" {
			return []api.UnsupportedAttribute{{
				Service: name,
				Path:    "ports[].mode",
				Reason:  "ports[].mode: host is only honored by the Swarm routing mesh",
			}}
		}
	}
	return nil
}
```

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 100/100 |

pkg/compose/unsupported_attributes.go
```diff
@@ -0,0 +1,301 @@
...
+// ignored outside Swarm mode.
+func checkSecretFileReferences(name string, service types.ServiceConfig) []api.UnsupportedAttribute {
+	var findings []api.UnsupportedAttribute
+	for _, secret := range service.Secrets {
+		findings = append(findings, fileReferenceFindings(name, "secrets[]", secret.UID, secret.GID, secret.Mode)...)
+	}
+	return findings
+}
+
+func fileReferenceFindings(name, prefix, uid, gid string, mode *types.FileMode) []api.UnsupportedAttribute {
```
docker-agent, line 222: **[low] fileReferenceFindings uses generic paths that don't identify which config/secret entry triggered the warning**

The `prefix` passed to `fileReferenceFindings` is always the bare bracket placeholder (`"configs[]"` or `"secrets[]"`), so when a service references two configs that both set `uid`, the caller gets two findings with identical `Service` and `Path` (`configs[].uid`) and the user cannot tell which config entry needs fixing.

Including the entry's `Source` name in the prefix — e.g. `fmt.Sprintf("configs[%s]", config.Source)` — makes each warning self-contained and actionable:

```
service "web": configs[myconfig].uid: uid is not supported outside Swarm mode and will be ignored
```

The change only requires updating the two call-sites in `checkConfigFileReferences` and `checkSecretFileReferences`:

```go
// in checkConfigFileReferences:
findings = append(findings, fileReferenceFindings(name, fmt.Sprintf("configs[%s]", config.Source), config.UID, config.GID, config.Mode)...)

// in checkSecretFileReferences:
findings = append(findings, fileReferenceFindings(name, fmt.Sprintf("secrets[%s]", secret.Source), secret.UID, secret.GID, secret.Mode)...)
```

(No change to `fileReferenceFindings` itself or to the test data — just update the expected `Path` strings in the unit tests.)

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 100/100 |

docker-agent (commented):
### Assessment: 🟢 APPROVE

This PR cleanly centralizes detection of unsupported compose-file attributes into `DetectUnsupportedAttributes`, wired once at `ToProject`. The old ad-hoc warnings in `create.go` and `service_containers.go` are correctly removed. The unit-test coverage and the e2e scenario are solid. Two low-severity usability nits in the new detection logic are noted inline.

cmd/compose/compose.go
```diff
@@ -356,6 +356,10 @@ func (o *ProjectOptions) ToProject(ctx context.Context, dockerCli command.Cli, b
 		return nil, metrics, err
 	}
 
+	for _, finding := range backend.DetectUnsupportedAttributes(project) {
```
docker-agent: **[medium] `DetectUnsupportedAttributes` fires on every `ToProject` call, including name-only lookups, potentially emitting duplicate warnings**

`ToProject` is the single implementation called by every code path that needs a project — including `projectOrName` and `toProjectName`, which invoke it with `cli.WithDiscardEnvFile, cli.WithoutEnvironmentResolution` solely to extract the project name. By wiring `DetectUnsupportedAttributes` unconditionally inside `ToProject`, warnings now fire on every call, not just the ones that need the full project analysis.

In practice, several CLI commands (`down`, `start`, `stop`, `kill`, `logs`, `ps`, `pause`, `remove`, `restart`, `volumes`, `wait`) route through `projectOrName` → `ToProject`. For these commands the user sees a warning even though they are operating on already-running containers — which can be surprising or noisy. More importantly, if any command calls `ToProject` more than once (e.g. to re-resolve after environment expansion via `WithServices`), the user would see each warning twice.

**Suggestion**: thread a boolean flag (or an `opts ...LoadOption` variant) through `ToProject` to suppress detection when the call is for name resolution only, or move the detection call to a point earlier in each command's RunE that is only reached once per user-visible invocation (e.g., inside `WithServices`).

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 57/100 |

docker-agent (commented):

- [medium] `pkg/e2e/testdata/TestUnsupportedAttributesWarning/compose.yaml`:52 — Cluster volume source `my-csi-volume` not declared in top-level `volumes:` may cause loader rejection before warnings are checked (confidence: weak 45/100)
  The e2e compose.yaml references `type: cluster, source: my-csi-volume` in the `cluster-vol` service, but there is no top-level `volumes:` declaration for `my-csi-volume`. If compose-go validates that cluster volume sources must be declared at the top level, `docker compose config` would fail before `DetectUnsupportedAttributes` runs, causing the e2e test to fail silently. The verifier notes this is speculative — `type: cluster` may bypass this validation since CSI volumes are externally managed — but it is worth confirming by running the e2e suite.

pkg/compose/unsupported_attributes.go
```diff
@@ -0,0 +1,307 @@
...
+	checkConfigsLabels,
+	checkSecretsDriverOptsAndLabels,
+}
+
+// DetectUnsupportedAttributes reports service/project attributes accepted by
+// the compose-file schema but not honored by this runtime outside Swarm
+// mode.
+func (s *composeService) DetectUnsupportedAttributes(project *types.Project) []api.UnsupportedAttribute {
+	var findings []api.UnsupportedAttribute
+	for name, service := range project.Services {
```
docker-agent: **[medium] `DetectUnsupportedAttributes` returns findings in non-deterministic map iteration order**

`project.Services` is a `map[string]ServiceConfig` (Go map), so `for name, service := range project.Services` iterates in a different random order on each invocation. The per-check helpers `checkConfigsLabels` and `checkSecretsDriverOptsAndLabels` also range over maps internally. The caller in `cmd/compose/compose.go` prints each returned finding immediately:

```go
for _, finding := range backend.DetectUnsupportedAttributes(project) {
    logrus.Warn(finding)
}
```

This means users running the same command twice may see warnings in a different order. The unit test works around non-determinism by sorting findings before comparing — a clear signal that the authors are aware the output is unordered. The API comment states it is a "pure function … safe to call multiple times", which implies repeatable output, but that contract is currently broken for multi-service projects.

**Suggested fix**: Sort the returned slice before returning it, e.g.:

```suggestion
	var findings []api.UnsupportedAttribute
	for name, service := range project.Services {
		for _, check := range serviceChecks {
			findings = append(findings, check(name, service)...)
		}
	}
	for _, check := range projectChecks {
		findings = append(findings, check(project)...)
	}
	slices.SortFunc(findings, func(a, b api.UnsupportedAttribute) int {
		if c := strings.Compare(a.Service, b.Service); c != 0 {
			return c
		}
		return strings.Compare(a.Path, b.Path)
	})
	return findings
```

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 92/100 |

pkg/compose/unsupported_attributes.go
```diff
@@ -0,0 +1,307 @@
...
+// non-empty value would fire on virtually every service that exposes ports.
+// Each matching port is reported individually, identified by its target
+// port, so multiple host-mode ports on one service stay distinguishable.
+func checkPortsMode(name string, service types.ServiceConfig) []api.UnsupportedAttribute {
+	var findings []api.UnsupportedAttribute
+	for _, port := range service.Ports {
+		if port.Mode == "host" {
+			findings = append(findings, api.UnsupportedAttribute{
+				Service: name,
+				Path:    fmt.Sprintf("ports[%d].mode", port.Target),
```
docker-agent: **[medium] `checkPortsMode` uses `port.Target` (port number) as array subscript, producing identical paths for same-port TCP/UDP entries**

The `Path` field is built as `fmt.Sprintf("ports[%d].mode", port.Target)`, where `port.Target` is the container port number (e.g. 80), not the slice index. This is intentional (documented in the comment: "identified by its target port"), but it has a correctness hazard: if a service exposes the same port for both TCP and UDP with `mode: host`, both findings produce the identical string `ports[80].mode`, making them indistinguishable in the output.

Docker Compose does support specifying the same target port with different protocols:
```yaml
ports:
  - target: 80
    published: "8080"
    protocol: tcp
    mode: host
  - target: 80
    published: "8080"
    protocol: udp
    mode: host
```

Both would emit `service "x": ports[80].mode: …`, making it impossible to tell which entry (or whether both) triggered the warning.

**Suggested fix**: Include the protocol in the path when it is non-empty:

```suggestion
				Path:    fmt.Sprintf("ports[%d/%s].mode", port.Target, port.Protocol),
```

(This matches the notation used in Docker port-mapping output, e.g. `0.0.0.0:8080->80/tcp`.) Alternatively use the slice index `i` with `for i, port := range service.Ports`.

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 92/100 |

pkg/compose/service_containers.go
```diff
@@ -223,7 +223,6 @@ func (s *composeService) waitDependency(ctx context.Context, dep string, config
 		case types.ServiceConditionCompletedSuccessfully:
 			done, err = s.checkDependencyCompleted(ctx, dep, config, waitingFor)
 		default:
-			logrus.Warnf("unsupported depends_on condition: %s", config.Condition)
 			return nil
```
docker-agent, line 240: **[high] Removing the runtime `logrus.Warn` from `waitDependency`'s `default` branch leaves `docker compose start` silently skipping dependency waits for unknown conditions**

The PR removes the `logrus.Warnf("unsupported depends_on condition: %s", config.Condition)` call from the `default:` branch and replaces it with a load-time check (`checkDependsOnCondition` called during `ToProject`). However, the `docker compose start` command takes a different code path that bypasses `ToProject` entirely:

1. `start` calls `projectFromName` to reconstruct a project from live container labels.
2. `projectFromName` reads the dependency condition directly from the `com.docker.compose.depends_on` container label (format: `"dep:condition:restart"`), setting it verbatim at `condition = dcArr[1]`.
3. This reconstructed project is passed directly to `waitDependency`  — `DetectUnsupportedAttributes` is never called.

If a container was created by an older Compose version with a non-standard condition stored in the label, or if the label is set manually, `waitDependency` will now silently hit `default: return nil` — no wait is performed, no error is returned, and no warning is emitted. The dependent service will start before its dependency reaches the expected state, which can cause silent startup failures.

The original runtime warning was a safety net precisely for this kind of runtime-only code path. Consider reinstating it as a defensive fallback:

```go
default:
    logrus.Warnf("unsupported depends_on condition %q, skipping wait for %q", config.Condition, dep)
    return nil
```

| Confidence | Score |
| :--: | :--: |
| 🟠 weak | 52/100 |

docker-agent (commented):

This PR cleanly centralizes the unsupported-attribute detection into a pure `DetectUnsupportedAttributes` function and wires it at the shared `ToProject` choke-point. The design is sound — the old ad-hoc `logrus.Warn` calls in `create.go` and `service_containers.go` are properly removed and the table-driven approach is well-tested. Three issues found:

1. **Non-deterministic warning order** (pkg/compose/unsupported_attributes.go:58) — map iteration over `project.Services` makes warning output non-deterministic across runs.
2. **Ambiguous port path for TCP/UDP collisions** (pkg/compose/unsupported_attributes.go:173) — `ports[80].mode` is used for both TCP and UDP port 80, producing identical `Path` strings when both use `mode: host`.
3. **Runtime warning gap in `waitDependency`** (pkg/compose/service_containers.go:226) — `docker compose start` reconstructs the project from container labels via `projectFromName`, bypassing `ToProject` and `DetectUnsupportedAttributes`; an unknown condition now silently skips the dependency wait with no warning or error.

pkg/compose/service_containers.go
```diff
@@ -223,7 +223,14 @@ func (s *composeService) waitDependency(ctx context.Context, dep string, config
 		case types.ServiceConditionCompletedSuccessfully:
 			done, err = s.checkDependencyCompleted(ctx, dep, config, waitingFor)
 		default:
-			logrus.Warnf("unsupported depends_on condition: %s", config.Condition)
+			// checkDependsOnCondition already warns about this at load time,
+			// but only for projects loaded with warnings enabled: start/restart
+			// intentionally skip that, and start can also rebuild the project
+			// from live container labels without loading a compose file at
+			// all. This is the only signal those paths get.
+			if !isKnownDependsOnCondition(config.Condition) {
```
docker-agent: **[medium] Double warning for unknown `depends_on` conditions when running `up` or `create`**

The PR adds `checkDependsOnCondition` to the load-time `DetectUnsupportedAttributes` machinery, and `WithServices` (used by `up` and `create`) always passes `warnUnsupportedAttributes`. As a result, an unknown condition such as `service_ready` now fires **two** warnings:

1. **Load time** — `ToProject` → `backend.DetectUnsupportedAttributes` → `checkDependsOnCondition` → `logrus.Warn(finding)`
2. **Runtime** — `waitDependencies` → `waitDependency` default branch → `!isKnownDependsOnCondition` == `true` → `logrus.Warnf(…)`

The comment on this branch acknowledges that the load-time warning exists but justifies keeping the runtime one for `start`/`restart` (which skip load-time detection). However, for `up` and `create` the user sees the same condition reported twice, which is confusing and regresses from the pre-PR behaviour of a single runtime warning.

**Suggested fix**: skip the runtime warning when the load-time warning already fired, for example by gating on whether the project was loaded with `warnUnsupportedAttributes`. Alternatively, move `checkDependsOnCondition` out of `serviceChecks` and keep the runtime warning as the sole signal — it already carries the richer message format introduced in this PR (`service %q: unsupported depends_on condition %q, skipping wait for %q`).

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 97/100 |

pkg/compose/unsupported_attributes.go
```diff
@@ -0,0 +1,331 @@
...
+var serviceChecks = []unsupportedAttributeCheck{
+	checkDeploySimpleFields,
+	checkDeployPlacement,
+	checkCredentialSpec,
+	checkLabelFile,
+	checkPortsMode,
+	checkClusterVolumes,
+	checkConfigFileReferences,
+	checkSecretFileReferences,
+	checkDependsOnCondition,
```
docker-agent: **[low] `checkDependsOnCondition` is not a Swarm-only check but is grouped with Swarm-only checks**

All other entries in `serviceChecks` flag attributes that are *accepted by the schema but only meaningful in Swarm mode* (`deploy.mode`, `credential_spec`, cluster volumes, etc.). An unknown `depends_on.X.condition` value is different: it is unrecognized in **any** context — Swarm mode would also silently skip an unknown condition. Grouping it here conflates two distinct categories:

- **Swarm-only attributes**: valid outside Swarm but have no effect (e.g., `deploy.placement`)
- **Unrecognized schema values**: not understood by the runtime in any mode (e.g., `service_ready`)

In practice the warning is useful either way, but the semantic mismatch could mislead users into thinking the condition _would_ work in Swarm mode when it would not. Consider either documenting this distinction in a comment, or surfacing the finding with a different message (e.g., `"unrecognized depends_on condition: …"` rather than implicitly grouping it with Swarm-specific attribute warnings).

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 67/100 |

docker-agent (commented):

ndeloof:
IMHO this would be simpler to implement this as a compose-go loader option, based on a set of yaml.Path pattern of supported attributs, so any unsupported (including newer) attribute get identified as unsupported/ignored

docker-agent (commented):
🟢 **No issues found** — LGTM! [View logs](https://github.com/docker/compose/actions/runs/34203614029).

ndeloof:
Alternative approach implemented by Claude: https://github.com/compose-spec/compose-go/pull/926

docker-agent (commented):
⏱️ **PR Review Timed Out** — The review agent hit the 2700 s time limit. This usually happens on large or complex diffs. Re-request a review from `docker-agent` to retry — if it times out again, consider splitting the PR into smaller pieces.

docker-agent (commented):
### Assessment: 🟢 APPROVE

This PR is well-structured — the allowlist-based fail-closed approach cleanly solves the silent-misconfiguration gap from #13150. The compose-go loader integration, the value-conditional attribute checks (ports mode, cluster volumes, file-reference uid/gid/mode), and the sync.OnceValue lazy schema parse are all sound. Test coverage is thorough.

One low-severity cosmetic issue is posted inline. One additional speculative finding is listed below.

- [medium] `pkg/compose/unsupported_attributes.go`:292 — `nil` passed as second argument to `loader.WithSupportedAttributes` could cause a nil pointer dereference if the fork's API dereferences it unconditionally; based on naming conventions an optional callback is more likely, making this safe (confidence: ⚪ negligible 26/100)

pkg/compose/unsupported_attributes.go
```diff
@@ -0,0 +1,294 @@
...
+// fileReferenceReport drives the configs/secrets entries above: uid/gid/mode
+// on a service-level configs:/secrets: reference are silently ignored
+// outside Swarm mode. Each non-nil sub-field on the matched reference
+// produces its own finding, identified by the reference's source, so
+// multiple references — or multiple flagged sub-fields on the same one —
+// stay distinguishable.
+func fileReferenceReport(kind string) func(loader.UnsupportedAttribute) []api.UnsupportedAttribute {
+	return func(finding loader.UnsupportedAttribute) []api.UnsupportedAttribute {
+		ref, _ := finding.Value.(map[string]any)
+		source, _ := ref["source"].(string)
```
docker-agent, line 207: **[low] Missing guard for empty `source` produces malformed path `configs..uid` in warning**

When a service-level `configs` or `secrets` reference is matched by `hasFileReferenceOverride`, `fileReferenceReport` reads `source` from the map via a comma-ok type assertion that silently returns `""` if the key is absent:

```go
source, _ := ref["source"].(string)
// ...
Path: fmt.Sprintf("%s.%s.%s", kind, source, field),
```

If `source` is empty the formatted path becomes `"configs..uid"` — two consecutive dots — which is syntactically malformed and produces a confusing warning like:
```
service "web": configs..uid: uid is not supported outside Swarm mode and will be ignored
```

The Compose Specification states that a service-level `configs` entry's `source` is required when `uid`/`gid`/`mode` are set, and compose-go typically normalises it, so this edge case is unlikely in practice. However, the code has no defensive fallback. A simple guard improves the user-facing message:

```suggestion
		source, _ := ref["source"].(string)
		if source == "" {
			source = "(anonymous)"
		}
```

Alternatively, keep it as a single fallback expression:
`source, _ := ref["source"].(string); if source == "" { source = "(anonymous)" }`

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 67/100 |

pkg/compose/service_containers.go
```diff
@@ -223,7 +240,18 @@ func (s *composeService) waitDependency(ctx context.Context, dep string, config
...
+			// The compose-spec schema declares condition as a closed enum,
+			// so a compose file can never carry an unrecognized value —
+			// schema.Validate rejects it at load time (see
+			// unsupported_attributes.go's comment on valueConditionalAttributes).
+			// This warning is for the two situations that never go through
+			// schema validation at all: a project rebuilt from live
+			// container labels (see projectFromName), or one written by an
+			// older/newer Compose version using a condition value this
+			// build doesn't recognize.
+			if !isKnownDependsOnCondition(config.Condition) {
```
docker-agent: **[low] `isKnownDependsOnCondition` guard in `default` branch is always-true dead code**

The `switch` in `waitDependency` has explicit `case` arms for every condition in `knownDependsOnConditions` (`ServiceConditionStarted`, `ServiceConditionHealthy`, `ServiceConditionCompletedSuccessfully`, `ServiceConditionRunningOrHealthy`). The `default` branch is therefore only reachable when the condition matches *none* of those — i.e. `isKnownDependsOnCondition` always returns `false` there, making the `if !isKnownDependsOnCondition(config.Condition)` guard permanently `true`. The warning fires unconditionally for every unknown condition (which is the correct behaviour), but the guard is dead code that creates a false impression it could suppress the warning.

```suggestion
			logrus.Warnf("service %q: unsupported depends_on condition %q, skipping wait for %q", dependant, config.Condition, dep)
```

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 100/100 |

pkg/compose/unsupported_attributes.go
docker-agent, line 207: **[low] `fileReferenceReport` may produce a malformed double-dot path when `source` is absent**

`source, _ := ref["source"].(string)` silently yields `""` if the raw map entry has no `source` key. `fmt.Sprintf("%s.%s.%s", kind, source, field)` then produces `"configs..uid"` — a double-dot path that is confusing in the warning output.

In practice compose-go's loader normalises every service-level config/secret reference so that `source` is always populated (defaulting to the config/secret name). However, this function receives the raw YAML map from the compose-go tree-walker, not the fully-normalised `types.ServiceConfig`, so the normalisation may not yet have run at call-time. Adding a fallback guard makes the output unambiguous regardless of loader ordering:

```suggestion
			source, _ := ref["source"].(string)
			if source == "" {
				source = "<unknown>"
			}
```

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 57/100 |

docker-agent (commented):
### Assessment: 🟢 APPROVE

- [low] `pkg/compose/unsupported_attributes.go`:60 — `label_file` warning message claims Swarm-only restriction; the feature is unimplemented in this runtime (not Swarm-specific) — users may incorrectly infer Swarm is required (confidence: weak 52/100)

go.mod
```diff
@@ -145,3 +145,5 @@ exclude (
 	github.com/davecgh/go-spew v1.1.2-0.20180830191138-d8f796af33cc
 	github.com/pmezard/go-difflib v1.0.1-0.20181226105442-5d4384ee4fb2
 )
+
+replace github.com/compose-spec/compose-go/v2 => github.com/glours/compose-go/v2 v2.0.0-20260908091111-b3a5d864f669
```
docker-agent: **[medium] go.mod replace directive points to a personal fork — shipping risk if not removed before release**

`go.mod` contains:
```
replace github.com/compose-spec/compose-go/v2 => github.com/glours/compose-go/v2 v2.0.0-20260908091111-b3a5d864f669
```

The PR description acknowledges this is **temporary** (pending an unreleased compose-go change), but shipping with this `replace` in place carries real risk:

1. **Fork availability** — `github.com/glours/compose-go` is a personal repository. If it is renamed, made private, force-pushed, or deleted after merge, all reproducible builds of this release break with an unresolvable module reference.
2. **Future-dated pseudo-version** — the commit timestamp `20260908` is from 2026 (today is 2026-09-08), indicating this was cut very recently from HEAD; there is no guarantee the fork's history is stable.
3. **No upstream review guarantee** — the personal fork's commit is not subject to the same review as upstream `compose-spec/compose-go`.

The `replace` must be removed (and the upstream `compose-go` module updated to the released version) before this PR is merged to a release branch. A merge without removing it is a ticking reproducibility failure.

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 70/100 |

docker-agent (commented):

This PR introduces a well-structured fail-closed allowlist mechanism for warning users about unsupported Compose-file attributes. The core detection logic in `pkg/compose/unsupported_attributes.go` is sound, with value-conditional checks for `ports[].mode: host`, cluster volumes, and config/secret file-reference overrides all correctly implemented.

One issue requires attention before merging: the `go.mod` `replace` directive points to a personal fork — see the inline comment.

#### Lower-confidence findings (reviewed but not posted inline)
- [medium] `cmd/compose/unsupported_attributes_test.go`:54 — `captureWarnings` mutates global logrus state (not parallel-test-safe) — DISMISSED by verifier: standard test helper pattern, no production risk
- [medium] `pkg/compose/unsupported_attributes.go`:249 — value-conditional findings Service field overwrite — DISMISSED by verifier: all patterns start with `services.*` so `splitServicePath` always returns non-empty service name

ndeloof (dismissed):

ndeloof:
https://github.com/compose-spec/compose-go/pull/927 has been merged, you can bump compose-go ref to latest commit then we can merge

ndeloof (approved):
CI green across the matrix, all docker-agent findings addressed at head (runtime warning kept in waitDependency's default branch, anonymous-source guard in fileReferenceReport, no more go.mod replace — compose-go pinned to an upstream pseudo-version). LGTM.

PR #14198, opened 2026-09-08
fix(publish): support short-form port mapping with variable interpolation

### LINKED ISSUE #13672: [BUG] docker compose publish rejects short-form port mapping

Hoomanghkhani:
**What I did**

Fixed an issue where `docker compose publish` failed with:
`'services[...].ports[0]' expected a map or struct, got "string"`
when a service defined short-form port syntax containing variable substitutions (e.g. `${PORT:-3000}:3000`).

During `preChecks`, `loadUnresolvedFile` runs with `options.SkipInterpolation = true` to detect uncommitted secrets and suspicious literals. Because interpolation is skipped, `types.ParsePortConfig` cannot parse un-interpolated port expressions into numeric ports, leaving them as raw string slices. Mapstructure subsequently failed attempting to decode strings into `types.ServicePortConfig` structs.

To fix this:
1. `loadUnresolvedFile` now loads the model via `loader.LoadModelWithContext` and deletes `ports` from services before calling `loader.Transform`. Neither `collectEnvCheckFindings` nor `checkForSensitiveData` inspects ports (only environment, env_files, extends, and configs are checked).
2. `composeFileAsByteReader` now reads the compose file directly from disk via `os.ReadFile`, guaranteeing the secret scanner examines the exact file bytes to be published without loss of comments or unnecessary decoding.
3. Added unit tests in `publish_test.go` verifying `loadUnresolvedFile`, `checkForSensitiveData`, and `collectEnvCheckFindings` with short-form port syntax containing variable substitutions.

**Related issue**
Fixes #13672

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -144,3 +144,5 @@ exclude (
 	github.com/davecgh/go-spew v1.1.2-0.20180830191138-d8f796af33cc
 	github.com/pmezard/go-difflib v1.0.1-0.20181226105442-5d4384ee4fb2
 )
+
+replace github.com/compose-spec/compose-go/v2 => github.com/compose-spec/compose-go/v2 v2.15.1-0.20260910130034-0e17c6437ebb
--- a/go.sum
+++ b/go.sum
@@ -32,8 +32,8 @@ github.com/cloudflare/circl v1.6.3 h1:9GPOhQGF9MCYUeXyMYlqTR6a5gTrgR/fBLXvUgtVcg
 github.com/cloudflare/circl v1.6.3/go.mod h1:2eXP6Qfat4O/Yhh8BznvKnJ+uzEoTQ6jVKJRn81BiS4=
 github.com/codahale/rfc6979 v0.0.0-20141003034818-6a90f24967eb h1:EDmT6Q9Zs+SbUoc7Ik9EfrFqcylYqgPZ9ANSbTAntnE=
 github.com/codahale/rfc6979 v0.0.0-20141003034818-6a90f24967eb/go.mod h1:ZjrT6AXHbDs86ZSdt/osfBi5qfexBrKUdONk989Wnk4=
-github.com/compose-spec/compose-go/v2 v2.15.1-0.20260908103050-cda18529aca7 h1:D9jScaeUAk7DjyB71SD9zvOFPR6klw0FxPcR2bgu72c=
-github.com/compose-spec/compose-go/v2 v2.15.1-0.20260908103050-cda18529aca7/go.mod h1:Q1+qtN4vhzEjGrnqRtzx1xa8raDZQlMUe3WJxndYNiQ=
+github.com/compose-spec/compose-go/v2 v2.15.1-0.20260910130034-0e17c6437ebb h1:2r/BSgm9NJLvTF84ULhXkq94woFgD2A5/p9PjY6Vmug=
+github.com/compose-spec/compose-go/v2 v2.15.1-0.20260910130034-0e17c6437ebb/go.mod h1:Q1+qtN4vhzEjGrnqRtzx1xa8raDZQlMUe3WJxndYNiQ=
 github.com/containerd/cgroups/v3 v3.1.3 h1:eUNflyMddm18+yrDmZPn3jI7C5hJ9ahABE5q6dyLYXQ=
 github.com/containerd/cgroups/v3 v3.1.3/go.mod h1:PKZ2AcWmSBsY/tJUVhtS/rluX0b1uq1GmPO1ElCmbOw=
 github.com/containerd/console v1.0.5 h1:R0ymNeydRqH2DmakFNdmjR2k0t7UPuiOV/N/27/qqsc=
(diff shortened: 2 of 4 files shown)


glours (changes requested):
Thanks for digging into this, the diagnosis is correct, but I see one blocker and a broader issue with the proposed fix.

## Blocker: secret detection regression

Changing `composeFileAsByteReader` to use `os.ReadFile` bypasses the canonicalization performed by `loadUnresolvedFile`. The secret scanner doesn’t detect list-form environment entries such as:

```yaml
environment:
  - MYSQL_ROOT_PASSWORD=secret
```

Canonicalizing them to map form is what makes the scanner detect the secret, which is why that round-trip was introduced in [[#12620](https://github.com/docker/compose/pull/12620)](https://github.com/docker/compose/pull/12620). Using the raw file would silently regress the `docker compose publish` secret check, so this needs to be addressed before merge.

## `ports` is only one symptom

Removing `ports` before `loader.Transform` fixes this case, but other typed fields fail for the same reason when interpolation is skipped, including:

```yaml
deploy:
  replicas: ${REPLICAS}
mem_limit: ${MEM}
healthcheck:
  retries: ${RETRIES}
```

The underlying problem is the strict conversion of an unresolved model into `types.Project`.

## Suggested approach

Neither consumer needs that full conversion:

* `collectEnvCheckFindings` can decode only the fields it uses.
* `composeFileAsByteReader` can keep `LoadModelWithContext`, normalize list-form `environment` values with `types.NewMappingWithEquals`, and marshal the raw model directly to YAML. This preserves full-file scanning without strict-decoding unrelated fields.

We’ll also need to temporarily pin [[compose-go#928](https://github.com/compose-spec/compose-go/pull/928)](https://github.com/compose-spec/compose-go/pull/928), as it fixes unresolved variables in `build.ssh` and `ulimits` during canonicalization:

```go
replace github.com/compose-spec/compose-go/v2 => github.com/compose-spec/compose-go/v2 v2.15.1-0.20260910130034-0e17c6437ebb
```

Finally, could you add a test using `mem_limit` or `deploy.replicas` so the fix covers the whole class of issue rather than only `ports`?

Hoomanghkhani:
Thanks for the thorough review and guidance @glours!

I have refactored the fix following the suggested approach:

1. **Secret Detection & Canonicalization:**
   - Reverted `os.ReadFile` in `composeFileAsByteReader`. It now retains `loader.LoadModelWithContext`, normalizes list-form environment variables to map form using `types.NewMappingWithEquals`, and marshals the raw dictionary directly to YAML.
   - Added `Test_checkForSensitiveData_list_form_secret` to ensure secrets declared in list-form environments are properly caught by the scanner.

2. **Selective Decoding for Pre-checks:**
   - Rather than stripping `ports` or decoding into a full `types.Project`, `loadUnresolvedFile` now decodes strictly into a lightweight `unresolvedFile` struct containing only the fields inspected by `collectEnvCheckFindings` (`environment`, `env_file`, `extends`, and config `content`).
   - Unrelated typed fields (`ports`, `deploy`, `mem_limit`, `healthcheck`, etc.) are ignored during mapstructure decoding and will no longer fail on un-interpolated variable syntax.

3. **Pinned compose-go#928:**
   - Added the temporary `replace` directive in `go.mod` to bring in `compose-go#928`.

4. **Extended Test Coverage:**
   - Updated existing tests (`loadUnresolvedFile`, `checkForSensitiveData`, and `collectEnvCheckFindings`) to include `mem_limit: ${MEM}`, `deploy.replicas: ${REPLICAS}`, and `healthcheck.retries: ${RETRIES}` alongside short-form port syntax with variable substitutions.

The PR has been updated with these changes.

PR #14200, opened 2026-09-08
plan engine: reconciler plans the start phase (epic #14081, lot 1)

ndeloof:

It teaches Compose's **plan engine** to describe *starting* an application — until now it could only describe *creating* it — without changing any current behavior: this is vocabulary, and nobody speaks it yet.

## Context

Today `up` works in two steps with two different engines: first the plan engine computes and executes "what to create" (deterministic, testable, comparable against a golden file), then a second, imperative engine improvises the start as it goes — waiting for the database to be healthy, running `pre_start` hooks, starting containers in dependency order, running `post_start`. That second engine is neither inspectable nor testable the same way, and the seam between the two is a known source of bugs (two different snapshots of the daemon state, two event-emission systems).

## What the PR brings

The plan can now **express** everything that start phase does: "wait for `db` to be healthy", "run `web`'s hooks", "start replica 1 then 2", "`app` only starts once `db` is ready". The whole start logic becomes data that can be displayed, reviewed and verified before being executed — instead of behavior that only exists at the moment it happens.

Two guardrails frame the PR:

- **Inert by construction**: the new capability sits behind an option no command passes yet. `compose create`, `up` — everything produces exactly the same plans as before, byte for byte.
- **Faithful to existing behavior**: every subtlety of the current imperative engine is reproduced and locked by tests — dependency conditions are re-verified even when everything is already running, a paused container being restarted gets neither a redundant restart nor a redundant injection, a container doomed to disappear (replica scale-down) is never started, and error cases produce today's messages word for word.

## Why this is the right first brick

This is the first item of Lot 1 of epic #14081: lay down the vocabulary and prove its fidelity *before* wiring anyone onto it. The next PRs will make the executor speak this vocabulary, then switch `up`, `start` and `scale` over one consumer at a time — each staying small and revertible, because the hard part (saying exactly the same thing as the existing engine) will already have been validated here.

---
Split out of #14156 to keep the review focused.

### DIFF
--- a/pkg/compose/reconcile.go
+++ b/pkg/compose/reconcile.go
@@ -17,6 +17,7 @@
 package compose
 
 import (
+	"cmp"
 	"context"
 	"fmt"
 	"slices"
@@ -45,8 +46,22 @@ func toReconcileOptions(options api.CreateOptions) ReconcileOptions {
 	}
 }
 
+// ReconcileScope selects which lifecycle phases the plan covers. The zero
+// value plans the Create phase only — the historical behavior, and what
+// `compose create` keeps using. ScopeCreateStart adds the Start phase to the
+// same plan; ScopeStart plans starting the observed containers without
+// converging them first.
+type ReconcileScope int
+
+const (
+	ScopeCreate ReconcileScope = iota
+	ScopeCreateStart
+	ScopeStart
+)
+
 // ReconcileOptions controls how the reconciler compares desired and observed state.
 type ReconcileOptions struct {
+	Scope                ReconcileScope // lifecycle phases to plan; zero = Create only
 	Services             []string       // targeted services (empty = all)
 	Recreate             string         // "diverged", "force", "never" for targeted services
 	RecreateDependencies string         // same for non-targeted services
@@ -77,6 +92,26 @@ type reconciler struct {
 	// serviceNodes tracks the last plan node per service, so dependent
 	// services can order their operations after dependencies.
 	serviceNodes map[string]*PlanNode
+	// containerNodes tracks, per service then replica number, the
+	// create-phase node that materializes that container (creation,
+	// recreation, or the exceptional-state restart), so start-phase nodes
+	// can depend on it and resolve their target from its result.
+	containerNodes map[string]map[int]*PlanNode
+	// startChainEnds tracks the nodes ending each service's start-phase
+	// visit: the last node of the replica chain, or — when nothing had to
+	// start — the dependency prerequisites the visit still evaluated. What a
+	// service_started dependent (or a wait node) hooks onto, matching
+	// InDependencyOrder semantics: a dependent's visit begins only once the
+	// dependency's whole visit (waits included) completed.
+	startChainEnds map[string][]*PlanNode
+	// waitNodes deduplicates OpWaitCondition nodes per (service, condition):
+	// several dependents awaiting the same condition share one node, like
+	// networkNodes deduplicates network creations.
+	waitNodes map[string]*PlanNode
+	// removedByPlan records replicas the create phase condemns (scale-down
+	// stop+remove): the start phase must never plan a start for them — the
+	// imperative engine only ever starts what survives the convergence.
+	removedByPlan map[string]bool
 	// stoppedByPlan records containers already stopped by an earlier stage
 	// of the plan (typically planRecreateNetwork) so that downstream stages
 	// can chain on the existing OpStopContainer instead of emitting a second
@@ -123,6 +158,10 @@ func reconcile(_ context.Context, project *types.Project, observed *ObservedStat
 		networkNodes:                map[string]*PlanNode{},
 		volumeNodes:                 map[string]*PlanNode{},
 		serviceNodes:                map[string]*PlanNode{},
+		containerNodes:              map[string]map[int]*PlanNode{},
+		startChainEnds:              map[string][]*PlanNode{},
+		waitNodes:                   map[string]*PlanNode{},
+		removedByPlan:               map[string]bool{},
 		stoppedByPlan:               map[string]*PlanNode{},
 		connectNodes:                map[string][]*PlanNode{},
 		recreatedServices:           map[string]bool{},
@@ -131,20 +170,28 @@ func reconcile(_ context.Context, project *types.Project, observed *ObservedStat
 
 	r.resolveObserved()
 
-	if err := r.reconcileNetworks(); err != nil {
-		return nil, err
+	if options.Scope != ScopeStart {
+		if err := r.reconcileNetworks(); err != nil {
+			return nil, err
+		}
+
+		if err := r.reconcileVolumes(); err != nil {
+			return nil, err
+		}
+
+		if err := r.reconcileContainers(); err != nil {
+			return nil, err
+		}
+
+		if r.options.RemoveOrphans {
+			r.reconcileOrphans()
+		}
 	}
 
-	if err := r.reconcileVolumes(); err != nil {
-		return nil, err
-	}
-
-	if err := r.reconcileContainers(); err != nil {
-		return nil, err
-	}
-
-	if r.options.RemoveOrphans {
-		r.reconcileOrphans()
+	if options.Scope != ScopeCreate {
+		if err := r.planStartPhase(); err != nil {
+			return nil, err
+		}
 	}
 
 	return r.plan, nil
@@ -261,7 +308,7 @@ func (r *reconciler) planRecreateNetworks(keys []string) {
 		var disconnectNodes []*PlanNode
 		for i := range containers {
 			oc := &containers[i]
-			resID := fmt.Sprintf("service:%s:%d", oc.Summary.Labels[api.ServiceLabel], oc.Number)
+			resID := serviceReplicaID(oc.Summary.Labels[api.ServiceLabel], oc.Number)
 			stopNode, alreadyStopped := r.stoppedByPlan[oc.ID]
 			if !alreadyStopped {
 				stopNode = r.plan.addNode(Operation{
@@ -324,7 +371,7 @@ func (r *reconciler) planRecreateNetworks(keys []string) {
 		// transitively through remove → create).
 		for i := range containers {
 			oc := &containers[i]
-			resID := fmt.Sprintf("service:%s:%d", oc.Summary.Labels[api.ServiceLabel], oc.Number)
+			resID := serviceReplicaID(oc.Summary.Labels[api.ServiceLabel], oc.Number)
 			deps := []*PlanNode{createNode}
 			if rename {
 				deps = append(deps, disconnectNodes[i])
@@ -449,7 +496,7 @@ func (r *reconciler) planRecreateVolumes(keys []string) {
 	var removeNodes []*PlanNode
 	for i := range containers {
 		oc := &containers[i]
-		resID := fmt.Sprintf("service:%s:%d", oc.Summary.Labels[api.ServiceLabel], oc.Number)
+		resID := serviceReplicaID(oc.Summary.Labels[api.ServiceLabel], oc.Number)
 		stopNode, alreadyStopped := r.stoppedByPlan[oc.ID]
 		if !alreadyStopped {
 			stopNode = r.plan.addNode(Operation{
@@ -593,12 +640,12 @@ func (r *reconciler) reconcileContainers() error {
 	}
 
 	// Visit in dependency order (leaves first = services with no deps)
-	return r.visitInDependencyOrder(graph)
+	return r.visitInDependencyOrder(graph, r.reconcileService)
 }
 
 // visitInDependencyOrder processes services from leaves to roots so that
 // dependencies are reconciled before the services that depend on them.
-func (r *reconciler) visitInDependencyOrder(g *Graph) error {
+func (r *reconciler) visitInDependencyOrder(g *Graph, visit func(types.ServiceConfig) error) error {
 	visited := map[string]bool{}
 	// Sort vertex keys for deterministic plan output in tests
 	keys := sortedKeys(g.Vertices)
@@ -631,7 +678,7 @@ func (r *reconciler) visitInDependencyOrder(g *Graph) error {
 		if err != nil {
 			return err
 		}
-		if err := r.reconcileService(service); err != nil {
+		if err := visit(service); err != nil {
 			return err
 		}
 	}
@@ -695,24 +742,27 @@ func (r *reconciler) reconcileService(service types.ServiceConfig) error {
 			// Scale down: stop + remove excess containers. Track the remove
 			// node so dependent services wait for the scale-down to finish
 			// even when no other operation runs on this service.
+			resID := serviceReplicaID(service.Name, oc.Number)
 			stopNode := r.plan.addNode(Operation{
 				Type:       OpStopContainer,
-				ResourceID: fmt.Sprintf("service:%s:%d", service.Name, oc.Number),
+				ResourceID: resID,
 				Cause:      "scale down",
 				Container:  &containers[i].Summary,
 				Timeout:    r.options.Timeout,
 			}, "")
 			lastNode = r.plan.addNode(Operation{
 				Type:       OpRemoveContainer,
-				ResourceID: fmt.Sprintf("service:%s:%d", service.Name, oc.Number),
+				ResourceID: resID,
 				Cause:      "scale down",
 				Container:  &containers[i].Summary,
 			}, "", stopNode)
+			r.removedByPlan[resID] = true
 			continue
 		}
 
 		if r.mustRecreate(service, expectedHash, parentRecreated, oc, strategy) {
 			lastNode = r.planRecreateContainer(service, &containers[i], infraDeps)
+			r.setContainerNode(service.Name, oc.Number, lastNode)
 			r.recreatedServices[service.Name] = true
 			continue
 		}
@@ -729,10 +779,11 @@ func (r *reconciler) reconcileService(service types.ServiceConfig) error {
 			// Any other state (paused, dead, ...): attempt to (re)start
 			lastNode = r.plan.addNode(Operation{
 				Type:       OpStartContainer,
-				ResourceID: fmt.Sprintf("service:%s:%d", service.Name, oc.Number),
+				ResourceID: serviceReplicaID(service.Name, oc.Number),
 				Cause:      "not running",
 				Container:  &containers[i].Summary,
 			}, "", infraDeps...)
+			r.setContainerNode(service.Name, oc.Number, lastNode)
 		}
 	}
 
@@ -744,12 +795,13 @@ func (r *reconciler) reconcileService(service types.ServiceConfig) error {
 		serviceCopy := service // copy for pointer stability
 		lastNode = r.plan.addNode(Operation{
 			Type:       OpCreateContainer,
-			ResourceID: fmt.Sprintf("service:%s:%d", service.Name, number),
+			ResourceID: serviceReplicaID(service.Name, number),
 			Cause:      "no existing container",
 			Service:    &serviceCopy,
 			Number:     number,
 			Name:       name,
 		}, "", infraDeps...)
+		r.setContainerNode(service.Name, number, lastNode)
 	}
 
 	if lastNode != nil {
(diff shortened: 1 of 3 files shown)


pkg/compose/reconcile.go
```diff
@@ -877,6 +922,289 @@ func (r *reconciler) hasVolumeMismatch(expected types.ServiceConfig, oc Observed
...
+		first := replicas[0]
+		op := Operation{
+			Type:         OpRunPreStart,
+			ResourceID:   first.resID,
+			Cause:        "pre_start hooks",
+			Service:      &serviceCopy,
+			Container:    first.container,
+			CreateNodeID: first.createNodeID,
+		}
+		deps := prev
```
docker-agent: **[medium] pre_start dep-list aliasing: `deps := prev` may silently mutate `depNodes` backing array**

`deps := prev` on this line is a shallow copy of the slice header — `deps` and `prev` (and therefore `depNodes`) all share the same backing array. When `first.after != nil`, the `append(deps, first.after)` call may write into that shared backing array without reallocating (if capacity is sufficient). Because `prev` is then immediately overwritten with `[]*PlanNode{preStart}`, subsequent `slices.Clone(prev)` calls in the replica loop are safe. However, the **`depNodes` slice** (which holds the same backing array) is silently left with `first.after` appended into its hidden capacity.

**Current impact**: `depNodes` is not read again after this block, so there is no observable defect today.

**Latent hazard**: If a future change reads `depNodes` after this block — e.g., to log, audit, or reuse the dependency list — it will see a corrupted slice with an extra element. The fix is a one-liner:

```suggestion
		deps := slices.Clone(prev)
```

This makes the intent explicit (no aliasing) and is consistent with the identical `slices.Clone(prev)` pattern used on line 1183 in the replica loop.

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 62/100 |

pkg/compose/reconcile.go
```diff
@@ -877,6 +922,289 @@ func (r *reconciler) hasVolumeMismatch(expected types.ServiceConfig, oc Observed
...
+		// container at all cannot be started — only reachable under scope
+		// Start, since CreateStart would have planned the missing creates
+		return fmt.Errorf("service %q has no container to start", service.Name)
+	}
+	// dependency conditions are evaluated even when nothing has to start:
+	// the imperative engine calls waitDependencies for every visited service
+	// before looking at what to start, so an up with everything running
+	// still fails on an unhealthy required dependency
+	depNodes := r.startPhaseDependencies(service)
+	if len(replicas) == 0 {
```
docker-agent, line 1165: **[medium] `startChainEnds` not populated for all-running service under `ScopeStart` — `service_started` dependents may have no ordering constraint**

When `planServiceStart` is called under `ScopeStart` for a service whose containers are all already running (`len(replicas) == 0`, `anyRunning == true`), the code sets `startChainEnds[service.Name]` only when `ends` is non-empty:

```go
ends := depNodes
if len(ends) == 0 {
    if node, ok := r.serviceNodes[service.Name]; ok {
        ends = []*PlanNode{node}
    }
}
if len(ends) > 0 {
    r.startChainEnds[service.Name] = ends  // only when non-empty
}
```

Under `ScopeStart`, `serviceNodes` is not populated (no create phase), so if the service also has no `depends_on` of its own (`depNodes` is empty), `startChainEnds[service.Name]` is never set.

A downstream service with `condition: service_started` on this service will then hit `startPhaseDependencies` where both `startChainEnds[dep]` and `serviceNodes[dep]` are empty, so `targets` is nil. For a `waitless` (service_started) condition, `depNodes = append(depNodes, targets...)` appends nothing — the dependent gets zero ordering edges relative to this service, violating the `service_started` semantics.

**Concrete scenario**: `db` → `app` (service_started) → `web` (service_started), all containers already running under `ScopeStart`. `db` is visited first, sets `startChainEnds["db"]` (it has a condition wait for itself). `app` is visited next, all running, no own `depends_on` except `db` — `depNodes` = wait-node-for-db. `app` sets `startChainEnds["app"] = [wait-node-for-db]`. So in this specific chain the bug may not manifest. However, for a **leaf** dependency (no `depends_on` of its own) that is all-running, `depNodes` is empty, `serviceNodes` is empty, and `startChainEnds` goes unset.

**Fix**: Unconditionally set `startChainEnds[service.Name]` even when `ends` is empty — an empty slice is still a valid signal that the service was visited and has no start-phase work. Alternatively, use a sentinel "already visited" marker so dependents know to add no ordering edges (which is correct: if dep is fully running, dependent can proceed freely).

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 57/100 |

docker-agent (commented):

2 inline findings (both moderate confidence). No high-severity issues found.

pkg/compose/reconcile.go
ndeloof: Fixed in dca392d96: the pre_start block clones prev before appending (with a comment naming the shared-backing-array hazard), matching the replica loop.

pkg/compose/reconcile.go
ndeloof: This is deliberate imperative parity, now documented in dca392d96: under ScopeStart with everything running and no depends_on, the visit had no effect — a service_started condition on a running dependency is already satisfied, and the imperative InDependencyOrder visit imposes no ordering either (waitDependencies no-ops, nothing to start). An edge here would create ordering the imperative engine does not have.

pkg/compose/reconcile.go
```diff
@@ -877,6 +922,295 @@ func (r *reconciler) hasVolumeMismatch(expected types.ServiceConfig, oc Observed
...
+			}
+			replicas = append(replicas, plannedReplica(resID, oc.Number, node))
+			continue
+		}
+		if oc.State == container.StateRunning {
+			anyRunning = true
+			continue
+		}
+		seen[resID] = true
+		replicas = append(replicas, startReplica{resID: resID, number: oc.Number, container: &oc.Summary})
```
docker-agent, line 1064: **[high] `startPhaseReplicas` emits start-phase nodes for containers targeted by scale-down**

When `ScopeCreateStart` is active and a service scales down (e.g. from 2 replicas to 1), `reconcileService` plans `OpStopContainer → OpRemoveContainer` for the excess container **but does not register it in `containerNodes`**. Later, `startPhaseReplicas` iterates `r.observed.Containers[service.Name]` — which includes *all* observed containers, including scale-down targets — and, for any container whose state is not `StateRunning`, falls through to line 1039:

```go
seen[resID] = true
replicas = append(replicas, startReplica{resID: resID, number: oc.Number, container: &oc.Summary})
```

A scale-down target whose observed state is `Exited`, `Created`, `Restarting`, `Paused`, or `Dead` satisfies neither the `containerNodes` lookup nor the `StateRunning` guard, so it is added to `replicas` and receives an `OpStartContainer` start-phase node. Because there is no DAG edge between the create-phase stop/remove and the start-phase start, the executor may attempt to start a container that has already been removed, or race a stop with a start.

**Fix**: Track scale-down container IDs during `reconcileService` so `startPhaseReplicas` can skip them. One minimal approach is to add a `scaledDownContainers map[string]bool` field to the reconciler, populate it in the `i >= expected` branch of `reconcileService`, and guard `startPhaseReplicas` with:

```go
if r.scaledDownContainers[resID] {
    continue // being removed by create phase; no start-phase node
}
```

A test covering this case (e.g. a service scaling from 2→1 with one exited replica) would prevent regressions.

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 75/100 |

docker-agent (commented):
### Assessment: 🔴 CRITICAL

One high-severity finding in the new start-phase planning logic.

pkg/compose/reconcile.go
ndeloof: Fixed in a2603d6e0: scale-down targets are recorded in removedByPlan at the moment reconcileService condemns them (the criterion is sortContainers-based, not a numeric cutoff, so the registration is the only faithful filter), and startPhaseReplicas skips them — no start node, and no anyRunning contribution, matching the imperative engine that only starts what survives the convergence. Under a pure ScopeStart nothing is condemned, so compose start keeps its current behavior of starting every observed non-running replica. Golden test added (running replica 1 + condemned exited replica 2 → zero start-phase nodes).

docker-agent (commented):
### Assessment: 🟢 APPROVE

docker-agent (commented):
### Assessment: 🟢 APPROVE

The start-phase planning logic is well-structured and the implementation is correct. All hypotheses raised during analysis were dismissed as false positives:

1. **`seen` map in `startPhaseReplicas`** — `seen[resID] = true` is set unconditionally before the `OpStartContainer` type check, so exceptional-state containers are correctly excluded from the second loop.
2. **`service_started` on fully-running dependency under `ScopeStart`** — producing no prerequisite edge is the intended behavior and matches the imperative engine's semantics: if the dependency is already running, the `service_started` condition is trivially satisfied and no ordering constraint is needed. The code comment at the call site makes this explicit.

The inert-by-construction design (gated behind an option no command passes yet), faithful parity with the imperative engine, and comprehensive test coverage are all solid.

pkg/compose/reconcile.go
```diff
@@ -877,6 +929,302 @@ func (r *reconciler) hasVolumeMismatch(expected types.ServiceConfig, oc Observed
...
+				// create-phase bare start already brings it to running
+				anyRunning = true
+				continue
+			}
+			replicas = append(replicas, plannedReplica(resID, oc.Number, node))
+			continue
+		}
+		if oc.State == container.StateRunning {
+			anyRunning = true
+			continue
```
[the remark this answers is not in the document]
ndeloof: Fixed in 8960d0ef9 — good catch, it was a real parity hole: a running container the create phase stops WITHOUT recreating (network recreate, via stoppedByPlan) is now planned for restart, ordered after its last reconnect node, and does not gate pre_start — exactly the imperative engine restarting it from its second snapshot. Containers stopped for a volume recreate are removed then re-created, so they were already covered through containerNodes. Unit test added.

pkg/compose/reconcile.go
```diff
@@ -877,6 +929,302 @@ func (r *reconciler) hasVolumeMismatch(expected types.ServiceConfig, oc Observed
...
+		}
+		seen[resID] = true
+		replicas = append(replicas, startReplica{resID: resID, number: oc.Number, container: &oc.Summary})
+	}
+	for resID, node := range r.containerNodes {
+		if !seen[resID] && strings.HasPrefix(resID, serviceReplicaPrefix(service.Name)) {
+			replicas = append(replicas, plannedReplica(resID, replicaNumber(resID), node))
+		}
```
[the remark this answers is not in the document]
ndeloof: Fixed in 8960d0ef9: containerNodes is now indexed by service then replica number (map[string]map[int]*PlanNode), so startPhaseReplicas reads its own service's entries only — no per-service scan of the global map. Bonus: the resource-ID prefix matching and number re-parsing disappear entirely (serviceReplicaPrefix/replicaNumber deleted).

pkg/compose/reconcile.go
```diff
@@ -877,6 +929,302 @@ func (r *reconciler) hasVolumeMismatch(expected types.ServiceConfig, oc Observed
...
+		// materialized container" error instead of panicking here
+	}
+	return rep
+}
+
+// serviceReplicaID is the canonical resource ID of one service replica.
+// Every producer and every parser of the "service:<name>:<number>" form goes
+// through these helpers so the format lives in exactly one place. Service
+// names cannot contain ':', which is what makes replicaNumber and
+// serviceReplicaPrefix unambiguous.
```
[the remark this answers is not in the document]
ndeloof: Fixed in 8960d0ef9: the seven remaining direct format sites now go through serviceReplicaID, and with the per-service index nothing parses the ID back — the helper comment now states exactly that (single producer, no parser).

docker-agent (commented):

Two medium-severity findings in the newly-introduced start-phase planning code. Both are LIKELY (moderate confidence, 67/100) and concern correctness of the DAG-building logic under specific conditions.

pkg/compose/reconcile.go
```diff
@@ -877,10 +928,317 @@ func (r *reconciler) hasVolumeMismatch(expected types.ServiceConfig, oc Observed
...
+			Type:       OpWaitCondition,
+			ResourceID: fmt.Sprintf("wait:%s:%s", dep, cfg.Condition),
+			Cause:      "depends_on condition",
+			Name:       dep,
+			Condition:  cfg.Condition,
+			BestEffort: !cfg.Required,
+		}, "", targets...)
+		wait.Phase = PhaseStart
+		r.waitNodes[key] = wait
+	} else if cfg.Required && wait.Operation.BestEffort {
```
docker-agent: **[medium] `waitConditionNode` silently discards `targets` from the second (and subsequent) callers**

When two services share the same `depends_on` dependency + condition (e.g., both `web` and `worker` declare `depends_on: db: condition: service_healthy`), `waitConditionNode` is called twice for key `"db:service_healthy"`. The first call creates the shared wait node with `targets` wired as prerequisites. The second call hits the `else if` branch, which upgrades `BestEffort` if needed — but never merges the second caller's `targets` into the wait node's `DependsOn`:

```go
} else if cfg.Required && wait.Operation.BestEffort {
    wait.Operation.BestEffort = false
    // ← second caller's `targets` are silently dropped here
}
```

**Today this is safe** because `visitInDependencyOrder` processes services in topological order and `startChainEnds[dep]` is written exactly once; every caller for the same `dep` therefore supplies identical `targets`. But the invariant is implicit and fragile: the function signature accepts `targets` on every call yet only honours it on the first. A future caller that supplies a legitimately different set of chain-end nodes (e.g., after an incremental plan mutation) would silently produce an incomplete wait node, breaking the ordering guarantee without any compile-time or runtime warning.

Consider asserting (or at least documenting) that subsequent callers always pass the same `targets`, or merge them when they differ:
```go
} else {
    // targets must be identical to first call — assert in tests or merge defensively
    if cfg.Required && wait.Operation.BestEffort {
        wait.Operation.BestEffort = false
    }
}
```

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 67/100 |

pkg/compose/reconcile.go
```diff
@@ -877,10 +928,317 @@ func (r *reconciler) hasVolumeMismatch(expected types.ServiceConfig, oc Observed
...
+		ends := depNodes
+		if len(ends) == 0 {
+			if node, ok := r.serviceNodes[service.Name]; ok {
+				ends = []*PlanNode{node}
+			}
+		}
+		if len(ends) > 0 {
+			r.startChainEnds[service.Name] = ends
+		}
+		// no entry otherwise (ScopeStart, all running, no depends_on): the
```
docker-agent, line 1177: **[medium] Under `ScopeStart`, a `service_started` dependency edge is silently dropped when the dependency has all containers running**

**Scenario**: `compose start` (`ScopeStart`) where `db` is already fully running and `web` declares `depends_on: db: condition: service_started`.

1. `planServiceStart(db)` finds no replicas to start (`anyRunning=true`, `replicas=[]`). `startPhaseDependencies(db)` returns `[]` (db has no own `depends_on`). `serviceNodes["db"]` is absent (no create phase ran under `ScopeStart`). Nothing is set in `startChainEnds["db"]`.

2. `planServiceStart(web)` calls `startPhaseDependencies(web)`. For dep `"db"`: `startChainEnds["db"]` → absent; `serviceNodes["db"]` → absent → `targets = nil`. The condition is `service_started` (waitless path), so `depNodes = append(depNodes, nil...)` → no-op. **No ordering edge is emitted.**

3. `web` can start concurrently with (or even before) `db` in the plan — even though the `service_started` condition is declared.

The code comment asserts "a `service_started` condition on a running dependency is already satisfied, exactly like the imperative engine's empty visit." This may be true for the execution semantics (the condition is satisfied), but drops the **ordering** guarantee: the plan does not record that `web` must start *after* `db`'s start-phase visit. If the imperative engine implicitly serialises these visits (e.g., via a sequential outer loop), the plan loses that implicit ordering and could start `web` before `db`'s hooks or post-start actions complete.

To be safe, consider synthesising a no-op chain-end node for fully-running services under `ScopeStart`, or explicitly documenting and testing that concurrent start of `web` and `db` is correct when `db` is already up.

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 67/100 |

pkg/compose/reconcile.go
```diff
@@ -877,10 +928,317 @@ func (r *reconciler) hasVolumeMismatch(expected types.ServiceConfig, oc Observed
...
+		replicas = append(replicas, startReplica{resID: resID, number: oc.Number, container: &oc.Summary})
+	}
+	for number, node := range planned {
+		if !seen[number] {
+			replicas = append(replicas, plannedReplica(service.Name, number, node))
+		}
+	}
+	// the map iteration above is unordered: this sort CARRIES the plan's
+	// determinism, on top of expressing the numeric start order
+	slices.SortFunc(replicas, func(a, b startReplica) int { return a.number - b.number })
```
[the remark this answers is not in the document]
ndeloof: Fixed in 057302417: cmp.Compare instead of subtraction.

pkg/compose/reconcile.go
ndeloof: Fixed in 057302417: subsequent callers now merge their targets into the shared wait node's DependsOn (deduplicated). As you note it is safe today — every caller passes the same startChainEnds[dep], fixed before any dependent is visited — so the merge is defensive, with a comment saying exactly that.

pkg/compose/reconcile.go
ndeloof: Deliberate imperative parity, documented at that exact spot in the code (the comment on the len(replicas)==0 early return): under ScopeStart with the dependency fully running, a service_started condition is already satisfied, and the imperative InDependencyOrder visit of that dependency is an empty no-op imposing no ordering either — compose start today starts web without any materialized wait on a running db. Emitting an edge would create ordering the imperative engine does not have. (If the dependency is only PARTIALLY running, it has replicas to start, a chain exists, and the edge IS emitted.)

docker-agent (commented):
### Assessment: 🟢 APPROVE

The plan engine's start-phase extension is well-structured and carefully reasoned. The new code is inert by construction (gated behind an option no existing command passes), and the comprehensive test suite locks every subtlety of the existing imperative engine's behavior.

Key design points reviewed and found sound:
- **Dependency ordering**: `visitInDependencyOrder` correctly traverses leaves-first so `startChainEnds[dep]` is always populated before any dependent is visited.
- **Replica deduplication**: the `seen` map correctly gates the second loop over `containerNodes`, preventing double-counting of observed+planned replicas.
- **Exceptional-state replicas**: create-phase bare restarts (paused, dead) are correctly counted as `anyRunning` and excluded from start-phase replicas, matching the imperative engine's second-snapshot behavior.
- **Pre_start scoping**: the `!anyRunning` gate faithfully reproduces the imperative engine — pre_start is skipped when any replica is already running.
- **Scale-down safety**: `removedByPlan` correctly prevents condemned replicas from receiving start-phase nodes.
- **Wait node deduplication**: shared `OpWaitCondition` nodes correctly merge prerequisites and propagate `Required` upgrades.
- **Short-circuit safety**: the `waitless` condition in `startPhaseDependencies` correctly short-circuits on `err != nil` before accessing `depService` fields.
- **Replica start ordering**: numeric sort via `cmp.Compare(a.number, b.number)` is correct and lexicography-safe.

pkg/compose/reconcile.go
```diff
@@ -877,10 +929,328 @@ func (r *reconciler) hasVolumeMismatch(expected types.ServiceConfig, oc Observed
...
+// the format lives in exactly one place (nothing parses it back — plan
+// bookkeeping is keyed by service and number, not by resource ID).
+func serviceReplicaID(service string, number int) string {
+	return fmt.Sprintf("service:%s:%d", service, number)
+}
+
+// startGroupID is the event-group key of one replica's start chain.
+func startGroupID(resID string) string {
+	return "start" + strings.TrimPrefix(resID, "service")
+}
```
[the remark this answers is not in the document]
ndeloof: Done in 2d84811c1: `"start:" + strings.TrimPrefix(resID, "service:")` — the separator is now explicit on both sides.

pkg/compose/reconcile.go
```diff
@@ -46,7 +47,21 @@ func toReconcileOptions(options api.CreateOptions) ReconcileOptions {
 }
 
 // ReconcileOptions controls how the reconciler compares desired and observed state.
+// ReconcileScope selects which lifecycle phases the plan covers. The zero
+// value plans the Create phase only — the historical behavior, and what
+// `compose create` keeps using. ScopeCreateStart adds the Start phase to the
+// same plan; ScopeStart plans starting the observed containers without
```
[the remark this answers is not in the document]
ndeloof: Done in 543903271: the stray first line now sits above `type ReconcileOptions struct` as its own doc comment, and `ReconcileScope` keeps only its own.

pkg/compose/reconcile_start_test.go
```diff
@@ -0,0 +1,511 @@
...
+	name := "myproject-" + service + "-" + strconv.Itoa(number)
+	return ObservedContainer{
+		ID:         name + "-id",
+		Name:       name,
+		State:      state,
+		ConfigHash: hash,
+		Number:     number,
+		Summary:    container.Summary{ID: name + "-id", Names: []string{"/" + name}, State: state},
+	}
+}
```
docker-agent, line 68: **[low] `observedServiceContainer` helper omits required Compose labels from `Summary.Labels`**

The new test helper populates `ObservedContainer.Number` (correct) but leaves `Summary.Labels` empty. Every existing helper in `reconcile_test.go` populates `api.ContainerNumberLabel`, `api.ServiceLabel`, and `api.ConfigHashLabel` in `Summary.Labels`. When `reconcileService` calls `nextContainerNumber`, it reads `api.ContainerNumberLabel` from `c.Labels`; a missing entry triggers a `logrus.Warnf` and treats the container number as 0, computing `maxNumber+1=1` regardless of the actual replica number. The result is spurious warning noise in every test that combines this helper with a scale-up path, and an incorrect `nextContainerNumber` result for those paths.

Suggested fix:

```suggestion
func observedServiceContainer(service string, number int, state container.ContainerState, hash string) ObservedContainer {
	name := "myproject-" + service + "-" + strconv.Itoa(number)
	return ObservedContainer{
		ID:         name + "-id",
		Name:       name,
		State:      state,
		ConfigHash: hash,
		Number:     number,
		Summary: container.Summary{
			ID:    name + "-id",
			Names: []string{"/" + name},
			State: state,
			Labels: map[string]string{
				api.ServiceLabel:         service,
				api.ContainerNumberLabel: strconv.Itoa(number),
				api.ConfigHashLabel:      hash,
			},
		},
	}
}
```

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 97/100 |

pkg/compose/reconcile.go
```diff
@@ -877,10 +929,328 @@ func (r *reconciler) hasVolumeMismatch(expected types.ServiceConfig, oc Observed
...
+		group := startGroupID(rep.resID)
+		op := Operation{
+			Type:         OpStartContainer,
+			ResourceID:   rep.resID,
+			Cause:        "start",
+			Service:      &serviceCopy,
+			Container:    rep.container,
+			CreateNodeID: rep.createNodeID,
+		}
+		deps := slices.Clone(prev)
```
docker-agent, line 1226: **[low] First replica's `OpStartContainer` has a redundant direct dependency on its create node when `pre_start` fires**

When `pre_start` is emitted, `first.after` (the create node for replica 1) is added to `preStart`'s dependencies. Then in the replica loop's first iteration, `rep.after` — the same node — is appended directly to the start node's `deps` as well. The start node therefore lists the create node as both a direct dependency and a transitive one (via `preStart`), producing `[1,3] -> #4` in `TestPlanStart_HooksAndReplicaChain` where #1 is already reachable through #3.

The DAG remains topologically valid and execution order is unaffected. The redundancy is visible in `plan.String()` output and in any tooling that enumerates direct edges, which may be confusing when displaying or diffing plan outputs. The second replica and beyond are not affected.

If the intent is that replica 1's start should express its dependency purely through `preStart`, the `if rep.after != nil` guard in the replica loop should be skipped for the first replica when `preStart` was emitted (since `preStart` already carries `first.after`). If the redundant edge is deliberate (e.g., to keep the loop uniform), a comment saying so would prevent future confusion.

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 57/100 |

docker-agent (commented):
### Assessment: 🟢 APPROVE

The start-phase planning logic is well-structured and faithfully reproduces the imperative engine's semantics. The new `planStartPhase` / `planServiceStart` code correctly handles:
- Dependency conditions (service_started as plain edges, others as deduplicated OpWaitCondition nodes)
- Replica ordering (numeric sort with `cmp.Compare`, tested with 11 replicas)
- Scale-down avoidance (`removedByPlan` correctly gates condemned replicas)
- Exceptional-state containers (counted as running for pre_start gating, excluded from start-phase replicas)
- Stopped-by-plan containers (network recreate path, ordered after reconnect)
- Scope guard (inert by default, CreateStart/Start scopes opt in)

Two minor points are noted below.

docker-agent (commented):
### Assessment: 🟢 APPROVE

The start-phase plan vocabulary is well-structured and carefully guarded behind `ReconcileScope` — no existing command passes a non-zero scope, so today's behavior is byte-identical (verified by `TestPlanStart_DefaultScopeIsInert`). The implementation faithfully reproduces every imperative-engine subtlety: dependency-condition re-evaluation even when everything is running, exceptional-state replicas counting as running to gate `pre_start`, scale-down victims excluded from the start phase, and numeric replica ordering to avoid lexicographic bugs with 10+ replicas.

The new test file covers all the critical invariants — paused containers, recreate chains, network-stopped-then-restarted containers, optional vs. required conditions, and the error path for missing containers under `ScopeStart`.

pkg/compose/reconcile_start_test.go
ndeloof: Done in cdb5ca92f: the helper now mirrors the typed fields into Summary.Labels (service, container-number, config-hash), consistent with the other reconcile_test.go helpers.

pkg/compose/reconcile.go
ndeloof: Done in cdb5ca92f: when a pre_start node was planned it already carries the first replica's create node, so the replica loop now skips the redundant direct edge for that replica (`[3] -> #4` in the golden). Later replicas keep their own edge.

docker-agent (commented):
### Assessment: 🟢 APPROVE

This PR carefully implements start-phase planning for the Compose plan engine. The implementation is faithful to the imperative engine's semantics, well-documented, and thoroughly tested.

**What was reviewed:**
- `planStartPhase` / `planServiceStart`: dependency ordering, replica chaining, pre/post-start hook sequencing
- `startPhaseReplicas`: all container-state branches (running, exceptional-state, stopped-by-plan, scale-down condemned), numeric sort via `cmp.Compare`
- `startPhaseDependencies` / `waitConditionNode`: condition deduplication, BestEffort upgrade, target merging
- `plannedReplica`: type switch correctness, unresolved-target default
- All 12 new test cases: logic verified against the code

**Notable strengths:**
- The nested `containerNodes map[string]map[int]*PlanNode` index (replacing the previous flat-map approach) eliminates the O(services × replicas) scan
- `slices.Clone(prev)` before the pre_start block correctly prevents backing-array aliasing
- `cmp.Compare(a.number, b.number)` for replica ordering is correct and overflow-safe
- All previously identified issues from earlier review cycles have been resolved

pkg/compose/reconcile.go
```diff
@@ -877,10 +929,331 @@ func (r *reconciler) hasVolumeMismatch(expected types.ServiceConfig, oc Observed
...
+			// disabled, scale-0 or provider dependencies
+			err != nil || depService.GetScale() == 0 || depService.Provider != nil
+		if waitless {
+			depNodes = append(depNodes, targets...)
+			continue
+		}
+		depNodes = append(depNodes, r.waitConditionNode(dep, cfg, targets))
+	}
+	return depNodes
+}
```
glours: `startPhaseDependencies` reimplements `shouldWaitForDependency` (`service_containers.go:314`) instead of calling it, and drops its distinction between a disabled dependency (`(false, nil)`) and a genuinely missing one (`(false, err)`, propagated as a hard error). Here `err != nil` is just folded into "nothing to wait for". Reachable via `compose start` without a compose file (`projectFromName` never populates `DisabledServices`), and untested. Breaks the PR's own "error cases produce today's messages word for word" promise.

```suggestion
func (r *reconciler) startPhaseDependencies(service types.ServiceConfig) ([]*PlanNode, error) {
      var depNodes []*PlanNode
      for _, dep := range sortedKeys(service.DependsOn) {
              cfg := service.DependsOn[dep]
              targets := r.startChainEnds[dep]
              if len(targets) == 0 {
                      if node, ok := r.serviceNodes[dep]; ok {
                              targets = []*PlanNode{node}
                      }
              }
              shouldWait, err := shouldWaitForDependency(dep, cfg, r.project)
              if err != nil {
                      return nil, err
              }
              if !shouldWait {
                      depNodes = append(depNodes, targets...)
                      continue
              }
              depNodes = append(depNodes, r.waitConditionNode(dep, cfg, targets))
      }
      return depNodes, nil
}
```

pkg/compose/reconcile.go
```diff
@@ -877,10 +929,331 @@ func (r *reconciler) hasVolumeMismatch(expected types.ServiceConfig, oc Observed
...
+		// imperative parity (startService): a scale>0 service with no
+		// container at all cannot be started — only reachable under scope
+		// Start, since CreateStart would have planned the missing creates
+		return fmt.Errorf("service %q has no container to start", service.Name)
+	}
+	// dependency conditions are evaluated even when nothing has to start:
+	// the imperative engine calls waitDependencies for every visited service
+	// before looking at what to start, so an up with everything running
+	// still fails on an unhealthy required dependency
+	depNodes := r.startPhaseDependencies(service)
```
glours: `planServiceStart` already returns `error`, so this is mechanical.

```suggestion
	depNodes, err := r.startPhaseDependencies(service)
      if err != nil {
              return err
      }
```

pkg/compose/reconcile.go
```diff
@@ -877,10 +929,331 @@ func (r *reconciler) hasVolumeMismatch(expected types.ServiceConfig, oc Observed
...
+
+func (r *reconciler) planServiceStart(service types.ServiceConfig) error {
+	if service.Provider != nil {
+		// a provider has no container to start: its create-phase RunProvider
+		// node is what dependents hook onto
+		if node, ok := r.serviceNodes[service.Name]; ok {
+			r.startChainEnds[service.Name] = []*PlanNode{node}
+		}
+		return nil
+	}
```
glours: `startService` (imperative) calls `waitDependencies` unconditionally for every service, providers included (`service_containers.go:582`). Here the provider branch returns before `startPhaseDependencies` is ever evaluated — a provider with its own `depends_on` would wait today but silently wouldn't under the plan.


```suggestion
func (r *reconciler) planServiceStart(service types.ServiceConfig) error {
      if service.Provider != nil {
              // a provider has no container to start, but its own depends_on still
              // applies: the imperative startService waits on it before deferring
              // to the provider mechanism
              depNodes, err := r.startPhaseDependencies(service)
              if err != nil {
                      return err
              }
              ends := depNodes
              if len(ends) == 0 {
                      if node, ok := r.serviceNodes[service.Name]; ok {
                              ends = []*PlanNode{node}
                      }
              }
              if len(ends) > 0 {
                      r.startChainEnds[service.Name] = ends
              }
              return nil
      }
```

pkg/compose/reconcile.go
```diff
@@ -877,10 +929,331 @@ func (r *reconciler) hasVolumeMismatch(expected types.ServiceConfig, oc Observed
...
+	if service.GetScale() == 0 {
+		return nil
+	}
+
+	replicas, anyRunning := r.startPhaseReplicas(service)
+	if len(replicas) == 0 && !anyRunning && r.options.Scope == ScopeStart {
+		// imperative parity (startService): a scale>0 service with no
+		// container at all cannot be started — only reachable under scope
+		// Start, since CreateStart would have planned the missing creates
+		return fmt.Errorf("service %q has no container to start", service.Name)
```
glours: Nit: This exact string is also produced by `startService` in `service_containers.go:591`. Dormant today since `ScopeStart` isn't wired to any command yet, but given this PR's explicit word-for-word fidelity contract with the imperative engine, it'd be worth factoring into one shared helper so the two can't drift apart later

pkg/compose/reconcile.go
```diff
@@ -77,6 +92,26 @@ type reconciler struct {
...
+	// visit: the last node of the replica chain, or — when nothing had to
+	// start — the dependency prerequisites the visit still evaluated. What a
+	// service_started dependent (or a wait node) hooks onto, matching
+	// InDependencyOrder semantics: a dependent's visit begins only once the
+	// dependency's whole visit (waits included) completed.
+	startChainEnds map[string][]*PlanNode
+	// waitNodes deduplicates OpWaitCondition nodes per (service, condition):
+	// several dependents awaiting the same condition share one node, like
+	// networkNodes deduplicates network creations.
+	waitNodes map[string]*PlanNode
```
glours, line 117: `reconcileContainers` and `planStartPhase` each build an identical `NewGraph(r.project, ServiceStopped)` (including its `HasCycles` DFS) within the same `reconcile()` call under `ScopeCreateStart`. Safe to cache: `visitInDependencyOrder` never mutates the `*Graph` (its `visited` set is local to each call, and `Children`/`Key`/`Service` are fixed once by `NewGraph`), and nothing between the two call sites reassigns `r.project` or its contents — only `r.observed`/`r.resolvedNetworks`/`r.resolvedVolumes` change in between.

```suggestion
	waitNodes map[string]*PlanNode
	// dependencyGraph memoizes NewGraph(project, ServiceStopped): both
	// reconcileContainers and planStartPhase need the same dependency-ordered
	// walk within one reconcile() call. Safe to share — visitInDependencyOrder
	// tracks visited state locally per call and never mutates the graph, and
	// r.project is never reassigned between the two calls.
	dependencyGraph *Graph
```

pkg/compose/reconcile.go
```diff
@@ -593,12 +640,12 @@ func (r *reconciler) reconcileContainers() error {
 	}
 
 	// Visit in dependency order (leaves first = services with no deps)
-	return r.visitInDependencyOrder(graph)
+	return r.visitInDependencyOrder(graph, r.reconcileService)
 }
```
glours, line 656:
```suggestion
// reconcileContainers processes each service in dependency order, comparing
// the desired scale and configuration with observed containers.
func (r *reconciler) reconcileContainers() error {
      graph, err := r.serviceGraph()
      if err != nil {
              return err
      }

      // Visit in dependency order (leaves first = services with no deps)
      return r.visitInDependencyOrder(graph, r.reconcileService)
}

// serviceGraph memoizes NewGraph(r.project, ServiceStopped): reconcileContainers
// and planStartPhase both need it within one reconcile() call.
func (r *reconciler) serviceGraph() (*Graph, error) {
      if r.dependencyGraph == nil {
              graph, err := NewGraph(r.project, ServiceStopped)
              if err != nil {
                      return nil, err
              }
              r.dependencyGraph = graph
      }
      return r.dependencyGraph, nil
}
```

pkg/compose/reconcile.go
```diff
@@ -877,10 +929,331 @@ func (r *reconciler) hasVolumeMismatch(expected types.ServiceConfig, oc Observed
...
+// deliberately re-observed at execution time — the plan only encodes what to
+// wait for, never a stale observation. service_started needs no node: a
+// plain DAG edge to the dependency's chain end expresses it.
+func (r *reconciler) planStartPhase() error {
+	graph, err := NewGraph(r.project, ServiceStopped)
+	if err != nil {
+		return err
+	}
+	return r.visitInDependencyOrder(graph, r.planServiceStart)
+}
```
glours, line 977:
```suggestion
func (r *reconciler) planStartPhase() error {
	graph, err := r.serviceGraph()
	if err != nil {
		return err
	}
	return r.visitInDependencyOrder(graph, r.planServiceStart)
}
```

glours (changes requested):
One blocker found: `startPhaseDependencies` (see inline) silently swallows an error the imperative engine treats as fatal, reachable via `compose start` without a compose file, contradicts the PR's own "error cases word for word" claim. 

Three should-fix items alongside it (provider services skip dependency waits, a duplicated error string, a graph built twice per reconcile). 
Two minor nits not called out inline: the `Scope` gating in `reconcile()` (~L173/L191) reads as double-negatives across a 3-value enum, a switch or two named helpers would be clearer, and `waitConditionNode`'s merge loop is confirmed dead code today (comment is accurate, no action needed).

pkg/compose/reconcile.go
ndeloof: Fixed in 911bf7d86, exactly as suggested: startPhaseDependencies now delegates to shouldWaitForDependency and propagates its error — the missing-vs-disabled distinction is the imperative engine's own. Covered by the new TestPlanStart_MissingDependencyFails (scope Start, depends_on naming an undeclared service, no DisabledServices).

pkg/compose/reconcile.go
ndeloof: Applied in 911bf7d86.

pkg/compose/reconcile.go
ndeloof: Fixed in 911bf7d86: the provider branch now evaluates startPhaseDependencies and hooks its chain end onto the prerequisites (falling back to the RunProvider node), per your suggestion — extracted as planProviderStart to keep planServiceStart under the gocognit budget. Covered by the new TestPlanStart_ProviderWaitsOnDependencies.

pkg/compose/reconcile.go
ndeloof: Done in 911bf7d86: both sites now share errNoContainerToStart (defined next to startService, the original owner).

pkg/compose/reconcile.go
ndeloof: Applied in 911bf7d86 (field + comment as suggested).

pkg/compose/reconcile.go
ndeloof: Applied in 911bf7d86.

pkg/compose/reconcile.go
ndeloof: Applied in 911bf7d86.

ndeloof:
All review items addressed in 911bf7d86 (single commit amended):
- **Blocker**: `startPhaseDependencies` delegates to `shouldWaitForDependency`, error propagated — with a regression test (`TestPlanStart_MissingDependencyFails`).
- Provider services now evaluate their own `depends_on` (`planProviderStart`), tested (`TestPlanStart_ProviderWaitsOnDependencies`).
- `errNoContainerToStart` shared between `startService` and the reconciler.
- `serviceGraph()` memoizes the dependency graph for both walks.
- The `Scope` gating nit from the review body: `reconcile()` now reads through positive helpers `plansCreatePhase()`/`plansStartPhase()` instead of the `!=` double-negatives.

Unit tests and lint green.

pkg/compose/reconcile.go
```diff
@@ -877,10 +955,352 @@ func (r *reconciler) hasVolumeMismatch(expected types.ServiceConfig, oc Observed
...
+func (r *reconciler) planServiceStart(service types.ServiceConfig) error {
+	if service.Provider != nil {
+		return r.planProviderStart(service)
+	}
+	if service.GetScale() == 0 {
+		return nil
+	}
+
+	replicas, anyRunning := r.startPhaseReplicas(service)
+	if len(replicas) == 0 && !anyRunning && r.options.Scope == ScopeStart {
```
docker-agent: **[medium] `errNoContainerToStart` returned before dependency validation, diverging from imperative parity**

`planServiceStart` short-circuits at line 1198 — returning `errNoContainerToStart` — before it ever calls `startPhaseDependencies`. The imperative counterpart `startService` (in `service_containers.go`) does the opposite: it calls `waitDependencies` **first**, then checks whether there are containers to start. The comment on line 1206 even documents the intent ("dependency conditions are evaluated even when nothing has to start"), but the early return at line 1202 means `startPhaseDependencies` is never reached for the `len(replicas)==0 && !anyRunning && ScopeStart` branch.

**Trigger**: `compose start` when a service has zero containers (`ScopeStart`) AND a `depends_on` dependency whose condition would cause `shouldWaitForDependency` to return an error (e.g. the dependency service is absent from the project after partial load, or removed from `DisabledServices` between calls).

**Impact**: The user sees `service "X" has no container to start` instead of the dependency error that is the actual root cause. This is a messaging-clarity regression relative to imperative parity. In practice, project-load validation makes `shouldWaitForDependency` returning an error rare, so the impact is narrow — but the ordering divergence is a correctness gap in the stated "word for word" fidelity contract.

**Fix**: Move the `errNoContainerToStart` guard to after `startPhaseDependencies` succeeds, matching the imperative `startService` order:

```go
replicas, anyRunning := r.startPhaseReplicas(service)
depNodes, err := r.startPhaseDependencies(service)
if err != nil {
    return err
}
if len(replicas) == 0 && !anyRunning && r.options.Scope == ScopeStart {
    return errNoContainerToStart(service.Name)
}
```

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 100/100 |

docker-agent (commented):

ndeloof:
Rebased onto current main: the base had moved under the PR and CI was building the merge against a newer compose-go where `DependsOn`/`HealthCheck` became promoted fields (`WorkloadSpec`, illegal in struct literals before go1.27) and `PreStart` changed to `[]PreStartHook` — the lint/test failures were compile errors in the merged tree, invisible to any local check against the branch's own go.mod. Test fixtures adapted (assignment style via a small `serviceWithDeps` helper, `PreStartHook` for pre_start). No functional change; review fixes from the previous push are unchanged.

pkg/compose/reconcile.go
ndeloof: Fixed in 6fa3f0614: startPhaseDependencies is now evaluated before the errNoContainerToStart guard, matching startService's order (waitDependencies first) — a dependency error wins over "no container to start". The regression test covers both orderings (TestPlanStart_MissingDependencyFails: dependency error with an observed container, and with no container at all under scope Start).

glours (approved):
Sounds good to me 👍

docker-agent (commented):
### Assessment: 🟢 APPROVE

PR #14201, opened 2026-09-09
plan the purge of stale pre_start hook runners

ndeloof:
### What this PR does, in one sentence

The reconciliation plan now covers the purge of stale `pre_start` hook runners, and observed state stops mistaking those runners for service replicas.

### Context

`pre_start` hooks run in ephemeral containers. When a hook fails, its runner is deliberately retained so the operator can inspect it, and the next run purges it — but that purge lives entirely inside the imperative primitive, invisible to the plan. Worse, observed-state collection had no notion of hook containers at all: carrying no container-number label, a leftover runner was classified as a service replica numbered 0.

### What the PR brings

- Observed state classifies hook containers into a dedicated `HookContainers` bucket — they no longer masquerade as replica 0.
- When `pre_start` is going to run again (hooks declared, no replica running at observation — the exact imperative gating), the plan emits one best-effort `RemoveContainer` per stale runner, dropping its anonymous volumes. Failure is warn-only and a running service keeps its retained-for-inspection runner untouched, exactly like today; the imperative purge remains in place as backstop, so no behavior is lost in the cases the plan cannot see.
- The imperative primitive is split into a lifecycle-free execution piece (`execPreStartHook`: start, wait, logs, retain-on-failure) and the create/remove pieces around it, recomposed identically in `runPreStart` — locked by the existing characterization tests.

### Why this is the right next brick

Once the start phase lands in the plan (#14200), hook-container **creation** and **post-success removal** can join it as plan nodes chained around a run-only `RunPreStart` — this PR puts the classification, the purge, and the split primitive in place without depending on it.

### DIFF
--- a/pkg/compose/observed_state.go
+++ b/pkg/compose/observed_state.go
@@ -46,6 +46,11 @@ type ObservedState struct {
 	// others as orphans (see selectNetwork/selectVolume).
 	Networks map[string][]ObservedNetwork // compose network key → observed
 	Volumes  map[string][]ObservedVolume  // compose volume key → observed
+	// HookContainers are ephemeral lifecycle-hook runners (HookLabel set),
+	// per service. Any observed at collection time is stale by definition —
+	// a previous run failed before removing it — and the reconciler plans
+	// its purge before re-running the hooks.
+	HookContainers map[string][]ObservedContainer // service name → hook containers
 }
 
 // selectNetwork picks, among the live networks recorded for a compose key, the
@@ -148,6 +153,8 @@ func (s *composeService) collectObservedState(ctx context.Context, project *type
 		Containers:  map[string][]ObservedContainer{},
 		Networks:    map[string][]ObservedNetwork{},
 		Volumes:     map[string][]ObservedVolume{},
+
+		HookContainers: map[string][]ObservedContainer{},
 	}
 
 	// --- Containers ---
@@ -170,6 +177,18 @@ func (s *composeService) collectObservedState(ctx context.Context, project *type
 
 	for _, ctr := range raw {
 		svcName := ctr.Labels[api.ServiceLabel]
+		if ctr.Labels[api.HookLabel] != "" && knownServices[svcName] {
+			// lifecycle-hook containers (ephemeral pre_start runners) are
+			// neither service replicas nor one-offs: classified apart, so
+			// they never masquerade as a replica (they carry no
+			// container-number label and would otherwise read as number 0)
+			// and the reconciler can plan purging stale ones. A hook
+			// container whose service left the model falls through to the
+			// orphan check below instead — nothing plans purges for an
+			// unknown service, and --remove-orphans must keep cleaning it.
+			state.HookContainers[svcName] = append(state.HookContainers[svcName], toObservedContainer(ctr))
+			continue
+		}
 		if isNotOneOff(ctr) && knownServices[svcName] {
 			state.Containers[svcName] = append(state.Containers[svcName], toObservedContainer(ctr))
 		} else if isOrphaned(project)(ctr) {
--- a/pkg/compose/reconcile.go
+++ b/pkg/compose/reconcile.go
@@ -662,6 +662,8 @@ func (r *reconciler) reconcileService(service types.ServiceConfig) error {
 		return err
 	}
 
+	r.planPurgeStaleHookRunners(service, expected)
+
 	containers := r.observed.Containers[service.Name]
 	actual := len(containers)
 
@@ -760,6 +762,41 @@ func (r *reconciler) reconcileService(service types.ServiceConfig) error {
 
 // mustRecreate decides whether oc must be recreated to match expected. The
 // expectedHash and parentRecreated inputs are precomputed once per service by
+// planPurgeStaleHookRunners plans the removal of hook-runner containers left
+// behind by a previous run that failed before removing them. It mirrors the
+// imperative purge living inside the gated runPreStart call: emitted only when
+// pre_start is going to run again — hooks declared, a replica to start
+// (scale > 0: the imperative start path returns before the hooks for a
+// scale-0 service) and no replica running at observation — so a genuinely
+// failed hook container stays retained for inspection as long as its service
+// is otherwise up. Removals are best-effort (the imperative purge is
+// warn-only) and independent of every other node.
+func (r *reconciler) planPurgeStaleHookRunners(service types.ServiceConfig, expectedScale int) {
+	stale := r.observed.HookContainers[service.Name]
+	if len(stale) == 0 || len(service.PreStart) == 0 || expectedScale == 0 {
+		return
+	}
+	for _, oc := range r.observed.Containers[service.Name] {
+		if oc.State == container.StateRunning {
+			return
+		}
+	}
+	serviceCopy := service
+	stale = slices.Clone(stale)
+	slices.SortFunc(stale, func(a, b ObservedContainer) int { return strings.Compare(a.ID, b.ID) })
+	for i := range stale {
+		r.plan.addNode(Operation{
+			Type:          OpRemoveContainer,
+			ResourceID:    fmt.Sprintf("hook:%s:stale:%s", service.Name, stale[i].ID[:min(12, len(stale[i].ID))]),
+			Cause:         "stale pre_start hook container",
+			Service:       &serviceCopy,
+			Container:     &stale[i].Summary,
+			RemoveVolumes: true,
+			BestEffort:    true,
+		}, "")
+	}
+}
+
 // reconcileService — see expectedConfigHash and parentNamespaceRecreated for
 // the rationale (issue #13878).
 func (r *reconciler) mustRecreate(expected types.ServiceConfig, expectedHash string, parentRecreated bool, oc ObservedContainer, policy string) bool {
--- a/pkg/compose/executor_ops.go
+++ b/pkg/compose/executor_ops.go
@@ -142,8 +142,15 @@ func (exec *planExecutor) execStopContainer(ctx context.Context, op Operation) e
 }
 
 func (exec *planExecutor) execRemoveContainer(ctx context.Context, op Operation) error {
-	_, err := exec.compose.apiClient().ContainerRemove(ctx, op.Container.ID, client.ContainerRemoveOptions{Force: true})
+	_, err := exec.compose.apiClient().ContainerRemove(ctx, op.Container.ID, client.ContainerRemoveOptions{Force: true, RemoveVolumes: op.RemoveVolumes})
 	if err != nil {
+		if op.BestEffort {
+			// warn-only removal (stale pre_start hook runner): the container
+			// stays visible to the operator, the plan carries on — and the
+			// live view below keeps it, since it was not removed
+			logrus.Warnf("failed to remove %s: %v", op.ResourceID, err)
+			return nil
+		}
 		return err
 	}
 	// Why: a dependent service's create may resolve `network_mode: service:X`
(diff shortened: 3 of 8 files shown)


pkg/compose/observed_state.go
```diff
@@ -170,6 +177,15 @@ func (s *composeService) collectObservedState(ctx context.Context, project *type
 
 	for _, ctr := range raw {
 		svcName := ctr.Labels[api.ServiceLabel]
+		if ctr.Labels[api.HookLabel] != "" {
```
docker-agent: **[medium] Hook containers for removed services bypass orphan detection and are never cleaned by `up --remove-orphans`**

The new `HookLabel` branch unconditionally routes every hook container into `HookContainers[svcName]` and `continue`s, skipping the `isOrphaned` check for all of them — including those whose service has been removed from `compose.yaml`.

**Trigger path**: A service with `pre_start` hooks runs, the hook container is left behind (hook failure or crash), then the service is deleted from `compose.yaml`. On the next `compose up --remove-orphans`:

1. The container has `HookLabel != ""` → goes into `HookContainers["deleted_svc"]` → `continue`.
2. `isOrphaned` never sees it → `state.Orphans` stays empty for it.
3. `planPreStart` only iterates project services → `HookContainers["deleted_svc"]` is never read.
4. `reconcileOrphans` only walks `state.Orphans` → the container leaks.

**Before this PR**: No `HookLabel` branch existed; the container would fall through to `isOrphaned` (service absent from project → returns `true`) → `state.Orphans` → cleaned by `reconcileOrphans`.

**Fix**: Guard the `HookContainers` route with `knownServices[svcName]`; for unknown-service hook containers, fall through to `isOrphaned` so the orphan removal path still applies:

```go
if ctr.Labels[api.HookLabel] != "" {
    if knownServices[svcName] {
        state.HookContainers[svcName] = append(state.HookContainers[svcName], toObservedContainer(ctr))
    } else if isOrphaned(project)(ctr) {
        state.Orphans = append(state.Orphans, toObservedContainer(ctr))
    }
    continue
}
```

The existing test (`observed_state_test.go`) only exercises a hook container for a known service — a test covering the removed-service case would lock in the corrected behavior.

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 97/100 |

docker-agent (commented):

pkg/compose/observed_state.go
docker-agent: **[medium] Hook containers from deleted services bypass orphan tracking and won't be cleaned by `--remove-orphans`**

The new `if ctr.Labels[api.HookLabel] != ""` block fires early with `continue`, bypassing the `isOrphaned` check below it. Before this PR, a hook container from a service that was subsequently deleted from the compose file would have taken this path:

1. `isNotOneOff(ctr)` → `true` (hook containers have no `OneoffLabel`)
2. `knownServices[svcName]` → `false` (service removed)
3. `isOrphaned(project)(ctr)` → `true` (service name not in `project.ServiceNames()`)
4. → classified as `Orphans`, cleaned by `compose up --remove-orphans`

After this PR, step 3 is never reached: the container lands in `HookContainers[deletedServiceName]` and `planPurgeStaleHookRunners` never visits it (the deleted service is not in the project, so the function is never called for it).

**Trigger**: User removes a service with `pre_start` hooks from the compose file (or renames it), leaving a stale hook runner behind. Then runs `compose up --remove-orphans` instead of `compose down` first.

**Impact**: The stale hook runner accumulates indefinitely in `compose up` runs. `compose down` still cleans it (via `removePreStartHookContainers`'s direct label query), but `--remove-orphans` no longer does.

**Fix**: Check `knownServices[svcName]` before classifying into `HookContainers`; if the service is unknown, fall through to the orphan check:
```suggestion
		if ctr.Labels[api.HookLabel] != "" && knownServices[svcName] {
```

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 100/100 |

docker-agent (commented):

One medium-severity correctness finding confirmed (strong confidence, 100/100).

#### Summary

The PR correctly classifies hook containers into a dedicated `HookContainers` bucket so they no longer masquerade as replica-0, and the plan-level purge of stale hook runners is well-gated and best-effort. The refactoring of `runPreStartHook` into `execPreStartHook` + caller-managed lifecycle is clean.

One issue was found: the early `continue` in `collectObservedState` bypasses the orphan check for hook containers, so stale hook runners from **deleted services** escape `compose up --remove-orphans` cleanup.

pkg/compose/observed_state.go
ndeloof: Done in 6682e7003: the HookContainers route is now guarded by knownServices[svcName]; a hook runner whose service left the model falls through to the orphan check, so --remove-orphans keeps cleaning it. TestCollectObservedState locks the case (c6: hook runner of the model-absent service classified as orphan, HookContainers empty for it).

ndeloof: Same finding as the later review's thread — fixed in 6682e7003 (knownServices guard + fall-through to the orphan check, with a test on the removed-service case).

pkg/compose/reconcile.go
```diff
@@ -760,6 +762,39 @@ func (r *reconciler) reconcileService(service types.ServiceConfig) error {
...
+// planPurgeStaleHookRunners plans the removal of hook-runner containers left
+// behind by a previous run that failed before removing them. It mirrors the
+// imperative purge living inside the gated runPreStart call: emitted only when
+// pre_start is going to run again — hooks declared and no replica running at
+// observation — so a genuinely failed hook container stays retained for
+// inspection as long as its service is otherwise up. Removals are best-effort
+// (the imperative purge is warn-only) and independent of every other node.
+func (r *reconciler) planPurgeStaleHookRunners(service types.ServiceConfig) {
+	stale := r.observed.HookContainers[service.Name]
+	if len(stale) == 0 || len(service.PreStart) == 0 {
```
docker-agent: **[medium] `planPurgeStaleHookRunners` purges stale hook runners even when service `scale=0`, contrary to the stated contract**

The function's guard checks for any _observed_ running replica but does not check whether the _desired_ scale is greater than zero. For a service with `scale: 0` (or `Deploy.Replicas == 0`) that has stale hook containers from a previous failed run, the guard loop over `r.observed.Containers[service.Name]` never fires (the slice is empty), so the function proceeds to emit `RemoveContainer` nodes.

The imperative path in `startService` (service_containers.go) has an explicit early-return on `scale == 0` that prevents `runPreStart` from ever being called. The reconcile-side guard is therefore weaker than what the comment claims:

> "emitted only when pre_start is going to run again … — the exact imperative gating"

**Operator-visible impact**: An operator who scales a service to zero specifically to preserve a retained-for-inspection hook container would see it purged on the next `compose up` reconciliation, contrary to the PR's own guarantee that "a genuinely failed hook container stays retained for inspection as long as its service is otherwise up."

**Suggested fix**: add `expected` (already computed in `reconcileService` before the call) as a parameter and gate on `expected > 0`, or move the call below the scale check:

```go
if expected == 0 {
    return nil  // nothing to purge — hooks will not run
}
r.planPurgeStaleHookRunners(service)
```

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 57/100 |

pkg/compose/observed_state.go
```diff
@@ -170,6 +177,18 @@ func (s *composeService) collectObservedState(ctx context.Context, project *type
 
 	for _, ctr := range raw {
 		svcName := ctr.Labels[api.ServiceLabel]
+		if ctr.Labels[api.HookLabel] != "" && knownServices[svcName] {
```
docker-agent, line 180: **[medium] Hook containers of disabled services are siloed into `HookContainers` with no cleanup path**

The new classification block guards on `knownServices[svcName]`, and `knownServices` is populated for **both** active AND disabled services:

```go
for _, ds := range project.DisabledServices {
    knownServices[ds.Name] = true   // ← disabled services included
}
```

So a stale pre_start hook container whose `ServiceLabel` refers to a disabled service hits the new `continue` and lands in `HookContainers[disabled-svc]` — never reaching the orphan check below.

`planPurgeStaleHookRunners` is called only from `reconcileService`, which is driven by `visitInDependencyOrder` over `NewGraph(project.Services)` — disabled services are **not** in that graph. As a result, `HookContainers["disabled-svc"]` is filled at collection time and never read again.

**Result**: stale hook containers of disabled services have no cleanup path:
- Not an orphan → `--remove-orphans` ignores them.
- Not processed by the plan → `planPurgeStaleHookRunners` never fires for them.

They accumulate indefinitely until manually removed.

**Fix**: use a separate `activeServices` set (active services only, not disabled ones) for the hook-container gate, so disabled-service hook containers still reach the orphan check:

```go
_, isActive := project.Services[svcName]
if ctr.Labels[api.HookLabel] != "" && isActive {
```

Or inline:

```suggestion
		if ctr.Labels[api.HookLabel] != "" && project.Services[svcName].Name != "" {
```

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 97/100 |

docker-agent (commented):

Two findings introduced by this PR's `+` lines:

**Inline comments (2)**:
1. `pkg/compose/reconcile.go:774` [medium] — `planPurgeStaleHookRunners` purges stale hook runners even when `scale=0`, contrary to the stated contract (confidence: 🟡 moderate 57/100)
2. `pkg/compose/observed_state.go:180` [medium] — Hook containers of disabled services are siloed into `HookContainers` with no cleanup path (confidence: 🟢 strong 97/100)

pkg/compose/reconcile.go
ndeloof: Done in b52def4df: planPurgeStaleHookRunners now takes the expected scale and returns early when it is 0 — the imperative start path returns before the hooks for a scale-0 service, so its stale runners stay retained. TestReconcileContainers_StaleHookRunnersKeptWhenScaleZero locks the case.

pkg/compose/observed_state.go
ndeloof: Deliberately kept as-is — the accumulation predates this PR rather than being introduced by it: before this change, a disabled service's hook runner sat in Containers[svc] as a bogus replica-0 entry (knownServices includes disabled services), never visited by reconcileService and never orphaned, so --remove-orphans did not clean it either. Routing it to the orphan check would be a behavior CHANGE: --remove-orphans would delete a retained-for-inspection runner while its service is merely profile-disabled — and while the service's real replicas are kept. The PR actually improves the cleanup path: once the service is re-enabled, reconcileService visits it and the plan purges the stale runner (as the imperative purge would). `down` keeps cleaning them via its label query. Leaving this thread open for the maintainer to weigh in.

docker-agent (commented):
### Assessment: 🟢 APPROVE

The PR is well-structured and the changes are correct.

**Summary of changes reviewed:**
- `observed_state.go`: Hook containers (identified by `HookLabel`) are now correctly classified into `HookContainers` map keyed by service name, preventing stale runners from masquerading as replica 0. The guard `knownServices[svcName]` ensures containers for services removed from the model fall through to the orphan path as expected.
- `reconcile.go`: `planPurgeStaleHookRunners` correctly mirrors the imperative gate (`len(service.PreStart) > 0 && expectedScale > 0 && no running replica`). The best-effort/warn-only semantics and the `RemoveVolumes: true` flag are both correct.
- `pre_start.go`: The refactor cleanly separates `execPreStartHook` (start + wait + logs + retain-on-failure) from `runPreStart` (create → exec → remove-on-success). The success-path removal is now correctly owned by the caller, and the failure retention / cancellation removal paths are unchanged.
- `executor_ops.go`: The `BestEffort` check and `RemoveVolumes` forwarding are placed correctly — a warn-only failure returns `nil` so the plan continues, and the live-view comment explains the container stays in the cache.
- Test coverage is thorough: executor, observed-state, and reconciler all have dedicated tests for the new behaviour.

PR #14202, opened 2026-09-09
fix(watch): coalesce rebuilds triggered during a burst of file changes

glours:
**What I did**
compose watch triggered one rebuild per debounced batch, so a burst of edits arriving faster than the debounce window (e.g. from a coding agent editing several files) queued up several rebuilds in a row instead of a single consolidated one.

Rebuilds now run asynchronously through a small scheduler that coalesces requests received while a rebuild is in flight into one trailing rebuild, without adding latency to an isolated edit. This also closes a race where a plain restart could run concurrently with an asynchronous rebuild of the same service.

**Related issue**
N/A

image

### DIFF
--- a/pkg/compose/rebuild_scheduler.go
+++ b/pkg/compose/rebuild_scheduler.go
@@ -1,0 +1,103 @@
+/*
+
+   Copyright 2020 Docker Compose CLI authors
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+       http://www.apache.org/licenses/LICENSE-2.0
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package compose
+
+import (
+	"sync"
+
+	"github.com/docker/compose/v5/pkg/utils"
+)
+
+// rebuildFunc performs a rebuild for the given services.
+type rebuildFunc func(services []string) error
+
+// rebuildScheduler coalesces rebuild requests received while a rebuild is
+// already running into a single trailing rebuild, so that a burst of file
+// change events never queues up more than one extra rebuild.
+type rebuildScheduler struct {
+	rebuild rebuildFunc
+
+	mu       sync.Mutex
+	building bool
+	active   utils.Set[string] // services being rebuilt by the current run, if any
+	pending  utils.Set[string] // services queued for the next trailing rebuild
+	wg       sync.WaitGroup
+}
+
+func newRebuildScheduler(rebuild rebuildFunc) *rebuildScheduler {
+	return &rebuildScheduler{
+		rebuild: rebuild,
+		pending: utils.NewSet[string](),
+	}
+}
+
+// Request asks for a rebuild of services. It never blocks on the rebuild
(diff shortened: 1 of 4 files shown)


ndeloof (commented):
Solid work overall: the scheduler is minimal and correct (decision taken synchronously under the lock, so requests racing a run's completion are neither lost nor doubled), the defer ordering in `watchEvents` guarantees no rebuild goroutine outlives the function, and the test suite is exemplary — deterministic channel-driven fake with the no-reentrancy invariant asserted inside the fake itself, and the `assertNoRebuildStarts` justification is exactly right. `-race` passes locally on the new tests.

One behavioral edge worth an explicit decision before merge, then small things:

**[medium] A `sync+restart` change arriving while a rebuild of the same service is ACTIVE is silently lost.**
Timeline: batch 1 triggers rebuild(A); while the image builds, batch 2 carries a `sync+restart` change for A. The sync lands in the doomed old container (the documented tradeoff), the restart is skipped by `InFlightOrPending`, and the new container comes from a build context snapshotted *before* the change. Net effect: the edit is silently dropped until the user touches the file again. Before this PR the synchronous rebuild made batch 2 converge; and the burst-editing regime this PR targets makes the overlap window common, not exotic.
The skip is fully correct for a **pending** rebuild (it will re-read the context, so the change is included and the restart is redundant). It is only the **active** case that loses data. Suggested fix: distinguish the two — skip on pending as today, and on active either requeue the restart to run after the current rebuild completes, or requeue a trailing `scheduler.Request([service])` (converges via context re-read, at the cost of an extra — usually cache-warm — rebuild in the rare overlap).

**[low]** After Ctrl-C, `run()` still drains the pending set with a cancelled ctx, so the user log gets an extra "Build failed. Error: context canceled". A `ctx.Err()` check before each drain iteration would make shutdown silent.

**[low]** `_ = s.rebuild(services)` is actually fine — `s.rebuild` reports everything to `options.LogTo` itself — but that contract deserves a one-line comment on the discard, or the next reader will read it as a swallowed error.

**[question]** `s.rebuild` (build/create/start) now runs concurrently with the event loop's own uses of the same `*types.Project` (restart, syncer, exec hooks). Worth double-checking none of those paths mutates the project in place — the scheduler serializes rebuilds against each other, but not against the loop.

**[nit]** The comment on `TestHandleWatchBatch_SkipsRestartForServiceWithRebuildInFlight` references "#14051 (follow-up)" — that's the TTY renderer PR; stale reference.
**[nit]** In `TestRebuildScheduler_NoConcurrentRebuilds`, the local `max` shadows the builtin.

ndeloof:
Concrete, tested proposal for the [medium] finding above: glours/compose#124 — targets this PR's branch, so merging it updates this PR. Design in one line: a rebuild made stale by a newer request for one of its own services is interrupted (per-run cancellable context) and its whole active set folds back into pending, upholding the invariant that every request ends up covered by a rebuild whose context snapshot postdates it. Also turns restart-during-in-flight into a Request (converged restart intent) and stops the drain on watch shutdown ([low] Ctrl-C noise). `go test -race` green.

ndeloof (approved):
All findings from my earlier review are addressed at this head — the [medium] sync+restart convergence loss (a rebuild made stale by a newer request is now interrupted, its active set folded back into pending, and a restart landing on an in-flight rebuild folds into a Request), the shutdown-drain noise, and the stale test reference. I validated the merged change locally before it landed here (full pkg/compose suite with -race, lint clean), and CI is green on this head. The scheduler invariant — every request ends up covered by a rebuild whose context snapshot postdates it — is exactly the right contract for watch. LGTM.

PR #14213, opened 2026-09-11
fix(port): preserve tie order for dual-stack port publishers

glours:
**What I did**
Follow-up to #13577, which made `docker compose port SERVICE [PRIVATE_PORT]` list all port mappings when `PRIVATE_PORT` is omitted, by routing through `containerPublishers()` (shared with `docker compose ps`).                                  
                                                                                                                                                                                                                                                        
 `containerPublishers()` sorts by `PrivatePort` using `sort.Slice`, which does not guarantee tie order. That didn't matter for `ps`, but the single-port lookup path added by #13577 now takes the first matching entry as *the* answer, so a dual-stack publish (same target port exposed on both an IPv4 and an IPv6 host address) could non-deterministically return either one.

- Switch `sort.Slice` to `sort.SliceStable` in `containerPublishers()` so ties preserve the daemon's original order.                                                                                                                               
- Add a regression test for that (large enough to force Go's sort past its 12-element insertion-sort fallback, since a smaller fixture can't tell a stable sort from an unstable one).                                                             
- Add table-driven coverage for `Ports()`'s protocol/port filtering and its not-found error path (previously only covered by two e2e assertions).                                                                                                  
- Drop a dead length check and a duplicated host:port formatter left over from #13577's review.   

**Related issue**
N/A

<img width="1500" height="1000" alt="image" src="https://github.com/user-attachments/assets/fc9885ac-a8f2-4f53-99c1-1c919730afdc" />

### DIFF
--- a/cmd/compose/port.go
+++ b/cmd/compose/port.go
@@ -19,7 +19,6 @@ package compose
 import (
 	"context"
 	"fmt"
-	"net"
 	"strconv"
 	"strings"
 
@@ -86,9 +85,8 @@ func runPort(ctx context.Context, dockerCli command.Cli, backendOptions *Backend
 		return err
 	}
 
-	if opts.port != 0 && len(publishers) > 0 {
-		p := publishers[0]
-		_, _ = fmt.Fprintf(dockerCli.Out(), "%s\n", net.JoinHostPort(p.URL, strconv.Itoa(p.PublishedPort)))
+	if opts.port != 0 {
+		_, _ = fmt.Fprintf(dockerCli.Out(), "%s\n", publishers[0].HostPort())
 		return nil
 	}
 	for _, p := range publishers {
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -609,7 +609,12 @@ type PortPublisher struct {
 }
 
 func (p PortPublisher) String() string {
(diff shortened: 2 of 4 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The changes are correct and well-scoped. The switch from `sort.Slice` to `sort.SliceStable` in `containerPublishers()` correctly fixes the non-deterministic tie-breaking for dual-stack port publishers. The removed redundant length guard in `cmd/compose/port.go` is safe because `Ports()` guarantees a non-empty slice when `port != 0` and no error is returned. The regression test correctly validates stable sort behaviour using 16 elements (>12 to force the quicksort path), and the new table-driven tests for `Ports()` provide solid coverage.

ndeloof (approved):
Reviewed in depth: the sort.SliceStable switch is exactly right — sort.Slice's pdqsort scrambles same-PrivatePort ties once past the small-slice insertion-sort path, and the single-port lookup added by #13577 takes publishers[0] as THE answer, so a dual-stack publish could non-deterministically print the ::-side. Verified the dropped len(publishers)>0 guard in cmd/port.go is safe: Ports() already errors on an unmatched port, so publishers[0] is unreachable empty. The tie-order test forcing the quicksort partitioning path (>12 tied elements) is a nice touch — a small fixture genuinely cannot catch this. Tests green locally with -race, CI green. LGTM.

PR #14214, opened 2026-09-11
prompt: swallow escape sequences, both backspaces, exit 130 on Ctrl+C

ndeloof:
**What this PR does, in one sentence:** three edges the survey-removal review (#14161) surfaced in the new interactive confirmation reader — escape sequences, the second backspace encoding, and Ctrl+C's exit status — plus the lost pipe-mode hint.

## Context

#14161 replaced the archived survey/v2 with a small raw-terminal reader, faithfully preserving the existing behavior. Raw mode changes what the reader receives, though: arrow keys arrive as byte sequences (`ESC [ A`) whose printable tail was echoed and taken as answer characters; terminals emitting `^H` for backspace could no longer erase; and Ctrl+C — which raw mode hides from the signal handler — surfaced as a generic error with exit 1, where a real SIGINT exits 130.

## What the PR brings

- **Escape sequences are swallowed whole** (CSI and SS3 forms): pressing an arrow or function key at the prompt no longer prints `[A` nor corrupts the answer.
- **Both backspace encodings erase**: DEL and `^H` (some terminals, legacy Windows console).
- **Ctrl+C at a prompt exits 130**, like a real SIGINT: the interrupt only exists as an error in raw mode, so `prompt.ErrInterrupt` (now exported) is mapped by `AdaptCmd` alongside the existing cancellation cases. This is the one deliberate behavior change — before and after #14161 alike, a prompt interrupt reported a generic failure.
- **The `[y/N]` hint shows in pipe mode too**: the message reaching a log or a piped consumer documents what was asked and what the default was, as it did before #14161 moved hints into the prompt.

Each point is locked by a test (pty-driven arrow-key round-trip, both-encodings erase, 130 mapping through `AdaptCmd`, updated pipe expectations).

### DIFF
--- a/cmd/compose/options_test.go
+++ b/cmd/compose/options_test.go
@@ -403,7 +403,7 @@ func TestConfirmRemoteIncludes(t *testing.T) {
 				"  - oci://registry.example.com/stack:latest\n" +
 				"  - git://github.com/user/repo.git\n" +
 				"\nRemote includes could potentially be malicious. Make sure you trust the source.\n" +
-				"Do you want to continue?",
+				"Do you want to continue? [y/N]: ",
 		},
 		{
 			name: "user rejects remote includes",
@@ -422,7 +422,7 @@ func TestConfirmRemoteIncludes(t *testing.T) {
 			wantOutput: "\nWarning: This Compose project includes files from remote sources:\n" +
 				"  - oci://registry.example.com/stack:latest\n" +
 				"\nRemote includes could potentially be malicious. Make sure you trust the source.\n" +
-				"Do you want to continue?",
+				"Do you want to continue? [y/N]: ",
 		},
 	}
 
--- a/cmd/prompt/prompt.go
+++ b/cmd/prompt/prompt.go
@@ -31,7 +31,11 @@ import (
 
 //go:generate mockgen -destination=./prompt_mock.go -self_package "github.com/docker/compose/v5/pkg/prompt" -package=prompt . UI
 
-var errInterrupt = errors.New("interrupt")
+// ErrInterrupt is returned by an interactive prompt when the user presses
+// Ctrl+C. The terminal being in raw mode, no SIGINT is ever delivered: this
+// error is the only interrupt signal callers get, and the command runner
+// maps it to the conventional 130 exit status like a real SIGINT.
+var ErrInterrupt = errors.New("interrupt")
 
 // UI - prompt user input
 type UI interface {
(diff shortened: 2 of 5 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The three fixes are clean and correct. Escape-sequence swallowing, dual backspace handling, and `ErrInterrupt`→130 mapping all look good. Each is covered by a test.

- [low] cmd/prompt/prompt.go:148 — `discardEscapeSequence` does not handle ANSI intermediate bytes (0x20–0x2F): the CSI parameter loop correctly handles numeric params and `;" (all ≥0x30), but true ECMA-48 intermediate bytes (0x20–0x2F, rare in practice) would not be recognized as non-final, causing the loop to stop early and treat the next byte as a new sequence. This is an extremely minor edge case for a y/n prompt and no practical terminal emits such sequences for cursor/function keys.

cmd/prompt/prompt.go
```diff
@@ -59,13 +63,8 @@ func (u User) Confirm(message string, defaultValue bool) (bool, error) {
...
 	defer u.stdin.RestoreTerminal()
 
-	prompt := " [y/N]: "
-	if defaultValue {
-		prompt = " [Y/n]: "
-	}
-
 	for {
-		_, _ = fmt.Fprint(u.stdout, message+prompt)
+		_, _ = fmt.Fprint(u.stdout, message+confirmHint(defaultValue))
```
thaJeztah, line 67: Interesting that it added a helper (thought from your Slack thread that it generally preferred just raw inline)?

cmd/prompt/prompt.go
```diff
@@ -31,7 +31,11 @@ import (
 
 //go:generate mockgen -destination=./prompt_mock.go -self_package "github.com/docker/compose/v5/pkg/prompt" -package=prompt . UI
 
-var errInterrupt = errors.New("interrupt")
+// ErrInterrupt is returned by an interactive prompt when the user presses
+// Ctrl+C. The terminal being in raw mode, no SIGINT is ever delivered: this
+// error is the only interrupt signal callers get, and the command runner
+// maps it to the conventional 130 exit status like a real SIGINT.
+var ErrInterrupt = errors.New("interrupt")
```
thaJeztah, line 38: When looking at the original PR, I was actually considering if we wanted to have context-cancellation somewhere. 

(I kept the `interrupt` error because the old library printed that, but I think it's not that common to print something normally, e.g. 

```bash
docker system prun
...
Are you sure you want to continue? [y/N] ^C
```

cmd/compose/options_test.go
```diff
@@ -403,7 +403,7 @@ func TestConfirmRemoteIncludes(t *testing.T) {
 				"  - oci://registry.example.com/stack:latest\n" +
 				"  - git://github.com/user/repo.git\n" +
 				"\nRemote includes could potentially be malicious. Make sure you trust the source.\n" +
-				"Do you want to continue?",
+				"Do you want to continue? [y/N]: ",
```
thaJeztah, line 406: Wait; did I mess this up? (if so; why didn't CI fail?)

PR #14215, opened 2026-09-11
fix(hash): pin the service config-hash byte layout, decoupled from compose-go struct refactorings

ndeloof:
**What this PR does, in one sentence:** the service config-hash byte layout is pinned to its historical form, so a compose-go struct refactoring can never again change hashes — and existing container stamps stay valid verbatim: no migration, no recreation.

## Context

`ServiceHash` — the value behind `com.docker.compose.config-hash`, which decides whether `up` recreates a container — digests `json.Marshal` of `types.ServiceConfig` directly. `encoding/json` emits struct fields in **declaration order** and flattens embedded structs at their embedding position, so every recorded hash is silently coupled to compose-go's struct layout. That has held only because compose-go never reordered its fields: the upcoming container-spec layering (compose-spec/compose-go#866, adopted by #14093) regroups the whole struct, and without this fix its adoption would recreate every running container on the first `up` after upgrade — the very kind of undisclosed side effect the hash exists to prevent.

## What the PR brings

- **The hash re-emits the marshaled object with its root keys in a frozen list** reproducing the historical order (generated by reflection over the last pre-layering compose-go — the last window where that order is observable in code), values byte-verbatim. Today the output is byte-identical to the direct marshal, **proven by a continuity test**; from the first struct reorder on, the frozen list alone carries that continuity, locked by golden-value tests. Full backward compatibility: not a single container, network or volume is touched by this change.
- **Future attributes stay cheap**: a root key missing from the list is appended in sorted order — deterministic, and thanks to `omitempty` only configurations using the new attribute see their hash move, exactly like a field addition always did. A reflection test fails when compose-go grows a root attribute absent from the list, keeping the hash surface a reviewed decision.
- **The residual exposure is named and guarded**: nested objects (`deploy`, `healthcheck`, …) keep their own struct marshal — a reorder *inside* one of them would still move hashes. The golden tests turn that from a silent side effect into a caught, reviewed event at the offending PR. Network and volume hashes are unchanged (their structs are not being reordered) and gain the same golden locks.
- The continuity test is deliberately transient: the compose-go upgrade that first reorders the struct deletes it in the same commit — the moment the frozen list takes over is explicit in history.

#14093 reorders `ServiceConfig` fields as a structural consequence of the jobs/container-spec work; landing this first means that PR ships with **zero** hash impact — nothing to put in its release notes — and any future compose-go layout change stays free as well.

### DIFF
--- a/pkg/compose/hash.go
+++ b/pkg/compose/hash.go
@@ -17,11 +17,132 @@
 package compose
 
 import (
+	"bytes"
 	"encoding/json"
+	"sort"
 
 	"github.com/compose-spec/compose-go/v2/types"
 	"github.com/opencontainers/go-digest"
 )
+
+// serviceHashKeyOrder freezes the top-level JSON key order of the service
+// config-hash. Hashing json.Marshal of the struct directly would couple every
+// recorded hash to the DECLARATION ORDER of compose-go's fields —
+// encoding/json emits struct fields in that order, and flattens embedded
+// structs at their embedding position — so any compose-go refactoring moving
+// a field (or grouping fields into embedded specs) would change the bytes,
+// and with them the hash, of configurations that did not change at all:
+// every container recreated on the first `up` after an upgrade.
+//
+// This list pins the byte layout to the historical form instead, generated
+// by reflection over compose-go v2.15.1-0.20260908103050 (the last layout
+// every released hash was computed from), so existing container stamps stay
+// valid verbatim: no migration, no recreation. A root attribute missing from
+// the list (added to compose-go later) is emitted after the listed ones, in
+// sorted order — deterministic, and thanks to omitempty only configurations
+// using the new attribute see their hash move, exactly like a field addition
+// always did. Nested objects keep the struct marshal of their own types; a
+// reorder inside one of them would still move hashes — the golden tests
+// exist to turn that into a reviewed decision instead of a side effect.
+var serviceHashKeyOrder = []string{
+	"profiles",
+	"annotations",
+	"attach",
+	"build",
+	"develop",
+	"blkio_config",
+	"cap_add",
+	"cap_drop",
+	"cgroup_parent",
+	"cgroup",
+	"cpu_count",
+	"cpu_percent",
+	"cpu_period",
+	"cpu_quota",
+	"cpu_rt_period",
+	"cpu_rt_runtime",
+	"cpus",
+	"cpuset",
+	"cpu_shares",
+	"command",
+	"configs",
+	"container_name",
+	"credential_spec",
+	"depends_on",
+	"deploy",
+	"device_cgroup_rules",
+	"devices",
+	"dns",
+	"dns_opt",
+	"dns_search",
+	"dockerfile",
+	"domainname",
+	"entrypoint",
+	"provider",
+	"environment",
+	"env_file",
+	"expose",
+	"extends",
+	"external_links",
+	"extra_hosts",
+	"group_add",
+	"gpus",
+	"hostname",
+	"healthcheck",
+	"image",
+	"init",
+	"ipc",
+	"isolation",
+	"labels",
+	"label_file",
+	"links",
+	"logging",
+	"log_driver",
+	"log_opt",
+	"mem_limit",
+	"mem_reservation",
+	"memswap_limit",
+	"mem_swappiness",
+	"mac_address",
+	"models",
+	"net",
+	"network_mode",
+	"networks",
+	"oom_kill_disable",
+	"oom_score_adj",
+	"pid",
+	"pids_limit",
+	"platform",
+	"ports",
+	"privileged",
+	"pull_policy",
+	"read_only",
+	"restart",
+	"runtime",
+	"scale",
+	"secrets",
+	"security_opt",
+	"shm_size",
+	"stdin_open",
+	"stop_grace_period",
+	"stop_signal",
+	"storage_opt",
+	"sysctls",
+	"tmpfs",
+	"tty",
+	"ulimits",
+	"use_api_socket",
+	"user",
+	"userns_mode",
+	"uts",
+	"volume_driver",
+	"volumes",
+	"volumes_from",
+	"working_dir",
+	"pre_start",
+	"post_start",
+	"pre_stop",
+}
 
 // ServiceHash computes the configuration hash for a service.
 func ServiceHash(o types.ServiceConfig) (string, error) {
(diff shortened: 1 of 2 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The canonical-hash migration is well-reasoned and correctly implemented. The core `canonicalHash` approach (marshal → `UseNumber` decode into `map[string]any` → re-marshal with sorted keys) is sound and will be stable across compose-go struct reorderings. The dual-acceptance logic for volumes in `reconcileVolumes` is also correct: it recomputes the legacy hash from the current desired config, which correctly diverges when the config actually changed and matches only when unchanged.

The one finding the drafter flagged as medium — that `legacyVolumeHash` would compute the wrong legacy hash if a future compose-go upgrade reordered `VolumeConfig` fields — was **dismissed by the verifier** as an inherent limitation of the pre-existing legacy approach, consciously accepted and documented in the PR. The bridge only needs to survive until users' volumes have been recreated/updated once; afterward the canonical hash takes over and becomes layout-independent. This is a reasonable tradeoff given that volume recreation is destructive.

No CONFIRMED or LIKELY bugs were found in the introduced code.

pkg/compose/hash_test.go
```diff
@@ -41,3 +45,151 @@ func serviceConfig(replicas int) types.ServiceConfig {
...
+	walk = func(rt reflect.Type) {
+		for i := 0; i < rt.NumField(); i++ {
+			f := rt.Field(i)
+			tag := f.Tag.Get("json")
+			name, _, _ := strings.Cut(tag, ",")
+			if name == "-" {
+				continue
+			}
+			if f.Anonymous && name == "" {
+				walk(f.Type)
```
docker-agent: **[medium] `TestServiceHashKeyOrderCoversStruct` panics on pointer-embedded anonymous struct fields**

The `walk` function recurses into anonymous struct fields via `walk(f.Type)`, but it does not check whether `f.Type` is a pointer kind first. If `types.ServiceConfig` ever gains an anonymous pointer-embedded field (e.g. `*EmbeddedSpec`), calling `rt.NumField()` on a pointer type panics with:

```
reflect: NumField of non-struct type *EmbeddedSpec
```

This is exactly the pattern `encoding/json` handles by dereferencing — the test should match. The fix is to dereference before recursing:

```suggestion
				ft := f.Type
				if ft.Kind() == reflect.Ptr {
					ft = ft.Elem()
				}
				walk(ft)
```

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 62/100 |

docker-agent (commented):

One medium-severity finding in new test code introduced by this PR.

pkg/compose/hash_test.go
ndeloof: Applied in ce4e45db1 — the walk now dereferences embedded pointers before recursing, matching encoding/json. Relevant precisely because #14093 introduces embedded specs: if they ever become pointer embeds, the guard test must keep working rather than panic.

docker-agent (commented):
### Assessment: 🟢 APPROVE

The PR correctly pins the service config-hash byte layout using a frozen top-level key order, decoupling it from compose-go struct field ordering. The implementation is sound:

- `pinRootKeyOrder` correctly reconstructs JSON with frozen key ordering, using `bytes.Buffer.Write` (which never errors) and `json.Marshal(key)` (safe for plain strings)
- The sorted-append for unknown keys is deterministic and correct
- The golden test mechanism will catch future unintended hash changes
- The struct coverage test (`TestServiceHashKeyOrderCoversStruct`) ensures the frozen list stays in sync with compose-go via reflection
- The continuity test correctly validates byte-for-byte identity before compose-go reorders fields

No bugs were found in the introduced code.

pkg/compose/hash.go
```diff
@@ -37,30 +158,74 @@ func ServiceHash(o types.ServiceConfig) (string, error) {
...
+		if val, ok := root[key]; ok {
+			write(key, val)
+			delete(root, key)
+		}
+	}
+	rest := make([]string, 0, len(root))
+	for key := range root {
+		rest = append(rest, key)
+	}
+	sort.Strings(rest)
```
glours, line 204: This hand-rolled "collect keys then sort" loop duplicates a helper that already exists twice in this package: `sortedKeys` in `reconcile.go:1081` (and `sortedMapKeys` in `publish.go:415`). Since `root` is `map[string]json.RawMessage`, `sortedKeys(root)` drops in directly here and lets you drop the now-unused `sort` import.

```suggestion
	for _, key := range sortedKeys(root) {
```
(remove the `rest := ...` / `sort.Strings(rest)` lines above, and the `"sort"` import at the top of the file)

pkg/compose/hash.go
```diff
@@ -37,30 +158,74 @@ func ServiceHash(o types.ServiceConfig) (string, error) {
 	o.DependsOn = nil
 	o.Profiles = nil
```
glours, line 159: Minor dedup opportunity: this field-trimming block (`Build`/`PullPolicy`/`Scale`/`Deploy.Replicas`/`DependsOn`/`Profiles`) is hand-copied verbatim in `TestServiceHashContinuity` (hash_test.go) to build its "legacy" comparison hash. If this trimming ever changes here, that test's private copy silently drifts instead of catching it. Suggest extracting it into a small unexported helper, e.g.:

```go
func trimServiceHashFields(o types.ServiceConfig) types.ServiceConfig {
	o.Build = nil
	o.PullPolicy = ""
	o.Scale = nil
	if o.Deploy != nil {
		deploy := *o.Deploy
		deploy.Replicas = nil
		o.Deploy = &deploy
	}
	o.DependsOn = nil
	o.Profiles = nil
	return o
}
```
called from both `ServiceHash` and the test.

pkg/compose/hash_test.go
```diff
@@ -41,3 +45,156 @@ func serviceConfig(replicas int) types.ServiceConfig {
...
+			if name == "-" {
+				continue
+			}
+			if f.Anonymous && name == "" {
+				// encoding/json dereferences embedded pointers; match it
+				ft := f.Type
+				if ft.Kind() == reflect.Pointer {
+					ft = ft.Elem()
+				}
+				walk(ft)
```
glours, line 189: This reflection walker has two robustness gaps relative to real `encoding/json` promotion rules — dormant today since `ServiceConfig` has no unexported or non-struct-embedded fields, but worth hardening before the anticipated compose-go layering refactor introduces either:

1. `reflect.Type.NumField()` panics if `ft` isn't a struct (e.g. an anonymous embed of a named `int` type or an interface). Guard with `ft.Kind() == reflect.Struct` before recursing.
2. Non-anonymous unexported fields aren't skipped — they'd fall through to `name = f.Name` and fail the `known[name]` assertion (a false test failure, not a silent miss, but still worth handling explicitly with `f.IsExported()`).

One subtlety if you do add the exported-field check: an *anonymous* unexported struct embed should still be recursed into, since `encoding/json` promotes its exported fields regardless of the embed's own export status — so the `IsExported()` check should come after the anonymous-struct-recursion branch, not before it, e.g.:

```go
if f.Anonymous && name == "" {
	ft := f.Type
	if ft.Kind() == reflect.Pointer {
		ft = ft.Elem()
	}
	if ft.Kind() == reflect.Struct {
		walk(ft)
		continue
	}
}
if !f.IsExported() {
	continue
}
```

glours (commented):
Reviewed the pinning mechanism itself (`pinRootKeyOrder`, the frozen `serviceHashKeyOrder` list, golden/continuity tests) and found it sound: verified the JSON re-serialization is byte-verbatim with no precision/escaping/duplicate-key issues, and cross-checked all 97 entries in `serviceHashKeyOrder` against the exact pinned compose-go version field-by-field with no gaps. `go build`, `go test ./pkg/...`, and `golangci-lint` are all clean on this branch.

Three small, non-blocking items below — a couple of reuse/dedup nits and one test-hardening suggestion for the reflection walker. None of these affect the correctness of the hash itself; happy to see this merge once you've had a look.

pkg/compose/hash.go
ndeloof: Done in ab98eda5d: pinRootKeyOrder now uses sortedKeys, sort import dropped.

pkg/compose/hash.go
ndeloof: Done in ab98eda5d: extracted trimServiceHashFields, shared by ServiceHash and the continuity test. Golden hashes unchanged.

pkg/compose/hash_test.go
ndeloof: Done in ab98eda5d, following your exact promotion-rules ordering: struct-kind guard before recursing into anonymous embeds (non-struct embeds fall through to regular handling), then IsExported() skip after that branch so an unexported anonymous struct embed still gets its exported fields promoted.

docker-agent (commented):
*This review covers only the commits pushed since `ce4e45db1dcd`.*

### Assessment: 🟢 APPROVE

The incremental changes are a clean refactoring with no bugs introduced:

- **`trimServiceHashFields` extraction** is correct: `types.ServiceConfig` is passed and returned by value; `ServiceHash` captures the result with `o = trimServiceHashFields(o)`, so all pointer-field nil-ings take effect. The `Deploy` deep-copy inside the function is preserved correctly.
- **`sortedKeys` call** replaces the equivalent inline sort-and-iterate loop. `sortedKeys[V any](m map[string]V) []string` is defined in `pkg/compose/reconcile.go` in the same package — in scope, correct signature, semantically identical to the removed code.
- **`"sort"` import removal** is safe: the only use of `sort` in `hash.go` was the now-replaced `sort.Strings(rest)` call.

glours (approved):
LGTM

PR #14217, opened 2026-09-12
fix(watch): exclude Dockerfile and compose files from the sync loop

Orivital:
**What I did**

#14117 restored the `initialSync` exclusion so `docker compose watch` does not copy the Dockerfile or compose files into the container on first sync. That PR called out that the continuous loop (`getWatchRules`) never had the same ignore, and that it would need a differently-anchored matcher because it matches absolute host paths rather than basenames.

Without that, a `develop.watch` sync on the project root still copies `Dockerfile` / `compose.yaml` (and a custom `build.dockerfile`) into the container when those files change after watch has started.

Shared the ignore as `dockerFileIgnoreMatcher` with `**/`-anchored patterns so it works for both `initialSync`'s basenames and the watch loop's full paths. Applied it only to copy actions (`sync`, `sync+restart`, `sync+exec`) so a rebuild trigger on the same tree still fires when those files change.

**Test plan**

- Unit: `go test ./pkg/compose/ -run 'TestGetWatchRules|TestInitialSync'`
- Repro with released `docker compose watch` on a root `sync` to `/work`: editing `Dockerfile` after start copies it into the container
- Same steps with this binary: `/work` stays empty
- Existing `TestWatch*` e2e cases do not cover this path; the full e2e suite was not run

**Related issue**
N/A

### DIFF
--- a/pkg/compose/watch.go
+++ b/pkg/compose/watch.go
@@ -336,6 +336,11 @@ func getWatchRules(config *types.DevelopConfig, service types.ServiceConfig) ([]
 		return nil, err
 	}
 
+	dockerFileIgnore, err := dockerFileIgnoreMatcher(service)
+	if err != nil {
+		return nil, err
+	}
+
 	for _, trigger := range config.Watch {
 		ignore, err := watch.NewDockerPatternMatcher(trigger.Path, trigger.Ignore)
 		if err != nil {
@@ -352,15 +357,22 @@ func getWatchRules(config *types.DevelopConfig, service types.ServiceConfig) ([]
 			}
 		}
 
+		ignores := []watch.PathMatcher{
+			dockerIgnores,
+			watch.EphemeralPathMatcher(),
+			dotGitIgnore,
+			ignore,
+		}
+		// Copy actions only: rebuild on the same tree must still fire.
+		switch trigger.Action {
+		case types.WatchActionSync, types.WatchActionSyncRestart, types.WatchActionSyncExec:
+			ignores = append(ignores, dockerFileIgnore)
+		}
+
 		rules = append(rules, watchRule{
 			Trigger: trigger,
 			include: include,
-			ignore: watch.NewCompositeMatcher(
-				dockerIgnores,
-				watch.EphemeralPathMatcher(),
-				dotGitIgnore,
-				ignore,
-			),
+			ignore:  watch.NewCompositeMatcher(ignores...),
 			service: service.Name,
 		})
 	}
@@ -760,6 +772,21 @@ func (s *composeService) pruneDanglingImagesOnRebuild(ctx context.Context, proje
 	}
 }
 
+// **/anchored so the matcher hits both initialSync's basenames and the
+// watch loop's absolute host paths.
+func dockerFileIgnoreMatcher(service types.ServiceConfig) (watch.PathMatcher, error) {
+	names := append([]string{"Dockerfile"}, cli.DefaultFileNames...)
+	names = append(names, cli.DefaultOverrideFileNames...)
+	if service.Build != nil && service.Build.Dockerfile != "" {
+		names = append(names, filepath.Base(service.Build.Dockerfile))
+	}
+	patterns := make([]string, len(names))
+	for i, name := range names {
+		patterns[i] = "**/" + name
+	}
+	return watch.NewDockerPatternMatcher("/", patterns)
+}
+
 // Walks develop.watch.path and checks which files should be copied inside the container
 // ignores develop.watch.ignore, Dockerfile, compose files, bind mounted paths and .git
 func (s *composeService) initialSync(ctx context.Context, service types.ServiceConfig, trigger types.Trigger, syncer sync.Syncer) error {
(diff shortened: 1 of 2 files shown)


pkg/compose/watch.go
```diff
@@ -760,6 +772,21 @@ func (s *composeService) pruneDanglingImagesOnRebuild(ctx context.Context, proje
 	}
 }
 
+// **/anchored so the matcher hits both initialSync's basenames and the
+// watch loop's absolute host paths.
+func dockerFileIgnoreMatcher(service types.ServiceConfig) (watch.PathMatcher, error) {
+	names := append([]string{"Dockerfile"}, cli.DefaultFileNames...)
```
ndeloof, line 778: this should only select service.Build.Dockerfile.

Orivital: Makes sense. The watch matcher now only uses `service.Build.Dockerfile` (or `Dockerfile` when that's empty). Compose files are still skipped on `initialSync`, same as #14117.
fe986b565b811751dc501d8cdb839f7c2c8518b6

pkg/compose/watch.go
```diff
@@ -336,6 +336,11 @@ func getWatchRules(config *types.DevelopConfig, service types.ServiceConfig) ([]
 		return nil, err
 	}
 
+	dockerFileIgnore, err := dockerFileIgnoreMatcher(service)
+	if err != nil {
+		return nil, err
+	}
```
glours: `composeFileIgnore` only exists in `initialSync`; `getWatchRules` never gets it, so compose files edited after `watch` starts are still synced, the bug this PR claims to fix.

```suggestion
	dockerFileIgnore, err := dockerFileIgnoreMatcher(service)
      if err != nil {
              return nil, err
      }

      composeFiles := append([]string{}, cli.DefaultFileNames...)
      composeFiles = append(composeFiles, cli.DefaultOverrideFileNames...)
      composeFilePatterns := make([]string, len(composeFiles))
      for i, f := range composeFiles {
              composeFilePatterns[i] = "**/" + f
      }
      composeFileIgnore, err := watch.NewDockerPatternMatcher("/", composeFilePatterns)
      if err != nil {
              return nil, err
      }
```

Plain filenames won't do, this loop matches full absolute paths (`/proj/compose.yaml`), not basenames like `initialSync`, so each pattern needs the same `**/` anchor `dockerFileIgnoreMatcher` uses.

pkg/compose/watch.go
```diff
@@ -352,15 +357,22 @@ func getWatchRules(config *types.DevelopConfig, service types.ServiceConfig) ([]
...
+			dockerIgnores,
+			watch.EphemeralPathMatcher(),
+			dotGitIgnore,
+			ignore,
+		}
+		// Copy actions only: rebuild on the same tree must still fire.
+		switch trigger.Action {
+		case types.WatchActionSync, types.WatchActionSyncRestart, types.WatchActionSyncExec:
+			ignores = append(ignores, dockerFileIgnore)
+		}
```
glours: Missing `composeFileIgnore` here too. Also duplicates `isSync()` (line 382) instead of reusing it.

```suggestion
		// Copy actions only: rebuild on the same tree must still fire.
		if isSync(trigger) || trigger.Action == types.WatchActionSyncExec {
				ignores = append(ignores, dockerFileIgnore, composeFileIgnore)
		}
```

pkg/compose/watch.go
```diff
@@ -793,6 +772,19 @@ func (s *composeService) pruneDanglingImagesOnRebuild(ctx context.Context, proje
...
+func dockerFileIgnoreMatcher(service types.ServiceConfig) (watch.PathMatcher, error) {
+	if service.Build == nil {
+		return watch.EmptyMatcher{}, nil
+	}
+	name := service.Build.Dockerfile
+	if name == "" {
+		name = "Dockerfile"
+	}
+	return watch.NewDockerPatternMatcher("/", []string{"**/" + filepath.Base(name)})
+}
```
glours, line 786: Basename-only match, needed for `initialSync` but now also applied to the continuous loop, which matches full paths and never needed it. Any unrelated file sharing the Dockerfile's basename anywhere in the tree silently stops syncing. Fix: give `getWatchRules` its own path-qualified matcher instead of reusing this one; not a same-line suggestion.

pkg/compose/watch.go
glours, line 370: `composeFileIgnore` only exists in `initialSync`; `getWatchRules` never builds it, so compose files edited after `watch` starts are still synced,  the bug this PR claims to fix. 
Fix: inside the `for _, trigger := range config.Watch` loop, build `composeFileIgnore` anchored on `trigger.Path` (same as `ignore`/`include` just above, using `cli.DefaultFileNames`/`DefaultOverrideFileNames`), and append it alongside `dockerFileIgnore` for sync/sync-restart/sync-exec triggers. Anchor on `trigger.Path`, not `**/`+basename — this only excludes the real compose file at the watched root, not a same-named file nested elsewhere. Verified working in an isolated build/test.

Also folds in a nit: the `switch trigger.Action { case Sync, SyncRestart, SyncExec: ... }` (line 367-370) duplicates `isSync()` (line 382) instead of reusing it — `isSync(trigger) || trigger.Action == types.WatchActionSyncExec` covers the same three cases.

glours (changes requested):
Compose-file exclusion isn't wired into `getWatchRules` (comment 1, `isSync` nit folded in) — fix described inline and verified by build+test, not posted as a suggestion since it touches too many lines cleanly. Basename matching over-applies to the continuous loop (comment 2, needs a new helper). Tests/gofmt clean otherwise.

PR #14221, opened 2026-09-15
pre_start hook runners become reconciliation-plan resources

ndeloof:

`pre_start` hook containers become first-class resources of the reconciliation plan: the plan creates them, the start phase only executes them.

## Context

`pre_start` hook containers have their own identity — a name, labels, and possibly an image different from the service's — yet they live entirely outside the reconciliation plan: today `runPreStart` creates, runs and removes them imperatively inside the start flow. The engine cannot see them, so it cannot order them, purge them deterministically, or account for their failures the way it does for every other resource. Leftover runners from a failed run are swept by ad-hoc label scans, and nothing ties a runner to the hook it was created for.

## What the PR brings

The reconciler now plans one `CreateHookContainer` operation per declared hook, whenever the start phase is going to run the hooks — a replica to start and no running replica *surviving the plan* (a recreated replica is not running at start time, so its runners are planned too). Runner creation is properly ordered in the DAG: after every replica operation of its service (so the executor resolves the `VolumesFrom` target against a final live view), after the infrastructure, and after the purge of previously observed runners. Runners get a deterministic name (`<project>-<service>-pre_start-<i>`) and a new `com.docker.compose.hook-index` label, so repeated plans converge on the same container instead of accumulating anonymous ones, and the start phase can match each declared hook with the runner prepared for it.

`runPreStart` becomes pure execution: it looks up the created-state runner for each hook and never creates one. A declared hook without a prepared runner is an actionable error naming the reconciliation command to run (`docker compose up <service>`). The accepted consequence — locked by an e2e scenario — is that `stop` then `start` of a hooked service errors, since runners are consumed on success; `up` covers the overwhelming majority of usage and always prepares fresh runners.

Guardrails: success removal, failure retention (container kept for post-mortem) and cancellation cleanup are byte-for-byte the previous behavior; runners stay invisible to start/ps listings (they carry no config-hash label, which the default filters require); `down` keeps removing them by hook label; runners from previous compose versions (no index label) are recognized and purged.

The plan engine currently covers the create phase only. Moving runner *creation* into that phase needs no start-phase operations, so this lands independently of the start-phase plan (#14200) while setting up the follow-up: with runners as plan resources, hook *execution and removal* can become start-phase operations, making failure retention a DAG property instead of imperative code.


docker-agent (commented):
⚠️ **Review did not complete** — The agent ran but did not post a review. [View logs](https://github.com/docker/compose/actions/runs/34956407078) for details. Re-request a review from `docker-agent` to retry.

pkg/compose/reconcile.go
```diff
@@ -750,14 +767,66 @@ func (r *reconciler) reconcileService(service types.ServiceConfig) error {
...
+	for i := range service.PreStart {
+		node := r.plan.addNode(Operation{
+			Type:       OpCreateHookContainer,
+			ResourceID: fmt.Sprintf("hook:%s:%s:%d", service.Name, preStartHookType, i),
+			Cause:      "pre_start hook",
+			Service:    &serviceCopy,
+			HookIndex:  i,
+			Name:       getHookContainerName(r.project.Name, service.Name, i),
+		}, "", deps...)
+		deps = []*PlanNode{node}
```
glours, line 826: The loop reassigns `deps` to just the previous hook's node, chaining every `OpCreateHookContainer` for a service into a serial dependency line — even though each targets the same already-live replica independently and has no reason to wait on a sibling hook's create. Keeping `deps` fixed at the shared set (`containerNodes+infraDeps+purges`) lets the executor create them in parallel instead.

```suggestion
	for i := range service.PreStart {
		r.plan.addNode(Operation{
			Type:       OpCreateHookContainer,
			ResourceID: fmt.Sprintf("hook:%s:%s:%d", service.Name, preStartHookType, i),
			Cause:      "pre_start hook",
			Service:    &serviceCopy,
			HookIndex:  i,
			Name:       getHookContainerName(r.project.Name, service.Name, i),
		}, "", deps...)
	}
```

pkg/compose/reconcile_test.go
```diff
@@ -1251,6 +1251,209 @@ func TestReconcileContainers_ExitedIsNoop(t *testing.T) {
...
+		Volumes:        map[string][]ObservedVolume{},
+	}
+
+	plan, err := reconcile(t.Context(), project, observed, defaultReconcileOptions(), noPrompt)
+	assert.NilError(t, err)
+
+	assert.Equal(t, plan.String(), strings.TrimSpace(`
+[] -> #1 service:app:1, CreateContainer, no existing container
+[1] -> #2 hook:app:pre_start:0, CreateHookContainer, pre_start hook
+[2] -> #3 hook:app:pre_start:1, CreateHookContainer, pre_start hook
```
glours, line 1324: Side effect of the fix suggested in reconcile.go: with `deps` no longer reassigned per iteration, hook[1]'s create would depend on `[1]` (the replica create) instead of `[2]` (hook[0]'s own node).

```suggestion
[1] -> #3 hook:app:pre_start:1, CreateHookContainer, pre_start hook
```

pkg/compose/pre_start.go
```diff
@@ -55,50 +64,93 @@ func lowestNumberedContainer(containers Containers) container.Summary {
...
+	if err != nil {
+		return err
 	}
-	for i, hook := range service.PreStart {
-		if err := s.runPreStartHook(ctx, project, service, ctr, i, hook, listener); err != nil {
+	for i := range service.PreStart {
+		runner, ok := runners[i]
+		if !ok {
+			return fmt.Errorf("service %q pre_start[%d]: no hook runner container found — runners are prepared when the service is created and consumed when its hooks run; run %q to prepare them again",
+				service.Name, i, "docker compose up "+service.Name)
```
glours, line 95: Narrow but real race: `keptRunning` is decided from the pre-plan `ObservedState` snapshot in `reconcile.go`, while `startService` re-checks container state independently via a fresh `ContainerList` *after* the whole create-phase plan has executed. If the sole surviving replica is stopped externally in that window, plan-time skips runner creation (replica looked like it would survive) but start-time now sees nothing running and hits this hard error — previously self-healing via the inline create, now a dead end pointing at `up` for a situation `up` didn't cause.

Not asking for the full fix here (that'd mean re-validating state at start time instead of trusting two independent daemon reads), but the message could at least hint this isn't necessarily a planning bug:

```suggestion
			return fmt.Errorf("service %q pre_start[%d]: no hook runner container found — either its runner was already consumed by a previous start, or the service's container state changed after reconciliation planned this run; run %q to prepare fresh runners",
				service.Name, i, "docker compose up "+service.Name)
```

glours (commented):
Nice one, turning hook runners into first-class plan resources removes a whole class of ad-hoc bookkeeping. Two non-blocking should-fixes found while going through the diff (AI-assisted review, verified against the code before posting), nothing here blocks merge.

pkg/compose/reconcile.go
ndeloof: Fixed in a51afaae0 -- deps is no longer reassigned, so every OpCreateHookContainer for a service depends on the same shared containerNodes+infraDeps+purges set instead of chaining serially.

pkg/compose/reconcile_test.go
ndeloof: Fixed in a51afaae0 -- the plan-string assertion now reads [1] -> #2 and [1] -> #3, both depending on the replica create rather than chaining.

pkg/compose/pre_start.go
ndeloof: Applied your suggested wording in a51afaae0.

pkg/compose/observed_state.go
```diff
@@ -170,6 +179,18 @@ func (s *composeService) collectObservedState(ctx context.Context, project *type
 
 	for _, ctr := range raw {
 		svcName := ctr.Labels[api.ServiceLabel]
+		if ctr.Labels[api.HookLabel] != "" && knownServices[svcName] {
+			// lifecycle-hook containers (ephemeral pre_start runners) are
+			// neither service replicas nor one-offs: classified apart, so
+			// they never masquerade as a replica (they carry no
+			// container-number label and would otherwise read as number 0)
```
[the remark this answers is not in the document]
ndeloof: Confirmed and fixed in 6a61f5680 -- reproduced the exact name-conflict-on-retry failure against a live daemon before fixing. Added getHookContainers (project+hook-label listing, no ConfigHashLabel requirement) merged into collectObservedState before classification, plus a new e2e scenario (TestPreStartHookRetryAfterFailureSucceeds) locking in the failure-then-retry sequence.

pkg/compose/service_containers.go
```diff
@@ -599,11 +599,10 @@ func (s *composeService) startService(ctx context.Context,
...
 	// pre_start runs once per service, only when no replica is already running
 	// (e.g. initial up, force-recreate, or spec change). per_replica: false is
-	// the only currently supported mode. Pick the replica with the lowest
-	// container-number so the choice is deterministic regardless of the order
-	// the daemon returns containers in.
+	// the only currently supported mode. The hooks execute in runner containers
+	// prepared by the reconciliation plan.
 	if len(service.PreStart) > 0 && len(serviceContainers) == len(toStart) {
-		if err := s.runPreStart(ctx, project, service, lowestNumberedContainer(toStart), listener); err != nil {
+		if err := s.runPreStart(ctx, project, service, listener); err != nil {
```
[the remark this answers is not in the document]
ndeloof: Fixed in 6a61f5680 -- added isNotHookContainer to startService's filter chain.

pkg/compose/reconcile.go
```diff
@@ -750,14 +767,66 @@ func (r *reconciler) reconcileService(service types.ServiceConfig) error {
...
+// so a genuinely stuck old runner surfaces as a name conflict on the create.
+// The creates also depend on every replica node and on the infrastructure, so
+// the executor resolves the target replica (VolumesFrom) against a final live
+// view with networks in place. When a running replica survives the plan, the
+// start phase skips pre_start entirely: nothing is created, and a hook
+// container retained after a failure stays available for inspection.
+func (r *reconciler) planPreStartHookRunners(service types.ServiceConfig, expectedScale int, keptRunning bool, containerNodes, infraDeps []*PlanNode) {
+	if len(service.PreStart) == 0 || expectedScale == 0 || keptRunning {
+		return
+	}
```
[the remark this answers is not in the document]
ndeloof: Fixed in 6a61f5680 -- planPreStartHookRunners now mirrors runPreStart's own up-front, all-or-nothing per_replica validation before planning any node (per_replica support itself is tracked separately in docker/compose#14259).

pkg/compose/pre_start.go
```diff
@@ -75,50 +83,96 @@ func lowestNumberedContainer(containers Containers) container.Summary {
...
+				"reconciliation planned this run; run %q to prepare fresh runners",
+				service.Name, i, "docker compose up "+service.Name)
+		}
+		if err := s.execPreStartHook(ctx, service, i, runner.ID, listener); err != nil {
 			return err
 		}
+		// Success: remove the hook container, mirroring the old AutoRemove behaviour
+		// (including its anonymous volumes). A removal failure is logged but does not
+		// gate service start — the hook already succeeded.
+		if _, removeErr := s.apiClient().ContainerRemove(ctx, runner.ID, client.ContainerRemoveOptions{RemoveVolumes: true}); removeErr != nil {
```
docker-agent: **[low] Success-removal uses the cancellable `ctx`; hook container lingers if context is cancelled between hook success and `ContainerRemove`**

`runPreStart` calls `s.apiClient().ContainerRemove(ctx, runner.ID, …)` after a successful hook run. If `ctx` is cancelled in the narrow window between `execPreStartHook` returning `nil` and `ContainerRemove` executing (e.g. a Ctrl-C arriving at that exact instant), the removal fails with a context-cancelled error. The warn path swallows it, leaving the hook container in exited state.

This is inconsistent with the cancellation-cleanup path in `execPreStartHook` (line 213), which already uses `context.Background()` to guarantee the container is removed even after cancellation. The success-removal path should use the same pattern:

```suggestion
		if _, removeErr := s.apiClient().ContainerRemove(context.WithoutCancel(ctx), runner.ID, client.ContainerRemoveOptions{RemoveVolumes: true}); removeErr != nil {
```

The practical impact is low — the container is warn-logged and will be purged by the next `planPreStartHookRunners` call — but the inconsistency with the already-established `context.Background()` pattern in the same file is worth fixing.

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 100/100 |

pkg/compose/reconcile.go
```diff
@@ -750,14 +767,76 @@ func (r *reconciler) reconcileService(service types.ServiceConfig) error {
...
+	for _, hook := range service.PreStart {
+		if hook.PerReplica {
```
docker-agent: **[medium] `planPreStartHookRunners` returns early on `PerReplica: true`, bypassing the stale-runner purge loop**

When any hook in `service.PreStart` has `PerReplica: true`, the function exits immediately before reaching the stale-container purge loop:

```go
for _, hook := range service.PreStart {
    if hook.PerReplica {
        return   // stale-runner purge below is never reached
    }
}
// stale purge loop starts here — never executed when PerReplica: true
stale := slices.Clone(r.observed.HookContainers[service.Name])
```

**Trigger**: a user who previously ran `up` with `per_replica: false` (which created hook runners), then sets `per_replica: true` in the compose file and runs `up` again. `runPreStart` rejects the hook before executing it, so the runners are never consumed, and `planPreStartHookRunners` returns early before scheduling their removal. The runners accumulate across `up` calls until `--remove-orphans` is passed or the hook is removed from the compose file.

**Fix**: separate the PerReplica guard from the purge — run the stale-purge loop unconditionally, then `return` before the create loop:

```go
for _, hook := range service.PreStart {
    if hook.PerReplica {
        // per_replica not yet supported: purge any stale runners left from a
        // previous run, but don't create new ones (they'd never be executed).
        break
    }
}
// purge stale runners
stale := ...
// only create new runners when all hooks are per_replica: false
for _, hook := range service.PreStart {
    if hook.PerReplica {
        return
    }
}
for i := range service.PreStart { ... }
```

Or more simply: collect the stale purge, then check PerReplica before the create loop only.

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 100/100 |

docker-agent (commented):

pkg/compose/observed_state.go
```diff
@@ -158,6 +167,18 @@ func (s *composeService) collectObservedState(ctx context.Context, project *type
...
+	// createPreStartContainer) -- so getContainers' filters never return
+	// them, and they must be listed separately, by project+hook label
+	// alone, or the classification loop below never sees them: stale
+	// runners would never be purged, and the next deterministic create
+	// would fail with a name conflict against the one left behind.
+	hookRaw, err := s.getHookContainers(ctx, project.Name)
+	if err != nil {
+		return nil, err
+	}
+	raw = append(raw, hookRaw...)
```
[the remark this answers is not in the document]
ndeloof: Fixed in 6c0fb017b -- deduplicated raw and hookRaw by container ID before classification. Locked in by TestCollectObservedState_LegacyHookRunnerNotDuplicated.

pkg/compose/reconcile.go
```diff
@@ -725,6 +735,9 @@ func (r *reconciler) reconcileService(service types.ServiceConfig) error {
 			// (start.go), which lists containers again and starts them in
 			// dependency order. Exited containers are deliberately left as-is
 			// here so that phase (or the user) decides.
+			if oc.State == container.StateRunning {
```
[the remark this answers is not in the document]
ndeloof: Fixed in 6c0fb017b -- a running replica only counts as kept when stoppedByPlan has no entry for it. Locked in by TestReconcileContainers_HookRunnerPlannedWhenCascadeStopsRunningReplica.

pkg/compose/service_containers.go
```diff
@@ -591,19 +591,20 @@ func (s *composeService) startService(ctx context.Context,
 		return fmt.Errorf("service %q has no container to start", service.Name)
 	}
 
-	serviceContainers := containers.filter(isService(service.Name), isNotOneOff)
+	serviceContainers := containers.filter(isService(service.Name), isNotOneOff, isNotHookContainer)
```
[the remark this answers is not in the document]
ndeloof: Fixed in 6c0fb017b -- added isNotHookContainer to waitDependencies' own filter too. Locked in by a new TestWaitDependencies subtest.

pkg/compose/pre_start.go
ndeloof: Fixed in 6c0fb017b -- matched the context.WithoutCancel pattern already used by the cancellation-cleanup path.

pkg/compose/reconcile.go
ndeloof: Fixed in 6c0fb017b -- the stale-runner purge now always runs; only the create loop is skipped for per_replica. Locked in by TestReconcileContainers_StaleHookRunnersPurgedEvenWhenPerReplica.

docker-agent (commented):
This review covers only the commits since `6a61f5680db4` (the second fix-up commit addressing the previous review round).

### Assessment: 🟢 APPROVE

The four changes in this incremental diff are all clean:

- **`observed_state.go`**: The `seen`-map deduplication correctly uses `container.ID` to prevent legacy runners (that still carry the config-hash label) from being appended twice and scheduled for double-removal.
- **`pre_start.go`**: `context.WithoutCancel(ctx)` on the success-removal path mirrors the already-established pattern on the cancellation-cleanup path — no resource leak introduced.
- **`reconcile.go`**: The `stoppedByPlan` guard is race-free (populated before `reconcileContainers` runs); the PerReplica guard repositioning is semantically correct — stale-runner purge always executes, orphaned-runner creation is still correctly prevented.
- **`service_containers.go`**: `isNotHookContainer` is defined correctly and prevents hook runner containers from blocking `waitDependencies`.

glours (commented):
`TestExecutePlugin_GetRelayInfo` and `TestExecutePlugin_GetRelayInfoDesktop` cover the happy path and the Desktop path, but nothing exercises `relayInfo`'s "gateway unresolved" branch (`NetworkInspect` error, or an IPAM config with no valid IPv4 gateway) — the exact case the `omitempty`/best-effort contract in `relay.go` is built around. Can you add a case mocking `NetworkInspect` to fail (see `notFoundError` in `executor_test.go`, already used this way in `observed_state_test.go:387`) and asserting the network is still listed with `Gateway` empty? Without it, a later change to the `err == nil`/`cfg.Gateway.IsValid()` guards could start sending a wrong non-empty gateway with nothing catching it.

ndeloof:
Rebased onto main now that #14200 (start-phase planning) is merged. The rebase surfaced the integration point between the two PRs: the start-phase `RunPreStart` node executed the hooks but nothing ordered it after the `CreateHookContainer` nodes that prepare the runners — the executor could have run hooks before their runners existed. The reconciler now records the runner-create nodes per service and the `RunPreStart` node depends on them (golden updated: `[1,3] -> #4 RunPreStart`). Unit tests, CI-pinned lint and the pre_start e2e scenarios are green.

pkg/compose/pre_start.go
```diff
@@ -75,50 +83,98 @@ func lowestNumberedContainer(containers Containers) container.Summary {
...
+// listPreStartRunners returns the service's created-state pre_start hook
+// runner containers, indexed by their HookIndexLabel. Runners in any other
+// state are ignored — a failed hook retained for inspection or an
+// old-generation container without an index label is not executable; only a
+// fresh runner prepared by the reconciliation plan is.
+func (s *composeService) listPreStartRunners(ctx context.Context, projectName, serviceName string) (map[int]container.Summary, error) {
+	f := projectFilter(projectName)
+	f.Add("label", serviceFilter(serviceName))
+	f.Add("label", hookFilter(preStartHookType))
+	res, err := s.apiClient().ContainerList(ctx, client.ContainerListOptions{
```
docker-agent, line 144: **[low] `listPreStartRunners`: missing `status=created` API filter causes over-fetch and a marginal TOCTOU window**

`listPreStartRunners` calls `ContainerList` with `All: true` and label-only filters, then discards every container whose state is not `created` client-side. Two effects:

1. **Over-fetch**: on a project where previous hook runs failed and their containers were not cleaned up (e.g. because the best-effort purge failed repeatedly), all those retained containers are transferred over the socket only to be thrown away. Adding `f.Add("status", "created")` before the call pushes the filter to the daemon and avoids this.

2. **TOCTOU window**: a runner found in `created` state at scan time can transition to another state before `ContainerStart` is called on it (though this window is inherent to any list-then-act pattern and a server-side filter does not close it entirely). With the filter it becomes marginally narrower.

```suggestion
	f.Add("status", "created")
	res, err := s.apiClient().ContainerList(ctx, client.ContainerListOptions{
		All:     true,
		Filters: f,
	})
```

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 100/100 |

docker-agent (commented):
### Assessment: 🟢 APPROVE

pkg/compose/pre_start.go
ndeloof: Applied in a175d5f1a (amended into the owning commit `feat: reconciliation plan creates pre_start hook runners`): `listPreStartRunners` now pushes `status=created` to the daemon via the list filter, avoiding the over-fetch. The client-side `StateCreated` check is kept as a last-line guard for the inherent list-then-act window.

docker-agent (commented):
### Assessment: 🟢 APPROVE

> ⚠️ **Verification note**: The verifier returned a malformed response (duplicate verdict ID for one finding), so findings could not be formally verified. The drafter's four hypotheses are surfaced below as unverified. Code inspection performed inline before and after delegation suggests all four are either false positives or pre-existing issues not introduced by this PR.

The PR design is sound: hook runners move cleanly into the reconciliation plan with deterministic naming, correct DAG ordering (stale-purge → replica nodes + infra → create runners), and `runPreStart` becoming pure execution. All three previously-fixed patterns from prior review threads are confirmed present:
- `context.WithoutCancel(ctx)` on success-removal (thread 1 ✓)
- Stale-runner purge always runs before the PerReplica guard (thread 2 ✓)
- `status=created` daemon-side filter in `listPreStartRunners` (thread 3 ✓)

---

#### Unverified drafter findings (surfaced for author review)

- **[low] `pkg/compose/reconcile.go`:838 — `keptRunning = true` for paused/dead containers in default restart path**: The `default:` block sets `keptRunning = true`, which causes `planPreStartHookRunners` to skip runner creation. This is consistent with the broader guard in `runPreStart` (the restarted container is both in `serviceContainers` and `toStart`, so hooks would be skipped anyway). Design inconsistency, not a correctness bug.

- **[medium] `pkg/compose/reconcile.go`:910 — PerReplica `return` inside check loop exits entire function**: When any hook has `PerReplica: true`, the function returns without creating runners for subsequent non-per_replica hooks. However: (1) the stale-purge loop runs unconditionally before this check; (2) `runPreStart` rejects the entire service upfront if ANY hook has `per_replica: true`, so there is no live execution path where mixed per_replica values would cause runner creation to be needed; (3) the comment documents this co-invariant explicitly. Assessed as intentional and correct.

- **[low] `pkg/compose/pre_start.go`:213 — cancellation-removal path uses `context.Background()`**: The cancellation-removal inside `execPreStartHook` uses `context.Background()` rather than `context.WithoutCancel(ctx)`. This is a pre-existing pattern unchanged by this PR (only the variable name `created.ID` → `containerID` changed). Not introduced by this PR.

- **[medium] `pkg/compose/observed_state.go`:206 — hook containers for removed services may not reach orphan classification**: When a hook container's service is removed from compose.yaml (`!knownServices[svcName]`), the code falls through to `isOrphaned`. The comment at line 212–214 explicitly documents this intent. `isOrphaned` (containers.go:175–183) checks `!slices.Contains(services, c.Labels[api.ServiceLabel])` with no `ConfigHashLabel` dependency — it correctly returns `true` for containers whose service is no longer in the project. Assessed as a false positive.

pkg/compose/observed_state.go
```diff
@@ -158,6 +167,30 @@ func (s *composeService) collectObservedState(ctx context.Context, project *type
...
+	// have, or it would be classified -- and scheduled for removal -- twice.
+	seen := make(map[string]bool, len(raw))
+	for _, ctr := range raw {
+		seen[ctr.ID] = true
+	}
+	for _, ctr := range hookRaw {
+		if !seen[ctr.ID] {
+			raw = append(raw, ctr)
+		}
+	}
```
glours: `getHookContainers` fires unconditionally on every `collectObservedState` call — one extra `ContainerList` round-trip per `up`/`create`, even for a project with zero `pre_start` hooks anywhere. `state.HookContainers[svc]` is only ever read in `planPreStartHookRunners`, which already bails out when `len(service.PreStart) == 0` (`reconcile.go:806`), so gating this fetch the same way is behavior-preserving.

```suggestion
	hasPreStartHooks := false
	for _, svc := range project.Services {
		if len(svc.PreStart) > 0 {
			hasPreStartHooks = true
			break
		}
	}
	if hasPreStartHooks {
		// Hook runner containers deliberately carry no ConfigHashLabel (kept
		// invisible to ps/start's default listing, which requires it -- see
		// createPreStartContainer) -- so getContainers' filters never return
		// them, and they must be listed separately, by project+hook label
		// alone, or the classification loop below never sees them: stale
		// runners would never be purged, and the next deterministic create
		// would fail with a name conflict against the one left behind.
		hookRaw, err := s.getHookContainers(ctx, project.Name)
		if err != nil {
			return nil, err
		}
		// A legacy runner created before this ConfigHashLabel exclusion existed
		// (or by an older compose version) can still carry the label, and so
		// already be present in raw: append only the IDs raw doesn't already
		// have, or it would be classified -- and scheduled for removal -- twice.
		seen := make(map[string]bool, len(raw))
		for _, ctr := range raw {
			seen[ctr.ID] = true
		}
		for _, ctr := range hookRaw {
			if !seen[ctr.ID] {
				raw = append(raw, ctr)
			}
		}
	}
```

glours (commented):
No blockers. Two should-fix items below (neither overlaps the prior Copilot/docker-agent threads on this PR), plus a few nits not worth their own inline comment.

---

### `pkg/compose/start.go:56-59` (not touched by this PR, so no inline anchor available)

This hand-rolled `ContainerList` doesn't filter on `ConfigHashLabel`, unlike `s.getContainers`/`getDefaultFilters`. That's the actual source of the hook-runner leak into `startService`/`waitDependencies` that `isNotHookContainer` now patches around at two separate call sites (`service_containers.go:176`, `:594`) — confirmed against a live daemon per `6a61f5680`'s commit message. Every other container this codebase creates stamps `ConfigHashLabel` (`create.go:264-266`); hook runners are the sole deliberate exception. Routing through the existing helper fixes it at the source instead of downstream, and matches the pattern already used at `containers.go:79` (`getContainersByService`):

```go
// current
res, err := s.apiClient().ContainerList(ctx, client.ContainerListOptions{
	Filters: projectFilter(project.Name).Add("label", oneOffFilter(false)),
	All:     true,
})
if err != nil {
	return err
}
containers := Containers(res.Items)

// proposed
containers, err := s.getContainers(ctx, project.Name, oneOffExclude, true)
if err != nil {
	return err
}
```

`isNotHookContainer` at the two `service_containers.go` call sites can stay as defense-in-depth, but becomes redundant after this.

---

Nits not called out individually: stale `HookLabel` doc comment in `pkg/api/labels.go` (still references `runPreStartHook`, split by this PR), a garbled doc comment on `isNotHookContainer`, duplicated `per_replica` validation between `runPreStart`/`planPreStartHookRunners` (both reference #14259 — a shared `hasPerReplicaHook` predicate would remove the duplication), and a `ConfigHashLabel` hash computed then immediately discarded in `createPreStartContainer`.

Otherwise: solid, well-tested piece of work.

pkg/compose/observed_state.go
ndeloof: Applied in df699dbbf, extracted as a mergeHookContainers helper (the inline gate pushed collectObservedState past the gocognit budget). One subtlety worth your confirmation: the gate is not strictly behavior-preserving — HookContainers is indeed only read by planPreStartHookRunners, but the classification loop also routes runners of a service REMOVED from the model into state.Orphans (TestCollectObservedState locks that: runner c6 of the model-absent service must reach Orphans, where --remove-orphans sweeps it). With the gate, that classification only happens while some remaining service still declares pre_start hooks; when none does, the leftover runner escapes `up --remove-orphans` and is cleaned by `down` instead (removePreStartHookContainers sweeps by project+hook label, no observed state involved). Accepted and documented in the helper's comment — the round-trip saving for the hook-less majority seems worth that corner. Tests adapted accordingly (fixtures now declare a hook where the listing must fire; the by-name-discovery helper expects a single ContainerList).

ndeloof:
Everything from the review is addressed in df699dbbf, each change amended into the commit owning the code it touches:
- **`start.go` listing** (body should-fix): now routes through `s.getContainers` — the ConfigHashLabel presence filter excludes hook runners at the source; the `isNotHookContainer` guards stay as defense-in-depth. Own commit, since start.go was previously untouched.
- **Hook listing gated** (inline thread): done as `mergeHookContainers`, with one documented corner — see the thread reply.
- **Nits**: `HookLabel` doc updated (createPreStartContainer, down's sweep); `isNotHookContainer` doc reworded; the duplicated per_replica validation is now one shared `perReplicaHookIndex` predicate used by both `runPreStart` and `planPreStartHookRunners`; and the computed-then-discarded ConfigHashLabel is gone — `createOptions` gained a `NoConfigHash` flag so `getCreateConfigs` never computes the hash for hook runners (replacing the post-hoc `delete`).

Unit tests, CI-pinned lint and the pre_start e2e scenarios are green.

docker-agent (commented):
### Assessment: 🟢 APPROVE

docker-agent (commented):
### Assessment: 🟢 APPROVE

pkg/compose/pre_start.go
```diff
@@ -75,50 +83,110 @@ func lowestNumberedContainer(containers Containers) container.Summary {
...
+	if err != nil {
+		return err
+	}
+	for i := range service.PreStart {
+		runner, ok := runners[i]
+		if !ok {
+			return fmt.Errorf("service %q pre_start[%d]: no hook runner container found — either its runner "+
+				"was already consumed by a previous start, or the service's container state changed after "+
+				"reconciliation planned this run; run %q to prepare fresh runners",
+				service.Name, i, "docker compose up "+service.Name)
```
docker-agent: **[low] Error hint wraps the entire `docker compose up <service>` command in `%q` double-quotes**

`%q` formats the concatenated string `"docker compose up "+service.Name` as a Go-quoted string, so the error message reads:

```
run "docker compose up web" to prepare fresh runners
```

A user who copy-pastes the quoted form literally (`"docker compose up web"`) will get a shell error on most shells because the whole string is treated as a single token. The intent is to show an executable shell command. Using a backtick pair instead makes the hint unambiguous and matches CLI documentation conventions:

```suggestion
			return fmt.Errorf("service %q pre_start[%d]: no hook runner container found — either its runner "+
				"was already consumed by a previous start, or the service's container state changed after "+
				"reconciliation planned this run; run `docker compose up %s` to prepare fresh runners",
				service.Name, i, service.Name)
```

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 70/100 |

ndeloof: Applied in 12099271b (amended into the owning commit `feat: reconciliation plan creates pre_start hook runners`): the hint now renders as `docker compose up <service>` in backticks instead of a Go-quoted string.

pkg/compose/executor_ops.go
```diff
@@ -161,6 +168,32 @@ func (exec *planExecutor) execRemoveContainer(ctx context.Context, op Operation)
...
+	exec.containersMu.Unlock()
+	if len(replicas) == 0 {
+		return fmt.Errorf("internal: no %q container to attach pre_start hook %d to", service.Name, op.HookIndex)
+	}
+	target := lowestNumberedContainer(replicas)
+	created, err := exec.compose.createPreStartContainer(ctx, exec.project, service, target, op.HookIndex, op.Name)
+	if err != nil {
+		return err
+	}
+	exec.pctx.set(node.ID, operationResult{
```
docker-agent: **[low] `execCreateHookContainer` stores its result in `pctx` but no downstream operation reads it**

`execCreateHookContainer` calls `exec.pctx.set(node.ID, operationResult{ContainerID: created.ID, ...})` after the hook runner container is created, but the stored value is never consumed by any downstream plan operation.

The `OpRunPreStart` node's `CreateNodeID` points to the **service container's** create node (not the hook runner's), and the start phase discovers runners at execution time via `listPreStartRunners` (an API round-trip). No other operation type retrieves a hook runner's ID from `pctx`.

**Impact**: The stored entry is dead code — a wasted map slot with no correctness consequence today. The risk is that a future maintainer wires a new operation to `pctx.get(hookRunnerNodeID)` believing the value will be there, only to find it never populated. Removing the `pctx.set` call (or adding a comment explaining why the result is deliberately not used by downstream nodes) would prevent that confusion.

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 60/100 |

pkg/compose/observed_state.go
```diff
@@ -140,6 +147,49 @@ type ObservedVolume struct {
...
+func (s *composeService) mergeHookContainers(ctx context.Context, project *types.Project, raw Containers) (Containers, error) {
+	hasPreStartHooks := false
+	for _, svc := range project.Services {
+		if len(svc.PreStart) > 0 {
+			hasPreStartHooks = true
+			break
+		}
+	}
+	if !hasPreStartHooks {
+		return raw, nil
```
docker-agent: **[low] `mergeHookContainers` skips the fetch when no service declares `pre_start` hooks, leaving orphaned runners invisible to `--remove-orphans`**

When a user removes **all** `pre_start` hook declarations from their `compose.yaml` and then runs `docker compose up --remove-orphans`, any hook runner containers left from a previous run are not cleaned up.

The reason: `mergeHookContainers` returns early (`!hasPreStartHooks → return raw, nil`) without fetching hook containers. Since hook runners deliberately carry no `ConfigHashLabel`, `getContainers` never returns them either. With nothing in `raw`, the classification loop never sees the runners, so they are never added to `state.Orphans`, and `--remove-orphans` cannot remove them.

The code comment at lines 157–161 of `observed_state.go` acknowledges this as an "accepted corner" and documents that `docker compose down` removes runners via hook label regardless. That mitigation is reliable, but users who remove hooks and subsequently run `up --remove-orphans` (a reasonable expectation after removing a feature) will find stale runners accumulating until they run `down`.

**Impact**: Leftover hook runner containers persist across `up --remove-orphans` invocations when all hooks are removed from the compose file. Severity is low given the `down` workaround, but the discrepancy between `--remove-orphans`'s promise and its actual behavior for this case is worth documenting more prominently (or unconditionally fetching hook containers at a small extra round-trip cost).

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 70/100 |

pkg/compose/reconcile.go
```diff
@@ -828,14 +859,77 @@ func (r *reconciler) reconcileService(service types.ServiceConfig) error {
...
+	slices.SortFunc(stale, func(a, b ObservedContainer) int { return strings.Compare(a.ID, b.ID) })
+	for i := range stale {
+		node := r.plan.addNode(Operation{
+			Type:          OpRemoveContainer,
+			ResourceID:    fmt.Sprintf("hook:%s:stale:%s", service.Name, stale[i].ID[:min(12, len(stale[i].ID))]),
+			Cause:         "stale pre_start hook container",
+			Service:       &serviceCopy,
+			Container:     &stale[i].Summary,
+			RemoveVolumes: true,
+			BestEffort:    true,
```
docker-agent: **[low] Stale-runner purge nodes carry no dependencies, allowing them to run concurrently with infrastructure operations**

In `planPreStartHookRunners`, stale hook runner containers are scheduled for removal via `r.plan.addNode(Operation{..., BestEffort: true}, "")` — the empty group and no `deps...` variadic means these purge nodes have zero `DependsOn` edges. The executor can schedule them at any time, including concurrently with network or volume recreation being performed for other services.

A stale hook runner that was previously connected to a network being concurrently torn down could generate a "network has active endpoints" conflict when Docker tries to remove it. The `Force: true` flag on `ContainerRemove` (in `execRemoveContainer`) does handle network disconnection, and `BestEffort: true` swallows any remaining errors, so the plan will still succeed. However, a failed stale purge leaves the old runner behind — meaning the subsequent `CreateHookContainer` operation will collide with the stale container's name (deterministic names prevent two containers sharing the same name) and fail, aborting the plan.

In practice, stale hook runners are in `created` or `exited` state and `Force: true` disconnects them before removal, so the race window is narrow. But adding `infraDeps` to the stale purge nodes (as the create nodes already receive them transitively via `deps`) would close the window and make the ordering guarantee explicit.

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 70/100 |

pkg/compose/pre_start.go
```diff
@@ -75,50 +83,110 @@ func lowestNumberedContainer(containers Containers) container.Summary {
...
+	runners := map[int]container.Summary{}
+	for _, ctr := range res.Items {
+		if ctr.State != container.StateCreated {
+			continue
+		}
+		index, err := strconv.Atoi(ctr.Labels[api.HookIndexLabel])
+		if err != nil {
+			continue
+		}
+		runners[index] = ctr
```
docker-agent: **[low] `listPreStartRunners` silently overwrites a runner if two `status=created` containers share the same `HookIndexLabel`**

The loop `runners[index] = ctr` does not check whether `runners[index]` is already populated. If two containers in `created` state carry the same `HookIndexLabel` value, the last one in the API response silently wins — `runPreStart` then executes whichever container the daemon happened to return last.

Under steady-state operation with current code this cannot happen: containers have deterministic names (`<project>-<service>-pre_start-<i>`) and Docker rejects a `ContainerCreate` that would produce a name conflict, so two `created` containers with the same index cannot coexist with the same name. However, a stale runner created by a **pre-deterministic-naming** compose version (no fixed name, same index label) could survive the best-effort purge (e.g., daemon temporarily unreachable) and then coexist with a freshly created runner under the new name. In that scenario, `runPreStart` starts the wrong container.

Adding a duplicate-index check (log a warning or return an error) would make unexpected daemon state visible rather than silently masking it:

```suggestion
		runners[index] = ctr
```

(A return-on-duplicate-index guard or a `logrus.Warnf` when `_, exists := runners[index]; exists` would surface the issue cleanly.)

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 60/100 |

docker-agent (commented):
### Assessment: 🟢 APPROVE

pkg/compose/executor_ops.go
ndeloof: Addressed in 6aeeb7d04 (amended into `feat: reconciliation plan creates pre_start hook runners`): the `pctx.set` now carries a comment stating it exists for uniformity with execCreateContainer (every create operation publishes its result) and that no plan operation consumes a hook runner's entry today — the start phase rediscovers runners through listPreStartRunners by design. Kept rather than removed so create operations stay uniform.

pkg/compose/observed_state.go
ndeloof: This is the trade-off deliberately documented in observed_state.go and already discussed in the human review round: fetching hook containers unconditionally would add a ContainerList round-trip to every up for a corner that `docker compose down` already covers via the hook label (regardless of hooks still being declared). `--remove-orphans` has never removed pre_start runners in the imperative engine either — they carry no ConfigHashLabel and were invisible to its listing too — so widening it here would exceed this PR's fidelity contract. Keeping as documented; a dedicated follow-up can revisit the `--remove-orphans` promise for hook runners across both engines.

pkg/compose/reconcile.go
ndeloof: Deliberate, and documented in planPreStartHookRunners' doc comment: purges are best-effort, mirroring the historical imperative purge which was warn-only and carried no ordering guarantee against infrastructure operations either. The create nodes do depend on the purges, so a genuinely stuck old runner surfaces deterministically as a name conflict on the create — exactly the visible failure mode you describe. Adding infraDeps to the purge nodes would introduce an ordering the imperative engine never had, which this PR's word-for-word fidelity contract avoids.

pkg/compose/pre_start.go
ndeloof: Applied in 6aeeb7d04 (amended into `feat: reconciliation plan creates pre_start hook runners`): listPreStartRunners now logs a warning when two created-state runners collide on the same HookIndexLabel (possible only with leftovers from a pre-deterministic-naming version), keeping the last-wins behavior but surfacing the unexpected daemon state.

pkg/dryrun/dryrunclient.go
```diff
@@ -170,15 +170,19 @@ func (d *DryRunClient) ContainerList(ctx context.Context, options client.Contain
...
-			return client.ContainerListResult{
-				Items: d.containers,
-			}, err
+			d.containers = res.Items
 		}
+		return client.ContainerListResult{
+			Items: d.containers,
+		}, nil
 	}
 	return d.apiClient.ContainerList(ctx, options)
```
docker-agent: **[high] `compose up --dry-run` fails for any service with `pre_start` hooks: `listPreStartRunners` is not routed in `DryRunClient.ContainerList` and falls through to the real daemon**

`DryRunClient.ContainerList` dispatches on the calling-function name, but the new `listPreStartRunners` caller introduced by this PR is not in the switch — it falls through to `d.apiClient.ContainerList`, which queries the real daemon. The real daemon has no hook runner containers (they were faked in `d.containers` by `ContainerCreate`), so `listPreStartRunners` returns an empty map.

`runPreStart` then errors on the very first hook:
```
service "web" pre_start[0]: no hook runner container found — ... run `docker compose up web` to prepare fresh runners
```
Before this PR, `runPreStart` created the hook container inline (via `createPreStartContainer`), so no `ContainerList` call occurred in the pre-start path during dry-run. This is a pure regression.

There are two issues that need to be fixed together:

1. **Routing**: add a `"listPreStartRunners"` case to `ContainerList` that filters `d.containers` by the hook labels (analogous to how `"start"` returns `d.containers` directly).

2. **State field**: `ContainerCreate` adds faked containers to `d.containers` without setting their `State` field (it defaults to `""`). `listPreStartRunners` filters by `ctr.State != container.StateCreated` and would silently discard every faked hook container even after the routing is fixed. Setting `State: containerType.StateCreated` on containers whose labels include `api.HookLabel` would close this second hole — or the routing case can skip the state re-check, since faked containers are always in `created` state by construction.

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 100/100 |

docker-agent (commented):
### Assessment: 🔴 CRITICAL

pkg/dryrun/dryrunclient.go
ndeloof: Both suggested fixes are applied in 173fe9b1a (amended into the owning commit `feat: reconciliation plan creates pre_start hook runners`): ContainerList now routes the `listPreStartRunners` caller to the faked-creations cache (filtered by the caller's label filters), and faked containers carry `State: created`.

One correction on the premise, verified by running both binaries: this is not a pure regression — `--dry-run up` with pre_start hooks already fails on pre-PR main (c6dffb26b) with `invalid container name or ID: value is empty (hook container retained for inspection)`, because the imperative runPreStartHook also delegates ContainerWait/ContainerLogs to the real daemon, which never saw the faked hook container. With this fix the plan-based flow now reaches that exact same pre-existing failure point (and names the container in the message). Making dry-run pre_start succeed end-to-end requires faking ContainerWait/ContainerLogs for cached containers — a pre-existing gap in the dry-run client, out of this PR's fidelity-to-the-imperative-engine scope; worth a dedicated follow-up.

docker-agent (commented):

pkg/compose/reconcile.go
```diff
@@ -828,14 +859,77 @@ func (r *reconciler) reconcileService(service types.ServiceConfig) error {
...
+// Purges are best-effort (mirroring the historical imperative purge, which was
+// warn-only) but the creates depend on them: runner names are deterministic,
+// so a genuinely stuck old runner surfaces as a name conflict on the create.
+// The creates also depend on every replica node and on the infrastructure, so
+// the executor resolves the target replica (VolumesFrom) against a final live
+// view with networks in place. When a running replica survives the plan, the
+// start phase skips pre_start entirely: nothing is created, and a hook
+// container retained after a failure stays available for inspection.
+func (r *reconciler) planPreStartHookRunners(service types.ServiceConfig, expectedScale int, keptRunning bool, containerNodes, infraDeps []*PlanNode) {
+	if len(service.PreStart) == 0 || expectedScale == 0 || keptRunning {
```
docker-agent: **[medium] `planPreStartHookRunners` early-returns before the stale-runner purge when a service removes all its `pre_start` hooks**

When a service drops all its `pre_start` declarations, `planPreStartHookRunners` fires the combined guard at line 890:

```go
if len(service.PreStart) == 0 || expectedScale == 0 || keptRunning {
    return   // ← stale purge loop never reached
}
// stale purge only runs below here
stale := slices.Clone(r.observed.HookContainers[service.Name])
for i := range stale { … }
```

**Trigger**: A multi-service project where service A still declares `pre_start` hooks and service B just removed all of its hooks. Because `mergeHookContainers` fetches **all** hook containers whenever `hasPreStartHooks` is true for **any** service, `r.observed.HookContainers[serviceB.Name]` is still populated with runners from the previous generation. `planPreStartHookRunners` returns early for service B before reaching the purge loop, so those stale runners are never scheduled for removal and accumulate until `compose down`.

**Why it matters**: The `per_replica` early-return path (line 909) was explicitly fixed to run the stale purge first and then return before the create loop — the same pattern is missing for the `len(service.PreStart) == 0` case.

**Fix**: separate the no-hooks guard from the stale purge:

```go
func (r *reconciler) planPreStartHookRunners(service types.ServiceConfig, expectedScale int, keptRunning bool, containerNodes, infraDeps []*PlanNode) {
    if expectedScale == 0 || keptRunning {
        return
    }
    serviceCopy := service
    deps := slices.Concat(containerNodes, infraDeps)
    // Always purge stale runners, even when the service no longer declares hooks.
    stale := slices.Clone(r.observed.HookContainers[service.Name])
    …
    if len(service.PreStart) == 0 || perReplicaHookIndex(service) >= 0 {
        return  // no creates needed
    }
    for i := range service.PreStart { … }
}
```

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 97/100 |

ndeloof: This mirrors the imperative engine exactly, and that is deliberate. At startService's call site (service_containers.go, pre-PR c6dffb26b) the whole pre_start flow — stale purge included — is gated by `len(service.PreStart) > 0 && len(serviceContainers) == len(toStart)`; `removeOrphanPreStartContainers` only ever runs inside `runPreStart`. So a service that dropped all its hooks never purged its leftover runners imperatively either: they accumulate until `docker compose down`, which removes them via the hook label regardless of declarations. The `len(service.PreStart) == 0` early return reproduces that gate word for word.

It is the same accepted corner already documented in observed_state.go and discussed on the mergeHookContainers thread — and the proposed always-purge would be inconsistent on its own terms: when no service declares hooks at all, mergeHookContainers doesn't fetch, so the purge loop would see nothing anyway; closing the corner for real requires the unconditional fetch already declined there. Both sides of this corner (observe + plan) are flagged for a dedicated follow-up on stale-runner cleanup beyond the imperative engine's behavior.

The per_replica purge-first path is different in kind: there the hooks are still declared, so the imperative engine did reach its purge before rejecting.

docker-agent (commented):
### Assessment: 🟢 APPROVE

pkg/compose/reconcile.go
ndeloof: Revisiting this after discussion: the imperative-parity defense was factually right but set the bar at the floor — this function's own per_replica path purges before returning for exactly the reason that applies here (a prior generation's runners must not accumulate until down), and deterministic purging is this PR's point. Fixed in 9c2a22dc8: the len(service.PreStart) == 0 clause is out of the guard — the purge now runs whenever scale > 0 and no running replica survives the plan, and the create loop naturally no-ops without hooks (no explicit second guard needed). When NO service declares hooks anymore, mergeHookContainers gates the listing off, so HookContainers is empty and this is a no-op — that outer corner keeps the documented trade-off (down sweeps by hook label). Covered by TestReconcileContainers_HooksRemovedStillPurgesStaleRunners (hook-less service alongside a hooked one: stale runner purged, nothing created); the test that locked the old behavior is removed.

pkg/compose/executor_ops.go
```diff
@@ -161,6 +168,36 @@ func (exec *planExecutor) execRemoveContainer(ctx context.Context, op Operation)
...
+// lowest-numbered replica matches the one the start phase hands the hooks.
+func (exec *planExecutor) execCreateHookContainer(ctx context.Context, node *PlanNode) error {
+	op := node.Operation
+	service := *op.Service
+	exec.containersMu.Lock()
+	replicas := slices.Clone(exec.containersByService[service.Name])
+	exec.containersMu.Unlock()
+	if len(replicas) == 0 {
+		return fmt.Errorf("internal: no %q container to attach pre_start hook %d to", service.Name, op.HookIndex)
+	}
```
docker-agent: **[medium] `execCreateHookContainer` may surface a misleading "internal: no container" error when a concurrent replica create fails**

When a replica create fails (non-BestEffort), the errgroup cancels the shared context, but its goroutine still calls `close(done[node.ID])` — which wakes up the hook-create goroutine that was waiting on that dependency channel. The hook goroutine then reaches `execCreateHookContainer` without any intervening `ctx.Done()` guard. At that point `containersByService[service.Name]` is empty (the failed create never appended to it), so line 182 fires:

```go
if len(replicas) == 0 {
    return fmt.Errorf("internal: no %q container to attach pre_start hook %d to", service.Name, op.HookIndex)
}
```

Go's errgroup returns the *first* non-nil error it receives. Because both goroutines are in flight concurrently, there is a small window where the hook goroutine's secondary error reaches errgroup before the replica goroutine's root-cause error, leaving the user with an opaque `internal:` message and no pointer to the actual failure.

A minimal defence is a `ctx.Err()` check at the top of `execCreateHookContainer` (before the empty-replicas guard), so the function returns the already-recorded cancellation reason and lets errgroup surface the upstream error:

```suggestion
func (exec *planExecutor) execCreateHookContainer(ctx context.Context, node *PlanNode) error {
	if err := ctx.Err(); err != nil {
		return err
	}
	op := node.Operation
	service := *op.Service
	exec.containersMu.Lock()
	replicas := slices.Clone(exec.containersByService[service.Name])
	exec.containersMu.Unlock()
	if len(replicas) == 0 {
		return fmt.Errorf("internal: no %q container to attach pre_start hook %d to", service.Name, op.HookIndex)
	}
```

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 65/100 |

docker-agent (commented):

pkg/compose/executor_ops.go
ndeloof: Applied in 5a8b4a549 (amended into the owning commit `feat: reconciliation plan creates pre_start hook runners`): execCreateHookContainer now bails out on ctx.Err() before the empty-replicas guard, with a comment noting that a dependency's done-channel also closes on failure. Worth noting the guard narrows the window rather than closing it — errgroup cancels the context only once the failing goroutine has returned, so the hook goroutine can still observe a nil ctx.Err() in between; the misleading `internal:` error remains possible in that residual sliver, but the common interleaving now surfaces the root cause.

pkg/compose/pre_start.go
```diff
@@ -173,19 +249,19 @@ func (s *composeService) runPreStartHook(
...
+// pre_start hook of service, named after getHookContainerName and stamped
+// with the hook labels so both the start phase (by index) and the purge (by
+// hook type) can find it. It only creates: the runner is left in created
+// state for the start phase to execute.
 func (s *composeService) createPreStartContainer(
 	ctx context.Context, project *types.Project, service types.ServiceConfig,
-	ctr container.Summary, hook types.PreStartHook,
+	ctr container.Summary, index int, name string,
 ) (client.ContainerCreateResult, error) {
+	hook := service.PreStart[index]
```
docker-agent: **[low] `createPreStartContainer` indexes `service.PreStart[index]` without a bounds check — potential panic**

`hook := service.PreStart[index]` performs a bare slice index with no guard. Today every caller is `planPreStartHookRunners`, which iterates `for i := range service.PreStart`, so `i` is always in bounds. However, the function's signature accepts an arbitrary `int` from any future caller, and a wrong index produces an unrecoverable panic rather than a wrapped error.

A one-line guard at the function entry makes the contract explicit and the failure mode graceful:

```suggestion
	if index < 0 || index >= len(service.PreStart) {
		return client.ContainerCreateResult{}, fmt.Errorf("service %q: hook index %d out of range (have %d hooks)", service.Name, index, len(service.PreStart))
	}
	hook := service.PreStart[index]
```

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 62/100 |

docker-agent (commented):
### Assessment: 🟢 APPROVE

pkg/compose/pre_start.go
ndeloof: Applied in 55b70c0ca (amended into the owning commit `feat: reconciliation plan creates pre_start hook runners`): createPreStartContainer now guards the hook index and returns a wrapped error instead of panicking on an out-of-range value.

pkg/dryrun/dryrunclient.go
```diff
@@ -170,15 +187,33 @@ func (d *DryRunClient) ContainerList(ctx context.Context, options client.Contain
 			Items: d.containers,
 		}, nil
 	case "getContainers":
+		// Seed the cache from the real daemon on first use, then keep
+		// answering from it: containers faked by ContainerCreate must stay
+		// visible to the start flow, which lists through getContainers.
 		if len(d.containers) == 0 {
```
docker-agent: **[medium] `DryRunClient.ContainerList` — `getContainers` cache guard `len(d.containers) == 0` conflates "never seeded" with "cache non-empty from a prior `ContainerCreate`"**

`getContainers` seeds `d.containers` from the real daemon only when the slice is empty.
This PR introduces `ContainerCreate` calls during the reconciliation plan phase (to fake
hook runner containers), which append to `d.containers` **before** the start flow calls
`getContainers`.

Trigger: `compose up --dry-run` on a project with `pre_start` hooks and at least one
service that was previously started (so the real daemon has existing containers).

Effect: when `start()` calls `getContainers()`, `len(d.containers) > 0` (it contains the
faked hook runner), so the guard fires and the real daemon is never queried.  The start
flow sees only the faked hook-runner containers — all real, pre-existing service containers
from the daemon are invisible.  Anything that depends on seeing real containers (e.g.
`waitDependencies`, container listings in the start phase) receives a truncated view.

The fix is to track seeding independently of the slice length — e.g. add a `containersSeeded bool`
field to `DryRunClient`, set it to `true` after the first daemon fetch, and change the guard from
`len(d.containers) == 0` to `!d.containersSeeded`.  The daemon result should then be merged via
`append(d.containers, res.Items...)` rather than assigned, so faked containers from `ContainerCreate`
are preserved alongside the real ones.

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 97/100 |

pkg/compose/reconcile.go
```diff
@@ -828,14 +859,79 @@ func (r *reconciler) reconcileService(service types.ServiceConfig) error {
...
+// Purges are best-effort (mirroring the historical imperative purge, which was
+// warn-only) but the creates depend on them: runner names are deterministic,
+// so a genuinely stuck old runner surfaces as a name conflict on the create.
+// The creates also depend on every replica node and on the infrastructure, so
+// the executor resolves the target replica (VolumesFrom) against a final live
+// view with networks in place. When a running replica survives the plan, the
+// start phase skips pre_start entirely: nothing is created, and a hook
+// container retained after a failure stays available for inspection.
+func (r *reconciler) planPreStartHookRunners(service types.ServiceConfig, expectedScale int, keptRunning bool, containerNodes, infraDeps []*PlanNode) {
+	if expectedScale == 0 || keptRunning {
```
docker-agent: **[medium] `planPreStartHookRunners` — `expectedScale == 0` early return bypasses the stale hook runner purge**

When a service is scaled to zero (`compose up --scale svc=0` or `deploy.replicas: 0`),
`planPreStartHookRunners` returns immediately before the stale-runner purge loop:

```go
if expectedScale == 0 || keptRunning {
    return            // ← purge loop never reached
}
stale := slices.Clone(r.observed.HookContainers[service.Name])
for i := range stale { … }   // scheduled here
```

Hook containers live in `r.observed.HookContainers`, not `r.observed.Containers`, so the
normal scale-down path (which removes service containers) never touches them.  Any hook
runner left behind by a previous failed run accumulates until `compose down` (which sweeps
by hook label) — a `scale=0` update alone is insufficient.

The asymmetry with the `per_replica` case is the tell: that path explicitly runs the purge
unconditionally and only returns before the **create** loop.  The `expectedScale == 0`
path skips the purge entirely without a comment explaining the design intent.

Fix: split the guard so the purge loop runs for scale=0 as well, and only the create loop
is gated:

```go
func (r *reconciler) planPreStartHookRunners(...) {
    if keptRunning {
        return
    }
    // purge stale runners unconditionally (scale=0 still needs cleanup)
    stale := slices.Clone(r.observed.HookContainers[service.Name])
    for i := range stale { … }
    if expectedScale == 0 || perReplicaHookIndex(service) >= 0 {
        return   // nothing to create
    }
    for i := range service.PreStart { … }
}
```

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 75/100 |

docker-agent (commented):

- [low] pkg/dryrun/dryrunclient.go:185 — `"start"` case in `ContainerList` is dead code after this PR routes `start()` through `getContainers` (confidence: moderate 67/100). The case can never match because `getCallingFunction()` resolves to `"getContainers"` (two frames up from `ContainerList`) when `start()` calls `getContainers()` which in turn calls `ContainerList`. Not posted inline because the anchor line is outside the diff hunks.

pkg/dryrun/dryrunclient.go
ndeloof: Hardened in c3c1c7e4d (amended into the owning commit `fix: start lists service containers through getContainers`): seeding is now tracked by a dedicated containersSeeded flag and the daemon result is merged with append, so a faked creation landing before the first listing can no longer mask the real containers. One precision on the trigger: in every current flow the first getContainers call happens at observe time (collectObservedState), before any plan execution, so the cache was always seeded before a fake create — the truncated view was not reachable today; the flag removes the implicit ordering assumption. The dead `"start"` case flagged in the review body is removed in the same commit — it became unreachable once start() routed through getContainers.

pkg/compose/reconcile.go
ndeloof: Applied in c3c1c7e4d (amended into the owning commit `feat: reconciliation plan creates pre_start hook runners`, like the analogous hooks-removed change): the guard is split — keptRunning still returns (runners retained for inspection), the stale purge now runs for scale=0 too, and only the create loop is gated on expectedScale == 0 / per_replica. Covered by the new TestReconcileContainers_ScaleZeroStillPurgesStaleRunners; the test locking the old kept-at-scale-0 behavior is removed, mirroring how the hooks-removed inversion was handled.

docker-agent (commented):
### Assessment: 🟢 APPROVE

- [low] pkg/e2e/hooks_test.go:121 — `TestPreStartHookRetryAfterFailureSucceeds`: missing `.Within()` timeout on `--wait` step (unverified / low severity)

  The step `ComposeCmd("up", "-d", "--wait")` at line 121 has no timeout bound. Every peer `--wait` call in the same file uses `.Within(60*time.Second)` (or `.Within(120*time.Second)` for builds). Without a bound, a slow CI agent, image-pull stall, or daemon regression can make this step block the test runner indefinitely. Suggest appending `.Within(60*time.Second)` for consistency.

docker-agent (commented):
### Assessment: 🟢 APPROVE

The reconciliation-plan approach for `pre_start` hook runners is architecturally sound. The deterministic naming, DAG ordering, stale-runner purge, and dry-run client routing are all well-structured. The 14 issues from previous review cycles have been cleanly addressed.

Two observations that did not reach the inline threshold (scored below 55/100) are noted below for awareness.

- [low] `pkg/dryrun/dryrunclient.go:204` — `getContainers` dry-run case returns all cached containers without real daemon filters, potentially leaking hook runners to `projectFromName` (confidence: weak 52/100)

  In production, `getContainers` filters by `api.ConfigHashLabel` presence, which hook runners deliberately omit — so they are invisible to the default listing. In the dry-run `getContainers` case the cache is returned verbatim, without applying those filters. The only caller that would be affected is `projectFromName` via the `project == nil` branch of `start()`, which is reached by `docker compose start --dry-run` (not `compose up --dry-run`, which passes a non-nil project). Since `start --dry-run` does not call `ContainerCreate` for hook runners, the cache would only contain real daemon containers or runners left by a prior failed `up`; `isNotHookContainer` guards the downstream start flow. The practical risk is narrow, but the structural asymmetry with production filtering is worth tracking.

- [low] `pkg/dryrun/dryrunclient.go:57` — `containersSeeded` and `d.containers` are accessed from concurrent errgroup goroutines without synchronisation (unverified low finding)

  `ContainerCreate` (called concurrently by the plan executor's errgroup) appends to `d.containers`, while `ContainerList`'s `getContainers` case reads `d.containersSeeded` and `d.containers` without a lock. The pre-existing `len(d.containers) == 0` guard had the same race surface; the new `containersSeeded` bool adds another field to it. In practice the plan executor seeds the cache at observe time before any concurrent creates run, so the window is narrow, but a `sync.Mutex` (or using `sync.Once` for seeding) would make the invariant explicit.

Issue #14224, opened 2026-09-15
When using docker compose OCI artifact, parameter --project-directory does not work for resolving relative volume paths


paulbourelly999:
### Description

I have the following [docker compose artifact](https://hub.docker.com/r/usdotfhwaops/v2xhub-application) I am attempting to run with this command.
```
docker compose \
               -f "oci://docker.io/usdotfhwaops/v2xhub-application:develop" \
               --env-file "/home/dev/cdasim-scenario-runner/tmp/.env.street_1" \
               -p "street_1" --project-directory "/home/dev/cdasim-scenario-runner/tmp/" \
               up -d --yes --force-recreate 
```

Under the /home/dev/cdasim-scenario-runner/tmp directory I have local configuration files that are intended to get volumed into my services using relative paths. One example is [here ](https://github.com/usdot-fhwa-OPS/V2X-Hub/blob/66fa36b3beaa6da4fc9695d14468f6582e475526/configuration/docker-compose.yml#L20) where I volume a .sql file into my mysql service to initialize my database from a relative path.

When running this however and inspecting the containers I see the following resolution for this volume path
```

       "Mounts": [
            {
                "Type": "bind",
                "Source": "/home/dev/.cache/docker-compose/f22c7d5e9d784c051618dd5ae2b75a2c8132cefffe0c9aaed666a0ae8f1ab702/v2xhub_db_init_street_1",
                "Destination": "/docker-entrypoint-initdb.d",
                "Mode": "rw",
                "RW": true,
                "Propagation": "rprivate"
            },
```

This --project-directory parameter works fine for local docker compose files and there is nothing in the documentation of the CLI parameter or the docker compose OCI artifacts that to me suggests that this should not work . Am I missing something?

### Steps To Reproduce

Run 
```
docker compose \
               -f "oci://docker.io/usdotfhwaops/v2xhub-application:develop" \
               --env-file "/home/dev/cdasim-scenario-runner/tmp/.env.street_1" \
               -p "street_1" --project-directory "/home/dev/cdasim-scenario-runner/tmp/" \
               up -d --yes --force-recreate 
```

### Compose Version

Docker Compose version v5.5.1

### Docker Environment

```Text
Client: Docker Engine - Community
 Version:           29.8.0
 API version:       1.56
 Go version:        go1.26.8
 Git commit:        88096ef
 Built:             Thu Sep  3 21:50:45 2026
 OS/Arch:           linux/amd64
 Context:           default

Server: Docker Engine - Community
[...]

PR #14226, opened 2026-09-16
e2e coverage for git and OCI remote stacks

ndeloof:

End-to-end coverage for remote compose sources: five scenarios deploying projects from git repositories and OCI artifacts, against fully hermetic in-process infrastructure.

## Context

The remote loaders are almost invisible to the e2e suite today: a single test consumes an OCI artifact — `config` output plus an `up` deliberately declined at the confirmation prompt — and nothing exercises the git loader at all. No test ever deploys a remote project, so the behaviors users actually rely on (relative files resolved against the fetched copy, branch and subdirectory selection, bundled env files, tag routing) are locked by nothing. Issue #14224 recently showed what lives in that blind spot.

## What the PR brings

Three git scenarios — deploy from the default branch with repository-relative `env_file`, select a `#branch`, select a `#ref:subdir` project past a decoy at the repository root — and two OCI scenarios: a full `up` from `oci://` consuming the bundled env-file layer (the first e2e that actually deploys an artifact), and tag selection between two published revisions.

The infrastructure stays hermetic. Git repositories are throwaway fixtures served over the **smart HTTP protocol by an in-process server**: `git http-backend` run as a CGI per `httptest` request — no daemon process, no container, no fixed port. Smart HTTP is a hard requirement, not a preference: the loader resolves the ref with `ls-remote` then shallow-fetches the raw commit, which the dumb protocol cannot serve (no shallow capability) and which needs `uploadpack.allowAnySHA1InWant`. OCI scenarios reuse the `TestPublish` pattern: publish to a throwaway local `registry:3`, consume with `--insecure-registry`. Every scenario isolates its remote-resource cache (`XDG_CACHE_HOME`), and the git helper neutralizes any developer configuration (`GIT_CONFIG_GLOBAL=/dev/null`) so signing or hook setups cannot leak into fixtures.

The scenario DSL gains the one primitive this needs: `FromRemote(source, rootFlags...)` switches the scenario's `-f` to a remote reference while the anchored testdata copy keeps serving as the local content the remote is built from — steps before the switch (the `publish`) still run against it. `ContainerEnv` joins the check vocabulary as the state-based observable that a remote project's bundled files were consumed.

Remote-source behavior is where reconciliation, path resolution and caching intersect (see #14224): before touching any of it, the intended behaviors need to be pinned. This suite establishes the harness those future fixes will extend — a regression test for #14224 becomes a one-scenario addition.


docker-agent (commented):
### Assessment: 🟢 APPROVE

pkg/e2e/remote_git_test.go
```diff
@@ -0,0 +1,142 @@
...
+// resulting repository over smart HTTP for the lifetime of the test.
+func serveGitRepo(t *testing.T, dir string) *gitRepo {
+	t.Helper()
+	gitPath, err := exec.LookPath("git")
+	if err != nil {
+		t.Skip("git is not available in PATH")
+	}
+	root := t.TempDir()
+	r := &gitRepo{t: t, work: dir, bare: filepath.Join(root, "repo.git")}
+	r.git(dir, "init", "-q", "-b", "main", ".")
```
docker-agent: **[medium] `git init -b main` requires git ≥ 2.28 — older versions silently ignore `-b`, breaking `TestGitRemoteSubdir`**

`serveGitRepo` uses `git init -q -b main .` to force the initial branch name to `main`. The `-b` flag was introduced in git 2.28.0 (2020-07). On a runner with git ≤ 2.27, the flag is silently ignored and git uses `init.defaultBranch` — which defaults to `master` on older installations.

This matters because `TestGitRemoteSubdir` references the branch explicitly with `#main:apps/web`. If the repo's head branch is actually `master`, compose's git loader will fail to resolve `main`, causing a confusing "branch not found" error with no indication that the git version is the culprit. The only skip guard today checks for git being absent entirely.

**Fix options:**
1. After `git init`, run `git symbolic-ref HEAD refs/heads/main` to rename the branch unconditionally (works on all git versions):
```suggestion
	r.git(dir, "init", "-q", ".")
	r.git(dir, "symbolic-ref", "HEAD", "refs/heads/main")
```

| Confidence | Score |
| :--: | :--: |
| 🟢 strong | 100/100 |

pkg/e2e/remote_git_test.go
```diff
@@ -0,0 +1,142 @@
...
+}
+
+// git runs a git command against a fully isolated configuration: no user or
+// system gitconfig (so a developer's signing or hook setup cannot leak into
+// the fixture) and a fixed identity.
+func (r *gitRepo) git(dir string, args ...string) {
+	r.t.Helper()
+	cmd := exec.Command("git", args...)
+	cmd.Dir = dir
+	cmd.Env = append(os.Environ(),
```
docker-agent, line 95: **[medium] `git()` appends env overrides to `os.Environ()` — first-match semantics on Linux mean pre-existing CI values shadow the `/dev/null` overrides**

`cmd.Env = append(os.Environ(), "GIT_CONFIG_GLOBAL=/dev/null", ...)` produces a slice where the original env vars come first and the isolation overrides come last. On Linux, glibc's `getenv()` — which git uses internally — returns the **first** match when a key appears multiple times. If the CI runner already exports `GIT_CONFIG_GLOBAL` (e.g. pointing at a signing config), the override at the tail of the slice has no effect: git reads the runner's config and the intended isolation breaks.

The same applies to `GIT_AUTHOR_NAME`/`GIT_COMMITTER_NAME` if the CI sets those via standard git env vars.

**Fix:** filter out the keys you're about to override before appending:

```suggestion
	env := slices.DeleteFunc(os.Environ(), func(e string) bool {
		return strings.HasPrefix(e, "GIT_CONFIG_GLOBAL=") ||
			strings.HasPrefix(e, "GIT_CONFIG_SYSTEM=") ||
			strings.HasPrefix(e, "GIT_AUTHOR_NAME=") ||
			strings.HasPrefix(e, "GIT_AUTHOR_EMAIL=") ||
			strings.HasPrefix(e, "GIT_COMMITTER_NAME=") ||
			strings.HasPrefix(e, "GIT_COMMITTER_EMAIL=")
	})
	cmd.Env = append(env,
```

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 75/100 |

pkg/e2e/remote_git_test.go
```diff
@@ -0,0 +1,142 @@
...
+	r.URL = server.URL + "/repo.git"
+	return r
+}
+
+// Branch publishes a variant of the fixture under a new branch: mutate edits
+// the working tree, and the resulting commit is pushed to the served
+// repository.
+func (r *gitRepo) Branch(name string, mutate func(dir string)) {
+	r.t.Helper()
+	r.git(r.work, "checkout", "-q", "-b", name)
```
docker-agent, line 79: **[medium] `Branch()` leaves the working tree on the new branch — a second `Branch()` call would branch from the wrong base commit**

After `git checkout -q -b <name>`, `r.work` stays on the newly created branch. The function returns without switching back to `main`. A single `Branch()` call per test works fine today, but any test that adds a second variant (e.g. two feature branches off `main`) would silently create the second branch from the tip of the first rather than from `main`, making it carry both sets of changes.

Add a `git checkout` back to `main` after the push so each `Branch()` call is independent:

```suggestion
	r.git(r.work, "checkout", "-q", "-b", name)
	mutate(r.work)
	r.git(r.work, "add", "-A")
	r.git(r.work, "commit", "-q", "-m", "branch "+name)
	r.git(r.work, "push", "-q", r.bare, name)
	r.git(r.work, "checkout", "-q", "main")
```

| Confidence | Score |
| :--: | :--: |
| 🟡 moderate | 70/100 |

docker-agent (commented):

3 findings in new test infrastructure. No issues with existing code.

pkg/e2e/remote_git_test.go
ndeloof: Fixed in d21fd0df6: git init then git symbolic-ref HEAD refs/heads/main, version-independent. (Nit on the finding: git < 2.28 actually errors on the unknown -b switch rather than silently ignoring it - the helper would have failed loudly via t.Fatalf - but the portability fix stands either way.)

pkg/e2e/remote_git_test.go
ndeloof: Fixed in d21fd0df6: Branch() checks main back out after the push, so successive calls cut from the same base.

pkg/e2e/remote_git_test.go
ndeloof: Not applicable to processes spawned through os/exec: exec.Cmd dedupes the environment before starting the child, keeping the LATER entry (dedupEnv, os/exec/exec.go: 'removed, in favor of later values'). Verified empirically: with GIT_CONFIG_GLOBAL=/ci/injected/config exported in the parent and cmd.Env = append(os.Environ(), "GIT_CONFIG_GLOBAL=/dev/null"), the child observes /dev/null. glibc first-match semantics never come into play because the duplicate never reaches the child's environment block. Keeping the idiomatic append.

docker-agent (commented):
This review covers only the commits pushed since 5109a754e8ef43e6cef6221eb9285e7f9a4a10a4.

### Assessment: 🟢 APPROVE

The two changes in this increment are both correct:

1. **Version-proof git fixture init** — replacing `git init -q -b main .` with `git init -q .` followed by `git symbolic-ref HEAD refs/heads/main` is the canonical, version-agnostic way to name an unborn branch. Git allows HEAD to be a symbolic ref pointing to a not-yet-existing ref (`refs/heads/main`), and the first commit creates that ref. The sequence (init → symbolic-ref → add → commit → clone) is correctly ordered. This directly addresses the finding from the previous review cycle.

2. **Branch cuts from main** — adding `r.git(r.work, "checkout", "-q", "main")` at the end of `Branch()` is safe: by that point `git add -A` and `git commit` have already run, so the working tree is clean and the checkout cannot fail on an unclean-tree basis. This ensures each `Branch()` call cuts from the same base commit, directly addressing the second finding from the previous review cycle.

No new issues found in the changed lines.

glours (approved):
LGTM

PR #14227, opened 2026-09-16
fix(up): drop benign context.Canceled noise from the final report

ndeloof:
This is a direct follow-up to #13990.

## Context

CI flaked right after #13990 merged: `TestUpAttachedTerminatesOnExternalStop` failed on the `standalone` job variants with `up` exiting 1 instead of 0, and this in the captured output:

```
Get "http://%2Fvar%2Frun%2Fdocker.sock/.../logs?follow=1&...": context canceled
```

Once the monitor detects the container's restart backoff was canceled by an external `stop`/`down` and returns, `up` cancels `globalCtx` to unwind its goroutines. The container's log-streaming goroutine (still blocked on `logs?follow=1` waiting for output that will never come) gets a `context.Canceled` error from that cancellation, and `appendErr` collected it unfiltered into the final `errors.Join`, turning what should be a clean, expected termination into a reported failure.

This could only surface once #13990 made the monitor loop actually return in this scenario — before that fix `up` hung forever and never reached this code path, so the race was unreachable.

## What this PR brings

`appendErr` now drops an error that is nothing more than fallout from `up`'s own shutdown (`errors.Is(err, context.Canceled)` once `u.globalCtx.Err()` is set) — the same treatment a SIGINT-triggered teardown's residual goroutine errors already get elsewhere. A genuine, unrelated error occurring after shutdown is still reported.


docker-agent (commented):
### Assessment: 🟢 APPROVE

glours (approved):
LGTM

